PR #270, opened 2015-08-03
Check whether files actually contain translatable strings.

dashea:
Instead of just checking whether gettext was used, use xgettext to look
for actual translatable strings before complaining that a file needs to
go in POTFILES.in.


vpodzime:
Looks good to me. Thanks for improving this!

dashea:
Pushed.

PR #271, opened 2015-08-03
More proxy tests

dashea:
The way we write proxy information to the .repo files is slightly different from the way we did it in yumpayload, so I wanted to make sure it still worked (it does).


vpodzime:
Looks good to me otherwise.

dashea:
Updated with the conditional in the %post scripts split up

vpodzime:
Looks even better to me now. Thanks!

dashea:
Pushed.

PR #273, opened 2015-08-03
Fix a line that still treated the fstype as a GtkComboBoxText

dashea:
Missed a conversion from ComboBoxText to ComboBox + ListStore on the
rhel7 side.

Resolves: rhbz#1248885


dashea:
Pushed.

PR #274, opened 2015-08-04
Fix crash when devices and connections are changing (#1245960)

jkonecny12:
First when user add new device to started anaconda in Welcome screen then
the anaconda will crash in NetworkSpoke initialization. This happens
because it's not created configuration file for the new device.
Fixed by generating this configuration in NetworkSpoke initialization.

Second when user remove existing ethernet connection for device. This could
happen specially in Live compose. This is fixed now that we skipping
devices without connection and add that device after the connection
appears.

*Resolves: rhbz#1245960*

### DIFF
--- a/pyanaconda/ui/gui/spokes/network.py
+++ b/pyanaconda/ui/gui/spokes/network.py
@@ -478,6 +478,12 @@ class NetworkControlBox(GObject.GObject):
         return True
 
     def initialize(self):
+        # There is a signal for newly added devices from NetworkManager but it
+        # is registered after the initialize method.
+        # So if someone adds a new device in the Welcome screen the ifconf file won't
+        # be created which causes anaconda to crash.
+        log.debug("Dump missing interfaces in NetworkControlBox initialize method")
+        network.dumpMissingDefaultIfcfgs()
         for device in self.client.get_devices():
             self.add_device_to_list(device)
 
--- a/pyanaconda/ui/gui/spokes/network.py
+++ b/pyanaconda/ui/gui/spokes/network.py
@@ -239,6 +239,17 @@ class DeviceConfiguration(object):
         elif con_uuid:
             self.device_type = self._setting_device_type(self.con_uuid)
 
+        # Found device for existing connection
+        if not self.device:
+            if self.device_type == NetworkManager.DeviceType.ETHERNET:
+                client = NMClient.Client.new()
+                dev_name = self.get_iface()
+                for device in client.get_devices():
+                    if dev_name == device.get_iface():
+                        self.device = device
+                        break
+
+        # Found connection for existing device
         if not self.con_uuid:
             if self.device_type != NetworkManager.DeviceType.WIFI:
                 uuid = nm.nm_device_setting_value(device.get_iface(), "connection", "uuid")
@@ -254,6 +265,11 @@ class DeviceConfiguration(object):
         return dev_type
 
     def get_iface(self):
+        """Get interface name
+
+           :return: Interface name or ``None`` if can't find device name for given uuid
+           :rtype: string or NoneType
+        """
         if self.device:
             iface = self.device.get_iface()
         else:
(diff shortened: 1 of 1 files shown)


pyanaconda/ui/gui/spokes/network.py
```diff
@@ -438,27 +443,36 @@ def add_connection_to_list(self, uuid):
...
         if dev_cfg.setting_value("connection", "read-only"):
             log.debug("network: GUI, not adding read-only connection %s", uuid)
             return False
         if dev_cfg.device_type not in self.supported_device_types:
             log.debug("network: GUI, not adding connection %s of unsupported type", uuid)
             return False
-        # Configs for ethernet has been already added,
-        # this must be some slave
-        if dev_cfg.device_type == NetworkManager.DeviceType.ETHERNET:
+        # skip slaves - only slaves has master
```
vpodzime: *have*^

vpodzime:

pyanaconda/ui/gui/spokes/network.py
jkonecny12: ahh thank you. I'll change this locally

pyanaconda/ui/gui/spokes/network.py
```diff
@@ -712,11 +726,19 @@ def add_device_to_list(self, device):
...
         except nm.SettingsNotFoundError:
-            # wireless devices
+            # wireless devices or device without a connection
             dev_cfg = None
+
         if dev_cfg:
             dev_cfg.device = device
         else:
+            # it's wireless or device with missing connection
+            # skip device with missing connection now, will be add when connection appears
```
bcl: 'skip the device'
'will be added'

pyanaconda/ui/gui/spokes/network.py
```diff
@@ -438,27 +443,36 @@ def add_connection_to_list(self, uuid):
...
         if dev_cfg.device_type == NetworkManager.DeviceType.WIFI:
             log.debug("network: GUI, not adding wireless connection %s", uuid)
             return False
+
         self.add_dev_cfg(dev_cfg)
         log.debug("network: GUI, adding connection %s", uuid)
         return True
 
     def initialize(self):
+        # There is signal for newly added devices from NetworkManager but this
```
bcl: 'is a signal'
'but it'

pyanaconda/ui/gui/spokes/network.py
```diff
@@ -438,27 +443,36 @@ def add_connection_to_list(self, uuid):
...
+        # is registered after the initialize method.
+        # So if someone add new device in Welcome screen ifconf file won't
```
bcl: 'adds a new device in the Welcome screen the ifconf'

pyanaconda/ui/gui/spokes/network.py
```diff
@@ -438,27 +443,36 @@ def add_connection_to_list(self, uuid):
...
+        # be created which causing anaconda to crash.
```
bcl: 'which causes'

bcl:
Looks ok to me, other than those subtle little grammar changes.

jkonecny12:
Ahh thank you @bcl for the grammar correction.
Sorry for that I have to look on a commit in more detail next time.

rvykydal:
I think we should find the NM device object and set device attribute of DeviceConfiguration when creating it in add_connection_to_list() for ETHERNET device case (ie the case when ethernet device is added). It seems the code assumes the DeviceConfiguration.device to be present for eth devices.

Also we should connect callbacks for the device in add_device_to_list, which does not happen due to earlier return before
        device.connect("notify::ip4-config", self.on_device_config_changed)
        device.connect("notify::ip6-config", self.on_device_config_changed)
        device.connect("state-changed", self.on_device_state_changed)
perhaps instead of the return we should just move #wireless device into else: branch

jkonecny12:
Updated PR.
I fixed the grammar and applied the @rvykydal changes.

pyanaconda/ui/gui/spokes/network.py
```diff
@@ -438,27 +454,36 @@ def add_connection_to_list(self, uuid):
...
             return False
+
         self.add_dev_cfg(dev_cfg)
         log.debug("network: GUI, adding connection %s", uuid)
         return True
 
     def initialize(self):
+        # There is a signal for newly added devices from NetworkManager but it
+        # is registered after the initialize method.
+        # So if someone adds a new device in the Welcome screen the ifconf file won't
```
vpodzime: *...the ifcfg file...* not *ifconf file* I believe.

vpodzime:
Looks good to me other than the nitpick above. But I'm leaving this to @rvykydal for a review as he knows these "waters" much better. :)

jkonecny12:
Thank you I'll change it locally.

jkonecny12:
Pushed

PR #276, opened 2015-08-04
Add a test for pre-install kickstart section and a fix for the bug it found.

bcl:


vpodzime:
Same suggestion as @dashea above, but otherwise this looks good to me. Good catch! I thought we covered all of these a while ago. Maybe this got in later?

bcl:
Fixed and pushed. There are a number of other cases of cat ... | grep that should also be cleaned up.

PR #277, opened 2015-08-04
Add test for liveimg kickstart command

bcl:
The environment needs to be setup with:
KSTEST_LIVEIMG_URL pointing to an image or .tar.gz file to install
KSTEST_LIVEIMG_CHECKSUM with the sha256 checksum of the image.


vpodzime:
Looks good to me.

clumens:
Looks good to me too.

bcl:
pushed

PR #278, opened 2015-08-05
Use someone else's code for PID file management.

dashea:
Someone else already solved this problem.

### DIFF
--- a/anaconda
+++ b/anaconda
@@ -45,7 +45,7 @@ if ("debug=1" in proc_cmdline) or ("debug" in proc_cmdline):
     cov.start()
 
 
-import atexit, sys, os, time, signal, stat
+import atexit, sys, os, time, signal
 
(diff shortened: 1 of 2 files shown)


bcl:
Absolutely!

anaconda
```diff
@@ -925,56 +925,14 @@ if __name__ == "__main__":
     # make sure we have /var/log soon, some programs fail to start without it
     iutil.mkdirChain("/var/log")
 
-    # Share these with the exit handler, installed later
-    pidfile_path = '/var/run/anaconda.pid'
-    pidfile_created = False
+    # Create a PID file. The exit handler, installed later, will clean it up.
+    import pid
```
vpodzime: Cannot agree more! But could we move this import next to the other "standard" (sys, os,...) ones?

PR #281, opened 2015-08-06
Use a disk label to find the filesystem for escrow results

dashea:
The escrow test does not use autopart, so the root filesystem does not
end up on /dev/mapper/fedora-root like most other tests. Use a label so
we don't have to depend on the partition order.


vpodzime:
Looks good to me. A clever change.

clumens:
If you get bored, there are several other tests that could use this same trick.

dashea:
Added label usage to btrfs-1 and 2 and raid-1

dashea:
Dang it, missed tmpfs. I'll push an update with that in a sec

clumens:
Yep, looks good to me too.

dashea:
Pushed.

PR #264, opened 2015-07-31
Add basic support for LVM cache creation in kickstart

vpodzime:
If user specifies we should create an LV as a cached LV, we need to transform
the parameters in to a CacheRequest instance and pass it to blivet to create a
new LV as cached.

This PR requires the [blivet's PR #190] (https://github.com/rhinstaller/blivet/pull/190) to be merged first and the appropriate bump of the required version of blivet will be done before this gets pushed.

PR #190, opened 2015-07-08
rhel7 backports

wgwoods:
This is the commits from #181 and #182, backported for rhel7. The differences are minor but _just barely_ enough to make me want another quick review.

The differences from the above commits are:

- 231ac70: also reverts a66cb15
- 99c97dd: dnfpayload vs. yumpayload; the fix is the same, it's just in a different file.
- d844db8: resolve some parse-kickstart conflicts - `commands.network.RHEL7_Network` vs. `commands.network.F22_Network`, etc.
- faf2432: resolve conflict in dracut/module-setup.sh: RHEL7 doesn't have `PYTHONHASHSEED=42`
- 679448a: RHEL7 handles chronyd and ntpd; rather than checking to see if the files exist, just try to modify them - the worst that can happen (AFAICT) is a warning in the log. NBD.

PR #282, opened 2015-08-06
Kickstart tests for the LVM cache kickstart support

vpodzime:
This adds two tests for the LVM cache kickstart support being added in PR #264 and [blivet's PR #190] (https://github.com/rhinstaller/blivet/pull/190). Please note that the test ``-2`` contains a commented-out check that reveals a minor bug in the LVM cache creation support which should be fixed and the check uncommented, but I don't want to hold everything off just because of that.


jkonecny12:
It seems good to me.

bcl:
Looks good to me too.

vpodzime:
Pushed.

PR #285, opened 2015-08-11
product.img buildstamp should override distribution buildstamp (#1240238)

bcl:
config.read() reads all the files in the order given, so later ones with
the same keys override the earlier ones. This changes the order so that
the product.img .buildstamp can be used for re-branding.

Resolves: rhbz#1240238


vpodzime:
Looks good to me.

bcl:
Pushed to all the places.

PR #286, opened 2015-08-12
Translate virtio vendor id into "Virtio Block Device"

bcl:
This will fix Fedora bug 1242117 and RHEL bug 1247861

### DIFF
--- a/pyanaconda/ui/lib/disks.py
+++ b/pyanaconda/ui/lib/disks.py
@@ -122,3 +122,21 @@ def checkDiskSelection(storage, selected_disks):
                            "unselected": ",".join(missing)})
 
     return errors
+
+def getDiskDescription(disk):
+    """ Return a string describing the disk.
+
+        :param :class:`blivet.devices.DiskDevice` disk: disk object
+        :returns: A string describing the disk
+        :rtype: str
+
+        This function tries to create a user friendly string describing
+        the disk. Usually this is made of up the vendor and model strings
+        from sysfs, with a few exceptions:
+
+        Virtio Block Device is used when the vendor is set to 0x1af4
+    """
+    if disk.vendor == "0x1af4":
+        return "Virtio Block Device"
+
+    return disk.description()
--- a/pyanaconda/ui/lib/disks.py
+++ b/pyanaconda/ui/lib/disks.py
@@ -45,7 +45,7 @@ class FakeDisk(object):
 
     @property
     def description(self):
-        return "%s %s" % (self.vendor, self.model)
+        return " ".join(s for s in (self.vendor, self.model) if s)
 
 def getDisks(devicetree, fake=False):
     if not fake:
@@ -139,4 +139,4 @@ def getDiskDescription(disk):
     if disk.vendor == "0x1af4":
         return "Virtio Block Device"
 
-    return disk.description()
+    return disk.description
--- a/pyanaconda/ui/gui/spokes/custom.py
+++ b/pyanaconda/ui/gui/spokes/custom.py
@@ -80,6 +80,7 @@ from pyanaconda import storage_utils
 from pyanaconda.ui.communication import hubQ
 from pyanaconda.ui.gui.spokes import NormalSpoke
 from pyanaconda.ui.helpers import StorageChecker
+from pyanaconda.ui.lib.disks import getDiskDescription
 from pyanaconda.ui.gui.spokes.lib.cart import SelectedDisksDialog
 from pyanaconda.ui.gui.spokes.lib.passphrase import PassphraseDialog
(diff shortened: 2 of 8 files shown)


vpodzime:
The changes look good to me, but I'm not really sure this should go into *anaconda*. Why not *blivet*? *blivet*'s ``DiskDevice`` already has the ``description`` property and doing the translation there would allow other user code to benefit from it.

jkonecny12:
[PR check](http://newcutlet.install.bos.redhat.com:8080/job/anaconda-master-PR/108/artifact/tests/nosetests.log) has failed on the new test. Please look on it before push.

bcl:
> The changes look good to me, but I'm not really sure this should go into anaconda. Why not blivet?

@dwlehman and I discussed this on IRC yesterday and decided that this is really a presentation issue that blivet doesn't want to deal with or maintain.

pyanaconda/ui/lib/disks.py
```diff
@@ -122,3 +122,21 @@ def checkDiskSelection(storage, selected_disks):
...
+        This function tries to create a user friendly string describing
+        the disk. Usually this is made of up the vendor and model strings
+        from sysfs, with a few exceptions:
+
+        Virtio Block Device is used when the vendor is set to 0x1af4
+    """
+    if disk.vendor == "0x1af4":
+        return "Virtio Block Device"
+
+    return disk.description()
```
dashea: This is the part jenkins is complaining about. Extra parenthesis?

dwlehman:
Doesn't this need a tweak to be able to handle nvme (and other non-scsi pci devices)?

A general solution that handles all of the foreseeable cases would probably be acceptable for blivet, but not if it's just encoding a partial pci ids table like this is doing.

bcl:
> Doesn't this need a tweak to be able to handle nvme (and other non-scsi pci devices)?

We don't have enough info about the nvme one yet, it may not be needed since it's new.

There is a pci id database provided by hwdata, in /usr/share/hwdata/pci.ids that we could use.

vpodzime:
> > The changes look good to me, but I'm not really sure this should go into anaconda. Why not blivet?
> 
> @dwlehman and I discussed this on IRC yesterday and decided that this is really a presentation issue that blivet doesn't want to deal with or maintain.

I'm sorry, but I don't really understand the reasoning here. From my point of view, this change/replacement/translation should be done as low as possible. It should ideally be fixed in kernel, but if that's not possible, it should be fixed in blivet. Otherwise every code using blivet (e.g. blivet-gui)  would have to implement something like this proposed function. parted did this for us, but if we no longer use parted, we need to do it. That would also keep things consistent.

vpodzime:
> A general solution that handles all of the foreseeable cases would probably be acceptable for blivet, but not if it's just encoding a partial pci ids table like this is doing.

Yes, such information should definitely be placed somewhere in a "declarative" way. At least in a dict outside of the function.

bonzini:
Checking the vendor only is not enough, for example 1af4:1004 is Virtio SCSI.

pci.ids has all the information, blivet should use it.

bcl:
I took a look at using the pci.ids file from hwdata today, but since /sys/block/* doesn't actually tell us the device id, just the vendor, there is no nice generic way to get from vendor id to 'virtio block device'.

bonzini:
Actually it does tell you the device id, but unfortunately the data you need (which corresponds to the _subsystem_ device id in the PCI device) is not yet in pci.ids.  So you need to handle PCI, virtio and SCSI all differently.

This needs to be applied to pci.ids for virtio subsystem ids to be recognized.

    --- pci.ids.old	2015-08-18 22:56:34.908591263 -0700
    +++ pci.ids	2015-08-18 23:00:23.010364669 -0700
    @@ -19121,14 +19121,23 @@
     # nee Qumranet, Inc.
     1af4  Red Hat, Inc
     	1000  Virtio network device
    +		1af4 0001  Virtio network device
     	1001  Virtio block device
    +		1af4 0002  Virtio network device
     	1002  Virtio memory balloon
    +		1af4 0005  Virtio network device
     	1003  Virtio console
    +		1af4 0003  Virtio network device
     	1004  Virtio SCSI
    +		1af4 0008  Virtio network device
     	1005  Virtio RNG
    +		1af4 0004  Virtio network device
     	1009  Virtio filesystem
    +		1af4 0009  Virtio network device
     	1010  Virtio GPU
    +		1af4 0010  Virtio network device
     	1012  Virtio input device
    +		1af4 0012  Virtio network device
     	1110  Inter-VM shared memory
     		1af4 1100  QEMU Virtual Machine
     1af5  Netezza Corp.

(Doesn't apply directly because github screws up TAB characters).

bcl:
pushed only to rhel7-branch.

PR #287, opened 2015-08-12
boot-options.rst: add a note about nfsiso

kparal:
Mention that nfs:<...>.iso can be used and is automatically detected.


vpodzime:
Looks good to me.

PR #288, opened 2015-08-12
Replace IPy package by ipaddress

jkonecny12:
ipaddress function is from python 3 standard library.
The IPy package requirement from anaconda was removed.


vpodzime:
Looks good to me too, but the ``check_ip_address`` is a good candidate for unit tests, I think.

jkonecny12:
The test was added. I hope this is enough.

vpodzime:
> The test was added. I hope this is enough.

Absolutely. Thanks!

jkonecny12:
Pushed.
Thank you

PR #289, opened 2015-08-12
Don't call setupDiskImages (#1252703)

bcl:
storageInitialize handles this itself, and if it is called first the
storageInitialize call resets the devicetree and doesn't re-setup the
images correctly.

Related: rhbz#1252703


jkonecny12:
Seems good to me

bcl:
No longer needed.

PR #290, opened 2015-08-13
boot-options.rst: add a note about hd iso source

kparal:
Mention that `inst.repo=hd:device:path.iso` can be used and is automatically detected.


kparal:
I think this commit is no longer needed at all.

PR #292, opened 2015-08-13
Fix LVM cache kickstart tests

vpodzime:
We should double quote variables in conditions to prevent syntax errors in case
something goes wrong, we cannot rely on the PV we expect to be 3100 MiB big to
have this exact size and we need to get rid of whitespace before comparing
strings. We can also use two disks instead of one to prevent issues caused by
everything hardly fitting into one 10GiB disk.

### DIFF
--- a/tests/kickstart_tests/lvm-cache-1.ks
+++ b/tests/kickstart_tests/lvm-cache-1.ks
@@ -26,65 +26,65 @@ shutdown
 
 %post
 root_lv="/dev/mapper/fedora-root"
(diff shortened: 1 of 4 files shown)


jkonecny12:
It works on my machine. 
Only problem is I needed ``device-mapper-persistent-data.x86_64 0.5.4-3.fc24`` package from rawhide because of some bug in the package.

clumens:
Ignore the bug in DNF in the test results, there.

vpodzime:
Pushed.

PR #293, opened 2015-08-13
network: compare with ssid bytes, not str

rvykydal:
followup of commit cf5bf6cfd65d0af06b9ef690016af87af9099f57


bcl:
Test failure is because it used the bad dnf, it can be ignored.

rvykydal:
Pushed

PR #294, opened 2015-08-13
Make sure proper installclass selected for RHEL installs.

bcl:
product.img is no longer being provided, and Fedora installclass
was being selected for installation. Bump the priority of the RHEL
installclass to ensure it gets selected instead.

Related: rhbz#1196721
(cherry picked from commit 859b10c8673052ca09169860ec1a5c161f46a1ac)


vpodzime:
Looks good to me too.

bcl:
pushed

PR #296, opened 2015-08-13
Write the empty dnf langpacks.conf to the right directory (#1253469)

dashea:


vpodzime:
Looks good to me too.

dashea:
Pushed.

PR #297, opened 2015-08-13
yum doesn't support nfs repos, don't write them to the target system.

bcl:


vpodzime:
These all look good to me.

bcl:
Pushed the "don't write" part of this to master for dnfpayload

bcl:
pushed

PR #298, opened 2015-08-14
Fix network startup when using kickstart from media

bcl:


vpodzime:
These both look good to me .

bcl:
Pushed to all the branches.

PR #299, opened 2015-08-14
Skip source url checks when network is off (#1251130)

bcl:
Otherwise you cannot exit the Source spoke to turn the network on.

Resolves: rhbz#1251130


vpodzime:
Looks good to me.

bcl:
Pushed to all the branches.

PR #301, opened 2015-08-17
Use the default mirrorlist instead of fixed repo URL in kickstart tests

vpodzime:
When we use a fixed repo URL we don't give DNF a chance to try multiple mirrors
if something cannot be downloaded resulting in tests failing because something
cannot be fetched from that particular URL. Mirrors are quite often broken
out-of-date, but by using multiple of them, we should be safe.


dashea:
Looks good, though I suspect this will force me to figure out how HTTPS proxies are supposed to work.

vpodzime:
New version leaving the proxy-* tests unchanged. We can change those later if it turns out to be worth the extra work.

vpodzime:
Pushed.

PR #302, opened 2015-08-17
Do not encode the geoloc timezone to bytes (#1240812)

dashea:
Yet string that ended up the wrong kind of string after switching to
python 3.


vpodzime:
Looks good to me.

dashea:
Pushed.

PR #209, opened 2015-07-15
fix storage writing for live and ostree installs (#1236937)

AdamWill:
3e800998 broke storage writing in two cases: regular live installs and ostree installs. The old write_storage_late check checked for flags.livecdInstall and ksdata.ostreesetup.seen as well as ksdata.method.method == "liveimg". The new check in writeStorageLate() dropped those two, only keeping the method check. The method is only 'liveimg' when doing a kickstart-driven install with the 'liveimg' parameter, which is a fairly niche case - the method is not 'liveimg' on a regular live install (it's 'harddrive').

PR #304, opened 2015-08-18
Add basic kickstart tests for LVM Thin Provisioning

vpodzime:
Two very simple kickstart tests for LVM Thin Provisioning. One with the very
basic setup using a single thin pool and two thin LVs all with fixed size plus
another one that tries to grow the thin pool to maximum possible size.

Let the fact that without [Blivet's PR #209] (https://github.com/rhinstaller/blivet/pull/209) these tests fail be the justification of their existence.

### DIFF
--- a/tests/kickstart_tests/lvm-thinp-1.ks
+++ b/tests/kickstart_tests/lvm-thinp-1.ks
@@ -1,0 +1,106 @@
+#version=DEVEL
+url --url="http://dl.fedoraproject.org/pub/fedora/linux/development/$releasever/$basearch/os/"
+install
+network --bootproto=dhcp
+
+bootloader --timeout=1
+zerombr
+clearpart --all --initlabel
+part /boot --fstype=ext4 --size=500
+part pv.1 --fstype=lvmpv --size=9004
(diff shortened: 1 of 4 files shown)


tests/kickstart_tests/lvm-thinp-1.ks
```diff
@@ -0,0 +1,106 @@
+#version=DEVEL
+url --url="http://dl.fedoraproject.org/pub/fedora/linux/development/$releasever/$basearch/os/"
```
bcl: You should use the mirrors here :)

tests/kickstart_tests/lvm-thinp-2.ks
```diff
@@ -0,0 +1,114 @@
+#version=DEVEL
+url --url="http://dl.fedoraproject.org/pub/fedora/linux/development/$releasever/$basearch/os/"
```
bcl: And here :) :)

vpodzime:
Ooops, caught myself in a loop. :) I'll fix the URL when pushing this.

vpodzime:
Pushed.

PR #305, opened 2015-08-18
Defer to Fedora distro-wide settings for password strength (#1250746)

dashea:
Remove the minlen and minquality settings from the default pwpolicy,
and instead pass the buck to libpwquality to determine what is and is
not an acceptable password.

### DIFF
--- a/data/interactive-defaults.ks
+++ b/data/interactive-defaults.ks
@@ -5,7 +5,7 @@ firstboot --enable
 
 %anaconda
 # Default password policies
-pwpolicy root --strict --minlen=8 --minquality=50 --nochanges --emptyok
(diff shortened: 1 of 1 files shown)


dashea:
Skimmed over the part of the bug about "The user should be able to override the exception and use the poor password if they wish," so --strict needs to go, too.

dashea:
While removing the minimum length and strength requirements is relatively inoffensive, since we depend on pwquality for this information anyway, I'm not interested in Fedora's removal of --strict being in upstream anaconda. Closing this PR, and tomorrow and I'll work on patches to the individual product.img packages to put the desired configuration into Fedora, where it belongs.

PR #306, opened 2015-08-18
Fix the single-spoke TUI message for Python 3.

dashea:
dict_values objects can't do much, so convert it to something that can.


vpodzime:
Looks good to me.

dashea:
Pushed.

PR #308, opened 2015-08-18
Add missing requirements

atodorov:
There are some missing requirements needed before you can build or test anaconda. So far the best place for them seems in Makefile. 

gettext-devel and libtool are required before running ./autogen.sh. The rest are required before make ci.

Btw I'm planning to add some documentation about building and testing anaconda and the first thing will be
```
make -f Makefile.am install-buildrequires
make -f Makefile.am install-requires
./autogen.sh
./configure
```

so it's safe that these dependencies live in the makefile.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -117,7 +117,7 @@ install-buildrequires:
 	if ! [[ $$(uname -m) =~ s390x? ]]; then \
 		pkglist=$$(echo "$$pkglist" | grep -v s390utils) ; \
 	fi ; \
-	dnf install $$pkglist
+	dnf install $$pkglist gettext-devel libtool
 
 # Install all packages specified as Requires in the Anaconda specfile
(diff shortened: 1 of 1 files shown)


bcl:
I think packages needed for testing, rc-release and ci should have their own target in the Makefile, install-test-requires maybe. the requires and buildrequires should be left clean, fed only by the spec file.

atodorov:
@bcl - how about gettext-devel and libtool then ? autogen.sh complains if those two are missing.

Should we add them to the spec file ?

atodorov:
Also I' in favor of making the new instll-test-requires target depend on the other two so we can install everything in one go.

clumens:
gettext-devel and libtool aren't BuildRequires because the tarball we ship as part of an SRPM will already have had ./autogen.sh run.  But yeah it would be handy to have them installed somewhere.

PR #310, opened 2015-08-19
Move locale environment logic into localization.py

dashea:
I'd also like this on master+f23, but I have a few more patches there for some python3 string encoding crud, so I'll just include the patch again in that PR.


dashea:
initial-setup wasn't on the email hook until just now, so for anyone reading on anaconda-patches, the initial-setup half of this is at https://github.com/rhinstaller/initial-setup/pull/2

vpodzime:
Looks good to me.

PR #311, opened 2015-08-19
Enable test coverage in CI

atodorov:
The important files resulting from this are .coverage and tests/coverage-report.log. 

In case the ci target fails there's a coverage-report target so you can combine the data files and do the reports manually.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,0 +1,5 @@
+# http://nedbatchelder.com/code/coverage/config.html
+
+[run]
+branch = True
+parallel = True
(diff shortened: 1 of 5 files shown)


atodorov:
See:
http://nedbatchelder.com/code/coverage/subprocess.html
https://docs.python.org/3/library/site.html


This is the way to instruct Python to start coverage recording before it loads any other processes. The user site-packages directory is the least intrusive way I've found. I'm not sure if it is possible to do the same with a wrapper script and configuring PYTHON=ourwrapper instead.

PR #312, opened 2015-08-20
Fix inst.repo=hd: is not working (#1252902)

jkonecny12:
isys method isIsoImage has problem with comparision of bytes with the
str. Another Python 3 problem.

Resolves: rhbz#1252902


vpodzime:
Looks good to me.

jkonecny12:
Pushed

PR #313, opened 2015-08-20
Fix crash when anaconda starts with repository on HDD

jkonecny12:
inst.repo=hd:LABEL= method with repositories on HDD (not ISO file)
is unsupported method but the anaconda do not need to crash anyway.


jkonecny12:
Closing. This needs to be taken care of by another way.

PR #314, opened 2015-08-20
Add test documentation

atodorov:
This is the start of what is supposed to be a human readable test documentation of the Anaconda test suite. Filing here for comments. 

I'd like to also include API docs for the Python files under tests/ but so far I'm having troubles building these. I will likely have to add __init__.py files to all directories that are missing them and fix the import failures from Sphinx (e.g. by altering PYTHONPATH if necessary).

Btw it looks like the generated HTML is under version control, not sure if that needs to be removed.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -21,6 +21,12 @@ import sys, os
 # documentation root, use os.path.abspath to make it absolute, like shown here.
 sys.path.insert(0, os.path.abspath('..'))
 
+# configuration required to import test modules
+for path in ["../pyanaconda/isys/.libs", "../pyanaconda", "../tests", "../tests/lib", "../dracut", "../widgets"]:
+    sys.path.append(os.path.abspath(path))
+if not 'ANACONDA_INSTALL_CLASSES' in os.environ:
+    os.environ['ANACONDA_INSTALL_CLASSES'] = os.path.abspath('../pyanaconda/installclasses')
+
 # -- General configuration -----------------------------------------------------
 
 # If your documentation needs a minimal Sphinx version, state it here.
--- a/.gitignore
+++ b/.gitignore
@@ -100,3 +100,4 @@ _sources
 modules.rst
 pyanaconda.*rst
 .coverage
+docs/tests/*.rst
--- a/docs/Makefile.am
+++ b/docs/Makefile.am
@@ -73,6 +73,8 @@ apidoc:
 	if [ -f /usr/bin/$(SPHINXAPIDOC) ]; then \
 	    rm $(addprefix ./, $(MODULE_NAMES)) ; \
 	    $(SPHINXAPIDOC) -o . $(SOURCEDIR) ; \
+	    rm $(addprefix ./, tests/*.rst) ; \
+	    $(SPHINXAPIDOC) -o ./tests ../tests ../tests/storage/run_storage_tests.py ; \
 	fi
 
 html: apidoc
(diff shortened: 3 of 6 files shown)


jkonecny12:
I like this, good work.
Seems good to me but this will be better if someone who can check grammar and have better understanding of the tests look on it too.

atodorov:
Can you comment on the inclusion of test API docs and the changes required to do so? Is this acceptable ?

jkonecny12:
I fear I can't help you much with this. I think this question is more for @bcl.

atodorov:
Update:

* The added init files are required for Sphinx to be able to discover all tests. They don't affect test results;
* The config changes in docs/conf.py are required so that we can import all tests without failures except pylint
```
************* Module ../docs/conf.py
W9940(environment-modify): 28,4: : Found potentially unsafe modification of environment
************* Module ../tests/gui/inside/__init__.py
/usr/lib/python3.4/site-packages/astroid/scoped_nodes.py:425: PyGIWarning: Atspi was imported without specifying a version first. Use gi.require_version('Atspi', '2.0') before import to ensure that the right version gets loaded.
  return MANAGER.ast_from_module_name(absmodname)
************* Unused False Positives Found:
^F0401.*: Unable to import 'dogtail.*'$
```
* Currently the building of the docs fails because of Dogtail for Python3
  - it is still not officially available in Rawhide (I've tested with a manually build package)
  - it fails because assistive technologies are disabled. Even with a patch to skip aborting it still fails because I don't have at-spi-bus-launcher running.

docs/conf.py
```diff
@@ -21,6 +21,12 @@
 # documentation root, use os.path.abspath to make it absolute, like shown here.
 sys.path.insert(0, os.path.abspath('..'))
 
+# configuration required to import test modules
+for path in ["../pyanaconda/isys/.libs", "../pyanaconda", "../tests", "../tests/lib", "../dracut", "../widgets"]:
```
vpodzime: I'm not sure if these "raw" (non-prefixed)  relative paths are a good way to go. Or do they work the same no matter which directory you run the sphinx stuff from?

vpodzime:
Looks good to me otherwise.

docs/conf.py
atodorov: Works even from the top level directory.

bcl:
Yes, including the generated docs is on purpose. I do a separate run to generate them and copy them over to the gh-pages branch so they'll show up on github. I'm all for documenting the test procedures, but I'm not sure how useful it is to actually document the test code with sphinx. It generally isn't very complicated and if it is going to cause problems with needing packages like dogtail and at-spi-bus-launcher it isn't worth it to block generating the rest of the docs.

atodorov:
@bcl - I'm removing the gui/ tests documentation for now b/c of the issues with Dogtail and b/c it's not clear if these tests will be available much longer.

bcl:
> Btw I see the following issues if I don't have X running:

Yeah, GTK doesn't really like sphinx on a non X system. I build the docs with my dev system and the helper script passes $DISPLAY through so that it doesn't complain.

atodorov:
@bcl - fixed and slightly updated the docs. Please review and merge if no objections.

atodorov:
ping, any update on this one ?

vpodzime:
Still looks good to me.

bcl:
Thanks, tested and added some new modules to make-sphinx-docs.

PR #315, opened 2015-08-20
Do not use libuser

dashea:
There are a couple other pieces needed before this can be merged. There's a bug in shadow-utils (https://bugzilla.redhat.com/show_bug.cgi?id=1255175) where useradd reads login.defs before calling chroot, and we need it to do the opposite of that, like the other tools do. Also the utilities need to be added to the stage2 in lorax or else cloud people are going to complain at us that creating a users crashes the install, and I haven't written that change yet.

### DIFF
--- a/pyanaconda/users.py
+++ b/pyanaconda/users.py
@@ -19,14 +19,14 @@
 # Author(s): Chris Lumens <clumens@redhat.com>
 #
 
-import libuser
 # Used for ascii_letters and digits constants
 import string # pylint: disable=deprecated-module
 import crypt
 import random
-import tempfile
 import os
 import os.path
+import subprocess
+from contextlib import contextmanager
 from pyanaconda import iutil
 import pwquality
 from pyanaconda.iutil import strip_accents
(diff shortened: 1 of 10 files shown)


dashea:
Also, the test will need to be run manually by jenkins, like it does for the storage test.

dashea:
> Also, the test will need to be run manually by jenkins, like it does for the storage test.

oh wait there's a makefile target for that. I'll push the make ci change after lunch.

dashea:
Rebased, resolved conflicts, and added the COVERAGE_PROCESS_START value to the new make check line.

pyanaconda/users.py
```diff
@@ -286,179 +251,166 @@ def createUser(self, user_name, *args, **kwargs):
...
+            os.mkdir(sshdir, 0o700)
+            iutil.eintr_retry_call(os.chown, sshdir, int(uid), int(gid))
+
+        authfile = os.path.join(sshdir, "authorized_keys")
+        authfile_existed = os.path.exists(authfile)
+        with iutil.open_with_perm(authfile, "a", 0o600) as f:
+            f.write(key + "\n")
+
+        # Only change ownership if we created it
+        if not authfile_existed:
```
jkonecny12: Maybe I missed something but I think we want to change ownership only when the out file "is" created.

jkonecny12:
Other than the one above this seems good to me.

pyanaconda/users.py
vpodzime: "is created" means "didn't exist" means "not existed" ;)

vpodzime:
Other than the comments/questions above this looks good to me.

pyanaconda/users.py
jkonecny12: Yeah I see it now as I said I missed something :D :D .

dashea:
Switched to keyword docs, removed mkmailspool, added a comment for if parent_dir part, replaced the _exit with an exception.

Also, it's a good thing I looked at the docs, because I messed up the handling of password. The empty string is an empty password, no (None, False) password is locked. Fixed that and updated the test to handle and check for it.

vpodzime:
Looks great to me now, thanks!

dashea:
One more try with fixes for problems found while testing:

I had changed some kwargs.get("whatever") calls to if "whatever" in kwargs, which often does not work because pykickstart sets default values for all that stuff. I changed the kwargs processing back, with the exception of the uid/gid stuff which I changed from

```
if kwargs.get("gid", -1) >= 0:
```

to

```
if kwargs.get("gid") is not None:
```

because I think using -1 as a sentinel value looks dumb, and the default value from pykickstart is going to be None and not a weird special int.

Also added checks in the user create test to make sure the default values aren't messed up.

vpodzime:
Still looks great to me.

dashea:
Pushed to master, leaving the PR open so I remember to push to F23 once the shadow-utils update (https://bodhi.fedoraproject.org/updates/FEDORA-2015-14416 ) goes to stable.

dashea:
Pushed to f23.

PR #316, opened 2015-08-20
Fix I/O issues when anaconda is started without a locale.

dashea:
Fun fact: python 3 lets input() crash with UnicodeDecodeError.

### DIFF
--- a/pyanaconda/localization.py
+++ b/pyanaconda/localization.py
@@ -27,6 +27,8 @@ import langtable
 import locale as locale_mod
 import glob
 from collections import namedtuple
+import sys
+import io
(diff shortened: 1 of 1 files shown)


vpodzime:
WTH, python 3?! I'm starting to think that python people should really get a huge badge for an incredible achievement -- doing unicode support in a bad way; twice!

vpodzime:
The change looks good to me.

jkonecny12:
Looks good to me too.
I think this should go to F23 branch too so I'm adding labels.

dashea:
Pushed to master and f23

PR #318, opened 2015-08-21
Exclude what we don't need in coverage reports

atodorov:
This produces better reports.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -3,3 +3,4 @@
 [run]
 branch = True
 parallel = True
+source = ..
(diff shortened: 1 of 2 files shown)


vpodzime:
Looks good to me.

atodorov:
This update adds the source parameter to coverage and changes the report parameters. The net result is that  we get a report on all files in the source dir, even ones which are never imported from the test suite, hence not covered at all so we know where to look next.

clumens:
Applied and pushed.  Thanks.

PR #320, opened 2015-08-21
sys.std* encoding, again

dashea:
I tested the crazy idea of detaching and reattaching std* using the TUI, and it worked fine there, but today a bunch of debug prints I was trying to add stopped showing up after the welcome spoke. Maybe this wasn't such a great idea.


vpodzime:
Let's keep our fingers crossed. Looks good to me.

dashea:
Pushed.

PR #321, opened 2015-08-21
Bad continue clicks

dashea:
Fix some issues where the continue button can be clicked when it shouldn't.

The first commit will be Resolves: rhbz#1255752. The second issue doesn't seem to be an issue on rhel7.


vpodzime:
These both look good to me.

dashea:
Pushed.

PR #322, opened 2015-08-24
Run AnacondaExceptionHandler in cmdline mode (#1256385)

bcl:
cmdline mode still needs to have the exception handler run. It exits
anaconda when there are problems, without it you can get stuck in a loop
in the summary hub when there are incomplete spokes.

This adds an interactive argument to it so that when the handler runs in
cmdline mode it only prints out the error and calls sys.exit(1) --
skipping the call to the superclass from meh which tries to run a UI to
prompt for debugging.

(cherry picked from commit 98ec24be63dcc66de695dbfc86ef0fb08ab31f01)

Resolves: rhbz#1256385

### DIFF
--- a/anaconda
+++ b/anaconda
@@ -1205,10 +1205,8 @@ if __name__ == "__main__":
     else:
         cleanPStore()
 
-    # only install interactive exception handler in interactive modes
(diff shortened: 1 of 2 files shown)


vpodzime:
Looks good to me.

bcl:
pushed

PR #323, opened 2015-08-25
Fix first run environment setup in software spoke

jkonecny12:
Without this change the anaconda will raise exception NoSuchGroup for
environment ``None``.

Thank you for helping me to find this @M4rtinK .

### DIFF
--- a/pyanaconda/ui/gui/spokes/software.py
+++ b/pyanaconda/ui/gui/spokes/software.py
@@ -396,7 +396,11 @@ class SoftwareSelectionSpoke(NormalSpoke):
             elif firstEnvironment:  # manual installation
                 # for manual installs that don't have a default provided by the install class
                 # just tick the first radio button and select the first environment
-                if not self.environment_valid:
+                #
+                # None indicates that an environment has not been set, which is a valid
+                # value of the environment variable.
+                # Only non existing environments are evaluate as invalid
+                if not self.environment_valid or self.environment is None:
                     radio.set_active(True)
                     self.environment = environmentid
                 firstEnvironment = False


pyanaconda/ui/gui/spokes/software.py
```diff
@@ -396,7 +396,7 @@ def refresh(self):
             elif firstEnvironment:  # manual installation
                 # for manual installs that don't have a default provided by the install class
                 # just tick the first radio button and select the first environment
-                if not self.environment_valid:
+                if not self.environment_valid or self.environment is None:
```
vpodzime: Am I wrong or does this mean that ``self.environment_valid`` is ``True`` for ``self.environment`` being ``None``? That seems at least really unfortunate.

M4rtinK: The idea is that None indicates that an environment has not been set, which is a valid value of the environment variable (which is just wrapping the environment variable from kickstart).
Existing environments (like for example @^infrastructure-server-environment) also evaluate as valid. Non existent environments (@^foo-bar) on the other hand evaluate as invalid.

Previously we didn't really do any environment validation at all and a wrong environment specification in kickstart was enough to crash the installation, so I've (among other changes) added this property.

I do agree that environment being None could be rather unintuitive and I'm open to improvement ideas. Maybe we can add a "environment_set" property that indicates that the environment has been set to some (not necessarily valid) value ? On the other hand the software spoke is already quite complex with all its various states and transitions.

vpodzime: Thanks for the explanation. I think a comment right before the condition explaining what ``self.environment`` being ``None`` means would do.

jkonecny12:
Updated:
Add comment from the @M4rtinK response to the condition.

pyanaconda/ui/gui/spokes/software.py
```diff
@@ -396,7 +396,11 @@ def refresh(self):
             elif firstEnvironment:  # manual installation
                 # for manual installs that don't have a default provided by the install class
                 # just tick the first radio button and select the first environment
-                if not self.environment_valid:
+                #
+                # None indicates that an environment has not been set, which is a valid
+                # value of the environment variable.
+                # Only non existing environments are evaluate as invalid
```
vpodzime: *"evaluated"* as valid

vpodzime:
Looks good to me now.

AdamWill:
+1, this seems to restore the correct behaviour for interactive install in Rawhide.

jkonecny12:
Pushed. I'm going to set the bug to modified.

PR #325, opened 2015-08-26
don't crash if no environment set in interactive (#1257036)

AdamWill:
634d2d9 fixed the kickstart case but broke the interactive
'nothing selected' case. Since we want to return "Nothing
selected" for None regardless of whether we're kickstarted or
interactive, just move the 'if not self.environment' check
out of the 'if flags.automatedInstall' block so it applies to
both.

### DIFF
--- a/pyanaconda/ui/gui/spokes/software.py
+++ b/pyanaconda/ui/gui/spokes/software.py
@@ -306,11 +306,15 @@ class SoftwareSelectionSpoke(NormalSpoke):
             elif not self.environment:
                 return _("Nothing selected")
 
-        if not flags.automatedInstall and not self.environment_valid:
-            # selected environment is not valid, this can happen when a valid environment
-            # is selected (by default, manually or from kickstart) and then the installation
-            # source is switched to one where the selected environment is no longer valid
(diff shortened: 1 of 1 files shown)


AdamWill:
whoops, first version wouldn't have been right, this should be better.

jkonecny12:
I had already created and Acked [this PR](https://github.com/rhinstaller/anaconda/pull/323) which was created with @M4rtinK.
Question is which one we will use.

AdamWill:
Yours looks better - as I mentioned in the bug, mine doesn't actually restore intended behaviour, only makes the crash go away. I'll check yours behaves as intended and ack it if so.

AdamWill:
In fact, perhaps we want to apply both? It usually shouldn't be possible to reach the 'None' state in interactive, I don't think, but it may be better to handle it than crash if somehow we do.

M4rtinK:
Yeah, even though with the patch from @jkonecny12 environment should never be None it does not hurt making it a bit more robust.

AdamWill:
In fact this seems to have been merged already as 3893e45a38b412ec4fb570970528fba69131d546 .

PR #327, opened 2015-08-26
Update partitioning and timeout configuration for gui tests

atodorov:
* I've bumped the size of the root partition but it may be too much; 
* updated the sed regex b/c the underlying kc configs in spin-kickstarts have probably changed. Without this change we end up with 3 root partitions in livecd.ks
* bumped the timeout value b/c it takes about 2:30 hrs only to perform the install.

### DIFF
--- a/tests/gui/anaconda-autogui-testing.ks
+++ b/tests/gui/anaconda-autogui-testing.ks
@@ -10,7 +10,7 @@ rootpw qweqwe
 bootloader --location=mbr
 zerombr
 clearpart --all
-part / --fstype="ext4" --size=6000
+part / --fstype="ext4" --size=10240
 
 %post
 cat >> /etc/rc.d/init.d/livesys << EOF
--- a/tests/gui/make_livecd.sh
+++ b/tests/gui/make_livecd.sh
@@ -54,7 +54,7 @@ REPO="$3"
 # (1) Remove the first two / partitions we inherit.
 # (2) Remove rawhide as a repo because it's already the installation source.
 # (3) Don't remove /boot/initramfs*.  Do what now?
-ksflatten -c anaconda-autogui-testing.ks | sed -e '\|part /.*--size=3.*|,+1 d' \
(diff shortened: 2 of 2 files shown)


atodorov:
Also set the releasever to 24

tests/gui/anaconda-autogui-testing.ks
```diff
@@ -10,7 +10,7 @@ rootpw qweqwe
 bootloader --location=mbr
 zerombr
 clearpart --all
-part / --fstype="ext4" --size=6000
+part / --fstype="ext4" --size=10240
```
bcl: Have you been hitting problems? The tests are all minimal package sets so really 4G should be enough and 6G more than enough.

atodorov: The same ks file is used to create the LiveCD and the available space was not enough. 10G is probably too much though.

tests/gui/make_livecd.sh
```diff
@@ -77,7 +77,7 @@ livemedia-creator --make-iso \
                   --iso "${ISO}" \
                   --title Fedora \
                   --project Fedora \
-                  --releasever 21 \
+                  --releasever 24 \
```
atodorov: Btw this should probably be set to rawhide

clumens:
Do we just want to drop the tests/gui directory entirely?  I've not worked on it in forever, and it only ever kind of worked, and there's still no python3 dogtail to even use.  I'm a little disappointed in how automated testing worked out, but perhaps I should not be surprised.  It's hard.

bcl:
I agree. Leave the GUI testing for people and focus on what we can automate.

atodorov:
Hi guys,
let me play around a little bit with these tests. I'll try to execute them using a local disk image/directory instead of a virtual machine. It this is successful it could be easier to execute the tests and see how things are going. If not we can easily remove the entire directory.

atodorov:
Closing, see https://github.com/rhinstaller/anaconda/pull/457

PR #328, opened 2015-08-26
Defer to Fedora distro-wide settings for password strength (#1250746)

dashea:
Try this again with all the changes.

The alternative proposed was to carry patches in a contrib directory and that didn't really sound any better to me, and also sounded like more work.

### DIFF
--- a/data/interactive-defaults.ks
+++ b/data/interactive-defaults.ks
@@ -5,7 +5,7 @@ firstboot --enable
 
 %anaconda
 # Default password policies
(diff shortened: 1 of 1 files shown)


dashea:
Two things: this needs to pass values to override the defaults (--minlen=0, --notstrict), and if I do pass --minlen=-0 and --minquality=0 then weak passwords don't throw errors. Looking into the second part of that.

dashea:
I guess it disabling the check on minquality=0 does make sense. Set it to 1 intead.

vpodzime:
Looks good to me.

PR #329, opened 2015-08-26
unbuffered read in python3 only works for binary

bcl:
and for text files it uses blocks, not lines, so drop ,0 from the UEFI
firmware language read.


jkonecny12:
Seems good to me

vpodzime:
Looks good to me too.

bcl:
pushed

PR #215, opened 2015-07-16
Tests

dashea:
Maybe this will make jenkins work again

PR #331, opened 2015-08-28
LVM on RAID kickstart test

vpodzime:
This adds a single kickstart test for installation to LVM on top of MD
RAID. Being one of the most complex storage configurations we support the LVM on
RAID definitely deserves testing. However, since such installation is quite a
bit resource-hungry, it would be great if we could get along with this single
test as running multiple such tests could be a problem for our infrastructure
running kickstart tests. That's why this tests the harder case with one of the
LVs requested to be grown to a maximum possible size (which tests blivet's
calculations of free space in a VG using MD RAID PVs and calculations of MD RAID
metadata size).

Extra info not in the commit message:
My goal here is to make the test stricter and stricter as I go through blivet's sources and remove all weird and magic calculations we do there. Especially once the changes from [blivet's PR #215] (https://github.com/rhinstaller/blivet/pull/215) are effective, this test should also check the free space in the VG because we should do a good-enough job filling it up with the root LV that is requested to grow to a maximum size.

### DIFF
--- a/tests/kickstart_tests/lvm-raid-1.ks
+++ b/tests/kickstart_tests/lvm-raid-1.ks
@@ -1,0 +1,81 @@
+#version=DEVEL
+url --mirror=http://mirrors.fedoraproject.org/mirrorlist?repo=fedora-$releasever&arch=$basearch
+install
+network --bootproto=dhcp
+
+bootloader --timeout=1
+zerombr
+clearpart --all --initlabel
+part /boot --fstype=ext4 --size=500
+part raid.01 --size=8000 --ondisk=sda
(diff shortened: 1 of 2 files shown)


clumens:
I couldn't find any obvious way of telling parallel that we want an individual test run by itself.  There were a couple things that looked like I could make them work, but nothing really solid.  We'll just have to trust that running two at a time will be alright and if not, try something else.

vpodzime:
Pushed with the fixed author and with ``$release`` replaced with ``rawhide`` in mirror URL as per https://lists.fedorahosted.org/pipermail/anaconda-patches/2015-August/023261.html

PR #332, opened 2015-08-28
Small fixes for run_gui_tests.sh

atodorov:


vpodzime:
These both look good to me.

clumens:
Yep, looks good to me too.

atodorov:
Update on this one: 
* add if=virtio,cache=none for the VM drives, although this doesn't appear to give a noticable performance boost
* update the rootpassword.py test b/c anaconda doesn't allow weak passwords.

Note: due to the fact that these tests are incredibly slow to execute and often fail with no obvious reason I've not been able to verify the changes. Also I'm not seeing any significant performance boost which is frustrating

clumens:
Applied and pushed.  Thanks.

PR #333, opened 2015-08-28
Force an encoding of utf-8 on liveinst installs (#1257060)

bcl:
This adds a copy of sitecustomize.py to a directory that can be added
to PYTHONPATH, and adds that directory to PYTHONPATH when running
liveinst. This way anaconda do perform early sys changes while not
affecting python globally and not messing up regular imports too badly.

This does not fix unicode problems for all cases (image/dir install are
still affected), but it papers over them for live installs.

(cherry picked from commit 2615c6d8de2f981262cff4618be38a36284dc9d5)

Resolves: rhbz#1257060


vpodzime:
Big ACK to this neat solution.

bcl:
pushed

PR #334, opened 2015-08-28
Disallow /boot on lvm until grub2 fully supports it. (#1252466)

bcl:
Still not supported. Also see bug 967880

(cherry picked from commit 0b9fe390f07d889d3a01d9d7196eb1fbef462e17)
Resolves: rhbz#1252466


vpodzime:
Looks good to me.

bcl:
pushed

PR #335, opened 2015-08-31
network: assume --device=link as default also for ks on hd (#1085310)

rvykydal:
Ie when network is not activated in dracut.

Related: rhbz#1085310

If not overriden by ksdevice.


vpodzime:
Might as well remove the blank line right above. Looks good to me.

rvykydal:
Pushed.

PR #336, opened 2015-08-31
Master port gui spoke to libnm

rvykydal:
Port of network gui spoke to new libnm.

- Moves to using libnm objects (connection, device, ap) instead of uuids and names which allows for less amount of more readable and maintainable code.
- Removes python-dbus workarounds for secrets flags (could have been done using older libnm-glib as well, if there were not a bug, reported, and fixed now by NM in libnm-glib. Still NM recommends to use the new libnm
- Also allows to get rid of dependency of NetworkControlBox on nm.py (one instance of nm.py function call is still to be removed). As a follow-up NetworkControlBox could be quite easily split out to a module with a glade file independent of ancaonda code which makes its development a lot less pain.
- Using only single client now, which fixes problems with update of some properties and objects (client.get_connections()).
- The first two revert patches are replaced by two patches on top of the set as the port allowed for (and required) this - I think more robust - approach.

### DIFF
--- a/pyanaconda/nm.py
+++ b/pyanaconda/nm.py
@@ -891,21 +891,6 @@ def nm_add_connection(values):
         raise
     return connection
 
-def nm_delete_connection(uuid):
-    """Delete connection specified by uuid.
-
-       :param uuid: uuid of connection to be deleted
-       :type uuid: str
-       :return: True if connection was deleted, False if it was not found
-       :rtype: bool
-    """
-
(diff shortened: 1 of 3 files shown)


rvykydal:
Strange thing, the order of commits in this PR is different from the order on my branch:
https://github.com/rvykydal/anaconda/commits/master-port-gui-spoke-to-libnm

vpodzime:
Looks good to me other than the above questions/suggestions.

rvykydal:
Pushed to master

PR #337, opened 2015-08-31
Driver Update Disk fixes

bcl:
There are several issues here, this fixes 2 of them.

1. arch and DD-net are no longer used as part of the DD-x path for the extracted rpms.
2. addRepo's refresh lambda has been broken for a long time, it is only used by driver updates so wasn't noticed in Fedora.

A 3rd problem is that the drivers are extracted 3 times instead of once, resulting in /run/install/DD-1,DD-2,DD-3 still working on figuring that one out.


vpodzime:
Looks good to me.

bcl:
pushed

PR #338, opened 2015-08-31
Do not allow passphrases to be typed with non-ASCII keymaps (#681250)

dashea:
This will require a string freeze exception for F23, and I'll start on (look up how to) that process tomorrow. If anyone has suggestions for the wording of the message I'm glad to hear them, since I'm not sure how well the one I have explains the situation.

### DIFF
--- a/pyanaconda/ui/gui/spokes/lib/passphrase.py
+++ b/pyanaconda/ui/gui/spokes/lib/passphrase.py
@@ -25,13 +25,16 @@ gi.require_version("Gtk", "3.0")
 from gi.repository import Gtk
 
 import pwquality
+import langtable
 
 from pyanaconda.ui.helpers import InputCheck
 from pyanaconda.ui.gui import GUIObject
 from pyanaconda.ui.gui.helpers import GUIInputCheckHandler
+from pyanaconda.ui.helpers import InputCheckHandler
 from pyanaconda.constants import PW_ASCII_CHARS
 from pyanaconda.i18n import _, N_
 from pyanaconda.ui.gui.utils import really_hide, really_show
+from pyanaconda.ui.gui.xkl_wrapper import XklWrapper
 
 __all__ = ["PassphraseDialog"]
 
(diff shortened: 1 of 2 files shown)


pyanaconda/ui/gui/spokes/lib/passphrase.py
```diff
@@ -73,6 +76,18 @@ def __init__(self, data):
...
+        # from non-ASCII keymaps. This check depends on the current active keymap
+        # and not any particular input, but add it as an input check so that the error
+        # message is displayed through the same mechanism. Use add_check from InputCheckHandler
+        # so that no real input object is needed and so there is no delay wrapper around the check
+        self._keymap_check = InputCheckHandler.add_check(self, None, self._checkKeymap)
+
+        # Run the keymap check on keymap changes
+        self._xkl_instance = XklWrapper.get_instance()
+        self._xkl_instance.engine.connect("X-state-changed", self.on_keymap_change)
+        self._xkl_instance.engine.connect("X-config-changed", self.on_keymap_change)
```
vpodzime: The handler needs to be disconnected from both these signals somewhere because the XklEngine is a per-process instance that exists for the whole process' lifetime and thus it would run these handlers even if the dialog is destroyed etc.

vpodzime:
Generally I must say I don't like this change because it works around half of the issues without fixing the root cause (which is not in our code). People can still paste whatever they want from a clipboard (which works between the host and guest) and even with a layout supporting ASCII people may have troubles typing in their passphrase because the letters may have completely different positions. So unless we want to restrict installations including passphrase typing to the 'en' keyboard layout, there's no real solution for this issue on our side. The real fixes should happen in systemd (giving us the matching console keymap for X layout) and dracut (being able to load such keymaps). Or we could just wait for the thing replacing consoles with some graphical-based UI supporting X layouts.

dashea:
Yeah, after thinking about this some more, I'm not real happy with it either. Even with an anaconda-based solution, the better thing to check would be whether the current X keymap is equivalent to what will be used as the VC keymap, and systemd can't even tell us that.

PR #303, opened 2015-08-18
Enable and improve the check for swap LV size in LVM cache kickstart tests

vpodzime:
The swap LV should actually have the requested size if the [blivet PR #207] (https://github.com/rhinstaller/blivet/pull/207) gets merged.

PR #218, opened 2015-07-16
a couple more things related to the tests

dashea:

PR #339, opened 2015-09-01
Stricter LVM --grow kickstart tests

vpodzime:
The first patch is already part of the PR #303 which is waiting for a new rawhide build of blivet. The second commit makes our tests that try to grow LVs to a maximum size stricter in a way that it checks the resulting free space in the VG. The requirements are hard, but with [blivet's PR #218] (https://github.com/rhinstaller/blivet/pull/218) merged we should be able to meet them. And that's what tests are good for, right? In case we need to make things more benevolent due to unpredictability of low-level storage calculations we can always update the requirements in our tests. But let's start with the hardest check possible.


clumens:
The first patch was already applied some time ago.  I've applied the second and pushed.  Thanks.

PR #340, opened 2015-09-02
Fix hiding curl 404 errors outside debug mode

vojtechtrefny:
New version of https://github.com/vojtechtrefny/anaconda/commit/2b542027b1f4c01e77f2e9ea40e70519b0f10cd9
Previous version actually worked only because of bug in dracut and that bug is already fixed.

### DIFF
--- a/dracut/anaconda-lib.sh
+++ b/dracut/anaconda-lib.sh
@@ -1,6 +1,7 @@
 #!/bin/bash
 
 command -v unpack_img >/dev/null || . /lib/img-lib.sh
+command -v getarg >/dev/null || . /lib/dracut-lib.sh
 
 # config_get SECTION KEY < FILE
 # read an .ini-style config file, find the KEY in the given SECTION, and return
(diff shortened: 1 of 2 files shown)


dracut/anaconda-lib.sh
```diff
@@ -181,6 +182,13 @@ tell_user() {
     fi
 }
 
+# print something only in if debug/inst.debug/rd.debug
+debug_msg() {
+   if getargbool 0 rd.debug || getargbool 0 debug || getargbool 0 inst.debug; then
+     echo $* >&2
+  fi
```
vpodzime: bad indentation of ``fi``

vpodzime:

dracut/anaconda-lib.sh
vojtechtrefny: Thanks, fixed locally.

vojtechtrefny:
Pushed.

PR #342, opened 2015-09-02
Install driver-updates@.service as a data instead of a script

dashea:
This way it doesn't end up with the executable bit set and systemd
doesn't complain about it.


vpodzime:
Looks good to me too.

dashea:
Pushed

PR #343, opened 2015-09-03
Install ca-bundle certs in the initrd (#1259880)

bcl:
Without these curl cannot verify any CA's from loader.

Resolves: rhbz#1259880


vpodzime:
For some reason, I find it interesting to see a pull request on GitHub for a rhel6-branch. Looks good to me though.

bcl:
Pushed

PR #344, opened 2015-09-03
kickstart test local http

dashea:
To make this work in jenkins, the bridge interface for the "default" network (probably virbr0) will need to be added to firewalld's "trusted" zone, if it's not there already: `firewall-cmd --permanent --zone=trusted --add-interface=virbr0`

Also the kickstart tests job would need to be updated for the new script locations.

This does make the test results noisier because I left the rpmbuild junk on stderr, so if that becomes a problem that can be redirected away, too.


vpodzime:
Other than the nitpicks above these all look good to me.

dashea:
Moved the shell portion of starting httpd to a function in functions.sh, and fixed the part where I misspelled "serve" after typing "server" like 15 times.

dashea:
Oops, forgot to switch all the $(pwd)'s.

vpodzime:
Looks good to me. Feel free to switch the ``$(pwd)``'s locally when pushing this.

dashea:
Pushed.

PR #347, opened 2015-09-04
Kickstart URL include

dashea:
The kickstart test depends on https://github.com/rhinstaller/anaconda/pull/344, but I rebased this branch to master so you don't have to look those commits twice.

As far as actually making things work, there's also a pykickstart change that's needed (return .text instead of .content in _load_url), but I haven't created a PR for that yet since I'm still working on adding a test case to pykickstart for it.


vpodzime:
Looks good to me.

dashea:
Pushed.

PR #348, opened 2015-09-07
Activate coverage for tests executed with sudo

atodorov:
Straight forward, just repeat what we already do but for root. Also remove the helper script and use Python directly.


vpodzime:
Looks good to me.

clumens:
Looks fine to me too.

vpodzime:
Pushed.

PR #349, opened 2015-09-07
Fix 'cat: /tmp/dd_disk: No such file or directory' (#1251394)

jkonecny12:
Missing /tmp/dd_disk file is not an error, changing to debug print instead.

*Resolves: rhbz#1251394*

### DIFF
--- a/dracut/driver-updates-genrules.sh
+++ b/dracut/driver-updates-genrules.sh
@@ -14,8 +14,15 @@ else
     DD_OEMDRV="LABEL=OEMDRV"
 fi
 
+DD_DISK=""
+if [ -f /tmp/dd_disk ]; then
+    DD_DISK=$(cat /tmp/dd_disk)
+else
+    debug_msg "/tmp/dd_disk file was not created"
+fi
+
 # Run driver-updates for LABEL=OEMDRV and any other requested disk
-for dd in $DD_OEMDRV $(cat /tmp/dd_disk); do
+for dd in $DD_OEMDRV $DD_DISK; do
     when_diskdev_appears "$(disk_to_dev_path $dd)" \
         driver-updates --disk $dd \$devnode
 done


vpodzime:
Looks good to me.

jkonecny12:
Pushed to rhel7-branch. Will go to master after dependency PR will be there too.

jkonecny12:
Pushed to master too. Closing this PR

dpward:
@jkonecny12 Please push to f23-branch as well. It is also affected and is breaking the ability to compose images using pungi.

jkonecny12:
We didn't push fixes to old branches because when the Fedora is already released the anaconda won't be updated anymore for the given release. There is no point in it because the GA compose won't be updated. After this deadline all patches goes only to master branch.

dpward:
@jkonecny12 Anaconda is not just for GA composes; it's also needed in order to create a [Fedora Remix](https://fedoraproject.org/wiki/Remix) using pungi, which currently does not work for Fedora 23 because of this issue. Pushing a fixed anaconda package to the updates repository would resolve this so it's not pointless.

What I am not entirely clear on is if the patches should only go directly into the anaconda source RPM for Fedora 23, rather than applying them to the f23-branch here first and bumping the version number.

dashea:
Could you explain how this is breaking pungi? The error message is related to driver update disks and is not fatal. It should have no effect on pungi.

mlopezcoria:
I'm trying to install fedora 23 from a net install iso and I can't because of this error. How do I fix it?

jkonecny12:
Hi @mlopezcoria this is change in the initial ramdisk which is harder to apply. You can use this overlay image https://jkonecny.fedorapeople.org/redhat/f23-dd-fix-overlay.img but for the application you need to use some PXE but unfortunately I can't tell you which one is suitable for your.

dashea:
@mlopezcoria the error message fixed by this PR is not fatal. The issue you are having is probably caused by something else. If you are able to gather more information, such as the log files from /tmp, or /run/initramfs/rdsosreport.txt if the error is occurring in the initramfs, please file a bug at http://bugzilla.redhat.com/

PR #350, opened 2015-09-08
Fix a DBus InvalidProperty handling

jkonecny12:
Rawhide anaconda crashing because the new InvalidProperty exception is
now raised.

DBus API is returing exception now so we have to catch the exception
instead of only testing for empty returned variable.

### DIFF
--- a/pyanaconda/nm.py
+++ b/pyanaconda/nm.py
@@ -263,7 +263,12 @@ def nm_device_property(name, prop):
             raise UnknownDeviceError(name, e)
         raise
 
-    retval = _get_property(device, prop, ".Device")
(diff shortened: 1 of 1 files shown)


jkonecny12:
@clumens I wanted to log this because I don't know why this exception is raising now but you are right that we don't need it so I can remove it.
Thank you for the misspelling catch.

@vpodzime I tried it and no, log.debug is logged always but with DEBUG label.

jkonecny12:
I have updated this PR.

jkonecny12:
I'm closing this one because I have found that this doesn't handle all problems and I have found better solution.

PR #351, opened 2015-09-08
set sysroot correctly when setting root password (#1260875)

AdamWill:
as @sgallagher noted, #1260875 happens because the `setRootPassword` call doesn't specify the system root (so we wind up changing the anaconda environment's root password, not the installed system's root password).

The default for `users.setRootPassword` itself could be changed instead of / as well as this, we weren't sure which would be the best place to do it. But this does work, I tested it.


dashea:
Looks fine, and I can add some test cases for root password.

vpodzime:
Pushed.

PR #353, opened 2015-09-08
Install all required packages.

dashea:
Append the packages from doInstall to requiredPackages instead of
replacing requiredPackages.

Resolves: rhbz#1260624


vpodzime:
Looks good to me.

dashea:
Pushed.

PR #354, opened 2015-09-09
Fix a DBus InvalidProperty handling

jkonecny12:
Rawhide anaconda crashing because the new InvalidProperty exception is
now raised.

DBus API is returing exception now so we have to catch the exception
instead of only testing for empty returned variable.

*Resolves: rhbz#1260239*

This is updated PR of this one https://github.com/rhinstaller/anaconda/pull/350. I have found out it will crash later in the anaconda with the old fix so I'm moving the solution closer to the source to fix both problems. Also it looks nicer here.

### DIFF
--- a/pyanaconda/nm.py
+++ b/pyanaconda/nm.py
@@ -122,7 +122,8 @@ def _get_property(object_path, prop, interface_name_suffix=""):
     try:
         prop = proxy.Get('(ss)', interface_name, prop)
     except GLib.GError as e:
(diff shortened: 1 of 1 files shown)


vpodzime:
*"Also it looks nicer here."* (not *it's*). Looks good to me otherwise.

jkonecny12:
Fixed this was only PR comment :D

jkonecny12:
Pushed.

PR #357, opened 2015-09-09
Use MDRaidArrayDevice.members instead of .devices

dashea:
.devices is deprecated in blivet.


vpodzime:
Looks good to me too.

dashea:
Pushed.

PR #360, opened 2015-09-10
Handle driver rpms retrieved via network (#1257916)

bcl:
Resolves: rhbz#1257916


vpodzime:
Looks good to me.

bcl:
pushed

PR #361, opened 2015-09-15
Tell gettext that anaconda is not a GNU package.

dashea:
This prevents gettext's auto-GNU-detection from picking up, for example,
the "grep 'GNU anaconda'" command that got written to a log file and
then failing in a confusing way.

The PACKAGE_GNU variable was added in gettext 0.19.1, so update the
requirements for autoconf and rpm accordingly.


vpodzime:
Looks good to me.

PR #362, opened 2015-09-16
Rename the gettext tests

dashea:
The tests/gettext directory looks too much like a python module, so
rename it to something else to avoid conflicts with python's gettext.


vpodzime:
Looks good to me too.

dashea:
Pushed

PR #364, opened 2015-09-17
Use yum to install the mock buildroot for now.

dashea:
Something is busted with dnf and mock, so switch to yum as a workaround
so we can get packages building again.


dashea:
https://bugzilla.redhat.com/show_bug.cgi?id=1264198 is the dnf bug and also where the --yum idea came from. I guess I should amend that it into the commit message.

This passed locally using the VM I had that was failing on the dnf stuff. Will it work in jenkins? Only time will tell!

dashea:
Pushed.

PR #365, opened 2015-09-18
Fix branding when iso is downloaded from nfs or hd (#1252756)

jkonecny12:
We should load `product.img` and `updates.img` from the same place as `install.img` is loaded which is `/run/install/repo`.

*Resolves: rhbz#1252756*

----------------
I'm still wondering why this was there if it had some special reason. If someone know that please tell me if this patch is alright.


bcl:
Looks good to me, it's likely that it wasn't caught because hardly anyone uses nfsiso with product.img until recently.

jkonecny12:
Thank you @bcl for confirmation.
Pushed

PR #368, opened 2015-09-21
Catch blkid failure in driver-updates (#1262963)

bcl:
Some systems are slow to bring up their devices, it is possible to enter
the DUD UI before they are available. blkid returns a 2 when there is
nothing to output so we need to catch CalledProcessError to give them
the chance to refresh a few times until they show up.

Resolves: rhbz#1262963


jkonecny12:
Seems good to me.

bcl:
Pushed.

PR #369, opened 2015-09-22
Use correct installclass for RHEL Atomic Host (#1265213)

rvykydal:
RHELAtomicInstallClass inherits from RHELBaseInstallClass so we can't
rely on BaseInstallClass default hidden=False

Resolves: rhbz#1265213


bcl:
cherry picked for tonight's build.

PR #370, opened 2015-09-22
rpmostreepayload: Forcibly unmount everything in sysroot

cgwalters:
There is an issue with 7.2 where something in the interaction
between blivet/systemd/rpmostreepayload is mounting filesystems
"deep" in `/mnt/sysimage/ostree/deploy/$x/sysroot/deploy/...`.

This then triggers another bug where the initramfs tmpfs gets remounted
read-only, and that in turns breaks dracut's ability to shutdown.

While these mounts do appear even in Fedora now, they're not actively
breaking things (for some reason), so let's just go with this crude
but functional hack for now, then re-investigate upstream and get the
mounts looking saner there.

### DIFF
--- a/anaconda
+++ b/anaconda
@@ -46,7 +46,7 @@ if ("debug=1" in proc_cmdline) or ("debug" in proc_cmdline):
 
 import atexit, sys, os, time, signal
 
-def exitHandler(rebootData, storage, exitCode=None):
+def exitHandler(rebootData, storage, payload, exitCode=None):
     # Clear the list of watched PIDs.
     iutil.unwatchAllProcesses()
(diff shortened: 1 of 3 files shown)


bcl:
In the commit message please refer to the bug in the title. eg. (#1234567) and in the body:

Resolves: rhbz#1234567

I like this version better than the previous pass, the behavior change is limited to just ostree for now and we can look at unmounting for other payloads later.

cgwalters:
Done.

vpodzime:
Looks good to me.

clumens:
Merged, thanks.

PR #372, opened 2015-09-25
Mount stage2 cdrom after running driver-updates (#1266478)

bcl:
driver-updates is now running from the initqueue. When booting from an
iso the stage2 would already be mounted, meaning it could not be ejected
to swap in a driver update cdrom.

This patch delays mounting the cdrom until after driver-updates has
been run, allowing it to be swapped for a DUD and swapped back in.

Resolves: rhbz#1266478


bcl:
pushed

PR #374, opened 2015-09-29
Raise an error if osimg cannot be found (#1248673)

bcl:
The os.stat(osimg.path) has become a false positive catch point for
other failures. Raise an error on purpose if it cannot be set. The real
error should be in the logs someplace.


jkonecny12:
Looks good to me

vpodzime:
Looks good to me too.

bcl:
pushed.

PR #375, opened 2015-09-29
Verify downloaded packages using key provided with "repo --gpgkey" option

marmarek:
This depends on pykickstart support
https://github.com/rhinstaller/pykickstart/pull/32

The `gpgsigcheck` function is almost 1-1 copy from dnf/cli/cli.py.

Warning: I haven't tested this in this particular case, but very similar code in [pungi](https://pagure.io/pungi/pull-request/63) and [livecd-tools](https://github.com/rhinstaller/livecd-tools/pull/14) works just fine.

One thing I'm not sure about is error reporting: currently signature verification failure causes `dnf.exceptions.Error`, which isn't handled by this code. Should some other exception be used? Or something else (`errors.errorHandler`?)?

### DIFF
--- a/pyanaconda/kickstart.py
+++ b/pyanaconda/kickstart.py
@@ -1575,7 +1575,7 @@ class RaidData(commands.raid.F23_RaidData):
                                      parents=request)
             storage.createDevice(luksdev)
 
-class RepoData(commands.repo.F21_RepoData):
+class RepoData(commands.repo.F24_RepoData):
     def __init__(self, *args, **kwargs):
         """ Add enabled kwarg
 
@@ -1585,7 +1585,7 @@ class RepoData(commands.repo.F21_RepoData):
         self.enabled = kwargs.pop("enabled", True)
         self.repo_id = kwargs.pop("repo_id", None)
 
-        commands.repo.F21_RepoData.__init__(self, *args, **kwargs)
+        commands.repo.F24_RepoData.__init__(self, *args, **kwargs)
 
 class ReqPart(commands.reqpart.F23_ReqPart):
     def execute(self, storage, ksdata, instClass):
--- a/pyanaconda/packaging/dnfpayload.py
+++ b/pyanaconda/packaging/dnfpayload.py
@@ -675,6 +675,30 @@ class DNFPayload(packaging.PackagePayload):
         self._base.read_comps()
         self._refreshEnvironmentAddons()
 
+    def gpgsigcheck(self, pkgs):
+        """Perform GPG signature verification on the given packages,
+        installing keys if possible.
+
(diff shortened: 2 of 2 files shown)


vpodzime:
> One thing I'm not sure about is error reporting: currently signature verification failure causes dnf.exceptions.Error, which isn't handled by this code. Should some other exception be used? Or something else (errors.errorHandler?)?

I think we need a better granularity here and adding a handler that would tell users what went wrong.

marmarek:
`packaging.PayloadInstallError` with appropriate message?

vpodzime:
> packaging.PayloadInstallError with appropriate message?

Yes, that would make sense to me.

marmarek:
Ok, done.

marmarek:
Updated this for F24 pykickstart. The branch name is somehow outdated now...

Frodox:
+1

jcpunk:
Would it make more sense to hand that off to DNF via the repo definition, catch the error in anaconda, and push it to the user?  That would prevent the duplication of the DNF code.

marmarek:
This will also require a change in DNF code structure. Currently DNF (as a library) do not check signatures, even if configured to do so in repo definition. It is implemented only in dnf _cli utility_:
https://github.com/rpm-software-management/dnf/blob/master/dnf/cli/cli.py#L231-L262

But this approach (including appropriate change in DNF) indeed may make more sense.

jcpunk:
I filed https://bugzilla.redhat.com/show_bug.cgi?id=1448164 to see about possibly coordinating the effort.

InStallker:

InStallker:

jkonecny12:
Hello everyone, we think this issue could be solved by this proposal: https://hackmd.io/@prJIqOTeSxC-W0RNGKIGbQ/ByzIz3kIH/edit

Please look and comment there if you are interested.

jkonecny12:
Not sure if everyone agrees on the proposed solution or it's not an issue anymore but I'm closing this PR now because it got stale. If you want to start this discussion again feel free to write here so I'll re-open it again.

j-mracek:
@Conan-Kudo I have a good news, DNF has brand new API to support GPG verification and import of GPG keys. Therefore I suggest that the code could be replaced by dnf.Base().package_import_key() (for impotr) or dnf.Base().package_signature_check() for detection if import is needed.

Conan-Kudo:
@j-mracek Sweet, when did this API get introduced?

j-mracek:
@Conan-Kudo dnf-4.2.23. See: https://github.com/rpm-software-management/dnf/commit/32ea7b347f859ed5e2d0d9ecb19da7a515192d4b

PR #376, opened 2015-09-29
Do not override StorageChecker.errors in StorageSpoke (#1252596)

vojtechtrefny:
When assigning checkDiskSelection result to self.errors, we can
lose errors from CustomPartitioningSpoke. Added separate "category"
for disk selection errors to fix this.

### DIFF
--- a/pyanaconda/ui/gui/spokes/storage.py
+++ b/pyanaconda/ui/gui/spokes/storage.py
@@ -268,6 +268,7 @@ class StorageSpoke(NormalSpoke, StorageChecker):
         self._last_selected_disks = None
         self._back_clicked = False
         self.autopart_missing_passphrase = False
+        self.disks_errors = []
 
         # This list contains all possible disks that can be included in the install.
         # All types of advanced disks should be set up for us ahead of time, so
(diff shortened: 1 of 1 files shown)


vpodzime:
Looks good to me.

AdamWill:
Uh. What about all the other places where the file checks for `self.errors` - i.e. `completed()`, `status()`, `refresh()`? This isn't going to do the right thing there, is it? The `disk_errors` will be disregarded.

vojtechtrefny:
> Uh. What about all the other places where the file checks for self.errors - i.e. completed(), status(), refresh()? This isn't going to do the right thing there, is it? The disk_errors will be disregarded.

These errors are specific just for the case when your disk selection is "wrong". The `checkDiskSelection` method returns errors only if you select just subset of disks with raid/lvm/btrfs. In this case you need either to select them all or quit the installation. It is not possible to leave the `StorageSpoke` with this selection and the check is performed every time you try to leave it (using the "Done" button).

pyanaconda/ui/gui/spokes/storage.py
```diff
@@ -1014,7 +1015,28 @@ def on_specialized_clicked(self, button):
...
+                errors = "\n".join(self.disks_errors)
+                dialog.refresh(errors)
+                rc = dialog.run()
+
+            dialog.window.destroy()
+
+            if rc == 0:
+                # Quit.
+                iutil.ipmi_report(constants.IPMI_ABORTED)
+                sys.exit(0)
```
jkonecny12: Could you please move the same code (below) to separate method and call it instead.

vojtechtrefny: Actually the same code is not as long -- the `dialog` is different (different buttons), `label` is different and `errors` are different. I'm not sure that a new method just for this case with five parameters would be better/more clear.

jkonecny12: Yeah right, sorry I missed this differences before. 
Looks good to me (now when I'm not so blind :) ), I'm returning the Ack now.

vojtechtrefny:
Comment added to explain why errors from `checkDiskSelection` can be in separate list and not in `self.errors`.

vpodzime:
Looks even better to me now, thanks!

vojtechtrefny:
Pushed.

Still not convinced if a special method is needed here, but I can always do a follow-up fixing this.

PR #377, opened 2015-09-29
Move UEFI to special category in Custom spoke (#1183880)

jkonecny12:
The UEFI partitions are moved to special category to prevent user from
unintentional deletion which could cause existing systems to be
unbootable. This category is used only if the UEFI is shared between
installed systems or if there are any unknown filesystems (could be
Windows UEFI).

*Resolves: rhbz#1183880*

--------------------
Screenshots:
### UEFI shared
UEFI shared

### UEFI doesn't show up in the existing system
UEFI doesn't show up in the existing system

### Removed RHEL system by "Delete all" button
UEFI removed system by "Delete all" button

### Move all UEFIs when any unknown partition present
Move all UEFIs when any unknown partition present

### UEFI without sharing between systems and without unknown partition
UEFI without sharing

### DIFF
--- a/pyanaconda/ui/gui/spokes/custom.py
+++ b/pyanaconda/ui/gui/spokes/custom.py
@@ -500,6 +500,7 @@ class CustomPartitioningSpoke(NormalSpoke, StorageChecker):
         log.debug("ui: new_devices=%s", [d.name for d in new_devices])
 
         ui_roots = self._storage_playground.roots[:]
+        page_efi = Page(_("UEFI partitions"))
 
         # If we've not yet run autopart, add an instance of CreateNewPage.  This
         # ensures it's only added once.
@@ -527,6 +528,16 @@ class CustomPartitioningSpoke(NormalSpoke, StorageChecker):
             new_root = Root(mounts=mounts, swaps=swaps, name=translated_new_install_name())
             ui_roots.insert(0, new_root)
 
+        # Find out if all efi partitions are shared or not
+        efi_shared = {}
+        for root in ui_roots:
+            for (mountpoint, device) in root.mounts.items():
+                if device.format.type == "efi":
+                    if device.name not in efi_shared:
+                        efi_shared[device.name] = False
+                    else:
+                        efi_shared[device.name] = True
+
         # Add in all the existing (or autopart-created) operating systems.
         for root in ui_roots:
             # Don't make a page if none of the root's devices are left.
@@ -543,6 +554,12 @@ class CustomPartitioningSpoke(NormalSpoke, StorageChecker):
                    not device.disks or \
                    (root.name != translated_new_install_name() and not device.format.exists):
                     continue
+                # Skip efi devices shared between existing systems or skip all efi devices if any
+                # unknown device is presented. Move them to the separate section.
+                if device.format.type == "efi":
+                    if efi_shared.get(device.name, False) or self.unusedDevices:
+                        page_efi.addSelector(device, self.on_selector_clicked, mountpoint=root.name)
+                        continue
 
                 selector = page.addSelector(device, self.on_selector_clicked,
                                             mountpoint=mountpoint)
@@ -556,18 +573,28 @@ class CustomPartitioningSpoke(NormalSpoke, StorageChecker):
                 selector = page.addSelector(device, self.on_selector_clicked)
                 selector.root = root
 
-            page.show_all()
-            self._accordion.addPage(page, cb=self.on_page_clicked)
+            # Do not populate empty page
+            if page.members:
+                page.show_all()
+                self._accordion.addPage(page, cb=self.on_page_clicked)
 
         # Anything that doesn't go with an OS we understand?  Put it in the Other box.
         if self.unusedDevices:
             page = UnknownPage(_("Unknown"))
 
             for u in sorted(self.unusedDevices, key=lambda d: d.name):
-                page.addSelector(u, self.on_selector_clicked)
+                if u.format.type == "efi": # move efi to "UEFI category"
+                    page_efi.addSelector(u, self.on_selector_clicked)
+                else:
+                    page.addSelector(u, self.on_selector_clicked)
 
             page.show_all()
             self._accordion.addPage(page, cb=self.on_page_clicked)
+
+        # Show efi partitions page if not empty
+        if page_efi.members:
+            page_efi.show_all()
+            self._accordion.addPage(page_efi, cb=self.on_page_clicked)
 
     def _do_refresh(self, mountpointToShow=None):
         # block mountpoint selector signal handler for now


vpodzime:
Looks good to me and I like the approach to the (real) issue.

clumens:
I have several general, nitpicky comments:

* I don't like the name "UEFI partitions".  We've avoided the word "partition" in general, and as adamw pointed out in IRC is seems awfully specific.  We'll likely have this same situation on other platforms.  Perhaps the name should be more generic.

* I think it's accidental that these are showing up under a DATA heading.  Regardless, I don't think it belongs under any heading.  I think there should probably be a different Page subclass that these guys live in, similar to how there's an Unknown page.  In fact you might just be able to borrow from that.

* I find it a little weird that there can be multiple selectors with different names but the same device.  I assume that's because it's part of multiple OSes?  I feel like that's hard to discover here.  Could they perhaps be grouped together but separated from other possibly shared UEFI partitions?  That is, in the next-to-last screenshot, could the two sda1s be their own page, vda1s be their own, etc.?  I don't know if that would look good though.  It might be more confusing.

* What happens when you delete just the first sda1 selector in the next-to-last screenshot?  Does the other one disappear as well, or do you have to click delete twice to actually delete one thing?

Thanks for taking a look at this, though.  UI stuff is just always tons of nitpicky details.

AdamWill:
@clumens note I wasn't just talking about the UI - I was thinking about the way the actual code is written. This code is very specifically tailored to UEFI, I was thinking it might make sense for it to be a bit more generic and use the data that's in the platform stuff about shared boot partitions and stuff (as a rough thought, I haven't looked into it in any detail).

pyanaconda/ui/gui/spokes/custom.py
```diff
@@ -500,6 +500,7 @@ def _populate_accordion(self):
         log.debug("ui: new_devices=%s", [d.name for d in new_devices])
 
         ui_roots = self._storage_playground.roots[:]
+        page_efi = Page(_("UEFI partitions"))
```
AdamWill: Perhaps this could be `UnknownPage`, as we don't need the SYSTEM/DATA sorting stuff?

AdamWill:
How do folks feel about this possible alternative?

    diff --git a/pyanaconda/ui/gui/spokes/custom.py b/pyanaconda/ui/gui/spokes/custom.py
    index 24c8474..4405097 100644
    --- a/pyanaconda/ui/gui/spokes/custom.py
    +++ b/pyanaconda/ui/gui/spokes/custom.py
    @@ -69,6 +69,7 @@ from blivet.errors import NoDisksError
     from blivet.errors import NotEnoughFreeSpaceError
     from blivet.devicelibs import raid, crypto
     from blivet.devices import LUKSDevice
    +from blivet.platform import platform
     
     from pyanaconda.storage_utils import ui_storage_logger, device_type_from_autopart
     from pyanaconda.storage_utils import DEVICE_TEXT_PARTITION, DEVICE_TEXT_MAP, DEVICE_TEXT_MD
    @@ -1962,10 +1963,28 @@ class CustomPartitioningSpoke(NormalSpoke, StorageChecker):
                         return
     
                 if dialog.deleteAll:
    -                for dev in (s._device for s in page.members):
    -                    self._destroy_device(dev)
    -            else:
    -                self._destroy_device(device)
    +                protected = platform.bootStage1ConstraintDict["format_types"]
    +                otherpgs = (pg for pg in self._accordion.allPages if
    +                            pg.pageTitle != page.pageTitle)
    +                otherdevs = []
    +                log.info("deleteall: otherdevs: %s", otherdevs)
    +                for otherpg in otherpgs:
    +                    otherdevs.append(mem._device.id for mem in otherpg.members)
    +                # we never want to delete known-shared devs here. we also
    +                # exclude 'device' as we'll unconditionally delete it later
    +                for dev in (s._device for s in page.members if
    +                            s._device.id not in otherdevs and
    +                            s._device.id != device.id):
    +                    # we only want to delete boot partitions if they're not
    +                    # shared *and* we have no unknown partitions
    +                    if (not self.unusedDevices or dev.format.type not in protected):
    +                        log.info("deleteall: removed %s", dev.name)
    +                        self._destroy_device(dev)
    +                    else:
    +                        log.info("deleteall: didn't remove %s", dev.name)
    +
    +            # We do this even in deleteAll case, see above comment
    +            self._destroy_device(device)
     
             log.info("ui: removed device %s", device.name)
     
Basically, when 'delete all' is clicked, don't delete any partitions that are 'shared' (appear in other roots), or bootloader partitions if there are any unknown partitions (stole that idea from @jkonecny12 for the same reason - we don't detect all OSes, so we use this as a proxy). I've tested it works in a simple UEFI case, haven't tested the `otherdevs` (shared partitions) stuff yet. I can test this and work on it some more if people are interested in the approach. If we do this we *could* adjust the 'delete all' text to explain more clearly what will happen, and maybe the delete confirm dialog text for bootloader partitions, but of course those are string freeze breaks.

jkonecny12:
> I don't like the name "UEFI partitions". We've avoided the word "partition" in general, and as adamw pointed out in IRC is seems awfully specific. We'll likely have this same situation on other platforms. Perhaps the name should be more generic.

Are you have some idea about how it should be named? Only "UEFI" seems little odd.

> I think it's accidental that these are showing up under a DATA heading. Regardless, I don't think it belongs under any heading. I think there should probably be a different Page subclass that these guys live in, similar to how there's an Unknown page. In fact you might just be able to borrow from that.
 
I missed this solution, good idea. I tested it and it works nice. I'll change it to `UnknownPage`.

> I find it a little weird that there can be multiple selectors with different names but the same device. I assume that's because it's part of multiple OSes? I feel like that's hard to discover here. Could they perhaps be grouped together but separated from other possibly shared UEFI partitions? That is, in the next-to-last screenshot, could the two sda1s be their own page, vda1s be their own, etc.? I don't know if that would look good though. It might be more confusing.

Yeah that could be but how you want to name the category `UEFI Shared` any idea? Also this will make the patch bigger. I think this part should go only to master. Question is, if only this part or whole patch and use the @AdamWill version instead.

> What happens when you delete just the first sda1 selector in the next-to-last screenshot? Does the other one disappear as well, or do you have to click delete twice to actually delete one thing?

Now it remove all sda1 items. I personally don't know which method is better if remove all or only one? From implementation point of view there could be some counter but as before it makes the patch bigger and it should go only to master though.

pyanaconda/ui/gui/spokes/custom.py
jkonecny12: I already tested it and it works nice. I'm going to change it to `UnknownPage`. Thank you for advice.

dwlehman:
The checkbox says "Delete all other file systems in the ___ root as well". We should not start doing something different unless we're also going to change that string to describe the silliness we're trying to accomplish here.

jkonecny12:
@AdamWill are you think that my patch is so invasive for F23? It's only one method change.

AdamWill:
@dwlehman right, as I said, a full version of my alternative would go along with a change to that string and also possibly a change to the delete confirmation dialog text for bootloader partitions. I just wanted to float the *idea* first, though, so I didn't waste time working on it further if no-one was interested.

@jkonecny12 Well, I'm not sure, I just wanted to provide a possible alternative. I'm a *little* worried about yours because it's a fairly significant UI change. For instance, the 'roots' are really supposed to be operating systems; now you're inventing one that isn't. We probably have docs that talk about these groups (has anyone looked at the 'Help' text?) - they'd have to change. The issues @clumens noted with the same ESP appearing multiple times in the 'UEFI partitions' root seem a bit difficult to deal with cleanly. And another consequence I thought of: when you delete a partition in the 'UEFI partitions' root you get that same 'do you want to delete all the other partitions in this root?' question, which really doesn't make a lot of sense.

We should probably ask mizmo about this?

AdamWill:
As for the naming question, I would call it something like 'System boot partitions'.

dwlehman:
I don't really think there is a legitimate problem to fix here, but if we are to address it, I would prefer we do it quietly, like @AdamWill is proposing. There's no need for more UI here aside from updating strings in existing dialog(s).

dwlehman:
```patch
diff --git a/pyanaconda/ui/gui/spokes/lib/custom_storage_helpers.py b/pyanaconda/ui/gui/spokes/lib/custom_storage_helpers.py
index e095fee..a3d7aa9 100644
--- a/pyanaconda/ui/gui/spokes/lib/custom_storage_helpers.py
+++ b/pyanaconda/ui/gui/spokes/lib/custom_storage_helpers.py
@@ -355,7 +355,9 @@ class ConfirmDeleteDialog(GUIObject):
             rootName = rootName.replace("_", "__")
         self._removeAll.set_label(
                 C_("GUI|Custom Partitioning|Confirm Delete Dialog",
-                    "Delete _all other file systems in the %s root as well.")
+                    "Delete _all other file systems in the %s root as well, "
+                    "including those used by other installed operating "
+                    "systems.")
                 % rootName)
         self._removeAll.set_sensitive(rootName is not None)
 ```

This is about as much as is warranted in my opinion.

AdamWill:
@dwlehman so current status is the bug was rejected as a blocker, so you can do whatever you like here, but it'd be nice if we could at least pick one thing (this, my #383 , or your text change) and do it? if that's just your text change, so be it, but even that is a definite improvement on current behaviour - at least it cues the user explicitly that the button might be dangerous and they should stop and think.

jkonecny12:
There were better solutions so I'm closing this PR.

PR #378, opened 2015-09-29
Use the bootloader raid levels for bootloader installation (#1266898)

bcl:
We already use bootloader.stage2_raid_levels to validate the RAID
selection in the UI. It should also be used when installing the
bootloader. This should allow grub2 on BIOS to work with RAID levels 0,
1, 4, 5, 6 and 10.

Note that UEFI only supports RAID1 due to needing a ESP partition on each
devices.


vpodzime:
Good point. :) Looks good to me.

bcl:
pushed

PR #379, opened 2015-09-30
Improve the creation of a user with a list of groups

dashea:


vpodzime:
Looks good to me.

clumens:
Yeah, I like it even if jenkins doesn't.

dashea:
Pushed.

PR #380, opened 2015-10-01
Check minimal memory requirements properly (#1267673)

jstodola:
isys.total_memory() returns memory size in kB, while isys.MIN_RAM and isys.MIN_GUI_RAM are in MB.

Resolves: rhbz#1267673

### DIFF
--- a/anaconda
+++ b/anaconda
@@ -460,7 +460,7 @@ def check_memory(anaconda, options, display_mode=None):
         display_mode = anaconda.displayMode
 
     reason = reason_strict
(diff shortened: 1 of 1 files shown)


vpodzime:
Ooops, looks good to me.

clumens:
This would apply to master too.  Of course, we can't push to rhel7-branch right now because it's not a blocker.

vpodzime:
Pushed to master. I'll make sure it gets to the rhel7-branch at the right time.

PR #381, opened 2015-10-01
Limit DNF's parallel downloads to 1 connection when using a proxy

bcl:
When using a proxy all connections are funneled through a single system,
parallel downloads can overwhelm the proxy and don't add any
benefit so limit it to 1.

### DIFF
--- a/pyanaconda/packaging/dnfpayload.py
+++ b/pyanaconda/packaging/dnfpayload.py
@@ -409,6 +409,9 @@ class DNFPayload(packaging.PackagePayload):
                 if proxy.password:
                     conf.proxy_password = proxy.password
                 log.info("Using %s as proxy", self.data.method.proxy)
+
+                # Prevent parallel downloads from swamping the proxy
+                conf.max_parallel_downloads = 1
             except ProxyStringError as e:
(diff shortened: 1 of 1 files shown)


vpodzime:
Looks good to me too.

kellinm:
The only time this would be an issue would be a manually configured non-transparent proxy.

For those cases, it is most likely not a home user, and would be up to the network administrator to deal with traffic and congestion. I would high recommend against this being the default behavior.

dashea:
> The only time this would be an issue would be a manually configured non-transparent proxy.

Which is why it's set when the user manually configures a non-transparent proxy?

kellinm:
Correct - but we are making the assumption that the proxy has less upload than the client can handle - having a non-transparent proxy is not the standard home-user scenario.

Given a non-transparent proxy already requires some level of advanced knowledge both in proxy and client configuration - an option like this might be better as an additional option in a kickstart or at the CLI placing the power into the network administrator's hands as to whether to throttle the number of client connections or increase the proxy's bandwidth.

bcl:
Yeah, making this non-configurable would end up causing problems. I'll ponder whether or not it should be yet another switch.

PR #382, opened 2015-10-01
Fix the locale pattern packages-instlangs-3 looks for.

dashea:
The names of the languages as actual English words ended up in the
localedef output, so filter those out, too.


dashea:
Pushed.

PR #383, opened 2015-10-01
don't delete shared/boot parts in deleteAll (#1183880)

AdamWill:
This is one potential approach to help with #1183880 . It
changes the deleteAll behaviour as follows:

* Partitions known to be shared with other OSes won't be deleted
* Boot partitions won't be deleted if there are unknown parts

You can delete these partitions only by specifically targeting
them. It also adds a variant of the ConfirmDeleteDialog for
boot partitions, with text specifically explaining that they
might be needed for other OSes to boot.

This is my alternative to #377 - I'm just providing it as an option. We could pull the ConfirmDeleteDialog change out of this as a standalone to go along with dlehman's minimal change to the 'delete other partitions' text, also?

### DIFF
--- a/pyanaconda/ui/gui/spokes/custom.py
+++ b/pyanaconda/ui/gui/spokes/custom.py
@@ -69,6 +69,7 @@ from blivet.errors import NoDisksError
 from blivet.errors import NotEnoughFreeSpaceError
 from blivet.devicelibs import raid, crypto
 from blivet.devices import LUKSDevice
+from blivet.platform import platform
 
 from pyanaconda.storage_utils import ui_storage_logger, device_type_from_autopart
 from pyanaconda.storage_utils import DEVICE_TEXT_PARTITION, DEVICE_TEXT_MAP, DEVICE_TEXT_MD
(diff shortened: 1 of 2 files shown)


pyanaconda/ui/gui/spokes/custom.py
```diff
@@ -1963,16 +1967,33 @@ def on_remove_clicked(self, button):
...
+
+            # We do this even in deleteAll case, see above comment
+            self._destroy_device(device)
 
         log.info("ui: removed device %s", device.name)
 
         # Now that devices have been removed from the installation root,
         # refreshing the display will have the effect of making them disappear.
         # It's like they never existed.
+        self._storage_playground.roots = findExistingInstallations(self._storage_playground.devicetree)
```
AdamWill: This is something of a big hammer, it re-runs the 'find existing OSes' probe. It causes a second or two delay after you confirm the delete. If that's a big problem we could perhaps just directly take the root we're acting on out of `self._storage_playground.roots` with a comprehension or whatever; in theory that'd be wrong if so many partitions are shared that the bucket should stick around, but in practice I don't think that's anything to worry about, since you shouldn't be sharing root partitions...

AdamWill:
https://www.happyassassin.net/temp/updates.img is an updates.img against Beta with this change, for testing. It's not against Final TC1 because detection of existing OSes is broken in that.

AdamWill:
https://youtu.be/e4PIOf8K538 and https://youtu.be/PgpOXgI1f-c are a couple of shakycam videos demonstrating what the results of this patch actually look like, since reproducing the scenarios can be a bit of a pain.

jkonecny12:
This is nice patch and I think this could be in F23. IMHO many people aren't using UEFI still so if they want to delete all partition of selected system and we skip one system it's not so painful. For master I should be more transparent solution where user will see what he/she doing though.

I'm going to PTO from Monday and return next week 12th Oct. I trust in your decision guys so if you want to fix this with your or @dwlehman patch you can push it or you can wait on me.

jkonecny12:
Slightly changed version of this PR https://github.com/rhinstaller/anaconda/pull/512 was pushed to master.
Thank you @AdamWill for the code.

PR #385, opened 2015-10-02
Lookup IPv6 address without brackets (#1267872)

bcl:
The brackets are added for display, gethostbyaddr doesn't like them.

Resolves: rhbz#1267872


vpodzime:
Looks good to me.

bcl:
Pushed.

PR #386, opened 2015-10-05
Add support for other systemd units to kickstart service command

bcl:
.service isn't the only unit type, if the user specifies a specific
unit, don't add .service to the end. If no service is specified, add
.service.


bcl:
> Why not just let systemctl do this? man systemctl ->

I didn't do it that way because we've been adding .service and I'm worried that if we stop it will change how some kickstarts behave. Maybe this is unfounded.

clumens:
I think that's just a holdover from when we were using the services command.  I don't think we'll see too many complaints if we move to just passing straight through to systemd.

bcl:
ok

PR #387, opened 2015-10-05
Add support for other systemd units to kickstart service command

bcl:
.service isn't the only unit type, stop adding it to the end to allow
units like fstrim.timer to be enabled. If there is no .service suffix
systemctl will figure out the right unit to use.

### DIFF
--- a/pyanaconda/kickstart.py
+++ b/pyanaconda/kickstart.py
@@ -1641,15 +1641,9 @@ class SELinux(commands.selinux.FC3_SELinux):
 class Services(commands.services.FC6_Services):
     def execute(self, storage, ksdata, instClass):
         for svc in self.disabled:
-            if not svc.endswith(".service"):
-                svc += ".service"
-
             iutil.execInSysroot("systemctl", ["disable", svc])
 
         for svc in self.enabled:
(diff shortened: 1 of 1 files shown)


vpodzime:
What I see in ``man systemctl`` on my Fedora 21 is this:
```
       For unit file commands, the specified NAME should be the full name of the unit file, or the absolute path to the unit file:
           # systemctl enable foo.service
       or
           # systemctl link /path/to/foo.service
```

vpodzime:
So I think #386 could really be the right way to go in the end.

bcl:
Interesting, that's not how it works though. systemctl enable and disable work without having .service

vpodzime:
> Interesting, that's not how it works though. systemctl enable and disable work without having .service

Maybe the word *should* is really important in there. Thanks for double-checking! I'm okay with this change then.

clumens:
It'd be worth filing a bug about their man page, then.  It seems pretty confusingly written in that aspect.

bcl:
Filed an issue - https://github.com/systemd/systemd/issues/1502

bcl:
pushed

PR #388, opened 2015-10-06
dracut: accept inst.dd=[file:]/dd.iso (#1268792)

wgwoods:
Like with inst.ks=/ks.cfg, some users want to be able to inject the
driver disk into the initramfs and load it from there.

The mechanisms for doing this are already basically present, we just
need to handle it the same way we do when the kickstart is already
inside the initramfs: check to see if the requested file exists, and if
it's already present, use it as-is.

The canonical form of this argument is:

    inst.dd=file:/path/to/dd.iso

`dd.iso` can be an ISO or a driver RPM, which must be injected into the
initramfs.
`file:` may be omitted, although that practice is discouraged.
`path:` can be used instead of `file:`, but that's also discouraged.

Resolves: rhbz#1268792

### DIFF
--- a/dracut/driver-updates-genrules.sh
+++ b/dracut/driver-updates-genrules.sh
@@ -21,10 +21,18 @@ else
     debug_msg "/tmp/dd_disk file was not created"
 fi
 
-# Run driver-updates for LABEL=OEMDRV and any other requested disk
+# Run driver-updates for LABEL=OEMDRV and any other requested disk/image
 for dd in $DD_OEMDRV $DD_DISK; do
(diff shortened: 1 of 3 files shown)


wgwoods:
(note: we should probably add some unit tests for this if possible..? or at least test it before pushing to rhel7-branch..)

bcl:
Tested with inst.dd=/dd.iso, inst.dd=file:/dd.iso, and inst.dd=file:/dd.iso inst.ks=file:/dd.ks and it works fine.

vpodzime:
Looks good to me too.

PR #389, opened 2015-10-07
network: create ifcfg files in tui if needed (#1267526)

rvykydal:
When a device is activated in dracut and ipv4 dhcp fails, while ipv6
autoconfiguration succeeds (it can happen when no ip= is specified and dracut
tries dhcp4 on all devices), ifcfg file is not created in dracut, and NM
creates in-memory connection for the active device. Installer needs to dump
this connection to ifcfg file for network tui.

In the future we want to get rid of this assumption (NICs have ifcfg files in
installer) and work with NM connections directly.


vpodzime:
Looks good to me.

rvykydal:
Pushed

PR #390, opened 2015-10-07
ibft: fix setting dracut boot args for static ibft nic configuration …

rvykydal:
…(#1267526)

ip=ibft was not passed to dracut.

Resolves: rhbz#1267526


vpodzime:
Looks good to me.

rvykydal:
Pushed

PR #394, opened 2015-10-07
Use "Enter" instead of "Return"

dashea:


vpodzime:
Good catches. Looks good tom e.

vojtechtrefny:
Not sure this actually fixes https://bugzilla.redhat.com/show_bug.cgi?id=1264235 -- the problem was that Zanata thinks "<return>" is a tag and there will be the same problem with "<Enter>" -- if you translate "<Enter>" to something else, Zanata will still complain about missing tags. Couldn't we just change the brackets into single quotes or something else that doesn't make it look like a tag?

dashea:
I don't think we should have to change strings just because Zanata isn't very configurable. We do have a lot of actual tags that would be nice to have actually verified, since if they're mangled in translations it can crash anaconda. If we just ensure that Zanata is configured to warn about tag changes instead of treating them as errors, this should not be an obstacle for translators.

But I think you're right that this doesn't actually address the issue. I'll remove the bug number from the commit.

dashea:
Pushed without the bug number.

PR #395, opened 2015-10-08
Do not limit ONBOOT default setting to url and nfs installation metho…

rvykydal:
…ds (#1269264)

It is a change of behavior caused by anaconda rebase: on media installs,
device activated in GUI is not activated automatically after reboot.

Resolves: rhbz#1269246


bcl:
Looks ok to me.

vpodzime:
Looks good to me too.

rvykydal:
Pushed

PR #396, opened 2015-10-08
Rename ./anaconda -> ./anaconda.py to work around coverage.py #425

atodorov:
This is a work around for 
https://bitbucket.org/ned/coveragepy/issues/425/

If we add the symlink anaconda.py then the ./anaconda file is included in the coverage data. This is to unify coverage results with the ones from within the installer (e.g. when doing kickstart install). 

This will make it easier to compare and analyze the differences between the two test runs.


bcl:
I'd rather see us move anaconda to anaconda.py and then install it as anaconda.

vpodzime:
> I'd rather see us move anaconda to anaconda.py and then install it as anaconda.

Sounds like a good idea to me.

atodorov:
Seems to work but I can't figure out how to avoid including ./anaconda in the tarball.

atodorov:
anyone ping. Can we move forward with this one ?

dashea:
Instead of making a second copy of the script in the build directory, I'd rather we just rename it as it's installed. Something like:

```
diff --git a/Makefile.am b/Makefile.am
index cc3974a..9687f08 100644
--- a/Makefile.am
+++ b/Makefile.am
@@ -42,7 +42,10 @@ CLEANFILES = *~
 
 dist_noinst_DATA      = $(PACKAGE_NAME).spec
 
-dist_sbin_SCRIPTS     = anaconda
+dist_sbin_SCRIPTS     = anaconda.py
+
+install-exec-hook:
+       cd $(DESTDIR)$(sbindir) && mv anaconda.py anaconda
 
 ARCHIVE_TAG   = $(PACKAGE_NAME)-$(PACKAGE_VERSION)-$(PACKAGE_RELEASE)
 
```

Plus the other stuff. That probably also needs something on the uninstall side.

anaconda.po would not be needed if the script is renamed, since the only reason we have that is that gettext can't autodetect the language with a .py extension. So all of that could go, including the piece in po/Rules-extract that creates it.

atodorov:
@bcl  - I've modified the patch to remove anaconda.po but I'm not sure if everything is correct. Please review and also probably test locally.

Btw: I saw po/anaconda.pot being updated with strings from anaconda.py after make dist-bzip2 so meybe I have to include this file in the PR as well, thoughts ?

bcl:
Looks good to me, I did a scratch build and a new boot.iso with it just to be sure.

PR #397, opened 2015-10-08
Point coverage.py to the full path of pyanaconda/

atodorov:
There's a problem with how coverage.py discovers all files in pyanaconda. When the source parameter is specified as module name the coverage data includes 109 files. If the source parameter is specified as full directory name then the data files contain 124 files. This is the similar to what `make ci` coverage reports produce.  Unifying those gives more accurate data and makes it easier to compare test executions in the future. 

Note: I had to use the getsitepackages() gimmick (proposed by dlehman) because I didn't wan't to import pyanaconda before coverage was started.

### DIFF
--- a/anaconda
+++ b/anaconda
@@ -31,6 +31,9 @@
 # This toplevel file is a little messy at the moment... (2001-06-22)
 # ...still messy (2013-07-12)
 
+import os
+import site
+
 coverage = None
 
 # If we get a signal immediately after starting that would be pretty messed up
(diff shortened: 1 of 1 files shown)


atodorov:
Coverage issue reported as:
https://bitbucket.org/ned/coveragepy/issues/426/difference-between-coverage-results-with

Note: I have a copy of the resulting coverage reports but GitHub doesn't want to upload them for some reason.

anaconda
```diff
@@ -37,10 +37,18 @@ coverage = None
 proc_cmdline = open("/proc/cmdline", "r").read()    # pylint: disable=interruptible-system-call
 proc_cmdline = proc_cmdline.split()
 if ("inst.debug=1" in proc_cmdline) or ("inst.debug" in proc_cmdline):
+    import os
+    import site
```
vpodzime: I think these two imports could safely go to the global scope of the file.

vpodzime:
Other that the above this looks good to me.

clumens:
Pushed, thanks.

PR #398, opened 2015-10-08
network: don't set NM_CONTROLLED=no for root on SAN.

rvykydal:
NM now takes over connections created in initramfs seamlessly, and our ifcfg is
overwritten by dracut anyway.


vpodzime:
Looks good to me too.

rvykydal:
Pushed

PR #399, opened 2015-10-08
Various user spoke improvements

dashea:


vpodzime:
Other than the question plus a nitpick above these all look good to me. Good job, @dashea!

dashea:
Removed that underscore and pushed.

PR #400, opened 2015-10-08
Kickstart test prereqs

dashea:
The prereqs idea is laying the groundwork for DUD tests that I haven't written yet.

The idea is that shared, or potentially sharable, files can be built by run_kickstart_tests before the tests are run. Building the prereq files is defined in Makefile.prereqs, and selecting the prereqs is handled by the prereqs function in <test>.sh which just echos what it wants.

The liveimg payload is only used by one test, so I could move that the prepare of liveimg.sh if anyone has opinions on that. Either way, it's two fewer environment variables that the kickstart tests need, leaving just the ostree repo and the NFS stuff.

The last two commits are because I tried testing one of the proxy tests and it ran real slow because it wasn't using the local mirror.

### DIFF
--- a/tests/kickstart_tests/functions.sh
+++ b/tests/kickstart_tests/functions.sh
@@ -16,6 +16,11 @@
 #
 # Red Hat Author(s): Chris Lumens <clumens@redhat.com>
 
+prereqs() {
+    # No prereqs by default
+    echo
+}
+
 kernel_args() {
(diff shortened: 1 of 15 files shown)


vpodzime:
Looks good to me too except for the above question.

dashea:
Pushed.

PR #403, opened 2015-10-13
Do not use deprecated getDevicesByInstance method

vojtechtrefny:
getDevicesByInstance has been deprecated in blivet 1.7

### DIFF
--- a/pyanaconda/bootloader.py
+++ b/pyanaconda/bootloader.py
@@ -800,7 +800,8 @@ class BootLoader(object):
         if usr_device:
             dracut_devices.extend([usr_device])
 
-        netdevs = storage.devicetree.getDevicesByInstance(NetworkStorageDevice)
+        netdevs = [d for d in storage.devices if (getattr(d, "complete", True) and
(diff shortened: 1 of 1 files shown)


pyanaconda/bootloader.py
```diff
@@ -800,7 +800,8 @@ def set_boot_args(self, *args, **kwargs):
         if usr_device:
             dracut_devices.extend([usr_device])
 
-        netdevs = storage.devicetree.getDevicesByInstance(NetworkStorageDevice)
+        netdevs = (d for d in storage.devices if (getattr(d, "complete", True) and
+            isinstance(d, NetworkStorageDevice)))
```
vpodzime: This should be a list comprehension because ``netdevs`` is used at least twice in the block below.

vpodzime:
Looks good to me otherwise. I like that this makes things more explicit and visible.

pyanaconda/bootloader.py
vojtechtrefny: Fixed, thanks.

vojtechtrefny:
Pushed.

PR #405, opened 2015-10-13
Show DNF exception instead of silent exit

jkonecny12:


jkonecny12:
@dashea good point but sad is I already tried it. The problem is you lose context when the exception exits the exception handler. Also you can't save stacktrace to queue I tried that too.
This was only working solution I have found.

It would be nice if someone found clearer solution.

vpodzime:
Looks good to me. This is definitely and improvement. If we found some way to get the exception to "the other side" in the future, we can always change it.

jkonecny12:
I added code to improve bug reporting to Bugzilla. Now the DNF exception will be in name of the bug.

Removing ACK -- complete change of code base.

jkonecny12:
Updated PR.
Implemented @vpodzime idea https://github.com/jkonecny12/anaconda/commit/f151da5fadc43fc0921fa6f7b8215635b9dfbaad#commitcomment-13809371 thank you.

vpodzime:
Looks even better to me now. Thanks!

jkonecny12:
Hmm looks like I broke something. With this patch I can't report bug raised by USR1 signal.

I will look on that in more detail why this happening. I will change status of this somehow when I found cause.

jkonecny12:
I did slight changes and now it should work properly as expected. 
Sad thing is that the DNF error is printed twice. First traceback is where we raised the exception and second is traceback of DNF error. I'm going to file a bug on DNF to make the exception shorter [this](https://bugzilla.redhat.com/show_bug.cgi?id=1277485) the packages there are exception description.

Another problem is that this exception has more than one line so we need to take only first line of the exception description. This was filed as bug on libreport because I think they should do this automatically.

jkonecny12:
Please another review of this PR. Thank you

vpodzime:
Looks good to me too.

jkonecny12:
Pushed. Thank you

PR #406, opened 2015-10-13
Use GtkResponseType values in the iso chooser dialog

dashea:
This will resolve 1265060

### DIFF
--- a/pyanaconda/ui/gui/spokes/source.glade
+++ b/pyanaconda/ui/gui/spokes/source.glade
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<!-- Generated with glade 3.18.3 -->
+<!-- Generated with glade 3.19.0 -->
 <interface>
   <requires lib="gtk+" version="3.0"/>
   <requires lib="AnacondaWidgets" version="1.0"/>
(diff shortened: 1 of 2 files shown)


vpodzime:
I think we should do this consistently in all places, not just a single one that is broken right now.

dashea:
> I think we should do this consistently in all places, not just a single one that is broken right now.

Unless glade decides to expose the GtkResponseType values, I don't think it will help. The iso file chooser is the only subclass of GtkDialog that anaconda uses, and for normal GtkDialogs there is no special meaning attached to any of the responses. The response IDs are always something we'll have to hardcode in glade and match in python, and I don't feel like having a "Yes" button return -8 instead of 1 is any better, especially since I had to grep through some header files to figure out that GTK_RESPONSE_YES is -8.

vpodzime:
Oh, in that case I'm definitely for using our own magic numbers we have at least some control of. Thanks for the extra info.

jkonecny12:
What about using `Enum` on our side for this. I known it won't be usable in glade file but we can use it on our side. It will be easier to read the code though.

dashea:
> What about using Enum on our side for this. I known it won't be usable in glade file but we can use it on our side. It will be easier to read the code though.

I started looking at doing this, and we stray from the Gtk stock buttons pretty quickly. Would it be more clear for the network device "Add" button to return Gtk.ResponseType.OK, or APPLY, or should it stay a custom response type? I don't know, so I'll let someone else try to figure out the answer to that.

dashea:
Pushed.

PR #407, opened 2015-10-14
network: handle bridge device appearing before its connection (#1265593)

rvykydal:
For example virbr0 during initial-setup. First the device is created and we add it
to spoke device/configuration list, then its connection appears and we want to
bind it to the device config already present in the list. We were already doing
it for ethernet devices, so do that also for other (virtual) devices.


vpodzime:
Looks good to me.

rvykydal:
Pushed

PR #408, opened 2015-10-14
Hide the places sidebar in the ISO chooser widget.

dashea:
The places make no sense in this context, so do not show them.


vpodzime:
The word ``child`` is a little bit confusing here since the function doesn't really go only over the children of the given widget, but I cannot come up with any better word. So ACK from me as well.

dashea:
Pushed.

PR #409, opened 2015-10-15
decode package name for /etc/sysconfig/kernel (RHBZ #1261569)

AdamWill:
Seems in Python 3 when we get a package name from RPM in this
way it comes out as a bytestring, we need a regular string as
we're going to write it into a text file. Without this fix
/etc/sysconfig/kernel gets a line:

DEFAULTKERNEL=b'kernel-core'

which obviously isn't what we wanted, and means new kernels
sometimes don't get set as the default boot option.

### DIFF
--- a/pyanaconda/bootloader.py
+++ b/pyanaconda/bootloader.py
@@ -2345,7 +2345,7 @@ def writeSysconfigKernel(storage, version, instClass):
         log.error("failed to get package name for default kernel")
         return
 
-    kernel = h.name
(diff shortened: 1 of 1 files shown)


vpodzime:
I wish Python3's ``str()`` returned just the string for a bytestring. Like anybody cared about/wanted the ``b'..'`` enclosing. This looks good to me.

snbueno:
I've pushed this manually to master and f23-branch, so I'm closing this.

PR #410, opened 2015-10-15
Warn user before using Delete all (#1183880)

jkonecny12:
Warn user that "Delete all" checkbox removes even shared partitions.
Thank you for this patch @dwlehman .

*Resolves: rhbz#1183880*

----------------------------------
This *should* be final patch for this bug to F23 based on https://github.com/rhinstaller/anaconda/pull/377#issuecomment-144808251 .

### DIFF
--- a/pyanaconda/ui/gui/spokes/lib/custom_storage_helpers.py
+++ b/pyanaconda/ui/gui/spokes/lib/custom_storage_helpers.py
@@ -383,9 +383,13 @@ class ConfirmDeleteDialog(GUIObject):
 
         if rootName and "_" in rootName:
             rootName = rootName.replace("_", "__")
+        delete_all_text = (C_("GUI|Custom Partitioning|Confirm Delete Dialog",
+                               "Delete _all other file systems in the %s root as well.") +
+                           "\n" +
+                           C_("GUI|Custom Partitioning|Confirm Delete Dialog",
+                               "(This includes those used by other installed operating systems.)"))
         self._removeAll.set_label(
-                C_("GUI|Custom Partitioning|Confirm Delete Dialog",
-                    "Delete _all other file systems in the %s root as well.")
+                delete_all_text
                 % rootName)
         self._removeAll.set_sensitive(rootName is not None)
 


pyanaconda/ui/gui/spokes/lib/custom_storage_helpers.py
```diff
@@ -383,9 +383,13 @@ def refresh(self, mountpoint, device, rootName, snapshots=False):
 
         if rootName and "_" in rootName:
             rootName = rootName.replace("_", "__")
+        delete_all_text = (C_("GUI|Custom Partitioning|Confirm Delete Dialog",
+                               "Delete _all other file systems in the %s root as well.") +
+                           "\n" +
+                           C_("GUI|Custom Partitioning|Confirm Delete Dialog",
+                               "(Including those used by other installed operating systems.)"))
```
dwlehman: It would be a bit more natural sounding if the second string began "This includes those used by..."

AdamWill:
I like this as it at least preserves existing translations of the first line, neat trick. ack from me.

vpodzime:
I agree with both @dwlehman about the wording and @AdamWill about removing the *Resolves* -- you can remove the bug ID from the first line of the commit and change *Resolves* to *Related*. Otherwise this looks good to me.

jkonecny12:
Good improvements. Thank you for the ideas.

Updated PR:
Implemented string change.
Changed to *Related* from *Resolves* and removed `(#bugnumber)` from label.

pyanaconda/ui/gui/spokes/lib/custom_storage_helpers.py
```diff
@@ -383,9 +383,13 @@ def refresh(self, mountpoint, device, rootName, snapshots=False):
 
         if rootName and "_" in rootName:
             rootName = rootName.replace("_", "__")
+        delete_all_text = (C_("GUI|Custom Partitioning|Confirm Delete Dialog",
+                               "Delete _all other file systems in the %s root as well.") +
+                           "\n" +
+                           C_("GUI|Custom Partitioning|Confirm Delete Dialog",
+                               "(This includes those used by other installed operating systems.)"))
```
vpodzime: What about *This includes those used also by other installed operating systems.*? Without the word *also* or something similar it sounds like it will delete all partitions from all operating systems.

jkonecny12: > What about This includes those used also by other installed operating systems.? Without the word also or something similar it sounds like it will delete all partitions from all operating systems.

"... those also used by ..." sounds a bit better to me but is it grammatically correct?

dwlehman: How about "This includes those shared with other installed operating systems."

Or, following more closely @vpodzime suggestion, "This includes those also used by other..."

vpodzime: > How about "This includes those shared with other installed operating systems."

Sounds best to me.

jkonecny12: +1 from me too. So I will change this locally and going to push it.

jkonecny12:
Pushed

PR #411, opened 2015-10-15
Changes in network state revalidate sources rhbz#1270354

jcpunk:
When altering the network state, it may result in selected
network sources becoming inaccessable due to typos in DNS
or GATEWAY.

With this patch, after the network spoke completes it tells
the software payload to revalidate its state.  If the state
becomes invalid, the source spoke errors will guide towards
what is wrong.

Resolves: rhbz#1270354

### DIFF
--- a/pyanaconda/ui/gui/spokes/network.py
+++ b/pyanaconda/ui/gui/spokes/network.py
@@ -313,6 +313,7 @@ class NetworkControlBox(GObject.GObject):
         self.builder = builder
         self._running_nmce = None
         self.spoke = spoke
+        self.settings_changed = False
 
         # button for creating of virtual bond and vlan devices
         self.builder.get_object("add_toolbutton").set_sensitive(True)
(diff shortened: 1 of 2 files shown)


jkonecny12:
Thank you for this patch I think it's valuable. 
It's late for 7.2 but we will consider it to 7.3. It will be good base for nice improvement to Fedora 24 now.

Good job :+1:

M4rtinK:
Also looks good to me, thanks!

jkonecny12:
Hello @jcpunk I finally looked on your PR in more detail and I think it should go to master and rhel branches but with some improvements.
I did patch on your patch here: https://github.com/jkonecny12/anaconda/commit/2b2f23fdb9c4069bfb97b8eba4348bfc0cddc71a .
You can use it or some parts of it and rework this patch or we can push both patches. It's on you.
But if we going to push both patches please fix your bad calling of the thread manager restart method. Without the install class seems the patch broken to me.

@rvykydal could you please take a brief look on the patches too? Thank you

jcpunk:
I've rebased my commit with most of the logic from jkonecny12@2b2f23f but a little bit of additional logging.

rvykydal:
Apart from the previous comment the change looks ok to me.

jkonecny12:
Pushed to master and RHEL7. Thank you for the PR.

PR #412, opened 2015-10-16
network: Don't set --device link default for hostname only network cm…

rvykydal:
…d (#1272274)

network --hostname=blah

used in %pre section could override a device configuration with dhcp which is a
regression caused by commit 7473be3d9c5f0c2b434e4c86231b73e72ffe64f9

Resolves: rhbz#1272274

### DIFF
--- a/pyanaconda/kickstart.py
+++ b/pyanaconda/kickstart.py
@@ -1070,6 +1070,19 @@ class Logging(commands.logging.FC6_Logging):
             logger.updateRemote(remote_server)
 
 class Network(commands.network.RHEL7_Network):
+    def parse(self, args):
+        nd = commands.network.RHEL7_Network.parse(self, args)
+        setting_only_hostname = nd.hostname and len(args) <= 2
+        if not setting_only_hostname:
+            if not nd.device:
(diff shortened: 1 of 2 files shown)


jkonecny12:
Looks good to me

pyanaconda/kickstart.py
```diff
@@ -1070,6 +1070,19 @@ def execute(self, *args):
             logger.updateRemote(remote_server)
 
 class Network(commands.network.RHEL7_Network):
+    def parse(self, args):
+        nd = commands.network.RHEL7_Network.parse(self, args)
+        setting_only_hostname = nd.hostname and len(args) <= 2
```
vpodzime: What are the two args that could be specified here? AFAICT, *network --hostname=blah* has no args.

rvykydal: Args are either
["--hostname=blah']
or
["--hostname", "blah"] for
network --hostname blah

vpodzime: Oh, I always thought these are parsed as options and args are only the extra args not related to any option.

rvykydal:
Pushed

PR #415, opened 2015-10-19
 Cleaner logging of .treeinfo return conditions in dependant function

jcpunk:
Use of the fall through value should be logged a bit more clearly.

### DIFF
--- a/pyanaconda/packaging/__init__.py
+++ b/pyanaconda/packaging/__init__.py
@@ -405,8 +405,10 @@ class Payload(object):
                 version = "rawhide"
             except configparser.Error:
                 pass
+            log.debug("using treeinfo release version of %s", version)
+        else:
+            log.debug("using default release version of %s", version)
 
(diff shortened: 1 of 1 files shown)


pyanaconda/packaging/__init__.py
```diff
@@ -405,8 +405,10 @@ def _getReleaseVersion(self, url):
                 version = "rawhide"
             except configparser.Error:
                 pass
+            log.debug("got a release version of %s", version)
+        else:
+            log.debug("no treeinfo returned, using fallthrough of %s", version)
```
bcl: I'd word them like this:
"using release version %s from treeinfo" and "using default release version of %s"

that way you can search the logs for 'release version' and find either result.

jcpunk:
Updated logging messages

vpodzime:
Looks good to me too.

clumens:
Merged, thanks.

PR #416, opened 2015-10-19
Don't set BOOTPROTO= when it isn't set

josefbacik:
BOOTPROTO= is ipv4 specific, so if we don't have an ipv4 configuration this just
confuses network manager.  With the fix to pykickstart we no longer set
net.bootProto if we haven't specified one, so don't write out a blank one to the
ifcfg script.

Signed-off-by: Josef Bacik <jbacik@fb.com>


vpodzime:
Looks good to me.

jkonecny12:
Looks good to me too.

josefbacik:
Needed to update the patch, I was testing against my old pykickstart.

PR #422, opened 2015-10-21
Do not run time_initialize for image and directory installations

bcl:


vpodzime:
Looks good to me.

AdamWill:
Um, does `can_touch_runtime_system("initialize time", touch_live=True)` imply `not flags.dirInstall`? If not, isn't this changing logic?

bcl:
Pushed

PR #423, opened 2015-10-23
Improve install space required estimation (#1224048)

jkonecny12:
The problem was that 35% add is not enough on the smaller installations.
That result in crash on RPM transaction check. RPM required more space
than we have.
This problem was consulted with RPM team and this is result.
The RPM rounding size of files to fragment size for better estimation
also they are required package header size * 4 for RPM DB.

We don't have header size or file size in the installation so I'm trying
to simulate worst case scenario for file rounding and using 2KiB for
every file to get something like RPM DB space required.
Also 10% is add to all this as a safeguard.

Seems as stable assumption right now.
+
Improved logging for packaging.

Resolves: rhbz#1224048

### DIFF
--- a/pyanaconda/packaging/dnfpayload.py
+++ b/pyanaconda/packaging/dnfpayload.py
@@ -71,6 +71,12 @@ REPO_DIRS = ['/etc/yum.repos.d',
              '/tmp/product/anaconda.repos.d']
 YUM_REPOS_DIR = "/etc/yum.repos.d/"
 
+# Bonus to required free space which depends on block size and rpm database size estimation.
+# Every file could be aligned to fragment size so 4KiB * number_of_files should be a worst
+# case scenario. 2KiB for RPM DB was acquired by testing.
+# 6KiB = 4K(max default fragment size) + 2K(rpm db could be taken for a header file)
+BONUS_SIZE_ON_FILE = Size("6 KiB")
+
 _DNF_INSTALLER_LANGPACK_CONF = DNF_PLUGINCONF_DIR + "/langpacks.conf"
 _DNF_TARGET_LANGPACK_CONF = "/etc/dnf/plugins/langpacks.conf"
 
@@ -115,8 +121,9 @@ def _pick_mpoint(df, download_size, install_size):
                    # for root we need to take in count both download and install size
                    if ((key != root_mpoint and val > requested)
                    or val > requested_root) and reasonable_mpoint(key)}
-    log.debug('Estimated size: download %s & install %s - df: %s', requested,
-              (requested_root - requested), df)
+    log.debug('Input mount points: %s', df)
+    log.debug('Estimated size: download %s & install %s', requested,
+              (requested_root - requested))
     log.info('Sufficient mountpoints found: %s', sufficients)
 
     if not len(sufficients):
@@ -555,11 +562,14 @@ class DNFPayload(packaging.PackagePayload):
             if key.startswith('/') and ((root_mpoint + new_key) not in valid_points):
                 valid_points[root_mpoint + new_key] = val.format.freeSpaceEstimate(val.size)
 
-        m_points = _pick_mpoint(valid_points, download_size, size)
-        if not m_points or m_points == root_mpoint:
+        m_point = _pick_mpoint(valid_points, download_size, size)
+        if not m_point or m_point == root_mpoint:
             # download and install to the same mount point
             size = size + download_size
-        log.debug("Instalation space required %s for mpoints %s", size, m_points)
+            log.debug("Install + download space required %s", size)
+        else:
+            log.debug("Download space required %s for mpoint %s (non-chroot)", download_size, m_point)
+            log.debug("Installation space required %s", size)
         return size
 
     def _spaceRequired(self):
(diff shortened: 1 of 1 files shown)


pyanaconda/packaging/dnfpayload.py
```diff
@@ -115,8 +115,8 @@ def reasonable_mpoint(mpoint):
                    # for root we need to take in count both download and install size
                    if ((key != root_mpoint and val > requested)
                    or val > requested_root) and reasonable_mpoint(key)}
-    log.debug('Estimated size: download %s & install %s - df: %s', requested,
-              (requested_root - requested), df)
+    log.debug('Estimated size: download %s & install %s', requested,
+              (requested_root - requested))
```
vpodzime: I don't quite see how this accounts for improved logging. ?

pyanaconda/packaging/dnfpayload.py
```diff
@@ -555,11 +555,14 @@ def spaceRequired(self):
...
-        m_points = _pick_mpoint(valid_points, download_size, size)
-        if not m_points or m_points == root_mpoint:
+        m_point = _pick_mpoint(valid_points, download_size, size)
+        if not m_point or m_point == root_mpoint:
             # download and install to the same mount point
             size = size + download_size
-        log.debug("Instalation space required %s for mpoints %s", size, m_points)
+            log.debug("Install + download space required %s", size)
+        else:
+            log.debug("Download space required %s for mpoint %s (non-chroot)", download_size, m_point)
```
vpodzime: *non-root* would be more precise here, but then you see it's obvious from the mount point so I think it could be dropped

pyanaconda/packaging/dnfpayload.py
```diff
@@ -567,11 +570,23 @@ def _spaceRequired(self):
...
+        size = 0
+        files_nm = 0
+        for tsi in transaction:
+            # space taken by all files installed by the packages
+            size += tsi.installed.installsize
+            # number of files installed on the system
+            files_nm += len(tsi.installed.files)
+        # add required space which depends on block size and rpm database taken space estimation
+        # 6144 = 4K(max default fragment size) + 2K(rpm db could be taken for a header file)
+        bonus_size = files_nm * 6144
```
vpodzime: This ``6144`` should be constant defined at the top of the file with appropriate name and, what's more important, it should be ``Size("6 KiB")``.

pyanaconda/packaging/dnfpayload.py
```diff
@@ -567,11 +570,23 @@ def _spaceRequired(self):
...
+        size = size + bonus_size
```
vpodzime: And ``size`` should be an instance of ``Size`` here.

pyanaconda/packaging/dnfpayload.py
```diff
@@ -567,11 +570,23 @@ def _spaceRequired(self):
...
+        # add another 10% as safeguard
+        total_space = Size(size) * 1.1
+        log.debug("Size %s files %s", Size(size), files_nm)
+        log.debug("Bonus size %s B", bonus_size)
+        log.debug("Complete size required %s", total_space)
```
vpodzime: The first message could better be something like *"Size of %s packages: %s"*, the second message could drop the unit if ``bonus_size`` is a ``Size`` instance and *Total size required* sounds better to me than *Complete...*

vpodzime:
Looks good to me other than the nitpicks above.

pyanaconda/packaging/dnfpayload.py
jkonecny12: This will remove 3 or more lines in the log of filesystem mountpoint which won't be used for downloading. 
I want to make logs more readable so I removed this. If we want these logs we want better formating or something though. I can try to create formatted version of this.

pyanaconda/packaging/dnfpayload.py
jkonecny12: I don't think so. The problem is that you can pick download location as '/' it's totally valid and it then downloads to ram where the stage 2 is loaded. (I hope I said it right). 
So or so when you get log "Download space required 650MiB for mpoint /" will you know it's stage 2 and not installed root after few year from now? I want to prevent confusion from logs.

pyanaconda/packaging/dnfpayload.py
jkonecny12: True I will update PR.

pyanaconda/packaging/dnfpayload.py
jkonecny12: Also true.

pyanaconda/packaging/dnfpayload.py
jkonecny12: 1) packages != files -- but I can change it to "Size: %s number of files: %s" or similar
2) Good point, will be changed
3) True, I somehow missed it when I changed complete_space variable to total_space. Will be changed.

jkonecny12:
Thank you for the valuable inputs @vpodzime . Really nice improvements.
I will wait until you reply and then I will update this PR. To prevent comments loss.

pyanaconda/packaging/dnfpayload.py
vpodzime: Makes sense, thanks for the extra explanation.

pyanaconda/packaging/dnfpayload.py
vpodzime: ok

pyanaconda/packaging/dnfpayload.py
vpodzime: Just please a comma or semicolon in the middle of the string in 1).

vpodzime:
/me is looking forward to seeing the updated version. :)

jkonecny12:
I updated the PR, thank you @vpodzime for the inputs. I hope I didn't forgot something.

Please look on it if it's OK.

vpodzime:
Looks good to me now. Thanks!

jkonecny12:
Pushed

PR #424, opened 2015-10-23
Fix bad space needed messages

jkonecny12:
When we reduce metadata from device size in space check we also
need to tell user how much space he needs on root to get installation
done. That means we need to add metadata in count when calculating this
space deficit.

I need this PR: https://github.com/rhinstaller/blivet/pull/255 in blivet to make this work.

### DIFF
--- a/pyanaconda/ui/gui/spokes/lib/resize.py
+++ b/pyanaconda/ui/gui/spokes/lib/resize.py
@@ -32,6 +32,7 @@ from pyanaconda.i18n import _, C_, N_, P_
 from pyanaconda.ui.gui import GUIObject
 from pyanaconda.ui.gui.utils import blockedHandler, escape_markup, timed_action
 from blivet.size import Size
+from blivet.formats.fs import FS
 
 __all__ = ["ResizeDialog"]
 
@@ -84,7 +85,8 @@ class ResizeDialog(GUIObject):
 
         self._required_label = self.builder.get_object("requiredSpaceLabel")
         markup = self._required_label.get_label()
-        self._required_label.set_markup(markup % escape_markup(str(self.payload.spaceRequired)))
+        required_dev_size = self.payload.requiredDeviceSize(FS.biggestOverheadFS())
+        self._required_label.set_markup(markup % escape_markup(str(required_dev_size)))
 
         self._reclaimDescLabel = self.builder.get_object("reclaimDescLabel")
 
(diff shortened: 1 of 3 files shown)


jkonecny12:
I need to test this more, but seems working right now. I will do more tests on Monday.

pyanaconda/ui/gui/spokes/lib/resize.py
```diff
@@ -95,6 +97,13 @@ def __init__(self, data, storage, payload):
         self._deleteButton = self.builder.get_object("deleteButton")
         self._resizeSlider = self.builder.get_object("resizeSlider")
 
+    def _updateRequiredSpace(self):
+        space_required = self.payload.spaceRequired
+        # get real required space
+        # BTRFS is used here because it's FS with biggest space taken by metadata
+        space_required = Size(Decimal(space_required) / Decimal(BTRFS.MetadataSizeFactor))
+        return space_required.roundToNearest(Size("1 MiB"), rounding=ROUND_UP)
+
```
vpodzime: A few suggestions here:
  * the newly added thing should be a property
  * and it should have a name that doesn't sound like does something else than calculating and returning a value (maybe something like ``_payloadRequiredSize``?)
  * I don't think working with BTRFS explicitly here is a good thing to do -- BTRFS' meta data requirements may change in a week and then it will be a different file system we would need to take into account here. So blivet should probably provide something like *the_most_hungry_FS* (giving the class and the size md size factor) and we should use that here

pyanaconda/ui/lib/space.py
```diff
@@ -77,7 +79,9 @@ def check(self):
         log.info("fs space: %s  needed: %s", free, needed)
         self.success = (free > needed)
         if not self.success:
-            self.deficit = needed - free
+            required_space = Size(Decimal(needed) / Decimal(self.storage.rootDevice.format.MetadataSizeFactor))
```
vpodzime: Should these two really be divided? It doesn't make much sense to me. Also, this calculation should be put into a method with a reasonable name.

jkonecny12:
I have updated this PR and created new PR for blivet https://github.com/rhinstaller/blivet/pull/265 . 
This should be better solution.

vpodzime:
Looks good to me.

vpodzime:
Just don't forget to update it with better names of Blivet's newly added methods.

jkonecny12:
Updated PR to be compatible with blivet PR update.

jkonecny12:
Pushed

PR #426, opened 2015-10-23
add kickstart_tests/driverdisk-auto

wgwoods:
This test does two things:
1. Test that the installer auto-loads driver disks labeled OEMDRV
2. Test that the installer can load multiple driver disks

Related: rhbz#1268792

### DIFF
--- a/tests/kickstart_tests/driverdisk-auto.sh
+++ b/tests/kickstart_tests/driverdisk-auto.sh
@@ -1,0 +1,34 @@
+# Copyright (c) 2015 Red Hat, Inc.
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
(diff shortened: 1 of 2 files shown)


tests/kickstart_tests/driverdisk-auto.sh
```diff
@@ -0,0 +1,34 @@
...
+    # main disk
+    qemu-img create -q -f qcow2 ${diskdir}/disk-a.img 10G
+    echo "${diskdir}/disk-a.img"
+
+    # OEMDRV driverdisk containing kernel modules
+    ${KSTESTDIR}/../lib/mkdud.py -k -L OEMDRV ${diskdir}/oemdrv.iso >/dev/null
+    echo "${diskdir}/oemdrv.iso,device=cdrom,readonly=on"
+
+    # second driverdisk contains binaries
+    ${KSTESTDIR}/../lib/mkdud.py -k -L TESTDD2 ${diskdir}/dd2.iso >/dev/null
```
vpodzime: Looks like it should be ``-b`` instead of ``-k`` here.

vpodzime:
Looks good to me otherwise.

tests/kickstart_tests/driverdisk-auto.sh
wgwoods: need an `echo ... dd2.iso` line here too.

jkonecny12:
Hi Will,

I know this is *late* reply, however, this test would still be handy. Are you willing to rebase it to https://github.com/rhinstaller/kickstart-tests and do changes to be able to use it there.

It should be enough to just change a path to `mkdud` binary.

When I tested this in kickstart tests I've got the error from modprobe`: ``Could not get modinfo from fake_dd: Exec format error``, I need to look on this closer. For this PR it will be enough to just rebase it and get it run to this point.

Of course, if you have time and want to look on that I'm not against it ;)

PR #427, opened 2015-10-25
rpmostreepayload: Fix two issues with mounting

cgwalters:
This makes the internal mounts show up a lot more cleanly
in the output of `mount` at least after installation.

This might fix:
https://bugzilla.redhat.com/show_bug.cgi?id=1274977
as it's possible Blivet/systemd will be less confused.

Basically I believe that becuase later mounts were traversing the
`/mnt/sysroot` -> `$deploy/sysroot` bind mount, they were being
`canonicalized` to point into the target.  This messed up the
hierarchy.

That's one bug - the other silly bug is that we were missing a `-o
remount` option for the `/usr` bind mount, causing it to be mounted
twice.

### DIFF
--- a/pyanaconda/packaging/rpmostreepayload.py
+++ b/pyanaconda/packaging/rpmostreepayload.py
@@ -232,9 +232,7 @@ class RPMOSTreePayload(ArchivePayload):
 
         # Set up bind mounts as if we've booted the target system, so
         # that %post script work inside the target.
-        binds = [(iutil.getTargetPhysicalRoot(),
-                  iutil.getSysroot() + '/sysroot'),
-                 (varroot,
+        binds = [(varroot,
                   iutil.getSysroot() + '/var'),
                  (iutil.getSysroot() + '/usr', None)]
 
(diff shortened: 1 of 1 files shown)


vpodzime:
Looks good to me, but I don't feel like I have a good enough insight into these things. So no ACK from me.

cgwalters:
One high level question I was uncertain about with this - should we use blivet to set up these mounts?  I only took a quick look, but I didn't see bind mount support in blivet.  An additional concern I had was that we *don't* want these mounts written out to the target `/etc/fstab`.

What we could still do is on top of this add similar logic from https://github.com/rhinstaller/anaconda/pull/370 to ensure they're unmounted before we do a reboot, if necessary.  I wonder if there are code path differences WRT blivet between livemedia-creator and normal anaconda that trigger the above bug.

cgwalters:
Pushed an additional commit :arrow_up:

dwlehman:
Here's an example of a bind mount using blivet:

```python
# set up a format that can do a bind mount
fmt = getFormat("bind", device=source_dir, mountpoint=dest_dir, exists=True)
# optionally, associate it with a device
dev = DirectoryDevice(dest_dir, fmt=fmt, exists=True)
# mount it
fmt.setup()
```

Anyway, I think your solution is fine -- especially if you don't want this stuff in /etc/fstab.

clumens:
Pushed, thanks.

PR #429, opened 2015-10-27
Ignore interfaces with invalid VLAN IDs.

dashea:
There are some special cases of interface names where something follows
the VLAN ID, such as -fcoe. We can't interact with these via
NetworkManager so raise SettingsNotFound instead.

Resolves: rhbz#1274893


pyanaconda/nm.py
```diff
@@ -634,7 +634,16 @@ def _device_settings(name):
...
             _parent, _sep, vlanid = name.partition(".")
-            settings = _find_settings(int(vlanid), 'vlan', 'id')
+
+            # If we are not able to convert the VLAN id to an int this
+            # is probably a FCoE interface and we're not going to be able
+            # to do much with it.
+            try:
+                vlanid = int(vlanid)
+            except ValueError:
+                return None
```
bcl: I think this should be []

bcl:
The commit message doesn't match the behavior, you aren't raising anything.

pyanaconda/nm.py
dashea: ok

dashea:
> The commit message doesn't match the behavior, you aren't raising anything.

All uses of _device_settings raise SettingsNotFoundError is _device_settings finds nothing, so from the point of view of users external to nm.py the commit changes a ValueError, which is a surprise, to a SettingsNotFound error, which is not.

dashea:
Changed the return from None to [] and amended the commit message.

bcl:
Ah, ok. thanks.

PR #432, opened 2015-10-30
Be more careful with incomplete device types (#1256582)

dashea:
Switch to isinstance to detect MD device types so that we correctly
report on subclasses of mdarray, and add another error string for
incomplete devices where we do not know how to count the members.


vpodzime:
Looks good to me too.

dashea:
Pushed.

PR #433, opened 2015-10-30
Use py3.4 crypt and salt

bcl:
There was some question as to whether /dev/urandom was being used for
generating user password salts, this switches to using the Python 3.4
crypt.mksalt which uses the /dev/urandom


vpodzime:
Nice! Thanks!

bcl:
Pushed.

PR #436, opened 2015-11-03
Improvements for running at low resolutions

dashea:


vpodzime:
Looks good to me.

jkonecny12:
In the first commit comment: " *Also also* the disk label column to be shrunk in case the user doesn't care about all of that."

I think there is miss-type and should be only one *also*.

jkonecny12:
Other than that seems good to me.

dashea:
Pushed.

PR #437, opened 2015-11-03
allow repo with only a name if it's a pre-defined one (#1277638)

AdamWill:
this recently-added check is too broad; a repo without a URL
or mirrorlist is allowed if the name is one of the pre-defined
repos in /etc/anaconda.repos.d , and in fact this style is
specifically mentioned in the documentation, as the way to
use one of the pre-defined-but-disabled-by-default repos.

### DIFF
--- a/pyanaconda/packaging/dnfpayload.py
+++ b/pyanaconda/packaging/dnfpayload.py
@@ -876,8 +876,9 @@ class DNFPayload(packaging.PackagePayload):
             log.debug("repo %s: mirrorlist %s, baseurl %s",
                       ksrepo.name, ksrepo.mirrorlist, ksrepo.baseurl)
             # one of these must be set to create new repo
-            if not (ksrepo.mirrorlist or ksrepo.baseurl):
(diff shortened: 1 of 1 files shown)


AdamWill:
I can try adding a kickstart test for this, but I've never worked with that stuff before so it's gonna take me a bit longer than the patch did :) I tested this with both a 'good' name-only repo ("updates") and a 'bad' name-only repo ("foobarmonkeys") and it does the appropriate thing in each case: in the first case it uses it properly, in the second case it errors as intended.

jkonecny12:
Sorry I missed that option. 
This PR looks good to me. Thank you.

clumens:
Adam - It's not hard.  If you can whip up a way of verifying whether anaconda used the updates repo in a %post, I can do the rest.  If you want to do the whole thing, you can basically just look at tests/kickstart_tests/ and use those as templates.  I've also got a presentation on it if you want to look through.

clumens:
Pushed, thanks.

PR #439, opened 2015-11-04
Glade tests

dashea:
Hopefully this will make it easier to add more glade tests.

As far as the translations: the markup test does catch one error in Hungarian (<bFigyelmeztetés</b>:Minden tárolóval kapcsolatos ...) so if someone could fix that it would be great, and there are a boatload of KeyError warnings because of a bug in python-polib that I'm still trying to figure out.


dashea:
...or maybe there is no polib bug. pocketlint builds the dict out of translated_entries(), so all those KeyErrors just means that there is no translation. So I think I can just remove that warning.

jkonecny12:
As far as I can tell looks good to me but I'm not giving ACK this wants more eye though.

dashea:
Updated to remove the warnings printed from untranslated strings, and simplified wantClass since it turned out my type problems were the result of earlier mistakes and I really can just compare types the normal way now.

Also added a commit to add tests/lib to $PYTHONPATH during the commit hook

vpodzime:
Other than the above tips these all look good to me.

dashea:
Added one more commit to get jenkins to try to use translations during the tests. That might require some configuration on the jenkins slaves.

vpodzime:
Looks like a nice "best-effort" thing to me. So still ACK.

tests/glade/check_hidden_translatable.py
```diff
@@ -0,0 +1,52 @@
...
+
+        # Look for
+        # <object class="GtkTreeView">
+        #   <property name="headers_visible">False</property>
+        #   ...
+        #   <child>
+        #     <object class="GtkTreeViewColumn">
+        #       <property name="title" translatable="yes">...
+
+        for translatable in glade_tree.xpath(".//object[@class='GtkTreeView' and ./property[@name='headers_visible' and ./text() = 'False']]/child/object[@class='GtkTreeViewColumn']/property[@name='title' and @translatable='yes']"):
```
jkonecny12: I'm thinking that these rows (one more few lines lower) could be split to shorter parts. Maybe it would be nice to store this in variable first before the for loop.

jkonecny12:
Other than one small idea, looks good to me.

PR #440, opened 2015-11-06
Use /usr/bin/mock for systems where /usr/sbin is listed before /usr/bin in PATH

atodorov:
When I try `make ci` on a Rawhide system I get the following error:

```
ERROR: [Errno 1] Operation not permitted

ERROR: The most common cause for this error is trying to run /usr/sbin/mock as an unprivileged user.
ERROR: Check your path to make sure that /usr/bin/ is listed before /usr/sbin, or manually run /usr/bin/mock to see if that fixes this problem.

$ echo $PATH
/usr/lib64/qt-3.3/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/root/bin

$ cat /etc/redhat-release 
Fedora release 24 (Rawhide)
```

Using the full path `/usr/bin/mock` fixes this.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -89,9 +89,9 @@ release:
 	$(MAKE) dist && $(MAKE) tag && git checkout -- $(srcdir)/po/$(PACKAGE_NAME).pot
 
 rc-release: scratch-bumpver scratch
-	mock -r $(MOCKCHROOT) --scrub all || exit 1
-	mock -r $(MOCKCHROOT) --buildsrpm  --spec ./$(PACKAGE_NAME).spec --sources . --resultdir $(PWD) || exit 1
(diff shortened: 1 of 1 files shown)


dashea:
I see this more as a problem with $PATH than anything. This is going to be an issue anything that uses consolehelper to gain privileges, since it depends on /usr/bin being earlier in the path for unprivileged users. Is the PATH something that you set up or something installed by Fedora?

atodorov:
@dashea - thanks for the hint. I've found that /etc/profile compiles the path differently based on whether we're root or regular user. I'm ssh'ing as root and then su as regular user which keeps the path settings for root. Closing.

PR #442, opened 2015-11-07
Add --sshkey to kickstart sshpw command

bcl:


bcl:
Pushed to master.

utils/handle-sshpw
```diff
@@ -46,9 +46,15 @@ u = users.Users()
...
+                              isCrypted=ud.isCrypted, lock=ud.lock)
     else:
         kwargs = ud.__dict__
         kwargs.update({"root": "/"})
         u.createUser(ud.username, **kwargs)
+
+    if ud.sshkey:
+        # Setup the account so that only the sshkey can be used
+        u.setUserPassword(username=ud.username, password="*", isCrypted=True, lock=False)
+        u.setUserSshKey(username=ud.username, key=ud.password, root="/")
```
vpodzime: should the key here really be ``ud.password``? I'd expect it to be ``ud.sshkey``.

clumens:
Yeah, in pykickstart the --sshkey is just a flag that says to interpret self.password as a key instead of a password.  There's only one place the key/password is stored.  Check out https://github.com/rhinstaller/pykickstart/commit/b1d1d1e80f94064b1124315f0661d3f50a02e01a.

vpodzime:
> Yeah, in pykickstart the --sshkey is just a flag that says to interpret self.password as a key instead of a password. There's only one place the key/password is stored. Check out rhinstaller/pykickstart@b1d1d1e.

Oh, I see it now. Thanks for the pointer!

PR #443, opened 2015-11-07
Don't clear the _currentIsoFile if another iso was selected

bcl:
It was clearing the selection even if another iso was selected.  This
resulted in the summary hub displaying 'Error setting up ISO file', but
still allowing installation to continue. It should only do that if
something other than an iso is selected.


jkonecny12:
Looks good to me.

bcl:
Pushed

PR #444, opened 2015-11-09
Remove uses of broad-except.

dashea:
Catching Exception hides bugs and we should avoid it whenever possible.
Keep the broad excepts in TUI since it's using them to send the
exception elsewhere, and be more particular everywhere else.


jkonecny12:
Looks good to me.

dashea:
Pushed.

PR #445, opened 2015-11-09
Cherry-pick patches for hd iso as the stage2/repo

bcl:


vpodzime:
These both look good to me.

bcl:
Pushed

PR #446, opened 2015-11-10
Always quote values in ifcfg- files (#1279131)

bcl:
parse-kickstart doesn't use SimpleConfig and wasn't quoting everything.
These files are shell executables so everything needs to be quoted to
prevent accidentally trying to run unexpected binaries.

Resolves: rhbz#1279131


jkonecny12:
Looks good to me.

vpodzime:
One would expect the ``shlex`` module to have a function for this, but there seems to be none. Looks good to me too.

bcl:
> One would expect the shlex module to have a function for this, but there seems to be none. Looks good to me too.

In pyanaconda we use SimpleConfig, but to get that into dracut we'd have to pull everything into the initrd :)

bcl:
pushed to master.

bcl:
Pushed.

PR #447, opened 2015-11-10
Include original kickstart in logs (#1227939)

bcl:
Having the original un-modified kickstart can sometimes be useful,
include it in /var/log/anaconda/ks.cfg

Resolves: rhbz#1227939

### DIFF
--- a/data/post-scripts/99-copy-logs.ks
+++ b/data/post-scripts/99-copy-logs.ks
@@ -9,4 +9,6 @@ cp /tmp/ks-script*.log $ANA_INSTALL_PATH/var/log/anaconda/
 journalctl -b > $ANA_INSTALL_PATH/var/log/anaconda/journal.log
 chmod 0600 $ANA_INSTALL_PATH/var/log/anaconda/*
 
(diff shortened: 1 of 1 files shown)


clumens:
One downside to this is that we're going to be grabbing root passwords, SSH keys, and who knows what else and stuffing it into the logs.  We take some pains to avoid this in other places by blanking out elements of the ksdata, but we won't have that same ability here if we're just copying the file over.

jkonecny12:
In much better in root but so or so maybe we want to erase these fields from the *original* version too.
Maybe to replace these with **removed** string or something similar.

vpodzime:
I think it's safe to have the original file unchanged in the /root. That's where we put the modified/resulting kickstart file anyway.

bcl:
pushed

jkonecny12:
Why are you reopened this PR?

bcl:
> Why are you reopened this PR?

I closed it on accident. I'm leaving things for rhel7-branch open until they get pushed there as well.

bcl:
Pushed.

PR #448, opened 2015-11-10
Allow a user's primary group to be created in --groups (#1279041)

dashea:
If, for example, a user is created with

  --gid=1000 --groups='newgroup(1000)'

recognize that the user's GID should be that of the newly created group
instead of trying to create a separate user group and crashing.

### DIFF
--- a/pyanaconda/users.py
+++ b/pyanaconda/users.py
@@ -270,10 +270,23 @@ class Users(object):
 
         args = ["-R", root]
 
-        # If a specific gid is requested, and that gid does not exist,
-        # create the user group with that GID.
+        # Split the groups argument into a list of (username, gid or None) tuples
+        # the gid, if any, is a string since that makes things simpler
+        group_gids = [GROUPLIST_FANCY_PARSE.match(group).groups()
+                      for group in kwargs.get("groups", [])]
+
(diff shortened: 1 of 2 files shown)


pyanaconda/users.py
```diff
@@ -270,10 +270,23 @@ def createUser(self, user_name, *args, **kwargs):
...
+        #     the GID.
+        #   - check if a new group is requested with that GID. If so, set the GID
+        #     and let the block below create the actual group.
+        #   - if neither of those are true, create a new user group with the requested
+        #     GID
+        # otherwise use -U to create a new user group with the next available GID.
         if kwargs.get("gid", None):
-            if not self._getgrgid(kwargs['gid'], root):
+            if not self._getgrgid(kwargs['gid'], root) and \
+                    not any(filter(lambda x: x[1] == str(kwargs['gid']), group_gids)):
```
vpodzime: Could be a matter of personal preferences, but I really like this more:
``any(gid[1] == str(kwargs['gid']) for gid in group_gids)``

(and it should do the same, AFAICT)

vpodzime:
Looks good to me too except for the above suggestion.

dashea:
Removed the unnecessary filter and pushed.

PR #449, opened 2015-11-10
Set the window icon

dashea:
Non-GNOME live window managers have a taskbar with an icon, and right
now that icon is the default window icon. fedora-logos already contains
an icon that we use for the liveinst launcher, so use it in this
context, too.

### DIFF
--- a/pyanaconda/ui/gui/__init__.py
+++ b/pyanaconda/ui/gui/__init__.py
@@ -396,6 +396,10 @@ class MainWindow(Gtk.Window):
         # inside the window, the titlebar text is redundant and should be disabled.
         self.set_title("")
 
+        # Set the icon used in the taskbar of window managers that have a taskbar
+        # The "anaconda" icon is part of fedora-logos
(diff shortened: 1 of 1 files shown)


dashea:
This might need a closer look. set_icon_name works in Xfce, but KDE still shows a question mark in the taskbar for some reason.

dashea:
I took a closer look, and my conclusion is I don't care about KDE. This line improves the situation in each of the other live spins (Cinnamon, LXDE, Mate, Xfce) by changing the icon in the taskbar from a default empty window icon to the liveinst icon. There is no change in Workstation/GNOME live since it was smart enough to get the icon information from the desktop file to begin with and setting the icon does not break that. In non-live there is no visible change since we go to some lengths to hide all that window managery stuff, and it also does not matter whether the icon is available or not. set_icon_name won't crash if it can't find the icon.

So KDE is busted and that's not our problem.

vpodzime:
Sounds and looks good to me.

dashea:
Pushed.

PR #450, opened 2015-11-11
Make sure unicode in kickstart works.

dashea:
Add a happy unicode snowman to the end of Kosieh's name and make sure it
makes it to /etc/passwd.


vpodzime:
Nice!

dashea:
Pushed.

PR #451, opened 2015-11-12
strip spaces from extlinux label and default (#ticket)

bcl:
Spaces are not allowed, if other kernels are added it will always select
the first one, not the one indicated by the default entry.


vpodzime:
Looks good to me.

bcl:
pushed

PR #453, opened 2015-11-13
rhel6 mk images fixes

tcely:
Fixed a pain point we're seeing on our CentOS 6 build machines.


rvykydal:
For RHEL branches, we require a specific bug filed against the respective RHEL
product so that we can properly triage it.
The RHEL 6 version is now in Production 3 Phase where only a very limited scope of changes is approved.

PR #454, opened 2015-11-13
Collect test-suite.log from all 'make check' invocations. Closes #452

### LINKED ISSUE #452: Multiple calls to make check overwrites test-suite.log
`make ci` calls `make check` several times which overwrites test-suite.log every time. In the end we've lost the nice summary of how many tests were executed and what their status was. 

I've tried adding a pid suffix to TEST_SUITE_LOG but that doesn't work b/c of the way make calls its internal targets. For example this will not work if you do `make test`

```
PID ?= $(shell pidof -s make)
LOG = test.log.$(shell pidof make)

test:
	echo $(LOG)
	$(MAKE) $(LOG)

$(LOG):
	echo $(LOG)
```

Another issue is that everything needs to pass if we're to collect output of pid stamped test-suite.log at the end. Otherwise make will simply abort after the failure. 

Any ideas how this can be fixed ?

atodorov:
Thanks to @dashea for the tip.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -80,7 +80,7 @@ tests/*.log
 tests/*.trs
 tests/.coverage.*
 tests/pylint/.pylint.d
-tests/test-suite.log
+tests/test-suite.log*
 updates.img
 utils/dd/dd_extract
 utils/dd/dd_list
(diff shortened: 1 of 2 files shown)


clumens:
Merged, thanks.

PR #455, opened 2015-11-17
Replace execReadlines with check_output in parse-kickstart_test.py

bcl:
When running from the cmdline with:
PYTHONPATH=../pykickstart/ ./tests/nosetests.sh ./tests/dracut_tests/parse-kickstart_test.py

The PYTHONPATH would get removed, parse-kickstart would fail, but no
traceback was visible because execReadlines raised an error instead of
processing the output.

### DIFF
--- a/tests/dracut_tests/parse-kickstart_test.py
+++ b/tests/dracut_tests/parse-kickstart_test.py
@@ -24,7 +24,7 @@ import os
 import unittest
 import tempfile
 import shutil
-from pyanaconda import iutil
+import subprocess
 
 class BaseTestCase(unittest.TestCase):
     def setUp(self):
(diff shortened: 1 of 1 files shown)


vpodzime:
So the exception produced by ``subprocess.check_output()`` contains the output? Otherwise I don't quite understand what is the difference between this and moving the ``execReadlines()`` call into a try-except block.

bcl:
No, parse-kickstart has the traceback so it's on stdin for execReadlines but since parse-kickstart also returned a 1 the stdio isn't logged anywhere.

vpodzime:
Ah, okay, now it makes sense and looks good to me.

bcl:
Pushed.

PR #456, opened 2015-11-19
Fix the replacement suggestion for "hostname"

dashea:
It's not real helpful if the thing being checked for and the suggested replacement are the same thing.


vpodzime:
Fair enough. :)

dashea:
Pushed.

PR #457, opened 2015-11-20
Rewrite tests/gui/

atodorov:
This is a fairly big change so here are the highlights:

- execute anaconda directly from git dropping our dependency on building a LiveCD and running virtual machines (I've tried it in the past and was incredibly slow and hard to debug things)
- renamed everything that calls itself a LiveCDSomething
- modify existing tests to match the latest anaconda behavior and the new environment
- get rid of OutsideMixin class and make the tests simpler, closer to regular unittest.TestCase
- because of the above moved all the tests into a single directory

In the second commit there are two disabled lines. One is already reported bug; the other is some AT-SPI error, which I guess is probably due to API change or Dogtail3 not being officially supported yet. I will be looking into this later.

What's missing:
 - run_gui_tests.sh needs to check for a gnome-session and accessibility and possibly start them before running the tests
 - I didn't execute pylint b/c Rawhide has Python 3.5 and the existing packages need 3.4. They have to be rebuilt. 
- when anaconda hits a bug the test will kill it at the end. This sometimes leaves /mnt/sysimage mounted and a whole lot of device mapper devices dangling around. I saw an anaconda.storage.unmountFilesystems() method which seems to be what we need but haven't looked into it in more details


atodorov:
Ping, any comments here ?

vpodzime:
All the changes look good to me except for the commit messages -- there should be a short one-line summary and then a paragraph with more detailed explanation of what's changed.

atodorov:
@vpodzime - I've rebased to master and updated the commit messages as much as I can. If there's something in particular you don't like please point it out and propose a replacement text.

PR #458, opened 2015-11-23
Make sure python3.5 code can run in early initrd

bcl:
There are 2 problems with running python3.5 early in the initrd when
entropy is low and the nonblocking pool hasn't been initialized yet:
    
    1. cpython runs getrandom() syscall to see the object id hash. This
    blocks until the pool is initialized and isn't critical from a security
    standpoint. Override this with PYTHONHASHSEED=42
    
    2. Python3.5 uses getrandom() syscall for urandom(), this also blocks
    instead of falling back to reading /dev/urandom like it did in py3.4
    this is fixed by monkeypatching os.urandom to just read /dev/urandom
    in parse-kickstart -- important because we still want unique UUIDs for
    networking.

### DIFF
--- a/dracut/anaconda-lib.sh
+++ b/dracut/anaconda-lib.sh
@@ -215,7 +215,7 @@ set_neednet() {
 }
 
 parse_kickstart() {
-    /sbin/parse-kickstart $1 > /etc/cmdline.d/80-kickstart.conf
+    PYTHONHASHSEED=42 /sbin/parse-kickstart $1 > /etc/cmdline.d/80-kickstart.conf
     unset CMDLINE  # re-read the commandline
     . /tmp/ks.info # save the parsed kickstart
     [ -e "$parsed_kickstart" ] && cp $parsed_kickstart /run/install/ks.cfg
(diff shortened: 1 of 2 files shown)


dracut/anaconda-lib.sh
```diff
@@ -214,7 +214,16 @@ set_neednet() {
     [ -z "$netroot" ] && >> /tmp/net.ifaces
 }
 
+wait_for_entropy() {
+    while [ "$(cat /proc/sys/kernel/random/entropy_avail)" -lt "10" ]; do
+        ENTROPY_AVAIL=$(cat /proc/sys/kernel/random/entropy_avail)
+        warn "Waiting for entropy ($ENTROPY_AVAIL) to be >= 10"
+        sleep 1
+    done
```
vpodzime: This is quite unfortunate because AFAICT every single of those ``$(cat ...)``s, ``warn``s and ``sleep``s eat some entropy from the pool because kernel does process-memory-placement randomization. Does this really improve things? To me it seems like it must make the wait longer. Of course we at least report something, but I'm not sure that's really worth it.

bcl: Good point, and no it really didn't make things noticeably better. Found a better solution anyway.

bcl:
Pushed.

PR #459, opened 2015-11-24
Automatically generate sr@latin.po

dashea:
The modern Serbian language exhibits synchronic digraphia, which is an
interesting few minutes on Wikipedia and means that it can be written in
two different ways. The Serbian translators provide a translation in
Cyrillic, and from that we can generate the translation in Latin. Stop
pulling sr@latin.po from Zanata, since no one updates it there, and add
a new Makefile rule to generate one from sr.po using a filter program
that comes with gettext.

### DIFF
--- a/po/Rules-sr-latin
+++ b/po/Rules-sr-latin
@@ -1,0 +1,10 @@
+# Generate sr@latin.po
+
+# The Serbian language is written in two scripts: Cyrillic and Latin. The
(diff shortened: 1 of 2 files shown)


vpodzime:
Looks good to me too.

jkonecny12:
Seems good to me too.

dashea:
Pushed.

PR #460, opened 2015-11-25
Replace <list>.delete() with <list>.remove() in user.py

sujithshankar:
Python doesn't support 'list'.delete() function. This results in an exception during installation when the admin status of an user  is toggled in User creation page during installation. Using 'list'.remove() instead will fix this issue.


vpodzime:
Oops. :) Thanks for the patch! Unless somebody has some issues with it, I'm gonna push this tomorrow.

vpodzime:
Pushed.

PR #461, opened 2015-11-26
Fix PropertyNotFoundError PermHwAddress (#1269298)

jkonecny12:
Only wired and wireless devices have ``PermHwAddress`` everything else seems to have ``HwAddress``.
Add fallback solution when device don't have ``PermHwAddress`` with warning log. Should be removed in future if everything will be as expected.

*Resolves: rhbz#1269298*


vpodzime:
Looks good to me.

rvykydal:
Looks ok to me too.

jkonecny12:
Pushed

PR #462, opened 2015-11-30
Fix IPoIB interfaces are shown as Ethernet (#1116633)

jkonecny12:
When device is type infiniband there will be message:
"ib0 - Infiniband device ib0 - [mac]"

Instead of previous:
"ib0 - Ethernet device ib0 - [mac]"

*Resolves: rhbz#1116633*

### DIFF
--- a/isys/devices.c
+++ b/isys/devices.c
@@ -153,6 +153,7 @@ storagedone:
             char path[64];
             int fd, type;
             char buf[64];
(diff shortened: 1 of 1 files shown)


vpodzime:
Looks good to me.

rvykydal:
Looks good to me too.

jkonecny12:
Pushed

PR #463, opened 2015-11-30
Translation tests

dashea:
The goal of this is to create and use a common framework for testing translatable and translated strings. I went ahead and added the code for that at https://github.com/rhinstaller/translation-canary. This is applicable to a lot of our projects, and in fact the catalyst for this was a problem in blivet's translated strings, but I figured that I would start here.

The "translatable" tests are run after updating anaconda.pot. Any problems detected in the translatable strings raise an error before the strings are pushed to Zanata. Part of this was previously a test in gettext_warnings.sh.

The "translated" tests are run after create the dist tarball. Here I can make a stronger case for running the tests at a weird time instead of just as a regular make check kind of test: since Zanata is out of our control, the translated strings can change right up to the last minute, so to ensure that translated strings aren't busted in a way that will crash anaconda we need to run checks on the actual source distribution.

"translated" also includes a warning about languages with a translation level below a certain threshold, currently 10% of the strings. This one is a warning instead of an error since, I don't know, maybe it's ok to have a mostly untranslated language or two, and also the lack of anonymous Zanata pulls make it impossible for this one to do anything useful in Jenkins.

Possible discussion topics:

  - Is this all just a terrible idea or what?
  - Is using a submodule appropriate or at all a thing we want to do?
  - Is causing 'make release' or pot-update to fail too big a hammer? Keeping in mind that the errors caught by these tests crash anaconda. Failing at pot-update might be too big a hammer.
  - better ideas for paths and names and such
  - ???

The other projects we have that could use this:

  - blivet
  - pykickstart
  - blivet-gui
  - libbytesize
  - meh
  - initial-setup
  - system-config-kickstart

The tests already found errors in translations which would need to be fixed before creating another release:

  - es: `Your current <b>%(product)s</b> software selection requires...` has an extra `<a>` tag
  - fr: `Your current <a href="" title="%(tooltip)s"><b>%(product)s</b> software selection</a> requires...` has an unclosed `<a>` tag
  - hr: `Your current <b>%(product)s</b> software selection requires...` has a stray `</a>`
  - zh_TW: `Your current <b>%(product)s</b> software selection requires...` has a stray `</a>`

Languages proposed for the currently very conservative chopping block:

  - Turkish (tr): 9% translated
  - Farsi (fa): 4% translated
  - Latvian (lv): 4% translated
  - Asturian (ast): 3% translated
  - Bengali (bn): 3% translated
  - British English (en_GB): 3% translated
  - Sinhala (si): 3% translated
  - Albanian (sq): 2% translated
  - Greek (el): 2% translated
  - Maithili (mai): 2% translated
  - Afrikaans (af): 1% translated
  - Amharic (am): 1% translated
  - Belarusian (be): 1% translated
  - Bosnian (bs): 1% translated
  - Icelandic (is): 2% translated
  - Iloko (ilo): 1% translated
  - Macedonian (mk): 1% translated
  - Malay (ms): 1% translated
  - Nepali (ne): 1% translated
  - Northern Sotho (nso): 1% translated
  - Slovenian (sl): 1% translated
  - Swiss German (de_CH): 1% translated
  - Tajik (tg): 1% translated
  - Urdu (ur): 1% translated
  - Vietnamese (vi): 1% translated
  - Welsh (cy): 1% translated
  - Zulu (zu): 1% translated


dashea:
Just tried checking out this code on another computer, and wow, that was a lot of unexpected steps. So, I think submodules are a pretty horrible idea now. Does anyone have some better ideas?

dashea:
Trying again with a subtree, suggested by @M4rtinK, which seems a lot less annoying. Instead of a remote ref, translation-canary gets merged into this project in a way that can it can be managed as a separate entity.

bcl:
I'm all for this, as long as it gets run with our normal Jenkins tests so that I'm not surprised when I try to do a build.

dashea:
> I'm all for this, as long as it gets run with our normal Jenkins tests so that I'm not surprised when I try to do a build.

I would like that too, so we'll need to untangle the anonymous pull situation before using any of this.

As far as reducing surprises at release time, what would you think of a release mode that just disables the language? So when run from make ci the tests would fail, but when run as part of making a release the translatable tests will just be a warning, and the translated tests will edit LINGUAS to remove any languages that do not pass.

For things like blivet that don't use autotools we'd need to add and use a LINGUAS file but that shouldn't be too bad.

dashea:
Actually, never mind about LINGUAS. Instead I could just delete the .mo file and create a new source archive.

dashea:
Updated the PR. It's probably kind of confusing to look at by now.

I redid the merge commit for translation-canary to point to the current head of https://github.com/rhinstaller/translation-canary. That repo is the five commits from Nov 30 and Dec 1, plus the merge commit d5dc7b8363e8090be3ef5ae4802cce8a631b12b2. That's the same as before.

I made some changes to the translation-canary subtree that I plan to split out and add to the upstream repo. Those are:

- 8c4a6483f87e404e0232447b215dcf7815960b73, a pylint thing
- 81653396e1c85d1c124e56e04afa726263de5c03, make the modules look more like real modules
- 50da8d455d69f14936971c28376f3f57371099e9, test directories instead of tarballs and
- 4086b144db22a09beba7fac2e15d138d06f4b6c2, the release mode for translated strings tests I mentioned above

The anaconda-specific changes are the last four commits: use xgettext_werror to reject our strings during make anaconda.pot-update, run all of the checks during make check, use the release-mode test and a dist-hook target to remove bad translations from the source tarball, and another pylint thing.

dashea:
> and another pylint thing.

which depends on https://github.com/rhinstaller/pocketlint/pull/9

dashea:
One more, 02037b10119e17ae4f5205550a0872d15c552f4f, look for a case I found removing things from pocketlint

dashea:
Rebased and re-did the translation-canary subtree as a squashed commit.

PR #465, opened 2015-12-01
Change parameter at ks.cfg for fcoe as --nic= (#1225877)

jkonecny12:
We wrote output ks.cfg with ``fcoe --nic eth0`` but it's different from the documentation where we have ``fcoe --nic=eth0``. The output was modified in order to be consistent with the documentation.

I don't have fcoe so I couldn't test it. Could you please have a look on this @rvykydal  if this seems good to you?

*Resolves: rhbz#1225877*


rvykydal:
Looks ok to me.

jkonecny12:
Pushed

PR #467, opened 2015-12-04
zerombr now shows in anaconda-ks.cfg if used in ks

japokorn:
Fixed the bug where 'zerombr' was not written to anaconda-ks.cfg when used in
kickstart

### DIFF
--- a/storage/__init__.py
+++ b/storage/__init__.py
@@ -1242,6 +1242,11 @@ class Storage(object):
             return (foundCreateFormat and not foundCreateDevice)
 
         log.warning("Storage.writeKS not completely implemented")
+
+        if self.zeroMbr:
+            f.write("# Clear the Master Boot Record\n"
+                    "#zerombr\n")
+
         f.write("# The following is the partition information you requested\n")
         f.write("# Note that any partitions you deleted are not expressed\n")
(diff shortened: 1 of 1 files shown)


storage/__init__.py
```diff
@@ -1242,6 +1242,10 @@ def useExisting(lst):
             return (foundCreateFormat and not foundCreateDevice)
 
         log.warning("Storage.writeKS not completely implemented")
+
+        if self.zeroMbr:
+            f.write('# Clear the Master Boot Record\nzerombr\n')
```
jkonecny12: Please split this to two lines. I think it will be better to find on separate line.

japokorn:
Minor code update: String shown in diff is now splitted into two lines; double quotes used to match the rest of the code.

clumens:
You should make sure that the zerombr line is also commented out (put a # at the beginning of the line).  We do this with all the other storage lines in the various writeKS methods on rhel6-branch.  This is to force people to review the partitioning that we generated, and comes from a time when we were less confident in our storage code.

japokorn:
Code updated: added a '#' at the beginning of the 'zerombr line' to comply with the existing rhel6-branch policy.

japokorn:
cherry-picked and pushed

PR #468, opened 2015-12-09
Add possibility to have /usr and /var parts on iscsi (#856446)

jkonecny12:
Prevents adding of ``_netdev`` fstab parameter for mountpoints which are on the same LUN where ``/`` mountpoint is. 
This will fix ``/usr`` and ``/var`` are not mounted on boot and ``/boot`` is not mounted in singleuser mode.

We support ``/var`` and ``/usr`` on a LUN only when the ``/`` is on the same LUN.

*Resolves: rhbz#856446*

### DIFF
--- a/storage/__init__.py
+++ b/storage/__init__.py
@@ -2397,6 +2397,8 @@ class FSSet(object):
         devices += self.swapDevices
         devices.extend([self.devshm, self.devpts, self.sysfs, self.proc])
         netdevs = self.devicetree.getDevicesByInstance(NetworkStorageDevice)
+        rootdev = self.rootDevice
+
         for device in devices:
             # why the hell do we put swap in the fstab, anyway?
             if not device.format.mountable and device.format.type != "swap":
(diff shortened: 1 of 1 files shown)


storage/__init__.py
```diff
@@ -2424,7 +2426,11 @@ def fstab (self):
                 if device.dependsOn(netdev):
                     if mountpoint == "/":
                         options = options + ",_rnetdev"
-                    else:
+                    # Don't add _netdev to fstab if it is the same network device where root is.
+                    # Or they can wait for network forever, because the mountpoints for
```
vpodzime: *mountpoints required for...*

vpodzime:
Looks good to me otherwise.

storage/__init__.py
jkonecny12: Thank you for this correction. It's fixed now.

rvykydal:
Looks ok to me.
I think this should not break any other valid SAN use cases.

jkonecny12:
Pushed. Thank you guys

PR #469, opened 2015-12-09
Skip kickstart sections in loader (#1251300)

bcl:
The loader kickstart parser was assuming that any section indicated the
end of the command section. But a strict order has never been enforced,
so a section, like %pre, could come first. This changes it to ignore the
sections and continue parsing.

Resolves: rhbz#1251300


vpodzime:
Looks good to me.

jkonecny12:
Seems good to me too.

bcl:
Pushed

PR #471, opened 2015-12-11
new changes for new pylint

dashea:
This leaves some exception warnings during enchant.Broker.__del__ which would go away if https://github.com/rfk/pyenchant/pull/56 ever got merged, and gi.require_version warnings for BlockDev, NM and AnacondaWidgets, which I cannot figure out.


jkonecny12:
Looks good to me too.

PR #472, opened 2015-12-14
Add nospacecheck argument to bypass space checking

jkonecny12:
Adding inst.nospacecheck boot option and --nospacecheck argument.

User still gets about the space needed but the installation could be started.

This is some kind of follow-up on space check patches.

### DIFF
--- a/anaconda
+++ b/anaconda
@@ -870,6 +870,7 @@ if __name__ == "__main__":
     flags.selinux = opts.selinux
     flags.eject = opts.eject
     flags.kexec = opts.kexec
+    flags.nospacecheck = opts.nospacecheck
 
     # Switch to tty1 on exception in case something goes wrong during X start.
     # This way if, for example, metacity doesn't start, we switch back to a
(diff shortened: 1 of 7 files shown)


clumens:
Why do we need to do this?  We've already got about a million command line options, very few of which ever see any testing.  I think what's going to happen after this is you're going to get a bug from someone who used this option, didn't have enough space so anaconda failed, and is angry about anaconda providing them a way to shoot themselves in the foot.

jkonecny12:
Anaconda is using new space check now and it needs a little more space requirement then before. Still it's more precise and the installation shouldn't fail because of low space now.

Another reason is when user will use partitioning where most files will be installed to another partitions than ``/``. Even when this installation could be installed with minimum size of the ``/``  mountpoint we still require the whole required size in root ( maybe root + usr I don't know right now) and we don't have enough information to create precise estimation for this. Disabling space check seems to me as nice way to solve this.

jkonecny12:
Anyone here who wants this boot option in anaconda? I'm going to close this PR if I'm the only one.

vpodzime:
> Anyone here who wants this boot option in anaconda? I'm going to close this PR if I'm the only one.

Do we have a real feature request for it?

jkonecny12:
I can't find anything on bugzilla so I think we don't have request for now. 
So I'm going to keep this change for future and if someone shows up who would like this feature I'll create new PR for this patch.

For now closing this one.

PR #475, opened 2015-12-18
Fix multipath UI representation (#914637)

kellinm:
Changed the depopulate method of the FilterWindow class to correctly
identify and remove outdated representations of a multipath device in
the user interface. Added a check to ensure the first member device of a
multipath device is not accidentally displayed and counted as both an
mpath device member and on the tab for other san devices.

The issues impacting the display of multipath devices in the user
interface were also contributing to an improper count of selected and
total devices in the user interface. Added a data structure to track
tallied storage devices and account for the movement of multipath
devices from single other san devices over to the multipath tab.

Related: rhbz#1014425
Resolves: rhbz#914637

### DIFF
--- a/iw/filter_gui.py
+++ b/iw/filter_gui.py
@@ -46,6 +46,12 @@ _ = lambda x: gettext.ldgettext("anaconda", x)
 import logging
 log = logging.getLogger("anaconda")
 
+# Define constants that will be used within the FilterWindow class
+# for the variable device_data
+UDEV_DATA = 0
+VISIBLE_IN_UI = 1
+ACTIVE_IN_UI = 2
+DEVICE_MUTABILITY = 3
 DEVICE_COL = 4
 MODEL_COL = 5
 CAPACITY_COL = 6
@@ -127,7 +133,7 @@ class Callbacks(object):
         self.sizeLabel = self.xml.get_widget("sizeLabel")
         self.sizeLabel.connect("realize", self.update)
 
-    def addToUI(self, tuple):
+    def addToUI(self, device_data):
         pass
 
     def deviceToggled(self, set, device):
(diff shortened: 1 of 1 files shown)


iw/filter_gui.py
```diff
@@ -130,6 +130,24 @@ def __init__(self, xml):
     def addToUI(self, tuple):
         pass
 
+    def incrementDeviceCount(self, tuple):
```
jkonecny12: Could you please change the ``tuple`` name to something more self explanatory and also add brief docstring for this method.

iw/filter_gui.py
```diff
@@ -209,6 +227,31 @@ def addToUI(self, tuple):
         if not tuple[INTERCONNECT_COL] in self._interconnects:
             self._interconnects.append(tuple[INTERCONNECT_COL])
 
+    def incrementDeviceCount(self, tuple):
```
jkonecny12: Same as the above

jkonecny12:
Seems good to me otherwise. Thank you for adding the comments there. It's much readable right now.

iw/filter_gui.py
```diff
@@ -130,6 +130,24 @@ def __init__(self, xml):
...
+            return
+
+        talliedDevices.append(tuple[0]["name"])
+
+        totalDevices += 1
+        totalSize += tuple[0]["XXX_SIZE"]
+
+        if tuple[2]:
+            selectedDevices += 1
+            selectedSize += tuple[0]["XXX_SIZE"]
```
vpodzime: I think these methods could be made clearer if the tuples were expanded somewhere. ``tuple[X]`` is really cryptic.

vpodzime:
Looks good to me otherwise too.

kellinm:
@jkonecny12 and @vpodzime I have made changes. @dcantrell recommended I make the initial name change for tuple in a separate patch in order to keep the history a little clearer.

If this is good, let me know and I will close this PR and push to the branch.

vpodzime:
The new name for the variable is definitely an improvement. But what bugs me more is the use of the magic index ``0``. I saw some constants being used as indices in the tuple, could you please replace those ``0``s with a/the properly-named constant for that index? Otherwise this looks good to me,.

kellinm:
This should be the final set of updates. Documented constants for the device_data data structure and replaced all usage in the old file in the initial patch with the variable name refactor. Used them in my second patch for multipath. Added docstrings for new methods.

@jkonecny12 @vpodzime and @dwlehman - let me know if there are any other requests/objections.

vpodzime:
Looks good to me now.

vpodzime:
Thanks for the extra work!

jkonecny12:
Looks good to me too :+1:

kellinm:
Pushed manually, closing pull request.

PR #477, opened 2016-01-05
Fix regression in lvm snapshot detection. (#1295897)

dwlehman:
The regression was caused by the patch(es) adding thinp support.


snbueno:
Ack.

vpodzime:
Looks good to me too.

dwlehman:
Cherry-picked and pushed.

PR #479, opened 2016-01-07
Add specification for the user interaction config file

M4rtinK:

Anaconda starts up (eq. a language sleection tool in a live environment)
and to provide initial configuration file snippets to disable screen in Anaconda

### DIFF
--- a/docs/user-interaction-config-file-spec.rst
+++ b/docs/user-interaction-config-file-spec.rst
@@ -1,0 +1,171 @@
+Specification of the user interaction configuration file
+========================================================
+
+:Version:
+    1.0
+
+:Authors:
+    Martin Kolman <mkolman@redhat.com>
+
+This specification aims to establish a configuration file format, that can be used
+to communicate information about which installation screens have been seen by the user
+during the installation. Optionally the configuration file might also contain information
+about which configuration options have been changed by the user and if the user has
+explicitly requested post installation setup tools to be disabled.
+
+While this configuration file is primarily meant to be read (and potentially changed)
+by *post* installation tools (such as for example Initial Setup and Gnome Initial setup),
+Anaconda will take an existing configuration file into account at startup. This is meant
+to accommodate tools that are run *before* Anaconda is started, such as a system-wide
+language selection tool that runs before Anaconda is started and which sets language
+for both Anaconda and a Live installation environment.
+
+
+Configuration file location
+---------------------------
+
+The user interaction configuration file is stored in: ``/etc/sysconfig/anaconda``
+
+
+General configuration file systax
+---------------------------------
+
+The configuration file is based on the INI file de-facto standard,
+eq.: key=value assignments and square bracket framed section headers.
+
+Example:
+
+::
+
+    top_level_key=some_value
+
+    [section_1]
+    key_a_in_section1=some_value
+    key_b_in_section1=some_value
+
+    [section_2]
+    key_a_in_section2=some_value
+
+Boolean values are marked with 1 for true and 0 for false.
+
+Example:
+
+::
+
+    true_key=1
+    false_key=0
+
+
+Toplevel namespace
+------------------
+
+The toplevel configuration file namespace can only contain the
+``post_install_tools_disabled`` key and section headers.
+
+The ``post_install_tools_disabled`` key corresponds to using ``firstboot --disable`` command in the
+installation kickstart, that requests the post installation setup tools to be skipped. If this key is present
+and set to true, then any post installation tools that parses the user interaction file should make sure
+it won't be started again and then terminate immediately.
+
+Section headers each correspond to a spoke screen shown in Anaconda. Hubs are not represented as section
+as there is nothing that can be directly set on a hub in Anaconda.
+All currently visible spokes in Anaconda will be enumerated as a section headers by Anaconda in the
+configuration file. Note that this can include spokes provided by addons, if any Anaconda addons that provide
+additional spokes are present during the installation run.
+
+
+Section naming
+--------------
+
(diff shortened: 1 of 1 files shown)


docs/user-interaction-config-file-spec.rst
```diff
@@ -0,0 +1,171 @@
...
+Configuration file location
+---------------------------
+
+The user interaction configuration file is stored in: ``/etc/sysconfig/anaconda``
+
+
+General configuration file systax
+---------------------------------
+
+The configuration file is based on the INI file de-facto standard,
```
bcl: Maybe mention python's configparser module as the standard way to read/write the file.

docs/user-interaction-config-file-spec.rst
```diff
@@ -0,0 +1,171 @@
...
+    false_key=0
+
+
+Toplevel namespace
+------------------
+
+The toplevel configuration file namespace can only contain the
+``post_install_tools_disabled`` key and section headers.
+
+The ``post_install_tools_disabled`` key corresponds to using ``firstboot --disable`` command in the
```
bcl: I would change the use of true/false to 1/0 just so that it is extra clear that numbers are being used not strings.

docs/user-interaction-config-file-spec.rst
```diff
@@ -0,0 +1,171 @@
...
+or ``KeyboardSpoke``.
+
+
+Section namespace
+-----------------
+
+Each section *must* contain the ``visited`` key with a value of either true if the user has visited
+the corresponding screen or false if not.
+
+Optional each section can contain one or more keys with the ``changed_`` prefix which track if the user
```
bcl: 'Optionally'

docs/user-interaction-config-file-spec.rst
```diff
@@ -0,0 +1,171 @@
...
+All currently visible spokes in Anaconda will be enumerated as a section headers by Anaconda in the
+configuration file. Note that this can include spokes provided by addons, if any Anaconda addons that provide
+additional spokes are present during the installation run.
+
+
+Section naming
+--------------
+
+Section headers are named according to the Anaconda spoke class name. For example ``DatetimeSpoke``
+or ``KeyboardSpoke``.
```
bcl: It would be a good idea to mention where they can find the names for the classes. The easiest is for them to look at the 'Entered spoke:' entries in the /tmp/anaconda.log file during an installation, but a pointer to the spoke code on github would also be useful.

docs/user-interaction-config-file-spec.rst
```diff
@@ -0,0 +1,171 @@
...
+
+::
+
+    [KeyboardSpoke]
+    visited=0
+
+Here the user has not visited the keyboard spoke and thus could not have changed any options,
+so all ``changed_*`` keys (if any) have been omitted.
+
+
```
bcl: They should also assume that if a spoke isn't mentioned in the configuration file that it hasn't been visited.

docs/user-interaction-config-file-spec.rst
```diff
@@ -0,0 +1,171 @@
...
+This can be for example a tool querying the user for language settings. Then once Anaconda starts it can
+skip the language selection screen as language has already been set by the tool.
+
+Once the installation process is done Anaconda will write out information about what screens the user has
+and has not visited and optionally which settings have been changed by the user.
+
+If Anaconda successfully parsed an existing user interaction configuration file, any valid settings present
+in the file will propagate to the configuration file when it is written-out by Anaconda.
+
```
bcl: What about any comments in the file? We should probably assume no comments are preserved.

bcl:
Looks good! Thanks for doing this.

docs/user-interaction-config-file-spec.rst
M4rtinK: Sure, I can mention it, but can't really define it as the standard way of reading/writing of these files, as at least Gnome Initial Setup is not written in Python but in C. And there could be presumably other non-Python tools making use of this file format in the future.

docs/user-interaction-config-file-spec.rst
M4rtinK: Yeah, while it might look a bit weird in the text, it will definitely be clearer.

docs/user-interaction-config-file-spec.rst
M4rtinK: Thanks, fixed! :)

docs/user-interaction-config-file-spec.rst
M4rtinK: I'm thinking that this might not be enough - what about adding a (human readable) text file listing all the spokes in Anaconda, including a short description:

    DateTimeSpoke - date and time configuration
    CustomSpoke - custom partitioning configuration

etc.
It should be easy to create and very easy to maintain - we don't add or rename spokes too much these days.

bcl: I don't really like adding another spot where things have to be maintained, and the less often they are updated the higher the chances we'll forget when we do change them :) Unless we can generate it automatically from the source.

M4rtinK: I guess we could somehow generate that based on the ancestry (eq. sub-classing among others the NormalSpoke).

docs/user-interaction-config-file-spec.rst
M4rtinK: Added.

docs/user-interaction-config-file-spec.rst
M4rtinK: Good point, I'll add a note for both Anaconda & the non-Anaconda tools section. Basically: "try to keep comments if possible, but don't depend on them".

M4rtinK:
Obsoleted by #485.

PR #480, opened 2016-01-07
Handle unexpected DNF exit

bcl:
If DNF hits an error without raising BaseException we still need to send
a quit message. This keeps anaconda from hanging unexpectedly when there
is a problem installing a package.


dwlehman:
It might be nice to note somewhere that the transaction finishing will cause a 'post' to be enqueued before the new default 'quit', but this looks good to me.

vpodzime:
> It might be nice to note somewhere that the transaction finishing will cause a 'post' to be enqueued before the new default 'quit', but this looks good to me.

Yeah, otherwise it is quite hard to see the behavior of valid/right/desired case.

bcl:
Pushed with some additional commentary.

PR #481, opened 2016-01-12
Move requiredDeviceSize to the main Payload class (#1297905)

dashea:
dnf is not the only payload type.


jkonecny12:
Good point, thank you.

dashea:
Pushed.

PR #483, opened 2016-01-15
Enable multiselection in Custom spoke

jkonecny12:
This series of commits should add multiselection feature to Anaconda.

It supports multiselection when holding CTRL and SHIFT keys and clicking by mouse. I do not support item selection by keyboard but I think it don't work before too, so no difference here.

In these commits I have changed MountpointSelector.[c/h] so new AnacondaWidgets package is mandatory to work with this Custom spoke. 

I have a question should I increment AnacondaWidgets version or this is done by someone after some time? Also ``gi.require_version()`` for Widgets should be changed after the version increment.

Here you found screenshots:
https://jkonecny.fedorapeople.org/redhat/screenshots/multiselection/

*Resolved: 1265620*

### DIFF
--- a/pyanaconda/ui/gui/spokes/custom.py
+++ b/pyanaconda/ui/gui/spokes/custom.py
@@ -1923,6 +1923,8 @@ class CustomPartitioningSpoke(NormalSpoke, StorageChecker):
         if not self._current_selector:
             return
 
+        skip_dialog = False
+        is_multiselection = self._accordion.is_multiselection
         for selector in self._accordion.selectedPages:
             page = self._current_page
             device = selector.device
@@ -1951,18 +1953,31 @@ class CustomPartitioningSpoke(NormalSpoke, StorageChecker):
                 # This is a device that exists on disk and most likely has data
                 # on it.  Thus, we first need to confirm with the user and then
                 # schedule actions to delete the thing.
-                dialog = ConfirmDeleteDialog(self.data)
-                snapshots = (device.direct and not device.isleaf)
-                dialog.refresh(getattr(device.format, "mountpoint", ""),
-                               device.name, root_name, snapshots=snapshots)
-                with self.main_window.enlightbox(dialog.window):
-                    rc = dialog.run()
+                if not skip_dialog:
+                    dialog = ConfirmDeleteDialog(self.data)
+                    snapshots = (device.direct and not device.isleaf)
+                    if not is_multiselection:
+                        if root_name and "_" in root_name:
+                            root_name = root_name.replace("_", "__")
+                            checkbox_text = (C_("GUI|Custom Partitioning|Confirm Delete Dialog",
+                                                "Delete _all other file systems in the %s root as well.")
+                                                % root_name)
+                    else:
+                        checkbox_text = C_("GUI|Custom Partitioning|Confirm Delete Dialog",
+                                           "Do _not show this dialog for other selected file systems.")
+                    dialog.refresh(getattr(device.format, "mountpoint", ""),
+                                   device.name, checkbox_text=checkbox_text, snapshots=snapshots)
+                    with self.main_window.enlightbox(dialog.window):
+                        rc = dialog.run()
+                        skip_dialog = dialog.option_checked
 
-                    if rc != 1:
-                        dialog.window.destroy()
-                        return
+                        if rc != 1:
+                            dialog.window.destroy()
+                            if skip_dialog:
+                                return
+                            continue
 
-                if dialog.deleteAll:
+                if skip_dialog and not is_multiselection:
                     for dev in (s._device for s in page.members):
                         self._destroy_device(dev)
                 else:
@@ -2305,8 +2320,11 @@ class CustomPartitioningSpoke(NormalSpoke, StorageChecker):
                 selectedDeviceLabel.set_text(selector.device.name)
                 desc = _(MOUNTPOINT_DESCRIPTIONS.get(selector.device.type, ""))
                 selectedDeviceDescLabel.set_text(desc)
+
+            if self._accordion.is_multiselection:
+                self._current_selector = selector
+
             selector.set_chosen(True)
-            self._current_selector = selector
             self._configButton.set_sensitive(False)
             self._removeButton.set_sensitive(True)
             return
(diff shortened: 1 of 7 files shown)


jkonecny12:
I hope I have fixed everything. 
New commit was added which will ask on new item removal when there are more items selected. Behavior before this commit were to skip them and remove them automatically but to ask on every other selected item.

clumens:
I have two UI concerns.

(1) The sentence "You can't edit device in multiselection mode." uses jargon that I don't think we should expect the user to understand.  I think a better sentence might be something like "You can't edit a mount point when you have more than one selected."  That's more in line with other text we display on that screen and is clearer what the problem is.

(2) I'm not sure the "(1/3)" text showing how many devices are selected is at all clear.  I missed it the first time through the screenshots, and it was only when I saw the patch adding it that I noticed.  My first question would be:  Do we definitely need to display this information?  Did you find it to be confusing or misleading without knowing how many devices were selected in each?  If so, perhaps adding a small string underneath the name of the OS of the format "X of Y mount points selected" might be clearer.

jkonecny12:
> (1) The sentence "You can't edit device in multiselection mode."

Thank you for the correction. I'll change it.

> (2) I'm not sure the "(1/3)" text showing how many devices are selected is at all clear.

IMHO when you use the shift click selection then you may select closed Pages between the selected items. It will immediately show "3/3" now. I'm not sure if adding new line under the text would be the best solution. The Page will jump when something is selected or there will be "0 of 3 mount points selected" or there will be blank line? Somehow I don't like any of these variants.
I'm thinking that I will ask on opinion our graphics guy.

kellinm:
What about this instead?  (more direct linguistically, declarative)

"Select a single mount point to edit properties."

kellinm:
For the multiple select issue - what about having that data show up on the right side, it both helps clarify multiple selections and show detail when the accordion is closed:

"Select a single mount mount to edit properties.

clumens:
> What about this instead? (more direct linguistically, declarative)
>
> "Select a single mount point to edit properties."

Yeah, that would work too.  I'm not tied to any exact wording, as long as it's something along these lines.  I was going for something with the same tone as what we have when you've got no mount points selected - "When you create mount points for..."

clumens:
> For the multiple select issue - what about having that data show up on the right side, it both helps
> clarify multiple selections and show detail when the accordion is closed:
>
> Select a single mount mount to edit properties.
>
> You have currently selected:
> 1 of 3 mount points in New Fedora rawhide installation.
> 1 of 3 mount points in Fedora Linux 24
> 2 of 3 mount points in Fedora Linux 24
> 0 of ? mount points in Unknown

Yeah, this is a pretty good.  I think it's definitely at least worth mocking up and seeing how it looks.

jkonecny12:
OK I'll change it and post here new screenshots.

jkonecny12:
New screenshots. Take which version do you like more and I'll update the PR in that way.
https://jkonecny.fedorapeople.org/redhat/screenshots/multiselection/new/

(The white background is GTK error in rawhide)

kellinm:
It looks OK - but I'm not sure I'm sold 100%...

It's a nitpick - but I'm not so sure about the mount point count lines "moving" as more are added.

I know it's because it's centered in the pane, and I know that's probably better than having uneven top and bottom borders - but it just feels awkward if someone is clicking/adding and the height level of the label for the mount point counting lines changes.

Also - what happens there if someone has more lines than we have space for? Does it get a scrollbar?

M4rtinK:
Nice! Also I think I like the multi-without-empty variant more.

BTW, a related idea for a separate patch in the future - appending a counter to duplicate OS names:

Fedora Linux 24 for x86_64 (1.)
Fedora Linux 24 for x86_64 (2.)
Fedora Linux 24 for x86_64 (3.)

or

Unknown (1.)
Unknown (2.)
Unknown (3.)

That would once added also make the right pane status message more useful when the system contains duplicate OS installations.

jkonecny12:
I don't think there would or should be scrollbar but this needs more tests. I will look on that tomorrow and I will see. 
So or so this needs to be solved for both variants.

After our IRC conversation I think I'll go for variant:
with-empty and pages that are not selected - grey, pages that are not selectable - not present

jkonecny12:
Updated.

New multiselection design

jkonecny12:
@kellinm 
Finally I tested this on 100 pages and it was scroll-able without any problem so I think it's OK.

vpodzime:
Looks good to me.

jkonecny12:
Pushed. Thank you all

pyanaconda/ui/gui/spokes/custom.py
```diff
@@ -489,9 +467,42 @@ def get_new_devices(self):
 
         return new_devices
 
+    def _set_page_label_text(self):
+        if self._accordion.is_multiselection:
+            select_tmpl = _("{fmt_start}{selected}{fmt_end} of "
```
dashea: This is going to be a mess to translate. There are a lot pieces that will mean nothing to translators, which creates the potential for the translations to break it, and I'm not really sure what the purpose of all the fmt_start and fmt_end is anyway. The entire string is going to be made black or gray, and not just the pieces wrapped in span tags, right? I can see that fmt_start+fmt_end will also make it large and bold, but I don't understand what the color adds at that point. Either way, I think it would be more readable for both us and for translators if all of, for example, "{fmt_start}{selected}{fmt_end}" was a single parameter in the format string.

Additionally, it would be nicer if this used the old-style % formatting that we use elsewhere in anaconda. For one, I'm old and I hate new things, and for two, the {} style is not regular which makes some of the analysis we do in the tests a giant pain.

jkonecny12: > This is going to be a mess to translate. There are a lot pieces that will mean nothing to translators, which creates the potential for the translations to break it, and I'm not really sure what the purpose of all the fmt_start and fmt_end is anyway. The entire string is going to be made black or gray, and not just the pieces wrapped in span tags, right? I can see that fmt_start+fmt_end will also make it large and bold, but I don't understand what the color adds at that point. Either way, I think it would be more readable for both us and for translators if all of, for example, "{fmt_start}{selected}{fmt_end}" was a single parameter in the format string.

Fine no problem to change this.

> Additionally, it would be nicer if this used the old-style % formatting that we use elsewhere in anaconda. For one, I'm old and I hate new things, and for two, the {} style is not regular which makes some of the analysis we do in the tests a giant pain.

If this would be a problem in the tests I'll change it to the old style. However I think the translators hate this too because ``%(name)s`` seems as the ``s`` on the end is part of the text and not part of the variable. Personally that is why I changed it on the first place.

PR #484, opened 2016-01-20
Make langpack work in DNF (#1297823)

jsilhan:
Then we can get rid of these bug reports about the same thing:

* https://bugzilla.redhat.com/show_bug.cgi?id=1297823
* https://bugzilla.redhat.com/show_bug.cgi?id=1114422
* https://bugzilla.redhat.com/show_bug.cgi?id=1297520
* https://bugzilla.redhat.com/show_bug.cgi?id=1297058

see [how it woks with DNF after anaconda installation](https://bugzilla.redhat.com/show_bug.cgi?id=1114422#c22) + [how the packaging look like](https://fedoraproject.org/wiki/PackagingDrafts/Langpack).

### DIFF
--- a/pyanaconda/packaging/dnfpayload.py
+++ b/pyanaconda/packaging/dnfpayload.py
@@ -78,8 +78,6 @@ YUM_REPOS_DIR = "/etc/yum.repos.d/"
 # 6KiB = 4K(max default fragment size) + 2K(rpm db could be taken for a header file)
 BONUS_SIZE_ON_FILE = Size("6 KiB")
 
-_DNF_INSTALLER_LANGPACK_CONF = DNF_PLUGINCONF_DIR + "/langpacks.conf"
-_DNF_TARGET_LANGPACK_CONF = "/etc/dnf/plugins/langpacks.conf"
 
 def _failure_limbo():
     progressQ.send_quit(1)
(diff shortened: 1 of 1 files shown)


pyanaconda/packaging/dnfpayload.py
```diff
@@ -797,23 +795,10 @@ def preInstall(self, packages=None, groups=None):
...
-                f.write("[main]\n")
-
-        # langpacks.conf is an INI style config file, read it and
-        # add or change the enabled and langpack_locales entries without
-        # changing anything else.
-        keys=[("langpack_locales", "langpack_locales=" + ", ".join(langs)),
-              ("enabled", "enabled=1")]
-        simple_replace(_DNF_INSTALLER_LANGPACK_CONF, keys)
+        # add langpacks into transaction
+        for lang in [self.data.lang.lang] + self.data.lang.addsupport:
```
vpodzime: A little bit extra effort needs to be spent here. Both ``self.data.lang.lang`` and ``self.data.lang.addsupport`` contain things like ``en_US.UTF-8``, ``cs_CZ.UTF-8`` and so on. ``pyanaconda.localization.parse_langcode()`` needs to be used here to separate the langcode.

jsilhan:
we have a langpack definition package in [rawhide](http://koji.fedoraproject.org/koji/packageinfo?packageID=21858) already. I've fixed the patch so now it should pick the right langpack based on language.

pyanaconda/packaging/dnfpayload.py
```diff
@@ -797,23 +795,20 @@ def preInstall(self, packages=None, groups=None):
...
-        # langpacks.conf is an INI style config file, read it and
-        # add or change the enabled and langpack_locales entries without
-        # changing anything else.
-        keys=[("langpack_locales", "langpack_locales=" + ", ".join(langs)),
-              ("enabled", "enabled=1")]
-        simple_replace(_DNF_INSTALLER_LANGPACK_CONF, keys)
+        # get all available languages in repos
+        available_langpacks = self._base.sack.query().available() \
+            .filter(name__glob="langpacks-*")
+        alangs = map(lambda p: p.name.split('-', 1)[1], available_langpacks)
```
vpodzime: I don't think this would work. ``map()`` returns a ``map`` object that behaves like a generator. If you iterate over it twice, you'll get no results in the second run. You can either call ``list()`` on the ``map()`` result or (preferably) use a list comprehension.

vpodzime:
Otherwise this looks good to me.

jsilhan:
Yeah, I tested it in Python2. Now it should work in Python3 too.

pyanaconda/packaging/dnfpayload.py
```diff
@@ -797,23 +795,20 @@ def preInstall(self, packages=None, groups=None):
...
+            .filter(name__glob="langpacks-*")
+        alangs = [p.name.split('-')[1] for p in available_langpacks]
+
+        # add base langpacks into transaction
+        for lang in [self.data.lang.lang] + self.data.lang.addsupport:
+            loc = pyanaconda.localization.find_best_locale_match(lang, alangs)
+            if not loc:
+                log.warning(
+                    "Selected lang %s does not match any available langpack",
+                    lang)
```
vpodzime: The string doesn't have to be on a separate line. We are not too strict about this in anaconda and it would just make the code better readable.

vpodzime:
Apart from the final, minor, nitpick this looks good to me. If nobody complains, I'm gonna push this tomorrow.

jsilhan:
changed, logging line now exceeds 80 chars.

jkonecny12:
Looks good to me too.

vpodzime:
Pushed.

PR #485, opened 2016-01-25
Master post instal tool config spec V2

M4rtinK:

Anaconda starts up (eq. a language slelection tool in a live environment)
and to provide initial configuration file snippets to disable a screen in Anaconda

Changes from V1:
* added a script for listing Anaconda spokes
* added note about possibly loosing comments when the config file is parsed and written out
* added a note about configparser being the default Python INI parser
* added a link to the Anaconda addon development guide
* replaced true/false by 1/0 everywhere in the text of the spec
* clarified that screens that are not present in the config should be assumed as not-visited
* some typo fixes

(V1 pull request: #479)

### DIFF
--- a/docs/user-interaction-config-file-spec.rst
+++ b/docs/user-interaction-config-file-spec.rst
@@ -1,0 +1,204 @@
+Specification of the user interaction configuration file
+========================================================
+
+:Version:
+    1.0
+
+:Authors:
+    Martin Kolman <mkolman@redhat.com>
+
+This specification aims to establish a configuration file format, that can be used
+to communicate information about which installation screens have been seen by the user
+during the installation. Optionally the configuration file might also contain information
+about which configuration options have been changed by the user and if the user has
+explicitly requested post installation setup tools to be disabled.
+
+While this configuration file is primarily meant to be read (and potentially changed)
+by *post* installation tools (such as for example Initial Setup and Gnome Initial setup),
+Anaconda will take an existing configuration file into account at startup. This is meant
+to accommodate tools that are run *before* Anaconda is started, such as a system-wide
+language selection tool that runs before Anaconda is started and which sets language
+for both Anaconda and a Live installation environment.
+
+
+Configuration file location
+---------------------------
+
+The user interaction configuration file is stored in: ``/etc/sysconfig/anaconda``
+
+
+General configuration file systax
+---------------------------------
+
+The configuration file is based on the INI file de-facto standard,
+eq.: key=value assignments and square bracket framed section headers.
+
+For Python programs this file format can be parsed and written by the ConfigParser[0] module availabe
+from the standrad library. Comparable INI file parsing and writing modules are available for
+most other programming languages.
+
+Example:
+
+::
+
+    top_level_key=some_value
+
+    [section_1]
+    key_a_in_section1=some_value
+    key_b_in_section1=some_value
+
+    [section_2]
+    key_a_in_section2=some_value
+
+Boolean values are marked with 1 for true and 0 for false.
+
+Example:
+
+::
+
+    true_key=1
+    false_key=0
+
+
+Toplevel namespace
+------------------
+
+The toplevel configuration file namespace can only contain the
+``post_install_tools_disabled`` key and section headers.
+
+The ``post_install_tools_disabled`` key corresponds to using ``firstboot --disable`` command in the
+installation kickstart, that requests the post installation setup tools to be skipped. If this key is present
+and set to 1, then any post installation tools that parses the user interaction file should make sure
+it won't be started again and then terminate immediately.
+
+Section headers each correspond to a spoke screen shown in Anaconda. Hubs are not represented as section
+as there is nothing that can be directly set on a hub in Anaconda.
+All currently visible spokes in Anaconda will be enumerated as a section headers by Anaconda in the
+configuration file. Note that this can include spokes provided by addons[1], if any Anaconda addons that provide
+additional spokes are present during the installation run.
+
+
+Section naming
+--------------
+
+Section headers are named according to the Anaconda spoke class name. For example ``DatetimeSpoke``
+or ``KeyboardSpoke``.
+
+To get a list of all such spokes run the ``list_spokes`` script from the ``scripts`` directory in the Anaconda source
+code tree:
+
+::
+
+    git clone https://github.com/rhinstaller/anaconda
+    cd anaconda/scripts
+    ./list_screens
+
+Note that this script only lists Anaconda spokes, not spokes provided by addons[1] or Initial Setup.
+
+It is also possible to check for the *Entered spoke:* entries in the ``/tmp/anaconda.log`` file during an installation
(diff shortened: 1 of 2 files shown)


jcpunk:
Are there plans to make this state information available to anaconda (and plugins ) during the install process?

I'm thinking in particular of the timezone information.  If the user hasn't modified the timezone when the network comes up a geoip lookup could set the value based on location, but if the use has set the timezone it should not be changed automatically.

docs/user-interaction-config-file-spec.rst
```diff
@@ -0,0 +1,204 @@
...
+
+
+General configuration file systax
+---------------------------------
+
+The configuration file is based on the INI file de-facto standard,
+eq.: key=value assignments and square bracket framed section headers.
+
+For Python programs this file format can be parsed and written by the ConfigParser[0] module availabe
+from the standrad library. Comparable INI file parsing and writing modules are available for
```
mcatanzaro: standrad -> standard

M4rtinK:
>Are there plans to make this state information available to anaconda (and plugins ) during the install process?

I plan to implement this as a singleton similar to for example the thread manager - you would tell it the spoke name and it will tell you if the spoke was visited or not - and possibly other information, such as what was changed (if the given spoke tracks changes done by users).

>I'm thinking in particular of the timezone information. If the user hasn't modified the timezone when the network comes up a geoip lookup could set the value based on location, but if the use has set the timezone it should not be changed automatically.

In this case the problem still is that you already show the timezone to the user via the spoke status on the hub. So if a GeoIP lookup is done automatically once the Network comes up after it was manually configured by the user, you are still effectively changing things behind his back. Eq. the user saw "New York" in the Date and Time spoke status message but there might be "Prague" there once the network is setup, without any user interaction.

A possible solution might be to prompt the user somehow - possibly by making the Date and Time spoke unready after the lookup, so that the user has to confirm the newly detected timezone is OK.

docs/user-interaction-config-file-spec.rst
```diff
@@ -0,0 +1,204 @@
...
+most other programming languages.
+
+Example:
+
+::
+
+    top_level_key=some_value
```
mcatanzaro: If you add some toplevel group then it would be much easier for GNOME to read and write this file, as it would allow us to use our standard keyfile parser, [GKeyFile](https://developer.gnome.org/glib/stable/glib-Key-value-file-parser.html), which is for a format extremely similar to ini, but requires that all keys and values be in some group. It would probably speed adoption if we could use GKeyFile.

docs/user-interaction-config-file-spec.rst
```diff
@@ -0,0 +1,204 @@
...
+Note that if a spoke section is missing, it should be assumed that the corresponding screen has not been visited.
+On othe other hand, if a screen *has been visited*, the section *must* be present, with the ``visited`` key being equal to 1.
+
+
+Full configuration file example
+-------------------------------
+
+::
+
+    post_install_tools_disabled=0
```
mcatanzaro: For example, instead of this:

```
    post_install_tools_disabled=0

    [DatetimeSpoke]
    visited=1
    changed_timezone=1
    changed_ntp=0
    changed_timedate=1

    [KeyboardSpoke]
    visited=0
```

I would prefer this:

```
    [General]
    post_install_tools_disabled=0

    [DatetimeSpoke]
    visited=1
    changed_timezone=1
    changed_ntp=0
    changed_timedate=1

    [KeyboardSpoke]
    visited=0
```

Otherwise we would probably have to bundle an ini-file parser in gdm and gnome-initial-setup. (The logic that reads post_install_tools_disabled will probably need to go into gdm.)

docs/user-interaction-config-file-spec.rst
```diff
@@ -0,0 +1,204 @@
...
+::
+
+    [KeyboardSpoke]
+    visited=0
+
+Here the user has not visited the keyboard spoke and thus could not have changed any options,
+so all ``changed_*`` keys (if any) have been omitted.
+
+Note that if a spoke section is missing, it should be assumed that the corresponding screen has not been visited.
+On othe other hand, if a screen *has been visited*, the section *must* be present, with the ``visited`` key being equal to 1.
```
mcatanzaro: othe -> the

docs/user-interaction-config-file-spec.rst
```diff
@@ -0,0 +1,204 @@
...
+This can be for example a tool querying the user for language settings. Then once Anaconda starts it can
+skip the language selection screen as language has already been set by the tool.
+
+Once the installation process is done Anaconda will write out information about what screens the user has
+and has not visited and optionally which settings have been changed by the user.
+
+If Anaconda successfully parsed an existing user interaction configuration file, any valid settings present
+in the file will propagate to the configuration file when it is written-out by Anaconda.
+
+Note that comments present in the configuration file at the time Anaconda parses it might not be present
```
mcatanzaro: Are comments denoted by semicolons (as in Windows ini files) or hashes (as in freedesktop keyfiles)? Again, this affects whether we can use GKeyFile.

mcatanzaro:
Martin, I think this works great for GNOME. Thanks!

docs/user-interaction-config-file-spec.rst
mcatanzaro: Matthias points out this is really easy to workaround:

mclasen:  mcatanzaro: you can probably get around the ini-file limbo problem by prepending a group to the data from the file, but getting the format fixes is certainly better
mcatanzaro:  mclasen: I suppose that is a trivial workaround indeed, just add the group, then remove it when we're done with it

docs/user-interaction-config-file-spec.rst
M4rtinK: Fixed, thanks! :)

docs/user-interaction-config-file-spec.rst
M4rtinK: Fixed.

docs/user-interaction-config-file-spec.rst
M4rtinK: Yeah, that looks good - let's add a toplevel group to hold the post_install_tools_disabled key (and possibly other such top-level keys if needed in the future). 

The only possible issue with this could be if a screen called "General" was added somewhere in the installation process, so we need to assure that does not happen (or use a different name for the toplevel section ?).

docs/user-interaction-config-file-spec.rst
M4rtinK: The python configparser module takes both "# "and ";", so lets go with "#" so that GKeyFile can be used - an I guess I'll add GKeyFile to the list of suggested parser for the format when I am at it.

Also lets disallow inline comments - they don't seem to be supported very well across parsers, configparser does not parse them by default & advises against using them.

So this would be valid:

    # comment
    [Foo]
    # comment
    bar=baz
    # comment

But this not:

    [Foo]
    bar=baz # comment

M4rtinK:
Obsoleted by #501.

PR #486, opened 2016-01-27
Log detailed information about installed packages

bcl:
Log the full nevra of the package along with the buildtime and
sha256 checksum to packaging.log. Each line starts with 'Installed: '


vpodzime:
Definitely a nice idea. Just out of curiosity, why is the checksum useful?

jkonecny12:
Looks good to me too. Nice work.

bcl:
> Definitely a nice idea. Just out of curiosity, why is the checksum useful?

I'm guessing that it's to make sure it really is the build you think it is, even though nevra should be unique. releng uses it to track what was actually built into the live iso.

vpodzime:
Makes sense, thanks for the extra info!

bcl:
Pushed.

PR #487, opened 2016-01-28
Fix network tracebacks when running docker addon

bcl:
Changing network during the installation causes the network callbacks to be triggered.


vpodzime:
Looks good to me otherwise.

bcl:
> I think we can just call the method as all spokes must have it due to inheritance.

I was thinking there may be a way for an addon spoke to not have it, but on 2nd though they all need to inherit from Spoke or one of its children so it should be safe to drop the check.

bcl:
Pushed

PR #489, opened 2016-02-01
Increment AnacondaWidgets version

jkonecny12:
Follow-up for the https://github.com/rhinstaller/anaconda/pull/483 .

### DIFF
--- a/widgets/src/Makefile.am
+++ b/widgets/src/Makefile.am
@@ -61,7 +61,7 @@ lib_LTLIBRARIES = libAnacondaWidgets.la
 libAnacondaWidgets_la_CFLAGS = $(GTK_CFLAGS) $(GLADEUI_CFLAGS) $(LIBXKLAVIER_CFLAGS) -Wall -g\
 			       -DWIDGETS_DATADIR=$(WIDGETSDATA)
 libAnacondaWidgets_la_LIBADD = $(GTK_LIBS) $(GLADEUI_LIBS) $(LIBXKLAVIER_LIBS)
-libAnacondaWidgets_la_LDFLAGS = $(LTLIBINTL) -version-info 3:1:1
+libAnacondaWidgets_la_LDFLAGS = $(LTLIBINTL) -version-info 4:0:2
 libAnacondaWidgets_la_SOURCES = $(SOURCES) $(HDRS) \
 	  glade-adaptor.c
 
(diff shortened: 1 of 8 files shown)


widgets/src/Makefile.am
```diff
@@ -61,7 +61,7 @@ lib_LTLIBRARIES = libAnacondaWidgets.la
 libAnacondaWidgets_la_CFLAGS = $(GTK_CFLAGS) $(GLADEUI_CFLAGS) $(LIBXKLAVIER_CFLAGS) -Wall -g\
 			       -DWIDGETS_DATADIR=$(WIDGETSDATA)
 libAnacondaWidgets_la_LIBADD = $(GTK_LIBS) $(GLADEUI_LIBS) $(LIBXKLAVIER_LIBS)
-libAnacondaWidgets_la_LDFLAGS = $(LTLIBINTL) -version-info 3:1:1
+libAnacondaWidgets_la_LDFLAGS = $(LTLIBINTL) -version-info 3:2:1
```
vpodzime: According to some slides about Autotools I have I think this change should be to ``4:0:2``

library_version_changes

jkonecny12: Thank you for the explanation. I couldn't found it anywhere.

jkonecny12:
PR updated.
After discussion with @vpodzime I have fixed the notes above and I also incremented AnacondaWidgets to version 4.0 to be consistent with the soname version.

Therefore AnacondaWidgets soname version is 4.0 and GI for them is also 4.0.
Anaconda GI now requires AnacondaWidgets in version 4.0 everywhere in the code. 

I hope I understand it correctly now :D.

widgets/src/Makefile.am
dashea: https://www.gnu.org/software/libtool/manual/html_node/Updating-version-info.html is the libtool documentation for -version-info. 4:0:2 is correct, since the change adds interfaces but does not remove or change existing ones.

dashea:
The version in -version-info (4:0:2) is not necessarily connected to the widgets version (3.2/4.0). I think incrementing the widgets version to 3.3 instead of 4.0 would be better, since the changes are backwards compatible.

jkonecny12:
Yeah I was thinking about that too but that looks to me bad when there are so many different versions of one piece of the software. It's still somehow crazy to me :D .

I'll change the widgets version to 3.3 ( the AnacondaWidgets_3_3_gir_FILES too) and leave the soname version the same as it is (-version-info (4:0:2)).

jkonecny12:
Updated. @dashea is it fine now.

vpodzime:
Looks good to me.

dashea:
Looks good to me too

jkonecny12:
Pushed. Thank you for helping me.

PR #490, opened 2016-02-01
Stylesheet changes

dashea:
Update anaconda for the CSS changes in gtk 3.19/20. This does not include the anaconda widget changes that I was talking about because I'm still thinking about that.


vpodzime:
Looks good to me.

dashea:
Pushed.

PR #491, opened 2016-02-02
Master network revalidate sources

jkonecny12:
New version of Source spoke re-validation caused by change in the network settings, based on the patch https://github.com/rhinstaller/anaconda/pull/411 of @jcpunk. Thank you for this good patch.

The first 2 commits should be usable for the rhel7 branch later.
The last commit is removing my old solution of this problem. I like this more then the old implementation.

It's fixing [this](https://bugzilla.redhat.com/show_bug.cgi?id=1270354) Rhel 7 problem but for master.

### DIFF
--- a/pyanaconda/ui/gui/spokes/network.py
+++ b/pyanaconda/ui/gui/spokes/network.py
@@ -284,7 +284,6 @@ class NetworkControlBox(GObject.GObject):
         self._running_nmce = None
         self.spoke = spoke
         self.client = client
-        self.settings_changed = False
 
         # button for creating of virtual bond and vlan devices
         self.builder.get_object("add_toolbutton").set_sensitive(True)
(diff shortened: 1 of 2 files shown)


jkonecny12:
@rvykydal could you please look on this one if it's fine with you.

jkonecny12:
Updated. 
The old commits was squashed and the one new commit was introduced. 
The new commit set settings_changed variable only if the connection was changed. This test is based on the connection timestamp.

bcl:
What is a connection timestamp? And will chronyd time changes have an effect on this?

jkonecny12:
Connection timestamp is value from NetworkManager which is epoch (I think, don't remember exactly now). This epoch is telling us the last connection change time.
It could be changed with chrony but I don't think it's so big issue here because I don't test that as time but only on equality. 
So this could be an issue but only if the chrony change the time to exact second when the last connection change was done. If you see this as real issue there is no problem to change it to compare whole NetworkManager settings this is only some optimization because this is only two numbers and not whole dictionary.

bcl:
Ok, that sounds good.

M4rtinK:
Looks good, ack!

jkonecny12:
Pushed. Thank you.

PR #492, opened 2016-02-02
Handle PEP 3101 strings in the gettext context check

dashea:


jkonecny12:
Thank you for this patch.

dashea:
Pushed.

PR #494, opened 2016-02-02
Improve RHS summary strings in multiselection (#1265620)

jkonecny12:
The code is better for reading and especially for translation now.
The strings are using old ``%`` formatting now which is better for the tests.

*Related: rhbz#1265620*

It's follow-up on the PR https://github.com/rhinstaller/anaconda/pull/483. @dashea is it good now?

### DIFF
--- a/pyanaconda/ui/gui/spokes/custom.py
+++ b/pyanaconda/ui/gui/spokes/custom.py
@@ -19,6 +19,7 @@
 #
 # Red Hat Author(s): Chris Lumens <clumens@redhat.com>
 #                    David Lehman <dlehman@redhat.com>
+#                    Jiri Konecny <jkonecny@redhat.com>
 #
 
 # TODO:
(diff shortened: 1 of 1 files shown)


dashea:
It like the new select_tmpl (minus the context), but this commit does change how the strings are displayed. The product name is no longer displayed in bold, and on the "0 of x mount points" lines, only the numbers are shown in gray instead of the full line.

dashea:
One other note on the product name, and using markup in format strings in general: parameters inserted into markup strings should be escaped using pyanaconda.ui.gui.utils.escape_markup. There's normally a check for it but the usage here tricks the pocketlint test.

jkonecny12:
Ahh sorry I'm a fool. How I could missed that. 
I'll do the updated version and I'll remove the context.

It's nice to see the patch you created. I'll leave this commit in the old style but I'm going to use the new style in future.

jkonecny12:
Updated.
I hope it's good now.

dashea:
Looks good now, thanks

jkonecny12:
Pushed. Thank you.

PR #497, opened 2016-02-09
Log payloadError so we know why installation failed.

bcl:
Otherwise it just quits and all we have to go by is this in the logs:
Sufficient mountpoints found: {}

### DIFF
--- a/pyanaconda/packaging/dnfpayload.py
+++ b/pyanaconda/packaging/dnfpayload.py
@@ -722,6 +722,7 @@ class DNFPayload(packaging.PackagePayload):
             self._download_location = self._pick_download_location()
         except packaging.PayloadError as e:
             if errors.errorHandler.cb(e) == errors.ERROR_RAISE:
+                log.error("Installation failed: %r", e)
                 _failure_limbo()
 
         pkgs_to_download = self._base.transaction.install_set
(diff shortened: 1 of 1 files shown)


jkonecny12:
There should be Python-meh attached on DNF errors now but sure, it will be much better to have this in the logs.

pyanaconda/packaging/dnfpayload.py
```diff
@@ -734,6 +735,7 @@ def install(self):
             msg = 'Failed to download the following packages: %s' % str(e)
             exc = packaging.PayloadInstallError(msg)
             if errors.errorHandler.cb(exc) == errors.ERROR_RAISE:
+                log.error("Installation failed: %r", exc)
                 _failure_limbo()
```
vpodzime: Wouldn't it be better to make ``_failure_limbo()`` take an (optional) exception and log it?

bcl:
> There should be Python-meh attached on DNF errors now but sure, it will be much better to have this in the logs.

Remember that we aren't always running with a user at the console. This came up as part of the koji livemedia-creator run, it fails but there isn't a clear reason in the logs.

pyanaconda/packaging/dnfpayload.py
bcl: > Wouldn't it be better to make _failure_limbo() take an (optional) exception and log it?

I don't don't think that's really needed, and it's clearer to have the log.error right at the point of the exception.

bcl:
Pushed.

PR #498, opened 2016-02-09
Add a uninstall hook for the renamed anaconda

dashea:


jkonecny12:
Looks good to me

vpodzime:
Looking at this I feel like autotools must already have a solution for this. Installing a file with a different file name is not an uncommon thing.

dashea:
> Looking at this I feel like autotools must already have a solution for this.

Not really. You can apply a sed expression to every filename at install time, which is broader of a brush than is helpful for this (not renaming pyanaconda/anaconda.py, for example, would probably be tricky), or you can create a copy in the build directory with the new name.

dashea:
Pushed.

PR #499, opened 2016-02-09
Use host storage for directory or image install dnf download

bcl:
dnf downloads all the packages first, when running a --dirinstall or
--image install use the host's /var/tmp/ instead of /mnt/sysimage

### DIFF
--- a/pyanaconda/packaging/dnfpayload.py
+++ b/pyanaconda/packaging/dnfpayload.py
@@ -60,6 +60,7 @@ DNF_PLUGINCONF_DIR = '/tmp/dnf.pluginconf'
 DNF_PACKAGE_CACHE_DIR_SUFFIX = 'dnf.package.cache'
 DOWNLOAD_MPOINTS = {'/tmp',
                     '/',
+                    '/var/tmp',
                     '/mnt/sysimage',
                     '/mnt/sysimage/home',
                     '/mnt/sysimage/tmp',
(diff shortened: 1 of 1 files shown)


bcl:
I'm not 100% sure about this patch. The problem I have is that it creates a situation where a kickstart could work with a --dirinstall but not with a regular installation. Maybe that doesn't really matter?

pyanaconda/packaging/dnfpayload.py
```diff
@@ -98,6 +99,11 @@ def _df_map():
         if not key.startswith('/'):
             continue
         structured[key] = Size(int(val)*1024)
+
+    # Add /var/tmp/ if this is a directory or image installation
+    if flags.dirInstall or flags.imageInstall:
+        var_tmp = os.statvfs("/var/tmp")
+        structured["/var/tmp"] = Size(var_tmp.f_frsize * var_tmp.f_bfree)
```
vpodzime: Is there (somewhere further in the code) some check testing that the directory is writable?

vpodzime:
If the answer for the above question is yes, this looks good to me.

bcl:
> If the answer for the above question is yes, this looks good to me.

No, it gets passed to dnf which handles creating its cache directory and writing the rpms. /var/tmp should never be read only though, it's a local disk backed tmp directory.

vpodzime:
> If the answer for the above question is yes, this looks good to me.
> 
> No, it gets passed to dnf which handles creating its cache directory and writing the rpms. /var/tmp should never be read only though, it's a local disk backed tmp directory.

I was thinking about some possibility to run anaconda in some restricted environment with everything outside of ``/mnt/sysimage`` (or the ``--dirinstall`` target) being read-only. But that's probably not gonna happen, so I'm okay with this.

bcl:
Thanks, pushed.

PR #501, opened 2016-02-10
Master post install config spec v3

M4rtinK:

Anaconda starts up (eq. a language selection tool in a live environment)
and to provide initial configuration file snippets to disable a screen in Anaconda

Changes from V2:
* moved the top level key to a group called *General*
* specified how comments should look like
* added GKeyFile as suggested parser for C applications parsing the file format

(V2 pull request: #485)

### DIFF
--- a/docs/user-interaction-config-file-spec.rst
+++ b/docs/user-interaction-config-file-spec.rst
@@ -1,0 +1,229 @@
+Specification of the user interaction configuration file
+========================================================
+
+:Version:
+    1.0
+
+:Authors:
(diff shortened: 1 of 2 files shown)


jkonecny12:
Seems good to me but someone with better English should look on this too.

clumens:
Okay, I claim to speak English and I've given it a review for grammar, spelling, and readability.  Thanks!

M4rtinK:
Thanks for the review everyone!

I've pushed this manually & I'm closing the PR now. :)

PR #504, opened 2016-02-10
More updates to tests/gui/

atodorov:
Update the code for the anaconda move to anaconda.py and also update run_tests_gui.sh in preparation to add it to make check and also to collect coverage reports.

CC @clumens


clumens:
Applied and pushed, thanks.

PR #505, opened 2016-02-11
Warn user before using Delete all

jkonecny12:
Warn user that "Delete all" checkbox removes even shared partitions.
Thank you for this patch David Lehman.

Applied to the new Rawhide code.

Related: rhbz#1183880
(based on commit b0f635a91e48cdc6e7d081dec6951fd3cd22b6c8)

*Based on the commit from f23*


jkonecny12:
Instead of this PR I'm going to use https://github.com/rhinstaller/anaconda/pull/508 .

PR #506, opened 2016-02-12
Move the langpacks install into to a separate function.

dashea:
preInstall is too early to add packages to the actual dnf transaction,
since the transaction is going to be reset in checkSoftwareSelection.
Instead, add langpacks packages in _apply_selections at the same time
the kernel package is selected.


vpodzime:
Looks good to me too.

dashea:
Pushed.

PR #507, opened 2016-02-12
Fix _find_by_title method in Accordion

jkonecny12:
It returned expander instead of page which I didn't know also this private method was called in the custom.py.

Fixed usage and renamed from _find_by_title to find_page_by_title. Now it really returns page. Please use .get_parent() method to get expander instead.


jkonecny12:
Pushed.

PR #508, opened 2016-02-12
UEFI final solution

jkonecny12:
This is almost the same as PR https://github.com/rhinstaller/anaconda/pull/383 from @AdamWill changed to be applicable on the actual custom spoke. With one small change in the second commit because pageTitle aren't unique.

This should be final solution for the UEFI problem. I hope :).

Thank you for the code @AdamWill. Good job.


vpodzime:
Looks good to me. The small change should be merged into the main commit.

jkonecny12:
I left the second commit as separate commit intentionally because the main commit is almost copy-paste (but on other place) of the original commit. But when I'm looking on it now it should be squashed.
I'll do that before push.

jkonecny12:
I have found some weird behavior so I'm closing this PR and I will create new one after I fix it.

PR #511, opened 2016-02-17
Wait for all udev events to finish (#1309348)

jstodola:
In some cases, a DASD device is not brought online by linuxrc.
Make sure that all udev events are finished before trying to
bring them online.

Resolves: rhbz#1309348


snbueno:
This looks ok to me.

vpodzime:
Looks good to me too as long as we are okay with the default timeout (120 s). Otherwise, please specify some with the ``--timeout`` option.

jstodola:
Yes, the default timeout is OK in this case.

snbueno:
Pushed.

PR #512, opened 2016-02-17
Master uefi fix2

jkonecny12:
This is the PR https://github.com/rhinstaller/anaconda/pull/508 but rebased to actual master (there were some conflicts) and with new commit.

When user click on any not-shared partition and check option Delete all it will remove all not-shared partitions from that page. 
If an user did this on shared partition it will remove all not-shared partitions in that page but it will also remove that selected partition (even when it's shared).

To stay consistent use the same rule for the selected partition.

I hope this will be good solution for all of you guys.

*Related: rhbz#1183880*


vpodzime:
I think the proposed behavior should be okay for users. And the changes look good to me apart from the minor comments above.

jkonecny12:
Updated. Thank you for review @vpodzime .

jkonecny12:
Pushed

PR #513, opened 2016-02-18
updated driver updates docs

wgwoods:
The old docs were kind of scant and out-of-date, so here's some brand
new maintainer documentation for driver updates disks!


vpodzime:
Wow, thanks!

PR #514, opened 2016-02-18
Manually set clock shifts on UI idle (#1251044)

kellinm:
Users reported the clock shifting a few seconds after manually setting
the clock when NTP was not available during installation.

The code was calculating the date to set the system clock by comparing
the GUI text fields to the epoch time. User input was not properly
converted thus the call to set system time was receiving the wrong
values. The GUI would begin updating from the system clock after sitting
idle a few seconds and use this incorrect time.

Resolves: rhbz#1251044
Resolves: rhbz#1293314

### DIFF
--- a/pyanaconda/isys/__init__.py
+++ b/pyanaconda/isys/__init__.py
@@ -135,10 +135,10 @@ def set_system_date_time(year=None, month=None, day=None, hour=None, minute=None
     minute = minute if minute is not None else now.minute
     second = second if second is not None else now.second
 
-    set_date = datetime.datetime(year, month, day, hour, minute, second, tzinfo=tz)
+    set_date = tz.localize(datetime.datetime(year, month, day, hour, minute, second))
 
     # Calculate the number of seconds between this time and timestamp 0
-    epoch = datetime.datetime.fromtimestamp(0, pytz.UTC)
+    epoch = tz.localize(datetime.datetime.fromtimestamp(0))
     timestamp = (set_date - epoch).total_seconds()
 
     set_system_time(timestamp)


kellinm:
If this is OK I'll close the PR and push to rhel7-branch.

jkonecny12:
Maybe I'm dump but this seems almost the same to me. What is the difference? Is it the ``pytz.UTC`` ?

Btw you should look if this is not happening in the master too and use the fix there.

kellinm:
Updated the subject title.

It does happen in master and it's documented behavior as part of pytz here: https://pypi.python.org/pypi/pytz

An excerpt:

> This library differs from the documented Python API for
> tzinfo implementations; if you want to create local wallclock
> times you need to use the ``localize()`` method documented in this
> document. In addition, if you perform date arithmetic on local
> times that cross DST boundaries, the result may be in an incorrect
> timezone (ie. subtract 1 minute from 2002-10-27 1:00 EST and you get
> 2002-10-27 0:59 EST instead of the correct 2002-10-27 1:59 EDT). A
> ``normalize()`` method is provided to correct this. Unfortunately these
> issues cannot be resolved without modifying the Python datetime
> implementation (see PEP-431).

Here's a code snippet that demonstrates the issue:

[kellin@hachiroku ~]$ cat datetest.py 
#!/usr/bin/python
import datetime
import pytz

epoch = datetime.datetime.fromtimestamp(1, pytz.timezone('America/Moncton'))
set_date = datetime.datetime(2016, 2, 18, 17, 4, 0, tzinfo=pytz.timezone('America/Moncton'))
print(epoch)
print(set_date)


my_timezone = pytz.timezone('America/Moncton')
my_date = my_timezone.localize(datetime.datetime(2016, 2, 18, 17, 4, 0))
print(my_timezone)
print(my_date)
[kellin@hachiroku ~]$ ./datetest.py 
1969-12-31 20:00:01-04:00
2016-02-18 17:04:00-04:19
America/Moncton
2016-02-18 17:04:00-04:00
[kellin@hachiroku ~]$

jkonecny12:
Ahh fine, this is fail.
Thank you for this fix and explanation. Good job.

kellinm:
New commit message to hopefully clarify for other folks who may not see this conversation in the future.

> Users reported the clock shifting a few seconds after manually setting
> the clock when NTP was not available during installation.
> 
> The code calculates the new system clock date by comparing the user
> defined time fields in the GUI to epoch time. This works if all
> calculations are performed in UTC, however, any conversions to and from
> UTC or calculations between different timezones currently return
> incorrect values due to a known issue in the Python datetime library.
> 
> Fixed the issue by using methods provided in the pytz library
> specifically designed to ameliorate these known issues when calculating
> the difference between the user selected time and epoch time.
> 
> Resolves: rhbz#1251044
> Resolves: rhbz#1293314

vpodzime:
Looks good to me too.

jkonecny12:
The commit message seems to me much better now. Thank you.

kellinm:
Closing / pushing manually.

PR #515, opened 2016-02-19
More descriptive message on invalid username

KosiehBarter:
Also checks against directory names in root (/) dir and
Checks against invalid chars as well as digits on start

Resolves: rhbz#1259284

### DIFF
--- a/pyanaconda/ui/tui/spokes/__init__.py
+++ b/pyanaconda/ui/tui/spokes/__init__.py
@@ -21,7 +21,7 @@
 from pyanaconda.ui.tui import simpleline as tui
 from pyanaconda.ui.tui.tuiobject import TUIObject, YesNoDialog
 from pyanaconda.ui.common import Spoke, StandaloneSpoke, NormalSpoke
-from pyanaconda.users import validatePassword, cryptPassword
+from pyanaconda.users import validatePassword, cryptPassword, check_name
 import re
 from collections import namedtuple
 from pyanaconda.iutil import setdeepattr, getdeepattr
@@ -103,7 +103,6 @@ class EditTUIDialog(NormalTUISpoke):
     To override the wrong input message set the wrong_input_message attribute
     to a translated string.
     """
-
     title = N_("New value")
     PASSWORD = re.compile(".*")
 
@@ -184,16 +183,20 @@ class EditTUIDialog(NormalTUISpoke):
             return _("Enter new value for '%s' and press enter\n") % entry.title
 
     def input(self, entry, key):
-        if entry.aux.match(key):
+        valid, err_msg = check_name(key)
+        if valid and entry.aux.match(key):
             self.value = key
             self.close()
             return True
         else:
+            self.wrong_input_message = err_msg
             if self.wrong_input_message:
                 print(self.wrong_input_message)
             else:
-                print(_("You have provided an invalid value\n"))
+                print(_("You have provided an invalid user name: %s\n"
+                        "Tip: Keep your user name shorter than 32 characters and do not use spaces.\n") % key)
             return NormalTUISpoke.input(self, entry, key)
+
 
 class OneShotEditTUIDialog(EditTUIDialog):
     """The same as EditTUIDialog, but closes automatically after
--- a/pyanaconda/users.py
+++ b/pyanaconda/users.py
@@ -28,6 +28,7 @@ import pwquality
 from pyanaconda.iutil import strip_accents
 from pyanaconda.constants import PASSWORD_MIN_LEN
 from pyanaconda.errors import errorHandler, PasswordCryptError, ERROR_RAISE
+import re
 
 import logging
 log = logging.getLogger("anaconda")
@@ -165,6 +166,22 @@ def validatePassword(pw, user="root", settings=None, minlen=None):
             message = e.args[1]
 
     return (valid, strength, message)
+
+def check_name(in_name):
+    in_name = in_name.replace(" ", "")
+    ign_cond = {r"^[\d]": "User name cannot start with a digit: ",
+                r"[^-_A-Za-z0-9]": "User name cannot contain any non-alphanumerical character: "}
+
+    for sys_name in os.listdir("/") + ["root", "home", "daemon", "system"]:
+        if in_name in sys_name:
+            return (False, "Username is reserved for system: %s" % sys_name)
+
+    for cond in ign_cond:
+        match = re.search(cond, in_name)
+        if match:
+            return (False, ign_cond[cond] + "'%s'" % match.group())
+
+    return (True, None)
 
 def guess_username(fullname):
     fullname = fullname.split()


pyanaconda/ui/tui/spokes/__init__.py
```diff
@@ -166,7 +166,11 @@ def prompt(self, entry = None):
             return _("Enter new value for '%s' and press enter\n") % entry.title
 
     def input(self, entry, key):
-        if entry.aux.match(key):
+        not_allowed_dirs = ['etc', 'boot', 'opt', 'proc', 'lib64', 'sbin', 'mnt',
+                            'lib', 'var', 'lost+found', 'sys', 'usr',
+                            '.esmtp_queue', 'dev', 'tmp', 'home', 'srv', 'root',
+                             'run', 'bin', 'media', 'daemon']
```
vpodzime: This list should definitely live somewhere else so that it can be used from other places in code as well. For example, how are we checking this in GUI?

vpodzime: Also it would be nice to add a comment about how this list was created.

pyanaconda/ui/tui/spokes/__init__.py
```diff
@@ -166,7 +166,11 @@ def prompt(self, entry = None):
...
+        if entry.aux.match(key) and key not in  not_allowed_dirs and not re.search("^[\d]", key):
```
vpodzime: Double space after ``in``.

pyanaconda/ui/tui/spokes/__init__.py
vpodzime: And it should be a set because the order of items doesn't matter and the ``in`` operation on a set is faster.

pyanaconda/ui/tui/spokes/__init__.py
```diff
@@ -175,9 +179,20 @@ def input(self, entry, key):
...
-                        "Tip: Keep your user name shorter than 32 characters and do not use spaces.\n") % key)
+                        "%s\n"
+                        "Tip: Keep your user name shorter than 32 characters and do not use spaces.\n")) % (key, self._error_mes(key, not_allowed_dirs))
             return NormalTUISpoke.input(self, entry, key)
 
+    def _error_mes(self, in_user, not_allowed_dirs):
+        if in_user in not_allowed_dirs:
+            return "This name is reserved for system"
+        elif re.search("^[\d]", in_user):
+            return "Username cannot start with digit"
```
vpodzime: *with a digit*

pyanaconda/ui/tui/spokes/__init__.py
```diff
@@ -175,9 +179,20 @@ def input(self, entry, key):
...
+        elif re.search("[\D]", in_user):
+            return "Invalid character in username: %s" % re.search("[\D]", in_user).group()
```
vpodzime: Are you sure about this check? AFAICT it means "if there's something else than digits in ``in_user``".

jkonecny12:
Please also change the "Resolves: RHBZ#1259284" line to "Resolves: rhbz#1259284" and move it below text and above the Sign-off... 
Something like:

...text...

Resolves....

Sign-off...

pyanaconda/users.py
```diff
@@ -168,6 +169,25 @@ def validatePassword(pw, user="root", settings=None):
 
     return (valid, strength, message)
 
+def check_name(in_name, ret_bool = False):
+    ign_dir = {}
+    for inc in os.listdir("/") + ["root", "home", "daemon"]:
+        ign_dir.update({inc: "Cannot set user name, reserved for system: "})
+    ign_dir.update({r"^[\d]": "User name cannot start with a digit: "})
+    ign_dir.update({r"[^A-Za-z0-9\s]": "User name cannot contain any non-ASCII character: "})
```
vpodzime: The regexp doesn't match the error. Either it needs to be changed/replaced or the error message needs to be more accurate.

pyanaconda/users.py
```diff
@@ -168,6 +169,25 @@ def validatePassword(pw, user="root", settings=None):
...
+
+    for inc in ign_dir:
+        tmp = regsearch(inc, in_name)
```
vpodzime: ``tmp`` is not a good name for a variable. What about ``match``?

pyanaconda/users.py
```diff
@@ -168,6 +169,25 @@ def validatePassword(pw, user="root", settings=None):
...
+        if tmp:
+            if ret_bool:
+                return False
+            else:
+                return ign_dir[inc] + "'%s'" % tmp.group()
+    if tmp == None:
```
vpodzime: use ``is None`` instead of ``== None``

pyanaconda/users.py
```diff
@@ -168,6 +169,25 @@ def validatePassword(pw, user="root", settings=None):
...
+        return "Unknown issue with username"
+    if ret_bool:
+        return False
+
```
vpodzime: In general I think the ``ret_bool`` parameter should be dropped and the function should just return the error message (or ``None`` if there's none) or, if the boolean value is really needed, it should return tuples like ``(True, None)`` or ``(False, error_msg)``.

pyanaconda/users.py
KosiehBarter: `match`

- can collide with function, if it is imported
- this is a temporary, I don't think we need to store it. This variable is changed often.

vpodzime: * That's why we don't import things like ``re.match`` directly to our namespace, but import just the module.
* That's not a reason for not giving the variable a proper name that describes what it holds.

pyanaconda/ui/tui/spokes/__init__.py
```diff
@@ -166,11 +166,12 @@ def prompt(self, entry = None):
             return _("Enter new value for '%s' and press enter\n") % entry.title
 
     def input(self, entry, key):
-        if entry.aux.match(key):
+        if entry.aux.match(key) and check_name(key)[0]:
             self.value = key
             self.close()
             return True
         else:
+            self.wrong_input_message = check_name(key)[1]
```
vpodzime: No need to call ``check_name()`` twice here. Just store the result first and then use its contents like this:
```
(valid, err_msg) = check_name(key)
if entry.aux.match(key) and valid:
    ...
```

pyanaconda/users.py
M4rtinK: Also what about translations ? The strings would show up in the UI so they should be translated.

jkonecny12:
Do not add new and new commits. That is not good way to do this. You need to force push the fixed commits or to create new pull request.

Also you can't push to rhel branch anything which doesn't have ``Resolves`` or ``Related`` tag in a commit message. The system will block you without this tag.

pyanaconda/ui/tui/spokes/__init__.py
```diff
@@ -97,11 +97,7 @@ class NormalTUISpoke(TUISpoke, NormalSpoke):
 # Inherit abstract methods from NormalTUISpoke
 # pylint: disable=abstract-method
 class EditTUIDialog(NormalTUISpoke):
-    """Spoke/dialog used to read new value of textual or password data
-
-    To override the wrong input message set the wrong_input_message attribute
-    to a translated string.
-    """
+    """Spoke/dialog used to read new value of textual or password data"""
```
vpodzime: Why is the above paragraph removed from the docstring?

pyanaconda/users.py
```diff
@@ -27,8 +27,11 @@
 import pwquality
 from pyanaconda.iutil import strip_accents
 from pyanaconda.constants import PASSWORD_MIN_LEN
+
 from pyanaconda.errors import errorHandler, PasswordCryptError, ERROR_RAISE
```
vpodzime: Why is the blank line added here?

pyanaconda/users.py
```diff
@@ -27,8 +27,11 @@
...
 
+from re import search as regsearch
```
vpodzime: ``import re`` and then using ``re.search`` would be better readable. This way when I hit the ``regsearch`` identifier in the code it looks to me like some custom thing.

pyanaconda/users.py
```diff
@@ -161,6 +164,20 @@ def validatePassword(pw, user="root", settings=None):
 
     return (valid, strength, message)
 
+def check_name(in_name):
+    ign_dir = {}
+    for inc in os.listdir("/") + ["root", "home", "daemon"]:
+        ign_dir.update({inc: "Cannot set user name, reserved for system: "})
```
vpodzime: If adding a key-value pair to the dict, use the ``some_dict[key] = value`` syntax not the ``update`` method.

pyanaconda/users.py
```diff
@@ -161,6 +164,20 @@ def validatePassword(pw, user="root", settings=None):
...
+    ign_dir.update({r"^[\d]": "User name cannot start with a digit: "})
+    ign_dir.update({r"[^A-Za-z0-9\s]": "User name cannot contain any non-alphanumerical character: "})
```
vpodzime: These two key-value pairs should be made part of the dictionary when it is created.

pyanaconda/ui/tui/spokes/__init__.py
```diff
@@ -170,15 +166,14 @@ def prompt(self, entry = None):
             return _("Enter new value for '%s' and press enter\n") % entry.title
 
     def input(self, entry, key):
-        if entry.aux.match(key):
-            self.value = key
-            self.close()
-            return True
-        else:
+        (valid, err_msg) = check_name(key)
+        if entry.aux.match(key) and valid:
```
vpodzime: Put the boolean value in the first place, it's cheaper to check it than do the ``entry.aux.match(key)`` call. And do we want to call ``check_name()`` before ``entry.aux.match()`` is called? To me it looks like we should first check if the intput is valid then check if it is a valid user name. Or do just one of the checks.

pyanaconda/users.py
```diff
@@ -161,6 +164,20 @@ def validatePassword(pw, user="root", settings=None):
...
+
+    for inc in ign_dir:
+        match = regsearch(inc, in_name)
+        if match:
+            return (False, ign_dir[inc] + "'%s'" % match.group())
```
vpodzime: What is the point of returning the match here when it is later just put into this: "You have provided an invalid user name: %s\n"? E.g. for ``"123my_username"`` this would put ``"1"`` there with this result shown to the user:
```
You have provided an invalid user name: 1
```

pyanaconda/users.py
```diff
@@ -161,6 +164,20 @@ def validatePassword(pw, user="root", settings=None):
 
     return (valid, strength, message)
 
+def check_name(in_name):
+    ign_dir = {}
+    for inc in os.listdir("/") + ["root", "home", "daemon"]:
+        ign_dir.update({inc: "Cannot set user name, reserved for system: "})
+    ign_dir.update({r"^[\d]": "User name cannot start with a digit: "})
```
vpodzime: The square brackets in the regular expression can be dropper here.

pyanaconda/ui/tui/spokes/__init__.py
```diff
@@ -103,7 +103,7 @@ class EditTUIDialog(NormalTUISpoke):
     To override the wrong input message set the wrong_input_message attribute
     to a translated string.
     """
-
+
```
vpodzime: Trailing white space added.

pyanaconda/ui/tui/spokes/__init__.py
```diff
@@ -184,17 +184,21 @@ def prompt(self, entry = None):
             return _("Enter new value for '%s' and press enter\n") % entry.title
 
     def input(self, entry, key):
-        if entry.aux.match(key):
+        valid, err_msg = check_name(key)
+        if entry.aux.match(key) and valid:
```
vpodzime: These two conditions should be swapped as checking the ``valid`` boolean is cheaper than trying to do a regexp match.

pyanaconda/users.py
```diff
@@ -166,6 +167,21 @@ def validatePassword(pw, user="root", settings=None, minlen=None):
 
     return (valid, strength, message)
 
+def check_name(in_name):
+    ign_cond = {r"^[\d]": "User name cannot start with a digit: ",
+                r"[^A-Za-z0-9\s]": "User name cannot contain any non-alphanumerical character: "}
```
vpodzime: I think ``_`` is missing and the ``\s`` is extra here. Or can user name contain a space?

vpodzime:
Looks good to me otherwise.

vpodzime:
Looks good to me now.

KosiehBarter:
Pushed.

PR #517, opened 2016-02-22
NTP should have better behavior (#1309396)

jkonecny12:
Fix when the ntp servers were in KS then chrony package was installed only with GUI but not in TUI.

Change a behavior of NTP package (chrony) installation and config file creation:

1) When --nontp is in KS then file don't install package and don't create config file.
2) When "-chrony" is in packages section of KS then don't install package but create the config file based on the config file from the installation environment.
3) If nothing used and there are ntp servers then create config file and install NTP package.

The "--nontp" option beats the "-chrony" packages rule.

Resolves: rhbz#1309396
Resolves: rhbz#1260725

### DIFF
--- a/pyanaconda/install.py
+++ b/pyanaconda/install.py
@@ -211,6 +211,9 @@ def doInstall(storage, payload, ksdata, instClass):
     ksdata.authconfig.setup()
     ksdata.firewall.setup()
     ksdata.network.setup()
+    # Setup timezone and add chrony as package if timezone was set in KS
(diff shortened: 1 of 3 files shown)


vpodzime:
Looks good to me.

jkonecny12:
Thank you for pointing me on this. I'll fix it locally.

jkonecny12:
Pushed to master and f24 but for RHEL it will need changes so I'll create new PR.
Thank you for review @vpodzime .

PR #518, opened 2016-02-22
Use CSS to style the internal widgets.

dashea:
Provide names for the stylable internal widgets, and use CSS instead of
Pango attributes and markup to set the style attributes. This allows the
appearance of the anaconda widgets to be more easily modified by
product-specific stylesheets.

The new widget names are only applied to the leaf widgets (labels and
buttons and such) since making the containers mess-with-able sounds like
trouble, and the type and existence of the containers could change if we
decide to switch to gadget types.

As part of the changes for moving things to CSS, what looks like the
intent of anaconda_mountpoint_selector_toggle_background (unselected
selectors are gray, selected selectors are black) actually works now,
and the incomplete state of AnacondaSpokeSelector is now tied to the
inconsistent widget flag.

The CSS data is stored in regular CSS files that get compiled into
libAnacondaWidgets.so, since that seemed less gross than a bunch of
multi-line C strings. This adds a bit of Makefile framework for creating
the resource bundle which could be used for other types of data files.

Resolves: rhbz#1243929

A version for master/rawhide will come later since I want to do a couple of extra things there (like moving the icons into the resource bundle) that I do not want to do in RHEL.


dashea:
Hey, guess who didn't look at any of the new generated docs. Copying things from gtk didn't do so hot, so the doc comments need a second look.

dashea:
The version of gtk-doc in RHEL 7.2 does not fully support the markdown I copied from gtk sources, so I changed it all to ugly SGML.

I'm still having trouble with the unicode line characters I also copied from gtk, but I'm going to assume that's a fixable font problem and ignore it.

jkonecny12:
Nice patch. Looks good to me and I also think this should be used in master. 
Do you have plans to commit this to master too?

dashea:
Pushed.

PR #519, opened 2016-02-23
makebumpver: Add a --dry-run option

bcl:
This allows the acks and jenkins state to be checked before running an
actual build.


bcl:
Pushed.

PR #520, opened 2016-02-23
Fix pki cert path in initrd (#ticket)

bcl:
It is under /etc/ not /

Resolves: rhbz#1303855


vpodzime:
Looks good to me.

bcl:
Pushed. With 'fixed' title. Except that the bz# was wrong and I didn't notice until too late.

PR #524, opened 2016-02-25
Generate resources.h before compiling widgets-common.c

dashea:
Pretty sure something was supposed to figure this out automatically, but oh well


jkonecny12:
Looks good to me.

PR #525, opened 2016-02-26
Fix nfs source crash when options change (#1264071)

bcl:
When stage2 is not nfs (eg. http) and the repo is on nfs the source
setup code would get confused on the 2nd pass through, thinking it was a
DRACUT_ISODIR that had been setup when in fact it was INSTALL_TREE. They
both point to the same place.

Resolves: rhbz#1264071


jkonecny12:
Seems good to me.

bcl:
Pushed

PR #526, opened 2016-02-26
NTP should have better behavior (#1309396)

jkonecny12:
Fix when the ntp servers were in KS then chrony package was
installed only with GUI but not in TUI.

Change a behavior of NTP package chrony installation and
config file creation:

1) When --nontp is in KS file then don't install ntp package and don't
create config file.
2) When "-chrony" is in packages section of KS then don't
install package but create the config files based on the config file
from the installation environment.
3) If nothing used and there are ntp servers then create config file and
install chrony package.

The "--nontp" option beats the "-chrony" packages rule.

Based on commit db800dc1bcb62cb8f811adae09a5b07c4f71ea5f but in RHEL
there is ntpd too. Which is handled as before this patch so we creating
the configuration files for ntp if ntp package is installed and --nontp
is not used.

Resolves: rhbz#1309396
Resolves: rhbz#1260725

*Based on PR https://github.com/rhinstaller/anaconda/pull/517 which is already in master now*

### DIFF
--- a/pyanaconda/install.py
+++ b/pyanaconda/install.py
@@ -204,6 +204,9 @@ def doInstall(storage, payload, ksdata, instClass):
     # Check for additional packages
     ksdata.authconfig.setup()
     ksdata.firewall.setup()
+    # Setup timezone and add chrony as package if timezone was set in KS
+    # and "-chrony" wasn't in packages section and/or --nontp wasn't set.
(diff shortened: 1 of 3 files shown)


vpodzime:
Looks good to me.

jkonecny12:
Pushed

PR #528, opened 2016-02-26
Change the install test URL.

dashea:
Everything moved because that is what Fedora does best.

The fedora-secondary URL will probably change at some point, but it
hasn't yet.


vpodzime:
Looks good to me.

jkonecny12:
I was adding the ACK here but than I started to change a rsync on our servers. 
Thank you for pointing me on this change.

dashea:
Pushed.

PR #529, opened 2016-02-26
Check to see if DD repo is already in addOn list (#1268357)

bcl:
When the repodata was re-downloaded it would add the DD repos again
without checking to see if they were already there.

Resolves: rhbz#1268357


vpodzime:
Looks good to me.

bcl:
Pushed.

PR #530, opened 2016-02-29
Hotfix for missing storage in payload class (#1271657)

jkonecny12:
The problem here is that payload is setting up storage when
PayloadManager run a thread but the user could do steps which
will call the spaceRequired() method sooner.

Now the behavior is the same as it was with the old spaceRequired
method for the variant when storage is not set.

Resolves: rhbz#1271657

*This is more general (hot)fix for the bugs "NoneType has no mountpoint"*

### DIFF
--- a/pyanaconda/packaging/dnfpayload.py
+++ b/pyanaconda/packaging/dnfpayload.py
@@ -585,6 +585,10 @@ class DNFPayload(packaging.PackagePayload):
     @property
     def spaceRequired(self):
         size = self._spaceRequired()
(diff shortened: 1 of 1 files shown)


vpodzime:
Looks good to me.

jkonecny12:
It's pushed to master already but I think this should go to F24 branch too. 
To give others time to react on push to F24 branch I'll push this tomorrow.

jkonecny12:
Pushed to f24 too.

PR #531, opened 2016-02-29
Behave better in the abscence of locales

dashea:
Concede that everything is an absolute disaster and try to muddle through


dashea:
Added a commit so that the GUI is not translated into an unsupported locale.

jkonecny12:
In the first commit the last word in the last sentence. Maybe it's my bad English but I think it shouldn't be there or you want there ``anywhere`` instead.
Otherwise looks good to me.

vpodzime:
> In the first commit the last word in the last sentence. Maybe it's my bad English but I think it shouldn't be there or you want there anywhere instead.

I think ``anymore`` makes sense there too.

dashea:
Pushed to master and f24.

PR #532, opened 2016-02-29
Use the default levelbar offset values.

dashea:
The effect is the same as the previous CSS rule, but now we're working
with instead of around.


jkonecny12:
Looks good to me.

dashea:
Pushed to master and f24.

PR #533, opened 2016-03-02
Added newline escapes to fit in 80-char length

KosiehBarter:
Resolves rhbz#1267881

### DIFF
--- a/pyanaconda/rescue.py
+++ b/pyanaconda/rescue.py
@@ -152,9 +152,9 @@ class RescueMode(NormalTUISpoke):
                          "to find your Linux installation and mount it under "
                          "the directory : %s.  You can then make any changes "
                          "required to your system.  Choose '1' to proceed with "
-                         "this step.\nYou can choose to mount your file "
+                         "this step. You can choose to mount your file "
                          "systems read-only instead of read-write by choosing "
-                         "'2'.\nIf for some reason this process does not work "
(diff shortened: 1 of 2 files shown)


pyanaconda/rescue.py
```diff
@@ -149,13 +149,13 @@ def refresh(self, args=None):
...
-                         "the directory : %s.  You can then make any changes "
+                         "to find your Linux installation and\nmount it under "
+                         "the directory : %s.  You can then make any changes\n"
                          "required to your system.  Choose '1' to proceed with "
                          "this step.\nYou can choose to mount your file "
-                         "systems read-only instead of read-write by choosing "
+                         "systems read-only instead of read-write by\nchoosing "
                          "'2'.\nIf for some reason this process does not work "
-                         "choose '3' to skip directly to a shell.\n\n") % (iutil.getSysroot())), ""]
+                         "choose '3' to skip directly\nto a shell.\n\n") % (iutil.getSysroot())), ""]
```
vpodzime: Please move the parts of the strings after ``\n`` to the next line so that the newlines are better visible.

vpodzime:
I don't think this is the right way to fix the issue. The ``TextWidget`` should take care about this implicitly.

rvykydal:
I've made a new PR https://github.com/rhinstaller/anaconda/pull/667.

PR #534, opened 2016-03-02
Changed INSTALL_PROGRESS to ERROR

KosiehBarter:
NOTE:
I am not able to force a error to reproduce error message for check.

My attempt to reproduce any of error messages:
1) Fix the bug (changing INSTALL_PROGRESS to ERROR)
2) Started VM, netinstall from PXE
3) Manually set options in Anaconda
4) Start the installation
5) During installation:
   - in shell: ip link set <dev> down
   - return back to GUI
6) Wait for error message to appear
 --> FAILED, installation crashed after some retries
 --> No message to confirm bug fix
7) Whole proccess repeated five times
 --> All of the attempts were unsuccessful.

However, changing some strings should not affect rest of the code in
a major way.

Resolves rhbz#1255094

### DIFF
--- a/pyanaconda/packaging/yumpayload.py
+++ b/pyanaconda/packaging/yumpayload.py
@@ -1446,6 +1446,7 @@ reposdir=%s
         progress_map = {
             "PROGRESS_PREP"    : _("Preparing transaction from installation source"),
             "PROGRESS_INSTALL" : _("Installing"),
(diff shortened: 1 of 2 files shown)


dashea:
Closing the PR since this was pushed as 1f77cb620d7d9e99a2bd00e864486acc47574762

PR #535, opened 2016-03-02
Now possible to use DUD from http (#1261024)

japokorn:
Fixed bug in driver update disk logic when using inst.dd=http:/... did not activated network.
This bug manifested itself when no other parameter activated network.

Fixed by adding logic to ask for network activation when needed.

Resolves: rhbz#1261024

### DIFF
--- a/dracut/parse-anaconda-options.sh
+++ b/dracut/parse-anaconda-options.sh
@@ -129,6 +129,15 @@ getargbool 0 vnc inst.vnc && warn "anaconda requiring network for vnc" && set_ne
 
 # Driver Update Disk
 warn_renamed_arg "dd" "inst.dd"
+# Net may be needed 
+if dd=$(getarg dd inst.dd); then
+    if [ -n "$dd" ]; then
+        case $dd in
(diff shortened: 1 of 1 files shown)


jkonecny12:
Looks good to me.

vpodzime:
Same here.

dashea:
There can be multiple inst.dd arguments, and I'm not sure that this would handle all of those cases correctly. (for example, inst.dd inst.dd=http://...). There's already a loop over the dd args in dracut/parse-anaconda-dd.sh, so I think it would be easier to add the set_neednet there.

This would be similar to how set_neednet is added for kickstart (https://github.com/rhinstaller/anaconda/blob/master/dracut/parse-anaconda-kickstart.sh#L17) and repo (https://github.com/rhinstaller/anaconda/blob/master/dracut/parse-anaconda-repo.sh#L18) options.

dashea:
I pushed https://github.com/rhinstaller/anaconda/pull/647, which removes the need for this one and works with multiple inst.dd arguments.

PR #536, opened 2016-03-02
Increase yum debug logging level from 3 to 6 (#1254368)

jkonecny12:
Resolves: rhbz#1254368


jkonecny12:
Tests are failing but it's because some pykickstart missing stuff.

vpodzime:
Looks good to me.

jkonecny12:
Pushed

PR #537, opened 2016-03-02
liveinst changes for Xwayland

dashea:
Clean up a couple of aspects of the liveinst xinitrc script, make it run in an xinitrc-less world, and add another environment variable to liveinst.


dashea:
Hopefully the autostart script doesn't do anything weird in non-GNOME window managers, where I assume X-GNOME-AutoRestart does not mean anything, but when I tried to test it I hit dependency errors in the KDE packages in rawhide right now. So really it's their own fault.

jkonecny12:
Looks good to me.

dashea:
Added some more shell crud and pushed to f24 and master

PR #539, opened 2016-03-04
Bios RAID traceback on Filter GUI (#1309069)

kellinm:
The patch resolving rhbz#914637 exposed issues with the filter gui and
its use of mocked udev data for bios RAID devices. Refactored the code
tracking selected devices, total devices, and disk size to use udev
wrappers instead of the mocked udev data.

Resolves: rhbz#1309069

### DIFF
--- a/iw/filter_gui.py
+++ b/iw/filter_gui.py
@@ -139,22 +139,30 @@ class Callbacks(object):
 
     def incrementDeviceCount(self, device_data):
         """Update the selected and total device counts when a device is added."""
+
+        deviceName = udev_device_get_name(device_data[UDEV_DATA])
+        deviceSize = device_data[UDEV_DATA]["XXX_SIZE"]
+
+        self._incrementDeviceCount(deviceName, deviceSize, device_data[ACTIVE_IN_UI])
+
+
+    def _incrementDeviceCount(self, deviceName, deviceSize, isActiveInUI):
         global totalDevices, totalSize
         global selectedDevices, selectedSize
         global talliedDevices
 
-        # Check to be sure this device has not already been counted
-        if device_data[UDEV_DATA]["name"] in talliedDevices or device_data[UDEV_DATA]["DEVNAME"] in talliedDevices:
+        # do not tally devices that were already counted
+        if deviceName in talliedDevices:
             return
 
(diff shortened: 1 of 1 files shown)


jkonecny12:
Look good to me AFAIK.

iw/filter_gui.py
```diff
@@ -240,29 +248,23 @@ def incrementDeviceCount(self, device_data):
...
-        if device_data[UDEV_DATA]["name"] in talliedDevices:
-            return
+        deviceName = udev_device_get_name(device_data[UDEV_DATA])
+        mpathName = udev_device_get_multipath_name(device_data[UDEV_DATA])
+        deviceSize = device_data[UDEV_DATA]["XXX_SIZE"]
 
-        # If the mpath member is already in the list, add the mpath device name
-        if device_data[UDEV_DATA]["DEVNAME"] in talliedDevices:
-            talliedDevices.append(device_data[UDEV_DATA]["name"])
+        if mpathName == None:
```
dwlehman: Should be `if mpathName is None:`.

iw/filter_gui.py
```diff
@@ -924,8 +926,6 @@ def _active(info):
             # We use a copy here, so as to not modify the original udev info
             # dict as that would break NameCache matching
             data = mpath[0].copy()
-            # This will swap the device name (eg: sda) for the mpath name (eg: mpatha)
-            data["name"] = udev_device_get_multipath_name(mpath[0])
```
dwlehman: What are you fixing by changing this?

kellinm: The first time an iSCSI mpath converts from single disk to an mpath, it won't detect the change because this overwrites the actual device name (sda) so it doesn't know to remove it.  This is an artifact from the earlier code where it used this instead of the udev call to look up the mpath name.

dwlehman: Okay, I think I understand as much as I want to now. If you're not changing the name you also shouldn't be making (and using) a copy of the actual device data.

dwlehman:
Any change to this file feels a bit like like a roll of the dice, but this looks okay from my perspective given the testing it has already received.

kellinm:
Pushed manually.

PR #540, opened 2016-03-04
Translated strings

dashea:
The rhel7 bug still has some process to go through, but I'd like to go ahead and get this on f24 before string freeze.


dashea:
Updated the help button commit to also translate the string used as the accessible name.

M4rtinK:
Looks good to me & thanks for looking into that! :)

Frodox:
+1 from bug reporter :)

dashea:
Pushed to master, f24, and rhel7

PR #541, opened 2016-03-07
Fix a bad usage of execWithRedirect (#1270319)

jkonecny12:
execWithRedirect returns an int, not a tuple, and the argv used to
execute realm still had the command as argv[0], which is no longer
needed.

(cherry picked from commit 08a55ebeb6ce6fa9e66332bb9f34d0d69ea80ecc)

Resolves: rhbz#1270319


vpodzime:
Looks good to me.

jkonecny12:
Pushed

PR #542, opened 2016-03-07
don't install kernel-PAE on x86_64 (#1313957)

AdamWill:
PAE is *available* on x86_64 systems, but we do not want to
install kernel-PAE on them even if it is in the repo for some
reason.

An alternative to this is to rename `isPaeAvailable()` to
`isPaeWanted()` and have it return False on x86_64, but I guess
that's technically an interface change, and this is also easy.

### DIFF
--- a/pyanaconda/packaging/__init__.py
+++ b/pyanaconda/packaging/__init__.py
@@ -751,7 +751,7 @@ class PackagePayload(Payload):
 
         kernels = ["kernel"]
 
-        if isys.isPaeAvailable():
+        if blivet.arch.isX86(32) and isys.isPaeAvailable():
             kernels.insert(0, "kernel-PAE")
 
         # most ARM systems use platform-specific kernels


AdamWill:
We would want this on master and f24, as the issue is a blocker for F24 Alpha.

AdamWill:
I tested this with an updates.img, it seems to work. An x86_64 network install from a current F24 Server `boot.iso` with this patch installs an x86_64 kernel and boots.

vpodzime:
This looks good to me as it makes the x86 code similar to what we do for arm.

AdamWill:
In case anyone wondered, note this, from `isPaeAvailable`:

    isPAE = False
    if not blivet.arch.isX86():
        return isPAE

i.e. it never returns `True` on any non-x86 arch. So there's definitely not some exotic arch other than x86_32 where we're expecting to install kernel-PAE and this change would break it.

vpodzime:
Pushed to master and f24-branch.

PR #543, opened 2016-03-08
except block in py3.5 undefines the variable

bcl:
try:
    foo = "I tried!"
    raise Exception(foo)
except Exception as foo:
    foo = str(foo) +  " bar"
finally:
    print(foo)

The except block will undefine foo, making finally crash.


vpodzime:
Looks good to me.

PR #544, opened 2016-03-08
Use a lock for repoStore access (#1315414)

bcl:
The spoke.changed method could be called while _reset_repoStore is
altering the list of repos. This adds a lock to _update_payload_repos
and _reset_repoStore to prevent problems.

Resolves: rhbz#1315414


bcl:
Should probably do this on all the branches.

vpodzime:
Looks good to me.

bcl:
Pushed to all the branches

PR #545, opened 2016-03-09
Fix a DBus InvalidProperty handling (#1315843)

jkonecny12:
Anaconda is crashing now because new InvalidProperty exception from
NetworkManager is now raised.

Resolves: rhbz#1315843

(cherry picked from commit 69a90585c3bb0f924964ff79d120b79ef738b0f4)

This is already in master.
NetworkManager upgraded version in rhel 7.3 so this error is in the rhel now and it's blocking testing of new composes. For this reason I'm going to break 24 hours rule to fix this error ASAP.


vpodzime:
Looks good to me.

jkonecny12:
Pushed.

PR #546, opened 2016-03-09
Load the system-wide Xresources (#1241724)

dashea:


jkonecny12:
Looks good to me.

dashea:
Pushed.

PR #547, opened 2016-03-10
Fix crash when new device appear in Welcome screen (#1216926)

jkonecny12:
Anaconda will crash when user add a new device in the Welcome screen.
This happens because the configuration file for the new device wasn't
created.
Fixed by generating this configuration in NetworkSpoke initialization.

Resolves: rhbz#1216926

Backporting part of the fix PR https://github.com/rhinstaller/anaconda/pull/274 .


vpodzime:
Looks good to me.

jkonecny12:
Pushed

PR #548, opened 2016-03-11
Rhel7 fix create bad dev name ks

jkonecny12:
This patch will do these 2 things:

1. Fixing issue when ifname is used in boot_args. We then can't use ``ksdev0`` as network device name but we should use the ifname one.
2. Fixing behavior with prioritizing network settings in  dracut and stage2.

It should work in this way:

* If there are any ``ip=<iface>:<IP>`` boot arg in Dracut this will be applied in Dracut and kickstart network commands will be applied later when an installation (stage2) starts.
* If there aren't any ``ip=<iface>:<IP>`` boot args in Dracut we will parse kickstart and apply that settings to Dracut.

In short: Dracut will prioritize the network boot args settings over kickstart one but in the stage2 kickstarts will be applied (if --activate is used).

I'm adding diagram with the network work-flow for easier explanation. If there is anything you think it could break please comment. Also comment if you think this behavior is bad.

Network Dracut diagram

I'm going to finish this diagram with (as best scenario) all network Dracut/Anaconda setups to exactly specify how it should behave for future.

### DIFF
--- a/dracut/parse-kickstart
+++ b/dracut/parse-kickstart
@@ -193,8 +193,9 @@ class Network(commands.network.RHEL7_Network, DracutArgsMixin):
                     return
                 else:
                     log.info("Using %s as first device with link found", net.device)
-            # tell dracut to bring this device up
-            netline = ksnet_to_dracut(args, lineno, net, bootdev=True)
+            # tell dracut to bring this device up if it's not already done by user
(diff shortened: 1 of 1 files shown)


rvykydal:
Looks good to me.

jkonecny12:
Pushed

PR #549, opened 2016-03-11
Language attributes

dashea:
In certain cases, Pango needs to know what language to render text in. The default language cannot be changed once the application has started, so work around that by explicitly setting language attributes on all labels instead.

### DIFF
--- a/pyanaconda/ui/gui/utils.py
+++ b/pyanaconda/ui/gui/utils.py
@@ -520,3 +520,70 @@ def find_first_child(parent, match_func):
             search_list.extend(widget.get_children())
 
     return None
+
+
+_widget_watch_list = {}
+def watch_children(widget, callback, user_data=None):
+    """
+    Call callback on widget and all children of widget as they are added.
+
+    Callback is a function that takes widget and user_data as arguments. No
+    return value is expected.
+
+    Callback will be called immediately for widget, and, if widget is a
+    GtkContainer, for all children of widget. If widget is a container it will
+    be then be watched for new widgets added to the container, and callback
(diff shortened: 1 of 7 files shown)


pyanaconda/ui/gui/utils.py
```diff
@@ -520,3 +520,70 @@ def find_first_child(parent, match_func):
...
+
+        watch_children(widget, callback, user_data)
+
+    def _remove_signal(container, widget, user_data):
+        callback, user_data = user_data
+
+        unwatch_children(widget, callback, user_data)
+
+    callback(widget, user_data)
+    if (isinstance(widget, Gtk.Container)):
```
jkonecny12: The brackets ``()`` aren't required here.

pyanaconda/ui/gui/utils.py
```diff
@@ -520,3 +520,70 @@ def find_first_child(parent, match_func):
...
+
+            _widget_watch_list[signal_key] = (add_signal, remove_signal)
+
+def unwatch_children(widget, callback, user_data=None):
+    """
+    Unregister a callback previously added with watch_children.
+
+    :param GtkWidget widget: the widget to unwatch
+    :param function callback: the callback that was previously added to the widget
+    :param user_data: The user_data that was previously added to the widget
```
jkonecny12: You are using lower case ``the`` in the first two :param and you have ``The`` in the third case.

jkonecny12:
Seems good to me except to those two small neat picks.

dashea:
Fixed the capitalization, removed the C-style parens, and pushed.

PR #551, opened 2016-03-14
fix multiple inst.dd=<path> args (rhbz#1268792)

wgwoods:
Using `--unique` here with means that each job with the same `--name`
will overwrite the previous one, which means that if you do:

    inst.dd=/dud/dd1.iso inst.dd=/dud/dd2.iso

...then the latter overwrites the former and we only end up loading one
driver disk.

Removing `--unique` means that both jobs will be added, as expected.

Note that this *also* means that you'll get two jobs if you add the same
path twice, like so:

    inst.dd=/dud/dd1.iso inst.dd=/dud/dd1.iso

but processing the same DUD twice should be harmless, so that's fine.

Resolves: rhbz#1268792


vpodzime:
Extra *with* in the commit message. Otherwise this looks good to me too.

wgwoods:
commit message fixed, merged into master, cherry-picked to rhel7-branch.

PR #552, opened 2016-03-15
Fix bad sensitivity on boxes in source spoke (#1262833)

jkonecny12:
~~The boxes sensitivity for source types are changed when user clicks on radio button to change type of source.
Fix problem when user enter source spoke -> nothing were clicked and everything is sensitive.~~

**EDIT:**
Remove default sensitivity on Network box. It caused more gtk boxes were sensitive in one time.
Enable network button when we have functional connectivity.

*Reported-by: Pavel Holica <pholica@redhat.com>*

*Resolves: rhbz#1262833*

I don't know why but it doesn't happened on rawhide/f24 so this is rhel7 only.

### DIFF
--- a/pyanaconda/ui/gui/spokes/source.py
+++ b/pyanaconda/ui/gui/spokes/source.py
@@ -792,10 +792,6 @@ class SourceSpoke(NormalSpoke, GUISpokeInputCheckHandler):
 
         self._updateURLEntryCheck()
 
-        # Set up the default state of UI elements.
-        self._networkButton.set_sensitive(True)
-        self._networkBox.set_sensitive(True)
-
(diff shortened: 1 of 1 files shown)


vpodzime:
I think all the boxes should be made insensitive first and then one of the radio buttons should be activated ("clicked from code").

jkonecny12:
I completely reworked this patch. I think this is happening because we are setting up default value. 

I tried test with network, iso repo and without repo and it looks it works correctly now.

jkonecny12:
Updated again. I hope I fixed the root cause of this problem now.

vpodzime:
Looks good to me.

jkonecny12:
Pushed

PR #553, opened 2016-03-15
Add reason when logging invalid repository (#1240379)

jkonecny12:
Print also exception message which caused the repository removal.

*Resolves: rhbz#1240379*

### DIFF
--- a/pyanaconda/packaging/yumpayload.py
+++ b/pyanaconda/packaging/yumpayload.py
@@ -566,8 +566,8 @@ reposdir=%s
                                  proxyurl=proxyurl, sslverify=sslverify)
                 self._addAddons(self._yum.repos.getRepo(BASE_REPO_NAME), url, proxyurl, sslverify)
             except (MetadataError, PayloadError) as e:
-                log.error("base repo (%s/%s) not valid -- removing it",
-                          method.method, url)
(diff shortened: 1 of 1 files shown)


pyanaconda/packaging/yumpayload.py
```diff
@@ -566,8 +566,8 @@ def updateBaseRepo(self, fallback=True, root=None, checkmount=True):
                                  proxyurl=proxyurl, sslverify=sslverify)
                 self._addAddons(self._yum.repos.getRepo(BASE_REPO_NAME), url, proxyurl, sslverify)
             except (MetadataError, PayloadError) as e:
-                log.error("base repo (%s/%s) not valid -- removing it",
-                          method.method, url)
+                log.error("base repo (%s/%s) not valid -- removing it\nreason: %s",
+                          method.method, url, e)
```
bcl: The \n in there is going to mess up how the logs look. either log it as a separate line, or "-- removing it because: %s"

jkonecny12:
Updated.
The reason message is now on separate line and the message changed slightly to be more self explanation. This could be handy when it's grep-ed without the first line.
Thank you for pointing me on this @bcl .

jkonecny12:
Pushed

PR #554, opened 2016-03-15
network: use NAME to find ifcfg on s390 with net.ifnames=0 (#1249750)

rvykydal:
Anaconda creates extra connection because it doesn't recognize ifcfg file
created in dracut because it has neither HWADDR (because it is s390) nor DEVICE
(because we are using kernel names instead of persistent names, net.ifnames=0)
binding info. Assuming dracut's ifcfg file is fine, we probably need to look
for NAME in this case.

Resolves: rhbz#1249750


jkonecny12:
Looks good to me.

rvykydal:
Pushed

PR #555, opened 2016-03-15
Don't use pyblock/dmraid to get info about imsm/isw in filter gui

dwlehman:
Using dmraid for imsm/isw was never supported, but we did it in the filter gui because it was simpler for us. Now that a case has arisen in which dmraid/pyblock actually fail to do the job (see bug reference below) we have to switch over.

Resolves: [rhbz#1269872](https://bugzilla.redhat.com/show_bug.cgi?id=1269872)


vpodzime:
Looks good to me otherwise.

snbueno:
This is cleaner than I thought it would be. From what I can tell, it looks ok to me as well.

kellinm:
This looks good to me. I would hold off on the sector size thing @vpodzime - if we change it here, we'd want to change it everywhere it shows up which, after a quick search, seems to be just over 70 potential locations in the codebase.

Also - after talking to @dwlehman it seems that for some appearances it's not so much sector size as an arbitrary constant that happens to be the same value for now. If we want to improve readability with constants, I'm 100% for that but let's push that to the next point release so we can do it one time cleanly for everything.

iw/filter_gui.py
```diff
@@ -878,7 +890,7 @@ def _active(info):
                                  list(rs.get_members()))
                 members = map(lambda m: m.get_devpath(), members)
                 for d in raids:
-                    if udev_device_get_name(d) in members:
+                    if udev_device_get_name(d) in members and d not in mdraids:
                         fstype = udev_device_get_format(d)
                         sysfs_path = udev_device_get_sysfs_path(d)
                         break
```
dwlehman: I think I want to replace this hunk with the following:

```diff
@@ -878,6 +890,9 @@ class FilterWindow(InstallWindow):
                                  list(rs.get_members()))
                 members = map(lambda m: m.get_devpath(), members)
                 for d in raids:
+                    if d in mdraids:
+                        break
+
                     if udev_device_get_name(d) in members:
                         fstype = udev_device_get_format(d)
                         sysfs_path = udev_device_get_sysfs_path(d)

```

I will retest with the change and update the PR when it checks out.

dwlehman:
Changes pushed just now:

- added `LINUX_SECTOR_SIZE` constant and used it in `filter_gui.py`
- break from loop instead of iterating over all of `raids` when the current raid is isw/imsm
- move `mduuids` definition/declaration closer to where it is used

dwlehman:
All of the changes made were in "Don't use pyblock..."

iw/filter_gui.py
```diff
@@ -902,6 +917,51 @@ def _active(info):
...
+                # time through. If the devicetree was already populated this
+                # should have no effect since mdadm.conf will already be
+                # written out. Note that the fwraid arrays themselves are
+                # not in mdadm.conf until after they are activated, resulting
+                # in them having names like md127.
+                uuid = udev_device_get_md_uuid(mdmember)
+                name = "md%d" % len(mduuids)
+                ensure_mdadm_conf_entry("/dev/" + name, uuid)
+                if uuid not in mduuids:
+                    mduuids.append(uuid)
```
vpodzime: This would be slightly more optimal if ``mduuids`` was a ``set()``, but that's just a nitpick.

iw/filter_gui.py
```diff
@@ -866,10 +874,13 @@ def _active(info):
...
 
         if raids and flags.dmraid:
+            # isw/imsm is handled by mdadm -- not dmraid/pyblock
+            mdraids = [d for d in raids
+                       if udev_device_get_format(d) == "isw_raid_member"]
             used_raidmembers = []
             for rs in block.getRaidSets():
                 # dmraid does everything in sectors
-                size = (rs.rs.sectors * 512) / (1024.0 * 1024.0)
+                size = (rs.rs.sectors * LINUX_SECTOR_SIZE) / (1024.0 * 1024.0)
```
vpodzime: This could be ``1024**2`` just like below.

vpodzime:
Looks good to me other than the two nitpicks above.

dwlehman:
Pushed manually to rhel6-branch.

PR #556, opened 2016-03-16
Fix only one address is shown in anaconda (#1264400)

jkonecny12:
Only ``NMIP[4,6]Config`` object was connected to signal on change. But when inner objects (Addresses, Nameservers, Gateway) changed it doesn't notify us about this change. 
Fix by connecting this inner properties too.

*Resolves: rhbz#1264400*

*Reported-by: ``Marek Hruscak <mhruscak@redhat.com>``*


rvykydal:
Looks good to me.
Just a note, on master we'll need to consider whether we need to disconnect the callbacks in disconnect_client_callbacks.

jkonecny12:
Pushed to rhel7, it seems it must be changed to master. So I'm closing this PR and I'm going to create new for the master.

PR #557, opened 2016-03-16
Switch to the adwaita icon theme.

dashea:
adwaita-icon-theme and gnome-icon-theme merged a while back, so we can
use the default icon theme settings now and still get the same icons
that GNOME uses. Also, the gnome-icon-theme packages in Fedora are no
longer updated, so this way we keep up with any changes.


dashea:
Pushed.

PR #558, opened 2016-03-16
Only use <> for markup (#1317297)

bcl:
Zanata error checking gets confused when a string looks like markup but
isn't. So convert <Enter> to [Enter] and change several other uses to
UPPER CASE.


jkonecny12:
Looks good to me.

bcl:
This now passes the tests.

jkonecny12:
Looks good to me again :) .

bcl:
Pushed.

PR #559, opened 2016-03-21
Show the close button in metacity (#1319590)

dashea:
This will not affect anaconda windows, since the titlebar is not shown
on anaconda windows.  However, for the two cases where anaconda runs an
external program (yelp and nm-connection-editor), this will display a
close button in the titlebar, allowing the program to actually be
closed.

Thanks to Ray Strode (halfline) for the change.


dashea:
Hold up, I've got a better one

jkonecny12:
Looks good to me.

dashea:
Pushed.

PR #560, opened 2016-03-21
Add the /EFI directory to the boot.iso (#1188876)

bcl:
The boot.iso should be able to boot on UEFI systems, currently they have
the UEFI boot image but are missing the contents of the /EFI directory.

Resolves: rhbz#1188876


snbueno:
Looks ok to me.

bcl:
Pushed.

PR #561, opened 2016-03-22
Add retry when downloading .treeinfo (#1292613)

jkonecny12:
Thank you for this patch Bryan.

Based on the patch from Bryan Wann.

Resolves: rhbz#1292613

Reported-by: Bryan Wann ``<bwann@wann.net>``

This is kind of workaround for the root issue where NetworkManager is set to ``CONNECTED`` state when it gets ``LOCAL`` address but this is not so invasive solution than changing the state behavior. The final solution will go to master where we will wait on connectivity to repository. After we get this connectivity then NetworkManager will be in ``CONNECTED`` state.

### DIFF
--- a/pyanaconda/packaging/__init__.py
+++ b/pyanaconda/packaging/__init__.py
@@ -36,6 +36,7 @@ import shutil
 from glob import glob
 import re
 import threading
+import time
 
 if __name__ == "__main__":
     from pyanaconda import anaconda_log
@@ -53,7 +54,7 @@ from pyanaconda import isys
 from pyanaconda.image import findFirstIsoImage
 from pyanaconda.image import mountImage
 from pyanaconda.image import opticalInstallMedia
-from pyanaconda.iutil import ProxyString, ProxyStringError
+from pyanaconda.iutil import ProxyString, ProxyStringError, xprogressive_delay
 from pyanaconda.regexes import VERSION_DIGITS
 from pyanaconda.threads import threadMgr, AnacondaThread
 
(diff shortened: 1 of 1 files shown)


pyanaconda/packaging/__init__.py
```diff
@@ -428,18 +431,51 @@ def _getTreeInfo(self, url, proxy_url, sslverify):
...
-            treeinfo = ug.urlgrab("%s/.treeinfo" % url,
-                                  "/tmp/.treeinfo", copy_local=True,
-                                  proxies=proxies, **ugopts)
-        except URLGrabError as e:
+
+        # Retry treeinfo downloads with a progressively longer pause,
+        # so NetworkManager have a chance setup a network and we have
+        # full connectivity before trying to download things. (#1292613)
+        xdelay = xprogressive_delay()
+        for retry_count in xrange(0, MAX_TREEINFO_DOWNLOAD_RETRIES + 1):
```
vpodzime: Why ``+ 1``?

vpodzime:
> The final solution will go to master where we will wait on connectivity to repository before NetworkManager will get CONNECTED state.

"before" NM will get CONNECTED state? I'd understand "after" here, but "before" doesn't make much sense. In general, why are we not waiting for NM to get a GLOBAL connection?

pyanaconda/packaging/__init__.py
jkonecny12: This code is borrowed from this commit https://github.com/rhinstaller/anaconda/commit/8c2544c6f537240179569d0068a7b22250e21a25. But from my understanding the first is not retry so that is why 0. +1 because the xrange is open interval from the end.

jkonecny12:
Updated I hope it's more understandable now. Thank you finding this.

We can't wait on ``GLOBAL`` because repository and everything could be only on ``LAN``, on network without ``GLOBAL`` connectivity.

pyanaconda/packaging/__init__.py
vpodzime: Makes sense, thanks for the extra info.

vpodzime:
Looks good to me.

jkonecny12:
Pushed

PR #562, opened 2016-03-22
Multiple Dogtail tests improvements

atodorov:
- use unittest.skipIf and execute via nosetests_root.sh
- remove support for anacondaArgs because it was not used
- screenshots are placed in the correct test results directory
- logs are collected on test failures too
- cleanup mounted filesystems and update anaconda-cleanup to
use the new BlockDev loop plugin along the way

This is in order to add these tests to make ci. The CI environment only needs X display configured and accessibility enabled.


atodorov:
Ping, any updates here ? 

Btw the CantReclaimTestCase fails b/c there is no more "No Space" dialog present. The user is left with the option to reclaim disk space, although the entire disk size is smaller that what is required to install the selected package set. Is this intentional ?

tests/gui/base.py
```diff
@@ -351,11 +361,11 @@ def die(self, terminate=False):
...
 
-            # tests will click the Reboot or Quit button which will shutdown anaconda
-            # we need to make sure /mnt/sysimage/* are unmounted and device mapper devices
-            # are removed before starting the next test. ATM I have no idea if pyanaconda
-            # provides an API for this so wait for a while instead!
-            doDelay(30)
+            # Tests will click the Reboot or Quit button which will shutdown anaconda.
+            # We need to make sure /mnt/sysimage/* are unmounted and device mapper devices
+            # are removed before starting the next test.
+            subprocess.call(["%s/scripts/anaconda-cleanup" % os.environ.get("top_srcdir", "")],
```
clumens: If $top_srcdir isn't set set, this will mean /scripts/anaconda-cleanup will get run.  You probably want ./scripts/anaconda-cleanup to get run instead.  So:

os.environ.get("top_srcdir", ".")

clumens:
Merged manually, thanks.

PR #565, opened 2016-03-23
Default locale

dashea:
Don't try to use C.UTF-8 anymore since, at least in the short term, it's too much trouble to deal with, and ensure that en_US.UTF-8 actually has the data for that regardless of how anaconda is installed.


jkonecny12:
Looks good to me.

dashea:
Pushed.

voxik:
Why anaconda forces EN od ALL langpacks? I have CS langpack installed and I hope that is enough for my system (why do I have Anaconda on my installed system anyway, going to remove it ...)

PR #566, opened 2016-03-23
Fix the use of CSS psuedo-classes in the widgets.

dashea:
The selectors that depend on state flags set in container classes
stopped working, probably because a more-specific selector in the
default theme overrode them. Propagate state flag changes to the leaf
widgets so that the state-sensitive selectors work again and so that
this can't happen again.

Also switch to the new pseudo-class names that match the CSS
psuedo-class names (but no longer match the names of the state flags
that drive them), which gets rid of all of those deprecation warnings.


jkonecny12:
Looks good to me.

PR #568, opened 2016-03-24
Don't crash if no groups are specified (#1316816)

dashea:
If the groups input field in the GUI is empty, set the groups list to an
empty list instead of a list containing an empty string.

### DIFF
--- a/pyanaconda/ui/gui/spokes/user.py
+++ b/pyanaconda/ui/gui/spokes/user.py
@@ -156,7 +156,8 @@ class AdvancedUserDialog(GUIObject, GUIDialogInputCheckHandler):
         else:
             self._user.gid = None
 
-        self._user.groups = [g.strip() for g in self._tGroups.get_text().split(",")]
+        # ''.split(',') returns [''] instead of [], which is not what we want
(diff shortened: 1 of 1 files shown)


pyanaconda/ui/gui/spokes/user.py
```diff
@@ -156,7 +156,12 @@ def apply(self):
         else:
             self._user.gid = None
 
-        self._user.groups = [g.strip() for g in self._tGroups.get_text().split(",")]
+        groups_string = self._tGroups.get_text()
+        # ''.split(',') returns [''] instead of [], which is not what we want
+        if groups_string:
+            self._user.groups = [g.strip() for g in groups_string.split(",")]
+        else:
+            self._user.groups = []
```
vpodzime: Wouldn't adding ``if g`` (or better ``if g.strip()``) to the end of the list comprehension do the same (plus check for empty strings as groups)?

dashea:
Updated to put the conditional into the comprehension

jkonecny12:
Looks good to me.

dashea:
Pushed to master and f24.

PR #569, opened 2016-03-24
Do not allow liveinst with --image or --dirinstall (#1276349)

dashea:


vpodzime:
Looks good to me.

dashea:
Pushed.

PR #570, opened 2016-03-24
Add resume= parameters for swap devices (#1224151)

dashea:
This way dracut knows where to potentially look for resume data


vpodzime:
I don't really understand why we should be doing this. Wouldn't it be better for the thing that does the hibernation (systemd?) to write the device somewhere? And has something changed over the years that we suddenly need to do this?

dashea:
I thought that this was because of encrypted swap, but we already add the rd.luks stuff to make that work, and it totally used to work, so I don't know what's going on. Reassigned the bug to dracut, and let's see if that goes anywhere.

PR #571, opened 2016-03-30
Add access to the payload from addons (#1288636)

jkonecny12:
This should add new payload argument to the ``setup()`` and ``execute()`` methods in the ``AddonData`` class without breaking existing API.

Add deprecation warning for old method signature.

*Resolves: rhbz#1288636*

Reported-by: Pat Riehecky ``<riehecky@fnal.gov>``

If you want to  test it. I'm using this https://github.com/jkonecny12/hello-world-anaconda-addon/tree/master-payload-testing for testing.

### DIFF
--- a/pyanaconda/addons.py
+++ b/pyanaconda/addons.py
@@ -23,7 +23,10 @@ __all__ = ["AddonSection", "AddonRegistry", "AddonData", "collect_addon_paths"]
 
 import os
 import functools
+import logging
 from pykickstart.sections import Section
+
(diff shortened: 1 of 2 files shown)


jkonecny12:
OSCAP and KDUMP addons are deprecated thanks to this.

I'll change the addons after this patch will be there as related changes.

vpodzime:
Looks good to me.

jkonecny12:
Pushed, thank you.
Now I'll take care of the addons :) .

PR #572, opened 2016-03-30
Stop kickstart when space check fails (#1320436)

bcl:
Text mode kickstart behavior was inconsistent, it would allow an
installation to continue even though the space check failed. Every other
install method stops, letting the user add more space before continuing.

Resolves: rhbz#1320436

### DIFF
--- a/pyanaconda/packaging/yumpayload.py
+++ b/pyanaconda/packaging/yumpayload.py
@@ -1332,6 +1332,10 @@ reposdir=%s
         self.selectRequiredPackages()
 
     def checkSoftwareSelection(self):
+        if not self.baseRepo:
+            log.warn("Skipping checkSoftwareSelect, no base repo setup.")
+            return
+
         log.info("checking software selection")
         self.txID = time.time()
 
--- a/pyanaconda/ui/gui/spokes/software.py
+++ b/pyanaconda/ui/gui/spokes/software.py
@@ -159,10 +159,7 @@ class SoftwareSelectionSpoke(NormalSpoke):
         hubQ.send_message(self.__class__.__name__, payloadMgr.error)
 
     def _apply(self):
-        if not self.environment:
-            return
-
-        if not (flags.automatedInstall and self.data.packages.seen):
+        if self.environment and not (flags.automatedInstall and self.data.packages.seen):
             addons = self._get_selected_addons()
             for group in addons:
                 if group not in self.selectedGroups:
--- a/pyanaconda/ui/tui/hubs/summary.py
+++ b/pyanaconda/ui/tui/hubs/summary.py
@@ -69,16 +69,20 @@ class SummaryHub(TUIHub):
         incompleteSpokes = [spoke for spoke in self._keys.values()
                                       if spoke.mandatory and not spoke.completed]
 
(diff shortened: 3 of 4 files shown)


jkonecny12:
I don't think we want this, or if we really do then we need to have something like this https://github.com/rhinstaller/anaconda/pull/472 .

The space check requires always more space than you really need and for some special use cases it's almost impossible to create even good space check. If you want really small installation you can't do it with this patch any way.
I always thought this KS behavior is on purpose.

bcl:
The problem is that we have inconsistent behavior, only available in a text kickstart. GUI install always stops, as does interactive text installations. We should make things consistent and then reevaluate #472 and the size calculation code.

jkonecny12:
OK but this should be added with the solution to disable space check or similar solution before a release.
I don't want to see f24 or RHEL7 with this patch but without an option to skip the check.

I'm giving you ACK for this one now and I'm waiting for the related later.

bcl:
New patch with another twist on this. In GUI mode if the kickstart didn't include an environment it wasn't running the software check thread so the size estimate could stay at the default of 3000MiB.

I still don't think we need a way to override this. We should depend on what information we have, and with this change I can make small / partitions that will pass the check and later fail during installation, needing a few extra megs.

jkonecny12:
> I still don't think we need a way to override this. We should depend on what information we have, and with this change I can make small / partitions that will pass the check and later fail during installation, needing a few extra megs.

Did you test it on RHEL or Fedora? RHEL is using old space check but the Rawhide and the Fedora 24 using new more strict test.

bcl:
> Did you test it on RHEL or Fedora? RHEL is using old space check but the Rawhide and the Fedora 24 using new more strict test now.

I've been testing on RHEL7.3

jkonecny12:
Yes this happening only on RHEL, but not on Fedora. On Fedora you always need min required space + some percent of space as a bonus. On Fedora you can't create small partition which will fail during installation. You even can't create small partition which will work during installation but it's practically unusable for normal purpose, but it could be useful for abnormal purpose.

Try this test on Fedora and see what is the minimal amount of space needed compared to minimal on RHEL.

Also on RHEL it could be nice to do some test with only ``@core`` or even ``--nocore`` installation.

bcl:
For now how about we limit this change to RHEL7?

jkonecny12:
I'm only thinking if you can create minimal Docker install (``--nocore`` if I'm right) but you have better understanding then me on this so I leave this to you.
In other words it's fine with me.

bcl:
Good idea. GUI works great with small (450M or so) / but TUI isn't updating the estimate.

bcl:
Ok, NOW it works consistently everywhere. I had to add checking software selection in tui kickstart using the execute() method on the software spoke. And further testing revealed a problem with GUI and network up, it tried running checkSoftwareSelection without a baseRepo setup so I added a check for that and re-tested all the variations (enough space, not enough space, nocore+nobase, minimal) for both GUI and TUI kickstarts as well as interactive.

jkonecny12:
Looks good to me again :).

bcl:
Thanks, pushed.

PR #573, opened 2016-03-30
Set CSS names on all of the anaconda classes. (#1322036)

dashea:
Without this, the name usable by CSS selectors ends up being "widget".
Set the CSS name to the type name so that CSS works right again.


jkonecny12:
Looks good to me.

dashea:
Pushed to master and f24.

PR #574, opened 2016-03-31
Fix only one address is shown in anaconda (#1264400)

jkonecny12:
Only NMIP[4,6]Config object was connected to signal on change. But when inner objects (Addresses, Nameservers, Gateway) changed it doesn't notify us about this change.
Fix by connecting this inner properties too.

Resolves: *rhbz#1264400*

Reported-by: Marek Hruscak <mhruscak@redhat.com>

Master/F24 version of https://github.com/rhinstaller/anaconda/pull/556 . On master we need to disconnect signals so more code was required here.

(cherry picked from commit b9323749c470ca3ef24bfee82f911a7b85d8f7fe)

### DIFF
--- a/pyanaconda/ui/gui/spokes/network.py
+++ b/pyanaconda/ui/gui/spokes/network.py
@@ -67,6 +67,9 @@ SECRET_AGENT_IFACE = 'org.freedesktop.NetworkManager.SecretAgent'
 AGENT_MANAGER_IFACE = 'org.freedesktop.NetworkManager.AgentManager'
 AGENT_MANAGER_PATH = "/org/freedesktop/NetworkManager/AgentManager"
 
+IPV4_CONFIG = "IPv4"
+IPV6_CONFIG = "IPv6"
+
 DEVICES_COLUMN_TITLE = 2
 DEVICES_COLUMN_OBJECT = 3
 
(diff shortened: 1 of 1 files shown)


jkonecny12:
Updated: Add comments to public functions.

pyanaconda/ui/gui/spokes/network.py
```diff
@@ -747,10 +750,36 @@ def add_device_to_list(self, device):
...
+        device.connect("notify::ip6-config", self.on_ip_obj_changed, IPV6_CONFIG)
         device.connect("state-changed", self.on_device_state_changed)
 
+    def on_ip_obj_changed(self, device, *args):
+        """Callback when ipX-config objects will be changed.
+
+        Register callback on properties (IP address, gateway...) of these ipX-config
+        objects when they are created.
+        """
+        log.debug("%s object changed", args[1])
```
rvykydal: anaconda.log is already rather cluttered with networking log messages, to mitigate this a bit I'd add "network:" prefix like in most of messages from network spoke

rvykydal:
Looks good to me.

pyanaconda/ui/gui/spokes/network.py
jkonecny12: I've missed that, thank you. I'll fix it locally.

jkonecny12:
Pushed

PR #576, opened 2016-03-31
Fix call to update optical media format. (#1322943)

dwlehman:
Goes with rhinstaller/blivet#352


bcl:
Should probably bump the required blivet version.

dwlehman:
Updated blivet requires version and manually pushed as 7f3568bf.

PR #581, opened 2016-04-05
Suppress signal handling when setting zone from location (#1322648)

dashea:
The new, improved timezone map solves some of the location confusion of
the old map (I clicked on Houston but the map says Chicago! I don't live
in Chicago! or, I clicked on my city right on the edge of a zone and
picked the wrong one!) by using a database of cities beyond just the
ones used as tzdata keys. Clicking the map will choose the nearest city,
drop the pin there, and set the zone info accordingly.

However, we managed to undo that. A click on the map in the datetime
spoke would set the location to the nearest city, which would call the
signal handler for a new location, which would update the combo boxes
with the zone for the new location, which would call the signal handler
for a change to the combo boxes, which would update the map with a
location based on the new zone.  Skip the last two steps of that.


jkonecny12:
Looks good to me.

dashea:
Pushed to master.

PR #583, opened 2016-04-06
Don't crash when trying to create a user/group that already exists.

bcl:


jkonecny12:
Looks good to me.

bcl:
Pushed

PR #584, opened 2016-04-07
Fix adding offload iSCSI devices

vojtechtrefny:
Offload iscsi devices have `node` set to `None` so we can't use it to get the port and other stuff.


dwlehman:
I would be tempted to do `if getattr(disk, 'node', None):`, but someone would probably tell me that explicit is better than implicit.

vpodzime:
Either way works for me. And both look good to me.

vojtechtrefny:
Pushed.

PR #587, opened 2016-04-12
Bring run_gui_tests.sh back

atodorov:


Makefile.am
```diff
@@ -202,6 +202,21 @@ ci: rc-release
...
+test-gui:
+	@rm -f tests/test-suite.log
+	@rm -rf tests/autogui-results-*/
+	@sudo mkdir -p $(ROOT_SITE_PACKAGES)
+	@sudo cp $(abs_builddir)/tests/usercustomize.py $(ROOT_SITE_PACKAGES)
+# We require SELinux in Permissive mode, see rhbz#1276376
+	@sudo setenforce 0
+	@sudo COVERAGE_PROCESS_START=$(abs_builddir)/.coveragerc \
+		$(abs_builddir)/tests/run_gui_tests.sh >tests/test-suite.log 2>&1
+	@sudo setenforce 1
```
clumens: I don't think you should worry about calling setenforce here and on line 211.  First off, we're already testing for it with that decorator on DogtailTestCase.  Second, if you just unconditionally "setenforce 1" here, you'll be enabling it on systems that might have been set to permissive in the first place.

atodorov: OK, I agree I might turn selinux on if it was turned off. But Dogtail tests will not run of I don't call setenforce 0. So I will simply restore whatever SELinux state was there previously.

atodorov:
Hi Chris,
objections on this one ? If not I will rebase so you can merge it without conflicts.

clumens:
I still don't think it's the test infrastructure's place to do anything with selinux.  I'd prefer that to just be a prerequisite to running tests in the first place - if you're setting up a machine to run these tests, you just need selinux in permissive mode.  If anything, I'd prefer it if we could use the same decorators to fail out/skip if selinux isn't set the way we want, just like we do for other things.

atodorov:
Hi Chris,
I've removed the bits where I modify the SELinux config. Note however that the system which will be executing these tests (aka the Jenkins box) needs to have SELinux configured properly, otherwise all the gui tests will be skipped which is pretty much the same as not executing them at all.

clumens:
Yeah that's not a problem - I've got to have selinux off for various reasons anyway.

clumens:
Merged, thanks.  I guess I need to add a jenkins job back for this, huh?  Also is there anything I need to verify is installed on the jenkins builders?

atodorov:
> I guess I need to add a jenkins job back for this, huh? 

well yes :)

> Also is there anything I need to verify is installed on the jenkins builders?

I don't think so. Everything needed should already be in install-test-requires so you should be all set.

clumens:
Alright sounds good.  I've re-enabled the jenkins job and made sure packages are installed and sudo is configured.  It should kick off overnight (for me).  I'm sure it'll fail, and tomorrow I'll figure out what I didn't set up right.

PR #588, opened 2016-04-12
network: handle null wireless AP SSID object (#1262556)

AdamWill:
With current NetworkManager, per Beniamino Galvani on the bug,
the Ssid for an AccessPoint can be null (if the AP is set not
to broadcast its SSID). Thus we could try and do
None.get_data() and blow up.

The case that definitely caused crashes is _get_strongest_
unique_aps. However it looks to me like _find_first_ap_setting
might also theoretically hit this (I don't think it's terribly
likely, but we might as well be safe).

At present _add_ap is only called after _get_strongest_unique
is done, so in theory we don't need to fix that one, but it's a
pretty generic method so probably safest to fix it too. I'm not
sure we really *need* to check both get_ssid() and get_data();
it may be that once get_ssid() has worked, get_data() always
will too. But I wasn't sure, so I played it safe.

The other cases that I didn't change are all, I think, only ever
going to be used on one of the APs in NM's list after it's been
updated, and we throw out all non-broadcast APs in _add_ap, so
I don't think we need to 'fix' those, we can safely assume that
they'll work.

Resolves: rhbz#1262556

### DIFF
--- a/pyanaconda/ui/gui/spokes/network.py
+++ b/pyanaconda/ui/gui/spokes/network.py
@@ -527,6 +527,9 @@ class NetworkControlBox(GObject.GObject):
 
     def _find_first_ap_setting(self, device, ap):
         for con in device.filter_connections(self.client.get_connections()):
+            if not con.get_setting_wireless().get_ssid():
+                # non-broadcast AP, we ignore these
+                return
             if con.get_setting_wireless().get_ssid().get_data() == ap.get_ssid().get_data():
                 return con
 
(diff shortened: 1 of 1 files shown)


AdamWill:
I tested this by just copying the modified file into place on a current F24 nightly live image before running the installer, it seems to do the trick.

rvykydal:
Looks good to me.
Thank you for the patch.

jkonecny12:
Nice patch, thank you.

bcl:
Pushed to master and f24-branch.

PR #591, opened 2016-04-14
Rename isS390 to match the renames in blivet.

dashea:


dwlehman:
Hopefully there aren't any other such mistakes in that set.

dashea:
Pushed.

PR #592, opened 2016-04-15
Add more requires to make password checking still work. (#1327411)

dashea:
Right now cracklib-dicts is still a Requires of libpwquality in f24, so right now this only needed on master.


jkonecny12:
Looks good to me.

dashea:
Pushed.

PR #594, opened 2016-04-15
Refresh metadata when updates checkbox changes (#1211907)

bcl:
When the state of a repo is changed the metadata needs to be refreshed.
To complicate matters updateBaseRepo rereads the repos from disk,
resetting the state of updates and updates-testing to enabled.

This patch adds a method to dnfpayload to control whether or not to
include updates. This is used right after the reload to set the correct
state on the repos so that the user selection won't be reset by the
upadate.


vpodzime:
Looks good to me.

bcl:
Pushed.

PR #595, opened 2016-04-19
Fix pylint errors on master

bcl:
Some are false positives, BUT some actual errors were also caught by turning E1101 errors back on in pocketlint.


jkonecny12:
Looks good to me.

bcl:
Pushed.

PR #596, opened 2016-04-19
Rhel7 network verify sources

jkonecny12:
This is basically PR https://github.com/rhinstaller/anaconda/pull/411 but with the implemented last @rvykydal's comment.

Resolves: rhbz#1270354


rvykydal:
Looks good to me.
Just a nitpick - given we set the value of the flag also when turning a device on/off, I'd name it connections_changed (or networking_changed?). I see "settings" more as a content of a connection. I mean, when I turn a device off, I am changing (active) connections, not settings.

jkonecny12:
Good point. I'm change it locally.

jkonecny12:
Pushed. Thank you.

PR #598, opened 2016-04-19
Master remove lock logging

bcl:
Remove the unused lock log level and the old yum logger target.


vpodzime:
Looks good to me.

bcl:
Pushed.

PR #600, opened 2016-04-20
Use rm -r to remove the temporary python site directory.

dashea:
Jeez, am I new here or something?


bcl:
Yes. Err, I mean ack :)

dashea:
Pushed.

PR #602, opened 2016-04-21
Master add treeinfo retry

jkonecny12:
This is reapplication of Rhel patch to master. I personally don't like this solution so if anybody has better idea I'll be happy. :)

### DIFF
--- a/pyanaconda/packaging/__init__.py
+++ b/pyanaconda/packaging/__init__.py
@@ -36,6 +36,7 @@ from fnmatch import fnmatch
 import threading
 import re
 import functools
+import time
 
 from blivet.size import Size
 from pyanaconda.iutil import requests_session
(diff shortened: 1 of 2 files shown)


vpodzime:
Cannot come up with anything better so this looks good to me.

dashea:
Jenkins, test this please

dashea:
one more try: jenkins, test this please

jkonecny12:
There were some regressions and in overall it didn't worked well. After that I have found that IPv6 url don't work in python3 requests library (bug filed).

The retry mechanism was reworked and the commit almost completely changed. Could someone do me new review after the changes? Thank you.

pyanaconda/packaging/__init__.py
```diff
@@ -374,18 +376,55 @@ def _getTreeInfo(self, url, proxy_url, sslverify):
                 log.info("Failed to parse proxy for _getTreeInfo %s: %s",
                          proxy_url, e)
 
-        response = None
+        # Retry treeinfo downloads with a progressively longer pause,
+        # so NetworkManager have a chance setup a network and we have
+        # full connectivity before trying to download things. (#1292613)
+        xdelay = xprogressive_delay()
+        response = res_1 = res_2 = None
```
vpodzime: I don't like this construct. What about having this as a list of responses? Altogether this code seems to be really complex. Maybe some of it could go into a function?

jkonecny12:
Updated.
Thank you for inspiration. I like the code much more now. :+1:

jkonecny12:
Fix one small pylint issue.

vpodzime:
Nice! Looks good to me now.

jkonecny12:
Pushed. Thank you for your help ;)

PR #603, opened 2016-04-21
Disable pylint warnings related to the log handler fixer.

dashea:
This is kind of an abuse of the inheritance system, so just disable
pylint for the parts that it isn't happy with.

Also remove ABCMeta since it isn't really getting us anything.


vpodzime:
Looks good to me.

PR #604, opened 2016-04-21
Add GRUB1 (legacy) support back to Anaconda

jaymzh:
In order to aid in installing RHEL6 from newer anacondas, provide the
ability to use legacy grub via a command line. It defaults to off.

This code was taken from 6bef2c903d1e64e087cc6a7a76dab8f6aa3a043d where it
was removed by @vathpela. After much discussion with @clumens @dcantrell and
@vathpela (and a bunch of other folks), we think this is a reasonable thing as
long as it's behind an option.

@bwann originally built this patch to add it back, I've done some trivial
fixups:

* Updated `install(self)` to `install(self, args=None)`
* Cleaned up some logs and the name of the option.

This needs some testing we will do soon.


pyanaconda/bootloader.py
```diff
@@ -2303,16 +2418,29 @@ def install(self, args=None):
...
+                              platform.MacEFI: MacEFIGRUB,
+                              platform.PPC: GRUB2,
+                              platform.IPSeriesPPC: IPSeriesGRUB2,
+                              platform.NewWorldPPC: MacYaboot,
+                              platform.S390: ZIPL,
+                              platform.Aarch64EFI: Aarch64EFIGRUB,
+                              platform.ARM: UBOOT,
+                              platform.omapARM: UBOOT}
+else:
+  bootloader_by_platform = {platform.X86: GRUB2,
```
dashea: indentation should be 4 spaces

pyanaconda/bootloader.py
```diff
@@ -2303,16 +2418,29 @@ def install(self, args=None):
...
+    log.info("Using legacy grub (0.9x)")
+    bootloader_by_platform = {platform.X86: GRUB,
+                              platform.EFI: EFIGRUB1,
+                              platform.MacEFI: MacEFIGRUB,
+                              platform.PPC: GRUB2,
+                              platform.IPSeriesPPC: IPSeriesGRUB2,
+                              platform.NewWorldPPC: MacYaboot,
+                              platform.S390: ZIPL,
+                              platform.Aarch64EFI: Aarch64EFIGRUB,
+                              platform.ARM: UBOOT,
```
dashea: "UBOOT" is not defined, and I think got pulled back in by accident

pyanaconda/bootloader.py
```diff
@@ -1677,6 +1679,119 @@ def check(self):
...
+            try:
+                (slot, _product) = line.split(None, 1)
+            except ValueError:
+                continue
+
+            if _product == productName:
+                slot_id = slot[4:8]
+                # slot_id is hex, we can't use .isint and use this regex:
+                if not re.match("^[0-9a-fA-F]+$", slot_id):
+                    log.warning("failed to parse efi boot slot (%s)" % slot)
```
dashea: I guess this came from before the great pylint change. s/ % slot/, slot/ plz

jaymzh:
OK updated for review. Will be testing today.

dashea:
test this please

dashea:
Well I guess that didn't work. Sorry for the confusing comment.

dashea:
Jenkins, this is ok to test

bwann:
fyi this is actually legacy grub for EFI, legacy grub for BIOS should still be there

pyanaconda/bootloader.py
```diff
@@ -2303,16 +2418,27 @@ def install(self, args=None):
...
-                          platform.EFI: EFIGRUB,
-                          platform.MacEFI: MacEFIGRUB,
-                          platform.PPC: GRUB2,
-                          platform.IPSeriesPPC: IPSeriesGRUB2,
-                          platform.NewWorldPPC: MacYaboot,
-                          platform.S390: ZIPL,
-                          platform.Aarch64EFI: Aarch64EFIGRUB,
-                          platform.ARM: EXTLINUX,
-                          platform.omapARM: EXTLINUX}
+if flags.cmdline.get("inst.legacygrub") == "1":
```
bwann: looks like flags.cmdline is actually command line args to anaconda (anaconda.py) and not /proc/cmdline so this doesn't work right now.

dashea: Arguments are defined in pyanaconda/anaconda_argparse.py in that big pile of add_argument calls. The help output, from all of those help_parser.help_text, is defined in data/anaconda_options.txt (or you could use help=argparse.SUPPRESS to just hide it from --help). If you add a --legacygrub option there it will also get tied to inst.legacygrub on the command line

bwann:
undeleting my earlier comment:  fyi this actually adds back legacy grub for EFI, legacy grub for BIOS is still in the code. The soon-to-be added --legacygrub flag will enable it for whichever way you boot (UEFI or BIOS)

pyanaconda/bootloader.py
```diff
@@ -1677,6 +1679,119 @@ def check(self):
...
+    #
+    # installation
+    #
+    def write(self):
+        """ Write the bootloader configuration and install the bootloader. """
+        if self.update_only:
+            self.update()
+            return
+
+        sync()
```
dashea: This is undefined, I think should be os.sync()

pyanaconda/bootloader.py
```diff
@@ -1010,6 +1009,9 @@ class GRUB(BootLoader):
 
     _serial_consoles = ["ttyS"]
 
+    def _config_dir(self):
```
bwann: @jaymzh: where did this come from? I don't remember this being in the original patch and I don't see it in master. It causes anaconda to die with a can't concatenate a string with an object (or something)

jaymzh: @bwann I was matching the updated API from the grub2 version, but I missed a thing. It needs `@property` decorator. Sorry about that.

jaymzh:
`_config_dir` is internal to the class and doesn't need to be a method, moved it back to a variable.
Updated `sync` to `os.sync`.

@bwann  tested tihs patch against anaconda23 and it worked... we need to finish getting the master building stuff that I'm working on setup to fully test master, but I feel confident in this PR now.

jaymzh:
tested against master and installing rhel6. I believe this is good to go.

bwann:
output from testing against master:  https://gist.github.com/bwann/ff75b97cdec1d55521babd1b0c0aff3b

pyanaconda/bootloader.py
```diff
@@ -2303,16 +2416,27 @@ def install(self, args=None):
...
+    bootloader_by_platform = {platform.X86: GRUB2,
+                              platform.EFI: EFIGRUB,
+                              platform.MacEFI: MacEFIGRUB,
+                              platform.PPC: GRUB2,
+                              platform.IPSeriesPPC: IPSeriesGRUB2,
+                              platform.NewWorldPPC: MacYaboot,
+                              platform.S390: ZIPL,
+                              platform.Aarch64EFI: Aarch64EFIGRUB,
+                              platform.ARM: EXTLINUX,
+                              platform.omapARM: EXTLINUX}
```
bcl: I would do this differently, setup the dict  as before and then modify the 2 entries using bootloader_by_platform.update({platform.X86: GRUB, platform.EFI: EFIGRUB1}) based on the flag.

pyanaconda/bootloader.py
```diff
@@ -1677,6 +1677,119 @@ def check(self):
 
         return ret
 
+class EFIGRUB1(GRUB):
+    packages = ["efibootmgr"]
+    can_dual_boot = False
+    _config_dir = "efi/EFI/redhat"
```
bcl: This should be modified to use the efi_dir like EFIGRUB does. That way it gets set to fedora or rhel based on the installclass.

pyanaconda/bootloader.py
jaymzh: I'm not opposed to that (it's how I would have written it) but this is just a revert of the diff that removed support for GRUB1 (with a few tweaks since the code is changed a bit). I was working to make it as close as possible to the previous code.... if the prefence is to change this in this PR, I can totally do that.

pyanaconda/bootloader.py
```diff
@@ -1677,6 +1677,119 @@ def check(self):
...
+        if self.update_only:
+            self.update()
+            return
+
+        os.sync()
+        self.stage2_device.format.sync(root=iutil.getTargetPhysicalRoot())
+        self.install()
+        self.write_config()
+
 class EFIGRUB(GRUB2):
```
bcl: I can't believe I'm going to say this. But I think we need a mixin class for EFI, we shouldn't be duplicating the code between EFIGRUB and EFIGRUB1. Take all the common parts and put it in there and then use class EFIGRUB(GRUB2, EFIMixin) or whatever the right name is for it.

pyanaconda/bootloader.py
jaymzh: will do.

pyanaconda/bootloader.py
jaymzh: OK, throwing the 'revert' right out the window then.

I actually don't think there's that much in common most of th efunctions differ slightly... but I can dig into that

jaymzh:
OK, we'll try to get this tested today or so, but in the mean time @bcl is that what you were looking for?  Code is definitely a lot cleaner... just wasn't sure how much you wanted integration like that... but I like it better.

jaymzh:
updated for s/Object/object/g

pyanaconda/bootloader.py
bcl: Yeah, at this point I don't think there's any reason to stick with reverted code, much has changed since then :)

bcl:
Looks pretty good to me, pylint complains a little though:

```
E1101(no-member):1684,31: EFIBase._config_dir: Instance of 'EFIBase' has no 'efi_dir' member
E1101(no-member):1716,46: EFIBase._add_single_efi_boot_target: Instance of 'EFIBase' has no '_efi_binary' member
E1101(no-member):1723,11: EFIBase.add_efi_boot_target: Instance of 'EFIBase' has no 'stage1_device' member
E1101(no-member):1724,45: EFIBase.add_efi_boot_target: Instance of 'EFIBase' has no 'stage1_device' member
E1101(no-member):1725,13: EFIBase.add_efi_boot_target: Instance of 'EFIBase' has no 'stage1_device' member
E1101(no-member):1726,26: EFIBase.add_efi_boot_target: Instance of 'EFIBase' has no 'stage1_device' member
E1101(no-member):1753,11: EFIBase.write: Instance of 'EFIBase' has no 'skip_bootloader' member
E1101(no-member):1756,11: EFIBase.write: Instance of 'EFIBase' has no 'update_only' member
E1101(no-member):1762,12: EFIBase.write: Instance of 'EFIBase' has no 'stage2_device' member
E1101(no-member):1765,12: EFIBase.write: Instance of 'EFIBase' has no 'write_config' member

```

I guess they should be set to a default to keep it happy.

jaymzh:
pylint is stupid, it comes from the other base class... it's a mixin. gah. I'll add defaults.

Anyway, @bwann  tested this and it works. Yay!

dwlehman:
@jaymzh it's probably better to disable the pylint checking for those lines, eg:

```python
class EFIBase(object):
    @property
    def _config_dir(self):
        return "efi/EFI/%s" % (self.efi_dir,)  # pylint: disable=no-member
```

jaymzh:
Yeah that is better. Lint silenced.

dashea:
Rebased and pushed as f251a30b26d55772752d8600973bb095bf3ff093..c53671f7ba55dca7fecc2ae95d523dfa99cf0a05

PR #605, opened 2016-04-22
Remove dnf from the list of required packages.

dashea:
This being part of the payload is leftover from the days when the
payload type was a choice, and those days are in the past. dnf is part
of @core now, so just let it live there.


jaymzh:
:+1: 

(this lets you install yum-based OSes, which is deliciously ironic)

reallylongword:
Jenkins, please test this

dashea:
(that was me making sure rando commenters can't tell jenkins what to do, in case anyone is wondering who the rando commenter is)

jkonecny12:
Looks good to me.

PR #606, opened 2016-04-25
Fix bad addon handling when addon import failed

jkonecny12:
An addon is added to AddonRegister even when the import of addon failed. This happens because pykickstart is caling method ``headerHandler()`` which thinks that addon wasn't add yet so the placeholder is add instead. Because of the placeholder the ``setup()`` and ``execute()`` methods of AddonData are called and in TUI install even the
*"Executing {AddonName} addon"* is printed.

Adding more warning messages when this happen and removing the placeholder when ``setup()`` method is called.

### DIFF
--- a/pyanaconda/addons.py
+++ b/pyanaconda/addons.py
@@ -31,6 +31,8 @@ from pyanaconda.i18n import N_
 
 log = logging.getLogger("anaconda")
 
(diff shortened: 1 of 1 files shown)


M4rtinK:
Looks good to me, ACK! :)

jkonecny12:
Pushed

PR #607, opened 2016-04-25
Use Blivet 2.0 for set_default_fstype

sgallagher:
If an InstallClass sets the DefaultFSType variable, this code path
gets executed and crashes anaconda because the method was renamed
in Blivet 2.0.

See: https://bugzilla.redhat.com/show_bug.cgi?id=1329019


dwlehman:
Looks good to me. Thanks for nothing, pylint.

PR #608, opened 2016-04-26
Remove detach-client from tmux.conf

dashea:
Whatever this was supposed to do, it's not doing it anymore.


jkonecny12:
I was wondering what changed or if this is another regression of tmux. I personally using it on daily basis and they have some minor one in later version but it's not related.
I don't see anything in the changelog so I guess this is a tmux regression.

So or so this looks good to me.

PR #609, opened 2016-04-26
Use systemctl --root instead of running systemctl in a chroot

dashea:


M4rtinK:
Looks good to me.

jkonecny12:
Nice work.

dashea:
Pushed.

AdamWill:
There's a big problem with this: it won't work for enabling non-systemd services, because when you do that, it redirects to `/usr/lib/systemd/systemd-sysv-install`, which in Fedora is a symlink to `/sbin/chkconfig` and - well, first of all, both of those files is stripped from the installer environment by lorax, so we'd have to put them back. But more importantly, `chkconfig` doesn't accept `--root`:

    [adamw@adam etc]$ systemctl --root=/tmp/fakesys enable network
    network.service is not a native service, redirecting to systemd-sysv-install.
    Executing: /usr/lib/systemd/systemd-sysv-install --root=/tmp/fakesys enable network
    --root=/tmp/fakesys: unknown option

See https://bugzilla.redhat.com/show_bug.cgi?id=1369794 .

PR #610, opened 2016-04-27
F24 ostree dirinstall

bcl:
These patches allow --dirinstall (and --dirinstall inside a mock) to be used for --make-pxe-live and --make-ostree-live

### DIFF
--- a/pyanaconda/packaging/__init__.py
+++ b/pyanaconda/packaging/__init__.py
@@ -636,29 +636,11 @@ class Payload(object):
            every payload except for dnf.  Payloads should only implement one of
            these methods by overriding the unneeded one with a pass.
         """
-        if not flags.dirInstall:
-            if iutil.getSysroot() != iutil.getTargetPhysicalRoot():
-                setSysroot(iutil.getTargetPhysicalRoot(), iutil.getSysroot())
-
-                # Now that we have the FS layout in the target, umount
-                # things that were in the legacy sysroot, and put them in
-                # the target root, except for the physical /.  First,
-                # unmount all target filesystems.
-                self.storage.umountFilesystems()
-
-                # Explicitly mount the root on the physical sysroot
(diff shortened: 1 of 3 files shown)


jkonecny12:
Seems good to me.

pyanaconda/packaging/__init__.py
```diff
@@ -636,29 +636,11 @@ def writeStorageLate(self):
...
-                # since the bootloader code will expect to find /boot
-                # inside the chroot.
-                self.storage.mountFilesystems(skipRoot=True)
-
-            self.storage.write()
+        if iutil.getSysroot() != iutil.getTargetPhysicalRoot():
+            setSysroot(iutil.getTargetPhysicalRoot(), iutil.getSysroot())
+            self.prepareMountTargets(self.storage)
+            if not flags.dirInstall:
+                self.storage.write()
```
rvykydal: I don't know if we should care but for
flags.dirInstall == False, iutil.getSysroot() == iutil.getTargetPhysicalRoot() we do not call self.storage.write() anymore with the patch.

bcl: I don't think it matters since a dirinstall isn't going to have anything managed by storage.

rvykydal:
My point was non-dirinstall case, sysroot == physical root.
For example live install - don't we need to call self.storage.write()?
(this may apply to tarinstall as well, for rpmostreepayload sysroot != physical root, for dnf storage is written early).

bcl:
Ah, I see now. For all other uses the storage.write() isn't done at this point, it is handled by writeStorageEarly. ostree is the only one that has sysroot != physical root and does the storage.write late instead of early.

bcl:
Pushed to master only. Too much change for post-freeze F24. And FWIW @rvykydal was correct, storage.write did still need to be called because of live installations. It was fixed by commit 4417eb1225b122a1cdeffa001646c74c6c46fcdc

PR #611, opened 2016-04-27
Ignore failure when disable services that do not exist

jaymzh:


dwlehman:
Looks good to me.

bwann:
Looks like this did the trick for installing RHEL6:

...

20:40:54,819 INFO anaconda: Running Thread: AnaConfigurationThread (140589322692352)
20:40:54,822 INFO anaconda: Configuring installed system
20:40:55,360 WARN anaconda: Disabling chronyd failed. It probably doesn't exist
20:40:55,544 WARN anaconda: there is no /etc/systemd/system directory, cannot update default.target!
20:40:55,545 WARN anaconda: there is no /etc/systemd/system directory, cannot update default.target!
20:40:55,546 INFO anaconda: Configuring installed system -- DONE
...

dwlehman:
Cherry-picked and pushed as b35fe094.

PR #613, opened 2016-04-29
Check for mounted partitions as part of sanity_check (#1330820)

bcl:
Sometimes partitions end up being mounted, especially during live
installations, so add a check to make sure none of the selected disks
have mounted partitions.


jkonecny12:
Looks good to me.

bcl:

PR #614, opened 2016-05-02
Implement %packages --excludeWeakdeps (#1331100)

dashea:


dashea:
I have a kickstart test, https://github.com/dashea/kickstart-tests/tree/exclude-weakdeps, which I'll turn into a real PR once all the pieces are in place. I got SUCCESS running it locally with a lovingly hand-crafted boot.iso.

pyanaconda/packaging/dnfpayload.py
```diff
@@ -448,6 +448,12 @@ def _configure(self):
 
         conf.reposdir = REPO_DIRS
 
+        # Two reasons to turn this off:
+        # 1. Minimal installs don't want all the extras this brings in.
+        # 2. Installs aren't reproducible due to weak deps. failing silently.
```
jkonecny12: I think these are two sentences so it would be nice to capitalize the **F**ailing word.

jkonecny12:
Other than the small issue above, this looks good to me.

james-antill:
FWIW it's just one sentence ending (unshortened) "weak dependencies failing silently"

dashea:

PR #615, opened 2016-05-03
Fix issues with glib schemas

dashea:


dashea:
Only the first commit is needed on master, the enums thing is already there.

M4rtinK:
Looks good to me - ACK! And thanks for tracking it down! :)

dashea:
Pushed.

PR #616, opened 2016-05-05
Look higher for the combobox associated with an entry (#1333530)

dashea:
Fun gtk fact: when creating a GtkComboBox with an associated entry,
gtk_bin_get_child() on the combobox returns the GtkEntry.
gtk_widget_get_parent() on the entry returns a GtkBox.


vpodzime:
I really think we need a "WTFACK" label in GitHub for cases like this. For now... WTF? ACK.

dashea:
Pushed to f24 and master.

PR #618, opened 2016-05-06
Use createrepo_c in the ci target.

dashea:
createrepo is old and busted, and its replacement has just about the
dumbest possible name. Use that one.


vpodzime:
Looks good to me.

dashea:
Pushed.

PR #619, opened 2016-05-06
Remove on_proxy_ok_clicked from Source spoke

jkonecny12:
This method is not used anywhere in the code now.


vpodzime:
Did you find the place where the things from this method happen? If there's no such place in the code, we should hook it back somewhere instead.

jkonecny12:
I'll look on that. I really don't know I found this by accident.

jkonecny12:
Results from my investigation:

*This method is a leftover from commit:*

*"Make dialogs behave better with timed input validation"*
*SHA bb1c211*

vpodzime:
Yeah, it looks like it can go. Thanks for the investigation. ;)

jkonecny12:
Pushed

PR #620, opened 2016-05-06
New canary tests

dashea:
This removes the tests on translated strings from make check and ci, and moves the release-mode check (the one that removes busted translations) to po-pull. This way errors that get caught by msgfmt are caught early and dealt with instead of being caught during the build (since we can't remove that pesky -c from the gettext makefiles).


jkonecny12:
Could you please change the first commit. Squashed information is totally unhelpful for anyone who doesn't have these commits. Also to know there were some merge from commit I don't have to commit I don't have also is only confusing.

If you rename it to small changes or similar and remove the unavailable SHAs it will be more readable and less confusing thought.

dashea:
When first adding translation-canary to projects I tried doing it without --squash, so that every commit in translation-canary would be added to anaconda and merged into a subtree. I felt like it was too much information. What really matters to anacoda is "added thing to subtree" or "synced to latest version of thing in subtree", and having each commit as part of that seemed like a lot of noise, especially when copied across every rhinstaller project.

The upstream SHAs are necessary for subtree split/merge to work.

jkonecny12:
OK then it's fine with me. Only one small thing, please fix the missing newline.

dashea:
Updated to fix the spacing issues.

jkonecny12:
It looks even better now. Thank you :+1:

PR #621, opened 2016-05-06
Lower required squashfs ram to 350M (#1333113)

bcl:
The squashfs.img is smaller than 750M, lower the required ram so that
the installation will run on lower memory systems.

This means the minimum for a text mode PXE or http stage2 install is
going to be 720M instead of 1070M

Resolves: rhbz#1333113


vpodzime:
Looks good to me. And an installation with 720MiB of RAM seems to keep moving on just quite slowly.

vpodzime:
Or not. The anaconda-yum process was killed in the end.

bcl:
> Or not. The anaconda-yum process was killed in the end.

Weird, how were you testing it? I did a minimal text PXE install with 720M and it worked just fine.

vpodzime:
> Weird, how were you testing it? I did a minimal text PXE install with 720M and it worked just fine.

Oh, I was trying a graphical installation. Let me try this again.

vpodzime:
Looks like this works just fine. Sorry for the noise. I just have the last question - how did you get to the number 720 MiB?

bcl:
> Looks like this works just fine. Sorry for the noise. I just have the last question - how did you get to the number 720 MiB?

Wild guess, nothing more.

bcl:
Pushed.

PR #622, opened 2016-05-09
Single language mode

M4rtinK:
This pull request adds a single language mode to Anaconda - in this mode no interactive language configuration is possible during the installation. Language can still be set via the _inst.lang_ boot option or the _lang_ kickstart command, but can't be changed during the installation. If no language is set aanconda will use the default language (en_US).

Single language mode can be enabled by the _inst.singlelang_ boot option and is enabled by default for all Atomic installations.

### DIFF
--- a/pyanaconda/ui/gui/spokes/keyboard.py
+++ b/pyanaconda/ui/gui/spokes/keyboard.py
@@ -322,6 +322,12 @@ class KeyboardSpoke(NormalSpoke):
 
     def initialize(self):
         NormalSpoke.initialize(self)
+
+        # set X keyboard defaults
+        # - this needs to be done early in spoke initialization so that
+        #   the spoke status does not show outdated keyboard selection
+        keyboard.set_x_keyboard_defaults(self.data, self._xkl_wrapper)
+
         self._add_dialog = AddLayoutDialog(self.data)
         self._add_dialog.initialize()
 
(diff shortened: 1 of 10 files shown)


rvykydal:
Looks good to me.

vpodzime:
The first commit shouldn't have the bug ID on the first line.

pyanaconda/ui/gui/spokes/keyboard.py
```diff
@@ -297,6 +297,9 @@ def apply(self):
 
     @property
     def completed(self):
+        # make sure the x_layouts list has at least one keyboard layout
+        if not self.data.keyboard.x_layouts:
+            self.data.keyboard.x_layouts.append(DEFAULT_KEYBOARD)
```
vpodzime: This really doesn't belong to the ``completed`` property.

atodorov:
@M4rtinK does this patch change the UI in the default case where multiple languages are available ? @clumens and I are working on integrating tests/gui in Jenkins and almost have it up and running. I'm wondering if this PR will break things or not.

M4rtinK:
@atodorov It should not change the default case in any way - if it does its a regression and should be fixed.

pyanaconda/ui/gui/spokes/keyboard.py
M4rtinK: Indeed! Works as well from `initialize()`.

M4rtinK:
@vpodzime Fixed the commit message.

M4rtinK:
Pushed.

PR #623, opened 2016-05-09
Replace IPv6 and URL regexes by python3 urllib

jkonecny12:
This is one way how we can remove part of our responsibility and leave this to python standard library. The problem here is that something like invalid URL doesn't exists in urllib solution. From my information even most of the DNS servers don't do such a check as validation of an URL because almost everything is valid and it's hard to find something which is not.

So our test of invalid URLs won't work and we can test only something like there is no hostname or there is no scheme.

Do we want this solution instead of that big ugly REG-EX or it will be better to leave it how it is now, what do you think?


bcl:
I think the tests that were removed should be rewritten to test the new/modified methods. There were a large number of patterns there that demonstrate the kinds of input we expect to be able to handle.

jkonecny12:
I'm closing this old PR. Can be finished later, however there is no time for it now.

PR #624, opened 2016-05-09
Do not override StorageChecker.errors in StorageSpoke (#1269195)

bcl:
When assigning checkDiskSelection result to self.errors, we can
lose errors from CustomPartitioningSpoke. Added separate "category"
for disk selection errors to fix this.

Signed-off-by: Vojtech Trefny <vtrefny@redhat.com>
(cherry picked from commit aec2c6f55c6970ef76cf21d4d1587f543330af17)

Resolves: rhbz#1269195


vpodzime:
Looks good to me.

bcl:
Pushed

PR #626, opened 2016-05-10
Move yum.cache to disk after partitioning (#1287438)

bcl:
In some installations the size of yum.cache can exceed that of the
tmpfs used for /tmp/yum.cache or even that of the available RAM so move
it to the disk after partitioning has finished.

Resolves: rhbz#1287438

### DIFF
--- a/pyanaconda/packaging/yumpayload.py
+++ b/pyanaconda/packaging/yumpayload.py
@@ -207,7 +207,7 @@ class YumPayload(PackagePayload):
         super(YumPayload, self).unsetup()
         self._setup = False
 
-    def _resetYum(self, root=None, keep_cache=False, releasever=None):
+    def _resetYum(self, root=None, keep_cache=False, releasever=None, cache_dir=_yum_cache_dir):
         """ Delete and recreate the payload's YumBase instance.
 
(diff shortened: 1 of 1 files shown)


jkonecny12:
Looks good to me.

pyanaconda/packaging/yumpayload.py
```diff
@@ -400,12 +400,17 @@ def _writeInstallConfig(self):
             except PayloadSetupError as e:
                 log.error(e)
 
-        releasever = self._yum.conf.yumvar['releasever']
-        self._writeYumConfig()
+        # Move the yum.cache to the new disk space to save memory
+        new_cache = iutil.getSysroot()+"/var/tmp/yum.cache"
+        os.makedirs(iutil.getSysroot()+"/var/tmp")
+        if execWithRedirect("mv", [_yum_cache_dir, new_cache]) > 0:
```
vpodzime: Any reason why not to use ``shutil.move()`` here?

bcl:
> Any reason why not to use shutil.move() here?

Good idea.

bcl:
Pushed.

PR #627, opened 2016-05-10
Remove the locale list from zanata.xml

dashea:
Zanata now manages this information on the server side and no longer
includes it in the exported config file. Also, if the information on the
server gets out of sync with the information in zanata.xml it causes a
really confusing 403 error.


dashea:
Added a patch to avoid an out-of-sync LINGUAS by getting out of the LINGUAS business.

And for what it's worth the test failure is because of the weak deps thing I pushed earlier, since that version of pykickstart hasn't hit dl.fp.o yet.

jkonecny12:
Looks good to me.

vpodzime:
Looks good to me. I hope you don't mind if I do the same in libbytesize. :)

dashea:
No good, need to take a different approach with LINGUAS.

PR #628, opened 2016-05-10
Remove the locales list from zanata.xml

dashea:
Zanata stores this information on the server and no longer exports it as
part of the project config. Also, if the information in zanata.xml gets
out of sync with the server, po-pull crashes with a confusing 403 error.

Run po-empty after po-pull so that if LINGUAS does not match the list of
enabled languages, we can ignore it.

Related: rhbz#1304327


dashea:
New patch based on the second commit in https://github.com/rhinstaller/anaconda/pull/627, without the sr@latin bit since we don't do that on this branch.

jkonecny12:
Looks good to me.

dashea:
No good, the gettext autoconf stuff reads po/LINGUAS during ./configure, so doing it after po-pull is too late.

PR #629, opened 2016-05-13
Deselect all addons correctly (#1333505)

bcl:
If depsolve failed with some (or all) addons selected it wasn't cleared
by deselecting all of them because the package and group set were never
cleared.

Add a check to skip this if the addons have not been changed (or never
set, as is the case with kickstart).


vpodzime:
Looks good to me.

bcl:

PR #630, opened 2016-05-16
translation changes

dashea:
Clean up the way we handle translations. No longer do we need a million files in the po directory getting out of sync with reality, and no longer do we need intltool doing part of gettext's job but worse.

Also remove the locales list from zanata.xml like I wanted to last week before I got stuck on LINGUAS getting messed up.


dashea:
Ha, whoops. I left a TODO comment in there. Fixed that and re-pushed.

dashea:
Removed the 'git checkout -- anaconda.pot' parts of the Makefile I missed.

bcl:
Very nice!

dashea:
```+ /usr/lib/rpm/find-lang.sh /builddir/build/BUILDROOT/anaconda-25.13-0.1.20160516232035.fc25.x86_64 anaconda
No translations found for anaconda in /builddir/build/BUILDROOT/anaconda-25.13-0.1.20160516232035.fc25.x86_64
RPM build errors:
error: Bad exit status from /var/tmp/rpm-tmp.lcSbw7 (%install)
    Bad exit status from /var/tmp/rpm-tmp.lcSbw7 (%install)
Child return code was: 1
```
Damn, need to find a way around that. Maybe we still do need a po-empty of some kind.

dashea:
Found the top secret rpm macro that lets you plow right past that error, trying again.

dashea:
Pushed one more tweak to dist-hook, so that the conditional is actually a conditional instead of trying to use "*.po" as a filename. Ugh, shell.

It probably doesn't matter either way, but my goal with that part was to avoid having a pocketlint dependency for creating language-less source tarballs.

bcl:
Good catch.

dashea:
Pushed.

PR #631, opened 2016-05-16
Fix UEFI installation after EFIBase refactor

bcl:
Python class search is left to right and depth first. So it was using
the GRUB/GRUB2 methods before the EFIBase methods.


vpodzime:
Looks good to me.

bcl:
Pushed.

PR #633, opened 2016-05-17
Fix writeStorageLate for live installations (#1334019)

bcl:
Ends up this is used by live, so the storage.write() call should be called.


snbueno:
Looks fine to me.

bcl:
Pushed

PR #634, opened 2016-05-18
network: don't crash on devices with zero MAC address (#1334632)

rvykydal:
Resolves: rhbz#1334632

Make handling of ifname= boot option more robust. In this case don't crash
(regardless of ifname= being used or no) if the device does not have permanent
MAC address.

Also add support for multiple ifname= options

And make one log message meaningful.

Details of infame= handling in anaconda:

When ifname=<DEVNAME>:<MAC> is used we bind DEVNAME to MAC in ifcfg file
when we create the ifcfg file in Anaconda.

We create the ifcfg in Anaconda if it was not created in dracut, ie if the
device was not activated by dracut. (Which means that the ifname= is not
applied by dracut, so this case actually occurs only when we just want to bind
the <MAC> to its <DEVNAME> without renaming it it dracut).  When creating ifcfg
file for <DEVNAME> we check that <MAC> from ifname= corresponds to actual mac
address of the device and if so we bind the <DEVNAME> to the <MAC> by writing
HWADDR value (additionally to DEVICE) into ifcfg file.

### DIFF
--- a/pyanaconda/constants.py
+++ b/pyanaconda/constants.py
@@ -161,7 +161,7 @@ SCREENSHOTS_DIRECTORY = "/tmp/anaconda-screenshots"
 SCREENSHOTS_TARGET_DIRECTORY = "/root/anaconda-screenshots"
 
 # cmdline arguments that append instead of overwrite
-CMDLINE_APPEND = ["modprobe.blacklist"]
+CMDLINE_APPEND = ["modprobe.blacklist", "ifname"]
 
 DEFAULT_AUTOPART_TYPE = AUTOPART_TYPE_LVM
 
--- a/pyanaconda/network.py
+++ b/pyanaconda/network.py
@@ -534,10 +534,11 @@ def add_connection_for_ksdata(networkdata, devname):
         dev_spec = None
     # type "802-3-ethernet"
     else:
-        mac = nm.nm_device_perm_hwaddress(devname)
-        if flags.cmdline.get("ifname", "").upper() == "{0}:{1}".format(devname, mac).upper():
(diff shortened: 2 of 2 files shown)


bcl:
Looks, good, could you also add the new behavior to the docs/boot-options.txt file?

vpodzime:
> Details of infame= handling in anaconda:

Typo.

vpodzime:
> when we just want to bind
> the to its without renaming it it dracut

?

vpodzime:
There seem to be a couple of words missing in the last paragraph. :)

vpodzime:
Looks good to me otherwise.

rvykydal:
@vpodizime:
1) I'll replace ifname= with ifname in the commit message
2) As for the missing words, please look at the commit message, the generated comment seems to drop strings enclosed in "<" ">"

rvykydal:
@bcl 
How about adding this note:
[NOTE]
==============================================================================
Dracut applies ifname option (which might involve renaming the device with 
given MAC) in initramfs only if the device is activated in initramfs stage
(based on ip= option). If it is not the case, installer still binds the
current device name to the MAC by adding HWADDR setting to the ifcfg file of
the device. 
==============================================================================

vpodzime:
> @vpodizime:
> 1) I'll replace ifname= with ifname in the commit message
> 2) As for the missing words, please look at the commit message, the generated comment seems to drop strings enclosed in "<" ">"

Wonderful. :) Thanks for the explanation.

bcl:
@rvykydal yeah, that looks good to me.

rvykydal:
Pushed

PR #635, opened 2016-05-18
Clean up yelp processes (#1282432)

dashea:


M4rtinK:
Good catch - it really [might not terminate correctly otherwise](https://docs.python.org/3/library/subprocess.html#subprocess.Popen.wait). :) Looks good to me, ACK!

PR #636, opened 2016-05-19
Rhel7 branch dont configure hostname with current

rvykydal:
The change affects the installations where user doesn't configure (static) hostname of target system [1]. In this case we used to set it (in /etc/hostname) to the name we found using getHostname function (ie from dhcp or DNS) in Anaconda [2].  As a result, if installed system was moved to another network environment, or had another transient hostname assigned by network configuration (NM) in general, the transient hostname was not applied. As a workaround users used to write localhost.localdomain in /etc/hostname in %post scripts.

With this pachset we set static hostname of target system only if explicitly configured by user (PATCH 1/5). If it is not, it has default "localhost.localdomain" value (which was already the case when dhcp or dns did not suggest any hostname).

To mitigate this change of behavior:

- Display current environment hostname in network spoke (PATCH 1/4). Formerly the current hostname was showed as value of the "Host name" entry because the ksdata hostname value had been set from it earlier, so it was visible information that could be missed now.

- Try to keep former behaviour of suggesting names for storage containers by offering current hostname in case static hostname was not configured (PATCH 2/4)

- Don't set current environment hostname from target system static hostname (the "Host name" entry) in network spoke's execute() method (eg when leaving the spoke) anymore. Especially don't set it to the default "localhost.localdomain" value which is in "Host name" entry if user does not configure target system hostname. This case was not an issue formerly because the target system hostname was initialized to current hostname of the system (hostname suggested by getHostname) during networking initialization. Setting the environment hostname from the "Host name" entry in execute() was actually added for initial-setup. To keep this functionality I am adding "Apply now" checkbox to the hostname entryi [3]. (PATCH 3/4)

- To prevent fallouts from changed behaviour, we keep writing out /etc/hostname even when static hostname is not configured - using value "localhost.localdomain". In this case target system still applies transient hostname (same as if we didn't write /etc/hostname). I bet there is something relying on /etc/hostname existing after installation.

[1] via network --hostname or "Host name" entry of network spoke.
[2] This hostname corresponds to the transient ("transient" as in man hostnamect) hostname of the installer environment set by NetworkManager during installation. To make things a bit more complicated we didn't actually read current system hostname set by NM, but we were using the getHostname function to suggest (and actually configure) the target system hostname.
[3] https://rvykydal.fedorapeople.org/apply_hostname_in_networ_spoke.webm - note this feature is to be used in initial-setup. I hope it is not disturbing too much in installer case.

### DIFF
--- a/pyanaconda/network.py
+++ b/pyanaconda/network.py
@@ -173,33 +173,6 @@ def prefix2netmask(prefix):
 def current_hostname():
     return socket.gethostname()
 
-# Try to determine what the hostname should be for this system
-def getHostname():
-
-    hn = None
-
-    # First address (we prefer ipv4) of last device (as it used to be) wins
-    for dev in nm.nm_activated_devices():
-        addrs = (nm.nm_device_ip_addresses(dev, version=4) +
-                 nm.nm_device_ip_addresses(dev, version=6))
-        for ipaddr in addrs:
-            try:
-                hinfo = socket.gethostbyaddr(ipaddr)
-            except socket.herror as e:
-                log.debug("Exception caught trying to get host name of %s: %s", ipaddr, e)
-            else:
-                if len(hinfo) == 3:
-                    hn = hinfo[0]
-                    break
-
-    if not hn or hn in ('(none)', 'localhost', 'localhost.localdomain'):
-        hn = socket.gethostname()
-
-    if not hn or hn in ('(none)', 'localhost', 'localhost.localdomain'):
-        hn = DEFAULT_HOSTNAME
-
-    return hn
-
 def logIfcfgFile(path, message=""):
     content = ""
     if os.access(path, os.R_OK):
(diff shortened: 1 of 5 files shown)


M4rtinK:
>To keep this functionality I am adding "Apply now" checkbox to the hostname entryi [3]. (PATCH 3/4)
I think an "apply now" button would be better - the user would see the change at once.

M4rtinK:
Other than the remark about the "apply now button" this all looks good to me, thanks & ACK! :)

rvykydal:
I agree with the button + immediate change (vs checkbox + change on Done). I'll update the patch.

rvykydal:
Also, as our automatic checks revealed, I shouldn't be removing the getHostname function as it is used in anaconda file. I'll make new PR.

rvykydal:
New PR: https://github.com/rhinstaller/anaconda/pull/643

rvykydal:
Closing

PR #637, opened 2016-05-20
Fix some residual issues of dasdfmt in text-mode. (#1259437)

snbueno:
Some of these are dumb errors. For example, need to make sure we pass
a list of device names, not device objects, to the methods that determine
whether a DASD needs to be formatted.

devicetree.populate() needs to be called so that the devicetree is
actually made aware of the changes.

Additionally, only call storageInitialize and self._initialize() after
all of this is we're doing a manual installation. This makes sure things
like the summary screen get updated with accurate information about any
newly-formatted DASDs.

Resolves: rhbz#1259437

### DIFF
--- a/pyanaconda/ui/tui/spokes/storage.py
+++ b/pyanaconda/ui/tui/spokes/storage.py
@@ -324,10 +324,10 @@ class StorageSpoke(NormalTUISpoke):
 
             if self.data.zerombr.zerombr:
                 # unformatted DASDs
-                unformatted += make_unformatted_dasd_list([d for d in self.data.ignoredisk.onlyuse])
+                unformatted += make_unformatted_dasd_list([d.name for d in getDisks(self.storage.devicetree)])
             if self.data.clearpart.cdl:
                 # LDL DASDs
-                ldl += [d for d in self.data.ignoredisk.onlyuse if is_ldl_dasd(d)]
+                ldl += [d for d in self.storage.devicetree.dasd if is_ldl_dasd(d.name)]
             # combine into one nice list
             to_format = list(set(unformatted + ldl))
         else:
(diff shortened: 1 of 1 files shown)


vpodzime:
> Additionally, only call storageInitialize and self._initialize() after
> all of this is we're doing a manual installation. This makes sure things

...all of this **if** we're doing... ?

pyanaconda/ui/tui/spokes/storage.py
```diff
@@ -324,10 +324,10 @@ def run_dasdfmt(self, to_format=None):
 
             if self.data.zerombr.zerombr:
                 # unformatted DASDs
-                unformatted += make_unformatted_dasd_list([d for d in self.data.ignoredisk.onlyuse])
+                unformatted += make_unformatted_dasd_list([d.name for d in getDisks(self.storage.devicetree)])
```
vpodzime: ``getDisks()`` gives a different list of devices so I think this change should be mentioned in the commit message.

pyanaconda/ui/tui/spokes/storage.py
```diff
@@ -361,9 +361,13 @@ def run_dasdfmt(self, to_format=None):
...
-        threadMgr.add(AnacondaThread(name=THREAD_STORAGE, target=storageInitialize,
-                                     args=(self.storage, self.data, self.storage.devicetree.protectedDevNames)))
-        self._initialize()
+        self.storage.devicetree.populate()
+        if not flags.automatedInstall:
+            # reinit storage
+            threadMgr.add(AnacondaThread(name=THREAD_STORAGE, target=storageInitialize,
+                                         args=(self.storage, self.data, self.storage.devicetree.protectedDevNames)))
+            # update the summary screen with the changes
+            self._initialize()
```
vpodzime: Does this need for the thread to finish or not? If not, I don't really see a point of running the thread and running it only in manual installations.

snbueno: It's necessary to wait on this thread in manual installations because otherwise the screen won't refresh with the updated config. (Obviously, it's not necessary in ks, since you don't see the storage spoke.)

pyanaconda/ui/tui/spokes/storage.py
snbueno: Will do.

pyanaconda/ui/tui/spokes/storage.py
vpodzime: Makes sense, thanks for the extra explanation.

vpodzime:
Looks good to me now.

snbueno:
Pushed.

PR #638, opened 2016-05-20
Fix TUI ErrorDialog processing (#1337427)

bcl:
There were a couple problems here. First was that the kickstart spoke
ready loop didn't process any events, so the dialog would never be
shown.

Second is that the dialog should do what it says and actually exit when
you hit enter.

The exit is placed in the input handler because we want to make sure
there is no further processing done. If the screen is allowed to close
and return to the mainloop you end up catching the traceback that
triggered this in other places, and falling into python-meh handling of
the error you wanted to use a nice dialog for.


vpodzime:
Looks good to me.

bcl:
pushed

PR #640, opened 2016-05-21
Catch DNF MarkingError during group installation (#1337731)

bcl:
dnf-1.1.9 has started raising an error when packages in a group are
missing. This catches the error, displays a dialog and exits the
installation.

### DIFF
--- a/pyanaconda/packaging/dnfpayload.py
+++ b/pyanaconda/packaging/dnfpayload.py
@@ -511,7 +511,10 @@ class DNFPayload(packaging.PackagePayload):
             types.add('optional')
         exclude = self.data.packages.excludedList
         try:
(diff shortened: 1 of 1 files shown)


vpodzime:
Looks good to me.

bcl:
Pushed to master. There's still some debate as to who what and how to handle this, but at least this shows what's wrong instead of crashing.

bcl:
Added strict=False to work around this in F24. For rawhide I think there needs to be a better solution, we shouldn't be ignoring missing packages within groups.

vpodzime:
> Added strict=False to work around this in F24. For rawhide I think there needs to be a better solution, we shouldn't be ignoring missing packages within groups.

Sounds reasonable to me.

PR #644, opened 2016-05-24
RHEL7 PW policy

vpodzime:
These are the ports of respective patches from the *master* branch which together implement the support for the ``pwpolicy`` kickstart command in the ``%anaconda`` section for restricting the password quality policies. This could be useful for administrators, people creating specific images/kickstarts. The OSCAP addon can translate requirements from the SCAP policy and require it to be followed in the installation by the means for the ``pwpolicy`` command.


bcl:
Looks good to me.

vpodzime:
Pushed.

PR #645, opened 2016-05-26
NFS DDs installation now works correctly (#1269915)

japokorn:
Only specified RPM file is now installed when specified by inst.dd and using NFS.
Added a feature that enables to install all RPMs in directory when using NFS.
(e.g. inst.dd=nfs://server/path/directory_with_rpms)

Resolves: rhbz#1269915

### DIFF
--- a/utils/dd/dd_list.c
+++ b/utils/dd/dd_list.c
@@ -171,10 +171,15 @@ int main(int argc, char *argv[])
     int option;
     int option_index;
 
-    char *directory = NULL;
+    char *path = NULL;
+    int path_len;
     int verbose = 0;
 
     struct _version_struct versions = {NULL, NULL};
+
+    char *globpattern;
+    glob_t globres;
+    char** globitem;
 
     while ((option = getopt_long(argc, argv, shortopts, longopts, &option_index)) != -1) {
         switch (option) {
@@ -183,7 +188,7 @@ int main(int argc, char *argv[])
             break;
 
         case 'd':
-            directory = strdup(optarg);
+            path = strdup(optarg);
             break;
 
         case 'k':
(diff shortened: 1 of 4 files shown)


utils/dd/dd_list.c
```diff
@@ -206,38 +206,55 @@ int main(int argc, char *argv[])
...
-        printf("Listing DD dir %s\n", directory);
-    }
-
     init_rpm();
 
-    char *globpattern;
-    checked_asprintf(&globpattern, "%s/*.rpm", directory);
+    /* determine if path is rpm filename or directory
+     * (paths ending with "/.rpm" will be still treated as directories) */
+    if((strlen(path) > 5) && (!strcmp(path + strlen(path) - 4, ".rpm")) &&
```
vpodzime: Please use ``strcmp(...) == 0``, the ``!strcmp(...)`` expression is confusing.

utils/dd/dd_list.c
```diff
@@ -206,38 +206,55 @@ int main(int argc, char *argv[])
...
+            (*(path + strlen(path) - 5) != '/')) {
```
vpodzime: It would make sense to compute the length of the ``path`` string once and then use the computed value.

utils/dd/dd_list.c
```diff
@@ -206,38 +206,55 @@ int main(int argc, char *argv[])
...
-    if (verbose) {
-        printf("Listing DD dir %s\n", directory);
-    }
-
     init_rpm();
 
-    char *globpattern;
-    checked_asprintf(&globpattern, "%s/*.rpm", directory);
+    /* determine if path is rpm filename or directory
+     * (paths ending with "/.rpm" will be still treated as directories) */
```
vpodzime: Why do we need to support crazy things like paths ending with ``/.rpm``? Maybe the comment could be a little bit more verbose?

utils/dd/dd_list.c
```diff
@@ -206,38 +206,55 @@ int main(int argc, char *argv[])
...
+
+        char *globpattern;
+
+        /* look for all .rpm files in a given direcory */
+        checked_asprintf(&globpattern, "%s/*.rpm", path);
 
-    glob_t globres;
-    char** globitem;
+        glob_t globres;
+        char** globitem;
```
vpodzime: Please move all the variables to the beginning of the function.

utils/dd/dd_list.c
```diff
@@ -206,38 +206,55 @@ int main(int argc, char *argv[])
...
-    char *globpattern;
-    checked_asprintf(&globpattern, "%s/*.rpm", directory);
+    /* determine if path is rpm filename or directory
+     * (paths ending with "/.rpm" will be still treated as directories) */
+    if((strlen(path) > 5) && (!strcmp(path + strlen(path) - 4, ".rpm")) &&
+            (*(path + strlen(path) - 5) != '/')) {
+
+        if (verbose) {
+            /* careful - standard output is being parsed */
+            printf("Listing DD dir %s\n", path);
```
vpodzime: Is this branch of code really listing a dir?

utils/dd/dd_list.c
japokorn: Fixed.

utils/dd/dd_list.c
japokorn: Fixed.

utils/dd/dd_list.c
japokorn: Original idea was that there is a possibility user will want to install rpm files from a hidden directory named ".rpm". Reconsidered as not probable - feature removed.

utils/dd/dd_list.c
japokorn: Fixed.

utils/dd/dd_list.c
japokorn: No it's not. Fixed.

utils/dd/dd_list.c
```diff
@@ -206,38 +211,50 @@ int main(int argc, char *argv[])
...
 
-    char *globpattern;
-    checked_asprintf(&globpattern, "%s/*.rpm", directory);
+    path_len = strlen(path);
 
-    glob_t globres;
-    char** globitem;
+    /* determine if the path is rpm filename or directory */
+    if ((path_len > 5) && (strcmp(path + path_len - 4, ".rpm") == 0)) {
+
```
vpodzime: I think this blank line could be removed.

vpodzime:

utils/dd/dd_list.c
japokorn: Fixed.

japokorn:
Pushed.

PR #646, opened 2016-05-26
Print an error when the kickstart file is missing (#1297380)

bcl:
If --kickstart has been passed we should make sure it actually exists.
If we don't the installation tries to continue but gets confused because
automatedInstall has been set.

(cherry picked from commit 8c06aeee7860d262ce8b1dec037d1cfaa670cc11)

Resolves: rhbz#1297380


vpodzime:
Looks good to me.

bcl:
Pushed

PR #647, opened 2016-05-26
Require network for network-based driver disks

dashea:
This will automatically activate the network if inst.dd requires it,
similar to the inst.ks and inst.updates parameters.

Resolves: rhbz#1261024


vpodzime:
I wonder how this managed to live unnoticed for so long. Maybe because we are all doing PXE installations? Anyway, LGTM.

dashea:
Whoops, missed that @japokorn already fixed this in https://github.com/rhinstaller/anaconda/pull/535. That one's fine, too.

PR #648, opened 2016-05-26
send all layouts to localed for keymap conversion (#1333998)

AdamWill:
Resolves: rhbz#1333998

localed's code for deciding what VConsole keymap matches a given
X11 layout/variant config is more likely to work well when we
give it the whole list of X11 layout/variants we have, not just
the first. If you pass it only a single X11 layout/variant, it
will only work if there is an entry in the model map table that
is for *exactly* that layout/variant; for instance, if you pass
it 'ru', it will not consider its entry for 'ru,us' to be a
match. If you pass it a list of layouts, it's actually rather a
lot more likely to find a match - it would consider its 'ru,us'
entry to be a (weak) match for 'ru,foo,bar', for example,
because as a last resort it tries matching just the first part
(up to the first comma) of the layout against the first part of
the table entry, but *only* if you give it multiple layouts.

### DIFF
--- a/pyanaconda/keyboard.py
+++ b/pyanaconda/keyboard.py
@@ -138,7 +138,7 @@ def populate_missing_items(keyboard):
         keyboard.vc_keymap = keyboard._keyboard
 
     if keyboard.x_layouts and not keyboard.vc_keymap:
-        keyboard.vc_keymap = localed.convert_layout(keyboard.x_layouts[0])
+        keyboard.vc_keymap = localed.convert_layouts(keyboard.x_layouts)
 
     if not keyboard.vc_keymap:
         keyboard.vc_keymap = DEFAULT_KEYBOARD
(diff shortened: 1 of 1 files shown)


AdamWill:
This is obviously an interface change. I just went with this simple version for now as I can't find anything outside this file which uses any of these methods. If compatibility is a worry, we can do the usual icky hacks to let the methods handle both a list of layouts or just a single layout, and leave the method names unchanged.

vpodzime:
I'm not sure we should be working around weird behavior of *localed*, but on the other hand it probably make sense to give it all information we have. I'm just wondering which unexpected changes this will introduce. :) Anyway, ACK from me.

AdamWill:
"I'm just wondering which unexpected changes this will introduce" - yeah, I know what you mean :/

AdamWill:
So systemd are improving things from their end too, but I think this change is still appropriate for anaconda; can I get another ack and merge? It probably doesn't need to go to F24 now, though, may be best to save it for Rawhide.

vpodzime:
> So systemd are improving things from their end too, but I think this change is still appropriate for anaconda; can I get another ack and merge? It probably doesn't need to go to F24 now, though, may be best to just do master.

Pushed to master. Let me know if/once this needs to go to the f24-branch too.

PR #650, opened 2016-05-27
Handling for corrupt or otherwise unsupported disklabels

dwlehman:
This is the anaconda piece to rhinstaller/blivet#427. The first patch is a touch-up for a recent commit. The next two are little tweaks to use blivet better, and the last one is the main event.

### DIFF
--- a/pyanaconda/storage_utils.py
+++ b/pyanaconda/storage_utils.py
@@ -489,3 +489,8 @@ class StorageSnapshot(object):
 # a snapshot of early storage as we got it from scanning disks without doing any
 # changes
 on_disk_storage = StorageSnapshot()
+
+def filter_unsupported_disklabel_devices(devices):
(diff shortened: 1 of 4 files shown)


pyanaconda/storage_utils.py
```diff
@@ -489,3 +489,8 @@ def reset_to_snapshot(self, storage, dispose=False):
 # a snapshot of early storage as we got it from scanning disks without doing any
 # changes
 on_disk_storage = StorageSnapshot()
+
+def filter_unsupported_disklabel_devices(devices):
```
vpodzime: This is a terrible name for a function, but I don't have a better one. Except for maybe ``get_suported_disklabel_devices()``? It's at least a bit shorter.

vpodzime:
Looks good to me.

dwlehman:
I had fixed that function name but then I accidentally pushed an older version of the patches without that fix.

PR #651, opened 2016-05-27
Handling for corrupt or otherwise unsupported disklabels (rhel7)

dwlehman:
This is the rhel7 version of #650. It goes along with rhinstaller/blivet#428.

Hide everything except the disk from the user when the disklabel is unsupported. The user can choose to leave the disk as-is or clear it completely.


vpodzime:
Looks good to me.

dwlehman:
Pushed.

PR #652, opened 2016-05-31
Don't try to do payload reset in Initial Setup

M4rtinK:
At the moment the network spoke tries to trigger a payload reset when network settings are changed even if running inside of Initial Setup. This is of course wrong, as there is no payload instance in Initial Setup, which can trigger a lot of various issues (crashes, weird log messages, etc.).

The first patch adds the environs flag so that spokes can easily check the environment at runtime. It also gets rid of the old _environs hub property.

The second patch then uses the environs flag to only trigger the payload reset when the network spoke is run from inside of Anaconda.

### DIFF
--- a/pyanaconda/flags.py
+++ b/pyanaconda/flags.py
@@ -20,7 +20,7 @@
 import selinux
 import shlex
 import glob
-from pyanaconda.constants import SELINUX_DEFAULT, CMDLINE_APPEND
+from pyanaconda.constants import SELINUX_DEFAULT, CMDLINE_APPEND, ANACONDA_ENVIRON
 from collections import OrderedDict
 
 import logging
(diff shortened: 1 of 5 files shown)


jkonecny12:
Please fix this pylint error. Jenkins is complaining about it.
``************* Module ../pyanaconda/ui/tui/spokes/network.py``
``E0611: 24,0: : No name 'environs' in module 'pyanaconda.flags'``

M4rtinK:
>Please fix this pylint error. Jenkins is complaining about it.
************* Module ../pyanaconda/ui/tui/spokes/network.py
E0611: 24,0: : No name 'environs' in module 'pyanaconda.flags'

Yep, that's indeed wrong. Should be fixed now.

jkonecny12:
Looks good to me too.

M4rtinK:
Pushed.

PR #654, opened 2016-06-01
Keep the subdir in driver disk update paths (#1296306)

dashea:
This is important for drivers that include a subdirectory in the
firmware name, since they will be looking for the subdirectory under
/lib/firmware/updates. Rearrange copy_files and move_files so that
subdirectories under /lib/firmware and /lib/modules/<version>/kernel are
retained.

### DIFF
--- a/dracut/driver_updates.py
+++ b/dracut/driver_updates.py
@@ -206,21 +206,41 @@ def ensure_dir(d):
     """make sure the given directory exists."""
     subprocess.check_call(["mkdir", "-p", d])
 
-def move_files(files, destdir):
+def move_files(files, destdir, basedir):
     """move files into destdir (iff they're not already under destdir)"""
-    ensure_dir(destdir)
     for f in files:
         if f.startswith(destdir):
             continue
-        subprocess.call(["mv", "-f", f, destdir])
+        dest = destdir+"/"+dest_strip(f, basedir)
+        ensure_dir(os.path.dirname(dest))
+        subprocess.call(["mv", "-f", f, dest])
 
-def copy_files(files, destdir):
+def dest_strip(dest, basedir):
+    """strip a base directory plus kernel version from a path"""
+    # Strip the basedir and any leftover leading /'s
+    dest = dest[len(basedir):]
+    while dest.startswith('/'):
+        dest = dest[1:]
+
+    # Look for a leading directory that is a version number
+    if "/" in dest and fnmatch.fnmatch(dest, "*.ko*") and dest[0].isdigit():
+        # Drop the leading directory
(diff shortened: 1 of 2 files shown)


dracut/driver_updates.py
```diff
@@ -206,21 +206,41 @@ def ensure_dir(d):
     """make sure the given directory exists."""
     subprocess.check_call(["mkdir", "-p", d])
 
-def move_files(files, destdir):
+def move_files(files, destdir, basedir):
     """move files into destdir (iff they're not already under destdir)"""
```
bcl: May as well fix the iff typo

dracut/driver_updates.py
```diff
@@ -206,21 +206,41 @@ def ensure_dir(d):
...
+        # Drop the leading directory
+        dest = "/".join(dest.split('/')[1:])
+
+        if dest.startswith("kernel/"):
+            dest = "/".join(dest.split('/')[1:])
+
+    return dest
+
+def copy_files(files, destdir, basedir):
     """copy files into destdir (iff they're not already under destdir)"""
```
bcl: Ha! copy-n-paste typos.

dracut/driver_updates.py
```diff
@@ -206,21 +206,41 @@ def ensure_dir(d):
...
 
-def move_files(files, destdir):
+def move_files(files, destdir, basedir):
     """move files into destdir (iff they're not already under destdir)"""
-    ensure_dir(destdir)
     for f in files:
         if f.startswith(destdir):
             continue
-        subprocess.call(["mv", "-f", f, destdir])
+        dest = destdir+"/"+dest_strip(f, basedir)
```
vpodzime: We usually write spaces around operators.

dracut/driver_updates.py
```diff
@@ -206,21 +206,41 @@ def ensure_dir(d):
...
+
+def copy_files(files, destdir, basedir):
     """copy files into destdir (iff they're not already under destdir)"""
-    ensure_dir(destdir)
     for f in files:
         if f.startswith(destdir):
             continue
-        subprocess.call(["cp", "-a", f, destdir])
+
+        dest = destdir+"/"+dest_strip(f, basedir)
```
vpodzime: Same here, but I see this file is a bit special regarding the spaces around operators.

dracut/driver_updates.py
dashea: I assumed that was intentional, for "if and only if"

dashea:
Pushed an update for the pylint error I introduced (changed the parameter name "dir" to "path")

dashea:
Pushed.

PR #655, opened 2016-06-01
Display storage errors that cause no disks to be selected (#1340240)

bcl:
Problems allocating space can result in no disks being selected, but
displaying 'No disks selected; ...' doesn't make it obvious what has
happened.

This checks for errors and warnings and displays them instead, hopefully
leading the user to the actual problem.

eg. this happens when a kickstart tries to allocate partitions that are
too big for the disk(s).

Related: rhbz#1340240


vpodzime:
Looks good to me.

bcl:
Pushed to master and rhel7-branch

PR #656, opened 2016-06-02
Add a refresh button to the storage spoke

dwlehman:
This allows users to hot-add disks to a vm without rebooting or having to find the other refresh buttons that exist in the filter spoke and the custom spoke.

The button is in the lower-right corner.

refresh-disks-rawhide 0


dwlehman:
Pushed.

PR #657, opened 2016-06-02
Add a button to refresh the disk list.

dwlehman:
(cherry picked from commit 1ea511667713fbe860ec34901d560840bdee7e5e)

Resolves: rhbz#1191305

This is the `rhel7-branch` version of #656. The button is in the lower right corner and it runs the same dialog as the one in the custom spoke:

refresh-disks-el7-0
refresh-disks-el7-1


dwlehman:
Pushed.

PR #658, opened 2016-06-02
Stop using undocumented DNF logging API

bcl:
DNF doesn't want users to access base.logging anymore.

This changes the dnf logging so that the dnf and dnf.rpm loggers go into
packaging.log now. It preserves the dnf.librepo.log file and the
hawkey.log files so that they can be examined after the installation is
complete or crashes.

### DIFF
--- a/pyanaconda/anaconda_log.py
+++ b/pyanaconda/anaconda_log.py
@@ -179,6 +179,13 @@ class AnacondaLog:
                             autoLevel=True)
         self.forwardToSyslog(packaging_logger)
 
+        # Create the dnf logger and link it to packaging
+        dnf_logger = logging.getLogger("dnf")
+        dnf_logger.setLevel(logging.DEBUG)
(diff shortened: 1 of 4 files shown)


pyanaconda/anaconda_log.py
```diff
@@ -179,6 +179,13 @@ def __init__(self):
                             autoLevel=True)
         self.forwardToSyslog(packaging_logger)
 
+        # Create the dnf logger and link it to packaging
+        dnf_logger = logging.getLogger("dnf")
+        dnf_logger.setLevel(logging.DEBUG)
```
vpodzime: Any reason why we are using a different log level than in lorax here?

vpodzime: Oh, I see it below now.

vpodzime:
Looks good to me.

bcl:
Pushed to master.

PR #660, opened 2016-06-03
Warn about broken keyboard layout switching in VNC (#1274228)

jkonecny12:
Keyboard layout switching doesn't work in VNC. VNC is using its own layout system. Overall the keyboard setting in VNC looks like it is ignored.
This commit replaces the message how to switch layout ("Alt+Shift"...) for VNC with message that shows warning to a user that keyboard settings isn't used for VNC, but will be applied after the installation.

*Resolves: rhbz#1274228*


vpodzime:
Looks good to me.

jkonecny12:
Pushed

PR #661, opened 2016-06-03
Fix crash when NM get_setting* methods return None (#1273497)

jkonecny12:
Test calling of these methods on few places.

*Resolves: rhbz#1273497*


vpodzime:
Looks good to me.

rvykydal:
Looks good to me as well.

jkonecny12:
Pushed

PR #662, opened 2016-06-03
Make the anaconda-generator exit early outside of the installation en…

M4rtinK:
…vironment (#1289179)

Check if the anaconda-generator is running in the installation
environment by checking where the default.target is pointing.

Proceed normally if it is pointing to the anaconda.target,
exit early if not.

This should get rid of the annoying "anaconda-generator failed with
error code 1" errors in journal when the generator tried to run in
outside of the installation environment.

Resolves: rhbz#1289179

### DIFF
--- a/data/systemd/anaconda-generator
+++ b/data/systemd/anaconda-generator
@@ -1,6 +1,14 @@
 #!/bin/bash
 # anaconda-generator: generate services needed for anaconda operation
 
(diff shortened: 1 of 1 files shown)


vpodzime:
Looks good to me.

M4rtinK:
Pushed.

PR #663, opened 2016-06-03
Overwrite network files when using ks liveimg (#1342639)

bcl:
A ks liveimg install may have pre-existing network files, overwrite them
on the target the same way we do for live installs. This effects:

/etc/hostname
/etc/resolv.conf
/etc/sysconfig/network

Resolves: rhbz#1342639


jkonecny12:
Looks good to me.

fabiand:
Looks good to me as well

bcl:
Pushed to master.

bcl:
Pushed to rhel7-branch

PR #664, opened 2016-06-07
Fix a covscan warning about fetch-driver-net (#1269915)

bcl:
warning: Declare and assign separately to avoid masking return values.
[SC2155]

Related: rhbz#1269915


snbueno:
Looks fine to me.

bcl:
Pushed.

PR #665, opened 2016-06-07
Fix the name sensitivity in the custom spoke.

dashea:
When switching from an existing device to a non-existing device of the
same type, the sensitivity of the "Name" field was not being updated.
Add an extra call to the combo box change signal handler to handle this
case.


vpodzime:
Looks good to me.

jkonecny12:
I'm missing the bug number in commit message for RHEL. Otherwise it looks good to me.

dashea:
Pushed to master. The rhel7 bug still needs acks.

PR #666, opened 2016-06-08
Fix old blivet identifiers (#1343907)

dashea:
Fix another couple identifiers that pylint didn't catch


dashea:
Pushed one more commit to handle the iscsi object moving. iscsi still fails because the storaged dbus service is missing, but I'll leave that for someone else.

vpodzime:
Looks good to me.

dashea:
Pushed.

PR #667, opened 2016-06-09
Add wordwrap to text mode and use it by default (#1267881)

rvykydal:
Resolves: rhbz#1267881


dashea:
Looks good to me

vpodzime:
Looks good to me too.

rvykydal:
Pushed

PR #668, opened 2016-06-09
Rhevh 4.0

dougsland:
Request for creation of rhevh-4.0 branch based on rhel7-branch for patches related to rhevh 4.0.


fabiand:
Awesome!

But instead of merging this into the `rhel7-branch` I'd suggest that we use this for the to be created `rhevh-4.0` branch.

The `rhevh-4.0` branch should be branched off the tip which was used for the rhel-7.2 anaconda build.

dougsland:
This is not updated pull request

PR #669, opened 2016-06-10
Turn translations back on.

dashea:
gettext 0.19.8 fixes the bug that we were hitting with msgfmt and
desktop files.  This reverts abae279fef8aeb69ba07db4b56c6f10622e939e3.


bcl:
Oh, whoops, I did this by just reverting the commit to turn it off, so we'll still need to update the gettext version.

PR #670, opened 2016-06-13
Use the RHEL 7 timezone kickstart command version

M4rtinK:
The RHEL 7 version of the command has been modified to make it possible
to use the --utc flag without requiring a timezone specification to be provided.

This way the hardware clock time format can be set from kickstart while
still requiring the user to select a timezone manually in the UI.

Resolves: rhbz#1312135

### DIFF
--- a/pyanaconda/kickstart.py
+++ b/pyanaconda/kickstart.py
@@ -1658,7 +1658,7 @@ class SshKey(commands.sshkey.F22_SshKey):
         for usr in self.sshUserList:
             users.setUserSshKey(usr.username, usr.key)
 
-class Timezone(commands.timezone.F18_Timezone):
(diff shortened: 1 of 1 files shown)


M4rtinK:
The Pykickstart PR is https://github.com/rhinstaller/pykickstart/pull/87.

vpodzime:
I think a little bit more needs to happen here. At minimum, the ``timezone.seen`` flag will now have a different meaning, AFAICT. But maybe that could be handled in the pykickstart part of the changes?

clumens:
.seen should really just mean "was this command given in the kickstart file?".  Any additional meaning would have to come from inspecting attributes for None vs. some preset vs. new value.

M4rtinK:
Pushed (a while ago).

PR #672, opened 2016-06-16
Update iscsi dialog for Blivet 2.0 API change

bcl:
iscsi.initiatorSet -> iscsi.initiator_set


pyanaconda/ui/gui/spokes/advstorage/iscsi.py
```diff
@@ -169,7 +169,7 @@ def refresh(self):
         self._storeFilter.set_visible_column(1)
 
         self._initiatorEntry.set_text(self.iscsi.initiator)
-        self._initiatorEntry.set_sensitive(not self.iscsi.initiatorSet)
+        self._initiatorEntry.set_sensitive(not self.iscsi.initiator-set)
```
dashea: I think that should _ instead of -

bcl:
Pushed.

PR #673, opened 2016-06-17
Re-configure proxy when updateBaseRepo is called (#1332472)

bcl:
The dnf proxy configuration was only being setup when _configure is
called, which only happens once. This means that the proxy dialog
wouldn't have any effect when using the closest mirrors. This can be
worked around by passing inst.proxy on the cmdline.

The proxy dialog works fine for http/https because it is setup
explicitly in updateBaseRepo call.

This adds setting up the proxy on the DNF Base object whenever it is
reset() so that the proxy settings will be current, and used for closest
mirror.


vpodzime:
Looks good to me.

bcl:
Pushed

PR #674, opened 2016-06-17
Fix adding new VG in Custom spoke can't be applied (#1263715)

jkonecny12:
When a new VG was created and nothing inside of a dialog was changed then the changes can't be confirmed by the "Update Settings" button.

Method which runs the dialog assumed, when nothing was modified the update button don't need to be updated. This method now gets info, if this is a new VG or modification of an existing one.

*Resolves: rhbz#1263715*

*Reported-by: Jan Stodola ``<jstodola@redhat.com>``*

### DIFF
--- a/pyanaconda/ui/gui/spokes/custom.py
+++ b/pyanaconda/ui/gui/spokes/custom.py
@@ -397,6 +397,10 @@ class CustomPartitioningSpoke(NormalSpoke, StorageChecker):
         self._updateSpaceDisplay()
         self._applyButton.set_sensitive(False)
 
+    def _get_container_names(self):
+        for data in self._containerStore:
+            yield data[0]
+
     @property
     def _current_page(self):
         # The current page is really a function of the current selector.
@@ -1947,7 +1951,7 @@ class CustomPartitioningSpoke(NormalSpoke, StorageChecker):
         disks = dialog.selected
         log.debug("new disks for %s: %s", device.name, [d.name for d in disks])
         if not disks:
-            self._error = "No disks selected. Keeping previous disk set."
+            self._error = _("No disks selected. Keeping previous disk set.")
             self.set_info(self._error)
             return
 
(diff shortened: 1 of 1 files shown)


pyanaconda/ui/gui/spokes/custom.py
```diff
@@ -2007,16 +2007,20 @@ def run_container_editor(self, container=None, name=None):
         if not disks:
             self._error = "No disks selected. Not saving changes."
```
bcl: Unrelated, but this should be translated.

pyanaconda/ui/gui/spokes/custom.py
```diff
@@ -2007,16 +2007,20 @@ def run_container_editor(self, container=None, name=None):
...
             self.set_info(self._error)
+            log.error("No disks selected for volume group.")
```
bcl: This should be the same text as the _error to prevent confusion.

bcl:
Looks ok to me, other than those nitpicks.

pyanaconda/ui/gui/spokes/custom.py
jkonecny12: Totally true if I didn't missed it would be fixed in this patch already. Thank you.

pyanaconda/ui/gui/spokes/custom.py
jkonecny12: OK, I'll change it locally.

pyanaconda/ui/gui/spokes/custom.py
```diff
@@ -2007,16 +2007,20 @@ def run_container_editor(self, container=None, name=None):
...
             return
 
         log.debug("new container name: %s", name)
-        if name != container_name and name in self._storage_playground.names:
+        if (name != container_name and name in self._storage_playground.names or
+            name in [data[0] for data in self._containerStore]):
```
vpodzime: This could use a generator instead of list comprehension.

jkonecny12: Hmm, good idea. I'm going to rework this.

jkonecny12:
Updated. 
Generator was added and I also fixed small bug there.

vpodzime:
Looks good to me.

jkonecny12:
Pushed, thank you.

PR #675, opened 2016-06-17
Fix SimpleConfigFile file permissions (#1346364)

bcl:
It was grabbing the permissions of the original file, and then applying
them to the original file, instead of the new temporary file.

Also adds a test to make sure the original file permissions are
preserved.


vpodzime:
Looks good to me.

bcl:
Pushed

PR #677, opened 2016-06-20
hostname: don't set installer env hostname to localhost.localdomain (…

rvykydal:
…#1290858)

Related: rhbz#1290858

Rather keep hostname set by NetworkManager from dhcp or DNS lookup.


rvykydal:
Pushed.

PR #680, opened 2016-06-22
Add wordwrap to text mode and use it by default (#1267881)

rvykydal:
This is just a backport from rhel7-branch to master, creating PR to run CI tests.

Resolves: rhbz#1267881


rvykydal:
Pushed

PR #682, opened 2016-06-23
Add support for device specification variants (#1200833)

M4rtinK:
Make it possible to have multiple device specification variants
delimited by | in the device specification string, for example:

"sd*|hd*|vda"

or

"sd*|hd*|vda,/dev/vdc"

The spec variants are processed individually from left to right
and at least one match needs to happen for the | delimited
spec to be considered successfully applied.

Also make sure that callers of the deviceMatches() function
really get disk devices when they expect them, not partitions
or other devices that are not disks.

Resolves: rhbz#1200833

### DIFF
--- a/pyanaconda/storage_utils.py
+++ b/pyanaconda/storage_utils.py
@@ -22,10 +22,12 @@
 
 import re
 import locale
+import os
 
 from contextlib import contextmanager
 
 from blivet import util
+from blivet import udev
 from blivet.size import Size
 from blivet.errors import StorageError
 from blivet.platform import platform as _platform
@@ -90,6 +92,8 @@ AUTOPART_DEVICE_TYPES = {AUTOPART_TYPE_LVM: DEVICE_TYPE_LVM,
 
 NAMED_DEVICE_TYPES = (DEVICE_TYPE_BTRFS, DEVICE_TYPE_LVM, DEVICE_TYPE_MD, DEVICE_TYPE_LVM_THINP)
 CONTAINER_DEVICE_TYPES = (DEVICE_TYPE_LVM, DEVICE_TYPE_BTRFS, DEVICE_TYPE_LVM_THINP)
+
+udev_device_dict_cache = None
 
 def size_from_input(input_str):
     """Get size from user's input"""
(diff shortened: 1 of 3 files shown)


M4rtinK:
Please disregard for now - some Blivet-side API changes are still being discussed[0], which might influence this PR.

[0] https://github.com/rhinstaller/blivet/pull/442

bcl:
Looks interesting, it should also include some tests if you proceed with it.

M4rtinK:
The blivet part (in https://github.com/rhinstaller/blivet/pull/442) should be done now & the code in this PR has been synced with it, so feel free to review this one for real. :)

M4rtinK:
>Looks interesting, it should also include some tests if you proceed with it.
I plan to do (or enhance) a kickstart test to cover this - it is unfortunately (at least IMHO) not a good candidate for meaningful unit tests due to all the back and forth between Anaconda and Blivet & dependency on system state.

pyanaconda/storage_utils.py
```diff
@@ -415,3 +419,114 @@ def filter_unsupported_disklabel_devices(devices):
...
+    normal local disks udev.resolve_devspec should suffice.
+    """
+
+    matches = []
+    # the device specifications might contain multiple "sub specs" separated by a |
+    # - the specs are processed from left to right
+    for single_spec in spec.split("|"):
+        full_spec = single_spec
+        if not full_spec.startswith("/dev/"):
+            full_spec = os.path.normpath("/dev/" + full_spec)
```
dwlehman: This is, and always has been, a mistake IMO. Users are capable of typing `/dev/` and that saves us from pretending we know whether prepending `/dev/` is a sane thing to do.

M4rtinK: >This is, and always has been, a mistake IMO. Users are capable of typing /dev/ and that saves us from pretending we know whether prepending /dev/ is a sane thing to do.

Well, it was in the original code, so I left it there. Do you think we can safely drop it without breaking the API/user assumptions ?

dwlehman: Probably not. Not ever. That's why it was a mistake.

vpodzime:
Looks good to me.

M4rtinK:
Pushed.

PR #686, opened 2016-06-27
Check if bridgeslave argument is valid MAC and use it

landgraf:
Allows users to specify bridgeslaves in format --bridgeslaves=<MACADDR>
in the same way it's done for --device to allow automatic bridge configuration

Signed-off-by: Pavel Zhukov pzhukov@redhat.com

### DIFF
--- a/dracut/parse-kickstart
+++ b/dracut/parse-kickstart
@@ -548,6 +548,8 @@ def ksnet_to_ifcfg(net, filename=None):
             ifcfg['BRIDGING_OPTS'] = " ".join(keyvalues)
 
         for i, slave in enumerate(net.bridgeslaves.split(","), 1):
+            if is_mac(slave):
+                slave = find_devname(slave)
             slave_ifcfg = {
                             'TYPE' : "Ethernet",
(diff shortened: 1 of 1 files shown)


rvykydal:
Thank you for the patch.
Supporting mac address for --bridgeslaves option would require also changes in stage 2 anaconda (eg pyanaconda/network.py) which could involve further related changes. We need to consider if we are able to support it consistently and also create and run kickstart tests (https://github.com/rhinstaller/kickstart-tests) for the new feature. If you still care about the PR, I propose to open a Fedora RFE bugzilla report for the feature.

InStallker:

InStallker:

rvykydal:
> Thank you for the patch.
> Supporting mac address for --bridgeslaves option would require also changes in stage 2 anaconda (eg pyanaconda/network.py) which could involve further related changes. We need to consider if we are able to support it consistently and also create and run kickstart tests (https://github.com/rhinstaller/kickstart-tests) for the new feature. If you still care about the PR, I propose to open a Fedora RFE bugzilla report for the feature.

... so closing.

PR #687, opened 2016-06-28
Add support for configuring NTP servers to the TUI

M4rtinK:


pyanaconda/ui/tui/spokes/time_spoke.py
```diff
@@ -19,34 +19,113 @@
...
+
+class TimeSpoke(FirstbootSpokeMixIn, NormalTUISpoke):
+    title = N_("Time settings")
     category = LocalizationCategory
 
     def __init__(self, app, data, storage, payload, instclass):
         NormalTUISpoke.__init__(self, app, data, storage, payload, instclass)
+        self._timezone_spoke = None
+        # we use an ordered dict to keep the NTP server insertion order
+        self._ntp_servers = OrderedDict()
```
bcl: I think access to this should have a lock. It's possible that it could be checking server status while a quick fingered user could be adding/deleting from the dict.

pyanaconda/ui/tui/spokes/time_spoke.py
```diff
@@ -63,8 +142,108 @@ def status(self):
...
+
+        if self.data.timezone.timezone:
+            timezone_option = _("Change timezone")
+        else:
+            timezone_option = _("Set timezone")
+
+        _options = [timezone_option, _("Configure NTP servers")]
+        text = [TextWidget(m) for m in _options]
+
+        def _prep(i, w):
```
bcl: _prep shows up in a number of places. Seems like a good candidate for a helper function, especially since it doesn't depend on anything from the class itself.

bcl:
Looks ok to me other than those 2 comments, although I'd like to see a 3rd opinion just because of the amount of change + threading :)

pyanaconda/ui/tui/spokes/time_spoke.py
M4rtinK: That's a good point - it would not be an issue if only adding was supported (as the thread only run on any initial servers and then on any server added), but since we also support removing of NTP servers, the thread might indeed try to update status of a server that is no longer in the dictionary, resulting in a traceback.

pyanaconda/ui/tui/spokes/time_spoke.py
M4rtinK: Yep, I've noticed _prep() and some similar functions copy pasted all around when working on this patch set. I plan to first de-duplicate the code on master and then also possibly backport it to the rhel7-branch (it might be quite a lot of changed lines in absolute numbers).

BTW, I've noticed a few other issues, like always expecting untranslated c/r/b and I plan to also fix that once I'm done with this patch set.

M4rtinK:
>Oh, and fix the pylint complaints first too.

AFAIK, Vratislav is working on enabling our Jenkins instance to check branches in our anaconda/blivet, etc. forks on demand. So if everything works out it should be possible in the future to check for these test/pylint issues using the same checks & environment Jenkins runs on PRs *before submitting the PR* from a forked repo. :)

pyanaconda/ui/tui/spokes/time_spoke.py
M4rtinK: Actually, it would not cause a traceback, but rather return the removed server back to the dictionary by setting it's status. :) But of course still an issue.

pyanaconda/ui/tui/spokes/time_spoke.py
```diff
@@ -63,8 +57,114 @@ def status(self):
...
+#    @property
+#    def mandatory(self):
+#        return True
+
+#    @property
+#    def status(self):
+#        if self.data.timezone.timezone:
+#            return _("%s timezone") % self.data.timezone.timezone
+#        else:
+#            return _("Timezone is not set.")
```
vpodzime: I think the commented-out lines should go.

pyanaconda/ui/tui/spokes/time_spoke.py
```diff
@@ -145,3 +249,89 @@ def prompt(self, args = None):
...
+            newspoke = NTPServersSpoke(self.app, self.data, self.storage,
+                                       self.payload, self.instclass)
+            self.app.switch_screen_modal(newspoke)
+            self.apply()
+            self.close()
+            return INPUT_PROCESSED
+        else:
+            # the user provided an invalid option number,
+            # just stay in the spoke
+            return INPUT_PROCESSED
```
vpodzime: What is the purpose of this code which is the same as in ``TimeSpoke``?

vpodzime:
There's a typo in the last commit's message.

pyanaconda/ui/tui/spokes/time_spoke.py
```diff
@@ -34,6 +56,69 @@ class TimeSpoke(FirstbootSpokeMixIn, NormalTUISpoke):
...
+    def _check_ntp_servers(self, servers):
+        for server in servers:
+            log.debug("checking ntp server %s", server)
+            result = ntp.ntp_server_working(server)
+            if result:
+                log.debug("ntp server %s appears to be working", server)
+                self._ntp_servers[server] = constants.NTP_SERVER_OK
+            else:
+                log.debug("ntp server %s appears not to be working", server)
+                self._ntp_servers[server] = constants.NTP_SERVER_NOK
```
vpodzime: Would be nice to do the checks in parallel here.

M4rtinK:
>There's a typo in the last commit's message.
Fixed, thanks! :)

M4rtinK:
Pushed a new version:
- added locks for NTP dict access
- fixed typo in commit message
- cleaned up the leftover work-in-progress commit :P
- fixed some pylint warnings

pyanaconda/ui/tui/spokes/time_spoke.py
M4rtinK: Well, why not - should not be very hard.

M4rtinK:
Another update:
- NTP server checker now runs checks on the provided NTP server list in parallel

pyanaconda/ui/tui/spokes/time_spoke.py
bcl: Sounds like a good plan.

pyanaconda/ui/tui/spokes/time_spoke.py
```diff
@@ -145,3 +363,154 @@ def prompt(self, args = None):
...
+            return INPUT_PROCESSED
+        else:
+            # the user provided an invalid option number,
+            # just stay in the spoke
+            return INPUT_PROCESSED
+
+    def apply(self):
+        pass
+
+class AddNTPServerSpoke(EditTUIDialog):
```
snbueno: This (and RemoveNTPServerSpoke) should probably have the `indirect` property set to `True`, since they are not spokes you would ordinarily access via the hub. Probably TimeZoneSpoke as well now, since that's accessed via the general Time spoke.

M4rtinK: I investigated this and it seems like the spoke gatherer only cares about spokes:
- listed in `__all__`, eq.: `__all__ = ["NetworkSpoke"]`
- **all spokes in the file** if `__all__` is not used

I found this the hard way when it tried to instantiate all the spokes to see if they should be shown, but throwing an exception due to `__init__()` taking more parameters than a _normal_ spoke. So even if you have `indirect == True` or `showable == False`, it will never get as far as checking those if your spoke has `__init__()` with a non-standard number of parameters.

In any case I'll add `indirect == True` to the sub-spokes just in case & for consistency.

But I'm thinking that this could be handled in a better way on _master_ - possibly by using __all__ where the spokes should never show up on the hub and using the `indirect` property only in cases where its value might change based on circumstances.

pyanaconda/ui/tui/spokes/time_spoke.py
```diff
@@ -19,34 +19,152 @@
...
+        elif constants.FIRSTBOOT_ENVIRON in flags.environs:
+            ntp_servers = ntp.get_servers_from_config()
+        else:
+            log.error("tui time spoke: unsupported environment configuration %s,"
+                      "can't decide where to get initial NTP servers", flags.environs)
 
-        self._selection = ""
+        # check if the NTP servers appear to be working or not
+        if ntp_servers:
+            for server in self.data.timezone.ntpservers:
```
rvykydal: You should iterate over ntp_servers here.

M4rtinK: Yep, good catch - thanks! :)

pyanaconda/ui/tui/spokes/time_spoke.py
```diff
@@ -19,34 +19,152 @@
...
+
+        :param str server: NTP server address to add
+        """
+        # the add & remove operations should (at least at the moment) be never
+        # called from different threads at the same time, but lets just use
+        # a lock there when we are at it
+        with self._ntp_servers_lock:
+            if server not in self._ntp_servers:
+                self._ntp_servers[server] = constants.NTP_SERVER_QUERY
+            self._check_ntp_servers_async([server])
```
rvykydal: Shouldn't the check be run also only if it has not already been added?

M4rtinK:
another `push --force` for the rescue:
- added indirect `property == True/False` to the spokes
- fixed the initial NTP server import

pyanaconda/ui/tui/spokes/time_spoke.py
M4rtinK: Indeed! Fixing that too, thanks! :)

rvykydal:
Other than the two comments the PR looks good to me.

M4rtinK:
Pushed.

PR #688, opened 2016-06-29
Rhel7 branch tui username check 1350375

rvykydal:


pyanaconda/ui/tui/spokes/user.py
```diff
@@ -130,11 +126,31 @@ def status(self):
...
+
     def apply(self):
         if self.args.gecos and not self.args.name:
             username = guess_username(self.args.gecos)
-            if not USERNAME_VALID.match(username):
-                self.errors.append(_("Invalid user name: %s.\n") % username)
+            valid, msg = check_name(username)
+            if not valid:
+                self.errors.append(_("Invalid user name: %(username)s.\n%(error_message)s")
+                        % {"username": username, "error_message": msg})
```
dashea: This is setting of the style_guide check, but it is valid. Either %(username) can be changed to something else, of you can add ```'pyanaconda/ui/tui/spokes/user.py': {'username': 1}``` to expected_badness in tests/gettext/style_guide.py

pyanaconda/ui/tui/spokes/user.py
```diff
@@ -130,11 +126,31 @@ def status(self):
...
 
+    def input(self, args, key):
+        self.dialog.wrong_input_message = None
+        try:
+            field = self.visible_fields[int(key)-1]
+        except (ValueError, IndexError):
+            pass
+        else:
+            if field.attribute == "gecos":
+                self.dialog.wrong_input_message = _("Full name can't contain ':' character")
```
M4rtinK: Just a simple suggestion: 

"Full name can't contain **the** ':' character"

That IMHO sounds better. :)

pyanaconda/ui/tui/spokes/user.py
```diff
@@ -130,11 +126,31 @@ def status(self):
...
+            elif field.attribute == "name":
+                # more granular message is returned by check_name
+                pass
+            elif field.attribute == "_groups":
+                self.dialog.wrong_input_message = _("Either some group name is invalid or groups are not separated by comma")
```
M4rtinK: Again just a nitpick/suggestion:

"Either **a** group name **in the group list** is invalid or groups are not separated by **a** comma"

Sounds tiny bit better that way IMHO.

M4rtinK:
Other than those nitpicks looks good to me. :)

rvykydal:
Changed %(username) to %(user_naem), applied Martin's suggestions and pushed.

PR #689, opened 2016-06-29
DD can now replace existing drivers (#1101653)

japokorn:
When using Driver Update Disks to insert new kernel module, system now
checks if the module alias is already loaded into kernel. In case it is,
the original module is replaced.

module-setup.sh changes add modinfo command support

Resolves: rhbz#1101653

### DIFF
--- a/dracut/driver_updates.py
+++ b/dracut/driver_updates.py
@@ -342,29 +342,41 @@ def grab_driver_files(outdir="/updates"):
 
     return module_dict
 
-def load_drivers(moddict):
-    """run depmod and try to modprobe all the given module names."""
+def get_all_loaded_modules():
+    """parse /proc/modules for all loaded kernel modules"""
+    all_modules = []
+    with open("/proc/modules", "r") as modules:
+        for line in modules:
+            module_name = line.split(" ")[0]
+            all_modules.append(module_name)
+    return all_modules
 
+def load_drivers(moddict):
+    """load all drivers based on given aliases. In case the drivers are
+    already present in the kernel, replace them with the new ones.
+    """
     # Step 1: try to unload everything that's being replaced
     # Using the current depmod data, resolve all the aliases to a module name,
     # and pass those names to modprobe -r.
     # modprobe can probably handle the aliases themselves, but this reduces this
     # list so we don't have to worry as much about what the maximum command line
     # length is.
+
+    # save snapshot of currently installed modules
+    all_modules_org = get_all_loaded_modules()
     unload_modules = set()
-    for _modname, alias_list in moddict.iteritems():
-        for alias in alias_list:
-            cmd = ["modprobe", "-R", alias]
-            try:
-                out = subprocess.check_output(cmd, stderr=DEVNULL)
-                if out:
-                    unload_modules.update(out.strip().split('\n'))
-            except subprocess.CalledProcessError:
-                pass
+    for modname in moddict.keys():
+        cmd = ["modprobe", "-R", modname]
+        try:
+            out = subprocess.check_output(cmd, stderr=DEVNULL)
+            if out:
+                unload_modules.update(out.strip().split('\n'))
+        except subprocess.CalledProcessError:
+            pass
 
     log.debug("unload drivers: %s", unload_modules)
     if unload_modules:
-        subprocess.call(["modprobe", "-a", "-r"] + list(unload_modules))
+        subprocess.call(["modprobe", "-r"] + list(unload_modules))
 
     # Step 2: Update the depmod data and try to load the new module list
     log.debug("load_drivers: %s", moddict.keys())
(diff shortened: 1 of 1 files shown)


dracut/driver_updates.py
```diff
@@ -320,11 +320,46 @@ def grab_driver_files(outdir="/updates"):
...
 
+def get_all_loaded_modules():
+    """parse /proc/modules for all loaded kernel modules"""
+    all_modules = []
+    with open("/proc/modules", "r") as modules:
+        line = modules.readline()
+        while line:
+            module_name = line.split(" ")[0]
+            all_modules.append(module_name)
+            line = modules.readline()
```
vpodzime: You can use a *for* cycle on the file object.

dashea:
Where is modinfo used? The modules removed are based on just the names passed to load_drivers and not their aliases.

Also, I think that process_* functions need to be rearranged so that the module removals and module loads happen outside of the mounted() context manager. In the example in the bug, the module in the bug is being used to access the DUD itself.

dracut/driver_updates.py
```diff
@@ -320,11 +320,46 @@ def grab_driver_files(outdir="/updates"):
...
+    """
     log.debug("load_drivers: %s", modnames)
+    # save snapshot of currently installed modules
+    all_modules_org = get_all_loaded_modules()
+    # find and unload already loaded modules
+    # (Replacement modules are not known, until 'depmod -a' is called)
+    modules_to_replace = [x for x in modnames if x in all_modules_org]
+    # unload all marked modules as well as dependant modules
+    subprocess.call(["modprobe", "-r"] + modules_to_replace)
+    if modules_to_replace == []:
```
vpodzime: ``if not modules_to_replace``

bcl: And the subprocess call should be inside the check.

dracut/driver_updates.py
```diff
@@ -320,11 +320,46 @@ def grab_driver_files(outdir="/updates"):
...
+        log.debug("replace_drivers: %s", modules_to_replace)
+
+    # original modules are all unloaded now
+    # refresh modules tree (i.e. switch to new modules)
     subprocess.call(["depmod", "-a"])
-    subprocess.call(["modprobe", "-a"] + modnames)
+    # add new modules
+    subprocess.call(["modprobe", "-a"] + modules_to_replace)
```
bcl: Should be inside a check of modules_to_replace since it could (and usually will) be an empty list.

dracut/driver_updates.py
```diff
@@ -320,11 +320,46 @@ def grab_driver_files(outdir="/updates"):
...
+
+    # get new snapshot of currently installed modules
+    all_modules_new = get_all_loaded_modules()
+    # compare snapshots and get modules removed from system due to dependencies
+    diff_modules = list(set(all_modules_org) - set(all_modules_new))
```
bcl: No need to turn this into a list and then back into a set in the next line.

dracut/driver_updates.py
```diff
@@ -320,11 +320,46 @@ def grab_driver_files(outdir="/updates"):
...
+    modules_to_add = list(set(diff_modules).union(set(modnames)))
+
+    # load all previously removed modules again
+    subprocess.call(["modprobe", "-a"] + modules_to_add)
```
bcl: Also needs a check for empty modules_to_add list.

dracut/driver_updates.py
japokorn: check redesigned. Fixed.

dracut/driver_updates.py
japokorn: Moved to check. Fixed.

dracut/driver_updates.py
japokorn: Fixed.

dracut/driver_updates.py
japokorn: modules_to_add is merged (using union) with modnames which is mandatory non-empty parameter of the function. This guarantees list nonemptiness.

japokorn:
modinfo command is a rudiment of an older code version. Not used anymore. File change reverted. Fixed.

dashea:
> modinfo command is a rudiment of an older code version. Not used anymore. File change reverted. Fixed.

I thought part of the point of the bug was to be able to replace modules with different modules that provide the same functionality, e.g., replacing ahci in the initrd with megasr from a DUD. That would require checking the aliases provided by the DUD modules, and like I mentioned before, possibly running all of this with the DUD unmounted.

japokorn:
I actually spent some time earlier today analyzing this and frankly - I am not sure if it is a good idea (at least at the moment). There are already multiple modules that are in use at the moment of unloading and thus cannot be unloaded. Releasing of used kernel modules which may be in use by some custom process would be extremely complicated and possibly dangerous. In this case I believe the best solution is to leave the stopping of the module up to the customer.
Another thing is that mounted() context manager is a part of (possibly) recursively called function. To refactor the code so this can be done outside mounted() would take some time and I think it would be out of scope of this bug.

dashea:
No, it's probably not a great idea, but it's what we agreed to do. On the other hand, I don't think it's all that bad. This is all running in the initrd, so the processes that are going to be running are:

- dracut modules that are part of the installer initrd
- the driver update disk, because it's one of those.

If someone manages to get something else in there that prevents an updated module from being unloaded, that's their own fault. If a module can't be unloaded for another reasons, we're no worse off for giving it a try and continuing afterwards.

As far as mounted being called from a recursive function: just move load_drivers to main.

https://github.com/dashea/anaconda/commit/3dd73eb80334f7accdedfedd7e4c98a67efee5d0 is more what I was thinking of. I have not yet tested it.

dashea:
To address the other part of this:

> In this case I believe the best solution is to leave the stopping of the module up to the customer.

We kind of explicitly disallow that. The scripts in the DUD rpms are not run. The files are extracted, and in the context of the drivers used in the initrd and stage2, that's all that happens. Blacklists may be insufficient since they make it impossible to update and use a driver for the installation itself.

jkonecny12:
Looks good to me and for Jenkins on Rhel too.

japokorn:
Pushed.

PR #693, opened 2016-07-01
Allow kickstart users to ignore the free space error

dashea:
When attempting to install via kickstart with insufficient space for the
selected packages, let the user try to make it work anyway. For
graphical and text modes, the installation will stop at the summary hub
with the warning message but allow the user to begin the installation,
and in cmdline mode the installation will begin on its own.

In all cases the insufficient space message is written to anaconda.log.

Resolves: rhbz#1287308

### DIFF
--- a/pyanaconda/ui/gui/hubs/__init__.py
+++ b/pyanaconda/ui/gui/hubs/__init__.py
@@ -92,6 +92,10 @@ class Hub(GUIObject, common.Hub):
         self._spokes = {}
 
         self._checker = None
+        # Flag to indicate the user can continue even if the checker indicates an error.
+        # The checker itself is left alone so the error message doesn't accidentally get
+        # cleaered.
+        self._checker_ignore = False
 
         self._spokesToStepIn = []
         self._spokeAutostepIndex = 0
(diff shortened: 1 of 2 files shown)


jkonecny12:
I have two notes to this.

1) There should be some information for user to know he/she can continue the installation. AFAIK in GUI the Begin button will still be blue so there it shouldn't be a problem but TUI could be non-intuitive.

2) What about completely automated installation. You can't block those and how it will work with ``cmdline`` mode?

dashea:
> 1) There should be some information for user to know he/she can continue the installation. AFAIK in GUI the Begin button will still be blue so there it shouldn't be a problem but TUI could be non-intuitive.

Ok. I'll see how it looks just adding a sentence to the end of self._checker.error_message.

> 2) What about completely automated installation. You can't block those and how it will work with cmdline mode?

In cmdline mode, it prints the error and then tries to continue the install anyway. This is using flags.ksprompt at https://github.com/rhinstaller/anaconda/pull/693/commits/1287b3945e8cae58ce3f10f51a04cad7cee1f604#diff-b67b0c8e44d0aead1286606c73674326L78

jkonecny12:
It looks good to me otherwise. So when the info message will be add then I will give you an ACK.

dashea:
Warning added to TUI.

tui_reqspace_warning

jkonecny12:
Only one small neat pick, I'm missing period on the end of the sentence. Otherwise it looks good to me ;), thank you.

PR #694, opened 2016-07-05
Fix the processing of device nodes as driver disks

dashea:
Device nodes generally do not end in .iso, so the change to require that
disks end in .iso broke the processing of driver disk devices. Correct
this.

Related: rhbz#1269915

### DIFF
--- a/dracut/driver_updates.py
+++ b/dracut/driver_updates.py
@@ -625,10 +625,17 @@ def main(args):
 
     if mode in ('--disk', '--net'):
         request, dev = args
+
+        # Guess whether this is an ISO or RPM based on the filename.
+        # If neither matches, assume it is a device node and processes as an ISO.
+        # This is relevant for both --disk and --net since --disk could be
+        # pointing to files within the initramfs.
         if dev.endswith(".iso"):
             process_driver_disk(dev)
-        else:
+        elif dev.endswith(".rpm"):
             process_driver_rpm(dev)
+        else:
+            process_driver_disk(dev)
 
     elif mode == '--interactive':
         log.info("starting interactive mode")


dracut/driver_updates.py
```diff
@@ -625,10 +625,17 @@ def main(args):
...
         request, dev = args
+
+        # Guess whether this is an ISO or RPM based on the filename.
+        # If neither matches, assume it is a device node and processes as an ISO.
+        # This is relevant for both --disk and --net since --disk could be
+        # pointing to files within the initramfs.
         if dev.endswith(".iso"):
             process_driver_disk(dev)
-        else:
+        elif dev.endswith(".rpm"):
```
vpodzime: I'm not sure I'm getting the commit 14c4fa right, but wasn't the purpose of the change to allow for installation of multiple DD RPMs from a given (NFS) directory?

dashea: The part that collects the rpms is in fetch-driver-net.sh (https://github.com/rhinstaller/anaconda/commit/14c4fa914324fd60e38f608b52d1511dfd4713d1#diff-7303c2903a26e1e3529239060b08765dR50), where it loops over *.rpm and runs driver-updates for each. I'm honestly not real sure why the conditional here was reversed, unless the goal was to process things not ending in .rpm as rpm. If that's the case, a lot more is going to have to change, since right now the most common use case is busted.

dashea: If there's no reason for the change, this conditional could just go back to how it was before 14c4fa9

dashea:
Including this commit in https://github.com/rhinstaller/anaconda/pull/695, closing

dracut/driver_updates.py
japokorn: There is also a possibility of user specifying neither .iso nor .rpm file but a directory containing .rpms
Hence the condition reversal.

PR #695, opened 2016-07-05
Driver update disks

dashea:
There is some previous discussion on the first commit at https://github.com/rhinstaller/anaconda/pull/694

The first commit attempts to fix a bug whereby driver update disks are not processed, because driver_updates.py now expects the device node to end in .iso, which is not usually the case.

The second commit attempts to unload drivers that were loaded from the initrd before loading updated drivers from the DUD.

### DIFF
--- a/dracut/driver_updates.py
+++ b/dracut/driver_updates.py
@@ -625,10 +625,17 @@ def main(args):
 
     if mode in ('--disk', '--net'):
         request, dev = args
+
+        # Guess whether this is an ISO or RPM based on the filename.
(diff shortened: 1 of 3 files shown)


dracut/driver_updates.py
```diff
@@ -346,6 +394,7 @@ def _process_driver_disk(dev, interactive=False):
     or ask which iso file to process (if no repos).
     """
```
bcl: Should mention that it returns a dict of the modules.

dashea:
Added a comment about the return value to _process_driver_disk and pushed.

PR #696, opened 2016-07-05
Make it possible to disable sshd service from running. (#1262707)

snbueno:
Add a conditional to check whether a user specified inst.sshd=0 in
the anaconda-sshd.service file. If so, this won't start the sshd
service.

Resolves: rhbz#1262707


vpodzime:
Looks good to me.

snbueno:
I got some of the logic wrong, so I'll be pushing an update for this in a bit.

vpodzime:
Still looks good to me. :)

snbueno:
Forgot to close this; but it's been pushed.

PR #697, opened 2016-07-05
Only run when_diskdev_appears for add events.

dashea:
This makes when_diskdev_appears behave more like its name: run a command
only when a device appears. Without the ACTION qualifier, the rule also
matches change events, which could cause the command to be run more than
once.

Resolves: rhbz#1260610

### DIFF
--- a/dracut/anaconda-lib.sh
+++ b/dracut/anaconda-lib.sh
@@ -155,8 +155,8 @@ when_diskdev_appears() {
     local dev="${1#/dev/}" cmd=""; shift
     cmd="/sbin/initqueue --settled --onetime --name $1 $*"
     {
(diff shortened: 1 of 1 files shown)


vpodzime:
I'm not sure this is going to work though because AFAICT many devices are not usable when the ``ADD`` event appears.

dashea:
So what should we do? Doing kickstart and updates fetches multiple times per disk isn't the worst thing in the world, but running anaconda-diskroot and driver-updates multiple times is a problem.

msekletar:
> I'm not sure this is going to work though because AFAICT many devices are not usable when the ADD event appears.

This is certainly true for DM based block devices (lvm) but do you care for those in this particular case?

rvykydal:
This change seems to be rather risky as is.
I am closing this PR as the BZ issue is not approved for current release. The PR is referred in the BZ so let's get back to it if/when the BZ is approved for resolution.

PR #700, opened 2016-07-12
Change bootloader boot drive fallback

jkonecny12:
When bootloader doesn't have ``--boot-drive`` parameter then first disk from Blivet was used.

Now when bootloader doesn't have ``--boot-drive`` parameter then use this logic as fallback:

1. If present use first disk from ``driveorder`` parameter
2. If present and usable, use disk where ``/boot`` partition is present
3. Use first partition from *Blivet*

------------------------------------------------------------------------------------------------------------
I think this is really an issue and it should go to rhel7-branch too. What do you think?

### DIFF
--- a/pyanaconda/kickstart.py
+++ b/pyanaconda/kickstart.py
@@ -396,17 +396,43 @@ class Bootloader(commands.bootloader.F21_Bootloader):
 
         storage.bootloader.disk_order = self.driveorder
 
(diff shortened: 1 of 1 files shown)


vpodzime:
I'd swear I have seen this being fixed before. This however looks good to me.

jkonecny12:
Updated: 1 small bug was fixed and some comment changes applied.

jkonecny12:
Pushed

PR #701, opened 2016-07-12
Use sshd-keygen.target instead of hardcoded sshd-keygen script

Jakuje:
For Fedora 25 (resolving [rhbz#1331753](https://bugzilla.redhat.com/show_bug.cgi?id=1331753), we are [removing legacy `sshd-keygen` script](https://fedoraproject.org/wiki/Changes/Remove_slogin_and_sshd-keygen) and moving to systemd instantiated services (under the umbrella of `sshd-keygen.target`).

I didn't test this on s390, but it is the same we have in [Fedora `sshd.service`](http://pkgs.fedoraproject.org/cgit/rpms/openssh.git/tree/sshd.service#n4).

### DIFF
--- a/data/systemd/anaconda-sshd.service
+++ b/data/systemd/anaconda-sshd.service
@@ -1,7 +1,8 @@
 [Unit]
 Description=OpenSSH server daemon
 Before=anaconda.target
-After=syslog.target network.target
+After=syslog.target network.target sshd-keygen.target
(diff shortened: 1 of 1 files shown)


jkonecny12:
Looks good to me.

jkonecny12:
Pushed. Thank you for this patch.

PR #702, opened 2016-07-13
Fix of Python3x uncompatible command

japokorn:
".iteritems()" is not supported by Python 3; replaced with ".items()"

### DIFF
--- a/tests/dracut_tests/test_driver_updates.py
+++ b/tests/dracut_tests/test_driver_updates.py
@@ -9,6 +9,7 @@ except ImportError:
 import os
 import tempfile
 import shutil
+import collections
 
 import sys
 sys.path.append(os.path.normpath(os.path.dirname(__file__)+'/../../dracut'))
@@ -433,10 +434,11 @@ class LoadDriversTestCase(unittest.TestCase):
     def test_basic(self, call):
         """load_drivers: runs depmod and modprobes all named modules"""
         modnames = ['mod1', 'mod2']
-        load_drivers({name: [name] for name in modnames})
+        moddict = collections.OrderedDict({name: [name] for name in modnames})
+        load_drivers(collections.OrderedDict(moddict))
         call.assert_has_calls([
             mock.call(["depmod", "-a"]),
-            mock.call(["modprobe", "-a"] + modnames)
+            mock.call(["modprobe", "-a"] + list(moddict.keys()))
         ])
 
     @mock.patch("driver_updates.subprocess.call")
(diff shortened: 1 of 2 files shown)


japokorn:
jenkins, test this please

M4rtinK:
Looks good to me, ACK!

jkonecny12:
jenkins, test this please

tests/dracut_tests/test_driver_updates.py
```diff
@@ -433,10 +434,11 @@ class LoadDriversTestCase(unittest.TestCase):
     def test_basic(self, call):
         """load_drivers: runs depmod and modprobes all named modules"""
         modnames = ['mod1', 'mod2']
-        load_drivers({name: [name] for name in modnames})
+        moddict = collections.OrderedDict({name: [name] for name in modnames})
```
jkonecny12: Ordered dict doesn't work this way. Now you create a normal dict (which is not ordered) and then add it to ``OrderedDict`` so in the end it won't be ordered at all. The normal dict randomize the order before it gets to ``OrderedDict``.

You need to add items to this dict one by one:
```
for name in modnames:
    ordered_dict[name] = name
```

tests/dracut_tests/test_driver_updates.py
```diff
@@ -433,10 +434,11 @@ class LoadDriversTestCase(unittest.TestCase):
...
+        load_drivers(collections.OrderedDict(moddict)
```
jkonecny12: I'm missing one ``)`` here.

jkonecny12:
Other than that it looks great and thank you for fixing this!

tests/dracut_tests/test_driver_updates.py
jkonecny12: Ignore this, solved on IRC.

In short, ``dict`` do a randomize output before but it doesn't matter because comparison is then still ordered in ``OrderedDict``.

japokorn: Actually I don't care about order of items in the dict, if the order does not change. Standard dictionary is created, retyped to OrderedDict and stored into variable `moddict`. This one is then used when dealing with the tested method. Since OrderedDict is a `dict` class child, tested method silently inherits "sorted" behavior when working with the dictionary.

jkonecny12:
Because the fixes were small and mainly it's fixing errors in jenkins tests. Please push it ASAP :). So or so the ACK is there more then 24hours now.

Good work :+1:

PR #703, opened 2016-07-15
Use the F25 timezone kickstart command version

M4rtinK:
The F25 version of the command has been modified to make it possible
to use the --utc flag without requiring a timezone specification to be provided.

This way the hardware clock time format can be set from kickstart while
still requiring the user to select a timezone manually in the UI.

Resolves: rhbz#1312135


M4rtinK:
Pushed.

PR #704, opened 2016-07-18
Add a contributing file

snbueno:


jkonecny12:
Looks good to me. It would be nice if someone with better english could look on this also.

M4rtinK:
Also looks good to me! :)

The only thing I'm thinking about is if we should call this CONTRIBUTING as it feels to me more like a branching/development policy specification while a "how to contribute" document usually goes more like "You want to contribute to X?, Then do these steps." (commits/patches should go there & in this form). 

All the information is certainly there but IMHO a short summary targeted to perspective contributors would be IMHO nice. :)

M4rtinK:
The summary could look like this, inserted between the main headline and the "Branching Policy" sub headline & modifying the "Branching Policy" sub headline:

```
How to Contribute to the Anaconda Installer (the short version)
---------------------------------------------------------------

a) I want to contribute to the upstream Anaconda Installer (used in Fedora):
- open a pull request for the ``<next fedora number>-devel`` branch (f25-devel, etc.)
- check the *Commit Messages* section below for how to format your commit messages

b) I want to contribute to the RHEL Anaconda installer:
- open a pull request for the ``<RHEL number>-branch``  branch (rhel7-branch, etc.)
- check the *Commits for RHEL Branches* section below for how to format your commit messages

If you want to contribute a change to both the upstream and RHEL Anaconda then follow both a) and b) separately.

Anaconda Installer Branching Policy (the long version)
------------------------------------------------------
```

rvykydal:
Looks good to me.

CONTRIBUTING
```diff
@@ -0,0 +1,87 @@
...
+---------
+
+For specific Fedora version, releases are made on designated ``-release`` branches and tagged.
+
+Concerning Fedora Rawhide, the release process is slightly different:
+- master is merged onto the unstable branch
+- a release commit is made (which bumps version in spec file) & tagged
+
+Concerning the ``<next Fedora number>`` branches (which could also be called ``next stable release`` if we wanted to decouple our versioning from Fedora in the future):
+- work which goes into the next Fedora goes to ``<next Fedora number>-devel``, which is periodically merged back to ``master``
```
rvykydal: Should we mention explicitly that <next Fedora number>-devel is merged to <next Fedora number>-release or is it obvious?

snbueno: It never hurts to mention things like that. I'll add a note.

snbueno:
@M4rtinK , right, I like the summary/main headline you've suggested, so I'll definitely amend the commit to include that.

jkonecny12:
I don't know if it should be here or if it is clear enough but when we are starting with merging instead of cherry-picking I would like to see rule that **every** change made to the PR **must** be force pushed to the PR branch before merging it to the *main-branch*. In other words no more *"I will fix it locally"*.

I think it will be better for consistent history and also I don't think GitHub will handle local changes so AFAIK it won't point the PR as merged.

snbueno:
@jkonecny12, I see what you mean. I will add something like that.

jkonecny12:
Looks good to me.

PR #705, opened 2016-07-19
fix style guide test false positive on username variable (#1350375)

rvykydal:
Related: rhbz#1350375


jkonecny12:
Looks good to me.

rvykydal:
Pushed

PR #706, opened 2016-07-19
Fix bootloader when re-using existing /boot part (#1355795)

jkonecny12:
When existing partition was used then a bootloader saved this partition to other variable which wasn't tested.

Do a more robust fallback and also try to find disk from devicetree instead of KS command. Sorry I had to create ugly hack with ``dry_run`` to achieve this.

*Related: rhbz#1355795*

### DIFF
--- a/pyanaconda/kickstart.py
+++ b/pyanaconda/kickstart.py
@@ -420,10 +420,10 @@ class Bootloader(commands.bootloader.RHEL7_Bootloader):
                                            msg=(_("Requested boot drive \"%s\" doesn't exist or cannot be used.")
                                                 % self.bootDrive)))
         # Take valid disk from --driveorder
-        elif len(self.driveorder) >= 1:
+        elif len(valid_disks) >= 1:
             log.debug("Bootloader: use '%s' first disk from driveorder as boot drive, dry run %s",
-                      self.driveorder[0], dry_run)
-            self.bootDrive = self.driveorder[0]
+                      valid_disks[0], dry_run)
+            self.bootDrive = valid_disks[0]
         else:
             # Try to find /boot
             #
--- a/pyanaconda/kickstart.py
+++ b/pyanaconda/kickstart.py
@@ -312,6 +312,8 @@ class Bootloader(commands.bootloader.RHEL7_Bootloader):
     def __init__(self, *args, **kwargs):
         commands.bootloader.RHEL7_Bootloader.__init__(self, *args, **kwargs)
         self.location = "mbr"
+        self._useBackup = False
+        self._origBootDrive = None
 
     def parse(self, args):
         commands.bootloader.RHEL7_Bootloader.parse(self, args)
@@ -326,9 +328,29 @@ class Bootloader(commands.bootloader.RHEL7_Bootloader):
 
         return self
 
-    def execute(self, storage, ksdata, instClass):
+    def execute(self, storage, ksdata, instClass, dry_run=False):
+        """ Resolve and execute the bootloader installation.
+
+            :param storage: object storing storage-related information
+                            (disks, partitioning, bootloader, etc.)
+            :type storage: blivet.Blivet
+            :param payload: object storing packaging-related information
+            :type payload: pyanaconda.packaging.Payload
+            :param instclass: distribution-specific information
+            :type instclass: pyanaconda.installclass.BaseInstallClass
+            :param dry_run: flag if this is only dry run before the partitioning
+                            will be resolved
+            :type dry_run: bool
+        """
         if flags.imageInstall and blivet.arch.isS390():
             self.location = "none"
+
+        if dry_run:
+            self._origBootDrive = self.bootDrive
+            self._useBackup = True
+        elif self._useBackup:
+            self.bootDrive = self._origBootDrive
+            self._useBackup = False
 
         if self.location == "none":
             location = None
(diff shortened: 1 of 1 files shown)


pyanaconda/kickstart.py
```diff
@@ -326,10 +326,30 @@ def parse(self, args):
...
+        """ Resolve and execute the bootloader installation.
+
+            :param storage: object storing storage-related information
+                            (disks, partitioning, bootloader, etc.)
+            :type storage: blivet.Blivet
+            :param payload: object storing packaging-related information
+            :type payload: pyanaconda.packaging.Payload
+            :param instclass: distribution-specific information
+            :type instclass: pyanaconda.installclass.BaseInstallClass
+            :param dry_run: flag if this is only fry run before the partitioning
```
vpodzime: *fry* run? :)

pyanaconda/kickstart.py
```diff
@@ -397,25 +418,30 @@ def execute(self, storage, ksdata, instClass):
...
+            # the second will be used.
+            try:
+                boot_dev = storage.mountpoints["/boot"]
+                boot_drive = boot_dev.parents[0].name
+                if boot_drive and boot_drive in disk_names:
+                    self.bootDrive = boot_drive
+                    log.debug("Bootloader: use /boot partition '%s' as boot drive, dry run %s",
+                              boot_drive, dry_run)
+            except KeyError as e:
+                log.debug("Bootloader: /boot partition is not present, dry run %s", dry_run)
```
vpodzime: Only the ``storage.mountpoints["/boot"]`` line should be enclosed in the try-except block. The rest should go into the ``else`` branch of the block.

pyanaconda/kickstart.py
jkonecny12: :D good one

pyanaconda/kickstart.py
```diff
@@ -397,25 +418,30 @@ def execute(self, storage, ksdata, instClass):
...
-                boot_drive = disk_names[0]
-
-            self.bootDrive = boot_drive
+            #
+            # This method is executed two times. Before and after partitioning.
+            # In the first run, the result is used for other partitioning but
+            # the second will be used.
+            try:
+                boot_dev = storage.mountpoints["/boot"]
+                boot_drive = boot_dev.parents[0].name
```
vpodzime: The first parent of a device mounted at ``/boot`` doesn't necessarily have to be a disk. E.g. in case of RAID it would be a partition. You might need to go through the ``ancestors`` attribute here and grab the first disk from there.

pyanaconda/kickstart.py
jkonecny12: Good point, I'll change that.

pyanaconda/kickstart.py
```diff
@@ -2200,7 +2226,7 @@ def doKickstartStorage(storage, ksdata, instClass):
     # snapshot free space now so that we know how much we had available
     storage.createFreeSpaceSnapshot()
 
-    ksdata.bootloader.execute(storage, ksdata, instClass)
+    ksdata.bootloader.execute(storage, ksdata, instClass, dry_run=True)
```
vpodzime: Shouldn't this be added somewhere else with ``dry_run=False`` too?

jkonecny12: ``False`` is default value so only the ``True`` is required.

pyanaconda/kickstart.py
jkonecny12: So I should do it this way?
```
for parent in parent.parents[0]:
    if parent.isDisk():
    break
```

pyanaconda/kickstart.py
vpodzime: Sure, but where is the second call? It's already been there somewhere?

pyanaconda/kickstart.py
vpodzime: Not actually. There might potentially be even more levels of devices. You need to iterate over ``ancestors`` as mentioned above.

pyanaconda/kickstart.py
jkonecny12: Line 2238: ``storage.setUpBootLoader()``, the second call is from Blivet.

pyanaconda/kickstart.py
jkonecny12: Ahh sorry my mistake:

```
run = True
while run:
   parent = dev.parents[0]
   dev = parent
   if dev.isDisk():
      run = False
```

I really should look more on my posted comments...

jkonecny12:
Updated, @vpodzime is everything good now?

jkonecny12:
Updated again. 
I found a next small issue with ``driveorder`` fallback. Sorry for the ``ancestors`` attribute I was thickheaded.

pyanaconda/kickstart.py
```diff
@@ -396,26 +419,39 @@ def execute(self, storage, ksdata, instClass):
...
+                boot_drive = ""
+                # Find disk ancestor
+                for ancestor in boot_dev.ancestors:
+                    if ancestor.isDisk:
+                        boot_drive = ancestor.name
+                        break
+
+                if boot_drive and boot_drive in disk_names:
+                    self.bootDrive = boot_drive
+                    log.debug("Bootloader: use /boot partition '%s' as boot drive, dry run %s",
```
vpodzime: It's using the boot partition's *disk* as the boot drive.

vpodzime:

jkonecny12:
Pushed to RHEL7.
For master I'm waiting for Martin's patch.

jkonecny12:
New PR for master and f25 created. closing this one.

PR #707, opened 2016-07-20
Add support for device specification variants (#1200833)

M4rtinK:
Make it possible to have multiple device specification variants
delimited by | in the device specification string, for example:

"sd*|hd*|vda"

or

"sd*|hd*|vda,/dev/vdc"

The spec variants are processed individually from left to right
and at least one match needs to happen for the | delimited
spec to be considered successfully applied.

Also make sure that callers of the device_matches() function
really get disk devices when they expect them, not partitions
or other devices that are not disks.

And move the function to storage_utils.py - it really doesn't make
sense to have it in kickstart.py.

Resolves: rhbz#1200833

### DIFF
--- a/anaconda.spec.in
+++ b/anaconda.spec.in
@@ -82,7 +82,7 @@ The anaconda package is a metapackage for the Anaconda installer.
 %package core
 Summary: Core of the Anaconda installer
 Requires: python3-dnf >= %{dnfver}
-Requires: python3-blivet >= 1:2.0.2
+Requires: python3-blivet >= 1:2.1.2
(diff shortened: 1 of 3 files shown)


M4rtinK:
The Blivet side pull request required by this PR is https://github.com/rhinstaller/blivet/pull/464.

vpodzime:
Looks good to me.

M4rtinK:
The Blivet-side PR has been merged & I've just tested that Anaconda with this PR works with the respective Blivet branch. So time to push this. :)

PR #710, opened 2016-07-25
Clarify a nosave related log message (#1285519)

M4rtinK:
It is the nosave option that might prevent the output kickstart
from being written to the installed system.

Related: rhbz#1285519


jkonecny12:
Looks good to me.

PR #712, opened 2016-07-25
Fix retry to download local treeinfo file (#1270354)

jkonecny12:
When no repository is specified then we are trying /run/install/repo folder.

We were trying to download .treeinfo file from there and failed six times before giving up.
Now when .treeinfo got an error code which means local file can't be reached it will fail immediately.

*Related: rhbz#1270354*

*Reported-by: Peter Kotvan ``<pkotvan@redhat.com>``*

### DIFF
--- a/pyanaconda/packaging/__init__.py
+++ b/pyanaconda/packaging/__init__.py
@@ -460,11 +460,22 @@ class Payload(object):
                 except URLGrabError as ex:
                     log.info("Error downloading 'treeinfo': %s", ex)
 
-                    # Don't retry if we got a 'HTTP Error 404' from both reponses
-                    # from curl (likely not to be a transient error), just move on.
-                    if "404" in str(e) and "404" in str(ex):
+                    # HTTP error
+                    if e.errno == ex.errno == 14:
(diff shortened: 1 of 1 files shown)


pyanaconda/packaging/__init__.py
```diff
@@ -460,11 +460,22 @@ def _getTreeInfo(self, url, proxy_url, sslverify):
                 except URLGrabError as ex:
                     log.info("Error downloading 'treeinfo': %s", ex)
 
-                    # Don't retry if we got a 'HTTP Error 404' from both reponses
-                    # from curl (likely not to be a transient error), just move on.
-                    if "404" in str(e) and "404" in str(ex):
+                    # HTTP error
+                    if e.errno == 14 and ex.errno == 14:
```
rvykydal: You could use something like
e.errno == ex.errno == 14
but it's a matter of taste.

rvykydal:
Looks good to me.

pyanaconda/packaging/__init__.py
jkonecny12: Why not, it's smaller that way.

jkonecny12:
Pushed, thank you.

PR #713, opened 2016-07-25
Fix restart payload only when repo needs network (#1358788)

jkonecny12:
When user change some settings in a network then the payload manager will restart payload thread to test if repositories are still reachable.

Fix use-case when repositories are local so changing of network settings don't require to test repositories.

*Resolves: rhbz#1358788*

*Reported-by: Peter Kotvan ``<pkotvan@redhat.com>``*


rvykydal:
Looks good to me.

jkonecny12:
Pushed to RHEL but for master the fix must be changed.

PR #717, opened 2016-07-29
network: don't set empty team slave config for NM (#1254929)

rvykydal:
Related: rhbz#1254929

Seems that NetworkManager started to require non-empty team port configration
when creating team slave configuration.


jkonecny12:
Looks good to me.

PR #722, opened 2016-07-29
Fix reset playload when repositories requires internet

jkonecny12:
Backported from RHEL7.

### DIFF
--- a/pyanaconda/ui/gui/spokes/network.py
+++ b/pyanaconda/ui/gui/spokes/network.py
@@ -1526,8 +1526,8 @@ class NetworkSpoke(FirstbootSpokeMixIn, NormalSpoke):
 
         # if installation media or hdd aren't used and settings have changed
         # try if source is available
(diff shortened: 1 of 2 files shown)


rvykydal:
Looks ok to me.

pyanaconda/ui/gui/spokes/network.py
```diff
@@ -1526,8 +1526,8 @@ def apply(self):
 
         # if installation media or hdd aren't used and settings have changed
         # try if source is available
-        if self.data.method.method not in ["cdrom", "harddrive"] and self.networking_changed:
```
jkonecny12: I have removed this check because it's already contained in needsNetwork method and it also skipping additional repositories now.

rvykydal: Looks ok to me

jkonecny12:
Pushed, thank you.

PR #726, opened 2016-08-05
Fix needsNetwork() to test only enabled repositories (#1361213)

jkonecny12:
By testing every repository no groups for installation was loaded even when all enabled repositories were reachable.

Additional repositories which are not reachable are automatically disabled at this moment. This is original/unmodified behavior.

*Resolves: rhbz#1361213*
*Related: rhbz#1358788*

### DIFF
--- a/pyanaconda/packaging/__init__.py
+++ b/pyanaconda/packaging/__init__.py
@@ -239,6 +239,7 @@ class Payload(object):
 
     @property
     def needsNetwork(self):
+        """ Test base and additional repositories if they requires network. """
         url = ""
         if self.data.method.method == "nfs":
             # NFS is always on network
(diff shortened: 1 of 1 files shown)


pyanaconda/packaging/__init__.py
```diff
@@ -239,6 +239,7 @@ def _sourceNeedsNetwork(self, sources):
 
     @property
     def needsNetwork(self):
+        """ Test base and additional repositories if they requires network. """
```
vpodzime: *require*

vpodzime:
"..no groups *were loaded for installation*..."

vpodzime:
Looks good to me otherwise.

jkonecny12:
Thank you, I'll fix it locally.

jkonecny12:
Pushed

PR #727, opened 2016-08-05
Make sure unformatted DASDs are dasdfmted in ks. (#1359865)

snbueno:
make_unformatted_dasd_list takes a list of device names, not devices,
so make sure to pass that.

Resolves: rhbz#1359865


jkonecny12:
Looks good to me.

PR #728, opened 2016-08-09
Fix screenshot taking logic (#1327456)

M4rtinK:
Unlike before (RHEL ~7.1) Anaconda now uses a window stack
(GtkStack) which gereally works nicely but unfortunately
in some ways breaks the assumptions the original screenshot taking code made,
resulting with various issues, such as:

- multiple simultaniously registered key bindings
-> as all windows are now initialized even though just one can be actually
   visible at a time
- content of a wrong window might be saved as a the screenshot
-> all the key bindings fire at once when the key is pressed
-> the first binding to fire saves the screenshot, independently
   on the currently visible spoke

To fix this:
- we now register just one screenshot key binding - in MainWindow
- the screenshot taking logic has been moved to MainWindow
- the window on top of the window stack is used when taking a screenshot,
  making sure we really take a screenshot of what is visible on the screen

Resolves: rhbz#1327456

### DIFF
--- a/pyanaconda/ui/gui/__init__.py
+++ b/pyanaconda/ui/gui/__init__.py
@@ -39,10 +39,6 @@ log = logging.getLogger("anaconda")
 
 __all__ = ["GraphicalUserInterface", "QuitDialog"]
 
-_screenshotIndex = 0
-_last_screenshot_timestamp = 0
-SCREENSHOT_DELAY = 1  # in seconds
-
 ANACONDA_WINDOW_GROUP = Gtk.WindowGroup()
 
 # Stylesheet priorities to use for product-specific stylesheets and our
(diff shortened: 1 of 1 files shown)


jkonecny12:
Please fix these issues:
```
************* Module ../pyanaconda/ui/gui/__init__.py
W0602:563,8: MainWindow.take_screenshot: Using global for '_screenshotIndex' but no assignment is done
W0105:569,-1: MainWindow.take_screenshot: String statement has no effect
```

M4rtinK:
@jkonecny12 Should be fixed.

pyanaconda/ui/gui/__init__.py
```diff
@@ -585,6 +556,34 @@ def enlightbox(self, dialog):
 
         self.lightbox_off()
 
+    def _handle_prnt_screen(self, *args, **kwargs):
```
jkonecny12: Shouldn't be here ``print`` instead of ``prnt`` ?

jkonecny12:
It looks good to me otherwise.

pyanaconda/ui/gui/__init__.py
M4rtinK: Yeah, why not. :)

M4rtinK:
Fedora 25 PR:
https://github.com/rhinstaller/anaconda/pull/810

PR #729, opened 2016-08-10
Fix getting network device's hwaddress

rvykydal:
cherry-pick of the patch from master for bug 1364977


jkonecny12:
Looks good to me.

PR #730, opened 2016-08-11
network: don't require gateway for static ipv4 config in TUI (#1365532)

rvykydal:
Resolves: rhbz#1365532


jkonecny12:
Looks good to me.

PR #731, opened 2016-08-11
Fix anaconda-pre.service wasn't properly installed (#1255659)

jkonecny12:
Service wasn't in a Makefile.

Small improvement on Wants in a ``anaconda-pre.service``. It should be started before anaconda.service and other anaconda services.

The ``anaconda-pre-log-gen`` was in ``/usr/libexec`` instead of ``/usr/bin``.
The ``dmesg`` command in ``anaconda-pre-log-gen`` doesn't have -a parameter.

*Related: rhbz#1255659*


M4rtinK:
Looks good to me! :+1:

PR #733, opened 2016-08-11
Fix reset payload only on network change (#1364367)

jkonecny12:
Reset payload when connection change.
This solve the issue only partially, problem here is that the ``NetworkManager-connection-editor`` will save the connection in other form than it is saved from the ``Dracut``.

So when you click on the save button (without changing anything) for the first time, the payload will be restarted but when you do it more times after, it won't restart the payload thread anymore.

*Resolves: rhbz#1364367*
*Reported-by: Jakub Vavra ``<jvavra@redhat.com>``*


jkonecny12:
I updated this PR, this solution is slightly better then the previous one but behavior is still the same.

PR #734, opened 2016-08-11
Disable strict password quality checking for partial kickstart instal…

M4rtinK:
…lations

We already disable strict password quality checking in fully interactive
installations (no user provided kickstart), so do the same thing by default
for partial kickstart installations.

Also clarify how the default password policy is set and when it actually applies.

Related: rhbz#1360263


jkonecny12:
Looks good to me.

PR #735, opened 2016-08-11
Run cache hack after moving cache to mounted storage (#1287438)

bcl:
Yum doubles up the /mnt/sysimage on the cache dir if we don't run the
cache hack on it. We DO want it to be under /mnt/sysimage, but not under
/mnt/sysimage/mnt/sysimage/

Without this yum creates the cache at the wrong place, and then when
anaconda-yum runs it uses the right place, leaving a
/mnt/sysimage/mnt/sysimage/var/tmp/yum.cache on the installed system.

Resolves: rhbz#1287438


jkonecny12:
Looks good to me.

PR #736, opened 2016-08-12
Master fix net reset payload

jkonecny12:
Improve connection network change detection

I'm reverting old unstable solution on master and introducing new solution based on https://github.com/rhinstaller/anaconda/pull/733 .


rvykydal:
Looks good to me.

PR #737, opened 2016-08-16
Add centos module to pyanaconda

jaymzh:
No reason for this to be patched in by centos or anyone else,
that's why it's modular!

### DIFF
--- a/pyanaconda/installclasses/centos.py
+++ b/pyanaconda/installclasses/centos.py
@@ -1,0 +1,64 @@
+#
+# centos.py
+#
+# Copyright (C) 2010  Red Hat, Inc.  All rights reserved.
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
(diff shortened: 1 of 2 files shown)


jkonecny12:
Hello @jaymzh , 

I don't know exactly which CentOS version you are targeting but in case you don't know this workflow. 
If you want this patches to CentOS 7, you need to create this PRs for the rhel7-branch too and also create bugs in https://bugzilla.redhat.com/ to get this patches to a RHEL release. 
All the RHEL commits then must have ``Related: rhbz#<bz-number>`` or ``Resolves: rhbz#<bz-number>`` in the commit message.

Thank you for your patches.

jaymzh:
@jkonecny12  - we only need it in master. This is part of the work we're doing with Redhat to get master to be able to install older versions. You can talk to @vathpela or @dcantrell about it for extra context.

jkonecny12:
Ahh OK, then you can ignore it ;)

snbueno:
This looks ok to me, but I want at least one other person on the team to add their ack as well.

PR #739, opened 2016-08-17
Add option to override efi_dir

jaymzh:
This is because in CentOS6, the efi_dir is actually 'redhat', not 'centos'.

This is on top of https://github.com/rhinstaller/anaconda/pull/738 and https://github.com/rhinstaller/anaconda/pull/737

### DIFF
--- a/pyanaconda/bootloader.py
+++ b/pyanaconda/bootloader.py
@@ -1677,7 +1677,10 @@ class GRUB2(GRUB):
 class EFIBase(object):
     @property
     def _config_dir(self):
-        return "efi/EFI/%s" % (self.efi_dir,) # pylint: disable=no-member
+        efi_dir = self.efi_dir
+        if flags.cmdline.get("force_efi_dir") is not None:
+            efi_dir = flags.cmdline.get("force_efi_dir")
+        return "efi/EFI/%s" % (efi_dir,) # pylint: disable=no-member
 
     def efibootmgr(self, *args, **kwargs):
         if flags.imageInstall or flags.dirInstall:
@@ -1788,6 +1791,8 @@ class EFIGRUB1(EFIBase, GRUB):
     def __init__(self):
         super(EFIGRUB1, self).__init__()
         self.efi_dir = 'BOOT'
+        if flags.cmdline.get("force_efi_dir") is not None:
+            self.efi_dir = flags.cmdline.get("force_efi_dir")
 
     #
     # configuration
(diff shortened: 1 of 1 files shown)


M4rtinK:
I've added the patch with and updates image on top of Fedora Rawhide boot ISO from August 22, then booted it in efi mode in a VM and added the legacygrub=1 option to the boot command line.

Unfortunately it crashes when it tries to write the bootloader:

```
anaconda 25.20-1 exception report
Traceback (most recent call first):
  File "/tmp/updates/pyanaconda/bootloader.py", line 1219, in write_device_map
    dev_map = open(map_path, "w")
  File "/tmp/updates/pyanaconda/bootloader.py", line 1259, in write_config
    self.write_device_map()
  File "/tmp/updates/pyanaconda/bootloader.py", line 1764, in write
    self.write_config() # pylint: disable=no-member
  File "/tmp/updates/pyanaconda/bootloader.py", line 2433, in writeBootLoaderFinal
    storage.bootloader.write()
  File "/tmp/updates/pyanaconda/bootloader.py", line 2511, in writeBootLoader
    writeBootLoaderFinal(storage, payload, instClass, ksdata)
  File "/tmp/updates/pyanaconda/install.py", line 252, in doInstall
    writeBootLoader(storage, payload, instClass, ksdata)
  File "/usr/lib64/python3.5/threading.py", line 862, in run
    self._target(*self._args, **self._kwargs)
  File "/tmp/updates/pyanaconda/threads.py", line 251, in run
    threading.Thread.run(self, *args, **kwargs)
FileNotFoundError: [Errno 2] No such file or directory: '/mnt/sysimage/boot/efi/EFI/fedora/device.map'

Local variables in innermost frame:
self: <pyanaconda.bootloader.EFIGRUB1 object at 0x7fa88f6cd0f0>
map_path: /mnt/sysimage/boot/efi/EFI/fedora/device.map
```

Is there something else I need to do to make the installation successful ? Also if needed I can supply logs from that installation run.

jaymzh:
EFIGRUB1 assumes a lot of things I can't imagine will work on a modern Fedora system. I'm guessing the bootmap isn't actually there these days. What did you pass in with `inst.force_efi_dir` on the kernel command line? To test this works, set it to something like 'foobar' and watch `/mnt/sysimage/boot/efi/EFI/fedora/device.map` turn into `/mnt/sysimage/boot/efi/EFI/foobar/device.map`.

This serious of patches is to enable anaconda master to be able to install a CentOS6 (or a RHEL6) system.

M4rtinK:
> EFIGRUB1 assumes a lot of things I can't imagine will work on a modern Fedora system. I'm guessing the bootmap isn't actually there these days. What did you pass in with inst.force_efi_dir on the kernel command line? To test this works, set it to something like 'foobar' and watch /mnt/sysimage/boot/efi/EFI/fedora/device.map turn into /mnt/sysimage/boot/efi/EFI/foobar/device.map.

Yep, that does happen, even though the crash still happens. :)

> 
> This serious of patches is to enable anaconda master to be able to install a CentOS6 (or a RHEL6) system.

Yeah, I understand that specific circumstances are needed to make use of this functionality. I just wanted to test it a bit locally. Also no Fedora code is making direct use of the EFI + GRUB1 functionality, so I guess is should not be an issue when it does not work during a normal Fedora installation run.

jaymzh:
> Also no Fedora code is making direct use of the EFI + GRUB1 functionality, so I guess is should
> not be an issue when it does not work during a normal Fedora installation run.

Yeah there's no reason for modern systems to ever touch that code. I imagine making Grub1 work with modern Fedora would take a lot more work than just in Anaconda :)

M4rtinK:
> Yeah there's no reason for modern systems to ever touch that code. I imagine making Grub1 work with modern Fedora would take a lot more work than just in Anaconda :)

Indeed! So lets merge this PR. :)

PR #742, opened 2016-08-19
Ensure LDL DASDs formatted in text ks. (#1259437)

snbueno:
Similar to 9840fd7, this patch makes sure to create a list of LDL
DASD names, not a list of LDL DASD device objects, since the dev
name must be passed to format_dasd.

Resolves: rhbz#1259437


vpodzime:
Looks good to me.

PR #749, opened 2016-08-24
don't use `systemctl --root` for now (#1369794)

AdamWill:
instead just run in a chroot (via execWithRedirect), because
`systemctl --root` will not work for non-systemd services at
present.

### DIFF
--- a/pyanaconda/iutil.py
+++ b/pyanaconda/iutil.py
@@ -758,10 +758,7 @@ def _run_systemctl(command, service, root="/"):
     """
 
     args = [command, service]
-    if root != "/":
-        args += ["--root", root]
-
-    ret = execWithRedirect("systemctl", args)
+    ret = execWithRedirect("systemctl", args, root=root)
 
     return ret
 


AdamWill:
this is mainly intended as a hotfix for 1369794 for F25 Alpha. Ideally we'd make chkconfig work with `--root`, then this could go back to the way it was.

AdamWill:
https://pagure.io/fedora-kickstarts/pull-request/52 is the alternative for this; working around the bug in kickstarts.

cgwalters:
I would have mentioned this came from  ec3ddc997fb854883b362e02a56e204e9c49eac6 - but it looks like an innocent side effect, we're unlikely to regress anything.  So LGTM.

AdamWill:
we have worked around this in kickstarts for now, so it's not urgent. But we should really either do this or get `chkconfig` fixed - https://bugzilla.redhat.com/show_bug.cgi?id=1369916

jkonecny12:
Jenkins, it's ok to test.

jkonecny12:
Hi @AdamWill. 
It looks to me like regression which should be properly fixed on the chkconfig or systemd side but as hotfix I think we can go with this.
If you want this hotfix only for Fedora 25 and proper fix will go to master later you need to create the PR against f25-release. If you want this hotfix for Fedora 25 and master create the PR for f25-devel. I personally think it would be better to have it in master and Fedora 25 and do a revert in master after proper fix will arrive to Rawhide.

If you want I can recreate the PR for you.

AdamWill:
@jkonecny12 done. Note: I don't think this is actually a 'regression' so far as `chkconfig` is concerned, I don't think `chkconfig` has *ever* supported this. systemd actually leaves the implementation of the fallback as something for distributions to sort out; upstream systemd only provides a sample called `systemd-sysv-install.SKELETON` and tells distributors to provide a real implementation for their distribution. So it's a Fedora/RHEL decision that we simply make `systemd-sysv-install` a symlink to `chkconfig`.

jkonecny12:
Thank you for explanation. Looks good to me.

snbueno:
NAK from me. The temp fix lives in the ks scripts for now -- why not leave it there and get bug 1369916 fixed properly instead?

cgwalters:
Yes, this isn't blocking anything now, and we can probably move forward without this patch.  However, in the end, there are so many interacting components in the anaconda/systemd/network/storage etc. universe that I believe it'd be a lot more efficient if we were more willing to carry some slight ugliness in one project when an apparently innocent cleanup patch does happen to trigger another bug.

AdamWill:
well, the kickstart workaround only fixes *our image building kickstarts*. if any users are out there trying to configure non-systemd services with kickstarts, it doesn't help them at all. Especially if 1369916 happened not to get fixed before F25 goes out, that would be a shame.

snbueno:
@AdamWill, that sounds like great justification to make 1369916 a release blocker. ;) And you're right about users, but this is again where I need to ask: who is our target audience? I don't want to accept things on a whim that somebody may someday want to do a very specific something, especially if the cost to us is basically a hack. How common is that anyway? There aren't that many non-systemd services left IIRC, especially among frequently used ones (network aside).

@cgwalters, we already, unfortunately, carry quite a lot of ugliness that was at one time just something small. I don't want anaconda to be a dumping ground for hacks any more than it already is, especially when a real fix is pretty clear cut. That's how we end up with things like rhel5-branch and rhel6-branch, which eventually makes our code very difficult to understand and continue working on.

AdamWill:
indeed there aren't many non-systemd services left, but `network` is a pretty huge exception, and I'd guess quite confidently that 'enable network service out of the box' is a fairly common use case (given that it's still necessary for some things; heck, I have a case myself - the openQA boxes, which use the network service's support for openvswitch interfaces - only I happen to use an ansible play to enable it for that case, rather than an install kickstart). And I wouldn't be surprised if there are still quite a few sites left installing and enabling custom services as sysv rather than systemd services.

snbueno:
Right, I said "network aside" -- I think it's been pretty clearly established that one needs to be fixed.

As for other legacy services, "quite a few" -- I'm sorry, but conjecture isn't going to convince me to accept a hack. Most people running a production server don't move to the latest and greatest right away, unless there's a very specific reason (necessary functionality; security; serious bugs that impacted them now fixed; etc). And how many of them do a fresh install as opposed to an upgrade? When you consider that, the pool gets pretty limited. Those people can file bugs for these less common cases, and we can fix them as appropriate or redirect them to the proper components then.

rvykydal:
Closing the NACKed PR.
Based also on Lukas's comment in
https://bugzilla.redhat.com/show_bug.cgi?id=1369794#c26
we should be able to address the issue by resolving https://bugzilla.redhat.com/show_bug.cgi?id=1369916

PR #751, opened 2016-08-25
Translate press-c-to-continue correctly in TUI (#1364539)

M4rtinK:
The press-c-to-continue shortcut has been partially translated
on some screens in the TUI. It has been translated in the prompt,
but a plan "c" character was still expected, regardless on what has been
shown to the user.

In some cases (such as in the storage configuration spoke) the user
could be instructed to press "p" for "Pokračovat" (in Czech language
installation), but TUI would expect "c" to be pressed.

This has been fixed and TUI should now always ask for the same
character it tells the user to input.

Resolves: rhbz#1364539

### DIFF
--- a/pyanaconda/rescue.py
+++ b/pyanaconda/rescue.py
@@ -300,7 +300,8 @@ class RootSpoke(NormalTUISpoke):
         try:
             keyid = int(key) - 1
         except ValueError:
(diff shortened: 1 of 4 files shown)


jkonecny12:
Looks good to me.

PR #137, opened 2015-06-12
Fix nfs4 stage2 and repo handling

bcl:

PR #752, opened 2016-08-25
network: don't stumble upon new Device.Statistics NM dbus iface (#137…

rvykydal:
…0099)

New org.freedesktop.NetworkManager.Device.Statistics inteface has been added to
NM.  We stumble upon it when looking for device type specific interface.

Resolves: rhbz#1370099


snbueno:
Since testing confirms the fix works, ACK.

PR #753, opened 2016-08-25
Don't offer encrypting for Atomic Host installclass (#1267905)

rvykydal:
Atomic Host does not include cryptsetup.

Resolves: rhbz#1267905

This is "work in progress" patch by Chris. To me it seems quite complete (ie ACK). I rebased it for top of rhel7-branch and tested with locally built RHELAH 7.3 installer. Also tested that it does not interfere with mainline installer (GUI and kickstart).

The only thing that came up to me is that we might rather want to make the "Encrypt" checkboxes insensitive instead of hiding them, but I don't have strong preference.

### DIFF
--- a/pyanaconda/installclass.py
+++ b/pyanaconda/installclass.py
@@ -52,6 +52,9 @@ class BaseInstallClass(object):
     # install the latest updates during installation source selection.
     installUpdates = True
 
+    # Do we want to present the option to encrypt devices?
+    offerEncryption = True
+
     _l10n_domain = None
 
     # The default filesystem type to use.  If None, we will use whatever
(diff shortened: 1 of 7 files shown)


jkonecny12:
As far as I understand, it looks good to me.

jkonecny12:
There is a style guide check error:
```
Bad string filesystem found at pyanaconda/kickstart.py:298 pyanaconda/kickstart.py:1080 pyanaconda/kickstart.py:1401 pyanaconda/kickstart.py:1595. Try file system instead.
```

vpodzime:
Looks good to me.

rvykydal:
Closing as cryptsetup will be added to 7.3 Atomic Host

PR #740, opened 2016-08-18
Replace deprecated method of DNF

j-mracek:
`dnf.callback.LoggingTransactionDisplay is deprecated now. It was considered part of API despite the fact that it has never been documented. Use `dnf.callback.TransactionProgress` instead.

PR #755, opened 2016-08-26
Apply replacement of DNF deprecated method

jkonecny12:
First commit is taken from PR https://github.com/rhinstaller/anaconda/pull/740 .
Add fix on top of this commit and change the installation progress loop to more readable code.


vpodzime:
Looks good to me.

j-mracek:
I think that there should be no problem from dnf side.

jkonecny12:
Thank you @j-mracek and @vpodzime

PR #741, opened 2016-08-19
Provides compatibility with DNF-2.0

j-mracek:
We try to help dnf api users to make kompatible theirs product with upcoming dnf-2.0. With this major release there are some api changes in dnf.

It would be also nice to mark anaconda without the patch as dnf-2.0.0 incompatible.

PR #759, opened 2016-08-30
Actual Anaconda is not compatible with DNF 2.0.0

jkonecny12:
Based on comment https://github.com/rhinstaller/anaconda/pull/741#issue-172130581,  I'm marking actual Anaconda as DNF 2.0 incompatible.

### DIFF
--- a/anaconda.spec.in
+++ b/anaconda.spec.in
@@ -21,6 +21,7 @@ Source0: %{name}-%{version}.tar.bz2
 %define gettextver 0.19.8
 %define pykickstartver 2.31-1
 %define dnfver 0.6.4
+%define dnfmaxver 2.0.0
 %define partedver 1.8.1
 %define pypartedver 2.5-2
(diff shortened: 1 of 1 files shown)


j-mracek:
Super, but if it will go into master branch, the patch is incompatible with #741 . Both patches together makes anaconda unistallable.

jkonecny12:
I know, but as I understand it DNF 2.0 is not in rawhide now, so when 2.0 reach rawhide we need to merge the PR https://github.com/rhinstaller/anaconda/pull/741 .
If we merge it before it will break all rawhide builds.

j-mracek:
Yes, you are completely right :+1:

M4rtinK:
Looks good to me.

jkonecny12:
I had this merge to master and it should be to f25-devel. This was fixed in git by merging this branch to f25-devel too.

PR #761, opened 2016-08-30
kickstart: Drop almost all `timezone` verb -> chrony bridging

cgwalters:
Previously, the `timezone` verb would e.g. automatically install
and enable chrony if `timezone --nontp` wasn't passed.  Since
Atomic Host switched to timesyncd, this caused the installer to
fail.

The other Fedora variants already have chrony installed and enabled,
so this shouldn't affect them.  The only impact I can think of is on
people doing custom kickstart installs, but they can really quite
easily pick/enable their own time service.

The only behavior we do explicitly retain is that if `timezone
--nontp` is specified (or the "disable NTP" box in the UI is checked),
we disable the chrony service.  That's bug 1135768.  I didn't see any
other likely regressions in the commits/bugzillas I found via git
annotate.

Resolves: rhbz#1331864

### DIFF
--- a/pyanaconda/kickstart.py
+++ b/pyanaconda/kickstart.py
@@ -1705,12 +1705,11 @@ class Timezone(commands.timezone.F25_Timezone):
     def __init__(self, *args):
         commands.timezone.F25_Timezone.__init__(self, *args)
 
-        self._added_chrony = False
-        self._enabled_chrony = False
-        self._disabled_chrony = False
-
     def setup(self, ksdata):
-        # do not install and use NTP package
+        # This code *used* to try to manage installation of chrony and
+        # the service enablement state, but no longer does so.
+        # For more information see:
+        # https://bugzilla.redhat.com/show_bug.cgi?id=1331864
         if self.nontp or NTP_PACKAGE in ksdata.packages.excludedList:
             if iutil.service_running(NTP_SERVICE) and \
                     can_touch_runtime_system("stop NTP service"):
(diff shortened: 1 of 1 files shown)


vpodzime:
I'm sorry, but I don't buy this. Saying that all Fedora variants install *chrony* by default **right now** is easy, but we can hardly base the code on such assumptions. This needs a more systematic solution. Maybe AH should blacklist *chrony* and this code should be made more clever to fallback to *timesyncd*? Or should we add a special configuration attribute to installclasses for NTP daemon?

cgwalters:
At some point I could see changing this code to optionally write a timesyncd config file.  But right now, this is a blocker for Fedora 25 AH, and I don't want to increase the scope too much.

cgwalters:
Basically, I am relatively confident[1] this won't regress anything.  The service enablement state is also managed via systemd presets, and chrony is [in the default preset](https://pagure.io/fedora-release/blob/master/f/90-default.preset#_20).

[1] I didn't actually test a Server/Workstation install though, just AH right now

M4rtinK:
Yeah, I also think the pull request is too invasive - especially if the potentially breaking change has not been tested outside of Atomic Host.

I second Vratislavs suggestion - blacklisting the "chrony" package in Atomic host for now and working towards a systematic solution (like NTP daemon configuration via the (Atomic Host) install class) that fixes the concrete issue Atomic Host is having without breaking anything else.

cgwalters:
I'm almost certain it's going to work fine, but it's not worth a battle right now.  We can revisit this when considering changing multiple editions at once.

https://lists.projectatomic.io/projectatomic-archives/atomic-devel/2016-September/msg00000.html

PR #762, opened 2016-08-30
Rawhide build improvements

M4rtinK:
A scratch-build made by the modified build scripts was successful, so time to send this for review. :)

In short this:
- fixes some static checker reported issues
- switches to argparse from optparse
- autodetects package name, version and bug reporting email if not provided
- makes it possible to provide Zanata branch name & automatically maps the "unstable" git branch to the "master" Zanata branch
- ignores GitHub PR merge commits when generating the changelog

### DIFF
--- a/scripts/makebumpver
+++ b/scripts/makebumpver
@@ -26,7 +26,7 @@ log.setLevel(logging.INFO)
 import bugzilla
 from contextlib import closing
 import datetime
-import getopt
+import argparse
 import getpass
 import json
 import os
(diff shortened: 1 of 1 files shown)


scripts/makebumpver
```diff
@@ -348,6 +348,7 @@ class MakeBumpVer:
         git_range = "%s-%s-%s.." % (self.name, self.version, self.release)
         proc = run_program(['git', 'log', '--pretty=oneline', git_range])
         lines = filter(lambda x: x.find('l10n: ') != 41 and
+                       x.find('Merge pull request') != 41 and
                        x.find('Merge commit') != 41 and
                        x.find('Merge branch') != 41,
```
vpodzime: Wouldn't it be better to just let git ignore all merge commits by using the ``--no-merges`` option?

vpodzime:
Looks good to me apart from the above suggestion.

M4rtinK:
@vpodzime Good point, that indeed works fine & looks much better, thanks! :) I've removed the whole filter block as we no longer do any l10n commits (I guess that was from back when the translations were stored in git).

scripts/makebumpver
```diff
@@ -36,16 +36,116 @@ import sys
...
+                                                     "Also runs some checks such as ensuring rhel*-branch commits correctly "
+                                                     "reference RHEL bugs.",
+                                         epilog="The -i switch is intended for use with utility commits that we do not need to "
+                                                "reference in the spec file changelog.\n"
+                                                "The -m switch is used to map a Fedora BZ number to a RHEL BZ number for "
+                                                "the spec file changelog.\n"
+                                                "Use -m if you have a commit that needs to reference a RHEL bug and have cloned "
+                                                "the bug, but the original commit was already pushed to the central repo.")
+
+        parser.add_argument("-n", "--name", dest="name", default=current_name, metavar="PACKAGE NAME", help="Package name.")
```
jkonecny12: I think this is some copy & paste error. From what I see here the two commands above are the same as two commands above them with the same comment, or am I missing something?

scripts/makebumpver
```diff
@@ -36,16 +36,116 @@ import sys
...
+
+        # gather all unprocessed command line arguments
+        parser.add_argument(nargs=argparse.REMAINDER, dest="unknown_arguments")
+
+        self.args = parser.parse_args()
+        if self.args.debug:
+            log.setLevel(logging.DEBUG)
+
+        if self.args.unknown_arguments:
+            sys.stderr.write("unknown arguments:%s" % self.args.unknown_arguments)
```
jkonecny12: I think it would be nice to print help here with some empty lines between error and help.

scripts/makebumpver
```diff
@@ -484,8 +577,8 @@ class MakeBumpVer:
                         if job["name"] == targetJob:
                             ret = job["color"] == "blue"
                             break
-        except IOError as e:
-            log.error("Jenkins check failed: %s", e)
+        except IOError:
+            log.exception("Jenkins check failed.")
```
jkonecny12: I think it would be better to use exclamation mark here.

scripts/makebumpver
M4rtinK: Indeed- it's there twice! :) Removing it now.

scripts/makebumpver
M4rtinK: Yeah, why not. :)

scripts/makebumpver
M4rtinK: Done. :)

jkonecny12:
Looks good to me now. Thank you for the fixes.

jkonecny12:
Ups bad button :D .

PR #764, opened 2016-08-31
network: don't show ibft configured devices in UI (#1309661)

rvykydal:
Resolves: rhbz#1309661

In UI we don't show devices having special connection created by NM iBFT plugin
from iBFT. NM is using the connection for the device (when taking it over from
dracut) if it matches device configuration created (also from iBFT) in dracut.
The matching is broken for static configuration, so NM just creates normal
connection for the device (ie not using the connection created by iBFT plugin
based on iBFT table). This does not prevent successful installation but a
device with such configuration is not filtered out for our UI. So take care
also of this case as we don't want to expose (and allow modification of)
configuration from iBFT in UI.

In text mode we also mistakenly used to generate kickstart commands for devices
configured from ibft which is fixed by this patch as well.

### DIFF
--- a/pyanaconda/network.py
+++ b/pyanaconda/network.py
@@ -45,7 +45,7 @@ from pyanaconda import nm
 from pyanaconda import constants
 from pyanaconda.flags import flags, can_touch_runtime_system
 from pyanaconda.i18n import _
-from pyanaconda.regexes import HOSTNAME_PATTERN_WITHOUT_ANCHORS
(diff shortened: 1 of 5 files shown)


vpodzime:
Looks good to me.

jkonecny12:
Looks good to me also, but Jenkins don't think it is :)

```
************* Module regex_tests/ibft_device_name_test.py
W0611: 23,0: : Unused import re
```

rvykydal:
Updated to fix pylint issue.

jkonecny12:
Looks even better now :). ACK from me.

PR #766, opened 2016-08-31
Fix multi-inheritance

jaymzh:
In order for cooperative multi-inheritance to work properly,
all classes in the tree have to call super() in their __init__

### DIFF
--- a/pyanaconda/bootloader.py
+++ b/pyanaconda/bootloader.py
@@ -238,6 +238,7 @@ class BootLoader(object):
     _trusted_boot = False
 
     def __init__(self):
+        super(BootLoader, self).__init__()
         self.boot_args = Arguments()
         self.dracut_args = Arguments()
 
@@ -1677,6 +1678,7 @@ class GRUB2(GRUB):
 class EFIBase(object):
 
     def __init__(self):
+        super(EFIBase, self).__init__()
         self.efi_dir = None
 
     @property


jkonecny12:
Jenkins, it's ok to test.

vpodzime:
Are you sure this is the issue and the right solution? I've never seen an issue like this and I think we have multiple inheritance in many other places none of them calling ``object.__init__()``

jaymzh:
Yes, I'm very sure. This is how python works - it'll only call the matching method in the first parent-class that matches. Here:

```
class First(object):
    def __init__(self):
        print("first")
                

class Second(object):
    def __init__(self):
        print("second")
                

class Third(First, Second):
    def __init__(self):
        super(Third, self).__init__()
        print("third")
                
Third()
```

gives you:

```
In [6]: Third()
first
third
```

But:

```

class First(object):
    def __init__(self):
        super(First, self).__init__()
        print("first")
                

class Second(object):
    def __init__(self):
        super(Second, self).__init__()
        print("second")
                

class Third(First, Second):
    def __init__(self):
        super(Third, self).__init__()
        print("third")

Third()
```

gives you:

```
In [10]: Third()
second
first
third
```

Also... the breakage people reported with fc26, we saw, but with this patch it works.

vpodzime:
I so much prefer the solution where the class using mutli-inheritance calls ``__init__()`` of the parent classes manually. This is really weird and the order above (second, first, third) is totally counter-intuitive.

vpodzime:
> I so much prefer the solution where the class using mutli-inheritance calls __init__() of the parent classes manually. This is really weird and the order above (second, first, third) is totally counter-intuitive.

And that's what we do in many other places.

jaymzh:
This is the standard python way of ensuring all subclasses call both trees completely all the way up. And it's done elsewhere in this file as well.

vpodzime:
> This is the standard python way of ensuring all subclasses call both trees completely all the way up. And it's done elsewhere in this file as well.

Don't see it anywhere else in combination with multi-inheritance.

jaymzh:
I'm adding Bootloader and EFIBase. It was already there for GRUB, GRUB2, EFIGRUB1, EFIGRUB, and ZIPL.

jkonecny12:
From my observations it shouldn't break anything and personally I like this solution. I'm giving you an ACK.
Thank you for finding this bug and for fixing it. Good job @jaymzh !

PR #768, opened 2016-09-01
network: set onboot correctly for vlan on bond device in ks (#1234849)

rvykydal:
Related: rhbz#1234849


jkonecny12:
Looks good to me.

PR #770, opened 2016-09-02
Drop T from cp command flags when copying driver disk RPMs (#1269915)

M4rtinK:
We are copying the driver disk RPMs to a directory, but cp used the T
flag, which corresponds to:

-T, --no-target-directory
       treat DEST as a normal file

As the destination should always be a directory the T flag doesn't make
sense there. And indeed, if T is removed from the cp flags the driver disk
RPM is copied correctly.

Resolves: rhbz#1269915


vpodzime:
Looks good to me.

jkonecny12:
Could you please look on the master code if this patch should be there too?

Looks good to me too.

M4rtinK:
@jkonecny12 Sure, that was the plan from the start: https://github.com/rhinstaller/anaconda/pull/772

PR #772, opened 2016-09-05
Drop T from cp command flags when copying driver disk RPMs (#1269915)

M4rtinK:
We are copying the driver disk RPMs to a directory, but cp used the T
flag, which corresponds to:

-T, --no-target-directory
       treat DEST as a normal file

As the destination should always be a directory the T flag doesn't make
sense there. And indeed, if T is removed from the cp flags the driver disk
RPM is copied correctly.

Resolves: rhbz#1269915


jkonecny12:
Looks good to me.

M4rtinK:
I've actually discovered some cases this PR does not cover & some more improvements that can be made. So I'm closing this PR as obsolete & will make a new one with the improved version.

PR #773, opened 2016-09-06
Improved driver disk copying (#1269915)

M4rtinK:
Turns out that just changing flags for the cp command won't do
as function needs to support both single file driver disk RPM copy
as well as copying all driver disk RPMs from a given folder.

So let's drop the cp command all together and replace it with Python shutil
based equivalent. As far as I can tell it seams that there should be
no sub-folders in a driver disk RPM repo, so don't copy them
but just log a warning instead.

Also improve the copy-RPM unit tests to cover more scenarios and
add add more logging in the stage2 driver disk package handling code.

Related: rhbz#1269915

### DIFF
--- a/dracut/driver_updates.py
+++ b/dracut/driver_updates.py
@@ -72,6 +72,7 @@ import os
 import subprocess
 import fnmatch
 import readline # pylint:disable=unused-import
+import shutil
 
 from contextlib import contextmanager
 from logging.handlers import SysLogHandler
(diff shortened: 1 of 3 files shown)


jkonecny12:
I don't think we want to add new dependency to ``shutil``. It will be grabbed to initrd and as Will wrote:

>   **Use as little Python as possible in initrd**
>
>   If you know Python better than dracut or C, it might be tempting to solve problems by hacking on the Python parts of the DUD code. This leads to redundant code - and more code for us to maintain.
>
>   `driver_updates.py`'s purpose is basically just to mount disks or disk images and find valid DUD repos therein (either automatically or interactively), and to save needed data to places where the installer can find it later.
>
>   *Any other tasks should be handled in dracut scriptlets, the C utilities, or the main installer program.*
>
>    (If you must add Python code, avoid pulling in new libraries. It makes building images much more fragile and bloats the initramfs.)

rvykydal:
We are already pulling shutil in via parse-kickstart:
https://github.com/rhinstaller/anaconda/blob/rhel7-branch/dracut/parse-kickstart#L472
OTOH it is the only occurrence I was able to find so we could also remove it from the parse-kickstart.

M4rtinK:
I could change it to two separate cp command calls (one with the T flag for folders and one without the T flag for files), but as this bug shows it's (at least IMHO) a lot less robust than the python-based copying functions.

vpodzime:
Looks good to me and I don't think trying to remove ``shutil`` from initrd is worth the struggle.

M4rtinK:
Fixed the potential issue reported by Pylint.

M4rtinK:
Fedora PR:
https://github.com/rhinstaller/anaconda/pull/811

PR #775, opened 2016-09-07
Revert "Show network spoke in the TUI reconfig mode (#1302165)"

M4rtinK:
Turns out the Network Spoke has some issues with running in Initial
Setup TUI, so revert this for now.

We will revisit it later once we are reasonably sure Network Spoke
can work correctly in Initial Setup.

This reverts commit f2e0ea93130f821138441938d1a1463040315271.


jkonecny12:
Looks good to me.

PR #777, opened 2016-09-08
It allows to always mark one env_group as installed

j-mracek:
It prevents if user install other env_group and later on try to remove it that
dnf will not try to remove 'core' group that will be required by
minimall-install group.

It will not install any additional packages, only it mark istead of only core
group installed, additional env_group minimal_install that has only one
mandatory group - core.

### DIFF
--- a/pyanaconda/packaging/dnfpayload.py
+++ b/pyanaconda/packaging/dnfpayload.py
@@ -345,8 +345,8 @@ class DNFPayload(packaging.PackagePayload):
             log.info("skipping core group due to %%packages --nocore; system may not be complete")
         else:
             try:
-                self._select_group('core', required=True)
-                log.info("selected group: core")
+                self._select_group('minimal-environment', required=True)
+                log.info("selected group: minimal-environment")
             except packaging.NoSuchGroup as e:
                 self._miss(e)
 


j-mracek:
It is related to RhBug:1359482

jkonecny12:
I don't think it is a good idea. You are changing default behavior for kickstart installation. Default environment is set by ``installclass`` class in Anaconda or from other package which is used when ``boot.iso`` is created. 
The ``boot.iso`` I gave you, doesn't have an ``installclass`` so it doesn't have default environment set. However it should be fine in Fedora compose.

j-mracek:
I just want to ensure that there will be impossible to install system only with group 'core' but always with env_group that could have only core group inside (only minor difference). Anyway there are not many users that plays with env_groups, therefore the chance appearance of problem is low. I just was surprised by code work-flow. It first marks for install 'core' group and than it marks env_group that is chosen for install that again contains core group. I would say that the first step is redundant. But I don't know, how it works in other than Fedora composes.

jkonecny12:
I'm not quite sure that every user of Anaconda want this behaviour. When there will be user request for this I'm happy to push the PR but for now I'm closing it.

PR #778, opened 2016-09-08
Fix the git branch name/Zanata branch name mismatch

M4rtinK:
This is a F25 specific hotfix to reconcile our
new branching scheme with Zanata branch naming.

In rawhide & further this is already fixed in a more robust manner.

### DIFF
--- a/scripts/makebumpver
+++ b/scripts/makebumpver
@@ -36,6 +36,9 @@ import sys
 import textwrap
 import urllib.request
 
+DEFAULT_ZANATA_GIT_BRANCH_ALIASES = {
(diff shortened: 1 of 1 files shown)


scripts/makebumpver
```diff
@@ -36,6 +36,9 @@ import sys
 import textwrap
 import urllib.request
 
+DEFAULT_ZANATA_GIT_BRANCH_ALIASES = {
+    "F25-release": "f25",
```
snbueno: Just change F25-release to f25-release, and this patch works otherwise. :)

PR #780, opened 2016-09-09
Rhel7 fix net when dud unload

jkonecny12:
This should fix the network driver unload problem. I'm thinking that this can use some tests so I'll some later to this PR.

### DIFF
--- a/dracut/Makefile.am
+++ b/dracut/Makefile.am
@@ -20,34 +20,36 @@
 dracutdir = $(prefix)/lib/dracut/modules.d/80$(PACKAGE_NAME)
 dist_dracut_SCRIPTS = module-setup.sh \
                       python-deps \
-		      anaconda-lib.sh \
-		      parse-anaconda-options.sh \
-		      parse-anaconda-repo.sh \
-		      parse-anaconda-kickstart.sh \
-		      parse-anaconda-net.sh \
-		      repo-genrules.sh \
-		      kickstart-genrules.sh \
-		      updates-genrules.sh \
-		      anaconda-ks-sendheaders.sh \
-		      anaconda-netroot.sh \
-		      anaconda-diskroot \
-		      anaconda-copy-ks.sh \
-		      anaconda-copy-cmdline.sh \
-		      anaconda-copy-dhclient.sh \
-		      anaconda-copy-s390ccwconf.sh \
-		      anaconda-ifcfg.sh \
-		      fetch-kickstart-net.sh \
-		      fetch-kickstart-disk \
-		      fetch-updates-disk \
-		      parse-kickstart \
-		      anaconda-modprobe.sh \
-		      save-initramfs.sh \
-		      anaconda-pre-shutdown.sh \
-		      parse-anaconda-dd.sh \
-		      fetch-driver-net.sh \
-		      driver-updates@.service \
-		      driver-updates-genrules.sh \
-		      anaconda-depmod.sh \
-		      driver_updates.py
+                      anaconda-lib.sh \
+                      parse-anaconda-options.sh \
+                      parse-anaconda-repo.sh \
+                      parse-anaconda-kickstart.sh \
+                      parse-anaconda-net.sh \
+                      repo-genrules.sh \
+                      kickstart-genrules.sh \
+                      updates-genrules.sh \
+                      anaconda-ks-sendheaders.sh \
+                      anaconda-netroot.sh \
+                      anaconda-diskroot \
+                      anaconda-copy-ks.sh \
+                      anaconda-copy-cmdline.sh \
+                      anaconda-copy-dhclient.sh \
+                      anaconda-copy-s390ccwconf.sh \
+                      anaconda-ifcfg.sh \
+                      fetch-kickstart-net.sh \
+                      fetch-kickstart-disk \
+                      fetch-updates-disk \
+                      parse-kickstart \
+                      anaconda-modprobe.sh \
+                      save-initramfs.sh \
+                      anaconda-pre-shutdown.sh \
+                      parse-anaconda-dd.sh \
+                      fetch-driver-net.sh \
+                      driver-updates@.service \
+                      driver-updates-genrules.sh \
+                      anaconda-depmod.sh \
+                      find-net-intfs-by-driver \
+                      anaconda-ifdown \
+                      driver_updates.py
 
 MAINTAINERCLEANFILES = Makefile.in
--- a/dracut/README-dd
+++ b/dracut/README-dd
@@ -50,6 +50,14 @@ driver-updates
     Pass inst.dd with no parameters and it will prompt the user for specific
     drivers to load using a simple text user interface.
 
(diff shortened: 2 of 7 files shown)


dracut/driver_updates.py
```diff
@@ -342,6 +342,35 @@ def grab_driver_files(outdir="/updates"):
...
 
+def net_intfs_by_modules(mods):
+    """get list of network interfaces which are depending on given kernel module"""
+    ret = set()
+    for mod in mods:
+        out = subprocess.check_output(["find-net-intfs-by-driver", mod])
+
+        for line in out.split('\n'):
+            if line:
+                ret.add(line.strip())
```
vpodzime: This could be a one-line comprehension.

dracut/anaconda-ifdown
```diff
@@ -0,0 +1,17 @@
...
+netif="$1"
+# ip down/flush ensures that routing info goes away as well
+ip link set $netif down
+ip addr flush dev $netif
+echo "#empty" > /etc/resolv.conf
+rm -f -- /tmp/*.$netif.*
+if [ -e /sys/class/net/$netif/address ]; then
+    address=$(cat /sys/class/net/$netif/address)
+    rm -f -- /tmp/*.$address.*
+fi
```
vpodzime: The ``rm`` utility fails if there's nothing to remove, doesn't it? Not sure if that's desirable here and above.

dracut/driver_updates.py
```diff
@@ -394,6 +430,13 @@ def load_drivers(moddict):
     if modules_to_add:
         subprocess.call(["modprobe", "-a"] + modules_to_add)
 
+    # Step 3: Retrigger udev event for net subsystem to force Dracut to load
+    # network settings for new interfaces and for replaced interfaces
+    if net_intfs_unload or pre_remove_intfs != list_net_intfs():
+        log.debug("Retrigger udev events for network interfaces")
+        cmd = ["udevadm", "trigger", "--action=change", "--subsystem-match=net"]
+        subprocess.check_call(cmd)
```
vpodzime: Isn't this going to cause stage2 image, kickstart,... re-download?

dracut/find-net-intfs-by-driver
```diff
@@ -0,0 +1,23 @@
...
+# List all interfaces in system
+for i in /sys/class/net/*/device/modalias; do
+    if [ -z $i ]; then
+        exit 0
+    fi
+    # Get kernel mods on which the interface is dependent
+    d=$(modprobe -R $(cat "$i"))
+
+    # Test if this is the mod we are looking for. If so return name of the interface.
+    if [[ " $d " == *\ $driver\ * ]]; then
```
vpodzime: do we really want the asterisks in the above? On both sides?

dracut/README-dd
```diff
@@ -50,6 +50,15 @@ driver-updates
     Pass inst.dd with no parameters and it will prompt the user for specific
     drivers to load using a simple text user interface.
 
+find-net-intfs-by-driver
+    Find all network interfaces which depends on the given driver (command
```
vpodzime: *depend*

dracut/README-dd
```diff
@@ -50,6 +50,15 @@ driver-updates
...
+    line argument), then return list of the network interfaces.
```
vpodzime: *a list*

dracut/README-dd
```diff
@@ -50,6 +50,15 @@ driver-updates
...
+
+anaconda-ifdown
+    This script will set the interface down and it removes all flags
```
vpodzime: *will set* -> *sets*

dracut/driver_updates.py
jkonecny12: It shouldn't, this is done before downloading of stage 2.

dracut/find-net-intfs-by-driver
jkonecny12: Yes we need them. This could return more modules and I need to test if the given module is in middle of other modules.

dracut/anaconda-ifdown
jkonecny12: Nope the `-f` stands for `ignore nonexistent files and arguments, never prompt` so there is no error it just do nothing.

jkonecny12:
Updated, thank you @vpodzime .

jkonecny12:
**Updated:** 
- Fixed small issue in the code. 
- Fix and add test case.

dracut/driver_updates.py
```diff
@@ -359,6 +359,33 @@ def grab_driver_files(outdir="/updates"):
 
     return module_dict
 
+def net_intfs_by_modules(mods):
+    """get list of network interfaces which are depending on given kernel module"""
+    ret = set()
+    for mod in mods:
+        out = subprocess.check_output(["find-net-intfs-by-driver", mod])
+
```
vpodzime: Extra blank line here.

dracut/driver_updates.py
```diff
@@ -411,6 +446,13 @@ def load_drivers(moddict):
     if modules_to_add:
         subprocess.call(["modprobe", "-a"] + modules_to_add)
 
+    # Step 3: Retrigger udev event for net subsystem to force Dracut to load
+    # network settings for new interfaces and for replaced interfaces
+    if net_intfs_unload or pre_remove_intfs != list_net_intfs():
+        log.debug("Retrigger udev events for network interfaces")
+        cmd = ["udevadm", "trigger", "--action=change", "--subsystem-match=net"]
```
vpodzime: I don't really understand the trigger here. Shouldn't it be triggered automatically as the new device appears in the system after its module is added? Or is just an *add* action triggered at that time?

vpodzime:
Looks good to me otherwise.

dracut/driver_updates.py
jkonecny12: Fixed, thank you.

jkonecny12:
**Updated:** 
* Removed udev event re-trigger. It will be done automatically by removing and adding the interface.
* Fix tests

dracut/Makefile.am
```diff
@@ -20,34 +20,36 @@
...
+                      save-initramfs.sh \
+                      anaconda-pre-shutdown.sh \
+                      parse-anaconda-dd.sh \
+                      fetch-driver-net.sh \
+                      driver-updates@.service \
+                      driver-updates-genrules.sh \
+                      anaconda-depmod.sh \
+                      find-net-intfs-by-driver \
+                      anaconda-ifdown \
+                      driver_updates.py
```
rvykydal: just checking - is the whitespace change intentional?

rvykydal:
Other then the white space question, looks ok to me.

dracut/Makefile.am
jkonecny12: Yes it is, because first line of this is only white space and all other lines was mix of a white space and tabs. I made it consistent.

jkonecny12:
Merged, thank you all for your valuable input.

PR #781, opened 2016-09-10
rpmostreepayload: Fix remote handling to use correct sysroot

cgwalters:
I'm working on Atomic Workstation, and putting the ostree remote
config into a package (same as CentOS is doing it), and this
is the current recommendation.

See
https://github.com/rhinstaller/anaconda/pull/474
for some previous discussion around this.

Recently, this broke; we end up deleting the remote in the package,
and the reason for this is we're using the *physical* sysroot rather
than the target, which means we look at the wrong
`/mnt/sysimage/etc/ostree/remotes.d`.

What I don't yet know is *why* this broke - I have a suspicion it's
related to some of the mount setup changes bcl did previously here,
but anyways, this patch fixes things.

### DIFF
--- a/pyanaconda/packaging/rpmostreepayload.py
+++ b/pyanaconda/packaging/rpmostreepayload.py
@@ -48,7 +48,6 @@ class RPMOSTreePayload(ArchivePayload):
     def __init__(self, data):
         super(RPMOSTreePayload, self).__init__(data)
         self._remoteOptions = None
-        self._sysroot_path = None
         self._internal_mounts = []
 
     @property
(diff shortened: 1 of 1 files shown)


pyanaconda/packaging/rpmostreepayload.py
```diff
@@ -292,10 +292,11 @@ def postInstall(self):
         # However, we ignore the case where the remote already exists,
         # which occurs when the content itself provides the remote
         # config file.
-        sysroot = OSTree.Sysroot.new(self._sysroot_path)
+        sysroot_path = Gio.File.new_for_path(iutil.getSysroot())
+        sysroot = OSTree.Sysroot.new(sysroot_path)
```
vpodzime: I have no idea what's going on in this patch, but maybe ``sysroot_file`` would be a better name for the variable?

cgwalters:
OK, done in :arrow_up:

rvykydal:
> Looks good to me. We'll need to rebase for f25-devel before pushing though.

I did it in: https://github.com/rhinstaller/anaconda/pull/804

rvykydal:
The patches was merged in https://github.com/rhinstaller/anaconda/pull/804

PR #782, opened 2016-09-12
screen_access: Ensure we write config to real sysroot

cgwalters:
The value of `iutil.getSysroot()` can change dynamically at
least in the case of an ostree payload, but Python will only
evaluate the default value once when the function is parsed.

Change the default to `None`, and take `None` to mean "use current
sysroot".

### DIFF
--- a/pyanaconda/screen_access.py
+++ b/pyanaconda/screen_access.py
@@ -64,7 +64,7 @@ class ScreenAccessManager(object):
                 with open(config_path, "rt") as f:
                     self._config.read_file(f)
 
-    def write_out_config_file(self, config_path=iutil.getSysroot() + CONFIG_FILE_PATH):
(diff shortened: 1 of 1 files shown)


vpodzime:
ooops. :) Looks good to me.

M4rtinK:
Oops indeed! Also looks good to me & thanks for the patch! :)

M4rtinK:
@cgwalters : BTW, it think this should also go to F25 - could you open a PR for the f25-devel branch ? We periodically merge the f25-devel branch to master (mostly right after a PR is merged) so it will also end in master/Rawhide. Thanks in advance!

M4rtinK:
Hmm, tried something but looks like you will still have to do the new PR yourself - the "change base" feature apparently does not work how it should. :)

PR #783, opened 2016-09-12
dud: fix multiple inst.dd=http:// instances stalling in dracut (#1268…

rvykydal:
…792)

Related: rhbz#1268792

It seems that fetch_url routine has been eating the second network driverdisk
target (2nd line) from dd_net sent to stdin of the while block. As a
consequence, the second driverdisk was never fetched and dracut stalled waiting
for dd tasks to be done.


vpodzime:
Funny...

PR #786, opened 2016-09-12
installclass: Add debugging information

cgwalters:
While debugging why my installclass isn't being loaded, might as well
have this patch I wrote to figure out why.


jkonecny12:
Hello,
change it to `log.debug` and please change `'` to `"` for the consistency of the code.

Other than that it looks good to me.

jkonecny12:
Jenkins, it's ok to test.

cgwalters:
Eh, not worth updating this.

PR #788, opened 2016-09-13
Fix traceback when payload have None as url (#1371494)

jkonecny12:
It's regression from commit 136961544060bd3891af981bbae5bc969117d08a

*Resolves: rhbz#1371494*

### DIFF
--- a/pyanaconda/packaging/__init__.py
+++ b/pyanaconda/packaging/__init__.py
@@ -230,7 +230,7 @@ class Payload(object):
         """
         network_protocols = ["http:", "ftp:", "nfs:", "nfsiso:"]
         for s in sources:
-            if any(s.startswith(p) for p in network_protocols):
+            if s and any(s.startswith(p) for p in network_protocols):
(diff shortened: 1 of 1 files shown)


pyanaconda/packaging/__init__.py
```diff
@@ -230,7 +230,7 @@ def _sourceNeedsNetwork(self, sources):
         """
         network_protocols = ["http:", "ftp:", "nfs:", "nfsiso:"]
         for s in sources:
-            if any(s.startswith(p) for p in network_protocols):
+            if any(s.startswith(p) for p in network_protocols if s):
```
rvykydal: Seems quite strange to me, I thing simple

if s and any(s.startswith(p) for p in network_protocols):

would be more readable

jkonecny12: OMG yeah, I'm blind I thought I'm testing the internal for part. Thank you for pointing this out. I'll fix that immediately.

rvykydal:
Looks good to me.

PR #789, opened 2016-09-13
Fix some issues in the password checking code

M4rtinK:
Thanks to Jiri Kortus for noticing the typos.

Related: rhbz#1365112
Related: rhbz#1365032

### DIFF
--- a/tests/gettext/gettext_potfiles.py
+++ b/tests/gettext/gettext_potfiles.py
@@ -24,9 +24,17 @@ def check_potfile(checkfile, potlist):
     global success
 
     potcheckfile = None
+
     if checkfile.endswith(".py"):
+        # Check whether the file should be ignored
+        # - there are some unhandled cases where this check returns false positives,
+        #   such as tests that just use the _() function to translate content of variables
+        #   without marking new strings for translation
+        # - so provide a way to skip the check in such a case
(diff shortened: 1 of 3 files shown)


M4rtinK:
I've added two more commits fixing issues reported by our testsuite, so I've changed the name of the PR & removed the ACK as I think a re-review is needed.

tests/gettext/gettext_potfiles.py
```diff
@@ -24,7 +24,16 @@ def check_potfile(checkfile, potlist):
...
 
     potcheckfile = None
+
     if checkfile.endswith(".py"):
+        # Check whether the file should be ignored
+        # - there are some unhandled cases where this check returns false positives,
+        #   such as tests that just use the _() function to translate content of variables
+        #   without marking new strings for translation
+        # - so provide a way to skip the check in such a case
+        if subprocess.call(["grep", "-q", "^# ignore-check: potfiles", checkfile]) == 0:
```
vpodzime: I think this condition needs to be moved to the ``if`` above otherwise the ``elif`` branch below will never be entered as it has the same condition.

M4rtinK: Thanks! Should be fixed now. :)

PR #790, opened 2016-09-13
screen_access: Ensure we write config to real sysroot

cgwalters:
The value of `iutil.getSysroot()` can change dynamically at
least in the case of an ostree payload, but Python will only
evaluate the default value once when the function is parsed.

Change the default to `None`, and take `None` to mean "use current
sysroot".


cgwalters:
Moved from https://github.com/rhinstaller/anaconda/pull/782

M4rtinK:
Thanks! Also as this PR was already ACKed & passed the 24 hour review rule before I'll merge it right away.

PR #794, opened 2016-09-14
F25 devel rhbz 1321288

rvykydal:
Port from rhel7-branch.

### DIFF
--- a/pyanaconda/network.py
+++ b/pyanaconda/network.py
@@ -686,7 +686,7 @@ def ksdata_from_ifcfg(devname, uuid=None):
     if devname in nm.nm_devices():
         # virtual devices (bond, vlan, ...) not activated in installer
         # are not created so guard these checks
-        if nm.nm_device_is_slave(devname):
+        if nm.nm_device_is_slave(devname) and nm.nm_device_type_is_ethernet(devname):
             return None
         if nm.nm_device_type_is_wifi(devname):
             # wifi from kickstart is not supported yet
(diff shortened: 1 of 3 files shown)


pyanaconda/network.py
```diff
@@ -686,7 +686,7 @@ def ksdata_from_ifcfg(devname, uuid=None):
     if devname in nm.nm_devices():
         # virtual devices (bond, vlan, ...) not activated in installer
         # are not created so guard these checks
-        if nm.nm_device_is_slave(devname):
+        if nm.nm_device_is_slave(devname) and nm.nm_device_type_is_ethernet(devname):
```
vpodzime: Doesn't this require a change of the comment right above?

vpodzime:
Looks good to me otherwise.

pyanaconda/network.py
rvykydal: The comment actually talks about the guard above (if devname in nm.nm_devices()), the changed line is just one of "these checks". (The change is that we want to ingore slaves only if they are ethernet devices, ie not bond devices which - unlike ethernet slaves - should have its own kickstart network command)
Anyway, the comment should be moved above the guard line so it is clear what it is talking about?

rvykydal:
jenkins, test this please

rvykydal:
jenkins, test this please

pyanaconda/network.py
vpodzime: Yeah I think it the comment would make more sense if moved above the guard. And maybe we need another comment for the added check?

PR #803, opened 2016-09-15
Fix password length in the password policy

M4rtinK:
Now that policy-based password length checking works correctly set the
minimal passowrd length in the policy to the value expected by users from
RHEL <=7.2 (6 characters long).

Related: rhbz#1365112
Related: rhbz#1365032


snbueno:
Looks good to me.

PR #805, opened 2016-09-20
Check /var/tmp directory before creating it (#1377761)

rvykydal:
It can be already existing as mount point configured by user.

Resolves: rhbz#1377761


rvykydal:
Note: the failing checks seem to be the same as in the last PR merged into rhel7-branch: https://github.com/rhinstaller/anaconda/pull/803

PR #806, opened 2016-09-20
Fix NTP server list fetching when running in IS (#1374810)

M4rtinK:
Fix fetching of the NTP server list when the NTP TUI spoke
is running in Initial Setup instead of Anaconda.

Unlike in Anaconda where we get the NTP server list directly
the config file parsing function used when running in Initial Setup
returns a (NTP pools, NTP servers) tuple, so just take the second item,
not the tuple itself.

Resolves: rhbz#1374810


jkonecny12:
jenkins, test this please

PR #810, opened 2016-09-26
Fix screenshot taking logic (#1327456)

M4rtinK:
Unlike before (RHEL ~7.1, Fedora <21) Anaconda now uses a window stack
(GtkStack) which gereally works nicely but unfortunately
in some ways breaks the assumptions the original screenshot taking code made,
resulting with various issues, such as:

- multiple simultaniously registered key bindings
-> as all windows are now initialized even though just one can be actually
   visible at a time
- content of a wrong window might be saved as a the screenshot
-> all the key bindings fire at once when the key is pressed
-> the first binding to fire saves the screenshot, independently
   on the currently visible spoke

To fix this:
- we now register just one screenshot key binding - in MainWindow
- the screenshot taking logic has been moved to MainWindow
- the window on top of the window stack is used when taking a screenshot,
  making sure we really take a screenshot of what is visible on the screen

Resolves: rhbz#1327456

### DIFF
--- a/pyanaconda/ui/gui/__init__.py
+++ b/pyanaconda/ui/gui/__init__.py
@@ -16,7 +16,7 @@
 # License and may only be used or replicated with the express permission of
 # Red Hat, Inc.
 #
-import inspect, os, sys, time, site, signal
+import inspect, os, sys, site, signal
 import meh.ui.gui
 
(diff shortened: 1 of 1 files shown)


M4rtinK:
Fixed Pylint warning.

PR #811, opened 2016-09-26
Improved driver disk copying (#1269915)

M4rtinK:
Turns out that just changing flags for the cp command won't do
as function needs to support both single file driver disk RPM copy
as well as copying all driver disk RPMs from a given folder.

So let's drop the cp command all together and replace it with Python shutil
based equivalent. As far as I can tell it seams that there should be
no sub-folders in a driver disk RPM repo, so don't copy them
but just log a warning instead.

Also improve the copy-RPM unit tests to cover more scenarios and
add add more logging in the stage2 driver disk package handling code.

Related: rhbz#1269915

### DIFF
--- a/dracut/driver_updates.py
+++ b/dracut/driver_updates.py
@@ -73,6 +73,7 @@ import fnmatch
 # because it breaks sometimes.
 if os.isatty(0):
     import readline # pylint:disable=unused-import
+import shutil
 
 from contextlib import contextmanager
 from logging.handlers import SysLogHandler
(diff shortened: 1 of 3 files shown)


dracut/driver_updates.py
```diff
@@ -263,7 +264,23 @@ def save_repo(repo, target="/run/install"):
...
+    if os.path.isfile(repo):
+        shutil.copy2(repo, newdir)
+    elif os.path.isdir(repo):
+        for item in os.listdir(repo):
+            item_path = os.path.join(repo, item)
+            if os.path.isfile(item_path):
+                log.debug("copying %s to %s", item_path, newdir)
+                shutil.copy2(item_path, newdir)
+            else:
+                log.warning("WARNING: DD repo content not a file: %s", item_path)
```
vpodzime: You should drop the *WARNING: * prefix here. Otherwise the log message would look like *timestamp WARNING: WARNING: ...*.

M4rtinK: OK. :)

PR #813, opened 2016-09-27
Fix network spoke being incorrectly marked as mandatory (#1374864)

M4rtinK:
The network spoke should only be marked mandatory during an installation
from network sources.

So only mark it as mandatory when it runs in Anaconda & the installation
source indicates that it needs network.

Also use a modern way of checking if the installation source needs
network when we are at it.

Resolves: rhbz#1374864


jkonecny12:
Did you try this in Live DVD installation? I don't know what will happen when there are no repositories.

M4rtinK:
Yeah - Live has no network spoke & if payload.needsNetwork does not work correctly on DVD installation then it's a bug and should be fixed.

PR #817, opened 2016-09-30
Skip live image on usb when checking storage for mounted partitions (#1369786)

rvykydal:
Do not check partitions that are protected (eg the usb stick partition with
live image).  They are not hidden by selection of disks, so they are included
in selected disks. Filter them out in the check then.

### DIFF
--- a/pyanaconda/storage_utils.py
+++ b/pyanaconda/storage_utils.py
@@ -360,6 +360,10 @@ def check_mounted_partitions(storage):
             continue
 
         for part in disk.format.partitions:
+            part_dev = storage.devicetree.get_device_by_path(part.path)
+            if part_dev and part_dev.protected:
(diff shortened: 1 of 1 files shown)


pyanaconda/storage_utils.py
```diff
@@ -360,6 +360,10 @@ def check_mounted_partitions(storage):
             continue
 
         for part in disk.format.partitions:
+            part_dev = storage.devicetree.get_device_by_path(part.path)
+            if part_dev and part_dev.protected:
+                log.debug("Not checking %s for being mounted, assuming live image mount", part.path)
```
vpodzime: I think this should mention that the partition is marked as protected.

AdamWill:
this is blocking F25 Beta, and Go/No-Go is Thursday, so we really need this merged ASAP...thanks :)

PR #818, opened 2016-09-30
SL Install Class

jcpunk:
Echoing 4de0ec44bdf0f68545bb55bb5fea00464b65fcab

May as well include the SL file

### DIFF
--- a/pyanaconda/installclasses/scientific.py
+++ b/pyanaconda/installclasses/scientific.py
@@ -1,0 +1,40 @@
+#
+# scientific.py
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+
+from pyanaconda.installclasses.rhel import RHELBaseInstallClass
+from pyanaconda.product import productName
+
+class ScientificBaseInstallClass(RHELBaseInstallClass):
+    '''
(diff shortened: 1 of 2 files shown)


InStallker:

InStallker:

vpodzime:
Jenkins, this is ok to test.

vpodzime:
Jenkins, test this, please.

pyanaconda/installclasses/scientific.py
```diff
@@ -0,0 +1,59 @@
...
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+
+from pyanaconda.installclass import BaseInstallClass
+from pyanaconda.product import productName
+from pyanaconda import network
+from pyanaconda import nm
+
+class RHELBaseInstallClass(BaseInstallClass):
```
jcpunk: Whoops, silly typo.  I'll fix it in a new commit.

There may be some shift over time between SL and RHEL.  In particular installUpdates is on for SL.

jkonecny12:
Please fix this Jenkins issue:
```
************* Module ../pyanaconda/installclasses/scientific.py
W9951(pointless-class-attribute-override): 34,21: ScientificBaseInstallClass: Assignment to class attribute installUpdates overrides identical assignment in ancestor BaseInstallClass.
```

jcpunk:
I'll see about getting that fixed.

I was thinking further on rvykydal's comment.  I may try inheriting RHEL and just altering the SL specifics.  Might make long term maintenance easier.

I'll post a new commit sometime today.

jcpunk:
Think I've got it.  While I was there I fixed up the CentOS class def so that it doesn't get overlooked.

pyanaconda/installclasses/scientific.py
```diff
@@ -0,0 +1,40 @@
...
+
+    installUpdates = True
+
+    efi_dir = "sl"
+
+    help_placeholder = "SLPlaceholder.html"
+    help_placeholder_with_links = "SLPlaceholder.html"
+
+    def __init__(self):
+        RHELBaseInstallClass.__init__(self)
```
rvykydal: I think there is no need to call ancestor's __init__ in otherwise empty __init__ method, but we should clean this up also in our rhel and fedora installclasses so I'll leave it for another patch. Other than that it looks good to me.

PR #820, opened 2016-10-05
Don't send intermediate pot files to zanata (gh#791)

AdamWill:
Translators reported that master and f25 branches in zanata
have two pot files, 'main' and 'extra', which never existed
before f25; this caused a sudden drop in translation completion
percentage.

0ee1f56d changed how we make pot files. Now `anaconda.pot` is
generated from two intermediate pot files (main and extra). These
are purely intermediate files, but because `Makefile.am` doesn't
delete them after `anaconda.pot` is done, passing `--srcdir` to
`zanata push` just sends all three `.pot` files to Zanata.

Just using `--srcfile` instead should be sufficient to fix it
(though this is all based on five minutes of reading zanata docs,
please check this). We could also  wipe the intermediate files
once `anaconda.pot` is made, I'm not totally sure whether we
should do that as well (or instead).

Resolves: gh#791

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -46,7 +46,7 @@ uninstall-hook:
 ARCHIVE_TAG   = $(PACKAGE_NAME)-$(PACKAGE_VERSION)-$(PACKAGE_RELEASE)
 
 ZANATA_PULL_ARGS = --transdir $(srcdir)/po/
-ZANATA_PUSH_ARGS = --srcdir $(srcdir)/po/ --push-type source --force
+ZANATA_PUSH_ARGS = --srcfile $(srcdir)/po/anaconda.pot --push-type source --force
 
 RC_RELEASE ?= $(shell date -u +0.1.%Y%m%d%H%M%S)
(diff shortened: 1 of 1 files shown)


InStallker:

InStallker:

AdamWill:
@InStallker go home, you're drunk

vpodzime:
> @InStallker go home, you're drunk

Looks like this worked!

vpodzime:
jenkins, this is ok to test

vpodzime:
Jenkins, test this, please.

jkonecny12:
> @InStallker go home, you're drunk

Sorry about this. It's Jenkins and it looks like there is some bug in plugin. I need to look on that in more detail.

jkonecny12:
I have created PR https://github.com/rhinstaller/anaconda/pull/842 containing this fix. The original fix didn't work but I think it's nice to have it there for future as a bonus protection.

I'm closing this PR now.

PR #823, opened 2016-10-11
Adjust the fraction of the progress bar (#1324706)

poncovka:
The value of a fraction in gtk.ProgressBar should be between 0.0 and 1.0.

Resolves: rhbz#1324706

### DIFF
--- a/iutil.py
+++ b/iutil.py
@@ -1054,3 +1054,16 @@ def copytree(src, dst, symlinks=False, preserveOwner=False,
         errors.extend((src, dst, e.strerror))
     if errors:
         raise Error, errors
+
+def adjust_fraction(fraction):
+    """ Adjusts the fraction to the interval from 0.0 to 1.0.
+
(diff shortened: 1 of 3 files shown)


iutil.py
```diff
@@ -1054,3 +1054,16 @@ def trySetfilecon(src, dest):
         errors.extend((src, dst, e.strerror))
     if errors:
         raise Error, errors
+
+def adjust_fraction(fraction):
+    """ Adjusts the fraction to the interval from 0.0 to 1.0.
+
+        Should be applied to a parameter of gtk.ProgressBar.set_fraction()
+        to ensure the correct value.
```
vpodzime: I think this part of the docstring is not necessary, the function is generic, after all.

jkonecny12: I think the doc string should stay here. Yes the code is self explanation but on the other hand, you have here information why this is required.

PR #839, opened 2016-10-21
Improve packaging logs without DEBUG logging

jkonecny12:
I don't know how but I have got packaging.log from customer without a ``DEBUG`` messages. I think more messages should be changed from debug to info.


vpodzime:
Shouldn't we rather focus on why DEBUG messages are ignored? This looks like a matter of time when somebody adds new DEBUG log messages which will then be silently ignored.

jkonecny12:
I'm completely for that but I don't have a lead why this is happening. The next log file from him has the DEBUG messages so hard to say where is this error. However I think these DEBUG messages should be INFO anyway.

PR #842, opened 2016-10-21
F25 dev rm zanata main extra pot

jkonecny12:
Remove main.pot and extra.pot files from zanata.

This PR contains https://github.com/rhinstaller/anaconda/pull/820 but there is an another commit because the original fix didn't work. Even it didn't work I think it's a nice bonus protection for future.


jkonecny12:
PR https://github.com/rhinstaller/anaconda/pull/843 is required for master because there is another place to fix this which is not on f25-devel.

PR #844, opened 2016-10-24
Fix picking mountpoint for package download (#1328151)

jkonecny12:
Mountpoint where we are downloading packages was picked by the same method as what is used in space check. This method is not working correctly when user skipped space check, it still taking install size in count.

When the first method failed, try again but don't include install size and try to prefer other than root mountpoint for downloading.

*Related: rhbz#1328151*

*Reported-by: Orion Poplawski `<orion@cora.nwra.com>`*

### DIFF
--- a/pyanaconda/packaging/dnfpayload.py
+++ b/pyanaconda/packaging/dnfpayload.py
@@ -113,27 +113,47 @@ def _paced(fn):
         return fn(self, *args)
     return paced_fn
 
-def _pick_mpoint(df, download_size, install_size):
+def _pick_mpoint(df, download_size, install_size, download_only):
     def reasonable_mpoint(mpoint):
         return mpoint in DOWNLOAD_MPOINTS
 
     requested = download_size
     requested_root = requested + install_size
     root_mpoint = pyanaconda.iutil.getSysroot()
-    sufficients = {key : val for (key, val) in df.items()
-                   # for root we need to take in count both download and install size
-                   if ((key != root_mpoint and val > requested)
-                   or val > requested_root) and reasonable_mpoint(key)}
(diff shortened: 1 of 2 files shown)


pyanaconda/packaging/dnfpayload.py
```diff
@@ -113,27 +113,47 @@ def paced_fn(self, *args):
...
     log.debug('Estimated size: download %s & install %s', requested,
               (requested_root - requested))
-    log.info('Sufficient mountpoints found: %s', sufficients)
+
+    # Find sufficient mountpoint to download and install packages.
+    sufficients = {key : val for (key, val) in df.items()
+                   if ((key != root_mpoint and val > requested) or val > requested_root)
+                      and reasonable_mpoint(key)}
+
+    # If no sufficient mountpoints for download and install was found and we are looking
```
vpodzime: *"were found"*

pyanaconda/packaging/dnfpayload.py
```diff
@@ -113,27 +113,47 @@ def paced_fn(self, *args):
...
+    # for download mountpoint only, ignore install size and try to find mountpoint just
+    # to download packages. This fallback is required when user skipped space check.
+    if not len(sufficients) and download_only:
```
vpodzime: I find ``if not sufficients`` better readable and understandable than testing the length. If you want to use the length in the condition, compare it to ``0``.

jkonecny12:
Updated, should be fixed now.

vpodzime:
> Updated, should be fixed now.

Perfect, thanks!

jkonecny12:
Update: Tests were added.

PR #847, opened 2016-10-26
A couple of small s390x fixes. (#1384532)

snbueno:
Some things changed in blivet, and this is just adapting to a new way of life. Otherwise we'll never be able to manually add extra DASD or zFCP disks.

### DIFF
--- a/pyanaconda/ui/gui/spokes/advstorage/dasd.py
+++ b/pyanaconda/ui/gui/spokes/advstorage/dasd.py
@@ -42,7 +42,8 @@ class DASDDialog(GUIObject):
     def __init__(self, data, storage):
         GUIObject.__init__(self, data)
         self.storage = storage
-        self.dasd = self.storage.dasd
+        self.dasd = [d for d in self.storage.devices if d.type == "dasd"]
+        self.dasd.sort(key=lambda d: d.name)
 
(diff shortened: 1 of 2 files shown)


pyanaconda/ui/gui/spokes/advstorage/dasd.py
```diff
@@ -42,7 +42,8 @@ class DASDDialog(GUIObject):
     def __init__(self, data, storage):
         GUIObject.__init__(self, data)
         self.storage = storage
-        self.dasd = self.storage.dasd
+        self.dasd = [d for d in self.storage.devices if d.type == "dasd"]
+        self.dasd.sort(key=lambda d: d.name)
```
jkonecny12: Shouldn't there be some better way to do this. I mean removing the `storage.dasd` without proper replacement looks somehow strange to me.
Could you please confirm that this is right way how to get DASD disks now, @vpodzime @dwlehman ?

vpodzime: Yes, this is correct. We don't want to have a static list of DASDs (which it used to be) as well as we don't want to have a property for each device type (dasd, cdrom, disk, iscsi,...).

PR #848, opened 2016-10-27
Correct deviceLinks to device_links (blivet renamed it)

AdamWill:
In 4e8f941b , blivet renamed the `deviceLinks` attribute of
`StorageDevice` and its subclasses to `device_links`. These
uses of it were never updated. The iSCSI one causes anaconda to
crash when logging into an iSCSI target; the other one probably
breaks something as well, but I don't know what.

Needed on master and f25.


jkonecny12:
Jenkins, it's ok to test.

vpodzime:
Still don't understand WTH pylint is doing with these things. Or more precisely what it is **not** doing with them.

PR #849, opened 2016-10-27
use blivet iSCSI singleton directly in storage spoke

AdamWill:
`Blivet` instances no longer have an `iscsi` attribute which
is the blivet iscsi singleton since 650505fb . Some bits of
anaconda were adjusted to use the singleton directly, but the
storage spoke GUI code was missed.


jkonecny12:
Jenkins, it's ok to test.

jkonecny12:
Thank you for this patch. Good job!

PR #850, opened 2016-10-27
iSCSI: adjust to change in how blivet passes auth info

AdamWill:
When blivet switched from using libiscsi to storaged, the node
attribute of an `iScsiDiskDevice` changed from being a libiscsi
`PyIscsiNode` instance to a blivet `NodeInfo` namedtuple, but
both blivet and anaconda continued trying to call the `getAuth`
method of the node - which `NodeInfo` doesn't have - to find
the authentication info for the node. I've sent a blivet PR
to store this information in the `NodeInfo` namedtuple instead:
https://github.com/rhinstaller/blivet/pull/518 . This patch
adjusts anaconda to use it.

Note: this assumes that the `self.iscsi.log_into_node()` call in `pyanaconda/ui/gui/spokes/advstorage/iscsi.py` will happen before this code gets hit; I think it should, but if that's not the case, it won't work right.

### DIFF
--- a/pyanaconda/ui/gui/spokes/storage.py
+++ b/pyanaconda/ui/gui/spokes/storage.py
@@ -419,14 +419,12 @@ class StorageSpoke(NormalSpoke, StorageChecker):
         if iscsi.ifaces:
             iscsi_data.iface = iscsi.ifaces[dev_node.iface]
 
-        auth = dev_node.getAuth()
-        if auth:
-            if auth.username and auth.password:
-                iscsi_data.user = auth.username
-                iscsi_data.password = auth.password
-            if auth.reverse_username and auth.reverse_password:
-                iscsi_data.user_in = auth.reverse_username
-                iscsi_data.password_in = auth.reverse_password
+        if dev_node.username and dev_node.password:
+            iscsi_data.user = dev_node.username
+            iscsi_data.password = dev_node.password
+        if dev_node.r_username and dev_node.r_password:
+            iscsi_data.user_in = dev_node.r_username
+            iscsi_data.password_in = dev_node.r_password
         return iscsi_data
 
     @property


AdamWill:
This should still be compatible with the revised version of https://github.com/rhinstaller/blivet/pull/518 .

jkonecny12:
Jenkins, it's ok to test.

AdamWill:
The blivet change is still not merged yet.

jkonecny12:
I know, these tests are failing because of other issue. So from tests point of view, this PR looks good.

AdamWill:
The tests probably don't cover this code at all, then, because neither the current code nor the new code can *possibly* work at this point unless the tests are running with an ancient version of blivet that still uses libiscsi :)

AdamWill:
The blivet change is merged now, this should be merged too.

jkonecny12:
I would like to merge it right away but it's the final freeze now so I don't want to merge it without a freeze exception or blocker.
I could of course merge it to master without a problem but first I want to know if we want to create the exceptions for f25.

AdamWill:
we can count it as part of https://bugzilla.redhat.com/show_bug.cgi?id=1378156 , I'd say, since I just found all these bugs while fixing that.

M4rtinK:
BTW, I suggest noting bug numbers in the commit message (if known) - that way we can easily check if the bugs are blockers or have freeze exceptions.

We have some documentation for that here:
https://github.com/rhinstaller/anaconda/blob/master/CONTRIBUTING#L72

If the commit fixes/is related to more than one bug the RHEL-style notation can be used:
https://github.com/rhinstaller/anaconda/blob/master/CONTRIBUTING#L86

Thanks in advance! :)

AdamWill:
eh, it doesn't *technically* fix that specific bug, I'm just suggesting we wedge all the iSCSI fixes in together because I don't feel like filing five new bugs, proposing them all as blockers, and getting them all voted on =)

jkonecny12:
I just tested it and it looks like I'm getting unhandled exception from blivet on iscsi timeout which is causing crash of Anaconda. 
Fix for this is here https://github.com/rhinstaller/anaconda/pull/862 .
I'm going to merge this now. Thank you for these patches.

PR #854, opened 2016-10-31
Let DNF do its own substitutions

### LINKED ISSUE #493: Should DNF payload use it's substitutions?
https://github.com/rhinstaller/anaconda/blob/master/pyanaconda/packaging/dnfpayload.py#L237 is a "_replace_vars" function.

DNF provides dnf.conf.substitutions which will pull in any DNF supported vars automatically.

This would allow for substitutions other than $basearch and $releasever while also moving the maintenance of expected vars out of anaconda.

jcpunk:
Resolves #493

### DIFF
--- a/pyanaconda/packaging/dnfpayload.py
+++ b/pyanaconda/packaging/dnfpayload.py
@@ -20,7 +20,6 @@
 import os
 
 from blivet.size import Size
-import blivet.arch
 from pyanaconda.flags import flags
 from pyanaconda.i18n import _, N_
 from pyanaconda.progress import progressQ, progress_message
@@ -50,6 +49,8 @@ import dnf.logging
 import dnf.exceptions
 import dnf.repo
 import dnf.callback
+import dnf.conf.parser
+import dnf.conf.substitutions
 import rpm
 import librepo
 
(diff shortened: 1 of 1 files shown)


jkonecny12:
Jenkins, it's ok to test.

jkonecny12:
Hello @jcpunk please fix this error:
```
************* Module ../pyanaconda/packaging/dnfpayload.py
W0611(unused-import): 23,0: : Unused import blivet.arch
```

jcpunk:
Should be fixed up now.

jcpunk:
hmmm, this has been stuck in pending check for a bit.....

jkonecny12:
jenkins, test this please

jkonecny12:
It looked good before, there were errors but those errors where because of some other bugs which are resolved already.

If the check is stuck for some time you can try rebase. Sorry, it is not public, I hope it will be but we don't have enough time now to work on it.

poncovka:
Hi, could you use `self._base.conf.substitutions` instead of `self._dnf_substitutions`? I don't think it is necessary to create a new instance of `dnf.conf.substitutions.Substitutions`. Also the substitution of `$releasever` does not work now, but the use of `conf` substitutions should fix that.

jcpunk:
I've taken a quick stab at it.

poncovka:
Looks good to me. Substitution works fine now. Thanks!

PR #855, opened 2016-11-01
Fix tui timezone region selection by name

poncovka:
In the timezone settings, the selected region was different from the chosen one.

### DIFF
--- a/pyanaconda/ui/tui/spokes/time_spoke.py
+++ b/pyanaconda/ui/tui/spokes/time_spoke.py
@@ -281,7 +281,7 @@ class TimeZoneSpoke(NormalTUISpoke):
         # so whatever
         self._regions = list(timezone.get_all_regions_and_timezones().keys())
         self._timezones = dict((k, sorted(v)) for k, v in timezone.get_all_regions_and_timezones().items())
-        self._lower_regions = [r.lower() for r in self._timezones]
+        self._lower_regions = [r.lower() for r in self._regions]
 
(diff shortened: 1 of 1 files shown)


pyanaconda/ui/tui/spokes/time_spoke.py
```diff
@@ -281,7 +281,7 @@ def __init__(self, app, data, storage, payload, instclass):
         # so whatever
         self._regions = list(timezone.get_all_regions_and_timezones().keys())
         self._timezones = dict((k, sorted(v)) for k, v in timezone.get_all_regions_and_timezones().items())
-        self._lower_regions = [r.lower() for r in self._timezones]
+        self._lower_regions = [r.lower() for r in self._regions]
```
vpodzime: Does this really make a difference? It looks to me like ``self._regions`` is the same as ``self._timezones.keys()``.

poncovka: It does make a difference. The order of regions in `self._regions` has to be the same as in `self._lower_regions`, but it was not.

poncovka:
Jenkins, test this please

pyanaconda/ui/tui/spokes/time_spoke.py
vpodzime: Oh, that's right, thanks for the extra info!

poncovka:
Merged only to master because of the final freeze.

PR #856, opened 2016-11-01
tui: New class for prompt

poncovka:
Prompt is now an instance of a class Prompt, so it is easier to
manipulate with it.

All prompt methods are rewritten and there were added new lines
before the prompt in refresh methods. Class Widget was moved to
other widgets to fix some problems with dependencies.

It is a preparation for introducing the help option.

### DIFF
--- a/pyanaconda/ui/tui/simpleline/prompt.py
+++ b/pyanaconda/ui/tui/simpleline/prompt.py
@@ -1,0 +1,141 @@
+# Class for the Anaconda TUI prompt.
+#
+# Copyright (C) 2016  Red Hat, Inc.
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# the GNU General Public License v.2, or (at your option) any later version.
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY expressed or implied, including the implied warranties of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General
+# Public License for more details.  You should have received a copy of the
+# GNU General Public License along with this program; if not, write to the
+# Free Software Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
+# 02110-1301, USA.  Any Red Hat trademarks that are incorporated in the
+# source code or documentation are not subject to the GNU General Public
+# License and may only be used or replicated with the express permission of
+# Red Hat, Inc.
+#
+# Author(s):  Vendula Poncova <vponcova@redhat.com>
+#
+
+from pyanaconda.i18n import C_, N_
+
+import logging
+log = logging.getLogger("anaconda")
+
+
+class Prompt(object):
+    """Class to create a prompt message with options."""
+
+    # Default message of the prompt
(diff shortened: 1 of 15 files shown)


pyanaconda/ui/tui/simpleline/prompt.py
```diff
@@ -0,0 +1,140 @@
...
+        :type message: str|None
+        """
+        self.message = message
+
+    def add_option(self, key, description):
+        """Add an option to the prompt.
+        Causes a warning if the option already exists.
+
+        :param key: the key for choosing the option
+        :type key: str
```
M4rtinK: I'll just note that while a separate `:type key: str` definition is 100% valid, it is also possible to specify the type inline when the type is not ambiguous, like this:

`:param str key: the key for choosing the option`

Or an advanced example from the Sphinx docs:
http://www.sphinx-doc.org/en/1.4.8/domains.html#info-field-lists

poncovka: Ok, thanks.

pyanaconda/ui/tui/simpleline/widgets.py
```diff
@@ -19,12 +19,222 @@
...
+        return functools.reduce(lambda acc, l: max(acc, len(l)), self._buffer, 0)
+
+    def clear(self):
+        """Clears this widgets buffer and resets cursor."""
+        self._buffer = list()
+        self._cursor = (0, 0)
+
+    @property
+    def content(self):
+        """This has to return list (rows) of lists (columns) with one character elements."""
```
jkonecny12: "Returns list...." seems better to me

pyanaconda/ui/tui/simpleline/widgets.py
```diff
@@ -19,12 +19,222 @@
...
+        return self._buffer
+
+    def render(self, width):
+        """This method has to redraw the widget's self._buffer.
```
jkonecny12: Please, try to avoid word "This" or it will be everywhere.

pyanaconda/ui/tui/simpleline/widgets.py
```diff
@@ -19,12 +19,222 @@
...
+
+        :param width: the width of buffer requested by the caller
+        :type width: int
+
+        This method will commonly call render of child widgets and then draw and write
```
jkonecny12: Same here.

pyanaconda/ui/tui/simpleline/widgets.py
poncovka: The `Widget `class is not mine. I've moved it from `base.py` to `widgets.py` to be able to use widgets in `base.py`. Should I change the comments anyway?

jkonecny12: I think it would be better, after all you are the one who will be blamed by Git now :).

poncovka: Fixed.

pyanaconda/ui/tui/simpleline/widgets.py
poncovka: Fixed.

poncovka:
Merged only to master because of the final freeze.

PR #858, opened 2016-11-03
network: index team slave connection names starting with 1

rvykydal:
Same as with bond slaves and as NetworkManager naming in general.


rvykydal:
The test should be fixed now.

PR #860, opened 2016-11-04
Fix the basis for percentage-based growth of LVs.

dwlehman:
Resolves: rhbz#1383817


rvykydal:
Looks good to me, thanks for the patch.

PR #861, opened 2016-11-07
tui: Available help system

poncovka:
The help system is now available from tui. Spokes with a help to
display have an extra option 'h' in their prompt.

### DIFF
--- a/pyanaconda/ui/tui/hubs/__init__.py
+++ b/pyanaconda/ui/tui/hubs/__init__.py
@@ -16,11 +16,13 @@
 # License and may only be used or replicated with the express permission of
 # Red Hat, Inc.
 #
+from pyanaconda import ihelp
+from pyanaconda.constants_text import INPUT_PROCESSED, INPUT_DISCARDED
 from pyanaconda.ui.tui import simpleline as tui
-from pyanaconda.ui.tui.tuiobject import TUIObject
+from pyanaconda.ui.tui.tuiobject import TUIObject, HelpScreen
 from pyanaconda.ui import common
 
-from pyanaconda.i18n import _, C_, N_
+from pyanaconda.i18n import _, N_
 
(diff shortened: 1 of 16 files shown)


pyanaconda/ui/tui/hubs/__init__.py
```diff
@@ -113,6 +114,12 @@ def input(self, args, key):
                     if not spoke.completed and spoke.mandatory:
                         print(_("Please complete all spokes before continuing"))
                         return False
+            # TRANSLATORS: 'h' to help
+            elif key == C_('TUI|Spoke Navigation', 'h'):
+                if self.has_help:
+                    help_path = ihelp.get_help_path(self.helpFile, self.instclass, True)
+                    self.app.switch_screen_modal(HelpScreen(self.app, help_path))
+                    return True
```
vpodzime: I know there's just ``False`` above, but I believe there were some constants for ``input()`` method's return values which made more sense than ``True, False``. Could you please check that and change this and the new ``input`` method added below to use the constants?

pyanaconda/ui/tui/spokes/langsupport.py
```diff
@@ -122,7 +123,7 @@ def input(self, args, key):
             self.app.switch_screen(self, None)
             return True
         else:
-            return key
+            return NormalTUISpoke.input(self, args, key)
```
jkonecny12: Isn't there a way how to solve this globally? I mean in some base class or similar place.

pyanaconda/ui/tui/hubs/__init__.py
poncovka: Fixed.

pyanaconda/ui/tui/spokes/langsupport.py
poncovka: Fixed.

PR #864, opened 2016-11-08
Don't write user interaction config for image & directory install (#1…

M4rtinK:
…379106)

Don't write out the user interaction config file for image and directory
installation modules.

In these modes Anaconda actually prepares the installation image,
so storing the config file could result in spokes being hidden
during the actual installation.

For example if the storage spokes was visited during an interactive
image or directory installation, it would not be shown during the
actual installation run from the generated image/directory,
as it would be considered already visited.

### DIFF
--- a/pyanaconda/install.py
+++ b/pyanaconda/install.py
@@ -130,8 +130,17 @@ def doConfiguration(storage, payload, ksdata, instClass):
     else:
         _writeKS(ksdata)
 
-    # write out the user interaction config file
-    screen_access.sam.write_out_config_file()
+    # Write out the user interaction config file.
+    #
+    # But make sure it's not written out in the image and directory installation mode,
+    # as that might result in spokes being inadvertedly hidden when the actual installation
+    # startes from the generate image or directory contents.
+    if flags.flags.imageInstall:
+        log.info("Not writing out user interaction config file due to image install mode.")
+    elif flags.flags.dirInstall:
+        log.info("Not writing out user interaction config file due to directory install mode.")
+    else:
+        screen_access.sam.write_out_config_file()
 
     progress_complete()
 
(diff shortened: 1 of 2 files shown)


M4rtinK:
Added another fix to the PR - don't read the user interaction config from the host during image and directory installs, for obvious reasons.

pyanaconda/install.py
```diff
@@ -130,8 +130,17 @@ def doConfiguration(storage, payload, ksdata, instClass):
     else:
         _writeKS(ksdata)
 
-    # write out the user interaction config file
-    screen_access.sam.write_out_config_file()
+    # Write out the user interaction config file.
+    #
+    # But make sure it's not written out in the image and directory installation mode,
+    # as that might result in spokes being inadvertedly hidden when the actual installation
+    # startes from the generate image or directory contents.
```
vpodzime: "*starts*". But I must say I don't really understand this. Shouldn't the configuration file go to the installed image or directory? I don't think we can rely on some assumption that every image or directory installation is then used to run installation again.

pyanaconda/screen_access.py
```diff
@@ -56,10 +57,21 @@ def __init__(self):
...
+            # directory installation modes.
+            # The config would be taken from the host system,
+            # which is certainly not what we would want to happen.
+            # But load the config if a path is specified,
+            # so that it is possible to hide spokes in
+            # image and directory installation modes.
+            image_or_dir_install = flags.imageInstall or flags.dirInstall
+            if config_path is None and not image_or_dir_install:
+                config_path = CONFIG_FILE_PATH
+            if config_path and os.path.exists(config_path):
```
jkonecny12: Wouldn't it be better to use `flags.can_touch_runtime_system()` for this?

M4rtinK: Good point - I've changed it to use can_touch_runtime_system(). :)

pyanaconda/install.py
vpodzime: Works for me.

jkonecny12:
Just a note, you are targeting master branch now. It would be nice to switch that if we get the freeze exception here.

PR #869, opened 2016-11-15
Branch specific things can now be programable decided by git-find-branch script

jkonecny12:
Because of our new branching policy all the code from `f**-devel` branch will be in the master branch. This is good for code but it is bad when we need specific things like `zanata` or `mock`. When we change `zanata.xml` on f25-devel the change will propagate to master branch and then we are pushing translation from master to f25 branch on `zanata`. 

New script `git-find-branch` is implemented to help us with this problem. This small script will resolve our actual branch to our main development branches (`f25-devel`, `master`, `unstable`, `rhel7-branch`...) and return the branch name without the `-*` suffix. So `f25-release` and `f25-devel` will be `f25`, this seems easier to me and this behavior can be of course changed in future. The `configure` can then substitute this variable to any file where this is needed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -82,3 +82,4 @@ modules.rst
 pyanaconda.*rst
 .coverage
 docs/tests/*.rst
+zanata.xml
--- a/scripts/git-find-branch
+++ b/scripts/git-find-branch
@@ -1,0 +1,104 @@
+#!/bin/bash
+#
+# git-find-branch - We need to have branch specific decision which will be
+#                   based on the main development branches. This script
+#                   will get nearest main branch and return it for others.
+#
+# Copyright (C) 2016  Red Hat, Inc.
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU Lesser General Public License as published
+# by the Free Software Foundation; either version 2.1 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Lesser General Public License for more details.
+#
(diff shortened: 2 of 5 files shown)


scripts/git-find-branch
```diff
@@ -0,0 +1,101 @@
...
+#
+# Red Hat Author(s): Jiri Konecny <jkonecny@redhat.com>
+#
+#
+# There are two ways how find parent branch of this project:
+#
+# 1) When environment variable ANACONDA_TARGET_BRANCH is set, use this
+#    variable as output without any checks. This is for automated tests
+#    mainly.
+# 2) Look on the Git find all interesting branches for us. Test if
```
vpodzime: "*Look at Git...*"

scripts/git-find-branch
```diff
@@ -0,0 +1,101 @@
...
+#    actual branch is not on of those and if not look on the parent and
+#    test parent. Repeat the process with parents.
+#
+#
+# WARNING!
+# This won't work without a merge commit! Git don't have any information
```
vpodzime: "*Git doesn't have...*"

scripts/git-find-branch
```diff
@@ -0,0 +1,101 @@
...
+# which branch is the parent without the merge commit, both branches tips
```
vpodzime: "*tips of both branches...*"

.gitignore
```diff
@@ -82,3 +82,4 @@ modules.rst
 pyanaconda.*rst
 .coverage
 docs/tests/*.rst
+zanata.xml
```
vpodzime: I think this also needs to be added to ``MAINTAINERCLEANFILES`` in the ``Makefile.am`` (just with the ``.in`` suffix).

jkonecny12:
**Updated:**
I hope everything will be alright now :)

vpodzime:
> Updated:
> I hope everything will be alright now :)

It is. :)

PR #871, opened 2016-11-16
gui: show supported locales on Atomic Host installs

jlebon:
In RHELAH 7.3.2, we will be adding all the officially supported locales
into the tree (see RHBZ#1186757). This means that users should now be
able to specify their language at install time, rather than defaulting
to en_US (which was done for RHBZ#1235726).

One issue right now is that many locales are presented, i.e. beyond what
is officially supported, which means that the original concerns in
RHBZ#1235726 are still valid. E.g. I shouldn't be able to select
"Asturian" if "ast_ES.utf8" isn't available in the tree.

In the future, we should be limiting the available languages to what the
install payload actually contains.

Related: https://bugzilla.redhat.com/show_bug.cgi?id=1186757
Related: https://bugzilla.redhat.com/show_bug.cgi?id=1235726

### DIFF
--- a/pyanaconda/installclasses/rhel.py
+++ b/pyanaconda/installclasses/rhel.py
@@ -17,6 +17,7 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #
 
+import os
 import logging
 log = logging.getLogger("anaconda")
 
@@ -24,6 +25,7 @@ from pyanaconda.installclass import BaseInstallClass
 from pyanaconda.product import productName
 from pyanaconda import network
 from pyanaconda import nm
+from pyanaconda import iutil
 from pyanaconda.kickstart import getAvailableDiskSpace
 from pyanaconda.flags import flags
 from blivet.partspec import PartSpec
@@ -76,11 +78,9 @@ class RHELAtomicInstallClass(RHELBaseInstallClass):
     sortPriority=21000
     hidden = not productName.startswith(("RHEL Atomic Host", "Red Hat Enterprise Linux Atomic"))
 
-    def configure(self, anaconda):
-        RHELBaseInstallClass.configure(self, anaconda)
-        # Atomic installations are always single language (#1235726)
-        log.info("Automatically enabling single language mode for %s installation.", self.name)
-        flags.singlelang = True
+    def __init__(self):
+        self.localemap = {} # loaded lazily
+        RHELBaseInstallClass.__init__(self)
 
     def setDefaultPartitioning(self, storage):
         autorequests = [PartSpec(mountpoint="/", fstype=storage.defaultFSType,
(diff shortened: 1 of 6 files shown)


M4rtinK:
We now have an ACKed bug for this change in Anaconda:
https://bugzilla.redhat.com/show_bug.cgi?id=1429576

So please add the line "Resolves: rhbz#1429576" your commit in the PR; after than we can merge it.

jlebon:
But this is still an issue right?

> One issue right now is that many locales are presented, i.e. beyond what
is officially supported, which means that the original concerns in
RHBZ#1235726 are still valid. E.g. I shouldn't be able to select
"Asturian" if "ast_ES.utf8" isn't available in the tree.

So merging this would cause a regression wrt https://bugzilla.redhat.com/show_bug.cgi?id=1235726. Would it be possible to hardcode the list of supported languages for now until we do this properly (i.e. inspect the payload)?

jlebon:
Sorry for the turnabout! It just feels like if we change the behaviour again, it should be towards a more correct one. If hardcoding the list is acceptable, I can take a look at doing that and update the PR.

M4rtinK:
> Sorry for the turnabout! It just feels like if we change the behaviour again, it should be towards a more correct one. If hardcoding the list is acceptable, I can take a look at doing that and update the PR.

I think it's fine as long as the hardcoding is temporary (eq. is eventually replaced by payload introspection) & the hardcoded locale list is part of the Atomic install class.

As for the implementation I would suggest adding a simple "supported languages" API to the base install class, which by default does no filtering (a supported() function that always returns True, etc.). The Atomic install class would then implement the API based on its hardcoded list of supported languages.

Also - how stable would the hardoced list be ?

I'm asking because if the list would be part of the Anaconda code base any change would generally have to go though a pull request & it might not be possible to update it at all if the given release branch is in a freeze. I guess an emergency downstream hotfix patch might be used though.

jlebon:
OK, I rebased and updated the patch to inspect the ostree. Though note that it requires https://github.com/rhinstaller/lorax/pull/194.

Can we port this to Fedora as well? I presume I can still open it against `f26-devel`?

M4rtinK:
Please add the machine readable bug reference at the end of the commit message:

Resolves: rhbz#1429576

jlebon:
OK, commit message updated!

jlebon:
I had forgotten to update the docs for `inst.singlelang`. Done in ⬆️.

pyanaconda/installclasses/rhel.py
```diff
@@ -104,3 +103,52 @@ def setDefaultPartitioning(self, storage):
...
+        return supported
+
+    def _initialize_localemap(self, ref):
+
+        if self.localemap:
+            return
+
+        iutil.mkdirChain("/install/ostree/tmp/usr/lib")
+        rc = iutil.execWithRedirect("/usr/bin/ostree",
+            ["checkout", "--repo", "/install/ostree", ref,
```
rvykydal: Well, I wonder if we should get the /install/ostree location from --url of ostreesetup command (supplied currently by interactive-defaults.ks via installer iso compose process), ie ksdata.ostreesetup.url so it survives possible changes in iso building process, or applies to user-supplied repo.

jlebon: Good idea! Done below ⬇️.

rvykydal: Thank you, looks good to me now.

M4rtinK:
The Lorax part has been merged and released, so merging this PR now.

PR #872, opened 2016-11-17
Don't crash if the UIC file can't be written (#1375702)

M4rtinK:
Don't crash if the user interaction config file can't be written,
which can happen for example on Atomic (read-only rootfs by default).

Print and error message and log the exception instead.

### DIFF
--- a/pyanaconda/screen_access.py
+++ b/pyanaconda/screen_access.py
@@ -93,12 +93,15 @@ class ScreenAccessManager(object):
 
         with self._lock:
             new_config_file = not os.path.exists(config_path)
-            with open(config_path, "wt") as f:
-                if new_config_file:
-                    # we are creating a new file, so add a header that it was created by Anaconda,
-                    # including its version number
-                    f.write(self._get_new_config_header())
-                self._config.write(f)
+            try:
+                with open(config_path, "wt") as f:
+                    if new_config_file:
+                        # we are creating a new file, so add a header that it was created by Anaconda,
+                        # including its version number
+                        f.write(self._get_new_config_header())
+                    self._config.write(f)
+            except OSError:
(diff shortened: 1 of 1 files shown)


pyanaconda/screen_access.py
```diff
@@ -93,12 +93,15 @@ def write_out_config_file(self, config_path=None):
...
-                self._config.write(f)
+            try:
+                with open(config_path, "wt") as f:
+                    if new_config_file:
+                        # we are creating a new file, so add a header that it was created by Anaconda,
+                        # including its version number
+                        f.write(self._get_new_config_header())
+                    self._config.write(f)
+            except OSError:
+                log.exception("Can't write user interaction config file.")
```
vpodzime: Shouldn't this be a warning instead?

M4rtinK: The open() can generate quite a lot of various exceptions - so the general idea is to for now just to prevent the crash & log the exception. Later on (and possibly once we have more data) we can narrow down the scope of the except clause & log  specific warnings/errors.

vpodzime: Sounds good to me, but shouldn't this also log details about the exception then? Or does ``log.exception()`` do that somehow?

M4rtinK: Yep - `log.exception()` logs the message you give it and then the full traceback from the exception caught by the except block. From [the docs](https://docs.python.org/3.5/library/logging.html#logging.Logger.exception):

```
Logger.exception(msg, *args, **kwargs)

    Logs a message with level ERROR on this logger. The arguments are interpreted as for debug(). Exception info is added to the logging message. This method should only be called from an exception handler.
```

vpodzime: nice!

PR #875, opened 2016-11-18
Fix calling of can_touch_runtime_system function

jkonecny12:
The `msg` parameter is mandatory.


jkonecny12:
I'll merge this to f25-devel with the new PR.

PR #877, opened 2016-11-21
Allow install classes to set alternate states for firstboot

jcpunk:
In my install class I'd like to be able to say "Just don't run firstboot unless someone actually asked for it".  Currently there isn't an easy way of setting that up.  Interactive installs get it automatically (if unit exists).

I don't want to outright remove it from the release.  This attribute seemed like a viable middle option.


jkonecny12:
Jenkins, it's ok to test.

PR #879, opened 2016-11-23
Move Anaconda tests to mock

jkonecny12:
To run tests in mock use `make isolated-ci`.

This is using copr repository from vtrefny so our packages will be
one day old from this repo and will fallback on Rawhide version when
Rawhide has newer packages (broken builds on COPR for example).

To run these tests the `nosync` package is required, it is for disabling `fsync`
in the mock and it will give us great performance boost. When you are
on a multilib (x86_64) system you need to install x86_64 and i686 versions
of the `nosync` library.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -31,7 +31,7 @@ MAINTAINERCLEANFILES =	Makefile.in config.guess config.h.in config.sub \
 						py-compile m4/* po/Makefile.in.in po/Rules-quot \
 						test-driver
 
-CLEANFILES = *~
+CLEANFILES = *~ utils/mock-out.cfg
 
 dist_noinst_DATA      = $(PACKAGE_NAME).spec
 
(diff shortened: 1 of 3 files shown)


rvykydal:
Looks good to me

Makefile.am
```diff
@@ -48,11 +48,17 @@ ARCHIVE_TAG   = $(PACKAGE_NAME)-$(PACKAGE_VERSION)-$(PACKAGE_RELEASE)
...
+				-e 's/master/rawhide/')
+ARCH_NAME ?= $(shell uname -m)
+
 RC_RELEASE ?= $(shell date -u +0.1.%Y%m%d%H%M%S)
-MOCKCHROOT ?= fedora-$(shell echo "$(GIT_BRANCH)" | sed \
-						-e 's/^f//' \
-						-e 's/master/rawhide/'\
-						)-$(shell uname -m)
+
+MOCK_CONF_TEMPLACE ?= $(srcdir)/utils/mock-template.cfg
```
jkonecny12: Should be fixed, thank you.

PR #880, opened 2016-11-23
Fix generated zanata.xml from https unstable branch

jkonecny12:
Protocol https doesn't end with `.git` suffix.

We need to have master instead of unstable here.


jkonecny12:
jenkins, test this please

jkonecny12:
jenkins, test this please

PR #884, opened 2016-11-24
Fix wrong attribute of AnacondaYumRepo (#1395076)

poncovka:
The name of the attribute is id, not repoid.

Resolves: rhbz#1395076


poncovka:
Bugzilla updated.

PR #885, opened 2016-11-25
Add checks to git-find-branch script

jkonecny12:
If remote or actual branch can't be found then print error to `stderr` and kill the script. `zanata.xml` file will have blank project tag.


jkonecny12:
jenkins, test this please

PR #891, opened 2016-11-30
Handle unexpected storage exception from blivet

jkonecny12:


vpodzime:
This unfortunately looks good to me. Could you please also file an issue for blivet that it requires it's users to do such things? Thanks!

jkonecny12:
Updated:
I also add solution to our problem with the LVM in `storage.py`.

jkonecny12:
@vpodzime issue has been created https://github.com/rhinstaller/blivet/issues/528

PR #895, opened 2016-12-02
Fix invalid characters in an installed package summary (#1324706)

poncovka:
Characters '<' and '>' in an installed package summary can cause
GtkWarning: Failed to set text from markup due to error parsing
markup.

Resolves: rhbz#1324706

### DIFF
--- a/yuminstall.py
+++ b/yuminstall.py
@@ -165,6 +165,12 @@ class AnacondaCallback:
             s = to_unicode(_("<b>Installing %(pkgStr)s</b> (%(size)s)\n")) \
                     % {'pkgStr': pkgStr, 'size': size_string(hdr['size'])}
             summary = to_unicode(gettext.ldgettext("redhat-dist", hdr['summary'] or ""))
+
+            # summary should not contain these characters
+            if any(c in summary for c in ('<', '>')):
+                log.warning("Package %s has an invalid summary: %s" % (pkgStr, summary))
+                summary = ""
+
             s += summary.strip()
             self.progress.set_label(s)
 


yuminstall.py
```diff
@@ -165,6 +165,12 @@ def callback(self, what, amount, total, h, user):
             s = to_unicode(_("<b>Installing %(pkgStr)s</b> (%(size)s)\n")) \
                     % {'pkgStr': pkgStr, 'size': size_string(hdr['size'])}
             summary = to_unicode(gettext.ldgettext("redhat-dist", hdr['summary'] or ""))
+
+            # summary should not contain these characters
+            if any(c in summary for c in ('<', '>')):
+                log.debug("Package " + pkgStr + " has an invalid summary: " + summary)
+                summary = ""
```
jkonecny12: Shouldn't it be better to just replace these symbols with something else with warning message instead?

vpodzime: +1 to this. Maybe by the SGML/XML entities so that Gtk just shows them correctly?

poncovka:
So after the discussion, it was decided to replace the gtk warning with anaconda warning and the problematic summary message with empty string.

yuminstall.py
```diff
@@ -165,6 +165,12 @@ def callback(self, what, amount, total, h, user):
             s = to_unicode(_("<b>Installing %(pkgStr)s</b> (%(size)s)\n")) \
                     % {'pkgStr': pkgStr, 'size': size_string(hdr['size'])}
             summary = to_unicode(gettext.ldgettext("redhat-dist", hdr['summary'] or ""))
+
+            # summary should not contain these characters
+            if any(c in summary for c in ('<', '>')):
+                log.warning("Package " + pkgStr + " has an invalid summary: " + summary)
```
jkonecny12: Please change this to:
```
log.warning("Package %s  has an invalid summary: %s", pkgStr, summary)
```

poncovka:
Fixed.

PR #897, opened 2016-12-05
Add none user interface.

poncovka:
None user interface (nui) should work the same way as the noninteractive tui, but with limited output. It is a substep to the further modularization. It starts with `inst.noui` option.


poncovka:
In case of failure:
nui_fail

In case of success:

The messages between the dots are not supposed to be there, it is a reported bug in `rpm`.

nui_success

jkonecny12:
jenkins, test this please

jkonecny12:
jenkins, test this please

cgwalters:
That's a scary amount of new code for this.  Can't this be a flag in the tui code or something?

jkonecny12:
Almost all the code here is appending and it will be executed only in a special mod. So or so it's now in discussion if this patch will be merged as it is now or it will be reworked.

PR #898, opened 2016-12-06
Structured installation tasks

M4rtinK:
The first two commits add structured installation task classes and the Signal class they use.
The third commit the installation phase to use the structured installation task classes.

### DIFF
--- a/pyanaconda/install.py
+++ b/pyanaconda/install.py
@@ -22,7 +22,7 @@ from blivet import callbacks
 from blivet.osinstall import turn_on_filesystems
 from blivet.devices import BTRFSDevice
 from pyanaconda.bootloader import writeBootLoader
-from pyanaconda.progress import progress_report, progress_message, progress_step, progress_complete, progress_init
+from pyanaconda.progress import progress_message, progress_step, progress_complete, progress_init
 from pyanaconda.users import Users
 from pyanaconda import flags
 from pyanaconda import iutil
@@ -34,8 +34,29 @@ from pyanaconda.threads import threadMgr
 from pyanaconda.ui.lib.entropy import wait_for_entropy
 from pyanaconda.kickstart import runPostScripts, runPreInstallScripts
 from pyanaconda.kexec import setup_kexec
+from pyanaconda.install_tasks import Task, TaskQueue
 import logging
 log = logging.getLogger("anaconda")
+
+
+class WriteResolvConfTask(Task):
+    """Custom task subclass for handling the resolv.conf copy task.
+
+    The main reason is to resolve the sysroot path right before the
+    copy operation, not at task & task queue creation time.
+
+    Secondary reason is to demonstrate how a lightweight Task subclass can be used.
+    """
+
+    def do_run(self):
+        """Resolve the sysroot path only right before doing the copy operatio.
+
+        If we just added the sysroot path as an argument, it would be resolved when the
+        task queue was created, not when the task is actually executed, which could
+        theoretically result in an incorrect path.
+        """
+        network.copyFileToPath("/etc/resolv.conf", iutil.getSysroot())
+
 
 def _writeKS(ksdata):
     path = iutil.getSysroot() + "/root/anaconda-ks.cfg"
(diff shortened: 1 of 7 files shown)


M4rtinK:
jenkins, test this please

rvykydal:
Looks good to me.

pyanaconda/install_tasks.py
```diff
@@ -0,0 +1,400 @@
...
+            self._callable()
+            # trigger the "completed" signal
+            self.completed(self)
+            # the task should be done, set the task state accordingly
+            with self._lock:
+                self._running = False
+                self._done = True
+                self._done_timestamp = time.time()
+
+def item_counter(item_count):
```
jkonecny12: Please move this global function to the top of this file. AFAIK it is like that in the other parts of our code.

pyanaconda/install_tasks.py
```diff
@@ -0,0 +1,400 @@
...
+        return self._status_message
+
+    @property
+    def running(self):
+        """Reports if the task queue processing is currently running.
+
+        :returns: if the task queue is currently running
+        :rtype: bool
+        """
+        with self._lock:
```
vpodzime: Maybe a decorator would be better than having (almost) everything in a ``with self._lock`` block?

pyanaconda/install_tasks.py
```diff
@@ -0,0 +1,400 @@
...
+            message += "Number of tasks: %d\n" % self.task_count
+            message += "Task & task group listing:\n"
+        else:
+            message = "Task queue: %s\n" % self.name
+        for item in self:
+            for line in item.summary.splitlines():
+                message += " %s\n" % line
+        # remove trailing newlines from the top level message
+        if self.parent is None and message[-1] == "\n":
+            message = message[0: -1]
```
vpodzime: ``message.rstrip("\n")``?

pyanaconda/install_tasks.py
vpodzime: I believe it should go to ``iutil.py``

pyanaconda/isignal.py
```diff
@@ -0,0 +1,54 @@
+""" A signal/slot implementation
+
+File:    signal.py
```
vpodzime: Wrong file name.

pyanaconda/install.py
```diff
@@ -51,74 +52,79 @@ def _writeKS(ksdata):
...
+    """Configure the installed system."""
+
+    configuration_queue = TaskQueue("Configuration queue")
+    # connect progress reporting
+    configuration_queue.queue_started.connect(lambda x: progress_message(x.status_message))
+    configuration_queue.queue_completed.connect(lambda x: progress_step("%s -- DONE" % x.status_message))
+
+    # schedule the execute methods of ksdata that require an installed system to be present
+    os_config = TaskQueue("Installed system configuration", N_("Configuring installed system"))
+    os_config.append(Task("Configure authconfig", lambda: ksdata.authconfig.execute(storage, ksdata, instClass)))
```
vpodzime: These ``lambda``s look really weird. Wouldn't it be better to pass the function/method and its attributes as separate things?

pyanaconda/install.py
```diff
@@ -51,74 +52,79 @@ def _writeKS(ksdata):
...
+    os_config.append(Task("Configure firewall", lambda: ksdata.firewall.execute(storage, ksdata, instClass)))
+    os_config.append(Task("Configure X", lambda: ksdata.xconfig.execute(storage, ksdata, instClass)))
+    os_config.append(Task("Configure skip-X", lambda: ksdata.skipx.execute(storage, ksdata, instClass)))
+    configuration_queue.append(os_config)
+
+    # schedule network configuration (if required)
+    will_write_network = not flags.flags.imageInstall and not flags.flags.dirInstall
+    if will_write_network:
+        network_config = TaskQueue("Network configuration", N_("Writing network configuration"))
+        network_config.append(Task("Network configuration", lambda: ksdata.network.execute(storage, ksdata, instClass)))
```
M4rtinK: Wrapped.

pyanaconda/install_tasks.py
```diff
@@ -0,0 +1,400 @@
...
+        """Task queue name.
+
+        :returns: task queue name
+        :rtype: str
+        """
+        return self._name
+
+    @property
+    def status_message(self):
+        return self._status_message
```
M4rtinK: Added.

pyanaconda/install_tasks.py
```diff
@@ -0,0 +1,400 @@
...
+
+    def insert(self, index, item):
+        item.parent = self
+        super(TaskQueue, self).insert(index, item)
+
+    def __setitem__(self, index, item):
+        item.parent = self
+        super(TaskQueue, self).__setitem__(index, item)
+
+    def start(self):
```
M4rtinK: I see it as starting the execution/processing of the Task or TaskQueue that then runs (as indicated by the `running` property) to completion (as indicated by the `completed` property). Also TaskQueues don't care what they are iterating over - they just call `start()` on it, so renaming it just for the TaskQueue would be problematic.

So I think `start` is fine. :)

pyanaconda/install_tasks.py
```diff
@@ -0,0 +1,400 @@
...
+            self.completed(self)
+
+
+class Task(object):
+    """Task is a wrapper for a single installation related task.
+
+    It has a name and a callable, which is the actual task to execute,
+    and some state variables to check if the task is running or
+    if it is already done.
+    The task can started by running the start() method. This can be done
```
M4rtinK: Fixed! Thanks for spotting the typo. :)

pyanaconda/install_tasks.py
```diff
@@ -0,0 +1,400 @@
...
+        self.completed = Signal()
+
+    @property
+    def name(self):
+        """Task name.
+
+        :returns: task name
+        :rtype: str
+        """
+
```
M4rtinK: Removed.

pyanaconda/install_tasks.py
```diff
@@ -0,0 +1,400 @@
...
+        return self._parent
+
+    @parent.setter
+    def parent(self, parent_item):
+        self._parent = parent_item
+        # Connect own signals "up" to the parent
+        self.started.connect(self._parent.task_started)
+        self.completed.connect(self._parent.task_completed)
+
+    def start(self):
```
M4rtinK: Done. :)

pyanaconda/install_tasks.py
M4rtinK: Yeah, it's not really task/task queue specific & could be useful elsewhere, so moving to iutil. :)

pyanaconda/install_tasks.py
```diff
@@ -0,0 +1,400 @@
...
+    def append(self, item):
+        item.parent = self
+        super(TaskQueue, self).append(item)
+
+    def insert(self, index, item):
+        item.parent = self
+        super(TaskQueue, self).insert(index, item)
+
+    def __setitem__(self, index, item):
+        item.parent = self
```
M4rtinK: Implemented in the setter for the parent property, so no need to do it individually. :)

pyanaconda/isignal.py
M4rtinK: Fixed. :)

pyanaconda/install_tasks.py
M4rtinK: Good point. :)

pyanaconda/install_tasks.py
jkonecny12: AFAICT this won't help you with all the cases. Now you will disconnect the old item when there will be a new item as replacement but what about simple pop. Nothing is added so the setter won't be called.

pyanaconda/install_tasks.py
jkonecny12: I thought about renaming all the `start` methods (inside of all classes) but it cloud be `start`, I don't mind.

pyanaconda/isignal.py
```diff
@@ -0,0 +1,54 @@
...
+
+class Signal(object):
+    def __init__(self):
+        # The original implementation used WeakSet to store functions,
+        # but that causes lambdas without any other reference to be
+        # garbage collected. So we use a normal set to avoid that.
+        self._functions = set()
+        self._methods = WeakKeyDictionary()
+
+    def __call__(self, *args, **kargs):
```
poncovka: Maybe there could be `emit` method instead of `__call__`, because `signal.emit()` looks more clear to me than just `signal()`. For example Qt also uses emit: http://doc.qt.io/qt-4.8/signalsandslots.html

pyanaconda/install_tasks.py
```diff
@@ -0,0 +1,424 @@
...
+    def __init__(self, name, status_message=None):
+        super(TaskQueue, self).__init__()
+        self._name = name
+        self._status_message = status_message
+        self._lock = RLock()
+        self._current_task_number = None
+        self._current_queue_number = None
+        self._running = False
+        self._done = False
+        self._parent = None
```
poncovka: Could `TaskQueue` be running and done at the same time? Maybe, the attributes `self._running` and `self._done` could be merged into one `self._state` attribute. Same for `Task`.

pyanaconda/install_tasks.py
```diff
@@ -0,0 +1,424 @@
...
+# Red Hat, Inc.
+#
+from threading import RLock
+from pyanaconda.isignal import Signal
+import time
+import logging
+log = logging.getLogger("anaconda")
+
+
+class TaskQueue(list):
```
poncovka: Shouldn't `Task` and `TaskQueue` share a common interface or a base class?

pyanaconda/isignal.py
M4rtinK: While I still like how you can just call the signal itself, I do agree that a separate `emit()` methods is more consistent. We already have `connect()`, `disconnect()` and `clear()`, so also having a method for triggering the signal makes sense.

Looks like [even GTK3 has an `emit()` method](http://python-gtk-3-tutorial.readthedocs.io/en/latest/objects.html#create-new-signals) for signals used from Python. :)

pyanaconda/install.py
jkonecny12: I wouldn't like to see mixing of args and lambdas just because I have there kwargs. Is there any reason why not to create ``args=(), kwargs={}`` like for example in ``Thread.__init__()``.

https://docs.python.org/3/library/threading.html#threading.Thread

pyanaconda/isignal.py
poncovka: Also with `emit`, you immediately see in a code that you are dealing with a signal and not a function.

M4rtinK: Yep, that's another benefit. :)

pyanaconda/install_tasks.py
M4rtinK: Nope, running and done should never be True at the same time. This is how it should go (running, done):
False, False - not yet started
True, False - once start has been called & as long as processing is ongoing
False, True - once finished

Switching between these states should be (hopefully ;-) ) atomic. :)

pyanaconda/install_tasks.py
M4rtinK: I've added the BaseTask class as a parent for Task and TaskQueue to share the common code. Task queue also no longer subclasses a list, but just [emulates it](https://docs.python.org/3/reference/datamodel.html#emulating-container-types) - this has the added benefit that we can skip implementing the list methods we *don't* want to support, such as `extend()` or `reverse()` and `sort()`.

pyanaconda/install.py
M4rtinK: The `task_kwargs` option has been added. ;-)

pyanaconda/install_tasks.py
M4rtinK: After investigating this a bit it doesn't seem that easy:
- AFAIK a lock instance can't be directly used as a decorator
- a custom decorator could be used as long **as the lock is named consistently** (eq. self._lock) - seem like quite a hack though :)

Based on [this](https://github.com/GrahamDumpleton/wrapt/blob/develop/blog/07-the-missing-synchronized-decorator.md#synchronizing-off-a-thread-mutex):

```
import functools
from threading import RLock

def synchronized(wrapped):
    @functools.wraps(wrapped)
    def _wrapper(*args, **kwargs):
        with args[0]._lock:
            return wrapped(*args, **kwargs)
    return _wrapper

class Test(object):
    def __init__(self):
        self._lock = RLock()

    @synchronized
    def function(self):
        print(self._lock)
        print("Hello world!")


test = Test()
test.function()
```

This actually works and is pretty lightweight, just a little crazy on the implementation side. If concerned parties are fine with this I can use this to replace the current locking usage on most places in the tasks PR. :)

vpodzime: The above looks good to me, I'd just explicitly used the ``self`` parameter as follows:
```
def synchronized(wrapped):
    @functools.wraps(wrapped)
    def _wrapper(self, *args, **kwargs):
        with self._lock:
            return wrapped(self, *args, **kwargs)
    return _wrapper
```

and documented in the docstring that the decorator can only be used on methods.

M4rtinK:
All suggestions (lock decorators, common base class, passing args/kwargs to callables, etc.) should now be implemented. :)

pyanaconda/install.py
```diff
@@ -34,9 +34,33 @@
...
+    """Custom task subclass for handling the resolv.conf copy task.
+
+    The main reason is to resolve the sysroot path right before the
+    copy operation, not at task & task queue creation time.
+
+    Secondary reason is to demonstrate how a lightweight Task subclass can be used.
+    """
+
+    def __init__(self, *args, **kwargs):
+        super(WriteResolvConfTask, self).__init__(*args, **kwargs)
```
vpodzime: An ``__init__`` like this can be dropped, AFAICT.

pyanaconda/install.py
```diff
@@ -210,57 +266,96 @@ def doInstall(storage, payload, ksdata, instClass):
...
+
+        if willInstallBootloader:
+            packages += storage.bootloader.packages
+
+        # don't try to install packages from the install class' ignored list and the
+        # explicitly excluded ones (user takes the responsibility)
+        packages = [p for p in packages
+                    if p not in instClass.ignoredPackages and p not in ksdata.packages.excludedList]
+        payload.preInstall(packages=packages, groups=payload.languageGroups())
+    pre_install.append(Task("Find additional & run preInstall()", run_pre_install))
```
vpodzime: "*additional packages*"?

pyanaconda/install.py
M4rtinK: OK, dropped. :)

PR #903, opened 2016-12-09
rpmostreepayload: Rework binds to be recursive

cgwalters:
Our hardcoded list of mount points missed critical ones
like `/sys/firmware/efi/efivars`, which broke installation
on EFI.

I started to add that to the list, but then I realized there's a much
simpler solution - make the binds recursive and let `mount` do the
work for us.  This is a bit more risky of a change, but it'll
clearly be more maintainable long term.

Down the line...we may want to simply walk over all the toplevel mount
points in `/mnt/sysimage`, but I doubt anyone is going to add anything
new that's not a subdirectory of `/sys`.

See https://pagure.io/atomic-wg/issue/185

### DIFF
--- a/pyanaconda/packaging/rpmostreepayload.py
+++ b/pyanaconda/packaging/rpmostreepayload.py
@@ -238,17 +238,25 @@ class RPMOSTreePayload(ArchivePayload):
                  (iutil.getTargetPhysicalRoot(), iutil.getSysroot() + "/sysroot"),
                  (iutil.getTargetPhysicalRoot() + "/boot", iutil.getSysroot() + "/boot")]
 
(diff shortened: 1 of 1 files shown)


cgwalters:
`updates.img` for this:

```
curl -L -O 'https://fedorapeople.org/~walters/updates.img{,.asc}'
gpg2 --verify updates.img.asc
```

jkonecny12:
Jenkins, it's ok to test.

rvykydal:
Looks good to me.

dustymabe:
can we get this merged? - we'd like to get things fixed so our next two week release of Atomic Host has a fix for this.

dustymabe:
can we get this merged in for f25 as well?

PR #907, opened 2016-12-15
ibft: fix dracut args and ifcfg values for ksdevice=bootif (non-ibft)…

rvykydal:
… (#1396949)

Resolves: rhbz#1396949

The use case is non-ibft device activated (eg pxe device with ksdevice=bootif)
and ibft devices on the same subnet. Pxe device is taking default route, ibft devices
are using the pxe device (default route) to reach iSCSI target and as a
consequence wrong dracut boot options (ip=<pxe device>) and ifcfg values
(NM_CONTROLLED=no is set not for ibft devices but for the device with route to
iscsi target, which is the pxe device) are written out.

### DIFF
--- a/network.py
+++ b/network.py
@@ -291,8 +291,12 @@ class NetworkDevice(IfcfgFile):
             if (isinstance(d, storage.devices.iScsiDiskDevice) and
                 rootdev.dependsOn(d)):
                 if d.nic == "default" or ":" in d.nic:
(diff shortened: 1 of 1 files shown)


vpodzime:
The amount of changed lines here scares me a bit, but I guess there's no better way now. Maybe the code needs some refactoring in the future?

rvykydal:
Closing. We are not going to update rhel6-branch unless strongly required.

PR #908, opened 2016-12-15
Add inst.ks.strict option to show kickstart warnings as errors.

poncovka:
With inst.ks.strict option, the warnings from reading the kickstart
file will be treated as errors. They will be printed on the output
and the installation will terminate immediately. By default, the
warnings are printed to logs and the installation continues.


pyanaconda/anaconda_argparse.py
```diff
@@ -412,6 +412,8 @@ def __call__(self, parser, namespace, values, option_string=None):
                     help=help_parser.help_text("ks"))
     ap.add_argument("--kickstart", dest="ksfile", metavar="KICKSTART_PATH",
                     help=help_parser.help_text("kickstart"))
+    ap.add_argument("--ks.strict", dest="ksstrict", action="store_true",
```
vpodzime: Extra ``.``?

pyanaconda/kickstart.py
```diff
@@ -2162,12 +2164,39 @@ def parseKickstart(f):
...
+        iutil.ipmi_report(IPMI_ABORTED)
+        time.sleep(10)
+        sys.exit(1)
+
+    except (UserWarning, SyntaxWarning, DeprecationWarning) as w:
+        log.error(w)
+        msg = _("\nA warning occurred during reading the kickstart file:"
+                "\n%s\n\nPlease modify your kickstart file or remove the "
+                "`strict` option.\nThe installer will now terminate.") % str(w).strip()
+        print(msg)
```
vpodzime: If I understand this correctly, it will only report the first error in the kickstart file. Or am I wrong? Is there a way to report all errors at once? Running the installation again just to find the next errors sounds like a lot of pain.

pyanaconda/anaconda_argparse.py
jkonecny12: If you want 2 words here use `-` instead of `.`, thought.

poncovka: Fixed to --ksstrict.

pyanaconda/kickstart.py
poncovka: Fixed. All errors are reported at once now.

PR #909, opened 2016-12-15
Fix ppc image being too big againagainagainagainagain (#1405089)

snbueno:


snbueno:
Things are dire, so I've got to break all of our rules and merge this in.

PR #911, opened 2016-12-19
Iutil PEP8 & formatting fixes

M4rtinK:
Make iutil more PEP8 compliant and fix formatting
(mostly indentation) in multiple places.

### DIFF
--- a/pyanaconda/iutil.py
+++ b/pyanaconda/iutil.py
@@ -25,7 +25,7 @@ import os.path
 import subprocess
 import unicodedata
 # Used for ascii_lowercase, ascii_uppercase constants
-import string # pylint: disable=deprecated-module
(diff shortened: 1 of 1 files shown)


pyanaconda/iutil.py
```diff
@@ -990,7 +991,7 @@ def strip_accents(s):
 
     """
     return ''.join((c for c in unicodedata.normalize('NFD', s)
-                      if unicodedata.category(c) != 'Mn'))
+                   if unicodedata.category(c) != 'Mn'))
```
vpodzime: While you're changing this I think on pair of brackets can be removed here.

M4rtinK: Fixed - it's interesting how the code sometimes almost organically mutates to accumulate stuff like this. :)

PR #912, opened 2016-12-20
Fix the name of StorageDiscoveryConfig attribute (#1395350)

poncovka:
Attribute protectedDevSpecs from StorageDiscoveryConfig was changed
to protected_dev_specs.

Resolves: rhbz#1395350


vpodzime:
The test failure seems to be unrelated.

jkonecny12:
When I'm looking on how much PR's still targeting F25, I think I should fix this error.

poncovka:
Merged to master.

PR #914, opened 2016-12-20
Fix selection of no software environment (#1400045)

poncovka:
In tui, do not apply changes when no environment is selected.

Resolves: rhbz#1400045


poncovka:
Merged to master.

PR #915, opened 2016-12-20
Use signals for Spoke & Hub entry/exit callbacks

M4rtinK:
Replace the custom callback calling implementation for the Spoke/Hub
entry/exit events by using the Signal class.

Like this we can remove quite a few lines of code while providing
improved functionality (entry/exit callbacks can now be added & removed at
runtime if needed, etc.).


vpodzime:
Nice!

PR #919, opened 2017-01-03
Add initialization-done signal

M4rtinK:
This pull request has been motivated by the current method of checking that all spokes have been initialized - joining all running threads - not being very robust.

The new method uses an initialization controller that track spoke initialization and triggers a signal (called init_done) once all spokes have been initialized on a given hub.

Due to the architecture of our GUI and TUI interface there are some parts of the PR I'm not really sure about, such as:
- accessing the controller from spokes
  - the controller is per-hub
  - but in our architecture spokes don't directly know on which hub they are running
  - to work around this spoke category matching is used to first find the hub on which the spoke is running and then then its controller

- initialize_start() needs to be always called when spoke initialization starts
  - this might look superfluous as it hub might as well just call that when running the initialize method
  - but that way any third party addon adding a spoke unaware of the initialization controller would prevent the init_done signal from being triggered by not calling initialize_done()
  - this is of course a trade-off - "unaware" addons won't block the init_done signal but we might consider initialization as done even if an "unaware" addon is still being initialized

- initialization in TUI
  - some spokes in TUI don't override the initialize() method
  - initialization (if any) seems to be done in __init__() in sich cases
  - initialize_start()/initialize_done() might look a little weird in such a case

### DIFF
--- a/pyanaconda/lifecycle.py
+++ b/pyanaconda/lifecycle.py
@@ -1,0 +1,162 @@
+# controller.py
+# Anaconda module lifecycle controller.
+#
+# Copyright (C) 2016  Red Hat, Inc.
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# the GNU General Public License v.2, or (at your option) any later version.
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY expressed or implied, including the implied warranties of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General
+# Public License for more details.  You should have received a copy of the
+# GNU General Public License along with this program; if not, write to the
+# Free Software Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
+# 02110-1301, USA.  Any Red Hat trademarks that are incorporated in the
+# source code or documentation are not subject to the GNU General Public
+# License and may only be used or replicated with the express permission of
+# Red Hat, Inc.
+#
+
+from threading import RLock
+from pyanaconda.isignal import Signal
+from pyanaconda.iutil import synchronized
+import logging
+log = logging.getLogger("anaconda")
+
+_controllers = {}
+_controller_categories_map = {}
+
+def get_controller_by_category(category_name):
+    """Return the controller instance that "owns" the corresponding category.
+
+    This is more or less a workaround to the fact that spokes
+    need to know what's their controller, but the controllers
+    are indexed by hub name and spokes don't have a direct reference
+    to the hub they are displayed on.
+
+    Spokes know their category and hubs know what categories the and
+    categories  should not be on two hubs. So we match the spoke category
+    to get the controller corresponding to the spoke.
+
+    :param str category_name: a spoke category name
+    :returns: a Controller instance "owning" the category or None if no matching instance is found
+    :rtype: str or None
+
+    """
+    for controller_name, controller_categories in _controller_categories_map.items():
+        if category_name in controller_categories:
+            return _controllers[controller_name]
+    # no controller has been found for the given spoke based on it's category
+    return None
+
+def get_controller_by_name(controller_name):
+    """Return controller instance by name.
+
+    This function wraps the internal controller dictionary, so that it is not directly exposed.
+
+    The controller names currently correspond to hub class name, eq.:
+    SummaryHub, ProgressHub, etc.
+
+    :param str controller_name: a name of a controller name
+    """
+    return _controllers.get(controller_name)
+
+def add_controller(controller_name, controller_categories):
+    # The controller name is currently based on Hub name
+    # and None indicates that the Hub subclass has not
+    # set the hub_name property to a non-default value.
+    if controller_name is None:
+        log.warning("Controller name is None.")
+    else:
+        log.info("Adding controller: %s", controller_name)
+    controller = Controller()
+    _controllers[controller_name] = controller
+    _controller_categories_map[controller_name] = controller_categories
+    return controller
+
+class Controller(object):
+    """A singleton that track initialization of Anaconda modules."""
+    def __init__(self):
+        self._lock = RLock()
+        self._modules = set()
+        self._all_modules_added = False
+        self.init_done = Signal()
+        self._init_done_triggered = False
+        self._added_module_count = 0
+
+    @synchronized
+    def module_init_start(self, module):
+        """Tell the controller that a module has started initialization.
+
+        :param module: a module which has started initialization
+        """
+        if self._all_modules_added:
+            log.warning("Late module_init_start() from: %s", self)
+        elif module in self._modules:
+            log.warning("Module already marked as initializing: %s", module)
+        else:
+            self._added_module_count += 1
+            self._modules.add(module)
+
+    def all_modules_added(self):
+        """Tell the controller that all expected modules have started initialization.
+
+        Tell the controller that all expected modules have been registered
+        for initialization tracking (or have already been initialized)
+        and no more are expected to be added.
+
+        This is needed so that we don't prematurely trigger the init_done signal
+        when all known modules finish initialization while other modules have not
+        yet been added.
+        """
+        init_done = False
+        with self._lock:
+            log.info("Initialization of all modules (%d) has been started.", self._added_module_count)
+            self._all_modules_added = True
+
+            # if all modules finished initialization before this was added then
+            # trigger the init_done signal at once
+            if not self._modules and not self._init_done_triggered:
+                self._init_done_triggered = True
+                init_done = True
+
+        # we should emit the signal out of the main lock as it doesn't make sense
+        # to hold the controller-state lock once we decide to the trigger init_done signal
+        # (and any callbacks registered on it)
+        if init_done:
+            self._trigger_init_done()
+
+    def module_init_done(self, module):
+        """Tell the controller that a module has finished initialization.
+
+        And if no more modules are being initialized trigger the init_done signal.
+
+        :param module: a module that has finished initialization
+        """
+        init_done = False
+        with self._lock:
+            # prevent the init_done signal from
+            # being triggered more than once
+            if self._init_done_triggered:
+                log.warning("Late module_init_done from module %s.", module)
+            else:
+                if module in self._modules:
+                    log.info("Module initialized: %s", module)
+                    self._modules.discard(module)
+                else:
+                    log.warning("Unknown module reported as initialized: %s", module)
+                # don't trigger the signal if all modules have not yet been added
+                if self._all_modules_added and not self._modules:
+                    init_done = True
+                    self._init_done_triggered = True
+
+        # we should emit the signal out of the main lock as it doesn't make sense
+        # to hold the controller-state lock once we decide to the trigger init_done signal
+        # (and any callbacks registered on it)
+        if init_done:
+            self._trigger_init_done()
+
+    def _trigger_init_done(self):
+        log.info("All modules have been initialized.")
+        self.init_done.emit()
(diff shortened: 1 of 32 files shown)


jkonecny12:
Please rebase this PR and fix conflicts.

pyanaconda/ui/tui/spokes/software.py
```diff
@@ -90,6 +91,9 @@ def _initialize(self):
         # Apply the initial selection
         self._apply()
 
+        # report that the software spoke has been initialized
+        self.initialize_done()
```
poncovka: Could be a problem, that there is a new thread created in `self._apply()` for the software selection check?

pyanaconda/ui/common.py
```diff
@@ -391,6 +391,49 @@ def finished(self):
...
+    #
+    # - initialize_start()
+    # -> Must be called by every spoke that calls initialize_start() or else the init_done
+    #    signal will never be emitted.
+
+    @property
+    def initialization_controller(self):
+        # standalone spokes don't have a category
+        if self.category:
+            return controller.get_controller(category_name=self.category.__name__)
```
poncovka: The call of `controller.get_controller()` is a little confusing. Same for `controller.controllers.get()`. Btw, I cannot find an import command for `controller` :-)

pyanaconda/ui/gui/hubs/__init__.py
```diff
@@ -211,6 +211,13 @@ def _createBox(self):
             if len(selectors) % 2:
                 row += 1
 
+        # initialization of all expected spokes has been started, so notify the controller
+        hub_controller = controller.controllers.get(self.__class__.__name__)
```
poncovka: Again, where is the `controller` imported?

pyanaconda/controller.py
```diff
@@ -0,0 +1,124 @@
...
+
+        # we should be able to safely emit the signal out of the main lock
+        if init_done:
+            self._trigger_init_done()
+
+    def _trigger_init_done(self):
+        log.info("All modules have been initialized.")
+        self.init_done.emit()
+
+def get_controller(category_name):
```
poncovka: Since I don't like to see `controller.controllers.get()` and `controller.get_controller()`, I suggest to add the methods and the variables to the `Controller` class as class methods and variables.

pyanaconda/controller.py
```diff
@@ -0,0 +1,124 @@
...
+
+        # if all modules finished initialization before this was added then
+        # trigger the init_done signal at once
+        if not self._modules and not self._init_done_triggered:
+            self._trigger_init_done()
+
+    def module_init_done(self, module):
+        """Tell the controller that a module has finished initialization.
+
+        An if no more modules are being initialized trigger the init_done signal.
```
M4rtinK: Fixed.

pyanaconda/controller.py
```diff
@@ -0,0 +1,124 @@
...
+                return
+            if module in self._modules:
+                log.info("Module initialized: %s", module)
+                self._modules.discard(module)
+            else:
+                log.warning("Unknown module reported as initialized: %s", module)
+            # don't trigger the signal if all modules have not yet been added
+            if self._all_modules_added and not self._modules:
+                init_done = True
+                self._init_done_triggered = True
```
M4rtinK: It should be done inside of the self._lock mutex as I *think* there could be an issue if the signal was triggered while the lock is held. 

That's why the `_trigger_init_done()` method, which does the signal triggering, is called outside of the lock. 

But in any case, the code could be written better & better documented so I'll try to improve it.

Also `all_modules_added()` might call `_trigger_init_done()` as well, so I'll make sure to also set `self._init_done_triggered` to True there.

pyanaconda/controller.py
M4rtinK: I don't know - that would basically mean storing mutable data in a class variable and that seems kinda weird. Also if I understand this correctly it would be called like this:

`controller.Controller.get_controller("foo")`

Which IMHO doesn't look any better than the current way.

Maybe we can rename the module to something else (`controllers`, etc.) or add a separate module (`spoke_startup`, `module_startup`, `module_init`, etc.) that has the dicts of  the `Controller` instances & the access functions while the `Controller` class still lives in the `controller` module.

Then the call could look like this:

controllers.get_controller("foo")
or
spoke_startup.get_controller("foo")
or
module_startup.get_controller("foo")
or 
module_init.get_controller("foo")

As for `controller.controllers.get()` I guess we could change the variable to something like `_controllers` (or even `__controllers`) if we want people to only use it via the access functions. and not directly.

pyanaconda/ui/common.py
```diff
@@ -391,6 +391,49 @@ def finished(self):
...
+    #    finished initialization.
+    # -> If there is no Hub (standalone spoke) the property is None
+    #
+    # - initialize_start()
+    # -> Should be called when Spoke initialization is started.
+    # -> Needs to be called explicitly, if we called it for every spoke by default
+    #    then any spoke that does not call initialize_done() would prevent the
+    #    controller form ever triggering the init_done signal.
+    #
+    # - initialize_start()
```
M4rtinK: Fixed, thanks! :)

pyanaconda/ui/common.py
M4rtinK: Oops, indeed! But it *worked* during my local testing - I kinda wonder how. :D

pyanaconda/ui/tui/spokes/software.py
M4rtinK: Good point - I'll add a `threadMgr.wait()` on the thread started by _apply() to make sure the software selection spoke is really initialized before initialize_done() is called.

The `_initialize()` method is already running in a thread so it should be OK. :)

pyanaconda/ui/gui/hubs/__init__.py
M4rtinK: Should be also fixed now.

pyanaconda/ui/common.py
```diff
@@ -365,8 +366,50 @@ def finished(self):
...
+
+    def initialize_done(self):
+        # get the correct controller for this spoke
+        spoke_controller = self.initialization_controller
+        # check if there actually is a controller for this spoke, there might not be one
+        # if this is a standalone spoke
+        if spoke_controller:
+            spoke_controller.module_init_done(self)
+
+     def __repr__(self):
```
jkonecny12: Bad indentation. This is breaking build and blocking tests.

pyanaconda/controller.py
```diff
@@ -0,0 +1,134 @@
...
+        # to hold the controller-state lock once we decide to the trigger init_done signal
+        # (and any callbacks registered on it)
+        if init_done:
+            self._trigger_init_done()
+
+    def _trigger_init_done(self):
+        log.info("All modules have been initialized.")
+        self.init_done.emit()
+
+def get_controller(category_name):
```
jkonecny12: Please move these top level function to top of the file.

pyanaconda/ui/common.py
```diff
@@ -640,4 +684,13 @@ def collectCategoriesAndSpokes(paths, klass, displaymode):
...
 
+    # As we now have a list of all categories this hub holds we can now register it's controller.
+    # We need the list of categories so that spokes can find out which controller they should use
+    # based on their category.
+    category_names = set()
+    for c in categories:
+        category_names.add(c.__name__)
+
+    controller.add_controller(klass.__name__, category_names)
+
```
jkonecny12: I think you can remove this blank line.

pyanaconda/ui/gui/spokes/keyboard.py
```diff
@@ -384,6 +385,9 @@ def _wait_ready(self):
         self._ready = True
         hubQ.send_ready(self.__class__.__name__, False)
 
+        # report that the keyboard spoke initialization has been finished
```
jkonecny12: `finished` is not the right word used here, thought.

pyanaconda/ui/gui/spokes/software.py
```diff
@@ -342,6 +343,9 @@ def _initialize(self):
         if not self._error:
             self._apply()
 
+        # report that software spoke initialization has been finished
```
jkonecny12: The same here. I think the word `finish` should be changed.

tests/pyanaconda_tests/controller_test.py
```diff
@@ -0,0 +1,129 @@
...
+
+from pyanaconda import controller
+
+class TestModule(object):
+    def __init__(self):
+        self._test = 1
+
+
+class InstallTasksTestCase(unittest.TestCase):
+
```
jkonecny12: You can remove the blank line here.

tests/pyanaconda_tests/controller_test.py
```diff
@@ -0,0 +1,129 @@
...
+    def setUp(self):
+        self._test_variable1 = 0
+
+    def _increment_var1(self):
+        self._test_variable1 += 1
+
+    def controller_test(self):
+        """Check that the module initialization controller works correctly."""
+
```
jkonecny12: Remove the blank line, please.

tests/pyanaconda_tests/controller_test.py
```diff
@@ -0,0 +1,129 @@
...
+        # When the last of the expected modules reports
+        # initialization as done the init_done signal of the
+        # controller should be triggered. So check that
+        # it really happened.
+        self.assertEqual(self._test_variable1, 1)
+
+    def controller_robustness_test(self):
+        """Check of controller handles various edge cases."""
+
```
jkonecny12: Remove the blank line, please.

pyanaconda/ui/common.py
M4rtinK: Looks like a rebase error, should be fixed now.

pyanaconda/ui/common.py
M4rtinK: There is a return statement under it:
```
# As we now have a list of all categories this hub holds we can now register it's controller.
# We need the list of categories so that spokes can find out which controller they should use
# based on their category.
category_names = set()
for c in categories:
    category_names.add(c.__name__)
    
controller.add_controller(klass.__name__, category_names)                                                                                                              
    
return ret
```
I think it's reasonable to have blank line above & bellow this statement as it doesn't really belong to either the for loop above of the return statement bellow. But I guess I can add a comment. :)

pyanaconda/controller.py
M4rtinK: Sure, why not. :)

tests/pyanaconda_tests/controller_test.py
M4rtinK: The one bellow, right ? Why not. :)

pyanaconda/ui/gui/spokes/keyboard.py
M4rtinK: Switched to "completed" - if I understand the issue correctly.

pyanaconda/controller.py
poncovka: I don't want to be a nitpicker here, so if others don't mind, it does not have to be changed.

pyanaconda/ui/gui/spokes/keyboard.py
jkonecny12: The `completed` sounds more appropriate here. With the finished it felt to me like the spoke has completed it's work completely not just initialized itself.

Thank you for change.

M4rtinK:
Yeah - lifecycle sounds good & we can keep the class & the dictionary of controller instances in a single file. :)

pyanaconda/lifecycle.py
```diff
@@ -0,0 +1,162 @@
...
+
+def get_controller_by_category(category_name):
+    """Return the controller instance that "owns" the corresponding category.
+
+    This is more or less a workaround to the fact that spokes
+    need to know what's their controller, but the controllers
+    are indexed by hub name and spokes don't have a direct reference
+    to the hub they are displayed on.
+
+    Sokes know their category and hubs know what categories the and
```
vpodzime: Spokes...

M4rtinK: Fixed! :)

PR #925, opened 2017-01-05
Master fix tests

jkonecny12:
Fix tests for master and ignore result directory with logs from tests in git.

### DIFF
--- a/scripts/makebumpver
+++ b/scripts/makebumpver
@@ -70,6 +70,9 @@ class MakeBumpVer:
             config_ac = f.read()
 
         # get current package name, version & bug reporting email from configure.ac in case they are needed
+        #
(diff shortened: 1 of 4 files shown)


scripts/makebumpver
```diff
@@ -70,6 +70,9 @@ class MakeBumpVer:
             config_ac = f.read()
 
         # get current package name, version & bug reporting email from configure.ac in case they are needed
+        #
+        # It looks like false positive raised on python 3.6
+        # pylint: disable=no-member
```
vpodzime: I wish pylint had some syntax for a version-specific disable.

jkonecny12: That cloud be handy. 👍

PR #928, opened 2017-01-11
There is no thread for dasd formatting in tui.

poncovka:
In tui, do not check the existence of a thread for dasd formatting,
because it is never created.


poncovka:
Fixed.

PR #935, opened 2017-01-23
The base group should be selected by default (#1361121)

poncovka:
Resolves: rhbz#1361121

### DIFF
--- a/pyanaconda/packaging/yumpayload.py
+++ b/pyanaconda/packaging/yumpayload.py
@@ -1296,6 +1296,9 @@ reposdir=%s
         else:
             self._selectYumGroup("core")
 
+        # Select the base group by default. It might be excluded later.
+        self._selectYumGroup("base")
+
         env = None
 
         if self.data.packages.default and self.environments:


pyanaconda/packaging/yumpayload.py
```diff
@@ -1296,6 +1296,9 @@ def _applyYumSelections(self):
         else:
             self._selectYumGroup("core")
 
+        # Select the base group by default. It might be excluded later.
+        self._selectYumGroup("base")
```
vpodzime: How does this work with the ``--nocore`` option? We don't want minimal (container) images to contain everything from the base group, do we?

Frodox: @vpodzime `--nocore` disables `core` group. But it is `base` group ;D

poncovka: According to the [documentation](https://access.redhat.com/documentation/en-US/Red_Hat_Enterprise_Linux/7/html/Installation_Guide/sect-kickstart-syntax.html#sect-kickstart-packages), this is the correct behaviour. Due to the removed `--nobase` option, the minimal installation should look like this:
```
%packages --nocore
-@base
%end
```

I can deselect the base group for `--nocore` option, but then we should change the documentation.

poncovka:
This fix can break minimal installations that do no exclude the base group, because it was never installed, or that use the deprecated --nobase option, because it looked like it worked. Therefore, the pull request will be closed and the documentation changed.

PR #937, opened 2017-01-26
Do not allow creating ntfs filesystem in custom spoke

vojtechtrefny:
We don't want to allow creating ntfs filesystem from Anaconda even
if it is supported.

### DIFF
--- a/pyanaconda/ui/gui/spokes/custom.py
+++ b/pyanaconda/ui/gui/spokes/custom.py
@@ -314,8 +314,7 @@ class CustomPartitioningSpoke(NormalSpoke, StorageChecker):
             obj = cls()
 
             # btrfs is always handled by on_device_type_changed
-            supported_fs = (obj.type != "btrfs" and
-                            obj.type != "tmpfs" and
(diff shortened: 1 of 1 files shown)


Frodox:
What is the idea? :)

vojtechtrefny:
The idea is we really don't want to deal with all the problems after someone creates `/home` or even `/` on NTFS.

We originally didn't allow to create NTFS in Blivet, but we changed that -- https://github.com/rhinstaller/blivet/pull/536 -- so Anaconda has to "remove" support fo creating NTFS.

pyanaconda/ui/gui/spokes/custom.py
```diff
@@ -314,8 +314,7 @@ def _initialize(self):
             obj = cls()
 
             # btrfs is always handled by on_device_type_changed
-            supported_fs = (obj.type != "btrfs" and
-                            obj.type != "tmpfs" and
+            supported_fs = (obj.type not in ("btrfs", "tmpfs", "ntfs") and
```
vpodzime: Just a late note on this -- might be worth to have the list stored somewhere else with an appropriate name like ``UNSUPPORTED_FILESYSTEMS``.

PR #938, opened 2017-01-26
Rhel7 branch rhbz#1416655 bridge ks pre activate onboot

rvykydal:


pyanaconda/network.py
```diff
@@ -636,10 +636,10 @@ def _get_s390_settings(devname):
 
     return cfg
 
-def _add_slave_connection(slave_type, slave, master, activate, values=None):
+def _add_slave_connection(slave_type, slave_idx, slave, master, activate, values=None):
     values = values or []
-    #slave_name = "%s slave %d" % (devname, slave_idx)
-    slave_name = slave
+    slave_name = "%s slave %d" % (master, slave_idx)
+    #slave_name = slave
```
jkonecny12: Please remove this if there is no higher meaning which I don't see.

pyanaconda/network.py
```diff
@@ -651,17 +651,15 @@ def _add_slave_connection(slave_type, slave, master, activate, values=None):
     mac = nm.nm_device_perm_hwaddress(slave)
     mac = [int(b, 16) for b in mac.split(":")]
     values.append(['802-3-ethernet', 'mac-address', mac, 'ay'])
+    # HACK preventing NM to autoactivate the connection
```
jkonecny12: If I understand it correctly, please correct me if not, the `ifconf` file will be changed later to have `ONBOOT=yes`. I think it would be nice to write it here.

rvykydal:
Ported to master with https://github.com/rhinstaller/anaconda/pull/940

PR #939, opened 2017-01-26
network: respect --activate value for bridge from kickstart (#1416687)

rvykydal:
Resolves: rhbz#1416687

We need to set correct ONBOOT value (which is "no" if we don't want to activate
the device in stage 2) for NM also for the slave.

Depends on
https://github.com/rhinstaller/anaconda/pull/938/commits/4458d7c7a2171a8b97aa895c8a3f332179a76554


rvykydal:
Ported to master with https://github.com/rhinstaller/anaconda/pull/940

PR #941, opened 2017-01-31
Fixed the default behaviour of the autopartitioning (#1380767)

poncovka:
Always set the default partitioning
* The default partitioning of the install class should be always set.

Enable to define the autopart type in an install class
* The default partitioning scheme can be defined in an install class.
  If the kickstart file defines the scheme, it will be used instead.
    
Resolves: rhbz#1380767

### DIFF
--- a/pyanaconda/installclass.py
+++ b/pyanaconda/installclass.py
@@ -115,6 +115,9 @@ class BaseInstallClass(object):
         anaconda.bootloader.timeout = self.bootloaderTimeoutDefault
         anaconda.bootloader.boot_args.update(self.bootloaderExtraArgs)
 
+        # The default partitioning should be always set.
+        self.setDefaultPartitioning(anaconda.storage)
+
     # sets default ONBOOT values and updates ksdata accordingly
     def setNetworkOnbootDefault(self, ksdata):
         pass
--- a/pyanaconda/installclass.py
+++ b/pyanaconda/installclass.py
@@ -34,6 +34,7 @@ import logging
 log = logging.getLogger("anaconda")
 
 from pyanaconda.kickstart import getAvailableDiskSpace
+from pyanaconda.constants import DEFAULT_AUTOPART_TYPE
 
 class BaseInstallClass(object):
     # default to not being hidden
@@ -57,6 +58,10 @@ class BaseInstallClass(object):
     # The default filesystem type to use.  If None, we will use whatever
     # Blivet uses by default.
     defaultFS = None
+
+    # The default partitioning scheme to use for autopartitioning.
+    # It has to be always set.
+    default_autopart_type = DEFAULT_AUTOPART_TYPE
 
     # help
     help_folder = "/usr/share/anaconda/help"
--- a/pyanaconda/iutil.py
+++ b/pyanaconda/iutil.py
@@ -1242,3 +1242,7 @@ def touch(file_path):
     # even when the path points to dirrectory
     if not os.path.exists(file_path):
         os.mknod(file_path)
+
+def firstNotNone(iterable):
+    """Return the first item that is not None."""
+    return next((item for item in iterable if item is not None), None)
--- a/pyanaconda/kickstart.py
+++ b/pyanaconda/kickstart.py
@@ -290,7 +290,7 @@ class AutoPart(commands.autopart.RHEL7_AutoPart):
 
         # sets up default autopartitioning.  use clearpart separately
         # if you want it
-        instClass.setDefaultPartitioning(storage)
+        refreshAutoSwapSize(storage)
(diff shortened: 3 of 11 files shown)


Frodox:
Great +1 for this!

pyanaconda/kickstart.py
```diff
@@ -290,7 +290,7 @@ def execute(self, storage, ksdata, instClass):
 
         # sets up default autopartitioning.  use clearpart separately
         # if you want it
-        instClass.setDefaultPartitioning(storage)
+        refreshAutoSwapSize(storage)
```
vpodzime: Please also describe this particular change in the commit message.

pyanaconda/installclass.py
```diff
@@ -58,6 +58,10 @@ class BaseInstallClass(object):
     # Blivet uses by default.
     defaultFS = None
 
+    # The default partitioning scheme to use for autopartitioning.
+    # If None, we will use DEFAULT_AUTOPART_TYPE.
```
vpodzime: Why is it being set to ``None`` then?

pyanaconda/iutil.py
```diff
@@ -1242,3 +1242,10 @@ def touch(file_path):
     # even when the path points to dirrectory
     if not os.path.exists(file_path):
         os.mknod(file_path)
+
+def firstNotNone(*args):
+    """Return the first argument that is not None."""
+    for arg in args:
+        if arg is not None:
+            return arg
```
vpodzime: This whole function can be written as:
```
return next((arg for arg in args if arg is not None), None)
```

vpodzime: And if it is a function, it should be generic -- i.e. take any iterable as a single argument and return the first item that is not None. AFAICT there's no need to use ``*args`` machinery here.

pyanaconda/ui/tui/spokes/storage.py
```diff
@@ -42,15 +42,15 @@
...
+from pykickstart.constants import CLEARPART_TYPE_ALL, CLEARPART_TYPE_LINUX, CLEARPART_TYPE_NONE
 from pykickstart.errors import KickstartValueError
 
 from collections import OrderedDict
 
 import logging
 log = logging.getLogger("anaconda")
 
-__all__ = ["StorageSpoke", "AutoPartSpoke"]
+__all__ = ["StorageSpoke"]
```
vpodzime: Is this change intentional? What is the purpose?

pyanaconda/installclass.py
poncovka: I could change `None` to `DEFAULT_AUTOPART_TYPE`, but then it is not necessary to have the constant `DEFAULT_AUTOPART_TYPE` at all, because it will be used only here.

pyanaconda/iutil.py
poncovka: I've seen this implementation before, but it is really hard to read. What are the advantages?

poncovka: I can make the function generic, though,

pyanaconda/ui/tui/spokes/storage.py
poncovka: In tui, it does not make sense to enable automatic import of indirect spokes, because we always create them and use them only in other spokes. With automatic import, they will be found by a hub, created, initialized and forgotten.

Also, I've added a new parameter to the `AutoPartSpoke.__init__` method and the initialization in a hub does not work anymore.

pyanaconda/kickstart.py
poncovka: Fixed.

pyanaconda/installclass.py
vpodzime: Still better to have a constant for this -- for other installclasses, addons,...

pyanaconda/iutil.py
vpodzime: There's no advantage, it just looks more "pythonic", whatever that means. :) Feel free to leave the for cycle in place. My main concern here is the (non-)genericity.

pyanaconda/ui/tui/spokes/storage.py
vpodzime: Thanks for the extra explanation. I think this deserves a separate commit or at least some explanation in the commit message, though.

poncovka: Fixed.

pyanaconda/installclass.py
poncovka: Fixed.

pyanaconda/iutil.py
poncovka: Fixed.

PR #945, opened 2017-02-02
Fix Anaconda forces payload restart when network (not)change

jkonecny12:
This problem is solved by downloading content of `repomd.xml` file and comparing its MD5 hashes with downloaded `repomd.xml` file from the old repo url.
    
MD5 hash is used here because it takes less memory and we don't need `repomd.xml` content for anything else.
    
The `postSetup()` method was added to the payload. This method can be overridden by `Payload` child classes which needs to do additional tasks on the end of the `restartThread` call from the `PayloadManager` class.
    
*Resolves: rhbz#1373449*
*Resolves: rhbz#1358778*

Second commit is to make docstrings consistent in `Payload` and `YumPayload`.

I want to backport this to Rawhide later.

### DIFF
--- a/pyanaconda/packaging/__init__.py
+++ b/pyanaconda/packaging/__init__.py
@@ -119,10 +119,9 @@ class PayloadInstallError(PayloadError):
     pass
 
 class Payload(object):
-    """ Payload is an abstract class for OS install delivery methods. """
+    """Payload is an abstract class for OS install delivery methods."""
     def __init__(self, data):
-        """ data is a kickstart.AnacondaKSHandler class
-        """
+        """Data is a kickstart.AnacondaKSHandler class."""
         if self.__class__ is Payload:
             raise TypeError("Payload is an abstract class")
 
@@ -135,35 +134,35 @@ class Payload(object):
         self.txID = None
 
     def setup(self, storage, instClass):
-        """ Do any payload-specific setup. """
+        """Do any payload-specific setup."""
         self.storage = storage
         self.instclass = instClass
 
     def unsetup(self):
-        """ Invalidate a previously setup paylaod. """
+        """Invalidate a previously setup payload."""
         self.storage = None
         self.instclass = None
 
     def preStorage(self):
-        """ Do any payload-specific work necessary before writing the storage
-            configuration.  This method need not be provided by all payloads.
+        """Do any payload-specific work necessary before writing the storage
+        configuration.  This method need not be provided by all payloads.
         """
         pass
 
     def release(self):
-        """ Release any resources in use by this object, but do not do final
-            cleanup.  This is useful for dealing with payload backends that do
-            not get along well with multithreaded programs.
+        """Release any resources in use by this object, but do not do final
+        cleanup.  This is useful for dealing with payload backends that do
+        not get along well with multithreaded programs.
         """
         pass
 
     def reset(self, root=None, releasever=None):
-        """ Reset the instance, not including ksdata. """
+        """Reset the instance, not including ksdata."""
         pass
 
     def prepareMountTargets(self, storage):
         """Run when physical storage is mounted, but other mount points may
-        not exist.  Used by the RPMOSTreePayload subclass. 
+        not exist.  Used by the RPMOSTreePayload subclass.
         """
         pass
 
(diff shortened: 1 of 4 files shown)


pyanaconda/packaging/__init__.py
```diff
@@ -307,6 +306,19 @@ def disableRepo(self, repo_id):
...
+        """Verify availability of existing repositories.
+
+        This method will test if repositories can be reached by URL from
+        actual repositories. It is useful when network settings is changed
+        so that we can verify if repositories are still reachable.
+
+        This method should be overriden.
+        """
+        log.debug("Install method %s is not able to verify availability",
+                  self.data.method.method)
```
vpodzime: I think this should log the name of the class instead.

pyanaconda/packaging/__init__.py
```diff
@@ -1382,11 +1402,16 @@ def _runThread(self, storage, ksdata, payload, instClass, fallback, checkmount):
...
+        # run payload specific post configuration tasks
+        payload.postSetup()
+
         self._setState(self.STATE_FINISHED)
 
 # Initialize the PayloadManager instance
 payloadMgr = PayloadManager()
 
+
+
```
vpodzime: Why three blank lines here?

pyanaconda/packaging/yumpayload.py
```diff
@@ -1611,3 +1637,69 @@ def postInstall(self):
...
+
+    def verifyRepoMD(self):
+        """Download and compare with stored repomd.xml file."""
+        new_repomd = self._downloadRepoMD(self._method)
+        new_repomd_hash = self._calculateHash(new_repomd)
+        return new_repomd_hash == self._repomd_hash
+
+    def _calculateHash(self, message):
+        m = hashlib.md5()
+        m.update(message)
```
vpodzime: Why ``message``?

pyanaconda/packaging/__init__.py
```diff
@@ -119,10 +119,9 @@ class PayloadInstallError(PayloadError):
     pass
 
 class Payload(object):
-    """ Payload is an abstract class for OS install delivery methods. """
+    """Payload is an abstract class for OS install delivery methods."""
```
vpodzime: One comment for this whole commit: tl;dr :)

tests/pyanaconda_tests/payload_test.py
```diff
@@ -0,0 +1,96 @@
...
+        """Test verification method."""
+        r = RepoMDMetaHash(DummyPayload(), self._dummyRepo)
+        r.storeRepoMDHash()
+
+        # test if repomd comparision works properly
+        self.assertTrue(r.verifyRepoMD())
+
+        # test if repomd change will be detected
+        with open(self._md_file, 'a') as f:
+            f.seek(os.SEEK_END)
```
vpodzime: Is this really necessary? I'd expect the ``'a'`` mode to take care of this.

pyanaconda/packaging/yumpayload.py
jkonecny12: First I used `input` here but then pylint complain about redefining `input`... so I changed it to a neutral `message`. Do you have better name for this?

pyanaconda/packaging/__init__.py
jkonecny12: I totally understand it :)

pyanaconda/packaging/__init__.py
jkonecny12: Mistake, sorry about that.

pyanaconda/packaging/__init__.py
jkonecny12: I guess it gives more sense. I'll fix that.

tests/pyanaconda_tests/payload_test.py
jkonecny12: Based on the test, it looks like I can remove the `f.seek()` method from here. Thanks for pointing me on this.

jkonecny12:
I just wonder how hard it will be to port this to master/DNF.

pyanaconda/packaging/yumpayload.py
vpodzime: ``data``?

jkonecny12: yes, why not :)

PR #947, opened 2017-02-03
Make lvdump be more verbose in pre-gen-log script (#1255659)

jkonecny12:
Add parameters '-l -u -s' to get more output from lvmdump.

This patch depends on this change in Lorax https://github.com/rhinstaller/lorax/pull/181 .

*Related: rhbz#1255659*


vpodzime:
I believe this deserves some more explanation in the commit message.

jkonecny12:
Is it better now?

PR #870, opened 2016-11-15
Password checking improvements

M4rtinK:
This patch set cleans up and generally improves password quality checking in Anaconda. While password quality checking sounds simple, it is actually quite complicated due to many states a password can be in, different password entry screens with different expectations and also due to interaction with the password quality policy.

The actual core password checking code has been refactored, as well as the individual password check that have been moved the the GUI helper module and are no longer duplicated in the user and root password spokes. Also the interactive & partial kickstart installations now use the same default policy and there is now also a set of unit tests for the password checking code.

Note: The LUKS passphrase quality checking has not yet been refactored to limit the already significant amount of changes. This will be the target of follow-up PR later on.

PR #952, opened 2017-02-07
Password checking fixes and improvements

M4rtinK:
This is a preliminary version of the password checking improvements - but the PR is rather big & changing quite a lot of stuff so I think it's good to make it available for review sooner rather then later.

In short what this pull request does:
- first commit setts a more sane default password policy
- the second commit refactors the password checking code, decreasing code duplication and hopefully making to code clearer and more easy to maintain
    - it also fixes a ton of small password checking issues if the maze that is the overall policy based password checking code

Looking forward to any feedback - thanks in advance!

Known issues - the following is a list of known issues with the PR. I plan to keep it up to date as issues are found & fixed, eventually ironing them all out so that the PR can be pushed:
- password status is not updated when whole password is selected and deleted - **FIXED**
- password policy does not properly affect LUKS passphrase entry - **FIXED**
- press-done-twice logic shows incorrect messages for empty password - **FIXED**
- the "create a user without password" checkbox is always inactive  & grayed out - **FIXED**
- non-ASCII passwords can't be used even if done is clicked repeatedly - **FIXED**
- password confirmation not applied if only the password and not the confirmation field change - **FIXED**
- [the RHEL7 docs](https://access.redhat.com/documentation/en-US/Red_Hat_Enterprise_Linux/7/html/Installation_Guide/sect-kickstart-syntax.html) will have to be changed to reflect the new saner password policy defaults

### DIFF
--- a/pyanaconda/constants.py
+++ b/pyanaconda/constants.py
@@ -136,22 +136,35 @@ UNSUPPORTED_HW = 1 << 28
 
 # Password validation
 PASSWORD_MIN_LEN = 6
-PASSWORD_EMPTY_ERROR = N_("The password is empty.")
-PASSWORD_CONFIRM_ERROR_GUI = N_("The passwords do not match.")
-PASSWORD_CONFIRM_ERROR_TUI = N_("The passwords you entered were different.  Please try again.")
-PASSWORD_WEAK = N_("The password you have provided is weak. %s")
-PASSWORD_WEAK_WITH_ERROR = N_("The password you have provided is weak: %s.")
-PASSWORD_WEAK_CONFIRM = N_("You have provided a weak password. Press Done again to use anyway.")
-PASSWORD_WEAK_CONFIRM_WITH_ERROR = N_("You have provided a weak password: %s. Press Done again to use anyway.")
-PASSWORD_ASCII = N_("The password you have provided contains non-ASCII characters. You may not be able to switch between keyboard layouts to login. Press Done to continue.")
+PASSWORD_EMPTY_ERROR = N_("The %(password)s is empty.")  # singular
+PASSWORD_CONFIRM_ERROR_GUI = N_("The %(passwords)s do not match.")  # plural
+PASSWORD_CONFIRM_ERROR_TUI = N_("The %(passwords)s you entered were different.  Please try again.")  # plural
+# The password-too-short constant is used to replace a libpwquality error message,
+# which is why it does not end with a ".", like all the other do.
+PASSWORD_TOO_SHORT = N_("The %(password)s is too short")  # singular
+PASSWORD_WEAK = N_("The %(password)s you have provided is weak.")  # singular
+PASSWORD_WEAK_WITH_ERROR = N_("The %(password)s you have provided is weak: %(error_message)s.")  # singular
+PASSWORD_FINAL_CONFIRM = N_("Press Done again to use anyway.")
+PASSWORD_ASCII = N_("The %(password)s you have provided contains non-ASCII characters. You may not be able to switch between keyboard layouts when typing it.")
+# ^ singular
 PASSWORD_DONE_TWICE = N_("You will have to press Done twice to confirm it.")
+PASSWORD_DONE_TO_CONTINUE = N_("Press Done to continue.")
 
+# password status
 PASSWORD_STATUS_EMPTY = N_("Empty")
 PASSWORD_STATUS_TOO_SHORT = N_("Too short")
 PASSWORD_STATUS_WEAK = N_("Weak")
 PASSWORD_STATUS_FAIR = N_("Fair")
 PASSWORD_STATUS_GOOD = N_("Good")
 PASSWORD_STATUS_STRONG = N_("Strong")
+
+# how should passwords be called in combined strings
+NAME_OF_PASSWORD = N_("password")
+NAME_OF_PASSWORD_PLURAL = N_("passwords")
+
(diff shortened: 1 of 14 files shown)


pyanaconda/users.py
```diff
@@ -235,8 +271,173 @@ def guess_username(fullname):
...
+
+    def __init__(self, password,
+                 username='root',
+                 minimum_length=None,
+                 empty_ok=False,
+                 pwquality_settings=None):
+        self._password = password
+        self._username = username
+        if minimum_length is None:
+            minimum_length = PASSWORD_MIN_LEN
```
jkonecny12: Move this few lines above. It will be more readable when this logic won't be inside of other assignments.

pyanaconda/constants.py
```diff
@@ -139,10 +139,12 @@
 PASSWORD_EMPTY_ERROR = N_("The password is empty.")
 PASSWORD_CONFIRM_ERROR_GUI = N_("The passwords do not match.")
 PASSWORD_CONFIRM_ERROR_TUI = N_("The passwords you entered were different.  Please try again.")
+PASSWORD_TOO_SHORT = N_("The password is too short")
```
M4rtinK: This is actually correct - the message replaces error messages from `libpwquality` if the password is too short and `libpwquality` apparently does not end it's errors with a dot (even though they start with an upper case letter).

But this is not apparent at all, so I'll add a comment explaining this to next to the constant. :)

pyanaconda/ui/gui/helpers.py
```diff
@@ -81,6 +115,17 @@ class GUISpokeInputCheckHandler(GUIInputCheckHandler):
 
     __metaclass__ = ABCMeta
 
+    def __init__(self):
```
M4rtinK: Indeed! Taking a second look on the class I now see it's also used in Source spoke for general input checking. So I've created the `GUISpokePasswordCheckHandler` class, which inherits from `GUISpokeInputCheckHandler` and has all the password checking specific functionality. The password and user spokes now use the new class while the source spoke still uses the old one. :)

pyanaconda/users.py
M4rtinK: Yeah, why not. :)

pyanaconda/constants.py
jkonecny12: OK, thanks for explanation.

jkonecny12:
You still have one `nosetest` error.

pyanaconda/users.py
```diff
@@ -112,87 +112,122 @@ def cryptPassword(password, algo=None):
 
     return cryptpw
 
-def validatePassword(pw, user="root", settings=None, minlen=None, empty_ok=False):
+class PwqualitySettingsCache(object):
+    """Cache for libpwquality settings used for password validation.
+
+    Libpwquality settings instantiation is probably not exactly cheep
```
poncovka: Cheap?

poncovka:
There is one more failure in tests. It looks like you should use named arguments instead of `%s` in string constants.

M4rtinK:
@poncovka I've seen it - and I'm currently figuring out how to fix it. :)

pyanaconda/users.py
M4rtinK: This is in the context of computationally expensive/cheap, but I guess I should make that more evident.

poncovka: There is `cheep` instead of `cheap`. The sentence is ok otherwise :-)

M4rtinK: Oh, right! :D

pyanaconda/ui/gui/helpers.py
```diff
@@ -121,3 +315,154 @@ def on_back_clicked(self, window):
...
+        GUISpokeInputCheckHandler.__init__(self)
+        self._password_required = False
+        self._waive_clicks = 0
+        self._waive_ASCII_clicks = 0
+
+    @property
+    def password_required(self):
+        """If password is required to be set for the current screen.
+
+        In generall this local on-screen represents stuff like the
```
rvykydal: In general

pyanaconda/ui/gui/helpers.py
```diff
@@ -121,3 +315,154 @@ def on_back_clicked(self, window):
...
+        :param int clicks: number of ASCII check waive clicks
+        """
+        self._waive_ASCII_clicks = clicks
+
+    def check_password_empty(self, inputcheck):
+        """Check whether a password has been specified at all.
+
+           This check is used for both the password and the confirmation.
+        """
+        # If the password was set by kickstart, skip the strength check
```
rvykydal: skip the strength check -> skip the check ?

pyanaconda/users.py
M4rtinK: Fixed.

pyanaconda/ui/gui/helpers.py
M4rtinK: Fixed. :)

pyanaconda/ui/gui/helpers.py
M4rtinK: Indeed! :) Fixed.

piotrdrag:
This change worries me. Splitting sentences this way makes them essentially untranslatable to many languages — https://wiki.gnome.org/TranslationProject/DevGuidelines/Never%20split%20sentences

rvykydal:
This has been ported to Fedora in https://github.com/rhinstaller/anaconda/pull/870

PR #954, opened 2017-02-09
Enable multiple inst.ks and inst.stage2 urls (#1391724)

poncovka:
If there are defined multiple `inst.ks` urls in boot options, we enable to try the urls sequentially one by one untill we successfully fetch the kickstart file. All defined locations must be urls (only http, https or ftp), otherwise we will try only the last location. It works in the same way for `inst.stage2` option.

Resolves: rhbz#1391724

### DIFF
--- a/dracut/anaconda-lib.sh
+++ b/dracut/anaconda-lib.sh
@@ -220,6 +220,28 @@ online_netdevs() {
     done
 }
 
+# Are all given locations only http, https or ftp urls?
+# Then succeed, otherwise fail.
+are_urls() {
+    local locations="${1}"
+    local location
+
+    # No locations, failure.
+    [ -z "$locations" ] && return 1
+
+    # Check all locations.
+    for location in $locations; do
+        case "${location%%:*}" in
+            http|https|ftp) continue ;;
+            # Another type, failure.
+            *) return 1 ;;
+        esac
+    done
+
+    # Success.
+    return 0
+}
+
 # This is where we actually run the kickstart. Whee!
 # We can't just add udev rules (we'll miss devices that are already active),
 # and we can't just run the scripts manually (we'll miss devices that aren't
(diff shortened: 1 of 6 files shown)


dracut/anaconda-lib.sh
```diff
@@ -220,6 +220,27 @@ online_netdevs() {
     done
 }
 
+# Are all given locations only http, https or ftp urls?
+# Then return true, otherwise false.
+are_urls() {
+    local locations="${1}" location
```
jkonecny12: Please use another line to declare `location`. This looks like bad assignment.

dracut/anaconda-lib.sh
```diff
@@ -220,6 +220,27 @@ online_netdevs() {
...
+
+    # No locations, return false.
+    [ -z "$locations" ] && return 1
```
jkonecny12: I think you can use `return false` and `return true` here.

dracut/fetch-kickstart-net.sh
```diff
@@ -18,27 +18,50 @@ command -v getarg >/dev/null || . /lib/dracut-lib.sh
...
-    kickstart="nfs:$server:$filename"
-fi
 
-# NFS kickstart URLs that end in '/' get '$IP_ADDR-kickstart' appended
-case "$kickstart" in
-    nfs*/) kickstart="${kickstart}${new_ip_address}-kickstart" ;;
+        # Kickstart URLs that end in '/' get '$IP_ADDR-kickstart' appended.
+        case "$kickstart" in
+            nfs*/) kickstart="${kickstart}${new_ip_address}-kickstart" ;;
+        esac
```
jkonecny12: Change this `case` to `if` please.

rvykydal: Hmm, I'd rather not change a thing that used to work ... in bash ;)

dracut/anaconda-lib.sh
poncovka: Fixed.

dracut/anaconda-lib.sh
poncovka: That does not work, so I've changed the comments.

dracut/fetch-kickstart-net.sh
poncovka: Fixed and tested.

PR #956, opened 2017-02-09
Add inst.waitfornet option (#1315160)

rvykydal:
inst.waitfornet=<TIMEOUT_IN_SECONDS>

At the beginning of installation (stage 2) we run a thread waiting for NM being
in connecting state (eg when configuring IP from DHCP) so that some actions
(getting NTP servers from DHCP, getting GeoIP info, downloading payload
metadata) can wait for network connectivity

This patch is adding option to wait for connectivity synchronously at the
beginning of installation (stage 2). It is addressing automatic installation
use case where we don't want to trigger user interaction when first attempt to
connect (configure IP via dhcp) fails but following attempts would succeed.

Resolves: rhbz#1315160


data/anaconda_options.txt
```diff
@@ -241,3 +241,6 @@ The "input_ks" value disables saving of the input kickstart (if any), "output_ks
 kickstart generated by Anaconda, "all_ks" disables saving of both input and output kickstarts, "logs" disables saving
 of all installation logs and "all" disables saving of all kickstarts and all logs. Multiple values can be combined
 as a comma separated list, for example: "all_ks,logs"
+
+waitfornet
+Wait for network connectivity at the beginning of installation.
```
jkonecny12: I think you should have a note about required parameter `timeout` here.

rvykydal:
Ported to master with https://github.com/rhinstaller/anaconda/pull/1068

PR #960, opened 2017-02-13
Rescue mode should wait for the storage and luks devices (#1376638)

poncovka:
In the rescue mode, we should wait till the storage is initialized, before we start to work with it. Also we should wait for a luks device to get ready after we try to unlock it. Otherwise, we don't have to find any Linux partitions.

Resolves: rhbz#1376638

### DIFF
--- a/pyanaconda/rescue.py
+++ b/pyanaconda/rescue.py
@@ -21,8 +21,9 @@ from blivet.devices import LUKSDevice
 from blivet.osinstall import mount_existing_system, find_existing_installations
 
 from pyanaconda import iutil
-from pyanaconda.constants import ANACONDA_CLEANUP
+from pyanaconda.constants import ANACONDA_CLEANUP, THREAD_STORAGE
 from pyanaconda.constants_text import INPUT_PROCESSED
+from pyanaconda.threads import threadMgr
 from pyanaconda.flags import flags
 from pyanaconda.i18n import _, N_, C_
 from pyanaconda.kickstart import runPostScripts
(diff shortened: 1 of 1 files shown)


poncovka:
My personal tip is that we wait for lvm2-pvscan, but this solution looks to be simple and save enough.

pyanaconda/rescue.py
```diff
@@ -248,6 +246,11 @@ def _unlock_devices(self):
                                               parents=[device],
                                               exists=True)
                         self.storage.devicetree._add_device(luks_dev)
+
+                        # Wait for the device.
+                        # Otherwise, we could get a message about no Linux partitions.
+                        time.sleep(2)
```
M4rtinK: I wonder if there is something blocking we could call (udev settle/some blivet thing - I guess the storage guys might know) instead of arbitrary sleep ? Well, I can image there is nothing like that though.

poncovka: I have tried udev settle, but that does not solve the problem. I was looking for inspiration in blivet and other places, but couldn't find anything. However, maybe I have overlooked something.

jkonecny12: @vpodzime could you please point us to the right solution?

jkonecny12:
jenkins, test this please

jkonecny12:
jenkins, test this please

poncovka:
jenkins, test this please

poncovka:
The Pylint error is from BlivetGuiSpoke. Merging.

PR #961, opened 2017-02-15
Add inst.notmux option

### LINKED ISSUE #958: allow user to specify to not use tmux
I'm trying to log output from the serial console during an installation and I notice that there is not a good way to tell anaconda to not use tmux during install (and thus all terminal control characters will get printed). There exists an `anaconda-direct.service` which is exactly what I want, but the only way to have this service started instead of `anaconda-tmux.service` is to be [running on s390 hardware](https://github.com/rhinstaller/anaconda/blob/3ab9c28518921fcb95a3b4aa423cc2c1cab289d4/data/systemd/anaconda-direct.service#L7).

Can we expose this to the user in a different way? Maybe we assume `cmdline` implies `anaconda-direct` or we create an option like `inst.notmux`. Thoughts?

dustymabe:
Allows a user to execute the anaconda-direct.service rather than
anaconda-tmux.service if they want to. This allows a user to print
output to a console without any terminal control characters.

Fixes #958

### DIFF
--- a/data/systemd/anaconda-direct.service
+++ b/data/systemd/anaconda-direct.service
@@ -5,6 +5,7 @@ After=instperf.service rsyslog.service systemd-udev-settle.service NetworkManage
 Requires=anaconda.service
 # TODO: use ConditionArchitecture in systemd v210 or later
 ConditionPathIsDirectory=|/sys/hypervisor/s390
+ConditionKernelCommandLine=|inst.notmux
 
 [Service]
 Environment=HOME=/root MALLOC_CHECK_=2 MALLOC_PERTURB_=204 PATH=/usr/bin:/bin:/sbin:/usr GDK_BACKEND=x11 XDG_RUNTIME_DIR=/tmp
--- a/data/systemd/anaconda-tmux@.service
+++ b/data/systemd/anaconda-tmux@.service
@@ -4,6 +4,7 @@ Requires=anaconda.service
 After=anaconda.target anaconda.service
 # TODO: use ConditionArchitecture in systemd v210 or later
 ConditionPathIsDirectory=!/sys/hypervisor/s390
+ConditionKernelCommandLine=!inst.notmux
 
 [Service]
 Type=idle
--- a/docs/boot-options.rst
+++ b/docs/boot-options.rst
@@ -522,6 +522,14 @@ inst.noshell
 
 Do not put a shell on tty2 during install.
 
+.. inst.notmux:
+
+inst.notmux
+^^^^^^^^^^^
+
+Do not use tmux during install. This allows for output to get generated without
+terminal control characters and is really meant for non-interactive uses.
+
 .. inst.syslog:
 
 inst.syslog


dustymabe:
so the only thing I have noticed as an issue while using this is the following stacktrace from logging.. It might be to related to the way I'm using it (see [this pr](https://github.com/clalancette/oz/pull/232) to oz) or it may be something else. Either way the error is harmless as the install completes fine:

```
[  269.256389] anaconda[1564]: --- Logging error ---
[  269.261420] anaconda[1564]: Traceback (most recent call last):
[  269.266965] anaconda[1564]:   File "/usr/lib64/python3.6/logging/__init__.py", line 989, in emit
[  269.278242] anaconda[1564]:     stream.write(msg)
[  269.285490] anaconda[1564]:   File "/usr/lib64/python3.6/site-packages/pyanaconda/anaconda_log.py", line 90, in write
[  269.301887] anaconda[1564]:     wrapped_self.__wrapped__.write(*args, **kwargs)
[  269.315687] anaconda[1564]: ValueError: underlying buffer has been detached
[  269.330448] anaconda[1564]: Call stack:
[  269.335107] anaconda[1564]:   File "/usr/sbin/anaconda", line 594, in <module>
[  269.341249] anaconda[1564]:     display.setup_display(anaconda, opts, addon_paths=addon_paths)
[  269.348247] anaconda[1564]:   File "/usr/lib64/python3.6/site-packages/pyanaconda/display.py", line 254, in setup_display
[  269.360576] anaconda[1564]:     stdout_log.warning(error_message)
[  269.370624] anaconda[1564]:   File "/usr/lib64/python3.6/logging/__init__.py", line 1313, in warning
[  269.377532] anaconda[1564]:     self._log(WARNING, msg, args, **kwargs)
[  269.383559] anaconda[1564]:   File "/usr/lib64/python3.6/logging/__init__.py", line 1437, in _log
[  269.396490] anaconda[1564]:     self.handle(record)
[  269.401781] anaconda[1564]:   File "/usr/lib64/python3.6/logging/__init__.py", line 1447, in handle
[  269.416399] anaconda[1564]:     self.callHandlers(record)
[  269.421962] anaconda[1564]:   File "/usr/lib64/python3.6/logging/__init__.py", line 1509, in callHandlers
[  269.435467] anaconda[1564]:     hdlr.handle(record)
[  269.441464] anaconda[1564]:   File "/usr/lib64/python3.6/site-packages/pyanaconda/anaconda_log.py", line 72, in handle
[  269.459188] anaconda[1564]:     self.emit(record)       # pylint: disable=no-member
```

jkonecny12:
This error needs to be fixed, even if it is harmless we don't want errors in our logs.
About the patch, we need to look on this more closely to decide if adding new kernel option is right thing to do here.

dustymabe:
any ideas for debugging the error message?

rvykydal:
I am for including the patch and adding the boot option.
@dustymabe, are you still seeing the traceback from a previous comment? Obviously I am not able to hit it testing the patch locally.

dustymabe:
> I am for including the patch and adding the boot option.
> @dustymabe, are you still seeing the traceback from a previous comment? Obviously I am not able to hit it testing the patch locally.

if you can't hit it then I'm not too worried about it, must have been something with my setup at the time. Since it's a new option, if we see the traceback happen again we can fix it in a later patch?

rvykydal:
> Since it's a new option, if we see the traceback happen again we can fix it in a later patch?
I agree.

rvykydal:
@dustymabe, could you please rebase the patch on f27-devel so it is included in F27.

dustymabe:
is there a "merge" button?

PR #962, opened 2017-02-15
Lock empty root password during kickstart installation (#1383656)

M4rtinK:
Lock the root password if it is empty during a kickstart installation
(it was not specified via rootpw in kickstart and not set in the UI).

Also make the locking conditions a bit more readable when we are at it.

Resolves: rhbz#1383656

### DIFF
--- a/pyanaconda/kickstart.py
+++ b/pyanaconda/kickstart.py
@@ -1647,8 +1647,17 @@ class ReqPart(commands.reqpart.RHEL7_ReqPart):
 
 class RootPw(commands.rootpw.F18_RootPw):
     def execute(self, storage, ksdata, instClass, users):
-        if not self.password and not flags.automatedInstall:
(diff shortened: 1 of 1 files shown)


pyanaconda/kickstart.py
```diff
@@ -1647,8 +1647,17 @@ def execute(self, storage, ksdata, instClass):
 
 class RootPw(commands.rootpw.F18_RootPw):
     def execute(self, storage, ksdata, instClass, users):
-        if not self.password and not flags.automatedInstall:
+        if flags.automatedInstall and not self.password and not self.seen:
+            # Lock the root password if during an installation witk kickstart
```
vpodzime: typo^ -- "witk"

M4rtinK: Fixed! :)

M4rtinK:
Fedora PR: https://github.com/rhinstaller/anaconda/pull/982

PR #963, opened 2017-02-15
rhev: add /home /tmp /var/log /var/log/audit

dougsland:
Add support for NIST 800-53 SC-32 System Partitioning schema.

Bug-Url: https://bugzilla.redhat.com/show_bug.cgi?id=1420068


rvykydal:
I've tested the PR in kvm with RHVH-4.1-20170209.0-RHVH-x86_64-dvd1.iso (RHEL 7.3 anaconda .93) and updates image on top of current rhel7-branch, using 40 GB disk and it seems to work (you just need to visit and leave storage spoke as it is reporting "no disk selected" - same as without the patch)

rvykydal:
Jenkins, it's OK to test

rvykydal:
This will be ported in Fedora in scope of https://github.com/rhinstaller/anaconda/pull/2137

PR #965, opened 2017-02-20
util: Add script to capture logs

jcpunk:
I figured now that we've got `%onerror` shipping an easy script to capture the system state and error logs might come in handy for unifying the troubleshooting.

In theory I'd see the workflow something like:
```
%onerror
/usr/libexec/anaconda/LogCapture /tmp/mylogs.tar
scp /tmp/mylogs.tar hostname:/some/path
%end
```

### DIFF
--- a/scripts/log-capture
+++ b/scripts/log-capture
@@ -1,0 +1,43 @@
+#!/bin/bash
+
+date=$(date +%F_%R)
+OUTDIR=/tmp/LogCapture-${date}
+ARCHIVE=${1:/tmp/log-capture.tar.bz2}
+
+_to_log() { local OutputFile=$(tr ' /' '_' <<<"$@"); $@ >${OUTDIR}/${OutputFile}; }
+
+mkdir -p ${OUTDIR}
+cd ${OUTDIR}
+
+if [[ ! -f ks.cfg ]];then
+  echo "No /tmp/ks.cfg present" > ${OUTDIR}/ks.cfg
+fi
+
+_to_log cat /mnt/sysimage/root/anaconda-ks.cfg
+_to_log date
+_to_log dmesg
+_to_log dmidecode
+_to_log lspci -vvnn
+_to_log fdisk -cul
+_to_log ls -lR /dev
+_to_log dmsetup ls --tree
+_to_log lvm pvs
+_to_log lvm vgs
+_to_log lvm lvs
+_to_log cat /proc/mdstat
+_to_log cat /proc/partitions
+_to_log mount
+_to_log df -h
+_to_log cat /proc/meminfo
+_to_log cat /proc/cpuinfo
+_to_log ps axf
+_to_log lsof
+_to_log ip -s li
+_to_log ip a
+_to_log ip r
+
+cp /tmp/*.log ${OUTDIR}
+
+tar cfa ${ARCHIVE} ${OUTDIR}
+
+echo -e "\nFinished gathering data\nUpload ${ARCHIVE} to another system\n"
--- a/scripts/Makefile.am
+++ b/scripts/Makefile.am
@@ -16,7 +16,7 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 scriptsdir = $(libexecdir)/$(PACKAGE_NAME)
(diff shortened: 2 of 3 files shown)


vpodzime:
AFAICT, there's already support for ``%traceback`` scripts.

jcpunk:
That is correct.

My aim here is mostly to provide a unified tool for collecting anaconda error logs that folks in support might require.  The script itself is suitable to run under any of the sections (`%traceback`, `%onerror`, `%post` i guess too).

M4rtinK:
jenkins, test this please

jcpunk:
I'm a bit surprised that Jenkins is failing.  The logs don't seem to be public... Any suggestions on what I should fix up?

M4rtinK:
Looking into the failure, it was caused by a missing dist_ prefix for utils_SCRIPTS. But looking at the pull request in more detail I have a few suggestions:

1. I think the script should be renamed, as all our other scripts are lower case and generally use dashes between words. Something like `log-capture`.
2. I think the script should just be placed in the `scripts` directory, wit the other script. Looking at the only other script in `utils` - `handle-sshpw` - there really is no good reason for it to be there and I'll probably also move it to scripts soon. Being in scripts also make the whole makefile business simpler.

I did the changes when I was at it:

```
diff --git a/scripts/Makefile.am b/scripts/Makefile.am
index c5100c0..cb3c184 100644
--- a/scripts/Makefile.am
+++ b/scripts/Makefile.am
@@ -16,7 +16,7 @@
 
 scriptsdir = $(libexecdir)/$(PACKAGE_NAME)
-dist_scripts_SCRIPTS = upd-updates run-anaconda zramswapon zramswapoff zram-stats anaconda-pre-log-gen
+dist_scripts_SCRIPTS = upd-updates run-anaconda zramswapon zramswapoff zram-stats anaconda-pre-log-gen log-capture
 dist_noinst_SCRIPTS  = upd-kernel makeupdates makebumpver
 
 dist_bin_SCRIPTS = analog anaconda-cleanup instperf anaconda-disable-nm-ibft-plugin
diff --git a/utils/LogCapture b/scripts/log-capture
similarity index 100%
rename from utils/LogCapture
rename to scripts/log-capture
diff --git a/scripts/makeupdates b/scripts/makeupdates
index a457453..f3ac744 100755
--- a/scripts/makeupdates
+++ b/scripts/makeupdates
@@ -343,7 +343,7 @@ def copyUpdatedFiles(tag, updates, cwd, builddir):
         elif gitfile == "utils/handle-sshpw":
             install_to_dir(gitfile, "usr/sbin")
         elif any(gitfile.endswith(libexec_script) for libexec_script in \
-                 ("zramswapon", "zramswapoff", "zram-stats")):
+                 ("zramswapon", "zramswapoff", "zram-stats", "log-capture")):
             install_to_dir(gitfile, "usr/libexec/anaconda")
         elif gitfile.endswith("AnacondaWidgets.py"):
             import gi
diff --git a/utils/Makefile.am b/utils/Makefile.am
index 7d13f5f..97c3396 100644
--- a/utils/Makefile.am
+++ b/utils/Makefile.am
@@ -19,7 +19,4 @@ SUBDIRS = dd
 
 dist_sbin_SCRIPTS    = handle-sshpw
 
-utilsdir             = $(libexecdir)/$(PACKAGE_NAME)
-utils_SCRIPTS        = LogCapture
-
 MAINTAINERCLEANFILES = Makefile.in
```

So if you are fine with these suggestions you can just apply the diff, move the file & update the pull request. After that I think we are all set to finally merge it. :)

jcpunk:
That all looks beautiful to me.  I'm happy with that diff.

scripts/log-capture
```diff
@@ -0,0 +1,43 @@
+#!/bin/bash
+
+date=$(date +%F_%R)
+OUTDIR=/tmp/LogCapture-${date}
+ARCHIVE=${1:/tmp/LogCapture.tar.bz2}
```
M4rtinK: Just a small nitpick - I think this should also be called "log-capture" - for consistency.

jcpunk:
Fixed

M4rtinK:
All set now, mergin to master. :)

PR #966, opened 2017-02-21
network: add dracut vlan= argument for iSCSI root via vlan (#1374003)

rvykydal:
Resolves: rhbz#1374003

### DIFF
--- a/pyanaconda/network.py
+++ b/pyanaconda/network.py
@@ -375,6 +375,33 @@ def dracutBootArguments(devname, ifcfg, storage_ipaddr, hostname=None):
             netargs.add("team=%s:%s" % (devname,
                                         ",".join(dev for dev, _cfg in slaves)))
 
+        if ifcfg.get("TYPE") == "Vlan":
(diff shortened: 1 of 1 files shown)


pyanaconda/network.py
```diff
@@ -375,6 +375,20 @@ def dracutBootArguments(devname, ifcfg, storage_ipaddr, hostname=None):
             netargs.add("team=%s:%s" % (devname,
                                         ",".join(dev for dev, _cfg in slaves)))
 
+        if ifcfg.get("TYPE") == "Vlan":
+            physdev = ifcfg.get("PHYSDEV")
+            if physdev not in nm.nm_devices() and len(physdev) > 20:
```
jkonecny12: Could you please write a comment here to explain why `len(physdev) > 20`.

rvykydal:
I had to rework the patch as it was not handling some cases (if PHYSDEV was specified by device name, it didn't add zd.net arguments as it was using ifcfg of vlan device instead of the parent)

rvykydal:
Jirka, could you please look at the new version of the patch? Thank you.

PR #967, opened 2017-02-22
Support DD rpm loading from local disk device (#1377233)

jkonecny12:
You can now use inst.dd=hd:LABEL=DD:/dd.rpm as kernel boot option.

Changes:
driver-updates-genrules.sh - parse LABEL=DD and similar correctly

driver_updates.py
Optional parameter RPMPATH on --disk param, this parameter is path to the rpm file. If this parameter is present mount device and process the rpm on the mounted device like usual.

*Resolves: rhbz#1377233*

### DIFF
--- a/dracut/driver_updates.py
+++ b/dracut/driver_updates.py
@@ -27,10 +27,11 @@ once for interactive mode (if requested).
 
 Usage is one of:
 
-    driver-updates --disk DISKSTR DEVNODE
+    driver-updates --disk DISKSTR DEVNODE [RPMPATH]
 
         DISKSTR is the string passed by the user ('/dev/sda3', 'LABEL=DD', etc.)
         DEVNODE is the actual device node or image (/dev/sda3, /dev/sr0, etc.)
+        RPMPATH is the path to the rpm file on the DEVNODE mountable device
 
         DEVNODE must be mountable, but need not actually be a block device
         (e.g. /dd.iso is valid if the user has inserted /dd.iso into initrd)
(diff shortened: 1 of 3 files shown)


Frodox:
What is meaning of dd ..?

jkonecny12:
As I already answered on IRC `Driver Disk` :)

dracut/driver_updates.py
```diff
@@ -491,13 +492,25 @@ def _process_driver_disk(dev, interactive=False):
...
-        return _process_driver_rpm(rpm)
+        if dev:
+            return _mount_and_process_driver_rpm(rpm, dev)
+        else:
+            return _process_driver_rpm(rpm)
     except (subprocess.CalledProcessError, IOError) as e:
         log.error("ERROR: %s", e)
         return {}
 
+def _mount_and_process_driver_rpm(rpm, dev):
```
M4rtinK: If I read this correctly, the `mounted` context managed unmounts the device once the block is processed, right ? In such case the name of the function (`_mount_and_process_driver_rpm`) seems misleading - it sounds like the device remains mounted after the function has been called.

Sp I recommend changing changing the name to something like `_process_driver_rpm_from_device()`. :)

dracut/driver_updates.py
```diff
@@ -491,13 +492,25 @@ def _process_driver_disk(dev, interactive=False):
 
     return modules
 
-def process_driver_rpm(rpm):
+def process_driver_rpm(rpm, dev = None):
```
rvykydal: I'd just leave out the spaces in default argument for consistency (dev=None).
Sorry I don't have any better remark to the patch, looks good to me.

dracut/driver_updates.py
jkonecny12: Good point, fixed.

dracut/driver_updates.py
jkonecny12: Thanks, fixed.
No need to be sorry for this, that is always a good mark.

PR #975, opened 2017-03-01
Fix tests by renaming packaging to payload

jkonecny12:
`Pylint` and `nosetests` are failing now because `python3-setuptools` have a new dependency which is called 'packaging'. So when our tests set the `PYTHONPATH` to `pyanaconda` directory then we will replace the python3 standard `packaging` module.

Renaming from `pyanaconda/packaging` to `pyanaconda/payload` should fix this issue.

### DIFF
--- a/anaconda.py
+++ b/anaconda.py
@@ -635,7 +635,7 @@ if __name__ == "__main__":
         anaconda.storage.setup_disk_images()
 
     from blivet.osinstall import storage_initialize
-    from pyanaconda.packaging import payloadMgr
+    from pyanaconda.payload import payloadMgr
     from pyanaconda.timezone import time_initialize
 
     if not flags.dirInstall:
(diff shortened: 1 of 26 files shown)


poncovka:
Looks good, but why do we need to have the `pyanaconda` submodules on `PYTHONPATH` instead of using the `pyanaconda` module itself? Are we going to rename our modules every time we get a collision?

jkonecny12:
Good point, `PYTHONPATH` for `pyanaconda` directory should be removed. I don't know why this is there and I think it should be removed as you said. So or so this new name give more sense to me and it is fixing the issue we have now. IMHO it should be merged.
We can create followup to fix the `PYTHONPATH` setting later.

**UPDATED:**
1) Fix typo in `pyanaconda/payload/Makefile.am` - missing `/`
2) Fix pylint error when overlapping `payload` global variable by local variable

**NOTE:**
Pylint still has an error, but it looks like it is because of Blivet-gui error.
If somebody will give ACK here, could you please merge this too. It would be nice to fix tests ASAP.

jkonecny12:
Thank you!

PR #976, opened 2017-03-02
Fixes in tui software selection

poncovka:
Some small fixes in the tui software selection spoke and the yum payload:

* If the software selection is defined by a kickstart file, we should remember it and use it until a new software selection is made. Resolves: rhbz#1371229
* The software spoke should not be initialized in the `status` method, but in the `initialize` method.
* If the `checkSoftwareSelection` finds a dependency error, we should report an error instead of showing the source changed message. Resolves: rhbz#979307
* Payload should be released also in a case that a dependency error is found.
* Release can raise an exception if we try to release resources from different threads. That should not happen and we should fix these cases. However, we don't want the installer to fail because of that either. Therefore, all exceptions from release will be catched and logged only.

### DIFF
--- a/pyanaconda/packaging/yumpayload.py
+++ b/pyanaconda/packaging/yumpayload.py
@@ -441,18 +441,27 @@ reposdir=%s
         """Release package sacks and close their connections to the database."""
         with _yum_lock:
             # Check if the meta package sack was created.
-            if self._yum._pkgSack is not None:
+            if self._yum._pkgSack is None:
+                return
 
+            try:
                 # Drop cached data in every sack.
                 self._yum.pkgSack.dropCachedData()
 
                 # Close every sack.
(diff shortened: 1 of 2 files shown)


rvykydal:
Looks good to me.
I wonder if we can be comparing _selected None and 0 in status property of the patched file but I think it wouldn't change the behavior anyhow so perhaps never mind.

M4rtinK:
jenkins, test this please

M4rtinK:
(re-triggering tests now that the redundant & broken help content generation step has been removed from the makefile)

poncovka:
Fixed and added new commits with fixes of bugs I have found.

rvykydal:
Looks good to me, but each commit should have a rhbz Resolves/Related reference so please add them (two are missing it).

poncovka:
Fixed.

PR #977, opened 2017-03-02
Use system Python when running Anaconda

M4rtinK:
Change the shebang in the main Anaconda script to use system Python.
This should result in get_python_lib() returning the correct path to
site packages even once the Python side change (bug 1426435) is reapplied.

### DIFF
--- a/anaconda.py
+++ b/anaconda.py
@@ -1,4 +1,4 @@
-#!/usr/bin/python3
+#!/usr/libexec/system-python
 #
 # anaconda: The Red Hat Linux Installation program
 #
--- a/anaconda.spec.in
+++ b/anaconda.spec.in
@@ -83,6 +83,7 @@ The anaconda package is a metapackage for the Anaconda installer.
 
 %package core
 Summary: Core of the Anaconda installer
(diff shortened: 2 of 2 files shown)


M4rtinK:
jenkins, test this please

M4rtinK:
@rvykydal Yeah, that's also something I have been thinking about - if we use system python & the planed "don't ignore PYTHONPATH" flag that should be added for system-python in the future there should be basically no change in functionality from our PoV. We might even drop that flag in places where it is not needed, this increasing the chance the given tool works if the non-system Python is broken.

But just in case I want to ask the Python guys a bit about some details about system-python, such as general policy for helper scripts (eq. not stuff that runs in the installation environment but is used for build for example - should it always use system Python ?) or system-python & Dracut, etc. 

And one more thing that's not really clear to me - I thought system-python was supposed to be used only by "system" stuff, which can then be coordinated more easily for system Python version and behavior changes. But at the moment it feels more like "if you don't want your stuff to break due to changes in behavior, use system-python" - so that's also something I want to clarify before we switch the rest. :)

PR #979, opened 2017-03-03
Remove redundant help content generation code (#1428787)

M4rtinK:
Remove redundant help content generation code from the makefile
thart snuck in during the rebase to F21 Anaconda.

It has been apparently running, unnoticed, during each "make release",
not interfering with the help content that is actually used,
which is provided by a separate package (anaconda-user-help).

With fedorahosted going away this peice of code finaly stopped
working, making use aware of its presence, so we can finally retire it.

Resolves: rhbz#1428787

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -66,11 +66,11 @@ po-empty:
 		exit 1 ; \
 	done
 
-scratch: po-empty get-help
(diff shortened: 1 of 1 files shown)


M4rtinK:
jenkins, test this please

PR #981, opened 2017-03-03
We should not have pyanaconda submodules on PYTHONPATH

poncovka:
If we add a path to the pyanaconda directory on PYTHONPATH, then
names of pyanaconda submodules can collide with names of other
modules. Therefore, we should avoid that.

Note: Tested with pyanaconda.unittest module and the tests passed.


M4rtinK:
Looks like Alpha freeze is tomorrow - so merging now. :)

PR #985, opened 2017-03-08
Add support for kickstart %onerror scripts (#1412538)

poncovka:
These run on most error conditions (including when %traceback scripts
would run), but note that they do not run on other kickstart script
failures. That would be strange.

The runOnErrorScripts function is in iutil, not kickstart, because it
gets imported all over the place and this would lead to import cycles.

(cherry picked from commit e470d17)

Resolves: rhbz#1412538

PR #986, opened 2017-03-08
Support --nohome option in the autopartitioning (#663099)

poncovka:
**Depends on a new version of pykickstart, that is changed here:** https://github.com/rhinstaller/anaconda/pull/985

Default partitioning can be modified in an install class's method
customizeDefaultPartitioning, where can be applied selections made in
a kickstart file. This method can be easily extended for future
--noswap and --noboot options.

The base install class will remove a request for a /home partition,
if --nohome option is selected in a kickstart file. In that case,
the /home partition will not be created.

Subclasses of the base install class can easily ignore this
customization if they override the customizeDefaultPartitioning
method with the pass statement.

Resolves: rhbz#663099

### DIFF
--- a/pyanaconda/installclass.py
+++ b/pyanaconda/installclass.py
@@ -116,12 +116,28 @@ class BaseInstallClass(object):
 
         storage.autoPartitionRequests = autorequests
 
+    def customizeDefaultPartitioning(self, storage, data):
+        # Customize the default partitioning with kickstart data.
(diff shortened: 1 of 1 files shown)


poncovka:
This needs to be tested after https://github.com/rhinstaller/anaconda/pull/985 is merged.

poncovka:
jenkins, test this please

PR #987, opened 2017-03-08
Rhel7 fix 1404158

jkonecny12:
Fix Software spoke.

1) Fix Software spoke radio buttons, you don't need to click twice on them now.
2) Fix Software spoke with partial kickstart. Now it will select groups from your kickstart file.

### DIFF
--- a/pyanaconda/packaging/yumpayload.py
+++ b/pyanaconda/packaging/yumpayload.py
@@ -1165,6 +1165,25 @@ reposdir=%s
 
             return (group.ui_name, group.ui_description)
 
+    def groupId(self, group_name):
+        """Translate group name to group ID.
+
+        :param group_name: Valid identifier for group specification.
+        :returns: Group ID.
+        :raise NoSuchGroup: If group_name doesn't exists.
+        :raise PayloadError: When Yum's groups are not available.
+        """
+        groups = self._yumGroups
+        if not groups:
+            raise PayloadError("Yum's groups are not available")
+
+        with _yum_lock:
+            if not groups.has_group(group_name):
+                raise NoSuchGroup(group_name)
+
+            group = groups.return_group(group_name)
+            return group.groupid
+
     def _isGroupVisible(self, groupid):
         groups = self._yumGroups
         if not groups:
--- a/pyanaconda/ui/gui/spokes/software.py
+++ b/pyanaconda/ui/gui/spokes/software.py
@@ -32,6 +32,7 @@ from pyanaconda.ui.gui.spokes import NormalSpoke
 from pyanaconda.ui.gui.spokes.lib.detailederror import DetailedErrorDialog
 from pyanaconda.ui.gui.utils import gtk_action_wait, escape_markup
 from pyanaconda.ui.categories.software import SoftwareCategory
+from pyanaconda.packaging import PayloadError
 
 import logging
 log = logging.getLogger("anaconda")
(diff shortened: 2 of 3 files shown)


pyanaconda/ui/gui/spokes/software.py
```diff
@@ -464,6 +461,19 @@ def _get_selected_addons(self):
 
         return retval
 
+    def _mark_addon_selection(self, grpid, active):
```
poncovka: Maybe `selected` instead of `active`? There is no gui in this function, so `active` does not make much sense.

pyanaconda/ui/gui/spokes/software.py
```diff
@@ -156,7 +160,12 @@ def _payload_error(self):
         hubQ.send_message(self.__class__.__name__, payloadMgr.error)
 
     def _apply(self):
-        if self.environment and not (flags.automatedInstall and self.data.packages.seen):
+        # Environment have to be set by GUI but not by kickstart
```
M4rtinK: Rewording suggestion:

`Environment needs to be set during a GUI installation, but is not required for a kickstart install (even partial)`

pyanaconda/packaging/yumpayload.py
```diff
@@ -1165,6 +1165,24 @@ def groupDescription(self, groupid):
...
+    def groupId(self, group_name):
+        """Translate group name to group ID.
+
+        :param group_name: Valid identifier for group specification.
+        :returns: Group ID or empty string if nothing is found.
+        :raise NoSuchGroup: If group_name doesn't exists.
+        """
+        groups = self._yumGroups
+        if not groups:
+            return ""
```
poncovka: This will cause that  `self._addonStates[""]` will be set several times to `self._ADDON_SELECTED`. Maybe it would be better to raise an exception, since we catch `PayloadError` anyway when `selectedGroupsIDs` is called.

pyanaconda/ui/gui/spokes/software.py
jkonecny12: Sounds better to me. Fixed.

pyanaconda/ui/gui/spokes/software.py
jkonecny12: Fixed.

pyanaconda/packaging/yumpayload.py
jkonecny12: I want to have the *same* behavior as `groupDescription()` method has, this method is returning `(groupid, groupid)`.

It would be unexpected behavior when one method is returning something and second similar method below raising an exception.

Also catching an exception takes more performance than rewriting value in a dictionary.

jkonecny12:
**UPDATED**
Comment should be fixed now. Thanks!

pyanaconda/packaging/yumpayload.py
poncovka: Ok, but we could at least return `None` and check the id before we modify `self._addonStates`.

jkonecny12: I don't want to see `return None`. I think it is straight way to hard debugging because you never know where this will propagate in future, and None is acting as normal value in many ways.

However you are right that the empty string is not the right solution here. So last man standing, use an exception. I don't like it here either for reasons above but it is smallest evil right now. Therefore I'm going to change this to exception.

vpodzime: It's an unexpected/exceptional case, right?

jkonecny12:
Is it OK now @poncovka ?

PR #995, opened 2017-03-13
Set liveimg min_size to be x3 the real size

yuvalturg:
When using liveimg to install from a fairly large sparse file, the
minimum required size for the partition is huge, even if the content of
the payload is very small

### DIFF
--- a/pyanaconda/payload/livepayload.py
+++ b/pyanaconda/payload/livepayload.py
@@ -294,7 +294,7 @@ class LiveImageKSPayload(LiveImagePayload):
         if not os.path.exists(self.data.method.url[7:]):
             return "file does not exist: %s" % self.data.method.url
 
-        self._min_size = os.stat(self.data.method.url[7:])[stat.ST_SIZE] * 3
+        self._min_size = os.stat(self.data.method.url[7:]).st_blocks * 512 * 3
         return None
 
     def setup(self, storage, instClass):


InStallker:

pyanaconda/payload/livepayload.py
```diff
@@ -294,7 +294,7 @@ def _setup_file_image(self):
         if not os.path.exists(self.data.method.url[7:]):
             return "file does not exist: %s" % self.data.method.url
 
-        self._min_size = os.stat(self.data.method.url[7:])[stat.ST_SIZE] * 3
+        self._min_size = os.stat(self.data.method.url[7:]).st_blocks * 512 * 3
```
sandrobonazzola: Any chance here block size can be different than 512, for example 4k?

yuvalturg: st_blocks is defined as "Number of 512B blocks allocated" (from man 2 stat)

jkonecny12:
Jenkins, it's ok to test.

jkonecny12:
jenkins, test this please

jkonecny12:
Tests above will not fail after merge or rebase. I've run these tests locally and they seems to be working fine. Also I've tested installation with a sparse files and it looks to be working properly. 
So I'm merging this now.

PR #996, opened 2017-03-14
Show the network spoke in Initial Setup TUI (#1302165)

M4rtinK:
Also disable some installation environment specific options
and make sure network configuration changes are applied immediately.

Resolves: rhbz#1302165


rvykydal:
Seems to be ported to master by https://github.com/rhinstaller/anaconda/commit/6487ba01c21ea290dfe95e50088f02b8a6c27493 but I'll better let @M4rtinK check.

rvykydal:
It is in the master, removing the label.

PR #998, opened 2017-03-17
util: Add script to capture logs (#1431313)

M4rtinK:
Also handle anaconda.ks not yet being created and run fdisk on
all harddrives available in the installation environment.

Resolves: rhbz#1431313

### DIFF
--- a/scripts/log-capture
+++ b/scripts/log-capture
@@ -1,0 +1,58 @@
+#!/bin/bash
+
+date=$(date +%F_%R)
+OUTDIR=/tmp/log-capture-${date}
+ARCHIVE=${1-/tmp/log-capture.tar.bz2}
+
+_to_log() { local OutputFile=$(tr ' /' '_' <<<"$@"); $@ >${OUTDIR}/${OutputFile}; }
+
+mkdir -p ${OUTDIR}
+cd ${OUTDIR}
+
+if [[ ! -f /tmp/ks.cfg ]];then
+  echo "# No /tmp/ks.cfg present" > ${OUTDIR}/ks.cfg
+else
+    _to_log cat /mnt/sysimage/root/ks.cfg
+fi
+
+# the /mnt/sysimage/root/anaconda-ks.cfg file
+# might not yet exist when this script is called
+if [[ ! -f  /mnt/sysimage/root/anaconda-ks.cfg ]];then
+    echo "# No /mnt/sysimage/root/anaconda-ks.cfg" > ${OUTDIR}/anaconda-ks.cfg
+else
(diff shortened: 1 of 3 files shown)


M4rtinK:
There are some fixes from the original Fedora version:
- the ARCHIVE variable is now resolved correctly
- handle /mnt/sysimage/root/anaconda-ks.cfg not yet being present
- run fdisk on all available disks

I'll backport these fixes back to Fedora once this rhel7-branch PR is merged.

M4rtinK:
@jcpunk Can you check the fixes look fine ? Thanks in advance!

jcpunk:
looks good to me

M4rtinK:
Good point, I've added the full path (`/tmp/ks.cfg`) on line 12. :)

jcpunk:
👍

scripts/log-capture
```diff
@@ -0,0 +1,57 @@
...
+cd ${OUTDIR}
+
+if [[ ! -f /tmp/ks.cfg ]];then
+  echo "No /tmp/ks.cfg present" > ${OUTDIR}/ks.cfg
+fi
+
+# the /mnt/sysimage/root/anaconda-ks.cfg file
+# might not yet exist when this script is called
+if [[ ! -f  /mnt/sysimage/root/anaconda-ks.cfg ]];then
+    echo "No /mnt/sysimage/root/anaconda-ks.cfg" > ${OUTDIR}/ks.cfg
```
rvykydal: Don't we want to append (>>) to the ks.cfg instead of creating it (>) so that previous message from line 13 is preserved if it is the case?

jcpunk: Good call, if we do that, it should probably pre-appended with a `#` so that any generated kickstarts remain valid.

M4rtinK: Actually, shouldn't the output be anaconda-ks.cfg ? Eq. two separate files ? Also I think we should include /tmp/ks.cfg in the tarball as well.

M4rtinK:
A couple changes:
- prepend "no kickstart file found" warning with "# "
- also include the input kickstart (`/tmp/ks.cfg`) in the tarball
- "no kickstart" warning for `anaconda-ks.cfg` is now in `anaconda-ks.cfg` in the tarball
- the output directory is now called the same as the script (`log-capture-<date>`) for better consistency

jcpunk:
Probably worthwhile to get these changes into `master`

M4rtinK:
@jcpunk Sure, I plan to do that. :)

M4rtinK:
@jcpunk https://github.com/rhinstaller/anaconda/pull/1005

PR #999, opened 2017-03-20
Enable flexible storage checking on Fedora

poncovka:
Cherry picked from https://github.com/rhinstaller/anaconda/pull/972.

* `StorageChecker` was renamed to a more specific name `StorageCheckHandler`, so we can define a class `StorageChecker `for actual storage checking.

* The function `sanity_check` was replaced with many small functions for storage checking that can be registered with `storage_checker` and called with defined constraints. This enables us to make the
    storage checking more flexible, since we can dynamically register new checks, change the constraints or just skip some checks.

* Enabled install classes to customize the storage checking, so they can modify the storage checker and customize it to their needs.

* `SanityError` and `SanityWarning` exceptions were removed, because they were never raised and we only used them to distinguish between errors and warnings. That is not necessary now.

### DIFF
--- a/anaconda.py
+++ b/anaconda.py
@@ -617,6 +617,15 @@ if __name__ == "__main__":
     if not anaconda.interactive_mode:
         flags.ksprompt = False
 
+    # Set minimal ram size to the storage checker.
+    if anaconda.display_mode == constants.DisplayModes.GUI:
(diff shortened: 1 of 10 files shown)


poncovka:
jenkins, test this please

jkonecny12:
jenkins, test this please

PR #1001, opened 2017-03-21
Display progress for the post installation phase

M4rtinK:
Instead of just telling the user that we are "Performing
post installation setup tasks" for many minutes display
detailed progress for the post installation phase.

Both post-scriptlets and installed package verification
are now displayed in the UI as well as logged in packaging.log.

Scriptlets that are run during the installation phase are just logged,
but not displayed in the UI to not needlessly interrupt the flow of
"installing foo-bar 15/345" messages.

### DIFF
--- a/anaconda.spec.in
+++ b/anaconda.spec.in
@@ -20,7 +20,7 @@ Source0: %{name}-%{version}.tar.bz2
 
 %define gettextver 0.19.8
 %define pykickstartver 2.32-1
-%define dnfver 2.0.0
+%define dnfver 2.2.0
 %define partedver 1.8.1
 %define pypartedver 2.5-2
 %define nmver 1.0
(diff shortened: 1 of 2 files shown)


M4rtinK:
To try this out latest DNF with https://github.com/rpm-software-management/dnf/pull/763 on top needs to be used. Also the python3-gpg and gpgme packages, recently added dnf dependencies, need to be added.

So I've prepared an updates image that contains what's needed:
https://github.com/rpm-software-management/dnf/pull/763

The updates image should work fine with recent Rawhide and likely with F26 as well.

M4rtinK:
Looks like the needed API will be present in DNF 2.2 - so we need to wait for that to hit F26 + Rawhide and than we can merge this.

M4rtinK:
Waiting for DNF 2.2 to hit F26, otherwise merging this would result in CI errors on the F26 branch.

PR #1002, opened 2017-03-22
Add snapshot feature to Rhel 7

jkonecny12:
This feature adding possibility to create snapshot for thin pool LVM devices. It is possible to create snapshot before and after the installation by using `--when` parameter which is mandatory.

Adding snapshot KS command:

``snapshot <origin_vg/origin_lv> --name=<snapshot_name> --when=<pre-install | post-install>``

*Resolves: rhbz#1113207*

This PR depends on: https://github.com/rhinstaller/pykickstart/pull/152 and https://github.com/rhinstaller/blivet/pull/563 .

This will be backported to Rawhide later.

### DIFF
--- a/pyanaconda/install.py
+++ b/pyanaconda/install.py
@@ -34,6 +34,7 @@ from pyanaconda.threads import threadMgr
 from pyanaconda.ui.lib.entropy import wait_for_entropy
 from pyanaconda.kexec import setup_kexec
 from pyanaconda.kickstart import runPostScripts, runPreInstallScripts
+from pykickstart.constants import SNAPSHOT_WHEN_POST_INSTALL
 import logging
 import blivet
 log = logging.getLogger("anaconda")
@@ -72,7 +73,7 @@ def doConfiguration(storage, payload, ksdata, instClass):
     if willRunRealmd:
         step_count += 1
 
-    if ksdata.snapshot:
+    if ksdata.snapshot and ksdata.snapshot.has_snapshot(SNAPSHOT_WHEN_POST_INSTALL):
         step_count += 1
 
     progress_init(step_count)
@@ -134,7 +135,7 @@ def doConfiguration(storage, payload, ksdata, instClass):
     # write out the user interaction config file
     screen_access.sam.write_out_config_file()
 
-    if ksdata.snapshot:
+    if ksdata.snapshot and ksdata.snapshot.has_snapshot(SNAPSHOT_WHEN_POST_INSTALL):
         with progress_report(N_("Creating snapshots")):
             ksdata.snapshot.execute(storage, ksdata, instClass)
 
--- a/pyanaconda/install.py
+++ b/pyanaconda/install.py
@@ -28,7 +28,7 @@ from pyanaconda import flags
 from pyanaconda import iutil
 from pyanaconda import timezone
 from pyanaconda import network
-from pyanaconda.i18n import _
+from pyanaconda.i18n import _, N_
 from pyanaconda import screen_access
 from pyanaconda.threads import threadMgr
 from pyanaconda.ui.lib.entropy import wait_for_entropy
@@ -70,6 +70,9 @@ def doConfiguration(storage, payload, ksdata, instClass):
     # increment the counter as the
     # real joining step will be executed
     if willRunRealmd:
+        step_count += 1
+
+    if ksdata.snapshot:
         step_count += 1
 
     progress_init(step_count)
@@ -130,6 +133,10 @@ def doConfiguration(storage, payload, ksdata, instClass):
 
     # write out the user interaction config file
     screen_access.sam.write_out_config_file()
+
+    if ksdata.snapshot:
+        with progress_report(N_("Creating snapshots")):
+            ksdata.snapshot.execute(storage, ksdata, instClass)
 
     progress_complete()
 
--- a/pyanaconda/kickstart.py
+++ b/pyanaconda/kickstart.py
@@ -21,7 +21,7 @@
 from pyanaconda.errors import ScriptError, errorHandler
 from blivet.deviceaction import ActionCreateFormat, ActionDestroyFormat, ActionResizeDevice, ActionResizeFormat
 from blivet.devices import LUKSDevice
-from blivet.devices.lvm import LVMCacheRequest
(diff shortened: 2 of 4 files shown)


pyanaconda/kickstart.py
```diff
@@ -1915,6 +1914,70 @@ def execute(self, *args):
...
+            Blivet will do a validity checking for future snapshot.
+        """
+        log.debug("Snapshot name %s setup", self.name)
+        if not self.origin.count('/') == 1:
+            raise KickstartParseError(
+                        formatErrorMsg(self.lineno,
+                                       msg=_("Incorrectly specified origin of the snapshot."
+                                             " Use format \"VolGroup/LV_name\"")))
+        origin = self.origin.replace("/","-")
+        origin_dev = storage.devicetree.getDeviceByName(origin)
```
vpodzime: Please check if this works with an *LV* that has ``-`` in its name -- could be (but shouldn't) that it is as ``VG_NAME-LV--NAME`` in devicetree (cause that's the name of the *DM* map).

pyanaconda/kickstart.py
```diff
@@ -1915,6 +1914,70 @@ def execute(self, *args):
...
+
+        if not isinstance(origin_dev, LVMThinLogicalVolumeDevice):
+            raise KickstartParseError(
+                        formatErrorMsg(self.lineno,
+                                       msg=(_("Snapshot: origin \"%(origin)s\" of snapshot \"%(name)s\""
+                                              " is not a valid thin pool device.") %
```
vpodzime: thin *LV* not a pool

pyanaconda/kickstart.py
```diff
@@ -1915,6 +1914,70 @@ def execute(self, *args):
...
+                                                       parents=[origin_dev.pool],
+                                                       segType="thin",
+                                                       origin=origin_dev)
+        except ValueError as e:
+            raise KickstartParseError(formatErrorMsg(self.lineno, msg=e))
+
+    def execute(self, storage, ksdata, instClass):
+        """ Register an action for snapshot creation.
+
+            Snapshot must be re-created because method blivet.reset() was called.
```
vpodzime: I don't think ``blivet.reset()`` is being called. And if so, the ``self.thin_snapshot`` is hardly in the the devicetree.

pyanaconda/install.py
```diff
@@ -131,6 +134,10 @@ def doConfiguration(storage, payload, ksdata, instClass):
     # write out the user interaction config file
     screen_access.sam.write_out_config_file()
 
+    if ksdata.snapshot:
+        with progress_report(N_("Creating snapshots")):
+            ksdata.snapshot.execute(storage, ksdata, instClass)
+
```
vpodzime: This (and the above ``step_count`` incrementation) should only happen when creating a ``post-install`` snapshot, right? I believe this should be fixed up in the next commit.

pyanaconda/kickstart.py
```diff
@@ -1915,6 +1914,70 @@ def execute(self, *args):
...
+        for snap_data in self.dataList():
+            snap_data.setup(storage, ksdata, instClass)
+
+    def execute(self, storage, ksdata, instClass):
+        """ Create ThinLV snapshot.
+
+            The Blivet must be reseted before creation of the snapshot. This is
+            required because the storage could be changed in post section.
+        """
+        try_populate_devicetree(storage.devicetree)
```
vpodzime: But that would require re-creation of the snapshot device (which could fail, actually). So maybe this is safer from users' perspective, after all.

pyanaconda/kickstart.py
jkonecny12: This is working correctly but also the *good* variant is working so I'm changing this to the correct new.

pyanaconda/kickstart.py
jkonecny12: Thanks, sorry about this. Fixed.

pyanaconda/kickstart.py
jkonecny12: This is leftover from previous version. Fixed.

pyanaconda/install.py
jkonecny12: True, fixed.
I've created new checking method for this which is used also before calling `setup` and `execute` for pre snapshots.

pyanaconda/kickstart.py
jkonecny12: I think I had this problem with re-creation failing so I changed it here to this. Maybe it won't be able to catch everything in post but it's better than nothing.

pyanaconda/kickstart.py
```diff
@@ -1915,6 +1914,70 @@ def execute(self, *args):
...
+
+            This will also do the checking of snapshot validity.
+        """
+        for snap_data in self.dataList():
+            snap_data.setup(storage, ksdata, instClass)
+
+    def execute(self, storage, ksdata, instClass):
+        """ Create ThinLV snapshot.
+
+            The Blivet must be reseted before creation of the snapshot. This is
```
jkonecny12: Fixed, thanks.

jkonecny12:
I hope everything is fixed now. Thanks for review @vpodzime .

pyanaconda/kickstart.py
```diff
@@ -1915,6 +1919,130 @@ def execute(self, *args):
...
+
+        if not isinstance(origin_dev, LVMThinLogicalVolumeDevice):
+            raise KickstartParseError(
+                        formatErrorMsg(self.lineno,
+                                       msg=(_("Snapshot: origin \"%(origin)s\" of snapshot \"%(name)s\""
+                                              " is not a valid thin LV device.") %
+                                            {"origin": self.origin,
+                                             "name": self.name})))
+
+        if storage.devicetree.getDeviceByName("%s-%s" % (origin_dev.vg.name, self.name)):
```
vpodzime: The same trick with ``replace('-','--')`` could/might be needed here too.

jkonecny12: It should be fixed now.

jkonecny12:
Add new commit which bump required version of pykickstart and blivet.

jkonecny12:
jenkins, test this please

jkonecny12:
jenkins, test this please

PR #1004, opened 2017-03-24
Use 64bit tools on 32bit firmware for x86 efi machines (#1310779)

vathpela:
This enables Baytrail and similar atom CPUs that typically ship with a
32-bit firmware, but have a 64-bit capable CPU.

Resolves: rhbz#1310779

Signed-off-by: Peter Jones <pjones@redhat.com>

### DIFF
--- a/pyanaconda/bootloader.py
+++ b/pyanaconda/bootloader.py
@@ -1388,7 +1388,7 @@ class GRUB2(GRUB):
 
     """
     name = "GRUB2"
-    packages = ["grub2"]
+    packages = ["grub2-pc", "grub2-tools"]
     _config_file = "grub.cfg"
     _config_dir = "grub2"
     _passwd_file = "user.cfg"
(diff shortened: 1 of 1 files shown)


M4rtinK:
I'm getting this when I try to run with the PR applied on latest rhel7-branch running on top of the 20170323 nightly in a uefi VM (created by virt manager in Fedora 25):
pjones_uefi_test_1

Is this expected behavior (eq. something else has to land for this to work correctly) ? Is there maybe some dependency that can be added/bumped to account for that ?

hadess:
@M4rtinK you need a kernel with support for this functionality. See https://bugzilla.redhat.com/show_bug.cgi?id=1434959

M4rtinK:
> @M4rtinK you need a kernel with support for this functionality. See https://bugzilla.redhat.com/show_bug.cgi?id=1434959

In such case we need to wait for the appropriate kernel to hit the repositories and also maybe add a Requires for the proper kernel version. 

Also I think the patch should be probably modified to not crash, but default to 64 bit uefi & log a warning if /sys/firmware/efi/fw_platform_size does not exist in case users still manage to run Anaconda with an older kernel version.

PR #1010, opened 2017-04-03
Pylint fixes (#1429576)

rvykydal:
Related: rhbz#1429576


rvykydal:
They slipped through perhaps because it was an external PR.

PR #1013, opened 2017-04-06
Call subprocess.Popen with absolute path to a binary (#1411407)

rvykydal:
Resolves: rhbz#1411407

Python is trying to execve the program (specified without the path) in the
fork, iterating over the path list and calling strerror when failing on paths
before it finds the right one. The strerror call can then hang waiting forever
for a lock held by parent process (__libc_setlocale_lock).

NOTE: in the future we want to replace this workaround by using subprocess32
module.

### DIFF
--- a/pyanaconda/iutil.py
+++ b/pyanaconda/iutil.py
@@ -37,6 +37,7 @@ import signal
 import crypt
 import random
 import functools
+from distutils import spawn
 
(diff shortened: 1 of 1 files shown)


rvykydal:
Created a PR also for blivet: https://github.com/rhinstaller/blivet/pull/572

PR #1014, opened 2017-04-06
Catch exception when reading from in-memory connection being removed …

rvykydal:
…(#1439051)

This can happen when a connection is removed by NM between we obtain connection
objects and try to read settings from them.

### DIFF
--- a/pyanaconda/nm.py
+++ b/pyanaconda/nm.py
@@ -757,7 +757,12 @@ def nm_get_all_settings():
     connections = proxy.ListConnections()
     for con in connections:
         proxy = _get_proxy(object_path=con, interface_name="org.freedesktop.NetworkManager.Settings.Connection")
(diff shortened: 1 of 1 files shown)


pyanaconda/nm.py
```diff
@@ -757,7 +757,11 @@ def nm_get_all_settings():
     connections = proxy.ListConnections()
     for con in connections:
         proxy = _get_proxy(object_path=con, interface_name="org.freedesktop.NetworkManager.Settings.Connection")
-        settings = proxy.GetSettings()
+        try:
+            settings = proxy.GetSettings()
+        except Exception as e:
+            log.debug("nm_get_all_settings: %s", e)
+            continue
```
M4rtinK: I think it would good to add a note why we are doing this - basically the content of the commit message. :)

PR #1018, opened 2017-04-07
Fix a property name of luks devices in storage checking (#1439411)

poncovka:
Changed hasKey to has_key in verify_luks_devices_have_key.

Resolves: rhbz#1439411


M4rtinK:
@jkonecny12 This was actually a backporting issue - Blivet on RHEL7 still uses the old names.

PR #1020, opened 2017-04-07
Pylint 1.6 fixes

M4rtinK:
A bunch of fixes needed to make the CI pass with Pylint 1.6. Looks like it might actually have found some real errors. :)

### DIFF
--- a/pyanaconda/ui/gui/__init__.py
+++ b/pyanaconda/ui/gui/__init__.py
@@ -783,7 +783,7 @@ class GraphicalUserInterface(UserInterface):
         return obj
 
     def run(self):
-        (success, args) = Gtk.init_check(None)
+        (success, _args) = Gtk.init_check(None)
         if not success:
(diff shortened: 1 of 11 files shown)


pyanaconda/ui/gui/__init__.py
```diff
@@ -783,7 +783,7 @@ def _instantiateAction(self, actionClass):
         return obj
 
     def run(self):
-        (success, args) = Gtk.init_check(None)
+        (success, _args) = Gtk.init_check(None)
```
poncovka: Since `_args` is not used anywhere else, maybe there could be `success, _ = Gtk.init_check(None)`.

M4rtinK: I'm afraid that could easily get mixed-up with the `_()` localization function. Also, even though the second part of the Gtk.init_check() output is currently not used it could still be useful to know what it contains.

PR #1021, opened 2017-04-10
Support --nohome, --noswap and --noboot options in autopartitioning (#1220866)

poncovka:
**Requires a new version of pykickstart!**

Default partitioning can be modified in an install class's method
customizeDefaultPartitioning, where can be applied selections made in
a kickstart file. 
    
The base install class will remove a request for a /home partition,
if --nohome option is selected in a kickstart file. In that case,
the /home partition will not be created.
    
If --noboot option is selected, then no /boot partition will be
created with autopartitioning.
    
If --noswap option is selected, then no swap partition will be
created with autopartitioning. Also the warning about the missing
swap partition will be suppressed in this case.

Resolves: rhbz#1220866

### DIFF
--- a/pyanaconda/installclass.py
+++ b/pyanaconda/installclass.py
@@ -127,15 +127,30 @@ class BaseInstallClass(object):
     def customizeDefaultPartitioning(self, storage, data):
         # Customize the default partitioning with kickstart data.
         skipped_mountpoints = set()
+        skipped_fstypes = set()
 
-        # Create a set of mountpoints to remove from autorequests.
-        # Add /home to the set if --nohome is selected.
-        if data.autopart.autopart and data.autopart.nohome:
-            skipped_mountpoints.add("/home")
+        # Create sets of mountpoints and fstypes to remove from autorequests.
+        if data.autopart.autopart:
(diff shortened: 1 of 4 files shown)


pyanaconda/installclass.py
```diff
@@ -124,13 +124,44 @@ def setDefaultPartitioning(self, storage):
...
+            if data.autopart.noboot:
+                skipped_mountpoints.add("/boot")
+
+            # Remove swap if --noswap is selected.
+            if data.autopart.noswap:
+                skipped_fstypes.add("swap")
+
+                # Swap will not be recommended by the storage checker.
+                from pyanaconda.storage_utils import storage_checker
+                storage_checker.add_constraint("swap_is_recommended", False)
```
jkonecny12: Can we use constants instead of a string?

jkonecny12:
As I see, you are fixing nosetests by this PR. So I'm removing the nosetest fix from my PR:
https://github.com/rhinstaller/anaconda/pull/1027

Thanks for that!

poncovka:
Jenkins, test this please

PR #1024, opened 2017-04-12
network tui: fix changing ipv4 config from static to dhcp (#1432886)

rvykydal:
- Clear static ip settings when changing the configuration
  (also for ipv6).
- Log the change.
- Allow removing of nameserver setting (empty string).

Resolves: rhbz#1432886


M4rtinK:
jenkins, test this please

PR #1031, opened 2017-04-20
Make geolocation with kickstart possible (#1358331)

M4rtinK:
Make it possible to use geolocation during a kickstart based
installation. Geolocation is disabled by default during a kickstart
installation to make it more deterministic and independent on geographic
location of the machine.

To enable geolocation during a kickstart installation, use the
inst.geoloc-use-with-ks=1 option (or the --geoloc-use-with-ks=1 option
when running Anaconda as an application from the command line).

There is also an API that install classes can use to enable geolocation
during kickstart installations - the use_geolocation_with_kickstart
attribute.

Resolves: rhbz#1358331

### DIFF
--- a/pyanaconda/geoloc.py
+++ b/pyanaconda/geoloc.py
@@ -115,13 +115,82 @@ slog = logging.getLogger("sensitive-info")
 from pyanaconda import constants
 from pyanaconda.threads import AnacondaThread, threadMgr
 from pyanaconda.timezone import get_preferred_timezone, is_valid_timezone
+from pyanaconda.flags import flags
 
+geolocation_enabled = False
 location_info_instance = None
 refresh_condition = threading.Condition()
 refresh_in_progress = False
 
 
-def init_geolocation(provider_id=constants.GEOLOC_DEFAULT_PROVIDER):
+def set_geolocation_enabled(options_override, install_class_override):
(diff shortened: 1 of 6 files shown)


pyanaconda/geoloc.py
```diff
@@ -115,13 +115,85 @@
...
+    Also the geoloc boot/CLI option can be used to make sure geolocation
+    will not be used during an installation, like this:
+
+    inst.geoloc=0
+
+    :returns: if geolocation can be used during this installation run
+    :rtype: bool
+    """
+
+    # check if geolocation should be enabled for this type of installation
```
jkonecny12: I'm missing the condition here.

pyanaconda/geoloc.py
```diff
@@ -115,13 +115,85 @@
...
+    use_geolocation = True
+
+    # don't use geolocation during image and directory installation
+    if flags.imageInstall or flags.dirInstall:
+        log.info("Geolocation is disabled for image or directory installation.")
+        use_geolocation = False
+
```
jkonecny12: Remove this blank line. `if` and `else` here are connected.

rvykydal:
Ported to master with https://github.com/rhinstaller/anaconda/pull/1101

PR #1033, opened 2017-04-21
F26 dev backport fix payload reset on same repo

jkonecny12:
Backport of RHEL7 PR: https://github.com/rhinstaller/anaconda/pull/945 with small optimization.

-----------------------------------

This problem is solved by downloading content of `repomd.xml` file and comparing its MD5 hashes with downloaded repomd.xml file from the old repo url.

MD5 hash is used here because it takes less memory and we don't need `repomd.xml` content for anything else.

The `postSetup()` method was added to the payload. This method can be overridden by `Payload` child classes which needs to do additional tasks at the end of the `restartThread` call from the `PayloadManager` class.

Second commit is to make docstrings consistent in `Payload` and `DNFPayload`.

### DIFF
--- a/pyanaconda/payload/__init__.py
+++ b/pyanaconda/payload/__init__.py
@@ -307,6 +307,19 @@ class Payload(object):
         if repo:
             repo.enabled = False
 
+    def verifyAvailableRepositories(self):
+        """Verify availability of existing repositories.
+
+        This method tests if URL links from active repositories can be reached.
+        It is useful when network settings is changed so that we can verify if repositories
+        are still reachable.
+
+        This method should be overriden.
+        """
+        log.debug("Install method %s is not able to verify availability",
+                  self.__class__.__name__)
+        return False
+
     ###
     ### METHODS FOR WORKING WITH GROUPS
     ###
@@ -815,6 +828,14 @@ class PackagePayload(Payload):
                 log.warning("Could not add %s to groups, not a list.", groupid)
         elif groupid:
             log.warning("Platform group %s not available.", groupid)
+
+    def postSetup(self):
+        """Run specific payload post-configuration tasks on the end of
+        the restartThread call.
+
+        This method could be overriden.
+        """
+        pass
 
     @property
     def kernelPackages(self):
@@ -1418,6 +1439,9 @@ class PayloadManager(object):
             self._setState(self.STATE_ERROR)
             payload.unsetup()
             return
+
+        # run payload specific post configuration tasks
+        payload.postSetup()
 
         self._setState(self.STATE_FINISHED)
 
--- a/pyanaconda/payload/__init__.py
+++ b/pyanaconda/payload/__init__.py
@@ -77,11 +77,13 @@ from distutils.version import LooseVersion
 REPO_NOT_SET = False
 MAX_TREEINFO_DOWNLOAD_RETRIES = 6
 
+
 def versionCmp(v1, v2):
-    """ Compare two version number strings. """
+    """Compare two version number strings."""
     firstVersion = LooseVersion(v1)
     secondVersion = LooseVersion(v2)
     return (firstVersion > secondVersion) - (firstVersion < secondVersion)
+
 
 ###
 ### ERROR HANDLING
@@ -89,21 +91,27 @@ def versionCmp(v1, v2):
 class PayloadError(Exception):
     pass
 
+
 class MetadataError(PayloadError):
     pass
 
+
 class NoNetworkError(PayloadError):
     pass
+
 
 # setup
 class PayloadSetupError(PayloadError):
     pass
 
+
 class ImageMissingError(PayloadSetupError):
     pass
 
+
 class ImageDirectoryMountError(PayloadSetupError):
     pass
+
 
 # software selection
 class NoSuchGroup(PayloadError):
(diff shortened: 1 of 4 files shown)


pyanaconda/payload/__init__.py
```diff
@@ -307,6 +307,19 @@ def disableRepo(self, repo_id):
         if repo:
             repo.enabled = False
 
+    def verifyAvailableRepositories(self):
+        """Verify availability of existing repositories.
+
+        This method will test if repositories can be reached by URL from
+        actual repositories. It is useful when network settings is changed
```
M4rtinK: > This method will test if repositories can be reached by URL from actual repositories.
I don't think I understand the meaning of this.

pyanaconda/payload/dnfpayload.py
```diff
@@ -1100,6 +1119,15 @@ def _writeDNFRepo(self, repo, repo_path):
             if ks_repo.excludepkgs:
                 f.write("exclude=%s\n" % ",".join(ks_repo.excludepkgs))
 
+    def postSetup(self):
+        """Save repomd.xml files for later test for availability."""
```
M4rtinK: But we are saving just the hash, right ?

pyanaconda/payload/dnfpayload.py
```diff
@@ -1125,3 +1153,74 @@ def postInstall(self):
...
+    This class can test availability of this repository by comparing hashes.
+    """
+    def __init__(self, dnf_payload, repo):
+        self._repoId = repo.id
+        self._method = dnf_payload.data.method.method
+        self._urls = repo.baseurl
+        self._repomd_hash = ""
+
+    @property
+    def repoMDHash(self):
```
M4rtinK: I think we might want to use underscores in the method names of new code F26/master branch.

pyanaconda/payload/dnfpayload.py
```diff
@@ -1125,3 +1153,74 @@ def postInstall(self):
...
+                         proxy_url, e)
+
+        session = requests_session()
+
+        # Test all urls for this repo. If any of these is working it is enough.
+        for url in self._urls:
+            try:
+                result = session.get("%s/repodata/repomd.xml" % url, headers=headers,
+                                     proxies=proxies, verify=sslverify)
+                # Server returned HTTP 4XX or 5XX codes
```
M4rtinK: I think this should be above the else ?

pyanaconda/ui/gui/spokes/network.py
```diff
@@ -1521,11 +1521,14 @@ def apply(self):
                 log.debug("network spoke (apply) refresh payload")
 
                 from pyanaconda.payload import payloadMgr
-                if not self.payload.verifyAvailableRepositories():
+                if payloadMgr.running:
+                    log.debug("Payload is in the process of restarting, skip the repo availability check.")
```
M4rtinK: What if network settings changed since the payload restart has been triggered ? I guess the chance is small, but I guess we might get false results if the network using part of the check did run before network configuration has been changed. That could be solved by scheduling an extra check, but I'm not sure if that's not an overkill.

pyanaconda/payload/dnfpayload.py
jkonecny12: True, I've missed this because of backporting. Thanks!

pyanaconda/payload/__init__.py
```diff
@@ -113,24 +121,27 @@ def __init__(self, group, adding=True, required=False):
...
     pass
 
+
 class Payload(object):
-    """ Payload is an abstract class for OS install delivery methods. """
+    """Payload is an abstract class for OS install delivery methods."""
     def __init__(self, data):
-        """ data is a kickstart.AnacondaKSHandler class
-        """
+        """Data is a kickstart.AnacondaKSHandler class."""
```
poncovka: Maybe this information could be in the `:param` and `:type` format.

pyanaconda/payload/__init__.py
```diff
@@ -145,31 +156,31 @@ def __init__(self, data):
...
-        """ Invalidate a previously setup paylaod. """
+        """Invalidate a previously setup payload."""
         self.storage = None
         self.instclass = None
 
     def preStorage(self):
-        """ Do any payload-specific work necessary before writing the storage
-            configuration.  This method need not be provided by all payloads.
+        """Do any payload-specific work necessary before writing the storage
+        configuration.  This method need not be provided by all payloads.
```
poncovka: There are also extra spaces between sentences in this file.

pyanaconda/payload/dnfpayload.py
jkonecny12: Rephrased, I hope it's ok now.

pyanaconda/payload/dnfpayload.py
jkonecny12: This shouldn't be here at all, it is leftover.

pyanaconda/ui/gui/spokes/network.py
jkonecny12: This problem is more general. It could happen even without this changes, because this part in `payload/__init__.py`.
```
# If a restart thread is already running, don't start a new one 
if threadMgr.get(THREAD_PAYLOAD_RESTART):
    return
```
So even without my test, it won't be able to run payload thread again and as a bonus it just silently skip restart of the payload.

Main question is, if it is a problem. I hope that DNF is able to somehow react on change of connectivity in a middle of this process. However if we want to solve this potential issue we can for example block network spoke in the middle of restarting.

So or so I think it shouldn't be part of this PR.

pyanaconda/payload/__init__.py
jkonecny12: Good point, I've just did the formatting but I didn't read the comments. Thanks I'll fix it.

pyanaconda/payload/__init__.py
jkonecny12: This is on purpose. I don't know exactly why clumes is using double space everywhere.

jkonecny12:
**UPDATED**

jkonecny12:
@poncovka @M4rtinK Is it fine now?

poncovka:
Looks good, thanks!

PR #1035, opened 2017-04-21
Tweak epoch definition to fix system clock setting (#1433560)

AdamWill:
This approach was proposed by rmarshall, I am just submitting
it as he didn't get around to it yet. I have tested this and
verified it works: the correct time is calculated and set even
when Python's concept of the 'current' timezone isn't the same
as the one the function is called with.

Proposed for F26 and Rawhide.

### DIFF
--- a/pyanaconda/isys/__init__.py
+++ b/pyanaconda/isys/__init__.py
@@ -114,9 +114,10 @@ def set_system_date_time(year=None, month=None, day=None, hour=None, minute=None
 
     """
 
+    utc = pytz.UTC
     # If no timezone is set, use UTC
     if not tz:
-        tz = pytz.UTC
+        tz = utc
 
     # get the right values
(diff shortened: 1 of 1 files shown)


jkonecny12:
Jenkins, it's ok to test.

AdamWill:
See my comment on #1036 for an explanation of the tweak. https://paste.fedoraproject.org/paste/TDHnLi10BXgy6XxXnMBg4V5M1UNdIGYhyRLivL9gydE=/ is a test script I adapted from @jkonecny12 's test (thanks for catching this).

AdamWill:
The error is basically a bug in pylint and has been reported as such:

https://github.com/PyCQA/pylint/issues/1104

it's to do with how `pytz.UTC` is implemented, I think. I'll add a pylint-disable comment.

PR #1037, opened 2017-04-24
Show or hide the content of the expander (#1441733)

poncovka:
When the expander is expanded or collapsed and we process the
`activate` signal, we should manually show or hide the content
of the expander. Otherwise, the content can be invisible but
still sensitive to user actions.

Resolves: rhbz#1441733


poncovka:
@jkonecny12, you are right. I have looked into this a little more and found a reported bug: https://bugzilla.gnome.org/show_bug.cgi?id=776937

I've added a comment about that to the code and the commit message.

PR #1038, opened 2017-04-25
network: handle multiple connections for one device better (#1444887)

rvykydal:
Resolves: rhbz#1444887

The connection we are working with (eg modifying in UI or based on definition
in %pre kickstart command) should be the persistent one (corresponding to ifcfg
files).  Not the connection generated for configuration set up in initramfs.

This should fix also #1433891.

Some details of the changes:

- nm.py: Don't assume having only single connection when reading or updating settings for given device - don't silently use the first connetction from a list but instead raise an exception. Add a new function to update setting specified by uuid to be used in cases where having multiple connections for a device is valid, when there is 1) active connection from initramfs and 2) ifcfg configuration/connection from kickstart which has not been activated in installer (--activate / --no-activate options))

- ensure_single_initramfs_connection: Be more robust ensuring that the ifcfg connection is used during installation if there is another connection generated by NM upon start as a result of failing matching of configuration from initramfs.

- dumpMissingdefaultIfcfgs: when dumping missing ifcfgs filter out the devices before handling irelevant exceptions

- update_settings_with_ksdata: handle multiple connections correctly - update the persistent (ifcfg) connection (the 2) from above)). This function is used when modifying configuration in TUI or via %pre kickstart section.

- setOnboot: handle multiple connections correctly as well - modify the ifcfg connection

- network gui spoke: expose the ifcfg connection in ui ... kickstart tests PR: https://github.com/rhinstaller/kickstart-tests/pull/84

- network tui spoke: expose the ifcfg connection in ui

- catch log the MultipleSettingsFoundError exception in other cases (athough they should not occur)


pyanaconda/nm.py
```diff
@@ -961,6 +969,38 @@ def nm_delete_connection(uuid):
...
+                          value: new value
+                          default_type_str: dbus type of new value to be used
+                                            if the setting does not already exist;
+                                            if None, the type is inferred from
+                                            value (string and bool only)
+       :type new_values: [[key1, key2, value, default_type_str], ...]
+                         key1: str
+                         key2: str
+                         value:
+                         default_type_str: str
```
jkonecny12: I'm not sure if this will be build correctly as the documentation. It would be nice to look on it before pushing the changes to avoid non-readable documentation.

Another thing is, wouldn't it be better to create object which will contain `key1`, `key2`, `value` and `default_type_str`. It could also handle validation and it will be easy to understand what to pass where.
I'm not sure if this is a right step now, because I think this is used somewhere else already. However it could be worth the effort even for future changes.

pyanaconda/nm.py
```diff
@@ -751,7 +751,11 @@ def _find_settings(value, key1, key2, format_value=lambda x:x):
     connections = proxy.ListConnections()
     for con in connections:
         proxy = _get_proxy(object_path=con, interface_name="org.freedesktop.NetworkManager.Settings.Connection")
-        settings = proxy.GetSettings()
+        try:
+            settings = proxy.GetSettings()
+        except GLib.GError as e:
+            log.debug("_find_settings: %s", e)
```
jkonecny12: Please append prefix "Exception raised in ...".

pyanaconda/nm.py
```diff
@@ -784,7 +788,11 @@ def nm_get_all_settings():
     connections = proxy.ListConnections()
     for con in connections:
         proxy = _get_proxy(object_path=con, interface_name="org.freedesktop.NetworkManager.Settings.Connection")
-        settings = proxy.GetSettings()
+        try:
+            settings = proxy.GetSettings()
+        except GLib.GError as e:
+            log.debug("nm_get_all_settings: %s", e)
```
jkonecny12: Same here.

pyanaconda/nm.py
rvykydal: For a debug message, to me it doesn't make much difference, but I can change it.

pyanaconda/nm.py
```diff
@@ -961,6 +969,38 @@ def nm_delete_connection(uuid):
...
+                                            if None, the type is inferred from
+                                            value (string and bool only)
+       :type new_values: [[key1, key2, value, default_type_str], ...]
+                         key1: str
+                         key2: str
+                         value:
+                         default_type_str: str
+       :raise SettingsNotFoundError: if settings were not found
+                                           (eg for "wlan0")
+    """
```
rvykydal: will do

pyanaconda/nm.py
rvykydal: AFAIK there is no such a place to look at, or I don't know of any (the generated documentation).
But I agree that we could perhaps use NamedTuple here.

pyanaconda/network.py
```diff
@@ -272,6 +272,55 @@ def unset(self, *args):
...
+    for dev_name in nm.nm_devices():
+        try:
+            nm.nm_device_setting_value(dev_name, "connection", "uuid")
+        except nm.SettingsNotFoundError:
+            pass
+        except nm.MultipleSettingsFoundError:
+            if nm.nm_device_is_slave(dev_name):
+                continue
+            ifcfg_path = find_ifcfg_file_of_device(dev_name)
+            if ifcfg_path:
```
rvykydal: Hm, I'd like to keep the outer function for networkInitialize readability, but I can't come with any reasonable name for the inner function (which tells something about the nature of the special-purpose tweak we are doing here). Also maybe for better place what to include in the function.

ensure_device_has_connection_from_ifcfg_if_the_ifcfg_was_generated_from_bootoptions

The point about testability seems valid to me in general.
I'll have to think about it. And perhaps, with handling multiple devices better we don't even need the whole function at all. Especially given that it would reactivate the device unconditionally with the risk of invalidating possible existing nfs mount via the connection.

jkonecny12: Hmm, this is really hard to name. I'll try to think of some smaller name.

pyanaconda/nm.py
jkonecny12: The `namedtuple` is really nice solution. However, I think it will be better to implement this to master instead of rhel7 now.

pyanaconda/nm.py
jkonecny12: Even debug logs are better to be readable for all of us ;).

pyanaconda/network.py
vpodzime: I think there could be 'device_has_ifcfg_from_kickstart()' for this first ``if``, ``with`` and the second ``if``. And then the part below ``if not "Generated by parse-kickstart" in f.read()`` could go into a function called ``ensure_device_activated()`` or something like that.

rvykydal:
Patches updated:
- exception log message modified as proposed by jkonecny12
- code from ensure_single_initramfs_connections moved into function as requested by jkonecny12. Similarly to what vpodzime proposed I created ensure_active_ifcfg_connection_for_device which is now used also by apply_kickstart.

Also added some more logging messages in initializeNetwork so we know what is actually calling the ensure_active_ifcfg_connection_for_device (fixing initramfs connections vs. applying kickstart)
and fixed a log message in kickstart.py

jkonecny12:
Please fix this pylint error before merge. 
``
W0612:337,48: ensure_active_ifcfg_connection_for_device: Unused variable 'e'
``

rvykydal:
jenkins, test this please

PR #1040, opened 2017-04-25
make anaconda working back again with tmux2.4

pallotron:
Rawhide pulls now tmux-2.4 (https://koji.fedoraproject.org/koji/packageinfo?packageID=9438) which has modified behavior.

See below for more details:

* https://bugzilla.redhat.com/show_bug.cgi?id=1445049
* https://github.com/tmux/tmux/issues/885

image

image

image

### DIFF
--- a/data/tmux.conf
+++ b/data/tmux.conf
@@ -5,10 +5,18 @@ bind -n F1 list-keys
 
 set-option -s exit-unattached off
 set-option -g base-index 1
-set-option -g set-remain-on-exit on
+
+# tmux 2.4 has different name for this option.
+# See https://github.com/rhinstaller/anaconda/pull/1040
+if-shell '[ $(echo "$(tmux -V | cut -d" " -f2) >= 2.4" | bc) -eq 1 ]' \
+    'set-option -g remain-on-exit on' \
+    'set-option -g set-remain-on-exit on'
+
 set-option -g history-limit 10000
 
-new-session -s anaconda -n main "anaconda"
+# The idea here is to detach the client started here via anaconda.service, and
+# then re-attach to it in the tmux service run on the console tty.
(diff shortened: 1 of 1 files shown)


data/tmux.conf
```diff
@@ -5,20 +5,16 @@ bind -n F1 list-keys
 
 set-option -s exit-unattached off
 set-option -g base-index 1
-set-option -g set-remain-on-exit on
```
Frodox: should it `Require: tmux >= 2.4` or not?

pallotron: yeah this is new in tmux2.4, does tmux config support conditionals? If so it might be a good idea to support both ways

pallotron: Something like this?
`if-shell` runs first statement if the condition is true, otherwise the second

```
if-shell '[ $(echo "$(tmux -V | cut -d" " -f2) >= 2.4" | bc) -eq 1 ]' \
    'set-option -g remain-on-exit on' \
    'set-option -g set-remain-on-exit on'
```

jkonecny12:
Jenkins, it's ok to test.

data/tmux.conf
jkonecny12: This looks to me like a good idea.

pallotron: tested on my laptop with tmux 2.2 and it works.
tested on a old server with tmux 1.8 and it does not (as in, it detaches on exit but I think it's because 1.8 does not support `set-remain-on-exit` :) )

pallotron: I updated this commit with the proposal.

pallotron:
some difficulties trying to rebase this into `f26-devel` it was agreed on IRC to instead land https://github.com/rhinstaller/anaconda/pull/1042
I will close this one.

PR #1041, opened 2017-04-25
Structured logging

M4rtinK:
This pull request adds support for structured logging to Anaconda and coverts all modules to use it. You can check the individual commit messages for how it works and and for examples.

Also as this PS is quite substantially changing how the logs look like (due to replacing the always present "anaconda" prefix with the structured logging prefixes), it is currently targeting master only.


M4rtinK:
This is an example how the resulting log output looks like: https://paste.fedoraproject.org/paste/qQhmZY~BLOMzpl9BWgQCgF5M1UNdIGYhyRLivL9gydE=

jkonecny12:
jenkins, test this please

M4rtinK:
jenkins, test this please

pyanaconda/anaconda_log.py
```diff
@@ -130,6 +133,25 @@ class AnacondaFileHandler(_AnacondaLogFixer, logging.FileHandler):
...
+
+    def filter(self, record):
+        record.log_prefix = ""
+        if record.name:
+            # messages going to the generic "anaconda" logger get the log prefix "misc"
+            if record.name == "anaconda":
+                record.log_prefix = "misc"
+            else:
+                # drop "anaconda." from the log prefix
+                record.log_prefix = re.sub(REMOVE_ANACONDA_PREFIX_RE, "", record.name)
```
M4rtinK: That's a good idea - and even though using a constant like this is slightly terrifying, this part of the code is indeed potentially quite performance sensitive & we are not likely to change the number any time soon. :)

jaymzh:
:feelsgood: 😀 

CC @dcavalca125

pyanaconda/kickstart.py
```diff
@@ -80,12 +80,32 @@
...
+anaconda_log = logging.getLogger("anaconda")
+script_log = logging.getLogger("anaconda.kickstart.script")
+parsing_log = logging.getLogger("anaconda.kickstart.parsing")
 stderrLog = logging.getLogger("anaconda.stderr")
 storage_log = logging.getLogger("blivet")
 stdoutLog = logging.getLogger("anaconda.stdout")
 from pyanaconda import anaconda_log
 
+# command specififc loggers
+authconfig_log = logging.getLogger("anaconda.kickstart.authconfig")
```
poncovka: Is it really necessary to create a new logger for every kickstart command? Isn't there a way, how to add these prefixes, when we log the messages? We could use `log.debug("Message about a command xyz", extra={"prefix": "xyz"})` or modify the logger to be able to do something like: `log.debug_with_prefix("xyz", "Message about a command xyz")`

Example with extra arguments is here: https://docs.python.org/3/library/logging.html#logging.debug

pyanaconda/addons.py
```diff
@@ -27,7 +27,7 @@
 from pyanaconda.progress import progress_message
 from pyanaconda.i18n import N_
 
-log = logging.getLogger("anaconda")
+log = logging.getLogger("anaconda.addons")
```
poncovka: Have you experimented with `%(module)s`? 
https://docs.python.org/3/library/logging.html#logrecord-attributes

pyanaconda/anaconda_log.py
```diff
@@ -32,6 +32,10 @@
 STDOUT_FORMAT = "%(asctime)s %(message)s"
 DATE_FORMAT = "%H:%M:%S"
 
+# the Anaconda log uses structured logging
+ANACONDA_ENTRY_FORMAT = "%(asctime)s,%(msecs)03d %(levelname)s %(log_prefix)s: %(message)s"
```
poncovka: Couldn't we have a boot option `inst.debug`, that would enable more detailed structured messages? 
It looks like there are a lot of attributes we could log: 
https://docs.python.org/3/library/logging.html#logrecord-attributes

pyanaconda/anaconda_log.py
```diff
@@ -130,6 +137,25 @@ class AnacondaFileHandler(_AnacondaLogFixer, logging.FileHandler):
...
+    general (or miss-directed) log message, set the log_prefix to "misc".
+    """
+
+    def filter(self, record):
+        record.log_prefix = ""
+        if record.name:
+            # messages going to the generic "anaconda" logger get the log prefix "misc"
+            if record.name == "anaconda":
+                record.log_prefix = "misc"
+            elif record.name.startswith("anaconda."):
```
poncovka: I have looked at the logs and I miss the `anaconda` prefix a little :-) But I can get used to it.

pyanaconda/anaconda_log.py
```diff
@@ -130,6 +137,25 @@ class AnacondaFileHandler(_AnacondaLogFixer, logging.FileHandler):
 class AnacondaStreamHandler(_AnacondaLogFixer, logging.StreamHandler):
     pass
 
+class AanacondaPrefixFilter(logging.Filter):
```
M4rtinK: Fixed, thanks! :)

pyanaconda/kickstart.py
jkonecny12: Or create the prefix automatically. Something like use class.__name__ or similar. 
Or implement create new "base class" with `log_prefix` variable which will be inherited by these kickstart classes and they just fill the `log_prefix` variable, everything else will be done inside of the "base class".

pyanaconda/anaconda_log.py
M4rtinK: Yeah, there is a lot of stuff to log - and especially stuff like path to file where the log message originated from or the line number could certainly be useful for debugging. But at the same time I'm not sure how readable the log file would be in such a case.

So I'm thinking about two things that could be done:
- we can do a detailed/verbose (`anaconda-verbose.log`, etc.) log that is created if inst.debug=1 is used, which h could contain path to file, function name and line number
  - like this we still have a readable `anaconda.log` for general debugging and the new detailed log for detailed debugging
- if we switched from using the SyslongHandler to actual Journal handler for forwarding of Anaconda log messages to journal, all the extra data fields should end up in Journal as well
  - we might need to change our log grabbing procedures to either grab the full journal or some expert that has the extra fields
  - we might also want to investigate how journalctl can present the extra fields in it's output, as just displaying all of them at once would make the log mostly unreadable
  - I guess that the extra data fields can be useful for kickstart tests and other forms of testing

pyanaconda/kickstart.py
M4rtinK: That sounds good - and it indeed looks like our kickstart command subclasses look like a good use case for the class instance log (eq. just call self.log.info() and the appropriate prefix will be added automatically based on the class). But as we subclass the kickstart commands in this case, it will probably have to a mixin that adds the logging functionality to the given kickstart command class.

BTW, a potential show-stopper - looks like the instantiated kickstart command classes for some reason get serialized when a storage snapshot is being created and I've seen it die when it tried to serialize a logger instance (eq. self.log = getLogger("kickstart.timezone") was enough to make it crash during serialization). It might not be an issue if self.log is is a proper property, or the serialization could be fixed to skip the kickstart command instances, but it's certainly something to watch for.

For that reason I would like to be careful and merge this method first (statically defined loggers on top of file) before investigating using the class instance loggers as there is less chance of immediate breakage like this.

pyanaconda/addons.py
M4rtinK: I thought about just basing the prefix on folder structure (which I think is basically what the module property contains), but then decided against that, for a couple reasons:
- the prefix would change if we move the file around, making it less stable
- manually specifying the prefix is simple (a single line per module) and provides greater flexibility
  - related modules can be grouped together with a single prefix without needlessly exposing filesystem details (eq. all non spoke & hub modules can just live in the ui.gui.* namespace while spokes log to ui.gui.spoke.* and hubs log to ui.gui.hub.*)
  - as the prefix is not based on folder structure a more meaningful & better readable prefix can be used
    - `ui.gui.spoke.network` instead of `ui.gui.spokes.network` - the log messages come from **the** network spoke, not one of the network spokes
    - `ui.gui.spoke.custom_storage_config` instead of `ui.gui.spokes.custom` - even though anaconda developers, contributors and the QA community probably know that "custom spoke" is actually the "custom storage configuration spoke", not everyone might know that, especially when reading a log message prefix
- it should be pretty easy to track message origin by grepping for the prefix in the Anaconda source code (the module based approach would have a similar benefit though)
- there might actually be more than one logger in a file, if the functionality contained in it is disparate enough (as for example in the kickstart module)

In short - I want the log prefix to understandably describe what is being logged and to be reasonably stable over time.

poncovka: I have done some research as well and I agree.

pyanaconda/anaconda_log.py
poncovka: I think that we should at least investigate the second option, but we can do that in a different issue :-)

pyanaconda/kickstart.py
poncovka: I like the statically defined loggers on top of a file per module, but I don't like the loggers for kickstart commands, because I think we should treat them the same way as the class loggers (in the future). Then the `self.log` might be a way to go. Good point about the serialization ;-)

pyanaconda/anaconda.py
```diff
@@ -31,9 +31,11 @@
 from pyanaconda import addons
 
 import logging
-log = logging.getLogger("anaconda")
 stdoutLog = logging.getLogger("anaconda.stdout")
 
+from pyanaconda.ilog import get_logger
```
poncovka: Why is the submodule called `ilog`? I know we had problems with name collisions between anaconda submodules and python modules, but that should be fixed now. Same for `ihelp`, `iutil` and `isignal`. Is there another reason for these names?

pyanaconda/network.py
```diff
@@ -34,6 +34,7 @@
 from uuid import uuid4
 import itertools
 import glob
+import logging
```
poncovka: This is needed in the function `setup_ifcfg_log`. Is the function fine as it is?

pyanaconda/startup_utils.py
```diff
@@ -20,9 +20,11 @@
 from pyanaconda.i18n import _
 
 import logging
-log = logging.getLogger("anaconda")
 stdout_log = logging.getLogger("anaconda.stdout")
```
poncovka: Isn't it a little confusing that we combine `logging` with `anaconda.ilog`?

pyanaconda/ilog.py
```diff
@@ -0,0 +1,40 @@
...
+# A: To avoid import cycles. The anaconda_log module would be a natural
+#    place for this function, but it unfrotunatelly imports the flags module.
+#    And the flags module would have to import anaconda_log to obtain the
+#    get_logger() function, resulting in an import cycle.
+#    We should be able to avoid this be placing the get_logger() function
+#    in its own module that does not import any Anaconda modules.
+
+import logging
+
+def get_logger(module_name):
```
poncovka: I guess that this could be done also in the `__init__` method of the custom `Logger` class, that could be set as default with `logging.setLoggerClass`. Then we could get loggers as:

```
import logging
log = logging.getLogger(__name__)
```
Have you considered that?

pyanaconda/storage_utils.py
```diff
@@ -47,7 +47,9 @@
 from pykickstart.constants import AUTOPART_TYPE_LVM, AUTOPART_TYPE_LVM_THINP
 
 import logging
```
poncovka: It would be nice to have all the logging logic in one file, so we wouldn't need to import the `logging` anymore.

pyanaconda/anaconda.py
M4rtinK: It's all due to naming collisions as far as I can tell (the i likely means Installer):

- ilog - we mostly call the logger we use "log", so a module name should not collide with it
- ihelp - there is built-in Python function called "help" and we should not shadow that
- isys - there is the standard library "sys" module and isys was probably originally meant as an installer specific extension of the sys module
- iutil - could be that there used to be a util module somewhere in the path so iutil was used to avoid naming collisions
- isignal - there is a module called signal in the Python standard library

So it's basically a sort of a custom - add in in front of stuff that would shadow existing modules and built-in functions. But that does not necessarily mean it's correct and could not be improved (using a different name that does not collide, using installer_ prefix instead of just i, etc.).

Still I think that should probably be (with the possible exception of the ilog module) a separate pull request as we already change quite a bit in this one.

pyanaconda/network.py
M4rtinK: It should be - it sets-up its own separate logger & log file for ficfg related logging, so I don't really see a need to change it.

pyanaconda/anaconda.py
poncovka: Ok, thanks for the explanation!

pyanaconda/storage_utils.py
M4rtinK: Should be done now - the anaconda_loggers module now has getters for the module loggers and for all the specialized loggers.

anaconda.py
```diff
@@ -291,16 +291,16 @@ def start_debugger(signum, frame):
         flags.dirInstall = True
 
     # Set up logging as early as possible.
-    import logging
-    from pyanaconda import anaconda_log
-    anaconda_log.init()
-    anaconda_log.logger.setupVirtio()
+    from pyanaconda import anaconda_logging
```
jkonecny12: Why the `anaconda_logging` doesn't provide us loggers?

I have a bad feeling for typos in `anaconda_logging` and `anaconda_loggers`.

M4rtinK: Check the header of the `anaconda_loggers` module[0] - it's to avoid circular imports. The `anaconda_logging` module imports the `flags` module and the `flags` module also does logging, so it would have to import the `get_module_logger()` function from the `anaconda_logging` module, creating an import circle.

It's likely there could be other such circles, especially once we account for all the specialized loggers the `anaconda_loggers` module now provides getters for.

[0] https://github.com/rhinstaller/anaconda/pull/1041/files#diff-5c8919b8315eecb9a03a32af4f2bc02eR20

M4rtinK:
PR for Blivet to account for the anaconda_log module name change:
https://github.com/rhinstaller/blivet/pull/598

pyanaconda/ui/gui/spokes/network.py
```diff
@@ -721,7 +721,7 @@ def selected_dev_cfg(self):
         return model[itr][DEVICES_COLUMN_OBJECT]
 
     def add_dev_cfg(self, dev_cfg):
-        log.debug("network: GUI, device configuration added: connection %s device %s",
+        log.debug("GUI, device configuration added: connection %s device %s",
```
rvykydal: we can remove also "GUI, " here

pyanaconda/ui/gui/spokes/network.py
```diff
@@ -734,14 +734,14 @@ def add_device_to_list(self, device):
         if device.get_device_type() not in self.supported_device_types:
             return
         if network.is_libvirt_device(device.get_iface()):
-            log.debug("network: GUI, not adding %s", device.get_iface())
+            log.debug("GUI, not adding %s", device.get_iface())
```
rvykydal: we can remove also "GUI, " here

pyanaconda/ui/gui/spokes/network.py
```diff
@@ -734,14 +734,14 @@ def add_device_to_list(self, device):
...
             return
         # ignore fcoe vlan devices
         # (can be chopped off to IFNAMSIZ kernel limit)
         if device.get_iface().endswith(('-fcoe', '-fco', '-fc', '-f', '-')):
             return
         if network.is_ibft_configured_device(device.get_iface() or ""):
-            log.debug("network: GUI, not adding connection for device %s configured from iBFT", device.get_iface())
+            log.debug("GUI, not adding connection for device %s configured from iBFT", device.get_iface())
```
rvykydal: we can remove also "GUI, " here

pyanaconda/ui/gui/spokes/network.py
```diff
@@ -752,11 +752,11 @@ def add_device_to_list(self, device):
             if rc:
                 con_setting = rc.get_setting_connection()
                 if con_setting and con_setting.get_read_only():
-                    log.debug("network: GUI, not adding read-only connection "
+                    log.debug("GUI, not adding read-only connection "
```
rvykydal: we can remove also "GUI, " here

pyanaconda/ui/gui/spokes/network.py
```diff
@@ -752,11 +752,11 @@ def add_device_to_list(self, device):
...
                               "(assuming iBFT) for device %s", device.get_iface())
                     return
                 else:
-                    log.debug("network: GUI, can't get remote connection of active connection "
+                    log.debug("GUI, can't get remote connection of active connection "
```
rvykydal: we can remove also "GUI, " here

pyanaconda/ui/gui/spokes/network.py
```diff
@@ -766,9 +766,9 @@ def add_device_to_list(self, device):
             cons = device.get_available_connections()
             ifcfg_uuid = None
             if not cons:
-                log.debug("network: GUI, no connection when adding device %s", device.get_iface())
+                log.debug("GUI, no connection when adding device %s", device.get_iface())
```
rvykydal: we can remove also "GUI, " here

pyanaconda/ui/gui/spokes/network.py
```diff
@@ -766,9 +766,9 @@ def add_device_to_list(self, device):
...
             if len(cons) > 1:
-                log.debug("network: GUI, %s has multiple connections: %s",
+                log.debug("GUI, %s has multiple connections: %s",
```
rvykydal: we can remove also "GUI, " here

pyanaconda/ui/gui/spokes/network.py
```diff
@@ -875,7 +875,7 @@ def dev_cfg(self, uuid=None, device=None, iface=None):
     def remove_device(self, device):
         # This should not concern wifi and ethernet devices,
         # just virtual devices e.g. vpn probably
-        log.debug("network: GUI, device removed: %s", device.get_iface())
+        log.debug("GUI, device removed: %s", device.get_iface())
```
rvykydal: we can remove also "GUI, " here

anaconda.py
poncovka: I was curious what we log in `flags.py` and maybe the function `can_touch_runtime_system` could be moved somewhere else?

jkonecny12: I think this is a good idea.

M4rtinK: That sounds promising - but that's just the firs module where the import cycle blew up. So I'm not sure there are not other places - or how easy it is to create a new cycle by incident.

Still certainly worth investigating in the future though.

pallotron:
Hey, two questions: can you tune this up using cmdline arguments or kickstart config statements? Are the docs updated?

the pastebin in the comments expired, can you post a permanent one? :)

Thanks!

M4rtinK:
> Hey, two questions: can you tune this up using cmdline arguments or kickstart config statements? Are the docs updated?
There are some existing options for configuring logging via [boot options](http://anaconda-installer.readthedocs.io/en/latest/boot-options.html):
`inst.loglevel` - can set log level
`inst.syslog` - logging to a remote syslog
`inst.virtiolog` - virtio logging
`inst.nosave` - can be used to set if logs are saved to the target system

And [kickstart](http://pykickstart.readthedocs.io/en/latest/kickstart-docs.html#logging):
The `logging` command can be used to set log level & configure remote logging.

Do you have some other use cases that are not cover yet by these options/commands ?

> 
> the pastebin in the comments expired, can you post a permanent one? :)
> 
> Thanks!
Will do! :)

M4rtinK:
anaconda.log with structured logging from latest Rawhide:
https://gist.github.com/M4rtinK/1bfa0a5913a404eb42cb78c5c79b57d8

PR #1042, opened 2017-04-26
make anaconda working back again with tmux2.4

jkonecny12:
Rebase PR https://github.com/rhinstaller/anaconda/pull/1040 on `f26-devel` branch.

### DIFF
--- a/data/tmux.conf
+++ b/data/tmux.conf
@@ -5,10 +5,18 @@ bind -n F1 list-keys
 
 set-option -s exit-unattached off
 set-option -g base-index 1
-set-option -g set-remain-on-exit on
+
+# tmux 2.4 has different name for this option.
+# See https://github.com/rhinstaller/anaconda/pull/1040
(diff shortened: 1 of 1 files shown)


jkonecny12:
Merged and it is also in master now.

AdamWill:
So this doesn't work properly, for two reasons.

One, bc isn't in the installer environment.

Two, it's not *right*. bc does a floating point number comparison, not a version comparison (given the conventions of how software is normally versioned). For instance, that means it thinks 1.9 > 1.10 (because as a number, it *is*).

This:

https://stackoverflow.com/questions/4023830/how-compare-two-strings-in-dot-separated-version-format-in-bash

suggests using sort -V to do a version sort, which seems like a reasonable solution.

I'll send a modified PR and test it.

AdamWill:
https://github.com/rhinstaller/anaconda/pull/1048 should fix this properly, I believe.

pallotron:
thanks for fixing. :)

PR #1044, opened 2017-04-27
Update access to blivet fcoe singleton. (#1440134)

dwlehman:
The fcoe singleton is no longer stored as an attribute of the `Blivet` object. I should have fixed this when I made the change in blivet (for 2.0 or 2.1 IIRC). Let me know if this is the wrong target branch, of course.

### DIFF
--- a/pyanaconda/ui/gui/spokes/advstorage/fcoe.py
+++ b/pyanaconda/ui/gui/spokes/advstorage/fcoe.py
@@ -17,6 +17,8 @@
 # Red Hat, Inc.
 #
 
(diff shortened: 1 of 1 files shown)


M4rtinK:
This should go to F26 & master, right ?

dwlehman:
Sounds correct to me @M4rtinK, but you know better than I do.

poncovka:
Looks good to me. Could you also change `self.fcoe.addSan` to `self.fcoe.add_san` on the line 90? It should fix the failed tests.

PR #1045, opened 2017-04-28
Fix device_name_is_disk to fully support raid devices (#1327439)

poncovka:
**Depends on blivet:** https://github.com/rhinstaller/blivet/pull/578

The cache we use in device_name_is_disk should contain also md names
of udev devices. This fixes the case when the given device name is
actually an md name. Also log the results of device_matches for debugging.

Generated kickstart file should not contain onlyuse devices of the
ignoredisk command that are not disks, otherwise the kickstart
installation with this file will fail.
 
Resolves: rhbz#1327439

### DIFF
--- a/pyanaconda/storage_utils.py
+++ b/pyanaconda/storage_utils.py
@@ -753,7 +753,15 @@ def device_name_is_disk(device_name, devicetree=None, refresh_udev_cache=False):
                 # so we cache it in this non-elegant way.
                 # An unfortunate side effect of this is that udev devices that show up after
                 # this function is called for the first time will not be taken into account.
-                udev_device_dict_cache = {udev.device_get_name(d): d for d in udev.get_devices()}
+                udev_device_dict_cache = dict()
+
+                for d in udev.get_devices():
+                    # Add the device name to the cache.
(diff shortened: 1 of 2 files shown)


pyanaconda/storage_utils.py
```diff
@@ -753,7 +753,15 @@ def device_name_is_disk(device_name, devicetree=None, refresh_udev_cache=False):
...
                 # this function is called for the first time will not be taken into account.
-                udev_device_dict_cache = {udev.device_get_name(d): d for d in udev.get_devices()}
+                udev_device_dict_cache = dict()
+
+                for d in udev.get_devices():
+                    # Add the device name to the cache.
+                    udev_device_dict_cache[udev.device_get_name(d)] = d
+                    # If the device is md, add the md name as well.
+                    if udev.device_is_md(d) and udev.device_get_md_name(d):
+                        udev_device_dict_cache[udev.device_get_md_name(d)] = d
```
jkonecny12: Just one small enhancement, you may call the `udev.device_get_md_name(d)` just once, save it to variable and use this. 

The `device_get_md_name()` method can do many calls so it'll be better for performance this way.

PR #1047, opened 2017-04-28
Show or hide the content of the expander on Fedora

poncovka:
When the expander is expanded or collapsed and we process the
`activate` signal, we should manually show or hide the content
of the expander. Otherwise, the content can be invisible but
still sensitive to user actions.

We need to set this manually because of a gtk bug:
https://bugzilla.gnome.org/show_bug.cgi?id=776937

(cherry picked from commit 855f6bc)


poncovka:
Of course, but no response so far.

PR #1048, opened 2017-04-28
Really fix with tmux 2.4 (version comparison was busted)

AdamWill:
The attempt in 6752b131 to conditionalize the tmux config on
tmux version didn't work right, for two reasons: bc isn't in
the installer environment, and bc doesn't do the right kind of
comparison anyway (we want to compare following versioning
conventions, bc does a floating point number comparison). This
ought to do it properly, I hope.

### DIFF
--- a/data/tmux.conf
+++ b/data/tmux.conf
@@ -8,7 +8,9 @@ set-option -g base-index 1
 
 # tmux 2.4 has different name for this option.
 # See https://github.com/rhinstaller/anaconda/pull/1040
-if-shell '[ $(echo "$(tmux -V | cut -d" " -f2) >= 2.4" | bc) -eq 1 ]' \
+# This checks if the current tmux version (from tmux -V) >= 2.4, using
+# sort -V to do a version-y comparison (and -r to reverse it)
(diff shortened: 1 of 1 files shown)


AdamWill:
I've built both F26 and Rawhide images with this patch applied, and both boot successfully.

AdamWill:
@M4rtinK

AdamWill:
Ping? Rawhide installer images are all completely broken till this is fixed.

pallotron:
did you mean f26? because for me it works ok on rawhide/master

AdamWill:
No, I meant Rawhide:

https://openqa.fedoraproject.org/tests/overview?distri=fedora&version=Rawhide&build=Fedora-Rawhide-20170502.n.0&groupid=1

All the 'boot_to_anaconda' failures are caused by this bug.

The original fix would work for the currently-relevant values if you just test it from a Rawhide system where `bc` is present, but it doesn't work in an actual Rawhide installer image because `bc` is not present in that environment.

pallotron:
gotcha, well I would approve this if I had the rights :)

jkonecny12:
Jenkins, it's ok to test.

PR #1053, opened 2017-05-10
Compare empty confirmation field against input (#1438832)

M4rtinK:
Even if the confirmation filed is empty, it should still be compared
against the input field. This should help reduce typos in passwords and
passphrases and it's also the behavior people are used to from RHEL >=7.2.

Resolves: rhbz#1438832


rvykydal:
This is working in master.

PR #1057, opened 2017-05-12
Show the text of completions in the datetime spoke in Fedora

poncovka:
The text_column property of GtkEntryCompletion shouldn't be set in
a glade file. It should be set in a code with the set_text_column
method instead. Otherwise, the completion doesn't show any text.
This is a documented issue and according to gtk it is not a bug.

(cherry picked from commit e29f169)


poncovka:
Jenkins, test this please

poncovka:
Jenkins, test this please

PR #1062, opened 2017-05-17
Show warnings in datetime and source spokes (#1418604)

poncovka:
When the network time cannot be enabled or disabled, we show a
warning message and reset the switch button to the previous state.
With the reset, the handler is run for the second time and clears
are warning messages, so the user doesn't see the message at all.
Unfortunately, we cannot use handler_block_by_func, because it does
not work, so we need to block the handler manually.

If the network installation source is set, but no network is
configured, we disable the network source settings and set
a warning message. However, updateURLEntryCheck clears all
messages, so it needs to be called only if the network is
configured.

Resolves: rhbz#1418604

### DIFF
--- a/pyanaconda/ui/gui/spokes/datetime_spoke.py
+++ b/pyanaconda/ui/gui/spokes/datetime_spoke.py
@@ -406,6 +406,7 @@ class DatetimeSpoke(FirstbootSpokeMixIn, NormalSpoke):
         self._update_datetime_timer_id = None
         self._start_updating_timer_id = None
         self._tz = None
+        self._block_on_ntp_switched = False
 
     def initialize(self):
         NormalSpoke.initialize(self)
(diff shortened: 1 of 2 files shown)


pyanaconda/ui/gui/spokes/datetime_spoke.py
```diff
@@ -1062,51 +1063,71 @@ def _show_no_network_warning(self):
...
+        # Unfortunately, handler_block_by_func does not work.
+        if self._block_on_ntp_switched:
+            self._block_on_ntp_switched = False
+            return
+
+        # Clear warning messages.
+        self.clear_info()
+
+        # Try to set NTP.
         if switch.get_active():
```
jkonecny12: I have a suggestion for an small improvement. What about moving the `if`-`else` parts to a separate methods. Something like:

```
if switch.get_active():
    self._try_to_enabled_ntp(...)
else:
    self._disable_ntp(...)
```

And the blocking could be done as a reaction on the output of these functions.

poncovka: Fixed.

PR #1063, opened 2017-05-17
rpmostreepayload: Handle /var as a user-specified mountpoint

cgwalters:
See: https://github.com/ostreedev/ostree/issues/855

Basically, if `/var` is in the user specified mounts, don't override
it with the "default ostree" `var`.

This allows people to easily keep all of their local state (apart from `/etc`)
in a single partition with potentially different quotas etc.


dustymabe:
LGTM

jkonecny12:
Jenkins, it's ok to test.

jkonecny12:
Ignore the test result. They are failing because of new pylint on rawhide.

PR #1065, opened 2017-05-18
Add support for IPoIB in tui (#1366935)

rvykydal:
Resolves: rhbz#1366935

### DIFF
--- a/pyanaconda/network.py
+++ b/pyanaconda/network.py
@@ -291,9 +291,7 @@ def dumpMissingDefaultIfcfgs():
     rv = []
 
     for devname in nm.nm_devices():
-        # for each ethernet device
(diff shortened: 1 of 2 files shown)


pyanaconda/network.py
```diff
@@ -1518,7 +1520,8 @@ def status_message():
             elif len(nonslaves) > 1:
                 devlist = []
                 for devname in nonslaves:
-                    if nm.nm_device_type_is_ethernet(devname):
+                    if nm.nm_device_type_is_ethernet(devname) \
+                       or nm.nm_device_type_is_infiniband(devname):
```
jkonecny12: Too much code duplication, what about creating new method:
`nm_device_type_is_ethernet_similar()`

PR #1068, opened 2017-05-18
Add inst.waitfornet option (#1315160)

rvykydal:
inst.waitfornet=<TIMEOUT_IN_SECONDS>

At the beginning of installation (stage 2) we run a thread waiting for NM being
in connecting state (eg when configuring IP from DHCP) so that some actions
(getting NTP servers from DHCP, getting GeoIP info, downloading payload
metadata) can wait for network connectivity

This patch is adding option to wait for connectivity synchronously at the
beginning of installation (stage 2). It is addressing automatic installation
use case where we don't want to trigger user interaction when first attempt to
connect (configure IP via dhcp) fails but following attempts would succeed.

Resolves: rhbz#1315160


anaconda.py
```diff
@@ -577,9 +577,26 @@ def _earlyExceptionHandler(ty, value, traceback):
...
+        else:
+            if timeout > 0:
+                log.info("network: waiting for network connection (timeout=%d)",
+                         timeout)
+                wait_for_connected_NM(timeout=timeout)
+            else:
+                log.warning("invalid inst.waitfornet option value: %s",
+                            opts.waitfornet)
+
+    # In any case do some actions only after NM finishes its connecting.
```
jkonecny12: Could you please extract this to the outside method. I don't want to feed this Godzilla method more than it is required.

docs/boot-options.rst
```diff
@@ -303,6 +303,17 @@ Set the DHCP vendor class identifier [#dhcpd]_. Defaults to ``anaconda-$(uname -
...
 
+.. inst.waitfornet:
+
+inst.waitfornet
+^^^^^^^^^^^^^^^
+
+``inst.waitfornet=<TIMEOUT_IN_SECONDS>``
+    Wait for network connectivity at the beginning of the second stage of
+    installation (after switchroot from early initramfs stage when the installer
+    process is run)
```
jkonecny12: You are missing the period here.

pyanaconda/network.py
```diff
@@ -1433,10 +1447,14 @@ def wait_for_network_devices(devices, timeout=constants.NETWORK_CONNECTION_TIMEO
     return False
 
 def wait_for_connecting_NM_thread(ksdata):
-    """This function is called from a thread which is run at startup
-    to wait for Network Manager to connect."""
-    # connection (e.g. auto default dhcp) is activated by NM service
-    connected = _wait_for_connecting_NM()
+    """Wait for connecting NM in thread, do some work and signal connectivity
```
jkonecny12: Missing the period here.

anaconda.py
rvykydal: You are right, parsing the option does not belong here.

rvykydal:
I've updated the PR for the remarks from jkonecny12, but I'd like M4tinK to look at parsing of the value (converting to int) if I am doing it at the right place.

pyanaconda/anaconda_argparse.py
```diff
@@ -404,6 +404,23 @@ def __call__(self, parser, namespace, values, option_string=None):
...
+            else:
+                if ivalue > 0:
+                    value = ivalue
+            if value is None:
+                value = 0
+            setattr(namespace, self.dest, value)
+
+    ap.add_argument("--waitfornet", dest="waitfornet", metavar="TIMEOUT_IN_SECONDS",
+                    action=SetWaitfornet, help=help_parser.help_text("waitfornet"))
+
```
rvykydal: I am not quite sure about this, I've also tried to convert the type using type= argument but id didn't work for me.

`    def positive_int(value):
        rvalue = None.
        try:
            ivalue = int(value)
        except ValueError:
            pass
        else:
            if ivalue > 0:
                rvalue = ivalue
        if rvalue is None:
            rvalue = 0
        return rvalue

    ap.add_argument("--waitfornet", dest="waitfornet", metavar="TIMEOUT_IN_SECONDS",
                    default=0, type=positive_int, help=help_parser.help_text("waitfornet"))
`

jkonecny12: Maybe this could be helpful for you:
https://github.com/rhinstaller/pykickstart/blob/master/pykickstart/commands/firewall.py#L84

rvykydal: Yes, I am doing the same with the positive_int, it just isn't called in anaconda for some reason.

rvykydal:
jenkins, test this please

rvykydal:
jenkins, test this please

PR #1071, opened 2017-05-18
F26 devel port 74 rhbz1444887 multiple connections

rvykydal:
Note: compared to original rhel7-branch patch there is one difference which is DeviceConfiguration hunk in pyanaconda/ui/gui/spokes/network.py


pyanaconda/network.py
```diff
@@ -277,6 +277,78 @@ def unset(self, *args):
...
+                         active connection
+    :type only_replace: bool
+
+    """
+    msg = "not activating"
+    ifcfg = IfcfgFile(ifcfg_path)
+    ifcfg.read()
+    con_uuid = ifcfg.get("UUID")
+    active_con_uuid = nm.nm_device_active_con_uuid(dev_name)
+    if active_con_uuid or not only_replace:
```
jkonecny12: Shouldn't this do something even when the `only_replace` will be `True`?
From documentation I understand it this way: 

If `True`:
* apply the connection only if the device has different active connection

If `False`:
* apply the connection always

pyanaconda/network.py
```diff
@@ -277,6 +277,78 @@ def unset(self, *args):
...
+            msg = "activating"
+            try:
+                nm.nm_activate_device_connection(dev_name, con_uuid)
+            except nm.UnknownConnectionError as e:
+                log.warning("network: can't activate connection %s on %s: %s",
+                            con_uuid, dev_name, e)
+    log.debug("network: ensure active ifcfg connection for %s (%s -> %s): %s",
+               dev_name, active_con_uuid, con_uuid, msg)
+
+def ifcfg_is_from_kickstart(path):
```
jkonecny12: Please add description what is path. Where it should points to.

pyanaconda/network.py
```diff
@@ -1151,14 +1242,17 @@ def write_hostname(rootpath, ksdata, overwrite=False):
...
-                for dev in nm.nm_devices() if nm.nm_device_type_is_ethernet(dev))):
-        log.info('Disabling ipv6 on target system')
-        with open(cfgfile, "a") as f:
-            f.write("# Anaconda disabling ipv6 (noipv6 option)\n")
-            f.write("net.ipv6.conf.all.disable_ipv6=1\n")
-            f.write("net.ipv6.conf.default.disable_ipv6=1\n")
+    try:
+        if ('noipv6' in flags.cmdline
+            and all(nm.nm_device_setting_value(dev, "ipv6", "method") == "ignore"
+                    for dev in nm.nm_devices() if nm.nm_device_type_is_ethernet(dev))):
```
jkonecny12: Can you please extract this:
``all(nm.nm_device_setting_value(dev, "ipv6", "method") == "ignore" for dev in nm.nm_devices() if nm.nm_device_type_is_ethernet(dev))``

to the special method `self._test_if_ipv6_is_ignored()`.

It will be much easier to read this `if` statement and it could be on one line.

pyanaconda/network.py
```diff
@@ -277,6 +277,78 @@ def unset(self, *args):
...
+                log.error("network: multiple settings but no ifcfg for %s", dev_name)
+                continue
+
+            # Handle only ifcfgs created from boot options in initramfs
+            # (Kickstart based ifcfgs are handled in apply_kickstart)
+            if ifcfg_is_from_kickstart(ifcfg_path):
+                continue
+
+            ensure_active_ifcfg_connection_for_device(ifcfg_path, dev_name, only_replace=True)
+            rv.append(dev_name)
```
rvykydal: I've already split out ensure_active_ifcfg_connection_for_device from the function during rhel7-branch patch review. I can't see a reason for going further and having something like _handle_multiple_setting_from_initramfs().

pyanaconda/network.py
rvykydal: Hm, I think it does exactly what the documentation says.
If only_replace is True it will execute the if clause only if there is already an active connection. And in the if clause it will apply the connection it if it is different.

pyanaconda/network.py
rvykydal: I'll rename the argument to ifcfg_path so it is more clear.

pyanaconda/network.py
rvykydal: I don't want to create another function with just one caller, but yes, the code is ugly and may be unreadable, but above all,  due to the usage of "all" it requires the exception block to include the stuff it shouldn't so I will rewrite it.

pyanaconda/network.py
jkonecny12: Thanks for clarification for some reason it is hard for me to read this properly but that could be just me.

pyanaconda/network.py
jkonecny12: It is much clearer to me what it should do now. I'm fine with this.

I just want to point out the difference:

Original:
```
 if ('noipv6' in flags.cmdline
            and all(nm.nm_device_setting_value(dev, "ipv6", "method") == "ignore"
                    for dev in nm.nm_devices() if nm.nm_device_type_is_ethernet(dev))):
```

New solution:
```
    if 'noipv6' in flags.cmdline:
        for devname in nm.nm_devices():
            if nm.nm_device_type_is_ethernet(devname):
                try:
                    ipv6_method = nm.nm_device_setting_value(devname, "ipv6", "method")
                except nm.MultipleSettingsFoundError as e:
                    log.debug("network: %s when getting ipv6 method of %s", e, devname)
                    ipv6_method = None
                if ipv6_method != "ignore":
                    return
    log.info('network: disabling ipv6 on target system')
    cfgfile = os.path.normpath(rootpath + ipv6ConfFile)
    with open(cfgfile, "a") as f:
        f.write("# Anaconda disabling ipv6 (noipv6 option)\n")
        f.write("net.ipv6.conf.all.disable_ipv6=1\n")
        f.write("net.ipv6.conf.default.disable_ipv6=1\n")
```

My proposed solution:
```
if 'noipv6' in flags.cmdline:
    if self._find_all_ethernet_devices_with_ipv6_ignored():
        return
   cfgfile = os.path.normpath(rootpath + ipv6ConfFile)
    with open(cfgfile, "a") as f:
        f.write("# Anaconda disabling ipv6 (noipv6 option)\n")
        f.write("net.ipv6.conf.all.disable_ipv6=1\n")
        f.write("net.ipv6.conf.default.disable_ipv6=1\n")
```

Line numbers will be almost the same (+2 empty lines and one function declaration) the only difference is readability for others. Of course this is just my point of view.

pyanaconda/network.py
jkonecny12: Thank you, it's much clearer to me now.

pyanaconda/network.py
jkonecny12: I described my point in the comment below. It is not required so I'm approving this.

pyanaconda/network.py
rvykydal: the function in your proposed solution should be
```
    if _find_all_ethernet_devices_with_ipv6_not_ignored():
        return
```

Then why not go furhter with the function (it will never be reused in any case) and do just
```
if 'noipv6' in flags.cmdline \
    and _all_ethernet_devices_have_ipv6_ignored():
```
Which just says what the doc text above:
`"""Disable ipv6 if noipv6 boot option is set and all ethernet devices ignore ipv6"""`
which feels strange to me (although in general I prefer the code to tell what is happening rather than adding comments).

rvykydal:
jenkins, test this please

pyanaconda/network.py
jkonecny12: In  the brackets that's my point. Why to write comment (doctext in this case) when you have self-explanation code. However I understand why it feels strange to you.

PR #1072, opened 2017-05-18
Fixes for pylint 1.7

poncovka:
* We should keep names of parameters of overiden methods.
* Removed useless super delegations in methods.
* Updated false positives.
* Threading objects are not recognised as context managers.

### DIFF
--- a/pyanaconda/ui/gui/__init__.py
+++ b/pyanaconda/ui/gui/__init__.py
@@ -954,7 +954,7 @@ class GraphicalUserInterface(UserInterface):
     ###
     ### SIGNAL HANDLING METHODS
     ###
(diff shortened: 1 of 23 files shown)


pyanaconda/ui/gui/__init__.py
```diff
@@ -1075,9 +1075,9 @@ def __init__(self, lightbox_func):
 
         self._lightbox_func = lightbox_func
 
-    def mainExceptionWindow(self, text, exn_file, *args, **kwargs):
+    def mainExceptionWindow(self, text, exnFile, *args, **kwargs):
```
jkonecny12: Why the `exnFile` it is not pep8

PR #1073, opened 2017-05-19
Set the default filesystem type from a kickstart file (#1449099)

poncovka:
If the kickstart file specifies the --fstype option in the autopart
command, we should use it to set the default filesystem type. If the
type is not set in a kickstart file, we will use the specified type
from an install class or the default type from blivet.

Resolves: rhbz#1449099

### DIFF
--- a/pyanaconda/anaconda.py
+++ b/pyanaconda/anaconda.py
@@ -165,11 +165,24 @@ class Anaconda(object):
         if not self._storage:
             import blivet
             self._storage = blivet.Blivet(ksdata=self.ksdata)
-
-            if self.instClass.defaultFS:
-                self._storage.setDefaultFSType(self.instClass.defaultFS)
+            self._setDefaultFSType(self._storage)
 
         return self._storage
+
+    def _setDefaultFSType(self, storage):
+        fstype = None
+
+        # Get the default fstype from a kickstart file.
+        if self.ksdata.autopart.autopart and self.ksdata.autopart.fstype:
+            fstype = self.ksdata.autopart.fstype
+        # Or from an install class.
+        elif self.instClass.defaultFS:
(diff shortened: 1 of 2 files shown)


pyanaconda/anaconda.py
```diff
@@ -165,9 +165,19 @@ def storage(self):
...
+            if self.ksdata.autopart.autopart and self.ksdata.autopart.fstype:
+                fstype = self.ksdata.autopart.fstype
+            # Or from an install class.
+            elif self.instClass.defaultFS:
+                fstype = self.instClass.defaultFS
+
+            # Set the default fstype.
+            if fstype:
+                self._storage.setDefaultFSType(fstype)
+                self._storage.setDefaultBootFSType(fstype)
```
jkonecny12: What about creating new method `self._set_default_fstype()` and moving this to the new method?

jkonecny12: What about missing `try...catch ValueError`` which was in `kickstart.py`? It is no longer required in that case why?

poncovka: I can create a new method for that.

poncovka: We already check the value of `fstype` in `AutoPart.parse`. Also do really we want to replace the message from blivet with another one? Blivet raises `ValueError("unrecognized value %s for new default fs type" % newtype)` or `ValueError("new value %s is not valid as a default fs type" % fmt)`.

jkonecny12: Good point, you can ignore my second question 👍

jkonecny12: Good point you can ignore my second question. 👍

poncovka: Fixed.

PR #1076, opened 2017-05-23
Add support for kickstart snippets in installation image drop directory

rvykydal:
This allows to have kickstart snippets in installer image drop directory
that would be prepended or appended to user supplied kickstart file (or
interactive-defaults.ks file in case of non-kickstart installation).

The drop directories are:
/usr/share/anaconda/kickstarts-prepend.d
/usr/share/anaconda/kickstarts-append.d

The kickstart snippet files will be concatenated in this order:
1) files from kickstart-prepend.d sorted by name
2) user supplied kickstart or interactive-defaults.ks in case of non-ks install
3) files from kickstart-append.d sorted by name

Use case example: atomic installer specific ostreesetup and (other stuff, like
services) configuration

https://www.redhat.com/archives/anaconda-devel-list/2017-February/msg00000.html

Formerly interactive-defaults.ks was modified with additional atomic-specific
kickstart commands (eg ostreesetup to specify the atomic repo tree) during
atomic installer image build process, which required including the (added)
content of the file in regular kickstart which makes very bad user experience.

Using this new mechanism, atomic could just drop its kickstart snippet into
/usr/share/anaconda/kickstarts-append.d/50-atomic.ks directory and it would be
applied both to kickstart and non-kickstart installations.  The file can be added
to image either right during atomic installer build process as before, but now
also via atomic product.img if it makes sense for the atomic installer image
build process POV.

### DIFF
--- a/anaconda.spec.in
+++ b/anaconda.spec.in
@@ -246,6 +246,10 @@ find %{buildroot} -type f -name "*.la" | xargs %{__rm}
 # Create an empty directory for addons
 mkdir %{buildroot}%{_datadir}/anaconda/addons
 
+# Create an empty directory for kickstart snippets
+mkdir %{buildroot}%{_datadir}/anaconda/kickstarts-prepend.d
+mkdir %{buildroot}%{_datadir}/anaconda/kickstarts-append.d
+
 %ifarch %livearches
 desktop-file-install ---dir=%{buildroot}%{_datadir}/applications %{buildroot}%{_datadir}/applications/liveinst.desktop
 %endif
--- a/pyanaconda/constants.py
+++ b/pyanaconda/constants.py
@@ -28,6 +28,8 @@ SELINUX_DEFAULT = -1
 
 # where to look for 3rd party addons
 ADDON_PATHS = ["/usr/share/anaconda/addons"]
+KICKSTART_PREPEND_DROPDIR_PATH = "/usr/share/anaconda/kickstarts-prepend.d"
+KICKSTART_APPEND_DROPDIR_PATH = "/usr/share/anaconda/kickstarts-append.d"
 
 from pykickstart.constants import AUTOPART_TYPE_LVM
 
--- a/pyanaconda/startup_utils.py
+++ b/pyanaconda/startup_utils.py
@@ -406,6 +406,34 @@ def set_installation_method_from_anaconda_options(anaconda, ksdata):
     else:
         log.error("Unknown method: %s", anaconda.methodstr)
 
+def add_kickstart_snippets_from_image(main_ks_candidate_paths, output_ks):
+    """Add kicstart snippets from image drop-down directory into main ks"""
+    ks_snippets = []
+    for main_ks in main_ks_candidate_paths:
+        if not os.path.exists(main_ks):
+            continue
+
+        prepend_dir = constants.KICKSTART_PREPEND_DROPDIR_PATH
+        prepend_snippets = [os.path.join(prepend_dir, file)
+                         for file in sorted(os.listdir(prepend_dir))]
+        append_dir = constants.KICKSTART_APPEND_DROPDIR_PATH
+        append_snippets = [os.path.join(append_dir, file)
+                           for file in sorted(os.listdir(append_dir))]
+
+        ks_snippets.extend(prepend_snippets)
+        ks_snippets.append(main_ks)
+        ks_snippets.extend(append_snippets)
+
+        with open(output_ks, "w") as f_out:
+            for ks in ks_snippets:
+                with open(ks, "r") as f_ks:
+                    f_out.write(f_ks.read())
+
+        # Only use the first main kickstart file we find.
+        break
+
+    return ks_snippets
+
 def parse_kickstart(options, addon_paths):
     """Parse the input kickstart.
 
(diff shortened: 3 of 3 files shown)


cgwalters:
It's probably worth thinking about - are there any other use cases for this?  We have a good generic design I think, but if we had other use cases in mind it might help drive the design.  One thing we could try to do is to make the ostree and rpm cases symmetric - i.e. when pungi goes to build a DVD (i.e. ISO+RPMs), it injects

```
repo --name=Anaconda-Fedora-20170524.0 --baseurl=file:///run/install/repo
```

And we drop the logic that lives in anaconda today for this?  It feels somewhat risky but probably makes things nicer in the end.

Are there any other use cases?

cgwalters:
As a thought experiment for example - would it make sense to use this instead of the `product.img` partitioning path?  Hm, probably not, because then `autopart` couldn't be used.

cgwalters:
I guess a general problem with trying to use this more widely is that kickstart doesn't have a consistent way to undo/replace earlier verbs.  Some of them (like `ostreesetup`) have simple last-one-wins semantics which is what we want here, others are always additive (like `repo` AIUI).

cgwalters:
Coincidentally, here's another possible use case:
https://pagure.io/fedora-kickstarts/pull-request/231
Basically change the Fedora Workstation ISO to inject `rhgb quiet`, and *only* Fedora workstation, and drop the bits in `dracutSetupArgs()` which injects them.

rvykydal:
Yes, this approach is rather limited - we don't (semantically) merge kickstarts, just concatenate them which means overriding the values in most of the cases (the last wins) etc... For example services command couldn't be used in user kickstart (it will either override or be overriden by the Atomic snippet). I just wanted to provide some slightly better mechanism for adding Atomic-specific default kickstart configuration both for kicstart and non-kickstart installs. Due to its limitations I didn't mean to advertise the drop direcrory feature, and we don't have other cases to use it for. It doesn't also seem right to have the atomic specific configuration partly in installclass, partly in the kickstart snippet (services).

If I am getting you correctly, you are rising the idea of using the same mechanism for repo on DVD both for Atomic and common rpm-based DVD. I think this is worth looking into, and it could mean:

- Setting ostree repo (I'd set ostreesetup command [1]) either via DVD boot options or by autodetecting it on media (the latter is what we actually do in common installer)
  - The autodetection logic would need to be updated in Anaconda when changing the approach (installer image vs iso) though. Or can we just assume the "repo on the iso" approach? (/run/install/repo)
  - We'd need some place for other metadata for repo configuration (--osname, --remote, --ref) on the iso (we get additional metadata from .treeinfo for rpm case)
- Moving other stuff from the kickstart snippet to Atomic installclass (services) or rpmostree payload class (the %post script stuff).

[1] partner of ostreesetup in kickstart for rpm installations is url (and liveimg for installation from images) and I don't think we want unification here. repo command is for additional rpm repos.

cgwalters:
Sorry to have left this on the back burner; this is still an issue for us for sure.

Maybe one thing we could do would be some sort of special kickstart flag that says it's defining defaults?

Something like this, taking the bits that live in `setDefaultPartitioning` today:

```
%defaults
part / --max-size=50000
part /home --size=500 --grow
```

The `%defaults` bit would tell pykickstart parser that the next time it sees any verb of the same type, to disregard the earlier bits.

I don't know offhand how hard this would be to implement or how many other bits of Anaconda could use it.

cgwalters:
(If we had something like this, then `interactive-defaults.ks` could just become `defaults.ks` right?)

dustymabe:
Can we pick up the discussion on this and make progress towards resolving this issue?

cgwalters:
I'm thinking that it's probably best for now to find a place to hack in the logic to say "if there's no payload specified, and we find ostree data in /ostree/repo, pick that up automatically".

rvykydal:
Seems that the PR is not good, I am closing it and moving the discussion back to anaconda-devel-list.
https://www.redhat.com/archives/anaconda-devel-list/2017-February/msg00000.html

> Maybe one thing we could do would be some sort of special kickstart flag that says it's defining defaults?
> 
> Something like this, taking the bits that live in setDefaultPartitioning today:
> 
> %defaults
> part / --max-size=50000
> part /home --size=500 --grow
> 

So one difference compared to just prepending the snippet would be that the additional verbs from %defaults would be reset by kickstart commands. But it gets beyond the level of syntactic preprocessing (which could be easy and which I hoped to be enough) given partitioning commands make up (semantically) a group - and all the partitioning commands should be reset. Moreover these commands are mutually exclusive with autopart command (checked by pykickstart).
What you seem to give as an example (moving default/auto partitioning from installclass to kickstart %default) is actually defining autopartitioning, not partitioning so it should use a new separate command (than it wouldn't confilict with autopart as I mentioned before).

> The %defaults bit would tell pykickstart parser that the next time it sees any verb of the same type, to disregard the earlier bits.
> 

Yes, the tricky part is "verb of the same type".

> I don't know offhand how hard this would be to implement or how many other bits of Anaconda could use it.

Going to the semantic level (as opposed to some kind of preprocessing which I was trying to keep the PR at) it would be a big change. Something we should probably think hard about in scope of our more long term plans. Kickstart has always been important installer interface so we need to be careful here.

PR #1078, opened 2017-05-23
Mark post inst tools as disabled only if firstboot --disable is used …

M4rtinK:
…(#1448940)

Only set the post_install_tools_disabled flag in the user interaction
config file if the user explicitly requests it via the firstboot
--disable kickstart command.

The main difference from current behavior is that while Anaconda has
been always disabling Initial Setup during a kickstart installation
(unless firstboot --enable) has been used in the kickstart, the same
behavior might not be expected for other post installation setup tools,
such as Gnome Initial Setup, which have been always start
unconditionally up till now.

User that want to make sure no post installation setup tools
(both Initial Setup and Gnome Initial Setup) will run can use
firstboot --disable in kickstart to achieve that.

Resolves: rhbz#1448940

### DIFF
--- a/pyanaconda/kickstart.py
+++ b/pyanaconda/kickstart.py
@@ -752,9 +752,13 @@ class Firstboot(commands.firstboot.FC3_Firstboot):
 
         if self.firstboot == FIRSTBOOT_SKIP:
             action = "disable"
-            # Also tell the screen access manager, so that the fact that post installation tools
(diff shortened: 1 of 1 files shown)


rvykydal:
The code has changed substantially with modularization and the setting is working as expected in Fedora. Removing the "port to Fedora" flag.

PR #1082, opened 2017-05-24
Add support for DNF-2.5.0

jkonecny12:
Callback method `start()` in the `DownloadProgress()` reporting class has the new attribute `total_drpms`.

Without this attribute the installation will crash.


jkonecny12:
The tests are failing here because the DNF-2.5.0 is in testing now.

PR #1084, opened 2017-05-24
Fix setting errors and warnings in the StorageCheckHandler in Fedora

poncovka:
Storage spoke and custom spoke communicate errors via
StorageCheckHandler, so we need to set errors and warnings
class attributes of the handler. Otherwise, the errors from
the custom spoke are not propagated to the storage spoke.


poncovka:
Jenkins, test this please.

poncovka:
Jenkins, test this please.

PR #1085, opened 2017-05-25
Fix pylint errors

jkonecny12:
Pylint errors introduced in commit:
62e95b9b070be62921cce4809de0dd4cb048ad31


pyanaconda/ui/tui/simpleline/base.py
```diff
@@ -157,7 +157,7 @@ def _thread_input(self, queue_instance, prompt, hidden):
         """
 
         if hidden:
-            data = self.simpleline_getpass(prompt)
+            data = self.simpleline_getpass = prompt
```
poncovka: This will not work, because `self.simpleline_getpass` is a function. Why is it also a property, I have no idea.

PR #1086, opened 2017-05-25
Fix simpleline_getpass related Pylint warning

M4rtinK:
Pylint is no fun and doesn't like executable properties,
so just use an instance attribute. Bonus - less code.


M4rtinK:
jenkins, test this please

PR #1087, opened 2017-05-25
Separate blivet-daily builds in mock config

jkonecny12:
It is used from `mock-fedora` and `mock-rawhide` so include it from `mock-blivet` instead of having it hardcoded inside of these configuration files.


jkonecny12:
Rebased to fixed tests.

PR #1090, opened 2017-05-29
Make it possible to change root password set in kickstart

M4rtinK:
Make it possible to change root password set in kickstart if the current password policy allows it (--changesok).

The first patch makes it possible to enter the root password spoke when root password is set in kickstart and and password policy for the root password contains the --changesok flag.

The second and third patch fix various UI issues that otherwise show up when entering the root password spoke while a root password is set in kickstart.

### DIFF
--- a/pyanaconda/ui/gui/helpers.py
+++ b/pyanaconda/ui/gui/helpers.py
@@ -52,6 +52,8 @@ class GUIInputCheckHandler(InputCheckHandler):
         super(GUIInputCheckHandler, self).__init__()
         self._policy = None
         self._input_enabled = True
+        self._initial_input = None
+        self._initial_input_confirmation = None
 
     @property
     def policy(self):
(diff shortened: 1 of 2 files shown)


pyanaconda/ui/gui/helpers.py
```diff
@@ -124,6 +126,68 @@ def input_confirmation(self):
...
+        # so remember the initial content of the fields so that we can detect if
+        # the content has changed.
+        self.initial_input = self.input
+        self.initial_input_confirmation = self.input_confirmation
+
+    @property
+    def initial_input(self):
+        """Content of the input field when the screen was entered.
+
+        And before user could influence it's contents.
```
jkonecny12: This should be standalone not something what is connecting to the name of the docstring. In other words, please remove the `And` part or remove this completely.

pyanaconda/ui/gui/helpers.py
```diff
@@ -124,6 +126,68 @@ def input_confirmation(self):
...
+
+    @initial_input.setter
+    def initial_input(self, new_input):
+        self._initial_input = new_input
+
+    @property
+    def initial_input_confirmation(self):
+        """Content of the input confirmation field when the screen was entered.
+
+        And before user could influence it's contents.
```
jkonecny12: The same here.

pyanaconda/ui/gui/spokes/password.py
```diff
@@ -39,7 +39,12 @@ class PasswordSpoke(FirstbootSpokeMixIn, NormalSpoke, GUISpokePasswordCheckHandl
     builderObjects = ["passwordWindow"]
 
     mainWidgetName = "passwordWindow"
-    focusWidgetName = "pw"
+    # Don't focus on the password entry field by default as is some cases
+    # a password might have been set in kickstart and focusing on the password
+    # filed might clear it (& hide the "password is set" message).
+    # Also there is logic in place that should focus the password entry field if
+    # password is *not* set in kickstart.
+    focusWidgetName = "passwordWindow"
```
jkonecny12: This should be local constant.

pyanaconda/ui/gui/helpers.py
M4rtinK: Actually - with the GUIInputCheckHandler being notified about the input fields being displayed to the user it doesn't really need to provide access to the initial values - it just needs to tell the "frontend" if the fields have changed. So indeed - we can scrap the properties and just keep some private attributes to hold the initial values.

pyanaconda/ui/gui/spokes/password.py
M4rtinK: I don't think that's needed - all the other spokes do the same thing & we are on the rhel7-branch. But it might make sense to refactor these out to local constants on the master branch.

jkonecny12: Good point, can you please change this on master then. Thanks!

rvykydal:
This has been ported to rhel-8 in https://github.com/rhinstaller/anaconda/pull/2034 so port to Fedora will be handled in scope of that PR.

PR #1092, opened 2017-06-02
Show warning if swap is smaller then recommended (#1290360)

poncovka:
The storage checker should produce a warning message if the swap
space is smaller then we recommend.

Resolves: rhbz#1290360

### DIFF
--- a/pyanaconda/storage_utils.py
+++ b/pyanaconda/storage_utils.py
@@ -30,6 +30,7 @@ from blivet import udev
 from blivet.size import Size
 from blivet.errors import StorageError
 from blivet.platform import platform as _platform
+from blivet.autopart import swap_suggestion
 from blivet.devicefactory import DEVICE_TYPE_LVM
 from blivet.devicefactory import DEVICE_TYPE_LVM_THINP
 from blivet.devicefactory import DEVICE_TYPE_BTRFS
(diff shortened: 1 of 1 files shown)


pyanaconda/storage_utils.py
```diff
@@ -347,6 +348,20 @@ def verify_swap(storage, constraints, report_error, report_warning):
                                  "it will significantly improve performance "
                                  "for most installations."))
 
+    else:
+        swap_space = sum((device.size for device in swaps), Size(0))
+        disk_space = sum((device.size for device in storage.mountpoints.values()), Size(0))
+        recommended = swap_suggestion(disk_space=disk_space, quiet=True)
```
jkonecny12: Shouldn't we show warning or raise exception when `disk_space` will be `Size(0)` or is it expected behavior?

poncovka: The recommended swap size for `disk_space` 0 B is 0 B, so the check will pass. We have other checks for not defined or too small partitions, that will fail in this case.

M4rtinK:
jenkins, test this please

pyanaconda/storage_utils.py
jkonecny12: Then it is fine. Thanks for explanation.

PR #1094, opened 2017-06-05
Delete file systems used only by a given installation (#1453097)

poncovka:
If user chooses to delete all file systems which are only used by
a given installation, we will skip the file systems that are shared
between more installations.

Before that we offered to delete all systems including those
used by other installed operating systems. That could be confusing
and result in data loss. The new behaviour is closer to Fedora's.

Resolves: rhbz#1453097

### DIFF
--- a/pyanaconda/ui/gui/spokes/custom.py
+++ b/pyanaconda/ui/gui/spokes/custom.py
@@ -1846,6 +1846,7 @@ class CustomPartitioningSpoke(NormalSpoke, StorageCheckHandler):
             factory.configure()
 
         self._remove_empty_parents(device)
+        log.debug("Removed device %s.", device.name)
 
     def _show_mountpoint(self, page=None, mountpoint=None):
         if not self._initialized:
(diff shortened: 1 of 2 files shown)


M4rtinK:
jenkins, test this please

pyanaconda/ui/gui/spokes/custom.py
```diff
@@ -1871,6 +1872,20 @@ def _show_mountpoint(self, page=None, mountpoint=None):
...
+    def _get_pages_with_device(self, dev):
+        pages = []
+
+        for page in self._accordion.allPages:
+            for member in page.members:
+                if dev == member._device:
+                    pages.append(page)
+                    break
+
+        return pages
```
jkonecny12: This could be optimized in a way that you will look for just two pages and kill the cycle. There could be systems where this cycle may take a big amount of time.

poncovka: Fixed.

M4rtinK:
jenkins, test this please

jkonecny12:
jenkins, test this please

poncovka:
Removed some of the comments :-)

pyanaconda/ui/gui/spokes/custom.py
```diff
@@ -1917,6 +1935,12 @@ def on_remove_clicked(self, button):
 
             if dialog.deleteAll:
                 for dev in (s._device for s in page.members):
+                    # Skip the device if it is shared with other installations.
+                    if self._is_device_shared(dev):
+                        log.debug("Shared device %s will be skipped.", dev.name)
+                        continue
+
+                    # Destroy the device.
```
jkonecny12: Please remove also this comment. It is only repeating name of the method below.
Also I'm thinking about removing the blank line above and the same for before the return inside of `_is_device_shared` method. However this is more recommendation than requirement for merge.

poncovka: Removed the comment.

PR #1097, opened 2017-06-05
Disable test-install in Makefile

jkonecny12:
Purpose of this test is to find out if all dependencies can be resolved on a target Fedora release. However, we don't want to block other tests because there is a missing dependency on Fedora.

This should be run as separate CI job and it should have informative purpose only.


M4rtinK:
Looks like we have test results again for master with this PR! :)  So merging now as this is basically a hotfix (as discussed with @jkonecny12 ).

PR #1098, opened 2017-06-14
Fix source.glade renaming mixup (#1461469)

M4rtinK:
The source.glade file has been renamed to installation_source.glade,
but the path has not been changed accordingly in all places.


M4rtinK:
Merging now as a hotfix - this is blocking graphical installation from working in the first Rawhide compose in two weeks - so the issue should be fixed quickly.

PR #1101, opened 2017-06-15
Make geolocation with kickstart possible (#1358331)

M4rtinK:
Make it possible to use geolocation during a kickstart based
installation. Geolocation is disabled by default during a kickstart
installation to make it more deterministic and independent on geographic
location of the machine.

To enable geolocation during a kickstart installation, use the
inst.geoloc-use-with-ks=1 option (or the --geoloc-use-with-ks=1 option
when running Anaconda as an application from the command line).

There is also an API that install classes can use to enable geolocation
during kickstart installations - the use_geolocation_with_kickstart
attribute.

Resolves: rhbz#1358331

### DIFF
--- a/pyanaconda/geoloc.py
+++ b/pyanaconda/geoloc.py
@@ -1,5 +1,5 @@
 #
-# Copyright (C) 2013  Red Hat, Inc.
+# Copyright (C) 2013, 2017  Red Hat, Inc.
 #
 # This copyrighted material is made available to anyone wishing to use,
 # modify, copy, or redistribute it subject to the terms and conditions of
@@ -20,14 +20,21 @@
 A GeoIP and WiFi location module - location detection based on IP address
 
 How to use the geolocation module
-   First call :func:`init_geolocation` - this creates the LocationInfo singleton and
-   you can also use it to set what geolocation provider should be used.
-   To actually look up current position, call :func:`refresh` - this will trigger
-   the actual online geolocation query, which runs in a thread.
+   First call init_geolocation() with appropriate parameters - this creates the Geolocation singleton and
+   specifies what geolocation provider will be used.
+
+   To actually look up current position, call the refresh() function of the singleton,
+   this will trigger the actual online geolocation query, which runs in a thread.
+
+   It's possible to wait for the lookup to finish by calling the wait_for_refresh_to_finish() method
+   of the singleton. If a lookup is in progress it will block until the lookup finishes or a timeout
+   is reached. If no lookup is in progress in till return at once.
+
    After the look-up thread finishes, the results are stored in the singleton
-   and can be retrieved using the :func:`get_territory_code` and :func:`get_result` methods.
-   If you call these methods without calling :func:`refresh` first or if the look-up
-   is currently in progress, both return ``None``.
+   and can be retrieved using the territory, timezone and result properties.
+
+   If you use these properties without calling refresh() first or if the look-up is currently
+   in progress or failed to return any results all properties will return None.
 
 ====================
 Geolocation backends
@@ -40,7 +47,7 @@ This module currently supports three geolocation backends:
 
 Fedora GeoIP backend
    This is the default backend. It queries the Fedora GeoIP API for location
-   data based on current public IP address. The reply is JSON formated and
+   data based on current public IP address. The reply is JSON formatted and
    contains the following fields:
    postal_code, latitude, longitude, region, city, country_code, country_name,
    time_zone, country_code3, area_code, metro_code, region_name and dma_code
@@ -51,8 +58,8 @@ Hostip backend
    from current public IP address. The public IP address is determined
    automatically when calling the API.
    GeoIP results from Hostip contain the current public IP and an approximate
-   address. To get this detail location info, use the get_result() method
-   to get an instance of the LocationResult class, used to wrap the result.
+   address. To get this detail location info, use the result property to get
+   an instance of the LocationResult class, used to wrap the lookup result.
 
 Google WiFi backend
    This backend is probably the most accurate one, at least as long as the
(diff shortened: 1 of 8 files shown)


pyanaconda/geoloc.py
```diff
@@ -117,22 +117,100 @@
...
 
+geolocation_enabled = False
 location_info_instance = None
 refresh_condition = threading.Condition()
 refresh_in_progress = False
 
 
-def init_geolocation(provider_id=constants.GEOLOC_DEFAULT_PROVIDER):
+def set_geolocation_enabled(options_override, install_class_override):
+    """Check if geolocation can be used during this installation run.
```
jkonecny12: I don't think that the name of the function `set_geolocation_enabled()` is doing what the name is telling us. It is even more clear when you look on the first row of the docstring: 
"**Check** if geolocation can be used during this installation run." 
There is no word about setting anything but more importantly there isn't a word about checking in the function name.

You could rename this function to `check_and_set_geolocation_enabled()` however then you have the `and` word which clearly telling us this function having a side effect and it also violate SRP principle.

I would recommend you to rewrite this to two functions. One will do just a checking and the second one will be a simple setter. It will be easier for any future modifications and bug hunting.

pyanaconda/geoloc.py
```diff
@@ -117,22 +117,100 @@
...
+    else:
+        if flags.imageInstall or flags.dirInstall:
+            log.info("Geolocation is disabled for image or directory installation.")
+        elif flags.automatedInstall:
+            log.info("Geolocation is disabled due to automated kickstart based installation.")
+        if not flags.cmdline.getbool('geoloc', True):
+            log.info("Geolocation is disabled by the geoloc option.")
+
+    global geolocation_enabled
+    geolocation_enabled = use_geolocation
```
jkonecny12: I don't think we want to have a global variable and it is even worse when it is used for the outside decisions. What about abstraction of this code to a Class with the variable inside of it. All the functions could be static to avoid requirement of holding the instance somewhere.
In any case I don't think we want to have `geolocation_enabled` global variable, IMHO it should be read only property for the outside of the geoloc.py source code.

M4rtinK:
V2 should be done now - looking forward to new reviews. :)

pyanaconda/geoloc.py
```diff
@@ -117,205 +124,194 @@
...
+                use_geolocation = False
+
+        # and also check if geolocation was not disabled by boot or command like option
+        if not flags.cmdline.getbool('geoloc', True):
+            # flags.cmdline.getbool is used as it handles values such as
+            # 0, no, off and also nogeoloc as False
+            # and other values or geoloc not being present as True
+            use_geolocation = False
+
+        # log the end result
```
jkonecny12: I would suggest to move this to separate method.

pyanaconda/geoloc.py
```diff
@@ -117,205 +124,194 @@
...
-    * the refresh is still in progress
-    * the geolocation module was not activated
-      (:func:`init_geolocation` & :func:`refresh` were not called)
+        If there is no lookup in progress (no Geolocation refresh thread
+        is running), this function returns at once.
 
-       * this is for example the case during image and directory installs
+        :param float timeout: how many seconds to wait before timing out
+        """
+        # check if there is a refresh in progress
```
jkonecny12: You can remove this comment the name of the method said that already.

pyanaconda/geoloc.py
```diff
@@ -338,12 +334,13 @@ def __init__(self,
                             once the class is initialized
         :type refresh_now: bool
         """
-        self._provider = _get_provider(provider_id)
+        provider = AVAILABLE_GEOLOCATION_PROVIDERS.get(provider_id, FedoraGeoIPProvider)
```
jkonecny12: This looks really weird to me. Everything is with capital letters but you are calling `get` with two parameters. Could you please change this to be less confusing for using?

pyanaconda/geoloc.py
```diff
@@ -489,52 +501,56 @@ def refresh(self, force=False):
...
         :rtype: LocationResult
         """
         with self._result_lock:
             return self._result
 
+    @result.setter
+    def result(self, new_result):
+        with self._result_lock:
+            self._result = new_result
+
```
jkonecny12: Do you really want public setter for `result`?

pyanaconda/geoloc.py
```diff
@@ -489,52 +501,56 @@ def refresh(self, force=False):
...
+            with self._refresh_condition:
+                self._refresh_in_progress = False
+                self._refresh_condition.notify_all()
             # check if there were any results
-            result = self.get_result()
-            if result:
+            if self.result:
                 log.info("got results from geolocation")
-                slog.info("geolocation result:\n%s", result)
+                slog.info("geolocation result:\n%s", self.result)
```
jkonecny12: Please rename `slog`. Not all abbreviations are great and I personally wouldn't guess that `slog` means `sensitive_log`. I would rather see `sensitive_log` and type little bit more than trying to find what `slog` is.

pyanaconda/geoloc.py
```diff
@@ -117,205 +124,194 @@
...
         else:
-            elapsed_time = time.time() - start_time
-            log.info("Waited %1.2f seconds for Geolocation", elapsed_time)
-    refresh_condition.release()
-    return location_info_instance
+            # fall back to the default provider
+            return None
 
 
 class GeolocationError(Exception):
```
jkonecny12: Please move this exception class to the top. AFAICT it is convention around here.

pyanaconda/geoloc.py
```diff
@@ -704,7 +720,8 @@ def reverse_geocode_coords(self, coordinates):
             return None  # unknown geocoder specified
```
jkonecny12: I can't place comment near to that place but make the `Geocoder` doc class one line please.

pyanaconda/geoloc.py
M4rtinK: Done.

pyanaconda/geoloc.py
M4rtinK: Done.

pyanaconda/geoloc.py
M4rtinK: Good point! :)

M4rtinK: Looks like it was at the top before I added the new class - oops! :)

pyanaconda/geoloc.py
M4rtinK: Yeah, we really don't need this as a constant - so I've moved it to a local variable.

pyanaconda/geoloc.py
M4rtinK: Good point - done. :)

pyanaconda/geoloc.py
M4rtinK: Yeah, that makes sense - turned it into a private setter function.

pyanaconda/geoloc.py
M4rtinK: All class-level docstrings should be now fixed.

M4rtinK:
Thanks! :)

M4rtinK:
jenkins, test this please

jkonecny12:
Jenkins, it's ok to test.

M4rtinK:
jenkins, test this please

PR #1103, opened 2017-06-16
Fix the 'non-ASCII characters in password' checks (#1413813)

AdamWill:
Currently, you can never successfully click through the 'non-
ASCII characters in password' warning. This is because the
`elif failed_check == self._pwASCIICheck:` blocks are wrongly
indented - they go with `if not self.policy.strict`, so they
would *only* kick in if the 'strict' policy was active. They
should be indented one additional level so they go with all
the *other* `if failed_check` conditionals. It also seems
logical to move them above the `elif failed_check:` catch-all.

### DIFF
--- a/pyanaconda/ui/gui/spokes/password.py
+++ b/pyanaconda/ui/gui/spokes/password.py
@@ -202,11 +202,11 @@ class PasswordSpoke(FirstbootSpokeMixIn, NormalSpoke, GUISpokeInputCheckHandler)
             elif failed_check == self._pwEmptyCheck:
                 self.waive_clicks += 1
                 self._pwEmptyCheck.update_check_status()
+            elif failed_check == self._pwASCIICheck:
+                self.waive_ASCII_clicks += 1
+                self._pwASCIICheck.update_check_status()
             elif failed_check:  # no failed checks -> failed_check == None
                 failed_check.update_check_status()
-        elif failed_check == self._pwASCIICheck:
-            self.waive_ASCII_clicks += 1
-            self._pwASCIICheck.update_check_status()
 
         if GUISpokeInputCheckHandler.on_back_clicked(self, button):
             NormalSpoke.on_back_clicked(self, button)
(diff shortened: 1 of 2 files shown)


AdamWill:
A little note here: the `check_password_ASCII` method says this in its docstring:

"Like the password strength check, this check can be bypassed by pressing Done twice."

However, this isn't actually how the code works. The code is written to only require you to press Done *once*, and that's actually also what the message seems to suggest:

"The password you have provided contains non-ASCII characters. You may not be able to switch between keyboard layouts to login. Press Done to continue."

Note it doesn't tell you to press Done *twice*, only to "Press Done". Since the code and the message are in accordance I didn't change this behaviour, but if it's really intended that you only have to click Done once, the docstring should be updated. If it was really intended to require the user to click twice (as with the 'weak password' check), the code and the message text should be changed.

M4rtinK:
jenkins, test this please

M4rtinK:
> A little note here: the check_password_ASCII method says this in its docstring:
> 
> "Like the password strength check, this check can be bypassed by pressing Done twice."
> 
> However, this isn't actually how the code works. The code is written to only require you to press Done once, and that's actually also what the message seems to suggest:
> 
> "The password you have provided contains non-ASCII characters. You may not be able to switch between keyboard layouts to login. Press Done to continue."
> 
> Note it doesn't tell you to press Done twice, only to "Press Done". Since the code and the message are in accordance I didn't change this behaviour, but if it's really intended that you only have to click Done once, the docstring should be updated. If it was really intended to require the user to click twice (as with the 'weak password' check), the code and the message text should be changed.

There has been quite a lot of password checking fixes and improvement on the rhel7-branch, likely including this case. I'll be backporting these fixes/improvements to Fedora (master branch) shortly, which should hopefully fix this issue. :)

M4rtinK:
jenkins, test this please

AdamWill:
For the record, now I look at the old code, this is still a slight change in behaviour. The old code looked like this:

       if not self.policy.strict and failed_check == self._pwStrengthCheck:
            self._waiveStrengthClicks += 1
            self._pwStrengthCheck.update_check_status()
        elif failed_check == self._pwASCIICheck:
            self._waiveASCIIClicks += 1
            self._pwASCIICheck.update_check_status()

i.e. it allowed non-ASCII passwords *even in strict policy mode*. The code as it exists now my patch has been merged only allows non-ASCII passwords in non-strict mode. So we might want to change it again if we want to keep previous behaviour.

PR #1104, opened 2017-06-19
Fix import from a renamed module (#1462538)

poncovka:
Module pyanaconda.threads has been renamed to pyanaconda.threading.

Resolves: rhbz#1462538

### DIFF
--- a/pyanaconda/kickstart.py
+++ b/pyanaconda/kickstart.py
@@ -19,7 +19,7 @@
 #
 
 from pyanaconda.errors import ScriptError, errorHandler
-from pyanaconda.threads import threadMgr
+from pyanaconda.threading import threadMgr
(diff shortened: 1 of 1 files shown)


M4rtinK:
Looks good to me as well. :)

AdamWill:
Can we get a package build ASAP? anaconda being DOA causes all kinds of problems (it prevents live and cloud images from building and being tested at all, for instance). Thanks!

M4rtinK:
> Can we get a package build ASAP? anaconda being DOA causes all kinds of problems (it prevents live and cloud images from building and being tested at all, for instance). Thanks!

Build running:
https://koji.fedoraproject.org/koji/taskinfo?taskID=20075262

:)

PR #1107, opened 2017-06-20
Use context manager to check KickstartError

jkonecny12:
Code for the checking if KickstartError raised were copied in the code. Use context manager to avoid copying.

*Resolves: rhbz#1463118*

*Reported-by: Peter Kotvan <pkotvan@redhat.com>*
(cherry picked from commit fa68680d0c13fb3edec2e942c587bea49f8b78da)

### DIFF
--- a/anaconda.py
+++ b/anaconda.py
@@ -719,9 +719,11 @@ if __name__ == "__main__":
 
     # Create pre-install snapshots
     from pykickstart.constants import SNAPSHOT_WHEN_PRE_INSTALL
+    from pyanaconda.kickstart import check_kickstart_error
     if ksdata.snapshot.has_snapshot(SNAPSHOT_WHEN_PRE_INSTALL):
-        ksdata.snapshot.pre_setup(anaconda.storage, ksdata, anaconda.instClass)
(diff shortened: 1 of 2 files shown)


jkonecny12:
I've already looked on that the problem is that it is not 100% applicable in this way, and can't find reasonable solution here.

The core issue is in these `except` parts here:
```
parsing_log.error(e)
...
# Print kickstart warnings in the strict mode.
if strict_mode and kswarnings:
    print(_("\nSome warnings occurred during reading the kickstart file:"))
    for w in kswarnings:
        print(str(w).strip())

...
time.sleep(10)
```

Suggestions are welcomed.

I'm thinking about using these variants everywhere, that could work but the `time.sleep(10)` may be something to worry about.

poncovka:
The time of sleep could be a parameter of the context manager. But probably it is not worth changing this part, since we need it only once.

jkonecny12:
I really like it but in combination with the strict mode prints I don't want to change it here.
I'm merging this now. However, if there is somebody with idea how to improve it, I would gladly help with that effort ;).

PR #1108, opened 2017-06-20
Refactor imports in kickstart.py

jkonecny12:
Other suggestions are welcomed.

### DIFF
--- a/pyanaconda/kickstart.py
+++ b/pyanaconda/kickstart.py
@@ -18,69 +18,63 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #
 
-from pyanaconda.errors import ScriptError, errorHandler
-from pyanaconda.threading import threadMgr
-from blivet.deviceaction import ActionCreateFormat, ActionResizeDevice, ActionResizeFormat
-from blivet.devices import LUKSDevice
-from blivet.devices.lvm import LVMVolumeGroupDevice, LVMCacheRequest, LVMLogicalVolumeDevice
-from blivet.devicelibs.lvm import LVM_PE_SIZE, KNOWN_THPOOL_PROFILES
-from blivet.devicelibs.crypto import MIN_CREATE_ENTROPY
-from blivet.formats import get_format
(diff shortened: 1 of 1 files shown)


pyanaconda/kickstart.py
```diff
@@ -18,69 +18,64 @@
...
+from blivet import autopart, udev
+from blivet.deviceaction import ActionCreateFormat, ActionResizeDevice, ActionResizeFormat
+from blivet.devicelibs.crypto import MIN_CREATE_ENTROPY
+from blivet.devicelibs.lvm import LVM_PE_SIZE, KNOWN_THPOOL_PROFILES
+from blivet.devices import LUKSDevice
+from blivet.devices.lvm import LVMVolumeGroupDevice, LVMCacheRequest, LVMLogicalVolumeDevice
+from blivet.errors import PartitioningError, StorageError, BTRFSValueError
+from blivet.formats.fs import XFS
+from blivet.formats import get_format
+from blivet.partitioning import do_partitioning
```
poncovka: There are two lines for `blivet.partitioning`, they could be merged together.

pyanaconda/kickstart.py
```diff
@@ -18,69 +18,64 @@
...
-from pyanaconda.simpleconfig import SimpleConfigFile
-from pyanaconda.users import getPassAlgo
+import warnings
+
+import blivet.arch
+import blivet.fcoe
+import blivet.iscsi
+import blivet.zfcp
+
+import pykickstart.commands as commands
```
poncovka: This import could be moved to other `pykickstart` imports.

jkonecny12: It is here intentionally because it is an `import` and the other `pykickstart` are from. This is looking better for an eye to me.

pyanaconda/kickstart.py
jkonecny12: I missed that, thanks!

pyanaconda/kickstart.py
poncovka: Ok, I didn't notice that.

PR #1110, opened 2017-06-20
rpmostreepayload: Stub out payload methods which use `import rpm`

cgwalters:
I'm not entirely sure what changed from 7.3 to 7.4 to cause this
to start breaking; we're getting an uninitialized rpm database in
`/var/lib/rpm` instead of the default rpmostree version of `/usr/share/rpm`.
This happens because of 2 methods which are doing `import rpm`
inside anaconda *before* we've had a chance to run systemd-tmpfiles and
set up the compatibility link.

Related: rhbz#1462979


jkonecny12:
jenkins, test this please

PR #1114, opened 2017-06-26
Install class shouldn't set the default boot fstype (#1463297)

poncovka:
We should let storage to choose the default boot fstype. This is
inconsistent with autopart --fstype, where autopart sets both the
default fstype and the default boot fstype.

(fix of the commit 3e7d9c8)

Resolves: rhbz#1463297

### DIFF
--- a/pyanaconda/anaconda.py
+++ b/pyanaconda/anaconda.py
@@ -221,18 +221,24 @@ class Anaconda(object):
 
     def _set_default_fstype(self, storage):
         fstype = None
+        boot_fstype = None
 
         # Get the default fstype from a kickstart file.
         if self.ksdata.autopart.autopart and self.ksdata.autopart.fstype:
             fstype = self.ksdata.autopart.fstype
+            boot_fstype = self.ksdata.autopart.fstype
         # Or from an install class.
         elif self.instClass.defaultFS:
             fstype = self.instClass.defaultFS
+            boot_fstype = None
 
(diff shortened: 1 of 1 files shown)


Frodox:
is it possible to set it from install_class.py ? :)

pyanaconda/anaconda.py
```diff
@@ -221,18 +221,24 @@ def noninteractive_tui_mode(self):
 
     def _set_default_fstype(self, storage):
         fstype = None
+        boot_fstype = None
 
         # Get the default fstype from a kickstart file.
         if self.ksdata.autopart.autopart and self.ksdata.autopart.fstype:
             fstype = self.ksdata.autopart.fstype
+            boot_fstype = self.ksdata.autopart.fstype
```
M4rtinK: Was this the same before, eq. - that the fstype set from kickstart with "autopart --fstype" was also used for the bootloader ?

M4rtinK:
@Frodox The problem is that the default fstype is also used for the boot partition, which might not work in some cases (Fedora on armv7 booted via uboot, etc.). As for install class being able to specify file system for /boot - I'm not sure we support that right now - if we don't & you think it could be useful then you can open a RFE bug for that. :)

pyanaconda/anaconda.py
poncovka: Yes.

PR #1118, opened 2017-06-29
Fix a typo in python-meh initialization (#1462825)

M4rtinK:
Fix a typo in initialization of the python-meh provided Config class.

A while ago the fileList parameter was renamed to file_list on Anaconda
side due to a small refactoring error. Unfortunately the __init__()
method of the Config class silently accepts *any parameter names whatsoever*,
even if they don't match the parameters it actually accepts.

End result: The additional log files Anaconda specifies are not included
            in the dump as file_list != fileList.

So let's:
- fix parameter name
- add "Make the __init__() method of the python-meh Config class sane"
  to our TODO list

### DIFF
--- a/pyanaconda/exception.py
+++ b/pyanaconda/exception.py
@@ -277,7 +277,7 @@ def initExceptionHandling(anaconda):
                                 "_bootloader.password",
                                 "payload._groups"],
                   localSkipList=["passphrase", "password", "_oldweak", "_password", "try_passphrase"],
-                  file_list=file_list)
+                  fileList=file_list)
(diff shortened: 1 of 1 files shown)


AdamWill:
Didn't test this yet, but I followed the code and it makes perfect sense, obviously someone just replaced all occurrences of `fileList` with `file_list` and that caused the problem. +1

M4rtinK:
Merging now to make sure this is included in tomorrows build. The fix is simple & has been ACKed so I think it should be fine.

Issue #1117, opened 2017-06-28
teach rpmostreepayload that /run/install/repo is a local mirror


cgwalters:
When doing `ostreesetup` kickstarts of Fedora Atomic Host, we still have the "default tree" in `/run/install/ostree`.  If we taught rpmostreepayload to use that as a local mirror, kickstart installs would only download the differences.

cgwalters:
Related: https://github.com/rhinstaller/lorax/issues/143

cgwalters:
Depends: https://github.com/ostreedev/ostree/issues/975

cgwalters:
PR in https://github.com/rhinstaller/anaconda/pull/1119

PR #1119, opened 2017-06-29
rpmostreepayload: Reuse the local repo as a cache

cgwalters:
When doing kickstart installs (can happen for both bare metal PXE as well as
building cloud images), we currently end up redownloading all of the ostree
content, even though we have a version in the squashfs already.

However, the version we have may not be the version we want; the installer image
is typically built at a different/slower cycle from the ostree content.

But this is where a strength of libostree's content-addressed storage shines; we
can use the local repo as an object cache, and avoid fetching what we already
have.

This just enables the logic that lives in libostree.

Closes: https://github.com/rhinstaller/anaconda/issues/1117

### DIFF
--- a/pyanaconda/payload/rpmostreepayload.py
+++ b/pyanaconda/payload/rpmostreepayload.py
@@ -170,8 +170,21 @@ class RPMOSTreePayload(ArchivePayload):
         progress = OSTree.AsyncProgress.new()
         progress.connect('changed', self._pullProgressCb)
 
+        pull_opts = {'refs': GLib.Variant('as', [ostreesetup.ref])}
+        # If we're doing a kickstart, we can at least use the content as a reference:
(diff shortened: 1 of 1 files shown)


cgwalters:
This is a WIP since the API hasn't landed in libostree yet, but I tested this and it works nicely!

jkonecny12:
Jenkins, it's ok to test.

cgwalters:
The upstream PR was merged, removing WIP.  Note that I added a version check, so we'll work with both old and new versions.  Hence this should be safe to merge at any time.

libostree releases usually are on a 3-4 week cadence, so it'll be a week or two.

cgwalters:
For reference the OSTree PR is: https://github.com/ostreedev/ostree/pull/982

PR #1120, opened 2017-06-30
Support for installclass command and refactorization

poncovka:
The install class can be now specified in the kickstart file by its name.

The logic for loading install classes is now hidden in the `InstallClassFactory`. 
The `availableClasses` method was removed, since we can use `collect` from 
`pyanaconda.iutil`. The class `DefaultInstall` was removed as well, because it 
was useless. Added tests for the `InstallClassFactory`.

Modules with install classes should define `__all__`.  Otherwise, when the 
module is loaded, anaconda again discovers also the imported install 
classes (like the base install class).

Depends on https://github.com/rhinstaller/pykickstart/pull/160

### DIFF
--- a/pyanaconda/anaconda.py
+++ b/pyanaconda/anaconda.py
@@ -91,7 +91,14 @@ class Anaconda(object):
     def instClass(self):
         if not self._instClass:
             from pyanaconda.installclass import factory
-            self._instClass = factory.get_best_install_class()
+
+            # Get install class by name.
+            if self.ksdata.installclass.seen:
+                name = self.ksdata.installclass.name
+                self._instClass = factory.get_install_class_by_name(name)
+            # Or just find the best one.
+            else:
+                self._instClass = factory.get_best_install_class()
 
         return self._instClass
 
--- a/pyanaconda/anaconda.py
+++ b/pyanaconda/anaconda.py
@@ -90,8 +90,8 @@ class Anaconda(object):
     @property
     def instClass(self):
         if not self._instClass:
-            from pyanaconda.installclass import DefaultInstall
-            self._instClass = DefaultInstall()
+            from pyanaconda.installclass import factory
+            self._instClass = factory.get_best_install_class()
 
         return self._instClass
 
--- a/pyanaconda/installclass.py
+++ b/pyanaconda/installclass.py
@@ -193,6 +193,17 @@ class InstallClassFactory(object):
         self._visible_classes = []
         self._paths = []
 
+    def get_install_class_by_name(self, name):
+        """Return an instance of an install class with a requested name."""
+        for install_class in self.classes:
+            if install_class.name == name:
+                log.info("Using the requested install class %s.",
+                         self._get_class_description(install_class))
+
+                return install_class()
+
+        raise RuntimeError("Unable to find the install class %s.", name)
+
     def get_best_install_class(self):
         """Return the instance of the best found install class."""
         if self.visible_classes:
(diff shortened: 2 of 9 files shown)


pyanaconda/anaconda.py
```diff
@@ -91,7 +91,14 @@ def bootloader(self):
     def instClass(self):
         if not self._instClass:
             from pyanaconda.installclass import factory
-            self._instClass = factory.get_base_install_class()
+
+            # Get required install class.
+            if self.ksdata.installclass.seen:
+                name = self.ksdata.installclass.name
+                self._instClass = factory.get_required_install_class(name)
+            # Or just find the best one.
```
jkonecny12: I wouldn't name `base` install class as "the best one"; or the function name below is incorrect.

pyanaconda/installclass.py
```diff
@@ -175,114 +178,105 @@ def setNetworkOnbootDefault(self, ksdata):
...
+                path += os.environ["ANACONDA_INSTALL_CLASSES"].split(":")
+
+            path += [
+                "installclasses",
+                "/tmp/updates/pyanaconda/installclasses",
+                "/tmp/product/pyanaconda/installclasses",
+                "%s/pyanaconda/installclasses" % get_python_lib(plat_specific=1)
+            ]
+
+            self._paths = list(filter(lambda d: os.access(d, os.R_OK), path))
```
jkonecny12: What about hiding this implementation to some function? For example `find_all_installclass_paths()`?

It would be easier to read:

```
if not self._paths:
    self.find_all_installclass_paths()
return self._paths
```

It is the same as you have for ``classes`` property.

pyanaconda/installclass.py
```diff
@@ -175,114 +178,116 @@ def setNetworkOnbootDefault(self, ksdata):
...
+            if install_class.name == name:
+                log.info("Using the required install class %s.",
+                         self._get_class_description(install_class))
+
+                return install_class()
+
+        raise RuntimeError("Unable to find the install class %s.", name)
+
+    def get_base_install_class(self):
+        """Return the instance of the best found install class."""
```
poncovka: Fixed.

pyanaconda/anaconda.py
poncovka: Fixed.

pyanaconda/installclass.py
```diff
@@ -175,114 +178,116 @@ def setNetworkOnbootDefault(self, ksdata):
...
-
-    return cobject
-
-baseclass = getBaseInstallClass()
-
-# we need to be able to differentiate between this and custom
-class DefaultInstall(baseclass):
+
+class InstallClassFactory(object):
+    """Class to get the install class instance."""
```
poncovka: Fixed.

pyanaconda/installclass.py
```diff
@@ -175,114 +178,116 @@ def setNetworkOnbootDefault(self, ksdata):
...
+
     def __init__(self):
-        baseclass.__init__(self)
+        self._classes = []
+        self._visible_classes = []
+        self._paths = []
+
+    def get_required_install_class(self, name):
```
poncovka: Fixed.

pyanaconda/installclass.py
poncovka: Fixed.

poncovka:
Jenkins, test this please

poncovka:
Fixed tests.

PR #1122, opened 2017-07-07
gui: show supported locales on Atomic Host installs

jlebon:
Fedora Atomic Host has never shipped with all supported locales. It
originally shipped with only `en_US`, though that was changed afterwards
to cover at least the same locales as RHEL7[1]. However, we would still
show all the locales in the installer GUI.

This patch adds support for allowing installclasses to filter locales
that should be shown based on the actual content carried in the ISO. The
Atomic installclass can then make use of this. This is essentially the
same patch as the one for rhel7-branch[2], but split out since the
Atomic installclass is not here.

[1] https://pagure.io/fedora-atomic/c/5a5edddc9a6b943e6fe902defad177009c43867a
[2] https://github.com/rhinstaller/anaconda/pull/871

Related: https://pagure.io/atomic-wg/issue/282

### DIFF
--- a/docs/boot-options.rst
+++ b/docs/boot-options.rst
@@ -351,9 +351,6 @@ If a language has been specified via the `inst.lang`_ boot option
 or the `lang` kickstart command it will be used.
 If no language is specified Anaconda will default to en_US.UTF-8.
 
-.. NOTE::
-         Atomic installations always run in single language mode.
-
 .. inst.geoloc:
 
 inst.geoloc
--- a/pyanaconda/installclass.py
+++ b/pyanaconda/installclass.py
@@ -172,6 +172,12 @@ class BaseInstallClass(object):
     def setNetworkOnbootDefault(self, ksdata):
         pass
 
+    def filterSupportedLangs(self, ksdata, langs):
+        return langs
+
+    def filterSupportedLocales(self, ksdata, lang, locales):
+        return locales
+
     def __init__(self):
         pass
 
--- a/pyanaconda/ui/gui/spokes/langsupport.py
+++ b/pyanaconda/ui/gui/spokes/langsupport.py
@@ -139,10 +139,16 @@ class LangsupportSpoke(LangLocaleHandler, NormalSpoke):
     def completed(self):
         return True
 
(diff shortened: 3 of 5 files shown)


jlebon:
Requires https://github.com/rhinstaller/lorax/pull/227 and https://bugzilla.redhat.com/show_bug.cgi?id=1468619 to actually be functional.

jkonecny12:
Jenkins, it's ok to test.

jlebon:
@poncovka @jkonecny12 Can we get an f26 update for this? Fedora AH releases every two weeks (including a new installer ISO), so new users would profit from this quite quickly.

jlebon:
Ping, any chance of an F26 update? Both the corresponding updated lorax and fedora-productimg-atomic packages are currently in updates-testing.

jkonecny12:
@jlebon  we are not releasing updates for once released Fedora. We are normally working only on rawhide version + Fedora in development because there is no point for Anaconda to do that in general.

However, we will discuss this (because it could be useful for AH) and I'll give you info.

dustymabe:
> @jlebon we are not releasing updates for once released Fedora. We are normally working only on rawhide version + Fedora in development because there is no point for Anaconda to do that in general.

We've had this discussion with at least some of the anaconda team in the past. We concluded that these changes would only affect atomic host (since no other media for f26 is getting built any longer) and thus it was not a big deal (doesn't need normal QE, etc) to release a new update for anaconda. This was during the f25 cycle that we had this discussion so it was an update to f25 that we released in the past.

dustymabe:
> We concluded that these changes would only affect atomic host (since no other media for f26 is getting built any longer)

One minor correction. I believe we do release updated docker images, but that is also kind of under the scope of the atomic working group, so we'd still only be affecting ourselves.

M4rtinK:
Sounds good - I'll do a new F26 build later today or tomorrow. :)

dustymabe:
> Sounds good - I'll do a new F26 build later today or tomorrow. :)

We have a build, now need it submitted as an update. Thanks @M4rtinK

Issue #1007, opened 2017-03-30
Handle /var being a mount point with rpmostreepayload


cgwalters:
In ostree, `/var` is a bind mount to outside the deployment root.  Today in RHEL7, the installer crashes trying to write `/root/anaconda-ks.cfg`.  In F25AH, we don't crash, but it looks like our `prepareMountTargets()` needs to handle the case where at least `/var` is a mount point.  We don't want to overwrite it.

cgwalters:
PR in https://github.com/rhinstaller/anaconda/pull/1063

cgwalters:
There's still an outstanding bug here; the `user` instruction writes to `/mnt/sysimage/var/home`.  A hackaround is:

```
# https://github.com/rhinstaller/anaconda/issues/1007
%post --nochroot
root=$(echo /mnt/sysimage/ostree/deploy/*/deploy/*.0/ | head -1)
for x in /var/home/*; do
  if test -d ${x} && ! test -d ${root}/var/${x}; then
    mv ${x} ${root}/var/
  fi
done
%end
```

cgwalters:
Reading the code a bit, this looks likely to be because `createLuserConf()` is called early on once, we need to rerun it after we call `iutil.setSysroot()`.

jkonecny12:
Nice investigation, could you please create a PR for this.

cgwalters:
Apparently when I was debugging this before, I was talking about RHEL7, since `createLuserConf()` hasn't existed since https://github.com/rhinstaller/anaconda/commit/7370f1b277f8082499c76391d017b4984c5e0230

Testing `Fedora-Atomic-ostree-x86_64-Rawhide-20170710.n.0.iso` we have an entirely different bug which is:

`/mnt/sysimage/ostree/deploy/fedora-atomic/deploy/14ee8ad14e076285d006503b6e6ee23eafebbf137bdf3f9f9dca0b1cf8ce2e60.0/sysroot/var`

That `/sysroot` in there is wrong.  I don't quite understand why it's happening; I am pretty sure this worked at one point.  Still debugging more.

cgwalters:
Combined with https://github.com/systemd/systemd/pull/6293 this should let the anaconda default mount for `/home` to work as well.

cgwalters:
More patches in https://github.com/rhinstaller/anaconda/pull/1124

PR #1124, opened 2017-07-12
A few more patches for rpmostreepayload + /var (and other) mounts

cgwalters:
This will close https://github.com/rhinstaller/anaconda/issues/1007

I'm tempted to squash the patches...but the series I think is quite useful to fully
see the chain of bugfixes and rationale.


pyanaconda/payload/rpmostreepayload.py
```diff
@@ -239,46 +239,73 @@ def install(self):
...
+            src = iutil.getTargetPhysicalRoot() + src
+        else:
+            src = iutil.getSysroot() + src
+        # Canonicalize dest to the full path
+        dest = iutil.getSysroot() + dest
+        if bind_ro:
+            self._safeExecWithRedirect("mount",
+                                       ["--bind", src, src])
+            self._safeExecWithRedirect("mount",
+                                       ["--bind", "-o", "remount,ro", src, src])
```
jkonecny12: Why do you have two steps for the `ro` mount?

pyanaconda/payload/rpmostreepayload.py
```diff
@@ -239,46 +239,73 @@ def install(self):
 
         mainctx.pop_thread_default()
 
+    # Internal API for setting up bind mounts between the physical root and
+    # sysroot, also ensures we track them in self._internal_mounts so we can
+    # cleanly unmount them.
```
jkonecny12: Add this documentation comment below the function name.

jkonecny12:
Jenkins, it's ok to test.

pyanaconda/payload/rpmostreepayload.py
cgwalters: That's how it works, see `man mount`.

>        mount(8) since v2.27 allows to change the mount options by passing the relevant options along with --bind.  For example:
>
>              mount --bind,ro foo foo
>
>       This feature is not supported by the Linux kernel; it is implemented in userspace by an additional mount(2) remounting system call.  This solution is not atomic.
>
>       The alternative (classic) way to create a read-only bind mount is to use the remount operation, for example:
>
>              mount --bind olddir newdir
>              mount -o remount,bind,ro olddir newdir

(We could use the `mount --bind,ro` feature but I'd need to undo it when backporting to RHEL)

pyanaconda/payload/rpmostreepayload.py
cgwalters: OK, I tweaked it into a pydoc comment as well. :arrow_down:

jkonecny12:
Please fix this issue:
```
************* Module ../pyanaconda/payload/rpmostreepayload.py
W0612(unused-variable):331,20: RPMOSTreePayload.prepareMountTargets: Unused variable 'device'
FAIL pylint/runpylint.py (exit status: 6)
```

I think we can merge this when this error above will be fixed.

cgwalters:
Updated :arrow_up:

PR #1125, opened 2017-07-13
Fix anaconda --help failed with traceback

jkonecny12:
There was missing import from `startup_utils` module.


cgwalters:
I hit this actually when I was trying to do `anaconda --version`.  LGTM.

PR #1126, opened 2017-07-14
rpmostreepayload: use correct secondary url location

dustymabe:
In f5263d4 we should have been using `/run/install/repo/content/repo`
since that is where we were storing ostree repos on media before
fedora 26. Using the incorrect location of `/run/install/repo/`
means that the install will fail if it is an ostree install and there
is no ostree repo at `/ostree/repo`.

### DIFF
--- a/pyanaconda/payload/rpmostreepayload.py
+++ b/pyanaconda/payload/rpmostreepayload.py
@@ -176,7 +176,7 @@ class RPMOSTreePayload(ArchivePayload):
         # The first path here is used by <https://pagure.io/fedora-lorax-templates>
         # and the second by <https://github.com/projectatomic/rpm-ostree-toolbox/>
         if OSTree.check_version(2017, 8):
-            for path in ['/ostree/repo', '/run/install/repo']:
+            for path in ['/ostree/repo', '/run/install/repo/content/repo']:
                 if os.path.isdir(path):
                     pull_opts['localcache-repos'] = GLib.Variant('as', [path])
                     break
--- a/pyanaconda/payload/rpmostreepayload.py
+++ b/pyanaconda/payload/rpmostreepayload.py
@@ -176,7 +176,7 @@ class RPMOSTreePayload(ArchivePayload):
         # The first path here is used by <https://pagure.io/fedora-lorax-templates>
         # and the second by <https://github.com/projectatomic/rpm-ostree-toolbox/>
         if OSTree.check_version(2017, 8):
-            for path in ['/ostree/repo', '/run/install/repo/content/repo']:
+            for path in ['/ostree/repo', '/install/ostree/repo']:
                 if os.path.isdir(path):
                     pull_opts['localcache-repos'] = GLib.Variant('as', [path])
                     break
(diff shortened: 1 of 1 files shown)


dustymabe:
please let @cgwalters review this before merging.

dustymabe:
see https://pagure.io/atomic-wg/issue/301 where we hit this in Fedora

cgwalters:
Ah...oh, I see.  Currently in Fedora pungi, we're installing AH from the Cloud installer?  Why are we doing that?  It defeats this optimization.

Hmm...so I think my intention there was to handle the RHELAH case, but looking at it it's in fact `/install/ostree` apparently.

pyanaconda/payload/rpmostreepayload.py
```diff
@@ -176,7 +176,7 @@ def install(self):
         # The first path here is used by <https://pagure.io/fedora-lorax-templates>
         # and the second by <https://github.com/projectatomic/rpm-ostree-toolbox/>
         if OSTree.check_version(2017, 8):
-            for path in ['/ostree/repo', '/run/install/repo']:
+            for path in ['/ostree/repo', '/run/install/repo/content/repo']:
                 if os.path.isdir(path):
```
cgwalters: And then here, change it to `if os.path.isdir(path + '/objects')`?

pyanaconda/payload/rpmostreepayload.py
```diff
@@ -176,7 +176,7 @@ def install(self):
         # The first path here is used by <https://pagure.io/fedora-lorax-templates>
         # and the second by <https://github.com/projectatomic/rpm-ostree-toolbox/>
         if OSTree.check_version(2017, 8):
-            for path in ['/ostree/repo', '/run/install/repo']:
+            for path in ['/ostree/repo', '/run/install/repo/content/repo']:
```
cgwalters: Let's ignore pre-F26, but we do want to work on RHELAH.  So can you change the second path to `/install/ostree`?

dustymabe:
> Ah...oh, I see. Currently in Fedora pungi, we're installing AH from the Cloud installer? Why are we doing that? It defeats this optimization.

yeah - i honestly don't know if it's even possible to tell the cloud images to use the ostree_installer. One bad thing this would cause is now the cloud images would depend on the ostree installer finishing first, so more serialization

> 
> Hmm...so I think my intention there was to handle the RHELAH case, but looking at it it's in fact /install/ostree apparently.

What all directories do we need to support? Also it would be great if it were possible to specify one of these alternate repos in the ostreesetup line (so it didn't have to be hardcoded into the installer).

cgwalters:
See my code review comment, I think we need `['/ostree/repo', '/install/ostree/repo']`.  Down the line I want to change the RHELAH one to use `/ostree/repo` as well for the same dedup/storage reasons, so there's no reason to make it configurable.

pyanaconda/payload/rpmostreepayload.py
dustymabe: i'll do this, but i still think it would be worth making ostree gracefully handle this as I proposed in https://pagure.io/atomic-wg/issue/301

cgwalters: I don't think so; if silently ostree ignored nonexistent repos, it would have taken us way longer to find this problem.

dustymabe:
@cgwalters ^^ updated

cgwalters:
LGTM, thanks!

jkonecny12:
Jenkins, it's ok to test.

PR #1127, opened 2017-07-14
Make kickstart rescue command noninteractive.

rvykydal:
It should decide on what to do (mount root, mount root readonly) just
based on rescue command options (in case only one system root is found,
otherwise it goes interactive asking for which root to mount).
This would make repairing systems via kickstart %pre and %post sections
automatically without tmux hacks possible.

This is quite dirty/pragmatic solution, I intend to rewrite rescue to remove more execution logic out of the tui spoke, perhaps in scope of moving the ui to simpleline.

### DIFF
--- a/pyanaconda/rescue.py
+++ b/pyanaconda/rescue.py
@@ -60,7 +60,7 @@ def makeFStab(instPath=""):
     except IOError as e:
         log.info("failed to write /etc/fstab: %s", e)
 
-def run_shell():
+def run_shell(reboot=False):
     """Launch a shell."""
     if flags.imageInstall:
         print(_("Run %s to unmount the system when you are finished.")
@@ -77,7 +77,7 @@ def run_shell():
             print(_("Unable to find /bin/bash to execute!  Not starting shell."))
             time.sleep(5)
 
-    if not flags.imageInstall:
+    if reboot:
         iutil.execWithRedirect("systemctl", ["--no-wall", "reboot"])
     else:
         return None
@@ -122,13 +122,13 @@ class RescueMode(NormalTUISpoke):
     # always. This is independent of any hub(s), so pass in some fake data
     def __init__(self, app, data, storage=None, payload=None, instclass=None):
         NormalTUISpoke.__init__(self, app, data, storage, payload, instclass)
-        if flags.automatedInstall:
-            self._ro = data.rescue.romount
-        else:
-            self._ro = False
-
         self._root = None
         self._choices = (_("Continue"), _("Read-only mount"), _("Skip to shell"), ("Quit (Reboot)"))
+        self._reboot = True
+        if flags.imageInstall:
+            self._reboot = False
+        if flags.automatedInstall and self.data.reboot.action not in [KS_REBOOT, KS_SHUTDOWN]:
+            self._reboot = False
 
     def initialize(self):
         NormalTUISpoke.initialize(self)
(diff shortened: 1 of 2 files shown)


jkonecny12:
Please fix this error:

```
************* Module ../pyanaconda/rescue.py
W0107(unnecessary-pass):143,16: RescueMode.initialize: Unnecessary pass statement
```

rvykydal:
@jkonecny12 I've addressed you remarks and added two more patches separating UI and execution logic. The first one is more about shuffling things. As for the execution functions I've tried to clean them up a bit in the second additional patch. There is still some room for cleanup of functions I haven't touched by the patches, but that was not my point in this set.
Also the prompt/input (mis)usage is still quite wierd in the spokes. Next step would be to move the UI to the new simpleline while sanitizing it.

rvykydal:
Adding also Read The Docs documentation for rescue.

rvykydal:
jenkins, test this please

pyanaconda/rescue.py
```diff
@@ -37,11 +37,13 @@
 import os
```
jkonecny12: The commit comment ``... would be to make the UI "will" more sane ..`` is confusing. I guess the `will` should not be there.

pyanaconda/rescue.py
```diff
@@ -115,41 +96,275 @@ def makeResolvConf(instPath):
...
-        NormalTUISpoke.__init__(self, app, data, storage, payload, instclass)
-        self._root = None
-        self._choices = (_("Continue"), _("Read-only mount"), _("Skip to shell"), ("Quit (Reboot)"))
-        self._reboot = True
-        if flags.imageInstall:
-            self._reboot = False
-        if flags.automatedInstall and self.data.reboot.action not in [KS_REBOOT, KS_SHUTDOWN]:
-            self._reboot = False
+def create_etc_symlinks():
+    """I don't know why I live. Maybe I should be killed."""
```
jkonecny12: I really like this doc string, however I don't understand the meaning :(

pyanaconda/rescue.py
```diff
@@ -182,105 +406,154 @@ def input(self, args, key):
...
-                        device.teardown(recursive=True)
-                        device.format.passphrase = None
-                        try_passphrase = None
+                    unlocked = self._rescue.unlock_device(device_name, passphrase)
+                    try_passphrase = passphrase if unlocked else None
+
+        return not self._rescue.get_locked_device_names()
+
+    def apply(self):
+        """Move along home."""
```
jkonecny12: Maybe I'm just dumb but I don't get it? Am I missing something?

pyanaconda/rescue.py
```diff
@@ -182,105 +406,154 @@ def input(self, args, key):
...
+            else:
+                delay = 0
+                self._rescue.run_shell()
+            self._rescue.finish(delay=delay)
+            return None
+        return Prompt(_("Please press %s to get a shell") % Prompt.ENTER)
 
-        self._root = None
+    def input(self, args, key):
+        """Move along home."""
```
jkonecny12: The same here.

pyanaconda/rescue.py
rvykydal: Yes, thanks.

pyanaconda/rescue.py
rvykydal: I'll remove it, I don't understand it either.

dedekind:
Hi, is there a chance to have this in F27?

jkonecny12:
The F27 Anaconda is still not branched so this will be in F27.

PR #1130, opened 2017-07-26
Add support for automatic generating of DBus specification.

poncovka:
With provided generators dbus_signal, dbus_interface and dbus_class,
it is possible to define a DBus interface with a Python class and
provide a DBus specification of a Python class. Types are derived
from type hints using types from pyanaconda.dbus.typing. Only class
members with CamelCase names are considered to be members of the DBus
interface.

It is supported the inheritance of DBus interfaces. Generated DBus
specification will contain all interfaces of the class ancestors.
If the class defines a new DBus interface, then class members with
names already defined in inherited interfaces are not considered to
be members of the new DBus interface.

Example of a class that defines a new DBus interface with a method,
a signal and a property. Instances of this class can be also exported
to DBus.

```
@dbus_interface("Interface")
class NewInterface(object):

  def Method(self, x: Int) -> List[Int]
    return [x]

  @dbus_signal
  def Signal(self, y: Double)
    pass

  @property
  def Property(self) -> Bool:
    return False
```

Example of a class that doesn't define a new DBus interface, but
instances can be exported to DBus.
```

@dbus_class
class ExportableClass(NewInterface):
  pass
```

It is required DBus library pydbus.

### DIFF
--- a/pyanaconda/dbus/interface.py
+++ b/pyanaconda/dbus/interface.py
@@ -1,0 +1,403 @@
+#
+# interface.py:  support for generating XML with interface description
+#
+# Copyright (C) 2017
+# Red Hat, Inc.  All rights reserved.
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+# Author(s):  Vendula Poncova <vponcova@redhat.com>
+#
+# For more info about DBus specification see:
+# https://dbus.freedesktop.org/doc/dbus-specification.html#introspection-format
+#
+import inspect
+import re
+
+from inspect import Parameter
+from typing import get_type_hints
+from pydbus.generic import signal
+
+from pyanaconda.dbus.typing import get_dbus_type
+from pyanaconda.dbus.xml import XMLGenerator
+
+__all__ = ["dbus_class", "dbus_interface", "dbus_signal"]
+
+
+class dbus_signal(signal):
+    """DBus signal.
+
+    Can be used as:
+        Signal = dbus_signal()
+
+    Or as a method decorator:
+
+        @dbus_signal
+        def Signal(x: Int, y: Double):
+            pass
+
+    Signal is defined by the type hints of a decorated method.
+    This method is accessible as: signal.definition
+
+    If the signal is not defined by a method, it is expected to
+    have no arguments and signal.definition is equal to None.
+    """
+    def __init__(self, method=None):
+        super(dbus_signal, self).__init__()
+        self.definition = method
+
+
+def dbus_interface(interface_name):
+    """DBus interface.
+
+    A new DBus interface can be defined as:
+
+        @dbus_interface
+        class Interface():
+            ...
+
+    The interface will be generated from the given class cls
+    with a name interface_name and added to the DBus XML
+    specification of the class.
+
+    The XML specification is accessible as:
+        Interface.dbus
+    """
+    def decorated(cls):
+        generator = DBusSpecification()
+        cls.dbus = generator.generate_specification(cls, interface_name)
+        return cls
+    return decorated
+
+
+def dbus_class(cls):
+    """DBus class.
+
+    A new DBus class can be defined as:
+
+        @dbus_class
+        class Class(Interface):
+            ...
+
+    DBus class can implement DBus interfaces, but it cannot
+    define a new interface.
+
+    The DBus XML specification will be generated from
+    implemented interfaces (inherited) and it will be
+    accessible as:
+        Class.dbus
+    """
+    # Get the interface decorator without a name.
+    decorated = dbus_interface(None)
+    # Apply the decorator on the given class.
+    return decorated(cls)
+
+
+class DBusSpecificationError(Exception):
+    pass
+
+
+class DBusSpecification(object):
+    """Class for generating DBus XML specification."""
+
+    DIRECTION_IN = "in"
+    DIRECTION_OUT = "out"
+
+    ACCESS_READ = "read"
+    ACCESS_WRITE = "write"
+    ACCESS_READWRITE = "readwrite"
+
+    RETURN_PARAMETER = "return"
+
+    NAME_PATTERN = re.compile(r'[A-Z][A-Za-z0-9]*')
+
+    def __init__(self, xml_generator=XMLGenerator()):
+        self.xml_generator = xml_generator
+
+    def generate_specification(self, cls, interface_name=None):
+        """Generates DBus XML specification for given class.
+
+        If class defines a new interface, it will be added to
+        the specification.
+
+        :param cls: class object to decorate
+        :param str interface_name: name of the interface defined by class
+        :return str: DBus specification in XML
+        """
+        # Collect all interfaces that class inherits.
+        interfaces = self._collect_interfaces(cls)
+
+        # Generate a new interface.
+        if interface_name:
+            interface = self._generate_interface(cls, interfaces, interface_name)
+            interfaces[interface_name] = interface
+
+        # Generate XML specification for the given class.
+        node = self._generate_node(cls, interfaces)
+        return self.xml_generator.element_to_xml(node)
+
+    def _collect_interfaces(self, cls):
+        """Collect interfaces implemented by the class.
+
+        Returns a dictionary that maps interface names
+        to interface elements.
+
+        :param cls: a class object
+        :return: a dictionary of implemented interfaces
+        """
+        interfaces = dict()
+
+        # Visit cls and base classes in reversed order.
+        for member in reversed(inspect.getmro(cls)):
+            # Skip classes with no specification.
+            if not getattr(member, "dbus", None):
+                continue
+
+            # Update found interfaces.
+            node = self.xml_generator.xml_to_element(member.dbus)
+            node_interfaces = self.xml_generator.get_interfaces_from_node(node)
+            interfaces.update(node_interfaces)
+
+        return interfaces
+
+    def _generate_interface(self, cls, interfaces, interface_name):
+        """Generate interface defined by given class.
+
+        :param cls: a class object that defines the interface
+        :param interfaces: a dictionary of implemented interfaces
+        :param interface_name: a name of the new interface
+        :return: an new interface element
+
+        :raises DBusSpecificationError: if a class member cannot be exported
+        """
+        interface = self.xml_generator.get_interface_element(interface_name)
+
+        # Search class members.
+        for member_name, member in inspect.getmembers(cls):
+            # Check it the name is exportable.
+            if not self._is_exportable(member_name):
+                continue
+
+            # Skip names already defined in implemented interfaces.
+            if self._is_defined(interfaces, member_name):
+                continue
+
+            # Generate XML element for exportable member.
+            if self._is_signal(member):
+                element = self._generate_signal(member, member_name)
+            elif self._is_property(member):
+                element = self._generate_property(member, member_name)
+            elif self._is_method(member):
+                element = self._generate_method(member, member_name)
+            else:
+                raise DBusSpecificationError("%s.%s cannot be exported."
+                                             % (cls.__name__, member_name))
+
+            # Add generated element to the interface.
+            self.xml_generator.add_child(interface, element)
+
+        return interface
+
+    def _is_exportable(self, member_name):
+        """Is the name of a class member exportable?
+
+        The name is exportable if it follows the DBus specification.
+        Only CamelCase names are allowed.
+        """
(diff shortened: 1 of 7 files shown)


pyanaconda/dbus/typing.py
```diff
@@ -0,0 +1,182 @@
...
+
+    @staticmethod
+    def _check_dictionary(type_hint):
+        """Check the type of a dictionary.
+
+        Raises ValueError, if the type is not valid.
+        """
+        key, _ = type_hint.__args__
+
+        if DBusType._is_container_type(key) or key == Variant:
```
rvykydal: Shouldn't we allow tuple as a key? I think it can be quite a common case.

poncovka: DBus specification doesn't allow it:
> A DICT_ENTRY works exactly like a struct, but rather than parentheses it uses curly braces, and it has more restrictions. The restrictions are: it occurs only as an array element type; it has exactly two single complete types inside the curly braces; the first single complete type (the "key") must be a basic type rather than a container type. Implementations must not accept dict entries outside of arrays, must not accept dict entries with zero, one, or more than two fields, and must not accept dict entries with non-basic-typed keys. A dict entry is always a key-value pair. 

https://dbus.freedesktop.org/doc/dbus-specification.html#type-system

pyanaconda/dbus/interface.py
```diff
@@ -0,0 +1,323 @@
...
+
+        if not member.declaration:
+            return element
+
+        for name, type_hint, direction in self._iterate_parameters(member.declaration):
+            # Only input parameters can be defined.
+            if direction == self.DIRECTION_OUT:
+                raise DBusSpecificationError("Signal %s has defined return type." % member_name)
+
+            direction = self.DIRECTION_OUT
```
jkonecny12: I'm don't understand this part. Why you are testing if the direction is not OUT and then you set the direction to OUT?

pyanaconda/dbus/interface.py
```diff
@@ -0,0 +1,323 @@
...
+        :param member: a method object
+        :return: an iterator
+        """
+        # Get type hints for parameters.
+        type_hints = get_type_hints(member)
+
+        # Get parameters.
+        parameters = inspect.signature(member).parameters
+
+        for index, name in enumerate(parameters):
```
jkonecny12: You can replace the `enumerate` by `for index, name in parameters[1:]:`.

pyanaconda/dbus/interface.py
```diff
@@ -0,0 +1,323 @@
...
+from pyanaconda.dbus.xml import XMLGenerator
+
+__all__ = ["dbus_class", "dbus_interface", "dbus_signal"]
+
+
+class dbus_signal(signal):
+    """DBus signal."""
+    def __init__(self, method=None):
+        super(dbus_signal, self).__init__()
+        self.declaration = method
```
jkonecny12: What is the method/declaration here?

pyanaconda/dbus/interface.py
```diff
@@ -0,0 +1,323 @@
...
+            raise DBusSpecificationError("Property %s has invalid parameters." % member_name)
+
+        # Property has both.
+        if member.fget and member.fset:
+            access = self.ACCESS_READWRITE
+
+        if access is None:
+            raise DBusSpecificationError("Property %s is not accessible." % member_name)
+
+        return self.xml_generator.create_property(member_name, get_dbus_type(type_hint), access)
```
jkonecny12: Now you are giving `access` to `direction` parameter, it is little bit confusing.

pyanaconda/dbus/interface.py
```diff
@@ -0,0 +1,323 @@
...
+
+        # Is the return type other than None?
+        if type_hints[self.RETURN_PARAMETER] is None:
+            return
+
+        yield self.RETURN_PARAMETER, type_hints[self.RETURN_PARAMETER], self.DIRECTION_OUT
+
+    def _is_method(self, member):
+        """Is the class member a DBus method?"""
+        return inspect.ismethod(member) or inspect.isfunction(member)
```
jkonecny12: I don't think we want to support `functions`. It needs to be part of the class all the time if I understand usage correctly.

tests/pyanaconda_tests/dbus_interface_test.py
```diff
@@ -0,0 +1,606 @@
+# -*- coding: utf-8 -*-
```
jkonecny12: You can remove this line.

tests/pyanaconda_tests/dbus_interface_test.py
```diff
@@ -0,0 +1,606 @@
...
+#
+# Red Hat Author(s): Vendula Poncova <vponcova@redhat.com>
+#
+
+import unittest
+
+from pyanaconda.dbus.typing import *  # pylint: disable=wildcard-import
+from pyanaconda.dbus.xml import XMLGenerator
+from pyanaconda.dbus.interface import DBusSpecification, DBusSpecificationError, dbus_interface, dbus_class, \
+    dbus_signal
```
jkonecny12: Please fix the indentation.

tests/pyanaconda_tests/dbus_interface_test.py
```diff
@@ -0,0 +1,606 @@
...
+        self.assertFalse(self.generator._is_exportable("invalid_name"))
+        self.assertFalse(self.generator._is_exportable("invalid_name_123"))
+        self.assertFalse(self.generator._is_exportable("Invalid_Name"))
+        self.assertFalse(self.generator._is_exportable("invalidname"))
+        self.assertFalse(self.generator._is_exportable("invalid123"))
+        self.assertFalse(self.generator._is_exportable("invalidName"))
+        self.assertFalse(self.generator._is_exportable("123InvalidName"))
+
+    def is_method_test(self):
+        """Test if methods as DBus methods."""
```
jkonecny12: I guess there should be `are` instead of `as`.

rvykydal:
Looks good to me, reading comments from @jkonecny12, I don't have anything to add.

pyanaconda/dbus/typing.py
```diff
@@ -0,0 +1,182 @@
...
+    # DBus representation of container types.
+    _container_type_mapping = {
+        Tuple:      "(%s)",
+        List:       "a%s",
+        Dict:       "a{%s}",
+    }
+
+    @staticmethod
+    def _is_basic_type(type_hint):
+        """Is is a basic type?"""
```
poncovka: Fixed.

pyanaconda/dbus/typing.py
```diff
@@ -0,0 +1,182 @@
...
+        if origin == Dict:
+            DBusType._check_dictionary(type_hint)
+
+        # Generate string.
+        container = DBusType._container_type_mapping[origin]
+        items = [DBusType.get_dbus_representation(arg) for arg in args]
+        return container % "".join(items)
+
+    @staticmethod
+    def _check_dictionary(type_hint):
```
poncovka: Fixed.

pyanaconda/dbus/typing.py
```diff
@@ -0,0 +1,182 @@
...
+
+        Raises ValueError, if the type is not valid.
+        """
+        key, _ = type_hint.__args__
+
+        if DBusType._is_container_type(key) or key == Variant:
+            raise ValueError("Dictionary key cannot be of type %s." % key)
+
+    @staticmethod
+    def get_dbus_representation(type_hint):
```
poncovka: Fixed.

pyanaconda/dbus/typing.py
```diff
@@ -0,0 +1,182 @@
...
+        """Return a DBus representation of the given type hint.
+
+        :param type_hint: a type hint
+        :return str: a DBus representation of the type hint
+
+        Raises ValueError for unknown types.
```
poncovka: Fixed.

tests/pyanaconda_tests/dbus_typing_test.py
```diff
@@ -0,0 +1,170 @@
+# -*- coding: utf-8 -*-
```
poncovka: Fixed.

tests/pyanaconda_tests/dbus_xml_test.py
```diff
@@ -0,0 +1,75 @@
+# -*- coding: utf-8 -*-
```
poncovka: Fixed.

tests/pyanaconda_tests/dbus_xml_test.py
```diff
@@ -0,0 +1,75 @@
...
+            'direction="ParameterDirection" '
+            'name="ParameterName" '
+            'type="ParameterType" />')
+
+    def property_test(self):
+        """Test the property element."""
+        self._compare(
+            XMLGenerator.create_property("PropertyName",
+                                           "PropertyType",
+                                           "PropertyAccess"),
```
poncovka: Fixed.

pyanaconda/dbus/interface.py
poncovka: Added comment.

pyanaconda/dbus/interface.py
```diff
@@ -0,0 +1,323 @@
...
+    def decorated(cls):
+        generator = DBusSpecification()
+        cls.dbus = generator.generate_specification(cls, interface_name)
+        return cls
+    return decorated
+
+
+def dbus_class(cls):
+    """Generate DBus XML specification for given class."""
+    return dbus_interface(None)(cls)
```
poncovka: Fixed.

pyanaconda/dbus/interface.py
```diff
@@ -0,0 +1,323 @@
...
+    ACCESS_READWRITE = "readwrite"
+
+    RETURN_PARAMETER = "return"
+
+    NAME_PATTERN = re.compile(r'[A-Z][A-Za-z0-9]*')
+
+    def __init__(self, xml_generator=XMLGenerator()):
+        self.xml_generator = xml_generator
+
+    def generate_specification(self, cls, interface_name=None):
```
poncovka: Fixed.

pyanaconda/dbus/interface.py
```diff
@@ -0,0 +1,323 @@
...
+        """Is the member name defined in given interfaces?
+
+        :param interfaces: a dictionary of interfaces
+        :param member_name: a name of the class member
+        :return: True if the name is defined, otherwise False
+        """
+        return any(self.xml_generator.is_member(member)
+                   and self.xml_generator.has_name(member, member_name)
+                   for interface in interfaces.values()
+                   for member in interface)
```
poncovka: Fixed.

pyanaconda/dbus/interface.py
poncovka: Kind of fixed, because `parameters` is an ordered dictionary. 
Using: `for name in list(signature.parameters)[1:]: ...`

pyanaconda/dbus/interface.py
```diff
@@ -0,0 +1,323 @@
...
+
+        return interfaces
+
+    def _generate_interface(self, cls, interfaces, interface_name):
+        """Generate interface defined by given class.
+
+        :param cls: a class object that defines the interface
+        :param interfaces: a dictionary of implemented interfaces
+        :param interface_name: a name of the new interface
+        :return: an new interface element
```
poncovka: Fixed.

tests/pyanaconda_tests/dbus_interface_test.py
poncovka: Fixed.

tests/pyanaconda_tests/dbus_interface_test.py
poncovka: Fixed.

tests/pyanaconda_tests/dbus_interface_test.py
poncovka: Fixed.

pyanaconda/dbus/interface.py
poncovka: The DBus specification says, that a signal can have only OUT parameters. Since a signal emits its parameters and doesn't return anything, we should forbid the OUT parameter and transform IN parameters to OUT parameters.

Added comment.

pyanaconda/dbus/interface.py
poncovka: The method `create_property` has `access` parameter: 
`def create_property(name, property_type, access)`

pyanaconda/dbus/interface.py
poncovka: Added comment with an explanation.

pyanaconda/dbus/interface.py
```diff
@@ -0,0 +1,323 @@
...
+
+            # Check if the type is defined.
+            if name not in type_hints:
+                raise DBusSpecificationError("Parameter %s doesn't have defined type." % name)
+
+            yield name, type_hints[name], self.DIRECTION_IN
+
+        # Is the return type defined?
+        if self.RETURN_PARAMETER not in type_hints:
+            return
```
poncovka: That kind of check doesn't exist. Functions always return at least `None`. We can force programmers to always define a type hint for the return value and use `None` to indicate, it doesn't return anything.

For example:
```
def Function(self) -> None:
    pass
```

pyanaconda/dbus/interface.py
jkonecny12: Ahh sorry, I've looked on the `create_parameter` instead of `create_property` my bad.

pyanaconda/dbus/interface.py
jkonecny12: Interesting, thanks for explanation.

jkonecny12:
I don't know what changed but it still looks good to me.

PR #1132, opened 2017-07-27
rhel7: rpmostree + rhelah

cgwalters:


jkonecny12:
Hi,

for `rhel7-branch` you need to append ``Resolves: rhbz#<bug_number>`` ``Related: rhbz#<bug_number>`` to the commit message.

Thanks for PR 👍

cgwalters:
Ok, amended the commits.

jkonecny12:
Jenkins, it's ok to test.

cgwalters:
Any issues with merging this?

pyanaconda/install.py
```diff
@@ -253,24 +253,9 @@ def doInstall(storage, payload, ksdata, instClass):
         if iutil.getSysroot() != iutil.getTargetPhysicalRoot():
             blivet.setSysroot(iutil.getTargetPhysicalRoot(),
                               iutil.getSysroot())
-
```
jkonecny12: I guess this part is specific only to ostree install and I guess it is mainly because of ``if iutil.getSysroot() != iutil.getTargetPhysicalRoot():``.

If it is that case, could you please add the comment explaining why this is working to this commit.

I understand that this should be done by us some time ago when this was created, so or so it could be nice improvement of this commit.

cgwalters: Ok, I adjusted the commit message :arrow_down:

cgwalters:
Updated the second commit to reference a new BZ against the `anaconda` package, per request.

PR #1134, opened 2017-08-02
Add setup-test-env target to the Makefile

jkonecny12:
This target will just copy Anaconda to mock and install test requires.
You can then do a `mock -r mock_conf --shell` and play with it there.


M4rtinK:
Nice idea! :)

PR #1136, opened 2017-08-02
Don't mock modules with sys in unit tests

poncovka:
We shouldn't change `sys.modules `to mock modules. Use patch instead.
Bad example: `sys.modules["module_name"] = Mock()`


poncovka:
In files `anaconda/tests/pyanaconda_tests/ks_version_test.py` and  `/anaconda/tests/pyanaconda_tests/pwpolicy.py` are mocked modules `anaconda_logging` and `block`. Should I try to fix it as well?

poncovka:
I have changed the mentioned files as well and the tests passed.

PR #1137, opened 2017-08-04
Also capture anaconda-pre logs if they exist

jcpunk:
It appears that anaconda also can generate logs in /tmp/pre-anaconda-logs.  Add those to the archive as well.


jkonecny12:
Jenkins, it's ok to test.

M4rtinK:
Tested just now - indeed works correctly! :)

PR #1138, opened 2017-08-07
80-setfilecons: Add a few paths (/var/run, /var/spool)

cgwalters:
First, add a comment to the top of the file with my understanding of the purpose
of this file. According to `git annotate`, this code was originally intended
to handle relabeling for files that anaconda itself created.  (A better
fix would be to create files with the right label in the first place, but
that's a whole other topic)

Since then though, I think we've ended up doing ad-hoc relabeling for
things that users/admins do in `%post`.  It's really easy to have
the same problem with shell scripts there, although modern SELinux
does have automatic filename transitions which simplifies some cases.

However, the way OSTree is defined, it's categorically the installer's
job to label `/var`.  See <https://github.com/ostreedev/ostree/pull/872>.
I'd like to apply that PR; from a default FAH install, we just need to
fix the labels for `/var/run` and `/var/spool`.  So add those to the list.

It's tempting to change this code to do something like:
`if <is ostree>; then restorecon -r /var; fi` or so, but let's go
with the conservative fix for now.

### DIFF
--- a/data/post-scripts/80-setfilecons.ks
+++ b/data/post-scripts/80-setfilecons.ks
@@ -1,10 +1,18 @@
 %post
+# We need to handle SELinux relabeling for a few reasons:
+# - %post scripts that write files into places in /etc, but don't do
+#   labeling correctly
+# - Anaconda code that does the same (e.g. moving our log files into
+#   /var/log/anaconda)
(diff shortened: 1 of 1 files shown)


cgwalters:
Dependency of: https://github.com/ostreedev/ostree/pull/872

PR #1139, opened 2017-08-08
Use SHA256 instead of MD5 for repoMDHash (#1341280)

bcl:
MD5 use is not allowed when FIPS mode is enabled. Use SHA256 instead.

Related: rhbz#1341280


jkonecny12:
I didn't know that we are not able to use MD5 hashing. 
Thanks! 

I'll create rawhide compatible of this patch to master.

jkonecny12:
jenkins, test this please

PR #1143, opened 2017-08-11
Fix rescue touched unavailable installclass attr

jkonecny12:
In the rescue mode the `ksdata.installclass` attribute is not available.


poncovka:
The `installclass` command should be available anyway. I'm not able to reproduce the problem.

PR #1144, opened 2017-08-11
Add message to setup-test-env is ran

jkonecny12:
Info what you should call to attach to the test environment.


jkonecny12:
**UPDATED:**
Add also info where the directory of anaconda is placed

PR #1146, opened 2017-08-11
Prevent crash in GUI with incomplete IMSM array (#1465944)

M4rtinK:
Use a different method of device type check to prevent a crash
in the custom partitioning GUI if an incomplete IMSM array is present.

Thanks a lot to Artur Paszkiewicz for the providing the patch!

Resolves: rhbz#1465944


pyanaconda/ui/gui/spokes/custom.py
```diff
@@ -60,6 +60,7 @@
 from blivet.errors import NotEnoughFreeSpaceError
 from blivet.devicelibs import raid, crypto
 from blivet.devices import LUKSDevice
+from blivet.devices.md import import MDRaidArrayDevice
```
poncovka: Two imports :-)

PR #1148, opened 2017-08-14
Support for installclass command in RHEL 7.5 (#1412159)

poncovka:
The install class can be now specified in the kickstart file by its name 
with the `installclass` command in the `%anaconda` section.

Backported from: https://github.com/rhinstaller/anaconda/pull/1120
Resolves: rhbz#1412159

### DIFF
--- a/tests/pyanaconda_tests/installclass_test.py
+++ b/tests/pyanaconda_tests/installclass_test.py
@@ -1,0 +1,209 @@
+#
+# Vendula Poncova <vponcova@redhat.com>
+#
+# Copyright 2017 Red Hat, Inc.
+#
+# This copyrighted material is made available to anyone wishing to use, modify,
+# copy, or redistribute it subject to the terms and conditions of the GNU
+# General Public License v.2.  This program is distributed in the hope that it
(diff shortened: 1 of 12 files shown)


tests/pyanaconda_tests/installclass_test.py
```diff
@@ -0,0 +1,218 @@
...
+        class InstallClassB2(BaseInstallClass):
+            name = "Install Class B 2"
+            hidden = False
+            sortPriority = 3
+
+        with self._patch_collect([InstallClassA1, InstallClassA2, InstallClassB1, InstallClassB2]):
+            factory = self._get_fake_factory()
+            self.assertEqual(factory.classes, [InstallClassB2, InstallClassB1, InstallClassA2, InstallClassA1])
+
+            self.assertTrue(isinstance(factory.get_best_install_class(), InstallClassB2))
```
jkonecny12: It would be nice to have a comment why this is the best install class.

poncovka: The explanation is in the `get_best_install_class` method now.

poncovka:
The `installclass` command is part of the `%anaconda` section now.

PR #145, opened 2015-06-18
Fix adding 'boot=' option in FIPS mode

vojtechtrefny:
'boot=' option shouldn't be added when there is no separate /boot
partition.

Resolves: rhbz#1190146

Signed-off-by: Vojtech Trefny <vtrefny@redhat.com>

PR #1149, opened 2017-08-14
timezone: for kickstart allow also timezones not offered by GUI (#145…

rvykydal:
…2873)

Resolves: rhbz#1452873

For example Japan, which is linked to Asia/Tokyo in Olson tz database,
and we offer only Asia/Tokyo in GUI.


rvykydal:
For master, we should think about separating UI valid and (kickstart) valid timezones better in modules, ie shuffle stuff between spoke and timezone module, give better names to timezone module api methods etc.

PR #1152, opened 2017-08-17
Add blivet-gui logs to python-meh file list

vojtechtrefny:
Do not merge, https://github.com/storaged-project/blivet-gui/pull/83 needs to be merged (and built) before this.

### DIFF
--- a/pyanaconda/exception.py
+++ b/pyanaconda/exception.py
@@ -248,7 +248,8 @@ def initExceptionHandling(anaconda):
                  "/tmp/program.log", "/tmp/storage.log", "/tmp/ifcfg.log",
                  "/tmp/dnf.librepo.log", "/tmp/hawkey.log",
                  "/tmp/lvm.log", iutil.getSysroot() + "/root/install.log",
-                 "/proc/cmdline", "/root/lorax-packages.log"]
+                 "/proc/cmdline", "/root/lorax-packages.log",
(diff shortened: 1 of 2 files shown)


pyanaconda/exception.py
```diff
@@ -248,7 +248,8 @@ def initExceptionHandling(anaconda):
                  "/tmp/program.log", "/tmp/storage.log", "/tmp/ifcfg.log",
                  "/tmp/dnf.librepo.log", "/tmp/hawkey.log",
                  "/tmp/lvm.log", iutil.getSysroot() + "/root/install.log",
-                 "/proc/cmdline", "/root/lorax-packages.log"]
+                 "/proc/cmdline", "/root/lorax-packages.log",
+                 "/tmp/blivet-gui-utils.log"]
```
M4rtinK: Why `blivet-gui-utils.log` and not for example `blivet-gui.log` ?

vojtechtrefny: I plan to add a separate log for the GUI process later. This one is just for the daemon ("BlivetUtils") process.

M4rtinK: Oh, so it's name of the process ? I guess that makes sense - I think it would still be better if the process & log file were called Blivet GUI daemon/Blivet GUI backend (or something similar), but I thats probably a substantial amount of work for little gain, at least for now.

PR #1153, opened 2017-08-17
Rhel7 fix repositories availability errors

jkonecny12:
Fix a problem with a proxy when testing repositories availability.

Also in a second commit fix that the Network spoke froze when this testing was in progress.

*Resolves: rhbz#1478970*

### DIFF
--- a/pyanaconda/ui/gui/spokes/network.py
+++ b/pyanaconda/ui/gui/spokes/network.py
@@ -1526,16 +1526,13 @@ class NetworkSpoke(FirstbootSpokeMixIn, NormalSpoke):
 
         if self.networking_changed:
             if ANACONDA_ENVIRON in anaconda_flags.environs and self.payload.needsNetwork:
-                log.debug("network spoke (apply) refresh payload")
                 from pyanaconda.packaging import payloadMgr
-                if not self.payload.verifyAvailableRepositories():
-                    payloadMgr.restartThread(self.storage, self.data, self.payload, self.instclass,
-                                             fallback=not anaconda_flags.automatedInstall)
-                else:
-                    log.debug("Payload isn't restarted, repositories are still available.")
+                payloadMgr.restartThread(self.storage, self.data, self.payload, self.instclass,
+                                         fallback=not anaconda_flags.automatedInstall, onlyOnChange=True)
+                self.networking_changed = False
             else:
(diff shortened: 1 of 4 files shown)


pyanaconda/ui/gui/spokes/network.py
```diff
@@ -1525,16 +1525,9 @@ def apply(self):
         log.debug("network: apply ksdata %s", self.data.network)
 
         if self.networking_changed:
-            if ANACONDA_ENVIRON in anaconda_flags.environs and self.payload.needsNetwork:
```
rvykydal: I don't see this conditions applied in the patched version.

jkonecny12: I totally missed that. Thanks!
Fixed.

jkonecny12:
**Updated:**
Fix issue pointed out by @rvykydal .

pyanaconda/ui/gui/spokes/network.py
```diff
@@ -1526,16 +1526,13 @@ def apply(self):
...
-                else:
-                    log.debug("Payload isn't restarted, repositories are still available.")
+                payloadMgr.restartThread(self.storage, self.data, self.payload, self.instclass,
+                                         fallback=not anaconda_flags.automatedInstall, onlyOnChange=True)
+                self.networking_changed = False
             else:
-                log.debug("network spoke (apply), payload refresh skipped (running outside of installation environment)")
-            self.networking_changed = False
+                log.debug("network spoke (apply), payload refresh skipped "
+                          "(running outside of installation environment)")
```
rvykydal: A small nitpick, the message is not quite correct - and was not before, but while on that we could fix it.
The thing is that the restart can be skipped also in case of self.payload.needsNetwork == False and the log message will be still written out.

Also, shouldn't we unset networking_changed (as before the patch) also if the restart is not needed based on the condition(s) so we don't think that networking has changed next time we visit the spoke ?
I don't know, maybe move the needsNetwork condition one new if level up?
```
if self.networking_changed:
    if self.payload.needsNetwork:
        if ANACONDA_ENVIRON ...
            ...
        else:
            log.debug(... the same message ...)
    self.networking_changed = False
```

jkonecny12: True and true. I'm going to create new patch for this.

PR #1154, opened 2017-08-17
Add support for the RHV branded help content variant (#1378010)

M4rtinK:
Show the RHV (Red Hat Virtualization) branded help content using the RHV
install class is being used.

The default help content also moved to the rhel subfolder, so also
account for that and bump the required version of the help content
package.

Resolves: rhbz#1378010

### DIFF
--- a/pyanaconda/installclass.py
+++ b/pyanaconda/installclass.py
@@ -67,7 +67,7 @@ class BaseInstallClass(object):
     default_autopart_type = DEFAULT_AUTOPART_TYPE
 
     # help
-    help_folder = "/usr/share/anaconda/help"
+    help_folder = "/usr/share/anaconda/help/rhel"
     help_main_page = "Installation_Guide.xml"
     help_placeholder = None
     help_placeholder_with_links = None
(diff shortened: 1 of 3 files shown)


M4rtinK:
NOTE: New version of the RHEL7 anaconda-user-help package needs to be released before this pull request can be merged.

pyanaconda/installclass.py
```diff
@@ -64,7 +64,7 @@ class BaseInstallClass(object):
     default_autopart_type = DEFAULT_AUTOPART_TYPE
 
     # help
-    help_folder = "/usr/share/anaconda/help"
+    help_folder = "/usr/share/anaconda/help/rhel"
```
poncovka: Maybe this change should be in `RHELBaseInstallClass` instead of `BaseInstallClass`.

M4rtinK: The RHEL content is basically the default help content for all variants, unless it specifies its own, so that's why it's defined this in the `BaseInstallClass`.

It could be also possible to put the help content to `default` subfolder, but having it in a `rhel` subfolder seemed more self-describing to me, as it is really just the converted RHEL installation guide without any variant specific changes that serves as the default help content.

poncovka: Ok.

rvykydal:
This will be ported to master with https://github.com/rhinstaller/anaconda/pull/2432

rvykydal:
Ported to master with https://github.com/rhinstaller/anaconda/pull/2432

PR #1156, opened 2017-08-17
Fix SL install class to use right efi dir

jcpunk:
Turns out when I initially committed the efi dir, I did it wrong.  This is the correct EFI dir for Scientific Linux.


jkonecny12:
Jenkins, it's ok to test.

PR #1159, opened 2017-08-21
Rhel7 branch add network bindto mac support

rvykydal:


dracut/parse-kickstart
```diff
@@ -88,6 +88,16 @@ def first_device_with_link():
 def setting_only_hostname(net, args):
     return net.hostname and (len(args) == 2 or (len(args) == 3 and "--hostname" in args))
 
+def nm_type_of_device(dev):
+    # uapi/linux/if_arp.h
+    arp_hw_id = readsysfile("/sys/class/net/%s/type" % dev)
+    if arp_hw_id == "1":
+        return "Ethernet"
+    elif arp_hw_id == "32":
+        return "Infiniband"
```
jkonecny12: Could these be constants please.

pyanaconda/network.py
```diff
@@ -688,6 +702,15 @@ def add_connection_for_ksdata(networkdata, devname):
...
     return added_connections
 
+def connection_binding_setting(devname, bindto):
+    if bindto == BIND_TO_MAC:
+        hwaddr = nm.nm_device_perm_hwaddress(devname)
+        hwaddr = [int(b, 16) for b in hwaddr.split(":")]
+        setting = ['802-3-ethernet', 'mac-address', hwaddr, 'ay']
+    else:
+        setting = ['connection', 'interface-name', devname, 's']
+    return setting
```
jkonecny12: Wouldn't it be better to have two methods here:

`connection_binding_setting()` and `connection_binding_to_mac_setting()` instead of the flag passed in?

pyanaconda/nm.py
```diff
@@ -691,19 +693,23 @@ def _device_settings(name):
 
             settings = _find_settings(vlanid, 'vlan', 'id')
     else:
+        # device name bound settings
         settings = _find_settings(name, 'connection', 'interface-name')
+        # mac address bound settings
+        try:
+            hwaddr_str = nm_device_valid_hwaddress(name)
+        except PropertyNotFoundError:
+            pass
```
jkonecny12: It would be nice to have a debug log here.

dracut/parse-kickstart
rvykydal: Sure

pyanaconda/network.py
rvykydal: Why?

pyanaconda/nm.py
rvykydal: Agreed

rvykydal:
jenkins, test this please

pyanaconda/network.py
jkonecny12: Because it is the same right now. This is just a composition of two functions. The only line which will be called in both cases is the condition and this condition can be also outside or somewhere else. And in general it is simpler to test separated functions than composition of functions.

So when the code is not shared it is more descriptive to have these as two methods which information what is changed.

Of course this is more recommendation than requirement.

rvykydal: Then they should be named
connection_binding_to_interface_name_setting() and connection_binding_to_mac_setting()
or better
bind_to_interface_name_connection_setting() and bind_to_mac_connection_setting()
or
bind_to_interface_name_setting() and bind_to_mac_setting()

Having two such functions doesn't make sense to me.

If I wanted to move the bindto argument branching one level up out of the function, I'd just create a function to find and mangle hwaddress and assign the 'setting' value directly as before, instead of calling functions.

The purpose of the connection_binding_setting function is: give me connection_binding_setting for this bindto parameter. It seems still valid to me.

rvykydal:
Updated required pykickstart version in anaconda.spec.in

M4rtinK:
jenkins, test this please

pyanaconda/network.py
jkonecny12: That was only suggestion so I'm fine with both solutions. You can leave it this way if you want.

rvykydal:
jenkins, test this please

rvykydal:
jenkins, test this please

rvykydal:
jenkins, test this please

rvykydal:
Ported to master with https://github.com/rhinstaller/anaconda/pull/1181/files

PR #1162, opened 2017-08-22
Add the boot option inst.xtimeout (#1462035)

poncovka:
Added a new boot option inst.xtimeout to set the timeout in seconds
for starting X.

Resolves: rhbz#1462035


jkonecny12:
Also it could be valuable to have kickstart test for this.

poncovka:
Fixed.

PR #1165, opened 2017-08-23
Remove the title bar in anaconda by default (#1468801)

poncovka:
By default, anaconda window doesn't need to be decorated at all,
because it is always maximized. It doesn't have to work for
some window managers, but most of the cases are covered.

With the inst.decorated option, the window in GUI will be decorated,
so it will have the title bar, resize controls, etc.

The metacity theme was removed, because it cannot be used to hide
the title bar anymore and anaconda uses a different workaround now.

Resolves: rhbz#1468801


data/anaconda_options.txt
```diff
@@ -258,3 +258,7 @@ Enable legacygrub (for installing older OSes). You probably do not want this.
 waitfornet
 Wait for network connectivity at the beginning of the installation with the
 timeout in seconds specified as a mandatory value of the option.
+
+decorated
+Run GUI installer in a decorated window. By default, the window is not decorated, so it doesn't have a title bar,
+resize controls, etc.
```
M4rtinK: A similar note should be also added to the boot option docs (`docs/boot-options.rst`) so that the new boot option also shows up in the Anaconda documentation on Read The Docs.

(I guess in the future it might make sense to generate that document from argparse output the same way Pykickstart does it for kickstart commands, so we don't need to keep multiple places in sync.)

poncovka: Fixed.

PR #1168, opened 2017-08-29
Branching instructions & improvement

M4rtinK:
Add options to `makebumpver` for easy & robust Rawhide & next Fedora Anaconda version bumps and document the whole branching process.

### DIFF
--- a/docs/release.rst
+++ b/docs/release.rst
@@ -195,3 +195,129 @@ Once Fedora enters a freeze:
 Once Fedora exits the freeze:
 
 - drop the downstream patches and do merge based releases as before
+
+
+Branching for the next Fedora release
+=====================================
+
+Anaconda uses separate branches for each Fedora release to make parallel Anaconda development for Rawhide and next Fedora possible.
+The branches are named like this:
+
+- f<number>-devel
+- f<number>-release
+
+The ``-devel`` branch is where code changes go and it is periodically merged to the master branch.
+The ``-release`` branch contains release commits and any Fedora version specific hotfixes.
+
+How to branch Anaconda
+----------------------
+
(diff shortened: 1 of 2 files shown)


docs/release.rst
```diff
@@ -195,3 +195,129 @@ Once Fedora enters a freeze:
...
+
+4. On the new page ``version id`` should be ``f<version>`` and make sure ``Copy from previous version`` is ticked
+
+5. Wait till the new branch is created.
+
+How to bump Rawhide Anaconda version
+------------------------------------
+
+- major version becomes major version +1
+- minor version is set to 1
```
jkonecny12: I would use here ``+1`` instead of +1

scripts/makebumpver
```diff
@@ -206,9 +214,15 @@ class MakeBumpVer:
...
+    def _incrementVersion(self, bump_major_version=False, add_version_number=False):
         fields = self.version.split('.')
-        fields[-1] = str(int(fields[-1]) + 1)
+        if add_version_number:  # add another version number to the end
+            fields.append(str(int(1)))
+        elif bump_major_version:  # increment major version
+            fields[0] = str(int(fields[0]) + 1) # bump major version
+            fields[-1] = str(int(1)) # reset minor version to 1
+        else:  # minor version bump
+            fields[-1] = str(int(fields[-1]) + 1)
```
jkonecny12: Could you please change these numbers to constants.

docs/release.rst
M4rtinK: Fixed.

scripts/makebumpver
M4rtinK: Sure. :)

PR #1170, opened 2017-09-01
Use /usr/bin/python3 shebang once again

hroncok:
This revert changes from b23760dcac91fcc77cc629d3e87757a8b4b6aae0

System Python was removed,
see https://fedoraproject.org/wiki/Changes/Platform_Python_Stack

Currently, /usr/libexec/system-python is just a symbolic link to /usr/bin/python3

### DIFF
--- a/anaconda.py
+++ b/anaconda.py
@@ -1,4 +1,4 @@
-#!/usr/libexec/system-python
+#!/usr/bin/python3
 #
 # anaconda: The Red Hat Linux Installation program
 #


hroncok:
Note that this is relevant for Fedora 27+

M4rtinK:
> This revert changes from b23760d
> 
> System Python was removed,
> see https://fedoraproject.org/wiki/Changes/Platform_Python_Stack

So do I understand it correctly that unlike System Python Platform Python will only be used by DNF ? So Anaconda will switch back to "Normal Python" and that would be it ? (eq. no switch to Platform Python once it's deployed)

We do have the DNF paylod which runs both in the controlled installation environment (during a normal installation) and on potentially arbitrary installed systems (during image generation), but I guess that could be out of scope as it's not really "the dnf" & worst case would break someones attempt for custom image generation.

> 
> Currently, /usr/libexec/system-python is just a symbolic link to /usr/bin/python3

hroncok:
> So do I understand it correctly that unlike System Python Platform Python will only be used by DNF ? So Anaconda will switch back to "Normal Python" and that would be it ? (eq. no switch to Platform Python once it's deployed)

Yes, Platform Python is for the Modularity Platform only and Anaconda is not part of that AFAIK. (Also, it cannot be, because it has plenty of Python deps not available in Platform.)

> We do have the DNF paylod which runs both in the controlled installation environment (during a normal installation) and on potentially arbitrary installed systems (during image generation), but I guess that could be out of scope as it's not really "the dnf" & worst case would break someones attempt for custom image generation.

Actually only dnf in Platform runs on Platform Python. In "normal" Fedora, dnf runs on `/usr/bin/python3`.

M4rtinK:
jenkins, test this please

PR #1171, opened 2017-09-01
Fix catch TUI not main thread exceptions

jkonecny12:
Exception which are not in a main thread was passed to the `hubQ` which is not used anymore in Simpleline. To resolve this issue, exceptions are now enqueued as `ExceptionSignal`.

This must be merged after https://github.com/rhinstaller/python-simpleline/pull/39

### DIFF
--- a/pyanaconda/exception.py
+++ b/pyanaconda/exception.py
@@ -17,34 +17,41 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #
-from meh import Config
-from meh.handler import ExceptionHandler
-from meh.dump import ReverseExceptionDump
-from pyanaconda import iutil, kickstart
-import sys
-import os
-import shutil
-import time
-import re
 import errno
 import glob
-import traceback
-import blivet.errors
-from pyanaconda.errors import NonInteractiveError
-from pyanaconda.ui.communication import hubQ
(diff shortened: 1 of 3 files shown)


pyanaconda/exception.py
```diff
@@ -190,9 +198,14 @@ def handleException(self, dump_info):
                 # data and let message handler run the exception handler in
                 # the main thread
                 exc_info = dump_info.exc_info
-                hubQ.send_exception((exc_info.type,
-                                     exc_info.value,
-                                     exc_info.stack))
+                if App.is_initialized():
```
M4rtinK: I think it might make sense to document around here that this is basically a split between handling exception via TUI and GUI.

Also I wonder - what about the ask-vnc-question TUI screen ? Could that result in Simpleline App being initialized while the the user is actually using the GUI over VNC ?

jkonecny12: I've added comment to explain why this `if` is here. If I'm not mistaken the `ask-vnc-question` screen is not using python-meh and if it is using python-meh, then the Simpleline would be initialized so or so.

jkonecny12:
**UPDATED:**
* Added comment about why we are looking if initialized.

jkonecny12:
Jenkins, it's ok to test.

jkonecny12:
jenkins, test this please

PR #1172, opened 2017-09-01
Add efi_dir to the BaseInstallClass (#1412391)

poncovka:
Attribute efi_dir should be defined in BaseInstallClass as well.

Also, it doesn't make sense to use hidden install classes. If we cannot
use a visible install class, we should raise an exception.

Resolves: rhbz#1412391

### DIFF
--- a/pyanaconda/installclass.py
+++ b/pyanaconda/installclass.py
@@ -59,6 +59,9 @@ class BaseInstallClass(object):
 
     _l10n_domain = None
 
+    # EFI directory.
+    efi_dir = "default"
+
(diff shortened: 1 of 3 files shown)


poncovka:
jenkins, test this please

jkonecny12:
Jenkins, it's ok to test.

M4rtinK:
jenkins, test this please

jkonecny12:
Please fix the failing test.

poncovka:
Fixed tests and added a default install class.

PR #1173, opened 2017-09-04
TUI progress reporting is handled by show_all

jkonecny12:
The TUI progress reporting was handled in the `refresh()` method, however, this should be handled by the `show_all()` method.

This is fixing problem with separator printed after the progress finished. Also we don't need to print name of the `Screen` manually.


jkonecny12:
Jenkins, it's ok to test.

jkonecny12:
jenkins, test this please

M4rtinK:
jenkins, test this please

M4rtinK:
jenkins, test this please

PR #1176, opened 2017-09-06
Make EFIGRUB._efi_binary a property, not a method

AdamWill:
This was previously a class attribute and still is one for the
other classes, and the code that uses it treats it as such. So
for EFIGRUB we need to make it a property. Without this fix,
UEFI installs crash with "TypeError: must be str, not method".


AdamWill:
@vathpela

PR #1178, opened 2017-09-10
Add missing dot to the availability status message

AsciiWolf:
Just a small cosmetic fix. ;-)

PR #1180, opened 2017-09-11
Missing dot in testing availability message

jkonecny12:
Port of the PR https://github.com/rhinstaller/anaconda/pull/1178 back to rhel7

*Related: rhbz#1478970*

*Suggested-by: AsciiWolf*


jkonecny12:
jenkins, test this please

AsciiWolf:
Would it be possible to fix this also in the f27 branch? Thanks!

jkonecny12:
I don't think we want to add this change, which is not breaking anything, after beta and with string freeze already behind us. It will be contained in F28.

PR #1184, opened 2017-09-13
Migrate fedora-server installclass into the anaconda source repositor…

rvykydal:
…y (#1466967)

We can remove it from fedora-productimg-server then.


rvykydal:
Wrong target branch, closing.

PR #1191, opened 2017-09-15
Log chroot in exec with redirect

jkonecny12:
This behavior was requested in mailing list and it will be useful even for us.

As a part of this change I've also added new blank line before top level functions and classes in iutil.py to do another step in pep8 migration.

### DIFF
--- a/pyanaconda/iutil.py
+++ b/pyanaconda/iutil.py
@@ -64,6 +64,7 @@ from pykickstart.constants import KS_SCRIPT_ONERROR
 
 _child_env = {}
 
+
 def setenv(name, value):
     """ Set an environment variable to be used by child processes.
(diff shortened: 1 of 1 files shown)


pyanaconda/iutil.py
```diff
@@ -193,7 +193,10 @@ def preexec():
             preexec_fn()
 
     with program_log_lock:
-        program_log.info("Running... %s", " ".join(argv))
+        if root:
```
rvykydal: Seems like this will be almost always true as default value for root is '/' and in this case we don't do chroot?

jkonecny12:
**UPDATED:**
* print `running in chroot` only when chroot is not '/'

PR #1193, opened 2017-09-21
Fix missing environment translation

jkonecny12:
Regression from Anaconda to Simpleline migration.

Resolves: rhbz#1491119


jkonecny12:
jenkins, test this please

M4rtinK:
jenkins, test this please

jkonecny12:
**UPDATED**:
* Add new commit which fix the race condition 
    * After this commit the original exception happens on all products. As a bonus it is fixing issue when the environment with index 0 wasn't selected at all.

PR #1194, opened 2017-09-21
Fix the race between StorageCheckHandler and updateCompleteness in ma…

KonstantinVolkov:
…in hub

This happens approximately in 1 of 100 deployment with next conditions:

- extremely slow storage IO
- kickstart installation with slightly reduced min_partition_size which causes
  appropriate warning

The StorageCheckHandler reports than StorageSpoke is ready, then start
logger for the warning. In separate thread hubQ got HUB_CODE_READY,
but StorageSpoke is not completed because of THREAD_CHECK_STORAGE is not
finished. So, installation stops on main Hub with the warning.

### DIFF
--- a/pyanaconda/ui/helpers.py
+++ b/pyanaconda/ui/helpers.py
@@ -98,8 +98,8 @@ class StorageCheckHandler(object):
         StorageCheckHandler.errors = report.errors
         StorageCheckHandler.warnings = report.warnings
 
-        hubQ.send_ready(self._mainSpokeClass, True)
         report.log(self.log)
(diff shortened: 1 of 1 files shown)


jkonecny12:
jenkins, test this please

rvykydal:
Thank you for investigating and the patch.
Actually I've been looking into this issue lately seeing the race running kickstart tests in kvm. I believe your fix should minimize the chance of hitting the race hugely but I tried to be even more defensive:
https://github.com/rhinstaller/anaconda/pull/1198

KonstantinVolkov:
OK, looks like #1198 is much better.

PR #1195, opened 2017-09-22
Deselect encryption when switching to blivet-gui partitioning

vojtechtrefny:
Hiding the encryption checkbox when switching to blivet-gui is
not enough, we need also to deselect/deactivate the checkbox.


M4rtinK:
jenkins, test this please

PR #1196, opened 2017-09-22
tui source spoke: initialize nfs values when switching to nfs

rvykydal:
Otherwise going to NFS and continuing without setting any values causes
traceback.


jkonecny12:
jenkins, test this please

PR #1197, opened 2017-09-22
Use build-time release (#1493952)

M4rtinK:
As we shown the Anaconda version including the
release we really should use the release as it
exists at the package build time instead of
the release at tarball creation time (which
is generally always 1).

Otherwise we would be lying to users every time
the Anaconda package has been rebuild without
changing the tarball (patch only rebuilds,
mass rebuilds, modular builds, etc.).

So ad a new configure variable that can be used
from the spec file at build time and use it
to set the built-time release version.

NOTE: The release version might not be a number,
but basically an arbitrary string as with
the modular builds.

Resolves: rhbz#1493952

### DIFF
--- a/anaconda.spec.in
+++ b/anaconda.spec.in
@@ -241,7 +241,8 @@ runtime on NFS/HTTP/FTP servers or local disks.
 %setup -q
 
 %build
-%configure
+# use actual build-time release number, not tarball creation time release number
+%configure ANACONDA_RELEASE=%{release}
 %{__make} %{?_smp_mflags}
 
 %install
--- a/configure.ac
+++ b/configure.ac
@@ -20,6 +20,15 @@ m4_define(python_required_version, 3.4)
 AC_PREREQ([2.63])
 AC_INIT([anaconda], [26.1], [anaconda-devel-list@redhat.com])
 
+# make it possible to set build info at build time
+# (patch only builds, modular builds, mass-rebuilds, etc.)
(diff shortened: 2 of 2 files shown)


M4rtinK:
Hmm, looks like this broke some weird hack in makebumpver. Good thing we have CI for stuff like this. :)

jkonecny12:
Please fix this:
```
  File "scripts/makebumpver", line 676, in <module>
    mbv = MakeBumpVer()
  File "scripts/makebumpver", line 87, in __init__
    self.current_release = re.search(regexp, config_ac).groups()[0]
AttributeError: 'NoneType' object has no attribute 'groups'
```

M4rtinK:
> However, is this handling --version? If so, should we really show release of the package? I would guess this is spec file specific and I would rather see it completely removed.

Yeah, that's a good point - it really should not show up for --version. Rather than that I guess we can save it (and possibly other data about the build) to a different variable in `version.py.in` and just log that at runtime.

> Please fix this:
> 
> Traceback (most recent call last):
>   File "scripts/makebumpver", line 676, in <module>
>     mbv = MakeBumpVer()
>   File "scripts/makebumpver", line 87, in __init__
>     self.current_release = re.search(regexp, config_ac).groups()[0]
> AttributeError: 'NoneType' object has no attribute 'groups'

I think I'll just drop the release version support from `makebumpver` - it also doesn't make sense to have it there as well.

M4rtinK:
Obsoleted by https://github.com/rhinstaller/anaconda/pull/1268.

PR #1204, opened 2017-09-27
Fix `shim` and `grub2-efi` to be backwards compatible

### LINKED ISSUE #1202: handle better the case where anaconda (master) can't find shim-x64 when installing from repo that contains old shim package
hi,

we are using anaconda from master to build squashfs.img.

when we try to install a system using Centos 7.3 repo the system cannot find `shim-x64`. we are in the process of switching to 7.4 but i will take a couple of weeks.

Can you please modify the tool to handle this package name transition? try `shim-x64` first, and if not there try just `shim`.

jaymzh:
`shim` was renamed to `shim-x64` in RHEL 7.4 and Rawhide. Similarly,
`grub2-efi` was renamed to `grub2-efi-x64`. Recently Anaconda started
using these new names, but this breaks backwards compatibility. This PR
moves back to the original names which *is still forward compatible*:

```
$ rpm -q --provides grub2-efi-x64
...
grub2-efi = 1:2.02-0.64.el7.centos
...

$ rpm -q --provides shim-x64
...
shim = 12-1.el7.centos
...
```

Fixes #1202

### DIFF
--- a/pyanaconda/bootloader.py
+++ b/pyanaconda/bootloader.py
@@ -1824,7 +1824,7 @@ class EFIGRUB1(EFIBase, GRUB):
 
 class EFIGRUB(EFIBase, GRUB2):
     _packages32 = [ "grub2-efi-ia32", "shim-ia32" ]
-    _packages64 = [ "grub2-efi-x64", "shim-x64" ]
+    _packages64 = [ "grub2-efi", "shim" ]
     _packages_common = [ "efibootmgr" ]
     can_dual_boot = False
     stage2_is_valid_stage1 = False


jkonecny12:
Jenkins, it's ok to test.

cgwalters:
Yes, but the problem is there is *also* a 32 bit provider (that's the idea), and if you install via Provides it's a roll of the dice 🎲 which one you get.  See https://pagure.io/fedora-atomic/pull-request/84

If we want new Anaconda to be able to install old systems (and that makes sense to me), I think it's going to have to be something like "look for -x64 version, fall back".

```
$ docker run --rm -ti registry.fedoraproject.org/fedora:27 
# dnf makecache
...
# dnf whatprovides grub2-efi
grub2-efi-ia32-1:2.02-18.fc27.x86_64 : GRUB for EFI systems.
Repo        : updates-testing

grub2-efi-x64-1:2.02-18.fc27.x86_64 : GRUB for EFI systems.
Repo        : updates-testing

grub2-efi-ia32-1:2.02-18.fc27.x86_64 : GRUB for EFI systems.
Repo        : fedora

grub2-efi-x64-1:2.02-18.fc27.x86_64 : GRUB for EFI systems.
Repo        : fedora
# 

```

jaymzh:
@cgwaltersdw Oh, weird.

So I'm happy to add try one and fall back, but I don't see anyway to do it in that layer...

cgwalters:
Honestly, I don't know how to do it here either without a bigger reworking of the code.  It'd probably be something like:

`_packages64 = [('grub2-efi-x64', 'grub2-efi'), ('shim-x64', 'shim')]` where for each tuple entry the package first is used if it exists, otherwise fall back to the second.  (Obviously the harder part is fixing the whole `@property packages()` thing to handle this)

In the short term I think you (and everyone else) are just going to have to use the installer associated with the major/minor.

vathpela:
Okay, so 2 questions (and maybe somebody knows the answer or can just test what I'm thinking):
1) does anaconda honor the "${NAME}.${ARCH}" convention on the packages properties?
2) does it error if something there is not found?

If the answers are "yes" and "no", then for the _packages64 version you can do ["grub2-efi-x64", "shim-x64", "grub2-efi.x86_64", "shim.x86_64"] and that should get the desired result.

cgwalters:
Actually if `1` is true (and it probably is), we can just unconditionally change it to
`_packages64 =  ["grub2-efi.x86_64", "shim.x86_64"]` I think.

vathpela:
I don't think so?  Provides aren't arched that way AIUI.

InStallker:

InStallker:

M4rtinK:
jenkins, test this please

martinezjavier:
@jaymzh are you still interested in this PR or can be closed? Do you plan to add the fallback logic ? Cc @jkonecny12 @poncovka

jaymzh:
Facebook stopped using Anaconda, and I'm no longer at FB, so I'm probably not the person to finish this up - and no one was even able to give me any sense of how to do it properly. :/

Also, with 9 about to come out, this probably isn't as-relevant.

I know that the Anaconda team liked the idea of Rawhide anaconda being able to install more than one OS, so being able to fix refactor the code a bit to handle something like this seems like a good idea, but that's gonna be up to you guys.

jkonecny12:
> Facebook stopped using Anaconda, and I'm no longer at FB, so I'm probably not the person to finish this up - and no one was even able to give me any sense of how to do it properly. :/
> 
> Also, with 9 about to come out, this probably isn't as-relevant.
> 
> I know that the Anaconda team liked the idea of Rawhide anaconda being able to install more than one OS, so being able to fix refactor the code a bit to handle something like this seems like a good idea, but that's gonna be up to you guys.

I like the idea, however, seems there are no users and we don't have time to implement everything so I don't think it's worth the effort now. We can return back to this in the future if there will be requirement.

Thanks @jaymzh for your reply and I wish you good luck in your current position. I hope we will meet again on some conference in the future :).

jaymzh:
That's fair. But CC @jaymzh and @davide125 just for heads up.

davide125:
@jaymzh is correct, and FB production doesn't currently use Anaconda. However, the desktop fleet does (https://github.com/facebookincubator/linux-desktop-kickstarts), and this could still be useful there, cc @michel-slm @steelcowboy

PR #1207, opened 2017-10-05
Enable SE/HMC file access to repo (#1289918)

poncovka:
Provide access to files on the USB/DVD drive of the SE/HMC of System
z hardware with a boot option: inst.repo=hmc

Resolves: rhbz#1289918

Depends on: https://github.com/rhinstaller/pykickstart/pull/179
Depends on: https://github.com/rhinstaller/lorax/pull/251

### DIFF
--- a/dracut/anaconda-hmcroot
+++ b/dracut/anaconda-hmcroot
@@ -1,0 +1,24 @@
+#!/bin/bash
+# anaconda-hmcroot:
+# Provide access to the root on the USB/DVD drive of the SE/HMC of System z hardware.
+
+command -v info >/dev/null || . /lib/dracut-lib.sh
+command -v anaconda_live_root_dir >/dev/null || . /lib/anaconda-lib.sh
+
+# Debug.
+debug_msg "Trying SE/HMC access with $1."
+
+# Do we already have a root device?
+# Then do not run again.
+[ -e "/dev/root" ] && exit 1
+
+# Do we have SE/HMC file access?
+if /usr/sbin/lshmc ; then
+    info "Anaconda using SE/HMC file access to root."
+    /usr/bin/hmcdrvfs $repodir || warn "Couldn't mount /dev/hmcdrv"
+    anaconda_live_root_dir $repodir
+else
+    debug_msg "No SE/HMC access to files."
+    exit 1
+fi
+
--- a/dracut/anaconda-modprobe.sh
+++ b/dracut/anaconda-modprobe.sh
@@ -19,6 +19,8 @@ done
 
 if [ "$ARCH" != "s390" -a "$ARCH" != "s390x" ]; then
     MODULE_LIST+=" floppy edd iscsi_ibft "
+else
+    MODULE_LIST+=" hmcdrv "
 fi
 
 if [ "$ARCH" = "ppc" ]; then
@@ -30,6 +32,9 @@ MODULE_LIST+=" raid0 raid1 raid5 raid6 raid456 raid10 linear dm-mod dm-zero  \
               sha256 lrw xts "
 
 for m in $MODULE_LIST; do
-    modprobe $m &>/dev/null
+    if modprobe $m ; then
+        debug_msg "$m was loaded"
+    else
+        debug_msg "$m was NOT loaded"
+    fi
 done
-
--- a/dracut/module-setup.sh
+++ b/dracut/module-setup.sh
@@ -14,6 +14,12 @@ depends() {
     return 0
 }
 
+installkernel() {
+    case "$(uname -m)" in
+        s390*) instmods hmcdrv ;;
+    esac
+}
+
 install() {
     # binaries we want in initramfs
     dracut_install eject -o pigz
@@ -66,5 +72,14 @@ install() {
             *) inst $dep ;;
         esac
     done
+
+    # support for specific architectures
+    case "$(uname -m)" in
+        s390*)
+            inst "/usr/sbin/lshmc"
+            inst "/usr/bin/hmcdrvfs"
+            inst "$moddir/anaconda-hmcroot" "/sbin/anaconda-hmcroot"
+        ;;
+    esac
 }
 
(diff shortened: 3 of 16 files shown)


dracut/anaconda-hmcroot
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash
+# anaconda-hmcroot:
+# Provide access to the root on the USB/DVD drive
+# of the SE/HMC of System z hardware.
```
jkonecny12: I think the comment could be just one line.

dracut/anaconda-hmcroot
```diff
@@ -0,0 +1,25 @@
...
+
+command -v getargbool >/dev/null || . /lib/dracut-lib.sh
+command -v config_get >/dev/null || . /lib/anaconda-lib.sh
```
jkonecny12: If I understand this `command -v` correctly it is testing command availability if so why do you testing these commands if they are not used here anywhere?

dracut/anaconda-modprobe.sh
```diff
@@ -30,6 +32,5 @@ MODULE_LIST+=" raid0 raid1 raid5 raid6 raid456 raid10 linear dm-mod dm-zero  \
               sha256 lrw xts "
 
 for m in $MODULE_LIST; do
-    modprobe $m &>/dev/null
+    modprobe $m && debug_msg "$m was loaded" || debug_msg "$m was NOT loaded"
```
jkonecny12: Could this be an `if-else` instead of this please.

dracut/module-setup.sh
```diff
@@ -14,6 +14,12 @@ depends() {
     return 0
 }
 
+installkernel() {
+    case "$(uname -m)" in
+        s390*) instmods hmcdrv ;;
+    esac
```
jkonecny12: Why to use `case` when it is only one value tested here?

dracut/module-setup.sh
```diff
@@ -66,5 +72,14 @@ install() {
...
     done
+
+    # support for specific architectures
+    case "$(uname -m)" in
+        s390*)
+            inst "/usr/sbin/lshmc"
+            inst "/usr/bin/hmcdrvfs"
+            inst "$moddir/anaconda-hmcroot" "/sbin/anaconda-hmcroot"
+        ;;
+    esac
```
jkonecny12: The same as above. Why to use `case`?

pyanaconda/packaging/__init__.py
```diff
@@ -1080,6 +1080,31 @@ def _setupInstallDevice(self, storage, checkmount):
...
+                log.debug("HMC is already mounted at %s.", DRACUT_REPODIR)
+                url = "file://" + DRACUT_REPODIR
+            else:
+                # Test the SE/HMC file access.
+                if iutil.execWithRedirect("/usr/sbin/lshmc", []):
+                    raise PayloadSetupError("The content of HMC media drive couldn't be accessed.")
+
+                # Test if a path is a mount point.
+                if os.path.ismount(INSTALL_TREE):
+                    log.debug("Don't mount the content of HMC media drive yet.")
```
jkonecny12: I think there should be more something like `Test if HMC is a mount point path.`. The actual message is without context what is happening, you just know what is not happening.

pyanaconda/packaging/__init__.py
```diff
@@ -1080,6 +1080,31 @@ def _setupInstallDevice(self, storage, checkmount):
...
+                else:
+                    # Make sure that the directories exists.
+                    iutil.mkdirChain(INSTALL_TREE)
+
+                    # Mount the device.
+                    if iutil.execWithRedirect("/usr/bin/hmcdrvfs", [INSTALL_TREE]):
+                        raise PayloadSetupError("The content of HMC media drive couldn't be mounted.")
+
+                # We are ready to use the HMC.
```
jkonecny12: I think you can remove this comment. The debug message is pretty clear.

dracut/module-setup.sh
poncovka: It is the same as in the function `depends` above. I think that the point is that we can more easily extend it with other cases.

dracut/anaconda-hmcroot
poncovka: Fixed.

dracut/anaconda-hmcroot
poncovka: Fixed, but you test the availability of a library, so it doesn't matter what command you test.

pyanaconda/packaging/__init__.py
poncovka: Fixed.

dracut/anaconda-modprobe.sh
poncovka: Fixed.

pyanaconda/packaging/__init__.py
poncovka: Fixed.

poncovka:
Fixed the suggested changes.

dracut/anaconda-hmcroot
jkonecny12: It depends, you can for example rename the command or deprecate it, so there should be commands which you want to use to have valid feedback.

dracut/module-setup.sh
jkonecny12: Ok.

pyanaconda/packaging/__init__.py
```diff
@@ -1080,6 +1080,32 @@ def _setupInstallDevice(self, storage, checkmount):
             url = method.url
             mirrorlist = method.mirrorlist
             sslverify = not (method.noverifyssl or flags.noverifyssl)
+        elif method.method == "hmc":
+            # Check if /dev/hmcdrv is already mounted.
```
jkonecny12: Could you please move this to its own method? This whole method is big as hell already and I don't think we want to have it even bigger.

poncovka:
Jenkins, test this please

poncovka:
Jenkins, test this please

jkonecny12:
jenkins, test this please

PR #1208, opened 2017-10-05
Support timeout and retries options in %packages section (#1448459)

poncovka:
Enable to set yum's timeout and retries with a kickstart file.

Keep in mind, that anaconda does its own retry with a progressive
delay. It is for case of a really unreliable connection. The messages
about retries in anaconda should be more clear now.

Resolves: rhbz#1448459

Depends on: https://github.com/rhinstaller/pykickstart/pull/180

### DIFF
--- a/pyanaconda/packaging/yumpayload.py
+++ b/pyanaconda/packaging/yumpayload.py
@@ -302,6 +302,12 @@ reposdir=%s
         if self.data.packages.multiLib:
             buf += "multilib_policy=all\n"
 
+        if self.data.packages.timeout is not None:
+            buf += "timeout=%s\n" % self.data.packages.timeout
+
+        if self.data.packages.retries is not None:
+            buf += "retries=%s\n" % self.data.packages.retries
+
         if hasattr(self.data.method, "proxy") and self.data.method.proxy:
             try:
                 proxy = ProxyString(self.data.method.proxy)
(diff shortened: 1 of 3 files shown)


pyanaconda/packaging/yumpayload.py
```diff
@@ -302,6 +302,12 @@ def _writeYumConfig(self, cache_dir=_yum_cache_dir):
         if self.data.packages.multiLib:
             buf += "multilib_policy=all\n"
 
+        if self.data.packages.timeout is not None:
+            buf += "timeout=%s\n" % (self.data.packages.timeout,)
```
jkonecny12: Why are you creating tuple from the `self.data.packages.timeout`? It should work with the plain variable.

pyanaconda/packaging/yumpayload.py
```diff
@@ -302,6 +302,12 @@ def _writeYumConfig(self, cache_dir=_yum_cache_dir):
...
+
+        if self.data.packages.retries is not None:
+            buf += "retries=%s\n" % (self.data.packages.retries,)
```
jkonecny12: The same here.

pyanaconda/packaging/yumpayload.py
poncovka: Fixed.

pyanaconda/packaging/yumpayload.py
poncovka: Fixed.

poncovka:
Fixed the suggested changes.

poncovka:
Jenkins, test this please

PR #1209, opened 2017-10-06
Add checks for group names (#1497676)

poncovka:
The group names should be checked the same way as user names.

Resolves: rhbz#1497676

### DIFF
--- a/tests/pyanaconda_tests/user_test.py
+++ b/tests/pyanaconda_tests/user_test.py
@@ -20,15 +20,21 @@
 #
 
 import unittest
-from pyanaconda.users import check_username
+from pyanaconda.users import check_username, check_groupname, check_grouplist
 
 
 class UserNameTests(unittest.TestCase):
 
-    def _assert_username(self, name, expected_validity):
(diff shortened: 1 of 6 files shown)


tests/pyanaconda_tests/user_test.py
```diff
@@ -90,3 +96,43 @@ def length_test(self):
...
+        self._assert_name(" \tfoo", True)
+        self._assert_name("foo ", True)
+        self._assert_name("foo \t", True)
+        self._assert_name("  foo  ", True)
+        self._assert_name("foo,bar", True)
+        self._assert_name("foo, bar", True)
+        self._assert_name("  foo,    bar", True)
+        self._assert_name("foo, bar, xxx", True)
+
+        self._assert_name(",", False)
```
jkonecny12: I'm missing test for `0` which is mentioned in the bug.

poncovka: Fixed.

poncovka:
Fixed the suggested changes.

PR #1211, opened 2017-10-10
Allow users to just assign mount points to block devices

vpodzime:
This PR adds a new way of storage configuration to the Anaconda installer. It is the most trivial way of storage configuration possible -- just assigning mount points to block devices and optionally reformatting them. Users are allowed and encouraged to do whatever they want with their storage using whichever tools they want eventually just telling the installer which block device to mount where.

There are still certain limitations to keep things consistent, though. In particular, the file system mounted as ``/`` has to be reformatted as part of the installation process and all ``swap`` formats need to be reformatted in order to be used by the newly installed system.

The first three patches are quite well isolated to limit the potential regression surface. The fourth one is invasive, but on the other hand only making things safer and less vulnerable. Without it, the repeated visits of the storage configuration screens are doomed to fail (regardless of the new mount point assignment functionality). It's much safer to just start from scratch if users change their mind about how they want to configure storage.

I plan to implement a few kickstart tests for this new functionality if it gets merged. The goal here is to make this new feature available in Fedora 27 as a new (maybe experimental?) way of storage configuration in text mode.

Note: This requires changes from https://github.com/rhinstaller/pykickstart/pull/178 and a matching bump of the required version of *pykickstart*

### DIFF
--- a/pyanaconda/kickstart.py
+++ b/pyanaconda/kickstart.py
@@ -1152,6 +1152,43 @@ class Logging(commands.logging.FC6_Logging):
                 remote_server = "%s:%s" % (self.host, self.port)
             anaconda_logging.logger.updateRemote(remote_server)
 
+
+class Mount(commands.mount.F27_Mount):
+    def execute(self, storage, ksdata, instClass):
+        storage.do_autopart = False
+
+        for md in self.dataList():
+            md.execute(storage, ksdata, instClass)
+
+class MountData(commands.mount.F27_MountData):
+    def execute(self, storage, ksdata, instClass):
+        dev = storage.devicetree.resolve_device(self.device)
+        if dev is None:
+            raise KickstartParseError(formatErrorMsg(self.lineno,
+                                      msg=_("Unknown or invalid device '%s' specified" % self.device)))
+        if self.reformat:
+            if self.format:
+                fmt = get_format(self.format)
+                if not fmt:
+                    msg = _("Unknown or invalid format '%s' specified for device '%s'" % \
+                            (self.format, self.device))
+                    raise KickstartParseError(formatErrorMsg(self.lineno, msg))
+            else:
+                old_fmt = dev.format
+                if not old_fmt or old_fmt.type is None:
+                    raise KickstartParseError(formatErrorMsg(self.lineno,
+                                              msg=_("No format on device '%s'" % self.device)))
+                fmt = get_format(old_fmt.type)
+            storage.format_device(dev, fmt)
+
+        # only set mount points for mountable formats
+        if dev.format.mountable:
+            dev.format.mountpoint = self.mount_point
+
+        # make sure swaps end up in /etc/fstab
+        if fmt.type == "swap":
+            storage.add_fstab_swap(dev)
+
 class Network(commands.network.F27_Network):
     def __init__(self, *args, **kwargs):
         commands.network.F27_Network.__init__(self, *args, **kwargs)
@@ -2194,6 +2231,7 @@ commandMap = {
     "lang": Lang,
     "logging": Logging,
     "logvol": LogVol,
+    "mount": Mount,
     "network": Network,
     "part": Partition,
     "partition": Partition,
@@ -2217,6 +2255,7 @@ commandMap = {
 dataMap = {
     "BTRFSData": BTRFSData,
     "LogVolData": LogVolData,
+    "MountData": MountData,
     "PartData": PartitionData,
     "RaidData": RaidData,
     "RepoData": RepoData,
@@ -2474,6 +2513,7 @@ def doKickstartStorage(storage, ksdata, instClass):
     ksdata.volgroup.execute(storage, ksdata, instClass)
     ksdata.logvol.execute(storage, ksdata, instClass)
     ksdata.btrfs.execute(storage, ksdata, instClass)
+    ksdata.mount.execute(storage, ksdata, instClass)
     # setup snapshot here, that means add it to model and do the tests
     # snapshot will be created on the end of the installation
     ksdata.snapshot.setup(storage, ksdata, instClass)
--- a/pyanaconda/kickstart.py
+++ b/pyanaconda/kickstart.py
@@ -2505,8 +2505,6 @@ def resetCustomStorageData(ksdata):
     for command in ["partition", "raid", "volgroup", "logvol", "btrfs"]:
         ksdata.resetCommand(command)
 
-    ksdata.clearpart.type = CLEARPART_TYPE_NONE
-
 def doKickstartStorage(storage, ksdata, instClass):
     """ Setup storage state from the kickstart data """
     ksdata.clearpart.execute(storage, ksdata, instClass)
--- a/pyanaconda/kickstart.py
+++ b/pyanaconda/kickstart.py
@@ -1160,6 +1160,15 @@ class Mount(commands.mount.F27_Mount):
         for md in self.dataList():
             md.execute(storage, ksdata, instClass)
 
+    def add_mount_data(self, md):
+        self.mount_points.append(md)
+
+    def remove_mount_data(self, md):
+        self.mount_points.remove(md)
+
+    def clear_mount_data(self):
+        self.mount_points = list()
+
 class MountData(commands.mount.F27_MountData):
     def execute(self, storage, ksdata, instClass):
         dev = storage.devicetree.resolve_device(self.device)
@@ -1180,14 +1189,17 @@ class MountData(commands.mount.F27_MountData):
                                               msg=_("No format on device '%s'" % self.device)))
                 fmt = get_format(old_fmt.type)
             storage.format_device(dev, fmt)
+            # make sure swaps end up in /etc/fstab
+            if fmt.type == "swap":
+                storage.add_fstab_swap(dev)
 
         # only set mount points for mountable formats
-        if dev.format.mountable:
+        if dev.format.mountable and self.mount_point is not None and self.mount_point != "none":
             dev.format.mountpoint = self.mount_point
 
-        # make sure swaps end up in /etc/fstab
-        if fmt.type == "swap":
-            storage.add_fstab_swap(dev)
+        dev.format.create_options = self.mkfs_opts
+        dev.format.options = self.mount_opts
+
 
 class Network(commands.network.F27_Network):
     def __init__(self, *args, **kwargs):
--- a/pyanaconda/ui/tui/spokes/storage.py
+++ b/pyanaconda/ui/tui/spokes/storage.py
@@ -87,7 +87,6 @@ class StorageSpoke(NormalTUISpoke):
         self.select_all = False
 
         self.autopart = None
-        self.clearPartType = None
 
         # This list gets set up once in initialize and should not be modified
         # except perhaps to add advanced devices. It will remain the full list
@@ -205,6 +204,11 @@ class StorageSpoke(NormalTUISpoke):
         # This print is foul.  Need a better message display
         print(_(PAYLOAD_STATUS_PROBING_STORAGE))
         threadMgr.wait(THREAD_STORAGE_WATCHER)
+
+        if not any(d in self.storage.disks for d in self.disks):
+            # something happened to self.storage (probably reset), need to
+            # reinitialize the list of disks
+            self._initialize(reinit=True)
 
         # synchronize our local data store with the global ksdata
         # Commment out because there is no way to select a disk right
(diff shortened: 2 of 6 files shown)


pyanaconda/ui/tui/spokes/storage.py
```diff
@@ -576,3 +596,256 @@ def input(self, args, key):
...
+        self._container = None
+        self._mds = None
+
+        self._gather_mount_data_info()
+
+    def _gather_mount_data_info(self):
+        self._mds = OrderedDict()
+
+        for device in (d for d in self.storage.devicetree.leaves if not d.protected and d.size != Size(0)):
+            if device.format.type != "none":
```
M4rtinK: Shouldn't there be a constant for this ?

pyanaconda/ui/tui/spokes/storage.py
```diff
@@ -576,3 +596,256 @@ def input(self, args, key):
...
+                    devspec += "*"
+                if md.mount_point:
+                    devspec += ", %s" % md.mount_point
+            w = TextWidget(devspec)
+            self._container.add(w, self._configure_device, device)
+
+        self.window.add_with_separator(self._container)
+
+        message = _("Choose device from above to assign mount point and set format.\n" +
+                    "Formats marked with * are new formats meaning all DATA on the original format WILL BE LOST!")
```
M4rtinK: I would suggest making all also capital, eq.: **ALL** DATA

pyanaconda/ui/tui/spokes/storage.py
```diff
@@ -576,3 +596,256 @@ def input(self, args, key):
...
+            elif key.lower() == C_('TUI|Spoke Navigation', 'c'):
+                self.apply()
+
+            return super(MountPointAssignSpoke, self).input(args, key)
+
+        return InputState.PROCESSED
+
+    def apply(self):
+        """ Apply our selections. """
+        for md in self._mds.values():
```
M4rtinK: What's actually `md` in this case ? I would suggest using a longer variable name for clarity.

pyanaconda/ui/tui/spokes/storage.py
```diff
@@ -576,3 +596,256 @@ def input(self, args, key):
...
+        NormalTUISpoke.refresh(self, args)
+
+        self._container = ListColumnContainer(1)
+
+        fmt = get_format(self._md.format)
+        if fmt and fmt.mountable:
+            self._container.add(TextWidget("Mount point: %s" % self._md.mount_point or "none"), self._assign_mount_point, self._md)
+        else:
+            self._container.add(TextWidget("Mount point: [%s]" % fmt.name), lambda x: self.redraw(), None)
+        self._container.add(TextWidget("Format: %s" % self._md.format or "none"), self._set_format, self._md)
```
M4rtinK: Same suggestion - I think `"none"` should be defined by a constant there as well.

pyanaconda/ui/tui/spokes/storage.py
```diff
@@ -576,3 +596,256 @@ def input(self, args, key):
...
+            return Prompt(_("New mount point for the device"))
+        else:
+            self.apply()
+            self.close()
+
+    def input(self, args, key):
+        if not key:
+            self.value = None
+            return InputState.DISCARDED
+        elif key.startswith("/"):
```
M4rtinK: I think a comment there might be helpful.

pyanaconda/ui/tui/spokes/storage.py
```diff
@@ -576,3 +596,256 @@ def input(self, args, key):
     def apply(self):
         """ Apply our selections. """
         self.data.autopart.type = self._selected_scheme_value
+
+class MountDataRec(kickstart.MountData):
+    """ An artificial subclass also recording changes """
```
jkonecny12: Please write period on the end of sentences in docstrings.

pyanaconda/ui/tui/spokes/storage.py
```diff
@@ -576,3 +596,256 @@ def input(self, args, key):
...
+                self.data.mount.add_mount_data(md)
+
+class ConfigureDeviceSpoke(NormalTUISpoke):
+    """ Assign mount point to a block device and (optionally) reformat it """
+    category = SystemCategory
+
+    def __init__(self, data, storage, payload, instclass, mount_data):
+        NormalTUISpoke.__init__(self, data, storage, payload, instclass)
+        self._container = None
+        self._md = mount_data
```
jkonecny12: Can you please change the name from `_md` to `_mount_data`.

I don't like two latter variable for something used in the whole class. It could be anything like for example `MetaData`...

pyanaconda/ui/tui/spokes/storage.py
```diff
@@ -576,3 +596,256 @@ def input(self, args, key):
...
+        if ((self._md.orig_format and self._md.orig_format != self._md.format) or self._md.mount_point == "/"):
+            # changing format implies reformat and so does "/" mount point
+            self._container.add(CheckboxWidget(title="Reformat", completed=self._md.reformat), lambda x: self.redraw(), None)
+        else:
+            self._container.add(CheckboxWidget(title="Reformat", completed=self._md.reformat), self._switch_reformat, self._md)
+
+        self.window.add_with_separator(self._container)
+
+        message = _("Choose from above to assign mount point and/or set format.")
+        self.window.add_with_separator(TextWidget(message))
```
jkonecny12: I guess you can inline the `message` here. You have bigger lines above which are ok.

pyanaconda/ui/tui/spokes/storage.py
```diff
@@ -576,3 +596,256 @@ def input(self, args, key):
...
+        ScreenHandler.push_screen(dialog)
+
+class SetFormatDialog(EditTUIDialog):
+    """ Set format for a device """
+    category = SystemCategory
+
+    def __init__(self, data, storage, payload, instclass, mount_data):
+        EditTUIDialog.__init__(self, data, storage, payload, instclass)
+        self.title = N_("Configure device format")
+        self._md = mount_data
```
jkonecny12: The same here.

pyanaconda/ui/tui/spokes/storage.py
```diff
@@ -576,3 +596,256 @@ def input(self, args, key):
...
+
+    @property
+    def indirect(self):
+        return True
+
+    def prompt(self, args=None):
+        if self.value is None:  # first run or nothing entered
+            return Prompt(_("New format for the device%s") % (" [%s]" % self._md.format if self._md.format else ""))
```
jkonecny12: First of all it is not clear what you are doing when you have `%` within `%`.
Second I would move the space to the original string so that `...device%s` will be `...device %s`.

pyanaconda/ui/tui/spokes/storage.py
```diff
@@ -201,6 +201,13 @@ def _update_summary(self):
     def refresh(self, args=None):
         NormalTUISpoke.refresh(self, args)
 
+        if not any(d in self.storage.disks for d in self.disks):
+            # something happened to self.storage (probably reset), need to
+            # reinitialize the list of disks
+            threadMgr.add(AnacondaThread(name=THREAD_STORAGE_WATCHER,
+                                         target=self._initialize))
+
+
```
jkonecny12: Remove one empty line here.

pyanaconda/ui/tui/spokes/storage.py
```diff
@@ -367,7 +374,7 @@ def apply(self):
         if self.autopart:
             self.clearPartType = CLEARPART_TYPE_ALL
         else:
-            self.clearPartType = CLEARPART_TYPE_NONE
+            self.clearPartType = None
```
jkonecny12: Why this change? What the `None` here means?

pyanaconda/ui/tui/spokes/storage.py
```diff
@@ -513,27 +526,51 @@ def apply(self):
...
+            self.data.clearpart.type = CLEARPART_TYPE_NONE
+            self.data.clearpart.initAll = False
+
+    def _ensure_init_storage(self):
+        """
+        If a different clearpart type was chosen or mount point assignment was
+        chosen instead, we need to reset/rescan storage to revert all changes
+        done by the previous run of doKickstartStorage() and get everything into
+        the initial state.
+
```
jkonecny12: Remove this empty line please.

pyanaconda/ui/tui/spokes/storage.py
```diff
@@ -616,10 +653,18 @@ def __init__(self, data, storage, payload, instclass):
 
         self._gather_mount_data_info()
 
+    def _is_dev_usable(self, dev):
+        maybe = not dev.protected and dev.size != Size(0)
+        if maybe and self.data.ignoredisk.onlyuse:
+            # all device's disks have to be in ignoredisk.onlyuse
+            maybe = set(self.data.ignoredisk.onlyuse).issuperset({d.name for d in dev.disks})
+
+        return maybe
```
jkonecny12: Isn't this code which should be in Blivet?

pyanaconda/ui/tui/spokes/storage.py
```diff
@@ -513,27 +526,51 @@ def apply(self):
...
             self.data.autopart.autopart = True
             self.data.clearpart.type = self.clearPartType
             self.data.clearpart.initAll = True
             self.data.mount.clear_mount_data()
         else:
             self.data.autopart.autopart = False
+            self.data.clearpart.type = CLEARPART_TYPE_NONE
+            self.data.clearpart.initAll = False
+
+    def _ensure_init_storage(self):
```
jkonecny12: Could this be done as common function somewhere else, maybe in Blivet directly.
This sounds to me like reset storage from any state to the original state, isn't this basically `Blivet.reset()`?

pyanaconda/kickstart.py
```diff
@@ -1180,14 +1189,17 @@ def execute(self, storage, ksdata, instClass):
                                               msg=_("No format on device '%s'" % self.device)))
                 fmt = get_format(old_fmt.type)
             storage.format_device(dev, fmt)
+            # make sure swaps end up in /etc/fstab
+            if fmt.type == "swap":
+                storage.add_fstab_swap(dev)
 
         # only set mount points for mountable formats
-        if dev.format.mountable:
+        if dev.format.mountable and self.mount_point is not None and self.mount_point != "none":
```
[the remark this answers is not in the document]
vpodzime: ``"none"``  can end up here from kickstart. See https://github.com/rhinstaller/pykickstart/pull/178/files#diff-626dd4fad4450cf40de2c467d3df3262R1163 for details.

pyanaconda/ui/tui/spokes/storage.py
vpodzime: Yes, there should be a constant for this, but it should be defined in *Blivet*, not here. And I don't want to introduce a dependency on a new release of blivet just because of this.

pyanaconda/ui/tui/spokes/storage.py
vpodzime: Why? It just says there is no format. Maybe it should be translated, though? Fixing.

vpodzime:
@M4rtinK's comments addressed and the new feature is marked as EXPERIMENTAL. I'll take a look at the other concerns tomorrow.

pyanaconda/ui/tui/spokes/storage.py
vpodzime: This change is a leftover from an older version. But you know what, the ``self.clearPartType`` attribute is only set in this class and never used. So I'm just removing it instead.

pyanaconda/ui/tui/spokes/storage.py
```diff
@@ -450,48 +455,55 @@ def _initialize(self):
...
-       .. inheritance-diagram:: AutoPartSpoke
+       .. inheritance-diagram:: PartTypeSpoke
           :parts: 3
     """
     category = SystemCategory
 
     def __init__(self, data, storage, payload, instclass):
         NormalTUISpoke.__init__(self, data, storage, payload, instclass)
-        self.title = N_("Autopartitioning Options")
+        self.title = N_("Partitioning Options")
```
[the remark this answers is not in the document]
vpodzime: Yeah, i was thinking the same. Better than having a misleading title and it's text mode after all.

pyanaconda/ui/tui/spokes/storage.py
vpodzime: ``storage_initialize()`` is a bit special beast which tries to make sure it behaves more friendly than ``reset()`` in the installation environment. And this method itself is very specific to text mode and users changing their mind. GUI uses a totally different approach.

pyanaconda/ui/tui/spokes/storage.py
vpodzime: Not really, it says which devices to show on this screen.

pyanaconda/ui/tui/spokes/storage.py
vpodzime: But that would mean there would always be a space after ``device`` which is exactly the result I want to prevent.

vpodzime:
I believe all the comments should be addressed now.

M4rtinK:
> I believe all the comments should be addressed now.
Yep, looks good to me! :) Once the Pykickstart PR is merged and a new build done I'll merge this and do a new Anaconda build as well + Bodhi.

PR #1214, opened 2017-10-11
Fix dnf exception repository not set (#1495211)

jkonecny12:
The `set_source_url()` is already called in `SpecifyRepoSpoke`. It shouldn't be called before because user may decide to not set anything as a repo.


jkonecny12:
**UPDATED:**
* Fix change of the installation method

PR #1216, opened 2017-10-12
Fix a translation check error

M4rtinK:
Apparently you need to use named substitution in this
case or msgid will complain and ruin the day.


M4rtinK:
Nominating as a hotfix as this is blocking a F27 Anaconda build.

PR #148, opened 2015-06-18
Do something about the new flood of pylint errors I caused

dashea:

PR #1218, opened 2017-10-16
packaging: clear downloaded packages repo cache before using it (#148…

rvykydal:
…0790)

So installer does not crash on reused home having invalid data in the case from
previous interrupted install attemps.

### DIFF
--- a/pyanaconda/payload/dnfpayload.py
+++ b/pyanaconda/payload/dnfpayload.py
@@ -893,6 +893,9 @@ class DNFPayload(payload.PackagePayload):
                 log.error("Installation failed: %r", e)
                 _failure_limbo()
 
+        if os.path.exists(self._download_location):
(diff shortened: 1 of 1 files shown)


pyanaconda/payload/dnfpayload.py
```diff
@@ -893,6 +893,9 @@ def install(self):
                 log.error("Installation failed: %r", e)
                 _failure_limbo()
 
+        if os.path.exists(self._download_location):
+            log.info("Removing existing download location: %s", self._download_location)
```
M4rtinK: I would suggest "Removing existing package download location:" instead. :)
IIRC there could be more stuff we might download, such as metadata, livepayload, tarpayload, etc.

rvykydal: Absolutely. Updated.

PR #1220, opened 2017-10-18
network: create default ifcfg also for missing default NM connection …

rvykydal:
…(#1478141)

In Fedora we create default ifcfg files by dumping default wired connections
created by NM.  NM does not create such connections for InfiniBand devices and
consequently installer crashes in text mode because of missing ifcfg file for
the IB device.

Create default ifcfg also when there is no NM default connection found for a
(in this BZ case InfiniBand) device.

This would be needed also when RHEL (server) policy is applied to the installer
turning creating of default connections by NM off in general. With the patch,
if default NM connection is not found, default ifcfg file is created in the
same way as in rhel7-branch.


rvykydal:
Duplicate PR, closing

PR #1222, opened 2017-10-19
Mac EFI installs need grub2-tools (#1503496)

AdamWill:
For the UEFI 64-on-32 stuff, pjones changed the MacEFIGRUB class
to just define the `packages` attribute directly as a list of
two packages; previously, MacEFIGRUB inherited the list from a
parent class. As reported in #1503496, the two packages listed
(grub2-tools-efi and mactel-boot) aren't enough for bootloader
install to actually work.

This changes things so MacEFIGRUB defines `_packages64`, like
Aarch64EFIGRUB, which should effectively cause it to include
those packages, plus the packages listed in the `GRUB2` class
and the `_packages_common` from the `EFIGRUB` class. I'm not
sure if that might be too many packages, but it's at least in
line with how another class does things and simple. The other
option would be just to extend the class's direct `packages`
list.

### DIFF
--- a/pyanaconda/bootloader.py
+++ b/pyanaconda/bootloader.py
@@ -1862,7 +1862,7 @@ class Aarch64EFIGRUB(EFIGRUB):
     _efi_binary = "\\shimaa64.efi"
 
 class MacEFIGRUB(EFIGRUB):
-    packages = [ "grub2-tools-efi", "mactel-boot" ]
+    _packages64 = [ "grub2-tools-efi", "mactel-boot" ]
     def mactel_config(self):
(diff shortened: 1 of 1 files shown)


AdamWill:
@vathpela

AdamWill:
For reference, this is the commit where @vathpela changed this:

https://github.com/rhinstaller/anaconda/commit/21f1a546c3d2d203f2718f02e7a4df4879dbe4d6

jkonecny12:
Jenkins, it's ok to test.

jkonecny12:
jenkins, test this please

PR #1224, opened 2017-10-25
Add support for the new 'mount' kickstart command

vpodzime:
It is supposed to set up mount points and optionally new formats
on devices.

(ported commit d3082ac8f8862224cb1e5e8b0dc5c2cb73d267e5 from f27-devel)

Resolves: rhbz#1450922

### DIFF
--- a/pyanaconda/kickstart.py
+++ b/pyanaconda/kickstart.py
@@ -1147,6 +1147,45 @@ class Logging(commands.logging.FC6_Logging):
                 remote_server = "%s:%s" %(self.host, self.port)
             logger.updateRemote(remote_server)
 
+class Mount(commands.mount.RHEL7_Mount):
+    def execute(self, storage, ksdata, instClass):
+        storage.do_autopart = False
+
+        for md in self.dataList():
+            md.execute(storage, ksdata, instClass)
+
+class MountData(commands.mount.RHEL7_MountData):
+    def execute(self, storage, ksdata, instClass):
+        dev = storage.devicetree.resolveDevice(self.device)
+        if dev is None:
+            raise KickstartValueError(formatErrorMsg(self.lineno,
+                                      msg=_("Unknown or invalid device '%s' specified") % self.device))
+        if self.reformat:
+            if self.format:
+                fmt = getFormat(self.format)
+                if not fmt:
+                    msg = _("Unknown or invalid format '%(format)s' specified for device '%(device)s'") % \
+                            {"format" : self.format, "device" : self.device}
+                    raise KickstartValueError(formatErrorMsg(self.lineno, msg))
+            else:
+                old_fmt = dev.format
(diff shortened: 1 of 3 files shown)


pyanaconda/kickstart.py
```diff
@@ -80,7 +80,7 @@
                                   SELINUX_DISABLED, SELINUX_ENFORCING, SELINUX_PERMISSIVE, \
                                   SNAPSHOT_WHEN_POST_INSTALL, SNAPSHOT_WHEN_PRE_INSTALL
 from pykickstart.base import BaseHandler, KickstartCommand
-from pykickstart.errors import formatErrorMsg, KickstartError, KickstartValueError
+from pykickstart.errors import formatErrorMsg, KickstartError, KickstartValueError, KickstartParseError
```
jkonecny12: You should use `KickstartValueError` for rhel7-branch as all the other commands do.

vpodzime: Any real reason for this? I mean, it makes no difference and only makes porting/cherry-picking things harder.

jkonecny12: The main reason here is the consistency. Yeah I know the difference on rhel and master is a pain but I don't want to introduce inconsistency because of harder cherry-picking.  

On the `master` it is because deprecation of `KickstartValueError` in Pykickstart and RHEL used this exception before the deprecation.

But I'm open for discussion about changing all of those exception to `KickstartParseError` to make it consistent between rhel7 and master branches. What do you think @poncovka , @rvykydal , @M4rtinK ?

poncovka: I agree with @jkonecny12 . `KickstartParseError` doesn't make sense here in a context of RHEL 7.
```

class KickstartParseError(KickstartError):
    """An exception class for errors when processing the input file, such as
       unknown options, commands, or sections.
    """

class KickstartValueError(KickstartError):
    """An exception class for errors when processing arguments to commands,
       such as too many arguments, too few arguments, or missing required
       arguments.
    """
```

rvykydal: For this PR I agree with Jirka and I'd just use KickstartValueError in rhel7-branch.
I wouldn't do the master/rhel7-branch unification in scope of this patch, but if Vratislav prefers to do that, I won't NACK it.

jkonecny12: So please @vpodzime change this to `KickstartValueError` and you have my ACK.

vpodzime:
> So please @vpodzime change this to KickstartValueError and you have my ACK.

Done!

vpodzime:
We still need the new release of pykickstart though.

poncovka:
Jenkins, test this please

jkonecny12:
jenkins, test this please

poncovka:
Jenkins, test this please

PR #1225, opened 2017-10-26
Split addon and environment refresh in software TUI

jkonecny12:
This is only the cosmetic change.

### DIFF
--- a/pyanaconda/ui/tui/spokes/software_selection.py
+++ b/pyanaconda/ui/tui/spokes/software_selection.py
@@ -196,34 +196,36 @@ class SoftwareSpoke(NormalTUISpoke):
         threadMgr.wait(THREAD_CHECK_SOFTWARE)
         self._container = ListColumnContainer(2, columns_width=38, spacing=2)
 
-        # Display the environments
         if args is None:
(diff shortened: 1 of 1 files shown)


pyanaconda/ui/tui/spokes/software_selection.py
```diff
@@ -196,34 +196,38 @@ def refresh(self, args=None):
...
+    def _refresh_addons(self, available_addons):
+        length = len(available_addons)
+
+        for addon_id in available_addons:
+            name = self.payload.groupDescription(addon_id)[0]
+            selected = addon_id in self._addons_selection
+            widget = CheckboxWidget(title="%s" % name, completed=selected)
+            self._container.add(widget, callback=self._set_addons_callback, data=addon_id)
+
+        if length > 0:
```
poncovka: This can be changed to `if available_addons` and `length` can be removed.

jkonecny12: And this is why it is important to polish code from time to time 👍 . Thanks!

jkonecny12:
**UPDATED:**
* Tweaked code based on @poncovka suggestions

PR #1227, opened 2017-10-27
Add modular server repo to the base repositories (#1506894)

jkonecny12:
The base repo is tested at the end of the installation and if the base repo is not enabled then the payload reset fails. This breaks closest source installation where this check is applied.

The base repo is recognized by name and this name could have the following values:

anaconda
rawhide
first part before "-" sign in the product name

This doesn't work for Fedora-Modular because it is fedora but that can't be used for repo id or there will be repo name collision.

Solved by adding fedora-modular-server to this list too.

Resolves: rhbz#1506894

### DIFF
--- a/pyanaconda/payload/__init__.py
+++ b/pyanaconda/payload/__init__.py
@@ -980,7 +980,7 @@ class ArchivePayload(ImagePayload):
 class PackagePayload(Payload):
     """A PackagePayload installs a set of packages onto the target system."""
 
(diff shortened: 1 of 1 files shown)


jkonecny12:
This was solved by applying this PR in side-tag build because that was blocking the fedora modular server. So or so if there will be time it would be nice to do a build on Monday.

You can contact **sgallagh** on freenode IRC :)

jkonecny12:
It could and it should this is only a hotfix. Proper solution should be implemented.

PR #1229, opened 2017-11-02
Really install all the right packages on Mac UEFI installs

AdamWill:
My previous fix still wasn't quite right - it overrode the
EFIGRUB `_packages64` list so those packages, which are needed,
weren't included. So let's *add* to that list, not replace it.
Note we have to put this in `__init__` because we can't use
`super()` from the class block, that just flat doesn't work.

### DIFF
--- a/pyanaconda/bootloader.py
+++ b/pyanaconda/bootloader.py
@@ -1821,7 +1821,6 @@ class EFIGRUB1(EFIBase, GRUB):
 
 class EFIGRUB(EFIBase, GRUB2):
     _packages32 = [ "grub2-efi-ia32", "shim-ia32" ]
-    _packages64 = [ "grub2-efi-x64", "shim-x64" ]
     _packages_common = [ "efibootmgr" ]
     can_dual_boot = False
     stage2_is_valid_stage1 = False
@@ -1832,6 +1831,7 @@ class EFIGRUB(EFIBase, GRUB2):
     def __init__(self):
         super(EFIGRUB, self).__init__()
         self.efi_dir = 'BOOT'
+        self._packages64 = [ "grub2-efi-x64", "shim-x64" ]
 
         try:
             f = open("/sys/firmware/efi/fw_platform_size", "r")
@@ -1857,12 +1857,18 @@ class EFIGRUB(EFIBase, GRUB2):
             super(EFIGRUB, self).packages
 
 class Aarch64EFIGRUB(EFIGRUB):
-    _packages64 = ["grub2-efi-aa64", "shim-aa64"]
     _serial_consoles = ["ttyAMA", "ttyS"]
     _efi_binary = "\\shimaa64.efi"
 
+    def __init__(self):
+        super(Aarch64EFIGRUB, self).__init__()
+        self._packages64 = ["grub2-efi-aa64", "shim-aa64"]
+
 class MacEFIGRUB(EFIGRUB):
-    _packages64 = [ "grub2-tools-efi", "mactel-boot" ]
+    def __init__(self):
+        super(MacEFIGRUB, self).__init__()
+        self._packages64.extend(["grub2-tools-efi", "mactel-boot"])
+
     def mactel_config(self):
         if os.path.exists(iutil.getSysroot() + "/usr/libexec/mactel-boot-setup"):
             rc = iutil.execInSysroot("/usr/libexec/mactel-boot-setup", [])


pyanaconda/bootloader.py
```diff
@@ -1862,7 +1862,10 @@ class Aarch64EFIGRUB(EFIGRUB):
     _efi_binary = "\\shimaa64.efi"
 
 class MacEFIGRUB(EFIGRUB):
-    _packages64 = [ "grub2-tools-efi", "mactel-boot" ]
+    def __init__(self):
+        super(MacEFIGRUB, self).__init__()
+        self._packages64 = super(MacEFIGRUB, self)._packages64 + ["grub2-tools-efi", "mactel-boot"]
```
jkonecny12: By doing this you are replacing that list not adding to it.
By assigning something to `self._packages64` variable you are creating `self._packages64` instance (not class) variable so the original array is not extended at all. You really don't want to do this.

Please use the same solution as it was before. Just manually copy & paste old values from the array of parent and extend it by typing new values or you can do this:

``self._packages64 = EFIGRUB._packages64 + ["grub2-tools-efi", "mactel-boot"]``

Sorry it is that way, I think we should change this whole class and replace these class variables by instance variables. I don't think there is a reason for having them as class variables.

AdamWill:
"By doing this you are replacing that list not adding to it.
By assigning something to self._packages64 variable you are creating self._packages64 instance (not class) variable so the original array is not extended at all."

Sure, I'm perfectly aware of that. But...

"You really don't want to do this."

Why not? What problem does it cause? AFAICS you didn't explain this, and it is not obvious to me.

"Please use the same solution as it was before. Just manually copy & paste old values from the array of parent and extend it by typing new values or you can do this:"

I don't like either of those. copy & paste approach means that *inevitably* someone will change one list and forget to change the other. Specifying the parent class name requires us to remember to update that line any time the class inheritance changes.

"Sorry it is that way, I think we should change this whole class and replace these class variables by instance variables. I don't think there is a reason for having them as class variables."

Agreed, that would make things simpler. I suspect it was just a case of trying to do things as short as possible, or something.

jkonecny12:
> "You really don't want to do this."
> 
> Why not? What problem does it cause? AFAICS you didn't explain this, and it is not obvious to me.

Because it behaves different from what it looks like. It is nice starting point for another bug to happen.
 
> "Please use the same solution as it was before. Just manually copy & paste old values from the array of parent and extend it by typing new values or you can do this:"
> 
> I don't like either of those. copy & paste approach means that inevitably someone will change one list and forget to change the other. Specifying the parent class name requires us to remember to update that line any time the class inheritance changes.

Yeah I know, the right solution would be to replace all these class variables by instance variables. But that is a big change and it should go only to `master` branch.

> "Sorry it is that way, I think we should change this whole class and replace these class variables by instance variables. I don't think there is a reason for having them as class variables."
> 
> Agreed, that would make things simpler. I suspect it was just a case of trying to do things as short as possible, or something.

Yeah it is possible. Honestly I don't know why this is in Anaconda on so many places. I guess it is simpler to just avoid having `__init__` method created.

AdamWill:
@jkonecny12 OK, sure. I just wanted to make sure there wasn't a more serious problem with this initial approach, as it's what we included in the F27 release candidates. If it's just that it creates a possibility of future problems due to non-obvious behaviour, then I think we're OK with shipping it in F27 but not merging it, agreed?

When I get a minute I'll update this PR to work by making the relevant variables instance variables for all classes, so things are clearer and consistent. Sounds like we all agree that's the way to go for the future.

jkonecny12:
> @jkonecny12 OK, sure. I just wanted to make sure there wasn't a more serious problem with this initial approach, as it's what we included in the F27 release candidates. If it's just that it creates a possibility of future problems due to non-obvious behaviour, then I think we're OK with shipping it in F27 but not merging it, agreed?

Yep, I agree with that :+1:

AdamWill:
OK, I've adjusted the PR to be against master (not f27-devel) and to make `_packages64` an instance variable for all classes. Look better?

jkonecny12:
jenkins, test this please

jkonecny12:
jenkins, test this please

jkonecny12:
jenkins, test this please

jkonecny12:
Could you please fix this error:
```
E1101(no-member):1870,27: MacEFIGRUB.__init__: Super of 'MacEFIGRUB' has no '_packages64' member
```

before we will merge your PR?

Thanks

AdamWill:
D'oh. Sigh. The perils of incremental revision. There.

jkonecny12:
jenkins, test this please

jkonecny12:
Thank you much for your help :+1:

PR #1231, opened 2017-11-06
F27 devel make vlan gui more robust

rvykydal:


rvykydal:
jenkins, test this please

pyanaconda/ui/gui/spokes/network.py
```diff
@@ -429,6 +436,25 @@ def add_connection_to_list(self, con):
...
+        the value as <PARENT_IFACE>.<VLAN_ID> - same as NetworkManager.
+        """
+        iface = connection.get_setting_connection().get_interface_name()
+        if not iface:
+            setting_vlan = connection.get_setting_vlan()
+            if setting_vlan:
+                vlanid = setting_vlan.get_id()
+                parent = setting_vlan.get_parent()
+                # if parent is specified by UUID
+                if len(parent) == 36:
```
jkonecny12: What if the parent will have name with 36 characters but it won't be UUID ? Could that happen?

pyanaconda/ui/gui/spokes/network.py
```diff
@@ -1020,12 +1020,16 @@ def _refresh_slaves(self, dev_cfg):
 
     def _refresh_parent_vlanid(self, dev_cfg):
         if dev_cfg.get_device_type() == NM.DeviceType.VLAN:
+            vlanidstr = parent = ""
```
jkonecny12: Please rename this variable. Sounds to me like someones name :)

rvykydal: hehe, I'll try

pyanaconda/ui/gui/spokes/network.py
rvykydal: No, it can be either UUID or network iface name and the name size is limited to 15 or 16 chars in kernel

jkonecny12: ok, then it should be fine :)

rvykydal:
I got rid of the vlanidstr variable completely.

PR #1232, opened 2017-11-07
Remove storage check for too small swap (#1466964)

poncovka:
The size of swap for autopartitioning is computed from the available
disk space. We should be able to repeat this computation to get the
same value for storage checking, so we can compare the actual size
of swap with the recommended one. However, the available disk space
is used for partitioning at that point, so we cannot use it to compute
the recommended size of swap. We could use the total space of
mountable partitions instead of it, but it doesn't work for btrfs and
lvm thinp.

We could also move the check to the custom partitioning spoke. Since
we cannot check swap that was created automatically, we can run the
check only when a new swap is created with Add new mountpoint dialog.
However, that means that the check will not be run when user changes
the size the swap. That can be confusing and users might report this
behavior as a bug.

In conclusion, we are not able to reliably do this check, so it will
be removed.

(reverted commit 00aab8d)

Resolves: rhbz#1466964


poncovka:
Jenkins, test this please.

poncovka:
Jenkins, test this please.

PR #1236, opened 2017-11-20
Add kickstart filter parser

rvykydal:
Filters kickstart for specific commands and section.
Does not parse commands or section headers during the filtering.


rvykydal:
Closing for another approach in
https://github.com/rhinstaller/anaconda/pull/1245

PR #1238, opened 2017-11-22
 Remove useless code

poncovka:
Removal of the dead code that was suggested by vulture. In some cases, I recommend a function for a removal with a comment, but if you think that the function can be useful in the future, we will keep it. The commit with comments will be removed before merging.

* Attributes that are set in anaconda, but never read in anaconda, pykickstart, blivet or any other library.
* Functions we never call and it doesn't make sense to keep them around.
* Classes for exceptions we never raise.
* Constants we never use.
* Fixed misspellings of some attributes and a method.
* Removed `pyanaconda.constants_text`.


pyanaconda/errors.py
```diff
@@ -20,8 +20,8 @@
 from pyanaconda.flags import flags
 
 __all__ = ["ERROR_RAISE", "ERROR_CONTINUE", "ERROR_RETRY",
-           "InvalidImageSizeError", "MissingImageError", "MediaUnmountError",
-           "MediaMountError", "ScriptError", "CmdlineError", "NonInteractiveError",
+           "InvalidImageSizeError", "MissingImageError",
+           "ScriptError", "CmdlineError", "NonInteractiveError",
```
jkonecny12: Modify this to use less lines.

pyanaconda/anaconda_loggers.py
```diff
@@ -50,6 +50,7 @@ def get_main_logger():
 def get_stdout_logger():
     return logging.getLogger(constants.LOGGER_STDOUT)
 
+# TODO: this function is never used
```
jkonecny12: This should stay. It will be harder to implement it again in future.

pyanaconda/iutil.py
```diff
@@ -618,7 +618,7 @@ def watchProcess(proc, name):
             del _forever_pids[proc.pid]
             _raise_exit_error([(name, proc.returncode)])
 
-
+# TODO: this function is never used
```
jkonecny12: Remove.

pyanaconda/network.py
```diff
@@ -203,6 +203,7 @@ def getHostname():
 
     return hn
 
+# TODO: this function is never used
```
jkonecny12: @rvykydal aren't we doing the same somewhere else? It think I have seen output similar to this.

pyanaconda/payload/__init__.py
```diff
@@ -472,6 +472,7 @@ def addRepo(self, ksrepo):
         # Add the repo to the ksdata so it'll appear in the output ks file.
         self.data.repo.dataList().append(ksrepo)
 
+    # TODO: this function is never used
     def removeRepo(self, repo_id):
```
jkonecny12: I think this functionality may stay here.

pyanaconda/payload/__init__.py
```diff
@@ -532,6 +533,7 @@ def selectedGroupsIDs(self):
         """
         return self.selectedGroups()
 
+    # TODO: this function is never used
```
jkonecny12: IMHO leave it here.

pyanaconda/payload/__init__.py
```diff
@@ -558,6 +560,7 @@ def selectGroup(self, groupid, default=True, optional=False):
 
         self.data.packages.groupList.append(grp)
 
+    # TODO: this function is never used
```
jkonecny12: IMHO leave it here.

pyanaconda/progress.py
```diff
@@ -40,6 +40,7 @@
 progressQ.addMessage("complete", 0)
 progressQ.addMessage("quit", 1)             # exit_code
 
+# TODO: this function is never used
```
jkonecny12: I think we won't use this code much in a future. Remove

pyanaconda/simpleconfig.py
```diff
@@ -202,7 +202,7 @@ def __str__(self):
 
         return s
 
-
+# TODO: this function is never used
```
jkonecny12: Remove

pyanaconda/ui/gui/spokes/blivet_gui.py
```diff
@@ -205,6 +205,7 @@ def _do_check(self):
 
         return self._error == ""
 
+    # TODO: this function is never used
```
jkonecny12: @vojtechtrefny what do you think?

pyanaconda/ui/gui/spokes/datetime_spoke.py
```diff
@@ -666,6 +666,7 @@ def _set_timezone(self, timezone):
     def add_to_store_xlated(self, store, item, xlated):
         store.append([item, xlated])
 
+    # TODO: this function is never used
```
jkonecny12: IMHO remove.

pyanaconda/ui/gui/spokes/blivet_gui.py
vojtechtrefny: This function is used from blivet-gui -- https://github.com/storaged-project/blivet-gui/blob/master/blivetgui/osinstall.py#L182

It's a not-so-nice hack -- these buttons are part of the spoke but are actually controlled by blivet-gui.

pyanaconda/network.py
rvykydal: I think we can remove it. We are using logIfcfgFiles everywhere now.

pyanaconda/ui/gui/spokes/welcome.py
```diff
@@ -62,7 +62,6 @@ class WelcomeLanguageSpoke(LangLocaleHandler, StandaloneSpoke):
     def __init__(self, *args, **kwargs):
         StandaloneSpoke.__init__(self, *args, **kwargs)
         LangLocaleHandler.__init__(self)
-        self._xklwrapper = XklWrapper.get_instance()
```
rvykydal: Given this is a singleton I wonder if we don't instantiate it here to do it early enough (even though it will be used only from keyboard spoke). The XklWrapper class doc says that initialization takes quite a lot of time.
But even if it is the case we could probably initialize it more explicitily, not just as a side effect of initializinng welcome spoke.

pyanaconda/exception.py
```diff
@@ -277,7 +277,6 @@ def initExceptionHandling(anaconda):
                   programVersion=startup_utils.get_anaconda_version_string(),
                   programArch=os.uname()[4],
                   attrSkipList=["_intf._actions",
-                                "_intf._currentAction._xklwrapper",
```
M4rtinK: Ar we sure this is not needed ? This seems to be the list of things to exclude from the python-meh dump, so it might be rather hard to track down if stuff is needed or not, especially as it's all just strings.

pyanaconda/payload/__init__.py
M4rtinK: Yeah, we definitely should keep complementary functions (eq. add/remove/modify) even if they are not used at the moment. Even though it would be good to have test coverage for them in such case, so that we are reasonably sure the code still works when real code starts using it.

pyanaconda/simpleconfig.py
M4rtinK: Do we really want to do that ? If we remove the __str__() method, the simpleconfig class instances will not have a string representation. That could be inconvenient during debugging & for python-meh dumps.

pyanaconda/ui/gui/__init__.py
```diff
@@ -731,38 +730,6 @@ def _widgetScale(self):
             # Export the scale so that Gtk programs launched by anaconda are also scaled
             iutil.setenv("GDK_SCALE", "2")
 
-    def _convertSignals(self):
```
M4rtinK: Are we sure we want to remove this ? It looks like something that should be in use, unless it was superseded by something else.

pyanaconda/simpleconfig.py
poncovka: It is not the `__str__` method, but `simple_replace`.

pyanaconda/ui/gui/spokes/welcome.py
poncovka: Good point. However, the instance was used before this commit https://github.com/rhinstaller/anaconda/commit/4e73305378a39e6b0db677ed48cb35ed8709b2e4. So it really looks like a leftover.

anaconda.py
```diff
@@ -284,7 +284,7 @@ def start_debugger(signum, frame):
 
     # do this early so we can set flags before initializing logging
     from pyanaconda.flags import flags, can_touch_runtime_system
-    (opts, depr) = parse_arguments(boot_cmdline=flags.cmdline)
+    (opts, _) = parse_arguments(boot_cmdline=flags.cmdline)
```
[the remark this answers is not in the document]

pyanaconda/anaconda_loggers.py
poncovka: Kept.

pyanaconda/constants.py
```diff
@@ -205,7 +204,6 @@ class PasswordStatus(Enum):
 
 # Payload status messages
 PAYLOAD_STATUS_PROBING_STORAGE = N_("Probing storage...")
-PAYLOAD_STATUS_TESTING_AVAILABILITY = N_("Testing availability...")
```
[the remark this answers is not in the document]
poncovka: Kept.

pyanaconda/errors.py
poncovka: Fixed.

pyanaconda/exception.py
poncovka: Kept.

pyanaconda/iutil.py
poncovka: Removed.

pyanaconda/kickstart.py
```diff
@@ -83,10 +83,9 @@
 from pykickstart.version import returnClassForVersion
 
 from pyanaconda import anaconda_logging
-from pyanaconda.anaconda_loggers import get_module_logger, get_stdout_logger, get_stderr_logger, get_blivet_logger, get_anaconda_root_logger
+from pyanaconda.anaconda_loggers import get_module_logger, get_stdout_logger, get_blivet_logger, get_anaconda_root_logger
```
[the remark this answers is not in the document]

pyanaconda/network.py
poncovka: Removed.

pyanaconda/payload/__init__.py
poncovka: Kept.

pyanaconda/payload/__init__.py
poncovka: Kept.

pyanaconda/payload/__init__.py
poncovka: Kept.

pyanaconda/progress.py
poncovka: Removed.

pyanaconda/simpleconfig.py
poncovka: Removed `simple_replace`.

pyanaconda/ui/gui/spokes/blivet_gui.py
poncovka: Kept.

pyanaconda/ui/gui/spokes/datetime_spoke.py
poncovka: Removed.

poncovka:
Jenkins, test this please

PR #1239, opened 2017-11-22
Add dbus tasks

jkonecny12:
*PoC PR*

This PR contains two possible ways how to use `TaskInterface`. It is more to show possible ways how to implement DBus interface and to decide which is the solution we want to use for future. Please do not review this PR from the point of bugs this PR will change after the decision will be made. 

The module `foo` has implementation with direct use of the `Task` class implementation.
The module `bar` is using facade pattern by inheriting this pattern.


jkonecny12:
Closing this PR  for now.

PR #1240, opened 2017-11-22
Tweak dbus code

jkonecny12:
Fix parts in the DBus implementation and do a small tweaks.

Also add missing Makefiles.

### DIFF
--- a/pyanaconda/dbus_addons/baz/__main__.py
+++ b/pyanaconda/dbus_addons/baz/__main__.py
@@ -1,4 +1,4 @@
 from pyanaconda.dbus_addons.baz.baz import Baz
 
-module = Baz()
-module.run()
+baz = Baz()
+baz.run()
--- a/pyanaconda/modules/boss/boss.py
+++ b/pyanaconda/modules/boss/boss.py
@@ -24,7 +24,6 @@ from gi.repository import GLib
 
 from pyanaconda.dbus import dbus_constants
 from pyanaconda.dbus.interface import dbus_interface
-from pyanaconda.dbus.typing import *  # pylint: disable=wildcard-import
 from pyanaconda.modules.base import BaseModule
 
 from pyanaconda.modules.boss.module_manager import ModuleManager  # pylint: disable=relative-beyond-top-level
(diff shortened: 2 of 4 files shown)


pyanaconda/dbus_addons/baz/__main__.py
```diff
@@ -1,4 +1,5 @@
 from pyanaconda.dbus_addons.baz.baz import Baz
 
-module = Baz()
-module.run()
+if __name__ == "__main__":
```
poncovka: This is not necessary here: https://docs.python.org/3.6/library/__main__.html

pyanaconda/modules/boss/boss.py
```diff
@@ -35,10 +34,11 @@
 @dbus_interface(dbus_constants.DBUS_BOSS_NAME)
 class Boss(BaseModule):
 
-    def __init__(self):
+    def __init__(self, module_manager=None):
```
poncovka: I am not sure if we want to do this with every manager.

configure.ac
```diff
@@ -132,6 +132,12 @@ AC_CONFIG_FILES([Makefile
                  pyanaconda/Makefile
                  pyanaconda/version.py
                  pyanaconda/dbus/Makefile
+                 pyanaconda/modules/Makefile
+                 pyanaconda/modules/boss/Makefile
+                 pyanaconda/modules/foo/Makefile
+                 pyanaconda/modules/bar/Makefile
+                 pyanaconda/dbus_addons/Makefile
+                 pyanaconda/dbus_addons/baz/Makefile
```
poncovka: I think that @M4rtinK has the makefiles ready in his next pull request.

pyanaconda/dbus_addons/baz/__main__.py
jkonecny12: Hmm, good to know. I'll remove this part then.

pyanaconda/modules/boss/boss.py
jkonecny12: Yep you want to. It is prevention from entangling the classes too much and be able to reuse the class for other purpose.

Also there shouldn't be more then a few managers. If there is more it should be redesigned a split otherwise you will get spaghetti at the end. Also it is helping to prevent Open Close Principle and SRP.

M4rtinK: What's the purpose of this ? Testing (eq. to more easily mock the manager class) ?

configure.ac
M4rtinK: Yep, I'll add it.

pyanaconda/modules/boss/boss.py
jkonecny12: The main reason is testing but also it is the correct way in general because you can replace parts used inside without changing the whole class. Look on the [Dependency Injection](https://en.wikipedia.org/wiki/Dependency_injection) if you want to find out more.

jkonecny12:
**UPDATED:**
* Makefiles commit removed
* `__main__` tests removed

M4rtinK:
jenkins, test this please

jkonecny12:
Jenkins, it's ok to test.

PR #1243, opened 2017-11-23
Run Boss & module prototypes in the installation environment

M4rtinK:
This pull request adds the necessary bits and pieces to run Boss and the module prototypes in the installation environment.

Boss is not yet started by default - `inst.run_boss` needs to be present on the boot command line.


jkonecny12:
Please fix these errors:
```
************* Module ../scripts/setup-updates
W0611(unused-import):  3,0: : Unused import sys
W0611(unused-import):  5,0: : Unused import pydbus
************* Module ../scripts/run_boss_locally.py
W0611(unused-import):  5,0: : Unused import shutil
```

data/systemd/Makefile.am
```diff
@@ -19,6 +19,7 @@ systemddir = $(prefix)/lib/systemd/system
 generatordir = $(prefix)/lib/systemd/system-generators
 
 dist_systemd_DATA = anaconda.service \
+					anaconda-boss.service \
```
jkonecny12: Please fix this indentation.

scripts/setup-updates
```diff
@@ -0,0 +1,78 @@
...
+                symlink_updates(dest_path, update_path)
+        else:
+            if f.endswith(".pyc") or f.endswith(".pyo"):
+                continue
+            os.symlink(dest_path, update_path)
+
+def setup_python_updates():
+    """Setup updates to Anaconda Python files."""
+    from distutils.sysconfig import get_python_lib
+    import gi.overrides
```
jkonecny12: These imports can be between the global imports.

scripts/setup-updates
```diff
@@ -0,0 +1,78 @@
...
+    # Temporary hack for F18 alpha to symlink updates and product directories
+    # into tmpfs.  To be removed after beta in order to directly use content
+    # from /run/install/ -- JLK
+    for dirname in ("updates", "product"):
+        if os.path.exists("/run/install/%s" % dirname):
+            if os.path.islink("/tmp/%s" % dirname):
+                # Assume updates have already been setup
+                return
+            os.symlink("/run/install/%s" % dirname,
+                       "/tmp/%s" % dirname)
```
jkonecny12: Can't we remove it now, when it was just a temporal hack which should be removed after beta?

scripts/setup-updates
```diff
@@ -0,0 +1,78 @@
...
+            if not os.access(dest, os.R_OK):
+                continue
+        # Symlink over everything that's in the python libdir but not in
+        # the updates directory.
+        symlink_updates(dest, d)
+
+    gi.overrides.__path__.insert(0, "/run/install/updates")
+
+    import glob
+    import shutil
```
jkonecny12: The same the imports should be on the top of the file.

scripts/setup-updates
```diff
@@ -0,0 +1,78 @@
...
+    for rule in glob.glob("/tmp/updates/*.rules"):
+        target = "/etc/udev/rules.d/" + rule.split('/')[-1]
+        shutil.copyfile(rule, target)
+
+print("setting up updates & product image")
+setup_python_updates()
+print("updates and product image setup")
```
jkonecny12: This should be in `__main__` part.

scripts/setup-updates
```diff
@@ -0,0 +1,78 @@
+#!/usr/bin/python3
```
jkonecny12: Thanks for doing this. It will help to clear the main script which is still too messy and hard to find anything in it.
Even if we don't need this in a future it is nice to have this as a separate script.

But wouldn't this be in conflict with the old variant when this is added but the old is not removed?

pyanaconda/dbus_addons/Makefile.am
```diff
@@ -0,0 +1,24 @@
+# dbus_addons/Makefile.am for anaconda
```
jkonecny12: You also need to add these Makefiles to the confiure.ac file.

scripts/makeupdates
```diff
@@ -52,6 +52,8 @@ VERSION_OP_MAP = {
     "<=": VERSION_LESS_OR_EQUAL
 }
 
+# Anaconda scripts that should be installed into the libexec folder
+LIBEXEC_SCRIPTS = ["zramswapon", "zramswapoff", "zram-stats", "log-capture", "setup-updates", "start-module"]
```
jkonecny12: Please wrap the line to 99 characters.

scripts/run_boss_locally.py
```diff
@@ -14,16 +14,23 @@
...
 DBUS_SERVICES_DIR = "../data/dbus/"
+STARTUP_SCRIPT = os.path.abspath("../scripts/start-module")
 
 print("creating a temporary directory for DBUS service files")
 temp_service_dir = tempfile.TemporaryDirectory(prefix="anaconda_dbus_")
 print(temp_service_dir.name)
 
-print("copying service files")
+print("copying & modifying DBUS service files")
+exec_path = 'Exec=/usr/libexec/anaconda/start-module'
```
jkonecny12: Please make this a constant variable.

pyanaconda/dbus_addons/Makefile.am
jkonecny12: You need to change `pyanaconda/Makefile.am` to add `SUBDIRS` there.

scripts/setup-updates
M4rtinK: The code seems to check for the folder to already exist before doing anything, so it should be fine.
My theory is that some other tool (Lorax ?) might be (or might have been) doing the updates/product image setup and the code needs to be able to handle that.

scripts/setup-updates
M4rtinK: Good point - another benefit of having the updates setup code as a separate script.

scripts/setup-updates
M4rtinK: Unfortunately it doesn't seem to work without the temporary hack - I assume something is using the the old & something the new location. But definitely something for a followup investigation & PR.

scripts/setup-updates
```diff
@@ -0,0 +1,78 @@
...
+
+import sys
+import os
+import pydbus
+
+def symlink_updates(dest_dir, update_dir):
```
jkonecny12: Indentation 2 lines for top level functions please.

scripts/setup-updates
```diff
@@ -0,0 +1,78 @@
...
+        if f in contents:
+            # recurse into directories, there might be files missing in updates
+            if os.path.isdir(dest_path) and os.path.isdir(update_path):
+                symlink_updates(dest_path, update_path)
+        else:
+            if f.endswith(".pyc") or f.endswith(".pyo"):
+                continue
+            os.symlink(dest_path, update_path)
+
+def setup_python_updates():
```
jkonecny12: Indentation 2 lines for top level functions please.

scripts/setup-updates
```diff
@@ -0,0 +1,78 @@
...
+
+    gi.overrides.__path__.insert(0, "/run/install/updates")
+
+    import glob
+    import shutil
+    for rule in glob.glob("/tmp/updates/*.rules"):
+        target = "/etc/udev/rules.d/" + rule.split('/')[-1]
+        shutil.copyfile(rule, target)
+
+print("setting up updates & product image")
```
jkonecny12: Indentation 2 lines for top level code please.

data/systemd/Makefile.am
M4rtinK: Fixed (looks like tabs instead of spaces get there somehow - weird).

pyanaconda/dbus_addons/Makefile.am
M4rtinK: Good catch - fixed!
But other than that - what insanity is this ? I kinda understand the idea that you have a Makefile.am in each folder, that specifies which files in the folder (if any) should be taken into account during build & defines any build-related subfolders (if any).

This should be enough for autotools to crawl the source tree, going over individual Makefile.am files. But it needs a static list of **all** the Makefiles as well ?? That's just totally redundant, fragile   and insane. :P

scripts/makeupdates
M4rtinK: Done.

scripts/run_boss_locally.py
M4rtinK: Done.

scripts/setup-updates
M4rtinK: Fixed.

scripts/setup-updates
M4rtinK: Fixed.

scripts/setup-updates
M4rtinK: Changed.

scripts/setup-updates
M4rtinK: Changed.

scripts/setup-updates
M4rtinK: Changed.

pyanaconda/dbus_addons/Makefile.am
jkonecny12: For that I could say only: "Die autotools..." <-- choose your language.

M4rtinK: It's "Die **A**utotools" in German. :)

scripts/setup-updates
jkonecny12: Sounds reasonable.

PR #1245, opened 2017-11-24
Master modularization kickstart filter

rvykydal:


pyanaconda/kickstart_dispatcher.py
```diff
@@ -0,0 +1,359 @@
...
+                name = ""
+            self._result.append_addon(name=name,
+                                      content=str(section),
+                                      header_lineno=section.header_lineno,
+                                      filename=self._current_ks_filename)
+        else:
+            self._result.append_section(name=section.args[0],
+                                        content=str(section),
+                                        header_lineno=section.header_lineno,
+                                        filename=self._current_ks_filename)
```
poncovka: You can set `name` to `section.args[0]` in the `else` branch and join the calls of `self._result.append_section`.

tests/pyanaconda_tests/kickstart_dispatcher.py
```diff
@@ -0,0 +1,514 @@
...
+        self.assertEqual(pony_addon, [element4])
+        self.assertEqual(elements.get_elements(), [])
+
+        # to get all elements of specified type use empty list as filter
+        all_addons_and_sections = elements.get_elements(sections=[], addons=[])
+        self.assertEqual(all_addons_and_sections, [element1, element4, element6, element7])
+
+        ### Tracking elements
+
+        processed_elements = elements.get_elements(commands=["network"], sections=["%pre"], addons=["pony"], process=True)
```
poncovka: It doesn't look very consistent, that a section has to start with `%` and an addon is missing `%addon`. Maybe modules could specify the kickstart elements with their "full" names. Then it could look like: 

`elements.get_elements(["network", "%pre", "%addon pony"], process=True)`

Then we don't have to care about the types of the elements we have and just keep their "full" names in `KickstartElement`. It will simplify the method `get_elements` as well.

Or we can just drop `%` from the section names.

pyanaconda/kickstart_dispatcher.py
```diff
@@ -0,0 +1,359 @@
...
+    def all_elements(self):
+        """List of all elements in the container.
+
+        :return: list of all elements in the container
+        :rtype: list(KickstartElement)
+        """
+
+        return list(self._elements)
+
+    def get_elements(self, commands=None, sections=None, addons=None):
```
poncovka: We probably shouldn't allow to request the `%addon` section in `sections`.

pyanaconda/kickstart_dispatcher.py
```diff
@@ -0,0 +1,355 @@
...
+        """List of all elements not tracked as processed.
+
+        :return: list of all elements not tracked as processed
+        :rtype: list(KickstartElement)
+        """
+        return [element for element in self._elements
+                if element not in self._processed_elements]
+
+
+class StoreSection(Section):
```
jkonecny12: This could be in separate file. It is not part using or handling `KickstartElements` class so why it should be here?

pyanaconda/kickstart_dispatcher.py
```diff
@@ -0,0 +1,355 @@
...
+        section to make it valid but just ignored. If you want to store the
+        content, supply a store argument.
+
+        Required kwargs:
+        sectionOpen - section name, including '%' starting character
+
+        Optional kwargs:
+        store - an instance of an object for storing the section
+                (SplitKickstartParser) which must provide
+                add_section(StoreSection) method
```
jkonecny12: Please merge this to two lines instead of three. I guess it should be doable in 99 characters limit.

pyanaconda/kickstart_dispatcher.py
```diff
@@ -0,0 +1,355 @@
...
+        super().__init__(*args, **kwargs)
+        self.sectionOpen = kwargs.get("sectionOpen")
+        self._store = kwargs.get("store", None)
+        self.header_lineno = 0
+        self.args = []
+        self._body = []
+
+    def handleHeader(self, lineno, args):
+        self.header_lineno = lineno
+        self.args = args
```
jkonecny12: What is `args`? I'm missing docstring here.

pyanaconda/kickstart_dispatcher.py
jkonecny12: I really hate these args solution it's so hard to find what means `section.args[0]` and `section.args[1]`.

pyanaconda/kickstart_dispatcher.py
```diff
@@ -0,0 +1,355 @@
...
+
+    Stores kickstart elements (commands, sections, addons) with their line
+    number and file name references to kickstart file.
+    Does not do any actual command or section parsing (ie command syntax
+    checking).
+
+    :raises KickstartParseError: on invalid section
+    :raises KickstartError: on missing %include unless instantiated with
+                            missing_include_is_fatal=False
+    """
```
jkonecny12: I'm able to imagine this as a library and this will be the interface of a library. I guess nothing else that this should be used outside in  the code. So why not to have this in `__init__.py` fine in different folder to separate all the implementation details (other classes) from developer when he/she wants to use this.

tests/pyanaconda_tests/kickstart_dispatcher.py
```diff
@@ -0,0 +1,486 @@
...
+        # process=False is just default (NOOP)
+        elements.get_elements(commands=["network"], sections=["%pre"], addons=["pony"], process=False)
+        self.assertEqual(elements.unprocessed_elements, unprocessed_elements)
+        # processing some more elements - firewall command
+        firewall_elements = elements.get_elements(commands=["firewall"], process=True)
+        self.assertEqual(set(elements.unprocessed_elements), set.difference(set(unprocessed_elements), set(firewall_elements)))
+        self.assertEqual(elements.unprocessed_elements, [element6, element7])
+        # now process everthing
+        elements.get_elements(commands=[], sections=[], addons=[], process=True)
+        self.assertEqual(elements.unprocessed_elements, [])
```
jkonecny12: Please split these to different test methods. These are already different tests but instead of `def...` they have comments between them.

tests/pyanaconda_tests/kickstart_dispatcher.py
```diff
@@ -0,0 +1,486 @@
...
+
+        ### Reading kickstart from string supplying filename
+
+        filename = "MY_FILENAME"
+        expected_result = get_expected_result(filename)
+
+        result = ksparser.split_from_string(ks_content, filename=filename)
+
+        self.assertEqual(result.all_elements, expected_result)
+        self.assertEqual(result.get_kickstart_from_elements(result.all_elements), ks_content)
```
jkonecny12: The same as above.

tests/pyanaconda_tests/kickstart_dispatcher.py
```diff
@@ -0,0 +1,486 @@
...
+
+        handler = returnClassForVersion()
+        ksparser = SplitKickstartParser(handler, valid_sections)
+
+        ### Reading kickstart from file
+
+        with open(filename, "w") as f:
+            f.write(ks_content)
+        result = ksparser.split(filename)
+        #os.remove(filename)
```
jkonecny12: Remove this completely I don't think there is a point on having code commented out.

tests/pyanaconda_tests/kickstart_dispatcher.py
```diff
@@ -0,0 +1,486 @@
...
+        expected_result = get_expected_result(filename)
+
+        result = ksparser.split_from_string(ks_content, filename=filename)
+
+        self.assertEqual(result.all_elements, expected_result)
+        self.assertEqual(result.get_kickstart_from_elements(result.all_elements), ks_content)
+
+
+    @contextmanager
+    def _create_ks_files(self, kickstart):
```
jkonecny12: It would be nice to have docstring here.

tests/pyanaconda_tests/kickstart_dispatcher.py
```diff
@@ -0,0 +1,486 @@
...
+
+        # Now do one more pass on resulting (flat) kickstart
+        result2 = ksparser.split_from_string(result1_kickstart)
+        # the result will be different in most cases because of different references
+        # but the kickstart should be the same now
+        result2_kickstart = result2.get_kickstart_from_elements(result2.all_elements)
+        self.assertEqual(result1_kickstart, result2_kickstart)
+
+    def split_kickstart_parser_test(self):
+        valid_sections = ["%pre", "%pre-install", "%post", "%onerror", "%traceback", "%packages", "%addon", "%anaconda"]
```
jkonecny12: Could you please hold limit 99 characters even for tests.

pyanaconda/kickstart_dispatcher.py
```diff
@@ -0,0 +1,355 @@
+# Anaconda kickstart dispatching for modules
```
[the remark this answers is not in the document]
rvykydal: Seemed like overkill to me at this point, but perhaps we should have "kickstart" directory (with parser, element subdirs) for this and at some point move all kickstart handling from kickstart.py there?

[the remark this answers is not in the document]
jkonecny12: That sounds great.

pyanaconda/kickstart_dispatcher.py
```diff
@@ -0,0 +1,355 @@
...
+        :param header_lineno: line number of the addon section header in a kickstart file
+        :type content: int
+        :param filename: name of a kickstart file
+        :type filename: str
+        """
+        addon = KickstartElement(args=("%addon", name),
+                                 content=content,
+                                 lineno=header_lineno,
+                                 filename=filename)
+        self._elements.append(addon)
```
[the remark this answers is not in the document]
rvykydal: I'll try this approach, I was thinking about having a class for element. Then it could/should also take care also of tracking (by setting a flag on the instance). Now the tracking takes advantage to element being immutable, so it is possible to use set for tracking processed elements in TrackedKickstartElements.

pyanaconda/kickstart_dispatcher.py
rvykydal: Hmm, I agree.

pyanaconda/kickstart_dispatcher.py
```diff
@@ -0,0 +1,359 @@
...
+                    continue
+                if sections == [] or name in sections:
+                    filtered_elements.append(element)
+            # command
+            else:
+                if commands is None:
+                    continue
+                if commands == [] or name in commands:
+                    filtered_elements.append(element)
+        return filtered_elements
```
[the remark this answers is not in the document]
rvykydal: I agree. What about having KickstartElementCommand etc subclasses then?

pyanaconda/kickstart_dispatcher.py
```diff
@@ -0,0 +1,355 @@
...
+
+
+class TrackedKickstartElements(KickstartElements):
+    """Container for kickstart elements with tracking."""
+    def __init__(self):
+        super().__init__()
+        self._processed_elements = set()
+
+    def get_elements(self, commands=None, sections=None, addons=None, process=False):
+        """Returns selected elements and tracks them as processed.
```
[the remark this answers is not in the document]
rvykydal: Yeah, I didn't want the line too long, I'd append 'optionally'.

pyanaconda/kickstart_dispatcher.py
rvykydal: I think i'll have one file for elements/tracking, another for kickstart parsing (including the section)

pyanaconda/kickstart_dispatcher.py
rvykydal: It is just overriding the parent method. Should I document it here?

pyanaconda/kickstart_dispatcher.py
rvykydal: I see the interface as one layer above.
This is just modified KickstartParser, although it seems a bit like providing THE interface.

pyanaconda/kickstart_dispatcher.py
rvykydal: Yes, I should consolidate this.

tests/pyanaconda_tests/kickstart_dispatcher.py
rvykydal: Actually I should uncomment it :)

pyanaconda/kickstart_dispatcher.py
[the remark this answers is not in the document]
rvykydal: hm, it would conflict with kickstart.py

pyanaconda/kickstart_dispatcher.py
```diff
@@ -0,0 +1,355 @@
...
+
+    def get_elements(self, commands=None, sections=None, addons=None):
+        """Returns selected elements.
+
+        :param commands: names of commands to be returned or [] for all commands
+        :type commands: list
+        :param sections: names of sections to be returned or [] for all sections
+        :type sections: list
+        :param addons: names of addons to be returned or [] for all addons
+        :type addons: list
```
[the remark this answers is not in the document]
rvykydal: Using [] seems non-intuitive to me as well, but it should be actually rare use case. Usually you'll want just (module-, or boss-) specific elements.
The reason for having get_elements is that it gives you all the elements in the order of appearance in kickstart.
which is important property when splitting kickstart to elements IMO. Having this you can also easily construct the filtered kickstart by using get_elements() and then feeding the elements to get_kickstart_from_elements.
You wouldn't be able to do this (keep order of elements) easily with separate get_commands / get_sections / get_addons methods. Merging would be quite complicated - you have order references in elements but you need to take into account also include files which would be pretty complicated.

I think a list models a split file more naturally. This is really basic processing and get_elements (or get_commands, ...) is more one kind of view. Another is get_kickstart_from_elements.

tests/pyanaconda_tests/kickstart_dispatcher.py
rvykydal: You are right.
I think I'll drop the % and try to work just with names and (typed) elements.

The
`elements.get_elements(["network", "%pre", "%addon pony"], process=True)`
seems dangerously low-level for me thinking about it now (like client needs to construct "%addon pony" if she wants a "pony" addon). We need to decompose / compose the "%addon pony" anyway when working with kickstart as the parser gives us only already parsed (shlex) args of section header (not the line itself) in the method overriden by the splitter. So I think we shouldn't work with kickstart line beginnings containing implicit element type info but with some elementary abstractions (element types:) - which would be also safer for future if some more complicated element is added to pykickstart.

poncovka: Ok.

pyanaconda/kickstart_dispatcher.py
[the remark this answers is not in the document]
jkonecny12: I would propose to just slightly change the name now to avoid conflict and rename it later when the kickstart.py will be moved in. What about `kickstart_handler` or `kickstart_dispatcher` for now?

pyanaconda/kickstart_dispatcher.py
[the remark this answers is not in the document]
jkonecny12: As we discussed, please remove the `[]` completely, and add an info about order to method docstring please.

pyanaconda/kickstart_dispatcher.py
jkonecny12: I think it should be here. You are doing here something else than the parent did (reason for overriding). But I guess it's not that mandatory.

tests/pyanaconda_tests/kickstart_dispatcher.py
jkonecny12: That is the solution too :)

rvykydal:
Notable changes:

- splitting into files and moving to pyanaconda/kickstart_dispatcher (may be ranamed to kickstart when we can remove kickstart.py)
- KickstartElement changed from namedtuple to a class encapsulating also parsing of output supplied by KickstartParser
- see comment for TrackedKickstartElements.get_and_process_element

pyanaconda/kickstart_dispatcher/element.py
```diff
@@ -0,0 +1,239 @@
...
+        return str(self._elements)
+
+
+class TrackedKickstartElements(KickstartElements):
+    """Container for kickstart elements with tracking."""
+    def __init__(self):
+        super().__init__()
+        self._processed_elements = set()
+
+    def get_and_process_elements(self, commands=None, sections=None, addons=None):
```
rvykydal: This used to be get_elements(self, commands=None, sections=None, addons=None, process=False) in the first PR, adding process option to the parent method signature, but pylint doesn't like changing the signature in overriden method so I used this.
I am not satisfied with method name name, get_and_track_elements may be better, plus maybe changing unprocessed_elements() to untracked_elements()

tests/pyanaconda_tests/kickstart_dispatcher.py
```diff
@@ -470,7 +470,7 @@ def simple_split_kickstart_parser_test(self):
 
         valid_sections = VALID_SECTIONS_ANACONDA
 
-        handler = returnClassForVersion()
+        handler = returnClassForVersion()()
```
poncovka: Actually, you can use `makeVersion`.

rvykydal: Nice, I'll update the PR, thank you!

pyanaconda/kickstart_dispatcher/element.py
```diff
@@ -0,0 +1,239 @@
...
+        :param args: tokens parsed from a line by shlex in KickstartParser.
+                     - for command the whole command
+                     - for section its header
+                     - for addon its header
+        :type args: list(str)
+        :param lines: lines parsed by KickstarParser
+                     - for command the whole command
+                     - for section its body
+                     - for addons its body
+        :type lines: list(str)
```
jkonecny12: I think `content` would be more appropriate than `lines` and save the original to something like `original_content` or better don't save it at all, what do you think?

pyanaconda/kickstart_dispatcher/element.py
```diff
@@ -0,0 +1,239 @@
...
+
+    @property
+    def unprocessed_elements(self):
+        """List of all elements not tracked as processed.
+
+        :return: list of all elements not tracked as processed
+        :rtype: list(KickstartElement)
+        """
+        return [element for element in self._elements
+                if element not in self._processed_elements]
```
jkonecny12: I would change this to:
```
elements = set(self._elements)
return self._processed_elements - elements
```
Or even better change the `self._elements` parent's variable to `set`.

rvykydal: That would not preserve the order of elements (also we are appending, not adding elements to the container).

pyanaconda/kickstart_dispatcher/element.py
rvykydal: I think content would be misleading here: the problem is that kickstart parser gives us different data for commands and sections. For commands - line is the content, for sections - lines is only the body content (there is no header line supplied, only in form of args - shlex parsed line).  If we find we want to construct KickstartElements in other contexts than kickstart parsing, let's add support/api for it later. (Trying to be generic when creating element I ended up with something like v1 of the patch was (append_command, append_section, ...). But I think we should be careful with allowing to construct the Elements in other ways than as it is done from kickstart parsing. The KickstartElement evolved to a class having also the knowledge about kickstart parser parsing callbacks output, and even re-composing kickstart content from parser output (for sections). And I think it is OK as the purpose of KickstartElements is to be basically a read-once structured storage for kickstart.

pyanaconda/kickstart_dispatcher/element.py
jkonecny12: Ohh ok, I didn't know the order is required here too.

pyanaconda/kickstart_dispatcher/element.py
jkonecny12: Ok, thanks for explanation.

PR #1247, opened 2017-11-27
Rhel8 alpha branch dist git patches update

rvykydal:
This should add f27 dist-git patches on top of the 27.20.4.
I'll then make zanata.ini static file containing the proper branch name (for this branch)
and build as 27.20.4.1

### DIFF
--- a/pyanaconda/bootloader.py
+++ b/pyanaconda/bootloader.py
@@ -1825,7 +1825,6 @@ class EFIGRUB1(EFIBase, GRUB):
 
 class EFIGRUB(EFIBase, GRUB2):
     _packages32 = [ "grub2-efi-ia32", "shim-ia32" ]
-    _packages64 = [ "grub2-efi-x64", "shim-x64" ]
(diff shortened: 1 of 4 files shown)


pyanaconda/bootloader.py
```diff
@@ -1861,12 +1861,18 @@ def packages(self):
...
 
+    def __init__(self):
+        super(Aarch64EFIGRUB, self).__init__()
+        self._packages64 = ["grub2-efi-aa64", "shim-aa64"]
+
 class MacEFIGRUB(EFIGRUB):
-    _packages64 = [ "grub2-tools-efi", "mactel-boot" ]
+    def __init__(self):
+        super(MacEFIGRUB, self).__init__()
+        self._packages64.extend(["grub2-tools-efi", "mactel-boot"])
```
jkonecny12: I really don't like this patch for alpha branch it is making instance variables from class variables so it is bug prone. I would like to see a clear solution to master and here. However, yeah it's already there so we can't do much about it here.

PR #1185, opened 2017-09-13
Migrate fedora-server installclass into the anaconda source repositor…

rvykydal:
…y (#1466967)

We can remove it from fedora-productimg-server then.

PR #1249, opened 2017-11-28
Add fedora-server.css

sgallagher:
PR https://github.com/rhinstaller/anaconda/pull/1185 moved the
fedora-server.py InstallClass to Anaconda, but did not move the
corresponding CSS definition referenced in the InstallClass.

Once this is merged, we can retire the fedora-productimg-server
package from Fedora.

### DIFF
--- a/data/fedora-server.css
+++ b/data/fedora-server.css
@@ -1,0 +1,63 @@
+@define-color fedora #2f4265;
+
+/* logo and sidebar classes for Fedora */
+
+/* The sidebar consists of three parts: a background, a logo, and a product logo,
+ * rendered in that order. The product logo is empty by default and is intended
+ * to be overridden by a stylesheet in product.img.
+ */
+.logo-sidebar {
+    background-image: url('/usr/share/anaconda/pixmaps/server/sidebar-bg.png');
+    background-color: @fedora;
+    background-repeat: no-repeat;
+}
+
+/* Add a logo to the sidebar */
+.logo {
+    background-image: url('/usr/share/anaconda/pixmaps/server/sidebar-logo.png');
+    background-position: 50% 20px;
+    background-repeat: no-repeat;
+    background-color: transparent;
+}
+
+/* This is a placeholder to be filled by a product-specific logo. */
+.product-logo {
+    background-image: none;
+    background-color: transparent;
+}
+
+AnacondaSpokeWindow #nav-box {
+    background-color: @fedora;
+    background-image: url('/usr/share/anaconda/pixmaps/server/topbar-bg.png');
+    background-repeat: no-repeat;
+    color: white;
+}
+
+/* Remove the box-shadow from buttons in the nav-box because it adds a white stripe
+ * below the buttons and makes them look dumb */
+AnacondaSpokeWindow #nav-box GtkButton {
+    box-shadow: none;
+}
+
+
+/* Change the sidebar background color */
+.logo-sidebar {
+  background-color: #006eb4;
+}
+
+AnacondaSpokeWindow #nav-box {
+  background-color: #006eb4;
+}
+
+/* Move the Fedora logo to the bottom */
+.logo {
+    background-position: 50% 90%;
+}
+
+/* Add the product logo at the top */
+.product-logo {
+    background-image: url('/usr/share/anaconda/pixmaps/server/sidebar-logo_flavor.png');
+    background-position: 12px 10.5em;
+    background-repeat: no-repeat;
+}
+
--- a/data/Makefile.am
+++ b/data/Makefile.am
@@ -21,7 +21,8 @@ CLEANFILES = *~
 
 dist_pkgdata_DATA          = interactive-defaults.ks \
 			     tmux.conf \
-			     anaconda-gtk.css
+			     anaconda-gtk.css \
+			     fedora-server.css
 
 helpdir               = $(datadir)/$(PACKAGE_NAME)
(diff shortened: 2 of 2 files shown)


jkonecny12:
Thanks for this PR. However, I don't think we want this in our code. The point is that the install class is interconnected with Anaconda but the css file is not.

Also we will fix the link in the install class but still css files are working with your png files from the product image. We definitely don't want to have these in Anaconda.

So the resolution -- I'm closing this PR and we will implement fallback to default graphics in case this CSS file is missing in the stage2.

rvykydal:
Yes, the fedora-productimg-server package also sets symlinks to proper product-specific images (although they come from yet another package).

sgallagher:
@rvykdal Those are actually unneeded and wrong (which is why I fixed those paths in this PR).

This is no different from what anaconda-gtk.css is doing as far as I can see. It also references images provided by the Fedora-logos package.

sgallagher:
@rvykydal My thoughts are that this DOES belong in Anaconda, because it's supplementary to anaconda-gtk.css. I'd rather that if anaconda-gtk.css undergoes changes, there's a single place to also change the supplement.

sgallagher:
FWIW, I opened this and submitted it because @M4rtinK asked me to (he agreed with me that it seemed more appropriate to be in the anaconda repository).
I'd much prefer to be able to retire the `fedora-productimg-server` package, since it really doesn't offer any value that couldn't just be carried upstream at this point.

M4rtinK:
It looks like I didn't have enough information - on a second look it much less clear to me now if the CSS file should be in Anaconda as well (for example - we have CentOS/Scientific Linux install class, but their CSS is separate). 

But I think that does not mean the `fedora-productimg-server` needs to still exist - the install class sets the path to the CSS file, so as long as it's available in the installation environment it should work (or fall back, as mentioned above). So the CSS could be moved to some other package (Fedora release/Fedora icons ?) and the product image can then be decommissioned.

rvykydal:
> @rvykydal Those are actually unneeded and wrong (which is why I fixed those paths in this PR).
> 
> This is no different from what anaconda-gtk.css is doing as far as I can see. It also references images provided by the Fedora-logos package.

Hm, from what I see, the product-specific .css files (fedora-server.css, fedora-workstation.css ..) refer all to the same file path for product-specific logo (eg /usr/share/anaconda/pixmaps/sidebar-logo_flavor.png). The mechanism by which Server shows server-specific image is that the fedora-productim-server package creates a symlink to (/usr/share/anaconda/pixmaps/server/sidebar-logo.png) at this path. So removing the package and just adding server .css to anaconda there would be no product-specific logo displayed.

And the productimg package creates also sidebar-bg.png and topbar-bg.png symlinks (referring to server specific folder in fedora-logos).

rvykydal:
So one option could be to make the .css refer the files from fedora-logos package located at /usr/share/anaconda/pixmaps/server directly and move the .css somewhere - anaconda? fedora-logos?

data/fedora-server.css
```diff
@@ -0,0 +1,63 @@
+@define-color fedora #2f4265;
+
+/* logo and sidebar classes for Fedora */
+
+/* The sidebar consists of three parts: a background, a logo, and a product logo,
+ * rendered in that order. The product logo is empty by default and is intended
+ * to be overridden by a stylesheet in product.img.
+ */
+.logo-sidebar {
+    background-image: url('/usr/share/anaconda/pixmaps/server/sidebar-bg.png');
```
rvykydal: symlink to fedora-logos created by productimg package

data/fedora-server.css
```diff
@@ -0,0 +1,63 @@
...
+
+/* This is a placeholder to be filled by a product-specific logo. */
+.product-logo {
+    background-image: none;
+    background-color: transparent;
+}
+
+AnacondaSpokeWindow #nav-box {
+    background-color: @fedora;
+    background-image: url('/usr/share/anaconda/pixmaps/server/topbar-bg.png');
```
rvykydal: symlink to fedora-logos created by productimg package

data/fedora-server.css
```diff
@@ -0,0 +1,63 @@
...
+}
+
+/* Move the Fedora logo to the bottom */
+.logo {
+    background-position: 50% 90%;
+}
+
+/* Add the product logo at the top */
+.product-logo {
+    background-image: url('/usr/share/anaconda/pixmaps/server/sidebar-logo_flavor.png');
```
rvykydal: symlink to fedora-logos created by productimg package

rvykydal:
> So one option could be to make the .css refer the files from fedora-logos package located at /usr/share/anaconda/pixmaps/server directly and move the .css somewhere - anaconda? fedora-logos?

Asking what fedora-logos thinks:
https://pagure.io/fedora-logos/issue/1

data/fedora-server.css
sgallagher: This is not a symlink created by productimg package. This is the literal path to the content in fedora-logos.

The productimg package *does* create symlinks right now, but this was a mistake and I fixed it in this PR. I'm not sure why I did that originally, but it was incorrect. I suspect it was because the CSS was written by someone else (modified from the official one) and I opted to move the images into the locations expected by the CSS instead of fixing the CSS.

sgallagher:
@rvykydal I've merged the CSS into fedora-logos and built it. I also sent a PR [1] to fix some paths.

[1] https://github.com/rhinstaller/anaconda/pull/1254

PR #1250, opened 2017-11-29
 Add the boot options inst.stage2.all and inst.ks.all

poncovka:
Anaconda will try to fetch stage2 from all defined locations that
are http, https or ftp if the boot option inst.stage2.all is set.
Other types of locations will be ignored.

(cherry-picked from a commit 6aba174)

Anaconda will try to fetch a kickstart file from all defined locations
that are http, https or ftp if the boot option inst.ks.all is set.
Other types of locations will be ignored.

(cherry-picked from a commit 98a1f0f)

### DIFF
--- a/dracut/anaconda-lib.sh
+++ b/dracut/anaconda-lib.sh
@@ -101,6 +101,46 @@ anaconda_live_root_dir() {
     anaconda_mount_sysroot $img
 }
 
+anaconda_net_root() {
+    local repo="$1"
+    info "anaconda: fetching stage2 from $repo"
+
+    # Try to get the local path to stage2 from treeinfo.
+    treeinfo=$(fetch_url $repo/.treeinfo 2> /tmp/treeinfo_err) && \
+        stage2=$(config_get stage2 mainimage < $treeinfo)
+
+    # No treeinfo available.
+    [ -z "$treeinfo" ] && debug_msg $(cat /tmp/treeinfo_err)
+
+    # Use the default local path to stage2.
+    if [ -z "$treeinfo" -o -z "$stage2" ]; then
+        warn "can't find installer main image path in .treeinfo"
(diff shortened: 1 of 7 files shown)


dracut/anaconda-lib.sh
```diff
@@ -101,6 +101,46 @@ anaconda_live_root_dir() {
...
+    # Try to get the local path to stage2 from treeinfo.
+    treeinfo=$(fetch_url $repo/.treeinfo 2> /tmp/treeinfo_err) && \
+        stage2=$(config_get stage2 mainimage < $treeinfo)
+
+    # No treeinfo available.
+    [ -z "$treeinfo" ] && debug_msg $(cat /tmp/treeinfo_err)
+
+    # Use the default local path to stage2.
+    if [ -z "$treeinfo" -o -z "$stage2" ]; then
+        warn "can't find installer mainimage path in .treeinfo"
```
jkonecny12: I have a feeling that `mainimage` is typo and I guess we can fix it now. What do you think?

dracut/anaconda-lib.sh
```diff
@@ -230,6 +270,27 @@ online_netdevs() {
     done
 }
 
+# Filter locations that are http, https or ftp urls.
+get_urls() {
+    local locations="${1}"
+    local location
+
+    # No locations.
+    [ -z "$locations" ] && return
```
jkonecny12: I guess you can remove this. When you have the `for` cycle below it will be skipped there without an additional code.

dracut/anaconda-lib.sh
poncovka: Fixed.

dracut/anaconda-lib.sh
poncovka: Fixed.

PR #1251, opened 2017-11-30
Use pykickstart 3

poncovka:
**DO NOT MERGE**
Adapt anaconda's commands to pykickstart 3.

We cannot do deepcopy of the method command since pykickstart 3,
because it tries to do deepcopy of patterns. However, we shouldn't
do it anyway and the code where we used it should be rewritten.

TODO: Run the kickstart tests.

NOTE: The kdump addon fails when it tries to parse its section in
a kickstart file. However, when it is not specified, the installer is fine.

**Requires: pykickstart-3**

### DIFF
--- a/anaconda.spec.in
+++ b/anaconda.spec.in
@@ -37,7 +37,7 @@ Source0: %{name}-%{version}.tar.bz2
 %define mehver 0.23-1
 %define nmver 1.0
 %define partedver 1.8.1
-%define pykickstartver 2.44-1
+%define pykickstartver 3.10-1
 %define pypartedver 2.5-2
 %define rpmver 4.10.0
 %define simplelinever 0.8-1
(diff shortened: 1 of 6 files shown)


M4rtinK:
Kdump addon PR for compatibility with Pykickstart 3:
https://github.com/daveyoung/kdump-anaconda-addon/pull/16

M4rtinK:
jenkins, test this please

poncovka:
Rebased, but there is still the problem with deep copy of kickstart data.

poncovka:
Added a fix for deepcopy of the kickstart data in the storage.

poncovka:
jenkins, test this please

M4rtinK:
BTW, this is the deepcopy upstream Python issue: https://bugs.python.org/issue32668

Looks like we might be lucky and it is already fixed in latest Python versions.

poncovka:
Jenkins, test this please.

PR #1252, opened 2017-11-30
Replace get_bus with the class DBus 

poncovka:
We should call static methods of DBus class instead of get_bus. The
class DBus keeps one instance of the DBus connection that is created
on demand. It also provides methods for easy publishing and creating
proxies.

We should never keep our own instances of the DBus connection unless
we have a very good reason for it.

The module pyanaconda.dbus.dbus_constants was renamed
to pyanaconda.dbus.constants.

The PYTHONPATH should be only modified in run_boss_locally, not
replaced, so we can change it and add our local repositories.

### DIFF
--- a/pyanaconda/dbus/__init__.py
+++ b/pyanaconda/dbus/__init__.py
@@ -1,21 +1,109 @@
+#
+# Representation of DBus connection.
+#
+# Copyright (C) 2017  Red Hat, Inc.  All rights reserved.
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+
 import os
 import pydbus
 
 from pyanaconda.anaconda_loggers import get_module_logger
 log = get_module_logger(__name__)
 
-def get_bus():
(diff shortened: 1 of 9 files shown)


pyanaconda/dbus/__init__.py
```diff
@@ -1,21 +1,86 @@
...
-    if bus_address:
-        return pydbus.connect(os.environ["DBUS_STARTER_ADDRESS"])
-    else:
-        log.info("DBUS_STARTER_ADDRESS not defined, using system bus")
+    _connection = None
+
+    @staticmethod
+    def get_connection():
+        """Returns a DBus connection."""
+        if not DBus._connection:
```
M4rtinK: I wonder if there is a chance for a race condition ? Eq. two requests getting two different connections, with the last one being cached.

pyanaconda/dbus/__init__.py
```diff
@@ -1,21 +1,86 @@
...
+
+        if bus_address:
+            log.info("Connecting to dbus at %s.", bus_address)
+            return pydbus.connect(bus_address)
+
+        log.info("Connecting to system dbus.")
         return pydbus.SystemBus()
 
+    @staticmethod
+    def publish(bus_name, *objects):
```
M4rtinK: So I guess this makes it possible to publish multiple class instances under the same dbus_name ? I wonder what's the use-case for that ?

pyanaconda/dbus/__init__.py
poncovka: I don't think that could be a problem. The cached connection is just recommended to use, but it is not required.

poncovka:
Reworking to add support for object paths.

poncovka:
Updated.

poncovka:
Updated.

pyanaconda/dbus/constants.py
```diff
@@ -17,6 +17,8 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #
 
+from pydbus.auto_names import auto_object_path
```
jkonecny12: It would be nice to have comment describing why we need it here. What's the reason.

PR #1253, opened 2017-12-01
Master add dbus tasks

jkonecny12:
Implement Task for example modules and Boss.

Add Task with interface and InstallManager with interface. Add tests for them.


pyanaconda/task/task.py
```diff
@@ -0,0 +1,203 @@
...
+        self.__cancel = False
+
+        self._task_number = Task._task_counter
+        Task._task_counter += 1
+
+        self._running_changed_signal = None
+        self._progress_changed_signal = None
+        self._error_raised_signal = None
+
+        self._dbus_name = "{}.{}".format(dbus_modul_path, self._task_number)
```
poncovka: I think that we should find a better solution for publishing objects and creating and keeping DBus names. But that is for another issue.

pyanaconda/async_utils.py
```diff
@@ -0,0 +1,86 @@
...
+#
+
+import gi
+
+from queue import Queue
+from pyanaconda.threading import threadMgr
+
+gi.require_version("GLib", "2.0")
+
+from gi.repository import GLib
```
poncovka: I am thinking if we could wrap all GLib functions we use in anaconda, so we don't have to do the crazy import every time. Just an idea for another issue.

pyanaconda/modules/boss/install_manager/installation_interface.py
```diff
@@ -0,0 +1,142 @@
...
+
+from pydbus.error import map_error
+
+from pyanaconda.dbus import get_bus
+from pyanaconda.dbus.dbus_constants import DBUS_BOSS_INSTALLATION_NAME
+from pyanaconda.dbus.interface import dbus_interface
+from pyanaconda.dbus.typing import *  # pylint: disable=wildcard-import
+
+
+@map_error("{}.InstallationNotRunning")
```
poncovka: Missing `format`?

pyanaconda/async_utils.py
jkonecny12: I have this as a future plan. We really need a wrapper for the GLib.

pyanaconda/modules/base.py
```diff
@@ -51,10 +59,25 @@ def run(self):
...
+        get_bus().publish(self._dbus_name, self)
         log.debug("module %s has been published", self._dbus_name)
 
+    def stop_module(self):
+        log.debug("module quiting: %s", self.dbus_name)
+        GLib.timeout_add_seconds(1, self.loop.quit)
+
+
+@dbus_interface(dbus_constants.DBUS_MODULE_NAMESPACE)
+class BaseModuleInterface(BaseModule, ABC):
```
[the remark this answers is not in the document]
jkonecny12: I have this on my TODO list already :)

pyanaconda/modules/boss/install_manager/installation_interface.py
```diff
@@ -0,0 +1,142 @@
...
+
+from pydbus.error import map_error
+
+from pyanaconda.dbus import get_bus
+from pyanaconda.dbus.dbus_constants import DBUS_BOSS_INSTALLATION_NAME
+from pyanaconda.dbus.interface import dbus_interface
+from pyanaconda.dbus.typing import *
+
+
+@map_error("{}.InstallationNotRunning")
```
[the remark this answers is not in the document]
jkonecny12: ups :D

pyanaconda/modules/foo/foo.py
```diff
@@ -48,6 +48,14 @@ def EchoString(self, s: Str) -> Str:
         log.debug(s)
         return s
 
+    def AvailableTasks(self) -> List((Str, Str)):
+        ret = List()
+
+        for task in self._tasks:
+            ret.append((task.name, task.dbus_name))
+
+        return ret
```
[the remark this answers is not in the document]
jkonecny12: Yep, but I think it should be done in the PR with BOSS interface splitting and similar stuff.

pyanaconda/task/task.py
jkonecny12: Could be nice topic for a future discussion.

pyanaconda/task/task_interface.py
```diff
@@ -0,0 +1,132 @@
...
+        return self._task_instance.description
+
+    @property
+    def Progress(self) -> Tuple[Int, Str]:
+        """Get immediate progress of this task.
+
+        :returns: Tuple with actual step count and description of this step.
+        """
+        return self._task_instance.progress
+
```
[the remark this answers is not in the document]
jkonecny12: I've missed that. Thanks!

jkonecny12:
**UPDATED:**
* Rebased on the new DBus class solution.
* Many and many other fixes.

pyanaconda/modules/boss/boss.py
```diff
@@ -38,15 +40,24 @@
...
+    def __init__(self, module_manager=None, install_manager=None):
         super().__init__()
+        self._module_manager = module_manager or ModuleManager()
+        self._install_manager = install_manager or InstallManager()
 
-        if module_manager is None:
-            self._module_manager = ModuleManager()
+    def _setup_install_manager(self):
+        # set modules to the installation manager
+        modules = self._module_manager.running_module_services
```
poncovka: This is really weird, because `running_module_services` is empty at this point. It works only because the managers share the same list of modules (the object).

pyanaconda/modules/boss/install_manager/install_manager.py
```diff
@@ -0,0 +1,264 @@
...
+            log.debug("Don't run another task. The installation was terminated.")
+            return
+
+        task_name = self._actual_task.Name
+
+        log.debug("Running installation task %s", task_name)
+        self._disconnect_task()
+        self._connect_task()
+        self._task_changed_signal.emit(task_name)
+        self._actual_task.Run()
```
poncovka: Shouldn't be this done asynchronously? I don't think you are able to do anything else while the task is running.

pyanaconda/modules/boss/install_manager/install_manager.py
```diff
@@ -0,0 +1,264 @@
...
+        else:
+            log.warn(log_msg)
+            return False
+
+    def _task_running_changed(self, is_running):
+        if not is_running:
+            self._tasks_done_step += self._actual_task.ProgressStepsCount
+            if self._tasks:
+                self._actual_task = self._tasks.pop()
+                self._run_task()
```
poncovka: Shouldn't we run next task from a callback of the `Run` method?

pyanaconda/modules/boss/install_manager/install_manager.py
```diff
@@ -0,0 +1,264 @@
...
+        if self._test_if_running("Can't get task progress when installation is not running."):
+            (step, msg) = self._actual_task.Progress
+            actual_progress = step + self._tasks_done_step
+
+            return actual_progress, msg
+        else:
+            return 0, ""
+
+    @property
+    def progress_float(self):
```
poncovka: Couldn't be the floating progress part of the regular progress?

jkonecny12:
**UPDATED:**
* Tests are fixed.

pyanaconda/modules/boss/install_manager/install_manager.py
poncovka: Ok, this is not true, because the Run method only starts the task. I am not sure if it shouldn't be Start method.

pyanaconda/modules/boss/install_manager/install_manager.py
poncovka: Based on the previous comment, this is also not true. However, I would prefer signals Started and Done instead of RunningChanged.

pyanaconda/modules/boss/boss.py
jkonecny12: Aha I've missed that, but this must be changed in overall structure. I'm not sure if `ModuleManager` should be passed to the `InstallManager` but I don't like this solution. I think there should be something more general to move modules around.

pyanaconda/modules/boss/install_manager/install_manager.py
jkonecny12: Could be renamed, I don't mind.

pyanaconda/modules/boss/install_manager/install_manager.py
jkonecny12: I don't think there is a reason to use both in one place ever. So you will use only the one you want to use.

pyanaconda/modules/boss/boss.py
poncovka: Maybe just add there a comment for now. We need to find a way how to synchronize the boss anyway.

pyanaconda/modules/boss/install_manager/install_manager.py
poncovka: Ok.

pyanaconda/modules/boss/boss.py
jkonecny12: Ok, I'll add a comment.

jkonecny12:
**UPDATED:**
* The `Task` `Run()` method is renamed to `Start()`.
* The `Task` `RunningChanged` signal changed to `Started()` and `Stopped()` signals.
* Similar changes in the `InstallManager` class and interface
* Add FIXME comment to setup of `InstallManager` in the Boss.

PR #1254, opened 2017-12-04
Point at new path for fedora-server.css

sgallagher:
In light of https://pagure.io/fedora-logos/pull-request/2 being merged into Anaconda, we just need to adjust the path that the `fedora-server.py` looks for the CSS file.

### DIFF
--- a/pyanaconda/installclasses/fedora-server.py
+++ b/pyanaconda/installclasses/fedora-server.py
@@ -30,7 +30,7 @@ __all__ = ["FedoraServerInstallClass"]
 
 class FedoraServerInstallClass(FedoraBaseInstallClass):
     name = "Fedora Server"
-    stylesheet = "/usr/share/anaconda/fedora-server.css"
+    stylesheet = "/usr/share/anaconda/pixmaps/server/fedora-server.css"
     defaultFS = "xfs"
     sortPriority = FedoraBaseInstallClass.sortPriority + 1
     if not productName.startswith("Fedora Server"):          # pylint: disable=no-member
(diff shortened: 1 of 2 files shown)


jkonecny12:
Jenkins, it's ok to test.

pyanaconda/installclasses/fedora-server.py
```diff
@@ -30,7 +30,7 @@
 
 class FedoraServerInstallClass(FedoraBaseInstallClass):
     name = "Fedora Server"
-    stylesheet = "/usr/share/anaconda/fedora-server.css"
+    stylesheet = "/usr/share/anaconda/pixmaps/server/fedora-server.css"
```
M4rtinK: It seems kinda weird to me to place a css file into a folder called pixmaps, which I would expect to contain bitmap images, possibly with some SVG files as well (as those result in pixmaps when rendered).

sgallagher: This is where it is stored in fedora-logos at the moment (because there is no more appropriate location in that package, currently). If that changes, we can update this path as well.

M4rtinK: Fair enough.

pyanaconda/installclasses/fedora-workstation.py
```diff
@@ -0,0 +1,6 @@
+from pyanaconda.installclasses.fedora import FedoraBaseInstallClass
```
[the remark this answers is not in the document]
sgallagher: Done

PR #1260, opened 2017-12-08
Tweak run_boss_locally script

jkonecny12:
The script can be started from anywhere now.
The bus address will be printed and wait for user to copy it and connect to the buss.
The bus address will be printed green if `python3-colorama` package is present, otherwise the warning about missing package will be printed but the script should work as usual.


jkonecny12:
*UPDATED:*
* Add controllable loop where you can just restart Boss without quitting the DBus session.
* By typing 'q' in the loop the DBuss session will quit.
* By pressing enter with anything else the DBus session will restart modules.

PR #1262, opened 2017-12-08
Add DBus observers for better access to proxies

poncovka:
We should use DBusServiceObserver and DBusObjectObserver to check
the availability of a service and to access its objects.

Module manager should use observers to access the modules and
check their availability.

### DIFF
--- a/pyanaconda/dbus/observer.py
+++ b/pyanaconda/dbus/observer.py
@@ -1,0 +1,243 @@
+#
+# Observers of remote DBus objects.
+#
+# Copyright (C) 2017  Red Hat, Inc.  All rights reserved.
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+from pyanaconda.dbus import DBus
+from pyanaconda.dbus.constants import DBUS_FLAG_NONE
+from pyanaconda.isignal import Signal
+
+from pyanaconda.anaconda_loggers import get_module_logger
+log = get_module_logger(__name__)
+
+__all__ = ["DBusObserverError", "DBusServiceObserver", "DBusObjectObserver"]
+
+
+class DBusObserverError(Exception):
+    """Exception class for the DBus observers."""
+    pass
+
+
+class DBusObserver(object):
+    """Base class for DBus observers.
+
+    This class is recommended to use only to watch the availability
+    of a service on DBus. It doesn't provide any support for accessing
+    objects provided by the service.
+
+    Usage:
+
+    def callback1(observer):
+        print("Service is available!")
+
+    def callback2(observer):
+        print("Service is unavailable!")
+
+    observer = DBusObserver("org.freedesktop.NetworkManager")
+    observer.service_available.connect(callback1)
+    observer.service_unavailable.connect(callback2)
+    observer.watch()
+    """
+
+    def __init__(self, service_name):
+        """Creates an DBus service observer.
+
+        :param service_name: a DBus name of a service
+        """
+        self._service_name = service_name
+        self._is_service_available = False
+
+        self._service_available = Signal()
+        self._service_unavailable = Signal()
+
+        self._watched_id = None
+
+    @property
+    def service_name(self):
+        """Returns a DBus name."""
+        return self._service_name
+
+    @property
+    def is_service_available(self):
+        """The proxy can be accessed."""
+        return self._is_service_available
+
+    @property
+    def service_available(self):
+        """Signal that emits when the service is available.
+
+        Signal emits this class as an argument. You have to
+        call the watch method to activate the signals.
+        """
+        return self._service_available
+
+    @property
+    def service_unavailable(self):
+        """Signal that emits when the service is unavailable.
+
+        Signal emits this class as an argument. You have to
+        call the watch method to activate the signals.
+        """
+        return self._service_unavailable
+
+    def watch(self):
+        """Watch the service name on DBus."""
+        bus = DBus.get_connection()
+        num = bus.watch_name(self.service_name,
+                             DBUS_FLAG_NONE,
+                             self._service_name_appeared_callback,
+                             self._service_name_vanished_callback)
+
+        self._watched_id = num
+
+    def unwatch(self):
+        """Stop to watch the service name on DBus."""
+        bus = DBus.get_connection()
+        bus.unwatch_name(self._watched_id)
+        self._watched_id = None
+
+    def _service_name_appeared_callback(self, *args):
+        """Callback for the watch method."""
+        self._is_service_available = True
+        self._service_available.emit(self)
+
+    def _service_name_vanished_callback(self, *args):
+        """Callback for the watch method."""
+        self._is_service_available = False
+        self._service_unavailable.emit(self)
+
+    def __str__(self):
+        """Returns a string version of this object."""
+        return self._service_name
+
+    def __repr__(self):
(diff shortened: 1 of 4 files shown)


pyanaconda/modules/boss/module_manager.py
```diff
@@ -19,111 +19,106 @@
...
-                log.exception("Quit failed for module: %s", service)
+        """Tell all running modules to quit."""
+        log.debug("Stop modules.")
+        for observer in self.module_observers:
+            if not observer.is_service_available:
+                continue
+
+            # Call asynchronously to avoid problems.
+            observer.proxy.Quit(callback=self._stop_modules_callback,
+                                callback_args=(observer,))
```
jkonecny12: What about catching exception from the observer here?

pyanaconda/modules/boss/module_manager.py
```diff
@@ -19,111 +19,106 @@
...
 
-    @property
-    def expected_module_services(self):
-        expected_modules = []
-        expected_modules.extend(ANACONDA_SERVICES)
-        expected_modules.extend(self.addon_module_services)
-        return expected_modules
+        for name in names:
+            if name.startswith(DBUS_ADDON_NAMESPACE):
+                self.add_module(name, auto_object_path(name))
```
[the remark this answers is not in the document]
poncovka: Well, the problem with addons is that we don't know the modules, so we have to guess the addon's object path.

pyanaconda/modules/boss/module_manager.py
poncovka: That's why we check if the service is available before we access the proxy.

pyanaconda/dbus/observer.py
```diff
@@ -0,0 +1,229 @@
...
+        self._proxy = None
+        super()._service_name_appeared_callback(*args)
+
+    def _service_name_vanished_callback(self, *args):
+        """Callback for the watch method."""
+        self._proxy = None
+        super()._service_name_vanished_callback(*args)
+
+    def __str__(self):
+        return self._object_path
```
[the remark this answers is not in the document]
poncovka: The logs are awful when you try to do that. Usually you don't have to know the service name, because the object path is unique. You can always print `observer.service_name` when you need to do that.

pyanaconda/dbus/observer.py
```diff
@@ -0,0 +1,229 @@
...
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+from pyanaconda.dbus import DBus
+from pyanaconda.dbus.constants import DBUS_FLAG_NONE
+from pyanaconda.isignal import Signal
+
+from pyanaconda.anaconda_loggers import get_module_logger
+log = get_module_logger(__name__)
+
```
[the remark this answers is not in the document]
poncovka: Fixed.

pyanaconda/dbus/observer.py
```diff
@@ -0,0 +1,229 @@
...
+        if object_path in self._proxies:
+            return self._proxies.get(object_path)
+
+        proxy = DBus.get_proxy(self._service_name, object_path)
+        self._proxies[object_path] = proxy
+        return proxy
+
+    def _service_name_appeared_callback(self, *args):
+        """Callback far the watch method."""
```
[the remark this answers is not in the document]
poncovka: Fixed.

pyanaconda/dbus/observer.py
```diff
@@ -0,0 +1,229 @@
...
+                                    self._service_name)
+
+        if not self._proxy:
+            self._proxy = DBus.get_proxy(self._service_name,
+                                         self._object_path)
+
+        return self._proxy
+
+    def _service_name_appeared_callback(self, *args):
+        """Callback far the watch method."""
```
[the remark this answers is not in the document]
poncovka: Fixed.

pyanaconda/dbus/observer.py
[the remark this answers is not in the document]
rvykydal: Maybe we could return DBusObjectObserver(service name, object path) for __repr__.

pyanaconda/modules/boss/module_manager.py
jkonecny12: I'm not sure if the checking is a good solution. It is basically race condition. Before we test availability and do that call could happen anything, and thanks to how multi threading works in python it could be a "long time" pause between these two instructions...

poncovka:
It looks like there is a misunderstanding about what this pull request is supposed to do. So there is a summary:

1. Caching of proxies: It is a nontrivial operation to create a new proxy. Therefore observers create the proxies on demand and keep them in a cache, so they can be used repeatedly.

2. Support for unavailable services: In some cases, we can decide not to run all modules. For example, the storage module will not be started in the initial setup. Or we could want to test and debug only a subset of modules. However, these are cases where the services are unavailable on purpose and they will not change their availability. Observers provide the property `is_service_available`. For the described use cases, it is a safe way how to test the service availability before we decide to access its objects.

3. Support for reload of services (for debugging and development only): We want Anaconda to be able to survive a reload of a service. It can be useful for a development and debugging when we want to test a new code without restarting the whole Anaconda. When a service is reloaded, the proxies of its objects need to be created again, because API of objects could have changed. In observers, the cache of proxies is cleared when a service name appears or disappears. Since proxies are created on demand, that is all we need to do. Keep in mind, that the reload of a service is not (and it was never supposed to be) a safe operation. It can cause errors in other modules, but the modules should be able to survive it.

What this pull request is not supposed to do:

1. Check the availability of a service before every DBus call: If a service is supposed to be available and it is not, then it is a bug. It will raise an exception, that will be handled in a loop and the module will survive it. Since it is a bug, it is an exception, that normally shouldn't happen. (Again, the reload of a service is not a safe operation. It is under our control, so we know that we can ignore the errors it will cause.)

2. Change the way how we handle DBus errors: This is Python. It is ok to raise exceptions.

pyanaconda/dbus/observer.py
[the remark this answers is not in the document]
poncovka: Fixed with `__repr__`.

pyanaconda/modules/boss/module_manager.py
M4rtinK: Well, I think as this is DBUS this is always a race condition - while the chance is small, it's possible the DBUS service/entity will no longer be there once you try to access it even right after a successful `is_service_available()` call.

Even though I guess it could be to tedious to handle this in individual modules, it might make sense to watch for this in Boss and other core places.

M4rtinK:
I don't think we want to add any result wrapping to stuff returned by DBUS function calls - if a service is expected to not be available in some cases the caller needs to catch the possible exception.

If the service _should_ be available but _isn't_ then that's a bug and stuff should crash to prevent undefined behavior during the installation.

Also, catching exception is pythonic, _result[1] == True_ is not.

pyanaconda/modules/boss/module_manager.py
poncovka: If a service is supposed to be available and it is not, then it is a bug - see the comment below.

jkonecny12:
> If the service should be available but isn't then that's a bug and stuff should crash to prevent undefined behavior during the installation.

Problem is that won't happen. The exception will be silently soaked by GLib loop so everything will continue as usual.

pyanaconda/dbus/observer.py
```diff
@@ -0,0 +1,243 @@
...
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+from pyanaconda.dbus import DBus
+from pyanaconda.dbus.constants import DBUS_FLAG_NONE
+from pyanaconda.isignal import Signal
+
+from pyanaconda.anaconda_loggers import get_module_logger
+log = get_module_logger(__name__)
+
+__all__ = ["DBusObserverError", "DBusObserver", "DBusServiceObserver", "DBusObjectObserver"]
```
jkonecny12: Is there a reason to have `DBusObserver` here? It is a base class so it shouldn't be used anywhere else so or os.

pyanaconda/modules/boss/module_manager.py
jkonecny12: OK, then it should be fine.

pyanaconda/dbus/observer.py
poncovka: Fixed.

poncovka:
@jkonecny12, I agree.

PR #1266, opened 2017-12-13
Add a base class for DBus interfaces

poncovka:
The InterfaceTemplate is a recommended base class for defining
a DBus interface, because it enables to separate the implementation
from the definition.

TaskInterface already uses the proxy pattern, so we just need to
modify it a little to use the InterfaceTemplate. Also, the Task class
doesn't need the module path.

InstallationInterface is already using the proxy pattern.

Fixing a connection to a signal in the install manager.
We should connect a callback to a signal.

### DIFF
--- a/pyanaconda/dbus/template.py
+++ b/pyanaconda/dbus/template.py
@@ -1,0 +1,101 @@
+#
+# template.py: Templates for DBus objects.
+#
+# Copyright (C) 2017  Red Hat, Inc.  All rights reserved.
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+from abc import ABC
+
+from pyanaconda.dbus import DBus
+
+__all__ = ["InterfaceTemplate"]
+
+
+class InterfaceTemplate(ABC):
+    """Template for DBus interface.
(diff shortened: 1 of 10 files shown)


jkonecny12:
jenkins, test this please

jkonecny12:
jenkins, test this please

pyanaconda/dbus/template.py
```diff
@@ -0,0 +1,99 @@
...
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+from pyanaconda.dbus import DBus
+
+__all__ = ["InterfaceTemplate"]
+
+
+class InterfaceTemplate(object):
```
jkonecny12: What about using ABC here and made the `connect_signals` method an abstract method.

pyanaconda/dbus/template.py
```diff
@@ -0,0 +1,99 @@
...
+        """Create a publishable DBus object.
+
+        :param implementation: an implementation of this interface
+        """
+        self._implementation = implementation
+        self._object_path = None
+        self.connect_signals()
+
+    @property
+    def implementation(self):
```
jkonecny12: I would suggest to change this to `impl`. Personally I don't like abbreviations but in this case it will be called a lot so it would be nice to have the name smaller.

poncovka: Unfortunately, an interface class full of calls like `self.impl.do_something()` can be very confusing since `impl` will be defined in the parent class. I will leave it as it is for now.

pyanaconda/dbus/template.py
poncovka: Used ABC.

pyanaconda/dbus/template.py
```diff
@@ -0,0 +1,99 @@
...
+        """Return the implementation of this interface.
+
+        :return: an implementation
+        """
+        return self._implementation
+
+    @property
+    def object_path(self):
```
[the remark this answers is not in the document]
poncovka: Specified in the comment.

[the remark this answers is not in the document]
jkonecny12: I guess it is enough.

PR #1269, opened 2017-12-14
Use observers in the install manager

poncovka:
Install manager should use observers to access the modules.

### DIFF
--- a/pyanaconda/modules/boss/install_manager/install_manager.py
+++ b/pyanaconda/modules/boss/install_manager/install_manager.py
@@ -17,8 +17,6 @@
 # Red Hat, Inc.
 #
 
-from pydbus.auto_names import auto_object_path
-
 from pyanaconda.isignal import Signal
 from pyanaconda.dbus import DBus
 from pyanaconda.modules.boss.install_manager.installation_interface import InstallationNotRunning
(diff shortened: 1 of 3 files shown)


pyanaconda/modules/boss/install_manager/install_manager.py
```diff
@@ -115,17 +113,17 @@ def start_installation(self):
...
 
-        if not self._modules:
+        if not self._module_observers:
             log.error("Starting installation without available modules.")
 
-        for module_service in self._modules:
-            # FIXME: This is just a temporary solution.
-            module_object = DBus.get_proxy(module_service, auto_object_path(module_service))
+        for observer in self._module_observers:
+            if not observer.is_service_available:
```
M4rtinK: Maybe we could catch the DBusObserverError instead ? Also I think it could be useful to log if a module observer we try to access is not actually usable.

rvykydal: I agree with the logging.
As for catching the exception, I think the approach of - check availability and raise exception (without catching/handling) when calling observer.proxy methods - seems OK to me here.

poncovka: The check is here so we can test the StartInstallation method even with some failed modules. It shouldn't be here once we will start to use it for real. I will add a log there and a comment.

PR #1274, opened 2017-12-15
Master kickstart dispatcher anaconda integration

rvykydal:
Integrate kickstart distpatcher in Anaconda:

- Create Boss interface for early stage of anaconda where there is no event loop so we will call boss methods synchronously or poll the state (modules available).
- Boss (KickstartManager) asks modules for lists of handled elements and gives them partial kickstart (string).
- Mapping between module kickstart and original kickstart lines is handled by Boss (KickstartManager).
- Kickstart parsing in modules is not implemented, there is a dummy implementation returns no-error return value for now and it is possible to trigger parsing error by having PARSE_ERROR string in kickstart (eg "network --hostname=PARSE_ERROR")
- I'll ask Chris if we can modify KickstartParseError to contain lineno attribute. If not, we can be getting it from modified KickstartParser (KickstartParser for modules).

### DIFF
--- a/pyanaconda/modules/boss/boss.py
+++ b/pyanaconda/modules/boss/boss.py
@@ -25,25 +25,70 @@ from pyanaconda.dbus import DBus
 gi.require_version("GLib", "2.0")
 from gi.repository import GLib
 
+from pyanaconda.dbus.typing import *  # pylint: disable=wildcard-import
 from pyanaconda.dbus.interface import dbus_interface
 from pyanaconda.modules.base import BaseModule
-from pyanaconda.dbus.constants import DBUS_BOSS_NAME, DBUS_BOSS_PATH, DBUS_BOSS_INSTALLATION_PATH
+from pyanaconda.dbus.constants import DBUS_BOSS_NAME, DBUS_BOSS_PATH, DBUS_BOSS_INSTALLATION_PATH,\
+                                      DBUS_BOSS_ANACONDA_NAME
 
 from pyanaconda.modules.boss.module_manager import ModuleManager
 from pyanaconda.modules.boss.install_manager.installation_interface import InstallationInterface
 from pyanaconda.modules.boss.install_manager.install_manager import InstallManager
+from pyanaconda.modules.boss.kickstart_manager import KickstartManager
 
 from pyanaconda import anaconda_logging
 log = anaconda_logging.get_dbus_module_logger(__name__)
 
 
-@dbus_interface(DBUS_BOSS_NAME)
-class Boss(BaseModule):
+@dbus_interface(DBUS_BOSS_ANACONDA_NAME)
+class AnacondaBossInterface(object):
+    """Temporary interface for anaconda.
 
-    def __init__(self, module_manager=None, install_manager=None):
+    Used for synchronization with anaconda during transition.
+    """
+
+    def SplitKickstart(self, path: Str) -> Str:
+        """Splits the kickstart for modules.
+
+        :returns: kickstart regenerated from elements after splitting
+
+        :raises SplitKickstartError: if parsing fails
+        """
+        log.info("splitting kickstart %s", path)
+        self._kickstart_manager.split(path)     # pylint: disable=no-member
+        return self._kickstart_manager.unprocessed_kickstart    # pylint: disable=no-member
+
+    def DistributeKickstart(self) -> List[Tuple[Str,Tuple[Int, Str],Str]]:
+        """Distributes kickstart to modules synchronously.
+
+        Assumes all modules are started.
+
+        :returns: list of (Module service, (Line number, File name), Error message)
+                  tuples for each kickstart parsing error.
+        """
+        log.info("distributing kickstart")
+        errors = self._kickstart_manager.distribute()   # pylint: disable=no-member
+        if errors:
+            log.info("distributing kickstart errors: %s", errors)
+        return errors
+
+    def UnprocessedKickstart(self) -> Str:
+        """Returns kickstart containing parts that are not handled by any module."""
+        return self._kickstart_manager.unprocessed_kickstart    # pylint: disable=no-member
+
+    def AllModulesAvailable(self) -> Bool:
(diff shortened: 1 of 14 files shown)


pyanaconda/modules/boss/kickstart_manager.py
```diff
@@ -0,0 +1,139 @@
...
+__all__ = ['KickstartManager', 'SplitKickstartError']
+
+
+@map_error("{}.SplitKickstartError".format(DBUS_BOSS_ANACONDA_NAME))
+class SplitKickstartError(Exception):
+    """Error when parsing kickstart for splitting."""
+    pass
+
+
+class SplitKickstartUnknownSectionError(SplitKickstartError):
```
poncovka: I am not sure, but maybe all DBus errors should be decorated.

pyanaconda/modules/boss/kickstart_manager.py
```diff
@@ -0,0 +1,139 @@
...
+
+    def split(self, path):
+        """Split the kickstart given by path into elements."""
+        self._elements = None
+        self._kickstart_path = path
+        handler = makeVersion()
+        ksparser = SplitKickstartParser(handler, valid_sections=VALID_SECTIONS_ANACONDA)
+        try:
+            result = ksparser.split(path)
+        except KickstartParseError as e:
```
poncovka: Actually, you can map the` KickstartParseError` and `KickstartError` to a DBus error if you want to.

pyanaconda/modules/boss/kickstart_manager.py
```diff
@@ -0,0 +1,139 @@
...
+                observer.service_name, commands, sections, addons))
+
+            elements = self._elements.get_and_process_elements(commands=commands,
+                                                               sections=sections,
+                                                               addons=addons)
+            kickstart = self._elements.get_kickstart_from_elements(elements)
+            log.info("distribute kickstart: {} will get kickstart elements: {}".format(
+                observer.service_name, elements))
+
+            error_lineno, error_msg = observer.proxy.ConfigureWithKickstart(kickstart)
```
poncovka: We might need to get the warnings are well, because in the strict mode we treat them as errors. But that can be extended in another issue.

rvykydal:
Updated with pylint fixes.

rvykydal:
jenkins, test this please

pyanaconda/startup_utils.py
```diff
@@ -433,6 +462,8 @@ def parse_kickstart(options, addon_paths):
         log.info("Parsing kickstart: " + ks)
         ksdata = kickstart.parseKickstart(ks, options.ksstrict)
 
+        if "run_boss" in flags.cmdline:
+            distribute_kickstart_with_boss(ks)
```
jkonecny12: I'm not sure if there is better place but this is side effect and it will be hard to find how the kickstart is distributed when it's here. I would rather see this call somewhere else.

tests/pyanaconda_tests/kickstart_manager.py
```diff
@@ -0,0 +1,298 @@
...
+# source code or documentation are not subject to the GNU General Public
+# License and may only be used or replicated with the express permission of
+# Red Hat, Inc.
+#
+
+import unittest
+from contextlib import contextmanager
+import os
+
+from pyanaconda.dbus.observer import DBusObjectObserver
```
jkonecny12: Group `import`s and `from`s together.

pyanaconda/modules/boss/boss.py
```diff
@@ -25,25 +25,69 @@
...
 from pyanaconda.modules.boss.install_manager.install_manager import InstallManager
+from pyanaconda.modules.boss.kickstart_manager import KickstartManager
 
 from pyanaconda import anaconda_logging
 log = anaconda_logging.get_dbus_module_logger(__name__)
 
 
+@dbus_interface(DBUS_BOSS_ANACONDA_NAME)
+class AnacondaBossInterface(object):
+    """Temporary interface for anaconda.
```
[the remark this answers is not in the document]
rvykydal: I mean temporary interface for the transition stage of development where we are moving functionality from anaconda to boss and modules.

pyanaconda/modules/boss/boss.py
```diff
@@ -25,25 +25,69 @@
...
+
+    def DistributeKickstart(self) -> List[Tuple[Str,Tuple[Int, Str],Str]]:
+        """Distributes kickstart to modules synchronously
+
+        Assumes all modules are started.
+
+        :returns: list of (Module service, (Line number, File name), Error message)
+                  tuples for each kickstart parsing error.
+        """
+        log.info("distirbuting kickstart")
```
M4rtinK: typo, should be "distributing kickstart"

pyanaconda/modules/boss/boss.py
```diff
@@ -25,25 +25,69 @@
...
+        errors = self._kickstart_manager.distribute()   # pylint: disable=no-member
+        if errors:
+            log.info("distributng kickstart errors: %s", errors)
```
M4rtinK: also typo, should be: "distributing kickstart errors: %s"

pyanaconda/startup_utils.py
rvykydal: I think we need either to split the parse_kickstart function or do passing to boss in this function. I'd clean this up (parse_kickstart) when we know better what we want to do (for example parse only modules-unprocessed part in anaconda?). For now, we just want to be able to pass kickstart to modules to be able to work on their POCs/integration.
What about - to be more explicit - adding pass_to_boss flag to the parse_kickstart function? I'll add a patch doing that to the PR.

rvykydal:
- Fixed typos.
- I am leaving the new Exceptions for now (@poncovka), but I am mapping specific child errors instead of the parent to dbus errors now. Also renamed one of the exceptions to tell what happened more correctly (can be seen in tests).
- I've added one more patch for @jkonecny12 (parse_kickstart in anaconda)

PR #1275, opened 2017-12-18
Differentiate upstream and build-time version (#1493952)

M4rtinK:
There are basically two version Anaconda can have:

- the upstream release version (eq. the version of the tarball)

example: 28.13

- the package version, which is generally a suffix to the upstream
  version and an upstream version can have multiple such suffixes
  due to rebuilds or patch only builds

example: 28.13-1, 28-13.2, 28.13-3, ...

So let's differentiate these two and show them where appropriate:

- show the upstream version for anaconda --version
- show the package version in the installation environment

This should match more closely to user expectations:
- Fedora QA & co. want to know which specific Anaconda patch release
  in used on the installation image during a freeze
- users of Anaconda from the command line mostly just want to
  know the general Anaconda version, many other CLI tools also
  show just the "upstream version" as well

Also handle the case of version.py being incorrectly generated, to
prevent an Anaconda crash in such case.

Resolves: rhbz#1493952

### DIFF
--- a/anaconda.py
+++ b/anaconda.py
@@ -324,7 +324,7 @@ if __name__ == "__main__":
             startup_utils.prompt_for_ssh()
             sys.exit(0)
 
-    log.info("%s %s", sys.argv[0], startup_utils.get_anaconda_version_string())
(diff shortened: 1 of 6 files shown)


jkonecny12:
jenkins, test this please

PR #1278, opened 2017-12-19
Tweak setup environment script

jkonecny12:
* Remove `/anaconda` in mock before copying new one
* Add `run-tests` parameter to setup-test-env script
* Add the `_prepare_command` helper function to setup-test-env

### DIFF
--- a/scripts/testing/setup-mock-test-env.py
+++ b/scripts/testing/setup-mock-test-env.py
@@ -31,6 +31,13 @@ def _prepare_command(mock_command):
     return cmd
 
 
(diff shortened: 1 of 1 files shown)


jkonecny12:
**UPDATED:**
* Add new `--result [folder]` param to copy out result folder.
* Multiple commands can be chained in one call now. You can for example do ``./setup-mock-env.py -i vim --copy --run-tests --result /tmp/results``.

jkonecny12:
**UPDATED:**
* Check if result dir exists before running any command.
* Do not fail when tests are failing. Thanks to this result was not saved when tests failed.

PR #1282, opened 2017-12-20
Improve password checking status and error messages

M4rtinK:
Most importantly have two variants, one for passwords and one for
passphrases. Related to that use "secret" for stuff that can be
password or passphrase (and possibly also something else in the future).

Instead of "password name" a secret-type enum is now used to
differentiate if the secret being checked should be called a
password or a passphrase.

And when we are at it make "Done" bold in the warning messages,
to hopefully improve their readability.

### DIFF
--- a/pyanaconda/constants.py
+++ b/pyanaconda/constants.py
@@ -137,39 +137,57 @@ FIRSTBOOT_ENVIRON = "firstboot"
 # Tainted hardware
 UNSUPPORTED_HW = 1 << 28
 
+# Password type
+class SecretType(Enum):
+    PASSWORD = "password"
+    PASSPHRASE = "passphrase"
+
 # Password validation
-PASSWORD_MIN_LEN = 6
-PASSWORD_EMPTY_ERROR = N_("The %(password_name)s is empty.")  # singular
-PASSWORD_CONFIRM_ERROR_GUI = N_("The %(password_name_plural)s do not match.")  # plural
-PASSWORD_CONFIRM_ERROR_TUI = N_("The %(password_name_plural)s you entered were different. Please try again.")  # plural
-# The password-too-short constant is used to replace a libpwquality error message,
+SECRET_MIN_LEN = 6
+SECRET_EMPTY_ERROR = {
+    SecretType.PASSWORD : N_("The password is empty."),
+    SecretType.PASSPHRASE : N_("The passphrase is empty.")
+}
+SECRET_CONFIRM_ERROR_GUI = {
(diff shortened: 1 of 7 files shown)


pyanaconda/constants.py
```diff
@@ -137,40 +137,58 @@
...
-PASSWORD_FINAL_CONFIRM = N_("Press Done again to use anyway.")
-PASSWORD_ASCII = N_("The %(password_name)s you have provided contains non-ASCII characters. You may not be able to switch between keyboard layouts when typing it.")
-# ^ singular
-PASSWORD_DONE_TWICE = N_("You will have to press Done twice to confirm it.")
-PASSWORD_DONE_TO_CONTINUE = N_("Press Done to continue.")
+SECRET_TOO_SHORT = {
+    SecretType.PASSWORD : N_("The password is too short"),
+    SecretType.PASSPHRASE : N_("The passphrase is too short")
+}
+SECRET_WEEK = {
```
rvykydal: SECRET_WEAK

pyanaconda/constants.py
```diff
@@ -137,40 +137,58 @@
...
+    SecretType.PASSWORD : N_("The password you have provided is weak."),
+    SecretType.PASSPHRASE : N_("The passphrase you have provided is weak.")
+}
+SECRET_WEEK_WITH_ERROR = {
```
rvykydal: SECRET_WEAK_WITH_ERROR

pyanaconda/ui/gui/spokes/root_password.py
```diff
@@ -214,6 +213,7 @@ def _checks_done(self, error_message):
             else:
                 # add suffix for the click twice logic
                 self.show_warning_message("{} {}".format(error_message, _(constants.PASSWORD_DONE_TWICE)))
+                stuff = "{} {}".format(error_message, _(constants.PASSWORD_DONE_TWICE))
```
rvykydal: doesn't seem like very useful stuff ;)

rvykydal:
Well, our CI pylint tests already found my 2 comments...

pyanaconda/ui/gui/spokes/root_password.py
M4rtinK: Indeed, a debugging leftover. :) Fixed now.

pyanaconda/constants.py
M4rtinK: Oops! :D Fixed now.

pyanaconda/constants.py
M4rtinK: Fixed as well.

PR #1283, opened 2017-12-20
Remove spurious echo call from tmux service file (#1526861)

M4rtinK:
Looks like systemd 236 does not like the escape sequence in the
call, which most likely tries to ring the terminal bell (!).

We have no evidence of this ever working & dropping the line makes
Tmux work again, so just drop the line - at least for now.


M4rtinK:
Merging as a hotfix.

PR #1286, opened 2018-01-03
Check payload is set before accessing its data (#1524785)

M4rtinK:
The payload is None when the Network spoke is running in Initial Setup,
so don't access its methods/properties without checking the payload is actually
set to something.

Resolves: rhbz#1524785


M4rtinK:
> I guess this should be backported to Rawhide.

Yep: https://github.com/rhinstaller/anaconda/pull/1287

PR #1289, opened 2018-01-03
Move system-logos to anaconda-core (#1529239)

bcl:
Some of the logo files (in this case syslinux-splash.png) are needed by
non-gui tools like syslinux and grub2. Install them as part of core so
that they are available for anaconda-tui and anaconda-gui.

Related: rhbz#1529239

### DIFF
--- a/anaconda.spec.in
+++ b/anaconda.spec.in
@@ -110,6 +110,7 @@ Requires: python3-dbus
 Requires: python3-pwquality
 Requires: python3-systemd
 Requires: python3-pydbus
+Requires: system-logos
 
 # pwquality only "recommends" the dictionaries it needs to do anything useful,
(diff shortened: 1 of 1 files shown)


jkonecny12:
Thanks for your patch but shouldn't this be required by the tools which are using them? Feels strange to me to add something to Anaconda which is required by syslinux and grub2 but we don't use it.

bcl:
> Thanks for your patch but shouldn't this be required by the tools which are using the splash image? Feels strange to me to add something to Anaconda which is required by syslinux and grub2 but we don't use it.

Installing only anaconda -tui will hit this problem as well. Since it's already required by (one part) of anaconda I think it makes sense to move it down into -core.

jkonecny12:
Jenkins, it's ok to test.

PR #1291, opened 2018-01-04
Add KickstartMerger for merging snippets in order defined by handler

rvykydal:
The add_command method is intended to be used for adding command snippets returned by modules. A module would return list of (Command Name, Kickstart snippet) tuples for each command it handles.

The add_command_line method is more a utility method that can be used during transition, eg if/when we want to merge also commands not handled by any module (firstboot? shutdown?). It is internally parsing the command name from the line. Maybe it will turn out to be redundant, ie anaconda/boss, or the thing handling commands not (yet) handled by any module will use add_command as well (like add_command("firstboot", str(ksdata.firstboot))).

### DIFF
--- a/pyanaconda/kickstart_dispatcher/parser.py
+++ b/pyanaconda/kickstart_dispatcher/parser.py
@@ -16,7 +16,8 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #
 
-__all__ = ["SplitKickstartParser", "VALID_SECTIONS_ANACONDA"]
+__all__ = ["SplitKickstartParser", "VALID_SECTIONS_ANACONDA", "KickstartMerger",
+           "MergeKickstartUnknownCommandError", "MergeKickstartCommandAlreadyAddedError"]
 
 from pykickstart.parser import KickstartParser
 from pykickstart.sections import Section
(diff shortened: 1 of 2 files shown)


rvykydal:
> The add_command method is intended to be used for adding command snippets returned by modules. A module would return list of (Command Name, Kickstart snippet) tuples for each command it handles.

There is one thing to keep in mind when referring to commands handled by a module by name (eg "network") as some commands map to multiple names:
F26_DisplayMode -> graphical, text, ...
F23_Reboot -> halt, poweroff, reboot, shutdown
Luckily these seem not to be handled by a module but rather by boss/anaconda
For these cases the add_command_line can become handy.

pyanaconda/kickstart_dispatcher/parser.py
```diff
@@ -178,3 +179,140 @@ def setupSections(self):
...
+    def _get_cmd_name_priorities(self, handler):
+        # Transform the mapping of command objects priorities from handler
+        # to a mapping of command names priorities.
+        command_names = {}
+        for name, cmd in handler.commands.items():
+            cmd_class = cmd.__class__.__name__
+            if cmd_class in command_names:
+                command_names[cmd_class].append(name)
+            else:
+                command_names[cmd_class] = [name]
```
jkonecny12: Just an idea. This method is basically two separate algorithms so I would suggest to move this part to a separate method name something like `_collect_command_names()`. I think it could be better for reading and easier to understand.

rvykydal:
Closing as we probably won't need it.

PR #1296, opened 2018-01-09
Remove btrfs support from UI.

rvykydal:
UI should stop offer btrfs when blivet/libblockdev (based on system) stops
supporting it (blivet.devicefactory.is_supported_device_type).

For handling existing btrfs devices just make sure we don't crash, don't offer
modifications (just removing) and present the info about the device type not
being supported in a decent way.

### DIFF
--- a/pyanaconda/ui/gui/spokes/custom_storage.py
+++ b/pyanaconda/ui/gui/spokes/custom_storage.py
@@ -68,7 +68,8 @@ from blivet.platform import platform
 
 from pyanaconda.storage_utils import ui_storage_logger, device_type_from_autopart, storage_checker, \
     verify_luks_devices_have_key, get_supported_filesystems
-from pyanaconda.storage_utils import DEVICE_TEXT_PARTITION, DEVICE_TEXT_MAP, DEVICE_TEXT_MD
+from pyanaconda.storage_utils import DEVICE_TEXT_PARTITION, DEVICE_TEXT_MAP, DEVICE_TEXT_MD, \
+    DEVICE_TEXT_UNSUPPORTED
 from pyanaconda.storage_utils import PARTITION_ONLY_FORMAT_TYPES, MOUNTPOINT_DESCRIPTIONS
 from pyanaconda.storage_utils import NAMED_DEVICE_TYPES, CONTAINER_DEVICE_TYPES
 from pyanaconda.storage_utils import try_populate_devicetree
@@ -115,6 +116,8 @@ DEVICE_CONFIGURATION_ERROR_MSG = N_("Device reconfiguration failed. <a href=\"\"
                                     "details.</a>")
 UNRECOVERABLE_ERROR_MSG = N_("Storage configuration reset due to unrecoverable "
                              "error. <a href=\"\">Click for details.</a>")
+
+DEVICE_TYPE_CONST_UNSUPPORTED = "DEVICE_TYPE_UNSUPPORTED"
 
 def dev_type_from_const(dev_type_const):
     """ Return integer corresponding to name for device type defined as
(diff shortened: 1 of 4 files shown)


pyanaconda/ui/gui/spokes/custom_storage.py
```diff
@@ -1458,15 +1471,17 @@ def _setup_device_type_combo(self, device, device_name):
 
             self._device_name_dict[_type] = name
 
-        itr = self._typeStore.get_iter_first()
-        while itr:
-            if dev_type_from_const(self._typeStore[itr][1]) == device_type:
+        if not self._set_device_type(device_type):
+            unsupported = device_type in (dev_type_from_const(dc) for dc
+                                          in should_appear - should_appear_supported)
```
jkonecny12: For better readability I would like to see that unsupported list creation in a function or property. It could be even generator.

Or you can create function `is_unsupported_device_type` which is complement of `is_supported_device_type` function.

rvykydal: Hm, it doesn't make sense to me. Did you see how should_appear and should_appear_supported is created above? I really can't imagine how such a function or property would look like and how it would be useful. Any suggestion? I don't think it is less readable then the rest of the code. Making this look better would require bigger rewrite of the function and related code and I didn't want to do it in this PR.

`Or you can create function is_unsupported_device_type which is complement of is_supported_device_type function.`
Why? I could use not is_supported_device_type (moreover this is a function from blivet). Still I don't see how should I use it to make things better.

pyanaconda/ui/tui/spokes/storage.py
```diff
@@ -587,7 +588,8 @@ def __init__(self, data, storage, payload, instclass):
         self._container = None
         self.part_schemes = OrderedDict()
         pre_select = self.data.autopart.type or DEFAULT_AUTOPART_TYPE
-        for item in AUTOPART_CHOICES:
+        autopart_choices = (c for c in AUTOPART_CHOICES if is_supported_device_type(AUTOPART_DEVICE_TYPES[c[1]]))
```
poncovka: This could be probably moved to a function.

pyanaconda/ui/gui/spokes/custom_storage.py
jkonecny12: Personally I would abstract all the `should_appear_supported`, `should_appear` and so on, because it is just a noise in this method and in general make this method over-complicated.

But I understand that this is not purpose of the PR and I'm not insisting on this change.

jkonecny12:
You can rebase this PR and the CI tests should work.

pyanaconda/ui/gui/spokes/custom_storage.py
```diff
@@ -1421,23 +1433,24 @@ def _setup_device_type_combo(self, device, device_name):
...
         # store.remove() updates or invalidates the passed iterator
         itr = self._typeStore.get_iter_first()
         valid = True
         while itr and valid:
             dev_type_const = self._typeStore[itr][1]
-            if dev_type_const not in should_appear:
+            if dev_type_const not in should_appear_supported:
                 valid = self._typeStore.remove(itr)
-            elif dev_type_const in should_appear:
+            elif dev_type_const in should_appear_supported:
```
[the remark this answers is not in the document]
rvykydal: Yes, I just wanted to touch the existing code as little as possible but you are right - this is too much - I'll change it.

pyanaconda/ui/tui/spokes/storage.py
rvykydal: Agreed and moved.

rvykydal:
jenkins, test this please

Conan-Kudo:
😭

PR #1299, opened 2018-01-10
Add support for Variant in .buildstamp

poncovka:
Added support for the Variant item in the .buildstamp file, so we can
choose install classes based on the name of the variant.

Give the modules with install classes proper python names, otherwise
it is not possible to import them.


pyanaconda/installclasses/fedora_server.py
```diff
@@ -1,5 +1,5 @@
 #
-# fedora-server.py
+# fedora_server.py
```
jkonecny12: Personally I would like to get rid of these file names as comments on top of a file. First of all, why we need them in a first place?

poncovka: Fixed.

poncovka:
@M4rtinK, we don't have to wait for lorax. The variable `productVariant` is set to an empty string by default.

jkonecny12:
Yeah but I think it would be wise to wait. In case some decision will change or this won't be a way how Lorax can implement it.

bcl:
Looks ok to me. Here's the lorax PR - https://github.com/rhinstaller/lorax/pull/302

poncovka:
The lorax PR was merged: https://github.com/rhinstaller/lorax/commit/dc635a97f66aeadd5e353d141c51a20b0024d391

PR #1304, opened 2018-01-15
RHEL 8 alpha - test dependencies have returned!

jkonecny12:
Revert commits which disabled tests.


jkonecny12:
To get these test successful I need to change configuration on Jenkins but that would broke all other PRs. So I will do that before merging this.

jkonecny12:
jenkins, test this please

PR #1305, opened 2018-01-15
kickstart: support firewall --use-system-defaults (#1526450)

rvykydal:
Needed for [1] where we would like to include firewalld
and configure firewalld in Atomic Host (in the ostree) and
have Anaconda leave the delivered "defaults" in place. The
action here is to do nothing if the user specified
--use-system-defaults.

[1] https://pagure.io/atomic-wg/issue/401

Resolves: rhbz#1526450


rvykydal:
Updated rhbz link

PR #1306, opened 2018-01-16
 Add --init as new parameter to setup-mock-test-env

jkonecny12:
This will help us to run initialize with running tests at once.

So it is possible to run now:

``setup-mock-test-env --init --copy --run-tests --result ./result_dir``

### DIFF
--- a/scripts/testing/setup-mock-test-env.py
+++ b/scripts/testing/setup-mock-test-env.py
@@ -7,7 +7,7 @@ import os
 import sys
 import subprocess
 
-from argparse import ArgumentParser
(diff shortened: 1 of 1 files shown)


jkonecny12:
I need to change CI configuration to make this works.

jkonecny12:
**UPDATED:**
* Stays backward compatible. No need to change Jenkins CI configuration.

jkonecny12:
**UPDATED:**
* The backward compatibility won't be keep. It results to confusing behavior and code.

jkonecny12:
jenkins, test this please

PR #1307, opened 2018-01-16
Added tests for the timezone module and other.

poncovka:
Added tests for the timezone module, recognition of standard
interfaces, emitting of changed properties, all types of observers
and the advanced template.

### DIFF
--- a/pyanaconda/threading.py
+++ b/pyanaconda/threading.py
@@ -45,6 +45,9 @@ class ThreadManager(object):
         self._objs_lock = threading.RLock()
         self._errors = {}
         self._errors_lock = threading.RLock()
+        self._main_thread = threading.main_thread()
+
+    def set_current_thread_as_main(self):
         self._main_thread = threading.current_thread()
 
     def __call__(self):
@@ -303,8 +306,7 @@ def initThreading():
        an event is triggered in the GTK main loop. And IT HAS TO BE CALLED IN
        THE MAIN THREAD.
     """
-    global threadMgr
-    threadMgr = ThreadManager()
+    threadMgr.set_current_thread_as_main()
 
 
-threadMgr = None
+threadMgr = ThreadManager()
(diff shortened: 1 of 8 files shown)


jkonecny12:
jenkins, test this please

poncovka:
**Updated:** It should be always safe to import the thread manager from now.

pyanaconda/threading.py
```diff
@@ -303,8 +306,7 @@ def initThreading():
        an event is triggered in the GTK main loop. And IT HAS TO BE CALLED IN
        THE MAIN THREAD.
     """
-    global threadMgr
-    threadMgr = ThreadManager()
+    threadMgr.set_current_thread_as_main()
 
 
-threadMgr = None
+threadMgr = ThreadManager()
```
jkonecny12: Yeah, this is basically what I had in mind, except I want to get rid off `initThreading()` method and `threadMgr`.

pyanaconda/threading.py
```diff
@@ -45,6 +45,9 @@ def __init__(self):
         self._objs_lock = threading.RLock()
         self._errors = {}
         self._errors_lock = threading.RLock()
+        self._main_thread = threading.main_thread()
+
+    def set_current_thread_as_main(self):
```
jkonecny12: I don't see a point in this. Why we want to have something which will be replaced immediately after import. 

Right now basically you will do:
```
self._main_thread = threading.main_thread()
self._main_thread = threading.current_thread()
```

Then you can remove the `initThreading()` completely and we don't need to call it at all. On the other hand I would leave this method here (but docstring should be here). It could be valuable to have the method to set current thread as main somewhere.

poncovka: The initThreading() should be called only from anaconda.py now, it is not called from the modules. I wasn't sure if main_thread is same as current_thread in anaconda.py. But yes, initThreading could be removed.

PR #1308, opened 2018-01-18
Small changes in modules and observers.

poncovka:
* Rename SetUTC to SetIsUTC in the timezone module. 
* Connect to the observed service and other stuff.
* Add the Kickstarted property to the kickstart modules.
* Fix tests for the timezone module.

### DIFF
--- a/pyanaconda/dbus/observer.py
+++ b/pyanaconda/dbus/observer.py
@@ -40,16 +40,26 @@ class DBusObserver(object):
 
     Usage:
 
+    # Create the observer and connect to its signals.
+    observer = DBusObserver("org.freedesktop.NetworkManager")
+
     def callback1(observer):
         print("Service is available!")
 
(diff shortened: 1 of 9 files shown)


pyanaconda/dbus/observer.py
```diff
@@ -93,6 +93,22 @@ def service_unavailable(self):
...
+
+        if not bus_proxy.NameHasOwner(self.service_name):
+            raise DBusObserverError("Service {} is not available."
+                                    .format(self._service_name))
+
+        self.watch()
+        self._enable_service()
+
+    def disconnect(self):
+        """Disconnect the service without waiting."""
```
jkonecny12: I would change the `without waiting` -> `immediately`. The same in doc for `connect()`.

Also is here a disconnect method with waiting?

poncovka:
Updated.

PR #1309, opened 2018-01-18
Prevent 99-copy-lgs.ks from exiting with a 1

bcl:
Using the previous form would cause the script to exit with a 1 when
/run/install/ks.cfg doesn't exist. Change to using an elif so that it
will return a 0.


jkonecny12:
jenkins, test this please

PR #1311, opened 2018-01-23
Add Atomic install class

jkonecny12:
Migrate Atomic install class to Anaconda.
Add `productVariant` check and fix stylesheet in the atomic install class.

The Atomic iso (Fedora-Atomic-ostree-x86_64-Rawhide-20180117.n.1.iso) already have `productVariant` in the `.buildstamp` file and stylesheet in the fedora logos package so this install class should work without problems -- tested with updates image. The only thing missing is to deprecate the fedora-productimg-atomic package.

*Resolves: rhbz#1536853
Resolves: rhbz#1491287*

### DIFF
--- a/pyanaconda/installclasses/fedora_server.py
+++ b/pyanaconda/installclasses/fedora_server.py
@@ -37,6 +37,10 @@ class FedoraServerInstallClass(FedoraBaseInstallClass):
         hidden = True
 
     def setDefaultPartitioning(self, storage):
+        self.createDefaultPartitioning(storage)
+
+    @staticmethod
+    def createDefaultPartitioning(storage):
         autorequests = [PartSpec(mountpoint="/", fstype=storage.default_fstype,
                                  size=Size("2GiB"),
                                  max_size=Size("15GiB"),
--- a/pyanaconda/installclasses/installclass_atomic.py
+++ b/pyanaconda/installclasses/installclass_atomic.py
@@ -22,13 +22,9 @@
 import os
 import shutil
 from pyanaconda.installclasses.fedora import FedoraBaseInstallClass
+from pyanaconda.installclasses.fedora_server import FedoraServerInstallClass
 from pyanaconda.product import productVariant
 from pyanaconda.core import util
-from pyanaconda.kickstart import getAvailableDiskSpace
-from pyanaconda.storage.partspec import PartSpec
-from pyanaconda.storage.autopart import swap_suggestion
-from pyanaconda.platform import platform
-from blivet.size import Size
 
 import logging
 log = logging.getLogger("anaconda")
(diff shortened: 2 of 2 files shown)


pyanaconda/installclasses/installclass_atomic.py
```diff
@@ -0,0 +1,143 @@
...
+    sortPriority = 11000
+    hidden = False
+    defaultFS = "xfs"
+
+    def __init__(self):
+        self.localemap = {} # loaded lazily
+        FedoraBaseInstallClass.__init__(self)
+
+    # This is intended right now to match Fedora Server; if changing this,
+    # please discuss on https://lists.projectatomic.io/projectatomic-archives/atomic-devel/
```
cgwalters: Side note...now that this is in the same codebase as Server, maybe we could actually dedup this?  Not a blocker/requirement for merging though from my PoV.

jkonecny12: I've used default partitioning from Fedora Server in the Fedora Atomic install class. Is that what you meant?

jkonecny12:
**UDPATED:**
* use `__all__` in Atomic install class
* use default partitioning in Atomic from Server variant

AdamWill:
ping on this? Atomic installer images for Rawhide are still broken.

pyanaconda/installclasses/fedora_server.py
```diff
@@ -37,6 +37,10 @@ class FedoraServerInstallClass(FedoraBaseInstallClass):
         hidden = True
 
     def setDefaultPartitioning(self, storage):
+        FedoraServerInstallClass.createDefaultPartitioning(storage)
```
poncovka: I think that you can use `self` here.

pyanaconda/installclasses/installclass_atomic.py
```diff
@@ -0,0 +1,120 @@
...
+
+    if productVariant != "Atomic":
+        hidden = True
+
+    def __init__(self):
+        self.localemap = {}  # loaded lazily
+        FedoraBaseInstallClass.__init__(self)
+
+    def setDefaultPartitioning(self, storage):
+        FedoraServerInstallClass.createDefaultPartitioning(storage)
```
poncovka: Could this install class be based on Fedora Server install class?

pyanaconda/installclasses/installclass_atomic.py
```diff
@@ -0,0 +1,120 @@
...
+import logging
+log = logging.getLogger("anaconda")
+
+__all__ = ['AtomicInstallClass']
+
+
+class AtomicInstallClass(FedoraBaseInstallClass):
+    name = "Atomic Host"
+    stylesheet = "/usr/share/anaconda/pixmaps/atomic/fedora-atomic.css"
+    sortPriority = 11000
```
poncovka: Priority should be set to `FedoraBaseInstallClass.sortPriority + 1`.

pyanaconda/installclasses/installclass_atomic.py
```diff
@@ -0,0 +1,120 @@
...
+    hidden = False
```
poncovka: Atomic install class can be used in Fedora and RHEL? Otherwise, this line should be removed.

pyanaconda/installclasses/fedora_server.py
```diff
@@ -37,6 +37,10 @@ class FedoraServerInstallClass(FedoraBaseInstallClass):
...
+
+    @staticmethod
+    def createDefaultPartitioning(storage):
```
poncovka: Maybe this method could be provided by `BaseInstallClass`.

jkonecny12: I don't think it is good idea. This is not the default partitioning. Or anything like that.
If we really want to do something like that then we should create new mixin class for it but I don't know where to put it how to name it and so on and I'm not sure if that would be that beneficial.

poncovka: Ok.

pyanaconda/installclasses/installclass_atomic.py
jkonecny12: Again I'm not sure we want to do that. Yeah this one method is the same but what if FedoraServer will change something else in future.

Basically I don't think we want to mix hierarchy in this. If something we can use that mixin but these install classes are not extending one to other by principle.

poncovka: Ok. I don't like this solution, but we will change install classes anyway.

pyanaconda/installclasses/fedora_server.py
jkonecny12: Fixed.

pyanaconda/installclasses/installclass_atomic.py
jkonecny12: Fixed.

pyanaconda/installclasses/installclass_atomic.py
jkonecny12: Fixed.

jkonecny12:
**UPDATED** 
Thanks for your suggestions, they should be implemented now.

jkonecny12:
@cgwalters Could you please verified usage of the default partitioning from Fedora Server install class?

I know there were different values but it seemed to me like they are not intentional.

cgwalters:
> I know there were different values but it seemed to me like they are not intentional.

Right, I don't think there was a strong rationale for having them different.  The new values look fine to me.  (Big picture, for bare metal installations we generally expect sysadmins will want to control things.  For cloud images we use separate kickstarts which is unrelated to this.)

jkonecny12:
jenkins, test this please

PR #1312, opened 2018-01-23
Provide comprehensive log messages about the display mode (#1534474)

poncovka:
Log message for the cmdline installation:
`Display mode is set to 'noninteractive text mode'.`

Log message for the interactive graphical installation:
`Display mode is set to 'interactive graphical mode'.`

Resolves: rhbz#1534474

### DIFF
--- a/pyanaconda/anaconda.py
+++ b/pyanaconda/anaconda.py
@@ -229,6 +229,15 @@ class Anaconda(object):
         return (self._display_mode == DisplayModes.TUI
                 and not self._interactive_mode)
 
+    def log_display_mode(self):
+        if not self.display_mode:
+            log.error("Display mode is not set!")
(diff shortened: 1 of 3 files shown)


pyanaconda/anaconda.py
```diff
@@ -229,6 +229,15 @@ def noninteractive_tui_mode(self):
         return (self._display_mode == DisplayModes.TUI
                 and not self._interactive_mode)
 
+    def report_display_mode(self):
```
jkonecny12: I would rather name this `log_display_mode()`. Report can be anything, log is unambiguous.

poncovka: Fixed.

poncovka:
Updated.

PR #1313, opened 2018-01-23
Provide comprehensive log messages about the display mode

poncovka:
Log message for the cmdline installation:
Display mode is set to 'noninteractive text mode'.

Log message for the interactive graphical installation:
Display mode is set to 'interactive graphical mode'.

(cherry-picked from a commit b17e16e)


poncovka:
Updated.

PR #1315, opened 2018-01-29
Network module master

rvykydal:
This is initial patch for network module.
I think I have quite a good idea of what is going into the module and will start to move the logic from anaconda.

### DIFF
--- a/pyanaconda/modules/network/network.py
+++ b/pyanaconda/modules/network/network.py
@@ -18,12 +18,16 @@
 # Red Hat, Inc.
 #
 
+import pydbus
 from pyanaconda.dbus import DBus
 from pyanaconda.dbus.constants import MODULE_NETWORK_NAME, MODULE_NETWORK_PATH
 from pyanaconda.core.signal import Signal
 from pyanaconda.modules.base import KickstartModule
 from pyanaconda.modules.network.network_interface import NetworkInterface
 from pyanaconda.modules.network.network_kickstart import NetworkKickstartSpecification
+
+HOSTNAME_SERVICE = "org.freedesktop.hostname1"
+HOSTNAME_PATH = "/org/freedesktop/hostname1"
 
 from pyanaconda import anaconda_logging
 log = anaconda_logging.get_dbus_module_logger(__name__)
@@ -35,6 +39,10 @@ class NetworkModule(KickstartModule):
         super().__init__()
         self.hostname_changed = Signal()
         self._hostname = "localhost.localdomain"
+        self.current_hostname_changed = Signal()
+        # TODO fallback solution (no hostnamed) ?
+        self._hostname_service_proxy = pydbus.SystemBus().get(HOSTNAME_SERVICE, HOSTNAME_PATH)
+        self._hostname_service_proxy.PropertiesChanged.connect(self._hostname_service_properties_changed)
 
     def publish(self):
         """Publish the module."""
@@ -70,3 +78,18 @@ class NetworkModule(KickstartModule):
         self._hostname = hostname
         self.hostname_changed.emit()
         log.debug("Hostname is set to %s", hostname)
+
+    def _hostname_service_properties_changed(self, interface, changed, invalidated):
+        if interface == "org.freedesktop.hostname1" and "Hostname" in changed:
+            hostname = changed["Hostname"]
+            self.current_hostname_changed.emit(hostname)
+            log.debug("Current hostname changed to %s", hostname)
+
+    def get_current_hostname(self):
+        """Return current hostname of the system."""
+        return self._hostname_service_proxy.Hostname
+
+    def set_current_hostname(self, hostname):
+        """Set current system hostname."""
+        self._hostname_service_proxy.SetHostname(hostname, False)
+        log.debug("Current hostname is set to %s", hostname)
--- a/pyanaconda/modules/network/network.py
+++ b/pyanaconda/modules/network/network.py
@@ -1,0 +1,72 @@
+#
+# Kickstart module for network and hostname settings
+#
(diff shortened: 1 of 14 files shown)


pyanaconda/modules/network/network.py
```diff
@@ -35,6 +39,10 @@ def __init__(self):
         super().__init__()
         self.hostname_changed = Signal()
         self._hostname = "localhost.localdomain"
+        self.current_hostname_changed = Signal()
+        # TODO fallback solution (no hostnamed) ?
```
rvykydal: I imagine setting Environment property (eg Anaconda, InitialSetup, LiveCD, System, Mock) of the module and make the module behave depending on it. For example either use hostnamed or use socket module or be noop. So do not try to connect to hostnamed service unconditionally like we do here.

pyanaconda/ui/gui/spokes/network.py
```diff
@@ -1541,15 +1544,23 @@ def __init__(self, *args, **kwargs):
...
+        self._network_module.proxy.CurrentHostnameChanged.connect(self._hostname_changed)
         self.network_control_box.connect("nm-state-changed",
                                          self.on_nm_state_changed)
         self.network_control_box.connect("device-state-changed",
                                          self.on_device_state_changed)
         self.network_control_box.connect("apply-hostname",
                                          self.on_apply_hostname)
 
+    def _hostname_changed(self, hostname):
+        gtk_call_once(self._update_hostname)
```
jkonecny12: I think you want to call `async_utils.run_in_loop()` method instead. Ideally as decorator.

pyanaconda/ui/gui/spokes/network.py
```diff
@@ -1541,15 +1544,23 @@ def __init__(self, *args, **kwargs):
         NormalSpoke.__init__(self, *args, **kwargs)
         self.networking_changed = False
         self.network_control_box = NetworkControlBox(self.builder, nmclient, spoke=self)
-        self.network_control_box.hostname = self.data.network.hostname
-        self.network_control_box.current_hostname = network.current_hostname()
+        self._network_module = DBusCachedObserver(MODULE_NETWORK_NAME,
```
poncovka: Use the DBusObjectObserver here.

pyanaconda/ui/gui/spokes/network.py
```diff
@@ -1677,8 +1689,15 @@ class NetworkStandaloneSpoke(StandaloneSpoke):
     def __init__(self, *args, **kwargs):
         StandaloneSpoke.__init__(self, *args, **kwargs)
         self.network_control_box = NetworkControlBox(self.builder, nmclient, spoke=self)
-        self.network_control_box.hostname = self.data.network.hostname
-        self.network_control_box.current_hostname = network.current_hostname()
+
+        self._network_module = DBusCachedObserver(MODULE_NETWORK_NAME,
```
poncovka: Same here.

pyanaconda/ui/tui/spokes/network.py
```diff
@@ -58,8 +60,12 @@ class NetworkSpoke(FirstbootSpokeMixIn, NormalTUISpoke):
     def __init__(self, data, storage, payload, instclass):
         NormalTUISpoke.__init__(self, data, storage, payload, instclass)
         self.title = N_("Network configuration")
+        self._network_module = DBusCachedObserver(MODULE_NETWORK_NAME,
```
poncovka: And here.

tests/pyanaconda_tests/network_test.py
```diff
@@ -280,7 +280,12 @@ def setUp(self):
         self.ksdata_mock = mock.Mock()
         self.ksdata_mock.network = mock.Mock()
 
-    def update_hostname_data_test(self):
+    @patch('pyanaconda.dbus.DBus.get_proxy')
+    def update_hostname_data_test(self, proxy_getter):
+        proxy = mock.Mock()
+        proxy.SetHostname = mock.Mock()
```
poncovka: I am not sure if you need to create a mock for `SetHostname` as well. It should work without it.

pyanaconda/network.py
```diff
@@ -1530,9 +1537,11 @@ def networkInitialize(ksdata):
         logIfcfgFiles(msg)
 
     # initialize ksdata hostname
-    if ksdata.network.hostname is None:
-        hostname = hostname_from_cmdline(flags.cmdline) or DEFAULT_HOSTNAME
-        update_hostname_data(ksdata, hostname)
+    network_proxy = DBus.get_proxy(MODULE_NETWORK_NAME, MODULE_NETWORK_PATH)
+    if network_proxy.Hostname == DEFAULT_HOSTNAME:
+        bootopts_hostname = hostname_from_cmdline(flags.cmdline)
```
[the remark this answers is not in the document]
rvykydal: I don't think modules should parse cmdline. It should be done by anaconda (boss) which later sets up respective module property, which effectively happens two lines below.

pyanaconda/ui/gui/spokes/network.py
rvykydal: I'll change it globally when I get to the network GUI spoke

rvykydal:
Replaced DBusCachedObserver with DBusObjectObserver.

pyanaconda/network.py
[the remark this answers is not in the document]
jkonecny12: If it will be handled by Boss than yes, but I would take caution on doing some complex logic based on the cmdline args. It could require multiple modules to do checks or setup which could result in nasty for synchronization issues.

rvykydal:
jenkins, test this please

rvykydal:
jenkins, test this please

PR #1316, opened 2018-01-31
Improve test run

jkonecny12:
* do not run test under root (all tests were ran as user and then again as sudo)
* add run only nosetests to makefile 
* add run only nosetests to setup-mock-test-env script
* move gui only dependencies to a special category

This will remove `sudo` and `python3-dogtail` from default test dependencies and tests are now almost twice as fast :speedboat:  .

### DIFF
--- a/scripts/testing/setup-mock-test-env.py
+++ b/scripts/testing/setup-mock-test-env.py
@@ -1,5 +1,24 @@
 #!/bin/python3
 #
+# Copyright (C) 2018  Red Hat, Inc.
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# the GNU General Public License v.2, or (at your option) any later version.
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY expressed or implied, including the implied warranties of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General
+# Public License for more details.  You should have received a copy of the
+# GNU General Public License along with this program; if not, write to the
+# Free Software Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
+# 02110-1301, USA.  Any Red Hat trademarks that are incorporated in the
+# source code or documentation are not subject to the GNU General Public
+# License and may only be used or replicated with the express permission of
+# Red Hat, Inc.
+#
+# Red Hat Author(s): Jiri Konecny <jkonecny@redhat.com>
+#
+#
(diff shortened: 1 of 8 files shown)


jkonecny12:
I'm not sure why there is that fail but it looks like a broken Rawhide. I'll try it later if that error will be fixed.

tests/pylint/runpylint.py
```diff
@@ -16,7 +21,8 @@ def __init__(self):
                                 # TODO: Should be fixed with https://github.com/PyCQA/astroid/pull/433
                                 FalsePositive(r"^E1129.*: Context manager 'lock' doesn't implement __enter__ and __exit__.$"),
 
-                                # XXX: This is temporary until koji has python3 versions.
+                                # We have koji used only in make_updates script there is no need to
+                                # add more test dependencies because of this script
```
poncovka: I don't understand the comment.

scripts/testing/setup-mock-test-env.py
```diff
@@ -49,6 +69,19 @@ def _resolve_top_dir():
     return os.path.split(top_dir)[0]
 
 
+def _replace_prefix_paths(paths, prefix):
+    result = []
+
+    for p in paths:
+        if os.path.exists(p):
+            basename = os.path.basename(p)
+            result.append(prefix + basename)
```
poncovka: I would prefer to use `os.path.join`.

poncovka:
Jenkins, test this please.

tests/pylint/runpylint.py
jkonecny12: It is not a temporary fake positive. We don't use this dependency in Anaconda anywhere and I don't think we want to have a new build requires in spec file only because it is used in `make_updates`. However the problem is that pylint is run on all the scripts in so also on `make_updates`.

poncovka: Couldn't we just disable pylint for `import koji` with a comment? Do we need to have a false positive here?

jkonecny12: True enough, ok I'll change that.

scripts/testing/setup-mock-test-env.py
jkonecny12: Fixed.

jkonecny12:
I've fixed what you mentioned and rebased this PR. However, it is still without successful tests. I need to fix those too.

jkonecny12:
jenkins, test this please

jkonecny12:
Remove splitting of GUI testing dependencies commit. Thanks to the pylint issues I don't think it has benefits bigger than drawbacks.

jkonecny12:
jenkins, test this please

PR #1317, opened 2018-01-31
UI support for configuring certain mitigations (#1534833)

M4rtinK:
Add the "KERNEL & PERFORMANCE" spoke to GUI and TUI, which makes
it possible to turn off certain security mitigations.

Mitigations that can currently be turned off:

- PTI
- IBRS
- IBPB

The mechanism is similar to (and compatible with) using the
--append option of the bootloader kickstart command.

If a mitigation is turned off via bootloader --append
(for example with "bootloader --append 'nopti') this
will be reflected in the UI. And turning mitigation on and
off will be reflected in the output kickstart.

Resolves: rhbz#1534833

### DIFF
--- a/pyanaconda/ui/gui/spokes/performance.py
+++ b/pyanaconda/ui/gui/spokes/performance.py
@@ -1,0 +1,103 @@
+# performance spoke class
+#
+# Copyright (C) 2018 Red Hat, Inc.
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# the GNU General Public License v.2, or (at your option) any later version.
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY expressed or implied, including the implied warranties of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General
+# Public License for more details.  You should have received a copy of the
+# GNU General Public License along with this program; if not, write to the
+# Free Software Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
+# 02110-1301, USA.  Any Red Hat trademarks that are incorporated in the
+# source code or documentation are not subject to the GNU General Public
+# License and may only be used or replicated with the express permission of
+# Red Hat, Inc.
+#
+# Red Hat Author(s): Martin Kolman <mkolman@redhat.com>
+#
+
+from pyanaconda.flags import flags
+from pyanaconda.i18n import _, CN_
+
+from pyanaconda.ui.helpers import find_bootopt_mitigations, set_bootopt_mitigations
+from pyanaconda.ui.gui.spokes import NormalSpoke
+from pyanaconda.ui.categories.system import SystemCategory
+
+import logging
+log = logging.getLogger("anaconda")
+
+__all__ = ["PerformanceSpoke"]
+
+
+class PerformanceSpoke(NormalSpoke):
+    """ The Kernel & Performance spoke provides tunables for
+    certain security mitigations, that are kernel based and
+    might incur a performance penalty.
+    """
+
+    builderObjects = ["performanceWindow"]
+
+    mainWidgetName = "performanceWindow"
+    uiFile = "spokes/performance.glade"
+    helpFile = "PerformanceSpoke.xml"
+
+    category = SystemCategory
+
+    title = CN_("GUI|Spoke", "_KERNEL & PERFORMANCE")
+    icon = "system-run-symbolic"
+
+    def __init__(self, *args):
+        NormalSpoke.__init__(self, *args)
+        self._pti_switch = self.builder.get_object("switch_pti")
+        self._ibrs_switch = self.builder.get_object("switch_ibrs")
+        self._ibpb_switch = self.builder.get_object("switch_ibpb")
+
+    def initialize(self):
+        NormalSpoke.initialize(self)
+        self.initialize_start()
+
+        # some mitigations might be disabled in the input kickstart
+        disabled_mitigations = find_bootopt_mitigations(self.data.bootloader.appendLine)
+
+        # set the switches accordingly
+        self._pti_switch.set_active(not disabled_mitigations.no_pti)
+        self._ibrs_switch.set_active(not disabled_mitigations.no_ibrs)
+        self._ibpb_switch.set_active(not disabled_mitigations.no_ibpb)
+
+        # report that we are done
+        self.initialize_done()
+
+    @property
+    def status(self):
+        # set based on mitigations being turned ON/OFF
+        all_enabled = all([self._pti_switch.get_active(),
+                           self._ibrs_switch.get_active(),
+                           self._ibpb_switch.get_active()])
+        if all_enabled:
+            return _("All mitigations enabled")
+        else:
+            return _("Some mitigations disabled")
+
+    @property
+    def mandatory(self):
+        return False
+
+    def apply(self):
+        # set boot options accordingly
+        opts = set_bootopt_mitigations(opts=self.data.bootloader.appendLine,
+                                       no_pti=not self._pti_switch.get_active(),
+                                       no_ibrs=not self._ibrs_switch.get_active(),
+                                       no_ibpb=not self._ibpb_switch.get_active())
+        self.data.bootloader.appendLine = opts
+
+    @property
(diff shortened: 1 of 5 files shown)


pyanaconda/ui/tui/spokes/performance.py
```diff
@@ -0,0 +1,91 @@
...
+from pyanaconda.flags import flags
+from pyanaconda.i18n import N_, _
+
+from pyanaconda.ui.categories.system import SystemCategory
+from pyanaconda.ui.tui.spokes import EditTUISpoke
+from pyanaconda.ui.tui.spokes import EditTUISpokeEntry as Entry
+
+__all__ = ["PerformanceSpoke"]
+
+class PerformanceSpoke(EditTUISpoke):
```
poncovka: And here.

pyanaconda/ui/tui/spokes/performance.py
```diff
@@ -0,0 +1,91 @@
...
+    category = SystemCategory
+
+    edit_fields = [
+        Entry("PTI (Page Table Isolation)", "pti", EditTUISpoke.CHECK, True),
+        Entry("IBRS (Indirect Branch Restricted Speculation)", "ibrs", EditTUISpoke.CHECK, True),
+        Entry("IBPB (Indirect Branch Prediction Barriers)", "ibpb", EditTUISpoke.CHECK, True),
+        ]
+
+    def __init__(self, app, data, storage, payload, instclass):
+        EditTUISpoke.__init__(self, app, data, storage, payload, instclass, "user")
```
rvykydal: Is the "user" okay here?

rvykydal:
Also there are some tests failing.

pyanaconda/iutil.py
```diff
@@ -1288,3 +1291,49 @@ def _wrapper(self, *args, **kwargs):
...
+        opt_set.remove("noibrs")
+
+    # IBPB
+    if no_ibpb:
+        opt_set.add("noibpb")
+    elif "noibpb" in opt_set:
+        opt_set.remove("noibpb")
+
+    # join back to space delimited string and return
+    return " ".join(opt_set)
```
jkonecny12: I really don't think this should be part of the `iutil.py`. It is not used anywhere else than performance spoke so why it is here. We shouldn't use `ituil.py` as code dump for everything. It is hard to find something there that way.

pyanaconda/ui/gui/spokes/performance.py
```diff
@@ -0,0 +1,99 @@
...
+
+        # report that we are done
+        self.initialize_done()
+
+    @property
+    def status(self):
+        # set based on mitigations being turned ON/OFF
+        all_enabled = self._pti_switch.get_active() and \
+                      self._ibrs_switch.get_active() and \
+                      self._ibpb_switch.get_active()
```
jkonecny12: `all()` should be used here instead of `and`.

pyanaconda/ui/tui/spokes/performance.py
```diff
@@ -0,0 +1,91 @@
...
+        return True
+
+    @property
+    def mandatory(self):
+        return False
+
+    @property
+    def status(self):
+        # set based on mitigations being turned ON/OFF
+        all_enabled = self.args.pti and self.args.ibrs and self.args.ibpb
```
jkonecny12: Again use `all` here instead.

pyanaconda/ui/tui/spokes/performance.py
```diff
@@ -0,0 +1,91 @@
...
+        # set boot options accordingly
+        opts = iutil.set_bootopt_mitigations(opts=self.data.bootloader.appendLine,
+                                             no_pti=not self.args.pti,
+                                             no_ibrs=not self.args.ibrs,
+                                             no_ibpb=not self.args.ibpb)
+        self.data.bootloader.appendLine = opts
+
+
+
+
```
jkonecny12: Remove these additional blank lines.

pyanaconda/ui/tui/spokes/performance.py
M4rtinK: Good catch - looks like one can specify the password checking policy like this (kinda strange to have this as part of EditTUISpoke) and the value can be safely left out.

pyanaconda/ui/gui/spokes/performance.py
```diff
@@ -0,0 +1,99 @@
...
+from pyanaconda.ui.gui.spokes import NormalSpoke
+from pyanaconda.ui.categories.system import SystemCategory
+
+import logging
+log = logging.getLogger("anaconda")
+
+__all__ = ["PerformanceSpoke"]
+
+
+class PerformanceSpoke(NormalSpoke):
```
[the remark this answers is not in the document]
M4rtinK: Added.

pyanaconda/ui/tui/spokes/performance.py
M4rtinK: Added as well.

pyanaconda/ui/gui/spokes/performance.py
M4rtinK: Changed.

pyanaconda/ui/tui/spokes/performance.py
M4rtinK: Changed.

pyanaconda/ui/tui/spokes/performance.py
M4rtinK: Removed.

pyanaconda/iutil.py
M4rtinK: Moved to ui/helpers.py - that looks like a good place for it.

pyanaconda/ui/gui/spokes/performance.py
```diff
@@ -0,0 +1,103 @@
...
+        self._ibrs_switch.set_active(not disabled_mitigations.no_ibrs)
+        self._ibpb_switch.set_active(not disabled_mitigations.no_ibpb)
+
+        # report that we are done
+        self.initialize_done()
+
+    @property
+    def status(self):
+        # set based on mitigations being turned ON/OFF
+        all_enabled = all(self._pti_switch.get_active(),
```
jkonecny12: I think you need to wrap these in array `[]` it won't work this way though.

pyanaconda/ui/tui/spokes/performance.py
```diff
@@ -0,0 +1,92 @@
...
+        return True
+
+    @property
+    def mandatory(self):
+        return False
+
+    @property
+    def status(self):
+        # set based on mitigations being turned ON/OFF
+        all_enabled = all(self.args.pti, self.args.ibrs, self.args.ibpb)
```
jkonecny12: Same here.

pyanaconda/ui/gui/spokes/performance.py
```diff
@@ -0,0 +1,103 @@
...
+import logging
+log = logging.getLogger("anaconda")
+
+__all__ = ["PerformanceSpoke"]
+
+
+class PerformanceSpoke(NormalSpoke):
+    """ The Kernel & Performance spoke provides tunables for
+    certain security mitigations, that are kernel based and
+    might incur a performance poenalty.
```
rvykydal: if we do one more round, "penoalty" -> "penalty"

pyanaconda/ui/tui/spokes/performance.py
```diff
@@ -0,0 +1,92 @@
...
+from pyanaconda.ui.categories.system import SystemCategory
+from pyanaconda.ui.tui.spokes import EditTUISpoke
+from pyanaconda.ui.tui.spokes import EditTUISpokeEntry as Entry
+
+__all__ = ["PerformanceSpoke"]
+
+class PerformanceSpoke(EditTUISpoke):
+    """ The Kernel & performance spoke provides tunables for
+    certain security mitigations, that are kernel based and
+    might incur a performance poenalty.
```
rvykydal: if we do one more round, "poenalty" -> "penalty"

pyanaconda/ui/gui/spokes/performance.py
M4rtinK: Bah, indeed! Should be fixed now.

pyanaconda/ui/tui/spokes/performance.py
M4rtinK: Fixed.

PR #1318, opened 2018-02-05
Fix access to the attributes of the method command.

poncovka:
We need to be more careful when we access attributes of the current installation method.

Resolves: rhbz#1530428
Related: rhbz#1515678

### DIFF
--- a/pyanaconda/ui/tui/spokes/installation_source.py
+++ b/pyanaconda/ui/tui/spokes/installation_source.py
@@ -271,7 +271,7 @@ class SpecifyRepoSpoke(NormalTUISpoke, SourceSwitchHandler):
         self._container = ListColumnContainer(1)
 
         dialog = Dialog(_("Repo URL"))
-        self._container.add(EntryWidget(dialog.title, self.data.method.url), self._set_repo_url, dialog)
+        self._container.add(EntryWidget(dialog.title, self._url), self._set_repo_url, dialog)
 
         self.window.add_with_separator(self._container)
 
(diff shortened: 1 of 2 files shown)


poncovka:
Jenkins, test this please.

pyanaconda/ui/tui/spokes/installation_source.py
```diff
@@ -271,7 +271,7 @@ def refresh(self, args=None):
         self._container = ListColumnContainer(1)
 
         dialog = Dialog(_("Repo URL"))
-        self._container.add(EntryWidget(dialog.title, self.data.method.url), self._set_repo_url, dialog)
+        self._container.add(EntryWidget(dialog.title, self._url), self._set_repo_url, dialog)
```
jkonecny12: Would it be the same?

The `self._url` is set in init by:

```
self._url = self.data.url.url
```
which I guess could be different in some situations.

poncovka: I am not sure what you mean. `SpecifyRepoSpoke` is created every time we want to set url, so `self._url` always holds the current value. In case that the current installation method is different, it still makes sense to use `self.data.url.url`, because it is last known value of an attribute we are trying to set.

poncovka:
Jenkins, test this please.

pyanaconda/ui/tui/spokes/installation_source.py
jkonecny12: Ok, then. It looks like it should work :+1:

poncovka:
The failing test is unrelated to this pull request. Merging.

PR #1320, opened 2018-02-07
Add payload_type attribute to payload classes

jcpunk:
For the addons I'm developing, it would be very helpful to have a simple attribute to check the payload type.

What I'm playing with right now only applies to 'dnf' and 'live' payloads.  I didn't find any clean ways to check the install type of the system so this adds attributes and tests.

### DIFF
--- a/pyanaconda/payload/__init__.py
+++ b/pyanaconda/payload/__init__.py
@@ -312,6 +312,7 @@ class Payload(object):
         self.storage = None
         self.instclass = None
         self.txID = None
+        self.payload_type = []
 
         # A list of verbose error strings from the subclass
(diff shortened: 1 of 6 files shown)


jkonecny12:
Hello,

I think you should be able to do ``isinstance(payload, DNFPayload)``.

However, I need to warn you that your addon will be most probably broken soon. We are trying to make Anaconda modular which means that addons would have stable API and they will have basically the same power as other Anaconda parts.
If you want to create your addon now then please be in contact with us and we will try to migrate your addon to avoid breaking.

jcpunk:
That is great solution.  I'll close this out

PR #1333, opened 2018-02-15
Use the KickstartHandler class

poncovka:
We can use the empty kickstart handler provided by pykickstart now.
this should fix the problem with extra `#version` comments in the final
kickstart file.

Since the kickstart handler doesn't hold any sections data by default,
they can be specified in the `sections_data` attribute of the
kickstart specification.

Kickstart modules can also provide their own kickstart handler and
parser if necessary.

**Depends on:** https://github.com/clumens/pykickstart/pull/211

### DIFF
--- a/anaconda.spec.in
+++ b/anaconda.spec.in
@@ -37,7 +37,7 @@ Source0: %{name}-%{version}.tar.bz2
 %define mehver 0.23-1
 %define nmver 1.0
 %define partedver 1.8.1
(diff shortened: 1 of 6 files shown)


poncovka:
Jenkins, test this please

poncovka:
Jenkins, test this please

poncovka:
Jenkins, test this please

PR #1335, opened 2018-02-16
Prevent anaconda-core requiring gjs-console

AdamWill:
Between 28.18-1 and 28.19-1, anaconda-core suddenly grew a
dependency on /usr/bin/gjs-console. This is ultimately due to
this change:
https://pagure.io/packaging-committee/issue/738
That thingy mungs the shebang in fedora-welcome.js from
/usr/bin/env gjs-console to /usr/bin/gjs-console, which in turn
causes RPM's automatic dependency code to make anaconda-core
depend on /usr/bin/gjs-console .

In a sense this is 'correct', but it's clearly problematic,
because it means we wind up depending on gjs (and gtk3!) in
all sorts of cases where we don't want to. This breaks stuff,
for instance arm minimal disk image composes. Ultimately perhaps
the GNOME welcome screen should be split out from anaconda-core,
but for now this will fix things up.

I have already made this change in dist-git and sent out the
28.20-2 build to Rawhide.

### DIFF
--- a/anaconda.spec.in
+++ b/anaconda.spec.in
@@ -1,5 +1,9 @@
 %define livearches %{ix86} x86_64 ppc ppc64 ppc64le
 
+# Avoid anaconda-core requiring gjs-console due to the GNOME welcome
+# screen that's shipped in it
+%global __requires_exclude_from ^%{_datadir}/anaconda/gnome/fedora-welcome.*$
+
 Summary: Graphical system installer
 Name:    anaconda
 Version: @PACKAGE_VERSION@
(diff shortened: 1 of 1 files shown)


jkonecny12:
Jenkins, it's ok to test.

AdamWill:
I put it in because I really did a 28.20-2 build in Rawhide...I wanted the changelog for future package builds to actually include the entry for the 28.20-2 build, since it really happened.

jkonecny12:
@M4rtinK could you please take a look on that. You are doing rawhide builds and you know those scripts.

M4rtinK:
The spec file change now collides with the lates release commit due to the changelog entry - so I'll close this PR and handle the change manually (including placing thwe changelog entry in the correct place).

AdamWill:
Rebased. Should merge fine now.

M4rtinK:
> Rebased. Should merge fine now.

Nice & thanks! :)

PR #1336, opened 2018-02-16
Only show Kernel & performance spoke on x86

M4rtinK:
The kernel and performance spoke provides options to turn
of the following mitigations: KPI, IBRS and IBPB.

Unfortunately it turns out these options are not generally available
on non-x86 RHEL 7 architectures.

So only show the Kernel and performance spoke on x86 and hide it on
all other architectures.

Related: rhbz#1534833


jkonecny12:
It looks like this won't be necessary. The whole GUI spoke may go away.

M4rtinK:
Obsoleted by #1339.

PR #1337, opened 2018-02-16
Use the KickstartError attributes

poncovka:
The kickstart modules are able to return the correct line number and
the original error message about an error in the parsed kickstart
file. It means that we are able to show correct error messages to the
user now.

The function formatErrorMsg from pykickstart is deprecated.

**Depends on:** https://github.com/clumens/pykickstart/pull/212


poncovka:
Jenkins, test this please

poncovka:
Jenkins, test this please

PR #1340, opened 2018-02-19
installclasses/rhelah: Allow LVM thin and dm-crypt

cgwalters:
We support both, no reason not to list them.  (Really we're just
omitting `btr=True` which may still be incompatible with ostree)

https://bugzilla.redhat.com/show_bug.cgi?id=1546252


jkonecny12:
Jenkins, it's ok to test.

PR #1343, opened 2018-02-20
kickstart: "clearpart --list" does not work (#1410335)

mtelka:
Resolves: rhbz#1410335


jkonecny12:
Jenkins, it's ok to test.

PR #1344, opened 2018-02-20
kickstart: "clearpart --list" does not work (#1410335)

mtelka:
Resolves: rhbz#1410335


jkonecny12:
Jenkins, it's ok to test.

jkonecny12:
Thanks for your contribution but to be able to merge this to RHEL we need a bug in bugzilla on RHEL. Without that we can't merge that. 
Bug you have in your commit is filed on Fedora and that is not enough.

mtelka:
Cloned 1410335 as 1561930 for RHEL.

mtelka:
Done. Thanks.

PR #1346, opened 2018-02-21
Use session bus and start session bus if not already running

jkonecny12:
Use DBus session instead of system bus.
Start DBus session bus in Anaconda in case it is not running already. Still requires `run_boss` flag enabled.

Name of the DBus session is written in `/var/run/anaconda/bus.address'.

Tested on DBus running on mock, lorax in mock, livemedia-creator in mock, installation environment.


pyanaconda/anaconda.py
```diff
@@ -337,3 +340,24 @@ def writeXdriver(self, root=None):
...
+
+    def run_dbus_if_required(self):
+        if not launcher.is_dbus_session_running():
+            launcher.start_dbus_session()
+
+    def run_boss(self):
+        bus_proxy = DBus.get_dbus_proxy()
+        bus_proxy.StartServiceByName(DBUS_BOSS_INSTALLATION_NAME,
+                                     DBUS_FLAG_NONE,
+                                     callback=self._check_boss_started)
```
poncovka: Does it really work when you call the method asynchronously? Because there is no loop yet.

pyanaconda/core/constants.py
```diff
@@ -59,6 +59,8 @@
 IMAGE_DIR = MOUNT_DIR + "/image"
 INSTALL_TREE = MOUNT_DIR + "/source"
 BASE_REPO_NAME = "anaconda"
+ANACONDA_RUN_DIR = "/run/anaconda"
+ANACONDA_BUS_ADDR_FILE = "bus.address"
```
poncovka: I think that there should be a full path to the file. It is not very useful this way and you could remove `_make_bus_address_file_path`.

pyanaconda/dbus/__init__.py
```diff
@@ -47,25 +48,30 @@ def get_connection():
         return DBus._connection
 
     @staticmethod
-    def get_new_connection():
+    def get_new_connection(bus_address=None):
```
poncovka: I don't think that this argument should be part this method, because it is not clear how it should be used. 

Since the only way how to use the method with the argument is this:
`DBus._connection = DBus.get_new_connection(TEST_ADDRESS)`

It doesn't look any better then this:
`DBus._connection = pydbus.connect(bus_address)`

I would prefer not to add this confusing option just for testing.

Btw. the best way is to set up the `DBUS_STARTER_ADDRESS` anyway.

pyanaconda/dbus/__init__.py
```diff
@@ -47,25 +48,30 @@ def get_connection():
...
 
-        log.info("Connecting to system DBus.")
-        return pydbus.SystemBus()
+        :param bus_address: Define your specific DBus session.
+                            This parameter is useful especially for testing.
+        """
+        if not bus_address:
+            if DBUS_STARTER_ADDRESS in os.environ:
+                bus_address = os.environ.get(DBUS_STARTER_ADDRESS)
+            elif DBUS_SESSION_ADDRESS in os.environ:
```
poncovka: Have you consider to set up the `DBUS_STARTER_ADDRESS` for the main anaconda process, so we don't have to handle here `DBUS_SESSION_ADDRESS`?

pyanaconda/dbus/launcher.py
```diff
@@ -0,0 +1,104 @@
...
+
+
+def is_dbus_session_running():
+    """Check if dbus session is running.
+
+    :returns: True if DBus is running, False otherwise
+    """
+    if os.environ.get(DBUS_SESSION_ADDRESS):
+        return True
+    elif os.path.exists(_make_bus_address_file_path()):
```
poncovka: If the file exists, but the `DBUS_STARTER_ADDRESS` is not set, anaconda will fail, no? Is the existence of the file really relevant?

pyanaconda/dbus/launcher.py
```diff
@@ -0,0 +1,104 @@
...
+        return True
+
+    return False
+
+
+def get_dbus_session_address():
```
poncovka: This function is never called.

pyanaconda/anaconda.py
jkonecny12: This callback is really called. Tested.
However, yeah it is interesting when GLib as you said is not running there. Maybe it is started by this?

pyanaconda/core/constants.py
jkonecny12: OK

pyanaconda/dbus/__init__.py
jkonecny12: Setting environment variable is not the easy way and it could be really unpleasant to work with. Also I don't see the confusion, you could specifying address for DBus?

pyanaconda/dbus/__init__.py
jkonecny12: But it is a lie, It is not started by DBUS. I think it is not that hard to set `DBUS_SESSION_ADDRESS` instead.

pyanaconda/dbus/launcher.py
jkonecny12: That is not `DBUS_STARTER_ADDRESS` but `DBUS_SESSION_ADDRESS` which must be set.

pyanaconda/dbus/launcher.py
jkonecny12: Nope, but it is really super nice for testing purpose. However, I can remove it if you don't want there "usable for test features".

pyanaconda/anaconda.py
poncovka: Or it might be called synchronously if there is no loop.

pyanaconda/dbus/__init__.py
poncovka: Let's discuss it tomorrow.

pyanaconda/dbus/launcher.py
poncovka: My mistake. I've updated the question.

jkonecny12: In the case that `DBUS_SESSION_ADDRESS` is not set but the file exists means that Anaconda was restarted but it created its own session bus which is not in the OS environments -- can't be. So use this existing session bus instead of creating new one.

pyanaconda/core/constants.py
jkonecny12: Fixed.

jkonecny12:
jenkins, test this please

pyanaconda/dbus/launcher.py
jkonecny12: So what about compromise. I will remove that `os.path.exists()` part and add comment for the whole file with warning that it is really NOT RECOMMENDED to use this in modules and it probably not work.

What do you think?

pyanaconda/anaconda.py
```diff
@@ -337,3 +340,27 @@ def writeXdriver(self, root=None):
...
+
+        launcher.write_bus_address()
+
+    def run_boss(self):
+        bus_proxy = DBus.get_dbus_proxy()
+        bus_proxy.StartServiceByName(DBUS_BOSS_INSTALLATION_NAME,
+                                     DBUS_FLAG_NONE,
+                                     callback=self._check_boss_started)
+
+    def _check_boss_started(self, returned, error):
```
poncovka: This callback is called when the loop is started (which is too late). Please, call `StartServiceByName` synchronously and fail if the Boss fails to start.

jkonecny12: Good point, I missed that it could be called later when the event is put in the queue... My fault.

pyanaconda/anaconda.py
```diff
@@ -337,3 +340,27 @@ def writeXdriver(self, root=None):
...
+    def ensure_running_dbus(self):
+        """Ensure suitable DBus is running. If not, start a new session."""
+        if not launcher.is_dbus_session_running():
+            launcher.start_dbus_session()
+
+        launcher.write_bus_address()
+
+    def run_boss(self):
+        bus_proxy = DBus.get_dbus_proxy()
+        bus_proxy.StartServiceByName(DBUS_BOSS_INSTALLATION_NAME,
```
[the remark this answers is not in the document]
jkonecny12: Interesting, I have a feeling that the boss was started before.

jkonecny12:
**UPDATED:**
* Fixed Boss name
* Fixed async call to check if boss started

PR #1347, opened 2018-02-21
Drop system-logos dependency from anaconda-core

M4rtinK:
The system-logos package is only needed in the installation
environment, so it should be part of the corresponding Lorax template.

Making it an Anaconda dependency results in it being dragged in
also when it is not needed, such as when Anaconda is installed as an
application or as Initial Setup dependency.


M4rtinK:
**NOTE:** Should be merged only after https://github.com/rhinstaller/lorax/pull/327.

M4rtinK:
jenkins, test this please

M4rtinK:
@bcl Do I understand things correctly that I can merge this now ? (eq. that the issues with system-logos in the Lorax template have been fixed)

PR #1349, opened 2018-02-26
 Fix logging of the DBus modules 

poncovka:
Force the message bus to use the system log for messages with the
`--syslog` option. We don't have to set up the loggers for that.

This will not work if the session bus is already running, but we will
move to our own bus anyway.

It doesn't make sense to use special loggers for the DBus modules,
because they import pyanaconda modules that use usual loggers anyway.
If we want to set up logging in DBus modules, we should do that in
the `__main__` files.

We want all levels of logs from the DBus modules for now.


poncovka:
The pylint error is fixed here: https://github.com/rhinstaller/anaconda/pull/1348

PR #1351, opened 2018-02-26
Move all DBus errors to pyanaconda.modules.common.errors

poncovka:
The classes and functions that are common to the modules, or need
to be importable by UI, are moved to pyanaconda.modules.common.

All DBus errors should be defined in pyanaconda.modules.common.errors,
so they can be imported from the client side as well. It is important,
because DBus errors are mapped on exception classes, that can be
reraised on the client side by pydbus.

Move classes for the kickstart dispatcher to the kickstart_manager
package in pyanaconda.modules.boss. Also improve the install_manager
package a little.


poncovka:
However, it might be moved to `pyanaconda.errors`.

jkonecny12:
jenkins, test this please

poncovka:
Updated.

pyanaconda/modules/common/errors/Makefile.am
```diff
@@ -0,0 +1,22 @@
+# Makefile.am for anaconda
```
jkonecny12: I think this shouldn't be here.

pyanaconda/modules/common/errors/__init__.py
```diff
@@ -0,0 +1,32 @@
+#
+# DBus errors.
```
jkonecny12: `Common DBus errors` or similar is better thought.

pyanaconda/modules/common/errors/installation.py
```diff
@@ -0,0 +1,33 @@
+#
+# DBus errors related to tasks and the installation.
```
jkonecny12: `installation tasks and the installation` sounds better to me.

pyanaconda/modules/common/base.py
```diff
@@ -24,7 +24,7 @@
 from pyanaconda.core.async_utils import run_in_loop
 from pyanaconda.core.timer import Timer
 from pyanaconda.dbus import DBus
-from pyanaconda.task import publish_task
+from pyanaconda.modules.common.task import publish_task
```
jkonecny12: I'm still thinking that `task` is not descriptive. Cloud you rename it in process to `installation_task` or similar please?

Or we can do that in special PR. I'll leave that on you.

pyanaconda/modules/common/task.py
```diff
@@ -26,11 +26,24 @@
 
 from pyanaconda.core.signal import Signal
 from pyanaconda.core.constants import THREAD_DBUS_TASK
-from pyanaconda.task.task_interface import TaskAlreadyRunningException
+from pyanaconda.modules.common.task_interface import TaskInterface
+from pyanaconda.modules.common.task_interface import TaskAlreadyRunningException
```
jkonecny12: I'm not fan of having everything in one folder. What about having `pyanaconda.modules.common.task.interface` instead of what is there now.

You know Tasks don't have anything related to base module classes so they shouldn't be together.

pyanaconda/modules/common/base.py
poncovka: But we might use tasks for all kinds of long running jobs.

pyanaconda/modules/common/errors/Makefile.am
poncovka: Fixed.

pyanaconda/modules/common/errors/__init__.py
poncovka: Fixed.

pyanaconda/modules/common/errors/installation.py
poncovka: Task errors moved to task.py.

pyanaconda/modules/common/task.py
poncovka: Moved to a new package pyanaconda.common.task.

poncovka:
Updated.

pyanaconda/modules/common/base.py
jkonecny12: Hmm true. I still don't like task name but I'm not able to find something better so I guess it is fine then.

PR #1352, opened 2018-02-26
Move the kickstart specification to pyanaconda.core.kickstart

poncovka:
The kickstart specification doesn't depend on DBus and it should be
stable, so it can be part of the core package.

Every kickstart module has only one kickstart specification, so the
files can be always called `kickstart.py`.


jkonecny12:
jenkins, test this please

PR #1353, opened 2018-02-27
Enable payload configuration for Install classes

jkonecny12:
Install classes are not able to set-up payload. This PR will enable that possibility.


jkonecny12:
**UPDATED:**
* Rename `configureInstallClass` to `postConfigureInstallClass`

PR #1355, opened 2018-02-27
Add tests for toplevel installclass attribs

jcpunk:
As installclasses get a bit more fancy, there should probably be a test to check for some of its base attributes.

This test adds some very simple checks.


M4rtinK:
jenkins, test this please

PR #1357, opened 2018-02-27
 Use observers to access the hostname service

poncovka:
On some systems, the system services and the system bus don't have
to run. We should set up observers to access the hostname service
and call connect_once_available to make sure that the system bus
is checked before we try to connect to it.

### DIFF
--- a/pyanaconda/dbus/connection.py
+++ b/pyanaconda/dbus/connection.py
@@ -63,6 +63,17 @@ class Connection(ABC):
         """
         pass
 
+    def check_connection(self):
+        """Check if the connection is set up.
+
+        :return: True if the connection is set up otherwise False
+        """
+        try:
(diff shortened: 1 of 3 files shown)


pyanaconda/dbus/connection.py
```diff
@@ -63,6 +63,17 @@ def get_new_connection(self):
         """
```
jkonecny12: I guess that commit message should be:
``Make safe to observe services on buses that don't have to run``
not actual:
``Make is safe to observe services on buses that don't have to run``

poncovka:
Updated.

pyanaconda/dbus/connection.py
poncovka: Fixed.

PR #1358, opened 2018-02-28
network: set TYPE value in ifcfg from kickstart in initrmfs

rvykydal:
Fixes a bunch of kickstart tests (eg ifname-httpks) for network module:
In network module we use the info from ifcfg files when looking for ifcfg file
(we don't get type of the device from NM).

Also be consistent with ifcfg files created by dracut.


rvykydal:
Replaced by f28-devel PR

PR #1360, opened 2018-02-28
Add the kernel option resume= by default (#1206936)

poncovka:
The option resume= with a path to the largest available swap device
will be added by default to the kernel options to enable hibernation.

Resolves: rhbz#1206936


jkonecny12:
jenkins, test this please

PR #1362, opened 2018-03-01
Clean dbus resources

jkonecny12:
Stop DBus session and stop modules when Anaconda stops.


jkonecny12:
jenkins, test this please

PR #1363, opened 2018-03-01
network: set TYPE value in ifcfg from kickstart in initrmfs

rvykydal:
Fixes a bunch of kickstart tests (eg ifname-httpks) for network module:
In network module we use the info from ifcfg files when looking for ifcfg file
(we don't get type of the device from NM).

Also be consistent with ifcfg files created by dracut.

### DIFF
--- a/dracut/parse-kickstart
+++ b/dracut/parse-kickstart
@@ -442,11 +442,15 @@ def ksnet_to_ifcfg(net, filename=None):
         if not os.path.isdir(TMPDIR+"/ifcfg"):
             os.makedirs(TMPDIR+"/ifcfg")
     hwaddr = readsysfile("/sys/class/net/%s/address" % dev)
+    devtype = nm_type_of_device(dev)
+    if devtype:
+        ifcfg['TYPE'] = devtype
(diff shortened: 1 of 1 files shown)


dracut/parse-kickstart
```diff
@@ -442,11 +442,14 @@ def ksnet_to_ifcfg(net, filename=None):
...
     hwaddr = readsysfile("/sys/class/net/%s/address" % dev)
+    devtype = nm_type_of_device(dev)
+    if devtype:
+        ifcfg['TYPE'] = devtype
     if net.bindto == BIND_TO_MAC and not (net.bridgeslaves or net.bondslaves or net.teamslaves or net.vlanid):
         # ifcfg-rh requires DEVICE or TYPE to be set
         ifcfg['HWADDR'] = hwaddr
         # fall back to Ethernet rather then crash the installation later
-        ifcfg['TYPE'] = nm_type_of_device(dev) or "Ethernet"
+        ifcfg['TYPE'] = devtype or "Ethernet"
```
jkonecny12: I don't see a point in trying `devtype` here. If the `devtype` exists then it is already set above. You should instead check if `devtype` exists or if `ifcfg['TYPE']` is set and react on that.

jkonecny12:
jenkins, test this please

rvykydal:
jenkins, test this please

PR #1365, opened 2018-03-02
F28 devel localization module

rvykydal:
This is the first patchset adding the module with support for lang command. I'll add keyboard in following patches.

Lang related kickstart tests passed for me:
- lang
- packages-instlangs-1
- packages-instlangs-2
- packages-instlangs-3


pyanaconda/modules/localization/Makefile.am
```diff
@@ -0,0 +1,22 @@
+# modules/localization/Makefile.am for anaconda
```
poncovka: Please, remove this line.

pyanaconda/modules/localization/localization_kickstart.py
```diff
@@ -0,0 +1,30 @@
+#
```
poncovka: Please, rename this file to `kickstart.py`.

PR #1366, opened 2018-03-02
Add the Security module

poncovka:
The Security module handles kickstart commands auth, authconfig,
authselect, selinux and realm. The module can parse them and generate
them and it provides DBus API to access parameters of these commands.

Modify the kickstart commands auth, authconfig, authselect, realm
and selinux in UI and use the Security module instead.


tests/lib/utils.py
```diff
@@ -0,0 +1,51 @@
...
+    # Generate a kickstart
+    ks_out = dedent(ks_out).strip()
+    test.assertEqual(ks_out, interface.GenerateKickstart().strip())
+
+    # Test the properties changed callback.
+    if ks_in is not None:
+        callback.assert_any_call(DBUS_MODULE_NAMESPACE, {'Kickstarted': True}, [])
+    else:
+        test.assertEqual(interface.Kickstarted, False)
+        callback.assert_not_called()
```
jkonecny12: Please move this file to `pyanaconda_tests/__init__.py` file instead.
This lib directory is for general use stuff not for just unit tests.

poncovka: Fixed.

pyanaconda/kickstart.py
```diff
@@ -293,25 +295,32 @@ def parse(self, args):
...
 
-class Authselect(commands.authselect.F28_Authselect):
+class Authselect(RemovedCommand):
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self.packages = []
 
+    def __str__(self):
+        return ""  # It is generated by the SELinux command.
```
M4rtinK: I think some more backstory for this is needed.

PR #1369, opened 2018-03-05
Always run anaconda private dbus session

jkonecny12:
We have our private DBus with our specific configuration.
Our private DBus is launched every time when Anaconda is started.
We are using our dbus variable instead of `DBUS_SESSION_BUS_ADDRESS`.


data/dbus/session.conf
```diff
@@ -0,0 +1,126 @@
+<!-- This configuration file controls Anaconda's special session message bus.
+     It is based on the /usr/share/dbus-1/session.conf file.
+
+     Author(s): Jiri Konecny <jkonecny@redhat.com
+-->
+
```
poncovka: Please, change the name of the file.

jkonecny12:
jenkins, test this please

data/dbus/session.conf
poncovka: The config file looks to be too strict, but that is fine for now.

jkonecny12: We can change the policy anytime we want.

jkonecny12:
**UPDATED:**
* Rename `session.conf` to `anaconda-bus.conf`.
* Fix check if dbus exits properly.

PR #1380, opened 2018-03-08
Replace DBus constants for names and paths

poncovka:
DBus objects, interfaces and services can be described with namespaces
and identifiers. They provide shorcuts to proxies and observers.

DBus errors should be defined with the dbus_error decorator that
requires the error name and the error namespace. Default DBus
error can be defined with dbus_error_by_default.

### DIFF
--- a/pyanaconda/dbus/identifier.py
+++ b/pyanaconda/dbus/identifier.py
@@ -1,0 +1,176 @@
+#
+# Identification of DBus objects, interfaces and services.
+#
+# Copyright (C) 2018  Red Hat, Inc.  All rights reserved.
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+from pyanaconda.dbus import DBus
+from pyanaconda.dbus.namespace import get_dbus_path, get_dbus_name
+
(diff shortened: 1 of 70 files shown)


pyanaconda/modules/common/constants/interfaces.py
```diff
@@ -0,0 +1,37 @@
...
+from pyanaconda.modules.common.constants.namespaces import ANACONDA_NAMESPACE, BOSS_NAMESPACE, \
+    MODULES_NAMESPACE
+
+
+KICKSTART_MODULE = DBusInterfaceIdentifier(
+    namespace=MODULES_NAMESPACE
+)
+
+
+BOSS_ANACONDA = DBusInterfaceIdentifier(
```
jkonecny12: Please remove this empty space or have two empty spaces everywhere.

pyanaconda/dbus/identifier.py
```diff
@@ -0,0 +1,177 @@
...
+def version_to_string(version):
+    """Convert version to a string.
+
+    :param version: a number or None
+    :return: a string
+    """
+    if version is None:
+        return ""
+
+    return str(version)
```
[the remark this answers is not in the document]
poncovka: Fixed.

pyanaconda/modules/common/constants/interfaces.py
poncovka: Fixed.

pyanaconda/dbus/identifier.py
```diff
@@ -0,0 +1,177 @@
...
+        return self.service_name
+
+    def get_proxy(self, object_path=None):
+        """Returns a proxy of the DBus object.
+
+        :param object_path: a DBus path an object
+        :return: a proxy object
+        """
+        if object_path is None:
+            object_path = self.object_path
```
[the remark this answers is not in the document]
poncovka: I've tried that, but the code is just too confusing. Also, I could use that method only at two places, because the third one has additional logic.

poncovka:
Rebased.

pyanaconda/modules/common/base/base.py
```diff
@@ -73,7 +73,7 @@ class KickstartModule(BaseModule):
 
     def __init__(self):
         super().__init__()
-        self._published_tasks = []
+        self._published_tasks = {}
```
M4rtinK: I think think a plain {} creates a set, not a dict, right ?

PR #1381, opened 2018-03-08
Fix protected device handling for live backing device. (#1524700)

dwlehman:
Because of isohybrid craziness, the live backing device is a partition on a block device that has multiple formats. Blivet treats these devices as containing only the iso9660 formatting. This patch changes `InstallerStorage._find_live_backing_device` so that it marks the disk containing a live backing device partition as protected instead of marking the partition itself. Blivet will never have a device to represent the partition, so this is the only way to mark the media as protected.

### DIFF
--- a/pyanaconda/storage/osinstall.py
+++ b/pyanaconda/storage/osinstall.py
@@ -1701,11 +1701,19 @@ class InstallerStorage(Blivet):
             if " /run/initramfs/live " not in mnt:
                 continue
 
-            live_device_name = mnt.split()[0].split("/")[-1]
-            log.info("%s looks to be the live device; marking as protected",
-                     live_device_name)
-            self.protected_dev_names.append(live_device_name)
-            self.live_backing_device = live_device_name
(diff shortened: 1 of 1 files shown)


dwlehman:
This requires storaged-project/blivet#668.

jkonecny12:
Just please fix these pylint errors:
```
************* Module ../pyanaconda/storage/osinstall.py
E1123(unexpected-keyword-arg):1705,26: InstallerStorage._find_live_backing_device: Unexpected keyword argument 'device_node' in function call
E1120(no-value-for-parameter):1705,26: InstallerStorage._find_live_backing_device: No value for argument 'sysfs_path' in function call
FAIL pylint/runpylint.py (exit status: 2)
```

dwlehman:
Both of the pylint errors are, I think, because it was tested against a blivet that doesn't have the changes from storaged-project/blivet#668.

jkonecny12:
jenkins, test this please

jkonecny12:
jenkins, test this please

PR #1385, opened 2018-03-09
network module: fix accessing org.freedesktop.hostname1 for current h…

rvykydal:
…ostname

The org.freedesktop.hostname1 service is expected to be started by a method
call and is shut down when not used so using observer does not fit here.

### DIFF
--- a/pyanaconda/modules/network/network.py
+++ b/pyanaconda/modules/network/network.py
@@ -18,7 +18,7 @@
 # Red Hat, Inc.
 #
 
-from pyanaconda.dbus import DBus
+from pyanaconda.dbus import DBus, SystemBus
 from pyanaconda.core.signal import Signal
 from pyanaconda.modules.common.base import KickstartModule
 from pyanaconda.modules.common.constants.services import NETWORK, HOSTNAME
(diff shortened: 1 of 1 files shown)


rvykydal:
jenkins, test this please

pyanaconda/modules/network/network.py
```diff
@@ -38,17 +42,10 @@ def __init__(self):
...
-        service = HOSTNAME.get_cached_observer()
-
-        service.cached_properties_changed.connect(
-            self._hostname_service_properties_changed)
-
-        service.connect_once_available()
-        return service
+        self._hostname_service_proxy = None
+        if SystemBus.check_connection():
+            self._hostname_service_proxy = SystemBus.get_proxy(HOSTNAME_SERVICE, HOSTNAME_PATH)
```
poncovka: You can create the proxy as `HOSTNAME.get_proxy()`.

pyanaconda/modules/network/network.py
```diff
@@ -85,25 +82,25 @@ def set_hostname(self, hostname):
         self.hostname_changed.emit()
         log.debug("Hostname is set to %s", hostname)
 
-    def _hostname_service_properties_changed(self, observer, changed, invalid):
-        if "Hostname" in changed:
-            hostname = self._hostname_service.cache.Hostname
+    def _hostname_service_properties_changed(self, interface, changed, invalid):
+        if interface == HOSTNAME_INTERFACE and "Hostname" in changed:
```
poncovka: Use `HOSTNAME.interface_name` instead of `HOSTNAME_INTERFACE`.

PR #1387, opened 2018-03-12
Remove spurious echo call from tmux service file

rvykydal:
Resolves: rhbz#1554243

Port of commit
b29bdff2e041ec4f354c3334d8aa15ea8366a17

Looks like systemd 236 does not like the escape sequence in the
call, which most likely tries to ring the terminal bell (!).

We have no evidence of this ever working & dropping the line makes
Tmux work again, so just drop the line - at least for now.


rvykydal:
See comments in https://github.com/rhinstaller/anaconda/commit/ab29bdff2e041ec4f354c3334d8aa15ea8366a17
We should probably review the fix some time.

PR #1388, opened 2018-03-12
Create the Services module

poncovka:
The Service module handles the kickstart commands:
`services`, `firstboot`, `skipx` and `xconfig`.

The module is able to parse and generate the commands
and it provides DBus API for changing their values.


poncovka:
Rebased. The issues with Enum will be fixed in another pull request.

PR #1391, opened 2018-03-13
Initial module enablement and installation support

M4rtinK:
Use new DNF API entry points to make it possible to enable install
modules. At the same time we can radically simplify our group exclusion
and environment resolution code by making full use of the new API.

The main change is that instead of incrementally calling "install package"
and "select group" we now gather all the things we want to include
and exclude in the installation transaction and feed it to DNF at once.

In short what this does is:
- makes it possible to install modules, including stream id & profile
- makes it possible to enable modules without installing them
- drops all our group exclusion logic as DNF can now handle that via the new API
- makes it possible to mark environments as installed, meaning we no longer need to resolve environments to groups and that environments are correctly marked as installed on the target system
- overall a nice code cleanup

Possible followup items:
- all in one transaction error handling: tell users all problem with a transaction, not one issue at a time like now
- make it possible to specify all supported package types per group

### DIFF
--- a/pyanaconda/payload/dnfpayload.py
+++ b/pyanaconda/payload/dnfpayload.py
@@ -37,6 +37,7 @@ import configparser
 import collections
 import itertools
 import multiprocessing
+import multiprocessing.dummy
 import operator
 import hashlib
 import shutil
@@ -59,6 +60,8 @@ import libdnf.conf
 import dnf.conf.substitutions
 import rpm
 import librepo
+
+from dnf.const import GROUP_PACKAGE_TYPES
 
 DNF_CACHE_DIR = '/tmp/dnf.cache'
 DNF_PLUGINCONF_DIR = '/tmp/dnf.pluginconf'
@@ -87,6 +90,18 @@ USER_AGENT = "%s (anaconda)/%s" % (productName, productVersion)
 # 6KiB = 4K(max default fragment size) + 2K(rpm db could be taken for a header file)
 BONUS_SIZE_ON_FILE = Size("6 KiB")
 
+class InstallSpecsMissing(Exception):
+    """Raised is some of the requested install specs seems to be missing from the repos."""
+    def __init__(self, missing_specs):
+        super().__init__()
+        self.missing_specs = missing_specs
+        self.missing_packages = []
+        self.missing_groups_and_modules=[]
+        for spec in missing_specs:
+            if spec.startswith("@"):
+                self.missing_groups_and_modules.append(spec)
+            else:
+                self.missing_packages.append(spec)
 
 def _failure_limbo():
     progressQ.send_quit(1)
@@ -307,6 +322,8 @@ class DNFPayload(payload.PackagePayload):
         # save repomd metadata
         self._repoMD_list = []
 
+        self._req_groups = set()
+        self._req_packages = set()
         self.requirements.set_apply_callback(self._apply_requirements)
 
     def unsetup(self):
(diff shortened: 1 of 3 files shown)


M4rtinK:
This PR is still rather preliminary at this point in time:
- a pykickstart PR is needed https://github.com/clumens/pykickstart/pull/213
- depends on PoC DNF API provided by mhatina (Thanks!)
- requires compose to include modular metadata (still missing AFAIK)

Only once all these issues land and are fixed we can think about merging this PR.

pyanaconda/payload/dnfpayload.py
```diff
@@ -412,80 +414,149 @@ def addRepo(self, ksrepo):
...
+                types = "mandatory"
+                if default:
+                    types = "{},default".format(types)
+                if optional:
+                    types = "{},optional".format(types)
+
+                group_spec = "@{group_name}/{types}".format(group_name=group.name, types=types)
+            else:
+                # if group is a regular group this is equal to @group/mandatory,default
+                group_spec = "@{group_name}".format(group_name=groups.name)
```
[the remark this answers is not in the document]
M4rtinK: Indeed! Fixed.

pyanaconda/payload/dnfpayload.py
```diff
@@ -412,80 +414,149 @@ def addRepo(self, ksrepo):
...
+        if kernel_package:
+            selected_packages.add(kernel_package)
+
+        # resolve packages and groups required by Anaconda
+        self.requirements.apply()
 
-        self._select_kernel_package()
+        # let's piece it all together and feed it to DNF at once
+        include_set = set()
+        exclude_set = set()
```
poncovka: Couldn't we use these sets from the beginning?

pyanaconda/payload/dnfpayload.py
```diff
@@ -412,80 +414,149 @@ def addRepo(self, ksrepo):
...
+        # note about formatting:
+        # - leading @ signifies a group or module
+        # - no leading @ means a package
+
+        # add environment
+        if env:
+            include_set.add("@^{}".format(env))
```
poncovka: Can we create this specification when we choose the environment?

pyanaconda/payload/dnfpayload.py
```diff
@@ -412,80 +414,149 @@ def addRepo(self, ksrepo):
...
+        # feed it to DNF
+        try:
+            self._base.install_specs(install=include_set, exclude=exclude_set)
+        except Exception as e:  # pylint: disable=broad-except
+            exception_message = str(e)
+            self._payload_setup_error(e)
 
     def _apply_requirements(self, requirements):
+        self._req_groups = set()
+        self._req_packages = set()
```
poncovka: The solution with a callback is not useful anymore. Ideally, this callback could be replaced with two methods `get_required_packages` and `get_required_groups` that could return a set of packages and a set of groups, so we can avoid these attributes.

pyanaconda/payload/dnfpayload.py
```diff
@@ -412,80 +414,149 @@ def addRepo(self, ksrepo):
...
-            try:
-                self._select_group(req.id, required=req.strong)
-                log.debug("selected group: %s, requirement for %s",
-                           req.id, req.reasons)
-            except payload.NoSuchGroup as e:
-                self._miss(e)
+            # NOTE: req.strong not handled yet
+            log.debug("selected group: %s, requirement for %s",
+                       req.id, req.reasons)
+            self._req_groups.add(req.id)
```
poncovka: Please, add here the group specification, so you don't have to do it later.

pyanaconda/payload/dnfpayload.py
```diff
@@ -412,80 +414,149 @@ def addRepo(self, ksrepo):
         super(DNFPayload, self).addRepo(ksrepo)
 
     def _apply_selections(self):
```
poncovka: This method could be probably split into several simpler ones, that can be tested with unittests.

pyanaconda/payload/dnfpayload.py
```diff
@@ -412,80 +414,149 @@ def addRepo(self, ksrepo):
...
+            else:
+                module_spec = module.name
+            log.debug("enabling module: {}".format(module_spec))
+            # TODO:
+            # - is this cleared by self._base.reset() ?
+            # - what does the save_immediately parameter of enable_module() do ?
+            self._base.enable_module([module_spec])
+
+        # handle "normal" groups
+        group_specs = set()
```
poncovka: I would prefer `selected_group_specs`, so it is consistent with other variables.

pyanaconda/payload/dnfpayload.py
rvykydal: +1, we should just think if/how we want to keep the possibility of checking if (all the) the requirements were actually applied by the payload

M4rtinK:
It is now actually possible to install modules with this PR while using F28 compose repositories with the following kickstart:

```
graphical
url --url=https://kojipkgs.fedoraproject.org/compose/branched/Fedora-28-20180322.n.0/compose/Server/x86_64/os/
repo --name=modular --baseurl https://kojipkgs.fedoraproject.org/compose/branched/Fedora-28-20180322.n.0/compose/Modular/x86_64/os/
lang en_US.UTF-8
keyboard us
rootpw 12345
timezone America/New_York --nontp
zerombr
clearpart --all --initlabel
autopart
#module --name=django --stream=1.6
module --name=nodejs --stream=6

%packages
#@nodejs:6/default
@django:1.6/default
vim
%end
```
A couple of notes though:
- note how we need to add the modular repository on top of the _normal_ server repo, this is the intended way of doing things as far as I can tell and it is supposed to work at least with the Server repo
- note that the timezone command uses the --nontp option, this is because the installed system is missing chrony for some reason and the installation would fail otherwise
- the installed system is **not bootable**
- the nodejs module is commented out in the %packages section as it fails to install at the moment

So I'll try to track down why chrony is missing & why the system is not bootable.

pyanaconda/payload/dnfpayload.py
```diff
@@ -412,80 +414,149 @@ def addRepo(self, ksrepo):
...
+            # stream definition is optional
+            if module.stream:
+                module_spec = "{name}:{stream}".format(name=module.name, stream=module.stream)
+            else:
+                module_spec = module.name
+            log.debug("enabling module: {}".format(module_spec))
+            # TODO:
+            # - is this cleared by self._base.reset() ?
+            # - what does the save_immediately parameter of enable_module() do ?
+            self._base.enable_module([module_spec])
```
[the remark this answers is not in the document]
M4rtinK: Done. :)

pyanaconda/payload/dnfpayload.py
```diff
@@ -412,80 +414,149 @@ def addRepo(self, ksrepo):
...
+
+        # handle "normal" groups
+        group_specs = set()
+        excluded_group_specs = set()
+        for excluded_group in [group.name for group in self.data.packages.excludedGroupList]:
+            excluded_group_specs.add("@{}".format(excluded_group))
```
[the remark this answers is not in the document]
M4rtinK: Yep, indeed & thanks! :)

pyanaconda/payload/dnfpayload.py
```diff
@@ -412,80 +414,149 @@ def addRepo(self, ksrepo):
...
 
-        excludedGroups = [group.name for group in self.data.packages.excludedGroupList]
-
         if env:
-            try:
-                self._select_environment(env, excludedGroups)
-                log.info("selected env: %s", env)
-            except payload.NoSuchGroup as e:
-                self._miss(e)
+            log.info("selected env: %s", env)
```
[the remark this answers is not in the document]
M4rtinK: Good point!

pyanaconda/payload/dnfpayload.py
```diff
@@ -412,80 +414,149 @@ def addRepo(self, ksrepo):
...
+            # only use type specififcations if --nodefault,
+            # --optional or both are used
+            if not default or optional:
+                types = "mandatory"
+                if default:
+                    types = "{},default".format(types)
+                if optional:
+                    types = "{},optional".format(types)
+
+                group_spec = "@{group_name}/{types}".format(group_name=group.name, types=types)
```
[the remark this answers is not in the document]
M4rtinK: Yep, that's definitely cleaner. I've also changed this section a bit so it takes the default group types to install from the DNF API.

pyanaconda/payload/dnfpayload.py
M4rtinK: Hmm, the original idea was that we keep the individual package/group/module sources separate for easier debugging and finding out what comes from where. But we should have most of these items logged already, so using the same two data structures from the beginning might should work now.

pyanaconda/payload/dnfpayload.py
M4rtinK: Done.

pyanaconda/payload/dnfpayload.py
M4rtinK: Definitely not looking nice at the moment. But on the other hand there are already many changes in this PR. So I'd personally prefer to handle this better in a followup PR.

pyanaconda/payload/dnfpayload.py
M4rtinK: As mentioned, I'd prefer to do the requirements handling adjustment in a followup PR.

pyanaconda/payload/dnfpayload.py
M4rtinK: Dropped in favor of common include/exclude sets.

pyanaconda/payload/dnfpayload.py
M4rtinK: I'm not sure this is required anymore with the common include/exclude sets as there are now just short pieces of code filling one of the sets.

pyanaconda/payload/dnfpayload.py

M4rtinK:
Updated the PR:
- fixed some Pylint errors (the DNF related error are due to using not yet released DNF APIs)
- split the module enabling code to a separate method
- used common include/exclude sets

jkonecny12:
jenkins, test this please

M4rtinK:
jenkins, test this please

PR #1394, opened 2018-03-14
Fix the kickstart specification handler

poncovka:
The kickstart specification handler should check if kickstart commands
are able to access their dataClass property. Otherwise, the given
kickstart specification might be invalid.

Also, the kickstart handler should be able to handle a specification
with one section and section data.

These cases are covered by new tests now.

### DIFF
--- a/pyanaconda/core/kickstart.py
+++ b/pyanaconda/core/kickstart.py
@@ -88,7 +88,11 @@ class KickstartSpecificationHandler(KickstartHandler):
         setattr(self, name, obj)
 
         # Write the section at the end of the output.
-        write_priority = max(self._writeOrder.keys()) + 100
+        write_priority = 0
+
+        if self._writeOrder:
+            write_priority = max(self._writeOrder.keys()) + 100
+
         self._writeOrder[write_priority] = [obj]
 
 
--- a/tests/pyanaconda_tests/kickstart_specification_test.py
+++ b/tests/pyanaconda_tests/kickstart_specification_test.py
@@ -1,0 +1,257 @@
+#
+# Copyright (C) 2018  Red Hat, Inc.
+#
+# This copyrighted material is made available to anyone wishing to use,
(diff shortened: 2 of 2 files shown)


pyanaconda/core/kickstart.py
```diff
@@ -72,23 +72,34 @@ def __init__(self, specification):
...
         for name, data in specification.commands_data.items():
             self.registerData(name, data)
 
+        for name, command in specification.commands.items():
+            self.registerCommand(name, command)
+
         for name, data in specification.sections_data.items():
             self.registerSectionData(name, data)
 
+    def registerCommand(self, name, command):  # pylint: disable=arguments-differ
```
jkonecny12: Why do we need here that pylint disable?

tests/pyanaconda_tests/kickstart_specification_test.py
```diff
@@ -0,0 +1,165 @@
...
+        }
+
+    class SpecificationD(KickstartSpecification):
+
+        commands = {
+            "user": F24_User,
+        }
+
+        commands_data = {
+            # "UserData": F19_UserData,
```
jkonecny12: Why is it commented out?

pyanaconda/core/kickstart.py
```diff
@@ -72,23 +72,34 @@ def __init__(self, specification):
...
+        """Register a command."""
+        super().registerCommand(name, command)
+
+        # Test if the handler provides the required data objects.
+        getattr(self.commands[name], "dataClass")
```
jkonecny12: What is the point of this test? If I understand it correctly this will raise AttributeError when the "dataClass" is not there. If I understand it correctly then please make it more obvious what will happen and why the test is required here.

poncovka:
**Updated**. The check for the kickstart handler was moved to the unittests and the unittests were extended

PR #1395, opened 2018-03-15
Don't autoquit by default if the last hub is empty (#1553935)

M4rtinK:
With the Reduce Initial Setup Redundancy change[0] in F28
we now encounter an edge case during the Fedora Workstation Live
installation which Anaconda does not handle particularly well.

There are no longer any spokes on the progress hub, which triggers
a piece of code originally added for Initial Setup, which autoclicks
the continue button, effectively quiting the installation as there are
no more hubs present.

While this actually does not influence the correctness of the
installation (this only happens after a successfull installation),
it's a pretty bad UX and as reported by Adam Wiliamson complicates
Open QA based installation testing.

So add a new property for hubs that specifies if a hub should
automaqtically continue if no spokes are present. This property
will be False by defaul but Initial Setup is expected to set it
to True for it's own hub.

Resolves: rhbz#1553935

[0] https://fedoraproject.org/wiki/Changes/ReduceInitialSetupRedundancy

### DIFF
--- a/pyanaconda/ui/gui/hubs/__init__.py
+++ b/pyanaconda/ui/gui/hubs/__init__.py
@@ -59,6 +59,11 @@ class Hub(GUIObject, common.Hub):
     handles_autostep = True
     _hubs_collection = []
 
+    # Should we automatically go to next hub if processing is done and there are no
+    # spokes on the hub ? The default value is False and Initial Setup will likely
+    # override it to True in it's hub.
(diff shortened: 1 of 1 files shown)


M4rtinK:
Related Initial Setup PR: https://github.com/rhinstaller/initial-setup/pull/49

M4rtinK:
jenkins, test this please

PR #1399, opened 2018-03-19
Use enums in DBus modules

poncovka:
Use enums for SELinux modes and first boot actions.
We want to use enums inside the modules, so we can
validate the given values with them.


jkonecny12:
Don't we need a change also in Initial-setup because of this?

poncovka:
@jkonecny12, we don't need to change the initial setup, because the DBus API is not affected by this change.

jkonecny12:
Than it is even better :+1:

PR #1375, opened 2018-03-05
Enable /boot on btrfs subvolume with GRUB2

npmccallum:
This depends on a version of grubby with btrfs subvolume support. For
example, grubby >= 8.40-10.fc29.

PR #1400, opened 2018-03-19
Add the "anaconda.experimental" flag

npmccallum:
The intent of this patch is to make it possible to test not yet
supported configurations in Anaconda without requiring a full compose.
Currently, it only allows you to bypass the stage2 whitelist. However,
this flag may be reused to enable other workflows in the future.

### DIFF
--- a/pyanaconda/bootloader.py
+++ b/pyanaconda/bootloader.py
@@ -44,6 +44,9 @@ from pyanaconda.core.i18n import _, N_
 from pyanaconda.anaconda_loggers import get_module_logger
 log = get_module_logger(__name__)
 
+with open("/proc/cmdline") as f:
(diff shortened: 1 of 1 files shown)


vpodzime:
Such a flag is a good idea. But there needs to be a well-defined list of experimental features. And it should disable reporting failures to bugzilla.

Conan-Kudo:
> And it should disable reporting failures to bugzilla.

I disagree with this statement. Instead, it should actually include prominently which experimental flag was used, because otherwise there's zero chance in hell anyone is going to know about problems to be able to fix them so that they can graduate to fully supported status.

jkonecny12:
Closing this PR. We are in favor of this https://github.com/rhinstaller/anaconda/pull/1375. 
The change will be community maintained and having a flag to turn feature on or off will not prevent bugs in happening so or so.

PR #1403, opened 2018-03-20
Make the class for removed kickstart commands more strict

poncovka:
The `RemovedCommand` class now requires to override its `__str__`
method, so we don't forget to use DBus modules to generate their
part of a kickstart file.

Use the `UselessCommand` class when the command doesn't do anything.


poncovka:
Jenkins, test this please

PR #1407, opened 2018-03-26
Use the disk selection and initialization modules in UI

poncovka:
Use the DBus modules to access values of the kickstart commands:
zerombr, ignoredisk and clearpart.

**This pull request might be merged only on the master branch.**

**TODO:** Run kickstart tests.

### DIFF
--- a/pyanaconda/storage/osinstall.py
+++ b/pyanaconda/storage/osinstall.py
@@ -32,7 +32,7 @@ gi.require_version("BlockDev", "2.0")
 
 from gi.repository import BlockDev as blockdev
 
-from pykickstart.constants import AUTOPART_TYPE_LVM, CLEARPART_TYPE_NONE, CLEARPART_TYPE_LINUX, CLEARPART_TYPE_ALL, CLEARPART_TYPE_LIST
+from pykickstart.constants import AUTOPART_TYPE_LVM
 
 from blivet import arch, udev
 from blivet import util as blivet_util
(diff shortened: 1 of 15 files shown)


pyanaconda/ui/gui/spokes/storage.py
```diff
@@ -457,9 +466,8 @@ def _doExecute(self):
             StorageCheckHandler.errors = str(e).split("\n")
             hubQ.send_message(self.__class__.__name__, _("Failed to save storage configuration..."))
             self.data.bootloader.bootDrive = ""
-            self.data.ignoredisk.drives = []
```
rvykydal: Hm, seems like a noop with original intent of resetting ignoredisk.ignoredisk. Interesting, but nothing that should be handled in scope of this PR I think.

pyanaconda/ui/gui/spokes/storage.py
```diff
@@ -525,27 +533,20 @@ def showable(self):
...
         if threadMgr.get(constants.THREAD_DASDFMT):
-            msg = _("Formatting DASDs")
+            return _("Formatting DASDs")
         elif flags.automatedInstall and not self.storage.root_device:
-            msg = _("Kickstart insufficient")
-        elif self.data.ignoredisk.onlyuse:
-            msg = P_(("%d disk selected"),
-                     ("%d disks selected"),
-                     len(self.data.ignoredisk.onlyuse)) % len(self.data.ignoredisk.onlyuse)
-
```
rvykydal: This message is no more displayed intentionally I guess ?

pyanaconda/storage/osinstall.py
```diff
@@ -378,17 +380,15 @@ def __init__(self):
         # disklabels depends on this flag.
         self.clear_non_existent = False
 
-    def update(self, ksdata):
-        """ Update configuration from ksdata source.
+    def update(self, *args, **kwargs):
```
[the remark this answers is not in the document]
poncovka: I just didn't want this method to fail if it is called with `ksdata`. I can use `ksdata=None` instead, or remove all arguments.

pyanaconda/ui/gui/spokes/storage.py
poncovka: I can replace it with `self._disk_select_observer.proxy.SetIgnoredDisks([])` or just remove it. I don't want to keep the original line since UI shouldn't access the `ignoredisk` command anymore. However, I could at least add a comment here.

pyanaconda/ui/gui/spokes/storage.py
poncovka: This message was never returned. It was always replaced with one of the messages below.

pyanaconda/ui/gui/spokes/storage.py
rvykydal: I guess it is there for quite a long time, so let's just ignore it (ie I'm fine with the current patch).

pyanaconda/ui/gui/spokes/storage.py
rvykydal: Ah, I see..., okay.

poncovka:
Rebased.

PR #1327, opened 2018-02-13
Authconfig is replaced with authselect (#1542968)

poncovka:
Authconfig is deprecated, so anaconda will run authselect instead.
If a kickstart file specifies the authconfig command, anaconda will
run authconfig-compat with given authconfig arguments for the backward
compatibility.

The --passalgo and --enableshadow options are not supported by
authselect.

Resolves: rhbz#1542968

**Depends on:** 
https://github.com/pbrezina/authselect/pull/36
https://github.com/pbrezina/authselect/issues/30
https://github.com/clumens/pykickstart/pull/209

PR #1410, opened 2018-03-27
RFC: kickstart: only call authselect if explicitly requested

jlebon:
The system image might have a curated version of `nsswitch.conf` that we
don't want to overwrite. Rather than indeterministically calling it
based on whether `fprintd` is in the tree, let's just only call it if it
was explicitly asked for in the kickstart.

See also: https://github.com/pbrezina/authselect/issues/48.

### DIFF
--- a/pyanaconda/kickstart.py
+++ b/pyanaconda/kickstart.py
@@ -344,15 +344,8 @@ class Authselect(RemovedCommand):
     def execute(self, *args):
         security_proxy = SECURITY.get_proxy()
 
-        # Enable fingerprint option by default (#481273).
-        if not flags.automatedInstall and self.fingerprint_supported:
-            self._run(
-                "/usr/bin/authselect",
-                ["select", "sssd", "with-fingerprints", "--force"],
-                required=False
-            )
-
-        # Apply the authselect options from the kickstart file.
+        # Only call authselect if explicitly specified in kickstart.
+        # https://github.com/pbrezina/authselect/issues/48
         if security_proxy.Authselect:
             self._run(
                 "/usr/bin/authselect",


jlebon:
(Untested for now.)

jlebon:
Tested as fixing FAW 28 Beta.

cgwalters:
The whole situation where `authselect` is sort of designed for an admin to explicitly invoke, but sometimes we invoke it behind their back in mysterious ways feels pretty broken.

The code here was added for a reason though - you're basically reverting it.  And then people clicking through interactive workstation installs then won't get fprintd enabled.

What would seem a lot more sane to me is that if `fprintd` is installed it's enabled via some sort of drop-in automatically, no "authselect" required.  But authselect could be used to explicitly disable it if installed.

AdamWill:
This doesn't feel right from the PoV of actually doing what this code was originally intended for, which was: enabling fingerprint authentication by default. Heck, that's still written in a comment line in your code:

    # Enable fingerprint option by default (#481273).

but it doesn't actually *do* that any more, does it? It'd only do it if it happened to be using a kickstart with an `authconfig` line, so not "by default" at all. The conditional also looks to be clearly wrong in the new context - why would be be checking whether this is an 'automated' install or not, in that context? Basically, the whole "move that part into the `if security_proxy.Authconfig:` conditional" part of the PR looks entirely bogus to me.

In an ideal world I think I'd want to just drop this chunk entirely (i.e. just the first part of the PR) and figure out a better way to enable fingerprint auth by default, if we think we still want that. It seems weird in the first place that this was *ever* done by having anaconda run authconfig in the installed system root.

jlebon:
Right, this was meant as a partial revert of #1327. I.e. we still do run `authselect` if a line is present in the kickstart. Agreed that we shouldn't mysteriously enable fingerprinting if `authconfig` is present. I updated the patch to drop that part! :arrow_up:

jlebon:
Dropping this since we have a nicer workaround in https://pagure.io/workstation-ostree-config/pull-request/81. We can follow-up with the details of better fprintd integration with authselect somewhere else.

PR #1413, opened 2018-04-03
Create the bootloader module

poncovka:
Add the bootloader module that can handle the bootloader kickstart
command and provides an access to its options.

Add tests for the bootloader kickstart command and properties of
the bootloader module.


poncovka:
**Updated:** Rebased to master and added a commit with UI changes.

pyanaconda/installation.py
```diff
@@ -200,8 +207,9 @@ def doInstall(storage, payload, ksdata, instClass):
        The two main tasks for this are putting filesystems onto disks and
        installing packages onto those filesystems.
     """
-    willInstallBootloader = not flags.flags.dirInstall and (not ksdata.bootloader.disabled
-                                                            and ksdata.bootloader != "none")
+    bootloader_proxy = STORAGE.get_proxy(BOOTLOADER)
+    bootloader_enabled = bootloader_proxy.BootloaderMode != BOOTLOADER_DISABLED
+    can_install_bootloader = not flags.flags.dirInstall and bootloader_enabled
```
jkonecny12: I would abstract this to a function. It feels better to me when I call `can_install_bootloader` function than variable.

poncovka: I don't want to create the DBus proxy multiple times just for this. However, I can rename the variable to something else.

jkonecny12: That variable name is fine and if you don't want to change this. You can ignore my previous suggestion.

PR #1418, opened 2018-04-06
Add possibility to set default password function

jkonecny12:
This is required for initial-setup after Simpleline's input rework pull request.

**This requires https://github.com/rhinstaller/python-simpleline/pull/51 before merge!**

### DIFF
--- a/pyanaconda/ui/tui/tuiobject.py
+++ b/pyanaconda/ui/tui/tuiobject.py
@@ -28,6 +28,9 @@ from simpleline.render.adv_widgets import ErrorDialog, GetInputScreen, GetPasswo
 from simpleline.render.screen import UIScreen, Prompt
 from simpleline.render.screen_handler import ScreenHandler
 
+from pyanaconda.anaconda_loggers import get_module_logger
+log = get_module_logger(__name__)
+
 
 def reporting_callback(message):
     """Callback used for general reporting from acceptance conditions.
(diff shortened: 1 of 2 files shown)


pyanaconda/ui/tui/tuiobject.py
```diff
@@ -93,12 +93,26 @@ class TUIObject(UIScreen, common.UIObject):
...
+
+    @classmethod
+    def set_default_password_func(cls, password_func):
+        """Set password function which will be used for all screens.
+
+        :param password_func: Function with one parameter which is text prompt returning user
+                              input.
+        :type password_func: Password function fn(text_prompt : str) -> str .
+        """
+        cls._pass_func = password_func
```
poncovka: Maybe I would use a static method. If `set_default_password_func` is called for a different class then `TUIObject`, it will have no effect.

jkonecny12: Actually that can be even beneficial. That way you can define subset of UIScreens which will have different password function than others.

poncovka: But `self.password_func = TUIObject._pass_func` will not set the password function to `_pass_func` of the current class. So it doesn't make sense to support `set_default_password_func`  for different classes if it actually doesn't change the default password function at all.

jkonecny12: Ohh, true I've missed that. But I think better solution would be to change that assignment to be work in child methods.
If I'm not mistaken abstract that to the classmethod and call this method through `self` should be enough to have this working as expected.

poncovka: Ok, that should work. Maybe also rename `_pass_func` to `_default_password_func`, so it is clear what the attribute does.

jkonecny12: okay

jkonecny12:
**UPDATED:**
* Fix correct setting of password function.

jkonecny12:
Thanks to the https://github.com/rhinstaller/python-simpleline/pull/52 this is no longer required. Closing.

PR #1421, opened 2018-04-09
authselect: fix typo in fingerprint case and enable silent last log

pbrezina:
Parameter "with-fingerprints" does not exist, it spells without 's'
at the end, therefore fingrprint authentication does not work after
this call.

Second patch is related to [1]. Loud pam_lastlog is required for
RHEL but it is not desirable for Fedora, since it delays gdm login
and thus is should be silent by default on Fedora.

[1] https://bugzilla.redhat.com/show_bug.cgi?id=1561092


pbrezina:
`with-silent-lastlog` is available since `authselect-0.4`, which I just released in F28 and rawhide.
F28 update: https://bodhi.fedoraproject.org/updates/authselect-0.4-1.fc28

pyanaconda/kickstart.py
```diff
@@ -349,7 +349,7 @@ def execute(self, *args):
         if not flags.automatedInstall and self.fingerprint_supported:
             self._run(
                 "/usr/bin/authselect",
-                ["select", "sssd", "with-fingerprints", "--force"],
+                ["select", "sssd", "with-fingerprint", "with-silent-lastlog", "--force"],
```
poncovka: This seems to be safe. I suppose, that it is ok, that the option `with-silent-lastlog` will not be applied if `self.fingerprint_supported` is False.

poncovka: Please, add to the commit message:
`Related: rhbz#1561092`

pbrezina:
I added rhbz and opend f28 pull request: https://github.com/rhinstaller/anaconda/pull/1423

pyanaconda/kickstart.py
pbrezina: Since Anaconda only calls authselect automatically in this case and otherwise does not touch the configuration, it is ok.

PR #1423, opened 2018-04-10
authselect: fix typo in fingerprint case and enable silent last log

pbrezina:
Parameter "with-fingerprints" does not exist, it spells without 's'
at the end, therefore fingrprint authentication does not work after
this call.

Second patch is related to [1]. Loud pam_lastlog is required for
RHEL but it is not desirable for Fedora, since it delays gdm login
and thus is should be silent by default on Fedora.

[1] https://bugzilla.redhat.com/show_bug.cgi?id=1561092


poncovka:
Jenkins, test this please.

PR #1424, opened 2018-04-10
Don't try to create required partitions if there are none (#1557485)

poncovka:
We shouldn't try to create required partitions, if there are none,
otherwise Blivet will use partitions from autopartitioning.

Resolves: rhbz#1557485


poncovka:
Jenkins, test this please.

jkonecny12:
jenkins, test this please

PR #1427, opened 2018-04-10
Clear errors when a user updates partitioning settings (#1535781)

poncovka:
In the custom partitioning spoke, we should clear the displayed errors
when a user clicks on the Update Settings button.

Resolves: rhbz#1535781

### DIFF
--- a/pyanaconda/ui/gui/spokes/custom.py
+++ b/pyanaconda/ui/gui/spokes/custom.py
@@ -2628,6 +2628,10 @@ class CustomPartitioningSpoke(NormalSpoke, StorageCheckHandler):
     @timed_action(delay=50, threshold=100)
     def on_update_settings_clicked(self, button):
         """ call _save_right_side, then, perhaps, populate_right_side. """
+        if button:
+            # Clear errors if we are handling a click.
+            self.clear_errors()
+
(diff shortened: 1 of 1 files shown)


pyanaconda/ui/gui/spokes/custom.py
```diff
@@ -2628,6 +2628,10 @@ def on_info_bar_clicked(self, *args):
     @timed_action(delay=50, threshold=100)
     def on_update_settings_clicked(self, button):
         """ call _save_right_side, then, perhaps, populate_right_side. """
+        if button:
```
jkonecny12: In what case the button won't be present?

poncovka: Sometimes we call the method directly with `None` and it is difficult to test if it is ok to clear the errors at that points. This code is enough to fix the bug, but it shouldn't change anything else.

poncovka:
Jenkins, test this please.

pyanaconda/ui/gui/spokes/custom.py
jkonecny12: When this is targeted to RHEL, then it should be fine with this patch. However, backport of this patch should be reworked to remove calls without button to this method.

PR #1428, opened 2018-04-10
Add %files for install-env-deps so it actually exists

AdamWill:
The package won't actually be built if it lacks a %files. This
broke Rawhide composes.

### DIFF
--- a/anaconda.spec.in
+++ b/anaconda.spec.in
@@ -288,7 +288,10 @@ update-desktop-database &> /dev/null || :
 update-desktop-database &> /dev/null || :
 %endif
 
+# main package and install-env-deps are metapackages
 %files
+
+%files install-env-deps
 
 # Allow the lang file to be empty
 %define _empty_manifest_terminate_build 0


mohanboddu:
+1

AdamWill:
Note, I've done a .fc29.1 build for Rawhide so we can hopefully get a compose done. It's fine for the changes in that to get overwritten by a 'proper' .2 build with this merged.

AdamWill:
This is also needed on the f28 branch.

jkonecny12:
Please rebase this to f28-devel branch. The f28-devel branch will be merged to master.

M4rtinK:
I've cherry picked the fix to f28-devel and did a new F28 build - anaconda-28.22.6-1.fc28. For some reason I can't add it to the Bohi update though (https://bodhi.fedoraproject.org/updates/anaconda-28.22.4-1.fc28%20lorax-28.10-1.fc28#comment-764198).

In any case it would be good if the update could be tested and if working forwarded to stable as soon as possible. As far as I know pbrobinson & the ARM group overall is really looking forward to the depndency changes to land. There will also be an Anaconda test day on Tuesday, so it really should be in the composes by then so we can test it. Thanks in advance! :)

M4rtinK:
The cherry-picked fix has been deployed & this PR could thus be closed. Thanks yet again for the patch! :)

PR #1430, opened 2018-04-11
Fix DD test missing patch for subprocess.check_output

jkonecny12:
I'm not sure why this worked before and I don't see any related change in python.

Basically when our testing infrastructure moved to a new Jenkins instance this test started to fail.

So or so there shouldn't be subprocess call to the system in unit-tests. To fix this I'm adding patch for the subprocess.check_output call as it is for other tests here already.

------------------------------

I need to set a bug for this fix and link it before merge. This is more test if the issue is really fixed.


jkonecny12:
Fixed by jenkins settings. Closing.

PR #1433, opened 2018-04-12
Fix double logging to stdout

poncovka:
Loggers should have different names, otherwise we can set up the
same logger several times. This will fix the double logging to stdout.


poncovka:
Maybe @M4rtinK could look at it as well, since he wrote this part: https://github.com/rhinstaller/anaconda/commit/a7285b6

PR #1434, opened 2018-04-12
Support fcoe --autovlan option (#1564096)

rvykydal:
This is a backward kickstart compatibility patch, the default behavior is using
autovlan so the option does not have any effect.


rvykydal:
We should merge the patch only after new pykickstart (pykickstart-3.13-1) is built

PR #1435, opened 2018-04-12
rpmostreepayload: do not require network for dvd installation (#1565369)

rvykydal:


pyanaconda/payload/rpmostreepayload.py
```diff
@@ -64,6 +64,14 @@ def spaceRequired(self):
         # We don't have this data with OSTree at the moment
         return Size("500 MB")
 
+    @property
+    def needsNetwork(self):
+        """Test ostree repository if it requires network."""
+        if self.data.ostreesetup.url and self.data.ostreesetup.url.startswith("file://"):
```
cgwalters: Could just be:
`return not (self.data.ostreesetup.url and self.data.ostreesetup.url.startswith("file://"))`
but LGTM as is too!

AdamWill: this is how pylint likes it, IIRC, yes :)

AdamWill:
+1!

pyanaconda/payload/rpmostreepayload.py
rvykydal: I decided for if command because somehow it felt more readable to my weary brain, but here you are :)

PR #1440, opened 2018-04-16
Start only the specified kickstart modules (#1566621)

poncovka:
We need to be able to specify which kickstart modules should run,
so we can disable some of them in the initial setup.

Related: rhbz#1566621


M4rtinK:
jenkins, test this please

PR #1441, opened 2018-04-17
Fix the install class selection on live installations

poncovka:
We should choose an install class for the variant Workstation
by default on live installations.

### DIFF
--- a/data/liveinst/liveinst
+++ b/data/liveinst/liveinst
@@ -65,6 +65,7 @@ for i in raid0 raid1 raid5 raid6 raid456 raid10 dm-mod dm-zero dm-mirror dm-snap
 if [ -f /etc/system-release ]; then
     export ANACONDA_PRODUCTNAME=$( cat /etc/system-release | sed -r -e 's/ *release.*//' )
     export ANACONDA_PRODUCTVERSION=$( cat /etc/system-release | sed -r -e 's/^.* ([0-9\.]+).*$/\1/' )
+    export ANACONDA_PRODUCTVARIANT="Workstation"
 fi
 
 if [ $IMAGE_INSTALL = 1 ]; then


jkonecny12:
I don't see any bug reference so do you really want this to go to f28?

M4rtinK:
> I don't see any bug reference so do you really want this to go to f28?

We are counting on @AdamWill for that. ;-)

jkonecny12:
jenkins, test this please

AdamWill:
This does not look correct to me on the face of it. What about all the live images that *aren't* Workstation?

AdamWill:
The bug, btw, is https://bugzilla.redhat.com/show_bug.cgi?id=1569083 .

poncovka:
Added the bug reference.

poncovka:
@AdamWill, I don't think that Anaconda is able to get the variant name from the live system right now. The best solution would be to include the `.buildstamp` file, but that needs to be done probably in `livemedia-creator`.

poncovka:
Jenkins, test this please.

M4rtinK:
Adding @bcl (Lorax & LMC maintainer) and @dustymabe (releng) to CC.

poncovka:
Jenkins, test this please.

dustymabe:
hey @M4rtinK, I'm actually not releng (though I appreciate the vote of confidence :smile: ) Tagging @mohanboddu instead

mohanboddu:
I agree with @AdamWill here, what about all the other live images that aren't workstation and hard coding it seems like a bad idea.

AdamWill:
If you're OK with parsing `/etc/system-release` it's presumably OK to parse `/etc/os-release` too? In which case you can just read the `VARIANT_ID` out of that, I think.

PR #1443, opened 2018-04-17
Create the dynamic module User

poncovka:
**Rename the main module User to Users:**
The main module should be named Users, so we can later create the
dynamic modules named User for every user setting.

**Create the dynamic module User:**
The User module represents settings of one user account. The modules
are created dynamically with the CreateUser method or from kickstart.
They can be accessed directly from DBus.

**Support temporary kickstart generating:**
Sometimes we extend a DBus module with new kickstart commands,
but the kickstart commands are still handled by UI for a while. Then
we need to be able to generate a temporary kickstart representation,
that doesn't contain these unsupported commands, because they will
be generated by UI. This functionality is now provided by the method
GenerateTemporaryKickstart.


jkonecny12:
jenkins, test this please

poncovka:
Rebased.

PR #1449, opened 2018-04-20
Network module add connectivity state

rvykydal:


tests/nosetests/pyanaconda_tests/module_network_test.py
```diff
@@ -56,6 +56,14 @@ def get_current_hostname_test(self):
...
         self.network_interface.GetCurrentHostname()
 
+    def connected_test(self):
+        """Test getting connectivity status does not fail."""
+        self.network_interface.Connected
+
+    def connecting_test(self):
+        """Test checking connecting status does not fail."""
+        self.network_interface.IsConnecting()
+
```
poncovka: You can probably extend the tests a little if you mock the `nm_client`.

pyanaconda/modules/network/network.py
```diff
@@ -100,3 +117,35 @@ def set_current_hostname(self, hostname):
 
         self._hostname_service_proxy.SetHostname(hostname, False)
         log.debug("Current hostname is set to %s", hostname)
+
+    @property
+    def connected(self):
+        """Is the system connected to the network?"""
+        if self.nm_client:
```
jkonecny12: I personally think this will deserve abstraction. It would be nice to call `self.nm_available` instead.

tests/nosetests/pyanaconda_tests/module_network_test.py
rvykydal: I can't figure out how to mock nm_client reasonbly with current code, I've only added one silly test. In the future when I start to use the client heavily for real work with connections and ifcfgs, I'll probably encapsulate it into a separate module so maybe then it will be easier. For now the client (and NM / libnm) seems to be too bound to the network module.

pyanaconda/modules/network/network.py
rvykydal: Added.

tests/nosetests/pyanaconda_tests/module_network_test.py
rvykydal: Added a test with mocked NMClient

rvykydal:
Also added module_properties_changed.emit() to set_connected

PR #1450, opened 2018-04-20
Rhel7 fix 1360223

jkonecny12:
These commits were lost during the rebase.

These will be backported to master too.


jkonecny12:
**UPDATED:**
* Fix style guide.

PR #1456, opened 2018-04-23
Show correct root spoke lock status in reconfig mode (#1507940)

M4rtinK:
It is currently possible (and expected on ARM images) to
re-enable a locked root account by setting a new root password
if the root password spoke is running in Initial Setup reconfig mode.

Unfortunately the spoke status in this case just says the password
is set without mentioning that the root account is locked.

So fix the root password spoke status to mention that the root account is locked
but can be re-enabled by setting a new root password.

Resolves: rhbz#1507940

### DIFF
--- a/pyanaconda/ui/gui/spokes/root_password.py
+++ b/pyanaconda/ui/gui/spokes/root_password.py
@@ -22,7 +22,7 @@ from pyanaconda.core.i18n import _, CN_
 from pyanaconda.users import cryptPassword
 from pyanaconda import input_checking
 from pyanaconda.core import constants
-from pyanaconda.modules.common.constants.services import USER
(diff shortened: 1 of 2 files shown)


M4rtinK:
> Makes me think if modules should provide the (localized) status message (eg for network tui and gui currently share status_message() from network.py). But it is probably something we want to discuss and figure out later and for all modules. And here it is also complicated by the mode which we may want to figure out for modules in a generic way (so later) as well.

I think we should avoid modules providing user visible strings if possible - ideally we would not need to handle localization at all for modules & all string localization would happen in the UI layer only.

As for the GUI/TUI strings - actually, they are different, most likely because TUI supports multi-line status messages while GUI is one line only and string that are too long get "compressed" in a weird way.

PR #1458, opened 2018-04-23
Create the auto partitioning module

poncovka:
he auto partitioning module can handle the autopart kickstart command
and provides API for its options. The module is static for now, but it
will be dynamic in the future.

Add new tests for the auto partitioning module and the autopart
command.


pyanaconda/modules/storage/partitioning/automatic.py
```diff
@@ -0,0 +1,268 @@
...
+        :param cipher: a name of an algorithm
+        """
+        self._cipher = cipher
+        self.cipher_changed.emit()
+        log.debug("Cipher is set to '%s'.", cipher)
+
+    @property
+    def passphrase(self):
+        """Default passphrase for all encrypted devices."""
+        return self._passphrase
```
jkonecny12: This just feels weird to me. I know we need it now, but we shouldn't provide password back from module. I don't think we will really need that when this spoke is finished. 

So could you please add TODO here to remove getter when module will be completed please?

pyanaconda/modules/storage/partitioning/automatic_interface.py
```diff
@@ -0,0 +1,186 @@
...
+    @property
+    def Enabled(self) -> Bool:
+        """Is the auto partitioning enabled?"""
+        return self.implementation.enabled
+
+    @emits_properties_changed
+    def SetEnabled(self, enabled: Bool):
+        """Is the auto partitioning enabled?
+
+        :param enabled: a boolean value
```
rvykydal: I think this can go away (or tell about semantics, not type).

pyanaconda/modules/storage/partitioning/automatic.py
M4rtinK: Yeah, something similar to the note in the user module could be used:
https://github.com/rhinstaller/anaconda/blob/master/pyanaconda/modules/user/user_interface.py#L66

One more reason to concentrate on moving the logic on these modules as soon as possible. :)

poncovka: Fixed in the DBus interface.

pyanaconda/modules/storage/partitioning/automatic_interface.py
poncovka: Fixed.

poncovka:
Fixed some small issues.

PR #1462, opened 2018-04-26
Fix crash on inst.noverifyssl with non-url method

rvykydal:
Resolves: rhbz#1515678

### DIFF
--- a/anaconda
+++ b/anaconda
@@ -1139,7 +1139,8 @@ if __name__ == "__main__":
             iutil.setenv("ftp_proxy", proxy.url)
             iutil.setenv("HTTPS_PROXY", proxy.url)
 
-    if flags.noverifyssl:
+    # only some installations methods support --noverifyssl option
+    if flags.noverifyssl and hasattr(ksdata.method, "noverifyssl"):
         ksdata.method.noverifyssl = flags.noverifyssl
     if opts.multiLib:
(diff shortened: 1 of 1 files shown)


anaconda
```diff
@@ -1139,7 +1139,7 @@ if __name__ == "__main__":
             iutil.setenv("ftp_proxy", proxy.url)
             iutil.setenv("HTTPS_PROXY", proxy.url)
 
-    if flags.noverifyssl:
+    if flags.noverifyssl and ksdata.method.method == "url":
         ksdata.method.noverifyssl = flags.noverifyssl
```
poncovka: We should probably check `hasattr(ksdata.method, "noverifyssl")`, because there can be several methods with this attribute (for example liveimg).

rvykydal: You are right, fixed, thank you.

anaconda
```diff
@@ -1139,7 +1139,7 @@ if __name__ == "__main__":
             iutil.setenv("ftp_proxy", proxy.url)
             iutil.setenv("HTTPS_PROXY", proxy.url)
 
-    if flags.noverifyssl:
+    if flags.noverifyssl and hasattr(ksdata.method, "noverifyssl"):
```
jkonecny12: Please also add comment why the `noverifyssl` may not be present.

rvykydal: Done

PR #1464, opened 2018-04-27
Help placeholders (#1551929)

M4rtinK:
Actual help content is not yet available, so just always
show a placeholder for now.

Resolves: rhbz#1551929


M4rtinK:
TODO:
- testing in GUI
- testing in TUI
- test packaging is correct

PR #1465, opened 2018-04-27
Update the pykickstart commands

poncovka:
The bootloader, ignoredisk and partition commands have some options
deprecated in F29. The deprecated command upgrade is removed in F29.

Related: https://github.com/clumens/pykickstart/pull/217
**Depends on: pykickstart-3.13**


poncovka:
Rebased.

PR #1468, opened 2018-05-03
Ask for a default passphrase if required (#1436304)

poncovka:
If the kickstart file requires a default LUKS passphrase to finish the partitioning, ask a user for one.

Resolves: rhbz#1436304

### DIFF
--- a/pyanaconda/ui/tui/spokes/__init__.py
+++ b/pyanaconda/ui/tui/spokes/__init__.py
@@ -26,7 +26,7 @@ import re
 from collections import namedtuple
 from pyanaconda.iutil import setdeepattr, getdeepattr
 from pyanaconda.i18n import N_, _
-from pyanaconda.constants import PASSWORD_CONFIRM_ERROR_TUI, PW_ASCII_CHARS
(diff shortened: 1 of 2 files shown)


pyanaconda/ui/tui/spokes/__init__.py
```diff
@@ -135,20 +145,21 @@ def prompt(self, entry = None):
             pw = self._app.raw_input(_("%s: ") % entry.title, hidden=True)
             confirm = self._app.raw_input(_("%s (confirm): ") % entry.title, hidden=True)
 
-            if (pw and not confirm) or (confirm and not pw):
-                print(_("You must enter your root password and confirm it by typing"
-                        " it a second time to continue."))
-                return None
```
jkonecny12: Are you sure that we want to loose this message?

M4rtinK: I would also expect it to be generally useful - unless the code is unreachable or something similar (that might be possible after the substantial changes we did on password checking a while back).

poncovka: If we want to keep this message, then I will have to change it anyway, because "root password" is hard coded, so for user passwords and passphrases is the text the same.

Now you will see a message: "The passwords you entered were different.  Please try again." I think that is clear what was wrong and what the user should do.

PR #1471, opened 2018-05-09
Only check space during a tui kickstart if ksprompt is enabled

bcl:
The problem is that if it isn't enabled it hits the check, which guesses
at the size (3000 MB) since depsolving isn't done yet and it crashes
(which is a separate problem).

When doing unattended ks installs it doesn't make sense to stop at this
point, the size is re-evaluated later, when the depsolve is finished and
it has the correct value to check against.

### DIFF
--- a/pyanaconda/ui/tui/hubs/summary.py
+++ b/pyanaconda/ui/tui/hubs/summary.py
@@ -92,7 +92,7 @@ class SummaryHub(TUIHub):
         if flags.automatedInstall and not incomplete_spokes:
 
             # Check the available space.
-            if self._checker and not self._checker.check():
+            if flags.ksprompt and self._checker and not self._checker.check():
 
                 # Space is not ok.
                 print(self._checker.error_message)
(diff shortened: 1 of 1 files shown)


bcl:
I'm pretty sure this patch is correct, but I have only tested it with --dirinstall from inside lorax-composer.

jkonecny12:
Hi @bcl, do you want this to go to Fedora 28?

jkonecny12:
OK, so then please rebase this to `modularization-devel` branch and create a new PR to `f28-release` branch when you will be ready for backport.

jkonecny12:
Hi @bcl, could you please rebase this to the `modularization-branch`?

jkonecny12:
Hi, @bcl is this still something we want to have? If yes, do you want to rebase it or should we re-take that?

bcl:
> Hi, @bcl is this still something we want to have? If yes, do you want to rebase it or should we re-take that?

I'd totally forgotten about this. lorax-composer is now gone, so that use case isn't valid anymore, but you can probably hit it with livemedia-creator. I'll try to reproduce it with the current release and see if it's a problem using lmc.

PR #1473, opened 2018-05-14
Fix firewall DBUS module API usage (#1577405)

M4rtinK:
The is-firewall-enabled check was using an incorrect
API of the Firewall DBUS module, use the correct one.
    
Resolves: rhbz#1577405


pyanaconda/kickstart.py
```diff
@@ -929,9 +929,8 @@ def execute(self, storage, ksdata, instClass):
 
         cmd = "/usr/bin/firewall-offline-cmd"
         if not os.path.exists(util.getSysroot() + cmd):
-            if firewall_proxy.FirewallEnabled:
-                msg = _("%s is missing. Cannot setup firewall.") % (cmd,)
-                raise KickstartError(msg)
+            msg = _("%s is missing. Cannot setup firewall.") % (cmd,)
+            raise KickstartError(msg)
```
poncovka: This error will abort the installation, no? Do we really want to do this if the firewall mode is disabled or default?

M4rtinK: Good point! I should have looked better and noticed it is not just not a simple log message. Code blindness I guess. :)

So I'll just put the check back with correct API usage.

PR #1475, opened 2018-05-14
Mod devel no ks check

bcl:


M4rtinK:
Is this storage space check backport for F28 ? I think something terrible happened with branching there. :)

bcl:
> Is this storage space check backport for F28 ? I think something terrible happened with branching there. :)

yeah, that's what it was supposed to be. I figured you just had a giant stack of pending stuff. How about I do this again...

PR #1477, opened 2018-05-15
Use the auto partitioning module in UI

poncovka:
Don't use the kickstart command autopart in UI anymore.
Use the auto partitioning module instead.

The autopart command was never part of the generated kickstart file,
because the mount command always disabled the autopartitioning flag.
The flag should be disabled in the mount data instead.


pyanaconda/ui/gui/spokes/storage.py
```diff
@@ -492,7 +497,7 @@ def _doExecute(self):
             hubQ.send_message(self.__class__.__name__, _("Unexpected storage error"))
             raise e
         else:
-            if self.autopart or (flags.automatedInstall and (self.data.autopart.autopart or self.data.partition.seen)):
+            if self.autopart or (flags.automatedInstall and (self._auto_part_observer.proxy.Enabled or self.data.partition.seen)):
```
jkonecny12: Please wrap this line.

poncovka: Fixed.

PR #1479, opened 2018-05-16
Add 10% for storage metadata to the total required space (#1578395)

poncovka:
In the Install Options dialog, we should add 10% of the disk space for
storage metadata to the total required space. Otherwise, the space
requirements might be confusing.

Resolves: rhbz#1578395

### DIFF
--- a/pyanaconda/ui/gui/spokes/storage.py
+++ b/pyanaconda/ui/gui/spokes/storage.py
@@ -71,7 +71,7 @@ from pyanaconda.flags import flags
 from pyanaconda.core.i18n import _, C_, CN_, P_
 from pyanaconda.core import util, constants
 from pyanaconda.core.constants import CLEAR_PARTITIONS_NONE, BOOTLOADER_DRIVE_UNSET, \
-    BOOTLOADER_ENABLED, AUTOPART_TYPE_DEFAULT
+    BOOTLOADER_ENABLED, STORAGE_METADATA_RATIO, AUTOPART_TYPE_DEFAULT
 from pyanaconda.bootloader import BootLoaderError
 from pyanaconda.storage import autopart
 from pyanaconda.storage_utils import on_disk_storage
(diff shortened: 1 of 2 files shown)


poncovka:
Now we will show a required size that is higher than the available space.

screenshot_vpn_test_02_2018-05-16_16 03 30

pyanaconda/ui/gui/spokes/storage.py
```diff
@@ -139,7 +139,7 @@ def _get_sw_needs_text(self, required_space, auto_swap):
                          "space, including <b>%(software)s</b> for software and "
                          "<b>%(swap)s</b> for swap space.")
                        % {"product": escape_markup(productName),
-                          "total": escape_markup(str(required_space + auto_swap)),
+                          "total": escape_markup(str(required_space + auto_swap + 0.1 * disk_free)),
```
jkonecny12: Why 0.1 constant? Could you please make the reason more obvious? Ideally make it `METADATA_SIZE` constant.

pyanaconda/ui/gui/spokes/storage.py
```diff
@@ -974,10 +974,10 @@ def _check_space_and_get_dialog(self, disks):
         log.debug("disk free: %s  fs free: %s  sw needs: %s  auto swap: %s",
                   disk_free, fs_free, required_space, auto_swap)
 
-        # compare only to 90% of disk space because fs takes some space for a metadata
-        if (disk_free * 0.9) >= required_space + auto_swap:
+        # We need enough space for the software, the swap and 10% of the disk space for metadata.
+        if disk_free >= required_space + auto_swap + 0.1 * disk_free:
```
jkonecny12: OK we are taking swap into account but what about `/boot`, `/home` and similar?

pyanaconda/ui/gui/spokes/storage.py
poncovka: Added a constant.

pyanaconda/ui/gui/spokes/storage.py
poncovka: Added a comment.

poncovka:
Updated.

PR #1484, opened 2018-05-21
Set locale to en_US.UTF-8 in every module (#1575415)

poncovka:
We should set locale to en_US.UTF-8 in every Anaconda DBus module,
otherwise the modules might not use the UTF-8 encoding by default.

Resolves: rhbz#1575415


poncovka:
@jkonecny12, I have found a way how to set the locale in lorax, but it doesn't help in our case. The problem was that we started the DBus daemon with `reset_lang=True` and set `LC_ALL=C`. I have fixed that in a new commit.

With this fix, it might seem redundant to set the locale in every module, but it is a recommended practice to always call `locale.setlocale` at the beginning of the application (https://docs.python.org/3.6/library/locale.html#locale.setlocale). Therefore, I would like to keep it.

Anyway, should I still fix the default locale in lorax?

PR #1485, opened 2018-05-23
Show a note about EULA at the end of the installation

M4rtinK:
Show a note about EULA at the end of the installation
in GUI & TUI.

Related: rhbz#1569596


M4rtinK:
Thanks for review!

PR #1486, opened 2018-05-24
Use subprocess32 to prevent concurrency issues.

rvykydal:
Resolves: rhbz#1581200

Related bugs: 1411407, 1268226

### DIFF
--- a/pyanaconda/iutil.py
+++ b/pyanaconda/iutil.py
@@ -25,7 +25,6 @@ import os
 import stat
 import os.path
 import errno
-import subprocess
 import unicodedata
 import shutil
 import string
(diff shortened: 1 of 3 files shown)


pyanaconda/iutil.py
```diff
@@ -38,6 +37,12 @@
 import random
 import functools
 from distutils import spawn
+import sys
+if os.name == 'posix' and sys.version_info[0] < 3:
+    import subprocess32 as subprocess
+else:
+    import subprocess
```
jkonecny12: Is this really required when you have subprocess32 as spec file dependency?

rvykydal:
jenkins, test this please

pyanaconda/iutil.py
rvykydal: I think we don't. Removed

PR #1487, opened 2018-05-25
Add hook to prevent mistake upstream pushes

jkonecny12:
This pre-push script will not allow push to the upstream branch. So if you need to push to the upstream then disable this check with --no-verify parameter.


scripts/githooks/pre-push-prevent-upstream-push
```diff
@@ -0,0 +1,29 @@
...
+url="$2"
+
+prevent_remote=$(git remote -v | grep rhinstaller | head -n 1 | cut -f 1)
+
+IFS=' '
+
+if [ "$remote" = "$prevent_remote" ]; then
+	echo "Do not push to remote branch"
+    echo "If you really want to push, use --no-verify to skip this check"
+	exit 1
```
poncovka: Please, fix the code alignment.

scripts/githooks/pre-push-prevent-upstream-push
```diff
@@ -0,0 +1,29 @@
...
+# This hook is called with the following parameters:
+#
+# $1 -- Name of the remote to which the push is being done
+# $2 -- URL to which the push is being done
+#
+
+remote="$1"
+url="$2"
+
+prevent_remote=$(git remote -v | grep rhinstaller | head -n 1 | cut -f 1)
```
poncovka: There should be probably quotation marks: `prevent_remote="$(...)"`

jkonecny12: it works even without them so they are not technically required but yeah, it won't hurt to add them

jkonecny12:
**UPDATED**

jkonecny12:
jenkins, test this please

PR #1489, opened 2018-05-29
Fix pylint errors

poncovka:
A new version of pylint was released.

PR #1490, opened 2018-05-29
Disable pylint run on bootloader.py

jkonecny12:
Python astroid 2.0.0 raising an exception on bootloader.py and because it doesn't change often we will disable the pylint check completely. 
This file should be enabled again when python-astroid will be fixed.


jkonecny12:
This PR and the second commit in the PR https://github.com/rhinstaller/anaconda/pull/1489 are fixing the same issue.

**You should decide destiny of Anaconda** -- In other words what PR should be used :D

jkonecny12:
The second PR won this fight. Closing this without merge.

PR #1496, opened 2018-06-04
Do not manually create LUKSDevice when unlocking a LUKS format

vojtechtrefny:
`populate` will create the device and add it to devicetree
automatically.
This also solves issues with LUKS 2 devices with integrity where
the LUKS device is not direct child of the partition that is
being unlocked. (So it is not possible to manually create it from
Anaconda and manually set its parents.)


vojtechtrefny:
> Should it be fixed as well?

Yes, thank you. Fixed.

poncovka:
Jenkins, test this please.

PR #1498, opened 2018-06-06
Wait for kickstart modules to quit

poncovka:
The boss should wait for the kickstart modules to quit before it will
quit itself. Otherwise, the modules might send a response while the
DBus connection is closed.

### DIFF
--- a/pyanaconda/modules/boss/module_manager.py
+++ b/pyanaconda/modules/boss/module_manager.py
@@ -114,10 +114,7 @@ class ModuleManager(object):
             if not observer.is_service_available:
                 continue
 
-            # Call asynchronously to avoid problems.
-            observer.proxy.Quit(callback=self._stop_modules_callback,
-                                callback_args=(observer,))
-
-    def _stop_modules_callback(self, observer, returned, error):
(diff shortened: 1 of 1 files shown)


M4rtinK:
That's a good question - ideally Boss should tell all the modules to quit and then wait (with a timeout) for them to quit (basically like join() on threads in Python, before it shuts down everything.

On the other hand I would be fine even with synchronous shutdown for the time being, as while fat, it's the correct way and the unclean module shutdown will only become a worse problem as modules start to process data, store caches, hold resources, etc.

Also, do we have any estimate how long the synchronous shutdown takes ? I don't think it would take so long (given that we have a 10 minute timeout for module startup...).

poncovka:
We could stop the modules the same way we start them. We could call QuitModules, the boss will asynchronously call Quit for each one of them and return, we will wait for the modules to become unavailable and we will call Quit to quit the boss.

If we want the boss to asynchronously call Quit and wait for the modules, then we need to create a task for that, because the boss is not able to process the callbacks inside of its Quit method.

I think that the synchronized quitting of modules is fine for now and we can change it later if necessary. It takes only seconds.

PR #1499, opened 2018-06-06
Use wwn attr instead of removed wwid. (#1565693)

dwlehman:
The 'wwid' attribute was removed in blivet-3.1.0. Now all disk-like
devices have a 'wwn' attribute that gets its value from udev's ID_WWN
property.

### DIFF
--- a/pyanaconda/ui/gui/spokes/storage.py
+++ b/pyanaconda/ui/gui/spokes/storage.py
@@ -742,7 +742,7 @@ class StorageSpoke(NormalSpoke, StorageCheckHandler):
         # We don't want to display the whole huge WWID for a multipath device.
         # That makes the DO way too wide.
         if isinstance(disk, MultipathDevice):
-            desc = disk.wwid.split(":")
+            desc = disk.wwn.split(":")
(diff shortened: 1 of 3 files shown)


pyanaconda/ui/gui/spokes/storage.py
```diff
@@ -742,7 +742,7 @@ def _add_disk_overview(self, disk, box):
         # We don't want to display the whole huge WWID for a multipath device.
         # That makes the DO way too wide.
         if isinstance(disk, MultipathDevice):
-            desc = disk.wwid.split(":")
+            desc = disk.wwn.split(":")
             description = ":".join(desc[0:3]) + "..." + ":".join(desc[-4:])
```
poncovka: Do wwn and wwid have the same value or the format of the value? Otherwise, the parsing might not work.

dwlehman: Good catch. The formatting is not identical.

poncovka:
Please, rebase the pull request to the `modularization-devel` branch.

dwlehman:
New pull request fixing the display formatting issue and rebased onto modularity-devel in #1503.

PR #1500, opened 2018-06-07
Fix the install class selection for Atomic (#1588389)

poncovka:
There are two variants of Atomic now. They are called AtomicHost nad
AtomicWorkstation. We should choose the Atomic install class for
both of them.

Resolves: rhbz#1588389


poncovka:
Could @dustymabe confirm that `AtomicHostInstallClass` should be used for `AtomicHost` and `AtomicWorkstation` variants? Otherwise, I can create new install classes.

dustymabe:
> Could @dustymabe confirm that AtomicHostInstallClass should be used for AtomicHost and AtomicWorkstation variants? Otherwise, I can create new install classes.

Preferably a new install class. Atomic Host and Atomic Workstation are different.

dustymabe:
also relevant is the "rebranding" of Atomic Workstation to Silverblue:  https://pagure.io/teamsilverblue/issue/20

cgwalters:
Right.  We do want a different install class eventually - specifically to switch to XFS (vs ext4) by default.  I'm not sure about going with the Server/AH partitioning for workstation systems though.  It'd force the "where'd my disk space go" and "how do I manage LVM" problems to the front, which is fine for servers, but I feel like would need some GUI for desktops.

cgwalters:
And not to mention the whole merger with CoreOS and how that's going to work that's not totally resolved yet.  So...I think my vote is "no action" here for now.

PR #1503, opened 2018-06-08
Use wwn attr instead of removed wwid. (#1565693)

dwlehman:
The 'wwid' attribute was removed in blivet-3.1.0. Now all disk-like
devices have a 'wwn' attribute that gets its value from udev's
ID_WWN_WITH_EXTENSION property.

This supercedes #1499 and also fixes the display formatting issue contained therein.

### DIFF
--- a/pyanaconda/ui/gui/spokes/advanced_storage.py
+++ b/pyanaconda/ui/gui/spokes/advanced_storage.py
@@ -240,7 +240,7 @@ class SearchPage(FilterPage):
         elif filterBy == self.SEARCH_TYPE_PORT_TARGET_LUN:
             return self._port_equal(device) and self._target_equal(device) and self._lun_equal(device)
         elif filterBy == self.SEARCH_TYPE_WWID:
-            return self._wwidEntry.get_text() in getattr(device, "wwid", self._long_identifier(device))
+            return self._wwidEntry.get_text() in getattr(device, "wwn", self._long_identifier(device))
 
     def visible_func(self, model, itr, *args):
         obj = DiskStoreRow(*model[itr])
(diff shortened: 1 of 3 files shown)


jkonecny12:
jenkins, test this please

jkonecny12:
It looks like it needs a little bit more polishing.
iscsi-multipath-bug

jkonecny12:
Also I'm not sure if this is happening because of my settings but when I'm trying to attach 1 disk from 2 different ip addresses there is no wwn value showed. Do you know why it is not there?

Also I think we need to change tab name to `WWN` when you look on that screenshot there is `WWID`.

jkonecny12:
This looks correct, however it looks that there is a bug in blivet. Based on my tests the wwn attribute is empty for multipath device.

New wwn:
screen_bad_wwn

Old wwid:
screen_good_wwn


I was looking into that in debugger and the `disk.wwn` is really empty. This is causing crash when I select mpath device and clicking on back button.

crash


Another question is if the `WWID` column title in the table shouldn't change to `WWN` too?

dwlehman:
I'm planning to build blivet-3.1.0.b2 (storaged-project/blivet@cd698c04) on Tuesday (tomorrow) for rawhide. I don't have firm plans for any other builds as of now.

jkonecny12:
OK thanks, I will test and merge this tomorrow.

rvykydal:
I've tested the PR with latest rawhide Fedora-Rawhide-20180802.n.0 and it seem to be good.
Once the patches mentioned above for blivet (https://github.com/storaged-project/blivet/pull/701) are built for rhel we can merge the PR.
wwn

PR #1509, opened 2018-06-12
passing a non existing disk to ignoredisk should not halt the installation

CaViCcHi:
I propose this small change to avoid an installation from exploding simply because a disk to ignore doesn't exist. 
Since it allows globs users can specify ignoredisk --drives=/dev/disk/by-path/\*usb\* if I wish to never install on a usb disk or something similar. But it shouldn't fail since I'm only excluding and if it doesn't exist, it should be ok.

### DIFF
--- a/pyanaconda/kickstart.py
+++ b/pyanaconda/kickstart.py
@@ -801,8 +801,7 @@ class IgnoreDisk(commands.ignoredisk.RHEL6_IgnoreDisk):
             if matched:
                 drives.extend(matched)
             else:
-                raise KickstartValueError(formatErrorMsg(self.lineno,
-                        msg=_("Disk \"%s\" given in ignoredisk command does not exist.") % spec))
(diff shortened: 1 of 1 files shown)


CaViCcHi:
I noticed it's a thing for the install to favor any usb media inserted as /boot and this was a way to prevent that. I can see how people could make mistakes, also getting a usb wiped in the process can't be good :)

M4rtinK:
Ignoring non existing disk specs/globs by default indeed seems to dangerous to me as well. A flag for the command that the user needs to supply to explicitly opt in for this behavior might work though.

CaViCcHi:
I can see how that would make it safer. I'll fiddle with a --nonfatal flag

CaViCcHi:
Reopened in 2 parts:
https://github.com/rhinstaller/anaconda/pull/1511 (warn if --nonfatal or fail)
https://github.com/clumens/pykickstart/pull/229 (--nonfatal parameter added to ignoredisk)

PR #1511, opened 2018-06-18
Introducing --nonfatal keeps the installation going on non-existent ignoredisk --drives=...

CaViCcHi:
POST conversation on https://github.com/rhinstaller/anaconda/pull/1509/commits/1dd12e8182eacc277f1577ebee9454e05f2d81a5
Part 1 of Pull request inside https://github.com/clumens/pykickstart/pull/229

I apologize if this is not the correct way to make a pull request on 2 different projects.

### DIFF
--- a/pyanaconda/kickstart.py
+++ b/pyanaconda/kickstart.py
@@ -801,7 +801,11 @@ class IgnoreDisk(commands.ignoredisk.RHEL6_IgnoreDisk):
             if matched:
                 drives.extend(matched)
             else:
-                log.warn("Disk \"%s\" given in ignoredisk command does not exist.", spec)
+                if self.nonfatal:
+                    log.warn("Disk \"%s\" given in ignoredisk command does not exist, --nonfatal", spec)
+                else:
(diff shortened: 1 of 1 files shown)


jkonecny12:
Hi, this is an interesting change. I think we can go with this. Of course others need to look on this too and it needs to be accepted to pykickstart first.
However, you need to do some changes before it can be merged.

* You have to have bug in bugzilla.redhat.com for RHEL 7 to be able to get something there. 
* You have to add `Resolves: rhbz#1581200` or `Related: rhbz#1534833`  in the commit message for every commit with bug number of your bug.
* Ideally you should make a PR to Fedora first (right now you should make PR for `modularization-devel` branch)
    * You may wait with this till the decision to accept this to RHEL 7 will be here. To avoid unnecessary work.

VladimirSlavik:
We are now well past the point where changes to RHEL 7 could be accepted. So is this still relevant @CaViCcHi ? As @jkonecny12 wrote, an upstream patch for Fedora would be a better starting point. Far less to discuss that way :) On the other hand, if you don't care any longer, maybe let's close this?

PR #1516, opened 2018-06-27
Add bootloader menu auto hide support

jwrdegoede:
These 2 commits implement:

1) Tracking for which installclasses the bootloader should (auto) hide its bootmenu by defaults; and
2) Actually enabling the menu_auto_hide feature for GRUB2 when the installclass indicates it should be enabled.

Together these implement the anaconda part of:
https://fedoraproject.org/wiki/Changes/HiddenGrubMenu

### DIFF
--- a/pyanaconda/bootloader.py
+++ b/pyanaconda/bootloader.py
@@ -212,6 +212,7 @@ class BootLoader(object):
     config_file_mode = 0o600
     can_dual_boot = False
     can_update = False
+    menu_auto_hide = False
     image_label_attr = "label"
 
     encryption_support = False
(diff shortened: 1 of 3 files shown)


pyanaconda/bootloader.py
```diff
@@ -1614,6 +1615,14 @@ def write_config(self):
             if rc:
                 log.error("failed to set default menu entry to %s", productName)
 
+        # set menu_auto_hide grubenv variable if we should enable menu_auto_hide
+        # set boot_success so that the menu is hidden on the boot after install
+        if self.menu_auto_hide:
+            rc = util.execInSysroot("grub2-editenv",
+                            ["-", "set", "menu_auto_hide=1", "boot_success=1"])
+            if rc:
+                log.error("failed to set menu_auto_hide=1")
```
jkonecny12: Also there is a question if this should be fatal error. In other words if we shouldn't raise an exception?

jwrdegoede: If this fails everything will still work fine, except that the grub2 bootmenu will show everyboot, as it was in F28 and before. So this is not really fatal. I based not making this fatal on the code which selects the default boot menu entry also simply logging an error, rather then throwing an exception.

jwrdegoede:
I cannot find a way to change the base-branch of a pull-req, so I will submit a new pull-req based on modularization-devel.

PR #1519, opened 2018-06-28
dnfpayload: repo in _fetch_md is a dnf repo not ksrepo

AdamWill:
In today's Rawhide compose, with DNF 3, the log message logged
by `dnfpayload._fetch_md` - "enabled repo: ..." - was slightly
broken in all the openQA tests that try to use it to verify a
repo was configured correctly. The message is supposed to
include an identifier for the repo ('anaconda', when the repo
in question is the base repo) but in these tests, the identifier
was blank, like this:

"enabled repo: '' - ..."

Looking into it a bit, I noticed that `_fetch_md` calls the repo
object it gets `ksrepo` and uses `ksrepo.name` as the identifier
string, but this is inconsistent with the rest of dnfpayload.
Elsewhere, we consistently use `ksrepo` as a name for instances
of the pykickstart RepoData classes, and `repo` as a name for
DNF Repo class instances. We use `ksrepo.name` as a string id
for RepoData instances, but `repo.id` as a string id for DNF
Repo instances.

The repo object in `_fetch_md` is clearly a DNF Repo class
instance (we get it from `self._base.repos`), but it's called
`ksrepo`, and we use `ksrepo.name` as the string identifier. I
think that *sometimes* DNF repo instances will have a working
`name` attribute but sometimes it'll just be empty, which is
when this bug happens. To be consistent with the rest of this
file and to avoid the bug, we should call the object `repo` and
use the `id` attribute, not the `name` attribute, as the string
identifier.

### DIFF
--- a/pyanaconda/payload/dnfpayload.py
+++ b/pyanaconda/payload/dnfpayload.py
@@ -410,26 +410,26 @@ class DNFPayload(payload.PackagePayload):
 
         log.info("added repo: '%s' - %s", ksrepo.name, url or mirrorlist or metalink)
 
-    def _fetch_md(self, repo):
+    def _fetch_md(self, repo_name):
         """Download repo metadata
 
-        :param repo: name/id of repo to fetch
-        :type repo: str
+        :param repo_name: name/id of repo to fetch
+        :type repo_name: str
         :returns: None
         """
-        ksrepo = self._base.repos[repo]
-        ksrepo.enable()
+        repo = self._base.repos[repo_name]
+        repo.enable()
         try:
             # Load the metadata to verify that the repo is valid
-            ksrepo.load()
+            repo.load()
         except dnf.exceptions.RepoError as e:
-            ksrepo.disable()
-            log.debug("repo: '%s' - %s failed to load repomd", ksrepo.name,
-                      ksrepo.baseurl or ksrepo.mirrorlist or ksrepo.metalink)
+            repo.disable()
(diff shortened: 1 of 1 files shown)


pyanaconda/payload/dnfpayload.py
```diff
@@ -393,26 +393,26 @@ def _add_repo(self, ksrepo):
 
         log.debug("added repo: '%s' - %s", ksrepo.name, url or mirrorlist or metalink)
 
-    def _fetch_md(self, repo):
+    def _fetch_md(self, reponame):
```
jkonecny12: Please change `reponame` to `repo_name`.

AdamWill: Done (what's 13 months between friends)

AdamWill:
OK, so this got a bit lost...but the bug is still there and this is still the fix. Resolved @jkonecny12 's issue and rebased.

jkonecny12:
@AdamWill you have to rebase on `f31-devel` if you want this change there.

jkonecny12:
@AdamWill please fix the commits. It looks that the rebase wasn't really successful. There should be only one commit.

jkonecny12:
Jenkins, it's ok to test.

jkonecny12:
Tested and everything looks great. The failing test is expected. I'll merge this as soon as the tests are fixed to be sure that these failing tests are not covering something else.

jkonecny12:
jenkins, test this please

PR #1520, opened 2018-06-29
New implementation of property changed signaling

jkonecny12:
This is just a draft I want to know your feedback before I'll spent more time on implementation of this.

### DIFF
--- a/pyanaconda/modules/common/base/base.py
+++ b/pyanaconda/modules/common/base/base.py
@@ -33,6 +33,46 @@ from pyanaconda.anaconda_loggers import get_module_logger
 log = get_module_logger(__name__)
 
 
+class SignalChangedProperty(object):
+    """Descriptor to implement signal changed 'property'."""
+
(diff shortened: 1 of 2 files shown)


jkonecny12:
> We talked about several solutions for signalling changed properties before we accepted the current one. This solution was one of them and it was rejected.
> 
> Problems:
> 
> When you want to connect to a signal, you have to access a magical attribute *_changed.

Yes you need.

> Developers and pylint will complain, that they don't know what this magical attribute is.

Yep.

> The solution cannot be used for passwords.

Why not?

> Basically, you are replacing two lines of code with other two lines, so I don't see a benefit of this.

I don't think so. The emit decorator can be used repeatedly and it is smaller copy & paste solution. Also more prone to typos.

jkonecny12:
OK, this doesn't solve the main issues and introducing new ones. I'm still not happy with existing copy & paste based solution but I agree this doesn't give is the required benefits.

PR #1525, opened 2018-07-04
Add support for LUKS2

poncovka:
**DO NOT MERGE!**

* Implemented a support for LUKS2 options in kickstart commands autopart, logvol, part and raid.
* Added a new button for the LUKS version in the custom partitioning spoke.

TODO:
* Add support for options `iterations` and  `time_ms`.
* Write and run new kickstart tests.
* Test GUI.

Depends on:
* https://github.com/storaged-project/blivet/pull/697
* https://github.com/clumens/pykickstart/pull/228


poncovka:
You can choose the LUKS version if the Encrypt option is selected.

luks_screenshot_vpn_test_01_2018-07-04_20 06 25

Otherwise, the widget is hidden.

luks_screenshot_vpn_test_01_2018-07-04_20 07 06

jkonecny12:
I like this solution. The only drawback I'm seeing is those jumping labels. Are we doing this also for something else or this is the first option doing that?

If this is the first maybe we should anchor those labels from bottom.

poncovka:
@jkonecny12, I am not sure what you mean by the jumping labels, but for example the UI for Volume Group depends on the Device Type and changes for every type.

jkonecny12:
@poncovka I mean the `Label:` and `Name:` labels with their text fields are moving based on the `Encrypt` checkbox. They are higher on the first image then on the second one.

So if I'm not mistaken when you are clicking on and off `Encryption` checkbox then they will `jump` up and down. That is bad user experience. I know that this is happening thanks to the added combobox but then those moving parts shouldn't be anchored to this but rather to the bottom of the page.

poncovka:
The jumping labels are fixed.

poncovka:
This pull request is ready for review.

pyanaconda/storage_utils.py
```diff
@@ -945,3 +946,24 @@ def get_supported_filesystems():
 
 def get_supported_autopart_choices():
     return [c for c in AUTOPART_CHOICES if is_supported_device_type(AUTOPART_DEVICE_TYPES[c[1]])]
+
+def get_pbkdf_args(luks_version, pbkdf_type=None, max_memory_kb=0, iterations=0, time_ms=0):
```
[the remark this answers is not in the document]
jkonecny12: OK in that case it is fine.

pyanaconda/kickstart.py
```diff
@@ -1270,12 +1271,29 @@ def execute(self, storage, ksdata, instClass):
             self.passphrase = self.passphrase or storage.encryption_passphrase
 
             cert = getEscrowCertificate(storage.escrow_certificates, self.escrowcert)
+
```
jkonecny12: You have this identical code 3 times in this file, please make that a method instead.

poncovka: The shared code is in the `get_pbkdf_args` method and I can put there also these lines:

```
if pbkdf_args and not luks_data.pbkdf_args:
    luks_data.pbkdf_args = pbkdf_args
```

Otherwise, I don't like it as well, but these execute methods are awful already and need to be refactorized. One weird method just for LUKS will not help with it. If you look above, you can see that the code for a passphrase and a certificate is also identical.

jkonecny12: The question is if you don't want to start with the refactorization on this point :). But I agree this is not something for NACK in this case.

poncovka:
Rebased and updated the lastest commit.

poncovka:
Updated function `get_pbkdf_args`.

PR #1526, opened 2018-07-06
Make pyanaconda.dbus.typing work with Python 3.7 (#1598574)

AdamWill:
As reported in RHBZ#1598574, the internals of typing changed in
Python 3.7 such that it's no longer so simple to find the 'base'
type of a type hint (it's not just its `__origin__` any more).
There doesn't appear to be any particularly great fix for this,
but this suggestion from Miro Hrončok seems as good as any other
option we have for now. This should work with both 3.6 and 3.7.

### DIFF
--- a/pyanaconda/dbus/typing.py
+++ b/pyanaconda/dbus/typing.py
@@ -125,13 +125,14 @@ class DBusType(object):
 
         :raises ValueError: for unknown types
         """
-        # Try base types.
+        # Try basic types.
         if DBusType._is_basic_type(type_hint):
             return DBusType._get_basic_type(type_hint)
 
         # Try container types.
-        if DBusType._is_container_type(type_hint):
-            return DBusType._get_container_type(type_hint)
+        base_type = _get_container_base_type(type_hint)
+        if base_type:
+            return DBusType._get_container_type(type_hint, base_type)
 
         # Or raise an error.
         raise TypeError("Unknown type: %s" % type_hint)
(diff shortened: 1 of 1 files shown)


pyanaconda/dbus/typing.py
```diff
@@ -149,24 +149,30 @@ def _get_basic_type(type_hint):
...
+        return False
 
     @staticmethod
     def _get_container_type(type_hint):
         """Return a container type."""
-        # Get the "base" type of the container.
-        origin = type_hint.__origin__
+        # Get the "base" type via the "origin" of the hint
+        # see https://bugzilla.redhat.com/show_bug.cgi?id=1598574
+        basetype = tuple(contype for contype in DBusType._container_type_mapping
```
hroncok: Could use next(...) instead if tuple(...)[0]

pyanaconda/dbus/typing.py
```diff
@@ -149,24 +149,30 @@ def _get_basic_type(type_hint):
...
     def _is_container_type(type_hint):
         """Is it a container type?"""
-        # Try to get the "base" type of the container type.
+        # Try to get the "origin" of the hint.
         origin = getattr(type_hint, "__origin__", None)
-        return origin in DBusType._container_type_mapping
+        if origin:
+            # Return true if the "origin" is a subclass of a container type
+            # see https://bugzilla.redhat.com/show_bug.cgi?id=1598574
+            return any(issubclass(origin, contype) for contype in DBusType._container_type_mapping)
```
hroncok: Could use tuple(DBus...) as the secon argument of issubclass and avoid any()

hroncok:
Long term, it would be nice to have a test that would cover this. I've built anaconda in copr when 3.7 was still beta and the build passed and I moved on. This particular problem only occurred during compose creation once the 3.7 side tag was merged. I'm not saying let's block the PR on this, just that it would be nice.

hroncok:
Oh, that's even worse.

jkonecny12:
Yeah no tests are run now because in general it is really hard to make them work. But we can at least create subset of tests which can run during build.

AdamWill:
There are `test-rpm` and `rpm-rpm-tests-only` targets in `Makefile.am`, which I've always assumed are/were for this purpose. I've never looked into why they are not used at present.

jkonecny12:
They are quite new and are check for us that all files are correctly contained in the rpm file and other similar checks. I'm more convinced to run nosetests (or their subset) and pylint.

hroncok:
To me it isn't clear what changes did you requested.

pyanaconda/dbus/typing.py
poncovka: Please, create a new method `_get_container_base_type`, that will get a type hint and return a base type or None. Then you can simplify the `_is_container_type` and `_get_container_type` methods.

poncovka:
We have unittests for the typing module here: https://github.com/rhinstaller/anaconda/blob/master/tests/nosetests/pyanaconda_tests/dbus_typing_test.py

The tests pass for python 3.6 and 3.7 with the current pull request.

AdamWill:
There, tweaked a bit as @poncovka requested - do you prefer that?There, tweaked a bit as @poncovka requested - do you prefer that?

poncovka:
@AdamWill, I had in mind something like this: https://github.com/poncovka/anaconda/commit/2fca5086b4509fe1e7820cc8d3de248dbd814767

AdamWill:
That's approximately what I did at first, but then I refined it again. I think what I came up with is better, because it avoids basically getting the base type twice, which is what happens if we do first `_is_container_type` and then `_get_container_type`.

Since all these methods are only used inside this one file anyway there doesn't seem any particularly great reason to retain `_is_container_type` rather than refactoring things so the discovery of the base type only happens *once*.

AdamWill:
Bonus note: funnily enough, I was reading about the PEP 572 controversy this weekend, and this would be a great place to use it - we could simplify these lines:

    base_type = _get_container_base_type(type_hint)
    if base_type:
        return DBusType._get_container_type(type_hint, base_type)

To:

    if base_type := _get_container_base_type(type_hint)
        return DBusType._get_container_type(type_hint, base_type)

:P

poncovka:
Jenkins, test this please.

poncovka:
We needed to do a new build for rawhide, so there is a new pull request at https://github.com/rhinstaller/anaconda/pull/1538 based on this one.

Thanks for your help!

AdamWill:
I still think mine's better, but it's up to you :P

AdamWill:
BTW, I'd already patched Rawhide downstream days ago.

M4rtinK:
> BTW, I'd already patched Rawhide downstream days ago.

Yep, we know & thanks for that! :)

PR #1529, opened 2018-07-12
3 column hub

rvykydal:
Mockups can be seen in the BZ:
https://bugzilla.redhat.com/show_bug.cgi?id=1584160

Screenshots:

1) Network spoke mocked into a separate category so there are 4 categories:
- english:
https://rvykydal.fedorapeople.org/hub_scale/4_categories_en_summary.png
https://rvykydal.fedorapeople.org/hub_scale/4_categories_en_progress.png
- german:
https://rvykydal.fedorapeople.org/hub_scale/4_categories_de_summary.png
https://rvykydal.fedorapeople.org/hub_scale/4_categories_de_progress.png
2) Current state (3 categories) english:
https://rvykydal.fedorapeople.org/hub_scale/current_en_summary.png
https://rvykydal.fedorapeople.org/hub_scale/current_en_progress.png
3) Using bigger icons (english):
https://rvykydal.fedorapeople.org/hub_scale/4_categories_en_summary_bigger_icons.png
https://rvykydal.fedorapeople.org/hub_scale/4_categories_en_progress_bigger_icons.png

### DIFF
--- a/pyanaconda/ui/gui/hubs/__init__.py
+++ b/pyanaconda/ui/gui/hubs/__init__.py
@@ -129,7 +129,8 @@ class Hub(GUIObject, common.Hub):
 
         grid = Gtk.Grid(row_spacing=12, column_spacing=12, column_homogeneous=True,
                         margin_bottom=12, margin_left=12, margin_right=12,
-                        halign=Gtk.Align.CENTER, valign=Gtk.Align.CENTER)
+                        halign=Gtk.Align.CENTER, valign=Gtk.Align.CENTER,
+                        row_homogeneous=True)
 
         max_row = category_row = 0
         col = 0
@@ -202,8 +203,9 @@ class Hub(GUIObject, common.Hub):
 
             row = category_row
 
-            label = Gtk.Label(label="<b>%s</b>" % escape_markup(_(obj.title)),
-                              use_markup=True, halign=Gtk.Align.START, margin_top=12, margin_bottom=6)
+            label = Gtk.Label(label="<span size=\"larger\" weight=\"bold\">%s</span>" % escape_markup(_(obj.title)),
+                              use_markup=True, halign=Gtk.Align.START,
+                              valign=Gtk.Align.END, margin_bottom=6)
             grid.attach(label, col, category_row, 1, 1)
             row += 1
 
--- a/pyanaconda/ui/gui/hubs/__init__.py
+++ b/pyanaconda/ui/gui/hubs/__init__.py
@@ -114,6 +114,8 @@ class Hub(GUIObject, common.Hub):
         self._spokesToStepIn = []
         self._spokeAutostepIndex = 0
 
+        self._gridColumns = 3
+
     def _createBox(self):
         import gi
 
@@ -125,10 +127,12 @@ class Hub(GUIObject, common.Hub):
         cats_and_spokes = self._collectCategoriesAndSpokes()
         categories = cats_and_spokes.keys()
 
-        grid = Gtk.Grid(row_spacing=6, column_spacing=6, column_homogeneous=True,
-                        margin_bottom=12)
+        grid = Gtk.Grid(row_spacing=12, column_spacing=12, column_homogeneous=True,
+                        margin_bottom=12, margin_left=12, margin_right=12,
+                        halign=Gtk.Align.CENTER, valign=Gtk.Align.CENTER)
 
-        row = 0
+        max_row = category_row = 0
+        col = 0
 
         for c in sorted(categories, key=lambda c: c.title):
             obj = c()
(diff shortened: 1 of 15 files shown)


pyanaconda/ui/gui/hubs/__init__.py
```diff
@@ -125,10 +127,13 @@ def _createBox(self):
...
+        grid = Gtk.Grid(row_spacing=12, column_spacing=12, column_homogeneous=True,
+                        margin_bottom=12, margin_left=12, margin_right=12,
+                        halign=Gtk.Align.CENTER, valign=Gtk.Align.CENTER,
+                        row_homogeneous=True)
 
-        row = 0
+        max_row = category_row = 0
+        col = 0
 
         for c in sorted(categories, key=lambda c: c.title):
```
poncovka: Shouldn't the categories be sorted by their `sortOrder`?

jkonecny12:
I like the summary hub but the progress hub feels too empty to me. Is it just me or are others feel it the same way?

poncovka:
@jkonecny12, I think, that we could move the user spoke to the summary hub and reorganize the progress hub as was suggested (the progress bar in the middle). It shouldn't be too difficult to do that. I just don't know if it is planned to do that soon.

rvykydal:
@poncovka 
To me, the original version with bigger icons:
https://rvykydal.fedorapeople.org/hub_scale/4_categories_en_summary_bigger_icons.png
seems fine, I don't mind the space around the grid, I like a bit more spacious feel. Also, and more importatntly it leaves some reserve for languages with even longer translations (or for new categories with longer names) for future.

For your proposal I tried to increment the grid by the smallest step I was able to achieve (while keeping the colums homogenous which is what we want, non-homogenous columns looked ugly to me) - which means changing spoke title font medium -> large, removing spacing between columns and margins for grid (and also I made category names a step bigger):
(1) https://rvykydal.fedorapeople.org/hub_scale/bigger/current_summary_de_bigger.png
(2) https://rvykydal.fedorapeople.org/hub_scale/bigger/current_progress_de_bigger.png
(3) https://rvykydal.fedorapeople.org/hub_scale/bigger/4_categories_summary_de_bigger.png
(4) https://rvykydal.fedorapeople.org/hub_scale/bigger/4_categories_progress_de_bigger.png
... these look ok to me as well, but my concern is that there is no reserve for potentially longer strings (see (1) above)
Using even larger icons than in my original bigger icons proposal would wasting of space IMO.

rvykydal:
@jkonecny12 
IMO the progress spoke doesn't look much worse (or emptier) than the current state:
https://rvykydal.fedorapeople.org/hub_scale/bigger/current_progress.png
It looks empty because it is almost empty...

@poncovka I don't think this is the right reason for moving user settings to summary hub. It does not seem to outweigh the original reason for having user settings on summary hub (doing configuration while the work is being done -> faster installation) to me. 
If we want to move user settings to summary hub I think it should be done in a separate step/PR. Perhaps the changes would be also more acceptable for the users in two steps and potential issues with moving user to summary would not block this change.

rvykydal:
New screenshots.
1) I added more space between columns to expand the grid a bit
2) I realized that russian is better for testing of overflowing:
https://rvykydal.fedorapeople.org/hub_scale/spacier/ru_problem.png
so I am now wrapping category label:
https://rvykydal.fedorapeople.org/hub_scale/spacier/ru_wrap.png

So, the german version with more space between columns:
https://rvykydal.fedorapeople.org/hub_scale/spacier/de_wrap.png
https://rvykydal.fedorapeople.org/hub_scale/spacier/de_wrap_progress.png

3) Not using capitals in spoke names (as in the mockup in the BZ) seems a lot nicer to me. I am showing only english version as the titles will have to be translated if we go with lower-case.
3s) Using lower-case:
https://rvykydal.fedorapeople.org/hub_scale/spacier/en_no_capitals_big_icon_small_title.png
https://rvykydal.fedorapeople.org/hub_scale/spacier/en_no_capitals_big_icon_small_title_progress.png
3b) Now with lower case it seems we can use bigger title size:
https://rvykydal.fedorapeople.org/hub_scale/spacier/en_no_capitals_big_icon_big_title.png
https://rvykydal.fedorapeople.org/hub_scale/spacier/en_no_capitals_big_icon_big_title_progress.png
current categories:
https://rvykydal.fedorapeople.org/hub_scale/spacier/current_no_capitals_big_icon_big_title.png
https://rvykydal.fedorapeople.org/hub_scale/spacier/current_no_capitals_big_icon_big_title_progress.png
3c) and perhaps smaller icons?
https://rvykydal.fedorapeople.org/hub_scale/spacier/en_no_capitals_small_icon_big_title.png
https://rvykydal.fedorapeople.org/hub_scale/spacier/en_no_capitals_small_icon_big_title_progress.png
current categories:
https://rvykydal.fedorapeople.org/hub_scale/spacier/current_no_capitals_small_icon_big_title.png
https://rvykydal.fedorapeople.org/hub_scale/spacier/current_no_capitals_small_icon_big_title_progress.png
3d) or smaller icons with smaller title size (I don't like this much):
https://rvykydal.fedorapeople.org/hub_scale/spacier/en_no_capitals_small_icon_small_title.png
https://rvykydal.fedorapeople.org/hub_scale/spacier/en_no_capitals_small_icon_small_title_progress.png

Personally I prefer 3b)

poncovka:
I like 3b) as well :-)

jkonecny12:
Yep, I also like 3b.

rvykydal:
Updated the PR for 3b version.
NOTE: we will need to de-capitalize addons.

rvykydal:
@poncovka: good point with the sorting, I added a patch using sortOrder.

rvykydal:
jenkins, test this please

PR #1530, opened 2018-07-13
Improve handling of unsupported filesystems in UI.

rvykydal:
UI should not offer filesystems that blivet/libblockdev does not support.

For handling of existing devices just make sure we don't crash, don't offer
modifications (just removing) and present the info about the device type not
being supported in a decent way.

Resolves: rhbz#1533904


rvykydal:
This is a port of https://github.com/rhinstaller/anaconda/commit/2f2ef62becb5c0c4d270d82acdaccb7763cfe273

rvykydal:
jenkins, test this please

PR #1532, opened 2018-07-16
Remove btrfs support

rvykydal:


jkonecny12:
> I think that we could have a python module for example in pyanaconda.core with all these imports of kickstart commands. The module would provide a version and version-less classes. Then the different versions of kickstart commands will not create conflicts in our code, but only in this module.

Interesting idea, you should create a ticked on that or something similar.

poncovka:
> Interesting idea, you should create a ticked on that or something similar.

I can implement it next week if everyone agrees with it. It should be merged before this pull request.

poncovka:
@jkonecny12, could you set up tests for the `rhel-devel` branch, please?

rvykydal:
@poncovka nice, I agree, thanks

poncovka:
Implemented in the pull request https://github.com/rhinstaller/anaconda/pull/1549.

rvykydal:
Closing, new PRs based on Vendy's kickstart commands patch:
https://github.com/rhinstaller/anaconda/pull/1551
https://github.com/rhinstaller/anaconda/pull/1552

PR #1533, opened 2018-07-17
Replace deprecated dracut options for booting with ibft.

rvykydal:
Generate correct options for installed system (part of the change happens
in blivet).

Accept rd.iscsi.ibft as installer option. Deprecated ip=ibft will keep working
until dracut removes it. When this happens, we can still use
parse-anaconda-options dracut script to translate it to the rd.iscsi.ibft
for some transitional timeframe.

Enabling NM ibft module via the patched service will still accept ip=ibft.

Resolves: rhbz#1566052


rvykydal:
jenkins, test this please

PR #1535, opened 2018-07-17
Change the pop-up text with the pre-release warning (#1542998)

rvykydal:
We should use the RHEL version of the pop-up text.

(cherry-picked from commits c02b585 and e033e50)

Resolves: rhbz#1542998


poncovka:
I would like to suggest to make this text default also for Fedora. What do others think?

jkonecny12:
You are suggesting to remove our timbuktu message, why? Are you agent from future?!?

But, yeah I not against nor for that. It doesn't matter to me.

rvykydal:
I agree with using the same text on Fedora:
https://github.com/rhinstaller/anaconda/pull/1546

PR #1537, opened 2018-07-18
Configure GRUB and zipl to use the "BootLoaderSpec" boot option configuration format.

martinezjavier:
This pull-request contains two commits from @vathpela that adds a inst.blscfg option to configure the GRUB and zipl bootloaders to use the BootLoaderSpec (BLS) boot option configuration format.

The first commit get rids of the calls to the new-kernel-pkg script since this no longer will be available once grubby is removed from the compose.

The second commit enables the BLS configuration by default. Since this is a new option, a inst.noblscfg is also added to allow anaconda to disable its use.

These changes are part of https://fedoraproject.org/wiki/Changes/BootLoaderSpecByDefault

### DIFF
--- a/data/anaconda_options.txt
+++ b/data/anaconda_options.txt
@@ -269,3 +269,6 @@ timeout in seconds specified as a mandatory value of the option.
 decorated
 Run GUI installer in a decorated window. By default, the window is not decorated, so it doesn't have a title bar,
 resize controls, etc.
+
+noblscfg
+Configure GRUB not to use the "BootLoader Spec" boot option configuration format.
--- a/pyanaconda/bootloader.py
+++ b/pyanaconda/bootloader.py
@@ -25,7 +25,7 @@ import re
 import blivet
 from parted import PARTITION_BIOS_GRUB
 from glob import glob
-from itertools import chain
+from itertools import chain, zip_longest
 import crypt
 from ordered_set import OrderedSet
 
@@ -36,7 +36,7 @@ from pyanaconda.product import productName
 from pyanaconda.flags import flags, can_touch_runtime_system
 from blivet.fcoe import fcoe
 import pyanaconda.network
-from pyanaconda.errors import errorHandler, ERROR_RAISE, ZIPLError
+from pyanaconda.errors import errorHandler, ERROR_RAISE, ZIPLError, FirmwareCompatError
 from pyanaconda.nm import nm_device_hwaddress
 from pyanaconda import platform
 from blivet.size import Size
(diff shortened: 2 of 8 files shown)


data/anaconda_options.txt
```diff
@@ -269,3 +269,9 @@ timeout in seconds specified as a mandatory value of the option.
 decorated
 Run GUI installer in a decorated window. By default, the window is not decorated, so it doesn't have a title bar,
 resize controls, etc.
+
+blscfg
+Configure GRUB to use the "BootLoader Spec" boot option configuration format.
+
+noblscfg
+Configure GRUB not to use the "BootLoader Spec" boot option configuration format.
```
jkonecny12: Why do we need to have both as boot options? Isn't one of them default behavior?

pyanaconda/bootloader.py
```diff
@@ -2257,6 +2311,9 @@ def boot_dir(self):
         return "/boot"
 
     def write_config_images(self, config):
+        if flags.cmdline.getbool("blscfg", True):
```
jkonecny12: Please do not use this string directly everywhere. The ideal solution would be to add this as bool flag to the `flags` class. However creating constant and moving it to constants would be enough.

pyanaconda/bootloader.py
```diff
@@ -2287,12 +2344,12 @@ def write_config_header(self, config):
         header = ("[defaultboot]\n"
                   "defaultauto\n"
                   "prompt=1\n"
-                  "timeout=%(timeout)d\n"
-                  "default=%(default)s\n"
+                  "timeout=%d\n"
                   "target=/boot\n"
-                  % {"timeout": self.timeout,
-                     "default": self.image_label(self.default)})
+                  % self.timeout)
```
jkonecny12: We are trying to replace `%` everywhere by `.format`. Not required but would be nice :)

data/anaconda_options.txt
vathpela: Well, you get them both anyway because that's how your option parser works, but yeah, the second description there is superfluous.

pyanaconda/bootloader.py
vathpela: Oh, I'd already done that and just didn't realize it would translate to flags.blscfg ; cool, will fix.

pyanaconda/bootloader.py
vathpela: Okay.

vathpela:
I've updated it in my branch (https://github.com/vathpela/anaconda/tree/blscfg here), but Javier is on PTO until Monday, so it won't get pushed to this PR until he gets back.

martinezjavier:
@jkonecny12 I've updated the branch with @vathpela modifications. The changes since v1 are:
- Remove `blscfg` option.
- Check `flags.blscfg` instead of `flags.cmdline.getbool("blscfg", True)`.
- Use `.format` instead of `%` for string formatting.

jkonecny12:
Thanks for the update but I'm closing this PR. 
The PR https://github.com/rhinstaller/anaconda/pull/1545 which is the same but targeted to the modularization-devel branch will propagate to master too. So this would make a duplicates.

PR #1538, opened 2018-07-18
Make pyanaconda.dbus.typing work with Python 3.7 (#1598574)

poncovka:
As reported in RHBZ#1598574, the internals of typing changed in
Python 3.7 such that it's no longer so simple to find the 'base'
type of a type hint (it's not just its `__origin__` any more).
There doesn't appear to be any particularly great fix for this,
but this suggestion from Miro Hrončok seems as good as any other
option we have for now. This should work with both 3.6 and 3.7.

Based on solution from Adam Williamson <awilliam@redhat.com>.

Resolves: rhbz#1598574

### DIFF
--- a/pyanaconda/dbus/typing.py
+++ b/pyanaconda/dbus/typing.py
@@ -149,24 +149,39 @@ class DBusType(object):
     @staticmethod
     def _is_container_type(type_hint):
         """Is it a container type?"""
-        # Try to get the "base" type of the container type.
+        return DBusType._get_container_base_type(type_hint) is not None
+
+    @staticmethod
+    def _get_container_base_type(type_hint):
+        """Return a container base type."""
+        # Try to get the "origin" of the hint.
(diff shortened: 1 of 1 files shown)


pyanaconda/dbus/typing.py
```diff
@@ -149,24 +149,39 @@ def _get_basic_type(type_hint):
     @staticmethod
     def _is_container_type(type_hint):
         """Is it a container type?"""
-        # Try to get the "base" type of the container type.
+        return DBusType._get_container_base_type(type_hint) is not None
```
AdamWill: why not just `bool(DBusType._get_container_base_type(type_hint))` ? that's  what i had in my version before i made it better ;)

poncovka: Version with `bool` is also acceptable, it is just a question of personal preference in this case :-)

PR #1541, opened 2018-07-19
Remove the python-wrapt dependency

poncovka:
We were using the wrapt package for logging to create object proxies
for streams with our own locking. Now, we are using our own proxy.


poncovka:
Jenkins, test this please.

poncovka:
Jenkins, test this please.

PR #1543, opened 2018-07-20
Show a note about EULA where relevant

M4rtinK:
Show a note about EULA at the end of the installation
in GUI & TUI if the currently active install class
specifies a path to an EULA file.

If no EULA file path is specified, no EULA note will
be shown.

### DIFF
--- a/pyanaconda/ui/tui/spokes/installation_progress.py
+++ b/pyanaconda/ui/tui/spokes/installation_progress.py
@@ -131,6 +131,14 @@ class ProgressSpoke(StandaloneTUISpoke):
 
         util.ipmi_report(IPMI_FINISHED)
 
+        if self.instclass.eula_path:
+            # Notify user about the EULA (if any).
+            print(_("Installation complete"))
+            print('')
+            print(_("Use of this product is subject to the license agreement found at:"))
+            print(self.instclass.eula_path)
+            print('')
+
         # kickstart install, continue automatically if reboot or shutdown selected
         if flags.automatedInstall and self.data.reboot.action in [KS_REBOOT, KS_SHUTDOWN]:
             # Just pretend like we got input, and our input doesn't care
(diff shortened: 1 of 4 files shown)


pyanaconda/ui/tui/spokes/installation_progress.py
```diff
@@ -131,14 +131,22 @@ def show_all(self):
 
         util.ipmi_report(IPMI_FINISHED)
 
+        if self.instclass.eula_path is not None:
```
poncovka: Please, remove `is not None`, so it is consistent with `pyanaconda/ui/gui/hubs/progress.py`.

pyanaconda/ui/tui/spokes/installation_progress.py
```diff
@@ -131,14 +131,22 @@ def show_all(self):
...
+
         # kickstart install, continue automatically if reboot or shutdown selected
         if flags.automatedInstall and self.data.reboot.action in [KS_REBOOT, KS_SHUTDOWN]:
             # Just pretend like we got input, and our input doesn't care
             # what it gets, it just quits.
             raise ExitMainLoop()
 
     def prompt(self, args=None):
-        return Prompt(_("Installation complete. Press %s to quit") % Prompt.ENTER)
+        return Prompt(_("Press %s to quit") % Prompt.ENTER)
```
poncovka: It looks like users will not see the `Installation complete.` message, if there is no EULA. Is it ok?

M4rtinK: That looks like some sort of cherry-pick artifacts, I'll drop it.

pyanaconda/ui/tui/spokes/installation_progress.py
M4rtinK: Fixed.

PR #1544, opened 2018-07-20
dracut: Add deps for fetch-kickstart-disk

cgwalters:
It looks like at some point after RHEL7 `rmdir` dropped out of
the initramfs as nothing else depends on it.  This broke
`fetch-kickstart-disk`.

### DIFF
--- a/dracut/module-setup.sh
+++ b/dracut/module-setup.sh
@@ -24,6 +24,8 @@ install() {
     # binaries we want in initramfs
     dracut_install eject -o pigz
     dracut_install depmod blkid
+    # Deps for fetch-kickstart-disk
+    dracut_install mount umount cp mkdir rmdir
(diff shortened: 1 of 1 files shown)


cgwalters:
(Note, not tested locally)

I was trying to use [OEMDRV functionality](https://bugzilla.redhat.com/show_bug.cgi?id=1057271) with Fedora Atomic Host on a USB stick.

jkonecny12:
Hi, what about having these tools in Dracut all the time. I mean not just add them to the Ananconda initrd but to the Dracut default initrd.

lnykryn:
@jkonecny12 rmdir is surprisingly used very rarely, so I don't think it should be included in default initramdisk (we try to keep it as small as possible). And I think that there is a good precedence in the 90crypt module that includes rmdir on its own https://github.com/dracutdevs/dracut/blob/master/modules.d/90crypt/module-setup.sh#L63

PR #1548, opened 2018-07-25
Use productmd library to parse .treeinfo (#1411673)

jkonecny12:
The python-productmd library is to parse and create treeinfo and other files used to create compose. By using the library we should be always compatible with current treeinfo structure.


jkonecny12:
**UPDATES:**
* Remove unused dependency.

PR #1550, opened 2018-07-25
Don't allow unlocking and using of LUKS2 devices (#1608251)

poncovka:
Users shouldn't be able to unlock or use LUKS2 devices.

* In GUI, users will see a message for uneditable devices.
* In TUI, the installer doesn't unlock any LUKS devices.
* In the rescue mode, it will not be possible to unlock LUKS2 devices.

Resolves: rhbz#1608251
Depends on: https://github.com/storaged-project/blivet/pull/708

### DIFF
--- a/anaconda.spec.in
+++ b/anaconda.spec.in
@@ -89,7 +89,7 @@ The anaconda package is a metapackage for the Anaconda installer.
 
 %package core
 Summary: Core of the Anaconda installer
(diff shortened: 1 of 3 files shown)


poncovka:
Can we use the UI for uneditable devices?

rhel76_lusk2

poncovka:
Fixed the rescue mode, thanks @pholica!

PR #1551, opened 2018-07-27
Remove kickstart btrfs support

rvykydal:


rvykydal:
@jkonecny12 this is all intentional, the two commits without the reference are only very loosely related to the btrfs removal (should be there in any case for rhel) and both the resolve commits resolve the issue.

PR #1553, opened 2018-07-27
Show better messages for NoSuchPackage and NoSuchGroup (#1599190)

poncovka:
Exceptions NoSuchPackage and NoSuchGroup should have a better string
representation, otherwise the installer in the cmdline mode shows only
the name of the missing package or group.

This change doesn't affect messages in other modes.

Resolves: rhbz#1599190


poncovka:
screenshot_vpn_test_01_2018-07-27_13 34 54

PR #1554, opened 2018-07-27
Fix disable additional repositories

jkonecny12:
I really wonder how is it possible but it looks like we didn't disabled DNF repositories if they were disabled in KS data.

Thanks to this we always used all the additional repositories even when user disabled them in the UI.


jkonecny12:
jenkins, test this please

jkonecny12:
I was looking into that but I don't see any line removal with the dnf repository disabling. So I don't think that commit is causing this issue.

PR #1556, opened 2018-07-30
Handle blivet gui support

rvykydal:


pyanaconda/ui/gui/spokes/storage.glade
```diff
@@ -214,6 +214,9 @@
       <action-widget response="0">need_space_cancel_button</action-widget>
       <action-widget response="3">need_space_reclaim_button</action-widget>
     </action-widgets>
+    <child>
+      <placeholder/>
+    </child>
```
poncovka: I think that you can remove this placeholder.

pyanaconda/ui/gui/spokes/storage.glade
```diff
@@ -427,6 +430,9 @@
       <action-widget response="4">no_space_quit_button</action-widget>
       <action-widget response="0">no_space_cancel_button</action-widget>
     </action-widgets>
+    <child>
+      <placeholder/>
+    </child>
```
poncovka: And this one.

pyanaconda/ui/gui/spokes/storage.py
```diff
@@ -1127,13 +1134,13 @@ def on_back_clicked(self, button):
...
+        self.autopart = self._get_selected_partitioning_method() == PartitioningMethod.AUTO
         disks = [d for d in self.disks if d.name in self.selected_disks]
         dialog = None
         if not self.autopart:
-            if self._customPart.get_active():
+            if self._get_selected_partitioning_method() == PartitioningMethod.CUSTOM:
                 self.skipTo = "CustomPartitioningSpoke"
-            if self._blivetGuiPart.get_active():
+            if self._get_selected_partitioning_method() == PartitioningMethod.BLIVET_GUI:
                 self.skipTo = "BlivetGuiSpoke"
```
poncovka: I think that `blivet_gui.py` needs to be changed as well. Imports from `blivegui` should be moved to the `initialize` method of `BlivetGuiSpoke`, because all spokes are imported dynamically by the hub at the beginning of the installation.

Have you tried to run tests with removed `blivet-gui`? Pylint doesn't complain about it?

pyanaconda/ui/gui/spokes/storage.py
```diff
@@ -314,23 +314,30 @@ def __init__(self, *args, **kwargs):
...
         # hide radio buttons for spokes that have been marked as visited by the
         # user interaction config file
         if sam.get_screen_visited("CustomPartitioningSpoke"):
             self._customPart.set_visible(False)
             self._customPart.set_no_show_all(True)
-        if sam.get_screen_visited("BlivetGuiSpoke"):
-            self._blivetGuiPart.set_visible(False)
-            self._blivetGuiPart.set_no_show_all(True)
+
+        blivet_gui_supported = True
```
jkonecny12: I would prefer to abstract this to a separate method (e.g. `enable_blivet_gui`) and call here that method only.

pyanaconda/ui/gui/spokes/storage.py
rvykydal: As we discussed, I'll disable the check on rhel branch.

pyanaconda/ui/gui/spokes/storage.py
rvykydal: Done

pyanaconda/ui/gui/spokes/storage.py
rvykydal: Updated https://github.com/rhinstaller/anaconda/pull/1557

pyanaconda/ui/gui/spokes/storage.glade
rvykydal: I'd rather keep it in so that
```
-<!-- Generated with glade 3.20.0 -->
+<!-- Generated with glade 3.20.3 -->
```
is true.

poncovka: Ok.

PR #1558, opened 2018-07-30
Starting from 3.0 DNF expects strings in comps queries

M4rtinK:
So make sure we only pass strings to the API, otherwise
DNF throws an assert error.

Basically Anaconda uses None internally to indicate the
environment has not been set and we need to prevent this
from leaking to the DNF comps API.

### DIFF
--- a/pyanaconda/payload/dnfpayload.py
+++ b/pyanaconda/payload/dnfpayload.py
@@ -912,6 +912,9 @@ class DNFPayload(payload.PackagePayload):
 
     def environmentId(self, environment):
         """Return environment id for the environment specified by id or name."""
+        # the enviroment must be string or else DNF >=3 throws an assert error
+        if not isinstance(environment, str):
+            log.warning("environmentId() called with non-string argument: %s", environment)
         env = self._base.comps.environment_by_pattern(environment)
         if env is None:
             raise payload.NoSuchGroup(environment)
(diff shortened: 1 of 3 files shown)


pyanaconda/payload/dnfpayload.py
```diff
@@ -912,6 +912,9 @@ def environmentDescription(self, environmentid):
 
     def environmentId(self, environment):
         """Return environment id for the environment specified by id or name."""
+        # the enviroment must be string or else DNF throws an assert error
+        if not isinstance(environment, str):
+            raise TypeError
```
poncovka: If `None` is an acceptable value for anaconda, then just test if `environment is None` and return `None`. It is dangerous to silently ignore all types of the environment.

jkonecny12: And also log this please.

M4rtinK: Makes sense, I've updated to PR to:
- shortcut on None to return None
- logged a warning if someone does try to feed the payload an environment that is not a string

pyanaconda/payload/dnfpayload.py
```diff
@@ -912,6 +912,10 @@ def environmentDescription(self, environmentid):
 
     def environmentId(self, environment):
         """Return environment id for the environment specified by id or name."""
+        # the enviroment must be string or else DNF throws an assert error
+        if not isinstance(environment, str):
+            log.warning("environmentId() called with non-string argument: %s", environment)
+            raise TypeError
```
poncovka: If the environment is not `None` or a string, then it is a bug in Anaconda. There is no reason to raise an exception, that will be later ignored. It is ok, that DNF fails in this case and it will cause Anaconda to crash.

M4rtinK: I'll remove the TypeError then will leave the warning & comment so that any issues can be more easily debugged (the assert error looks pretty confusing at first glance).

PR #1560, opened 2018-07-30
Only show the "closest mirror" source option where appropriate

M4rtinK:
Only show it on distros that actually have public mirrors,
based on install class provided hint.

Also when we are at it cleanup the relevant code a bit,
changing the property name to something more reasonable
and remove the needless override in dnfpayload.

Resolves: rhbz#1608241

### DIFF
--- a/pyanaconda/payload/__init__.py
+++ b/pyanaconda/payload/__init__.py
@@ -324,11 +324,20 @@ class Payload(object):
         # Additional packages required by installer based on used features
         self.requirements = PayloadRequirements()
 
+        # are mirrors available ?
+        # - this is an initial default value that will
+        #   be overridden by value from the current
+        #   install class in the setup() method
+        self._mirrors_available = True
+
     def setup(self, storage, instClass):
         """Do any payload-specific setup."""
         self.storage = storage
         self.instclass = instClass
         self.verbose_errors = []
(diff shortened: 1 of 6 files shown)


pyanaconda/payload/__init__.py
```diff
@@ -324,11 +324,17 @@ def __init__(self, data):
         # Additional packages required by installer based on used features
         self.requirements = PayloadRequirements()
 
+        # are mirrors available ?
```
jkonecny12: Please change this comment to something telling that this is only a placeholder and it will be set in setup method.

pyanaconda/payload/dnfpayload.py
```diff
@@ -785,10 +785,6 @@ def groups(self):
         groups = self._base.comps.groups_iter()
         return [g.id for g in groups]
 
-    @property
-    def mirrorEnabled(self):
-        return True
```
poncovka: Is it ok, that `mirrors_available` can be True for all types of payload?

jkonecny12: Please look on the code but if I'm not mistaken than this property will be ignored for payloads which doesn't support this.

M4rtinK: Yep, that's how it works AFAIK.

PR #1562, opened 2018-07-31
Add help placeholders

M4rtinK:
Actual RHEL help content is not yet available, so just always
show a placeholder on RHEL for now.

The placeholders and their use are temporary and will be removed
once correct help content becomes available.

### DIFF
--- a/pyanaconda/installclasses/rhel.py
+++ b/pyanaconda/installclasses/rhel.py
@@ -42,8 +42,9 @@ class RHELBaseInstallClass(BaseInstallClass):
 
     efi_dir = "redhat"
 
-    help_placeholder = "RHEL7Placeholder.html"
-    help_placeholder_with_links = "RHEL7PlaceholderWithLinks.html"
+    help_placeholder = "rhel_help_placeholder.xml"
+    help_placeholder_with_links = "rhel_help_placeholder.xml"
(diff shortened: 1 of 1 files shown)


pyanaconda/ihelp.py
```diff
@@ -96,6 +96,15 @@ def get_help_path(help_file, instclass, plain_text=False):
...
     :return str: full path to the help file requested or to a placeholder
     """
+
+    if instclass.name.startswith("Red Hat "):
+        # NOTE: Help content is not yet available for RHEL, so just always return
+        #       a placeholder for now.
+        if plain_text:
+            return "/usr/share/anaconda/rhel_help_tui_placeholder.txt"
+        else:
+            return "/usr/share/anaconda/rhel_help_gui_placeholder.xml"
```
poncovka: I think, that if you change `help_folder`, `help_placeholder` and `help_placeholder_plain_text` in the install class for RHEL, the `get_help_path` function might work without these modification.

poncovka:
Why doesn't the package `anaconda-user-help` provide the right placeholders? Then anaconda doesn't need to be changed at all. 

Otherwise, rebase this pull request to rhel-devel. It shouldn't be part of Fedora, since it is going to be reverted anyway.

PR #1563, opened 2018-08-02
DD: Use text mode when calling tools with subprocess

rvykydal:
Resolves: rhbz#1609744


rvykydal:
I've also updated the (failing) tests.

PR #1566, opened 2018-08-07
Detect that there is not enough space on a device (#1613232)

poncovka:
In live installations, we ignore all rsync errors except for the ones
that indicate not enough space on a device (see #868755 and #1592520).

The valid return code of rsync is 11 (Error in file I/O) now.

Resolves: rhbz#1613232


poncovka:
@jkonecny12, we should support only the code 11 based on https://bugzilla.redhat.com/show_bug.cgi?id=1592520#c24.

PR #1570, opened 2018-08-10
Use SimpleConfigFile to get PLATFORM_ID from /etc/os-release

M4rtinK:
SimpleConfigFile looks like a better match for parsing of
/etc/os-release than the primitive custom key value parser.

Also unlike the primitive parser it can handle /etc/os-release files
with blank lines.

Related: rhbz#1613296


M4rtinK:
Thanks! :)

PR #1571, opened 2018-08-10
Revert back to running DNF in a subprocess

M4rtinK:
When working on the modula installation support via the install_specs()
API we hit a weird issue with DNF crashing on a "foregin key error"
when accessing one of its sqlite databases.

When looking into it with the DNF developers we found out the issue
goes away when we just run the DNF transaction in a dummy process
instead of a real one. That seemed to work fine and there were no
apparent regressions at that time so we went with the workaround.

Looks like we might have been wrong - bug 1614511 strongly hints
on chroot being used by rpm od DNF during package installation trasaction
is causing GTK to crash by basically swapping the root filesystem under
it.

A similar issue back with yum was the main reason for isolating
the yum payload into a separate process. So switch back to the previous
behavior where the DNF trasaction runs in a separate sub-process to
shield the main Anaconda process from rpm/DNF chroot shenanigans.

Resolves: rhbz#1614511
Related: rhbz#1613296

### DIFF
--- a/pyanaconda/payload/dnfpayload.py
+++ b/pyanaconda/payload/dnfpayload.py
@@ -37,7 +37,6 @@ import configparser
 import collections
 import itertools
 import multiprocessing
-import multiprocessing.dummy
 import operator
 import hashlib
 import shutil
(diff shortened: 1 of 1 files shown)


M4rtinK:
**WARNING**

With dnf 3.2 this causes the following sqlite related crash:

dnf_transaction_in_subprocess_crash

If we want to apply this fix DNF needs to be first fixed to not crash if the transaction runs in a sub-process.

AdamWill:
Together with https://github.com/rpm-software-management/libdnf/pull/546 in a custom test ISO, this looks good.

PR #1573, opened 2018-08-14
Disable failing test because of change in RPM

jkonecny12:
This should be disabled until the BZ 1581757 is fixed.

This is happening only on Rawhide.


jkonecny12:
This needs to be solved on our side. Most probably I will remove shebang from the test.

PR #1540, opened 2018-07-19
Add inst.addrepo cmdline option

jkonecny12:
Add support for the `inst.addrepo` command line option.

This option supports:
* http, https, ftp
* nfs in the `repo` kickstart command variant
* file:// -- new solution for loading local installation tree structures

This will basically read the new `inst.addrepo` and pass it to the installation kickstart data. That data are then processed as usual. This boot option can be used multiple times so you can use it to specify multiple additional repositories.

The `file://` solution is brand new and it is used to load installation tree from the installation environment. It is working even when specified by the `repo` kickstart command so the output kickstart file works as expected.

*Resolves: rhbz#1595415*

PR #1574, opened 2018-08-15
Warn when repo names are not unique

jkonecny12:
When there are multiple repositories with the same name then the first repo is used. This is opposite from our usual kickstart behavior where the last command is used.

Unfortunately, we can't change this because users might depend on that so only log this in the package log.

*Related: rhbz#1595415*


jkonecny12:
This will be implemented in the https://github.com/rhinstaller/anaconda/pull/1540

PR #1581, opened 2018-08-27
Show a message for InstallSpecsMissing (#1599190)

poncovka:
The exception InstallSpecsMissing should have a string representation,
otherwise the installer shows an empty string in the cmdline mode.

This change doesn't affect messages in other modes.

Resolves: rhbz#1599190


poncovka:
Rebased to f29-devel.

poncovka:
This fix will be reverted again by https://github.com/rhinstaller/anaconda/pull/1592/. The bug is not worth fixing until the DNF payload is stabilized.

PR #1582, opened 2018-08-27
Fix branching documentation

M4rtinK:
There were two small issues:
- kickstart version no longer needs to be set manually
- pre-formatted text was not marked up correctly in one place

### DIFF
--- a/docs/release.rst
+++ b/docs/release.rst
@@ -293,17 +293,22 @@ First checkout the ``f<version>-release`` branch and merge ``f<version>-devel`` 
     git merge --no-ff f<version>-devel
 
 
-Then correct pykickstart version for the new Fedora release by changing all occurrences of:
+Then correct pykickstart version for the new Fedora release by changing all occurrences of
+the DEVEL constant imported from pykickstart for the F<version> constant, for example:
 
 ::
 
-    returnClassForVersion()
+    from pykickstart.version import DEVEL as VERSION
 
 to
 
 ::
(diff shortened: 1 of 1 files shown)


docs/release.rst
```diff
@@ -293,24 +293,10 @@ First checkout the ``f<version>-release`` branch and merge ``f<version>-devel``
...
-
-to
-
-::
-
-    returnClassForVersion(version=F<new version>)
-
-Commit the result. The commit will become one of the few exclusive release branch commits,
-as we can't let it be merged back to master via the devel branch for obvious reasons.
-
```
poncovka: It is still necessary to update the pykickstart version in these files:

https://github.com/rhinstaller/anaconda/blob/master/pyanaconda/core/kickstart/version.py#L21
https://github.com/rhinstaller/anaconda/blob/master/dracut/parse-kickstart#L46

M4rtinK: So DEVEL should be replaced by F<version> in these files ? I though that that was not needed and the F29 pykickstart package will just continue to point DEVEL to F29, while rawhide pykickstart builds will switch to DEVEL == F30.

poncovka: Yes. I don't think that pykickstart does that. It doesn't have a branch for F29.

M4rtinK: Ah, OK. So I'll adjust the docs & set the kickstart version on f29-release branch.

PR #1586, opened 2018-08-31
Add support for DBus structures

poncovka:
DBus structure is a dictionary of type Dict[Str, Variant] that maps
field names to variants of field values. However, it might be easier
to work with objects instead of dictionaries.

A data class that describes a DBus structure should be decorated
with the @dbus_structure decorator. Then instances of this class can
be transformed to a DBus structure with the get_structure function, or
a DBus structure can be transformed to an instance of this class with
the apply_structure function.

The class RealmData is an example of such a data class.

### DIFF
--- a/pyanaconda/dbus/typing.py
+++ b/pyanaconda/dbus/typing.py
@@ -28,7 +28,8 @@ from pydbus import Variant
 
 __all__ = ["Bool", "Double", "Str", "Int", "Byte", "Int16", "UInt16",
            "Int32", "UInt32", "Int64", "UInt64", "File", "ObjPath",
-           "Tuple", "List", "Dict", "Variant", "get_variant"]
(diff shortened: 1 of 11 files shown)


poncovka:
Jenkins, test this please.

pyanaconda/dbus/typing.py
```diff
@@ -56,6 +57,8 @@
 # Container types.
 # Use Tuple, Dict and List from typing.
 # Use Variant from pydbus and get_variant.
+# Use Structure instead of Dict[Str, Variant].
+Structure = Dict[Str, Variant]
```
jkonecny12: Wouldn't the `Variant` here be a cause of bad optimization and make the DBus API harder to read?

jkonecny12:
Thanks for explanation face to face.

PR #1588, opened 2018-09-03
network: remove dumpMissingDefaultIfcfgs

lkundrak:
This would be a lot nicer if done by Lorax.
    
See-Also: https://github.com/weldr/lorax/pull/454


lkundrak:
This actually crashed for me when there was a VLAN connection present with an unconfigured parent. Not sure why, but I believe it's a better idea to just remote this instead of fixing it, since it does the wrong thing anyway (creates connections the user didn't ask for).

Screenshot

poncovka:
@rvykydal could you look at this PR, please?

rvykydal:
This is rather fragile area with a lot of use cases to keep in mind.
Our kickstart tests for networking are a minimum, and there is a lot of cases which can't be covered by them.
The function you are removing does not concern only autoconnections. By the patch you are removing the assumption that each device supported/configurable in installer has its ifcfg file during installation (the code can rely to it for various reasons, eg manipulating ONBOOT value directly in ifcfg files) and also on installed system - which might be bigger problem - users may rely on it in their tooling, kickstarts, post scripts, etc...

I am not saying we can't try to get rid of this assumption but it needs big care and we have to dedicate some time to work on it.

PR #1592, opened 2018-09-06
DNF 3.5 compatibility

M4rtinK:
DNF 3.5 changes some of it's package & module installation interfaces
and we need to adjust Anaconda accordingly:

- the module-enable method was moved from Base to a separate module
- the install_specs() method now raises exceptions instead of returning
  a value if something is wrong
- the exception contains separate list of for missing missing or broken
  packages, groups or modules

Related: rhbz#1613296

### DIFF
--- a/pyanaconda/errors.py
+++ b/pyanaconda/errors.py
@@ -222,28 +222,47 @@ class ErrorHandler(object):
             else:
                 return ERROR_RAISE
 
-    def _install_specs_missing(self, exn):
+    def _install_specs_handler(self, exn):
+        missing_packages = exn.no_match_pkg_specs
+        missing_groups_modules = exn.no_match_group_specs
+        broken_packages = exn.error_pkg_specs
+        broken_groups_modules = exn.error_group_specs
+
         message = ""
-        if exn.missing_packages:
-            if len(exn.missing_packages) > 1:
-                packages = ", ".join(exn.missing_packages)
-                message = message + _("The following packages are missing:\n%s\n\n") % packages
-            else:
-                message = message + _("The following package is missing:\n%s\n\n") % exn.missing_packages[0]
 
-        if exn.missing_groups_and_modules:
-            if len(exn.missing_groups_and_modules) > 1:
-                groups_modules = ", ".join(exn.missing_groups_and_modules)
-                message = message + _("The following groups or modules are missing:\n%s\n\n") % groups_modules
-            else:
-                message = message + _("The following group or module is missing:\n%s\n\n") % exn.missing_groups_and_modules[0]
+        # missing packages/groups/modules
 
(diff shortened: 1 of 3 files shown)


pyanaconda/errors.py
```diff
@@ -222,28 +222,64 @@ def _noSuchPackageHandler(self, exn):
...
+            else:
+                message = message + _("The following group or module is broken:\n%s\n\n") % broken_groups_modules[0]
+
+        # if we have at least one broken package, group or module we will abort the installation
+        if broken_packages or broken_groups_modules:
+            broken_item_count = len(broken_packages) + len(broken_groups_modules)
+            if broken_item_count >1:
+                message = message + _("Some of packages, groups or modules are broken, the installation will be aborted.")
+            else:
+                message = message + _("Some of packages, groups or modules are broken, the installation will be aborted.")
```
poncovka: Please, remove the messages for just one package/group/module. I think that a message like `The following packages are broken: python` is readable enough.

Btw. the message on this line is even the same as the previous one.

pyanaconda/payload/dnfpayload.py
```diff
@@ -566,13 +555,23 @@ def _apply_selections(self):
...
+        except dnf.exceptions.MarkingErrors as e:
+            log.debug("install_specs(): some packages, groups or modules are missing or broken")
+            if e.no_match_pkg_specs:
+                log.debug("install_specs(): missing packages: %s", e.no_match_pkg_specs)
+            if e.no_match_group_specs:
+                log.debug("install_specs(): missing groups or modules: %s", e.no_match_group_specs)
+            if e.error_pkg_specs:
+                log.debug("install_specs(): broken packages: %s", e.error_pkg_specs)
+            if e.error_group_specs:
+                log.debug("install_specs(): broken groups or modules: %s", e.error_group_specs)
```
poncovka: Please, replace this code with `log.debug("install_specs(): %s", e)` and ask DNF to provide a nice string representation of this error.

pyanaconda/errors.py
M4rtinK: OK, it will be simpler & there will be less strings to translate. :)

pyanaconda/payload/dnfpayload.py
M4rtinK: I think we should keep this until DNF provides the improved string representation of the exception (I'll ask them for it) as the the information about missing or broken groups/packages/modules is important for debugging and should always be logged.

pyanaconda/errors.py
poncovka: It looks better, but you can simplify also the part for `Some of packages, groups or modules are broken, the installation will be aborted`.

M4rtinK: Oh, right, done.

PR #1594, opened 2018-09-07
rpmostreepayload: detect `/var` and `/var/` (#1626557)

dustymabe:
Handle if the user pases `/var/` in their kickstart instead of
`/var`.

Resolves: rhbz#1626557

### DIFF
--- a/pyanaconda/payload/rpmostreepayload.py
+++ b/pyanaconda/payload/rpmostreepayload.py
@@ -319,7 +319,8 @@ class RPMOSTreePayload(ArchivePayload):
         # to do the default ostree one.
         # https://github.com/ostreedev/ostree/issues/855
         varroot = '/ostree/deploy/' + ostreesetup.osname + '/var'
-        if storage.mountpoints.get("/var") is None:
+        if storage.mountpoints.get("/var") is None \
+           and storage.mountpoints.get("/var/") is None:
             self._setupInternalBindmount(varroot, dest='/var', recurse=False)
         else:
             # Otherwise, bind it


pyanaconda/payload/rpmostreepayload.py
```diff
@@ -319,7 +319,8 @@ def prepareMountTargets(self, storage):
         # to do the default ostree one.
         # https://github.com/ostreedev/ostree/issues/855
         varroot = '/ostree/deploy/' + ostreesetup.osname + '/var'
-        if storage.mountpoints.get("/var") is None:
+        if storage.mountpoints.get("/var") is None \
```
cgwalters: Oh, whee.  It's unfortunate that Blivet doesn't do canonicalization here.

cgwalters: @vojtechtrefny 
```
$ git diff
diff --git a/blivet/formats/fs.py b/blivet/formats/fs.py
index 81e367f4..9843b0e9 100644
--- a/blivet/formats/fs.py
+++ b/blivet/formats/fs.py
@@ -123,6 +123,10 @@ class FS(DeviceFormat):
         self._current_info = None  # info obtained by _info task
 
         self.mountpoint = kwargs.get("mountpoint")
+        # The rpmostreepayload for example looks for /var, and hence
+        # won't find /var/.  Canonicalize.
+        if self.mountpoint is not None:
+            self.mountpoint = self.mountpoint.rstrip('/')
         self.mountopts = kwargs.get("mountopts")
         self.label = kwargs.get("label")
         self.fsprofile = kwargs.get("fsprofile")
```
?

dustymabe:
There are multiple ways to solve the problem. I trust the anaconda teams judgment here so please implement fix in best way. This PR was just because I saw an easy answer.

dustymabe:
:+1: can someone make that happen in order to close out #1626557 ?

poncovka:
Ok, I will implement the validation in Anaconda.

dustymabe:
> Opened a pull request with the alternative solution: [clumens/pykickstart#243](https://github.com/clumens/pykickstart/pull/243)

thanks!

PR #1595, opened 2018-09-07
Require python3-librepo explicitly

jwboyer:
Anaconda imports the librepo module at runtime.  Previously this was transitively pulled in via dnf RPM dependencies.  However, as of dnf-3.4.0, dnf no longer requires python3-librepo itself.  This causes anaconda to traceback at runtime as the librepo python module is no longer present.

Add an explicit Requires for the python3-librepo package to fix this.


jkonecny12:
Thanks for your PR but we need to do the opposite. DNF doesn't use librepo anymore and we use the library only to save logs to an appropriate location. We need to remove this from Anaconda completely.

Closing this PR to create and I'll open a new one.

PR #1597, opened 2018-09-10
Remove librepo imports from Anaconda (#1626609)

jkonecny12:
Anaconda used the librepo library to redirect logs to a proper place. New DNF 3.4.0/3.5.0 is not using librepo library so there is no need to have this in Anaconda.

Also this library wasn't mentioned in the Anaconda spec file so it nows breaks Anaconda.

*Resolves: rhbz#1626609*


jkonecny12:
@cgwalters thanks for info! There were misinformation, I'll create a patch to solve this correctly.

PR #1602, opened 2018-09-10
Save lsblk output to the Anaconda traceback file

vojtechtrefny:
Currently the lsblk output is only attached to the reported bug
but it isn't saved to /tmp or the anaconda-tb file so it is
missing for manually reported bugs.
lsblk output is really useful for storage related bugs, so it
would be nice to have it always.

Resolves: rhbz#1627220


jkonecny12:
jenkins, test this please

PR #1608, opened 2018-09-17
Rhel devel prefixdevname

rvykydal:


dracut/anaconda-copy-prefixdefname.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh
+# Copy over persistent network device names to anaconda environment
+
+CMDLINE=$(cat /proc/cmdline)
+if [[ ${CMDLINE} =~ "net.ifnames.prefix=" ]]; then
+  mkdir -p /run/initramfs/state/etc/systemd/network
+  cp /etc/systemd/network/71-net-ifnames-prefix* /run/initramfs/state/etc/systemd/network
```
jkonecny12: Is the `71` prefix fixed number? Couldn't this change in the future for some reason?

pyanaconda/network.py
```diff
@@ -1828,6 +1828,9 @@ def update_onboot_value(devname, value, ksdata=None, root_path=None):
 def is_using_team_device():
     return any(nm.nm_device_type_is_team(d) for d in nm.nm_devices())
 
+def is_using_persistent_device_names():
+    return 'net.ifnames.prefix' in flags.cmdline
```
jkonecny12: Shouldn't this be parsed in the flags and then used as property here? @poncovka what do you think?

dracut/anaconda-copy-prefixdefname.sh
rvykydal: I discussed this with Michal Sekletar and he recommended this approach (eg rather than copying content of /etc/systemd/network which at this moment contains only the prefixdevname configuation files).

rvykydal:
Ported to master with https://github.com/rhinstaller/anaconda/pull/2340

PR #1609, opened 2018-09-17
Flags shouldn't process the kernel options

poncovka:
There is no reason why we should process the kernel options in the
flags, because these flags are later set with the Anaconda options
anyway. The kernel options are handled there.

Also, the flag `noefi` can be removed, because we don't use it and the
flag `leavebootorder` should be set from the Anaconda option.

The option `inst.gpt` should be fully supported now.


poncovka:
Tested interactive installation, kickstart installation and live installation. It seems to be ok.

poncovka:
Rebased.

PR #1611, opened 2018-09-17
Create the Anaconda configuration file

poncovka:
Added a very basic and incomplete version of the Anaconda
configuration file, that provides default values for the temporary
configuration file used by DBus modules.

### DIFF
--- a/pyanaconda/core/configuration/anaconda.py
+++ b/pyanaconda/core/configuration/anaconda.py
@@ -1,0 +1,225 @@
+#
+# Copyright (C) 2018 Red Hat, Inc.
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# the GNU General Public License v.2, or (at your option) any later version.
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY expressed or implied, including the implied warranties of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General
+# Public License for more details.  You should have received a copy of the
+# GNU General Public License along with this program; if not, write to the
+# Free Software Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
+# 02110-1301, USA.  Any Red Hat trademarks that are incorporated in the
+# source code or documentation are not subject to the GNU General Public
+# License and may only be used or replicated with the express permission of
+# Red Hat, Inc.
+#
+#  Author(s):  Vendula Poncova <vponcova@redhat.com>
+#
+import os
+from abc import ABC
+
+from pyanaconda.core.constants import ANACONDA_CONFIG_TMP, ANACONDA_CONFIG_DIR
+from pyanaconda.core.configuration.base import create_parser, read_config, write_config, \
+    get_option, set_option
+
+__all__ = ["conf", "AnacondaConfiguration"]
+
+
+class Section(ABC):
+    """A base class for representation of a configuration section."""
+
+    def __init__(self, section_name, parser):
+        self._section_name = section_name
+        self._parser = parser
+
+    def _get_option(self, option_name, converter=None):
+        """Get a converted value of the option.
+
+        :param option_name: an option name
+        :param converter: a function or None
+        :return: a converted value
+        """
+        return get_option(self._parser, self._section_name, option_name, converter)
+
+    def _set_option(self, option_name, value):
+        """Set the option.
+
+        :param option_name: an option name
+        :param value: an option value
+        """
+        set_option(self._parser, self._section_name, option_name, value)
+
+
+class AnacondaSection(Section):
+    """The Anaconda section."""
+
+    @property
+    def debug(self):
+        """Run Anaconda in the debugging mode."""
+        return self._get_option("debug", bool)
+
+
+class ServicesSection(Section):
+    """The Services section."""
+
+    @property
+    def selinux(self):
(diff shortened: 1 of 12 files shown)


poncovka:
Jenkins, test this please.

pyanaconda/configuration.py
```diff
@@ -0,0 +1,222 @@
...
+         -1  The value is not set.
+          0  SELinux is disabled.
+          1  SELinux is enabled.
+        """
+        value = self._get_option("selinux", int)
+
+        if value not in (-1, 0, 1):
+            raise ValueError("Invalid value: {}".format(value))
+
+        return value
```
jkonecny12: What about converting these values to enum in a process?

pyanaconda/configuration.py
```diff
@@ -0,0 +1,222 @@
...
+
+        :return: an instance of AnacondaConfiguration
+        """
+        config = cls()
+
+        # Read the temporary configuration file.
+        config_path = os.environ.get("ANACONDA_CONFIG_TMP", ANACONDA_CONFIG_TMP)
+
+        # Or use the defaults if it doesn't exist.
+        if not config_path or not os.path.exists(config_path):
```
jkonecny12: It would be nice to log when the `ANACONDA_CONFIG_TMP` variable is set but the folder doesn't exists.

pyanaconda/configuration.py
```diff
@@ -0,0 +1,222 @@
...
+
+        :param path: a path to the file
+        """
+        write_config(self._parser, path)
+
+    def validate(self):
+        """Validate the configuration."""
+        self._validate_members(self)
+
+    def _validate_members(self, obj):
```
jkonecny12: I would remove the `obj` object. It adds superfluous complexity.

pyanaconda/core/configuration.py
```diff
@@ -0,0 +1,133 @@
...
+        raise ConfigurationFileError(str(e), path)
+
+
+def get_option(parser, section_name, option_name, converter=None):
+    """Get a converted value of the option.
+
+    :param parser: an instance of ConfigParser
+    :param section_name: a section name
+    :param option_name: an option name
+    :param converter: a function or None
```
jkonecny12: Could you please describe the `converter` function declaration?

data/anaconda.conf.d/anaconda.conf
```diff
@@ -0,0 +1,34 @@
...
+# Enable SELinux usage in the installed system.
+# Valid values:
+#
+#  -1  The value is not set.
+#   0  SELinux is disabled.
+#   1  SELinux is enabled.
+#
+selinux = -1
+
+[Storage]
```
jkonecny12: Please add spaces only between major section. Otherwise it  is hard to keep track where is the major section starts and ends.

pyanaconda/core/configuration.py
```diff
@@ -0,0 +1,133 @@
...
+    """
+    try:
+        # Make sure that the section and option exists.
+        parser.get(section_name, option_name)
+
+        # Set the option to the string representation of the value.
+        parser[section_name][option_name] = str(value)
+
+    except (configparser.Error, ValueError) as e:
+        raise ConfigurationDataError(str(e), section_name, option_name)
```
jkonecny12: Sorry, but I'm not able to see why is there API for `pyanaconda/configuration.py` and here. What is the benefit why we have two APIs? Wouldn't it be better to convert `pyanaconda/configuration.py` to a module (python module) and have this as an internal implementation details, so to provide only the `pyanaconda/configuration.py` API?

pyanaconda/configuration.py
poncovka: Let's use obj=self.

pyanaconda/configuration.py
poncovka: Discussed.

pyanaconda/configuration.py
poncovka: Discussed, it is too early to log anything

pyanaconda/configuration.py
poncovka: It is not possible to set obj to self in the arguments.

pyanaconda/core/configuration.py
poncovka: Fixed.

data/anaconda.conf.d/anaconda.conf
poncovka: Fixed.

pyanaconda/configuration.py
jkonecny12: That's shame. We can still remove this by abstracting the recursion to a separate method. But that adds one more method just for this and even now it is not part of the class API.
So I'm leaving that decision on you. It is fine for me even with this form.

poncovka:
Updated.

pyanaconda/core/configuration/anaconda.py
```diff
@@ -0,0 +1,225 @@
...
+    """The Services section."""
+
+    @property
+    def selinux(self):
+        """Enable SELinux usage in the installed system.
+
+        Valid values:
+
+         -1  The value is not set.
+          0  SELinux is disabled.
```
M4rtinK: I think it might be good to mention the SELinux terminology there rather than enabled/disabled:
disabled = permissive
enabled = enforcing

poncovka: Fixed.

poncovka:
Updated. I will open a new pull request for changes in the DBus launcher.

PR #1613, opened 2018-09-18
Don't set Anaconda-specific flags in Blivet

poncovka:
The Anaconda-specific flags are going to be removed in Blivet, because
they are not used or useful anymore.

See: https://github.com/storaged-project/blivet/pull/723


poncovka:
Jenkins, test this please.

PR #1617, opened 2018-09-20
Set the VNC password directly (#1592686)

poncovka:
The vncpasswd tool was waiting for an input from a pipe, because the
file descriptor for writing wasn't closed. Let's not use pipes at all.
We can use the input argument of the communicate method instead.

Resolves: rhbz#1592686

### DIFF
--- a/pyanaconda/vnc.py
+++ b/pyanaconda/vnc.py
@@ -82,21 +82,24 @@ class VncServer(object):
 
     def setVNCPassword(self):
         """Set the vnc server password. Output to file. """
-
-        r, w = os.pipe()
         password_string = "%s\n" % self.password
-        os.write(w, password_string.encode("utf-8"))
+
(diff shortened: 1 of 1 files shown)


pyanaconda/vnc.py
```diff
@@ -82,21 +82,22 @@ def __init__(self, root="/", ip=None, name=None,
...
+            stdout=subprocess.PIPE
+        )
 
-            os.close(r)
-            os.close(w)
+        out = proc.communicate(password_string.encode("utf-8"))[0]
 
-        return rc
+        if proc.returncode != 0:
+            raise OSError("Unable to set the VNC password.")
```
jkonecny12: You should also log stderr in this case please if possible.

poncovka: Fixed.

poncovka:
Updated.

PR #1618, opened 2018-09-21
Fix addrepo parameter issues

jkonecny12:
Fix issues with the addrepo parameter and add tests for it.

### DIFF
--- a/pyanaconda/argument_parsing.py
+++ b/pyanaconda/argument_parsing.py
@@ -214,12 +214,10 @@ class AnacondaArgumentParser(ArgumentParser):
                 # we hate you.
 
                 continue
-            elif option.nargs in ("*", "?", "+"):
-                # store multiple values under one key
-                # parsing of these values to list is done in BootArgs object
-                if type(val) is list:
-                    setattr(namespace, option.dest, val)
-                    continue
+            elif type(val) is list:
+                for item in val:
+                    option(self, namespace, item)
+                continue
 
             option(self, namespace, val)
         return namespace
--- a/pyanaconda/argument_parsing.py
+++ b/pyanaconda/argument_parsing.py
@@ -454,8 +454,7 @@ def getArgumentParser(version_string, boot_cmdline=None):
     ap.add_argument("--stage2", dest="stage2", default=None, metavar="STAGE2_URL",
                     help=help_parser.help_text("stage2"))
     ap.add_argument("--addrepo", dest="addRepo", default=[], metavar="NAME,ADDITIONAL_REPO_URL",
-                    nargs='*', action="append",
-                    help=help_parser.help_text("addrepo"))
+                    action="append", help=help_parser.help_text("addrepo"))
     ap.add_argument("--noverifyssl", action="store_true", default=False,
                     help=help_parser.help_text("noverifyssl"))
     ap.add_argument("--liveinst", action="store_true", default=False,
--- a/pyanaconda/anaconda.py
+++ b/pyanaconda/anaconda.py
@@ -182,7 +182,11 @@ class Anaconda(object):
 
     @staticmethod
     def _get_additional_repo_name(repo):
-        name, rest = repo.split(',', maxsplit=1)
+        try:
+            name, rest = repo.split(',', maxsplit=1)
+        except ValueError:
+            raise RuntimeError("addrepo boot option has incorrect format. Correct format is: "
+                               "inst.addrepo=<name>,<url>") from None
         return name, rest
 
     @property


pyanaconda/errors.py
```diff
@@ -66,6 +67,11 @@ class ExitError(RuntimeError):
     pass
 
 
+class AnacondaArgumentError(Exception):
```
poncovka: Please, use an exception from the argparser if necessary.

anaconda.py
```diff
@@ -341,14 +360,10 @@ def start_debugger(signum, frame):
     util.ipmi_report(constants.IPMI_STARTED)
 
     if (opts.images or opts.dirinstall) and opts.liveinst:
```
poncovka: This check that should be done in `validate_args`.

anaconda.py
```diff
@@ -341,14 +360,10 @@ def start_debugger(signum, frame):
...
-        stdout_log.error("--liveinst cannot be used with --images or --dirinstall")
-        util.ipmi_report(constants.IPMI_ABORTED)
-        sys.exit(1)
+        exit_on_error(stdout_log, "--liveinst cannot be used with --images or --dirinstall")
 
     if opts.images and opts.dirinstall:
```
poncovka: This check that should be done in `validate_args`.

anaconda.py
```diff
@@ -687,9 +701,8 @@ def _earlyExceptionHandler(ty, value, traceback):
             image_count += 1
             flags.imageInstall = True
     except ValueError as e:
```
poncovka: This should be handled in `validate_args`.

anaconda.py
```diff
@@ -201,6 +201,18 @@ def parse_arguments(argv=None, boot_cmdline=None):
...
+    from pyanaconda.argument_parsing import ArgumentChecker
+    from pyanaconda.errors import AnacondaArgumentError
+
+    ac = ArgumentChecker()
+
+    try:
+        ac.validate_args(namespace)
+    except AnacondaArgumentError as e:
+        print("")  # print blank line to make the message easier to read
+        exit_on_error(logger, str(e))
```
poncovka: Please, consider https://docs.python.org/3/library/argparse.html#exiting-methods to report the errors.

anaconda.py
```diff
@@ -201,6 +201,18 @@ def parse_arguments(argv=None, boot_cmdline=None):
...
+def check_arguments(namespace, logger):
+    from pyanaconda.argument_parsing import ArgumentChecker
+    from pyanaconda.errors import AnacondaArgumentError
+
+    ac = ArgumentChecker()
+
+    try:
+        ac.validate_args(namespace)
+    except AnacondaArgumentError as e:
+        print("")  # print blank line to make the message easier to read
```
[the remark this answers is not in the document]
jkonecny12: That's the point. I don't want to log this I just want to make it more readable for user in command line. So this is on purpose print call.

pyanaconda/argument_parsing.py
```diff
@@ -247,8 +248,47 @@ def parse_args(self, args=None, boot_cmdline=None):
...
         # NOTE: this means that CLI options override boot options
         namespace = super().parse_args(args, namespace)
+
         return namespace
 
+class ArgumentChecker(object):
+    def validate_args(self, namespace):
+        errors = []
+
+        if hasattr(namespace, "addRepo"):
```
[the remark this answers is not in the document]
jkonecny12: Probably. This is rather make check which is not required than crash.

pyanaconda/argument_parsing.py
```diff
@@ -247,8 +248,47 @@ def parse_args(self, args=None, boot_cmdline=None):
...
+            errors.extend(self._addrepo_check(namespace.addRepo))
+
+        if errors:
+            raise AnacondaArgumentError("\n".join(errors))
```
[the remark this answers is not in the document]
jkonecny12: We can but what is the benefit of that?

anaconda.py
jkonecny12: The point is that this error is not only thanks to bad arguments. It could be because some resources are not valid or present or similar.

jkonecny12:
We agreed to rewrite this to have validation checks during parsing process but just store errors and raise them after logging is set.

poncovka:
This might be helpful: https://www.python.org/dev/peps/pep-0389/#discussion-sys-stderr-and-sys-exit

jkonecny12:
This is changed to hotfix only because the correct solution would be too invasive now.

PR #1626, opened 2018-09-26
Fix RHEL tests

M4rtinK:
Use appropriate tool and package versions on RHEL.


M4rtinK:
Obsoleted by other work.

PR #1628, opened 2018-09-27
Make sure fips is correctly enabled on target system (#1619568)

M4rtinK:
If installation is running in fips mode (fips=1 on boot command line)
Anaconda needs to make sure that also the target system is correctly
configured
for fips.

To achieve this we must call the fips-mode-setup tool after dracut finishes
initrd regeneration.

Also we can drop the optional requirement for the dracut-fips package,
which is no long needed or even available.

Resolves: rhbz#1619568


M4rtinK:
We should be able to backport this once the BLS patches land on Rawhide.

PR #1634, opened 2018-10-02
Move the glade adaptor to a separate plugin

dashea:
The glade adaptor is only needed when running the glade UI, so it
doesn't need to be in the main widgets library. Move glade-adaptor.c to
a plugin in the widgets-devel package.


dashea:
whoops, forgot to add the cppcheck script change. Fixed and re-pushed.

dashea:
I verified that anaconda still works using an updates.img with the new libAnacondaWidgets.so.4. I also confirmed that glade still works as expected with an updated anaconda-widgets and anaconda-widgets-devel.

PR #1636, opened 2018-10-04
Ignore errors when trying to activate unsupported swaps (#1635252)

vojtechtrefny:
Anaconda shouldn't crash when trying to use preexisting swaps
that can't be activated.

**Do not merge**, https://github.com/storaged-project/libblockdev/pull/405 needs to be merged first.

### DIFF
--- a/pyanaconda/storage/osinstall.py
+++ b/pyanaconda/storage/osinstall.py
@@ -676,6 +676,10 @@ class FSSet(object):
                 try:
                     device.setup()
                     device.format.setup()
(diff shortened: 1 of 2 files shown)


jkonecny12:
I guess we need to wait on new libblockdev version based on the tests.

M4rtinK:
jenkins, test this please

M4rtinK:
I'm still seeing some pylint errors in one of the test runs, but we just got a FE one an issue the requires this change to be batched in one Bodhi update together with DNF, so let's merge this now. We can fix the pylint issues if they are real in a follow-up patch.

PR #1639, opened 2018-10-04
Add option to set kernel.hung_task_timeout_secs option

rvykydal:
Adds installer boot option

inst.kernel.hung_task_timeout_secs

which sets

kernel.hung_task_timoeut_secs sysctl option

Used for VM guests in kickstart tests.

Resolves: rhbz#1633549


dracut/anaconda-set-kernel-hung-timeout.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/sh
+# anaconda-set-kernel-hung-timoeut.sh - set kernel hung timeout value
```
jkonecny12: Typo: `timoeut` -> `timeout`

rvykydal:
Typo fixed.

PR #1641, opened 2018-10-04
Rhel devel adjust to new help content

M4rtinK:
This pull requests adapts anaconda to changes in help content formatting.
- the first commit changes the helpFile properties to help_id
- the second commit updates help related paths in install classes
- the third commit adapts the help handling code to work with the new help content format


pyanaconda/ihelp.py
```diff
@@ -71,77 +117,67 @@ def _get_best_help_file(help_folder, help_file):
...
+    if plain_text:
+        placeholder = installclass.help_placeholder_plain_text
+    else:
+        placeholder = instclass.help_placeholder
+    placeholder_path = _get_best_help_file(instclass.help_folder, placeholder)
+    if placeholder_path:
+        return placeholder_path
+    else:
+        log.error("placeholder {placeholder_name} not found in {help_folder}".format(placeholder_name=instclass.help_placeholder,
+                                                                                     help_folder=instclass.help_folder))
```
jkonecny12: In this particular case I would prefer to use just `{}`  instead of the long name. It is not used for translation and it should be able to get to one line, thought.

PR #1642, opened 2018-10-05
Adjust to some DNF 3.6 changes

M4rtinK:
Adjust to some changes in DNF 3.6:
- catch ModuleBase.enable() now raises the MarkingErrors exception
- the exceptions raise by enable() & install_specs() should now have a
  proper string representation, so drop our custom code for that

### DIFF
--- a/pyanaconda/payload/dnfpayload.py
+++ b/pyanaconda/payload/dnfpayload.py
@@ -443,20 +443,27 @@ class DNFPayload(payload.PackagePayload):
         ksrepo.disable()
         self._add_repo(ksrepo)
         super().addDisabledRepo(ksrepo)
+
     def _enable_modules(self):
         """Enable modules (if any)."""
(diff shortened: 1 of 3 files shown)


pyanaconda/payload/dnfpayload.py
```diff
@@ -443,20 +443,27 @@ def addDisabledRepo(self, ksrepo):
...
                 module_spec = module.name
-            log.debug("enabling module: %s", module_spec)
-            try:
-                module_base = dnf.module.module_base.ModuleBase(self._base)
-                module_base.enable([module_spec])
-            except dnf.module.exceptions.NoModuleException as e:
-                self._payload_setup_error(e)
+            module_specs.append(module_spec)
+
+        # for ward the module specs to enable to DNF
```
jkonecny12: Please fix the typo `for ward` -> `forward`.

PR #1645, opened 2018-10-08
Set the VNC password directly

rvykydal:
The vncpasswd tool was waiting for an input from a pipe, because the
file descriptor for writing wasn't closed. Let's not use pipes at all.
We can use the input argument of the communicate method instead.

Resolves: rhbz#1634667

Port of master commit de88ab569091b503ced8f1a05da1afdb9eb40b32


rvykydal:
Tested locally.

PR #1646, opened 2018-10-08
Enable CI scripts on RHEL devel

jkonecny12:
Add pip support to our testing scripts and change them to work on new RHEL.

### DIFF
--- a/scripts/testing/setup-mock-test-env.py
+++ b/scripts/testing/setup-mock-test-env.py
@@ -155,6 +155,15 @@ One of these commands must be used. These commands can be combined.
     group.add_argument('--install', '-i', metavar='<packages>', action='store', type=str,
                        dest='install',
                        help="""install additional packages to the mock""")
+    group.add_argument('--install-pip', '-P', metavar='<pip packages>', action='store', type=str,
+                       dest='install_pip', default=None,
+                       help="""
+                       install additional packages from Python Package Index repository to the
+                       mock environment via the pip tool;
+                       there is a default set of pip packages which can be replaced by this
+                       parameter;
+                       set this command with empty string to disable pip completely
+                       """)
 
     group.add_argument('--run-tests', '-t', action='store_true', dest='run_tests',
                        help="""
@@ -205,7 +214,7 @@ def get_required_packages():
     return proc_res.stdout.decode('utf-8').strip()
 
 
-def get_pip_packages():
+def get_required_pip_packages():
     """Get pip packages for running Anaconda tests."""
     script = _get_script_dir() + os.path.sep + DEPENDENCY_SOLVER
     cmd = [script, "--pip"]
@@ -219,8 +228,10 @@ def install_required_packages(mock_command):
     packages = get_required_packages()
     install_packages_to_mock(mock_command, packages)
 
-    pip_packages = get_pip_packages()
-    install_pip_packages_to_mock(mock_command, pip_packages)
+
+def install_required_pip_packages(mock_command):
+    packages = get_required_pip_packages()
+    install_pip_packages_to_mock(mock_command, packages)
 
 
 def remove_anaconda_in_mock(mock_command):
@@ -353,7 +364,7 @@ if __name__ == "__main__":
     mock_cmd = create_mock_command(ns.mock_config, ns.uniqueext)
     success = True
 
-    if not any([ns.init, ns.copy, ns.run_tests, ns.install]):
+    if not any([ns.init, ns.copy, ns.run_tests, ns.install, ns.install_pip]):
         print("You need to specify one of the main commands!", file=sys.stderr)
         print("Run './setup-mock-test-env.py --help' for more info.", file=sys.stderr)
         exit(1)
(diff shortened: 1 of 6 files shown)


jkonecny12:
**DO NOT MERGE**

Pocketlint changes https://github.com/rhinstaller/pocketlint/pull/28 and https://github.com/rhinstaller/pocketlint/pull/26 must be merged and new pip version released before this will work.

scripts/testing/setup-mock-test-env.py
```diff
@@ -155,6 +155,13 @@ def parse_args():
     group.add_argument('--install', '-i', metavar='<packages>', action='store', type=str,
                        dest='install',
                        help="""install additional packages to the mock""")
+    group.add_argument('--install-pip', '-P', metavar='<pip packages>', action='store', type=str,
+                       dest='install_pip', default=None,
+                       help="""
+                       install pip additional packages in the mock
```
M4rtinK: Just an enhancement suggestion:

"install additional packages from  Python Package Index repository to the mock environment via the pip tool"

M4rtinK:
A small typo in one commit message:

"Small optimalization ni setup-mock-test-env" -> "Small optimalization in setup-mock-test-env"

scripts/testing/setup-mock-test-env.py
jkonecny12: Good idea, thanks!

jkonecny12:
**UPDATED**
Fixed suggestions from @M4rtinK

jkonecny12:
Still work to do:
/bin/sh: coverage3: command not found

jkonecny12:
**UPDATED**
We need to fix another binary name.

scripts/testing/setup-mock-test-env.py
```diff
@@ -343,13 +383,15 @@ def setup_mock(mock_command):
 
     if ns.init:
         setup_mock(mock_cmd)
-        mock_init_run = True
-        if ns.install:
-            install_packages_to_mock(mock_cmd, ns.install)
+        if ns.install_pip is None:
+            install_required_pip_packages(mock_cmd)
```
poncovka: I am a little confused. Why don't we install the required pip packages in `setup_mock`?

scripts/testing/setup-mock-test-env.py
```diff
@@ -155,6 +159,15 @@ def parse_args():
...
                        dest='install',
                        help="""install additional packages to the mock""")
+    group.add_argument('--install-pip', '-P', metavar='<pip packages>', action='store', type=str,
+                       dest='install_pip', default=None,
+                       help="""
+                       install additional packages from Python Package Index repository to the
+                       mock environment via the pip tool;
+                       there is a default set of pip packages which can be replaced by this
+                       parameter;
+                       set this command with empty string to disable pip completely
```
poncovka: This seems to be pretty hackish. When do we need to disable pip completely? It would be better to have an extra option for that, like --no-pip.

scripts/testing/setup-mock-test-env.py
```diff
@@ -197,19 +210,39 @@ def check_args(namespace):
...
     packages = get_required_packages()
     install_packages_to_mock(mock_command, packages)
 
 
+def install_required_pip_packages(mock_command):
+    packages = get_required_pip_packages()
+    install_pip_packages(mock_command, packages)
+
+
+def install_pip_packages(mock_command, addon_packages):
```
poncovka: It looks like this method could be removed and replaced with `install_pip_packages_to_mock`, because it doesn't do anything useful.

scripts/testing/setup-mock-test-env.py
```diff
@@ -343,13 +383,15 @@ def setup_mock(mock_command):
...
-        if ns.install:
-            install_packages_to_mock(mock_cmd, ns.install)
+        if ns.install_pip is None:
+            install_required_pip_packages(mock_cmd)
 
-    if ns.install and not mock_init_run:
+    if ns.install:
         install_packages_to_mock(mock_cmd, ns.install)
 
+    if ns.install_pip is not None:
```
poncovka: Does it makes sense to call pip if `ns.install_pip` is an empty string?

scripts/testing/setup-mock-test-env.py
jkonecny12: Because this way it is done on one place and you can easily ignore the default set.

scripts/testing/setup-mock-test-env.py
jkonecny12: I guess we can, should we also add the same for the rpm packages to be consistent?

scripts/testing/setup-mock-test-env.py
jkonecny12: Yes if you want to have the behavior in the help, no otherwise.

scripts/testing/setup-mock-test-env.py
jkonecny12: Yes it doesn't but it is consistent. The *_mock methods are something like private. They are working in mock directly and they shouldn't be used in the main.
I want to have this structure. Ideally it should be placed to multiple files but that would complicate things...

jkonecny12:
**UPDATED**:
Add `--no-pip`

PR #1649, opened 2018-10-09
Fix tests errors for rhel-devel

jkonecny12:
These are issues which needs to be resolved on rhel-devel to make tests working again.

### DIFF
--- a/pyanaconda/core/kickstart/commands.py
+++ b/pyanaconda/core/kickstart/commands.py
@@ -27,7 +27,7 @@ from pykickstart.commands.authselect import F28_Authselect as Authselect
 from pykickstart.commands.autopart import RHEL8_AutoPart as AutoPart
 from pykickstart.commands.autostep import FC3_AutoStep as AutoStep
 from pykickstart.commands.bootloader import RHEL8_Bootloader as Bootloader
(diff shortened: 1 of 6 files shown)


jkonecny12:
**UPDATED**
* fix rpm test issue

tests/nosetests/pyanaconda_tests/ks_version_test.py
```diff
@@ -27,6 +27,7 @@ class CommandVersionTestCase(unittest.TestCase):
 
     # Names of the kickstart commands and data that should be temporarily ignored.
     IGNORED_NAMES = {
+        "btrfs"
```
poncovka: You should just update the version of the command btrfs to `RHEL8_BTRFS` in `pyanaconda/core/kickstart/commands.py`.

jkonecny12:
**UPDATED**:
* Implemented correct fix for BTRFS error

Thanks for pointing that out @poncovka . I thought that this is a problem thanks to the removal.

jkonecny12:
**UPDATED**:
* Fix new errors in `ihelp.py`
* Add missing pylint false positive
* one line code polishing

PR #1654, opened 2018-10-11
Remove EXPERIMENTAL label for mountpoint assignment in TUI (#1636940)

M4rtinK:
The mountpoint assignment support in TUI is no longer considered
experimental, so drop the EXPERIMENTAL label.

Resolves: rhbz#1636940


M4rtinK:
Rawhide PR: https://github.com/rhinstaller/anaconda/pull/1660

PR #1659, opened 2018-10-16
Fix unified nfs iso

jkonecny12:
Add RepoLoader to abstract work with TreeInfo object from productmd library.

Use this class to check if at least one repodata folder is present on the installation media.


jkonecny12:
I also moved the `DEFAULT_REPOS` outside of `PackagePayload` object. I wanted to use this in the check but on the end I'm not convinced that we really want to use it instead of this solution. What is the benefit to test only the base repo? The changes to move the `DEFAULT_REPOS` are still there and I think it gives more sense now so I want to keep it that way.

If we agree that the test should be on the main repo instead of any repo I can change that easily.

pyanaconda/payload/__init__.py
```diff
@@ -76,6 +75,12 @@
 
 MAX_TREEINFO_DOWNLOAD_RETRIES = 6
 
+# Get list of repo names witch should be used as base repo
+DEFAULT_REPOS = [productName.split('-')[0].lower(),
```
M4rtinK: Would it make sense to have this in the main constants file so one could easily find the default repo names ? Or alternatively, I wonder if this should be even hardcoded in Anaconda VS provided as config file from some package for example, like we do with other things.

pyanaconda/payload/install_tree_metadata.py
```diff
@@ -0,0 +1,244 @@
...
+
+        for repo_md in self._meta_repos:
+            if repo_md.name in repos:
+                return repo_md
+
+        return None
+
+    def _check_variants(self):
+        if not self._meta_repos:
+            self._read_variants()
```
poncovka: Please, make this code part of `get_metadata_repos`.

pyanaconda/payload/install_tree_metadata.py
```diff
@@ -0,0 +1,244 @@
...
+
+    def get_repo_metadata_by_name(self, name):
+        """Get repository metadata object with given name.
+
+        :param name: Name of the variant to return.
+        :rtype name: VariantRepo object or None.
+        """
+        self._check_variants()
+
+        for variant in self._meta_repos:
```
poncovka: Please, iterate over `get_metadata_repos`.

pyanaconda/payload/install_tree_metadata.py
```diff
@@ -0,0 +1,244 @@
...
+
+    def get_base_repo_metadata(self, additional_names=None):
+        """Get repo metadata about base repository."""
+        self._check_variants()
+        repos = constants.DEFAULT_REPOS
+
+        if additional_names:
+            repos.extend(additional_names)
+
+        for repo_md in self._meta_repos:
```
poncovka: Please, iterate over `get_metadata_repos`.

pyanaconda/payload/install_tree_metadata.py
```diff
@@ -0,0 +1,244 @@
...
+        self._obj = obj
+        self._root_path = root_path
+
+    @property
+    def name(self):
+        """Get name of this repository."""
+        return self._name
+
+    @property
+    def repo_rel_path(self):
```
poncovka: Is the `repo` in the name necessary? Also, please don't use shortcuts. Also, we don't use this, so it might be hidden.

pyanaconda/payload/install_tree_metadata.py
```diff
@@ -0,0 +1,244 @@
...
+        """Get relative repository root path."""
+        return self._obj.paths.repository
+
+    @property
+    def repo_path(self):
```
poncovka: Same here.

pyanaconda/payload/install_tree_metadata.py
```diff
@@ -0,0 +1,244 @@
...
+        """Get absolute repository root path."""
+        if self.repo_rel_path == ".":
+            return self._root_path
+        else:
+            return self._root_path + "/" + self.repo_rel_path
+
+    def check_valid_repo(self):
```
poncovka: Could it be `is_valid`?

pyanaconda/payload/install_tree_metadata.py
```diff
@@ -0,0 +1,244 @@
...
+
+MAX_TREEINFO_DOWNLOAD_RETRIES = 6
+
+
+class InstallTreeMetadata(object):
+
+    def __init__(self):
+        self._tree_info = TreeInfo()
+        self._meta_repos = []
+        self._session = util.requests_session()
```
poncovka: It looks like this could be created locally in `load_url`.

pyanaconda/payload/install_tree_metadata.py
```diff
@@ -0,0 +1,244 @@
...
+        self._errors = []
```
poncovka: I would prefer exceptions.

pyanaconda/payload/install_tree_metadata.py
```diff
@@ -0,0 +1,244 @@
...
+        """Get relative repository root path."""
+        return self._obj.paths.repository
+
+    @property
+    def repo_path(self):
+        """Get absolute repository root path."""
+        if self.repo_rel_path == ".":
+            return self._root_path
+        else:
+            return self._root_path + "/" + self.repo_rel_path
```
poncovka: Please, use `os.path.join`.

pyanaconda/payload/__init__.py
jkonecny12: Preferably I want to get rid of this constant at all and move this to repo metadata. This hardcoding is bad from purpose.

Moving this to configuration file won't help. The problem is the first line where we are taking base repo  based on `productName`.

pyanaconda/payload/install_tree_metadata.py
```diff
@@ -0,0 +1,244 @@
...
+        return None
+
+    def _check_variants(self):
+        if not self._meta_repos:
+            self._read_variants()
+
+    def _read_variants(self):
+        for variant_name in self._tree_info.variants:
+            variant_object = self._tree_info.variants[variant_name]
+            self._meta_repos.append(RepoMetadata(variant_name, variant_object, self._path))
```
poncovka: I see the usage of `self._path` of potentially dangerous, because this will work only if `load_file` or `load_url` was called exactly once.

pyanaconda/payload/repo_loader.py
```diff
@@ -0,0 +1,95 @@
...
+        """
+        self._check_variants()
+
+        for variant in self._variants:
+            if name == variant.name:
+                return variant
+
+    def _check_variants(self):
+        if not self._variants:
+            self._read_variants()
```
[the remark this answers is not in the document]
jkonecny12: This is more elegant from the code point of view but it is basically the same. Fixed.

pyanaconda/payload/repo_loader.py
```diff
@@ -0,0 +1,95 @@
...
+
+    def get_variant_by_name(self, name):
+        """Get repository variant with given name.
+
+        :param name: Name of the variant to return.
+        :rtype name: VariantRepo object or None.
+        """
+        self._check_variants()
+
+        for variant in self._variants:
```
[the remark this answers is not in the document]
jkonecny12: When there will be a use than we can change it. Till that time I don't see a point.

pyanaconda/payload/repo_loader.py
```diff
@@ -0,0 +1,95 @@
...
+            if name == variant.name:
+                return variant
```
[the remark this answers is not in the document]
jkonecny12: Fixed.

pyanaconda/image.py
```diff
@@ -110,6 +110,28 @@ def findFirstIsoImage(path):
...
+    if os.path.exists(mount_path + '/.treeinfo'):
+        repo_loader.load_file(mount_path + "/.treeinfo")
+    elif os.path.exists(mount_path + "/treeinfo"):
+        repo_loader.load_file(mount_path + "/treeinfo")
+    else:
+        return False
+
+    # find first valid repo
+    for variant in repo_loader.get_variants():
+        import pdb; pdb.set_trace()
```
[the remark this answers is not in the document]
jkonecny12: Fixed

pyanaconda/payload/install_tree_metadata.py
jkonecny12: Fixed

pyanaconda/payload/install_tree_metadata.py
jkonecny12: Fixed

pyanaconda/payload/install_tree_metadata.py
jkonecny12: Fixed

pyanaconda/payload/install_tree_metadata.py
jkonecny12: Name changed. I don't see point in hiding that. It can be used in the future and it is used by member methods now.

pyanaconda/payload/install_tree_metadata.py
jkonecny12: Renamed.

pyanaconda/payload/install_tree_metadata.py
jkonecny12: Fixed

pyanaconda/payload/install_tree_metadata.py
jkonecny12: Fixed

pyanaconda/payload/install_tree_metadata.py
jkonecny12: Truth is that I forgot to clear repo list when new repo is loaded. There should be clean method for this.

pyanaconda/payload/install_tree_metadata.py
jkonecny12: Fixed

pyanaconda/payload/install_tree_metadata.py
jkonecny12: Fixed.

pyanaconda/payload/install_tree_metadata.py
```diff
@@ -0,0 +1,239 @@
...
+            log.debug("Retrieved '%s' from %s", file_name, url)
+        except requests.exceptions.RequestException as e:
+            log.info("Error downloading '%s': %s", file_name, e)
+            return (None, None)
+        return result, result.status_code
+
+    def _clear(self):
+        """Clear metadata repositories."""
+        self._meta_repos = []
+        self._path = ""
```
poncovka: Maybe `self._tree_info` should be cleared as well.

pyanaconda/payload/install_tree_metadata.py
```diff
@@ -0,0 +1,239 @@
...
+        else:
+            return os.path.join(self._root_path, self.relative_path)
+
+    def is_valid(self):
+        """Quick check if the repo is a valid repository.
+
+        This is only check if the repository is invalid. If success it doesn't mean it is valid.
+
+        :returns: True if repository is not invalid, False otherwise."""
+        return os.access(self.path + "/repodata", os.R_OK)
```
poncovka: Please, use `os.path.join`.

pyanaconda/payload/install_tree_metadata.py
jkonecny12: Yep it should. I totally forgot that.

pyanaconda/payload/install_tree_metadata.py
jkonecny12: OK

jkonecny12:
**UPDATED**

PR #1661, opened 2018-10-18
installclass: fix variant string for Atomic Host (#1640409)

dustymabe:
We need to fix the variant string for Atomic Host so it can properly be
detected on Atomic Host ISO images.

Resolves: rhbz#1640409


M4rtinK:
Rebased on f29-devel: https://github.com/rhinstaller/anaconda/pull/1662

dustymabe:
do we not want this on the master branch too? I was going to cherry-pick my commit once this had merged. Thanks for opening #1662 though.

PR #1663, opened 2018-10-18
Use the configuration file in DBus modules

poncovka:
The Anaconda DBus launcher is able to set up and clean up everything
related to the DBus modules. It will create a temporary configuration
file, start a DBus session, set up the environment, start the boss
and kickstart modules and wait for the modules to start. It raises
TimeoutError if the modules don't start on time.

The list of enabled kickstart modules is provided by the configuration file
now.

### DIFF
--- a/pyanaconda/dbus/launcher.py
+++ b/pyanaconda/dbus/launcher.py
@@ -25,91 +25,109 @@
 #
 
 import os
+import time
 from subprocess import TimeoutExpired
 
-from pyanaconda.dbus.constants import DBUS_ANACONDA_SESSION_ADDRESS
+from pyanaconda.core.configuration.anaconda import conf
 from pyanaconda.core.util import startProgram
-from pyanaconda.core.constants import ANACONDA_BUS_ADDR_FILE, ANACONDA_DATA_DIR
-from pyanaconda.anaconda_loggers import get_anaconda_root_logger
+from pyanaconda.core.constants import ANACONDA_BUS_ADDR_FILE, ANACONDA_CONFIG_TMP,\
+    ANACONDA_BUS_CONF_FILE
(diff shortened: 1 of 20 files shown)


pyanaconda/dbus/launcher.py
```diff
@@ -130,3 +148,53 @@ def stop(self):
...
+        file_name = ANACONDA_BUS_ADDR_FILE
+        run_dir = os.path.dirname(file_name)
+
+        if not os.path.exists(run_dir):
+            os.mkdir(run_dir)
+
+        with open(file_name, 'wt') as f:
+            f.write(self._bus_address)
+
+    def _remove_bus_address(self):
```
jkonecny12: I would rename this to `_remove_bus_address_file` to be more explicit.

pyanaconda/modules/common/__init__.py
```diff
@@ -31,3 +31,10 @@ def init():
     import locale
     from pyanaconda.core.constants import DEFAULT_LANG
     locale.setlocale(locale.LC_ALL, DEFAULT_LANG)
+
+    from pyanaconda.core.configuration.anaconda import conf
+    # The configuration is validated on the first import.
```
jkonecny12: Please move this comment above the line.

pyanaconda/dbus/launcher.py
poncovka: Fixed.

pyanaconda/modules/common/__init__.py
poncovka: Fixed.

poncovka:
Added support for loading configuration files from /etc/anaconda/conf.d, so the Initial Setup can use it.

PR #1664, opened 2018-10-18
Create a class for the Anaconda bus connection

poncovka:
Created a new class for the Anaconda bus connection, that can be
specified by an environmental variable or by a file. This class
is a parent of the default connection.

It is possible to use our DBus support to access the Anaconda DBus
modules without setting up the the environment now.

### DIFF
--- a/pyanaconda/dbus/connection.py
+++ b/pyanaconda/dbus/connection.py
@@ -21,6 +21,7 @@ from abc import ABC, abstractmethod
 
 import pydbus
 
+from pyanaconda.core.constants import ANACONDA_BUS_ADDR_FILE
 from pyanaconda.dbus.constants import DBUS_ANACONDA_SESSION_ADDRESS, DBUS_STARTER_ADDRESS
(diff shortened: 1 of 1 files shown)


pyanaconda/dbus/connection.py
```diff
@@ -188,22 +189,38 @@ def get_new_connection(self):
...
 
-class DBusDefaultConnection(Connection):
+class DBusAnacondaConnection(Connection):
+    """Representation of an Anaconda bus connection."""
+
+    def get_new_connection(self):
+        """Get an Anaconda bus connection."""
+        if DBUS_ANACONDA_SESSION_ADDRESS in os.environ:
+            bus_address = os.environ.get(DBUS_ANACONDA_SESSION_ADDRESS)
+
```
jkonecny12: Please remove these empty lines between `if-elif-else`.

poncovka: Fixed.

PR #1666, opened 2018-10-19
Fix local repo files aren't enabled (#1636739)

jkonecny12:
This is fall of PR:

3496e955f0763ab4016f96ef2af4988796b32005

Line to enable know repositories where accidentally removed.

*Resolves: rhbz#1636739*

*Reported-by: Edgar Hoch <edgar.hoch@ims.uni-stuttgart.de>*

### DIFF
--- a/pyanaconda/payload/dnfpayload.py
+++ b/pyanaconda/payload/dnfpayload.py
@@ -387,7 +387,9 @@ class DNFPayload(payload.PackagePayload):
         #     and use this new one.  The highest profile user of this is livecd
         #     kickstarts.
         if repo.id in self._base.repos:
-            if url or mirrorlist or metalink:
+            if not url and not mirrorlist and not metalink:
+                self._base.repos[repo.id].enable()
+            else:
                 with self._repos_lock:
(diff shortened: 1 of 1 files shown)


jkonecny12:
Also @jcpunk can you please verify that this won't break your SL workflow?

poncovka:
Btw. it would be great to have a kickstart test for this use case.

jkonecny12:
I was thinking about that too. It would be Fedora only but even that is valuable.

jkonecny12:
We will merge it now to get it as soon as possible to F29 (it will take some time, thanks to the bodhi). It can be reverted if there will be a problem.

jkonecny12:
Thanks @jcpunk :+1:

jkonecny12:
Thanks @jcpunk :+1:

jkonecny12:
Thanks @jcpunk :+1:

jkonecny12:
Thanks @jcpunk :+1:

jkonecny12:
Thanks @jcpunk :+1:

jkonecny12:
Thanks @jcpunk :+1:

PR #1668, opened 2018-10-19
Do not include unbootable disks in bootloader list.

dwlehman:
This is untested, but I think something along these lines makes sense to prevent SAN disks from being selected as the boot device.

### DIFF
--- a/pyanaconda/bootloader/execution.py
+++ b/pyanaconda/bootloader/execution.py
@@ -125,13 +125,18 @@ class BootloaderExecutor(object):
     def _is_usable_disk(self, d):
         """Is the disk usable for the bootloader?
 
-        Throw out drives that don't exist or cannot be used
-        (iSCSI device on an s390 machine).
+        Throw out drives that cannot be used as the boot device:
+            - disk w/ 'hidden' format (member of a RAID/multipath set)
+            - 'protected' disk (we cannot modify the disk's contents)
+            - any iSCSI disk on s390
+            - any non-local disk other than an iSCSI disk using iBFT (these are the
+              only type of remotely-connected disk that we know we can boot from)
         """
         return \
             not d.format.hidden and \
             not d.protected and \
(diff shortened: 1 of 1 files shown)


jkonecny12:
Hi Dave could you please rebase this to `master` branch.

pyanaconda/kickstart.py
```diff
@@ -557,7 +557,8 @@ def _is_usable_disk(self, d):
         return \
             not d.format.hidden and \
             not d.protected and \
-            not (blivet.arch.is_s390() and isinstance(d, iScsiDiskDevice))
+            not (blivet.arch.is_s390() and isinstance(d, iScsiDiskDevice)) and \
```
M4rtinK: I think there should be a comment about what this code does:

`('local' in d.tags or getattr(d, 'ibft', False))`

And ideally also about the other conditions on the line (the s390 & iScsiDiskDevice) are there and preclude the device from being used for for the bootloder.

jkonecny12: Or move this to a separate method.

jkonecny12:
Could you please explain why do you think this is happening? I'm not sure I understand this change set.

dwlehman:
Sure! If there is no explicit bootloader ordering given, we select the "first" bootable disk. We cannot boot from iSCSI disks unless they are configured using iBFT. We do filter out iSCSI disks on s390, but we don't have anything to rule out non-iBFT iSCSI disks in general. That's what this adds.

jkonecny12:
Thanks for explanation, finally I understand now.
Could you please add this to the commit message.

poncovka:
Hi, it looks like the problem was somewhere else. Anaconda was able to choose a valid disk, but it didn't specify it for the allocation of partitions in Blivet and Blivet chose an invalid one. See: https://github.com/rhinstaller/anaconda/pull/2680

I am closing this PR.

PR #1612, opened 2018-09-17
Configure GRUB and zipl to use the "BootLoaderSpec" boot option configuration format

martinezjavier:
This PR is basically the same than PR #1545, that contains @vathpela patches to add BootLoaderSpec support to Anaconda.

The only differences are that it's based on top of the `rhel-devel` branch, and adds a `Resolves: rhbz#1615442` to the commit message as asked by @poncovka.

PR #1545, opened 2018-07-23
Configure GRUB and zipl to use the "BootLoaderSpec" boot option configuration format

vathpela:
Same as the other branch: https://github.com/rhinstaller/anaconda/pull/1537

(cc @martinezjavier )

PR #1670, opened 2018-10-24
Configure GRUB and zipl to use the "BootLoaderSpec" boot option configuration format

martinezjavier:
This pull-request contains @vathpela patches to add BootLoaderSpec (BLS) support to Anaconda. These were already merged in the `rhel-devel` branch in PR #1612. So this is the same for the `master` branch, to add BLS support to Fedora rawhide.

It also contains some fixes that were part of PR #1633 and #1635.

### DIFF
--- a/pyanaconda/bootloader.py
+++ b/pyanaconda/bootloader.py
@@ -24,7 +24,7 @@ import re
 import blivet
 from parted import PARTITION_BIOS_GRUB
 from glob import glob
-from itertools import chain, zip_longest
+from itertools import chain
 import crypt
 from ordered_set import OrderedSet
 
@@ -35,7 +35,7 @@ from pyanaconda.product import productName
 from pyanaconda.flags import flags, can_touch_runtime_system
 from blivet.fcoe import fcoe
 import pyanaconda.network
-from pyanaconda.errors import errorHandler, ERROR_RAISE, ZIPLError, FirmwareCompatError
+from pyanaconda.errors import errorHandler, ERROR_RAISE, ZIPLError
 from pyanaconda.nm import nm_device_hwaddress
 from pyanaconda import platform
 from blivet.size import Size
@@ -97,34 +97,6 @@ def _is_on_ibft(device):
     """Tells whether a given device is ibft disk or not."""
 
     return all(getattr(disk, "ibft", False) for disk in device.disks)
-
-def _get_petitboot_version():
-    """
-    get the version of the petitboot loader that OF expresses as:
-        v1.6.1-pd8c7a0a
-    in the form:
-        ('v1.6.1-pd8c7a0a', (1, 6, 1), 'pd8c7a0a')
-    """
-    path = "/sys/firmware/devicetree/base/ibm,firmware-versions/petitboot"
-    if not os.access(path, os.F_OK):
-        raise FileNotFoundError
-    for line in open(path).readlines():
-        if line.startswith('v'):
-            line = line[1:]
-        (components, digest) = line.split('-')
-        components = [int(x) for x in components.split('.')]
-        return (line, tuple(components), digest)
-
-    return (None, (), None)
-
-def _version_ge(version, goal):
-    """
-    Compare two version tuples and tell you if version is >= goal
-    """
-    for ver0, ver1 in zip_longest(version, goal, fillvalue=0):
-        if ver0 != ver1:
-            return ver0 > ver1
-    return True
 
 
 class BootLoaderError(Exception):
@@ -1568,9 +1540,6 @@ class GRUB2(GRUB):
             defaults.write("GRUB_ENABLE_BLSCFG=true\n")
         defaults.close()
 
-    def _test_firmware_compat(self):
-        """ Check that this platform is configured in a compatible way """
-
     def _encrypt_password(self):
         """ Make sure self.encrypted_password is set up properly. """
         if self.encrypted_password:
@@ -1683,9 +1652,6 @@ class GRUB2(GRUB):
         if self.update_only:
             self.update()
             return
-
-        if flags.blscfg:
-            self._test_firmware_compat()
 
         try:
             self.write_device_map()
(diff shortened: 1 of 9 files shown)


pyanaconda/bootloader.py
```diff
@@ -98,6 +98,35 @@ def _is_on_ibft(device):
 
     return all(getattr(disk, "ibft", False) for disk in device.disks)
 
+def _get_petitboot_version():
```
poncovka: I think that this method is never called.

pyanaconda/bootloader.py
```diff
@@ -98,6 +98,35 @@ def _is_on_ibft(device):
...
+    for line in open(path).readlines():
+        if line.startswith('v'):
+            line = line[1:]
+        (components, digest) = line.split('-')
+        components = [int(x) for x in components.split('.')]
+        return (line, tuple(components), digest)
+
+    return (None, (), None)
+
+def _version_ge(version, goal):
```
poncovka: Same here.

poncovka:
Jenkins, test this please.

pyanaconda/bootloader.py
martinezjavier: @poncovka you are right, I've pushed a new version that also removes those two.

poncovka:
Jenkins, test this please.

pyanaconda/payload/livepayload.py
```diff
@@ -166,14 +166,25 @@ def install(self):
...
+                util.execInSysroot("new-kernel-pkg",
+                                   ["--rpmposttrans", kernel])
+            else:
+                files = glob.glob(util.getSysroot() + "/etc/kernel/postinst.d/*")
+                srlen = len(util.getSysroot())
+                files = sorted([f[srlen:] for f in files
+                                if os.access(f, os.X_OK)])
+                for file in files:
+                    util.execInSysroot(file,
+                                       [kernel, "/boot/vmlinuz-%s" % kernel])
```
keszybz: The standard way to install kernels is to invoke `/bin/kernel-install`. `kernel-install` was written by dracut author to provide a uniform interface for kernel installation and so that random tools don't try to implement all the logic. Finding and calling the `postinst.d` scripts like that looks very iffy. @haraldh comments?

haraldh: Yes, the kernel-core package also calls ```kernel-install```:
```
preuninstall scriptlet (using /bin/sh):
/bin/kernel-install remove 4.18.14-200.fc28.x86_64 /lib/modules/4.18.14-200.fc28.x86_64/vmlinuz || exit $?
posttrans scriptlet (using /bin/sh):
/bin/kernel-install add 4.18.14-200.fc28.x86_64 /lib/modules/4.18.14-200.fc28.x86_64/vmlinuz || exit $?
```
Why call the scripts manually?

pyanaconda/bootloader.py
```diff
@@ -24,7 +24,7 @@
 import blivet
 from parted import PARTITION_BIOS_GRUB
 from glob import glob
-from itertools import chain
+from itertools import chain, zip_longest
```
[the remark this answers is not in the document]
martinezjavier: Indeed, I've removed it.

pyanaconda/payload/livepayload.py
martinezjavier: @keszybz @haraldh I agree for kernel installs, but in this case the code doesn't do a kernel install, it just executes `new-kernel-pkg --rpmposttrans`. In the `new-kernel-pkg` script that is:

```
rpmposttrans()
{
    local files
    local f
    files="/etc/kernel/postinst.d/*[^~] /etc/kernel/postinst.d/$version/*[^~]"
    for f in $files ; do
        [ -f $f ] && [ -x $f ] || continue
        $f $version $kernelImage
    done
}
```

And that's why the `/etc/kernel/postinst.d/` scripts are called directly when `new-kernel-pkg` doesn't exists.

I'm not that familiar with this part of the Anaconda code to comment if a whole `kernel-install` is suitable here instead. But that will certainly be a change in behaviour with respect to just calling `kernel-pkg --rpmposttrans` as it's the case when `new-kernel-pkg` is available.

I'll let @vathpela comment on this since he is more familiar with Anaconda and this is part of his patch.

martinezjavier:
> Looks good to me from the code perspective.

Is anything else needed to land this change in rawhide?

poncovka:
Hi, we are waiting for you based on the comment at https://github.com/rhinstaller/anaconda/pull/1545#issuecomment-427901321.

martinezjavier:
> Hi, we are waiting for you based on the comment at [#1545 (comment)](https://github.com/rhinstaller/anaconda/pull/1545#issuecomment-427901321).

@poncovka sorry, this is scattered over so many pull-requests that I forgot about this comment...

The change in `fedora-comps` isn't needed anymore since after discussing it with @vathpela we decided to keep `grubby` in the `@core` group and update the package to drop `new-kernel-pkg`.

So we will push a `grubby` build with those changes once the anaconda changes land in rawhide.

M4rtinK:
> So we will push a `grubby` build with those changes once the anaconda changes land in rawhide.

Is it safe to do it that way - eq. first new Anaconda is built and only then grubby ? We could coordinate more closely if not & not sure if chainbuild is doable distgit permission wise.

martinezjavier:
> Is it safe to do it that way - eq. first new Anaconda is built and only then grubby ? We could coordinate more closely if not & not sure if chainbuild is doable distgit permission wise.

Yes, both the grubby and anaconda changes need to happen together. I already pushed the grubby changes to dist-git but didn't build it yet, will do that once the anaconda changes are included.

M4rtinK:
> Yes, both the grubby and anaconda changes need to happen together. I already pushed the grubby changes to dist-git but didn't build it yet, will do that once the anaconda changes are included.

OK, so I'll do a new Anaconda build with this and will add a new comment to this PR once I start it, soi we can coordinate. :)

martinezjavier:
> OK, so I'll do a new Anaconda build with this and will add a new comment to this PR once I start it, soi we can coordinate. :)

Ok, thanks!

M4rtinK:
> > OK, so I'll do a new Anaconda build with this and will add a new comment to this PR once I start it, soi we can coordinate. :)
> 
> Ok, thanks!

Anaconda build running.

martinezjavier:
> Anaconda build running.

Great, thanks for the info, [grubby build](https://koji.fedoraproject.org/koji/buildinfo?buildID=1161013) finished. I'll ping @vathpela so he does the same for grub2 as well, to make sure  that we have all the latest fixes.

M4rtinK:
> Great, thanks for the info, [grubby build](https://koji.fedoraproject.org/koji/buildinfo?buildID=1161013) finished. I'll ping @vathpela so he does the same for grub2 as well, to make sure that we have all the latest fixes.
Nice! The [Anaconda build](https://koji.fedoraproject.org/koji/taskinfo?taskID=30706323) finished successfully in the meantime, so all should be in place from Anaconda side.

AdamWill:
Broken for live installs.

https://bugzilla.redhat.com/show_bug.cgi?id=1648472

PR #1671, opened 2018-10-24
Do not enable or disable updates-testing, use setting from repo file …

M4rtinK:
…(#1642089)

This is a F29 specific hotfix to drop code from Anaconda that enables or
disables updates-testing based on user input & isFinal flag.

With this change the updates-testing repository will be enabled or
disabled only depending on the "enabled" value in it's repo file.
It will not be possible to enable or disable it from the GUI.

This is a downstream patch for F29 only and a better solution is
expected to be worked out for F30+.

Resolves: rhbz#1642089

### DIFF
--- a/pyanaconda/payload/dnfpayload.py
+++ b/pyanaconda/payload/dnfpayload.py
@@ -907,12 +907,14 @@ class DNFPayload(payload.PackagePayload):
 
         if self._updates_enabled:
             self.enableRepo("updates")
-            if not constants.isFinal:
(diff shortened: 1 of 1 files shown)


M4rtinK:
Looks like the current behavior might be good enough once isFinal=True is set in bootstamp and thus no fix in Anaconda will be needed.

But let's keep this PR open till GA in case the hotfix is still needed.

jkonecny12:
I think you can close this PR now.

PR #1672, opened 2018-10-24
Remove flags from anaconda_logging

poncovka:
Remove flags from pyanaconda.anaconda_logging and make the module
more flexible and independent of the remaining code.

### DIFF
--- a/anaconda.py
+++ b/anaconda.py
@@ -296,8 +296,8 @@ if __name__ == "__main__":
     # Set up logging as early as possible.
     from pyanaconda import anaconda_logging
     from pyanaconda import anaconda_loggers
-    anaconda_logging.init()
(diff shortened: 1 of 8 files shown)


anaconda.py
```diff
@@ -296,8 +296,8 @@ def start_debugger(signum, frame):
     # Set up logging as early as possible.
     from pyanaconda import anaconda_logging
     from pyanaconda import anaconda_loggers
-    anaconda_logging.init()
-    anaconda_logging.logger.setupVirtio()
+    anaconda_logging.init(not flags.imageInstall and not flags.dirInstall)
```
jkonecny12: Could you please make more obvious what the parameters means here? 
Maybe use keyword parameter or store this to a variable with name.

poncovka:
Updated.

PR #1681, opened 2018-11-02
Hotfix - disable unsupported hardware detection

M4rtinK:
There are apparently cases where the unsupported hardware
detection algorithm might be a bit too eager, resulting
in unfavourable user experience.

So let's disable unsupported hardware detection for now until the
algorithm is fixed.


M4rtinK:
Seems to work fine in both TUI & GUI - the message will not show up with the patch applied.

PR #1683, opened 2018-11-06
Workaround: Systemd-tmpfiles fails to create files/directories with ostreesetup, when the user of the directory is missing in live environment

adityashah1212:
systemd-tmpfiles queries nss to resolve users and groups. nss doesn't have --root flag or equivalent, hence the resolution happens in context of live environment. This causes systemd-tmpfiles to fail, if a user required by the installation is missing from the live environment. So as a work around we copy the passwd and group files from the ostree deployment (system users only) then run systemd-tmpfiles and immediately revert to the original values to not cause any problems with the rest of the installation procedures.

I encountered this issue when creating a custom tree for centos 7.5.

The following issue talks about the systemd-tmpfiles problem https://github.com/systemd/systemd/issues/7032


jkonecny12:
Hello, please create a bug on a https://bugzilla.redhat.com and link this PR there. We can't do much without a bug on RHEL systems.
Thank you.

PR #1684, opened 2018-11-06
Define the default partitioning statically

poncovka:
In the install classes, the setDefaultPartitioning method was replaced
with the default_partitioning attribute. The attribute provides a list
of partitioning requirements that will be automatically completed with
platform-specific requirements and user data.

Most of the code was moved to the pyanaconda.storage module.

### DIFF
--- a/pyanaconda/storage/partitioning.py
+++ b/pyanaconda/storage/partitioning.py
@@ -1,0 +1,147 @@
+#
+# Copyright (C) 2018  Red Hat, Inc.
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# the GNU General Public License v.2, or (at your option) any later version.
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY expressed or implied, including the implied warranties of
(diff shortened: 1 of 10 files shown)


pyanaconda/storage/partitioning.py
```diff
@@ -0,0 +1,146 @@
...
+        if auto_part_proxy.NoBoot:
+            skipped_mountpoints.add("/boot")
+
+        # Remove swap if --noswap is selected.
+        if auto_part_proxy.NoSwap:
+            skipped_fstypes.add("swap")
+
+            # Swap will not be recommended by the storage checker.
+            from pyanaconda.storage_utils import storage_checker
+            storage_checker.add_constraint(STORAGE_SWAP_IS_RECOMMENDED, False)
```
jkonecny12: I can help myself but I see this as a big side effect. Can you please change this to make this change more on purpose?

poncovka: This function will have to be changed anyway once it is moved to the storage module. I have added a `TODO` comment.

jkonecny12: Thanks, that is enough for me.

PR #1686, opened 2018-11-06
Replace setNetworkOnbootDefault

poncovka:
The method setNetworkOnbootDefault was removed from the install classes
and replaced with the attribute network_on_boot. The logic from the
install classes was moved to pyanaconda.network.

### DIFF
--- a/pyanaconda/network.py
+++ b/pyanaconda/network.py
@@ -35,6 +35,7 @@ from uuid import uuid4
 import itertools
 import glob
 import logging
+from enum import Enum
 
 from pyanaconda.simpleconfig import SimpleConfigFile
 from blivet.devices import FcoeDiskDevice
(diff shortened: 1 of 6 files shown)


pyanaconda/network.py
```diff
@@ -64,6 +65,13 @@
 network_connected_condition = threading.Condition()
 
 
+class NetworkOnBoot(Enum):
+    """The default network on boot."""
```
rvykydal: """Device to be activated on boot if none was configured so."""
... don't like it much but can't come up with better.

jkonecny12: I would add `Network devices...`

pyanaconda/network.py
```diff
@@ -1836,3 +1844,45 @@ def is_ibft_configured_device(iface):
...
+
+        dev = default_route_device() or default_route_device(family="inet6")
+        if not dev:
+            return
+
+        # ignore wireless (its ifcfgs would need to be handled differently)
+        if nm.nm_device_type_is_wifi(dev):
+            return
+
+        update_onboot_value(dev, True, ksdata=ksdata)
```
jkonecny12: Could you please move the `if-elif` content to a separated private methods? I would be easier to read.

pyanaconda/network.py
poncovka: Fixed.

pyanaconda/network.py
poncovka: Fixed.

poncovka:
Updated.

poncovka:
Rebased and tested.

PR #1688, opened 2018-11-09
Re-generate BLS loader file snippets on live installs (#1648472)

AdamWill:
Currently, on live installs, the BLS loader snippets that are
part of the live image itself - in /boot/loader/entries - are
simply rsync'ed into the installed system along with everything
else. This is no good in most cases because they point to
`/boot/vmlinuz-(foo)` and `/boot/initramfs-(foo)` (since the
live image itself does not have a separate /boot partition),
whereas most installs will have a separate /boot partition so
the path specified in the config snippet should not include
`/boot`. (The files are also named for the machine ID that was
used during live image generation, which will not be the machine
ID of the installed system; I'm not sure exactly what the
consequences of this are, probably nothing major but it may be
somehow an issue when the kernel is removed, perhaps.)

To fix this, we should exclude these files from the rsync, and
instead do something to have them re-generated after the rsync
has run. There are a few different ways and places we could do
this, but my choice is to run `kernel-install` in live payload
postInstall. This seems like it should be fairly robust. In
testing it does the job, both BIOS and UEFI installs that I
tested installed cleanly and booted successfully. Note that
`kernel-install` must run *after* `/etc/machine-id` exists in
the install root, or it will bail immediately and do nothing.
Note that at present, the 'Generating rescue image' step in
`install()` actually results in the generation of a machine ID,
so this could be done any time after that, but doing it after
the explicit failsafe machine ID generation in `postInstall()`
seems safest.

### DIFF
--- a/pyanaconda/payload/livepayload.py
+++ b/pyanaconda/payload/livepayload.py
@@ -48,6 +48,7 @@ from pyanaconda.anaconda_loggers import get_packaging_logger
 log = get_packaging_logger()
 
 from pyanaconda.errors import errorHandler, ERROR_RAISE
+from pyanaconda.flags import flags
 from pyanaconda.progress import progressQ
 from blivet.size import Size
 import blivet.util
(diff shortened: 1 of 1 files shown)


pyanaconda/payload/livepayload.py
```diff
@@ -143,6 +144,7 @@ def install(self):
         # file system boundaries
         args = ["-pogAXtlHrDx", "--exclude", "/dev/", "--exclude", "/proc/",
                 "--exclude", "/sys/", "--exclude", "/run/", "--exclude", "/boot/*rescue*",
+                "--exclude", "/boot/loader/", "--exclude", "/boot/efi/loader/",
```
jkonecny12: Will this work if `inst.noblscfg`  is used?

martinezjavier: @jkonecny12 yes, because with `inst.noblscfg` the `/boot/loader` directory is not used.

jkonecny12: OK, thanks for answer!

PR #1690, opened 2018-11-13
Remove configurePayload

poncovka:
The method configurePayload was removed from the install classes.
It is used only by the Scientific Linux to add optional repositories.
We will not support this feature in the Anaconda configuration file,
but we are going to find a different way.

### DIFF
--- a/anaconda.py
+++ b/anaconda.py
@@ -744,8 +744,6 @@ if __name__ == "__main__":
     from pyanaconda import exception
     anaconda.mehConfig = exception.initExceptionHandling(anaconda)
 
-    anaconda.postConfigureInstallClass()
-
     # add additional repositories from the cmdline to kickstart data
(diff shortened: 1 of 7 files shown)


poncovka:
@jcpunk, as we are continuing to work on the Anaconda modularization, we are currently working on cleaning up the install classes and basically turning them into a declarative configuration file. To make this possible we are going to temporarily remove the support for optional repositories from the upstream Anaconda codebase (so just on Fedora).

However, this should be only temporary and we are going to find a way how to bring the support back in a different way. We are considering to load the optional repositories from repo files in a dedicated directory. Would that work for you?

jcpunk:
That sounds viable.  My primary usage is with the RHEL stream.  So long as it stays "basically the same" for the release over there, I'm fine.

poncovka:
We will implement the alternative solution in another pull request. Merging.

PR #1691, opened 2018-11-15
Replace filterSupportedLangs and filterSupportedLocales

poncovka:
The methods for filtering supported languages and locales were removed
from the install classes and replaced with methods is_language_supported
and is_locale_supported in the Payload class. The logic from the install
classes was moved to pyanaconda.localization.


pyanaconda/payload/__init__.py
```diff
@@ -559,6 +559,21 @@ def verifyAvailableRepositories(self):
...
     ###
+    def is_language_supported(self, language):
+        """Is the given language supported by the payload?
+
+        :param language: a name of the language
+        """
+        return True
+
+    def is_locale_supported(self, language, locale):
+        """Is the given locale supported by the payload:
```
jkonecny12: I guess that `:` should be `?`.

poncovka: Fixed.

poncovka:
Rebased and updated.

poncovka:
@cgwalters and @dustymabe, as we are continuing to work on the Anaconda modularization, we are currently working on cleaning up the install classes and basically turning them into declarative configuration files. To make this possible, we need to do something with methods for filtering supported languages and locales in the class `AtomicHostInstallClass`.

Solution A is to enable this filtering for all rpmostree-based installations. Then the code could be part of the class `RPMOSTreePayload` and we don't have to deal with this in the install classes. However, I don't know if it is a desirable change.

Solution B is to keep the filtering optional as it is now. The code could be part of a new object and the install classes could choose the filter via a variable. This solution makes sense if the filter should be really optional and might be useful also for other products and variants. However, then we need to know a little more about the code, so we can better document it.

Could you please provide us some context about the language and locale filtering in the Atomic Host? Why it is required in the Atomic Host? Why is it not required in other rpmostree-based variants?

cgwalters:
> Why it is required in the Atomic Host? 

https://pagure.io/atomic-wg/issue/408
https://pagure.io/atomic-wg/issue/282

> Why is it not required in other rpmostree-based variants?

For Silverblue...it's already large, dropping languages doesn't save much.

Personally I'm OK if support for rpm-ostree locale filtering is dropped from Anaconda - we aren't going to be using Anaconda for Fedora CoreOS, and Silverblue doesn't need it.

poncovka:
@cgwalters, thanks! Based on your reply I have implemented something between solutions A and B. The code is part of the payload class, but install classes can still choose to skip the filtering. This will not change the current behaviour. Anyway, as soon as you decide to drop this code, we will be happy to remove it.

poncovka:
Tested with Fedora Server, Atomic Host and Silverblue.

pyanaconda/ui/gui/spokes/lib/lang_locale_handler.py
```diff
@@ -40,7 +40,7 @@ class LangLocaleHandler(object):
 
     """
 
-    def __init__(self, payload):
+    def __init__(self, data, storage, payload, instclass):
```
jkonecny12: Shouldn't it be better to just pass intclass and payload here?

poncovka: Fixed.

poncovka:
Jenkins, test this please.

PR #1692, opened 2018-11-16
Don't start TUI main loop if no screens are scheduled (#1646568)

M4rtinK:
We need to make sure the TUI main loop is only started if some screens
are scheduled.

This (no screens scheduled as there is nothing to do) is a valid
usecase for Initial Setup if the user has already been created,
there is no EULA to show and reconfig mode is not active.

Unfortunately instead of the expected behavior (do nothing and exit)
the main loop would still be started even with no screens, effectively
stopping the boot process while not showing anything on the console.

To avoid this scenario we need to check that there are some screens
scheduled before we start the TUI main loop.

Resolves: rhbz#1646568

### DIFF
--- a/pyanaconda/ui/tui/__init__.py
+++ b/pyanaconda/ui/tui/__init__.py
@@ -223,7 +223,19 @@ class TextUserInterface(ui.UserInterface):
         This should do little more than just pass through to something else's run method,
         but is provided here in case more is needed.  This method must be provided by all subclasses.
         """
-        return App.run()
+        # App.run() starts the Simpline managed GLib main loop, so we need to make sure
+        # it's only called if some screens are actually scheduled.
+        #
+        # While not likely to ever happen in Anaconda, this is a valid case for Initial Setup,
+        # where it can happen that no screens are needed to be shown and thus not scheduled
+        # and calling App.run() in such a case will effectively deadlock the boot process without
+        # showing anything on the console (prompt is provided by screens, if no screens are scheduled,
+        # not even a prompt is shown). By not calling App.run() Initial Setup will simply exit,
+        # which is the expected behavior as nothing needs to be done.
+        if App.get_scheduler().nothing_to_render:
+            log.info("TUI main loop will not be started as no TUI screens have been scheduled")
+        else:
+            return App.run()
 
     ###
     ### MESSAGE HANDLING METHODS


pyanaconda/ui/tui/__init__.py
```diff
@@ -223,7 +223,19 @@ def run(self):
...
+        # While not likely to ever happen in Anaconda, this is a valid case for Initial Setup,
+        # where it can happen that no screens are needed to be shown and thus not scheduled
+        # and calling App.run() in such a case will effectively deadlock the boot process without
+        # showing anything on the console (prompt is provided by screens, if no screens are scheduled,
+        # not even a prompt is shown). By not calling App.run() Initial Setup will simply exit,
+        # which is the expected behavior as nothing needs to be done.
+        if App.get_scheduler().nothing_to_render:
+            log.info("TUI main loop will not be started as no TUI screens have been scheduled")
+        else:
+            return App.run()
```
poncovka: @jkonecny12, why isn't this logic part of `App.run()`? I would expect the `run` method to quit if there are no screens scheduled.

jkonecny12: This is not there on purpose. The reason is that you can start the application and have a scheduler (futures, other thread...) add the screen later but your even loop will work during that time.

What I'm thinking of is that this could be more as a opt-in. So maybe it would be better to have this as a configuration option to start event loop if scheduler is empty. That may be an alternative fix for this issue. What do you think?

poncovka: Ok, but even in that case you want to show the user something, no? If you cannot schedule the screens before starting the app then you are waiting for something and you might at least show a message to the user to inform him and let him to quit the app if he wants to.

I am ok with the opt-in variant, but the default behaviour should be to quit the app.

jkonecny12: You really don't want to show anything which wasn't requested by a developer. If he is not able to remove that. And I don't want to add logic for this.
I don't see problem with the opt-in solution.

poncovka: I meant that it is not so unreasonable to require at least one scheduled screen. That's all :-) However, the opt-in solution gives more flexibility.

M4rtinK:
So in the end this has been handled in simpleline and Initial Setup, so this PR can be closed.

PR #1693, opened 2018-11-17
Install grubby-deprecated when using the extlinux bootloader

martinezjavier:
Most bootloaders used in Fedora already have BootLoaderSpec but extlinux
doesn't yet. So when using extlinux the old grubby must to be installed.

Resolves: rhbz#1649778


jkonecny12:
Jenkins, it's ok to test.

PR #1695, opened 2018-11-21
Convert a keymap into a list of layouts

poncovka:
We should return a list of layouts when we try to convert a keymap.


poncovka:
Updated the commit message.

poncovka:
Jenkins, test this please.

PR #1696, opened 2018-11-22
Remove help-related constants from install classes

poncovka:
It looks like we can remove some of the help-related variables from
the install classes and replace them with constants.


poncovka:
Rebased.

PR #1697, opened 2018-11-22
Remove setup_on_boot from the install classes

poncovka:
Let's remove the variable setup_on_boot from the install classes,
because it's value doesn't change.


poncovka:
Rebased.

PR #158, opened 2015-06-24
autopart root snapshot support

dwlehman:
After installing the system, take a snapshot of the root device and
reconfigure the system to use the snapshot as the root device.

This is only supported with thinp and btrfs autopart.

Resolves: rhbz#1113207

Related to rhinstaller/blivet/pull/166 and rhinstaller/pykickstart/pull/29

PR #1699, opened 2018-11-23
Drop xorg-x11-server-Xorg check from graphical target detection (#158…

M4rtinK:
…3958)

While this package will still likely continue to be installed on Wayland systems
due to xwayland, it should be enough that we check for a package that
provides "service(graphical-login)" to correctly detect if a graphical
system target should be used instead of multi-user.

This might fix some edge cases where users install x server for
on a system without the intent to ever log into the system
graphically (no package providing "service(graphical-login)" is
installed). Previously such users could erroneously get graphical
target set.

Resolves: rhbz#1583958

### DIFF
--- a/pyanaconda/payload/__init__.py
+++ b/pyanaconda/payload/__init__.py
@@ -894,7 +894,6 @@ class Payload(object):
 
             # XXX one day this might need to account for anaconda's display mode
             if ts.dbMatch("provides", 'service(graphical-login)').count() and \
-               ts.dbMatch('provides', 'xorg-x11-server-Xorg').count() and \
                not flags.usevnc:
                 # We only manipulate the ksdata.  The symlink is made later
                 # during the config write out.


halfline:
> This might fix some edge cases where users install x server for
on a system without the intent to ever log into the system
graphically (no package providing "service(graphical-login)" is
installed). Previously such users could erroneously get graphical
target set.

Don't think this is true. the conditional is `and` not `or` so graphical.target will only get set if both are there.  still the Xorg bit seems to be extraneous and maybe even wrong down the road in a fully waylandized world.

M4rtinK:
> > This might fix some edge cases where users install x server for
> > on a system without the intent to ever log into the system
> > graphically (no package providing "service(graphical-login)" is
> > installed). Previously such users could erroneously get graphical
> > target set.
> 
> Don't think this is true. the conditional is `and` not `or` so graphical.target will only get set if both are there. still the Xorg bit seems to be extraneous and maybe even wrong down the road in a fully waylandized world.

OK, so I'll just drop that part from the commit message.

PR #1701, opened 2018-11-27
Add can_write_network_config system property

rvykydal:
True on boot_iso and live_io (ie in live environments).

### DIFF
--- a/pyanaconda/core/configuration/anaconda.py
+++ b/pyanaconda/core/configuration/anaconda.py
@@ -180,6 +180,16 @@ class InstallationSystem(Section):
         return self._is_boot_iso or self._is_live_os
 
     @property
+    def provides_network_config(self):
+        """Can we copy network configuration to the target system?
+
+        We can do it only if the current system configuration is created by
+        anaconda (or installation process in general, as on Live OS) and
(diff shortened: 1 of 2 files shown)


pyanaconda/core/configuration/anaconda.py
```diff
@@ -179,6 +179,16 @@ def can_configure_network(self):
         """Can we configure the network?"""
         return self._is_boot_iso or self._is_live_os
 
+    @property
+    def can_write_network_config(self):
```
poncovka: Could the rule be called `provides_network_config`? It seems to be similar to for example `provides_resolver_config`. The current name suggests that we can write to the network configuration of the installation system and that is not true.

pyanaconda/core/configuration/anaconda.py
```diff
@@ -179,6 +179,16 @@ def can_configure_network(self):
...
+        """Can we pass network configuration to installed system?
```
poncovka: I would use `the target system` here.

rvykydal:
The suggestions by @poncovka applied.

rvykydal:
jenkins, test this please

PR #1704, opened 2018-11-27
dracut/parse-kickstart: don't abort on --device=link

lkundrak:
It never works. parse-anaconda-kickstart.sh generates command line options
and as such is launched from a cmdline hook. That is long before the
network devices could be brought up, or even their drivers are loaded.

Not aborting and generating a device-less "ip=dhcp" option is good enough:
NetworkManager only brings up devices with a carrier anyway.

This is needed to address https://bugzilla.redhat.com/show_bug.cgi?id=1641832


rvykydal:
external PR Jenkins

jkonecny12:
Jenkins, it's ok to test.

PR #1708, opened 2018-11-29
Disable BLS config if new-kernel-pkg is installed

martinezjavier:
The kernel-install scripts installs kernels and adds BLS snippets in the
/boot/loader/entries directory if new-kernel-pkg isn't present or a BLS
setup was explicility configured (i.e: setting GRUB_ENABLE_BLSCFG=true).

But the kernel package is installed before writing any bootloader config,
so the kernel-install script only rely on the existence of new-kernel-pkg
to choose if a BLS or non-BLS kernel installation will be made.

So if new-kernel-pkg is present and the inst.noblscfg boot option wasn't
used, then kernels will be installed without a BLS snippet but bootloader
config will expect them, leading to an unbootable installed system.

Resolves: rhbz#1654036

### DIFF
--- a/pyanaconda/bootloader.py
+++ b/pyanaconda/bootloader.py
@@ -268,6 +268,8 @@ class BootLoader(object):
         self._update_only = False
         self.skip_bootloader = False
 
+        self.use_bls = flags.blscfg
+
         self.errors = []
         self.warnings = []
 
@@ -1544,7 +1546,12 @@ class GRUB2(GRUB):
         defaults.write("GRUB_CMDLINE_LINUX=\"%s\"\n" % self.boot_args)
         defaults.write("GRUB_DISABLE_RECOVERY=\"true\"\n")
         #defaults.write("GRUB_THEME=\"/boot/grub2/themes/system/theme.txt\"\n")
-        if flags.blscfg:
+
+        if self.use_bls and os.path.exists(util.getSysroot() + "/usr/sbin/new-kernel-pkg"):
+            log.warning("BLS support disabled due new-kernel-pkg being present")
+            self.use_bls = False
+
+        if self.use_bls:
             defaults.write("GRUB_ENABLE_BLSCFG=true\n")
         defaults.close()
 
@@ -2325,7 +2332,7 @@ class ZIPL(BootLoader):
                 args.update(["rootflags=subvol=%s" % image.device.name])
             log.info("bootloader.py: used boot args: %s ", args)
 
-            if flags.blscfg:
+            if self.use_bls:
                 self.update_bls_args(image, args)
             else:
                 self.write_config_image(config, image, args)
@@ -2337,7 +2344,12 @@ class ZIPL(BootLoader):
                   "timeout={}\n"
                   "target=/boot\n")
         config.write(header.format(self.timeout))
-        if not flags.blscfg:
+
+        if self.use_bls and os.path.exists(util.getSysroot() + "/usr/sbin/new-kernel-pkg"):
+            log.warning("BLS support disabled due new-kernel-pkg being present")
+            self.use_bls = False
+
+        if not self.use_bls:
             config.write("default={}\n".format(self.image_label(self.default)))
 
     #


M4rtinK:
I think this issue is kinda similar to why Anaconda re-creates the initramdisk _after_ the package installation transaction. As package ordering is not deterministic it could very well happen that the kernel package scriptlet that triggers the Dracut image generation will run before essential files (such as keyboard layouts) are in places and they will be missing from the image.

I guess this will also happen on a regular system, but it is less likely to cause issues as there will usually be at least an older version of the files and there is obviously no previous version of any files in the initial Anaconda packages installation transaction.

AdamWill:
I think I'd like it if this logged a message. It's quite obscure behaviour, in a way. I just find the whole 'we configure the bootloader one way if a certain binary is installed, and a completely different way if it isn't' behaviour to be a bit jarring in general.

martinezjavier:
> I think I'd like it if this logged a message. It's quite obscure behaviour, in a way. I just find the whole 'we

Indeed, I should had done that. I'll re-spin the patches tomorrow. Thanks for the feedback!

M4rtinK:
> I think I'd like it if this logged a message. It's quite obscure behaviour, in a way. I just find the whole 'we configure the bootloader one way if a certain binary is installed, and a completely different way if it isn't' behaviour to be a bit jarring in general.

Good point. But I still hope we can hopefully simplify the bootloader code quite a bit once the BLS is more established and all the fallbacks and workarounds for can be dropped.

pyanaconda/bootloader.py
```diff
@@ -2337,6 +2341,10 @@ def write_config_header(self, config):
                   "timeout={}\n"
                   "target=/boot\n")
         config.write(header.format(self.timeout))
+
+        if os.path.exists(util.getSysroot() + "/usr/sbin/new-kernel-pkg"):
+            flags.blscfg = False
+
         if not flags.blscfg:
```
poncovka: Is there a reason why the flag needs to be modified? Could you just change the condition to:
`if not flags.blscfg or os.path.exists(util.getSysroot() + "/usr/sbin/new-kernel-pkg"):`

I am afraid of the side effects.

pyanaconda/bootloader.py
```diff
@@ -1544,6 +1544,10 @@ def write_defaults(self):
         defaults.write("GRUB_CMDLINE_LINUX=\"%s\"\n" % self.boot_args)
         defaults.write("GRUB_DISABLE_RECOVERY=\"true\"\n")
         #defaults.write("GRUB_THEME=\"/boot/grub2/themes/system/theme.txt\"\n")
+
+        if os.path.exists(util.getSysroot() + "/usr/sbin/new-kernel-pkg"):
+            flags.blscfg = False
+
         if flags.blscfg:
```
poncovka: Same here.

martinezjavier:
@AdamWill @M4rtinK I've been thinking a little more about this. Should we disable BLS and print something or fail the installation completely if `ins.blscfg` is set and `grubby-deprecated` was installed?

Telling the user to use `inst.noblscfg` instead?

pyanaconda/bootloader.py
martinezjavier: @poncovka thanks a lot for your feedback. The reason why was disabled was to prevent having the same check done multiple times, for example in the case of `ZIPL` class, the `flags.blscfg` are used in both `write_config_header()` and `write_config_images()`.

But you are correct about the possibly side effects, I'll re-spin the patch just adding a check for  `new-kernel-pkg` existence on every place where `flags.blscfg` is checked.

martinezjavier:
@poncovka @AdamWill @M4rtinK I've pushed a new version of the patch that I think that addresses the issues pointed out by you.

Changes in v2:
- Log a warning message that BLS is getting disabled due new-kernel-pkg being present.
- Don't modify `flags.blscfg`.

martinezjavier:
> I am really looking forward to the removal of the flag.

Indeed, me too.

AdamWill:
I also thought about whether setting the flag was correct, but I actually thought rather it was. After all, if *this* code behaves as if BLS is disabled, surely all *subsequent* actions should *also* behave as if BLS is disabled? Do we actually ever *want* the scenario where this code behaves in non-BLS fashion, but later code reads `flags.blscfg` as `True` and acts in BLS fashion?

AdamWill:
> > I am really looking forward to the removal of the flag.
> 
> Indeed, me too.

Me three.

In fact...do we really want/need it? Who requested that it be added, and do they still want it? Can we maybe get rid of it now?

martinezjavier:
> I also thought about whether setting the flag was correct, but I actually thought rather it was. After all, if _this_ code behaves as if BLS is disabled, surely all _subsequent_ actions should _also_ behave as if BLS is disabled? Do we actually ever _want_ the scenario where this code behaves in non-BLS fashion, but later code reads `flags.blscfg` as `True` and acts in BLS fashion?

Yes, nothing should behave as if BLS was enabled after the bootloader was configured without BLS support. In fact, the problem with rhbz#1654036 was exactly this mismatch between how the kernels were installed (without BLS support) and the bootloader config (with BLS support).

But what I understood from @poncovka comment was that it's a bad pattern in Anaconda to modify global flags in general.

martinezjavier:
> In fact...do we really want/need it? Who requested that it be added, and do they still want it? Can we maybe get rid of it now?

I don't remember now who asked for it, but it was part of the contingency plan in the [Changes page](https://fedoraproject.org/wiki/Changes/BootLoaderSpecByDefault) and also to allow users to do a non-BLS installation if they didn't want to use BLS.

poncovka:
Jenkins, test this please.

poncovka:
My mistake, sorry.

AdamWill:
"I don't remember now who asked for it, but it was part of the contingency plan in the Changes page"

I don't see it there currently, at least...

" and also to allow users to do a non-BLS installation if they didn't want to use BLS."

http://islinuxaboutchoice.com/ :P

martinezjavier:
> I don't see it there currently, at least...

I think it was implicit in "Contingency mechanism: Revert the anaconda changes".

As said, I don't mind dropping the option, @vathpela wrote the Anaconda patches so I would like him to decide though.

AdamWill:
AIUI, what is meant by 'contingency' for the Change process is "this is what we do *if we decide to cancel the Change* - it's the *backup plan*. It doesn't mean you have to implement both choices...simultaneously. To me, that 'contingency mechanism' just means "if we decide not to have this Change in Fedora XX, we revert any related anaconda commits so anaconda *only* will do non-BLS installs" - and indeed that's more or less what happened for F29 (we rejected the anaconda commits, thus effectively rejecting the Change and invoking the 'contingency plan' by...not changing anything :>).

It doesn't mean "we should make sure anaconda can do both BLS and non-BLS installs". Still, I agree we should get @vathpela 's opinion indeed.

vathpela:
I don't think the inst.blscfg flag actually serves any currently meaningful purpose.

martinezjavier:
@AdamWill @poncovka I've proposed PR #1715 that drops the `inst.noblscfg` option.

The idea is to use as criterion to decide whether a BLS configuration would be used, the existence of the `new-kernel-pkg` script instead. That is consistent to what the kernel-install scripts do and won't leave the system configured as no BLS but with a broken kernel update mechanism.

I tested that with the following setups:

* The default install when `grubby-deprecated` isn't installed so it defaults to BLS.
* A kickstart that installs `grubby-deprecated` package so BLS support is disabled.

PR #1711, opened 2018-11-29
Fix module installation error reporting (#1644635)

M4rtinK:
Starting with the 3.6 release DNF now provides much improved
module installation error reporting & includes it in the
string representation of the relevant exception it raises.

So drop our custom error reporting code and just use the
string representation of the DNF exception.

Also adjust the module enable try-block to catch the new exception
and submit all requests for enabling modules at once so we get
a combined error for all broken modules at once.

Backported from commit 1caf765733a1cdcb58e9227047c2468725bdbf13 and
has been in use in Fedora since F29.

Resolves: rhbz#1644635

### DIFF
--- a/anaconda.spec.in
+++ b/anaconda.spec.in
@@ -23,7 +23,7 @@ Source0: %{name}-%{version}.tar.bz2
 # match the requires versions of things).
 
 %define dbusver 1.2.3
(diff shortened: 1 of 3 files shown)


M4rtinK:
jenkins, test this please

PR #1712, opened 2018-11-29
Ensure grubby-deprecated is installed if blscfg is disabled

AdamWill:
If blscfg is diasbled (with e.g. inst.noblscfg), we must install
grubby-deprecated or else the installed system will not be able
to update the bootloader configuration properly when the kernel
is updated.

### DIFF
--- a/pyanaconda/bootloader.py
+++ b/pyanaconda/bootloader.py
@@ -1437,9 +1437,6 @@ class GRUB2(GRUB):
 
     """
     name = "GRUB2"
-    # grub2 is a virtual provides that's provided by grub2-pc, grub2-ppc64le,
-    # and all of the primary grub components that aren't grub2-efi-${EFIARCH}
-    packages = ["grub2", "grub2-tools"]
     _config_file = "grub.cfg"
     _config_dir = "grub2"
     _passwd_file = "user.cfg"
@@ -1455,6 +1452,16 @@ class GRUB2(GRUB):
 
     # XXX we probably need special handling for raid stage1 w/ gpt disklabel
     #     since it's unlikely there'll be a bios boot partition on each disk
+
+    @property
+    def packages(self):
+        # grub2 is a virtual provides that's provided by grub2-pc, grub2-ppc64le,
+        # and all of the primary grub components that aren't grub2-efi-${EFIARCH}
+        packages = ["grub2", "grub2-tools"]
+        # non-BLS installs require grubby-deprecated for kernel updates to work
+        if not flags.blscfg:
+            packages.append("grubby-deprecated")
+        return packages
 
     @property
     def stage2_format_types(self):


AdamWill:
@vathpela @martinezjavier please check.

I believe this should also go to the RHEL 8 branch. I've checked and RHEL 8 has the same setup as F30, with grubby-deprecated.

AdamWill:
Without this patch, if you install with `inst.noblscfg`, you'll get some files in `/boot/loader/entries` (not sure why), but `/boot/grub2/grub.cfg` is in the old, non-BLS format (as intended)...but grubby-deprecated is not installed, so when a kernel update comes out, that old, non-BLS grub2.cfg isn't going to get updated properly, at least AIUI.

martinezjavier:
> I believe this should also go to the RHEL 8 branch. I've checked and RHEL 8 has the same setup as F30, with grubby-deprecated.

I'm not sure about RHEL8, I remember we had a conversation with the Anaconda folks that the `inst.noblscfg` would not even be a documented option there, and the supported configuration is BLS.

pyanaconda/bootloader.py
```diff
@@ -1456,6 +1453,18 @@ class GRUB2(GRUB):
     # XXX we probably need special handling for raid stage1 w/ gpt disklabel
     #     since it's unlikely there'll be a bios boot partition on each disk
 
+    # this is a property because the state of flags.blscfg changes
+    # during anaconda run, so it can't just be a class attribute
```
poncovka: The flag shouldn't change during the Anaconda run. It should be set before this module is imported. If you have in mind the pull request https://github.com/rhinstaller/anaconda/pull/1708, that code is run after the package installation, so this property shouldn't be used after that (at least I think so).

martinezjavier:
> ithout this patch, if you install with `inst.noblscfg`, you'll get some files in `/boot/loader/entries` (not sure why), but `/boot/grub2/grub.cfg` is in the old, non-BLS format (as intended)...but grubby-deprecated is not installed, so when a kernel update comes out, that old, non-BLS grub2.cfg isn't going to get updated properly, at least AIUI.

The reason why BLS snippets are added even when the `inst.noblscfg` is used, is because the `kernel-install` scripts adds BLS entries if either `GRUB_ENABLE_BLSCFG=true` was set in `/etc/default/grub` or `new-kernel-pkg` ins't present.

Since the `grubby-deprecated` package isn't installed as a part of the `@core` package group, then `kernel-install` will add the BLS entries.

Later at the end of the installation, `anaconda` will check if `inst.noblscfg` was set and in that case won't set `GRUB_ENABLE_BLSCFG=true` and so `grub2-mkconfig` will generate a non-BLS `grub.cfg` picking the kernels from `/boot` that were copied by `kernel-install`.

Note that this is the same behavior before the BLS changes, `kernel-install` used to call `new-kernel-pkg`, that then called `grubby` that would fail to add menu entries to `grub.cfg` (because the file didn't exist yet), but it doesn't matter because at the end of the installation `grub2-mkconfig` would genrate a `grub.cfg` with the installed kernels.

On the other hand, if `inst.blscfg` is set (the default) then `GRUB_ENABLE_BLSCFG=true` will be set and `grub2-mkconfig` will generate a BLS `grub.cfg` (without boot entries and just calling the `blscfg` command).

So I think your patch is correct, because now `grubby-deprecated` will be installed and so the `kernel-install` scripts won't add BLS entries since `new-kernel-pkg` would be present.

pyanaconda/bootloader.py
AdamWill: I actually tested it, and nope. If I just put this conditional in the setting of the class variable, it doesn't work. You can try it for yourself easily enough - just generate an updates.img with that version of the change, and try an install of https://kojipkgs.fedoraproject.org/compose/rawhide/Fedora-Rawhide-20181130.n.0/compose/Server/x86_64/iso/Fedora-Server-dvd-x86_64-Rawhide-20181130.n.0.iso with `inst.noblscfg`. It doesn't install `grubby-deprecated`. With this version of the change, it does.

I agree it's unexpected, but that's the behaviour I experienced in testing. I can only assume that somehow `bootloader.py` gets imported *before* the code in `anaconda.py` that sets `flags.blscfg` is run.

AdamWill:
> > I believe this should also go to the RHEL 8 branch. I've checked and RHEL 8 has the same setup as F30, with grubby-deprecated.
> 
> I'm not sure about RHEL8, I remember we had a conversation with the Anaconda folks that the `inst.noblscfg` would not even be a documented option there, and the supported configuration is BLS.

I guess I'd say if the code actually exists (even if it's not documented), it should behave correctly. If it's actually been removed from RHEL 8, then no need of course.

pyanaconda/bootloader.py
AdamWill: Well, I suppose it's possible that's to do with the updates mechanism, thinking about it (though I thought updates were applied during initramfs phase these days). I'll test it with a new image rather than just an updates.img.

poncovka: I am ok with the property, but the comment is not necessary and just confusing.

AdamWill: I'm going to re-test with custom images containing patched anaconda package builds to verify whether the property is really needed. If it is, I'll tweak the comment somehow.

AdamWill: So I tested, and it really does seem to need to be a property (or I guess maybe an instance variable? didn't test that yet). It just doesn't work if it's a class variable.

I'll twiddle the comment a bit I guess.

AdamWill:
Tweaked the comment and commit message.

poncovka:
Jenkins, test this please.

poncovka:
The pull request can be merged, but I would wait for the conclusion of the conversation at https://github.com/rhinstaller/anaconda/pull/1708.

AdamWill:
sure, sounds sensible.

poncovka:
The pull request https://github.com/rhinstaller/anaconda/pull/1715 was merged. I am closing this one.

PR #1715, opened 2018-12-05
Drop the inst.noblscfg option

martinezjavier:
This option was added to allow installing a system configured without BLS
support. But the option is misleading because even when the bootloader is
configured without BLS, the grubby-deprecated package must be installed
in order to install new kernels and add entries in the bootloader config.

The kernel-install scripts already use the existence of the new-kernel-pkg
to decide if a BLS snippet will be added or not when a kernel is installed
and also is used to configure a bootloader with BLS support or not.

So the inst.noblscfg option doesn't really serve a purpose at this point
and can just be dropped. Instead, a user can choose to avoid configuring
the system with BLS by installing the grubby-deprecated package.

### DIFF
--- a/anaconda.py
+++ b/anaconda.py
@@ -426,7 +426,6 @@ if __name__ == "__main__":
     flags.rescue_mode = opts.rescue
     flags.noverifyssl = opts.noverifyssl
     flags.extlinux = opts.extlinux
-    flags.blscfg = opts.blscfg
     flags.nombr = opts.nombr
     flags.debug = opts.debug
     flags.askmethod = opts.askmethod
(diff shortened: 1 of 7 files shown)


AdamWill:
:+1: LGTM.

poncovka:
Jenkins, test this please.

poncovka:
Please, fix the warning from pylint:

```
************* Module ../pyanaconda/payload/livepayload.py
W0611(unused-import): 51,0: : Unused flags imported from pyanaconda.flags
```

martinezjavier:
> Please, fix the warning from pylint:

@poncovka done, thanks a lot for your feedback.

poncovka:
Jenkins, test this please.

AdamWill:
I'd suggest this should go to RHEL 8 too, if the feature is still in that branch.

jkonecny12:
@AdamWill look here: https://github.com/rhinstaller/anaconda/pull/1728

PR #1716, opened 2018-12-05
Remove install classes from Anaconda

poncovka:
Install classes are completely replaced with configuration files and
removed from the code.

The kickstart command `installclass` is removed and replaced
with boot options `inst.product` and `inst.variant`.


pyanaconda/ui/gui/__init__.py
```diff
@@ -56,10 +56,10 @@
 ANACONDA_WINDOW_GROUP = Gtk.WindowGroup()
 
 # Stylesheet priorities to use for product-specific stylesheets.
-# installclass stylesheets should be higher than our base stylesheet, and
+# custom stylesheets should be higher than our base stylesheet, and
```
rvykydal: custom -> Custom

pyanaconda/core/configuration/payload.py
```diff
@@ -0,0 +1,68 @@
...
+    def default_environment(self):
+        """Default package environment."""
+        return self._get_option("default_environment", str)
+
+    @property
+    def ignored_packages(self):
+        """List of ignored packages.
+
+        Anaconda flags several packages to be installed based on the configuration
+        of the system -- things like fs utilities, boot loader, &c. This is a list
```
M4rtinK: &c == etc. or a typo ? :)

pyanaconda/core/configuration/payload.py
```diff
@@ -0,0 +1,68 @@
...
+        mechanism.
+        """
+        return self._get_option("ignored_packages", str).split()
+
+    @property
+    def enable_updates(self):
+        """Enable installation of latest updates.
+
+        This flag controls whether or not Anaconda should provide an option to
+        install the latest updates during installation source selection.
```
M4rtinK: It sounds unclear to me from the docstring if this option just shows an otherwise invisible checkbox for enabling updates or if it also enables updates themselves (eq. the checkbox is checked by default). Also, how does this influence a kickstart installation ? So it might be good to clarify this in the docstring.

pyanaconda/core/configuration/payload.py
```diff
@@ -0,0 +1,68 @@
...
+        """
+        return self._get_option("enable_updates", bool)
+
+    @property
+    def enable_closest_mirror(self):
+        """Enable installation from the closest mirror.
+
+        A hint if mirrors are expected to be available for the distribution
+        installed by the given install class. At the moment this just used
```
M4rtinK: I think we should change "install class" to something else there (product/variant/config/specification) ?

PR #1717, opened 2018-12-05
Add comment to /etc/default/grub explaining how cmdline can be set wi…

martinezjavier:
…th BLS

On a BLS configuration the menu entries aren't defined in the grub.cfg but
instead defined in BLS snippets in the /boot/loader/entries directory. So
re-generating the grub.cfg with grub2-mkconfig has no effect on the menu
entries or their kernel command line parameters.

That's why modifying GRUB_CMDLINE_LINUX in /etc/default/grub and executing
grub2-mkconfig can't be used anymore to modify the entries kernel cmdline.

This can be confusing for users that are used to those steps to modify the
cmdline, so let's add some comments in /etc/default/grub explaining that's
not supported anymore and how the cmdline can be modified on a BLS setup.

### DIFF
--- a/pyanaconda/bootloader.py
+++ b/pyanaconda/bootloader.py
@@ -1539,10 +1539,18 @@ class GRUB2(GRUB):
         else:
             defaults.write("GRUB_TERMINAL_OUTPUT=\"%s\"\n" % self.terminal_type)
 
+        bls_comment = ("# If GRUB_ENABLE_BLSCFG is enabled, GRUB_CMDLINE_LINUX is only used to set\n"
+                       "# the kernelopts environment variable defined in /boot/grub2/grubenv once.\n"
+                       "# If kernelopts is already set then grub2-mkconfig will ignore the value in\n"
+                       "# GRUB_CMDLINE_LINUX. To edit kernel command line parameters the kernelopts\n"
+                       "# variable needs to be modified using the grub2-editenv tool.\n")
+
         # this is going to cause problems for systems containing multiple
         # linux installations or even multiple boot entries with different
         # boot arguments
         log.info("bootloader.py: used boot args: %s ", self.boot_args)
+        if self.use_bls:
+            defaults.write(bls_comment)
         defaults.write("GRUB_CMDLINE_LINUX=\"%s\"\n" % self.boot_args)
         defaults.write("GRUB_DISABLE_RECOVERY=\"true\"\n")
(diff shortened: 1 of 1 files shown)


AdamWill:
this seems like a good idea.

poncovka:
Jenkins, test this please.

pyanaconda/bootloader.py
```diff
@@ -1539,10 +1539,18 @@ def write_defaults(self):
         else:
             defaults.write("GRUB_TERMINAL_OUTPUT=\"%s\"\n" % self.terminal_type)
 
+        bls_comment = ("# If GRUB_ENABLE_BLSCFG is enabled, GRUB_CMDLINE_LINUX is only used to set\n"
+                       "# the kernelopts environment variable defined in /boot/grub2/grubenv once.\n"
+                       "# If kernelopts is already set then grub2-mkconfig will ignore the value in\n"
+                       "# GRUB_CMDLINE_LINUX. To edit kernel command line parameters the kernelopts\n"
+                       "# variable needs to be modified using the grub2-editenv tool.\n")
```
kyle-walker: Could we get another section right below this one?

    #
    # To use GRUB_CMDLINE_LINUX in the legacy fashion, 'grub2-editenv' will 
    # need to be used to unset the kernelopts variable.
    #
    # Example:
    #    $ sudo grub2-editenv - unset kernelopts
    #    $ sudo grub2-mkconfig -o /<path>/<to>/grub.cfg

The switch to BLS is absolutely causing quite a bit confusion, but it is also quite frustrating to users that rely on the GRUB_CMDLINE_LINUX variable for changing kernel boot parameters. For the frustrated, learning the `new` way of doing things is pretty terrible. This at least gives users the (not recommended) method as an option.

jnpkrn: +1 for whatever will make users/admins grasp how the landscape has
changed and how to adapt the old habits without investing much effort
~ alienating "non-proactive" (alleged) majority

martinezjavier: @kyle-walker we have discussed this with the rest of the bootloader folks and agreed that you are right and that `$kernelopts` should be set with the values of `GRUB_CMDLINE_LINUX{,_DEBUG}`. 

Even when the kernel command line parameters are not define in `grub.cfg` anymore and so `grub2-mkconfig` shouldn't affect it, users are too familiar with updating the cmdline with `GRUB_CMDLINE_LINUX` so we should keep that behaviour.

PR #1718, opened 2018-12-06
Run zipl again after generating initramfs (#1652727)

poncovka:
Anaconda recreates the initrds by calling new-kernel-pkg or dracut.
However, dracut doesn't invoke zipl on s390x, so we have to do it to
fix the bootloader.

Resolves: rhbz#1652727

### DIFF
--- a/pyanaconda/installation.py
+++ b/pyanaconda/installation.py
@@ -18,7 +18,7 @@
 # Red Hat, Inc.
 #
 
-from blivet import callbacks
+from blivet import callbacks, arch
 from blivet.devices import BTRFSDevice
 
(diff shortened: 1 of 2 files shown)


pyanaconda/installation.py
```diff
@@ -138,6 +138,11 @@ def doConfiguration(storage, payload, ksdata, instClass):
 
     if flags.flags.livecdInstall and boot_on_btrfs and bootloader_enabled:
         generate_initramfs.append(Task("Write BTRFS bootloader fix", writeBootLoader, (storage, payload, instClass, ksdata)))
+
+    # Invoking zipl should the last thing done on a s390x installation (see #1652727).
```
M4rtinK: should **be** that last thing

PR #1726, opened 2018-12-13
Add --skippackages option and null payload

larskarlitski:
This option is only valid with --dirinstall. It assumes that all
necesary packages are already installed in the given tree.

This is useful debugging anaconda and when installing packages from
sources that anaconda doesn't have access to.

### DIFF
--- a/pyanaconda/payload/nullpayload.py
+++ b/pyanaconda/payload/nullpayload.py
@@ -1,0 +1,38 @@
+# nullpayload.py
+#
+# Copyright (C) 2018  Red Hat, Inc.
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# the GNU General Public License v.2, or (at your option) any later version.
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY expressed or implied, including the implied warranties of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General
+# Public License for more details.  You should have received a copy of the
+# GNU General Public License along with this program; if not, write to the
+# Free Software Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
+# 02110-1301, USA.  Any Red Hat trademarks that are incorporated in the
+# source code or documentation are not subject to the GNU General Public
+# License and may only be used or replicated with the express permission of
+# Red Hat, Inc.
+#
+
+from pyanaconda.anaconda_loggers import get_module_logger
+log = get_module_logger(__name__)
+
+from pyanaconda.core import util
+from pyanaconda.payload import Payload, PayloadError, versionCmp
+
+class NullPayload(Payload):
+    """ A NullPayload assumes the target alrady contains a system and does nothing."""
+
+    @property
+    def spaceRequired(self):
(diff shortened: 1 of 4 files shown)


M4rtinK:
On a quick look looks reasonable to me given current Anaconda architecture. If I understand things correctly, you basically have the content of the installed system and you want Anaconda to configure it, right ?

In the mid-term future we would like ti make the Anaconda payload more modular and plugable, which would make things like this easier (possibly by not running a payload at all). But till we get there your solution looks fine to me. :)

pyanaconda/payload/nullpayload.py
```diff
@@ -0,0 +1,38 @@
...
+
+class NullPayload(Payload):
+    """ A NullPayload assumes the target alrady contains a system and does nothing."""
+
+    @property
+    def spaceRequired(self):
+        return 0
+
+    @property
+    def kernelVersionList(self):
```
bcl: I think this should borrow the implementation from livepayload _upateKernelVersionList from the /boot/ contents. This would possibly allow bootloader installation to work in some cases (eg. UEFI).

poncovka:
I think that this use case is really great, because we would like to support something similar for all Anaconda DBus modules. I have an alternative solution based on that.

From the Composer point of view, you could just disable the Anaconda DBus module for packaging. We already have support for this in our configuration file. You can create a new file in `/etc/anaconda/conf.d/` or we can add a new option that will allow to specify a path to your configuration file.

The configuration file should look like this:
```
[Anaconda]
kickstart_modules =
     org.fedoraproject.Anaconda.Modules.Timezone
     org.fedoraproject.Anaconda.Modules.Localization
     org.fedoraproject.Anaconda.Modules.Users
```

I would like to avoid the option `--skippackages`, because it doesn't look like something users should consider to use.

From the Anaconda point of view, we have the Payload module, but its functionality is very limited. The UI still uses an instance of the `Payload` class, so we could use the  `NullPayload` class as a temporary workaround. Once the code for payload is moved to the module, we can drop this class and make sure that Anaconda can survive with a disabled module.

The code in `pyanaconda/anaconda.py` might look like this:
```
if PAYLOAD.service_name not in conf.anaconda.kickstart_modules:
  from pyanaconda.payload.nullpayload import NullPayload
  klass = NullPayload
```

@jkonecny12, I would like to know your opinion since you will implement the Payload module. What do you think?

bcl:
I prefer the idea of being able to specify our own configuration file on the cmdline instead of modifying the system-wide one. Normally composer is going to be the only user of anaconda on an installed system, but I don't really want to overwrite the packaged configuration file, just in case there is another use for it.

larskarlitski:
> I think that this use case is really great, because we would like to support something similar for all Anaconda DBus modules. I have an alternative solution based on that. [...]

Ah, that's much better indeed, thanks! It doesn't work right now, because payload unconditionally loaded in [anaconda.py](https://github.com/rhinstaller/anaconda/blob/master/anaconda.py#L728). I guess that's just a holdover from the old way of doing things?

I found out that I can achieve the same thing in the mean time by asking anaconda to install from a live image (`liveimg` kickstart option) with an empty tarball. Your solution seems much more in line with the new architecture and doesn't require a (quite useless) new command line argument, so I'll close this PR.

Plus 1 to @bcl's suggestion to be able to specify the configuration file on the command line.

PR #1733, opened 2018-12-13
Require password confirmation in GUI (#1584064)

M4rtinK:
If a non empty password is entered in the root or user spoke
the content of the confirmation field needs to match it.

Also, the RHEL7 GUI already behaves this way.

We enable this behavior by always enabling the password confirmation
check unconditionally. This can be done as bug 1620135,
which was preventing users from escaping an empty root/user spoke,
has been fixed by disabling all the checks if nothing has
been entered.

Resolves: rhbz#1584064


M4rtinK:
jenkins, test this please

PR #1735, opened 2018-12-13
Only quit GTK mainloop on GUI quit request (#1643111)

M4rtinK:
If the user selects the quit action from GUI only quit the GTK mainloop
instead of killing the whole process by calling exit(0).

Initial Setup starts the GUI mainloop and then expects it to always to quit,
so that it can run various post-configuration actions, such as rebooting
the machine on RHEL if EULA has not been accepted.

This change should not influence "normal" non-IS Anaconda runs,
as in such case ending the mainloop should be equivalent to killing
the process.

Resolves: rhbz#1643111

### DIFF
--- a/pyanaconda/ui/gui/__init__.py
+++ b/pyanaconda/ui/gui/__init__.py
@@ -1032,7 +1032,7 @@ class GraphicalUserInterface(UserInterface):
         if rc == 1:
             self._currentAction.exited.emit(self._currentAction)
             util.ipmi_abort(scripts=self.data.scripts)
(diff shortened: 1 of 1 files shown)


rvykydal:
Not sure if this needs to be ported to Fedora, added "port to Fedora" for Martin to check.

rvykydal:
PR with prot to Fedora: #2443

rvykydal:
Ported to master with https://github.com/rhinstaller/anaconda/pull/2443

PR #1742, opened 2018-12-17
Change fedora-welcome to rhel-welcome

bcl:


jkonecny12:
Hello, we need a bug to fix anything in RHEL.

Please create a bug first or link it here.

bcl:
> Please create a bug first or link it here.

You can make it Related to rhbz#1650295

jkonecny12:
@bcl I was looking what was change and if I didn't missed something (could happen) I think the desktop icon should be also changed.

screenshot_jk_test_2019-01-11_10 53 16

When you look on this screenshot there is a Fedora logo with an arrow pointing to the HDD. For RHEL it shouldn't be a Fedora.

jkonecny12:
The same works here I guess.

screenshot_jk_test_2019-01-11_12 34 11

jstodola:
screenshot2

jkonecny12:
Thanks @jstodola for verification, then you can ignore my previous comments.

PR #1744, opened 2018-12-19
Rhel dev enable repo from hd

jkonecny12:
Enable installation from an installation tree extracted on HDD.

*Resolves: rhbz#1502323*

### DIFF
--- a/pyanaconda/image.py
+++ b/pyanaconda/image.py
@@ -116,6 +116,22 @@ def findFirstIsoImage(path):
     return None
 
 
+def verify_valid_installtree(path):
+    """Check if the given path is a valid installtree repository
+
+    :param str path: install tree path
+    :returns: True if repository is valid false otherwise
+    :rtype: bool
+    """
+    # TODO: This can be enhanced to check for repodata folder.
+    if os.path.exists(os.path.join(path, ".treeinfo")):
+        return True
+    elif os.path.exists(os.path.join(path, "treeinfo")):
+        return True
+
+    return False
+
+
 def _check_repodata(mount_path):
     install_tree_meta = InstallTreeMetadata()
     if not install_tree_meta.load_file(mount_path):
--- a/pyanaconda/payload/__init__.py
+++ b/pyanaconda/payload/__init__.py
@@ -51,7 +51,7 @@ from pyanaconda.core import util
 from pyanaconda import isys
 from pyanaconda.image import findFirstIsoImage
 from pyanaconda.image import mountImage
-from pyanaconda.image import opticalInstallMedia, verifyMedia
+from pyanaconda.image import opticalInstallMedia, verifyMedia, verify_valid_installtree
 from pyanaconda.core.util import ProxyString, ProxyStringError
 from pyanaconda.threading import threadMgr, AnacondaThread
 from pyanaconda.core.regexes import VERSION_DIGITS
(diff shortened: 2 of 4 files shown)


pyanaconda/image.py
```diff
@@ -116,6 +116,20 @@ def findFirstIsoImage(path):
     return None
 
 
+def verify_valid_installtree(path):
+    """Check if the given path is a valid installtree repository
+
```
M4rtinK: Maybe add:

`:param str path: install three path`

pyanaconda/image.py
```diff
@@ -116,6 +116,20 @@ def findFirstIsoImage(path):
...
+    :returns: True if repository is valid false otherwise
```
M4rtinK: I suggest:

```
:returns: True if repository is valid False otherwise
:rtype: bool
```

pyanaconda/image.py
```diff
@@ -116,6 +116,20 @@ def findFirstIsoImage(path):
...
+    """
+    # TODO: This can be enhanced for check for repodata folder.
```
M4rtinK: *to* check for repodata folder ?

pyanaconda/payload/__init__.py
```diff
@@ -1186,13 +1196,21 @@ def _find_and_mount_iso(self, device, device_mount_dir, iso_path, iso_mount_dir)
...
 
         return iso_path
 
+    def _setup_install_tree(self, device, install_tree_path, device_mount_dir):
+        self._setupDevice(device, mountpoint=device_mount_dir)
+        path = os.path.normpath("%s/%s" % (device_mount_dir, install_tree_path))
+
+        if not verify_valid_installtree(path):
+            device.teardown(recursive=True)
+            raise PayloadSetupError("failed to find valid installation tree")
```
poncovka: Make a real sentence from the string, please.

pyanaconda/ui/gui/spokes/installation_source.py
```diff
@@ -658,13 +659,26 @@ def status(self):
...
         else:
             return _("Nothing selected")
 
+    def _get_harddrive_partition_name(self):
+        devices = device_matches(self.data.method.partition)
+        if not devices:
+            log.warning("Device for method harddrive partition can't be found!")
+            return ""
+        elif len(devices) > 1:
+            log.warning("More than one device is found for method harddrive partition!")
```
poncovka: I am not sure if the term "method harddrive partition" makes sense.

pyanaconda/ui/gui/spokes/installation_source.py
```diff
@@ -863,6 +875,17 @@ def _initialize(self):
         # report that the source spoke has been initialized
         self.initialize_done()
 
+    def _enable_autodetect_with_device(self, device):
+        device_label = getattr(device.format, "label", "") or ""
+        self._enable_autodetect(device.name, device_label)
+
+    def _enable_autodetect(self, device_name, device_label):
```
poncovka: This method is more about showing autodetected devices than about enabling the autodetection. Could you change the name, please?

pyanaconda/ui/gui/spokes/installation_source.py
```diff
@@ -863,6 +875,17 @@ def _initialize(self):
         # report that the source spoke has been initialized
         self.initialize_done()
 
+    def _enable_autodetect_with_device(self, device):
```
poncovka: Same here.

pyanaconda/ui/gui/spokes/installation_source.py
jkonecny12: Did `harddrive method partition` sounds better?

jkonecny12:
**Updated:**
* Most of the items above. Thanks for suggestions!

pyanaconda/ui/gui/spokes/installation_source.py
VladimirSlavik: If there's internally something called `harddrive partition`, how about quotes or such?

`log.warning("Device for 'harddrive partition' method can't be found!")`

poncovka: What about `Device for installation from HDD can't be found!`? I guess I don't like the `method` there. It is an internal name.

jkonecny12: Thanks @VladimirSlavik and @poncovka for a great suggestions! Fixed.

jkonecny12:
**UPDATED:**
* Warning message changed.

jkonecny12:
**UPDATED:**
* Fix tests by adding python3-coverage as test dependency.

PR #1746, opened 2018-12-26
anaconda: add option to lock root account

fepitre:
Allow locking the root account in the password setting spoke.

### DIFF
--- a/pyanaconda/ui/gui/spokes/root_password.glade
+++ b/pyanaconda/ui/gui/spokes/root_password.glade
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<!-- Generated with glade 3.19.0 -->
+<!-- Generated with glade 3.22.1 -->
 <interface>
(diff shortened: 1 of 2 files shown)


poncovka:
Jenkins, test this please.

fepitre:
Jenkins seems happy. Is it normal that we cannot access to the build log? Above, in Show all checks, the links to Jenkins log are unavailable.

jkonecny12:
Yes it is normal. Sorry about that, we are using inner infrastructure which is not public. We have a solution how to get logs outside, however, we don't have a time to implement that. :(

M4rtinK:
I'll try to test this during this week.

fepitre:
Thank you for your feedback. Indeed it looks like that currently and it would be better to put it under the confirm password field with some padding. I will do it soon.

fepitre:
I edited the file under Glade. And put it below the confirmation box with an empty row. Is it enough?

PR #1748, opened 2019-01-03
Remove the selinux flag

poncovka:
The `selinux` option in the Anaconda configuration file should be part
of the `Security` section to keep it consistent with the DBus modules.
So let's rename the `Services` section. Then replace the `selinux` flag
with an option in the Anaconda configuration.


poncovka:
Rebased.

PR #1749, opened 2019-01-03
Add a new configuration option allow_imperfect_devices

poncovka:
Add a new configuration option allow_imperfect_devices, so we can use
it to set up blivet's flags for the rescue mode.

### DIFF
--- a/data/anaconda.conf
+++ b/data/anaconda.conf
@@ -104,6 +104,9 @@ gpt = False
 # Tell multipathd to use user friendly names when naming devices during the installation.
 multipath_friendly_names = True
 
+# Do you want to allow imperfect devices (for example, degraded mdraid array devices)?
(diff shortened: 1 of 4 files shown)


data/anaconda.conf
```diff
@@ -101,6 +101,9 @@ gpt = False
 # Tell multipathd to use user friendly names when naming devices during the installation.
 multipath_friendly_names = True
 
+# Do you want to allow imperfect devices?
+allow_imperfect_devices = False
```
M4rtinK: I would suggest to either define what an imperfect device is (I guess a broken RAID with missing memebers or similar) or link to a place where this is defined.

poncovka: I have added an example.

poncovka:
Rebased.

PR #1752, opened 2019-01-04
Remove unused code

poncovka:
Let's clean up the code again.


pyanaconda/payload/tarpayload.py
```diff
@@ -1,85 +0,0 @@
-# tarpayload.py
```
M4rtinK: This is strange - I think this should have been one of the lesser known but still supported payloads. Also IIRC the talk by a Yahoo employee about sub minute baremetal provisioning a couple years on Flock in Prague was IIRC using the tar payload. Might be good to investigate if there is really no way to use a tar payload with Anaconda and decide what to do about that, possibly asking Composer as well.

poncovka: I cannot find any proof that `TarPayload` was ever used in Anaconda. Also, notice the comment `TarPayload is not yet fully implemented`.

Btw. the `liveimg` command supports tarfiles:
https://pykickstart.readthedocs.io/en/latest/commands.html?highlight=liveimg#liveimg

pyanaconda/bootloader.py
```diff
@@ -2009,7 +2009,7 @@ class Yaboot(YabootBase):
 
     # stage2 device requirements
     stage2_device_types = ["partition", "mdarray"]
-    stage2_device_raid_levels = [raid.RAID1]
+    stage2_raid_levels = [raid.RAID1]
```
jkonecny12: Wow this may fix some ugly issues or introduce them?

pyanaconda/bootloader.py
```diff
@@ -2320,7 +2320,7 @@ class EXTLINUX(BootLoader):
     _config_dir = "/boot/extlinux"
 
     # stage1 device requirements
-    stage1_device_types = ["disk"]
+    _boot_stage1_device_types = ["disk"]
```
jkonecny12: How do you know this should be the correct name? I can't see it used anywhere.

pyanaconda/input_checking.py
```diff
@@ -334,53 +334,6 @@ def run(self, check_request):
...
-
-    def run(self, check_request):
-        """Run the function on the provided data.
-
-        :param str check_request: a string to run the function on
-        """
-        success, error_message = self._function(check_request)
-        self.result.error_message = error_message
-        self.result.success = success
-
```
jkonecny12: Can't this be used in the future? Maybe we want to leave it here to avoid coding this again in the future.

pyanaconda/payload/dnfpayload.py
```diff
@@ -936,17 +935,6 @@ def environmentId(self, environment):
...
-        env = self._base.comps.environment_by_pattern(environmentid)
-        if env is None:
-            raise payload.NoSuchGroup(environmentid)
-        group_ids = (id_.name for id_ in env.group_ids)
-        option_ids = (id_.name for id_ in env.option_ids)
-        if optional:
-            return list(itertools.chain(group_ids, option_ids))
-        else:
-            return list(group_ids)
-
```
jkonecny12: @M4rtinK what do you think about this removal?

pyanaconda/payload/tarpayload.py
jkonecny12: I'm also not convinced about removing this. It is weird to have it here. I want to look on this more before the removal.

pyanaconda/bootloader.py
poncovka: This may enable installation of the Yaboot boot loader on RAID1 again. The attribute was renamed in a commit cc17c1b019d17400cccc1e006b5314adcb531537, so this is clearly a bug.

pyanaconda/bootloader.py
poncovka: You are right, the attribute should be removed. Fixed.

poncovka:
Updated.

poncovka:
Jenkins, test this please.

pyanaconda/payload/tarpayload.py
jkonecny12: After some more investigation it looks like tar payload was implemented in the Great RHEL 7 rewrite as some first version and from that time no-one touch that. So I think it is safe to remove this, especially when this functionality is already implemented in the liveimg payload.

pyanaconda/bootloader.py
jkonecny12: Ok thanks!

pyanaconda/input_checking.py
jkonecny12: I agree that it is not that difficult to get it back and smaller code-base means less problems. 
Good arguments, you can remove this.

pyanaconda/payload/dnfpayload.py
M4rtinK: I think this can be safely removed. It likely became dead code when we delegated all group and package exclusion to DNF. Before that we needed to find out what groups are contained in an environment so that we can drop the excluded ones. And apparently we don't need that functionality anywhere else.

PR #1758, opened 2019-01-09
Move code from Blivet

poncovka:
Anaconda-specific attributes can be moved from the `Blivet` class
in Blivet to the `InstallerStorage` class in Anaconda.

Remove the Blivet's `gpt` flag. Blivet doesn't use this flag, so let's
use our `config` instead.


poncovka:
Jenkins, test this please.

PR #1765, opened 2019-01-14
Move code to pyanaconda.bootloader

poncovka:
Move the boot-loader-related code to `pyanaconda.bootloader`
and split `bootloader.py` to smaller files.


poncovka:
@jhlavac, @martinezjavier, could you check if the new arrangement makes sense, please?

martinezjavier:
@poncovka the reorganization looks good to me, is great to see that huge file to be split.

As a follow-up, I think that are several things that can be removed. For example, I the following classes are no longer needed AFAIK:

* `GRUB()`
* `EFIGRUB1()`
* `Yaboot()`

PR #1768, opened 2019-01-15
Change Skylake Spectre_V2 default to Retpolines (#1659626)

poncovka:
Set the boot option `spectre_v2` to `retpoline` on Intel Skylake systems.

Resolves: rhbz#1659626

### DIFF
--- a/pyanaconda/bootloader.py
+++ b/pyanaconda/bootloader.py
@@ -30,7 +30,7 @@ from parted import PARTITION_BIOS_GRUB
 
 from pyanaconda import iutil
 from blivet.devicelibs import raid
-from pyanaconda.isys import sync
+from pyanaconda.isys import sync, is_skylake
 from pyanaconda.product import productName
 from pyanaconda.flags import flags
 from blivet.errors import StorageError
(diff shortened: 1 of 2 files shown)


pyanaconda/bootloader.py
```diff
@@ -826,6 +826,12 @@ def set_boot_args(self, *args, **kwargs):
         if flags.cmdline.get("fips") == "1" and boot_device:
             self.boot_args.add("boot=%s" % self.stage2_device.fstabSpec)
 
+        #
+        # Spectre
```
jkonecny12: I would rename this to something like ``Mitigation to Spectre/Meltdown``

pyanaconda/isys/__init__.py
```diff
@@ -99,6 +99,27 @@ def isLpaeAvailable():
...
+def get_cpu_model():
+    """Get the cpu model number."""
+    with open("/proc/cpuinfo", "r") as f:
+        for line in f:
+            name, value = line.split(":")
+
+            if name.strip() == "model":
+                return int(value.strip())
+
+    return None
```
jkonecny12: I hoped that python will have something in the standard library but I wasn't able to find anything :(.

Please add information about output format to the function documentation.

pyanaconda/bootloader.py
poncovka: Fixed.

pyanaconda/isys/__init__.py
poncovka: Fixed.

jkonecny12: I think there should be also `an integer value *or None*`. The None part possibility is missing here.

poncovka: It should be fixed now. Thanks!

poncovka:
Tested with my cpu model listed in the set of Skylake models.

PR #1769, opened 2019-01-16
Fix keyboard selinux label

jkonecny12:
Backport of PR https://github.com/rhinstaller/anaconda/pull/1766

It is even better here. On RHEL we get system with label `var_lib_t` on Fedora there is `mock_var_lib_t`?!?

### DIFF
--- a/data/post-scripts/80-setfilecons.ks
+++ b/data/post-scripts/80-setfilecons.ks
@@ -7,7 +7,7 @@
 # - ostree payloads, where all of the labeling of /var is the installer's
 #   responsibility (see https://github.com/ostreedev/ostree/pull/872 )
 
-restorecon -ir /etc/sysconfig/network-scripts /etc/lvm \
+restorecon -ir /etc/sysconfig/network-scripts /etc/lvm /etc/X11/xorg.conf.d \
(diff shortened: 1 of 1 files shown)


jkonecny12:
**UPDATED:**
* Added Fedora bug [#1666892](https://bugzilla.redhat.com/show_bug.cgi?id=1666892) to the commit message

M4rtinK:
OK, `mock_var_lib_t` is pretty weird for something installed on the system. :) Maybe it get's "burned in" somehow at package build time ? That could be a rpmbuild/Koji bug - if the packages ship labeled files, they should IMHO have labels corresponding to to their destination, not to the temporary folder where they were build.

jkonecny12:
> OK, `mock_var_lib_t` is pretty weird for something installed on the system. :) Maybe it get's "burned in" somehow at package build time ? That could be a rpmbuild/Koji bug - if the packages ship labeled files, they should IMHO have labels corresponding to to their destination, not to the temporary folder where they were build.

I guess so but in general I also wonder how it gets in. I guess something changed recently in the build process so it began to create this errors. Could it be something which will reflect also to composer images?

PR #1776, opened 2019-01-22
Remove code from osinstall

poncovka:
* Remove `update_ksdata` from the `InstallerStorage` class
* Move `get_ignored_nvdimm_blockdevs` to `pyanaconda.storage.utils `

### DIFF
--- a/pyanaconda/storage/kickstart.py
+++ b/pyanaconda/storage/kickstart.py
@@ -1,0 +1,199 @@
+#
+# Copyright (C) 2019  Red Hat, Inc.
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# the GNU General Public License v.2, or (at your option) any later version.
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY expressed or implied, including the implied warranties of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General
+# Public License for more details.  You should have received a copy of the
+# GNU General Public License along with this program; if not, write to the
+# Free Software Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
+# 02110-1301, USA.  Any Red Hat trademarks that are incorporated in the
+# source code or documentation are not subject to the GNU General Public
+# License and may only be used or replicated with the express permission of
+# Red Hat, Inc.
+#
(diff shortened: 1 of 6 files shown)


pyanaconda/storage/kickstart.py
```diff
@@ -0,0 +1,199 @@
...
+        if not destroy_actions:
+            pass
+        elif partial:
+            # make a list of removed partitions
+            disk_init_proxy.SetInitializationMode(CLEAR_PARTITIONS_LIST)
+            disk_init_proxy.SetDevicesToClear(cleared_partitions)
+        else:
+            # if they didn't partially clear any disks, use the shorthand
+            disk_init_proxy.SetInitializationMode(CLEAR_PARTITIONS_ALL)
+            disk_init_proxy.SetDrivesToClear(fresh_disks)
```
jkonecny12: Could you please move this big `if` part under an abstraction of the separate method?

pyanaconda/storage/kickstart.py
```diff
@@ -0,0 +1,199 @@
...
+
+
+def _update_custom_storage(storage, ksdata):
```
jkonecny12: Could you please break this 70 lines long method to a smaller methods if possible?

pyanaconda/storage/kickstart.py
```diff
@@ -0,0 +1,199 @@
...
+    DISK_INITIALIZATION
+from pyanaconda.modules.common.constants.services import STORAGE
+
+log = get_module_logger(__name__)
+
+__all__ = ["update_storage_ksdata"]
+
+
+def update_storage_ksdata(storage, ksdata):
+    """Update kickstart data to reflect the current storage configuration..
```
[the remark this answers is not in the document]
poncovka: Fixed.

pyanaconda/storage/kickstart.py
```diff
@@ -0,0 +1,199 @@
...
+
+    :param storage: an instance of the storage
+    :param ksdata: an instance of kickstart data
+    """
+    if not ksdata or not storage.mountpoints:
+        return
```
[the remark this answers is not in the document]
poncovka: I guess this `if` handles dir installations? I don't want to dig into this code right now. It will be eventually moved to the storage submodules.

pyanaconda/storage/kickstart.py
poncovka: Same here.

poncovka:
Updated. The goal of this pull request is to remove the code from `osinstall.py`. It will be eventually moved to the storage submodules, so it can be refactorized then.

poncovka:
The errors are not related to this PR.

PR #1779, opened 2019-01-23
Remove the preStorage method in payload

poncovka:
Remove the useless method `preStorage` from the payload base class
and the related installation task.


poncovka:
Jenkins, test this please.

PR #1780, opened 2019-01-23
Don't create an extra instance of the bootloader

poncovka:
The Anaconda class doesn't have to create its own instance of
the bootloader. Use storage.bootloader instead.


poncovka:
Jenkins, test this please.

PR #1786, opened 2019-01-28
Remove obsolete ldconfig scriptlets

M4rtinK:
References:
https://fedoraproject.org/wiki/Changes/RemoveObsoleteScriptlets
Signed-off-by: Igor Gnatenko <ignatenkobrain@fedoraproject.org>


M4rtinK:
Found this change in distgit, ideally should be part of the next build so that we don't put ldconfig back in. :)

M4rtinK:
Managed to work around this in distgit, but we should definitely get this in before the *next* build. :)

PR #1790, opened 2019-01-31
Set default entry to the BLS id instead of the entry index

martinezjavier:
The GRUB default boot entry is set to the default's entry index. This made
sense on a non-BLS configuration because the entries where defined in the
GRUB config file, so they had a strict definition order.

But that's not the case on a BLS configuration, users can add or remove an
entry by just dropping or removing files from the /boot/loader/entries dir.

That means the default won't necessary be the installed kernel, but rather
the BLS that's first in the sorted list of BLS snippets.

So instead the default entry shold be explicit and be set to the BLS id of
the default kernel (the machine ID followed by a dash and kernel version).

That is also what is used when a new kernel is installed, so the initial
grubenv should be consistent.

Resolves: rhbz#1671047

### DIFF
--- a/pyanaconda/bootloader/grub2.py
+++ b/pyanaconda/bootloader/grub2.py
@@ -327,16 +327,16 @@ class GRUB2(BootLoader):
 
         # make sure the default entry is the OS we are installing
         if self.default is not None:
-            # find the index of the default image
-            try:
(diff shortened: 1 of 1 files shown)


poncovka:
Jenkins, test this please.

PR #1793, opened 2019-02-01
Remove multiboot support for tboot

martinezjavier:
The multiboot support for tboot was added in 2011 by commit 4ac57face30
("add multiboot support for tboot") but it has been unused since commit
e2db119e6cd ("Remove some modules obsoleted by the packaging module.").

The latter removed the code that set the BootLoader class trusted_boot
property if the tboot package was installed, so now it's never enabled.

Also the multiboot support has been removed from the grubby BLS wrapper
and it's not even built into the EFI grub2 binary anymore. Since could
be used by an attacker to bypass the Secure Boot mechanism and execute
unsigned binaries.

### DIFF
--- a/pyanaconda/bootloader/base.py
+++ b/pyanaconda/bootloader/base.py
@@ -146,8 +146,6 @@ class BootLoader(object):
                             "net.ifnames"]
     preserve_args = []
 
-    _trusted_boot = False
-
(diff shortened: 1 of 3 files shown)


jkonecny12:
Jenkins, it's ok to test.

PR #1795, opened 2019-02-04
Allow non-ASCII characters in passphrases again (#1619813)

poncovka:
We show a warning for using non-ASCII characters in passphrases, but
it should be allowed to use them anyway.

Resolves: rhbz#1619813


M4rtinK:
IIRC the original idea was to prevent users from using characters in passphrases that users might not be able to type in the limited early boot environment where LUKS is generally unloked.

But as we show a warning & worst case people can just rerun the install as there will not be any real data on the new LUKS anyway. So it should be IMHO fine to enable.

PR #1796, opened 2019-02-04
Clean up the InstallerStorage from the DBus calls

poncovka:
Configure the storage and reload the additional storage modules before
we call the `reset` method of the `InstallerStorage` class. Let's move the
code to the `reset_storage` function.

Let's create a new function `write_storage_configuration` from the
`write` method of the `InstallerStorage` class, so we can later create
an installation task from it.


poncovka:
Rebased.

PR #1805, opened 2019-02-11
Clean up write_boot_loader

poncovka:
The method `set_boot_args` in the `Bootloader` class now depends only on
the storage object. The method was split to several smaller methods.

The function `write_boot_loader` was split to several smaller functions
and reorganized, so we can later drop the dependency on the payload
object.

### DIFF
--- a/pyanaconda/bootloader/base.py
+++ b/pyanaconda/bootloader/base.py
@@ -18,9 +18,9 @@
 import collections
 import os
 from glob import glob
-from itertools import chain
 
 import blivet
+from blivet.devices import NetworkStorageDevice
 from blivet.formats.disklabel import DiskLabel
 from blivet.size import Size
 from ordered_set import OrderedSet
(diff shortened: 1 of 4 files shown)


pyanaconda/bootloader/base.py
```diff
@@ -826,20 +800,27 @@ def set_boot_args(self, *args, **kwargs):
...
-                setup_string = cfg_obj.dracutSetupString()
-                self.boot_args.add(setup_string)
-                self.dracut_args.add(setup_string)
+    def _set_graphical_boot_args(self):
+        """Set up the graphical boot."""
+        args = []
+
+        try:
+            import rpm
+        except ImportError:
```
M4rtinK: I wonder if logging a warning there might make sense ? Not sure how serious the situation is though & if it warrants a warning.

poncovka: Fixed.

poncovka:
Updated.

poncovka:
These changes are merged in https://github.com/rhinstaller/anaconda/pull/1804. I messed up the rebase, sorry.

jkonecny12:
That sometimes happened, it's fine :)

PR #1806, opened 2019-02-11
Create installation tasks for the storage

poncovka:
Create the DBus installation tasks for the storage configuration.

### DIFF
--- a/pyanaconda/modules/storage/installation.py
+++ b/pyanaconda/modules/storage/installation.py
@@ -1,0 +1,75 @@
+#
+# Installation tasks
+#
+# Copyright (C) 2019 Red Hat, Inc.
+#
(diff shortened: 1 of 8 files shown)


pyanaconda/modules/storage/installation.py
```diff
@@ -0,0 +1,75 @@
...
+        """Create a new task."""
+        super().__init__()
+        self._storage = storage
+
+    @property
+    def name(self):
+        return "Mount filesystems"
+
+    def run(self):
+        """Mount the filesystems.."""
```
jkonecny12: Please remove one period.

poncovka: Fixed.

PR #1807, opened 2019-02-11
Schedule the kickstart custom partitioning in the Storage module

poncovka:
Let's remove the data argument form the execute methods and set it
only in the `__init__` method of the custom partitioning executor.
Also introduce a base class for the partitioning executors.

Create a storage module for custom partitioning defined by kickstart.

### DIFF
--- a/pyanaconda/storage/execution.py
+++ b/pyanaconda/storage/execution.py
@@ -15,6 +15,8 @@
 # License and may only be used or replicated with the express permission of
 # Red Hat, Inc.
 #
+from abc import ABC, abstractmethod
+
 import blivet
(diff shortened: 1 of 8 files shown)


pyanaconda/storage/execution.py
```diff
@@ -120,15 +123,23 @@ def clear_partitions(storage):
     storage.clear_partitions()
 
 
-class AutomaticPartitioningExecutor(object):
-    """The executor of the automatic partitioning."""
+class PartitioningExecutor(object):
```
jkonecny12: Ideally you should also inherit from ABC to avoid base class instantiation.

poncovka: Fixed.

PR #1808, opened 2019-02-11
Remove a confusing message for headless systems (#1638791)

poncovka:
Let's remove the confusing message 'DISPLAY variable not set'.

Resolves: rhbz#1638791

PR #1809, opened 2019-02-11
Change a confusing message for headless systems

poncovka:
Let's change the confusing message 'DISPLAY variable not set'.

Related: rhbz#1638791

### DIFF
--- a/pyanaconda/display.py
+++ b/pyanaconda/display.py
@@ -277,7 +277,7 @@ def setup_display(anaconda, options, addon_paths=None):
 
     # X on a headless (e.g. s390) system? Nonsense!
     if want_x and anaconda.isHeadless:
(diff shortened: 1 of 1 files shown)


poncovka:
Ported from https://github.com/rhinstaller/anaconda/pull/1808.

jkonecny12:
Do we really need this message translated? If not please remove the translation function.

poncovka:
> Do we really need this message translated? If not please remove the translation function.

We translate most of the messages at this point of the installation. If we don't want them to be translated, we should change all of them.

jkonecny12:
Maybe, on the other hand we are adding unnecessary work to translators. So if we can decide now, I would rather decide and change it before we add new work for people.

PR #1810, opened 2019-02-11
Don't load storage plugins in dir installations (#1674605)

poncovka:
We shouldn't load storage plugins in dir installations.

Resolves: rhbz#1674605


poncovka:
Jenkins, test this please.

PR #1813, opened 2019-02-13
Change the overflow color

poncovka:
Change the overflow color from 2f4265 to 2d2d2d.

Related: rhbz#1637686


poncovka:
Changed the related bug.

PR #1814, opened 2019-02-13
Fix missing space in translatable string

AsciiWolf:
/cc @jkonecny12


jkonecny12:
Jenkins, it's ok to test.

PR #1815, opened 2019-02-14
Use unformatted DASDs for the partitioning (#1676630)

poncovka:
When an unformatted DASD disk is available during kickstart
installation and it is formatted, it should be selected for
the partitioning by default.

Resolves: rhbz#1676630

### DIFF
--- a/pyanaconda/storage/osinstall.py
+++ b/pyanaconda/storage/osinstall.py
@@ -2314,16 +2314,26 @@ def storage_initialize(storage, ksdata, protected):
        not any(d.protected for d in storage.devices):
         raise UnknownSourceDeviceError(protected)
 
-    # kickstart uses all the disks
-    if flags.automatedInstall:
-        disk_select_proxy = STORAGE.get_proxy(DISK_SELECTION)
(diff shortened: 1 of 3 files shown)


pyanaconda/storage/osinstall.py
```diff
@@ -2314,16 +2314,26 @@ def storage_initialize(storage, ksdata, protected):
...
-    if flags.automatedInstall:
-        disk_select_proxy = STORAGE.get_proxy(DISK_SELECTION)
-        selected_disks = disk_select_proxy.SelectedDisks
-        ignored_disks = disk_select_proxy.IgnoredDisks
 
-        if not selected_disks:
-            selected_disks = [d.name for d in storage.disks if d.name not in ignored_disks]
-            disk_select_proxy.SetSelectedDisks(selected_disks)
-            log.debug("onlyuse is now: %s", ",".join(selected_disks))
+def select_all_disks_by_default(storage):
```
jkonecny12: The same as on the other PR.
`by_default` seems to me like usage of the method not something what the method is doing. Can you please rename that?

PR #1816, opened 2019-02-14
Apply the partitioning in the Storage module

poncovka:
Call `ApplyPartitioning` to use the partitioned storage from the given
partitioning module in the Storage module.

### DIFF
--- a/pyanaconda/modules/storage/storage.py
+++ b/pyanaconda/modules/storage/storage.py
@@ -23,6 +23,8 @@ from pyanaconda.core import util
 from pyanaconda.core.signal import Signal
 from pyanaconda.dbus import DBus
 from pyanaconda.modules.common.base import KickstartModule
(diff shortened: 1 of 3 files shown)


pyanaconda/modules/storage/storage.py
```diff
@@ -89,15 +83,31 @@ def __init__(self):
...
 
     def _add_module(self, storage_module):
         """Add a base kickstart module."""
         self._modules.append(storage_module)
 
+    def _add_partitioning_module(self, object_path, partitioning_module):
+        """Add a partitioning module."""
+        # Add the module.
+        self._modules.append(partitioning_module)
+        self._partitioning_modules[object_path] = partitioning_module
```
jkonecny12: It could be interesting to check if the `object_path` is already in the dict and raise an exception if it is. It couldn't happen but during all the refactorization it could happen.

PR #1818, opened 2019-02-14
Use unformatted DASDs for the partitioning

poncovka:
When an unformatted DASD disk is available during kickstart
installation and it is formatted, it should be selected for
the partitioning by default.

Related: rhbz#1676630

### DIFF
--- a/pyanaconda/storage/initialization.py
+++ b/pyanaconda/storage/initialization.py
@@ -27,7 +27,6 @@ from blivet.iscsi import iscsi
 from pyanaconda.anaconda_logging import program_log_lock
 from pyanaconda.core.configuration.anaconda import conf
 from pyanaconda.errors import errorHandler as error_handler, ERROR_RAISE
-from pyanaconda.flags import flags
 from pyanaconda.modules.common.constants.objects import DISK_SELECTION, AUTO_PARTITIONING, \
     DISK_INITIALIZATION, FCOE, ZFCP
 from pyanaconda.modules.common.constants.services import STORAGE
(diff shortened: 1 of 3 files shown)


poncovka:
Ported from https://github.com/rhinstaller/anaconda/pull/1815.

pyanaconda/storage/initialization.py
```diff
@@ -150,16 +149,26 @@ def initialize_storage(storage):
...
-    if flags.automatedInstall:
-        disk_select_proxy = STORAGE.get_proxy(DISK_SELECTION)
-        selected_disks = disk_select_proxy.SelectedDisks
-        ignored_disks = disk_select_proxy.IgnoredDisks
 
-        if not selected_disks:
-            selected_disks = [d.name for d in storage.disks if d.name not in ignored_disks]
-            disk_select_proxy.SetSelectedDisks(selected_disks)
-            log.debug("onlyuse is now: %s", ",".join(selected_disks))
+def select_all_disks_by_default(storage):
```
jkonecny12: Not sure about having `by_default` in the method name. It is more how the method is used not what the method is doing.

poncovka: If there are already some disks selected, this method will do nothing. So it will really select all disks by default. I think that eventually, we will be able to this in a better way, so consider this to be a temporary fix.

jkonecny12: If it is just temporary then I may close one eye. But I'm still convinced there should be better name used.

poncovka:
Rebased.

PR #1820, opened 2019-02-14
Check for dirinstall target on s390

bcl:
Don't prompt for ssh if --dirinstall mode is being used.

### DIFF
--- a/anaconda.py
+++ b/anaconda.py
@@ -315,7 +315,7 @@ if __name__ == "__main__":
     # see if we're on s390x and if we've got an ssh connection
     uname = os.uname()
     if uname[4] == 's390x':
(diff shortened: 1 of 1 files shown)


anaconda.py
```diff
@@ -315,7 +315,7 @@ def setup_environment():
     # see if we're on s390x and if we've got an ssh connection
     uname = os.uname()
     if uname[4] == 's390x':
-        if 'TMUX' not in os.environ and 'ks' not in flags.cmdline and not conf.target.is_image:
+        if 'TMUX' not in os.environ and 'ks' not in flags.cmdline and not conf.target.is_image and not conf.target.is_directory:
```
poncovka: Please, change `not conf.target.is_image and not conf.target.is_directory` to `conf.target.is_hardware`.

PR #1831, opened 2019-02-20
Create the Snapshot module

poncovka:
Create a new module for the kickstart command `snapshot` and use it in UI when possible.

### DIFF
--- a/pyanaconda/modules/storage/snapshot/device.py
+++ b/pyanaconda/modules/storage/snapshot/device.py
@@ -1,0 +1,63 @@
+#
+# Snapshot device
+#
+# Copyright (C) 2019 Red Hat, Inc.
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# the GNU General Public License v.2, or (at your option) any later version.
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY expressed or implied, including the implied warranties of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General
+# Public License for more details.  You should have received a copy of the
+# GNU General Public License along with this program; if not, write to the
+# Free Software Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
+# 02110-1301, USA.  Any Red Hat trademarks that are incorporated in the
+# source code or documentation are not subject to the GNU General Public
+# License and may only be used or replicated with the express permission of
+# Red Hat, Inc.
+#
+from blivet.devices import LVMLogicalVolumeDevice
+from pykickstart.errors import KickstartParseError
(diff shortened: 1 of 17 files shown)


pyanaconda/modules/storage/snapshot/device.py
```diff
@@ -0,0 +1,63 @@
...
+        raise KickstartParseError(_("Snapshot: origin \"%(origin)s\" of snapshot "
+                                    "\"%(name)s\" is not a valid thin LV device.")
+                                  % {"origin": request.origin, "name": request.name},
+                                  lineno=request.lineno)
+
+    if devicetree.get_device_by_name("%s-%s" % (origin_dev.vg.name, snap_name)):
+        raise KickstartParseError(_("Snapshot %s already exists.") % request.name,
+                                  lineno=request.lineno)
+    try:
+        return LVMLogicalVolumeDevice(
```
poncovka: @vojtechtrefny, is it safe to create this object just for a verification and then create it again when we really want to create the device? It seems to work fine.

tests/nosetests/pyanaconda_tests/module_snapshot_test.py
```diff
@@ -0,0 +1,133 @@
...
+        self.assertEqual(self.interface.IsRequested(SNAPSHOT_WHEN_PRE_INSTALL), False)
+        self.assertEqual(self.interface.IsRequested(SNAPSHOT_WHEN_POST_INSTALL), False)
+
+        self.module._requests = [Mock(when=SNAPSHOT_WHEN_PRE_INSTALL)]
+        self.assertEqual(self.interface.IsRequested(SNAPSHOT_WHEN_PRE_INSTALL), True)
+        self.assertEqual(self.interface.IsRequested(SNAPSHOT_WHEN_POST_INSTALL), False)
+
+        self.module._requests = [Mock(when=SNAPSHOT_WHEN_POST_INSTALL)]
+        self.assertEqual(self.interface.IsRequested(SNAPSHOT_WHEN_PRE_INSTALL), False)
+        self.assertEqual(self.interface.IsRequested(SNAPSHOT_WHEN_POST_INSTALL), True)
```
jkonecny12: You may add test when both are set. The result should be the last wins if I'm not mistaken.

pyanaconda/modules/storage/snapshot/device.py
vojtechtrefny: Yes, it is safe. If you don't add it to the devicetree, just creating the device doesn't affect the Blivet() instance state.

tests/nosetests/pyanaconda_tests/module_snapshot_test.py
jkonecny12: By the both I mean `SNAPSHOT_WHEN_(PRE|POST)_INSTALL` :).

poncovka: Fixed.

PR #1837, opened 2019-02-25
dracut/parse-kickstart: Set DEVICETYPE instead of TYPE in a team mast…

ptalbert:
…er ifcfg file

Both NetworkManager and the legacy network service scripts expect a team master
ifcfg file to have DEVICETYPE="Team" set. Currently, only TYPE is set which
means the legacy network service will not assign IP addressing as expected.

Signed-off-by: Patrick Talbert <ptalbert@redhat.com>

### DIFF
--- a/dracut/parse-kickstart
+++ b/dracut/parse-kickstart
@@ -566,7 +566,7 @@ def ksnet_to_ifcfg(net, filename=None):
     if net.teamslaves:
 
         ifcfg.pop('HWADDR', None)
-        ifcfg['TYPE'] = "Team"
+        ifcfg['DEVICETYPE'] = "Team"
         ifcfg['NAME'] = "Team connection %s" % dev
         ifcfg['TEAM_CONFIG'] = net.teamconfig
 


jkonecny12:
Jenkins, it's ok to test.

rvykydal:
I am for including of the patch but it needs some updates so we handle the DEVICETEAM where needed:

- update for stage2 (which should be included in any case, not related directly to this patch, we are already taking DEVICETEAM into account at other places in stage 2):
https://github.com/rhinstaller/anaconda/pull/1844
- update of this PR:
https://github.com/rhinstaller/anaconda/pull/1845
The latter PR is also done for f30-devel where we are targeting our devel PRs at this stage.

My manual testing passed.

@ptalbert, would you be OK if we close this PR and handle the issue in the https://github.com/rhinstaller/anaconda/pull/1845 ?

ptalbert:
@rvykydal, that is fine by me; thank you!

rvykydal:
Closing, see https://github.com/rhinstaller/anaconda/pull/1845.

PR #1840, opened 2019-02-26
Network remove api for boot options

rvykydal:


rvykydal:
@poncovka @jkonecny12 
I've added one more patch removing also DisableIPv6 property.

PR #1841, opened 2019-02-26
Fix pep8 errors in TUI source spoke

jkonecny12:


pyanaconda/ui/tui/spokes/installation_source.py
```diff
@@ -426,7 +432,8 @@ def refresh(self, args=None):
         # check if there are any mountable devices
         if self._mountable_devices:
             for d in self._mountable_devices:
-                self._container.add(TextWidget(d[1]), callback=self._select_mountable_device, data=d[0])
+                self._container.add(TextWidget(d[1]), callback=self._select_mountable_device,
+                                    data=d[0])
```
poncovka: Personally, I don't like this style. I prefer a new line for every argument.

jkonecny12: Fair, I'll change that.

jkonecny12:
Will be solved only in F30 branch.

rvykydal:
Applied to master with https://github.com/rhinstaller/anaconda/pull/1843

PR #1846, opened 2019-02-27
network: Get FCoE nics from the DBus module

poncovka:
Use the method GetNics of the FCoE DBus module to get the nics.


poncovka:
@rvykydal, can you look at it, please? I am not sure if this is correct.

PR #1847, opened 2019-02-27
Set up the kickstart partitioning from the storage by default

poncovka:
A partitioning module should set up the kickstart data from the storage
playground by default. The automatic and manual partitioning modules
override this method, but the custom partitioning module will use it.

### DIFF
--- a/pyanaconda/modules/storage/partitioning/base.py
+++ b/pyanaconda/modules/storage/partitioning/base.py
@@ -19,6 +19,9 @@
 #
 from abc import abstractmethod
 
+from blivet.devices import PartitionDevice, TmpFSDevice, LVMLogicalVolumeDevice, \
+    LVMVolumeGroupDevice, MDRaidArrayDevice, BTRFSDevice
+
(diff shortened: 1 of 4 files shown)


pyanaconda/modules/storage/partitioning/base.py
```diff
@@ -76,3 +79,72 @@ def validate_with_task(self):
...
+            bootloader_devices.append(storage.bootloader_device)
+
+        for device in storage.devices:
+            if device.format.type == 'biosboot':
+                bootloader_devices.append(device)
+
+        # Make a list of ancestors of all used devices
+        used_devices = list(storage.mountpoints.values()) + storage.swaps + bootloader_devices
+        all_devices = list(set(a for d in used_devices for a in d.ancestors))
+        all_devices.sort(key=lambda d: len(d.ancestors))
```
jkonecny12: This is a nice candidate for a separate method.

poncovka: It is not my code and I don't want to mess with it.

PR #1850, opened 2019-02-27
Get the required device size for the given space from DBus

poncovka:
Remove the method required_device_size from Payload and use the
method GetRequiredDeviceSize from the Storage module instead.

Let's try to always use the format class with largest space taken
by metadata for the calculation of the required size.

### DIFF
--- a/pyanaconda/ui/gui/spokes/lib/resize.py
+++ b/pyanaconda/ui/gui/spokes/lib/resize.py
@@ -27,10 +27,10 @@ gi.require_version("Gtk", "3.0")
 from gi.repository import Gdk, Gtk
 
 from pyanaconda.core.i18n import _, C_, N_, P_
+from pyanaconda.modules.common.constants.services import STORAGE
 from pyanaconda.ui.gui import GUIObject
 from pyanaconda.ui.gui.utils import blockedHandler, escape_markup, timed_action
 from blivet.size import Size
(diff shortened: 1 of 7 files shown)


pyanaconda/ui/gui/spokes/lib/resize.py
```diff
@@ -68,6 +68,13 @@ def __init__(self, data, storage, payload):
         self.storage = storage
         self.payload = payload
 
+        self._storage_proxy = STORAGE.get_proxy()
+
+        # Get the required device size.
+        required_space = self.payload.space_required.get_bytes()
+        required_size = self._storage_proxy.GetRequiredDeviceSize(required_space)
+        self._required_size = Size(required_size)
```
jkonecny12: Not sure if this is the correct way, what if we would want to re-use the dialog? 

I would rather see the original variant.

poncovka: We always create a new dialog. I am not even sure if a dialog can be run twice.

poncovka: Btw. we set a label with this value in the `__init__` method, so the original variant is pointless.

jkonecny12: In that case this makes the think more obvious. Thanks!

PR #1851, opened 2019-02-28
Offer only supported sector sizes in NVDIMM reconfiguration dialog

rvykydal:
Resolves: rhbz#1614049

### DIFF
--- a/pyanaconda/ui/gui/spokes/advstorage/nvdimm.py
+++ b/pyanaconda/ui/gui/spokes/advstorage/nvdimm.py
@@ -38,6 +38,9 @@ PAGE_ACTION = 1
 PAGE_RESULT_ERROR = 2
 PAGE_RESULT_SUCCESS = 3
 
+NVDIMM_SECTOR_SIZE_OPTIONS = ['512', '4096']
+
+
 class NVDIMMDialog(GUIObject):
     """
        .. inheritance-diagram:: NVDIMMDialog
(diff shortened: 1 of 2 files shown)


pyanaconda/ui/gui/spokes/advstorage/nvdimm.py
```diff
@@ -38,6 +38,9 @@
 PAGE_RESULT_ERROR = 2
 PAGE_RESULT_SUCCESS = 3
 
+NVDIMM_SUPPORTED_SECTOR_SIZES = ['512', '4096']
```
jkonecny12: Aren't we able to get these values from blivet?

rvykydal: We are not.
@vojtechtrefny - would it make sense to provide such values ?

rvykydal: After talking to Vojta I am just going to rename the constant.
There seems to be about 8 sizes supported for sector mode currently, but we don't want to offer all of them in UI. Vojta will probably propagate info about supported sizes by libblockdev api, and we may use it to check kickstart values, but it should be solved in a separate issue.

jkonecny12: Nice cooperation thanks!

rvykydal:
The test have actually passed.

rvykydal:
Ported to master by https://github.com/rhinstaller/anaconda/pull/2323

PR #1852, opened 2019-02-28
Set up the disk initialization module from the partitioned storage

poncovka:
The disk initialization module should set up its configuration from
the partitioned storage. The callback is not connected to a signal
for now, however the code will be used in UI.

Move the tests for the disk initialization module to a new file.


poncovka:
Rebased.

PR #1854, opened 2019-02-28
Replace the default autopart type in the Storage module

poncovka:
If the autopart type is default, the Storage module should provide
the real value of the type, so UI doesn't have to deal with it.


jkonecny12:
jenkins, test this please

jkonecny12:
jenkins, test this please

PR #1861, opened 2019-03-04
Fix the import of ZFCP (#1684583)

poncovka:
We should import the DBus object identifier for ZFCP.

Resolves: rhbz#1684583


sharkcz:
ACK from me too, it fixes the problem.

PR #1869, opened 2019-03-05
Fix kickstart liveimg support

bcl:


jkonecny12:
Please fix this pylint issue:
```
W0703(broad-except):285,11: do_transaction: Catching too general exception BaseException
```
Also @martinezjavier could you please verify this patch?

martinezjavier:
@jkonecny12 @bcl the patches looks good to me, but this is very similar to @AdamWill's fix for Fedora in commit e5619c9d0948 "Re-generate BLS loader file snippets on live installs (#1648472)".

There are just some minor differences in the patches AFAICT (i.e: removing the BLS vs excluding /boot/loader/entries from the tar command, calling kernel-install after the machine-id was generated vs generating the machine-id earlier, etc).

So I think that would be good to consolidate the two and avoid the RHEL and Fedora branches to diverge, that would make maintenance easier in the long term.

jkonecny12:
The https://github.com/rhinstaller/anaconda/commit/e5619c9d094833fd9d256c2c8d1ad22fc43b3b13 is already there so this should be based on that I guess.
@bcl what do you think?

bcl:
Ok, rewrote these with adamw's patch and the ones from my f30 version. Tested against tar, squashfs and bare ext4 filesystem using liveimg.

jkonecny12:
Please fix this pylint error:
```
************* Module ../pyanaconda/payload/dnfpayload.py
W0703(broad-except):285,11: do_transaction: Catching too general exception BaseException
```

I have to look on the other test. Looks like the test environment could be broken.

martinezjavier:
@jkonecny12 @bcl looks good to me now. And I noticed PR #1870 was also opened to get these fixes into master.

bcl:
> Please fix this pylint error:

That's in dnfpayload.py and wasn't introduced by this patchset, it should be handled separately, shouldn't it?

jkonecny12:
Ah you are correct. I guess this happened thanks to the new pylint or something similar. You can ignore that.

jkonecny12:
jenkins, test this please

jkonecny12:
One test is fixed. One more to go...

jkonecny12:
jenkins, test this please

jikortus:
@jkonecny12 Is there anything that blocks merging this PR (especially for rhel-8)? Thanks.

poncovka:
Tested automated installation with `liveimg` using image and tar.

jikortus:
@poncovka @rvykydal Thanks for taking care of this.

PR #1871, opened 2019-03-06
Handle missing support for Blivet-GUI in the Storage module (#1685645)

poncovka:
The Blivet partitioning module should be able to exist without support
for Blivet-GUI. In this case, the method `SendRequest` will raise the
`UnsupportedPartitioningError` exception.

In the future, we will raise this exception once the Blivet
partitioning is requested.

Resolves: rhbz#1685645


AdamWill:
I ran openQA tests on a scratch build with this patch applied, [all looks good](https://openqa.stg.fedoraproject.org/tests/overview?distri=fedora&version=30&build=Kojitask-33236649-NOREPORT&groupid=2).

PR #1872, opened 2019-03-06
Simplify the storage reset

poncovka:
Simplify the code that handles the storage reset.

### DIFF
--- a/pyanaconda/storage/initialization.py
+++ b/pyanaconda/storage/initialization.py
@@ -23,6 +23,7 @@ from blivet import util as blivet_util, udev, arch
 from blivet.errors import StorageError
 from blivet.flags import flags as blivet_flags
 from blivet.iscsi import iscsi
+from blivet.storage_log import log_exception_info
 
 from pyanaconda.anaconda_logging import program_log_lock
 from pyanaconda.core.configuration.anaconda import conf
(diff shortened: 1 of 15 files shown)


pyanaconda/storage/initialization.py
```diff
@@ -137,6 +138,11 @@ def reset_storage(storage, teardown=False, retry=True):
         except Exception:  # pylint: disable=broad-except
             log_exception_info(log.error, "Failure tearing down device tree.")
 
+    # Clear the exclusive disks to scan all devices in the system.
+    if scan_all:
+        disk_select_proxy = STORAGE.get_proxy(DISK_SELECTION)
+        disk_select_proxy.SetExclusiveDisks([])
```
jkonecny12: I would be cautious about this. In general any device which is not supported by blivet may crash the system.
If I'm not mistaken all the disks in the protected and excluded list should be scanned by blivet or we will crash in some situations.

poncovka: This only replaces the workaround with selected disks. We already scan all devices if the storage configuration fails and if a user requests to rescan a storage in TUI.

The only place I have changed is the logic for the Rescan button in GUI to make it consistent with TUI. However, I am ready to revert this change if it causes problems.

poncovka:
Rebased.

pyanaconda/storage/initialization.py
jkonecny12: I won't block merging of this but if I'm not mistaken for example if you have broken VG group (missing disk) the blivet library will fail on scan. Now this behavior can be work-around by using the `ignoredisk` command in kickstart. If I'm not mistaken then this won't work now for this part of the code.

You can merge this now but please test how it behaves on broken VG (the easiest test I know) with ignoreddisks command. If didn't introduce a regression by this and if yes then please make a follow-up PR.

poncovka:
Fixed the function `_reset_storage`.

pyanaconda/storage/initialization.py
poncovka: This pull request should not affect the logic of the `ignoredisk` command at all. Merging.

PR #1873, opened 2019-03-06
payload: don't force host-only mode when executing dracut

martinezjavier:
When the initramfs are generated on kernel installs, dracuts default for
enabling or disabling host-only mode is used. But when recreating them
at the end of the installation, dracut is executed with host-only mode.

This means that host-only mode can't be disabled (i.e: by installing the
dracut-config-generic package) which is needed to create aarch64 images.

Also don't force the persistent policy to be by-uuid and use whatever is
the default, which is also the case when initrds are generated on kernel
package installs.

### DIFF
--- a/pyanaconda/payload/__init__.py
+++ b/pyanaconda/payload/__init__.py
@@ -587,8 +587,7 @@ class Payload(metaclass=ABCMeta):
                 if use_dracut:
                     util.execInSysroot("depmod", ["-a", kernel])
                     util.execInSysroot("dracut",
-                                       ["-H", "--persistent-policy", "by-uuid",
-                                        "-f",
+                                       ["-f",
                                         "/boot/initramfs-%s.img" % kernel,
                                         kernel])
                 else:


jkonecny12:
Jenkins, it's ok to test.

jkonecny12:
This PR should solve https://bugzilla.redhat.com/show_bug.cgi?id=1686326

martinezjavier:
> This PR should solve https://bugzilla.redhat.com/show_bug.cgi?id=1686326

Indeed, I've included the Resolves: tag in the commit message now.

jkonecny12:
I've just did a smoke test (I don't have an ARM device) and it looks it works good so after the tests will finish. I'm merging this.

jkonecny12:
Did also a check proposed by @martinezjavier .
After the installation with `dracut-config-generic` package the initrd have 46M with the patch and 19M without it.
So the without the patch it is created host only specific which is wrong.

PR #1874, opened 2019-03-07
efi: don't include the grub2-pc package on EFI installs

martinezjavier:
The GRUB2() class adds the grub2 package to the list of packages to be
installed and the EFIGRUB() class inherits this which leads to grub2-pc
package being installed even on EFI installs. That's wrong because this
isn't needed for EFI and also causes a broken /etc/grub2.cfg symlink to
be created.

### DIFF
--- a/pyanaconda/bootloader/efi.py
+++ b/pyanaconda/bootloader/efi.py
@@ -133,7 +133,7 @@ class EFIBase(object):
 class EFIGRUB(EFIBase, GRUB2):
     """EFI GRUBv2"""
     _packages32 = [ "grub2-efi-ia32", "shim-ia32" ]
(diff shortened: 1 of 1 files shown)


jkonecny12:
If you do want to add this to Fedora 30, please change the target to `f30-devel` branch.

jkonecny12:
Jenkins, it's ok to test.

martinezjavier:
> If you do want to add this to Fedora 30, please change the target to `f30-devel` branch.

done, thanks for the pointer.

poncovka:
Jenkins, test this please.

poncovka:
Looks good to me. Thanks!

PR #1881, opened 2019-03-11
Validate post-installation snapshot requests with the storage checker

poncovka:
Don't validate the post-installation snapshot requests in the function
`do_kickstart_storage`. Register a new callback with the storage
checker, so the verification will be done as part of the storage checking.

We don't need to validate the pre-installation snapshot requests,
because there is no way how to handle the validation error. Since
the post-installation snapshot requests are validated by the storage
checker now, we can remove the `SnapshotValidateTask` class.

Don't run the storage checker to check the result of autopart.
The storage checker will be run after the storage configuration.
It doesn't make sense to run it just to check the result of the
autopart. The errors and warnings don't have to be related to that.


poncovka:
> However, I don't agree with the pre-installation snapshot validation removal. I understand why you want to do this but I don't think it is a good way to go. The pre-installation snapshots could be used mainly as a backup and by removing their validation then we are not able to tell if the your backup works before it is re-installed during the installation.

The `SnapshotValidateTask` runs basically the same code as `SnapshotCreateTask`, so the the pre-installation snapshot requests are still validated. I can reorganize the code a little to make sure that all requests are valid before we start to create the snapshots. Otherwise, I don't see why we should run a task that does the same as the next one we run immediately after that.

> I even think that in non-interactive installation mode this check should break the installation and not even start.

I don't understand this comment.

jkonecny12:
> > However, I don't agree with the pre-installation snapshot validation removal. I understand why you want to do this but I don't think it is a good way to go. The pre-installation snapshots could be used mainly as a backup and by removing their validation then we are not able to tell if the your backup works before it is re-installed during the installation.
> 
> The `SnapshotValidateTask` runs basically the same code as `SnapshotCreateTask`, so the the pre-installation snapshot requests are still validated. I can reorganize the code a little to make sure that all requests are valid before we start to create the snapshots. Otherwise, I don't see why we should run a task that does the same as the next one we run immediately after that.

Ahh OK, I see that now. This is validation before the snapshot are created not after the creation. In that case this is fine. You have an ACK.

> 
> > I even think that in non-interactive installation mode this check should break the installation and not even start.
> 
> I don't understand this comment.
This part is not important we can ignore that for the sake of simplicity.

PR #1882, opened 2019-03-12
payload: separate blivet.utils & blivet.arch call

jkonecny12:
First step in separating all the storage calls to one place.

### DIFF
--- a/pyanaconda/payload/__init__.py
+++ b/pyanaconda/payload/__init__.py
@@ -39,15 +39,13 @@ from pyanaconda.image import opticalInstallMedia, verifyMedia, verify_valid_inst
 from pyanaconda.core.util import ProxyString, ProxyStringError
 from pyanaconda.core.regexes import VERSION_DIGITS
 from pyanaconda.payload.errors import PayloadError, PayloadSetupError, NoSuchGroup
-from pyanaconda.payload.utils import version_cmp
+from pyanaconda.payload import utils as payload_utils
 from pyanaconda.payload.install_tree_metadata import InstallTreeMetadata
 from pyanaconda.payload.requirement import PayloadRequirements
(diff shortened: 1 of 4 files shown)


pyanaconda/payload/__init__.py
```diff
@@ -38,6 +38,7 @@
 from pyanaconda.image import opticalInstallMedia, verifyMedia, verify_valid_installtree
 from pyanaconda.core.util import ProxyString, ProxyStringError
 from pyanaconda.core.regexes import VERSION_DIGITS
+from pyanaconda.payload import storage as payload_storage
```
poncovka: Please, move the functions to `pyanaconda.payload.utils`, this is too confusing.

pyanaconda/payload/storage.py
```diff
@@ -0,0 +1,61 @@
...
+import blivet.util
+import blivet.arch
+
+from pyanaconda.anaconda_loggers import get_module_logger
+from pyanaconda.payload.errors import PayloadSetupError
+
+log = get_module_logger(__name__)
+
+
+def get_mount_device(mountpoint):
```
poncovka: Just a note, don't expect that these functions will be provided by the Storage module. It is a very basic functionality that doesn't require a device tree or any additional storage support.

pyanaconda/payload/__init__.py
jkonecny12: Will do.

pyanaconda/payload/storage.py
jkonecny12: We can discuss this later. It is definitely not a priority now.

jkonecny12:
**UPDATED:**
*  Moved to payload.utils instead of payload.storage

PR #1885, opened 2019-03-13
Simplify the partitioning

poncovka:
Move the code for the storage partitioning to the Storage module and
drop `autopart`-related attributes from the `InstallerStorage` class.


pyanaconda/storage/autopart.py
```diff
@@ -36,76 +34,11 @@
 from pykickstart.constants import AUTOPART_TYPE_BTRFS, AUTOPART_TYPE_LVM, AUTOPART_TYPE_LVM_THINP, AUTOPART_TYPE_PLAIN
 
 from pyanaconda.core.i18n import _
-from pyanaconda.storage.utils import get_available_disk_space
+from pyanaconda.storage.utils import get_available_disk_space, suggest_swap_size
```
jkonecny12: Change `swap_suggestion` in the 1st commit message. It could be name of the function which is `suggest_swap_size` or without `_` character.

jkonecny12: The same trend follow on some other commits. Is that intentional?

poncovka:
New tests in the last commit.

PR #1894, opened 2019-03-14
Update the kickstart data for iSCSI and NVDIMM during the installation

poncovka:
We don't need to update the kickstart data for iSCSI and NVDIMM after
every partitioning. Let's move the code to the update_storage_ksdata
function that we run during the installation.

This code should be eventually moved to the Storage module.

### DIFF
--- a/pyanaconda/storage/kickstart.py
+++ b/pyanaconda/storage/kickstart.py
@@ -15,10 +15,13 @@
 # License and may only be used or replicated with the express permission of
 # Red Hat, Inc.
 #
+from blivet.devices import MultipathDevice, iScsiDiskDevice, NVDIMMNamespaceDevice
+from blivet.iscsi import iscsi
+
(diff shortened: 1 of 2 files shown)


poncovka:
@rvykydal, could you help me with testing, please?

rvykydal:
Yes I can, I think we just need to test that
- nvdimm device reconfigured in GUI (or selected in GUI if it is already in sector mode - if we want more testing)
- iscsi device attached in GUI
ends up in generated kickstart as nvdimm or iscsi command
I can do the testing for you.

poncovka:
@rvykydal, I have figured out how to test the NVDIMM device (Anaconda had to forget that the device was reconfigured), so I need help only with iSCSI.

rvykydal:
I tested the patch with iscsi target added in GUI and the commands seem to be still generated correctly /root/anaconda-ks.cfg

poncovka:
@rvykydal, thanks!

PR #1896, opened 2019-03-15
Simplify get_free_space

poncovka:
Let's remove the code that estimates the free space based on the
clearpart settings and use the original get_free_space method of
the Blivet class that calculates the current free space of the
storage model. We don't really need to know the estimation.

Use get_disk_free_space, get_disk_reclaimable_space and
get_file_system_free_space to calculate the available space,
because later we will provide DBus support only for these
methods.

### DIFF
--- a/pyanaconda/modules/storage/partitioning/custom_partitioning.py
+++ b/pyanaconda/modules/storage/partitioning/custom_partitioning.py
@@ -37,8 +37,7 @@ from pyanaconda.modules.storage.partitioning.noninteractive_partitioning import 
 from pyanaconda.modules.storage.partitioning.schedule import get_candidate_disks, \
     schedule_partitions
 from pyanaconda.platform import platform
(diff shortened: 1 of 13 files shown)


pyanaconda/modules/storage/partitioning/custom_partitioning.py
```diff
@@ -37,7 +37,7 @@
 from pyanaconda.modules.storage.partitioning.schedule import get_candidate_disks, \
     schedule_partitions
 from pyanaconda.platform import platform
-from pyanaconda.storage.utils import get_available_disk_space, suggest_swap_size, get_pbkdf_args, \
+from pyanaconda.storage.utils import suggest_swap_size, get_pbkdf_args, \
```
jkonecny12: Please remove the `\` from this import, it should fit on one line now.

poncovka: Fixed.

PR #1898, opened 2019-03-15
Hide the encryption checkbox for the interactive partitioning

poncovka:
Users should set up the encryption in the CustomPartitioningSpoke.
It doesn't make sense to choose this option before they enter this
spoke.


poncovka:
The checkbox is visible if the automatic partitioning is selected:
auto

The checkbox will be hidden if the custom partitioning is selected:
custom

jkonecny12:
Thanks for these screenshots.

PR #1899, opened 2019-03-18
Always specify the LUKS version in CustomPartitioningSpoke (#1689699)

poncovka:
We should specify the LUKS version when we add a new device in the
CustomPartitioningSpoke.

Resolves: rhzb#1689699


wbob:
this will only partly solve rhzb#1689699 _"luks-version dropdown dysfunctional in manual partitioning"_. The widget introduced in #1525 will have no effect on update if a user tries to change the luks version. A video of the behaviour (still present in current rawhide with this PR merged) is attached to the original bug ticket.

poncovka:
@wbob, the other issue might not be so trivial to fix, so I will leave your bug open and return to it later. Anyway, thanks for the video, it is very helpful!

PR #1901, opened 2019-03-18
Add DBus support for the device tree

poncovka:
Add DBus support for devices in the device tree and provide it by
the Storage module. It might be provided by some of the partitioning
modules as well in the future.

### DIFF
--- a/pyanaconda/modules/storage/devicetree/devicetree.py
+++ b/pyanaconda/modules/storage/devicetree/devicetree.py
@@ -1,0 +1,133 @@
+#
+# Handler of the device tree.
+#
+# Copyright (C) 2019 Red Hat, Inc.
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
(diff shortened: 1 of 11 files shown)


pyanaconda/modules/storage/devicetree/devicetree.py
```diff
@@ -0,0 +1,133 @@
...
+
+    def _get_device(self, name):
+        """Find a device by its name.
+
+        :param name: a name of the device
+        :return: an instance of the Blivet's device
+        """
+        return self.storage.devicetree.get_device_by_name(name, hidden=True)
+
+    def _get_device_attrs(self, device):
```
jkonecny12: Shouldn't this method be part of the `DeviceData` instead of the solution when `DeviceData` own an array which is not used only outside of the class?

poncovka: No, `DeviceData` shouldn't be able to work with Blivet's objects. I have updated the docstring of the `attrs` property to point to the `SUPPORTED_ATTRIBUTES` list.

poncovka:
Updated.

PR #1904, opened 2019-03-19
Detect and warn if SMT is enabled (#1678353)

poncovka:
Detect whether hyperthreading (SMT) is enabled on a system. If so,
we should display a warning in the info bar of the summary hub.

Resolves: rhbz#1678353


poncovka:
Updated `POTFILES.in`.

jkonecny12:
Thanks for solving this!

PR #1907, opened 2019-03-21
Remove nosslverify flag

jkonecny12:
Replace the `inst.nosslverify` option by the configuration file.

---------------------------------
Looking for advice. Based on the [documentation](https://anaconda-installer.readthedocs.io/en/latest/boot-options.html#inst-noverifyssl) it should not be used for additional repositories but based on the code it is not completely true. The point is that it will use also this flag for new additional repositories but if I remove this option then you basically can't add additional repositories from GUI with disabled ssl verification. So I leaved it the way it was before, if it is not set from the ksrepo (kickstart) then it will use this option.

What do you think, should I change it somehow?

### DIFF
--- a/data/anaconda.conf
+++ b/data/anaconda.conf
@@ -68,6 +68,9 @@ enable_closest_mirror = True
 # Check if payload supports the locales.
 check_supported_locales = False
 
+# Enable ssl verification for all HTTP connection
+verify_ssl = True
+
 
 [Security]
 # Enable SELinux usage in the installed system.
--- a/pyanaconda/payload/__init__.py
+++ b/pyanaconda/payload/__init__.py
@@ -394,7 +394,7 @@ class Payload(metaclass=ABCMeta):
         #   - the path to a cert file
         #   - True, to use the system's certificates
         #   - False, to not verify
-        ssl_verify = getattr(self.data.method, "sslcacert", not flags.noverifyssl)
+        ssl_verify = getattr(self.data.method, "sslcacert", conf.payload.verify_ssl)
 
         ssl_client_cert = getattr(self.data.method, "ssl_client_cert", None)
         ssl_client_key = getattr(self.data.method, "ssl_client_key", None)
--- a/pyanaconda/payload/dnfpayload.py
+++ b/pyanaconda/payload/dnfpayload.py
@@ -363,7 +363,7 @@ class DNFPayload(payload.PackagePayload):
             repo.mirrorlist = mirrorlist
         if metalink:
             repo.metalink = metalink
-        repo.sslverify = not (ksrepo.noverifyssl or flags.noverifyssl)
+        repo.sslverify = not ksrepo.noverifyssl and conf.payload.verify_ssl
         if ksrepo.proxy:
             try:
                 repo.proxy = ProxyString(ksrepo.proxy).url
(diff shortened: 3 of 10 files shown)


pyanaconda/payload/dnfpayload.py
```diff
@@ -364,7 +364,7 @@ def _add_repo(self, ksrepo):
             repo.mirrorlist = mirrorlist
         if metalink:
             repo.metalink = metalink
-        repo.sslverify = not (ksrepo.noverifyssl or flags.noverifyssl)
+        repo.sslverify = not (ksrepo.noverifyssl or not self.is_ssl_verification_enabled)
```
poncovka: I don't think this is correct. This will check `method.noverifyssl` for additional repositories now.

pyanaconda/payload/dnfpayload.py
```diff
@@ -1403,6 +1403,7 @@ class RepoMDMetaHash(object):
     def __init__(self, dnf_payload, repo):
         self._repoId = repo.id
         self._method = dnf_payload.data.method
+        self._ssl_verify = dnf_payload.is_ssl_verification_enabled
```
poncovka: This doesn't look correct either. It will use the value of `method.noverifyssl` for additional repositories again.

pyanaconda/payload/rpmostreepayload.py
```diff
@@ -206,7 +205,7 @@ def install(self):
         if hasattr(ostreesetup, 'nogpg') and ostreesetup.nogpg:
             self._remoteOptions['gpg-verify'] = Variant('b', False)
 
-        if flags.noverifyssl:
+        if not self.is_ssl_verification_enabled:
```
poncovka: The kickstart command `ostreesetup` is not in `data.method` and doesn't have the `noverifyssl` attribute anyway.

poncovka:
> Looking for advice. Based on the [documentation](https://anaconda-installer.readthedocs.io/en/latest/boot-options.html#inst-noverifyssl) it should not be used for additional repositories but based on the code it is not completely true. The point is that it will use also this flag for new additional repositories but if I remove this option then you basically can't add additional repositories from GUI with disabled ssl verification. So I leaved it the way it was before, if it is not set from the ksrepo (kickstart) then it will use this option.
> 
> What do you think, should I change it somehow?

From the code, it looks like `inst.nosslverify` really works as a global option, so I would update the documentation and continue to use it that way.

data/anaconda.conf
```diff
@@ -68,6 +68,9 @@ enable_closest_mirror = True
 # Check if payload supports the locales.
 check_supported_locales = False
 
+# Disable ssl verification for all HTTP connection with exception to additional repositories.
+no_verify_ssl = False
```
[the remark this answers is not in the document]
jkonecny12: Fixed

pyanaconda/payload/__init__.py
```diff
@@ -81,6 +81,23 @@ def __init__(self, data):
     def first_payload_reset(self):
         return self._first_payload_reset
 
+    @property
+    def is_ssl_verification_enabled(self):
```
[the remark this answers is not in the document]
jkonecny12: I would say it makes a standard because thanks to the fact that inst.nosslverify is set to the ``ksdata.method`` and then it doesn't make sense to use the flag with the method later.
However, you are right that this will be solved better later. So it is removed.

pyanaconda/payload/dnfpayload.py
jkonecny12: True. Removed.

pyanaconda/payload/dnfpayload.py
jkonecny12: This is now handled completely other way. I think it should be correct now.

pyanaconda/payload/rpmostreepayload.py
jkonecny12: That shouldn't be a problem. The method is skipped in case it is not present.

jkonecny12:
**UPDATED:**
* This version won't change the code only replace flags by conf.
* Removed the negation from configuration so it is now `sslverify`.

PR #1908, opened 2019-03-21
Remove askmethod flag

jkonecny12:
Based on the [documentation](https://anaconda-installer.readthedocs.io/en/latest/boot-options.html#askmethod-asknetwork) this option is not used for a long time so remove this from the code completly.


jkonecny12:
**UPDATED:**
* Add more info to commit message.
* Remove also parameter parsing.

PR #1910, opened 2019-03-22
network module: add network module unit tests

rvykydal:


tests/nosetests/pyanaconda_tests/module_network_test.py
```diff
@@ -682,3 +683,170 @@ def set_disabled_services_test(self):
...
+        self.network_module = NetworkModule()
+
+    def apply_boot_options_ksdevice_test(self):
+        """Test _apply_boot_options function for 'ksdevice'."""
+        self.assertEqual(
+            self.network_module.default_device_specification,
+            DEFAULT_DEVICE_SPECIFICATION
+        )
+        mocked_kernel_args = {"something": "else"}
+        self.network_module._apply_boot_options(mocked_kernel_args)
```
jkonecny12: I would rather see using mock on flags module to get avoid calling protected API.
What do you think?

rvykydal:
After discussion with @jkonecny12 I am merging the PR as is.

PR #1912, opened 2019-03-22
Simplify the code for clearing partitions

poncovka:
Move the code for clearing partitions from the `InstallerStorage`
class to modules and simplify it.


pyanaconda/modules/storage/partitioning/custom_partitioning.py
```diff
@@ -324,8 +324,7 @@ def _execute_partition_data(self, storage, data, partition_data):
                     lineno=partition_data.lineno
                 )
 
-            should_clear = storage.should_clear(disk)
-            if disk and (disk.partitioned or should_clear):
+            if disk and disk.partitioned:
```
poncovka: @vojtechtrefny, we run this code after the disks are cleared and initialized, so I think that it should be enough to check the `partitioned` property. What do you think?

pyanaconda/modules/storage/partitioning/schedule.py
```diff
@@ -57,10 +56,6 @@ def get_candidate_disks(storage):
         if not disk.format.supported or disk.protected:
             continue
 
-        if storage.config.clear_part_disks and \
-           (disk.name not in storage.config.clear_part_disks):
-            continue
```
poncovka: @vojtechtrefny, we collect the candidate disks for autopart and reqpart after the disks are cleared and initialized and all unselected devices are hidden, so it seems safe to drop this check. What do you think?

pyanaconda/modules/storage/partitioning/custom_partitioning.py
vojtechtrefny: `partitioned` just checks if the disk is formatted to a "disklabel", so it should be enough to check only `partitioned` after disk initialization.

pyanaconda/modules/storage/partitioning/schedule.py
vojtechtrefny: Yes, it should be safe to remove this, if this functions after clearpart.

pyanaconda/modules/storage/disk_initialization/configuration.py
```diff
@@ -32,3 +36,82 @@ def __init__(self):
         self.initialize_labels = False
         self.format_unrecognized = False
         self.clear_non_existent = False
+
+    def can_remove(self, storage, device):
```
jkonecny12: Why switching the name from `should_clear` to `can_remove`? AFAIK the disk which can be passed in won't be removed it will be cleared.

pyanaconda/storage/osinstall.py
```diff
@@ -454,40 +437,40 @@ def should_clear(self, device, **kwargs):
...
-        clear_part_type = kwargs.get("clear_part_type", self.config.clear_part_type)
-        clear_part_disks = kwargs.get("clear_part_disks",
-                                      self.config.clear_part_disks)
-        clear_part_devices = kwargs.get("clear_part_devices",
-                                        self.config.clear_part_devices)
+        initialization_mode = kwargs.get("initialization_mode", self.config.initialization_mode)
+        drives_to_clear = kwargs.get("drives_to_clear",
+                                      self.config.drives_to_clear)
+        devices_to_clear = kwargs.get("devices_to_clear",
+                                        self.config.devices_to_clear)
```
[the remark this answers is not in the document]
poncovka: This code is removed in the following commits.

pyanaconda/storage/osinstall.py
```diff
@@ -454,40 +437,40 @@ def should_clear(self, device, **kwargs):
...
-                                       :attr:`self.config.clear_part_devices`
-            :type clear_part_devices: list
+            :keyword initialization_mode: overrides :attr:`self.config.initialization_mode`
+            :type initialization_mode: int
+            :keyword drives_to_clear: overrides
+                                     :attr:`self.config.drives_to_clear`
+            :type drives_to_clear: list
+            :keyword devices_to_clear: overrides
+                                       :attr:`self.config.devices_to_clear`
+            :type devices_to_clear: list
```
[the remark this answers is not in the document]
poncovka: This code is removed in the following commits.

pyanaconda/modules/storage/disk_initialization/configuration.py
```diff
@@ -0,0 +1,34 @@
+#
+# Configuration of the disk initialization
```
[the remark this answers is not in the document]
poncovka: Fixed.

pyanaconda/modules/storage/disk_initialization/configuration.py
poncovka: I have decided to use the names `can_remove` and `can_initialize`, because we call Blivet's methods `recursive_remove` and `initialize_disk`. The cleared disks are removed and initialized from the Blivet's perspective, so these names better reflect the implementation. These methods are not going to be exposed on DBus, so it is just an implementation detail.

poncovka:
Jenkins, test this please.

pyanaconda/modules/storage/disk_initialization/configuration.py
jkonecny12: Truth is it is an implementation detail so I won't block this PR because of that.
However, my point is that we shouldn't use names based on the library implementation we should go opposite way. The whole point in using the library is that we want abstraction and by naming things based on how they are implemented you are loosing this abstraction level.

Just keep this in mind, you don't need to change the code here.

poncovka:
Jenkins, test this please.

poncovka:
Jenkins, test this please.

PR #1913, opened 2019-03-25
Disable updates-testing (#1670091)

poncovka:
Always disable `updates-testing` and `updates-testing-modular` repos.
Only `updates` and `updates-modular` repos should be controlled by the
checkbox of the installation source spoke.

Resolves: rhbz#1670091


kparal:
Thanks for merging this. Getting a new anaconda build before F30 Final would be great.

poncovka:
> Thanks for merging this. Getting a new anaconda build before F30 Final would be great.

@kparal, that's the plan :-)

PR #1892, opened 2019-03-14
Prepare for fips-mode-setup being split out of crypto-policies

keszybz:

PR #1916, opened 2019-03-25
[F30] Prepare for fips-mode-setup being split out of crypto-policies

keszybz:
Same as https://github.com/rhinstaller/anaconda/pull/1892.

### DIFF
--- a/pyanaconda/installation.py
+++ b/pyanaconda/installation.py
@@ -331,6 +331,9 @@ def doInstall(storage, payload, ksdata):
 
         if can_install_bootloader:
             payload.requirements.add_packages(storage.bootloader.packages, reason="bootloader")
+        if flags.flags.cmdline.getbool("fips"):
+            payload.requirements.add_packages(['/usr/bin/fips-mode-setup'], reason="bootloader")
+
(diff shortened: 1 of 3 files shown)


jkonecny12:
Jenkins, it's ok to test.

jkonecny12:
You have to solve this pylint issue before merging:
```
E1101(no-member):334,11: doInstall.run_pre_install: Module 'pyanaconda.flags' has no 'cmdline' member
```

pyanaconda/installation.py
```diff
@@ -331,6 +331,9 @@ def run_pre_install():
 
         if can_install_bootloader:
             payload.requirements.add_packages(storage.bootloader.packages, reason="bootloader")
+        if flags.cmdline.get("fips") == "1":
```
jkonecny12: To make this work you have to change the code to this.
```suggestion
        if flags.flags.cmdline.getbool("fips"):
```

Sorry for the ``flags.flags`` that is just a wrong import but we will fix that in the future later.

keszybz:
Thanks for the suggestion. Updated.

PR #1919, opened 2019-03-26
Update the protected devices in the Storage module

poncovka:
We need to update the protected devices in the storage model of the
Storage module, when the list of protected devices is changed.

### DIFF
--- a/pyanaconda/modules/storage/storage.py
+++ b/pyanaconda/modules/storage/storage.py
@@ -121,6 +121,9 @@ class StorageModule(KickstartModule, DeviceTreeHandler):
         self.storage_changed.connect(
             self._bootloader_module.on_storage_reset
         )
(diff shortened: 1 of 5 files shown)


pyanaconda/modules/storage/storage.py
```diff
@@ -177,6 +182,13 @@ def set_storage(self, storage):
         self.storage_changed.emit(storage)
         log.debug("The storage model has changed.")
 
+    def on_protected_devices_changed(self, protected_devices):
+        """Update the protected devices in the storage model."""
+        if not self._storage:
+            return
```
jkonecny12: Please at least log that nothing happened.

poncovka: I have added logging to the method `protect_devices` to log that something happened.

poncovka:
Jenkins, test this please.

PR #1922, opened 2019-03-27
Add DBus support for disk selection

poncovka:
Move the code from the function `get_available_disks` to the
`InstallerStorage` class. Use the property `usable_disks` to get a
list of disks usable for the installation and the method `select_disks`
to select the disks that should be used for the installation.

We need to hide the unselected disks before we start the partitioning.
Therefore, we can apply the selection only on the copy of the storage
model and leave the original model unchanged.

Call `GetUsableDisks` of the disk selection module to get a list
of disk names that can be used for the installation.


poncovka:
Rebased.

PR #1926, opened 2019-04-01
Give payload a hook for adding disabled repos.

jcpunk:
I'm beginning to rework my anaconda addon for the dbus interface and the new products.d directory.  The ability to easily add disabled repos through anaconda cuts a significant number of lines from my code base and would probably be helpful to others too.

### DIFF
--- a/pyanaconda/payload/__init__.py
+++ b/pyanaconda/payload/__init__.py
@@ -239,6 +239,16 @@ class Payload(metaclass=ABCMeta):
         ksrepo.enabled = True
         self.data.repo.dataList().append(ksrepo)
 
+    def add_disabled_repo(self, ksrepo):
+        """Add the repo given by the pykickstart Repo object ksrepo to the
+        list of known repos.  The repo will be automatically disabled.
+
+        Duplicate repos will not raise an error.  They should just silently
+        take the place of the previous value.
+        """
+        ksrepo.enabled = False
+        self.data.repo.dataList().append(ksrepo)
+
     def remove_repo(self, repo_id):
         repos = self.data.repo.dataList()
         try:


jkonecny12:
Jenkins, it's ok to test.

jkonecny12:
If you don't need this API now then please wait, I'm working on the payload module and we can add this directly there instead of adding this now and moving it later.
Is this a priority or do you have a dealine @jcpunk ?

jcpunk:
This is something I would begin using immediatly. It is at the core of what
I'm currently doing.

Is there a rough ETA on the payload changes you are building?

jkonecny12:
Hi,
It will take some time. I have to create tests first because right now there is easy to get burn by missing something or by some side effect. So it will take some time (my rough ETA could be about 4 months but that is really just my guess).

Do you have your code upstream somewhere, so at least I could make a PRs to your codebase if something will change?

Thanks for your patient.

jcpunk:
Right now the extension code isn't public because it is way too ugly....  I can work on getting it cleaned up for more public consumption.

In the mean time, can you merge this PR while I get my end cleaner?  This way we are working from both ends (and the behavior doesn't get missed in the port over).

pyanaconda/payload/dnfpayload.py
```diff
@@ -446,6 +446,16 @@ def add_repo(self, ksrepo):
...
+    def add_disabled_repo(self, ksrepo):
+        """Add a disabled repo to dnf and kickstart repo lists.
+
+        :param ksrepo: Kickstart Repository to add
+        :type ksrepo: Kickstart RepoData object.
+        :returns: None
+        """
+        ksrepo.enabled = False
+        super().add_disabled_repo(ksrepo)
+
```
jkonecny12: In this case you can remove this dnfpayload code completely. It will call the same code from the `payload/__init__.py` so there is no reason to overload this method.

The only problem is that the repo won't be checked on validity before it is enabled but that is I guess even benefit because it won't slow down the installation in case the user don't want to use that repository.

jcpunk: I'll have it fixed shortly.

jcpunk:
Updated per request.

PR #1928, opened 2019-04-01
We inherit self.data.repo.dataList() from parent class.

jcpunk:
Avoiding a direct call to permit clean change out if needed.

If payload decides to store things a different way, we shouldn't need to edit the specific instance class.

Semi-related:  I noticed https://github.com/rhinstaller/anaconda/commit/9a02cda015672a9eab3ce8d956ed1e05f2b63b61 set `enabled=True` for addon repos (then inherited by https://github.com/rhinstaller/anaconda/commit/57c6786b656ec57a84a4cf88c1b16cd16c2c3660).  Wasn't this `enabled=False` historically?

### DIFF
--- a/pyanaconda/payload/dnfpayload.py
+++ b/pyanaconda/payload/dnfpayload.py
@@ -1289,15 +1289,16 @@ class DNFPayload(payload.PackagePayload):
             if base_repo_url is not None:
                 existing_urls.append(base_repo_url)
 
-            for ksrepo in self.data.repo.dataList():
-                existing_urls.append(ksrepo.baseurl)
+            for ksrepo in self.addons:
+                baseurl = self.get_addon_repo(ksrepo).baseurl
+                existing_urls.append(baseurl)
 
             for repo_md in self._install_tree_metadata.get_metadata_repos():
                 if repo_md.path not in existing_urls:
                     repo = RepoData(name=repo_md.name, baseurl=repo_md.path,
                                     install=False, enabled=True)
                     repo.treeinfo_origin = True
-                    self.data.repo.dataList().append(repo)
+                    self.add_repo(repo)
 
         return install_tree_url
 


jkonecny12:
Jenkins, it's ok to test.

jcpunk:
> The feature enables you to add url which has a `.treeinfo` file with multiple repositories, so by adding this url you will add multiple repositories at once.

Would extending the `.treeinfo` format to note "default enabled" and "default disabled" be helpful?  The spec seems not to specify 
https://productmd.readthedocs.io/en/latest/treeinfo-1.1.html

jkonecny12:
If you want to go this way, I think it would be better to use repo files which has this functionality now. You would just add repo files with your addon to the system and we would work with that as expected.

However, even this variant means that we have to adjust Fedora boot.iso creation, because right now they are all enabled by default.

jcpunk:
The big driver of my `.treeinfo` interest is creating clean entry points into various CentOS SIGs.  If a SIG appears after the release of the install media, it wouldn't be on the media.  If one gets retired, the artifacts on the media might be confusing (or even break stuff?).  But if they are using a netinstall, we would be able to inject the repos as needed.

On the enabled/disabled front, I do feel that the spec should have some way of indicating the expected default state of a repo.  I dislike the ambiguity... but that could just be in my head.

jkonecny12:
Understood, yeah it gives sense from the SIG point of view.

If you really want to get this in then you have to start discussion about changing the productmd specs. I would start on the fedora-devel mailing list. If they will extend the treeinfo specs then we can implement that into Anaconda.

jkonecny12:
The drawback I see in this, is that it won't be tested properly and we are not able to fix errors which may happen thanks to the .treeinfo changes. So I'm not that convinced it is worth the benefit.

I would like to know if there is an interest in this use-case on the first place.

jcpunk:
I think if I can get a few changes into `.treeinfo` that would help drive some of the ideas from https://www.redhat.com/archives/anaconda-devel-list/2018-April/msg00010.html in terms of UI.

I can reach out to the productmd folks.

jkonecny12:
I finally was able to create a script which will make a unified ISO so I was able to test this. It works fine.

Thanks for your contribution!

PR #1929, opened 2019-04-02
network: for vlan set DEVICE only if interfacename is specified in ki…

rvykydal:
…ckstart


rvykydal:
I need to fix the code and the tests.

rvykydal:
Should be OK now.

PR #1930, opened 2019-04-02
Change fips package requirement reason

jkonecny12:
Reason was set as bootloader but that is not accurate reason. So changing to compliance for better clarity.

Minor improvement of PR https://github.com/rhinstaller/anaconda/pull/1916 .


jkonecny12:
jenkins, test this please

PR #1932, opened 2019-04-03
Fix the string representation of the data objects

poncovka:
We shouldn't implicitly set the `__repr__ `methods of the data classes
decorated with the `@dbus_structure` decorator, because then it is
difficult to override them. We should define these methods explicitly.

We should skip attributes with sensitive data in the string representations
of the data classes. For example, passwords and passphrases.

### DIFF
--- a/pyanaconda/dbus/structure.py
+++ b/pyanaconda/dbus/structure.py
@@ -21,7 +21,8 @@ from typing import get_type_hints
 
 from pyanaconda.dbus.typing import get_variant, Structure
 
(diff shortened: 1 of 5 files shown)


poncovka:
> Couldn't we create a parent class which will be used for inheritance and which will implement the `__repr__` method instead of implementing that in the every class?

This is a quick fix that is required by the Users module. I am rethinking the support for DBus structures, but it will take a while, so I would like to do this first.

PR #1933, opened 2019-04-04
network: fix copying of resolv.conf for cases where target /etc does …

rvykydal:
…not exist

For example on Silverblue.

We used to check for this case before it was removed in cleanup in
8bdcce8b62d54770ae175e5809a37b8ddae05bd0


rvykydal:
I've just added BZ reference to the commit message.

PR #1935, opened 2019-04-05
Handle bytes and strings from RPM (#1693766)

poncovka:
RPM can return bytes or strings, so let's use the function
decode_bytes that can handle both and always returns a string.

Resolves: rhbz#1693766


pmatilai:
Looks correct to me. And thanks for tackling this so quickly!

PR #1936, opened 2019-04-05
network tui: fix updating of a connection from UI (#1692677)

rvykydal:
The patch should fix the change in
commit 1740cc1e39e8a2e1de2fb8b69d141545dd9323ea

The connection should be checked and upated once all the values are set, not
after change of each value.


pyanaconda/ui/tui/spokes/network.py
```diff
@@ -563,26 +576,8 @@ def indirect(self):
         return True
 
     def apply(self):
```
poncovka: Btw. I think that the problem is that the `apply` method is called after every processed user input. You could change the `input` method and call `apply` only `if key == Prompt.CONTINUE`.

rvykydal: I'll update the PR to do so.

rvykydal: Done

PR #1940, opened 2019-04-08
network: create systemd .link files for ifname= boot options (#1695894)

rvykydal:
We can't rely on initscripts renaming a device based on DEVICE and HWADDR
values in ifcfg file as initscripts may not be installed. Instead, use systemd
.link configuration file mechanism.


pyanaconda/modules/network/installation.py
```diff
@@ -106,6 +120,28 @@ def _write_sysconfig_network(self, root, overwrite):
...
+        :type root: str
+        :param ifname_option_values: list of ifname boot option values
+        :type ifname_option_values: list(str)
+        :param overwrite: overwrite existing configuration file
+        :type overwrite: bool
+        """
+        for ifname_value in ifname_option_values:
+            iface, mac = ifname_value.split(":", 1)
+            content = self.INTERFACE_RENAME_FILE_CONTENT_TEMPLATE.format(mac, iface)
+            file_path = self.INTERFACE_RENAME_FILE_TEMPLATE.format(iface)
```
jkonecny12: Just an idea.
Wouldn't it be better to have methods to create a string from these templates. These methods could have a documented arguments so it would be easier to not miss `{}` in a templates.

PR #1941, opened 2019-04-09
network: do not pass None value from NM device object to data holder …

rvykydal:
…(#1695967)

It would crash when generating dbus structure from the object.
https://bugzilla.redhat.com/show_bug.cgi?id=1695967#c4

### DIFF
--- a/pyanaconda/modules/common/structures/network.py
+++ b/pyanaconda/modules/common/structures/network.py
@@ -116,13 +116,13 @@ class NetworkDeviceInfo(object):
         :param device: NetworkManager device object
         :type device: NMDevice
         """
-        self._device_name = device.get_iface()
-        self._device_type = device.get_device_type()
+        self._device_name = device.get_iface() or ""
+        self._device_type = device.get_device_type() or self.DEVICE_TYPE_UNKNOWN
         try:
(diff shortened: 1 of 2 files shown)


pyanaconda/modules/common/structures/network.py
```diff
@@ -116,13 +116,13 @@ def set_from_nm_device(self, device):
...
-        self._device_type = device.get_device_type()
+        self._device_name = device.get_iface() or ""
+        self._device_type = device.get_device_type() or self.DEVICE_TYPE_UNKNOWN
         try:
-            hw_address = device.get_permanent_hw_address()
+            perm_hw_address = device.get_permanent_hw_address()
         except AttributeError:
-            hw_address = device.get_hw_address()
-        self.hw_address = hw_address
+            perm_hw_address = None
```
jkonecny12: It could be interesting to have a log here.

rvykydal: Virtual devices don't have permanent hw address so I think this would produce a lot of useless info.

jkonecny12: OK. Thanks for info.

PR #1942, opened 2019-04-10
Blivet-GUI should wait for the storage threads to finish (#1696478)

poncovka:
At this moment, we run the execute method of the Storage spoke before
we skip to the Blivet-GUI spoke. Then the Blivet-GUI spoke should wait
for the execute method to finish.

Resolves: rhbz#1696478


poncovka:
@vojtechtrefny, can you test it please? This is the simple fix that can go the Fedora 30, I will implement the other fix on rawhide.

PR #1943, opened 2019-04-10
Extend the function generate_string_from_data

poncovka:
Set the argument 'add' to add other values to the string
representation. The attributes in the string representation
will be sorted by their names.

### DIFF
--- a/pyanaconda/dbus/structure.py
+++ b/pyanaconda/dbus/structure.py
@@ -207,17 +207,23 @@ def generate_fields(cls):
     return fields
 
 
-def generate_string_from_data(obj, skip=None):
+def generate_string_from_data(obj, skip=None, add=None):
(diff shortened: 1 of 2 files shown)


pyanaconda/dbus/structure.py
```diff
@@ -207,17 +207,23 @@ def generate_fields(cls):
...
 
 
-def generate_string_from_data(obj, skip=None):
+def generate_string_from_data(obj, skip=None, add=None):
     """Generate a string representation of a data object.
 
     Set the argument 'skip' to skip attributes with sensitive data.
 
+    Set the argument 'add' to add other values to the string
+    representation. The attributes in the string representation
```
M4rtinK: I suggest just "sorted alphabetically" as this is just a list of strings which don't have a name.

PR #1945, opened 2019-04-10
Add the class DBusData

poncovka:
Classes derived from the `DBusData` class should represent specific
types of DBus structures. They will support a conversion from a DBus
structure of the specified type to a Python object and back.

Replace the decorator `@dbus_structure` with the class `DBusData`,
the function `apply_structure` with the method `from_structure` and
the function `get_structure` with the method `to_structure`.


poncovka:
Rebased to master.

PR #1949, opened 2019-04-16
Make default state of addon repos configurable by product

jcpunk:
For Scientific Linux we are planning to make use of the addon repos for optional software.  So I've added a flag where a product or variant can define if it wants them enabled or disabled by default.

I set Fedora to disabled (the pre https://github.com/rhinstaller/anaconda/commit/9a02cda015672a9eab3ce8d956ed1e05f2b63b61 behavior) as Fedora doesn't use addon repos in its `.treeinfo`

### DIFF
--- a/data/anaconda.conf
+++ b/data/anaconda.conf
@@ -64,6 +64,15 @@ ignored_packages =
 # Enable installation of latest updates.
 enable_updates = True
 
+# List of .treeinfo variant types to enable.
+# Valid items:
+#
+#   addon
+#   optional
+#   variant
+#
+enabled_repositories_from_treeinfo = addon optional variant
+
 # Enable installation from the closest mirror.
 enable_closest_mirror = True
 
--- a/pyanaconda/core/configuration/payload.py
+++ b/pyanaconda/core/configuration/payload.py
@@ -52,6 +52,16 @@ class PayloadSection(Section):
         return self._get_option("enable_updates", bool)
 
     @property
+    def enabled_repositories_from_treeinfo(self):
+        """List of .treeinfo variant types to enable
+
+        This flag controls which .treeinfo variant repos Anaconda
+        will automatically enable.  For a list of valid types see
+        .treeinfo documentation.
+        """
+        return self._get_option("enabled_repositories_from_treeinfo", str).split()
+
+    @property
     def enable_closest_mirror(self):
         """Enable installation from the closest mirror.
 
--- a/pyanaconda/payload/dnfpayload.py
+++ b/pyanaconda/payload/dnfpayload.py
@@ -1301,10 +1301,14 @@ class DNFPayload(payload.PackagePayload):
                 baseurl = self.get_addon_repo(ksrepo).baseurl
                 existing_urls.append(baseurl)
 
+            enabled_repositories_from_treeinfo = conf.payload.enabled_repositories_from_treeinfo
+
             for repo_md in self._install_tree_metadata.get_metadata_repos():
                 if repo_md.path not in existing_urls:
+                    repo_treeinfo = self._install_tree_metadata.get_treeinfo_for(repo_md.name)
+                    repo_enabled = repo_treeinfo.type in enabled_repositories_from_treeinfo
                     repo = RepoData(name=repo_md.name, baseurl=repo_md.path,
-                                    install=False, enabled=True)
+                                    install=False, enabled=repo_enabled)
                     repo.treeinfo_origin = True
                     self.add_repo(repo)
 
--- a/pyanaconda/payload/install_tree_metadata.py
+++ b/pyanaconda/payload/install_tree_metadata.py
@@ -199,6 +199,10 @@ class InstallTreeMetadata(object):
 
         return None
 
+    def get_treeinfo_for(self, variant_name):
+        """Return the productmd.Variant object for variant_name."""
+        return self._tree_info.variants[variant_name]
+
     def _read_variants(self):
         for variant_name in self._tree_info.variants:
             variant_object = self._tree_info.variants[variant_name]


jkonecny12:
Jenkins, it's ok to test.

jcpunk:
circling back around to this....

data/product.d/scientific-linux.conf
```diff
@@ -8,6 +8,7 @@ product_name = Red Hat Enterprise Linux
 
 [Payload]
 enable_updates = True
+enable_addon_repos = False
```
poncovka: Please, change the name of the option to `enable_repositories_from_treeinfo`.

pyanaconda/core/configuration/payload.py
```diff
@@ -61,6 +61,17 @@ def enable_closest_mirror(self):
...
         return self._get_option("enable_closest_mirror", bool)
 
+    @property
+    def enable_addon_repos(self):
+        """Enable installation from .treeinfo addon repos.
+
+        This flag controls whether or not Anaconda will automatically
+        enable any addon repos defined in .treeinfo for your product.
+
+        The use of addon repos is disabled by default.
```
poncovka: Please, use `additional repositories` instead of `addon repos`.

pyanaconda/payload/dnfpayload.py
```diff
@@ -1291,8 +1291,12 @@ def _add_treeinfo_repositories(self, install_tree_url, base_repo_url=None):
...
             for repo_md in self._install_tree_metadata.get_metadata_repos():
                 if repo_md.path not in existing_urls:
-                    repo = RepoData(name=repo_md.name, baseurl=repo_md.path,
-                                    install=False, enabled=True)
+                    if conf.payload.enable_addon_repos:
+                        repo = RepoData(name=repo_md.name, baseurl=repo_md.path,
+                                        install=False, enabled=True)
+                    else:
+                        repo = RepoData(name=repo_md.name, baseurl=repo_md.path,
+                                        install=False, enabled=False)
```
poncovka: Please, create a new instance of `RepoData` with `enabled=conf.payload.enable_addon_repos` to avoid the branching.

pyanaconda/core/configuration/payload.py
```diff
@@ -61,6 +61,17 @@ def enable_closest_mirror(self):
         """
         return self._get_option("enable_closest_mirror", bool)
 
+    @property
+    def enable_addon_repos(self):
+        """Enable installation from .treeinfo addon repos.
+
+        This flag controls whether or not Anaconda will automatically
+        enable any addon repos defined in .treeinfo for your product.
```
poncovka: Please, mention that the base repository is always enabled.

data/product.d/fedora.conf
```diff
@@ -6,6 +6,9 @@ product_name = Fedora
 [Network]
 default_on_boot = FIRST_WIRED_WITH_LINK
 
+[Payload]
+enable_addon_repos = False
```
poncovka: Please, set this value in `anaconda.conf` to `True` and drop the changes in `fedora.conf` and `rhel.conf`. We would like to keep these repositories enabled in Fedora.

data/product.d/scientific-linux.conf
jcpunk: I'll make this change shortly.

pyanaconda/payload/dnfpayload.py
jcpunk: I'll make this change shortly.

pyanaconda/core/configuration/payload.py
jcpunk: I'll make this change shortly.

pyanaconda/core/configuration/payload.py
jcpunk: I'll make this change shortly.

data/product.d/fedora.conf
jcpunk: I'll make this change shortly.

jcpunk:
Would it make sense to make this a bit more granular?

Per https://github.com/release-engineering/productmd there are:
- Variant
- Optional
- Addon

types of repos that can be listed within `.treeinfo`.

Folks may want `Variant` repos enabled, but not `Optional` for example.

Thoughts?

jkonecny12:
> Would it make sense to make this a bit more granular?
> 
> Per https://github.com/release-engineering/productmd there are:
> 
> * Variant
> * Optional
> * Addon
> 
> types of repos that can be listed within `.treeinfo`.
> 
> Folks may want `Variant` repos enabled, but not `Optional` for example.
> 
> Thoughts?

Yeah we were thinking about that too. However, we also have repositories added from the command line and have to differentiate the base repo which makes things even more complicated. Also I'm not sure if there is a use for anything else than the Variant now.

I would rather not complicate things by going too much in the details. What do you think @poncovka ?

pyanaconda/core/configuration/payload.py
```diff
@@ -61,6 +61,19 @@ def enable_closest_mirror(self):
...
         return self._get_option("enable_closest_mirror", bool)
 
+    @property
+    def enable_repositories_from_treeinfo(self):
+        """Enable installation from .treeinfo extra repos.
+
+        This flag controls whether or not Anaconda will automatically
+        enable any repos defined in .treeinfo for your product.
+
+        The use of additional repos from .treeinfo is disabled by default.
```
poncovka: I think that it is enabled by default, no? Maybe just remove this line completely. The defaults are defined by `anaconda.conf` and it can be difficult to synchronize them with the docstrings.

pyanaconda/core/configuration/payload.py
```diff
@@ -61,6 +61,19 @@ def enable_closest_mirror(self):
...
+
+        The BaseOS repo is always enabled.
```
poncovka: I would change it to `The base repo is ...`. I think that Fedora's base repo has a different name.

poncovka:
> I would rather not complicate things by going too much in the details.

I agree with @jkonecny12. Let's focus on the existing use cases.

poncovka:
Jenkins, test this please.

poncovka:
Jenkins, test this please.

pyanaconda/core/configuration/payload.py
jcpunk: Good call, I'll drop that line.

pyanaconda/core/configuration/payload.py
jcpunk: I'll make this change.

jcpunk:
> Let's focus on the existing use cases.

That seems viable to me.

If a need to add the more precise behavior appears, should we alter the variable name to future proof this?

jkonecny12:
> > Let's focus on the existing use cases.
> 
> That seems viable to me.
> 
> If a need to add the more precise behavior appears, should we alter the variable name to future proof this?

From my understanding of the productmd library - all the repos are variants and we don't make a difference between variants right now. So if I'm not mistaken we will load all the repositories right now not looking on a type. I don't think it's bad, maybe we can in the future have a logic that optional repositories will be disabled by default.
Which make me think, wouldn't that be a better solution for you than having a configuration option. Basically to load all the repositories and if it is optional then load it as disabled. Both variants are fine so you can tell us what do you think is a better solution for you. If the second one, then I will make sure that we won't break anything but I don't think we will.

jcpunk:
I'll need to think about this a bit....

What I know in the wild is:
RHEL8 is shipping AppStream as a `variant` on the install DVD
RHEL7 is shipping HighAvailibility as an `addon` on the install DVD

I might take a stab at another set of revisions around these....

jcpunk:
I've taken a stab at it.

jkonecny12:
> I'll need to think about this a bit....
> 
> What I know in the wild is:
> RHEL8 is shipping AppStream as a `variant` on the install DVD
> RHEL7 is shipping HighAvailibility as an `addon` on the install DVD
> 
> I might take a stab at another set of revisions around these....

You shouldn't think about RHEL 7 here. These changes won't be applied to RHEL 7 ever, if you on your own won't install RHEL 7 content by RHEL 8 Anaconda what I would not recommend.

jcpunk:
My thoughts on the older RHELs was mostly to try and follow the use case bits, should they reappear.

data/anaconda.conf
```diff
@@ -62,6 +62,11 @@ ignored_packages =
 # Enable installation of latest updates.
 enable_updates = True
 
+# Set state for repos found in .treeinfo
+enable_addon_repositories_from_treeinfo = True
+enable_variant_repositories_from_treeinfo = True
+enable_optional_repositories_from_treeinfo = False
```
poncovka: We would like to have only one option that would return a list of variant types that will be enabled. Please, enable all variant types by default.

```
# List of .treeinfo variant types to enable.
# Valid items:
#
#   variant
#   optional
#   addon
#
enabled_repositories_from_treeinfo = variant optional addon

```

It is much easier to work with a list and possibly extend the supported types in the future. You don't have to implement any validation, just split the string into a list of items (see `preserved_arguments`). Then you can easily check if `repo_treeinfo.type` is in this list.

data/anaconda.conf
```diff
@@ -64,6 +64,15 @@ ignored_packages =
...
 enable_updates = True
 
+# List of .treeinfo variant types to enable.
+# Valid items:
+#
+#   addon
+#   optional
+#   variant
+#
+enabled_repositories_from_treeinfo = addon optional variant
```
poncovka: The file `scientific-linux.conf` doesn't change the default settings. Is it correct?

pyanaconda/payload/dnfpayload.py
poncovka: Again, please avoid the branching. For example, you can use a new variable:
`repo_enabled = repo_treeinfo.type in enabled_repositories_from_treeinfo`

data/anaconda.conf
jcpunk: That is correct.

jcpunk:
I believe I've fixed the requested bits.

poncovka:
Jenkins, test this please.

poncovka:
The unit tests pass when run locally.

PR #1950, opened 2019-04-16
Handle bytes and strings from RPM (#1689909)

poncovka:
RPM can return bytes or strings, so let's use the function
decode_bytes that can handle both and always returns a string.

(cherry-picked from a commit 1c45e7b3a)

Resolves: rhbz#1689909


poncovka:
Rebased.

PR #1951, opened 2019-04-16
Fix the rescue mode (#1628653)

poncovka:
Anaconda in the rescue mode failed with a traceback if there was no
root found. Let's use containers and callbacks in a spoke to select
the root, so we can abandon the solution with arrays and indexes.

(cherry-picked from a commit 4fdb8e2)

Resolves: rhbz#1628653


poncovka:
Rebased.

PR #1952, opened 2019-04-16
Set the flag lvm_metadata_backup (#1673901)

poncovka:
The flag `lvm_metadata_backup` should be set to False for image
installations. We shouldn't use the Blivet's flag `image_install`
to check the type of the installation, because it is not set at
this point. Let's use the Anaconda's flag `imageInstall` instead.

Resolves: rhbz#1673901


rvykydal:
jenkins, test this please

jkonecny12:
This PR have to be rebased to rhel-8 branch.

poncovka:
Rebased.

PR #1955, opened 2019-04-25
get_iface_from_hwaddr: be more careful about hwaddr (#1703152)

AdamWill:
From https://bugzilla.redhat.com/show_bug.cgi?id=1703152 it
seems clear that in at least some case, we can get `None` from
`device.get_hw_address` or `device.get_permanent_hw_address`.
This falls back from permanent to non-permanent if the return
from permanent is None (or other false-y value), and protects
against the value still being None even with that fallback by
checking address's truthiness before trying to call `upper` on
it. That should make it impossible for this to blow up in the
same way.

### DIFF
--- a/pyanaconda/modules/network/nm_client.py
+++ b/pyanaconda/modules/network/nm_client.py
@@ -58,12 +58,16 @@ def get_iface_from_hwaddr(nm_client, hwaddr):
                                         NM.DeviceType.WIFI):
             try:
                 address = device.get_permanent_hw_address()
+                if not address:
+                    address = device.get_hw_address()
             except AttributeError as e:
                 log.warning("Device %s: %s", device.get_iface(), e)
                 address = device.get_hw_address()
         else:
             address = device.get_hw_address()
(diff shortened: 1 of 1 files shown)


AdamWill:
Note this is quite similar to https://github.com/rhinstaller/anaconda/pull/1941 , just in a different place. I'm guessing the same weird conditions pwhalen is able to provide trigger both bugs...or this one comes *after* the one that #1941 fixed.

pyanaconda/modules/network/nm_client.py
```diff
@@ -58,12 +58,16 @@ def get_iface_from_hwaddr(nm_client, hwaddr):
                                         NM.DeviceType.WIFI):
             try:
                 address = device.get_permanent_hw_address()
+                if not address:
+                    address = device.get_hw_address()
```
rvykydal: I'd maybe keep only get_pernmanent_hw_address in the try: block and assign None to address in the except:, then using get_hw_address() if address is None, but I think it is OK this way as well.

PR #1961, opened 2019-05-07
Don't check the storage in a separate thread

poncovka:
Let's run the storage checks in the same thread as the storage
configuration, so there will be no race conditions what might
cause a failure of the noninteractive installation.


poncovka:
Jenkins, test this please.

poncovka:
Updated based on the test run with this fix. @rvykydal, could you test it again, please?

rvykydal:
I'll test it with today's nightly run.

rvykydal:
Latest nightly results (using the updated fix) are far better. Very close to the "vnc" runs.

PR #1963, opened 2019-05-09
Fix the installation from NFS (#1708070)

poncovka:
The kickstart command nfs doesn't support the noverifyssl option.
Use the function getattr to get the value of this attribute.

Resolves: rhbz#1708070

### DIFF
--- a/pyanaconda/packaging/__init__.py
+++ b/pyanaconda/packaging/__init__.py
@@ -136,7 +136,7 @@ class SSLOptions(object):
 
     @classmethod
     def createFromMethod(cls, method):
-        sslverify = not (method.noverifyssl or flags.noverifyssl)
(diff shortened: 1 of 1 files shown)


pyanaconda/packaging/__init__.py
```diff
@@ -136,7 +136,7 @@ def sslverify(self, value):
 
     @classmethod
     def createFromMethod(cls, method):
-        sslverify = not (method.noverifyssl or flags.noverifyssl)
+        sslverify = not (getattr(method, "noverifyssl", False) or flags.noverifyssl),
```
jkonecny12: Why you have the `,` at the end?

poncovka: Fixed.

PR #1964, opened 2019-05-09
Use a DBus structure for the mount point requests

poncovka:
Use the class `MountPoint` as a DBus structure in the Storage module and UI.

Rename the class `MountPoint` to `MountPointRequest` and the DBus property
`MountPoints` to `Requests`, so there is no confusion between requested
and assigned mount points.


pyanaconda/modules/storage/partitioning/manual.py
```diff
@@ -87,25 +87,25 @@ def setup_kickstart(self, data):
 
         data_list = []
 
-        for mount_point in self.mount_points:
+        for request in self.requests:
             mount_data = data.MountData()
-            self._setup_mount_data(mount_data, mount_point)
+            self._setup_mount_data(mount_data, request)
```
jkonecny12: Is there a reason why the `mount_data` are not return value?

pyanaconda/ui/tui/spokes/storage.py
```diff
@@ -584,7 +584,7 @@ def _ensure_init_storage(self):
         # else
         print(_("Reverting previous configuration. This may take a moment..."))
         reset_storage(self.storage, scan_all=True)
-        self._manual_part_proxy.SetMountPoints([])
+        self._manual_part_proxy.SetRequests([])
```
jkonecny12: From this part I wouldn't understand what kind of storage requests are we talking about. Could you make this more obvious that this is mount point requests?

pyanaconda/modules/storage/partitioning/manual.py
poncovka: Because you need `data` to create `mount_data` and `request` to set up `mount_data`. I am also not very happy about the methods `_setup_mount_data` and `_process_mount_data`. I am thinking about pasting their code back to the methods `setup_kickstart` and `process_kickstart`. What do you think?

pyanaconda/ui/tui/spokes/storage.py
poncovka: Added a comment.

pyanaconda/modules/storage/partitioning/manual.py
jkonecny12: Maybe I'm missing something but why not change this to:
```
for request in self.requests:
    mount_data = self._setup_mount_data(data, request)

....
def _setup_mount_data(self, data, request):
    mount_data = data.MountData()
    <fill up mount data>
    return mount_data
```

The point is that you will not hide the side effect of the method. Before you didn't have a clue what it does.

pyanaconda/modules/storage/partitioning/manual.py
```diff
@@ -53,32 +53,32 @@ def publish(self):
...
 
-        mount_points = []
+        requests = []
 
         for obj in data.mount.mount_points:
-            mount_point = self.get_new_mount_point()
-            self._process_mount_data(obj, mount_point)
-            mount_points.append(mount_point)
+            request = MountPointRequest()
+            self._process_mount_data(obj, request)
```
jkonecny12: The same could be applied here.

pyanaconda/modules/storage/partitioning/manual.py
poncovka: I have considered it, but I don't think that there is a good reason for keeping these methods at this moment. They are removed in https://github.com/rhinstaller/anaconda/commit/166cdf51109d2a07d02a581116ce7b60b253882b.

PR #1965, opened 2019-05-13
Parse the output of df correctly (#1708701)

poncovka:
Split each line only once and in the rightmost way to get a mount
point and a number of available blocks. The mount point can contain
white spaces.

Resolves: rhbz#1708701

### DIFF
--- a/pyanaconda/payload/dnfpayload.py
+++ b/pyanaconda/payload/dnfpayload.py
@@ -103,9 +103,7 @@ def _df_map():
     lines = output.splitlines()
     structured = {}
     for line in lines:
-        items = line.split()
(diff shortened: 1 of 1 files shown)


poncovka:
@jkonecny12, I was able to reproduce the bug with a device mounted at `/tmp/My Volume`. Then the output looked for example like this:

```
Mounted on          Avail
/tmp/My Volume    1234567
```

jkonecny12:
Oh I see space in the mount name. Thanks for info.

Could you please add this example to the commit message. Thanks!

poncovka:
Updated the commit message.

PR #1967, opened 2019-05-13
Handle Firstboot kickstart command in DBus Task

M4rtinK:


pyanaconda/core/util.py
```diff
@@ -666,19 +666,33 @@ def service_running(service):
     return ret == 0
 
 
-def enable_service(service):
```
jkonecny12: Please fix typo `moth` in the first commit message.

pyanaconda/core/util.py
```diff
@@ -666,19 +666,33 @@ def service_running(service):
...
-    """ Enable a systemd service in the sysroot """
-    ret = _run_systemctl("enable", service, root=getSysroot())
+def enable_service(service, root=None):
+    """ Enable a systemd service in the sysroot.
+
+    :param str: name of the service to enable
+    :param: path to the sysroot or None to use default sysroot path
```
jkonecny12: Please remove double space before `  path...`

pyanaconda/modules/services/installation.py
```diff
@@ -0,0 +1,67 @@
...
+
+
+class ConfigureInitialSetupTask(Task):
+    """Installation task for Initial Setup configuration."""
+
+    def __init__(self, sysroot, setup_on_boot):
+        """Create a new root password configuration task.
+
+        :param str sysroot: a path to the root of the target system
+        :param int setup_on_boot: setup-on-boot mode for Initial Setup
```
jkonecny12: But what the int here means? What will happen if I pass `-100` or `100`?
Could you please improve the documentation here?

pyanaconda/modules/services/services.py
```diff
@@ -187,3 +188,19 @@ def set_setup_on_boot(self, value):
...
+        :param str sysroot: a path to the root of the installed system
+        :returns: list of object paths of installation tasks
+        """
+        tasks = [
+            ConfigureInitialSetupTask(sysroot=sysroot, setup_on_boot=self.setup_on_boot)
+        ]
+
+        paths = [
+            self.publish_task(SERVICES.namespace, task) for task in tasks
+        ]
```
jkonecny12: Do we really need `for` here for just one task?

pyanaconda/core/util.py
M4rtinK: Fixed! GitHub really should have a better mechanism for commit message enhancement suggestions.

pyanaconda/core/util.py
M4rtinK: Fixed up the function headers.

pyanaconda/modules/services/installation.py
M4rtinK: Indeed! Changed the type to "enum" and added a note that this is being defined by the SetupOnBoot enum. 

Doe we have something similar elsewhere ? Can definitely make it the same if we do.

pyanaconda/modules/services/services.py
M4rtinK: At the moment there is just a single task, but there should be more soon for the other kickstart commands handled by the services module - "skipx", "services", "xconfig"

M4rtinK:
Updated based on review feedback.

M4rtinK:
Added a bunch of kickstart tests checking if we configure Initial Setup correctly:
https://github.com/rhinstaller/kickstart-tests/pull/252

pyanaconda/modules/services/installation.py
jkonecny12: Not sure but we can change it later when we hit it.

pyanaconda/modules/services/services.py
jkonecny12: In that case it could be here fine.

pyanaconda/modules/services/installation.py
```diff
@@ -0,0 +1,70 @@
...
+log = get_module_logger(__name__)
+
+__all__ = ["ConfigureInitialSetupTask"]
+
+
+class ConfigureInitialSetupTask(Task):
+    """Installation task for Initial Setup configuration."""
+
+    def __init__(self, sysroot, setup_on_boot):
+        """Create a new root password configuration task.
```
poncovka: Please, fix the doc string.

pyanaconda/installation.py
```diff
@@ -95,7 +97,26 @@ def doConfiguration(storage, payload, ksdata):
...
     os_config.append(Task("Configure authselect", ksdata.authselect.execute))
     os_config.append(Task("Configure SELinux", ksdata.selinux.execute))
-    os_config.append(Task("Configure first boot tasks", ksdata.firstboot.execute))
+
+    services_proxy = SERVICES.get_proxy()
+    services_dbus_tasks = services_proxy.InstallWithTasks(util.getSysroot())
+    # add one Task instance per DBUS task
+    for dbus_task in services_dbus_tasks:
+        task_proxy = SERVICES.get_proxy(dbus_task)
+        os_config.append(Task(task_proxy.Name, sync_run_task, (task_proxy,)))
```
poncovka: There should be a method that will get and append all installation tasks of the given DBus service. This code is the same as for the Users module.

pyanaconda/installation.py
```diff
@@ -95,7 +97,26 @@ def doConfiguration(storage, payload, ksdata):
...
+
+    # If Initial Setup is disabled notify Screen Access Manager so that
+    # this information ends up correctly in user interaction config file.
+    # Otherwise "firstboot --disable" in kickstart would only disable Initial Setup
+    # but not other tools reading the config file, such as Gnome Initial Setup.
+    if services_proxy.SetupOnBoot == SETUP_ON_BOOT_DISABLED:
+        # we need a simple function that does the assignment we can put into a task
+        def mark_post_inst_tools_disabled():
+            screen_access.sam.post_install_tools_disabled = True
```
poncovka: I would like to see this code in `screen_access.py`. For example, in a new method `update_config_file`.

pyanaconda/installation.py
M4rtinK: Yeah, it looks like we will be using these task queues for now, so some tooling like this could be useful.

pyanaconda/installation.py
M4rtinK: So basically a hook that would be called from installation.py ? The hook would then check the SetupOnBootState and set post_install_tools disabled accordingly. Does that sound right ?

poncovka: Yes :-)

pyanaconda/modules/services/installation.py
M4rtinK: Indeed! A clear copy-paste error. I'll combine it in a PR with the adding the screen access hook-

PR #1968, opened 2019-05-14
Remove HMC from flags

jkonecny12:
Do not replace HMC with a configuration value. This feature is arch dependent (only S390) so I expect the configuration would not change ever.

### DIFF
--- a/pyanaconda/payload/__init__.py
+++ b/pyanaconda/payload/__init__.py
@@ -81,6 +81,10 @@ class Payload(metaclass=ABCMeta):
     def first_payload_reset(self):
         return self._first_payload_reset
 
+    @property
+    def is_hmc_enabled(self):
+        return self.data.method.method == "hmc"
+
     def setup(self, storage):
         """Do any payload-specific setup."""
         self.storage = storage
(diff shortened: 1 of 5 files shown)


pyanaconda/payload/__init__.py
```diff
@@ -81,6 +81,10 @@ def __init__(self, data):
     def first_payload_reset(self):
         return self._first_payload_reset
 
+    @property
+    def is_hmc_enabled(self):
+        return self.data.method.method == "hmc"
```
poncovka: Just a note, the point of the `hmc` flag was to enable the selection of SE/HMC during the installation, so you can change the installation source to something different and then set SE/HMC again. I think that it will still work this way, because the spokes remember the initial value of this property, but it could be easy to miss it next time. It is not really documented or anything, so as long as the basic use case works, I am fine with it. It is just something to consider.

jkonecny12: You mean that the SE/HMC can change but the value should still be set to True. In other words, it can't be disabled?

jkonecny12: Thanks for the explanation.

PR #1969, opened 2019-05-15
A couple Initial Setup related fixes

M4rtinK:
The first commit exposes the individual user group, user and root password tasks via the DBUS interface so that Initial Setup can run them on demand. Initial Setup does not expose SSH key configuration so we don't need that task to be available on demand via DBUS.

The second commit fixes a minor logic error that was preventing the user spoke from correctly showing up in Initial Setup. Also we still need to continue checking for the "data" variable in should_run as that's used by the User spoke to access the password policy data.

### DIFF
--- a/pyanaconda/modules/users/users.py
+++ b/pyanaconda/modules/users/users.py
@@ -135,25 +135,56 @@ class UsersModule(KickstartModule):
 
         return str(data)
 
+    def configure_groups_with_task(self, sysroot):
+        """Return the user group configuration task.
+
+        :param str sysroot: a path to the root of the installed system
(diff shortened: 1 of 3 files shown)


M4rtinK:
@poncovka Yep, this is a less than ideal solution aimed at unbreaking Rawhide ARM image deployments & other important Initial Setup use cases.

Proper solution indeed looks like adding some more introspection or at least some constants that the API consumers can use to find matching tasks from the module task listing ? Or maybe some sort of task registry (as part of Boss ?) where one could request tasks by task-id/type/tag ?

PR #1971, opened 2019-05-16
Firstboot Task followup fixes

M4rtinK:
The first commit fixes a typo, the second one makes user interaction config file updates nicer.

### DIFF
--- a/pyanaconda/screen_access.py
+++ b/pyanaconda/screen_access.py
@@ -42,6 +42,8 @@ from threading import RLock
 from pyanaconda import startup_utils
 from pyanaconda.core import util
 from pyanaconda.core.configuration.anaconda import conf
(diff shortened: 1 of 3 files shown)


pyanaconda/screen_access.py
```diff
@@ -236,6 +238,19 @@ def get_screen_option_changed(self, screen_name, option_name):
                           option_name, screen_name)
             return False
 
+    def synchronize_config_file_state(self):
```
poncovka: I don't really like the name `synchronize`, because the method just updates some values in the config file.

M4rtinK: I see it as synchronizing the config state with state of the DBus modules - but might as well change it to update. :)

PR #1973, opened 2019-05-19
Define the __eq__ method of DBusData

poncovka:
Compare the values of all data fields by default.

Use a dictionary from `generate_dictionary_from_data`
to generate a string representation of the data. Change the
argument `add` to a dictionary of attributes to make it more
user-friendly.

### DIFF
--- a/pyanaconda/dbus/structure.py
+++ b/pyanaconda/dbus/structure.py
@@ -182,6 +182,10 @@ class DBusData(ABC):
         """Convert this data object to a string."""
         return generate_string_from_data(self)
 
+    def __eq__(self, other):
+        """Compare data of the data objects."""
+        return compare_data(self, other)
+
 
 def get_fields(obj):
     """Return DBus fields of a data object.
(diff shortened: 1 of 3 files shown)


pyanaconda/dbus/structure.py
```diff
@@ -273,23 +273,21 @@ def generate_string_from_data(obj, skip=None, add=None):
...
-        attributes.append(attribute)
+    for name in add or dict():
+        dictionary[name] = add[name]
 
-    if extra_attributes:
-        attributes.extend(extra_attributes)
-        attributes.sort()
+    attributes = sorted([
+        "{}={}".format(name, repr(value))
+        for name, value in dictionary.items()
```
jkonecny12: Can you please merge this to one line?

pyanaconda/dbus/structure.py
```diff
@@ -273,23 +273,21 @@ def generate_string_from_data(obj, skip=None, add=None):
...
 
-    for field in get_fields(obj).values():
-        if field.data_name in skipped_names:
-            continue
+    for name in skip or list():
+        dictionary.pop(name, None)
 
-        attribute = "{}={}".format(field.data_name, repr(field.get_data(obj)))
-        attributes.append(attribute)
+    for name in add or dict():
```
jkonecny12: I don't like this construct but it works so if you want, you can leave it this way.

pyanaconda/dbus/structure.py

poncovka:
Updated.

PR #1976, opened 2019-05-21
rpmostree: use nss files from sysroot for tmpfiles

yuvalturg:
Executing systemd-tmpfiles with --root does not recognize users and
groups from the root filesystem.  To overcome this, copy the nss files
from the sysroot over to the stage2 filesystem just before
systemd-tmpfiles

Signed-off-by: Yuval Turgeman <yturgema@redhat.com>

### DIFF
--- a/pyanaconda/payload/rpmostreepayload.py
+++ b/pyanaconda/payload/rpmostreepayload.py
@@ -368,15 +368,19 @@ class RPMOSTreePayload(Payload):
         # both mounts like /home (really /var/home) and %post scripts might
         # want to write to e.g. `/srv`, `/root`, `/usr/local`, etc. The
         # /var/lib/rpm symlink is also critical for having e.g. `rpm -qa` work
-        # in %post. We don't iterate *all* tmpfiles because we don't have the
-        # matching NSS configuration inside Anaconda, and we can't "chroot" to
-        # get it because that would require mounting the API filesystems in the
-        # target.
-        for varsubdir in ('home', 'roothome', 'lib/rpm', 'opt', 'srv',
-                          'usrlocal', 'mnt', 'media', 'spool', 'spool/mail'):
-            self._safe_exec_with_redirect("systemd-tmpfiles",
-                                          ["--create", "--boot", "--root=" + util.getSysroot(),
-                                           "--prefix=/var/" + varsubdir])
+        # in %post.
+        nsspaths = ("/etc/", "/usr/lib/")
+        for nfile in ("passwd", "group"):
+            infiles = [util.getSysroot() + path + nfile for path in nsspaths]
+            if all([os.path.isfile(f) for f in infiles]):
+                log.debug("copying over %s from sysroot", nfile)
+                with open("/etc/{}".format(nfile), "w") as outf:
+                    for fname in infiles:
+                        with open(fname) as inf:
+                            outf.write(inf.read())
+        self._safe_exec_with_redirect("systemd-tmpfiles",
+                                      ["--create", "--boot", "--root=" + util.getSysroot(),
+                                       "--prefix=/var/"])
 
         # Handle mounts like /boot (except avoid /boot/efi; we just need the
         # toplevel), and any admin-specified points like /home (really


jkonecny12:
Jenkins, it's ok to test.

jkonecny12:
@cgwalters could you please look on this and say what do you think about it?

pyanaconda/payload/rpmostreepayload.py
```diff
@@ -368,15 +368,16 @@ def prepare_mount_targets(self, storage):
...
-                          'usrlocal', 'mnt', 'media', 'spool', 'spool/mail'):
-            self._safe_exec_with_redirect("systemd-tmpfiles",
-                                          ["--create", "--boot", "--root=" + util.getSysroot(),
-                                           "--prefix=/var/" + varsubdir])
+        # in %post.
+        log.debug("copying over NSS files from sysroot")
+        nssfiles = ["passwd", "group"]
+        srcfiles = [util.getSysroot() + "/usr/lib/" + f for f in nssfiles]
+        cp_args = ["-a", "-bS", ".orig"] + srcfiles + ["/etc"]
+        self._safe_exec_with_redirect("cp", cp_args)
```
cgwalters: First, please have this code silently do nothing if the files don't exist, so things continue to work after we [switch to sysusers](https://github.com/projectatomic/rpm-ostree/issues/49).

Second...I think you need to temporarily *concatenate* them, otherwise the `root` user won't be present, etc.

yuvalturg: Thanks for the review @cgwalters

pyanaconda/payload/rpmostreepayload.py
```diff
@@ -368,15 +368,19 @@ def prepare_mount_targets(self, storage):
...
-            self._safe_exec_with_redirect("systemd-tmpfiles",
-                                          ["--create", "--boot", "--root=" + util.getSysroot(),
-                                           "--prefix=/var/" + varsubdir])
+        # in %post.
+        nsspaths = ("/etc/", "/usr/lib/")
+        for nfile in ("passwd", "group"):
+            infiles = [util.getSysroot() + path + nfile for path in nsspaths]
+            if all([os.path.isfile(f) for f in infiles]):
+                log.debug("copying over %s from sysroot", nfile)
+                with open("/etc/%s" % nfile, "w") as outf:
```
jkonecny12: Please change this to `.format()` instead of `%`.

pyanaconda/payload/rpmostreepayload.py
```diff
@@ -368,15 +368,19 @@ def prepare_mount_targets(self, storage):
...
+                    for fname in infiles:
+                        with open(fname) as inf:
+                            outf.write(inf.read())
```
jkonecny12: Is there a reason to do this manually and not use the `copy` library?

jkonecny12:
Could you please explain how we can test this? Also is there a bug you are trying to fix by this contribution?

jkonecny12:
jenkins, test this please

yuvalturg:
Hi guys, thanks for the review, I didn't file a bug, but the issue that is reported in systemd/systemd#7032 seems to fit.

This patch will concatenate nss files from /etc and /usr/lib from the image and in case one of them is missing it will fail silently as @cgwalters asked so it won't break the move to sysusers.

As for testing, I created my own ostree image with custom file ownership under /var, repacked stage2 with this change, then created my own iso, and saw that the installation was successful with correct ownership under /var.

Why do we need to restore the old files ?

pyanaconda/payload/rpmostreepayload.py
yuvalturg: Done

poncovka:
@yuvalturg, Anaconda doesn't run only in the stage2 environment. It can be installed on your system and used for installation to a directory or an image. Also, the users and groups of the stage2 environment can be set up with a kickstart file to configure login via ssh. Therefore, it is not desirable to change the users and groups of the host system. Even with restored files, the installation can fail before that and leave the host system changed.

We recommend to file a bug to libc nss, so they can implement support for the root option and systemd-tmpfiles can use it.

I am closing this pull request.

PR #1977, opened 2019-05-22
Protect cdroms during tree population for image installs.

rvykydal:
Resolves: rhbz#1671713

Depends on https://github.com/storaged-project/blivet/pull/777

### DIFF
--- a/pyanaconda/storage/osinstall.py
+++ b/pyanaconda/storage/osinstall.py
@@ -82,6 +82,7 @@ def enable_installer_mode():
     # into the *host's* /etc/lvm. This can get real messy on build systems.
     if flags.imageInstall:
         blivet_flags.lvm_metadata_backup = False
+        blivet_flags.protect_cdroms = True
 
     blivet_flags.auto_dev_updates = True
     blivet_flags.selinux_reset_fcon = True
--- a/anaconda.spec.in
+++ b/anaconda.spec.in
@@ -85,7 +85,7 @@ The anaconda package is a metapackage for the Anaconda installer.
 Summary: Core of the Anaconda installer
 Requires: python3-libs
 Requires: python3-dnf >= %{dnfver}
-Requires: python3-blivet >= 1:3.1.0-1
+Requires: python3-blivet >= 1:3.1.0-12
 Requires: python3-blockdev >= %{libblockdevver}
 Requires: python3-meh >= %{mehver}
 Requires: libreport-anaconda >= 2.0.21-1
--- a/pyanaconda/payload/__init__.py
+++ b/pyanaconda/payload/__init__.py
@@ -1167,7 +1167,7 @@ class PackagePayload(Payload):
 
         # Check to see if the device is already mounted, in which case
         # we don't need to mount it again so skip this part
-        elif not (method.method == "cdrom" and blivet.util.get_mount_paths(device.path)):
+        elif not (method.method == "cdrom" and INSTALL_TREE in blivet.util.get_mount_paths(device.path)):
             device.format.setup(mountpoint=INSTALL_TREE)
 
         return "file://" + INSTALL_TREE


pyanaconda/storage/osinstall.py
```diff
@@ -82,6 +82,7 @@ def enable_installer_mode():
     # into the *host's* /etc/lvm. This can get real messy on build systems.
     if blivet_flags.image_install:
         blivet_flags.lvm_metadata_backup = False
+        blivet_flags.protect_cdroms = True
```
poncovka: I think that the code for protecting cdroms should be in anaconda. It could be probably done the same way as protecting zRAM swap devices: https://github.com/rhinstaller/anaconda/blob/rhel-devel/pyanaconda/anaconda.py#L177

rvykydal: That wouldn't protect the device from teardown_all() that is called by

storage_initialize -> InstallerStorage.reset() -> Blivet.reset() -> DeviceTree.populate() which is calling DeviceTree.teardown_all().

The specs from anaconda.protected are applied after reset/populate (_mark_protected_device(dev)).

I didn't find a way how to prevent teardown of (just populated) cdrom in Blivet.reset() outside blivet.

poncovka: In that case, I would like to be able to propagate the list of protected devices to Blivet, so they can be skipped during the teardown, or the teardown shouldn't be part of the populate method and we can call it in Anaconda in our reset method. It seems weird to me that it is possible to teardown devices that should be protected. @vojtechtrefny, what do you think?

dwlehman:
Does this mean that all cdrom/dvd installations are broken in this way?

rvykydal:
@dwlehman  For now the only problem we are aware of is the referred BZ  - https://bugzilla.redhat.com/show_bug.cgi?id=1671713 - during image install on a system, DVD used by other tools is unmounted.

Other option could be to set auto_dev_update blivet flag to False and do the teardown on our own in installer. I am not quite sure what is the purpose of the flag and why the devices are torn down after populating (in Blivet.reset) though.

pyanaconda/storage/osinstall.py
jkonecny12: @rvykydal did you tried if the CDROM installation works with this code. I fear that this could block us from installing from the media. Ideally also test if you are able to install from a different cdrom than you booted.

poncovka:
The flag `auto_dev_updates` was introduced in https://github.com/storaged-project/blivet/commit/d15a7dd3aeb74fb1a7f08092975a4fb04d45fc6f. It seems to be installer-specific, so it might be possible to do some changes.

dwlehman:
> @dwlehman For now the only problem we are aware of is the referred BZ - https://bugzilla.redhat.com/show_bug.cgi?id=1671713 - during image install on a system, DVD used by other tools is unmounted.
> 
> Other option could be to set auto_dev_update blivet flag to False and do the teardown on our own in installer. I am not quite sure what is the purpose of the flag and why the devices are torn down after populating (in Blivet.reset) though.

We replaced the old `installer_mode` flag with a few more-specific flags. One of the behaviors specific to the installer was freely activating and deactivating devices, automatically running commands to get information about resize for file systems, etc. That's what `auto_dev_update` controls.

The reason we tear down the devices right away is so we know if there is something that will prevent us from doing so early -- rather than late -- in the installation process.

pyanaconda/storage/osinstall.py
rvykydal: @jkonecny12 The patch is limited to image installs. But I'll check if it does not break installing from cdrom.

rvykydal: Actually we shouldn't differentiate by target (image) but system (system) in this case though - but that would be  for upstream. I think for rhel branch we just should go with flags.imageInstall.

rvykydal:
Protecting the DVD from unmounting is causing one issue for DVD installs in environments where the installation DVD is already mounted. It should be fixed by the additional patch https://github.com/rhinstaller/anaconda/pull/1977/commits/cb9953810e43e27fde68d13a864787ddf5aa624a . Here is the table with testing results without ("patched") and with ("patched+") the additional patch.

1) installation DVD mounted

ks with http repo specified - the BZ case, DVD should be used by other tools
[-] original: umounts the DVD
[+] patched: DVD stays mounted
[+] patched+: DVD stays mounted

ks with cdrom command
[-] original - DVD is unmounted (then mounted to /run/install/source)
[-] patched - DVD source does not work, stays mounted
[+] patched+ - DVD stays mounted (also mounted to /run/install/source)

interactive using the DVD by autodetection
[-] original - DVD is unmounted (mounted to /run/install/source)
[-] patched - DVD source does not work, stays mounted
[+] patched+ - DVD stays mounted (also mounted to /run/install/source)

2) installation DVD not mounted

ks with http repo specified
[-] original: umounts the DVD
[+] patched: DVD stays mounted
[+] patched+: DVD stays mounted

ks with cdrom set
[+] original - mounted to /run/install/source
[+] patched - mounted to /run/install/source
[+] patched+ - mounted to /run/install/source

interactive using the DVD by autodetection
[+] original - mounted to /run/install/source)
[+] patched - mounted to /run/install/source
[+] patched+ - mounted to /run/install/source

3) non-installation DVD mounted

ks with http repo specified
[-] original: umounts the DVD
[+] patched: DVD stays mounted
[+] patched+: DVD stays mounted

jkonecny12:
If I'm reading this correctly it solves even more errors than the original issue. Great work and thanks for testing!

rvykydal:
I've run the tests from comment above with updated patch for blivet https://github.com/storaged-project/blivet/pull/777#issuecomment-500399834 and all the results are same as patched+ above.

rvykydal:
For the reference: better (and less isolated) solution will be based on https://github.com/storaged-project/blivet/pull/780.

rvykydal:
Ported to master by https://github.com/rhinstaller/anaconda/pull/2275

PR #1980, opened 2019-05-23
Add DBus support for the rescue mode

poncovka:
Call the DBus method GetExistingSystems of the device tree module to
get a list of existing installations on devices from the device tree.

Call the DBus method FindExistingSystemsWithTask to update this list.

Call the DBus method MountExistingSystemWithTask of the device tree
module to mount an existing system.


rvykydal:
"operational system" -> "operating system" ?

poncovka:
Fixed.

PR #1981, opened 2019-05-27
Hide spokes with the Anaconda configuration file

poncovka:
Run a new installation task of the Services module to configure
the post-installation tools. It will enable or disable post-install
tools via the user interaction configuration file.

Add a new option hidden_spokes in the UI section of the Anaconda
configuration file and use it to hide spokes in UI. Don't use the
user interaction file for that.

Remove the support for the screen access management. It is not
necessary to generate data about visited spokes anymore and the
spokes can be hidden with the Anaconda configuration file.

Hide some of the spokes in variants Silverblue and Workstation Live.


poncovka:
**TODO:**
* Add tests for the installation task.
* Figure out how to handle the Workstation variant - some spokes are hidden only for live installations.

poncovka:
Rebased and added tests.

PR #1983, opened 2019-05-29
Add DBus support for teardown tasks

poncovka:
We set up the disk images with `setup_disk_images` of the device tree
instance, so we should call `teardown_disk_images` to tear them down.

Call the DBus method `TeardownWithTasks` of the Storage module before
the module shutdown.

### DIFF
--- a/anaconda.py
+++ b/anaconda.py
@@ -80,15 +80,13 @@ def exitHandler(rebootData, storage):
     if anaconda.payload:
         anaconda.payload.unsetup()
 
+    # Unmount the filesystems.
     if not conf.target.is_hardware:
         anaconda.storage.umount_filesystems(swapoff=False)
-        devicetree = anaconda.storage.devicetree
-        devicetree.teardown_all()
-        for imageName in devicetree.disk_images:
-            dev = devicetree.get_device_by_name(imageName)
-            for loop in dev.parents:
-                loop.controllable = True
-            dev.deactivate(recursive=True)
+
+    # Tear down disk images.
+    if conf.target.is_image:
+        anaconda.storage.devicetree.teardown_disk_images()
 
     # Clean up the PID file
     if pidfile:
--- a/pyanaconda/modules/common/base/base_interface.py
+++ b/pyanaconda/modules/common/base/base_interface.py
@@ -127,9 +127,16 @@ class KickstartModuleInterface(KickstartModuleInterfaceTemplate):
         """Returns installation tasks of this module.
 
         :param sysroot: a path to the root of the installed system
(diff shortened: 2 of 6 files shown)


pyanaconda/modules/common/base/base_interface.py
```diff
@@ -123,6 +123,13 @@ def InstallWithTasks(self, sysroot: Str) -> List[ObjPath]:
         """
         return self.implementation.install_with_tasks(sysroot)
 
+    def TeardownWithTasks(self) -> List[ObjPath]:
+        """Returns teardown tasks for this module.
+
+        :returns: list of object paths of installation tasks.
```
jkonecny12: Remove the dot here.

anaconda.py
```diff
@@ -80,15 +80,13 @@ def exitHandler(rebootData, storage):
     if anaconda.payload:
         anaconda.payload.unsetup()
 
+    # Unmount the filesystems.
     if not conf.target.is_hardware:
         anaconda.storage.umount_filesystems(swapoff=False)
-        devicetree = anaconda.storage.devicetree
-        devicetree.teardown_all()
```
jkonecny12: Are you sure this will work correctly? It looks like a behavior change.

pyanaconda/modules/common/base/base_interface.py
poncovka: Fixed.

anaconda.py
poncovka: The method `teardown_all` is called in `teardown_disk_images`, so it shouldn't change image installations. And the device tree is not populated for dir installations, so it shouldn't change them either.

anaconda.py
```diff
@@ -80,15 +80,13 @@ def exitHandler(rebootData, storage):
...
-        for imageName in devicetree.disk_images:
-            dev = devicetree.get_device_by_name(imageName)
-            for loop in dev.parents:
-                loop.controllable = True
-            dev.deactivate(recursive=True)
+
+    # Tear down disk images.
+    if conf.target.is_image:
+        anaconda.storage.devicetree.teardown_disk_images()
```
poncovka: @vojtechtrefny, could take a quick look? Is it correct to call `teardown_disk_images` instead of the removed part?

vojtechtrefny: Yes, `teardown_disk_images` does basically the same. The `loop.controllable = True` looks odd, but it shouldn't be needed.

anaconda.py
jkonecny12: OK, thanks for explanation.

PR #1985, opened 2019-05-30
Fix non-root dir of install tree HD installation (#1689194)

jkonecny12:
Installation didn't worked if you placed an install tree on the other place than / of the partition. This was because `method.dir` was not correctly propagated in this case.

The reason is, before installation from an install tree was supported only the ISO installation was supported from HD. In that case you would use `method.dir` to find the ISO but the path to the installation source would be always root of the mounted image.
This does not work for install tree because we will mount partition and then have to use the path to adjust installation source on that mount point.

Solution is to know what type of installation we have and adjust the path based on that.

Resolves: rhbz#1689194

### DIFF
--- a/pyanaconda/payload/__init__.py
+++ b/pyanaconda/payload/__init__.py
@@ -1151,6 +1151,7 @@ class PackagePayload(Payload):
 
     def _setupMedia(self, device):
         method = self.data.method
+
         if method.method == "harddrive":
             try:
                 method.dir = self._find_and_mount_iso(device, ISO_DIR, method.dir, INSTALL_TREE)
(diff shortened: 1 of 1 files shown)


pyanaconda/payload/__init__.py
```diff
@@ -1272,8 +1275,11 @@ def _setup_harddrive_device(self, storage, method, isodev, device):
             if not isodevice:
                 raise PayloadSetupError("device for HDISO install %s does not exist" % devspec)
 
-            self._setupMedia(isodevice)
-            url = "file://" + INSTALL_TREE
+            if self._setupMedia(isodevice):
+                url = "file://" + INSTALL_TREE + method.dir
+            else:
+                url = "file://" + INSTALL_TREE
```
poncovka: Couldn't you return the url instead of the boolean value? This solution looks like a hack to me.

jkonecny12:
**UPDATED:**
* Thanks for the suggestion Vendy.

PR #1986, opened 2019-05-30
Initial Setup and post install tool configuration improvements and fixes

M4rtinK:
This pull request contains various fixes and improvements for Initial Setup & post install tool configuration in general.

The first commit adds a separate property that controls if all post installation tools (that read the config file we create) will be disabled or not. This makes it explicit if what will happen and makes it possible to control this option not only via kickstart, but also over DBus. The next commit makes use of this property when updating the user interaction config file, which hosts the actual `post_install_tools_disabled` key.

The third commit switches the default for Initial Setup to be disabled by default, dropping a bogus `firstboot --disable` from the output kickstart of every kickstart installation that has no `firstboot` command in its input kickstart.

And the last commit forwards the sysroot path to service enable/disable method invocations, that we apparently forgot to apply a while ago.

### DIFF
--- a/anaconda.py
+++ b/anaconda.py
@@ -739,14 +739,14 @@ if __name__ == "__main__":
 
     # Finish the initialization of the setup on boot action.
     # This should be done sooner and somewhere else once it is possible.
-    from pyanaconda.core.constants import SETUP_ON_BOOT_DEFAULT, SETUP_ON_BOOT_DISABLED
+    from pyanaconda.core.constants import SETUP_ON_BOOT_DEFAULT, SETUP_ON_BOOT_ENABLED
     from pyanaconda.modules.common.constants.services import SERVICES
     services_proxy = SERVICES.get_proxy()
 
     if services_proxy.SetupOnBoot == SETUP_ON_BOOT_DEFAULT:
-        if flags.automatedInstall:
-            # Disable by default after kickstart installations.
-            services_proxy.SetSetupOnBoot(SETUP_ON_BOOT_DISABLED)
+        if  not flags.automatedInstall:
+            # Enable by default for interactive installations.
+            services_proxy.SetSetupOnBoot(SETUP_ON_BOOT_ENABLED)
 
     # Create pre-install snapshots
     from pykickstart.constants import SNAPSHOT_WHEN_PRE_INSTALL
--- a/pyanaconda/modules/services/installation.py
+++ b/pyanaconda/modules/services/installation.py
@@ -31,6 +31,8 @@ __all__ = ["ConfigureInitialSetupTask"]
 class ConfigureInitialSetupTask(Task):
     """Installation task for Initial Setup configuration."""
 
+    INITIAL_SETUP_UNIT_NAME = "initial-setup.service"
+
     def __init__(self, sysroot, setup_on_boot):
         """Create a new Initial Setup configuration task.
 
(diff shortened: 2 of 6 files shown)


pyanaconda/modules/services/services.py
```diff
@@ -80,6 +83,15 @@ def process_kickstart(self, data):
         setup_on_boot = self._map_firstboot(data.firstboot.firstboot)
         self.set_setup_on_boot(setup_on_boot)
 
+        # Having "firstboot --disable" in the input kickstart
+        # sets post_install_tools_disable property,
+        # which is then reflected in the user interaction config file
+        # parsed by Gnome Initial Setup.
+        # This makes it possible to disable both with a single
+        # kickstart command.
```
poncovka: Please, make a nice text block from this.

pyanaconda/modules/services/services.py
```diff
@@ -189,6 +201,30 @@ def set_setup_on_boot(self, value):
         self.setup_on_boot_changed.emit()
         log.debug("Setup on boot is set to %s.", value)
 
+    @property
+    def post_install_tools_disabled(self):
```
poncovka: Please, change it to `post_install_tools_enabled`. It is difficult to work with the negations.

M4rtinK: The original idea was to mirror the `post_install_tools_disabled` key in the user interaction config file, but this really looks weird as part of the DBus module API. 

I guess adding a note that the key will be written out only if `post_install_tools_enabled` == `False` should be able to achieve the same result.

pyanaconda/modules/services/installation.py
```diff
@@ -50,21 +50,28 @@ def name(self):
...
+                log.debug("Initial Setup will be enabled.")
+                util.enable_service(unit_name, root=self._sysroot)
+                return
 
-        log.debug("Initial Setup will be enabled.")
-        util.enable_service(unit_name)
+        # the Initial Setup service is disabled by default
+        if os.path.exists(os.path.join(self._sysroot, "lib/systemd/system/", unit_name)):
+            log.debug("Initial Setup will be disabled.")
+            util.disable_service(unit_name, root=self._sysroot)
```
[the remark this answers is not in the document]
M4rtinK: Indeed, that looks much nicer! :) Let's do it that way.

pyanaconda/modules/services/services.py
M4rtinK: The block has been improved.

M4rtinK:
The PR has been updated:
- applied feedback from @poncovka (Thanks! :) )
- reduced number of commits as it made a logical sense
- added a commit that adds a utility function for checking in a unit file exists

pyanaconda/core/util.py
```diff
@@ -697,6 +697,19 @@ def disable_service(service, root=None):
     if ret != 0:
         log.warning("Disabling %s failed. It probably doesn't exist", service)
 
+def unit_file_exists(service, root=None):
```
poncovka: I don't think that this function is general enough to be useful for other cases.

pyanaconda/modules/services/installation.py
```diff
@@ -48,23 +50,32 @@ def __init__(self, sysroot, setup_on_boot):
...
+        """Disable the Initial Setup service."""
+        if util.unit_file_exists(self.INITIAL_SETUP_UNIT_NAME, root=self._sysroot):
+            util.disable_service(self.INITIAL_SETUP_UNIT_NAME, root=self._sysroot)
 
-        log.debug("Initial Setup will be enabled.")
-        util.enable_service(unit_name)
+    def _enable_reconfig_mode(self):
+        """Write the reconfig mode trigger file."""
+        log.debug("Initial Setup reconfiguration mode will be enabled.")
+        f = open(os.path.join(self._sysroot, "etc/reconfigSys"), "w+")
```
poncovka: Please, use `with`.

anaconda.py
```diff
@@ -739,14 +739,14 @@ def _earlyExceptionHandler(ty, value, traceback):
...
     from pyanaconda.modules.common.constants.services import SERVICES
     services_proxy = SERVICES.get_proxy()
 
     if services_proxy.SetupOnBoot == SETUP_ON_BOOT_DEFAULT:
-        if flags.automatedInstall:
-            # Disable by default after kickstart installations.
-            services_proxy.SetSetupOnBoot(SETUP_ON_BOOT_DISABLED)
+        if  not flags.automatedInstall:
+            # Enable by default for interactive installations.
+            services_proxy.SetSetupOnBoot(SETUP_ON_BOOT_ENABLED)
```
poncovka: Shouldn't we set also `PostInstallToolsEnabled`? Couldn't we remove `firstboot --enable` from `interactive-defaults.ks` now?

pyanaconda/core/util.py
M4rtinK: I was making the assumption we have other cases where we need to check if a systemd unit exists, but it does not seem to be the case. So it definitely makes sense to have this as an internal method.

pyanaconda/modules/services/installation.py
```diff
@@ -48,23 +50,32 @@ def __init__(self, sysroot, setup_on_boot):
...
-            return
-
-        if not os.path.exists(os.path.join(self._sysroot, "lib/systemd/system/", unit_name)):
+    def _enable_service(self):
+        """Enable the Initial Setup service."""
+        if util.unit_file_exists(self.INITIAL_SETUP_UNIT_NAME, root=self._sysroot):
+            util.enable_service(self.INITIAL_SETUP_UNIT_NAME, root=self._sysroot)
+        else:
             log.debug("Initial Setup will not be started on first boot, because "
-                      "it's unit file (%s) is not installed.", unit_name)
```
rvykydal: it's -> its

pyanaconda/modules/services/services.py
```diff
@@ -189,6 +203,36 @@ def set_setup_on_boot(self, value):
         self.setup_on_boot_changed.emit()
         log.debug("Setup on boot is set to %s.", value)
 
+    @property
+    def post_install_tools_enabled(self):
+        """"Should all post installation tools be disabled ?
```
rvykydal: """" -> """

pyanaconda/modules/services/installation.py
M4rtinK: Even better, turns out we have a utility function for empty file creation that can be used.

anaconda.py
M4rtinK: `PostInstallToolsEnabled` is True by default (that's one of the reason we wanted to split if from the `SetupOnBoot` property), so we don't need to set it to True here. 

And I believe we can indeed drop `firstboot --enable` command from `interactive-defaults.ks` once this PR is merged - but preferably after some manual testing of a custom built livecd (we have docs for that).

pyanaconda/modules/services/services.py
M4rtinK: Fixed. :)

pyanaconda/modules/services/installation.py
M4rtinK: Fixed. :)

PR #1988, opened 2019-05-31
Replace Firewall execute() method by DBus Task

M4rtinK:
The first commit extends the execInSysroot utility function to support setting a custom sysroot path.
Second commit adds a new DBus Task for Firewall configuration and the third commit replaces the execute() method by the Task.

BTW, it looks like the built-in InstallWithTasks interface is not available for sub-modules and one also has to get a task proxy for the tasks from the top level module (NETWORK in this case) to make things work, which is kinda confusing. Looks like something we might want to improve in the future.


pyanaconda/core/util.py
```diff
@@ -334,15 +334,16 @@ def _run_program(argv, root='/', stdin=None, stdout=None, env_prune=None, log_ou
     return (proc.returncode, output_string)
 
 
-def execInSysroot(command, argv, stdin=None):
+def execInSysroot(command, argv, stdin=None, root=getSysroot()):
```
poncovka: The sysroot can change during the installation. You should set it inside the function.

pyanaconda/modules/network/firewall/firewall.py
```diff
@@ -203,3 +204,18 @@ def set_disabled_services(self, disabled_services):
...
         self.disabled_services_changed.emit()
         log.debug("Services that will be explicitly disabled on the firewall: %s", self._disabled_services)
+
+    def install_with_task(self, sysroot):
+        """Return the installation tasks of this module.
+
+        :param str sysroot: a path to the root of the installed system
+        :returns: list of object paths of installation tasks
+        """
+        firewall_configuration_task = ConfigureFirewallTask(sysroot=sysroot,
```
poncovka: Please, start the list of arguments on a new line and fix the indentation.

pyanaconda/modules/network/firewall/firewall_interface.py
```diff
@@ -114,3 +114,13 @@ def SetDisabledServices(self, disabled_services: List[Str]):
         :param disabled_services: a list of services to be enabled
         """
         self.implementation.set_disabled_services(disabled_services)
+
+    def InstallWithTask(self, sysroot: Str) -> ObjPath:
```
poncovka: Please, use the method `InstallWithTasks` that is defined for all modules.

pyanaconda/modules/network/firewall/firewall.py
```diff
@@ -203,3 +204,18 @@ def set_disabled_services(self, disabled_services):
         self._disabled_services = list(disabled_services)
         self.disabled_services_changed.emit()
         log.debug("Services that will be explicitly disabled on the firewall: %s", self._disabled_services)
+
+    def install_with_task(self, sysroot):
```
poncovka: Please, use the predefined method `install_with_tasks`.

pyanaconda/modules/network/firewall/firewall_interface.py
poncovka: I didn't realize that this is a submodule of the Network module. It can stay like this.

pyanaconda/modules/network/firewall/firewall.py
poncovka: I didn't realize that this is a submodule of the Network module. It can stay like this.

pyanaconda/core/util.py
jkonecny12: Yes, ideally just use `root=None` in the header and if `None` set the sysroot in the function body.

M4rtinK: Fixed. :)

pyanaconda/modules/network/firewall/firewall.py
M4rtinK: Should be fixed now.

poncovka:
It would be nice to have a test for the installation task. Please, create an issue for that.

M4rtinK:
Indeed - issue created for Firewall configuration task testing. BTW, we already have one kickstart test testing the `firewall` command.

PR #1990, opened 2019-06-04
Migrate payload requirements

jkonecny12:
**WIP - do not merge**
This is not working version of the PR so please bear that in mind. I will update it when fixing these issues. The PR is created mainly to show the structure of the module and start a discussion about possible changes.

* Migrate ``payload/requirement.py`` new ``Payload/Requirements`` module. 
* Add new Requirement DBus structure.

This module will collect groups and packages requirements during the installation and then they will be used by the payload module in the future changes. Now it will be used instead of the old solution.

There are a few differences between the old solution and the new one:
- `Strong` value is only one value for the whole Requirement structure.
  Reasoning for this is that this was the only use case. The `strong` property looked on all the reasons and if any was `True` it returned `True`. This is a code simplification so we don't have to store named tuple and the functionality stays the same.  
- The set_callback and call it later was removed. 
  There is not much a reason and it is not trivial to migrate that to DBus structure. The only caller and also who sets the callback is the dnf payload so why complicate things. New planning solution is that the payload will have a responsibility to collect all the requirements without itself (which is doing now so or so).

### DIFF
--- a/pyanaconda/modules/common/constants/objects.py
+++ b/pyanaconda/modules/common/constants/objects.py
@@ -109,3 +109,8 @@ DNF_PACKAGES = DBusObjectIdentifier(
     namespace=DNF_NAMESPACE,
     basename="Packages"
 )
+
+REQUIREMENTS = DBusObjectIdentifier(
+    namespace=PAYLOAD_NAMESPACE,
+    basename="Requirements"
+)
--- a/pyanaconda/modules/common/structures/payload.py
+++ b/pyanaconda/modules/common/structures/payload.py
@@ -1,0 +1,88 @@
+#
+# DBus structures for the payload data.
+#
+# Copyright (C) 2018 Red Hat, Inc.
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# the GNU General Public License v.2, or (at your option) any later version.
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY expressed or implied, including the implied warranties of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General
+# Public License for more details.  You should have received a copy of the
+# GNU General Public License along with this program; if not, write to the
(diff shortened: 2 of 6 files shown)


poncovka:
Hi, we talked about it with @M4rtinK and at this point we just need a way how to get the requirements from the DBus modules, so we can set up `payload.requirements` directly in `pyanaconda.installation.py`. This is implemented at https://github.com/rhinstaller/anaconda/pull/1995.

Eventually, the Payload module should collect the requirements from other modules in an installation task and set up its internal data, so the requirements are applied during the package installation if possible.

From this point of view, I am not sure what should be the role of the Requirements module anymore.

@rvykydal, what do you think?

rvykydal:
It seems to me like there are 2 approaches in play:

1) (the comment above) payload module is responsible to gather the info via common module API
vs
2) (this PR) module is responsible for using payload (requirements) API to set the info


I think we could use 1) with eventual option to do 2) via simple payload module api if really needed (maybe if also something other than module wants to add requirement)

Given the way we use requirements now I think 1) makes sense for now.

Anyway, we'll discuss it in today's meeting

(There is also a question of synchronization (ie is the module ready to provide the info when payload asks for it) but it would be probably something that would need to be handled with both approaches and we don't have to think about it now.)

jkonecny12:
> It seems to me like there are 2 approaches in play:
> 
> 1. (the comment above) payload module is responsible to gather the info via common module API
>    vs
> 2. (this PR) module is responsible for using payload (requirements) API to set the info
> 
> I think we could use 1) with eventual option to do 2) via simple payload module api if really needed (maybe if also something other than module wants to add requirement)
> 
> Given the way we use requirements now I think 1) makes sense for now.
> 
> Anyway, we'll discuss it in today's meeting
> 
> (There is also a question of synchronization (ie is the module ready to provide the info when payload asks for it) but it would be probably something that would need to be handled with both approaches and we don't have to think about it now.)

I don't think that this PR needs any synchronization is required. The reason for that is, that the requirements are collected during the installation phase and you should not change requirements there, even if you wanted to do that, then you should add task before payload installation so the requirements will be there.

jkonecny12:
This PR lost a fight with PR https://github.com/rhinstaller/anaconda/pull/1995. Closing now :)

PR #1993, opened 2019-06-05
Use new ssl certificate kickstart options

rvykydal:
The `repo` and `url` commands can now contain --sslcacert,
--sslclientcert, and --sslclientkey options. Use those when accessing
repositories.

Resolves: rhbz#1663954


rvykydal:
jenkins, test this please

PR #1997, opened 2019-06-06
Replace the lang execute() method

M4rtinK:
We actually already have a DBus task for this, but it's called in a
rather antiquated manner. So let's switch it to that standardized
install_with_tasks() interface and run the task from installation.py
so that we can drop the execute() method.

### DIFF
--- a/tests/nosetests/pyanaconda_tests/module_localization_test.py
+++ b/tests/nosetests/pyanaconda_tests/module_localization_test.py
@@ -132,7 +132,7 @@ class LocalizationInterfaceTestCase(unittest.TestCase):
     def install_language_with_task_test(self, publisher):
         """Test InstallLanguageWithTask."""
         self.localization_interface.SetLanguage("cs_CZ.UTF-8")
-        task_path = self.localization_interface.InstallLanguageWithTask("/")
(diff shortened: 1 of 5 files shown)


tests/nosetests/pyanaconda_tests/module_localization_test.py
```diff
@@ -128,21 +128,6 @@ def set_language_kickstarted_test(self):
         self.assertEqual(self.localization_interface.LanguageKickstarted, True)
         self.callback.assert_called_once_with(LOCALIZATION.interface_name, {'LanguageKickstarted': True}, [])
 
-    @patch('pyanaconda.dbus.DBus.publish_object')
-    def install_language_with_task_test(self, publisher):
-        """Test InstallLanguageWithTask."""
-        self.localization_interface.SetLanguage("cs_CZ.UTF-8")
-        task_path = self.localization_interface.InstallLanguageWithTask("/")
```
poncovka: Please, just modify the test for `InstallWithTasks`.

M4rtinK: Oh, good point, that's really simple - added. :)

PR #2001, opened 2019-06-07
Initialize DNF plugins (#1717016)

M4rtinK:
For the upcoming support for swid tag handling during the
installation RPM transaction the swidtags DNF plugin needs to run.

As Anaconda uses DNF via its Python API, DNF plugins are not
loaded by default unless init_plugins() is called on the DNF
Base class instance.

So call init_plugins() at appropriate time so that DNF
plugins, including the swidtags plugin, are properly
loaded.

Resolves: rhbz#1717016

NOTE - PROTOTYPE, NOT MERGEABLE YET

there are a couple things that need to be done/checked before this PR can be merged:
- the DNF debug level override should be dropped
- `init_plugins()` currently enables **all** plugins DNF can find, regardless if the plugins are useful or even suitable for the installation environment, we should proably either filter what plugins are enabled or at least check all the plugins not cause detectable issues

### DIFF
--- a/pyanaconda/payload/dnfpayload.py
+++ b/pyanaconda/payload/dnfpayload.py
@@ -666,6 +666,9 @@ class DNFPayload(payload.PackagePayload):
         conf.logdir = '/tmp/'
         # enable depsolver debugging if in debug mode
         self._base.conf.debug_solver = flags.debug
+        # REMOVE BEFORE MERGE
+        # - temporarily increse DNF debug level
+        self._base.conf.debuglevel = 10
         # set the platform id based on the /os/release
         # present in the installation environment
         platform_id = self.get_platform_id()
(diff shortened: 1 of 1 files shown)


poncovka:
Looks good to me. The list of enabled/disabled plugins could be in the Anaconda configuration file on Fedora, so it could be easy to modify if necessary.

pyanaconda/payload/dnfpayload.py
```diff
@@ -666,6 +666,9 @@ def _configure(self):
         conf.logdir = '/tmp/'
         # enable depsolver debugging if in debug mode
         self._base.conf.debug_solver = flags.debug
+        # REMOVE BEFORE MERGE
+        # - temporarily increse DNF debug level
```
jkonecny12: `increse` -> `increase`

M4rtinK:
Looks like this has stalled outside of Anaconda, so let's close this PR.

M4rtinK:
Since this was never merged to RHEL8, we can also drop the `port to Fedora` label.

PR #1978, opened 2019-05-22
Rename add-ons to additional software (#1674011)

jkonecny12:
Rename TUI and GUI add-ons in the software spoke to the 'Additional software'.

Resolves: rhbz#1674011

PR #2003, opened 2019-06-11
Use rhel-8.0 zanata version.

rvykydal:
rhel-8 is going to diverge from rhel-8.0 regarding the translations

Related: rhbz#1666319


rvykydal:
This is the PR introducing the divergence: https://github.com/rhinstaller/anaconda/pull/1978

PR #2004, opened 2019-06-11
Replace the XConfig execute() method by DBus Tasks

M4rtinK:
Replace the execute() method by the ConfigureSystemdDefaultTargetTask
and ConfigureDefaultDesktopTask.

Also drop the ancient desktop.py module and extend the task tests.


pyanaconda/modules/services/installation.py
```diff
@@ -123,3 +126,77 @@ def run(self):
...
+            msg = "Unsupported systemd default target: {} Specify one of: {}".format(
+                self._default_target, self.SUPPORTED_SYSTEMD_TARGETS
+            )
+            raise SystemdDefaultTargetConfigurationError(msg)
+        else:
+            log.debug("Setting systemd default target to: %s", self._default_target)
+
+        if not os.path.isdir(os.path.join(self._sysroot, 'etc/systemd/system')):
+            msg = "There is no /etc/systemd/system directory, cannot set systemd default target!"
+            raise SystemdDefaultTargetConfigurationError(msg)
```
poncovka: How likely is this case? Also, you are changing behavior by raising the exception.

pyanaconda/modules/services/installation.py
```diff
@@ -123,3 +126,77 @@ def run(self):
...
+
+        default_target_path = os.path.join(self._sysroot, 'etc/systemd/system/default.target')
+        # unlike an links already in place
```
poncovka: I don't understand the comment.

pyanaconda/modules/services/services.py
```diff
@@ -247,7 +247,8 @@ def install_with_tasks(self, sysroot):
                 disabled_services=self.disabled_services,
                 enabled_services=self.enabled_services
             ),
-
+            ConfigureSystemdDefaultTargetTask(sysroot=sysroot, default_target=self.default_target),
+            ConfigureDefaultDesktopTask(sysroot=sysroot, default_desktop=self.default_desktop),
```
poncovka: Ideally, this task shouldn't be created if `default_desktop` is not set. However, it is ok for now.

pyanaconda/modules/services/installation.py
```diff
@@ -123,3 +126,77 @@ def run(self):
...
+        """Create a new systemd target configuration task.
+
+        :param str sysroot: a path to the root of the target system
+        :param default_target: systemd default target to be set
+        """
+        super().__init__()
+        self._sysroot = sysroot
+        # default to the text only target
+        if not default_target:
+            default_target = TEXT_ONLY_TARGET
```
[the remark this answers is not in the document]
M4rtinK: It's apparently set to "" in the module as otherwise it would always show up in the output kickstart. But I guess we could do that better - add a flag if it showed up in the input kickstart or was set via DBus, which would trigger it being written out.

pyanaconda/modules/services/installation.py
M4rtinK: I would say pretty unlikely - lot of the code in the desktop.py file likely predates systemd being in wider use. I think we can drop it.

M4rtinK:
Added a set for kickstart tests for the default desktop & default systemd target configuration:
https://github.com/rhinstaller/kickstart-tests/pull/263

pyanaconda/modules/services/installation.py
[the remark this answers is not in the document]
poncovka: You can return the default value in the property `default_target`, if `_default_target` is not set, and generate kickstart based on the attribute `_default_target`. I think we already do something similar somewhere else.

pyanaconda/modules/services/installation.py
```diff
@@ -123,3 +126,77 @@ def run(self):
...
+        # symlink the selected target
+        selected_target_path = os.path.join(self._sysroot, 'etc/systemd/system', self._default_target)
+        log.debug("Linking {} as systemd default target.".format(selected_target_path) )
+        os.symlink(selected_target_path, default_target_path)
+
+
+class ConfigureDefaultDesktopTask(Task):
+    """Installation task for configuring the default desktop."""
+
+    def __init__(self, sysroot, default_desktop=None):
```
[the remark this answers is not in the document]
M4rtinK: Fixed.

pyanaconda/modules/services/installation.py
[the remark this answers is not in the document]
M4rtinK: OK, let's do that, but that still pretty ugly. :D

pyanaconda/modules/services/installation.py
```diff
@@ -123,3 +126,77 @@ def run(self):
...
+
+    Set the correct systemd default target for the target system.
+
+    We support setting either the text only "multi-user"
+    target or the graphical target called "graphical".
+    """
+
+    SUPPORTED_SYSTEMD_TARGETS = [TEXT_ONLY_TARGET, GRAPHICAL_TARGET]
+
+    def __init__(self, sysroot, default_target=None):
```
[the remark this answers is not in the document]
M4rtinK: Fixed.

pyanaconda/modules/services/installation.py
```diff
@@ -123,3 +126,77 @@ def run(self):
...
+        if not default_target:
+            default_target = TEXT_ONLY_TARGET
+        self._default_target = default_target
+
+    @property
+    def name(self):
+        return "Configure systemd default target"
+
+    def run(self):
+        if self._default_target not in self.SUPPORTED_SYSTEMD_TARGETS:
```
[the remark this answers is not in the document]
M4rtinK: Sure, but I wonder how we handle stuff like this in general - you set call SetFoo() over DBus and get an exception. So lets do that and see what happens. :)

pyanaconda/modules/services/installation.py
M4rtinK: should have been "unlink any links already in place", fixed

pyanaconda/modules/services/services.py
M4rtinK: Yeah, we have other tasks like this. We should handle this better later on.

pyanaconda/modules/services/services.py
```diff
@@ -165,13 +172,23 @@ def set_disabled_services(self, services):
...
     def set_default_target(self, target):
         """Set the default target of the installed system.
 
         :param target: a string with the target
         """
+        if target not in self.SUPPORTED_SYSTEMD_TARGETS:
+            msg = "Unsupported systemd default target: {} Specify one of: {}".format(
+                target, self.SUPPORTED_SYSTEMD_TARGETS
+            )
+            raise SystemdDefaultTargetConfigurationError(msg)
```
poncovka: Please, raise `ValueError` here. There no need for a new exception.

pyanaconda/modules/common/errors/installation.py
```diff
@@ -48,3 +48,8 @@ class NetworkInstallationError(InstallationError):
 class FirewallConfigurationError(InstallationError):
     """Exception for the firewall configuration errors."""
     pass
+
+@dbus_error("SystemdDefaultTargetConfigurationError", namespace=ANACONDA_NAMESPACE)
+class SystemdDefaultTargetConfigurationError(InstallationError):
```
poncovka: We don't need this exception anymore.

pyanaconda/modules/services/installation.py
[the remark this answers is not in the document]
poncovka: This code can be removed now.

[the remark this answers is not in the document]
M4rtinK: Removed.

pyanaconda/modules/services/services.py
M4rtinK: Good point, indeed no longer needed if this check is not in a DBUS Task subclass.

pyanaconda/modules/common/errors/installation.py
M4rtinK: Removed.

tests/nosetests/pyanaconda_tests/module_services_test.py
```diff
@@ -23,14 +23,19 @@
 from textwrap import dedent
 from unittest.mock import patch
 
-from mock import Mock
+from mock import Mock, patch
```
poncovka: You should use the other `patch` imported above.

M4rtinK: Fixed.

PR #2005, opened 2019-06-12
Fix setting and reporting ready state in Source Spoke.

rvykydal:
Actually set _ready internal attribute before reporting being ready to the
summary hub.  Not doing that resulted in hangs (UI input request) for kickstart
noniteractive installations if the ready state was sent to the hub based on
payload thread finishing callback (_payload_finished()) earlier then from the
source spoke initialization thread (_initialize()).

### DIFF
--- a/pyanaconda/ui/gui/spokes/installation_source.py
+++ b/pyanaconda/ui/gui/spokes/installation_source.py
@@ -817,6 +817,7 @@ class SourceSpoke(NormalSpoke, GUISpokeInputCheckHandler):
 
     def _payload_finished(self):
         hubQ.send_ready("SoftwareSelectionSpoke", False)
+        self._ready = True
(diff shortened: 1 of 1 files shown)


rvykydal:
> What about creating a method which would always set the `_ready` attribute and then send the ready to the `hubQ`? What do you think?

I was thinking about it but it didn't seem worth it to me, but if you want...

rvykydal:
> > What about creating a method which would always set the `_ready` attribute and then send the ready to the `hubQ`? What do you think?
> 
> I was thinking about it but it didn't seem worth it to me, but if you want...

Ah, and now I remember why: Do we want to wait for self.ready as in _initialize in the method then? I didn't want to spare time on deciding that now. I'd propose to go with this patch and maybe open another PR for moving it to the method (I wouldn't to that)?

PR #2007, opened 2019-06-12
Call teardown_all explicitly

poncovka:
The Blivet's `populate` method doesn't call `teardown_all` anymore,
so we have to call it explicitly after populating the device tree and
resetting the storage. It allows us to protect devices from teardown
during the reset.

**Depends on:** https://github.com/storaged-project/blivet/pull/780

### DIFF
--- a/pyanaconda/storage/osinstall.py
+++ b/pyanaconda/storage/osinstall.py
@@ -218,12 +218,14 @@ class InstallerStorage(Blivet):
 
         super().reset(cleanup_only=cleanup_only)
 
+        # Protect devices from teardown.
+        self._mark_protected_devices()
+        self.devicetree.teardown_all()
+
         self.fsset = FSSet(self.devicetree)
 
         # Clear out attributes that refer to devices that are no longer in the tree.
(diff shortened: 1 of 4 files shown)


pyanaconda/storage/osinstall.py
```diff
@@ -230,13 +230,15 @@ def reset(self, cleanup_only=False):
...
+        self._mark_protected_devices()
+        self.devicetree.teardown_all()
+
         self.fsset = FSSet(self.devicetree)
 
         # Clear out attributes that refer to devices that are no longer in the tree.
         self.bootloader.reset()
 
-        self._mark_protected_devices()
-
```
jkonecny12: What is the reason to move this code before FSSet and reset of the bootloader?

poncovka: To protect devices from teardown and make the teardown as soon after reset as possible.

jkonecny12: OK, thanks for the explanation.

poncovka:
Changed version of python3-blivet.

PR #2010, opened 2019-06-13
Add the apply-updates script

poncovka:
Run apply-updates to download and extract the given updates image.
Anaconda will use the updated files during the next run.

### DIFF
--- a/scripts/apply-updates
+++ b/scripts/apply-updates
@@ -1,0 +1,43 @@
+#!/bin/bash
+#
+# Copyright (C) 2019  Red Hat, Inc.
+#
+# Download and extract the updates image.
+#
+# How to apply the updates image?
+#   apply-updates http://myserver/updates.img
+#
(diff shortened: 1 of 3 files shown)


scripts/apply-updates
```diff
@@ -0,0 +1,38 @@
...
+if [[ -z "${URL}" ]]; then
+    echo "Specify the location of the updates image." 1>&2
+    exit 1
+fi
+
+# Clear /tmp/updates, so Anaconda can set it up again.
+rm -rf /tmp/updates
+
+# Download and extract the updates image.
+(cd / ; curl -f "${URL}" | gzip -dc  | cpio -idu )
```
jkonecny12: For future extension it should also support tarball image files which are supported by Anaconda if I'm not mistaken. But this is fine for now.

poncovka:
Updated.

PR #2011, opened 2019-06-13
Add support for disabling modules via module --disable (#1655523)

M4rtinK:
Make it possible to disable modules via the module command --disable option.

Resolves: rhbz#1655523
Related: rhbz#1719347

### DIFF
--- a/pyanaconda/payload/dnfpayload.py
+++ b/pyanaconda/payload/dnfpayload.py
@@ -453,23 +453,37 @@ class DNFPayload(payload.PackagePayload):
         self._add_repo(ksrepo)
         super().addDisabledRepo(ksrepo)
 
-    def _enable_modules(self):
-        """Enable modules (if any)."""
+    def _process_module_command(self):
+        """Enable/disable modules (if any)."""
         # convert data from kickstart to module specs
-        module_specs = []
+        module_specs_to_enable = []
+        module_specs_to_disable = []
         for module in self.data.module.dataList():
             # stream definition is optional
             if module.stream:
                 module_spec = "{name}:{stream}".format(name=module.name, stream=module.stream)
             else:
                 module_spec = module.name
-            module_specs.append(module_spec)
 
-        # forward the module specs to enable to DNF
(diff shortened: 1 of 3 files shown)


M4rtinK:
updated:
- bump pykickstart version in spec file to expected value, might need to be checked later if it matches actual pykickstart version
- fixed some pylint issues, remaining test errors seem to be caused by the pykickstart part of this change missing from test environment

pyanaconda/payload/dnfpayload.py
```diff
@@ -453,23 +453,37 @@ def addDisabledRepo(self, ksrepo):
         self._add_repo(ksrepo)
         super().addDisabledRepo(ksrepo)
 
-    def _enable_modules(self):
-        """Enable modules (if any)."""
+    def _handle_modules(self):
```
jkonecny12: I don't like handle, could we switch this to something more descriptive?

pyanaconda/payload/dnfpayload.py
```diff
@@ -453,23 +453,37 @@ def addDisabledRepo(self, ksrepo):
...
                 module_spec = module.name
-            module_specs.append(module_spec)
+
+            if module.enable:
+                module_specs_to_enable.append(module_spec)
+            else:
+                module_specs_to_disable.append(module_spec)
+
+        # forward the module specs to disable to DNF
+        log.debug("disabling modules: %s", module_specs_to_disable)
```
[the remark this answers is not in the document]
M4rtinK: If empty `module_specs_to_disable` should be an empty list, so `[]`. So no quotes needed IMHO. :)

pyanaconda/payload/dnfpayload.py
M4rtinK: At the moment (pre-payload modularization) it processes the kickstart module command, so let's rename it that way.

M4rtinK:
jenkins, test this please

rvykydal:
Ported to master with https://github.com/rhinstaller/anaconda/pull/2084

PR #2014, opened 2019-06-14
Move updates repo names to constants

jcpunk:
Rather than having the updates repo names hard coded this moves them into the constants list so that they can be more centrally reviewed.  And maybe moved into the products.d definitions....

### DIFF
--- a/pyanaconda/core/constants.py
+++ b/pyanaconda/core/constants.py
@@ -66,6 +66,10 @@ DEFAULT_REPOS = [productName.split('-')[0].lower(),
                  "rawhide",
                  "BaseOS"]
 
+# Get list of repo names which should be used as updates repos
+DEFAULT_UPDATE_REPOS = ["updates",
+                        "updates-modular"]
+
 ANACONDA_BUS_CONF_FILE = "/usr/share/anaconda/dbus/anaconda-bus.conf"
 ANACONDA_BUS_ADDR_FILE = "/run/anaconda/bus.address"
 
--- a/pyanaconda/payload/dnfpayload.py
+++ b/pyanaconda/payload/dnfpayload.py
@@ -907,11 +907,11 @@ class DNFPayload(payload.PackagePayload):
 
         # Enable or disable updates.
         if self._updates_enabled:
-            self.enable_repo("updates")
-            self.enable_repo("updates-modular")
+            for repo in constants.DEFAULT_UPDATE_REPOS:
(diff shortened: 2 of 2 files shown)


pyanaconda/payload/dnfpayload.py
```diff
@@ -907,11 +907,11 @@ def set_updates_enabled(self, state):
 
         # Enable or disable updates.
         if self._updates_enabled:
-            self.enable_repo("updates")
-            self.enable_repo("updates-modular")
+            for repo in constants.DEFAULT_REPOS:
```
poncovka: It should be `DEFAULT_UPDATE_REPOS`, same below.

pyanaconda/core/constants.py
```diff
@@ -66,6 +66,11 @@
                  "rawhide",
                  "BaseOS"]
 
+# Get list of repo names which should be used as updates repos
+DEFAULT_UPDATE_REPOS = [productName.split('-')[0].lower() + '-updates',
```
poncovka: Please, remove the first repo from the list. We enable or disable only `updates` and `updates-modular` at this moment.

pyanaconda/payload/dnfpayload.py
jcpunk: I'll fix this shortly.

pyanaconda/core/constants.py
jcpunk: Right now `DEFAULT_REPOS` can find a repo off of productName for the BaseOS.  Adding similar behavior for a productName specific updates repo would make them more parallel.

jkonecny12: Yes, you are right but we would like to change this solution completely. We would like to avoid matching stuff based on product name it is not that clear how this should work and it breaks custom images often. The only thing using this is Fedora now so there is no reason to have this shadow logic just for one system. Also you are blocking systems based on Fedora in renaming their repos / product names.

Or do you want to use this somewhere else?

jkonecny12:
Jenkins, it's ok to test.

pyanaconda/core/constants.py
jcpunk: My use case is fairly soft and I can solve another way.  I'll drop the productName updates and update the patch.

jcpunk:
Patch updated based on comments.

poncovka:
Jenkins, it's ok to test.

PR #2015, opened 2019-06-14
default bootfs to product file_system_type if valid

jcpunk:
Right now anaconda has a special path for bootloaders that doesn't consider the product default filesystem type.

With this fix, each bootloader can list its preferred filesystems, but if the product Default File System is compatible it will become the default file system for the bootloader.

This PR will drop two patches carried by rebuilds and eliminate the need for a RHEL specific path.

### DIFF
--- a/pyanaconda/bootloader/grub2.py
+++ b/pyanaconda/bootloader/grub2.py
@@ -183,10 +183,12 @@ class GRUB2(BootLoader):
 
     @property
     def stage2_format_types(self):
-        if productName.startswith("Red Hat "): # pylint: disable=no-member
-            return ["xfs", "ext4", "ext3", "ext2"]
-        else:
-            return ["ext4", "ext3", "ext2", "btrfs", "xfs"]
+        # If the product file_system_type is valid, use it by default
+        valid_fs = ["ext4", "ext3", "ext2", "btrfs", "xfs"]
+        if conf.storage.file_system_type in valid_fs:
+            valid_fs.insert(0, conf.storage.file_system_type)
+
+        return valid_fs
 
     #
     # grub-related conveniences
--- a/pyanaconda/bootloader/zipl.py
+++ b/pyanaconda/bootloader/zipl.py
@@ -46,10 +46,12 @@ class ZIPL(BootLoader):
 
     @property
     def stage2_format_types(self):
-        if productName.startswith("Red Hat "):          # pylint: disable=no-member
-            return ["xfs", "ext4", "ext3", "ext2"]
-        else:
-            return ["ext4", "ext3", "ext2", "xfs"]
+        # If the product file_system_type is valid, use it by default
+        valid_fs = ["ext4", "ext3", "ext2", "xfs"]
+        if conf.storage.file_system_type in valid_fs:
+            valid_fs.insert(0, conf.storage.file_system_type)
+
+        return valid_fs
 
     image_label_attr = "short_label"
     preserve_args = ["cio_ignore", "rd.znet", "rd_ZNET", "zfcp.allow_lun_scan"]
--- a/pyanaconda/bootloader/base.py
+++ b/pyanaconda/bootloader/base.py
@@ -152,7 +152,12 @@ class BootLoader(object):
 
     @property
     def stage2_format_types(self):
-        return ["ext4", "ext3", "ext2"]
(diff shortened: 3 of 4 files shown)


pyanaconda/bootloader/grub2.py
```diff
@@ -183,10 +183,12 @@ def write_config_images(self, config):
 
     @property
     def stage2_format_types(self):
-        if productName.startswith("Red Hat "): # pylint: disable=no-member
-            return ["xfs", "ext4", "ext3", "ext2"]
-        else:
-            return ["ext4", "ext3", "ext2", "btrfs", "xfs"]
+        # If the product file_system_type is valid, use it by default
+        valid_fs = ["ext4", "ext3", "ext2", "btrfs", "xfs"]
```
poncovka: `btrfs` is not a valid format type for RHEL.

pyanaconda/bootloader/zipl.py
```diff
@@ -46,10 +46,12 @@ class ZIPL(BootLoader):
...
     @property
     def stage2_format_types(self):
-        if productName.startswith("Red Hat "):          # pylint: disable=no-member
-            return ["xfs", "ext4", "ext3", "ext2"]
-        else:
-            return ["ext4", "ext3", "ext2", "xfs"]
+        # If the product file_system_type is valid, use it by default
+        valid_fs = ["ext4", "ext3", "ext2", "xfs"]
+        if conf.storage.file_system_type in valid_fs:
+            valid_fs.insert(0, conf.storage.file_system_type)
```
poncovka: I don't like the repetitive code. There are two places where this property is used and only one of them is affected by this change, so it makes more sense to move this code to the `default_boot_fstype` property.

pyanaconda/bootloader/grub2.py
jcpunk: Perhaps the best way to do that is add an 'invalid_fs' to the Products?

Grub2 seems to work with `btrfs`, so I wouldn't want to drop it from the supported grub2 fs list.

pyanaconda/bootloader/zipl.py
jcpunk: My worry with having a `default_boot_fstype` is if someone sets that to a value that isn't supported by the bootloader Anaconda would do as instructed, but the system wouldn't be able to boot.

This logic permits a fall back to a boot fs that is compatible.

jkonecny12:
Jenkins, it's ok to test.

pyanaconda/bootloader/zipl.py
poncovka: The class `InstallerStorage` has a property `default_boot_fstype` that uses the first item from `stage2_format_types` if no default boot fstype is set.

```
@property
def default_boot_fstype(self):
    """The default filesystem type for the boot partition."""
    if self._default_boot_fstype:
        return self._default_boot_fstype

    return self.bootloader.stage2_format_types[0]
```

This property can be extended like this:

```
@property
def default_boot_fstype(self):
    """The default filesystem type for the boot partition."""
    if self._default_boot_fstype:
        return self._default_boot_fstype

    if self._default_fstype in self.bootloader.stage2_format_types:
        return self._default_fstype 

    return self.bootloader.stage2_format_types[0]
```

We set the default fstype to `conf.storage.file_system_type` in the function `create_storage` when we create the storage object, so the logic should be the same at the end.

This solution is more flexible and it looks like we might be able to remove the method `set_default_boot_fstype`. That would fix the issue with the unsupported default boot fstype that you mentioned, because it looks like this problem could happen with the `autopart` command.

pyanaconda/bootloader/grub2.py
poncovka: It looks like it will not be necessary to change `stage2_format_types` at all, so I would prefer to leave it as it is for now.

pyanaconda/bootloader/zipl.py
jcpunk: That is a much better solution!

I'll be out of the office for the next week or so... it may be a bit for me to reimplement.  Did you want to do it since you actually found the elegant path?

poncovka: Implemented at: https://github.com/rhinstaller/anaconda/pull/2026

poncovka:
I am closing this pull request, since we have decided to choose a different implementation.

PR #2017, opened 2019-06-19
Fix misleading comment

jcpunk:
This fixes a misleading comment so that the documented behavior matches the code.

### DIFF
--- a/pyanaconda/payload/dnfpayload.py
+++ b/pyanaconda/payload/dnfpayload.py
@@ -1141,8 +1141,9 @@ class DNFPayload(payload.PackagePayload):
         # Enable or disable updates.
         self.set_updates_enabled(self._updates_enabled)
 
-        # Repos on disk are always enabled. When reloaded their state needs to
-        # be synchronized with the user selection.
+        # Repo files are always loaded from the system.
+        # When reloaded their state needs to be synchronized with the user configuration.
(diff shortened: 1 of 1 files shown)


poncovka:
Jenkins, test this please.

pyanaconda/payload/dnfpayload.py
```diff
@@ -1141,8 +1141,8 @@ def update_base_repo(self, fallback=True, checkmount=True):
         # Enable or disable updates.
         self.set_updates_enabled(self._updates_enabled)
 
-        # Repos on disk are always enabled. When reloaded their state needs to
-        # be synchronized with the user selection.
+        # Repos from .treeinfo are always enabled.
+        # When reloaded their state needs to be synchronized with the user selection.
```
jkonecny12: This is not correct. The `.treeinfo` is not taken into account here at all. They are loaded later, line 1159. This is only about  repo files loaded from the local system. I would suggest this:

```
Repo files are always loaded from the system.
When reloaded their state needs to be synchronized with the user configuration. So disable them now and enable them later if required.
```

jcpunk: Just pushed a fix.

jkonecny12:
Jenkins, it's ok to test.

PR #2018, opened 2019-06-20
Fix list-screens script reference

gpchelkin:
Docs lead to wrong script filename for list-screens.


M4rtinK:
Thanks as well & merging. :)

PR #2019, opened 2019-06-20
Add DBus support for device trees of partitioning modules

poncovka:
Call the DBus method GetDeviceTree of a partitioning module to get
access to its device tree. The method will create a new device tree
module and publish it on a unique DBus path.

Create a new partitioning module for the custom partitioning spoke.

### DIFF
--- a/pyanaconda/modules/storage/devicetree/devicetree_interface.py
+++ b/pyanaconda/modules/storage/devicetree/devicetree_interface.py
@@ -19,6 +19,7 @@
 #
 
 from pyanaconda.dbus.interface import dbus_class
+from pyanaconda.dbus.namespace import get_dbus_path
 from pyanaconda.modules.storage.devicetree.handler_interface import DeviceTreeHandlerInterface
 from pyanaconda.modules.storage.devicetree.viewer_interface import DeviceTreeViewerInterface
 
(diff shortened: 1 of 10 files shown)


pyanaconda/modules/storage/devicetree/devicetree_interface.py
```diff
@@ -28,4 +29,18 @@
...
+        """Get the unique object path in the given namespace.
+
+        This method is not thread safe for now.
+
+        :param namespace: a sequence of names
+        :return: a DBus path of a device tree
+        """
+        tree_number = DeviceTreeInterface._tree_counter
+        DeviceTreeInterface._tree_counter += 1
+        return get_dbus_path(*namespace, "DeviceTree", str(tree_number))
```
jkonecny12: Wouldn't it be better to make this method classmethod and use cls variable?

pyanaconda/modules/storage/partitioning/base.py
```diff
@@ -67,6 +72,25 @@ def on_selected_disks_changed(self, selection):
...
+            self._device_tree_module = module
+
+        if not self._device_tree_path:
+            self._device_tree_path = publish_device_tree(
+                message_bus=DBus,
+                namespace=STORAGE.namespace,
+                device_tree=self._device_tree_module
+            )
+
+        return self._device_tree_path
```
jkonecny12: What about making this a property `device_tree`?

pyanaconda/modules/storage/devicetree/devicetree_interface.py
poncovka: No, because every subclass would have its own counter and that could cause name collisions.

pyanaconda/modules/storage/partitioning/base.py
poncovka: Properties have signals and other synchronization. I don't want to create this module until someone really asks for it.

pyanaconda/modules/storage/devicetree/devicetree_interface.py
jkonecny12: Fair point, thanks for explanation.

pyanaconda/modules/storage/partitioning/base.py
jkonecny12: I meant "private" property not part of the module API. That should be fine I guess.

poncovka: I don't think that this property could be useful to the python module, since the method returns a DBus path. I will probably eventually create a property for the `self._device_tree_module`, but it is not necessary at this moment.

jkonecny12: Ohh, OK, I thought it's an object not a path.

PR #2021, opened 2019-06-21
Drop redundant fallbacks to multi-user.target

M4rtinK:
With the Services DBus module making sure that
the multi-user.target is used if no target is set
we can drop the old fallbacks that are now redundant.


M4rtinK:
Found a better way to handle this.

PR #2022, opened 2019-06-21
Migrate liveos payload

jkonecny12:
Implement Live OS payload handler with tasks.

**This PR will be merged as partial solution to avoid blocking of others work.**

TODO list:
* update kernel list task (won't be a task but a method)
* space required attribute
* pre-install task?
* install task
* post-install task
* progress reporting
* tests


pyanaconda/modules/common/errors/payload.py
```diff
@@ -0,0 +1,28 @@
...
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+
+from pyanaconda.dbus.error import dbus_error
+from pyanaconda.modules.common.constants.namespaces import PAYLOAD_NAMESPACE
+from pyanaconda.modules.common.errors import AnacondaError
+
+
+@dbus_error("PayloadSourceSetupError", namespace=PAYLOAD_NAMESPACE)
+class PayloadSourceSetupError(AnacondaError):
```
poncovka: I would remove `Payload` from both names of the error. The class will be imported from `payload.py`, so it is clear that it will be related to payload, and the DBus error uses the payload namespace, so `Payload` is already part of its name.

pyanaconda/modules/payload/live/initialization.py
```diff
@@ -0,0 +1,81 @@
...
+
+    @property
+    def name(self):
+        return "Setup Installation Source"
+
+    def run(self):
+        """Run live installation source setup."""
+        # Mount the live device and copy from it instead of the overlay at /
+        device_tree = STORAGE.get_proxy(DEVICE_TREE)
+        dev = device_tree.ResolveDevice(self._live_partition)
```
poncovka: Please, rename it to `device_name`.

pyanaconda/modules/payload/live/live_os.py
```diff
@@ -0,0 +1,84 @@
...
+        self.image_path_changed.emit()
+        log.debug("LiveOS image path is set to '%s'", self._image_path)
+
+    def setup_installation_source_with_task(self):
+        """Setup installation source device."""
+        task = SetupInstallationSourceTask(self.image_path, INSTALL_TREE)
+
+        return self.publish_task(LIVE_OS_HANDLER.namespace, task)
+
+    def unsetup_installation_source_with_task(self):
```
poncovka: Maybe I would use `teardown` to make it consistent with the Storage module. Also, `unsetup` doesn't exist :-)

pyanaconda/modules/payload/payload.py
```diff
@@ -60,7 +63,9 @@ def process_kickstart(self, data):
         self._payload_handler.process_kickstart(data)
 
     def _create_correct_handler(self, data):
-        if data.liveimg.seen:
+        if conf.system.provides_live_base_image:
+            self._payload_handler = LiveOSHandlerModule()
```
poncovka: This is not necessary, right? Since this handler cannot be requested by kickstart.

pyanaconda/modules/payload/live/live_os.py
jkonecny12: I'm not sure about `teardown`. It sounds to me like you want teardown device especially because it's used in the storage but there's more about the installation source.
The `unsetup` is from the original payload.

pyanaconda/modules/payload/payload.py
jkonecny12: Yes, this can be removed.

pyanaconda/modules/common/errors/payload.py
jkonecny12: Fixed.

pyanaconda/modules/payload/live/initialization.py
jkonecny12: Fixed.

pyanaconda/modules/payload/live/live_os.py
jkonecny12: Fixed.

pyanaconda/modules/payload/payload.py
jkonecny12: Fixed.

jkonecny12:
**UPDATED**:
* Fixed issues noted by Vendy.

jkonecny12:
**UPDATED:**
* Add live os base image interface
* Add tests for this new method

The reason why the code is not called automatically on handler creation is, that it makes testing of the code a lot harder. Also I don't like classes which are touching (even reading) system stuff on creation, it's prone to hard to track behavior errors.

pyanaconda/dbus/structure.py
```diff
@@ -258,6 +258,21 @@ def generate_dictionary_from_data(obj):
     }
 
 
+def generate_dbus_dictionary_from_data(obj):
```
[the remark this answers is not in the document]
jkonecny12: Fixed. Thanks!

jkonecny12:
**UPDATED:**
* Switched to `get_native` solution from https://github.com/rhinstaller/anaconda/pull/2029 in tests.

jkonecny12:
jenkins, test this please

jkonecny12:
jenkins, test this please

PR #2023, opened 2019-06-21
Move graphical login detection to task & related fixes

M4rtinK:
Move graphical target detection to a DBus task & change how we detect if the default target has been set.

### DIFF
--- a/pyanaconda/modules/services/services.py
+++ b/pyanaconda/modules/services/services.py
@@ -105,12 +105,9 @@ class ServicesModule(KickstartModule):
         data.services.enabled = self.enabled_services
         data.services.disabled = self.disabled_services
 
-        # Use the _default_target attribute directly instead of the
-        # default_target property to differentiate if default
-        # target has been set.
-        if self._default_target == TEXT_ONLY_TARGET:
+        if self.default_target == TEXT_ONLY_TARGET:
             data.skipx.skipx = True
-        elif self._default_target == GRAPHICAL_TARGET:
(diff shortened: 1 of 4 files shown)


jkonecny12:
Please fix the tests first.

pyanaconda/modules/services/services.py
```diff
@@ -105,12 +105,9 @@ def generate_kickstart(self):
         data.services.enabled = self.enabled_services
         data.services.disabled = self.disabled_services
 
-        # Use the _default_target attribute directly instead of the
-        # default_target property to differentiate if default
-        # target has been set.
-        if self._default_target == TEXT_ONLY_TARGET:
+        if self.default_target == TEXT_ONLY_TARGET:
             data.skipx.skipx = True
-        elif self._default_target == GRAPHICAL_TARGET:
```
AdamWill: I don't think this is right, is it? If you're not changing the property implementation any more, you should leave this as is...

M4rtinK: Well, on the other hand I think we should avoid calling the variable backing the property where possible, or else we might hit an issue if we ever put some code in the property implementation that returns something else than the raw variable.

And we already check the value to be equal to something know, so the default value being returned sometimes should not be an issue.

AdamWill: d'oh, never mind - I thought you hadn't changed the property definition back to the simple version, but you did. So this is fine.

PR #2026, opened 2019-06-24
Use the default file system type for /boot

poncovka:
Use the default file system type for /boot if this type is supported
by the bootloader, otherwise choose the first one from the list
of supported file system types.

Remove the method set_default_boot_fstype. It is enough to set up
the default file system type with set_default_fstype now.

Suggested-by: Pat Riehecky


poncovka:
Jenkins, test this please.

PR #2028, opened 2019-06-25
Fix the unit tests for the Services module

poncovka:
There is no default value of the default target.


poncovka:
Jenkins, test this please.

PR #2031, opened 2019-06-26
Add payload state to Payload Live OS handler

jkonecny12:
This state is defined in main payload module because it will be shared between payloads.


jkonecny12:
Closed. After our discussion with @poncovka we will move the state to Sources and do only a validation in the handlers.

PR #2032, opened 2019-06-26
Add the option decorated_window to the Anaconda configuration

poncovka:
Add a new configuration option decorated_window to the User Interface
section of the Anaconda configuration file and use it instead of the
Anaconda option decorated.


poncovka:
Updated the commit message.

PR #2033, opened 2019-06-26
Fix default value for pwpolicy emptyok (#1664704)

M4rtinK:
The pwpolicy command supported in the %anaconda
section of kickstart uses KSOptionParser provided
by the Pykickstart module.

In RHEL7 Pykickstart still uses OptionParser for option
parsing, while on RHEL8 & in Fedora Pykickstart now uses
Argparse. This has apprently caused a subtle regression
in the default value of the emptyok option of the
pwpolicy command. On RHEL7 the default value was True
which on RHEL8 it changed to False.

So explicitely set the default value of the emptyok option
to be True, to fix this issue.

Resolves: rhbz#1664704

### DIFF
--- a/pyanaconda/pwpolicy.py
+++ b/pyanaconda/pwpolicy.py
@@ -131,8 +131,8 @@ class F22_PwPolicy(KickstartCommand):
                         version=F22, help="""
                         Do not allow UI to be used to change the password/user
                         if it has been set in the kickstart.""")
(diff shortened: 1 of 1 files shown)


rvykydal:
PR with port to Fedora: https://github.com/rhinstaller/anaconda/pull/2444

rvykydal:
Ported to master with https://github.com/rhinstaller/anaconda/pull/2444

PR #2037, opened 2019-06-28
Add root password SSH login override checkbox (#1716282)

M4rtinK:
Open SSH, which provides the SSH server Fedora uses already for some
time disallowed password based root logins. Fedora used to patch this
out so far, but this patch has been recently dropped, restoring the
upstream behavior.

This change does not affect key based SSH logins and it will continue
to be possible to login as root with key based authentication.

To make the transition easier for user who might have valid use cases
for logging in as root with password over SSH we will add a checkbox
to the root password spoke. This checkbox makes it possible to manually
enable password based SSH logins for the root account.

Resolves: rhbz#1716282

**TODO:**
- ~unit tests for DBus API~
- ~unit tests for DBus Task~

### DIFF
--- a/pyanaconda/modules/users/installation.py
+++ b/pyanaconda/modules/users/installation.py
@@ -16,6 +16,8 @@
 # Red Hat, Inc.
 #
 
+import os
+
 from pyanaconda.core import users
 
 from pyanaconda.modules.common.task import Task
@@ -24,7 +26,8 @@ from pyanaconda.modules.common.structures.user import USER_GID_NOT_SET, USER_UID
 from pyanaconda.anaconda_loggers import get_module_logger
 log = get_module_logger(__name__)
 
-__all__ = ["SetRootPasswordTask", "CreateUsersTask", "CreateGroupsTask", "SetSshKeysTask"]
+__all__ = ["SetRootPasswordTask", "CreateUsersTask", "CreateGroupsTask", "SetSshKeysTask",
+           "ConfigureRootPasswordSSHLoginTask"]
 
 
 class SetRootPasswordTask(Task):
@@ -161,3 +164,37 @@ class SetSshKeysTask(Task):
     def _set_ssh_keys(self):
         for key_data in self._ssh_key_data_list:
             users.set_user_ssh_key(key_data.username, key_data.key)
+
+class ConfigureRootPasswordSSHLoginTask(Task):
+    """Optionally add an override allowing root to login with password over SSH."""
+
+    SSHD_OVERRIDE_DROP_DIR = "etc/systemd/system/sshd.service.d/"
+    OVERRIDE_CONFIG_NAME = "anaconda.conf"
+
+    def __init__(self, sysroot, password_allowed):
+        """Create a new root password SSH login configuration task.
+
+        :param str sysroot: a path to the root of the installed system
+        :param bool password_allowed: True allows root to login via SSH with password auth.
+                                      False prevents it by not changing the default OpenSSH
+                                      behavior
+        """
+        super().__init__()
+        self._sysroot = sysroot
+        self._password_allowed = password_allowed
+
+    @property
+    def name(self):
+        return "Configure optional root password SSH login"
+
+    def run(self):
+        # TODO: don't instantiate and run this Task if the override is not needed
+        if self._password_allowed:
+            log.debug("adding an override allowing root login with password via SSH")
+            dropdir_path = os.path.join(self._sysroot, self.SSHD_OVERRIDE_DROP_DIR)
+            os.makedirs(dropdir_path)
+            config_path = os.path.join(dropdir_path, self.OVERRIDE_CONFIG_NAME)
+            with open(config_path, "wt") as f:
+                f.write('[Service]\nEnvironment=OPTIONS="-oPermitRootLogin=yes"')
+        else:
+            log.debug("not adding an override allowing root login with password via SSH")
--- a/pyanaconda/modules/users/users.py
+++ b/pyanaconda/modules/users/users.py
@@ -28,7 +28,8 @@ from pyanaconda.modules.common.structures.sshkey import SshKeyData
 from pyanaconda.modules.users.kickstart import UsersKickstartSpecification
 from pyanaconda.modules.users.users_interface import UsersInterface
 from pyanaconda.modules.users.installation import SetRootPasswordTask, CreateUsersTask, \
-                                                  CreateGroupsTask, SetSshKeysTask
+                                                  CreateGroupsTask, SetSshKeysTask, \
+                                                  ConfigureRootPasswordSSHLoginTask
 
 from pyanaconda.anaconda_loggers import get_module_logger
 log = get_module_logger(__name__)
@@ -48,6 +49,9 @@ class UsersModule(KickstartModule):
 
         self.root_account_locked_changed = Signal()
         self._root_account_locked = True
+
+        self._root_password_ssh_login_allowed = False
+        self.root_password_ssh_login_allowed_changed = Signal()
 
         self.users_changed = Signal()
         self._users = []
@@ -173,6 +177,18 @@ class UsersModule(KickstartModule):
         task = SetSshKeysTask(sysroot=sysroot, ssh_key_data_list=self.ssh_keys)
         return self.publish_task(USERS.namespace, task)
 
+    def configure_root_password_ssh_login_with_task(self, sysroot):
+        """Return the root password SSH login configuration task.
+
+        :param str sysroot: a path to the root of the installed system
+        :param bool root_password_ssh_login_allowed: if root login via SSH with password
+                                                     should be allowed
+        :returns: object path of the root password SSH login configuration task
+        """
+        task = ConfigureRootPasswordSSHLoginTask(sysroot=sysroot,
+                                                 password_allowed=self.root_password_ssh_login_allowed)
+        return self.publish_task(USERS.namespace, task)
+
     def install_with_tasks(self, sysroot):
         """Return the installation tasks of this module.
 
@@ -183,7 +199,8 @@ class UsersModule(KickstartModule):
             self.configure_groups_with_task(sysroot=sysroot),
             self.configure_users_with_task(sysroot=sysroot),
             self.set_root_password_with_task(sysroot=sysroot),
-            self.set_ssh_keys_with_task(sysroot=sysroot)
+            self.set_ssh_keys_with_task(sysroot=sysroot),
+            self.configure_root_password_ssh_login_with_task(sysroot=sysroot)
         ]
         return paths
 
@@ -353,6 +370,25 @@ class UsersModule(KickstartModule):
     def root_account_locked(self):
         """Is the root account locked ?"""
         return self._root_account_locked
+
+    def set_root_password_ssh_login_allowed(self, root_password_ssh_login_allowed):
+        """Allow/disable root login via SSH with password.
+
+        (Login as root with key is always allowed)
+
+        param bool root_password_ssh_login_allowed: True to allow, False to disallow
+        """
+        self._root_password_ssh_login_allowed = root_password_ssh_login_allowed
+        self.root_password_ssh_login_allowed_changed.emit()
+        if root_password_ssh_login_allowed:
+            log.debug("SSH login as root with password will be allowed.")
+        else:
+            log.debug("SSH login as root with password will not be allowed.")
+
+    @property
+    def root_password_ssh_login_allowed(self):
+        """Is logging in as root via SSH with password allowed ?"""
+        return self._root_password_ssh_login_allowed
 
     @property
     def check_admin_user_exists(self):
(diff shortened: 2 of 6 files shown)


pyanaconda/modules/users/installation.py
```diff
@@ -161,3 +163,33 @@ def run(self):
...
+        self._sysroot = sysroot
+        self._root_password_ssh_login_allowed = root_password_ssh_login_allowed
+
+    @property
+    def name(self):
+        return "Configure optional root password SSH login"
+
+    def run(self):
+        # TODO: don't instantiate and run this Task if the override is not needed
+        if self._root_password_ssh_login_allowed:
```
poncovka: Please, log a message if the configuration is not allowed.

pyanaconda/modules/users/installation.py
```diff
@@ -161,3 +163,33 @@ def run(self):
...
         for key_data in self._ssh_key_data_list:
             users.set_user_ssh_key(key_data.username, key_data.key)
+
+class ConfigureRootPasswordSSHLogin(Task):
+    """Optionally add an override allowing root to login with password over SSH."""
+
+    SSHD_OVERRIDE_DROP_DIR = "etc/systemd/system/sshd.service.d/"
+    OVERRIDE_CONFIG_NAME = "anaconda.conf"
+
+    def __init__(self, sysroot, root_password_ssh_login_allowed):
```
poncovka: Please, make the name of the parameter shorter. The meaning is clear from the description of the class.

pyanaconda/modules/users/installation.py
```diff
@@ -161,3 +163,33 @@ def run(self):
...
+
+    @property
+    def name(self):
+        return "Configure optional root password SSH login"
+
+    def run(self):
+        # TODO: don't instantiate and run this Task if the override is not needed
+        if self._root_password_ssh_login_allowed:
+            log.debug("adding an override allowing root login with password via SSH")
+            os.makedirs(os.path.join(self.sysroot, self.SSHD_OVERRIDE_DROP_DIR))
```
poncovka: Please, keep the path to the drop directory in a variable, so you can reuse it in the line below.

pyanaconda/modules/users/installation.py
```diff
@@ -161,3 +163,33 @@ def run(self):
     def _set_ssh_keys(self):
         for key_data in self._ssh_key_data_list:
             users.set_user_ssh_key(key_data.username, key_data.key)
+
+class ConfigureRootPasswordSSHLogin(Task):
```
poncovka: The class is not defined in `__all__`.

pyanaconda/modules/users/users.py
```diff
@@ -173,6 +177,18 @@ def set_ssh_keys_with_task(self, sysroot):
         task = SetSshKeysTask(sysroot=sysroot, ssh_key_data_list=self.ssh_keys)
         return self.publish_task(USERS.namespace, task)
 
+    def configure_root_password_ssh_login_with_task(self, sysroot):
```
poncovka: I guess that this is required because of the initial setup.

pyanaconda/modules/users/installation.py
```diff
@@ -161,3 +163,33 @@ def run(self):
...
+                                                     behavior
+        """
+        self._sysroot = sysroot
+        self._root_password_ssh_login_allowed = root_password_ssh_login_allowed
+
+    @property
+    def name(self):
+        return "Configure optional root password SSH login"
+
+    def run(self):
```
poncovka: Please, add a unit test for this task.

pyanaconda/modules/users/users_interface.py
```diff
@@ -107,6 +108,18 @@ def IsRootAccountLocked(self) -> Bool:
         """
         return self.implementation.root_account_locked
 
+    @emits_properties_changed
+    def SetRootPasswordSSHLoginAllowed(self, root_password_ssh_login_allowed: Bool):
+        """Allow or disallow the root from logging in via SSH with password authetication."""
+        self.implementation.set_root_password_ssh_login_allowed(root_password_ssh_login_allowed)
+
+    @property
+    def IsRootPasswordSSHLoginAllowed(self) -> Bool:
```
poncovka: Please, remove `Is`.

pyanaconda/modules/users/users_interface.py
```diff
@@ -107,6 +108,18 @@ def IsRootAccountLocked(self) -> Bool:
         """
         return self.implementation.root_account_locked
 
+    @emits_properties_changed
+    def SetRootPasswordSSHLoginAllowed(self, root_password_ssh_login_allowed: Bool):
```
poncovka: Please, add a unit test for the new DBus property.

pyanaconda/modules/users/installation.py
M4rtinK: Fixed, thanks! :)

pyanaconda/modules/users/installation.py
M4rtinK: Good point, changing.

pyanaconda/modules/users/installation.py
M4rtinK: Indeed, that can be shorter.

pyanaconda/modules/users/installation.py
M4rtinK: Will do, though ideally (as in the TODO) this Task should just be skipped in the future in such a case.

pyanaconda/modules/users/users.py
M4rtinK: Yep, IIRC it needs to be able to cherry pick these tasks at the moment. Actually, I think it will have to run the new task as well as it does have the root password spoke in some cases.

pyanaconda/modules/users/installation.py
M4rtinK: Added.

pyanaconda/modules/users/users_interface.py
M4rtinK: Removed.

pyanaconda/modules/users/users_interface.py
M4rtinK: Added.

pyanaconda/modules/users/installation.py
```diff
@@ -161,3 +164,37 @@ def run(self):
...
+        # TODO: don't instantiate and run this Task if the override is not needed
+        if self._password_allowed:
+            log.debug("adding an override allowing root login with password via SSH")
+            dropdir_path = os.path.join(self._sysroot, self.SSHD_OVERRIDE_DROP_DIR)
+            os.makedirs(dropdir_path)
+            config_path = os.path.join(dropdir_path, self.OVERRIDE_CONFIG_NAME)
+            with open(config_path, "wt") as f:
+                f.write('[Service]\nEnvironment=OPTIONS="-oPermitRootLogin=yes"')
```
Jakuje: Now, I am actually wondering whether it would make sense to use separate environment variable since the OPTIONS could be already used by some other applications/configuration management tool from `/etc/sysconfig/sshd` and would overwrite what we set here.

This worked as a proof of concept, but for the final solution, I would propose to use variable ANACONDA or something like that, which could be separately handled in the openssh service file. What do you think?

tests/nosetests/pyanaconda_tests/module_users_test.py
```diff
@@ -86,6 +91,12 @@ def lock_root_account_test(self):
         self.assertEqual(self.users_interface.IsRootAccountLocked, True)
         self.callback.assert_called_once_with(USERS.interface_name, {'IsRootAccountLocked': True}, [])
 
+    def allow_root_password_ssh_login_test(self):
+        """Test if root password SSH login can be allowed."""
+        self.users_interface.SetRootPasswordSSHLoginAllowed(True)
+        self.assertEqual(self.users_interface.RootPasswordSSHLoginAllowed, True)
+        self.callback.assert_called_once_with(USERS.interface_name, {'RootPasswordSSHLoginAllowed': True}, [])
```
poncovka: Usually, we use the function `check_dbus_property` for testing DBus properties.

pyanaconda/modules/users/installation.py
M4rtinK: > Now, I am actually wondering whether it would make sense to use separate environment variable since the OPTIONS could be already used by some other applications/configuration management tool from `/etc/sysconfig/sshd` and would overwrite what we set here.

I don't think this can really happen during an interactive graphical installation, which is the override checkbox targeting & you need to explicitly click the checkbox for the config file to be created here.
As for the file being overwritten later on - I wonder how often do tools add overrides to this place ? Do we have some concrete often used examples ? 

> 
> This worked as a proof of concept, but for the final solution, I would propose to use variable ANACONDA or something like that, which could be separately handled in the openssh service file. What do you think?

Wouldn't that mean that the openssh would have to carry & maintain an anaconda specific change ? I actually rather like how the current implementation is rather self contained. Do you possibly have some other more robust solution in mind that would not require the openssh package to cary some Anaconda specific behavior ?

M4rtinK:
> I think this will also require a modification of either commandline or sshd_config.anaconda (I am not sure where does this file come from).

I think it is created by the Lorax template when it creates the boot iso:
https://github.com/weldr/lorax/blob/8c2184d59eb2b8298e9012a3d92c2404191f2507/share/templates.d/99-generic/runtime-postinstall.tmpl#L83

Looks like it just takes `/etc/ssh` and renames it to `sshd_config.anaconda`. So I guess the template would have to be modified to add the option here ?

pyanaconda/modules/users/installation.py
Jakuje: > I don't think this can really happen during an interactive graphical installation, which is the
> override checkbox targeting & you need to explicitly click the checkbox for the config file to be
> created here.

Yes, this would not appear during the installation, but it could get broken afterward during some later configuraiton.

> As for the file being overwritten later on - I wonder how often do tools add overrides to this
> place ? Do we have some concrete often used examples ?

I do not know about any tool that would overwrite this at this moment. But it might be just because it just always worked so far and there were no issues with it. If anaconda will hijack this environment variable, nothing else can simply use it anymore. Additionally if we will want some other applications to modify this variable, we would have conflict at least with anaconda.

Therefore I would really rather see using separate variable in the service file, which will make it more fail-proof and we have this "linked" from both sides.

> Wouldn't that mean that the openssh would have to carry & maintain an anaconda specific
> change ? I actually rather like how the current implementation is rather self contained. 

Yes, in that case, OpenSSH would need to be modified to carry this variable, but at least it would clearly signalize "this variable comes from anaconda" and not "there is some variable from somewhere".

> Do you possibly have some other more robust solution in mind that would not require the
> openssh package to cary some Anaconda specific behavior ?

Unfortunately I do not have any better ideas if we do not want to modify the config directly (hint: we don't want).

Jakuje:
> > I think this will also require a modification of either commandline or sshd_config.anaconda (I am not sure where does this file come from).
> 
> I think it is created by the Lorax template when it creates the boot iso:
> https://github.com/weldr/lorax/blob/8c2184d59eb2b8298e9012a3d92c2404191f2507/share/templates.d/99-generic/runtime-postinstall.tmpl#L83
> 
> Looks like it just takes `/etc/ssh` and renames it to `sshd_config.anaconda`. So I guess the template would have to be modified to add the option here ?

Yes. Can you take care of it or ask someone who will know more about it? What is needed is only to modify the sshd_config.anaconda to contain `PermitRootLogin prohibit-password` if the original file does not allow it by default.

pyanaconda/modules/users/installation.py
M4rtinK: I think it would be good to consult the openssh package maintainers about this - about the ANACONDA option or possible other ways how the override could be implemented. I guess this could be done in the FESCO ticket if it is still active or perhaps in the bug ?

BTW, I'll be on vacation starting from tomorrow (3.7.2019) till the end of July basically (26.7.2019). So either this can wait till I'm back or you will have to reach out to others on the installer team (@poncovka, @rvykydal, @jkonecny12).

Jakuje: I am the OpenSSH package maintainer in Fedora (sorry for different name here).

Jakuje: @poncovka ping. Martin is away. Do any of you have some strong opinions about the above discussion? We will indeed have some time to get this into Fedora after he gets back, but I would like to avoid rushing it on last minute for Fedora 31 change.

Jakuje: @poncovka The openssh change is in place [1] as we agreed with a name `PERMITROOTLOGIN` so we can change it in this PR and progress to the merge. Thank you for the help to both of you!

[1] https://src.fedoraproject.org/rpms/openssh/c/0b10752b

Jakuje:
For the record, the sshd used in s390 installation and for debugging are using separate sshd config from lorax [1], which already has `PermitRootLogin yes` so there is really no action needed and this should continue working as it used to.

[1] https://github.com/weldr/lorax/blob/8c2184d59eb2b8298e9012a3d92c2404191f2507/share/templates.d/99-generic/config_files/common/sshd_config.anaconda

pyanaconda/modules/users/installation.py
poncovka: @Jakuje, thanks! Here is the new PR: https://github.com/rhinstaller/anaconda/pull/2042.

poncovka:
I am closing this pull request. The fix with `PERMITROOTLOGIN` is implemented at https://github.com/rhinstaller/anaconda/pull/2042.

PR #2044, opened 2019-07-10
Move all spokes to the Summary hub

poncovka:
Move spokes from the User Settings category to the Summary hub.

There should never be spokes on the Progress hub, so replace the hub
with a spoke. Join the task queues for installation and configuration
into one.

There is only one hub where the spokes can be displayed, so we can
remove the class attributes displayOnHubGUI and displayOnHubTUI,
because they are always set to the Summary hub.


poncovka:
no_banners_0

poncovka:
no_banners_1

poncovka:
no_banners_2

pyanaconda/ui/gui/spokes/installation_progress.py
```diff
@@ -218,28 +208,34 @@ def _cycle_rnotes(self):
...
-            continueText = self.builder.get_object("rebootLabel")
-            continueText.set_text(_("%s is now successfully installed on your system and ready "
-                                    "for you to use!  When you are ready, reboot your system to start using it!"))
-            continueText.set_line_wrap(True)
-            self.window.get_continue_button().set_label(C_("GUI|Progress", "_Quit"))
+        # Set the label of the continue button.
+        if conf.target.is_hardware and conf.system.can_reboot:
+            continue_label = C_("GUI|Progress", "_Reboot System")
+        else:
+            continue_label = C_("GUI|Progress", "_Finish Installation")
```
jkonecny12: Do we need this is there are no spokes and nothing can interrupt the installation?

poncovka: `Finish Installation` will appear instead of `Reboot System` when you are not allowed to reboot the system. That includes live, image and dir installations.

jkonecny12: Ahh true. OK in that case I guess we need it :D

jkonecny12:
Good point @rvykydal but on the other hand what is the probability that we will use that in the future?

pyanaconda/ui/gui/spokes/installation_progress.py
AdamWill: in the live case, the label previously was 'Quit', not 'Finish Installation'...

PR #2045, opened 2019-07-11
Tweak payload handlers

jkonecny12:
Handler is now property which will raise an exception when no handler is set.
Remove logic choosing active handler based on the KS data. This logic will be moved to Boss.
Add base class for all the handlers.
Create handler API of the main payload module now returns the path to the handler.
Add API to get the active handler.

EDIT: This is prepared for a merge.

### DIFF
--- a/pyanaconda/modules/payload/handler_base.py
+++ b/pyanaconda/modules/payload/handler_base.py
@@ -1,0 +1,39 @@
+#
+# Base object of all payload handlers.
+#
+# Copyright (C) 2018 Red Hat, Inc.
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# the GNU General Public License v.2, or (at your option) any later version.
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY expressed or implied, including the implied warranties of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General
+# Public License for more details.  You should have received a copy of the
+# GNU General Public License along with this program; if not, write to the
+# Free Software Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
+# 02110-1301, USA.  Any Red Hat trademarks that are incorporated in the
+# source code or documentation are not subject to the GNU General Public
+# License and may only be used or replicated with the express permission of
+# Red Hat, Inc.
+#
+from abc import ABCMeta, abstractmethod
+
+from pyanaconda.modules.common.base import KickstartBaseModule
+
+
+class PayloadHandlerBase(KickstartBaseModule, metaclass=ABCMeta):
+    """Base class for all the payload handler modules.
+
+    This will contain all API specific to payload handlers which will be called
+    by the base payload module.
+    """
+
+    @abstractmethod
+    def get_handler_path(self):
+        """Get path of this payload handler.
+
+        :returns: path to this handler
+        :rtype: string
+        """
+        pass
--- a/pyanaconda/modules/payload/handler_base.py
+++ b/pyanaconda/modules/payload/handler_base.py
@@ -37,12 +37,3 @@ class PayloadHandlerBase(KickstartBaseModule, metaclass=ABCMeta):
         :rtype: string
         """
         pass
-
-    @abstractmethod
-    def get_handler_path(self):
-        """Get path of this payload handler.
-
-        :returns: path to this handler
-        :rtype: string
-        """
-        pass
--- a/pyanaconda/modules/payload/handler_base.py
+++ b/pyanaconda/modules/payload/handler_base.py
@@ -30,6 +30,15 @@ class PayloadHandlerBase(KickstartBaseModule, metaclass=ABCMeta):
     """
 
     @abstractmethod
+    def publish_handler(self):
+        """Publish object on DBus and return its path.
+
+        :returns: path to this handler
+        :rtype: string
+        """
+        pass
+
+    @abstractmethod
     def get_handler_path(self):
         """Get path of this payload handler.
 
--- a/pyanaconda/modules/payload/live/live_os.py
+++ b/pyanaconda/modules/payload/live/live_os.py
@@ -26,7 +26,7 @@ from pyanaconda.core.signal import Signal
 from pyanaconda.core.constants import INSTALL_TREE
 
 from pyanaconda.modules.common.constants.objects import LIVE_OS_HANDLER
-from pyanaconda.modules.common.base import KickstartBaseModule
+from pyanaconda.modules.payload.handler_base import PayloadHandlerBase
 from pyanaconda.modules.payload.live.live_os_interface import LiveOSHandlerInterface
 from pyanaconda.modules.payload.live.initialization import SetupInstallationSourceTask, \
     TeardownInstallationSourceTask
@@ -35,7 +35,7 @@ from pyanaconda.anaconda_loggers import get_module_logger
 log = get_module_logger(__name__)
 
 
-class LiveOSHandlerModule(KickstartBaseModule):
+class LiveOSHandlerModule(PayloadHandlerBase):
     """The Live OS payload module."""
 
     def __init__(self):
@@ -53,6 +53,10 @@ class LiveOSHandlerModule(KickstartBaseModule):
 
     def setup_kickstart(self, data):
         """Setup the kickstart data."""
+
+    def get_handler_path(self):
+        """Get path of this payload handle."""
+        return LIVE_OS_HANDLER.object_path
 
     @property
     def image_path(self):
--- a/pyanaconda/modules/payload/live/live_os.py
+++ b/pyanaconda/modules/payload/live/live_os.py
@@ -44,9 +44,10 @@ class LiveOSHandlerModule(PayloadHandlerBase):
         self._image_path = ""
         self.image_path_changed = Signal()
 
-    def publish(self):
-        """Publish the module."""
+    def publish_handler(self):
+        """Publish the handler."""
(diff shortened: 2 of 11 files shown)


pyanaconda/modules/payload/payload_interface.py
```diff
@@ -27,9 +28,10 @@
...
-    def CreateLiveOSHandler(self):
-        """Create Live OS payload handler and publish it on dbus.
+    def CreateLiveOSHandler(self) -> ObjPath:
+        """Create Live OS payload handler and publish it on dbus."""
+        return self.implementation.create_live_os_handler()
 
-        # FIXME: This is a temporary solution which will change in the future commits
-        """
-        self.implementation.create_live_os_handler()
+    def GetActiveHandler(self) -> ObjPath:
```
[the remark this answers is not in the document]
jkonecny12: This name is on purpose and it is a preparation for the multiple payloads in the future. This will then return the main payload and we don't have to change the API later.

pyanaconda/modules/payload/live/live_os.py
```diff
@@ -54,6 +55,10 @@ def process_kickstart(self, data):
     def setup_kickstart(self, data):
         """Setup the kickstart data."""
 
+    def get_handler_path(self):
+        """Get path of this payload handle."""
+        return LIVE_OS_HANDLER.object_path
```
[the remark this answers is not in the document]
jkonecny12: I'm thinking opposite. I want to have this information in the handlers and easily accessible. The reason is again based on the multiple payloads modules in the future. It looks to me unnecessary complication to have logic in the payload to decide what will be the name of the handler instead of the handler, which knows where it should be published and how many of them are there now.

pyanaconda/modules/payload/handler_base.py
```diff
@@ -0,0 +1,37 @@
...
+# Free Software Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
+# 02110-1301, USA.  Any Red Hat trademarks that are incorporated in the
+# source code or documentation are not subject to the GNU General Public
+# License and may only be used or replicated with the express permission of
+# Red Hat, Inc.
+#
+from abc import ABC, abstractmethod
+
+
+class PayloadHandlerBase(ABC):
```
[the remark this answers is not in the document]
jkonecny12: Fixed.

pyanaconda/modules/payload/payload.py
```diff
@@ -44,41 +45,58 @@ def publish(self):
...
-            self._payload_handler = DNFHandlerModule()
+        self.payload_handler.process_kickstart(data)
 
     def generate_kickstart(self):
         """Return the kickstart string."""
         log.debug("Generating kickstart data...")
         data = self.get_kickstart_handler()
 
-        self._payload_handler.setup_kickstart(data)
+        self.payload_handler.setup_kickstart(data)
```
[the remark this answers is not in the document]
jkonecny12: Fixed.

pyanaconda/modules/payload/payload.py
```diff
@@ -44,41 +45,58 @@ def publish(self):
...
-        """Create the live os payload handler and publish it on dbus.
+        """Create the live os payload handler and publish it and return DBus path.
 
-        FIXME: This is only temporary and the solution to manage handlers will change.
+        :returns: DBus path to the handler
+        :rtype: str
         """
-        self._payload_handler = LiveOSHandlerModule()
+        self.set_payload_handler(LiveOSHandlerModule())
         self._publish_handler()
```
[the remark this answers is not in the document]
jkonecny12: Fixed, but I don't think it really matters. When there will be problem with publishing then the whole payload is screwed up anyway.

pyanaconda/modules/payload/payload.py
```diff
@@ -44,41 +45,58 @@ def publish(self):
...
+        :raises: HandlerNotSetError
+        """
+        log.debug("Processing kickstart data...")
 
-    def _create_correct_handler(self, data):
-        if data.liveimg.seen:
-            self._payload_handler = LiveImageHandlerModule()
-        else:
-            self._payload_handler = DNFHandlerModule()
+        self.payload_handler.process_kickstart(data)
```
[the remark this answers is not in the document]
jkonecny12: Fixed.

pyanaconda/modules/payload/payload_interface.py
```diff
@@ -27,9 +28,10 @@
...
+        """Get path to the payload which is used now."""
+        return self.implementation.payload_handler.get_handler_path()
```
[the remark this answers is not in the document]
jkonecny12: Fixed.

jkonecny12:
**UPDATED**:
* Fix conflicts
* Solve most of the issues above
* Add DNF payload
* Add tests

-------------------------
**TODO**:
* Add Live Image handler support
* Add configuration for the default payload

jkonecny12:
**UPDATED:**
* Add live image support
* Add default handler configuration
* Add tests

jkonecny12:
**UPDATED:**
* solve the conflict

jkonecny12:
**UPDATED:**
* Remove the default handler

pyanaconda/modules/payload/payload.py
```diff
@@ -42,43 +43,108 @@ def publish(self):
...
+        except HandlerNotSetError:
+            return ""
+
     def process_kickstart(self, data):
         """Process the kickstart data."""
         log.debug("Processing kickstart data...")
 
-        self._create_correct_handler(data)
-        self._publish_handler()
+        handler_class = self._get_correct_handler_class(data)
```
poncovka: I don't think that we need to do this in another method. The method `process_kickstart` is not going to get too big.

pyanaconda/modules/payload/payload.py
```diff
@@ -42,43 +43,108 @@ def publish(self):
...
+        self._publish_handler(handler)
+        self.set_payload_handler(handler)
+
+    @staticmethod
+    def _publish_handler(handler):
+        """Publish handler passed in.
+
+        This method is really helpful for testing purpose.
+        """
+        handler.publish()
```
poncovka: This method is really not necessary.

pyanaconda/modules/payload/payload_interface.py
```diff
@@ -27,9 +28,18 @@
 class PayloadInterface(KickstartModuleInterface):
     """DBus interface for Payload module."""
 
-    def CreateLiveOSHandler(self):
-        """Create Live OS payload handler and publish it on dbus.
+    def CreateDNFHandler(self) -> ObjPath:
```
poncovka: I think that we should have only one method `CreatePayloadHandler` with a type of the requested handler as an argument.

pyanaconda/modules/payload/payload.py
```diff
@@ -42,43 +43,108 @@ def publish(self):
...
-        self._create_correct_handler(data)
-        self._publish_handler()
+        handler_class = self._get_correct_handler_class(data)
+
+        if not handler_class:
+            log.warning("No handler was created. Kickstart data passed in are lost.")
+            return
 
-        self._payload_handler.process_kickstart(data)
+        self._initialize_handler(handler_class)
```
poncovka: The payload handler shouldn't be published and set in the main module until the kickstart data are processed.

pyanaconda/modules/payload/payload.py
```diff
@@ -42,43 +43,108 @@ def publish(self):
...
+
+    def _get_correct_handler_class(self, data):
         if data.liveimg.seen:
-            self._payload_handler = LiveImageHandlerModule()
+            return LiveImageHandlerModule
         else:
-            self._payload_handler = DNFHandlerModule()
+            return None
+
+    def _initialize_handler(self, handler_class):
```
poncovka: I would prefer to create handlers based on a requested type.

pyanaconda/modules/payload/live/live_os.py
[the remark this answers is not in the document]
poncovka: No Anaconda modules or submodules remember their DBus path, because they don't really care about that. In general, modules can be published multiple times with different DBus paths. The main payload module will have to keep the list of created payload handlers anyway, because the client can request to change the active payload handler. The request will be based on a DBus path of the handler, so the main payload module will have to find the handler with the given path. So the dictionary is the best choice anyway.

The payload handlers are not going to have static DBus paths, so they will have to be published on a path that is given to them.

pyanaconda/modules/payload/payload.py
jkonecny12: Fixed by moving this completely to Factory class.

pyanaconda/modules/payload/payload.py
jkonecny12: Fixed.

pyanaconda/modules/payload/payload.py
jkonecny12: Fixed.

pyanaconda/modules/payload/payload_interface.py
jkonecny12: Fixed.

pyanaconda/modules/payload/live/live_os.py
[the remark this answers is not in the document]
jkonecny12: Fixed.

pyanaconda/modules/payload/payload.py
jkonecny12: Fixed, found a better way how to test this.

jkonecny12:
**UPDATED:**
* Fixed notes above.

Thanks for your ideas @poncovka !

jkonecny12:
**UPDATED:**
* Small fix to removed not useful code from the factory.

jkonecny12:
**UPDATED:**
* Fixed the tests issue

PR #2046, opened 2019-07-11
Remove system root from DBus methods

poncovka:
Remove the argument for the system root from the DBus methods that
return installation and configuration tasks. The path to the system
root doesn't change anymore.

Replace functions getSysroot, setSysroot and getTargetPhysicalRoot.

**Depends on:**
https://github.com/rhinstaller/initial-setup/pull/73
https://github.com/daveyoung/kdump-anaconda-addon/pull/23

### DIFF
--- a/pyanaconda/core/util.py
+++ b/pyanaconda/core/util.py
@@ -85,21 +85,6 @@ def augmentEnv():
     return env
 
 
-def getTargetPhysicalRoot():
-    """Returns the path to the "physical" storage root, traditionally /mnt/sysimage.
-
-    This may be distinct from the sysroot, which could be a
-    chroot-type subdirectory of the physical root.  This is used for
-    example by all OSTree-based installations.
-    """
-
-    # We always use the traditional /mnt/sysimage - the physical OS
-    # target is never mounted anywhere else.  This API call just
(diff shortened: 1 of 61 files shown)


pyanaconda/core/util.py
```diff
@@ -85,7 +85,7 @@ def augmentEnv():
     return env
 
 
-def setSysroot(path):
+def set_system_root(path):
```
jkonecny12: Don't you want to rename this to `mount_system_root` or something similar. You are not setting path or anything now.

poncovka: I was thinking about it, but from the "user" point of view, you just want to tell Anaconda what is the actual path to the system root. The bind mount behind it is just an implementation detail, that could change again in the future. The name `mount_system_root` suggests that the real system root will be mounted somewhere and that is not correct.

jkonecny12: Still looks to me like a giant side effect. If we don't want to rename the method at least document that mount will happen and why it is happening.

poncovka: Fixed.

poncovka:
Rebased.

PR #2047, opened 2019-07-15
Add link for our translations to README

jkonecny12:
Resolves: rhbz#1729788

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -21,6 +21,8 @@ I want to contribute!
 
 Checkout out our easy guide at: https://anaconda-installer.readthedocs.io/en/latest/contributing.html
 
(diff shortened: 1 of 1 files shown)


bexelbie:
I think adding this is awesome, but we need to seriously consider going further and including the .po files used to build the project releases.  This both allows people to get all the code and recreate our work, but also ensures we have that data.  Zanata is not a supported platform and is literally on lights and power life support.  Alternatives are being explored, but a proper backup is crucial.

jkonecny12:
We should discuss this on a bug which reminds me that I forget to add BZ number here...

jkonecny12:
**UPDATED:**
* Add bz number to the commit message

PR #2051, opened 2019-07-23
Don't prune and sort actions in the partitioning task

poncovka:
Don't prune and sort actions in the interactive partitioning task,
because the scheduled partitioning doesn't have to be valid for the
storage checker. Organize the actions before displaying them to users
in the dialog window.


poncovka:
@vojtechtrefny, can you test Blivet-GUI, please?

PR #2052, opened 2019-07-24
Always schedule the installation task for joining realm (#1732620)

poncovka:
We don't know the result of the realm discovery during the scheduling,
because the task for the realm discovery hasn't run yet. So always
schedule the task for joining the realm and check the result there.

Resolves: rhbz#1732620


poncovka:
The commit message was updated.

PR #2053, opened 2019-07-24
Improve publisher patching decorator

jkonecny12:
We are patching publisher a lot so make a faster variant of this patch.

### DIFF
--- a/tests/nosetests/pyanaconda_tests/module_blivet_test.py
+++ b/tests/nosetests/pyanaconda_tests/module_blivet_test.py
@@ -22,7 +22,7 @@ import pickle
 import unittest
 from unittest.mock import patch, Mock
 
-from tests.nosetests.pyanaconda_tests import check_task_creation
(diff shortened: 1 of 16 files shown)


tests/nosetests/pyanaconda_tests/module_blivet_test.py
```diff
@@ -101,7 +101,7 @@ def send_request_test(self):
         self.assertIsInstance(answer, ProxyID)
         self.assertEqual(answer.id, 0)
 
-    @patch('pyanaconda.dbus.DBus.publish_object')
+    @patch_publisher
```
poncovka: Please, rename it to `patch_dbus_publish_object`, so it is clear what the publisher is. Otherwise, it is just confusing.

jkonecny12: Fixed.

PR #2055, opened 2019-07-25
Don't change the storage if the reset fails

poncovka:
Don't change the storage model of the Storage module unless the task
for the storage reset succeeds.


jkonecny12:
Please fix test before merge.

poncovka:
The test is fixed.

PR #2059, opened 2019-07-29
Remove the storage object from the rescue mode

poncovka:
**DO NOT MERGE!**

* Don't create the local storage object.
* Don't reset the local storage object.
* Don't use the local storage object to initialize time.
* Remove the storage object from the rescue mode.
* Don't use the local storage object during the start of anaconda.

### DIFF
--- a/anaconda.py
+++ b/anaconda.py
@@ -50,7 +50,7 @@ if ("inst.debug=1" in proc_cmdline) or ("inst.debug" in proc_cmdline):
 import atexit, sys, time, signal
 import pid
 
-def exitHandler(rebootData, storage):
+def exitHandler(rebootData):
     # Clear the list of watched PIDs.
     from pyanaconda.core.process_watchers import WatchProcesses
     WatchProcesses.unwatch_all_processes()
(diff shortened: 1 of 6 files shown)


anaconda.py
```diff
@@ -80,13 +80,12 @@ def exitHandler(rebootData, storage):
...
-        anaconda.storage.umount_filesystems(swapoff=False)
+    from pyanaconda.modules.common.task import sync_run_task
+    storage_proxy = STORAGE.get_proxy()
 
-    # Tear down disk images.
-    if conf.target.is_image:
-        anaconda.storage.devicetree.teardown_disk_images()
+    for task_path in storage_proxy.TeardownWithTasks():
+        task_proxy = STORAGE.get_proxy(task_path)
+        sync_run_task(task_proxy)
```
jkonecny12: I think this could be as a function like ``sync_run_all_tasks()``.

pyanaconda/rescue.py
```diff
@@ -251,41 +240,25 @@ def mount_root(self, root):
...
+            if not format_data.type == "luks":
+                continue
 
-        device = device_state.device
-        unlocked = unlock_device(self._storage, device, passphrase)
+            device_names.append(device_name)
 
-        if unlocked:
-            device_state.set_unlocked(passphrase)
+        return device_names
```
jkonecny12: I'm confused. Every device what does not have format type "luks" is a locked device? What am I missing? 

Also why is the method name ``get_locked_device_names`` and method description ``List of names of unlocked LUKS devices``?

poncovka: Yes, because we don't support devices unlocked outside of Anaconda.

The docstring was updated.

anaconda.py
poncovka: The teardown should be eventually handled by the boss, so I am not going to implement the new function for now.

jkonecny12: OK, I still think it has it's usages on the other places but that would be better done in a separate PR.

PR #2060, opened 2019-07-30
Do not try to mount cdrom source also if it has been mounted in dracut.

rvykydal:
Fix of commit 4ff5a55d905a43fe3455929cff1be0db80d9fb23

Resolves: rhbz#1733507

### DIFF
--- a/pyanaconda/payload/__init__.py
+++ b/pyanaconda/payload/__init__.py
@@ -1149,6 +1149,10 @@ class PackagePayload(Payload):
             else:
                 blivet.util.umount(mount_point)
 
+    def _device_is_mounted_as_source(self, device):
+        device_mounts = blivet.util.get_mount_paths(device.path)
+        return INSTALL_TREE in device_mounts or DRACUT_REPODIR in device_mounts
+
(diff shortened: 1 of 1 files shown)


pyanaconda/payload/__init__.py
```diff
@@ -1167,7 +1167,8 @@ def _setupMedia(self, device):
 
         # Check to see if the device is already mounted, in which case
         # we don't need to mount it again so skip this part
-        elif not (method.method == "cdrom" and INSTALL_TREE in blivet.util.get_mount_paths(device.path)):
+        elif not (method.method == "cdrom" and (INSTALL_TREE in blivet.util.get_mount_paths(device.path)
+                                                or DRACUT_REPODIR in blivet.util.get_mount_paths(device.path))):
```
jkonecny12: You are over 99 characters limit and this elif is not really readable. Could you please rewrite this to something more sane? For example to abstract this to a function or split to multiple if's.

rvykydal: Done

PR #2066, opened 2019-07-31
Specify sizes of nonexistent devices in tests

poncovka:
Specify sizes of nonexistent devices in tests. Otherwise, these tests
will fail with a message "device is too small for new format".


vojtechtrefny:
This was caused by blivet changing minimal size of filesystems from 0 to 2 MiB, see https://github.com/storaged-project/blivet/pull/690

PR #2068, opened 2019-07-31
Add dedicated log handler for addons

jcpunk:
While developing a new addon, I discovered the `anaconda_loggers` file.  I thought it would be nice to have a dedicated log handler for addons to use.  This way addon logs can go into a dedicated file for ease of development and troubleshooting.

Additionally, it would be good to demo logging best practices in https://github.com/rhinstaller/hello-world-anaconda-addon


jkonecny12:
Jenkins, it's ok to test.

M4rtinK:
Dedicated log handler for addons looks like a good idea, it can add a prefix & that can be easily filtered from Journal. As for separate log file for addons - this could be a bit more complicated as each modular addon is a separate Python process and it's not clear the Python logging machinery will be able to correctly handled that:

https://stackoverflow.com/questions/15096090/python-separate-processes-logging-to-same-file

One option is to have a per addon log file, alternatively we could start a process early on that would "sit" on journal, filter Anaconda stuff from it and then split it into log files for individual categories, such as addons, storage, network, etc.

jkonecny12:
Also we would have to add the log file to all the places which grabs Anaconda logs which wouldn't be easy.

poncovka:
Hi, it is an interesting idea for now, but this handler will not be usable in the future. Every addon will have to be implemented as a DBus service and multiple processes shouldn't log to the same file as @M4rtinK pointed out. We plan to do some changes in logging, but it is not really a priority at this moment. Thank you for your suggestion, but I am closing this pull request.

PR #2069, opened 2019-07-31
More detailed addon logging

jcpunk:
It would be handy to have anaconda spit out logs on what all it is loading for addons.

### DIFF
--- a/pyanaconda/addons.py
+++ b/pyanaconda/addons.py
@@ -57,14 +57,17 @@ def collect_addon_paths(toplevel_addon_paths, ui_subdir="gui"):
             addon_ks_path = os.path.join(path, addon_id, "ks")
             if os.path.isdir(addon_ks_path):
                 module_paths["ks"].append(("%s.ks.%%s" % addon_id, addon_ks_path))
+                log.debug('Loading ks section into module path for addon %s', addon_id)
 
             addon_spoke_path = os.path.join(path, addon_id, ui_subdir, "spokes")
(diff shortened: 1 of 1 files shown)


jkonecny12:
Jenkins, it's ok to test.

pyanaconda/addons.py
```diff
@@ -88,6 +91,7 @@ def execute(self, storage, ksdata, users, payload):
         for v in self.__dict__.values():
             if hasattr(v, "execute"):
                 progress_message(N_("Executing %s addon") % v.name)
+                log.debug("Executing %s addon", v.name)
```
poncovka: I think that the progress message above is logged as well, so it is not necessary to log it again.

pyanaconda/addons.py
```diff
@@ -103,6 +107,7 @@ def setup(self, storage, ksdata, payload):
         self.__dict__ = d
         for v in self.__dict__.values():
             if hasattr(v, "setup"):
+                log.debug("Loading %s addon", v.name)
```
poncovka: Please, change it to `progress_message(N_("Setting up %s addon") % v.name)`

pyanaconda/addons.py
jcpunk: Fixing now.

pyanaconda/addons.py
jcpunk: Fixing now.

jkonecny12:
Tested, everything works as expected.

PR #2070, opened 2019-07-31
liveinst: Add support for plain squashfs root filesystem

bcl:
livemedia-creator-31.9-1 can now create a live iso where the
LiveOS/squashfs.img is the root filesystem instead of wrapping an ext4
filesystem inside of it.

dracut's dmsquash-live module already knows how to handle this and mount
an overlayfs on top of it.

After booting the base filesystem will be mounted at /run/rootfsbase,
this examines that mountpoint and sets the LIVE_BLOCK device, the device
to use as the source for the live install, to the mounted loop device.

### DIFF
--- a/data/liveinst/liveinst
+++ b/data/liveinst/liveinst
@@ -39,6 +39,8 @@ if [ -z "$LIVE_BLOCK" ]; then
        LIVE_BLOCK="/dev/mapper/live-base"
     elif [ -b "/dev/mapper/live-osimg-min" ]; then
        LIVE_BLOCK="/dev/mapper/live-osimg-min"
+    elif [ -e "/run/rootfsbase" ]; then
+        LIVE_BLOCK="$(findmnt -n -o SOURCE /run/rootfsbase)"
     fi
 fi
 
--- a/pyanaconda/modules/payload/live/live_os.py
+++ b/pyanaconda/modules/payload/live/live_os.py
@@ -24,6 +24,7 @@ from pyanaconda.dbus import DBus
 
 from pyanaconda.core.signal import Signal
 from pyanaconda.core.constants import INSTALL_TREE
+from pyanaconda.core.util import execWithCapture
 
 from pyanaconda.modules.common.constants.objects import LIVE_OS_HANDLER
 from pyanaconda.modules.payload.handler_base import PayloadHandlerBase
@@ -84,6 +85,16 @@ class LiveOSHandlerModule(PayloadHandlerBase):
                     log.debug("Detected live base image %s", block_device)
                     return block_device
             except FileNotFoundError:
+                pass
+
+        # Is it a squashfs+overlayfs base image?
+        if os.path.exists("/run/rootfsbase"):
+            try:
+                block_device = execWithCapture("findmnt", ["-n", "-o", "SOURCE", "/run/rootfsbase"]).strip()
+                if block_device:
+                    log.debug("Detected live base image %s", block_device)
+                    return block_device
+            except (OSError, FileNotFoundError):
                 pass
 
         log.debug("No live base image detected")


jkonecny12:
Will this be a replacement for the existing Fedora Workstation Live DVD solution?
Do you have links for a bug or issue where this was discussed, if yes, could you please add them to the commit message?

From my curiosity is the benefit that you will have free space based on free RAM and not fixed as it is now?

poncovka:
There are more places in Anaconda that will have to be changed. Search `/dev/mapper/live-base`. And what about the mount point `/run/initramfs/live`? Does is exist in this setup?

Is there a live ISO that we could use for testing?

bcl:
> Will this be a replacement for the existing Fedora Workstation Live DVD solution?

That's up to Fedora, I'm just adding the ability to switch if they want to.

The lorax bug for this is here - https://bugzilla.redhat.com/show_bug.cgi?id=1717728

In my testing it worked just fine with only this change. But I didn't look any deeper.

The created image ends up being slightly smaller (about 44M), I didn't look closely at CPU usage but it didn't appear any worse.

You can make a live iso using livemedia-creator and the included fedora-livemedia.ks example, just pass --squashfs-only to the normal commandline (http://weldr.io/lorax/livemedia-creator.html#quickstart).

cmurf:
@jkonecny12 
I think it's up to Fedora Release Engineering. But I think it's appropriate to coordinate with livecd-tools folks and Fedora Spins, because they'll need to be aware of any changes involving rw persistence, e.g. SoaS use case depends on not being installed but directly using the Fedora LiveOS with rw persistence on a USB stick. The current device-mapper based overlay doesn't tell the true amount of free space whether memory or file backed, and blows up spectacularly when the fixed ext4 volume's free space is busted. It's a big benefit for that to go away by using an overlayfs based overlay instead.

bcl:
> The current device-mapper based overlay doesn't tell the true amount of free space whether memory or file backed

Right, I forgot to mention I tested that as well. Much nicer experience with overlayfs, you just get an out of space error.

jkonecny12:
Could you please @bcl extend this changes here:

https://github.com/rhinstaller/anaconda/blob/master/pyanaconda/modules/payload/live/live_os.py#L78

It's our temporary solution which will be solved other way in the future but it would be nice to have it consistent on both places.

poncovka:
Jenkins, test this please.

jkonecny12:
I've tested that and everything seems fine. Could you @bcl please rebase it on a newly created branch `f31-devel`. Sorry for the inconveniences. This PR is ready for merge.

jkonecny12:
Looking forward to see the new Live solution. Hope that Workstation will adapt it soon :)

jkonecny12:
Ahh nice, I was able to just switch branch. :)

cmurf:
I tested this by out of band modification: make plain squashfs image, zstd compression, using xorriso to make a new ISO keeping prior Fedora LiveOS bits while substituting the squashfs.img payload; and then applying the changes in this PR once the LiveOS booted. And I successfully installed the system, and it reboots. Pretty cool!

jkonecny12:
I've ran the tests locally to avoid the issue with gpg keys and everything is fine -> merging.

@cmurf thanks for testing. I'm looking forward to see this on Live Workstation :)

cmurf:
Question: does the change affect only LiveOS? e.g. Fedora-Workstation-Live-x86_64-Rawhide-20190820.n.0.iso
Or does it also affect netinstall + DVD? e.g. Fedora-Server-dvd-x86_64-Rawhide-20190820.n.0.iso, and Fedora-Server-netinst-x86_64-Rawhide-20190820.n.0.iso

The benefit on Live way bigger of course, because of the constant random reads during installation, but it might be nice in Fedora to standardize on a single kind of rootfs and overlay method for troubleshooting purposes.

bcl:
> Question: does the change affect only LiveOS?

At this time, yes. I think we should experiment with this in the spins before switching the boot.iso, get the bugs worked out, and then we can change. FWIW lorax already supports --squashfs-only, it just isn't the default.

PR #2073, opened 2019-08-02
Master move realm handling to tasks

M4rtinK:
The first few commits add some infrastructure and the new tasks, while the last commit replaces the old code by using the new tasks & there is also a bunch of tests covering the new code.

### DIFF
--- a/pyanaconda/installation.py
+++ b/pyanaconda/installation.py
@@ -27,6 +27,7 @@ from pyanaconda.modules.common.constants.objects import BOOTLOADER, SNAPSHOT, FI
 from pyanaconda.modules.common.constants.services import STORAGE, USERS, SERVICES, NETWORK, SECURITY, \
     LOCALIZATION
 from pyanaconda.modules.common.task import sync_run_task
+from pyanaconda.modules.common.structures.requirement import Requirement
 from pyanaconda.modules.storage.snapshot.create import SnapshotCreateTask
 from pyanaconda.storage.kickstart import update_storage_ksdata
 from pyanaconda.storage.installation import turn_on_filesystems, write_storage_configuration
@@ -167,10 +168,10 @@ def _prepare_configuration(storage, payload, ksdata):
 
     configuration_queue.append(generate_initramfs)
 
-    # join a realm (if required)
-    join_realm = TaskQueue("Realm join", N_("Joining realm"))
-    join_realm.append(Task("Join a realm", ksdata.realm.execute))
-    configuration_queue.append(join_realm)
+    # realm join
+    # - this can run only after network is configured in the target system chroot
+    task_proxy = SECURITY.get_proxy(security_proxy.JoinRealmWithTask())
+    configuration_queue.append(Task(task_proxy.Name, sync_run_task, (task_proxy,)))
 
     post_scripts = TaskQueue("Post installation scripts", N_("Running post-installation scripts"))
     post_scripts.append(Task("Run post installation scripts", runPostScripts, (ksdata.scripts,)))
@@ -286,14 +287,9 @@ def _prepare_installation(storage, payload, ksdata):
     pre_install_scripts.append(Task("Run %pre-install scripts", runPreInstallScripts, (ksdata.scripts,)))
     installation_queue.append(pre_install_scripts)
 
-    # Do packaging.
-
-    # Discover information about realms to join to determine the need for additional packages.
-    realm_discover = TaskQueue("Realm discover", N_("Discovering realm to join"))
-    realm_discover.append(Task("Discover realm to join", ksdata.realm.setup))
-    installation_queue.append(realm_discover)
-
-    # Check for other possibly needed additional packages.
+    # Do various pre-installation tasks
+    # - try to discover a realm (if any)
+    # - check for possibly needed additional packages.
     pre_install = TaskQueue("Pre install tasks", N_("Running pre-installation tasks"))
     pre_install.append(Task("Setup authselect", ksdata.authselect.setup))
     pre_install.append(Task("Setup firewall", ksdata.firewall.setup))
@@ -308,13 +304,17 @@ def _prepare_installation(storage, payload, ksdata):
         # only when the task is actually started, not at task creation time
         pre_install.append(WriteResolvConfTask("Copy resolv.conf to sysroot"))
 
+    # realm discovery
+    security_proxy = SECURITY.get_proxy()
+    task_proxy = SECURITY.get_proxy(security_proxy.DiscoverRealmWithTask())
+    pre_install.append(Task(task_proxy.Name, sync_run_task, (task_proxy,)))
+
     def run_pre_install():
         """This means to gather what additional packages (if any) are needed & executing payload.pre_install()."""
         # anaconda requires storage packages in order to make sure the target
         # system is bootable and configurable, and some other packages in order
         # to finish setting up the system.
         payload.requirements.add_packages(storage.packages, reason="storage")
-        payload.requirements.add_packages(ksdata.realm.packages, reason="realm")
         payload.requirements.add_packages(ksdata.authselect.packages, reason="authselect")
         payload.requirements.add_packages(ksdata.firewall.packages, reason="firewall")
         payload.requirements.add_packages(ksdata.network.packages, reason="network")
@@ -327,6 +327,17 @@ def _prepare_installation(storage, payload, ksdata):
 
         payload.requirements.add_groups(payload.language_groups(), reason="language groups")
         payload.requirements.add_packages(payload.langpacks(), reason="langpacks", strong=False)
+
+        # add package requirements from modules
+        # - iterate over all modules we know have valid package requirements
+        # - add any requirements found to the payload requirement tracking
+        modules_with_package_requirements = [SECURITY]
+        for module in modules_with_package_requirements:
+            module_proxy = module.get_proxy()
+            module_requirements = Requirement.from_structure_list(module_proxy.CollectRequirements())
+            log.debug("Adding requirements for module %s : %s", module, module_requirements)
+            payload.requirements.add_requirements(module_requirements)
+
         payload.pre_install()
 
     pre_install.append(Task("Find additional packages & run pre_install()", run_pre_install))
--- a/pyanaconda/modules/common/structures/realm.py
+++ b/pyanaconda/modules/common/structures/realm.py
@@ -29,6 +29,8 @@ class RealmData(DBusData):
         self._name = ""
         self._discover_options = list()
         self._join_options = list()
+        self._discovered = False
+        self._required_packages = []
 
     @property
     def name(self) -> Str:
@@ -71,3 +73,26 @@ class RealmData(DBusData):
     @join_options.setter
     def join_options(self, options: List[Str]):
         self._join_options = options
+
+    @property
+    def discovered(self) -> Bool:
+        """Reports if a realm has been successfully discovered."""
+        return self._discovered
+
+    @discovered.setter
+    def discovered(self, discovered : Bool):
+        self._discovered = discovered
+
+    @property
+    def required_packages(self) -> List[Str]:
+        """Packages required for joining a realm.
+
+        For example: ['realmd']
+
+        :return: a list of required packages
+        """
+        return self._required_packages
+
+    @required_packages.setter
+    def required_packages(self, packages: List[Str]):
+        self._required_packages = packages
--- a/pyanaconda/modules/security/installation.py
+++ b/pyanaconda/modules/security/installation.py
@@ -16,6 +16,9 @@
 # Red Hat, Inc.
 #
 import os
+import copy
+
+from pyanaconda.core import util
 
 from pyanaconda.simpleconfig import SimpleConfigFile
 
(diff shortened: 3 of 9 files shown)


M4rtinK:
@AdamWill Hi! :) I'm afraid we don't really have anything at hand to test realm discovery/join works correctly with this change, but IIRC OpenQA has something - would it be possible to do a run on this PR, to check I have not inadvertently broken something? 

Thanks in advance! :)

And let me know if something (updates image ?) else is needed to test this. :)

AdamWill:
Thanks for the headsup, I should be able to do it...I'm working on something else ATM, though, I'll try it later today if I can.

M4rtinK:
> Thanks for the headsup, I should be able to do it...I'm working on something else ATM, though, I'll try it later today if I can.

Thanks! :)

M4rtinK:
Updated with a few fixes & pylint fixes.

pyanaconda/installation.py
```diff
@@ -327,6 +325,23 @@ def run_pre_install():
...
+        modules_with_package_requirements = [SECURITY]
+        for module in modules_with_package_requirements:
+            module_proxy = module.get_proxy()
+            module_requirements = Requirement.from_structure_list(module_proxy.CollectRequirements())
+            for requirement in module_requirements:
+                # check requirement type and add a payload requirement appropriately
+                if requirement.type == "package":
+                    payload.requirements.add_packages([requirement.name], reason=requirement.reason)
+                elif requirement.type == "group":
+                    payload.requirements.add_groups([requirement.name], reason=requirement.reason)
```
poncovka: I recommend to implement a new method `add_requirements(requirements)` in the class `PayloadRequirements` and move this code there.

pyanaconda/modules/security/security.py
```diff
@@ -176,15 +180,64 @@ def set_realm(self, realm):
...
         :returns: list of object paths of installation tasks
         """
-        tasks = [
-            ConfigureSELinuxTask(sysroot=conf.target.system_root, selinux_mode=self.selinux)
-        ]
+        tasks = [ConfigureSELinuxTask(sysroot=conf.target.system_root, selinux_mode=self.selinux),
+                 RealmJoinTask(sysroot=conf.target.system_root,
+                               realm_discovered=self.realm.realm_discovered,
+                               realm_name=self.realm.name,
+                               join_options=self.realm.join_options)]
```
poncovka: It looks like `RealmJoinTask` could get just the system root and  `self.realm`.

pyanaconda/modules/security/security.py
```diff
@@ -176,15 +180,64 @@ def set_realm(self, realm):
...
+        for name in self._realm_required_packages:
+            requirements.append(Requirement.for_package(name, reason="realm join"))
+
+        return requirements
+
+    def setup_with_tasks(self):
+        """Return the setup tasks for this module."""
+        realm_task = RealmDiscoverTask(sysroot=conf.target.system_root,
+                                      realm_name=self.realm.name,
+                                      discover_options=self.realm.discover_options)
```
poncovka: It looks like `RealmDiscoverTask` could get just the system root and `self.realm`.

pyanaconda/modules/common/structures/realm.py
```diff
@@ -71,3 +72,12 @@ def join_options(self) -> List[Str]:
     @join_options.setter
     def join_options(self, options: List[Str]):
         self._join_options = options
+
+    @property
+    def realm_discovered(self) -> Bool:
```
poncovka: I would rename to  `discovered`. It is clear, that it is related to a realm.

pyanaconda/modules/security/security.py
```diff
@@ -52,6 +54,8 @@ def __init__(self):
         self.realm_changed = Signal()
         self._realm = RealmData()
 
+        self._realm_required_packages = []
```
poncovka: I would keep packages in the realm data as well. Please, create a new `packages` property.

pyanaconda/modules/security/security.py
```diff
@@ -176,15 +180,64 @@ def set_realm(self, realm):
...
+
+        return requirements
+
+    def setup_with_tasks(self):
+        """Return the setup tasks for this module."""
+        realm_task = RealmDiscoverTask(sysroot=conf.target.system_root,
+                                      realm_name=self.realm.name,
+                                      discover_options=self.realm.discover_options)
+
+        realm_task.succeeded_signal.connect(lambda: self.handle_realm_discover_results(realm_task.get_result()))
```
poncovka: In this case, I would just update the data structure for realm in the task. You don't need anything else for keeping the results.

pyanaconda/modules/common/base/base_interface.py
```diff
@@ -123,6 +123,14 @@ def CollectRequirements(self) -> List[Structure]:
         """
         return Requirement.to_structure_list(self.implementation.collect_requirements())
 
+    def SetupWithTasks(self) -> List[ObjPath]:
+        """Returns setup tasks of this module.
```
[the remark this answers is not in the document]
M4rtinK: Yeah, I'll just use a specific temporary API & we can do this properly once we got task introspection going.

pyanaconda/modules/security/security.py
M4rtinK: Yep, this should work out of the box.

pyanaconda/modules/security/security.py
M4rtinK: Yep, this should work as long as we update the realm data when realm_changed signal is triggered. :)

M4rtinK:
Updated the PR with a bunch of feedback to @poncovka suggestions & even more tests. :)

pyanaconda/modules/security/security.py
M4rtinK: Good point, added. :)

pyanaconda/modules/common/structures/realm.py
M4rtinK: Yep, renamed. :)

pyanaconda/installation.py
M4rtinK: Good point! Added. :)

AdamWill:
Sorry, I'm having trouble testing this via openQA as Rawhide is too broken ATM :( selinux issues are a problem, and freeipa-server is not installable due to changes in systemd. I'll try and get it cleaned up...

pyanaconda/payload/requirement.py
```diff
@@ -118,6 +118,20 @@ def add_groups(self, group_ids, reason, strong=True):
...
+        """Add requirements from a list of Requirement instances.
+
+        :param requirements: list of Requirement instances
+        """
+        for requirement in requirements:
+            # check requirement type and add a payload requirement appropriately
+            if requirement.type == "package":
+                self.add_packages([requirement.name], reason=requirement.reason)
+            elif requirement.type == "group":
+                self.add_groups([requirement.name], reason=requirement.reason)
```
poncovka: You can use the enum `PayloadRequirementType` and the method `_add` to simplify this code if you want to. However, this implementation is good enough for now.

M4rtinK: I think its fine for now. :)

pyanaconda/modules/security/security.py
```diff
@@ -176,15 +179,51 @@ def set_realm(self, realm):
...
+
+        return requirements
+
+    def discover_realm_with_task(self):
+        """Return the setup task for discovering a realm."""
+        # We do a deep copy of the realm data holder to avoid
+        # changes to the data in the Task from changing the backing
+        # data structure in the module, without triggering the changed signal.
+        realm_task = RealmDiscoverTask(sysroot=conf.target.system_root,
+                                      realm_data=copy.deepcopy(self.realm))
```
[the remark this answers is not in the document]
M4rtinK: Well, this is also about the realm data potentially changing after the `RealmDiscoverTask` has been instantiated. By doing a deepcopy outright we avoid both data changes comming by reference from the module to task and the other way around.

pyanaconda/modules/security/security.py
```diff
@@ -176,15 +179,51 @@ def set_realm(self, realm):
...
+    def collect_requirements(self):
+        """Return installation requirements for this module.
+
+        :return: a list of requirements
+        """
+        requirements = []
+
+        # Add realm requirements.
+        for name in self.realm.required_packages:
+            requirements.append(Requirement.for_package(name, reason="needed to join a realm"))
```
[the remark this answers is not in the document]
M4rtinK: I think a more descriptive reason is good & should be more widely used.

The main use case for the reason field is people reading logs to find out why some packages have been automatically added to their system, so it the reson should IMHO be easy to read & descriptive.

AdamWill:
openQA test came up [no good](https://openqa.stg.fedoraproject.org/tests/582808). Problem seems to be this:

    var/log/anaconda/syslog:04:20:34,861 WARNING org.fedoraproject.Anaconda.Modules.Security:INFO:program:realm: Couldn't join realm: This computer's host name is not set correctly.
    var/log/anaconda/syslog:04:20:34,862 WARNING org.fedoraproject.Anaconda.Modules.Security:INFO:anaconda.modules.security.installation:Joining realm ipa001.domain.local failed

pyanaconda/modules/security/security.py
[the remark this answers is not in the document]
poncovka: You can create the copy in the `__init__` method of the task.

pyanaconda/installation.py
```diff
@@ -167,11 +168,6 @@ def _prepare_configuration(storage, payload, ksdata):
 
     configuration_queue.append(generate_initramfs)
 
-    # join a realm (if required)
-    join_realm = TaskQueue("Realm join", N_("Joining realm"))
```
poncovka: The task for joining the realm needs to run after setting up the network, but the tasks of the Security module are running before that. I guess that the best solution is to create a new DBus method for this task for now. Please, document this requirement in the docstring of `DiscoverRealmWithTask`.

pyanaconda/modules/security/security.py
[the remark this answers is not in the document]
poncovka: The use case you are talking about can be solved on the module level. You can easily log why you add the specific packages and groups to the list of requirements. I don't understand why we should propagate this detailed information on DBus.

I prefer to use just a name of a component or a tool that requires it (as it was before), so the requirements can be easily processed by a third party. For example, when you want to remove all requirements for `realm`.

We should discuss how the attribute `reason` should look like. One of the solutions could be to extend the DBus structure for the requirement to be able to specify the name of the component and the detailed description.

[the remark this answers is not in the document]
M4rtinK: > We should discuss how the attribute `reason` should look like. One of the solutions could be to extend the DBus structure for the requirement to be able to specify the name of the component and the detailed description.

This sounds like an ideal solution to me.

pyanaconda/modules/security/security.py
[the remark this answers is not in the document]
M4rtinK: Good point, that indeed looks nice! :) Changed.

pyanaconda/installation.py
M4rtinK: Indeed! I've split out the task geter to a separate & temporary DBUS method and moved its invocation to the original place where the kickstart execute method has been previously. That should hopefully fix things. :)

M4rtinK:
> openQA kickstart join test came up [no good](https://openqa.stg.fedoraproject.org/tests/582808). Problem seems to be this:
> 
> ```
> var/log/anaconda/syslog:04:20:34,861 WARNING org.fedoraproject.Anaconda.Modules.Security:INFO:program:realm: Couldn't join realm: This computer's host name is not set correctly.
> var/log/anaconda/syslog:04:20:34,862 WARNING org.fedoraproject.Anaconda.Modules.Security:INFO:anaconda.modules.security.installation:Joining realm ipa001.domain.local failed
> ```

@AdamWill Thanks! Should be hopefully fixed now, so please retest with the updated PR. :)

AdamWill:
OK, will do.

AdamWill:
sorry, some other shiny things appeared in front of me first...will try it again soon.

AdamWill:
openQA test with current version of the PR [passed](https://openqa.stg.fedoraproject.org/tests/58432), yay.

M4rtinK:
@AdamWill Nice! Thanks a lot! :)

pyanaconda/modules/security/installation.py
```diff
@@ -72,3 +77,119 @@ def run(self):
...
+            for line in lines:
+                parts = line.split(":", 1)
+                if len(parts) == 2 and parts[0].strip() == "required-package":
+                    package_spec = parts[1].strip()
+                    # "" is not a valid package specification
+                    if package_spec:
+                        required_packages.append(package_spec)
+
+            log.info("Realm %s needs packages %s",
+                      discovered_realm_data, ", ".join(required_packages))
```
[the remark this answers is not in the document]
M4rtinK: Good point, split the parsing off. :)

tests/nosetests/pyanaconda_tests/module_security_test.py
```diff
@@ -154,3 +165,598 @@ def realm_kickstart_test(self):
...
+
+        publisher.assert_called()
+
+        # realm discover
+        object_path = publisher.call_args_list[0][0][0]
+        obj = publisher.call_args_list[0][0][1]
+
+        self.assertEqual(realm_discover_task_path, object_path)
+        self.assertIsInstance(obj, TaskInterface)
+        self.assertIsInstance(obj.implementation, RealmDiscoverTask)
```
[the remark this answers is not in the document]
M4rtinK: Yep, this is indeed handy for the `*WithTask()` methods that return a single task, but I think we will need to extend this wot cover also the `*WithTasks()` methods that return a list of task paths.

M4rtinK:
jenkins, test this please

tests/nosetests/pyanaconda_tests/module_security_test.py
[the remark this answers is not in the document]
jkonecny12: Yep, it would be handy.

PR #2075, opened 2019-08-05
Add support for localization of modules

rvykydal:
The pull request is still modifying os.enviorn which is non-thread-safe.
For how we do it currently look for os.environ assignments in https://github.com/rhinstaller/anaconda/blob/master/pyanaconda/localization.py and https://github.com/rhinstaller/anaconda/blob/master/pyanaconda/ui/gui/spokes/welcome.py
Hopefully the PR is not doing something more dangerous than our current code does.

### DIFF
--- a/pyanaconda/localization.py
+++ b/pyanaconda/localization.py
@@ -28,6 +28,7 @@ from collections import namedtuple
 
 from pyanaconda.core import constants, util
 from pyanaconda.core.util import upcase_first_letter, setenv, execWithRedirect
+from pyanaconda.modules.common.constants.services import BOSS
 
 from pyanaconda.anaconda_loggers import get_module_logger
 log = get_module_logger(__name__)
@@ -277,7 +278,16 @@ def setup_locale(locale, localization_proxy=None, text_mode=False):
         setenv("LANG", locale)
         locale_mod.setlocale(locale_mod.LC_ALL, locale)
 
+    set_modules_locale(locale)
+
     return locale
+
+
+def set_modules_locale(locale):
+    """Set locale of all modules."""
+    boss_proxy = BOSS.get_proxy()
+    boss_proxy.SetLocale(locale)
+
 
 def get_english_name(locale):
     """
--- a/pyanaconda/modules/boss/boss.py
+++ b/pyanaconda/modules/boss/boss.py
@@ -132,3 +132,12 @@ class Boss(MainModule):
         task = self._install_manager.install_system_with_task()
         path = self.publish_task(BOSS.namespace, task)
         return path
+
+    def set_locale(self, locale):
+        """Set locale of boss and all modules.
+
+        :param str locale: locale to set
+        """
+        log.info("Setting locale of all modules to %s.", locale)
+        super().set_locale(locale)
+        self._module_manager.set_modules_locale(locale)
--- a/pyanaconda/modules/boss/boss_interface.py
+++ b/pyanaconda/modules/boss/boss_interface.py
@@ -84,3 +84,10 @@ class AnacondaBossInterface(BossInterface):
             "line_number": get_variant(Int, result["line_number"]),
             "error_message": get_variant(Str, result["error_message"])
         } for result in results]
+
+    def SetLocale(self, locale: Str):
+        """Set locale of boss and all modules.
+
+        Examples: "cs_CZ.UTF-8", "fr_FR"
+        """
+        self.implementation.set_locale(locale)
--- a/pyanaconda/modules/common/base/base.py
+++ b/pyanaconda/modules/common/base/base.py
@@ -18,10 +18,13 @@
 # License and may only be used or replicated with the express permission of
 # Red Hat, Inc.
 #
+import os
 from abc import ABC
+from locale import setlocale, LC_ALL
 
 from pyanaconda.core.event_loop import EventLoop
 from pyanaconda.core.timer import Timer
+from pyanaconda.core.util import setenv
 from pyanaconda.dbus import DBus
 from pyanaconda.modules.common.task import publish_task, TaskInterface
 from pyanaconda.core.signal import Signal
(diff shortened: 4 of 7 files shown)


pyanaconda/localization.py
```diff
@@ -277,8 +280,22 @@ def setup_locale(locale, localization_proxy=None, text_mode=False):
         setenv("LANG", locale)
         locale_mod.setlocale(locale_mod.LC_ALL, locale)
 
+    set_modules_locale(locale)
+
     return locale
 
+
+def set_modules_locale(locale):
```
poncovka: Please, add a docstring.

pyanaconda/localization.py
```diff
@@ -277,8 +280,22 @@ def setup_locale(locale, localization_proxy=None, text_mode=False):
...
+    log.debug("setting modules locale to: %s", locale)
+    boss_proxy = BOSS.get_proxy()
+    modules = boss_proxy.GetModules()
+    for service_name in modules:
+        namespace = get_namespace_from_name(service_name)
+        object_path = get_dbus_path(*namespace)
+        module_proxy = DBus.get_proxy(service_name, object_path)
+        module_proxy.SetLocale(locale)
```
poncovka: We will probably need to set up the locale in the boss module as well, so this code might be moved to its method `set_locale`. Then you can call here just `boss_proxy.SetLocale(locale)`.

pyanaconda/modules/common/base/base_interface.py
```diff
@@ -123,6 +123,22 @@ def CollectRequirements(self) -> List[Structure]:
         """
         return Requirement.to_structure_list(self.implementation.collect_requirements())
 
+    def SetLocale(self, locale: Str):
```
poncovka: Please, add a unit test for this method.

pyanaconda/modules/boss/boss_interface.py
```diff
@@ -84,3 +84,10 @@ def DistributeKickstart(self) -> List[Dict[Str, Variant]]:
             "line_number": get_variant(Int, result["line_number"]),
             "error_message": get_variant(Str, result["error_message"])
         } for result in results]
+
+    def GetModules(self) -> List[Str]:
```
poncovka: Please, add a unit test for this method.

pyanaconda/localization.py
rvykydal: Done

pyanaconda/localization.py
rvykydal: Done

pyanaconda/modules/boss/boss_interface.py
rvykydal: I removed the method

rvykydal:
TODO: I still need to add test for SetLocale module method. I am not sure where and how to test it.

pyanaconda/modules/boss/boss.py
```diff
@@ -132,3 +132,12 @@ def install_system_with_task(self):
...
         return path
+
+    def set_locale(self, locale):
+        """Set locale of all modules.
+
+        :param str locale: locale to set
+        """
+        log.info("Setting locale of all modules to %s.", locale)
+        for observer in self._module_manager.module_observers:
+            observer.proxy.SetLocale(locale)
```
poncovka: The boss should always check whether the service is available before accessing the proxy.

pyanaconda/modules/boss/boss.py
```diff
@@ -132,3 +132,12 @@ def install_system_with_task(self):
         task = self._install_manager.install_system_with_task()
         path = self.publish_task(BOSS.namespace, task)
         return path
+
+    def set_locale(self, locale):
+        """Set locale of all modules.
+
+        :param str locale: locale to set
+        """
+        log.info("Setting locale of all modules to %s.", locale)
```
poncovka: Shouldn't the boss also call `super.set_locale()` to set up its own locale?

pyanaconda/modules/boss/boss_interface.py
```diff
@@ -84,3 +84,7 @@ def DistributeKickstart(self) -> List[Dict[Str, Variant]]:
             "line_number": get_variant(Int, result["line_number"]),
             "error_message": get_variant(Str, result["error_message"])
         } for result in results]
+
+    def SetLocale(self, locale: Str):
+        """Set locale of all modules."""
```
poncovka: Please, document an example of the locale.

pyanaconda/localization.py
```diff
@@ -28,6 +28,9 @@
 
 from pyanaconda.core import constants, util
 from pyanaconda.core.util import upcase_first_letter, setenv, execWithRedirect
+from pyanaconda.modules.common.constants.services import BOSS
+from pyanaconda.dbus.namespace import get_namespace_from_name, get_dbus_path
```
poncovka: This import can be removed, I guess.

pyanaconda/modules/common/base/base.py
```diff
@@ -133,6 +136,26 @@ def collect_requirements(self):
         """
         return []
 
+    def set_locale(self, locale):
```
poncovka: This method should be probably implemented in the class MainModule, so the boss can use it as well.

pyanaconda/modules/boss/boss.py
rvykydal: Done (moved to module_manager)

pyanaconda/modules/boss/boss.py
rvykydal: Done

pyanaconda/modules/boss/boss_interface.py
rvykydal: Added.

pyanaconda/localization.py
rvykydal: Removed.

pyanaconda/modules/common/base/base.py
rvykydal: Done.

pyanaconda/modules/boss/module_manager.py
```diff
@@ -107,6 +107,17 @@ def check_modules_availability(self):
...
 
+    def set_modules_locale(self, locale):
+        """Set locale of all modules.
+
+        :param str locale: locale to set
+        """
+        log.info("Setting locale of all modules to %s.", locale)
+        for observer in self.module_observers:
+            if not observer.is_service_available:
+                log.warning("%s is not available when setting locale", observer)
```
poncovka: You are missing `continue` here.

rvykydal: Shame on me and thank you for this endless review, I think I should focus on a single thing at a time.

PR #2079, opened 2019-08-08
Change the documentation of the 'reason' attribute

poncovka:
Update the documentation of the 'reason' attribute in the Requirement
class. The attribute should provide a message that explains why this
requirement is required.

Provide explanations for requirements of the Storage module.

### DIFF
--- a/pyanaconda/modules/common/structures/requirement.py
+++ b/pyanaconda/modules/common/structures/requirement.py
@@ -65,9 +65,15 @@ class Requirement(DBusData):
 
     @property
     def reason(self) -> Str:
-        """A reason of the requirement.
+        """An explanation of the requirement.
 
-        Usually a name of a component that requires this requirement.
+        Provides a message that explains why this requirement
+        is required.
+
+        For example:
+            Required to join a realm.
+            Requested by the firewall kickstart command.
+            Necessary for network team device configuration.
 
         :return: a reason
         """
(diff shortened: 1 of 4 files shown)


pyanaconda/modules/common/structures/requirement.py
```diff
@@ -65,9 +65,15 @@ def name(self, value: Str):
...
     def reason(self) -> Str:
-        """A reason of the requirement.
+        """An explanation of the requirement.
 
-        Usually a name of a component that requires this requirement.
+        Provides a message that explains why this requirement
+        is required.
+
+        For example:
+            Required to join the realm.
```
M4rtinK: I would recommend: "Required to join a realm."

pyanaconda/modules/common/structures/requirement.py
```diff
@@ -65,9 +65,15 @@ def name(self, value: Str):
...
+            Requested by the firewall kickstart command.
+            Necessary for the team device configuration.
```
M4rtinK: I would recommend: "Necessary for network team device configuration."

pyanaconda/modules/common/structures/requirement.py
poncovka: Fixed.

pyanaconda/modules/common/structures/requirement.py
poncovka: Fixed.

PR #2080, opened 2019-08-08
Add a DBus structure for the partitioning request 

poncovka:
Use a DBus structure for the partitioning request in the automatic
partitioning module and UI.

Move the code from pyanaconda.storage.partitioning to the task
for the automatic partitioning.


pyanaconda/modules/common/structures/partitioning.py
```diff
@@ -18,10 +18,232 @@
...
+        """
+        return self._partitioning_scheme
+
+    @partitioning_scheme.setter
+    def partitioning_scheme(self, scheme: Int):
+        self._partitioning_scheme = scheme
+
+    @property
+    def file_system_type(self) -> Str:
+        """Type of a file system used on the partitions.
```
M4rtinK: I wonder if it would make sense to have these specified by an enum/constnats or some other way where one could easily check what are the valid values ?

pyanaconda/modules/common/structures/partitioning.py
```diff
@@ -18,10 +18,232 @@
...
+        """
+        return self._passphrase
+
+    @passphrase.setter
+    def passphrase(self, value: Str):
+        self._passphrase = value
+
+    @property
+    def cipher(self) -> Str:
+        """Encryption algorithm used to encrypt the filesystem.
```
M4rtinK: Again I wonder if we want to specify somehow what are the valid values - maybe even by pointing to docs ?

pyanaconda/modules/common/structures/partitioning.py
```diff
@@ -18,10 +18,232 @@
...
+        self._file_system_type = value
+
+    @property
+    def excluded_mount_points(self) -> List[Str]:
+        """Excluded mount points.
+
+        Don't create partitions for the excluded
+        mount points during the partitioning.
+
+        For example: /home, /boot, swap
```
M4rtinK: I think it would be good to point to some document about what mountpoints are supported (/<something> seems logical, but do we have other special cases other than `swap` ?)

pyanaconda/modules/common/structures/partitioning.py
```diff
@@ -18,10 +18,232 @@
...
+        """
+        return self._cipher
+
+    @cipher.setter
+    def cipher(self, cipher: Str):
+        self._cipher = cipher
+
+    @property
+    def luks_version(self) -> Str:
+        """Version of LUKS.
```
M4rtinK: do we have something else than luks1 & luks2 ? I guess list of valid values could be specified.

pyanaconda/modules/common/structures/partitioning.py
```diff
@@ -18,10 +18,232 @@
...
+        self._luks_version = version
+
+    @property
+    def pbkdf(self) -> Str:
+        """The PBKDF algorithm.
+
+        Set Password-Based Key Derivation Function (PBKDF)
+        algorithm for LUKS keyslot.
+
+        Example: 'argon2i'
```
M4rtinK: What are the other supported values & what happens if unknown value (or "") is supplied ?

pyanaconda/modules/common/structures/partitioning.py
```diff
@@ -18,10 +18,232 @@
...
+        self._pbkdf_iterations = iterations
+
+    @property
+    def escrow_certificate(self) -> Str:
+        """URL of an X.509 certificate.
+
+        Store the data encryption keys of all encrypted volumes created during
+        installation, encrypted using the certificate, as files in /root.
+
+        :return: URL of an X.509 certificate
```
M4rtinK: is this a file:// URL or regular URL, eq.:

`file://foo.cert`

or

`/foo.cert`

pyanaconda/modules/common/structures/partitioning.py
poncovka: No, we don't.

pyanaconda/modules/common/structures/partitioning.py
poncovka: Well, Blivet will fail at some point if you specify an invalid value. We cannot validate these values in the DBus structure, because that would require to import objects from Blivet.

pyanaconda/modules/common/structures/partitioning.py
poncovka: Well, Blivet will fail at some point if you specify an invalid value. I don't know where to find the valid values, even the kickstart documentation doesn't mention that.

pyanaconda/modules/common/structures/partitioning.py
poncovka: Fixed.

pyanaconda/modules/common/structures/partitioning.py
poncovka: I don't know.

pyanaconda/modules/common/structures/partitioning.py
poncovka: I don't know.

poncovka:
> Looks good to me overall, just seems like the documentation for the DBus structure could be extended and maybe some values possible handled via enums/constants, DBus API considerations permitting.

@M4rtinK, the extension of the documentation and improved validation of values is not a goal of this pull request. It just grabs the existing DBus API of the automatic partitioning module and transforms that into a DBus structure.

pyanaconda/modules/common/structures/partitioning.py
jkonecny12: Maybe ask @vojtechtrefny about the list of valid values?

vojtechtrefny: Unfortunately we don't have a list of valid ciphers here. We just take the string and give it to cryptsetup. List of actually supported combinations of cipher, mode and iv depends on kernel support. Some examples are available here -- https://gitlab.com/cryptsetup/cryptsetup/wikis/DMCrypt

pyanaconda/modules/storage/partitioning/automatic_partitioning.py
```diff
@@ -17,24 +17,90 @@
...
+        lv=True,
+        thin=True,
+        encrypted=True),
+    PartSpec(
+        fstype="swap",
+        grow=False,
+        lv=True,
+        encrypted=True
+    )
+]
```
jkonecny12: I guess this will change to something which can be dynamically changed by a spin later. If that is correct, could you please add a comment with explanation of our plans?

pyanaconda/modules/storage/partitioning/automatic_partitioning.py
```diff
@@ -70,40 +66,85 @@ def _configure_partitioning(self, storage):
...
+        )
+
+        return {
+            "passphrase": request.passphrase,
+            "cipher": request.cipher,
+            "luks_version": luks_version,
+            "pbkdf_args": pbkdf_args,
+            "escrow_cert": escrow_cert,
+            "add_backup_passphrase": request.backup_passphrase_enabled,
+            "min_luks_entropy": MIN_CREATE_ENTROPY,
```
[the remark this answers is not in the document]
poncovka: This data structure is required by Blivet, so I don't see the point of doing that unless Blivet provides us these constants. Anyway, this is a common way how to propagate arguments to Blivet's classes and functions and I don't really want to mess with it.

pyanaconda/modules/storage/partitioning/automatic.py
```diff
@@ -162,290 +128,38 @@ def set_enabled(self, enabled):
...
-        return {
-            "passphrase": self.passphrase,
-            "cipher": self.cipher,
-            "luks_version": luks_version,
-            "pbkdf_args": self.pbkdf_args,
-            "escrow_cert": escrow_cert,
-            "add_backup_passphrase": self.backup_passphrase_enabled,
-            "min_luks_entropy": MIN_CREATE_ENTROPY,
-        }
+        request = copy.deepcopy(self.request)
```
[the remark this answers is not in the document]
poncovka: Fixed.

pyanaconda/modules/storage/partitioning/automatic_partitioning.py
poncovka: I don't understand. We already dynamically modify this partitioning based on the platform and the kickstart data. These are just default definitions.

poncovka:
Updated.

pyanaconda/modules/storage/partitioning/automatic_partitioning.py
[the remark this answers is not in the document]
jkonecny12: I'm not saying changing the dict but using constants instead of plain strings. Pylint will find typo in the constant but no-one is able to do that for you in the plain strings.

pyanaconda/modules/storage/partitioning/automatic_partitioning.py
jkonecny12: That's exactly what I asked for. To add comment that these are the default values and how they can be overridden from the outside. However if I look little bit lower where these are used then it make it's pretty self-explanation so you can ignore this.

PR #2077, opened 2019-08-07
Replace network related setup calls in kickstart.py

M4rtinK:
Move the network related `setup()` calls in kickstart.py, that set package requirements for the main RPM transaction and move the code to the Network DBUS module.

~~**NOTE:** This PR depends on https://github.com/rhinstaller/anaconda/pull/2073, and can be only merged once #2073 is merged. This is also the reason why the very last commit is the only relevant one. :-)~~

PR #2081, opened 2019-08-09
Extend TaskQueue for dbus tasks

M4rtinK:
Extend the TaskQueue API to make adding DBus Tasks easy.

### DIFF
--- a/pyanaconda/installation.py
+++ b/pyanaconda/installation.py
@@ -26,7 +26,6 @@ from pyanaconda.core.constants import BOOTLOADER_DISABLED
 from pyanaconda.modules.common.constants.objects import BOOTLOADER, SNAPSHOT, FIREWALL
 from pyanaconda.modules.common.constants.services import STORAGE, USERS, SERVICES, NETWORK, SECURITY, \
     LOCALIZATION
-from pyanaconda.modules.common.task import sync_run_task
 from pyanaconda.modules.common.structures.requirement import Requirement
 from pyanaconda.modules.storage.snapshot.create import SnapshotCreateTask
 from pyanaconda.storage.kickstart import update_storage_ksdata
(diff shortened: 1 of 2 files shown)


M4rtinK:
NOTE: This should be merged after #2073 and #2077 are merged, so the new tasks they add can use this API as well.

pyanaconda/installation_tasks.py
```diff
@@ -321,6 +322,17 @@ def append(self, item):
         item.parent = self
         self._list.append(item)
 
+    @synchronized
+    def append_dbus_tasks(self, module, dbus_tasks):
+        """Append DBus Tasks from a module to the TaskQueue.
+
+        :param module: DBusServiceIdentifier instance corresponding to an Anaconda DBus module
```
poncovka: I would rename it to `service_id`, so it cannot be confused with a proxy of a DBus module.

M4rtinK: Sounds good, changed. :)

pyanaconda/installation.py
```diff
@@ -90,34 +89,28 @@ def _prepare_configuration(storage, payload, ksdata):
     os_config = TaskQueue("Installed system configuration", N_("Configuring installed system"))
     os_config.append(Task("Configure authselect", ksdata.authselect.execute))
 
+    # add installation tasks for the Security DBus module
```
jkonecny12: Please remove these comments. By this comment you are telling the same as you can easily read from the code. If you want to have a comment here, then merge them together to ``add installation tasks from DBus modules``.

M4rtinK: The order of the modules is sometimes important & it simply seems more readable this way.

jkonecny12: Both ways are good to me. One is just slightly better.
Still you have my ACK.

M4rtinK:
jenkins, test this please

PR #2082, opened 2019-08-09
Simplify the storage-related code in the payload

poncovka:
Propagate the storage object and the kickstart data to an instance
of the Payload class during its initialization. Then it is not
necessary to propagate these objects to this instance again.

Rename variables and arguments of methods in the Payload class that
provide a device path, so it is clear that it is not an instance of
a device.

Move the support methods for CD/DVD and ISO image installations
to the the subpackage for payload, so it is clear that it is only
related to the payload.

Extend the payload utils and use them in the payload-related code.


pyanaconda/payload/__init__.py
```diff
@@ -863,29 +863,29 @@ def _setup_install_device(self, checkmount):
         metalink = None
 
         # See if we already have stuff mounted due to dracut
-        isodev = payload_utils.get_mount_device(DRACUT_ISODIR)
-        device = payload_utils.get_mount_device(DRACUT_REPODIR)
+        iso_device_path = payload_utils.get_mount_device(DRACUT_ISODIR)
+        device_path = payload_utils.get_mount_device(DRACUT_REPODIR)
```
jkonecny12: Rename this to `repo_device_path` instead.

pyanaconda/payload/__init__.py
```diff
@@ -779,7 +779,7 @@ def _find_mounted_additional_repos(self):
     def _unmount_source_directory(self, mount_point):
```
jkonecny12: The commit comment shouldn't be ``Use the function resolve_device from payload utils`` but instead ``Add new function resolve_device to payload utils``

pyanaconda/payload/__init__.py
```diff
@@ -32,9 +32,8 @@
```
jkonecny12: Typo in commit message 13d46bd8c0347fb7ee919ed2007051d34aa42422:
``Move the support methods for CD/DVD and ISO image installations to the the subpackage for payload``

Double ``the the``.

pyanaconda/payload/__init__.py
```diff
@@ -472,7 +472,7 @@ def _setup_device(device, mountpoint):
             device.format.setup(mountpoint=mountpoint)
```
jkonecny12: Again please rename the commit messages from ``Use the function...`` to ``Add a new function...``

pyanaconda/payload/dnfpayload.py
```diff
@@ -291,9 +291,9 @@ def do_transaction(base, queue_instance):
 
 
 class DNFPayload(payload.PackagePayload):
-    def __init__(self, data):
-        super().__init__(data)
 
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
```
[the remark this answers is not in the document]
poncovka: It is more difficult to change arguments, when subclasses don't use `*args` and `**kwargs`. In this case, the arguments are going to slowly disappear and the `Payload` class will be eventually removed, so I think that we can use this solution for now.

pyanaconda/payload/utils.py
```diff
@@ -28,37 +28,63 @@
...
 
 
-def mount(url, mountpoint, fstype, options):
+def mount(device_path, mount_point, fstype, options):
+    """Mount a filesystem.
+
+    :param device_path: a device path
+    :param mount_point: a mount point
+    :param fstype: a filesystem type
+    :param options: a string of mount options
```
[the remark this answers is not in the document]
poncovka: Fixed.

pyanaconda/payload/__init__.py
```diff
@@ -457,13 +457,13 @@ def _setup_device(device, mountpoint):
...
         # instance.
         real_mountpoint = os.path.realpath(mountpoint)
+        mount_device_path = payload_utils.get_mount_device(real_mountpoint)
 
-        mdev = payload_utils.get_mount_device(real_mountpoint)
-        if mdev:
-            if mdev:
-                log.warning("%s is already mounted on %s", mdev, mountpoint)
+        if mount_device_path:
+            if mount_device_path:
```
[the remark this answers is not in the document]
poncovka: Fixed.

pyanaconda/payload/utils.py
```diff
@@ -28,37 +28,63 @@
 log = get_module_logger(__name__)
 
 
-def get_mount_device(mountpoint):
-    if os.path.ismount(mountpoint):
-        return blivet.util.get_mount_device(mountpoint)
+def get_mount_device(mount_point):
```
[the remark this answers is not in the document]
poncovka: Fixed.

pyanaconda/payload/__init__.py
poncovka: Fixed.

pyanaconda/payload/__init__.py
poncovka: Fixed.

pyanaconda/payload/__init__.py
poncovka: Fixed.

pyanaconda/payload/__init__.py
poncovka: Fixed.

poncovka:
Updated.

pyanaconda/payload/dnfpayload.py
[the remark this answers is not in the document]
jkonecny12: OK

PR #2083, opened 2019-08-09
Don't check the format status in UI

poncovka:
Don't check the format status to find out whether the device is
mounted. Use the result of the function `get_mount_paths`.


poncovka:
@vojtechtrefny, can you have a look at this pull request, please?

PR #2085, opened 2019-08-12
Fix missing requests timeout

jkonecny12:
Adding timeouts to `session.get` calls and increasing timeout value based on the requests upstream documentation.

See:
https://3.python-requests.org/user/advanced/#timeouts

This unfortunately also change also three timeouts in the network.py which is a side effect but I'm not convinced we want to create a new constant for just 1 second difference.

### DIFF
--- a/pyanaconda/core/constants.py
+++ b/pyanaconda/core/constants.py
@@ -103,7 +103,11 @@ TEXT_ONLY_TARGET = 'multi-user.target'
 GRAPHICAL_TARGET = 'graphical.target'
 
 # Network
-NETWORK_CONNECTION_TIMEOUT = 45  # in seconds
+
+# Requests package (where this constant is used) recommends to have timeout slightly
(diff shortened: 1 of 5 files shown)


pyanaconda/core/constants.py
```diff
@@ -103,7 +103,7 @@
 GRAPHICAL_TARGET = 'graphical.target'
 
 # Network
-NETWORK_CONNECTION_TIMEOUT = 45  # in seconds
+NETWORK_CONNECTION_TIMEOUT = 46  # in seconds
```
poncovka: The explanation should be documented in a comment above this constant in case that we will need to modify it later.

jkonecny12: Good point.

jkonecny12:
**UPDATED:**
* Implemented @poncovka note.

PR #2086, opened 2019-08-12
Don't measure code coverage during the installation

poncovka:
Remove the support for measuring the code coverage during the
installation. The results don't include DBus modules, only UI,
so they are not so useful anymore. We focus on the coverage of
the unit tests instead.


poncovka:
Fixed.

hroncok:
Thanks!

PR #2087, opened 2019-08-13
Collect addons after running the %pre section

poncovka:
Don't collect addons until we run the `%pre` section, so the addons
can be generated in this section. Simplify the code for processing
a kickstart file and run specific steps in `anaconda.py`.

### DIFF
--- a/anaconda.py
+++ b/anaconda.py
@@ -268,8 +268,6 @@ if __name__ == "__main__":
     # init threading before Gtk can do anything and before we start using threads
     from pyanaconda.threading import AnacondaThread, threadMgr
     from pyanaconda.core.i18n import _
-
-    from pyanaconda.addons import collect_addon_paths
     from pyanaconda.core import util, constants
     from pyanaconda import startup_utils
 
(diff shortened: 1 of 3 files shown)


anaconda.py
```diff
@@ -455,13 +453,21 @@ def _earlyExceptionHandler(ty, value, traceback):
...
+    from pyanaconda.addons import collect_addon_paths
     addon_paths = collect_addon_paths(constants.ADDON_PATHS)
 
-    # If we were given a kickstart file on the command line, parse (but do not
-    # execute) that now.  Otherwise, load in defaults from kickstart files
-    # shipped with the installation media.
-    ksdata = startup_utils.parse_kickstart(opts, addon_paths, pass_to_boss=True)
+    # Parse the kickstart file.
+    ksdata = kickstart.AnacondaKSHandler(addon_paths["ks"])
+    kickstart.parseKickstart(ksdata, kspath, strict_mode=opts.ksstrict, pass_to_boss=True)
```
jkonecny12: Could you please make a function from this part. It will be much readable to see that these parts are connected.

poncovka:
@jkonecny12, this pull request also simplifies the code for processing a kickstart file, so it will be easier to use our DBus solution in the future.

jkonecny12:
> @jkonecny12, this pull request also simplifies the code for processing a kickstart file, so it will be easier to use our DBus solution in the future.

OK, in that case there is a good reason. I'm in! :)

PR #2091, opened 2019-08-15
Rename the icon (#1740864)

poncovka:
The new icon provided by fedora-logos has a different name.
We should use org.fedoraproject.AnacondaInstaller now.

Resolves: rhbz#1740864


jkonecny12:
Thanks for solving this!

PR #2094, opened 2019-08-21
Rename the icon (#1740864)

poncovka:
**Merged to master at https://github.com/rhinstaller/anaconda/pull/2091, but it should be also part of F31.**

The new icon provided by fedora-logos has a different name.
We should use org.fedoraproject.AnacondaInstaller now.

Resolves: rhbz#1740864


poncovka:
Jenkins, test this please.

PR #2095, opened 2019-08-21
Handle the missing mountpoint attribute (#1743853)

poncovka:
The functions generate_device_info and validate_device_info shouldn't
fail if the device format doesn't have the mountpoint attribute.
Use None instead.

Don't require to specify a mount point for swap.

Resolves: rhbz#1743853


poncovka:
Jenkins, test this please.

PR #2096, opened 2019-08-21
Flatpak support for rpmostree payload

jkonecny12:
This will add logic to install all the flatpak apps and runtimes from the local repository on stage2 path `/flatpak/repo`.

This change set will add flatpak "payload" which contains logic to install flatpaks after rpmostree installation. Also I have to change how installation tasks for rpmostreepayload are registered so the installed flatpaks won't be erased by the prepare mount points installation task in the rpmostree payload.

This solution will be used as a base for creating a proper payload after rpmostree will be a module.

Possible improvements:
* Better error reporting and reaction 
    * Ask user if we can skip the failing flatpak?
* Better progress
    * Report user progress of the current flatpak. Right now it only tells you what flatpak is installed but runtimes could have plenty of gigs so it could take some time before user see a change.

*Resolves: rhbz#1734970*


pyanaconda/payload/flatpak.py
```diff
@@ -0,0 +1,353 @@
...
+        :param operation: object describing the operation
+        :type operation: Flatpak.TransactionOperation instance
+        :param error: object containing error description
+        :type error: GLib.Error instance
+        :param details: information if the error was fatal
+        :type details: int value of Flatpak.TransactionErrorDetails
+        """
+        self._log_operation(operation, "failed")
+        log.error("Flatpak installation failed with message: '%s' -- error is fatal %s",
+                  error.message, self._error_fatal)
```
rvykydal: Is there _error_fatal attribute in the instance?

pyanaconda/payload/rpmostreepayload.py
```diff
@@ -53,6 +54,10 @@ def __init__(self, *args, **kwargs):
         self._internal_mounts = []
         self._locale_map = None
 
+        self._flatpak_payload = FlatpakPayload(conf.target.system_root)
+        # Initialize temporal repo to enable reading of the remote
+        self._flatpak_payload.initialize_with_path("/var/tmp/anaconda-flatpak-temp")
```
rvykydal: Is it safe to have FlatpackPayload (even initialized) in RPMOSTreePayload by default? I mean from the perspective of use cases without flatpacks - now running code (and hitting potential bugs) which is not necessary to run.
But maybe we don't care at this point.
And shouldn't we check if the payload is available before trying to initialize it?

jkonecny12:
jenkins, test this please

jkonecny12:
Jenkins, it's ok to test.

pyanaconda/payload/rpmostreepayload.py
jkonecny12: Good point maybe it would make sense to promote the `is_available` as a static method.

jkonecny12:
**UPDATED:**
* Fixed pylint errors

pyanaconda/payload/flatpak.py
jkonecny12: A valid bug found also by pylint. I should have run the pylint before posting PR.

jkonecny12:
**UPDATED:**
* Switch `is_available()` method to a class method
* Avoid creating of the flatpak object when repository is not available.

pyanaconda/payload/flatpak.py
jkonecny12: Fixed.

pyanaconda/payload/rpmostreepayload.py
jkonecny12: Fixed by checking is available before creation.

jkonecny12:
Hi, @cgwalters this is a heads-up if you agree with the work done here.

pyanaconda/payload/rpmostreepayload.py
```diff
@@ -53,6 +54,14 @@ def __init__(self, *args, **kwargs):
         self._internal_mounts = []
         self._locale_map = None
 
+        # flatpak_payload will be None if it can't be used in this environment (don't have repo)
+        self._flatpak_payload = None
+
+        if FlatpakPayload.is_available():
```
poncovka: There could be a new type of the `RPMOSTreePayload` with the flatpack support always enabled, so you don't have to check whether `self._flatpak_payload` is None all the time.

jkonecny12: Great solution. Working on that!

pyanaconda/payload/flatpak.py
```diff
@@ -0,0 +1,354 @@
...
+                  error.message, details == TransactionErrorDetails.FATAL)
+
+    @staticmethod
+    def _log_operation(operation, state):
+        operation_type_str = TransactionOperationType.to_string(operation.get_operation_type())
+        log.debug("Flatpak operation: %s of ref %s state %s",
+                  operation_type_str, operation.get_ref(), state)
+
+
+class BaseRefsList(ABC):
```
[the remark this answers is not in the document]
jkonecny12: Yeah I know but I'm using official names for that.

jkonecny12:
**UPDATED:**
* Split flatpaks logic to a separate rpmostree payload (thanks for suggestion @poncovka!)

pyanaconda/payload/flatpak.py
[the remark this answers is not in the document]
poncovka: That is ok, but it should be explained in the docstring.

pyanaconda/installation.py
```diff
@@ -331,14 +331,11 @@ def run_pre_install():
     installation_queue.append(payload_install)
 
     # for some payloads storage is configured after the payload is installed
-    if not payload.needs_storage_configuration:
+    if not payload.needs_storage_configuration and not conf.target.is_directory:
```
cgwalters: This was separated in this way for a reason I think...maybe using lorax/LMC to create live rpm-ostree? I forget.  May not be relevant anymore.  `git annotate` would probably know though.

jkonecny12: Ups you are correct. We would probably broke the --dirinstall by calling the part of this code there.
Thanks a lot for your idea!

jkonecny12: Also there is a question, would dirinstall without this code be usable. The code creates /var and similar mountpoints so I can't imagine system usable without this. Also is dirinstall even useful for ostree?

What do you think @cgwalters ?

jkonecny12: So or so behavior is the same as before so I'm merging this. We can fix this later.

jkonecny12: I was able to find out when the check for dir install was removed and it's this commit:
664ef7b43f9102aa9332d0db5b7d13f8ece436f0

It was removed by Brian and the reason was migration to bind mounts. So this seems as a targeted change which improved the situation. In that case I'm taking this as works as expected.

PR #2097, opened 2019-08-22
Add a simple unit test for TUI and GUI

poncovka:
Test collecting hubs and spokes of UI interfaces.


poncovka:
The spokes are sorted by their names now.

PR #2101, opened 2019-08-26
Set the flag lvm_metadata_backup (#1673901)

rvykydal:
The flag lvm_metadata_backup should be set to False for image
installations. We shouldn't use the Blivet's flag image_install
to check the type of the installation, because it is not set at
this point. Let's use the Anaconda's flag imageInstall instead.

Related: rhbz#1715344
Resolves: rhbz#1673901


rvykydal:
The patch (cherry-picked from rhel 8.1 / rhel-8 branch) is required for the fix of the z-stream BZ.

PR #2102, opened 2019-08-26
Use a DBus container for tasks

poncovka:
Make tasks publishable, remove `publish_task` and publish them using `TaskContainer`.


poncovka:
The unit tests pass when run locally.

PR #2104, opened 2019-08-27
Use numbers for user/group names when unpacking tar live image

carlzhc:
By default when using tar to unpack a live image, it maps files' uid/gid to the installer system user's uid/gid if the user exists in current system. 
But this behavior can cause issue for unpacking a live image. For example if the anaconda environment has the user 'chrony' and the live image also contains 'chrony' but with different uid/gid as anaconda environment. After unpacking the live image, the files in live image belonging to 'chrony' user will be changed to different uid/gid. It is not correct.

This PR is addressing the issue.

### DIFF
--- a/pyanaconda/payload/livepayload.py
+++ b/pyanaconda/payload/livepayload.py
@@ -524,7 +524,7 @@ class LiveImageKSPayload(LiveImagePayload):
 
         cmd = "tar"
         # preserve: ACL's, xattrs, and SELinux context
-        args = ["--selinux", "--acls", "--xattrs", "--xattrs-include", "*",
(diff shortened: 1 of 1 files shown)


jkonecny12:
Hello, thanks a lot for your contribution. We will definitely look on that.
If you want to get this to Fedora 31 please rebase this PR on `f31-devel` branch and we will try to get this to the new Fedora. If you are ok with Rawhide then you have the correct branch.

jkonecny12:
Jenkins, it's ok to test.

rvykydal:
Note: I am also going to update the options in our WIP payload module:
https://github.com/rhinstaller/anaconda/blob/7cb173edb8b85f3a8336fb608eeaf1481696f157/pyanaconda/modules/payload/live/installation.py#L93

PR #2105, opened 2019-08-27
Don't set up a warning label to None (#1745933)

poncovka:
If the error is None, set the warning label to an empty string.

Resolves: rhbz#1745933


poncovka:
Jenkins, test this please.

PR #2108, opened 2019-08-28
Fall back to a boot drive with a valid stage1 device (#1168118)

poncovka:
When the boot drive is not specified, try to find a disk with a valid
stage1 device. Don't look for a disk with a /boot partition, because
the partition doesn't have to be a valid stage1 device.

Resolves: rhbz#1168118


poncovka:
The kickstart tests for the bootloader and storage passed.

poncovka:
Jenkins, test this please.

PR #2112, opened 2019-09-02
Compare normalized XML strings

poncovka:
Use the function compare_xml to compare two XML strings in tests.
The order of the XML attributes has changed in Python 3.8, so we
have to call the function canonicalize to normalize the strings.


poncovka:
Fixed the pylint issues.

PR #2114, opened 2019-09-03
Drop dhclient requirement

rvykydal:
With all networking configured by NetworkManager and NM now defaulting to
using it's own internal dhcp client there is no need to depend on dhclient.

Signed-off-by: Peter Robinson <pbrobinson@gmail.com>


rvykydal:
I accidentally merged the PR https://github.com/rhinstaller/anaconda/pull/2109 into master branch so I am adding the PR for f31-devel.

PR #2119, opened 2019-09-06
Don't add more devices in boot-device NVRAM than the maximum allowed

martinezjavier:
The boot-device NVRAM variable has a maximum number of devices allowed,
that is specified in /sys/firmware/devicetree/base/ibm,max-boot-devices
OF property.

Adding more devices than the maximum number allowed in the boot-device
NVRAM variable can cause issues so the device list need to be restricted
to this maximum before updating the variable with nvram --update-config.

### DIFF
--- a/pyanaconda/bootloader.py
+++ b/pyanaconda/bootloader.py
@@ -2214,6 +2214,12 @@ class IPSeriesGRUB2(GRUB2):
         # Remove all other occurances of it.
         boot_list = [boot_disk] + [x for x in boot_list if x != boot_disk]
 
+        # The boot-device NVRAM variable has a maximum number of devices allowed,
+        # don't add more than the limit in the ibm,max-boot-devices OF property.
+        with open("/sys/firmware/devicetree/base/ibm,max-boot-devices", "rb") as f:
+            limit = int.from_bytes(f.read(4), byteorder='big')
+            boot_list = boot_list[:limit]
+
         update_value = "boot-device=%s" % " ".join(boot_list)
 
         rc = util.execWithRedirect("nvram", ["--update-config", update_value])


pyanaconda/bootloader.py
```diff
@@ -2214,6 +2214,12 @@ def updateNVRAMBootList(self):
         # Remove all other occurances of it.
         boot_list = [boot_disk] + [x for x in boot_list if x != boot_disk]
 
+        # The boot-device NVRAM variable has a maximum number of devices allowed,
+        # don't add more than the limit in the ibm,max-boot-devices OF property.
+        f = open("/sys/firmware/devicetree/base/ibm,max-boot-devices", "rb")
+        limit = int.from_bytes(f.read(4), byteorder='big')
+        boot_list = boot_list[0:limit]
```
jkonecny12: Do you need the ``boot_list[0:limit]`` shouldn't it be enough to have just a ``boot_list[limit]``?

jkonecny12:
Jenkins, it's ok to test.

pyanaconda/bootloader.py
```diff
@@ -2214,6 +2214,12 @@ def updateNVRAMBootList(self):
         # Remove all other occurances of it.
         boot_list = [boot_disk] + [x for x in boot_list if x != boot_disk]
 
+        # The boot-device NVRAM variable has a maximum number of devices allowed,
+        # don't add more than the limit in the ibm,max-boot-devices OF property.
+        f = open("/sys/firmware/devicetree/base/ibm,max-boot-devices", "rb")
```
poncovka: Please, use `with` to close the file.

pyanaconda/bootloader.py
martinezjavier: @jkonecny12 I guess you meant `boot_list[:limit]`? Yes, I changed if you think that's more clear.

pyanaconda/bootloader.py
martinezjavier: @poncovka done.

martinezjavier:
@jkonecny12 @poncovka pushed a new version addressing your comments. Thanks a lot for the feedback.

pyanaconda/bootloader.py
jkonecny12: For me it's more obvious. When the `0` is there I wonder what's the reason and looking for catches.

rvykydal:
jenkins, test this please

rvykydal:
@martinezjavier 
Javier, could you add a rhbz reference to an approved BZ to the commit message? I'll merge and build then.

martinezjavier:
> @martinezjavier
> Javier, could you add a rhbz reference to an approved BZ to the commit message? I'll merge and build then.

@rvykydal done.

PR #2120, opened 2019-09-09
network tui: fix addr_str referenced before assignment (#1731415)

rvykydal:


rvykydal:
You are right, updated.

PR #2121, opened 2019-09-09
Fix the second screen of Welcome to Fedora (#1748203)

poncovka:
Don't show an outdated image in the second screen of the Welcome
to Fedora application. Use the Anaconda icon instead.

Resolves: rhbz#1748203


poncovka:
live1
live2fixed

poncovka:
The second screen without the fix:

live2broken

PR #2133, opened 2019-09-17
Extend the support for storage installation

poncovka:
Extend the DBus API to support all storage and bootloader installation tasks required by UI.

### DIFF
--- a/pyanaconda/modules/storage/bootloader/installation.py
+++ b/pyanaconda/modules/storage/bootloader/installation.py
@@ -18,6 +18,7 @@
 # Red Hat, Inc.
 #
 from blivet import arch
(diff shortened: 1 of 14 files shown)


pyanaconda/modules/storage/bootloader/installation.py
```diff
@@ -95,16 +97,63 @@ def run(self):
         install_boot_loader(storage=self._storage)
 
 
+class FixBTRFSBootloaderTask(Task):
+    """Installation task fixing the bootloader on BTFS.
```
jkonecny12: You're missing `R` here.

poncovka: Fixed.

PR #2137, opened 2019-09-20
Add the install class for Red Hat Virtualization (#1753316)

poncovka:
The default partitioning scheme can be defined in an install class.
If the kickstart file defines the scheme, it will be used instead.

(based on a commit 106ffe9)

Red Hat Virtualization has some extra requirements on partitioning
that need to be checked.

(based on a commit a5a8aca)

Backport the install class from RHEL 7.

Resolves: rhbz#1753316

### DIFF
--- a/pyanaconda/installclasses/rhv.py
+++ b/pyanaconda/installclasses/rhv.py
@@ -1,0 +1,129 @@
+#
+# rhv.py
+#
+# Copyright (C) 2019  Red Hat, Inc.  All rights reserved.
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+
+from blivet.devicefactory import DEVICE_TYPE_LVM_THINP
+from blivet.size import Size
+from pykickstart.constants import AUTOPART_TYPE_LVM_THINP
+
+from pyanaconda import network
+from pyanaconda import nm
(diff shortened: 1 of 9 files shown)


pyanaconda/installclasses/rhv.py
```diff
@@ -0,0 +1,132 @@
+#
+# rhv.py
+#
+# Copyright (C) 2016  Red Hat, Inc.  All rights reserved.
```
jkonecny12: Please fix the year.

pyanaconda/installclasses/rhv.py
```diff
@@ -0,0 +1,132 @@
...
+    hidden = not productName.startswith("oVirt")
+
+    efi_dir = "centos"
+    default_autopart_type = AUTOPART_TYPE_LVM_THINP
+
+    # there is a RHV branded help content variant
+    help_folder = "/usr/share/anaconda/help/rhv"
+
+    def configure(self, anaconda):
+        BaseInstallClass.configure(self, anaconda)
```
jkonecny12: Is this really required. Wouldn't it be called automatically if not specified?

pyanaconda/storage_utils.py
```diff
@@ -242,6 +251,12 @@ def verify_partition_sizes(storage, constraints, report_error, report_warning):
                            % {'mount': mount, 'size': size,
                               'productName': productName})
 
+    for (mount, size) in constraints[STORAGE_REQ_PARTITION_SIZES].items():
+        if mount in filesystems and filesystems[mount].size < size:
+            report_error(_("Your %(mount)s partition is less than "
+                           "%(size)s which is lower than required.")
+                         % {'mount': mount, 'size': size})
```
jkonecny12: This sentence sounds weird. Could it be rephrased in something like this?
`Your %(mount)s partition size is lower than required %(size)s.`

pyanaconda/installclasses/rhv.py
poncovka: Fixed.

pyanaconda/installclasses/rhv.py
poncovka: Removed.

pyanaconda/storage_utils.py
poncovka: This message is based on the warning above.

poncovka: Fixed.

jkonecny12: Looks great now. Thanks a lot!

poncovka:
Backported to upstream in https://github.com/rhinstaller/anaconda/pull/2432.

PR #2141, opened 2019-09-23
Remove pointless '../../' to clean up NFS mounts

jcpunk:
NFS mounts on CentOS 8 get confused when `../../../` takes you below the mounted NFS volume.  The repo base is being mounted by anaconda, but the pointed to repo is well below the repo base.

For non-NFS users this will result in cleaner looking URLs.


jkonecny12:
Jenkins, it's ok to test.

PR #2142, opened 2019-09-25
Verify block sizes of VGs on top of PVs (#1754446)

poncovka:
LVM no longer allows creating VGs on top of PVs with different block
sizes, so report a storage validation error in this case.

Resolves: rhbz#1754446

### DIFF
--- a/pyanaconda/storage_utils.py
+++ b/pyanaconda/storage_utils.py
@@ -519,6 +519,34 @@ def verify_mounted_partitions(storage, constraints, report_error, report_warning
                                "installation. Please unmount it and retry.") % part.path)
 
 
+def verify_lvm_block_sizes(storage, constraints, report_error, report_warning):
+    """ Verify block sizes of VGs on top of PVs.
+
+    LVM no longer allows creating VGs on top of PVs with different block sizes.
+
+    :param storage: a storage to check
+    :param constraints: a dictionary of constraints
+    :param report_error: a function for error reporting
+    :param report_warning: a function for warning reporting
+    """
+    for vg in storage.vgs:
+        if vg.exists:
+            continue
+
+        disks = set(pv.disk for pv in vg.pvs)
+        sector_sizes = set(d.format.sector_size for d in disks)
+
+        if len(sector_sizes) > 1:
+            report_error(_(
+                "The volume group {name} cannot be created. Disks "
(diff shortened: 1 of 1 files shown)


poncovka:
@vojtechtrefny, can you have a look at this pull request, please?

poncovka:
sector_sizes

dwlehman:
We are going to have to have code to do this check in blivet as well, so maybe anaconda should only contain the UI part. @vojtechtrefny or am I missing something?

pyanaconda/storage_utils.py
```diff
@@ -519,6 +519,34 @@ def verify_mounted_partitions(storage, constraints, report_error, report_warning
...
+    :param storage: a storage to check
+    :param constraints: a dictionary of constraints
+    :param report_error: a function for error reporting
+    :param report_warning: a function for warning reporting
+    """
+    for vg in storage.vgs:
+        if vg.exists:
+            continue
+
+        disks = set(pv.disk for pv in vg.pvs)
```
dwlehman: You should be able to use `vg.disks` for this.

pyanaconda/storage_utils.py
```diff
@@ -519,6 +519,34 @@ def verify_mounted_partitions(storage, constraints, report_error, report_warning
...
+        sector_sizes = set(d.format.sector_size for d in disks)
```
dwlehman: Since its possible for the PVs to be unpartitioned disks you should check that the format type is 'disklabel' before accessing the `sector_size` attribute.

vojtechtrefny:
> We are going to have to have code to do this check in blivet as well, so maybe anaconda should only contain the UI part. @vojtechtrefny or am I missing something?

Our check will be in `_pre_create` which is too late for Anaconda. We can't check this sooner because you can have preexisting VGs created with older version of LVM and checking in the `LVMVolumeGroupDevice` constructor would break reset/populate.

pyanaconda/storage_utils.py
vojtechtrefny: I didn't realize this. Looks like we need to add a `sector_size` property for `DiskDevice`.

dwlehman:
I thought we could add the check to `LVMVolumeGroupDevice._add_parent` where we hopefully know whether or not the VG exists.

poncovka:
This issue will be handled by Blivet. Closing this pull request.

PR #2144, opened 2019-09-25
Run the manual partitioning task for the given requests

poncovka:
Provide the list of requests as an argument of the task.
Don't retrieve them in the task using DBus.

### DIFF
--- a/pyanaconda/modules/storage/partitioning/manual_partitioning.py
+++ b/pyanaconda/modules/storage/partitioning/manual_partitioning.py
@@ -20,9 +20,6 @@ from pykickstart.errors import KickstartParseError
 
 from pyanaconda.anaconda_loggers import get_module_logger
 from pyanaconda.core.i18n import _
-from pyanaconda.modules.common.constants.objects import MANUAL_PARTITIONING
-from pyanaconda.modules.common.constants.services import STORAGE
-from pyanaconda.modules.common.structures.partitioning import MountPointRequest
 from pyanaconda.modules.storage.partitioning.noninteractive_partitioning import \
     NonInteractivePartitioningTask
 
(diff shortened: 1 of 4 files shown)


pyanaconda/modules/storage/partitioning/manual_partitioning.py
```diff
@@ -34,19 +32,25 @@
...
     """A task for the manual partitioning configuration."""
 
+    def __init__(self, storage, requests):
+        """Create a task.
+
+        :param storage: an instance of Blivet
+        :param requests: a list of requests
+        """
+        super().__init__(storage)
+        self._requests = list(requests)
```
jkonecny12: If it should be list (based on the doc string) why are you making a list from the `requests` variable?

pyanaconda/modules/storage/partitioning/manual_partitioning.py
```diff
@@ -34,19 +32,25 @@
...
-        manual_part_proxy = STORAGE.get_proxy(MANUAL_PARTITIONING)
-
-        # Set up mount points.
-        for mount_data in manual_part_proxy.Requests:
-            self._setup_mount_point(storage, MountPointRequest.from_structure(mount_data))
+        for mount_data in self._requests:
+            self._setup_mount_point(storage, mount_data)
 
-    def _setup_mount_point(self, storage, mount_data):
+    def _setup_mount_point(self, storage, mount_data: MountPointRequest):
```
jkonecny12: It's already in the documentation, what is the reason to specify it here for a protected method?

pyanaconda/modules/storage/partitioning/manual_partitioning.py
poncovka: Fixed.

pyanaconda/modules/storage/partitioning/manual_partitioning.py
poncovka: Fixed.

poncovka:
Updated.

PR #2147, opened 2019-09-25
Improve updates repo configuration in GUI (#1670471)

M4rtinK:
Instead of the previous "do not install updates" checkbox use a
radio button group with install updates/do not install updates options.
This also gets rid of the very long line of text that was next to the
old checkbox.

Also fix a two UI element glitches:
- update configuration elements being sensitive for certain sources
  that don't support them
- HTTP URL options checkbox not being available for closest mirror after
  switching to a non-HTTP remote installation source and back

Resolves: rhbz#1670471

This is how it looked before:
old_no_updates_check_box

This is how it looks now:
new_updates_radio_button

### DIFF
--- a/pyanaconda/ui/gui/spokes/installation_source.py
+++ b/pyanaconda/ui/gui/spokes/installation_source.py
@@ -709,7 +709,7 @@ class SourceSpoke(NormalSpoke, GUISpokeInputCheckHandler):
         self._url_type_combo_box = self.builder.get_object("urlTypeComboBox")
         self._url_type_label = self.builder.get_object("urlTypeLabel")
 
-        self._no_updates_checkbox = self.builder.get_object("noUpdatesCheckbox")
+        self._updates_radio_button = self.builder.get_object("updatesRadioButton")
 
         self._verify_iso_button = self.builder.get_object("verifyIsoButton")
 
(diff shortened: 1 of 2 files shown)


M4rtinK:
@kparal @AdamWill Just a heads up that this will likely need OpenQA adjustments once it reaches Rawhide.

AdamWill:
thanks.

pyanaconda/ui/gui/spokes/installation_source.py
```diff
@@ -709,7 +709,8 @@ def _grab_objects(self):
         self._url_type_combo_box = self.builder.get_object("urlTypeComboBox")
         self._url_type_label = self.builder.get_object("urlTypeLabel")
 
-        self._no_updates_checkbox = self.builder.get_object("noUpdatesCheckbox")
+        self._updates_radio_button = self.builder.get_object("updatesRadioButton")
+        self._no_updates_radio_button = self.builder.get_object("noUpdatesRadioButton")
```
poncovka: It looks like this attribute is not used.

M4rtinK: Indeed! Fixed.

PR #2152, opened 2019-09-26
Change the bases of the install classes for RHVH

poncovka:
The install classes should be based on the install classes for
Red Hat Enterprise Linux and CentOS Linux to get correct default
values.

They provide the same implementation of setNetworkOnbootDefault,
so this code can be removed.

Related: rhbz#1753316


poncovka:
Backported to upstream in https://github.com/rhinstaller/anaconda/pull/2432.

PR #2154, opened 2019-09-27
Don't touch storage until it is ready

poncovka:
UI should never do anything with the storage until all long-term
tasks are finished, because the access to all Blivet objects is
locked with one global lock, so we might block the main thread.


pyanaconda/ui/gui/spokes/blivet_gui.py
```diff
@@ -141,6 +139,7 @@ def refresh(self):
 
         self._back_already_clicked = False
 
+        config.default_fstype = self._storage.default_fstype
```
poncovka: @vojtechtrefny, can I move this line here?

vojtechtrefny: Yes, this should be save. blivet-gui uses this value in the add and edit dialogs so it's not picked up before running `refresh`.

PR #2155, opened 2019-10-01
Rhel 8 ks from hdd with network and url

rvykydal:


rvykydal:
Ported to master by https://github.com/rhinstaller/anaconda/pull/2184 (rather different fix due to NM in initramfs).

PR #2157, opened 2019-10-03
Switch to pypi pylint from RPM

jkonecny12:
~The reason to do this is that Fedora `pylint` have again some problems with the newer version on the other hand this works fine on pypi.~

~I also don't prefer that Fedora is packaging some random commits instead of the released versions on Rawhide.~

---------------
_EDIT:_
Fedora RPM version of astroid has a problem with dependency auto generation which works correctly in the pypi repository. This is a work-around for the issue.

See: rhbz#1758141


jkonecny12:
**UPDATED**
The commit message to explain why this is happening.

PR #2158, opened 2019-10-03
Support basic nested DBus structures

poncovka:
Move the code for generating DBus fields to the class `DBusFieldFactory`,
so it will be easier to extend it.

Use the function `is_base_type` to check the base of a type hint and
the function `get_type_arguments` to get type arguments of a type hint.

Add support for object representation of basic nested DBus structures.
The attributes of the class DBusData can be instances of DBusData and
lists of instances of DBusData.

### DIFF
--- a/pyanaconda/dbus/structure.py
+++ b/pyanaconda/dbus/structure.py
@@ -118,7 +118,7 @@ class DBusData(ABC):
         super().__init_subclass__(*args, **kwargs)
 
         # Generate the DBus fields from the members of the class cls.
-        setattr(cls, DBUS_FIELDS_ATTRIBUTE, generate_fields(cls))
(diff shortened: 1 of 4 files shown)


poncovka:
Jenkins, test this please.

pyanaconda/dbus/structure.py
```diff
@@ -197,53 +197,96 @@ def get_fields(obj):
     return fields
 
 
-def generate_fields(cls):
-    """Generate DBus fields from properties of a class.
+class DBusFieldFactory(object):
+    """A DBus filed factory."""
```
rvykydal: filed -> field

poncovka:
Fixed a typing error.

PR #2161, opened 2019-10-04
Skip version when processing kernel boot args (#1637472)

jkonecny12:
Without this patch booting with version on kernel boot cmdline will just print anaconda version and quit. This is not expected behavior.

(based on commit ef48dab58bdc5ca388aa4a181a6215bc4f22394f)

*Resolves: rhbz#1637472*


jstodola:
Could you also skip the "help" argument, please?

jkonecny12:
**UPDATED:**
* Disable also a `help` parameter. Thanks @jstodola for pointing that out.

rvykydal:
Ported to master with https://github.com/rhinstaller/anaconda/pull/2169

PR #2165, opened 2019-10-07
Fix a unit test for a boot loader task

poncovka:
Change the order of the patched tasks and call `assert_called_once_with`
instead of `called_once_with` on them.


poncovka:
The test is failing, because the kernel module btrfs couldn't be loaded.

jkonecny12:
jenkins, test this please

poncovka:
Jenkins, test this please.

PR #2166, opened 2019-10-07
Port the makeupdates script to Python 3

M4rtinK:
This actually was rather easy, as we have also decided to drop the long unused code for Fetching packages from Koji based on requirement changes in Anaconda spec file. The removed code would likely require the most work, even just due to using urllib. :)

With the code removed it was basically just switching the shebang & some minor cleanups to make the code look a bit nicer. :)


jkonecny12:
Please fix the pylint errors first.

M4rtinK:
Pylint issues fixed (actually identified some more dead code indirectly :) ).

PR #2167, opened 2019-10-08
Remove ldconfig spec macros

jkonecny12:
This macros are not required from Fedora 28 (base of RHEL-8):

https://fedoraproject.org/wiki/Changes/Removing_ldconfig_scriptlets

The macros will be called on transaction and not for every package.

This will solve the bug when anaconda-widgets package raised an error when it was removed or updated on S390x or aarch64 architectures.

*Resolves: rhbz#1700685*


jkonecny12:
Fedora PR: https://github.com/rhinstaller/anaconda/pull/2168

PR #2169, opened 2019-10-08
Skip buildins when processing kernel boot args (#1637472)

jkonecny12:
Skip version and help from a kernel parameters.

Without this patch booting with version or help on kernel boot cmdline will just print anaconda version/help and quit. This is not the expected behavior.

(based on commit ef48dab58bdc5ca388aa4a181a6215bc4f22394f)

Resolves: rhbz#1637472


jkonecny12:
Backport PR: https://github.com/rhinstaller/anaconda/pull/2161

PR #2170, opened 2019-10-09
Make the nomount boot option work again

VladimirSlavik:
Previously, handling of the `inst.nomount` boot option was somehow removed. As a consequence, the command did nothing. This adds back logic to make the option do something. As a result, users can again reject mounting storage in rescue mode from the boot editing line. (Was a kickstart-only capability requiring editing the kickstart file.)

### DIFF
--- a/pyanaconda/rescue.py
+++ b/pyanaconda/rescue.py
@@ -169,7 +169,7 @@ class Rescue(object):
                       <pyanaconda.kickstart.AnacondaKSScript>
 
     """
-    def __init__(self, storage, rescue_data=None, reboot=False, scripts=None):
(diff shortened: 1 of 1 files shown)


pyanaconda/rescue.py
```diff
@@ -184,7 +184,7 @@ def __init__(self, storage, rescue_data=None, reboot=False, scripts=None):
 
         if rescue_data:
             self.automated = True
-            self.mount = not rescue_data.nomount
+            self.mount = (not rescue_data.nomount) and (not rescue_nomount)
```
jkonecny12:
```suggestion
            self.mount = not (rescue_data.nomount or rescue_nomount)
```

VladimirSlavik:
Thank you, merging then...

PR #2171, opened 2019-10-10
Fix regexp for iscsi initiator iqn name validation (#1750865)

rvykydal:


rvykydal:
I proposed the patch for F31 final freeze exception.

PR #2172, opened 2019-10-10
Add PARTUUID to lsblk_output log

japokorn:
Modified lsblk_output log creation so it includes partition UUID.

UUID collisions are commonly reported issue. It isusually caused by cloning partitions/disks, so it is not something we can fix.
Since blivet works with partition UUID which is not included in `lsblk_output` log, this change
should improve understanding what is actually going on.

### DIFF
--- a/pyanaconda/exception.py
+++ b/pyanaconda/exception.py
@@ -321,7 +321,9 @@ def initExceptionHandling(anaconda):
 def lsblk_callback():
     """Callback to get info about block devices."""
 
-    return util.execWithCapture("lsblk", ["--perms", "--fs", "--bytes"])
+    options = "NAME,SIZE,OWNER,GROUP,MODE,FSTYPE,LABEL,UUID,PARTUUID,MOUNTPOINT"
+
+    return util.execWithCapture("lsblk", ["--bytes", "-o", options])
 
 
 def nmcli_dev_list_callback():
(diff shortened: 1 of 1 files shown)


poncovka:
Jenkins, test this please.

VladimirSlavik:
Jenkins, please test this.

pyanaconda/exception.py
```diff
@@ -321,7 +321,9 @@ def initExceptionHandling(anaconda):
 def lsblk_callback():
     """Callback to get info about block devices."""
 
-    return util.execWithCapture("lsblk", ["--perms", "--fs", "--bytes"])
+    options = "NAME,SIZE,OWNER,GROUP,MODE,FSTYPE,LABEL,UUID,PARTUUID,MOUNTPOINT"
```
VladimirSlavik: Hi, as I'm comparing the outputs, it lacks the columns FSAVAIL and FSUSE% that are in the output of the previous lsblk incantation. They are apparently added by the `--fs` option, even if manpage says nothing about these.

```suggestion
    options = "NAME,SIZE,OWNER,GROUP,MODE,FSTYPE,LABEL,UUID,PARTUUID,FSAVAIL,FSUSE%,MOUNTPOINT"
```

VladimirSlavik:
See also https://github.com/karelzak/util-linux/pull/884 for fixing the likely review nitpick origin.

VladimirSlavik:
Works as intended.

PR #2174, opened 2019-10-11
Do not use module_debsolv_errors DNF method

jkonecny12:
We should instead use `module_depsolv_errors`.


jkonecny12:
**UPDATED:**
* Resolve not that related conflict.

PR #2179, opened 2019-10-15
Add support for the ZIPL Secure Boot (#1659400)

poncovka:
Use the DBus property `ZIPLSecureBoot` of the Bootloader module to set up the ZIPL
Secure Boot on s390x. The property can be set with the kickstart command `zipl` and
one of the options `--secure-boot`, `--no-secure-boot` and `--force-secure-boot`.

Allow to set up the ZIPL Secure Boot on s390x in the dialog window for selected
disks and the boot loader of the Storage spoke in GUI.

**Depends on:** https://github.com/pykickstart/pykickstart/pull/281
Resolves: rhbz#1659400

### DIFF
--- a/pyanaconda/modules/storage/bootloader/bootloader_interface.py
+++ b/pyanaconda/modules/storage/bootloader/bootloader_interface.py
@@ -22,7 +22,7 @@ from pyanaconda.dbus.property import emits_properties_changed
 from pyanaconda.dbus.typing import *  # pylint: disable=wildcard-import
 from pyanaconda.modules.common.base import KickstartModuleInterfaceTemplate
 from pyanaconda.modules.common.constants.objects import BOOTLOADER
-from pyanaconda.modules.storage.constants import BootloaderMode, BootloaderType
+from pyanaconda.modules.storage.constants import BootloaderMode, BootloaderType, ZIPLSecureBoot
 
 
 @dbus_interface(BOOTLOADER.interface_name)
@@ -43,6 +43,7 @@ class BootloaderInterface(KickstartModuleInterfaceTemplate):
         self.watch_property("KeepBootOrder", self.implementation.keep_boot_order_changed)
         self.watch_property("ExtraArguments", self.implementation.extra_arguments_changed)
         self.watch_property("Timeout", self.implementation.timeout_changed)
+        self.watch_property("ZIPLSecureBoot", self.implementation.zipl_secure_boot_changed)
         self.watch_property("IsPasswordSet", self.implementation.password_is_set_changed)
 
     @property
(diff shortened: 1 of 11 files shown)


poncovka:
secure_boot_window

pyanaconda/modules/storage/bootloader/bootloader_interface.py
```diff
@@ -232,3 +234,22 @@ def IsPasswordSet(self) -> Bool:
...
+        return self.implementation.secure_boot.value
+
+    @emits_properties_changed
+    def SetSecureBoot(self, value: Int):
+        """Set up the Secure Boot mode.
+
+        Supported values:
+           -1  Default.
+            0  Disable Secure Boot.
+            1  Enable Secure Boot unconditionally.
```
M4rtinK: What happens if Secure Boot is unconditionally enabled yet hardware does not support it ? I guess the installation will fail ? If so, that should be documented somewhere.

pyanaconda/ui/gui/spokes/lib/cart.py
```diff
@@ -217,3 +235,26 @@ def on_set_as_boot_clicked(self, button):
             self._previousID = self._store[itr][ID_COL]
 
         self._toggle_button_text(self._store[itr])
+
+    def _initialize_secure_boot(self):
+        if not arch.is_s390():
+            self._secure_boot_box.hide()
```
M4rtinK: So if I understand this correctly, this is "just" a s390 variant of Secure Boot & is not adding support for "regular" UEFI Secure Boot? In such case I think this needs to be explicitly noted in the pull request that this is s390 specific and all the relevant boot options, kickstart options & GUI elements should be appropriately namespaced. 

Otherwise this could be easily confused with what most people know under Secure Boot (eq. UEFI Secure Boot) & could conflict with attempts to support UEFI Secure Boot in the future as well.

pyanaconda/modules/storage/bootloader/bootloader_interface.py
poncovka: Yes, it will fail. The docstring is updated.

pyanaconda/ui/gui/spokes/lib/cart.py
poncovka: Fixed. It is called ZIPL Secure Boot now.

poncovka:
Updated.

pyanaconda/ui/gui/spokes/lib/cart.py
```diff
@@ -217,3 +235,26 @@ def on_set_as_boot_clicked(self, button):
...
+
+    def _initialize_zipl_secure_boot(self):
+        if not arch.is_s390():
+            self._secure_boot_box.hide()
+            return
+
+        secure_boot = self._bootloader_proxy.ZIPLSecureBoot
+
+        if secure_boot == SECURE_BOOT_DEFAULT:
+            secure_boot = SECURE_BOOT_AUTO
```
jkonecny12: Not sure if the GUI should be responsible about what means `DEFAULT`. Shouldn't it be more like the `DEFAULT` is decided by the module and module will give you the "real" value?

pyanaconda/ui/gui/spokes/lib/cart.py
```diff
@@ -217,3 +235,26 @@ def on_set_as_boot_clicked(self, button):
...
+
+        for name, value in SECURE_BOOT_VALUES.items():
+            if secure_boot == value:
+                self._secure_boot_combo.set_active_id(name)
+                break
+
+    def _apply_zipl_secure_boot(self):
+        if not arch.is_s390():
+            return
```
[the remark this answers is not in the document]
poncovka: It is an interesting idea, but I would try to implement it in rawhide first.

poncovka:
Cherry-picked from https://github.com/rhinstaller/anaconda/pull/2199.

pyanaconda/ui/gui/spokes/lib/cart.py
poncovka: Fixed.

jkonecny12:
Just thinking aloud. Aren't we able to detect if we are running on IBM z14 series and disable the option automatically to avoid user confusion?

poncovka:
> Just thinking aloud. Aren't we able to detect if we are running on IBM z14 series and disable the option automatically to avoid user confusion?

@jkonecny12, it was discussed at bugzilla, but unfortunately it is not possible.

poncovka:
Jenkins, test this please.

poncovka:
Jenkins, test this please.

rvykydal:
jenkins, test this please

jkonecny12:
jenkins, test this please

jkonecny12:
jenkins, test this please

jkonecny12:
jenkins, test this please

rvykydal:
Adding "port to Fedora" for Vendy to determine if this should be ported to Fedora.

poncovka:
Backported to upstream in https://github.com/rhinstaller/anaconda/pull/2199.

PR #2181, opened 2019-10-15
Rename main modules to services

poncovka:
Rename the class MainModule, the class KickstartModule and all subclasses
of KickstartModule. Use the names Service and KickstartService instead.


pyanaconda/modules/users/users.py
```diff
@@ -36,8 +36,9 @@
 from pyanaconda.anaconda_loggers import get_module_logger
 log = get_module_logger(__name__)
 
-class UsersModule(KickstartModule):
-    """The Users module."""
+
+class UsersService(KickstartService):
+    """The Users services."""
```
M4rtinK: Shouldn't this the "The Users service" ? We have in singular everywhere else as far as I can tell.

poncovka: Fixed.

poncovka:
Updated.

PR #2183, opened 2019-10-15
Re-factorize payload module

jkonecny12:
Create two directories `payloads` and `sources`. Each source will have its own folder under `sources` directory and the same works for `payloads`.


jkonecny12:
I was just moving the code so this is tested by unit tests which should be enough in this case.

jkonecny12:
**UPDATED:**
* Fixed bad path in the packages Makefile. Fortunately we have the RPM tests now :)

jkonecny12:
**UPDATED:**
* Fix conflicts

PR #2184, opened 2019-10-17
Fix initramfs network configuration from kickstart from hd

rvykydal:


rvykydal:
First three commits are port from rhel-8 branch,
tha fourth is a modification for NetworkManager in intramfs.

rvykydal:
@jkonecny12 I've updated the messages and also the last commit based on feedback from Lubomir from NM (he suggested the detection of the NM case by existence of the file) to still support also non-NM case.

PR #2186, opened 2019-10-17
Return back AutoPart kickstart command

jkonecny12:
The `AutoPart` command was completely removed by the commit d28a76fe1b19d4d0ac98b4f89020a99117278310. The command is tested in pykickstart for collisions with other storage commands by `handler.autopart.seen` call. That works until this command was replaced by `UselessCommand` which removed the `handler.autopart` command. It does not exists because this handler mapping is mapped based on the `__class__.__name__` so instead of having the `handler.autopart` we end up with `handler.uselesscommand`.

This is a hotfix for the issue to unblock installations. The correct fix should go to pykickstart and test if handler contains `autopart` attribute before touching it.

### DIFF
--- a/pyanaconda/kickstart.py
+++ b/pyanaconda/kickstart.py
@@ -265,6 +265,12 @@ class Authselect(RemovedCommand):
         except RuntimeError as msg:
             authselect_log.error("Error running %s %s: %s", cmd, args, msg)
 
+# FIXME: We have to fix self.handler.autopart in the pykickstart project before replacing this by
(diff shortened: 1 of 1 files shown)


rvykydal:
Looks good to me as a hot fix.
Maybe I'd add a FIXME comment in the pykickstart.py so we don't forget to fix it properlyl.

jkonecny12:
**UPDATED:**
* Added fixme

M4rtinK:
Storage kickstart tests all passed or failed for unrelated reasons, so lets merge this & do a release. :)

PR #2187, opened 2019-10-17
Add pure community features documentation

jkonecny12:
Add documentation about pure community features. This will make possible to add and maintain pure community related parts which won't be maintained by the Anaconda team.

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -187,9 +187,9 @@ Below is a list of pure community features, their community maintainers, and mai
 /boot on btrfs subvolume
 ^^^^^^^^^^^^^^^^^^^^^^^^
 
-- Origin: https://github.com/rhinstaller/anaconda/pull/1375
-- Bugzilla: https://bugzilla.redhat.com/show_bug.cgi?id=1418336
-- Maintainer:
-- Description:
+* Origin: https://github.com/rhinstaller/anaconda/pull/1375
+* Bugzilla: https://bugzilla.redhat.com/show_bug.cgi?id=1418336
+* Maintainer: Neal Gompa <ngompa13@gmail.com>
+* Description:
 
 ``Enable boot of the installed system from a BTRFS subvolume.``


CONTRIBUTING.rst
```diff
@@ -174,3 +174,25 @@ Then you are done.
...
+
+Pure community features
+=======================
+
+The pure community features are features which are part of the Anaconda code base but they are maintained and extended mainly by a community. These features are not
+a priority for the Anaconda project.
+
+In case there will be an issue in the pure community features, the Anaconda team will provide only a sanity checking. It is the responsibility of the
+community (maintainers of the feature) to provide fix for the issue. If the issue will have bigger impact on other parts of the Anaconda project or if it will
+block a release or another priority feature and the fix won't be provided in a reasonable time the Anaconda team have the rights to remove or disable this feature from
```
VladimirSlavik: Less words, probably same meaning, less lawyerish. Feel free to not use.
```suggestion
block a release or another priority feature and the fix won't be provided in a reasonable time the Anaconda team can remove or disable this feature from
```

CONTRIBUTING.rst
```diff
@@ -174,3 +174,25 @@ Then you are done.
...
+the Anaconda code base.
```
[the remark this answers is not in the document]
jkonecny12: My bad I thought I'm consistent with the rest of the document but I was wrong. I'll fix that.

CONTRIBUTING.rst
jkonecny12: Not sure about this one. Seems to me like it has a different meaning.

VladimirSlavik: Ah, the meaning sure is different - I'd consider "have the rights" to hint at a quite formal relationship between those involved in a potential disagreement. In fact I would expect that kind of a word from a formal document, which this is not. (Is it?) That's why I went with the "can", which in my mind captures better the one-sided "I can because I am the BDFL here" reality. Because honestly, as a maintainer, you do have the right implicitly.

One further consideration is that features could be perhaps also disabled instead of removed, but I guess spending time on all details of all possibilities is not worth it. Also, the current text does go with the "worst case" outcome, so I see no problem there.

All things considered, I write all of this in the spirit of a documentation review: This is what I see and think. Make of it what you want. Therefore it's always a suggestion, not a demand/blocker :)

CONTRIBUTING.rst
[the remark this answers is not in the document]
VladimirSlavik: Thank you!

CONTRIBUTING.rst
```diff
@@ -174,3 +174,21 @@ Then you are done.
...
+The pure community features are features which are part of the Anaconda code base but they are maintained and extended mainly by the community. These features are not a priority for the Anaconda project.
+
+In case of issues in pure community features, the Anaconda team will provide only sanity checking. It is the responsibility of the community (maintainers of the feature) to provide fix for the issue. If the issue will have bigger impact on other parts of the Anaconda project or if it will block a release or another priority feature and the fix won't be provided in a reasonable time the Anaconda team have the rights to remove or disable this feature from the Anaconda code base.
+
+Below is a list of these features, their community maintainers, and contacts to them:
+
+/boot on btrfs subvolume
+^^^^^^^^^^^^^^^^^^^^^^^^
+
+- Origin: https://github.com/rhinstaller/anaconda/pull/1375
```
poncovka: Could we list here also the number of the bug?

jkonecny12: Good point.

CONTRIBUTING.rst
```diff
@@ -174,3 +174,22 @@ Then you are done.
 
 If the pull request has been opened for the ``f25-devel`` branch, then you also need to merge the ``f25-devel``
 branch back to ``master`` once you merge your pull request (see "Merging the Fedora devel branch back to the master branch" above).
+
+Pure community features
+-----------------------
+
+The pure community features are features which are part of the Anaconda code base but they are maintained and extended mainly by the community. These features are not a priority for the Anaconda project.
+
+In case of issues in pure community features, the Anaconda team will provide only sanity checking. It is the responsibility of the community (maintainers of the feature) to provide fix for the issue. If the issue will have bigger impact on other parts of the Anaconda project or if it will block a release or another priority feature and the fix won't be provided in a reasonable time the Anaconda team have the rights to remove or disable this feature from the Anaconda code base.
```
M4rtinK: `the Anaconda team have the rights to remove or disable this feature`
This seams a bit weird to me, I suggest using this:
`the Anaconda team reserves the right to remove or disable this feature`

CONTRIBUTING.rst
```diff
@@ -174,3 +174,22 @@ Then you are done.
...
+
+Below is a list of these features, their community maintainers, and contacts to them:
```
M4rtinK: `Below is a list of these features, their community maintainers, and contacts to them:`
I suggest:
`Below is a list of pure community features, their community maintainers, and maintainer contact information:`

CONTRIBUTING.rst

CONTRIBUTING.rst

jkonecny12:
**UPDATED:**
* Implemented notes from @M4rtinK. Thanks!

CONTRIBUTING.rst
```diff
@@ -174,3 +174,22 @@ Then you are done.
...
+In case of issues in pure community features, the Anaconda team will provide only sanity checking. It is the responsibility of the community (maintainers of the feature) to provide fix for the issue. If the issue will have bigger impact on other parts of the Anaconda project or if it will block a release or another priority feature and the fix won't be provided in a reasonable time the Anaconda team reserves the rights to remove or disable this feature from the Anaconda code base.
+
+Below is a list of pure community features, their community maintainers, and maintainers contact information:
+
+/boot on btrfs subvolume
+^^^^^^^^^^^^^^^^^^^^^^^^
+
+- Origin: https://github.com/rhinstaller/anaconda/pull/1375
+- Bugzilla: https://bugzilla.redhat.com/show_bug.cgi?id=1418336
+- Maintainer:
```
Conan-Kudo: How is this defined? GitHub user? FAS ID? A person's name?

jkonecny12: I would not insist on one specific type of contact. Let's leave everyone to decide what contact is the best to reach that person. However, at least email should be always marked here as the base ground. We can always add more contacts if needed.

So for you, I will add a new commit with:
Maintainer: ``Neal Gompa <ngompa13@gmail.com>``

Do you want to add more contacts?

------------------------

About @npmcallum I will not add him here until his reply or PR to this document will arrive. Just to be sure that he really want to be a maintainer.

Conan-Kudo: That's fine with me.

PR #2188, opened 2019-10-18
Remove network command from anaconda

rvykydal:


pyanaconda/network.py
```diff
@@ -296,6 +298,33 @@ def initialize_network():
     log.debug("Initialization finished.")
 
 
+def write_configuration(payload):
+    """Install network configuration to target system."""
+    fcoe_proxy = STORAGE.get_proxy(FCOE)
+    fcoe_nics = fcoe_proxy.GetNics()
+    fcoe_ifaces = [dev.device_name for dev in get_supported_devices()
```
jkonecny12: Just want to be sure. Is FCoE from storage module completely moved to the module from UI?

poncovka: Yes, it is.

pyanaconda/installation.py
```diff
@@ -117,9 +117,10 @@ def _prepare_configuration(storage, payload, ksdata):
 
     # schedule network configuration (if required)
     if conf.system.provides_network_config:
+        overwrite = isinstance(payload, LiveImagePayload)
         network_config = TaskQueue("Network configuration", N_("Writing network configuration"))
         network_config.append(Task("Network configuration",
-                                   ksdata.network.execute, (payload, )))
+                                   network.write_configuration, (overwrite, )))
```
poncovka: We should call `InstallWithTasks` of the Network module here, but it will be easier to implement it after the Payload module is done.

pyanaconda/network.py
jkonecny12: Thanks for the confirmation.

PR #2189, opened 2019-10-18
Add packages ignore broken parameter support

jkonecny12:
This option is not recommended but there are a valid use-cases. It should be used only if user know what he/she is doing!

**Blocked by https://github.com/pykickstart/pykickstart/pull/282**

*Resolves: rhbz#1642013*

### DIFF
--- a/data/anaconda.conf
+++ b/data/anaconda.conf
@@ -76,6 +76,9 @@ enabled_repositories_from_treeinfo = addon optional variant
 # Enable installation from the closest mirror.
 enable_closest_mirror = True
 
+# Enable possibility to skip packages with conflicts and broken dependencies.
+enable_ignore_broken_packages = True
+
 # Check if payload supports the locales.
 check_supported_locales = False
 
--- a/pyanaconda/core/configuration/payload.py
+++ b/pyanaconda/core/configuration/payload.py
@@ -72,6 +72,19 @@ class PayloadSection(Section):
         return self._get_option("enable_closest_mirror", bool)
 
     @property
+    def enable_ignore_broken_packages(self):
+        """Enable possibility to skip packages with conflicts and broken dependencies.
+
+        This enables --ignorebroken parameter of the packages section and related DBus API.
+
+        If this feature is disabled Anaconda won't start the installation when
+        --ignorebroken paramater is used. Instead print error message to user
+        when Anaconda is started and quit the installation process.
+        It will also block use of related DBus API.
+        """
+        return self._get_option("enable_ignore_broken_packages", bool)
+
+    @property
     def check_supported_locales(self):
         """Check if payload supports the locales.
 
--- a/pyanaconda/payload/dnfpayload.py
+++ b/pyanaconda/payload/dnfpayload.py
@@ -57,7 +57,7 @@ import rpm
 from dnf.const import GROUP_PACKAGE_TYPES
 
 from blivet.size import Size
-from pykickstart.constants import GROUP_ALL, GROUP_DEFAULT, KS_MISSING_IGNORE
+from pykickstart.constants import GROUP_ALL, GROUP_DEFAULT, KS_MISSING_IGNORE, KS_BROKEN_IGNORE
 
 from pyanaconda.anaconda_loggers import get_packaging_logger, get_dnf_logger
 log = get_packaging_logger()
@@ -573,8 +573,10 @@ class DNFPayload(payload.PackagePayload):
 
         # feed it to DNF
         try:
+            # FIXME: Remove self._base.conf.strict workaround when bz1761518 is fixed
             # install_specs() returns a list of specs that appear to be missing
-            self._base.install_specs(install=include_list, exclude=exclude_list)
+            self._base.install_specs(install=include_list, exclude=exclude_list,
+                                     strict=self._base.conf.strict)
         except dnf.exceptions.MarkingErrors as e:
             log.debug("install_specs(): some packages, groups or modules "
                       " are missing or broken:\n%s", e)
@@ -695,6 +697,13 @@ class DNFPayload(payload.PackagePayload):
 
         if self.data.packages.retries is not None:
             config.retries = self.data.packages.retries
+
+        if self.data.packages.handleBroken == KS_BROKEN_IGNORE:
+            log.warning("\n*********************************************************************\n"
+                        "User has requested to skip broken packages. Using this option may result "
+                        "in an UNUSABLE system!\n"
+                        "*********************************************************************")
+            config.strict = False
 
         self._configure_proxy()
 
--- a/anaconda.spec.in
+++ b/anaconda.spec.in
@@ -35,7 +35,7 @@ Source0: %{name}-%{version}.tar.bz2
 %define libxklavierver 5.4
 %define mehver 0.23-1
 %define nmver 1.0
-%define pykickstartver 3.21-1
+%define pykickstartver 3.22-1
 %define pypartedver 2.5-2
 %define rpmver 4.10.0
 %define simplelinever 1.1-1
--- a/data/product.d/rhel.conf
+++ b/data/product.d/rhel.conf
@@ -16,6 +16,12 @@ ignored_packages = ntfsprogs
 enable_updates = False
 enable_closest_mirror = False
 
+# This feature may result in an installed system which won't have everything installed
+# as expected. Because of this problem this feature will be disabled on RHEL based variants.
+# When disabled the installation won't be started with --ignorebroken, instead it will inform user
+# that the feature is not supported on their product. This will also block usage of related DBus API.
+enable_ignore_broken_packages = False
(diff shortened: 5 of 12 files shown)


pyanaconda/payload/dnfpayload.py
```diff
@@ -696,6 +698,11 @@ def _configure(self):
         if self.data.packages.retries is not None:
             config.retries = self.data.packages.retries
 
+        if self.data.packages.handleBroken == KS_BROKEN_IGNORE:
+            log.warning("User requested to skip broken packages. The resulting system could be "
+                        "in UNDEFINED state!")
```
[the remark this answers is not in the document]
jkonecny12: Fixed.

jkonecny12:
**UPDATED:**
* Make the log message more visible and change wording.

jkonecny12:
jenkins, test this please

jkonecny12:
**UPDATED:**
* Add possibility to turn this off by configuration. Disable automatically on RHEL based systems.

data/anaconda.conf
```diff
@@ -76,6 +76,14 @@ enabled_repositories_from_treeinfo = addon optional variant
 # Enable installation from the closest mirror.
 enable_closest_mirror = True
 
+# Enable %packages --ignorebroken feature.
```
poncovka: This configuration option doesn't enable only the kickstart option, but also the DBus support, so this description is not correct. I would keep it more general.

data/anaconda.conf
```diff
@@ -76,6 +76,14 @@ enabled_repositories_from_treeinfo = addon optional variant
...
+# This will automatically skip conflicting and broken packages. This may result
+# in getting installed system which won't have everything installed as expected.
+# Because of this problem this feature will be disabled for some platforms. If disabled
+# the installation won't be started, instead it will inform user that the feature is not
+# supported on their platform.
```
poncovka: I don't think that this super long explanation needs to be here. This file should be easy to read. Also, it is not clear whether you are talking about the configuration option or the kickstart option. I would move the explanation to `rhel.conf` to document why we have decided to disable it in RHEL.

pyanaconda/core/configuration/payload.py
```diff
@@ -71,6 +71,21 @@ def enable_closest_mirror(self):
...
+    @property
+    def enable_ignore_broken_packages(self):
+        """Enable --ignorebroken parameter of packages section.
+
+        This parameter could be dangerous for users because it will skip packages
+        with conflicting files so installed system may not contain all the required
+        functionality.
+
+        In case this feature will be disabled we won't start the installation when
+        --ignorebroken paramater is used.
```
poncovka: Again, this documentation should be more general and more focus on what happens when the configuration option is set to False or True. At the end, it is still the user who decides whether he wants to ignore the broken packages or not. It is not very clear from the description.

pyanaconda/modules/payload/payloads/dnf/packages/packages.py
```diff
@@ -355,6 +369,30 @@ def set_missing_ignored(self, missing_ignored):
...
+        return self._broken_ignored
+
+    def set_broken_ignored(self, broken_ignored):
+        """Set if the packages that have conflicts with other packages should be ignored.
+
+        :param missing_ignored: True if broken packages should be ignored.
+        :type missing_ignored: bool
+        :raise: UnsupportedValueError if ignorebroken is disabled on this platform.
+        """
+        if not conf.payload.enable_ignore_broken_packages:
```
poncovka: It should be always ok to call the setter with `False`, no?

jkonecny12: Yes it should but on the other hand what's the point to set False something which is disabled and can't be True. I rather taken this approach, basically you can't use the API if the feature is disabled, it seems more clear to me.

jkonecny12:
**UPDATED:**
* Tried to make the documentation better based on @poncovka suggestions.
* Resolved conflicts.

data/anaconda.conf
jkonecny12: Reworded.

data/anaconda.conf
jkonecny12: Reworded.

pyanaconda/core/configuration/payload.py
jkonecny12: Reworded.

pyanaconda/modules/payload/payloads/dnf/packages/packages.py
poncovka: I can imagine an addon that would like to always disable this feature. With this approach, the addon will have to check `conf` first and then call the DBus method. I don't really like it. You should be allowed to disable unsupported features, because you might not really know whether they are unsupported in the current configuration.

data/anaconda.conf
```diff
@@ -76,6 +76,16 @@ enabled_repositories_from_treeinfo = addon optional variant
...
 enable_closest_mirror = True
 
+# Enable possibility to skip packages with conflicts and broken dependencies.
+#
+# This enables --ignorebroken parameter of the packages section and related DBus API.
+#
+# If this feature is disabled Anaconda won't start the installation when
+# --ignorebroken paramater is used. Instead print error message to user
+# when Anaconda is started and quit the installation process.
+# It will also block use of related DBus API.
```
poncovka: The first sentence it sufficient for this file. The remaining text is too detailed and can easily become obsolete. If you want to make sure that the feature for skipping broken packages is properly documented, I suggest to write a short paragraph in the Anaconda documentation. However, it is also documented in `pyanaconda.core.configuration.payload`, so it is not necessary.

pyanaconda/modules/payload/payloads/dnf/packages/packages.py
```diff
@@ -355,6 +368,30 @@ def set_missing_ignored(self, missing_ignored):
...
+            raise UnsupportedValueError(
+                "The %packages --ignorebroken feature is not supported on this platform"
```
poncovka: This error shouldn't mention the kickstart option, because that is handled elsewhere.

pyanaconda/modules/payload/payloads/dnf/packages/packages.py
```diff
@@ -355,6 +368,30 @@ def set_missing_ignored(self, missing_ignored):
...
+        :rtype: bool
+        """
+        return self._broken_ignored
+
+    def set_broken_ignored(self, broken_ignored):
+        """Set if the packages that have conflicts with other packages should be ignored.
+
+        :param missing_ignored: True if broken packages should be ignored.
+        :type missing_ignored: bool
+        :raise: UnsupportedValueError if ignorebroken is disabled on this platform.
```
poncovka: I think that we say `product` instead of `platform`, when we are talking about Fedora and RHEL.

pyanaconda/modules/payload/payloads/dnf/packages/packages.py
jkonecny12: I don't agree.

The point is in general the default state for some future feature could be enabled not disabled. In that case, you want to be able to only enable the feature? Also when something is disabled it shouldn't be usable all the time, not just accepting something but not everything, that seems really confusing to me. If we really want to go this way, then we should probably instead of raising an exception just silently not assign the value. It will than behave consistently all the time but I really don't want to do that because than user of the API don't know if something really happened or not.

Checking the conf is not such a pain and also it is not needed. You can try to set the value and just check for exception, I don't see a problem in that.

data/anaconda.conf
jkonecny12: I agree with everything except the last sentence.

The point is that ``pyanaconda.core.configuration.payload`` is in Anaconda code so when someone wants to change just the configuration he/she don't know about this place.

So or so I will leave here only the first sentence, it seems enough to me.

pyanaconda/modules/payload/payloads/dnf/packages/packages.py
jkonecny12: Fixed.

pyanaconda/modules/payload/payloads/dnf/packages/packages.py
jkonecny12: Fixed.

jkonecny12:
**UPDATED:**
* Implemented changes pointed by @poncovka . Thanks for suggestions!

pyanaconda/modules/payload/payloads/dnf/packages/packages.py
poncovka: Ok, it is not really that important to me.

jkonecny12:
**UPDATED:**
* Fix tests and solve two more platform occurrences.

InStallker:
test

jkonecny12:
**EDIT:**
* Rebased to fix conflicts.

PR #2194, opened 2019-10-21
Add code conventions to upstream

jkonecny12:
It will be better to find and better to maintain.


CONTRIBUTING.rst
```diff
@@ -111,6 +111,26 @@ Please note that there is a minimum review period of 24 hours for any patch. The
 
 All subsequent changes made to patches must be force-pushed to the PR branch before merging it into the main branch.
 
+Code conventions
+----------------
+
+It is important to have consistency across the codebase. This won't necessarily make your code work better, but it might help make it more understandable, time-consuming and less irritating to go through when doing a code review of your changes.
```
M4rtinK: `but it might help make it more understandable, time-consuming and less irritating to go through when doing a code review of your changes.`

I think we should reword this:

`but it might help make he codebase more understandable, easier to work with and more pleasant to go through when doing a code review.`

This should address:
- removes the implication that following code convention makes working with the codebase more time consuming
- removes the implication that code review is irritating ;-)

CONTRIBUTING.rst
```diff
@@ -111,6 +111,26 @@ Please note that there is a minimum review period of 24 hours for any patch. The
...
+
+In general we are trying to be as close as possible to `PEP8 <https://www.python.org/dev/peps/pep-0008/>`_ but also extending or modifying minor rules when it looked more readable to us. See list of the conventions below:
```
M4rtinK: `but also extending or modifying minor rules when it looked more readable to us.`
I suggest:
`but also extending or modifying minor PEP8 rules when it seems suitable in the context of our project.`

VladimirSlavik: +1, the suggestion makes it more neutral

CONTRIBUTING.rst

CONTRIBUTING.rst

jkonecny12:
**UPDATED:**
* Implemented changes from @M4rtinK. Thanks!

jkonecny12:
I would like to have an ACK from everyone on the team before merge. It will be consistent with the acceptance process for conventions.

PR #2196, opened 2019-10-22
Don't add more devices in boot-device NVRAM than the maximum allowed

jkonecny12:
Backport of https://github.com/rhinstaller/anaconda/pull/2119

The boot-device NVRAM variable has a maximum number of devices allowed, that is specified in /sys/firmware/devicetree/base/ibm,max-boot-devices OF property.

Adding more devices than the maximum number allowed in the boot-device NVRAM variable can cause issues so the device list need to be restricted to this maximum before updating the variable with nvram --update-config.

*Resolves: rhbz#1748756*

(cherry picked from commit 8d51f7659fd3abfb955d6edede96afaf6e0bbdac)

### DIFF
--- a/pyanaconda/bootloader/grub2.py
+++ b/pyanaconda/bootloader/grub2.py
@@ -584,9 +584,11 @@ class IPSeriesGRUB2(GRUB2):
 
         # The boot-device NVRAM variable has a maximum number of devices allowed,
         # don't add more than the limit in the ibm,max-boot-devices OF property.
-        with open("/sys/firmware/devicetree/base/ibm,max-boot-devices", "rb") as f:
-            limit = int.from_bytes(f.read(4), byteorder='big')
-            boot_list = boot_list[:limit]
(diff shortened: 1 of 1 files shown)


jkonecny12:
@martinezjavier I've created a backport of the rhel-8 PR but I'm not able to test this. Is it fine to skip testing in this case and go with a blind merge?

martinezjavier:
@jkonecny12 I would just test that it doesn’t cause regression on ppc64le OF machines for the common case where there are no more devices than the maximum number allowed.

In other words, just testing that installation is successful on any ppc64le machine should be enough for testing I think.

martinezjavier:
@jkonecny12 @poncovka so it seems that there are cases where the sysfs entry isn't present, so we also need #2197.

jkonecny12:
Added related commit to fix issue in the original one https://github.com/rhinstaller/anaconda/pull/2197.

jkonecny12:
Tested on ppc64le and x86_64 installation. Both went fine so I'm merging this now.

PR #2197, opened 2019-10-22
Only attempt to open the ibm,max-boot-devices sysfs entry if it exists

martinezjavier:
The /sys/firmware/devicetree/base/ibm,max-boot-devices sysfs entry defines
a OF property that contains the maximum number of devices that are allowed
in the boot-device NVRAM variable.

But this property may not be present in some ppc64le machines so it should
only be read if the file exists.

Related: rhbz#1748756


jkonecny12:
Jenkins, it's ok to test.

PR #2200, opened 2019-10-24
Forbid trailing fullstop (dot) in hostname validation (#1648107)

VladimirSlavik:
Since some time we use systemd for setting hostname immediately, and it does not accept the dot. Systemd claims this is according to kernel and only propagates the limitation closer to user. Survey of other hostname change methods shows this is consistently obeyed, by either:
a) refusing the dot (hostnamectl --static)
b) silently dropping the dot (hostnamectl)
c) sanitizing the input and using pretty hostname (GNOME control center)

For more details see bug 1648107: https://bugzilla.redhat.com/show_bug.cgi?id=1648107

Hostname ks module validation test is also extended for this case.

### DIFF
--- a/pyanaconda/network.py
+++ b/pyanaconda/network.py
@@ -94,10 +94,11 @@ def check_ip_address(address, version=None):
         return False
 
 
-def is_valid_hostname(hostname):
+def is_valid_hostname(hostname, local=False):
     """Check if the given string is (syntactically) a valid hostname.
 
-    :param hostname: a string to check
+    :param str hostname: a string to check
+    :param bool local: is the hostname static (for this system) or not (on the network)
     :returns: a pair containing boolean value (valid or invalid) and
               an error message (if applicable)
     :rtype: (bool, str)
@@ -107,6 +108,9 @@ def is_valid_hostname(hostname):
 
     if len(hostname) > 255:
         return (False, _("Host name must be 255 or fewer characters in length."))
+
+    if local and hostname[-1] == ".":
+        return (False, _("Local host name must not end with period '.'."))
 
     if not re.match('^' + HOSTNAME_PATTERN_WITHOUT_ANCHORS + '$', hostname):
         return (False, _("Host names can only contain the characters 'a-z', "
--- a/tests/nosetests/pyanaconda_tests/network_test.py
+++ b/tests/nosetests/pyanaconda_tests/network_test.py
@@ -52,6 +52,9 @@ class NetworkTests(unittest.TestCase):
 
         self.assertFalse(network.is_valid_hostname("Lennart's Laptop")[0])
 
+        self.assertFalse(network.is_valid_hostname("own.hostname.cannot.end.in.dot.",
+                                                   local=True)[0])
+
     def prefix2netmask2prefix_test(self):
         lore = [
                 (0, "0.0.0.0"),
--- a/tests/nosetests/pyanaconda_tests/network_test.py
+++ b/tests/nosetests/pyanaconda_tests/network_test.py
@@ -19,9 +19,11 @@
 from pyanaconda import network
 import unittest
 
+
 class NetworkTests(unittest.TestCase):
 
     def is_valid_hostname_test(self):
+        """Test hostname validation."""
 
         self.assertFalse(network.is_valid_hostname("")[0])
         self.assertFalse(network.is_valid_hostname(None)[0])
@@ -56,6 +58,7 @@ class NetworkTests(unittest.TestCase):
                                                    local=True)[0])
 
     def prefix2netmask2prefix_test(self):
+        """Test netmask and prefix conversion."""
         lore = [
                 (0, "0.0.0.0"),
                 (1, "128.0.0.0"),
@@ -98,6 +101,7 @@ class NetworkTests(unittest.TestCase):
         self.assertEqual(network.prefix_to_netmask(33), "255.255.255.255")
 
     def nm_check_ip_address_test(self,):
+        """Test IPv4 and IPv6 address checks."""
         good_IPv4_tests = [
                 '1.2.3.4',
                 '0.0.0.0',
@@ -222,6 +226,7 @@ class NetworkTests(unittest.TestCase):
             self.assertFalse(network.check_ip_address(i, version=4))
 
     def hostname_from_cmdline_test(self):
+        """Test extraction of hostname from cmdline."""
         cmdline = {"ip": "10.34.102.244::10.34.102.54:255.255.255.0:myhostname:ens9:none"}
         self.assertEqual(network.hostname_from_cmdline(cmdline), "myhostname")
         # ip takes precedence
--- a/pyanaconda/ui/gui/spokes/network.py
+++ b/pyanaconda/ui/gui/spokes/network.py
@@ -1516,7 +1516,7 @@ class NetworkSpoke(FirstbootSpokeMixIn, NormalSpoke):
 
     def on_apply_hostname(self, *args):
         hostname = self.network_control_box.hostname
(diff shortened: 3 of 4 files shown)


tests/nosetests/pyanaconda_tests/network_test.py
```diff
@@ -47,11 +47,13 @@ def is_valid_hostname_test(self):
         self.assertFalse(network.is_valid_hostname(
             "section.may.not.end.with.hyphen-")[0])
 
-        self.assertTrue(network.is_valid_hostname("0-0.")[0])
-        self.assertTrue(network.is_valid_hostname("0.")[0])
+        #self.assertTrue(network.is_valid_hostname("0-0.")[0])
+        #self.assertTrue(network.is_valid_hostname("0.")[0])
```
VladimirSlavik: I am not sure what this actually tests. It could be that ``0`` is the minimal viable hostname component and thus also minimal hostname, and ``0-0`` a minimal variant that contains a hyphen. But I have seen the specs for that literally a few minutes ago, so I don't trust myself with that.

@rvykydal, you added this half a decade ago, so if you have any advice, I'd be very grateful. But if you have no idea any more, that's ok too :-)

pyanaconda/core/regexes.py
```diff
@@ -103,7 +103,7 @@
 # with a period, but it can end with one.
```
poncovka: Please, update also the comment about the dot.

pyanaconda/core/regexes.py
```diff
@@ -103,7 +103,7 @@
...
 # This regex uses negative lookahead and lookback assertions to enforce the
 # hyphen rules and make it way more confusing
-HOSTNAME_PATTERN_WITHOUT_ANCHORS = r'(?:(?!-)[A-Za-z0-9-]{1,63}(?<!-)(?:\.(?!-)[A-Za-z0-9-]{1,63}(?<!-))*\.?)'
+HOSTNAME_PATTERN_WITHOUT_ANCHORS = r'(?:(?!-)[A-Za-z0-9-]{1,63}(?<!-)(?:\.(?!-)[A-Za-z0-9-]{1,63}(?<!-))*)'
```
jkonecny12: Every time I'm seeing a change in these rexeps I am getting hives. :(

Don't get me wrong, I like regexps but these giant things are not good way to go.

poncovka: I think they are pretty ;-)

VladimirSlavik: With some assistive technology, this one is still quite readable...

poncovka:
It should be `Resolves:` in the commit message.

Just a note, this is documented in [Contribution guidelines](https://anaconda-installer.readthedocs.io/en/latest/contributing.html#commits-for-rhel-branches). While it isn't required for the Fedora branches, we always mention the bug if there is one. It is really handy when you are searching the history of the code.

tests/nosetests/pyanaconda_tests/network_test.py
rvykydal: Sorry I don't remember. I'd guess the dot at the end is the main point of these checks but your guess may be true as well. I would remove it.

VladimirSlavik:
@rvykydal you're right. It seems this regexp is used everywhere as apparently until now hostname==FQDN, for everything. I'll have to look at the actual usages to make sure we have a way of dealing with that.

VladimirSlavik:
The current rewrite tries to keep modifications to minimum while satisfying all the objectives. I chose this way because renaming everything would make the internals diverge from what we call things on the outside.

pyanaconda/network.py
```diff
@@ -56,7 +56,7 @@
            "wait_for_connecting_NM_thread", "wait_for_network_devices", "wait_for_connected_NM",
```
jkonecny12: You have a typo in your first commit message:
`coplication` -> `complication`

pyanaconda/network.py
```diff
@@ -94,6 +94,23 @@ def check_ip_address(address, version=None):
         return False
 
 
+def is_valid_hostname_to_set(hostname):
+    """Check if the given string is (syntactically) a valid hostname for the running system.
```
jkonecny12: I know it's not easy but we should definitely have better description for how `is_valid_hostname_to_set` differs from `is_valid_hostname`. From what is written here it doesn't explain why we have both methods. If this is for running system for what is the second method?

Maybe call this method `is_valid_hostname_without_trailing_dot` and explain in description why we need to have hostname without the trailing dot?

tests/nosetests/pyanaconda_tests/network_test.py
```diff
@@ -19,9 +19,11 @@
 from pyanaconda import network
 import unittest
 
+
 class NetworkTests(unittest.TestCase):
 
     def is_valid_hostname_test(self):
+        """Test hostname validation."""
```
jkonecny12: We don't have any convention for this. So I guess @rvykydal should tell if this is fine because he is looking onto this code most of the time.

pyanaconda/network.py
VladimirSlavik: I don't think `_without_trailing_dot` is the right way out. The underlying reason is that different rules apply, not just the dot. Related to that is that I see the real hurdle for future comprehension as "hostname" now becoming an ambiguous word. Are we talking about hostname for us, or for anyone else on the network? So I tried to follow that split (us/others).

How about this?
- `is_valid_hostname_for_this_system`
- `is_valid_hostname_of_other_servers`

There are also these possible changes:
- Have two parallel implementations of the check (I really dislike that as it touches too much to my liking... but don't rule it out yet)
- Rename everything (I don't like that but it seems more clear than just hostname)
- Put more information into doc strings of the two methods

tests/nosetests/pyanaconda_tests/network_test.py
VladimirSlavik: For the record, I'm totally fine with dropping the commit, if having human readable names on tests is not something we need.

pyanaconda/network.py
poncovka: Maybe, we could just extend the function is_valid_hostname to:
`is_valid_hostname(hostname, allow_trailing_dot=False)`

VladimirSlavik: Conclusion from team discussion: Keep one function, add a parameter `static` or `local` that defaults to `False` and activates the more strict check.

VladimirSlavik:
Update: Changed the setup so that there is an additional parameter.

pyanaconda/network.py
```diff
@@ -108,6 +109,9 @@ def is_valid_hostname(hostname):
     if len(hostname) > 255:
         return (False, _("Host name must be 255 or fewer characters in length."))
 
+    if local and hostname[-1] == ".":
+        return (False, "Local host name must not end with period '.'.")
```
poncovka: This string should be translated. Use: `_("Local host name must not end with period '.'.")`

pyanaconda/network.py
```diff
@@ -94,10 +94,11 @@ def check_ip_address(address, version=None):
         return False
 
 
-def is_valid_hostname(hostname):
+def is_valid_hostname(hostname, local=False):
     """Check if the given string is (syntactically) a valid hostname.
 
-    :param hostname: a string to check
+    :param str hostname: a string to check
+    :param bool local: is the hostname static (for this system) or not (anywhere on the network)
```
jkonecny12: I would remove `anywhere` here, `on the network` seems good enough to me.

VladimirSlavik: Done!

VladimirSlavik:
Thank you for all your help, everyone!

PR #2201, opened 2019-10-24
Don't enable BLS for Xen machines

martinezjavier:
PV and PVH Xen DomU guests boot with pygrub that doesn't have BLS support.
Also Xen Dom0 use the menuentries from 20_linux_xen and not the ones from
10_linux. So BLS support needs to be disabled for both Xen Dom0 and DomU
and use a traditional grub.cfg file generated by the grub2-mkconfig tool.

Related: rhbz#1703700

### DIFF
--- a/pyanaconda/bootloader/grub2.py
+++ b/pyanaconda/bootloader/grub2.py
@@ -273,6 +273,13 @@ class GRUB2(BootLoader):
             log.warning("BLS support disabled due new-kernel-pkg being present")
             self.use_bls = False
 
+        hv_type_path = "/sys/hypervisor/type"
+        if self.use_bls and os.access(hv_type_path, os.F_OK):
+            with open(hv_type_path, "r") as fd:
+                if fd.readline().strip() == "xen":
+                    log.warning("BLS support disabled because is a Xen machine")
+                    self.use_bls = False
+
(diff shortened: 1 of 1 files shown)


martinezjavier:
I don't have a Xen setup to test this patch but @AdamWill said that he could give it a try (thanks a lot for that!)

AdamWill:
This seems to do what it's intended to do - I get a non-BLS install when installing to a Xen domU - but the installed system still doesn't boot for me. Not sure what's going on yet.

jkonecny12:
Jenkins, it's ok to test.

jkonecny12:
We don't have access to the Xen machine too so we will only provide a sanity checking for this PR.

M4rtinK:
I've run a couple kickstart tests in this (all from the "bootloader" group & a generic selection) and all passed, so I think this is good to go from the sanity perspective.

M4rtinK:
@AdamWill How it it look like on your side (eq. on actual XEN) ? Should we merge this or rather wait a bit ?

AdamWill:
I think it should be fine to merge.

M4rtinK:
> I think it should be fine to merge.

OK then. :)

PR #2203, opened 2019-10-29
Do not run hwclock when using image install mode

bcl:
Currently anaconda --image=... results in it calling the hwclock
program which copies the hardware clock value to the system time. And
depending on the kickstart settings, can cause large time jumps eg.
using --localtime on a --utc based hwclock system.

Resolves: rhbz#1766785


poncovka:
Jenkins, test this please.

PR #2204, opened 2019-10-30
Add shared payload module and API

jkonecny12:
Add new classes for shared API and implementation what have to be implemented by modules.

This is new required API for modules:
* RequiredSpace
* SupportedSourceTypes
* Sources
* SetSources
* HasSources
* PreInstallWithTasks
* InstallWithTasks
* PostInstallWithTasks
* SetUpSourcesWithTask
* TearDownSourcesWithTask

I've tried to adapt existing tests and add new ones to at least Live OS. Others than Live OS are harder because they can't really implement the APIs because of missing sources implementation.

Please share feedback, we can rework this or add/remove some of the API above. It would be great to have a good base ground for further implementation.


pyanaconda/modules/common/constants/namespaces.py
```diff
@@ -96,6 +96,11 @@
     "Baz"
 )
 
+PAYLOAD_BASE_NAMESPACE = (
+    *PAYLOAD_NAMESPACE,
+    "Base"
```
poncovka: This is inconsistent with our usual approach. Ideally, the base namespace for payloads would be `Payload.Payload` and example of a subnamespace `Payload.Payload.DNF`. Otherwise, you are using a namespace of the main module. It can be confusing and cause problems, because we might need these names  in the future.

I wonder if we should use the same solution as for the `Users` service and rename the `Payload` service to `Payloads` since we plan to eventually support multiple payloads anyway.

pyanaconda/modules/payload/payloads/live_image/live_image.py
```diff
@@ -63,16 +63,15 @@ def __init__(self):
...
-        self.required_space_changed = Signal()
-
         self._kernel_version_list = []
         self.kernel_version_list_changed = Signal()
 
         self._image_path = conf.target.system_root + "/disk.img"
 
         self._requests_session = None
 
+        self.set_required_space(1024 * 1024 * 1024)
```
poncovka: Why is this default value necessary? Do you know something about it?

tests/nosetests/pyanaconda_tests/module_payload_live_os_test.py
```diff
@@ -42,25 +44,100 @@ def setUp(self):
...
+        source2 = self._prepare_source()
+
+        paths = {
+            "Source1": source1
+        }
+        paths2 = {
+            "Source2": source2
+        }
+
+        set_up_dbus_container_mock(source_container, paths)
```
poncovka: This is really not necessary. If you patch only `DBus.publish_object`, you can call `PayloadSourceContainer.to_object_path` directly to create the mapping and get unique DBus paths for your source objects. Then `SetSources` will work as expected.

tests/nosetests/pyanaconda_tests/__init__.py
```diff
@@ -239,3 +239,27 @@ def patch_dbus_get_proxy(func):
     This is a shortcut to avoid creating of DBus proxies using DBus.
     """
     return patch('pyanaconda.dbus.DBus.get_proxy')(func)
+
+
+def set_up_dbus_container_mock(container_mock, paths_dict):
```
poncovka: I don't believe that this function is useful.  DBus containers can be used in unit tests decorated with `patch_dbus_publish_object`. They do all the mapping and generating of DBus paths, they just don't publish anything on DBus.

pyanaconda/modules/payload/payloads/payload_base_interface.py
```diff
@@ -0,0 +1,76 @@
...
+
+    @property
+    def Sources(self) -> List[ObjPath]:
+        """Get list of sources attached to this payload."""
+        return PayloadSourceContainer.to_object_path_list(
+            self.implementation.sources
+        )
+
+    def SetSources(self, sources: List[ObjPath]):
+        """Attached list of sources to this payload.
```
rvykydal: Attached -> Attach ?

pyanaconda/modules/payload/payloads/payload_base_interface.py
```diff
@@ -0,0 +1,76 @@
...
+
+        Before setting the sources, please make sure the sources are not initialized otherwise
+        the SourceSetupError exception will be raised. Payload have to cleanup after itself.
```
rvykydal: Payload has to

pyanaconda/modules/payload/payloads/live_os/live_os.py
```diff
@@ -95,31 +95,28 @@ def _check_source_availability(self, message):
...
-        TODO: Add missing check if source is ready. Until then you shouldn't call this when
-        source is not ready.
-
-        TODO: This is not that fast as I thought (a few seconds). Caching or task?
-
-        :return: required size in bytes
-        :rtype: int
-        """
+    @staticmethod
+    def _get_requiered_space():
```
rvykydal: a typo in the name, should  be _get_required_space ?

pyanaconda/modules/payload/payloads/payload_base.py
```diff
@@ -123,3 +123,27 @@ def publish_handler(self):
         :rtype: string
         """
         pass
+
+    @abstractmethod
+    def pre_install_with_tasks(self):
+        """Create pre installation tasks.
```
rvykydal: Regarding the with_tasks and WithTasks method, for the consistency, it seems to me that we are describing what is the task doing rather than that we are creating the task(s). So something like "Prepare for installation" or "Do pre-installation steps / tasks?".
And for the install_with_tasks below: "Install the payload" instead of "Create tasks for installation."

tests/nosetests/pyanaconda_tests/__init__.py
jkonecny12: Fixed by removing.

tests/nosetests/pyanaconda_tests/module_payload_live_os_test.py
jkonecny12: Fixed.

pyanaconda/modules/payload/payloads/live_os/live_os.py
```diff
@@ -95,47 +95,44 @@ def _check_source_availability(self, message):
...
         return get_dir_size("/") * 1024
 
     def set_up_sources_with_task(self):
-        """Set up installation source."""
+        """Set up installation sources."""
         self._check_source_availability("Set up source failed - source is not set!")
 
-        return SetUpSourcesTask(self._sources)
+        task = SetUpSourcesTask(self._sources)
+        # TODO: This is not that fast as I thought (a few seconds). Caching or solved in task?
```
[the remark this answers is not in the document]
jkonecny12: The problem is function called by this signal. I moved the TODO to the function it is not clear here.

pyanaconda/modules/payload/payloads/payload_base_interface.py
```diff
@@ -0,0 +1,115 @@
...
+        the initialization task will run with the old list.
+
+        :raise: IncompatibleSourceError when source is not a supported type
+                SourceSetupError when attached sources are initialized
+        """
+        self.implementation.set_sources(
+            PayloadSourceContainer.from_object_path_list(
+                sources
+                )
+            )
```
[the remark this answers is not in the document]
jkonecny12: Fixed.

pyanaconda/modules/common/constants/namespaces.py
jkonecny12: I go with Payloads

tests/nosetests/pyanaconda_tests/module_payload_live_os_test.py
jkonecny12: Fixed.

pyanaconda/modules/payload/payloads/payload_base_interface.py
```diff
@@ -0,0 +1,115 @@
...
+
+    def HasSource(self) -> Bool:
+        """Check if any source is attached to this payload."""
+        return self.implementation.has_source()
+
+    def PreInstallWithTasks(self) -> List[ObjPath]:
```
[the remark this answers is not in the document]
jkonecny12: Fixed, added for Pre, Post and Install API.

tests/nosetests/pyanaconda_tests/module_payload_live_image_test.py
```diff
@@ -64,7 +73,7 @@ def liveimg_simple_kickstart_test(self):
         # Use live disk image installation
         liveimg --url="http://my/super/path"
         """
-        self.check_kickstart(ks_in, ks_out)
+        self.shared_tests.check_kickstart(ks_in, ks_out)
```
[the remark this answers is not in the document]
jkonecny12: Could be, I'm still fear that the test file will be too big to find anything useful there.

However, truth is that it is now easier to do, thanks to the fact that payloads are created based on kickstart.

pyanaconda/modules/payload/payloads/live_image/live_image.py
jkonecny12: I was able to find the source of this value and it was added in commit which added live image support. No explanation why this value.

However, when I'm looking into the code I guess we should keep this. It is required when server won't provide us `content-length` in the header. If we don't know this value we can only give a reasonable guess which is this `1G` value.

jkonecny12: OK, when I'm looking on the new moved code it won't work as it should. I'll change that.

pyanaconda/modules/payload/payloads/payload_base_interface.py
jkonecny12: Fixed.

pyanaconda/modules/payload/payloads/payload_base_interface.py
jkonecny12: Fixed.

pyanaconda/modules/payload/payloads/live_os/live_os.py
jkonecny12: Fixed.

pyanaconda/modules/payload/payloads/payload_base.py
jkonecny12: Fixed. Thanks for pointing that out.

jkonecny12:
**UPDATED:**
* Fix Live Image payload default required size value.
* Implemented notes from @poncovka and @rvykydal -- Thanks a lot for your input!

pyanaconda/modules/payload/payloads/live_image/live_image.py
```diff
@@ -63,16 +63,24 @@ def __init__(self):
...
+    def set_required_space(self, required_space):
+        """Set required space for the installation.
+
+        When the value is not known (0) change the value to 1G as reasonable default.
+        """
+        if required_space == 0:
+            log.debug("Required size not known, use 1G instead.")
+            super().set_required_space(1024 * 1024 * 1024)
+        else:
+            super().set_required_space(required_space)
```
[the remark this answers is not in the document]
jkonecny12: The reason why is it here is that it will signal PropertiesChanged with the value. We want this signal but with correct value.

jkonecny12:
**UPDATED:**
* Fix the payload base namespace name.
* Add implementation for default required space.

pyanaconda/modules/common/constants/namespaces.py
```diff
@@ -96,6 +96,11 @@
     "Baz"
 )
 
+PAYLOAD_BASE_NAMESPACE = (
+    *PAYLOAD_NAMESPACE,
+    "Payloads"
```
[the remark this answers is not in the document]
jkonecny12: Fixed.

pyanaconda/modules/payload/payloads/live_image/live_image.py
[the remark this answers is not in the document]
jkonecny12: Fixed by adding default space solution.

jkonecny12:
**UPDATED:**
* Remove reasonable from the default required method name.

pyanaconda/modules/payload/payloads/live_os/live_os.py
```diff
@@ -95,50 +95,54 @@ def _check_source_availability(self, message):
...
 
-        return TearDownSourcesTask(self._sources)
+        task = TearDownSourcesTask(self._sources)
+        task.stopped_signal.connect(lambda: self.set_required_space(0))
+
+        return task
 
-    def pre_install_with_task(self):
-        """Prepare intallation task."""
+    def pre_install_with_tasks(self):
```
VladimirSlavik: pre -> prepare? To me it is really confusing. Why is one word shortened and the other not? I'd expect prepare_install vs pre_inst but here it's a bit of both. Or is it accepted terminology to call things "pre" in Anaconda?

jkonecny12: It's more historical reason. We are changing most of these things. Good point about this one, it should be changed.

PR #2207, opened 2019-11-05
Add support for installation tasks of DBus addons

poncovka:
DBus addons should return their "setup" tasks in the `ConfigureWithTasks` method
and their "execute" tasks in the `InstallWithTasks` method.

Anaconda will schedule and run the installation tasks of the DBus addons during
the installation.

### DIFF
--- a/tests/nosetests/pyanaconda_tests/module_boss_install_test.py
+++ b/tests/nosetests/pyanaconda_tests/module_boss_install_test.py
@@ -20,7 +20,7 @@ import unittest
 from mock import Mock, patch, call
 
 from pyanaconda.modules.boss.install_manager import InstallManager
-from pyanaconda.modules.boss.install_manager.installation import SystemInstallationTask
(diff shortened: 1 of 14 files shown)


tests/nosetests/pyanaconda_tests/module_boss_install_test.py
```diff
@@ -106,3 +106,39 @@ def install_three_tasks_test(self, proxy_getter):
         self.assertIsInstance(main_task, DBusMetaTask)
         self.assertEqual(main_task.name, "Install the system")
         self.assertEqual(main_task._subtasks, [task_proxy, task_proxy, task_proxy])
+
+    @patch('pyanaconda.dbus.DBus.get_proxy')
+    def configure_runtime_test(self, proxy_getter):
+        """Configure the runtime system with three tasks."""
+        observers = []
+
+        observer = Mock()
```
jkonecny12: What about start using `create_autospec()` here?

PR #2209, opened 2019-11-06
Remove the enum PartitioningMethod from the Storage spoke in GUI

poncovka:
The enum is otherwise useless. Instead, use the constants of the DBus
Storage module. Later, we can use these values directly to create the
requested partitioning via DBus.


poncovka:
Updated the commit message.

PR #2213, opened 2019-11-11
Cmdline deflagification

VladimirSlavik:
This adds `pyanaconda.kernel.cmdline` so that `pyanaconda.flags.flags.cmdline` can be eventually removed.
- All functionality is identical.
- The old name still exists, works, and in fact returns the same instance of `KernelArguments`.
- All internal uses of the old are replaced with the new.
- Same for tests.

A quick check of other potential consumers shows no use in:
- Initial setup
- KDdump addon
- OSCAP addon

Up for discussion: What level of refactoring is desired? I'm not entirely happy that `KernelArguments` exposes its various `read_` and `from_` methods which makes it possible to potentially call them on the global shared instance and fill it with nonsense. So far it was apparently not a problem though.

### DIFF
--- a/anaconda.py
+++ b/anaconda.py
@@ -39,7 +39,7 @@ def exitHandler(rebootData, storage):
     if flags.usevnc:
         vnc.shutdownServer()
 
-    if "nokill" in flags.cmdline:
+    if "nokill" in kernel_arguments:
         util.vtActivate(1)
         print("anaconda halting due to nokill flag.")
         print("The system will be rebooted when you press Ctrl-Alt-Delete.")
@@ -248,7 +248,8 @@ if __name__ == "__main__":
 
     # do this early so we can set flags before initializing logging
     from pyanaconda.flags import flags
-    (opts, depr) = parse_arguments(boot_cmdline=flags.cmdline)
+    from pyanaconda.core.kernel import kernel_arguments
+    (opts, depr) = parse_arguments(boot_cmdline=kernel_arguments)
 
     from pyanaconda.core.configuration.anaconda import conf
     conf.set_from_opts(opts)
@@ -284,7 +285,7 @@ if __name__ == "__main__":
     # see if we're on s390x and if we've got an ssh connection
     uname = os.uname()
     if uname[4] == 's390x':
-        if 'TMUX' not in os.environ and 'ks' not in flags.cmdline and conf.target.is_hardware:
+        if 'TMUX' not in os.environ and 'ks' not in kernel_arguments and conf.target.is_hardware:
             startup_utils.prompt_for_ssh()
             sys.exit(0)
 
@@ -293,7 +294,7 @@ if __name__ == "__main__":
         log.info("Using updates in /tmp/updates/ from %s", opts.updateSrc)
 
     # TODO: uncomment this when we're sure that we're doing the right thing
-    # with flags.cmdline *everywhere* it appears...
+    # with kernel_arguments *everywhere* it appears...
     #for arg in depr:
     #    stdout_log.warn("Boot argument '%s' is deprecated. "
     #                   "In the future, use 'inst.%s'.", arg, arg)
--- a/pyanaconda/core/kernel.py
+++ b/pyanaconda/core/kernel.py
@@ -21,12 +21,17 @@ import glob
 from collections import OrderedDict
 from pyanaconda.core.constants import CMDLINE_APPEND, CMDLINE_LIST, CMDLINE_FILES
 
+__all__ = ['KernelArguments', 'kernel_arguments']
 
-class KernelArguments(OrderedDict):
+
+class KernelArguments():
     """The kernel arguments.
 
     Hold boot arguments as an OrderedDict.
     """
+
+    def __init__(self):
+        self._data = OrderedDict()
 
     @classmethod
     def from_defaults(cls):
@@ -110,17 +115,17 @@ class KernelArguments(OrderedDict):
                 val = None
 
             # Some duplicate args create a space separated string
-            if key in CMDLINE_APPEND and self.get(key, None):
+            if key in CMDLINE_APPEND and self._data.get(key, None):
                 if val:
-                    self[key] = self[key] + " " + val
+                    self._data[key] = self._data[key] + " " + val
             # Some arguments can contain spaces so adding them in one string is not that helpful
             elif key in CMDLINE_LIST:
                 if val:
-                    if not self.get(key, None):
-                        self[key] = []
-                    self[key].append(val)
+                    if not self._data.get(key, None):
+                        self._data[key] = []
+                    self._data[key].append(val)
             else:
-                self[key] = val
+                self._data[key] = val
 
     def getbool(self, arg, default=False):
         """Return the boolean value of the given argument.
@@ -134,15 +139,36 @@ class KernelArguments(OrderedDict):
         :return: a boolean value of the argument
         """
         result = default
-        for a in self:
+        for a in self._data:
             if a == arg:
-                if self[arg] in ("0", "off", "no"):
+                if self._data[arg] in ("0", "off", "no"):
                     result = False
                 else:
                     result = True
             elif a == 'no' + arg:
                 result = False  # XXX: should noarg=off -> True?
         return result
+
+    def get(self, arg, default=None):
+        """Return the value of the given argument.
+
+        Propagates the call verbatim to the underlying dictionary.
+        """
+        return self._data.get(arg, default)
+
+    def __contains__(self, arg):
+        """Check for presence of an argument.
+
+        Propagates the call verbatim to the underlying dictionary.
+        """
+        return arg in self._data
+
+    def items(self):
+        """Return an iterator over all arguments.
+
+        Propagates the call verbatim to the underlying dictionary.
+        """
+        return self._data.items()
 
 
 kernel_arguments = KernelArguments.from_defaults()
--- a/pyanaconda/core/kernel.py
+++ b/pyanaconda/core/kernel.py
@@ -143,3 +143,6 @@ class KernelArguments(OrderedDict):
             elif a == 'no' + arg:
                 result = False  # XXX: should noarg=off -> True?
         return result
+
+
+kernel_arguments = KernelArguments.from_defaults()
--- a/pyanaconda/core/kernel.py
+++ b/pyanaconda/core/kernel.py
@@ -127,6 +127,21 @@ class KernelArguments():
             else:
                 self._data[key] = val
 
+    def is_enabled(self, arg):
+        """Return boolean value for the given argument.
+
+        Rules:
+        - 0, off, not present -> False
+        - the rest -> True
+        """
+        # None is stored when arg is present but had no value when parsing.
+        # So the "miss" value must be something else.
+        val = self._data.get(arg, False)
+        if val in ["0", "off", False]:
+            return False
+        else:
+            return True
+
     def getbool(self, arg, default=False):
         """Return the boolean value of the given argument.
 
--- a/pyanaconda/geoloc.py
+++ b/pyanaconda/geoloc.py
@@ -193,10 +193,7 @@ class Geolocation(object):
                 geolocation_enabled = False
 
         # and also check if geolocation was not disabled by boot or command like option
-        if not kernel_arguments.getbool('geoloc', True):
-            # kernel_arguments.getbool is used as it handles values such as
-            # 0, no, off and also nogeoloc as False
-            # and other values or geoloc not being present as True
+        if not kernel_arguments.is_enabled('geoloc'):
             geolocation_enabled = False
 
         # log the result
@@ -218,7 +215,7 @@ class Geolocation(object):
                 log.info("Geolocation is disabled for image or directory installation.")
             elif flags.automatedInstall:
                 log.info("Geolocation is disabled due to automated kickstart based installation.")
-            if not kernel_arguments.getbool('geoloc', True):
+            if not kernel_arguments.is_enabled('geoloc'):
                 log.info("Geolocation is disabled by the geoloc option.")
 
     def refresh(self):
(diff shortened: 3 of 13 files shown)


pyanaconda/core/kernel.py
```diff
@@ -143,3 +143,6 @@ def getbool(self, arg, default=False):
             elif a == 'no' + arg:
                 result = False  # XXX: should noarg=off -> True?
         return result
+
+
+cmdline = KernelArguments.from_defaults()
```
poncovka: I am not sure if we want to keep the name `cmdline`. I would use `kernel_arguments` instead. Then the imports could be `from pyanaconda.core.kernel import kernel_arguments`. What do others think?

poncovka:
> Up for discussion: What level of refactoring is desired? I'm not entirely happy that `KernelArguments` exposes its various `read_` and `from_` methods which makes it possible to potentially call them on the global shared instance and fill it with nonsense. So far it was apparently not a problem though.

I don't think that it is a problem. However, I don't like that the class `KernelArguments` extends a dictionary, because it is much easier to mess with the values this way. Ideally, the dictionary should be used only as an internal implementation and the class should provide only methods required by Anaconda.

VladimirSlavik:
Ah, oops, I thought I ran the tests but I did not. Back to the editor...

pyanaconda/core/kernel.py
VladimirSlavik: Ah, in one word "yes". Solves a lot of baggage.

VladimirSlavik:
**UPDATE**: Class changed to provide the currently used interface instead of inheriting from OrderedDict.

anaconda.py
```diff
@@ -248,7 +248,8 @@ def setup_environment():
 
     # do this early so we can set flags before initializing logging
     from pyanaconda.flags import flags
```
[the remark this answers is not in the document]
VladimirSlavik: Yes, they are used a few lines below.

pyanaconda/modules/network/network.py
```diff
@@ -652,20 +652,20 @@ def get_dracut_arguments(self, iface, target_ip, hostname, ibft):
...
-        if 'BOOTIF' in kernel_arguments:
-            self.bootif = kernel_arguments['BOOTIF'][3:].replace("-", ":").upper()
-        if 'ifname' in kernel_arguments:
-            self.ifname_option_values = kernel_arguments.get("ifname", "").split()
-        if 'noipv6' in kernel_arguments:
+        log.debug("Applying boot options %s", kernel_args)
+        if 'ksdevice' in kernel_args:
+            self.default_device_specification = kernel_args.get('ksdevice')
+        if 'BOOTIF' in kernel_args:
+            self.bootif = kernel_args['BOOTIF'][3:].replace("-", ":").upper()
```
poncovka: I think that this will not work.

pyanaconda/core/kernel.py
```diff
@@ -22,12 +22,15 @@
 from pyanaconda.core.constants import CMDLINE_APPEND, CMDLINE_LIST, CMDLINE_FILES
```
poncovka: Please, all the `__all__` statement to this file.

pyanaconda/core/kernel.py
```diff
@@ -134,12 +137,23 @@ def getbool(self, arg, default=False):
...
-                if self[arg] in ("0", "off", "no"):
+                if self._data[arg] in ("0", "off", "no"):
                     result = False
                 else:
                     result = True
             elif a == 'no' + arg:
                 result = False  # XXX: should noarg=off -> True?
         return result
+
+    def get(self, *args):
```
poncovka: The method `get` has only one optional argument `default=None`, so I think you can explicitly define it here.

pyanaconda/core/kernel.py
```diff
@@ -134,12 +137,23 @@ def getbool(self, arg, default=False):
...
+        return self._data.get(*args)
+
+    def __contains__(self, item):
+        return item in self._data
+
+    def items(self):
```
poncovka: All new methods should have some defined docstrings.

pyanaconda/core/kernel.py
```diff
@@ -22,12 +22,15 @@
 from pyanaconda.core.constants import CMDLINE_APPEND, CMDLINE_LIST, CMDLINE_FILES
 
 
-class KernelArguments(OrderedDict):
+class KernelArguments():
```
poncovka: We don't have tests for this class. Please, write some or create an issue for that.

pyanaconda/modules/network/network.py
VladimirSlavik: Thank you! Agreed. Makes me wonder two things:
- Is there some method you use to find these issues? I must be missing _something_, as I searched for all uses, and then got some more help from pylint, and then tests passed, and all of that was still not enough.
- Does anyone see any value in keeping the "I am a dictionary" behavior? I'd rather keep the interface small with just `get()`. But there is the question of addon usage, even if so far I did not find anyone using `flags.cmdline`...

poncovka: > Is there some method you use to find these issues?

Unfortunately, no. I have noticed it during reading your pull request. That is the reason why the reviews are so valuable. It is very common to overlook something during the implementation, so it is good to have a backup ;-)

>  Does anyone see any value in keeping the "I am a dictionary" behavior?

I think that the methods that we have now are sufficient. Btw. it looks like only the kdump addon uses `flags.cmdline`. The link to the upstream is https://github.com/daveyoung/kdump-anaconda-addon.

VladimirSlavik: I checked both kdump and oscap and I was unable to find any place where cmdline would be referenced, same as KernelArguments. I think that means case closed for them? Kdump only writes bootloader arguments, which is a different thing.

poncovka: https://github.com/daveyoung/kdump-anaconda-addon/blob/master/com_redhat_kdump/gui/spokes/kdump.py#L54

VladimirSlavik: Argh, it's a forked repo! I see where I made the mistake.

VladimirSlavik:
**Update**: Added tests, added docstrings, fixed all from the feedback, rebased.

pyanaconda/core/kernel.py
```diff
@@ -110,17 +115,17 @@ def read_string(self, cmdline):
...
-                        self[key] = []
-                    self[key].append(val)
+                    if not self._data.get(key, None):
+                        self._data[key] = []
+                    self._data[key].append(val)
             else:
-                self[key] = val
+                self._data[key] = val
 
     def getbool(self, arg, default=False):
```
poncovka: I wonder whether we shouldn't also get rid of the `getbool` method. We could merge it with `get` and use the logic with the converter from the function `get_option` in `pyanaconda.core.configuration.base`. The converter shouldn't be used if the return value is None (returned as a default).

jkonecny12: Interesting idea. Just thinking if that won't take part of the expressiveness. However, you probably know what you should expect from the boot option...

VladimirSlavik: It's actually easily possibile: There are just 3 uses of `getbool` in Anaconda (geoloc, fips). Most of the boot "nopes" are handled in argument_parsing explicitly with full name, so the magic for "no" prefix is actually mostly useless. Kdump uses `getbool` 3 times, but that's easily fixed too.

I propose removing all _use_ of `getbool` in Anaconda now, and addressing the eradication of external usage in a new work unit. Is that ok?

VladimirSlavik:
**UPDATE**: Changes to make `getbool()` unused in Anaconda. Feel free to reject independently of the rest.

pyanaconda/geoloc.py
```diff
@@ -192,8 +193,9 @@ def _check_if_geolocation_should_be_used(self, options_override):
                 geolocation_enabled = False
 
         # and also check if geolocation was not disabled by boot or command like option
-        if not flags.cmdline.getbool('geoloc', True):
-            # flags.cmdline.getbool is used as it handles values such as
+        if not kernel_arguments.get('geoloc', default=True, interpret_booleans=True) or \
```
poncovka: I would still prefer to call `kernel_arguments.get('geoloc', default=True, converter=bool)`, because it is a more universal solution. For example, we would be able to call `kernel_arguments.get('timeout', default=0, converter=int)` to retrieve a number from the kernel arguments. However, the current solution is all right, so I don't mind.

pyanaconda/geoloc.py
```diff
@@ -217,7 +219,8 @@ def _log_geolocation_status(self, geolocation_enabled, options_override):
                 log.info("Geolocation is disabled for image or directory installation.")
             elif flags.automatedInstall:
                 log.info("Geolocation is disabled due to automated kickstart based installation.")
-            if not flags.cmdline.getbool('geoloc', True):
+            if not kernel_arguments.get('geoloc', default=True, interpret_booleans=True) or \
+                    "nogeoloc" in kernel_arguments:
```
poncovka: Btw. I don't think that we should parse the kernel argument `geoloc` here. It is our option, so it should be handled during parsing kernel and cmdline arguments and merged with `opt.geoloc`. This is just wrong, but out of scope of this pull request.

pyanaconda/installation.py
```diff
@@ -318,7 +319,8 @@ def run_pre_install():
 
         if can_install_bootloader:
             payload.requirements.add_packages(storage.bootloader.packages, reason="bootloader")
-        if flags.flags.cmdline.getbool("fips"):
+        if kernel_arguments.get("fips", default=False, interpret_booleans=True) and \
+                "nofips" not in kernel_arguments :
```
poncovka: I couldn't found any proof any the `nofips` option is documented or used. It is always documented as `fips=1`.

pyanaconda/geoloc.py
```diff
@@ -192,8 +193,9 @@ def _check_if_geolocation_should_be_used(self, options_override):
                 geolocation_enabled = False
 
         # and also check if geolocation was not disabled by boot or command like option
-        if not flags.cmdline.getbool('geoloc', True):
-            # flags.cmdline.getbool is used as it handles values such as
+        if not kernel_arguments.get('geoloc', default=True, interpret_booleans=True) or \
+                "nogeoloc" in kernel_arguments:
```
poncovka: Actually, the option `nogeoloc` is not documented either. It is recommended to use  `inst.geoloc=0` to disable a geolocation.

pyanaconda/core/kernel.py
poncovka: Sounds good to me.

pyanaconda/geoloc.py
jkonecny12: I think it could be even more robust by finding bugs when the value is not a given type.

pyanaconda/geoloc.py
jkonecny12: Agree, let's remove this. If it is not documented we can use that as argument for removal.

VladimirSlavik: Removed.

pyanaconda/installation.py
VladimirSlavik: Removed.

VladimirSlavik:
**Update**:
* Removed the "noxxx" option handling as discussed in review comments
* Rebased

poncovka:
There is still an unresolved issue from https://github.com/rhinstaller/anaconda/pull/2213#discussion_r347348643. Otherwise, it looks good to me.

pyanaconda/geoloc.py
VladimirSlavik: I don't really agree. In hindsight, I should have argued this earlier. The core of my argument is that I don't like making the interface look as if it can just take any callable, then unexpected magic springs from around the corner, making booleans behave weirdly, when the syntax is the same as for any other callables.

The legacy rules for parsing "bools" are "no/off/0" || not present VS any other case. Let's compare with the other place you suggested which is based on `configparser.RawConfigParser.BOOLEAN_STATES`: "false" is *not* going to give False, and presence is interpreted too. So these are not consistent. 

I have no other word other than *magical* for the behavior that would result from making `bool` get the magic in place of a callable. The method signature implies a `bool` that could be interchangeable with `int`, except it does something completely unexpected. So I don't want to make it look like you can just plug in anything with equal results and rules, because you can *not*.

If I were to go to the full extent of my argument, then even adding the parameter there (as I eventually did) makes no sense - containing the magic is best done by letting it have its own clearly separate method. It's not like the calls do not have to be different anyway:  `get(..., interpret_booleans=True)` != `getbool(...)` != `get(..., converter=bool)`. People will have to type different things anyway, and the legacy logic for the magic must be retained too, so we are not getting better in any metric.

Perhaps that's too subtle, but it is also the reason I went with "interpret" instead of "convert" in the parameter name as it (to me) implies more processing.

As to callables. The only advantage allowed by using a callable parameter is that one could add their own conversion later, but one can do `get(...,converter=int)` equally well as just `int(get(...))`. So why? Callable is a callable, whether in or out of the method signature. And making it so for a mere *possibility* of somebody using a converter in the future seems a premature optimization.

...so yes, I am convinced the converter argument in the other place is not good either.

I don't want to make it look like I care too much about this - the previous state is suboptimal, I don't see how the next one could be much better, so if you think defaults+callables+magic in single method is the way to go, I am not saying *never!* It's just, I don't see any argument for doing it that way. So I expect a convincing argument "why".

VladimirSlavik: @jkonecny12 @M4rtinK @rvykydal - After offline discussion with @poncovka, we decided the way forward is: Remove `interpret_booleans` parameter, and use a new method `is_enabled(str)` instead. We'll also drop the yes/no and keep only 0/1 & on/off.

VladimirSlavik:
Update: Exchanged adding `interpret_booleans` with `is_enabled()` per discussion above.

VladimirSlavik:
Thank you too!

PR #2217, opened 2019-11-14
Keyboard move logic to module 1

rvykydal:
This is first phase of moving keyboard logic (including LocaledWrapper) into keyboard module.
Posted mainly for review of separation into tasks and API before I move furhter.
Next steps:
- write tests
- perhaps polish moved content
- replace LocaledWrapper and safe_dbus with something wrapping systemd-localed service using dasbus, robust enough to survive in mock (environments without the service).


rvykydal:
Okay, let me resolve the conflicts firs.

pyanaconda/modules/localization/runtime.py
```diff
@@ -69,3 +75,104 @@ def run(self):
             x_layouts = localed.convert_keymap(vc_keymap)
 
         return x_layouts, vc_keymap
+
+
+class ApplyKeyboardTask(Task):
+    """Task for applying keyboard settings to current system."""
+
+    def __init__(self, keyboard, x_layouts, vc_keymap, switch_options):
+        """Create a new task,
```
jkonecny12: You should use `.` instead of `,` here.

pyanaconda/modules/localization/installation.py
```diff
@@ -97,102 +102,106 @@ def name(self):
...
 
-        if root != "/":
-            # writing to a different root, we need to save these values, so that
-            # we can restore them when we have the file written out
-            layouts_variants = localed_wrapper.layouts_variants
-            options = localed_wrapper.options
+        # copy the file to the chroot
+        try:
+            shutil.copy2(xconf_file_path,
+                            os.path.normpath(root + "/" + xconf_file_path))
```
jkonecny12: Bad indentation.

pyanaconda/modules/localization/localed.py
```diff
@@ -0,0 +1,253 @@
...
+    def set_keymap(self, keymap, convert=False):
+        """
+        Method that sets VConsole keymap via systemd-localed's DBus API.
+
+        :param keymap: VConsole keymap that should be set
+        :type keymap: str
+        :param convert: whether the keymap should be converted to a X11 layout
+                        (see set_and_convert_keymap)
+        :type convert: bool
+
```
jkonecny12: Please remove this empty lines at the end of docstrings.

pyanaconda/modules/localization/installation.py
```diff
@@ -58,3 +70,139 @@ def _write_language_configuration(self, lang, root):
...
+    if root != "/":
+        # writing to a different root, we need to save these values, so that
+        # we can restore them when we have the file written out
+        layouts_variants = localed_wrapper.layouts_variants
+        options = localed_wrapper.options
+
+        # set systemd-localed's layouts, variants and switch options, which
+        # also generates a new conf file
+        localed_wrapper.set_layouts(x_layouts,
+                                    switch_options)
```
jkonecny12: This could be one line.

pyanaconda/modules/localization/installation.py
```diff
@@ -58,3 +70,139 @@ def _write_language_configuration(self, lang, root):
...
+        try:
+            shutil.copy2(xconf_file_path,
+                            os.path.normpath(root + "/" + xconf_file_path))
+        except IOError:
+            # The file may not exist (eg. text install) so don't raise
+            pass
+
+        # restore the original values
+        localed_wrapper.set_layouts(layouts_variants,
+                                    options)
```
jkonecny12: This could also be one line.

pyanaconda/modules/localization/installation.py
```diff
@@ -58,3 +70,139 @@ def _write_language_configuration(self, lang, root):
...
+                os.makedirs(rooted_xconf_dir)
+        except OSError:
+            errors.append("Cannot create directory {}".format(rooted_xconf_dir))
+
+        # copy the file to the chroot
+        try:
+            shutil.copy2(xconf_file_path,
+                            os.path.normpath(root + "/" + xconf_file_path))
+        except IOError:
+            # The file may not exist (eg. text install) so don't raise
```
jkonecny12: Don't you think it could be helpful to have debug output. I think it could be handy to know that something is missing even though it's not required.

pyanaconda/modules/localization/installation.py
```diff
@@ -58,3 +70,139 @@ def _write_language_configuration(self, lang, root):
...
+    xconf_file_path = os.path.normpath(X_CONF_DIR + "/" + X_CONF_FILE_NAME)
+    errors = []
+
+    try:
+        if not os.path.isdir(X_CONF_DIR):
+            os.makedirs(X_CONF_DIR)
+    except OSError:
+        errors.append("Cannot create directory {}".format(X_CONF_DIR))
+
+    if root != "/":
```
jkonecny12: These large ``if-else`` should be abstracted, thought.

pyanaconda/modules/localization/installation.py
```diff
@@ -16,9 +16,21 @@
...
+from pyanaconda.core.constants import DEFAULT_VC_FONT, DEFAULT_KEYBOARD
+from pyanaconda.keyboard import InvalidLayoutVariantSpec
+from pyanaconda.modules.localization.localed import LocaledWrapper
+from pyanaconda.anaconda_loggers import get_module_logger
+
+log = get_module_logger(__name__)
+
+X_CONF_DIR = "/etc/X11/xorg.conf.d"
+X_CONF_FILE_NAME = "00-keyboard.conf"
+VC_CONF_FILE_PATH = "/etc/vconsole.conf"
```
poncovka: These values are defined also in `KeyboardInstallationTask`.

pyanaconda/modules/localization/runtime.py
```diff
@@ -0,0 +1,178 @@
...
+from pyanaconda.modules.common.task import Task
+from pyanaconda.modules.localization.localed import LocaledWrapper
+from pyanaconda.anaconda_loggers import get_module_logger
+from pyanaconda.modules.localization.installation import write_x_configuration, \
+    write_vc_configuration
+
+log = get_module_logger(__name__)
+
+
+class ConvertMissingKeyboardConfigurationTask(Task):
```
poncovka: This task is a little confusing. I wonder if the logic could be somehow automated using signals of the setters.

pyanaconda/modules/localization/runtime.py
```diff
@@ -0,0 +1,178 @@
...
+        if new_vc_keymap:
+            vc_keymap = new_vc_keymap
+
+        if not x_layouts:
+            x_layouts = localed.convert_keymap(vc_keymap)
+
+        return x_layouts, vc_keymap
+
+
+class ApplyKeyboardTask(Task):
```
poncovka: I would try to move to this task the whole keyboard configuration from anaconda.py:

```

    configured = any((localization_proxy.Keyboard,
                      localization_proxy.VirtualConsoleKeymap,
                      localization_proxy.XLayouts))

    if opts.keymap and not configured:
        localization_proxy.SetKeyboard(opts.keymap)
        configured = True

    if configured:
        if conf.system.can_activate_keyboard:
            keyboard.activate_keyboard(localization_proxy)
        else:
            # at least make sure we have all the values
            keyboard.populate_missing_items(localization_proxy)
```

pyanaconda/installation.py
```diff
@@ -100,15 +100,17 @@ def _prepare_configuration(storage, payload, ksdata):
...
-
     # add installation tasks for the Timezone DBus module
     timezone_proxy = TIMEZONE.get_proxy()
     timezone_dbus_tasks = timezone_proxy.InstallWithTasks()
     os_config.append_dbus_tasks(TIMEZONE, timezone_dbus_tasks)
 
-    # add installation tasks for the Localization DBus module
+    # get missing keyboard values
     localization_proxy = LOCALIZATION.get_proxy()
+    convert_task = localization_proxy.ConvertMissingKeyboardConfigurationWithTask()
```
poncovka: I would include this task in `InstallWithTasks` if possible.

pyanaconda/modules/localization/runtime.py
rvykydal: Fixed

pyanaconda/modules/localization/installation.py
rvykydal: Fixed

pyanaconda/modules/localization/localed.py
rvykydal: Fixed

pyanaconda/modules/localization/installation.py
rvykydal: Fixed

pyanaconda/modules/localization/installation.py
rvykydal: Added debug message

pyanaconda/modules/localization/installation.py
rvykydal: It is not so easy.
This belongs to the point
- perhaps polish moved content

from my initial comment.

I am starting to thing that the two usages of the functions (runtime, chroot) should use separate functions, they share only a little of code and behave quite differently (eg root="/" version is using default). I may come with follow-up patch for this, maybe in scope of this PR, maybe not because I didn't want to introduce any non-necessary bugs/regressions caused by refactoring while moving (this PR).

pyanaconda/modules/localization/installation.py
rvykydal: Removed form the KeyboardInstallationTask

pyanaconda/modules/localization/installation.py
jkonecny12: agree, lets go with the current solution for this PR

rvykydal:
I've added 7 patches, 2 of them fixing Rawhide BZs I filed recently.

pyanaconda/modules/localization/runtime.py
rvykydal: After splitting missing values handling https://github.com/rhinstaller/anaconda/pull/2217/commits/69d12b4a6edd6059b410ab579a93c56a14f59841
this task becomes PopulateMissingKeyboardConfiguration and does only x_layouts <-> vc_keymap update of missing values. I think such update should be explicit to module client, especially when the task is still used as installation task (see below).

pyanaconda/modules/localization/runtime.py
rvykydal: After looking at the code I still think we want Apply and PopulateMissing API used by client. It is possible that the latter will be dropped? Or even handled by the client.

pyanaconda/installation.py
rvykydal: Modified to PopulateMissing... task. It is used for example when nothing is set (by ks or boot options) and x_layout is inferred from lang. In this case the task updates vc_keymap. I think we could remove this task when looking at updating of the values from lang. One thing to consider is that this update seems to happen only from GUI. It is for another PR.

rvykydal:
In scope of this PR I need to add tests for the changes.

pyanaconda/modules/localization/localization.py
```diff
@@ -200,5 +191,65 @@ def install_with_tasks(self):
...
+            self.vc_keymap = DEFAULT_KEYBOARD
+
+        task = ConvertMissingKeyboardConfigurationTask(
+            x_layouts=self.x_layouts,
+            vc_keymap=self.vc_keymap,
+        )
+        task.succeeded_signal.connect(lambda: self.update_settings_from_task(task.get_result()))
+        return task
+
+    def update_settings_from_task(self, result):
```
poncovka: I think that this method could be private.

pyanaconda/modules/localization/localization.py
```diff
@@ -200,5 +191,65 @@ def install_with_tasks(self):
...
+            )
         ]
+
+    def populate_missing_keyboard_configuration_with_task(self):
+        """Get missing keyboard configuration by conversion.
+
+        :returns: a task converting the configuration
+        """
+        if not self.x_layouts and not self.vc_keymap:
+            self.vc_keymap = DEFAULT_KEYBOARD
```
poncovka: Shouldn't we wait and set this property later with `update_settings_from_task`?

pyanaconda/modules/localization/localization.py
```diff
@@ -200,5 +191,65 @@ def install_with_tasks(self):
...
+        """
+        log.debug("Setting keyboard from generic setting value '%s'.", keyboard)
+        if self.vc_keymap or self.x_layouts:
+            log.debug("Ignoring generic keyboard setting as we have a specific one.")
+            return
+
+        task = AssignGenericKeyboardSettingTask(
+            keyboard=keyboard
+        )
+        x_layouts, vc_keymap = task.run()
```
poncovka: You can call `update_settings_from_task` for the result of the task.

pyanaconda/modules/localization/localization_interface.py
```diff
@@ -164,3 +156,31 @@ def KeyboardKickstarted(self) -> Bool:
         :return: True if keyboard command was seen in kickstart, otherwise False
         """
         return self.implementation.keyboard_seen
+
+    @emits_properties_changed
+    def SetKeyboardKickstarted(self, keyboard_seen: Bool):
```
poncovka: Just a note, I would like to eventually get rid of the `Kickstarted` properties.

rvykydal: Yes, for now it became kind of consistent with locale settings of the localization module.

pyanaconda/modules/localization/localization.py
rvykydal: https://github.com/rhinstaller/anaconda/pull/2217/commits/ca85dad74ca838176b8f128e02dec24c0b2da0e3

pyanaconda/modules/localization/localization.py
rvykydal: https://github.com/rhinstaller/anaconda/pull/2217/commits/ca85dad74ca838176b8f128e02dec24c0b2da0e3

pyanaconda/modules/localization/localization.py
rvykydal: Following on our discussion, finally I decided to move the default setting into the task:
https://github.com/rhinstaller/anaconda/pull/2217/commits/d462bc9a70b5db294521a371ae988a46747f63c8

rvykydal:
Added tests, some fixups, (mostly discovered when working on the tests).
I will add a test for write_x_configuration in another PR (probably refactoring the function itself).

pyanaconda/installation.py
```diff
@@ -100,15 +100,17 @@ def _prepare_configuration(storage, payload, ksdata):
...
-
     # add installation tasks for the Timezone DBus module
     timezone_proxy = TIMEZONE.get_proxy()
     timezone_dbus_tasks = timezone_proxy.InstallWithTasks()
     os_config.append_dbus_tasks(TIMEZONE, timezone_dbus_tasks)
 
-    # add installation tasks for the Localization DBus module
+    # get missing keyboard values
     localization_proxy = LOCALIZATION.get_proxy()
+    convert_task = localization_proxy.PopulateMissingKeyboardConfigurationWithTask()
```
poncovka: I think that this method will have to be called in the initial setup.

rvykydal: https://github.com/rhinstaller/initial-setup/pull/89

PR #2218, opened 2019-11-14
RHEL install time registration, CDN install source and Insights configuration support

M4rtinK:
**Commit structure**
The commit structure has been selected for easy review & feedback incorporation, merge time commit structure will most likely be different:

- changes outside of DBus modules
- changes in DBus modules
- changes in payload
- changes in GUI

+ a number of self contained smaller commits that should over time grow in number, possibly replacing the bigger commits over time

Looking forward to your feedback! :)

**TODO**

- ~HTTP proxy support in kickstart~
- ~registration/unregistration & attach should run in a thread~
- ~nicer Source spoke integration (similar to HMC)~
- ~default to Red Hat CDN installation source~
- ~put back the Register button  (based on user feedback)~
- ~show attached entitlements~
- ~implement network connectivity check~
- ~add correct Requires for Subscription Manager packages to the spec file~
- ~fix insights client detection on the target system~
- ~subscription status tab wonky after failed manual register followed by successful register~
- ~software spoke status "Nothing selected" after manual registration~
- ~RHSM service returns error messages in Japanese for non en (cs_cz for example) locale (!?)~ was likely an intermittent backend issue
- ~make sure the checked-by-default Insights checkbox in the GUI is not overridden by kickstart not mentioning Insights~
- ~proxy URL in GUI should have placeholder showing HOSTNAME:PORT syntax~
- ~get rid of remaining debug log messages~
- ~make sure repo set via kickstart overrides the default Red Hat CDN installation source~
- ~check things work with SSL verification enabled~
- ~handle all TODO & FIXME~

**FOLLOWUP PR STUFF**

- add check for subscription tokens being generated to attach task (jstodola discovered that some accounts can fail to attach even without an exception being raised, yay! :P)
- improve handling of custom server hostname & custom baseurl with & without protocol specification (http://, https://)
- improve start & end date display & coordinate with RHSM Cockpit plugin on the result
- check why Pykickstart rhsm command connect_to_insights property seems to be None if not set, instead of False
- check how CDN + custom additional repos behave
- check if Source spoke integration could be improved code wise
- improve how we handled error reporting from Subscription DBus tasks
- improve how we handle exceptions raised by the ConnectToInsights task
- plumb Anaconda debugging level to RHSM
- split non-API code from Subscription module to a separate file

Pykickstart PR adding the **rhsm** command:
https://github.com/pykickstart/pykickstart/pull/290
Merged.


M4rtinK:
jenkins, test this please

pyanaconda/modules/subscription/subscription.py
```diff
@@ -17,24 +17,58 @@
...
+    """Representation of a RHSM private bus connection."""
+
+    def __init__(self, rhsm_private_bus_address):
+        super().__init__()
+        self._rhsm_private_bus_address = rhsm_private_bus_address
+
+    def get_new_connection(self):
+        """Connect to the RHSM private bus."""
+        log.info("Connecting to RHSM private bus at %s.", self._rhsm_private_bus_address)
+        connection = pydbus.bus.Gio.DBusConnection.new_for_address_sync(self._rhsm_private_bus_address,
```
jkonecny12: I think we have a better solution than using `pydbus` directly. @poncovka could you please look on this?

pyanaconda/modules/subscription/subscription.py
```diff
@@ -242,5 +441,438 @@ def set_system_purpose_with_task(self, sysroot):
...
+
+    def install_with_tasks(self, sysroot):
+        # subscription token transfer
+        token_transfer_task = TransferSubscriptionTokensTask(sysroot, self.entitlement_attached)
+        token_transfer_task_path = self.publish_task(SUBSCRIPTION.namespace, token_transfer_task)
+
+        # insights
+        insights_connect_task = ConnectToInsightsTask(sysroot, self.connect_to_insights)
+        insights_connect_task_path = self.publish_task(SUBSCRIPTION.namespace, insights_connect_task)
+        return [token_transfer_task_path, insights_connect_task_path]
```
jkonecny12: The `subscription.py` is too big. It should definitely be broken into smaller pieces for better maintenance. First step would be to separate RHSM service connection to a separate file.

In general the module main service class should be as small as possible and should not contain almost no logic.

pyanaconda/modules/subscription/installation.py
```diff
@@ -15,32 +15,446 @@
...
-                                                 self._addons)
+                                               self._role,
+                                               self._sla,
+                                               self._usage,
+                                               self._addons)
+        self._done()
+
+    @async_action_nowait
+    def _done(self):
+        # tell nayone listening that we are done
```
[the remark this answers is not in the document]
M4rtinK: Fixed, thanks! :)

pyanaconda/modules/subscription/installation.py
```diff
@@ -15,32 +15,446 @@
...
+        self.succeeded.emit()
+
+
+class StartRHSMTask(Task):
+    """Start the RHSM DBUS service.
+
+    TODO: this might not be needed once the RHSM DBUS services
+          is a regular part of the installation environment
```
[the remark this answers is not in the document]
M4rtinK: This was needed mostly for updates image based development, but I now have custom boot ISOs, so it should be now possible to verify if this is needed in practice. Still, I think we still can drop this only in the final review round or else this will simply not work if an updates image is used.

pyanaconda/modules/subscription/installation.py
```diff
@@ -15,32 +15,446 @@
...
+        config_proxy = RHSM.get_proxy(RHSM_CONFIG)
+        config_proxy.Set("server.hostname", self._subscription_url, "")
+
+        # restart the service
+        log.debug("start-rhsm-task: restarting RHSM service")
+        util.restart_service("rhsm.service")
+        log.debug("start-rhsm-task: RHSM service should be now ready for use")
+
+
+class RegisterWithUsernamePasswordTask(Task):
```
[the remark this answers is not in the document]
M4rtinK: Well, the organization name + activation key are also credentials, so I think it could be ambiguous. Still, the current name is indeed not the best. So maybe `RegisterWithPasswordTask` ? And the same thing for the other one - `RegisterWithActivationKeyTask` ?

pyanaconda/modules/subscription/installation.py
```diff
@@ -15,32 +15,446 @@
...
+        return "Register with organization name and activation key"
+
+    def run(self):
+        """Register the system with organization name and activation key."""
+        log.debug("RHSM: connecting to RHSM private bus")
+        private_connection = RHSMPrivateBusConnection(self._rhsm_private_bus_address)
+        private_register_proxy = private_connection.get_proxy("com.redhat.RHSM1","/com/redhat/RHSM1/Register")
+        log.debug("RHSM: registering with organization and activation key")
+        try:
+            private_register_proxy.RegisterWithActivationKeys(self._organization, [self._activation_key], {}, {}, "")
```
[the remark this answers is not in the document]
M4rtinK: Shortened.

pyanaconda/modules/subscription/installation.py
```diff
@@ -15,32 +15,446 @@
...
+        if not self._copy_file_to_path(self.RHSM_CONFIG_FILE_PATH, target_rhsm_config_path):
+            msg = "RHSM config file (%s) is missing." % self.RHSM_CONFIG_FILE_PATH
+            raise SubscriptionTokenTransferError(msg)
+
+        # transfer the consumer key
+        log.debug("transfer-subscription-task: transferring consumer key")
+        target_consumer_key_path = self._sysroot + self.RHSM_CONSUMER_KEY_PATH
+        if not self._copy_file_to_path(self.RHSM_CONSUMER_KEY_PATH, target_consumer_key_path):
+            msg = "RHSM consumer key (%s) is missing." % self.RHSM_CONSUMER_KEY_PATH
+            raise SubscriptionTokenTransferError(msg)
```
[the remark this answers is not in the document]
M4rtinK: Hmm, that might work & might make unit testing easier. Lets do that.

pyanaconda/modules/common/constants/namespaces.py
```diff
@@ -90,3 +90,9 @@
     *ADDONS_NAMESPACE,
     "Baz"
 )
+
+## System service namespaces ##
+
+RHSM_NAMESPACE = (
+    "com", "redhat", "RHSM1"
```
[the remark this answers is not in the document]
M4rtinK: We are certainly calling stuff over DBus, so what's the correct way to do this ? @poncovka

M4rtinK:
@jkonecny12 This should certainly be split to smaller commits eventually, but the current commit structure is IMHO quite necessary for quick feedback incorporation.

pyanaconda/modules/subscription/subscription.py
M4rtinK: Hmm, yeah, I think we should move the "helper" non API functions to a separate file in the module directory. Looking on the `master` Anaconda that is definitely farther module wise than what we have currently on RHEL, there are couple such helper files in some modules.

pyanaconda/ui/gui/spokes/subscription.py
```diff
@@ -0,0 +1,787 @@
...
+
+    @property
+    def organization(self):
+        """Organization name."""
+        return self._organization_entry.get_text()
+
+    @organization.setter
+    def organization(self, organization):
+        self._organization_entry.set_text(organization)
+        self.authentication_data_changed.emit()
```
poncovka: I think that you could have a callback that will emit the `authentication_data_changed` signal when a widget is changed and register the callback in the `.glade` file in all related widgets. This way we have about 200-300 lines of boilerplate.

pyanaconda/ui/gui/spokes/subscription.py
```diff
@@ -0,0 +1,787 @@
...
+        # try to register
+        log.debug("Subscription GUI: attempting to register")
+        task_path = self._subscription_module.proxy.RegisterWithTask()
+        task_proxy = SUBSCRIPTION.get_proxy(task_path)
+        sync_run_task(task_proxy)
+
+        # TODO: error handling
+
+        # try to attach subscription
+        log.debug("Subscription GUI: attempting to attach attach an entitlement")
```
poncovka: ...attach attach...

pyanaconda/modules/subscription/subscription.py
```diff
@@ -242,5 +441,438 @@ def set_system_purpose_with_task(self, sysroot):
...
         task = SystemPurposeConfigurationTask(sysroot, self.role, self.sla, self.usage, self.addons)
+        # set system purpose as applied once the task finishes running
+        task.stopped_signal.connect(lambda: self._set_is_system_purpose_applied(True))
         path = self.publish_task(SUBSCRIPTION.namespace, task)
         return path
+
+    # subscription
+
+    @property
+    def registered(self):
```
poncovka: Does the property describe the state of the installation environment or does it describe the state of the DBus module? The question applies also for `entitlement_attached` and `is_system_purpose_applied`. During the backport, we might need to reconsider the design of this module if the properties describe its state.

pyanaconda/modules/subscription/installation.py
```diff
@@ -15,32 +15,466 @@
...
+            log.debug("RHSM: registered with username and password")
+        except:
+            # FIXME: catch a concrete exception
+            log.exception("RHSM: failed to register with username and password")
+            msg = "Registration with username & password failed."
+            raise PasswordRegistrationError(msg)
+
+
+class RegisterWithOrganizationKeyTask(Task):
+    """Register the system via username + password."""
```
rvykydal: copy and paste leftover?

pyanaconda/modules/subscription/installation.py
```diff
@@ -15,32 +15,466 @@
...
+        if not os.path.isfile(file_path):
+            return False
+        if not os.path.isdir(os.path.dirname(target_file_path)):
+            util.mkdirChain(os.path.dirname(target_file_path))
+        shutil.copy(file_path, target_file_path)
+        return True
+
+    def _transfer_system_purpose(self):
+        # transfer the system purpose file if present
+        # - this might be needed even if the system has not bee subscribed
```
rvykydal: bee -> been

pyanaconda/modules/subscription/installation.py
```diff
@@ -15,32 +15,466 @@
...
+        #   during the installation and is thereofre always attempted
```
rvykydal: thereofre -> therefore

pyanaconda/modules/subscription/subscription.py
```diff
@@ -233,6 +403,29 @@ def is_system_purpose_set(self):
...
+        environment will be transferred to the target system.
+        """
+        return self._is_system_purpose_applied
+
+    def _set_is_system_purpose_applied(self, system_purpose_applied):
+        """Set if system purpose is applied.
+
+        :param bool system_purpose_applied: True if applied, False otherwise
+
+        NOTE: We keep this as a private method, called the completed signal of the
```
rvykydal: called the -> called by the ?

pyanaconda/modules/subscription/subscription.py
```diff
@@ -242,5 +435,437 @@ def set_system_purpose_with_task(self, sysroot):
...
+        return self._registered
+
+    @registered.setter
+    def registered(self, system_registered):
+        """Set that the system has been registered.
+
+        :param bool system_registered: system registered state
+        """
+        self._registered = system_registered
+        self.registered_changed.emit()
```
rvykydal: Do you want to log setting of the property similarly to the entitlement_attached.setter ?

pyanaconda/modules/subscription/subscription.py
```diff
@@ -242,5 +435,437 @@ def set_system_purpose_with_task(self, sysroot):
...
+        self._red_hat_account_password = password
+        self.red_hat_account_password_changed.emit()
+        log.debug("Red Hat account password has been set.")
+
+    @property
+    def authentication_method(self):
+        """Return what authentication method has been selected (if any)."""
+        return self._authentication_method
+
+    def set_authentication_method(self, method):
```
rvykydal: Missing docstring?

pyanaconda/modules/subscription/subscription.py
```diff
@@ -242,5 +435,437 @@ def set_system_purpose_with_task(self, sysroot):
...
+
+        :param str password: Red Hat account password
+        """
+        self._red_hat_account_password = password
+        self.red_hat_account_password_changed.emit()
+        log.debug("Red Hat account password has been set.")
+
+    @property
+    def authentication_method(self):
+        """Return what authentication method has been selected (if any)."""
```
rvykydal: Missing return value tags in docstring?

pyanaconda/modules/subscription/subscription.py
```diff
@@ -242,5 +435,437 @@ def set_system_purpose_with_task(self, sysroot):
...
+        :rtype: bool
+        """
+        return bool(self._server_proxy_password)
+
+    # pylint: disable=too-many-function-args
+    def set_server_proxy(self, hostname, port, username, password):
+        """Set configuration of the RHSM HTTP proxy.
+
+        :param str hostname: RHSM HTTP proxy hostname
+        :param int post: RHSM HTTP proxy port (set to -1 to clear)
```
rvykydal: post -> port

pyanaconda/modules/subscription/subscription.py
```diff
@@ -242,5 +435,437 @@ def set_system_purpose_with_task(self, sysroot):
...
+        else:
+            port_string = "{}".format(port)
+        # set data to RHSM
+        self.rhsm_config_proxy.Set("server.proxy_hostname", get_variant(Str, hostname), "")
+        self.rhsm_config_proxy.Set("server.proxy_port", get_variant(Str, port_string, ""))
+        self.rhsm_config_proxy.Set("server.proxy_user", get_variant(Str, username), "")
+        self.rhsm_config_proxy.Set("server.proxy_password", get_variant(Str, password), "")
+
+        # trigger the changed signal
+        self.server_proxy_configuration_changed.emit()
```
rvykydal: Do we want to log this change ?

pyanaconda/modules/subscription/installation.py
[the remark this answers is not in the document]
jkonecny12: Seems like a good solution.

pyanaconda/modules/subscription/installation.py
jkonecny12: is that really a typo? :D

pyanaconda/installclasses/rhel.py
```diff
@@ -50,7 +50,7 @@ class RHELBaseInstallClass(BaseInstallClass):
 
     eula_path="/usr/share/redhat-release/EULA"
 
-    mirrors_available = False
+    mirrors_available = True
```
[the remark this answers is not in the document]
M4rtinK: Dropped with proper CDN source support.

pyanaconda/ui/gui/spokes/installation_source.glade
```diff
@@ -772,7 +772,7 @@
                                           <item id="https">https://</item>
                                           <item id="ftp">ftp://</item>
                                           <item id="nfs">nfs</item>
-                                          <item id="Closest mirror" translatable="yes">Closest mirror</item>
+                                          <item id="Closest mirror" translatable="yes">Red Hat CDN</item>
```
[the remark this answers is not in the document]
M4rtinK: Replaced with proper CDN source support.

pyanaconda/installation.py
```diff
@@ -334,13 +335,21 @@ def run_pre_install():
...
         subscription_proxy = SUBSCRIPTION.get_proxy()
-        if subscription_proxy.IsSystemPurposeSet:
-            # at least one value has been set, so we need the utility being installed in the target system
-            # chroot so we can call it
+        # check if we need the syspurpose package needed for system purpose support
+        if subscription_proxy.IsSystemPurposeSet and not subscription_proxy.IsSystemPurposeApplied:
+            # The system purpose installation task (which needs the syspurpose utility to be installed) runs:
+            # - if system purpose configuration has been requested
+            # - but system purpose has not been applied during the installation
+            # Onl in such a case it will run on the target system and needs the syspurpose utility to be available.
```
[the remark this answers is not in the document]
M4rtinK: Fixed, thanks! :)

pyanaconda/kickstart.py
```diff
@@ -980,9 +981,14 @@ def __str__(self):
 
     def execute(self, *args, **kwargs):
         subscription_proxy = SUBSCRIPTION.get_proxy()
-        task_path = subscription_proxy.SetSystemPurposeWithTask(util.getSysroot())
-        task_proxy = SUBSCRIPTION.get_proxy(task_path)
-        sync_run_task(task_proxy)
+        log.debug("KICKSTART EXECUTE")
```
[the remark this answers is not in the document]
M4rtinK: No longer there.

pyanaconda/kickstart.py
```diff
@@ -2196,6 +2202,33 @@ def execute(self, storage, ksdata, instClass, users):
             except ValueError as e:
                 user_log.warning(str(e))
 
+class RHSM(RemovedCommand):
+
+    def __str__(self):
+        subscription_proxy = SUBSCRIPTION.get_proxy()
+        return subscription_proxy.GenerateKickstart()
+
+    def __parse(self, ksdata):
```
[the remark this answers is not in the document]
M4rtinK: Dropped.

pyanaconda/kickstart.py
```diff
@@ -2196,6 +2202,33 @@ def execute(self, storage, ksdata, instClass, users):
...
+        sync_run_task(task_proxy)
+
+        # next try to register if we have enough data available from kickstart
+        if subscription_proxy.Username and subscription_proxy.IsPasswordSet:
+            rhsm_log.debug("RHSM kickstart SETUP - RHSM registration task")
+            task_path = subscription_proxy.StartRHSMWithTask()
+            task_proxy = SUBSCRIPTION.get_proxy(task_path)
+            sync_run_task(task_proxy)
+
+    def execute(self, storage, ksdata, instClass, users):
```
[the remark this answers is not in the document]
M4rtinK: Dropped.

pyanaconda/modules/common/errors/installation.py
```diff
@@ -37,3 +37,19 @@ class InstallationNotRunning(InstallationError):
...
+    pass
+
+
+@dbus_error("InsightsClientMissingError", namespace=ANACONDA_NAMESPACE)
+class InsightsClientMissingError(InstallationError):
+    """Exception for missing Red Hat Insights utility."""
+    pass
+
+@dbus_error("InsightsConnectError", namespace=ANACONDA_NAMESPACE)
+class InsightsConnectError(InstallationError):
```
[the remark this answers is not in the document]
M4rtinK: They do happen at install time, as Insights is configured in the post installation phase in the target system chroot, so I guess they are installation errors.

pyanaconda/modules/common/errors/installation.py
```diff
@@ -37,3 +37,19 @@ class InstallationNotRunning(InstallationError):
 class LanguageInstallationError(InstallationError):
     """Exception for the language installation errors."""
     pass
+
```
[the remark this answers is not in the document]
M4rtinK: Indeed! Fixed. :)

pyanaconda/modules/subscription/installation.py
```diff
@@ -15,32 +15,446 @@
...
+
 class SystemPurposeConfigurationTask(Task):
     """Installation task for setting system purpose."""
 
     def __init__(self, sysroot, role, sla, usage, addons):
         """Create a new task.
-
-        :param sysroot: a path to the root of the installed system
-        :param lang: a value for LANG locale variable
+        :param str sysroot: a path to the root of the installed system
```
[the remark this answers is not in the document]
M4rtinK: Good point, docstrings added. :)

pyanaconda/modules/subscription/installation.py
```diff
@@ -15,32 +15,446 @@
...
         self._usage = usage
         self._addons = addons
+        self.succeeded = Signal()
 
     @property
     def name(self):
         return "Set system purpose"
 
     def run(self):
         system_purpose.give_the_system_purpose(self._sysroot,
```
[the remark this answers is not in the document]
M4rtinK: Done.

pyanaconda/modules/subscription/installation.py
```diff
@@ -15,32 +15,446 @@
...
+        if rc:
+            raise InsightsConnectError("Connecting to Red Hat Insights failed.")
+
+
+
+
+
+
+
+
```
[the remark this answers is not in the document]
M4rtinK: Done.

pyanaconda/modules/subscription/subscription.py
```diff
@@ -17,24 +17,58 @@
...
+from pyanaconda.modules.subscription.installation import TransferSubscriptionTokensTask
+from pyanaconda.modules.subscription.installation import ConnectToInsightsTask
 from pyanaconda.modules.subscription import system_purpose
 
 from pyanaconda.anaconda_loggers import get_module_logger
 log = get_module_logger(__name__)
 
 
+# FIXME: only leave this in installation.py ?
+class RHSMPrivateBusConnection(Connection):
```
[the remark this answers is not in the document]
M4rtinK: No longer needed in this file.

pyanaconda/modules/subscription/subscription.py
```diff
@@ -242,5 +441,438 @@ def set_system_purpose_with_task(self, sysroot):
...
+        Organization name is needed when using an activation key
+        and is not needed when registering via Red Hat account
+        credentials.
+
+        :return: organization name
+        :rtype: str
+        """
+        return self._organization
+
+    @organization.setter
```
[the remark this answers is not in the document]
M4rtinK: Oh, right - fixed.

pyanaconda/modules/subscription/subscription.py
```diff
@@ -242,5 +441,438 @@ def set_system_purpose_with_task(self, sysroot):
...
+        # set data to RHSM
+        self.rhsm_config_proxy.Set("server.proxy_hostname", get_variant(Str, hostname), "")
+        self.rhsm_config_proxy.Set("server.proxy_port", get_variant(Str, port_string, ""))
+        self.rhsm_config_proxy.Set("server.proxy_user", get_variant(Str, username), "")
+        self.rhsm_config_proxy.Set("server.proxy_password", get_variant(Str, password), "")
+
+        # trigger the changed signal
+        self.server_proxy_configuration_changed.emit()
+
+    def start_rhsm_with_task(self):
```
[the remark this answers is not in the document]
M4rtinK: Dropped.

pyanaconda/modules/subscription/subscription.py
```diff
@@ -242,5 +441,438 @@ def set_system_purpose_with_task(self, sysroot):
...
+        self._server_proxy_port = port
+        self._server_proxy_user = username
+        self._server_proxy_password = password
+        # negative port number means no port has been set
+        if port < 0:
+            port_string = ""
+        else:
+            port_string = "{}".format(port)
+        # set data to RHSM
+        self.rhsm_config_proxy.Set("server.proxy_hostname", get_variant(Str, hostname), "")
```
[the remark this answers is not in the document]
M4rtinK: The last parameter is for setting locale for RHSM output (mainly for error messages I think). While we do use this in some of the installation tasks, as the results are user visible, I this case I think we should keep the default in place. Otherwise, RHSM error messages in logs would be translated to currently selected language and potentially unreadable as a result.

pyanaconda/modules/subscription/subscription.py
```diff
@@ -242,5 +441,438 @@ def set_system_purpose_with_task(self, sysroot):
...
+        :param bool system_purpose_applied: True if attached, False otherwise
+
+        NOTE: We keep this as a private method, called the completed signal of the
+              tasks that attach entitlements or unregister the system.
+        """
+        self.entitlement_attached = entitlement_attached
+
+    def attach_subscription_with_task(self):
+        task = AttachSubscriptionTask(sla=self.sla)
+        # TODO: check return code to see if the attach procedure was successful
```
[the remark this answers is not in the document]
M4rtinK: Indeed, since dropped.

pyanaconda/modules/subscription/subscription.py
```diff
@@ -242,5 +441,438 @@ def set_system_purpose_with_task(self, sysroot):
...
+        """Set if an entitlement has been attached.
+
+        :param bool system_purpose_applied: True if attached, False otherwise
+
+        NOTE: We keep this as a private method, called the completed signal of the
+              tasks that attach entitlements or unregister the system.
+        """
+        self.entitlement_attached = entitlement_attached
+
+    def attach_subscription_with_task(self):
```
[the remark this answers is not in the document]
M4rtinK: Done. :)

pyanaconda/modules/subscription/subscription_interface.py
```diff
@@ -139,3 +183,196 @@ def SetSystemPurposeWithTask(self, sysroot: Str) -> ObjPath:
...
+
+        Setting "" will restore initial value before
+        it was first overriden using this method.
+
+        :param str url: Red Hat subscription service URL
+        """
+        self.implementation.set_rhsm_baseurl(baseurl)
+
+    @property
+    def ConnectToInsights(self) -> Bool:
```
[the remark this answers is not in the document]
M4rtinK: Good point, let's call it `InsightsEnabled`. :)

pyanaconda/ui/gui/spokes/subscription.py
```diff
@@ -0,0 +1,787 @@
...
+        if entered_text:
+            self.show_activation_key_placeholder = False
+
+        self._update_registration_state()
+
+    def on_system_purpose_checkbox_toggled(self, checkbox):
+        self.system_purpose_visible = checkbox.get_active()
+
+    def on_role_combo_changed(self, combo):
+        self._subscription_module.proxy.SetRole(combo.get_active_id())
```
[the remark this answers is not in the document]
M4rtinK: Replaced by batched setting to the DBus module.

pyanaconda/ui/gui/spokes/subscription.py
M4rtinK: Fixed.

pyanaconda/modules/subscription/subscription.py
M4rtinK: Both - the DBus module mirrors the RHSM API & system state in this regard.

pyanaconda/modules/subscription/installation.py
M4rtinK: Fixed. :)

pyanaconda/modules/subscription/installation.py
M4rtinK: Bzzzz, fixed. :)

pyanaconda/modules/subscription/installation.py
M4rtinK: Thanks, fixed. :)

pyanaconda/modules/subscription/installation.py
```diff
@@ -15,32 +15,466 @@
...
+            log.debug("transfer-subscription-task: transferring syspurpose file")
+            target_syspurpose_file_path = self._sysroot + self.RHSM_SYSPURPOSE_FILE_PATH
+            self._copy_file_to_path(self.RHSM_SYSPURPOSE_FILE_PATH, target_syspurpose_file_path)
+
+    def _transfer_rhsm_config_file(self):
+        """Transfer the RHSM config file."""
+        log.debug("transfer-subscription-task: transferring RHSM config file")
+        target_rhsm_config_path = self._sysroot + self.RHSM_CONFIG_FILE_PATH
+        if not self._copy_file_to_path(self.RHSM_CONFIG_FILE_PATH, target_rhsm_config_path):
+            msg = "RHSM config file (%s) is missing." % self.RHSM_CONFIG_FILE_PATH
```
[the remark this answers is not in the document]
M4rtinK: Indeed, should be fixed now. :)

pyanaconda/modules/subscription/subscription.py

pyanaconda/modules/subscription/subscription.py

pyanaconda/modules/subscription/subscription.py
M4rtinK: Docstring added. :)

pyanaconda/modules/subscription/subscription.py
M4rtinK: Added. :)

pyanaconda/modules/subscription/subscription.py
M4rtinK: Fixed. :)

pyanaconda/ui/gui/spokes/subscription.py
M4rtinK: I don't think we actually need this signal as long as the relevant data in the Subscription DBus module is not changed after kickstart is parsed by other actors than the Subscription spoke itself.

So I'll just drop that signal & make sure registration update is run at the end of `initialize`.

pyanaconda/modules/subscription/subscription.py
M4rtinK: Log message added.

pyanaconda/modules/subscription/subscription.py
M4rtinK: The signal has been added & redundant setters dropped.

pyanaconda/modules/subscription/installation.py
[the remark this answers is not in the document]
M4rtinK: Actually, it looks like the service will get started automatically even with an updates image, likely due to auto-activation. :) So looks like we should be able to drop this completely right away.

pyanaconda/payload/dnfpayload.py
```diff
@@ -380,12 +380,15 @@ def _add_repo(self, ksrepo):
             repo.exclude = ksrepo.excludepkgs
 
         if ksrepo.sslcacert:
+            log.repo(ksrepo.sslcacert)
```
poncovka: Are you sure this is correct?

pyanaconda/subscription.py
```diff
@@ -0,0 +1,178 @@
...
+    # possible, we need to first unregister the already
+    # registered system, as a registration attempt on
+    # an already registered system would fail.
+    if subscription_proxy.IsRegistered:
+        log.debug("RHSM: subscription thread: system already registered, unregistering")
+        progress_callback(SubscriptionPhase.UNREGISTER)
+        task_path = subscription_proxy.UnregisterWithTask()
+        task_proxy = SUBSCRIPTION.get_proxy(task_path)
+        sync_run_task(task_proxy)
+        error = task_proxy.Error
```
poncovka: This is just wrong. You should handle an exception.

pyanaconda/ui/gui/spokes/installation_source.py
```diff
@@ -618,12 +621,18 @@ def changed(self):
         update_payload_repos = self._update_payload_repos()
         return method_changed or update_payload_repos or self._error
 
+    @property
+    def CDN_used_as_installation_source(self):
```
poncovka: Please, use only lowercase letters.

poncovka: This property looks like a hack to me.

pyanaconda/ui/gui/spokes/installation_source.py
```diff
@@ -641,6 +650,8 @@ def ready(self):
 
     @property
     def status(self):
+        if self.CDN_used_as_installation_source:
```
poncovka: I am afraid of the side effects of this `if` that overrides all other states.

pyanaconda/ui/gui/spokes/installation_source.py
```diff
@@ -665,7 +676,10 @@ def status(self):
                 return _("Error setting up installation from HDD")
             return os.path.basename(self._currentIsoFile)
         elif self.payload.baseRepo:
-            return _("Closest mirror")
+            if self._payload_module.proxy.IsRedHatCDNEnabled:
+                return _("Red Hat CDN")
```
poncovka: I don't understand why `Red Hat CDN` is defined twice in this property.

pyanaconda/ui/gui/spokes/software_selection.py
```diff
@@ -36,6 +36,8 @@
 from pyanaconda.core.async_utils import async_action_wait
 from pyanaconda.ui.categories.software import SoftwareCategory
 
+from pyanaconda.modules.common.constants.services import SUBSCRIPTION, PAYLOAD
```
poncovka: I don't see any changes in TUI. Does the installation in the text mode with a subscription in a kickstart file work as expected? What happens if the subscription fails?

pyanaconda/ui/gui/spokes/installation_source.py
M4rtinK: By this we basically delegate the detailed source configuration to the Subscription spoke, which makes sure the subscription is attached properly enabling CDN use.

I guess there could still be some edge cases related to using CDN + custom additional repos, I'll make a note to investigate those as a followup.

anaconda.py
```diff
@@ -725,6 +726,23 @@ def _earlyExceptionHandler(ty, value, traceback):
...
+    from pyanaconda.modules.common.constants.services import SUBSCRIPTION
+    subscription_proxy = SUBSCRIPTION.get_proxy()
+
+    org_set = bool(subscription_proxy.Organization)
+    key_set = subscription_proxy.IsActivationKeySet
+
+    if org_set and key_set:
+        threadMgr.add(AnacondaThread(name=constants.THREAD_SUBSCRIPTION,
+                                     target=subscribe,
+                                     args=()))
```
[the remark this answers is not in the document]
M4rtinK: Good point, removed. :)

pyanaconda/modules/payload/payload.py
```diff
@@ -49,3 +57,12 @@ def generate_kickstart(self):
         log.debug("Generating kickstart data...")
         data = self.get_kickstart_handler()
         return str(data)
+
+    @property
+    def red_hat_cdn_enabled(self):
+        return self._red_hat_cdn_enabled
+
+    @red_hat_cdn_enabled.setter
+    def red_hat_cdn_enabled(self, enabled):
```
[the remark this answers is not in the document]
M4rtinK: Indeed, fixed!

pyanaconda/modules/subscription/installation.py
```diff
@@ -15,32 +15,491 @@
...
+from pyanaconda.anaconda_loggers import get_module_logger
+log = get_module_logger(__name__)
+
+
+def start_rhsm_private_bus():
+    """Start the RHSM private DBus.
+
+    The RHSM private DBus session is used for secure credential passing.
+
+    NOTE: There is no locking invovlved, in starting/stopping of
```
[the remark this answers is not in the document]
M4rtinK: Fixed, thanks.

pyanaconda/modules/subscription/installation.py
```diff
@@ -15,32 +15,491 @@
...
+          the private bus. So it's important to make sure only
+          one task using the RHSM private bus is running at a time.
+          Otherwise unexpected behavior might happen (one task shutting down
+          private bus another task is still using, etc.).
+
+    return: address of the RHSM private DBus session
+    rtype: str
```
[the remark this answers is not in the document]
M4rtinK: Fixed.

pyanaconda/ui/gui/spokes/installation_source.py
M4rtinK: Still seems less problematic to me than touching the method based logic in any way. I'll ad a note to look at this as a followup, but I'm not sure if we have better options with the state of the Source spoke code on RHEL 8.

M4rtinK: Switched to lowercase.

pyanaconda/ui/gui/spokes/installation_source.py
M4rtinK: Good point, the second one is redundant, dropping it now.

pyanaconda/ui/gui/spokes/software_selection.py
M4rtinK: Good point, I'll check. Even with no Subscription spoke, we might need to tune up the Source spoke status for partial kickstart installations.

M4rtinK: So a simple TUI based kickstart installation from CDN seems to work fine.

pyanaconda/installation_tasks.py
```diff
@@ -435,7 +442,15 @@ def run_task(self):
...
         """
         if self._task:
-            self._task(*self._task_args, **self._task_kwargs)
+            try:
+                self._task(*self._task_args, **self._task_kwargs)
+            except Exception as e:
+                # If an exception is raised when processing the task, run the exception through
+                # the error handler and raise it back if it does not match any of the exceptions
+                # we can handle.
+                if errors.errorHandler.cb(e) == errors.ERROR_RAISE:
```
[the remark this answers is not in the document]
M4rtinK: I'm sure we can find a better way & implement that in a followup PR. But at least it seems to work fine so far in all the cases I have tried & it has been like this since the initial implementation.

pyanaconda/payload/dnfpayload.py
M4rtinK: Good point, a debugging leftover. Removed. :)

pyanaconda/subscription.py
M4rtinK: OK, lets do that in the context of the error handling improvements PR.

pyanaconda/modules/subscription/installation.py
```diff
@@ -15,32 +15,491 @@
...
+        # signal that we are done
+        self._done()
+
+    @async_action_nowait
+    def _done(self):
+        # tell anyone listening that we are done
+        self.succeeded.emit()
+
+
+class SubscriptionTask(Task, metaclass=ABCMeta):
```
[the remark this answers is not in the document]
M4rtinK: Docstring added.

pyanaconda/errors.py
```diff
@@ -350,8 +375,14 @@ def cb(self, exn):
                 "PasswordCryptError": self._passwordCryptErrorHandler,
                 "ZIPLError": self._ziplErrorHandler}
 
+        _dbus_map = {
+                "org.fedoraproject.Anaconda.InsightsConnectError" : self._insightsConnectErrorHandler,
+                "org.fedoraproject.Anaconda.InsightsClientMissingError" : self._insightsClientMissingErrorHandler}
```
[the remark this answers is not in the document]
M4rtinK: Let's move improving this to a followup PR. At least this seems to be working fine currently as far as I can tell.

pyanaconda/errors.py
```diff
@@ -19,6 +19,9 @@
 from pyanaconda.core.i18n import _, C_
 from pyanaconda.flags import flags
 
+from pyanaconda.anaconda_loggers import get_module_logger
+log = get_module_logger(__name__)
```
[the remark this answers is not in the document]
M4rtinK: Indeed, removed. :)

pyanaconda/ui/gui/spokes/subscription.py
```diff
@@ -0,0 +1,850 @@
...
+        self.initialize_done()
+
+    def _fill_combobox(self, combobox, clicked_callback, user_provided_value, valid_values):
+        """Fill the given list box with data based on current value & valid values.
+
+        Please note that it is possible that the list box will be empty if no
+        list of valid values are available and the user has not supplied any value
+        via kickstart or the DBUS API.
+
+        :param listbox: the listbox to fill
```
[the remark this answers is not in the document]
M4rtinK: Fixed. :)

pyanaconda/modules/payload/payload_interface.py
```diff
@@ -20,8 +20,27 @@
...
+        super().connect_signals()
+        self.watch_property("IsRedHatCDNEnabled", self.implementation.red_hat_cdn_enabled_changed)
+
+    @property
+    def IsRedHatCDNEnabled(self) -> Bool:
+        """Report if the Red HAt CDN is enabled as the installation source."""
+        return self.implementation.red_hat_cdn_enabled
+
+    @emits_properties_changed
+    def SetRedHatCDNEnabled(self, cdn_enabled: Bool):
```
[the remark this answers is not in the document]
M4rtinK: OK, let's drop the `Is`. :)

pyanaconda/payload/dnfpayload.py
```diff
@@ -383,9 +383,10 @@ def _add_repo(self, ksrepo):
             repo.sslcacert = ksrepo.sslcacert
 
         if ksrepo.sslclientcert:
-            repo.sslclientcert = ksrepo.sslclientcert
+            repo.sslcacert = ksrepo.sslclientcert
 
         if ksrepo.sslclientkey:
+            log.repo(ksrepo.sslclientkey)
```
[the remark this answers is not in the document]
M4rtinK: Gone! :)

pyanaconda/payload/dnfpayload.py
```diff
@@ -820,10 +821,18 @@ def _sync_metadata(self, dnf_repo):
     def baseRepo(self):
         # is any locking needed here?
         repo_names = [constants.BASE_REPO_NAME] + constants.DEFAULT_REPOS
+        subscription_proxy = SUBSCRIPTION.get_proxy()
         with self._repos_lock:
-            for repo in self._base.repos.iter_enabled():
-                if repo.id in repo_names:
+            if subscription_proxy.IsSubscriptionAttached:
+                # If we have a subscription attached, any of the enabled repos
+                # should be fine as the base repo.
```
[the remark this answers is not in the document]
M4rtinK: AFAIK, the priority does not matter here - as long as both are enabled, all should be fine.

The CDN repo files generated by RHSM also have generated names without a clear machine readable pattern that we could use to find out which one of them is BaseOS. So this is the best we can do. But this code has been in place from the beginning & I have not seen any issues with it so far (eq. one of the repos missing, etc.).

pyanaconda/subscription.py
```diff
@@ -0,0 +1,178 @@
...
+# 02110-1301, USA.  Any Red Hat trademarks that are incorporated in the
+# source code or documentation are not subject to the GNU General Public
+# License and may only be used or replicated with the express permission of
+# Red Hat, Inc.
+#
+# Why is this needed ?
+#
+# The Red Hat subscription related tasks communicate over network and might
+# take some time to finish (up to tens of seconds). We definitely don't
+# to block neither automated installation or the UI for these to finish.
```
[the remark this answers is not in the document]
M4rtinK: Fixed, thanks! :)

pyanaconda/subscription.py
```diff
@@ -0,0 +1,178 @@
...
+#
+# Some of these tasks (register + attach) need to run in threads
+# and these threads need to be started from at least two places:
+# - from early startup code for automated installations
+# - from UI based on user interaction
+#
+# Also in some cases, multiple individual DBus tasks will need to be run.
+# Anaconda modularity on RHEL8 is not advanced enough to handle this by itself,
+# so we need s imple sheduler living in the context of the main Anaconda
```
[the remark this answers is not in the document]
M4rtinK: Fixed. :)

pyanaconda/modules/subscription/subscription_data.py
```diff
@@ -0,0 +1,221 @@
...
+    def _parse_subscription_json(self, subscription_json):
+        """Parse JSON data corresponding to subscriptions attached to the system.
+
+        :param str subscription_json: JSON describing what subscriptions have been attached
+
+        The expected JSON is at top level a list of rather complex dictionaries,
+        with each dictionary describing a single subscription that has been attached
+        to the system.
+
+        In this function we actually only parse the JSON, split ti per subscription
```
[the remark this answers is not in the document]
M4rtinK: Fixed. :)

PR #2219, opened 2019-11-15
The great Payload renaming!

jkonecny12:
Payload handler changed to payload. Rename everything now before it will be too late!

### DIFF
--- a/pyanaconda/modules/common/constants/interfaces.py
+++ b/pyanaconda/modules/common/constants/interfaces.py
@@ -50,7 +50,7 @@ DEVICE_TREE_HANDLER = DBusInterfaceIdentifier(
     basename="Handler"
 )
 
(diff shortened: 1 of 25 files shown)


pyanaconda/modules/common/containers.py
```diff
@@ -37,6 +37,11 @@
     message_bus=DBus
 )
 
+PayloadContainer = DBusContainer(
+    namespace=PAYLOAD_BASE_NAMESPACE,
```
poncovka: Btw. this container will publish objects at for example `.../Modules/Payload/Base/1`.

jkonecny12: Ups this shouldn't be here. It's part of the other PR.

jkonecny12: Fixed. I really should more look on what I'm posting as PR before posting it...

PR #2220, opened 2019-11-15
Clean up the spoke for the advanced storage

poncovka:
Simplify the code and rename camelCase names of methods, attributes and variables.

### DIFF
--- a/pyanaconda/ui/gui/spokes/advanced_storage.py
+++ b/pyanaconda/ui/gui/spokes/advanced_storage.py
@@ -16,12 +16,6 @@
 # License and may only be used or replicated with the express permission of
 # Red Hat, Inc.
 #
-
-import gi
-gi.require_version("Gtk", "3.0")
-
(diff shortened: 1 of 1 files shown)


pyanaconda/ui/gui/spokes/advanced_storage.py
```diff
@@ -127,98 +166,93 @@ def clear(self):
...
-        active = self._portCombo.get_active_text()
+        active = self._port_combo.get_active_text()
         if active:
             if hasattr(device, "port"):
                 return device.port == int(active)
-            else:
-                return False
-        else:
-            return True
+
```
jkonecny12: Please add else here, it will be easier to follow the
```
    if <condition>:
        return
    return
return
```

```suggestion
        else:
```

poncovka: Fixed.

poncovka:
Updated.

PR #2222, opened 2019-11-19
Rename payload modules

jkonecny12:
Rest of the payload renaming. Hopefully this is the last part. 

Rename:

`Payload` -> `Payloads`
`Payload.Sources` -> `Payloads.Source`
`Payload.Payloads` -> `Payloads.Payload`

So in the DBus after dynamic provisioning (will follow in different PR) it should be like:

```
payloads/payload/0
payloads/payload/1
...
payloads/source/0
payloads/source/1
...
```

This change was chosen to be consistent with Users and Services because we support multiple payloads now and planning to support even using more payloads in during one installation.


pyanaconda/modules/common/constants/objects.py
```diff
@@ -122,11 +122,11 @@
...
 
 PAYLOAD_LIVE_IMAGE = DBusObjectIdentifier(
-    namespace=PAYLOAD_NAMESPACE,
+    namespace=PAYLOADS_NAMESPACE,
     basename="LiveImage"
 )
 
 PAYLOAD_LIVE_OS = DBusObjectIdentifier(
-    namespace=PAYLOAD_NAMESPACE,
+    namespace=PAYLOADS_NAMESPACE,
```
poncovka: Payloads should use the namespace `PAYLOAD_BASE_NAMESPACE`. Then, for example, the DBus name of `PAYLOAD_LIVE_OS` will be `.Payloads.Payload.LiveOS`. Is this change planned for the pull request that will introduce dynamic payload modules?

pyanaconda/modules/common/constants/namespaces.py
```diff
@@ -96,12 +86,22 @@
...
+    *MODULES_NAMESPACE,
+    "Payloads"
+)
+
+DNF_NAMESPACE = (
+    *PAYLOADS_NAMESPACE,
+    "DNF"
+)
+
 PAYLOAD_BASE_NAMESPACE = (
```
poncovka: What about using the names `PAYLOADS_PAYLOAD_NAMESPACE` and `PAYLOADS_SOURCE_NAMESPACE`? Is it too confusing? Btw, I am fine also with the name `PAYLOAD_NAMESPACE`.

pyanaconda/modules/common/constants/objects.py
jkonecny12: Yes, that will be introduced by the dynamic payloads PR.

pyanaconda/modules/common/constants/namespaces.py
jkonecny12: Really not sure. I'm thinking the same. It's really hard to find some good solution for payloads and payload...

That would be inconsistent with `PAYLOAD_LIVE_OS` or similar and I don't think we want to rename payload to `PAYLOADS_LIVE_OS`.

Maybe the best solution would be to have `PAYLOAD_BASE_NAMESPACE` and `PAYLOAD_SOURCE_BASE_NAMESPACE` or just `PAYLOAD_BASE_NAMESPACE`?

poncovka: I don't like `BASE` in the name, because I feel we usually use it in a little different context. What about `PAYLOAD_GENERAL`? Anyway, feel free to merge the pull request as it is. It is not so important.

PR #2223, opened 2019-11-19
Escape spaces in dracut partition specifications

bcl:
For example, when using a kickstart with:
driverdisk "CDLABEL=Fedora 23 x86_64"

it is translated into "inst.dd:hd:CDLABEL=Fedora 23 x86_64"
on the cmdline. The spaces need to be escaped as \\x20

Resolves: rhbz#1452770


jkonecny12:
Replaced by https://github.com/rhinstaller/anaconda/pull/2374.

PR #2224, opened 2019-11-20
Covert VG reserved_space to Size

vojtechtrefny:
We get amount of reserved space in MiB from kickstart but blivet
needs Size.

Resolves: rhbz#1774477

### DIFF
--- a/pyanaconda/kickstart.py
+++ b/pyanaconda/kickstart.py
@@ -2271,7 +2271,7 @@ class VolGroupData(COMMANDS.VolGroupData):
 
             storage.create_device(request)
             if self.reserved_space:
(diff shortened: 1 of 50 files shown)


pyanaconda/kickstart.py
```diff
@@ -2271,7 +2271,7 @@ def execute(self, storage, ksdata, instClass):
 
             storage.create_device(request)
             if self.reserved_space:
-                request.reserved_space = self.reserved_space
+                request.reserved_space = Size("%d MiB" % self.reserved_space)
```
jkonecny12: Please change this to .format().

vojtechtrefny: Fixed.

PR #2231, opened 2019-11-25
Calculate the space on disks with supported disk labels

poncovka:
Skip disks with disk labels that are not supported on the current platform
in functions get_disk_free_space and get_disk_reclaimable_space.

### DIFF
--- a/pyanaconda/ui/gui/spokes/storage.py
+++ b/pyanaconda/ui/gui/spokes/storage.py
@@ -64,7 +64,6 @@ from pyanaconda.storage.kickstart import reset_custom_storage_data
 
 from blivet.size import Size
 from blivet.devices import MultipathDevice, ZFCPDiskDevice, NVDIMMNamespaceDevice
-from blivet.formats.disklabel import DiskLabel
 from pyanaconda.threading import threadMgr, AnacondaThread
 from pyanaconda.product import productName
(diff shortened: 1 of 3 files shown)


pyanaconda/ui/gui/spokes/storage.py
```diff
@@ -853,21 +852,8 @@ def _check_dasd_formats(self):
...
-        # you need to have at least one of the platform's labels in order for
-        # any of the free space to be useful.
-        disk_labels = set(disk.format.label_type for disk in disks
-                              if hasattr(disk.format, "label_type"))
-        platform_labels = set(DiskLabel.get_platform_label_types())
-        if disk_labels and platform_labels.isdisjoint(disk_labels):
-            disk_free = 0
-            fs_free = 0
-            log.debug("Need disklabel: %s have: %s", ", ".join(platform_labels),
-                                                     ", ".join(disk_labels))
```
VladimirSlavik: Is removing this from logs a good thing? I can see how it is "not nice" to have it logged in the storage code, but it seems an useful piece of information to have. On the other hand I am not sure how one would correlate the lack of space with the log message, except during debugging. If the user is not learning that "_No present disks satisfy labels needed for the platform to be installed._" then I can see how losing it is not really much of a loss. Make of that what you will...

PR #2236, opened 2019-11-28
Create the boot loader factory

poncovka:
Create the option `type` in the Bootloader section of the Anaconda configuration
and set it to EXTLINUX if the cmdline option `extlinux` is set.

Create a factory of the boot loader instances. Call its method `set_default_class`
to set up the default boot loader class. It can be used for the EXTLINUX override.

Use a configuration option or a kickstart option to set up the factory.
Remove the DBus property `BootloaderType` of the Bootloader module. Instead,
use the temporary method `GetDefaultType` to set up the factory from UI.


poncovka:
Jenkins, test this please.

poncovka:
Rebased.

PR #2239, opened 2019-11-28
Use id_path as a long identifier of a disk

poncovka:
Remove the function `get_long_identifier` and use the attribute `id_path` instead.
Extend the attributes of the device data with a new attribute `path-id`.

**Depends on:** https://github.com/storaged-project/blivet/pull/814


poncovka:
Rebased and tested.

poncovka:
The Blivet version was changed to 3.2.0-1.

PR #2240, opened 2019-11-29
Reduce progress spinner repaints to save CPU (rhbz#1204242)

VladimirSlavik:
This makes the spinner redraw less (at ~24 FPS), which in turn frees CPU for
actual install work. Consequently, this speeds up testing in single-processor
virtual machines.

The savings are proportional to the time the spinner would be visible, so
the more packages, the more time saved. Thus the effect is more significant
when installing Workstation with rich software selection.

Resolves: rhbz#1204242

(edited: 12 -> 24 according to changes)

### DIFF
--- a/data/anaconda-gtk.css
+++ b/data/anaconda-gtk.css
@@ -139,3 +139,11 @@ AnacondaLayoutIndicator {
     background-color: #fdfdfd;
     color: black;
 }
+
+/* Fix the cpu-hogging spinner in progress spoke. This works by overriding the css property
+"animation-timing-function". The default Adwaita value "linear" requests infinitely fine-grained
+progress of the animation, so the component redraws every time it can. Using "steps(N)" makes the
+animation have only N discrete states with no interpolation, so there are effectively N intervals
+where the look is constant. This reduces redraws significantly and saves the CPU. See rhbz#1204242
+for measurements. */
+spinner { animation-timing-function: steps(24); }


VladimirSlavik:
Kamil [reports](https://bugzilla.redhat.com/show_bug.cgi?id=1204242#c21) this change makes the spinner add only 14% instead of 77% in the more pessimistic case usually tested. The actual FPS rate is up for debate - 12 offers some savings while not being too ugly, but the difference between 12 and 16 is quite noticeable.

data/anaconda-gtk.css
```diff
@@ -139,3 +139,11 @@ AnacondaLayoutIndicator {
...
     color: black;
 }
+
+/* Fix the cpu-hogging spinner in progress spoke. This works by overriding the css property
+"animation-timing-function". The default Adwaita value "linear" requests infinitely fine-grained
+progress of the animation, so the component redraws every time it can. Using "steps(12)" makes the
+animation have only 12 discrete states with no interpolation, so there are effectively 12 intervals
+where the look is constant. This reduces redraws significantly and saves the CPU. See rhbz#1204242
+for measurements. */
+AnacondaStandaloneWindow spinner#progressSpinner { animation-timing-function: steps(12); }
```
poncovka: I would use 24 FPS if we are going to animate the rotation of the current icon. That shouldn't be notable at all and still improve the performance a little. If we want to use a lower number of steps, we need a different icon, that can be animated in a different way.

Please, redefine the style for all spinners of the installer.

VladimirSlavik: Done.

VladimirSlavik:
Update: Changed FPS to 24.

pyanaconda/ui/gui/spokes/installation_progress.glade
```diff
@@ -57,6 +57,7 @@
                         <property name="spacing">6</property>
                         <child>
                           <object class="GtkSpinner" id="progressSpinner">
+                            <property name="name">progressSpinner</property>
```
poncovka: You don't need to define the name of the spinner anymore.

PR #2243, opened 2019-12-05
Localization populate missing in installation task

rvykydal:


pyanaconda/modules/localization/installation.py
```diff
@@ -98,80 +98,86 @@ def name(self):
...
+        if vc_keymap:
             write_vc_configuration(
-                self._vc_keymap,
+                vc_keymap,
                 self._sysroot
             )
 
 
-def write_x_configuration(x_layouts, switch_options, root):
+def write_x_configuration(localed_wrapper, x_layouts, switch_options, x_conf_dir_path, root):
```
VladimirSlavik: I'm not familiar with this, so I am not sure why the `localed_wrapper` instance is passed from the caller. It seems redundant to pass when it wraps the same thing and you could just construct a new one, as you did? I can see how it makes tests easier though...

rvykydal: The tests are the reason here. I am even considering passing the wrapper instance to the task(s) (to be shared by the module).

VladimirSlavik: Thank you, noted.

PR #2244, opened 2019-12-05
Extend DBus support for the Resize dialog window

poncovka:
Extend the DBus support of the Storage module to be able to
remove the local storage object from the Resize dialog window.

### DIFF
--- a/pyanaconda/modules/common/structures/storage.py
+++ b/pyanaconda/modules/common/structures/storage.py
@@ -32,6 +32,7 @@ class DeviceData(DBusData):
         self._size = 0
         self._parents = []
         self._is_disk = False
+        self._protected = False
         self._removable = False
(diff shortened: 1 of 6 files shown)


pyanaconda/modules/common/structures/storage.py
```diff
@@ -96,6 +97,15 @@ def is_disk(self) -> Bool:
     def is_disk(self, is_disk: Bool):
         self._is_disk = is_disk
 
+    @property
+    def protected(self) -> Bool:
+        """It this device protected?"""
```
rvykydal: It -> Is

poncovka: Fixed.

poncovka:
Fixed and rebased.

PR #2248, opened 2019-12-11
Calculate the space on uninitialized disks (#1782449)

poncovka:
Don't skip uninitialized disks in functions get_disk_free_space
and get_disk_reclaimable_space.

Resolves: rhbz#1782449


poncovka:
The docstring is changed.

PR #2250, opened 2019-12-11
Fix text color on info bars

M4rtinK:
Due to some external changes - likely in GTK or the way
it processes CSS styles - our info bar text color overrides
no longer work.

Make the overrides more specific, so they match again and override
the text color on info bars as intended.

This should finally make the warnings Anaconda displays readable
again, as they no longer will be while on orange but black on orange
as intended.

Resolves: rhbz#1782174

### DIFF
--- a/data/anaconda-gtk.css
+++ b/data/anaconda-gtk.css
@@ -46,34 +46,45 @@ levelbar.discrete trough block.filled.high {
 @define-color error_fg_color white;
 @define-color error_bg_color rgb (237, 54, 54);
 
-.info {
+infobar.info {
     background-color: @info_bg_color;
-    color: @info_fg_color;
     border-color: darker(@info_bg_color);
 }
 
-.warning {
+infobar.info label {
+    color: @info_fg_color;
+}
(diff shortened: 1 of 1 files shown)


M4rtinK:
So this is how it looks like with the fix:

RHEL_8 2_fixed_warning_text_color

poncovka:
Btw. I have tested it in upstream and it doesn't work at all.

VladimirSlavik:
Looks good to me too.

(I think the icon is rather lost in the background color, but that is not part of the bug, so never mind.)

andreasn:
Looks great, thank you for the quick fix!

andreasn:
> I think the icon is rather lost in the background color, but that is not part of the bug, so never mind.

I agree, and style-wise it's different from the icons used on top of the spokes.
Filed as https://bugzilla.redhat.com/show_bug.cgi?id=1782832

M4rtinK:
Fedora PR opened (#2418), removing the `port to Fedora` flag.

PR #2252, opened 2019-12-16
Move the support for resizing devices

poncovka:
Move the DBus support for the Resize dialog to the device tree
of the automatic partitioning module.

The workaround for resizing and removing of devices is too fragile, so move
the code to new functions and call these functions from the Resize dialog.


poncovka:
The Resizable interface is going to be unsupported, so we can modify it later.

PR #2253, opened 2019-12-17
Delegate locale parsing to langtable

VladimirSlavik:
Use parsing function from last (.49) release of langtable instead of our custom parsing with regexp. This simplifies considerably large portions of the localization code and  enables a few more languages. Tests are adjusted accordingly. Some errors found during testing are fixed too.

All of this is possible thanks to @mike-fabian who exposed the parsing method in langtable!

### DIFF
--- a/pyanaconda/localization.py
+++ b/pyanaconda/localization.py
@@ -105,7 +105,7 @@ def is_supported_locale(locale):
     :type locale: str
     :return: whether the given locale is supported or not
     :rtype: bool
-    :raise InvalidLocaleSpec: if an invalid locale is given (see LANGCODE_RE)
+    :raise InvalidLocaleSpec: if an invalid locale is given (see is_valid_langcode)
 
     """
 
@@ -122,7 +122,7 @@ def locale_supported_in_console(locale):
     :param str locale: locale to test
     :return: whether the given locale is supported by the console or not
     :rtype: bool
-    :raise InvalidLocaleSpec: if an invalid locale is given (see LANGCODE_RE)
+    :raise InvalidLocaleSpec: if an invalid locale is given (see is_valid_langcode)
 
     """
 
--- a/pyanaconda/localization.py
+++ b/pyanaconda/localization.py
@@ -35,15 +35,18 @@ log = get_module_logger(__name__)
 
 SCRIPTS_SUPPORTED_BY_CONSOLE = {'Latn', 'Cyrl', 'Grek'}
 
+
 class LocalizationConfigError(Exception):
     """Exception class for localization configuration related problems"""
 
     pass
 
+
 class InvalidLocaleSpec(LocalizationConfigError):
     """Exception class for the errors related to invalid locale specs"""
 
     pass
+
 
 def is_valid_langcode(langcode):
     """Check if the given locale has a language specified.
@@ -54,6 +57,7 @@ def is_valid_langcode(langcode):
     parsed = langtable.parse_locale(langcode)
     return bool(parsed.language)
 
+
 def raise_on_invalid_locale(arg):
     """Helper to abort when a locale is not valid.
 
@@ -62,13 +66,14 @@ def raise_on_invalid_locale(arg):
     if not is_valid_langcode(arg):
         raise InvalidLocaleSpec("'{}' is not a valid locale".format(arg))
 
+
 def get_language_id(locale):
     """Return language id without territory or anything else."""
     return langtable.parse_locale(locale).language
 
+
 def is_supported_locale(locale):
-    """
-    Function that tells if the given locale is supported by the Anaconda or
+    """Function that tells if the given locale is supported by the Anaconda or
     not. We consider locales supported by the langtable as supported by the
     Anaconda.
 
@@ -77,15 +82,13 @@ def is_supported_locale(locale):
     :return: whether the given locale is supported or not
     :rtype: bool
     :raise InvalidLocaleSpec: if an invalid locale is given (see is_valid_langcode)
-
     """
-
     en_name = get_english_name(locale)
     return bool(en_name)
 
+
 def locale_supported_in_console(locale):
-    """
-    Function that tells if the given locale can be displayed by the Linux console.
+    """Function that tells if the given locale can be displayed by the Linux console.
 
     The Linux console can display Latin, Cyrillic and Greek characters reliably,
     but others such as Japanese, can't be correctly installed.
@@ -94,11 +97,10 @@ def locale_supported_in_console(locale):
     :return: whether the given locale is supported by the console or not
     :rtype: bool
     :raise InvalidLocaleSpec: if an invalid locale is given (see is_valid_langcode)
-
     """
-
     locale_scripts = get_locale_scripts(locale)
     return set(locale_scripts).issubset(SCRIPTS_SUPPORTED_BY_CONSOLE)
+
 
 def langcode_matches_locale(langcode, locale):
     """Function that tells if the given langcode matches the given locale. I.e. if
@@ -128,6 +130,7 @@ def langcode_matches_locale(langcode, locale):
             return False
 
     return True
+
 
 def find_best_locale_match(locale, langcodes):
     """Find the best match for the locale in a list of langcodes. This is useful
@@ -183,9 +186,9 @@ def find_best_locale_match(locale, langcodes):
     else:
         return None
 
+
 def setup_locale(locale, localization_proxy=None, text_mode=False):
-    """
-    Procedure setting the system to use the given locale and store it in to the
+    """Procedure setting the system to use the given locale and store it in to the
     localization module (if given). DOES NOT PERFORM ANY CHECKS OF THE GIVEN
     LOCALE.
 
@@ -205,9 +208,7 @@ def setup_locale(locale, localization_proxy=None, text_mode=False):
     :param bool text_mode: if the locale is being setup for text mode
     :return: the locale that was actually set
     :rtype: str
-
     """
-
     if localization_proxy:
         localization_proxy.SetLanguage(locale)
 
@@ -240,9 +241,9 @@ def setup_locale(locale, localization_proxy=None, text_mode=False):
         if not font_set:
             log.warning("can't set console font for locale %s", locale)
             # report what exactly went wrong
-            if not(script_supported):
+            if not script_supported:
                 log.warning("script not supported by console for locale %s", locale)
-            if not(console_fonts):  # no fonts known for locale
+            if not console_fonts:  # no fonts known for locale
                 log.warning("no console font found for locale %s", locale)
             if script_supported and console_fonts:
                 log.warning("none of the suggested fonts can be set for locale %s", locale)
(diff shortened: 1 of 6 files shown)


pyanaconda/localization.py
```diff
@@ -148,20 +148,20 @@ def langcode_matches_locale(langcode, locale):
...
+    langcode_parsed = langtable.parse_locale(langcode)
+    locale_parsed = langtable.parse_locale(locale)
+
     # Check parts one after another. If some part appears in the langcode and
     # doesn't match the one from the locale (or is missing in the locale),
     # return False, otherwise they match
     for part in ("language", "territory", "script", "encoding"):
-        if langcode_parts[part] and langcode_parts[part] != locale_parts.get(part):
+        value = langcode_parsed.__getattribute__(part)
+        if value and value!= locale_parsed.__getattribute__(part):
```
jkonecny12: Please change the `__getattribute__` to [getattr()](https://docs.python.org/3/library/functions.html#getattr) calls.

pyanaconda/localization.py
```diff
@@ -189,20 +189,25 @@ def find_best_locale_match(locale, langcodes):
...
 
+        locale_parsed = langtable.parse_locale(locale)
+        langcode_parsed = langtable.parse_locale(langcode)
+
         for part, part_score in SCORE_MAP.items():
-            if locale_parts[part] and langcode_parts[part]:
-                if locale_parts[part] == langcode_parts[part]:
+
+            locale_value = locale_parsed.__getattribute__(part)
+            langcode_value = langcode_parsed.__getattribute__(part)
```
jkonecny12: The same here also I'm not sure if this will work. If I'm reading the documentation correctly then if the part is missing you will get `AttributeError` exception.

tests/nosetests/pyanaconda_tests/localization_test.py
```diff
@@ -17,9 +17,191 @@
...
+from pyanaconda.core.constants import DEFAULT_LANG
 from pyanaconda.core.util import execWithCaptureBinary
 import locale as locale_mod
 import unittest
+from unittest.mock import call, patch, MagicMock
+from io import StringIO
+
+
+class LangcodeLocaleParsingTests(unittest.TestCase):
+    def is_valid_test(self):
```
jkonecny12: You should also have an invalid test case.

tests/nosetests/pyanaconda_tests/localization_test.py
```diff
@@ -17,9 +17,191 @@
...
+        locale = localization.setup_locale("sk", localization_proxy=loc_proxy)
+
+        loc_proxy.SetLanguage.assert_called_once_with("sk")
+        setenv_mock.assert_called_once_with("LANG", "sk")
+        setlocale_mock.assert_called_once_with(locale_mod.LC_ALL, "sk")
+        set_modules_locale_mock.assert_called_once_with("sk")
+
+        self.assertEqual(locale, "sk")
+
+    @patch("pyanaconda.localization.os.environ", new_callable=dict)
```
jkonecny12: Shouldn't this be replaced by [patch.dict()](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.patch.dict)?

VladimirSlavik: I need to manipulate the dictionary inside the test and patch.dict() gives me no variable to access it.

pyanaconda/localization.py
VladimirSlavik: The attributes are now always present, see https://github.com/mike-fabian/langtable/blob/master/langtable/langtable.py#L1100

pyanaconda/localization.py
VladimirSlavik: Will do!

tests/nosetests/pyanaconda_tests/localization_test.py
VladimirSlavik: On a further thought, perhaps I am reading the docs wrong, it does return a context manager... let's try.

VladimirSlavik: OK, so no variable, but I can access the thing via full path as `localization.os.environ`. Do you feel it is worth the change?

tests/nosetests/pyanaconda_tests/localization_test.py
VladimirSlavik: Will do. The name is a bit misleading because the names are taken from the functions tested, so there won't be a new is_invalid_test ...

VladimirSlavik: Added.

tests/nosetests/pyanaconda_tests/localization_test.py
VladimirSlavik: Used patch.dict anyway.

VladimirSlavik:
Update:
* Tests for failures in the parsing test too
* Use patch.dict

VladimirSlavik:
Apparently the `C` and `POSIX` locales are still not solved with this. The langtable parsing only understands locales that have a language.

Besides that, I have missed a few bits here and there, so this is not done :(

tests/nosetests/pyanaconda_tests/localization_test.py
jkonecny12: This was only a suggestion. I'll leave the solution on you.

jkonecny12:
> Apparently the `C` and `POSIX` locales are still not solved with this. The langtable parsing only understands locales that have a language.
> 
> Besides that, I have missed a few bits here and there, so this is not done enough :(
> 
> Additionally, it turns out that langtable does not "support" a lot of stuff that is in glibc. Output of `localectl list-locales` gives me eg. `fi_FI.iso885915@euro` which should parse, but it seems the data in langtable is lacking the `euro` part, so it gets dropped and then of course it is not the same locale so we do not actually "support" it! Given this, I am now really pessimistic we can define well what we do support.

Did you asked Mike about this?

VladimirSlavik:
> Did you asked Mike about this?
Not yet, I am still "loading" enough knowledge about where what happens.

VladimirSlavik:
Just dumping data here so I won't forget.

Taking `C.UTF-8` as the first offender: It looks like the problem happens when `setup_locale_environment` calls `get_language_locales` which has the `raise_on_invalid_locale` decorator, which finally checks that parsing the locale yields a language. So we could either special-case this, or langtable could. Similarly for POSIX, I guess. I have no idea about the considerations for this, though.

For the fi_FI.iso885915@euro, it seems langtable's parsing lets through only scripts that are specified in its internal list. Others are silently dropped. See https://github.com/mike-fabian/langtable/blob/master/langtable/langtable.py#L139 . Somewhat curiously, the link included in the source leads to a super-long list of valid scripts, including eg. Greek, Hangul, Katakana, and Han variants that I would expect to be represented?

I'll ask Mike in the other thread where we already have all the discussion history.
https://github.com/pykickstart/pykickstart/pull/273

pyanaconda/localization.py
```diff
@@ -163,20 +160,25 @@ def find_best_locale_match(locale, langcodes):
...
-        locale_parts = parse_langcode(locale)
-        langcode_parts = parse_langcode(langcode)
-        if not locale_parts or not langcode_parts:
+        if not is_valid_langcode(locale) or not is_valid_langcode(langcode):
             return score
 
+        locale_parsed = langtable.parse_locale(locale)
+        langcode_parsed = langtable.parse_locale(langcode)
+
         for part, part_score in SCORE_MAP.items():
```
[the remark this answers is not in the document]
VladimirSlavik: I like the idea, but what happens if the order of fields in `langtable.Locale` changes, or new ones are added - perhaps inserted in the middle? The score won't update automatically. As `zip` just silently crops length to the shortest common thing, there would be no indication anything is off.

I suppose I could add a test to validate a number of exact combinations, except the score calculation is an inner function...

pyanaconda/localization.py
```diff
@@ -51,28 +46,30 @@ class InvalidLocaleSpec(LocalizationConfigError):
...
 
-    or None if the given string doesn't match the LANGCODE_RE.
+def raise_on_invalid_locale(fun):
+    """Decorator to return early from a function(locale) when the locale is not valid.
 
+    :raise: InvalidLocaleSpec
     """
+    @wraps(fun)
+    def inner(*args, **kwargs):
+        if not is_valid_langcode(args[0]):
```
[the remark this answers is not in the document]
VladimirSlavik: Sorry, that was stupid indeed :D Now it takes exactly one argument - is that what you meant?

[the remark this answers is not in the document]
VladimirSlavik: Decorator removed.

jkonecny12:
jenkins, test this please

pyanaconda/localization.py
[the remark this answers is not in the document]
VladimirSlavik: Rewritten completely to avoid loop.

VladimirSlavik:
So, I think this is done? Currently testing actual install again to see if there is anything unobvious.

@jkonecny12 if you want, you can take a look at [`find_best_locale_match`](https://github.com/VladimirSlavik/anaconda/blob/locale_parsing/pyanaconda/localization.py#L134) as that received a rather substantial [rewrite](https://github.com/rhinstaller/anaconda/pull/2253/commits/fe6a0468fd5ae78747e6f1515f5094e837c5f4fd#diff-54b102530cdde94395082bd464b4bb69L176) in the process. Vendy says you'll like it :P

tests/nosetests/pyanaconda_tests/localization_test.py
```diff
@@ -17,72 +17,214 @@
...
+        set_modules_locale_mock.assert_called_once_with(DEFAULT_LANG)
+
+        self.assertEqual(locale, DEFAULT_LANG)
+
+    @patch("pyanaconda.localization.setenv")
+    @patch("pyanaconda.localization.locale_mod.setlocale", side_effect=[locale_mod.Error, None])
+    @patch("pyanaconda.localization.set_modules_locale")
+    def setup_locale_setlocale_fail_test(self, set_modules_locale_mock, setlocale_mock, setenv_mock):
+        """Test setup_locale with failure in setlocale"""
+
```
jkonecny12: Please make the newlines on a function/method start consistent. I saw three types here.

1:
```
def test():
"""documentation

description
"""
   code
```

2:
```
def test():
"""documentation

description
"""

   code
```

3:
```
def test():
"""documentation

description

"""
   code
```


You should be consistent with [PEP-257](https://www.python.org/dev/peps/pep-0257/) which means the first variant.

pyanaconda/localization.py
```diff
@@ -496,13 +454,11 @@ def get_xlated_timezone(tz_spec_part):
     """
 
     locale = os.environ.get("LANG", constants.DEFAULT_LANG)
-    parts = parse_langcode(locale)
-    if "language" not in parts:
+
+    if not is_valid_langcode(locale):
         raise InvalidLocaleSpec("'%s' is not a valid locale" % locale)
```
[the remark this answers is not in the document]
VladimirSlavik: Ah, right - that was refactored to be not a decorator after writing this bit. Makes sense here too then. Will fix.

PR #2258, opened 2020-01-02
Calculate ip address for kickstart URL ending in /

imsedgar:
In the past (until Fedora 30) anaconda uses dhclient which sets environment variable new_ip_address before calling script fetch-kickstart.net in dracut hook initqueue.
This variable is used for creating the filename of the kickstart file when the URL for the kickstart file is of type nfs and ends with a slash.
Now dhclient is not used anymore, NetworkManager uses it's own dhcp client, but NetworkManager does not define variable new_ip_address. So this variable is empty when this script is called, the result was that anaconda wants to download a kickstart file with name "-kickstart" instead of "${new_ip_address}-kickstart" which does not exist.
This patch calculates the ipv4 address of the network interface if this variable is not set or empty.
See also rhbz #1769825 .

### DIFF
--- a/dracut/fetch-kickstart-net.sh
+++ b/dracut/fetch-kickstart-net.sh
@@ -44,7 +44,7 @@ case $kickstart in
 
         # URLs that end in '/' get '$IP_ADDR-kickstart' appended.
         if [[ $kickstart == nfs*/ ]]; then
-            kickstart="${kickstart}${new_ip_address}-kickstart"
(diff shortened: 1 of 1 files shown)


jkonecny12:
Jenkins, it's ok to test.

rvykydal:
Checking with latest rawhide it seems to me that *all* the ip addresses found by ip are appendend in some way to the kickstart url, ie the suffix appended to the $kickstart url would be for example:
"127.0.0.1 10.43.26.115 192.168.122.1-kickstart"
I think we should:
- filter out lo device (127.0.0.1)
- use only single address (the last one?)

imsedgar:
I think variable netif is set by the caller of the script in parameter $1.

rvykydal:
You are right, sorry I missed that.

jkonecny12:
jenkins, test this please

PR #2261, opened 2020-01-07
Dynamic payload provisioning

jkonecny12:
Add dynamic provisioning to our Payloads module.

New payloads will be provisioned under path:

```
    /org/fedoraproject/Anaconda/Modules/Payloads/Payload/1
    /org/fedoraproject/Anaconda/Modules/Payloads/Payload/2
...
```

So the payload type is now just interface not DBus object.

As part of this changes I also moved dnf/packages to a separate part which does not belong to specific payload. That way it can be easily extended and shared by other payloads.

Further question:
Should I implement type as payload property? Right now owner (or who created the payload) knows what payload type is it so they can control that. However, if someone else want to look/control these payloads, it's pretty hard to find out what interfaces it has.


pyanaconda/modules/payloads/payload/dnf/dnf.py
```diff
@@ -17,43 +17,35 @@
...
 from pyanaconda.modules.payloads.payload.payload_base import PayloadBase
 from pyanaconda.modules.payloads.payload.dnf.dnf_interface import DNFInterface
-from pyanaconda.modules.payloads.payload.dnf.packages.packages import PackagesModule
 
 from pyanaconda.anaconda_loggers import get_module_logger
 log = get_module_logger(__name__)
 
 
-class DNFModule(PayloadBase):
+class DNFModule(PayloadBase, Publishable):
```
poncovka: I think that the class `PayloadBase` should be `Publishable`. It is an abstract class anyway.

pyanaconda/modules/payloads/payloads_interface.py
```diff
@@ -30,9 +30,14 @@
 class PayloadsInterface(KickstartModuleInterface):
     """DBus interface for Payload module."""
 
-    def GetActivePayloadPath(self) -> ObjPath:
-        """Get path to the payload which is used now."""
-        return self.implementation.get_active_payload_path()
+    def GetActivePayload(self) -> ObjPath:
```
poncovka: This should be a property to be able to monitor the changes with the `PropertiesChanged` signal. Also, there should be a property that returns all created payloads. It will ensure that all created payloads are published.

pyanaconda/modules/payloads/payloads_interface.py
```diff
@@ -30,9 +30,14 @@
...
+        """Get active payload.
+
+        :raise: PayloadNotSetError if payload is not set
+        """
+        return PayloadContainer.to_object_path(
+            self.implementation.get_active_payload()
+        )
 
     def IsPayloadSet(self) -> Bool:
```
poncovka: Btw. I don't think this method is necessary.  You can always ask for the active payload directly. It can be unsafe to use it, for example to check if we can call `GetActivePayload`, because the value can change between the calls. I don't remember if there was a discussion about it, but there is no `FIXME` comment in the docstring.

poncovka:
> Further question:
> Should I implement type as payload property? Right now owner (or who created the payload) knows what payload type is it so they can control that. However, if someone else want to look/control these payloads, it's pretty hard to find out what interfaces it has.

I think it is a good idea. When the payload is created from a kickstart file, UI doesn't know its type. For example, partitioning modules have the `PartitioningMethod` property.

pyanaconda/modules/payloads/payloads.py
```diff
@@ -83,15 +82,15 @@ def is_payload_set(self):
...
 
-        :rtype: str
-        """
-        if self._payload_path:
-            return self._payload_path
+        FIXME: Merge get_active_payload and payload property. They are doing the same.
 
-        return ""
+        :rtype: instance of active payload
+        :raise: PayloadNotSetError is no payload is set
```
rvykydal: a typo: "is" -> "if"

pyanaconda/modules/payloads/payloads_interface.py
jkonecny12: Yeah, I'm thinking if there is a reason to have this. I'm thinking if it would be better to drop the ActivePayload idea and do not recognize what payload is active.

However, there is a question then how we will solve switching between payloads. I mean if user switch from one payload to another how we will disable payloads which will not be used? 

So or so this will be better to solve in a separate PR.

pyanaconda/modules/payloads/payloads_interface.py
jkonecny12: The same as above, not sure about how it should be handled.

I'll add `FIXME`.

pyanaconda/modules/payloads/payload/dnf/dnf.py
jkonecny12: Fixed.

pyanaconda/modules/payloads/payloads.py
jkonecny12: Fixed.

jkonecny12:
**UPDATED:**
* Fix conflicts.
* Fix issues above.
* Implement payload type.
* Implement payload type tests.
* Tweak payload shared testing stuff.

pyanaconda/modules/payloads/payload/payload_base_interface.py
```diff
@@ -39,6 +39,14 @@ def connect_signals(self):
         self.watch_property("Sources", self.implementation.sources_changed)
         self.watch_property("RequiredSpace", self.implementation.required_space_changed)
 
+    @property
+    def Type(self) -> Str:
+        """Get type of this payload.
+
+        Possible types are values of the payload.base.constants.PayloadType enum
+        """
+        return self.implementation.type
```
poncovka: You are returning the enum value here instead of the string.

tests/nosetests/pyanaconda_tests/module_payload_live_os_test.py
```diff
@@ -45,22 +45,25 @@ def setUp(self):
...
 
     def _prepare_and_use_source(self):
         source = self._prepare_source()
-        self.source_tests.set_sources([source])
+        self.shared_tests.set_sources([source])
 
         return source
 
+    def type_test(self):
+        self.shared_tests.check_type(PayloadType.LIVE_OS)
```
poncovka: I would also extend the test for `PayloadFactory`, because the type of the created payload should be the same as the requested type. It should be easy to check.

pyanaconda/modules/payloads/payload/payload_base_interface.py
jkonecny12: Hmm and I was wondering why I don't getting string in my tests... I should have investigate more before accepting that.

Thanks for pointing that out.

tests/nosetests/pyanaconda_tests/module_payload_live_os_test.py
jkonecny12: Added. Thanks for a good idea.

jkonecny12:
**UPDATED:**
* Fixed types are returning values.
* Implement a new test for PayloadFactory.

jkonecny12:
**UPDATED:**
* Rebased to solve conflicts.

jkonecny12:
jenkins, test this please

jkonecny12:
**UPDATED:**
* Rebased to solve conflict.

PR #2263, opened 2020-01-08
Do not crash on disk.wwn value being None (#1711571)

rvykydal:


rvykydal:
jenkins, test this please

PR #2265, opened 2020-01-09
Remove duplicated exception handler usage

VladimirSlavik:
Now exception handling is always on, with same effect as previous duplication.


VladimirSlavik:
Jenkins, could you test this again please?

VladimirSlavik:
Jenkins, could you please test this again?

PR #2267, opened 2020-01-09
Implement additional support for advanced storage

poncovka:
Add the DBus method `IsSupported` to all modules for advanced storage
and use it in UI. Always create and publish the DASD and ZFCP modules.

### DIFF
--- a/pyanaconda/ui/gui/spokes/advanced_storage.py
+++ b/pyanaconda/ui/gui/spokes/advanced_storage.py
@@ -18,7 +18,6 @@
 #
 from collections import namedtuple
 
-from blivet import arch
 from blivet.devices import DASDDevice, FcoeDiskDevice, iScsiDiskDevice, MultipathDevice, \
     ZFCPDiskDevice, NVDIMMNamespaceDevice
 
@@ -27,7 +26,7 @@ from pyanaconda.core.i18n import CN_, CP_
 from pyanaconda.storage.utils import try_populate_devicetree, apply_disk_selection, \
     filter_disks_by_names
 from pyanaconda.storage.snapshot import on_disk_storage
-from pyanaconda.modules.common.constants.objects import DISK_SELECTION, FCOE, ISCSI
+from pyanaconda.modules.common.constants.objects import DISK_SELECTION, FCOE, ISCSI, DASD
 from pyanaconda.modules.common.constants.services import STORAGE
 
 from pyanaconda.ui.gui.utils import timed_action
@@ -427,9 +426,6 @@ class ZPage(FilterPage):
 
     def setup(self, store, selected_names, disks):
         """ Set up our Z-page, but only if we're running on s390x. """
-        if not arch.is_s390():
-            return
-
         ccws = []
         wwpns = []
         luns = []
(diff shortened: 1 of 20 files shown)


pyanaconda/ui/gui/spokes/advanced_storage.py
```diff
@@ -427,9 +426,6 @@ def is_member(self, device):
 
     def setup(self, store, selected_names, disks):
         """ Set up our Z-page, but only if we're running on s390x. """
-        if not arch.is_s390():
-            return
```
jkonecny12: Shouldn't this be replaced by ``if not STORAGE.get_proxy(DASD).IsSupported():``?

tests/nosetests/pyanaconda_tests/module_dasd_test.py
```diff
@@ -42,6 +42,9 @@ def setUp(self):
         self.dasd_module = DASDModule()
         self.dasd_interface = DASDInterface(self.dasd_module)
 
+    def is_supported_test(self):
+        self.assertEqual(self.dasd_interface.IsSupported(), False)
```
jkonecny12: What will happen if this test will be started on S390?

poncovka: Many of our tests will probably fail on s390x. I can mock the architecture for this test, but I don't think our unit tests are ready to run on a different architecture in general.

pyanaconda/ui/gui/spokes/advanced_storage.py
poncovka: It doesn't have to be replaced. This page is removed from the notebook if the DASD module is not supported, so the page is never set up in that case.

poncovka:
> Also wouldn't it be better to have one base class which would provide `IsSupported()` everywhere?

I was considering it, but the modules don't share any other API at this moment and it doesn't looks like they will. I don't want to introduce a new DBus interface just for this method. It would be nice to get rid of this "duplication" of the code, but the interface wouldn't make any sense from the API point of view.

tests/nosetests/pyanaconda_tests/module_iscsi_test.py
```diff
@@ -61,6 +61,9 @@ def setUp(self):
         self.callback = PropertiesChangedCallback()
         self.iscsi_interface.PropertiesChanged.connect(self.callback)
 
+    def is_supported_test(self):
+        self.assertEqual(self.iscsi_interface.IsSupported(), False)
```
VladimirSlavik: Do I read this right as "_these tests will never run in a situation where iSCSI is actually configured_" ? Given the s390 expectations, it's fair I guess...

poncovka: Fixed.

poncovka:
Rebased. All tests are patched now.

tests/nosetests/pyanaconda_tests/module_dasd_test.py
poncovka: Anyway, I have fixed the tests.

poncovka:
The failing test is not related to this PR.

pyanaconda/ui/gui/spokes/advanced_storage.py
jkonecny12: Thanks for explanation.

tests/nosetests/pyanaconda_tests/module_dasd_test.py
jkonecny12: Thanks for fixing this. We want to move these tests to gating tests so it would be great to make them robust if possible.

PR #2269, opened 2020-01-09
Improve the protection of live devices

poncovka:
Raise the BootLoaderError instead of the IndexError, if there are no usable boot
drives, so the UI can handle the exception and show the error message to the user.

If the live device is not in the device tree, try to find the disk instead.
Otherwise, we will not be able to protect these devices.

Resolves: rhbz#1773412

### DIFF
--- a/pyanaconda/bootloader/execution.py
+++ b/pyanaconda/bootloader/execution.py
@@ -17,6 +17,7 @@
 #
 import blivet.arch
 from blivet.devices import iScsiDiskDevice
+from pyanaconda.bootloader.base import BootLoaderError
 from pykickstart.errors import KickstartParseError
 
 from pyanaconda.core.configuration.anaconda import conf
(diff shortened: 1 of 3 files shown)


poncovka:
Rebased and extended.

poncovka:
> Is that the right place to put the check? The call stack will be:
> 
> ```
> (...)
> BootloaderExecutor.execute
> BootloaderExecutor._apply_boot_drive
> BootloaderExecutor._get_boot_drive
> ```
> 
> You could also return None and raise in the applying method... I grant that it will do the same and is a few lines less, so maybe it isn't the right kind of thing to care about.

@VladimirSlavik, in general, when you return None, then the caller is responsible for checking the return value. It is very easy to forget to do this check and do operations that are not valid for None. This is a very common cause of the NoneType exceptions.

In this case, `_get_boot_drive` is a private method, so it doesn't really matter when we will raise the exception. However, the missing boot drive is a critical error in the context of this code, so it is better to end the execution as soon as possible.

> As an aside, the docstring for _apply... is likely wrong as it's the same as on _get...

The docstring is fixed.

PR #2273, opened 2020-01-13
Add scrolling for the registration tab

M4rtinK:
Add scrolling to the registration tab on the Connect to Red Hat spoke.
Otherwise if all options are filled in, the notification bar and
possibly also the "Register" button might get pushed off the screen.

Resolves: rhbz#1788422

### DIFF
--- a/pyanaconda/ui/gui/spokes/subscription.glade
+++ b/pyanaconda/ui/gui/spokes/subscription.glade
@@ -46,56 +46,391 @@
                 <child>
                   <object class="GtkNotebook" id="main_notebook">
                     <property name="visible">True</property>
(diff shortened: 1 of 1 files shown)


M4rtinK:
So this is how it looked like before the fix:
rhel_8 2_registration_no_tab_scrolling
It might no look wrong at first, but the info bar showing what went wrong actually got pushed off screen. This can be seen if you untick one of the options:
rhel_8 2_registration_no_tab_scrolling_1
With the fix, the infobar is always visible & all content of the registration tab is always accessible via scrolling:
rhel_8 2_registration_tab_scrolling

PR #2275, opened 2020-01-14
Cdrom protect port

rvykydal:


pyanaconda/storage/osinstall.py
```diff
@@ -287,6 +287,11 @@ def _mark_protected_devices(self):
                 protected.append(dev)
                 protected.extend(dev.parents)
 
+        # Prevents unmounting of cdrom devices during teardown.
+        # FIXME: This should based on conf.system (regular) not conf.target
+        if conf.target.is_image:
+            protected.extend(dev for dev in self.devicetree.devices if dev.type == "cdrom")
```
poncovka: I am a little worried about this change, because we might not be able to unmount our own mount points created by the payload. Anyway, the comment is not sufficient, because it took me a while to understand the problem. I would mention that during the image installation the method `setup_disk_images` marks all local storage disks as ignored, so they are protected from the teardown, but other devices need to be handled here. For now, we protect only cdroms, but we might need to add more types, ideally all except for the images. From that point of view, I would keep `conf.target.is_image` in the condition, because it seems to be related to the code in this case.

rvykydal: Agreed, I'll update the comment

rvykydal:
jenkins, test this please

PR #2276, opened 2020-01-14
Remove pyanaconda.flags.cmdline

VladimirSlavik:
Once kdump addon gets a new rawhide release, we can drop `pyanaconda.flags.cmdline` and leave only the replacement - `pyanaconda.kernel.kernel_arguments`. Also `getbool()` will be unused and removed then.

Blocked by https://github.com/daveyoung/kdump-anaconda-addon/pull/24

### DIFF
--- a/pyanaconda/core/kernel.py
+++ b/pyanaconda/core/kernel.py
@@ -103,8 +103,7 @@ class KernelArguments():
         inst_prefix = "inst."
 
         for i in lst:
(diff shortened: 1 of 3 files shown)


VladimirSlavik:
kdump addon is now in bodhi: https://bodhi.fedoraproject.org/updates/FEDORA-2020-810f9eb17d

PR #2277, opened 2020-01-14
Do not allow selecting ignored disk in GUI even if it is the only one

rvykydal:
~~DO NOT MERGE!~~

### DIFF
--- a/pyanaconda/ui/gui/spokes/storage.py
+++ b/pyanaconda/ui/gui/spokes/storage.py
@@ -819,7 +819,10 @@ class StorageSpoke(NormalSpoke, StorageCheckHandler):
 
         # if there's only one disk, select it by default
         if len(self.disks) == 1 and not self.selected_disks:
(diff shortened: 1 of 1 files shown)


jkonecny12:
I agree with @poncovka here. I wonder where the hidden disks are originally filtered out. There should be the correct fix.

rvykydal:
@jkonecny12: The disks are hidden because it is default for nvdimm disks. Ignored disks are hidden (by blivet). 
I think the fix is pretty straightforward, we just shouldn't be automatically selecting disks which are ignored. Maybe the comment should be:
"# if the disk is not ignored".

As for porting to master, the code seems to be changing so much when migrating from local storage instance to STORAGE service that it is bound to be solved from scratch there (if it is not actually fixed by the migration).

rvykydal:
Ported to Fedora by https://github.com/rhinstaller/anaconda/pull/2328

PR #2279, opened 2020-01-15
Add support for device factory changes

poncovka:
Move the code for adding a new device to a device tree to the new task
called `AddDeviceTask` and fix the related code.

Move the code for changing a device in a device tree to the new task called
`ChangeDeviceTask` and fix the related code.

Call the DBus method `ChangeDevice` to change a device in the storage model.

### DIFF
--- a/pyanaconda/modules/storage/partitioning/interactive/change_device.py
+++ b/pyanaconda/modules/storage/partitioning/interactive/change_device.py
@@ -1,0 +1,269 @@
+#
+# Copyright (C) 2020 Red Hat, Inc.
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# the GNU General Public License v.2, or (at your option) any later version.
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY expressed or implied, including the implied warranties of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General
+# Public License for more details.  You should have received a copy of the
+# GNU General Public License along with this program; if not, write to the
+# Free Software Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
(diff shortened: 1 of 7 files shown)


pyanaconda/modules/storage/partitioning/interactive/change_device.py
```diff
@@ -0,0 +1,268 @@
...
+
+        log.debug("Changing device size: %s", size)
+        resize_device(self._storage, self._device, size, original_size)
+
+    def _should_reformat_device(self):
+        """Should we reformat the device?
+
+        It's possible that reformat is requested but the format
+        type is unchanged, in which case we're not going to schedule
+        another reformat unless encryption has changed.
```
jkonecny12: I'm not quite sure this is correct. We should plan an action if device reformat is requested and format does not changed, because we want to erase data. Opposite is true. We don't want to reformat if the format changed because then the reformat action is already there.

The code below seems correct to me. Do I miss something here?

pyanaconda/modules/storage/partitioning/interactive/change_device.py
```diff
@@ -0,0 +1,268 @@
...
+                encrypted=False,
+                luks_version=luks_version
+            )
+
+            device = change_encryption(
+                storage=storage,
+                device=device,
+                encrypted=True,
+                luks_version=luks_version
+            )
```
jkonecny12: Why you have two calls here. One without encryption and one with encryption? Could you please add comment explaining that? Or abstract these calls somewhere to make the reason more obvious.

pyanaconda/modules/storage/partitioning/interactive/change_device.py
poncovka: Fixed.

pyanaconda/modules/storage/partitioning/interactive/change_device.py
poncovka: Fixed.

PR #2281, opened 2020-01-16
Rhel 8 fix status for manual org key registration

M4rtinK:
The aim of this pull request is to make sure, that correct subscription status is shown to users after a successful subscription attempt.

The first commit fixes a case where manually entered system purpose data has not been set to the installation environment, resulting in the data not being used at registration time and not displayed on the subscription status tab.

The second commit switches to a different DBus method for fetching overall subscription status as the current way of doing that turned out to be unreliable.

**Update**

Looks like this PR actually fixes not only rhbz#1788036, but also rhbz#1788207. :)

### DIFF
--- a/pyanaconda/modules/subscription/subscription.py
+++ b/pyanaconda/modules/subscription/subscription.py
@@ -27,7 +27,8 @@ from pyanaconda.core.signal import Signal
 from pyanaconda.modules.common.base import KickstartModule
 from pyanaconda.modules.common.constants.services import SUBSCRIPTION
 from pyanaconda.modules.common.constants.services import RHSM
-from pyanaconda.modules.common.constants.objects import RHSM_CONFIG
+from pyanaconda.modules.common.constants.objects import RHSM_CONFIG, RHSM_SYSPURPOSE, \
+        RHSM_ENTITLEMENT
 from pyanaconda.modules.subscription.subscription_interface import SubscriptionInterface
 from pyanaconda.modules.subscription.kickstart import SubscriptionKickstartSpecification
 from pyanaconda.modules.subscription.constants import AuthenticationMethod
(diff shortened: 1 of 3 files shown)


rvykydal:
I'd use the BZ to create pep8speak configuration for rhel-8 branch:
https://github.com/rhinstaller/anaconda/pull/2282

M4rtinK:
@rvykydal Sounds good to me, thanks! :)

pyanaconda/modules/subscription/subscription.py
```diff
@@ -844,6 +845,42 @@ def _unregister_task_callback(self, task_instance):
...
+        )
+        subscription_data_length = 0
+        # Log how much subscription data we got for debugging purposes.
+        # By only logging length, we should be able to debug cases of no
+        # or incomplete data being logged, without logging potentially
+        # sensitive subscription status detail in the installation logs stored
+        # on the target system.
+        if subscription_json:
+            subscription_data_length = len(subscription_json)
+        log.debug("RHSM: fetched subscription status data: %d characters",
```
jkonecny12: I think you should move this under the `if` otherwise you will log `0`.

pyanaconda/modules/subscription/subscription_data.py
```diff
@@ -173,49 +176,51 @@ def as_dict(self):
...
+        # fallback in case of the parsing fails
+        date_string = date_from_json
+        try:
+            # The start/end date in GetPools() output seems to be formatted as
+            # "Locale’s appropriate date representation.".
+            date = datetime.datetime.strptime(date_from_json, "%m/%d/%y")
+            # get a nice human readable date
+            date_string = date.strftime("%b %d, %Y")
+        except ValueError:
+            log.debug("subscription date parsing failed: %s", date_from_json)
```
jkonecny12: I would guess this is a warning.

jkonecny12:
jenkins, test this please

pyanaconda/modules/subscription/subscription.py
```diff
@@ -928,11 +933,15 @@ def _set_rhsm_log_level(self, log_level):
...
+    def _apply_syspurpose(self):
+        """Apply system purpose information to the installation environment.
+
+        If this method is called, then the token transfer installation task will
+        make sure to transfer the result, so the system purpose installation task
+        does not have to run afterwards.
 
-        The token transfer installation task will then make sure to transfer the result,
-        so the system purpose installation task does not have to run in a such a case.
+        For this reson se record if this method has run via the
```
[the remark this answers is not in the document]

pyanaconda/modules/subscription/subscription.py
M4rtinK: This is actually intended, so that we detect if nothing was returned by the API. But I guess a can add a separate warning level log message for that. :)

pyanaconda/modules/subscription/subscription_data.py
M4rtinK: Good point, warning makes more sense here.

jkonecny12:
jenkins, test this please

PR #2283, opened 2020-01-17
Add pep8 check

rvykydal:
* Example:

scripts/testing/setup-mock-test-env.py --init --copy --run-pep8-check --result results0 fedora-rawhide-x86_64
scripts/testing/setup-mock-test-env.py --init --copy --run-pep8-check "pyanaconda/modules/ pyanaconda/network.py" --result results0 fedora-rawhide-x86_64

make tests-pep8
make tests-pep8 PEP8_TARGETS="pyanaconda/modules/ pyanaconda/network.py"

Default PEP8_TARGETS are defined in tests/pep8/config.targets

* Configuration:

tests/pep8/config

It would be possible to share the configuration with pep8speaks in setup.cfg
located in the repository root but I think we might want to have the both checks
configured separately with some differences.

* Including into ci tests:

Add pep8/runpep8.sh to TESTS in tests/Makefile.am.
Configure targets to be tested in tests/pep8/config.targets

### DIFF
--- a/scripts/testing/dependency_solver.py
+++ b/scripts/testing/dependency_solver.py
@@ -36,6 +36,9 @@ TEST_DEPENDENCIES = ["e2fsprogs", "git", "bzip2", "cppcheck", "rpm-ostree", "pyk
                      "python3-mock", "python3-nose-testconfig", "python3-sphinx_rtd_theme",
                      "python3-lxml", "python3-pip", "python3-coverage",
 
+                     # pep8 check
+                     "python-pycodestyle",
+
                      # contains restorecon which was removed in Fedora 28 mock
                      "policycoreutils"
                      # "python3-pocketlint"
--- a/tests/pep8/config
+++ b/tests/pep8/config
@@ -1,0 +1,7 @@
+# Configuration file for pep8 checker (pycodestyle or pep8)
+
+[pycodestyle]
+max-line-length = 99
+
+[pep8]
+max-line-length = 99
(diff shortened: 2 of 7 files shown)


tests/pep8/config
```diff
@@ -0,0 +1,9 @@
+# Configuration file for pep8 checker (pycodestyle or pep8)
+
+[pycodestyle]
+max-line-length = 99
+# Ignored by default:
+#ignore = E121,E123,E126,E133,E226,E241,E242,E704,W503,W504,W505
```
jkonecny12: What is the reason to keep this here?

rvykydal: So we know what pycodestyle ignores by default. But I can remove it, it can be found in "man pycodestyle"

scripts/testing/dependency_solver.py
```diff
@@ -36,6 +36,11 @@
                      "python3-mock", "python3-nose-testconfig", "python3-sphinx_rtd_theme",
                      "python3-lxml", "python3-pip", "python3-coverage",
 
+                     # pep8 check
+                     "python-pycodestyle",
```
poncovka: Maybe we could install the pypi package instead of the rpm package to avoid issues we had with pylint.

scripts/testing/dependency_solver.py
```diff
@@ -36,6 +36,11 @@
...
+                     # can't find
+                     # "python-pep8",
```
poncovka: I would remove this comment. It is a little confusing.

tests/pep8/config
poncovka: I would remove it.

rvykydal: Done

PR #2288, opened 2020-01-23
Make sure RHSM HTTP proxy port from kickstart is an integer

M4rtinK:
Make sure RHSM HTTP proxy port from kickstart is always an integer
before being passed to set_server_proxy(). Also make sure -1 is
passed if proxy port is unknown.

Resolves: rhbz#1794025

### DIFF
--- a/pyanaconda/modules/subscription/subscription.py
+++ b/pyanaconda/modules/subscription/subscription.py
@@ -230,8 +230,11 @@ class SubscriptionModule(KickstartModule):
             try:
                 proxy = util.ProxyString(data.rhsm.proxy)
                 if proxy.host:
+                    # ensure port is an integer and set to -1 if unknown
+                    port = int(proxy.port) if proxy.port else -1
+
                     self.set_server_proxy(hostname=proxy.host,
-                                          port=proxy.port,
+                                          port=port,
                                           username=proxy.username,
                                           password=proxy.password)
             except util.ProxyStringError as e:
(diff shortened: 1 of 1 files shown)


pyanaconda/modules/subscription/subscription.py
```diff
@@ -229,8 +229,13 @@ def process_kickstart(self, data):
             try:
                 proxy = util.ProxyString(data.rhsm.proxy)
                 if proxy.host:
+                    # ensure port is an integer and set to -1 if unknown
+                    port = -1
+                    if proxy.port:
+                        port = int(proxy.port)
```
poncovka: Please, use `port = int(proxy.port) if proxy.port else -1`.

pyanaconda/modules/subscription/subscription.py
```diff
@@ -801,19 +806,28 @@ def set_server_proxy(self, hostname, port, username, password):
...
         # negative port number means no port has been set
         if port < 0:
             port_string = ""
         else:
             port_string = "{}".format(port)
+        # use "" when username is not known
+        username_string = ""
+        if username:
+            username_string = username
+        self._server_proxy_user = username_string
```
poncovka: Please, use `self._server_proxy_user = username or ""`.

pyanaconda/modules/subscription/subscription.py
```diff
@@ -801,19 +806,28 @@ def set_server_proxy(self, hostname, port, username, password):
...
+        # use "" when password is not known
+        password_string = ""
+        if password:
+            password_string = password
+        self._server_proxy_password = password_string
```
poncovka: Please, use `self._server_proxy_password = password or ""`.

pyanaconda/modules/subscription/subscription.py
M4rtinK: Hmm, yeah, this looks better. Changed.

pyanaconda/modules/subscription/subscription.py
M4rtinK: Changed.

pyanaconda/modules/subscription/subscription.py
M4rtinK: Changed.

PR #2290, opened 2020-01-24
Support ksstrict in modules

VladimirSlavik:
Propagate kickstart warnings from modules, handle them in UI if asked for with `ksstrict`.

### DIFF
--- a/pyanaconda/kickstart.py
+++ b/pyanaconda/kickstart.py
@@ -45,7 +45,7 @@ from pyanaconda.pwpolicy import F22_PwPolicy, F22_PwPolicyData
 
 from pykickstart.base import BaseHandler, KickstartCommand
 from pykickstart.constants import KS_SCRIPT_POST, KS_SCRIPT_PRE, KS_SCRIPT_TRACEBACK, KS_SCRIPT_PREINSTALL
-from pykickstart.errors import KickstartError
+from pykickstart.errors import KickstartError, KickstartParseWarning
 from pykickstart.parser import KickstartParser
 from pykickstart.parser import Script as KSScript
 from pykickstart.sections import NullSection, PackageSection, PostScriptSection, PreScriptSection, PreInstallScriptSection, \
@@ -573,15 +573,13 @@ def parseKickstart(handler, f, strict_mode=False, pass_to_boss=False):
 
     ksparser = AnacondaKSParser(handler)
     kswarnings = []
-    ksmodule = "pykickstart"
-    kscategories = (UserWarning, SyntaxWarning, DeprecationWarning)
     showwarning = warnings.showwarning
 
     def ksshowwarning(message, category, filename, lineno, file=None, line=None):
         # Print the warning with default function.
         showwarning(message, category, filename, lineno, file, line)
         # Collect pykickstart warnings.
-        if ksmodule in filename and issubclass(category, kscategories):
+        if issubclass(category, KickstartParseWarning):
             kswarnings.append(message)
 
     try:
@@ -590,9 +588,7 @@ def parseKickstart(handler, f, strict_mode=False, pass_to_boss=False):
 
             # Set up the warnings module.
             warnings.showwarning = ksshowwarning
-
-            for category in kscategories:
-                warnings.filterwarnings(action="always", module=ksmodule, category=category)
+            warnings.simplefilter("always", category=KickstartParseWarning)
 
             # Parse the kickstart file in DBus modules.
             if pass_to_boss:
@@ -600,6 +596,8 @@ def parseKickstart(handler, f, strict_mode=False, pass_to_boss=False):
                 report = KickstartReport.from_structure(
                     boss.ReadKickstartFile(f)
                 )
+                for warn in report.warning_messages:
+                    warnings.warn(warn.message, KickstartParseWarning)
                 if not report.is_valid():
                     message = "\n\n".join(map(str, report.error_messages))
                     raise KickstartError(message)
--- a/pyanaconda/modules/common/base/base.py
+++ b/pyanaconda/modules/common/base/base.py
@@ -19,10 +19,11 @@
 # Red Hat, Inc.
 #
 import os
+import warnings
 from abc import ABC
 from locale import setlocale, LC_ALL
 
-from pykickstart.errors import KickstartError
+from pykickstart.errors import KickstartError, KickstartParseWarning
 
 from pyanaconda.core.event_loop import EventLoop
 from pyanaconda.core.timer import Timer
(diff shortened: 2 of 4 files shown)


pyanaconda/kickstart.py
```diff
@@ -682,6 +679,8 @@ def ksshowwarning(message, category, filename, lineno, file=None, line=None):
                 if not report.is_valid():
                     message = "\n\n".join(map(str, report.error_messages))
                     raise KickstartError(message)
+                if strict_mode:
```
poncovka: I think that you can always add warnings from DBus modules to `kswarnings`. I suggest to do that before raising  `KickstartError`, because we print the collected warnings in the exception handler in the strict mode.

pyanaconda/kickstart.py
```diff
@@ -669,9 +668,7 @@ def ksshowwarning(message, category, filename, lineno, file=None, line=None):
 
             # Set up the warnings module.
             warnings.showwarning = ksshowwarning
-
-            for category in kscategories:
-                warnings.filterwarnings(action="always", module=ksmodule, category=category)
+            warnings.filterwarnings(action="always", module=ksmodule, category=KickstartParseWarning)
```
poncovka: I think that we might be able to use `simplefilter` now. We don't need to check the module, because there is a special category for kickstart warnings.

pyanaconda/kickstart.py
```diff
@@ -653,14 +653,13 @@ def parseKickstart(handler, f, strict_mode=False, pass_to_boss=False):
...
     ksmodule = "pykickstart"
-    kscategories = (UserWarning, SyntaxWarning, DeprecationWarning)
     showwarning = warnings.showwarning
 
     def ksshowwarning(message, category, filename, lineno, file=None, line=None):
         # Print the warning with default function.
         showwarning(message, category, filename, lineno, file, line)
         # Collect pykickstart warnings.
-        if ksmodule in filename and issubclass(category, kscategories):
+        if ksmodule in filename and issubclass(category, KickstartParseWarning):
```
poncovka: It should be enough to check the category now.

pyanaconda/kickstart.py
```diff
@@ -682,6 +679,8 @@ def ksshowwarning(message, category, filename, lineno, file=None, line=None):
                 if not report.is_valid():
                     message = "\n\n".join(map(str, report.error_messages))
                     raise KickstartError(message)
+                if strict_mode:
+                    kswarnings += report.warning_messages
```
poncovka: I am afraid that warnings from the DBus modules are not going to be logged in a non-strict mode.

pyanaconda/kickstart.py
```diff
@@ -653,14 +653,13 @@ def parseKickstart(handler, f, strict_mode=False, pass_to_boss=False):
...
-    kscategories = (UserWarning, SyntaxWarning, DeprecationWarning)
     showwarning = warnings.showwarning
 
     def ksshowwarning(message, category, filename, lineno, file=None, line=None):
         # Print the warning with default function.
         showwarning(message, category, filename, lineno, file, line)
         # Collect pykickstart warnings.
-        if ksmodule in filename and issubclass(category, kscategories):
+        if ksmodule in filename and issubclass(category, KickstartParseWarning):
             kswarnings.append(message)
```
poncovka: I would investigate whether we can use `warnings.catch_warnings(record=True)` to collect warnings.

pyanaconda/kickstart.py
poncovka: I would consider to "re-issue" the remote warnings here to make sure that they are processed the same way as the local warnings.

pyanaconda/modules/common/base/base.py
```diff
@@ -222,11 +223,31 @@ def read_kickstart(self, s):
...
+                report.warning_messages.append(data)
+            # No need to print the warning here, the UI will take care of that
+
         try:
             handler = self.get_kickstart_handler()
             parser = self.get_kickstart_parser(handler)
-            parser.readKickstartFromString(s)
-            self.process_kickstart(handler)
+
+            with warnings.catch_warnings():
```
poncovka: I am pretty sure that we can use `warnings.catch_warnings(record=True)` here to avoid the function `log_ks_warning`.

poncovka:
Please, write a simple test for the warning handler in a DBus module. Just add a new kickstart test for an existing DBus module. The method `check_kickstart_interface` returns a kickstart report that can be checked further.

pyanaconda/kickstart.py
VladimirSlavik: Thank you - done.

pyanaconda/kickstart.py
VladimirSlavik: Yes, this seems to work.

VladimirSlavik:
Update:
- Fix all feedback from above
- Add test

tests/nosetests/pyanaconda_tests/module_timezone_test.py
```diff
@@ -188,6 +189,16 @@ def install_with_tasks_configured_test(self, publisher):
         self.assertIsInstance(ntp_obj.implementation, ConfigureNTPTask)
         self.assertEqual(ntp_obj.implementation._ntp_enabled, False)
 
+    def deprecated_warnings_test(self):
+        response = self.timezone_interface.ReadKickstart("timezone --isUtc Europe/Bratislava")
+        report = KickstartReport.from_structure(response)
+        warning_found = False
+        for warning in report.warning_messages:
```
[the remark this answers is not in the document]
VladimirSlavik: Rewritten to expect exactly one warning.

pyanaconda/kickstart.py
VladimirSlavik: Rewritten to re-issue these warnings.

VladimirSlavik:
Thank you, updated per our discussion (I hope!).

pyanaconda/kickstart.py
```diff
@@ -669,16 +668,17 @@ def ksshowwarning(message, category, filename, lineno, file=None, line=None):
...
+                                    category=KickstartParseWarning)
 
             # Parse the kickstart file in DBus modules.
             if pass_to_boss:
                 boss = BOSS.get_proxy()
                 report = KickstartReport.from_structure(
                     boss.ReadKickstartFile(f)
                 )
+                for warn in report.warning_messages:
+                    warnings.warn(warn.message)
```
poncovka: Please, set the category to `KickstartParseWarning`, otherwise these warnings will not be collected in `kswarnings`.

pyanaconda/kickstart.py
poncovka: You can remove the check for `ksmodule`. It is enough to check the category.

pyanaconda/kickstart.py
```diff
@@ -669,16 +668,17 @@ def ksshowwarning(message, category, filename, lineno, file=None, line=None):
 
             # Set up the warnings module.
             warnings.showwarning = ksshowwarning
-
-            for category in kscategories:
-                warnings.filterwarnings(action="always", module=ksmodule, category=category)
+            warnings.filterwarnings(action="always", module=ksmodule,
```
poncovka: You can still use `simplefilter` here and remove the argument `module`.

PR #2292, opened 2020-01-25
[WIP] Use unsquashfs to install the Live system

bkhomuts:
This will allow for multi-threaded decompression and will result in lower installation time.
The unsquashfs command is more efficient than rsync to decompress the entire image.

Note, that this patch will crash anaconda installer.
I did not investigate why exactly. The unsquashfs command will complete and you'll see Fedora installed in /mnt/sysimage, but other steps will not be performed.

This patch requires polishing before merging:
0. Investigate and fix the crash, so that unsquashfs can be used to install the Live image.
1. Remove assumption that squashfs is always plain, a check should be added.
2. Remove assumption, that the installation is performed from local media, see comments in the patch.
   I'd suggest falling back to /dev/loopX in case the file in /run/initramfs is not found.

This is a follow-up effort of optimizing the squashFS image size: https://fedoraproject.org/wiki/Category:Changes/OptimizeSquashFS

### DIFF
--- a/pyanaconda/payload/live/payload_liveimg.py
+++ b/pyanaconda/payload/live/payload_liveimg.py
@@ -287,7 +287,7 @@ class LiveImagePayload(BaseLivePayload):
             Otherwise fall back to rsync of INSTALL_TREE
         """
         # If it doesn't look like a tarfile use the super's install()
-        if not self.is_tarfile:
+        if not self.is_tarfile and not self.is_plain_squashfs_image:
             super().install()
             return
 
--- a/pyanaconda/payload/live/payload_liveos.py
+++ b/pyanaconda/payload/live/payload_liveos.py
@@ -16,18 +16,23 @@
 # Red Hat, Inc.
 #
 import os
+import re
 import stat
 
 from blivet.size import Size
 from pyanaconda.anaconda_loggers import get_packaging_logger
-from pyanaconda.core.constants import PAYLOAD_TYPE_LIVE_OS, INSTALL_TREE, SOURCE_TYPE_LIVE_OS_IMAGE
+from pyanaconda.core.configuration.anaconda import conf
+from pyanaconda.core.constants import PAYLOAD_TYPE_LIVE_OS, INSTALL_TREE, \
+    SOURCE_TYPE_LIVE_OS_IMAGE, THREAD_LIVE_PROGRESS
 from pyanaconda.core.i18n import _
+from pyanaconda.core.util import execInSysroot, execWithCapture, execWithRedirect
 from pyanaconda.errors import errorHandler, ERROR_RAISE
 from pyanaconda.modules.common.constants.services import PAYLOADS
 from pyanaconda.payload import utils as payload_utils
 from pyanaconda.payload.errors import PayloadInstallError, PayloadSetupError
 from pyanaconda.payload.live.payload_base import BaseLivePayload
 from pyanaconda.progress import progressQ
+from pyanaconda.threading import threadMgr, AnacondaThread
 
 log = get_packaging_logger()
 
@@ -55,6 +60,27 @@ class LiveOSPayload(BaseLivePayload):
 
         source_proxy = PAYLOADS.get_proxy(source_path)
         return source_proxy.DetectLiveOSImage()
+
+    def get_number_of_inodes(self):
+        """ Examine the contents of the filename to determine whether it's a plain squashfs.
+        " If the return value is True, then the image is considered plain. This means
+        " it does not contain any embedded filesystem inside. If the value returned is False,
+        " it means that the filesystem has an embedded, in Fedora -- ext4, filesystem inside
+        " a squashfs image.
+        " :param: self.osimg_path -- a path to the target file or a block device
+        " :returns: bool
+        """
+        search_string = r"Number of inodes (\d+)"
+        file_squashfs_information = execWithCapture("unsquashfs", ["-s", self.osimg_path])
+        match_obj = re.search(search_string, file_squashfs_information)
+        try:
+            number_of_inodes = int(match_obj.group(1))
+        except AttributeError as e:
+            log.debug("The filesystem is either non-plain or an error has occured")
+            debug_message = str(e)
+            log.debug(debug_message)
+            return -1
+        return number_of_inodes
 
     def setup(self):
         super().setup()
@@ -94,6 +120,48 @@ class LiveOSPayload(BaseLivePayload):
         """ Perform pre-installation tasks. """
         super().pre_install()
         progressQ.send_message(_("Installing software") + (" %d%%") % (0,))
+
+    def install(self):
+        # Define a number of files from which a filesystem is considered plain
+        plain_squashfs_threshold = 50
+        if self.get_number_of_inodes() < plain_squashfs_threshold:
+            # Proceed with the standard installation.
+            super().install()
+            return
+        # Use an optimization in case the SQUASHFS image is plain.
+        # Include only the directories from the squashfs as specified below.
+
+        cmd = "unsquashfs"
+        args = ["-f", "-n", "-d", conf.target.system_root, self.osimg_path]
+
+        try:
+            rc = execWithRedirect(cmd, args)
+        except (OSError, RuntimeError) as e:
+            msg = None
+            err = str(e)
+            log.error(err)
+        else:
+            err = None
+            msg = "%s exited with code %d" % (cmd, rc)
+            log.info(msg)
+        if err or rc == 11:
+            exn = PayloadInstallError(err or msg)
+            if errorHandler.cb(exn) == ERROR_RAISE:
+                raise exn
+        # Wait for progress thread to finish
+        with self.pct_lock:
+            self.pct = 100
+        threadMgr.wait(THREAD_LIVE_PROGRESS)
+
+        # This will cleanup files that are by default in the SquashFS image
+        # and are not wanted in the target system. The parent directories will be preserved
+        # Ideally those should be not present in the image in the first place.
+        find_arguments = ["/boot/loader", "/tmp", "-mindepth", "1", "-delete"]
+        execInSysroot("find", find_arguments)
+        # Live needs to create the rescue image before bootloader is written
+        self._create_rescue_image()
+
+
 
     @property
     def space_required(self):


VladimirSlavik:
Hi @bkhomuts , thank you for your work on this! I don't know much about this area, but I would just quickly raise two points:

a) You mention speed as the rationale. Unless rsync does something we don't need, there must be some tradeoff? I found some mentions of memory usage being far higher for unsquashfs. How does that comparison look, and how does it compare with the minimal viable system configuration that we claim to support? See eg. the [RHEL](https://access.redhat.com/articles/rhel-limits#minimum-required-memory-3) values, which is up to date (as opposed to various Fedora docs...) and says 1.5 GB.

b) Your changes remove the comment about all that is preserved, and options that prevent overwriting some stuff. Is that going to work with unsquashfs? Because you mention it does not work as a whole, and, well, this sounds like this could be a possible cause...

bcl:
> Your changes remove the comment about all that is preserved,

That's my biggest concern about using unsquashfs, I need to take a closer look at it to make sure it can handle xattrs, etc.

> options that prevent overwriting some stuff. Is that going to work with unsquashfs? Because you mention it does not work as a whole, and, well, this sounds like this could be a possible cause...

The rsync --exclude options are there because it is copying a live mounted filesystem, they should be unneeded for unsquashfs.

In general I like this idea, but one thing to keep in mind is that rsync is still useful for uncompressed image support so we need to make sure that still works.

jkonecny12:
Hi @bkhomuts could you also please fix the @pep8speaks errors in your code.

Thanks!

jkonecny12:
Jenkins, it's ok to test.

M4rtinK:
> In general I like this idea, but one thing to keep in mind is that rsync is still useful for uncompressed image support so we need to make sure that still works.

Yeah, I think this will need to be a new codepath while using the current rsync based codepath in place, both for the uncompressed images and in case we hit some functionality regressions later on and might have to (temporarily ?) revert to using rsync.

bkhomuts:
> That's my biggest concern about using unsquashfs, I need to take a closer look at it to make sure it can handle xattrs, etc.

Yes, unsquashfs handles everything that's needed for Fedora root filesystem. This includes extended attributes. So, the functionality is similar to the RSYNC.


> a) You mention speed as the rationale. Unless rsync does something we don't need, there must be some **tradeoff**? 

That's not the case here. The unsquashfs is faster because it's aware of the filesystem structure. For example, it may read squashFS file sequentially. That will speed up the decompression from slow installation media.

> b) Your changes remove the comment about all that is preserved, and options that prevent overwriting some stuff. Is that going to work with unsquashfs? Because you mention it does not work as a whole, and, well, this sounds like this could be a possible cause...
I'm not sure. I don't have resources to test and debug further.

I found some mentions of memory usage being far higher for unsquashfs. How does that comparison look, and how does it compare with the minimal viable system configuration that we claim to support? See eg. the RHEL values, which is up to date (as opposed to various Fedora docs...) and says 1.5 GB.
I measured memory consumption of unsquashfs process. By default it peaks at ~540MiB on a 4-core machine. Doubling the number of cores to 8 leads to peak consumption of ~555MiB.
The high memory usage is due to extreme queuing of data, to optimize CPU usage. This can be adjusted with flags -data-queue and -fragment-queue. I found no difference in decompression time when reducing the data and fragment queue by half to 128MiB. But I think, the memory consumption of default options is alright even for low-end machines. This could be implemented in a smart way by polling the available memory on the machine and adjusting above mentioned options accordingly.

>Hi @bkhomuts could you also please fix the @pep8speaks errors in your code.
Thanks!
> Yeah, I think this will need to be a new codepath while using the current rsync based codepath in place, both for the uncompressed images and in case we hit some functionality regressions later on and might have to (temporarily ?) revert to using rsync.

I agree, there should be two code paths to support old squashfs+ext4 and new 'plain' squash filesystem.
Unfortunately, I don't have resources to work on this patch to finalize it. I just proposed a concept how it should look like. I'll monitor this ticket and will advise, if there will be more questions.

jkonecny12:
Hello @bkhomuts, thanks a lot for this PR and idea. Right now, we want to migrate payload to a module so this code will be moved/changed. Also we should make a performance checks on our sides first to give you proper answers.

I would like to postpone this until everything mentioned above is solved and then we can return back to this PR. If this is a priority for some reason then please tell us here in the PR. If it's not a priority than please, bear with us for some time.

Adding blocked flag as blocked by Anaconda modularization.

bkhomuts:
> Hello @bkhomuts, thanks a lot for this PR and idea. Right now, we want to migrate payload to a module so this code will be moved/changed. Also we should make a performance checks on our sides first to give you proper answers.

My pleasure to work on this. I've done draft performance verifications on my side. I can see almost 2 times improvement in the installation time, and that's when the performance is not CPU-bound.

- Default configuration - 284s
- Using the patch - 150s [1]
Note, I performed testing on Fedora 32 Beta image 
CPU: 4 vCPU of Ryzen 5 3600.
[1] You can get the ISO at: https://khomutsky.com/fedora-dvd/Fedora-Workstation-dvd-x86_64-32_Beta-1.2-bkhomuts_squashfs_XZ_plain.iso


You could try applying this patch on a live system and performing the installation. The slower the single-core performance of the processor, the more visible the effect of the speed-up will be...

The patch is not perfect at this point, but the concept can be verified and performance measured. It still requires adding the checks and determining how to not install unwanted files, such as /etc/machineid, to the target system. The unsquashfs does not support excluding files, so most likely a removal of those will be needed from the target system. I plan to work on this in the coming weekends.

> I would like to postpone this until everything mentioned above is solved and then we can return back to this PR. If this is a priority for some reason then please tell us here in the PR. If it's not a priority than please, bear with us for some time.

This is not a priority right now, but considering the triviality of the change and benefit to the end user, I'd personally would like to have this merged sooner rather than later.

jkonecny12:
Feel free to work this we will try to answer your questions if we will have time.
I just want to warn you about possible issues:
- excludes have to be solved (as you already mentioned)
- we have to support not only images but also tar payloads
- think also about a note at https://github.com/rhinstaller/anaconda/pull/2292#issuecomment-578877767 that the rsync is still useful to have for uncompressed images.

bkhomuts:
Hello @jkonecny12 , I have edited my change to be compatible with behavior of existing installation process.
In case the plain squashfs is determined on the installation media, then the optimization will be used. Otherwise, rsync will be used.

I also approached the problem of not having `--exclude` option in unsquashfs by removing the files from the target system after unpacking of  the squashfs is complete. Not all files that are specified by exclusion filter for rsync are in fact present in the 'plain' squashfs image, therefore I removed extra exclude filters from the `if plain_squashfs_image:` section of the program.

All your previous concerns:

> * excludes have to be solved (as you already mentioned)
> * we have to support not only images but also tar payloads
> * think also about a note at #2292 (comment) that the rsync is still useful to have for uncompressed images.

Are now resolved.

I also opened a merge request to f32-devel branch #2436. Hopefully this change will be merged sooner rather than later.

bkhomuts:
I have re-based this change on top of master, therefore you may see a big number of commits in this change.

jkonecny12:
> I also approached the problem of not having `--exclude` option in unsquashfs by removing the files from the target system after unpacking of the squashfs is complete. Not all files that are specified by exclusion filter for rsync are in fact present in the 'plain' squashfs image, therefore I removed extra exclude filters from the `if plain_squashfs_image:` section of the program.

Just a question. How will unsquashfs deal with a situation that some files in the iso can't be written? Is it going to ignore these or abort?

bkhomuts:
> Just a question. How will unsquashfs deal with a situation that some files in the iso can't be written? Is it going to ignore these or abort?

To answer this question, I performed a computer simulation of such failure: by decompressing the squash filesystem to a device with insufficient free storage space available.
In case of error, the _unsquashfs_ utility will continue decompressing, while returning errors to _standard error_. In the end, the program will return the exit code 0.

I see a room for improvement in this patch set. There's a need to perform a verification.

I propose to do it in the following way: compare the number of files the _unsquashfs_ utility planned to decompress with actually decompressed files count. Precise digits could be calculated by the formula:
number of... inodes = files + symbolic links + devices + FIFOs

All the necessary information for error verification is present in the output of the _unsquashfs_. This information should be used to ensure the correct installation.

jkonecny12:
Thanks a lot for doing this test and sharing result.

cmurf:
> Just a question. How will unsquashfs deal with a situation that some files in the iso can't be written? Is it going to ignore these or abort?

If it's on write to the target, I'd expect EIO and unsquashfs should stop. If it's on read, there's no integrity checking at all for ISO9660/UDF/squashfs. Same if rsync is copying.

It's a separate feature request, but I'd really like to see the isomd5sum check go away, because users skip it, it's slow, one time only, and USB flash media can have transient read errors, even silent corruption. And instead use dmverity/dmintegrity/btrfs to provide integrity checking on every read, and EIO if it fails. A dmverity image also provides authenticity, and optionally includes RS ECC, so it could somewhat cheaply correct for small errors, if desired.

keszybz:
> In case of error, the _unsquashfs_ utility will continue decompressing, while returning errors to _standard error_. In the end, the program will return the exit code 0.

That'd mean `unquashfs` is fundamentally broken. Let's fix it, instead of piling up workarounds on top.

plougher:
> > In case of error, the _unsquashfs_ utility will continue decompressing, while returning errors to _standard error_. In the end, the program will return the exit code 0.
> 
> That'd mean `unquashfs` is fundamentally broken. Let's fix it, instead of piling up workarounds on top.

In the case of I/O error or out of space Unsquashfs will abort instantly, and return an error status code.
 
See section 6 of the release notes for Unsquashfs 4.4 (https://github.com/plougher/squashfs-tools/blob/master/README-4.4)

<pre>
phillip@phoenix:/tmp$ sudo unsquashfs -d /mnt3/squashfs-root /mnt/casper/filesystem.squashfs
Parallel unsquashfs: Using 4 processors
186383 inodes (206701 blocks) to write

[=/                                                                                                                                     ]   2123/206701   1%
Write on output file failed because No space left on device

FATAL ERROR:writer: failed to write file /mnt3/squashfs-root/etc/libreoffice/registry/res/registry_fr.xcd
phillip@phoenix:/tmp$ echo $?
1
</prev>

bkhomuts:
@plougher , I think I tested with SquashFS-tools version 4.3 on Fedora, that is why I've got different results.

> 6. Unsquashfs is now more strict about error handling.
This behavior is new to version 4.4.

PR #2294, opened 2020-01-27
[WIP] Fix long lines

VladimirSlavik:
This is an attempt to fix too long lines (`C0301` aka too-long-line) in the whole `pyanaconda` codebase.

Given there are more possibilities how to break lines (styles?), I think we should have some discussion about what is preferred, why etc. This PR as of now is only a starting point for this discussion - I actually mixed various styles, depending on the context and what looked better.

To give some examples:
```
call(foo(bar(
    "too long string")))
call(many_parameters, just_so_many, 
     that, it_is_too, long_to_fit)
call(_("some text") % (foo,
     bar, baz))
```
vs
```
call(foo(bar("too long"
   " string")))
call(many_parameters,
     just_so_many, 
     that,
     it_is_too,
     long_to_fit)
call(_("some text")
     % (foo, bar, baz))
```

Then there is also the question of replacing % with format, and whether that should be part of fixing line lengths... etc...


poncovka:
I prefer this style:

```
call(foo(bar(
    "too long string"
)))

call(
    many_parameters,
    just_so_many, 
    that,
    it_is_too,
    long_to_fit
)

call(_("some text").format(
    foo, bar, baz
))
```

jkonecny12:
> Then there is also the question of replacing % with format, and whether that should be part of fixing line lengths... etc...

It's not a question we have guidelines that only `format` should be used. If it should be part depends on you, it's your PR :).

I agree with @poncovka on the style question.

VladimirSlavik:
Jenkins, please test this.

poncovka:
This PR is no longer blocked.

VladimirSlavik:
As we keep refactoring, the diff gets stale fast. With pep8check running on all PRs, we also keep fixing long lines by default. I think this is no longer relevant, so closing.

PR #2298, opened 2020-01-30
network: fix GUI crash on invalid devices in the list

rvykydal:
Resolves: rhbz#1697256

This fix only tries to make sure GUI does not crash on invalid devices in
device list. Making sure such devices are not there is for another fix.

The invalid devices are appearing for example when vlan device IP configuration
fails.


rvykydal:
Ported to Fedora in https://github.com/rhinstaller/anaconda/pull/2328

rvykydal:
Ported to master with https://github.com/rhinstaller/anaconda/pull/2337

PR #2299, opened 2020-01-30
Disable package download to / /tmp (#1781517)

jkonecny12:
The `/` and `/tmp` mount points aren't good candidates for package download.

During stage2 installation they are on TMPFS or squashfs. Both these systems gives you an abstract number when you asking for a free space. It's getting even more complicated in case when zram/zswap is involved.

During the image install you have /tmp as TMPFS and / as your root. This is pretty similar situation, we don't want to download packages there neither.

We are not able reliably tell user that the packages will be downloaded
correctly.

*Resolves: rhbz#1781517*

### DIFF
--- a/pyanaconda/payload/dnfpayload.py
+++ b/pyanaconda/payload/dnfpayload.py
@@ -67,9 +67,7 @@ DNF_CACHE_DIR = '/tmp/dnf.cache'
 DNF_PLUGINCONF_DIR = '/tmp/dnf.pluginconf'
 DNF_PACKAGE_CACHE_DIR_SUFFIX = 'dnf.package.cache'
 DNF_LIBREPO_LOG = '/tmp/dnf.librepo.log'
-DOWNLOAD_MPOINTS = {'/tmp',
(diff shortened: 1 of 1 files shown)


bcl:
How about also sorting by size and picking the largest one?

jkonecny12:
> How about also sorting by size and picking the largest one?

Not sure if it's really helpful. Biggest size doesn't mean most space for the package download because we don't know how much space will be taken by the installation. I would rather stay conservative here and set priority by us.

poncovka:
> How about also sorting by size and picking the largest one?

I think we already do that (at least on Fedora, I haven't checked RHEL).

jkonecny12:
> > How about also sorting by size and picking the largest one?
> 
> I think we already do that (at least on Fedora, I haven't checked RHEL).

Heh, you are right we are doing that even on RHEL https://github.com/rhinstaller/anaconda/blob/rhel-8/pyanaconda/payload/dnfpayload.py#L161 .

PR #2300, opened 2020-01-31
Use opts in geoloc instead of kernel_arguments

VladimirSlavik:
This makes `geoloc` behave like a normal Anaconda module should - read boot options from `opts` instead of looking at kernel cmdline with `kernel_arguments`.

First commit is the actual switch,
second cleans up the surrounding code,
third cleans up style & conventions in the file.


VladimirSlavik:
Thank you, typos fixed!

PR #2301, opened 2020-01-31
Change "not...in" to "...not in"

VladimirSlavik:
Eradicates PEP8 message E713. Questions:
a) Do we want this?
b) If yes, lists or tuples?


poncovka:
> a) Do we want this?

Yes.

> b) If yes, lists or tuples?

It depends. We should use sets for a higher number of items. Otherwise, we should prefer tuples, because they are faster to create. However, I don't really think it will have a very big impact in these cases, so you can keep it as it is.

PR #2304, opened 2020-02-03
Fix network & CDN installation source switching

M4rtinK:
If the user requested a custom network installation source
(say via the inst.repo boot option), keep it as installation
source even if the system is then registered.

Also make sure that user can switch from such a network
installation source to CDN and back.

Resolves: rhbz#1791376

**NOTE:** WiP, not yet to be merged


M4rtinK:
Superseded by #2319.

PR #2307, opened 2020-02-06
Check text of nicCombo before split it,fix bug1798402

skycastlelily2:


rvykydal:
Thank you for the patch.
I have some suggestions:
- I'd return empty string instead of None from nic property in case there are no nics so we use the same type (string).
- Instead of raising error message, I'd rather prevent trying to add FCoE in case there are no nics by making the "Add FCoE Disk(s)" button insensitive. It also seems that with your patch there is no way out of the dialog ("Cancel" button is insensitive).
Screenshot from 2020-02-06 11-18-53
The solution with "Add FCoE Disk(s)" disabled:
Screenshot from 2020-02-06 11-38-32

So the solution modified with my suggestions could be:
https://github.com/rhinstaller/anaconda/pull/2311/files

skycastlelily2:
Mine is ugly,gonna to close it

PR #2313, opened 2020-02-10
Handle the variant from GetResult (#1798392)

poncovka:
The method GetResult of the interface TaskInterface returns a variant.
Unwrap or unpack the variant, so we can work with the variant value.

Resolves: rhbz#1798392


poncovka:
Anyway, dasbus should provide a better error. See: https://github.com/rhinstaller/dasbus/pull/22

PR #2314, opened 2020-02-10
Add missing logs to the log-capture utility

jkonecny12:
The log-capture utility seems like a good thing for bug reports. However, it's missing a few interesting logs.

**Question:**
Do we want to add this to ``/usr/bin/``. Right now users have to type ``/usr/libexec/anaconda/log-capture`` to invoke this script. Installing this as a `anaconda-log-capture` could be easier to discover for users.

This could be even part of the documentation.

**Further improvements:**
We may extend this to automatically upload the tar file to an existing bugzilla with a BZ number as argument.

### DIFF
--- a/scripts/log-capture
+++ b/scripts/log-capture
@@ -50,8 +50,10 @@ _to_log lsof
 _to_log ip -s li
 _to_log ip a
 _to_log ip r
+_to_log journalctl
 
 cp /tmp/*.log ${OUTDIR}
+cp /root/lorax-packages.log ${OUTDIR}
 
 if [[ -e /tmp/pre-anaconda-logs/ ]];then
     cp -r /tmp/pre-anaconda-logs ${OUTDIR}
--- a/scripts/log-capture
+++ b/scripts/log-capture
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-date=$(date +%F_%R)
+date=$(date +%F_%H%M%S)
 OUTDIR=/tmp/log-capture-${date}
 ARCHIVE=${1-/tmp/log-capture.tar.bz2}
 


jkonecny12:
Note:

This script is raising traceback thanks to the error from pyparted:
https://bugzilla.redhat.com/show_bug.cgi?id=1801355

scripts/log-capture
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-date=$(date +%F_%R)
+date=$(date +%F_%H%M%S)
```
poncovka: `%R` represents only `%H:%M`.

poncovka:
> **Question:**
> Do we want to add this to `/usr/bin/`. Right now users have to type `/usr/libexec/anaconda/log-capture` to invoke this script. Installing this as a `anaconda-log-capture` could be easier to discover for users.

I don't think that we should have scripts in `/usr/bin/`.

> **Further improvements:**
> We may extend this to automatically upload the tar file to an existing bugzilla with a BZ number as argument.

I would love to add the new button for reporting problems from UI. It is the most usable solution for everyone, easy to test and easy to maintain. This does not apply to our scripts. I think it is a waste of time to try to extend them.The scripts should be as simple as possible.

scripts/log-capture
jkonecny12: I know it does but I think seconds should be included too. You can have two reports under one minute when it describes just one click difference.

jkonecny12:
> > **Question:**
> > Do we want to add this to `/usr/bin/`. Right now users have to type `/usr/libexec/anaconda/log-capture` to invoke this script. Installing this as a `anaconda-log-capture` could be easier to discover for users.
> 
> I don't think that we should have scripts in `/usr/bin/`.

We have at least `liveinst` and `anaconda`. From the spec file we also have:

```
%files core -f %{name}.lang
%{_bindir}/instperf
%{_bindir}/anaconda-disable-nm-ibft-plugin
%{_sbindir}/anaconda
%{_sbindir}/handle-sshpw
%{_bindir}/analog
%{_bindir}/anaconda-cleanup

%files live
%{_bindir}/liveinst
%{_sbindir}/liveinst
```

> 
> > **Further improvements:**
> > We may extend this to automatically upload the tar file to an existing bugzilla with a BZ number as argument.
> 
> I would love to add the new button for reporting problems from UI. It is the most usable solution for everyone, easy to test and easy to maintain. This does not apply to our scripts. I think it is a waste of time to try to extend them.The scripts should be as simple as possible.

I don't think it's a correct way to think about it. Text ui reporting is valuable the same way as GUI and for that script is great way to do that.
Also it's much easier to have one script to grab all the things on one place than solving this for each UI. Even thought it could be python-meh here.

PR #2318, opened 2020-02-16
Don't modify the list of device names

poncovka:
When we are renaming a container, we shouldn't modify the list of device names.
This list is created on demand in a property now, so the changes have no effect.


poncovka:
Jenkins, test this please.

poncovka:
Jenkins, test this please.

poncovka:
Jenkins, test this please.

PR #2319, opened 2020-02-17
Fix installation source switching to/from CDN

M4rtinK:
Fix various issues with switching between the Red Hat CDN
and a different source, such a on-media repos on a DVD image
or a user specified repository URL.

Also make sure repository priorities work as expected:
- if user provides repository URL via boot options or
  kickstart, it is always used, even on the boot ISO,
  where otherwise CDN is the default installation source
- on the DVD image the CDN should be automatically enabled
  once the user registers, as the on media content is generally
  outdated, while CDN content should be always up to date
- for a kickstart installation without a valid rhsm command
  from a DVD image the DVD image repository should be used
  as installation source without the need for specifying
  the "cdrom" command in kickstart

And lastly make sure Anaconda correctly falls back to the most
suitable installation source if CDN is set as installation source,
registers and then unregisters:
- if user provides repository URL via boot options or
  kickstart, then selects CDN, registers and then
  for some reason unregisters -> fall back to the URL source
- is user boots a DVD image, registers (this auto-activates CDN)
  and then unregisters -> fall back to local DVD repository

Resolves: rhbz#1791376
Resolves: rhbz#1790383
Resolves: rhbz#1788487

### DIFF
--- a/pyanaconda/payload/__init__.py
+++ b/pyanaconda/payload/__init__.py
@@ -43,7 +43,7 @@ from pyanaconda.core.constants import DRACUT_ISODIR, DRACUT_REPODIR, DD_ALL, DD_
     DD_RPMS, INSTALL_TREE, ISO_DIR, THREAD_STORAGE, THREAD_PAYLOAD, THREAD_PAYLOAD_RESTART, \
     THREAD_WAIT_FOR_CONNECTING_NM, THREAD_SUBSCRIPTION, PayloadRequirementType, \
     GRAPHICAL_TARGET, TEXT_ONLY_TARGET
-from pyanaconda.modules.common.constants.services import SERVICES
+from pyanaconda.modules.common.constants.services import SERVICES, SUBSCRIPTION, PAYLOAD
 from pykickstart.constants import GROUP_ALL, GROUP_DEFAULT, GROUP_REQUIRED
 from pyanaconda.flags import flags
 from pyanaconda.core.i18n import _, N_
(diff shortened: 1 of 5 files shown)


VladimirSlavik:
I don't really know what I'm doing, but... why is url special?

pyanaconda/payload/__init__.py
```diff
@@ -1420,6 +1420,19 @@ def _setup_cdrom_device(self, storage, method, isodev, device):
                     method.server = host
                     method.dir = path
                 else:
+                    if flags.automatedInstall:
+                        # During a kickstart installation if we auto-detected
+                        # a local media, we should also disable the CDN in case
+                        # it was set previously and but not enough authetication
```
rvykydal: redundant "and", authetication -> authentication ?

pyanaconda/payload/__init__.py
```diff
@@ -1420,6 +1420,19 @@ def _setup_cdrom_device(self, storage, method, isodev, device):
...
+                        # data is available.
+                        subscription_proxy = SUBSCRIPTION.get_proxy()
+                        org_set = bool(subscription_proxy.Organization)
+                        key_set = subscription_proxy.IsActivationKeySet
+                        kickstart_sufficient = org_set and key_set
+                        if not kickstart_sufficient:
+                            log.debug("payload: making sure cdrom is used instead of CDN")
+                            payload_proxy = PAYLOAD.get_proxy()
```
rvykydal: payload_proxy is not used

pyanaconda/ui/gui/spokes/installation_source.py
```diff
@@ -623,7 +649,16 @@ def changed(self):
 
     @property
     def cdn_used_as_installation_source(self):
-        return self._cdnButton.get_active()
+
```
rvykydal: Maybe this empty line shouldn't be here for consistency.

M4rtinK:
@VladimirSlavik There is a subtle difference - we want to always override the automatically detected cdrom source with CDN, as the CDN will basically always have more up to date content. 

We don't want to ever override the URL source, as that is always set by the user via boot options or kickstart.

But in both cases we want to revert back (to cdrom/URL) if the user decides to unregister.

It's definitely a bit complicated & the way installation source switching has been designed in the past makes it rather hard to implement correctly.

VladimirSlavik:
Ah, now I remember you said something about priority of sources earlier. So that's it.

pyanaconda/ui/gui/spokes/installation_source.py
M4rtinK: Fixed. :)

pyanaconda/payload/__init__.py
M4rtinK: Oops, there should be one more line using the proxy - good catch! :) Fixed.

pyanaconda/payload/__init__.py
M4rtinK: Indeed! Fixed. :)

PR #2324, opened 2020-02-19
Migrate localization to weblate

jkonecny12:
Switch our workflow from Zanata to Weblate.

The main changes are in the Makefile.

We have disabled automatic pot file push during the release because we want to be cautious for now. In the future we can create automation to push pot file after each merge.

**Warning:**
Some parts of this PR is hard to test. So makebumpver.py is not really tested and it should be. Could you please @M4rtinK look on that if the script is working correctly after this change?


Makefile.am
```diff
@@ -99,7 +99,33 @@ po-pull:
 po-push:
 	$(MAKE) -C po $(PACKAGE_NAME).pot-update
 	rm $(srcdir)/po/{main,extra}.pot
-	$(ZANATA_CLIENT_BIN) push $(ZANATA_PUSH_ARGS)
+# This algorithm will make these steps:
+# - clone localization repository
+# - copy pot file to this repository
+# - check if pot file is changed (ignore the POT-Creation-Date otherwise it's always changed)
+# - it does not changed:
```
M4rtinK: I think this should be:
"if not changed"

Makefile.am
```diff
@@ -99,7 +99,33 @@ po-pull:
...
+#   - remove cloned repository
+# - it does changed:
```
M4rtinK: "if changed"

Makefile.am
```diff
@@ -99,7 +99,33 @@ po-pull:
...
+#   - add pot file
+#   - commit pot file
+#   - tell user to verify this file and push it in the temp dir
```
M4rtinK: "push it to the remote from the temp dir" ?

Makefile.am
```diff
@@ -180,10 +175,8 @@ bumpver: po-pull
 	if [ ! -z "$(SKIP_ACKS)" ]; then \
 		opts="$${opts} -s" ; \
 	fi ; \
-	( cd $(srcdir) && scripts/makebumpver $${opts} ) || exit 1 ; \
-	$(MAKE) -C po $(PACKAGE_NAME).pot-update && \
-	rm $(srcdir)/po/{main,extra}.pot
-	$(ZANATA_CLIENT_BIN) push $(ZANATA_PUSH_ARGS)
+	( cd $(srcdir) && scripts/makebumpver $${opts} ) || exit 1 ;
```
rvykydal: I wonder if there should bre '\' at the and of the line. Perhaps not necessary, but there are places we do that, so I better ask.

Makefile.am
jkonecny12: Fixed.

Makefile.am
jkonecny12: Fixed.

Makefile.am
jkonecny12: Sounds much better. Fixed.

Makefile.am
jkonecny12: It's correct. We don't share anything from the above in the last line so there is no reason to have it as one shell.
However, I removed have `;` on the last line. :)

jkonecny12:
**UPDATED:**
* Fix grammar issues
* Fix builds (removing *.txt from a make dist)
* Remove unneeded semicolon from Makefile

jkonecny12:
This PR was closed by accident and can't be reopened...

PR #2326, opened 2020-02-20
Remove dead dracut code

jkonecny12:
**Clean-up time in Dracut!!!** :fireworks: :fireworks: 

Dracut has code to copy stage2 image to RAM. However, this is not used anymore (from 2012).

This code was added by:
99854685688efd9ee5d582953c790e264919dcc8

And disabled / not completely removed by:
adfcd4ba7bae513d4f51bf36ab37bc6fc0bdc28e

This code was used only for Anaconda upgrades and Anaconda does not handle upgrades long time. Git rid of it.


jkonecny12:
I've tested this and everything looks correctly.

PR #2328, opened 2020-02-21
Fix the default disk selection (#1805553)

poncovka:
If there are some disks already selected, do nothing. In the automatic
installation, select all disks. In the interactive installation, select
a disk if there is only one available.

Convert bytes to a readable string representation of the size.


rvykydal:
I think with this PR merged we can consider the https://github.com/rhinstaller/anaconda/pull/2298 as ported to Fedora.

PR #2330, opened 2020-02-21
Fix weblate migration issues

jkonecny12:
Fix issues from weblate migration process.


jkonecny12:
This is merged but there will be follow-up because the translation repository will change it's structure.

PR #2334, opened 2020-02-24
Wait for entropy in the DBus Storage module

poncovka:
The installation task ActivateFilesystemsTask should report Blivet's messages.

The task ActivateFilesystemsTask should report that the entropy is lower than
requested and wait for a limited time to increase it.

Connect to the ProgressChanged signal of the DBus tasks and show their progress
messages. Always drop the first message, because it is the same as the name of
the DBus task, so it was probably already shown.

If a DBus installation task raises an error, call the error handler to
show a correct dialog.

### DIFF
--- a/pyanaconda/modules/storage/installation.py
+++ b/pyanaconda/modules/storage/installation.py
@@ -17,7 +17,10 @@
 # License and may only be used or replicated with the express permission of
 # Red Hat, Inc.
 #
+from blivet import callbacks
+
 from pyanaconda.anaconda_loggers import get_module_logger
+from pyanaconda.core.i18n import _
 from pyanaconda.core.configuration.anaconda import conf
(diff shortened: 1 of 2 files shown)


poncovka:
Updated.

pyanaconda/modules/storage/installation.py
```diff
@@ -68,6 +77,56 @@ def _report_message(self, data):
...
+            return
+
+        if time == 0:
+            self.report_progress(_("Gathering entropy (time ran out)"))
+            return
+
+        message = _("Gathering entropy {percents}% (remaining time {time})").format(
+            percents=percents,
+            time=timedelta(seconds=time)
+        )
```
jkonecny12: Please change this to `if-elif-else` with `self.report_progress` on one place with `message` variable.

PR #2336, opened 2020-02-24
Don't remove unknown devices (#1806233)

poncovka:
Don't try to remove a device that is not in the device tree. It could have been
already removed in one of the previous iterations.

Get the format data of the specified device, so we can check if its
format type is protected before a removal of this device.

Resolves: rhbz#1806233


poncovka:
Updated.

PR #2339, opened 2020-02-25
Handle invalid optical install media (#1806520)

poncovka:
The function find_optical_install_media should ignore optical media that cannot
be mounted. Handle the DBus error MountFilesystemError, instead of the Blivet's
exception.

Resolves: rhbz#1806520


VladimirSlavik:
Jenkins, please test this.

PR #2340, opened 2020-02-25
Port prefixdevname to fedora

rvykydal:


pyanaconda/modules/network/installation.py
```diff
@@ -112,7 +114,8 @@ class NetworkInstallationTask(Task):
 """.strip()
 
     def __init__(self, sysroot, disable_ipv6, overwrite,
-                 network_ifaces, ifname_option_values):
+                 network_ifaces, ifname_option_values,
+                 configure_prefixdevname):
```
jkonecny12: I don't really agree with the variable name. I don't think we should use the same name in dracut which is inconsistent with the rest of the code here.

pyanaconda/modules/network/installation.py
```diff
@@ -254,6 +262,19 @@ def _copy_dhclient_config_files(self, root, network_ifaces):
             dhclient_file = self.DHCLIENT_FILE_TEMPLATE.format(device_name)
             self._copy_file_to_root(root, dhclient_file)
 
+    def _copy_prefixdevname_files(self, root):
+        """Copy prefixdevname persistent conifuration to target system.
```
jkonecny12:
```suggestion
        """Copy prefixdevname persistent configuration to target system.
```

tests/nosetests/pyanaconda_tests/module_network_test.py
```diff
@@ -1158,3 +1160,16 @@ def apply_boot_options_test(self):
...
+        mocked_kernel_args = {'ifname': "bla bla"}
+        self.assertEqual(
+            self.network_module._is_using_persistent_device_names(mocked_kernel_args),
+            False
+        )
+        mocked_kernel_args = {'net.ifnames.prefix': "net"}
+        self.assertEqual(
+            self.network_module._is_using_persistent_device_names(mocked_kernel_args),
+            True
+        )
```
jkonecny12: I don't think this test is what should be here. Point is to test the use case, that means that we are trying to achieve something by the call of something. This test is testing only our protected function which is only a part in the whole solution and should be hidden as uninteresting part of the whole test. 
In general testing of private API is not what should be tested.

pyanaconda/modules/network/installation.py
rvykydal: Renamed

pyanaconda/modules/network/installation.py
rvykydal: Fixed

tests/nosetests/pyanaconda_tests/module_network_test.py
rvykydal: Removed. After merging the branch I'll open a PR for update of installation task unit tests that should cover also the private method.

PR #2342, opened 2020-02-26
Adapt po makefile L10N commands

jkonecny12:
Adapt to a new directory structure of our localization repository.
Remove git-find-branch feature, we should focus on having specific stuff for branches working properly.

This have to be ported to `f32-release` branch with a small change in `GIT_BRANCH` constant value.


M4rtinK:
And as for `f32-release` - I think the best way is to merge this PR to `f32-devel`, then merge it to `f32-release`, add a commit changing `GIT_BRANCH ?= master` to `GIT_BRANCH ?= f32-release`, then push `f32-release` to origin.

Does that sounds correct ?

jkonecny12:
It is correct to avoid conflicts from `f32-devel` but it's nonsense from a PoV that you are giving master specific configuration to `f32-devel` branch. :(
We really have to solve this other way.

jkonecny12:
Closing this and I'll create a new f32-devel PR.

PR #2343, opened 2020-02-26
Adapt localization makefile commands

jkonecny12:
Adapt to a new directory structure of our localization repository.
Remove git-find-branch feature, we should focus on having specific stuff for branches working properly.

This have to be ported to f32-release branch with a small change in GIT_BRANCH constant value.

Replacement of https://github.com/rhinstaller/anaconda/pull/2342 for `f32-devel`.


jkonecny12:
This was already added to Fedora by this PR.

PR #2350, opened 2020-03-03
Handle storage errors in UI (#1808650)

poncovka:
Update the error handlers of the storage-related exceptions.
Replace exceptions with DBus errors and make sure that these
errors will be raised when it is expected.

Remove the error handler from the DBus Storage module.

Resolves: rhbz#1808650

### DIFF
--- a/pyanaconda/bootloader/zipl.py
+++ b/pyanaconda/bootloader/zipl.py
@@ -21,18 +21,12 @@ import re
 from pyanaconda.bootloader.base import BootLoader, Arguments, BootLoaderError
 from pyanaconda.core import util
 from pyanaconda.core.configuration.anaconda import conf
-from pyanaconda.errors import errorHandler
 from pyanaconda.product import productName
 
 from pyanaconda.anaconda_loggers import get_module_logger
 log = get_module_logger(__name__)
 
 __all__ = ["ZIPL"]
-
-
-class ZIPLError(Exception):
-    """An exception class for ZIPL errors."""
-    pass
 
 
 class ZIPL(BootLoader):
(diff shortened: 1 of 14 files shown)


pyanaconda/ui/gui/spokes/lib/detailederror.glade
```diff
@@ -24,6 +24,7 @@
             <property name="xalign">0</property>
             <property name="label" translatable="yes">An unknown error occurred during installation.  Details are below.</property>
             <property name="wrap">True</property>
+            <property name="max_width_chars">60</property>
```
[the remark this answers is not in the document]
poncovka: Yes, it is. When I was testing the exception UnusableStorageError, the dialog was really ugly.

pyanaconda/bootloader/zipl.py
```diff
@@ -162,7 +156,7 @@ def install(self, args=None):
             # exceed 896 bytes; there is nothing we can do about this, so just
             # catch the error and show it to the user instead of crashing
             elif line.startswith("Error: The length of the parameters "):
-                errorHandler.cb(ZIPLError(line))
+                raise BootLoaderError(line)
```
jkonecny12: Did you looked on the possible output. The problem I'm seeing here is that it could be hard to find out this is a ZIPL problem. Raising bootloader error with output of zipl command could be pretty confusing.

poncovka: The removed error handler `_ziplErrorHandler` doesn't say anything about zipl neither, so the user will see pretty much the same message as before. The example of the content of `line` is below. I'd rather not change it.

`Error: The length of parameters line (1185 bytes) exceeds the allowed maximum (896 bytes)`.

jkonecny12: Agree. Thanks for explanation.

PR #2351, opened 2020-03-03
Add small badges for Read the Docs and translations

poncovka:
Remove the bigger badge for translations and add two smaller ones.


poncovka:
@VladimirSlavik, the github badges are usually placed at the top of the README file in one row.

PR #2353, opened 2020-03-03
Verify that LVM destruction is orderly (#1688478)

VladimirSlavik:
Fixes the infamous LVM bug somewhat: There is now a proper error message.

Resolves: rhbz#1688478
Resolves: rhbz#1584729
Resolves: rhbz#1698570
...and more than we ever wanted to know :-/

### DIFF
--- a/pyanaconda/storage/checker.py
+++ b/pyanaconda/storage/checker.py
@@ -19,6 +19,8 @@ import gi
 gi.require_version("BlockDev", "2.0")
 from gi.repository import BlockDev as blockdev
 
+from collections import defaultdict
+
 from blivet import arch, util
 from blivet.size import Size
 
@@ -411,6 +413,43 @@ def verify_mounted_partitions(storage, constraints, report_error, report_warning
                                "installation. Please unmount it and retry.") % part.path)
 
 
+def verify_lvm_destruction(storage, constraints, report_error, report_warning):
+    """Verify that destruction of LVM devices is correct.
+
+    A VG and all its PVs must be all destroyed together, nothing is allowed to remain.
+
+    :param storage: a storage to check
+    :param constraints: a dictionary of constraints
+    :param report_error: a function for error reporting
+    :param report_warning: a function for warning reporting
+    """
+    # Implementation detects VGs that should be destroyed and aren't. This is because of how
+    # blivet implements ignoring devices: Ignoring a device hides also all dependent devices - for
+    # any disk, it could be a PV partition, which would then hide also all VGs and LVs that depend
+    # on the PV. It does not matter that the VG needs other PVs too - the hiding wins.
+    destroyed_vg_names = set()
+    all_touched_disks_by_vg = defaultdict(list)
+
+    for action in storage.devicetree.actions:
+        if action.is_destroy and action.is_device and action.device.type == "lvmvg":
+            destroyed_vg_names.add(action.device.name)
+
+        elif action.is_destroy and action.is_format and action.orig_format.type == "lvmpv":
+            # Check if the PV actually had any VG assigned
+            if action.orig_format.vg_name:
+                disk_name = action.device.disk.name
+                vg_name = action.orig_format.vg_name
+                all_touched_disks_by_vg[vg_name].append(disk_name)
+
+    for vg_name, disks in all_touched_disks_by_vg.items():
+        if vg_name not in destroyed_vg_names:
+            report_error(_(
+                "Selected disks {} contain volume group '{}' that also uses further unselected "
+                "disks. You must select or de-select all these disks as a set."
+                .format(", ".join(disks), vg_name)
+            ))
+
+
 class StorageCheckerReport(object):
     """Class for results of the storage checking."""
 
@@ -620,6 +659,7 @@ class StorageChecker(object):
         self.add_check(verify_luks_devices_have_key)
         self.add_check(verify_luks2_memory_requirements)
         self.add_check(verify_mounted_partitions)
+        self.add_check(verify_lvm_destruction)
 
 
 # Setup the storage checker.
--- a/tests/nosetests/pyanaconda_tests/module_storage_checker_test.py
+++ b/tests/nosetests/pyanaconda_tests/module_storage_checker_test.py
@@ -18,13 +18,13 @@
 # Red Hat Author(s): Vendula Poncova <vponcova@redhat.com>
 #
 import unittest
-from unittest.mock import patch
+from unittest.mock import patch, Mock
 
 from dasbus.typing import get_variant, Int
 from pyanaconda.core.constants import STORAGE_MIN_RAM
 from pyanaconda.modules.storage.checker import StorageCheckerModule
 from pyanaconda.modules.storage.checker.checker_interface import StorageCheckerInterface
-from pyanaconda.storage.checker import storage_checker
+from pyanaconda.storage.checker import storage_checker, verify_lvm_destruction
 
 
 class StorageCheckerInterfaceTestCase(unittest.TestCase):
@@ -39,3 +39,58 @@ class StorageCheckerInterfaceTestCase(unittest.TestCase):
         with patch.dict(storage_checker.constraints):
             self.interface.SetConstraint(STORAGE_MIN_RAM, get_variant(Int, 987))
             self.assertEqual(storage_checker.constraints[STORAGE_MIN_RAM], 987)
+
+
+class StorageCheckerVerificationTestCase(unittest.TestCase):
+
+    @patch("pyanaconda.storage.checker._", side_effect=lambda x: x)
+    def lvm_verification_test(self, _):
+        """Test the LVM destruction test."""
+        # VG that is destroyed correctly
+        action1 = Mock()
+        action1.is_destroy = True
+        action1.is_device = True
+        action1.device.type = "lvmvg"
+        action1.device.name = "VOLGROUP-A"
+
+        # something to ignore
+        action2 = Mock()
+        action2.is_destroy = False
+
+        # PV that belongs to VG from #1
+        action3 = Mock()
+        action3.is_destroy = True
+        action3.is_format = True
+        action3.orig_format.type = "lvmpv"
+        action3.device.disk.name = "PHYSDISK-1"
+        action3.orig_format.vg_name = "VOLGROUP-A"
+
+        # PV that belongs to a missing group
+        action4 = Mock()
+        action4.is_destroy = True
+        action4.is_format = True
+        action4.orig_format.type = "lvmpv"
+        action4.device.disk.name = "PHYSDISK-2"
+        action4.orig_format.vg_name = "VOLGROUP-B"
+
+        # PV that does not belong to any group
+        action5 = Mock()
+        action5.is_destroy = True
+        action5.is_format = True
+        action5.orig_format.type = "lvmpv"
+        action5.device.disk.name = "PHYSDISK-3"
+        action5.orig_format.vg_name = ""
+
+        storage = Mock()
+        storage.devicetree.actions = [action1, action2, action3, action4, action5]
+        error_handler = Mock()
+        warning_handler = Mock()
+
+        verify_lvm_destruction(storage, None, error_handler, warning_handler)
+
+        warning_handler.assert_not_called()
+        error_handler.assert_called_once_with(
+            "Selected disks {} contain volume group '{}' that also uses further unselected disks. "
+            "You must select or de-select all these disks as a set."
+            .format("PHYSDISK-2", "VOLGROUP-B")
+        )


pyanaconda/storage/checker.py
```diff
@@ -411,6 +412,51 @@ def verify_mounted_partitions(storage, constraints, report_error, report_warning
...
+    log.debug("called verify_lvm_destruction()")
+    destroyed_vg_names = set()
+    vg_names_seen_in_destroyed_pvs = set()
+    for action in storage.devicetree.actions:
+        if isinstance(action, ActionDestroyDevice) and action.device.type == "lvmvg":
+            log.debug("ActionDestroyDevice: %s %s", action.device.type, action.device.name)
+            destroyed_vg_names.add(action.device.name)
+            if not action.device.complete:
+                report_warning("Disk selection will result in removing an incomplete volume "
+                               "group '{}'.".format(action.device.name))
```
VladimirSlavik: This part is up for discussion - do we want to warn when we'll destroy an incomplete VG? Apparently we can currently just steamroll over that, but it still feels not exactly right.

VladimirSlavik:
Probably also TODO: write test for this.

pyanaconda/storage/checker.py
poncovka: @vojtechtrefny, what do you think?

pyanaconda/storage/checker.py
```diff
@@ -411,6 +412,51 @@ def verify_mounted_partitions(storage, constraints, report_error, report_warning
...
+        elif isinstance(action, ActionDestroyFormat) and action.orig_format.type == "lvmpv":
+            log.debug("ActionDestroyFormat: %s %s", action.orig_format.type, action.device.name)
```
poncovka: Please, remove the debugging logs. We always log the result of the validation, that should be enough.

pyanaconda/storage/checker.py
```diff
@@ -411,6 +412,51 @@ def verify_mounted_partitions(storage, constraints, report_error, report_warning
...
+            if action.orig_format.vg_name:
+                vg_names_seen_in_destroyed_pvs.add(action.orig_format.vg_name)
+    missing_vg_names = vg_names_seen_in_destroyed_pvs - destroyed_vg_names
+    log.debug("destroyed VGs: %s", destroyed_vg_names)
+    log.debug("seen VGs: %s", vg_names_seen_in_destroyed_pvs)
+    log.debug("missing VGs: %s", missing_vg_names)
+    if missing_vg_names:
+        for vg_name in missing_vg_names:
+            report_error(_("Disk selection contains only some physical volumes for volume group "
+                           "'{}'. Adjust your 'ignoredisk' command, or remove the physical "
```
poncovka: This validation runs for all types of partitioning, so these messages should be generic. I would recommend to ask users to adjust their disk selection, or remove the physical volumes. In some cases, it will be possible to solve this validation error by changing the disk selection in the storage spoke.

VladimirSlavik: First - by generic, you mean they should not mention kickstart, right?

Second - Does it make sense to see the LVM details here, or do you think walking the changes to speak only about disks is better? The error _is_ caused specifically by LVM stuff, but if I changed it to list disks, I could reuse the existing error message from GUI.

poncovka: > First - by generic, you mean they should not mention kickstart, right?

Yes, they should not mention kickstart. I would replace only `Adjust your 'ignoredisk' command` with `Adjust your disk selection`. Otherwise, the message looks good to me.

> Second - Does it make sense to see the LVM details here, or do you think walking the changes to speak only about disks is better? The error _is_ caused specifically by LVM stuff, but if I changed it to list disks, I could reuse the existing error message from GUI.

 It is definitely good to mention the cause of the error and provide some suggestions how to fix it or avoid it, so the LVM details should be part of the message.

pyanaconda/storage/checker.py
```diff
@@ -411,6 +412,51 @@ def verify_mounted_partitions(storage, constraints, report_error, report_warning
                                "installation. Please unmount it and retry.") % part.path)
 
 
+def verify_lvm_destruction(storage, constraints, report_error, report_warning):
+    """ Verify that destruction of LVM devices is correct.
```
jkonecny12:
```suggestion
    """Verify that destruction of LVM devices is correct.
```

pyanaconda/storage/checker.py
```diff
@@ -411,6 +412,51 @@ def verify_mounted_partitions(storage, constraints, report_error, report_warning
...
+
+    This means: If a PV is destroyed, the VG on it must be destroyed too.
+    Additionally, destroying an incomplete VG warns about this.
```
jkonecny12: Maybe I'm misunderstanding something but this seems incorrect to me.

In general you should be able to destroy PV and just make VG smaller. I'm not sure if we support this in Anaconda, thought. The main problem we are trying to solve is that you have to have a complete VG (with all PVs) to be able to work with it. Am I missing something? Is this some kind of simplification?

pyanaconda/storage/checker.py
jkonecny12: I don't think this message is correct. I would rather see something like:
```Disk selection does not contain all parts of existing volume groups```

File: pyanaconda/storage/checker.py [hunk 6]
```diff
@@ -411,6 +412,51 @@ def verify_mounted_partitions(storage, constraints, report_error, report_warning
...
+            destroyed_vg_names.add(action.device.name)
+            if not action.device.complete:
+                report_warning("Disk selection will result in removing an incomplete volume "
+                               "group '{}'.".format(action.device.name))
+        elif isinstance(action, ActionDestroyFormat) and action.orig_format.type == "lvmpv":
+            log.debug("ActionDestroyFormat: %s %s", action.orig_format.type, action.device.name)
+            # Must check if the PV actually has a VG assigned
+            if action.orig_format.vg_name:
+                vg_names_seen_in_destroyed_pvs.add(action.orig_format.vg_name)
+    missing_vg_names = vg_names_seen_in_destroyed_pvs - destroyed_vg_names
```
jkonecny12: I would split this code to two private methods. Collect and check

poncovka: I don't think it is necessary. One the debugging messages are gone, the function will be small and readable enough.

pyanaconda/storage/checker.py
```diff
@@ -411,6 +412,51 @@ def verify_mounted_partitions(storage, constraints, report_error, report_warning
...
+    # that has a VG that isn't going to be destroyed, that is an error. This is because we ask
+    # blivet to ignore some disks, and these might contain a PV. Crucially, blivet implements this
+    # such that the ignoring hides also all dependent devices - for any disk, it could be a PV
+    # partition, which would then hide also all VGs and LVs that depend on the PV. It does not
+    # matter that the VG needs other PVs too - the hiding wins.
+    log.debug("called verify_lvm_destruction()")
+    destroyed_vg_names = set()
+    vg_names_seen_in_destroyed_pvs = set()
+    for action in storage.devicetree.actions:
+        if isinstance(action, ActionDestroyDevice) and action.device.type == "lvmvg":
```
vojtechtrefny: You can use `action.is_destroy` here.

pyanaconda/storage/checker.py
vojtechtrefny: I don't think it's necessary to warn about this.

VladimirSlavik: Removing then.

pyanaconda/storage/checker.py
VladimirSlavik: Done.

pyanaconda/storage/checker.py
VladimirSlavik: Done.

pyanaconda/storage/checker.py
VladimirSlavik: Ah, thanks for the tip - actually I think `action.is_destroy and action.is_device` would be better?

pyanaconda/storage/checker.py
VladimirSlavik: Done.

pyanaconda/storage/checker.py
VladimirSlavik: Yes, this is a simplification, at least compared to what you are thinking about. This does not check if there is no possible sequence of actions that could satisfy the high-level objectives. Instead, this checks if we can do what we do - that the selected disks can be just wiped. So the check is, will we leave a known dangling PV behind?

The comment was half-truth, so I changed it anyway. Now it says: A VG and all its PVs must be all destroyed together, nothing is allowed to remain.

As an aside, the action you are hinting at is not very useful, IMO. Effectively, you would have to successfully downsize the filesystems on all the LVs so as to fit on the remaining PVs. Is that possible? Yes. But IMO most users don't want a magical shrinking of their previous installation, but rather wipe everything and start anew.

pyanaconda/storage/checker.py
VladimirSlavik: Yes, that works, and is_format for the other one. Changed.

VladimirSlavik:
Update: Changed logic somewhat so that the disks can be reported.

@vojtechtrefny , it seems `get_related_disks` won't work in this situation - I am getting back *all* disks. (Would have expected none.) Is this a final verdict?

pyanaconda/storage/checker.py
```diff
@@ -411,6 +413,60 @@ def verify_mounted_partitions(storage, constraints, report_error, report_warning
...
+                vg_name = action.orig_format.vg_name
+                all_touched_vgs_and_disks[vg_name].append(disk_name)
+
+    log.debug("verify_lvm_destruction : destroyed_vg_names, all_touched_vgs_and_disks :")
+    log.debug(destroyed_vg_names)
+    log.debug(all_touched_vgs_and_disks)
+    for vg_name in set(all_touched_vgs_and_disks):
+        if vg_name not in destroyed_vg_names:
+            selected_disks = all_touched_vgs_and_disks[vg_name]
+            helper_disk = storage.devicetree.get_device_by_name(selected_disks[0])
```
poncovka: Why don't you keep the disks instead of their names?

pyanaconda/storage/checker.py
```diff
@@ -411,6 +413,60 @@ def verify_mounted_partitions(storage, constraints, report_error, report_warning
...
+            related_disks = storage.devicetree.get_related_disks(helper_disk) # THIS IS DOING WRONG THINGS
```
poncovka: I don't think this is going to work, because the destroyed LVM devices that define the relations between disks are not in the device tree at this moment.

pyanaconda/storage/checker.py
```diff
@@ -411,6 +413,60 @@ def verify_mounted_partitions(storage, constraints, report_error, report_warning
...
+        if action.is_destroy and action.is_device and action.device.type == "lvmvg":
+            destroyed_vg_names.add(action.device.name)
+
+        elif action.is_destroy and action.is_format and action.orig_format.type == "lvmpv":
+            # Check if the PV actually had any VG assigned
+            if action.orig_format.vg_name:
+                pv_name = action.device.name
+                disk_name = action.device.disk.name
+                vg_name = action.orig_format.vg_name
+                all_touched_vgs_and_disks[vg_name].append(disk_name)
```
poncovka: This implementation of the verification check is a little harder to follow. I liked the previous one.

pyanaconda/storage/checker.py
VladimirSlavik: Yep, Vojta confirms that the function is not going to work once actions happen. The untouched disks are linked via the to-be-destroyed LVM bits, while the touched disks via the to-be-created bits. That's impossible to reconcile, so in the end it spews nonsense.

pyanaconda/storage/checker.py
VladimirSlavik: Mainly it's due to trying to find what *all* the involved disks are/were. I will remove that and make it more simple.

pyanaconda/storage/checker.py
VladimirSlavik: Sorry, this was just a prototype of sorts. I tried both and there wasn't as much difference. Anyway I'll change this, see other comments below.

pyanaconda/storage/checker.py
```diff
@@ -411,6 +413,46 @@ def verify_mounted_partitions(storage, constraints, report_error, report_warning
...
+            destroyed_vg_names.add(action.device.name)
+
+        elif action.is_destroy and action.is_format and action.orig_format.type == "lvmpv":
+            # Check if the PV actually had any VG assigned
+            if action.orig_format.vg_name:
+                disk_name = action.device.disk.name
+                vg_name = action.orig_format.vg_name
+                all_touched_disks_by_vg[vg_name].append(disk_name)
+
+    for vg_name in set(all_touched_disks_by_vg):
```
poncovka: It is not necessary to create a set of keys. You can iterate over keys with `for vg_name in all_touched_disks_by_vg`. However, I would iterate over keys and items with `for vg_name, disks in all_touched_disks_by_vg.items()`. Then you can replace `all_touched_disks_by_vg[vg_name]` with `disks` in the formatting below.

pyanaconda/storage/checker.py
```diff
@@ -411,6 +413,46 @@ def verify_mounted_partitions(storage, constraints, report_error, report_warning
...
+        if vg_name not in destroyed_vg_names:
+            report_error(_(
+                "Selected disks {} contain volume group {} that also uses further unselected disks. "
+                "You must select or de-select these disks as a set."
```
poncovka: Maybe `all these disks`?

VladimirSlavik: Good idea - will do.

pyanaconda/storage/checker.py
VladimirSlavik: You're right - sorry, will fix this.

pyanaconda/storage/checker.py
VladimirSlavik: I changed the comment to match the final logic.

VladimirSlavik:
Update:
- All comments above resolved, hopefully
- Added a test

Up for discussion:
- Test location
- Should the whole checking move under modules?

pyanaconda/storage/checker.py
```diff
@@ -411,6 +413,43 @@ def verify_mounted_partitions(storage, constraints, report_error, report_warning
                                "installation. Please unmount it and retry.") % part.path)
 
 
+def verify_lvm_destruction(storage, constraints, report_error, report_warning):
+    """Verify that destruction of LVM devices is correct.
+
+    This means: A VG and all its PVs must be all destroyed together, nothing is allowed to remain.
```
jkonecny12: It's superfluous: 
```suggestion
    A VG and all its PVs must be all destroyed together, nothing is allowed to remain.
```

VladimirSlavik:
Hm, right now it is more of an everything test. It has two actions that are ok, multiple actions that should be ignored, and one that is wrong. The reason it's not multiple tests is that setting up all the storage in detail is not something I'd like to duplicate.

tests/nosetests/pyanaconda_tests/module_storage_checker_test.py
```diff
@@ -39,3 +39,58 @@ def set_constraint_test(self):
...
+        action5.is_destroy = True
+        action5.is_format = True
+        action5.orig_format.type = "lvmpv"
+        action5.device.disk.name = "PHYSDISK-3"
+        action5.orig_format.vg_name = ""
+
+        storage = Mock()
+        storage.devicetree.actions = [action1, action2, action3, action4, action5]
+        error_handler = Mock()
+        warning_handler = Mock()
```
poncovka: I think that you might be able to use Blivet's devices and actions to model this state, but I don't insist on it.

pyanaconda/storage/checker.py
VladimirSlavik: Fixed.

tests/nosetests/pyanaconda_tests/module_storage_checker_test.py
VladimirSlavik: I'm open to suggestions. But would that be better? Mock lets me specify values arbitrarily deep in the "tree", and only these I care about, so this is still super short compared to alternatives...

VladimirSlavik:
Thank you for all the help, again, everyone... Merging.

PR #2355, opened 2020-03-04
Fix action buttons in the resize dialog (#1809950)

poncovka:
If the selected row in the resize dialog is not editable, just make all action
buttons insensitive and quit. The shrink button can be sensitive only for
editable rows with resizable devices.

Never update action buttons for other rows then the selected one. Otherwise,
the resize dialog will be in an invalid state and could crash with an exception.

Resolves: rhbz#1809950

### DIFF
--- a/pyanaconda/ui/gui/spokes/lib/resize.py
+++ b/pyanaconda/ui/gui/spokes/lib/resize.py
@@ -366,7 +366,14 @@ class ResizeDialog(GUIObject):
         self._resize_slider.add_mark(min_size, Gtk.PositionType.BOTTOM, str(Size(min_size)))
         self._resize_slider.add_mark(max_size, Gtk.PositionType.BOTTOM, str(Size(max_size)))
 
-    def _update_action_buttons(self, row):
+    def _update_action_buttons(self):
+        # Update buttons for the selected row.
+        itr = self._selection.get_selected()[1]
+
+        if not itr:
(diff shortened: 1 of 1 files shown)


pyanaconda/ui/gui/spokes/lib/resize.py
```diff
@@ -368,21 +368,20 @@ def _setup_slider(self, min_size, max_size, default_size):
...
-        device_name = obj.name
-        device_data = DeviceData.from_structure(
-            self._device_tree.GetDeviceData(device_name)
-        )
 
-        # Disks themselves may be editable in certain ways, but they are never
-        # shrinkable.
         self._preserve_button.set_sensitive(obj.editable)
-        self._shrink_button.set_sensitive(obj.editable and not device_data.is_disk)
+        self._shrink_button.set_sensitive(obj.editable)
```
jkonecny12: Why we don't have to disable shrink button for disk anymore?

poncovka: It happens a few lines below. See: `self._shrink_button.set_sensitive(is_resizable)`

poncovka:
> IMHO right now the changes looks like they are changing behavior of the dialog but I'm somehow missing explanation why.

@jkonecny12, these are bug fixes, the resize dialog crashes with an exception without them.

PR #2357, opened 2020-03-05
Move functions from pyanaconda.storage.utils

poncovka:
Move functions that are used only by UI to pyanaconda.ui.lib.storage.
Move functions that are used by the Storage module and UI to pyanaconda.core.storage.


poncovka:
Jenkins, test this please.

PR #2358, opened 2020-03-05
Remove unused code

poncovka:
Remove code that is no longer used.


pyanaconda/ui/gui/spokes/lib/entropy_dialog.glade
```diff
@@ -1,85 +0,0 @@
-<?xml version="1.0" encoding="UTF-8"?>
```
M4rtinK: We did move move this functionality somewhere else previously, right ?

M4rtinK:
I think @rvykydal should definitely take a look due the the iscsi & fcoe related parts. Also I think running kickstart tests on the PR could be a good smoke test in case something that was dropped was still in use via some obscure mechanism.

pyanaconda/ui/gui/spokes/lib/entropy_dialog.glade
poncovka: Partially, the logic is the same, but we don't show a dialog, just the progress messages. It is possible, that someone will want the dialog back, but the dialog will have to be rewritten anyway, because the logic is part of the Storage module now. Therefore, I suggest to remove this code for now and bring back the updated version if requested.

M4rtinK: I'm afraid this could be an issue in the future once this change lands in a RHEL release as I can image people expecting the dialog & not noticing this is now reported via a status message.

PR #2360, opened 2020-03-05
Fix the unit tests for BTRFS

poncovka:
The unit tests should be able to run without the btrfs kernel module enabled.


jkonecny12:
Thanks a lot for this!

PR #2365, opened 2020-03-09
Execute a kickstarted partitioning (#1811242)

poncovka:
Skip the execute method of the Storage spoke in GUI only for the interactive
and Blivet-GUI partitioning methods. Kickstarted methods should be always
executed.

Resolves: rhbz#1811242


poncovka:
Jenkins, test this please.

PR #2369, opened 2020-03-09
Avoid downloading payload image in set up phase

jkonecny12:
Before this change Anaconda will download stage2 image in the set up phase. Result was that with big image Payload reset took ages and whole image was downloaded to RAM without any check if the image could be downloaded.

In my case the above resulted in this process: Anaconda started, waited on thread and then OOM killer killed it to free up RAM. Then systemd again started Anaconda and the whole cycle repeated in an never ending cycle.


jkonecny12:
**UPDATED:**
* Backport this fix to modules.

jkonecny12:
Works as expected for module too. Merging.

PR #2370, opened 2020-03-10
Refactorize source and payload factories

poncovka:
Reorganize and extend tests for the source and payload factories
and simplify kickstart processing in the Payloads module.

The changes are consistent with the code of the Storage module
and the partitioning factory.


pyanaconda/modules/payloads/payloads.py
```diff
@@ -97,19 +97,18 @@ def process_kickstart(self, data):
...
+            payload_module = self.create_payload(payload_type)
+            payload_module.process_kickstart(data)
 
-        self._packages.process_kickstart(data)
+            # FIXME: This is a temporary workaround.
+            PayloadContainer.to_object_path(payload_module)
 
-        self.set_payload(payload)
-        PayloadContainer.to_object_path(payload)
+        # FIXME: Process packages in the DNF module.
```
jkonecny12: Add here please that `Packages` module will be converted to a structure as explanation why.

poncovka: Fixed.

poncovka:
Updated. I have renamed only the test file for the Payloads module. We don't really use the service name as a prefix of all related test files. However, we could eventually move these files to new subdirectories.

VladimirSlavik:
I agree we should start thinking about the test file organization - some of the "module test" files looked ridiculously large to me.

PR #2371, opened 2020-03-10
Add initial Subscription DBus module & system purpose data support

M4rtinK:
The first commit creates the initial empty structure of the Subscription DBus module.
The second commit then adds support for handling system purpose data via kickstart and DBus API.

### DIFF
--- a/data/product.d/rhel.conf
+++ b/data/product.d/rhel.conf
@@ -3,6 +3,19 @@
 [Product]
 product_name = Red Hat Enterprise Linux
 
+[Anaconda]
+# List of enabled Anaconda DBus modules for RHEL.
+kickstart_modules =
+     org.fedoraproject.Anaconda.Modules.Timezone
+     org.fedoraproject.Anaconda.Modules.Network
+     org.fedoraproject.Anaconda.Modules.Localization
+     org.fedoraproject.Anaconda.Modules.Security
+     org.fedoraproject.Anaconda.Modules.Users
+     org.fedoraproject.Anaconda.Modules.Payloads
+     org.fedoraproject.Anaconda.Modules.Storage
+     org.fedoraproject.Anaconda.Modules.Services
+     org.fedoraproject.Anaconda.Modules.Subscription
+
 [Installation System]
 # The detection is disabled since #1645686.
 # can_detect_unsupported_hardware = True
--- a/pyanaconda/modules/subscription/Makefile.am
+++ b/pyanaconda/modules/subscription/Makefile.am
@@ -1,6 +1,5 @@
-# modules/Makefile.am for anaconda
 #
-# Copyright (C) 2017  Red Hat, Inc.
+# Copyright (C) 2020  Red Hat, Inc.
 #
 # This program is free software; you can redistribute it and/or modify
 # it under the terms of the GNU Lesser General Public License as published
@@ -15,10 +14,8 @@
 # You should have received a copy of the GNU Lesser General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-SUBDIRS = common boss timezone network localization security users payloads storage services subscription
-
 pkgpyexecdir = $(pyexecdir)/py$(PACKAGE_NAME)
-modulesdir = $(pkgpyexecdir)/modules
-modules_PYTHON = $(srcdir)/*.py
+subscriptiondir = $(pkgpyexecdir)/modules/subscription
+subscription_PYTHON = $(srcdir)/*.py
 
 MAINTAINERCLEANFILES = Makefile.in
--- a/pyanaconda/modules/subscription/__main__.py
+++ b/pyanaconda/modules/subscription/__main__.py
@@ -1,0 +1,6 @@
+from pyanaconda.modules.common import init
+init()
+
+from pyanaconda.modules.subscription.subscription import SubscriptionService
+service = SubscriptionService()
+service.run()
--- a/pyanaconda/modules/subscription/kickstart.py
+++ b/pyanaconda/modules/subscription/kickstart.py
@@ -1,0 +1,28 @@
+#
+# Kickstart handler for the subscription module.
+#
+# Copyright (C) 2020 Red Hat, Inc.
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# the GNU General Public License v.2, or (at your option) any later version.
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY expressed or implied, including the implied warranties of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General
+# Public License for more details.  You should have received a copy of the
+# GNU General Public License along with this program; if not, write to the
+# Free Software Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
+# 02110-1301, USA.  Any Red Hat trademarks that are incorporated in the
+# source code or documentation are not subject to the GNU General Public
+# License and may only be used or replicated with the express permission of
+# Red Hat, Inc.
+#
+from pyanaconda.core.kickstart import VERSION, KickstartSpecification, commands as COMMANDS
+
+
+class SubscriptionKickstartSpecification(KickstartSpecification):
+
+    commands = {
+        "syspurpose": COMMANDS.Syspurpose,
+        "rhsm": COMMANDS.RHSM
+    }
--- a/pyanaconda/modules/subscription/subscription.py
+++ b/pyanaconda/modules/subscription/subscription.py
@@ -17,7 +17,19 @@
 # License and may only be used or replicated with the express permission of
 # Red Hat, Inc.
 #
+import os
+
+from pyanaconda.core import util
+from pyanaconda.core.signal import Signal
+
 from pyanaconda.modules.common.base import KickstartService
+from pyanaconda.modules.common.structures.subscription import SystemPurposeData
+
+from pyanaconda.modules.subscription import system_purpose
+from pyanaconda.modules.subscription.kickstart import SubscriptionKickstartSpecification
+
+from pyanaconda.anaconda_loggers import get_module_logger
+log = get_module_logger(__name__)
 
 
 class SubscriptionService(KickstartService):
@@ -26,3 +38,139 @@ class SubscriptionService(KickstartService):
     def __init__(self):
         super().__init__()
 
+        # system purpose
+
+        self._valid_roles = []
+        self._valid_slas = []
+        self._valid_usage_types = []
+
+        self._system_purpose_data = SystemPurposeData()
+        self.system_purpose_data_changed = Signal()
+
+        self._load_valid_system_purpose_values()
+
+        # FIXME: handle rhsm.service startup in a safe manner
+
+    @property
+    def kickstart_specification(self):
+        """Return the kickstart specification."""
+        return SubscriptionKickstartSpecification
+
+    def process_kickstart(self, data):
+        """Process the kickstart data."""
+        log.debug("Processing kickstart data...")
+
+        # system purpose
+        #
+        # Try if any of the values in kickstart match a valid field.
+        # If it does, write the valid field value instead of the value from kickstart.
+        #
+        # This way a value in kickstart that has a different case and/or trailing white space
+        # can still be used to preselect a value in a UI instead of being marked as a custom
+        # user specified value.
+        system_purpose_data = SystemPurposeData()
+
+        system_purpose_data.role = system_purpose.process_field(
+            data.syspurpose.role,
+            self.valid_roles,
+            "role"
+        )
+
+        system_purpose_data.sla = system_purpose.process_field(
+            data.syspurpose.sla,
+            self.valid_slas,
+            "sla"
+        )
+
+        system_purpose_data.usage = system_purpose.process_field(
+            data.syspurpose.usage,
+            self.valid_usage_types,
+            "usage"
+        )
+
+        if data.syspurpose.addons:
+            # As we do not have a list of valid addons available, we just use what was provided
+            # by the user in kickstart verbatim.
+            system_purpose_data.addons = data.syspurpose.addons
+
+        self.set_system_purpose_data(system_purpose_data)
+
+    def setup_kickstart(self, data):
+        """Return the kickstart string."""
+
+        # system purpose
+        data.syspurpose.role = self.system_purpose_data.role
+        data.syspurpose.sla = self.system_purpose_data.sla
+        data.syspurpose.usage = self.system_purpose_data.usage
+        data.syspurpose.addons = self.system_purpose_data.addons
+
+    # system purpose configuration
+
+    def _load_valid_system_purpose_values(self):
+        """Load lists of valid roles, SLAs and usage types.
+
+        About role/sla/validity:
+        - an older installation image might have older list of valid fields,
+          missing fields that have become valid after the image has been released
+        - fields that have been valid in the past might be dropped in the future
+        - there is no list of valid addons
+
+        Due to this we need to take into account that the listing might not always be
+        comprehensive and that we need to allow what might on a first glance look like
+        invalid values to be written to the target system.
+        """
+        roles, slas, usage_types = system_purpose.get_valid_fields()
+        self._valid_roles = roles
+        self._valid_slas = slas
+        self._valid_usage_types = usage_types
+
+    @property
+    def valid_roles(self):
+        """Return a list of valid roles.
+
+        :return: list of valid roles
+        :rtype: list of strings
+        """
+        return self._valid_roles
+
+    @property
+    def valid_slas(self):
+        """Return a list of valid SLAs.
+
+        :return: list of valid SLAs
+        :rtype: list of strings
+        """
+        return self._valid_slas
+
+    @property
+    def valid_usage_types(self):
+        """Return a list of valid usage types.
+
+        :return: list of valid usage types
+        :rtype: list of strings
+        """
+        return self._valid_usage_types
+
+    @property
+    def system_purpose_data(self):
+        """System purpose data.
+
+        A DBus structure holding information about system purpose,
+        such as role, sla, usage and addons.
+
+        :return: system purpose DBus structure
+        :rtype: DBusData instance
+        """
+        return self._system_purpose_data
+
+    def set_system_purpose_data(self, system_purpose_data):
+        """Set system purpose data.
+
+        Set the complete DBus structure containing system purpose data.
+
+        :param system_purpose_data: system purpose data structure to be set
+        :type system_purpose_data: DBus structure
+        """
+        self._system_purpose_data = system_purpose_data
+        self.system_purpose_data_changed.emit()
+        log.debug("System purpose data set to %s.", system_purpose_data)
--- a/pyanaconda/modules/subscription/subscription.py
+++ b/pyanaconda/modules/subscription/subscription.py
@@ -1,0 +1,28 @@
+#
+# Kickstart module for subscription handling.
+#
+# Copyright (C) 2020 Red Hat, Inc.
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# the GNU General Public License v.2, or (at your option) any later version.
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY expressed or implied, including the implied warranties of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General
+# Public License for more details.  You should have received a copy of the
+# GNU General Public License along with this program; if not, write to the
+# Free Software Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
+# 02110-1301, USA.  Any Red Hat trademarks that are incorporated in the
+# source code or documentation are not subject to the GNU General Public
+# License and may only be used or replicated with the express permission of
+# Red Hat, Inc.
+#
+from pyanaconda.modules.common.base import KickstartService
+
+
+class SubscriptionService(KickstartService):
+    """The Subscription service."""
+
+    def __init__(self):
+        super().__init__()
+
--- a/pyanaconda/modules/subscription/subscription_interface.py
+++ b/pyanaconda/modules/subscription/subscription_interface.py
@@ -1,0 +1,29 @@
+#
+# DBus interface for the subscription module.
+#
+# Copyright (C) 2020 Red Hat, Inc.
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# the GNU General Public License v.2, or (at your option) any later version.
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY expressed or implied, including the implied warranties of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General
+# Public License for more details.  You should have received a copy of the
+# GNU General Public License along with this program; if not, write to the
+# Free Software Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
+# 02110-1301, USA.  Any Red Hat trademarks that are incorporated in the
+# source code or documentation are not subject to the GNU General Public
+# License and may only be used or replicated with the express permission of
+# Red Hat, Inc.
+#
+from pyanaconda.modules.common.constants.services import SUBSCRIPTION
+from pyanaconda.modules.common.base import KickstartModuleInterface
+from pyanaconda.dbus.interface import dbus_interface
+
+
+@dbus_interface(SUBSCRIPTION.interface_name)
+class SubscriptionInterface(KickstartModuleInterface):
+    """DBus interface for the Subscription service."""
+
+
--- a/pyanaconda/modules/subscription/subscription_interface.py
+++ b/pyanaconda/modules/subscription/subscription_interface.py
@@ -19,11 +19,63 @@
 #
 from pyanaconda.modules.common.constants.services import SUBSCRIPTION
 from pyanaconda.modules.common.base import KickstartModuleInterface
-from pyanaconda.dbus.interface import dbus_interface
+from pyanaconda.modules.common.structures.subscription import SystemPurposeData
+from dasbus.server.interface import dbus_interface
+from dasbus.server.property import emits_properties_changed
+from dasbus.typing import *  # pylint: disable=wildcard-import
 
 
 @dbus_interface(SUBSCRIPTION.interface_name)
 class SubscriptionInterface(KickstartModuleInterface):
     """DBus interface for the Subscription service."""
 
+    def connect_signals(self):
+        super().connect_signals()
+        self.watch_property("SystemPurposeData", self.implementation.system_purpose_data_changed)
 
+    def GetValidRoles(self) -> List[Str]:
+        """Return all valid system purpose roles.
+
+        These are OS release specific, but could look like this:
+
+        "Red Hat Enterprise Linux Server"
+        "Red Hat Enterprise Linux Workstation"
+        "Red Hat Enterprise Linux Compute Node"
+        """
+        return self.implementation.valid_roles
+
+    def GetValidSLAs(self) -> List[Str]:
+        """Return all valid system purpose SLAs.
+
+        These are OS release specific, but could look like this:
+
+        "Premium"
+        "Standard"
+        "Self-Support"
+        """
+        return self.implementation.valid_slas
+
+    def GetValidUsageTypes(self) -> List[Str]:
+        """List all valid system purpose usage types.
+
+        These are OS release specific, but could look like this:
+
+        "Production",
+        "Development/Test",
+        "Disaster Recovery"
+        """
+        return self.implementation.valid_usage_types
+
+    @property
+    def SystemPurposeData(self) -> Structure:
+        """Return DBus structure holding current system purpose data."""
+        return SystemPurposeData.to_structure(self.implementation.system_purpose_data)
+
+    @emits_properties_changed
+    def SetSystemPurposeData(self, system_purpose_data: Structure):
+        """Set a new DBus structure holding system purpose data.
+
+        :param system_purpose_data: DBus structure corresponding to SystemPurposeData
+        """
+        converted_data = SystemPurposeData.from_structure(system_purpose_data)
+        self.implementation.set_system_purpose_data(converted_data)
--- a/pyanaconda/modules/subscription/system_purpose.py
+++ b/pyanaconda/modules/subscription/system_purpose.py
@@ -1,0 +1,140 @@
+#
+# System purpose library.
+#
+# Copyright (C) 2020 Red Hat, Inc.
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# the GNU General Public License v.2, or (at your option) any later version.
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY expressed or implied, including the implied warranties of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General
+# Public License for more details.  You should have received a copy of the
+# GNU General Public License along with this program; if not, write to the
+# Free Software Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
+# 02110-1301, USA.  Any Red Hat trademarks that are incorporated in the
+# source code or documentation are not subject to the GNU General Public
+# License and may only be used or replicated with the express permission of
+# Red Hat, Inc.
+#
+
+import os
+import json
+
+from pyanaconda.core import util
+from pyanaconda.anaconda_loggers import get_module_logger
+log = get_module_logger(__name__)
+
+VALID_FIELDS_FILE_PATH = "/etc/rhsm/syspurpose/valid_fields.json"
+
+
+def get_valid_fields(valid_fields_file_path=VALID_FIELDS_FILE_PATH):
+    """Get valid role, sla and usage fields for system purpose use.
+
+    If no valid fields are provided, the fields file is not found or can not be
+    parsed then all three lists will be empty.
+
+    :param str valid_fields_file_path: path to a JSON file holding the valid field listings
+
+    :return: role, sla and usage type lists
+    :rtype: [roles], [slas], [usages]
+    """
+    valid_roles = []
+    valid_slas = []
+    valid_usage_types = []
+    if os.path.exists(valid_fields_file_path):
+        try:
+            with open(valid_fields_file_path, "rt") as f:
+                valid_fields_json = json.load(f)
+                valid_roles = valid_fields_json.get("role", [])
+                valid_slas = valid_fields_json.get("service_level_agreement", [])
+                valid_usage_types = valid_fields_json.get("usage", [])
+        except (IOError, json.JSONDecodeError):
+            log.exception("parsing of syspurpose valid fields file at %s failed",
+                          valid_fields_file_path)
+    else:
+        log.warning("system purpose valid fields file not found at %s", valid_fields_file_path)
+    return valid_roles, valid_slas, valid_usage_types
+
+
+def _normalize_field(raw_field):
+    """Normalize a field for matching.
+
+    Fields specified in free form by users can have different case or trailing white space,
+    while still technically being a match on a valid field.
+
+    So convert the field to lower case and strip any trailing white space and return the result.
+
+    :param str raw_field: raw not normalized field
+    :return: normalized field suitable for matching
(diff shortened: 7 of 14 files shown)


tests/nosetests/pyanaconda_tests/module_subscription_tests.py
```diff
@@ -0,0 +1,261 @@
...
+            "role" : get_variant(Str, "foo"),
+            "sla" : get_variant(Str, "bar"),
+            "usage" : get_variant(Str, "baz"),
+            "addons" : get_variant(List[Str], ["a", "b", "c"])
+        }
+
+        # compare the two
+        self.assertDictEqual(SystemPurposeData.to_structure(system_purpose_data), expected_dict)
+
+    def default_property_values_test(self):
```
poncovka: This test is a mess. Please, focus on testing units of the code. There should be always one test dedicated to testing one member of the DBus API of the Subscription module. So one test for the DBus property `SystemPurposeData`, one test the DBus method `GetValidRoles`, etc. Then there should be kickstart tests for the most common use cases. And after all that you should check the coverage of the code and if there are use cases that are not covered, you can add and extend tests for these use cases.

Otherwise, we can end up with a lot of useless tests, but miss the most critical ones.

pyanaconda/modules/subscription/subscription.py
```diff
@@ -0,0 +1,204 @@
...
+# License and may only be used or replicated with the express permission of
+# Red Hat, Inc.
+#
+import os
+
+from pyanaconda.core import util
+from pyanaconda.core.signal import Signal
+
+from pyanaconda.modules.common.base import KickstartService
+from pyanaconda.modules.common.structures.subscription import SystemPurposeData
```
poncovka: Where is this class?

M4rtinK:
@VladimirSlavik No problem, feedback like this is still appreciated - especially here where I am in many cases porting bigger chunks of code from rhel-8 branch, so there is a more chance I overlook something. :) Thanks! :)

pyanaconda/modules/subscription/subscription.py
```diff
@@ -0,0 +1,204 @@
...
+        self._valid_usage_types = usage_types
+
+    def _process_role(self, data):
+        if data.syspurpose.role:
+            role_match = system_purpose.match_field(data.syspurpose.role, self.valid_roles)
+        else:
+            role_match = None
+
+        if role_match:
+            log.info("role value %s from kickstart matched to know valid field %s", data.syspurpose.role, role_match)
```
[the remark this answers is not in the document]
M4rtinK: Fixed, thanks! :)

data/product.d/rhel.conf
```diff
@@ -3,6 +3,19 @@
 [Product]
 product_name = Red Hat Enterprise Linux
 
+[Anaconda]
```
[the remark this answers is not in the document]
M4rtinK: What it does it to make sure the Subscription module is run once it's upstream code lands in RHEL in the future. 

The base config file has the same list of modules already, minus the Subscription module & as we can't append to values when when override them in the config, we need to set the whole list here even if we want to add a single value.

But good point, I think we should make this at least a separate commit with the explanation above attached so that it is evident why the set of modules to be run for RHEL has been configured in such a way.

tests/nosetests/pyanaconda_tests/module_subscription_tests.py
```diff
@@ -0,0 +1,261 @@
...
+        system_purpose_data.sla = "bar"
+        system_purpose_data.usage = "baz"
+        system_purpose_data.addons = ["a", "b", "c"]
+
+        structure = SystemPurposeData.to_structure(system_purpose_data)
+        self.subscription_interface.SetSystemPurposeData(structure)
+        self.callback.assert_called_once()
+
+        output_structure = self.subscription_interface.SystemPurposeData
+        output_system_purpose_data = SystemPurposeData.from_structure(output_structure)
```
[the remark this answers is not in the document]
M4rtinK: Indeed! Looks like I kinda forgot to complete that one test function. Fixed now. :P

tests/nosetests/pyanaconda_tests/module_subscription_tests.py
```diff
@@ -0,0 +1,261 @@
...
+        # - or they contain a non zero amount of items
+        valid_roles = self.subscription_interface.GetValidRoles()
+        valid_slas = self.subscription_interface.GetValidSLAs()
+        valid_usage_types = self.subscription_interface.GetValidUsageTypes()
+        self.assertTrue(len(valid_roles) > 0 or valid_roles == [])
+        self.assertTrue(len(valid_slas) > 0 or valid_slas == [])
+        self.assertTrue(len(valid_usage_types) > 0 or valid_usage_types == [])
+
+    def set_system_purpose_test(self):
+        """Test if setting system prupose data from DBUS works correctly."""
```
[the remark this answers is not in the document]
M4rtinK: Fixed. :)

pyanaconda/modules/subscription/Makefile.am
```diff
@@ -0,0 +1,21 @@
+#
+# Copyright (C) 2018  Red Hat, Inc.
```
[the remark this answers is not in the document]
M4rtinK: Fixed.

pyanaconda/modules/subscription/Makefile.am
```diff
@@ -0,0 +1,21 @@
...
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+pkgpyexecdir = $(pyexecdir)/py$(PACKAGE_NAME)
+securitydir = $(pkgpyexecdir)/modules/security
+security_PYTHON = $(srcdir)/*.py
```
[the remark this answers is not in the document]
M4rtinK: Good catch! Which reminds me of the fact, that for some reasons CI tests have not started on this automatically, which should have found this...

pyanaconda/modules/subscription/kickstart.py
```diff
@@ -0,0 +1,28 @@
...
+# 02110-1301, USA.  Any Red Hat trademarks that are incorporated in the
+# source code or documentation are not subject to the GNU General Public
+# License and may only be used or replicated with the express permission of
+# Red Hat, Inc.
+#
+from pyanaconda.core.kickstart import VERSION, KickstartSpecification, commands as COMMANDS
+
+class SubscriptionKickstartSpecification(KickstartSpecification):
+
+    version = VERSION
```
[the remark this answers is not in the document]
M4rtinK: Nice simplification, dropping this.

pyanaconda/modules/subscription/subscription.py
```diff
@@ -0,0 +1,204 @@
...
+from pyanaconda.modules.common.base import KickstartService
+from pyanaconda.modules.common.structures.subscription import SystemPurposeData
+
+from pyanaconda.modules.subscription import system_purpose
+from pyanaconda.modules.subscription.kickstart import SubscriptionKickstartSpecification
+
+from pyanaconda.anaconda_loggers import get_module_logger
+log = get_module_logger(__name__)
+
+class SubscriptionModule(KickstartService):
```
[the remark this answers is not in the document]
M4rtinK: Yep, makes sense given the parent changed name as well.

tests/nosetests/pyanaconda_tests/module_subscription_tests.py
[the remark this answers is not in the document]
VladimirSlavik: On a second thought, it might be a good idea to grep for that? You have the same in name of this PR, so it's probably something you (mis)type automatically without thinking...

pyanaconda/modules/subscription/subscription.py
```diff
@@ -0,0 +1,204 @@
...
+        # If it does, write the valid field value instead of the value from kickstart.
+        #
+        # This way a value in kickstart that has a different case and/or trailing white space
+        # can still be used to preselect a value in a UI instead of being marked as a custom
+        # user specified value.
+        role = self._process_role(data)
+        sla = self._process_sla(data)
+        usage = self._process_usage(data)
+
+        system_purpose_data = self.system_purpose_data
```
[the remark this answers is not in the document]
M4rtinK: Yeah, we do not expect clients to set system purpose data via DBus before kickstart has been parsed & it can'T be set via other mechanisms (boot options/configs), so there should be no issue with using a newly created SystemPurposeData instance at kickstart parsing time.

pyanaconda/modules/subscription/subscription.py
```diff
@@ -0,0 +1,204 @@
...
+        system_purpose_data.usage = usage
+
+        if data.syspurpose.addons:
+            # As we do not have a list of valid addons available, we just use what was provided
+            # by the user in kickstart verbatim.
+            system_purpose_data.addons = data.syspurpose.addons
+
+        self.set_system_purpose_data(system_purpose_data)
+
+    def generate_kickstart(self):
```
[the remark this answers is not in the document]
M4rtinK: Changed.

pyanaconda/modules/subscription/subscription.py
```diff
@@ -0,0 +1,204 @@
...
+        # system purpose
+
+        self._valid_roles = {}
+        self._valid_slas = {}
+        self._valid_usage_types = {}
+
+        self._system_purpose_data = SystemPurposeData()
+        self.system_purpose_data_changed = Signal()
+
+        self._load_valid_system_purpose_values()
```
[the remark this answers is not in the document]
M4rtinK: Added.

pyanaconda/modules/subscription/subscription.py
```diff
@@ -0,0 +1,204 @@
...
+        Due to this we need to take into account that the listing might not always be
+        comprehensive and that we need to allow what might on a first glance look like
+        invalid values to be written to the target system.
+        """
+        roles, slas, usage_types = system_purpose.get_valid_fields()
+        self._valid_roles = roles
+        self._valid_slas = slas
+        self._valid_usage_types = usage_types
+
+    def _process_role(self, data):
```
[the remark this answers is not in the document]
M4rtinK: You mean subclass of the Subscription kickstart service, right ? Like we already have for firewall handling in the network DBus module for example.

pyanaconda/modules/subscription/subscription.py
```diff
@@ -0,0 +1,204 @@
...
+
+        A DBus structure holding information about system purpose,
+        such as role, sla, usage and addons.
+
+        :return: system purpose DBus structure
+        :rtype: DBusData instance
+        """
+        return self._system_purpose_data
+
+    def set_system_purpose_data(self, system_purpose_data):
```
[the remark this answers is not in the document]
M4rtinK: Added.

pyanaconda/modules/subscription/system_purpose.py
```diff
@@ -0,0 +1,94 @@
+#
+# System purpose module.
```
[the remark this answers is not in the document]
M4rtinK: Sounds good.

pyanaconda/modules/subscription/subscription_interface.py
```diff
@@ -0,0 +1,56 @@
...
+@dbus_interface(SUBSCRIPTION.interface_name)
+class SubscriptionInterface(KickstartModuleInterface):
+    """DBus interface for the Subscription module."""
+
+    def connect_signals(self):
+        super().connect_signals()
+        self.watch_property("SystemPurposeData", self.implementation.system_purpose_data_changed)
+
+    def GetValidRoles(self) -> List[Str]:
+        """Return all valid system purpose roles."""
```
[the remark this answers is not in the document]
M4rtinK: I'll add some with a disclaimer that these are really environment & time specific. Eq different RHEL releases could have different valid values, even over time for a single release.

tests/nosetests/pyanaconda_tests/module_subscription_tests.py
```diff
@@ -0,0 +1,261 @@
...
+        # create its expected representation
+        expected_dict = {
+            "role" : get_variant(Str, "foo"),
+            "sla" : get_variant(Str, "bar"),
+            "usage" : get_variant(Str, "baz"),
+            "addons" : get_variant(List[Str], ["a", "b", "c"])
+        }
+
+        # compare the two
+        self.assertDictEqual(SystemPurposeData.to_structure(system_purpose_data), expected_dict)
```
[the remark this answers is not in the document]
M4rtinK: Changed.

tests/nosetests/pyanaconda_tests/module_subscription_tests.py
```diff
@@ -0,0 +1,261 @@
...
+
+        structure = SystemPurposeData.to_structure(system_purpose_data)
+        self.subscription_interface.SetSystemPurposeData(structure)
+        self.callback.assert_called_once()
+
+        output_structure = self.subscription_interface.SystemPurposeData
+        output_system_purpose_data = SystemPurposeData.from_structure(output_structure)
+
+
+    def system_purpose_valid_json_parsing_test(self):
```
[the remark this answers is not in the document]
M4rtinK: Good point, moved system_purpose library tests to a separate TestCase sub class.

tests/nosetests/pyanaconda_tests/module_subscription_tests.py
```diff
@@ -0,0 +1,261 @@
...
+    def ks_out_set_usage_test(self):
+        """Check kickstart with usage being used."""
+        ks_in = 'syspurpose --usage="FOO USAGE"'
+        ks_out = '# Intended system purpose\nsyspurpose --usage="FOO USAGE"'
+        self._test_kickstart(ks_in, ks_out)
+
+    def ks_out_set_addons_test(self):
+        """Check kickstart with addons being used."""
+        ks_in = 'syspurpose --addon="Foo Product" --addon="Bar Feature"'
+        ks_out = '# Intended system purpose\nsyspurpose --addon="Foo Product" --addon="Bar Feature"'
```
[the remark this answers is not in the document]
M4rtinK: Cool! That's really nice. :)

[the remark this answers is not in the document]
M4rtinK: Specially to solve the issue of keeping the strings having a sane line length.

tests/nosetests/pyanaconda_tests/product_test.py
```diff
@@ -68,6 +68,16 @@ def _check_default_product(self, product_name, variant_name, file_names):
...
         config.validate()
 
+    def _get_config(self, product_name, variant_name=""):
+        """Get parsed config file."""
+        config = AnacondaConfiguration.from_defaults()
+        paths = self._loader.collect_configurations(product_name, variant_name)
+
+        for path in paths:
+            config.read(path)
+
```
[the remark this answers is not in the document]
M4rtinK: Added.

tests/nosetests/pyanaconda_tests/product_test.py
```diff
@@ -110,6 +120,41 @@ def rhel_products_test(self):
...
+        expected_modules = [
+            "org.fedoraproject.Anaconda.Modules.Timezone",
+            "org.fedoraproject.Anaconda.Modules.Network",
+            "org.fedoraproject.Anaconda.Modules.Localization",
+            "org.fedoraproject.Anaconda.Modules.Security",
+            "org.fedoraproject.Anaconda.Modules.Users",
+            "org.fedoraproject.Anaconda.Modules.Payloads",
+            "org.fedoraproject.Anaconda.Modules.Storage",
+            "org.fedoraproject.Anaconda.Modules.Services",
+            "org.fedoraproject.Anaconda.Modules.Subscription"
```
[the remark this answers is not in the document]
M4rtinK: Good point, that should be much more maintainable. Also I guess we could use a similar technique later if we ever have other such cases, possibly including Fedora only modules.

pyanaconda/modules/subscription/subscription.py
M4rtinK: Should be there now. :)

tests/nosetests/pyanaconda_tests/module_subscription_tests.py
M4rtinK: Yeah, this should be already covered by the individual API tests. And another issue is keeping this one test actually up to date when additional methods/properties are added. So let's scrap this. :)

pyanaconda/modules/subscription/subscription.py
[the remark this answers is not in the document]
poncovka: I thought this is the method that that causes the activation of the RHSM service. My mistake :-) Maybe I would just call it right after definitions of the `valid_` attributes. Otherwise, it can stay as it is. Btw. the attributes are initialized as dictionaries, but they should be lists.

pyanaconda/modules/subscription/subscription.py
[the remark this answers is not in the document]
poncovka: I have changed my previous comment, because it didn't make sense.

poncovka:
Please, rebase your PR to the master branch.

tests/nosetests/pyanaconda_tests/module_subscription_tests.py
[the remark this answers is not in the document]
M4rtinK: Yep, should be fixed now as far as I can tell. :)

pyanaconda/modules/subscription/subscription.py
```diff
@@ -26,3 +38,183 @@ class SubscriptionService(KickstartService):
...
+    def _process_sla(self, data):
+        if data.syspurpose.sla:
+            sla_match = system_purpose.match_field(data.syspurpose.sla, self.valid_slas)
+        else:
+            sla_match = None
+
+        if sla_match:
+            log.info("SLA value %s from kickstart matched to known valid field %s",
+                     data.syspurpose.sla,
+                     sla_match)
```
jkonecny12: Write this on two lines, I don't think there is a reason to add each parameter on a separate line in this case.

pyanaconda/modules/subscription/subscription.py
```diff
@@ -26,3 +38,183 @@ class SubscriptionService(KickstartService):
...
+            return data.syspurpose.sla
+        else:
+            return ""
+
+    def _process_usage(self, data):
+        if data.syspurpose.usage:
+            usage_match = system_purpose.match_field(data.syspurpose.usage,
+                                                     self._valid_usage_types)
+        else:
+            usage_match = None
```
jkonecny12: What about returning `None` if  the argument doesn't exists. You then can remove this from three places to just one.
Or in general, wouldn't it be better to have all the `process_*` private methods as part of the `system_purpose` library. It seems to me that it would make the main service cleaner and avoid repetition.

pyanaconda/modules/subscription/subscription_interface.py
```diff
@@ -19,11 +19,60 @@
...
+        return self.implementation.valid_usage_types
+
+    @property
+    def SystemPurposeData(self) -> Structure:
+        """Return DBus structure holding current system purpose data."""
+        return SystemPurposeData.to_structure(self.implementation.system_purpose_data)
+
+    @emits_properties_changed
+    def SetSystemPurposeData(self, system_purpose_data: Structure):
+        """Set a new DBus structure holding system purpose data."""
```
jkonecny12: It would be great to point what DBus structure you are expecting. From the DBus point of view it could be any dictionary.

pyanaconda/modules/subscription/system_purpose.py
```diff
@@ -0,0 +1,96 @@
...
+                valid_usage_types = valid_fields_json.get("usage", [])
+        except (IOError, json.JSONDecodeError):
+            log.exception("parsing of syspurpose valid fields file at %s failed",
+                          valid_fields_file_path)
+    else:
+        log.warning("system purpose valid fields file not found at %s", valid_fields_file_path)
+    return valid_roles, valid_slas, valid_usage_types
+
+
+def normalize_field(raw_field):
```
jkonecny12: This should be private method from the sense that the only communication with the outer world should be from this file.

pyanaconda/modules/subscription/system_purpose.py
```diff
@@ -0,0 +1,96 @@
...
+    :rtype: str
+    """
+    return raw_field.strip().lower()
+
+
+def match_field(raw_field, valid_fields):
+    """Try to match the field on an item in a list of fields.
+
+    If a match is found return the first matching item from the list.
+    If np match is found, return None.
```
jkonecny12: Probably a typo?

```suggestion
    If no match is found, return None.
```

pyanaconda/modules/subscription/system_purpose.py
```diff
@@ -0,0 +1,96 @@
...
+
+    :param raw_field str: field to match
+    :param list valid_fields: list of valid fields to match against
+    :return: a matching valid field or None if no match is found
+    "rtype: str or None
```
jkonecny12:
```suggestion
    :rtype: str or None
```

pyanaconda/modules/subscription/system_purpose.py
```diff
@@ -0,0 +1,96 @@
...
+    """
+    matching_valid_field = None
+    normalized_field = normalize_field(raw_field)
+
+    for valid_field in valid_fields:
+        if normalized_field == normalize_field(valid_field):
+            # looks like the fields match, no need to search
+            # any further
```
jkonecny12: You can place this comment on one line. I don't think it will exceed 99 characters limit.

tests/nosetests/pyanaconda_tests/module_subscription_tests.py
```diff
@@ -0,0 +1,278 @@
...
+            self.assertListEqual(roles, ["role_a", "role_b", "role_c"])
+            self.assertListEqual(slas, ["sla_a", "sla_b", "sla_c"])
+            self.assertListEqual(usage_types, ["usage_a", "usage_b", "usage_c"])
+
+    def normalize_field_test(self):
+        """Test that the system purpose valid field normalization works."""
+        # this should basically just lower case the input
+        self.assertEqual(normalize_field("AAA"), "aaa")
+        self.assertEqual(normalize_field("Ab"), "ab")
+        self.assertEqual(normalize_field("A b C"), "a b c")
```
jkonecny12: Do we need this test? Isn't this already tested by the above test?

data/product.d/rhel.conf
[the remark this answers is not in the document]
M4rtinK: Moved to a separate commit.

pyanaconda/modules/subscription/subscription.py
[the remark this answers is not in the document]
M4rtinK: > Btw. the attributes are initialized as dictionaries, but they should be lists.

Indeed! Fixed, thanks! :)

pyanaconda/modules/subscription/subscription.py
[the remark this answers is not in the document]
M4rtinK: Nice idea, merge the three into single function & put it into the system_purpose library. :)

pyanaconda/modules/subscription/Makefile.am
```diff
@@ -0,0 +1,21 @@
+#
```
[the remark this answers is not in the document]
M4rtinK: Indeed! Now that tests have starter running on this PR it is apparent. :P Fixed!

(grumbles about questionable autotools requirements)

pyanaconda/modules/subscription/__main__.py
```diff
@@ -0,0 +1,6 @@
+from pyanaconda.modules.common import init
```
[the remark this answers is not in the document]
M4rtinK: Looks like none of our `__main__.py` in DBus modules has a license in the file itself ? I guess we might want to to handle that for all modules in a separate PR for consistency.

pyanaconda/modules/subscription/subscription.py
```diff
@@ -0,0 +1,28 @@
...
+# 02110-1301, USA.  Any Red Hat trademarks that are incorporated in the
+# source code or documentation are not subject to the GNU General Public
+# License and may only be used or replicated with the express permission of
+# Red Hat, Inc.
+#
+from pyanaconda.modules.common.base import KickstartService
+
+
+class SubscriptionService(KickstartService):
+    """The Subscription module."""
```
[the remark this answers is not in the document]

pyanaconda/modules/subscription/subscription_interface.py
```diff
@@ -0,0 +1,29 @@
...
+# Red Hat, Inc.
+#
+from pyanaconda.modules.common.constants.services import SUBSCRIPTION
+from pyanaconda.modules.common.base import KickstartModuleInterface
+from pyanaconda.dbus.interface import dbus_interface
+
+
+@dbus_interface(SUBSCRIPTION.interface_name)
+class SubscriptionInterface(KickstartModuleInterface):
+    """DBus interface for the Subscription module."""
```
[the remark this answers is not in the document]
M4rtinK: Fixed.

pyanaconda/modules/subscription/subscription.py
M4rtinK: Code has been refactored and moved to the system_purpose library.

pyanaconda/modules/subscription/subscription.py
M4rtinK: Code was moved to `system_purpose` library.

pyanaconda/modules/subscription/subscription_interface.py
M4rtinK: I'll add a docstring mentioning that DBus structure corresponding to/parseable by the SystemPurposeData is expected.

pyanaconda/modules/subscription/system_purpose.py
M4rtinK: Well, yeah - also since the full matching code is now part of `system_purpose`, let's make all the other functions private.

M4rtinK: The only downside is that we are now importing methods starting with an `_` in the unit tests file, but oh well.

pyanaconda/modules/subscription/system_purpose.py
M4rtinK: Fixed.

pyanaconda/modules/subscription/system_purpose.py
M4rtinK: Fixed, thanks! :)

pyanaconda/modules/subscription/system_purpose.py
M4rtinK: You newer know! :D Changed. :)

tests/nosetests/pyanaconda_tests/module_subscription_tests.py
M4rtinK: I tihnk it is good to have all the `system_purpose` library field processing functions covered.

M4rtinK:
@poncovka Rebased to master. :)

pyanaconda/modules/subscription/system_purpose.py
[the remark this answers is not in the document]
M4rtinK: Fixed.

pyanaconda/modules/subscription/__main__.py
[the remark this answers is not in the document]
jkonecny12: Good point. I agree on that solution.

pyanaconda/modules/subscription/system_purpose.py
```diff
@@ -0,0 +1,140 @@
...
+        if normalized_field == _normalize_field(valid_field):
+            # looks like the fields match, no need to search any further
+            matching_valid_field = valid_field
+            break
+
+    return matching_valid_field
+
+
+def process_field(syspurpose_value, valid_values, value_name):
+    """Process a single system prupose value provided by the user.
```
jkonecny12:
```suggestion
    """Process a single system purpose value provided by the user.
```

tests/nosetests/pyanaconda_tests/module_subscription_tests.py
```diff
@@ -0,0 +1,261 @@
...
+    def set_system_purpose_test(self):
+        """Test if setting system prupose data from DBUS works correctly."""
+        system_purpose_data = SystemPurposeData()
+        system_purpose_data.role = "foo"
+        system_purpose_data.sla = "bar"
+        system_purpose_data.usage = "baz"
+        system_purpose_data.addons = ["a", "b", "c"]
+
+        structure = SystemPurposeData.to_structure(system_purpose_data)
+        self.subscription_interface.SetSystemPurposeData(structure)
```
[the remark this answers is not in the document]
M4rtinK: OK, thanks for the suggestion - lets use that. :)

tests/nosetests/pyanaconda_tests/module_subscription_tests.py
```diff
@@ -0,0 +1,261 @@
...
+    """Test DBus interface for the subscription module."""
+
+    def setUp(self):
+        """Set up the subscription module."""
+        self.subscription_module = SubscriptionModule()
+        self.subscription_interface = SubscriptionInterface(self.subscription_module)
+
+        # Connect to the properties changed signal.
+        self.callback = Mock()
+        self.subscription_interface.PropertiesChanged.connect(self.callback)
```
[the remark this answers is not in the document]
M4rtinK: Sounds good, changed. :)

pyanaconda/modules/subscription/system_purpose.py
M4rtinK: Thanks, fixed! :)

pyanaconda/modules/subscription/subscription.py
```diff
@@ -0,0 +1,170 @@
...
+        # FIXME: handle rhsm.service startup in a safe manner
+
+    @property
+    def kickstart_specification(self):
+        """Return the kickstart specification."""
+        return SubscriptionKickstartSpecification
+
+    def process_kickstart(self, data):
+        """Process the kickstart data."""
+        log.debug("Processing kickstart data...")
```
poncovka: You can remove this line. The logging is handled in the method that calls this one.

pyanaconda/modules/subscription/subscription.py
```diff
@@ -0,0 +1,170 @@
...
+        # can still be used to preselect a value in a UI instead of being marked as a custom
+        # user specified value.
+        role = system_purpose.process_field(data.syspurpose.role, self.valid_roles, "role")
+        sla = system_purpose.process_field(data.syspurpose.sla, self.valid_slas, "sla")
+        usage = system_purpose.process_field(data.syspurpose.usage,
+                                             self.valid_usage_types,
+                                             "usage")
+
+        system_purpose_data = SystemPurposeData()
+        system_purpose_data.role = role
```
poncovka: I would try to avoid the extra variables.

```
system_purpose_data.role = system_purpose.process_field(
    data.syspurpose.role,
    self.valid_roles,
    "role"
)

```

pyanaconda/modules/subscription/subscription.py
```diff
@@ -0,0 +1,170 @@
...
+        if data.syspurpose.addons:
+            # As we do not have a list of valid addons available, we just use what was provided
+            # by the user in kickstart verbatim.
+            system_purpose_data.addons = data.syspurpose.addons
+
+        self.set_system_purpose_data(system_purpose_data)
+
+    def setup_kickstart(self, data):
+        """Return the kickstart string."""
+        log.debug("Generating kickstart data...")
```
poncovka: You can also remove this line.

pyanaconda/modules/subscription/subscription.py
```diff
@@ -0,0 +1,170 @@
...
+
+        # system purpose
+        data.syspurpose.role = self.system_purpose_data.role
+        data.syspurpose.sla = self.system_purpose_data.sla
+        data.syspurpose.usage = self.system_purpose_data.usage
+        data.syspurpose.addons = self.system_purpose_data.addons
+
+        return str(data)
```
poncovka: The method `setup_kickstart` modifies the data object. It doesn't return anything.

tests/nosetests/pyanaconda_tests/module_subscription_tests.py
```diff
@@ -0,0 +1,311 @@
...
+        structure = self.subscription_interface.SystemPurposeData
+        system_purpose_data = SystemPurposeData.from_structure(structure)
+        self.assertEqual(system_purpose_data.role, 'FOO')
+        self.assertEqual(system_purpose_data.sla, 'BAR')
+        self.assertEqual(system_purpose_data.usage, 'BAZ')
+        self.assertEqual(system_purpose_data.addons, ["Foo Product", "Bar Feature"])
+
+    def ks_set_nothing_test(self):
+        """Test what happens if just the syspurpose command is used."""
+        self.subscription_interface.ReadKickstart('syspurpose')
```
poncovka: Please, merge this test with `ks_out_command_only_test`. It is not a good practice to call `ReadKickstart` directly in the tests. You would have to check the return value to make sure that the parsing was sucesfull.

tests/nosetests/pyanaconda_tests/module_subscription_tests.py
```diff
@@ -0,0 +1,311 @@
...
+        self.assertEqual(output_system_purpose_data.role, "foo")
+        self.assertEqual(output_system_purpose_data.sla, "bar")
+        self.assertEqual(output_system_purpose_data.usage, "baz")
+        self.assertEqual(output_system_purpose_data.addons, ["a", "b", "c"])
+
+    def ks_set_all_test(self):
+        """Test if setting all options from kickstart works correctly."""
+        self.subscription_interface.ReadKickstart(
+            'syspurpose --role="FOO" --sla="BAR" --usage="BAZ" '
+            '--addon="Foo Product" --addon="Bar Feature"'
```
poncovka: Please, merge this test with `ks_out_set_all_usage_test`,

pyanaconda/modules/subscription/subscription.py
M4rtinK: Oh, right, `setup_kickstart()` does that automatically.

pyanaconda/modules/subscription/subscription.py
M4rtinK: OK, lets do that.

pyanaconda/modules/subscription/subscription.py
M4rtinK: Oh, right, `setup_kickstart()` does that automatically.

pyanaconda/modules/subscription/subscription.py
M4rtinK: OK, dropping that.

tests/nosetests/pyanaconda_tests/module_subscription_tests.py
M4rtinK: Good point! Looks like I have not realized the `check_kickstart_interface` function also changes module state, so I can check it, not just the output kickstart. That simplifies things quite a bit, thanks! :)

tests/nosetests/pyanaconda_tests/module_subscription_tests.py
M4rtinK: Merging as well.

pyanaconda/modules/subscription/system_purpose.py
```diff
@@ -0,0 +1,140 @@
...
+    return matching_valid_field
+
+
+def process_field(syspurpose_value, valid_values, value_name):
+    """Process a single system purpose value provided by the user.
+
+    At the moment this value generally comes from kickstart
+    as we don't support free form system purpose value entry in the UI.
+
+    We try to match the user prodived value to value in a lit of well
```
jkonecny12:
```suggestion
    We try to match the user provided value to value in a list of well
```

tests/nosetests/pyanaconda_tests/module_subscription_tests.py
```diff
@@ -0,0 +1,303 @@
...
+
+        output_structure = self.subscription_interface.SystemPurposeData
+        output_system_purpose_data = SystemPurposeData.from_structure(output_structure)
+
+        self.assertEqual(output_system_purpose_data.role, "foo")
+        self.assertEqual(output_system_purpose_data.sla, "bar")
+        self.assertEqual(output_system_purpose_data.usage, "baz")
+        self.assertEqual(output_system_purpose_data.addons, ["a", "b", "c"])
+
+    def _test_kickstart(self, ks_in, ks_out):
```
jkonecny12: For private test method it would be nice to name them `check` instead of test. To make it clear that these are not test units.

data/product.d/rhel.conf
```diff
@@ -3,6 +3,19 @@
 [Product]
```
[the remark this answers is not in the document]
M4rtinK: Fixed. :)

pyanaconda/modules/subscription/system_purpose.py
M4rtinK: Fixed. :)

tests/nosetests/pyanaconda_tests/module_subscription_tests.py
M4rtinK: As with the license in `__main__.py` file - this is how we currently do it in most if not all tests. If we want to change it we should do it in a separate PR for all tests. :)

PR #2372, opened 2020-03-11
Module backport to better handling of sparse images

jkonecny12:
We are taking image file size as input for required space check. This
will not work good for sparse images. To fix this, check for occupied
blocks instead.

Backport of d8af00ce30d39cee9876906e0daed88245549050

Suggested-by: Yuval Turgeman <yturgema@redhat.com>


jkonecny12:
Tested and it works as expected.

PR #2373, opened 2020-03-12
Ellipsize description in configure mount point

rvykydal:
The patches make Description of a disk in disk list which can be sometimes (eg iSCSI disks) pretty long shorter by ellipsizing so that other columns don't run out of the dialog (with scrollbar added).
Like in the screenshot of https://bugzilla.redhat.com/show_bug.cgi?id=1530410#c2

Now to make the complete description available, two approaches are used:
1) Adding a tooltip with complete description. It might be considered too distracting given it is enabled for all descriptions - including the non-ellipsized, but perhaps it is fine, the delay before showing the tooltip seems making this no problem.
2) Making the description column resizable. Added value of this approach can be it makes the info Copy-able, but I am not sure if it would be actually useful for current use case.

A screencast is missing cursor, you have to think it up a bit but I hope it shows enough.
https://rvykydal.fedorapeople.org/ellipsize_disk_description.webm

Personally I'd go only with 2), but I don't mind 1) and see 2) as nice complement.
(I wonder which of the two is actually more discoverable;))

### DIFF
--- a/pyanaconda/ui/gui/spokes/lib/custom_storage_helpers.glade
+++ b/pyanaconda/ui/gui/spokes/lib/custom_storage_helpers.glade
@@ -218,6 +218,9 @@
                       <object class="GtkTreeViewColumn" id="description_column">
                         <property name="spacing">6</property>
                         <property name="title" translatable="yes">Description</property>
+                        <property name="expand">True</property>
+                        <property name="resizable">True</property>
+                        <property name="min-width">100</property>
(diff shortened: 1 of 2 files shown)


jkonecny12:
Hi @AdamWill , this is a nice improvement but it can break OpenQA tests. So heads-up that it's comming.

AdamWill:
Thanks for the heads-up. It'll be easy to adjust the tests when it lands.

PR #2376, opened 2020-03-13
Add the DBus module for SE/HMC

poncovka:
Use the SE/HMC module to install packages from SE/HMC.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -178,6 +178,7 @@ AC_CONFIG_FILES([Makefile
                  pyanaconda/modules/payloads/payload/live_image/Makefile
                  pyanaconda/modules/payloads/base/Makefile
                  pyanaconda/modules/payloads/source/Makefile
+                 pyanaconda/modules/payloads/source/hmc/Makefile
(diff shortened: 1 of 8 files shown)


pyanaconda/modules/payloads/source/hmc/hmc_module.py
```diff
@@ -0,0 +1,67 @@
+#
```
jkonecny12: Do you have a reason to call this `/hmc/hmc_module` instead of `/hmc/hmc` as other sources are?

poncovka: I have started to use that in the Storage module, because it makes the file structure easier to understand, but I can change it back.

jkonecny12: It would be great to stay consistent in this. If we want to go the `_module` way let's create an issue to rename everything or stay as it is now ideally.

poncovka: Ok, fixed.

poncovka:
Updated and tested.

PR #2378, opened 2020-03-13
Support logging of Anaconda DBus modules to files (#1812380)

poncovka:
Add support for writing logs of the Anaconda DBus modules to files.
Enable logging to the file /tmp/storage.log for the Storage module.

Don't set up the Blivet's logger in UI. Blivet is used mostly by the Storage
module now, so the logging should be handled there as well.

Resolves: rhbz#1812380

### DIFF
--- a/pyanaconda/modules/common/__init__.py
+++ b/pyanaconda/modules/common/__init__.py
@@ -15,21 +15,40 @@
 # License and may only be used or replicated with the express permission of
 # Red Hat, Inc.
 #
+import sys
 
 __all__ = ["init"]
 
 
-def init():
+def init(log_filename=None, log_stream=sys.stderr):
     """Do initial configuration of an Anaconda DBus module.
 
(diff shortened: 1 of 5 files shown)


pyanaconda/modules/common/__init__.py
```diff
@@ -15,21 +15,40 @@
 # License and may only be used or replicated with the express permission of
 # Red Hat, Inc.
 #
+import sys
 
 __all__ = ["init"]
 
 
-def init():
+def init(filename=None, stream=sys.stderr):
```
M4rtinK: I think calling those two `log_filename` & `log_stream` would be better - even though both are documented as log related, it would be good to have this also in the name for parameters of such a top level function.

poncovka: Fixed.

poncovka:
Updated.

PR #2381, opened 2020-03-16
Fix fd leak while fchmod failed

yubihong:
In explodeDDRPM(), we will break the while loop if fchmod() exec failed, which
cause the open fd leak. So we close the fd before brak.

### DIFF
--- a/utils/dd/dd_extract.c
+++ b/utils/dd/dd_extract.c
@@ -155,14 +155,17 @@ int main(int argc, char *argv[])
             break;
 
         case 'd':
+            free(directory);
             directory = strdup(optarg);
             break;
 
         case 'k':
+            free(kernel);
             kernel = strdup(optarg);
             break;
 
         case 'r':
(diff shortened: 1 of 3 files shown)


VladimirSlavik:

VladimirSlavik:
Hi @yubihong, thank you! At a first glance it looks ok. Somebody from the team will have to check this - we don't touch the C parts much so it's always an adventure...

yubihong:
Hi,  @VladimirSlavik, I'm sorry about to close the pull request, because my some bad operation, I will add a new pull request to submit these patch.

yubihong:
> Hi @yubihong, thank you! At a first glance it looks ok. Somebody from the team will have to check this - we don't touch the C parts much so it's always an adventure...

Hi, @VladimirSlavik, thank you for your apply. I'm sorry about to close the pull request, because my some bad operation, I will add a new pull request to submit these patch.

PR #2383, opened 2020-03-17
Fix memory leak and fd leak

yubihong:
fd leak：
In explodeDDRPM(), we will break the while loop if fchmod() exec failed, which cause the open fd leak. So we close the fd before break.
memory leak：
The normal usage of dd_extract/dd_list is:
1.dd_extract [-vbmlfh] -k -d -r
2.dd_list [-vh] -k -d -a
But if we use dd_extract -k -k .../dd_list -k -k ...we will lost the first strdup string pionter which cause memory leak; And in the explodeDDRPM(), if we check the status of archive_open failed, we will retrun -1, which cause cpio leak;

### DIFF
--- a/utils/dd/dd_extract.c
+++ b/utils/dd/dd_extract.c
@@ -155,14 +155,32 @@ int main(int argc, char *argv[])
             break;
 
         case 'd':
+            if (directory) {
+                logMessage(ERROR, "Multiple -d arguments\n");
+                show_help();
+                rc = 1;
+                goto cleanup;
+            }
             directory = strdup(optarg);
             break;
 
         case 'k':
+            if (kernel) {
+                logMessage(ERROR, "Multiple -k arguments\n");
+                show_help();
+                rc = 1;
+                goto cleanup;
+            }
             kernel = strdup(optarg);
             break;
 
         case 'r':
+            if (rpm) {
+                logMessage(ERROR, "Multiple -r arguments\n");
+                show_help();
+                rc = 1;
+                goto cleanup;
+            }
             rpm = strdup(optarg);
             break;
 
--- a/utils/dd/dd_list.c
+++ b/utils/dd/dd_list.c
@@ -190,14 +190,32 @@ int main(int argc, char *argv[])
             break;
 
         case 'd':
+            if (path) {
+                logMessage(ERROR, "Multiple -d arguments\n");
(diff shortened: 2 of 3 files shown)


jkonecny12:
Hello, thanks for your contribution!
I have one question, don't we want to block parameter repeating in this case? From my understanding this is not expected behavior and raising an error instead seems to me like a better solution in this case to make it obvious that user shouldn't do this.

Is it useful to use these arguments multiple times?

yubihong:
> Hello, thanks for your contribution!
> I have one question, don't we want to block parameter repeating in this case? From my understanding this is not expected behavior and raising an error instead seems to me like a better solution in this case to make it obvious that user shouldn't do this.
> 
> Is it useful to use these arguments multiple times?

Hello, @jkonecny12 ! Yes, your suggestion is another good solution. If one parameter repeat multiple times, in my patch, we can using the last parameter to continue the process. If you think raising an error is a better solution, I will push another patch.
And I have test using these arguments multiple times, it will enter the case multiple times.

yubihong:
Hello, @jkonecny12, please review if this is what you wanted?

jkonecny12:
I would rather prefer to raise an error. It seems to me that it is then easier to catch calling problem.

utils/dd/dd_extract.c
```diff
@@ -155,14 +155,32 @@ int main(int argc, char *argv[])
             break;
 
         case 'd':
+            if (directory) {
+                logMessage(ERROR, "Multiple -d arguments\n");
+                show_help();
+                rc =1;
```
jkonecny12: Please fix the white spaces:
```
rc = 1;
```

jkonecny12:
Jenkins, it's ok to test.

utils/dd/dd_extract.c
yubihong: Done！

yubihong:
> Looks great. I also tested this code and fixed our existing tests.
> 
> Thanks a lot for this contribution and pointing me on our broken tests!
You're welcome! I‘’m glad to contribute。

yubihong:
> Looks great. I also tested this code and fixed our existing tests.
> 
> Thanks a lot for this contribution and pointing me on our broken tests!

You're welcome! I'm glad to contribute.

PR #2384, opened 2020-03-17
Don't return anything from setup_kickstart

poncovka:
The method `setup_kickstart` shouldn't return anything. It sets up the given
kickstart handler, so there is no reason why it should return the same object.

### DIFF
--- a/pyanaconda/modules/common/base/base.py
+++ b/pyanaconda/modules/common/base/base.py
@@ -134,9 +134,8 @@ class KickstartBaseModule(BaseModule):
         Use the module attributes to set the kickstart data.
 
         :param data: a kickstart handler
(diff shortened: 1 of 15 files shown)


poncovka:
@jkonecny12, I am not changing behaviour, I am just fixing a bug, that was caused by an overlook and copying the same code over and over. We have 25 implementations of `setup_kickstart`. I am fixing only 15 of them, because the remaining ones don't return anything.

Nothing cares about the return value of `setup_kickstart`, so it really doesn't matter whether you return the same value or a different one. It has no effect. The current code could actually make you think that you have this option, while you don't.

jkonecny12:
I understand that, it was just an idea if the proposed approach wouldn't be a better solution.

PR #2389, opened 2020-03-19
Create pyanaconda.payload.dnf

poncovka:
Move the class Payload to pyanaconda.payload.base.
Move the DNF-related code to pyanaconda.payload.dnf.


pyanaconda/payload/dnf/utils.py
```diff
@@ -0,0 +1,139 @@
...
+        base.do_transaction(display=display)
+        exit_reason = "DNF quit"
+    except BaseException as e:  # pylint: disable=broad-except
+        log.error('The transaction process has ended abruptly')
+        log.info(e)
+        import traceback
+        exit_reason = str(e) + traceback.format_exc()
+    finally:
+        base.close()  # Always close this base.
+        queue_instance.put(('quit', str(exit_reason)))
```
jkonecny12: I don't think this is utils code. It's one of the main things the payload is doing not just an utility to achieve something..

pyanaconda/payload/dnf/progress.py
```diff
@@ -0,0 +1,82 @@
+#
```
jkonecny12: I think progress and display should be together. Also the name doesn't really explain what it does.

poncovka:
> These changes looks great but I don't quite understand the reasoning behind that. Could you please tell me first what is the goal?

It is really difficult to orient yourself in the payload-related code. The file `dnfpayload.py` has 1492 lines, so it makes sense to split it to separate files.

About the base classes, there is no good reason why they should be in the `__init__.py` file. It just led to some very weird imports from `pyanaconda.payload`. 

About the class `PackagePayload`, I would prefer to merge it with `DNFPayload`. `PackagePayload` has some private methods that are used only by `DNFPayload`, it just doesn't make any sense to keep them separate at this moment. I know that we talked about it already, so I haven't merge them.

I know that this code will be eventually replaced by the DBus modules, but we still need pieces of this code that we can move to the modules. At this moment, it is very difficult to identify these pieces or create new ones, because there is not a good place where to put them.

pyanaconda/payload/dnf/utils.py
poncovka: I don't think it matters, but I don't care about the name of this file. I just want a library of DNF-related functions and constants. What do you suggest?

pyanaconda/payload/dnf/progress.py
poncovka: I can rename the files, but I don't see why they should be together. They are used at different phases of the installation.

poncovka:
Updated.

poncovka:
Jenkins, test this please.

PR #2394, opened 2020-03-19
Add url payload source

jkonecny12:
Ready for review with the new solution.

Create a new payload url source. This source will handle `http`, `https`, `ftp` and `file` protocols.

This source is not expected to contain plenty of logic but it tries to change how we are handling input settings for URL. Also this source should handle two kickstart commands parts `url` and `repo`.

Progress of the source implementation:
- [x] URL link 
- [x] SSL verification toggle flag
- [x] SSL verification configuration
- [x] Proxy settings 
- [x] Packages/Groups exclude/include
- [x] Repo file (``repo --install``) creation flag
- [x] Repo `cost` value
- [ ] ~Solve `is_ready()` method always return `True` (this will block payload source change)~
- [x] Repo name


pyanaconda/modules/payloads/constants.py
```diff
@@ -33,3 +36,13 @@ class SourceType(Enum):
     LIVE_OS_IMAGE = "LIVE_OS_IMAGE"
     HMC = "HMC"
     CDROM = "CDROM"
+    URL = "URL"
+
+
+@unique
+class URLSourceType(Enum):
```
VladimirSlavik: I'm curious - why are the constants and type definition in two places?

pyanaconda/core/constants.py
```diff
@@ -390,3 +390,9 @@ class PayloadRequirementType(Enum):
 PARTITIONING_METHOD_MANUAL = "MANUAL"
 PARTITIONING_METHOD_INTERACTIVE = "INTERACTIVE"
 PARTITIONING_METHOD_BLIVET = "BLIVET"
+
+# Payload URL source types.
+URL_SOURCE_BASE = "BASE"
```
poncovka: I would call it `URL_SOURCE_BASEURL` and `BASEURL` to keep the names same as in DNF.

pyanaconda/modules/payloads/constants.py
jkonecny12: It's in `core` which is shared for everyone. Modules stuff on the other hand are specific for modules only.

pyanaconda/core/constants.py
```diff
@@ -390,3 +390,9 @@ class PayloadRequirementType(Enum):
...
+URL_SOURCE_MIRRORLIST = "MIRRORLIST"
+URL_SOURCE_METALINK = "METALINK"
+URL_SOURCE_FILE = "FILE"
```
poncovka: This is confusing. I wouldn't mix protocols with types.

pyanaconda/modules/common/structures/payload.py
```diff
@@ -0,0 +1,60 @@
...
+        """Get link to the repository."""
+        return self._url
+
+    @url.setter
+    def url(self, url: Str):
+        """Set link to the repository."""
+        self._url = url
+
+    @property
+    def url_type(self) -> Str:
```
poncovka: This property could be called just `type`.

pyanaconda/modules/payloads/constants.py
poncovka: Maybe `URLType` could be enough?

tests/nosetests/pyanaconda_tests/module_source_url_test.py
```diff
@@ -0,0 +1,144 @@
...
+            self.url_source_interface,
+            *args, **kwargs
+        )
+
+    def type_test(self):
+        """Test URL source has a correct type specified."""
+        self.assertEqual(SourceType.URL.value, self.url_source_interface.Type)
+
+    def set_url_base_source_properties_test(self):
+        url = URLData()
```
poncovka: Could you rename it to `data`, please?

tests/nosetests/pyanaconda_tests/module_source_url_test.py
```diff
@@ -0,0 +1,144 @@
...
+        url.url = "http://example.com/repo"
+        url.url_type = URLSourceType.BASE.value
+
+        self._check_dbus_property(
+            "URL",
+            URLData.to_structure(url)
```
poncovka: It is good to have one test with the raw representation of the data structures to make sure that there are no surprises.

tests/nosetests/pyanaconda_tests/module_source_url_test.py
```diff
@@ -0,0 +1,144 @@
...
+        )
+
+    def set_url_mirrorlist_properties_test(self):
+        url = URLData()
+        url.url = "http://forthehorde.com/mirrorlist?url"
+        url.url_type = URLSourceType.MIRRORLIST.value
```
poncovka: Please, use the constants instead of the enum.

pyanaconda/core/constants.py
```diff
@@ -390,3 +390,9 @@ class PayloadRequirementType(Enum):
 PARTITIONING_METHOD_MANUAL = "MANUAL"
 PARTITIONING_METHOD_INTERACTIVE = "INTERACTIVE"
 PARTITIONING_METHOD_BLIVET = "BLIVET"
+
+# Payload URL source types.
+URL_SOURCE_BASE = "BASE"
+URL_SOURCE_MIRRORLIST = "MIRRORLIST"
```
poncovka: Maybe the prefix of the constants could be `URL_TYPE_`?

pyanaconda/core/constants.py
jkonecny12: Already removed :). Thanks for suggestion.

pyanaconda/core/constants.py
jkonecny12: Fixed.

pyanaconda/core/constants.py
jkonecny12: Fixed.

pyanaconda/modules/payloads/constants.py
jkonecny12: Fixed.

pyanaconda/modules/common/structures/payload.py
jkonecny12: Fixed.

tests/nosetests/pyanaconda_tests/module_source_url_test.py
jkonecny12: Fixed.

tests/nosetests/pyanaconda_tests/module_source_url_test.py
jkonecny12: Fixed.

tests/nosetests/pyanaconda_tests/module_source_url_test.py
jkonecny12: Fixed.

pyanaconda/modules/payloads/constants.py
VladimirSlavik: Makes sense - thank you!

jkonecny12:
**UPDATED:**
* Added ssl configuration
* Implemented changes suggested above

tests/nosetests/pyanaconda_tests/module_source_url_test.py
```diff
@@ -30,6 +37,70 @@ def setUp(self):
...
+            )
+
+    def set_raw_url_with_properties_test(self):
+        data = {
+            "url": get_variant(Str, "http://NaNaNaNaNaNa/Batmaaan"),
+            "type": get_variant(Str, URL_TYPE_METALINK)
+        }
+        self._check_dbus_property(
+            "URL",
+            data)
```
rvykydal: I'd move the ')' to the next line for consistency and beauty.

jkonecny12: Fixed.

jkonecny12:
**UPDATED:**
* Implemented suggestion from @rvykydal 
* Add proxy

pyanaconda/modules/common/structures/payload.py
```diff
@@ -0,0 +1,99 @@
...
+    def url(self, url: Str):
+        """Set link to the repository."""
+        self._url = url
+
+    @property
+    def type(self) -> Str:
+        """Get type of the url.
+
+        Possible values:
+        - BASE
```
poncovka: It should be `BASEURL`.

pyanaconda/modules/payloads/constants.py
```diff
@@ -33,3 +36,12 @@ class SourceType(Enum):
     LIVE_OS_IMAGE = "LIVE_OS_IMAGE"
     HMC = "HMC"
     CDROM = "CDROM"
+    URL = "URL"
+
+
+@unique
+class URLType(Enum):
+    """Types of supported URL sources."""
+    BASE = URL_TYPE_BASEURL
```
poncovka: Please, use `BASEURL` here as well.

pyanaconda/modules/payloads/source/url/url.py
```diff
@@ -0,0 +1,162 @@
...
+    def set_url(self, url):
+        """Set URL data.
+
+        :param url: data structure containing url and type
+        :type url: URLData data structure
+        """
+        try:
+            URLType(url.type)
+        except ValueError:
+            raise InvalidValueError("Invalid source type set '{}'".format(url.type))
```
poncovka: Maybe, I would move this code to a  new method `_validate_url`.

pyanaconda/modules/payloads/source/url/url.py
```diff
@@ -0,0 +1,162 @@
...
+        :type proxy: string in format [protocol://][username[:password]@]host[:port]
+        """
+        log.debug("XXXXX %s", proxy)
+        self._validate_proxy(proxy)
+        self._proxy = proxy
+        self.proxy_changed.emit(proxy)
+        log.debug("Proxy has changed to %s", self._proxy)
+
+    def _validate_proxy(self, proxy):
+        if proxy:
```
poncovka: Please, return if no proxy is set.

pyanaconda/modules/payloads/source/url/url_interface.py
```diff
@@ -0,0 +1,105 @@
...
+        return self.implementation.ssl_configuration
+
+    @emits_properties_changed
+    def SetSSLConfiguration(self, ssl_configuration: Structure):
+        """Set custom SSL configuration for this repository.
+
+        :param ssl_configuration: ssl configuration data structure
+        :type ssl_configuration: SSLConfigurationData data structure
+        """
+        self.implementation.set_ssl_configuration(ssl_configuration)
```
poncovka: You should use `SSLConfigurationData` to transform the structure into a data object. Same above.

tests/nosetests/pyanaconda_tests/module_source_url_test.py
```diff
@@ -0,0 +1,196 @@
...
+
+        with self.assertRaises(InvalidValueError):
+            self._check_dbus_property(
+                "URL",
+                URLData.to_structure(data)
+            )
+
+    def set_raw_url_with_properties_test(self):
+        data = {
+            "url": get_variant(Str, "http://NaNaNaNaNaNa/Batmaaan"),
```
poncovka: :bat:

pyanaconda/modules/payloads/constants.py
jkonecny12: True, sorry for missing that.

pyanaconda/modules/common/structures/payload.py
jkonecny12: Fixed.

pyanaconda/modules/payloads/source/url/url.py
jkonecny12: Good point. Fixed.

pyanaconda/modules/payloads/source/url/url.py
jkonecny12: Fixed.

pyanaconda/modules/payloads/source/url/url_interface.py
jkonecny12: Hmm, true. I just wonder why this wasn't caught by tests.

jkonecny12:
**UPDATED:**
* Implemented notes from @poncovka. Thanks!

jkonecny12:
**UPDATED:**
* Add default values tests.

jkonecny12:
**UPDATED:**
* Add repository cost.

Here's a question about how to implement it. In the old payload you can set cost for the repository or not set it. However, with DBus API we should be able to unset the value too which is little bit challenging. The cost is an integer which is compared I did not found any information if it has to be positive so my guess is that it could also be a negative value. In that case we have to have other way how to disable the cost value.

* The current solution is to add a structure with cost and enabled values. If enabled is False then the cost value is ignored.

Another solutions:
* Add cost and is_enabled properties to the DBus interface. The structure seems like a better solution (easily extensible).
* Add cost setter with two values but then you can't ask if the cost is enabled or not.

jkonecny12:
**UPDATED:**
* Add packaging constrains
  * include packages
  * exclude packages

Ignore ``repo --ignoregroups`` because it's not used in Anaconda anywhere. Instead https://github.com/pykickstart/pykickstart/pull/312 .

This is the last part implemented by this PR. The `is_ready` problem will be solved in other PR.

pyanaconda/modules/common/structures/payload.py
```diff
@@ -0,0 +1,205 @@
...
+        This is a shortcut to avoid calling enable = True all the time.
+
+        :param cost: cost value of this repository
+        :type cost: int
+        """
+        self.cost = cost
+        self.enabled = True
+
+    @property
+    def cost(self) -> Int:
```
poncovka: The default value of the cost is 1000, so I would use the default value and avoid this DBus structure. See: https://dnf.readthedocs.io/en/latest/conf_ref.html#repo-options

You can still remember whether you are using a default or a user value internally in the module for the kickstart and dnf outputs.

pyanaconda/modules/common/structures/payload.py
```diff
@@ -0,0 +1,205 @@
...
+    @enabled.setter
+    def enabled(self, enabled: Bool):
+        """Set if the cost value should be used.
+
+        If True then change the default repository value.
+        """
+        self._enabled = enabled
+
+
+class PackagingConstrainsData(DBusData):
```
poncovka: I wonder whether we shouldn't have just one structure called `RepoConfiguration`. I know it looks pretty to have all these  classes, but there is not benefit for having them. It is much safer and easier to just set up one data object and call one DBus method. I would merge at least `URLData`, `PackagingConstrainsData`, `RepoCostData` and `SSLVerificationEnabled`.

pyanaconda/modules/common/structures/payload.py
jkonecny12: Fixed. Drawback of this is to that we have to have private API for payload to know that this value has changed.

pyanaconda/modules/common/structures/payload.py
jkonecny12: Fixed.

jkonecny12:
**UPDATED:**
* Completely changed all commits.
* New solution is using `RepoConfigurationData` and `SSLConfigurationData` structures for everything.

jkonecny12:
**UPDATED:**
* Conflicts from NFS source merge  has been resolved

pyanaconda/modules/payloads/source/url/url.py
```diff
@@ -0,0 +1,145 @@
...
+        self._url_source_id = self._generate_source_id()
+        self._set_repo_config_name()
+
+        self._install = False
+        self.install_changed = Signal()
+
+    @classmethod
+    def _generate_source_id(cls):
+        source_id = cls.REPO_NAME_ID
+        cls.REPO_NAME_ID = cls.REPO_NAME_ID + 1
```
poncovka, line 52: This will generate same names for any subclasses of `URLSourceModule` if we ever create such classes.

pyanaconda/modules/payloads/source/url/url.py
```diff
@@ -0,0 +1,145 @@
...
+        """
+        return self._repo_configuration
+
+    def set_repo_configuration(self, repo_configuration):
+        """Set repository configuration data.
+
+        :param repo_configuration: configuration for this repository
+        :type repo_configuration: RepoConfigurationData data structure
+        """
+        self._repo_configuration = repo_configuration
```
poncovka, line 103: Please, set the attribute after the validation.

pyanaconda/modules/payloads/source/url/url.py
```diff
@@ -0,0 +1,145 @@
...
+
+        self._validate_url(self._repo_configuration.type)
+        self._validate_proxy(self._repo_configuration.proxy)
+
+        if not self._repo_configuration.name:
+            self._set_repo_config_name()
+
+        self.repo_configuration_changed.emit(self._repo_configuration)
+        log.debug("The repo_configuration has changed")
```
poncovka, line 112: Please, log the configuration. You can simply write:
`log.debug("The repo configuration is set to: %s", repo_configuration)`

pyanaconda/modules/payloads/source/url/url.py
```diff
@@ -0,0 +1,145 @@
...
+            raise InvalidValueError("Proxy URL does not have valid format: {}".format(str(e)))
+
+    def _validate_url(self, url_type):
+        try:
+            URLType(url_type)
+        except ValueError:
+            raise InvalidValueError("Invalid source type set '{}'".format(url_type))
+
+    @property
+    def install(self):
```
poncovka, line 130: It sounds like a method. Maybe `installation_enabled`? Same for the DBus API.

pyanaconda/modules/payloads/source/url/url.py
jkonecny12: I'm not sure I understand you correctly. This is a class variable so it is shared between all subclasses with the same value. So if any subclass will generate new id it will increment the number for all the subclasses and parent class too?

Or are you talking about something else?

pyanaconda/modules/payloads/source/url/url.py
jkonecny12: Good point. Fixed.

pyanaconda/modules/payloads/source/url/url.py
jkonecny12: Fixed.

pyanaconda/modules/payloads/source/url/url.py
jkonecny12: Fixed by renaming to `install_repo_enabled`.

jkonecny12:
**UPDATED:**
* Resolve conflict.
* Implement notes from @poncovka .

jkonecny12:
jenkins, test this please

pyanaconda/modules/payloads/source/url/url.py
poncovka: Not exactly. If you try this script:

```
class X(object):
    id = 0

    @classmethod
    def get_id(cls):
        my_id = cls.id
        cls.id = cls.id + 1
        return my_id

class Y(X):
    pass

class Z(X):
    pass

for cls in [X, Y, Z, X, Y, Z, X, Y, Z]:
    print(cls.get_id())

```

The output is:
```
0
1
1
1
2
2
2
3
3
```

It should be a static method that will set up specifically `URLSourceModule.REPO_NAME_ID`. Also, please rename `REPO_NAME_ID` to `_repo_counter`. It is not a public constant.

pyanaconda/modules/payloads/source/url/url.py
```diff
@@ -0,0 +1,145 @@
...
+
+        self._install_repo_enabled = False
+        self.install_repo_enabled_changed = Signal()
+
+    @classmethod
+    def _generate_source_id(cls):
+        source_id = cls.REPO_NAME_ID
+        cls.REPO_NAME_ID = cls.REPO_NAME_ID + 1
+
+        return source_id
```
poncovka, line 54: Btw. I would generate here the full repo name. The source ID is not really interesting.

pyanaconda/modules/payloads/source/url/url.py
jkonecny12: Interesting, I've tried this and it worked. I wonder what I did wrong before...
Fixed.

pyanaconda/modules/payloads/source/url/url.py
jkonecny12: Fixed.

jkonecny12:
**UPDATED:**
* Implement notes from @poncovka .

PR #2406, opened 2020-03-26
subscription: SubscriptionRequest DBus structure

M4rtinK:
Add a SubscriptionRequest DBus structure, holding data needed
for a successful subscription attempt.

SubscriptionRequest can hold the following credentials:
- username & password
- org id & activation key/-s

As possibly both set of credentials could be set,
SubscriptionRequest has a type, that specifies which
credentials should be used for the subscription attempt.

SubscriptionRequest also holds the proxy configuration
and custom URL configuration (if any).

Red Hat account password, activation keys and HTTP proxy
password are all considered sensitive data, due to which
they are set-only. Once they are set to the Subscription
module, they will be stored internally and wiped in
SubscriptionRequest instances returned by the module.

It is still possible to check if a given sensitive value
has been set by inspecting the account_password_set,
activation_keys_set ans server_proxy_password_set properties.

Also included is the InsightsEnabled property that tracks
the intent to connect the target system to Red Hat Insights.

And of course, all this is covered by extensive unit tests.

### DIFF
--- a/pyanaconda/modules/common/structures/subscription.py
+++ b/pyanaconda/modules/common/structures/subscription.py
@@ -19,8 +19,11 @@
 from dasbus.structure import DBusData
 from dasbus.typing import *  # pylint: disable=wildcard-import
 
-__all__ = ["SystemPurposeData"]
+from pyanaconda.core.constants import DEFAULT_SUBSCRIPTION_REQUEST_TYPE
 
+from pyanaconda.modules.common.structures.secret import SecretData, SecretDataList
+
+__all__ = ["SystemPurposeData", "SubscriptionRequest"]
 
 class SystemPurposeData(DBusData):
     """System purpose data."""
@@ -78,3 +81,274 @@ class SystemPurposeData(DBusData):
     @addons.setter
     def addons(self, addons: List[Str]):
         self._addons = addons
+
+
+class SubscriptionRequest(DBusData):
+    """Data for a subscription request.
+
+    NOTE: Names of some of the fields are based on
+          how the given keys are called in rhsm.conf.
+    """
+
+    def __init__(self):
+        # subscription request type
+        # (based on authentication method used)
+        self._type = DEFAULT_SUBSCRIPTION_REQUEST_TYPE
+        # user identification
+        # - in case an account is member
+        #   of multiple organizations, both
+        #   organization and account username
+        #   need to be set
+        self._organization = ""
+        self._redhat_account_username = ""
+        # Candlepin instance
+        self._server_hostname = ""
+        # CDN base url
+        self._rhsm_baseurl = ""
+        # RHSM HTTP proxy
+        self._server_proxy_hostname = ""
+        self._server_proxy_port = -1
+        self._server_proxy_user = ""
+        # private data
+        # - we are using SecretData & SecretDataList
+        #   nested DBus structures to protect this
+        #   sensitive data
+        # - this way they can be set-only & easily
+        #   removed from SubscriptionRequest on
+        #   output from the Subscription module
+        # - they also support a robust way of clearing
+        #   previously set sensitive data if required
+        self._redhat_account_password = SecretData()
+        self._activation_keys = SecretDataList()
+        self._server_proxy_password = SecretData()
+
+    @property
+    def type(self) -> Str:
+        """Subscription request type.
+
+        Subscription request type is based on the authentication method used.
+
+        At the moment the following two are supported:
+        - username + password
+        - organization id + one or more activation keys
+
+        By default username + password is used.
+
+        Valid values are:
+        "username_password"
+        "org_activation_key"
+
+        :return: subscription request type
+        :rtype: str
+        """
+        return self._type
+
+    @type.setter
+    def type(self, request_type: Str):
+        self._type = request_type
+
+    @property
+    def organization(self) -> Str:
+        """Organization id for subscription purposes.
+
+        In most cases one of the following will be used:
+        - org id + one or more activation keys
+        - username + password
+
+        There is also a less often expected use case,
+        which applies if the same user account exists
+        in multiple organizations on the same Candlepin
+        instance. In such a case both username and
+        organization id needs to be set.
+
+        :return: organization id
+        :rtype: str
+        """
+        return self._organization
+
+    @organization.setter
+    def organization(self, organization: Str):
+        self._organization = organization
+
+    @property
+    def account_username(self) -> Str:
+        """Red Hat account username for subscription purposes.
+
+        In case the account for the given username is member
+        of multiple organizations, organization id needs to
+        be specified as well or else the registration attempt
+        will not be successful.
+
+        :return: Red Hat account username
+        :rtype: str
+        """
+        return self._redhat_account_username
+
+    @account_username.setter
+    def account_username(self, account_username: Str):
+        self._redhat_account_username = account_username
+
+    @property
+    def server_hostname(self) -> Str:
+        """Subscription server hostname.
+
+        This is basically a URL pointing to a Candlepin
+        instance to be used. It could be the one handling
+        general subscriptions hosted by Red Hat or one
+        embedded in a Satellite deployment.
+
+        If no custom server hostname is set, the default
+        value used by subscription manager will be used,
+        which is usually the URL pointing to the general
+        purpose Red Hat hosted Candlepin instance.
+
+        :return: Candlepin instance URL
+        :rtype: str
+        """
+        return self._server_hostname
+
+    @server_hostname.setter
+    def server_hostname(self, server_hostname: Str):
+        self._server_hostname = server_hostname
+
+    @property
+    def rhsm_baseurl(self) -> Str:
+        """CDN repository base URL.
+
+        Sets the base URL for the RHSM generated
+        repo file.
+
+        Setting this to a non default value only
+        makes sense if registering against Satellite
+        (as you would want to use the repos hosted
+        on the given Satellite instance) or possibly
+        during testing.
+
+        If no custom rhsm baseurl is set, the default
+        value used by subscription managed will be used,
+        which is generally baseurl for the Red Hat CDN.
+
+        :return: RHSM base url
+        :rtype: str
+        """
+        return self._rhsm_baseurl
+
+    @rhsm_baseurl.setter
+    def rhsm_baseurl(self, rhsm_baseurl: Str):
+        self._rhsm_baseurl = rhsm_baseurl
+
+    @property
+    def server_proxy_hostname(self) -> Str:
+        """RHSM HTTP proxy - hostname.
+
+        This is the hostname of the RHSM HTTP
+        proxy, which will be used for subscription
+        purposes only, eq. this will not configure
+        a system wide HTTP proxy.
+
+        :return: RHSM HTTP proxy hostname
+        :rtype: str
+        """
+        return self._server_proxy_hostname
+
+    @server_proxy_hostname.setter
+    def server_proxy_hostname(self, hostname: Str):
+        self._server_proxy_hostname = hostname
+
+    @property
+    def server_proxy_port(self) -> Int:
+        """RHSM HTTP proxy - port number.
+
+        -1 means port has not been set.
+
+        :returns: RHSM HTTP proxy port number
+        :rtype: int
+        """
+        return self._server_proxy_port
+
+    @server_proxy_port.setter
+    def server_proxy_port(self, port_number: Int):
+        self._server_proxy_port = port_number
+
+    @property
+    def server_proxy_user(self) -> Str:
+        """RHSM HTTP proxy - access username.
+
+        :return: RHSM HTTP proxy access username
+        :rtype: str
+        """
+        return self._server_proxy_user
+
+    @server_proxy_user.setter
+    def server_proxy_user(self, username: Str):
+        self._server_proxy_user = username
+
+    # private data
+    # - generally sensitive data such as passwords
+    #   or activation keys
+    # - these values should be "write only",
+    #   meaning data goes to the Subscription
+    #   module but can't be read out later one
+    #   via the public API
+    # - only the Subscription module should have
+    #   have access to these data internally &
+    #   use them appropriately (eq. register
+    #   the system, authenticate to HTTP proxy, etc.)
+    # - it should be also possible to explicitly
+    #   clear a previously set secret
+    # - to protect these values we are using
+    #   SecretData & SecretDataList, see their
+    #   implementation for more information
+
+    @property
+    def account_password(self) -> SecretData:
+        """Red Hat account password.
+
+        NOTE: This property is stored in SecretData
+              nested DBus structure to protect its contents.
+
+        :return: Red hat account password stored in a SecretData instance
+        :rtype: SecretData instance
+        """
+        return self._redhat_account_password
+
+    @account_password.setter
+    def account_password(self, password: SecretData):
+        if password:
+            self.account_password_set = True
+        self._redhat_account_password = password
+
+    @property
+    def activation_keys(self) -> SecretDataList:
+        """List of activation keys.
+
+        For a successful activation key based registration
+        at least one activation key needs to be set.
+
+        NOTE: This property is stored in SecretDataList
+              nested DBus structure to protect its contents.
+
+        :return: list of activation keys stored in SecretDataList instance
+        :rtype: SecretDataList instance
+        """
+        return self._activation_keys
+
+    @activation_keys.setter
+    def activation_keys(self, activation_keys: SecretDataList):
+        self._activation_keys = activation_keys
+
+    @property
+    def server_proxy_password(self) -> SecretData:
+        """RHSM HTTP proxy - access password.
+
+        NOTE: This property is stored in SecretData
+              nested DBus structure to protect its contents.
+
+        :return: RHSM HTTP proxy password stored in SecretData instance
+        :rtype: SecretData instance
+        """
+        return self._server_proxy_password
+
+    @server_proxy_password.setter
+    def server_proxy_password(self, password: SecretData):
+        self._server_proxy_password = password
--- a/pyanaconda/modules/subscription/subscription.py
+++ b/pyanaconda/modules/subscription/subscription.py
@@ -17,16 +17,24 @@
 # License and may only be used or replicated with the express permission of
 # Red Hat, Inc.
 #
-import os
+import copy
+import warnings
 
 from pyanaconda.core import util
 from pyanaconda.core.signal import Signal
+from pyanaconda.core.constants import SECRET_TYPE_HIDDEN, SUBSCRIPTION_REQUEST_TYPE_ORG_KEY, \
+    SUBSCRIPTION_REQUEST_VALID_TYPES
 
+from pyanaconda.modules.common.errors import InvalidValueError
 from pyanaconda.modules.common.base import KickstartService
-from pyanaconda.modules.common.structures.subscription import SystemPurposeData
+from pyanaconda.modules.common.structures.subscription import SystemPurposeData, \
+    SubscriptionRequest
+from pyanaconda.modules.common.structures.secret import get_public_copy
 
 from pyanaconda.modules.subscription import system_purpose
 from pyanaconda.modules.subscription.kickstart import SubscriptionKickstartSpecification
+
+from pykickstart.errors import KickstartParseWarning
 
 from pyanaconda.anaconda_loggers import get_module_logger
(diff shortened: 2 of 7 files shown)


M4rtinK:
Please let me know what you think about how the enum is being used for the SubscriptionRequest type. While it seems natural to me that an enum should be used for this, I'm still not sure if it is really worth it, given that we need to convert its values to string and back all the time due to DBus API requirements.

pyanaconda/modules/subscription/subscription.py
```diff
@@ -174,3 +240,80 @@ def set_system_purpose_data(self, system_purpose_data):
...
+        if subscription_request.server_proxy_password == "":
+            subscription_request.server_proxy_password = current_request.server_proxy_password
+
+        self._subscrition_request = subscription_request
+        self.subscription_request_changed.emit()
+        log.debug("A subscription request has been set.")
+
+    @property
+    def connect_to_insights(self):
+        """Indicates if the target system should be connected to red Hat Insights.
```
poncovka: Red Hat :-)

tests/nosetests/pyanaconda_tests/module_subscription_tests.py
```diff
@@ -219,6 +228,186 @@ def set_system_purpose_test(self):
...
+
+    def subscription_request_data_defaults_test(self):
+        """Test the SubscriptionRequest DBus structure defaults."""
+
+        # create empty SubscriptionRequest structure
+        empty_request = SubscriptionRequest()
+
+        # compare with expected default values
+        expected_default_dict = {
+            "type": get_variant(Str, str(DEFAULT_SUBSCRIPTION_REQUEST_TYPE.value)),
```
poncovka: This is not even correct. Please, always use constants when you set up the DBus structure.

pyanaconda/modules/subscription/subscription.py
```diff
@@ -174,3 +240,80 @@ def set_system_purpose_data(self, system_purpose_data):
...
+        # set private data indicators
+        subscription_request.account_password_set = bool(
+                subscription_request.account_password
+        )
+        subscription_request.activation_keys_set = bool(
+                subscription_request.activation_keys
+        )
+        subscription_request.server_proxy_password_set = bool(
+                subscription_request.server_proxy_password
+        )
```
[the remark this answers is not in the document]
M4rtinK: These properties will be used by the UI to check if a password/key has been set previously outside of the UI, via kickstart DBus or similar.

As for setting the value of the property in the DBus structure itself - I guess we could implement that & possibly drop the public setter for the _set properties hen we are at it.

But we still need to validate this when we set the structure to the module as well, otherwise in a case of a client not written in Python that does not have our SubscriptionRequest class available we can't really assure it will all be set correctly & should not blindly accept the data into the module.

pyanaconda/modules/common/structures/subscription.py
```diff
@@ -78,3 +84,326 @@ def addons(self) -> List[Str]:
...
+
+        :return: subscription request type
+        :rtype: str
+        """
+        return str(self._type.value)
+
+    @type.setter
+    def type(self, request_type: Str):
+        self._validate_type(request_type)
+        self._type = SubscriptionRequestType(request_type)
```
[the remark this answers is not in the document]
M4rtinK: Yeah, it really looked less good in the final form with all those conversions to string and back.
Let's scrap it and just use constants instead.

pyanaconda/modules/common/structures/subscription.py
```diff
@@ -16,11 +16,17 @@
...
 #
+from enum import Enum
+
 from dasbus.structure import DBusData
 from dasbus.typing import *  # pylint: disable=wildcard-import
 
-__all__ = ["SystemPurposeData"]
+from pyanaconda.modules.common.errors import InvalidValueError
+
+from pyanaconda.modules.subscription.constants import SubscriptionRequestType, \
```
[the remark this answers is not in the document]
M4rtinK: Replaced by constants in pyanaconda/modules/common/structures/subscription.py.

pyanaconda/modules/common/structures/subscription.py
```diff
@@ -78,3 +84,326 @@ def addons(self) -> List[Str]:
...
+        """Subscription request type.
+
+        Subscription request type is based on the authentication method used.
+
+        At the moment the following two are supported:
+        - username + password
+        - organization id + one or more activation keys
+
+        By default username + password is used.
+
```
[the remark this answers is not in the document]
M4rtinK: Added.

pyanaconda/modules/subscription/subscription.py
```diff
@@ -95,6 +117,50 @@ def process_kickstart(self, data):
...
+        # credentials
+        if data.rhsm.organization:
+            subscription_request.organization = data.rhsm.organization
+        if data.rhsm.activation_keys:
+            subscription_request.activation_keys = data.rhsm.activation_keys
+
+        # if org id and at least one activation key is set, switch authentication
+        # type to ORG & KEY
+        if data.rhsm.organization and data.rhsm.activation_keys:
+            subscription_request.type = str(SubscriptionRequestType.ORG_KEY.value)
```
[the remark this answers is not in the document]
M4rtinK: Switched to constant.

pyanaconda/modules/subscription/subscription.py
```diff
@@ -95,6 +117,50 @@ def process_kickstart(self, data):
...
+                # should not be fatal, but definitely logged as error
+                log.error("Failed to parse proxy for the rhsm command: %s", str(e))
+
+        # set the resulting subscription request
+        self.set_subscription_request(subscription_request)
+
+        # insights
+        self.set_connect_to_insights(bool(data.rhsm.connect_to_insights))
+
     def setup_kickstart(self, data):
```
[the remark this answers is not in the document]
M4rtinK: Good point, note added. :)

pyanaconda/modules/subscription/subscription.py
```diff
@@ -95,6 +117,50 @@ def process_kickstart(self, data):
...
+                    # ensure port is an integer and set to -1 if unknown
+                    port = int(proxy.port) if proxy.port else -1
+
+                    subscription_request.server_proxy_hostname = proxy.host
+                    subscription_request.server_proxy_port = port
+                    subscription_request.server_proxy_user = proxy.username
+                    subscription_request.server_proxy_password = proxy.password
+            except util.ProxyStringError as e:
+                # should not be fatal, but definitely logged as error
+                log.error("Failed to parse proxy for the rhsm command: %s", str(e))
```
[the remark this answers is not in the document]
M4rtinK: Sounds, like a good idea, did that.

M4rtinK:
jenkins, test this please

pyanaconda/modules/common/structures/subscription.py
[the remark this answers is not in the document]
M4rtinK: The Enum has been dropped.

pyanaconda/modules/subscription/subscription.py
```diff
@@ -174,3 +240,80 @@ def set_system_purpose_data(self, system_purpose_data):
...
+    def set_subscription_request(self, subscription_request):
+        """Set a subscription request.
+
+        Set the complete DBus structure containing subscription
+        request data.
+
+        :param subscription_request: subscription request structure to be set
+        :type subscription_request: DBus structure
+        """
+        current_request = copy.deepcopy(self._subscrition_request)
```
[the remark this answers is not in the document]
M4rtinK: Oh, right, we are replacing it anyway. Dropping the deepcopy.

pyanaconda/modules/subscription/subscription.py
```diff
@@ -174,3 +240,80 @@ def set_system_purpose_data(self, system_purpose_data):
...
+                subscription_request.activation_keys
+        )
+        subscription_request.server_proxy_password_set = bool(
+                subscription_request.server_proxy_password
+        )
+
+        # prevent currently set private data from being wiped if the
+        # new structure has them set to an empty string
+        if subscription_request.account_username == "":
+            subscription_request.account_username = current_request.account_username
```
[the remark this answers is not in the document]
M4rtinK: Indeed! Fixed.

pyanaconda/modules/subscription/subscription.py
```diff
@@ -174,3 +240,80 @@ def set_system_purpose_data(self, system_purpose_data):
...
+        if subscription_request.activation_keys == []:
+            subscription_request.activation_keys = current_request.activation_keys
+        if subscription_request.server_proxy_password == "":
+            subscription_request.server_proxy_password = current_request.server_proxy_password
```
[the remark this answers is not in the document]
M4rtinK: Expanded the docstring.

pyanaconda/modules/subscription/subscription.py
M4rtinK: Fixed! :)

tests/nosetests/pyanaconda_tests/module_subscription_tests.py
M4rtinK: Enum dropped.

pyanaconda/modules/subscription/subscription.py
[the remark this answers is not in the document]
jkonecny12: Shouldn't we generate comment in that case instead?

pyanaconda/modules/subscription/subscription.py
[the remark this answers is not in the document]
poncovka: I think that the design with the boolean property `<secret>_set` is not going to work in DBus structures, because the value of the property can be interpreted in different ways and confuse everyone. It should be more obvious what is going on.

I suggest to use a property `<secret>_type` that can be set to values `NONE`, `HIDDEN` and `TEXT`. The service will return `NONE`, when the secret is not set, and `HIDDEN`, when the secret is set, but its value is not provided in the property `<secret>`. The client will return `NONE`, when the secret should be cleared, `HIDDEN`, when the hidden value of the secret should be used, and `TEXT`, when the plain text value of the property `<secret>` should be used. This design could be used also in the Users module. Types can be extended with `ENCRYPTED`.

Then the structure could have a method called `set_<secret>` with an argument `secret`. If the argument `secret` is None, then the type will be set to `NONE` and the secret to an empty string. Otherwise, we will set the type to `TEXT` and set the secret. This will be useful for the client and the service.

The service could use a method called `hide_<secret>`, that will change the type of the secret from `TEXT` to `HIDDEN` and set the secret to an empty string.

Ideally, we should have a DBus structure called `SecretData` that can be shared by all modules that provide secrets. We support nested DBus structures, so you would have just a property called for example `server_proxy_password` that would return an instance of `SecretData`. The problem is with the list of activation keys. I will have to think about it a little more.

pyanaconda/modules/subscription/subscription.py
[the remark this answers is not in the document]
jkonecny12: I think this should be a method in the `SubscriptionRequest` in the same way as `get_public_copy()`.

pyanaconda/modules/subscription/subscription.py
```diff
@@ -174,3 +252,102 @@ def set_system_purpose_data(self, system_purpose_data):
...
+        return self._connect_to_insights
+
+    def set_connect_to_insights(self, connect):
+        """Set if the target system should be connected to Red Hat Insights.
+
+        :param bool connect: set to True to connect, set to False not to connect
+        """
+        self._connect_to_insights = connect
+        self.connect_to_insights_changed.emit()
+        log.debug("Connect target system to Insights set to: %s", self._connect_to_insights)
```
jkonecny12: This should have been in a separate commit. Please try to split the code to multiple commits if possible next time.

tests/nosetests/pyanaconda_tests/module_subscription_tests.py
```diff
@@ -161,6 +165,10 @@ def setUp(self):
         self.callback = PropertiesChangedCallback()
         self.subscription_interface.PropertiesChanged.connect(self.callback)
 
+        # some of the diffs might be long if a test fails, but we still
+        # want to see them
+        self.maxDiff = None
+
     def _test_dbus_property(self, *args, **kwargs):
```
jkonecny12: Please rename this to `_check_dbus_property`().

tests/nosetests/pyanaconda_tests/module_subscription_tests.py
```diff
@@ -219,6 +227,197 @@ def set_system_purpose_test(self):
...
+            "server-proxy-password-set": get_variant(Bool, True),
+        }
+
+        self._test_dbus_property(
+          "SubscriptionRequest",
+          subscription_request,
+          expected_subscription_request
+        )
+
+    def set_subscription_request_custom_urls_test(self):
```
jkonecny12: I think you should have at least one test using those data structures.

pyanaconda/modules/subscription/subscription.py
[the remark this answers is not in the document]
poncovka: https://github.com/rhinstaller/anaconda/pull/2412

pyanaconda/modules/subscription/subscription.py
```diff
@@ -174,3 +240,80 @@ def set_system_purpose_data(self, system_purpose_data):
...
+        if subscription_request.account_username == "":
+            subscription_request.account_username = current_request.account_username
+        if subscription_request.activation_keys == []:
+            subscription_request.activation_keys = current_request.activation_keys
+        if subscription_request.server_proxy_password == "":
+            subscription_request.server_proxy_password = current_request.server_proxy_password
+
+        self._subscrition_request = subscription_request
+        self.subscription_request_changed.emit()
+        log.debug("A subscription request has been set.")
```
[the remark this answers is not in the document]
M4rtinK: Good point, added! :)

pyanaconda/modules/subscription/subscription.py
M4rtinK: Given that the Insights hint is manipulated by the rhsm command via the `--connect-to-insights` parameter I think it makes sense to include it in the same commit that includes the SubscriptionRequest structure which is mainly filled by data from the same kickstart command. Same thing for tests that test that the rhsm command is parsed correctly.

tests/nosetests/pyanaconda_tests/module_subscription_tests.py
M4rtinK: I did that. Though, looking at our tests, it looks like we are using both at the moment. I guess we might want to "standardise" one one if them in the future ?

tests/nosetests/pyanaconda_tests/module_subscription_tests.py
M4rtinK: I'm not sure what you mean by "use" ? The tests check that data in the structures can be set & read out. Actual use will come in subsequent PRs when tasks and the GUI use the SubscriptionReqeust structure to process and exchange data.

pyanaconda/modules/subscription/subscription.py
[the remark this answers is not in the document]
M4rtinK: Sounds perfect, let's use that! :) And thanks a lot! :)

pyanaconda/modules/subscription/subscription.py
jkonecny12: It's not problem to add something dependent in separate commits if you won't broke the tests during the way.

tests/nosetests/pyanaconda_tests/module_subscription_tests.py
jkonecny12: Yep, it's copied from somewhere. I have to change this globally everywhere to avoid another wrong copy.

tests/nosetests/pyanaconda_tests/module_subscription_tests.py
jkonecny12: I mean create `SubscriptionRequest` and `SystemPurposeData` here and use them in the interface.

pyanaconda/modules/subscription/subscription.py
```diff
@@ -174,3 +254,99 @@ def set_system_purpose_data(self, system_purpose_data):
...
+                               (this will also switch the secret data
+                                instance to SECRET_TYPE_TEXT so that
+                                the Subscription module can read it
+                                internally)
+        - SECRET_TYPE_TEXT - this is new secret entry, we can keep it as is
+        """
+        current_request = self._subscription_request
+        # Red Hat account password
+        if new_request.account_password.type == SECRET_TYPE_HIDDEN:
+            new_request.account_password.set_secret(
```
poncovka, line 322: In this case, I would set `new_request.account_password = copy.deepcopy(current_request.account_password)`, because the hidden secret has no useful data. You shouldn't use `set_secret`, because it will set the secret to an empty string if the type of the current secret is `NONE`.

poncovka:
Hi, it is not possible to run the Subscription module. Please, add files `__init__.py`, `org.fedoraproject.Anaconda.Modules.Subscription.conf` and `org.fedoraproject.Anaconda.Modules.Subscription.service`, implement the `publish` method, fix the pylint issues and make sure it is possible to start the module in the installation environment.

M4rtinK:
@poncovka I have actually not tried running the module just yet, given how incomplete it still is. In any case a good point about the missing files, but I think this should be a separate PR.

pyanaconda/modules/common/structures/subscription.py
```diff
@@ -16,11 +16,31 @@
...
 
+SUBSCRIPTION_REQUEST_VALID_TYPES = {
+    SUBSCRIPTION_REQUEST_TYPE_USERNAME_PASSWORD,
+    SUBSCRIPTION_REQUEST_TYPE_ORG_KEY,
+}
+
+# Default authentication for subscription requests is
+# username password - this is basically to avoid the invalid
+# case of request not having a type set.
+DEFAULT_SUBSCRIPTION_REQUEST_TYPE = SUBSCRIPTION_REQUEST_TYPE_USERNAME_PASSWORD
```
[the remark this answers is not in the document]
M4rtinK: Yeah, looking at the other DBus structures, they do not really define their constants (if any) locally, so let's move it to the main constants file.

poncovka:
Btw. it is not just about the files. Pylint is ignoring `pyanaconda.modules.subscription` at this moment because of the missing `__init__` file.

pyanaconda/modules/subscription/subscription.py
M4rtinK: Good point, changed! :)

tests/nosetests/pyanaconda_tests/module_subscription_tests.py
M4rtinK: Oh, right - yeah, I can add a test like that. :)

M4rtinK:
@poncovka Opened #2430 with the missing bits.

pyanaconda/modules/common/structures/subscription.py
```diff
@@ -78,3 +81,281 @@ def addons(self) -> List[Str]:
...
+    @server_proxy_password.setter
+    def server_proxy_password(self, password: SecretData):
+        self._server_proxy_password = password
+
+    def __repr__(self):
+        """Provide a string representation of SubscriptionRequest."""
+        # we don't need to skip the sensitive data as the
+        # SecretData & SecretDataList classes handle that
+        # automatically
+        return generate_string_from_data(self)
```
poncovka, line 361: This is the default implementation. You can remove it.

pyanaconda/modules/subscription/subscription.py
```diff
@@ -174,3 +254,108 @@ def set_system_purpose_data(self, system_purpose_data):
...
+                                the Subscription module can read it
+                                internally)
+        - SECRET_TYPE_TEXT - this is new secret entry, we can keep it as is
+        """
+        current_request = self._subscription_request
+
+        # Red Hat account password
+        new_hidden = new_request.account_password.type == SECRET_TYPE_HIDDEN
+        current_set = current_request.account_password.type == SECRET_TYPE_TEXT
+        if new_hidden and current_set:
```
poncovka, line 324: I think you want to always replace the hidden values. If the new password is hidden and the current one is unset, the new password will stay hidden. It doesn't seem to be right.

pyanaconda/modules/common/structures/subscription.py
M4rtinK: Oh, cool, that's handy. :)

pyanaconda/modules/subscription/subscription.py
M4rtinK: Yup, always using old value if new one is SECRET_TYPE_HIDDEN is simpler and works. :)

PR #2407, opened 2020-03-27
Don't call the DBus method IsNodeFromIbft from the Storage module (#1…

poncovka:
…817529)

We cannot call this method from the main thread of the Storage module,
call a local function instead.

Resolves: rhbz#1817529

### DIFF
--- a/pyanaconda/bootloader/base.py
+++ b/pyanaconda/bootloader/base.py
@@ -22,6 +22,7 @@ from glob import glob
 import blivet
 from blivet.devices import NetworkStorageDevice
 from blivet.formats.disklabel import DiskLabel
+from blivet.iscsi import iscsi
 from blivet.size import Size
 from ordered_set import OrderedSet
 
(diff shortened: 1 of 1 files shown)


pyanaconda/bootloader/base.py
```diff
@@ -61,16 +62,27 @@ def _get_iscsi_node_from_device(device):
...
             return False
     return True
 
 
+def _is_node_from_ibft(node):
+    """Is the node configured from iBFT table?.
+
+    :param node: the node information
+    """
+    for ibft_node in iscsi.ibft_nodes:
```
rvykydal: Looking into blivet/iscsi.py it seems that the value of ibft_nodes is not set until iscsi.setup() is called, so maybe here it is not set?

poncovka: I guess that you mean `iscsi.startup()`. We call that during the initialization of the Storage module and this code is used only by the Storage module, so ibft_nodes should be already set when we call this function.

rvykydal: Ah, it is used only by the Storage module, sorry for the noise then...

poncovka:
@jkonecny12, the commit message was updated.

PR #2410, opened 2020-03-27
Improve tests speed

jkonecny12:
This will add parameter `-u` `--update` which will basically copy anaconda directory to temporary location in a mock and then rsync to the original location.

First version with removing only python files did not work because you can't run tests if you remove directories containing Makefiles. Makefile is that "smart" that it will check all these files first. :(


jkonecny12:
Closing this PR. Will be replaced by other with a correct name :D.

PR #2412, opened 2020-03-30
Add support for secrets in DBus structures

poncovka:
Use the DBus structure SecretData to represent a secret string. Use the DBus
structure SecretDataList to represent a secret list of strings. Call the
function hide_secrets to hide all secret fields of a data object. Call
the function get_public_copy to get a copy of a data object with hidden
secret fields.


jkonecny12:
Maybe thinking if secrets couldn't be solved by inheritance instead of types. Also it would be great to have an example with explanation in the code.

poncovka:
There is an issue with a string representation of nested data objects with secrets. The string shouldn't contain secret values. It should be fixed with https://github.com/rhinstaller/dasbus/pull/24.

poncovka:
Jenkins, test this please.

tests/nosetests/pyanaconda_tests/secret_data_test.py
```diff
@@ -0,0 +1,264 @@
...
+            self._b = SecretData()
+            self._c = SecretDataList()
+
+        @property
+        def a(self) -> Str:
+            return self._a
+
+        @a.setter
+        def a(self, value):
+            self._a = value
```
jkonecny12: Isn't it mandatory to provide type for setters too?

poncovka: No, it is not mandatory for DBus structures.

jkonecny12: Hmm, interesting. Thanks for confirmation.

pyanaconda/modules/common/structures/secret.py
```diff
@@ -0,0 +1,168 @@
...
+
+from pyanaconda.anaconda_loggers import get_module_logger
+from pyanaconda.core.constants import SECRET_TYPE_NONE, SECRET_TYPE_TEXT, SECRET_TYPE_HIDDEN
+
+log = get_module_logger(__name__)
+
+__all__ = ["get_public_copy", "hide_secrets", "SecretData", "SecretDataList"]
+
+
+def get_public_copy(obj: DBusData):
```
M4rtinK, line 33: So if I understand it correctly, I can call `get_public_copy()` on an arbitrary DBus structure & it will automatically handle any nested `SecretData` in the DBus structure, right ?

If this is correct, I think it should be mentioned in the docstring, as at the moment I don't think it is really clear this is possible (without reading the implementation) & that the method actually works on any DBus structure, not just SecretData instances.

poncovka:
Updated docstrings.

PR #2413, opened 2020-03-31
Handle translation of an empty string in widgets (#1815461)

poncovka:
Don't try to translate an empty string. It will return translation metadata.

Resolves: rhbz#1815461


poncovka:
Updated.

PR #2415, opened 2020-03-31
Rename _test_dbus_property to _check_dbus_property

jkonecny12:
First make it consistent in all tests. Second we should use test word only to define test case.


jkonecny12:
To avoid conflicts with master patches.

PR #1884, opened 2019-03-12
Make SMT warning message smaller (#1684056)

jkonecny12:
Related: rhbz#1684056

PR #1883, opened 2019-03-12
Remove blank line in the SMT warning (#1684056)

jkonecny12:
We are fighting with too long message so make it one line smaller this way.

*Related: rhbz#1684056*

PR #1878, opened 2019-03-08
Update the SMT warning message

rvykydal:
Related: rhbz#1684056

PR #1863, opened 2019-03-04
Detect and warn if SMT is enabled (#1684056)

poncovka:
Detect whether hyperthreading (SMT) is enabled on a system. If so,
we should display a warning in the info bar of the summary hub.

Resolves: rhbz#1684056

PR #2417, opened 2020-03-31
Detect and warn if SMT is enabled (#1684056)

poncovka:
Detect whether hyperthreading (SMT) is enabled on a system. If so,
we should display a warning in the info bar of the summary hub.

(cherry-picked from commits 30fdc00, 0fb8e42, 9a8485a, a4b7e1f)

Resolves: rhbz#1684056

### DIFF
--- a/data/anaconda.conf
+++ b/data/anaconda.conf
@@ -32,6 +32,9 @@ can_detect_unsupported_hardware = False
 # Should the installer show a warning about removed support for hardware?
 can_detect_support_removed = False
 
+# Should the installer show a warning about enabled SMT?
+can_detect_enabled_smt = False
+
 
 [Installation Target]
 # Type of the installation target.
--- a/pyanaconda/core/constants.py
+++ b/pyanaconda/core/constants.py
@@ -176,6 +176,25 @@ WARNING_NO_DISKS_SELECTED = N_(
     "No disks selected; please select at least one disk to install to."
 )
 
+# Kernel messages.
+WARNING_SMT_ENABLED_GUI = N_(
+    "Simultaneous Multithreading (SMT) technology can provide performance "
+    "improvements for certain workloads, but introduces several publicly "
+    "disclosed security issues. You have the option of disabling SMT, which "
+    "may impact performance. If you choose to leave SMT enabled, please read "
+    "https://red.ht/rhel-smt to understand your potential risks and learn "
+    "about other ways to mitigate these risks."
+)
+
+# This message is shorter to fit on the screen.
+WARNING_SMT_ENABLED_TUI = N_(
+    "Simultaneous Multithreading (SMT) may improve performance for certain "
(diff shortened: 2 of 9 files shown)


poncovka:
Backport from https://github.com/rhinstaller/anaconda/pull/1863, https://github.com/rhinstaller/anaconda/pull/1878, https://github.com/rhinstaller/anaconda/pull/1883, https://github.com/rhinstaller/anaconda/pull/1884.

poncovka:
Can we show this warning on Fedora or do we have to disable it (via the product configuration files)?

M4rtinK:
@poncovka I think the warning by itself is actually valid on Fedora as well. Still, I would recommend giving a public heads-up for this fedora-devel, even just due to the possibility of this breaking Fedora QAs graphical tests and there might be some good feedback why this is/is not a good idea. :)

pyanaconda/core/constants.py
```diff
@@ -176,6 +176,25 @@
...
+    "https://red.ht/rhel-smt to understand your potential risks and learn "
+    "about other ways to mitigate these risks."
+)
+
+# This message is shorter to fit on the screen.
+WARNING_SMT_ENABLED_TUI = N_(
+    "Simultaneous Multithreading (SMT) may improve performance for certain "
+    "workloads, but introduces several publicly disclosed security issues. "
+    "You can disable SMT, which may impact performance. Please read "
+    "https://red.ht/rhel-smt to understand potential risks and learn about "
```
M4rtinK, line 194: If we want to show this in Fedora pointing to access.redhat.com (which is where this shortened URL goes) might look a bit strange, but on the other hand I have not been able to find anything neutral that would be as authoritative.

jkonecny12: I'm not sure if we want to show this in Fedora at all. Isn't this just a RHEL decision?

rvykydal:
I'd add guard by product (via anaconda configuration) and disable on Fedora.

pyanaconda/core/constants.py
M4rtinK: Yeah, that in the end seems like a simpler solution.

poncovka:
Rebased and added a new commit with the configuration option.

data/anaconda.conf
```diff
@@ -32,6 +32,9 @@ can_detect_unsupported_hardware = False
 # Should the installer show a warning about removed support for hardware?
 can_detect_support_removed = False
 
+# Should the installer show a warning about enabled SMT?
+can_detect_enabled_smt = False
```
jkonecny12, line 36: I think it should be named ``show_warning_on_enabled_smt`` because the `can_detect...` does not explaining what will happen if we detect that.

poncovka: I am following the pattern of the options above. If we want to use a different pattern, we should rename all these options, but that is for another PR.

PR #2418, opened 2020-03-31
Fix text color on info bars

M4rtinK:
Due to some external changes - likely in GTK or the way
it processes CSS styles - our info bar text color overrides
no longer work.

Make the overrides more specific, so they match again and override
the text color on info bars as intended.

This should finally make the warnings Anaconda displays readable
again, as they no longer will be while on orange but black on orange
as intended.

Resolves: rhbz#1782174

(cherry picked from commit 39f8fcd9297b2db82b26546432d4bcee094b445b)

port from RHEL PR #2250

### DIFF
--- a/data/anaconda-gtk.css
+++ b/data/anaconda-gtk.css
@@ -46,34 +46,45 @@ levelbar.discrete trough block.filled.high {
 @define-color error_fg_color white;
 @define-color error_bg_color rgb (237, 54, 54);
 
-.info {
+infobar.info {
     background-color: @info_bg_color;
-    color: @info_fg_color;
     border-color: darker(@info_bg_color);
 }
 
-.warning {
+infobar.info label {
+    color: @info_fg_color;
+}
+
+infobar.warning {
     background-color: @warning_bg_color;
-    color: @warning_fg_color;
     border-color: darker(@warning_bg_color);
 }
 
-.question {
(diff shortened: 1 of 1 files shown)


M4rtinK:
For some reason this actually does not influence the infobar background color with the GTK (3.24.16-1) & theme in Rawhide. Surprisingly, the same code work with GTK (3.22.30-5) and the RHEL theme on RHEL8.

In short, with this patch or without it, in both cases the warning info bar is beige on Fedora:
infobar_warning_beige

On the other hand, as far as I can tell from the documentation, this is all correct.The `GTK.InfoBar` has as far as I can tell the name `infobar` in CSS:
https://lazka.github.io/pgi-docs/#Gtk-3.0/classes/InfoBar.html#class-details
And anything GTK should have the `background-color` CSS property:
https://developer.gnome.org/gtk3/stable/chap-css-properties.html
See "Table 10. Background properties". 

So I'm not really sure what's happening there, possibly other than the Fedora theme being more aggressive then the RHEL theme and overriding the `anaconda-gtk.css`.

But as we know this works correctly with RHEL 8 GTK, is correct according to the docs, looks nice code wise & does not cause any regressions on Fedora, I say we should merge this patch. Then we can look into making reproducers and possibly filling a bug on GTK if the issue is not caused by the Fedora theme.

PR #2419, opened 2020-04-01
Fix non-root dir of install tree HD installation (#1689194)

jkonecny12:
Installation didn't worked if the install tree was placed on the other place than / of the partition. This was because `method.dir` was not correctly propagated in this case.

In the past the only supported installation from HD was with ISO image. In that case you would use `method.dir` to find the ISO but path to the installation source was always root of the mounted image. This does not work for the install tree. Installation path have to be adjusted from the mount point to the specified path.

Backport of https://github.com/rhinstaller/anaconda/pull/1985.

*Resolves: rhbz#1689194*
(cherry picked from commit 559ad599790e8791b70a8cf0e95b002fb0035274)

### DIFF
--- a/pyanaconda/payload/dnf/payload.py
+++ b/pyanaconda/payload/dnf/payload.py
@@ -1369,8 +1369,7 @@ class DNFPayload(Payload):
             if not iso_device:
                 raise PayloadSetupError("device for HDISO install %s does not exist" % dev_spec)
 
-            self._setup_media(iso_device)
-            url = "file://" + INSTALL_TREE
+            url = self._setup_media(iso_device)
             self.install_device = iso_device
 
         return url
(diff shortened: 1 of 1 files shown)


jkonecny12:
jenkins, test this please

jkonecny12:
**UPDATED:**
* Fix error with not processing `harddrive --dir` with paths not starting with '/'.

jkonecny12:
jenkins, test this please

jkonecny12:
Tested with new KS tests.

jkonecny12:
**UPDATED:**
* Use ``join_paths`` to simplify solution

jkonecny12:
Tested with KS tests and it seems to be working correctly.

PR #2420, opened 2020-04-02
Fix support for string representations of DBus structures

poncovka:
Rewrite the function generate_request_description without the function
generate_dictionary_from_data that is going to be removed from dasbus.

Dasbus has fixed the bug with the unsafe string representation of nested DBus
structures. Update the test and change the version of dasbus.


poncovka:
The tests pass when run locally.

PR #2422, opened 2020-04-02
Fix the permission for changing a mount point (#1818500)

poncovka:
We shouldn't check whether the current format of a device is mountable, but
whether the requested format type is mountable. Otherwise, it might not be
possible to reformat a device.

We shouldn't clear errors in the callbacks on_page_clicked and
on_selector_clicked of the custom spoke. Otherwise, we might
clear the errors before the user sees them.

Resolves: rhbz#1818500


poncovka:
@jkonecny12, the commit message was updated.

PR #2423, opened 2020-04-02
Fix the encryption checkbox in the custom spoke (#1819360)

poncovka:
In the custom spoke, we shouldn't manually trigger the callback on_encrypt_toggled
that overrides the attribute device_encrypted of the device factory request. Call
the set_active method of the checkbox to change the value.

Show the checkbox for device encryption in an inconsistent state if the
device's container is encrypted. It will not trigger the on_encrypt_toggled
callback. Show the LUKS version if the checkbox is active or inconsistent.

Don't use the checkbox's method get_active for that, otherwise the callback
on_encrypt_toggled will to set the device encryption to an invalid value and
it will result in an unexpected partitioning.

Related: rhbz#1819360

### DIFF
--- a/pyanaconda/ui/gui/spokes/custom_storage.py
+++ b/pyanaconda/ui/gui/spokes/custom_storage.py
@@ -646,7 +646,7 @@ class CustomPartitioningSpoke(NormalSpoke, StorageCheckHandler):
             if data[0] == luks_version
         )
         self._luksCombo.set_active(idx)
-        self.on_encrypt_toggled(self._encryptCheckbox)
+        self._update_luks_combo()
 
     def _get_current_device_type(self):
         """ Return integer for type combo selection.
@@ -1253,7 +1253,7 @@ class CustomPartitioningSpoke(NormalSpoke, StorageCheckHandler):
                 self._encryptCheckbox.set_active(False)
 
             fancy_set_sensitive(self._encryptCheckbox, self._permissions.device_encrypted)
-            self.on_encrypt_toggled(self._encryptCheckbox)
+            self._update_luks_combo()
 
         # Update the UI.
         self._set_devices_label()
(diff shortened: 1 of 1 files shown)


poncovka:
Screenshot from 2020-04-02 21-31-54

pyanaconda/ui/gui/spokes/custom_storage.py
```diff
@@ -1509,7 +1508,10 @@ def on_fs_type_changed(self, combo):
     def on_encrypt_toggled(self, widget):
         self._request.device_encrypted = self._encryptCheckbox.get_active()
         self.on_luks_version_changed(self._luksCombo)
+        self._update_luks_combo()
+        self.on_value_changed()
```
jkonecny12, line 1510: Is it really required to have here `on_value_changed()` when this is already called by `on_luks_version_changed()`?

pyanaconda/ui/gui/spokes/custom_storage.py
```diff
@@ -1506,21 +1503,23 @@ def on_fs_type_changed(self, combo):
...
         self._update_luks_combo()
         self.on_value_changed()
 
     def _update_luks_combo(self):
-        hide_or_show = really_show if self._encryptCheckbox.get_active() else really_hide
-
-        for widget in [self._luksLabel, self._luksCombo]:
-            hide_or_show(widget)
+        visible = self._encryptCheckbox.get_active() or self._encryptCheckbox.get_inconsistent()
+        sensitive = self._encryptCheckbox.get_active()
```
jkonecny12, line 1514: Shouldn't this be instead?

```suggestion
        sensitive = self._encryptCheckbox.get_sensitive()
```

jkonecny12, line 1514: Move this line under `if visible:` part. It's not used elsewhere.

pyanaconda/ui/gui/spokes/custom_storage.py
poncovka: I don't really want to rely on that. Yes, this call is redundant at this moment, but it can change in the future.

pyanaconda/ui/gui/spokes/custom_storage.py
poncovka: No, it shouldn't. The LUKS combo box should be sensible only if the encryption checkbox is active. When you don't want to encrypt the device, you shouldn't be able to change the LUKS version.

poncovka: Thanks to the suggestion, but this looks more readable to me.

PR #2424, opened 2020-04-02
Allow to use an existing unlocked LUKS in one special case (#1772902)

poncovka:
In general, we don't allow to use an existing unlocked LUKS for the installation,
because Blivet is not able to set it up without a key and the whole installation
fails with an exception.

However, when the LUKS is unlocked with a different name than the standard
`luks-<UUID>`, Blivet will keep it open due to a bug in its code and the
installation will be successful. Let's allow this use case again and later
figure out how to allow unlocked LUKS devices in general.

Resolves: rhbz#1772902

### DIFF
--- a/pyanaconda/storage/checker.py
+++ b/pyanaconda/storage/checker.py
@@ -336,6 +336,7 @@ def verify_unlocked_devices_have_key(storage, constraints, report_error, report_
         and d.format.exists
         and not d.format.has_key
         and d.children
+        and any(c.name == d.format.map_name for c in d.children)
     ]
 
     for dev in devices:


poncovka:
Jenkins, test this please.

vtwaldo21:
three years later... i don't see this change actually working. i don't fully understand the change but kickstart still fails with the error and the map_name is "tcs-custom-pkgs".  but so is the child name, so it seems this check is basically a no-op? Or is there some way to force them to be different?  latest RHEL9 which has this patch.

the fix done over at https://linuxconfig.org/how-to-install-fedora-rhel-centos-via-kickstart-on-an-existing-luks-device does however work

vtwaldo21:
ok, well to answer my own question, I found that map_name in old versions of anaconda/blevit was that luks-{uuid} format, so this change actually did allow it to work. but some future change in blevit (haven't tracked down what yet), changes map_name to equal the <name> passed to cryptsetup, so this check is a no-op now and a better fix is required to overcome this bug of not allowing an unlocked LUKS partition.

VladimirSlavik:
@vojtechtrefny any thoughts about the above?

vojtechtrefny:
The blivet bug which this feature relied on was fixed in 3.5: https://github.com/storaged-project/blivet/pull/991

vtwaldo21:
Thanks for pointing that out @vojtechtrefny .  So this bug is therefore back, not sure what you all prefer, should a new ticket be opened or can this one be? Seems like a pretty big regression that was introduced.

Some more details of this bug are at are at https://bugzilla.redhat.com/show_bug.cgi?id=1839795

vtwaldo21:
changing my kickstart file and adding the option 
ignoredisk --drives=sde

I was able to work around this problem in my use case at least. No longer do I need to do all the steps over at

https://linuxconfig.org/how-to-install-fedora-rhel-centos-via-kickstart-on-an-existing-luks-device

PR #2425, opened 2020-04-03
Use constants to check the type of the payload object (#1820418)

poncovka:
Don't import the payload classes to check whether the payload object is an
instance of the specified class. Use the constants and the payload's property
called type. Otherwise, we might import some unrelated libraries that could
cause the installation to fail.

Resolves: rhbz#1820418


poncovka:
The issue was probably introduced in the commit https://github.com/rhinstaller/anaconda/commit/8a3c1182b34fd1514cca28c5d025691d40557da9. It looks like the class `PackagePayload` was safe to import, but the class `DNFPayload` isn't.

pyanaconda/ui/gui/spokes/storage.py
```diff
@@ -102,7 +102,7 @@ def _modify_sw_link_clicked(self, label, uri):
     def _get_sw_needs_text(self, required_space, sw_space, auto_swap):
         tooltip = _("Please wait... software metadata still loading.")
 
-        if isinstance(self.payload, LiveImagePayload):
+        if self.payload.type in (PAYLOAD_TYPE_LIVE_OS, PAYLOAD_TYPE_LIVE_IMAGE):
```
jkonecny12, line 105: I'm thinking there should be smarter solution than checking to these two.

One idea is to create a constant in CORE:
``PAYLOAD_TYPE_LIVE = (PAYLOAD_TYPE_LIVE_OS, PAYLOAD_TYPE_LIVE_IMAGE)``

VladimirSlavik: I almost agreed, but after reflecting on this, I think spelling out the multiple options is better. The test uses a different operator for multiple values. If you created a "merged types constant" then you would still have to use a different operator, but you could not tell easily why.

Actually - is there a chance this could get out of hand? Such as, testing for even more types? Two is fine, I guess, but more would be weird.

poncovka: Fixed with `PAYLOAD_LIVE_TYPES`.

poncovka:
Rebased and updated.

poncovka:
Tested with RPM OSTree, DNF and Live OS.

PR #2428, opened 2020-04-05
Update the dasbus-related code

poncovka:
Create a new instance of the error mapper, assign it to the message buses,
generate a new decorator for DBus errors and use it for the error registration.
Register all DBus errors right after the error mapper is created.

Remove the mapping to a DBus error for KickstartError and the Anaconda's
implementation of the event loop.

**Depends on:** https://github.com/rhinstaller/dasbus/pull/25


pyanaconda/modules/common/base/base.py
```diff
@@ -23,9 +23,9 @@
 from abc import ABC
 from locale import setlocale, LC_ALL
 
+from dasbus.loop import EventLoop
 from pykickstart.errors import KickstartError, KickstartParseWarning
 
-from pyanaconda.core.event_loop import EventLoop
```
jkonecny12, line 28: I don't think we should take EventLoop from a library. We are the main application the event loop should go opposite way. We should provide event loop to the library.

In case that we would have two libraries with their event loop we would end up with broken scenario because event loops can't be started twice.

poncovka: TODO: Use `create_main_loop` instead.

poncovka: Fixed.

poncovka:
Updated. The tests pass when run locally.

PR #2431, opened 2020-04-08
Fix bad import of general errors

jkonecny12:
This was introduced by merging PR checked on the old code-base.


jkonecny12:
**UPDATED:**
* There was one more in the `subscription.py`

PR #2432, opened 2020-04-08
Create product configuration files for Red Hat Virtualization

poncovka:
Create new files for products Red Hat Virtualization and oVirt Node Next.

(based on commits b497482 and 9a06e82)

### DIFF
--- a/data/anaconda.conf
+++ b/data/anaconda.conf
@@ -154,6 +154,16 @@ file_system_type =
 #
 default_partitioning = WORKSTATION
 
+# Default partitioning scheme.
+# Valid values:
+#
+#    PLAIN      Create standard partitions.
+#    BTRFS      Use the Btrfs scheme.
+#    LVM        Use the LVM scheme.
+#    LVM_THINP  Use LVM Thin Provisioning.
+#
+default_scheme = LVM
+
 # Default version of LUKS.
 # Valid values:
 #
--- a/data/anaconda.conf
+++ b/data/anaconda.conf
@@ -174,6 +174,64 @@ default_scheme = LVM
 luks_version = luks2
 
 
+[Storage Constraints]
+
+# Minimal size of the total memory.
+min_ram = 320 MiB
+
+# Minimal size of the available memory for LUKS2.
+luks2_min_ram = 128 MiB
+
+# Should we recommend to specify a swap partition?
+swap_is_recommended = True
+
+# Recommended minimal sizes of partitions.
+# Specify a mount point and a size on each line.
+min_partition_sizes =
+    /      250 MiB
+    /usr   250 MiB
+    /tmp   50  MiB
+    /var   384 MiB
+    /home  100 MiB
+    /boot  200 MiB
+
+# Required minimal sizes of partitions.
+# Specify a mount point and a size on each line.
+req_partition_sizes =
+
+# Allowed device types of the / partition if any.
+# Valid values:
+#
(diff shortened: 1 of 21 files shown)


pyanaconda/core/configuration/storage.py
```diff
@@ -87,6 +98,31 @@ def default_partitioning(self):
...
+
+    def _convert_partitioning_scheme(self, value):
+        """Convert the given value into a partitioning scheme."""
+        try:
+            member = PartitioningScheme.__members__[value]
+            return member.value
+        except KeyError:
+            pass
+
+        raise ValueError("'{}' is not a valid partitioning scheme".format(value))
```
jkonecny12, line 126: Please move this into the enum as method. I don't think we should access ``__member__`` private member out of the class.

data/anaconda.conf
```diff
@@ -174,6 +174,64 @@ default_scheme = LVM
...
+# Specify a list of mount points.
+must_not_be_on_root =
+
+# Mount points that are recommended to be reformatted.
+#
+# It will be recommended to create a new file system on a mount point
+# that has a whitelisted prefix, but doesn't have a blacklisted one.
+# Specify lists of mount points.
+reformat_whitelist = /boot /var /tmp /usr
+reformat_blacklist = /home /usr/local /opt /var/www
```
jkonecny12, line 232: What if something is not on whitelist neither blacklist? What then will happen?
Also why `/` is not specified here?

pyanaconda/core/configuration/storage_constraints.py
```diff
@@ -0,0 +1,159 @@
...
+    def must_be_on_linuxfs(self):
+        """Mount points that must be on a linux file system.
+
+        :return: a set of mount points
+        """
+        return set(self._get_option("must_be_on_linuxfs").split())
+
+    @property
+    def must_be_on_root(self):
+        """Mount points that must be on the / file system.
```
jkonecny12, line 125: Shouldn't it be:
``Mount points that must be on the / partition.``

IMHO FS is in most cases used for format used on the storage device but this is not what we are talking here.

pyanaconda/core/configuration/storage.py
poncovka: Fixed.

data/anaconda.conf
poncovka: > What if something is not on whitelist neither blacklist? What then will happen?

We will not show a message with the recommendation. Is it really so surprising?

> Also why `/` is not specified here?

We require to reformat the file system on `/`. It is a different check.

pyanaconda/core/configuration/storage_constraints.py
poncovka: Fixed.

poncovka:
Rebased and fixed.

pyanaconda/core/configuration/storage_constraints.py
VladimirSlavik: I know I'm late to the party, but is device possibly a good word here?

poncovka: I have changed it to:
`Paths that must be directories on the / file system.`

data/anaconda.conf
jkonecny12: OK, I guess this should be enough.

poncovka:
Rebased and tested.

PR #2433, opened 2020-04-09
subscription: Add service observer

M4rtinK:
Add ServiceObserver, which makes it possible to start
systemd unit backed DBus services on demand & to wait for
them to start up to a pre-defined timeout.

The timeout is set to 90 seconds by default, as this is also
the default timeout for systemd unit startup.

If the DBus service has not yet been requested, the corresponding
systemd unit will be started and get_proxy() will block up to
the timeout. If the DBus service becomes available before the timeout
DBus proxy for the service is cached and returned to the caller.

If the DBus service does not start in time a DBusObserverError is raised.

ServiceObserver is thread safe, so multiple threads can call
get_proxy() and safely wait for the DBus service to be available.

Once a proxy has been cached, subsequent get_proxy() calls should
not block ans will return the cached proxy.

On the other hand, if the service failed to start once, all subsequent
calls to get_proxy() will not block but will always raise the
DBusObserverError exception.

### DIFF
--- a/pyanaconda/core/constants.py
+++ b/pyanaconda/core/constants.py
@@ -450,3 +450,8 @@ SUBSCRIPTION_REQUEST_VALID_TYPES = {
 # username password - this is basically to avoid the invalid
 # case of request not having a type set.
 DEFAULT_SUBSCRIPTION_REQUEST_TYPE = SUBSCRIPTION_REQUEST_TYPE_USERNAME_PASSWORD
+
+# How long to wait for the RHSM service to become available after it is started.
+# - in seconds
+# - based on the default 90 second systemd service activation timeout
+RHSM_SERVICE_TIMEOUT = 90.0
--- a/pyanaconda/modules/common/constants/objects.py
+++ b/pyanaconda/modules/common/constants/objects.py
@@ -19,7 +19,7 @@
 from dasbus.identifier import DBusObjectIdentifier
 from pyanaconda.modules.common.constants.namespaces import STORAGE_NAMESPACE, NETWORK_NAMESPACE, \
     PARTITIONING_NAMESPACE, DEVICE_TREE_NAMESPACE, \
-    PAYLOADS_NAMESPACE
+    PAYLOADS_NAMESPACE, RHSM_NAMESPACE
 
 # Storage objects.
 
@@ -114,4 +114,13 @@ FIREWALL = DBusObjectIdentifier(
 PAYLOAD_PACKAGES = DBusObjectIdentifier(
     namespace=PAYLOADS_NAMESPACE,
     basename="Packages"
+)
+
+# System services
+
+# Subscription objects.
+
+RHSM_CONFIG = DBusObjectIdentifier(
+    namespace=RHSM_NAMESPACE,
+    basename="Config"
 )
--- a/pyanaconda/modules/subscription/initialization.py
+++ b/pyanaconda/modules/subscription/initialization.py
@@ -1,0 +1,104 @@
+#
+# Kickstart module for subscription handling.
+#
+# Copyright (C) 2020 Red Hat, Inc.
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# the GNU General Public License v.2, or (at your option) any later version.
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY expressed or implied, including the implied warranties of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General
+# Public License for more details.  You should have received a copy of the
+# GNU General Public License along with this program; if not, write to the
+# Free Software Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
+# 02110-1301, USA.  Any Red Hat trademarks that are incorporated in the
+# source code or documentation are not subject to the GNU General Public
+# License and may only be used or replicated with the express permission of
+# Red Hat, Inc.
+#
+
+from dasbus.typing import get_variant, Str
+
+from pyanaconda.core import util
+from pyanaconda.core.constants import RHSM_SERVICE_TIMEOUT
+
+from pyanaconda.threading import threadMgr
+
+from pyanaconda.modules.common.task import Task
+from pyanaconda.modules.common.errors.task import NoResultError
+from pyanaconda.modules.common.constants.services import RHSM
+from pyanaconda.modules.common.constants.objects import RHSM_CONFIG
+
+from pyanaconda.anaconda_loggers import get_module_logger
+log = get_module_logger(__name__)
+
+
+class StartRHSMTask(Task):
+    """Task for starting the RHSM DBus service."""
+
+    RHSM_SYSTEMD_UNIT_NAME = "rhsm.service"
+
+    @property
+    def name(self):
+        return "Start RHSM DBus service"
+
+    def run(self):
+        """Start the RHSM DBus service.
+
+        And also some related tasks, such as setting RHSM log levels.
+        """
+        # start the rhsm.service
+        # - this is blocking, but as we are effectively running in a thread
+        # it should not be an issue
+        # - if the return code is non-zero, return False immediately
+        rc = util.start_service(self.RHSM_SYSTEMD_UNIT_NAME)
+        if rc:
+            log.warning(
+                "subscription: RHSM systemd service failed to start with error code: %s",
+                rc
+            )
+            return False
+        # create a temporary proxy to set the log levels
+        rhsm_config_proxy = RHSM.get_proxy(RHSM_CONFIG)
+
+        # set RHSM log levels to debug
+        # - otherwise the RHSM log output is not usable for debugging subscription issues
+        log.debug("subscription: setting RHSM log level to DEBUG")
+        rhsm_config_proxy.Set("logging.default_log_level", get_variant(Str, "DEBUG"), "")
+        # all seems fine
+        log.debug("subscription: RHSM service start successfully.")
+        return True
+
+    def is_service_available(self, timeout=RHSM_SERVICE_TIMEOUT):
+        """Return if RHSM service is available or wait if startup is ongoing."""
+        if self.is_running:
+            # Wait up to defined timeout for the service to startup
+            # by joining the thread running the task. We specify a timeout when
+            # joining to prevent a deadlocked task blocking this method forever.
+            thread = threadMgr.get(self._thread_name)
+            if thread:
+                log.debug("subscription: waiting for RHSM service to start for up to %f seconds.",
+                          timeout)
+                thread.join(timeout)
+            else:
+                log.error("subscription: RHSM startup task is running but no thread found.")
+                return False
+
+        # now check again if the task is still running
+        if self.is_running:
+            # looks like we timed out
+            log.debug("subscription: RHSM service not available after waiting for %f seconds.",
+                      timeout)
+            return False
+        else:
+            # If we got this far, the task has finished running. If the result is True
+            # it was able to successfully start the systemd unit and connect to the DBus API.
+            # If the result is False, then the service failed to start.
+            try:
+                result = self.get_result()
+            except NoResultError:
+                # if the task fails in weird ways, there could apparently be no result
+                log.error("subscription: got no result from StartRHSMTask")
+                result = False
+            return result
--- a/pyanaconda/modules/subscription/rhsm_observer.py
+++ b/pyanaconda/modules/subscription/rhsm_observer.py
@@ -1,0 +1,60 @@
+#
+# Copyright (C) 2020  Red Hat, Inc.  All rights reserved.
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+from pyanaconda.modules.common.constants.services import RHSM
+from pyanaconda.core.constants import RHSM_SERVICE_TIMEOUT
+from pyanaconda.anaconda_loggers import get_module_logger
+
+from dasbus.client.observer import DBusObserver, DBusObserverError
+
+log = get_module_logger(__name__)
+
+
+class RHSMObserver(DBusObserver):
+    """Observer of the systemd unit backed RHSM DBus service."""
+
+    def __init__(self, startup_check_method, timeout=RHSM_SERVICE_TIMEOUT):
+        """Creates a RHSM service observer.
+
+        The observer will wait (up to a timeout) for the task starting the RHSM service
+        to finish and then provide access to the corresponding RHSM DBus API.
+
+        The waiting is only expected to happen at Anaconda startup, once the RHSM
+        startup task has finished running no further waiting is expected.
+
+        :param startup_check_method: a blocking method that waits for RHSM
+                                     startup to finish up to a given timeout
+                                     and then returns a boolean value corresponding
+                                     to True if startup was successful or False
+                                     otherwise
+        :param float timeout: how long to wait for RHSM service to start in seconds
+        """
+        super().__init__(RHSM._message_bus, RHSM.service_name)
+        self._startup_check_method = startup_check_method
+        self._timeout = timeout
+
+    def get_proxy(self, object_identifier):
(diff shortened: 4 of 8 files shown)


poncovka:
I think that you should split the code into two classes.

There should be a task called for example `StartRHSMTask`. The task should start the RHSM service with `util.start_service`, create a private proxy, set the logging level and restart the service with `util.restart_service`. It could return False if one of the steps fails, otherwise True.  The task should be started in the `run` method of the `SubscriptionService` (before the call to `super()`).

Then there could an observer called for example `RHSMObserver`. The observer shouldn't use signals of the task, because emitting of these signals is scheduled in the main loop. However, you can use methods `is_running`, `get_result` and the name of the thread. When you want to create a proxy of the RHSM service, but `is_service_available` returns False and the task is still running, you can wait with a timeout for the task to finish and then set `_is_service_available` to the result of the task. If the service is still not available, you can raise an exception. I suggest to implement this logic in the task. It could be in a method `is_service_available(timeout)`. Then the observer needs access only to this method.

pyanaconda/modules/subscription/service_observer.py
```diff
@@ -0,0 +1,117 @@
...
+        :param float timeout: how long to wait for the DBus service to start in seconds
+        """
+        super().__init__(message_bus, dbus_service_name)
+        self._message_bus = message_bus
+        self._dbus_service_name = dbus_service_name
+        self._systemd_unit_name = systemd_unit_name
+        self._timeout = timeout
+
+        namespace = get_namespace_from_name(dbus_service_name)
+        self._object_path = get_dbus_path(*namespace)
```
poncovka, line 57: How do you plan to access other DBus objects that belong to the RHSM service?

pyanaconda/modules/subscription/service_observer.py
```diff
@@ -0,0 +1,117 @@
...
+    def get_proxy(self):
+        """Returns a proxy of the remote object."""
+        with self._proxy_lock:
+            # if there is a proxy, return it (this also releases the lock)
+            if self._proxy:
+                return self._proxy
+            # check if service startup is in progress
+            elif not self._timer:
+                # looks like service was not yet started, so start it's systemd unit
+                util.start_service(self._systemd_unit_name)
```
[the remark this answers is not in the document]
M4rtinK: Good point, we should definitely use that. Looking at start service, it does not yet support adding that flag but it should be easy to add.

pyanaconda/modules/subscription/service_observer.py
```diff
@@ -0,0 +1,117 @@
...
+                # start a timer that will mark service startup as failed if it is
+                # not interrupted by successful service activation
+                self._timer = Timer(self._timeout, self._service_timed_out)
+                self._timer.start()
+                # connect to the service_available signal of the Observer and
+                # stop the timer once the DBus service becomes accessible
+                self.service_available.connect(self._service_activated)
```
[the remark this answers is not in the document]
M4rtinK: I was a bit afraid the signal might not work out, still decided to use it for the first draft as this way I could avoid any explicit polling. But given that this is not really performance critical, some polling should not be a problem.

pyanaconda/modules/subscription/service_observer.py
M4rtinK: Good point, I'll need to look more into that as we do use a bunch of them.

M4rtinK:
> I think that you should split the code into two classes.
> 
> There should be a task called for example `StartRHSMTask`. The task should start the RHSM service with `util.start_service`, create a private proxy, set the logging level and restart the service with `util.restart_service`. It could return False if one of the steps fails, otherwise True. The task should be started in the `run` method of the `SubscriptionService` (before the call to `super()`).
So this would be unconditional & would delay the Subscription module startup unit all this is done, right ? Well, I think that's not a problem and much simpler than some lazy startup of the RHSM service that would be done only once someone tries to call it. So let's do it that way. :)

> 
> Then there could an observer called for example `RHSMObserver`. The observer shouldn't use signals of the task, because emitting of these signals is scheduled in the main loop. However, you can use methods `is_running`, `get_result` and the name of the thread. When you want to create a proxy of the RHSM service, but `is_service_available` returns False and the task is still running, you can wait with a timeout for the task to finish and then set `_is_service_available` to the result of the task. If the service is still not available, you can raise an exception. I suggest to implement this logic in the task. It could be in a method `is_service_available(timeout)`. Then the observer needs access only to this method.
Sounds good, nice & simple. Let's do that as well. :)

poncovka:
> > There should be a task called for example `StartRHSMTask`. The task should start the RHSM service with `util.start_service`, create a private proxy, set the logging level and restart the service with `util.restart_service`. It could return False if one of the steps fails, otherwise True. The task should be started in the `run` method of the `SubscriptionService` (before the call to `super()`).

> So this would be unconditional & would delay the Subscription module startup unit all this is done, right ? Well, I think that's not a problem and much simpler than some lazy startup of the RHSM service that would be done only once someone tries to call it. So let's do it that way. :)

It would be unconditional, but it should be still safe for all types of products and environments. For example, the task should be able to handle a missing or failing service.

The task cannot delay the start of the Subscription module, because we could get another DBus timeout. The task should be started with the `start` method to run in a separate thread. The benefit would be that we will try to start the service as soon as possible, but we will not block the module or the event loop until it is necessary.

M4rtinK:
BTW, discussed the need for restarting RHSM service for log levels to take effect with the RHSM team and it looks like the bug causing this has been fixed, so we should be able to drop that workaround. :)

M4rtinK:
@poncovka Updated the PR based on your suggestions. (At the moment it is rebased on top of the insights task PR).

pyanaconda/modules/common/constants/objects.py
```diff
@@ -115,3 +116,12 @@
     namespace=PAYLOADS_NAMESPACE,
     basename="Packages"
 )
+
+# System services
+
+# Subscription objects.
+
+RHSM_CONFIG = DBusObjectIdentifier(
+    namespace=RHSM.namespace,
```
poncovka, line 125: Please, use `RHSM_NAMESPACE`.

pyanaconda/modules/subscription/initialization.py
```diff
@@ -0,0 +1,90 @@
...
+# Free Software Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
+# 02110-1301, USA.  Any Red Hat trademarks that are incorporated in the
+# source code or documentation are not subject to the GNU General Public
+# License and may only be used or replicated with the express permission of
+# Red Hat, Inc.
+#
+
+RHSM_SYSTEMD_UNIT_NAME = "rhsm.service"
+
+SERVICE_CHECK_INTERVAL = 0.1  # how often to check in s
```
poncovka, line 23: I would move these constants to the task as class members.

pyanaconda/modules/subscription/initialization.py
```diff
@@ -0,0 +1,90 @@
...
+
+        And also some related tasks, such as setting RHSM log levels.
+        """
+        # record when the task was started
+        self._start_time = time.time()
+
+        # start the rhsm.service
+        # - this is blocking, but as we are effectively running in a thread
+        # it should not be an issue
+        # - if the return code is non-zero, raise an exception
```
poncovka, line 60: We don't raise an exception.

pyanaconda/modules/subscription/initialization.py
```diff
@@ -0,0 +1,90 @@
...
+        rc = util.start_service(RHSM_SYSTEMD_UNIT_NAME)
+        if rc:
+            return False
```
poncovka, line 63: Please, log here that the service couldn't be started.

pyanaconda/modules/subscription/initialization.py
```diff
@@ -0,0 +1,90 @@
...
+        # create a temporary proxy to set the log levels
+        rhsm_config_proxy = RHSM.get_proxy(RHSM_CONFIG)
+
+        # set RHSM log levels to debug
+        # - otherwise the RHSM log output is not usable for debugging subscription issues
+        log.debug("subscription: setting RHSM log level to DEBUG")
+        rhsm_config_proxy.Set("logging.default_log_level", get_variant(Str, "DEBUG"), "")
+        # all seems fine
+        return True
```
poncovka, line 72: Please, log here that the service was successfully started.

pyanaconda/modules/subscription/initialization.py
```diff
@@ -0,0 +1,90 @@
...
+
+    def is_service_available(self, timeout, check_interval=SERVICE_CHECK_INTERVAL):
+        """Return if RHSM service is available or wait if startup is ongoing."""
+        while self.is_running:
```
poncovka, line 76: Active waiting is not necessary. You can get the running thread with `thread = threadMgr.get(self._thread_name)` and wait with a timeout with `thread.join(timeout)`.

See: https://docs.python.org/3.9/library/threading.html#threading.Thread.join

pyanaconda/modules/subscription/initialization.py
```diff
@@ -0,0 +1,90 @@
...
+                # This means the task is still running even after exceeding
+                # the define timeout and is likely stuck. We can't expect
+                # the RHSM DBus API be available in such a state, so return False,
+                # which will also terminate the while cycle.
+                log.debug("RHSM service failed to start in %f seconds", timeout)
+                return False
+        # If we got this far, the task has finished running. If the result is True
+        # it was able to successfully start the systemd unit and connect to the DBus API.
+        # If the result is False, then the service failed to start.
+        return self.get_result()
```
poncovka, line 90: I would handle the exception `NoResultError` just to be sure.

pyanaconda/modules/subscription/initialization.py
```diff
@@ -0,0 +1,90 @@
...
+        rhsm_config_proxy = RHSM.get_proxy(RHSM_CONFIG)
+
+        # set RHSM log levels to debug
+        # - otherwise the RHSM log output is not usable for debugging subscription issues
+        log.debug("subscription: setting RHSM log level to DEBUG")
+        rhsm_config_proxy.Set("logging.default_log_level", get_variant(Str, "DEBUG"), "")
+        # all seems fine
+        return True
+
+    def is_service_available(self, timeout, check_interval=SERVICE_CHECK_INTERVAL):
```
poncovka, line 74: I would provide a default timeout.

pyanaconda/modules/subscription/rhsm_observer.py
```diff
@@ -0,0 +1,73 @@
...
+        :raises: DBusObserverError if the given proxy can't be returned
+        """
+        can_return_proxy = False
+        # first check if the DBus API seems to be up
+        if self.is_service_available:
+            can_return_proxy = True
+        else:
+            log.debug("rhsm-observer: API not yet available, checking startup task state")
+            # the RHSM startup task is either still running or has already failed
+            can_return_proxy = self._startup_check_method(self._timeout)
```
poncovka, line 68: You don't need the variable `can_return_proxy`. If `is_service_available` is False and `_startup_check_method` returns False, you can raise the exception. Otherwise, you can create the proxy.

pyanaconda/modules/subscription/rhsm_observer.py
```diff
@@ -0,0 +1,73 @@
...
+                                     and then returns a boolean value corresponding
+                                     to True if startup was successful or False
+                                     otherwise
+        :param float timeout: how long to wait for RHSM service to start in seconds
+        """
+        super().__init__(message_bus, RHSM.service_name)
+        self._startup_check_method = startup_check_method
+        self._timeout = timeout
+
+    def get_proxy(self, object_identifier):
```
poncovka, line 54: The argument `object_identifier` could be None by default, if you need to be able to access also `RHSM.get_proxy()`.

pyanaconda/modules/subscription/rhsm_observer.py
```diff
@@ -0,0 +1,73 @@
...
+        startup task has finished running no further waiting is expected.
+
+        :param startup_check_method: a blocking method that waits for RHSM
+                                     startup to finish up to a given timeout
+                                     and then returns a boolean value corresponding
+                                     to True if startup was successful or False
+                                     otherwise
+        :param float timeout: how long to wait for RHSM service to start in seconds
+        """
+        super().__init__(message_bus, RHSM.service_name)
```
poncovka, line 50: I would use `RHSM._message_bus` for now. It looks like it would be useful to make the message bus public.

pyanaconda/modules/common/constants/objects.py
M4rtinK: OK, will do - I used this as we at the moment don't have any name spaces defined for non-Anaconda DBus services.

pyanaconda/modules/subscription/initialization.py
M4rtinK: Moved. :)

pyanaconda/modules/subscription/initialization.py
M4rtinK: Indeed, a leftover from a previous revision. Removed.

pyanaconda/modules/subscription/initialization.py
M4rtinK: Good point, added.

pyanaconda/modules/subscription/initialization.py
M4rtinK: Added. :)

pyanaconda/modules/subscription/initialization.py
M4rtinK: Oh, right - that'c clever. :)

BTW, I wonder if we might want to implement a mechanism like this in the Task itself in the future ? Seems useful.

pyanaconda/modules/subscription/initialization.py
M4rtinK: Added.

pyanaconda/modules/subscription/initialization.py
M4rtinK: Added.

pyanaconda/modules/subscription/rhsm_observer.py
M4rtinK: Good point, let's do that.

pyanaconda/modules/subscription/rhsm_observer.py
M4rtinK: I've checked the code on RHEL-8 and it looks like we don't do that, so I did I like this.

pyanaconda/modules/subscription/rhsm_observer.py
M4rtinK: Sounds good, changed.

M4rtinK:
Rebased on `master`.

pyanaconda/core/constants.py
```diff
@@ -450,3 +450,8 @@ class PayloadRequirementType(Enum):
 # username password - this is basically to avoid the invalid
 # case of request not having a type set.
 DEFAULT_SUBSCRIPTION_REQUEST_TYPE = SUBSCRIPTION_REQUEST_TYPE_USERNAME_PASSWORD
+
+# How long to wait for the RHSM service to become available after is is started.
```
poncovka, line 454: `it is ...`

pyanaconda/modules/subscription/initialization.py
```diff
@@ -0,0 +1,98 @@
...
+        log.debug("subscription: RHSM service start successfully.")
+        return True
+
+    def is_service_available(self, timeout=RHSM_SERVICE_TIMEOUT):
+        """Return if RHSM service is available or wait if startup is ongoing."""
+        if self.is_running:
+            # Wait up to defined timeout for the service to startup
+            # by joining the thread running the task. We specify a timeout when
+            # joining to prevent a deadlocked task blocking this method forever.
+            thread = threadMgr.get(self._thread_name)
```
poncovka, line 79: This can return None.

pyanaconda/modules/subscription/rhsm_observer.py
```diff
@@ -0,0 +1,62 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+from pyanaconda.modules.common.constants.services import RHSM
+
+from pyanaconda.core.constants import RHSM_SERVICE_TIMEOUT
+
```
poncovka, line 20: Please, don't add new lines between imports from the same packages.

pyanaconda/core/constants.py
M4rtinK: Fixed.

pyanaconda/modules/subscription/initialization.py
M4rtinK: Adding a check for the return value, that logs an error & returns False if thread is None.

pyanaconda/modules/subscription/rhsm_observer.py
M4rtinK: New lines removed.

M4rtinK:
@poncovka Thanks for the final review! :)

M4rtinK:
jenkins, test this please

PR #2434, opened 2020-04-09
Make the kickstart support for the btrfs command optional

poncovka:
Raise the exception KickstartParseError if the btrfs command is specified
in a kickstart file, but btrfs is not a supported file system type.

(based on a commit 3b28b3e)


rvykydal:
No objections so far.

poncovka:
Reimplemented and tested.

PR #2435, opened 2020-04-09
Fix spokes ordering and add tests for priority uniqueness

jkonecny12:
~The "KernelWarningSpoke" and "UnsupportedHardwareSpoke" seems to be switched. Not sure if this is happening only sometimes or all the time but sorting the spokes in the test should solve this for good.~

~Probably introduced by: 0aa6f1507a8b4bacebca7dcccf7bda4b873e6fba~

~**EDIT:**~
~If two standalone spokes have the same priority then we depending on implementation of sorted algorithm which can differ between architectures.~

**EDIT2:**
See https://github.com/rhinstaller/anaconda/pull/2435#issuecomment-614684320 for explanation.

This already happened in the past see:
https://bugzilla.redhat.com/show_bug.cgi?id=929177

### DIFF
--- a/pyanaconda/ui/__init__.py
+++ b/pyanaconda/ui/__init__.py
@@ -136,7 +136,8 @@ class UserInterface(object):
         """
         raise NotImplementedError
 
-    def _collectActionClasses(self, module_pattern_w_path, standalone_class):
+    @staticmethod
+    def _collectActionClasses(module_pattern_w_path, standalone_class):
         """Collect all the Hub and Spoke classes which should be enqueued for processing.
 
         :param module_pattern_w_path: the full name patterns (pyanaconda.ui.gui.spokes.%s)
@@ -165,7 +166,8 @@ class UserInterface(object):
 
         return standalones
 
-    def _orderActionClasses(self, spokes, hubs):
+    @staticmethod
+    def _orderActionClasses(spokes, hubs):
         """Order all the Hub and Spoke classes.
 
         These should be enqueued for processing according to their pre/post dependencies.
@@ -183,13 +185,13 @@ class UserInterface(object):
         for hub in hubs:
             action_classes.extend(
                 sorted(
-                    self._filter_spokes_by_pre_for_hub_reference(spokes, hub),
+                    UserInterface._filter_spokes_by_pre_for_hub_reference(spokes, hub),
                     key=lambda obj: obj.priority)
             )
             action_classes.append(hub)
             action_classes.extend(
                 sorted(
-                    self._filter_spokes_by_post_for_hub_reference(spokes, hub),
+                    UserInterface._filter_spokes_by_post_for_hub_reference(spokes, hub),
                     key=lambda obj: obj.priority)
             )
 
--- a/pyanaconda/ui/__init__.py
+++ b/pyanaconda/ui/__init__.py
@@ -75,7 +75,7 @@ class UserInterface(object):
 
     @classmethod
     def update_paths(cls, pathdict):
-        """Receives pathdict and appends it's contents to the current class defined search path dictionary."""
+        """Receives path dict and appends it's contents to the class defined search path."""
         for k, v in pathdict.items():
             cls.paths.setdefault(k, [])
             cls.paths[k].extend(v)
@@ -91,18 +91,23 @@ class UserInterface(object):
         """Run the interface.
 
         This should do little more than just pass through to something else's run method,
-        but is provided here in case more is needed.  This method must be provided by all subclasses.
+        but is provided here in case more is needed. This method must be provided
+        by all subclasses.
         """
         raise NotImplementedError
 
     @property
     def meh_interface(self):
-        """Returns an interface for exception handling (defined by python-meh's AbstractIntf class)."""
+        """Returns an interface for exception handling.
+
+        Defined by python-meh's AbstractIntf class.
+        """
         raise NotImplementedError
 
     ###
-    ### MESSAGE HANDLING METHODS
+    # MESSAGE HANDLING METHODS
     ###
+
     def showError(self, message):
         """Display an error dialog with the given message.
 
@@ -146,14 +151,24 @@ class UserInterface(object):
         """
         standalones = []
 
+        def check_standalone_spokes(obj):
+            return issubclass(obj, standalone_class) and \
+                getattr(obj, "preForHub", False) or \
+                getattr(obj, "postForHub", False)
+
         for module_pattern, path in module_pattern_w_path:
-            standalones.extend(collect(module_pattern, path, lambda obj: issubclass(obj, standalone_class) and \
-                                       getattr(obj, "preForHub", False) or getattr(obj, "postForHub", False)))
+            standalones.extend(
+                collect(module_pattern,
+                        path,
+                        check_standalone_spokes)
+            )
 
         return standalones
 
     def _orderActionClasses(self, spokes, hubs):
-        """Order all the Hub and Spoke classes which should be enqueued for processing according to their pre/post dependencies.
+        """Order all the Hub and Spoke classes.
+
+        These should be enqueued for processing according to their pre/post dependencies.
 
         :param spokes: the classes we are to about order according
                        to the hub dependencies
@@ -163,13 +178,27 @@ class UserInterface(object):
                      attribute of Spokes to pick up
         :type hubs: common.Hub based types
         """
+        action_classes = []
 
-        actionClasses = []
         for hub in hubs:
-            actionClasses.extend(sorted(filter(lambda obj, h=hub: getattr(obj, "preForHub", None) == h, spokes),
-                                        key=lambda obj: obj.priority))
-            actionClasses.append(hub)
-            actionClasses.extend(sorted(filter(lambda obj, h=hub: getattr(obj, "postForHub", None) == h, spokes),
-                                        key=lambda obj: obj.priority))
+            action_classes.extend(
+                sorted(
+                    self._filter_spokes_by_pre_for_hub_reference(spokes, hub),
+                    key=lambda obj: obj.priority)
+            )
+            action_classes.append(hub)
+            action_classes.extend(
+                sorted(
+                    self._filter_spokes_by_post_for_hub_reference(spokes, hub),
+                    key=lambda obj: obj.priority)
+            )
 
-        return actionClasses
+        return action_classes
+
+    @staticmethod
+    def _filter_spokes_by_pre_for_hub_reference(spokes, hub):
+        return filter(lambda obj, h=hub: getattr(obj, "preForHub", None) == h, spokes)
+
+    @staticmethod
+    def _filter_spokes_by_post_for_hub_reference(spokes, hub):
+        return filter(lambda obj, h=hub: getattr(obj, "postForHub", None) == h, spokes)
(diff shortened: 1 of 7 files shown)


jkonecny12:
jenkins, test this please

jkonecny12:
**UPDATED:**
* Or better changed the implementation. Remove the old solution and change the code with a adding new test.

tests/nosetests/pyanaconda_tests/simple_ui_test.py
```diff
@@ -156,3 +197,30 @@ def gui_test(self, window_group, window, meh, builder, proxy_getter):
...
+        If we have the same priority for two standalone spokes which anchor to the same hub
+        then we are depending on sorted function implementation which can differ between
+        architectures and python versions.
+
+        This already happened in the past:
+        https://bugzilla.redhat.com/show_bug.cgi?id=929177
+        """
+        # Create the interface.
+        from pyanaconda.ui.gui import GraphicalUserInterface
+        self.interface = GraphicalUserInterface(self.storage, self.payload)
```
poncovka, line 218: Please, extend the existing tests. Don't copy-paste them.

pyanaconda/ui/common.py
```diff
@@ -480,6 +480,8 @@ class StandaloneSpoke(Spoke):
     preForHub = None
     postForHub = None
 
+    priority = 100
```
poncovka, line 483: It looks like the default priority should be zero.

pyanaconda/ui/gui/spokes/network.py
```diff
@@ -1578,7 +1578,6 @@ class NetworkStandaloneSpoke(StandaloneSpoke):
     uiFile = "spokes/network.glade"
 
     preForHub = SummaryHub
-    priority = 10
```
poncovka, line 1581: Please, don't change the priorities when it is not necessary. There might be addons that rely on these values.

tests/nosetests/pyanaconda_tests/simple_ui_test.py
```diff
@@ -56,6 +56,10 @@ def hubs(self):
     def action_classes(self):
         return self.interface._collectActionClasses(self.paths["spokes"], StandaloneSpoke)
 
+    @property
+    def ordered_action_classes(self):
+        return self.interface._orderActionClasses(self.action_classes, self.hubs)
```
poncovka, line 61: Why don't we just fix the method `_orderActionClasses`?

poncovka:
From https://docs.python.org/3/library/functions.html#sorted:

> The built-in sorted() function is guaranteed to be stable. A sort is stable if it guarantees not to change the relative order of elements that compare equal — this is helpful for sorting in multiple passes (for example, sort by department, then by salary grade).

The problem is with the initial list of spokes. We need to sort these spokes by name first.

tests/nosetests/pyanaconda_tests/simple_ui_test.py
jkonecny12: I'm seeing this as another test. The first test is testing if the spokes are correctly assigned to hubs and they are there. This new test is testing if the spokes are explicitly positioned.
It feels like overloading test cases.

pyanaconda/ui/common.py
jkonecny12: This is default priority from the past but we can change it to make it stable with the current state.

pyanaconda/ui/gui/spokes/network.py
jkonecny12: OK

jkonecny12:
It's totally my fault that I did not looked on the official documentation to verify my discoveries from a past. I guess this could be a difference between the old Python 2 code and current Python 3. Good idea to fix spokes ordering instead of this solution.
However, there is a question if we don't want to require explicit ordering than using the sort. For me personally it seems that the explicit ordering is the better option and much easier to spot what is happening in the code.

tests/nosetests/pyanaconda_tests/simple_ui_test.py
jkonecny12: See my answer below in https://github.com/rhinstaller/anaconda/pull/2435#issuecomment-614055745.

jkonecny12:
**UPDATED:**
* Implemented notes from @poncovka.
* Added ordering of spokes before dividing them to the action classes in `_orderActionClasses()`.
* Added test for this ordering to prevent happening this in the future.
* Change the default priority to `0`.
* Move priority test to the existing test functions to remove those giant patching a preparation.
* Make ``ui/__init__.py`` almost pep8 compliant (without changing API).

I'm solving this on two fronts.
1. Adding priority to solve the Unsupported HW and Kernel spokes ordering and adding test to require setting different priority if there are two spokes with the same priority on the same hub.
2. Sorting spokes by name before sorting them by priority.

The first one will solve the second for us but not for spokes. This way we will make sure that we are not depending just on priority/names and addons can make use of that.

pyanaconda/ui/__init__.py
```diff
@@ -147,13 +153,21 @@ def _collectActionClasses(self, module_pattern_w_path, standalone_class):
         standalones = []
 
         for module_pattern, path in module_pattern_w_path:
-            standalones.extend(collect(module_pattern, path, lambda obj: issubclass(obj, standalone_class) and \
-                                       getattr(obj, "preForHub", False) or getattr(obj, "postForHub", False)))
+            standalones.extend(
+                collect(module_pattern,
+                        path,
+                        lambda obj: issubclass(obj, standalone_class) and
```
poncovka, line 159: Sometimes it is better to define a new function (it can be nested) instead of of using lambda functions.

pyanaconda/ui/__init__.py
```diff
@@ -163,13 +177,22 @@ def _orderActionClasses(self, spokes, hubs):
...
-            actionClasses.extend(sorted(filter(lambda obj, h=hub: getattr(obj, "preForHub", None) == h, spokes),
-                                        key=lambda obj: obj.priority))
-            actionClasses.append(hub)
-            actionClasses.extend(sorted(filter(lambda obj, h=hub: getattr(obj, "postForHub", None) == h, spokes),
-                                        key=lambda obj: obj.priority))
-
-        return actionClasses
+            action_classes.extend(
+                sorted(
+                    filter(lambda obj, h=hub: getattr(obj, "preForHub", None) == h,
```
poncovka, line 186: I don't think this is readable. I would replace the lambda function with a function. It looks like it could be used also in tests. Same below.

tests/nosetests/pyanaconda_tests/simple_ui_test.py
```diff
@@ -70,6 +75,20 @@ def _get_category_names(self, hub_class):
...
+
+        for spoke in spokes:
+            priority = spoke.priority
+            name = spoke.__name__
+
+            if priority in res:
+                self.assertNotIn(
+                    priority,
+                    res,
+                    "Spokes {} and {} have the same priority!".format(res[priority], name))
```
poncovka, line 89: I would create the message first and then call the assert method, so all arguments can be on the same line.

tests/nosetests/pyanaconda_tests/simple_ui_test.py
```diff
@@ -110,6 +129,16 @@ def tui_test(self, proxy_getter):
             ]
         })
 
+        # Force us to always decide order of standalone spokes based on priority not by name.
+        # This will ordering errors easier to spot.
+        spokes = self.ordered_action_classes
+
+        for hub in self.hubs:
+            pre_spokes = filter(lambda obj, h=hub: getattr(obj, "preForHub", None) == h, spokes)
```
poncovka, line 137: I don't understand why we define `h=hub` in these lambda functions.

jkonecny12: I've just copied this. I guess it's some old convention?

pyanaconda/ui/__init__.py
jkonecny12: I've only reformatted the code not change it. But yeah why not.

pyanaconda/ui/__init__.py
jkonecny12: Fixed.

tests/nosetests/pyanaconda_tests/simple_ui_test.py
jkonecny12: Fixed.

tests/nosetests/pyanaconda_tests/simple_ui_test.py
jkonecny12: Fixed.

tests/nosetests/pyanaconda_tests/simple_ui_test.py
```diff
@@ -156,3 +185,36 @@ def gui_test(self, window_group, window, meh, builder, proxy_getter):
...
+
+        # Force us to always decide order of standalone spokes based on priority not by name.
+        # This will ordering errors easier to spot.
+        spokes = self.ordered_action_classes
+
+        for hub in self.hubs:
+            pre_spokes = filter(lambda obj, h=hub: getattr(obj, "preForHub", None) == h, spokes)
+            self._check_spokes_priority_uniqueness(pre_spokes)
+            post_spokes = filter(lambda obj, h=hub: getattr(obj, "postForHub", None) == h, spokes)
+            self._check_spokes_priority_uniqueness(post_spokes)
```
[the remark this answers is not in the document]
jkonecny12: Fixed.

tests/nosetests/pyanaconda_tests/simple_ui_test.py
```diff
@@ -156,3 +185,36 @@ def gui_test(self, window_group, window, meh, builder, proxy_getter):
...
+            preForHub = hub
+
+        class SpokeC(StandaloneSpoke):  # pylint: disable=abstract-method
+            preForHub = hub
+
+        list1 = [SpokeA, SpokeB, SpokeC]
+        list2 = [SpokeC, SpokeB, SpokeA]
+
+        for spoke in list1:
+            spoke.preForHub = hub
```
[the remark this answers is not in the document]
jkonecny12: Fixed.

tests/nosetests/pyanaconda_tests/simple_ui_test.py
```diff
@@ -156,3 +185,36 @@ def gui_test(self, window_group, window, meh, builder, proxy_getter):
...
+
+        # the input list ordering shouldn't matter sorting should be based on class names
+        self.assertEqual(UserInterface._orderActionClasses(list1, [hub]),
+                         UserInterface._orderActionClasses(list2, [hub]))
```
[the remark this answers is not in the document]
jkonecny12: Fixed.

jkonecny12:
**UPDATED:**
* Implemented suggestions from Vendy.

pyanaconda/ui/__init__.py
```diff
@@ -163,13 +180,28 @@ def _orderActionClasses(self, spokes, hubs):
...
-        return actionClasses
+            action_classes.extend(
+                sorted(
+                    UserInterface._filter_spokes_by_pre_for_hub_reference(ordered_spokes, hub),
+                    key=lambda obj: obj.priority)
+            )
+            action_classes.append(hub)
+            action_classes.extend(
+                sorted(
+                    UserInterface._filter_spokes_by_post_for_hub_reference(ordered_spokes, hub),
```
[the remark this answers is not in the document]
jkonecny12: This is a static method so I don't have self.

PR #2436, opened 2020-04-10
Use unsquashfs to install the Live system when possible

bkhomuts:
This is an optimization of the installation process.
Before this change, the image was installed using rsync utility. However, this is not efficient: rsync is not aware of the squash filesystem's internal structure. Therefore it reads the file with sub-optimal speed. On the other hand, unsquashfs uses optimized functions to quickly decompress the image. Unsquashfs takes advantage of multiple processors available on the system and handles several files at a time. The decompression threads will span across all available logical cores saturating the CPU. In the end, this reduces the total installation time.

On my setup, I measured 2 times installation time reduction, when the CPU is constraint in the installation. (approx. 2 minutes in absolute terms -- benchmarks to be provided later).
This benefit in installation time will be clearly seen on multi-processor systems, in case single-core performance is a constraint during the installation.

One of the drawbacks of unsquashfs is that the utility does not support for "exclude" behavior, similar to rsync. I handle this by removing the not wanted files, that are in fact present in squashfs.img

The compatibility with older types of images is preserved. If the squashfs on the DVD is not 'plain', the old rsync approach will be used.

This change is very similar to #2292, but targets Fedora 32 installer branch.
I tested this change myself on Fedora 32 Workstation x86_64 Beta image with both types of the squashfs image.
In both cases, the resulting system is installed and booted normally.

You can download the plain squashfs image from https://khomutsky.com/fedora-dvd/Fedora-Workstation-dvd-x86_64-32_Beta-1.2-bkhomuts_squashfs_XZ_plain.iso , apply this patch there and test yourself.

Note: I noticed a bug in /usr/sbin/liveinst, that does not allow multiple launches of the installation. The bug will be addressed in a different merge request. See #2292 how to address the bug in liveinst.

cc: @VladimirSlavik , @bcl , @jkonecny12

### DIFF
--- a/pyanaconda/payload/livepayload.py
+++ b/pyanaconda/payload/livepayload.py
@@ -31,6 +31,7 @@ import stat
 import requests
 import hashlib
 import glob
+import re
 import functools
 from time import sleep
 from threading import Lock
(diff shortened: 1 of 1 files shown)


jkonecny12:
Thank you for this PR but it's too late for Fedora 32. We are not releasing new versions after GA so and right now we have final freeze. Please finish this feature on Rawhide and we can include this to Fedora 33.

kkocdko:
This is a very helpful feature, maybe I'll try to port it later?

PR #2441, opened 2020-04-14
Fix order of spokes in TUI test

VladimirSlavik:
Something was broken, some time ago, somewhere. Not sure about any of that. But this fixes the test.


VladimirSlavik:
OK, looks like this is different on my own system and where tests run... Any ideas?

VladimirSlavik:
See #2435

PR #2442, opened 2020-04-14
Mounting sources unification

VladimirSlavik:
This makes sources that mount share common ancestors.

Pending:

- [x] Pylint has a problem because I have a subclass that is meant to be abstract, and does not implement some stuff. ~~Not sure what to do with that.~~ It's broken so adding ignores.
- [x] Naming and locations of classes need feedback.
- [x] Tests for the base classes.
- [x] What to do about ancestry of the HDISO stuff. -> Another PR.
- [x] Generating the mount points

### DIFF
--- a/pyanaconda/modules/payloads/source/mount_tasks.py
+++ b/pyanaconda/modules/payloads/source/mount_tasks.py
@@ -1,0 +1,74 @@
+#
+# Base object for payload sources that use mounting.
+#
+# Copyright (C) 2020 Red Hat, Inc.
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# the GNU General Public License v.2, or (at your option) any later version.
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY expressed or implied, including the implied warranties of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General
+# Public License for more details.  You should have received a copy of the
+# GNU General Public License along with this program; if not, write to the
+# Free Software Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
+# 02110-1301, USA.  Any Red Hat trademarks that are incorporated in the
+# source code or documentation are not subject to the GNU General Public
+# License and may only be used or replicated with the express permission of
+# Red Hat, Inc.
+#
+import os.path
+from abc import ABC, abstractmethod
+
+from pyanaconda.modules.common.task import Task
+from pyanaconda.modules.common.errors.payload import SourceSetupError
+
+from pyanaconda.payload.utils import unmount
+from pyanaconda.anaconda_loggers import get_module_logger
+
+log = get_module_logger(__name__)
+
+
+class TearDownMountTask(Task):
+    """Task to tear down a mounting installation source.
+
+    This is universal for any source that can be directly unmounted with no additional work.
+    """
+
+    def __init__(self, target_mount):
+        super().__init__()
+        self._target_mount = target_mount
+
+    @property
+    def name(self):
+        return "Tear down mount installation source"
+
+    def run(self):
+        """Run source un-setup."""
+        log.debug("Unmounting installation source")
+        unmount(self._target_mount)
+
+
+class SetUpMountTask(Task, ABC):
+    """Abstract base class for set up tasks that need mounting."""
+
+    def __init__(self, target_mount):
+        super().__init__()
+        self._target_mount = target_mount
+
+    def run(self):
+        self._check_mount()
+        return self._do_mount()
+
+    def _check_mount(self):
+        if os.path.ismount(self._target_mount):
+            raise SourceSetupError("The mount point {} is already in use.".format(
+                self._target_mount
+            ))
+
+    @abstractmethod
+    def _do_mount(self):
+        """Override this method in descendants to do the actual work of mounting.
+
+        Return the result you want returned from the task."""
+        pass
--- a/pyanaconda/modules/payloads/source/source_base.py
+++ b/pyanaconda/modules/payloads/source/source_base.py
@@ -17,10 +17,17 @@
 # License and may only be used or replicated with the express permission of
 # Red Hat, Inc.
 #
-from abc import ABCMeta, abstractmethod
+import os.path
+from abc import ABC, ABCMeta, abstractmethod
 
 from dasbus.server.publishable import Publishable
+
 from pyanaconda.modules.common.base import KickstartBaseModule
+from pyanaconda.modules.payloads.source.mount_tasks import TearDownMountTask
+from pyanaconda.modules.payloads.source.utils import MountPointGenerator
+from pyanaconda.anaconda_loggers import get_module_logger
+
+log = get_module_logger(__name__)
 
 
 class PayloadSourceBase(KickstartBaseModule, Publishable, metaclass=ABCMeta):
@@ -84,3 +91,41 @@ class PayloadSourceBase(KickstartBaseModule, Publishable, metaclass=ABCMeta):
         :rtype: list[task]
         """
         pass
+
+
+class MountingSourceBase(PayloadSourceBase, ABC):
+    """Base class for sources that use mounting.
+
+    Implements some common functionality, most notably generation of mount point paths.
+    """
+    # pylint: disable=abstract-method
+
+    def __init__(self):
+        super().__init__()
+        self._mount_point = MountPointGenerator.generate_mount_point(self.type.value.lower())
+
+    def is_ready(self):
+        """This source is ready for the installation to start.
+
+        :return: ready or not
+        :rtype: bool
+        """
+        return os.path.ismount(self._mount_point)
+
+    @property
+    def mount_point(self):
+        """Where the source will be mounted.
+
+        :return: path to the mount point
+        :rtype: str
+        """
+        return self._mount_point
+
+    def tear_down_with_tasks(self):
+        """Tear down the installation source.
+
+        :return: list of tasks required for the source clean-up
+        :rtype: [TearDownMountTask]
+        """
+        task = TearDownMountTask(self._mount_point)
+        return [task]
--- a/tests/nosetests/pyanaconda_tests/module_source_base_test.py
+++ b/tests/nosetests/pyanaconda_tests/module_source_base_test.py
@@ -1,0 +1,122 @@
+#
+# Copyright (C) 2020  Red Hat, Inc.
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# the GNU General Public License v.2, or (at your option) any later version.
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY expressed or implied, including the implied warranties of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General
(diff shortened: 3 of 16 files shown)


VladimirSlavik:
Pylint has a problem because I have a subclass that is meant to be abstract, and does not implement some stuff. Not sure what to do with that.

tests/nosetests/pyanaconda_tests/module_source_hmc_test.py
```diff
@@ -58,7 +57,9 @@ def ready_state_test(self, ismount):
         ismount.return_value = True
 
         self.assertTrue(self.module.is_ready())
-        ismount.assert_called_once_with(INSTALL_TREE)
+        ismount.assert_called_once_with(
+            "/mnt/anaconda-sources/hmc-{:0>4}".format(self.module.counter - 1)
```
poncovka, line 61: Please, use `self.module.mount_point`. Same in other tests.

tests/nosetests/pyanaconda_tests/module_source_nfs_test.py
```diff
@@ -115,9 +114,12 @@ def ready_state_test(self, ismount):
 
         self.assertTrue(self.source_module.is_ready())
```
poncovka: Btw. I think that you can remove most of the tests for the `is_ready` method and keep only one.

pyanaconda/modules/payloads/source/harddrive/harddrive.py
```diff
@@ -43,8 +41,8 @@ def __init__(self):
         self._device = ""
         self.device_changed = Signal()
         self._install_tree_path = ""
-        self._device_mount = INSTALL_TREE + "_device"
-        self._iso_mount = INSTALL_TREE + "_iso"
+        self._device_mount = self._mount_point + "-device"
```
poncovka, line 44: Maybe I would override the `mount_point` property to return the device mount point. Then you can use the `super().is_ready` method.

pyanaconda/modules/payloads/source/harddrive/harddrive.py
```diff
@@ -43,8 +41,8 @@ def __init__(self):
...
+        self._iso_mount = self._mount_point + "-iso"
```
poncovka, line 45: Maybe, you could replace `TearDownHardDriveSourceTask` with two instances of `TearDownMountTask`. It would be good to also use `SetUpMountTask` to check the mount points.

poncovka:
About the pylint issues, maybe try to remove ABCMeta and ABC from `PayloadSourceBase` and `MountingSourceBase`. They already inherit from ABC. For example, `AbstractTask` seems to work this way.

VladimirSlavik:
> About the pylint issues (...)

Apparently pylint puts too little work into deciding what is abstract so I just added an ignore for that in the class.

pyanaconda/modules/payloads/source/source_base.py
```diff
@@ -84,3 +89,50 @@ def tear_down_with_tasks(self):
...
+    # pylint: disable=abstract-method
+
+    counter = 0
+
+    def __init__(self):
+        super().__init__()
+        self._mount_point = self._generate_mount_point()
+
+    def _generate_mount_point(self):
+        result = "/mnt/anaconda-sources/{}-{:0>4}".format(self.type.value.lower(), self.counter)
```
jkonecny12, line 108: Please use the same path we are using now. That means ``/run/install/...``

pyanaconda/modules/payloads/source/source_base.py
```diff
@@ -84,3 +89,43 @@ def tear_down_with_tasks(self):
...
+class MountingSourceBase(PayloadSourceBase, ABC):
+
+    counter = 0
+
+    def __init__(self):
+        super().__init__()
+        self._counter_value = MountingSourceBase.counter
+        self._mount_point = "/mnt/anaconda-sources/{}-{:0>4}".format(self.type.value.lower(),
+                                                                     self.counter)
+        MountingSourceBase.counter = MountingSourceBase.counter + 1
```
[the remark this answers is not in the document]
jkonecny12: I think that we want to store the id but not here but in the base class. It could be usable for plenty of other things (including logging) in the future and it would be great to have ID for every source. What do you think?

pyanaconda/modules/payloads/source/harddrive/harddrive.py
jkonecny12: Not sure if this suggestion is not bending the purpose of the `MountingSourceBase` class too much.

VladimirSlavik: Maybe the real problem here is whether harddrive belongs with other mountables or not?

In a very broad sense it does fit: it mounts, and wants a number. Then again it has multiple mount points, actually a variable number at that, so the actual functionality is different.

tests/nosetests/pyanaconda_tests/module_source_nfs_test.py
VladimirSlavik: Haha, I really am braindead :) Yes, the tests need way more changes, namely for the base classes. Sorry.

jkonecny12:
Also I'm thinking that we should have a separate test file for base source implementation. To test this code separated from an existing sources.
Right now I'm thinking about the same for payloads.

VladimirSlavik:
@jkonecny12 yes, that's precisely what I am doing now...

jkonecny12:
Yeah, I see that now. So I'll do the same for the payload base class. It would help me to test functionality for the `is_ready` solution.

pyanaconda/modules/payloads/source/source_base.py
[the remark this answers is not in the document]
VladimirSlavik: Do we want to decide that now? I agree keeping an ID would be somehow "proper", but  at this moment it is not really needed, and the mount point acts as a kind of ID, even if that is rather crude.

tests/nosetests/pyanaconda_tests/module_source_nfs_test.py
VladimirSlavik: I think I have addressed that by now...

pyanaconda/modules/payloads/source/harddrive/harddrive.py
VladimirSlavik: The former is another facepalm moment for me. The latter, I am not sure. I'll try to remember and bring it up at the next opportunity.

pyanaconda/modules/payloads/source/source_base.py
VladimirSlavik: Done: Now paths are like /run/install/source-TYPE-NNNN.

pyanaconda/modules/payloads/source/harddrive/harddrive.py
VladimirSlavik: Haha, another facepalm - I finally got what you meant by the other half. Check implemented in-place now, discussion still needed. I don't see how a class that assumes exactly one of something can be subclassed easily into handling more than one of that. I'd need to override all of the ancestor code anyway.

VladimirSlavik:
Update:
- Addressed all clear feedback above
- Tests now exist
- Rebased

Left to discuss:
- Should the Harddrive inherit from the common mounting ancestors, and how? It's a poor match.

pyanaconda/modules/payloads/source/source_base.py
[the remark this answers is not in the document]
poncovka: If we want to have ID for every source, than it is not definitely going to be `counter_value` in `MountingSourceBase`, so please remove that attribute. We can easily modify the code for generating mount points with this future ID.

pyanaconda/modules/payloads/source/harddrive/harddrive.py
poncovka: This solution seems to be fine.

pyanaconda/modules/payloads/source/harddrive/harddrive.py
poncovka: If the `mount_point` property cannot return a real mount point, then `HardDriveSourceModule` shouldn't be a subclass of `MountingSourceBase`. However, we need to be able to generate the "base" mount point. I still think that there should be a separate class called `MountPointGenerator` with one class method `create_mount_point` that will return a unique mount point. There might be more cases like this one in the future.

pyanaconda/modules/payloads/source/source_base.py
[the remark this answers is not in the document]
VladimirSlavik: Removed already - somehow GitHub keeps showing the old diff even if the other lines changed :(

In terms of keeping the ID, leaving that open for the future.

VladimirSlavik:
Update:
- Changed generation of mount points
- Harddrive is left for another PR
- Removed all commits with other concerns to separate PRs

poncovka:
> I think it would be better to move the `MountPointGenerator` to `core.utils` or somewhere like that. It seems to me like something general to be used. If that will be done than it should be renamed to something more general. Something like `UniquePathGenerator` and the INSTALLTREE should be removed. It also should have it's own tests to easily making sure that changing this class will still result in the same output used elsewhere.
> 
> `MountPointGenerator` than could be specialization of this class if needed.
> 
> Do you agree with me @poncovka ?

I would prefer to have as much code in the modules as possible. Also, I don't really see where else we could use that.

tests/nosetests/pyanaconda_tests/module_source_base_test.py
```diff
@@ -0,0 +1,122 @@
...
+        return SourceType.URL
+
+    def for_publication(self):
+        return None
+
+    def set_up_with_tasks(self):
+        return []
+
+
+class DummySetUpMountTaskSubclass(SetUpMountTask):
```
poncovka, line 44: In general, I don't think it is necessary to write tests with dummy implementations. It is enough the test one of the real implementations.

pyanaconda/modules/payloads/source/source_base.py
```diff
@@ -84,3 +91,43 @@ def tear_down_with_tasks(self):
...
+        self._mount_point = MountPointGenerator.generate_mount_point(self.type.value.lower())
+
+    def is_ready(self):
+        """This source is ready for the installation to start.
+
+        :return: ready or not
+        :rtype: bool
+        """
+        res = os.path.ismount(self._mount_point)
+        log.debug("Source is set to %s ready state", res)
```
poncovka, line 114: @jkonecny12, isn't this message a leftover from debugging? I would expect the payload to log this kind of messages when it has sources in invalid state. Otherwise, it is not really useful.

VladimirSlavik: Removed.

pyanaconda/modules/payloads/source/mount_tasks.py
```diff
@@ -0,0 +1,77 @@
...
+    def run(self):
+        """Run source un-setup."""
+        log.debug("Unmounting installation source")
+        unmount(self._target_mount)
+
+
+class SetUpMountTask(Task, ABC):
+    """Abstract base class for set up tasks that need mounting.
+
+    Subclasses must implement _do_mount().
```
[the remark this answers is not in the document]
VladimirSlavik: Done!

VladimirSlavik:
Update: Removed some of the cruft you found in last round of reviews.

VladimirSlavik:
> I would like to see separate tests just for `MountPointGenerator` but I'll approve even this variant.

I'm thinking tests for the related "utils and stuff" would be my next PR after the harddrive source changes...

PR #2446, opened 2020-04-15
Remove the tip about the user name (#1823015)

poncovka:
We show a warning when the user name is too long or it contains spaces.

Resolves: rhbz#1823015

### DIFF
--- a/pyanaconda/ui/gui/spokes/user.glade
+++ b/pyanaconda/ui/gui/spokes/user.glade
@@ -110,6 +110,7 @@
                       <object class="GtkEntry" id="username_entry">
                         <property name="visible">True</property>
                         <property name="can_focus">True</property>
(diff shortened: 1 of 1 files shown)


VladimirSlavik:
As a former writer, I propose these:

- Maximum user name length is 32 characters.
- User name length is limited to 32 characters.

Why:

- Maximum and limit are pretty universally understood words in most western languages
- With the words used, understanding sentence structure is not required to grasp the meaning
- 32 as a number feels more natural than 33

poncovka:
Maybe we could actually remove this tip.

username_long
username_spaces

jkonecny12:
Yep, remove the tip. We will make easier life for us and translators. Users will not be annoyed by the same message twice.

poncovka:
The tip is removed.

username_notip

PR #2447, opened 2020-04-16
Install test dependencies from pip when possible

poncovka:
We should prefer to install test dependencies via pip. The required RPMs might
not be available for all products. It should also help to reduce the number of
packages that we maintain.

### DIFF
--- a/scripts/testing/setup-mock-test-env.py
+++ b/scripts/testing/setup-mock-test-env.py
@@ -370,7 +370,7 @@ def install_pip_packages_to_mock(mock_command, packages):
     cmd = _prepare_command(mock_command)
 
     cmd = _run_cmd_in_chroot(cmd)
-    cmd.append('pip3 install {}'.format(packages))
+    cmd.append('pip3 install --install-option="--install-scripts=/usr/bin" {}'.format(packages))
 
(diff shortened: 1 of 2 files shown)


scripts/testing/setup-mock-test-env.py
```diff
@@ -370,7 +370,7 @@ def install_pip_packages_to_mock(mock_command, packages):
     cmd = _prepare_command(mock_command)
 
     cmd = _run_cmd_in_chroot(cmd)
-    cmd.append('pip3 install {}'.format(packages))
+    cmd.append('pip3 install --install-option="--install-scripts=/usr/bin" {}'.format(packages))
```
poncovka, line 373: Or we could add `/usr/local/bin` to `PATH` of the mock environment, but I couldn't figure out, how to do that.

poncovka:
Jenkins, test this please.

scripts/testing/setup-mock-test-env.py
jkonecny12: This solution seems to me better because we then don't depend on a mock more than required. It will be little bit easier to switch to the podman if we decide that in the future.

PR #2448, opened 2020-04-16
Translate names of OS installations on demand (#1823126)

poncovka:
Don't translate names of found OS installations immediately, otherwise
the translations might not match the current locale.

Resolves: rhbz#1823126

### DIFF
--- a/pyanaconda/storage/root.py
+++ b/pyanaconda/storage/root.py
@@ -124,28 +124,21 @@ def _find_existing_installations(devicetree):
             device.teardown()
             continue
 
-        try:
-            (architecture, product, version) = get_release_string(chroot=sysroot)
-        except ValueError:
-            name = _("Linux on %s") % device.name
-        else:
-            # I'd like to make this finer grained, but it'd be very difficult
-            # to translate.
-            if not product or not version or not architecture:
-                name = _("Unknown Linux")
-            elif "linux" in product.lower():
-                name = _("%(product)s %(version)s for %(arch)s") % \
-                    {"product": product, "version": version, "arch": architecture}
-            else:
-                name = _("%(product)s Linux %(version)s for %(arch)s") % \
-                    {"product": product, "version": version, "arch": architecture}
-
+        architecture, product, version = get_release_string(chroot=sysroot)
         (mounts, swaps) = _parse_fstab(devicetree, chroot=sysroot)
         blivet_util.umount(mountpoint=sysroot)
+
         if not mounts and not swaps:
             # empty /etc/fstab. weird, but I've seen it happen.
             continue
-        roots.append(Root(mounts=mounts, swaps=swaps, name=name))
+
+        roots.append(Root(
+            product=product,
+            version=version,
+            arch=architecture,
+            mounts=mounts,
+            swaps=swaps
+        ))
 
     return roots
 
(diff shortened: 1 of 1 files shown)


pyanaconda/storage/root.py
```diff
@@ -124,28 +124,28 @@ def _find_existing_installations(devicetree):
...
             (architecture, product, version) = get_release_string(chroot=sysroot)
         except ValueError:
-            name = _("Linux on %s") % device.name
-        else:
-            # I'd like to make this finer grained, but it'd be very difficult
-            # to translate.
-            if not product or not version or not architecture:
-                name = _("Unknown Linux")
-            elif "linux" in product.lower():
-                name = _("%(product)s %(version)s for %(arch)s") % \
```
VladimirSlavik, line 137: format?

pyanaconda/storage/root.py
```diff
@@ -312,32 +312,42 @@ def _parse_fstab(devicetree, chroot):
...
+        if self._name:
+            return self._name
 
-        if not self.name and "/" in self.mounts:
-            self.name = self.mounts["/"].format.uuid
+        if not self._product or not self._version or not self._arch:
+            return _("Unknown Linux")
+
+        if "linux" in self._product.lower():
+            return _("%(product)s %(version)s for %(arch)s") % \
```
VladimirSlavik, line 344: format?

pyanaconda/storage/root.py
```diff
@@ -312,32 +312,42 @@ def _parse_fstab(devicetree, chroot):
...
 
-        self.name = name    # eg: "Fedora Linux 16 for x86_64", "Linux on sda2"
+    @property
+    def name(self):
+        """The name of the OS."""
+        if self._name:
+            return self._name
 
-        if not self.name and "/" in self.mounts:
-            self.name = self.mounts["/"].format.uuid
```
jkonecny12, line 339: I'm missing this part in the new code. Did I overlooked or is it removed on purpose?

pyanaconda/storage/root.py
```diff
@@ -124,28 +124,28 @@ def _find_existing_installations(devicetree):
...
-            # to translate.
-            if not product or not version or not architecture:
-                name = _("Unknown Linux")
-            elif "linux" in product.lower():
-                name = _("%(product)s %(version)s for %(arch)s") % \
-                    {"product": product, "version": version, "arch": architecture}
-            else:
-                name = _("%(product)s Linux %(version)s for %(arch)s") % \
-                    {"product": product, "version": version, "arch": architecture}
+            pass
```
jkonecny12, line 134: I'm not quite happy with this solution. It's not really visible that you depend on the fact that `architecture`, `product`, `version` variables are `None` because they are not set. Could you please make this somehow more obvious?

pyanaconda/storage/root.py
poncovka: Fixed.

pyanaconda/storage/root.py
poncovka: The name was always specified to something, so this code was never used.

pyanaconda/storage/root.py
poncovka: The function `get_release_string` can already return None values, so I have moved the try-except block there and simplified this code.

poncovka:
Updated.

pyanaconda/storage/root.py
jkonecny12: Oh it can even return None. In that case it's even more confusing... :(

PR #2449, opened 2020-04-16
Make sure that the summary button is really hidden (#1823467)

poncovka:
Call the functions really_hide and really_show to really hide or show the
summary button in the spoke for the advanced storage. Don't set up attributes
that have no effect.

Resolves: rhbz#1823467


poncovka:
Jenkins, test this please.

PR #2450, opened 2020-04-17
Change payload source ready state

jkonecny12:
Payload ready state is not working for sources like URL which does not have a state. Problem with this is that Payload is not able to switch sources which are ready and on the other hand can't start installation with sources which are not ready.

Make this a 3 state - ready, unready and not implemented. That should be enough to solve any problem.

### DIFF
--- a/pyanaconda/modules/payloads/constants.py
+++ b/pyanaconda/modules/payloads/constants.py
@@ -56,9 +56,9 @@ class SourceState(Enum):
 
     These will be used only internally. Not with a DBus API.
     """
+    NOT_APPLICABLE = auto()
     READY = auto()
     UNREADY = auto()
-    NOT_SUPPORTED = auto()
 
     @staticmethod
     def from_bool(value):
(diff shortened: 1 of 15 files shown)


poncovka:
I think this PR should be rebased on a top of https://github.com/rhinstaller/anaconda/pull/2442.

jkonecny12:
**UPDATED:**
* Rebased and fixed conflicts

pyanaconda/modules/payloads/constants.py
```diff
@@ -48,3 +48,28 @@ class URLType(Enum):
...
+    READY = auto()
+    UNREADY = auto()
+    NOT_SUPPORTED = auto()
+
+    @staticmethod
+    def from_bool(value):
+        """Get state from a bool value.
+
+        This way we can't return NONE state but that is not problem. NONE state is specific
+        for sources which does not have a state, so they don't have to convert it from bool.
```
M4rtinK, line 68: not problem -> not a problem

does not have -> do not have

jkonecny12:
**UPDATED:**
* Fixed issue pointed by @M4rtinK

jkonecny12:
**UPDATED:**
* Renamed `NOT_SUPPORTED` state to `NOT_APPLICABLE`

PR #2454, opened 2020-04-17
Unify module source test variable names

VladimirSlavik:
They are shorter now, and same in all tests.


VladimirSlavik:
I think this is low priority and should wait for #2450, #2457, and potentially anything else that is more vital.

VladimirSlavik:
Rebased. Will discuss before merging.

PR #2455, opened 2020-04-19
Pre-select disk initialization mode in TUI

jstodola:
No disk initialization method was pre-selected in TUI. Pre-select "Use All Space" by default, but allow the clearpart kickstart command to override the method.

### DIFF
--- a/pyanaconda/ui/tui/spokes/storage.py
+++ b/pyanaconda/ui/tui/spokes/storage.py
@@ -426,8 +426,11 @@ class PartTypeSpoke(NormalTUISpoke):
         # Choose the initialization mode.
         self._disk_init_proxy = STORAGE.get_proxy(DISK_INITIALIZATION)
         self._orig_init_mode = self._disk_init_proxy.InitializationMode
-        self._init_mode = self._disk_init_proxy.InitializationMode
+        self._init_mode = self._orig_init_mode
         self._init_mode_list = sorted(INIT_MODES.keys())
+
+        if self._init_mode == CLEAR_PARTITIONS_DEFAULT:
+            self._init_mode = CLEAR_PARTITIONS_ALL
 
         # Choose the partitioning method.
         self._storage_module = storage_module


jkonecny12:
Jenkins, it's ok to test.

poncovka:
Fedora 31 and RHEL 8 don't pre-select the disk initialization method as well, so this is not a fix of a regression. The question is whether we want to change the behaviour. At this moment, the user has to choose one of the options.

clearpart

jstodola:
@VladimirSlavik , yes, it's possible that the idea of the change is completely wrong and this logic needs to be somewhere else, thanks for checking this attempt :-)
@poncovka right, this is also broken in RHEL-8. On the other hand, the pre-selected option in RHEL-7 is "Use All Space". I wrote "broken" intentionally - even though it might be OK to have nothing pre-selected, anaconda should not allow to continue to the partitioning scheme screen unless any option has been selected. And that's not true right now.

poncovka:
Ok, I have found the commit that changed the behaviour (see https://github.com/rhinstaller/anaconda/commit/ed9d69dd805c75e8068a6f22f619c205d87ffd4f#diff-c16f13027c8253a68b206b751513b4acL477). Based on that, we should select "Use All Space" by default.

pyanaconda/ui/tui/spokes/storage.py
```diff
@@ -426,7 +426,11 @@ def __init__(self, data, storage, payload, storage_module, partitioning):
         # Choose the initialization mode.
         self._disk_init_proxy = STORAGE.get_proxy(DISK_INITIALIZATION)
         self._orig_init_mode = self._disk_init_proxy.InitializationMode
-        self._init_mode = self._disk_init_proxy.InitializationMode
+        # Preserve clearpart mode from kickstart.
+        if self._orig_init_mode != CLEAR_PARTITIONS_DEFAULT:
+            self._init_mode = self._orig_init_mode
+        else:
+            self._init_mode = CLEAR_PARTITIONS_ALL
         self._init_mode_list = sorted(INIT_MODES.keys())
```
poncovka, line 434: I would reorganize the code a little to make it easier to follow. Also, the comment about the kickstart is not really relevant, because the property could be set directly via DBus.

```suggestion
        self._init_mode = self._orig_init_mode
        self._init_mode_list = sorted(INIT_MODES.keys())

        if self._init_mode == CLEAR_PARTITIONS_DEFAULT:
            self._init_mode = CLEAR_PARTITIONS_ALL
```

jstodola:
Thanks for the suggestion, Vendy, it looks better now.
I've applied your suggestion, which created an additional commit. Would you prefer this PR is fixed as a single commit or leave it as it is?

poncovka:
@jstodola, it might be better to squash them into one commit if it is not a problem :-)

jstodola:
The two commits have been squashed into one.

PR #2458, opened 2020-04-21
Reset the partitioning of Blivet-GUI (#1826286)

poncovka:
Fix the "Reset all" button in Blivet-GUI and reset the current partitioning.

Resolves: rhbz#1826286


M4rtinK:
Merging now so that a fixed Anaconda build can be done for F32 and Rawhide.

PR #2460, opened 2020-04-21
Stop passing rd.{dm,md,lvm,luks}=0 in installer environment

AdamWill:
This blob has been maturing here ever since 2012. @dwlehman was
surprised it was still there, and thinks we've really been
aiming for anaconda/blivet to work with boot-time activation of
storage devices since 2014 or so, so this blob should go away
and we should fix any bugs that show up.

### DIFF
--- a/dracut/parse-anaconda-options.sh
+++ b/dracut/parse-anaconda-options.sh
@@ -10,18 +10,6 @@ mkdir -p $repodir $isodir
 # add some modules
 modprobe -q edd
 
-# NOTE: anaconda historically activated all the fancy disk devices itself,
-# and it would get very confused if they were already active when it started.
-# F17 has some support for handling already-active devices, but it's still
-# currently safer to disable these things and let anaconda activate them.
-# TODO FIXME: remove this and make anaconda handle active devices!
-{
-    for t in dm md lvm luks; do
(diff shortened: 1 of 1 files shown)


AdamWill:
I have not tested this *at all*, and it probably breaks stuff. I am more proposing it for discussion and as a 'tracker' so we don't forget this whole thing (which @dwlehman and @vojtechtrefny and I discussed a bit on IRC yesterday) as we go forward.

jkonecny12:
Hmm, interesting. Personally I'm for every removal, but I fear this will introduce more code than remove :D.

Let's wait on reply from blivet guys what they think about it.

jkonecny12:
Jenkins, it's ok to test.

dwlehman:
In general this should work. There may be hiccups, but blivet is generally capable of handling already-active devices. In fact, most of the time devices are already active when blivet starts -- possibly because of the `blivet.udev.trigger` call in `pyanaconda.storage.initialization.enable_installer_mode`.

poncovka:
Jenkins, test this please.

poncovka (approved):
Looks good to me. Thanks!

poncovka:
I have done some basic testing and it seems to work fine.

PR #2461, opened 2020-04-22
Don't initialize disks that are not empty (#1825067)

poncovka:
The interactive partitioning shouldn't try to initialize disks that are not empty.
The method `_is_device_empty` of the class `DiskInitializationConfig` should
reflect the logic of the Blivet's method `initialize_disk`.

Resolves: rhbz#1825067

### DIFF
--- a/pyanaconda/modules/storage/disk_initialization/configuration.py
+++ b/pyanaconda/modules/storage/disk_initialization/configuration.py
@@ -118,10 +118,15 @@ class DiskInitializationConfig(object):
 
     def _is_device_empty(self, device):
         """Is the given device empty?"""
+        # Is the device partitioned?
         if not device.partitioned:
             return device.format.type is None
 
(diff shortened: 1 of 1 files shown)


vojtechtrefny:
There are currently some problems with partitions in rawhide and possibly also in F32. Basically it is possible that partitions are not discovered by kernel after boot (meaning no nodes in `/dev` and no records in sysfs) so we don't add partitions to devicetree during reset. But after running some parted checks the partitions magically appears, that's why `device.format.partitions` works (this is list of parted partitions, not "devicetree partitions"). Generally this should never happen and your original check should always work so no change is needed.
But I agree the `_is_device_empty` function with checking for magic partitions is too "low level" and shouldn't be in Anaconda and we should provide API for checking this in blivet.

poncovka:
I am closing this PR based on Vojta's comment.

vojtechtrefny:
Promised API in blivet: https://github.com/storaged-project/blivet/pull/839

PR #2462, opened 2020-04-22
date time gui: no empty space in combo for day and month selection (#…

rvykydal:
…1823130)

### DIFF
--- a/pyanaconda/ui/gui/spokes/datetime_spoke.py
+++ b/pyanaconda/ui/gui/spokes/datetime_spoke.py
@@ -148,6 +148,7 @@ def _new_date_field_box(store):
 
     # idx is column 0, string we want to show is 1
     combo.add_attribute(renderer, "text", 1)
+    combo.set_wrap_width(1)
 
     box.pack_start(combo, False, False, 0)
     box.pack_start(suffix_label, False, False, 0)


rvykydal:
For the description and background of the issue see: https://bugzilla.redhat.com/show_bug.cgi?id=1823130#c1
Current state (day, month, year):
date_time_current_day
date_time_current_month
date_time_current_year

Patched (day, month, year):
date_time_1c_day
date_time_1c_month
date_time_1c_year

We could go even further and use `combo.set_wrap_width(3)` but it seems maybe too uncommon to me (day, month, year):
date_time_3c_day
date_time_3c_month
date_time_3c_year

jkonecny12:
I think the third option is the best one. It's not taking whole screen height which is great. Even thought, I'm not sure if it's not confusing when those values are above other combo boxes...

jstodola:
Radek, the patched version looks great!
The combo.set_wrap_width(3) version is weird, it requires an extra effort to find the number of your choice, since you need to figure out what is the flow of the numbers in the first place.

jkonecny12:
> Radek, the patched version looks great!
> The combo.set_wrap_width(3) version is weird, it requires an extra effort to find the number of your choice, since you need to figure out what is the flow of the numbers in the first place.

He's a tester he's right :). No, but really I'm not that convinced what to pick so feel free with the current solution.

poncovka:
Couldn't we use GtkCalendar? https://lazka.github.io/pgi-docs/Gtk-3.0/classes/Calendar.html

jkonecny12:
Seems like the best idea really.

rvykydal:
> Couldn't we use GtkCalendar? https://lazka.github.io/pgi-docs/Gtk-3.0/classes/Calendar.html

That was my option 3) in https://bugzilla.redhat.com/show_bug.cgi?id=1823130#c1
The reason I went with this simple patch was that it would be bigger change
- I didn't want to invest more time into it.
- It would require update of documentation (new screenshots)
- Gnome Date and Time setting is not using it (why?) - but that is in F30, I haven't checked with latest Fedora.

rvykydal:
Hm, seems you really care so I'll try to use the Calendar widget.

rvykydal:
>     * Gnome Date and Time setting is not using it (why?) - but that is in F30, I haven't checked with latest Fedora.
Still using this:
rawhide_f33_date_time

rvykydal:
I'd prefer to open another issue/bz: use GtkCalendar in Date Time spoke.
It will take more time than we think to agree on design / how the link/button taking us to the dialog with calendar will look, how the current date would be displayed, etc and I can imagine it would in the end lead also to update of left hand side (time).

poncovka:
@rvykydal, GtkCalendar feels more natural to me. However, this is really out of scope of this bug, so feel free to open a new issue for that.

rvykydal:
> @rvykydal, GtkCalendar feels more natural to me. However, this is really out of scope of this bug, so feel free to open a new issue for that.

https://bugzilla.redhat.com/show_bug.cgi?id=1827265

PR #2463, opened 2020-04-22
Don't mix kickstart commands for installation methods

poncovka:
Make sure that the kickstart command `method` is used only
for the RPM sources, so we can later move the RPM sources
to the DBus module.


pyanaconda/payload/live/payload_liveos.py
```diff
@@ -41,20 +42,36 @@ def type(self):
...
+    def _get_live_os_image():
+        """Detect live os image on the system.
+
+        :return: a path to the image
+        """
+        payloads_proxy = PAYLOADS.get_proxy()
+        source_path = payloads_proxy.CreateSource(SOURCE_TYPE_LIVE_OS_IMAGE)
+
+        source_proxy = PAYLOADS.get_proxy(source_path)
+        return source_proxy.DetectLiveOSImage()
```
jkonecny12, line 55: This should be handled better way I think. It's fine for now but I don't think we should create source to be able to detect if there is a source. This should be ideally part of the payload?

Please add TODO here.

poncovka: Added a comment.

poncovka:
Rebased.

poncovka:
Tested.

PR #2464, opened 2020-04-22
subscription: Configure RHSM service

M4rtinK:
The first two commits are small fixups, while the third commits introduces the code for setting RHSM service configuration.

### DIFF
--- a/pyanaconda/modules/subscription/runtime.py
+++ b/pyanaconda/modules/subscription/runtime.py
@@ -1,0 +1,116 @@
+#
+# Copyright (C) 2020 Red Hat, Inc.
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# the GNU General Public License v.2, or (at your option) any later version.
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY expressed or implied, including the implied warranties of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General
+# Public License for more details.  You should have received a copy of the
+# GNU General Public License along with this program; if not, write to the
+# Free Software Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
+# 02110-1301, USA.  Any Red Hat trademarks that are incorporated in the
+# source code or documentation are not subject to the GNU General Public
+# License and may only be used or replicated with the express permission of
+# Red Hat, Inc.
+#
+from dasbus.typing import get_variant, Str
+
+from pyanaconda.modules.common.task import Task
+
+from pyanaconda.anaconda_loggers import get_module_logger
+log = get_module_logger(__name__)
+
+
+class SetRHSMConfigurationTask(Task):
+    """Task for setting configuration to the RHSM service.
+
+    Set configuration options of the RHSM service via it's
+    DBus interface, based on the provided SubscriptionRequest
+    structure.
+
+    Also in case one of the configuration options was unset,
+    restore the key to its original value. This way for example
+    a user decides at runtime to use the default server hostname
+    or RHSM baseurl, they can just delete the value in the UI,
+    triggering the original value to be restored when we encounter
+    the empty value for a key that originally was set to
+    a non empty value.
+    """
+
+    # Keys in the RHSM config key/value store we care about and
+    # should be able to restore to original value.
+    #
+    # NOTE: These keys map 1:1 to rhsm.conf. To see what they do the
+    #       best bet is to check the /etc/rhsm/rhsm.conf file on a system
+    #       with the subscription-manager package installed. The file
+    #       is heavily documented with comment's explaining what
+    #       the different keys do.
+    CONFIG_KEY_SERVER_HOSTNAME = "server.hostname"
+    CONFIG_KEY_SERVER_PROXY_HOSTNAME = "server.proxy_hostname"
+    CONFIG_KEY_SERVER_PROXY_PORT = "server.proxy_port"
+    CONFIG_KEY_SERVER_PROXY_USER = "server.proxy_user"
+    CONFIG_KEY_SERVER_PROXY_PASSWORD = "server.proxy_password"
+    CONFIG_KEY_RHSM_BASEURL = "rhsm.baseurl"
+
+    def __init__(self, rhsm_config_proxy, rhsm_config_defaults, subscription_request):
+        """Create a new task for setting RHSM configuration.
+
+        :param rhsm_config_proxy: DBus proxy for the RHSM Config object
+        :param dict rhsm_config_defaults: a dictionary of original RHSM configuration values
+        :param subscription_request: subscription request DBus Structure
+        :type subscription_request: SubscriptionRequest instance
+        """
+        super().__init__()
+        self._rhsm_config_proxy = rhsm_config_proxy
+        self._request = subscription_request
+        self._rhsm_config_defaults = rhsm_config_defaults
+
+    @property
+    def name(self):
+        return "Set RHSM configuration."
+
+    def run(self):
+        log.debug("subscription: setting RHSM config values")
+        # We will use the SetAll() dbus method and we need to
+        # assemble a dictionary that we will feed to it.
+        # Start by preparing a SubscriptionData property mapping
+        # to the RHSM config keys.
+        #
+        # A note about constructing the dict:
+        # - DBus API needs all values to be strings, so we need to convert the
+        #   port number to string
+        # - all values need to be string variants
+        # - proxy password is stored in SecretData instance and we need to retrieve
+        #   its value
+        property_key_map = {
+            self.CONFIG_KEY_SERVER_HOSTNAME: self._request.server_hostname,
+            self.CONFIG_KEY_SERVER_PROXY_HOSTNAME: self._request.server_proxy_hostname,
+            self.CONFIG_KEY_SERVER_PROXY_PORT: str(self._request.server_proxy_port),
+            self.CONFIG_KEY_SERVER_PROXY_USER: self._request.server_proxy_user,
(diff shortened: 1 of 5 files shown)


pyanaconda/modules/subscription/runtime.py
```diff
@@ -0,0 +1,111 @@
...
+    CONFIG_KEY_SERVER_PROXY_USER = "server.proxy_user"
+    CONFIG_KEY_SERVER_PROXY_PASSWORD = "server.proxy_password"
+    CONFIG_KEY_RHSM_BASEURL = "rhsm.baseurl"
+
+    def __init__(self, rhsm_config_proxy, rhsm_config_defaults, subscription_request):
+        """Create a new task for setting RHSM configuration.
+
+        :param rhsm_config_proxy: DBus proxy for the RHSM Config object
+        :param dict rhsm_config_defaults: a dictionary of original RHSM configuration values
+        :param subscription_request: subscription request DBus
```
rvykydal, line 61: DBus -> DBus Structure ?

pyanaconda/modules/subscription/runtime.py
```diff
@@ -0,0 +1,111 @@
...
+                config_dict[key] = value
+            else:
+                # if no value is present in request, use
+                # value from the original RHSM config state
+                # (if any)
+                log.debug("subscription: restoring original value for RHSM config key %s", key)
+                config_dict[key] = self._rhsm_config_defaults.get(key, "")
+
+        # and finally set the dict to RHSM via the DBus API
+        self._rhsm_config_proxy.SetAll(config_dict, "")
```
poncovka, line 111: I don't see the DBus method `SetAll` in the documentation of the RHSM Configuration object. Are you sure this is working? I see only the method `Set` and it requires a variant.

pyanaconda/modules/subscription/subscription.py
```diff
@@ -509,3 +515,41 @@ def rhsm_observer(self):
...
+        first "warming up" this property by reading it or else
+        it might return non-default (Anaconda overwritten) data.
+
+        :return : dictionary of default RHSM configuration values
+        :rtype: dict
+        """
+        if self._rhsm_config_defaults is None:
+            # config defaults cache not yet populated, do it now
+            proxy = self.rhsm_observer.get_proxy(RHSM_CONFIG)
+            self._rhsm_config_defaults = proxy.GetAll("")
```
poncovka, line 540: Could you please create a constant for the empty string? Otherwise, it is a little confusing. It can be done in another PR.

pyanaconda/modules/subscription/subscription.py
```diff
@@ -510,3 +515,41 @@ def rhsm_observer(self):
...
+        return self._rhsm_config_defaults
+
+    def set_rhsm_config_with_task(self):
+        """Set RHSM config values based on current subscription request.
+
+        :return: a DBus path of an installation task
+        """
+        # NOTE: we access self._subscription_request directly
+        #       to avoid the sensitive data clearing happening
+        #       in the subscription_request property getter
```
jkonecny12, line 550: This note can be solved by new read-only getter with self describing name. It would take less lines than this explanation.

pyanaconda/modules/subscription/runtime.py
M4rtinK: > don't see the DBus method `SetAll` in the documentation of the RHSM Configuration object. Are you sure this is working? 

Yes, it was recently added but for some reason has not yet shown up in the DBus API documentation:
https://github.com/candlepin/subscription-manager/pull/2224
I have reported this to the Subscription Manager team on our Tuesday sync meeting, so it should be fixed soon.

>I see only the method `Set` and it requires a variant.

So this is the definition for SetAll as far as I can tell:

```python3
    @util.dbus_service_method(
        constants.CONFIG_INTERFACE,
        in_signature='a{sv}s')
    @util.dbus_handle_exceptions
    def SetAll(self, configuration, locale, sender=None):
        """
        Method for setting multiple configuration options. Of course all of them could be set.
        :param configuration: d-bus dictionary with configuration. Keys have to include section. e.g.
                              server.hostname. Configuration file is saved to the file at the end of method.
        :param locale: string with locale
        :param sender: not used
        :return: None
        """
        configuration = dbus_utils.dbus_to_python(configuration, expected_type=dict)
        locale = dbus_utils.dbus_to_python(locale, expected_type=str)
        Locale.set(locale)

```
So `a{sv}s` = dictionary with string to variant mapping (for the config values) & string (for locale). So is it enough to pass it a dictionary (which has string keys and only string values) and dasbus will do the magic, or do I need to do some more processing on it ?

pyanaconda/modules/subscription/subscription.py
M4rtinK: Yeah, that's a good point. I'll do a separate PR for it as it's already used in other places.

BTW, about the locale string - its used to specify locale for the error messages RHSM returns, so we only set it to the actual locale when the error message is going to be shown to the user in the UI. In other cases we keep the default value so that any error messages that get logged are not translated.

pyanaconda/modules/subscription/runtime.py
```diff
@@ -0,0 +1,111 @@
...
+    def __init__(self, rhsm_config_proxy, rhsm_config_defaults, subscription_request):
+        """Create a new task for setting RHSM configuration.
+
+        :param rhsm_config_proxy: DBus proxy for the RHSM Config object
+        :param dict rhsm_config_defaults: a dictionary of original RHSM configuration values
+        :param subscription_request: subscription request DBus
+        :type subscription_request: SubscriptionRequest instance
+        """
+        super().__init__()
+        self._rhsm_config_proxy = rhsm_config_proxy
```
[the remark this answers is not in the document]
M4rtinK: Due to issues with the RHSM system unit that provides the RHSM DBus API taking long to start, we have decided to pass all access to the RHSM DBus API through the `RHSMObserver` class instance, which makes sure attempts to get a proixy on 

We decided to pass all RHSM DBus API access via the `RHSMObserver`, which waits for the RHSM service to start before returning a proxy. Otherwise we could get crashes when we attempt to get a proxy for the RHSM DBus objects but the service is still starting.

Due to this we can't just call `get_proxy()` directly and need to go through `RHSMObserver`, which then the task needs to have access to. Rather then passing a reference to the Subscription module or the observer instance it seem like the best option to pass the proxy. This also simplifies testing.

pyanaconda/modules/subscription/subscription.py
M4rtinK: Yeah, I like that suggestion - it makes more sense for a method to have side effects like that rather than for a property. Lets do that. :)

pyanaconda/modules/subscription/runtime.py
M4rtinK: Thanks, fixed! :)

pyanaconda/modules/subscription/runtime.py
poncovka: Ah, ok. But you still have to create the variants. Also `Get` and `GetAll` return variants, but you can just use `get_native` for the return values.

Dasbus doesn't unpack the variants, so the arguments and return values match the signatures of the methods. Pydbus used to unpack the received variants, but still required to provide variants to send. It was a mess.

M4rtinK: OK, thanks for deciphering it for me. :) So I've added `get_native()` over the `GetAll()` call and changed the `SetAll()` invocation to use a `Dict[Str, Str]` variant. I hope that's correct ?

M4rtinK:
jenkins, test this please

M4rtinK:
jenkins, test this please

pyanaconda/modules/subscription/subscription.py
```diff
@@ -509,3 +516,43 @@ def rhsm_observer(self):
...
+        or else the method might return non-default (Anaconda overwritten)
+        data.
+
+        :return : dictionary of default RHSM configuration values
+        :rtype: dict
+        """
+        if self._rhsm_config_defaults is None:
+            # config defaults cache not yet populated, do it now
+            proxy = self.rhsm_observer.get_proxy(RHSM_CONFIG)
+            # turn the variant into a duct with get_native()
```
poncovka, line 542: `a dict`

pyanaconda/modules/subscription/runtime.py
```diff
@@ -0,0 +1,115 @@
...
+                config_dict[key] = value
+            else:
+                # if no value is present in request, use
+                # value from the original RHSM config state
+                # (if any)
+                log.debug("subscription: restoring original value for RHSM config key %s", key)
+                config_dict[key] = self._rhsm_config_defaults.get(key, "")
+
+        # turn the dict into variant
+        config_variant = get_variant(Dict[Str, Str], config_dict)
```
poncovka, line 112: It needs to be a dictionary that maps strings to variants, so you need to create a string variant for every value in the dictionary.

M4rtinK: This seems like something that we could have a generic function for, that would take the values in a Python dict and turn them to appropriate variants - "foo" -> String variant, 1 -> Int variant, etc.

It would be basically the inverse of `get_native()`.

pyanaconda/modules/subscription/subscription.py
M4rtinK: Fixed. :)

pyanaconda/modules/subscription/runtime.py
M4rtinK: Anyway, switched the `SetAll()` call to use a regular dict that has string keys and string variant values - hopefully got it right this time. :P

poncovka: > This seems like something that we could have a generic function for, that would take the values in a Python dict and turn them to appropriate variants - "foo" -> String variant, 1 -> Int variant, etc.

We are not able to do that, because 1 can be also Byte, Int16, UInt16, Int32, UInt32, Int64, UInt64. This is the reason why we have DBus structures.

PR #2468, opened 2020-04-24
Reload NTP server list in dialog on every run

VladimirSlavik:
Fixes weird behavior of removals and changes. Basically the dialog was reloaded only on cancel, but needs it on every show.
    
See also: rhbz#1823129


VladimirSlavik:
Thank you, fixed the commit message, will merge once tests pass.

VladimirSlavik:
Pylint is broken and can't run now. + Previous version was ok. = Merging

VladimirSlavik:
Given the bug was later split, the commit should have read "Resolves" instead of "Related".

poncovka:
> Given the bug was later split, the commit should have read "Resolves" instead of "Related".

@VladimirSlavik, never mind. Just update the bug in bugzilla :-)

PR #2469, opened 2020-04-24
Run %onerror and %traceback scripts for all types of exceptions

poncovka:
Run these kickstart scripts also for exceptions like NonInteractiveError
in the cmdline mode.

### DIFF
--- a/pyanaconda/exception.py
+++ b/pyanaconda/exception.py
@@ -111,12 +111,16 @@ class AnacondaExceptionHandler(ExceptionHandler):
                              "The exact error message is:\n\n%s.\n\n "
                              "The installer will now terminate.") % str(value)
             self.intf.messageWindow(_("Hardware error occurred"), hw_error_msg)
+            self._run_kickstart_scripts(dump_info)
             sys.exit(0)
         elif isinstance(value, UnusableStorageError):
+            self._run_kickstart_scripts(dump_info)
             sys.exit(0)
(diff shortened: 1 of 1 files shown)


pyanaconda/exception.py
```diff
@@ -111,12 +111,16 @@ def _main_loop_handleException(self, dump_info):
                              "The exact error message is:\n\n%s.\n\n "
                              "The installer will now terminate.") % str(value)
             self.intf.messageWindow(_("Hardware error occurred"), hw_error_msg)
+            self._run_kickstart_scripts(dump_info)
```
jkonecny12, line 114: I think we should separate these two lines to a private method.

poncovka: I have created an issue for simplifying the code for exiting the installation. It is a little messy in general.

StykMartin (commented):
Is this change rebased to any current RHEL release?

VladimirSlavik:
@StykMartin git says it's in `rhel-8` which has the RHEL 8.3 rebase:
```
git branch --contains 714e342
* master
  rhel-8
  (...)
```

That's the merge commit, ditto for ac11b78 which is the actual change.

StykMartin:
I believe that https://bugzilla.redhat.com/show_bug.cgi?id=1826644 can be closed then. There is nothing to do more. We should use mode cmdline and hook will take care of rest

PR #2472, opened 2020-04-25
subscription: Add subscription token transfer task

M4rtinK:
The first commit adds a task that restores RHSM log level (so that `rhsm.conf` does not contain the `DEBUG` level setting). The second commit adds the token transfer task, which transfers subscription related tokens from the installation environment to the target system.

### DIFF
--- a/pyanaconda/modules/subscription/installation.py
+++ b/pyanaconda/modules/subscription/installation.py
@@ -17,6 +17,8 @@
 #
 import os
 
+from dasbus.typing import get_variant, Str
+
 from pyanaconda.core import util
 
 from pyanaconda.modules.common.task import Task
@@ -104,3 +106,32 @@ class SystemPurposeConfigurationTask(Task):
             usage=self._system_purpose_data.usage,
             addons=self._system_purpose_data.addons
         )
+
+
+class RestoreRHSMLogLevelTask(Task):
+    """Restore RHSM log level back to INFO."""
+
+    def __init__(self, rhsm_config_proxy):
+        """Create a new task.
+        :param rhsm_config_proxy: DBus proxy for the RHSM Config object
+        """
+        super().__init__()
+        self._rhsm_config_proxy = rhsm_config_proxy
+
+    @property
+    def name(self):
+        return "Restoring subscription manager log level"
+
+    def run(self):
+        """Set RHSM log level back to INFO.
+
+        We previously set the RHSM log level to DEBUG, which is also
+        reflected in rhsm.conf. This would mean RHSM would continue to
+        log in debug mode also on the system once rhsm.conf has been
+        copied over to the target system.
+
+        So set the log level back to INFO before we copy the config file.
+        """
+        log.debug("subscription: setting RHSM log level back to INFO")
+        self._rhsm_config_proxy.Set("logging.default_log_level",
+                                    get_variant(Str, "INFO"), "")
--- a/pyanaconda/modules/subscription/installation.py
+++ b/pyanaconda/modules/subscription/installation.py
@@ -16,6 +16,8 @@
 # Red Hat, Inc.
 #
 import os
+import glob
+import shutil
 
 from dasbus.typing import get_variant, Str
 
@@ -23,7 +25,7 @@ from pyanaconda.core import util
 
 from pyanaconda.modules.common.task import Task
 from pyanaconda.modules.common.errors.installation import InsightsConnectError, \
-    InsightsClientMissingError
+    InsightsClientMissingError, SubscriptionTokenTransferError
 
 from pyanaconda.modules.subscription import system_purpose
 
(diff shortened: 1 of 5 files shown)


pyanaconda/modules/subscription/installation.py
```diff
@@ -135,3 +137,138 @@ def run(self):
...
+        # transfer all the pem files in the input folder
+        for pem_file_path in glob.glob(os.path.join(input_folder, "*.pem")):
+            shutil.copy(pem_file_path, output_folder)
+        # if we got this far the pem copy operation was a success
+        return True
+
+    def _copy_file_to_path(self, file_path, target_file_path):
+        if not os.path.isfile(file_path):
+            return False
+        if not os.path.isdir(os.path.dirname(target_file_path)):
```
jkonecny12, line 199: You can remove this check based on our documentation `util.mkdirChain` should not fail if the target already exists.

pyanaconda/modules/subscription/installation.py
```diff
@@ -135,3 +137,138 @@ def run(self):
...
+        # create the output folder path if it does not exist
+        if not os.path.isdir(output_folder):
+            util.mkdirChain(output_folder)
+        # transfer all the pem files in the input folder
+        for pem_file_path in glob.glob(os.path.join(input_folder, "*.pem")):
+            shutil.copy(pem_file_path, output_folder)
+        # if we got this far the pem copy operation was a success
+        return True
+
+    def _copy_file_to_path(self, file_path, target_file_path):
```
jkonecny12, line 196: I think `_copy_file` should be sufficient name.

pyanaconda/modules/subscription/installation.py
```diff
@@ -135,3 +137,138 @@ def run(self):
...
+        # - this might be needed even if the system has not been subscribed
+        #   during the installation and is therefore always attempted
+        # - this means the syspurpose tool has been called in the installation
+        #   environment & we need to transfer the results to the target system
+        if os.path.exists(self.RHSM_SYSPURPOSE_FILE_PATH):
+            log.debug("subscription: transferring syspurpose file")
+            target_syspurpose_file_path = self._sysroot + self.RHSM_SYSPURPOSE_FILE_PATH
+            self._copy_file_to_path(self.RHSM_SYSPURPOSE_FILE_PATH, target_syspurpose_file_path)
+
+    def _transfer_rhsm_config_file(self):
```
jkonecny12, line 215: These method can be one method with just two or three arguments.

M4rtinK:
@jkonecny12 
- restoring the log level was part of the token transfer task on RHEL8 and IMHO that was wrong, it does not really have anything in common with token transfer and should not be part of the token transfer task & would complicate testing
- I agree there is similarity between the test methods, I think it's not enough to warrant adding helper function at the cost of test readability

pyanaconda/modules/subscription/installation.py
M4rtinK: Yeah, lets do that. :)

pyanaconda/modules/subscription/installation.py
M4rtinK: OK, why not. :)

M4rtinK:
Looks like using the common `_transfer_file()` method actually reduced the number of similar test methods. :)

pyanaconda/modules/subscription/installation.py
M4rtinK: Cool, makes the code simpler. :)

pyanaconda/modules/subscription/subscription.py
```diff
@@ -488,9 +488,26 @@ def set_subscription_attached(self, system_subscription_attached):
     def install_with_tasks(self):
         """Return the installation tasks of this module.
 
+        Order of execution is important:
```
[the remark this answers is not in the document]
M4rtinK: Yes, it is expected installation tasks will be run in series and in the order they are returned.

Actually due to the rudimentary introspection options we have at the moment one would kinda have to go out of their way to find out which task is which in the returned list before deciding about them. This is also one of the reasons why some our our tasks take a boolean option and don't do anything if the option is False.

pyanaconda/modules/subscription/installation.py
```diff
@@ -104,3 +108,150 @@ def run(self):
...
+        if not self._copy_file(target_path, target_repo_file_path):
+            msg = "{} ({}) is missing".format(target_name, self.RHSM_REPO_FILE_PATH)
+            raise SubscriptionTokenTransferError(msg)
+
+    def _transfer_system_purpose(self):
+        # transfer the system purpose file if present
+        # - this might be needed even if the system has not been subscribed
+        #   during the installation and is therefore always attempted
+        # - this means the syspurpose tool has been called in the installation
+        #   environment & we need to transfer the results to the target system
```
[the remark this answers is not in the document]
M4rtinK: Good point, let's do that.

pyanaconda/modules/subscription/installation.py
```diff
@@ -104,3 +108,150 @@ def run(self):
...
+            return False
+        # make sure the output folder exist
+        util.mkdirChain(output_folder)
+        # transfer all the pem files in the input folder
+        for pem_file_path in glob.glob(os.path.join(input_folder, "*.pem")):
+            shutil.copy(pem_file_path, output_folder)
+        # if we got this far the pem copy operation was a success
+        return True
+
+    def _copy_file(self, file_path, target_file_path):
```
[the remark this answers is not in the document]
M4rtinK: What the task does is copying a bunch of files from the installation environment to the target system in similar yet slightly different ways.

So there could either be a whole bunch of pretty similar pieces of code, doing the slightly different things for each file/bunch of files, which would likely not look very good and would be cumbersome to unit test.

So instead an effort has been made to extract the similar pieces to private helper methods to reduce code duplication & make the task easier to test.

For the individual methods:

`_copy_pem_files()` - this copies `*.pem` files related to the attached subscription and needs to handle extension based filtering (so that only `pem` files are transferred), the input folder not existing as well as making sure the output folder exists

`_copy_file()` - general file copy function that checks if the target is file and makes sure the target folder exists

`_transfer_file()` - nice wrapper for `_copy_file()` with logging and exception raising based on what is being transferred

`_ transfer_system_purpose()` - wrapper for `_copy_file()` with system purpose specifics, mainly due to missing system purpose data not being a fatal error unlike ale the other files

`_transfer_entitlement_keys()` - wrapper for `_copy_pem_files()` with logging and exception raising

You could unify some of these methods (eq. copy-past _copy_file() in place to `_transfer_file()` and `_transfer_system purpose()` and roll in the pem file transfer method to `_transfer_entitlement_keys()`) but that would IMHO be at the cost of complexity and harder unit testing.

tests/nosetests/pyanaconda_tests/module_subscription_tests.py
```diff
@@ -1042,14 +1042,31 @@ def set_rhsm_config_with_task_test(self, publisher):
     @patch_dbus_publish_object
     def install_with_tasks_default_test(self, publisher):
         """Test install tasks - Subscription module in default state."""
+        # mock the rhsm config proxy
+        observer = Mock()
+        observer.get_proxy = Mock()
+        self.subscription_module._rhsm_observer = observer
+        config_proxy = Mock()
+        observer.get_proxy.return_value = config_proxy
```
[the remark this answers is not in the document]
M4rtinK: Looks like a nice improvement, but I'll do that as a separate PR given that I can indeed also improve the test above, which is not a part of this PR.

M4rtinK:
@VladimirSlavik 
End to end testing is definitely planned via kickstart tests once all is in place. That should test everything working at once complete with interaction with real RHSM service.

As for more complete testing as part of unit tests, there will be a registration/unregistration management code running in a thread similar [to the one on RHEL8] (https://github.com/rhinstaller/anaconda/blob/d168dcaf23bf67939d230cfd3f5b62d9c6ea807b/pyanaconda/subscription.py#L71) that will orchestrate the tasks as needed during a registration/unregistration attempt. This management code than can be unit tested to verify it runs the  (mocked) tasks in correct order and with correct inputs and correctly handles their (mocked) outputs. This together with tests for the individual tasks should provide sufficient test coverage without duplicating a lot of the tests during end-to-end testing.

VladimirSlavik:
Nice, thanks for the explanation!

pyanaconda/modules/subscription/installation.py
```diff
@@ -104,3 +108,155 @@ def run(self):
...
+
+        # the other subscription tokens are only relevant if the system has been subscribed
+        if not self._transfer_subscription_tokens:
+            log.debug("subscription: transfer of subscription tokens not requested")
+            return
+
+        # transfer entitlement keys
+        self._transfer_entitlement_keys()
+
+        # transfer the consumer key
```
jkonecny12, line 252: I don't think you need these comments but I don't mind to have it them here.

PR #2475, opened 2020-04-27
subscription: Add registration tasks

M4rtinK:
Add the username+password and organization+activation key(s) registration
tasks. Also add helper methods needed to active a private DBus session
that is needed for safe credential passing to RHSM.

For success/error reporting the tasks return to value tuple with the
first value indicating success/failure via a boolean and the second
value contains the error message, which is an empty string for success.

### DIFF
--- a/pyanaconda/modules/subscription/runtime.py
+++ b/pyanaconda/modules/subscription/runtime.py
@@ -15,12 +15,56 @@
 # License and may only be used or replicated with the express permission of
 # Red Hat, Inc.
 #
+import os
+import json
+
 from dasbus.typing import get_variant, Str
+from dasbus.connection import MessageBus
+from dasbus.error import DBusError
+
+from pyanaconda.core.i18n import _
 
 from pyanaconda.modules.common.task import Task
+from pyanaconda.modules.common.constants.services import RHSM
+from pyanaconda.modules.common.constants.objects import RHSM_REGISTER
+from pyanaconda.modules.common.errors.subscription import RegistrationError
 
 from pyanaconda.anaconda_loggers import get_module_logger
 log = get_module_logger(__name__)
+
+
+class RHSMPrivateBus(MessageBus):
+    """Representation of RHSM private bus connection that can be used as a context manager."""
+
+    def __init__(self, rhsm_register_server_proxy, *args, **kwargs):
+        """Representation of RHSM private bus connection that can be used as a context manager.
+
+        :param rhsm_register_server_proxy: DBus proxy for the RHSM RegisterServer object
+        """
+        super().__init__(*args, **kwargs)
+        self._rhsm_register_server_proxy = rhsm_register_server_proxy
+        self._private_bus_address = None
+
+    def __enter__(self):
+        log.debug("subscription: starting RHSM private DBus session")
+        locale = os.environ.get("LANG", "")
+        self._private_bus_address = self._rhsm_register_server_proxy.Start(locale)
+        log.debug("subscription: RHSM private DBus session has been started")
+        return self
+
+    def __exit__(self, exc_type, exc_value, exc_traceback):
+        log.debug("subscription: shutting down the RHSM private DBus session")
+        self.connection.disconnect()
+        locale = os.environ.get("LANG", "")
+        self._rhsm_register_server_proxy.Stop(locale)
+        log.debug("subscription: RHSM private DBus session has been shutdown")
+
+    def _get_connection(self):
+        """Get a connection to RHSM private DBus session."""
+        # the RHSM private bus address is potentially sensitive
+        # so we will not log it
+        log.info("Connecting to the RHSM private DBus session.")
+        return self._provider.get_addressed_bus_connection(self._private_bus_address)
 
 
 class SetRHSMConfigurationTask(Task):
@@ -114,3 +158,106 @@ class SetRHSMConfigurationTask(Task):
 
         # and finally set the dict to RHSM via the DBus API
         self._rhsm_config_proxy.SetAll(config_dict, "")
+
+
+class RegisterWithUsernamePasswordTask(Task):
+    """Register the system via username + password."""
+
+    def __init__(self, rhsm_register_server_proxy, username, password):
+        """Create a new registration task.
+
+        It is assumed the username and password have been
+        validated before this task has been started.
+
+        :param rhsm_register_server_proxy: DBus proxy for the RHSM RegisterServer object
+        :param str username: Red Hat account username
+        :param str password: Red Hat account password
+        """
+        super().__init__()
+        self._rhsm_register_server_proxy = rhsm_register_server_proxy
+        self._username = username
+        self._password = password
+
+    @property
+    def name(self):
+        return "Register with Red Hat account username and password"
+
+    def run(self):
+        """Register the system with Red Hat account username and password.
+
+        :return: True on success, raises RegistrationError on failure
+        :rtype: bool
+        :raises: RegistrationError
+        """
+        log.debug("subscription: registering with username and password")
+        with RHSMPrivateBus(self._rhsm_register_server_proxy) as private_bus:
+            try:
+                locale = os.environ.get("LANG", "")
+                private_register_proxy = private_bus.get_proxy(RHSM.service_name,
+                                                               RHSM_REGISTER.object_path)
+                # We do not yet support setting organization for username & password
+                # registration, so organization is blank for now.
+                organization = ""
+                private_register_proxy.Register(organization,
+                                                self._username,
+                                                self._password,
+                                                {},
+                                                {},
+                                                locale)
+                log.debug("subscription: registered with username and password")
+            except DBusError as e:
+                log.debug("subscription: failed to register with username and password: %s",
+                          str(e))
+                # RHSM exception contain details as JSON due to DBus exception handling limitations
+                exception_dict = json.loads(str(e))
+                # return a generic error message in case the RHSM provided error message is missing
+                message = exception_dict.get("message", _("Registration failed."))
+                raise RegistrationError(message) from None
+
+
+class RegisterWithOrganizationKeyTask(Task):
+    """Register the system via organization and one or more activation keys."""
+
+    def __init__(self, rhsm_register_server_proxy, organization, activation_keys):
+        """Create a new registration task.
+
+        :param rhsm_register_server_proxy: DBus proxy for the RHSM RegisterServer object
+        :param str organization: organization name for subscription purposes
+        :param activation keys: activation keys
+        :type activation_keys: list of str
+        """
+        super().__init__()
+        self._rhsm_register_server_proxy = rhsm_register_server_proxy
+        self._organization = organization
+        self._activation_keys = activation_keys
+
+    @property
+    def name(self):
+        return "Register with organization name and activation key"
+
+    def run(self):
+        """Register the system with organization name and activation key.
+
+        :return: True on success, raises RegistrationError on failure
+        :rtype: bool
+        :raises: RegistrationError
+        """
+        log.debug("subscription: registering with organization and activation key")
+        with RHSMPrivateBus(self._rhsm_register_server_proxy) as private_bus:
+            try:
+                locale = os.environ.get("LANG", "")
+                private_register_proxy = private_bus.get_proxy(RHSM.service_name,
+                                                               RHSM_REGISTER.object_path)
+                private_register_proxy.RegisterWithActivationKeys(self._organization,
+                                                                  self._activation_keys,
+                                                                  {},
+                                                                  {},
+                                                                  locale)
+                log.debug("subscription: registered with organization and activation key")
+            except DBusError as e:
+                log.debug("subscription: failed to register with organization & key: %s", str(e))
+                # RHSM exception contain details as JSON due to DBus exception handling limitations
+                exception_dict = json.loads(str(e))
+                # return a generic error message in case the RHSM provided error message is missing
+                message = exception_dict.get("message", _("Registration failed."))
+                raise RegistrationError(message) from None
--- a/tests/nosetests/pyanaconda_tests/module_subscription_tasks_tests.py
+++ b/tests/nosetests/pyanaconda_tests/module_subscription_tasks_tests.py
@@ -24,19 +24,24 @@ from unittest.mock import patch, Mock, call
 import tempfile
 
 from dasbus.typing import get_variant, Str
+from dasbus.error import DBusError
 
 from pyanaconda.core import util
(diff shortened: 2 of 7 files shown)


VladimirSlavik:
Jenkins, test this please.

pyanaconda/modules/subscription/runtime.py
```diff
@@ -114,3 +167,121 @@ def run(self):
...
+            exception_dict = json.loads(str(e))
+            # return a generic error message in case the RHSM provided error message is missing
+            message = exception_dict.get("message", _("Registration failed."))
+            error_message = message
+        finally:
+            # registration done (one way or another), so shutdown the private bus used for
+            # credential passing
+            stop_rhsm_private_bus(self._rhsm_register_server_proxy)
+        # return the result
+        return success, error_message
```
poncovka, line 231: Why don't you raise an exception?

pyanaconda/modules/subscription/runtime.py
```diff
@@ -15,14 +15,67 @@
...
 
 from pyanaconda.modules.common.task import Task
+from pyanaconda.modules.common.constants.services import RHSM
+from pyanaconda.modules.common.constants.objects import RHSM_REGISTER
 
 from pyanaconda.anaconda_loggers import get_module_logger
 log = get_module_logger(__name__)
 
 
+def start_rhsm_private_bus(rhsm_register_server_proxy, message_bus_class=AddressedMessageBus):
```
poncovka, line 37: I would consider to create a subclass of `MessageBus` and move this code to the method `_get_connection` (see the implementation of `AnacondaMessageBus`). Also, I would use a context manager instead of these start and stop functions. Ideally, the code in the tasks could look like this:

```
with RHSMPrivateBus() as bus:
    private_register_proxy = bus.get_proxy(...)
    ...

```

pyanaconda/modules/subscription/runtime.py
```diff
@@ -15,14 +15,67 @@
...
+    The RHSM private DBus session is used for secure credential passing.
+    We are only enabling the private bus for credential passing and
+    shut it down once the credentials have been transferred.
+
+    :param rhsm_register_server_proxy: DBus proxy for the RHSM RegisterServer object
+    """
+
+    log.debug("subscription: shutting down the RHSM private DBus session")
+    locale = os.environ.get("LANG", "")
+    rhsm_register_server_proxy.Stop(locale)
```
poncovka, line 75: I would recommend to call the `disconnect` method of the message bus if possible. It is not really necessary, just a nice thing to do.

pyanaconda/modules/subscription/runtime.py
```diff
@@ -114,3 +167,121 @@ def run(self):
...
+            exception_dict = json.loads(str(e))
+            # return a generic error message in case the RHSM provided error message is missing
+            message = exception_dict.get("message", _("Registration failed."))
+            error_message = message
+        finally:
+            # registration done (one way or another), so shutdown the private bus used for
+            # credential passing
+            stop_rhsm_private_bus(self._rhsm_register_server_proxy)
+        # return the result
+        return success, error_message
```
poncovka, line 287: I would prefer an exception.

M4rtinK:
@jkonecny12 The private DBus session is needed for secure credential passing where normally other clients on the system and session bus could could eavesdrop and capture the sensitive credentials. For the installation environment this is less of an issue, but could still be problematic in cases Anaconda runs as a process on a normal like dir installs and similar.

In any case, this is the design of credential passing the subscription manager team decided to use & tha API we have available. As for starting and stopping the bus, from what I remember there are two main reasons for stopping it after use:
- this seems to be the expected behavior from the documentation and RHSM use cases
- if the RHSM service gets restarted, this also kills the private bus, so holding reference to it for longer than necessary is dangerous

pyanaconda/modules/subscription/runtime.py
M4rtinK: I was not sure we can safely survive an exception being raised in the task, so opted for the tuple for now. The thing is that the registration task will fail quite often in practice due to things such as typos in usernames and passwords & Anaconda needs to be able to survive that, notify user & try again.

If it is not a problem to safely catch the exception (mostly likely from the `run_sync()` method that runs the task in the registration handling thread in the UI) the we should definitely use an exception for this. :)

pyanaconda/modules/subscription/runtime.py
M4rtinK: Sounds good, lets do that. :)

tests/nosetests/pyanaconda_tests/module_subscription_tasks_tests.py
```diff
@@ -241,3 +246,182 @@ def set_rhsm_config_tast_restore_default_value_test(self):
...
+
+    @patch("os.environ.get")
+    @patch("pyanaconda.modules.subscription.runtime.stop_rhsm_private_bus")
+    @patch("pyanaconda.modules.subscription.runtime.start_rhsm_private_bus")
+    def username_password_failure_test(self, start_private_bus, stop_private_bus, environ_get):
+        """Test the RegisterWithUsernamePasswordTask - failure."""
+        # register server proxy
+        register_server_proxy = Mock()
+        # private register proxy
+        private_register_proxy = start_private_bus.return_value
```
VladimirSlavik, line 335: I suspect this in fact reads from the mock API (`return_value`) a thing that is unset, and so evaluates to just another `Mock()`. That would mean the line is misleading, as it implies some relationship where there is none. But I am not 100% sure.

The pattern is repeated everywhere.

tests/nosetests/pyanaconda_tests/module_subscription_tests.py
```diff
@@ -1039,6 +1040,76 @@ def set_rhsm_config_with_task_test(self, publisher):
...
+
+    @patch_dbus_publish_object
+    def register_with_organization_key_test(self, publisher):
+        """Test RegisterWithOrganizationKeyTask creation."""
+        # prepare the module with dummy data
+        full_request = SubscriptionRequest()
+        full_request.type = SUBSCRIPTION_REQUEST_TYPE_ORG_KEY
+        full_request.organization = "123456789"
+        full_request.account_username = "foo_user"
+        full_request.server_hostname = "candlepin.foo.com"
```
VladimirSlavik, line 1086: Not sure if applicable, so just a note - foo.com does exist.

pyanaconda/modules/subscription/runtime.py
M4rtinK: Good point, added.

pyanaconda/modules/subscription/runtime.py
M4rtinK: Implemented a `RHSMPrivateBus` that inherits `MessageBus` and is also a context manager. The only really difference from the suggestion is that the constructor takes a single argument, which is the RHSM registration server proxy (acquired via the `RHSMObserver` instance).

pyanaconda/modules/subscription/runtime.py
```diff
@@ -114,3 +167,121 @@ def run(self):
...
+    def name(self):
+        return "Register with Red Hat account username and password"
+
+    def run(self):
+        """Register the system with Red Hat account username and password.
+
+        :return: a (success/failure, error_message) tuple
+        :rtype: a (bool, str) tuple
+        """
+        log.debug("subscription: connecting to RHSM private bus")
```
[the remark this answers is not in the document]
M4rtinK: Indeed, cleaned it up. :)

tests/nosetests/pyanaconda_tests/module_subscription_tasks_tests.py
```diff
@@ -241,3 +246,182 @@ def set_rhsm_config_tast_restore_default_value_test(self):
...
+        # check the result is the mock proxy we expect
+        self.assertEqual(result, connection.get_proxy.return_value)
+
+    @patch("os.environ.get")
+    def stop_rhsm_private_bus_test(self, environ_get):
+        """Test the stop_rhsm_private_bus method."""
+        # mock the register server proxy
+        register_server_proxy = Mock()
+        # make sure locate is always the same at test time
+        environ_get.return_value = "en_US.UTF-8"
```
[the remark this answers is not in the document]
M4rtinK: Yeah, that's indeed cleaner. In some cases it could hit our 99 character line length limit, but in this case it is fine. :)

tests/nosetests/pyanaconda_tests/module_subscription_tasks_tests.py
M4rtinK: Yeah, I have been looking a bit more into how the `Mock` objects work and it really looks like `return_value` is always another Mock instance that gets cached and stays the same after fist access. So getting the mock instance from `return_value` should be equivalent to setting your own.

Anyway, cleaned up that part when changing the unit tests to work with `RHSMPrivateBus`.

tests/nosetests/pyanaconda_tests/module_subscription_tests.py
M4rtinK: Looking at it example.com also resolves and returns a web site (noting that it is an example web site). Anyway, as no network access is being done here I think we are fine with the ones we have there.

rvykydal:
I didn't find anything to add. Looks good to me.

pyanaconda/modules/subscription/runtime.py
poncovka: It is definitely safe, we use it a lot :-) The task's exception is propagated via DBus when we call it's `Finish` method in the function `sync_run_task`. So you can safely catch it on the client side when you call  `sync_run_task`.

pyanaconda/modules/subscription/runtime.py
```diff
@@ -15,14 +15,57 @@
...
 from pyanaconda.modules.common.task import Task
+from pyanaconda.modules.common.constants.services import RHSM
+from pyanaconda.modules.common.constants.objects import RHSM_REGISTER
 
 from pyanaconda.anaconda_loggers import get_module_logger
 log = get_module_logger(__name__)
 
 
+class RHSMPrivateBus(MessageBus):
+    """Representation of an Anaconda bus connection."""
```
poncovka, line 36: Please, fix the docstring.

pyanaconda/modules/subscription/runtime.py
M4rtinK: Sounds good, I'll switch the tasks to return exceptions then. :)

pyanaconda/modules/subscription/runtime.py
M4rtinK: Oops, fixed. :)

pyanaconda/modules/subscription/runtime.py
M4rtinK: `RegistrationError` is now raised. :)

M4rtinK:
@poncovka Tasks now raise `RegistrationError` when they fail. :)

tests/nosetests/pyanaconda_tests/module_subscription_tasks_tests.py
```diff
@@ -241,3 +246,149 @@ def set_rhsm_config_tast_restore_default_value_test(self):
...
+                                                                "en_US.UTF-8")
+
+    @patch("os.environ.get", return_value="en_US.UTF-8")
+    @patch("pyanaconda.modules.subscription.runtime.RHSMPrivateBus")
+    def username_password_failure_test(self, private_bus, environ_get):
+        """Test the RegisterWithUsernamePasswordTask - failure."""
+        # register server proxy
+        register_server_proxy = Mock()
+        # private register proxy
+        get_proxy = private_bus.return_value.__enter__.return_value.get_proxy
```
VladimirSlavik, line 326: I don't think this is how you are supposed to do it. But maybe I should stop giving you "ideas" ;-)

pyanaconda/modules/subscription/runtime.py
```diff
@@ -114,3 +158,110 @@ def run(self):
...
+            except DBusError as e:
+                log.debug("subscription: failed to register with username and password: %s",
+                          str(e))
+                # RHSM exception contain details as JSON due to DBus exception handling limitations
+                exception_dict = json.loads(str(e))
+                # return a generic error message in case the RHSM provided error message is missing
+                message = exception_dict.get("message", _("Registration failed."))
+                raise RegistrationError(message)
+        # all good
+        return True
```
poncovka, line 217: It looks like a leftover.

pyanaconda/modules/subscription/runtime.py
```diff
@@ -114,3 +158,110 @@ def run(self):
...
+                                                locale)
+                log.debug("subscription: registered with username and password")
+            except DBusError as e:
+                log.debug("subscription: failed to register with username and password: %s",
+                          str(e))
+                # RHSM exception contain details as JSON due to DBus exception handling limitations
+                exception_dict = json.loads(str(e))
+                # return a generic error message in case the RHSM provided error message is missing
+                message = exception_dict.get("message", _("Registration failed."))
+                raise RegistrationError(message)
```
poncovka, line 215: Please use `raise RegistrationError(message) from None`.

pyanaconda/modules/subscription/runtime.py
```diff
@@ -114,3 +158,110 @@ def run(self):
...
+                                                                  {},
+                                                                  locale)
+                log.debug("subscription: registered with organization and activation key")
+            except DBusError as e:
+                log.debug("subscription: failed to register with organization & key: %s", str(e))
+                # RHSM exception contain details as JSON due to DBus exception handling limitations
+                exception_dict = json.loads(str(e))
+                # return a generic error message in case the RHSM provided error message is missing
+                message = exception_dict.get("message", _("Registration failed."))
+                raise RegistrationError(message)
```
poncovka, line 265: Same here.

pyanaconda/modules/subscription/runtime.py
```diff
@@ -114,3 +158,110 @@ def run(self):
...
+        # all good
+        return True
```
poncovka, line 267: Same here.

M4rtinK: Removed.

pyanaconda/modules/subscription/runtime.py
M4rtinK: Removed.

pyanaconda/modules/subscription/runtime.py
M4rtinK: Interesting, haven't known this. Let's do it. :)

pyanaconda/modules/subscription/runtime.py
M4rtinK: Changed.

PR #2477, opened 2020-04-27
Fix the error reporting from ReadKickstart

poncovka:
Change the type of the attribute `line_number` in the class `KickstartMessage` to
`UInt32` and change the default value to zero. Set the line number in the method
`for_error` method to zero if it is not set in the given exception.

Warning messages returned by the DBus method `ReadKickstart` shouldn't contain
meta data about caught warnings.

Test propagation of errors and warnings from the DBus method `ReadKickstart`.

### DIFF
--- a/tests/nosetests/pyanaconda_tests/module_base_test.py
+++ b/tests/nosetests/pyanaconda_tests/module_base_test.py
@@ -1,0 +1,112 @@
+#
+# Copyright (C) 2020  Red Hat, Inc.
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# the GNU General Public License v.2, or (at your option) any later version.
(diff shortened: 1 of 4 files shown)


tests/nosetests/pyanaconda_tests/module_base_test.py
```diff
@@ -0,0 +1,113 @@
...
+            'message': get_variant(Str, 'Second warning!'),
+            'module-name': get_variant(Str, ""),
+        }
+
+        report = {
+            'error-messages': get_variant(List[Structure], []),
+            'warning-messages': get_variant(List[Structure], [first_warning, second_warning])
+        }
+
+        print(interface.ReadKickstart(""))
```
jkonecny12, line 112: Do you really want a `print` here?

poncovka: Removed. Good catch :-)

PR #2478, opened 2020-04-27
Separate the "Reset All" button from the edge of the screen

jstodola:
This change makes the "Reset All" button on the custom partitioning screen aligned with the "Help!" button and increases the user experience by 2%.
It is worth mentioning that this change slightly reduces the usable area of the screen, compare the attached screenshots.

Before:
before
After:
after

### DIFF
--- a/pyanaconda/ui/gui/spokes/custom_storage.glade
+++ b/pyanaconda/ui/gui/spokes/custom_storage.glade
@@ -118,6 +118,7 @@
             <property name="vexpand">True</property>
             <property name="yalign">0</property>
             <property name="left_padding">24</property>
+            <property name="right_padding">6</property>
             <child internal-child="action_area">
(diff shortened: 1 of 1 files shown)


VladimirSlavik:
Just a side note - others will correct me, but IIRC we are aiming for default spacing of 4. That it's 6 somewhere means things are not entirely ideal :)

VladimirSlavik:
After checking other dialogs, 6 is the right value, but maybe I'd put it on the `grid2` thingy instead of just the button - that will work equally well, even for any future changes?

VladimirSlavik:

jstodola:
The padding is added to `AnacondaSpokeWindow-alignment1`, so it's not just the button (right, the title is misleading a bit).
I can add padding to `grid2` instead, it should not be a problem.

VladimirSlavik:
Works as expected. Thank you!

PR #2480, opened 2020-04-29
Show warning message when entered size is not valid (#1809573)

VladimirSlavik:
It looks like this:
desired_capacity_warning

~~Suggestions for the warning text are welcome!~~ Screenshot no longer reflects the actual text.

### DIFF
--- a/pyanaconda/ui/gui/spokes/custom_storage.py
+++ b/pyanaconda/ui/gui/spokes/custom_storage.py
@@ -63,7 +63,8 @@ from pyanaconda.ui.gui.spokes.lib.custom_storage_helpers import get_size_from_en
     get_selected_raid_level, get_default_raid_level, get_container_type, AddDialog,\
     ConfirmDeleteDialog, DisksDialog, ContainerDialog, NOTEBOOK_LABEL_PAGE, NOTEBOOK_DETAILS_PAGE,\
     NOTEBOOK_LUKS_PAGE, NOTEBOOK_UNEDITABLE_PAGE, NOTEBOOK_INCOMPLETE_PAGE, NEW_CONTAINER_TEXT,\
-    CONTAINER_TOOLTIP, get_supported_device_raid_levels, generate_request_description
+    CONTAINER_TOOLTIP, DESIRED_CAPACITY_ERROR, get_supported_device_raid_levels, \
+    generate_request_description
 from pyanaconda.ui.gui.spokes.lib.passphrase import PassphraseDialog
 from pyanaconda.ui.gui.spokes.lib.refresh import RefreshDialog
 from pyanaconda.ui.gui.spokes.lib.summary import ActionSummaryDialog
@@ -1552,14 +1553,10 @@ class CustomPartitioningSpoke(NormalSpoke, StorageCheckHandler):
         self._request.device_raid_level = get_selected_raid_level(self._raidLevelCombo)
         self.on_value_changed()
 
-    def on_size_changed(self, widget):
+    @timed_action(750, 1500, False)
+    def on_size_changed(self, *args):
+        """Callback for text change in "desired capacity" widget"""
         if not self._sizeEntry.get_sensitive():
-            return
-
-        current_size = Size(self._request.device_size)
-        displayed_size = current_size.human_readable(max_places=self.MAX_SIZE_PLACES)
-
-        if displayed_size == self._sizeEntry.get_text():
             return
 
         size = get_size_from_entry(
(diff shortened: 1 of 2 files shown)


pyanaconda/ui/gui/spokes/custom_storage.py
```diff
@@ -1569,22 +1569,28 @@ def on_raid_level_changed(self, widget):
...
             units=SIZE_UNITS_DEFAULT
         )
 
         if size is None:
+            self.set_warning(_(
+              "Values for Desired Capacity can contain only numbers, decimal separator, and unit."
+            ))
+            return
+        else:
+            self.clear_info()
```
poncovka, line 1588: I think that this should be called before `set_warning` as well.

VladimirSlavik: That would animate the bar at every keystroke :) I changed it a bit though, so it clears only if the warning is the same.

It might be better to actually validate the value only on applying with `on_configure_clicked`, but I think this requires less change to the overall structure.

pyanaconda/ui/gui/spokes/custom_storage.py
```diff
@@ -106,6 +106,17 @@ class CustomPartitioningSpoke(NormalSpoke, StorageCheckHandler):
     # If the user enters a smaller size, the GUI changes it to this value
     MIN_SIZE_ENTRY = Size("1 MiB")
 
+    DESIRED_CAPACITY_ERROR = N_(
+        "Specify the Desired Capacity in whole or decimal numbers, with an appropriate unit.\n\n"
+        "Spaces separating digit groups are not allowed. Units consist of a decimal or binary "
+        "prefix, and optionally the letter B. Letter case does not matter for units. The default "
+        "unit used when units are left out is MiB.\n\n"
```
jkonecny12, line 113: I would change the last sentence to:
``The default unit is MiB if no letter is specified.``

Seems easier to understand to me.

pyanaconda/ui/gui/spokes/custom_storage.py
```diff
@@ -106,6 +106,17 @@ class CustomPartitioningSpoke(NormalSpoke, StorageCheckHandler):
     # If the user enters a smaller size, the GUI changes it to this value
     MIN_SIZE_ENTRY = Size("1 MiB")
 
+    DESIRED_CAPACITY_ERROR = N_(
+        "Specify the Desired Capacity in whole or decimal numbers, with an appropriate unit.\n\n"
```
jkonecny12, line 110: I would simplify this:
```
Specify the Desired Capacity in numbers, with an appropriate unit letter.\n\n
```

pyanaconda/ui/gui/spokes/custom_storage.py
```diff
@@ -106,6 +106,17 @@ class CustomPartitioningSpoke(NormalSpoke, StorageCheckHandler):
...
+        "Spaces separating digit groups are not allowed. Units consist of a decimal or binary "
+        "prefix, and optionally the letter B. Letter case does not matter for units. The default "
```
jkonecny12, line 112:
```
... and optionally the letter (B, M, MiB, G, GiB...).
```

pyanaconda/ui/gui/spokes/custom_storage.py
```diff
@@ -106,6 +106,17 @@ class CustomPartitioningSpoke(NormalSpoke, StorageCheckHandler):
     # If the user enters a smaller size, the GUI changes it to this value
     MIN_SIZE_ENTRY = Size("1 MiB")
 
+    DESIRED_CAPACITY_ERROR = N_(
```
poncovka, line 109: Maybe I would move the constant to `custom_storage_helpers`.

pyanaconda/ui/gui/spokes/custom_storage.py
poncovka: Good point. I think that is the reason why we use `show_warning_message` for password entries.

pyanaconda/ui/gui/spokes/custom_storage.py
```diff
@@ -1569,22 +1580,35 @@ def on_raid_level_changed(self, widget):
...
             lower_bound=self.MIN_SIZE_ENTRY,
             units=SIZE_UNITS_DEFAULT
         )
 
+        # Show warning if the size string is invalid. Field self._error is used as a "flag" that
+        # the last error was the same. This is done because this warning can fire on every change,
+        # so it would keep flickering at the bottom as you type.
         if size is None:
+            if self._error != self.DESIRED_CAPACITY_ERROR:
+                self.clear_info()
```
poncovka, line 1598: Btw. this will not clear `self._error`. However, I have no idea what could be the consequences.

pyanaconda/ui/gui/spokes/custom_storage.py
```diff
@@ -1569,22 +1580,35 @@ def on_raid_level_changed(self, widget):
...
+                self.set_detailed_warning(
```
poncovka, line 1599: This solution is somehow less readable. I guess we need to somehow combine `show_warning_message` with `set_detailed_warning`.

VladimirSlavik:
@poncovka, just to be clear, you are not asking to rewrite the mechanism as part of this, correct?

poncovka:
@VladimirSlavik, no, I was just complaining about it ;-)

VladimirSlavik:
Ah, oops, I misread that and put this aside for later as it seemed too large a project at the time.

VladimirSlavik:
Jenkins, test this please.

VladimirSlavik:
Rebased to master as of now.

This is somehow fundamentally broken. Sometimes the changes take and give me the warning, sometimes not, sometimes the thing starts to behave differently after playing with the input field for a while. I'm not sure what to do about this (yet).

pyanaconda/ui/gui/spokes/custom_storage.py
VladimirSlavik: Oh, I know now! It will not do what I want. Fixed.

VladimirSlavik:
Update: this now works as expected. The silly dances with checking `_error` are sadly needed. `set_detailed_warning` should replace the previous error and not animate if the new one is the same. But the replacing does not happen, so I have to clear `_error` manually, and then the animation will always play since `_error` is always empty and thus different. So I think this is the minimal version.

poncovka (approved):
It looks great! Thanks!

poncovka:
Do you want the change in Fedora 33? You should edit the branch in that case.

VladimirSlavik:
Of course I'd like to get it out faster. But this is a sort-of-RFE, so I'd rather not upset the stable release just for that.

PR #2482, opened 2020-04-30
network: don't try to use DeviceConfigurations on live cd (#1827999)

rvykydal:
On LiveDVD the network configuration is done by the the environment tools, so
anaconda does not track its network device configurations, therefore they can't
be used when generating dracut network arguments for installations on network
storage (eg iSCSI). Let's use NM api to find the right configuration if
network module DeviceConfigurations are not available.

### DIFF
--- a/pyanaconda/modules/network/network.py
+++ b/pyanaconda/modules/network/network.py
@@ -637,18 +637,26 @@ class NetworkService(KickstartService):
             log.error("Get dracut arguments for %s: device not found", iface)
             return dracut_args
 
-        connections = self.nm_client.get_connections()
-
         target_connections = []
         if ibft:
-            target_connections = [con for con in connections if is_ibft_connection(con)]
+            target_connections = [con for con in self.nm_client.get_connections()
+                                  if is_ibft_connection(con)]
         else:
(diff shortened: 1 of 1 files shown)


pyanaconda/modules/network/network.py
```diff
@@ -637,18 +637,26 @@ def get_dracut_arguments(self, iface, target_ip, hostname, ibft):
...
+            target_connections = [con for con in self.nm_client.get_connections()
+                                  if is_ibft_connection(con)]
         else:
-            for cfg in self._device_configurations.get_for_device(iface):
-                uuid = cfg.connection_uuid
-                if uuid:
-                    connection = self.nm_client.get_connection_by_uuid(uuid)
-                    if connection:
-                        target_connections.append(connection)
+            if self._device_configurations:
```
VladimirSlavik, line 645: Would this work as `elif` instead of nesting?

rvykydal: Hm, I wanted the ibft and non-ibft branch to be apparent so I'd keep two levels here. I mean we are not dealing with 3 cases here, but with 2 cases of which one is a bit complicated.

PR #2486, opened 2020-04-30
Add the DBus method GetModules to the Boss service

poncovka:
Call the DBus method `GetModules` to get service names of all running DBus modules
(including addons) that were discovered and started by the boss.


M4rtinK:
To add some back story - this is needed to check if a module providing some functionality is available before trying to use it's API. Technically, one could check the list of modules that were scheduled to start in the config object, but this new method actually checks the module is there up and running, which should be mode robust.

The main use case is for the subscription support, to check if the subscription module has been loaded and can be used or if the subscription related functionality (such as the Subscription spoke, subscription related installation tasks, the CDN installation source, etc.) should be turned off.

PR #2487, opened 2020-04-30
Destroy the disk selection dialog on escape

jstodola:
The "Configure mount point" dialog window for selecting disk devices for standard partitions doesn't close after pressing Esc (it's necessary to press Esc twice).


jkonecny12:
Jenkins, it's ok to test.

jkonecny12:
This is hell of the code, every dialog behaves differently. Some calling destroy inside of the dialog some don't and it's handled outside.

So or so, looks it is working correctly.

PR #2488, opened 2020-05-03
Do not offer disk selection when encrypting existing partition

jstodola:
Anaconda allowed to enter the "Configure mount point" window when
re-using an existing partition and making it encrypted.


jkonecny12:
Jenkins, it's ok to test.

jkonecny12:
jenkins, test this please

poncovka:
Tested.

poncovka:
@M4rtinK, I think we can merge this PR.

PR #2489, opened 2020-05-03
Improve behavior of the boot loader selection window

jstodola:
The UI in the "Select disks and boot loader" window  was not properly updated after removing the disk marked as a boot device:
* The first disk was correctly marked as a new boot device, but it was possible to set one more disk as a boot device.
* The label of the "set as boot" button was not updated properly.

Activate the set/remove buttons in the "Select disks and boot loader" window only when a disk is selected


jkonecny12:
Jenkins, it's ok to test.

rvykydal:
Works as expected to me with the patch.

PR #2490, opened 2020-05-04
subscription: Attach subscription task

M4rtinK:
The AttachSubscription task should be run after the system
has been successfully registered. It makes sure that appropriate
subscriptions are attached to the system based on system architecture,
account used for registration and system purpose data.

Note that this effectively subscribes the running installation
environment, so that it can communicate with the CDN and install
packages from it. Separate TransferSubscriptionTokens install time
task then makes sure the subscription tokens and configuration
files are transferred to the target system.

### DIFF
--- a/pyanaconda/modules/subscription/subscription_interface.py
+++ b/pyanaconda/modules/subscription/subscription_interface.py
@@ -180,3 +180,12 @@ class SubscriptionInterface(KickstartModuleInterface):
         return TaskContainer.to_object_path(
             self.implementation.unregister_with_task()
         )
+
+    def AttachSubscriptionWithTask(self) -> ObjPath:
+        """Attach subscription using a runtime DBus task.
+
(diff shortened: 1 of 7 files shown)


pyanaconda/modules/subscription/subscription_interface.py
```diff
@@ -171,3 +171,13 @@ def RegisterOrganizationKeyWithTask(self) -> ObjPath:
         return TaskContainer.to_object_path(
             self.implementation.register_organization_key_with_task()
         )
+
+    @emits_properties_changed
```
poncovka, line 175: I was wrong. The decorator is not necessary, because the properties won't be changed by the method, but by the task.

M4rtinK: No problem, dropping the decorator. :)

PR #2491, opened 2020-05-04
subscription: Run install time subscription tasks

M4rtinK:
The first commit introduces a helper function for easy checking if the Subscription module is running (due to this the PR is temporarily based on top of #2486). The second commit adds the Subscription module provided installation tasks to the installation task queue, provided the Subscription module is running.

~**NOTE:** When working on this I've realized there is no longer a reason to handle the System Purpose configuration task separately & there will be a followup PR that drops it's `with_task()`/`WithTask()` methods and rolls it to the list of regular installation tasks.~
See PR #2496.

### DIFF
--- a/pyanaconda/installation.py
+++ b/pyanaconda/installation.py
@@ -22,9 +22,10 @@ from pyanaconda.core.constants import PAYLOAD_LIVE_TYPES
 from pyanaconda.core.kernel import kernel_arguments
 from pyanaconda.modules.common.constants.objects import BOOTLOADER, SNAPSHOT, FIREWALL
 from pyanaconda.modules.common.constants.services import STORAGE, USERS, SERVICES, NETWORK, SECURITY, \
-    LOCALIZATION, TIMEZONE, BOSS
+    LOCALIZATION, TIMEZONE, BOSS, SUBSCRIPTION
 from pyanaconda.modules.common.structures.requirement import Requirement
 from pyanaconda.modules.common.task import sync_run_task
+from pyanaconda.modules.common.util import is_module_available
 from pyanaconda.progress import progress_message, progress_step, progress_complete, progress_init
 from pyanaconda import flags
 from pyanaconda.core import util
@@ -77,6 +78,16 @@ def _prepare_configuration(payload, ksdata):
     # connect progress reporting
     configuration_queue.queue_started.connect(lambda x: progress_message(x.status_message))
     configuration_queue.task_completed.connect(lambda x: progress_step(x.name))
+
+    # add installation tasks for the Subscription DBus module
+    if is_module_available(SUBSCRIPTION):
+        # we only run the tasks if the Subscription module is available
+        subscription_config = TaskQueue("Subscription configuration",
+                                        N_("Configuring Red Hat subscription"))
+        subscription_proxy = SUBSCRIPTION.get_proxy()
+        subscription_dbus_tasks = subscription_proxy.InstallWithTasks()
+        subscription_config.append_dbus_tasks(SUBSCRIPTION, subscription_dbus_tasks)
+        configuration_queue.append(subscription_config)
 
     # schedule the execute methods of ksdata that require an installed system to be present
     os_config = TaskQueue("Installed system configuration", N_("Configuring installed system"))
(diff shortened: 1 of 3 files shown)


pyanaconda/subscription.py
```diff
@@ -0,0 +1,32 @@
...
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+
+from pyanaconda.modules.common.constants.namespaces import SUBSCRIPTION_NAMESPACE
+from pyanaconda.modules.common.constants.services import BOSS
+
+
+def check_subscription_module_available():
```
poncovka, line 24: Maybe I would create a more general function `is_module_available` and put it to `pyanaconda.modules.common.utils`. The function could have one argument of the type `DBusServiceIdentifier`. Then we could use it like this:

```
is_module_available(SUBSCRIPTION)
```

pyanaconda/installation.py
```diff
@@ -319,6 +320,16 @@ def run_pre_install():
...
         installation_queue.append(late_storage)
 
+    # Add installation tasks for the Subscription DBus module
+    # - to see if this is appropriate for the current environment we check if
+    #   the Subscription module is running via a helper function
+    if check_subscription_module_available():
+        subscription_config = TaskQueue("Subscription configuration",
+                                        N_("Configuring Red Hat subscription"))
+        subscription_proxy = SUBSCRIPTION.get_proxy()
+        subscription_dbus_tasks = subscription_proxy.InstallWithTasks()
```
poncovka, line 330: It there a reason why this task should be run here? Couldn't we move it at the beginning of `_prepare_configuration`?

pyanaconda/installation.py
```diff
@@ -319,6 +320,16 @@ def run_pre_install():
         late_storage.append_dbus_tasks(STORAGE, [conf_task])
         installation_queue.append(late_storage)
 
+    # Add installation tasks for the Subscription DBus module
+    # - to see if this is appropriate for the current environment we check if
+    #   the Subscription module is running via a helper function
+    if check_subscription_module_available():
+        subscription_config = TaskQueue("Subscription configuration",
```
poncovka, line 327: I don't think this queue is connected to the installation queue.

pyanaconda/subscription.py
```diff
@@ -0,0 +1,32 @@
...
+    """Check if the Subscription module appears to be running.
+
+    :return: True if Subscription module is running, False otherwise
+    :rtype: bool
+    """
+    boss_proxy = BOSS.get_proxy()
+    subscription_module_name = ".".join(SUBSCRIPTION_NAMESPACE)
```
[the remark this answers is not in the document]
M4rtinK: Cool, that looks better. Thanks! :)

pyanaconda/subscription.py
M4rtinK: Sounds good, I'll do that.

pyanaconda/installation.py
M4rtinK: I simply put it to the same location where it was in RHEL 8.2. It needs storage to be configured and mounted and packages to be installed (in case the Insights task runs). Given that tasks added in `_prepare_configuration` run even later than where it is now it should be fine to move it there & I'll do that. :)

pyanaconda/installation.py
M4rtinK: Fixed during the move.

pyanaconda/subscription.py
M4rtinK: Using `util` for consistency in the end as we already have `core/util.py`.

pyanaconda/modules/common/util.py
```diff
@@ -0,0 +1,32 @@
...
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+
+from pyanaconda.modules.common.constants.services import BOSS
+
+
+def is_module_available(module_service_identifier):
+    """Check if the module appears to be running be running.
```
poncovka, line 24: Please, fix the docstring.

M4rtinK: Fixed. :)

PR #2474, opened 2020-04-27
2/2 DNF kickstart support

jkonecny12:
Add kickstart support for DNF module.

In this PR I'm disabling `%packages` which will be solved by the main process for the time being. Right now, we have to add support for the base repositories to support subscription spoke.

Will be implemented:
- [x] KS support for HDD
- [x] KS support for NFS
- [x] KS support for URL

Part of this PR was split to https://github.com/rhinstaller/anaconda/pull/2484 to make this smaller.

PR #2493, opened 2020-05-05
Add get_repo_configuration() for DNF payload

jkonecny12:
This will provide temporary API which we can use to get sources working in the main process payload. It will provide the necessary data to be able to work with these in the dnf library.

We are re-using the existing `RepoConfigurationData` structure to handle all the sources data.

~This is not finished. I'm waiting for https://github.com/rhinstaller/anaconda/pull/2474 code to finish URL source support.~

### DIFF
--- a/pyanaconda/modules/payloads/source/source_base.py
+++ b/pyanaconda/modules/payloads/source/source_base.py
@@ -144,3 +144,17 @@ class MountingSourceBase(PayloadSourceBase, ABC):
         """
         task = TearDownMountTask(self._mount_point)
         return [task]
+
+
+class RPMSourceMixin(ABC):
+    """Interface class which has to be implemented by all sources used by DNF payload."""
+
+    @abstractmethod
+    def generate_repo_configuration(self):
+        """Generate RepoConfigurationData structure.
+
+        This structure will be used by DNF payload in the main process.
+
+        FIXME: This is a temporary solution. Will be removed after DNF payload logic is moved.
+        """
+        pass
--- a/pyanaconda/modules/payloads/source/url/url.py
+++ b/pyanaconda/modules/payloads/source/url/url.py
@@ -24,14 +24,14 @@ from pyanaconda.core.payload import ProxyString, ProxyStringError
 from pyanaconda.modules.common.errors.general import InvalidValueError
 from pyanaconda.modules.common.structures.payload import RepoConfigurationData
 from pyanaconda.modules.payloads.constants import SourceType, URLType, SourceState
-from pyanaconda.modules.payloads.source.source_base import PayloadSourceBase
+from pyanaconda.modules.payloads.source.source_base import PayloadSourceBase, RPMSourceMixin
 from pyanaconda.modules.payloads.source.url.url_interface import URLSourceInterface
 
 from pyanaconda.anaconda_loggers import get_module_logger
 log = get_module_logger(__name__)
 
 
-class URLSourceModule(PayloadSourceBase):
+class URLSourceModule(PayloadSourceBase, RPMSourceMixin):
     """The URL source payload module."""
 
     REPO_NAME_ID = 0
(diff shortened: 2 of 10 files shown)


pyanaconda/modules/payloads/source/source_base.py
```diff
@@ -134,3 +136,31 @@ def tear_down_with_tasks(self):
         """
         task = TearDownMountTask(self._mount_point)
         return [task]
+
+
+class RPMSourceInterface(ABC):
```
VladimirSlavik, line 141: I believe interfaces go to the file `source_base_interface.py`...

pyanaconda/modules/payloads/source/source_base.py
```diff
@@ -134,3 +136,31 @@ def tear_down_with_tasks(self):
...
+        """Generate RepoConfigurationData structure.
+
+        This structure will be used by DNF payload in the main process.
+
+        FIXME: This is a temporary solution. Will be removed after DNF payload logic is moved.
+        """
+        pass
+
+
+class RPMMountingSourceBase(MountingSourceBase, ABC):
```
poncovka, line 155: I would go the other way, rename `MountingSourceBase` to `Mountable` and remove its parents, so it could be used as a mixin. Then we can have a proper subclass of `PayloadSourceBase` called `RPMSource`. I think it makes more sense. The current solution is little confusing.

pyanaconda/modules/payloads/source/source_base.py
```diff
@@ -134,3 +136,31 @@ def tear_down_with_tasks(self):
...
+    """Enhanced MountingSourceBase class to handle code specific for RPM sources."""
+    # pylint: disable=abstract-method
+
+    def generate_repo_configuration(self):
+        """Generate RepoConfigurationData structure."""
+        data = RepoConfigurationData()
+
+        data.name = self.type.value
+        data.url = join_paths("file:///", self.mount_point)
```
poncovka, line 164: I would prefer to see the implementation in every RPM source. Maybe, the class `RepoConfigurationData` could have a class method `for_file` that would accept a name and a path and a return a new instance of the data object.

pyanaconda/modules/payloads/source/source_base.py
```diff
@@ -134,3 +136,31 @@ def tear_down_with_tasks(self):
...
+
+class RPMMountingSourceBase(MountingSourceBase, ABC):
+    """Enhanced MountingSourceBase class to handle code specific for RPM sources."""
+    # pylint: disable=abstract-method
+
+    def generate_repo_configuration(self):
+        """Generate RepoConfigurationData structure."""
+        data = RepoConfigurationData()
+
+        data.name = self.type.value
```
poncovka, line 163: About the repo name, shouldn't we investigate it a little further? Maybe DNF will generate its own name if none is set.

pyanaconda/modules/payloads/source/url/url.py
```diff
@@ -67,6 +67,15 @@ def for_publication(self):
         """Get the interface used to publish this source."""
         return URLSourceInterface(self)
 
+    def generate_repo_configuration(self):
+        """Generate RepoConfigurationData structure."""
+        data = RepoConfigurationData()
+
+        data.name = self.type.value
+        # TODO: Use solution https://github.com/rhinstaller/anaconda/pull/2474
```
poncovka, line 75: I think that we should just return `self.repo_configuration` in this case.

jkonecny12: Ahh, good point. Truth is that I don't need to fill it up when it's already filed from KS data.

pyanaconda/modules/payloads/source/source_base.py
jkonecny12: AFAIK (and I'll check that to be sure) we are always setting name for DNF Repo. For the base repo it is `anaconda`. I've created this as a type identifier which could be handy but maybe we can remove this completely and just not set it.

jkonecny12:
**UPDATED:**
* Changed implementation based on discussion with @poncovka and @VladimirSlavik.

I will change `MountpointSourceBase` to a mixin in a separate PR.

jkonecny12:
**UPDATED:**
* Rebased

PR #2494, opened 2020-05-05
Remove the HasSource D-Bus function

VladimirSlavik:
It was redundant, Sources can be used instead.

### DIFF
--- a/tests/nosetests/pyanaconda_tests/module_payload_shared.py
+++ b/tests/nosetests/pyanaconda_tests/module_payload_shared.py
@@ -103,7 +103,6 @@ class PayloadSharedTest(object):
     def check_empty_sources(self):
         """Default check for payload with no sources set."""
         self._test.assertEqual([], self.payload_interface.Sources)
-        self._test.assertFalse(self.payload_interface.HasSource())
 
     def set_and_check_sources(self, test_sources, exception=None):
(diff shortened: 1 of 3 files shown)


tests/nosetests/pyanaconda_tests/module_payload_shared.py
```diff
@@ -152,6 +151,6 @@ def check_sources(self, expected_sources=None):
         self._test.assertEqual(self.payload_interface.Sources, expected_paths)
 
         if expected_sources:
-            self._test.assertTrue(self.payload_interface.HasSource())
+            self._test.assertGreater(len(self.payload_interface.Sources), 0)
```
poncovka, line 154: I think this is the same as `self._test.assertTrue(self.payload_interface.Sources)`.

VladimirSlavik: I had something similar there, but didn't like the implicit (unstated) conversion to `bool`. Actually, how about this:
```
self._test.assertEqual(
  bool(expected_sources),
  self.payload_interface.Sources == []
)
```

poncovka: Actually, you can remove this block, because the property Sources is tested above.

PR #2495, opened 2020-05-05
Ignore the required package dmraid in RHEL

jstodola:
The package dmraid is not available in RHEL 8, so ignore this requirement.

Resolves: rhbz#1831790


poncovka:
Jenkins, test this please.

poncovka:
I'm not sure how to test it, but it is a very small change that should be safe to merge.

PR #2496, opened 2020-05-06
subscription: Simplified system purpose configuration

M4rtinK:
This PR substantially simplifies system purpose configuration in comparison to what was done back in RHEL 8.0 and 8.2 due to the specific circumstances back then. 

In short, we now simply set system purpose data to the installation environment every time a new system purpose data DBus structure is supplied & a separate installation task will take care of transferring the system purpose configuration file to the target system. Due to this we need just a simple runtime DBus task that des not need to be directly invoked via DBus API. Also we no longer need the `syspurpose` package to be installed to the target system as `syspurpose` is no longer called on the target system anymore.

For more details you can see the commit message of the third commit.

The PR is structured as follows:
- commit 1 adds a simple helper function for checking if system purpose has been set
- commit 2 adds equality checking for SystemPurpose data, that wee need to decide if system purpose data changed and should be re-applied
- commit 3 does the actual simplification of system purpose configuration

### DIFF
--- a/pyanaconda/modules/subscription/installation.py
+++ b/pyanaconda/modules/subscription/installation.py
@@ -22,12 +22,12 @@ import shutil
 from dasbus.typing import get_variant, Str
 
 from pyanaconda.core import util
+from pyanaconda.core.constants import RHSM_SYSPURPOSE_FILE_PATH
+from pyanaconda.core.subscription import check_system_purpose_set
 
 from pyanaconda.modules.common.task import Task
 from pyanaconda.modules.common.errors.installation import InsightsConnectError, \
     InsightsClientMissingError, SubscriptionTokenTransferError
-
-from pyanaconda.modules.subscription import system_purpose
 
 from pyanaconda.anaconda_loggers import get_module_logger
 log = get_module_logger(__name__)
@@ -81,35 +81,6 @@ class ConnectToInsightsTask(Task):
             raise InsightsConnectError("Connecting to Red Hat Insights failed.")
 
 
-class SystemPurposeConfigurationTask(Task):
-    """Installation task for setting system purpose."""
-
-    def __init__(self, sysroot, system_purpose_data):
-        """Create a new system purpose configuration task.
-
-        :param str sysroot: a path to the root of the installed system
-        :param system_purpose_data: system purpose data DBus structure
-        :type system_purpose_data: DBusData instance
-        """
-        super().__init__()
-        self._sysroot = sysroot
-        self._system_purpose_data = system_purpose_data
-
-    @property
-    def name(self):
-        return "Set system purpose"
-
-    def run(self):
-        # apply System Purpose data
-        return system_purpose.give_the_system_purpose(
-            sysroot=self._sysroot,
-            role=self._system_purpose_data.role,
-            sla=self._system_purpose_data.sla,
-            usage=self._system_purpose_data.usage,
-            addons=self._system_purpose_data.addons
-        )
-
-
 class RestoreRHSMLogLevelTask(Task):
     """Restore RHSM log level back to INFO."""
 
@@ -144,7 +115,6 @@ class TransferSubscriptionTokensTask(Task):
 
     RHSM_REPO_FILE_PATH = "/etc/yum.repos.d/redhat.repo"
     RHSM_CONFIG_FILE_PATH = "/etc/rhsm/rhsm.conf"
-    RHSM_SYSPURPOSE_FILE_PATH = "/etc/rhsm/syspurpose/syspurpose.json"
     RHSM_ENTITLEMENT_KEYS_PATH = "/etc/pki/entitlement"
     RHSM_CONSUMER_KEY_PATH = "/etc/pki/consumer/key.pem"
     RHSM_CONSUMER_CERT_PATH = "/etc/pki/consumer/cert.pem"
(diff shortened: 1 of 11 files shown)


pyanaconda/modules/subscription/subscription.py
```diff
@@ -302,68 +296,42 @@ def set_system_purpose_data(self, system_purpose_data):
...
-        log.debug("System purpose is applied set to: %s", system_purpose_applied)
-
-    def _apply_syspurpose(self):
+        # If new structure contains some data and the data is different than
+        # what was in the old structure, set system purpose data to the system.
+        #
+        # NOTE: We do this only after setting the value and logging the "set"
+        #       as having a set -> apply sequence in the logs makes more sense
+        #       than the other way around.
+        data_available = system_purpose_data.check_data_available()
```
poncovka, line 312: I don't get it. If the client sets the system purpose data with no data available, we will set the property, but not the system. Therefore, the system will use the previous configuration, that is now different from our property.

pyanaconda/modules/subscription/subscription_interface.py
```diff
@@ -91,25 +89,6 @@ def SetSystemPurposeData(self, system_purpose_data: Structure):
...
-    def IsSystemPurposeApplied(self) -> Bool:
-        """Report if system purpose data has been applied.
-
-        We don't differentiate between that installation environment and the target system in this
-        case as we will make sure the system purpose data will always end up on the target system
-        if requested by the user, regardless of where it is initially set.
-        """
-        return self.implementation.is_system_purpose_applied
-
-    def SetSystemPurposeWithTask(self) -> ObjPath:
```
poncovka, line 104: I don't think it is a good idea to have a setter with side effects. If you want to go this way, then remove the setter from the DBus API and extend the method `SetSystemPurposeWithTask` with an argument for the system purpose data. If the task succeeds, set the property for the system purpose data. However, this solution might have some new drawbacks.

In general, I am not sure if it is worth it to modify the current solution. You can achieve the same effect in the client by calling the method `SetSystemPurposeWithTask` directly after setting the property `SystemPurposeData`. The DBus API should be straightforward.

pyanaconda/modules/subscription/subscription.py
M4rtinK: This is due to the fact that you need at least one of the values to call the `syspurpose` utility - with no data, you can't really call it. But indeed, like this it is not possibly to clear previously set data, which I think should be possible - eq. the user changes their mind, you want to clear something set by kickstart.

So I'll remove the data available check here & in the task & have the task remove existing syspurpose.json if it gets empty `SystemPurposeData`. Does that sound right ? :)

pyanaconda/modules/subscription/subscription.py
```diff
@@ -302,68 +296,42 @@ def set_system_purpose_data(self, system_purpose_data):
         :param system_purpose_data: system purpose data structure to be set
         :type system_purpose_data: DBus structure
         """
+        # record if the new data is different from current data
+        data_different = self._system_purpose_data != system_purpose_data
```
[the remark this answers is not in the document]
M4rtinK: My idea was that the GUI will simply set new SystemPurposeData based on spoke content every time the user exits the spoke - with this check in place, it can do that an avoid triggering call to the `syspurpose` utility and associated log messages polluting the logs.

But thinking about it, it would still have a lot of "System purpose data set" even if nothing changes in the spoke in regards to system purpose, which is not that much better. So yeah, let's drop this check and do it in the GUI, the added code complexity should be not that bad there.

pyanaconda/modules/subscription/subscription.py
poncovka: I guess we should always remove the previous configuration. Based on the code, if I set `role` and nothing else, then the system will use the previous configuration for `sla`, `usage` and `addons`.

pyanaconda/modules/subscription/subscription_interface.py
M4rtinK: I don't think having the task set the system purpose property is a good idea - that seems like something that could cause race conditions & complicate setting system purpose data overall.

Yeah, let's revert to the current solution with the expectation of the client having to call `SetSystemPurposeWithTask()` to actually apply the system purpose data that has been set (hopefully this does not cause race condition by itself, we can change it later if it does). 

Looking into the attached subscription parsing, this might actually be needed - activation keys might have system purpose data attached, that override what was set on the system before registration, so we might need to set system purpose data just to the Subscription module property, so that it can be shown in the GUI easily. 

In this case the client (GUI in this case) would check if system is subscribed and would not run the task (changing system purpose of subscribed system does is not a good idea). This subscription stuff just needs to be complicated. :P

pyanaconda/modules/subscription/subscription.py
M4rtinK: That's a very good point - seems to illustrate a bit that the syspurpose tool was made for simpler usecases that what we have in Anaconda. :)

pyanaconda/subscription.py
```diff
@@ -0,0 +1,40 @@
+#
+# Subscription related helper functions.
```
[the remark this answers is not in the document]
M4rtinK: Let's move it to `pyanaconda.core` as it's not used just by UI bits.

pyanaconda/modules/subscription/installation.py
```diff
@@ -219,7 +189,7 @@ def _transfer_system_purpose(self):
          - this means the syspurpose tool has been called in the installation
            environment & we need to transfer the results to the target system
         """
-        if os.path.exists(self.RHSM_SYSPURPOSE_FILE_PATH):
+        if check_system_purpose_set("/"):
```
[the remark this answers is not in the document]
M4rtinK: Good point, added. :)

pyanaconda/modules/subscription/subscription.py
M4rtinK: Changed `give_system_purpose()` function to remove existing `syspurpose.json` before running the `syspurpose` tool.

pyanaconda/subscription.py
```diff
@@ -0,0 +1,40 @@
...
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+
+import os
+
+from pyanaconda.core import util
+from pyanaconda.core.constants import RHSM_SYSPURPOSE_FILE_PATH
+
+
+def check_system_purpose_set(sysroot):
```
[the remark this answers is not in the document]
M4rtinK: Let's do it! :)

pyanaconda/modules/subscription/installation.py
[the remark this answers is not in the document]
M4rtinK: Added.

M4rtinK:
Thanks for the reviews & feedback! :)

PR #2497, opened 2020-05-06
subscription: Add AttachedSubscription DBus structure

M4rtinK:
The AttachedSubscription DBus structure describes a single subscription
that has been attached to the target system.

A list of AttachedSubscription instances together with the InsightsEnabled property
and SystemPurposeData property describe the system state post registration and subscription.

Note that all the fields are effectively string "blobs" we are getting
from a JSON file returned by RHSM DBus API.

### DIFF
--- a/pyanaconda/modules/common/structures/subscription.py
+++ b/pyanaconda/modules/common/structures/subscription.py
@@ -362,3 +362,147 @@ class SubscriptionRequest(DBusData):
     @server_proxy_password.setter
     def server_proxy_password(self, password: SecretData):
         self._server_proxy_password = password
+
+
+class AttachedSubscription(DBusData):
+    """Data for a single attached subscription."""
+
+    def __init__(self):
+        self._name = ""
+        self._service_level = ""
+        self._sku = ""
+        self._contract = ""
+        self._start_date = ""
+        self._end_date = ""
+        # we can expect at least one entitlement
+        # to be consumed per attached subscription
+        self._consumed_entitlement_count = 1
+
+    @property
+    def name(self) -> Str:
+        """Name of the attached subscription.
+
+        Example: "Red Hat Beta Access"
+
+        :return: subscription name
+        :rtype: str
+        """
+        return self._name
+
+    @name.setter
+    def name(self, name: Str):
+        self._name = name
+
+    @property
+    def service_level(self) -> Str:
+        """Service level of the attached subscription.
+
+        Example: "Premium"
+
+        :return: service level
+        :rtype: str
+        """
+        return self._service_level
+
+    @service_level.setter
+    def service_level(self, service_level: Str):
+        self._service_level = service_level
+
+    @property
+    def sku(self) -> Str:
+        """SKU id of the attached subscription.
+
+        Example: "MBT8547"
+
+        :return: SKU id
+        :rtype: str
+        """
+        return self._sku
+
+    @sku.setter
+    def sku(self, sku: Str):
+        self._sku = sku
+
+    @property
+    def contract(self) -> Str:
+        """Contract identifier.
+
+        Example: "32754658"
+
+        :return: contract identifier
+        :rtype: str
+        """
+        return self._contract
+
+    @contract.setter
+    def contract(self, contract: Str):
+        self._contract = contract
+
+    @property
+    def start_date(self) -> Str:
+        """Subscription start date.
+
+        We do not guarantee fixed date format,
+        but we aim for the date to look good
+        when displayed in a GUI and be human
+        readable.
+
+        For context see the following bug, that
+        illustrates the issues we are having with
+        the source date for this property, that
+        prevent us from providing a consistent
+        date format:
+        https://bugzilla.redhat.com/show_bug.cgi?id=1793501
+
+        Example: "Nov 04, 2019"
+
+        :return: start date of the subscription
+        :rtype: str
+        """
+        return self._start_date
+
+    @start_date.setter
+    def start_date(self, start_date: Str):
+        self._start_date = start_date
+
+    @property
+    def end_date(self) -> Str:
+        """Subscription end date.
+
+        We do not guarantee fixed date format,
+        but we aim for the date to look good
+        when displayed in a GUI and be human
+        readable.
+
+        For context see the following bug, that
+        illustrates the issues we are having with
+        the source date for this property, that
+        prevent us from providing a consistent
+        date format:
+        https://bugzilla.redhat.com/show_bug.cgi?id=1793501
+
+        Example: "Nov 04, 2020"
+
+        :return: end date of the subscription
+        :rtype: str
+        """
+        return self._end_date
+
+    @end_date.setter
+    def end_date(self, end_date: Str):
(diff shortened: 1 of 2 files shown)


pyanaconda/modules/common/structures/subscription.py
```diff
@@ -362,3 +362,107 @@ def server_proxy_password(self) -> SecretData:
...
+        """
+        return self._name
+
+    @name.setter
+    def name(self, name: Str):
+        self._name = name
+
+    @property
+    def service_level(self) -> Str:
+        """Service level of the attached subscription.
```
poncovka, line 394: Could you provide an example in the docstring?

pyanaconda/modules/common/structures/subscription.py
```diff
@@ -362,3 +362,107 @@ def server_proxy_password(self) -> SecretData:
...
+        """
+        return self._service_level
+
+    @service_level.setter
+    def service_level(self, service_level: Str):
+        self._service_level = service_level
+
+    @property
+    def sku(self) -> Str:
+        """SKU id of the attached subscription.
```
poncovka, line 407: Same here.

pyanaconda/modules/common/structures/subscription.py
```diff
@@ -362,3 +362,107 @@ def server_proxy_password(self) -> SecretData:
...
+        """
+        return self._sku
+
+    @sku.setter
+    def sku(self, sku: Str):
+        self._sku = sku
+
+    @property
+    def contract(self) -> Str:
+        """Contract identifier.
```
poncovka, line 420: Same here?

pyanaconda/modules/common/structures/subscription.py
```diff
@@ -362,3 +362,107 @@ def server_proxy_password(self) -> SecretData:
...
+        """
+        return self._contract
+
+    @contract.setter
+    def contract(self, contract: Str):
+        self._contract = contract
+
+    @property
+    def start_date(self) -> Str:
+        """Subscription start date.
```
poncovka, line 433: Are we able to specify the format? I know we had some issues with that.

tests/nosetests/pyanaconda_tests/module_subscription_tests.py
```diff
@@ -798,6 +798,57 @@ def set_subscription_request_sensitive_data_keep_test(self):
...
+            "service-level": "really good",
+            "sku": "ABC123456",
+            "contract": "987654321",
+            "start-date": "1 January 1970",
+            "end-date": "19 January 2038",
+            "consumed-entitlement-count": "9001"
+        }
+        # compare the full structure with expected values
+        self.assertEqual(
+            get_native(AttachedSubscription.to_structure(full_request)),
```
poncovka, line 848: It would be good to have one test with variants so we can verify the types.

pyanaconda/modules/common/structures/subscription.py
```diff
@@ -362,3 +362,107 @@ def server_proxy_password(self) -> SecretData:
...
+        :rtype: str
+        """
+        return self._end_date
+
+    @end_date.setter
+    def end_date(self, end_date: Str):
+        self._end_date = end_date
+
+    @property
+    def consumed_entitlement_count(self) -> Str:
```
poncovka, line 458: Shouldn't this be an integer?

M4rtinK: The idea is that we are not really sure - the JSON returned by the RHSM DBus API from which this data is returned has what seems to be a string containing an integer. But given that the format is not well documented and I fear we might one day get "3 entitlements" or something like that.

So it seems prudent to me to treat this data as an opaque string that we do not really interpret to avoid surprises later on.

pyanaconda/modules/common/structures/subscription.py
M4rtinK: The problem is in the JSON from which this is parsed providing dates in an inconsistent manner - for more info see [bug 1793501](https://bugzilla.redhat.com/show_bug.cgi?id=1793501).

The code that does the parsing (coming soon in followup PR) handles these differences and provides a consistent human readable data that is set to the DBus structure and then displayed.

Another issues is that we would like to harmonize the date display between all the tools handling this data (Anaconda, Cockpit RHSM plugin, RHSM CLI, Satellite, etc.) but this is still an ongoing effort and no shared date format has yet been agreed on.

To summarize, the only date format we can specify at the moment for these fields in the DBus structure is "nice human readable date with consistent formatting" - which I will add to the docstring. :)

Later on (possibly before declaring our DBus API stable) we might able able to improve this.

pyanaconda/modules/common/structures/subscription.py
M4rtinK: This is one of the bugs for the JSON output not being documented: [bug 1790987](https://bugzilla.redhat.com/show_bug.cgi?id=1790987)

poncovka: In that case, could you document the explanation in the docstring of the property?

pyanaconda/modules/common/structures/subscription.py
poncovka: Ok, I would document this in the docstring.

pyanaconda/modules/common/structures/subscription.py
M4rtinK: Added an example.

pyanaconda/modules/common/structures/subscription.py
M4rtinK: Added an example.

pyanaconda/modules/common/structures/subscription.py
M4rtinK: Added.

pyanaconda/modules/common/structures/subscription.py
M4rtinK: Added a docstring explaining what can be expected in the data property & linked to the bug.

pyanaconda/modules/common/structures/subscription.py
M4rtinK: So I looked deeper into this and in the end it looks rather unlikely we will get something else than an integer in the key we are parsing in the JSON file. Also the "consumed" suffix is added by us at the moment. So I've switched the property to be an `Int`.

tests/nosetests/pyanaconda_tests/module_subscription_tests.py
M4rtinK: Good point, switched the tests to variants. :)

pyanaconda/modules/common/structures/subscription.py
M4rtinK: Added.

pyanaconda/modules/common/structures/subscription.py
M4rtinK: Added.

M4rtinK:
PR updated
- examples added
- entitlement count switched to Int
- improved tests

PR #2498, opened 2020-05-07
subscription: Initial subscription GUI bits

M4rtinK:
The first commit adds the full Glade file "XML blob" providing all the spoke widget layout similar to what is currently in RHEL 8.2.

The second commit adds a couple helper functions in a separate file for easy unit testing. It is expected more will be added there later.

The third commit adds the main Subscription spoke structure. This is mostly boilerplate for interacting with the many controls on the Subscription spoke, with many of the logic methods intentionally left out for now. It is expected followup PRs of much more manageable size will add these parts. See the commit message for list of what has been left out in detail.

### DIFF
--- a/pyanaconda/ui/gui/spokes/lib/subscription.py
+++ b/pyanaconda/ui/gui/spokes/lib/subscription.py
@@ -1,0 +1,102 @@
+# Helper methods for the Subscription spoke.
+#
+# Copyright (C) 2020  Red Hat, Inc.
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# the GNU General Public License v.2, or (at your option) any later version.
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY expressed or implied, including the implied warranties of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General
+# Public License for more details.  You should have received a copy of the
+# GNU General Public License along with this program; if not, write to the
+# Free Software Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
+# 02110-1301, USA.  Any Red Hat trademarks that are incorporated in the
+# source code or documentation are not subject to the GNU General Public
+# License and may only be used or replicated with the express permission of
+# Red Hat, Inc.
+#
+
+from collections import namedtuple
+
+from pyanaconda.core.i18n import _
+
+TextComboBoxItem = namedtuple("TextComboBoxItem", ["value", "display_string", "is_preselected"])
+
+
+def handle_user_provided_value(user_provided_value, valid_values):
+    """Handle user provided value (if any) based on list of valid values.
+
+    There are three possible outcomes:
+    - the value matches one of the valid values, so we preselect the valid value
+    - the value does not match a valid value, so we append a custom value
+      to the list and preselect it
+    - the user provided value is not available (empty string), no matching will be done
+      and no value will be preselected
+
+    :param str user_provided_value: a value provided by user
+    :param list valid_values: a list of valid values
+    :returns: list of values with one value preselected
+    :rtype: list of TextComboBoxItem tuples
+    """
+    preselected_value_list = []
+    value_matched = False
+    for valid_value in valid_values:
+        preselect = False
+        if user_provided_value and not value_matched:
+            if user_provided_value == valid_value:
+                preselect = True
+                value_matched = True
+        item = TextComboBoxItem(value=valid_value,
+                                display_string=valid_value,
+                                is_preselected=preselect)
+        preselected_value_list.append(item)
+    # check if the user provided value matched a valid value
+    if user_provided_value and not value_matched:
+        # user provided value did not match any valid value,
+        # add it as a custom value to the list and preselect it
+        other_value_string = _("Other ({})").format(user_provided_value)
+        item = TextComboBoxItem(value=user_provided_value,
+                                display_string=other_value_string,
+                                is_preselected=True)
+        preselected_value_list.append(item)
+    return preselected_value_list
+
+
+def fill_combobox(combobox, user_provided_value, valid_values):
+    """Fill the given ComboBoxText instance with data based on current value & valid values.
+
+    Please note that it is possible that the list box will be empty if no
+    list of valid values are available and the user has not supplied any value
+    via kickstart or the DBUS API.
+
+    NOTE: Removes any existing values from the GTK ComboBoxText instance before
+          filling it.
+
+    :param combobox: the combobox to fill
+    :param user_provided_value: the value provided by the user (if any)
+    :type user_provided_value: str or None
+    :param list valid_values: list of known valid values
+    """
+    preselected_value_list = handle_user_provided_value(user_provided_value,
+                                                        valid_values)
+    # make sure the combo box is empty
+    combobox.remove_all()
+
+    # add the "Not Specified" option as the first item
+    # - otherwise the user would not be able to unselect option clicked previously
+    #   or selected via kickstart
+    # - set the active id to this value by default
+
+    active_id = ""
+    combobox.append("", _("Not Specified"))
+
+    if preselected_value_list:
+        for value, display_string, preselected in preselected_value_list:
+            combobox.append(value, display_string)
+            # the value has been preselected, set the active id accordingly
+            if preselected:
+                active_id = value
+
+    # set the active id (what item should be selected in the combobox)
+    combobox.set_active_id(active_id)
--- a/pyanaconda/ui/gui/spokes/subscription.py
+++ b/pyanaconda/ui/gui/spokes/subscription.py
@@ -1,0 +1,830 @@
+# Subscription spoke class
+#
+# Copyright (C) 2020 Red Hat, Inc.
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# the GNU General Public License v.2, or (at your option) any later version.
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY expressed or implied, including the implied warranties of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General
+# Public License for more details.  You should have received a copy of the
+# GNU General Public License along with this program; if not, write to the
+# Free Software Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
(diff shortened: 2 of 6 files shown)


M4rtinK:
For some context, this is how the spoke currently looks like on RHEL 8.2:

rhel82_rhsm_hub
spoke status on the Hub

rhel82_rhsm_not_registered
not yet registered with some fields filled in

rhel82_rhsm_registered
registered & subscription attached (there could be multiple subscriptions attached, so the subscription delegates are inside a scrollable list view)

M4rtinK:
jenkins, test this please

pyanaconda/ui/gui/spokes/subscription.py
```diff
@@ -0,0 +1,750 @@
...
+    def http_proxy_location(self):
+        return self._http_proxy_location_entry.get_text()
+
+    @http_proxy_location.setter
+    def http_proxy_location(self, location):
+        self._http_proxy_location_entry.set_text(location)
+
+    @property
+    def http_proxy_username(self):
+        return self._http_proxy_username_entry.get_text()
```
poncovka, line 173: I am not a big fan of this approach. Instead of keeping the data in UI elements, I would prefer to keep the data in the subscription request. These properties could be replaced with signal handlers that would update the request with a new value of the changed UI element. Then you can always use the request to check the current values.

This works pretty well in the custom partitioning spoke (see https://github.com/rhinstaller/anaconda/blob/master/pyanaconda/ui/gui/spokes/custom_storage.py#L1563). I use the request to log the current state or to check if something has changed.

However, it might be a too big change that could complicate next backports, so it is up to you.

pyanaconda/ui/gui/spokes/subscription.py
```diff
@@ -0,0 +1,750 @@
...
+        log.debug("Subscription GUI: apply() running")
+        # Set System Purpose data to the Subscription DBus module,
+        # it makes sense to send the rest only right before a
+        # registration attempt.
+        self._set_system_purpose_data_to_subscription_module()
+
+    def refresh(self):
+        log.debug("Subscription GUI: refresh() running")
+        # update subscription request from the Subscription module
+        self._subscription_request = self._get_subscription_request()
```
poncovka, line 132: This method should set up the UI elements from the current request.

pyanaconda/ui/gui/spokes/subscription.py
```diff
@@ -0,0 +1,750 @@
...
+        self._attached_subscriptions_label = self.builder.get_object(
+            "attached_subscriptions_label"
+        )
+        self._subscriptions_listbox = self.builder.get_object("subscriptions_listbox")
+
+        # unregister button
+        self._unregister_revealer = self.builder.get_object("unregister_revealer")
+        self._unregister_button = self.builder.get_object("unregister_button")
+
+        # setup spoke state based on Subscription DBus module state
```
poncovka, line 590: This part should be in the refresh method.

pyanaconda/ui/gui/spokes/subscription.py
M4rtinK: I think it is a good idea - this always seemed cumbersome, but with the DBus module infrastructure on RHEL 8.2 & the available timeframe I was not able to come up with something better.

I like the idea & if its known to work elsewhere I think I can make it work here as well.

pyanaconda/ui/gui/spokes/subscription.py
```diff
@@ -0,0 +1,750 @@
...
+    REGISTRATION_PAGE = 0
+    SUBSCRIPTION_STATUS_PAGE = 1
+
+    @classmethod
+    def should_run(cls, environment, data):
+        """The Subscription spoke should run only if the Subscription module is available."""
+        return is_module_available(SUBSCRIPTION)
+
+    def __init__(self, *args):
+        NormalSpoke.__init__(self, *args)
```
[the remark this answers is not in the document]
M4rtinK: Fixed.

pyanaconda/ui/gui/spokes/subscription.py
M4rtinK: Split into a content refresh method.

pyanaconda/ui/gui/spokes/subscription.py
M4rtinK: Added a content refresh method called from here.

M4rtinK:
**PR updated:**
- content of the spoke is now always set based on DBus module state on spoke entry
- subscription request and system purpose data DBus structure mirrors are now updated by signal handlers when content of a field changes due to user interaction
- added a couple changed signals to the glade file, so it is no longer 100% identical to the RHEL 8.2
- DBus structure mirrors are set to the module on `apply()` or (in the future) before a subscription attempt
- support for password placeholders added (would have to be set by another client via DBus as kickstart only supports org & key authentication

**Known issues:**
- the visibility checkboxes actually have no effect on the values being set or not to the DBus module, a reasonable implementation if this needs to be added that keeps expected user behavior from 8.2 (fields are not cleared when you just toggle the checkbox) and does not complicate the code too much
- handling of the Insights default (off by default for kickstart installs, on by default for interactive GUI installs) is ugly & cold be done better if Subscription module could check if installation is a kickstart driven installation

pyanaconda/ui/gui/spokes/subscription.py
```diff
@@ -0,0 +1,769 @@
...
+
+        # Fixup Insights defaults
+        #
+        # FIXME: could be done better in module if module can see if kickstart was used
+        #        for the given installation and set InsightsEnabled to True
+        #
+        # If we have a kickstart, use the value from the module, otherwise enable Insights
+        # as that's the default for interactive installations.
+        if flags.automatedInstall:
+            self._insights_checkbox.set_active(self._subscription_module.InsightsEnabled)
```
poncovka, line 552: This is already set up in `_update_spoke_state`. I think you should set the DBus property `InsightsEnabled` to `True` for interactive installations before you call `_update_spoke_state` (ideally at the beginning of the `initialize` or `__init__` method).

pyanaconda/ui/gui/spokes/lib/subscription.py
```diff
@@ -0,0 +1,92 @@
...
+    - the value matches one of the valid values, so we preselect the valid value
+    - the value does not match a valid value, so we append a custom value
+      to the list and preselect it
+    - the user provided value is not available (empty string), no matching will be done
+      and no value will be preselected
+
+    :param str user_provided_value: a value provided by user
+    :param list valid_values: a list of valid values
+    :returns: list of values with one value preselected
+    :rtype: list of (str, str, bool) tuples in (<value>, <display string>, <is_preselected>) format
```
jkonecny12, line 36: What about using named tuple. It's really inconvenient to read code with three tuple because you don't know what is what and [2] doesn't really help here.

Or mark this as private method and leave it as it is. I don't mind if it is not an API.

pyanaconda/ui/gui/spokes/subscription.py
```diff
@@ -0,0 +1,769 @@
...
+        self.http_proxy_revealer.set_reveal_child(checkbox.get_active())
+
+    def on_http_proxy_location_entry_changed(self, editable):
+        hostname = ""
+        port = -1  # not set == -1
+        entered_text = editable.get_text()
+        if entered_text:
+            proxy_obj = ProxyString(url=entered_text)
+            hostname = proxy_obj.host
+            port = proxy_obj.port
```
jkonecny12, line 344: I think you want to remove this line. It's `-1` in case of not set and if set then the value is changed in the if below.

pyanaconda/ui/gui/spokes/subscription.py
M4rtinK: Nice solution, let's do that! :)

pyanaconda/ui/gui/spokes/subscription.py
```diff
@@ -0,0 +1,769 @@
...
+            self.authentication_method = AuthenticationMethod.ORG_KEY
+            self._organization_entry.set_text(self.subscription_request.organization)
+            set_in_entry = bool(self._activation_key_entry.get_text())
+            set_in_module = self.subscription_request.activation_keys.type == SECRET_TYPE_HIDDEN
+            if set_in_module:
+                if not set_in_entry:
+                    self.enable_activation_key_placeholder(True)
+            else:
+                self._activation_key_entry.set_text("")
+                self.enable_activation_key_placeholder(False)
```
jkonecny12, line 618: This could be separated to a private method with the comment having as docstring.

pyanaconda/ui/gui/spokes/subscription.py
```diff
@@ -0,0 +1,769 @@
...
+        # HTTP proxy password
+        set_in_entry = bool(self._http_proxy_password_entry.get_text())
+        secret_type = self.subscription_request.server_proxy_password.type
+        set_in_module = secret_type == SECRET_TYPE_HIDDEN
+        if set_in_module:
+            if not set_in_entry:
+                self.enable_http_proxy_password_placeholder(True)
+        else:
+            self._http_proxy_password_entry.set_text("")
+            self.enable_http_proxy_password_placeholder(False)
```
jkonecny12, line 666: Second part which can be moved to a private method.

pyanaconda/ui/gui/spokes/lib/subscription.py
M4rtinK: Yep, named tuple looks better, let's use it. :)

pyanaconda/ui/gui/spokes/subscription.py
M4rtinK: Yep, that's redundant, lets remove it.

pyanaconda/ui/gui/spokes/subscription.py
M4rtinK: Yeah, that looks a bit better.

pyanaconda/ui/gui/spokes/subscription.py
M4rtinK: Moved.

pyanaconda/ui/gui/spokes/subscription.py
```diff
@@ -0,0 +1,830 @@
...
+        self._registration_error = ""
+        self._registration_phase = None
+        self._registration_controls_enabled = True
+
+        # Red Hat Insights should be enabled by default for non-kickstart installs.
+        #
+        # For kickstart installations we will use the value from the module, which
+        # False by default & can be set to True via the rhsm kickstart command.
+        if not flags.automatedInstall:
+            self._subscription_module.InsightsEnabled = True
```
poncovka, line 109: I think that you need to call `SetInsightsEnabled`.

M4rtinK: Indeed! Fixed. :)

PR #2499, opened 2020-05-07
Remove a missed use of has_source() to fix tests

VladimirSlavik:
Followup to the botched #2494


tests/nosetests/pyanaconda_tests/module_payload_dnf_test.py
```diff
@@ -48,10 +48,7 @@ def _check_properties(self, expected_source_type):
         self.assertIsInstance(payload, DNFModule)
 
         # verify sources set
-        if expected_source_type is None:
-            self.assertFalse(payload.has_source())
```
poncovka, line 52: I guess we should check `self.assertFalse(payload.sources)`.

VladimirSlavik: Fixed...

PR #2500, opened 2020-05-11
Add __repr__ to payload sources

VladimirSlavik:
SSIA...

### DIFF
--- a/pyanaconda/modules/payloads/source/harddrive/harddrive.py
+++ b/pyanaconda/modules/payloads/source/harddrive/harddrive.py
@@ -64,6 +64,13 @@ class HardDriveSourceModule(PayloadSourceBase, RPMSourceMixin):
         """Get description of this source."""
         return "{}:{}".format(self._device, self._directory)
 
+    def __repr__(self):
+        result = "Source(type='HDD', partition='{}', directory='{}')".format(
+            self._device,
+            self.directory,
+        )
+        return result
+
     def get_state(self):
         """Get state of this source."""
         res = os.path.ismount(self._device_mount) and bool(self._install_tree_path)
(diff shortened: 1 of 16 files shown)


tests/nosetests/pyanaconda_tests/module_source_nfs_test.py
```diff
@@ -114,6 +114,13 @@ def url_property_test(self):
         self.module.set_url(nfs_url)
         self.assertEqual(nfs_url, self.module.url)
 
+    def repr_test(self):
+        self.module.set_url(nfs_url)
+        self.assertEqual(
+            repr(self.module),
+            "Source(type='NFS', url='{}')".format(nfs_url)
```
poncovka, line 121: Please, don't use the final string instead of the formatting.

pyanaconda/modules/payloads/source/harddrive/harddrive.py
```diff
@@ -64,6 +64,15 @@ def description(self):
         """Get description of this source."""
         return "{}:{}".format(self._device, self._directory)
 
+    def __repr__(self):
+        result = "Source(type='HDD', partition='{}', directory='{}', install_tree_path='{}')" \
```
poncovka, line 68: I am not sure if there should `install_tree_path` in the string. What do you think, @jkonecny12?

tests/nosetests/pyanaconda_tests/module_source_nfs_test.py
VladimirSlavik: You mean, I should get "NFS" there via `format` too?

pyanaconda/modules/payloads/source/harddrive/harddrive.py
VladimirSlavik: I agree it should be removed, all other sources have only "inputs" displayed there.

VladimirSlavik: Removed.

tests/nosetests/pyanaconda_tests/module_source_nfs_test.py
poncovka: My mistake. It should be "use the final string instead of the formatting". It is better to be sure that the output really matches the expected value.

VladimirSlavik: ACK, will do.

PR #2501, opened 2020-05-11
Work around pylint 2.5.0 issue with _ in warning id

VladimirSlavik:
Pylint 2.5.0 chokes on disable pragmas that contain `_`. See https://github.com/PyCQA/pylint/issues/3604

*This does NOT fix all errors after upgrading to pylint 2.5.0, only one of them.*

### DIFF
--- a/anaconda.py
+++ b/anaconda.py
@@ -376,7 +376,8 @@ if __name__ == "__main__":
             # The module-level _() calls are ok here because the language may
             # be set from the live environment in this case, and anaconda's
             # language setup hasn't happened yet.
-            # pylint: disable=found-_-in-module-class
+            # FIXME: change the line below back to found-_-in-module-class once it works in pylint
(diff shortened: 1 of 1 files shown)


VladimirSlavik:
@bcl just a heads up - this might hit you too, I've seen the same line in pykickstart.

anaconda.py
```diff
@@ -376,7 +376,7 @@ def setup_environment():
             # The module-level _() calls are ok here because the language may
             # be set from the live environment in this case, and anaconda's
             # language setup hasn't happened yet.
-            # pylint: disable=found-_-in-module-class
+            # pylint: disable=W9902
```
jkonecny12, line 379: Please add a fixme to set this back to the name because we want to get rid of these ID's which don't give you a hint about what is ignored.

VladimirSlavik: Done!

PR #2502, opened 2020-05-12
Create a DBus proxy of the DNF module

poncovka:
An instance of class DNFPayload can access its DBus counterpart via the property
proxy. The property will be eventually moved to the class Payload and created
for each type of the payload.

Refactorize the DBus API for working with payload modules. The current design
is based on the API for the partitioning modules.

### DIFF
--- a/pyanaconda/ui/lib/payload.py
+++ b/pyanaconda/ui/lib/payload.py
@@ -1,0 +1,61 @@
+#
+# Copyright (C) 2020  Red Hat, Inc.
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# the GNU General Public License v.2, or (at your option) any later version.
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY expressed or implied, including the implied warranties of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General
+# Public License for more details.  You should have received a copy of the
+# GNU General Public License along with this program; if not, write to the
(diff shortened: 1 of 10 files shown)


pyanaconda/ui/lib/payload.py
```diff
@@ -0,0 +1,58 @@
+#
```
jkonecny12, line 1: I think this code should be somewhere in the `/pyanaconda/payload` not UI code.

pyanaconda/ui/lib/payload.py
```diff
@@ -0,0 +1,58 @@
...
+    object_path = payloads_proxy.CreatePayload(payload_type)
+
+    if activate:
+        payloads_proxy.ActivatePayload(object_path)
+
+    return PAYLOADS.get_proxy(object_path)
+
+
+def get_payload(payload_type=PAYLOAD_TYPE_DNF):
+    """Get a payload of the specified type.
```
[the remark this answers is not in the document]
poncovka: Fixed.

pyanaconda/ui/lib/payload.py
poncovka: `pyanaconda.payload` will be removed. We usually put these functions to `pyanaconda.core` when they are shared with modules or to `pyanaconda.ui.lib` when they are used only by the UI. Look at `pyanaconda.ui.lib.storage` and `pyanaconda.ui.lib.users`.

poncovka: It is true, that we might want to merge `pyanaconda.ui.lib` with `pyanaconda.core`.

jkonecny12: OK, thanks for explanation.

PR #2503, opened 2020-05-12
Add ISOPath API to Hard drive payload source

jkonecny12:
This API will return name of the ISO. It's a replacement for `dnf.payload.ISO_image` to get name of the ISO to GUI.

### DIFF
--- a/pyanaconda/modules/payloads/source/harddrive/harddrive.py
+++ b/pyanaconda/modules/payloads/source/harddrive/harddrive.py
@@ -52,7 +52,7 @@ class HardDriveSourceModule(PayloadSourceBase, RPMSourceMixin):
         self._iso_mount = MountPointGenerator.generate_mount_point(
             self.type.value.lower() + "-iso"
         )
-        self._uses_iso_mount = False
+        self._iso_name = ""
 
     @property
     def type(self):
@@ -160,12 +160,13 @@ class HardDriveSourceModule(PayloadSourceBase, RPMSourceMixin):
         :rtype: [Task]
         """
         tasks = []
-        if self._uses_iso_mount:
+        if self._iso_name:
             tasks.append(TearDownMountTask(self._iso_mount))
         tasks.append(TearDownMountTask(self._device_mount))
         return tasks
 
     def _handle_setup_task_result(self, task):
         result = task.get_result()
-        self._install_tree_path, self._uses_iso_mount = result
+        self._install_tree_path = result.install_tree_path
+        self._iso_name = result.iso_name
         log.debug("Hard drive install tree path is set to '%s'", self._install_tree_path)
--- a/pyanaconda/modules/payloads/source/harddrive/harddrive.py
+++ b/pyanaconda/modules/payloads/source/harddrive/harddrive.py
@@ -55,6 +55,11 @@ class HardDriveSourceModule(PayloadSourceBase, RPMSourceMixin):
         self._iso_name = ""
 
     @property
+    def is_iso_mounted(self):
+        """Is ISO file mounted from set up task?"""
+        return bool(self._iso_name)
+
+    @property
     def type(self):
         """Get type of this source."""
         return SourceType.HDD
@@ -160,7 +165,7 @@ class HardDriveSourceModule(PayloadSourceBase, RPMSourceMixin):
         :rtype: [Task]
         """
         tasks = []
-        if self._iso_name:
+        if self.is_iso_mounted:
             tasks.append(TearDownMountTask(self._iso_mount))
         tasks.append(TearDownMountTask(self._device_mount))
         return tasks
--- a/pyanaconda/modules/payloads/source/harddrive/harddrive.py
+++ b/pyanaconda/modules/payloads/source/harddrive/harddrive.py
@@ -24,6 +24,7 @@ from pykickstart.errors import KickstartParseError
 from pyanaconda.anaconda_loggers import get_module_logger
 from pyanaconda.core.i18n import _
 from pyanaconda.core.signal import Signal
+from pyanaconda.core.util import join_paths
 from pyanaconda.modules.common.structures.payload import RepoConfigurationData
 from pyanaconda.modules.payloads.constants import SourceType, SourceState
 from pyanaconda.modules.payloads.source.source_base import PayloadSourceBase, RPMSourceMixin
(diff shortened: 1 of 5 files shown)


pyanaconda/modules/payloads/source/harddrive/harddrive.py
```diff
@@ -160,12 +166,24 @@ def tear_down_with_tasks(self):
         :rtype: [Task]
         """
         tasks = []
-        if self._uses_iso_mount:
+        if self.is_iso_mounted:
             tasks.append(TearDownMountTask(self._iso_mount))
         tasks.append(TearDownMountTask(self._device_mount))
         return tasks
 
+    def iso_path(self):
```
poncovka, line 174: Same here.

pyanaconda/modules/payloads/source/harddrive/initialization.py
```diff
@@ -51,8 +57,8 @@ def run(self):
         order again.
 
         :raise: SourceSetupError
-        :return: path to the install tree
-        :rtype: str
+        :return: named tuple with path to the install tree and name of ISO if set or empty string
+        :rtype: namedtuple("SetupHardDriveResult", ["install_tree_path", "iso_name"])
```
poncovka, line 61: I think it should be enough to specify the type as an instance of `SetupHardDriveResult`.

pyanaconda/modules/payloads/source/harddrive/initialization.py
```diff
@@ -71,9 +72,15 @@ def run(self):
...
 
-        if find_and_mount_iso_image(full_path_on_mounted_device, self._iso_mount):
-            return self._iso_mount, True
-        elif verify_valid_installtree(full_path_on_mounted_device):
+        iso_name = find_first_iso_image(full_path_on_mounted_device)
+
+        full_path_to_iso = join_paths(full_path_on_mounted_device, iso_name)
+
+        if iso_name:
+            if mount_iso_image(full_path_to_iso, self._iso_mount):
```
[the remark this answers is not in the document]
jkonecny12: Because the second one is a function doing an actual work. It seems easy to miss that if it's an "complicated" `if` statement.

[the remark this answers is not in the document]
VladimirSlavik: Thanks! To be honest, to me this also looks more readable than one line with AND... So I'm not exactly agreeing with the thumbs up I got ;-)

(It's fascinating how much of "engineering" can be perception and opinion-driven.)

pyanaconda/modules/payloads/source/harddrive/harddrive.py
```diff
@@ -170,6 +171,17 @@ def tear_down_with_tasks(self):
...
+        """Get path to the ISO from the partition root.
+
+        This could be an empty string if the source is pointing to
+        installation tree instead of ISO.
+
+        :return: path to the ISO or empty string if no ISO is involved
+        :rtype: str
+        """
+        return join_paths(self.directory, self._iso_name)
```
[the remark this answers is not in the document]
jkonecny12: Ahh, good point. I will add condition if `_iso_name` is empty to return an empty string.

pyanaconda/modules/payloads/source/harddrive/harddrive_interface.py
```diff
@@ -52,3 +52,11 @@ def Partition(self) -> Str:
     def SetPartition(self, partition: Str):
         """Set the partition containing the repository."""
         self.implementation.set_device(partition)
+
+    def IsoPath(self) -> Str:
```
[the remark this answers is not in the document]
jkonecny12: Fixed.

pyanaconda/modules/payloads/source/harddrive/harddrive.py
jkonecny12: Fixed.

pyanaconda/modules/payloads/source/harddrive/initialization.py
jkonecny12: Fixed.

pyanaconda/modules/payloads/source/harddrive/harddrive.py
```diff
@@ -54,6 +54,11 @@ def __init__(self):
         )
         self._iso_name = ""
 
+    @property
+    def is_iso_mounted(self):
```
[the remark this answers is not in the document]
jkonecny12: We can change this in future. But in general I've rather add code than comment to the code. 

For me it's much self explanation to use:
```
if self.is_iso_mounted:
```
than
```
if self._iso_name:
```

The first is check if the iso is mounted so I can react to that.
The second one is check if the iso name is set what can have several reasons so I don't know how to approach this when reading the code. I know it can be explained by comment but then you are using explanation of your code instead of changing the code to be self-explanatory. 

But yes, I agree that both approaches have the same result here. I just like the second one more.

jkonecny12:
**UPDATED:**
* Implemented notes from @VladimirSlavik and @poncovka . Thanks for your suggestions!

tests/nosetests/pyanaconda_tests/module_source_harddrive_test.py
```diff
@@ -122,23 +135,22 @@ def ready_state_test(self, ismount):
...
         ismount.assert_called_once_with(self.module._device_mount)
 
-
     def return_handler_test(self):
         """Hard drive source setup result propagates back."""
         task = _create_setup_task()
         # Test only the returning. To do that, fake what the magic in start() does.
         # Do not run() the task at all, less mocking needed that way.
-        task._set_result((iso_mount_location, True))
+        task._set_result(SetupHardDriveResult(iso_mount_location, "iso_name.iso"))
```
VladimirSlavik, line 149: An explicit test for the inverse situation, where you'd call `_set_result` with the second string empty, is missing. I believe that's why you didn't catch the "always true" problem.  Coverage is 100%, so the code path is taken in some other test. 

Maybe you could add the assert for `is_iso_mounted==False` in `success_find_dir_test`? And then also for `is_iso_mounted==True` in `success_find_iso_test`.

jkonecny12: > An explicit test for the inverse situation, where you'd call _set_result with the second string empty, is missing. I believe that's why you didn't catch the "always true" problem. 

Good point I'll extend that test.

> Coverage is 100%, so the code path is taken in some other test.

Coverage is 100% because I've added this test part when fixing this.
https://github.com/rhinstaller/anaconda/pull/2503/files#diff-74093229cac143f46d63bb39fee26e08R98

> Maybe you could add the assert for is_iso_mounted==False in success_find_dir_test? And then also for is_iso_mounted==True in success_find_iso_test.

I'm not sure I follow you here. I can't check `is_iso_mounted` because these tests are working with task only. Module/Interface is not involved as far as I know.

jkonecny12:
**UPDATED:**
* Add new test for return handler without ISO image situation.
* Improve return handler tests. They are not using private methods anymore.

tests/nosetests/pyanaconda_tests/module_source_harddrive_test.py
VladimirSlavik: Sorry, my bad, I only skimmed that part.

PR #2504, opened 2020-05-12
Add DeviceName API to CDrom source

jkonecny12:
Add `DeviceName` API to the CDrom source which will provide name of the device for the GUI.


jkonecny12:
@poncovka could you please look on the DBus property signaling for the read-only property here?

jkonecny12:
**UPDATED:**
* Resolve conflict.

PR #2505, opened 2020-05-12
Pylint 2.5.0 fixes

VladimirSlavik:
This consists of fixes to anaconda codebase that I couldn't match to pylint bugs.

### DIFF
--- a/pyanaconda/modules/storage/bootloader/base.py
+++ b/pyanaconda/modules/storage/bootloader/base.py
@@ -88,8 +88,9 @@ class BootLoaderError(Exception):
     pass
 
 
-class Arguments(OrderedSet):
+class Arguments(OrderedSet):  # pylint: disable=duplicate-bases
     """An ordered set of arguments."""
+    # FIXME: the pylint pragma is needed since 2.5.0 for unclear reasons
 
     def _merge_ip(self):
         """Find ip= arguments targeting the same interface and merge them."""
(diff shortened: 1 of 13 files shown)


pyanaconda/modules/storage/bootloader/base.py
```diff
@@ -88,7 +88,7 @@ class BootLoaderError(Exception):
     pass
 
 
-class Arguments(OrderedSet):
+class Arguments(OrderedSet):  # pylint: disable=duplicate-bases
```
jkonecny12, line 91: This seems to me like a pylint error. Or do you know why it is happening?

VladimirSlavik:
@jkonecny Good point about the gui tests. Not sure if we want to spend time on them now, though? I will file a task for later anyway.

pyanaconda/modules/storage/bootloader/base.py
VladimirSlavik: Yes, I think it's a pylint error. `Arguments` inherits from a standard library class, and as far as I can reach, there is no sign of multiple parents.

Do you think a FIXME would be good here?

jkonecny12:
Thanks @VladimirSlavik !

VladimirSlavik:
Jenkins, test this please.

pyanaconda/modules/storage/bootloader/base.py
poncovka: Btw. OrderedSet is not a standard library class :-)

VladimirSlavik: Mea culpa...

VladimirSlavik: Added the FIXME for this one.

VladimirSlavik:
Update: added a fixme note for the duplicate bases pragma.

VladimirSlavik:
Jenkins, test this please.

PR #2506, opened 2020-05-12
subscription: Add the ParseAttachedSubscriptionsTask

M4rtinK:
The task parses attached subscriptions as well as final
system purpose data from the installation environment
after a successful subscription.

(System purpose data can change if an activation key
with attached system purpose data is used. Due to this we
need to parse system purpose data after subscription has
been attached to avoid the Subscription spoke showing stale
system purpose data after a subscription has been attached.)

We connect to the completed signal of the task to automatically
set attached subscriptions and system purpose data when the task
succeeds. And we do the same in reverse for the unregister task.

### DIFF
--- a/pyanaconda/modules/subscription/runtime.py
+++ b/pyanaconda/modules/subscription/runtime.py
@@ -17,6 +17,8 @@
 #
 import os
 import json
+import datetime
+from collections import namedtuple
 
 from dasbus.typing import get_variant, Str
 from dasbus.connection import MessageBus
@@ -29,6 +31,8 @@ from pyanaconda.modules.common.constants.services import RHSM
 from pyanaconda.modules.common.constants.objects import RHSM_REGISTER
 from pyanaconda.modules.common.errors.subscription import RegistrationError, \
     UnregistrationError, SubscriptionError
+from pyanaconda.modules.common.structures.subscription import AttachedSubscription, \
+    SystemPurposeData
 
 from pyanaconda.modules.subscription import system_purpose
 
@@ -68,6 +72,10 @@ class RHSMPrivateBus(MessageBus):
         # so we will not log it
         log.info("Connecting to the RHSM private DBus session.")
         return self._provider.get_addressed_bus_connection(self._private_bus_address)
+
+
+SystemSubscriptionData = namedtuple("SystemSubscriptionData",
+                                    ["attached_subscriptions", "system_purpose_data"])
 
 
 class SystemPurposeConfigurationTask(Task):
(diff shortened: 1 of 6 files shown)


pyanaconda/modules/subscription/runtime.py
```diff
@@ -360,3 +368,200 @@ def run(self):
...
+            # - looks like productId == SKU in this JSON output
+            attached_subscription.sku = subscription_info.get(
+                "sku",
+                _("unknown")
+            )
+
+            # contract number
+            attached_subscription.contract = subscription_info.get(
+                "contract",
+                _("Not Available")
```
poncovka, line 456: This string is title cased, but the other strings are lower cased. Maybe it could be unified?

M4rtinK: Good point, let's do that. :)

pyanaconda/modules/subscription/runtime.py
```diff
@@ -360,3 +368,200 @@ def run(self):
...
+        super().__init__()
+        self._rhsm_entitlement_proxy = rhsm_entitlement_proxy
+        self._rhsm_syspurpose_proxy = rhsm_syspurpose_proxy
+
+    @property
+    def name(self):
+        return "Parse attached subscription data"
+
+    @classmethod
+    def _pretty_date(self, date_from_json):
```
jkonecny12, line 391: This should be ``@staticmethod`` and the `self` should be removed.

pyanaconda/modules/subscription/runtime.py
```diff
@@ -360,3 +368,200 @@ def run(self):
...
+            #   *or* the pool-level "quantity" key
+            quantity_string = int(subscription_info.get("quantity_used", 1))
+            attached_subscription.consumed_entitlement_count = quantity_string
+            # add attached subscription to the list
+            attached_subscriptions.append(attached_subscription)
+        # return the list of attached subscriptions
+        return attached_subscriptions
+
+    @classmethod
+    def _parse_system_purpose_json(cls, final_syspurpose_json):
```
jkonecny12, line 487: Switch this to `@staticmethod`.

pyanaconda/modules/subscription/runtime.py
```diff
@@ -360,3 +368,200 @@ def run(self):
...
+            # See bug 1793501 for possible issues with RHSM provided date parsing.
+            date = datetime.datetime.strptime(date_from_json, "%m/%d/%y")
+            # get a nice human readable date
+            date_string = date.strftime("%b %d, %Y")
+        except ValueError:
+            log.warning("subscription: date parsing failed: %s", date_from_json)
+        return date_string
+
+    @classmethod
+    def _parse_subscription_json(cls, subscription_json):
```
jkonecny12, line 407: Switch this to `@staticmethod`.

pyanaconda/modules/subscription/subscription.py
```diff
@@ -355,6 +360,32 @@ def set_subscription_request(self, subscription_request):
...
+        The list holds DBus structures with each structure holding information about
+        one attached subscription. A system that has been successfully registered and
+        subscribed usually has one or more subscriptions attached.
+
+        :return: list of DBus structures, one per attached subscription
+        :rtype: list of AttachedSubscription instances
+        """
+        return self._attached_subscriptions
+
+    def set_attached_subscriptions(self, attached_subscriptions):
```
jkonecny12, line 376: Shouldn't this be a private method if it is not an API of this class?

pyanaconda/modules/subscription/runtime.py
M4rtinK: This needs to call `_pretty_date()`, so needs access to the class to call it in a clean way.

pyanaconda/modules/subscription/runtime.py
M4rtinK: Yep, that's cleaner. :)

pyanaconda/modules/subscription/runtime.py
M4rtinK: Yep, that's a good idea. :)

pyanaconda/modules/subscription/subscription.py
M4rtinK: We have it like this in a couple places for other properties & I think I saw something similar in other modules. So I think such a change is more suitable for a followup PR.

M4rtinK:
Updated based on @poncovka and @jkonecny12 feedback, thanks! :)

pyanaconda/modules/subscription/runtime.py
```diff
@@ -360,3 +368,200 @@ def run(self):
...
+            attached_subscription.name = subscription_info.get(
+                "subscription_name",
+                _("product name unknown")
+            )
+
+            # subscription support level
+            # - this does *not* seem to directly correlate to system purpose SLA attribute
+            attached_subscription.service_level = subscription_info.get(
+                "service_level",
+                _("unknown")
```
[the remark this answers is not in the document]
M4rtinK: Making a note to add context in followup PR.

M4rtinK:
> LGTM, insofar I can understand it. Aren't the tests duplicated though?
I don't see any duplication - we test the methods the task calls, the property & task creation. Pretty much the same we do with other tasks.

PR #2509, opened 2020-05-12
Add functions for DBus sources

poncovka:
Add helpers functions create_source, set_source and get_source that should
simplify manipulation with DBus sources of the Payloads module.

### DIFF
--- a/pyanaconda/ui/lib/payload.py
+++ b/pyanaconda/ui/lib/payload.py
@@ -15,6 +15,7 @@
 # License and may only be used or replicated with the express permission of
 # Red Hat, Inc.
 #
+from dasbus.client.proxy import get_object_path
 from pyanaconda.anaconda_loggers import get_module_logger
 from pyanaconda.core.constants import PAYLOAD_TYPE_DNF
 from pyanaconda.modules.common.constants.services import PAYLOADS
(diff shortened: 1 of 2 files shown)


pyanaconda/ui/lib/payload.py
```diff
@@ -56,3 +57,58 @@ def get_payload(payload_type=PAYLOAD_TYPE_DNF):
...
+
+def get_source(payload_proxy, source_type=None):
+    """Get a source of the given payload.
+
+    If the payload has one or more sources, return the first one.
+
+    If the payload has no sources and the default source type
+    is specified, create a default source.
+
+    If the payload has not sources and the default source type
```
jkonecny12, line 91:
```suggestion
    If the payload has no sources and the default source type
```

pyanaconda/ui/lib/payload.py
```diff
@@ -56,3 +57,58 @@ def get_payload(payload_type=PAYLOAD_TYPE_DNF):
...
+    """Attach the source to the payload.
+
+    :param payload_proxy: a DBus proxy of a payload
+    :param source_proxy: a DBus proxy of a source
+    """
+    object_path = get_object_path(source_proxy)
+    payload_proxy.SetSources([object_path])
+
+
+def get_source(payload_proxy, source_type=None):
```
jkonecny12, line 83: Please rename `source_type` to `default_source_type`. This way it seems to return source of the given type.

pyanaconda/ui/lib/payload.py
poncovka: Fixed.

pyanaconda/ui/lib/payload.py
poncovka: Fixed.

poncovka:
Rebased and fixed.

pyanaconda/ui/lib/payload.py
```diff
@@ -59,3 +60,58 @@ def get_payload(payload_type=PAYLOAD_TYPE_DNF):
...
+    :param source_proxy: a DBus proxy of a source
+    """
+    object_path = get_object_path(source_proxy)
+    payload_proxy.SetSources([object_path])
+
+
+def get_source(payload_proxy, default_source_type=None):
+    """Get a source of the given payload.
+
+    If the payload has one or more sources, return the first one.
```
VladimirSlavik, line 89: What's that good for? I get something, and don't know if there is more?

poncovka: It is a temporary workaround, because the DBus API is designed to work with multiple sources, but UI doesn't support them, so it will ignore other sources for now.

VladimirSlavik: Ah, ok.

PR #2510, opened 2020-05-13
subscription: Asynchronous registration/unregistration handling

M4rtinK:
The first commit adds the `IsRegistered` property to the Subscription DBus module API. We need this property to detect if system is currently registered but not subscribed.

The second commit adds the asynchronous registration+subscription and unregistration helper functions. Both functions use the same "API" - a progress callback and and error reporting callback.

During normal operations the functions call the progress callback with values specified by the `SubscriptionPhase` enum. On success `SubscriptionPhase.DONE` phase is used and the function terminates. On error, the error callback is called with appropriate error message and the function terminates.

It is expected these helper functions will be started in a thread, as the work they are doing (especially in the register & subscribe) case can take in extreme cases tens of seconds or even up to a minute.

**TODO**
- use the helper functions during automated installation
- use the helper functions from the GUI


M4rtinK:
jenkins, test this please

pyanaconda/core/constants.py
```diff
@@ -467,3 +467,11 @@ class PayloadRequirementType(Enum):
 
 # Path to the System Purpose configuration file on a system.
 RHSM_SYSPURPOSE_FILE_PATH = "/etc/rhsm/syspurpose/syspurpose.json"
+
+
+# Asynchronous subscription state tracking
+class SubscriptionPhase(Enum):
```
poncovka, line 473: I would put this enum to the same file as `register_and_subscribe` and `unregister`.

pyanaconda/core/subscription.py
```diff
@@ -38,3 +53,191 @@ def check_system_purpose_set(sysroot="/"):
...
+    """Dummy progress reporting function used if no custom callback is set."""
+    pass
+
+
+def dummy_error_callback(error_message):
+    """Dummy error reporting function used if no custom callback is set."""
+    pass
+
+
+def register_and_subscribe(progress_callback=None, error_callback=None):
```
poncovka, line 90: I would put these functions to `pyanaconda.ui.lib.subscription`, because they are interesting only for UI.

tests/nosetests/pyanaconda_tests/module_subscription_tests.py
```diff
@@ -897,6 +897,29 @@ def insights_property_test(self):
...
+
+        def custom_setter(value):
+            self.subscription_module.set_registered(value)
+
+        # check the property is True and the signal was emitted
+        # - we use fake setter as there is no public setter
+        self._check_dbus_property(
+          "IsRegistered",
+          True,
+          setter=custom_setter
```
jkonecny12, line 917: Couldn't you just add here ``self.subscription_module.set_registered`` directly?

pyanaconda/core/subscription.py
```diff
@@ -38,3 +53,191 @@ def check_system_purpose_set(sysroot="/"):
...
+    """Try to register and subscribe the installation environment.
+
+    :param progress_callback: progress callback function, takes one argument, subscription phase
+    :type progress_callback: a callable
```
jkonecny12, line 94: I think you should have a declaration here.
```
callable(subscription_phase)
```

pyanaconda/core/subscription.py
```diff
@@ -38,3 +53,191 @@ def check_system_purpose_set(sysroot="/"):
...
+        task_proxy = SUBSCRIPTION.get_proxy(task_path)
+        try:
+            task.sync_run_task(task_proxy)
+        except UnregistrationError as e:
+            log.debug("subscription thread: unregistration failed: %s", e)
+            # Failing to unregister the system is an unrecoverable error,
+            # so we end there.
+            error_callback(str(e))
+            return
+        log.debug("Subscription GUI: unregistration succeeded")
```
jkonecny12, line 142: I would abstract this to a private method.

pyanaconda/core/subscription.py
jkonecny12, line 90: I see this a too big method to be just one method. If I'm counting correctly it can be split to about 5 parts.

I won't block this PR on this.

pyanaconda/core/subscription.py
```diff
@@ -38,3 +53,191 @@ def check_system_purpose_set(sysroot="/"):
...
+    log.debug("subscription thread: attempting to auto attach an entitlement")
+    progress_callback(SubscriptionPhase.ATTACH_SUBSCRIPTION)
+    task_path = subscription_proxy.AttachSubscriptionWithTask()
+    task_proxy = SUBSCRIPTION.get_proxy(task_path)
+    try:
+        task.sync_run_task(task_proxy)
+    except SubscriptionError as e:
+        log.debug("subscription thread: failed to attach subscription: %s", e)
+        error_callback(str(e))
+        return
```
jkonecny12, line 194: Another nice part for abstraction.

pyanaconda/core/subscription.py
M4rtinK: On RHEL 8.2 this is invoked from the top-level `anaconda.py` file for kickstart based installations & it seems weird to me to import from `pyanaconde.ui` inside there. But we apparently already do that: https://github.com/rhinstaller/anaconda/blob/8afb0d050936ba35d966cc056ee49afe91b166bb/anaconda.py#L650

(IIRC we wanted to be able to run Anaconda without the UI, but maybe it was just without the `tui` & `gui` folders ?)

pyanaconda/modules/subscription/subscription.py
```diff
@@ -415,6 +419,35 @@ def set_connect_to_insights(self, connect):
...
+              the system is registered but has not subscription
+              attached. This is generally a sign something went
+              wrong, usually when trying to attach subscription.
+
+        :return: True if the system has been registered, False otherwise
+        :rtype: bool
+        """
+        return self._registered
+
+    def set_registered(self, system_registered):
```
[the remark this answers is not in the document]
M4rtinK: As mentioned in https://github.com/rhinstaller/anaconda/pull/2506#discussion_r424475326 we something similar in couple other places already, so if we want to change it we should do it consistently in a separate PR.

tests/nosetests/pyanaconda_tests/module_subscription_tests.py
M4rtinK: AFAIK we want to use `_check_dbus_property()` to test properties where possible & all the other properties with private setters are tested like this.

pyanaconda/core/subscription.py
```diff
@@ -38,3 +53,191 @@ def check_system_purpose_set(sysroot="/"):
...
+    pass
+
+
+def register_and_subscribe(progress_callback=None, error_callback=None):
+    """Try to register and subscribe the installation environment.
+
+    :param progress_callback: progress callback function, takes one argument, subscription phase
+    :type progress_callback: a callable
+    :param error_callback: error callback function, takes one argument, the error message
+    :param error_callback: a callable
```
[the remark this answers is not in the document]
M4rtinK: Fixed!

pyanaconda/core/subscription.py
M4rtinK: I tried that and it was not worth it - it's ~5 lines and you would have to pass it the subscription proxy and then detect if there was an error or not. The invocation & return checking have the same or more lines.

pyanaconda/core/subscription.py
M4rtinK: This is basically a recipe on how to register and subscribe a system, that should consist of individual simple steps. Splitting it into individual private micro methods would make it harder to follow and introduce a lot of boilerplate.

pyanaconda/core/subscription.py
M4rtinK: See above.

pyanaconda/core/constants.py
M4rtinK: Yeah, we can move it.

pyanaconda/core/subscription.py
M4rtinK: Yep, that looks better. :)

pyanaconda/core/subscription.py
M4rtinK: Also, longer term we will likely want to use some generic way of running a batch of DBus tasks both during installation and cases like this with runtime tasks.

M4rtinK:
Updated the PR with a couple commits that hook the asynchronous code to automated installation and to GUI.

A couple notes per commit:


**subscription: Register & subscribe during automated installation** f6cb59b
* I hope this is the correct spot for triggering the subscription thread. In 8.2 it was run earlier, possibly even for rescue mode, which could potentially cause issues. If you think the thread should be started earlier or later, please let me know. :)
* Also I'm not really sure about error handling at this early stage - at the moment the progress and error callback are simply not connected. If registration fails, it will not crash the installation but will simply result in the machine not getting registered.

**subscription: Make it possible to register/unregister from the GUI** 989f44e
* This hooks up the registration/unregistration code to the GUI, making sure to handle the various bits and pieces (progress reporting, error reporting, disabling UI interaction when operations are ongoing, etc.)
* The progress and error reporting hooks are the same both for registration and unregistration.
* On RHEL 8.2 unregistration was actually not run in a thread but was a UI blocking operation, it was just not that visible due to usually not taking long.

**subscription: Hook up state update methods** f462a38
* This hooks up state update methods to the appropriate places.
* At the moment we have ~4 state update methods:
  * `_update_spoke()` which updates spoke state based on data from the module
  * `_update_registration_state()` updates the registration part of the spoke
  * `_update_registration_state()` updates the subscription part of the spoke
  * `_update_register_button_state()` updates state of the registration button
* Once all is in place, I guess we might want to check if this is fine or if we want to combine/split the methods differently.

pyanaconda/core/subscription.py
poncovka: `pyanaconda.ui.lib` doesn't depend on TUI or GUI and it is part of `anaconda-core`. We might want to eventually merge it with `pyanaconda.core`, but I would like to keep it separate for now.

anaconda.py
```diff
@@ -691,6 +691,27 @@ def _earlyExceptionHandler(ty, value, traceback):
...
+        from pyanaconda.modules.common.structures.secret import SECRET_TYPE_HIDDEN
+        from pyanaconda.core.subscription import register_and_subscribe
+        subscription_proxy = SUBSCRIPTION.get_proxy()
+        subscription_request_struct = subscription_proxy.SubscriptionRequest
+        subscription_request = SubscriptionRequest.from_structure(subscription_request_struct)
+        organization_set = bool(subscription_request.organization)
+        key_set = subscription_request.activation_keys.type == SECRET_TYPE_HIDDEN
+        if organization_set and key_set:
+            threadMgr.add(AnacondaThread(name=constants.THREAD_SUBSCRIPTION,
+                                         target=register_and_subscribe))
```
poncovka, line 713: Maybe `register_and_subscribe` could have an optional argument that would allow to skip the registration if the organization and the key are not set. Or there could be a new function called `try_register_and_subscribe`. Anyway, it would be good to move this block somewhere else if possible.

pyanaconda/core/subscription.py
poncovka: However, it is up to you. You can leave the functions in `pyanaconda.core` if you want to.

M4rtinK: Sounds good, I can move it. :)

anaconda.py
M4rtinK: Indeed, that really should be simpler. So I put that code to function (`org_keys_sufficient()` & `username_password_sufficient()`) that can be imported & used to check if starting the subscription thread makes sense.

I've also used these methods in the `register_and_subscribe()` method to avoid code duplication. In this case they are bit less efficient as they create they own proxies for the Subscription module, but given that this will be called basically once per installation, I don't think that's a problem.

tests/nosetests/pyanaconda_tests/module_subscription_tests.py
jkonecny12: I know. I don't mean to not use _check_dbus_property. I was talking about creating your custom setter which only call set_registered with the same value. So you can skip creating the custom_setter and just use the set_registered directly instead.

I should have point this to the custom setter to make it more obvious.

PR #2321, opened 2020-02-19
Detect HTTPS as a protocol that requires the network

michel-slm:
Currently the network is not automatically set up if all the repos are
HTTPS, because HTTPS is not in the whitelist of protocols that require
a network connection.

Signed-off-by: Michel Alexandre Salim <michel@fb.com>

---
This change is [Reviewable](https://reviewable.io/reviews/rhinstaller/anaconda/2321)

PR #2513, opened 2020-05-13
Add the DBus method IsNetworkRequired

poncovka:
Call the DBus method IsNetworkRequired of a payload module to check whether
the payload's sources require a network.

### DIFF
--- a/pyanaconda/modules/payloads/source/repo_files/repo_files.py
+++ b/pyanaconda/modules/payloads/source/repo_files/repo_files.py
@@ -43,6 +43,14 @@ class RepoFilesSourceModule(PayloadSourceBase):
         """Get description of this source."""
         return _("Closest mirror")
 
+    @property
+    def network_required(self):
+        """Does the source require a network?
+
+        :return: True or False
+        """
+        return True
+
     def __repr__(self):
         return "Source(type='REPO_FILES')"
 
--- a/pyanaconda/modules/payloads/source/source_base.py
+++ b/pyanaconda/modules/payloads/source/source_base.py
@@ -58,6 +58,15 @@ class PayloadSourceBase(KickstartBaseModule, Publishable, metaclass=ABCMeta):
         """
         pass
 
+    @property
+    @abstractmethod
+    def network_required(self):
+        """Does the source require a network?
+
+        :return: True or False
+        """
+        pass
+
     @abstractmethod
     def get_state(self):
         """Get state of this source.
--- a/pyanaconda/modules/payloads/source/utils.py
+++ b/pyanaconda/modules/payloads/source/utils.py
@@ -25,8 +25,18 @@ from pyanaconda.core.util import join_paths
 from pyanaconda.anaconda_loggers import get_module_logger
 log = get_module_logger(__name__)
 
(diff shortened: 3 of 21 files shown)


pyanaconda/modules/payloads/source/utils.py
```diff
@@ -27,7 +27,20 @@
...
+def has_network_protocol(url):
+    """Does the given URL have a network protocol?
+
+    :param url: a string with URL
+    :return: True or False
+    """
+    if not url:
+        return False
+
+    network_protocols = ["http:", "ftp:", "nfs:", "nfsiso:"]
```
jstodola, line 42: https is missing

VladimirSlavik: See #2321...

pyanaconda/modules/payloads/source/source_base.py
```diff
@@ -58,6 +58,14 @@ def description(self):
         """
         pass
 
+    @property
+    def network_required(self):
+        """Does the source require a network?
+
+        :return: True or False
+        """
+        return False
```
VladimirSlavik, line 67: Maybe this should be abstract like the rest of these properties? Not sure.

pyanaconda/modules/payloads/source/repo_files/repo_files.py
```diff
@@ -43,6 +43,14 @@ def description(self):
         """Get description of this source."""
         return _("Closest mirror")
 
+    @property
+    def network_required(self):
+        """Does the source require a network?
+
+        :return: True or False
+        """
+        return True
```
VladimirSlavik, line 52: Is this true in the mathematical sense, or just the best approximation? I could see repo files with only file://, but maybe these are special cases handled by other sources?

pyanaconda/modules/payloads/source/utils.py
poncovka: Fixed. I have used the commit from https://github.com/rhinstaller/anaconda/pull/2321.

pyanaconda/modules/payloads/source/source_base.py
poncovka: Fixed.

pyanaconda/modules/payloads/source/repo_files/repo_files.py
poncovka: It is the best approximation. This code will replace the implementation of the property `needs_network` from `DNFPayload` and we assume there that closest mirror requires network. See: https://github.com/rhinstaller/anaconda/blob/master/pyanaconda/payload/dnf/payload.py#L158

poncovka:
Rebased and fixed.

pyanaconda/modules/payloads/source/utils.py
VladimirSlavik: Thank you!

pyanaconda/modules/payloads/source/repo_files/repo_files.py
VladimirSlavik: Sorry, here I'm stupid again - of course that's how the code worked previously. Thank you for the explanation!

PR #2514, opened 2020-05-14
Consistent method ordering for payload sources

jkonecny12:
I don't insist on the second commit if you don't agree with that.


jkonecny12:
**UPDATED:**
* Rebase to solve conflicts.

PR #2517, opened 2020-05-15
subscription: Show data about attached subscriptions

M4rtinK:
This basically updates the second part of the spoke that
is shown after a successful subscrioption attempt. Type
of authentication used, final system purpose data as
well as information about attached subscriptions is shown.

The code that creates the listbox delegates describing an
attached subscription is not very nice, as it needs to
imperatively construct the rather complex listbox delegate
using the GTK/Pango API.

Ideally this would be done be describing the delegate in
Glade and then just instantiating it per item and filling
it with data. Unfortunately there is zero information
available for dynamically creating widgets with Glade
and I was simply not able to get it working in this way.

I guess this might explain why some of the dynamically
created widgets we have in Anaconda are implemented in C.

Also add a CSS file we use to make the list box background
transparent to make the delegates look better, especially
in case there is just one or two of them.

### DIFF
--- a/pyanaconda/ui/gui/spokes/lib/subscription.py
+++ b/pyanaconda/ui/gui/spokes/lib/subscription.py
@@ -17,12 +17,20 @@
 # Red Hat, Inc.
 #
 
+import gi
+gi.require_version("Gtk", "3.0")
+gi.require_version("Pango", "1.0")
+from gi.repository import Gtk, Pango
+
 from collections import namedtuple
 
 from pyanaconda.core.i18n import _
 
-TextComboBoxItem = namedtuple("TextComboBoxItem", ["value", "display_string", "is_preselected"])
+from pyanaconda.anaconda_loggers import get_module_logger
+log = get_module_logger(__name__)
 
+
+TextComboBoxItem = namedtuple("TextComboBoxItem", ["value", "display_string", "is_preselected"])
 
 def handle_user_provided_value(user_provided_value, valid_values):
     """Handle user provided value (if any) based on list of valid values.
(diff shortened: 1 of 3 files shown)


M4rtinK:
Not sure how to fix the PEP8 error in a clean way - maybe we have just not touched `gi.repository` importing code since we enabled the checker ?

poncovka:
> Not sure how to fix the PEP8 error in a clean way - maybe we have just not touched `gi.repository` importing code since we enabled the checker ?

You can ignore this error when it is caused by imports from `gi.repository`.

pyanaconda/ui/gui/spokes/lib/subscription.py
```diff
@@ -100,3 +108,135 @@ def fill_combobox(combobox, user_provided_value, valid_values):
 
     # set the active id (what item should be selected in the combobox)
     combobox.set_active_id(active_id)
+
+    def clear_listbox(listbox):
```
poncovka, line 112: It looks like these functions are nested in the function `fill_combobox`. Is it necessary?

pyanaconda/ui/gui/spokes/subscription.css
```diff
@@ -0,0 +1,4 @@
+.subscriptions_listbox {
```
poncovka, line 1: How does it work? I couldn't find the code that loads this file.

pyanaconda/ui/gui/spokes/lib/subscription.py
M4rtinK: Yeah, if it'S used just in a single place, we can drop the function and just include it.

pyanaconda/ui/gui/spokes/subscription.css
M4rtinK: It should bind to the GtkListBox named "subscriptions_listbox" in glade:
https://github.com/rhinstaller/anaconda/blob/master/pyanaconda/ui/gui/spokes/subscription.glade#L941

I *think* the dot in the front makes it bind to the a name, while other prefixes make ti bind to something else (I think I saw #, which does something else than the dot).

In any case, this should make the listbox invisible by itself, so it should not look bad in the usual case of systems having just a couple attached subscriptions yet it will provided the needed scrolling in cases there are more subscription delegates then what can fit on the screen.

VladimirSlavik: I think the problem was, "How is this file even loaded"?

pyanaconda/ui/gui/spokes/lib/subscription.py
poncovka: That is not necessary, I'm just not sure how it works. The functions `add_attached_subscription_delegate` and `populate_attached_subscriptions_listbox` are nested in the function `fill_combobox`, but `fill_combobox` doesn't call them.

M4rtinK: Oh, now I  see it! Yeah, the indentation is totally wrong - but given that it was wrong for all the functions I've added, I have completely missed that. :P Should be fixed now.

pyanaconda/ui/gui/spokes/subscription.css
M4rtinK: The answer is simple - it is not loaded, at all! GAH, damn you Glade & GTK! :P

 Turns out this got solved by including the CSS definitions in the main `anaconda-gtk.css` file and this was just some local leftover from an earlier attempt back then to have these definitions separately. 

Anyway, scrapping this file & putting the definitions to `anaconda-gtk.css` - that should be the correct state.

pyanaconda/ui/gui/spokes/subscription.py
```diff
@@ -970,7 +971,64 @@ def _update_subscription_state(self):
...
+        # role
+        final_role_string = _("Role: {}").format(self.system_purpose_data.role)
+        self._role_status_label.set_text(final_role_string)
+
+        # SLA
+        final_sla_string = _("SLA: {}").format(self.system_purpose_data.sla)
+        self._sla_status_label.set_text(final_sla_string)
+
+        # usage
+        final_usage_string = ("Usage: {}").format(self.system_purpose_data.usage)
```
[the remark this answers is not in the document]
M4rtinK: Indeed! There is even the `()` but not the underscore... On the bright sight, that's one more "fixed by rebase" bug. ;-)

Should be fixed now.

M4rtinK:
PR updated:
* fixed the indentation issues with the attached subscription listbox filling code
* added sane CSS definition
* added missing string translation

poncovka:
@M4rtinK, I think you didn't push the changes to the branch :-)

M4rtinK:
@poncovka Gah! I have really been looking into this for far too long. ;-) Should be finally up to date now. :)

M4rtinK:
jenkins, test this please

M4rtinK:
Even though GitHub shows "in progress" for the CI, when looking to Jenkins directly, both test runs have finished with success. So this should be clear for merging. :)

PR #2519, opened 2020-05-15
Rhel 8 rebase rhel8 ks commands

rvykydal:


dracut/parse-kickstart
```diff
@@ -43,7 +43,7 @@ from pykickstart.constants import *
 from collections import OrderedDict
 
 # Import the kickstart version.
-from pykickstart.version import DEVEL as VERSION
+from pykickstart.version import RHEL8 as VERSION
 
 # Import all kickstart commands as version-less.
```
poncovka, line 48: We should also replace the bootloader command below.

rvykydal: Fixed, thank you!

PR #2521, opened 2020-05-15
Update the bootloader configuration after live installation

martinezjavier:
The GRUB configuration was changed to not store the kernel cmdline as a
variable in the grubenv file anymore, but instead in the BLS snippets.

But this change broke the live installations because the kernel-install
grub2 plugin now takes the kernel cmdline from /proc/cmdline. That means
the BLS snippets will have the cmdline used to boot the live install and
not the one needed to boot the installed rootfs.

To make sure that the BLS files have the expected cmdline, call to the
grub2-mkconfig tool after all the kernel have been installed.

Resolves: rhbz#1836020

### DIFF
--- a/pyanaconda/payload/live/payload_base.py
+++ b/pyanaconda/payload/live/payload_base.py
@@ -27,6 +27,9 @@ from pyanaconda.core.configuration.anaconda import conf
 from pyanaconda.core.constants import INSTALL_TREE, THREAD_LIVE_PROGRESS
 from pyanaconda.core.i18n import _
 from pyanaconda.errors import errorHandler, ERROR_RAISE
+from pyanaconda.modules.common.constants.objects import BOOTLOADER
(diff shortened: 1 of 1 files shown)


poncovka:
Jenkins, test this please.

martinezjavier:
@jkonecny12 thanks for the review, I've done the suggested changes. Now `pylint` doesn't complain anymore.

jkonecny12:
Jenkins, it's ok to test.

AdamWill:
using `execInSysroot` should automatically log the command the result, shouldn't it?

poncovka:
@AdamWill, yes, it should be in `program.log`.

VladimirSlavik:
Yes, it does. However my suggestion was more about following the pattern in the method - other "sections" of code there always log what is about to happen.

PR #2522, opened 2020-05-15
Fix RHEL-8 tests on the rebase branch

jkonecny12:
These changes should not be problem for Rawhide tests. However, it will fail until blivet-gui is removed from the code base and tests which is correct failure.

Tests are still blocked by https://github.com/rhinstaller/anaconda/pull/2520 which should be merged first.

### DIFF
--- a/anaconda.spec.in
+++ b/anaconda.spec.in
@@ -19,7 +19,6 @@ Source0: %{name}-%{version}.tar.bz2
 # Versions of required components (done so we make sure the buildrequires
 # match the requires versions of things).
 
-%define blivetguiver 2.1.12-1
 %define dbusver 1.2.3
 %define dnfver 3.6.0
 %define dracutver 034-7
@@ -168,12 +167,6 @@ Requires: isomd5sum >= %{isomd5sum}
 %ifarch %{ix86} x86_64
 Requires: fcoe-utils >= %{fcoeutilsver}
 %endif
-# likely HFS+ resize support
-%ifarch %{ix86} x86_64
-%if ! 0%{?rhel}
-Requires: hfsplus-tools
-%endif
-%endif
 # kexec support
 Requires: kexec-tools
 Requires: createrepo_c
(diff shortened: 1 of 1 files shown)


scripts/testing/setup-mock-test-env.py
```diff
@@ -370,7 +370,9 @@ def install_pip_packages_to_mock(mock_command, packages):
     cmd = _prepare_command(mock_command)
 
     cmd = _run_cmd_in_chroot(cmd)
-    cmd.append('pip3 install --install-option="--install-scripts=/usr/bin" {}'.format(packages))
+    cmd.append(
+        'python3 -m pip install --install-option="--install-scripts=/usr/bin" {}'.format(packages)
```
VladimirSlavik, line 374: Will this work on RHEL? IIRC there is platform-python, python3.6 and such, and you use alternatives to set up what is python2, python3, python etc.

poncovka: `python3` should call python3.6 on RHEL 8, so it should be fine.

scripts/testing/dependency_solver.py
```diff
@@ -41,6 +41,8 @@
     "pykickstart",
     "python3-pip",
     "policycoreutils",  # contains restorecon which was removed in Fedora 28 mock
+    "libxml2-devel",  # sphinx dependency
+    "libxslt-devel",  # sphinx dependency
```
poncovka, line 45: I am confused. Why do we need sphinx dependencies?

jkonecny12:
**UPDATED:**
* Remove commits solved by other PR's.

https://github.com/rhinstaller/anaconda/pull/2520
https://github.com/rhinstaller/anaconda/pull/2528 <- thanks a lot for suggestion @poncovka !

scripts/testing/setup-mock-test-env.py
jkonecny12: platform-python is pre-installed but we are dragging in "user" python.

jkonecny12: It would be interesting to switch but I don't think it's really worth the effort.

jkonecny12:
This PR will be replaced by https://github.com/rhinstaller/anaconda/pull/2535.

jkonecny12:
This PR is replaced by https://github.com/rhinstaller/anaconda/pull/2539 and others :D

PR #2527, opened 2020-05-18
Extend the class DNFPayload

poncovka:
Call the payload's method get_source_proxy get the DBus proxy of the source.

Use the payload's property source_type to simply the check the type
of the current source.

Ask the payload whether the source configuration is complete.

### DIFF
--- a/pyanaconda/payload/base.py
+++ b/pyanaconda/payload/base.py
@@ -56,6 +56,11 @@ class Payload(metaclass=ABCMeta):
         return None
 
     @property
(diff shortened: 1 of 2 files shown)


M4rtinK:
I think there is a typo in one of the commit messages:

```
Use the payload's property source_type to simply the check the type
of the current source.
```

I guess that should be "simplify the check" ?

poncovka:
Fixed the commit messages.

poncovka:
The tests passed.

PR #2530, opened 2020-05-18
subscription: Registration button sensitivity handling

M4rtinK:
The conditions for the Register/Unregister button being sensitive
are as follows:
- no registration/unregistration is currently in progress
- credentials for the currently selected authentication method are
  sufficient (we use the SECRET_TYPE_* constant to detect both locally
  set credentials as well as credentials stored in the module)

### DIFF
--- a/pyanaconda/ui/gui/spokes/subscription.py
+++ b/pyanaconda/ui/gui/spokes/subscription.py
@@ -41,6 +41,7 @@ from pyanaconda.ui.gui.spokes import NormalSpoke
 from pyanaconda.ui.gui.spokes.lib.subscription import fill_combobox
 from pyanaconda.ui.categories.system import SystemCategory
 from pyanaconda.ui.communication import hubQ
+from pyanaconda.ui.lib.subscription import username_password_sufficient, org_keys_sufficient
 
 from pyanaconda.anaconda_loggers import get_module_logger
 log = get_module_logger(__name__)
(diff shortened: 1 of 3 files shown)


pyanaconda/ui/gui/spokes/subscription.py
```diff
@@ -1008,4 +1009,24 @@ def _check_connectivity(self):
...
     def _update_register_button_state(self):
         """Update register button state."""
-        # TODO
+        if self._registration_controls_enabled:
+            # check if credentials are sufficient for registration
+            if self.authentication_method == AuthenticationMethod.USERNAME_PASSWORD:
+                username_set = self.subscription_request.account_username
+                password_type = self.subscription_request.account_password.type
+                password_set = password_type in (SECRET_TYPE_TEXT, SECRET_TYPE_HIDDEN)
+                if username_set and password_set:
```
poncovka, line 1018: I think I have already seen these checks.

pyanaconda/ui/gui/spokes/subscription.py
```diff
@@ -34,6 +34,7 @@
 from pyanaconda.modules.common.constants.services import SUBSCRIPTION, NETWORK
 from pyanaconda.modules.common.structures.subscription import SystemPurposeData, \
     SubscriptionRequest
+from pyanaconda.modules.common.structures.secret import SECRET_TYPE_TEXT, SECRET_TYPE_HIDDEN
```
poncovka, line 37: These should be imported from `pyanaconda.core.constants`.

M4rtinK: Fixed.

pyanaconda/ui/gui/spokes/subscription.py
M4rtinK: Well we have this checks:
https://github.com/rhinstaller/anaconda/blob/master/pyanaconda/ui/lib/subscription.py#L93
But that creates a new Subscription module DBus proxy on each invocation, which I don't think is something we want to here, as this check is invoked basically on every key press on the authentication. But I guess I could add an optional `subscription_request` parameter to the checking function, that would make it skip fetching the subscription request structure over DBus.

M4rtinK: OK, so with the helper function it really looks quite a bit nicer! :)

M4rtinK:
PR updated:
- fixed `SECRET_TYPE_*` imports
- reused helper function for checking if authentication data is sufficient
- handled the subscribed state explicitly, not just as a side effect of sufficient authentication data being a pre-condition for successful subscription

PR #2531, opened 2020-05-18
Fix call to super() with missing parameters

VladimirSlavik:
Not sure if this is needed as such, but pylint does not like it.

What could go wrong?

### DIFF
--- a/pyanaconda/ui/gui/spokes/subscription.py
+++ b/pyanaconda/ui/gui/spokes/subscription.py
@@ -78,7 +78,7 @@ class SubscriptionSpoke(NormalSpoke):
         return is_module_available(SUBSCRIPTION)
 
     def __init__(self, *args):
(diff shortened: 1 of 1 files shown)


pyanaconda/ui/gui/spokes/subscription.py
```diff
@@ -78,7 +78,7 @@ def should_run(cls, environment, data):
         return is_module_available(SUBSCRIPTION)
 
     def __init__(self, *args):
-        super().__init__()
+        super().__init__(self, *args)
```
poncovka, line 81: I don't think that `self` should be there.

VladimirSlavik: Fixed.

VladimirSlavik:
Jenkins, test this please.

PR #2536, opened 2020-05-18
Permit blacklisting modules from inherited parent product

jcpunk:
This patch will make it easy for CentOS and SL to inherit RHEL, but not subscription-manager.

### DIFF
--- a/data/anaconda.conf
+++ b/data/anaconda.conf
@@ -20,6 +20,8 @@ kickstart_modules =
      org.fedoraproject.Anaconda.Modules.Storage
      org.fedoraproject.Anaconda.Modules.Services
 
+# List of disabled Anaconda DBus modules.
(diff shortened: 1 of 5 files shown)


jkonecny12:
Jenkins, it's ok to test.

jkonecny12:
jenkins, test this please

jkonecny12:
So we discussed this in the team and we have decided that we would like to avoid having exclude lists. Mainly because we have plenty of options which could be excluded so it could grow our configuration file a lot.

Please change this PR to just copy the list and remove one item from the CentOS and SL. You can leave the test you have to avoid problems in the future.

jcpunk:
That is a bit sad, though I do understand the reasoning.

I'll close this PR out and build a new one with the alternate workflow.

PR #2548, opened 2020-05-19
Move lxml test dependency from pip to RPM

jkonecny12:
The lxml on pip is just a wrapper so it needs the package installed in the system during the installation. There is no real benefit in having it from pip.

Also it is not installed by default in RHEL-8 build root.


jkonecny12:
Created from a bad branch. Sorry for the noise.

PR #2552, opened 2020-05-20
Always clear kickstarted status for GUI time&date spoke (#1665428)

VladimirSlavik:
Clear the (internal) kickstarted status of Time and date GUI spoke on every call to `apply()`, because only human interaction can cause its execution.

This allows setting the default timezone America/New York in a single spoke visit when using kickstart with no timezone specified.

Resolves: rhbz#1665428

### DIFF
--- a/pyanaconda/ui/gui/spokes/datetime_spoke.py
+++ b/pyanaconda/ui/gui/spokes/datetime_spoke.py
@@ -587,22 +587,15 @@ class DatetimeSpoke(FirstbootSpokeMixIn, NormalSpoke):
         # Etc/XXXXXX timezone is selected
         region = self._get_active_region()
         city = self._get_active_city()
-        # nothing selected, just leave the spoke and
-        # return to hub without changing anything
+
+        # nothing selected, just leave the spoke and return to hub without changing anything
         if not region or not city:
             return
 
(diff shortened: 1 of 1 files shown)


VladimirSlavik:
Not sure what to do about the TODO for seen. It might be invalid by now?

pyanaconda/ui/gui/spokes/datetime_spoke.py
```diff
@@ -592,17 +592,11 @@ def apply(self):
...
-        self._timezone_module.SetTimezone(new_tz)
-
-        if old_tz != new_tz:
-            # new values, not from kickstart
-            # TODO: seen should be set from the module
-            self._kickstarted = False
-
+        self._timezone_module.SetTimezone(region + "/" + city)
         self._timezone_module.SetNTPEnabled(self._ntpSwitch.get_active())
+        # TODO: seen should be set from the module
```
poncovka, line 597: I guess you can remove this comment.

VladimirSlavik: Removed...

VladimirSlavik:
> If this is tested even with partial kickstart installation
Yes.

PR #2558, opened 2020-05-20
Run the set-up tasks of sources with signals enabled

poncovka:
The sources are connected to signals of their tasks, so make sure that these
signals are emitted as expected.


pyanaconda/modules/common/task/task.py
```diff
@@ -107,6 +107,29 @@ def _task_failed_with_info_callback(self, *exc_info):
         log.error("Thread %s has failed: %s", self._thread_name, formatted_info)
         super()._task_failed_callback()
 
+    def run_with_signals(self):
```
jkonecny12, line 110: I think it should be named `run_in_current_thread` or something similar. Signals are important but still an implementation detail.

poncovka:
Tests pass when run locally.

PR #2560, opened 2020-05-20
Small changes in the HDD source

poncovka:
The indicator of a mounted ISO is not really reliable outside the task, so
always try to unmount the HDD ISO. The task will log an error message, but
it will not fail if the mount point doesn't exist or it is not mounted.

There are two possible results of the task, so make sure the result
is readable from the logs.

### DIFF
--- a/pyanaconda/modules/payloads/source/harddrive/harddrive.py
+++ b/pyanaconda/modules/payloads/source/harddrive/harddrive.py
@@ -180,10 +180,10 @@ class HardDriveSourceModule(PayloadSourceBase, RPMSourceMixin):
         :return: list of tasks required for the source clean-up
         :rtype: [Task]
         """
-        tasks = []
-        if self.is_iso_mounted:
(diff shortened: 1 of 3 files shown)


poncovka:
The tests are passing.

pyanaconda/modules/payloads/source/harddrive/harddrive.py
```diff
@@ -180,10 +180,10 @@ def tear_down_with_tasks(self):
         :return: list of tasks required for the source clean-up
         :rtype: [Task]
         """
-        tasks = []
-        if self.is_iso_mounted:
```
jkonecny12, line 184: Please remove whole `is_iso_mounted`. This was it's only use.

poncovka: Fixed.

poncovka:
Updated.

PR #2561, opened 2020-05-21
Rhel 8 rebase root policy

rvykydal:
The first two commits are fixes applied also to master.
The third commit changes the defaults of root account options from these:
root_opts_original

to current rhel behaviour:
root_opts_rhel_policy

The fourth commit hides the options from the GUI:
root_opts_rhel_hidden


rvykydal:
@M4rtinK based our discussion I've updated the PR so that the we keep the behaviour of root account being locked if the root password is empty. So the second screenshot would look like:
root_opts_rhel_policy2

rvykydal:
jenkins, test this please

PR #2565, opened 2020-05-21
Collect package requirements of the Subscription module

poncovka:
We should use the requirements of the Subscription module if the module
is available.

### DIFF
--- a/pyanaconda/installation.py
+++ b/pyanaconda/installation.py
@@ -307,8 +307,12 @@ def _prepare_installation(payload, ksdata):
         # add package requirements from modules
         # - iterate over all modules we know have valid package requirements
         # - add any requirements found to the payload requirement tracking
-        modules_with_package_requirements = [SECURITY, NETWORK, TIMEZONE, STORAGE]
+        modules_with_package_requirements = [SECURITY, NETWORK, TIMEZONE, STORAGE, SUBSCRIPTION]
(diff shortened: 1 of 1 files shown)


pyanaconda/installation.py
```diff
@@ -307,8 +307,12 @@ def run_pre_install():
         # add package requirements from modules
         # - iterate over all modules we know have valid package requirements
         # - add any requirements found to the payload requirement tracking
-        modules_with_package_requirements = [SECURITY, NETWORK, TIMEZONE, STORAGE]
+        modules_with_package_requirements = [SECURITY, NETWORK, TIMEZONE, STORAGE, SUBSCRIPTION]
```
M4rtinK, line 310: The PR itself is totally fine but this is strange - why don't we just iterate over all currently available modules and gather all the requirements they provide ? I was kinda expecting it to be that way already - thus I only implemented the `collect_requirements()` method in the Subscription Module.

AFAIK we would do this just once per installation run so there should not be performance issues and modules that have no requirements would (at least I would expect) just return an empty list. That way we could also drop the module availability test as by definition only available modules would be iterated over.

poncovka: That's the plan.

M4rtinK: Sounds good. :)

PR #2570, opened 2020-05-21
Fix bad GetIsoPath calls of hard drive source module

jkonecny12:


jkonecny12:
**UPDATED:**
* Remove unused import.

PR #2573, opened 2020-05-21
Extend the support for CDN

poncovka:
The repo files source will represent a general support for local repo files.
Special cases like the CDN or the closest mirror should be handled in sources
of different types.

Add a new source of the type closest mirror. It will be a new default on Fedora.

Add support for sources of the type closest mirror and CDN.

Use the closest mirror as a default type for Fedora. We will use CDN for RHEL.


poncovka:
@VladimirSlavik, yes, the type REPO_FILES is not used at this moment, but we will eventually use it for additional repositories.

PR #2574, opened 2020-05-21
Fixes in the Source spoke in GUI

poncovka:
Use the default property that always returns True and move the code to the
method apply. The method apply will reset the payload if anything has changed.

Otherwise, the property could cause tearing down of the source without setting
it up again in the apply method. Then the installation started with unmounted
sources.

If the current source is SE/HMC, don't replace it with a new SE/HMC source.

### DIFF
--- a/pyanaconda/ui/gui/spokes/installation_source.py
+++ b/pyanaconda/ui/gui/spokes/installation_source.py
@@ -433,10 +433,17 @@ class SourceSpoke(NormalSpoke, GUISpokeInputCheckHandler, SourceSwitchHandler):
         self._treeinfo_repos_already_disabled = False
 
     def apply(self):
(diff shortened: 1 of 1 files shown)


pyanaconda/ui/gui/spokes/installation_source.py
```diff
@@ -433,10 +433,17 @@ def __init__(self, *args, **kwargs):
...
+        if source_changed or repo_changed or self._error:
+            payloadMgr.restart_thread(self.payload, checkmount=False)
+        else:
+            log.debug("Nothing has changed - skipping payload restart.")
+
         self.clear_info()
 
-    def _method_changed(self):
+    def _update_payload_source(self):
         """ Check to see if the install method has changed.
```
VladimirSlavik, line 447: The naming doesn't exactly fit together, but I suppose that is for another time.

PR #2576, opened 2020-05-21
Use platform-python instead of python3

VladimirSlavik:


M4rtinK:
I think there are some files missing that should be using platform-python:
`dracut/parse-kickstart`
`dracut/python-deps`

Also the following scripts run or can be run in the installation environment and should use `platform-python`:
`scripts/anaconda-cleanup`
`scripts/instperf`
`scripts/start-module`
`scripts/anaconda-read-journal`
`utils/handle-sshpw`
`data/command-stubs/list-harddrives-stub`

On the other hand, I don't think we want (or should) use `platform-python` in utility scripts and tests (we kept `/usr/bin/python3` in those scripts for RHEL 8.0-8.2).

AFAIK `platform-python` should only be used by system services and tools & is the only Python available in the installation environment. On the other hand on a normal installed RHEL8 system the user is expected to install regular (non platform) Python 3 and use that to run Python code as on any other distro (`/usr/bin/python3` is available & python3 is in `PATH`).

The idea is that by decoupling these two Python versions, they can be updated on a separate cadence as needed & multiple non-system Python versions being available. For comparison on RHEL<=7 there is a single Python version officially available that is used both by system tools and services and customers, so it's basically impossible to upgrade it as even if all the (Red Hat maintained) system tools and services were adjusted to the new Python version in sync, the version bump could still break customer code.

VladimirSlavik:
Thank you Martin, fixed... hopefully...

PR #2580, opened 2020-05-22
Set up LANG for tests

poncovka:
Set the environment variable LANG to en_US.UTF-8 in the testing environment.


poncovka:
The commit message was updated.

PR #2587, opened 2020-05-22
subscription: Use SetAll to set RHSM log level

M4rtinK:
The Set() RHSM DBus API method unfortunately does clash with default
DBus Set() method, resulting in a type conflict. So at least until
this is fixed, let's use the SetAll() method that is unaffected by this.


poncovka:
This is deprecated by https://github.com/rhinstaller/anaconda/pull/2588.

PR #2588, opened 2020-05-22
Create a proxy of the RHSM Config object for a specific interface

poncovka:
The DBus object Config re-defines methods of the standard interface
org.freedesktop.DBus.Properties, so create a proxy only with the
Config's interface.


poncovka:
Fixed tests.

PR #2589, opened 2020-05-22
Use a specific kickstart version in handle-sshpw

poncovka:
Import a kickstart version from pyanaconda and use it to create the handler.
Otherwise, pykickstart will use the DEVEL version that can be different
from the expected one.


jkonecny12:
I would suggest to avoid having the "default" here. It will make this confusion in the future releases.

PR #2590, opened 2020-05-22
Change connection flags for the RHSM private bus

poncovka:
The RHSM private bus connection requires only the AUTHENTICATION_CLIENT flag.


poncovka:
Fixed tests.

PR #2595, opened 2020-05-23
subscription: Set RHSM configuration before registration attempt

M4rtinK:
We should set configuration of the RHSM service before we attempt
to register (or unregister) or else custom configuration options
will be ignored (server hostname, rhsm baseurl, HTTP proxy) during
the attempt.

### DIFF
--- a/tests/nosetests/pyanaconda_tests/subscription_helpers_test.py
+++ b/tests/nosetests/pyanaconda_tests/subscription_helpers_test.py
@@ -239,6 +239,7 @@ class AsynchronousRegistrationTestCase(unittest.TestCase):
         # we were successful, so no error callback calls
         error_callback.assert_not_called()
         # we should have requested the appropriate tasks
+        subscription_proxy.SetRHSMConfigWithTask.assert_called_once()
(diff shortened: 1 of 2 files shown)


tests/nosetests/pyanaconda_tests/subscription_helpers_test.py
```diff
@@ -451,9 +461,13 @@ def register_password_missing_test(self, get_proxy, thread_mgr_wait, run_task):
...
-        # int this case we fails before requesting any task and we should not
-        # attempt to run any
-        run_task.assert_not_called()
+        # in this case we fails before requesting any other task than
+        # the config one
+        subscription_proxy.SetRHSMConfigWithTask.assert_called_once()
+        run_task.assert_called()
+        # setting CDN as installation source does not make sense
+        # when we were not able to attach a subscription
+        set_cdn.assert_not_called()
```
poncovka, line 470: This variable is not defined.

M4rtinK: The variable was introduced by #2571, which has been just merged, so this should be good now.

PR #2604, opened 2020-05-25
subscription: Fix authentication method switching in GUI

M4rtinK:
Make sure the authetication method revealers are activated
appropriatelly when the authetication radio buttons are
clicked.

Related: rhbz#1691319


M4rtinK:
jenkins, test this please

PR #2608, opened 2020-05-26
Fix adding repository from unified URL source

jkonecny12:
The problem is that we added ``os.path.normpath()`` to clean-up relative paths with NFS mounts (`../../`). However, this also resulted in change of `http://` -> `http:/` issue which resulted in crash of DNF `repo.load()` method.

Add new helper function to split protocol from a url so we can use normpath to just url without the protocol.

### DIFF
--- a/pyanaconda/core/payload.py
+++ b/pyanaconda/core/payload.py
@@ -64,6 +64,34 @@ def create_nfs_url(host, path, options=None):
     return "nfs:{server}:{path}".format(server=host, path=path)
 
 
+def split_protocol(url):
+    """Split protocol from url
+
+    The function will look for ://.
+
+    - If found more than once in the url then raising an error.
+    - If found exactly once it will return tuple with [protocol, rest_of_url].
+    - If an empty string is given it will return tuple with empty strings ("", "").
+
+    :param str url: base url we want to split protocol from
+    :return: tuple of (protocol, rest of url)
+    :raise: ValueError if url is invalid
+    """
+    ret = url.split("://")
+
+    if len(ret) > 2:
+        raise ValueError("Invalid url to split protocol '{}'".format(url))
+
+    if len(ret) == 2:
+        # return back part removed when splitting
+        return (ret[0] + "://", ret[1])
+
+    if len(ret) == 1:
+        return ("", ret[0])
+
+    return ("", "")
+
+
 class ProxyStringError(Exception):
     pass
 
--- a/pyanaconda/payload/install_tree_metadata.py
+++ b/pyanaconda/payload/install_tree_metadata.py
@@ -23,6 +23,7 @@ import os
 
 from productmd.treeinfo import TreeInfo
 from pyanaconda.core import util, constants
+from pyanaconda.core.payload import split_protocol
 
 from pyanaconda.anaconda_loggers import get_packaging_logger
 log = get_packaging_logger()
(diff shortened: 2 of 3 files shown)


pyanaconda/core/payload.py
```diff
@@ -64,6 +64,35 @@ def create_nfs_url(host, path, options=None):
...
+    :raise: ValueError if url is invalid
+    """
+    ret = url.split("://")
+
+    if len(ret) > 2:
+        raise ValueError("Invalid url to split protocol '{}'".format(url))
+
+    if len(ret) == 2:
+        # return back part removed when splitting
+        ret[0] = ret[0] + "://"
```
poncovka, line 87: Maybe, the code could be easier to follow, if we call here `return ret[0] + "://", ret[1]`.

pyanaconda/payload/install_tree_metadata.py
```diff
@@ -236,7 +238,10 @@ def path(self):
         if self.relative_path == ".":
             return self._root_path
         else:
-            return os.path.normpath(os.path.join(self._root_path, self.relative_path))
+            [protocol, url] = split_protocol(os.path.join(self._root_path, self.relative_path))
```
poncovka, line 241: This should be a tuple.

tests/nosetests/pyanaconda_tests/payload_test.py
```diff
@@ -645,3 +645,21 @@ def nfs_combine_test(self):
...
+                         ["http://", "abc/cde"])
+        self.assertEqual(util.split_protocol("https://yay/yay"),
+                         ["https://", "yay/yay"])
+        self.assertEqual(util.split_protocol("ftp://ups/spu"),
+                         ["ftp://", "ups/spu"])
+        self.assertEqual(util.split_protocol("nfs:ups/spu:/abc:opts"),
+                         ["", "nfs:ups/spu:/abc:opts"])
+        self.assertEqual(util.split_protocol("http:/typo/test"),
+                         ["", "http:/typo/test"])
+        self.assertEqual(util.split_protocol(""), ["", ""])
```
poncovka, line 662: The result should be always a tuple.

pyanaconda/core/payload.py
poncovka: Also, you will not return a tuple in this case.

pyanaconda/core/payload.py
```diff
@@ -64,6 +64,35 @@ def create_nfs_url(host, path, options=None):
...
 
 
+def split_protocol(url):
+    """Split protocol from url
+
+    The function will look for ://.
+
+    - If found more than once in the url then raising an error.
+    - If found exactly once it will return tuple with [protocol, rest_of_url].
+    - If an empty string is given it will return tuple with empty strings ["", ""].
```
poncovka, line 74: This is not a tuple.

jkonecny12:
**UPDATED:**
* Implemented suggestion from @poncovka. Thanks!

pyanaconda/payload/install_tree_metadata.py
```diff
@@ -236,7 +238,10 @@ def path(self):
         if self.relative_path == ".":
             return self._root_path
         else:
-            return os.path.normpath(os.path.join(self._root_path, self.relative_path))
+            protocol, url = split_protocol(os.path.join(self._root_path, self.relative_path))
+            # resolve problems with relative path especially useful for NFS (root/path/../new_path)
+            url = os.path.normpath(url)
```
VladimirSlavik, line 243: See #2610 for details.

jkonecny12: Answered there.

PR #2609, opened 2020-05-26
subscription: Make sure CDN repos are loaded and usable

M4rtinK:
We need to do two things to make the CDN provided repositories
available for installation:

1) we need to make sure the CDN provided repos loaded from the
   RHSM generated redhat.repo file

Normally no repo file are loaded for kickstart installations to
prevent the closest mirror repo files from interfering with
automated installations.

But if the CDN is set as the installation source and we are
subscribed, we want to override this and load the RHSM generated
repo file.

2) we need to make sure a base repo is available if CDN is used
   as the installation source

On Fedora for an installation from closest mirror we check
the available repos for a whitelist of well known repo names
and select one of them as base repo if it matches this whitelist.

In comparison, the repo names on the Red Hat CDN is no concept
of such stable well known repo names and the names are effectively
considered to be an implementation detail. Still, any enabled
repo in the redhat.repo file should be usable as a base repo.

So for the CDN, we just pick one of the enabled CDN repos to prevent
the payload code from falling over due to no base repo being set.

Based on commit eafef5ab85faa52b2efea3ad6526e12426551692.

Related: rhbz#1691319

### DIFF
--- a/pyanaconda/payload/dnf/payload.py
+++ b/pyanaconda/payload/dnf/payload.py
@@ -52,7 +52,7 @@ from pyanaconda.anaconda_loggers import get_dnf_logger, get_packaging_logger
 from pyanaconda.core import constants, util
 from pyanaconda.core.configuration.anaconda import conf
 from pyanaconda.core.constants import INSTALL_TREE, ISO_DIR, PAYLOAD_TYPE_DNF, \
-    DNF_DEFAULT_SOURCE_TYPE, SOURCE_TYPE_HMC, SOURCE_TYPE_URL, \
+    DNF_DEFAULT_SOURCE_TYPE, SOURCE_TYPE_HMC, SOURCE_TYPE_URL, SOURCE_TYPE_CDN, \
     URL_TYPE_BASEURL, URL_TYPE_MIRRORLIST, URL_TYPE_METALINK, SOURCE_REPO_FILE_TYPES
 from pyanaconda.core.i18n import N_, _
 from pyanaconda.core.payload import ProxyString, ProxyStringError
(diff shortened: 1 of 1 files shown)


pyanaconda/payload/dnf/payload.py
```diff
@@ -810,10 +810,21 @@ def base_repo(self):
...
+        subscription_proxy = SUBSCRIPTION.get_proxy()
         with self._repos_lock:
-            for repo in self._base.repos.iter_enabled():
-                if repo.id in repo_names:
-                    return repo.id
+            if self.source_type == SOURCE_TYPE_CDN:
+                if subscription_proxy.IsSubscriptionAttached:
+                    # If CDN is used as the installation source and we have a subscription attached,
+                    # any of the enabled repos should be fine as the base repo.
+                    # If CDN is used but subscription has not be attached, there will not be
```
jkonecny12, line 819:
```suggestion
                    # If CDN is used but subscription has not been attached, there will not be
```

pyanaconda/payload/dnf/payload.py
```diff
@@ -1521,8 +1532,15 @@ def update_base_repo(self, fallback=True, checkmount=True):
...
         if source_type in SOURCE_REPO_FILE_TYPES:
 
-            # If this is a kickstart install, just return now
-            if flags.automatedInstall:
+            # If this is a kickstart install, just return now as we normally do not
+            # want to read the on media repo files in such a case. On the other hand,
+            # the local repo files are a valid use case if the system is subscribed
+            # and the CDN is selected as the installation source.
+            subscription_proxy = SUBSCRIPTION.get_proxy()
+            cdn_source = self.source_type == SOURCE_TYPE_CDN
```
jkonecny12, line 1540: Better name would be `is_cdn_source` here.

pyanaconda/payload/dnf/payload.py
```diff
@@ -1521,8 +1532,15 @@ def update_base_repo(self, fallback=True, checkmount=True):
...
+            subscription_attached = subscription_proxy.IsSubscriptionAttached
+            load_cdn_repos = cdn_source and subscription_attached
```
jkonecny12, line 1542: I think you can inline the `subscription_attached` variable here. It does not help here.

pyanaconda/payload/dnf/payload.py
M4rtinK: Fixed. :)

pyanaconda/payload/dnf/payload.py
M4rtinK: Changed.

pyanaconda/payload/dnf/payload.py
M4rtinK: Inlined.

PR #2610, opened 2020-05-26
Fix adding repository from unified URL source

jkonecny12:
The problem is that we added os.path.normpath() to clean-up relative paths with NFS mounts (../../). However, this also resulted in change of http:// -> http:/ issue which resulted in crash of DNF repo.load() method.

Add new helper function to split protocol from a url so we can use normpath to just url without the protocol.

Backport PR: https://github.com/rhinstaller/anaconda/pull/2608

### DIFF
--- a/pyanaconda/payload/install_tree_metadata.py
+++ b/pyanaconda/payload/install_tree_metadata.py
@@ -23,6 +23,7 @@ import os
 
 from productmd.treeinfo import TreeInfo
 from pyanaconda.core import util, constants
+from pyanaconda.core.payload import split_protocol
 
 from pyanaconda.anaconda_loggers import get_packaging_logger
 log = get_packaging_logger()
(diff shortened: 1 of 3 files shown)


pyanaconda/payload/install_tree_metadata.py
```diff
@@ -236,7 +238,10 @@ def path(self):
         if self.relative_path == ".":
             return self._root_path
         else:
-            return os.path.normpath(os.path.join(self._root_path, self.relative_path))
+            protocol, url = split_protocol(os.path.join(self._root_path, self.relative_path))
+            # resolve problems with relative path especially useful for NFS (root/path/../new_path)
+            url = os.path.normpath(url)
```
VladimirSlavik, line 243: Actually, do we want to do this if there is something in the protocol?

jkonecny12: We don't but there is no need to avoid that. This is just simplification instead of adding unnecessary if.

VladimirSlavik: OK, I think "URLs won't contain double slashes" is a reasonable assumption, even if not bulletproof.

jkonecny12: I'm merging this now. If we will find better solution in the future we can update this.

PR #2603, opened 2020-05-25
Fix RHEL tests as much as possible

VladimirSlavik:
This should fix our tests to work on the RHEL branch after rebase. There are lots of commits because the problems are rather diverse...

TODOs:

* [ ] ~~Solve crashy `blivet-gui` imports~~
* [x] Silence or appease `cppcheck`

PR #2611, opened 2020-05-26
Test fixes from rhel8

VladimirSlavik:
Backport the usable bits from #2603.


tests/nosetests/pyanaconda_tests/dbus_test.py
```diff
@@ -72,8 +73,8 @@ def default_bus_test(self, getter):
         """Test the default bus."""
         message_bus = DefaultMessageBus()
 
-        with patch.dict("os.environ") as environment:
```
poncovka, line 75: I think that we don't have to fix Python 3.8 stuff in upstream.

PR #2613, opened 2020-05-27
Fix changelog to keep rhel history

rvykydal:
Do not overwrite rhel history with history of upstream we rebase to.
Also include commits added after rebase.


rvykydal:
I've just added rhbz# reference.

PR #2614, opened 2020-05-27
Fixes for the CDROM source

poncovka:
Use the current source to set up the GUI elements in the refresh method of
the Source spoke, but don't change it. The source should be changed only in
the apply method.

Set up the spoke for all supported types of sources and raise an exception
if the current type is unsupported.

When we restart the payload thread for the first time, we should change
the default source to CDROM if there is a valid install media.

Anaconda uses the argument checkmount to control this behavior, but we
should try to find a better way how to handle this case.

### DIFF
--- a/pyanaconda/ui/gui/spokes/installation_source.py
+++ b/pyanaconda/ui/gui/spokes/installation_source.py
@@ -55,7 +55,6 @@ from pyanaconda.modules.common.constants.services import NETWORK, STORAGE
 from pyanaconda.modules.common.constants.objects import DEVICE_TREE
 from pyanaconda.modules.common.structures.storage import DeviceData
 from pyanaconda.core.storage import device_matches
-from pyanaconda.ui.lib.payload import set_source, tear_down_sources, create_source
 
 from pyanaconda.anaconda_loggers import get_module_logger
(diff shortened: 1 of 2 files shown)


pyanaconda/ui/gui/spokes/installation_source.py
```diff
@@ -973,20 +972,17 @@ def refresh(self):
...
-            # guess.  Go with autodetected media if that was provided, and then
-            # fall back to closest mirror.
+        elif source_type == SOURCE_TYPE_CDROM:
+            # Go with autodetected media if that was provided,
+            # otherwise fall back to the closest mirror.
             if not self._autodetect_button.get_no_show_all():
                 self._autodetect_button.set_active(True)
-                source_type = SOURCE_TYPE_CDROM
             else:
                 self._network_button.set_active(True)
```
rvykydal, line 981: Should the network_button be active for SOURCE_TYPE_CDROM ?

poncovka: The option for CDROM ( `autodetect_button`) doesn't have to be visible if we are not able to find a valid installation media. In this case, we have to fall back to another option - the closest mirror (`network_button`).

PR #2615, opened 2020-05-28
Add back NFS ISO support (ups)

jkonecny12:
We missed support for ISO in the NFS payload source.

This will implement it back in a reasonable manner. It has plenty of duplicate code with `HardDriveSourceModule` but merging those would mean changing almost all the hard drive tests and nfs tests also designing completely new solution (probably mixin class). This is good enough solution for now and I'll try to make next PR with better solution soon to get it to RHEL-8 when we have time.

### DIFF
--- a/pyanaconda/modules/payloads/source/nfs/nfs.py
+++ b/pyanaconda/modules/payloads/source/nfs/nfs.py
@@ -17,6 +17,8 @@
 # License and may only be used or replicated with the express permission of
 # Red Hat, Inc.
 #
+import os.path
+
 from pyanaconda.core.i18n import _
 from pyanaconda.core.payload import create_nfs_url, parse_nfs_url
 from pyanaconda.core.signal import Signal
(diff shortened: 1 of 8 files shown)


pyanaconda/modules/payloads/source/nfs/nfs.py
```diff
@@ -111,19 +125,39 @@ def set_url(self, url):
...
+        tasks = [
+            TearDownMountTask(self._iso_mount),
+            TearDownMountTask(self._device_mount),
+        ]
+        return tasks
+
+    def _handle_setup_task_result(self, task):
+        result = task.get_result()
+        self._install_tree_path = result
+        log.debug("Hard drive install tree path is set to '%s'", self._install_tree_path)
```
poncovka, line 163: NFS install tree path ....

jkonecny12:
**UPDATED:**
* Implemented notes from @poncovka . Thanks!

PR #2617, opened 2020-05-28
subscription: Replace helpFile property with help_id

M4rtinK:
Previously we used to have one help content file per screen,
which no longer works for the new help content, which is using
a single docbook file with anchors.

So replace the helpFile property with a help_id property, which
holds a unique string id corresponding to one screen needing help.

The help id can then be mapped into an anchor in the main docbook file,
or even something else if needed.

Based on 56777977c9bfbd97b97ef0d72e59d0702f7b725c.

Related: rhbz#1593723
Related: rhbz#1691319

### DIFF
--- a/pyanaconda/ui/gui/__init__.py
+++ b/pyanaconda/ui/gui/__init__.py
@@ -45,7 +45,7 @@ from pyanaconda.ui.gui.utils import gtk_call_once, unbusyCursor
 from pyanaconda.core.async_utils import async_action_wait
 from pyanaconda.ui.gui.utils import watch_children, unwatch_children
 from pyanaconda.ui.gui.helpers import autoinstall_stopped
-from pyanaconda.ui.lib.help import start_yelp, get_help_path
+from pyanaconda.ui.lib.help import start_yelp
 import os.path
 
 from pyanaconda.anaconda_loggers import get_module_logger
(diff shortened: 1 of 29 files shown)


pyanaconda/ui/tui/spokes/__init__.py
```diff
@@ -101,7 +101,7 @@ def input(self, args, key):
         # TRANSLATORS: 'h' to help
         if key.lower() == Prompt.HELP:
             if self.has_help:
-                help_path = get_help_path(self.helpFile, True)
+                help_path = get_placeholder_path(plain_text=True)
```
poncovka, line 104: Please, just call get_help_path with None for now.

pyanaconda/ui/tui/hubs/__init__.py
```diff
@@ -139,7 +139,7 @@ def input(self, args, key):
             # TRANSLATORS: 'h' to help
             elif key == Prompt.HELP:
                 if self.has_help:
-                    help_path = get_help_path(self.helpFile, True)
+                    help_path = get_placeholder_path(plain_text=True)
```
poncovka, line 142: Same here.

pyanaconda/ui/gui/__init__.py
```diff
@@ -1028,7 +1029,7 @@ def _on_continue_clicked(self, window, user_data=None):
     def _on_help_clicked(self, window, obj):
         # the help button has been clicked, start the yelp viewer with
         # content for the current screen
-        start_yelp(get_help_path(obj.helpFile))
+        start_yelp(obj.help_id)
```
poncovka, line 1032: Same here.

M4rtinK:
In the end the help ids were handled in #2567.

PR #2619, opened 2020-05-28
Add back NFS ISO support (ups)

jkonecny12:
We missed support for ISO in the NFS payload source.

This will implement it back in a reasonable manner. It has plenty of duplicate code with HardDriveSourceModule but merging those would mean changing almost all the hard drive tests and nfs tests also designing completely new solution (probably mixin class). This is good enough solution for now and I'll try to make next PR with better solution soon to get it to RHEL-8 when we have time.

Backport of https://github.com/rhinstaller/anaconda/pull/2615

### DIFF
--- a/pyanaconda/modules/payloads/source/harddrive/initialization.py
+++ b/pyanaconda/modules/payloads/source/harddrive/initialization.py
@@ -19,11 +19,10 @@ import os.path
 
 from collections import namedtuple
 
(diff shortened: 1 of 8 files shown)


jkonecny12:
Feel free to merge this if the ACKs will be here to avoid divergence between `rhel-8` and `master`.

PR #2620, opened 2020-05-28
subscription: Make sure CDN repos are loaded and usable

M4rtinK:
We need to do two things to make the CDN provided repositories
available for installation:

1) we need to make sure the CDN provided repos loaded from the
   RHSM generated redhat.repo file

Normally no repo file are loaded for kickstart installations to
prevent the closest mirror repo files from interfering with
automated installations.

But if the CDN is set as the installation source and we are
subscribed, we want to override this and load the RHSM generated
repo file.

2) we need to make sure a base repo is available if CDN is used
   as the installation source

On Fedora for an installation from closest mirror we check
the available repos for a whitelist of well known repo names
and select one of them as base repo if it matches this whitelist.

In comparison, the repo names on the Red Hat CDN is no concept
of such stable well known repo names and the names are effectively
considered to be an implementation detail. Still, any enabled
repo in the redhat.repo file should be usable as a base repo.

So for the CDN, we just pick one of the enabled CDN repos to prevent
the payload code from falling over due to no base repo being set.

Based on commit eafef5ab85faa52b2efea3ad6526e12426551692.

Related: rhbz#1691319

Master version of #2609.

### DIFF
--- a/pyanaconda/payload/dnf/payload.py
+++ b/pyanaconda/payload/dnf/payload.py
@@ -54,7 +54,8 @@ from pyanaconda.core import constants, util
 from pyanaconda.core.configuration.anaconda import conf
 from pyanaconda.core.constants import INSTALL_TREE, ISO_DIR, PAYLOAD_TYPE_DNF, \
     DNF_DEFAULT_SOURCE_TYPE, SOURCE_TYPE_HMC, SOURCE_TYPE_URL, SOURCE_TYPE_CDROM, \
-    URL_TYPE_BASEURL, URL_TYPE_MIRRORLIST, URL_TYPE_METALINK, SOURCE_REPO_FILE_TYPES
+    URL_TYPE_BASEURL, URL_TYPE_MIRRORLIST, URL_TYPE_METALINK, SOURCE_REPO_FILE_TYPES, \
+    SOURCE_TYPE_CDN
 from pyanaconda.core.i18n import N_, _
 from pyanaconda.core.payload import ProxyString, ProxyStringError
 from pyanaconda.core.regexes import VERSION_DIGITS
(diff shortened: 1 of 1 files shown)


pyanaconda/payload/dnf/payload.py
```diff
@@ -822,10 +823,23 @@ def base_repo(self):
         """Get the identifier of the current base repo or None."""
         # is any locking needed here?
         repo_names = [constants.BASE_REPO_NAME] + constants.DEFAULT_REPOS
+        subscription_proxy = SUBSCRIPTION.get_proxy()
```
poncovka, line 826: This will fail on Fedora. You should reorganize the code and make sure that we can access the Subscription module.

pyanaconda/payload/dnf/payload.py
```diff
@@ -1540,8 +1554,14 @@ def update_base_repo(self, fallback=True, checkmount=True):
         # We need to check this again separately in case REPO_FILES were set above.
         if source_type in SOURCE_REPO_FILE_TYPES:
 
-            # If this is a kickstart install, just return now
-            if flags.automatedInstall:
+            # If this is a kickstart install, just return now as we normally do not
+            # want to read the on media repo files in such a case. On the other hand,
+            # the local repo files are a valid use case if the system is subscribed
+            # and the CDN is selected as the installation source.
+            subscription_proxy = SUBSCRIPTION.get_proxy()
```
poncovka, line 1561: Same here.

pyanaconda/payload/dnf/payload.py
M4rtinK: Good, point! Changed! :)

pyanaconda/payload/dnf/payload.py
M4rtinK: Changed here to move the check into the condition that checks if CDN is used as the installation source, so we can log an error message ig CDN source is used yet Subscription module is not available.

pyanaconda/payload/dnf/payload.py
```diff
@@ -1540,8 +1558,20 @@ def update_base_repo(self, fallback=True, checkmount=True):
         # We need to check this again separately in case REPO_FILES were set above.
         if source_type in SOURCE_REPO_FILE_TYPES:
 
-            # If this is a kickstart install, just return now
-            if flags.automatedInstall:
+            # If this is a kickstart install, just return now as we normally do not
+            # want to read the on media repo files in such a case. On the other hand,
+            # the local repo files are a valid use case if the system is subscribed
+            # and the CDN is selected as the installation source.
+            if is_module_available(SUBSCRIPTION):
```
poncovka, line 1565: I think that you should check the type first as well.

pyanaconda/payload/dnf/payload.py
poncovka: Please, fix also the RHEL 8 branch. It will fail on Cent OS as well.

pyanaconda/payload/dnf/payload.py
M4rtinK: OK, did that.

pyanaconda/payload/dnf/payload.py
M4rtinK: Good point, I'll do that based on the final form of this PR.

M4rtinK:
Manual testing of  the final version of this PR indicates no issue on both interactive and kickstart installation, so I think we can merge this.

PR #2621, opened 2020-05-29
Update requirements for fips - update-crypto-policies script

rvykydal:
Resolves: rhbz#1841471

### DIFF
--- a/anaconda.spec.in
+++ b/anaconda.spec.in
@@ -126,6 +126,7 @@ Requires: python3-ordered-set >= 2.0.0
 
 # Required by the systemd service anaconda-fips.
 Requires: crypto-policies
+Requires: /usr/bin/update-crypto-policies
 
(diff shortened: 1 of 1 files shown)


t184256:
Please specify both crypto-policies and crypto-policies-scripts, I find it more explicit and it depends on both.

t8m:
> Please specify both crypto-policies and crypto-policies-scripts, I find it more explicit and it depends on both.

Except if for any catastrophical reason we are forced to revert the package split anaconda will have to be updated again.

t8m:
So the best is:
Requires: crypto-policies 
Requires: /usr/bin/update-crypto-policies

t184256:
no objections

rvykydal:
> So the best is:
> Requires: crypto-policies
> Requires: /usr/bin/update-crypto-policies

PR updated

PR #2623, opened 2020-05-29
request for comments: Implement a faster way to retest updates.img

t184256:
Add a script that checkpoints a QEMU VM before it fetches updates.img,
so that you can re-test different ones faster.

Trying it out:
```
./scripts/reanaconda prime --sensible --tree http://ftp.fi.muni.cz/pub/linux/fedora/linux/releases/32/Everything/x86_64/os
./scripts/reanaconda updates updates.img
./scripts/reanaconda cleanup
```

I'm open for implementing additional features, redesigning it as a single command, auto-reloading as you save the changes --- whatever saves you time in the future. I have little experience with anaconda though, so I'm interested in what the more active developers think about it.

### DIFF
--- a/scripts/reanaconda
+++ b/scripts/reanaconda
@@ -1,0 +1,253 @@
+#!/usr/bin/python3
+
+"""
+reanaconda, a script to make testing different updates.img in QEMU faster.
+This is achieved by separating the installation into two stages,
+with a checkpoint before downloading updates.img.
+
+First use `prime` to perform all steps preceeding updates.img downloading.
+Then do `updates` to resumes/restart with an updated updates.img.
+
+Requires Python 3.6+ (3.7+?), docopt, QEMU, netcat.
+
+Usage:
+  ./reanaconda prime [--tree <url>] [--append <extra_cmdline>] [--sensible]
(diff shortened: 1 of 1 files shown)


jkonecny12:
Hello @t184256 and thanks for your contribution!

This does not really reflect Anaconda team way of development but I see it pretty useful for folks from the community. So we are pretty happy to merge this, however, please move this PR to [Devel tools repository](https://github.com/rhinstaller/devel-tools/). 
One reason for that is that Anaconda repo has some restrictions about branching and merging so it is easier to make the changes in a separate repository. Another benefit is that scripts in the Anaconda repository are tend to be forgotten, they are created with an interesting idea but no-one new know it is there. 

I'm closing this PR and looking forward to see it on devel-tools repo.

t184256:
@jkonecny12, sure, I'll re-file it. I've filed it here because I happened to discover this place first.

> This does not really reflect Anaconda team way of development

That was kinda expected, given I wrote it with just a vague idea of what your workflow is. If you decide the idea is worth integrating into what you really use, I'd like to do that. I just like writing tools, I guess.

PR #2628, opened 2020-06-01
Use sshd_config drop in directory to allow root login

Jakuje:
Last year, anaconda introduced a checkbox override permission to root login through ssh (#2042). It was implemented in suboptimal way, because at that time, OpenSSH's sshd_config parser did not have include capability so we needed to take it through environment file and command-line arguments, which is ugly.

This patch is changing the configuration file generated by anaconda to new drop-in directory for sshd to pick up configuration files from, which will make it easier to maintain on both sides and more discoverable for users.

Let me know if I missed some case. I was not able to configure anaconda locally.

Resolves rhbz#1832900.


Jakuje:
Note, that this can go to Fedora 32 and 33 but not in older releases as they miss the include functionality.

pyanaconda/modules/users/installation.py
```diff
@@ -195,7 +195,7 @@ def run(self):
                 f.write(
                     '# This file has been generated by the Anaconda Installer.\n'
                     '# Allow root to log in using ssh. Remove this file to opt-out.\n'
-                    'PERMITROOTLOGIN="-oPermitRootLogin=yes"\n'
+                    'PermitRootLogin yes"\n'
```
VladimirSlavik, line 198: I think you have a stray double quote " there?

VladimirSlavik:
Besides Anaconda writing the file, there is this:
```
[root@yavstest /]# grep -rnw '/usr/' -e 'PERMITROOTLOGIN'
/usr/lib/systemd/system/sshd@.service:11:ExecStart=-/usr/sbin/sshd -i $OPTIONS $CRYPTO_POLICY $PERMITROOTLOGIN
/usr/lib/systemd/system/sshd.service:12:ExecStart=/usr/sbin/sshd -D $OPTIONS $CRYPTO_POLICY $PERMITROOTLOGIN
```
@Jakuje do you know anything about this part?

jkonecny12:
> Note, that this can go to Fedora 32 and 33 but not in older releases as they miss the include functionality.

We can't update already released Fedora so it will go to Fedora 33 only.

jkonecny12:
Jenkins, it's ok to test.

jkonecny12:
@Jakuje looks good except what was pointed by @VladimirSlavik already.

Thanks a lot for implementing the drop dir for ssh. Great job to improve the current situation! Also thanks for this PR!

pyanaconda/modules/users/installation.py
Jakuje: Right. Will fix in next push.

Jakuje:
> Besides Anaconda writing the file, there is this:
> 
> ```
> [root@yavstest /]# grep -rnw '/usr/' -e 'PERMITROOTLOGIN'
> /usr/lib/systemd/system/sshd@.service:11:ExecStart=-/usr/sbin/sshd -i $OPTIONS $CRYPTO_POLICY $PERMITROOTLOGIN
> /usr/lib/systemd/system/sshd.service:12:ExecStart=/usr/sbin/sshd -D $OPTIONS $CRYPTO_POLICY $PERMITROOTLOGIN
> ```
> 
> I was not able to find a reference to the variable anywhere else on the system. @Jakuje do you know anything about this part?

This is part of openssh package using your generated file. I am a bit hesitant to remove this part right after this update will land, because it could break users who configured root login this "old" way through anaconda. But it might make sense to handle this conversion in openssh scriptlet as it is pretty straight-forward.

VladimirSlavik:
Thank you, that makes sense. I was just wondering if the service file is something we (anaconda) should care about too.

Sorry to keep dragging this on, but the test needed fixing the stray " too, then it should pass. Currently tests are broken though :(

Jakuje:
> Thank you, that makes sense. I was just wondering if the service file is something we (anaconda) should care about too.
> 
> Sorry to keep dragging this on, but the test needed fixing the stray " too, then it should pass. Currently tests are broken though :(

My bad. I would run the tests locally, but it looks like Fedora 31 is missing some dependencies if I saw right. Fixed now in the tests too.

jkonecny12:
jenkins, test this please

PR #183, opened 2015-07-07
Test fixes

dashea:
That is, fixes to the tests

PR #2630, opened 2020-06-02
Fix wireless network configuration on get_ssid() returning None (#183…

rvykydal:
…8357)

Having connecntion found ensures the ssid would be found as well.


rvykydal:
jenkins, test this please

PR #2631, opened 2020-06-02
Set up LD_PRELOAD for the Payloads module

poncovka:
The environment variable LD_PRELOAD needs to be set up during the start of
the Payloads module, but it needs to be dropped after that.

After the modularization, only the Payloads module should require LD_PRELOAD.


sharkcz:
I did a quick test with Fedora-Rawhide-20200601.n.1 and an updates.img provided by @M4rtinK and GUI is up and working.

poncovka:
Jenkins, test this please.

sharkcz:
I think reference to https://bugzilla.redhat.com/show_bug.cgi?id=1722181 should be added.

poncovka:
The commit messages are updated.

PR #2632, opened 2020-06-03
Catch Blivet's exceptions when we reset a device (#1843278)

poncovka:
Make sure we catch the Blivet's exceptions and raise the DBus errors instead
when we call the DBus method ResetDevice of the Storage module.

Resolves: rhbz#1843278


poncovka:
Jenkins, test this please.

PR #2633, opened 2020-06-03
localization: do not crash on failed parsing of X layout (#1836047)

rvykydal:
Also adds unit test for the case and more localed wrapper testing
(67% -> 98% coverage of localed.py)

### DIFF
--- a/pyanaconda/modules/localization/localed.py
+++ b/pyanaconda/modules/localization/localed.py
@@ -18,7 +18,8 @@
 from pyanaconda.core.dbus import SystemBus
 from pyanaconda.modules.common.constants.services import LOCALED
 from pyanaconda.core.configuration.anaconda import conf
-from pyanaconda.keyboard import join_layout_variant, parse_layout_variant
+from pyanaconda.keyboard import join_layout_variant, parse_layout_variant, \
+    InvalidLayoutVariantSpec
 from pyanaconda.core.constants import DEFAULT_KEYBOARD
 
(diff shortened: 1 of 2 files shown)


pyanaconda/modules/localization/localed.py
```diff
@@ -160,12 +161,21 @@ def set_layouts(self, layouts_variants, options=None, convert=False):
...
+                log.debug("Parsing of %s failed: %s", layout_variant, e)
+                parsing_failed = True
+                continue
             layouts.append(layout)
             variants.append(variant)
 
+        if not layouts and parsing_failed:
+            return
+
         layouts_str = ",".join(layouts)
```
jkonecny12, line 179: What will happen here if parsing will be successful but there won't be any layouts?

rvykydal: Hm, I think it that by the logic here, it can't happen.
Or if you mean the case where the method is called with no nonempty layout_variants, then the behaviour should be preserved (which I think is effectively unsetting the X Layouts by supplying empty strings to the SetX11Keyboard). I don't intend to change it in this PR.

jkonecny12: ok, thanks for explanation.

rvykydal:
jenkins, test this please

jkonecny12:
jenkins, test this please

jkonecny12:
jenkins, test this please

PR #2636, opened 2020-06-03
Fix more SElinux contexts

VladimirSlavik:
Apparently we have them wrong in `/boot/loader/entries` and `/etc/dnf/modules.d/` See the linked bugs for more details. This is a bit workaroundish, but works.

Resolves: rhbz#1775975
Resolves: rhbz#1834189


poncovka:
Please, don't forget to backport it to Fedora. You should remove the label **port to Fedora** after that.

PR #2638, opened 2020-06-03
Don't use data.method.proxy

poncovka:
The command 'method' is no longer provided by the Anaconda's kickstart handler.

Related: rhbz#1691319


poncovka:
Tested.

PR #2639, opened 2020-06-03
Do not build Anaconda on i686

M4rtinK:
As native i686 installations are not supported on RHEL8
and Anaconda is not a i686 compatibility library, it makes
no sense to build it for the i686 architecture.

This also fixes an issue with Anaconda depending on kexec-tools,
which has dropped i686 build. This other packages from being
built due to transitive dependency on i686 kexec-tools package that
no longer exists if those packages have BuildRequires: on anaconda.

Related: rhbz#1691319

### DIFF
--- a/anaconda.spec.in
+++ b/anaconda.spec.in
@@ -148,6 +148,11 @@ Obsoletes: anaconda-runtime < %{version}-%{release}
 Provides: anaconda-runtime = %{version}-%{release}
 Obsoletes: booty <= 0.107-1
 
+# native i686 installations are not supported on RHEL8 and Anaconda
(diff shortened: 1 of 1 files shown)


jkonecny12:
Also please don't forget to backport this to master.

M4rtinK:
@jkonecny I don't think we want to do this on Fedora, given that unlike RHEL8 32 bit x86 is still kinda supported on Fedora.

jkonecny12:
Is it? I thought it is just for multilib which is not the case here.

VladimirSlavik:
Isn't there a 32-bit SIG? My memory is unclear, it could have been also a proposed x32 one...

PR #2642, opened 2020-06-04
Fix the pylint warning in the Payloads module

poncovka:
Disable the warning about potentially unsafe modification of environment
for the environment variable LD_PRELOAD in the Payloads module.


poncovka:
Tests pass when run locally.

PR #2648, opened 2020-06-07
Allow having /boot on a btrfs volume

ignatenkobrain:
It works just fine and in case you go with btrfs-on-luks there is no
reason to create a btrfs volume and just one boot subvolume in it.

Signed-off-by: Igor Raits <i.gnatenko.brain@gmail.com>

### DIFF
--- a/pyanaconda/modules/storage/bootloader/grub2.py
+++ b/pyanaconda/modules/storage/bootloader/grub2.py
@@ -116,7 +116,7 @@ class GRUB2(BootLoader):
     stage2_must_be_primary = False
 
     # requirements for boot devices
-    stage2_device_types = ["partition", "mdarray", "btrfs subvolume"]
+    stage2_device_types = ["partition", "mdarray", "btrfs volume", "btrfs subvolume"]
     stage2_raid_levels = [raid.RAID0, raid.RAID1, raid.RAID4,
                           raid.RAID5, raid.RAID6, raid.RAID10]
     stage2_raid_member_types = ["partition"]


jkonecny12:
Jenkins, it's ok to test.

VladimirSlavik:
It works :) I'll remove the label, check with the rest of the team, and merge.

One thing I noticed was, if I select the `Btrfs` partitioning scheme, I still get `/boot` as a plain old partition. Please keep in mind I do not know enough to say if that should be so, or why. I don't know a thing about btrfs! It's just an observation; maybe it helps, maybe not.

Conan-Kudo:
@VladimirSlavik The default partitioning arrangement when selecting Btrfs has not been adjusted yet, so this is still correct. This change will make it easier to consider making the Btrfs proposal choose to make a subvolume or a volume depending on whether disk encryption is enabled.

VladimirSlavik:
One more thing - @ignatenkobrain, does this fix [bug 1844780](https://bugzilla.redhat.com/show_bug.cgi?id=1844780)? It would be nice if you amended the commit message to include `Resolves: rhbz#1844780` if you think it's true,  or `Related: rhbz#1844780` if not. But I am not going to say it _must_ be there...

ignatenkobrain:
No, that bug is different. It is about that you can't select btrfs in advanced partitioning. This bug makes it actually possible to use btrfs volume for /boot.

VladimirSlavik:
Results look ok, only expected breakage :)

PR #2649, opened 2020-06-08
Don't prompt the user to ssh in VNC installations

poncovka:
Move the checks for the function prompt_for_ssh from anaconda.py. The function
returns True, if the prompt is printed, otherwise False.

When the installation is started with the boot option inst.vnc, don't ask
the user to ssh to the installation environment. Proceed to establish the
VNC connection.

### DIFF
--- a/pyanaconda/startup_utils.py
+++ b/pyanaconda/startup_utils.py
@@ -183,8 +183,23 @@ def setup_logging_from_options(options):
             log.error("Could not setup remotelog with %s", options.remotelog)
 
 
-def prompt_for_ssh():
-    """Prompt the user to ssh to the installation environment on the s390."""
+def prompt_for_ssh(options):
+    """Prompt the user to ssh to the installation environment on the s390.
+
+    :param options: Anaconda command line/boot options
+    :return: True if the prompt is printed, otherwise False
+    """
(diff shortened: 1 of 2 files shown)


jstodola:
Successfully tested on rawhide+updates.img, no prompt for ssh login when inst.vnc is present on the kernel command line, or running a kickstart installation, or running an installation in a KVM guest.
This change looks good to me, thanks! :+1:

pyanaconda/startup_utils.py
```diff
@@ -183,8 +183,23 @@ def setup_logging_from_options(options):
...
-def prompt_for_ssh():
-    """Prompt the user to ssh to the installation environment on the s390."""
+def prompt_for_ssh(options):
+    """Prompt the user to ssh to the installation environment on the s390.
+
+    :param options: Anaconda command line/boot options
+    :return: True if the prompt is printed, otherwise False
+    """
+    if not blivet.arch.is_s390():
+        return False
```
VladimirSlavik, line 193: If it's s390x-only, why not have that in the function name?

poncovka: It doesn't have to be s390x only in the future.

poncovka:
@jstodola, thanks for testing!

PR #2651, opened 2020-06-10
Change the default_partitioning option

poncovka:
Change the default_partitioning option of the Storage section in the Anaconda
configuration file. Instead of the identifier, specify a mount point and its
attributes on each line.

### DIFF
--- a/pyanaconda/core/configuration/storage.py
+++ b/pyanaconda/core/configuration/storage.py
@@ -19,17 +19,12 @@
 #
 from enum import Enum
 
+from blivet.size import Size
+
 from pykickstart.constants import AUTOPART_TYPE_PLAIN, AUTOPART_TYPE_BTRFS, AUTOPART_TYPE_LVM, \
     AUTOPART_TYPE_LVM_THINP
 
 from pyanaconda.core.configuration.base import Section
-
-
-class PartitioningType(Enum):
-    """Type of the default partitioning."""
-    SERVER = "SERVER"
-    WORKSTATION = "WORKSTATION"
-    VIRTUALIZATION = "VIRTUALIZATION"
 
 
 class PartitioningScheme(Enum):
@@ -98,20 +93,6 @@ class StorageSection(Section):
         return self._get_option("file_system_type", str)
 
     @property
-    def default_partitioning(self):
-        """Default partitioning.
-
-        Valid values:
-
-          SERVER          Choose partitioning for servers.
-          WORKSTATION     Choose partitioning for workstations.
-          VIRTUALIZATION  Choose partitioning for virtualizations.
-
-        :return: an instance of PartitioningType
-        """
-        return self._get_option("default_partitioning", PartitioningType)
-
-    @property
     def default_scheme(self):
         """Default partitioning scheme.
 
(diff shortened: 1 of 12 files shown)


pyanaconda/core/configuration/storage.py
```diff
@@ -142,3 +123,78 @@ def luks_version(self):
...
+        """Convert a partitioning string into a list of dictionaries."""
+        return list(map(self._convert_partitioning_line, value.strip().split("\n")))
+
+    @classmethod
+    def _convert_partitioning_line(cls, line):
+        """Convert a partitioning line into a dictionary."""
+        # Parse the line.
+        name, attributes = cls._split_string(line)
+
+        attributes = attributes.strip("()").split(",")
```
jkonecny12, line 158: Please rename `attributes` and `attrs` to `raw_attrs` and `attrs` or the long variants. Point is to avoid having two attributes names here. Or another option is to name `attrs` as `result_attrs`.

pyanaconda/core/configuration/storage.py
```diff
@@ -142,3 +123,78 @@ def luks_version(self):
...
+
+        if attrs.get("max") and not attrs.get("min"):
+            raise ValueError("The attribute 'max' cannot be set without 'min'.")
+
+    @staticmethod
+    def _split_string(value):
+        """Split the given value into two strings."""
+        first, *second = value.strip().split(maxsplit=1)
+        second = "".join(second)
+        return first, second
```
jkonecny12, line 200: The combination of `maxsplit` and `*second` to avoid problems during parsing items without values (encrypted) is really confusing. Could you please solve this not that elegant way but pythonic way, by using exceptions here?

pyanaconda/core/configuration/storage.py
```diff
@@ -142,3 +123,78 @@ def luks_version(self):
...
+        return list(map(self._convert_partitioning_line, value.strip().split("\n")))
+
+    @classmethod
+    def _convert_partitioning_line(cls, line):
+        """Convert a partitioning line into a dictionary."""
+        # Parse the line.
+        name, attributes = cls._split_string(line)
+
+        attributes = attributes.strip("()").split(",")
+        attributes = map(cls._split_string, filter(None, attributes))
```
jkonecny12, line 159: Why do you need to filter out `None` values?

pyanaconda/modules/storage/partitioning/automatic/utils.py
```diff
@@ -213,6 +216,36 @@ def schedule_implicit_partitions(storage, disks, scheme, encrypted=False, luks_f
...
+        swap = name == "swap"
+        spec = PartSpec(
+            mountpoint=name if not swap else None,
+            fstype=None if not swap else "swap",
+            lv=True,
+            thin=not swap,
+            btr=not swap,
+            size=attrs.get("min") or attrs.get("size"),
+            max_size=attrs.get("max"),
+            grow=bool("min" in attrs),
```
jkonecny12, line 239: There is no need to wrap this in `bool`.

pyanaconda/modules/storage/partitioning/automatic/utils.py
```diff
@@ -213,6 +216,36 @@ def schedule_implicit_partitions(storage, disks, scheme, encrypted=False, luks_f
...
+        name = attrs.get("name")
+        swap = name == "swap"
+        spec = PartSpec(
+            mountpoint=name if not swap else None,
+            fstype=None if not swap else "swap",
+            lv=True,
+            thin=not swap,
+            btr=not swap,
+            size=attrs.get("min") or attrs.get("size"),
+            max_size=attrs.get("max"),
```
jkonecny12, line 238: Is it fine to store `None` values here?

pyanaconda/core/configuration/storage.py
poncovka: Fixed.

pyanaconda/core/configuration/storage.py
poncovka: Fixed.

pyanaconda/core/configuration/storage.py
poncovka: Added a comment.

pyanaconda/modules/storage/partitioning/automatic/utils.py
poncovka: Fixed.

pyanaconda/modules/storage/partitioning/automatic/utils.py
poncovka: Yes, it is the default values of the `PartSpec` class.

pyanaconda/modules/storage/partitioning/automatic/utils.py
```diff
@@ -213,6 +216,36 @@ def schedule_implicit_partitions(storage, disks, scheme, encrypted=False, luks_f
...
+            grow=bool("min" in attrs),
+            required_space=attrs.get("free") or 0,
```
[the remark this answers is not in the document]
poncovka: It would work, but this method guarantees that there will never be `None`.

tests/nosetests/pyanaconda_tests/product_test.py
```diff
@@ -29,12 +35,106 @@
 
 PRODUCT_DIR = os.path.join(os.environ.get("ANACONDA_DATA"), "product.d")
 
+SERVER_PARTITIONING = [
```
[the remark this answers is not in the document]
poncovka: The main reason why these constants are moved to the tests is that it is the easiest way to validate this configuration option. This is a big change with many small details that could be easily overlooked, so it is great that we can confirm that the new option will create the same partitioning as before. We might want to rework this part in the future, but there will be always some hard coded values, because you can not really test something without them. I don't see how tests can enforce something.

poncovka:
Updated.

Conan-Kudo:
@jkonecny12 @poncovka Can someone explain why all these mount points are encrypted by default? As far as I know, nobody asked for this yet...

poncovka:
@Conan-Kudo, thanks for pointing that out! The description of the attribute is not correct and the name is indeed unfortunate. See: https://github.com/rhinstaller/anaconda/pull/2691

PR #2653, opened 2020-06-11
Use modinfo to check ko before modprobe

ft272781150:
Hello, everyone:
I install CentOS8  recently, and i found some ERROR info during dracut-pre-udev. such as:
`
dracut-pre-udev[534]: modprobe: FATAL: Module floppy not found in directory /lib/modules/4.18.0-80.el8.x86_64
`
when i finish the installer and reboot os, i check the kernel config for floppy: _CONFIG_BLK_DEV_FD_ is not set.

so, would it be better if we use modinfo to check it exists before we probe it? And, we can still got what ko not loaded if we add inst.debug in cmdline.

### DIFF
--- a/dracut/anaconda-modprobe.sh
+++ b/dracut/anaconda-modprobe.sh
@@ -38,6 +38,10 @@ MODULE_LIST+=" raid0 raid1 raid5 raid6 raid456 raid10 linear dm-mod dm-zero  \
               sha256 lrw xts "
 
 for m in $MODULE_LIST; do
+    if ! modinfo $m >/dev/null 2>&1 ; then
+        echo "anaconda-modprobe: Module $m not found" >&2
+        continue
+    fi
     if modprobe $m ; then
         debug_msg "$m was loaded"
     else


ft272781150:
@poncovka hi~, can  you review my commit ：）

jkonecny12:
Hi, I'm taking this one ;). Thanks for the contribution!

jkonecny12:
Jenkins, it's ok to test.

rvykydal:
I'd consider logging of modinfo not finding a module, so the information is in the log also without using rd.debug. But I don't insist.

ft272781150:
> I'd consider logging of modinfo not finding a module, so the information is in the log also without using rd.debug. But I don't insist.

OK, logging of modinfo not find a module is more friendly for user. So, i will update the commit

VladimirSlavik:
@ft272781150 sorry about this, we are currently really busy with something else, so it might take Jirka some time to get to this :-(

Personally I think maybe you could use the `debug_msg` function, because you mentioned it should be visible with inst.debug? At least that is my interpretation. But I know very little about bash, dracut, and this part of anaconda in general. So it is very possible you know better than me :-)

Anyway, the tests are no longer broken and should pass, except they didn't rerun since the last submission...

Jenkins, test this please.

ft272781150:
> @ft272781150 sorry about this, we are currently really busy with something else, so it might take Jirka some time to get to this :-(
> 
> Personally I think maybe you could use the `debug_msg` function, because you mentioned it should be visible with inst.debug? At least that is my interpretation. But I know very little about bash, dracut, and this part of anaconda in general. So it is very possible you know better than me :-)
> 
> Anyway, the tests are no longer broken and should pass, except they didn't rerun since the last submission...
> 
> Jenkins, test this please.

Hi, it's OK.
In the first version of this commit, i used the _debug_msg_ funciton, just like:
`
if modinfo $m >/dev/null 2>&1 && modprobe $m ; then
    debug_msg "$m was loaded"
else
    debug_msg "$m was NOT loaded"
fi
`
But, just like @rvykydal mentioned,  it cloud be better for user to check which ko is not loaded. we use modinfo before modprobe aimed to avoid _FATAL_ print in dracut. Although current solution also has print for which ko is not loaded, it directly shows info to user and it's not a _FATAL_ print.

ft272781150:
@jkonecny12 hello, do you want to merge this PR :-)

PR #2655, opened 2020-06-11
Remove the dasbus workarounds

poncovka:
Don't use the private attribute for a message bus. The class
DBusServiceIdentifier has the message_bus property now.

Use the service identifier to create a proxy for a specific interface.
Always create this type of a proxy for the RHSM objects.

**Depends on:** https://github.com/rhinstaller/dasbus/pull/36


poncovka:
Jenkins, test this please.

PR #2656, opened 2020-06-11
Prevent crash on unregistration

M4rtinK:
Previously it was possible that the registration thread was still
running when the user clicks the "Unregister" button.

To prevent that, we now wait for the previous subscription thread to
finish before starting the next registration/unregistration thread.

Resolves: rhbz#1845962


M4rtinK:
Port to Fedora: https://github.com/rhinstaller/anaconda/pull/2730

PR #2657, opened 2020-06-11
Fix regression reading kernel list when collecting configurations

jkonecny12:
We have to have payload prepared to be able to read list of kernels from the installation source. However, during transitioning to storage module we moved reading list of kernels to place where the installation tasks are collected instead of where they are executed.

Create a function which will read this list and execute everything later during the installation tasks execution.

*Resolves: rhbz#1846156*


bcl:
Looks good, thanks!

PR #2661, opened 2020-06-16
Take spoke refresh after enter it (#1847493)

ft272781150:
If firset netdev is unplugged, and when we click Network in summary page, the switch which is used to turn device on/off is default visible.
How to reproduce:
- Environmental constraints：The First netdev is unplugged！
- In anaconda summary page, click Network & HostName,you cloud find a switch to turn on/off (even the device is unplugged)
- turn the unplugged device on, and turn the unplugged device off.

View [bugzilla 1847493](https://bugzilla.redhat.com/show_bug.cgi?id=1847493) for detials.

To fix this, i change the refresh order when we click the Network Spoke in Summary Hub.

### DIFF
--- a/pyanaconda/ui/gui/hubs/__init__.py
+++ b/pyanaconda/ui/gui/hubs/__init__.py
@@ -424,8 +424,8 @@ class Hub(GUIObject, common.Hub):
 
         # Enter the spoke
         self._inSpoke = True
-        spoke.refresh()
         self.main_window.enterSpoke(spoke)
+        spoke.refresh()
         # the new spoke should be now visible, trigger the entered signal
         spoke.entered.emit(spoke)
 


ft272781150:
> Thank you for the PR.
> It seems to me that this solution is a kind of too big gun that could cause regressions and the issue does not legitimate such a change. I've come with a solution which is more focused on the place where hiding of the widget is failing:
> #2664

Nice, your solution is better then mine. It is really helpful for me, and make me learned a log. Thanks

jkonecny12:
Can we close this as solved by https://github.com/rhinstaller/anaconda/pull/2664?

ft272781150:
> Can we close this as solved by #2664?

Yes, it's OK

PR #2664, opened 2020-06-17
Fix hiding of network device activation switch (#1847493)

rvykydal:
Sometimes we need to call set_no_show_all(True) to make sure that
the widget disappears.

### DIFF
--- a/pyanaconda/ui/gui/spokes/network.py
+++ b/pyanaconda/ui/gui/spokes/network.py
@@ -1019,8 +1019,9 @@ class NetworkControlBox(GObject.GObject):
 
         switch = self.builder.get_object("device_%s_off_switch" % dev_type_str)
         if dev_type_str == "wired":
-            switch.set_visible(state not in (NM.DeviceState.UNAVAILABLE,
-                                             NM.DeviceState.UNMANAGED))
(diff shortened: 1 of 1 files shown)


pyanaconda/ui/gui/spokes/network.py
```diff
@@ -1019,8 +1019,10 @@ def _refresh_header_ui(self, dev_cfg, state=None):
 
         switch = self.builder.get_object("device_%s_off_switch" % dev_type_str)
         if dev_type_str == "wired":
-            switch.set_visible(state not in (NM.DeviceState.UNAVAILABLE,
-                                             NM.DeviceState.UNMANAGED))
+            visible = state not in (NM.DeviceState.UNAVAILABLE, NM.DeviceState.UNMANAGED)
+            switch.set_visible(visible)
+            if not visible:
+                switch.set_no_show_all(True)
```
poncovka, line 1025: Shouldn't it be `switch.set_no_show_all(not visible)`? Is it a one-time thing?

VladimirSlavik: I guess the call with "reversed" argument is to be avoided?

rvykydal: Modified according to @poncovka's suggestion. Also the rhel-8 branch PR.

PR #2665, opened 2020-06-17
Fix hiding of network device activation switch

rvykydal:
Resolves: rhbz#1628521

Sometimes We need to call set_no_show_all(True) to make sure that
the widget disappears.


rvykydal:
PR with port to Fedora: https://github.com/rhinstaller/anaconda/pull/2664

rvykydal:
Updated to be consistent with master PR.

PR #2666, opened 2020-06-17
Show pylint version in logs

VladimirSlavik:


VladimirSlavik:
Tests should be unborked again, I think... Jenkins, test this please.

PR #2667, opened 2020-06-17
Fix the mock function for DBus.get_proxy

poncovka:
The function should be able to accept other arguments.


poncovka:
Jenkins, test this please.

PR #2669, opened 2020-06-17
Correctly set up the LUKS version when we click on a mount point

poncovka:
When the encryption is changed, set the LUKS version to the current version of
the device factory request or use a default one. Don't use the LUKS version
selected in UI. It might be a version of the previously selected mount point.

Resolves: rhbz#1689699


jkonecny12:
jenkins, test this please

poncovka:
Jenkins, test this please.

PR #2670, opened 2020-06-18
Exclude stderr from returned output when executing powerpc-utils tools

martinezjavier:
The GRUB2 class updateNVRAMBootList() method calls to some tools to update
ppc64le's boot device order list but it doesn't exclude the error messages
which leads to an error message being stored in the NVRAM instead of the
boot list.

To prevent this to happen, exclude the standard error messages from the
output when calling the tools.

Related: rhbz#1847604


poncovka:
Jenkins, test this please.

PR #2672, opened 2020-06-18
subscription: Handle CDN related source switching

M4rtinK:
This pull request implements whats needed to properly handle the CDN souce being default when needed as well as the appropriate sources having priority over the CDN source if available, as well as support for manually switching to and from the CDN source.

**TODO:**
~* fix unit tests~


pyanaconda/payload/dnf/payload.py
```diff
@@ -141,7 +141,11 @@ def set_from_opts(self, opts):
         """
         # Set the source based on opts.method if it isn't already set
         # - opts.method is currently set by command line/boot options
-        if opts.method and (not self.proxy.Sources or self.source_type == DNF_DEFAULT_SOURCE_TYPE):
+
+        # in some cases the CDN is a valid DNF default source
+        default_sources = (DNF_DEFAULT_SOURCE_TYPE, SOURCE_TYPE_CDN)
+        default_source_set = self.source_type in default_sources
```
poncovka, line 147: This should be solved differently. CDN shouldn't be a special case. Also, you want to check `opts.method` before working with DBus sources, because `self.proxy` creates a default one.

pyanaconda/ui/gui/spokes/installation_source.py
```diff
@@ -427,15 +429,22 @@ def __init__(self, *args, **kwargs):
         self._repo_store_lock = threading.Lock()
 
         self._network_module = NETWORK.get_proxy()
+        self._subscription_module = SUBSCRIPTION.get_proxy()
```
poncovka, line 432: This will not work on Fedora.

pyanaconda/ui/gui/spokes/installation_source.py
```diff
@@ -427,15 +429,22 @@ def __init__(self, *args, **kwargs):
...
         repo_changed = self._update_payload_repos()
-
-        if source_changed or repo_changed or self._error:
+        cdn_source = check_cdn_is_installation_source(self.payload)
+        subscribed = self._subscription_module.IsSubscriptionAttached
+        # If CDN is the current installation source but no subscription is
+        # attached there is no need to refresh the installation source,
+        # as without the subscription tokens the refresh would fail anyway.
+        if cdn_source and not subscribed:
+            log.debug("CDN source but no subscribtion attached - skipping payload restart.")
```
poncovka, line 446: Do we really have to skip it the restart? Why?

pyanaconda/ui/gui/spokes/installation_source.py
```diff
@@ -451,7 +460,11 @@ def _update_payload_source(self):
         source_proxy = self.payload.get_source_proxy()
         source_type = source_proxy.Type
 
-        if self._autodetect_button.get_active():
+        if self._cdn_button.get_active():
+            if source_type == SOURCE_TYPE_CDN:
+                return False
+            switch_source(self.payload, SOURCE_TYPE_CDN)
```
poncovka, line 466: Please, extend the SourceSwitchHandler class.

pyanaconda/ui/gui/spokes/installation_source.py
```diff
@@ -602,12 +615,19 @@ def _get_protocol_row(self, protocol):
 
         return None
 
+    @property
+    def cdn_used_as_installation_source(self):
```
poncovka, line 619: Please, use `payload.get_source_type == SOURCE_TYPE_CDN` instead of this property and the function below.

pyanaconda/ui/gui/spokes/installation_source.py
```diff
@@ -625,7 +645,18 @@ def ready(self):
...
+        # is not yet subscribed, the automatic repo refresh will
+        # fail. This is expected as CDN can't be used until the
+        # system has been registered. So prevent the error
+        # message and show CDN is used instead. If CDN still
+        # fails after registration, the regular error message
+        # will be displayed.
+        cdn_source = check_cdn_is_installation_source(self.payload)
+        subscribed = self._subscription_module.IsSubscriptionAttached
+        if cdn_source and not subscribed:
+            return _("Red Hat CDN")
```
poncovka, line 658: Please, use `source_proxy.Description`.

pyanaconda/ui/gui/spokes/installation_source.py
```diff
@@ -922,7 +955,18 @@ def refresh(self):
         self._protocol_combo_box.set_active_id(PROTOCOL_MIRROR)
         self._url_type_combo_box.set_active_id(URL_TYPE_BASEURL)
 
-        if source_type == SOURCE_TYPE_URL:
+        # unset CDN as installation source if the system has been
+        # unregistered in the meantime
+        cdn_button_active = self._cdn_button.get_active()
+        if cdn_button_active and not source_type == SOURCE_TYPE_CDN:
```
poncovka, line 961: This logic is chaotic. Why you cannot set up the source and let the payload restart fail?

pyanaconda/ui/gui/spokes/software_selection.py
```diff
@@ -296,6 +299,13 @@ def status(self):
         if self._error_msgs:
             return _("Error checking software selection")
 
+        cdn_source = check_cdn_is_installation_source(self.payload)
+        subscription_proxy = SUBSCRIPTION.get_proxy()
```
poncovka, line 303: This will not work on Fedora.

pyanaconda/payload/dnf/payload.py
```diff
@@ -194,7 +194,12 @@ def get_source_proxy(self):
 
         :return: a DBus proxy
         """
-        return get_source(self.proxy, DNF_DEFAULT_SOURCE_TYPE)
+        default_source_type = DNF_DEFAULT_SOURCE_TYPE
+        # Default to the CDN if the if the Subscription
+        # DBus module is available.
+        if is_module_available(SUBSCRIPTION):
+            default_source_type = SOURCE_TYPE_CDN
```
poncovka, line 205: I thought that we will set `DNF_DEFAULT_SOURCE_TYPE` to `SOURCE_TYPE_CDN` on the rhel-8 branch. This is not a very good solution, because it breaks every logic that involves `DNF_DEFAULT_SOURCE_TYPE`.

pyanaconda/payload/dnf/payload.py
```diff
@@ -1527,7 +1531,13 @@ def update_base_repo(self, fallback=True, checkmount=True):
...
         # Change the default source to CDROM if there is a valid install media.
         # FIXME: Set up the default source earlier.
-        if checkmount and source_type == DNF_DEFAULT_SOURCE_TYPE and find_optical_install_media():
+
+        # Both DNF default source as well as CDN source should be overriden by
+        # on media source when found.
+        # NOTE: On successfull registration, CDN will be set as installation source
+        #       even if CDROM was set before.
+        can_override_source = source_type in (DNF_DEFAULT_SOURCE_TYPE, SOURCE_TYPE_CDN)
+        if checkmount and can_override_source and find_optical_install_media():
```
poncovka, line 1540: This should be solved differently.

pyanaconda/ui/gui/spokes/subscription.py
```diff
@@ -840,6 +843,16 @@ def _register(self):
         # wait for the previous subscription thread to finish
         threadMgr.wait(THREAD_SUBSCRIPTION)
 
+        # check if the current installation source will be overriden
+        # and remember it if it is the case
+        source_proxy = self.payload.get_source_proxy()
+        source_type = source_proxy.Type
+        if source_type in SOURCE_TYPES_OVERRIDEN_BY_CDN:
+            self._overridden_source_type = source_type
```
poncovka, line 851: Please, keep `self.payload.Sources`. There is no reason to make it so complicated.

pyanaconda/ui/lib/subscription.py
```diff
@@ -66,14 +66,15 @@ def _tear_down_existing_source(payload):
     tear_down_sources(payload.proxy)
 
 
-def set_source_cdn(payload):
-    """Switch to the CDN installation source.
+def switch_source(payload, source_type):
```
poncovka, line 69: This doesn't make sense.

pyanaconda/ui/lib/subscription.py
```diff
@@ -267,20 +268,29 @@ def register_and_subscribe(payload, progress_callback=None, error_callback=None)
     task_proxy = SUBSCRIPTION.get_proxy(task_path)
     task.sync_run_task(task_proxy)
 
-    # report attaching subscription was successful
+    # check if the current installation source should be overridden by
+    # the CDN source we can now use
+    # - at the moment this is true only for the CDROM source
+    source_proxy = payload.get_source_proxy()
+    if source_proxy.Type in SOURCE_TYPES_OVERRIDEN_BY_CDN:
```
poncovka, line 275: Why?

pyanaconda/core/configuration/payload.py
```diff
@@ -71,6 +89,17 @@ def enable_closest_mirror(self):
...
+    @property
+    def default_source(self):
+        """Default installation source.
+
+        Valid values:
+
+        CLOSEST_MIRROR
+        CDN
+        """
+        return self._get_option("default_source", SourceType)
```
poncovka, line 101: We don't need an enum, but the original value. Look at the validation of the selinux option.

pyanaconda/payload/dnf/payload.py
poncovka: You should remove the constant `DNF_DEFAULT_SOURCE_TYPE` from the code and replace it with `conf.payload.default_source`, so you can drop these ugly workarounds.

pyanaconda/payload/dnf/payload.py
poncovka: Same here.

data/anaconda.conf
```diff
@@ -79,6 +79,14 @@ enabled_repositories_from_treeinfo = addon optional variant
 # Enable installation from the closest mirror.
 enable_closest_mirror = True
 
+# Default installation source.
+# Valid values:
+#
+#    CLOSEST_MIRROR  Use closest public repository mirror.
+#    CDN             Use Content Delivery Network (CDN).
+#
+default_source = CLOSEST_MIRROR
```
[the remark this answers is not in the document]
M4rtinK: Oh, right, they inherit from the RHEL config - fixed now. :)

pyanaconda/core/configuration/payload.py
M4rtinK: I was under the impression we wanted to support all the installation sources but for now document just the two we want to use now, so for that I used the Enum validation as is used for example for partitioning type.

If we want to support only closest mirror and CDN for now, I can indeed use the simplified method as used by SELinux.

pyanaconda/core/configuration/payload.py
```diff
@@ -71,6 +89,17 @@ def enable_closest_mirror(self):
...
         return self._get_option("enable_closest_mirror", bool)
 
+    @property
+    def default_source(self):
+        """Default installation source.
+
+        Valid values:
+
+        CLOSEST_MIRROR
+        CDN
```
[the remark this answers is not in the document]
M4rtinK: Done.

pyanaconda/core/configuration/payload.py
```diff
@@ -17,7 +17,25 @@
...
 #
+from enum import Enum
+
 from pyanaconda.core.configuration.base import Section
+from pyanaconda.core.constants import SOURCE_TYPE_LIVE_OS_IMAGE, SOURCE_TYPE_HMC, \
+    SOURCE_TYPE_CDROM, SOURCE_TYPE_CLOSEST_MIRROR, SOURCE_TYPE_REPO_FILES, \
+    SOURCE_TYPE_NFS, SOURCE_TYPE_URL, SOURCE_TYPE_HDD, SOURCE_TYPE_CDN
+
+
+class SourceType(Enum):
```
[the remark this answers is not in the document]
M4rtinK: Removed.

pyanaconda/ui/gui/spokes/installation_source.py
M4rtinK: If installation source is set to the CDN but we are not yet registered, any attempt to interact with the CDN repos will fail. The payload refresh also sets a "Failed to set up installation source" (IIRC) error message even though nothing really is wrong. 

Only after the user has registered is it possible to use the CDN as a source (`redhat.repo` has been generated and the needed authentication tokens are in place) and payload restart has a chance to succeed.

Do you think there is a specific reason to still restart the payload in this case ? In that case we would need some option to skip repo refresh, but I don't see anything like that in the `restart_thread()` method:

https://github.com/rhinstaller/anaconda/blob/be7d7f6cbe96885657351c338254c34b37268a1c/pyanaconda/payload/manager.py#L115

pyanaconda/ui/gui/spokes/installation_source.py
M4rtinK: We need the `switch_source()` for use in the subscription handling thread, so I wanted to avoid code duplication by using that.

pyanaconda/ui/gui/spokes/installation_source.py
M4rtinK: Changed.

pyanaconda/ui/gui/spokes/installation_source.py
M4rtinK: Yep, that's indeed cleaner. :)

pyanaconda/ui/gui/spokes/installation_source.py
M4rtinK: I might be getting a bit lost here - how is this related to the payload restart ?

These conditionals make sure the CDN radio button is enabled/disabled based on CDN being set as installation source and the other conditionals do the same. The one specifics o the CDN source is that in some cases (CDROM source) the CDN source might be switched to another source (CDROM) outside of the Source spoke and we need to handle that.

pyanaconda/ui/lib/subscription.py
M4rtinK: Back during 8.2 the conclusion was reached that RHEL DVD images should default to sing the CDN installation source by default as by the time customers actually get their hands on the image, there will be many updates, many of them security sensitive. And over time more and more packages on the DVD will be outdated while the packages from the CDN will be always up to date.

By defaulting to installation from the CDN even on the DVD installation media the resulting system will be secure and the customer can also avoid installing updates right on the fist start after the installation, resulting in better experience.

To implement this, we need to track if the currently set source is one of those that can be overridden and then also roll it back on un-registration to maintain consistency.

pyanaconda/payload/dnf/payload.py
M4rtinK: Yup, much cleaner with the config option. :)

pyanaconda/payload/dnf/payload.py
M4rtinK: Used the config option, much cleaner now.

pyanaconda/ui/gui/spokes/installation_source.py
poncovka: My mistake, it is not related to the payload restart. However, this logic is still too complicated for no reason.

| Is the button active? | Is the current type CDN? | Should the button stay active? | 
| --- | --- | --- |
| 0 | 0 | 0 |
| 0 | 1 | 1 |
| 1 | 0 | 0 |
| 1 | 1 | 1 |

Based on the table, the button activation depends only on the current type of the source. Therefore, you can replace this code:
```
        cdn_button_active = self._cdn_button.get_active()
        if cdn_button_active and not source_type == SOURCE_TYPE_CDN:
            log.debug("source spoke: unregistered since last visit, turning off CDN button")
            self._cdn_button.set_active(False)

        # Set CDN radio button active, in case external conditions changed since the last refresh.
        if source_type == SOURCE_TYPE_CDN and not self._cdn_button.get_active():
            log.debug("source spoke: CDN enabled, enabling CDN button")
            self._cdn_button.set_active(True)
        elif source_type == SOURCE_TYPE_URL:
            ...
```

With this one:

```    
        if source_type == SOURCE_TYPE_CDN:
            self._cdn_button.set_active(True)
        elif source_type == SOURCE_TYPE_URL:
            ...

```

pyanaconda/payload/dnf/payload.py
```diff
@@ -141,7 +141,8 @@ def set_from_opts(self, opts):
         """
         # Set the source based on opts.method if it isn't already set
         # - opts.method is currently set by command line/boot options
-        if opts.method and (not self.proxy.Sources or self.source_type == DNF_DEFAULT_SOURCE_TYPE):
+        current_source_is_default = self.source_type == conf.payload.default_source
```
poncovka, line 144: The property `self.source_type` will create a default source if there is none. You don't really want that in this case. Please, move this code back to the condition or create a new method `is_source_default` and call it AFTER we check `opts.method` and `not self.proxy.Sources`.

pyanaconda/ui/lib/subscription.py
poncovka: It is just a very confusing implementation.

pyanaconda/ui/gui/spokes/installation_source.py
M4rtinK: I don't think this is correct - how do you turn the button OFF when CDN is not the installation source at spoke entry with this ? This what the first condition does.

Ideally, there would be just:

`self._cdn_button.set_active(source_type == SOURCE_TYPE_URL)`

Still, then the if/else tree down there would get executed resulting in:

```
        else:
            ValueError("Unsupported source type: '{}'".format(source_type))
```

pyanaconda/payload/dnf/payload.py
M4rtinK: Sounds good, moving to a separate method.

pyanaconda/ui/gui/spokes/installation_source.py
M4rtinK: Actually, a lot of this is no longer needed with the rebased Anaconda - in 8.2 there wasn't really a CDN source and the CDN radio button held part of the state.

Now there is a proper CDN source, so all of these previous considerations are not really needed anymore and a simple condition should be enough. :)

pyanaconda/payload/dnf/payload.py
```diff
@@ -1511,6 +1515,14 @@ def _unmount_source_directory(self, mount_point):
             else:
                 payload_utils.unmount(mount_point, raise_exc=True)
 
+    def is_source_default(self):
+        """Report if the current source type is the default source type.
+
+        NOTE: If not source was set previously a new default one
```
rvykydal, line 1521: "not" -> "no" ?

M4rtinK: Fixed. :)

poncovka:
Please, port all commits that change the payload support to upstream.

M4rtinK:
Port to Fedora: #2754

PR #2674, opened 2020-06-19
Do not require treeinfo

jkonecny12:
For mountable sources we want to make stupid check if the folder is a repository. We did that by checking .treeinfo file but that works only for composes and not for repositories created by `createrepo_c` utility.

Instead check if `repodata/repomd.xml` file exists.

*Resolves: rhbz#1844287*
*Resolves: rhbz#1849093 (rhel-8)*

### DIFF
--- a/pyanaconda/modules/payloads/source/harddrive/initialization.py
+++ b/pyanaconda/modules/payloads/source/harddrive/initialization.py
@@ -22,8 +22,7 @@ from collections import namedtuple
 from pyanaconda.modules.common.errors.payload import SourceSetupError
 from pyanaconda.modules.common.task import Task
 from pyanaconda.modules.payloads.source.utils import find_and_mount_device, \
-    find_and_mount_iso_image
-from pyanaconda.payload.image import verify_valid_installtree
+    find_and_mount_iso_image, verify_valid_installtree
 from pyanaconda.payload.utils import unmount
 from pyanaconda.anaconda_loggers import get_module_logger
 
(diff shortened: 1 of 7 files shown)


AdamWill:
Shouldn't the function be renamed, then? It seems like the name is a lie: you're not checking if the thing is an 'install tree', and you're not using it as one. You're checking if it's a repository and using it as a repository.

jkonecny12:
> Shouldn't the function be renamed, then? It seems like the name is a lie: you're not checking if the thing is an 'install tree', and you're not using it as one. You're checking if it's a repository and using it as a repository.

Honestly I don't see much difference between install tree and repository but yeah. Repository is more clear meaning.

I'll change that on Monday.

AdamWill:
I'm not sure if something can be an install tree without being a repository (I'd have to check if the netinst images have a `.installtree` file or not), but it can definitely be a repository without being an install tree :)

jkonecny12:
The netinst images should not have it as far as I know. `.treeinfo` file is abstraction of a repository describing metada and/or joining multiple repositories together. Basically if `.treeinfo` is found it will be used instead of the repository in that folder.

However, I'm not sure how we should approach this. The problem is that another option how to specify repository could be by pointing it to the directory where is only this .treeinfo file. This treeinfo file will then redirect us to an existing path where the repository is. It will work right (I think) now because the resolution is done outside of the source modules code but we need to take this into account for the future complete modular payload where consumers of the API should create sources directly. 
The question is, how should we react on empty dir with just a `.treeinfo` file? We will discuss this in a team on Monday, I guess.

jkonecny12:
**UPDATED:**
* Rename installtree to repository in the function name.
* Add RHEL-8 bug for easier backporting.

jkonecny12:
We have an approach for the future development. For now this should be good enough solution. @AdamWill I've changed the name. Thanks for the suggestion.

PR #2678, opened 2020-06-23
Don't fail in the exit handler if no storage is available

poncovka:
The exit handler shouldn't fail if there is no storage model available in the
Storage module. It means that Anaconda hasn't finished the device scanning.
We should handle the UnavailableStorageError exception and continue.

### DIFF
--- a/pyanaconda/startup_utils.py
+++ b/pyanaconda/startup_utils.py
@@ -38,6 +38,11 @@ from pyanaconda import safe_dbus
 from pyanaconda import kickstart
 from pyanaconda.flags import flags
 from pyanaconda.screensaver import inhibit_screensaver
+from pyanaconda.modules.common.constants.objects import DEVICE_TREE
+from pyanaconda.modules.common.constants.services import STORAGE
+from pyanaconda.modules.common.errors.storage import UnavailableStorageError
+from pyanaconda.modules.common.structures.storage import DeviceData
+from pyanaconda.modules.common.task import sync_run_task
 
 import blivet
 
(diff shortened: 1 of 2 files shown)


pyanaconda/startup_utils.py
```diff
@@ -373,3 +378,38 @@ def parse_kickstart(ks, addon_paths, strict_mode=False):
...
+        device_tree = STORAGE.get_proxy(DEVICE_TREE)
+
+        for device_name in device_tree.FindOpticalMedia():
+            device_data = DeviceData.from_structure(
+                device_tree.GetDeviceData(device_name)
+            )
+            optical_media.append(device_data.path)
+
+    except UnavailableStorageError:
+        log.debug("No storage model is available.")
```
jkonecny12, line 400: I think this should be warning.

pyanaconda/startup_utils.py
```diff
@@ -373,3 +378,38 @@ def parse_kickstart(ks, addon_paths, strict_mode=False):
...
+    """Tear down the storage module."""
+    try:
+        storage_proxy = STORAGE.get_proxy()
+
+        for task_path in storage_proxy.TeardownWithTasks():
+            task_proxy = STORAGE.get_proxy(task_path)
+            sync_run_task(task_proxy)
+
+    except UnavailableStorageError:
+        log.debug("No storage model is available.")
```
jkonecny12, line 415: The same here.

pyanaconda/startup_utils.py
poncovka: Fixed.

pyanaconda/startup_utils.py
poncovka: Fixed.

poncovka:
> Just a question what will be the behavior now? Anaconda will work but without showing any disks?

The exit handler will able to finish the clean-up if Anaconda fails too soon.

poncovka:
Updated.

poncovka:
The issue will be fixed with https://github.com/rhinstaller/anaconda/pull/2703.

PR #2680, opened 2020-06-23
Always specify the boot disk (#1630299)

poncovka:
We should always specify the boot disk when we allocate partitions. Otherwise,
Blivet will choose one of the available disks that don't have to be valid.

Resolves: rhbz#1630299


jstodola:
Testing on s390x was successful, /boot was placed on the correct device (DASD) instead of an iSCSI LUN.
Thank you for this fix!

PR #2681, opened 2020-06-23
Reconfigure DNF payload after options are set

VladimirSlavik:
Fixes broken multilib boot option.

Resolves: rhbz#1847603

### DIFF
--- a/pyanaconda/payload/dnf/payload.py
+++ b/pyanaconda/payload/dnf/payload.py
@@ -174,6 +174,9 @@ class DNFPayload(Payload):
         if opts.multiLib:
             self.data.packages.multiLib = opts.multiLib
 
+        # Reset all the other things now that we have new configuration.
+        self._configure()
+
     @property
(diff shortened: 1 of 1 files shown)


pyanaconda/payload/dnf/payload.py
```diff
@@ -174,6 +174,9 @@ def set_from_opts(self, opts):
         if opts.multiLib:
             self.data.packages.multiLib = opts.multiLib
 
+        # Reset all the other things now that we have new configuration.
+        self.unsetup()
```
poncovka, line 178: I was looking into that a little more and might be better to directly call `self._configure()`, because `unsetup` does other unnecessary steps (like the tear down of the sources).

VladimirSlavik:
Thank you for the suggestions - changed to `_configure()`.

VladimirSlavik:
@rvykydal do you want this merged now?

PR #2684, opened 2020-06-24
Add table of branches to README

jkonecny12:
This table can be used by contributors to pick correct branch where they want to make their contribution.

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -41,6 +41,19 @@ How to Contribute to the Anaconda Installer (the short version)
 
 If you want to contribute a change to both the upstream and RHEL Anaconda then follow both a) and b) separately.
 
+Which is my target git branch?
+------------------------------
+
+Depending on where you want to make your contribution please choose your correct branch based on the table below.
+
++--------------------------+--------------+
+| Fedora Rawhide           | master       |
++--------------------------+--------------+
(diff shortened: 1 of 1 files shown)


jkonecny12:
I think we want to make this table visible on the first look in the repository. Because we're getting questions about where to contribute from people who knows the contribution guidelines.

Other than that, yes, we can replace the order even though that this seems more logical to me. You are looking for a branch based on where you want to contribute not opposite.

poncovka:
I guess that the problem is that the style of our README doesn't really work anymore. It will always look weird to add some reasonable content to all these shouts and one-liner sentences.

About the order, it is about the length of the cells. The branch names will have very similar length, but the descriptions can be very different. With the current order, it might be harder to read the table. But that's up to you.

jkonecny12:
True, OK. I'll change that.

jkonecny12:
**UPDATED:**
* Moved from readme to contributions document

VladimirSlavik (approved):
Looks good to me. Thank you!

poncovka (approved):
Looks good to me. Thanks!

jkonecny12:
**UPDATED:**
* Removed RHEL-7 because Anaconda can't update that now.
* Switch `rhel-8` to `rhel-X` to make this more general.

poncovka (approved):

PR #2685, opened 2020-06-24
network: do not enforce network standalone spoke on default source

rvykydal:
We used to show standalone network spoke asking for network configuration (not
mandatory) before summary hub only in cases of explicitly set url or nfs
software sources.  We should keep this behaviour in rhel.

On Fedora there are automatic connections created at the beginning of stage2 so
the chances of the standalone spoke being showed are rather small anyway.

### DIFF
--- a/pyanaconda/ui/gui/spokes/network.py
+++ b/pyanaconda/ui/gui/spokes/network.py
@@ -1628,10 +1628,10 @@ class NetworkStandaloneSpoke(StandaloneSpoke):
 
     @property
     def completed(self):
-        # If we can't configure network, don't require it
         return (not conf.system.can_configure_network
                 or self._network_module.GetActivatedInterfaces()
-                or not self.payload.needs_network)
+                or not (self.payload.source_type != conf.payload.default_source
+                        and self.payload.needs_network))
 
     def initialize(self):
         register_secret_agent(self)


pyanaconda/ui/gui/spokes/network.py
```diff
@@ -1631,7 +1631,7 @@ def completed(self):
         # If we can't configure network, don't require it
         return (not conf.system.can_configure_network
                 or self._network_module.GetActivatedInterfaces()
-                or not self.payload.needs_network)
+                or self.payload.source_type not in (SOURCE_TYPE_URL, SOURCE_TYPE_NFS))
```
poncovka, line 1634: The commit message says something else then the code. If we want to enforce this behaviour, please add here a comment with an explanation. Otherwise, we might change it again in the future.

It might be still better to combine both conditions:

`not (self.payload.source_type in (SOURCE_TYPE_URL, SOURCE_TYPE_NFS) and self.payload.needs_network)`

However, I think that a more universal solution would be:

`not (self.payload.source_type != conf.payload.default_source and self.payload.needs_network)`

It would require the first commit from https://github.com/rhinstaller/anaconda/pull/2672.

rvykydal: I like the more universal solution, kind of of expected some suggestion like this - thank you.
I'll update the PR when the required commit is merged.
And I'll remove the comment - all should be clear from the condition.

rvykydal: Done

rvykydal:
PR with port to master: https://github.com/rhinstaller/anaconda/pull/2714

PR #2686, opened 2020-06-25
Fix software spoke message when source changes

M4rtinK:
With the Red Hat CDN now easily available the users will now
trigger the software spoke message that tells them to re-enter
the software spoke and confirm the software selection.

Unfortunately the current message - "Nothing selected" is
rather confusing. It does not tell what is not selected and
might actually be incorrect, as in some cases the previous
selection might transfer over when the installation source
changes.

So change the confusing message to "Please confirm software
selection" which is also more in line with our other spoke
status messages.

Thanks a lot to Jan Stodola for coming up with the text of
the new message. :)

Resolves: rhbz#1788458

### DIFF
--- a/pyanaconda/ui/gui/spokes/software_selection.py
+++ b/pyanaconda/ui/gui/spokes/software_selection.py
@@ -313,12 +313,12 @@ class SoftwareSelectionSpoke(NormalSpoke):
                     return _("Invalid environment specified in kickstart")
             # we have no packages section in the kickstart and no environment has been set
             elif not self.environment:
-                return _("Nothing selected")
+                return _("Please confirm software selection")
 
         if not flags.automatedInstall:
             if not self.environment:
                 # No environment yet set
-                return _("Nothing selected")
+                return _("Please confirm software selection")
             elif not self.environment_valid:
                 # selected environment is not valid, this can happen when a valid environment
                 # is selected (by default, manually or from kickstart) and then the installation


rvykydal:
I don't want to complicate things, but my view is:
In the proposed message, the "software" seems redundant to me.
Also, for consistency, I'd rather display status then hint here - a hint is quite unprecedented in the status of a spoke but I might be wrong. Something like:
"Selection not confirmed"

pyanaconda/ui/gui/spokes/software_selection.py
```diff
@@ -313,12 +313,12 @@ def status(self):
                     return _("Invalid environment specified in kickstart")
             # we have no packages section in the kickstart and no environment has been set
             elif not self.environment:
-                return _("Nothing selected")
+                return _("Please confirm software selection")
```
poncovka, line 316: There is a big difference between `Nothing selected` and `Please confirm software selection`. We shouldn't ask users to confirm something if we are not sure that there is something to confirm. We should have something like this in the `status` property:

```
if not self.completed:
  return _("Please confirm software selection")
```

rvykydal: Hm, I am not able to figure how using self.completed can help here.
If we get to the point of "Nothing selected" messages (happens only in case %packages are not specified) then self.completed is either (processing_done is always true at this point)
- True if self.environment exists and is invalid
- False otherwise

M4rtinK:
Port to fedora: #2734

PR #2689, opened 2020-06-25
Add product data for CentOS Stream

bstinsonmhk:
We need product data for CentOS Stream in addition to CentOS Linux. This is just a copy of the CentOS Linux data with the product_name changed.

Can we be sure this also makes it into a RHEL build of anaconda?

### DIFF
--- a/data/product.d/centos-stream.conf
+++ b/data/product.d/centos-stream.conf
@@ -1,0 +1,29 @@
+# Anaconda configuration file for CentOS Stream.
+
+[Product]
+product_name = CentOS Stream
+
+[Base Product]
+product_name = Red Hat Enterprise Linux
+
+[Anaconda]
+# List of enabled Anaconda DBus modules for RHEL.
+#  but without org.fedoraproject.Anaconda.Modules.Subscription
+kickstart_modules =
+     org.fedoraproject.Anaconda.Modules.Timezone
+     org.fedoraproject.Anaconda.Modules.Network
+     org.fedoraproject.Anaconda.Modules.Localization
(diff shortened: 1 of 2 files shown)


data/product.d/centos-stream.conf
```diff
@@ -0,0 +1,29 @@
+# Anaconda configuration file for CentOS Stream.
+
+[Product]
+product_name = CentOS Stream
+
+[Base Product]
+product_name = Red Hat Enterprise Linux
```
poncovka, line 7: Hi, it looks like the configuration file is just a copy-paste of `centos.conf`. In that case, you can change the `product_name` of the base product to `CentOS Linux` and remove the sections below. We will load them automatically from the base products.

data/product.d/centos-stream.conf
```diff
@@ -0,0 +1,29 @@
+# Anaconda configuration file for CentOS Stream.
+
+[Product]
+product_name = CentOS Stream
```
poncovka, line 4: Please, extend the `rhel_products_test` method in `product_test.py`, so the file can be validated by unit tests.

poncovka:
> Can we be sure this also makes it into a RHEL build of anaconda?

Please, open a new pull request for the rhel-8 branch. The commit message should contain a line `Resolves: rhbz#12345` or `Related: rhbz#12345` with a number of the related bugzilla bug.

VladimirSlavik:

jkonecny12:
Hi @bstinsonmhk , do you want me to handle the commit squashing?

poncovka:
Fixed in https://github.com/rhinstaller/anaconda/pull/3014.

PR #2690, opened 2020-06-25
Fix RHSM HTTP proxy configuration crash in the GUI

M4rtinK:
The RHSM HTTP proxy that configured in the Subscription
spoke had a couple issues that have been addressed:

- there was a typo in one of the variables, preventing the correct
  port from being set
- the proxy password is a secret value and needs to be set to
  the DBus structure accordingly
- some hostnames users type into the proxy hostname fields
  might be incorrect - generally the intermediate states of
  a hostname as they type it - and we need to ignore those
  until the final valid hostname is typed in

Resolves: rhbz#1849792

### DIFF
--- a/pyanaconda/ui/gui/spokes/subscription.py
+++ b/pyanaconda/ui/gui/spokes/subscription.py
@@ -26,7 +26,7 @@ from pyanaconda.core.i18n import _, CN_
 from pyanaconda.core.constants import SECRET_TYPE_HIDDEN, \
     SUBSCRIPTION_REQUEST_TYPE_USERNAME_PASSWORD, SUBSCRIPTION_REQUEST_TYPE_ORG_KEY, \
     THREAD_SUBSCRIPTION, SOURCE_TYPES_OVERRIDEN_BY_CDN
-from pyanaconda.core.payload import ProxyString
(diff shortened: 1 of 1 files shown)


pyanaconda/ui/gui/spokes/subscription.py
```diff
@@ -390,28 +390,41 @@ def on_http_proxy_checkbox_toggled(self, checkbox):
...
-        self.subscription_request.server_proxy_password = editable.get_text()
+        password = editable.get_text()
+        # check if password field is non empty
+        if password:
+            # there is a password, set it
+            self.subscription_request.server_proxy_password.set_secret(password)
+        else:
+            # there is no password, clear the password (if any) by setting
+            # the secret to None
+            self.subscription_request.server_proxy_password.set_secret(None)
```
poncovka, line 427: You can replace the `if-else` block with:

`self.subscription_request.server_proxy_password.set_secret(password or None)`

M4rtinK: Good idea, changed. :)

M4rtinK:
Port to fedora: #2732

PR #2691, opened 2020-06-26
Remove the encrypted attribute

poncovka:
The encrypted attribute of the default_partitioning configuration option
can be removed. It will be allowed to encrypt all mount points specified
in the Anaconda configuration files. It won't affect boot partitions that
are defined somewhere else. It will affect the default partitioning of Red
Hat Virtualization and oVirt Node Next, but we are able to with deal that.

### DIFF
--- a/data/anaconda.conf
+++ b/data/anaconda.conf
@@ -152,11 +152,10 @@ file_system_type =
 #   min <MIN_SIZE> The size will grow from MIN_SIZE to MAX_SIZE.
 #   max <MAX_SIZE> The max size is unlimited by default.
 #   free <SIZE>    The required available space.
-#   encrypted      The mount point will be encrypted.
 #
 default_partitioning =
-    /     (min 1 GiB, max 70 GiB, encrypted)
-    /home (min 500 MiB, free 50 GiB, encrypted)
+    /     (min 1 GiB, max 70 GiB)
+    /home (min 500 MiB, free 50 GiB)
 
 # Default partitioning scheme.
 # Valid values:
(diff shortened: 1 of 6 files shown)


poncovka:
I am also thinking about changing the attribute. There are several options:

1. Rename it to `allow_encryption` (or something like that).
2. Rename it to `no_encryption` (or something like that) and revert the logic.
3. Rename it to `encryption` with supported values `never` and `allowed` and possibly change the default. 
4. Remove the attribute and decide it automatically from the mount point.

The last option would require to ask someone from RHEV if we can change their partitioning. They never allow encryption of their mount points, but I think it is not intentional.

data/anaconda.conf
```diff
@@ -155,7 +155,7 @@ file_system_type =
 #   min <MIN_SIZE> The size will grow from MIN_SIZE to MAX_SIZE.
 #   max <MAX_SIZE> The max size is unlimited by default.
 #   free <SIZE>    The required available space.
-#   encrypted      The mount point will be encrypted.
+#   encrypted      Allow to encrypt the mount point if requested.
```
Conan-Kudo, line 158: Maybe change this to `encryptable`?

poncovka:

poncovka:
We are considering to remove the `encrypted` attribute from the config file, but first I need to ask someone from RHEV about their default partitioning.

Conan-Kudo:
@poncovka If you did that, wouldn't that mean all partitions would be assumed to be encryptable?

poncovka:
@Conan-Kudo, yes, all partitions defined in the configuration files would be cryptable. However, that is pretty much the current state. Boot partitions are defined somewhere else, so it wouldn't affect them.

poncovka:
Updated.

poncovka:
Updated.

PR #2692, opened 2020-06-26
Remove the support for Fedora Atomic Host

poncovka:
The Atomic Host platform is being replaced by CoreOS and the Fedora project is
no longer producing new versions of Atomic Host, so let's remove the support
in Anaconda.


poncovka:
@cgwalters, is it a good time to remove support for Fedora Atomic Host?

cgwalters:
Yep, sounds fine to me!

pyanaconda/localization.py
```diff
@@ -649,66 +648,3 @@ def setup_locale_environment(locale=None, text_mode=False, prefer_environment=Fa
     for varname in ("LANGUAGE", "LC_ALL", "LC_MESSAGES"):
         if varname in os.environ:
             del os.environ[varname]
-
-
-def get_locale_map_from_ostree(repo, ref):
```
jlebon, line 654: Hmm, this code should still be relevant to Fedora Silverblue I think (which at least for the time being is still using Anaconda).

poncovka: This function is used only when the configuration option `check_supported_locales` is enabled. Based on https://github.com/rhinstaller/anaconda/pull/1691#issuecomment-440320475, Fedora Silverblue isn't supposed to ship a subset of locales unlike Fedora Atomic Host. @tpopela, could you confirm that, please?

tpopela: Yes, we don't (at least for now) plan to ship/use some subset of available locales. But I leave the decision on @jlebon or @cgwalters as they will know more from technical side of things.

jlebon: Ahh yup, you're correct. Thanks for the reminder. And anyway if we ever want to add this to Silverblue or some other variant, we can always revive it from git. So :+1:  on nuking this too.

PR #2696, opened 2020-06-30
Don't override the eula command with the same command

poncovka:
Remove the lines that override the eula kickstart command.


VladimirSlavik:
Why?

poncovka:
The current code is basically doing:
```
x = y
y = x
```

There is no good reason to do that.

PR #2701, opened 2020-07-02
lsblk in RHEL-8 doesn't support FSAVAIL and FSUSE%

jstodola:
This reverts commit 02ca57d0f7ab3d1214647b228e88b7d4a9e3202c.

Resolves: rhbz#1853470


poncovka:
Jenkins, test this please.

VladimirSlavik:
Let's check if we should not revert more with it? Originally we used an option, then replaced that with the column list because we added something. Let's check if the result isn't missing anything useful that was there before.

PR #2708, opened 2020-07-07
Set up FIPS during the installation

poncovka:
Add the systemd service anaconda-fips that will set up the FIPS mode in stage2
if there is fips=1 on the kernel cmdline. It should run before network.target.

Bring the service anaconda-sshd unit closer to sshd.service and respect
the crypto policy of the installation environment.

Copy the crypto policy from the installation environment to target system
the before the package installation. The RPM scriptlets needs to be executed
in the FIPS mode if there is fips=1 on the kernel cmdline.

**Ported from:** https://github.com/rhinstaller/anaconda/pull/2537 https://github.com/rhinstaller/anaconda/pull/2621

### DIFF
--- a/anaconda.spec.in
+++ b/anaconda.spec.in
@@ -123,6 +123,10 @@ Requires: systemd
 Requires: python3-pid
 Requires: python3-ordered-set >= 2.0.0
 
+# Required by the systemd service anaconda-fips.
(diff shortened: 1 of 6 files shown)


poncovka:
Rebased and tested.

PR #2710, opened 2020-07-07
Exclude stderr from returned output when executing powerpc-utils tools

poncovka:
The GRUB2 class updateNVRAMBootList() method calls to some tools to update
ppc64le's boot device order list but it doesn't exclude the error messages
which leads to an error message being stored in the NVRAM instead of the
boot list.

To prevent this to happen, exclude the standard error messages from the
output when calling the tools.

(cherry-picked from a commit c790ad9)

**Ported from:** https://github.com/rhinstaller/anaconda/pull/2670


VladimirSlavik:
Jenkins, test this please.

PR #2711, opened 2020-07-07
subscription: Fix rhsm --proxy kickstart command usage with no userna…

M4rtinK:
…me specified

The ProxyString class used for parsing proxy definitions in Anaconda
set's it's username property to None if the proxy definition does not
have any username specified.

This is a problem if the None get's forwarded to the SubscriptionRequest
DBus structure, which expects an empty string instead, leading to
a crash once such incorrectly filled structure is first serialized.

To prevent that always check if ProxyString returned something for
the username and use "" if not, preventing the serialization crash.

Resolves: rhbz#1849792


M4rtinK:
Port to fedora: #2732

PR #2718, opened 2020-07-08
Fix the position of the info bar in standalone spokes (#1847423)

poncovka:
Move the info bar of standalone spokes at the bottom of the window. Otherwise,
the Quit and Continue buttons will be incorrectly placed below the info bar.

Resolves: rhbz#1847423


poncovka:
Before:

Screenshot_vs_main_2020-06-16_12 56 00

After:

eula_warning

PR #2720, opened 2020-07-08
ostree: set rootflags when installing on btrfs (#1753485)

davide125:
When / is on a btrfs subvolume, we need to pass `rootflags` so it can be mounted correctly. This is an attempt at fixing https://bugzilla.redhat.com/show_bug.cgi?id=1753485

Full disclosure: I have no idea what I'm doing here, and this is completely untested so far.

### DIFF
--- a/pyanaconda/payload/rpmostreepayload.py
+++ b/pyanaconda/payload/rpmostreepayload.py
@@ -28,6 +28,7 @@ from pyanaconda.core.constants import PAYLOAD_TYPE_RPM_OSTREE
 from pyanaconda.core.i18n import _
 from pyanaconda.modules.common.constants.objects import BOOTLOADER, DEVICE_TREE
 from pyanaconda.modules.common.constants.services import STORAGE
+from pyanaconda.modules.common.structures.storage import DeviceData
 from pyanaconda.progress import progressQ
 from pyanaconda.payload.base import Payload
 from pyanaconda.payload import utils as payload_utils
@@ -445,11 +446,19 @@ class RPMOSTreePayload(Payload):
             # such.
             bootloader = STORAGE.get_proxy(BOOTLOADER)
             device_tree = STORAGE.get_proxy(DEVICE_TREE)
-            root_device = device_tree.GetRootDevice()
+
+            root_name = device_tree.GetRootDevice()
+            root_data = DeviceData.from_structure(
+                device_tree.GetDeviceData(root_name)
+            )
 
             set_kargs_args = ["admin", "instutil", "set-kargs"]
             set_kargs_args.extend(bootloader.GetArguments())
-            set_kargs_args.append("root=" + device_tree.GetFstabSpec(root_device))
+            set_kargs_args.append("root=" + device_tree.GetFstabSpec(root_name))
+
+            if root_data.type == "btrfs subvolume":
+                set_kargs_args.append("rootflags=subvol=" + root_name)
+
             self._safe_exec_with_redirect("ostree", set_kargs_args, root=conf.target.system_root)
 
 


pyanaconda/payload/rpmostreepayload.py
```diff
@@ -450,6 +450,8 @@ def post_install(self):
             set_kargs_args = ["admin", "instutil", "set-kargs"]
             set_kargs_args.extend(bootloader.GetArguments())
             set_kargs_args.append("root=" + device_tree.GetFstabSpec(root_device))
+            if root_device.type == "btrfs subvolume":
+                set_kargs_args.append("rootflags=subvol=" + image.device.name)
```
mark-otaris, line 454: Should possibly be:

```suggestion
                set_kargs_args.append("rootflags=subvol=" + root_device.name)
```

davide125: Yes, definitely. Thanks!

Conan-Kudo:
@davide125 You should probably add `Resolves: rhbz#1753485` as a second line in your commit message, per the Anaconda patch guidelines.

poncovka:
Jenkins, test this please.

jkonecny12:
Jenkins, it's ok to test.

jkonecny12:
You can ignore the test result. Problem is a new version of pykickstart.

felipeborges:
https://twitter.com/pid_eins/status/1281493584081027072
> That bug looks weird. You shouldn't have to set rootflags= to pick the default subvolume, that's just fragile. Use `btrfs subvol set-default` to set the default subvolume for a btrfs fs, so that this is implicit. I mean, it's great of rootflags= works, but requiring it sucks.

Conan-Kudo:
Someone else yesterday brought up that our default partitioning arrangement "looks weird" too, and I need to find out why...

poncovka:
@Conan-Kudo, can we merge this pull request or you want more time?

davide125:
Can we merge this now to unblock silverblue? We can look at doing set-default as a followup (as that will require changing a bunch more places)

Conan-Kudo:
@poncovka it's fine to merge now, some investigations will happen later and we can change as needed.

Conan-Kudo:
> https://twitter.com/pid_eins/status/1281493584081027072
> 
> > That bug looks weird. You shouldn't have to set rootflags= to pick the default subvolume, that's just fragile. Use `btrfs subvol set-default` to set the default subvolume for a btrfs fs, so that this is implicit. I mean, it's great of rootflags= works, but requiring it sucks.

So I found out why we do this: upstream grub does not support reading the default subvolume. SUSE has patches for this in their GRUB2 package, some of those we have (hello from 2017 me!) and some of those we don't. So this is unfortunately the correct fix until we deal with that, and I don't know when we can get around to dealing with that...

At some point, I'd like to get the RH and SUSE bootloader folks to come together on maybe getting us to BLS for btrfs boot to snapshot (which is the feature enabled by these patches that would help in this case), because this Jenga tower of patches is bonkers...

martinezjavier:
> So I found out why we do this: upstream grub does not support reading the default subvolume. SUSE has patches for this in their GRUB2 package, some of those we have (hello from 2017 me!) and some of those we don't. So this is unfortunately the correct fix until we deal with that, and I don't know when we can get around to dealing with that...

If you point me to those patches, I can include them in our package if that helps. Or you could propose a pull request for the https://github.com/rhboot/grub2/tree/fedora-33 branch.

> At some point, I'd like to get the RH and SUSE bootloader folks to come together on maybe getting us to BLS for btrfs boot to snapshot (which is the feature enabled by these patches that would help in this case), because this Jenga tower of patches is bonkers...

Nobody is happy with having so many patches on top of mainline. Specially since most distros are carrying more or less the same set. 

But things have been improving lately and upstream got much better at reviewing the posted patches. We managed to get ~40 patches merged for the upcoming 2.06 release. So we will get there, it just takes time.

poncovka:
Hi, I have updated the pull request, because the installation was failing. The current code seems to work fine.

Conan-Kudo:
> > So I found out why we do this: upstream grub does not support reading the default subvolume. SUSE has patches for this in their GRUB2 package, some of those we have (hello from 2017 me!) and some of those we don't. So this is unfortunately the correct fix until we deal with that, and I don't know when we can get around to dealing with that...
> 
> If you point me to those patches, I can include them in our package if that helps. Or you could propose a pull request for the https://github.com/rhboot/grub2/tree/fedora-33 branch.
> 
> > At some point, I'd like to get the RH and SUSE bootloader folks to come together on maybe getting us to BLS for btrfs boot to snapshot (which is the feature enabled by these patches that would help in this case), because this Jenga tower of patches is bonkers...
> 
> Nobody is happy with having so many patches on top of mainline. Specially since most distros are carrying more or less the same set.
> 
> But things have been improving lately and upstream got much better at reviewing the posted patches. We managed to get ~40 patches merged for the upcoming 2.06 release. So we will get there, it just takes time.

@martinezjavier The patches can be found here: https://build.opensuse.org/package/show/Base:System/grub2

You can see _only_ the Btrfs patches by putting "btrfs" into the Search filter on the file list.

Conan-Kudo:
> > So I found out why we do this: upstream grub does not support reading the default subvolume. SUSE has patches for this in their GRUB2 package, some of those we have (hello from 2017 me!) and some of those we don't. So this is unfortunately the correct fix until we deal with that, and I don't know when we can get around to dealing with that...
> 
> If you point me to those patches, I can include them in our package if that helps. Or you could propose a pull request for the https://github.com/rhboot/grub2/tree/fedora-33 branch.
> 
> > At some point, I'd like to get the RH and SUSE bootloader folks to come together on maybe getting us to BLS for btrfs boot to snapshot (which is the feature enabled by these patches that would help in this case), because this Jenga tower of patches is bonkers...
> 
> Nobody is happy with having so many patches on top of mainline. Specially since most distros are carrying more or less the same set.
> 
> But things have been improving lately and upstream got much better at reviewing the posted patches. We managed to get ~40 patches merged for the upcoming 2.06 release. So we will get there, it just takes time.

@martinezjavier The patches can be found here: https://build.opensuse.org/package/show/Base:System/grub2

You can see _only_ the Btrfs patches by putting "btrfs" into the Search filter on the file list. The patches were made by @jeffmahoney and @WenhuaChang primarily.

davide125:
Thanks @poncovka, the updated PR lgtm.

PR #2721, opened 2020-07-09
Temporarily ignore the new timezone kickstart command

poncovka:
Ignore the new 'timezone' kickstart command in the unit tests.

**Related:** https://github.com/rhinstaller/anaconda/pull/2704


jkonecny12:
I won't require that change.

PR #2722, opened 2020-07-09
Don't require fcoe-utils

poncovka:
Don't require to install the fcoe-utils package, so it can be excluded
from Live OS. The package should be only recommended.


jwrdegoede:
Thank you. Question, any idea when you will do the next 33.## tag / build and this will land in F33/rawhide ?

poncovka:
@jwrdegoede, we plan to release a new rawhide version tomorrow, so it might get to Wednesday's compose.

PR #2724, opened 2020-07-09
Add Anaconda version to saved kickstart

VladimirSlavik:
This is only version, not build/release from NEVRA.

Resolves: rhbz#1410595

Any advice on better places to put the string are welcome?


poncovka:
You could change the `__str__` method `AnacondaKSHandler`, but I prefer your solution.

VladimirSlavik:
Added a timestamp as asked originally at https://bugzilla.redhat.com/show_bug.cgi?id=1410595

bcl:
A problem with including the timestamp is that will make efforts to create reproducible builds fail. I know we don't officially support this, but there are some people who are trying :)

VladimirSlavik:
There is an implicit timestamp as the file's creation date, so perhaps that is not strictly needed in the file itself. I'll change the code.

Brian also changed pykickstart similarly, see https://github.com/pykickstart/pykickstart/pull/324 .

Do we want to add also blivet version to the output?

Conan-Kudo:
@VladimirSlavik If we could, that'd be useful for reporting partitioning bugs

VladimirSlavik:
Done, now with 0% timestamp but 50% blivet :-)

jkonecny12:
jenkins, test this please

VladimirSlavik:
Tests pass locally, except for the timezone changes which is expected and unrelated.

pyanaconda/installation.py
```diff
@@ -68,6 +70,10 @@ def _writeKS(ksdata):
 
     # Make it so only root can read - could have passwords
     with util.open_with_perm(path, "w", 0o600) as f:
+        f.write("# Generated by Anaconda {} with blivet {}\n".format(
```
poncovka, line 73: I guess it should be `Anaconda with Blivet` or `anaconda with blivet`.

VladimirSlavik: Ah, right. I was not even sure if Blivet has a name... but their docs are unambiguous on that :-)

pyanaconda/installation.py
```diff
@@ -68,6 +70,10 @@ def _writeKS(ksdata):
 
     # Make it so only root can read - could have passwords
     with util.open_with_perm(path, "w", 0o600) as f:
+        f.write("# Generated by Anaconda {} with Blivet {}\n".format(
```
poncovka, line 73: I think that the Blivet version should be generated by the Storage module. This code is going to live in the Boss module and it shouldn't have to import Blivet because of this string.

VladimirSlavik: I'll take a look at this, then...

VladimirSlavik:
Update: This now implements @poncovka suggestion above to add blivet version inside the storage module. Not sure if I got this right yet, but here we go!

pyanaconda/modules/storage/storage.py
```diff
@@ -177,6 +178,14 @@ def setup_kickstart(self, data):
         if self.applied_partitioning:
             self.applied_partitioning.setup_kickstart(data)
 
+    def generate_kickstart(self):
+        """Generate kickstart string representation of this module's data
+
+        Adds Blivet version to the output because most of the strings come from Blivet anyway.
+        """
+        return "# Generated using Blivet version {}\n".format(blivet_version) + \
+               super().generate_kickstart()
```
poncovka, line 187: Please, add this string to the format above to avoid the concatenation.

VladimirSlavik:
The rpmfluff problem is messing up tests and I can't tell what is actually working and what not - somehow all tests result in the same logs. The commits should be in the right order, and I plan to squash the test fixes with the changes. But without a way to verify, I am forced to wait.

pyanaconda/modules/storage/storage.py
VladimirSlavik: Done...

tests/nosetests/pyanaconda_tests/__init__.py
```diff
@@ -101,8 +117,9 @@ def check_kickstart_interface(test, interface, ks_in, ks_out=None, ks_valid=True
         return result
 
     # Generate a kickstart
-    ks_out = dedent(ks_out).strip()
-    test.assertEqual(ks_out, interface.GenerateKickstart().strip())
+    ks_out = clear_version_from_kickstart_string(dedent(ks_out)).strip()
```
poncovka, line 120: `ks_out` is the expected kickstart string, so it shouldn't contain the `# Generated` comment. Could you try to run the tests without modifying `ks_out`?

tests/nosetests/pyanaconda_tests/module_nvdimm_test.py
```diff
@@ -134,8 +135,8 @@ def _reconfigure(self, namespace, sector_size):
     def _check(self, expected_ks):
         """Check the generated kickstart."""
         self.assertEqual(
-            self.storage_module.generate_kickstart().strip(),
-            dedent(expected_ks).strip()
+            clear_version_from_kickstart_string(self.storage_module.generate_kickstart()).strip(),
+            clear_version_from_kickstart_string(dedent(expected_ks)).strip()
```
poncovka, line 139: Same here.

tests/nosetests/pyanaconda_tests/__init__.py
VladimirSlavik: Ha, you're right, I got confused by the reversed order at the other place I guess...

PR #2725, opened 2020-07-09
subscription: Prevent Subscription spoke being mandatory with DVD source set

M4rtinK:
Looks like there is a chance for the Subscription spoke to be set mandatory if payload initialization takes too long and the Subscription spoke gets initialized before payload is done.

The first commit make the Subscription spoke initialization to wait on the payload thread to finish.

The second commit enhances this by moving the waiting to a separate thread, preventing it from potentially
blocking the GUI startup for a noticeable amount of time.

~**NOTE: testing in progress.**~
Works as expected with simulated CDROM detection delay.

### DIFF
--- a/pyanaconda/ui/gui/spokes/subscription.py
+++ b/pyanaconda/ui/gui/spokes/subscription.py
@@ -25,7 +25,7 @@ from pyanaconda.threading import threadMgr, AnacondaThread
 from pyanaconda.core.i18n import _, CN_
 from pyanaconda.core.constants import SECRET_TYPE_HIDDEN, \
     SUBSCRIPTION_REQUEST_TYPE_USERNAME_PASSWORD, SUBSCRIPTION_REQUEST_TYPE_ORG_KEY, \
-    THREAD_SUBSCRIPTION, SOURCE_TYPES_OVERRIDEN_BY_CDN
+    THREAD_SUBSCRIPTION, THREAD_PAYLOAD, SOURCE_TYPES_OVERRIDEN_BY_CDN
 from pyanaconda.core.payload import ProxyString, ProxyStringError
 from pyanaconda.ui.lib.subscription import register_and_subscribe, \
     unregister, SubscriptionPhase
@@ -636,6 +636,10 @@ class SubscriptionSpoke(NormalSpoke):
 
         # wait for subscription thread to finish (if any)
         threadMgr.wait(THREAD_SUBSCRIPTION)
+        # also wait for the payload thread, which migh still be processing
+        # a CDROM source, to avoid the Subscription being mandatory by mistake
+        # due to CDN still being default at the time of evaulation
+        threadMgr.wait(THREAD_PAYLOAD)
 
         # update overall state
         self._update_registration_state()
--- a/pyanaconda/ui/gui/spokes/subscription.py
+++ b/pyanaconda/ui/gui/spokes/subscription.py
@@ -25,7 +25,8 @@ from pyanaconda.threading import threadMgr, AnacondaThread
 from pyanaconda.core.i18n import _, CN_
 from pyanaconda.core.constants import SECRET_TYPE_HIDDEN, \
     SUBSCRIPTION_REQUEST_TYPE_USERNAME_PASSWORD, SUBSCRIPTION_REQUEST_TYPE_ORG_KEY, \
-    THREAD_SUBSCRIPTION, THREAD_PAYLOAD, SOURCE_TYPES_OVERRIDEN_BY_CDN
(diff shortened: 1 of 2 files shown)


pyanaconda/ui/gui/spokes/subscription.py
```diff
@@ -634,8 +635,19 @@ def initialize(self):
...
+
+    def _initialize(self):
         # wait for subscription thread to finish (if any)
         threadMgr.wait(THREAD_SUBSCRIPTION)
+        # also wait for the payload thread, which migh still be processing
+        # a CDROM source, to avoid the Subscription being mandatory by mistake
+        # due to CDN still being default at the time of evaulation
+        threadMgr.wait(THREAD_PAYLOAD)
 
         # update overall state
```
poncovka, line 652: I think that the methods bellow that update UI should have the `async_action_wait` decorator.

pyanaconda/ui/gui/spokes/subscription.py
```diff
@@ -634,8 +635,19 @@ def initialize(self):
         # setup spoke state based on data from the Subscription DBus module
         self._update_spoke_state()
 
+        # start the rest of spoke initialization which might take some time
+        # (mainly due to waiting for various initialization threads to finish)
+        # in a separate thread
+        threadMgr.add(AnacondaThread(name=constants.THREAD_SUBSCRIPTION_SPOKE_INIT,
```
VladimirSlavik, line 641: Isn't `constants.` redundant here? You refer to all the other threads directly by constant name...

poncovka:
Jenkins, test this please.

pyanaconda/ui/gui/spokes/subscription.py
M4rtinK: Indeed, this was a copy paste typo & has been fixed.

pyanaconda/ui/gui/spokes/subscription.py
M4rtinK: Very good point! They are now running in a thread, which indeed could make GTK go boom.

pyanaconda/ui/gui/spokes/subscription.py
```diff
@@ -126,8 +127,16 @@ def __init__(self, *args):
...
     def ready(self):
-        """The subscription spoke is always ready."""
-        return True
+        """The subscription spoke is ready once its initialization thread finishes.
+
+        We do this to avoid the Subscription spoke being set mandatory in cases
+        where the current installation source is the CDN, but payload refres is still
+        running and it might change to CDROM later one. We achieve this by waiting
+        for tha payload refresh thread to finish in the Subscription spoke initialization
+        thread.
```
VladimirSlavik, line 136: refres -> refresh ; tha - > the

pyanaconda/ui/gui/spokes/subscription.py
```diff
@@ -634,8 +643,19 @@ def initialize(self):
...
+    def _initialize(self):
         # wait for subscription thread to finish (if any)
         threadMgr.wait(THREAD_SUBSCRIPTION)
+        # also wait for the payload thread, which migh still be processing
+        # a CDROM source, to avoid the Subscription being mandatory by mistake
+        # due to CDN still being default at the time of evaulation
+        threadMgr.wait(THREAD_PAYLOAD)
 
         # update overall state
         self._update_registration_state()
```
VladimirSlavik, line 661: Why do this twice?

pyanaconda/ui/gui/spokes/subscription.py
M4rtinK: Added the decorates, all seems fine after quick testing. (I have not hit a crash or freeze even before without the decorators, but that could have been luck as most of these GTK threading issues manifest as race conditions.

pyanaconda/ui/gui/spokes/subscription.py
M4rtinK: Fixed, thanks! :)

pyanaconda/ui/gui/spokes/subscription.py
M4rtinK: I'm not sure I understand, these are different methods:

```
        self._update_registration_state()
        self._update_subscription_state()
```

Or do you mean something else ?

VladimirSlavik: Oh, wow, they are :O never mind then...

M4rtinK:
jenkins, test this please

M4rtinK:
Port to Fedora: #2807

PR #2728, opened 2020-07-10
Add Fedora IoT product override

Conan-Kudo:
Fedora IoT was previously relying on the base Fedora product definition,
but it should not, since it should be using the classic storage configuration.

Fixes https://pagure.io/fedora-iot/issue/31

(cc: @AdamWill, @nullr0ute)

Signed-off-by: Neal Gompa <ngompa13@gmail.com>

### DIFF
--- a/data/product.d/fedora-iot.conf
+++ b/data/product.d/fedora-iot.conf
@@ -1,0 +1,11 @@
+# Anaconda configuration file for Fedora IoT.
+
+[Product]
+product_name = Fedora
+variant_name = IoT
+
+[Base Product]
+product_name = Fedora
+
+[Storage]
+default_scheme = LVM
--- a/data/product.d/fedora-iot.conf
+++ b/data/product.d/fedora-iot.conf
@@ -1,7 +1,7 @@
 # Anaconda configuration file for Fedora IoT.
 
 [Product]
-product_name = Fedora
+product_name = Fedora-IoT
 variant_name = IoT
 
 [Base Product]
--- a/tests/nosetests/pyanaconda_tests/product_test.py
+++ b/tests/nosetests/pyanaconda_tests/product_test.py
@@ -218,7 +218,7 @@ class ProductConfigurationTestCase(unittest.TestCase):
             WORKSTATION_PARTITIONING
         )
         self._check_default_product(
-            "Fedora", "IoT",
+            "Fedora-IoT", "IoT",
             ["fedora.conf", "fedora-iot.conf"],
             WORKSTATION_PARTITIONING
         )
--- a/tests/nosetests/pyanaconda_tests/product_test.py
+++ b/tests/nosetests/pyanaconda_tests/product_test.py
@@ -217,6 +217,11 @@ class ProductConfigurationTestCase(unittest.TestCase):
              "fedora-silverblue.conf"],
             WORKSTATION_PARTITIONING
         )
+        self._check_default_product(
+            "Fedora", "IoT",
+            ["fedora.conf", "fedora-iot.conf"],
+            WORKSTATION_PARTITIONING
+        )
 
     def rhel_products_test(self):
         self._check_default_product(


Conan-Kudo:
cc: @cmurf, @davide125

data/product.d/fedora-iot.conf
```diff
@@ -0,0 +1,11 @@
+# Anaconda configuration file for Fedora IoT.
```
poncovka, line 1: Could you please extend the method `fedora_products_test`, so the config file can be validated by our unit tests?

Conan-Kudo: Done.

poncovka:
Jenkins, it's ok to test.

data/product.d/fedora-iot.conf
```diff
@@ -0,0 +1,11 @@
...
+
+[Product]
+product_name = Fedora
```
poncovka, line 4: I have tested it and the product name is supposed to be `Fedora-IoT`. The variant name is correct.

Conan-Kudo: Something is not right here, because as far as I can tell, no other Edition installed through Anaconda is configured that way. Silverblue, Workstation, and old Fedora Atomic Host were just `Fedora` product with the Edition name as the variant. Why is this different?

poncovka: Anaconda reads the product and variant names from the `.buildstamp` file that is generated by `lorax`. I guess that `lorax` is run with arguments `--product "Fedora-IoT" --variant "IoT"`. I don't know where this configuration comes from.

Conan-Kudo: That sounds like a bug that needs to be fixed.

AdamWill: I think it's [Basically this](https://pagure.io/fedora-iot/issue/24) - I think lorax uses the productmd 'short name' here. We might actually want to allow some kind of wildcard specification here, as we may well wind up with IoT images with `product_name` "Fedora" at some point, and there are other cases where we have images for the same variant produced with different `product_name`s, e.g. the [post-release nightly Cloud composes](https://kojipkgs.fedoraproject.org/compose/cloud/) will likely use "Fedora-Cloud", if I'm understanding correctly here.

jkonecny12: I can't help myself but I see this as logically bad decision. It's still product Fedora not Fedora-IoT. For me it feels that even when they changed how they are making composes it should not touch the product name.

poncovka: I have replied to the pagure issue.

AdamWill: Well, there are practical reasons we need more-or-less that value to differ between different compose streams. We can't have two "Fedora" compose streams because it'd produce all kinds of ambiguities - the dist name is expected to be a unique identifier in, for instance, the compose ID (we'd often wind up with two or more entirely different composes with the same ID) and many deliverable file names (if two compose streams with some difference between them both build an image with the same subvariant, for instance, and we run both streams on the same day, they'd likely produce two different but identically-named image files).

At the Pungi and productmd levels this property is not referred to as the "product" name, FWIW. It's the "release" "short" name. It sort of 'becomes' the "product" name at the lorax level.

jkonecny12: Thanks a lot for explanation.

In that case there is a question if these shouldn't be separate? I mean if Pungi should not put something else to Lorax to avoid this problem.

AdamWill: I mean, that's something we can look at, but it's still the sort of thing I'd worry will unexpectedly break something else. We're fairly clear at least on how the current behaviour works. Can we not just teach anaconda wildcards for this or something, so we can do product configs with 'Fedora*'?

Conan-Kudo: @jkonecny12 @AdamWill How about I make a separate commit that just changes the product to `Fedora-IoT` that can be reverted as soon as this mess is fixed in the compose tooling?

Conan-Kudo: @poncovka @jkonecny12 @AdamWill I've added a commit to mangle the name until this is fixed.

tests/nosetests/pyanaconda_tests/product_test.py
```diff
@@ -217,6 +217,11 @@ def fedora_products_test(self):
              "fedora-silverblue.conf"],
             WORKSTATION_PARTITIONING
         )
+        self._check_default_product(
+            "Fedora", "IoT",
```
poncovka, line 221: Could you please also change the product name here? It should fix the failing test.

Conan-Kudo: Done.

PR #2729, opened 2020-07-10
Add LVM with inconsistent sector size disks to common bugs

jkonecny12:
This bug seems to be hard to understand for users even with the message. Explain it here too.

### DIFF
--- a/docs/common-bugs.rst
+++ b/docs/common-bugs.rst
@@ -187,6 +187,16 @@ Failing hardware
 :Solution: It looks like a hardware failure. Please, check your hardware.
 :Example: `rhbz#1685047 <https://bugzilla.redhat.com/show_bug.cgi?id=1685047>`_
 
+LVM on disks with inconsistent sector size
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+:Detection: The storage configuration fails with an error message mentioning "inconsistent sector
+    size".
+:Solution: LVM is now demanding that all disks have consistent sector size, otherwise they can't
+    be used together. Please adjust your disk selection to use only disks with the consistent
+    sector size.
+:Example: `rhbz#1754683 <https://bugzilla.redhat.com/show_bug.cgi?id=1754683>`_
+
 Unlocked LUKS
 ^^^^^^^^^^^^^
 


docs/common-bugs.rst
```diff
@@ -187,6 +187,14 @@ Failing hardware
...
 :Example: `rhbz#1685047 <https://bugzilla.redhat.com/show_bug.cgi?id=1685047>`_
 
+LVM on disks with inconsistent sector size
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+:Detection: Error with message mentioning "inconsistent sector size".
+:Solution: LVM is now demanding that all disks have consistent sector size, otherwise they can't
+    be used together. Please look on
+    `rhbz#1669751 <https://bugzilla.redhat.com/show_bug.cgi?id=1669751>`_ for more info.
+
```
poncovka, line 197: I would mention the bug https://bugzilla.redhat.com/show_bug.cgi?id=1754683 as an example. It is a public duplicate of the Blivet bug that was dealing with this problem.

docs/common-bugs.rst
```diff
@@ -187,6 +187,14 @@ Failing hardware
 :Solution: It looks like a hardware failure. Please, check your hardware.
 :Example: `rhbz#1685047 <https://bugzilla.redhat.com/show_bug.cgi?id=1685047>`_
 
+LVM on disks with inconsistent sector size
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+:Detection: Error with message mentioning "inconsistent sector size".
```
poncovka, line 193: Maybe it could be a sentence? `The storage configuration fails with an error message ...`

docs/common-bugs.rst
```diff
@@ -187,6 +187,14 @@ Failing hardware
...
+:Solution: LVM is now demanding that all disks have consistent sector size, otherwise they can't
+    be used together. Please look on
+    `rhbz#1669751 <https://bugzilla.redhat.com/show_bug.cgi?id=1669751>`_ for more info.
```
poncovka, line 196: The link doesn't provide useful recommendations. It is mentioned in the example below, I think that is enough. Instead of the link, I would recommend to change the disk selection or the sector sizes of selected disks.

docs/common-bugs.rst
jkonecny12: I wasn't able to find it. Thanks!

docs/common-bugs.rst
jkonecny12: It does not but it's the bug where they decided to go this way. I would like to have it here but I'll try to change the wording.

PR #2726, opened 2020-07-09
Only pass one initrd image to kexec

martinezjavier:
The kexec command line tool can only get a single initrd, but a boot entry
can have multiple ones. For example the tuned package adds a variable that
could be set to a second initrd image.

If that's the case, it will lead to the following error:

FileNotFoundError: [Errno 2] No such file or directory:
'/mnt/sysroot/boot/initramfs-4.18.0-223.el8.x86_64.img $tuned_initrd'

Resolves: rhbz#1855290

Signed-off-by: Javier Martinez Canillas <javierm@redhat.com>

PR #2731, opened 2020-07-10
Only pass one initrd image to kexec

jkonecny12:
The kexec command line tool can only get a single initrd, but a boot entry can have multiple ones. For example the tuned package adds a variable that could be set to a second initrd image.

If that's the case, it will lead to the following error:

FileNotFoundError: [Errno 2] No such file or directory: '/mnt/sysroot/boot/initramfs-4.18.0-223.el8.x86_64.img $tuned_initrd'

*Resolves: rhbz#1855290*

Backport of https://github.com/rhinstaller/anaconda/pull/2726 .

### DIFF
--- a/pyanaconda/kexec.py
+++ b/pyanaconda/kexec.py
@@ -72,6 +72,11 @@ def run_grubby(args=None):
     if boot_info_fields:
         raise GrubbyInfoError("Missing values: %s" % ", ".join(boot_info_fields))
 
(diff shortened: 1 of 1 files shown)


jkonecny12:
I've tested it and everything seems to be working correctly. However, this bug does not show-up on Fedora. I guess it's because of other behavior of tuned/grub...
So or so, it does not fail with this fix so there is no reason to not merge this.

PR #2732, opened 2020-07-13
subscription: Fixes for the the RHSM proxy (port to Fedora)

M4rtinK:
The Subscription Manager has it's own HTTP proxy and these two commits fix the configuration of it by Anaconda.

The first commit fixes use in GUI, the second fixes use in kickstart.

### DIFF
--- a/pyanaconda/ui/gui/spokes/subscription.py
+++ b/pyanaconda/ui/gui/spokes/subscription.py
@@ -26,7 +26,7 @@ from pyanaconda.core.i18n import _, CN_
 from pyanaconda.core.constants import SECRET_TYPE_HIDDEN, \
     SUBSCRIPTION_REQUEST_TYPE_USERNAME_PASSWORD, SUBSCRIPTION_REQUEST_TYPE_ORG_KEY, \
     THREAD_SUBSCRIPTION
-from pyanaconda.core.payload import ProxyString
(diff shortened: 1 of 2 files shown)


M4rtinK:
Original RHEL PRs:
#2690
#2711

pyanaconda/ui/gui/spokes/subscription.py
```diff
@@ -390,28 +390,35 @@ def on_http_proxy_checkbox_toggled(self, checkbox):
...
             hostname = proxy_obj.host
             if proxy_obj.port:
                 # the DBus API expects an integer
                 port = int(proxy_obj.port)
+        except ProxyStringError:
+            hostname = ""
+            port = -1  # not set == -1
+        # set the resulting values to the DBus structure
         self.subscription_request.server_proxy_hostname = hostname
         self.subscription_request.server_proxy_port = port
```
VladimirSlavik, line 413: What happens if no exception is raised, but the condition is not fulfilled?

M4rtinK: Hmm, that's a good point - while technically this is correct as ProxyString has a default port (3128), thes part of the code still looks confusing with the `port` variable. I'll try to do something about that.

poncovka:
Jenkins, test this please.

PR #2734, opened 2020-07-13
Fix software spoke message when source changes

M4rtinK:
With the Red Hat CDN now easily available the users will now
trigger the software spoke message that tells them to re-enter
the software spoke and confirm the software selection.

Unfortunately the current message - "Nothing selected" is
rather confusing. It does not tell what is not selected and
might actually be incorrect, as in some cases the previous
selection might transfer over when the installation source
changes.

So change the confusing message to "Please confirm software
selection" which is also more in line with our other spoke
status messages.

Thanks a lot to Jan Stodola for coming up with the text of
the new message. :)

Resolves: rhbz#1788458
(cherry picked from commit 6867d6eef6ce777c2059fe3ae278c858790c5cf7)

### DIFF
--- a/pyanaconda/ui/gui/spokes/software_selection.py
+++ b/pyanaconda/ui/gui/spokes/software_selection.py
@@ -313,12 +313,12 @@ class SoftwareSelectionSpoke(NormalSpoke):
                     return _("Invalid environment specified in kickstart")
             # we have no packages section in the kickstart and no environment has been set
             elif not self.environment:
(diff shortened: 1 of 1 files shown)


M4rtinK:
Original RHEL PR: #2686

poncovka:
The upstream fix should be different. See https://github.com/rhinstaller/anaconda/pull/2686#discussion_r445443630.

PR #2735, opened 2020-07-13
Run bash instead of sh in rescue mode

VladimirSlavik:
What it says on the label.

Not sure yet what else it affects, testing still todo.

### DIFF
--- a/pyanaconda/core/util.py
+++ b/pyanaconda/core/util.py
@@ -478,10 +478,10 @@ def execReadlines(command, argv, stdin=None, root='/', env_prune=None, filter_st
 ## Run a shell.
 def execConsole():
     try:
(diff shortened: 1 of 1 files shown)


VladimirSlavik:
Is it the same? The prompt says "sh" which is why I noticed this at all...

VladimirSlavik:
This needs both manual testing, and checking kickstart tests. Hence still a draft.

VladimirSlavik:
Manual testing ok, onwards to kickstart tests...

M4rtinK:
> Is it the same? The prompt says "sh" which is why I noticed this at all...

Looks like this can be used to check what is actually providing the shell running on a console:

`$SHELL --version`

VladimirSlavik:
Coverage Kickstart tests passed, and there isn't anything that would seem related, so that's that - undrafting.

PR #2737, opened 2020-07-15
Run the user instance of systemd

poncovka:
The service org.a11y.Bus runs the dbus-broker-launch in the user scope that
requires the user instance of systemd.

Resolves: rhbz#1832362

### DIFF
--- a/pyanaconda/core/configuration/system.py
+++ b/pyanaconda/core/configuration/system.py
@@ -69,6 +69,11 @@ class SystemSection(Section):
         return self._is_boot_iso or self._is_booted_os
 
     @property
+    def can_start_user_systemd(self):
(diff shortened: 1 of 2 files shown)


M4rtinK:
Thinking about it, I wonder if we couldn't use this when Anaconda is running in mock and "normal" Systemd & dbus broker are not available. Then we might be able to finally switch from DBus daemon to dbus broker. :)

poncovka:
Added a check of the installation environment.

jstodola:
Thanks for working on this, Vendy.
I tried the steps from https://bugzilla.redhat.com/show_bug.cgi?id=1832362#c6 - confirmed that the change fixes the problem described there.

VladimirSlavik (approved):
Even better now.

PR #2738, opened 2020-07-16
Add NTS support

mlichvar:
These patches add support for the client Network Time Security (NTS) option in chrony.

### DIFF
--- a/pyanaconda/ntp.py
+++ b/pyanaconda/ntp.py
@@ -97,18 +97,24 @@ def get_ntp_servers_summary(servers, states):
     return summary
 
 
-def ntp_server_working(server_hostname):
+def ntp_server_working(server_hostname, nts_enabled):
     """Tries to do an NTP request to the server (timeout may take some time).
+
+    If NTS is enabled, try making a TCP connection to the NTS-KE port instead.
 
     :param server_hostname: a host name or an IP address of an NTP server
     :type server_hostname: string
     :return: True if the given server is reachable and working, False otherwise
     :rtype: bool
     """
-    client = ntplib.NTPClient()
-
     try:
-        client.request(server_hostname)
+        # ntplib doesn't support NTS
+        if nts_enabled:
+            s = socket.create_connection((server_hostname, 4460), 2)
+            s.close()
+        else:
+            client = ntplib.NTPClient()
+            client.request(server_hostname)
     except ntplib.NTPException:
         return False
     # address related error
@@ -324,8 +330,9 @@ class NTPServerStatusCache(object):
 
         :param TimeSourceData server: an NTP server
         """
-        # Get a hostname.
+        # Get a hostname and NTS option.
         hostname = server.hostname
+        nts_enabled = "nts" in server.options
 
         # Reset the current status.
         self._set_status(hostname, NTP_SERVER_QUERY)
@@ -334,7 +341,7 @@ class NTPServerStatusCache(object):
         threadMgr.add(AnacondaThread(
             prefix=THREAD_NTP_SERVER_CHECK,
             target=self._check_status,
-            args=(hostname, ))
+            args=(hostname, nts_enabled))
         )
 
     def _set_status(self, hostname, status):
@@ -345,13 +352,13 @@ class NTPServerStatusCache(object):
         """
         self._cache[hostname] = status
 
-    def _check_status(self, hostname):
+    def _check_status(self, hostname, nts_enabled):
         """Check if an NTP server appears to be working.
 
         :param str hostname: a hostname of an NTP server
         """
         log.debug("Checking NTP server %s", hostname)
-        result = ntp_server_working(hostname)
+        result = ntp_server_working(hostname, nts_enabled)
 
         if result:
             log.debug("NTP server %s appears to be working.", hostname)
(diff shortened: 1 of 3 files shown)


mlichvar:
This doesn't currently work well with the "one_time_sync" that's performed after enabling the NTP switch. As ntplib doesn't support NTS, the function would need to call `chronyd -q 'server {server} nts iburst maxsamples 1' to make a quick step using an NTS-enabled server.

Maybe the one_time_sync code could be dropped and just rely on the restarted chronyd service to make the step? It should take about 6-8 seconds. Is that too long?

poncovka:
Jenkins, it's ok to test.

pyanaconda/ntp.py
```diff
@@ -106,7 +110,12 @@ def ntp_server_working(server_hostname):
     client = ntplib.NTPClient()
```
M4rtinK, line 110: I think we might move this in the else branch as the `client` variable is now used only there ?

pyanaconda/ntp.py
```diff
@@ -143,7 +152,16 @@ def get_servers_from_config(conf_file_path=NTP_CONFIG_FILE):
...
-                server.options = ["iburst"]
+                server.options = []
+
+                words = match.group(3).lower().split()
+                for i in range(len(words)):
+                    if words[i] in SRV_NOARG_OPTIONS:
+                        server.options.append(words[i])
+                    elif words[i] in SRV_ARG_OPTIONS and i + 1 < len(words):
+                        server.options.append(' '.join(words[i:i+2]))
+                    # ignore unknown options
```
M4rtinK, line 163: Maybe we can log those unknown options ? That would also make the comment look better and not a bit dangling like this.

pyanaconda/ntp.py
```diff
@@ -106,7 +110,12 @@ def ntp_server_working(server_hostname):
     client = ntplib.NTPClient()
 
     try:
-        client.request(server_hostname)
+        # ntplib doesn't support NTS
+        if "nts" in options:
+            s = socket.create_connection((server_hostname, 4460), 2)
```
poncovka, line 115: How can I test it? I wasn't able to find a server that would be marked as working.

poncovka:
Here are screenshots of the UI changes:

Screenshot_vpn_test_03_2020-07-16_16:42:48-nts1

Screenshot_vpn_test_03_2020-07-16_16:43:08-nts2

pyanaconda/ntp.py
```diff
@@ -316,7 +334,7 @@ def check_status(self, server):
         threadMgr.add(AnacondaThread(
             prefix=THREAD_NTP_SERVER_CHECK,
             target=self._check_status,
-            args=(hostname, ))
+            args=(hostname, server.options))
```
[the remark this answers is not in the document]
mlichvar: Ok, thanks.

pyanaconda/ntp.py
mlichvar: Makes sense. Thanks.

pyanaconda/ntp.py
mlichvar: You can test it with `nts-test.strangled.net` and `ntp1.glypnod.com` for example. Cloudflare and others will hopefully move to the new port soon. Please note that we will also need to update the chrony package to use this port by default.

poncovka:
> This doesn't currently work very well with the "one_time_sync" that's performed after enabling the NTP switch. An NTS-enabled server may have the NTP server running on different host or port (that's negotiated in NTS-KE), so the ntplib request may not reach the NTP server, and even if it does, the response won't be protected by NTS. I'm not sure how much that matters, as the installer seems to sync to a pool.ntp.org server before the servers are configured in the dialog.
> 
> For a proper NTS-enabled sync, the one_time_sync function would need to call something like `chronyd -q 'server {server} nts iburst maxsamples 1'`.
> 
> But maybe it could be dropped and just rely on the restarted chronyd service to make the step? It should take about 6-8 seconds. Is that too long?

It looks like the logic was introduced in the commit https://github.com/rhinstaller/anaconda/commit/edad0eb1504bde987d2bda7d8ca225266629b81d. It is from 2012, so I don't really know how much time it saved back then. I guess that 6-8 seconds could be fine. @jkonecny12, what do you think?

jkonecny12:
> This doesn't currently work very well with the "one_time_sync" that's performed after enabling the NTP switch. An NTS-enabled server may have the NTP server running on different host or port (that's negotiated in NTS-KE), so the ntplib request may not reach the NTP server, and even if it does, the response won't be protected by NTS. I'm not sure how much that matters, as the installer seems to sync to a pool.ntp.org server before the servers are configured in the dialog.
> 
> For a proper NTS-enabled sync, the one_time_sync function would need to call something like `chronyd -q 'server {server} nts iburst maxsamples 1'`.
> 
> But maybe it could be dropped and just rely on the restarted chronyd service to make the step? It should take about 6-8 seconds. Is that too long?

I just wonder, I remember that chrony when started on system with bad clocks it starts slowly convergent to the correct time but not immediately. Is that still true? In that case will the service restart work at all?

mlichvar:
It depends on the configuration. The default configuration always allowed step on the service (re)start, but it takes at least 6 seconds for that to happen, which I suspect may be confusing when the clock is obviously wrong and the user is staring at the large time displayed in the timezone dialog?

mlichvar:
Thanks for the review.

poncovka:
Tested. The changes for the `one_time_sync` function can be done in another pull request, so I am going to merge this one. I like the idea of dropping the function and relying on the chronyd service.

PR #2739, opened 2020-07-16
Fix root password handling

VladimirSlavik:
Now correct:
- Spoke locking / sensitivity
- Fields get grayed out
- Both with and without Kickstart

Resolves: rhbz#1851220

CC: @jstodola

### DIFF
--- a/pyanaconda/ui/gui/spokes/root_password.glade
+++ b/pyanaconda/ui/gui/spokes/root_password.glade
@@ -180,35 +180,10 @@
                   </packing>
                 </child>
                 <child>
-                  <object class="GtkCheckButton" id="lock">
-                    <property name="label" translatable="yes">Lock root account</property>
-                    <property name="visible">True</property>
-                    <property name="can_focus">True</property>
-                    <property name="receives_default">False</property>
-                    <property name="halign">start</property>
-                    <property name="draw_indicator">True</property>
-                    <signal name="clicked" handler="on_lock_clicked" swapped="no"/>
-                  </object>
-                  <packing>
-                    <property name="expand">False</property>
-                    <property name="fill">True</property>
-                    <property name="position">1</property>
-                  </packing>
+                  <placeholder/>
                 </child>
                 <child>
-                  <object class="GtkCheckButton" id="root_password_ssh_login_override">
-                    <property name="label" translatable="yes">Allow root SSH login with password</property>
-                    <property name="visible">True</property>
-                    <property name="can_focus">True</property>
-                    <property name="receives_default">False</property>
-                    <property name="halign">start</property>
-                    <property name="draw_indicator">True</property>
-                  </object>
-                  <packing>
-                    <property name="expand">False</property>
-                    <property name="fill">True</property>
-                    <property name="position">2</property>
-                  </packing>
+                  <placeholder/>
                 </child>
                 <child>
                   <placeholder/>
--- a/pyanaconda/ui/gui/spokes/root_password.py
+++ b/pyanaconda/ui/gui/spokes/root_password.py
@@ -1,6 +1,6 @@
 # root password spoke class
 #
-# Copyright (C) 2012-2014 Red Hat, Inc.
+# Copyright (C) 2012-2020 Red Hat, Inc.
 #
 # This copyrighted material is made available to anyone wishing to use,
 # modify, copy, or redistribute it subject to the terms and conditions of
@@ -49,7 +49,6 @@ class PasswordSpoke(FirstbootSpokeMixIn, NormalSpoke, GUISpokeInputCheckHandler)
     uiFile = "spokes/root_password.glade"
     help_id = "RootPasswordSpoke"
 
-
     category = UserSettingsCategory
 
     icon = "dialog-password-symbolic"
@@ -60,8 +59,6 @@ class PasswordSpoke(FirstbootSpokeMixIn, NormalSpoke, GUISpokeInputCheckHandler)
         GUISpokeInputCheckHandler.__init__(self)
         self._users_module = USERS.get_proxy()
         self._services_module = SERVICES.get_proxy()
-        self._refresh_running = False
-        self._manually_locked = False
 
     def initialize(self):
         NormalSpoke.initialize(self)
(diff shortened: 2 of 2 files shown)


VladimirSlavik:
For updates image, see https://bugzilla.redhat.com/show_bug.cgi?id=1851220#c9

poncovka:
It is not very clear what the fix is supposed to do based on the commit message.

poncovka:
@M4rtinK, could you look at this pull request, please?

pyanaconda/ui/gui/spokes/root_password.py
```diff
@@ -118,9 +119,11 @@ def initialize(self):
         self.checker.add_check(self._validity_check)
         self.checker.add_check(self._ascii_check)
 
-        # the password entries should be sensitive on first entry to the spoke
-        self.password_entry.set_sensitive(True)
-        self.password_confirmation_entry.set_sensitive(True)
+        # the password entries should be sensitive only if root is not locked from kickstart
+        allow_changes = self._users_module.CanChangeRootPassword or self.checker.policy.changesok
+        # or not self._users_module.IsRootAccountLocked
```
M4rtinK, line 124: I think this is a leftover commented out ?

pyanaconda/ui/gui/spokes/root_password.py
```diff
@@ -118,9 +119,11 @@ def initialize(self):
         self.checker.add_check(self._validity_check)
         self.checker.add_check(self._ascii_check)
 
-        # the password entries should be sensitive on first entry to the spoke
-        self.password_entry.set_sensitive(True)
-        self.password_confirmation_entry.set_sensitive(True)
+        # the password entries should be sensitive only if root is not locked from kickstart
+        allow_changes = self._users_module.CanChangeRootPassword or self.checker.policy.changesok
```
M4rtinK, line 123: I think this should be "and" instead of "or" - eq. to change the root password, both the module and the policy need to allow that. Or was it done differently in the past ?

pyanaconda/ui/gui/spokes/root_password.py
```diff
@@ -339,8 +346,5 @@ def on_lock_clicked(self, lock):
                 # if the checkbox has not yet been manipulated by the user
                 # we can ignore this run and not adjust the fields
                 return True
-        self.password_entry.set_sensitive(not lock.get_active())
```
M4rtinK, line 342: Is it necessary to remove this code ? I though we disabled the root password locking checkbox on RHEL so I would assume this callback to be never called (or only during refresh & that should be interrupted by the check).

pyanaconda/ui/gui/spokes/root_password.py
VladimirSlavik: Yes, sorry, fixed.

pyanaconda/ui/gui/spokes/root_password.py
VladimirSlavik: The removal of `grab_focus` below was overzealous, but the `set_sensitive` must go for this to work.

pyanaconda/ui/gui/spokes/root_password.py
VladimirSlavik: tl;dr OR is right here.

The names are kind of misleading. Either can allow entering the spoke. `CanChangeRootPassword` is actually kickstart's `seen`, just inverted, so that one means "yes you can change it, changes policy does not apply".

VladimirSlavik:
@M4rtinK @jstodola the only difference now is that in 8.2, the spoke says that root password is not set, while in 8.3 with this change it says root account is disabled. Is that a problem?

Personally I see the new behavior as more consistent, but it *is* technically different.

poncovka:
> @M4rtinK @jstodola the only difference now is that in 8.2, the spoke says that root password is not set, while in 8.3 with this change it says root account is disabled. Is that a problem?
> 
> Personally I see the new behavior as more consistent, but it _is_ technically different.

Based on the code, we show `Root account is disabled`, when the root account is locked, and `Root password is not set`, when it is not locked and the password is not set. I think we changed the behaviour when we introduced the checkbox for locking, but @M4rtinK should remember that better.

It is not clear what are the expectations here. Personally, I would prefer to make the checkbox visible and keep the upstream logic, because we use it for a while now, it is stable and well tested.

If it is expected to keep everything the same as in RHEL 8.2, I suggest to revert all changes in the password spoke and update only the necessary lines. It will be safer then trying to fix the current solution.

VladimirSlavik:
Just a dirty push to keep it in the open and keep the previous version too. It's a bit more tricky, even if I adapt the old code for new api, the messages are still changed...

pyanaconda/ui/gui/spokes/root_password.py
```diff
@@ -118,16 +115,10 @@ def initialize(self):
...
-
-        # Set placeholders if the password has been set outside of the Anaconda
-        # GUI we either don't really know anything about it if it's crypted
-        # and still would not really want to expose it if its set in plaintext,
-        # and thus can'treally show it in the UI in any meaningful way.
-        if self._users_module.IsRootPasswordSet:
-            password_set_message = _("Root password has been set.")
+        # set placeholders if the password has been kickstarted as we likely don't know
+        # nothing about it and can't really show it in the UI in any meaningful way
+        password_set_message = _("The password was set by kickstart.")
```
poncovka, line 120: I think you can use the new string.

pyanaconda/ui/gui/spokes/root_password.glade
```diff
@@ -182,12 +182,11 @@
                 <child>
                   <object class="GtkCheckButton" id="lock">
                     <property name="label" translatable="yes">Lock root account</property>
```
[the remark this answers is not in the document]
VladimirSlavik: Done...

pyanaconda/ui/gui/spokes/root_password.py
VladimirSlavik: Done.

VladimirSlavik:
Update: review applied, commits squashed, dropped the previous version.

PR #2743, opened 2020-07-20
Log the information about the original exception

poncovka:
When we re-raise an exception from None, log the information about the original
exception. Otherwise, we might not be able to find the cause of the error.


poncovka:
Jenkins, test this please.

PR #2744, opened 2020-07-20
Revert "Do not build Anaconda on i686"

M4rtinK:
This reverts commit 63f85f6226bcdca800129d2ed311c37736a3b6d7.

Turns out that anaconda-widgets.i686 has been shipped to customers
so there is a slight chance that dropping the i686 might cause issues.

So revert the commit that disables the i686 Anaconda build.

Resolves: rhbz#1857180

### DIFF
--- a/anaconda.spec.in
+++ b/anaconda.spec.in
@@ -148,11 +148,6 @@ Obsoletes: anaconda-runtime < %{version}-%{release}
 Provides: anaconda-runtime = %{version}-%{release}
 Obsoletes: booty <= 0.107-1
 
-# native i686 installations are not supported on RHEL8 and Anaconda
-# is not a i686 compatibility library, so building it for i686 does not
-# make sense
-ExcludeArch: i686
-
 %description core
 The anaconda-core package contains the program which was used to install your
 system.


poncovka:
From https://github.com/rhinstaller/anaconda/pull/2639:
 
> This also fixes an issue with Anaconda depending on kexec-tools,
> which has dropped i686 build. This other packages from being
> built due to transitive dependency on i686 kexec-tools package that
> no longer exists if those packages have BuildRequires: on anaconda.

What is the alternative solution of the original bug?

M4rtinK:
> From #2639:
> 
> > This also fixes an issue with Anaconda depending on kexec-tools,
> > which has dropped i686 build. This other packages from being
> > built due to transitive dependency on i686 kexec-tools package that
> > no longer exists if those packages have BuildRequires: on anaconda.
> 
> What is the alternative solution of the original bug?

In Initial Setup this was solved by dropping the i686 of Initial Setup and as that was apparently not shipped to customers, all seems to be fine. And all other packages depending on Anaconda seem to be fine as well, as they are either noarch or do not have a i686 build.

jkonecny12:
jenkins, test this please

PR #2746, opened 2020-07-21
Allow to detect devices with the iso9660 file system as optical media

poncovka:
Test that the DBus method FindOpticalMedia identifies devices with the iso9660 file
system as optical media, so it is able to find NVDIMM devices with iso9660.

The DBus method GetDevicesToIgnore of the NVDIMM module shouldn't return NVDIMM
devices with the iso9660 file system. They can be used as an installation source.

Protect all devices with the iso9660 file system. It will protect, for example, NVDIMM
devices with the iso9660 file system that can be used only as an installation source
anyway.

Related: rhbz#1856264


jkonecny12:
jenkins, test this please

jkonecny12:
jenkins, test this please

PR #2748, opened 2020-07-21
Don't return negative container free space

poncovka:
The DBus method GetContainerFreeSpace is supposed to return an unsigned integer,
so adjust the return value in case that Blivet returned a negative value.

Resolves: rhbz#1853071


poncovka:
Reported to upstream: https://bugzilla.redhat.com/show_bug.cgi?id=1868623

poncovka:
Fixed in upstream: https://github.com/storaged-project/blivet/pull/890

PR #2747, opened 2020-07-21
Fix stage2 as default sources

jkonecny12:
We should prioritize stage2 device as the default source. This is especially needed for DVD ISO because it is booting with inst.stage2 instead and we should use the DVD as the source for the default CDROM. The situation is even worse thanks to the fact that DVD ISOs are using inst.stage2=hd:...

Find stage2 device and test this device first during the auto-discover feature of CDRom source.

This PR depends on https://github.com/rhinstaller/anaconda/pull/2746.

*Resolves: rhbz#1856264*

PR #2749, opened 2020-07-22
Fix stage2 as default sources

jkonecny12:
We should prioritize stage2 device as the default source. This is especially needed for DVD ISO because it is booting with inst.stage2 instead and we should use the DVD as the source for the default CDROM. The situation is even worse thanks to the fact that DVD ISOs are using inst.stage2=hd:...

Find stage2 device and test this device first during the auto-discover feature of CDRom source.

Backport of https://github.com/rhinstaller/anaconda/pull/2747.

### DIFF
--- a/pyanaconda/modules/payloads/source/cdrom/cdrom.py
+++ b/pyanaconda/modules/payloads/source/cdrom/cdrom.py
@@ -32,7 +32,12 @@ log = get_module_logger(__name__)
 
 
 class CdromSourceModule(PayloadSourceBase, MountingSourceMixin, RPMSourceMixin):
-    """The CD-ROM source payload module."""
+    """The CD-ROM source payload module.
+
+    This source will try to automatically detect installation source. First it tries to look only
+    stage2 device used to boot the environment then it will use first valid iso9660 media with a
+    valid structure.
+    """
 
     def __init__(self):
         super().__init__()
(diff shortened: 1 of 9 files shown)


jkonecny12:
Unfortunately, this does not really work. I tested this on Fedora and the problem is that the pmem device is not mountable because it's already mounted. What is weird that the original mount is to /run/install/repo but as RO so it should not be the problem?

I have to dig deeper into this and why it's a problem on Fedora and not on RHEL...

jkonecny12:
OK, might be problem of old blivet. When I downloaded new Fedora server DVD it seems to be working correctly.

Interesting thing is that blivet still returns bad device but this device has the correct data... 

```
Aug 20 17:16:07 fedora org.fedoraproject.Anaconda.Modules.Storage[1887]: DEBUG:blivet:resolved 'LABEL=Fedora-S-dvd-x86_64-33' to 'pmem0' (nvdimm)
```

```
[anaconda root@fedora ~]# ls -lh /dev/disk/by-label/Fedora-S-dvd-x86_64-33
lrwxrwxrwx. 1 root root 13 Aug 20 17:15 /dev/disk/by-label/Fedora-S-dvd-x86_64-33 -> ../../pmem0p1
```

```
Aug 20 17:16:07 fedora org.fedoraproject.Anaconda.Modules.Payloads[1888]: INFO:program:Running... mount -t iso9660 -o ro /dev/pmem0 /run/install/sources/mount-0000-cdrom
Aug 20 17:16:07 fedora org.fedoraproject.Anaconda.Modules.Payloads[1888]: DEBUG:program:Return code: 0
```

@vojtechtrefny is this correct behavior or it should not work like this?

VladimirSlavik:
Can we reject the device on anaconda side when it has children, and search with that one excluded, or something similar?

jkonecny12:
jenkins, test this please

jkonecny12:
I want to test this again soon because it seems that the problem was just with the old ISO.

jkonecny12:
jenkins, test this please

jkonecny12:
jenkins, test this please

jkonecny12:
Working correctly on newer Server DVD.

PR #2756, opened 2020-07-29
Apply onboot policy even when network was configured in UI.

rvykydal:
Resolves: rhbz#1856632

Brings back behaviour from RHEL 7 and RHEL 8.2

### DIFF
--- a/pyanaconda/modules/network/network.py
+++ b/pyanaconda/modules/network/network.py
@@ -366,9 +366,6 @@ class NetworkService(KickstartService):
         # Not if any network device was configured via kickstart.
         if self._original_network_data:
             return False
-        # Not if any network device was configured in UI.
-        if self._use_device_configurations:
-            return False
         # Not if there is no configuration to apply the policy to
         if not self._device_configurations or not self._device_configurations.get_all():
             return False


rvykydal:
> LGTM
> If this behavior is also in Fedora than maybe we should use configuration files to solve this in future.

Yes, I was thinking about it but it would require reworking of policy setting we already have in configuration files, and here the situation is still quite unclean. I'll think about it more when porting the patch to Fedora.

jkonecny12:
> > LGTM
> > If this behavior is also in Fedora than maybe we should use configuration files to solve this in future.
> 
> Yes, I was thinking about it but it would require reworking of policy setting we already have in configuration files, and here the situation is still quite unclean. I'll think about it more when porting the patch to Fedora.

In that case please create an issue for the future development. We can look on this later.

rvykydal:
> > > LGTM
> > > If this behavior is also in Fedora than maybe we should use configuration files to solve this in future.
> > 
> > 
> > Yes, I was thinking about it but it would require reworking of policy setting we already have in configuration files, and here the situation is still quite unclean. I'll think about it more when porting the patch to Fedora.
> 
> In that case please create an issue for the future development. We can look on this later.

INSTALLER-2154

rvykydal:
PR with port to Fedora: https://github.com/rhinstaller/anaconda/pull/2833

PR #2757, opened 2020-07-29
Rhel 8 always recreate unified repos

jkonecny12:
This will solve problem that additional repositories pointing to an invalid path. This is happening because we moved source mounting to a different folders for each source. By this patch we will always reload new treeinfo repository if user change the base source repository.

Also this changed behavior of Source spoke. The Source spoke does not disable repository on base repo switch but instead it will remove this repository and the repository will be reloaded when user re-enter the spoke. It's not ideal solution but doable with avoiding great amount of changes everywhere.

Resolves: rhbz#1851207

### DIFF
--- a/pyanaconda/payload/dnf/payload.py
+++ b/pyanaconda/payload/dnf/payload.py
@@ -1931,7 +1931,11 @@ class DNFPayload(Payload):
                     repo.treeinfo_origin = True
                     log.debug("Adding new treeinfo repository: %s enabled: %s",
                               repo_md.name, repo_enabled)
-                    self.add_repo(repo)
+
+                    if repo_enabled:
+                        self.add_repo(repo)
+                    else:
+                        self.add_disabled_repo(repo)
 
     def _cleanup_old_treeinfo_repositories(self):
         """Remove all old treeinfo repositories before loading new ones.
--- a/pyanaconda/payload/dnf/payload.py
+++ b/pyanaconda/payload/dnf/payload.py
@@ -1531,6 +1531,8 @@ class DNFPayload(Payload):
         log.info("Configuring the base repo")
         self.reset()
 
+        self._cleanup_old_treeinfo_repositories()
+
         # Find the source and its type.
         source_proxy = self.get_source_proxy()
         source_type = source_proxy.Type
@@ -1583,11 +1585,9 @@ class DNFPayload(Payload):
                 self._refresh_install_tree(data)
                 self._base.conf.releasever = self._get_release_version(install_tree_url)
                 base_repo_url = self._get_base_repo_location(install_tree_url)
-
-                if self.first_payload_reset:
-                    self._add_treeinfo_repositories(install_tree_url, base_repo_url)
-
                 log.debug("releasever from %s is %s", base_repo_url, self._base.conf.releasever)
+
+                self._load_treeinfo_repositories(base_repo_url)
             except configparser.MissingSectionHeaderError as e:
                 log.error("couldn't set releasever from base repo (%s): %s", source_type, e)
 
@@ -1896,12 +1896,11 @@ class DNFPayload(Payload):
         log.debug("No base repository found in treeinfo file. Using installation tree root.")
         return install_tree_url
 
-    def _add_treeinfo_repositories(self, install_tree_url, base_repo_url=None):
-        """Add all repositories from treeinfo file which are not already loaded.
+    def _load_treeinfo_repositories(self, base_repo_url):
+        """Load new repositories from treeinfo file.
 
-        :param install_tree_url: Url to the installation tree root.
-        :param base_repo_url: Base repository url. This is not saved anywhere when the function
-        is called. It will be add to the existing urls if not None.
+        :param base_repo_url: base repository url. This is not saved anywhere when the function
+                              is called. It will be add to the existing urls if not None.
         """
         if self._install_tree_metadata:
             existing_urls = []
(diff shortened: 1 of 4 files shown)


pyanaconda/payload/dnf/payload.py
```diff
@@ -275,125 +275,6 @@ def _replace_vars(self, url):
 
         return None
 
-    def _add_repo(self, ksrepo):
```
poncovka, line 278: Is this refactorization necessary?

pyanaconda/payload/dnf/payload.py
```diff
@@ -1928,9 +1938,19 @@ def _add_treeinfo_repositories(self, install_tree_url, base_repo_url=None):
                     repo = RepoData(name=repo_md.name, baseurl=repo_md.path,
                                     install=False, enabled=repo_enabled)
                     repo.treeinfo_origin = True
+                    log.debug("Adding new treeinfo repository %s", repo_md.name)
                     self.add_repo(repo)
 
-        return install_tree_url
+    def _cleanup_old_treeinfo_repositories(self):
```
poncovka, line 1954: I think we should do this after reset. The repositories won't be deleted if the source is changed to for example to CDN.

pyanaconda/payload/dnf/payload.py
```diff
@@ -1588,12 +1588,10 @@ def update_base_repo(self, fallback=True, checkmount=True):
             try:
                 self._refresh_install_tree(data)
                 self._base.conf.releasever = self._get_release_version(install_tree_url)
-                base_repo_url = self._get_base_repo_location(install_tree_url)
-
-                if self.first_payload_reset:
-                    self._reload_treeinfo_repositories(base_repo_url)
-
                 log.debug("releasever from %s is %s", base_repo_url, self._base.conf.releasever)
```
poncovka, line 1591: This logging is broken.

pyanaconda/payload/dnf/payload.py
```diff
@@ -1932,24 +1934,43 @@ def _load_treeinfo_repositories(self, base_repo_url):
...
     def _cleanup_old_treeinfo_repositories(self):
         """Remove all old treeinfo repositories before loading new ones.
 
         Find all repositories added from treeinfo file and remove them. After this step new
         repositories will be loaded from the new link.
+
+        :return: list of repositories which were disabled before removal
+        :rtype: [str]
         """
+        disabled_repo_list = []
```
poncovka, line 1963: It is not clear that it is a list of repo names.

pyanaconda/payload/dnf/payload.py
```diff
@@ -1908,14 +1908,16 @@ def _reload_treeinfo_repositories(self, base_repo_url=None):
         :param base_repo_url: base repository url. This is not saved anywhere when the function
                               is called. It will be add to the existing urls if not None.
         """
-        self._cleanup_old_treeinfo_repositories()
-        self._load_treeinfo_repositories(base_repo_url)
+        disabled_repo_list = self._cleanup_old_treeinfo_repositories()
+        self._load_treeinfo_repositories(base_repo_url, disabled_repo_list)
 
-    def _load_treeinfo_repositories(self, base_repo_url):
+    def _load_treeinfo_repositories(self, base_repo_url, disable_repositories):
```
poncovka, line 1914: Maybe `disabled_repositories`? But I guess `disabled_names` would be less confusing.

poncovka:
Also, you should probably remove the workaround in the Subscription spoke:
https://github.com/rhinstaller/anaconda/blob/rhel-8/pyanaconda/ui/lib/subscription.py#L280

pyanaconda/payload/dnf/payload.py
jkonecny12: It's not disabled but a list you want to disable.

pyanaconda/payload/dnf/payload.py
jkonecny12: It's not necessary but also it's not really dangerous. It's just easier to find the pieces in the code. However, if you insist on this I can definitely drop the first commit.

pyanaconda/payload/dnf/payload.py
jkonecny12: I'm not sure I understand you. When we do this after reset then we have to do the reset again because these repositories will be added back.

jkonecny12: Fixed.

pyanaconda/payload/dnf/payload.py
jkonecny12: Fixed.

pyanaconda/payload/dnf/payload.py
jkonecny12: Good catch. Fixed.

pyanaconda/payload/dnf/payload.py
jkonecny12: I've tried better name.

pyanaconda/payload/dnf/payload.py
jkonecny12: Tried better name.

jkonecny12:
**UPDATED:**
* Implemented notes from @poncovka, thanks a lot for them!
* Squashed first two commits into one.

pyanaconda/payload/dnf/payload.py
poncovka: I think it is the same. Could you at least fix the docstring, please?

jkonecny12: Ohh, sorry I missed that one.

jkonecny12:
**UPDATED:**
* Fixed missed rename of variable.
* Unsquash commits accidentally squashed in the last version.

pyanaconda/payload/dnf/payload.py
```diff
@@ -1896,12 +1896,13 @@ def _get_base_repo_location(self, install_tree_url):
         log.debug("No base repository found in treeinfo file. Using installation tree root.")
         return install_tree_url
 
-    def _add_treeinfo_repositories(self, install_tree_url, base_repo_url=None):
-        """Add all repositories from treeinfo file which are not already loaded.
+    def _load_treeinfo_repositories(self, base_repo_url, repo_names_to_disable):
+        """Load new repositories form treeinfo file.
```
poncovka, line 1900: `from`

jkonecny12:
**UPDATED:**
* Fixed typo found by @poncovka . Thanks!

PR #2758, opened 2020-07-29
Implement get_source_proxy() in payload base class

M4rtinK:
Even though some payloads (such as for example the OSTree payload)
do not have a source proxy, it's better to implement this method
universally. We can just return None for the payloads that don't
have a source.
    
As the method now can return None, adjust the code that can be using
it with a payload that does not have source proxy to handle this case.
    
Resolves: rhbz#1861456

### DIFF
--- a/pyanaconda/ui/lib/subscription.py
+++ b/pyanaconda/ui/lib/subscription.py
@@ -25,6 +25,7 @@ from pyanaconda.core.constants import THREAD_WAIT_FOR_CONNECTING_NM, \
     SUBSCRIPTION_REQUEST_TYPE_USERNAME_PASSWORD, SUBSCRIPTION_REQUEST_TYPE_ORG_KEY, \
     SOURCE_TYPE_HDD, SOURCE_TYPE_CDN, SOURCE_TYPES_OVERRIDEN_BY_CDN
 from pyanaconda.core.i18n import _
+from pyanaconda.core.constants import PAYLOAD_TYPE_DNF
 from pyanaconda.ui.lib.payload import create_source, set_source, tear_down_sources
 from pyanaconda.ui.lib.storage import unmark_protected_device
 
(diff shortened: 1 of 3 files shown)


pyanaconda/ui/lib/subscription.py
```diff
@@ -53,14 +53,33 @@ class SubscriptionPhase(Enum):
...
+    :return: the source proxy or None is payload has no source proxy
+    :rtype: DBus proxy object or None
+
+    In some cases, such as for the OSTree payload there
+    is no source proxy and we need to handle that,
+    at least until payload refactorization has been finished.
+    """
+    try:
+        source_proxy = payload.get_source_proxy()
+    except AttributeError:
```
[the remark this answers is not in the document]
M4rtinK: Done and done. Thanks for the suggestion. :)

pyanaconda/ui/lib/subscription.py
```diff
@@ -59,8 +60,8 @@ def _tear_down_existing_source(payload):
     :param payload: Anaconda payload instance
     """
     source_proxy = payload.get_source_proxy()
-
-    if source_proxy.Type == SOURCE_TYPE_HDD and source_proxy.Partition:
+    should_unamrk_protected = source_proxy.Type == SOURCE_TYPE_HDD and source_proxy.Partition
+    if payload.type == PAYLOAD_TYPE_DNF and should_unamrk_protected:
```
poncovka, line 64: I don't get it. The payload type is already checked before you call `switch_source` that calls ` _tear_down_existing_source`, so the type should be correct at this point.

pyanaconda/ui/lib/subscription.py
```diff
@@ -59,8 +60,8 @@ def _tear_down_existing_source(payload):
     :param payload: Anaconda payload instance
     """
     source_proxy = payload.get_source_proxy()
-
-    if source_proxy.Type == SOURCE_TYPE_HDD and source_proxy.Partition:
+    should_unamrk_protected = source_proxy.Type == SOURCE_TYPE_HDD and source_proxy.Partition
```
jkonecny12, line 63:
```suggestion
    should_unmark_protected = source_proxy.Type == SOURCE_TYPE_HDD and source_proxy.Partition
```

M4rtinK: Fixed, thanks! :)

pyanaconda/ui/lib/subscription.py
M4rtinK: Yeah, you are right - I don't think we need to be that careful if it's checked before `switch_source()`. I'll drop that. :)

M4rtinK:
Port to Fedora: #2808

PR #2759, opened 2020-07-30
Match names from ignoredisk after setting up all devices

poncovka:
We should match names specified by the ignoredisk kickstart command after
we set up devices defined by iscsi, fcoe and other kickstart commands. Parse
the ignoredisk command in UI and update the Disk Selection module to fix it.


jkonecny12:
Ahh, I hope we won't be merging this to rhel-8.2-prerebase...

PR #2761, opened 2020-08-03
Always recreate unified repos

jkonecny12:
This will solve problem that additional repositories pointing to an invalid path. This is happening because we moved source mounting to a different folders for each source. By this patch we will always reload new treeinfo repository if user change the base source repository.

Also this changed behavior of Source spoke. The Source spoke does not disable repository on base repo switch but instead it will remove this repository and the repository will be reloaded when user re-enter the spoke. It's not ideal solution but doable with avoiding great amount of changes everywhere.

Backport of https://github.com/rhinstaller/anaconda/pull/2757 .

### DIFF
--- a/pyanaconda/payload/base.py
+++ b/pyanaconda/payload/base.py
@@ -39,8 +39,6 @@ class Payload(metaclass=ABCMeta):
         """
         self.data = data
 
-        self._first_payload_reset = True
-
         # A list of verbose error strings from the subclass
         self.verbose_errors = []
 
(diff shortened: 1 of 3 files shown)


jkonecny12:
jenkins, test this please

jkonecny12:
**UPDATED:**
* Add refactorization.
* Remove `add_disabled_repo()` instead use `add_repo()`.

I should make sure that addons are not using this API.

jkonecny12:
jenkins, test this please

jkonecny12:
Tested with modified Rawhide ISO to test unified repo functionality. Everything seems to be working correctly here.

PR #2762, opened 2020-08-03
subscription: Convert the RHSM default config values to expected format

M4rtinK:
Turns out the RHSM config DBus API GetAll() returns a nested
distionary instead of a flat dict we were expecting, as SetAll()
uses such a flat dict.

Due to this empty values from GUI would be written to the RHSM
config file as no original values would be found due to the
unexpected format of the GetAll() returned dict.

To fix this, convert the nested dict to the flat one we expect &
which is much easier to work with in our use case.

This way the Candlepin URL & CDN URL should no longer be overwritten
by empty values from the GUI.

Resolves: rhbz#1862116

### DIFF
--- a/pyanaconda/modules/subscription/subscription.py
+++ b/pyanaconda/modules/subscription/subscription.py
@@ -556,6 +556,26 @@ class SubscriptionService(KickstartService):
         """
         return self._rhsm_observer
 
+    def _flatten_rhsm_nested_dict(self, nested_dict):
+        """Convert the GetAll() returned nested dict into a flat one.
+
+        RHSM returns a nested dict with categories on top
+        and category keys & values inside. This is not convenient
+        for setting keys based on original values, so
+        let's normalize the dict to the flat key based
+        structure similar to what's used by SetAll().
+
+        :param dict nested_dict: the nested dict returned by GetAll()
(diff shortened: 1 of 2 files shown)


pyanaconda/modules/subscription/subscription.py
```diff
@@ -572,14 +572,29 @@ def get_rhsm_config_defaults(self):
...
             proxy = self.rhsm_observer.get_proxy(RHSM_CONFIG)
             # turn the variant into a dict with get_native()
-            self._rhsm_config_defaults = get_native(proxy.GetAll(""))
+            nested_dict = get_native(proxy.GetAll(""))
+            # RHSM returns a nested dict with categories on top
+            # and category keys & values inside. This is not convenient
+            # for setting keys based on original values, so
+            # let's normalize the dict to the flat key based
+            # structure similar to what's used by SetAll().
+            flat_dict = {}
```
poncovka, line 592: I would prefer to keep the logic in a function that could be somewhere else.

M4rtinK: Moved to a helper function. :)

M4rtinK:
Port to Fedora: #2809

PR #2763, opened 2020-08-03
Don't try to pass nonexistent fields (#1860058)

VladimirSlavik:
This fixes the bug where "homedir" does not exist on F24_SshPwData and no user is created for SSH access to installation environment.

Resolves: [rhbz#1860058](https://bugzilla.redhat.com/show_bug.cgi?id=1860058)

### DIFF
--- a/utils/handle-sshpw
+++ b/utils/handle-sshpw
@@ -47,8 +47,8 @@ for ud in userdata:
             users.set_user_password(username=ud.username, password=ud.password,
                                     is_crypted=ud.isCrypted, lock=ud.lock)
     else:
-        users.create_user(username=ud.username, password=ud.password, is_crypted=ud.isCrypted, lock=ud.lock,
-                          homedir=ud.homedir, shell=ud.shell, gecos=ud.gecos, root="/")
+        users.create_user(username=ud.username, password=ud.password, is_crypted=ud.isCrypted,
+                          lock=ud.lock, root="/")
 
     if ud.sshkey:
         # Setup the account so that only the sshkey can be used


VladimirSlavik:
I think this doesn't need porting, but I'm not entirely sure.

utils/handle-sshpw
```diff
@@ -47,8 +47,11 @@ for ud in userdata:
             users.set_user_password(username=ud.username, password=ud.password,
                                     is_crypted=ud.isCrypted, lock=ud.lock)
     else:
-        users.create_user(username=ud.username, password=ud.password, is_crypted=ud.isCrypted, lock=ud.lock,
-                          homedir=ud.homedir, shell=ud.shell, gecos=ud.gecos, root="/")
+        users.create_user(
+            username=ud.username, password=ud.password, is_crypted=ud.isCrypted, lock=ud.lock,
+            homedir=getattr(ud, "homedir", None), shell=getattr(ud, "shell", None),
```
poncovka, line 52: This bug was introduced in https://github.com/rhinstaller/anaconda/commit/1278b3f612a55e0b8dda82b0ca65b0afc46fceba. The `sshpw` command doesn't have options for `homedir`, `shell` or `gecos` (see https://pykickstart.readthedocs.io/en/latest/kickstart-docs.html#sshpw), so you can remove the optional arguments instead of using `getattr`.

VladimirSlavik: Right, that's better...

VladimirSlavik:
Heh, thank you for your help, as always...

PR #2765, opened 2020-08-04
Create ssh user using only existing fields (#1860058)

VladimirSlavik:
This fixes the bug where `homedir` and other fields do not exist on `F24_SshPwData` and no user is created for SSH access to installation environment.

Resolves: [rhbz#1860058](https://bugzilla.redhat.com/show_bug.cgi?id=1860058)

Port of #2763

### DIFF
--- a/utils/handle-sshpw
+++ b/utils/handle-sshpw
@@ -47,8 +47,8 @@ for ud in userdata:
             users.set_user_password(username=ud.username, password=ud.password,
                                     is_crypted=ud.isCrypted, lock=ud.lock)
     else:
-        users.create_user(username=ud.username, password=ud.password, is_crypted=ud.isCrypted, lock=ud.lock,
-                          homedir=ud.homedir, shell=ud.shell, gecos=ud.gecos, root="/")
+        users.create_user(username=ud.username, password=ud.password, is_crypted=ud.isCrypted,
+                          lock=ud.lock, root="/")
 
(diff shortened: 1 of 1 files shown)


VladimirSlavik:
Jenkins, test this please.

poncovka:
Btw. I am not sure about including RHEL bug numbers into upstream commit messages this way. For cherry picks, I usually add `(cherry-picked from a commit 1234567)` at the end of the message. You can always find the bug from in the original commit. Or you can use something like `See the bug 12345` in the message if the bug is really important. Otherwise, I think that bugs from `Resolves:` and `Related:` should match the current branch. @jkonecny12, what do you think?

jkonecny12:
> Btw. I am not sure about including RHEL bug numbers into upstream commit messages this way. For cherry picks, I usually add `(cherry-picked from a commit 1234567)` at the end of the message. 

Just FYI look on ``git cherry-pick -x``

> You can always find the bug from in the original commit. Or you can use something like `See the bug 12345` in the message if the bug is really important. Otherwise, I think that bugs from `Resolves:` and `Related:` should match the current branch. @jkonecny12, what do you think?

Personally, I don't change commit message on cherry-picking so I don't think the bug number is a problem here. It could be even helpful to find original bug sooner.

VladimirSlavik:
Commit message extended with the cherrypick source commit ref.

PR #2769, opened 2020-08-05
Drop the dependency on python3-ntplib

poncovka:
* Use `chronyd -Q` to check the status of an NTP server.
* Don't synchronize the system time with a working NTP server.
* Let the chronyd service to synchronize the time.

### DIFF
--- a/pyanaconda/ntp.py
+++ b/pyanaconda/ntp.py
@@ -25,14 +25,12 @@ import re
 import os
 import tempfile
 import shutil
-import ntplib
-import socket
 
-from pyanaconda import isys
 from pyanaconda.anaconda_loggers import get_module_logger
 from pyanaconda.core.i18n import N_, _
-from pyanaconda.core.constants import THREAD_SYNC_TIME_BASENAME, NTP_SERVER_QUERY, \
+from pyanaconda.core.constants import NTP_SERVER_TIMEOUT, NTP_SERVER_QUERY, \
     THREAD_NTP_SERVER_CHECK, NTP_SERVER_OK, NTP_SERVER_NOK
+from pyanaconda.core.util import execWithRedirect
 from pyanaconda.modules.common.structures.timezone import TimeSourceData
 from pyanaconda.threading import threadMgr, AnacondaThread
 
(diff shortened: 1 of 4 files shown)


pyanaconda/ntp.py
```diff
@@ -107,25 +105,14 @@ def ntp_server_working(server_hostname, nts_enabled):
     :return: True if the given server is reachable and working, False otherwise
     :rtype: bool
     """
-    try:
-        # ntplib doesn't support NTS
-        if nts_enabled:
-            s = socket.create_connection((server_hostname, 4460), 2)
```
poncovka, line 113: @mlichvar, can I replace your solution for nts with the one below? It seems to work fine.

mlichvar: Yes, I think that's a good idea. Thanks.

mlichvar: NETWORK_CONNECTION_TIMEOUT seems to be 46 seconds. Isn't that a bit too long for this check? I'd suggest something like 8 seconds, or so.

pyanaconda/ntp.py
```diff
@@ -107,25 +105,14 @@ def ntp_server_working(server_hostname, nts_enabled):
...
-    except socket.error:
-        return False
+    timeout = str(NETWORK_CONNECTION_TIMEOUT)
+    directive = ["server", server_hostname, "iburst", "maxsamples", "1"]
+
+    if nts_enabled:
+        directive.append("nts")
 
-    return True
+    arguments = ["-Q", " ".join(directive), "-t", timeout]
```
jkonecny12, line 114: I would remove the `timeout` variable instead just add the ``str(NETWORK_CONNECTION_TIMEOUT)`` directly.

poncovka: Fixed.

pyanaconda/ntp.py
poncovka: Yep, that is too long. The python3-ntplib library sets the timeout to 5 seconds, so I have changed it to the same number.

poncovka:
Updated.

PR #2770, opened 2020-08-06
network: fix parsing of hostname from ip= if mac is defined in dhcp

rvykydal:
Resolves: rhbz#1852560


rvykydal:
PR with port to Fedora: https://github.com/rhinstaller/anaconda/pull/2834

PR #2771, opened 2020-08-06
Fix kickstart file error with user groups

parr0tr1ver:
When fill the "Group Membership" in "ADVANCED USER CONFIGURATION"
dialog with the provided example text:

wheel, my-team (1245), project-x (29935)

It keeps the spaces between group name(project-x) and group id('(29935)'),
and then write them to kickstart file. When boot anaconda with the kickstart
file, it fails with:

| unrecognzed arguments: (1245),project-x (29935)

Filter out the spaces between group name and group id to avoid such
issue.

Signed-off-by: Kai Kang <kai.kang@windriver.com>


jkonecny12:
Jenkins, it's ok to test.

jkonecny12:
Tested and it works as expected.

PR #2775, opened 2020-08-07
network: pass also keyfile NM configuration to target system (#1858439)

rvykydal:


rvykydal:
This is just a start of support for NM keyfiles in Anaconda and unblocking of some of Adam's tests. The plan is that Anaconda would be able to handle both at the same time (as NM is), but the direct handling (reading/writing) of NM config files would be drastically reduced (to none ideally) and replaced by using NM API and its configuration objects.

PR #2776, opened 2020-08-07
Enable Packit for smoke testing

jkonecny12:
This is a first step to support Packit. In the future we would like to run our tests and make releases by Packit.

**Ready for review.**

### DIFF
--- a/.packit.yml
+++ b/.packit.yml
@@ -1,0 +1,11 @@
+specfile_path: anaconda.spec
+upstream_package_name: anaconda
+actions:
+  post-upstream-clone:
+    - ./autogen.sh
+    - ./configure
+  get-current-version:
+    - bash -c "git describe --tags --match '*.*' | awk -F '-' '{ print $2}' "
+  create-archive:
+    - "make release"
+    - "bash -c 'ls | grep tar.bz2'"
--- a/.packit.yml
+++ b/.packit.yml
@@ -9,3 +9,9 @@ actions:
   create-archive:
     - "make release"
     - "bash -c 'ls | grep tar.bz2'"
+jobs:
+  - job: tests
+    trigger: pull_request
+    metadata:
+      targets:
+        - fedora-rawhide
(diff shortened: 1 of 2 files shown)


TomasTomecek:
wow, very nice progress! I can see the failure now is that copr can't find the correct version of kickstart:
```
No matching package to install: 'python3-kickstart >= 3.27-1'
```

jkonecny12:
True I have to run these tests only on Rawhide because we are using Rawhide dependencies here.

.packit.yml
```diff
@@ -0,0 +1,17 @@
+specfile_path: anaconda.spec
+upstream_package_name: anaconda
+actions:
+  post-upstream-clone:
+    - ./autogen.sh
+    - ./configure
+  get-current-version:
+    - bash -c "git describe --tags --match '*.*' | awk -F '-' '{ print $2}' "
+  create-archive:
+    - "make release"
```
TomasTomecek, line 10:
```suggestion
    - "make release"
    - 'bash -c "tar -tf *.tar.bz2 | tail"'
```
let's make sure the tarball is valid, `tail` will make the output brief

TomasTomecek (commented):
and in the specfile I'd do something like this:
```
%prep
ls -lha %{SOURCE0}
tar -tf %{SOURCE0}
%autosetup
```

jkonecny12:
Yep, it's definitely not an archive:
```
-rw-rw-r--. 1 mockbuild mock 14 Aug  7 14:35 /builddir/build/SOURCES/anaconda-33.24.tar.bz2
```

The size is wrong. I'll try to add cat, just to see if it is not response or something like that.

jkonecny12:
Hmm, I'm not really smart from the output. Do you see something @TomasTomecek ?

TomasTomecek:
I'm not sure I understand why the tarball did not get created: maybe something changed in upstream/master in the meantime?

Please let me know if there is anything I can help with in future.

jkonecny12:
Sorry, I'm a bit swamped by other work right now but I plan to look on that soon.

Thanks a lot for all your help!

VladimirSlavik (approved):
It works! \o/

poncovka (approved):
It looks great! Thank you!

PR #2778, opened 2020-08-07
Fix our tests with a newer rpmfluff library

jkonecny12:
They made breaking changes in the library.

https://pagure.io/rpmfluff/issue/42

This will probably happen even on RHEL branch. I guess we want to pin version there to avoid this issue instead of creating bug to fix tests. :(


M4rtinK:
Merging now as arranged with @jkonecny12 on Friday.

PR #2780, opened 2020-08-12
Skip auto updates if inst.updates is given

cheese:
Anaconda tried to load updates image from both where specified by inst.updates and
the default path in repo. But user expects the one specified by inst.updates should
take effect.

What even worse is when repo is loaded from disk, loading updates from network and
disk are triggered by different udev rules and there is raise condition. Updates
from repo default path may override what user specified in inst.updates.

This change will skip loading updates from repo when inst.updates is given.

### DIFF
--- a/dracut/anaconda-lib.sh
+++ b/dracut/anaconda-lib.sh
@@ -117,14 +117,15 @@ anaconda_net_root() {
         info "anaconda: successfully fetched stage2 from $repo"
 
         # NOTE: Should be the same as anaconda_auto_updates()
-        updates=$(fetch_url $repo/images/updates.img 2> /tmp/updates_err)
-        [ -z "$updates" ] && debug_msg $(cat /tmp/updates_err)
-        [ -n "$updates" ] && unpack_updates_img $updates /updates
+        if ! getargbool 0 updates inst.updates; then
+            updates=$(fetch_url $repo/images/updates.img 2> /tmp/updates_err)
+            [ -z "$updates" ] && debug_msg $(cat /tmp/updates_err)
+            [ -n "$updates" ] && unpack_updates_img $updates /updates
 
-        product=$(fetch_url $repo/images/product.img 2> /tmp/product_err)
-        [ -z "$product" ] && debug_msg $(cat /tmp/product_err)
-        [ -n "$product" ] && unpack_updates_img $product /updates
-
+            product=$(fetch_url $repo/images/product.img 2> /tmp/product_err)
+            [ -z "$product" ] && debug_msg $(cat /tmp/product_err)
+            [ -n "$product" ] && unpack_updates_img $product /updates
+        fi
         anaconda_mount_sysroot $runtime
         return 0
     fi
@@ -146,6 +147,10 @@ anaconda_mount_sysroot() {
 # find updates.img/product.img/RHUpdates and unpack/copy them so they'll
 # end up in the location(s) that anaconda expects them
 anaconda_auto_updates() {
+    if getargbool 0 updates inst.updates; then
+        info "anaconda: skip auto updates since inst.updates is given"
+        return
+    fi
     local dir="$1"
     if [ -d $dir/RHupdates ]; then
         copytree $dir/RHupdates /updates


jkonecny12:
Jenkins, it's ok to test.

dracut/anaconda-lib.sh
```diff
@@ -117,14 +117,15 @@ anaconda_net_root() {
...
-        [ -n "$product" ] && unpack_updates_img $product /updates
-
+        if ! getargbool 0 updates inst.updates; then
+            updates=$(fetch_url $repo/images/updates.img 2> /tmp/updates_err)
+            [ -z "$updates" ] && debug_msg $(cat /tmp/updates_err)
+            [ -n "$updates" ] && unpack_updates_img $updates /updates
+
+            product=$(fetch_url $repo/images/product.img 2> /tmp/product_err)
+            [ -z "$product" ] && debug_msg $(cat /tmp/product_err)
+            [ -n "$product" ] && unpack_updates_img $product /updates
```
jkonecny12, line 127: I don't think updates images should be skipped based on ``inst.updates`` kernel param.

Their are basically the same but product image is specialized on rebranding so it shouldn't be excluded based on updates_image.

jkonecny12 (changes requested):
Also we should look on ``OEMDRV``, I think we are using this to load updates_images too so the same logic should be applied there.

jkonecny12:
I would just make a small modification to exclude the product.img from the `if` condition.

Optional is to investigate the OEMDRV but even without that it's fine by me.

VladimirSlavik (changes requested):
I can confirm that the patch as presented works as described. However, I agree with @jkonecny12 that the product image should not be excluded. Updates are a debug/devel thing, while product is meant for regular use on non-code parts of the environments. That isn't too big of a change anyway :-)

The OEMDRV thing still need additional investigation.

jkonecny12:
In that case @cheese please change the condition to include only updates image and I think we should be ready for merge. ;)

cheese:
If auto loading of product.img is kept, then my use case is not fulfilled. In my use case, image specified with inst.updates is the new version of product.img. I want only the data from inst.updates.

If loading default product.img is kept, we should make sure what from inst.updates is overriding. So we should figure out a complete solution other than just this patch.

VladimirSlavik:
I did not know, so I ran a quick test - looks like the order in which the images are applied is:

1. `inst.updates=`
2. `product.img`
3. `updates.img`

To me that seems just wrong. I would expect the boot option image to be applied at the end of the list. It's the most accessible one, and sees most debugging use where one wants to override everything.

I have no idea what are the considerations for backwards compatibility here.

jkonecny12:
> I have no idea what are the considerations for backwards compatibility here.

That's fine. Updates images are used specifically for debugging purposes so the ordering should not matter. I agree that the ordering is not good.

@cheese, what do you think about merging this with the moved ``if`` as discussed above and we can look on adjusting order to this:

- product.img
- updates.img

Whould that work for your issue?

jkonecny12:
Another possibility is to introduce new boot option to avoid loading of product image but I would like to avoid that if possible.

VladimirSlavik:
I'd like to add that since you want something to be not overwritten by `product.img`, you can use `updates.img` in the repo, right now.

Alternatively, maybe try applying the `inst.updates` image after these? I'm not sure if this could work, but maybe it's the fastest compromise for now:

```
        if [[ -n ${anac_updates} ]]; then
            updates_boot=$(fetch_url $anac_updates 2> /tmp/product_err)
            [ -z "$updates_boot" ] && debug_msg $(cat /tmp/updates_boot_err)
            [ -n "$updates_boot" ] && unpack_updates_img updates_boot /updates
        fi
```

`anac_updates` should be set earlier when `inst.updates` is handled, or so I believe.

VladimirSlavik:
Looks like this is another piece for the graveyard collection...

jkonecny12:
@cheese is this patch still interesting to you or we can close it?

cheese:
This is actually a minor case. I don't have time to work out a better patch for now.

PR #2781, opened 2020-08-12
Reset the state of the custom partitioning spoke (#1868330)

poncovka:
We should reset the state of the custom partitioning spoke after every user
interaction. If we don't reset the _back_already_clicked attribute, the final
check of the storage configuration might be skipped and the status of the
storage spoke might show an error from the previously failed check.

Resolves: rhbz#1868330


poncovka:
Jenkins, test this please.

PR #2783, opened 2020-08-12
Remove usage of OrderedSet

VladimirSlavik:
...by turning the class into a ~~list descendant~~ standalone thing with checks on adding.

### DIFF
--- a/pyanaconda/modules/storage/bootloader/base.py
+++ b/pyanaconda/modules/storage/bootloader/base.py
@@ -40,7 +40,7 @@ from pyanaconda.modules.common.structures.network import NetworkDeviceInfo
 
 log = get_module_logger(__name__)
 
-__all__ = ["BootLoaderError", "Arguments", "BootLoader"]
+__all__ = ["BootLoaderError", "BootLoaderArguments", "BootLoader"]
 
 
 def _is_on_sw_iscsi(device):
@@ -87,7 +87,7 @@ class BootLoaderError(Exception):
     pass
 
 
-class Arguments:
+class BootLoaderArguments:
     """An ordered set of bootloader arguments."""
 
     def __init__(self):
@@ -186,8 +186,8 @@ class BootLoader(object):
 
     def __init__(self):
         super().__init__()
-        self.boot_args = Arguments()
-        self.dracut_args = Arguments()
+        self.boot_args = BootLoaderArguments()
+        self.dracut_args = BootLoaderArguments()
 
         # the device the bootloader will be installed on
         self.stage1_device = None
--- a/pyanaconda/modules/storage/bootloader/base.py
+++ b/pyanaconda/modules/storage/bootloader/base.py
@@ -24,7 +24,6 @@ from blivet.devices import NetworkStorageDevice
 from blivet.formats.disklabel import DiskLabel
 from blivet.iscsi import iscsi
 from blivet.size import Size
-from ordered_set import OrderedSet
 
 from pyanaconda.network import iface_for_host_ip
 from pyanaconda.modules.storage.platform import platform
@@ -88,9 +87,11 @@ class BootLoaderError(Exception):
     pass
 
 
-class Arguments(OrderedSet):  # pylint: disable=duplicate-bases
-    """An ordered set of arguments."""
-    # FIXME: the pylint pragma is needed since 2.5.0 for unclear reasons
+class Arguments:
+    """An ordered set of bootloader arguments."""
+
+    def __init__(self):
+        self._arguments = []
 
     def _merge_ip(self):
         """Find ip= arguments targeting the same interface and merge them."""
@@ -101,8 +102,8 @@ class Arguments(OrderedSet):  # pylint: disable=duplicate-bases
             # automatic network setup:
             return arg.startswith("ip=") and arg.count(":") == 1
 
-        ip_params = filter(partition_p, self)
-        rest = OrderedSet(filter(lambda p: not partition_p(p), self))
+        ip_params = list(filter(partition_p, self._arguments))
+        rest = list(filter(lambda p: not partition_p(p), self._arguments))
 
         # split at the colon:
         ip_params = map(lambda p: p.split(":"), ip_params)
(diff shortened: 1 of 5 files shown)


VladimirSlavik:
For consideration: Maybe the inheritance should be entirely removed, and the list should be an internal implementation detail of the class? Using extend and such is still possible, making the "solution" full of holes.

poncovka:
> Maybe the inheritance should be entirely removed, and the list should be an internal implementation detail of the class? Using extend and such is still possible, making the "solution" full of holes.

I agree. It is more safe to remove the inheritance and implement the necessary methods than the other way around. Otherwise, it would be too easy to start to use the arguments in the "unsupported" way. You might also consider to use a dictionary instead of the list, because it keeps its items in the insertion order.

VladimirSlavik:
I implemented pretty much everything from my earlier comment.

> use a dictionary instead of the list, because it keeps its items in the insertion order.

@poncovka I am not really sure what you mean by this. AFAICT a list does keep items in the order... unless you meant some other implication? The only place where the code can lose order of args is in `_merge_ip` where a `set` is used. But that's the way it was before, so I guess that's ok?

VladimirSlavik:
Apart from the above, the object now needs being iterable only so as to enable the test. I can see some value in making it iterable regardless, but as it is not used, maybe it would be better to just rewrite the test?

pyanaconda/modules/storage/bootloader/base.py
```diff
@@ -116,24 +117,31 @@ def partition_p(arg):
...
+
+        Merges ip= arguments targeting the same interface as a side effect.
+        """
         self._merge_ip()
-        return " ".join(list(self))
+        return " ".join(list(self._arguments))
 
-    def add(self, key):
-        self.discard(key)
-        super().add(key)
```
jkonecny12, line 131: Please leave the `add` there. It's not nice to making list from everything. Or better add new `extend` method which will work with lists and change the `update` method.

pyanaconda/modules/storage/bootloader/base.py
```diff
@@ -116,24 +117,31 @@ def partition_p(arg):
...
     def update(self, sequence):
+        """Add arguments from some other iterable object.
+
+        :param sequence: iterable object
+        """
         for key in sequence:
-            self.discard(key)
-            self.add(key)
+            if key not in self._arguments:
+                self._arguments.append(key)
```
jkonecny12, line 144: This breaks behavior. The reason there was `discard` and then `add` was to move this item to the last in the list (set). In your implementation you are not doing that. I guess because of this the [recommended way](https://github.com/LuminosoInsight/ordered-set#comparisons) is to use dict instead, it's much faster for look-ups.

pyanaconda/modules/storage/bootloader/base.py
```diff
@@ -116,24 +117,31 @@ def partition_p(arg):
...
+        return iter(self._arguments)
 
     def update(self, sequence):
+        """Add arguments from some other iterable object.
+
+        :param sequence: iterable object
+        """
         for key in sequence:
-            self.discard(key)
-            self.add(key)
```
poncovka, line 136: I think that the effect of `discard` and `add` was that the argument was always appended at the end. That's no longer true.

pyanaconda/modules/storage/bootloader/extlinux.py
```diff
@@ -61,12 +62,12 @@ def write_config_console(self, config):
...
         if self.console_options:
             console_arg += ",%s" % self.console_options
-        self.boot_args.add(console_arg)
+        self.boot_args.update([console_arg])
 
     def write_config_images(self, config):
         self.write_config_console(config)
         for image in self.images:
-            args = Arguments()
+            args = BootLoaderArguments()
```
poncovka, line 70: Could you please write some unit tests and test them on the `Arguments` class first to make sure that the `BootLoaderArguments` class works the same way?

pyanaconda/modules/storage/bootloader/base.py
```diff
@@ -88,9 +87,11 @@ class BootLoaderError(Exception):
     pass
 
 
-class Arguments(OrderedSet):  # pylint: disable=duplicate-bases
-    """An ordered set of arguments."""
-    # FIXME: the pylint pragma is needed since 2.5.0 for unclear reasons
+class BootLoaderArguments:
```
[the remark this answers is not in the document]
jkonecny12: `Object` is not mandatory but `()` should be there if I'm not mistaken ...

pyanaconda/modules/storage/bootloader/extlinux.py
VladimirSlavik: I guess this was the smart thing to start with :-/

pyanaconda/modules/storage/bootloader/base.py
poncovka: I think there are `add` and `update` method, because the class behaves like an ordered set. I think we shouldn't change that. The list is only an implementation detail.

poncovka:
> I implemented pretty much everything from my earlier comment.
> 
> > use a dictionary instead of the list, because it keeps its items in the insertion order.
> 
> @poncovka I am not really sure what you mean by this. AFAICT a list does keep items in the order... unless you meant some other implication? The only place where the code can lose order of args is in `_merge_ip` where a `set` is used. But that's the way it was before, so I guess that's ok?

Well, the `dict` is an ordered dictionary, so it is the closest thing to the ordered set as you can get if you only work with keys. However, I haven't realized that we change the ordering and always move the newest items at the end of the sequence, so I guess it is better to use the list and have a full control over the ordering.

pyanaconda/modules/storage/bootloader/base.py
poncovka: Btw. the `update` method should just call `self.add` on all items of the sequence and only the `add` method should implement the logic for adding an item into the list.

VladimirSlavik: Sorry, fixed.

pyanaconda/modules/storage/bootloader/extlinux.py
VladimirSlavik: Done.

VladimirSlavik:
Updated with tests, onwards to kickstart tests, I guess?

VladimirSlavik:
Kickstart tests had mixed success, didn't pass bootloader-4 and -5. It's pretty hard to interpret the results though, so I have no idea if these are actually broken by this PR or not.

pyanaconda/modules/storage/bootloader/base.py
```diff
@@ -725,7 +743,7 @@ def _set_storage_boot_args(self, storage):
         # FIPS
         boot_device = storage.mountpoints.get("/boot")
         if kernel_arguments.get("fips") == "1" and boot_device:
-            self.boot_args.add("boot=%s" % self.stage2_device.fstab_spec)
+            self.boot_args.update(["boot=%s" % self.stage2_device.fstab_spec])
```
jkonecny12, line 746: Please use the `add` method when it's available now.

tests/nosetests/pyanaconda_tests/module_storage_bootloader_args_test.py
```diff
@@ -0,0 +1,66 @@
...
+        """Arguments.update reorders things as expected."""
+        args = Arguments()
+
+        args.update(["one", "two", "three"])
+        args.update("abc")
+        self.assertEqual(str(args), "one two three a b c")
+        args.update(["three", "two"])
+        self.assertEqual(str(args), "one a b c three two")
+
+    def ip_merge_test(self):
```
jkonecny12, line 49: The IP solution should be split from the class and make it separate, however, lets do that in the future in some other PR.

VladimirSlavik: Sounds good.

pyanaconda/modules/storage/bootloader/base.py
VladimirSlavik: Done! Thanks for catching these.

VladimirSlavik:
The last commit is optional.

VladimirSlavik:
Rebased from master to f33-devel.

PR #2784, opened 2020-08-12
Propagate verify_ssl to RHSM

M4rtinK:
At the moment the inst.noverifyssl boot option and the corresponding
Anaconda config file option verify_ssl only influence SSL certification
validation when anaconda itself downloads some installation related
artifacts.

For consistency this option should be also propagated to the RHSM
install time configuration.

Also this way users that acknowledge the risks by passing the
inst.noverifyssl boot option can successfully register against
a Candlepin instance with certificate that the installation
environment considers unknown or invalid.

Due to how this option is set (effectively overwriting the RHSM config
file) we need to make RHSM SSL certificate validation is turned back ON
at the end of the installation. This is again consistent with how
SSL certificate verification overrides are implemented elsewhere in
Anaconda. The overrides are always install time only and never propagate
to the installed system.

On the impelementation side we rename one of the installation tasks
as it no longer only reverts the log level changes but also makes
sure SSL certification validation is enabled in RHSM on the installed
system. Also change from Set() to SetAll() as se no longer just set a
single value in the two tasks that now interfacw with the RHSM SSL
certificate validation toggle.

Resolves: rhbz#1858126

### DIFF
--- a/pyanaconda/modules/subscription/initialization.py
+++ b/pyanaconda/modules/subscription/initialization.py
@@ -38,6 +38,19 @@ class StartRHSMTask(Task):
 
     RHSM_SYSTEMD_UNIT_NAME = "rhsm.service"
 
+    def __init__(self, verify_ssl=True):
+        """Create a new task for starting the RHSM DBus service.
+
+        :param bool verify_ssl: True if RHSM should be configured to verify SSL certificates,
+                                False if RHSM should be set to *not* verify SSL certificates
+
+        NOTE: If RHSM SSL verification is disabled, this is install time only, as we will
+              always turn it back on unconditionally at the same time we roll back the RHSM
+              log level change.
+        """
+        super().__init__()
+        self._verify_ssl = verify_ssl
+
     @property
     def name(self):
         return "Start RHSM DBus service"
@@ -71,7 +84,15 @@ class StartRHSMTask(Task):
         # set RHSM log levels to debug
         # - otherwise the RHSM log output is not usable for debugging subscription issues
         log.debug("subscription: setting RHSM log level to DEBUG")
-        rhsm_config_proxy.Set("logging.default_log_level", get_variant(Str, "DEBUG"), "")
+        config_dict = {"logging.default_log_level": get_variant(Str, "DEBUG")}
+        # turn OFF SSL certificate validation (if requested)
+        if not self._verify_ssl:
+            log.debug("subscription: disabling RHSM SSL certificate validation")
+            config_dict["server.insecure"] = get_variant(Str, "1")
+
+        # set all the values at once atomically
+        rhsm_config_proxy.SetAll(config_dict, "")
+
         # all seems fine
         log.debug("subscription: RHSM service start successfully.")
         return True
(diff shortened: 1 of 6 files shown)


pyanaconda/modules/subscription/initialization.py
```diff
@@ -38,6 +38,19 @@ class StartRHSMTask(Task):
...
     RHSM_SYSTEMD_UNIT_NAME = "rhsm.service"
 
+    def __init__(self, verify_ssl=True):
+        """Create a new task for starting the RHSM DBus service.
+
+        :param bool verify_ssl: True if RHSM should be configured to verify SSL certificates,
+                                False if RHSM should be set to *not* verify SSL certificates
+
+        NOTE: If RHSM SSL verification is disabled, this is install time only, as we will
+              always turn it back on uncoditionally at the same time we roll back the RHSM
```
rvykydal, line 48: unconditionally

pyanaconda/modules/subscription/initialization.py
```diff
@@ -38,6 +38,19 @@ class StartRHSMTask(Task):
...
+              log lvevel change.
```
rvykydal, line 49: level

pyanaconda/modules/subscription/installation.py
```diff
@@ -81,8 +81,14 @@ def run(self):
...
 
 
-class RestoreRHSMLogLevelTask(Task):
-    """Restore RHSM log level back to INFO."""
+class RestoreRHSMDefaultsTask(Task):
+    """Restore RHSM defaults we changed for install time purposes.
+
+    At the moment this means setting the RHSM log level back to INFO
+    from DEBUG and making sure SSL certificate validation is enabled
+    (as we might turn it of for the installation run if requested by
```
rvykydal, line 89: turn it off

tests/nosetests/pyanaconda_tests/rhsm_observer_test.py
```diff
@@ -55,9 +55,42 @@ def success_test(self, start_service, get_proxy):
...
+        # run the task and expect it to succeed
+        self.assertTrue(task.run())
+        # check service was started correctly
+        start_service.assert_called_once_with("rhsm.service")
+        # check proxy was requested
+        get_proxy.assert_called_once_with()
+        # check expected values were set on the RHSM config proxy
+        # - logging should be always set to DEBUG
+        # - SSL certificate validation should be enabled by default
+        #   (insecure == 0)
```
rvykydal, line 88: insecure == 1 ?

rvykydal, line 88: SSL certificate validation should be disabled ?

M4rtinK:
@jkonecny12 commit message fixed :)

pyanaconda/modules/subscription/initialization.py
M4rtinK: Fixed. :)

pyanaconda/modules/subscription/initialization.py
M4rtinK: Fixed. :)

pyanaconda/modules/subscription/installation.py
M4rtinK: Fixed. :)

tests/nosetests/pyanaconda_tests/rhsm_observer_test.py
M4rtinK: Good catch, fixed! :)

M4rtinK:
Port to Fedora: #2821

PR #2785, opened 2020-08-13
Reset the RAID level of the device request (#1828092)

poncovka:
In the custom partitioning spoke, always reset the RAID level of the
device request when the device type changes. Otherwise, the new type
doesn't have to support the old RAID level.

Resolves: rhbz#1828092


M4rtinK (approved):
Looks good to me, thanks! :)

PR #2786, opened 2020-08-13
Print warning for boot args without `inst.` prefix

jkonecny12:
Deprecate use of Anaconda boot arguments without `inst.` prefix. Our intention is to disable these completely in the future.

### DIFF
--- a/anaconda.py
+++ b/anaconda.py
@@ -303,11 +303,11 @@ if __name__ == "__main__":
     if os.path.exists("/tmp/updates"):
         log.info("Using updates in /tmp/updates/ from %s", opts.updateSrc)
 
-    # TODO: uncomment this when we're sure that we're doing the right thing
-    # with kernel_arguments *everywhere* it appears...
-    #for arg in depr:
-    #    stdout_log.warn("Boot argument '%s' is deprecated. "
-    #                   "In the future, use 'inst.%s'.", arg, arg)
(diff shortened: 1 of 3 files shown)


poncovka:
The warning message is very vague. Shouldn't it say that the support will be removed in Fedora 34?

jkonecny12:
Good point. I'll try to change it. However, I would like to avoid setting deadline for that, you never know what could happen.

jkonecny12:
**UPDATED:**
* Improve both deprecation messages.

jkonecny12:
What do you think now @poncovka ?

VladimirSlavik:
I think you might want to change also `boot-options.rst`, especially the following sentence:

> Options specified without the ``inst`` prefix are
> recognized, but the prefix may be required in a future release.

jkonecny12:
> I think you might want to change also `boot-options.rst`, especially the following sentence:
> 
> > Options specified without the `inst` prefix are
> > recognized, but the prefix may be required in a future release.

Good point. I'll do that.

jkonecny12:
**UPDATED:**
* Implemented note from @VladimirSlavik, thanks!

M4rtinK (approved):
Looks good to me. :)

jkonecny12:
It's two weeks without objection. Merging now.

PR #2787, opened 2020-08-13
Reset the state of the custom partitioning spoke

poncovka:
We should reset the state of the custom partitioning spoke after every user
interaction. If we don't reset the _back_already_clicked attribute, the final
check of the storage configuration might be skipped and the status of the
storage spoke might show an error from the previously failed check.

(cherry picked from commit 0aa3abfb6fb137746a393588ff9315807a65f7b9)


poncovka:
Ported from https://github.com/rhinstaller/anaconda/pull/2781.

M4rtinK (approved):
Looks good to me as well. :)

PR #2789, opened 2020-08-17

martinezjavier:


sharkcz:
We should have it in master/Fedora too :-)

jkonecny12:
Jenkins, it's ok to test.

martinezjavier:
> We should have it in master/Fedora too :-)

Indeed, proposed a PR for master as #2790.

martinezjavier:
For some reasons this PR isn't able to find the source anymore, created a new PR #3012.

PR #2790, opened 2020-08-17

martinezjavier:


jkonecny12:
Thanks a lot for this improvement!

jkonecny12:
Jenkins, it's ok to test.

jkonecny12:
From our yesterday conversation it looked like it's pretty hard to test this. So we will only do a sanity check here.

jkonecny12:
jenkins, test this please

jkonecny12:
jenkins, test this please

M4rtinK (approved):
Looks good to me - nice clean-up! :)

jkonecny12:
Tested standard installation and everything seems to work OK.

PR #2792, opened 2020-08-18
Handle exceptions from threads without new instances

VladimirSlavik:
It is not possible to instantiate some exceptions with just an instance as the only argument, for example `UnicodeError` and descendants. However, these days it is possible to raise directly with the provided instance, no need to instantiate the class. The original came to be so due to incrementally rewriting the python2 3-argument form of `raise`. See previous commits affecting this line, in chronological order: 07b7034, d16512e, a6085b8.

Resolves: [rhbz#1835027](https://bugzilla.redhat.com/show_bug.cgi?id=1835027)

### DIFF
--- a/pyanaconda/threading.py
+++ b/pyanaconda/threading.py
@@ -168,7 +168,7 @@ class ThreadManager(object):
         with self._errors_lock:
             exc_info = self._errors.pop(name)
         if exc_info:
-            raise exc_info[0](exc_info[1]).with_traceback(exc_info[2])
+            raise exc_info[1]
 
     def in_main_thread(self):
         """Return True if it is run in the main thread."""


VladimirSlavik:
I tested this (briefly) by appending raising of the following various exceptions to `pyanaconda.modules.storage.reset.ScanDevicesTask._reset_Storage()`:

```
        raise UnicodeDecodeError("blah", b"\x00\x00", 0, 1, "testing exception handling")
        raise UnusableStorageError("just testing")
        raise TypeError("hello from an exception")
```

pyanaconda/threading.py
```diff
@@ -168,7 +168,7 @@ def raise_if_error(self, name):
         with self._errors_lock:
             exc_info = self._errors.pop(name)
         if exc_info:
-            raise exc_info[0](exc_info[1]).with_traceback(exc_info[2])
+            raise exc_info[1].with_traceback(exc_info[2])
```
jkonecny12, line 171: Do we need the `.with_traceback()` part, isn't that already part of the instance?

VladimirSlavik: Ha, nice idea. It didn't occur to me that might be the case. I'll check that!

VladimirSlavik: Yes, that works too! Less code \o/

jkonecny12 (approved):
Looks much simpler. The question is if this won't break some other use-case..., hopefully it won't.

VladimirSlavik:
I think we should be good as-is. tl;dr: Exception handling evolved over the ages, and we have it better than our ancestors.

The following is result of some quick research, so it might not be 100% accurate. Anyway:

The 3-parameter form of `raise` was that way because at some time in the past, exceptions lacked tracebacks stored directly "on" themselves, and also you apparently might have needed a way to cast the exception instance as the right class. Hence the three-tuple returned by `sys.exc_info()` - that gave you the stuff directly reusable with the most explicit form of `raise`.

As stuff modernized, we are now in a place where the instance is good enough, we don't need to cast it, it has the traceback, it has the message, everything is in place.

Finally I am tempted to think that our previous call `raise class(instance).with_traceback(tb)` in fact constructed a new instance, and the original instance was `str()`-ified so as to set the message, thus also losing the traceback...

M4rtinK (approved):
Looks good to me - a nice improvement! :)

poncovka (commented):
Thanks for fixing it!

PR #2795, opened 2020-08-20
network: fix configuration of virtual devices by boot options

rvykydal:
The configuration was not passed to installed system via ifcfg files.

Resolves: rhbz#1851218

This is a port to Fedora.


M4rtinK (approved):
Looks good to me. :)

PR #2797, opened 2020-08-20
Clean up s390 partitioning check (#1855724)

VladimirSlavik:
a) The single_pv property does not exist in blivet since ca. 2004. See blivet's commits e4549cf and b7df921.

b) The complicated restrictions on /boot are better solved by relying on the general check across all architectures.

c) The comments about root were all wrong.

Deleting both that check and comments related to it resolves the issue. Only the LDL check remains now.

Resolves: rhbz#1855724


jstodola:
Tested manually and it works fine, anaconda shows an expected error message when /boot is on LVM (or when no separate /boot exists and / is on LVM). No traceback.
Thanks for the PR, Vlado!

VladimirSlavik:
Nah, thank *you* for the find and suggestions, Honzo :-)

PR #2793, opened 2020-08-18
Avoid unnecessarily pulling in glibc-langpack-en

sgallagher:
Back in 2016[1], this was changed to a strict dependency because the
releng tools of the time could not handle boolean dependencies in
spec files. Nowadays, these work just fine, so we should switch
back to ensuring that we only pull it in when glibc-all-langpacks
is not installed.

Related to https://pagure.io/releng/issue/9689

[1] https://bugzilla.redhat.com/show_bug.cgi?id=1323314

Signed-off-by: Stephen Gallagher <sgallagh@redhat.com>

PR #2798, opened 2020-08-20
Fix dependency_solver failure with spec file boolean logic syntax

jkonecny12:
There is proposed change to have requires in spec file like

``Requires: (... or ...)``

Problem with this is that we take package name with '('. To avoid this problem, remove the brackets before processing and take the first item. Taking the first is not the best solution but we don't have all the info to decide so go with first and hope is our way now. It really should be fine in most of the cases.

This is blocking: https://github.com/rhinstaller/anaconda/pull/2793


VladimirSlavik:
Could Mock take the whole bracket?

PR #2799, opened 2020-08-21
Dracut do not mount source as rw

jkonecny12:
Dracut do not have reason to mount sources as RW it's more that it wasn't specified and until recently we just did not bother. However, we changed the logic that installation sources in stage2 environment does not re-using mounts from the Dracut so this will now fail to set the environment because you can't mount one thing as RW and RO at once.

**DEV NOTE:**
When backporting to Fedora add test that there is no RW mount in Dracut code. Just scan the code for mount without ``-o ro`` should be enough for now.

*Resolves: rhbz#1871049*


jkonecny12:
jenkins, test this please

PR #2800, opened 2020-08-21
Fix traceback when removing additional repository

jkonecny12:
I made a mistake by missing `not` in the condition. That way the repo removal as user action crashed Anaconda because variable which was not set was immediately used.

Unfortunately, I tested this just on the failing use-case and in that case it looked like it is working correctly (but didn't really).

*Resolves: rhbz#1871037*


jkonecny12:
jenkins, test this please

VladimirSlavik (approved):
Looks good to me. Thank you!

PR #2802, opened 2020-08-24
Fix dracut rw mount

jkonecny12:
We changed a logic how to source modules so each module is mounting it's on source to it's own created folder. If anything is mounted as RW in Dracut it will fail to be mounted in sources. This works correctly if all mounts are RO.

This test will check that each mount command is specified with '-o ro' otherwise it have to be ignored.

Backport of https://github.com/rhinstaller/anaconda/pull/2799.

### DIFF
--- a/tests/nosetests/dracut_tests/dracut_source_test.py
+++ b/tests/nosetests/dracut_tests/dracut_source_test.py
@@ -1,0 +1,92 @@
+#
+# Copyright 2020 Red Hat, Inc.
+#
+# This copyrighted material is made available to anyone wishing to use, modify,
+# copy, or redistribute it subject to the terms and conditions of the GNU
+# General Public License v.2.  This program is distributed in the hope that it
+# will be useful, but WITHOUT ANY WARRANTY expressed or implied, including the
+# implied warranties of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+# See the GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License along with
+# this program; if not, write to the Free Software Foundation, Inc., 51
+# Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.  Any Red Hat
+# trademarks that are incorporated in the source code or documentation are not
+# subject to the GNU General Public License and may only be used or replicated
+# with the express permission of Red Hat, Inc.
+
+import unittest
+import os
+import re
+
+import pyanaconda
+
+from os.path import dirname
+from functools import reduce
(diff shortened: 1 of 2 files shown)


jkonecny12:
jenkins, test this please

jkonecny12:
jenkins, test this please

M4rtinK (approved):
Looks good to me. :)

tests/nosetests/dracut_tests/dracut_source_test.py
```diff
@@ -0,0 +1,92 @@
...
+            if os.path.isdir(file_name):
+                print("ignoring directory", file_name)
+                return result
+
+            result.append(file_name)
+            return result
+
+        files = reduce(_filter_ignore_files, files, [])
+
+        mount_line_regex = re.compile(r'\bmount ')
```
VladimirSlavik, line 67: Does this even need a regexp?

VladimirSlavik (approved):
Looks good to me. Thank you!

(Just a few notes...)

tests/nosetests/dracut_tests/dracut_source_test.py
```diff
@@ -0,0 +1,92 @@
...
+    def find_rw_mounts_test(self):
+        # everything what starts by string in this list will not be tested
+        ignore_prefixes = ("Makefile", "README", "python-deps", "module-setup.sh")
+        # define false positives
+        false_positives = (re.compile(r'\bmount +(--move|--make-rprivate)'),
+                           re.compile(r'\bmount /dev/mapper/live-rw'))
+
+        # filter not interesting content
+        comment_regex = re.compile(r'#[^\n]*')
+        log_regex = re.compile(r'\b(info|warn|error) +["][^"]*')
```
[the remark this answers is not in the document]
jkonecny12: Hmm, probably a leftover. Thanks for finding it.

tests/nosetests/dracut_tests/dracut_source_test.py
jkonecny12: Unfortunately yes it needs. The problem is white spaces.
I can do (which I done in earlier versions) just plain ``if "mount" in line`` however, that would also take all ..``something-mount-something` and I can't just use ``if " mount " in line`` because that would not work when there is ``(mount`` or ``^mount``. :(

jkonecny12:
**UPDATE:**
* fixed issue found by @VladimirSlavik . Thanks!

poncovka (approved):
Looks good to me. Thanks!

PR #2803, opened 2020-08-24
Use "raise from" to link exceptions

VladimirSlavik:
Silences all appearances of Pylint error W0707.

In a nutshell, this means both exceptions are printed. This is always in the `except` clause, which printed the other exception anyway. However, the previous message was "_During handling of the above exception, another exception occurred_", which is sort of meant to be for cases where there is a legitimate error. By explicitly linking the exceptions, they are now described as "_The above exception was the direct cause of the following exception_".

At this point we need a discussion: What do we really want to achieve, besides making pylint shut up? Do we want to mask the original traceback somewhere? (`raise * from None`) Do we want to not care?

### DIFF
--- a/pyanaconda/core/configuration/base.py
+++ b/pyanaconda/core/configuration/base.py
@@ -71,7 +71,7 @@ def read_config(parser, path):
             parser.read_file(f, path)
 
     except (configparser.Error, IOError) as e:
-        raise ConfigurationFileError(str(e), path)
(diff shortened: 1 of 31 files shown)


jkonecny12 (approved):
LGTM!

M4rtinK (approved):
LGTMAW! :)

jkonecny12 (approved):
Still looks good.

PR #2804, opened 2020-08-25
Correctly work with package boolean logic in our setup scripts

jkonecny12:
There is support in spec file to have (package1 or package2). This was already fixed by taking the first package by 9c500836d595c0c16cb530efc79cbcf7bf89e26f but there is better way by taking whole brackets.

Add logic to take whole brackets and put that to mock directly.

### DIFF
--- a/scripts/testing/dependency_solver.py
+++ b/scripts/testing/dependency_solver.py
@@ -104,11 +104,9 @@ def runtime_dependencies(spec_content):
     """Find all Requires from spec file."""
     # find all Requires: statements
     # take first word from these statements to avoid these problems:
-    # - Requires: (take_package or not_taken)             # take first from the boolean is the
-    #                                                       best what we can do
-    # - Requires: ( taken_package or not_taken_package )  # work even with space after bracket
+    # - Requires: (take_package or not_taken)             # take the whole (...)
     # - Requires: taken_package >= 3.2.1                  # not taking the version
-    packages = re.findall(r"\n *Requires:[ (]*([^ \n]*)", spec_content)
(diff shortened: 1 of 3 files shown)


scripts/testing/dependency_solver.py
```diff
@@ -104,11 +104,9 @@ def runtime_dependencies(spec_content):
     """Find all Requires from spec file."""
     # find all Requires: statements
     # take first word from these statements to avoid these problems:
-    # - Requires: (take_package or not_taken)             # take first from the boolean is the
-    #                                                       best what we can do
-    # - Requires: ( taken_package or not_taken_package )  # work even with space after bracket
+    # - Requires: (take_package or not_taken)             # take the whole (...)
     # - Requires: taken_package >= 3.2.1                  # not taking the version
-    packages = re.findall(r"\n *Requires:[ (]*([^ \n]*)", spec_content)
+    packages = re.findall(r"\n *Requires: +([^#><=\n]*)", spec_content)
```
VladimirSlavik, line 109: Just an idea: Would the other characters be a problem? If we can happily make the brackets someone else's problem, maybe the rest could be handled the same way?

jkonecny12:
**UPDATE:**
* Fix version specification in spec file.

Answer to question from @VladimirSlavik. It would be great but thanks to the nature that we are using rpm macros to specify version the logic to make this working would be really not trivial. Also the benefit is not that positive, because most of these packages have to be called just because we want to run autogen and configure to be able to use Makefile, however, it does not really have impact on the tests.

VladimirSlavik (approved):
Looks good to me. Thank you!

PR #2808, opened 2020-08-26
subscription: Implement get_source_proxy() in payload base class (port to Fedora)

M4rtinK:
Even though some payloads (such as for example the OSTree payload)
do not have a source proxy, it's better to implement this method
universally. We can just return None for the payloads that don't
have a source.

As the method now can return None, adjust the code that can be using
it with a payload that does not have source proxy to handle this case.

Resolves: rhbz#1861456

### DIFF
--- a/pyanaconda/payload/base.py
+++ b/pyanaconda/payload/base.py
@@ -60,6 +60,17 @@ class Payload(metaclass=ABCMeta):
         """The DBus type of the payload."""
         return None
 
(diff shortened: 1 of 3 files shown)


M4rtinK:
Original RHEL PR: #2758

jkonecny12 (approved):
Looks good to me.

VladimirSlavik (approved):
Looks good to me. Thank you!

PR #2809, opened 2020-08-26
subscription: Convert the RHSM default config values to expected format (port to Fedora)

M4rtinK:
Turns out the RHSM config DBus API GetAll() returns a nested
distionary instead of a flat dict we were expecting, as SetAll()
uses such a flat dict.

Due to this empty values from GUI would be written to the RHSM
config file as no original values would be found due to the
unexpected format of the GetAll() returned dict.

To fix this, convert the nested dict to the flat one we expect &
which is much easier to work with in our use case.

This way the Candlepin URL & CDN URL should no longer be overwritten
by empty values from the GUI.

Resolves: rhbz#1862116

### DIFF
--- a/pyanaconda/modules/subscription/subscription.py
+++ b/pyanaconda/modules/subscription/subscription.py
@@ -556,6 +556,26 @@ class SubscriptionService(KickstartService):
         """
         return self._rhsm_observer
 
+    def _flatten_rhsm_nested_dict(self, nested_dict):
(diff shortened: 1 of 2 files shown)


M4rtinK:
Original RHEL PR: #2762

jkonecny12 (approved):
Looks good to me.

VladimirSlavik (approved):
Looks good to me. Thank you!

rvykydal (approved):
Looks good to me.

PR #2814, opened 2020-08-28
subscription: Assure payload restart on DVD install after registration

M4rtinK:
Make sure that after successful registration on an install from a DVD image
Anaconda does not just switch the installation source to the CDN (as it
has more up to date and secure packages than the on media repos), but
also restarts the payload. Otherwise we can end in an in-between state
with CDN source being assumed, yet most things still pointing to the
on media repo, resulting in an install-time crash.

To fix this, make sure to get a new source proxy - the previously used
one is no longer valid after a source switch.
    
Also fix unit tests accordingly.

Resolves: rhbz#1873164

### DIFF
--- a/pyanaconda/ui/lib/subscription.py
+++ b/pyanaconda/ui/lib/subscription.py
@@ -315,6 +315,10 @@ def register_and_subscribe(payload, progress_callback=None, error_callback=None,
             switch_source(payload, SOURCE_TYPE_CDN)
         # If requested, also restart the payload if CDN is the installation source
         # The CDN either already was the installation source or we just switched to it.
+        #
+        # Make sure to get fresh source proxy as the old one might be stale after
+        # after a source switch.
+        source_proxy = payload.get_source_proxy()
         if restart_payload and source_proxy.Type == SOURCE_TYPE_CDN:
             log.debug("subscription thread: restarting payload after registration")
             _do_payload_restart(payload)
--- a/pyanaconda/ui/lib/subscription.py
+++ b/pyanaconda/ui/lib/subscription.py
@@ -186,6 +186,20 @@ def register_and_subscribe(payload, progress_callback=None, error_callback=None,
     :type progress_callback: callable(subscription_phase)
     :param error_callback: error callback function, takes one argument, the error message
     :type error_callback: callable(error_message)
+    :param bool restart_payload: should payload restart be attempted if it appears necessary ?
+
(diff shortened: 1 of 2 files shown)


VladimirSlavik (approved):
Looks good to me. Thank you!

jkonecny12 (approved):
Looks good to me.

pyanaconda/ui/lib/subscription.py
```diff
@@ -296,12 +310,19 @@ def register_and_subscribe(payload, progress_callback=None, error_callback=None,
...
         # If requested, also restart the payload if CDN is the installation source
-        # The CDN either already was the installation source or we just switched to it.
-        if restart_payload and source_proxy.Type == SOURCE_TYPE_CDN:
+        # or if the source was switched.
+        #
+        # If the source was just switched, then source_proxy.Type will not be updated
+        # until payload refresh has been done. For this reason we also need to check
+        # the value of source_switched variable and restart payload if it is true,
+        # even if CDN does not appear to be the current installation source.
+        if restart_payload and (switched_source or source_proxy.Type == SOURCE_TYPE_CDN):
```
poncovka, line 325: The statement `source_proxy.Type` doesn't return a correct type, because `source_proxy` is a proxy of the previous source. You should call `source_proxy = payload.get_source_proxy()` to work with the new source.

M4rtinK: Oh, so that's why it happens! Still, I wonder if we want to change the code ? Or maybe just fix the documentation but keep the logic, as that has already been tested by RTT & appears to be working fine ?

rvykydal (approved):
Looks good to me.

pyanaconda/ui/lib/subscription.py
jkonecny12: I would prefer correct fix here.

M4rtinK: Hmm, looking into it in more detail, it really looks better, let's do that. :)

poncovka (approved):
Looks good to me. Thanks!

pyanaconda/ui/lib/subscription.py
M4rtinK: Switched to the better fix. :)

jkonecny12 (approved):
Looks even better now!

M4rtinK:
Port to Fedora: #2842

PR #2815, opened 2020-08-30
The underline character should not be displayed

jstodola:
Fix anaconda showing "1 _storage device selected" on the advanced storage spoke.

The call was accidentally? removed in 7b7616f546


VladimirSlavik:
Thank you, looks good to me. However, it would be better to rebase this on f33-devel.

jstodola:
uhh, that's wrong... will try next time

VladimirSlavik:
No that is perfectly ok, as long as you change also the target  branch up there under the name...

jstodola:
https://github.com/rhinstaller/anaconda/pull/2817 created instead.

PR #2819, opened 2020-09-01
Make custom storage summary dialog resizeable (1626555)

M4rtinK:
The dialog that shows the summary of a custom storage spoke session
is rather information dense and not that big by default. So lets enable
decorations, which make it possible for users to grad the window edges
to make it bigger as needed.

Also as we now have the dialog name in the header, drop the label that
was previously holding it.

Resolves: rhbz#1626555

### DIFF
--- a/pyanaconda/ui/gui/spokes/lib/summary.glade
+++ b/pyanaconda/ui/gui/spokes/lib/summary.glade
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<!-- Generated with glade 3.18.1 -->
+<!-- Generated with glade 3.36.0 -->
 <interface>
   <requires lib="gtk+" version="3.2"/>
   <object class="GtkListStore" id="actionStore">
@@ -27,7 +27,8 @@
     <property name="window_position">center-on-parent</property>
     <property name="destroy_with_parent">True</property>
     <property name="type_hint">dialog</property>
-    <property name="decorated">False</property>
+    <property name="deletable">False</property>
+    <property name="has_resize_grip">True</property>
     <child internal-child="vbox">
       <object class="GtkBox" id="dialog-vbox1">
         <property name="can_focus">False</property>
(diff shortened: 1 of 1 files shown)


M4rtinK:
So, this is how it looked like so far:
F33_custom_storage_summary_original
With the change, there is now a header for the dialog window:
F33_custom_storage_summary_v1_1
And user can drag the window edges to make it bigger, like this:
F33_custom_storage_summary_v1_2

poncovka:
Could you paste a screenshot of the dialog, please?

poncovka:
I like how the dialog window looks like, but I think that the design should be consistent for all dialogs. Btw. what does the X button do?

M4rtinK:
@poncovka The `x` button does the same thing as the `cancel` button - it just closes the dialog. BTW, I've managed to find the right magic checkbox in Glade designer & the dialog no longer has the x button:
F33_custom_storage_summary_v2
IMHO, this really looks better than before & does not needlessly duplicate functionality already available elsewhere in the dialog.

I guess if this catches on we might want to apply a similar change to other Anaconda provided dialogs as well.

VladimirSlavik:
Does this mean that the dialog can be now moved around too? It's a small thing but being in control of the GUI makes life a bit easier.

VladimirSlavik (approved):
Looks good to me. Thank you!

rvykydal (approved):
Looks good to me.

M4rtinK:
@VladimirSlavik Yep, you can grab the dialog header and move it as well if you want. :)

jkonecny12 (approved):
Looks good to me too.

VladimirSlavik:
Great! As to changing he rest of the dialogs, this might be worth a moment to discuss with some UX person? I can't see why the dialogs should be headerless in the first place. So there could be some exotic reason...

M4rtinK:
@VladimirSlavik I think we can just proceed with adding headers to all the other dialogs - this one needs it to fix the bug & all utilities we launch (connection editor, yelp help viewer, keyboard layout display) already have a header. So making all dialogs have a header is IMHO a good idea.

As for why dialogs did not have a header before - I'm not sure if we would be able to dig that out, but these are my guesses:
- the idea might have been it looks better, until it was noticed the external utilities will have a header and there is no way around it
- back when the current GTK3 UI was being designed (2012-2013) screen resolutions and sizes were much smaller, usually making a fixed size dialog sufficient, which is I'm afraid no longer sufficient for 2020 with 2k+ 30+ inch screens many users have

PR #2820, opened 2020-09-01
Add support for booting installation media with plain SquashFS

bkhomuts:
This patch allows booting the image even if it's SquashFS images/install.img
does not contain the embedded rootfs.img in it.
In dracut documentation, this is called direct compression of the root filesystem.

In case the plain SquashFS is found, the filesystem is mounted as overlay.
Otherwise, the current method of mounting the /dev/mapper/live-rw applies.

An example of a netinstall image with the plain SquashFS:
https://kojipkgs.fedoraproject.org/compose/rawhide/Fedora-Rawhide-20200828.n.0/compose/Server/x86_64/iso/Fedora-Server-netinst-x86_64-Rawhide-20200828.n.0.iso

The following problem is solved by this change:
```
Starting dracut mount hook...
dracut-mount: mount /sysroot: special device /dev/mapper/live-rw does not exist
Entering emergency mode...
```

The simplest way you can test this change is by interrupting the boot process,
And editing the /usr/lib/anaconda-lib.sh
You can interrupt the boot process by adding rd.break=pre-udev kernel argument.

The plain SquashFS is already supported if the image is Live. But because non-Live images
use a different code path for mounting the root filesystem, this change is needed.

The following patch relates to the following change proposal:
https://fedoraproject.org/wiki/Changes/OptimizeSquashFS

### DIFF
--- a/dracut/anaconda-lib.sh
+++ b/dracut/anaconda-lib.sh
@@ -137,9 +137,20 @@ anaconda_mount_sysroot() {
     local img="$1"
     if [ -e "$img" ]; then
         /sbin/dmsquash-live-root $img
-        # dracut & systemd only mount things with root=live: so we have to do this ourselves
-        # See https://bugzilla.redhat.com/show_bug.cgi?id=1232411
-        printf 'mount /dev/mapper/live-rw %s\n' "$NEWROOT" > $hookdir/mount/01-$$-anaconda.sh
+        if [ -d /run/rootfsbase ]; then
+            # /run/rootfsbase has been created
+            # Which means that the Squash filesystem is plain
+            # and does not contain the embedded EXT4 inside.
+            # Also known as flattened SquashFS or directly compressed SquashFS.
+            printf "mount -t overlay LiveOS_rootfs \
+                   -o lowerdir=/run/rootfsbase,upperdir=/run/overlayfs,workdir=/run/ovlwork \
+                   ${NEWROOT}" > ${hookdir}/mount/01-$$-anaconda.sh
+        else
+            # Otherwise, assumption is that /dev/mapper/live-rw should have been created.
+            # dracut & systemd only mount things with root=live: so we have to do this ourselves
+            # See https://bugzilla.redhat.com/show_bug.cgi?id=1232411
+            printf 'mount /dev/mapper/live-rw %s\n' "$NEWROOT" > $hookdir/mount/01-$$-anaconda.sh
+        fi
     fi
 }
 


poncovka:
Jenkins, test this please.

poncovka (commented):

jkonecny12:
Test fail is solved by https://github.com/rhinstaller/anaconda/pull/2824 .

jkonecny12 (commented):
Looks good to me too. Thanks for this improvement!

jkonecny12:
I see that you have the change set on F33 but it's not accepted. I guess this should be changed to F34 but if you really want to add this to F33 then please change target branch to f33-devel.

bcl (approved):
Looks good to me. I've built a new boot.iso using this and it boots and installs just fine. Also tested building a non-squashfs-only iso which continues to work as before. Thanks!

jkonecny12:
Thanks a lot for testing this @bcl !

jkonecny12:
jenkins, test this please

jkonecny12 (approved):
I believe @bcl tested it properly and the code looks good to me so from my PoV  we are free with the merge here.

bkhomuts:
> I guess this should be changed to F34 but if you really want to add this to F33 then please change target branch to f33-devel.
@jkonecny12 , I'd like this change to be included in both master and Fedora 33 (if possible).

Just in case, I filed a separate pull request for f33-devel: https://github.com/rhinstaller/anaconda/pull/2829

jkonecny12:
The ``f33-devel`` branch is automatically merged to ``master`` so you don't have to have two PR's created.

However, until this is accepted as F33 change I would avoid closing this PR or merging ``f33-devel`` one to avoid confusion on re-opening/closing issues. So for now, let's wait on FESCO and please ping us here when the change will have result.

bkhomuts:
@jkonecny12 , I see. Thanks for explaining.

I'd like to share my perspective: this patch relates https://fedoraproject.org/wiki/Changes/OptimizeSquashFS, but does not depend on it.

Therefore I'd recommend merging this change if the merge window is still open for Fedora 33. This is because the user might try building the image using `--squashfs-only` option of `lorax` or `livemedia-creator`. Without this patch in place, the user will find the resulting ISO unbootable.

Also, this patch does not affect existing behavior of Anaconda, so even if the Change proposal is not accepted in Fedora, nothing will change for the end user.

jkonecny12:
Thanks a lot for explanation. Seems reasonable to me. If it's just for people who creates image manually and not the default I don't see a reason to block this. 

Closing this PR because as explained above the f33 branch will be merged to master so merging both would just create unnecessary commit duplication.

PR #2821, opened 2020-09-02
Propagate verify_ssl to RHSM (port to Fedora)

M4rtinK:
At the moment the inst.noverifyssl boot option and the corresponding
Anaconda config file option verify_ssl only influence SSL certification
validation when anaconda itself downloads some installation related
artifacts.

For consistency this option should be also propagated to the RHSM
install time configuration.

Also this way users that acknowledge the risks by passing the
inst.noverifyssl boot option can successfully register against
a Candlepin instance with certificate that the installation
environment considers unknown or invalid.

Due to how this option is set (effectively overwriting the RHSM config
file) we need to make RHSM SSL certificate validation is turned back ON
at the end of the installation. This is again consistent with how
SSL certificate verification overrides are implemented elsewhere in
Anaconda. The overrides are always install time only and never propagate
to the installed system.

On the impelementation side we rename one of the installation tasks
as it no longer only reverts the log level changes but also makes
sure SSL certification validation is enabled in RHSM on the installed
system. Also change from Set() to SetAll() as se no longer just set a
single value in the two tasks that now interface with the RHSM SSL
certificate validation toggle.

Resolves: rhbz#1858126
(cherry picked from commit b7ca4c14061bd812d2baae1615b6835d1956d6ba)

### DIFF
--- a/pyanaconda/modules/subscription/initialization.py
+++ b/pyanaconda/modules/subscription/initialization.py
@@ -39,6 +39,19 @@ class StartRHSMTask(Task):
 
     RHSM_SYSTEMD_UNIT_NAME = "rhsm.service"
 
+    def __init__(self, verify_ssl=True):
+        """Create a new task for starting the RHSM DBus service.
+
+        :param bool verify_ssl: True if RHSM should be configured to verify SSL certificates,
+                                False if RHSM should be set to *not* verify SSL certificates
+
+        NOTE: If RHSM SSL verification is disabled, this is install time only, as we will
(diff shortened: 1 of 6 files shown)


poncovka:
Please, fix the unit tests.

M4rtinK:
jenkins, test this please

M4rtinK:
@poncovka Tests fixed. :)

M4rtinK:
RHEL PR: #2784

VladimirSlavik (approved):
Looks good to me. Thank you!

PR #2823, opened 2020-09-02
Propagate a lazy proxy of the storage model

poncovka:
The storage model for the partitioning modules should be always created lazily.
When we reset the partitioning, the new model shouldn't be created until we try
to work with it. Then we create a new copy of the storage model, hide disks that
are not selected and possibly initialize empty disks.

There is a problem with modules that need to be able to work with the same copy
of the storage model as the partitioning modules. At this moment, it is only the
device tree module. The suggested solution is to propagate a lazy proxy of the
storage model. It will not trigger the creation of the copy until we try to
access the attributes of the storage model.

Basically, the device tree module always gets the storage model on demand from
the storage property of the partitioning module.

Related: rhbz#1868577

### DIFF
--- a/pyanaconda/core/util.py
+++ b/pyanaconda/core/util.py
@@ -1453,3 +1453,40 @@ def is_smt_enabled():
     except (IOError, ValueError):
         log.warning("Failed to detect SMT.")
         return False
+
+
+class LazyObject(object):
+    """The lazy object."""
+
+    def __init__(self, getter):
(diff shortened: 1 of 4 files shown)


jkonecny12 (commented):
Looks good to me. It just feel fragile. Will this work even if some thread will be slower or similar situations?

jkonecny12 (approved):
Thanks a lot for the explanation. You are free to merge this.

PR #2826, opened 2020-09-03
Always clear treeinfo metadata (#1872056)

jkonecny12:
Metadata from the treeinfo were loaded only during the load of new metadata. However, this does not work if we have source without metadata (e.g. mirrorlist). In that case we are loading additional repositories from the old metadata (not mounted anymore) and not the new ones which may have unexpected results.

*Resolves: rhbz#1872056*


poncovka (approved):
Looks good to me. Thanks!

PR #2827, opened 2020-09-03
Differentiate between RAID levels of a device and its container

poncovka:
The current logic returned the same RAID level for the device and its container,
but we expect that only one of them will have the RAID level set.

If the current device type is not a container device type, don't generate
container data for the device factory request.

### DIFF
--- a/pyanaconda/modules/storage/partitioning/interactive/utils.py
+++ b/pyanaconda/modules/storage/partitioning/interactive/utils.py
@@ -696,18 +696,39 @@ def get_device_raid_level(device):
     if hasattr(device, "data_level"):
         return device.data_level
 
-    if hasattr(device, "volume"):
-        return device.volume.data_level
-
-    if not hasattr(device, "vg") and hasattr(device, "lvs") and len(device.parents) == 1:
-        return get_device_raid_level(device.parents[0])
-
     return None
 
 
 def get_device_raid_level_name(device):
     """Get the RAID level name of the given device."""
     raid_level = get_device_raid_level(device)
+    return raid_level.name if raid_level else ""
+
+
+def get_container_raid_level(container):
+    """Get the RAID level of the given container.
+
+    :param container: a container
+    :return: a RAID level
+    """
+    # Try to get a RAID level of this device.
+    raid_level = get_device_raid_level(container)
+
+    if raid_level:
+        return raid_level
+
+    device = container.raw_device
+
+    # Or get a RAID level of the LVM container.
+    if hasattr(device, "lvs") and len(device.parents) == 1:
+        return get_container_raid_level(device.parents[0])
+
+    return None
+
+
+def get_container_raid_level_name(device):
+    """Get the RAID level name of the given container."""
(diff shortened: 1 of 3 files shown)


pyanaconda/modules/storage/partitioning/interactive/utils.py
```diff
@@ -711,6 +705,31 @@ def get_device_raid_level_name(device):
...
+    :return: a RAID level
+    """
+    raid_level = get_device_raid_level(container)
+
+    if raid_level:
+        return raid_level
+
+    device = container.raw_device
+
+    if hasattr(device, "lvs") and len(device.parents) == 1:
```
jkonecny12, line 721: Could you please make this condition more self-explanatory?

jkonecny12 (approved):
Looks good to me otherwise. I just wonder, this looks like a bugfix but I don't see bug attached in these commits.

pyanaconda/modules/storage/partitioning/interactive/utils.py
poncovka: This is copy-pasted from `get_device_raid_level`. @vojtechtrefny, is there a simpler way to implement `get_container_raid_level`?

poncovka:
> Looks good to me otherwise. I just wonder, this looks like a bugfix but I don't see bug attached in these commits.

It doesn't cause any issues at this moment, because GUI is able to deal with it. However, the logs were confusing and it could eventually start to cause problems.

VladimirSlavik (approved):
Looks good to me, thank you!

pyanaconda/modules/storage/partitioning/interactive/utils.py
jkonecny12: I didn't meant to change the whole algorithm. I was just thinking about something like:
```
is_lvm_container = hasattr(device, "lvs") and len(device.parents) == 1

if is_lvm_container:....
```

The point is that it's not easy to know what you are trying to find out by looking on number of parents and if there is "lvs" attribute.

jkonecny12:
> > Looks good to me otherwise. I just wonder, this looks like a bugfix but I don't see bug attached in these commits.
> 
> It doesn't cause any issues at this moment, because GUI is able to deal with it. However, the logs were confusing and it could eventually start to cause problems.

Thanks for explanation!

pyanaconda/modules/storage/partitioning/interactive/utils.py
poncovka: I have added comments.

poncovka:
Added some comments.

pyanaconda/modules/storage/partitioning/interactive/utils.py
VladimirSlavik: What if it's a btrfs subvolume?

poncovka: That should be covered by `get_device_raid_level`.

M4rtinK (approved):
Looks good to me.

PR #2828, opened 2020-09-04
network: fix using of values of NMConnectionType enum

rvykydal:


jkonecny12 (commented):
I would rather see change on the nmcli side that they would always return the enum and avoid spreading strings where it's not necessary. However, I would not block on that.

VladimirSlavik (approved):
Does this fix the crash on startup? If so, I guess we want this ASAP, and then think about the rest... Otherwise +1 to what Jirka wrote.

pyanaconda/modules/network/nm_client.py
```diff
@@ -213,7 +213,7 @@ def _update_bond_connection_from_ksdata(connection, network_data):
     :type network_data: pykickstart NetworkData
     """
     s_con = connection.get_setting_connection()
-    s_con.props.type = NMConnectionType.BOND
+    s_con.props.type = NMConnectionType.BOND.value
```
poncovka, line 216: I was a little worried that the enum might not work as expected. However, I don't think the new solution is safe. It will be too easy to forget to specify the value. I suggest to remove the enum and use the constants instead (for example, `NM_CONNECTION_TYPE_BOND`).

poncovka: Btw. you can also try to define the enum as `class NMConnectionType(str, Enum)`. Then the members of the enum are really strings, but I would still prefer to use the constants. https://docs.python.org/3/library/enum.html#others

rvykydal: The main reason for using Enum was that I didn't want to have to import all the constants.

> I was a little worried that the enum might not work as expected. However, I don't think the new solution is safe. It will be too easy to forget to specify the value. I suggest to remove the enum and use the constants instead (for example, `NM_CONNECTION_TYPE_BOND`).

The main reason for using Enum was that I didn't want to import all the constants.

rvykydal: > Btw. you can also try to define the enum as `class NMConnectionType(str, Enum)`. Then the members of the enum are really strings, but I would still prefer to use the constants. https://docs.python.org/3/library/enum.html#others

I was considering this as well, but assignment would still need using the .value
I think I will use just the constants.

rvykydal: Added commit replacing enum with constants.

VladimirSlavik (approved):
Still LGTM

jkonecny12 (approved):
Looks good to me now.

poncovka (approved):
Looks good to me. Thank you!

rvykydal:
jenkins, test this please

PR #2829, opened 2020-09-04
Add support for booting installation media with plain SquashFS (Fedora 33)

bkhomuts:
This patch allows booting the image even if it's SquashFS images/install.img
does not contain the embedded rootfs.img in it.
In dracut documentation, this is called direct compression of the root filesystem.

In case the plain SquashFS is found, the filesystem is mounted as overlay.
Otherwise, the current method of mounting the /dev/mapper/live-rw applies.

An example of a netinstall image with the plain SquashFS:
https://kojipkgs.fedoraproject.org/compose/rawhide/Fedora-Rawhide-20200828.n.0/compose/Server/x86_64/iso/Fedora-Server-netinst-x86_64-Rawhide-20200828.n.0.iso

The following problem is solved by this change:

Starting dracut mount hook...
dracut-mount: mount /sysroot: special device /dev/mapper/live-rw does not exist
Entering emergency mode...
The simplest way you can test this change is by interrupting the boot process,
And editing the /usr/lib/anaconda-lib.sh
You can interrupt the boot process by adding rd.break=pre-udev kernel argument.

The plain SquashFS is already supported if the image is Live. But because non-Live images
use a different code path for mounting the root filesystem, this change is needed.

The following patch relates to the following change proposal:
https://fedoraproject.org/wiki/Changes/OptimizeSquashFS

This change is the same as https://github.com/rhinstaller/anaconda/pull/2820, but targets the f33-devel branch.

### DIFF
--- a/dracut/anaconda-lib.sh
+++ b/dracut/anaconda-lib.sh
@@ -137,9 +137,20 @@ anaconda_mount_sysroot() {
     local img="$1"
     if [ -e "$img" ]; then
         /sbin/dmsquash-live-root $img
-        # dracut & systemd only mount things with root=live: so we have to do this ourselves
-        # See https://bugzilla.redhat.com/show_bug.cgi?id=1232411
-        printf 'mount /dev/mapper/live-rw %s\n' "$NEWROOT" > $hookdir/mount/01-$$-anaconda.sh
+        if [ -d /run/rootfsbase ]; then
+            # /run/rootfsbase has been created
+            # Which means that the Squash filesystem is plain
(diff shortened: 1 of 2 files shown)


jkonecny12:
Jenkins, it's ok to test.

poncovka:
The PR is blocked by the Fedora 33 change (see https://github.com/rhinstaller/anaconda/pull/2820#issuecomment-688164328).

poncovka (approved):
Looks good to me.

PR #2830, opened 2020-09-04
Do not push pot files just tell user that he should update

jkonecny12:
Reason for this is that pushing something is pretty hard thing to do. We need to have git created and set. This complicates automation a lot.

Adding new pot-update-check with similar code which will just tell you that the pot file is prepared for update.


poncovka (approved):
Looks good to me. Thanks!

M4rtinK (approved):
Looks good to me as well. :)

PR #2833, opened 2020-09-07
Apply onboot policy even when network was configured in UI.

rvykydal:
Resolves: rhbz#1856632

Port from rhel-8


poncovka (approved):
Looks good to me.

jkonecny12 (approved):
LGTM! I just wonder why this condition was here on the first place? Did the policy for this changed?

VladimirSlavik (approved):

PR #2836, opened 2020-09-07
Add the DBus method IsDeviceShrinkable (#1875677)

poncovka:
Replace the DBus method IsDeviceResizable with IsDeviceShrinkable and fix its
implementation. A shrinkable device has to be resizable and its minimal size
has to be lower then the current size. This should fix the issue with XFS, that
is resizable, but not shrinkable.

Resolves: rhbz#1875677


VladimirSlavik (approved):
Looks good to me. Thank you!

(Does this mean we assume all devices are growable?)

poncovka:
> (Does this mean we assume all devices are growable?)

This change affects only DBus support for the resize dialog. In this dialog, devices can be shrunken or removed, so it doesn't matter whether they are growable. We don't really care.

M4rtinK (approved):
Looks good to me as well. :)

PR #2837, opened 2020-09-07
Never convert translated strings to uppercase

poncovka:
The conversion might produce invalid strings in some languages, for example in Georgian.

Resolves: rhbz#1619530

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -132,6 +132,7 @@ In general we are trying to be as close as possible to `PEP8 <https://www.python
 * Use ``raise`` & ``return`` in the doc string. Do not use ``raises`` or ``returns``.
 * Methods that return a task should have the suffix ‘_with_task’ (for example discover_with_task and DiscoverWithTask).
 * Prefer to use ``pyanaconda.util.join_paths`` over ``os.path.join``. See documentation for more info.
+* Never call ``upper()`` on translated strings. See the bug `1619530 <https://bugzilla.redhat.com/show_bug.cgi?id=1619530>`_
 
 Merging examples
 ----------------
--- a/anaconda.spec.in
+++ b/anaconda.spec.in
@@ -27,7 +27,7 @@ Source0: %{name}-%{version}.tar.bz2
 %define gtk3ver 3.22.17
 %define helpver 22.1-1
 %define isomd5sumver 1.0.10
-%define langtablever 0.0.49
+%define langtablever 0.0.53
 %define libarchivever 3.0.4
 %define libblockdevver 2.1
 %define libreportanacondaver 2.0.21-1
--- a/pyanaconda/core/util.py
+++ b/pyanaconda/core/util.py
@@ -972,6 +972,8 @@ def upcase_first_letter(text):
     lowercases all the others. string.title() capitalizes all words in the
     string.
 
+    Note: Never use on translated strings!
+
     :type text: str
     :return: the given text with the first letter upcased
     :rtype: str
--- a/pyanaconda/localization.py
+++ b/pyanaconda/localization.py
@@ -301,8 +301,7 @@ def get_native_name(locale):
     """
     raise_on_invalid_locale(locale)
 
-    name = langtable.language_name(languageId=locale)
-    return upcase_first_letter(name)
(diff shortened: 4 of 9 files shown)


poncovka:
Before:

georgia_01

After:

georgia_02

VladimirSlavik (approved):
Looks good to me, both code and idea. I agree that Anaconda is not the special snowflake to make things right here.

I'm a bit concerned about the impact of this - users will just see things suddenly became ugly. Will langtable or someone else need to capitalize their data?

poncovka:
@mike-fabian, we would like to stop converting the first letters of language and locale names to uppercase. It means that we will show these strings as they are in the langtable. It looks like some of them are capitalized and some of them are not. Is it expected or could it be unified?

mike-fabian:
> @mike-fabian, we would like to stop converting the first letters of language and locale names to uppercase. It means that we will show these strings as they are in the langtable. It looks like some of them are capitalized and some of them are not. Is it expected or could it be unified?

It is intentional as in some languages these start *always* with an uppercase letter and in some languages they do not.
When it appears "standalone" or at the beginning of a sentence, it seems that the first letter should be in uppercase almost always with the single exception of Georgian where uppercase  seems to be not common even in that case. 

See https://bugzilla.redhat.com/show_bug.cgi?id=1619530 for Georgian.

I think  these values returned by langtable are used by Anaconda never inside a longer sentence, they are always used standalone or at the beginning of a sentence, right?

So I could change langtable to return them with uppercase in the first letter except for Georgian.

I.e. I could return the form useful for using it standalone in a list like the list offered by anaconda.
I could maybe add an optional keyword option "standalone=True" (would there be a better name for such an option?). Then the names would be returned by default in a form suitable for use in a list and one could still use standalone=False if one really wants the original value from CLDR to use it in a sentence.

Would that help?

mike-fabian:
I think it is  probably better to return the right thing from langtable and not make anaconda do the upper casing.

jkonecny12:
> I think it is probably better to return the right thing from langtable and not make anaconda do the upper casing.
> 
> Also, I cannot really think of a use case for using the returned names in a longer context where capitalizing them might be wrong for some languages. So I think I should make a change to returned the names capitalized (except for Georgian). And if we later find a use case where capitalizing is wrong, I could add a keyword option later.

Agree. Thanks a lot for this!

jkonecny12 (approved):
Looks good to me with combination of Langtable proposed change!

poncovka:
> I think it is probably better to return the right thing from langtable and not make anaconda do the upper casing.
> 
> Also, I cannot really think of a use case for using the returned names in a longer context where capitalizing them might be wrong for some languages. So I think I should make a change to returned the names capitalized (except for Georgian). And if we later find a use case where capitalizing is wrong, I could add a keyword option later.

Sounds good to me!

mike-fabian:
https://github.com/mike-fabian/langtable/releases/tag/0.0.53

mike-fabian:
https://bodhi.fedoraproject.org/updates/FEDORA-2020-6426385750

poncovka:
Tested with langtable-0.0.53-1.fc33:

georgia_03

poncovka:
The spec file is updated.

VladimirSlavik (approved):

jkonecny12 (approved):
Looks great now. Thanks for your help @mike-fabian !

poncovka:
Jenkins, test this please.

VladimirSlavik:
It's the test for čeština(česko) that fails on the now lowercase letter.

poncovka:
Jenkins, test this please.

mike-fabian:
> It's the test for čeština(česko) that fails on the now lowercase letter.

That is fixed now, right?

VladimirSlavik:
> That is fixed now, right?

Yes, apparently the test ran still without your changes then :(

poncovka:
> It's the test for čeština(česko) that fails on the now lowercase letter.

The test failed only on Fedora 33 because of the beta freeze. I have uploaded the RPM with the new version of langtable to our repository that we use for these cases and that fixed the unit test. We can drop this workaround after the freeze.

It means that all our tests use the new langtable.

PR #2838, opened 2020-09-08
Improve the error dialog for storage reset

poncovka:
The storage reset might fail because of a problem with the existing storage
configuration or the initial settings (such as a kickstart file). Change the
message in the error dialog for storage reset to cover both cases.

Related: rhbz#1866243

### DIFF
--- a/pyanaconda/errors.py
+++ b/pyanaconda/errors.py
@@ -116,8 +116,8 @@ class ErrorHandler(object):
         return ERROR_RAISE
 
     def _storage_reset_handler(self, exn):
-        message = (_("There is a problem with your existing storage "
-                     "configuration. "
+        message = (_("There is a problem with your existing storage configuration "
+                     "or your initial settings, for example a kickstart file. "
                      "You must resolve this matter before the installation can "
                      "proceed. There is a shell available for use which you "
                      "can access by pressing ctrl-alt-f1 and then ctrl-b 2."


poncovka:
@vojtechtrefny

VladimirSlavik (approved):
Looks good to me. Thank you for caring for these details, my documentation alter ego totally approves.

pyanaconda/errors.py
```diff
@@ -117,7 +117,7 @@ def _storage_install_handler(self, exn):
 
     def _storage_reset_handler(self, exn):
         message = (_("There is a problem with your existing storage "
-                     "configuration. "
+                     "configuration or your initial settings. "
```
M4rtinK, line 120: "initial settings" seems a bit ambiguous to me to what initial settings refers to - intial settings of the *installation* or something else ?

So I think something like "initial installation settings" or even "initial settings (kickstart, etc.)" could be more clear.

VladimirSlavik: Hm, true... How is this message displayed? It's the "some errors with your setup" yellow bar with "click here for details", rigth? If so, there's enough space so I would go for explicit list of places things might be specified.

vojtechtrefny: Message from blivet is also part of the error message so we can be more specific there (feel free to suggest a better message in the blivet PR https://github.com/storaged-project/blivet/pull/896).

poncovka:
This is the dialog with the new blivet's error message:

reset_error

VladimirSlavik:
Thank you! In that context, "initial settings" really is too open to interpretation. Can it be anything else than kickstart and boot options?

But then, I find this a very superficial problem. Right in this very example, the message gives details about what is wrong, but the actual devices involved are not mentioned at all. Good luck user, you're on your own anyway... :-/

poncovka:
> Thank you! In that context, "initial settings" really is too open to interpretation. Can it be anything else than kickstart and boot options?

The exclusive and ignored disks are set via a DBus module, so at the moment of error, we have no idea who set the DBus module. However, we can guess: `There is a problem with your existing storage configuration or your initial settings, for example a kickstart file.`

poncovka:
Updated.

VladimirSlavik (approved):
Looks good to me. Thank you!

PR #2839, opened 2020-09-08
Check if original partitions are mounted too

VladimirSlavik:
Resolves: rhbz#1751698

### DIFF
--- a/pyanaconda/modules/storage/checker/utils.py
+++ b/pyanaconda/modules/storage/checker/utils.py
@@ -417,11 +417,15 @@ def verify_luks2_memory_requirements(storage, constraints, report_error, report_
 def verify_mounted_partitions(storage, constraints, report_error, report_warning):
     """ Check the selected disks to make sure all their partitions are unmounted.
 
+    Check both the currently known and original partitions.
+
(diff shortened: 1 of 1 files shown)


VladimirSlavik:
This could be also moved to master (34) only, if it's the wrong time to break things...

pyanaconda/modules/storage/checker/utils.py
```diff
@@ -429,15 +431,30 @@ def verify_mounted_partitions(storage, constraints, report_error, report_warning
         if not disk.partitioned:
             continue
 
-        for part in disk.format.partitions:
-            part_dev = storage.devicetree.get_device_by_path(part.path)
+        already_checked_partitions = set()
```
poncovka, line 434: It might be easier to first collect a set of partitions that need to be checked and then check these partitions in another loop. This solution is a little hard to follow.

VladimirSlavik: Done...

poncovka (approved):
Looks good to me. Thanks!

jkonecny12 (approved):
Looks good to me.

PR #2777, opened 2020-08-07
subscription: Only restart payload when needed

M4rtinK:
Only restart the payload after registration/unregistration
when necessary - eq. when CDN is in use or when CDN overrides
another installation source (CDROM).

Do not restart the payload in cases when it is not needed,
such as when the URL installation source is in use.

To do this, move the payload restart logic to the subscription
helper thread, and make it conditional, as payload restart is
currently not needed during registration triggered by an
automated kickstart installation, only during manual interaction
in the GUI.

Resolves: rhbz#1861351

PR #2842, opened 2020-09-09
subscription: Payload restart fixes

M4rtinK:
This backport PR combined the following RHEL8 PRs:

**#2777: Only restart payload when needed**

Only restart the payload after registration/unregistration
when necessary - eq. when CDN is in use or when CDN overrides
another installation source (CDROM).

Do not restart the payload in cases when it is not needed,
such as when the URL installation source is in use.

To do this, move the payload restart logic to the subscription
helper thread, and make it conditional, as payload restart is
currently not needed during registration triggered by an
automated kickstart installation, only during manual interaction
in the GUI.

Resolves: rhbz#1861351

**#2814 : Assure payload restart on DVD install after registration**

Make sure that after successful registration on an install from a DVD image
Anaconda does not just switch the installation source to the CDN (as it
has more up to date and secure packages than the on media repos), but
also restarts the payload. Otherwise we can end in an in-between state
with CDN source being assumed, yet most things still pointing to the
on media repo, resulting in an install-time crash.

To fix this, make sure to get a new source proxy - the previously used
one is no longer valid after a source switch.

Also fix unit tests accordingly.

Resolves: rhbz#1873164


VladimirSlavik (approved):
Looks good to me. Thank you!

poncovka (approved):
Looks good to me.

pyanaconda/ui/lib/subscription.py
```diff
@@ -315,6 +315,10 @@ def register_and_subscribe(payload, progress_callback=None, error_callback=None,
             switch_source(payload, SOURCE_TYPE_CDN)
         # If requested, also restart the payload if CDN is the installation source
         # The CDN either already was the installation source or we just switched to it.
+        #
+        # Make sure to get fresh source proxy as the old one might be stale after
+        # after a source switch.
```
jkonecny12, line 320: Typo ``after after``.

jkonecny12 (approved):
Looks good to me otherwise.

pyanaconda/ui/lib/subscription.py
M4rtinK: Fixed! :)

PR #2847, opened 2020-09-10
Create the RPM OSTree DBus module

poncovka:
Create the basic structure for the DBus support of the RPM OSTree payload.


pyanaconda/core/constants.py
```diff
@@ -443,6 +443,7 @@ class PayloadRequirementType(Enum):
 SOURCE_TYPE_URL = "URL"
 SOURCE_TYPE_HDD = "HDD"
 SOURCE_TYPE_CDN = "CDN"
+SOURCE_TYPE_OSTREE = "OSTREE"
```
jkonecny12, line 446: I think this should be named ``RPM_OSTREE`` not just ``OSTREE``. One reason for that is payload is named ``rpm_ostree`` and second reason is because ostree is something different than rpm-ostree. 

Also I'm thinking if this shouldn't be named ``RPM_OSTREE_REMOTE`` and also add ``RPM_OSTREE_LOCAL`` because source purpose should be to distinct between different sources and for local image we don't need most of what is here.

jkonecny12 (changes requested):
Looks good to me but let's discuss the current approach before merge please.

pyanaconda/core/constants.py
poncovka: I have used `OSTREE` for the source, because it reflects the kickstart command `ostreesetup` and its documentation mentions only OSTree. https://pykickstart.readthedocs.io/en/latest/kickstart-docs.html#ostreesetup

Also, I don't understand the code enough to split the source to `REMOTE` and `LOCAL`. Pykickstart sets `remote` to `osname`, if `remote` is not set, and the RPMOSTreePayload class happily runs all commands for `remote` in all cases.

jkonecny12: > I have used `OSTREE` for the source, because it reflects the kickstart command `ostreesetup` and its documentation mentions only OSTree. https://pykickstart.readthedocs.io/en/latest/kickstart-docs.html#ostreesetup
> 

Yes I know, but I don't think we should follow here. As I said ostree is something different than rpm ostree.

> Also, I don't understand the code enough to split the source to `REMOTE` and `LOCAL`. Pykickstart sets `remote` to `osname`, if `remote` is not set, and the RPMOSTreePayload class happily runs all commands for `remote` in all cases.

It would be more about url parameter than remote. From my understanding remote is always set.

VladimirSlavik (approved):
Looks good to me. Thank you!

pyanaconda/core/constants.py
poncovka: The source was renamed to `RPM_OSTREE`.

poncovka:
Updated.

VladimirSlavik (commented):
Still LGTM, though some of the @pep8speaks warnings could use attention...

poncovka:
The fixable pep8 issues are solved.

M4rtinK (commented):
Looks good to me.

jkonecny12 (approved):
Looks good to me. Thanks for all the work!

PR #2849, opened 2020-09-10
Clean up the DNF payload class

poncovka:
Move or remove some parts of the DNF payload class.

### DIFF
--- a/pyanaconda/installation.py
+++ b/pyanaconda/installation.py
@@ -18,7 +18,7 @@
 # Red Hat, Inc.
 #
 from pyanaconda.core.configuration.anaconda import conf
-from pyanaconda.core.constants import PAYLOAD_LIVE_TYPES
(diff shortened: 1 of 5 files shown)


jkonecny12 (approved):
Looks good to me.

However, I don't really agree with the last commit to remove ``needs_storage_configuration``. I understand why you are doing this hence I won't block this PR but it's much harder to read the code now.

poncovka:
> However, I don't really agree with the last commit to remove `needs_storage_configuration`. I understand why you are doing this hence I won't block this PR but it's much harder to read the code now.

It is not ideal, but its just a temporary solution. Eventually, this should be solved via dependencies between tasks.

VladimirSlavik (approved):
Looks good to me. Thank you!

poncovka:
Rebased.

PR #2850, opened 2020-09-11
Fix CDN button visibility

M4rtinK:
Only show the Red Hat CDN button if the Subscription
module appears to be running. To achieve that, we do
the same thing as with the HMC button - the CDN
button invisible by default. And enable it only if it
looks like the Subscription module is running.


VladimirSlavik (approved):
Looks good to me. Thank you!

poncovka (approved):
Looks good to me.

PR #2853, opened 2020-09-14
Replace the Packages DBus module with a DBus structure

poncovka:
Add a new DBus structure PackagesConfigurationData and the DBus
property Packages. Implement support for the `%packages` section
in the DNF module. Remove the Packages module.


tests/nosetests/pyanaconda_tests/module_payload_dnf_test.py
```diff
@@ -146,6 +147,166 @@ def url_metalink_kickstart_test(self):
...
+        package1
+        package2
+
+        %end
+        """
+        self.shared_ks_tests.check_kickstart(
+            ks_in, ks_out="", ks_tmp=ks_out
+        )
+
+    def packages_section_with_attribute_kickstart_test(self):
```
VladimirSlavik, line 233: What does this do that the next test does not?

VladimirSlavik (approved):
LGTM, except for a few bits. Thank you!

pyanaconda/modules/common/structures/payload.py
```diff
@@ -215,3 +216,261 @@ def included_packages(self) -> List[Str]:
...
+        """
+        return self._environment
+
+    @environment.setter
+    def environment(self, value: Str):
+        self._environment = value
+
+    @property
+    def groups(self) -> List[Str]:
+        """A list of package groups marked for installation.
```
M4rtinK, line 283: If I understand this correctly, this basically comes from the kickstart packages section, right ? I that case it might be good to note that this might also contain module specification, as they use the same notation in kickstart as groups, for example:

```
@nodejs:6/default
@django:1.6/default
```

I think just noting that in the docstring would be enough, no need to rename the properties IMHO.

pyanaconda/modules/payloads/kickstart.py
```diff
@@ -43,6 +45,13 @@ def handleHeader(self, lineno, args):
             )
 
 
+class AnacondaPackages(Packages):
+    """The representation of the %packages kickstart section."""
+
+    # The class that represents a package group.
+    Group = Group
```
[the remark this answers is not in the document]
M4rtinK: Indeed, this seems strange.

pyanaconda/modules/payloads/payload/dnf/dnf.py
```diff
@@ -91,11 +102,123 @@ def process_kickstart(self, data):
...
+        """Process the kickstart packages."""
+        packages = PackagesConfigurationData()
+        packages.core_group_enabled = not data.packages.nocore
+        packages.default_environment_enabled = data.packages.default
+
+        if data.packages.environment is not None:
+            packages.environment = data.packages.environment
+
+        packages.packages = list(data.packages.packageList)
+        packages.excluded_packages = data.packages.excludedList
```
M4rtinK, line 115: Is `data.packages.excludeList` is different and does not need `list()` ?

pyanaconda/modules/payloads/payload/dnf/dnf.py
```diff
@@ -91,11 +102,123 @@ def process_kickstart(self, data):
...
+        self._set_up_kickstart_packages(data)
+
+    def _set_up_kickstart_sources(self, data):
+        """Set up the kickstart sources."""
         for source in self.sources:
             source.setup_kickstart(data)
 
+    def _set_up_kickstart_packages(self, data):
+        """Set up the kickstart packages."""
+        # The empty packages section won't be printed without seen set to True.
```
M4rtinK, line 166: I'm a bit confused here - do we usually print the `%packages` section if it is empty ? Also, in case setting seen is needed always even for non-empty `%packages` that would seem like a Pykickstart bug.

pyanaconda/modules/payloads/payloads.py
```diff
@@ -114,7 +114,14 @@ def generate_kickstart(self):
             log.debug("Generating kickstart... (skip)")
             return ""
 
-        return super().generate_kickstart()
+        log.debug("Generating kickstart...")
+        handler = self.get_kickstart_handler()
+        self.setup_kickstart(handler)
+
+        # FIXME: This is a temporary workaround for packages.
```
M4rtinK, line 115: I think some information about why and till when the workaround is needed would be useful.

tests/nosetests/pyanaconda_tests/module_payload_dnf_test.py
```diff
@@ -149,6 +147,166 @@ def url_metalink_kickstart_test(self):
...
 
+    def packages_section_empty_kickstart_test(self):
+        """Test the empty packages section."""
+        ks_in = """
+        %packages
+        %end
+        """
+        ks_out = """
+        %packages
+
```
M4rtinK, line 158: We really always used to put a blank line there ? Wow.

tests/nosetests/pyanaconda_tests/module_payload_dnf_test.py
```diff
@@ -149,6 +147,166 @@ def url_metalink_kickstart_test(self):
...
+        %end
+        """
+        ks_out = """
+        %packages
+        @^environment
+        @group
+        @module2:1.5/server
+        @module:10
+        package
+
```
M4rtinK, line 183: Hmm, on non empty as well. Well, if it was like this in the past...

tests/nosetests/pyanaconda_tests/module_payload_dnf_test.py
M4rtinK: I guess the case of just `--nocore` without anything else ?

M4rtinK (approved):
Looks good to me, just some notes in places that were not clear to me. :)

jkonecny12 (commented):
Looks good to me from the structure PoV. However, I was not looking for bugs in the code so I don't giving an ACK here.

pyanaconda/modules/payloads/kickstart.py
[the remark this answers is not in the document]
Conan-Kudo: 🤪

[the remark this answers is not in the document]
poncovka: Replaced with a function.

tests/nosetests/pyanaconda_tests/module_payload_dnf_test.py
poncovka: The tests are reorganized a little.

pyanaconda/modules/payloads/payload/dnf/dnf.py
poncovka: Fixed.

pyanaconda/modules/payloads/payload/dnf/dnf.py
poncovka: There is a logical difference between an empty `%packages` section and a missing `%packages` section. Pykickstart needs to keep the missing section missing during pre-processing of a kickstart file (for example, flattening), so it uses the `seen` attribute to decide whether to generate the section or not. It is ugly, but this is how it works.

pyanaconda/modules/payloads/payloads.py
poncovka: Added another comment.

pyanaconda/modules/common/structures/payload.py
poncovka: Good point. I will have to think it through, so it might be fixed in another pull request.

poncovka:
Updated.

pyanaconda/modules/common/structures/payload.py
jkonecny12: Please add FIXME keyword than.

poncovka: I have updated the doc strings and added examples.

poncovka:
The docstrings of the DBus structure are updated.

jkonecny12 (approved):
Looks great to me. Thanks!

PR #2855, opened 2020-09-15
Import RPM certificates at end of installation

VladimirSlavik:
Resolves: [rhbz#1858006](https://bugzilla.redhat.com/show_bug.cgi?id=1858006)

I think this belongs elsewhere, just not sure where.

### DIFF
--- a/data/anaconda.conf
+++ b/data/anaconda.conf
@@ -93,6 +93,10 @@ default_source = CLOSEST_MIRROR
 # Enable ssl verification for all HTTP connection
 verify_ssl = True
 
+# GPG keys to import to RPM database by default.
+# Specify paths on the installed system, each on a line.
+# Substitutions for $releasever and $basearch happen automatically.
+default_rpm_gpg_keys =
 
 [Security]
 # Enable SELinux usage in the installed system.
--- a/pyanaconda/core/configuration/payload.py
+++ b/pyanaconda/core/configuration/payload.py
@@ -111,3 +111,8 @@ class PayloadSection(Section):
         this option.
         """
         return self._get_option("verify_ssl", bool)
+
+    @property
+    def default_rpm_gpg_keys(self):
+        """List of GPG keys to import into RPM database at end of installation."""
+        return self._get_option("default_rpm_gpg_keys", str).split()
--- a/pyanaconda/core/util.py
+++ b/pyanaconda/core/util.py
@@ -47,6 +47,7 @@ from pyanaconda.core.constants import DRACUT_SHUTDOWN_EJECT, TRANSLATIONS_UPDATE
     IPMI_ABORTED, X_TIMEOUT, TAINT_HARDWARE_UNSUPPORTED, TAINT_SUPPORT_REMOVED, \
     WARNING_HARDWARE_UNSUPPORTED, WARNING_SUPPORT_REMOVED
 from pyanaconda.core.constants import SCREENSHOTS_DIRECTORY, SCREENSHOTS_TARGET_DIRECTORY
+from pyanaconda.core.regexes import OS_RELEASE_OS_VERSION
 from pyanaconda.errors import RemovedModuleError, ExitError
 
 from pyanaconda.anaconda_logging import program_log_lock
@@ -1490,3 +1491,22 @@ class LazyObject(object):
             return super().__setattr__(name, value)
 
         return setattr(self._object, name, value)
+
+
+def get_os_version(sysroot=""):
+    """Find version of the OS from the os-release file.
+
+    See os-release(5).
+
+    :param str sysroot: Where to look.
+    :return str: The version
+    """
+    for filename in ("/etc/os-release", "/usr/lib/os-release"):
+        try:
+            with open(sysroot + filename, "r") as f:
+                data = f.read(4096)  # 4 kB should be enough for everyone!
+            return OS_RELEASE_OS_VERSION.findall(data)[0]
+        except (FileNotFoundError, IndexError):
+            pass
+
+    return None
--- a/pyanaconda/payload/dnf/payload.py
+++ b/pyanaconda/payload/dnf/payload.py
@@ -1986,6 +1986,17 @@ class DNFPayload(Payload):
             if ks_repo.excludepkgs:
                 f.write("exclude=%s\n" % ",".join(ks_repo.excludepkgs))
 
+    def _import_rpm_keys(self):
+        """Import GPG keys to RPM database."""
+        if conf.payload.default_rpm_gpg_keys:
+            # TODO: replace the interpolation with DNF once possible
+            arch = util.execWithCapture("uname", ["-i"]).strip().replace("'", "")
+            vers = util.get_os_version(conf.target.system_root)
+            for key in conf.payload.default_rpm_gpg_keys:
+                interpolated_key = key.replace("$releasever", vers).replace("$basearch", arch)
+                log.info("Importing GPG key to RPM database: %s", interpolated_key)
+                util.execInSysroot("rpm", ["--import", interpolated_key])
+
     def post_setup(self):
         """Perform post-setup tasks.
 
@@ -2024,6 +2035,9 @@ class DNFPayload(Payload):
         # We don't need the mother base anymore. Close it.
         self._base.close()
         super().post_install()
+
+        # rpm needs importing installed certificates manually, see rhbz#748320 and rhbz#185800
+        self._import_rpm_keys()
 
     @property
     def kernel_version_list(self):
--- a/data/product.d/fedora.conf
+++ b/data/product.d/fedora.conf
@@ -20,3 +20,5 @@ default_help_pages =
 
 [Payload]
 default_source = CLOSEST_MIRROR
+default_rpm_gpg_keys =
+    /etc/pki/rpm-gpg/RPM-GPG-KEY-fedora-$releasever-$basearch
--- a/pyanaconda/core/regexes.py
+++ b/pyanaconda/core/regexes.py
@@ -194,3 +194,6 @@ IPV6_ADDRESS_IN_DRACUT_IP_OPTION = re.compile(r'\[[^\]]+\]')
 
 # Octet of MAC address
 MAC_OCTET = re.compile(r'[a-fA-F0-9][a-fA-F0-9]')
+
+# OS version in os-release file
+OS_RELEASE_OS_VERSION = re.compile(r"VERSION_ID=[\"']?([0-9a-z._-]+)[\"']?")
--- a/tests/nosetests/pyanaconda_tests/util_test.py
+++ b/tests/nosetests/pyanaconda_tests/util_test.py
@@ -856,6 +856,36 @@ class MiscTests(unittest.TestCase):
         self.assertEqual(get_anaconda_version_string(), "1.0")
         self.assertEqual(get_anaconda_version_string(build_time_version=True), "1.0-1")
 
+    def get_os_version_test(self):
+        """Test get_os_version."""
+        with tempfile.TemporaryDirectory() as root:
+            # prepare paths
+            util.mkdirChain(root + "/usr/lib")
+            util.mkdirChain(root + "/etc")
+
+            # backup file only
+            with open(root + "/usr/lib/os-release", "w") as f:
+                f.write("# blah\nVERSION_ID=foo256bar  \n VERSION_ID = wrong\n\n")
+            version = util.get_os_version(root)
+            self.assertEqual(version, "foo256bar")
+
+            # main file and backup too
+            with open(root + "/etc/os-release", "w") as f:
+                f.write("# blah\nVERSION_ID=more-important\n")
+            version = util.get_os_version(root)
+            self.assertEqual(version, "more-important")
+
+            # both, main file twice
+            with open(root + "/etc/os-release", "w") as f:
+                f.write("# blah\nVERSION_ID=more-important\nVERSION_ID=not-reached\n \n")
+            version = util.get_os_version(root)
+            self.assertEqual(version, "more-important")
+
+            # no files
+            os.remove(root + "/usr/lib/os-release")
+            os.remove(root + "/etc/os-release")
+            version = util.get_os_version(root)
+            self.assertEqual(version, None)
 
 class LazyObjectTestCase(unittest.TestCase):
 


kparal:
I'd love to have the missing keys problem finally solved. I'd vote for including this in F33 Beta release so that we can test it properly, if you can make a new Anaconda release in time...

pyanaconda/payload/dnf/payload.py
```diff
@@ -2045,6 +2046,18 @@ def post_install(self):
             except PayloadSetupError as e:
                 log.error(e)
 
+        # rpm needs importing installed certificates manually, see rhbz#748320 and rhbz#185800
+        if conf.payload.import_keys_to_rpm:
+            log.info("Importing keys to RPM database.")
+            #releasever = util.execWithCapture("rpm", ["--eval", "'%{fedora}'"])
+            #basearch = util.execWithCapture("uname", ["-i"])
+            util.execInSysroot("rpm", [
+                "--import", "/etc/pki/rpm-gpg/RPM-GPG-KEY-fedora-{releasever}-{basearch}".format(
```
poncovka, line 2055: This is not work for other products. The configuration file should provide paths to gpg keys that should be imported.

pyanaconda/payload/dnf/payload.py
```diff
@@ -2045,6 +2046,18 @@ def post_install(self):
...
+                    releasever=util.execWithCapture("rpm", ["--eval", "'%{fedora}'"]),
+                    basearch=get_arch()
```
poncovka, line 2057: I think you can use `_replace_vars` for the substitution.

poncovka: Then the Fedora configuration file could set the option for the gpg keys to:

`default_gpg_keys = /etc/pki/rpm-gpg/RPM-GPG-KEY-fedora-$releasever-$basearch`

VladimirSlavik:
Update: Now with 100% more `_replace_vars` and it's not as great as could be, because the keys use numbers while DNF substitutes "rawhide". More ideas welcome.

pyanaconda/payload/dnf/payload.py
VladimirSlavik: Thank you - solved...

pyanaconda/payload/dnf/payload.py
VladimirSlavik: Done. See below / diff for the aftermath.

data/product.d/fedora-workstation.conf
```diff
@@ -9,6 +9,7 @@ product_name = Fedora
 
 [Payload]
 default_environment = workstation-product-environment
+import_keys_to_rpm = /etc/pki/rpm-gpg/RPM-GPG-KEY-fedora-$releasever-$basearch
```
Conan-Kudo, line 12: This should be in `fedora.conf`, not `fedora-workstation.conf`.

Conan-Kudo (changes requested):

data/anaconda.conf
```diff
@@ -93,6 +93,8 @@ default_source = CLOSEST_MIRROR
 # Enable ssl verification for all HTTP connection
 verify_ssl = True
 
+# Import keys to RPM database
+import_keys_to_rpm =
```
poncovka, line 97: I would change the name to something like `default_gpg_keys`.

data/anaconda.conf
```diff
@@ -93,6 +93,8 @@ default_source = CLOSEST_MIRROR
 # Enable ssl verification for all HTTP connection
 verify_ssl = True
 
+# Import keys to RPM database
```
poncovka, line 96: It is not clear what the option contains.

pyanaconda/core/configuration/payload.py
```diff
@@ -111,3 +111,8 @@ def verify_ssl(self):
         this option.
         """
         return self._get_option("verify_ssl", bool)
+
+    @property
+    def import_keys_to_rpm(self):
+        """Import keys into RPM database at end of installation."""
```
poncovka, line 117: Again, it is not clear what the property returns.

pyanaconda/core/configuration/payload.py
```diff
@@ -111,3 +111,8 @@ def verify_ssl(self):
...
+        return self._get_option("import_keys_to_rpm", str)
```
poncovka, line 118: It should be a list of keys. Please, call `split()` on the result.

pyanaconda/payload/dnf/payload.py
```diff
@@ -2005,6 +2005,16 @@ def _write_dnf_repo(self, repo, repo_path):
             if ks_repo.excludepkgs:
                 f.write("exclude=%s\n" % ",".join(ks_repo.excludepkgs))
 
+    def _import_rpm_keys(self):
+        """Import keys to RPM database."""
+        if bool(conf.payload.import_keys_to_rpm):
```
poncovka, line 2010: `if conf.payload.import_keys_to_rpm` is enough.

data/product.d/fedora-workstation.conf
VladimirSlavik: Done.

VladimirSlavik:
Next thing to consider: Just one key, or multiple ones?

pyanaconda/payload/dnf/payload.py
```diff
@@ -2005,6 +2005,16 @@ def _write_dnf_repo(self, repo, repo_path):
...
+            key_path = self._replace_vars(conf.payload.import_keys_to_rpm)
+            if "rawhide" in key_path:
+                releasever = util.execWithCapture("rpm", ["--eval", "'%{fedora}'"]).strip().replace("'", "")
+                key_path = key_path.replace("rawhide", releasever)
```
poncovka, line 2014: Maybe `_replace_vars` is not a good solution.

VladimirSlavik:
Note to self: Original inspiration calls `uname -i` while we have access to:

* `os.uname` which contains stuff equivalent to `uname -m`,
* `_replace_vars` which does presumably do something else, and then also
* blivet's `get_arch()` which mangles the python one into something closer to `uname -i`.

"Enjoy the choice!"

pyanaconda/core/configuration/payload.py
VladimirSlavik: Done.

VladimirSlavik:
Update: I have tried to address the feedback. From my PoV, it's good enough...

pyanaconda/payload/dnf/payload.py
```diff
@@ -2005,6 +2005,17 @@ def _write_dnf_repo(self, repo, repo_path):
             if ks_repo.excludepkgs:
                 f.write("exclude=%s\n" % ",".join(ks_repo.excludepkgs))
 
+    def _import_rpm_keys(self):
+        """Import GPG keys to RPM database."""
+        if conf.payload.default_rpm_gpg_keys:
+            arch = util.execWithCapture("uname", ["-i"]).strip().replace("'", "")
+            vers = util.execWithCapture("rpm", ["--eval", "'%{fedora}'"]).strip().replace("'", "")
```
Conan-Kudo, line 2012: I'm pretty sure you can get these things from the DNF API... @dmach, @j-mracek?

VladimirSlavik: I can, but they'll give me "rawhide" instead of "34" and similar niceties, so this seemed really easier.

Conan-Kudo: Yeah, but this doesn't work either, since `%fedora` is not defined everywhere.

mikhailnov: Yeah, what if it is not fedora or rh?

cgwalters: See https://github.com/rpm-software-management/libdnf/issues/43

poncovka, line 2012: This line needs to work for all products.

poncovka (changes requested):

VladimirSlavik:
Update: Now using the dnf api to do the substitutions. The string "rawhide" still needs to be replaced with a number, but that happens only if rawhide is there, so all should be well.

Note that this all happens only if the config file for the product asks for import of keys. If it does not, then this does not happen.

Conan-Kudo:
Why does `rawhide` need to be replaced? That works just as well with importing GPG keys and getting mirrors from MirrorManager2.

VladimirSlavik:
Jenkins, test this please.

VladimirSlavik:
> Why does `rawhide` need to be replaced? That works just as well with importing GPG keys and getting mirrors from MirrorManager2.

Try this: `ls -1 /etc/pki/rpm-gpg | grep hide` See anything? I do not. For reference, you want to find things like `/etc/pki/rpm-gpg/RPM-GPG-KEY-fedora-$releasever-$basearch`. All have numbers. I think it's because rawhide is a pointer/alias, not an actual release number: currently 34, in a few months 35,...

j-mracek:
What about to add gpg key url into repo.gpgkey. Then you don't need any additional import function in anaconda. The game with releasever `rawhide` vs. number like '34' must be resolved by different way, because it is also issue in dnf. But it has to be resolved by distribution, not by hack in dnf/anaconda.

VladimirSlavik:
Update: I think the game some exit actually :)

I also think that what gets lost in the discussion is that this is opt-in - use the config option to import certs, or not.

jkonecny12 (approved):
Looks good to me but I agree with others that the numbering should be read from a correct place to avoid hack and such. However, the ``os-release`` solution seems like a good idea to me.

VladimirSlavik:
Update: regexp now conforms to the spec.

pyanaconda/core/util.py
```diff
@@ -1490,3 +1490,30 @@ def __setattr__(self, name, value):
...
+    except IndexError:
+        pass
+
+    with open(sysroot + "/usr/lib/os-release", "r") as f:
+        data = f.read(4096)
+    try:
+        return _os_version_finder.findall(data)[0]
+    except IndexError:
+        pass
+    return None
```
Conan-Kudo, line 1519: This might be a better way to actually _read_ `os-release(5)` data: https://pagure.io/fedora-rust/rust2rpm/blob/master/f/rust2rpm/__main__.py#_37-44

Conan-Kudo (changes requested):

j-mracek:
I suggest that Fedora distribution should drop to use `rawhide` value for releasever and shim rawhide repository signed only with a single key.
In DNF we have a problem when user want to jump from released Fedora to rawhide. The problem is not only in key `rawhide`, because then you are unable to import gpg_keys, but also that rawhide repository contains packages signed by multiple keys. It means that you have to import key for Fedora 33 and 34/rawhide for proper functionality.

pyanaconda/payload/dnf/payload.py
```diff
@@ -1986,6 +1986,17 @@ def _write_dnf_repo(self, repo, repo_path):
             if ks_repo.excludepkgs:
                 f.write("exclude=%s\n" % ",".join(ks_repo.excludepkgs))
 
+    def _import_rpm_keys(self):
```
j-mracek, line 1989: I think that you don't need this function. You need only properly setup path in DNF repo object.

j-mracek: It could be replaced by `dnf.Base().package_import_key()`. Also check `dnf.Base().package_signature_check()` for detection if import is needed.

VladimirSlavik: This could be far better. Thank you for the pointers!

VladimirSlavik: @j-mracek does this take the certs from chroot or host?

poncovka: @j-mracek, we are trying to import gpg keys that are installed on the target system by rpm packages. I don't see how we could use `dnf.Base()` configured by the installation environment that might use different keys and different repositories than the installed ones.

pyanaconda/core/util.py
```diff
@@ -1490,3 +1490,30 @@ def __setattr__(self, name, value):
             return super().__setattr__(name, value)
 
         return setattr(self._object, name, value)
+
+
+_os_version_finder = re.compile(r"VERSION_ID=[\"']?([0-9a-z._-]+)[\"']?")
```
poncovka, line 1495: Please, move this line to `pyanaconda.core.regexes`.

pyanaconda/core/util.py
```diff
@@ -1490,3 +1490,30 @@ def __setattr__(self, name, value):
...
+
+
+def get_os_version(sysroot=""):
```
poncovka, line 1498: Could you write some unit tests for this function, please?

pyanaconda/core/util.py
```diff
@@ -1490,3 +1490,30 @@ def __setattr__(self, name, value):
...
+    """
+    try:
+        with open(sysroot + "/etc/os-release", "r") as f:
+            data = f.read(4096)  # 4 kB should be enough for everyone!
+        return _os_version_finder.findall(data)[0]
+    except (FileNotFoundError, IndexError):
+        pass
+
+    try:
+        with open(sysroot + "/usr/lib/os-release", "r") as f:
```
poncovka, line 1514: It looks like a duplicate of the code above. You can have a loop that iterates over the file names and returns the first usable result or None.

pyanaconda/payload/dnf/payload.py
```diff
@@ -1986,6 +1986,17 @@ def _write_dnf_repo(self, repo, repo_path):
...
+        """Import GPG keys to RPM database."""
+        if conf.payload.default_rpm_gpg_keys:
+            arch = util.execWithCapture("uname", ["-i"]).strip().replace("'", "")
+            vers = util.get_os_version(conf.target.system_root)
+            for key in conf.payload.default_rpm_gpg_keys:
+                interpolated_key = key.replace("$releasever", vers).replace("$basearch", arch)
+                log.info("Importing GPG key to RPM database: %s", interpolated_key)
+                util.execInSysroot("rpm", ["--import", interpolated_key], None,
+                                   conf.target.system_root)
```
poncovka, line 1998: Use can remove `None` and `conf.target.system_root` from the arguments. These are defaults.

pyanaconda/payload/dnf/payload.py
j-mracek: I suggest that keys used for verification are taken from RPM_DB from chroot and also imported to the same DB. The location of source keys for import are taken from repo.gpgkey (see all repos from base.repos). Paths or urls in repo.gpgkey are not related or modified according to chroot.

poncovka (changes requested):
The code requires some small fixes. Otherwise, it looks good to me.

pyanaconda/payload/dnf/payload.py
j-mracek: > eys that are installed on the target system by rpm packages. I don't see how we could use `dnf.Base()` configured by the installation environment that might use different keys and different

We can discuss it on call if you like. In case that you want to install, upgrade packages, rpm verify it if the gpg key is valid and inserted in rpm_db. DNF provide a functionality how to verify and import missing gpg keys for a package. You only need to set properly path in dnf.repo.Repo() class using urllist attribute `gpgkey`.

Both methods `dnf.Base().package_import_key()` and `dnf.Base().package_signature_check()` are brand new APIs to allow user of DNF (anaconda, dnfdeamon) to not have their own implementation of gpg import mechanism.

VladimirSlavik:
Addressed the code feedback above, tests would come after the direction is confirmed.

VladimirSlavik:
Update: tests.

poncovka (approved):
Looks good to me. Thanks!

PR #2856, opened 2020-09-15
network: clone connections from intramfs to persistent config

rvykydal:
In case of common connection ("Wired connection") being used for multiple
devices in initramfs (which happens in case of ip=dhcp or no ip= specified) the
connection was not used for persistent configuration of the devices and instead
a default connection created by Anaconda was used.  Let's clone the "Wired
connection" into the default connection created by Anaconda so that its
properties (like default ipv6.addr-gen-mode or
ipv4.dhcp-vendor-class-identifier) are preserved.

Related: rhbz#1870692


poncovka (approved):
Looks good to me.

VladimirSlavik (approved):
Looks good to me. Thank you!

PR #2857, opened 2020-09-15
Fix the combo box for an URL type of additional repositories (#1879127)

poncovka:
In the commit ff9a7e1, we introduced new constants for the URL source types with
values `BASEURL`, `MIRRORLIST` and `METALINK`. In the commit cc2c3d3, we started
to use these constants in the Installation Source spoke and removed the old ones
with values `url`, `mirrorlist` and `metalink`. We updated the combo box for the
base repository, but forgot to update the combo box for additional repositories.

The combo box items have to have ids that match the values of the constants,
otherwise the URL type of additional repositories will be always `BASEURL`.

Resolves: rhbz#1879127

### DIFF
--- a/pyanaconda/ui/gui/spokes/installation_source.glade
+++ b/pyanaconda/ui/gui/spokes/installation_source.glade
@@ -1296,9 +1296,9 @@
                                         <property name="can_focus">False</property>
                                         <property name="halign">start</property>
                                         <items>
-                                          <item id="url" translatable="yes">repository URL</item>
-                                          <item id="mirrorlist" translatable="yes">mirrorlist</item>
-                                          <item id="metalink" translatable="yes">metalink</item>
+                                          <item id="BASEURL" translatable="yes">repository URL</item>
+                                          <item id="MIRRORLIST" translatable="yes">mirrorlist</item>
+                                          <item id="METALINK" translatable="yes">metalink</item>
                                         </items>
                                         <signal name="changed" handler="on_repo_url_type_changed" swapped="no"/>
                                       </object>


VladimirSlavik (approved):
Not sure what to say to this kind of diff. LGTM?

AdamWill:
So, hum, I'm thinking the commits that changed stuff here were ff9a7e1342a6f19b503506bae2efb4ed7d73b41b and cc2c3d38448d500debc25d784e1536ed31a286f3 ? That latter one seems to have changed *one* place in `installation_source.glade` in the same way as this PR, but not the place this PR touches...

so on the face of it this seems like it's probably going in the right direction, but...it'd be nice to know that someone had checked that *everything* is now consistent and there isn't anything else still looking for the 'old' strings...

VladimirSlavik:
@AdamWill if you want to do more digging yourself: Without any more info, I'd hazard a guess that somewhere we used to read `id` of currently selected item in GUI, and fed that to `...method` in the kickstart data tree, directly or indirectly. Or compared with possible values thereof. That would imply that now that we have abolished during modularization the Cthulhu of `method` as source of truth, the new code no longer matched the ids to the right stuff. We now generally go UPPERCASE on constants while `method` was from the age of kickstart so camelCase or lowercase. This commit/PR message and diff hints at all this, I think.

So maybe try seeing where `__init__` or `initialize` puts the elements changed in glade (or parent element / dropdown), and trace usages of that for reading the id or some such. That would then be the code where the constants no longer matched.

Myself, I'm content with the hypothesis, knowing Vendy wrote a lot of the implied changed code, and seeing the manual testing label dropped. I also think it's not worth chasing if she can just describe the thing for us...

poncovka:
> So, hum, I'm thinking the commits that changed stuff here were [ff9a7e1](https://github.com/rhinstaller/anaconda/commit/ff9a7e1342a6f19b503506bae2efb4ed7d73b41b) and [cc2c3d3](https://github.com/rhinstaller/anaconda/commit/cc2c3d38448d500debc25d784e1536ed31a286f3) ? That latter one seems to have changed _one_ place in `installation_source.glade` in the same way as this PR, but not the place this PR touches...

That is correct. I have updated the commit message to make it easier to understand the proposed changed.

> so on the face of it this seems like it's probably going in the right direction, but...it'd be nice to know that someone had checked that _everything_ is now consistent and there isn't anything else still looking for the 'old' strings...

I have searched for the old strings and the code seems to be fine now.

poncovka:
> @AdamWill if you want to do more digging yourself: Without any more info, I'd hazard a guess that somewhere we used to read `id` of currently selected item in GUI, and fed that to `...method` in the kickstart data tree, directly or indirectly. Or compared with possible values thereof. That would imply that now that we have abolished during modularization the Cthulhu of `method` as source of truth, the new code no longer matched the ids to the right stuff. We now generally go UPPERCASE on constants while `method` was from the age of kickstart so camelCase or lowercase. This commit/PR message and diff hints at all this, I think.

This bug is not related to the `method` issue. The support for additional repositories was barely touched during the transition of payload sources. This bug was caused by replacing one set of constants with another.

jkonecny12 (approved):
LGTM!

PR #2861, opened 2020-09-18
network: add support for bridged bond to stage 2 kickstart (%pre)

rvykydal:
This is required to be able handle kickstart completely in stage 2.


poncovka (approved):
Looks good to me.

rvykydal:
jenkins, test this please

rvykydal:
I've added fixup for pylint which I'll squash before merging.

pyanaconda/modules/network/nm_client.py
```diff
@@ -492,41 +529,71 @@ def add_connection_from_ksdata(nm_client, network_data, device_name, activate=Fa
...
         device_to_activate = device_name if activate else None
-        nm_client.add_connection2(
-            con.to_dbus(NM.ConnectionSerializationFlags.ALL),
-            (NM.SettingsAddConnection2Flags.TO_DISK |
-             NM.SettingsAddConnection2Flags.BLOCK_AUTOCONNECT),
-            None,
-            False,
-            None,
-            _connection_added_cb,
+        add_and_activate_connection_sync(
```
jkonecny12, line 540: I don't really like the `and` word in the function name. Most of the time it means that the functions should be split to two.

jkonecny12 (changes requested):
Could you please squash these commits. The second does not really give benefit.

pyanaconda/modules/network/nm_client.py
rvykydal: Why do you think this is the case ?
Also see: https://developer.gnome.org/libnm/stable/NMClient.html#nm-client-add-and-activate-connection-async

rvykydal:
> Could you please squash these commits. The second does not really give benefit.

Please see the previous comment, it is for easier review for additional change for people that already reviewed the PR.

poncovka (approved):
Looks good to me.

pyanaconda/modules/network/nm_client.py
jkonecny12: Because smaller function is easier to call and simpler for modification. Also having:
```
add_and_activate_connection_sync()
```
is not giving you much benefit compared to
```
add_connection_sync()
activate_connection_sync()
```

You will strip dependencies to just the parts requiring it. And the last reason is that it's simpler for testing because it's smaller.

However, this is not something for blocking this PR. It's just note. Reason for block was the commit squashing.

jkonecny12:
> > Could you please squash these commits. The second does not really give benefit.
> 
> Please see the previous comment, it is for easier review for additional change for people that already reviewed the PR.

In general yes, but fixing pylint issues does not give any benefit to anyone. These are just changing how the code is written not what the code is doing so not really something requiring review. However, it will make harder to list history of the code and lines in the code basically it will shadow reason for the code change in the blame.

pyanaconda/modules/network/nm_client.py
VladimirSlavik: If you wanted to get pedantic, it's used exactly once so maybe mark it as _private. Also, "activate" could be dropped from name since it's in the parameter list. But I don't think it's important...

VladimirSlavik (approved):
Looks good to me. Thank you!

rvykydal:
> > > Could you please squash these commits. The second does not really give benefit.
> > 
> > 
> > Please see the previous comment, it is for easier review for additional change for people that already reviewed the PR.
> 
> In general yes, but fixing pylint issues does not give any benefit to anyone. These are just changing how the code is written not what the code is doing so not really something requiring review. However, it will make harder to list history of the code and lines in the code basically it will shadow reason for the code change in the blame.

In my comment above I am just saying:
"I've added fixup for pylint which I'll squash before merging."

pyanaconda/modules/network/nm_client.py
rvykydal: I split the function:
https://github.com/rhinstaller/anaconda/pull/2861/commits/ce593eed2b4706fd82876276b315c6010d5b8079

jkonecny12:
> > > > Could you please squash these commits. The second does not really give benefit.
> > > 
> > > 
> > > Please see the previous comment, it is for easier review for additional change for people that already reviewed the PR.
> > 
> > 
> > In general yes, but fixing pylint issues does not give any benefit to anyone. These are just changing how the code is written not what the code is doing so not really something requiring review. However, it will make harder to list history of the code and lines in the code basically it will shadow reason for the code change in the blame.
> 
> In my comment above I am just saying:
> "I've added fixup for pylint which I'll squash before merging."

Ohh sorry my bad.

jkonecny12 (approved):
Looks good to me.

PR #2865, opened 2020-09-22
Fix traceback when starting installation with isnt.console (no args)

jkonecny12:
`None` is default value if no value is specified which would fail on os.path.basename call.

### DIFF
--- a/pyanaconda/modules/storage/bootloader/base.py
+++ b/pyanaconda/modules/storage/bootloader/base.py
@@ -903,9 +903,15 @@ class BootLoader(object):
 
     def _set_console(self):
         """ Set console options based on boot arguments. """
-        console = kernel_arguments.get("console", "")
+        console = kernel_arguments.get("console")
+
+        if not console:
+            return
+
         console = os.path.basename(console)
         self.console, _x, self.console_options = console.partition(",")
+
+        log.debug("Console is set to %s with options '%s'", self.console, self.console_options)
 
     def write_config_console(self, config):
         """Write console-related configuration lines."""


VladimirSlavik (approved):
Looks good to me. Thank you!

(Is returning `None` the right thing?)

jkonecny12:
jenkins, test this please

pyanaconda/modules/storage/bootloader/base.py
```diff
@@ -902,6 +902,9 @@ def boot_prefix(self):
     def _set_console(self):
         """ Set console options based on boot arguments. """
         console = kernel_arguments.get("console", "")
+        if console is None:
```
poncovka, line 905: We should do that for an empty string as well. Therefore, the code can be simplified into:

```
console = kernel_arguments.get("console")
if not console:
    return
```

poncovka (changes requested):

pyanaconda/modules/storage/bootloader/base.py
jkonecny12: OK and what about the log message. Seems that you want to drop it.

poncovka: I would suggest to log only the successful path with a message "The console is set to something with options something". It will be more informative anyway.

jkonecny12:
**UPDATED:**
* Implemented changes suggested by @poncovka

poncovka (approved):
Thanks!

jkonecny12:
**UPDATED:**
* Fixed typo in the commit message `isnt` -> `inst`.
* Changed the commit message to make it more clear.

poncovka (approved):

PR #2866, opened 2020-09-23
Never mount partitions on a disk with the iso9660 filesystem

poncovka:
Blivet doesn't recognize these partitions, so we mount the disk in stage2.
However, it will fail if one of its partitions is already mounted, for example
in stage1. Therefore, skip these partitions in the script that looks for a root
image on devices and use the disk instead.

The boot.iso has the following structure:

```
NAME   TYPE  SIZE FSTYPE  LABEL
sda    disk  8.8G iso9660 RHEL-8-3-0-BaseOS-x86_64
|-sda1 part  8.8G iso9660 RHEL-8-3-0-BaseOS-x86_64
`-sda2 part   10M vfat    ANACONDA

```
And the following default boot options:

`inst.stage2=hd:LABEL=RHEL-8-3-0-BaseOS-x86_64 rd.live.check quiet`

Anaconda runs the anaconda-diskroot script for every device that matches the
specification from the inst.stage2 option. In this example, it is sda and sda1,
because both of them have the specified label RHEL-8-3-0-BaseOS-x86_64. With
the fix, the script will detect that sda1 is a partition on a disk with the
iso9660 filesystem and stop to process sda1. The script will succeed with sda.

We should never rely on the order of the devices.

(cherry-picked from a commit 301a09d)

Related: rhbz#1878784

### DIFF
--- a/dracut/anaconda-diskroot
+++ b/dracut/anaconda-diskroot
@@ -41,7 +41,22 @@ dev="$1"
 path="$2" # optional, could be empty
 kickstart="$(getarg ks= inst.ks=)"
 
-[ -e "/dev/root" ] && exit 1 # we already have a root device!
+# Log the device that triggered this job.
+debug_msg "Trying to find a root image on the device $dev."
+
+# Do we already have a root device?
+# Then do not run again.
+[ -e "/dev/root" ] && exit 1
+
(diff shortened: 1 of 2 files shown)


VladimirSlavik (approved):
Thanks for the port!

jkonecny12 (approved):
LGTM!

PR #2870, opened 2020-09-25
Process network kickstart in stage 2

rvykydal:
~~The first comit is already on f33-devel.~~
~~For the second there is a PR for f33-devel.~~
~~I will remove them after merging them into master with f33-devel.~~

~~All the rest~~ The PR is only for master (so we catch fallouts before getting into Fedora 34) where we move from:

A: Network kickstart parsed in initramfs into ifcfg files based on which NM started after switch root creates and activates the respective connections.

to:

B: Network kickstart parsed in Anaconda after switch root which creates and activates NM connections via NM API.

The B: is what we already used to do for networking configured in %pre section so it has already got some usage and testing. We just needed to add a few more complex cases (eg vlan over bond).

The change from A to B was made possible by NM replacing dracut network module in initramfs.

The PR makes the code fully independent of the configuration format NM is using (ifcfg file of keyfile). Also network configuration (and perhaps Anaconda as whole) is no more dependent on fedora-import-service (passing configuration files from initramfs to rootfs), which used to cause issues like rhbz#1537977. Also it allows to get rid of hacks around ONBOOT value used for activating/not activating ifcfgs created in initramfs by NM.

The only thing that is changing with this PR is that NM default autoconnections are no more blocked from being created by ifcfgs with ONBOOT=no created in initramfs.


pyanaconda/modules/network/nm_client.py
```diff
@@ -672,7 +672,7 @@ def bound_hwaddr_of_device(nm_client, device_name, ifname_option_values):
     :return: hwaddress of the device if bound, or None
     :rtype: str or None
     """
-    for ifname_value in ifname_option_values:
+    for ifname_value in ifname_option_values or []:
```
poncovka, line 675: I would replace `ifname_option_values` with an empty list in `update_connection_from_ksdata` to make sure it is always a list. The current workaround in `bound_hwaddr_of_device` is a little hard to follow.

pyanaconda/modules/network/initialization.py
```diff
@@ -148,103 +148,6 @@ def _find_initramfs_connection_of_iface(self, iface):
         return None
 
 
-class ConsolidateInitramfsConnectionsTask(Task):
```
poncovka, line 151: I recommend to run [vulture](https://pypi.org/project/vulture/). There might be some functions and constants that are no longer used thanks to the cleanup.

poncovka (approved):
It looks great! Thank you!

jkonecny12 (approved):
Looks good to me as far as I can tell :).

VladimirSlavik (approved):
Looks good to me, as far as I can tell. Thank you! Massive removal is massive.

pyanaconda/modules/network/initialization.py
rvykydal: Good tip. I've found only https://github.com/rhinstaller/anaconda/pull/2894 which is unrelated to this patchset.

pyanaconda/modules/network/nm_client.py
rvykydal: Done.

rvykydal:
I've replaced the first two commits and rebased the branch to the current master.

poncovka (approved):
Looks good to me. Thanks!

PR #2871, opened 2020-09-25
Use the DBus source for RPM OSTree in UI

poncovka:
Use the RPM OSTree DBus module and its source to install the payload.
The `ostreesetup` kickstart command should be handled only by the DBus module now.

### DIFF
--- a/pyanaconda/modules/payloads/source/rpm_ostree/rpm_ostree.py
+++ b/pyanaconda/modules/payloads/source/rpm_ostree/rpm_ostree.py
@@ -121,4 +121,9 @@ class RPMOSTreeSourceModule(PayloadSourceBase):
         return []
 
     def __repr__(self):
-        return "Source(type='RPM_OSTREE')"
+        """Return a string representation of the source."""
+        return "Source(type='{}', osname='{}', url='{}')".format(
+            self.type.value,
+            self.configuration.osname,
+            self.configuration.url
+        )
--- a/pyanaconda/modules/payloads/source/rpm_ostree/rpm_ostree.py
+++ b/pyanaconda/modules/payloads/source/rpm_ostree/rpm_ostree.py
@@ -81,7 +81,6 @@ class RPMOSTreeSourceModule(PayloadSourceBase):
 
     def get_state(self):
         """Get state of this source."""
-        # TODO: Implement this method
         return SourceState.NOT_APPLICABLE
 
     def process_kickstart(self, data):
@@ -108,7 +107,6 @@ class RPMOSTreeSourceModule(PayloadSourceBase):
         :return: list of tasks required for the source setup
         :rtype: [Task]
         """
-        # TODO: Implement this method
         return []
 
     def tear_down_with_tasks(self):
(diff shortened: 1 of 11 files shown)


pyanaconda/modules/payloads/source/rpm_ostree/rpm_ostree.py
```diff
@@ -76,8 +77,7 @@ def network_required(self):
 
         :return: True or False
         """
-        # TODO: Implement this method
-        return False
+        return has_network_protocol(self.configuration.url)
```
jkonecny12, line 80: I'm not sure if this will work. I have a feeling that mirrors for rpm-ostree could be specified also as ```oci+https://...```, however, I wasn't able to find this anywhere.

Hi @dustymabe , could you please help us here and tell us. If rpm-ostree remote sources could be specified only by "common" prefixes (http, https, ftp...) or it supports also something special like the ``oci+http:///`` ? Thanks!

jkonecny12 (changes requested):
I'm concern about my comment below. Other than that looks good to me. Great work!

But first let's get the answer.

VladimirSlavik (approved):
Looks good to me. Thank you!

poncovka:
Rebased.

pyanaconda/modules/payloads/source/rpm_ostree/rpm_ostree.py
poncovka: I have changed the implementation, so it doesn't block this PR. See the commit https://github.com/rhinstaller/anaconda/pull/2871/commits/5a770dac2598b2c8d0a674d808f650442340820e.

pyanaconda/modules/payloads/source/rpm_ostree/rpm_ostree.py
```diff
@@ -76,12 +77,10 @@ def network_required(self):
 
         :return: True or False
         """
-        # TODO: Implement this method
-        return False
+        return not (self.configuration.url and self.configuration.url.startswith("file://"))
```
VladimirSlavik, line 80: Why not extend `has_network_protocol` with an option to include the ostree url styles?

(Alternatively, the import is unnecessary now and will fail the tests, I think.)

pyanaconda/modules/payloads/source/rpm_ostree/rpm_ostree.py
owtaylor: 'oci+https://' is Flatpak-specific. (means "link to an index of a container registry") ostree only uses http/https.

jkonecny12: Thanks great for the reply. In that case I'm fine with this solution!

jkonecny12 (approved):
Based on the above I agree with the old and current solution. :)

poncovka:
The workaround was dropped.

VladimirSlavik (approved):

jkonecny12 (approved):
Looks good to me.

PR #2872, opened 2020-09-29
fix remove unkown partition in sda failed

xqrustc2020:
fix: cannot create partition when sda exists a ext4 filesystem.
When you remove unkown partition in sda, only the format in sda should be destroyed without removing sda from the device tree, but sda is also destroyed. As a result, sda cannot be found during disk initialization and an error is reported. 
Resolves: rhbz#1878661

### DIFF
--- a/pyanaconda/modules/storage/partitioning/interactive/utils.py
+++ b/pyanaconda/modules/storage/partitioning/interactive/utils.py
@@ -1072,8 +1072,10 @@ def _destroy_device(storage, device):
     :param device: an instance of a device
     """
     # Remove the device.
-    if device.is_disk and device.partitioned and not device.format.supported:
-        storage.recursive_remove(device)
+    if device.is_disk:
+        if device.partitioned and not device.format.supported:
+            storage.recursive_remove(device)
+        storage.initialize_disk(device)
     elif device.direct and not device.isleaf:
         # We shouldn't call this method for with non-leaf devices
         # except for those which are also directly accessible like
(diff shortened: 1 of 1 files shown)


poncovka:
Jenkins, test this please.

TomasTomecek:
dang, seems like testing farm is having problems :/

jkonecny12:
Yep, I asked them and the cluster is dying. They are in process of migration to a new cluster (which should also switch away from crusher). However, they expect the work took a few weeks...

VladimirSlavik (approved):
@xqrustc2020, thank you for the contribution! The change looks good to me - originally the code was exactly that way after c6e41f88d0d2f3e13d2fa922f97e512413d30e21 and got broken in 23c7e6803a7349304c7a0f4d187030d005418429 with some simplifications during modularization.

VladimirSlavik:
Given the changes are 1:1 with what we used to have, it isn't probable that this could break something.

But if I forget about that, what I see is a bunch of complicated `elif`s chained together. Changes in the frontmost condition (especially making it more catch-all) could break the chain. So I'm wondering what is the best way to verify not just that this works, but also that it does not bring some obscure regression.

jkonecny12 (approved):
Seems good to me from the code PoV but it has to be approved by the assignee.

poncovka (approved):
Looks good to me. Thanks!

VladimirSlavik:
This works well when just trying out by hand, and in "storage" kickstart tests too. I think we're good to go with this one.

PR #2873, opened 2020-09-29
Document the mount points of the target system

poncovka:
We should document the difference between /mnt/sysroot and /mnt/sysimage.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -25,4 +25,5 @@ Contents:
    multipath
    list-harddrives
    sysconfig-file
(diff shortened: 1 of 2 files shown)


VladimirSlavik (approved):
Looks good to me. Thank you!

VladimirSlavik:
It might be useful to add also something about the source-specific mount points. But I don't think that's a requirement, IMO the PR is pretty useful already as such.

jkonecny12 (approved):
LGTM!

PR #2874, opened 2020-09-29
Replace metacity with mutter

bcl:
mutter is the modern replacement for metacity, used by gnome-shell and
supporting x11 or wayland.

### DIFF
--- a/pyanaconda/display.py
+++ b/pyanaconda/display.py
@@ -177,9 +177,9 @@ def do_startup_x11_actions():
     else:
         xdg_data_dirs = datadir + '/window-manager:/usr/share'
 
-    childproc = util.startProgram(["metacity", "--display", ":1", "--sm-disable"],
+    childproc = util.startProgram(["mutter", "--display", ":1", "--sm-disable", "--x11"],
                                   env_add={'XDG_DATA_DIRS': xdg_data_dirs})
-    WatchProcesses.watch_process(childproc, "metacity")
+    WatchProcesses.watch_process(childproc, "mutter")
 
 
 def set_x_resolution(runres):


bcl:
I did some experiments with lorax and anaconda, trying to use gnome-shell and couldn't get it to start. It complains about javascipt and void not being a valid type. It also drags in a bunch of extra things that aren't really needed.

So I tried mutter, which is what gnome-shell runs on top of, and surprisingly enough it works. Mostly.

As far as I can tell the only thing that doesn't work is the keyboard switcher. Adding keyboards works, and when you move one of them to the top of the list the layout indicator widget does change. But clicking on it doesn't do anything.

I suspect something is wrong with libxklavier. I wrote a little script to cycle through the layouts every 5 seconds, and to print the group id that it thought was selected. On metacity it does cycle through them, but the output doesn't make sense. It never changes group.state to anything other than 0, and get_next_group always returns 1. Actually, mutter's behavior is more like what I would expect from this output than metacity (ie. always selecting the default).

Here's the script:

```python
#!/usr/bin/python3
import sys
import time

import gi
gi.require_version("GdkX11", "3.0")
gi.require_version("Xkl", "1.0")
from gi.repository import GdkX11, Xkl

display = GdkX11.x11_get_default_xdisplay()
engine = Xkl.Engine.get_instance(display)
state = engine.get_current_state()
print(dir(state))
print("before")
print(state.group)
num_groups = engine.get_num_groups()
print(num_groups)

group = int(sys.argv[1]) 
while True:
    engine.lock_group(group)
    state = engine.get_current_state()
    print("after")
    print(f"current: {state.group}")

    next_group = engine.get_next_group()
    print(f"next: {next_group}")

    time.sleep(5)
    group = (group + 1) % num_groups
    state = engine.get_current_state()
    print("afterer")
    print(f"current: {state.group}")
```

So, my conclusion is that this is pretty close, with minimal changes, and that we should solve the switcher problem instead of trying to drag in gnome-shell and all of its various dependencies.

See https://github.com/weldr/lorax/pull/1086 for the Lorax template changes.

jkonecny12 (commented):
This is pretty interesting. I wonder what the desktop team will tell us about using this solution.

Thanks a lot Brian for looking on this.

bcl:
One other thing to mention, @M4rtinK mentioned that first boot also has a keyboard switcher which appears to work with different window managers. I haven't looked at it yet, but it's worth exploring what they do differently, if anything.

VladimirSlavik:
We settled on gnome-kiosk as the replacement - see #3307 - so this is not needed anymore.

PR #2875, opened 2020-09-30
Drop obsolete Py_Initialize link check in configure

martinpitt:
Current Python 3 does not link extensions to libpython3.so any more [1],
`and consequently `pkg-config --libs python3` is empty. This causes the
"Checking for Python libraries" check to always fail, resulting in a
scary message

    *** Anaconda encountered the following issues during configuration:
    Unable to use python library

    *** Anaconda will not successfully build without these missing dependencies

The build will still work anyway though, and pyanaconda does not link to
libpython3 either, so the check is superfluous, obsolete, and always
fails anyway.

[1] https://bugs.python.org/issue21536

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -86,17 +86,7 @@ AC_CHECK_LIB([audit], [audit_open], [:],
 AM_PATH_PYTHON(python_required_version)
 
 # Check for the python extension paths
-PKG_CHECK_MODULES([PYTHON3], [python3], [
-    LIBS_save="$LIBS"
-    LIBS="$LIBS $PYTHON3_LIBS"
-    AC_MSG_CHECKING([Python libraries])
-    AC_TRY_LINK_FUNC([Py_Initialize], 
-                     [AC_MSG_RESULT([yes])], 
-                     [AC_MSG_RESULT([no])
-                      ANACONDA_SOFT_FAILURE([Unable to use python library])])
-    LIBS="$LIBS_save"
-    ],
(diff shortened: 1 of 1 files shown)


martinpitt:
The same message [happens in Koji as well](https://kojipkgs.fedoraproject.org//packages/anaconda/34.7/1.fc34/data/logs/x86_64/build.log). This was a bit scary for me as first-timer,  took a while to realize that it's harmless.

martinpitt:
Can someone please guide me about the [packit failure](https://prod.packit.dev/srpm-build/8027/logs)? The log looks like the build environment doesn't have python3-devel installed. We still need the `PKG_CHECK_MODULES([PYTHON3], [python3])` check, as pyanaconda requires `PYTHON3_CFLAGS` at least. But that packit log looks entirely different from e. g. PR #2874 ([log](https://console-testing-farm.apps.ci.centos.org/pipeline/cf39eaeb-8096-40a4-b2e4-c75b2a4fb9f6)).

martinpitt:
It seems we want to run `make release` in packit, need `./configure` to do that, but its environment does not install build dependencies. In master, ./configure will succeed and just print the warning. This isn't really nice, but also doesn't hurt that much, so let's just go back to that. Fix pushed.

poncovka (approved):
Looks good to me. Thanks!

jkonecny12 (approved):
Thanks a lot for fixing this. It was on my TODO list deeply hidden for a long time. LGTM!

PR #2877, opened 2020-09-30
Set up proxy environmental variables with a function

poncovka:
Move the code that sets up the proxy environmental variables into a new function
`set_up_proxy_variables` in startup utils. Call this function before we start the
DBus modules, so the environment of the modules has these variables set.

### DIFF
--- a/anaconda.py
+++ b/anaconda.py
@@ -332,7 +332,6 @@ if __name__ == "__main__":
     from pyanaconda import startup_utils
     from pyanaconda import rescue
     from pyanaconda import geoloc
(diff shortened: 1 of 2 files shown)


poncovka:
Jenkins, test this please.

jkonecny12 (changes requested):
Could you please explain me the reasoning behind this. I can't image a reason to have configuration option for proxy settings because proxy is dependent on where the computer is placed and not about by what you are doing the installation.

poncovka:
In progress.

poncovka:
Updated.

jkonecny12 (approved):
Looks good to me now. Thanks!

PR #2878, opened 2020-10-01
Fix instructions for unit tests

martinpitt:
With these I am able to set up a mock and successfully run unit tests. On Fedora 33 one first needs to manually install the 
[latest pykickstart](https://koji.fedoraproject.org/koji/buildinfo?buildID=1617300), as the one in current 33 isn't sufficient.

### DIFF
--- a/tests/README.rst
+++ b/tests/README.rst
@@ -49,8 +49,8 @@ produce a human readable report.
 Run tests inside Mock
 ---------------------
 
-When using the `ci' target in a mock you need to use a regular user account which
-is a member of the `mock' group. You can update your account by running
+When using the `ci` target in a mock you need to use a regular user account which
+is a member of the `mock` group. You can update your account by running
 the command::
 
     # usermod -a -G mock <username>
--- a/tests/README.rst
+++ b/tests/README.rst
@@ -35,7 +35,7 @@ Executing the tests can be done with::
 
 To run a single test do::
 
-    make TESTS=install/nosetests.sh check
+    make TESTS=nosetests.sh check
 
 See `tests/Makefile.am` for possible values. Alternatively you can try::
 
--- a/tests/README.rst
+++ b/tests/README.rst
@@ -57,14 +57,17 @@ the command::
 
 To prepare testing mock environment call::
 
-    ./scripts/testing/setup-mock-test-env.py [mock-configuration]
+    ./scripts/testing/setup-mock-test-env.py --init [mock-configuration]
 
 Mock configuration can be path to a file or name of file in `/etc/mock/*.cfg`
 without suffix. For detail configuration look on the script help output.
 
 Then you can run tests by::
 
-    mock -r [mock_configuration] --chroot -- "cd /anaconda && ./autogen.sh && ./configure && make ci"
+    ./scripts/testing/setup-mock-test-env.py -ut [mock-configuration]
+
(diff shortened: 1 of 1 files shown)


VladimirSlavik:
So, I'm not the most qualified person to speak about this, but the "in mock" part seems to be more concerned with understanding how the sausage is made, rather than succeeding in running the tests. Maybe that's a side effect of the tooling evolving and the readme not.

What I'll do on a daily basis is more or less like this:

1. Do the full build with `--init` and avoid running pylint which takes forever:  `rm -rf ./test_results ; scripts/testing/setup-mock-test-env.py --init --copy --run-nosetests --result test_results anaconda-rawhide-x86_64`

2. If I then just edit the python code which needs no building, I'll replace `--init --copy` with `--update` for subsequent runs which shaves off some 30% by not building the package. I will also want to save yet more time by isolating the specific test(s) I want to see: `--run-nosetests tests/nosetests/pyanaconda_tests/xxx.py

3. Once I'm approaching some kind of conclusion for the work "unit", I'll re-run the full suite again with `--run-tests` to see what pylint thinks.

Maybe that could be reflected somehow in the edit too?

martinpitt:
Talking to @jkonecny12 , I'll update that to the usual workflow instead. Also, the "leaking mounts" is annoying mostly because I still had an ancient ~/.config/mock.cfg which disabled nspawn. So that part is fine now.

martinpitt:
@jkonecny12 : Updated now, and confirmed that this works well. Thanks for your insights on IRC!

tests/README.rst
```diff
@@ -57,14 +57,17 @@ the command::
...
-    ./scripts/testing/setup-mock-test-env.py [mock-configuration]
+    ./scripts/testing/setup-mock-test-env.py --init [mock-configuration]
 
 Mock configuration can be path to a file or name of file in `/etc/mock/*.cfg`
 without suffix. For detail configuration look on the script help output.
 
 Then you can run tests by::
 
-    mock -r [mock_configuration] --chroot -- "cd /anaconda && ./autogen.sh && ./configure && make ci"
+    ./scripts/testing/setup-mock-test-env.py -ut [mock-configuration]
```
martinpitt, line 67: Note: There's `--copy` and `--update`, but it seems `--update` works just fine for the first time as well. So for someone who isn't interested in the details, this seems to be a reasonable standard command?

martinpitt:
@VladimirSlavik : Indeed, this was mostly my ignorance, and the README leading off to the wrong track. Thanks!

martinpitt:
BTW, I'm using `sudo mount -t tmpfs tmpfs /var/lib/mock`, which speeds up `--init` really nicely.

jkonecny12:
> BTW, I'm using `sudo mount -t tmpfs tmpfs /var/lib/mock`, which speeds up `--init` really nicely.

I would suggest rather using https://github.com/rpm-software-management/mock/wiki/Plugin-Tmpfs

tests/README.rst
martinpitt: "need" in setup-mock-test-env.py, of course. Do you want me to document both in README, though? Or leave that to the "see --help" part?

jkonecny12 (approved):
LGTM! Thanks for improvement!

VladimirSlavik (approved):
Looks good to me. Thank you!

PR #2879, opened 2020-10-01
Fix show missing inst. prefix warning appropriately

jkonecny12:
The original solution removed the inst. prefix before it went to the cmdline options parsing. That way the code couldn't react appropriately on missing inst. prefix.

*Resolves: rhbz#1875561*

### DIFF
--- a/pyanaconda/core/kernel.py
+++ b/pyanaconda/core/kernel.py
@@ -24,6 +24,11 @@ from pyanaconda.core.constants import CMDLINE_APPEND, CMDLINE_LIST, CMDLINE_FILE
 __all__ = ['KernelArguments', 'kernel_arguments']
 
 
+# options might have the inst. prefix (used to differentiate
+# boot options for the installer from other boot options)
+BOOT_ARG_PREFIX = "inst."
+
+
 class KernelArguments():
     """The kernel arguments.
 
@@ -32,6 +37,7 @@ class KernelArguments():
 
     def __init__(self):
         self._data = OrderedDict()
+        self._args_with_prefix = set()
 
     @classmethod
     def from_defaults(cls):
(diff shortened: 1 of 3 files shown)


tests/nosetests/pyanaconda_tests/kernel_test.py
```diff
@@ -27,7 +27,8 @@ def value_retrieval_test(self):
         """KernelArguments value retrieval test."""
 
         ka = KernelArguments.from_string(
-            "blah foo=anything bar=1 baz=0 nowhere nothing=indeed beep=off derp=no nobody=0")
+            "inst.blah foo=anything inst.bar=1 baz=0 nowhere inst.nothing=indeed beep=off "
+            "derp=no nobody=0")
```
VladimirSlavik, line 31: Any reason to change this if you're changing neither the code nor the test?

pyanaconda/core/kernel.py
```diff
@@ -162,5 +169,18 @@ def items(self):
         """
         return self._data.items()
 
+    def items_raw(self):
+        """Return an iterator over all arguments in their raw form (with prefixes).
+
+        TODO: DO NOT USE THIS METHOD! This workaround will be removed
+              when 'inst.' prefix is not supported.
```
VladimirSlavik, line 176: ...lack of... ?

VladimirSlavik (changes requested):
I think you might want to clean this up a bit? Otherwise ready to ack the main idea.

poncovka (approved):

pyanaconda/core/kernel.py
```diff
@@ -32,6 +37,7 @@ class KernelArguments():
 
     def __init__(self):
         self._data = OrderedDict()
+        self._prefixes = set()
```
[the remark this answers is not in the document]
poncovka: Or `_has_prefix`?

tests/nosetests/pyanaconda_tests/kernel_test.py
jkonecny12: Because it's more appropriate. Before the change we won't test that the prefix is correctly removed.

pyanaconda/core/kernel.py
```diff
@@ -162,5 +169,18 @@ def items(self):
...
+    def items_raw(self):
+        """Return an iterator over all arguments in their raw form (with prefixes).
+
+        TODO: DO NOT USE THIS METHOD! This workaround will be removed
+              when 'inst.' prefix is not supported.
+        """
+        for key, val in self._data.items():
+            if key in self._prefixes:
+                yield ("{}{}".format(BOOT_ARG_PREFIX, key), val)
+                continue
```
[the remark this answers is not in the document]
jkonecny12: Without the `continue` statement it will yield `inst.stage2` and then `stage2` because it won't take next item from the loop but return on the same place and take the value from 2 lines below.

pyanaconda/core/kernel.py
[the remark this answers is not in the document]
jkonecny12: I changed this to ``args_with_prefix``

pyanaconda/core/kernel.py
jkonecny12: Fixed.

VladimirSlavik (approved):

PR #2882, opened 2020-10-02
Unit test fixes

martinpitt:
This paves the way for running the unit tests in a container. I will submit a draft PR which does that, but as that may need a little more discussion and fine-tuning, I'm submitting these parts separately.

This can be reproduced with running 

    make tests-nose-only NOSE_TESTS_ARGS='nosetests/pyanaconda_tests/module_storage_test.py'

in mock with a non-root user, in other words

    mock -r fedora-33-x86_64 --shell --unpriv

### DIFF
--- a/tests/nosetests/pyanaconda_tests/module_storage_test.py
+++ b/tests/nosetests/pyanaconda_tests/module_storage_test.py
@@ -1457,13 +1457,26 @@ class StorageTasksTestCase(unittest.TestCase):
         storage.assert_not_called()
 
     @patch("pyanaconda.core.util.mkdirChain")
-    @patch("pyanaconda.core.util.execWithRedirect")
-    def mount_filesystems_test(self, execute, mkdir):
+    @patch("pyanaconda.core.util._run_program")
+    @patch("os.makedirs")
+    @patch("blivet.util._run_program")
+    def mount_filesystems_test(self, blivet_run_program, makedirs, core_run_program, mkdirChain):
         """Test MountFilesystemsTask."""
         storage = create_storage()
         storage._bootloader = Mock()
-        execute.return_value = 0
+        blivet_run_program.return_value = (0, "")
+        core_run_program.return_value = (0, "")
         MountFilesystemsTask(storage).run()
+        # created the mount points
+        makedirs.assert_any_call('/mnt/sysimage/dev', 0o755)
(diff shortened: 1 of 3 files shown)


martinpitt:
/me gives PEP-8 a pacifier and pours some extra love into his whitespace :heart_eyes:

martinpitt:
Note that this is still quite far from being able to run *all* nosetests as unprivileged user -- there are still 7 failures and 19 skips. There isn't anything insurmountable there, just some more mocking out to do. But it's not really a blocker for moving the tests off of Jenkins, so I don't do that for now.

poncovka (approved):
Looks good to me. Thank you!

tests/nosetests/pyanaconda_tests/module_storage_test.py
```diff
@@ -1457,13 +1457,26 @@ def activate_filesystems_test(self, patched_conf):
...
+    def mount_filesystems_test(self, blivet_run_program, makedirs, core_run_program, mkdirChain):
         """Test MountFilesystemsTask."""
         storage = create_storage()
         storage._bootloader = Mock()
-        execute.return_value = 0
+        blivet_run_program.return_value = (0, "")
+        core_run_program.return_value = (0, "")
         MountFilesystemsTask(storage).run()
+        # created the mount points
+        makedirs.assert_any_call('/mnt/sysimage/dev', 0o755)
```
VladimirSlavik, line 1471: Not important, but if I wanted to be pedantic, I'd ask if you could go with `assert_called_once_with` instead.

VladimirSlavik (approved):
Looks good to me. Thank you!

tests/nosetests/pyanaconda_tests/module_storage_test.py
martinpitt: But that's not true -- It gets called with a lot of other mount points, like /mnt/sysimage/proc and /sys, and /selinux etc. I don't think it makes much sense to replicate that entire list here, and e.g. the /selinux submount is conditional (it doesn't happen in a container with SELinux disabled). So I picked /dev as a representative example, and leave the rest to blivet -- we are not testing the blivet internals here.

jkonecny12 (approved):
Looks great to me. Thanks a lot!

PR #2883, opened 2020-10-02
Run unit tests in a GitHub action

martinpitt:
Set up a Fedora rawhide container with
scripts/testing/dependency_solver.py and run the unit tests. Export the 
test log on failure.

This is a bit inefficient (dependency_solver.py setup takes about 2
mintues) and brittle (as rawhide may break at any time), but that is
exactly what happens in Jenkins through setup-mock-test-env.py right
now. It should be improved later by regularly building and publishing a
custom container image.

------

 - [x] Builds on top of PR #2882, land that first
 - [x] Export coverage reports as artifacts

Follow-ups:
 - [ ] Dynamically pick container version based on [target branch](https://docs.github.com/en/free-pro-team@latest/actions/reference/context-and-expression-syntax-for-github-actions)
 - [ ] Provide a script to locally prepare a podman container and run tests in it

This is supposed to replace the "Anaconda Rawhide tests" on Jenkins. It's a first working PoC that runs the same build/test than that, aside from exporting the coverage logs (which is easy enough to do).

I would also like to replace the hardcoded `:master` container tag name with something that picks `master` vs `33` etc. based on the PR's target branch, so that we can use the exact same workflow file on branches.

But first I'd like to get some review/consensus if that is suitable for you. I recommend it because it's super-simple to set up, and there is not that much magic sauce involved -- e.g. it's fairly straightforward to write a .travis.yml or a Semaphore 2.0 CI. For running on a public OpenShift/k8s instance it's not good enough yet, as they usually don't give you containers with root access -- but that's just [some more well-understood work](https://github.com/rhinstaller/anaconda/pull/2882#issuecomment-702552830) if that ever comes up.

If you generally like the approach, I'll do the remaining work in the above TODO list.

### DIFF
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -1,0 +1,35 @@
+name: validate
+on: [push, pull_request]
+jobs:
+  unit-tests:
+    runs-on: ubuntu-latest
+    container:
+      image: docker.io/fedora:rawhide
+    steps:
+      - name: Clone repository
+        uses: actions/checkout@v2
+
+      # FIXME: the following should be done statically in a custom container image
+      - name: set up dependencies
+        run: |
+          dnf install -y /usr/bin/xargs
+          scripts/testing/dependency_solver.py -brt | xargs -d '\n' dnf install -y
+          scripts/testing/dependency_solver.py --pip | xargs pip install
+
+      - name: build
+        run: |
+          ./autogen.sh
+          ./configure
+          make
+
+      - name: run unit tests
+        run: |
+          make ci
+
+      - name: Upload test and coverage logs
+        uses: actions/upload-artifact@v2
+        with:
+          name: logs
+          path: |
+            tests/test-suite.log
+            tests/coverage-*.log


martinpitt:
Hmm, I wonder why the action doesn't trigger here -- it worked quite fine on [a PR against my forked repo](https://github.com/martinpitt/anaconda/pull/1/checks) , see the [check output](https://github.com/martinpitt/anaconda/runs/1197306137?check_suite_focus=true). https://github.com/rhinstaller/anaconda/settings/actions is enabled, maybe they don't run on draft PRs or so?

martinpitt:
Actions [need to be enabled in the rhinstaller org](https://docs.github.com/en/free-pro-team@latest/github/setting-up-and-managing-organizations-and-teams/disabling-or-limiting-github-actions-for-your-organization), it should be [here](https://github.com/organizations/rhinstaller/settings/actions). I don't have admin for the organization, actions may be restricted there? I suppose it does make sense to restrict actions to those we have in the repository, but as we currently have none, could this be temporarily allowed to test this PR? @jkonecny12 and @M4rtinK , you should both have admin privs there ("Owner" in the [team](https://github.com/orgs/rhinstaller/people))

VladimirSlavik:
It seems we have the most permissive setting there, so I'd go with the "not for draft" hypothesis.

martinpitt:
@VladimirSlavik : https://github.com/martinpitt/anaconda/pull/2 is draft as well, so most probably not -- but I'll try.

VladimirSlavik:
Oops, spoke too soon - it was disabled for PRs. I did not enable any of the tokens under there, ask if you need them too.

On a different note, I think you should have access to the page too? I'm not aware of any special treatment for me, although maybe @jkonecny12 added me somewhere in the past as I'm the infrastructure backup...

martinpitt:
@VladimirSlavik : ah, thanks, that might be it. @dvdhrm just pointed out that I had the wrong `on:` setting, so I added `pull_request`. But indeed they need to be enabled in the org.

> I think you should have access to the page too

If "should" means "by policy", then I leave that to you. If it means "it ought to work already", then no -- I'm only a "Member" on https://github.com/orgs/rhinstaller/people

martinpitt:
Actions are enabled in the org, so trying one more force-push.

martinpitt:
So, the actions [do run](https://github.com/martinpitt/anaconda/actions/runs/283848795), but they only appear against my own repo, not in the "Checks" here. WTH??

.github/workflows/validate.yml
```diff
@@ -0,0 +1,34 @@
...
+          ./autogen.sh
+          ./configure
+          make
+
+      - name: run unit tests
+        run: |
+          make check
+
+      - name: Upload unit test log
+        if: failure()
```
jkonecny12, line 30: Shouldn't we want to upload logs every time? It could be interesting from the debugging purposes.

jkonecny12 (approved):
Other than the below it looks great to me. Thanks!

.github/workflows/validate.yml
martinpitt: Sure, I can do that -- I thought let's not spam github with success logs, I also wanted to attach the coverage logs anyway, I'll look into that.

martinpitt: Done.

martinpitt:
Re-pushed with rebasing after #2882 landed, unconditionally uploading test logs, and including coverage logs into the artifacts. See https://github.com/martinpitt/anaconda/actions/runs/283947303 for the run/result. As discussed, I think we need to land this before it will appear as "Checks" in PRs properly, until then it will just appear in my own repo.

@jkonecny12 , PTAL?

martinpitt:
Ooops, lies! logs.zip doesn't contain the coverage results, hang on.

martinpitt:
I called `make check` instead of `make ci`, fixed now. Now the [Action run](https://github.com/martinpitt/anaconda/actions/runs/283992820) has coverage logs. @jkonecny12 : Now for real :grin:

jkonecny12 (approved):
Looks great to me. Great work!

PR #2885, opened 2020-10-02
Show in the first screen only translated locales

VladimirSlavik:
This decision is delegated to gettext. The most usual situations are:

- The locale does not have exactly matching translation and falls back to translation for the language.
- The locale has an exactly matching translation.
- The locale has neither a translation nor anything to fall back to.

Languages are already filtered in a similar manner, so there is no danger of having a language with no locale leading to an empty locale list and failing callbacks.

Resolves: [rhbz#1678787](https://bugzilla.redhat.com/show_bug.cgi?id=1678787)

### DIFF
--- a/pyanaconda/localization.py
+++ b/pyanaconda/localization.py
@@ -24,6 +24,7 @@ import langtable
 import locale as locale_mod
 import glob
 from collections import namedtuple
+import functools
 
 from pyanaconda.core import constants
 from pyanaconda.core.util import upcase_first_letter, setenv, execWithRedirect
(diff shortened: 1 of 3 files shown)


VladimirSlavik:
This is straight call to gettext stuff and interpreting its results, so I decided a test would be redundant. The thing to test would be the result in the gui, anyway.

TUI does not have this problem, since it is always in English.

martinpitt:
Ooooh, a proper GitHub action run!  :heart_eyes:

poncovka (approved):
It looks great! Thanks!

AdamWill:
This seems to have broken two things:

* https://bugzilla.redhat.com/show_bug.cgi?id=1890340
* https://bugzilla.redhat.com/show_bug.cgi?id=1890342

VladimirSlavik:
@AdamWill thanks for researching and filing these, I noticed the UK only yesterday too. Unfortunately "en" is special due to being "no translation", it seems. Not sure about the untranslated text.

PR #2888, opened 2020-10-02
Fix spelling/name of OSTree

VladimirSlavik:
It's spelled that way everywhere. I think it might be better to stick to that?

See https://translate.fedoraproject.org/translate/anaconda/master/en/?checksum=e440ad521e7a3ffe

### DIFF
--- a/tests/gettext_tests/style_guide.py
+++ b/tests/gettext_tests/style_guide.py
@@ -31,6 +31,7 @@ bad_strings = {'(?i)bootloader': 'boot loader',
                '[Cc]an not':     'cannot',
                '(?i)mountpoint': 'mount point',
                'Ok':             'OK',
+               '(?i)os tree ':   'OSTree',
                # Find instances of "return" that are referring to a keyboard key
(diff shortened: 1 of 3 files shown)


martinpitt (approved):
Thanks! That is the official spelling indeed.

poncovka (approved):
Thank you! I think that we have some tests that checks for the bad strings. Could you extend them, please?

VladimirSlavik:
Thanks for the idea and pointers, done!

tests/gettext_tests/style_guide.py
```diff
@@ -31,6 +31,7 @@
                '[Cc]an not':     'cannot',
                '(?i)mountpoint': 'mount point',
                'Ok':             'OK',
+               '(?i)os tree ':   'OSTree',
```
martinpitt, line 34: 💿🌲 clearly! :wink:

martinpitt (approved):
Cheers!

poncovka (approved):
Thanks!

PR #2890, opened 2020-10-05
Run rpm tests in a GitHub action

martinpitt:
This replaces the "Anaconda Rawhide test RPM files" Jenkins job. (When
that gets removed, a lot of Makefile rules can get cleaned up.)

As with the unit tests, use a standard Fedora rawhide container for now,
and install the anaconda build dependencies into it. Like with
unit tests, this should eventually be replaced with a custom container
image that gets refreshed regularly (gated by tests), to increase both
speed and reliability. But the Jenkins job has the exact same problems,
so this is not a regression.

### DIFF
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -33,3 +33,33 @@ jobs:
           path: |
             tests/test-suite.log
             tests/coverage-*.log
+
+  rpm-tests:
+    runs-on: ubuntu-latest
+    # start from a minimal container and install only our build deps; mock does
+    # not work in an unprivileged container (no CAP_SYS_ADMIN for mount), and
+    # dnf --installroot is too broken (rhbz#1885103, #1885101, #1738233)
+    container:
+      image: docker.io/fedora:rawhide
+    steps:
+      - name: Clone repository
+        uses: actions/checkout@v2
+
+      - name: install build dependencies
+        run: |
+          dnf install -y /usr/bin/xargs
+          scripts/testing/dependency_solver.py -b | xargs -d '\n' dnf install -y --setopt=install_weak_deps=False --nodocs rpm-build
+
+      - name: build RPMs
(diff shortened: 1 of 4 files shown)


martinpitt:
Not sure what that [packit failure](https://prod.packit.dev/srpm-build/8209/logs) means -- I didn't touch any of these bits here. Is that a common  flake? (I've heard that the packit infra is currently a bit wonky)

.github/workflows/validate.yml
```diff
@@ -33,3 +33,33 @@ jobs:
           path: |
             tests/test-suite.log
             tests/coverage-*.log
+
+  rpm-tests:
+    runs-on: ubuntu-latest
+    # start from a minimal container and install only our build deps; mock does
```
martinpitt, line 39: Some further context here: I deliberately do *not* want to run this in the same container (image) as the unit tests -- the latter require a lot more runtime/test dependencies, and I want to test the rpm build in a minimal environment (to avoid missing BuildRequires). I initially tried to do this with mock or a simple chroot inside the container, but that requires too many hacks. We really don't want to run a `--privileged` container, as that would severely hamper test infra portability -- you don't get  privileged containers on shared infrastructure.

martinpitt:
The failed jenkins job is due to

    Error: Will not install a source rpm package (anaconda-34.8-1.fc34.src).

(the jenkins log unhelpfully doesn't show that, I had to reproduce it in a VM).

mock puts the src.rpm into the built binary rpm. That's the reason for the weird `/[0-9]*.rpm` glob, so I put that back.

jkonecny12 (approved):
Looks great to me. Thanks for these improvements!

poncovka (approved):
Looks good to me. Thank you!

VladimirSlavik (approved):
Looks good to me. Thank you!

PR #2892, opened 2020-10-05
Add Dockerfile for anaconda unit-tests

jkonecny12:
This will install all dependencies to run the tests and add simple script to run tests by default or anything in the Anaconda directory.

Thanks to replacing the dependency_solver.py here we should be able to drop the script completely.

TODO:
- [x] Change README file in tests folder to explain container workflow.

### DIFF
--- a/dockerfile/ci-tasks/Dockerfile
+++ b/dockerfile/ci-tasks/Dockerfile
@@ -1,0 +1,40 @@
+FROM fedora:rawhide
+LABEL maintainer=anaconda-list@redhat.com
+
+# Prepare environment and install build dependencies
+RUN dnf update -y && \
+  dnf install -y \
+  'dnf-command(copr)' \
+  curl \
+  /usr/bin/xargs \
+  rpm-build \
+  e2fsprogs \
+  git \
+  bzip2 \
+  cppcheck \
+  rpm-ostree \
+  pykickstart \
+  python3-pip \
+  python3-lxml \
+  policycoreutils \
+  python3-gobject-base \
+  python3-pip && \
+  dnf copr enable -y @rhinstaller/Anaconda && \
+  dnf copr enable -y @storage/blivet-daily && \
+  curl -L https://raw.githubusercontent.com/rhinstaller/anaconda/master/anaconda.spec.in | sed 's/@PACKAGE_VERSION@/0/; s/@PACKAGE_RELEASE@/0/; s/%{__python3}/python3/' > /tmp/anaconda.spec && \
+  rpmspec -q --buildrequires /tmp/anaconda.spec | xargs -d '\n' dnf install -y && \
+  rpmspec -q --requires /tmp/anaconda.spec | grep -v anaconda | xargs -d '\n' dnf install -y && \
+  dnf clean all
+
+RUN pip install \
+  pocketlint \
+  coverage \
+  pycodestyle \
+  dogtail \
+  nose-testconfig \
+  rpmfluff
+
+RUN mkdir /anaconda
+
+WORKDIR /anaconda
+CMD ./autogen.sh && ./configure && make && make ci
--- a/tests/README.rst
+++ b/tests/README.rst
@@ -46,6 +46,29 @@ In case the *ci* target fails there is also a *coverage-report* target
 which can be used to combine the multiple `.coverage` files into one and
 produce a human readable report.
 
+Run tests inside of container
+-----------------------------
+
+Right now only unit tests are supported by the container, not rpm-tests.
+Before being able to run the tests you have to build the container.
+To build the container run::
+
+    make container-build
+
+Then you are free to run the tests without dependency installation by
+running::
+
+    make container-ci
+
+This will run all the tests.
+
+Logs from the run are stored in the `tests` folder.
+
+Note:
+
+Please update your container from time to time to have newest dependencies.
+To do that just run the build again.
+
 Run tests inside Mock
 ---------------------
 
--- a/Makefile.am
+++ b/Makefile.am
@@ -81,6 +81,12 @@ CONTAINER_BUILD_NAME ?= anaconda/ci-tasks
 CONTAINER_BUILD_TAG ?= latest
 CONTAINER_BUILD_ARGS ?=
 
+# Default argument to execute tests in the container
+CONTAINER_TEST_NAME ?= anaconda/ci-tasks
+CONTAINER_TEST_TAG ?= latest
+CONTAINER_RUNNER_TEST_ARGS ?=--rm -v .:/anaconda:Z
+CONTAINER_TEST_ARGS ?=
+
 SKIP_BRANCHING_CHECK ?= "false"
 
 # If translations are present, run tests on the .po files before tarring them
@@ -216,6 +222,9 @@ ci:
 		rm $(srcdir)/tests/error_occured; \
 		exit $$status; \
 	fi
+
+container-ci:
+	podman run $(CONTAINER_RUNNER_TEST_ARGS) $(CONTAINER_TEST_NAME):$(CONTAINER_TEST_TAG) $(CONTAINER_TEST_ARGS)
 
 check-branching:
 # checking if branching can be finished and all the pieces are in place
--- a/Makefile.am
+++ b/Makefile.am
@@ -75,6 +75,12 @@ SRPM_BUILD_DIR = $(BUILD_RESULT_DIR)/00-srpm-build
 RPM_BUILD_DIR = $(BUILD_RESULT_DIR)/01-rpm-build
 TEST_INST_BUILD_DIR = $(BUILD_RESULT_DIR)/02-test-install
 
+# Default variables for container build call
+DOCKERFILE_PATH ?= $(srcdir)/dockerfile/ci-tasks
+CONTAINER_BUILD_NAME ?= anaconda/ci-tasks
+CONTAINER_BUILD_TAG ?= latest
+CONTAINER_BUILD_ARGS ?=
+
 SKIP_BRANCHING_CHECK ?= "false"
 
 # If translations are present, run tests on the .po files before tarring them
@@ -151,6 +157,9 @@ release:
 release-and-tag:
 	$(MAKE) dist
 	$(MAKE) tag
+
+container-build:
+	podman build $(CONTAINER_BUILD_ARGS) -t $(CONTAINER_BUILD_NAME):$(CONTAINER_BUILD_TAG) $(DOCKERFILE_PATH)
 
 bumpver: po-pull
 	@opts="-n $(PACKAGE_NAME) -v $(PACKAGE_VERSION) -r $(PACKAGE_RELEASE) -b $(PACKAGE_BUGREPORT)" ; \


poncovka (approved):
It looks great! Thanks!

dockerfile/Dockerfile
```diff
@@ -0,0 +1,48 @@
...
+  dnf copr enable -y @rhinstaller/Anaconda && \
+  dnf copr enable -y @storage/blivet-daily
+
+RUN curl -L https://raw.githubusercontent.com/rhinstaller/anaconda/master/anaconda.spec.in -o /tmp/anaconda.spec && \
+  sed -i 's/@PACKAGE_VERSION@/0/' /tmp/anaconda.spec && \
+  sed -i 's/@PACKAGE_RELEASE@/0/' /tmp/anaconda.spec && \
+  sed -i 's/%{__python3}/python3/' /tmp/anaconda.spec && \
+  dnf builddep -y /tmp/anaconda.spec && \
+  rpmspec -q --requires /tmp/anaconda.spec | grep -v anaconda | xargs -d '\n' dnf install -y && \
+  dnf clean all
```
martinpitt, line 21: This does not help, and is just counter-productive here. The next RUN step again calls dnf, so this wipes the cache for that, and doesn't clean up after. Please merge two dnf steps into one `RUN` layer, and run `dnf clean all` as the last step.

dockerfile/Dockerfile
```diff
@@ -0,0 +1,48 @@
...
+  pocketlint \
+  coverage \
+  pycodestyle \
+  dogtail \
+  nose-testconfig \
+  rpmfluff
+
+COPY entrypoint.sh /root/entrypoint.sh
+
+VOLUME /workplace
```
martinpitt, line 46: This isn't being used anywhere, and will just cause pointlessly piling up nameless volumes. Let's please drop this until we actually need it (I don't think we ever will).

dockerfile/entrypoint.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+cd /workplace
```
martinpitt, line 3: This doesn't have to be a separate volume, the whole operation here doesn't take *that* much space? (Certainly less than the 10 GiB that one gets by default).

dockerfile/Dockerfile
```diff
@@ -0,0 +1,48 @@
...
+
+ENTRYPOINT ["/root/entrypoint.sh"]
```
martinpitt, line 48: I'm a bit sceptical to hardcode this into `ENTRYPOINT`, as that's relatively inconvenient to override. How about making this a `CMD`?

dockerfile/entrypoint.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
```
martinpitt, line 1: bikeshedding again -- how about naming that file `make-ci.sh`? We may have other scripts in the future that this container can run.

martinpitt (changes requested):
Thanks! This is a nice start!

jkonecny12:
**UPDATED:**
* Remove dnf-buiddep and use rpmspec twice instead. Avoid dragging in not needed dependencies.

dockerfile/entrypoint.sh
jkonecny12: It should be just a mount point for anaconda to enter. But yes, I guess I know where you are pointing.

dockerfile/Dockerfile
```diff
@@ -0,0 +1,48 @@
+FROM fedora:rawhide
```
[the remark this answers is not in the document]
jkonecny12: Not sure I understand you here. I will create a separate directory but why labels? They are just metadata (AFAIK) so I would add them when they are needed.

dockerfile/Dockerfile
```diff
@@ -0,0 +1,48 @@
...
+  /usr/bin/xargs \
+  rpm-build \
+  python3-pip && \
+  dnf copr enable -y @rhinstaller/Anaconda && \
+  dnf copr enable -y @storage/blivet-daily
+
+RUN curl -L https://raw.githubusercontent.com/rhinstaller/anaconda/master/anaconda.spec.in -o /tmp/anaconda.spec && \
+  sed -i 's/@PACKAGE_VERSION@/0/' /tmp/anaconda.spec && \
+  sed -i 's/@PACKAGE_RELEASE@/0/' /tmp/anaconda.spec && \
+  sed -i 's/%{__python3}/python3/' /tmp/anaconda.spec && \
```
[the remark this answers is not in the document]
jkonecny12: Done.

dockerfile/Dockerfile
jkonecny12: My bad, I missed that the next RUN is also dnf. Fixed.

dockerfile/Dockerfile
jkonecny12: I thought it's just telling container runtime that here you should expect the anaconda mounted from outside. I guess I missed something here.

jkonecny12:
**UPDATED:**
* Implemented points above.

dockerfile/Dockerfile
martinpitt: Yes, for each `VOLUME` that you declare, but don't have a `-v` for, docker/podman will create an "anonymous" volume (`podman volume ls`).

dockerfile/entrypoint.sh
martinpitt: Sure -- please do the `mkdir -p /workplace; cd /workplace`, that's fine -- it should just not be a "mandatory" mount point.

dockerfile/make-ready/Dockerfile
```diff
@@ -0,0 +1,38 @@
+FROM fedora:rawhide
```
martinpitt, line 1: In the future we probably want to turn this into a variable, but let's just start with `:rawhide` indeed.

martinpitt (changes requested):
Thanks for the updates! I would still like to drop the `VOLUME`, otherwise LGTM!

martinpitt (commented):
Oh, and can you please update the commit message to something like "Add unit test container" or something?

dockerfile/make-ready/Dockerfile
jkonecny12: You can have variables in the Dockefile?

jkonecny12:
> Thanks for the updates! I would still like to drop the `VOLUME`, otherwise LGTM!

It's dropped already, isn't it?

jkonecny12:
**UPDATED:**
* Renamed to ``ci-tasks``.

martinpitt (approved):
Thanks! And of course it said `WORKDIR`, not `VOLUME`, sorry for mis-reading :-)

dockerfile/ci-tasks/Dockerfile
```diff
@@ -0,0 +1,38 @@
...
+RUN pip install \
+  pocketlint \
+  coverage \
+  pycodestyle \
+  dogtail \
+  nose-testconfig \
+  rpmfluff
+
+WORKDIR /anaconda
+CMD ./autogen.sh && ./configure && make && make ci
```
martinpitt, line 38: Actually, I'm still not really happy about this. Just running that container will just fail, as /anaconda doesn't exist:
```
❱❱❱ podman run -it --rm anaconda/ci-tasks
Error: chdir: No such file or directory: OCI runtime command not found error
```

You would have to use `-v` to bind your local checkout into the container, which usually gives problems with SELinux, and it would also clutter your tree. So *if* using a bind mount, it should be a read-only (`:r`) one, and the container should copy the tree.

The GitHub pipeline would of course still use its own commands to do the checkout recipe etc. Thus what we do here should be optimized for human developers, not the machinery. One idea: If `/source` exists, copy it to /worktree and run tests there -- then you can safely run the container with `-v .:/source:ro` (→ make rule). And if /source does not exist, just clone current master?

Or, of course, if that's too much bikeshedding, let's just not specify a CMD for now, and optimize this later?

dockerfile/ci-tasks/Dockerfile
```diff
@@ -0,0 +1,38 @@
...
+  python3-pip \
+  python3-lxml \
+  policycoreutils \
+  python3-gobject-base \
+  python3-pip && \
+  dnf copr enable -y @rhinstaller/Anaconda && \
+  dnf copr enable -y @storage/blivet-daily && \
+  curl -L https://raw.githubusercontent.com/rhinstaller/anaconda/master/anaconda.spec.in | sed 's/@PACKAGE_VERSION@/0/; s/@PACKAGE_RELEASE@/0/; s/%{__python3}/python3/' > /tmp/anaconda.spec && \
+  rpmspec -q --buildrequires /tmp/anaconda.spec | xargs -d '\n' dnf install -y && \
+  rpmspec -q --requires /tmp/anaconda.spec | grep -v anaconda | xargs -d '\n' dnf install -y && \
```
poncovka, line 26: Does it have to be all part of one command?

poncovka (approved):

martinpitt (changes requested):
I tried `podman build -t anaconda/ci-tasks:latest .` on that, and it fails with

    Error unpacking rpm package iputils-20190515-8.fc33.x86_64
    error: unpacking of archive failed on file /usr/bin/arping;5f7b2384: cpio: cap_set_file
error: iputils-20190515-8.fc33.x86_64: install failed

This is essentially https://bugzilla.redhat.com/show_bug.cgi?id=1738233 again. Did that work for you @jkonecny12 ? Maybe that only happens with the podman/etc. versions on a Fedora 33 host. But it'll hopefully not be a problem with docker (we'll find out soon enough when we build this on dockerhub or a github pipeline), and with podman we can avoid this by actually allowing that capability:

    podman build --cap-add=CAP_SETFCAP -t anaconda/ci-tasks:latest .

(this should be part of the `Makefile` rule or whatever we use to locally build/run this).

After that, I can use the container with

    podman run -it --rm --workdir /tmp/ -v .:/source:ro anaconda/ci-tasks bash -i

(needs `--workdir` as /workdir doesn't exist) -- this is rather inconvenient. And inside,
```
cp -r /source/ /worktree
./autogen.sh && ./configure && make ci
```

This works fine :tada: 

Can you please add a `make build-ci-tasks` rule to do the build, and add to tests/README.rst how to build and use it? We'll need to extend it later to use `docker` or `podman` depending on what's available, but that's easy enough. Let's sort out the podman workflow first, for developers.

After that, let's land this, push a first container build to dockerhub (or the github container registry, doesn't matter much?), and change our workflow to use that container image.

dockerfile/ci-tasks/Dockerfile
martinpitt: Yes, due to the way [docker layers](https://dzone.com/articles/docker-layers-explained) work. Otherwise  the whole /var/cache/dnf/ would still be part of the container image, even though a later layer removes it. That'd be a waste of space and download time.

dockerfile/ci-tasks/Dockerfile
jkonecny12: You should use ``-v .:/anaconda:Z`` to solve the issue.

jkonecny12:
> I tried `podman build -t anaconda/ci-tasks:latest .` on that, and it fails with
> 
> ```
> Error unpacking rpm package iputils-20190515-8.fc33.x86_64
> error: unpacking of archive failed on file /usr/bin/arping;5f7b2384: cpio: cap_set_file error: iputils-20190515-8.fc33.x86_64: install failed
> ```
> 
> This is essentially https://bugzilla.redhat.com/show_bug.cgi?id=1738233 again. Did that work for you @jkonecny12 ? Maybe that only happens with the podman/etc. versions on a Fedora 33 host. But it'll hopefully not be a problem with docker (we'll find out soon enough when we build this on dockerhub or a github pipeline), and with podman we can avoid this by actually allowing that capability:

I don't have this problem at all on Fedora 32. Just running:

```
podman build -t anaconda:latest ./dockerfile/ci-tasks/
```

> 
> ```
> podman build --cap-add=CAP_SETFCAP -t anaconda/ci-tasks:latest .
> ```
> 
> (this should be part of the `Makefile` rule or whatever we use to locally build/run this).
> 
> After that, I can use the container with
> 
> ```
> podman run -it --rm --workdir /tmp/ -v .:/source:ro anaconda/ci-tasks bash -i
> ```
> 
> (needs `--workdir` as /workdir doesn't exist) -- this is rather inconvenient. And inside,
> 
> ```
> cp -r /source/ /worktree
> cd /worktree
> ./autogen.sh && ./configure && make ci
> ```
> 
> This works fine tada
> 
> Can you please add a `make build-ci-tasks` rule to do the build, and add to tests/README.rst how to build and use it? We'll need to extend it later to use `docker` or `podman` depending on what's available, but that's easy enough. Let's sort out the podman workflow first, for developers.

I'll add the make and README note. It would be great to also have the readme here but I think we don't have to create it here. Could be done later if needed.

> 
> After that, let's land this, push a first container build to dockerhub (or the github container registry, doesn't matter much?), and change our workflow to use that container image.

dockerfile/ci-tasks/Dockerfile
poncovka: I see. Thanks for the explanation!

dockerfile/ci-tasks/Dockerfile
martinpitt: Well yes, but that still clutters the tree on the host. That may be acceptable for a PoC, but it's easy enough to avoid?

jkonecny12:
**UPDATED:**
* Add new Makefile targets.

dockerfile/make-ready/Dockerfile
martinpitt: Yes, with [ARG](https://docs.docker.com/engine/reference/builder/#arg). A caveat is that they don't yet work in `FROM` on older docker versions (such as RHEL 7), but they work fine with podman.

martinpitt:
@jkonecny12 : This works great now, thanks! As I said, an r/w bind mount of the source dir is a bit weird, but (1) in the end that's your decision, (2) easy to change later on, and (3) not that important, so let's not block on this. There is a trivial merge conflict, so this needs rebasing. You added a README update todo item, do you want to do that in this PR still? Either way, good for me now. Dekuji!

jkonecny12:
**UPDATED:**
* Added README file note how to use container
  * We have a problem with running just a single test (will fix that in separate PR)

dockerfile/ci-tasks/Dockerfile
martinpitt: Let's do this in a follow-up, this involves some bike-shedding, and collecting some opinions what workflow is desired.

tests/README.rst
```diff
@@ -46,6 +46,29 @@ In case the *ci* target fails there is also a *coverage-report* target
 which can be used to combine the multiple `.coverage` files into one and
 produce a human readable report.
 
+Run tests inside of container
+-----------------------------
+
+Right now only unit tests are supported by the container not rpm-tests.
```
martinpitt, line 52:
```suggestion
Right now only unit tests are supported by the container, not rpm-tests.
```

tests/README.rst
```diff
@@ -46,6 +46,29 @@ In case the *ci* target fails there is also a *coverage-report* target
...
+    make container-build
+
+Then you are free to run the tests without dependency installation by
+running::
+
+    make container-ci
+
+This will run all the tests.
+
+Logs from the run are stored in the ``tests`` folder.
```
martinpitt, line 65:
```suggestion
Logs from the run are stored in the `tests` folder.
```

tests/README.rst
```diff
@@ -46,6 +46,29 @@ In case the *ci* target fails there is also a *coverage-report* target
...
+
+Note:
+
+Please update your container from time to time to have newest dependencies.
+To do that just run the build again.
```
martinpitt, line 70: We will actually fix that to refresh/autobuild the container on docker.io (or github registry) every week, so this note should soon go away :-)

martinpitt (approved):
Thanks! Two rather trivial typo fixes, otherwise LGTM!

tests/README.rst
jkonecny12: I would leave it there because it could be valuable to know that rebuild could solve your dependencies issues. And even if it is on dockerhub automatically updated you still have to do a pull.

jkonecny12:
**UPDATED:**
* Fixed issues found by @martinpitt . Thanks!

poncovka (approved):
It looks great! Thanks!

rvykydal (approved):
Looks good to me.

VladimirSlavik (approved):
Looks good to me. Thank you!

martinpitt:
This [packit crash](https://console-testing-farm.apps.ci.centos.org/pipeline/4d5782fc-1925-471e-a555-b70b5d42408f) is unrelated, and nothing here affects the RPM build.

M4rtinK (commented):

PR #2893, opened 2020-10-05
Stop ignoring "Run" in pylint

VladimirSlavik:
These days we have nothing with such a name.

### DIFF
--- a/tests/pylint/pylintrc
+++ b/tests/pylint/pylintrc
@@ -182,7 +182,6 @@ good-names=i,
            j,
            k,
            ex,
-           Run,
            _
 
 # Include a hint for the correct naming format with invalid-name.


poncovka (approved):
Looks good to me. However, I am not sure whether we run the check that uses this option.

martinpitt:
@poncovka : I would assume here? https://github.com/rhinstaller/anaconda/tree/master/tests/pylint

poncovka:
> @poncovka : I would assume here? https://github.com/rhinstaller/anaconda/tree/master/tests/pylint

Yes, but we have a lot of disabled pylint checks. See: https://github.com/rhinstaller/anaconda/blob/master/tests/pylint/pylintrc#L69

VladimirSlavik:
After digging a bit more, this is a default value. Let's ignore this instead.

PR #2895, opened 2020-10-05
Allow to format selected DASDs (#1874394)

poncovka:
TUI should allow to format selected DASDs the same way as GUI.

Resolves: rhbz#1874394

### DIFF
--- a/pyanaconda/ui/tui/spokes/storage.py
+++ b/pyanaconda/ui/tui/spokes/storage.py
@@ -258,6 +258,11 @@ class StorageSpoke(NormalTUISpoke):
                         # Wait for storage.
                         threadMgr.wait(THREAD_STORAGE)
 
+                        # Allow to format DASDs.
+                        # FIXME: Don't allow to initialize all disks with unrecognized formats.
+                        self._disk_init_module.SetFormatUnrecognizedEnabled(True)
+                        self._disk_init_module.SetFormatLDLEnabled(True)
+
                         # Get selected disks.
                         disks = filter_disks_by_names(self._available_disks, self._selected_disks)
 
--- a/pyanaconda/ui/gui/spokes/storage.py
+++ b/pyanaconda/ui/gui/spokes/storage.py
@@ -727,6 +727,7 @@ class StorageSpoke(NormalSpoke, StorageCheckHandler):
             return rc
 
         # Allow to format DASDs.
+        # FIXME: Don't allow to initialize all disks with unrecognized formats.
         self._disk_init_module.SetFormatUnrecognizedEnabled(True)
         self._disk_init_module.SetFormatLDLEnabled(True)
 


jstodola:
It works fine, thanks. 0.0.3527 unformatted, 0.0.3627 LDL formatted DASD:

```
Installation Destination

1) [x] DISK: 6.88 GiB (dasda), 0.0.3227
2) [x] DISK: 6.88 GiB (dasdb), 0.0.3027
3) [x] DISK: 6.88 GiB (dasdc), 0.0.3127
4) [x] DISK: 6.88 GiB (dasdd), 0.0.3327
5) [x] DISK: 6.88 GiB (dasde), 0.0.3427
6) [x] DISK: 0 B (dasdf), 0.0.3527
7) [x] DISK: 6.88 GiB (dasdg), 0.0.3627
8) [x] DISK: 6.88 GiB (dasdh), 0.0.3727
9) [x] Select all

8 disks selected; 48.15 GiB capacity; 5.95 GiB free

Please make a selection from the above ['c' to continue, 'q' to quit, 'r' to
refresh]: c
================================================================================
================================================================================
Question

The following unformatted or LDL DASDs have been detected on your system. You
can choose to format them now with dasdfmt or cancel to leave them unformatted.
Unformatted DASDs cannot be used during installation.

/dev/dasdg (0.0.3627)
/dev/dasdf (0.0.3527)

Warning: All storage changes made using the installer will be lost when you
choose to format.

Proceed to run dasdfmt?

Please respond 'yes' or 'no': yes
This may take a moment.
Formatting DASDs
Formatting dasdg
Formatting dasdf
Probing storage
================================================================================
================================================================================
Installation Destination

1) [x] DISK: 6.88 GiB (dasda), 0.0.3227
2) [x] DISK: 6.88 GiB (dasdb), 0.0.3027
3) [x] DISK: 6.88 GiB (dasdc), 0.0.3127
4) [x] DISK: 6.88 GiB (dasdd), 0.0.3327
5) [x] DISK: 6.88 GiB (dasde), 0.0.3427
6) [x] DISK: 6.88 GiB (dasdf), 0.0.3527
7) [x] DISK: 6.88 GiB (dasdg), 0.0.3627
8) [x] DISK: 6.88 GiB (dasdh), 0.0.3727
9) [x] Select all

8 disks selected; 55.03 GiB capacity; 19.7 GiB free
```

jstodola (approved):

VladimirSlavik (approved):
Looks good to me. Thank you!

pyanaconda/ui/tui/spokes/storage.py
```diff
@@ -258,6 +258,10 @@ def input(self, args, key):
                         # Wait for storage.
                         threadMgr.wait(THREAD_STORAGE)
 
+                        # Allow to format DASDs.
+                        self._disk_init_module.SetFormatUnrecognizedEnabled(True)
```
jkonecny12, line 262: Isn't this a bit dangerous? What about other disks which are not DASD, wouldn't this call enable us to destroy them too?

jkonecny12 (changes requested):
Not sure if I understand the call correctly. Please feel free to correct me if I'm wrong.

NACK just to be sure we won't miss something. If the above is not a problem then I'm changing it to ACK ;)

pyanaconda/ui/tui/spokes/storage.py
poncovka: Hmm, it looks like that. The partitioning module will initialize all disks with unrecognized formats. However, we already do the same in GUI and it was implemented this way from the beginning. The fix of this behaviour won't be trivial, so I suggest to go ahead with this solution and implement the proper one in upstream. What do you think?

VladimirSlavik: What would be these other unrecognized formats?

poncovka: I guess that every format that isn't supported by blivet? It is a DBus alternative to the kickstart command `zerombr`. This issue can happen only on 390x, so that might be the reason why it is not reported.

VladimirSlavik: So this is set to bite us in the future, once something (not-yet-)supported appears? Maybe extending the comment could help.

jkonecny12: Could you please add a comment with FIXME here?

poncovka: Fixed.

poncovka:
Added comments.

jkonecny12 (approved):
LGTM now!

VladimirSlavik (approved):
Re-acking.

PR #2897, opened 2020-10-06
Backport unit/rpm test fixes

martinpitt:
These are all cherry-picked from master. After that we will be able to run these tests in containers as well, and move them off Jenkins.

This is my first non-master PR, so please have an extra thorough look at the  branch target ("rhel-8" seems to be the active one, but there's also "rhel-8.0", "rhel8-alpha-branch", and "rhel8-branch"), and the labels.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -159,12 +159,6 @@ release-and-tag:
 	$(MAKE) dist
 	$(MAKE) tag
 
-rc-release: scratch
-	-rm -rf $(BUILD_RESULT_DIR)
-	mock -r $(MOCKCHROOT) $(MOCK_EXTRA_ARGS) --clean || exit 1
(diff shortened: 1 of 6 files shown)


poncovka:
The branch is correct, but we will need an approved bug for these commits. @jkonecny12, could you take care of that, please?

martinpitt:
@poncovka : OOI, why? Nothing here changes any runtime behaviour or the contents of the RPM.

martinpitt:
I filed https://bugzilla.redhat.com/show_bug.cgi?id=1885635 and added a "Related:" tag to all commit messages. No code changes.

poncovka (approved):
Looks good to me.

VladimirSlavik (approved):
Looks good to me. Thank you!

PR #2898, opened 2020-10-06
Add GitHub actions runner to ci-tasks container

martinpitt:
This allows us to start a self-hosted runner pretty much anywhere,
including locally (for debugging) or on internal infrastructure (for
RHEL tests).

Include a `github-action-run-once` script that does everything that's
required for this, so that the container invocation becomes fairly
simple (see comment at the top of the file).

 - [x] Builds on top of PR #2899, which changes the entrypoint; land that first

### DIFF
--- a/dockerfile/ci-tasks/github-action-run-once
+++ b/dockerfile/ci-tasks/github-action-run-once
@@ -1,0 +1,38 @@
+#!/bin/sh
+# Register a self-hosted runner on GitHub, wait for and run one request, and stop again.
+# This ensures that broken/malicious PRs cannot leave anything permanent behind on the host machine.
+# Inputs:
+#   $GITHUB_TOKEN: must have "public_repo" scope
+#   $GITHUB_REPOSITORY: owner/name
+#   $RUNNER_NAME: string, optional (defaults to container host name)
+#
+# Example invocation:
+# podman run -h local-runner-1 --rm -e GITHUB_TOKEN=$(< ~/.config/github-token) -e GITHUB_REPOSITORY=rhinstaller/anaconda \
+#        --entrypoint /github-action-run-once anaconda/ci-tasks
+#
+# Documentation:
+# https://docs.github.com/en/free-pro-team@latest/actions/hosting-your-own-runners/about-self-hosted-runners
+# https://developer.github.com/v3/actions/self-hosted-runners/
+
+set -eu
+
+export RUNNER_ALLOW_RUNASROOT=1
+cd /actions-runner
+
+# get the runner registration token
+RUNNER_TOKEN=$(curl --silent --show-error -X POST -u "token:$GITHUB_TOKEN" \
+    "https://api.github.com/repos/$GITHUB_REPOSITORY/actions/runners/registration-token" | \
+    sed -n '/"token"/ { s/.*: "//; s/".*$//; p }')
+if [ -z "$RUNNER_TOKEN" ]; then
+    echo "Failed to acquire runner token" >&2
+    exit 1
+fi
+
+# register runner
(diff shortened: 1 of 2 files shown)


martinpitt:
I tested this with an [extra commit](https://github.com/martinpitt/anaconda/commit/a58b98735e9ce00f06207f7bdacf2f6c333b0d77) that essentially changes `runs-on: ubuntu-latest` to `runs-on: ["self-hosted", "X64", "ci-tasks"]` (plus some extra noise). I started a local runner with

    podman run -h pittirunner1 -e GITHUB_TOKEN=$(< ~/.config/github-token) -e GITHUB_REPOSITORY=martinpitt/anaconda  --rm anaconda/ci-tasks github-action-run-once

which then appeared on [my fork's actions](https://github.com/martinpitt/anaconda/settings/actions):

image

The local container picks this up, runs that one request, the output [gets published on my repo's Actions](https://github.com/martinpitt/anaconda/runs/1214254712?check_suite_focus=true); then the container exits cleanly, and the runner gets deregistered:

```
√ Connected to GitHub

2020-10-06 10:26:31Z: Listening for Jobs
2020-10-06 10:26:41Z: Running job: unit-tests
2020-10-06 10:35:38Z: Job unit-tests completed with result: Succeeded

# Runner removal

√ Runner removed successfully
√ Removed .credentials
√ Removed .runner
```

In a real deployment, systemd would now start a fresh container for the next request.

This provides the necessary building blocks for my proposed schema how to run PRs on a self-hosted runner in a reasonably isolated manner, except that the systemd unit does not exist yet (I just started the container by hand for the first test):

sequence

dockerfile/ci-tasks/github-action-run-once
```diff
@@ -0,0 +1,37 @@
...
+RUNNER_TOKEN=$(curl --silent --show-error -X POST -u "token:$GITHUB_TOKEN" \
+    "https://api.github.com/repos/$GITHUB_REPOSITORY/actions/runners/registration-token" | \
+    sed -n '/"token"/ { s/.*: "//; s/".*$//; p }')
+if [ -z "$RUNNER_TOKEN" ]; then
+    echo "Failed to acquire runner token" >&2
+    exit 1
+fi
+
+# register runner
+./config.sh --labels 'ci-tasks' ${RUNNER_NAME:+--name} ${RUNNER_NAME:-} --replace --token "$RUNNER_TOKEN" --unattended --url "https://github.com/$GITHUB_REPOSITORY"
```
jkonecny12, line 31: This is a bit crazy bashism to me but I was able to understand what is happening there.

jkonecny12 (approved):
Looks good to me. Thanks!

dockerfile/ci-tasks/github-action-run-once
martinpitt: It's not a bashism, it's POSIX shell. That said, if you think it's weird, I'm happy to replace with an if condition and building some `$EXTRA_ARGS` or so. (That's why we have reviews, thanks for pointing out!)

VladimirSlavik: This one seems self-explanatory to me, and I have not been exposed to shells too much... Then again maybe trusting that hunch is not such a great idea. It's shell after all...

martinpitt:
**UPDATES**:
 - Rebase on  top of #2899, which changes from CMD to ENTRYPOINT. This changes the invocation of the container.
 - I updated the README and the detailled "how to test" comment above.
 - I also moved the default entrypoint from /root to / for convenience and consistency.

@jkonecny12 : thus nothing substantial, but do you want to re-review? (Also, should we configure github to drop approvals when the PR changes?)

martinpitt:
Rebased after landing #2899, no code changes.

PR #2900, opened 2020-10-06
Make it easier to run make commands

jkonecny12:
Use toolbox which will creates you container for current Fedora. It should make dependency hunting for autotools much easier.

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -6,6 +6,26 @@ help with implementing changes in Anaconda, please follow our
 `blog series <https://rhinstaller.wordpress.com/2019/10/11/anaconda-debugging-and-testing-part-1/>`_ or
 an `addon guide <http://rhinstaller.github.io/anaconda-addon-development-guide/index.html>`_ to create Anaconda addon.
 
+How to run make commands
+------------------------
+
+Anaconda has plenty of dependencies and because of that it's hard to set environment to
+with Anaconda properly. To get all the dependencies you are free to use script in the Anaconda
+repository.
+
+Follow these steps to keep your machine clean from all the Anaconda dependencies. It will
+create a container where you can install all the dependencies. If you are not interested in
+dealing with container just skip this part and continue on the next one::
+
+    sudo dnf install toolbox
+    toolbox create
+    toolbox enter
+
+To prepare the environment in the container or on your system just run these commands::
+
+    sudo ./scripts/testing/install_dependencies.sh
+    ./autogen.sh && ./configure
+
 How to Contribute to the Anaconda Installer (the short version)
 ----------------------------------------------------------------
 
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -9,12 +9,12 @@ an `addon guide <http://rhinstaller.github.io/anaconda-addon-development-guide/i
 How to Contribute to the Anaconda Installer (the short version)
 ----------------------------------------------------------------
 
-a) I want to contribute to the upstream Anaconda Installer (used in Fedora):
+1) I want to contribute to the upstream Anaconda Installer (used in Fedora):
 
 - open a pull request for the ``<next Fedora number>-devel`` branch (f25-devel, etc.)
 - check the *Commit Messages* section below for how to format your commit messages
 
-b) I want to contribute to the RHEL Anaconda installer:
+2) I want to contribute to the RHEL Anaconda installer:
 
 - open a pull request for the ``<RHEL number>-branch``  branch (rhel7-branch, etc.)
 - check the *Commits for RHEL Branches* section below for how to format your commit messages
@@ -141,18 +141,18 @@ Merging the Fedora ``devel`` branch back to the ``master`` branch
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 (Fedora 25 is used as an example, don't forget to use appropriate Fedora version.)
 
-Checkout and pull the master branch:
+Checkout and pull the master branch::
 
-``git checkout master``
-``git pull``
+    git checkout master
+    git pull
 
-Merge the Fedora devel branch to the master branch:
+Merge the Fedora devel branch to the master branch::
 
-``git merge --no-ff f25-devel``
+    git merge --no-ff f25-devel
 
-Push the merge to the remote:
+Push the merge to the remote::
 
-``git push origin master``
+    git push origin master
 
 Merging a GitHub pull request
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.rst
```diff
@@ -6,6 +6,34 @@ help with implementing changes in Anaconda, please follow our
...
+
+Anaconda has plenty of dependencies and because of that it's hard to set environment to
+with Anaconda properly. To get all the dependencies you are free to use script in the Anaconda
+repository. There are two ways how to approach.
+
+a) I want to have my host clean and run autotools from a container:
+
+``sudo dnf install toolbox``
+``toolbox create``
+``toolbox enter``
```
martinpitt, line 20: Please make this a single pre-formatted paragraph, like this it [renders wrongly](https://github.com/rhinstaller/anaconda/blob/0322be290e02b489eb71e9b52c5b03ca2215d9be/CONTRIBUTING.rst).

CONTRIBUTING.rst
```diff
@@ -6,6 +6,34 @@ help with implementing changes in Anaconda, please follow our
...
+
+Now, you are inside of the container.
+
+``sudo dnf install rpm-build``
```
martinpitt, line 24: Let's please move this into install_dependencies.sh.

CONTRIBUTING.rst
```diff
@@ -6,6 +6,34 @@ help with implementing changes in Anaconda, please follow our
...
+``sudo ./scripts/testing/install_dependencies.sh``
+``./autogen.sh && ./configure``
```
martinpitt, line 26: Same paragraph formatting as above.

CONTRIBUTING.rst
```diff
@@ -6,6 +6,34 @@ help with implementing changes in Anaconda, please follow our
...
 an `addon guide <http://rhinstaller.github.io/anaconda-addon-development-guide/index.html>`_ to create Anaconda addon.
 
+How to run make commands
+------------------------
+
+Anaconda has plenty of dependencies and because of that it's hard to set environment to
+with Anaconda properly. To get all the dependencies you are free to use script in the Anaconda
+repository. There are two ways how to approach.
+
+a) I want to have my host clean and run autotools from a container:
```
martinpitt, line 16: Let's link to the [toolbox docs](https://docs.fedoraproject.org/en-US/fedora-silverblue/toolbox/) here.

Nitpick: The ReST says "a)", but it gets rendered as "1.". Maybe just give in and say "1." in the source as well :grin:

scripts/testing/install_dependencies.sh
```diff
@@ -0,0 +1,33 @@
+#!/bin/bash
```
martinpitt, line 1: Please add `set -eu` to make sure our scripts are sane.

scripts/testing/install_dependencies.sh
```diff
@@ -0,0 +1,33 @@
...
+#
+# For direct host installation run:
+#
+#   $ sudo dnf install rpm-build  # to get rpmspec binary
+#   $ sudo ./scripts/testing/install_dependencies.sh
+#
+
+
+# remove all problematic pieces from anaconda spec to be able to get dependencies
+cat ./anaconda.spec.in | sed 's/@PACKAGE_VERSION@/0/; s/@PACKAGE_RELEASE@/0/; s/%{__python3}/python3/' > /tmp/anaconda.spec
```
martinpitt, line 28: Useless use of cat :-) `sed 'prog' file > destfile`

But more importantly, this is a script that likely runs on real systems, so no hardcoded file names in /tmp please. That can range anywhere from DoS to exploits. Please use `mktemp`.

What's problematic about `%{__python3}`?

scripts/testing/install_dependencies.sh
```diff
@@ -0,0 +1,33 @@
...
+
+# get all build requires dependencies from the spec file and strip out version
+# version could be problematic because of fedora version you are running and
+# they are mostly not important for automake
+rpmspec -q --buildrequires /tmp/anaconda.spec | sed 's/>=.*$//' | xargs -d '\n' dnf install
```
martinpitt, line 33: I suggest to append `$@` here. That way, one can call the script as `install_dependencies.sh -y` and make it non-interactive. An usual workflow is to rebuild your toolboxes regularly and often, and interactive questions are annoying.

martinpitt (changes requested):
Toolbox! :heart_eyes:

scripts/testing/install_dependencies.sh
jkonecny12: __python3 will break rpmspec, I added the same to the Dockerfile.

scripts/testing/install_dependencies.sh
jkonecny12: Good idea.

CONTRIBUTING.rst
jkonecny12: It's used in the document below already. So I'm making it consistent.

jkonecny12:
**UPDATED:**
* Implemented notes above from @martinpitt 
* Fix formatting of the contribution guidelines document

CONTRIBUTING.rst
```diff
@@ -6,6 +6,32 @@ help with implementing changes in Anaconda, please follow our
...
+``sudo ./scripts/testing/install_dependencies.sh``
+``./autogen.sh && ./configure``
+
+You can exit the container by pressing `CTRL+d` and call make commands locally until any
+Makefile file is changed.
+
+b) To install the dependencies to my machine directly:
+
+``sudo ./scripts/testing/install_dependencies.sh``
+``./autogen.sh && ./configure``
```
martinpitt: The commands are now exactly the same in toolbox and on the host (which is kinda the point). So how about instead of a) or b) you have a first paragraph which points out toolbox, and the second just says what to do regardless of the environment?

martinpitt (changes requested):
Thanks! There's still two outstanding issues from the previous review round (link to toolbox docs and mktemp), I resolved the other threads.

CONTRIBUTING.rst
```diff
@@ -13,34 +13,34 @@ Anaconda has plenty of dependencies and because of that it's hard to set environ
...
 
-a) I want to have my host clean and run autotools from a container:
+1) I want to have my host clean and run autotools from a container::
 
-``sudo dnf install toolbox``
-``toolbox create``
-``toolbox enter``
+    sudo dnf install toolbox
+    toolbox create
+    toolbox enter
```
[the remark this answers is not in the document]
jkonecny12: Fixed.

CONTRIBUTING.rst
jkonecny12: Good idea, fixed.

martinpitt (approved):
Make make easier! :rofl:  Dekuji!

scripts/testing/install_dependencies.sh
```diff
@@ -0,0 +1,39 @@
...
+# non-interactive you can call:
+#
+#   $ sudo ./scripts/testing/install_dependencies.sh -y
+
+set -eu
+
+dnf install $@ rpm-build
+
+# remove all problematic pieces from anaconda spec to be able to get dependencies
+sed 's/@PACKAGE_VERSION@/0/; s/@PACKAGE_RELEASE@/0/; s/%{__python3}/python3/' ./anaconda.spec.in > /tmp/anaconda.spec
```
martinpitt, line 34: Ah sorry, this really needs a proper temporary file, or put it somewhere else -- like ./anaconda-installdeps.spec (that's safe as well).

martinpitt (changes requested):

scripts/testing/install_dependencies.sh
jkonecny12: Good point!

jkonecny12: Fixed.

martinpitt (approved):
:+1: for real now, thank you!

VladimirSlavik (approved):
Looks good to me. Thank you!

PR #2902, opened 2020-10-06
post-scripts need to be sorted and anaconda log files relabeled

jstodola:
Sort the order of execution of post-scripts, since glob.glob() doesn't return a sorted list.
Relabel the copied anaconda log files - they could be labeled incorrectly, like in case of hawkey.log, see bug [1885772](https://bugzilla.redhat.com/show_bug.cgi?id=1885772).

Resolves: rhbz#[1870493](https://bugzilla.redhat.com/show_bug.cgi?id=1870493)

### DIFF
--- a/data/post-scripts/99-copy-logs.ks
+++ b/data/post-scripts/99-copy-logs.ks
@@ -28,3 +28,12 @@ elif [ -e /run/install/ks.cfg ]; then
 fi
 
 %end
+
+%post
+# Relabel the anaconda logs we've just coppied, since they could be incorrectly labeled, like
+# hawkey.log: https://bugzilla.redhat.com/show_bug.cgi?id=1885772.
+# Execution of this %post script will not be logged in the log files on the installed system.
+
+restorecon -ir /var/log/anaconda/
+
+%end
--- a/pyanaconda/kickstart.py
+++ b/pyanaconda/kickstart.py
@@ -562,7 +562,7 @@ def appendPostScripts(ksdata):
     scripts = ""
 
     # Read in all the post script snippets to a single big string.
-    for fn in glob.glob("/usr/share/anaconda/post-scripts/*ks"):
+    for fn in sorted(glob.glob("/usr/share/anaconda/post-scripts/*ks")):
         f = open(fn, "r")
         scripts += f.read()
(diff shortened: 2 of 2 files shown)


VladimirSlavik:
From the Anaconda perspective, we want to log all executions of `%post` scripts for debugging/truobleshooting purposes. These logs go to the install environment, so they need to be copied. That is done by a post script too. So there's a bit of a chicken and egg problem: To correct log contexts, they need to be in the final location, but the context fixing should be logged too.

Personally I think specifically fixing the one and only hawkey context in the "copy logs" script is what we should do - but only as a concession. It's a known problem, we know it's that, blah blah, still needs to be done somehow. In an ideal world, the file would have the right context straight away, just like all the other log files.

VladimirSlavik (approved):
Besides the meta-bikeshedding above, the change proposed here looks very good to me. Thank you for fighting nondeterminism!

VladimirSlavik (changes requested):
Sorry, spoke too soon. If this does not fix the bug, then use `Related:` instead of `Resolves:` in the commit message, please?

poncovka (approved):
Good catch! About the labeling, I've suggested to replace the Anaconda's %post scripts with tasks (for several reasons). Then we could easily do the relabeling twice. Once before we copy the files (so the call will be logged) and once after we copy the files (so the copied files have the right label).

jkonecny12 (approved):
Looks great to me other than what was already mentioned here ;).

jstodola:
Used `Related:` in the commit message of the first commit.
Additional commit included to relabel anaconda logs after copying them to the installed system.

VladimirSlavik (approved):
I think this is as good as can be. Let's see what others think...

poncovka (approved):
It looks great! Thanks!

jkonecny12 (approved):
LGTM!

VladimirSlavik:
Blocked since bug has not been accepted yet.

jstodola:
master branch PR: #2928

VladimirSlavik:
Port to master merged.

VladimirSlavik:
Bug acked, PR no longer blocked.

PR #2903, opened 2020-10-07
Fix SECTION headers in docstrings

martinpitt:
Make them coincide with the class names. The missing "Anaconda" prefix
causes a build failure with gobject-introspection ≥ 1.61.1 due to [1]:

    Namespace conflict: DocSection('BaseWindow')

See https://bugzilla.redhat.com/show_bug.cgi?id=1885825 for some
details.

[1] https://gitlab.gnome.org/GNOME/gobject-introspection/-/commit/346c0690fe62f614ecb1f55857ea72939d9c0f83

### DIFF
--- a/widgets/src/BaseStandalone.c
+++ b/widgets/src/BaseStandalone.c
@@ -22,7 +22,7 @@
 #include "intl.h"
 
 /**
- * SECTION: BaseStandalone
+ * SECTION: AnacondaBaseStandalone
  * @title: AnacondaBaseStandalone
  * @short_description: Abstract base class for standalone Anaconda windows.
  *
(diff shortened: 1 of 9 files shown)


martinpitt:
See [this workflow run](https://github.com/martinpitt/anaconda/runs/1218578334?check_suite_focus=true) for how the failure looks like with current rawhide.

This PR is urgent, as it will block every other PR. This also nicely demonstrates why we shouldn't run CI against rawhide, but do the ci-tests container build every week and gate on it -- so that regressions like this can be fixed orderly instead of in panic mode. :wink:

rvykydal (approved):
Looks good to me.

VladimirSlavik (approved):
Looks good to me. Thank you!

poncovka (approved):
Thanks!

jkonecny12 (approved):
Looks good to me. Thanks for fixing this!

PR #2905, opened 2020-10-07
Small fixes in unit tests

poncovka:
Partially revert the commit ca07d04 that allowed to temporarily ignore the new
`timezone` kickstart command. We use the latest version of this command now.

### DIFF
--- a/tests/nosetests/pyanaconda_tests/kickstart_specification_test.py
+++ b/tests/nosetests/pyanaconda_tests/kickstart_specification_test.py
@@ -398,7 +398,6 @@ class ModuleSpecificationsTestCase(unittest.TestCase):
 
     # Names of the kickstart commands and data that should be temporarily ignored.
     IGNORED_NAMES = {
-        "timezone",
     }
 
     def setUp(self):


tests/nosetests/pyanaconda_tests/module_storage_test.py
```diff
@@ -1467,11 +1467,6 @@ def mount_filesystems_test(self, blivet_run_program, makedirs, core_run_program,
         blivet_run_program.return_value = (0, "")
         core_run_program.return_value = (0, "")
         MountFilesystemsTask(storage).run()
-        # created the mount points
-        makedirs.assert_any_call('/mnt/sysimage/dev', 0o755)
-        # sysimage mounts happened
-        blivet_run_program.assert_any_call(
-                ['mount', '-t', 'bind', '-o', 'bind,defaults', '/dev', '/mnt/sysimage/dev'])
```
VladimirSlavik, line 1474: I guess these mockings still need to happen, even if you don't check them?

VladimirSlavik (approved):
Looks good to me. Thank you!

tests/nosetests/pyanaconda_tests/module_storage_test.py
poncovka: Yes, we have to mock these calls to prevent the actual mounting.

martinpitt: Yes, the calls must be mocked out, as actually doing them is both forbidden in containers and actively harmful on developer machines.

However, OOI: How is /dev target/architecture specific? I can't imagine a platform (aside from building a simple chroot maybe) where /dev isn't required. If /dev somehow gets in the way, then perhaps let's test /run? That is a firm OS requirement these days (/run *must* be a tmpfs).

FTR, I added these as the marching orders were "mock them out, but make sure the mounts will happen". So, not much objection to dropping this, but it improves coverage and I'm mostly curious where they get in the way. Thanks!

poncovka: This check was failing when I ran unit tests locally in my mock environment, but I am not able to reproduce it anymore. Weird. I have removed this commit from the pull request.

poncovka:
Updated.

M4rtinK (approved):
Looks good to me. :)

jkonecny12 (approved):
Looks good to me as far as I understand.

PR #2906, opened 2020-10-07
Fix UID and GID handling in GUI and D-Bus

VladimirSlavik:
Fix the one-too-high UID and GID limit in GUI, as well as D-Bus limiting the values as they pass through.

Resolves: [rhbz#1786680](https://bugzilla.redhat.com/show_bug.cgi?id=1786680)

### DIFF
--- a/pyanaconda/core/constants.py
+++ b/pyanaconda/core/constants.py
@@ -486,3 +486,7 @@ RHSM_SERVICE_TIMEOUT = 90.0
 
 # Path to the System Purpose configuration file on a system.
 RHSM_SYSPURPOSE_FILE_PATH = "/etc/rhsm/syspurpose/syspurpose.json"
+
+# GID and UID modes
+ID_MODE_USE_VALUE = "ID_MODE_USE_VALUE"
+ID_MODE_USE_DEFAULT = "ID_MODE_USE_DEFAULT"
--- a/pyanaconda/modules/common/structures/group.py
+++ b/pyanaconda/modules/common/structures/group.py
@@ -22,7 +22,7 @@ from dasbus.typing import *  # pylint: disable=wildcard-import
 
 __all__ = ["GROUP_GID_NOT_SET", "GroupData"]
 
-GROUP_GID_NOT_SET = -1
+GROUP_GID_NOT_SET = 2**32-1
 
 
 class GroupData(DBusData):
@@ -50,7 +50,7 @@ class GroupData(DBusData):
         self._name = name
 
     @property
-    def gid(self) -> Int:
+    def gid(self) -> UInt32:
         """The GID of the group.
 
         If not provided, defaults to the next available non-system GID.
@@ -65,5 +65,5 @@ class GroupData(DBusData):
         return self._gid
 
     @gid.setter
-    def gid(self, gid: Int):
+    def gid(self, gid: UInt32):
         self._gid = gid
--- a/pyanaconda/modules/common/structures/group.py
+++ b/pyanaconda/modules/common/structures/group.py
@@ -20,9 +20,9 @@ from dasbus.structure import DBusData
 
 from dasbus.typing import *  # pylint: disable=wildcard-import
 
-__all__ = ["GROUP_GID_NOT_SET", "GroupData"]
+from pyanaconda.core.constants import ID_MODE_USE_DEFAULT, ID_MODE_USE_VALUE
 
-GROUP_GID_NOT_SET = 2**32-1
+__all__ = ["GroupData"]
 
 
 class GroupData(DBusData):
@@ -30,7 +30,8 @@ class GroupData(DBusData):
 
     def __init__(self):
         self._name = ""
-        self._gid = GROUP_GID_NOT_SET
+        self._gid = 0
+        self._gid_mode = ID_MODE_USE_DEFAULT
 
     @property
     def name(self) -> Str:
@@ -50,12 +51,29 @@ class GroupData(DBusData):
         self._name = name
 
     @property
+    def gid_mode(self) -> Str:
+        """Mode of the GID.
+
+        Contains a string describing the mode of the group's GID: Use the value or default.
+
+        Possible values are:
+        - "ID_MODE_USE_VALUE"
+        - "ID_MODE_USE_DEFAULT"
+
+        :return: the mode
+        :rtype str:
+        """
+        return self._gid_mode
+
+    @gid_mode.setter
+    def gid_mode(self, status: Str):
+        self._gid_mode = status
+
+    @property
     def gid(self) -> UInt32:
         """The GID of the group.
 
-        If not provided, defaults to the next available non-system GID.
-
-        GID equal to -1 means that a valid GID has not been set.
+        If ignored due to gid_mode, defaults to the next available non-system GID.
 
         For examples: 1234
 
@@ -67,3 +85,31 @@ class GroupData(DBusData):
     @gid.setter
     def gid(self, gid: UInt32):
         self._gid = gid
+
+    def get_gid(self):
+        """Return a GID value which can be a number or None.
+
+        Prefer using this method instead of directly reading gid and gid_mode.
+
+        :return: GID or None if not set
+        :rtype: int or None
+        """
+        if self._gid_mode == ID_MODE_USE_DEFAULT:
+            return None
+        else:
+            return self._gid
+
+    def set_gid(self, new_gid):
+        """Set GID value and mode from a value which can be None.
+
+        Prefer using this method instead of directly writing gid and gid_mode.
+
+        :param new_gid: new GID
+        :type new_gid: int or None
+        """
+        if new_gid is not None:
+            self._gid = new_gid
+            self._gid_mode = ID_MODE_USE_VALUE
+        else:
+            self._gid = 0
+            self._gid_mode = ID_MODE_USE_DEFAULT
--- a/pyanaconda/modules/common/structures/user.py
+++ b/pyanaconda/modules/common/structures/user.py
@@ -21,10 +21,9 @@ from dasbus.structure import DBusData
 from dasbus.structure import generate_string_from_data
 from dasbus.typing import *  # pylint: disable=wildcard-import
 
-__all__ = ["USER_GID_NOT_SET", "USER_UID_NOT_SET", "UserData"]
+from pyanaconda.core.constants import ID_MODE_USE_DEFAULT, ID_MODE_USE_VALUE
 
-USER_GID_NOT_SET = 2**32-1
-USER_UID_NOT_SET = 2**32-1
+__all__ = ["UserData"]
 
 
 class UserData(DBusData):
@@ -32,9 +31,11 @@ class UserData(DBusData):
 
     def __init__(self):
         self._name = ""
-        self._uid = USER_UID_NOT_SET
+        self._uid = 0
+        self._uid_mode = ID_MODE_USE_DEFAULT
         self._groups = list()
-        self._gid = USER_UID_NOT_SET
+        self._gid = 0
+        self._gid_mode = ID_MODE_USE_DEFAULT
         self._homedir = ""
         self._password = ""
         self._is_crypted = True
@@ -60,16 +61,33 @@ class UserData(DBusData):
         self._name = name
 
     @property
+    def uid_mode(self) -> Str:
+        """Mode of UID.
+
+        Contains a string describing the mode of the user's UID: Use the value or default.
+
+        Possible values are:
+        - "ID_MODE_USE_VALUE"
+        - "ID_MODE_USE_DEFAULT"
+
+        :return: the mode
+        :rtype str:
+        """
+        return self._uid_mode
+
+    @uid_mode.setter
+    def uid_mode(self, status: Str):
+        self._uid_mode = status
+
+    @property
     def uid(self) -> UInt32:
         """The users UID.
 
-        If not provided, this defaults to the next available non-system UID.
+        If ignored due to uid_mode, this defaults to the next available non-system UID.
 
-        For examples: 1234
+        For example: 1234
 
-        UID equal to -1 means that a valid UID has not been set.
-
-        :return: users UID
+        :return: user's UID
         :rtype: int
         """
         return self._uid
@@ -77,6 +95,34 @@ class UserData(DBusData):
     @uid.setter
     def uid(self, uid: UInt32):
         self._uid = uid
+
+    def get_uid(self):
+        """Return a UID value which can be a number or None.
+
+        Prefer using this method instead of directly reading uid and uid_mode.
+
+        :return: UID or None if not set
+        :rtype: int or None
+        """
+        if self._uid_mode == ID_MODE_USE_DEFAULT:
+            return None
+        else:
+            return self._uid
+
+    def set_uid(self, new_uid):
+        """Set UID value and mode from a value which can be None.
+
+        Prefer using this method instead of directly setting uid and uid_mode.
+
+        :param new_uid: new UID
+        :type new_uid: int or None
+        """
+        if new_uid is not None:
+            self._uid = new_uid
+            self._uid_mode = ID_MODE_USE_VALUE
+        else:
+            self._uid = 0
+            self._uid_mode = ID_MODE_USE_DEFAULT
 
     @property
     def groups(self) -> List[Str]:
@@ -96,14 +142,31 @@ class UserData(DBusData):
         self._groups = groups
 
     @property
+    def gid_mode(self) -> Str:
+        """Mode of GID.
+
+        Contains a string describing the mode of the user's GID: Use the value or default.
+
+        Possible values are:
+        - "ID_MODE_USE_VALUE"
+        - "ID_MODE_USE_DEFAULT"
+
+        :return: the mode
+        :rtype str:
+        """
+        return self._gid_mode
+
+    @gid_mode.setter
+    def gid_mode(self, status: Str):
+        self._gid_mode = status
+
+    @property
     def gid(self) -> UInt32:
         """The GID of the user’s primary group.
 
-        If not provided,  defaults to the next available non-system GID.
+        If ignored due to gid_mode, defaults to the next available non-system GID.
 
-        GID equal to -1 means that a valid GID has not been set.
-
-        For examples: 1234
+        For example: 1234
 
         :return: primary group GID
         :rtype: int
@@ -113,6 +176,34 @@ class UserData(DBusData):
     @gid.setter
     def gid(self, gid: UInt32):
         self._gid = gid
+
+    def get_gid(self):
+        """Return a GID value which can be a number or None.
+
+        Prefer using this method instead of directly reading gid and gid_mode.
+
+        :return: GID or None if not set
+        :rtype: int or None
+        """
+        if self._gid_mode == ID_MODE_USE_DEFAULT:
+            return None
+        else:
+            return self._gid
+
+    def set_gid(self, new_gid):
+        """Set GID value and mode from a value which can be None.
+
+        Prefer using this method instead of directly writing gid and gid_mode.
+
+        :param new_gid: new GID
+        :type new_gid: int or None
+        """
+        if new_gid is not None:
+            self._gid = new_gid
+            self._gid_mode = ID_MODE_USE_VALUE
+        else:
+            self._gid = 0
+            self._gid_mode = ID_MODE_USE_DEFAULT
 
     @property
     def homedir(self) -> Str:
--- a/pyanaconda/modules/common/structures/user.py
+++ b/pyanaconda/modules/common/structures/user.py
@@ -23,8 +23,8 @@ from dasbus.typing import *  # pylint: disable=wildcard-import
 
 __all__ = ["USER_GID_NOT_SET", "USER_UID_NOT_SET", "UserData"]
 
-USER_GID_NOT_SET = -1
-USER_UID_NOT_SET = -1
+USER_GID_NOT_SET = 2**32-1
+USER_UID_NOT_SET = 2**32-1
 
 
 class UserData(DBusData):
@@ -60,7 +60,7 @@ class UserData(DBusData):
         self._name = name
 
     @property
-    def uid(self) -> Int:
+    def uid(self) -> UInt32:
         """The users UID.
 
         If not provided, this defaults to the next available non-system UID.
@@ -75,7 +75,7 @@ class UserData(DBusData):
         return self._uid
 
     @uid.setter
-    def uid(self, uid: Int):
+    def uid(self, uid: UInt32):
         self._uid = uid
 
     @property
@@ -96,7 +96,7 @@ class UserData(DBusData):
         self._groups = groups
 
     @property
-    def gid(self) -> Int:
+    def gid(self) -> UInt32:
         """The GID of the user’s primary group.
 
         If not provided,  defaults to the next available non-system GID.
@@ -111,7 +111,7 @@ class UserData(DBusData):
         return self._gid
 
     @gid.setter
-    def gid(self, gid: Int):
+    def gid(self, gid: UInt32):
         self._gid = gid
 
     @property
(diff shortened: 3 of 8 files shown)


pyanaconda/modules/common/structures/group.py
```diff
@@ -50,7 +50,7 @@ def name(self, name: Str):
         self._name = name
 
     @property
-    def gid(self) -> Int:
+    def gid(self) -> Int64:
```
martinpitt, line 53: That's still not correct -- u/gids can never be negative, nor so large. Can you please use a proper `UInt32` instead?

martinpitt (changes requested):
Thanks!

martinpitt (changes requested):
Thanks! But can you please fix the typo in the commit message? In this case I think that's not trivial, as it's plausible that someone wants to grep `git log` for "UInt32".

VladimirSlavik:
So, we did put a negative value in *IDs, to signal "not set". Somewhat confusingly, the raw bits somewhere in the actual system also reserve the last value for something, which can be re-interpreted as -1. That threw me off a bit. But it is actually possible to use just `SInt32` to pass the value, if I change the "not set" value from -1 to the maximum which is reserved.

I'll see if the tests pass and then reverse the order of commits to remove the "differently broken" state with the d-bus fix but w/o the gui fix. Then I'll see about the commit messages that need more explaining I think.

VladimirSlavik:
Pitti yup, got that. Actually I wrote that first, submitted, and only then saw you re-reviewed in between.

jkonecny12 (approved):
Interesting change. Thanks for the fix!

VladimirSlavik:
On a second thought, let's run this through the user-focused kickstart tests to see if the renumbering brings any problems.

VladimirSlavik:
Looks like all the (kickstart) tests blew up on some circular lock in btrfs, all I have is a very large collection of kernel stack traces.

pyanaconda/modules/common/structures/group.py
```diff
@@ -22,7 +22,7 @@
 
 __all__ = ["GROUP_GID_NOT_SET", "GroupData"]
 
-GROUP_GID_NOT_SET = -1
+GROUP_GID_NOT_SET = 4294967295
```
poncovka, line 25: This is a very unfriendly default. Also, we cannot be really sure that this value will be never valid in the future. Could you please, keep `-1` as a default and use `Int64`?

An alternative solution is to introduce new properties `gid_type` and `uid_type` that will return `DEFAULT` to indicate that the installer will use a default ID, or `VALUE` to indicate that the installer will use a specified ID.

poncovka (commented):
I have some suggestions.

pyanaconda/modules/common/structures/group.py
VladimirSlavik: Yes, agreed. I just did s/UInt32/Int64/ to return this to the state before @martinpitt review. Basically it's now 2:1 that we want to waste four bytes so we can keep our -1.

martinpitt: It's not really about waste -- that's trivial. It's about picking the data type that matches what's going on in the kernel, and that's 32 bit. It can neither be negative, nor be > 2³², so having an API that accepts either is just buggy.

martinpitt: Also, `(uint32) -1` is exactly that magic number, so you can keep -1 even with uint32.

martinpitt (changes requested):
see above

pyanaconda/modules/common/structures/group.py
VladimirSlavik: I don't think the API really matters as much here. The whole user creation is just *wrong*. By design of the ancients, it must fail silently, and it must also keep backwards compatibility ad infinitum. Since there is no option to deprecate, it is impossible to change. *That* is the true reason why the value is not checked anywhere, it would be useless.

For that reason it's also pointless to have hard crashes when the untreated user inputs exceed the data type while sending the inputs over some transport layer (current state with `Int`). This change basically ensures the failures stay silent, same as the rest of the program does. The error surfaces with execution of `useradd` where it is caught and ignored.

VladimirSlavik:
Team discussion outcome: We want to separate the actual number and any special values. So there should be some kind of additional status flag, preferably as a string.

VladimirSlavik:
Update:
This is not pretty, just the first step so we can discuss the important things like arrangement of things and names. Also shows what else needs to be done so that Anaconda does not crash with *some* of the corner cases related to UIDs & GIDs.

I'm not squashing yet, let's see first what needs changing.

pyanaconda/modules/common/structures/group.py
```diff
@@ -67,3 +66,12 @@ def gid(self) -> Int64:
     @gid.setter
     def gid(self, gid: Int64):
         self._gid = gid
+
+    @property
+    def gid_status(self) -> Str:
+        """Status of the GID."""
```
jkonecny12, line 72: You should have example of possible values here.

pyanaconda/modules/common/structures/user.py
```diff
@@ -78,6 +77,15 @@ def uid(self) -> Int64:
     def uid(self, uid: Int64):
         self._uid = uid
 
+    @property
+    def uid_status(self) -> Str:
+        """Status of UID."""
```
jkonecny12, line 82: Same here.

pyanaconda/modules/common/structures/user.py
```diff
@@ -114,6 +120,15 @@ def gid(self) -> Int64:
     def gid(self, gid: Int64):
         self._gid = gid
 
+    @property
+    def gid_status(self) -> Str:
+        """Status of GID."""
```
jkonecny12, line 125: Same here.

pyanaconda/modules/users/users.py
```diff
@@ -217,8 +220,14 @@ def _ksdata_to_user_data(self, user_ksdata):
         # We need to make sure we correctly convert between these two.
         if user_ksdata.uid is not None:
             user_data.uid = user_ksdata.uid
+            user_data.uid_status = ""
+        else:
+            user_data.uid_status = ID_NOT_SET
```
jkonecny12, line 226: You can simplify this by abstracting this logic into the structure class. It would still have data as strings but we don't have to bother when working with those.

pyanaconda/ui/gui/spokes/user.py
```diff
@@ -118,8 +117,8 @@ def refresh(self):
         self._tHome.set_text(homedir)
         self._origHome = homedir
 
-        self._cUid.set_active(self.user.uid != USER_UID_NOT_SET)
-        self._cGid.set_active(self.user.gid != USER_GID_NOT_SET)
+        self._cUid.set_active(self.user.uid_status != constants.ID_NOT_SET)
+        self._cGid.set_active(self.user.gid_status != constants.ID_NOT_SET)
```
jkonecny12, line 121: The same here, just create a method in the structure class for this check.

jkonecny12 (changes requested):
I like the new solution. It just needs a bit more polishing.

pyanaconda/modules/users/users.py
```diff
@@ -215,10 +218,17 @@ def _ksdata_to_user_data(self, user_ksdata):
...
         if user_ksdata.uid is not None:
-            user_data.uid = user_ksdata.uid
+            user_data.uid = min(user_ksdata.uid, MAX_ID_VALUE)
+            user_data.uid_status = ""
+        else:
+            user_data.uid_status = ID_NOT_SET
         if user_ksdata.gid is not None:
-            user_data.gid = user_ksdata.gid
+            user_data.gid = min(user_ksdata.gid, MAX_ID_VALUE)
+            user_data.gid_status = ""
```
poncovka, line 229: So the status will be an empty string or `not set`. I don't know what I would expect from a status that doesn't tell me anything. And I have no idea how to interpret a configuration with `gid` set to `12345` and `gid_status` set to `not_set`.

First of all, I would like to avoid `status` in the property name. I think it has a little different meaning. What do you think about `uid_mode` and `gid_mode`?

Second, the property has to use meaningful values. I would suggest `USE_DEFAULT` and `USE_VALUE` (or just `DEFAULT` and `VALUE`). When `gid` is not set, then `groupadd` chooses a reasonable default. It works the same way with `uid` and `useradd`.

pyanaconda/modules/users/users.py
poncovka: Ideally, there would be methods `set_uid` and `get_uid` that would allow to set and get `None`. It will require a lot of additional code, because we need the same set of methods three times, but I don't know about a better solution.

pyanaconda/ui/gui/spokes/user.py
poncovka: I don't think we need an extra method just for the check. We might be able to use `get_uid` and `get_gid` as suggested above.

pyanaconda/modules/users/users.py
```diff
@@ -215,10 +218,17 @@ def _ksdata_to_user_data(self, user_ksdata):
...
+        else:
+            user_data.gid_status = ID_NOT_SET
```
poncovka, line 231: Please, use the fact that `user_data` is a new instance with default values. You don't have to rewrite defaults with defaults.

pyanaconda/modules/users/users.py
```diff
@@ -99,7 +99,10 @@ def process_kickstart(self, data):
             group_data = GroupData()
             group_data.name = group_ksdata.name
             if group_ksdata.gid is not None:
-                group_data.gid = group_ksdata.gid
+                group_data.gid = min(group_ksdata.gid, MAX_ID_VALUE)
+                group_data.gid_status = ""
+            else:
+                group_data.gid_status = ID_NOT_SET
```
poncovka, line 105: Same here.

pyanaconda/modules/users/users.py
```diff
@@ -215,10 +218,17 @@ def _ksdata_to_user_data(self, user_ksdata):
         # -> as user data is -1 by default ve only set it if kickstart has something,
         #    that is not None
         # We need to make sure we correctly convert between these two.
+        # We also cannot use values larger then the accepted range.
         if user_ksdata.uid is not None:
-            user_data.uid = user_ksdata.uid
+            user_data.uid = min(user_ksdata.uid, MAX_ID_VALUE)
+            user_data.uid_status = ""
+        else:
+            user_data.uid_status = ID_NOT_SET
```
poncovka, line 226: Same here.

poncovka (commented):
I have some suggestions.

pyanaconda/modules/common/structures/group.py
```diff
@@ -65,5 +64,14 @@ def gid(self) -> Int:
         return self._gid
 
     @gid.setter
-    def gid(self, gid: Int):
+    def gid(self, gid: UInt32):
         self._gid = gid
+
+    @property
+    def gid_status(self) -> Str:
```
poncovka, line 71: Please, define these properties before the original ones since these properties decide whether the original ones will be used or not. It is just easier to read the code that way.

pyanaconda/modules/users/users.py
jkonecny12: Good idea.

VladimirSlavik:
WIP, no need to re-review yet. (Will delete the comment once no longer true.)

pyanaconda/modules/users/users.py
VladimirSlavik: Done (I think).

pyanaconda/modules/users/users.py
VladimirSlavik: Thank you, I did just that.

pyanaconda/modules/common/structures/group.py
VladimirSlavik: Done.

VladimirSlavik:
Where we stand now: This is a WIP.

- Did not squash the commits yet.
- Tests still TODO.
- Needs actual testing after the rewrite too.

VladimirSlavik:
So, apart from the above, I do not really understand why the mount tests fail. This should be entirely unrelated.

pyanaconda/core/constants.py
```diff
@@ -486,3 +486,7 @@ class PayloadRequirementType(Enum):
 
 # Path to the System Purpose configuration file on a system.
 RHSM_SYSPURPOSE_FILE_PATH = "/etc/rhsm/syspurpose/syspurpose.json"
+
+# GID and UID status
+ID_USE_VALUE = "USE VALUE"
+ID_USE_DEFAULT = "USE DEFAULT"
```
poncovka, line 492: I would use the names `ID_MODE_USE_VALUE` and `ID_MODE_USE_DEFAULT` to avoid possible confusion. Please, use underscores instead of spaces in the values. It is pretty much our convention.

pyanaconda/core/constants.py
```diff
@@ -486,3 +486,7 @@ class PayloadRequirementType(Enum):
 
 # Path to the System Purpose configuration file on a system.
 RHSM_SYSPURPOSE_FILE_PATH = "/etc/rhsm/syspurpose/syspurpose.json"
+
+# GID and UID status
```
poncovka, line 490: `The UID and GID modes`

pyanaconda/modules/common/structures/group.py
```diff
@@ -50,12 +51,27 @@ def name(self, name: Str):
...
     @property
-    def gid(self) -> Int:
-        """The GID of the group.
+    def gid_mode(self) -> Str:
+        """Mode of the GID.
+
+        Contains a string describing the mode of the group's GID: Use the value or default.
 
-        If not provided, defaults to the next available non-system GID.
+        Example value: "USE DEFAULT"
```
poncovka, line 59: Please, provide a full list of supported values and their meanings.

pyanaconda/modules/common/structures/group.py
```diff
@@ -50,12 +51,27 @@ def name(self, name: Str):
...
+    @gid_mode.setter
+    def gid_mode(self, status: Str):
+        self._gid_mode = status
+
+    @property
+    def gid(self) -> UInt32:
+        """The GID of the group.
 
-        GID equal to -1 means that a valid GID has not been set.
+        If ignored due to mode, defaults to the next available non-system GID.
```
poncovka, line 74: Please, be more specific. Which mode?

pyanaconda/modules/common/structures/group.py
```diff
@@ -65,5 +81,29 @@ def gid(self) -> Int:
         return self._gid
 
     @gid.setter
-    def gid(self, gid: Int):
+    def gid(self, gid: UInt32):
         self._gid = gid
+
+    def get_gid(self):
+        """Return a value suitable for use in Kickstart.
```
poncovka, line 88: Please, don't use the name `kickstart` in DBus structures. These methods are useful for other use cases, that's why they are implemented here and not in the module.

pyanaconda/modules/common/structures/group.py
```diff
@@ -65,5 +81,29 @@ def gid(self) -> Int:
...
+        :return: GID or None if not set
+        :rtype: int or None
+        """
+        if self._gid_mode == ID_USE_DEFAULT:
+            return None
+        else:
+            return self._gid
+
+    def set_gid(self, new_gid):
+        """Set value and mode from a Kickstart value which can be None.
```
poncovka, line 99: Same here.

pyanaconda/modules/users/users.py
```diff
@@ -215,10 +212,8 @@ def _ksdata_to_user_data(self, user_ksdata):
         # -> as user data is -1 by default ve only set it if kickstart has something,
         #    that is not None
         # We need to make sure we correctly convert between these two.
```
poncovka, line 214: I think that you can completely remove the comment.

pyanaconda/modules/users/users.py
```diff
@@ -244,14 +239,12 @@ def _user_data_to_ksdata(self, user_ksdata, user_data):
         user_ksdata.groups = user_data.groups
         # To denote that a value has not been set:
         # - kickstart uses None
-        # - our DBus API uses -1
+        # - our DBus API uses a separate property
         # -> as ksdata has None as default, we simply only set the value if
-        #    it is != -1 on our side
+        #    *id_status != ID_NOT_SET on our side
         # We need to make sure we correctly convert between these two.
```
poncovka, line 245: Same here.

pyanaconda/ui/gui/spokes/user.py
```diff
@@ -118,8 +117,8 @@ def refresh(self):
         self._tHome.set_text(homedir)
         self._origHome = homedir
 
-        self._cUid.set_active(self.user.uid != USER_UID_NOT_SET)
-        self._cGid.set_active(self.user.gid != USER_GID_NOT_SET)
+        self._cUid.set_active(self.user.uid_mode != constants.ID_USE_DEFAULT)
+        self._cGid.set_active(self.user.gid_mode != constants.ID_USE_DEFAULT)
```
poncovka, line 121: Maybe you could use `get_gid() is not None`.

poncovka (changes requested):
The code looks very nice with the new methods. It just requires a few improvements.

VladimirSlavik:
Thank you, that all looks valid.

I wonder, do we even need the mode field accessible from the outside, apart from the need to have it visible for transmission? It certainly feels like no. Maybe there could be some note that {get_,set_}*id are the preferred mode of manipulation for these structures.

VladimirSlavik:
Update: Still WIP, so no need to re-review. Slowly getting there, TODOs left:
1. Rename the constants throughout the whole history, and
2. make tests work, then do so backwards in the commit flow.

VladimirSlavik:
This might actually work now, all the way through the history. WIP and TODOs still the same.

VladimirSlavik:
Update: Ready for review!

The history has been edited to resemble a realistic timeline with always working program and passing tests.

In-depth manual testing still TBD, but everything seems to work as it should.

poncovka (approved):
It looks great. Thank you!

pyanaconda/modules/common/structures/user.py
```diff
@@ -59,17 +60,32 @@ def name(self) -> Str:
     def name(self, name: Str):
         self._name = name
 
+    @property
+    def uid_mode(self) -> Str:
+        """Mode of UID.
+
+        Contains a string describing the mode of the user's UID: Use the value or default.
+
+        Possible values are: "ID_MODE_USE_VALUE" and "ID_MODE_USE_DEFAULT".
```
jkonecny12, line 69: Could you please write the items on lines below one on each line. It's more consistent with how the code is written.

pyanaconda/ui/gui/spokes/user.py
```diff
@@ -146,16 +146,14 @@ def apply(self):
...
-            self.user.uid_mode = constants.ID_MODE_USE_DEFAULT
+            self.user.set_uid(None)
 
         if self._cGid.get_active():
-            self.user.gid = int(self._gid.get_value())
+            self.user.set_gid(int(self._gid.get_value()))
         else:
-            self.user.gid = 0
-            self.user.gid_mode = constants.ID_MODE_USE_DEFAULT
+            self.user.set_gid(None)
```
jkonecny12, line 156: Just wonder if instead of setting `None` if we shouldn't have another setter. I mean something like `set_empty_gid()`.

jkonecny12 (approved):
Looks good to me. Just some nitpicks below.

pyanaconda/modules/common/structures/user.py
VladimirSlavik: Thanks, sounds good, will do.

pyanaconda/ui/gui/spokes/user.py
VladimirSlavik: Is that worth the work? There are two places where this could be used instead of calling with `None` explicitly. In 8 other places it just happens depending on what comes via a variable, so there is nothing really saved with such a change.

VladimirSlavik:
Now just wait a second! I fixed the comment according to nitpicks, and all acks are gone? That didn't happen yesterday!

pyanaconda/modules/users/users.py
```diff
@@ -100,6 +100,9 @@ def process_kickstart(self, data):
             group_data.name = group_ksdata.name
             if group_ksdata.gid is not None:
                 group_data.gid = group_ksdata.gid
+                group_data.gid_mode = ""
```
martinpitt: This looks unclean to me. You are using the _mode field like an enum, thus it should have a defined value (`ID_MODE_USE_{DEFAULT,VALUE}`), and `_VALUE` seems appropriate here as you just set one? Having a third nameless mode is confusing IMHO. Same thing several times below.

pyanaconda/core/constants.py
```diff
@@ -486,3 +486,7 @@ class PayloadRequirementType(Enum):
 
 # Path to the System Purpose configuration file on a system.
 RHSM_SYSPURPOSE_FILE_PATH = "/etc/rhsm/syspurpose/syspurpose.json"
+
+# GID and UID modes
+ID_MODE_USE_VALUE = "ID_MODE_USE_VALUE"
+ID_MODE_USE_DEFAULT = "ID_MODE_USE_DEFAULT"
```
martinpitt, line 492: To reduce the confusion and possible gotchas about this "fake" enum, how about using a boolean instead? Or even more pythonic, set `self._gid = None` initially if it's not set? Then you don't need an explicit _mode in the first place.

pyanaconda/ui/gui/spokes/user.py
martinpitt: FWIW, :+1 to setting it to `None` explicitly (as it does here)

martinpitt (dismissed):
Thanks! Found one bug (I think), and in general that whole back and forth with commits that change the logic three times is a bit unwieldy; could this be squashed? The normal Pythonic way of expressing "unset" is `None`, not a separate `_mode = UNSET` thing -- it's too easy to get out of sync, or to set `_mode` to an invalid value (as you do in three cases)

poncovka (dismissed):
Still looks good.

pyanaconda/modules/users/users.py
VladimirSlavik: Ha, good catch, that empty string should definitely not be there.

poncovka: This code is outdated.

pyanaconda/core/constants.py
poncovka: The `GroupData` and `UserData` classes are representations of DBus dictionaries, so we are limited by DBus data types. That is why we cannot use enums or set the properties to `None`.

poncovka:
> Thanks! Found one bug (I think), and in general that whole back and forth with commits that change the logic three times is a bit unwieldy; could this be squashed? The normal Pythonic way of expressing "unset" is `None`, not a separate `_mode = UNSET` thing -- it's too easy to get out of sync, or to set `_mode` to an invalid value (as you do in three cases)

Unfortunately, it is not a DBus way. We are not able to send `None` via DBus.

pyanaconda/core/constants.py
poncovka: Btw. we have considered the boolean, but the majority voted for the string.

martinpitt: Ack, thanks. Then let's fix the empty string modes.

jkonecny12 (approved):
Looks even better to me now!

VladimirSlavik:
@martinpitt all your comments are resolved, can you please amend the NACK?

martinpitt:
As I wrote above, I find it very unclean to first introduce a broken commit (`mode = ""`) and then immediately changing it with a subsequent commit. This breaks bisecting, and endangers back/forward-porting. (This ought to be squashed). So I really don't want to approve this, but at least I'll remove my NACK. Thanks!

VladimirSlavik:
Oh, I see now what you meant! I did not think you would review the middle commits specifically.

martinpitt:
@VladimirSlavik : Oh yes, I always review commits individually. They should be independent changes, and thus it's generally easier to review multiple smallerl changes than a big one which has multiple unrelated changes.

VladimirSlavik:
Update: Fixed what was discussed, rebased to get the latest test fixes, squashed.

martinpitt (approved):
Thanks! Will you send this to master as well?

VladimirSlavik:
Definitely, maybe even hold off merging until that is done. I want to run kickstart tests on this and that seems to have better success chance in master...

VladimirSlavik:
Master port merged.

PR #2907, opened 2020-10-07
Translate spoke category title directly at runtime

M4rtinK:
At the moment we mark spoke category title stored in class attribute for
translation and then translate it when hub sets up the category widget.

This had to be done as the class attribute is evaluated on code parsing
time and Anaconda might have a different locale (due to user selecting
installation language) by the tame the string is being displayed &
We would then translate the string at runtime.

This is unfortunately problematic for Initial Setup and addons as there
the category title will be marked for translation in the IS/addon
context and then the translation will be attempted in the Anaconda
context, which of course does not have the translated string.

So lets drop this split and instead translate the category title always
at runtime by providing it via a static class method. IS/addons can then
do the same, making sure the category title will be always translated at
the appropriate time & in their translation context.

Also clean up and document the category widget handling code a bit.

### DIFF
--- a/pyanaconda/ui/categories/__init__.py
+++ b/pyanaconda/ui/categories/__init__.py
@@ -17,12 +17,12 @@
 # Red Hat, Inc.
 #
 
-from pyanaconda.core.i18n import N_
+from abc import ABC, abstractmethod
 
 __all__ = ["SpokeCategory"]
 
 
-class SpokeCategory(object):
+class SpokeCategory(ABC):
     """A SpokeCategory is an object used to group multiple related Spokes
        together on a hub.  It consists of a title displayed above, and then
        a two-column grid of SpokeSelectors.  Each SpokeSelector is associated
@@ -31,11 +31,26 @@ class SpokeCategory(object):
 
        Class attributes:
 
-       sortOrder     -- A number indicating the order in which this Category
-                        will be displayed.  A lower number indicates display
-                        higher up in the Hub.
-       title         -- The title of this SpokeCategory, to be displayed above
-                        the grid.
     """
-    sortOrder = 1000
-    title = N_("DEFAULT TITLE")
+
+    @staticmethod
+    @abstractmethod
+    def get_title():
+        """The translated title of this category, to be displayed above the grid.
+
+        :return: translated category title
+        :rtype: str
+        """
+        return ""
+
+    @staticmethod
+    @abstractmethod
+    def get_sort_order():
+        """A number indicating the order in which this Category will be displayed.
+
+        A lower number indicates display higher up in the Hub.
+
+        :return: sort order number
+        :rtype: int
+        """
+        return 0
--- a/pyanaconda/ui/common.py
+++ b/pyanaconda/ui/common.py
@@ -665,7 +665,7 @@ def collectCategoriesAndSpokes(paths, klass):
     ret = {}
     # Collect all the categories this hub displays, then collect all the
     # spokes belonging to all those categories.
-    categories = sorted(collect_categories(paths["categories"]), key=lambda c: c.sortOrder)
+    categories = sorted(collect_categories(paths["categories"]), key=lambda c: c.get_sort_order())
 
     for c in categories:
         ret[c] = collect_spokes(paths["spokes"], c.__name__)
@@ -683,3 +683,22 @@ def collectCategoriesAndSpokes(paths, klass):
     lifecycle.add_controller(klass.__name__, category_names)
 
     return ret
+
+
+def sort_categories(categories):
+    """Sort categories based on sort order & calls name.
+
+    While sort order is the primarily sorting key, we can't always
+    assure all categories have unique sort order values.
+    So make the class name part of the sorting key, to at least
+    make the resulting category sort order deterministic.
+
+    To do this we concatenate we create a tuple as the sorting key,
+    with the sort order integer being followed by the by the class
+    name. This way the categories will be sorted by the sort order
+    first and then by class name if sort order happens to be the same.
+
+    :param categories: list of categories
+    :type categories: SpokeCategory sub classes
+    """
+    return sorted(categories, key=lambda i: (i.get_sort_order(), i.__class__))
--- a/tests/nosetests/pyanaconda_tests/common_ui_code_tests.py
+++ b/tests/nosetests/pyanaconda_tests/common_ui_code_tests.py
@@ -1,6 +1,5 @@
-# Localization category classes
 #
-# Copyright (C) 2011, 2013  Red Hat, Inc.
+# Copyright (C) 2020  Red Hat, Inc.
 #
 # This copyrighted material is made available to anyone wishing to use,
 # modify, copy, or redistribute it subject to the terms and conditions of
@@ -16,19 +15,57 @@
 # License and may only be used or replicated with the express permission of
 # Red Hat, Inc.
 #
+# Red Hat Author(s): Martin Kolman <mkolman@redhat.com>
+#
 
+import unittest
 from pyanaconda.core.i18n import _
+from pyanaconda.ui.common import sort_categories
 from pyanaconda.ui.categories import SpokeCategory
 
-__all__ = ["CustomizationCategory"]
 
-
-class CustomizationCategory(SpokeCategory):
+class AaaCategory(SpokeCategory):
 
     @staticmethod
     def get_title():
-        return _("CUSTOMIZATION")
+        return _("FOO")
 
     @staticmethod
     def get_sort_order():
         return 100
+
+
+class BbbCategory(SpokeCategory):
+
+    @staticmethod
+    def get_title():
+        return _("BAR")
+
+    @staticmethod
+    def get_sort_order():
+        return 100
+
+
+class CccCategory(SpokeCategory):
+
+    @staticmethod
+    def get_title():
+        return _("BAZ")
+
+    @staticmethod
+    def get_sort_order():
+        return 50
+
+
+class CommonCodeTestCase(unittest.TestCase):
+    """Test common UI code."""
+
+    def category_sorting_test(self):
+        """Test category sorting works as expected."""
+
+        category_list = [AaaCategory, BbbCategory, CccCategory]
+        # We expect the C category to be dorted first due to sort order and
+        # then A & B as they have the same sort order but A comes before B
+        # on the alphabet.
+        expected_category_list = [CccCategory, AaaCategory, BbbCategory]
+        self.assertEqual(sort_categories(category_list), expected_category_list)
--- a/pyanaconda/ui/categories/localization.py
+++ b/pyanaconda/ui/categories/localization.py
@@ -17,12 +17,18 @@
 # Red Hat, Inc.
 #
 
-from pyanaconda.core.i18n import N_
+from pyanaconda.core.i18n import _
 from pyanaconda.ui.categories import SpokeCategory
 
 __all__ = ["LocalizationCategory"]
 
 
 class LocalizationCategory(SpokeCategory):
-    sortOrder = 100
-    title = N_("LOCALIZATION")
+
+    @staticmethod
+    def get_title():
+        return _("LOCALIZATION")
+
+    @staticmethod
+    def get_sort_order():
+        return 100
--- a/pyanaconda/ui/categories/software.py
+++ b/pyanaconda/ui/categories/software.py
@@ -17,12 +17,18 @@
 # Red Hat, Inc.
 #
 
-from pyanaconda.core.i18n import N_
+from pyanaconda.core.i18n import _
 from pyanaconda.ui.categories import SpokeCategory
 
 __all__ = ["SoftwareCategory"]
 
 
 class SoftwareCategory(SpokeCategory):
-    sortOrder = 200
-    title = N_("SOFTWARE")
+
+    @staticmethod
+    def get_title():
+        return _("SOFTWARE")
+
+    @staticmethod
+    def get_sort_order():
+        return 200
--- a/pyanaconda/ui/categories/system.py
+++ b/pyanaconda/ui/categories/system.py
@@ -17,12 +17,18 @@
 # Red Hat, Inc.
 #
 
-from pyanaconda.core.i18n import N_
+from pyanaconda.core.i18n import _
 from pyanaconda.ui.categories import SpokeCategory
 
 __all__ = ["SystemCategory"]
 
 
 class SystemCategory(SpokeCategory):
-    sortOrder = 300
-    title = N_("SYSTEM")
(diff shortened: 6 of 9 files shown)


M4rtinK:
Dependent fix in Initial Setup: https://github.com/rhinstaller/initial-setup/pull/110

VladimirSlavik (approved):
Looks good to me. Thank you!

poncovka (commented):
I have suggested some small improvements. Otherwise, it looks great. Thanks!

poncovka:
Could you open a pull request for the hello world addon, please?

VladimirSlavik:
You will need to rebase for tests, I think...

pyanaconda/ui/categories/customization.py
```diff
@@ -17,12 +17,16 @@
...
 from pyanaconda.ui.categories import SpokeCategory
 
 __all__ = ["CustomizationCategory"]
 
 
 class CustomizationCategory(SpokeCategory):
-    sortOrder = 100
-    title = N_("CUSTOMIZATION")
+
+    def get_title():
```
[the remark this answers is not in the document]
M4rtinK: Yep, lets change this.

Hmm, I though I saw we were doing this somewhere else and did the same but 
can't find it now.

pyanaconda/ui/categories/__init__.py
```diff
@@ -31,11 +31,26 @@ class SpokeCategory(object):
...
+    @abstractmethod
+    def get_sort_order():
+        """A number indicating the order in which this Category will be displayed.
+
+        A lower number indicates display higher up in the Hub.
+
+        :return: sort order number
+        :rtype: int
+        """
+        pass
```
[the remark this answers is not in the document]
M4rtinK: Sure.

BTW, should it be always done ? I see some places we just have pass for an abstract method:

https://github.com/rhinstaller/anaconda/blob/master/pyanaconda/modules/payloads/payload/payload_base.py#L47

Or is it somehow related to these being properties ?

pyanaconda/ui/categories/__init__.py
```diff
@@ -31,11 +31,26 @@ class SpokeCategory(object):
...
+
+    @staticmethod
+    @abstractmethod
+    def get_title():
+        """The title of this SpokeCategory, to be displayed above the grid.
+
+        :return: category title
+        :rtype: str
+        """
+        pass
```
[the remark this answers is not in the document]
M4rtinK: Changed.

M4rtinK:
> Could you open a pull request for the hello world addon, please?

Will do. :)

pyanaconda/ui/categories/__init__.py
```diff
@@ -31,11 +31,26 @@ class SpokeCategory(object):
...
     """
-    sortOrder = 1000
-    title = N_("DEFAULT TITLE")
+
+    @staticmethod
+    @abstractmethod
+    def get_title():
+        """The title of this SpokeCategory, to be displayed above the grid.
+
+        :return: category title
```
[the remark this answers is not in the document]
M4rtinK: Good point, changed.

pyanaconda/ui/categories/__init__.py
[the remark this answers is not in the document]
poncovka: From the functionality point of view, it doesn't matter. However, I think that the code is more readable that way.

poncovka (approved):
Looks good to me. Thanks!

M4rtinK:
PR adjusting category handling  in Hello World Addon: https://github.com/rhinstaller/hello-world-anaconda-addon/pull/22

jkonecny12 (commented):
As I wrote elsewhere. This seems fine from the code PoV but are we sure about switching ``N_`` to ``_``? This won't work when compare is used. 

Also thinking that we may change the ``get_title()`` to ``get_localized_title()`` to avoid issues with using this in the future. The problem here is that it will work for English but for nothing else.

poncovka:
> As I wrote elsewhere. This seems fine from the code PoV but are we sure about switching `N_` to `_`? This won't work when compare is used.

The goal of this change is to use `_`, because addons are able to translate their own strings with the right context. Anaconda translates  `N_` strings with its own context and it results in untranslated strings.

I don't understand why you are worried about comparison of titles of categories. Categories are sorted by their sort order. This issue will be valid for spokes, but that's for another pull request.

> Also thinking that we may change the `get_title()` to `get_localized_title()` to avoid issues with using this in the future. The problem here is that it will work for English but for nothing else.

What kind of issues?

poncovka:
My bad. We sort categories by titles in TUI. However, I think we should use the sort order as well.

jkonecny12:
> > As I wrote elsewhere. This seems fine from the code PoV but are we sure about switching `N_` to `_`? This won't work when compare is used.
> 
> The goal of this change is to use `_`, because addons are able to translate their own strings with the right context. Anaconda translates `N_` strings with its own context and it results in untranslated strings.
> 
> I don't understand why you are worried about comparison of titles of categories. Categories are sorted by their sort order. This issue will be valid for spokes, but that's for another pull request.
> 
> > Also thinking that we may change the `get_title()` to `get_localized_title()` to avoid issues with using this in the future. The problem here is that it will work for English but for nothing else.
> 
> What kind of issues?

I mean that it's pretty easy to use this for something like:
```
if spoke.get_title() == "SourceSpoke":
```
without realizing that it won't work on non-english locale.

poncovka:
> I mean that it's pretty easy to use this for something like:
> 
> ```
> if spoke.get_title() == "SourceSpoke":
> ```
> 
> without realizing that it won't work on non-english locale.

The title will be `Installation Source` in this case. I think it is pretty easy to spot this kind of error.

VladimirSlavik:
It makes sense that translation should be mentioned in the method name. Otherwise I think this might be fine. Sorting by name which is untranslated-but-should-be sounds mostly not ok, so if this makes a fix for that necessary, I'd still consider that a good thing.

M4rtinK:
So to summarize the current issues & possible solutions (**for both spokes and categories to have the discussion in one place**):

* `get_title()` vs `get_localized_title()`
Personally, having the `get_localized_title()` does not make sense to me unless we also provide `get_title()` that returns the untranslated string. If we can avoid that, documenting get_title() returns translated strings should be IMHO enough.

* comparisons on category/spoke titles
These should not be a thing at all - as any other attempts to match on human readable strings in the GUI. If the human readable string is translated or not IMHO does not make a difference here.

For robust comparisons some machine readable spoke descriptor should be used - possibly class name or something new we can introduce similar to the `help_id` we already have.

* translations vs category & spoke ordering
I'm a bit confused at the moment if we are all taking about the same problem. 

If I understand things correctly, the current status quo is that we sort the spokes and (categories in TUI) based on their untranslated English name. This has the benefit of spokes (and categories in TUI) being always in the same place in the UI regardless of the selected locale. On the other hand, users that are not native in English might be confused why the spokes are not ordered according to their expected native alphabet ordering.

With change in spoke (and category localization) we might be be able to sort the spokes (and categories) in TUI based on the correct native alphabet ordering (as long as we make sure to only do it once locale has been set in the welcome spoke).  The downside here is that spokes might be located in different parts of the hub based on currently selected language. This can make debugging of l10n related issues as one can't simply put things side by side as before, but navigate based on spoke icon, which will be the same. I can imagine this also might momentarily break GUI testing tools that expect a certain spoke order and/or work based on screenshot bitmaps.

* inconsistency in category ordering
GUI orders categories by `sortOrder` but TUI seems to ignore it ? Do we want (or need ?) to keep it this way ? Or should we make the TUI `sortOder` aware to make this consistent ?

If there is a good reason we might want to sort the categories in a different way in the GUI and TUI I guess we could have a `sortOrderTUI` property if really needed.

* do we actually want to sort spokes alphabetically ?

As mentioned above it has some serious pitfalls & it might make sense to manually set the sort order to both make it consistent as well as possible order the spokes on the hub in a more ergonomic way. I remember @andreasn recommended an improved spoke & category ordering layout but I'm afraid it might have got lost in all the install-time-subscription related work & I can't find it right now.

VladimirSlavik:
Thanks for the summary, I agree with most of what's there.

The order should reflect something logical, eg. how people will move through these, not random, such as first letter ;-) At least that's what seems sane to me. Following from that, I completely agree that comparisons should not be based on strings.

jkonecny12:
> > I mean that it's pretty easy to use this for something like:
> > ```
> > if spoke.get_title() == "SourceSpoke":
> > ```
> > 
> > 
> > without realizing that it won't work on non-english locale.
> 
> The title will be `Installation Source` in this case. I think it is pretty easy to spot this kind of error.

OK bad example, the point is that for "CS" localization it will not be 'Installation Source' anymore.

jkonecny12:
> So to summarize the current issues & possible solutions (**for both spokes and categories to have the discussion in one place**):
> ### `get_title()` vs `get_localized_title()`
> 
> Personally, having the `get_localized_title()` does not make sense to me unless we also provide `get_title()` that returns the untranslated string. If we can avoid that, documenting get_title() returns translated strings should be IMHO enough.

I don't think there is a reason to have both. If we don't require to have ``get_title()`` then I don't see a reason to have it. On the other hand I still think that the name should explicitly tell that the string is already localized.

> ### comparisons on category/spoke titles
> 
> These should not be a thing at all - as any other attempts to match on human readable strings in the GUI. If the human readable string is translated or not IMHO does not make a difference here.
> 
> For robust comparisons some machine readable spoke descriptor should be used - possibly class name or something new we can introduce similar to the `help_id` we already have.

Good point on that but I still think the same about the above. Even though, I won't block this PR because of it.

M4rtinK:
After a meeting with all interested parties, we have reached the following conclusion in regards to the spoke related localization and attribute fixes/cleanups:

* `sort_order()` abstract property will be added to spokes, to provide deterministic ordering of spokes
* initial spoke sort order will be based on current alphabetical ordering (English) spoke titles, in the future the sort order can be adjusted as necessary
* category sorting in TUI will also be based on sort order instead of current alphabetical sort (GUI already displays categories based on sort order)
* `get_title()` will be used for category titles and `title()` for spokes, with the documentation clearly mentioning the output is expected to be localized (as is output from `status()` already)

andreasn:
@M4rtinK maybe it was this mockup?
Been a while since I last looked at this, but what I was attempting to solve at the time was that some spokes were dependant on others. So _Connect to Red Hat_ was dependant on the network being turned on (off by default in RHEL) in _Network & Host name_ and _Software selection_ dependant on _Installation source_ being set up etc.

image

Didn't look to deep into the problem though, so not sure how crucial it is.

M4rtinK:
@andreasn Thanks, that's the mockup I meant (and failed to find myself :P)! And it's definitely not a pressing concern at the moment - it's just that once this PR & #2960 are merged, we will actually have full controls over category and spoke placement in both GUI & TUI. Then we can actually do something about it - rest assured we will loop you in once that time comes. :)

M4rtinK:
@poncovka @jkonecny12 @VladimirSlavik So I did some final testing with TUI category sort based on sort order.

This is how it looked before:
TUI_pre_localization_fixes

This is how it looks now:
TUI_post_localization_fixes_category_sort

GUI also looks good:
GUI_post_localization_fixes_category_sort

So I think we can merge this (and the IS & Hello World Addon PRs), so that #2960 can be rebase on top of it & there is less change at one point in time.

pyanaconda/ui/common.py
```diff
@@ -683,3 +683,22 @@ def collectCategoriesAndSpokes(paths, klass):
...
+
+    To do this we concatenate we create a tuple as the sorting key,
+    with the sort order integer being followed by the by the class
+    name. This way the categories will be sorted by the sort order
+    first and then by class name if sort order happens to be the same.
+
+    :param categories: list of categories
+    :type categories: SpokeCategory sub classes
+    """
+    return sorted(categories, key=lambda i: (i.get_sort_order(), i.__class__))
```
poncovka, line 704: It should be sorted by `__class__.__name__`.

pyanaconda/ui/common.py
```diff
@@ -665,7 +665,7 @@ def collectCategoriesAndSpokes(paths, klass):
     ret = {}
     # Collect all the categories this hub displays, then collect all the
     # spokes belonging to all those categories.
-    categories = sorted(collect_categories(paths["categories"]), key=lambda c: c.sortOrder)
+    categories = sorted(collect_categories(paths["categories"]), key=lambda c: c.get_sort_order())
```
poncovka, line 668: This sort doesn't seem to be useful.

tests/nosetests/pyanaconda_tests/common_ui_code_tests.py
```diff
@@ -0,0 +1,71 @@
...
+        return 50
+
+
+class CommonCodeTestCase(unittest.TestCase):
+    """Test common UI code."""
+
+    def category_sorting_test(self):
+        """Test category sorting works as expected."""
+
+        category_list = [AaaCategory, BbbCategory, CccCategory]
```
poncovka, line 66: A and B is already in the input.

poncovka (commented):
I would prefer if this pull request wasn't merged without a final review.

pyanaconda/ui/common.py
poncovka: I couldn't figure out why your test is not failing, because classes are not comparable. The problem is that categories are not instances, but classes. Therefore `i` is a class and `i.__class__` is `type`. And `type` is comparable. Anyway, categories should be sorted by `__name__`. Please, make sure that your tests are testing this use case.

M4rtinK:
@poncovka Sorry for that, I've mentioned my intent to merge the PR on our daily standup & the changes in the send seemed quite minor. I guess pinging the reviewers on the PR would work better in the end though.

pyanaconda/ui/common.py
M4rtinK: I assumed there is some reason we sort the categories when they are being collected (stable listing in logs ?) so I just kept the logic the same. If you think it's safe to drop this, I can definitely do that. :)

pyanaconda/ui/common.py
M4rtinK: Oops! Yeah, that's really wrong. :P I'll fix it up in a folowup PR.

tests/nosetests/pyanaconda_tests/common_ui_code_tests.py
M4rtinK: Good point, I'll change the sort order or else, it will indeed just be kept the same in output, making the test kinda pohtless.

pyanaconda/ui/common.py
poncovka: I would try to remove drop that. It is just confusing.

PR #2908, opened 2020-10-08
Robustify GitHub actions runner download in ci-tasks container

martinpitt:
- Don't use the API, as that is very prone to running into rate
  limiting, especially behind large NATed environments such as the Red
  Hat VPN. Parse the releases HTML page instead, which is just as
  simple.

- Skip the intermediate writing and deletion of the tarball, just
  pipeline it.

- Enable verbosity for tar, for easier debugging.

### DIFF
--- a/dockerfile/ci-tasks/Dockerfile
+++ b/dockerfile/ci-tasks/Dockerfile
@@ -38,11 +38,9 @@ RUN pip install \
 
 # see https://github.com/martinpitt/anaconda/settings/actions/add-new-runner
 RUN mkdir actions-runner && cd actions-runner && \
-  URL=$(curl -s https://api.github.com/repos/actions/runner/releases/latest | \
-        sed -n '/browser_download_url.*linux-x64/ { s/^.*https:/https:/; s/"$//; p }') && \
-  curl -O -L "$URL" && \
-  tar xzf actions-runner-*.tar.gz && \
-  rm actions-runner-*.tar.gz
+  URL_BASE=https://github.com/actions/runner/releases && \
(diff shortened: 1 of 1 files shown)


martinpitt:
/packit build

Another [packit infra failure](https://console-testing-farm.apps.ci.centos.org/pipeline/befdd8cb-b0f3-4474-8ec9-598f12674bb6), let's try again

VladimirSlavik (approved):
Looks good to me. Thank you!

jkonecny12 (changes requested):
Looks a bit fragile to depend on the HTML page structure.

What about using some tool to parse json data? Maybe jq for example?

martinpitt:
@jkonecny12 : api.github.com was a JSON structure, but that's the one that I'd like to move away from. It has fierce rate limiting for unauthenticated callers (I was never able to run it even once from an upshift instasnce). So we'd need to get an authenticated token into the container build just for that, which I feel is way overkill and makes these container builds much less useful. I'd rather just hardcode the runner version and bump it from time to time.

Also, this doesn't really depend on the HTML "structure" as such, just that the page contains a link to a linux-x64 tarball. If they ever change the release artifact structure in a way that this does not exist any more (for example, renaming to linux-x86_64 or from .tar.xz to tar.gz or so), a JSON/jq based parser would fail as well.

I do realize that it's less than elegant, but still an okay compromise for the simplicity.

martinpitt:
@jkonecny12 : I changed this from "grep for the tarball" to "read the /release/latest redirection target" to get the current version number. Do you like this better?

jkonecny12 (approved):
Thanks for explanation. I think these arguments are meaningful and ACK this PR. 

This version seems a bit better than the previous one.

PR #2909, opened 2020-10-08
Simplify container make calls

jkonecny12:
It's not really required to run `./autogen.sh && ./configure` to run containers so let's skip that part and just support `make -f` call instead.


M4rtinK (approved):
Looks good to me - nice simplification! :)

martinpitt:
Taking the liberty to land this now, as I'd like to build on top of it.

PR #2910, opened 2020-10-08
Don't stop unit tests when the tests failed

jkonecny12:
Otherwise we would be missing logs which are uploaded as next step.


martinpitt (approved):
Indeed, in a [recent failure](https://github.com/martinpitt/anaconda/runs/1218578334?check_suite_focus=true) there was no log. I thought that this was due to the build failure (there are no logs to upload), but this makes sense. Dekuji!

VladimirSlavik (approved):
Looks good to me. Thank you!

M4rtinK (approved):
LGTM, thanks! :)

PR #2911, opened 2020-10-08
Add variable for extra labels to GitHub action runner entry point

martinpitt:
This will be useful to indicate for which branch/OS a runner can run
tasks. The workflow in a particular branch will then request running on
a particular label like "master" or "rhel-8".


martinpitt:
With the example it looks like this:

image

I also tested this with just one value and with not giving the new env variable at all, then it's just as before.

VladimirSlavik (approved):
Looks good to me. Thank you!

PR #2914, opened 2020-10-09
CI tasks cleanup

martinpitt:
With this we can re-use the same make rules for the rhel-8 and other branches, and they become extensible for other containers.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -79,17 +79,15 @@ SRPM_BUILD_DIR = $(BUILD_RESULT_DIR)/00-srpm-build
 RPM_BUILD_DIR = $(BUILD_RESULT_DIR)/01-rpm-build
 TEST_INST_BUILD_DIR = $(BUILD_RESULT_DIR)/02-test-install
 
-# Default variables for container build call
-DOCKERFILE_PATH ?= $(srcdir)/dockerfile/ci-tasks
-CONTAINER_BUILD_NAME ?= anaconda/ci-tasks
-CONTAINER_BUILD_TAG ?= latest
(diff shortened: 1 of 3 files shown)


VladimirSlavik (approved):
I don't really understand the first commit, but the second definitely LGTM.

Makefile.am
```diff
@@ -80,14 +80,15 @@ RPM_BUILD_DIR = $(BUILD_RESULT_DIR)/01-rpm-build
...
+CONTAINER_ENGINE ?= podman
 CONTAINER_BUILD_ARGS ?=
 CONTAINER_TEST_ARGS ?=
 
 # ci-tasks container
 CI_TASKS_DOCKERFILE ?= $(srcdir)/dockerfile/ci-tasks
 CI_TASKS_NAME ?= anaconda/ci-tasks
 CI_TASKS_TAG ?= $(GIT_BRANCH)
-CI_TASKS_TEST_ARGS ?= --rm -v .:/anaconda:Z
+CI_TASKS_TEST_ARGS ?= --rm -v $(CURDIR):/anaconda:Z
```
jkonecny12, line 91: I did not know about this variable. Pretty nice!

Don't we want to use this instead of ``$(srcdir)`` above too?

jkonecny12 (approved):
Looks good to me otherwise.

Makefile.am
martinpitt: Good point, thanks! I cleaned this up by having srcdir default to `$(CURDIR)` instead of `.` and using `$(srcdir)` here.

jkonecny12 (approved):
Looks great now. Thanks!

PR #2915, opened 2020-10-09
Add ci-tasks container for running unit tests

martinpitt:
Port dockerfile/ci-tasks from the master branch, with some RHEL specific
changes:

 - Derive from the UBI8 image instead of Fedora Rawhide. (UBI is public)

 - Disable the default repositories (which require a subscription), and
   replace them with the Red Hat internal devel ones, plus the Anaconda
   overlay repositories. This requires VPN, thus the image can be built
   only internally.

 - Drop the COPRs, they are not relevant for testing the RHEL branch.

 - Get the BuildRequires from the rhel-8 branch.

 - Pin pylint to version 2.5.3 (just like the internal Jenkins job
   does), as the current version has a lot of complaints about exception
   handling (see master commit d07e90b93145).

 - Disable full upgrade, this is brittle in unprivileged podman.

Copy the convenience Make rules from master as well, they work
unchanged.

-----

This backports some stuff which has not landed in master yet, so this is blocked:

 - [ ] PR #2908
 - [ ] PR #2911
 - [ ] PR #2914


martinpitt:
This mirrors the Jenkins+mock config that we currently use (I had some fun finding all these bits :grin:).

Test this with:

    make -f Makefile.am ci-tasks-build
    make -f Makefile.am container-ci

I did not yet add a workflow as we still need to set up some self-hosted runners. I'm almost there, but I need this PR to land first to test it properly.

martinpitt:
All depending PRs landed now, so this is unblocked and ready for review.

martinpitt:
There is a new internal github-runner upshift instance (fully automated deployment) which registers [two runners](https://github.com/rhinstaller/anaconda/settings/actions). As these aren't able to pick up actions from my fork, I started another container for martinpitt/anaconda instead of rhinstaller/anaconda:
```
podman run --name ci-tasks-m1 -h github-runner-1 --rm -e GITHUB_TOKEN=$(cat ~/github-token) -e GITHUB_REPOSITORY=martinpitt/anaconda -e RUNNER_LABELS=rhel-8 --entrypoint /github-action-run-once anaconda/ci-tasks:rhel-8
```

and have a branch (not PRed yet, as it is still blocked) that [ran the unit tests](https://github.com/martinpitt/anaconda/runs/1241655819?check_suite_focus=true) on that upshift instance. The pylint test hung (there was a zombie), I had to kill it. I still need to investigate that. But a local test run is fine, and that's all what this PR adds.

dockerfile/ci-tasks/overlays.repo
```diff
@@ -0,0 +1,11 @@
+[rhinstaller]
+name=Copr repo for rhinstaller daily builds for RHEL 8
+baseurl=http://coprbe.devel.redhat.com/results/rhinstaller-group/Anaconda/rhel-8.dev-$basearch/
+enabled=1
+
+[cobra2-rhel-fallback]
+name=Cobra02 custom repository used to solve temporary missing new versions
+baseurl=http://cobra02.anaconda.englab.brq.redhat.com/trees/tests_fallback_repos/rhel8/
```
jkonecny12, line 8: Is it fine to have these links here?

jkonecny12 (changes requested):
Looks good to me otherwise but is it fine to have internal links in public repo? I know they are not accessible but still.

dockerfile/ci-tasks/overlays.repo
martinpitt: If you google for them, there are several hits in bugzilla and fedorapeople that already expose them. There's a lot more hits for the download.devel.redhat.com repos. We also expose them in Cockpit bots, etc. Their names are not secret really, they don't expose anything about customers, projects, etc. That's why I wouldn't worry too much about that -- the benefits for making it really easy to build/use these containers outweigh that IMHO.

jkonecny12 (approved):
OK thanks for explanations!

martinpitt:
Yay, a run with `jobs=2` [works](https://github.com/martinpitt/anaconda/runs/1242813525?check_suite_focus=true). So I'll work on a PR to specify/reduce the number of pylint jobs first, and cherry-pick that into this PR.

VladimirSlavik (approved):
Looks good to me. Thank you!

martinpitt:
I sent PR #2918 to set a more reasonable `--jobs` for pylint. Once that landed, I'll cherry-pick it here.

martinpitt:
PR #2918 landed, I cherry-picked the fix here. No other changes. @jkonecny12 , do you want to sign off again?

jkonecny12 (approved):
LGTM!

PR #2916, opened 2020-10-09
Fix crash on nonexisting network config directories

martinpitt:
In the network config module, don't crash on nonexisting
/etc/sysconfig/network-scripts/ directory. This is owned by
NetworkManager, which is not present in a minimal installation. Just
return an empty file list in this case.

Detected by the "container" kickstart test.

### DIFF
--- a/pyanaconda/modules/network/config_file.py
+++ b/pyanaconda/modules/network/config_file.py
@@ -44,6 +44,10 @@ def get_config_files_paths(root_path=""):
     paths = []
     for directory in (os.path.normpath(root_path + IFCFG_DIR),
                       os.path.normpath(root_path + KEYFILE_DIR)):
+        if not os.path.exists(directory):
+            log.info("network device config directory %s does not exist, skipping", directory)
+            continue
+
         for filename in os.listdir(directory):
             if not (_has_keyfile_basename(filename) or _has_ifcfg_basename(filename)):
                 continue


martinpitt:
Once the initial "missing rpm" crash in the "container" kickstart test gets fixed, it failed 
```
11:50:08,085 WARNING org.fedoraproject.Anaconda.Modules.Network:  File "/usr/lib64/python3.9/site-packages/pyanaconda/core/async_utils.py", line 68, in _idle_method
11:50:08,097 WARNING org.fedoraproject.Anaconda.Modules.Network:    func(*args, **kwargs)
11:50:08,097 WARNING org.fedoraproject.Anaconda.Modules.Network:  File "/usr/lib64/python3.9/site-packages/pyanaconda/modules/common/task/runnable.py", line 110, in _task_succeeded_callback
11:50:08,098 WARNING org.fedoraproject.Anaconda.Modules.Network:    self._succeeded_signal.emit()
11:50:08,098 WARNING org.fedoraproject.Anaconda.Modules.Network:  File "/usr/lib64/python3.9/site-packages/pyanaconda/core/signal.py", line 36, in emit
11:50:08,101 WARNING org.fedoraproject.Anaconda.Modules.Network:    func(*args, **kargs)
11:50:08,102 WARNING org.fedoraproject.Anaconda.Modules.Network:  File "/usr/lib64/python3.9/site-packages/pyanaconda/modules/network/network.py", line 354, in <lambda>
11:50:08,102 WARNING org.fedoraproject.Anaconda.Modules.Network:    lambda: self.log_task_result(task, root_path=conf.target.system_root)
11:50:08,102 WARNING org.fedoraproject.Anaconda.Modules.Network:  File "/usr/lib64/python3.9/site-packages/pyanaconda/modules/network/network.py", line 702, in log_task_result
11:50:08,103 WARNING org.fedoraproject.Anaconda.Modules.Network:    self.log_configuration_state(task.name, root_path)
11:50:08,103 WARNING org.fedoraproject.Anaconda.Modules.Network:  File "/usr/lib64/python3.9/site-packages/pyanaconda/modules/network/network.py", line 715, in log_configuration_state
11:50:08,103 WARNING org.fedoraproject.Anaconda.Modules.Network:    for line in get_config_files_content(root_path=root_path).splitlines():
11:50:08,103 WARNING org.fedoraproject.Anaconda.Modules.Network:  File "/usr/lib64/python3.9/site-packages/pyanaconda/modules/network/config_file.py", line 35, in get_config_files_content
11:50:08,103 WARNING org.fedoraproject.Anaconda.Modules.Network:    for file_path in get_config_files_paths(root_path):
11:50:08,103 WARNING org.fedoraproject.Anaconda.Modules.Network:  File "/usr/lib64/python3.9/site-packages/pyanaconda/modules/network/config_file.py", line 47, in get_config_files_paths
11:50:08,104 WARNING org.fedoraproject.Anaconda.Modules.Network:    for filename in os.listdir(directory):
11:50:08,104 WARNING org.fedoraproject.Anaconda.Modules.Network:FileNotFoundError: [Errno 2] No such file or directory: '/mnt/sysroot/etc/sysconfig/network-scripts'
```

With both fixes, the test succeeds again.

VladimirSlavik (approved):
Looks good to me. Thank you!

pyanaconda/modules/network/config_file.py
```diff
@@ -44,10 +44,14 @@ def get_config_files_paths(root_path=""):
...
-            if not (_has_keyfile_basename(filename) or _has_ifcfg_basename(filename)):
-                continue
-            paths.append(os.path.join(directory, filename))
+        try:
+            for filename in os.listdir(directory):
+                if not (_has_keyfile_basename(filename) or _has_ifcfg_basename(filename)):
+                    continue
+                paths.append(os.path.join(directory, filename))
+        except FileNotFoundError:
+            # ignore missing directories
```
jkonecny12, line 53: Could you please add debug info. It could help with the debugging in the future.

jkonecny12 (changes requested):
Looks great. Thanks for this fix.

Just please we should log that this code does not happened.

pyanaconda/modules/network/config_file.py
rvykydal: Maybe just checking directory existence instead of catching an exception would be more self-explaining here?

martinpitt: "easier to ask for forgiveness than permission" is more pythonic than "look before you leap" -- and the latter is a bad style to get used to, as it is TOCTOU race condition. Sure, in this case it isn't relevant (most probably), but easier/more robust to just handle the exception.

rvykydal: I agree with the general rule, in this case I was considering the asking because
- the race condition would not apply
- the readability
- the block that is caught seemed too big to me (compared to the os.listdir call)
But I am fine with catching the exception either.

rvykydal (approved):
Looks good to me.

pyanaconda/modules/network/config_file.py
poncovka: I would agree, but in this case, it is not very clear which file might be missing since we are working with directories and files in these directories. Also, the robustness is not always helpful. Anaconda sometimes runs in very weird and unstable environments (for example, with failing hardware) and it doesn't make sense to pretend that everything is fine when it clearly isn't.

martinpitt: OK, changed to LBYL now.

poncovka (approved):
It looks good to me. Thank you!

jkonecny12 (approved):
LGTM!

PR #2917, opened 2020-10-12
fix “0 storage devices selected”abnormality

xqrustc2020:
"0 storage devices selected" is printed because the format_type of sda is "ext4" rather than "disklabel", and disk 'sda' is filtered by filter_disks_by_names(partitioned_devices, selected_disks) .
Resolves: rhbz#1878661

### DIFF
--- a/pyanaconda/ui/gui/spokes/custom_storage.py
+++ b/pyanaconda/ui/gui/spokes/custom_storage.py
@@ -47,7 +47,7 @@ from pyanaconda.modules.common.structures.partitioning import PartitioningReques
 from pyanaconda.modules.common.structures.device_factory import DeviceFactoryRequest, \
     DeviceFactoryPermissions
 from pyanaconda.product import productName, productVersion
-from pyanaconda.ui.lib.storage import reset_bootloader, create_partitioning, filter_disks_by_names
+from pyanaconda.ui.lib.storage import reset_bootloader, create_partitioning
 from pyanaconda.core.storage import DEVICE_TYPE_UNSUPPORTED, DEVICE_TEXT_MAP, \
     MOUNTPOINT_DESCRIPTIONS, NAMED_DEVICE_TYPES, CONTAINER_DEVICE_TYPES, device_type_from_autopart, \
     PROTECTED_FORMAT_TYPES, DEVICE_TYPE_BTRFS, DEVICE_TYPE_MD, Size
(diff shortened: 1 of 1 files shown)


pyanaconda/ui/gui/spokes/custom_storage.py
```diff
@@ -302,9 +302,7 @@ def refresh(self):
         self._default_file_system = self._device_tree.GetDefaultFileSystem()
 
         # Initialize the selected disks.
-        selected_disks = self._disk_selection.SelectedDisks
-        partitioned_devices = self._device_tree.GetPartitioned()
-        self._selected_disks = filter_disks_by_names(partitioned_devices, selected_disks)
```
poncovka, line 307: I am a little worried about the side effects of this change. Did it work before?

poncovka:
@vojtechtrefny, could you look at the bug [1878661](https://bugzilla.redhat.com/show_bug.cgi?id=1878661), please? Do we support this configuration?

pyanaconda/ui/gui/spokes/custom_storage.py
xqrustc2020: In this situation, self._device_tree.GetPartitioned()  is used to check whether the disk format type is 'disklabel', which means that I can not use a formatted disk without partition table to install OS. I think it is unjustified.  In other cases, it truly works.

vojtechtrefny:
You should definitely support using raw format disks in the custom spoke. But looking at #2872, this should be already solved and "removed" disk with an ext4 format should be properly initialized. If this change is only about showing the disk in the UI, it shouldn't be dangerous. The comment `Initialize the selected disks` is a little confusing here -- if this is only about what disks should be shown in the UI, I wouldn't use the word "initialize" because it's used when creating a new partition table on the disk.
But in general both changes look good to me.

poncovka (approved):
Looks good to me. Thank you!

poncovka:
> The comment `Initialize the selected disks` is a little confusing here -- if this is only about what disks should be shown in the UI, I wouldn't use the word "initialize" because it's used when creating a new partition table on the disk.

The comment is really confusing. It is about the initialization of the UI, the disks are not initialized by these calls.

xqrustc2020:
@vojtechtrefny Thanks for your reply. This change is not only about what disks should be shown in the UI, but also determines whether disk can be performed custom partition after ext4 format is removed. If 0 disk is selected,  I can not choose the disk to add new partitions one by one.

vojtechtrefny (approved):
LGTM

VladimirSlavik:
Thank you @poncovka and @vojtechtrefny, looks like you actually solved this. Is there anything left for me to do?

xqrustc2020:
> Thank you @poncovka and @vojtechtrefny, looks like you actually solved this. Is there anything left for me to do?
> 
> Edit: Besides testing and later merging, of course.

Thank you @VladimirSlavik , I have no more quetions.

xqrustc2020:
> Thank you @poncovka and @vojtechtrefny, looks like you actually solved this. Is there anything left for me to do?
> 
> Edit: Besides testing and later merging, of course.

Thank you @VladimirSlavik , I have no more questions.

VladimirSlavik:
I apologize for lack of activity on this. I wanted to manually test this change and then run the full range of storage tests with it, but I am not able to: Recent issues with metacity prevent the installer from starting so I cannot test this (yet). Hopefully the issue will be resolved soon, or I will try to find a way to work around this.

VladimirSlavik:
Ran some of the storage kickstart tests, results looked ok-ish. Manual testing looked good too. Let's merge!

PR #2918, opened 2020-10-13
Restrict pylint parallelism to available RAM

martinpitt:
pylint uses *lots* of RAM (more than 1 GiB per process), and the
`jobs=0` default runs `nproc` parallel processes by default. This
exceeds the available memory in many standard environments (such as
m1.medium == 2 CPUs/4 GiB RAM, or m1.large == 4 CPUs/8 GiB RAM) and
causes eternal hangs of the pylintrc test when one job process gets
OOM-killed.

Instead, set the parallelism depending on the amount of free memory --
one job for each 2 GB of available RAM. This is an underestimation, but
let's not use up the entire system memory. Also still cap it to the
number of available CPUs, just in case this runs on a system with lots
of RAM and only few CPUs.

### DIFF
--- a/tests/pylint/censorship.py
+++ b/tests/pylint/censorship.py
@@ -32,6 +32,7 @@ __all__ = ["CensorshipLinter", "CensorshipConfig"]
 
 import sys
 import re
+import multiprocessing
 
 from io import StringIO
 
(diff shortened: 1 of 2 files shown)


tests/pylint/censorship.py
```diff
@@ -116,6 +117,23 @@ def run(self):
     def _prepare_args(self):
         args = []
 
+        # pylint uses a lot of memory and doesn't handle ENOMEM well, so set --jobs based on memory
```
poncovka, line 120: Do you think it might be worth it to report the issue to pylint?

poncovka (approved):
Looks good to me. Thanks!

tests/pylint/censorship.py
martinpitt: https://github.com/PyCQA/pylint/issues/1495 has existed for years, but I'll report a more specific one.

VladimirSlavik (approved):
Looks good to me. Thank you!

jkonecny12 (approved):
Looks good to me. Thanks for this improvement!

martinpitt:
I filed https://github.com/PyCQA/pylint/issues/3899 about the hang after OOM, and followed up to https://github.com/PyCQA/pylint/issues/1495 with a current reproducer about the memory leak.

PR #2919, opened 2020-10-13
Allow hiding the help button selectively

VladimirSlavik:
Resolves: [rhbz#1890092](https://bugzilla.redhat.com/show_bug.cgi?id=1890092)

Before & after - see "help" button absence in top right corner:
Screenshot_vs_main_2020-10-21_15:56:07
Screenshot_vs_main_2020-10-21_15:51:36

~~Blocked since we don't have the bug (yet).~~

~~There does not seem to be a good place to put this, since spokes generally inherit from `NormalSpoke` and `StandaloneSpoke` which then are based on the GUI/TUI-agnostic classes of the same name before a common ancestor comes up. I tried a mixin, but it seemed pretty weird to have it depend on unknown attributes, and making it inherit from `GUIObject` meant introducing another diamond. Additionally it would mean the calls to `super()...` would need to be specified. So in the end I just made it a helper function. This needs the smallest amount of changes, both in structural sense and LoC.~~

### DIFF
--- a/pyanaconda/ui/gui/__init__.py
+++ b/pyanaconda/ui/gui/__init__.py
@@ -106,6 +106,8 @@ class GUIObject(common.UIObject):
                            given GUI object. The default value of None indicates
                            that the object has not specific help file assigned
                            and the default help placeholder should be used.
+
+       hide_help_button -- Hide the button for showing help.
     """
     builderObjects = []
     mainWidgetName = None
@@ -117,6 +119,7 @@ class GUIObject(common.UIObject):
 
     uiFile = ""
     help_id = None
+    hide_help_button = False
     translationDomain = "anaconda"
 
     handles_autostep = False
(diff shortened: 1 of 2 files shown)


pyanaconda/ui/gui/utils.py
```diff
@@ -550,3 +550,14 @@ def set_password_visibility(entry, visible):
...
     entry.set_icon_from_icon_name(position, icon)
     entry.set_icon_tooltip_text(position, text)
+
+
+def hide_help_button_if_no_help(spoke):
+    """Hide help button for spokes with no help target set.
+
+    :param spoke: The spoke or hub to modify.
+    """
+    if not spoke.help_id:
```
jkonecny12, line 560: Shouldn't this be ``if spoke.help_id is None:``

jkonecny12 (dismissed):
Looks good to me! Just a note below.

VladimirSlavik:
Looks like there are multiple ideas what this should be. Please @ShwetaNaresh or @mkolman file the bug and get it acked so it's clear.

If a generic approach is to be taken, then `GUIObject` is the best place to put this, overriding `initialize` there, and checking a class attribute.

M4rtinK:
@VladimirSlavik Actually Anacondas Martin Kolman == @M4rtinK (yes, it's a bit confusing ;-) )

VladimirSlavik:
Completely changed according to previous notes.

poncovka (approved):
Looks good to me. Thanks!

VladimirSlavik:
Actually blocked by bug lacking acks, so let's get the label back.

jkonecny12 (approved):
LGTM!

M4rtinK (approved):

VladimirSlavik:
Now waiting for [the bug](https://bugzilla.redhat.com/show_bug.cgi?id=1890092) to get qe ack.

PR #2920, opened 2020-10-13
Run unit tests on self-hosted GitHub runner

martinpitt:
Adapted the workflow from master, but skip the `container:` bits -- our
registered runners are already one-time throwaway container
environments for security reasons.

Related: rhbz#1885635

 - [x] Wait for and include PR #2925

### DIFF
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -1,0 +1,33 @@
+name: validate
+on: [push, pull_request_target]
+jobs:
+  unit-tests:
+    # this already is a ci-tasks:rhel-8 container environment
+    runs-on: [self-hosted, ci-tasks, rhel-8]
+    steps:
+      - name: Clone repository
+        uses: actions/checkout@v2
+
+      - name: Run tests
+        id: run-unit-tests
+        run: |
+          ./autogen.sh
+          ./configure
+          make
+          make ci
+        continue-on-error: true
+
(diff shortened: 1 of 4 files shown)


martinpitt:
The infrastructure part is in the internal rhinstaller/builders repo PR 21. This can't actually trigger the "real" runner yet, as the workflow needs to be *on* the target branch (rhel-8) before it will be displayed here. But I ran a container instance on the github-runner VM with my token, and [it succeeded](https://github.com/martinpitt/anaconda/runs/1249382612).

martinpitt:
A second run still ran into the pylint OOM failure. So I sent PR #2923 to fix this for master, and cherry-picked the fix here as well. I ran it again, and [it succeeded](https://github.com/martinpitt/anaconda/runs/1251826754?check_suite_focus=true).

It also makes sense now:

    Using 2 parallel jobs based on 7255904 kB available RAM and 4 CPUs

(it used 4 before, and it crashed even with 3)

martinpitt:
Blocking again. I modified #2923 , and there's #2925 to consider. Once these have landed, I'll re-do the backport.

martinpitt:
I backported #2925 as well. [successful test run](https://github.com/martinpitt/anaconda/runs/1253348685?check_suite_focus=true)

jkonecny12 (approved):
LGTM!

PR #2921, opened 2020-10-13
Move constants to prevent circular dependency

VladimirSlavik:
~~To prevent a circular dependency it is imported only in the function.~~

Resolves: [rhbz#1883359](https://bugzilla.redhat.com/show_bug.cgi?id=1883359)

### DIFF
--- a/pyanaconda/ui/gui/spokes/lib/storage_dialogs.py
+++ b/pyanaconda/ui/gui/spokes/lib/storage_dialogs.py
@@ -28,7 +28,19 @@ from pyanaconda.ui.gui.utils import escape_markup
 
 log = get_module_logger(__name__)
 
-__all__ = ["NeedSpaceDialog", "NoSpaceDialog"]
+__all__ = ["NeedSpaceDialog", "NoSpaceDialog", "RESPONSE_CANCEL", "RESPONSE_OK",
+           "RESPONSE_MODIFY_SW", "RESPONSE_RECLAIM", "RESPONSE_QUIT", "DASD_FORMAT_NO_CHANGE",
+           "DASD_FORMAT_REFRESH", "DASD_FORMAT_RETURN_TO_HUB"]
+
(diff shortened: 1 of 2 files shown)


jkonecny12 (approved):
Looks good to me.

poncovka (approved):
Looks good to me. Thanks!

pyanaconda/ui/gui/spokes/lib/storage_dialogs.py
```diff
@@ -49,7 +49,8 @@ def run(self):
 
     def _modify_sw_link_clicked(self, label, uri):
         if self._software_is_ready():
-            self.window.response(RESPONSE_MODIFY_SW)  # pylint: disable=undefined-variable
+            from pyanaconda.ui.gui.spokes.storage import RESPONSE_MODIFY_SW
```
poncovka, line 52: Btw. I would consider to move all these constants to `storage_dialogs.py` to prevent these issues.

VladimirSlavik: Let's see about that, then...

VladimirSlavik:
Update: Moved the constants instead, as suggested.

poncovka (approved):
Looks good to me. Thanks!

jkonecny12 (approved):
Looks good to me. I just wonder if it wouldn't be simpler to have these as `Enum`. It would make importing much easier. However, I did not looked on the code so I'm not sure how many places needs to change.

VladimirSlavik:
These constants are used in dialog return values, just in these two files. Typically you'll get from the dialog a raw `4` and compare to `RESPONSE_QUIT`. I think having to do `if rc = ReturnValueType.RESPONSE_QUIT.value` is not really better.

VladimirSlavik:
Tests pass locally. Merging.

PR #2922, opened 2020-10-13
Verify RPM exists before running it

VladimirSlavik:
SSIA. Some error messaging added as well.

Thanks to @martinpitt and @poncovka finding this in https://github.com/rhinstaller/kickstart-tests/pull/366.

### DIFF
--- a/pyanaconda/payload/dnf/payload.py
+++ b/pyanaconda/payload/dnf/payload.py
@@ -1992,10 +1992,20 @@ class DNFPayload(Payload):
             # TODO: replace the interpolation with DNF once possible
             arch = util.execWithCapture("uname", ["-i"]).strip().replace("'", "")
             vers = util.get_os_version(conf.target.system_root)
+
+            if not os.path.exists(conf.target.system_root + "/usr/bin/rpm"):
+                log.error("Can not import GPG keys to RPM database because the 'rpm' executable "
+                          "is missing on the target system. The following keys were not "
+                          "imported:\n%s",
+                          "\n".join(conf.payload.default_rpm_gpg_keys))
+                return
+
             for key in conf.payload.default_rpm_gpg_keys:
                 interpolated_key = key.replace("$releasever", vers).replace("$basearch", arch)
                 log.info("Importing GPG key to RPM database: %s", interpolated_key)
-                util.execInSysroot("rpm", ["--import", interpolated_key])
+                rc = util.execInSysroot("rpm", ["--import", interpolated_key])
+                if rc:
+                    log.error("Failed to import key.")
 
     def post_setup(self):
         """Perform post-setup tasks.


martinpitt:
Unit tests fail, but that's almost certainly fallout from some change in rawhide, not this PR:

```
======================================================================
FAIL: Check if the empty_ok flag works correctly.
----------------------------------------------------------------------
  File "/anaconda/tests/nosetests/pyanaconda_tests/password_quality_test.py", line 98, in password_empty_ok_test
    self.assertEqual(check.result.password_score, 1)
AssertionError: 2 != 1

======================================================================
FAIL: Check if minimal password length is checked properly.
----------------------------------------------------------------------
  File "/anaconda/tests/nosetests/pyanaconda_tests/password_quality_test.py", line 121, in password_length_test
    self.assertEqual(check.result.password_score, 1)
AssertionError: 2 != 1

======================================================================
FAIL: Check if libpwquality gives reasonable numbers & score is assigned correctly.
----------------------------------------------------------------------
  File "/anaconda/tests/nosetests/pyanaconda_tests/password_quality_test.py", line 164, in password_quality_test
    self.assertEqual(check.result.password_quality, 0)
AssertionError: 34 != 0
```

martinpitt (approved):
Nice, thank you! So we should revert https://github.com/rhinstaller/kickstart-tests/pull/366 after this. Did you run the container test with this update and without rpm?

VladimirSlavik:
> Did you run the container test with this update and without rpm?

I didn't actually - let's try.

poncovka (approved):
Looks good to me. Thanks!

pyanaconda/payload/dnf/payload.py
```diff
@@ -1992,10 +1992,19 @@ def _import_rpm_keys(self):
             # TODO: replace the interpolation with DNF once possible
             arch = util.execWithCapture("uname", ["-i"]).strip().replace("'", "")
             vers = util.get_os_version(conf.target.system_root)
+
+            if not os.path.exists(conf.target.system_root + "/usr/bin/rpm"):
+                log.error("Can not import GPG keys to RPM database because rpm is missing. The "
```
jkonecny12, line 1997:
```suggestion
                log.error("Can not import GPG keys to RPM database because rpm tool is missing. The "
```
It's a bit unclear otherwise.

jkonecny12 (approved):
Please change the message. Otherwise, it's good to me.

pyanaconda/payload/dnf/payload.py
martinpitt: Bikeshedding of course, but an [RPM tool](https://www.rpmtool.com/) is something else :grin: (I've never seen rpm to be referred to as a "tool".)

jkonecny12: hmm... good point. When someone tell me RPM, I always first think about some package. I guess that's the problem on my side?

martinpitt: Oh! Now I understand how you read it before :) I. e. "some random rpm is missing", not "rpm the package manager" is missing -- i.e. `rpm.rpm`. Interesting semantics. So, just ignore me then, it's fine for me either way.

VladimirSlavik: Thanks, good catch! I think you want "executable".

VladimirSlavik:
The "container" test works well with this. Will make a PR for that shortly.

VladimirSlavik:
Update: Fixed the message per discussion above.

martinpitt (approved):
Thanks! That's very clear now. It's the rpm rpm :-)

VladimirSlavik:
And now the rpm tests fail! It's the revenge of rpm!

martinpitt:
@VladimirSlavik : They fail everywhere, we need to urgently land #2926

jkonecny12 (approved):
Looks great to me now!

M4rtinK (approved):
Looks good to me. :)

VladimirSlavik:
Rebased for test fixes...

VladimirSlavik:
Tests pass locally, merging...

PR #2923, opened 2020-10-14
unit test improvement/fixes

martinpitt:


martinpitt:
As the unit tests currently fail due to a change in rawhide(first seen in [PR #2922](https://github.com/rhinstaller/anaconda/pull/2922/checks?check_run_id=1249458384)), this provides a good opportunity to validate that the log cat works as intended. :grin:

martinpitt:
rpm-tests failure is https://bugzilla.redhat.com/show_bug.cgi?id=1887969 FTR

martinpitt:
Yes, so we can now see the failure directly in the output, and have a convenient way to link to it: https://github.com/rhinstaller/anaconda/pull/2923/checks?check_run_id=1251662037#step:6:1485

poncovka (approved):

VladimirSlavik (approved):
Thank you!

martinpitt:
Ported to rhel-8 branch in PR #2920

.github/workflows/validate.yml
```diff
@@ -22,6 +22,7 @@ jobs:
           name: logs
           path: |
             tests/test-suite.log
+            tests/pylint/runpylint*.log
```
jkonecny12, line 25: Please add also ``tests/nosetests.log``.

tests/README.rst
```diff
@@ -79,7 +79,7 @@ Logs from the run are stored in the ``tests`` folder.
 
 For debugging of the container please run the container as::
 
-    make container-ci CONTAINER_TEST_ARGS="--rm -ti -v .:/anaconda:Z --entrypoint /bin/bash"
+    make container-ci CONTAINER_TEST_ARGS="-it --entrypoint /bin/bash"
```
jkonecny12, line 82: Please look here https://github.com/rhinstaller/anaconda/pull/2925

We have to decide the correct solution here.

jkonecny12 (changes requested):
We have to find the correct solution first.

Other than that looks good to me.

.github/workflows/validate.yml
martinpitt: nosetests output is already contained in test-suite.log, so that seemed redundant to me. But it's also tiny, so I updated the commit to collect it.

tests/README.rst
martinpitt: OK, let's do that in #2925. I dropped that commit from this PR.

.github/workflows/validate.yml
jkonecny12: It is but it gives you nice overview when there is plenty of pylint errors too.

jkonecny12 (approved):
Looks great now. Thanks!

martinpitt:
I put back the "Fix ci-tasks container shell instructions" commit as per discussion with @jkonecny12 , and we'll fix the "run single test" separately in #2925.

PR #2925, opened 2020-10-14
Fix local tests run inside of container

jkonecny12:
Following the documentation to run single test will not work for two reasons.

- it will remove --rm -v ... from the podman call
- it will try to download image from the $(CI_TASKS_TEST_ARGS) variable so this
  have to be last parameter

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -89,6 +89,7 @@ CI_TASKS_DOCKERFILE ?= $(srcdir)/dockerfile/ci-tasks
 CI_TASKS_NAME ?= anaconda/ci-tasks
 CI_TASKS_TAG ?= $(GIT_BRANCH)
 CI_TASKS_TEST_ARGS ?= --rm -v $(srcdir):/anaconda:Z
+CI_TASKS_CMD ?=
 
 SKIP_BRANCHING_CHECK ?= "false"
 
@@ -227,7 +228,7 @@ ci:
 	fi
 
 container-ci:
-	$(CONTAINER_ENGINE) run $(CONTAINER_TEST_ARGS) $(CI_TASKS_TEST_ARGS) $(CI_TASKS_NAME):$(CI_TASKS_TAG)
+	$(CONTAINER_ENGINE) run $(CONTAINER_TEST_ARGS) $(CI_TASKS_TEST_ARGS) $(CI_TASKS_NAME):$(CI_TASKS_TAG) $(CI_TASKS_CMD)
 
 check-branching:
 # checking if branching can be finished and all the pieces are in place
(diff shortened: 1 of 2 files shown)


Makefile.am
```diff
@@ -82,13 +82,13 @@ TEST_INST_BUILD_DIR = $(BUILD_RESULT_DIR)/02-test-install
 # build/run tweaks for all containers, such as --cap-add
 CONTAINER_ENGINE ?= podman
 CONTAINER_BUILD_ARGS ?=
-CONTAINER_TEST_ARGS ?=
+CONTAINER_TEST_ARGS ?= --rm -v $(srcdir):/anaconda:Z
```
martinpitt, line 85: Admittedly I liked [this version](https://github.com/martinpitt/anaconda/commit/d68ba304861d) better as it simplifies the user-facing invocation for a shell, but that's not a veto.

I intended the generic `CONTAINER_*_ARGS` to be empty by default and user-definable, so that the user can add stuff like `--cap-add`, `--device`, or additional `-v`s without breaking the default behaviour.

Makefile.am
```diff
@@ -227,7 +227,7 @@ ci:
 	fi
 
 container-ci:
-	$(CONTAINER_ENGINE) run $(CONTAINER_TEST_ARGS) $(CI_TASKS_TEST_ARGS) $(CI_TASKS_NAME):$(CI_TASKS_TAG)
+	$(CONTAINER_ENGINE) run $(CONTAINER_TEST_ARGS) $(CI_TASKS_NAME):$(CI_TASKS_TAG) $(CI_TASKS_TEST_ARGS)
```
martinpitt, line 230: I see we need a trailing variable to fix the "run a single test" use case (I didn't think of that in #2923 as I never use that -- iterating in an interactive shell is just so much nicer). We could have a `$(CI_TASKS_CMD)` for that, reflecting the container's `CMD`?

martinpitt (commented):
WFM if you prefer this route, at least it fixes the invocation. I just wrote a proposal here, if you don't like it I'll approve. Dekuji!

jkonecny12:
OK so what about replacing this whole commit with just the new ``CI_TASKS_CMD``?

martinpitt:
I put back the doc fix for "run shell" to #2923.

jkonecny12:
**UPDATED:**
* Add just a new makefile variable to be able to specify entrypoint parameter.

martinpitt (approved):
Dekuji! I'll backport this to #2920

VladimirSlavik (approved):
Looks good to me. Thank you!

M4rtinK (approved):
Looks good to me as well. :)

poncovka (approved):
Looks good to me. Thanks!

PR #2928, opened 2020-10-14
post-scripts need to be sorted and anaconda log files relabeled

jstodola:
Sort the order of execution of post-scripts, since glob.glob() doesn't return a sorted list.
Relabel the copied anaconda log files - they could be labeled incorrectly, like in case of hawkey.log, see bug 1885772.

Resolves: rhbz#1870493


VladimirSlavik (approved):
Thank you for porting!

poncovka (approved):
Looks good to me. Thanks!

jkonecny12 (approved):
LGTM!

PR #2929, opened 2020-10-15
Fix password quality test

M4rtinK:
Adjust to password quality scoring changes introduced in libpwquality 1.4.4-1.

### DIFF
--- a/tests/nosetests/pyanaconda_tests/password_quality_test.py
+++ b/tests/nosetests/pyanaconda_tests/password_quality_test.py
@@ -95,8 +95,8 @@ class PasswordQuality(unittest.TestCase):
         request.password = "1234567891"
         check = input_checking.PasswordValidityCheck()
         check.run(request)
-        self.assertEqual(check.result.password_score, 1)
-        self.assertEqual(check.result.status_text, _(constants.SecretStatus.WEAK.value))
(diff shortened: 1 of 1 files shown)


M4rtinK:
@t8m Are you sure libpwquality 1.4.4 is working fine ? On a first look it seem to me that some pretty weak passwords are now getting a higher score & as if dictionary attach checking was not being used (the "anaconda" case).

martinpitt:
Just like this the test wouldn't work on e.g. Fedora 33 any more, so if we need something like that, the test should accept either answer or a range, up to the degree where the answers are reasonable.

However, now the [tests fail](https://github.com/rhinstaller/anaconda/pull/2929/checks?check_run_id=1256596165#step:6:1500) exactly the other way around -- in other words, everything is fine again in master. I just retried #2928, and it's green again.

There has been [no further libpwquality update](https://bodhi.fedoraproject.org/updates/?packages=libpwquality), so apparently something else changed in rawhide?

I suppose we can just close this then. Thanks anyway!

t8m:
There was 1.4.3 version for one day that was broken. The 1.4.4 version should be OK and there should be no difference from the 1.4.2 in scores and rejections.

martinpitt:
A, thanks @t8m!

PR #2931, opened 2020-10-15
Add GH action to build and push master container to registry

jkonecny12:
Build and push container to Quay.io container registry every day. Also make it possible to manually dispatch this action.

This container should be tested before pushing to the registry but that would be follow up when this is working.

The container will be then used as source to run unit-tests GH action.
Images are published to our new [Quay.io repository](https://quay.io/repository/rhinstaller/anaconda-ci).

You can see results on my [fork repository](https://github.com/jkonecny12/anaconda/actions?query=workflow%3A%22Build%2C+test+and+push+container%22).

**TODO:**
~- [ ] Move this workflow file into the existing workflow. [Seems](https://github.community/t/how-do-i-specify-job-dependency-running-in-another-workflow/16482) that it's not possible to use jobs from a different workflow :(~
- [x] Run tests on the container first to prevent regressions
- [x] Change Makefile paths to pull the image by default and build the correct name
- [x] Change validate task to pull image instead of building it


.github/workflows/container-autoupdate.yml
```diff
@@ -0,0 +1,34 @@
+name: Build, test and push container
+on:
+  schedule:
+    - cron: 0 0 * * *
+  workflow_dispatch:
+
+jobs:
+  push-container:
+    name: Push container to Quay.io
```
martinpitt, line 9: Nitpick: "Refresh container", as it also includes the build step. And let's not repeat the destination too much.

.github/workflows/container-autoupdate.yml
```diff
@@ -0,0 +1,34 @@
...
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v1
+
+      - name: Login to Quay.io
+        uses: docker/login-action@v1
+        with:
+          registry: quay.io
+          username: ${{ secrets.QUAY_USERNAME }}
+          password: ${{ secrets.QUAY_PASSWORD }}
```
martinpitt, line 24: I suggest committing them to our internal secrets repo. Also, any chance to use a less privileged token for the upload instead of the full user account? (not sure if quay supports that)

.github/workflows/container-autoupdate.yml
```diff
@@ -8,6 +8,7 @@ jobs:
   push-container:
     name: Push container to Quay.io
     runs-on: ubuntu-latest
+    needs: ["unit-tests", "rpm-tests"]
```
martinpitt, line 11: rpm-tests uses a separate container, no need to depend on that.

martinpitt (changes requested):
Nice, thanks! Can you please also rename the "ci-tasks" container in the Makefile and workflow, so that the container name is consistent between quay and local builds?

.github/workflows/container-autoupdate.yml
jkonecny12: That is a bot account from Quay.io. It has Write access to push the images.

.github/workflows/container-autoupdate.yml
jkonecny12: Good point! This was just a test and as you can see above I have to merge workflow files first to make this work.

.github/workflows/container-autoupdate.yml
jkonecny12: KK, will fix that.

jkonecny12:
> Nice, thanks! Can you please also rename the "ci-tasks" container in the Makefile and workflow, so that the container name is consistent between quay and local builds?

I have that in my TODO above ;).

jkonecny12:
This new version seems to be working correctly on my fork:

https://github.com/jkonecny12/anaconda/actions/runs/315838229

However, it's not nice that the build is done twice so it may not have the same result... Do we want to solve this or is it good enough?

.github/workflows/container-autoupdate.yml
```diff
@@ -25,8 +25,8 @@ jobs:
       - name: Build and push container
         uses: docker/build-push-action@v2
         with:
-          context: dockerfile/ci-tasks/
-          file: dockerfile/ci-tasks/Dockerfile
+          context: dockerfile/anaconda-ci/
+          file: dockerfile/anaconda-ci/Dockerfile
```
martinpitt: I suggest to rebase this PR to do the renaming first; then it's less confusing to look at.

Makefile.am
```diff
@@ -84,9 +84,9 @@ CONTAINER_ENGINE ?= podman
 CONTAINER_BUILD_ARGS ?=
 CONTAINER_TEST_ARGS ?=
 
-# ci-tasks container
-CI_TASKS_DOCKERFILE ?= $(srcdir)/dockerfile/ci-tasks
-CI_TASKS_NAME ?= anaconda/ci-tasks
+# anaconda-ci container
+CI_TASKS_DOCKERFILE ?= $(srcdir)/dockerfile/anaconda-ci
+CI_TASKS_NAME ?= rhinstaller/anaconda-ci
```
martinpitt, line 89: The variables should be renamed as well for consistency. You could probably ignore the "anaconda-" prefix though, i. e. `CI_DOCKERFILE` seems fine? `</bikeshedding>`

.github/workflows/container-autoupdate.yml
```diff
@@ -0,0 +1,34 @@
...
+
+      - name: Login to Quay.io
+        uses: docker/login-action@v1
+        with:
+          registry: quay.io
+          username: ${{ secrets.QUAY_USERNAME }}
+          password: ${{ secrets.QUAY_PASSWORD }}
+
+      - name: Build and push container
+        uses: docker/build-push-action@v2
```
martinpitt, line 53: This is a bit awkward. It's one step for two distinct actions (building and pushing a container), and we want to put the validation *in between*. This causes the big code duplication and also a race condition -- the second build may look different from the first one. Instead, I suggest to do the build manually with `make anaconda-ci-build` (which will also validate that our Makefile rule works as expected), then run tests (`make container-ci`), and only *afterwards* push the container. That's a simple `docker tag` and `docker push`, so honestly I don't see the need to invoke an external action here.

martinpitt (changes requested):
Thanks! I have a proposal how to make this simpler and just build once.

Makefile.am
jkonecny12: I did not want to break existing stuff but it's still early so let's rename even these.

.github/workflows/container-autoupdate.yml
jkonecny12: Fixed.

.github/workflows/container-autoupdate.yml
jkonecny12: I'll try to experiment with this a bit more.

jkonecny12:
**UPDATED:**
I've found a way how to not build the image twice. Or more precisely it's build twice but second time it's just using cache from the first build so it's basically no-op. I'm still using the ``push-and-build`` action because that's recommended way based on the documentation.

What do you think now @martinpitt ?

jkonecny12:
Now I'm looking on validate to use our new images from quay.

.github/workflows/container-autoupdate.yml
```diff
@@ -0,0 +1,34 @@
+name: Build, test and push container
+on:
+  schedule:
+    - cron: 0 0 * * *
+  workflow_dispatch:
+
+jobs:
+  push-master-container:
+    name: Refresh Rawhide container
```
martinpitt, line 10: I'd avoid putting "Rawhide" here. It doesn't coincide with how the name/tag is going to be (`:master` or `:f33` etc.), and it's a source of diff noise between branches. Let's just say "anaconda-ci container"?

.github/workflows/container-autoupdate.yml
martinpitt: OK, fair enough.

.github/workflows/container-autoupdate.yml
```diff
@@ -0,0 +1,34 @@
...
+      - name: Build and push container
+        uses: docker/build-push-action@v2
+        with:
+          context: dockerfile/ci-tasks/
+          file: dockerfile/ci-tasks/Dockerfile
+          platforms: linux/amd64
+          push: true
+          tags: |
+            quay.io/rhinstaller/anaconda-ci:latest
+            quay.io/rhinstaller/anaconda-ci:master
```
martinpitt, line 69: Again, let's not hardcode `:master` here, but use `$(GIT_BRANCH)` like `Makefile` does. That way you can keep the workflow file in sync between branches. Also, I think `:latest` makes relatively little sense here. At least it should be pushed conditionally only for master.

.github/workflows/container-autoupdate.yml
```diff
@@ -2,18 +2,53 @@ name: Build, test and push container
...
       - name: Checkout anaconda repository
         uses: actions/checkout@v2
 
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v1
+        with:
+          driver-opts: network=host
+
+      - name: Build and store image locally for testing
+        uses: docker/build-push-action@v2
```
martinpitt, line 28: Ugh, why? A plain docker build (preferably `make -f Makefile.am anaconda-ci-build`) already creates it as a local docker image. It's totally unnecessary to involve yet another registry here, it's just needlessly copying around the image on disk.

Also, we really want to use the same command that a developer would do -- namely, `make anaconda-ci-build`. Let's please test this code path as well, and simplify things a lot.

.github/workflows/container-autoupdate.yml
martinpitt: I stand by my opinion that replacing build-push-action@v2 with a simple `docker tag && docker push` run command after `make anaconda-ci-build && make container-ci` is both simpler, structurally what we want, and much easier to reproduce locally.

martinpitt (changes requested):
Hmm, this is now even more complicated rather than simpler.. If you really like build-and-push so much, then I suggest to push to a `:staging-$branchname` tag first, run tests against that, and re-tag it as `:branchname` on success. At least that has the added benefit that it's easy for a developer to download the staging image and investigate failures. However, let's do what developers do and use our Makefile commands as much as possible.

Thanks!

.github/workflows/container-autoupdate.yml
jkonecny12: kk will do

.github/workflows/container-autoupdate.yml
jkonecny12: Good point!

jkonecny12:
Good idea with the simplification. By not using that action and just using make solution I was able to reduce number of steps.

Introduced:
* Add support for multiple branches for future need.
* Add latest tag push (master branch only).
* Add new Makefile target to push to registry.
 - This will really simplify decision about what should be used as name, tag... Just leave makefile to decide.
* First implementation of validate change to use registry instead of building it's own containers is in place.

martinpitt (approved):
Ooh, this looks really nice now! :heart_eyes: My only nitpick now is that the commit message for "push new image only if unit tests went well" is misleading now, as there is no code duplication any more. Nice that you addressed the "rebuild container on changes to dockerfiles/" part as well already!

jkonecny12:
> Ooh, this looks really nice now! heart_eyes My only nitpick now is that the commit message for "push new image only if unit tests went well" is misleading now, as there is no code duplication any more. Nice that you addressed the "rebuild container on changes to dockerfiles/" part as well already!

Good point! I'll change the commit and have to test if my changes in validate really works.

jkonecny12:
I finally managed to put all the pieces together. This PR is now ready for review.

I've managed to run PR tests on the registry container or build container if ``dockerfile/anaconda-ci`` files has changed. We have this condition to avoid regression from moving Rawhide but also from introducing bug in the ``anaconda-ci`` dockerfile.

To see results of the latest test look [here](https://github.com/jkonecny12/anaconda/pull/4/checks?check_run_id=1297599529).

Tests on this PR are failing which is expected and couldn't be fixed because I'm changing Makefile command name which is used in the workflow in the target branch (the one used here).

This should be partially backported to ``rhel-8`` branch.

martinpitt (approved):
This looks and works really nicely, thanks! It's also fairly generic, so we can apply the same pattern for the rpm-tests container.

When running `make -f Makefile.am container-ci` on current master + this PR, I do get a nosetests failure:

```
======================================================================
ERROR: Test a fake image.
----------------------------------------------------------------------
  File "/anaconda/tests/nosetests/pyanaconda_tests/module_source_live_image_test.py", line 204, in fake_image_test
    self.assertGreater(result.required_space, 0)
  File "/usr/lib64/python3.9/unittest/case.py", line 1227, in assertGreater
    if not a > b:
TypeError: '>' not supported between instances of 'NoneType' and 'int'
-------------------- >> begin captured logging << --------------------
anaconda.modules.payloads.source.live_image.initialization: DEBUG: Unknown required space.
--------------------- >> end captured logging << ---------------------
```

But this is certainly not related to this PR, and such regressions should be caught by both PR and container refresh gating.

PR #2932, opened 2020-10-19
Document how to start a self-hosted GitHub actions runner locally

martinpitt:

PR #2934, opened 2020-10-19
TEST: Try to capture internal runner

martinpitt:
Don't land this! :-) We want to see what happens, in the context of PR #2933.

### DIFF
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -2,35 +2,14 @@ name: validate
 on: [push, pull_request_target]
 jobs:
   unit-tests:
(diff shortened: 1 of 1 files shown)


martinpitt:
So this *did* run on the internal runner, which is not that nice. It's heavily guarded, but not quite what we expected. Let's get in #2932 and try this again.

jkonecny12:
Heh, interesting what could be done by an easy PR.

martinpitt:
#2933 landed, re-running.

martinpitt:
@jkonecny12 : That now went as intended, it runs the workflow from master.  *phew*

PR #2935, opened 2020-10-19
rhbz#1888697 list important major languages

sundeep-anand:
Listing the most important major languages make them super-easy to find for international users. (_lesser scroll_)
Identification of those languages is a challenge, however we may refer [gnome-control-center](https://gitlab.gnome.org/GNOME/gnome-control-center/-/blob/master/panels/common/cc-common-language.c#L222).

#### Implementation

- One option is to join (_prepend_) those identified languages to `locales` in `pyanaconda/ui/gui/spokes/welcome.py`. Here, we may have separators to avoid confusion: separate local/national languages to common/major ones. So, `major languages | national languages | other languages`. However, in a many cases the count of national languages remain less than 4 which may leave the separators very close! And may not look nice.

- Another option could be `major languages + national languages | other languages`. I feel this may create confusion.

- As we are identifying local/national languages already to help ease user's search (_for their languages_), it makes sense to populate the first list with identified common languages in the event when `territory lookup` fails instead of just `en_US`.

Should we try moving `get_common_languages` to **langtable** instead?

Please feel free, to suggest better implementation way.

### DIFF
--- a/pyanaconda/localization.py
+++ b/pyanaconda/localization.py
@@ -72,6 +72,11 @@ def get_language_id(locale):
     return langtable.parse_locale(locale).language
 
 
+def get_common_languages():
+    """Return common languages to prioritize them"""
+    return langtable.list_common_languages()
+
+
 def is_supported_locale(locale):
     """Function that tells if the given locale is supported by the Anaconda or
     not. We consider locales supported by the langtable as supported by the
--- a/pyanaconda/ui/gui/spokes/welcome.py
+++ b/pyanaconda/ui/gui/spokes/welcome.py
@@ -162,7 +162,12 @@ class WelcomeLanguageSpoke(StandaloneSpoke, LangLocaleHandler):
         # get language codes for the locales
         langs = [localization.get_language_id(locale) for locale in locales]
 
-        # check which of the geolocated languages have translations
+        # get common languages and append them here
+        common_langs = localization.get_common_languages()
+        common_langs_len = len(set(common_langs).difference(set(langs)))
+        langs += common_langs
+
+        # check which of the geolocated or common languages have translations
         # and store the iterators for those languages in a dictionary
         langs_with_translations = {}
         itr = store.get_iter_first()
@@ -181,7 +186,7 @@ class WelcomeLanguageSpoke(StandaloneSpoke, LangLocaleHandler):
             langs_with_translations[DEFAULT_LANG] = lang_itr
             locales = [DEFAULT_LANG]
 
-        # go over all geolocated languages in reverse order
+        # go over all geolocated and common languages in reverse order
         # and move those we have translation for to the top of the
         # list, above the separator
         for lang in reversed(langs):
@@ -195,8 +200,13 @@ class WelcomeLanguageSpoke(StandaloneSpoke, LangLocaleHandler):
 
         # And then we add a separator after the selected best language
         # and any additional languages (that have translations) from geoip
-        newItr = store.insert(len(langs_with_translations))
+        langs_with_translations_len = len(langs_with_translations)
+        newItr = store.insert(langs_with_translations_len)
         store.set(newItr, 0, "", 1, "", 2, "", 3, True)
+
+        # add a separator before common languages as well
+        commonLangsItr = store.insert(langs_with_translations_len - common_langs_len)
+        store.set(commonLangsItr, 0, "", 1, "", 2, "", 3, True)
 
         # setup the "best" locale
         locale = localization.setup_locale(locales[0], self._l12_module)
--- a/anaconda.spec.in
+++ b/anaconda.spec.in
@@ -27,7 +27,7 @@ Source0: %{name}-%{version}.tar.bz2
 %define gtk3ver 3.22.17
 %define helpver 22.1-1
 %define isomd5sumver 1.0.10
-%define langtablever 0.0.53
+%define langtablever 0.0.54
 %define libarchivever 3.0.4
 %define libblockdevver 2.1
 %define libreportanacondaver 2.0.21-1


jkonecny12:
Hello @sundeep-anand thanks for your PR. Yes, this seems something which should be part of the langtable. What do you think @mike-fabian ? Is it appropriate to have list of these languages in langtable or is there a better solution?

mike-fabian:
Yes, we can add this to langtable.

I must say that I don't like such a list of common languages very much, it always annoys me a bit when I open gnome-control-center. My own native language (German) is in the list of common languages, but I feel that it might annoy people that their language isn't. 

But anyway, I am not against adding something like that to langtable, it is probably better to have it there than in anaconda.

juhp:
Just to comment, I believe Gnome used number of speakers as the main criteria for the top listed languages.

VladimirSlavik:
As to the technical matters:

- I tested this and it works, probably as intended. Thank you!
- The language support spoke does not seem to be affected, just as it should be.
- Yes, langtable is a better place for such a list. Thank you Mike for considering this.
- *This happens only if geolocation is off*. I think it's an important distinction for the rest of the discussion. I am also not sure if that was the intention?
- The discussion in first comment seems for more than the one commit. Is this part of some larger plan? Will more be added? Having a full picture would help.

Apart from that, there is the "soft" side of this. It's a potentially loaded decision / change.

- Someone (@juhp?) asked for this and someone else (@sundeep-anand) implemented this. [Bug](https://bugzilla.redhat.com/show_bug.cgi?id=1888697) has no links to anywhere else. So two people definitely think this is a good idea. Who else?
- Do we need to run this by some sort of Fedora board?
- In the Gnome thing at least there is some balance: The list is initially very short and lists only the most common stuff, so that it fits on the screen space. So many benefit from having the choice "above the fold". Others can expand the list for everything. Here, we already do have a "full" list in some sense, there is no clear split between these important items and the rest. Also note the Gnome list changes to alphabetical after expanding.
- Should we also order these? How? Alphabetically? By estimated amount of speakers? By expected % of speakers using Fedora? By expected share of computer users among speakers of these languages?

At any rate, I am told @M4rtinK already had some discussions about this, so switching to him...

sundeep-anand:
depends https://github.com/mike-fabian/langtable/pull/12
**edit**: common languages are being appended to the priority languages. (_determined by geolocation_)
a discussion around the use case is [here](https://lists.fedoraproject.org/archives/list/devel@lists.fedoraproject.org/thread/L5G3KAHKG5BLLP7NDUWK24RO3UX25TLQ/).

pyanaconda/ui/gui/spokes/welcome.py
```diff
@@ -162,6 +162,10 @@ def initialize(self):
         # get language codes for the locales
         langs = [localization.get_language_id(locale) for locale in locales]
 
+        # get common languages and append them here
+        common_locales = localization.get_common_locales()
+        langs += [localization.get_language_id(locale) for locale in common_locales]
```
Zokormazo, line 167: Will this duplicate languages on territories where a common language is a territory languace too?

sundeep-anand: no.. the language will appear once.. in territory and be omitted from common.

pyanaconda/localization.py
```diff
@@ -72,6 +72,11 @@ def get_language_id(locale):
     return langtable.parse_locale(locale).language
 
 
+def get_common_locales():
+    """Return common locales to prioritize them"""
+    return langtable.list_common_locales()
```
poncovka, line 77: Please, update the required `langtable` version in the `anaconda.spec.in` file.

poncovka (commented):
Otherwise, the code looks good to me. Thank you!

jkonecny12:
Failure of the tests is expected here:
```
E1101(no-member):/anaconda/pyanaconda/localization.py:77,11: get_common_locales: Module 'langtable' has no 'list_common_locales' member
```

VladimirSlavik:
/packit build

Sorry about that.

jkonecny12:
Seems that the langtable is still not in the Fedora yet.

M4rtinK (approved):

M4rtinK:
Sorry for taking me so long to review this. :P
The code looks fine & the updated langtable seems to be in place as well. :)

The RPM test is failing, but this might be fixed by rebasing the PR on latest master. Cold you please do that ? :)

I also did some manual testing and everything seems to work fine:

lang_en_sorting

See the 7 languages under the selected language and *Afrikaans* - these are the major languages this PR is about. :) 

In comparison this is how it looks like at the moment without the PR being applied:

lang_en_no_sorting

As you can there is just an alphabetical list of all languages Anaconda knowns under the currently selected language.

I've also checked with geolocation enabled & all seems to be fine as well:

lang_cs_sorting

One thing I have noticed is, that the separators we use in the language listbox are really hard to see. I don't think it used to be like this & it's possibly due to some theming changes. In any case, this is not really something this PR needs to fix & I'll make a note for myself to look into fixing all the Anaconda separators to be more visible.

M4rtinK:
Also one more thing - the Packit test failure seems to be related to ongoing packit test infra issues & not an issue with this PR.

sundeep-anand:
sorry - not sure how the review got dismissed ..

jkonecny12:
> sorry - not sure how the review got dismissed ..

This is expected it's change of behavior in the project.

jkonecny12 (approved):
Looks good to me too

M4rtinK:
I did a new Anaconda build yesterday, so this change should be now hitting the Fedora Rawhide nightly composes as part of Anaconda 34.14-1. :)

sundeep-anand:
> I did a new Anaconda build yesterday, so this change should be now hitting the Fedora Rawhide nightly composes as part of Anaconda 34.14-1. :)

thank you!

PR #2940, opened 2020-10-21
Fix testing target branch instead of PR branch

jkonecny12:
Hardening done in the commit baafcef39acbf0df181dd4b87176308c07a261e8 work more than expected. Right now we are testing target branch instead of the PR ref which means that we are not testing commits on the PR at all.
Ups...

We have to change what we are checking out to fix this.

We have to merge this ASAP.

### DIFF
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -6,6 +6,9 @@ jobs:
     steps:
       - name: Clone repository
         uses: actions/checkout@v2
+        with:
+          # otherwise we are testing target branch instead of the PR branch (see pull_request_target trigger)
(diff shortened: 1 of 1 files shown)


VladimirSlavik (approved):

rvykydal (approved):
Looks good to me.

M4rtinK (approved):

martinpitt:
@jkonecny12: Ouch, well spotted! We need that in the rhel-8 branch, too.

jkonecny12:
@martinpitt I totally missed that it's also in rhel-8 already. Thanks for info.
However, you should be enjoying your vacation ;).

PR #2941, opened 2020-10-21
DO NOT MERGE -- test PR

jkonecny12:
**DO NOT MERGE* 

This is just a test


jkonecny12:
This was a test if I can create PR from the history before we switched to ``pull_request_target`` trigger. It does not work which is good.

However, it seems that you have to rebase your PR when some workflow fix is merged to master otherwise you will use the old master. Seems that the PR metadata are created when PR changed and it does not take updated default branch changes into account afterwards.

PR #2943, opened 2020-10-22
Do not exclude English from available languages

VladimirSlavik:
English is the untranslated state, so it needs no translation files.
All its locales are always "available".

Resolves: [rhbz#1890340](https://bugzilla.redhat.com/show_bug.cgi?id=1890340)
Resolves: [rhbz#1890342](https://bugzilla.redhat.com/show_bug.cgi?id=1890342)

### DIFF
--- a/pyanaconda/localization.py
+++ b/pyanaconda/localization.py
@@ -342,9 +342,15 @@ def locale_has_translation(locale):
     locale has a translation, or because there is another translation to fall back onto. In
     reality, the fallback is mostly of the type "ja_JP" -> "ja".
 
+    For English, always return true, because that is the "untranslated" state which does not need
+    translation files present to work.
+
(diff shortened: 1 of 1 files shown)


poncovka (approved):
Looks good to me. Thanks!

AdamWill:
Tested, confirmed it fixes both bugs.

jkonecny12 (approved):
Thanks for fixing this!

jkonecny12:
We should make a release with this soon.

VladimirSlavik:
Thanks for testing, Adam! Merging...

AdamWill:
Can we please do a Rawhide build with this included ASAP? It breaks openQA and I don't really want to add a workaround needle for a bug that's already fixed. Thanks!

PR #2901, opened 2020-10-06
Organize nosetests in directories (1)

VladimirSlavik:
This works but it's just an initial proof of concept. There is one most important question for this:

**How do we group the tests? By module? By function? By directory?**

eg.: Kickstart handler tests could go with boss, or with other kickstart. Same for subscription stuff; there is the module, and there are also tests for the gui.

PR #2944, opened 2020-10-22
Master nosetests directories 2

VladimirSlavik:
Second round of test moves.

(Previous: #2901)


VladimirSlavik:
I think this should not conflict with anything recently opened, but I might be wrong.

jkonecny12 (approved):
LGTM! Not affecting me.

poncovka (approved):
Looks good to me. Thanks!

poncovka:
> I think this should not conflict with anything recently opened, but I might be wrong.

I might have a conflict, but it doesn't matter.

VladimirSlavik:
Thank you! Merging.

PR #2945, opened 2020-10-23
Remove unused code again

poncovka:
Remove everything that is no longer used.


pyanaconda/modules/storage/partitioning/interactive/scheduler_interface.py
```diff
@@ -173,13 +173,6 @@ def UpdateContainerData(self, request: Structure, container_name: Str) -> Struct
         self.implementation.update_container_data(request, container_name)
         return DeviceFactoryRequest.to_structure(request)
 
-    def GetPartitioned(self) -> List[Str]:
```
VladimirSlavik, line 176: This is not used after the very recent #2917. Not sure if that means anything.

VladimirSlavik (approved):
Another "yes please" for me.

However, I am not really the one to judge many of the changes. Maybe it would have been useful to split this into more PRs by affected area/dev?

pyanaconda/modules/subscription/subscription.py
```diff
@@ -330,13 +328,6 @@ def set_system_purpose_with_task(self):
 
     # subscription request
 
-    def _validate_subscription_request_type(self, request_type):
```
M4rtinK, line 333: All these look like aritifacts of the Fedora <-> RHEL rebase passes & should be fine to drop.

M4rtinK (approved):
Looks good to me - nice cleanup. :) 

If we need some of the utility stuff later I guess we can always grab it from the Git history.

pyanaconda/localization.py
```diff
@@ -102,36 +102,6 @@ def locale_supported_in_console(locale):
     return set(locale_scripts).issubset(SCRIPTS_SUPPORTED_BY_CONSOLE)
 
 
-def langcode_matches_locale(langcode, locale):
```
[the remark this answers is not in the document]
VladimirSlavik: Possibly. I think I did not check if these things are actually used.

poncovka:
Rebased.

jkonecny12 (approved):
LGTM! Nice Anaconda code diet!

From the Payload PoV, I think it's not a problem to remove these parts. I think none of these will be usable with the new solution so there is not really a reason to have them.
However, It would be great to check if addons are not using those... Did you looked there?

poncovka:
> However, It would be great to check if addons are not using those... Did you looked there?

Addons should be fine.

PR #2946, opened 2020-10-23
Names of signal handlers should have the on_ prefix

poncovka:
Add a new rule to the code conventions and rename signal handles that break it.
These handles are reported as unused code and it is easier to filter them out
when they use the same prefix.


VladimirSlavik (approved):
Yes please - thank you!

jkonecny12 (approved):
LGTM!

PR #2948, opened 2020-10-23
Add workaround to fix support for installtree without repo

jkonecny12:
Sources should always point to the repository, however, right now we are in
middle of the migration process to Payloads sources so only base
repositories are migrated and they are more or less only reflecting user data.
To solve this for now we are adding back support for .treeinfo file and not
just repodata. This should be removed when payload source migration will be
finished.

Resolves: rhbz#1854825
(cherry picked from commit 4db89b9be8007b093042f94b18a73042f3526bc9)

Backport of https://github.com/rhinstaller/anaconda/pull/2717.

### DIFF
--- a/pyanaconda/modules/payloads/source/utils.py
+++ b/pyanaconda/modules/payloads/source/utils.py
@@ -160,6 +160,17 @@ def verify_valid_repository(path):
     if os.path.exists(repomd_path) and os.path.isfile(repomd_path):
         return True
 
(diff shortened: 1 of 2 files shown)


jkonecny12:
Smoke test done. I'm not able to easily test this on Fedora and it's working on RHEL so it should be fine.

poncovka (approved):
Looks good to me.

PR #2949, opened 2020-10-25
Clean up the Payload class

poncovka:
Remove functions and attributes that don't have to be shared by all subclasses.


poncovka:
Rebased and tested.

M4rtinK (approved):

jkonecny12 (approved):
LGTM!

All these function have to change during the migration so why not make the code smaller before the transition.

PR #2952, opened 2020-10-26
Improve the documentation of inst.variant

poncovka:
Document that `inst.variant` has to be used together with `inst.product`.


jkonecny12 (approved):
Thanks a lot for this!

PR #2955, opened 2020-10-26
Add container image build badge

jkonecny12:
This would be super helpful to get info that our image build is failing!

And it looks cool too ;).

Blocked by https://github.com/rhinstaller/anaconda/pull/2954 . I want to be sure that the links here are fine and they will work after first run with the new name.

----------------------
Adding badge is not a notable change but having automatic container builds for testing is ;).


martinpitt (approved):
Nice!

poncovka (approved):
Looks good to me.

PR #2958, opened 2020-10-27
Mock os.stat in the unit tests

poncovka:
Fix a failing test by mocking the `os.stat` module.


jkonecny12 (approved):
Oh my, this again. Thanks for the fix.

jkonecny12:
I guess we want to backport this to `rhel-8` branch too?

poncovka:
> I guess we want to backport this to `rhel-8` branch too?

This test is not in the `rhel-8` branch, so the fix doesn't have to be backported.

PR #2959, opened 2020-10-27
Don't include unknown files in the updates image

poncovka:
For example, Docker files and Git Hub workflows.

### DIFF
--- a/scripts/makeupdates
+++ b/scripts/makeupdates
@@ -216,7 +216,7 @@ def copy_updated_files(tag, updates, cwd, builddir):
 
         if gitfile.endswith('.spec.in') or (gitfile.find('Makefile') != -1) or \
            gitfile.endswith('.c') or gitfile.endswith('.h') or \
-           gitfile.endswith('.sh') or gitfile == 'configure.ac':
+           gitfile.endswith('.sh'):
             continue
 
         if gitfile.endswith('.glade'):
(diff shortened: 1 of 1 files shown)


martinpitt (dismissed):
Makes sense, thank you!

scripts/makeupdates
```diff
@@ -289,19 +289,10 @@ def copy_updated_files(tag, updates, cwd, builddir):
...
-            fields = gitfile.split('/')
-            subdir = fields[0]
-            if subdir in ['po', 'scripts','command-stubs', 'tests',
-                          'docs', 'fonts', 'utils',
-                          'liveinst', 'dracut', 'data']:
-                continue
-            else:
-                sys.stdout.write("Including %s\n" % (gitfile,))
-                install_to_dir(gitfile, tmpupdates)
         else:
```
jkonecny12, line 292: What about removing this whole `else` part?

jkonecny12 (approved):
Looks good to me otherwise.
Thanks for this!

Don't we want to remove also ``configure.ac`` from the ignores above?

scripts/makeupdates
poncovka: Fixed.

poncovka:
Updated.

jkonecny12 (approved):
LGTM!

PR #2960, opened 2020-10-28
Fix localization context issues in GUI

M4rtinK:
Fix localization context issues in GUI by direct string translations
instead of just marking strings as translatable in spokes and hubs
and translating them before use. This is fragile as strings might
be marked in Initial Setup or Anaconda Addons yet translated and
displayed in Anaconda, with context that does not match.

We can't do this directly in class attributes due to these being
evaluated much to early. Instead, put strings into properties &
make sure these properties are implemented by turning the appropriate
base classes into abstract base classes and the required properties
into abstract properties.

Also when we are at it convert some non-localization class attributes
to properties for better consistency.

**NOTE:**
- still fighting some issues related to the code that collects spokes, so not really tested as working yet

**TODO:**
- adjust Initial Setup
- adjust the Hello World addon
- adjust Kdump addon
- adjust OSCAP addon
- run full kstest suite on this change

### DIFF
--- a/pyanaconda/ui/common.py
+++ b/pyanaconda/ui/common.py
@@ -17,7 +17,7 @@
 # Red Hat, Inc.
 #
 
-from abc import ABCMeta, abstractproperty
+from abc import ABC, ABCMeta, abstractproperty, abstractmethod
 
 from pyanaconda.core.configuration.anaconda import conf
 from pyanaconda.core.constants import ANACONDA_ENVIRON, FIRSTBOOT_ENVIRON, SETUP_ON_BOOT_RECONFIG
@@ -159,7 +159,7 @@ class FirstbootOnlySpokeMixIn(object):
             return False
 
 
-class Spoke(object, metaclass=ABCMeta):
+class Spoke(ABC):
     """A Spoke is a single configuration screen.  There are several different
        places where a Spoke can be displayed, each of which will have its own
        unique class.  A Spoke is typically used when an element in the Hub is
@@ -170,26 +170,8 @@ class Spoke(object, metaclass=ABCMeta):
        to be shown.  Regardless, the UI of a Spoke should be given by an
        interface description file like glade as often as possible, though this
        is not a strict requirement.
-
-       Class attributes:
-
-       category   -- Under which SpokeCategory shall this Spoke be displayed
-                     in the Hub?  This is a reference to a Hub subclass (not an
-                     object, but the class itself).  If no category is given,
-                     this Spoke will not be displayed.  Note that category is
-                     not required for any Spokes appearing before or after a
-                     Hub.
-       icon       -- The name of the icon to be displayed in the SpokeSelector
-                     widget corresponding to this Spoke instance.  If no icon
-                     is given, the default from SpokeSelector will be used.
-       title      -- The title to be displayed in the SpokeSelector widget
-                     corresponding to this Spoke instance.  If no title is
-                     given, the default from SpokeSelector will be used.
     """
 
-    category = None
-    icon = None
-    title = None
 
     def __init__(self, storage, payload):
         """Create a new Spoke instance.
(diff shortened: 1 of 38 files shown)


pyanaconda/ui/common.py
```diff
@@ -501,6 +537,8 @@ def status(self):
         return None
 
 
+
+
```
poncovka, line 541: Please, remove the extra lines.

pyanaconda/ui/gui/spokes/custom_storage.py
```diff
@@ -130,6 +127,14 @@ def __init__(self, data, storage, payload):
         self._boot_loader = STORAGE.get_proxy(BOOTLOADER)
         self._disk_selection = STORAGE.get_proxy(DISK_SELECTION)
 
+    @property
+    def category(self):
+        return SystemCategory
+
+    @property
+    def title(self):
+        return _("MANUAL PARTITIONING")
```
poncovka, line 136: This looks like a bug. Please, replace the upper-cased string with a capitalized string.

pyanaconda/ui/gui/spokes/blivet_gui.py
```diff
@@ -133,6 +127,16 @@ def __init__(self, data, storage, payload):
...
 
+    @property
+    def category(self):
+        """Category this spoke belongs to."""
+        return SystemCategory
+
+    @property
+    def title(self):
+        """Title of the spoke (will be displayed on the hub)."""
+        return _("GUI|Spoke", "_Blivet-GUI Partitioning")
```
poncovka, line 138: Are you sure that  this works? Shouldn't you use `C_`?

pyanaconda/ui/gui/hubs/__init__.py
```diff
@@ -172,7 +172,7 @@ def _createBox(self):
                     spoke.initialize()
                     continue
 
-                spoke.selector = AnacondaWidgets.SpokeSelector(C_("GUI|Spoke", spoke.title),
```
poncovka, line 175: Shouldn't you use `C_("GUI|Spoke", "...")` for all titles of our spokes?

pyanaconda/ui/tui/hubs/__init__.py
```diff
@@ -61,6 +62,11 @@ def __init__(self, data, storage, payload):
         # we want user input
         self.input_required = True
 
+    @property
+    @abstractmethod
+    def title(self):
+        return _("Default HUB title")
```
poncovka, line 68: Just return `None`, please. The property has to be redefined anyway.

pyanaconda/ui/tui/spokes/__init__.py
```diff
@@ -54,7 +54,10 @@ def __init__(self, data, storage, payload):
         Spoke.__init__(self, storage, payload)
 
         self.input_required = True
-        self.title = N_("Default spoke title")
+
+    @property
+    def title(self):
+        return _("Default spoke title")
```
poncovka, line 60: I don't think that we want to have TUI spokes with this default titles.

pyanaconda/ui/tui/tuiobject.py
```diff
@@ -97,7 +98,11 @@ class TUIObject(UIScreen, common.UIObject):
     def __init__(self, data):
         UIScreen.__init__(self)
         common.UIObject.__init__(self, data)
-        self.title = u"Default title"
+
+    @property
+    @abstractmethod
+    def title(self):
+        return _("Default title")
```
poncovka, line 105: Please, return `None`.

poncovka (changes requested):
I don't think that this pull request is working. Spokes are sorted by the category and the title before they are initialized. Also there is the issue with spokes sorted by translated strings, that @jkonecny12 mentioned in the other pull request.

pyanaconda/ui/gui/spokes/blivet_gui.py
M4rtinK: Oh right, looks like I was a bit to eager in my search and replace - BTW, looking at the spokes in TUI, those don't have a context set. I think we might as well do it now when we are at it ("TUI|Spoke").

pyanaconda/ui/gui/hubs/__init__.py
M4rtinK: Yep, definitely! Fixing it now. :)

pyanaconda/ui/tui/hubs/__init__.py
M4rtinK: Sounds good - this is one of the things I was not sure how to do correctly, so thanks a lot for feedback! :)

pyanaconda/ui/tui/spokes/__init__.py
M4rtinK: OK, let's return `None` here as well.

pyanaconda/ui/tui/tuiobject.py
M4rtinK: Done. :)

pyanaconda/ui/common.py
M4rtinK: Dropped. :)

pyanaconda/ui/gui/spokes/custom_storage.py
M4rtinK: Will do - I wonder if the custom partitioning spoke title is ever displayed anywhere given that it is only ever accessed via the Storage spoke and never directly presented on the a hub, where the title string can be seen.

That could explain why this might have been overlooked in the past.

PR #2961, opened 2020-10-29
Fix issue when ns_info cannot be retrieved for NVDimm namespace

jkonecny12:
If we don't skip this part the uncaught exception will raise because we are
trying to concatenate string and None types.

*Resolves: rhbz#1891827*

### DIFF
--- a/pyanaconda/modules/storage/nvdimm/nvdimm.py
+++ b/pyanaconda/modules/storage/nvdimm/nvdimm.py
@@ -101,6 +101,12 @@ class NVDIMMModule(KickstartBaseModule):
         devices_to_ignore = set()
 
         for ns_name, ns_info in nvdimm.namespaces.items():
+            # this is happening when namespace is set to DEVDAX mode - block device is not present
+            if ns_info.blockdev is None:
+                log.debug("%s will be skipped - NVDIMM namespace block device information "
+                          "can't be retrieved", ns_name)
+                continue
+
             info = udev.get_device(device_node="/dev/" + ns_info.blockdev)
 
             if info and udev.device_get_format(info) == "iso9660":
@@ -116,8 +122,7 @@ class NVDIMMModule(KickstartBaseModule):
             else:
                 continue
 
-            if ns_info.blockdev:
-                devices_to_ignore.add(ns_info.blockdev)
+            devices_to_ignore.add(ns_info.blockdev)
 
         return devices_to_ignore
 
--- a/tests/nosetests/pyanaconda_tests/module_nvdimm_test.py
+++ b/tests/nosetests/pyanaconda_tests/module_nvdimm_test.py
@@ -145,6 +145,7 @@ class NVDIMMKickstartTestCase(unittest.TestCase):
                 "namespace0.0": Mock(blockdev="pmem0", mode=blockdev.NVDIMMNamespaceMode.SECTOR),
                 "namespace1.0": Mock(blockdev="pmem1", mode=blockdev.NVDIMMNamespaceMode.SECTOR),
                 "namespace2.0": Mock(blockdev="pmem2", mode=blockdev.NVDIMMNamespaceMode.MEMORY),
+                "namespace3.0": Mock(blockdev=None, mode=blockdev.NVDIMMNamespaceMode.DEVDAX),
             }
 
             ignored_devices = self.nvdimm_module.get_devices_to_ignore()


pyanaconda/modules/storage/nvdimm/nvdimm.py
```diff
@@ -101,6 +101,9 @@ def get_devices_to_ignore(self):
         devices_to_ignore = set()
 
         for ns_name, ns_info in nvdimm.namespaces.items():
+            if ns_info is None and ns_info.blockdev is None:
```
poncovka, line 104: The `ns_info` variable shouldn't be `None`.

pyanaconda/modules/storage/nvdimm/nvdimm.py
```diff
@@ -101,6 +101,9 @@ def get_devices_to_ignore(self):
...
+                continue
```
poncovka, line 105: Please, log here why this namespace is skipped.

pyanaconda/modules/storage/nvdimm/nvdimm.py
poncovka: Please, remove `if ns_info.blockdev:` below.

poncovka: Please, extend the unit tests that call this method. I think it should be enough to define a namespace without a block device in `_check_ignored`.

poncovka (commented):

jkonecny12:
**UPDATED:**
* Fixed notes from @poncovka . Thanks for that!

pyanaconda/modules/storage/nvdimm/nvdimm.py
```diff
@@ -101,6 +101,11 @@ def get_devices_to_ignore(self):
         devices_to_ignore = set()
 
         for ns_name, ns_info in nvdimm.namespaces.items():
+            if ns_info.blockdev is None:
+                log.debug("%s - NVDIMM namespace block device information "
```
poncovka, line 105: I would use `%s will be skipped - ` to follow the pattern of the messages below.

poncovka (approved):
Looks good to me. Thanks!

poncovka (approved):
It looks great!

tests/nosetests/pyanaconda_tests/module_nvdimm_test.py
```diff
@@ -145,6 +145,7 @@ def _check_ignored(self, expected_devices):
                 "namespace0.0": Mock(blockdev="pmem0", mode=blockdev.NVDIMMNamespaceMode.SECTOR),
                 "namespace1.0": Mock(blockdev="pmem1", mode=blockdev.NVDIMMNamespaceMode.SECTOR),
                 "namespace2.0": Mock(blockdev="pmem2", mode=blockdev.NVDIMMNamespaceMode.MEMORY),
+                "namespace3.0": Mock(blockdev=None, mode=blockdev.NVDIMMNamespaceMode.SECTOR),
```
vojtechtrefny, line 148: I know it's just a test, but changing the mode to `blockdev.NVDIMMNamespaceMode.DEVDAX`  would make this more realistic, sector mode namespaces will always have a blockdev.

pyanaconda/modules/storage/nvdimm/nvdimm.py
```diff
@@ -101,6 +101,11 @@ def get_devices_to_ignore(self):
         devices_to_ignore = set()
 
         for ns_name, ns_info in nvdimm.namespaces.items():
+            if ns_info.blockdev is None:
+                log.debug("%s will be skipped - NVDIMM namespace block device information "
```
vojtechtrefny, line 105: You could make this a little more elegant by ignoring all namespaces that are not in devdax or sector mode. You are already ignoring all non-sector mode namespaces in the `elif` below, so it would make the code a little bit more straightforward. (But that's just a nitpick, feel free to ignore me.)

vojtechtrefny (approved):
Looks good to me.

pyanaconda/modules/storage/nvdimm/nvdimm.py
jkonecny12: Good point, however, I'm thinking if I wouldn't prefer this solution with a comment to explain in case there would be other similar cases. It's just more robust. What do you think @poncovka ?

poncovka: I agree. We should still check `blockdev`. Please, add a new comment or update the commit  message.

jkonecny12:
**UPDATED:**
* Update test
* Update commit message
* Add comment

poncovka (approved):
Still looks good to me.

jkonecny12:
Smoke test went fine.

PR #2963, opened 2020-10-29
gui: specify unit in which disk capacity is measured in traditional partitioning module

mikhailnov:
GUI requests to specify size of a new partition, but it was not clear if it is in MiB, GiB etc.

Thanks to Mikhail Mosolov <m.mosolov@rosalinux.ru> for help.

Screenshot:

DeepinScreenshot_выберите-область_20201029174337

### DIFF
--- a/pyanaconda/ui/gui/spokes/lib/custom_storage_helpers.glade
+++ b/pyanaconda/ui/gui/spokes/lib/custom_storage_helpers.glade
@@ -510,6 +510,30 @@ use.  Try something else?</property>
               </packing>
             </child>
             <child>
+              <object class="GtkLabel" id="desired_capacity_units">
+                <property name="visible">True</property>
+                <property name="can_focus">False</property>
+                <property name="label" translatable="yes">MiB</property>
+                <property name="track_visited_links">False</property>
+              </object>
+              <packing>
+                <property name="left_attach">2</property>
+                <property name="top_attach">4</property>
+              </packing>
+            </child>
+            <child>
+              <placeholder/>
+            </child>
+            <child>
+              <placeholder/>
+            </child>
+            <child>
+              <placeholder/>
+            </child>
+            <child>
+              <placeholder/>
+            </child>
+            <child>
               <placeholder/>
             </child>
           </object>


poncovka:
Hi @mikhailnov, thanks for the fix! I agree that it should be clear what is the default unit, but now it looks like we accept only sizes in MiB. I would prefer to use the `placeholder_text` attribute and provide more information about the expected value. It could say, for example: `Enter size and unit (MiB by default).`

jkonecny12:
I agree, it's a great suggestion @poncovka.

mikhailnov:
30.10.2020 12:32, Vendula Poncova пишет:
>
> Hi @mikhailnov <https://github.com/mikhailnov>, thanks for the fix! I agree that it should be clear what is the default unit, but now it looks like we accept only sizes in MiB. I would prefer to use the |placeholder_text| attribute and provide more information about the expected value. It could say, for example: |Enter size and unit (MiB by default).|
>
Do you suggest to make the user and units, e.g. type MiB or GiB? Or do you mean choosing from a list of possible values? In the first case I don't think it is a good idea, in the second one it is probably useful, expecially if someone does not know how many MiBs there are in one GiB

poncovka:
Hi, I suggest to remove the `MiB` label and set `placeholder_text` of the GtkEntry for "Desired Capacity".

> Do you suggest to make the user and units, e.g. type MiB or GiB? 

I don't understand the question.

> Or do you mean choosing from a list of possible values?

That is not necessary. You can already write a number (for example, `200`) or a number with an unit (for example, `2 GiB`). The installer accepts both cases.

VladimirSlavik:
@mikhailnov it is possible to enter the value with units, eg. 500 GiB. The MiB is only a default unit which is used if you type only numbers. So putting "MiB" there statically makes little sense :( Hope that helps understanding why @poncovka disagrees.

Maybe a better change would be to make the user aware of the possibility to write units, and that MiB is the default if omitted. The suggestions so far go in that direction. (Although I would add there could be more ways to do it besides placeholder text...)

mikhailnov:
>  it is possible to enter the value with units, eg. 500 GiB

Wow :-) How should users guess about it?)) Yes, it makes @poncovka 's thoughts clearer, thanks for explaining. And what is the difference between Gb (1000 MB), GiB (1024 MB), GB (== GiB?), Гб, ГБ, Гигабайт, Гигабит, Гибибайт, gigabyte, gibybyte etc? It looks like a good experience for a CLI, but not for GUI, I believe that the GUI must be as straighforward and simple as possible.

> Hi, I suggest to remove the MiB label and set placeholder_text of the GtkEntry for "Desired Capacity".

I thought about a placeholder before making this pull request, but the text of the placeholder disappers once typing is strated, it is not easy to remember what was written there, I believe that such information must be in front of the eyes all the time.

> (Although I would add there could be more ways to do it besides placeholder text...)

I would make a drop down list to choose units, probably. But then "GiB" etc. inside the entered text must not be parsed and must be treated as an invalid input.

VladimirSlavik:
I am not an expert, but there's a lot of space - maybe adding a paragraph about the units could be put between the input field and buttons? That's just an idea, I am not really sure about that one...

PS: You can search for `DESIRED_CAPACITY_ERROR` definition and usages for some related stuff.

mikhailnov:
02.11.2020 12:00, Vladimír Slávik пишет:
>
> I am not an expert, but there's a lot of space - maybe adding a paragraph about the units could be put between the input field and buttons? That's just an idea, I am not really sure about that one...
>
Most people won't read big texts carefully and try to understand them unfortunately...

mikhailnov:
In other words, my suggestion is to move to making UX more straightforward. To my mind disk partitioning is the worst pice of Anaconda, it is non-intuituive at all. Blivet is better, but it lacks good automatic partitioning for non-experienced users.

snbueno:
> Maybe a better change would be to make the user aware of the possibility to write units, and that MiB is the default if omitted. The suggestions so far go in that direction. (Although I would add there could be more ways to do it besides placeholder text...)

I think this is a good compromise that @VladimirSlavik suggested, and the suggested text from @poncovka works, `Enter size and unit (MiB by default)`. This is not too verbose, and so people will read it.

@mikhailnov, sorry for the delay in responding to you here. The UX could do with being more straightforward, absolutely. But we must strike a balance between what is supportable/manageable from a maintenance perspective as well. What do you think of this?

mikhailnov:
>
>@mikhailnov, sorry for the delay in responding to you here. The UX
>could do with being more straightforward, absolutely. But we must
>strike a balance between what is supportable/manageable from a
>maintenance perspective as well. What do you think of this?

What do you mean under maintenance? I think the more straightforward the UI is, the easier maintenance is. Maintaining chosing if units from a drop down list should be easier than parsing manual input, but parsing will be still supported for TUI and kickstarts...

VladimirSlavik:
Actually, the current solution requires the least maintenance. There is the 3rd party (?) Size library that does the work. In terms of code, we just do the same thing handling inputs from TUI, GUI, Kickstart, config files... take the string and give it to the library. And then you also overlook that this exists and works right now, and has been that way for years, so the bugs have been solved. So the "inertia" is very strong.

You wrote that a dropdown is less maintenance. But that would need new, additional, unique code for actually combining the number from text input and selected unit. This would be very little, but still. And then there is the big question, what should happen when users type not just number but also unit as they did before. We actually had bug reports like that before, it's not a hypothetical problem. But then the code must be more complicated, to somehow handle both kinds of input. So in my mind, a "correct" solution would have to do three things: 1) When typing, check if the input has an unit, if not, use unit from dropdown, and if yes, change the dropdown. 2) Changing dropdown would erase the unit from text, if present. 3) All changes must be internally combined to the correct representation. Now, that is a lot more than just the dropdown... And now repeat it in all the places where size can be provided.

poncovka:
I have discussed your suggestion with the team and we would prefer to use a less invasive solution. It is implemented in the pull request https://github.com/rhinstaller/anaconda/pull/3066. Based on that, I am closing this pull request. Thanks for bringing the issue out!

mikhailnov:
Great, thanks!

PR #2964, opened 2020-10-29
workflows: Add manual triggering of tests

martinpitt:
Run unit tests in a PR triggered by a "/test [...]" comment from an
organization member.  This avoids running untrusted and unreviewed code
on self-hosted runners.

For now this only applies for PRs against the rhel-8 branch, but an
`issue_comment` trigger needs to be on master. (In the future we *could*
support running rhel/ELN tests on master as well, though.)

------

This will go along with dropping the `pull_request` action from the rhel-8 branch. I'll send a PR for that and link it from here.

I tested this pretty thoroughly [on my fork](https://github.com/martinpitt/anaconda/pull/4) and a local github action runner. The "unit-tests" check is "the real thing" (on self-hosted runner), the "demo test" was just some `sleep 10` thing for quickly iterating the workflow mechanics.

### DIFF
--- a/.github/workflows/validate-rhel-8.yml
+++ b/.github/workflows/validate-rhel-8.yml
@@ -1,0 +1,76 @@
+# Run rhel-8 unit tests in a PR triggered by a "/test [...]" comment from an organization member.
+# This avoids running untrusted and unreviewed code on self-hosted runners.
+name: unit-tests-rhel-8
+on:
+  issue_comment:
+    types: [created]
+
+jobs:
+  pr-info:
+    # restrict running of tests to trusted organization members
+    # see https://developer.github.com/v4/enum/commentauthorassociation/
+    if: startsWith(github.event.comment.body, '/test') && contains('OWNER MEMBER', github.event.comment.author_association)
+    runs-on: ubuntu-latest
+    steps:
+      - name: Get information for pull request
+        uses: octokit/request-action@v2.x
+        id: pr_api
+        with:
+          route: GET /repos/${{ github.repository }}/pulls/${{ github.event.issue.number }}
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+    outputs:
+      base_ref: ${{ fromJson(steps.pr_api.outputs.data).base.ref }}
+      sha: ${{ fromJson(steps.pr_api.outputs.data).head.sha }}
+
+  unit-tests:
+    needs: pr-info
+    if: needs.pr-info.outputs.base_ref == 'rhel-8'
+    runs-on: [self-hosted, ci-tasks, rhel-8]
+    steps:
+      # we post statuses manually as this does not run from a pull_request event
+      # https://developer.github.com/v3/repos/statuses/#create-a-status
+      - name: Create in-progress status
+        uses: octokit/request-action@v2.x
+        with:
+          route: 'POST /repos/${{ github.repository }}/statuses/${{ needs.pr-info.outputs.sha }}'
+          context: unit-tests
+          state: pending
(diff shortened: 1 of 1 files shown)


martinpitt:
I tested the success, failure, and "someone outside the org tries the comment" cases (in which case you just get a [skipped action](https://github.com/martinpitt/anaconda/runs/1327538906?check_suite_focus=true).

.github/workflows/validate-rhel-8.yml
```diff
@@ -0,0 +1,76 @@
...
+    steps:
+      - name: Get information for pull request
+        uses: octokit/request-action@v2.x
+        id: pr_info
+        with:
+          route: GET /repos/${{ github.repository }}/pulls/${{ github.event.issue.number }}
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      #- name: Dump PR info
```
jkonecny12, line 23: Before merge please remove these comments.

.github/workflows/validate-rhel-8.yml
```diff
@@ -0,0 +1,76 @@
...
+      #  run: |
+      #    echo "${{ steps.pr_info.outputs.data }}"
+
+      # the following steps only run for PRs against rhel-8 branch
+
+      # we post statuses manually as this does not run from a pull_request event
+      # https://developer.github.com/v3/repos/statuses/#create-a-status
+      - name: Create in-progress status
+        if: fromJson(steps.pr_info.outputs.data).base.ref == 'rhel-8'
```
jkonecny12, line 32: I'm thinking if having another job 'needing' this one would improve the state of having the ``if rhel-8`` everywhere. But maybe that's just unnecessary complication?

jkonecny12 (changes requested):
Looks great, just please remove the commented code first.

Thanks!

.github/workflows/validate-rhel-8.yml
martinpitt: I thought about it, but then indeed it would be slower at least, as that second job would need an entirely fresh VM instance. It would be more elegant, and indeed I don't particularly like the repetition. I thought we could use that approach for now, and fine-tune/optimize it later; but with this I arrived at a solution that works, and I had the impression that you'd rather land this sooner than later.

In other words, I have no strong opinion.

.github/workflows/validate-rhel-8.yml
martinpitt: Ack, done. I thought I'd leave it in, as it's really useful for debugging/development (and it took me > 1 minute to figure this out).

martinpitt:
@jkonecny12 : I learned about job dependencies and outputs now, and updated this PR accordingly. The workflow [became larger](https://github.com/rhinstaller/anaconda/compare/5b9959f207cf39fc36e500e9a4eeb58ae533e654..ed91510ec5bb80c41d05710191bcfe40dae3d419) now, but it's more elegant and easier to re-use.

I tested this again [on my fork](https://github.com/martinpitt/anaconda/pull/4), and also made sure that [everything gets skipped for unauthorized users](https://github.com/martinpitt/anaconda/actions/runs/337287331).

WDYT?

.github/workflows/validate-rhel-8.yml
martinpitt: Done, see below. Let me know which one you prefer.

jkonecny12 (approved):
I really like this solution. Great work!

martinpitt:
Oops, I forgot to take out the "Dump info" part again, development leftover. I force-pushed to take that out again, and rebased to current master. @jkonecny12 , can you please re-approve?

VladimirSlavik (approved):
Looks good to me. Thank you!

PR #2966, opened 2020-10-30
Fix and robustify validate workflow

martinpitt:
- Use `HEAD` instead of `${{ github.sha }}`. The `checkout` step
   fetches the latter, so HEAD will point to that. This makes the shell
   commands easier to reproduce locally, and also easier to read.

 - `git diff HEAD` is a no-op, as it will compare HEAD against itself.
   Thus the container never built even for PRs that changed it. To fix
   this, diff current origin/master against the PR, so that (1) it does
   anything at all, and (2) also checks all commits in the PR instead of
   just the topmost one.

 - Use an output variable instead of an exit code for the
   `check-dockerfile-changed` step, so that we don't hide syntax/command
   errors in the shell script. These would lead to silently not
   refreshing the container when it should be.

 - Don't use `continue-on-error` in the main test run, but instead
   always run the log upload (even on failure). This is more robust, and
   avoids the need for the explicit "Set job result" step.

### DIFF
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -14,24 +14,24 @@ jobs:
       - name: Check if container changed in this PR
         id: check-dockerfile-changed
         run: |
-          # the SHA is target branch thanks to pull_request_target trigger
-          # add prints for debugging purposes
-          git log -1 --oneline ${{ github.sha }}
-          git diff --name-only ${{ github.sha }}
-          git diff --name-only ${{ github.sha }} | grep -q dockerfile/anaconda-ci/ || exit 2
-        continue-on-error: true
+          changes=$(git diff origin/master..HEAD -- dockerfile/anaconda-ci/)
+          # print for debugging
+          echo "$changes"
+          [ -z "$changes" ] || echo "::set-output name=changed::true"
 
       # build container if files for dockerfile changed in the PR
       - name: Build anaconda-ci container
-        if: ${{ steps.check-dockerfile-changed.outcome == 'success' }}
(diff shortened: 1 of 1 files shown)


martinpitt:
I tested both cases on my fork:

 - [This  run](https://github.com/martinpitt/anaconda/runs/1331095572?check_suite_focus=true) is exactly the version that is pushed here. The container build is skipped.

 - [This run](https://github.com/martinpitt/anaconda/runs/1331093780?check_suite_focus=true) has an extra commit that changes the container, and the container build runs.

.github/workflows/validate.yml
```diff
@@ -14,24 +14,24 @@ jobs:
       - name: Check if container changed in this PR
         id: check-dockerfile-changed
         run: |
-          # the SHA is target branch thanks to pull_request_target trigger
-          # add prints for debugging purposes
-          git log -1 --oneline ${{ github.sha }}
-          git diff --name-only ${{ github.sha }}
-          git diff --name-only ${{ github.sha }} | grep -q dockerfile/anaconda-ci/ || exit 2
```
jkonecny12, line 21: Just a note this is not `HEAD` as you are writing in the commit message but master because of `pull_request_target`. So it should work as expected, I've tested that. However, I like your solution more so no problem with changing this.

jkonecny12 (approved):
Looks good to me otherwise.

.github/workflows/validate.yml
martinpitt: Ah right, `github.event.pull_request.head.sha != github.sha`. Then the previous code makes more sense indeed. Sorry for the mix-up.

PR #2967, opened 2020-10-30
Don't clobber host checkout on `make container-ci`

martinpitt:
Copy the work tree into the container's file system, so that autotools,
built/pyc files etc. don't clobber the host checkout with files (which
are not even owned by the user, so are a bit difficult to delete).

Just copy out the logs to the host's ./test-logs/*.log directory. Adjust
the validate workflow accordingly.

 - [x] Builds on top of PR #2966 (for proper validation)


martinpitt:
I played around with various other options, such as `cp -rl` (does not work, as hardlinks can't span across mounts), `cp -rs` (that by itself succeeds, but our build system doesn't get along with a symlink farm), and in general I don't like a `cp -r` tree for interactive debugging -- it's too brittle, like it doesn't see new or removed files, and it's all too easy to replace a symlink with a diverging file copy.

So I think this should address @poncovka 's woes about not clobbering the host's checkout, while keeping the behaviour for the interactive shell (but that's a different style of working, and in that case I believe you really want to work directly on the bind mount -- and `git clean -fdx` is your friend!)

WDYT?

martinpitt:
/packit build

Hmm, not sure about that [packit failure](https://prod.packit.dev/srpm-build/9216/logs) -- this PR does not change anything that's relevant for the RPM build or runtime, or anything that this test touches. I'll retry it.

poncovka (approved):
I have played with the new version and it works as expected. Thank you!

dockerfile/anaconda-ci/run-build-and-arg
```diff
@@ -0,0 +1,22 @@
...
+./autogen.sh
+./configure
+make
+
+# always copy the results back to the host
+copy_logs() {
+    mkdir -p /anaconda/test-logs/
+    # clean up from previous run
+    rm -rf /anaconda/test-logs/*
+    find -name '*.log' -exec cp '{}' /anaconda/test-logs/ \;
```
poncovka, line 17: Btw. this will find logs in `/tmp/anaconda/test-logs` that will override everything else, so I never get a new version of the logs.

martinpitt: Sorry, I'm afraid I don't understand what you mean. Can you give a step-by-step description how this goes wrong?

My intention was that each run cleans up the host directory first, to avoid having outdated logs. Otherwise you could have an early failure, and still find the previous run's nosetests.log or so in the log directory, and think "all is well".

poncovka: How to reproduce?

1. Clone the repository.
2. Create an error in the code.
3. Run unit tests for the first time. This step will create the `test-logs` directory on my host.
4. Remove the error from the code.
5. Run unit tests for the second time. The stdout shows that the tests passed, but the `test-logs` directory contains logs from the first run.

What happened in the step 5?

1. The existing `test-logs` directory with logs from the first run is copied to `/tmp/anaconda`.
2. Tests are run for the second time. The `/tmp/anaconda` directory contains new logs, but `/tmp/anaconda/test-logs` still contains logs from the first run.
3. The files in `/anaconda/test-logs/` are removed, but the files in `/tmp/anaconda/test-logs/` still exists.
4. The command `find` copies logs from `/tmp/anaconda` to `/anaconda/test-logs/`. The files are processed in the alphanumeric order, so `/tmp/anaconda/tests` is processed before `/tmp/anaconda/test-logs`.
5. Files copied from `/tmp/anaconda/test-logs` overwrites files copied from `/tmp/anaconda/tests`.
6. The `/anaconda/test-logs/` directory contains logs from the first run.

poncovka (changes requested):
The existing `test-logs` directory causes problems.

dockerfile/anaconda-ci/run-build-and-arg
martinpitt: Should be fixed now. PTAL?

Makefile.am
```diff
@@ -234,7 +234,10 @@ ci:
 	fi
 
 container-ci:
-	$(CONTAINER_ENGINE) run $(CONTAINER_TEST_ARGS) $(CI_TEST_ARGS) $(CI_NAME):$(CI_TAG) $(CI_CMD)
+	$(CONTAINER_ENGINE) run --entrypoint /anaconda/dockerfile/anaconda-ci/run-build-and-arg $(CONTAINER_TEST_ARGS) $(CI_TEST_ARGS) $(CI_NAME):$(CI_TAG) $(CI_CMD)
```
jkonecny12, line 237: This is pretty smart. Do not copy the script in the container but use the mount already present. Interesting.

jkonecny12 (approved):
LGTM! Great improvement!

dockerfile/anaconda-ci/run-build-and-arg
poncovka: It is fixed. Thanks!

poncovka (approved):
Looks good to me.

VladimirSlavik:
I'm eagerly awaiting this, as I don't mind the files too much, but would be far happier w/o these too :-)

martinpitt:
I'll send a backport to rhel-8 tomorrow.

PR #2969, opened 2020-11-01
Do not check ssl certificate if --noverifyssl is in kickstart (#1745064)

jstodola:
This PR fixes two problems with checking of ssl certificates:
1. Downloading of metadata ignored --noverifyssl in the kickstart file.
2. Additional repositories loaded from treeinfo were always checking ssl certificate. This is a problem for RHEL-8 and unified DVD.

Resolves: rhbz#1745064

### DIFF
--- a/pyanaconda/payload/dnf/payload.py
+++ b/pyanaconda/payload/dnf/payload.py
@@ -1451,7 +1451,7 @@ class DNFPayload(Payload):
                 base_repo_url = self._get_base_repo_location(install_tree_url)
                 log.debug("releasever from %s is %s", base_repo_url, self._base.conf.releasever)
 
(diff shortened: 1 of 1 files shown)


poncovka (commented):
Thanks for the fix! It looks good to me, but @jkonecny12 should see it as well.

jkonecny12:
jenkins, test this please

jkonecny12 (commented):
Looks great but it needs to be tested properly first.

jkonecny12 (approved):
This is really hell to test it on Rawhide but finally. I've manage to do that. Works as expected, without this change the Anaconda will use repository as installation root and ignore other ``.treeinfo`` repositories and with this change it will correctly load ``.treeinfo`` and all repositories in it.

Thanks for this fix @jstodola!

PR #2970, opened 2020-11-03
Don't clobber host checkout on `make container-ci`

martinpitt:
Backport of #2967 . Still draft, I am still validating this locally, and it also needs a corresponding PR to master to adjust grabbing the logs.

 - [x] Adjust validate-rhel-8.yml workflow for changed log paths; land this in lockstep with PR #2971


martinpitt:
Ugh, seems the `/test` magic doesn't work here, it just [gets skipped](https://github.com/rhinstaller/anaconda/runs/1351401417?check_suite_focus=true). That will need some investigation. It did work against my fork, but I didn't use an actual org there.

martinpitt:
I tested this in my personal test org now, and [figured out what is wrong](https://github.com/mpitt-test-org/anaconda/pull/1#issuecomment-721598987). TL/DR: The rhinstaller org has mostly private memberships, so the "association" is never OWNER or MEMBER for these, just CONTRIBUTOR. But the latter would be true for people who are outside of the rhinstaller org, so that's no use to us.

The easy solution would be to make our [memberships](https://github.com/orgs/rhinstaller/people) public, and that's maybe also more useful for the open source spirit. The other solution would be to create a github token with `read:org` permission, and replace the declarative `if` with a code check for membership that uses this token.

@jkonecny12, @rvykydal , @poncovka , @VladimirSlavik : Opinions? Would you rather keep your membership in rhinstaller private? (It's not *really* private -- the whole git log and contributor history is all public, so you can't really hide the fact that you work on anaconda :grin: )

poncovka (approved):
Looks good to me.

PR #2972, opened 2020-11-04
Fix user permission checking for rhel-8 validate task

jkonecny12:
It does not work correctly for users with private visibility on the organization. However, this is problem because the private visibility is default on GitHub and it can't be changed globally.

Use GH API to check user permissions on the repository. This does not require special privileges on token. Allow only admin or write access to start the tests.

This was tested on my test organization with help of our bot. Should work as expected.
https://github.com/Test-anaconda-org/anaconda/actions?query=workflow%3Aunit-tests-rhel-8
Last failure is because I removed the bot from the organization.

### DIFF
--- a/.github/workflows/validate-rhel-8.yml
+++ b/.github/workflows/validate-rhel-8.yml
@@ -7,11 +7,25 @@ on:
 
 jobs:
   pr-info:
-    # restrict running of tests to trusted organization members
-    # see https://developer.github.com/v4/enum/commentauthorassociation/
-    if: startsWith(github.event.comment.body, '/test') && contains('OWNER MEMBER', github.event.comment.author_association)
+    if: startsWith(github.event.comment.body, '/test')
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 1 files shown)


martinpitt (approved):
That is a very elegant approach, thank you!

It's a bit of a wart that the workflow will fail instead of being skipped -- this will generate error emails to everyone who is subscribed to the project.

This could be refined further by setting a job result, and `if`'ing that in the `unit-tests:` job. But that cleanup doesn't need to hold up fixing the PR workflow, so :+1: from me.

jkonecny12:
> That is a very elegant approach, thank you!
> 
> It's a bit of a wart that the workflow will fail instead of being skipped -- this will generate error emails to everyone who is subscribed to the project.
> 
> This could be refined further by setting a job result, and `if`'ing that in the `unit-tests:` job. But that cleanup doesn't need to hold up fixing the PR workflow, so +1 from me.

Great idea, I'll create follow up PR.

VladimirSlavik (approved):

PR #2973, opened 2020-11-04
Select disks for implicit partitions (#1642391)

poncovka:
Don't use disks that will be used for allocation of requested partitions
(for example, the /boot partition). Calculate the number of slots that
will be used by requested partitions and skip disks that will not have
free slots for implicit partitions.

Otherwise, we might schedule implicit partitions, that wouldn't be possible
to allocate on the specified disk, and the partitioning might fail.

Resolves: rhbz#1642391

### DIFF
--- a/pyanaconda/modules/storage/partitioning/automatic/utils.py
+++ b/pyanaconda/modules/storage/partitioning/automatic/utils.py
@@ -167,6 +167,45 @@ def get_candidate_disks(storage):
     return free_disks
 
 
+def get_disks_for_implicit_partitions(disks, scheme, requests):
+    """Return a list of disks that can be used for implicit partitions.
+
+    :param disks: a list of candidate disks
+    :param scheme: a type of the partitioning scheme
+    :param requests: a list of partitioning requests
+    :return: a list of disks that can be used for implicit partitions
+    """
+    # There will be no implicit partitions.
+    if scheme == AUTOPART_TYPE_PLAIN:
+        return []
+
+    # Calculate slots for requested partitions.
+    requested_slots = 0
+
+    for request in requests:
+        if request.is_partition(scheme):
+            requested_slots += 1
+
+    # Collect extra disks for implicit partitions.
+    extra_disks = []
+
+    for disk in disks:
+        parted_disk = disk.format.parted_disk
+        supports_extended = parted_disk.supportsFeature(parted.DISK_TYPE_EXTENDED)
+        available_slots = parted_disk.maxPrimaryPartitionCount - parted_disk.primaryPartitionCount
+
+        # Skip disks that will be used for requested partitions.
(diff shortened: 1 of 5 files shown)


martinpitt:
@poncovka , @jkonecny12 : FYI, I just marked "unit-tests" as a required test, so that we don't have a PR where we forget to run it. It's too easy to just see the green Jenkins rpm-tests and forget to trigger the uni-tests one.

poncovka:
@vojtechtrefny, could you look at this pull request, please?

VladimirSlavik (approved):
Looks good to me, thank you!

I wonder about the commit messages though. Basically this seems to be: "Consider not just free space but also if partitions can be created, because there can be only 4 standard partitions." These words are nowhere to be seen in the commit, bug, anywhere. It's not critical but if I saw the commit three years later, the description will not help me unless I actually read a specific part of the changes.

poncovka:
> I wonder about the commit messages though. Basically this seems to be: "Consider not just free space but also if partitions can be created, because there can be only 4 standard partitions." These words are nowhere to be seen in the commit, bug, anywhere. It's not critical but if I saw the commit three years later, the description will not help me unless I actually read a specific part of the changes.

Unfortunately, the situation is much more complicated. There are primary partitions and extended partitions. When the disk supports extended partitions, then you are not limited by the number of slots. And the disk doesn't have to support exactly 4 primary partitions. Therefore, I cannot use anything from the sentence `because there can be only 4 standard partitions`.

VladimirSlavik:
Thanks, makes sense.

pyanaconda/modules/storage/partitioning/automatic/utils.py
```diff
@@ -167,6 +167,45 @@ def get_candidate_disks(storage):
...
+        supports_extended = parted_disk.supportsFeature(parted.DISK_TYPE_EXTENDED)
+        available_slots = parted_disk.maxPrimaryPartitionCount - parted_disk.primaryPartitionCount
+
+        # Skip disks that will be used for requested partitions.
+        if requested_slots and not supports_extended and available_slots <= requested_slots:
+            requested_slots -= available_slots
+            log.debug("Don't use %s for implicit partitions.", disk.name)
+        else:
+            requested_slots = 0
+            extra_disks.append(disk)
```
jkonecny12, line 203: Can't help myself but I think this should live in blivet. These are exactly the calculation which we should avoid in Anaconda.

jkonecny12 (approved):
Looks good to me otherwise.

pyanaconda/modules/storage/partitioning/automatic/utils.py
poncovka: @vojtechtrefny, what do you think?

poncovka:
Ported in https://github.com/rhinstaller/anaconda/pull/3093.

PR #2974, opened 2020-11-04
Create an abstraction of the DNF base

poncovka:
Use the DNF manager to work with the DNF base.

### DIFF
--- a/pyanaconda/core/util.py
+++ b/pyanaconda/core/util.py
@@ -46,7 +46,6 @@ from pyanaconda.core.process_watchers import WatchProcesses
 from pyanaconda.core.constants import DRACUT_SHUTDOWN_EJECT, TRANSLATIONS_UPDATE_DIR, \
     IPMI_ABORTED, X_TIMEOUT, TAINT_HARDWARE_UNSUPPORTED, TAINT_SUPPORT_REMOVED, \
     WARNING_HARDWARE_UNSUPPORTED, WARNING_SUPPORT_REMOVED
-from pyanaconda.core.regexes import OS_RELEASE_OS_VERSION
 from pyanaconda.errors import RemovedModuleError, ExitError
 
 from pyanaconda.anaconda_logging import program_log_lock
@@ -1476,20 +1475,36 @@ class LazyObject(object):
         return setattr(self._object, name, value)
 
 
-def get_os_version(sysroot=""):
-    """Find version of the OS from the os-release file.
+def get_os_release_value(name, sysroot=""):
+    """Read os-release files and return a value of the specified parameter.
 
-    See os-release(5).
-
-    :param str sysroot: Where to look.
-    :return str: The version
+    :param name: a name of the parameter (for example, "VERSION_ID")
+    :param sysroot: a path to the system root
+    :return: a string with the value of None if nothing found
     """
-    for filename in ("/etc/os-release", "/usr/lib/os-release"):
+    # Match the variable assignment (for example, "VERSION_ID=").
+    name += "="
+
+    # Search all os-release files in the system root.
+    paths = ("/etc/os-release", "/usr/lib/os-release")
+
+    for path in paths:
         try:
-            with open(sysroot + filename, "r") as f:
-                data = f.read(4096)  # 4 kB should be enough for everyone!
-            return OS_RELEASE_OS_VERSION.findall(data)[0]
-        except (FileNotFoundError, IndexError):
+            with open(join_paths(sysroot, path), "r") as f:
+                for line in f:
+                    # Match the current line.
+                    if not line.startswith(name):
+                        continue
+
+                    # Get the value.
+                    value = line[len(name):]
+
+                    # Strip spaces and then quotes.
+                    value = value.strip().strip("\"'")
+                    return value
+        except FileNotFoundError:
             pass
 
+    # No value found.
+    log.debug("%s not found in os-release files")
     return None
--- a/pyanaconda/modules/payloads/payload/dnf/dnf_manager.py
+++ b/pyanaconda/modules/payloads/payload/dnf/dnf_manager.py
@@ -26,6 +26,7 @@ from pyanaconda.core.payload import ProxyString, ProxyStringError
 from pyanaconda.core.util import get_os_release_value
 from pyanaconda.modules.payloads.constants import DNF_REPO_DIRS
 from pyanaconda.modules.payloads.payload.dnf.utils import get_product_release_version
+from pykickstart.constants import KS_BROKEN_IGNORE
 
 log = get_module_logger(__name__)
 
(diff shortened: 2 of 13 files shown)


pyanaconda/core/util.py
```diff
@@ -1496,20 +1495,35 @@ def __setattr__(self, name, value):
...
+                    # Get the value.
+                    value = line[len(name):]
+
+                    # Strip spaces and then quotes.
+                    value = value.strip().strip("\"'")
+                    return value
+        except FileNotFoundError:
             pass
 
+    # No value found.
```
jkonecny12, line 1528: Maybe adding debug logging could help in some situations.

tests/nosetests/pyanaconda_tests/module_payload_dnf_manager_test.py
```diff
@@ -152,3 +152,14 @@ def configure_base_test(self):
             "strict = 0",
             "install_weak_deps = 0",
         )
+
+    def dump_configuration_test(self):
+        """Test the dump of the DNF configuration."""
+        with self.assertLogs(level="DEBUG") as cm:
+            self.dnf_manager.dump_configuration()
```
jkonecny12, line 160: Again, I don't think we want to call ``dump()`` DNF function, it could fail based on environment. As example Fedora Silverblue where no DNF is installed.

tests/nosetests/pyanaconda_tests/module_payload_dnf_manager_test.py
```diff
@@ -163,3 +164,24 @@ def dump_configuration_test(self):
...
+        # Fake transaction.
+        tsi_1 = Mock()
+        tsi_1.pkg.installsize = 1024 * 100
+        tsi_1.pkg.files = ["/file"] * 10
+
+        tsi_2 = Mock()
+        tsi_2.pkg.installsize = 1024 * 200
+        tsi_2.pkg.files = ["/file"] * 20
+
+        self.dnf_manager._base.transaction = [tsi_1, tsi_2]
```
jkonecny12, line 183: This code basically testing DNF. We don't want to test DNF code just that we are calling it correctly.

tests/nosetests/pyanaconda_tests/module_payload_dnf_manager_test.py
```diff
@@ -185,3 +185,21 @@ def get_installation_size_test(self):
...
+        self.assertEqual(size, Size(0))
+
+        # Fake transaction.
+        tsi_1 = Mock()
+        tsi_1.pkg.downloadsize = 1024 * 1024 * 100
+
+        tsi_2 = Mock()
+        tsi_2.pkg.downloadsize = 1024 * 1024 * 200
+
+        self.dnf_manager._base.transaction = [tsi_1, tsi_2]
```
jkonecny12, line 202: The same here.

pyanaconda/modules/payloads/payload/dnf/dnf_manager.py
```diff
@@ -191,6 +191,27 @@ def get_installation_size(self):
...
+    def get_download_size(self):
+        """Calculate the download size.
+
+        :return: a space required for packages
+        :rtype: an instance of Size
+        """
+        download_size = Size(0)
+
+        if self._base.transaction is None:
+            return Size(0)
```
jkonecny12, line 203: Moving this check to the top of the function would be better.

jkonecny12 (changes requested):
This looks really great! Thanks for that.

However, I think we should look on how the code is tested. I don't think we want to test DNF code.

pyanaconda/core/util.py
poncovka: Fixed.

tests/nosetests/pyanaconda_tests/module_payload_dnf_manager_test.py
```diff
@@ -0,0 +1,91 @@
...
+# License and may only be used or replicated with the express permission of
+# Red Hat, Inc.
+#
+import unittest
+from unittest.mock import patch
+
+from pyanaconda.modules.payloads.payload.dnf.dnf_manager import DNFManager
+
+
+class DNFMangerTestCase(unittest.TestCase):
```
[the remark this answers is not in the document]
poncovka: I don't want to mock `read_comps`. It seems to work fine and the only reason we call it during the initialization is to set up `base.comps`. Otherwise, it would be `None`. Anyway, I plan to remove it from the initialization and handle `None`, but it will take a while.

Also, I am not going to use a mock instead of DNF, because that doesn't test anything. We don't mock Blivet, just the necessary parts. I don't see a difference. This solution already helped me to discover some bugs in our code.

tests/nosetests/pyanaconda_tests/module_payload_dnf_manager_test.py
poncovka: I don't understand the comment. We don't run unit tests on Fedora Silverblue.

tests/nosetests/pyanaconda_tests/module_payload_dnf_manager_test.py
poncovka: Unfortunately, this is testing our code.

```
    def get_installation_size(self):
        """Calculate the installation size.

        :return: a space required by packages
        :rtype: an instance of Size
        """
        packages_size = Size(0)
        files_number = 0

        if self._base.transaction is None:
            return Size("3000 MiB")

        for tsi in self._base.transaction:
            # Space taken by all files installed by the packages.
            packages_size += tsi.pkg.installsize
            # Number of files installed on the system.
            files_number += len(tsi.pkg.files)

        log.debug("Space required for packages: %s", packages_size)

        # Calculate the files size depending on number of files.
        files_size = Size(files_number * DNF_EXTRA_SIZE_PER_FILE)
        log.debug("Space required for installed files: %s", files_size)

        # Get the total size. Add another 10% as safeguard.
        total_space = Size((packages_size + files_size) * 1.1)
        log.debug("Total required size: %s", total_space)

```

tests/nosetests/pyanaconda_tests/module_payload_dnf_manager_test.py
poncovka: Same as above.

pyanaconda/modules/payloads/payload/dnf/dnf_manager.py
poncovka: Fixed.

pyanaconda/core/constants.py
```diff
@@ -475,6 +475,14 @@ class DisplayModes(Enum):
 DNF_DEFAULT_TIMEOUT = -1
 DNF_DEFAULT_RETRIES = -1
 
+# Locations of repo files.
+DNF_REPO_DIRS = [
+    '/etc/yum.repos.d',
+    '/etc/anaconda.repos.d',
+    '/tmp/updates/anaconda.repos.d',
+    '/tmp/product/anaconda.repos.d'
+]
```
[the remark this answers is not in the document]
poncovka: Fixed.

jkonecny12 (approved):
Looks good to me. I still don't think it's a good idea to test other project in our unit tests code but true is that we already decided this elsewhere so I won't be blocking this PR because of that.

PR #2975, opened 2020-11-05
Do not fail user check job in rhel-8 validate workflow

jkonecny12:
The check job fail will generate mails and it is really not a fail. Let's instead set output which will be checked on unit-tests job.

As you can see [here](https://github.com/Test-anaconda-org/anaconda/actions/runs/345978664) the job for unauthorized user is succeeded but unit-test job is not started.


martinpitt (approved):
Niiiice! :+1:

PR #2978, opened 2020-11-05
Remove autostep functionality

VladimirSlavik:
Autostep did not work for some time already, this merely removes the no longer used code. Automatic screenshots are gone with this, too. Manual screenshots remain available.

----

This was so much code, it was expensive to write... and now...

Throw away


poncovka (approved):
It looks great. Thanks!

M4rtinK (approved):
Looks good to me as well - a really nice cleanup! :)

jkonecny12:
I love commits which removes more than add :)!

PR #2980, opened 2020-11-06
Master nosetests directories 3

VladimirSlavik:
Another episode of the Organizing Tests series. Let's rename stuff so we don't repeat the `module_xxx_` bit?

(Trying to avoid touching the high traffic bits so far.)


M4rtinK (approved):
Looks good to me, thanks! :)

poncovka (approved):
Looks good to me.

PR #2982, opened 2020-11-06
Fix UID and GID handling in GUI and D-Bus

VladimirSlavik:
Fixes the range for UID and GIDs in GUI input element, D-Bus transmission, and changes how it is transmitted to prevent future clashes.

Port of #2906 that fixes [rhbz#1786680](https://bugzilla.redhat.com/show_bug.cgi?id=1786680).


poncovka (approved):
Looks good to me.

VladimirSlavik:
Manual testing OK. Crashing with kickstart-supplied *ID over 2**32-1 is still possible; that should be eventually fixed with https://github.com/pykickstart/pykickstart/pull/346

PR #2983, opened 2020-11-07
gui: updating current_hostname lable when clicking apply button in network spoke

Gel0bmstu:
When you click on the "Apply" button the current_hostname label does't change, while the hostname itself changes. This may mislead users.

### DIFF
--- a/pyanaconda/ui/gui/spokes/network.py
+++ b/pyanaconda/ui/gui/spokes/network.py
@@ -663,6 +663,7 @@ class NetworkControlBox(GObject.GObject):
             con.delete()
 
     def on_apply_hostname(self, *args):
+        self.label_current_hostname.set_text(self.hostname)
         self.emit("apply-hostname")
 
     def add_device(self, ty):


poncovka:
@rvykydal, could you look at this pull request, please?

rvykydal:
This (not updating the value of the current hostname) is the case only for localhost.localdomain (or perhaps localhost) values, right? This behaviour is expected (compare with hostnamectl: https://www.freedesktop.org/software/systemd/man/hostnamectl.html) - in this case the current (transient) hostname is "is a fallback value received from network configuration". You can check that the value displayed as the "Current host name" is the same as value of transient hostname returned by hostnamectl command.

Also the patch would probably not work as expected because the value set in on_apply_hostname would be replaced value obtained receiving signal about change of current hostname by network module (_hostname_changed callback).

pyanaconda/ui/gui/spokes/network.py
```diff
@@ -663,6 +663,7 @@ def on_remove_device_clicked(self, *args):
             con.delete()
 
     def on_apply_hostname(self, *args):
+        self.label_current_hostname = self.hostname
```
rvykydal, line 666: This makes Anaconda crashing, the label would need to be set by something like:
```
self.label_current_hostname.set_text(self.hostname)
```

Gel0bmstu:
oh, sorry, it seemed to me that current_hostnme is some kind of the current state of static_hostname (returned by hostnamectl command) on the installed system. Anyway, the idea of this PR is to give to user a 'feedback' when clicking on 'Apply' key in UI of installer. 

Setting an _Host Name_ (_Имя узла_ on screenshot) and pressing 'Apply' button will successfully update _static hostname_ on installed system , but in installer user did't get any confirmation message that in the installed system the _hostname_ will be updated to _test_.



At my opinion current_hostname should not point to transient hostname, but to the static hostname that the user will receive on the installed system, and it should update with changing a _Host Name_ input in installer

jkonecny12:
To fix the tests this PR have to be rebased.

Gel0bmstu:
> To fix the tests this PR have to be rebased.
Am i doing it right ? I just rebase 2 commits in one and force-pushed it to PR branch

mikhailnov:
+1 for this idea, users think that the apply button does not work because the histname does not change.

Installer is always run in LiveCD, if I understand correctly what transient hostname is, it is a hostname received from DHCP, but I don't think that DHCP is often configured to give a special hostname to the system on which livecd is being run.

jkonecny12:
I've rebased the PR on behalf of you. Hope you don't mind that. Tests should be fine now.

rvykydal:
> oh, sorry, it seemed to me that current_hostnme is some kind of the current state of static_hostname (returned by hostnamectl command) on the installed system. Anyway, the idea of this PR is to give to user a 'feedback' when clicking on 'Apply' key in UI of installer.
> 
> Setting an _Host Name_ (_Имя узла_ on screenshot) and pressing 'Apply' button will successfully update _static hostname_ on installed system , but in installer user did't get any confirmation message that in the installed system the _hostname_ will be updated to _test_.
> 
> 
> 
> At my opinion current_hostname should not point to transient hostname, but to the static hostname that the user will receive on the installed system, and it should update with changing a _Host Name_ input in installer

Static hostname for target system is set and displayed in the entry, current (transient) hostname of installer environment is displayed on the right side. So the meaning of "Apply" is not "Apply the change from the entry to the setting of target system hostname" as you probably understand it, but "Apply the hostname configured for target system to the current installer environment" - and it is used mainly in reconfiguration mode of initial setup UI which may be run on installed system after reboot in some cases, for example to reconfigure of finish configuration of pre-installed image.
Apparently this is not so clear from the UI.

mikhailnov:
If we change hostname in the LiveCD as this patch does, won't it cause issues with refreshing IP via DHCP after the initial lease reaches end of life?

rvykydal:
For the reference: https://github.com/rhinstaller/anaconda/pull/3044

rvykydal:
Could you please sum up the issue you are trying to address in this PR (the issue from the description as I understand it seems to be resolved with #3044) - it could be better to file the issue a ticket in bugzilla with a reference to PR with patch so there is a place for discussion if the PR is not found acceptable by reviewers.

rvykydal:
> Could you please sum up the issue you are trying to address in this PR (the issue from the description as I understand it seems to be resolved with #3044) - it could be better to file the issue a ticket in bugzilla with a reference to PR with patch so there is a place for discussion if the PR is not found acceptable by reviewers.

Closing the PR, please open a BZ ticket if you still have an issue to address.

PR #2985, opened 2020-11-09
Remove deviceprobe and install commands

VladimirSlavik:
They have been deprecated a long time ago.

### DIFF
--- a/pyanaconda/core/kickstart/commands.py
+++ b/pyanaconda/core/kickstart/commands.py
@@ -32,7 +32,6 @@ from pykickstart.commands.cdrom import FC3_Cdrom as Cdrom
 from pykickstart.commands.clearpart import F28_ClearPart as ClearPart
 from pykickstart.commands.displaymode import F26_DisplayMode as DisplayMode
 from pykickstart.commands.device import F24_Device as Device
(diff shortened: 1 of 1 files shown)


VladimirSlavik:
See also https://github.com/pykickstart/pykickstart/pull/347

poncovka (approved):
Looks good to me. Thanks!

jstodola:
Can you please wait a bit before merging this PR? Although the install command has been deprecated for a while, this change would cause problems for installations via Beaker (https://beaker-project.org/), where the install command is still provided in the kickstart file even for rawhide.
Bug against Beaker: https://bugzilla.redhat.com/show_bug.cgi?id=1896479

VladimirSlavik:
Sure, let's do that.

However, specifically this change does not matter. It's only two lines because even before, the commands did not do anything. It's the pykickstart change that actually changes outward behavior - that makes these commands not parse.

poncovka:
This PR is safe to merge, because it only removes some unused imports. I don't think it's a notable change.

jstodola:
OK, feel free to merge this change, I will follow the pykickstart PR to avoid unpleasant surprises before the Beaker bug is resolved.
Sorry for the noise :-)

PR #2986, opened 2020-11-11
Use podman in GitHub workflows

martinpitt:
This is closer to what developers use locally.

Also run the workflows on Ubuntu 20.04 instead of 18.04, to use the
latest LTS version.

---

Note that this is totally unrelated to the `test -r` / `access()` trouble that I had with ELN. It's just some nice little improvement/housekeeping.

### DIFF
--- a/.github/workflows/container-autoupdate.yml
+++ b/.github/workflows/container-autoupdate.yml
@@ -8,7 +8,7 @@ on:
 jobs:
   refresh-container:
     name: Refresh anaconda-ci container
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
     # we need to have matrix to cover all branches because schedule is run only on default branch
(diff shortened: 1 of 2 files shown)


martinpitt:
[Successful run on my fork](https://github.com/martinpitt/anaconda/runs/1385420621?check_suite_focus=true)

VladimirSlavik (approved):
Looks good to me.

Hopefully this will reduce the differences in behavior, if any?

martinpitt:
@VladimirSlavik : Yes, that was the idea. Right now I'm not aware of any actual differences between podman vs. docker, or 18.04 vs. 20.04, but let's not challenge our luck too much :grin:

martinpitt:
The nightly container refresh is broken due to that weird `m4 not found` issue that I debugged yesterday. I need to send an urgent PR for that. In order to avoid conflicts, I'll land this now, even if it's not yet 24 hours old.

PR #2987, opened 2020-11-11
Run and fix master tests on Fedora ELN

martinpitt:
Let's make sure that Anaconda keeps building/working on ELN, to avoid breakage in RHEL 9.

This requires adjusting the unit tests to not require blivet-gui on ELN.

 - [x] Builds on top of PR #2991
 - [x] Enable [Everything compose](https://odcs.fedoraproject.org/composes/production/latest-Fedora-ELN/compose/Everything) to replace rawhide packages

### DIFF
--- a/tests/nosetests/pyanaconda_tests/modules/storage/partitioning/module_part_blivet_test.py
+++ b/tests/nosetests/pyanaconda_tests/modules/storage/partitioning/module_part_blivet_test.py
@@ -17,6 +17,7 @@
 #
 # Red Hat Author(s): Vendula Poncova <vponcova@redhat.com>
 #
+import os
 import sys
 import pickle
 import unittest
@@ -25,13 +26,15 @@ from unittest.mock import patch
 from pyanaconda.modules.storage.devicetree import create_storage
 from tests.nosetests.pyanaconda_tests import patch_dbus_publish_object, check_task_creation
 
-from blivetgui.communication.proxy_utils import ProxyID
-
 from pyanaconda.modules.storage.partitioning.blivet.blivet_module import BlivetPartitioningModule
 from pyanaconda.modules.storage.partitioning.blivet.blivet_interface import \
     BlivetPartitioningInterface
 from pyanaconda.modules.storage.partitioning.interactive.interactive_partitioning import \
     InteractivePartitioningTask
+
+
+# blivet-gui is supported on Fedora, but not ELN/CentOS/RHEL
+HAVE_BLIVET_GUI = os.path.exists("/usr/bin/blivet-gui")
 
 
 class BlivetPartitioningInterfaceTestCase(unittest.TestCase):
@@ -79,18 +82,21 @@ class BlivetPartitioningInterfaceTestCase(unittest.TestCase):
             request = pickle.dumps(("call", "get_disks", []))
             blivet_module.send_request(request)
 
+    @unittest.skipUnless(HAVE_BLIVET_GUI, "blivet-gui not installed")
     def storage_handler_test(self):
         """Test the storage_handler property."""
         self.module.on_storage_changed(create_storage())
         self.assertIsNotNone(self.module.storage_handler)
         self.assertEqual(self.module.storage, self.module.storage_handler.storage)
 
+    @unittest.skipUnless(HAVE_BLIVET_GUI, "blivet-gui not installed")
     def request_handler_test(self):
         """Test the request_handler property."""
         self.module.on_storage_changed(create_storage())
         self.assertIsNotNone(self.module.request_handler)
         self.assertEqual(self.module.storage_handler, self.module.request_handler.blivet_utils)
 
+    @unittest.skipUnless(HAVE_BLIVET_GUI, "blivet-gui not installed")
     def send_request_test(self):
         """Test SendRequest."""
         self.module.on_storage_changed(create_storage())
@@ -99,6 +105,7 @@ class BlivetPartitioningInterfaceTestCase(unittest.TestCase):
         answer = self.interface.SendRequest(request)
         answer = pickle.loads(answer)
 
+        from blivetgui.communication.proxy_utils import ProxyID  # pylint: disable=import-error
         self.assertIsInstance(answer, ProxyID)
         self.assertEqual(answer.id, 0)
 
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -3,6 +3,17 @@ on: [push, pull_request_target]
 jobs:
   unit-tests:
     runs-on: ubuntu-20.04
+    strategy:
+      matrix:
+        # empty = release that corresponds to current branch name
+        release: ['', 'eln']
+        include:
+          - release: eln
+            build-args: '--build-arg=eln=1'
+    env:
+      CI_TAG: '${{ matrix.release }}'
+      CONTAINER_BUILD_ARGS: '${{ matrix.build-args }}'
+
     steps:
       - name: Clone repository
         uses: actions/checkout@v2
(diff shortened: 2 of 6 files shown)


martinpitt:
[Example run on my fork](https://github.com/martinpitt/anaconda/runs/1385415761?check_suite_focus=true). This fails right now (thus marking as blocked), as blivet-gui is not available in ELN/RHEL 9 and the tests don't account for it. These should be fixed first before landing  this -- but if you do, please slurp in this PR to be able to build and locally run an ELN container locally.

martinpitt:
I added a second commit which conditionally skips/adjusts the blivet-gui tests for ELN. I first thought about something like:
```py
try:
    import blivetgui.communication.proxy_utils
except ImportError:
   no_blivet_gui = True
```
but I find that too brittle -- it would silently make the tests succeed if blivet-gui is broken in Fedora. So I went for an explicit declaration and "fail-if-forgotten" approach, which takes a little more work, but is much more robust.

[Successful run on my fork](https://github.com/martinpitt/anaconda/runs/1389101999?check_suite_focus=true) -- just look at the eln test, ignore the two others -- that's an independent issue, and I'll send a PR ASAP.

@poncovka , WDYT?

tests/nosetests/pyanaconda_tests/modules/storage/partitioning/module_part_blivet_test.py
```diff
@@ -25,15 +26,18 @@
...
 from pyanaconda.modules.storage.partitioning.blivet.blivet_module import BlivetPartitioningModule
 from pyanaconda.modules.storage.partitioning.blivet.blivet_interface import \
     BlivetPartitioningInterface
 from pyanaconda.modules.storage.partitioning.interactive.interactive_partitioning import \
     InteractivePartitioningTask
 
 
+# blivet-gui is supported on Fedora, but not ELN/CentOS/RHEL
+# define this env to skip the corresponding tests
+NO_BLIVET_GUI = os.getenv('NO_BLIVET_GUI')
```
poncovka, line 38: I could argue that this solution has the same issues as the `ImportError` solution. If we mess up the environment (and we will definitely do that one day), we will start to skip these tests without noticing. Therefore, the tests should run based on the importability of `blivetgui`, but we can have a better control over the skipped tests in general.

This kind of error will affect the coverage of the code. I am planning to enable `codecov.io` anyway, so any change in skipped tests will be visible from every pull request.

More strict solution would be to work with a number of skipped tests and verify that we have skipped exactly that number of tests. I don't think it is necessary, but at least it is more universal and not specific for blivetgui.

martinpitt: I'd argue that it's much harder to "accidentally" set `$NO_BLIVET_GUI` in our container than blivet-gui suddenly breaking in Rawhide. The former is "explicitly opt in" (and if you do nothing, the test fails), the latter (or any approach with catching `ImportError`) is "if you do nothing, the test wrongly passes".

> verify that we have skipped exactly that number of tests

That has the same principal problem, though -- you need to control *that* condition somehow, based on whether you are in a Rawhide or ELN container.

So, I certainly appreciate better ideas how to tighten the control over the skips. Gating on reduced code coverage certainly sounds like a great idea!

But I strongly veto an `except ImportError:` approach, as that's just laying a trap for ourselves.

poncovka: > I'd argue that it's much harder to "accidentally" set $NO_BLIVET_GUI in our container than blivet-gui suddenly breaking in Rawhide. 

I like your optimism. Anyway, if blivet-gui is broken, is it really our problem? Do we really want to block anaconda because of that? It should be possible to finish the installation with broken blivet-gui. Therefore, it will not block anyone, but our CI will fail until it is fixed.

Also, blivet-gui is gated in Rawhide ;-)

> That has the same principal problem, though -- you need to control that condition somehow, based on whether you are in a Rawhide or ELN container.

Sure, but it doesn't create additional noise in the unit tests. The number of skipped tests will be kept outside the container and checked outside the container. It is not necessary to propagate this information into the container, where we can accidentally change it.

martinpitt: > Anyway, if blivet-gui is broken, is it really our problem? Do we really want to block anaconda because of that?

I suppose that's eventually your team's decision. In cockpit's team I'd say "absolutely, yes" - if my project's dependencies don't work, then I'm screwed as well, and want to know about it, report it, and see that it gets fixed. Just covering your eyes and let users figure it out isn't really better IMHO. :grin: 

> Therefore, it will not block anyone, but our CI will fail until it is fixed.

That is a  good point. Note that this PR is mostly just the initial workflow. Once it lands, we want to build the container in a nightly workflow and gate it on tests, so that (1) we don't block our PRs on Rawhide regressions, and (2) make the tests faster. I.e. exactly the same that we do for the normal rawhide anaconda-ci right now.

But aside from that, it's obvious that you don't like this current approach. What would you rather have instead?

martinpitt: I thought about this again, and I'm actually not sure what your main objection is: The way we get that information from the outside into the unit test? (I do agree that the current approach is rather inelegant)? Or the fact that we "insist" on testing blivet-gui on Rawide? In our call I had the impression that it was the latter, but we are doing that right now, and always have for rawhide. So that wouldn't be anything new.

But anyway, I'll work on a patch tomorrow where the unit test itself will check if blivet-gui is installed, and drop the whole env var business. The .spec already enforces the installation of  blivet-gui on Rawhide, that's good enough. And it will make the structure much simpler. Thanks for the review!

martinpitt:
I reworked this heavily, see the [successful run on my fork](https://github.com/martinpitt/anaconda/actions/runs/361149570).

@poncovka : The tests now check themselves whether blivet-gui is installed, which is more local and easier to understand. I think/hope that's the direction you wanted, but please let me know if you have something else in mind.

I checked both logs carefully, and the Rawhide one has no skips, while the ELN one has the three expected skips:
```
Test the request_handler property. ... SKIP: blivet-gui not installed
Test SendRequest. ... SKIP: blivet-gui not installed
Test the storage_handler property. ... SKIP: blivet-gui not installed
```

@jkonecny12 : I also made the container build more "regular", it now builds `anaconda-ci:eln`, i.e. uses the `:eln` tag instead of "anaconda-ci-eln:master". That's easier to see on quay, and we won't ever build something like a "f33-eln" container as that's just not a thing. That allowed me to avoid duplicating the whole `unit-tests:` job, and instead use a strategy matrix. The remaining little issue is that copying that workflow unmodified to e.g. the F33 branch will still run the ELN test -- we either need to drop it from the matrix, or shrug and don't care about the extra run (it's not going to take longer, and the stable branches don't get that many runs anyway). A matrix can't be built dynamically. I also tried something like

```
    # only run ELN test on master branch
    if: matrix.release != 'eln' || github.base_ref == 'master'
```
but that doesn't work -- `if:` is evaluated once for the entire job, *before* the matrix gets expanded. What we could do is to put that if on the individual steps (like building and running the container), but that's also a bit awkward.

The only other way would be to entirely duplicate the job like I [did earlier](https://github.com/rhinstaller/anaconda/commit/e296e020cfdbdc9a99a7c6446ac01bae984d716a) -- but that's a lot more code duplication.

So in summary, I think adjusting the matrix on the F3x branches is the easiest option.

poncovka (approved):
It looks great. Thanks!

jkonecny12 (approved):
Looks good to me. Thanks!

PR #2988, opened 2020-11-11
Handle disabled and failing add-ons

poncovka:
If the add-on started by the `start-module` script fails before the
timeout of sixty seconds, return its exit status. Otherwise, return zero.

If the add-on fails to start, just remove it from the list of available modules
and continue. Don't abort the whole installation because of that.

This will allow the kdump add-on to refuse to run its DBus module if it is
not enabled via the `inst.kdump_addon` boot option.


VladimirSlavik (dismissed):
Looks good to me. Thank you!

jkonecny12 (dismissed):
Looks good to me. Nice improvement with the timeout!

poncovka:
Rebased to fix the tests.

PR #2990, opened 2020-11-11
Fix the priority of the ostreesetup kickstart command (#1896761)

poncovka:
Keep the same priorities as in the payload property of the Anaconda class.
If ostreesetup is in the kickstart file, always create the RPM OSTree module.

Resolves: rhbz#1896761


jkonecny12 (commented):
Good catch. LGTM!

jkonecny12 (dismissed):

VladimirSlavik (approved):
Looks good to me. Thank you!

poncovka:
Rebased to fix the tests.

PR #2991, opened 2020-11-12
Fix GitHub container failure with latest Fedora Rawhide

martinpitt:
Turns out that the weird "m4 not found" error that I was brooding over yesterday was in fact not specific to Fedora ELN, but affects Rawhide as well. [Last night's container refresh broke due to this](https://github.com/rhinstaller/anaconda/actions/runs/358720903), and the RPM tests (which don't use a static container) now [fail as well](https://github.com/rhinstaller/anaconda/pull/2990/checks?check_run_id=1386774946).

This PR avoids the error in a more generic fashion than I did in #2987 (and I'll rebase/adjust that PR on top of this).

This is a bit urgent, as currently our CI is broken, so I'd like to skip the 24 hour thing for this. @jkonecny12 , @poncovka , can you please review this quickly?

### DIFF
--- a/.github/workflows/container-autoupdate.yml
+++ b/.github/workflows/container-autoupdate.yml
@@ -26,10 +26,12 @@ jobs:
 
       - name: Run tests in anaconda-ci container
         id: run-tests
-        run: make -f Makefile.am container-ci
-        continue-on-error: true
+        run: |
+          # put the log in the output, where it's easy to read and link to
+          make -f Makefile.am container-ci || { cat test-logs/test-suite.log; exit 1; }
 
       - name: Upload test and coverage logs from local testing
+        if: always()
         uses: actions/upload-artifact@v2
         with:
           name: logs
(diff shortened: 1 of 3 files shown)


martinpitt:
I also took the liberty to robustify/simplify the container-autoupdate workflow. That will also make the [actions output](https://github.com/rhinstaller/anaconda/runs/1387956306?check_suite_focus=true) easier to read -- it's weird that the "Run tests in anaconda container" step appears as successful when it in fact failed.

martinpitt:
[Successful run on my fork](https://github.com/martinpitt/anaconda/actions/runs/359226847). The tests appear failing here because of `pull_request_target`.

poncovka (approved):
Looks good to me. Thank you!

.github/workflows/container-autoupdate.yml
```diff
@@ -39,12 +39,6 @@ jobs:
             tests/pylint/runpylint*.log
             tests/coverage-*.log
 
-      - name: Fail if local test did not worked
-        if: ${{ steps.run-tests.outcome == 'failure' }}
-        run: |
-          cat tests/test-suite.log
```
jkonecny12, line 45: Don't we want to have this `cat` call as part of ``run-tests`` step?

jkonecny12 (approved):
LGTM! Just a nitpick below.

.github/workflows/container-autoupdate.yml
martinpitt: Good point, thanks! Fixed.

jkonecny12 (approved):
Looks even better now!

martinpitt:
[Successful result on my fork](https://github.com/martinpitt/anaconda/actions/runs/359531390) after adjusting the daily container refresh.

PR #2992, opened 2020-11-12
Drop testing in mock

martinpitt:
By now we have settled on running tests in the container, both locally
and on CI. Testing in mock isn't validated in CI any more and thus not
supported any more, and it's also confusing to document two different ways.

----
:musical_score: :musical_note: Are you ready, hey! Are you ready for this? Are you hanging on the edge of your seat?  :notes: 

This is very low-priority, just something to get off my TODO list. To make sure that everyone is ok with this, I'll ask for review approval of everyone. Please do speak up if you think this is too early, or something in the container based workflow doesn't suit you. Thanks!

 - [x] Builds on top of PR #2993


martinpitt:
[oh-uh](https://github.com/rhinstaller/anaconda/pull/2992/checks?check_run_id=1390126612) , that has [worked previously on my fork](https://github.com/martinpitt/anaconda/actions/runs/359494179). But that part of the cleanup is much less contentious, I'll send that as a separate PR first.

martinpitt:
Oh, of course -- again I fell into the `pull_request_target` trap. It's using the workflow from master and thus not picking up the change to install_dependencies.

jkonecny12 (approved):
You are again solving items from my TODO list :+1: . Looks good to me!

martinpitt:
#2993 landed, rebased on top of that. This should clear up the failing rpm-tests.

VladimirSlavik (commented):
Code-wise LGTM, but the docs absolutely not. I'd be happy to work on that part if you're ok with that.

Why: You delete the part about installing dependencies, but neglect its replacement. The never explicitly mentioned requirement to run `./configure && make` is then even more hidden. Given that the `make -F Makefile.am` bit is hidden in text, the default seems to be to not install deps, but then all the commands after that are as if they were. Overall it sets up the reader for failure.

VladimirSlavik (dismissed):
ACKing to signal that I'm OK with the code.

martinpitt:
Thanks @VladimirSlavik for the review! Indeed I meant to update the `-f Makefile.am` bits, and forgot. I did that now. Now the commands work from a clean checkout.

> You delete the part about installing dependencies, but neglect its replacement.

The replacement is running tests or an interactive shell in the container. That already has all the dependencies installed.

tests/README.rst
```diff
@@ -80,46 +37,48 @@ Logs from the run are stored in the ``test-logs/`` folder; no other files are
...
-
-Please update your container from time to time to have newest dependencies.
-To do that just run the build again.
+Prepare the environment and build the sources::
 
-Run tests inside Mock
----------------------
+    ./autogen.sh
+    ./configure
+    make
```
VladimirSlavik, line 56: Won't this need the dependencies too?

martinpitt: Yes, it does. But the rawhide/ELN/rhel-8 container have all the dependencies (that's the point of having the container). Note that this is now in the "Interactively work inside of container" section. That may be a bit hard to see from the diff, it may be easier to look at [the rendered README](https://github.com/martinpitt/anaconda/blob/remove-mock/tests/README.rst)?

poncovka (approved):
It looks great. Thank you!

tests/README.rst
VladimirSlavik: Ah yes, you're completely right.

VladimirSlavik (approved):
Thank you!

Looking forward to dropping the old workflow.

M4rtinK (approved):
Looks good to me, nice cleanup! :)

jkonecny12 (approved):
Looks great to me still :)

rvykydal (approved):
Looks good to me, thank you.

jkonecny12:
+1 for @martinpitt . You have a winning score for our outdated infra!

PR #2993, opened 2020-11-12
Drop obsolete dependency_solver.py script from rpm-tests

martinpitt:
We want to remove that soon, so use the much more lightweight
install_dependencies.sh script instead.


martinpitt:
[successful run on my fork](https://github.com/martinpitt/anaconda/actions/runs/359569747) -- the tests here use the master workflow, so they are not representative.

VladimirSlavik (approved):
Looks good to me, as far as I can tell. Thank you!

jkonecny12 (approved):
LGTM!

PR #3000, opened 2020-11-16
Move the package-related code to the DNF module

poncovka:
Move code that works with the `%packages` section to the DNF module.

It will simplify the removal of the `%packages` section from UI, because
it can be tested with unit tests now.


martinpitt:
Congratulations to PR #3000! :tada: :balloon:

jkonecny12 (approved):
Looks great, thanks a lot! Finally, we have some test coverage here!

poncovka:
Rebased.

PR #3001, opened 2020-11-17
Install python-copr and RedHat CA into anaconda-ci container

martinpitt:
This is a dependency of https://github.com/vojtechtrefny/copr-builder

Related: rhbz#1885635

----

This is a prerequisite for replacing the anaconda_daily_rhel_copr Jenkins job.

PR #3002, opened 2020-11-17
Add workflow for daily RHEL COPR build

martinpitt:
Generate the configuration right in the workflow, so that it is easy to
change and obvious where it comes from.

 - [x] Install python-copr into anaconda-ci container: PR #3001 
 - [x] Make RHEL COPR token available in self-hosted github runners: builders PR 29
 - [x] Disable anaconda_daily_rhel_copr Jenkins job

### DIFF
--- a/.github/workflows/daily-rhel-copr.yml
+++ b/.github/workflows/daily-rhel-copr.yml
@@ -1,0 +1,32 @@
+name: Build current anaconda rhel-8 branch in RHEL COPR
+on:
+  schedule:
+    - cron: 0 2 * * *
+  # be able to start this action manually from a actions tab when needed
+  workflow_dispatch:
+
+jobs:
+  build:
+    name: copr-builder
+    runs-on: [self-hosted, ci-tasks, rhel-8]
+    steps:
+      - name: Checkout copr-builder repository
+        uses: actions/checkout@v2
+        with:
+          repository: vojtechtrefny/copr-builder
+
+      - name: Create copr-builder configuration
+        run: |
+          cat <<EOF > copr-builder-rhel.conf
+          [anaconda]
+          copr_user = rhinstaller-group
+          copr_repo = Anaconda
+          package = anaconda
+          git_url = https://github.com/rhinstaller/anaconda
+          git_branch = rhel-8
+          pre_archive_cmd = ./autogen.sh && ./configure
+          archive_cmd = make po-pull && make dist
+          EOF
+
+      - name: Run copr-builder
+        run: ./copr-builder -v -c copr-builder-rhel.conf -C /copr-rhel


martinpitt:
This is a bit tricky to test, as the workflow needs to be on master so that it can work all the way through (credentials, and such). So I tested this in two parts.

First, I [added an extra debug hack](https://github.com/martinpitt/anaconda/commit/7b08d08ce969304f597e290710a53a54516fa4db) which runs the workflow on Ubuntu instead of self-hosted, and adds ssh. 
I [triggered the workflow](https://github.com/martinpitt/anaconda/runs/1411337573?check_suite_focus=true), and entered with ssh. There I manually did the setup that our self-hosted container already has (or rather, will have with PR #3001):
```
sudo apt-get install python3-setuptools
sudo pip3 install copr
```

Then I checked that the config file looks sane:
```
$ cat copr-builder-rhel.conf 
[anaconda]
copr_user = rhinstaller-group
copr_repo = Anaconda
package = anaconda
git_url = https://github.com/rhinstaller/anaconda
git_branch = rhel-8
pre_archive_cmd = ./autogen.sh && ./configure
archive_cmd = make po-pull && make dist
```

and finally that copr-builder can run. I didn't give it a valid credential, and it can't reach the internal COPR anyway, but this at least makes sure that dependencies and the command line is correct:
```
$ echo pwned | sudo tee /copr-rhel
$ ./copr-builder -v -c copr-builder-rhel.conf -C /copr-rhel
copr.builder: Failed to get Copr token expiration date from config file.
  File "./copr-builder", line 43, in <module>
    builder = CoprBuilder(args.config, args.copr_config)
  File "/home/runner/work/anaconda/anaconda/copr_builder/copr_builder.py", line 34, in __init__
    self.copr = copr.create_client2_from_file_config(filepath=self.copr_config)
  File "/usr/local/lib/python3.6/dist-packages/copr/client_v2/client.py", line 196, in create_from_file_config
    if not raw_config.read(filepath):
  File "/usr/lib/python3.6/configparser.py", line 697, in read
    self._read(fp, filename)
  File "/usr/lib/python3.6/configparser.py", line 1080, in _read
    raise MissingSectionHeaderError(fpname, lineno, line)
configparser.MissingSectionHeaderError: File contains no section headers.
file: '/copr-rhel', line: 1
'pwned\n'
```

Secondly, I `podman exec`'ed our github-runner-1 container on the internal runner, and ran the git clone and two commands from the workflow (create config file and run copr-builder):
```
# ./copr-builder -v -c copr-builder-rhel.conf -C /copr-rhel
copr.builder: Package anaconda (repo rhinstaller-group/Anaconda): New SRPM build started.
copr.builder: Package anaconda (repo rhinstaller-group/Anaconda): Found latest build: anaconda-33.16.4.1-5.20201117gitb9d38cfb9.fc31 (ID: 59658)
copr.builder: Package anaconda: Running prepare archive commands.
copr.builder: Package anaconda (repo rhinstaller-group/Anaconda): Newest version is already built (git hash: b9d38cfb9).
```

So this is all expected and seems to work.

martinpitt:
The infra PR (builders 29) landed, so this is unblocked. @jkonecny12 , after review/approval I suggest to disable the Jenkins job first and then land this, so that it doesn't happen twice.

jkonecny12 (approved):
Looks good to me.
Also I've got a heads-up from @vojtechtrefny that the copr-builder is failing now. Seems that the problem is in the copr or copr library. So do not care much if the job will fail.

Jenkins job is disabled. You are free to go ;)

Thanks a lot for this!

martinpitt:
@jkonecny12 : WDYM with "broken"? I [triggered the action manually](https://github.com/rhinstaller/anaconda/runs/1442306863), it succeeded and [did a COPR build](https://copr.devel.redhat.com/coprs/rhinstaller-group/Anaconda/build/60464/).

jkonecny12:
hmm, maybe it's fixed already?

vojtechtrefny:
> hmm, maybe it's fixed already?

I love weird bugs: The problem with copr happens only for "normal" users, not groups and this made me think the problem exists only on the internal copr, because I use my personal account there not a group account...

PR #3007, opened 2020-11-19
Rebase branches to current master before testing

martinpitt:
This makes sure that we test what we will actually land: The PR's
commits will be rebased on top of master. It also avoids rebuilding the
anaconda-ci container when the proposed branch is older, and doesn't
include recent container changes.


jkonecny12 (approved):
This will be great improvement! Thanks!

VladimirSlavik:
Thank you, that's a really important distinction. IMO it finally brings the new tests on par with the old ones which tested a merge commit :-)

PR #2977, opened 2020-11-05
Network select wireless ap with dialog

rvykydal:

PR #3009, opened 2020-11-20
network: do not use dialog for just single wireless connection to con…

rvykydal:
…figure

Follow up to https://github.com/rhinstaller/anaconda/pull/2977

### DIFF
--- a/pyanaconda/ui/gui/spokes/network.py
+++ b/pyanaconda/ui/gui/spokes/network.py
@@ -497,6 +497,20 @@ class NetworkControlBox(GObject.GObject):
             dialog.refresh(device_name)
             dialog.run()
 
+    def _get_the_only_wireless_connection(self, device):
+        con_uuid = con_ssid = ""
+        if device:
+            cons = device.filter_connections(self.client.get_connections())
+            if len(cons) == 1:
+                connection = cons[0]
+                con_uuid = connection.get_setting_connection().get_uuid()
+                wireless_setting = connection.get_setting_wireless()
+                if wireless_setting:
+                    ssid_variant = wireless_setting.get_ssid()
+                    if ssid_variant:
+                        con_ssid = ssid_variant.get_data()
+        return con_uuid, con_ssid
+
     def on_edit_connection(self, *args):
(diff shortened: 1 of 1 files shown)


rvykydal:
Show dialog for network selection only if we have multiple profiles:
https://rvykydal.fedorapeople.org/new_ui_configure_more_cons_available.webm

If there is only single connection, edit it right away:
https://rvykydal.fedorapeople.org/new_ui_configure_single_con.webm
... The connection is created when selecting the network and in this case entering wrong password. Typical use case is WPA Enterprise where we don't ask for credentials in secret agent dialog when selecting the network (too complex), but we just create the connection and let the user enter credentials via configuration using nm-c-e.

poncovka (approved):
Looks good to me. Thanks!

jkonecny12 (approved):
LGTM!

jstodola:
I'm a bit late, but I found it confusing to allow to click the "Configure..." button if there is nothing to be configured. (https://rvykydal.fedorapeople.org/new_ui_configure_single_con.webm). Shouldn't the button be disabled if the list of networks to configure is empty?

rvykydal:
> I'm a bit late, but I found it confusing to allow to click the "Configure..." button if there is nothing to be configured. (https://rvykydal.fedorapeople.org/new_ui_configure_single_con.webm). Shouldn't the button be disabled if the list of networks to configure is empty?

I will add this in another PR when NM crashes which would make the configuration with this addition even more fragile are fixed (https://bugzilla.redhat.com/show_bug.cgi?id=1873561)

rvykydal:
> > I'm a bit late, but I found it confusing to allow to click the "Configure..." button if there is nothing to be configured. (https://rvykydal.fedorapeople.org/new_ui_configure_single_con.webm). Shouldn't the button be disabled if the list of networks to configure is empty?
> 
> I will add this in another PR when NM crashes which would make the configuration with this addition even more fragile are fixed (https://bugzilla.redhat.com/show_bug.cgi?id=1873561)

https://github.com/rhinstaller/anaconda/pull/3098

PR #3010, opened 2020-11-21
gui: fix window width with long product titles

mikhailnov:
It was:
image

It is now:
image

Also, by the way, a similar fix may be done in the dashboard of available spokes
image
by adding something like `<property name="wrap">True</property>` into `*.glade`.

### DIFF
--- a/pyanaconda/ui/gui/spokes/welcome.py
+++ b/pyanaconda/ui/gui/spokes/welcome.py
@@ -249,6 +249,11 @@ class WelcomeLanguageSpoke(StandaloneSpoke, LangLocaleHandler):
         welcomeLabel.set_text(_("WELCOME TO %(name)s %(version)s.") %
                 {"name" : productName.upper(), "version" : productVersion})         # pylint: disable=no-member
 
+        # Product title may be long and make the window of Anaconda so wide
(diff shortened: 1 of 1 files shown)


VladimirSlavik:
Hello @mikhailnov, thank you for the fix! The change is good, only you guessed the code/glade fix the other way :) What you changed is a static heading, so please change it in the ui file instead. It's right next to the code you changed - same file name, only `.glade` instead of `.py`. If it turns out you need `AnacondaWidgets` and can't load these, just ask?

For the spoke titles, that would be a good thing too - if you want to take a look at that: These are generated dynamically in hubs. See https://github.com/rhinstaller/anaconda/blob/master/pyanaconda/ui/gui/hubs/__init__.py#L171. You'll have to dig into the widgets for the actual place, but that should be a good starting point. If you're interested in category titles too, these are on line 199 just below.

Generally speaking, we would be happy to fix this anywhere it can happen - if you're happy to go hunt for these, patches (or even just scoping in a bug report) are welcome.

VladimirSlavik:
All ideas mentioned thus far implemented in #3160. Closing.

mikhailnov:
Thanks!

PR #3011, opened 2020-11-23
Fix issue when ns_info cannot be retrieved for NVDimm namespace

jkonecny12:
If we don't skip this part the uncaught exception will raise because we are trying to concatenate string and None types.

This is happening when NVDIMM namespace is set to DEVDAX mode. In this mode there is no device to be returned so we will got None from blivet.

Resolves: rhbz#1891827
(cherry picked from commit 6afc375b164a802e26802ec4ba54d3446c078091)

Backport of https://github.com/rhinstaller/anaconda/pull/2961


jkonecny12:
Smoke test done.

poncovka (approved):
Looks good to me.

PR #3012, opened 2020-11-23

martinezjavier:


jkonecny12:
Smoke test already done on the old PR.

jkonecny12 (approved):
Smoke test done. Everything seems to be working correctly.
Thanks a lot for this!

jkonecny12:
The 24 hour rule was already satisfied on the old PR.

PR #3013, opened 2020-11-23
Rebase test pull request also for rhel-8 contributors

jkonecny12:
By this change the tests will run always on top of the rhel-8 branch.

### DIFF
--- a/.github/workflows/validate-rhel-8.yml
+++ b/.github/workflows/validate-rhel-8.yml
@@ -65,7 +65,11 @@ jobs:
           fetch-depth: 0
 
       - name: Rebase to current rhel-8
-        run: git rebase origin/rhel-8
+        run: |
+          git config user.name github-actions
+          git config user.email github-actions@github.com
+          git log --oneline -1 origin/rhel-8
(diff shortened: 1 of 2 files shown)


jkonecny12:
**UPDATED:**
* Added temp username and email to git, otherwise the rebase will fail (fix also on master).

Tested here: 
* https://github.com/jkonecny12/anaconda/runs/1443215340?check_suite_focus=true

martinpitt (approved):
Thanks! I'd rather see this squashed to avoid confusion, but not important (you can do that with the green button, too).

martinpitt:
We need the "Set workflow temp mail" also on master, see [failed PR](https://github.com/rhinstaller/anaconda/runs/1446766540). @jkonecny12 , will you send a master PR as well, or want me to?

martinpitt:
Erk, nevermind -- this PR already does it for both master and rhel8.

jkonecny12:
Yes, it does and that is main reason why the commits are split.

jkonecny12:
merging to avoid more breakage of the tests.

PR #3014, opened 2020-11-23
Add CentOS Stream config

jcpunk:
CentOS Stream tracks ahead of CentOS Linux.  These config changes note this.

### DIFF
--- a/data/product.d/centos.conf
+++ b/data/product.d/centos.conf
@@ -4,29 +4,4 @@
 product_name = CentOS Linux
 
 [Base Product]
-product_name = Red Hat Enterprise Linux
-
-[Anaconda]
-# List of enabled Anaconda DBus modules for RHEL.
-#  but without org.fedoraproject.Anaconda.Modules.Subscription
-kickstart_modules =
-    org.fedoraproject.Anaconda.Modules.Timezone
-    org.fedoraproject.Anaconda.Modules.Network
-    org.fedoraproject.Anaconda.Modules.Localization
-    org.fedoraproject.Anaconda.Modules.Security
-    org.fedoraproject.Anaconda.Modules.Users
-    org.fedoraproject.Anaconda.Modules.Payloads
-    org.fedoraproject.Anaconda.Modules.Storage
-    org.fedoraproject.Anaconda.Modules.Services
-
-[Bootloader]
-efi_dir = centos
-
-[User Interface]
-default_help_pages =
-    CentOSPlaceholder.txt
-    CentOSPlaceholder.html
-    CentOSPlaceholder.html
-
-[Payload]
-default_source = CLOSEST_MIRROR
+product_name = CentOS Stream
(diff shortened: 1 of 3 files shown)


VladimirSlavik:
@jcpunk you'll need to adjust the list of config files for "oVirt Node Next" too. Just change the line to ```["rhel.conf", "centos-stream.conf", "centos.conf", "ovirt.conf"],```

jcpunk:
In theory, I've just updated it correctly :)

VladimirSlavik:
Tests succeeded, so you did!

data/product.d/centos.conf
```diff
@@ -4,29 +4,4 @@
 product_name = CentOS Linux
 
 [Base Product]
-product_name = Red Hat Enterprise Linux
```
poncovka, line 7: I am not sure if we want to complicate dependencies between configurations. CentOS Stream could be just an alias for CentOS Linux. Therefore, I would prefer Red Hat Enterprise Linux -> CentOS Linux -> CentOS Stream. @jkonecny12, what do you think?

jcpunk: With CentOS Stream ahead of CentOS Linux, if one was to become an alias of the other, I'd prefer CentOS Stream to be the 'real' release with CentOS Linux as the alias.

jkonecny12: If we want to be completely correct here than it should be ``CentOS Stream`` -> ``RHEL`` -> ``CentOS`` but I don't think we want to complicate things so much.

For that reason I agree with @poncovka . We won't be reflecting how things really happen so or so. But if we go path
``RHEL`` -> ``CentOS`` -> ``CentOS Stream`` then we would make something like stability order. I mean most stable systems on the left and the most prone to change on the right. Do you agree @jcpunk ?

Conan-Kudo: What should probably be the way this works is that there's a base CentOS config that both CentOS Linux and CentOS Stream derive from, since there's a lot that's in common between the two.

jcpunk: This is something I'll want to think through a bit... with the holiday coming up it can certainly sit for a time.

jcpunk: In light of today's announcement I'd like to revisit this.

VladimirSlavik: IMO the two centoses already were related mostly by name and the announcement means any divergences will only grow. So the correct approach would be to derive both from RHEL, which is the real centerpiece anyway. @Conan-Kudo's idea expands on that, and sounds almost good, except I wouldn't go as far as creating abstract classes in config files...

Just one more opinion.

poncovka (approved):
Looks good to me. Thanks!

data/product.d/centos.conf
poncovka: Let's use @jcpunk's solution.

PR #3015, opened 2020-11-23
Force hide CDN button when no CDN sources

jcpunk:
In CentOS Linux and CentOS Stream this patch seems needed to correctly hide the CDN button.

### DIFF
--- a/pyanaconda/ui/gui/spokes/installation_source.glade
+++ b/pyanaconda/ui/gui/spokes/installation_source.glade
@@ -636,6 +636,7 @@
                                 <property name="can_focus">True</property>
                                 <property name="receives_default">False</property>
                                 <property name="no_show_all">True</property>
+                                <property name="visible">False</property>
                                 <property name="margin_left">12</property>
                                 <property name="use_underline">True</property>
                                 <property name="xalign">0</property>
--- a/pyanaconda/ui/gui/spokes/installation_source.py
+++ b/pyanaconda/ui/gui/spokes/installation_source.py
@@ -853,9 +853,10 @@ class SourceSpoke(NormalSpoke, GUISpokeInputCheckHandler, SourceSwitchHandler):
     def _initialize(self):
         threadMgr.wait(constants.THREAD_PAYLOAD)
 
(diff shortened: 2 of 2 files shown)


VladimirSlavik:
I tested this and I think it works as intended - @M4rtinK can you please check too?

VladimirSlavik (approved):

poncovka (approved):
I don't understand why it doesn't work in CentOS, but it works in Fedora. Anyway, it looks good to me. Thanks!

VladimirSlavik:
Actually I'm not sure **why**, either. I like that this inverts the logic, because the overall intent seems to be hidden->visible, which this makes happen on all levels. I also checked that it works, hence the ack. But I, too, do not know why the original was not enough - thus asking @M4rtinK ...

M4rtinK:
So I downloaded a CentOS Stream (http://merlin.fit.vutbr.cz/mirrors/centos/8-stream/isos/x86_64/) boot ISO and looked what Anaconda version is being used - it's 33.16.3.26-1. This is the Anaconda version used in RHEL 8.3 GA.

Thinking about it, that makes sense as CentOS Stream is basically a rolling rebuild of RHEL 8 at the moment, so it has the RHEL 8 version of Anaconda. This also explains why the CDN button is correctly hidden in Rawhide but not in CentOS stream - Rawhide has the #2850, which fixes the issues. I've just tried applying the fix on the CentOS stream boot iso and it works:

centos_stream_cdn_button_fix

So to fix the issues I recommend using the existing fix from #2850 to avoid having the issues fixes in two different ways in Fedora & RHEL8. 

Also the fix needs to be applied to the rhel-8 branch - putting it in master would have no effect. All RHEL 8 branch changes require a RHEL 8 bug on Anaconda to be created and attached to the commit/s - could you please create the bug ? Thanks in advance! :)

Once it's created I can take things forward from there - I'll backport the fix and make sure it finds its way to CentOS stream.

poncovka:
I guess this PR can be closed based on https://github.com/rhinstaller/anaconda/pull/3015#issuecomment-733694146.

PR #3016, opened 2020-11-24
Set shorter workflow timeouts

martinpitt:
The default timeout is 6 hours, which just causes unnecessary hangs,
delays, and blocked self-hosted runners if anything goes wrong (such as
https://github.com/PyCQA/pylint/issues/3899).

A normal run takes about 10 minutes, so set a timeout of 30 minutes.
Use one hour for the full container rebuild+validation+upload to be on
the safe side (it usually takes between 15 and 18 minutes).

### DIFF
--- a/.github/workflows/container-autoupdate.yml
+++ b/.github/workflows/container-autoupdate.yml
@@ -15,6 +15,7 @@ jobs:
       matrix:
         branch:
           - master
+    timeout-minutes: 60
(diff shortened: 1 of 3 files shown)


rvykydal (approved):
Looks good to me.

VladimirSlavik (approved):
Looks good to me. Thank you!

PR #3017, opened 2020-11-24
[r8] Apply pylint excessive memory usage fix

martinpitt:
With current astroid 2.4.2, pylint requires a ginormous amount of
memory. Apply the recent fix during container build, until a newer
version gets released.

Revert the hack from commit 283561559b0 and go back to pylint
auto-selecting the number of jobs based on number of CPUs.

Backported from master PR #3018
    
Related: rhbz#1885635

PR #3018, opened 2020-11-24
Apply pylint excessive memory usage fix

martinpitt:
With current astroid 2.4.2, pylint requires a ginormous amount of
memory. Apply the recent fix during container build, until a newer
version gets released.

Revert the hack from commit 283561559b0c and go back to pylint
auto-selecting the number of jobs based on number of CPUs.

----

The corresponding fix for rhel-8 branch is in PR #3017 (landed already)


jkonecny12 (approved):
Thanks for solving this!

PR #3020, opened 2020-11-26
Fix CDN button visibility (Backport to RHEL to fix CentOS stream)

M4rtinK:
Only show the Red Hat CDN button if the Subscription
module appears to be running. To achieve that, we do
the same thing as with the HMC button - the CDN
button invisible by default. And enable it only if it
looks like the Subscription module is running.

### DIFF
--- a/pyanaconda/ui/gui/spokes/installation_source.glade
+++ b/pyanaconda/ui/gui/spokes/installation_source.glade
@@ -633,9 +633,9 @@
                             <child>
                               <object class="GtkRadioButton" id="cdnRadioButton">
                                 <property name="label" translatable="yes" context="GUI|Software Source">Red Hat _CDN</property>
-                                <property name="visible">True</property>
                                 <property name="can_focus">True</property>
                                 <property name="receives_default">False</property>
+                                <property name="no_show_all">True</property>
                                 <property name="margin_left">12</property>
                                 <property name="use_underline">True</property>
                                 <property name="xalign">0</property>
--- a/pyanaconda/ui/gui/spokes/installation_source.py
+++ b/pyanaconda/ui/gui/spokes/installation_source.py
@@ -854,9 +854,9 @@ class SourceSpoke(NormalSpoke, GUISpokeInputCheckHandler, SourceSwitchHandler):
     def _initialize(self):
         threadMgr.wait(constants.THREAD_PAYLOAD)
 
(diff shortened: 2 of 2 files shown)


M4rtinK:
@jcpunk Please create a RHEL 8 bug describing the issues CentOS Stream has with the CDN button & let me know the bug number. I'll handle the rest after that. Thanks! :)

jkonecny12 (approved):
LGTM!

poncovka (commented):
I am not sure why we are doing this since CentOS already uses a patch with a fix. I think that the point of https://github.com/rhinstaller/anaconda/pull/3015 was to make sure that this bug is fixed in upstream.

jcpunk:
I've filed https://bugzilla.redhat.com/show_bug.cgi?id=1903178

M4rtinK:
> I've filed https://bugzilla.redhat.com/show_bug.cgi?id=1903178

@jcpunk Thanks! I've updated the PR to reference the bug number with take it from there process wise. :)



> I am not sure why we are doing this since CentOS already uses a patch with a fix. I think that the point of #3015 was to make sure that this bug is fixed in upstream.

@poncovka So I think this is the patch CentOS is using at the moment:
https://git.centos.org/rpms/anaconda/blob/c8/f/SOURCES/0002-CentOS-disable-cdn-radiobutton.patch

I guess that works, but I think the patch is simple enough that it could be better to just include it in our RHEL 8 branch so that CentOS (and possibly other community rebuilds) don't need to carry a patch for this themselves.

jcpunk:
When reasonable, I prefer to patches 'upstream'.  In theory it prevents folks downstream from rehashing solved problems.

carlwgeorge:
I just checked anaconda-33.16.5.2-1, and every patch CentOS has been applying has been included except this one.  Any chance this can be merged for the next tag?

poncovka (approved):

VladimirSlavik (approved):
Looks good to me. Thank you!

M4rtinK:
Still working fine on latest RHEL 8 as far as I can tell. :)

M4rtinK:
> I just checked anaconda-33.16.5.2-1, and every patch CentOS has been applying has been included except this one. Any chance this can be merged for the next tag?

I did some more testing & rebased the commit on top of latest RHEL 8 branch and it should be merged shortly. :)

PR #3022, opened 2020-11-27
Lower geolocation connection timeout to 5 seconds (#1774873)

M4rtinK:
Lower the connection timeout to 5 seconds as the GUI will wait
only up to 3 seconds for the geolocation thread to finish anyway.

Up till now we used the Anaconda default timeout, which is 45 seconds.
Apparently the actual waiting time could be up to 2x the timeout value,
so 45 seconds was far too long.

Resolves: rhbz#1774873

### DIFF
--- a/pyanaconda/core/constants.py
+++ b/pyanaconda/core/constants.py
@@ -141,9 +141,13 @@ GEOLOC_GEOCODER_NOMINATIM = "geocoder_nominatim"
 # default providers
 GEOLOC_DEFAULT_PROVIDER = GEOLOC_PROVIDER_FEDORA_GEOIP
 GEOLOC_DEFAULT_GEOCODER = GEOLOC_GEOCODER_NOMINATIM
-# timeout (in seconds)
+# how long should the GUI wait for the geolocation thread to finish (in seconds)
+# - GUI starts this count once it finishes its initialization
+# - the geoloc thread is started early and in most cases will be already done
+#   when GUI finishes its initialization, so no delays will be introduced
 GEOLOC_TIMEOUT = 3
-
+# timeout for the network connection used for geolocation (in seconds)
+GEOLOC_CONNECTION_TIMEOUT = 5
 
 ANACONDA_ENVIRON = "anaconda"
 FIRSTBOOT_ENVIRON = "firstboot"
--- a/pyanaconda/geoloc.py
+++ b/pyanaconda/geoloc.py
@@ -475,7 +475,7 @@ class FedoraGeoIPProvider(GeolocationBackend):
     def _refresh(self):
         try:
             reply = self._session.get(self.API_URL,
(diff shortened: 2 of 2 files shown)


VladimirSlavik (approved):
Code lgtm, reasoning sounds ok...

pyanaconda/core/constants.py
```diff
@@ -145,9 +145,13 @@
 # default providers
 GEOLOC_DEFAULT_PROVIDER = GEOLOC_PROVIDER_FEDORA_GEOIP
 GEOLOC_DEFAULT_GEOCODER = GEOLOC_GEOCODER_NOMINATIM
-# timeout (in seconds)
+# how long should the GUI wait for the geolocation thread to finish (in seconds)
+# - GUI starts this count once it fnishes it's initialization
```
poncovka, line 149: `finishes`

pyanaconda/geoloc.py
```diff
@@ -475,7 +475,7 @@ def name(self):
     def _refresh(self):
         try:
             reply = self._session.get(self.API_URL,
-                                      timeout=constants.NETWORK_CONNECTION_TIMEOUT,
+                                      timeout=constants.GEOLOC_CONNECTION_TIMEOUT,
```
poncovka, line 478: Could please change the timeout for all backends?

poncovka (changes requested):

pyanaconda/core/constants.py
M4rtinK: Fixed, thanks! :)

pyanaconda/geoloc.py
M4rtinK: Indeed, oops! :P Should be fixed now. :)

poncovka: The wrong timeout is also used in the `GoogleWiFiLocationProvider` class and the `Geocoder` class.

pyanaconda/core/constants.py
VladimirSlavik: Also its, not it's... Didn't want to play spellchecker, but once we're going there, let's do it right :)

jkonecny12 (approved):
LGTM!

pyanaconda/geoloc.py
M4rtinK: Actually, let's just scrap all that! :D 

That WiFi location and geocoding code was basically a proof of concept that we never documented as supported by the boot options & the API URLs no longer work anyway. I wanted to drop it as part of moving the geolocation code to a task but might as well do it now. :)

See: https://github.com/rhinstaller/anaconda/pull/3030

pyanaconda/core/constants.py
M4rtinK: Fixorzed as well! :)

poncovka (approved):
Looks good to me.

PR #3025, opened 2020-11-30
Change the default_partitioning option (#1887370)

poncovka:
Change the default_partitioning option of the Storage section in the Anaconda
configuration file. Instead of the identifier, specify a mount point and its
attributes on each line.

(cherry-picked from commits f6f373f,  b3d59e6, a58afad and 08093ad)

Resolves: rhbz#1887370

### DIFF
--- a/pyanaconda/modules/storage/partitioning/automatic/automatic_partitioning.py
+++ b/pyanaconda/modules/storage/partitioning/automatic/automatic_partitioning.py
@@ -16,138 +16,20 @@
 # Red Hat, Inc.
 #
 from blivet.partitioning import do_partitioning, grow_lvm
-from blivet.size import Size
 from blivet.static_data import luks_data
 
 from pyanaconda.anaconda_loggers import get_module_logger
-from pyanaconda.core.configuration.anaconda import conf
-from pyanaconda.core.configuration.storage import PartitioningType
 from pyanaconda.modules.common.structures.partitioning import PartitioningRequest
 from pyanaconda.modules.storage.partitioning.automatic.noninteractive_partitioning import \
     NonInteractivePartitioningTask
 from pyanaconda.modules.storage.partitioning.automatic.utils import get_candidate_disks, \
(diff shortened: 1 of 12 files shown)


jkonecny12 (approved):
LGTM!

pyanaconda/modules/storage/partitioning/automatic/automatic_partitioning.py
```diff
@@ -16,140 +16,22 @@
...
 
 __all__ = ["AutomaticPartitioningTask"]
 
 
-SERVER_PARTITIONING = [
-    PartSpec(
-        mountpoint="/",
-        size=Size("2GiB"),
-        max_size=Size("15GiB"),
-        grow=True,
```
jstodola, line 45: grow, btr, lv, thin, encrypted - these arguments have no alternative in the new implementation? After this change, only sizes for mount points are defined, right?
For example PartSpec for swap didn't allow to allocate it as thin LV or btrfs - how this is ensured with this change?

poncovka: Hi, the important part of the implementation is here:

```
        spec = PartSpec(
            mountpoint=name if not swap else None,
            fstype=None if not swap else "swap",
            lv=True,
            thin=not swap,
            btr=not swap,
            size=attrs.get("min") or attrs.get("size"),
            max_size=attrs.get("max"),
            grow="min" in attrs,
            required_space=attrs.get("free") or 0,
            encrypted=True,
        )
```

The arguments `btr`, `lv`, `thin` and `encrypted` are not configurable, because they can be set up automatically. It means that swap still won't be allocated as thin LV or Btrfs. To set up the `grow` argument, we use the `min` and `max` configuration options (instead of `size`).

The previous definitions with `PartSpec` objects were moved to the unit tests where we check that the `PartSpec` objects generated from the configuration files are the same as before (with some minor clean-ups).

jstodola: Ahh, oukej, thanks for the explanation, it makes sense to me now.

PR #3027, opened 2020-11-30
Fix the get_os_release_value function

poncovka:
* Log the name of the parameter that wasn't found in the os-release files.
* Don't use an empty string as a system root.


rvykydal (approved):
Looks good to me.

VladimirSlavik (approved):
Thank you for fixing this!

jkonecny12 (approved):
LGTM!

PR #3028, opened 2020-11-30
Don't run the execute method in the kickstart installation in GUI by default

poncovka:
We always run the `execute` method from the `initialize` method of the spoke when
it is required. The `execute` method will be still called when we leave a spoke.

Call `hubQ.send_ready` without the `justUpdate` argument.
Remove unused `hubQ` messages.

### DIFF
--- a/pyanaconda/ui/common.py
+++ b/pyanaconda/ui/common.py
@@ -215,8 +215,6 @@ class Spoke(object, metaclass=ABCMeta):
         self._payload = payload
         self.applyOnSkip = False
 
-        self.visitedSinceApplied = True
-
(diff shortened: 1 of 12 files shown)


poncovka:
I would like to eventually drop all support for automated runs of the `execute` method in GUI and TUI. The method can be always called explicitly from the `initialize` and `apply` methods. The current solution is just confusing and prone to errors

rvykydal (approved):
Looks good to me.

jkonecny12 (approved):
LGTM! I agree with you. It's not helpful to have this invisible magic happening, it's easy to miss that.

PR #3030, opened 2020-12-02
Drop old WiFi & geocoding based geolocation support

M4rtinK:
This code was developed in the initial stages of geolocation
support in Anaconda with the aim of providing an alternative
to purely IP based territory & timezone determination.

In was considered a prototype back then and was never documented as
something you can enable via the inst.geoloc boot option.

In the end we never did get to making the code production ready
while the IP based geolocation worked fine. Also all the WiFi
location and geocoding API URLs no longer work anyway.

So lets drop that old code & cleanup the docstrings a bit when we are at it.

### DIFF
--- a/pyanaconda/core/constants.py
+++ b/pyanaconda/core/constants.py
@@ -135,12 +135,8 @@ THREAD_SUBSCRIPTION_SPOKE_INIT = "AnaSubscriptionSpokeInitThread"
 # - values are used by the geoloc CLI/boot option
 GEOLOC_PROVIDER_FEDORA_GEOIP = "provider_fedora_geoip"
 GEOLOC_PROVIDER_HOSTIP = "provider_hostip"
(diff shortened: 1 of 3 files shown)


VladimirSlavik (approved):
Apart from the pylint complaints with unused imports, LGTM. Thank you!

```
W0611(unused-import):/tmp/anaconda/pyanaconda/geoloc.py:108,0: : Unused import dbus
W0611(unused-import):/tmp/anaconda/pyanaconda/geoloc.py:107,0: : Unused import urllib.parse
```

M4rtinK:
@VladimirSlavik Pylint should be good to go.

jkonecny12 (approved):
Nice cleanup. Thanks!

poncovka (approved):
Looks good to me. Thanks!

PR #3031, opened 2020-12-02
Fixup category sorting

M4rtinK:
Turns out the category sorting code is wrong
(as categories are classes, not class instances :P)
yet it still manages to provide the expected output,
so this incorrect behavior is not readily apparent.

Thanks vponcova for spotting this! :)

Also when we are at it drop apparently unnecessary sorting
of categories in the collect function and fixup the unit test.
Make sure the input is not sorted in the expected order and
move the test to the appropriate sub folder.

### DIFF
--- a/pyanaconda/ui/common.py
+++ b/pyanaconda/ui/common.py
@@ -665,7 +665,7 @@ def collectCategoriesAndSpokes(paths, klass):
     ret = {}
     # Collect all the categories this hub displays, then collect all the
     # spokes belonging to all those categories.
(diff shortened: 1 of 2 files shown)


rvykydal (approved):
Looks good to me.

jkonecny12 (approved):
LGTM!

poncovka (approved):
Looks good to me. Thanks!

VladimirSlavik (approved):
Looks good to me. Thank you!

PR #3033, opened 2020-12-03
Remove the support for updates on a floppy disk

poncovka:
This is the last floppy-related code that we have in our code base.

### DIFF
--- a/dracut/parse-kickstart
+++ b/dracut/parse-kickstart
@@ -163,9 +163,7 @@ class DracutURL(Url, DracutArgsMixin):
 
 class DracutUpdates(Updates, DracutArgsMixin):
     def dracut_args(self, args, lineno, obj):
-        if self.url == "floppy":
-            return "live.updates=/dev/fd0"
-        elif self.url:
+        if self.url:
             return "live.updates=%s" % self.url
 
 class DracutMediaCheck(MediaCheck, DracutArgsMixin):


jkonecny12 (approved):
Looks good to me from the code PoV.
I wonder, are we sure that floppy disks are not used anymore. I mean some outpost somewhere in the middle of nowhere may still use these.

dracut/parse-kickstart
```diff
@@ -163,9 +163,7 @@ class DracutURL(Url, DracutArgsMixin):
 
 class DracutUpdates(Updates, DracutArgsMixin):
     def dracut_args(self, args, lineno, obj):
-        if self.url == "floppy":
-            return "live.updates=/dev/fd0"
-        elif self.url:
+        if self.url:
```
VladimirSlavik, line 166: What happens if "floppy" is there? Intentional error?

poncovka: There will be no officially supported way how to set `url` to `floppy`. Users were supposed to specify `updates` without any arguments to use the floppy disk and that won't be possible with the pykickstart update. It is possible to specify `updates floppy` in a kickstart file, but the documentation says that the argument should be a URL.

VladimirSlavik: OK, changing to ACK then.

I understand this as "a device can be there too but we are not telling users that". Which is fine.

VladimirSlavik (approved):

poncovka:
> I wonder, are we sure that floppy disks are not used anymore. I mean some outpost somewhere in the middle of nowhere may still use these.

I guess that users can still use `/dev/fd0`, so we might want to keep the kernel module loaded. Btw, the `updates` kickstart command is not documented at all in RHEL 8.

poncovka:
Pykickstart is fixed in: https://github.com/pykickstart/pykickstart/pull/361

jkonecny12:
> > I wonder, are we sure that floppy disks are not used anymore. I mean some outpost somewhere in the middle of nowhere may still use these.
> 
> I guess that users can still use `/dev/fd0`, so we might want to keep the kernel module loaded. Btw, the `updates` kickstart command is not documented at all in RHEL 8.

That may be on purpose. It's really just for debugging, nothing what users should use on daily basis.

poncovka:
The kernel module remains loaded.

jkonecny12 (approved):
Looks great to me now!

PR #3035, opened 2020-12-03
Fix missing inst. prefixes for bootloader options in documentation

jkonecny12:
We want to disable options without the 'inst.' prefix so our documentation should be correct.


poncovka (approved):
Looks good to me.

jstodola (approved):

VladimirSlavik (approved):
Looks good to me. Thank you!

PR #3036, opened 2020-12-03
Migrate Anaconda daily COPR builds to Packit

jkonecny12:
Packit should handle this for us.

Unfortunately, I don't know how to test this. So I guess we have to tune it here in the repository directly? I'll try to figure out something with my fork.

### DIFF
--- a/.packit.yml
+++ b/.packit.yml
@@ -21,3 +21,16 @@ jobs:
     metadata:
       targets:
         - fedora-eln
+
+  - job: copr_build
+    trigger: commit
+    metadata:
+      targets:
+        - fedora-rawhide
+        - fedora-eln
+      branch: master
+      owner: "@rhinstaller"
+      project: Anaconda
+      preserve_project: True
+      additional_repos:
+        - "copr://@storage/blivet-daily"


poncovka (approved):
Looks good to me.

martinpitt (approved):
I guess that's a bit difficult to test indeed. But the values [match our COPR](https://copr.fedorainfracloud.org/coprs/g/rhinstaller/Anaconda/), so let's try this and see how it works. Thanks!

jkonecny12:
I was able to test it on my fork, it wasn't that hard at the end.

You can see the results here:
https://github.com/jkonecny12/anaconda/commit/b9f045ef5f02985b007d6ebe36c93a94f3ccf1f3

Unfortunately, blivet does not support daily builds of ELN right now so it won't work with one configuration. @vojtechtrefny do you have ELN daily builds somewhere else or do you want me to add you a packit support for "commit" builds like we have here?

vojtechtrefny:
> Unfortunately, blivet does not support daily builds of ELN right now so it won't work with one configuration. @vojtechtrefny do you have ELN daily builds somewhere else

I've enabled ELN builds in our repo. Lets hope it will be more stable than last time I tried.

> or do you want me to add you a packit support for "commit" builds like we have here?

Is packit stable enough now? Builds on our PRs fail very often, I'm not sure if I want something like that for our daily builds too. Our existing solution works and that's good enough for now. Maybe someday when TODO list is emptier, I'll give a try.

jkonecny12:
> > Unfortunately, blivet does not support daily builds of ELN right now so it won't work with one configuration. @vojtechtrefny do you have ELN daily builds somewhere else
> 
> I've enabled ELN builds in our repo. Lets hope it will be more stable than last time I tried.
> 
> > or do you want me to add you a packit support for "commit" builds like we have here?
> 
> Is packit stable enough now? Builds on our PRs fail very often, I'm not sure if I want something like that for our daily builds too. Our existing solution works and that's good enough for now. Maybe someday when TODO list is emptier, I'll give a try.

Packit is pretty stable. What is failing is Testing Farm Cluster which should be migrated in January (last info). However, this is required only for running tests so if you are doing releases or COPR builds than it's pretty stable just some flakes from time to time but not really that problematic.

jkonecny12:
**UPDATED:**
* Add back ELN builds, seems to be [working correctly](https://github.com/jkonecny12/anaconda/commit/29f8b3200dbde3998e7c1f0fba29eeafe973430b)

rvykydal (approved):
Looks good to me.

VladimirSlavik (approved):
Looks good to me. Thank you!

PR #3038, opened 2020-12-04
Add bot PR comment to run kickstart tests

martinpitt:
Taking inspiration (and code) from the validate-rhel-8 workflow,
introduce a `/kickstart-tests <launch args>` magic comment, that must be
sent by an organization member (to guard running untrusted code on our
self-hosted runners).

This builds an update.img from the anaconda PR and runs it through the
usual kickstart-tests container launch script. The result is reported
through a "kickstart <...>" status.

### DIFF
--- a/.github/workflows/kickstart-tests.yml
+++ b/.github/workflows/kickstart-tests.yml
@@ -1,0 +1,131 @@
+# Run kickstart tests in a PR triggered by a "/kickstart-test <launch args>" comment from an organization member.
+name: kickstart-tests
+on:
+  issue_comment:
+    types: [created]
+
+jobs:
+  pr-info:
+    if: startsWith(github.event.comment.body, '/kickstart-test')
+    runs-on: ubuntu-latest
+    steps:
+      - name: Query comment author repository permissions
+        uses: octokit/request-action@v2.x
+        id: user_permission
+        with:
+          route: GET /repos/${{ github.repository }}/collaborators/${{ github.event.sender.login }}/permission
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      # restrict running of tests to users with admin or write permission for the repository
+      # see https://docs.github.com/en/free-pro-team@latest/rest/reference/repos#get-repository-permissions-for-a-user
+      # store output if user is allowed in allowed_user job output so it has to be checked in downstream job
+      - name: Check if user does have correct permissions
+        if: contains('admin write', fromJson(steps.user_permission.outputs.data).permission)
+        id: check_user_perm
+        run: |
+          echo "User '${{ github.event.sender.login }}' has permission '${{ fromJson(steps.user_permission.outputs.data).permission }}' allowed values: 'admin', 'write'"
+          echo "::set-output name=allowed_user::true"
+
+      - name: Get information for pull request
+        uses: octokit/request-action@v2.x
+        id: pr_api
+        with:
+          route: GET /repos/${{ github.repository }}/pulls/${{ github.event.issue.number }}
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Parse launch arguments
+        id: parse_launch_args
+        run: |
+          # extract first line and cut out the "/kickstart-tests" first word
+          LAUNCH_ARGS=$(echo '${{ github.event.comment.body }}' | sed -n '1 s/^[^ ]* *//p')
+          echo "::set-output name=launch_args::${LAUNCH_ARGS}"
+
+    outputs:
+      allowed_user: ${{ steps.check_user_perm.outputs.allowed_user }}
+      base_ref: ${{ fromJson(steps.pr_api.outputs.data).base.ref }}
+      sha: ${{ fromJson(steps.pr_api.outputs.data).head.sha }}
+      launch_args: ${{ steps.parse_launch_args.outputs.launch_args }}
+
+  run:
+    needs: pr-info
+    # only do this for master for now; once we have RHEL 8/9 boot.iso builds working, also support these
+    if: needs.pr-info.outputs.base_ref == 'master' && needs.pr-info.outputs.allowed_user == 'true' && needs.pr-info.outputs.launch_args != ''
+    runs-on: [self-hosted, kstest]
+    timeout-minutes: 300
+    steps:
+      # self-hosted runners don't do this automatically; also useful to keep stuff around for debugging
+      # need to run sudo as the launch script and the container create root/other user owned files
+      - name: Clean up previous run
+        run: |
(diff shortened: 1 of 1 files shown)


martinpitt:
I tested this on my test org:

 - https://github.com/mpitt-test-org/anaconda/pull/1 shows what happens with an invalid comment without a testtype (the actual test run just gets skipped), and with a valid testtype (I chose "keyboard" as that's the smallest one).
 - https://github.com/mpitt-test-org/anaconda/pull/2 shows what happens when a proposed change actually breaks stuff.

Note that I spun up a temporary "ghks-pitti" runner on upshift and attached it to my test org. That was both cleaner, and also I wouldn't know how to test the workflow otherwise, as such comment workflows must be on the target master branch already.

martinpitt:
FTR: This implements phase 1 of https://issues.redhat.com/browse/INSTALLER-2241 - after this lands and has worked for a bit, I'll look into building a real boot.iso.

VladimirSlavik (approved):
Yes please! This is the ultimate interface to kickstart tests.

martinpitt:
Hmm, considering the [breaking PR example](https://github.com/mpitt-test-org/anaconda/runs/1498315513?check_suite_focus=true): WDYT, is it right to do auto-retries for anaconda PRs? They are definitively right for nightly scheduled builds, and definitively wrong for kickstart-tests PRs, as we want to avoid introducing flaky tests. Anaconda is somewhat in the middle, but I think we should *not* retry tests for these, as it could also be that we have an Anaconda change that only works sometimes. So for now I'm going to drop the `--retry`, but of course feel free to put it back if you regularly run large test suites and random failures become too annoying.

.github/workflows/kickstart-tests.yml
```diff
@@ -0,0 +1,127 @@
...
+        run: git clone --depth=1 https://github.com/rhinstaller/kickstart-tests
+
+      - name: Ensure http proxy is running
+        run: |
+          [ -n "$(sudo podman ps -q -f 'name=^squid$')" ] || sudo kickstart-tests/containers/squid.sh start
+
+      - name: Run testtype ${{ needs.pr-info.outputs.testtype }} in container
+        working-directory: kickstart-tests
+        run: |
+          sudo TEST_JOBS=16 containers/runner/launch --testtype '${{ needs.pr-info.outputs.testtype }}' --skip-testtypes 'rhel-only,knownfailure' --updates ../updates.img
```
poncovka, line 106: I wanted to ask about the `knownfailure` tests, but you already took care of them :-)

poncovka (approved):
My dream came true! Thank you!

rvykydal (approved):
Looks great to me. Thank you.

rvykydal:
I have some use cases that others might find interesting as well (for a separate PR in the future maybe):
- Running a single test by name (or list of names) - for example I'd like to run the two tests for hostname we have but running the whole "network" group (~ 70-80 tests) would seem as resources wasting to me. Of course we could add tags to the tests instead, but still I think the option could be useful.
- Using specified kickstart-tests repository and branch - not sure if the interface is appropriate for such a level of configuration.

martinpitt:
Both is possible, we just need to find a way to encode this into the magic comment string.  We could make individual tests the default with `/kickstart-tests keyboard foo bar`, and run all tests of a type with `/kickstart-tests --testtype network`. In other words, the whole arguments would be passed to the containers/runner/launch script verbatim, after `--skip-testtypes 'rhel-only,knownfailure' --updates ../updates.img` (typing that all the time is clumsy and error-prone, so let's use that by default). Then you could also just say `all` to run all tests.

This would be much more powerful, but of course also more error prone. But if you typo an option, at least the test will fail pretty fast :grin: 

My gut feeling is that it can't hurt to run an entire suite to make sure that a proposed change does not regress, so I'd like to keep the "testtype" invocation. The above proposal would break the syntax, so maybe let's agree on that first before landing this PR?

The kstests repo/branch sounds interesting to co-develop something in anaconda with a test. Again we mostly need to invent a syntax for it. Maybe `/kickstart-tests @user/repo:branch [launch args...]`?

So the most simple invocation would still look like

    /kickstart-tests keyboard

A recommended one would be

    /kickstart-tests --testtype=coverage

And one with all bells and whistles would be

    /kickstart-tests @martinpitt/kickstart-tests:new-wifi-dialog --testtype=network 

@rvykydal , @poncovka , @jkonecny12 : Opinions?

jkonecny12:
I really like your idea! You have :+1:  from me ;)

jkonecny12:
> Hmm, considering the [breaking PR example](https://github.com/mpitt-test-org/anaconda/runs/1498315513?check_suite_focus=true): WDYT, is it right to do auto-retries for anaconda PRs? They are definitively right for nightly scheduled builds, and definitively wrong for kickstart-tests PRs, as we want to avoid introducing flaky tests. Anaconda is somewhat in the middle, but I think we should _not_ retry tests for these, as it could also be that we have an Anaconda change that only works sometimes. So for now I'm going to drop the `--retry`, but of course feel free to put it back if you regularly run large test suites and random failures become too annoying.

I don't think we want to have automatic retries here. It's simple to run the tests again manually if we are not sure.

.github/workflows/kickstart-tests.yml
```diff
@@ -0,0 +1,127 @@
...
+          git rebase origin/master
+
+      # FIXME: build a complete boot.iso with lorax/osbuild
+      - name: Build updates.img
+        run: |
+          scripts/makeupdates
+          gzip -cd updates.img | cpio -tv
+
+      - name: Check out kickstart-tests
+        run: git clone --depth=1 https://github.com/rhinstaller/kickstart-tests
```
jkonecny12, line 97: Please use checkout action here. I don't think there is a reason to call ``git clone`` manually.

jkonecny12 (changes requested):
Looks great to me. Just please use the ``checkout`` action to use standard tools here ;).

VladimirSlavik:
That sounds fantastic!

.github/workflows/kickstart-tests.yml
martinpitt: OK. I originally didn't because I don't want it to overwrite the anaconda checkout -- but I see that the action has a `path:` option. I'm happy to use that if you prefer.

martinpitt: Done.

martinpitt:
Reworked to accept generic launch options, as discussed above. I re-tested this on my test org again: https://github.com/mpitt-test-org/anaconda/pull/1 and https://github.com/mpitt-test-org/anaconda/pull/2

PTAL?

jkonecny12 (approved):
Looks great to me now!

jkonecny12:
The Packit failure is flake. From time to time they are missing some files. Not happening that often fortunately.

PR #3039, opened 2020-12-04
Disable the debug solver debugging

poncovka:
Never automatically enable the DNF debug solver debugging. It slows down the
installation and blocks the main thread, so the UI freezes for tens of seconds.
The debugging can be easily enabled with an updates image or we can introduce
a new option for that.


VladimirSlavik (approved):
Looks good to me. Thank you!

I wonder how many BZs are resolved by this, actually.

jkonecny12 (approved):
Hah, interesting. Thanks a lot for this fix!

rvykydal (approved):
Looks good to me.

M4rtinK (approved):
Looks good to me, thanks! :)

IIRC it has been a long time since the debug solver logs were needed for DNF debugging & we can indeed enable the debug solver via updates image if needed, so really no reason to keep this in. :)

AdamWill:
Ooh! Thanks for finally figuring this out :)

PR #3040, opened 2020-12-04
Stop trying to set the default hostname

keszybz:
Not tested. I'd like to make a change along the following lines:

If the user configured a hostname, then great, let's use it. If not,
let's just do nothing. This is totally appropriate, we know that
systemd sets some hostname when the machine is booted and we can just
keep using it.

In https://github.com/rhinstaller/anaconda/pull/643:
> To prevent fallouts from changed behaviour, we keep writing out
> /etc/hostname even when static hostname is not configured - using
> value "localhost.localdomain". In this case target system still
> applies transient hostname (same as if we didn't write
> /etc/hostname). I bet there is something relying on /etc/hostname
> existing after installation.

I want to drop this behaviour from systemd since it's just pointless
complicated. All the layers of guesswork mean that the whole system is
frozen in place. Instead of trying to second-guess the user and figure
out if the hostname was configured by the user or not based on the
specific string used, let's just write the config file only when the
user specifies a hostname.

### DIFF
--- a/pyanaconda/modules/network/installation.py
+++ b/pyanaconda/modules/network/installation.py
@@ -62,7 +62,7 @@ class HostnameConfigurationTask(Task):
     def run(self):
         _write_config_file(
             self._sysroot, self.HOSTNAME_CONF_FILE_PATH,
-            "{}\n".format(self._hostname),
+            "{}\n".format(self._hostname or ""),
             "Cannot write hostname configuration file",
             self._overwrite
         )
--- a/pyanaconda/modules/network/network.py
+++ b/pyanaconda/modules/network/network.py
@@ -61,7 +61,7 @@ class NetworkService(KickstartService):
         self._firewall_module = FirewallModule()
 
         self.hostname_changed = Signal()
-        self._hostname = "localhost.localdomain"
+        self._hostname = None
 
         self.current_hostname_changed = Signal()
         self._hostname_service_proxy = None
--- a/pyanaconda/network.py
+++ b/pyanaconda/network.py
@@ -47,8 +47,6 @@ from gi.repository import NM
 
 log = get_module_logger(__name__)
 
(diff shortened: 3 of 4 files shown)


rvykydal:
Thank you for the draft, seems like a reasonable change to me.
I'll look at it, check how the patch goes along with our UI for setting hostname and also how the unit tests should be updated with the patch.

jkonecny12:
Seems pretty interesting to me. I always welcome when we are reducing number of things Anaconda has to do!

pyanaconda/modules/network/network.py
```diff
@@ -61,7 +61,7 @@ def __init__(self):
         self._firewall_module = FirewallModule()
 
         self.hostname_changed = Signal()
-        self._hostname = "localhost.localdomain"
+        self._hostname = None
```
poncovka, line 64: Could you please change it to an empty string? `None` is not a valid DBus value and it can lead to errors.

pyanaconda/modules/network/installation.py
```diff
@@ -62,7 +62,7 @@ def name(self):
     def run(self):
         _write_config_file(
             self._sysroot, self.HOSTNAME_CONF_FILE_PATH,
-            "{}\n".format(self._hostname),
+            "{}\n".format(self._hostname or ""),
```
poncovka, line 65: Then this fix can be dropped.

poncovka (commented):
Otherwise, the code looks good to me.

rvykydal:
I've opened a PR with updates requied for UI and tests and also Vendy's suggestions for not using None value:
https://github.com/rhinstaller/anaconda/pull/3044

keszybz:
> I've opened a PR with updates requied for UI and tests and also Vendy's suggestions for not using None value:
> #3044

Great, thanks! Let's close this one then.

PR #3043, opened 2020-12-09
Import imp on demand

poncovka:
Import imp when we are actually going to use it to avoid deprecation warnings
from every module and addon that imports `pyanaconda.core.util`.

### DIFF
--- a/pyanaconda/core/util.py
+++ b/pyanaconda/core/util.py
@@ -30,7 +30,6 @@ import re
 import gettext
 import signal
 import sys
-import imp
 import types
 import inspect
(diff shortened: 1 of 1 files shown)


martinpitt (approved):
Looks good! That's the kind of feature that may benefit from a more wide-scale `--testtype coverage`, though? (You can trigger that in addition, as many as you like -- they get queued when running out of idle runners)

poncovka:
The commit message is fixed.

VladimirSlavik (approved):
Looks good to me. Thank you!

PR #3045, opened 2020-12-09
network: fix double-free using libnm function filter_connections()

thom311:
The libnm functions `nm_{device,access_pointer}_filter_connections()`
return ownership of a newly create GPtrArray, which also owns the
content by having the `g_ptr_array_set_free_func()` function set to
`g_object_unref()`.

Consequently, this functions should be annotate with "(transfer
container)" or "(transfer full)", but various bindings treat that
differently. To make pygobject happy, it would require "(transfer
container)", but that breaks vala (and vice versa).

This leads to a double-free and a crash in anaconda.

This is not really fixable in libnm. But the workaround is simple.
Let's just not use these functions.

See-also: https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/commit/b012877445bddd772eec79e3ce38f93cc40de720
See-also: https://gitlab.gnome.org/GNOME/gobject-introspection/-/issues/305

https://bugzilla.redhat.com/show_bug.cgi?id=1847868
https://bugzilla.redhat.com/show_bug.cgi?id=1873561

---

it was easy to reproduce the crash, especially when using https://copr.fedorainfracloud.org/coprs/networkmanager/NetworkManager-CI-master/, which has more assertions enabled. With this patch, I couldn't reproduce the crash anymore... but I still ran into other issues. So it is not clear to me whether this patch completely fixes all libnm related issues. What is also strange, is that this problem should exist at least since NetworkManager 1.16.0 (March 2019). Dunno why this was not discovered earlier...

### DIFF
--- a/pyanaconda/ui/gui/spokes/network.py
+++ b/pyanaconda/ui/gui/spokes/network.py
@@ -485,10 +485,16 @@ class NetworkControlBox(GObject.GObject):
             dialog.refresh(device_name)
             dialog.run()
 
+    def _device_filter_connections(self, device, connections):
+        return [c for c in connections if device.connection_valid(c)]
+
+    def _ap_filter_connections(self, ap, connections):
+        return [c for c in connections if ap.connection_valid(c)]
+
     def _get_the_only_wireless_connection(self, device):
         con_uuid = con_ssid = ""
         if device:
-            cons = device.filter_connections(self.client.get_connections())
+            cons = self._device_filter_connections(device, self.client.get_connections())
             if len(cons) == 1:
                 connection = cons[0]
                 con_uuid = connection.get_setting_connection().get_uuid()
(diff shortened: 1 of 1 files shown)


rvykydal:
Thank you for the patch Thomas, I'll look at it and try to test it locally.

pyanaconda/ui/gui/spokes/network.py
```diff
@@ -1185,7 +1191,7 @@ def refresh(self, device_name):
             log.warnig("device for interface %s not found", device)
             return
 
-        cons = device.filter_connections(self._nm_client.get_connections())
+        cons = self._device_filter_connections(device, self._nm_client.get_connections())
```
rvykydal, line 1194: This method is not defined in the class, but nevermind, I'll modify the patch a bit and submit it for rhel-8 branch. We can then port it to master after review.

rvykydal:
@thom311 I've opened PR for rhel-8 branch with your patch slightly modified.
Could you please review, and would you be ok with keeping you asauthor of the patch?

pyanaconda/ui/gui/spokes/network.py
thom311: right!

rvykydal:
Closing with PR porting modified fix from rhel-8 branch to master:
https://github.com/rhinstaller/anaconda/pull/3061

PR #3047, opened 2020-12-10
Build real boot.iso for /kickstart-tests runs

martinpitt:
Using an updates.img is really fast, but does not catch changes to
dracut or file removals. So move to building a boot.iso from Rawhide and
locally built anaconda RPMs, and using that.

Do the lorax build *after* checking out kickstart-tests and starting the
squid container, so that lorax already benefits from the HTTP caching
and seeds the cache for the subsequent kickstart tests.

Keep the old updates.img build in a comment, as that may come in handy
in the future -- we may want to control that via the magic comment to
select between fast and thorough?

-----

As before, I tested this on my test org on a [deliberate failure](https://github.com/mpitt-test-org/anaconda/pull/2), which proves that it's really using the new code; and on a [successful run](https://github.com/mpitt-test-org/anaconda/pull/1) which proves that all the boot.iso/kickstart-tests plumbing is working.

This adds 8 minutes to the test run to build boot.iso. This is certainly not a small price compared to the 15 mins that an average set of kicktart-tests needs, but it may be worth the trouble. Also, as explained in the commit message, there's always the option to select between the two approaches if we want that.

### DIFF
--- a/.github/workflows/kickstart-tests.yml
+++ b/.github/workflows/kickstart-tests.yml
@@ -113,7 +113,7 @@ jobs:
         if: always()
         uses: actions/upload-artifact@v2
         with:
-          name: 'logs-${{ matrix.scenario }}'
+          name: 'logs'
           # skip the /anaconda subdirectories, too large
           path: |
             kickstart-tests/data/logs/kstest.log
--- a/.github/workflows/kickstart-tests.yml
+++ b/.github/workflows/kickstart-tests.yml
@@ -88,12 +88,6 @@ jobs:
           git log --oneline -1 origin/master
           git rebase origin/master
 
-      # FIXME: build a complete boot.iso with lorax/osbuild
-      - name: Build updates.img
-        run: |
-          scripts/makeupdates
-          gzip -cd updates.img | cpio -tv
-
       - name: Check out kickstart-tests
         uses: actions/checkout@v2
         with:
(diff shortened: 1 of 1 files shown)


VladimirSlavik:
So, hm, what is the timing difference? Obviously if I want to run the full suite, it matters less than when running just one test...

martinpitt:
@VladimirSlavik : As I said, about 8 minutes. You can see that in the "Build boot.iso" step [in the workflow run](https://github.com/mpitt-test-org/anaconda/runs/1530890254?check_suite_focus=true)

.github/workflows/kickstart-tests.yml
```diff
@@ -88,26 +88,58 @@ jobs:
...
         uses: actions/checkout@v2
         with:
           repository: rhinstaller/kickstart-tests
           path: kickstart-tests
 
       - name: Ensure http proxy is running
+        run: sudo kickstart-tests/containers/squid.sh start
+
+      # This is really fast, but does not catch file removals or dracut changes
+      # maybe this becomes opt-in via a magic comment for efficiency reasons?
```
rvykydal, line 101: I think this option could be useful. Perhaps also as an easily accessible fallback when iso generating is broken for whatever reasons.

rvykydal (approved):
Looks good to me.

VladimirSlavik:
Oh. Sorry, apparently I'm blind :( 8 minutes looks good enough, given that kickstart tests still are the "slow" option. I agree with @rvykydal that a fast path might be nice, but I have no idea how hard it is to keep both.

PR #3048, opened 2020-12-10
Use Linux HOST_NAME_MAX hostname length limit

xqrustc2020:
Linux defines HOST_NAME_MAX is 64, so an error was reported when hostname was set more than 64 characters in anaconda.

### DIFF
--- a/tests/nosetests/pyanaconda_tests/network_test.py
+++ b/tests/nosetests/pyanaconda_tests/network_test.py
@@ -32,10 +32,6 @@ class NetworkTests(unittest.TestCase):
         self.assertTrue(network.is_valid_hostname("h"*63)[0])
         self.assertFalse(network.is_valid_hostname("h"*64)[0])
 
-        # length < 256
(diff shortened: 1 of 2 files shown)


tests/nosetests/pyanaconda_tests/network_test.py
```diff
@@ -32,10 +32,6 @@ def is_valid_hostname_test(self):
         self.assertTrue(network.is_valid_hostname("h"*63)[0])
         self.assertFalse(network.is_valid_hostname("h"*64)[0])
 
-        # length < 256
-        self.assertTrue(network.is_valid_hostname("section." * 31+"section")[0])
```
VladimirSlavik, line 36: To make the tests pass, simply adjust the counts.

Length of the `section.` string is 8 characters, so the previous situation was 8*31 + 7 = 255 which was ok, and with the one "x" more it was 256 which was expected to fail.

If you want 64 instead of 256, just reduce the 31 to 7: 8*7 + 7 = 63, and one more is 64.

VladimirSlavik (commented):
This almost worked!

PR #3049, opened 2020-12-10
Use Linux HOST_NAME_MAX hostname length limit

xqrustc2020:
Linux defines HOST_NAME_MAX is 64, so an error was reported when hostname was set more than 64 characters in anaconda.

### DIFF
--- a/pyanaconda/network.py
+++ b/pyanaconda/network.py
@@ -109,8 +109,8 @@ def is_valid_hostname(hostname, local=False):
     if not hostname:
         return (False, _("Host name cannot be None or an empty string."))
 
-    if len(hostname) > 255:
-        return (False, _("Host name must be 255 or fewer characters in length."))
+    if len(hostname) > 64:
+        return (False, _("Host name must be 64 or fewer characters in length."))
 
     if local and hostname[-1] == ".":
         return (False, _("Local host name must not end with period '.'."))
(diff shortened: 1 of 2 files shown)


VladimirSlavik:
@rvykydal , can you please take a look at this? Networking is mostly your domain...

xqrustc2020:
> @xqrustc2020 , thank you for the contribution! @rvykydal, can you please take a look at this? Networking is mostly your domain..

VladimirSlavik:
It seems that this value could have both limits in practice, so I'm no longer sure...

https://stackoverflow.com/questions/8724954/what-is-the-maximum-number-of-characters-for-a-host-name-in-unix

rvykydal (approved):
Looks good to me.
We will become consistent with hostnamectl and systemd-firstboot tools.

poncovka (approved):
The code looks good to me.

rvykydal:
Rebasing to f34-devel so the patch gets into F34.

VladimirSlavik:
For the record:
* This applies also to kickstart due to #3135
* I considered changing also the GUI input to limit the length, but decided against that. I think it's better to have it as now, free-form with a detailed error message, rather than just "suddenly can't type any more".

rvykydal:
>     * I considered changing also the GUI input to limit the length, but decided against that. I think it's better to have it as now, free-form with a detailed error message, rather than just "suddenly can't type any more".

Agreed.

PR #3050, opened 2020-12-10
Add support to run RPM tests in a container

jkonecny12:
This new  make command will do this:

- build RPM files in a container
- run rpm-tests
- test installation of these RPM files

This commit is extending run-build-and-arg script because it's not reasonably
possible to put multiple commands from Makefile in the podman. To avoid this
issue it's adding --build-rpms as first argument which when detected will build
RPM and test installation before the command execution.

Also fix issues that nosetests-3 executable is not available in the container.

*Related: rhbz#1885635*

This is first part of the ``rpm-test`` support for rhel-8. I won't be able to do more this year so more to come ;)

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -91,6 +91,9 @@ CI_TAG ?= $(GIT_BRANCH)
 CI_TEST_ARGS ?= --rm -v $(srcdir):/anaconda:Z
 CI_CMD ?= make ci
 
+# --build-rpm will be processed by run-build-and-arg script to build rpm before command execution
+RPM_TEST_CMD ?= --build-rpms make run-rpm-tests-only
+
 SKIP_BRANCHING_CHECK ?= "false"
 
 # If translations are present, run tests on the .po files before tarring them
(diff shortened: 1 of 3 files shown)


VladimirSlavik:
I don't understand this enough to meaningfully ack, but LGTM.

martinpitt:
TBH I don't like this approach much -- the anaconda-ci container is "too big", it has too many packages installed. This breaks the idea of this RPM test to make sure that all the build and runtime requires are sufficient and available. We should either build a second "anaconda-rpm" test container, or just use a `fedora:rawhide` container. The latter will be much simpler to do and avoid the overhead of maintaining another image, for the price of  taking a bit longer to run locally -- but I don't think that people actually run it manually that often.

I'll send a PR against master soon with a proposal.

martinpitt:
I sent PR #3054 with an alternate proposal which I think is more uniform between master and rhel, and more correct.

jkonecny12:
Replaced by https://github.com/rhinstaller/anaconda/pull/3054.

VladimirSlavik:

PR #3052, opened 2020-12-11
network: fix double-free using libnm function filter_connections()

rvykydal:
The libnm functions nm_{device,access_pointer}_filter_connections()
return ownership of a newly create GPtrArray, which also owns the
content by having the g_ptr_array_set_free_func() function set to
g_object_unref().

Consequently, this functions should be annotate with "(transfer
container)" or "(transfer full)", but various bindings treat that
differently. To make pygobject happy, it would require "(transfer
container)", but that breaks vala (and vice versa).

This leads to a double-free and a crash in anaconda.

This is not really fixable in libnm. But the workaround is simple.
Let's just not use these functions.

See-also: https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/commit/b012877445bddd772eec79e3ce38f93cc40de720
See-also: https://gitlab.gnome.org/GNOME/gobject-introspection/-/issues/305

Resolves: rhbaz#1873561

### DIFF
--- a/pyanaconda/ui/gui/spokes/network.py
+++ b/pyanaconda/ui/gui/spokes/network.py
@@ -485,7 +485,7 @@ class NetworkControlBox(GObject.GObject):
     def _get_the_only_wireless_connection(self, device):
         con_uuid = con_ssid = ""
         if device:
-            cons = device.filter_connections(self.client.get_connections())
+            cons = _safe_device_filter_connections(device, self.client.get_connections())
             if len(cons) == 1:
                 connection = cons[0]
                 con_uuid = connection.get_setting_connection().get_uuid()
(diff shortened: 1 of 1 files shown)


thom311:
Maybe add  a `Co-authored-by: Radek Vykydal <rvykydal(at)redhat.com>` to the commit message. Or change the patch's main author and add "Based-on-patch-by: Thomas Haller <thaller(at)redhat.com>`.

Patch lgtm, of course. I did not test.

poncovka (approved):
Looks good to me.

rvykydal:
> Maybe add a `Co-authored-by: Radek Vykydal <rvykydal(at)redhat.com>` to the commit message. Or change the patch's main author and add "Based-on-patch-by: Thomas Haller <thaller(at)redhat.com>`.

Did the latter, thank you.

rvykydal:
PR with port to Fedora: https://github.com/rhinstaller/anaconda/pull/3061

PR #3053, opened 2020-12-11
Move to official fedora ELN container image

martinpitt:
quay.io/fedoraci/fedora is the official ELN image as per [1], so use
that instead of a cross-grade from rawhide.

That container image only has BaseOS and AppStream repositories
available unfortunately, so we still need to add our eln.repo file.

[1] https://docs.fedoraproject.org/en-US/eln/deliverables/

----

[Tested on my branch](https://github.com/martinpitt/anaconda/actions/runs/415131425)

### DIFF
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -9,7 +9,7 @@ jobs:
         release: ['', 'eln']
         include:
           - release: eln
-            build-args: '--build-arg=eln=1'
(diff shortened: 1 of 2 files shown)


VladimirSlavik (approved):
Looks good to me. Thank you!

martinpitt:
PR #3054 also uses that new approach, and @jkonecny12 approved it there. As this blocks some further work, I'm landing this now.

PR #3054, opened 2020-12-11
Define container for running rpm-tests

martinpitt:
Define and document the  "anaconda-rpm" container similar to the
"anaconda-ci" one for running the RPM tests. We don't want to use the
anaconda-ci one as that has too many packages pre-installed, and thus we
would not detect missing dependencies.

Fix the workflow to rebase the branch to current master just as the
unit tests do.

-----

This is *prepared* for running for ELN, but the workflow does not yet enable it. I'd first like to get @jkonecny12 's and others opinion whether you like this direction. I think it's better than PR #3050 as it uses a smaller environment, which is quite important for this particular test. It should also port well to the rhel8 branch (I'll do that once this lands).

Note that for ELN it already uses the official image, like I do in PR #3053 .

Successfully [tested on my fork](https://github.com/martinpitt/anaconda/runs/1536764299?check_suite_focus=true).

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -88,7 +88,9 @@ CONTAINER_REGISTRY ?= quay.io
 
 # anaconda-ci container
 CI_DOCKERFILE ?= $(srcdir)/dockerfile/anaconda-ci
+RPM_DOCKERFILE ?= $(srcdir)/dockerfile/anaconda-rpm
 CI_NAME ?= $(CONTAINER_REGISTRY)/rhinstaller/anaconda-ci
+RPM_NAME ?= $(CONTAINER_REGISTRY)/rhinstaller/anaconda-rpm
 CI_TAG := $(or $(CI_TAG),$(GIT_BRANCH))
 CI_TEST_ARGS ?= --rm -v $(srcdir):/anaconda:Z
 CI_CMD ?= make ci
@@ -178,6 +180,9 @@ anaconda-ci-build:
 anaconda-ci-push:
 	$(CONTAINER_ENGINE) push $(CI_NAME):$(CI_TAG)
 
+anaconda-rpm-build:
+	$(CONTAINER_ENGINE) build $(CONTAINER_BUILD_ARGS) -t $(RPM_NAME):$(CI_TAG) $(RPM_DOCKERFILE)
+
 bumpver: po-pull
 	@opts="-n $(PACKAGE_NAME) -v $(PACKAGE_VERSION) -r $(PACKAGE_RELEASE) -b $(PACKAGE_BUGREPORT)" ; \
 	if [ ! -z "$(IGNORE)" ]; then \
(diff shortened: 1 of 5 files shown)


Makefile.am
```diff
@@ -233,6 +238,13 @@ container-ci:
 container-shell:
 	$(CONTAINER_ENGINE) run -it $(CONTAINER_TEST_ARGS) $(CI_TEST_ARGS) $(CI_NAME):$(CI_TAG)
 
+container-rpm-test:
+	$(CONTAINER_ENGINE) run -it $(CONTAINER_TEST_ARGS) $(CI_TEST_ARGS) -v $(CI_DOCKERFILE)/run-build-and-arg:/run-build-and-arg $(RPM_NAME):$(CI_TAG) sh -exc ' \
+	    /run-build-and-arg make rpms; \
+	    cd /tmp/anaconda; \
+	    make run-rpm-tests-only; \
+	    dnf install -y result/build/01-rpm-build/*.rpm'
```
jkonecny12, line 246: I've tried to add this as Makefile variable the same way as we have for ``container-ci`` target but I failed there because it always lost quotes along the way. However, I must say this works pretty well too.

It's really unfortunate that with bash you are still fighting with quoting :(.

dockerfile/anaconda-rpm/Dockerfile
```diff
@@ -0,0 +1,20 @@
+ARG image=docker.io/fedora:rawhide
+FROM ${image}
```
jkonecny12, line 2: This is pretty nice solution. :clap:

Makefile.am
```diff
@@ -178,6 +180,9 @@ anaconda-ci-build:
 anaconda-ci-push:
 	$(CONTAINER_ENGINE) push $(CI_NAME):$(CI_TAG)
 
+anaconda-rpm-build:
+	$(CONTAINER_ENGINE) build $(CONTAINER_BUILD_ARGS) -t $(RPM_NAME):$(CI_TAG) $(RPM_DOCKERFILE)
```
jkonecny12, line 184: I'm missing here possibility to make advanced of the ``ARG`` definition from the Dockerfile.

jkonecny12 (approved):
I agree with this solution more than with mine. It's definitely cleaner than my and it's also more precise because of not installing everything (although much of the dependencies are same in the end because you have to install them for the build).

However, I think this should have been split to at least 2 commits but I won't block this on such a nitpick. I also have a few notes below.

jkonecny12:
Also this is my last review this year, so don't expect me to react on the answers on Monday :).

Makefile.am
martinpitt: Good point. I'll have to do that anyway in a follow-up to also run rpm-tests for ELN. But for that I'd really like to land #3053 first, so that we have some consistency between the two tests.

martinpitt: Actually, no -- this is exactly what `$(CONTAINER_BUILD_ARGS)` is for, and how we use it for the unit-tests ELN run.

PR #3056, opened 2020-12-14
Reorganize the error handler

poncovka:
We are not able to use the error handler in the DBus modules, so move
it from the low-level code to the top level if possible, so we don't have to
deal with it during the Anaconda modularization.

Remove error handlers for deprecated or invalid use cases and clean up
the code.

### DIFF
--- a/pyanaconda/core/users.py
+++ b/pyanaconda/core/users.py
@@ -25,7 +25,6 @@ from contextlib import contextmanager
 from pyanaconda.core import util
 from pyanaconda.core.configuration.anaconda import conf
 from pyanaconda.core.util import strip_accents
-from pyanaconda.errors import errorHandler, PasswordCryptError, ERROR_RAISE
 from pyanaconda.core.regexes import GROUPLIST_FANCY_PARSE, NAME_VALID, PORTABLE_FS_CHARS, GROUPLIST_SIMPLE_VALID
 import crypt
 from pyanaconda.core.i18n import _
@@ -45,10 +44,12 @@ def crypt_password(password):
     :rtype: str
     """
     cryptpw = crypt.crypt(password, crypt.METHOD_SHA512)
+
     if cryptpw is None:
-        exn = PasswordCryptError(algo=crypt.METHOD_SHA512)
-        if errorHandler.cb(exn) == ERROR_RAISE:
-            raise exn
+        raise RuntimeError(_(
+            "Unable to encrypt password: unsupported "
+            "algorithm {}").format(crypt.METHOD_SHA512)
+        )
 
     return cryptpw
 
--- a/pyanaconda/errors.py
+++ b/pyanaconda/errors.py
@@ -23,12 +23,6 @@ from pyanaconda.modules.common.errors.installation import BootloaderInstallation
 from pyanaconda.modules.common.errors.storage import UnusableStorageError
 
 
-class InvalidImageSizeError(Exception):
-    def __init__(self, message, filename):
-        Exception.__init__(self, message)
-        self.filename = filename
-
-
 class ScriptError(Exception):
     def __init__(self, lineno, details):
         Exception.__init__(self)
@@ -116,20 +110,6 @@ class ErrorHandler(object):
                    C_("GUI|Storage Detailed Error Dialog", "_Retry"))
         if self.ui.showDetailedError(message, details, buttons=buttons):
             return ERROR_RETRY
-        else:
-            return ERROR_RAISE
-
-    def _invalidImageSizeHandler(self, exn):
-        message = _("The ISO image %s has a size which is not "
-                    "a multiple of 2048 bytes.  This may mean "
-                    "it was corrupted on transfer to this computer."
-                    "\n\n"
-                    "It is recommended that you exit and abort your "
-                    "installation, but you can choose to continue if "
-                    "you think this is in error. Would you like to "
-                    "continue using this image?") % exn.filename
-        if self.ui.showYesNoQuestion(message):
-            return ERROR_CONTINUE
         else:
             return ERROR_RAISE
 
@@ -252,7 +232,6 @@ class ErrorHandler(object):
         _map = {
             StorageInstallationError.__name__: self._storage_install_handler,
             UnusableStorageError.__name__: self._storage_reset_handler,
-            "InvalidImageSizeError": self._invalidImageSizeHandler,
             "NoSuchGroup": self._noSuchGroupHandler,
             "NoStreamSpecifiedException": self._no_module_stream_specified,
             "InstallMoreStreamsException": self._multiple_module_streams_specified,
--- a/pyanaconda/errors.py
+++ b/pyanaconda/errors.py
@@ -113,29 +113,6 @@ class ErrorHandler(object):
         else:
             return ERROR_RAISE
 
-    def _noSuchGroupHandler(self, exn):
-        if exn.required:
-            message = _("The group '%s' is required for this installation. "
-                        "This group does not exist. This is a fatal error and "
-                        "installation will be aborted.") % exn.group
-            self.ui.showError(message)
-            return ERROR_RAISE
-        elif exn.adding:
-            message = _("You have specified that the group '%s' should be "
-                        "installed.  This group does not exist.  Would you like "
-                        "to ignore this group and continue with "
-                        "installation?") % exn.group
-        else:
-            message = _("You have specified that the group '%s' should be "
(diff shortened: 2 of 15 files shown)


VladimirSlavik:
Code looks far better, but won't our users miss all the wonderful detail in error popups?

pyanaconda/payload/dnf/utils.py
```diff
@@ -32,16 +30,9 @@
 
 DNF_PACKAGE_CACHE_DIR_SUFFIX = 'dnf.package.cache'
 YUM_REPOS_DIR = "/etc/yum.repos.d/"
-
 USER_AGENT = "%s (anaconda)/%s" % (productName, productVersion)
 
 
-def go_to_failure_limbo():
```
jkonecny12, line 39: Wouldn't this change just kill the Anaconda with a traceback instead of showing GUI error to user?

pyanaconda/payload/rpmostreepayload.py
```diff
@@ -247,12 +243,7 @@ def _install(self, data):
                                    Variant('a{sv}', pull_opts),
                                    progress, cancellable)
         except GError as e:
-            exn = PayloadInstallError("Failed to pull from repository: %s" % e)
-            log.error(str(exn))
-            if errors.errorHandler.cb(exn) == errors.ERROR_RAISE:
-                progressQ.send_quit(1)
-                util.ipmi_abort(scripts=self.data.scripts)
```
M4rtinK, line 254: Is dropping the [IPMI](https://en.wikipedia.org/wiki/Intelligent_Platform_Management_Interface) call intentional here (eq. the task will do the IPMI calls by itself) ?

pyanaconda/errors.py
```diff
@@ -106,8 +106,6 @@ def _get_default_mapping(self):
             PayloadInstallError.__name__: self._payload_install_handler,
 
             # DNF errors
-            "NoStreamSpecifiedException": self._no_module_stream_specified,
-            "InstallMoreStreamsException": self._multiple_module_streams_specified,
```
jkonecny12: It would be great if you link the DNF changes to the commit message but if it would be pain to find those just leave it as it is now. Thanks!

jkonecny12 (changes requested):
Great cleanup, thanks a lot for this!

I have a few question below which I would like to have answered before merge.

pyanaconda/errors.py
M4rtinK: Hmm, could have been something that was planned but never implemented/handled in a different manner for modularity in the end. Should be safe to drop.

M4rtinK (approved):
Looks good to me - nice cleanup & removal of effectively dead code in hiding! :)

pyanaconda/payload/rpmostreepayload.py
jkonecny12: If I'm not mistaken, IPMI calls should definitely stay there.

pyanaconda/core/users.py
```diff
@@ -45,10 +44,12 @@ def crypt_password(password):
     :rtype: str
     """
     cryptpw = crypt.crypt(password, crypt.METHOD_SHA512)
+
     if cryptpw is None:
-        exn = PasswordCryptError(algo=crypt.METHOD_SHA512)
```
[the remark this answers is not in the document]
poncovka: The commit message was changed.

pyanaconda/payload/dnf/payload.py
```diff
@@ -1244,7 +1244,7 @@ def _find_and_mount_iso(self, device, device_mount_dir, iso_path, iso_mount_dir)
         if not os.path.ismount(iso_mount_dir):
             # mount the ISO on a loop
             image = os.path.normpath("%s/%s" % (path, image))
-            mountImage(image, iso_mount_dir)
+            payload_utils.mount(image, iso_mount_dir, fstype='iso9660', options="ro")
```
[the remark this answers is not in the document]
poncovka: The commit message was fixed.

pyanaconda/errors.py
```diff
@@ -119,20 +113,6 @@ def _storage_reset_handler(self, exn):
...
 
-    def _invalidImageSizeHandler(self, exn):
-        message = _("The ISO image %s has a size which is not "
-                    "a multiple of 2048 bytes.  This may mean "
-                    "it was corrupted on transfer to this computer."
-                    "\n\n"
-                    "It is recommended that you exit and abort your "
-                    "installation, but you can choose to continue if "
-                    "you think this is in error. Would you like to "
-                    "continue using this image?") % exn.filename
```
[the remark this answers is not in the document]
poncovka: That would be great, but I don't think that it is a good idea to store unused strings in our code base. It is not likely that we will need the same strings in the future. Also, do we want people to translate strings that we keep just in case?

pyanaconda/errors.py
```diff
@@ -113,29 +113,6 @@ def _storage_reset_handler(self, exn):
...
-            message = _("You have specified that the group '%s' should be "
-                        "excluded from installation.  This group does not exist.  "
-                        "Would you like to ignore this group and continue with "
-                        "installation?") % exn.group
-
-        if self.ui.showYesNoQuestion(message):
-            return ERROR_CONTINUE
-        else:
-            return ERROR_RAISE
-
```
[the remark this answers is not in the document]
poncovka: Yes, we do.

pyanaconda/errors.py
```diff
@@ -113,29 +113,6 @@ def _storage_reset_handler(self, exn):
         else:
             return ERROR_RAISE
 
-    def _noSuchGroupHandler(self, exn):
-        if exn.required:
-            message = _("The group '%s' is required for this installation. "
-                        "This group does not exist. This is a fatal error and "
-                        "installation will be aborted.") % exn.group
```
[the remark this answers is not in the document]
poncovka: Yes, we are. See https://github.com/rhinstaller/anaconda/pull/3056#discussion_r551330128.

pyanaconda/payload/dnf/utils.py
poncovka: The traceback will be processed in the installation task (from `pyanaconda.installation_tasks`).

pyanaconda/payload/rpmostreepayload.py
poncovka: Fix in progress...

pyanaconda/payload/live/payload_liveos.py
```diff
@@ -70,9 +69,7 @@ def setup(self):
 
         osimg_path = payload_utils.get_device_path(osimg)
         if not stat.S_ISBLK(os.stat(osimg_path)[stat.ST_MODE]):
-            exn = PayloadSetupError("{} is not a valid block device".format(osimg_spec))
-            if errorHandler.cb(exn) == ERROR_RAISE:
-                raise exn
+            raise PayloadSetupError("{} is not a valid block device".format(osimg_spec))
```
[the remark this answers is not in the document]
poncovka: Fix in progress...

pyanaconda/payload/rpmostreepayload.py
M4rtinK: Thanks! :)

pyanaconda/errors.py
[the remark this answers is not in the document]
VladimirSlavik: At least for me, the real issue is that we will now have "generic error" instead of useful details. Keeping the strings is meaningless if they are not used.

jkonecny12 (changes requested):
Looks good to me, except that I wasn't able to spot the removed IPMI calls replacement.

pyanaconda/errors.py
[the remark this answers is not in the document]
poncovka: I have checked to code and there is no easy way how to propagate this error to the user. We cannot really keep this error handler, because it won't work after the modularization. A DBus task cannot and shouldn't wait for a user input. I have moved part of the error message to the warning log and I can create an issue for improving the user experience, but that is out of scope of this PR.

pyanaconda/payload/rpmostreepayload.py
poncovka: I have fixed some places where the IPMI events where not reported before the exit, but in general, the exit handling is a mess. I have created a new issue and described some of the issues, but the fixes are out of scope of this PR.

poncovka:
All issues should be resolved and the code is tested.

jkonecny12 (approved):
Great work! Thanks a lot for this!

PR #3063, opened 2021-01-05
Change absolute shebangs to /usr/bin/env equivalents

Epsilqn:
Current shebangs for executed files gave a direct path to where they would be located on most systems. With the current official Anaconda targets, this is not an issue. Some distributions however place their installations of Bourne, BASH, and Python at other locations. They may not have /bin symlinked to /usr/bin or may install in an entirely different location. /usr/bin/env is more consistent as it can support wherever an interpreter is located no matter the target system, and env is found at /usr/bin/env on almost every distribution.

### DIFF
--- a/anaconda.py
+++ b/anaconda.py
@@ -1,4 +1,4 @@
-#!/usr/bin/python3
+#!/usr/bin/env python3
 #
 # anaconda: The Red Hat Linux Installation program
 #
--- a/dracut/driver_updates.py
+++ b/dracut/driver_updates.py
@@ -1,4 +1,4 @@
-#!/usr/bin/python3
+#!/usr/bin/env python3
 #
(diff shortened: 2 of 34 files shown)


Epsilqn:
The PEP8 issue has been corrected.

jkonecny12:
Thank you for your contribution but we had ``/usr/bin/env python`` in the past and it was removed because of Fedora required to do that.

For more info please read [here](https://fedoraproject.org/wiki/Features/SystemPythonExecutablesUseSystemPython).

Closing this PR now.

jkonecny12:
Re-opening this PR for the further discussion. This can't be merged in this shape but maybe there is another solution to solve this.

Can you please tell us what distributions are you talking about? Are they already using Anaconda or they want to do that?

jkonecny12:
No problem at all :).

PR #3064, opened 2021-01-05
set native layout for virtual console

sundeep-anand:
the console layout configured should be "native" by default,
setting that explicitly for non-ascii layouts where we prepend "us"

refer: https://bugzilla.redhat.com/show_bug.cgi?id=1912609

### DIFF
--- a/pyanaconda/keyboard.py
+++ b/pyanaconda/keyboard.py
@@ -183,6 +183,12 @@ def set_x_keyboard_defaults(localization_proxy, xkl_wrapper):
     if conf.system.can_configure_keyboard:
         xkl_wrapper.replace_layouts(new_layouts)
 
+    # the console layout configured should be "native" by default,
+    # setting that explicitly for non-ascii layouts where we prepend "us"
+    # refer: https://bugzilla.redhat.com/show_bug.cgi?id=1912609
+    if len(new_layouts) >= 2 and not langtable.supports_ascii(new_layouts[1]):
+        localization_proxy.SetVirtualConsoleKeymap(new_layouts[1])
+
     if len(new_layouts) >= 2 and not localization_proxy.LayoutSwitchOptions:
         # initialize layout switching if needed
         localization_proxy.SetLayoutSwitchOptions(["grp:alt_shift_toggle"])


AdamWill:
I *think* this should do the trick. This stuff is way too complicated, though :( Wish something had come of the various plans to use xkb for consoles...

VladimirSlavik:
I think I either fail to verify this works, or it does not.

I booted into installer, set Russian, named my user vslavik (in ASCII), switched to Russian keyboard and entered my password, using muscle memory and keyboard markings. The cursed eye thing shows cyrillic, so I think that's as wanted. After reboot, I can enter my user name just fine, but apparently not the password. Of course I can't really claim I did enter the right thing, given I did not see the password either time, but I went with a pattern I have been typing for almost two decades, so I give it 99% it was right.

On a tangent, I can't read cyrillic, so I simply mash the right buttons. There was a warning displayed for the password entry, but I think it just said I might have problems with non-ASCII passwords? It was of the "Done twice to keep going" type.

sundeep-anand:
hey @VladimirSlavik ,
thanks for giving it a try! when I tried - I had given root password in Russian. After reboot, it was `us` by default, so given username as **root**, switched to `ru` (_by `ctrl + shift`_)  and entered my password. It worked. Switched again to `us` and ran `localectl`.

sundeep-anand:
hey @VladimirSlavik ,
thanks for giving it a try! when I tried - I had given root password in Russian. After reboot, it was `us` by default, so given username as **root**, switched to `ru` (_by `ctrl + shift`_)  and entered my password. It worked. Switched again to `us` and ran `localectl`.

VladimirSlavik:
So, it seems tests pass ok, and Sundeep says it works for him, while it just won't work for me. Let's make sure all the rules are ok, and merge, so I'm not the one blocking this.

VladimirSlavik (approved):

AdamWill:
@VladimirSlavik it would be easier to test while typing the *username*, no, so you can see what you're typing? Note that as @sundeep-anand says, when you first boot with the 'ru' console layout, it's set to Latin "mode". You have to hit ctrl+shift to switch to entering Cyrillic characters. That's normal.

VladimirSlavik:
@AdamWill indeed, that's one of the many things I tried since the first comment. No luck.

PR #3065, opened 2021-01-05
Fix python3 rpm package name in spec file

jkonecny12:
Python3 bindings for rpm module are named python3-rpm as all other python packages. However, this package also provides rpm-python3, most probably for backward compatibility.


poncovka (approved):
Thanks!

VladimirSlavik (approved):
Looks good to me. Thank you!

M4rtinK (approved):
Looks good to me as well.

PR #3066, opened 2021-01-05
Provide a hint for the desired capacity

poncovka:
Clarify that it is possible to enter the size with a unit or that we will use
MiB if the unit is not specified.

Suggested-by: Mikhail Novosyolov <m.novosyolov@rosalinux.ru>

There is a new placeholder in the Desired Capacity entry:
desired_capacity_01

A new tooltip appears when you keep the mouse above the Desired Capacity entry:
desired-capacity-03

### DIFF
--- a/pyanaconda/ui/gui/spokes/lib/custom_storage_helpers.py
+++ b/pyanaconda/ui/gui/spokes/lib/custom_storage_helpers.py
@@ -49,7 +49,7 @@ CONTAINER_TOOLTIP = N_("Create or select %(container_type)s")
 CONTAINER_DIALOG_TITLE = N_("CONFIGURE %(container_type)s")
 CONTAINER_DIALOG_TEXT = N_("Please create a name for this %(container_type)s "
                            "and select at least one disk below.")
-DESIRED_CAPACITY_ERROR = N_(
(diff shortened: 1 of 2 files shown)


pyanaconda/ui/gui/spokes/lib/custom_storage_helpers.py
```diff
@@ -60,6 +60,8 @@
     "'123456789' = 123 terabytes and a bit less than a half\n"
 )
 
+DESIRED_CAPACITY_ERROR = DESIRED_CAPACITY_HINT
```
VladimirSlavik, line 63: I think the other usage is mine, and I'm happy with the rename, if that helps.

VladimirSlavik (approved):
Looks good to me. Thank you!

pyanaconda/ui/gui/spokes/lib/custom_storage_helpers.py
poncovka: I was considering it, but `DESIRED_CAPACITY_ERROR` makes more sense in the context of the validation error. It might be useful to have both constants.

M4rtinK (approved):
Looks good to me as well, nice improvement! :)

PR #3067, opened 2021-01-05
network: handle wireless configure button sensitiveness

rvykydal:
When configuring a wireless device (clicking Configure button) without any
access point selected before (by clicking Select network) we used to present
the dialog with empty list of access points. With the patch make the Configure
button sensitive only if there already is some access point to be selected.

Related: rhbz#1847681


jkonecny12 (approved):
LGTM!

rvykydal:
Updated RHBZ reference.

rvykydal:
PR with port to Fedora: https://github.com/rhinstaller/anaconda/pull/3098

PR #3069, opened 2021-01-06
Initialize librepo logger (#1908286)

jkonecny12:
Due to a change in dnf and libdnf, librepo logs are now redirected to "librepo" logger in python, so a handler needs to be added.

Resolves: rhbz#1908286

### DIFF
--- a/pyanaconda/anaconda_logging.py
+++ b/pyanaconda/anaconda_logging.py
@@ -40,6 +40,7 @@ ANACONDA_SYSLOG_FORMAT = "anaconda: %(log_prefix)s: %(message)s"
 MAIN_LOG_FILE = "/tmp/anaconda.log"
 PROGRAM_LOG_FILE = "/tmp/program.log"
 PACKAGING_LOG_FILE = "/tmp/packaging.log"
+LIBREPO_LOG_FILE = "/tmp/dnf.librepo.log"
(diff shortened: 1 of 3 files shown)


pyanaconda/anaconda_logging.py
```diff
@@ -216,6 +217,13 @@ def __init__(self, write_to_journal=False):
                             minLevel=logging.NOTSET)
         self.forwardToJournal(dnf_logger)
 
+        # Create the librepo logger.
+        librepo_logger = logging.getLogger(constants.LOGGER_LIBREPO)
```
poncovka, line 221: Are the `libdnf.repo.LibrepoLog` calls in the DNF payload still valid?

jkonecny12: Yes, they are still necessary. Reason for that is:

> pkratoch: it still needs to be there in libdnf, so that the logs from librepo get the callback that writes them into the libdnf log (with different source), so that dnf can then process them
> pkratoch: it's a bit tangled, but this was the easiest way to unite the librepo log handling with other logs, so that the log rotation is consistent

jkonecny12: Updated commit message to add explanation.

poncovka (approved):
Looks good to me. Thanks!

PR #3070, opened 2021-01-06
Rhel 8 disable some modules based on configuration

rvykydal:
Example of configuration disabling the modules:
`data/conf.d/09-rhel-image-disable-modules.conf`:
```
# Anaconda configuration file ISOs generated by ImageBuilder for Bare Metal Image Deployment
  
[Anaconda]
# List of enabled Anaconda DBus modules for RHEL.
#  but without org.fedoraproject.Anaconda.Modules.Subscription
kickstart_modules =
#     org.fedoraproject.Anaconda.Modules.Timezone
     # TODO turn off but allow to configure with kickstart
#     org.fedoraproject.Anaconda.Modules.Localization
#     org.fedoraproject.Anaconda.Modules.Users
     org.fedoraproject.Anaconda.Modules.Payloads
     # TODO possibly turn off
#     org.fedoraproject.Anaconda.Modules.Subscription

```

### DIFF
--- a/pyanaconda/installation.py
+++ b/pyanaconda/installation.py
@@ -97,21 +97,23 @@ def _prepare_configuration(payload, ksdata):
     security_dbus_tasks = security_proxy.InstallWithTasks()
     os_config.append_dbus_tasks(SECURITY, security_dbus_tasks)
 
-    # add installation tasks for the Timezone DBus module
-    # run these tasks before tasks of the Services module
-    timezone_proxy = TIMEZONE.get_proxy()
-    timezone_dbus_tasks = timezone_proxy.InstallWithTasks()
-    os_config.append_dbus_tasks(TIMEZONE, timezone_dbus_tasks)
+    if is_module_available(TIMEZONE):
+        # add installation tasks for the Timezone DBus module
+        # run these tasks before tasks of the Services module
+        timezone_proxy = TIMEZONE.get_proxy()
+        timezone_dbus_tasks = timezone_proxy.InstallWithTasks()
+        os_config.append_dbus_tasks(TIMEZONE, timezone_dbus_tasks)
 
     # add installation tasks for the Services DBus module
     services_proxy = SERVICES.get_proxy()
     services_dbus_tasks = services_proxy.InstallWithTasks()
     os_config.append_dbus_tasks(SERVICES, services_dbus_tasks)
 
-    # add installation tasks for the Localization DBus module
-    localization_proxy = LOCALIZATION.get_proxy()
-    localization_dbus_tasks = localization_proxy.InstallWithTasks()
-    os_config.append_dbus_tasks(LOCALIZATION, localization_dbus_tasks)
+    if is_module_available(LOCALIZATION):
+        # add installation tasks for the Localization DBus module
+        localization_proxy = LOCALIZATION.get_proxy()
+        localization_dbus_tasks = localization_proxy.InstallWithTasks()
+        os_config.append_dbus_tasks(LOCALIZATION, localization_dbus_tasks)
 
     # add the Firewall configuration task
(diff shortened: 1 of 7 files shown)


rvykydal:
I am going to open a separate PR for network module which is a bit different:
- kickstart setting is not mandatory so we don't have to disable the spoke
- in case of kickstart configuration existing we want to apply it

rvykydal:
> I am going to open a separate PR for network module which is a bit different:
> 
>     * kickstart setting is not mandatory so we don't have to disable the spoke
> 
>     * in case of kickstart configuration existing we want to apply it

https://github.com/rhinstaller/anaconda/pull/3074

pyanaconda/ui/gui/spokes/datetime_spoke.py
```diff
@@ -564,6 +565,10 @@ def _initialize(self):
         # report that we are done
         self.initialize_done()
 
+    @property
+    def showable(self):
```
poncovka, line 569: I think that we should use `should_run` instead of `showable`. There is too much code that might crash before we check the `showable` property.

pyanaconda/ui/gui/spokes/welcome.py
```diff
@@ -112,9 +113,10 @@ def apply(self):
     @property
     def completed(self):
         # Skip the welcome screen if we are in single language mode
+        # or module is not available.
         # If language has not been set the default language (en_US)
         # will be used for the installation and for the installed system.
-        if flags.flags.singlelang:
+        if flags.flags.singlelang or not is_module_available(LOCALIZATION):
```
poncovka, line 119: It would be nice to use `should_run` as well if possible, but this solution is good enough for now.

pyanaconda/ui/gui/spokes/user.py
```diff
@@ -437,6 +438,10 @@ def refresh(self):
         # rerun checks so that we have a correct status message, if any
         self.checker.run_checks()
 
+    @property
+    def showable(self):
+        return is_module_available(USERS)
```
poncovka, line 443: Do we need to support this feature in GUI for now? What about TUI?

poncovka (commented):
I have some suggestions.

pyanaconda/installation.py
```diff
@@ -97,21 +97,23 @@ def _prepare_configuration(payload, ksdata):
...
 
-    # add installation tasks for the Localization DBus module
-    localization_proxy = LOCALIZATION.get_proxy()
-    localization_dbus_tasks = localization_proxy.InstallWithTasks()
-    os_config.append_dbus_tasks(LOCALIZATION, localization_dbus_tasks)
+    if is_module_available(LOCALIZATION):
+        # add installation tasks for the Localization DBus module
+        localization_proxy = LOCALIZATION.get_proxy()
+        localization_dbus_tasks = localization_proxy.InstallWithTasks()
+        os_config.append_dbus_tasks(LOCALIZATION, localization_dbus_tasks)
```
jkonecny12, line 116: I'm thinking about abstraction here. Something like:
```
self.add_install_tasks_for_available_module(os_config, LOCALIZATION)
```

jkonecny12 (commented):
Looks good to me otherwise aside of what Vendy already mentioned.

poncovka:
Deprecated by the following pull requests:

* Timezone: https://github.com/rhinstaller/anaconda/pull/3081
* Users: https://github.com/rhinstaller/anaconda/pull/3082
* Localization: https://github.com/rhinstaller/anaconda/pull/3083
* Users: https://github.com/rhinstaller/anaconda/pull/3082
* Services: https://github.com/rhinstaller/anaconda/pull/3084
* Security: https://github.com/rhinstaller/anaconda/pull/3085

PR #3072, opened 2021-01-06
Use should_run instead of showable

poncovka:
Create a new object that will provide access to all persistent objects of the
user interface. The goal is to replace all these objects with this one. In the
future, the context might be replaced with a DBus module.

Define the `should_run` method instead of the `showable` property to prevent
the initialization of spokes that shouldn't run at all.

### DIFF
--- a/pyanaconda/ui/context.py
+++ b/pyanaconda/ui/context.py
@@ -1,0 +1,54 @@
+#
+# Copyright (C) 2021 Red Hat, Inc.
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# the GNU General Public License v.2, or (at your option) any later version.
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY expressed or implied, including the implied warranties of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General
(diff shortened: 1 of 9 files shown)


M4rtinK (approved):
Looks good to me - nice improvement! :+1: 

BTW, I guess some addons might be using the showable property and might need to be adjusted ? Initial Setup does not seem to be touching it.

pyanaconda/ui/context.py
```diff
@@ -0,0 +1,60 @@
...
+
+    @property
+    def anaconda(self) -> Anaconda:
+        """The Anaconda object."""
+        return self._anaconda
+
+    @property
+    def flags(self) -> Flags:
+        """The Anaconda flags."""
+        return flags
```
jkonecny12, line 47: I don't think you need this property here. Let's remove it because it's not used anywhere.

jkonecny12 (approved):
Looks good to me otherwise. Although, I would recommend to run complete KS tests suite to this to have at least some check that it's working as expected.

EDIT:
Of course if you need it fixed soon then not wait for KS tests.

rvykydal (approved):
Looks good to me.

pyanaconda/ui/context.py
poncovka: I have removed the flags for now.

poncovka:
The failing coverage tests are not related to this issue.

PR #3074, opened 2021-01-06
Allow disabling of applying of network module configuration

rvykydal:
For automatic installation if network module is disabled the network
configuration will be written to target system only if it is explicitly
configured by kickstart command. Which also means that configuration files
based on network boot options and default configuration files will not be
written to target system.

The point is to allow to prevent overwriting network configuration on installed
image while still offering option to explicitly configure networking via
kickstart.

Known limitations of this approach:

Defining only hostname via kickstart network command would cause also writing
the default configuration to the target system. To overcome this we'd need the
network module to offer something like IsNetworkDeviceKickstarted property.

Network configuration from boot options will not be passed to installed system
if there is no kickstart network configuration.

Resolves: rhbz#1834535

### DIFF
--- a/pyanaconda/network.py
+++ b/pyanaconda/network.py
@@ -43,6 +43,7 @@ from pyanaconda.modules.common.constants.services import NETWORK, TIMEZONE, STOR
 from pyanaconda.modules.common.constants.objects import FCOE
 from pyanaconda.modules.common.task import sync_run_task
 from pyanaconda.modules.common.structures.network import NetworkDeviceInfo
+from pyanaconda.modules.common.util import is_module_available
 
 from pyanaconda.anaconda_loggers import get_module_logger
 log = get_module_logger(__name__)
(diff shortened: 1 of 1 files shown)


pyanaconda/network.py
```diff
@@ -318,13 +319,17 @@ def write_configuration(overwrite=False):
...
 
-    task_path = network_proxy.InstallNetworkWithTask(overwrite)
-    task_proxy = NETWORK.get_proxy(task_path)
-    sync_run_task(task_proxy)
-
-    task_path = network_proxy.ConfigureHostnameWithTask(overwrite)
-    task_proxy = NETWORK.get_proxy(task_path)
-    sync_run_task(task_proxy)
+    if is_module_available(NETWORK) \
+            or network_proxy.Kickstarted:
```
poncovka, line 323: I think we shouldn't disable modules that we still plan to use in some cases. This wasn't the intended use case. It is just a nice fallback that prevents anaconda from crashing if we forget to handle the disabled module somewhere. In the future, it shouldn't be allowed at all.

It looks like these cases could be configurable via our configuration files.

pyanaconda/network.py
```diff
@@ -318,13 +319,17 @@ def write_configuration(overwrite=False):
...
+        task_path = network_proxy.InstallNetworkWithTask(overwrite)
+        task_proxy = NETWORK.get_proxy(task_path)
+        sync_run_task(task_proxy)
+
+    if is_module_available(NETWORK) \
+            or network_proxy.Hostname and network_proxy.Hostname != DEFAULT_HOSTNAME:
```
poncovka, line 329: I would move the conditions inside the DBus module or the DBus task if possible. Same above.

poncovka (commented):
I have some suggestions.

jkonecny12 (commented):
Other than what @poncovka said seems good to me.

poncovka:
Deprecated by https://github.com/rhinstaller/anaconda/pull/3086.

PR #3075, opened 2021-01-07
Don't run the execute method in the kickstart installation in TUI by default

poncovka:
We should run the `execute` method from the `initialize` method of the spoke when it
is required. In GUI, we have removed this functionality in the commit 31d35e9.

This patch will fix the broken dir installation.


M4rtinK (approved):
Looks good to me, thanks! :)

poncovka:
The failing coverage tests are not related to this issue.

jkonecny12 (approved):
LGTM!

PR #3077, opened 2021-01-07
Fix kickstart tests failure when loop device is not yet created

jkonecny12:
The loop device is not in the container automatically it has to be created first. Let's pre-create the loop 0 and 1 devices always before podman run.

Also add update of container images used by this workflow.

Tested [here](https://github.com/Test-anaconda-org/anaconda/runs/1663957864?check_suite_focus=true).

### DIFF
--- a/.github/workflows/kickstart-tests.yml
+++ b/.github/workflows/kickstart-tests.yml
@@ -54,6 +54,8 @@ jobs:
     if: needs.pr-info.outputs.base_ref == 'master' && needs.pr-info.outputs.allowed_user == 'true' && needs.pr-info.outputs.launch_args != ''
     runs-on: [self-hosted, kstest]
     timeout-minutes: 300
+    env:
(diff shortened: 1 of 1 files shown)


M4rtinK (approved):
Looks good to me, thanks for the fix! :)

VladimirSlavik (commented):
Changes look good to me, although I'm curious: Why is the linked test a keyboard one, if this affects network?

jkonecny12:
> Changes look good to me, although I'm curious: Why is the linked test a keyboard one, if this affects network?

It's just one of the simplest :D.

VladimirSlavik (approved):
My bad, it's loop, not loopback :D

jkonecny12:
Not your bad but my :D. I somehow mixed up those two.

jkonecny12:
@poncovka  this should fix issue with running kickstart tests on Anaconda PR

PR #3078, opened 2021-01-07
Fix duplicated rd.shell=0 (#1907258)

VladimirSlavik:
Contributed in [rhbz#1907258](https://bugzilla.redhat.com/show_bug.cgi?id=1907258). First commit is 1:1 with the patch except for message edited for our guidelines. In the second commit I tried to pythonize the code.

Not tested yet.

### DIFF
--- a/pyanaconda/modules/storage/bootloader/grub2.py
+++ b/pyanaconda/modules/storage/bootloader/grub2.py
@@ -323,15 +323,14 @@ class GRUB2(BootLoader):
         # See if we have a password and if so update the boot args before we
         # write out the defaults file.
         if self.password or self.encrypted_password:
-            rd_shell_not_exist = 1
-            grub_d_file = "%s%s" % (util.getSysroot(), self.grub_d_file)
-            with open(grub_d_file, "r") as fo:
-                for line in fo.readlines():
-                    match = re.search("rd.shell=0", line)
-                    if match:
-                        rd_shell_not_exist = 0
-                        break
-            if rd_shell_not_exist == 1:
+            # Add rd.shell=0 but only if not present yet, see rhbz#1907258
+            grub_d_file = "%s%s" % (conf.target.system_root, self.grub_d_file)
+            try:
+                with open(grub_d_file, "r") as fo:
+                    if "rd.shell=0" not in " ".join(fo.readlines()):
+                        self.boot_args.add("rd.shell=0")
+            except OSError:
+                # Could not open file because it did not exist
                 self.boot_args.add("rd.shell=0")
         self.write_defaults()
 
--- a/pyanaconda/modules/storage/bootloader/grub2.py
+++ b/pyanaconda/modules/storage/bootloader/grub2.py
@@ -105,6 +105,7 @@ class GRUB2(BootLoader):
     _config_dir = "grub2"
     _passwd_file = "user.cfg"
     defaults_file = "/etc/default/grub"
+    grub_d_file = "/etc/grub.d/10_linux"
     terminal_type = "console"
     stage2_max_end = None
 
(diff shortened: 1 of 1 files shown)


pyanaconda/modules/storage/bootloader/grub2.py
```diff
@@ -105,6 +105,7 @@ class GRUB2(BootLoader):
     _config_dir = "grub2"
     _passwd_file = "user.cfg"
     defaults_file = "/etc/default/grub"
+    grub_d_file = "/etc/grub.d/10_linux"
```
VladimirSlavik, line 108: Not sure if this is the right place to have the constant.

poncovka:
I don't think that this is a good idea for several reasons.

1. I wasn't able to reproduce the issue.
2. The `/etc/grub.d/10_linux` file is a script, so I don't recommend to parse the file.
3. There are more files in `/etc/grub.d/` that might add something on the cmdline.

pyanaconda/modules/storage/bootloader/grub2.py
```diff
@@ -323,15 +323,14 @@ def write_config(self):
...
-            grub_d_file = "%s%s" % (util.getSysroot(), self.grub_d_file)
-            with open(grub_d_file, "r") as fo:
-                for line in fo.readlines():
-                    match = re.search("rd.shell=0", line)
-                    if match:
-                        rd_shell_not_exist = 0
-                        break
-            if rd_shell_not_exist == 1:
+            # Add rd.shell=0 but only if not present yet, see rhbz#1907258
+            grub_d_file = "%s%s" % (conf.target.system_root, self.grub_d_file)
```
jkonecny12, line 327: Please use ``format`` we are trying to get rid of the ``%`` in the code.

jkonecny12 (changes requested):
Looks great to me otherwise. Thanks for those improvements!
Just please do the small change requested below.

jkonecny12:
> I don't think that this is a good idea for several reasons.
> 
>     1. I wasn't able to reproduce the issue.

We don't have the reproducer. I guess there is a condition you need to do to have ``/etc/grub.d/10_linux`` with this parameter. Maybe asking reporter first would be a good idea here.

> 
>     2. The `/etc/grub.d/10_linux` file is a script, so I don't recommend to parse the file.

Scripts used for the configuration of the system are common. I don't see a problem with parsing that. Or do you have a specific reason why we don't want to do that?

> 
>     3. There are more files in `/etc/grub.d/` that might add something on the cmdline.

Yes, but the file mentioned above is specifically created for this purpose (I think). The same solution is used also for example for ``/etc/X11/xinit/xinitrc.d/`` for example.

jstodola:
>     2. The `/etc/grub.d/10_linux` file is a script, so I don't recommend to parse the file.
I agree with this point.
Trying to find a string **anywhere** in a ~500 lines shell script (even in a comment) simply cannot be a correct solution.

jkonecny12:
> > ```
> > 2. The `/etc/grub.d/10_linux` file is a script, so I don't recommend to parse the file.
> > ```
> 
> I agree with this point.
> Trying to find a string **anywhere** in a ~500 lines shell script (even in a comment) simply cannot be a correct solution.

OK my bad, I did not looked on that script. If it's that long it seems really as a bad decision.

VladimirSlavik:
Blocked until we learn more, it seems.

VladimirSlavik:
It has been a month and there was no reply - neither here nor [in the bug](https://bugzilla.redhat.com/show_bug.cgi?id=1907258). Therefore I'm closing this. We can't accept patches without knowing why something is a problem, much less patches that need further work to be acceptable.

To be clear, this is not a rejection of the problem or any fix for it, just of this particular PR. The bug remains open. Do feel free to continue the discussion there, open new PRs etc. related to this problem.

PR #3079, opened 2021-01-08
Flush stdout during the TUI installation

poncovka:
Make sure that we flush stdout after every progress message and after handling
the cmdline error.


jkonecny12 (approved):
Looks good to me. However, wouldn't it be better to have our printing which would flush all the time? I think it would be helpful to avoid forgetting the ``flush`` parameter.

poncovka:
> Looks good to me. However, wouldn't it be better to have our printing which would flush all the time? I think it would be helpful to avoid forgetting the `flush` parameter.

Maybe. The current flushing works mostly fine and it would be a huge and annoying change.

PR #3080, opened 2021-01-11
Always show pykickstart parse warnings

VladimirSlavik:
Previously, we would show warnings only in strict mode.

Resolves: [rhbz#1897663](https://bugzilla.redhat.com/show_bug.cgi?id=1897663)

### DIFF
--- a/pyanaconda/kickstart.py
+++ b/pyanaconda/kickstart.py
@@ -533,21 +533,19 @@ def parseKickstart(handler, f, strict_mode=False, pass_to_boss=False):
             # Parse the kickstart file in anaconda.
             ksparser.readKickstart(f)
 
-            # Process pykickstart warnings in the strict mode:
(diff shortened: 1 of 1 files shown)


VladimirSlavik:
Together with https://github.com/pykickstart/pykickstart/pull/343, it looks like this:

Screenshot_vs_main_2021-01-11_16:14:29

jstodola (approved):
It looks good to me, thank you!
A possible improvement would be avoiding splitting lines in the middle of words.

poncovka (approved):
Looks good to me. Thanks!

VladimirSlavik:
> A possible improvement would be avoiding splitting lines in the middle of words.

Unfortunately, that's what `print` gives me, so probably pretty deep in cPython... :(

jkonecny12 (approved):
LGTM!

PR #3081, opened 2021-01-11
Allow to disable the Timezone module

poncovka:
Check whether the Timezone DBus module is enabled before it is used.

Related: rhbz#1834535

### DIFF
--- a/pyanaconda/startup_utils.py
+++ b/pyanaconda/startup_utils.py
@@ -38,6 +38,9 @@ from pyanaconda import safe_dbus
 from pyanaconda import kickstart
 from pyanaconda.flags import flags
 from pyanaconda.screensaver import inhibit_screensaver
+from pyanaconda.modules.common.constants.services import TIMEZONE
+from pyanaconda.modules.common.util import is_module_available
+from pyanaconda.threading import AnacondaThread, threadMgr
 
 import blivet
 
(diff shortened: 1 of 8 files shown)


pyanaconda/startup_utils.py
```diff
@@ -353,3 +356,46 @@ def parse_kickstart(ks, addon_paths, strict_mode=False):
...
+
+
+def initialize_system_clock():
+    """Initialize the system clock."""
+    if not conf.system.can_initialize_system_clock:
+        log.debug("Skip the clock initialization.")
+        return
+
+    if not is_module_available(TIMEZONE):
+        return
```
VladimirSlavik, line 368: I'm not sure if it helps, but this could be logged too?

VladimirSlavik (approved):
Looks good to me. Thank you!

pyanaconda/ui/gui/spokes/welcome.py
```diff
@@ -89,6 +90,9 @@ def apply(self):
         if flags.flags.automatedInstall and not geoloc.geoloc.enabled:
             return
 
+        if not is_module_available(TIMEZONE):
+            return
```
jkonecny12, line 94: This would be better to refactor in the future. I don't think it's wise to have code .. conditions with early return ... code again. It's pretty simple to add code at the bottom but miss the early returns in the middle.

Also shouldn't we merge these two conditions. They are both disabling geoloc at the end.

jkonecny12 (approved):
Looks good to me otherwise.

pyanaconda/ui/gui/spokes/welcome.py
poncovka: I will change it in upstream. This code should be in a separate function.

poncovka:
Fixed one incorrect import.

poncovka:
Ported in https://github.com/rhinstaller/anaconda/pull/3094.

PR #3083, opened 2021-01-11
Allow to disable the Localization module

poncovka:
Check whether the Localization DBus module is enabled before it is used.

The installer should skip standalone GUI spokes that return False from their
`should_run` method. There are no standalone TUI spokes.

Related: rhbz#1834535

### DIFF
--- a/anaconda.py
+++ b/anaconda.py
@@ -323,7 +323,6 @@ if __name__ == "__main__":
 
     from pyanaconda import vnc
     from pyanaconda import kickstart
-    from pyanaconda import keyboard
     # we are past the --version and --help shortcut so we can import display &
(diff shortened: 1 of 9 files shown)


VladimirSlavik (approved):
As far as I understand it, looks good to me.

I really like how this series of changes cleans up `anaconda.py`.

jkonecny12 (approved):
LGTM!
I just wonder if `startup_utils` are not starting to be a bit overloaded.

M4rtinK (approved):
Looks good to me, just a suggestion to extend a docstring. :)

poncovka:
Rebased.

poncovka:
Ported in https://github.com/rhinstaller/anaconda/pull/3106.

PR #3086, opened 2021-01-11
Allow to disable the network installation

poncovka:
Check the `can_configure_network` configuration option of the `Installation Target`
section before the installation of the network configuration on the target
system.

Related: rhbz#1834535

### DIFF
--- a/data/anaconda.conf
+++ b/data/anaconda.conf
@@ -46,6 +46,10 @@ physical_root = /mnt/sysimage
 # A path to the system root of the target.
 system_root = /mnt/sysroot
 
+# Should we install the network configuration?
+can_configure_network = True
+
+
 [Network]
 # Network device to be activated on boot if none was configured so.
 # Valid values:
--- a/pyanaconda/core/configuration/target.py
+++ b/pyanaconda/core/configuration/target.py
@@ -61,3 +61,11 @@ class TargetSection(Section):
     def is_directory(self):
         """Are we installing to a directory?"""
         return self.type is TargetType.DIRECTORY
+
+    @property
+    def can_configure_network(self):
+        """Should we install the network configuration?
+
+        :return: True or False
+        """
+        return self._get_option("can_configure_network", bool)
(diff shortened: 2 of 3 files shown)


M4rtinK (approved):
Looks good to me, thanks! :)

data/anaconda.conf
```diff
@@ -56,6 +56,9 @@ system_root = /mnt/sysroot
 #
 default_on_boot = NONE
 
+# Enable installation of the network configuration on the target system.
+enable_installation = True
```
VladimirSlavik, line 60: I'm not sure if the name is enough to understand what it does. Not a blocker though.

VladimirSlavik (approved):
Looks good to me. Thank you!

data/anaconda.conf
poncovka: Maybe it could be in the `Installation Target` section as `can_change_network` or `can_configure_network`. It would nicely complement the `Installation System` section that has similar "rules" for the installation environment.

VladimirSlavik: The latter sounds good. But your call :)

poncovka: Fixed.

poncovka:
Updated.

data/anaconda.conf
```diff
@@ -46,6 +46,10 @@ physical_root = /mnt/sysimage
 # A path to the system root of the target.
 system_root = /mnt/sysroot
 
+# Should we install the network configuration?
+can_configure_network = True
```
jkonecny12, line 50: Not sure if the name of the value shouldn't be more specific. I mean `can_configure_network` is by name applicable even to installation environment not just to resulting system.

Maybe something like `can_install_network`?

jkonecny12 (changes requested):
I think we should change the name here but let's discuss this a bit.

jkonecny12 (approved):
I've missed the section name before. Sorry for the noise.

VladimirSlavik (approved):
Re-acking. Thank you for the name change!

poncovka:
Ported in https://github.com/rhinstaller/anaconda/pull/3096.

PR #3087, opened 2021-01-12
Save logs correctly after container update GH action run

jkonecny12:
Logs were moved to test-logs folder so we should adapt the workflow to the correct path.

Trying to fix [this](https://github.com/rhinstaller/anaconda/actions/runs/479018492) situation.

Tested [here](https://github.com/Test-anaconda-org/anaconda/actions/runs/479886238). Job has failed which is expected, the point is that logs are stored correctly.

### DIFF
--- a/.github/workflows/container-autoupdate.yml
+++ b/.github/workflows/container-autoupdate.yml
@@ -37,10 +37,10 @@ jobs:
         with:
           name: logs
           path: |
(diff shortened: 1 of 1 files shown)


poncovka (approved):
Looks good to me.

VladimirSlavik (approved):
Looks good to me, thank you for fixing this!

PR #3088, opened 2021-01-13
Stop the timer that calls a callback for clicking on the continue button

poncovka:
Otherwise, the summary hub repeatedly runs this callback even after it is left.


poncovka:
/kickstart-test ui_graphical_interactive ui_graphical_noninteractive

jkonecny12 (approved):
LGTM!

PR #3091, opened 2021-01-14
Grab GitHub notifications for periodic workflows

jkonecny12:
Unfortunately, GitHub will sent notifications about failures for on:schedule workflows just to the last person who did some changes to the workflow. :(

So let's go and steal it from pitti!

### DIFF
--- a/.github/workflows/daily-rhel-copr.yml
+++ b/.github/workflows/daily-rhel-copr.yml
@@ -1,3 +1,4 @@
+# Do a build of the rhel-8 branch to RH internal COPR
 name: Build current anaconda rhel-8 branch in RHEL COPR
 on:
   schedule:


martinpitt:
I hear a lot of cyborg teams are using sentry.io. I don't know about it myself yet, but that may be a better way to store and distribute failure notifications than these simple emails.

martinpitt (approved):

VladimirSlavik (approved):

poncovka (approved):

jkonecny12:
> I hear a lot of cyborg teams are using sentry.io. I don't know about it myself yet, but that may be a better way to store and distribute failure notifications than these simple emails.

Yes, I was looking on that briefly and need a deeper look to find out more. It's not usable for Anaconda as project but it may be nice for a service monitoring we are using.

PR #3092, opened 2021-01-14
Always show pykickstart parse warnings

VladimirSlavik:
Previously, we would show warnings only in strict mode.

Resolves: rhbz#1897663

Port of #3080


poncovka (approved):
Looks good to me.

jkonecny12 (approved):
LGTM!

PR #3093, opened 2021-01-18
 Select disks for implicit partitions

poncovka:
Don't use disks that will be used for allocation of requested partitions
(for example, the /boot partition). Calculate the number of slots that
will be used by requested partitions and skip disks that will not have
free slots for implicit partitions.

Otherwise, we might schedule implicit partitions, that wouldn't be possible
to allocate on the specified disk, and the partitioning might fail.

Related: rhbz#1642391

**Ported from** https://github.com/rhinstaller/anaconda/pull/2973.

### DIFF
--- a/pyanaconda/modules/storage/partitioning/specification.py
+++ b/pyanaconda/modules/storage/partitioning/specification.py
@@ -19,6 +19,8 @@
 #
 
 from blivet.util import stringize, unicodeize
+from pykickstart.constants import AUTOPART_TYPE_PLAIN, AUTOPART_TYPE_BTRFS, AUTOPART_TYPE_LVM, \
+    AUTOPART_TYPE_LVM_THINP
 
 
 class PartSpec(object):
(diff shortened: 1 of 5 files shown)


poncovka:
kickstart --testtype smoke

poncovka:
kickstart --testtype smoke

pyanaconda/modules/storage/partitioning/specification.py
```diff
@@ -82,6 +84,52 @@ def _to_string(self):
...
+
+        :param scheme: a partitioning scheme
+        :return: True or False
+        """
+        if not self.is_lvm_volume(scheme):
+            return False
+
+        return scheme == AUTOPART_TYPE_LVM_THINP and self.thin
+
+    def is_btrfs_subvolume(self, scheme):
```
VladimirSlavik, line 125: Is no test needed for this?

VladimirSlavik (approved):
Looks good to me, thank you!

pyanaconda/modules/storage/partitioning/specification.py
poncovka: It is tested via the `is_volume` and `is_partition` methods.

PR #3096, opened 2021-01-18
Allow to disable the network installation

poncovka:
Check the `can_configure_network` configuration option of the `Installation Target`
section before the installation of the network configuration on the target
system.

(cherry-picked from a commit 76d52f2)

Related: rhbz#1834535


poncovka:
kickstart --testtype smoke

rvykydal (approved):
Looks good to me.
I am a bit concerned about configuring hostname setting separately but I think we should handle it if/when we have request / use case for that.

poncovka:
kickstart --testtype smoke

PR #3098, opened 2021-01-19
network: handle wireless configure button sensitiveness

rvykydal:
When configuring a wireless device (clicking Configure button) without any
access point selected before (by clicking Select network) we used to present
the dialog with empty list of access points. With the patch make the Configure
button sensitive only if there already is some access point to be selected.

Resolves: rhbz#1917437


rvykydal:
kickstart --testtype smoke

VladimirSlavik (approved):
Looks good to me. Thank you!

PR #3099, opened 2021-01-19
Move the password policy into the configuration files and on DBus

poncovka:
Use the configuration options `can_change_root` and `can_change_users` 
from the `User Interface` section instead of the `changesok` attribute of the
`pwpolicy` kickstart command.

Use the `password_policies` configuration option of the `User Interface` section to
define the default password policies.

Use the User Interface module of the Boss service to get and set password policies
during the runtime. This support is required for the OSCAP addon.

Apply the password policy specified in the kickstart file if there is any
and set up the UI DBus module and the Anaconda configuration options.

Deprecate the `%anaconda` kickstart section and its `pwpolicy` kickstart command.
The section specified in a kickstart file will generate a deprecation warning.

The `interactive-defaults.ks` file is deprecated and will be removed in in the future.
Use the Anaconda configuration files instead. Anaconda no longer uses this file for
the configuration of the interactive defaults, but some of the products do. Make sure
that the products use the configuration files before the removal of this support.

### DIFF
--- a/data/anaconda.conf
+++ b/data/anaconda.conf
@@ -270,6 +270,12 @@ hidden_spokes =
 # Run GUI installer in a decorated window.
 decorated_window = False
 
+# Should the UI allow to change the configured root account?
+can_change_root = False
+
+# Should the UI allow to change the configured user accounts?
+can_change_users = False
+
 
 [License]
 # A path to EULA (if any)
--- a/data/anaconda.conf
+++ b/data/anaconda.conf
@@ -276,6 +276,21 @@ can_change_root = False
 # Should the UI allow to change the configured user accounts?
 can_change_users = False
 
+# Define the default password policies.
+# Specify a policy name and its attributes on each line.
+#
+# Valid attributes:
+#
+#   quality <NUMBER>  The minimum quality score (see libpwquality).
+#   length <NUMBER>   The minimum length of the password.
+#   empty             Allow an empty password.
+#   strict            Require the minimum quality.
+#
+password_policies =
+    root (quality 1, length 6)
+    user (quality 1, length 6, empty)
+    luks (quality 1, length 6)
+
 
 [License]
 # A path to EULA (if any)
--- a/pyanaconda/core/configuration/ui.py
+++ b/pyanaconda/core/configuration/ui.py
@@ -18,6 +18,7 @@
 #  Author(s):  Vendula Poncova <vponcova@redhat.com>
 #
 from pyanaconda.core.configuration.base import Section
+from pyanaconda.core.configuration.utils import split_name_and_attributes
 
 
 class UserInterfaceSection(Section):
@@ -93,3 +94,64 @@ class UserInterfaceSection(Section):
         :return: True or False
         """
         return self._get_option("can_change_users", bool)
+
+    @property
+    def password_policies(self):
+        """The password policies.
+
+        Returns a list of dictionaries with password policy attributes.
+        The name of the policy is represented by the attribute 'name'
+        in the dictionary representation.
+
+        Valid attributes:
+
+            name        The name of the policy.
+            quality     The minimum quality score (see libpwquality).
+            length      The minimum length of the password.
+            empty       Allow an empty password.
+            strict      Require the minimum quality.
+
+        :return: a list of dictionaries with policy attributes
+        """
+        return self._get_option("password_policies", self._convert_policies)
+
+    def _convert_policies(self, value):
+        """Convert a policies string into a list of dictionaries."""
+        return list(map(self._convert_policy_line, value.strip().split("\n")))
+
+    @classmethod
+    def _convert_policy_line(cls, line):
+        """Convert a policy line into a dictionary."""
+        # Parse the line.
+        name, raw_attrs = split_name_and_attributes(line)
+
+        # Generate the dictionary.
+        attrs = {"name": name}
+
+        for name, value in raw_attrs.items():
+            if not value and name in ("strict", "empty"):
+                # Handle a boolean attribute.
+                attrs[name] = True
+            elif value and name in ("length", "quality"):
+                # Handle an integer attribute.
+                attrs[name] = int(value)
+            else:
+                # Handle an invalid attribute.
+                raise ValueError("Invalid attribute: " + name)
+
+        # Validate the dictionary.
+        cls._validate_policy_attributes(attrs)
+
+        return attrs
+
+    @staticmethod
+    def _validate_policy_attributes(attrs):
+        """Validate the dictionary with policy attributes."""
+        if not attrs.get("name"):
+            raise ValueError("The name of the policy is not specified.")
+
+        if "length" not in attrs:
+            raise ValueError("The minimal length is not specified.")
+
+        if "quality" not in attrs:
+            raise ValueError("The minimal quality is not specified.")
--- a/pyanaconda/core/configuration/ui.py
+++ b/pyanaconda/core/configuration/ui.py
@@ -69,3 +69,27 @@ class UserInterfaceSection(Section):
         :return: True or False
         """
         return self._get_option("decorated_window", bool)
+
+    @property
+    def can_change_root(self):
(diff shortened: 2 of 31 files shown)


poncovka:
kickstart --testtype smoke

data/anaconda.conf
```diff
@@ -266,6 +266,12 @@ hidden_spokes =
 # Run GUI installer in a decorated window.
 decorated_window = False
 
+# Should the UI allow to change the configured root?
```
M4rtinK, line 269: I think we should always talk about "root password" in this case as otherwise it's easy to confuse with install root/rootfs/root directory/etc.

Also it seems to me that we might want to clarify a bit more what "configured" means - maybe "configured via kickstart" or possibly "configured via the DBus API" ?

pyanaconda/modules/common/structures/policy.py
```diff
@@ -0,0 +1,139 @@
...
+
+    def __init__(self):
+        self._quality = UInt16(0)
+        self._length = UInt16(0)
+        self._empty = True
+        self._strict = False
+
+    @property
+    def quality(self) -> UInt16:
+        """The minimal quality of the password.
```
M4rtinK, line 41: I think he we should also mention that quality in this case means the `libpwquality` defined password quality score.

data/anaconda.conf
```diff
@@ -272,6 +272,21 @@ can_change_root = False
 # Should the UI allow to change the configured users?
 can_change_users = False
 
+# Define the default password policies.
+# Specify a policy name and its attributes on each line.
+#
+# Valid attributes:
+#
+#   quality <NUMBER>  The minimal quality of the password.
```
M4rtinK, line 280: I think there should be a note about what "qaility" means - eq. that it's the password quality score defined by `libpwquality`.

pyanaconda/kickstart.py
```diff
@@ -241,64 +240,24 @@ def __str__(self):
...
-
-    def finalize(self):
-        """Let %anaconda know no additional data will come."""
-        Section.finalize(self)
-
+        warnings.warn(_(
+            "The %%anaconda section has been deprecated and "
+            "no longer has any effect. It may be removed from "
+            "future releases, which will result in a fatal error "
+            "from kickstart. Please modify your kickstart file "
```
M4rtinK, line 258: Maybe:
"from kickstart" -> "when parsing kickstart"/"during kickstart processing" ?

M4rtinK (approved):
Looks good to me - quite a bit cleaner now than with the old solution! :)

Just added dome docstring improvement suggestions. :)

jstodola:
:warning: `pwpolicy` is a documented kickstart command, which can be used by users:

- https://docs.fedoraproject.org/en-US/fedora/f33/install-guide/appendixes/Kickstart_Syntax_Reference/#sect-kickstart-commands-pwpolicy
- https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/performing_an_advanced_rhel_installation/kickstart-commands-and-options-reference_installing-rhel-as-an-experienced-user#pwpolicy_commands-used-in-anaconda

Removing the support from kickstart without deprecating the command first is not correct, especially when it is related to security of the installed system. Because of that I cannot agree with this change.

VladimirSlavik (commented):
Overall, I like the idea and think it is a very logical step. I see no particular reason why UI should be configured in Kickstart.

Before, we had the data-driven thing, so everything was Kickstart, sometimes just because that was the paradigm. Now, we are putting everything on D-Bus, much in the same way. So far so good.

Configuration files as the data source also make sense, the persona you'd expect changing this seems to fit (IMO).

No issues with the code itself.

However, I really disagree with the (re)naming of some fields, see review below. Generally I find the old names self-explanatory, while the new ones are shorter but often misleading by omission. Some of the changed docstrings could also use some help in this direction. However, these are easily changed details.

data/anaconda.conf
poncovka: Fixed.

data/anaconda.conf
poncovka: Fixed.

pyanaconda/core/configuration/base.py
```diff
@@ -51,6 +51,21 @@ def __str__(self):
                "'{}': {}".format(self._option, self._section, super().__str__())
 
 
+def split_string(value, delimiter=None):
```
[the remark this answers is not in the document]
poncovka: Fixed.

pyanaconda/modules/common/structures/policy.py
poncovka: Fixed.

pyanaconda/modules/common/structures/policy.py
```diff
@@ -0,0 +1,139 @@
...
+        else:
+            log.debug(
+                "No default %s password policy is "
+                "configured.", policy_name
+            )
+
+        return policy
+
+    @classmethod
+    def to_structure_dict(cls, objects) -> Dict[Str, Structure]:
```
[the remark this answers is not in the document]
poncovka: The type hints are not supposed to be used for something like that and I don't really want to test their limits.

pyanaconda/input_checking.py
```diff
@@ -379,14 +382,14 @@ def run(self, check_request):
             # PWQError values are built as a tuple of (int, str)
             error_message = e.args[1]
 
-        if check_request.policy.emptyok:
+        if check_request.policy.empty and not check_request.password:
```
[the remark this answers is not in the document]
poncovka: Fixed.

tests/nosetests/pyanaconda_tests/password_quality_test.py
```diff
@@ -66,8 +69,8 @@ def password_empty_ok_test(self):
         # empty_ok with password length
         request = input_checking.PasswordCheckRequest()
         request.policy = get_policy()
-        request.policy.emptyok = True
-        request.policy.minlen = 10
+        request.policy.empty = True
+        request.policy.length = 10
```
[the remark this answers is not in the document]
poncovka: Fixed.

pyanaconda/ui/tui/tuiobject.py
```diff
@@ -264,7 +264,7 @@ def _validate_password(self, password, confirm):
             self._report(password_check.result.error_message)
             return None
 
-        if password_check.result.password_quality < self._policy.minquality:
+        if password_check.result.password_quality < self._policy.quality:
```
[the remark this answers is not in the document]
poncovka: Fixed.

pyanaconda/kickstart.py

pyanaconda/core/configuration/ui.py
```diff
@@ -85,3 +85,69 @@ def can_change_users(self):
         :return: True or False
         """
         return self._get_option("can_change_users", bool)
+
+    @property
+    def password_policies(self):
+        """The default password policies.
```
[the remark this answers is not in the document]
poncovka: Fixed.

poncovka:
> Removing the support from kickstart without deprecating the command first is not correct, especially when it is related to security of the installed system. Because of that I cannot agree with this change.

I have updated the pull request. The `pwpolicy` kickstart command will be only deprecated for now. It means that we will modify the policy and the configuration based on values of this command if it is specified in the kickstart file.

@jstodola, could we stop to write the `%anaconda` section into the output kickstart file? Otherwise, we would have to update the `pwpolicy` kickstart command before it is written. Or we can write the original values from the input kickstart file.

jstodola:
I think it makes sense to stop writing `pwpolicy` after interactive installation. (it is written there automatically if I remember correctly, right? ). 
As for Kickstart installations - do whatever you do with deprecated Ks commands, writing the original commands is fine.
If you want to make me super happy, you can write a comment in the output Kickstart that the `pwpolicy` command is deprecated :-)

poncovka:
Rebased and tested.

VladimirSlavik (approved):
Looks great, thank you!

PR #3100, opened 2021-01-19
Update versions of the kickstart commands

poncovka:
Some of the commands or their options were removed or deprecated.

### DIFF
--- a/anaconda.spec.in
+++ b/anaconda.spec.in
@@ -35,7 +35,7 @@ Source0: %{name}-%{version}.tar.bz2
 %define libxklavierver 5.4
 %define mehver 0.23-1
 %define nmver 1.0
-%define pykickstartver 3.31-1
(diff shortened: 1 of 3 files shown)


VladimirSlavik:
Thanks! Builds aren't up yet, though :(

poncovka:
> Thanks! Builds aren't up yet, though :(

Rawhide is not, but ELN unit tests started to fail on other pull requests.

poncovka:
The tests pass locally.

VladimirSlavik (approved):
Looks good to me. Thank you!

jkonecny12:
The `master` test won't work here because it's running in the old container with the old pykickstart. However, the new container is not build because it has broken test (fixed by this PR). So let's merge this and build new container to solve the issue.
Tested locally on newly build container works correctly.

PR #3104, opened 2021-01-20
Move flatpak installation code to a task

VladimirSlavik:
Split flatpak installation code to a task


pyanaconda/modules/payloads/payload/rpm_ostree/flatpak_installation.py
```diff
@@ -0,0 +1,58 @@
...
+        self._sysroot = sysroot
+
+    @property
+    def name(self):
+        return "Install Flatpaks"
+
+    def run(self):
+        self.report_progress(_("Starting Flatpak installation"))
+
+        flatpak_payload = FlatpakPayload(self._sysroot)
```
poncovka, line 42: Please, move the `FlatpakPayload` class to a new file in `pyanaconda.modules.payload.rpm_ostree` and rename it to something like `FlatpackManager`. I was very confused by this solution until I realized that this is not the typical payload class.

pyanaconda/modules/payloads/payload/rpm_ostree/flatpak_installation.py
```diff
@@ -0,0 +1,58 @@
...
+from pyanaconda.core.i18n import _
+from pyanaconda.modules.common.task import Task
+from pyanaconda.payload.errors import PayloadInstallError, FlatpakInstallError
+from pyanaconda.payload.flatpak import FlatpakPayload
+
+
+class InstallFlatpaksTask(Task):
+    """Task to install flatpaks from the local source on SilverBlue."""
+
+    def __init__(self, sysroot):
```
poncovka, line 27: The existing instance of `FlatpakPayload` is initialized in the `RPMOSTreePayloadWithFlatpaks` class, so you cannot create a new one for the task. However, it can be one of the arguments of the installation task.

pyanaconda/modules/payloads/payload/rpm_ostree/flatpak_installation.py
```diff
@@ -0,0 +1,58 @@
...
+# Red Hat, Inc.
+#
+from pyanaconda.core.i18n import _
+from pyanaconda.modules.common.task import Task
+from pyanaconda.payload.errors import PayloadInstallError, FlatpakInstallError
+from pyanaconda.payload.flatpak import FlatpakPayload
+
+
+class InstallFlatpaksTask(Task):
+    """Task to install flatpaks from the local source on SilverBlue."""
```
poncovka, line 25: I think it is `Silverblue`.

poncovka (changes requested):
I have some suggestions.

pyanaconda/modules/payloads/payload/rpm_ostree/flatpak_installation.py
VladimirSlavik: I don't think it matters in this case, same as for the OSTree stuff. As I see it:

The instance is persistent mainly to provide the install size on demand, but that could be easily extracted once and cached. The inputs are static, so the result must be too. In fact that would be a good refactoring step before further changes.

The remaining productive usage of the instance is preceded by `cleanup` which amounts to rmtree-ing whatever happened earlier and overwriting all private fields. A new instance will work just as well.

pyanaconda/modules/payloads/payload/rpm_ostree/flatpak_installation.py
VladimirSlavik: Good idea.

VladimirSlavik: And it is done.

pyanaconda/modules/payloads/payload/rpm_ostree/flatpak_installation.py
VladimirSlavik: I have verified the the caching idea and used it. The manager class now looks like a refactoring candidate, as I have identified two moderately independent code paths that it semi-randomly collects under one name. But that probably also includes a conversation with @jkonecny12 about potential future flatpak usage.

VladimirSlavik:
Should the flatpak size collection (now 4 lines in  `RPMOSTreePayloadWithFlatpaks.__init__`) also become a `Task`?

pyanaconda/modules/payloads/payload/rpm_ostree/flatpak_installation.py
poncovka: I see. Thanks for the explanation!

poncovka:
> Should the flatpak size collection (now 4 lines in `RPMOSTreePayloadWithFlatpaks.__init__`) also become a `Task`?

That is a good idea.

pyanaconda/anaconda.py
```diff
@@ -83,7 +83,7 @@ def payload(self):
             payload_type = self._get_dbus_payload_type()
 
             if payload_type == PAYLOAD_TYPE_RPM_OSTREE:
-                if FlatpakPayload.is_available():
+                if FlatpakManager.is_available():
```
poncovka, line 86: I am not sure if the `is_available` name still makes sense after the renaming. Maybe it could be `is_supported`?

VladimirSlavik:
Size-getting taskified, onwards to tests, let's see what other renaming comes up once the rest of the team is back.

pyanaconda/anaconda.py
VladimirSlavik: `is_supported` sounds to me like a (abstract) class method that could tell me if I should even instantiate the manager.

What this method tells me is, do the directories the manager "manages" give us some flatpak? (It checks only existence of dir though!) So I think `is_correctly_configured` might work better?

poncovka: Or we can keep the payload in the name and name it `is_payload_available` or `is_source_available`.

VladimirSlavik: Thanks, the source variant sounds good to me.

jkonecny12 (approved):
Looks great to me. If you tested these changes on SilverBlue media. Than you have my ACK!

Great work!

pyanaconda/anaconda.py
VladimirSlavik: Done.

poncovka (approved):
It looks great. Thanks!

VladimirSlavik:
I have some issues trying to run this with recent Silverblue composes, they install RPMs instead :O

VladimirSlavik:
Resolved, seems to work correctly.

PR #3105, opened 2021-01-20
Skip storage-related spokes in the initial setup (#1918048)

poncovka:
The initial setup doesn't configure storage. It was broken by the commit 2e305f8.

Resolves: rhbz#1918048


poncovka:
All these regressions where caused by the pull request https://github.com/rhinstaller/anaconda/pull/3072. I have checked all `should_run` methods that were changed in this pull request, so this should be hopefully the last fix.

@M4rtinK, could you test the initial setup with this change, please?

M4rtinK (approved):
Looks good to me & seems to fix the issue on my Initial Setup testing VM. Thanks for the fix! :)

PR #3106, opened 2021-01-20
Allow to disable the Localization module

poncovka:
Check whether the Localization DBus module is enabled before it is used.

The installer should skip standalone GUI spokes that return False from their
`should_run` method. There are no standalone TUI spokes.

(cherry-picked from commits efc6278 and 1df4c77)

Related: rhbz#1834535


poncovka:
Updated.

VladimirSlavik (approved):
Looks good to me. Thank you!

PR #3109, opened 2021-01-21
Fix failing tests after nose dependency was removed from the spec

jkonecny12:
We removed this dependency from spec file in commit
2be2374

However, we missed that these dependencies are not yet installed by pip. Fix this issue now.

Also apply a small fix on the rpm tests execution which is required now.


VladimirSlavik (approved):
Thank you for fixing my breakage!

poncovka (approved):
Looks good to me. Thanks!

PR #3110, opened 2021-01-21
Specify version 3 of GTK+ and GDK for fedora-welcome

AdamWill:
Now version 4 exists and is on the live image, if we don't specify
a version, we get 4. And the code doesn't work with 4.

Resolves: rhbz#1918940


poncovka (approved):
Looks good to me. Thanks!

M4rtinK (approved):
Looks good to me as well & thanks for the fix! :)

M4rtinK:
I wonder where else we might hit issues due to the system Gnome switch to GTK4 - can't come up with anything concrete from top of my head. I guess we will find out. ;-)

PR #3111, opened 2021-01-22
anaconda should add only s390utils-core

sharkcz:
The s390utils-core is the sufficient part of s390utils that's required for a s390x
system to be functional. Thus add only s390utils-core to the package set instead of
the complete s390utils.

Fixes: #1918642

### DIFF
--- a/pyanaconda/modules/storage/bootloader/zipl.py
+++ b/pyanaconda/modules/storage/bootloader/zipl.py
@@ -35,7 +35,7 @@ class ZIPL(BootLoader):
 
     name = "ZIPL"
     config_file = "/etc/zipl.conf"
-    packages = ["s390utils-base"]
+    packages = ["s390utils-core"]
 
     # stage2 device requirements
     stage2_device_types = ["partition"]
--- a/pyanaconda/modules/storage/platform.py
+++ b/pyanaconda/modules/storage/platform.py
@@ -398,11 +398,6 @@ class PS3(PPC):
 class S390(Platform):
 
     @property
-    def packages(self):
-        """Packages required for this platform."""
-        return ["s390utils"]
-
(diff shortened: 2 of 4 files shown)


VladimirSlavik (changes requested):
Thanks for the change! Unit tests fail, because you need to change them too:
- https://github.com/rhinstaller/anaconda/blob/master/tests/nosetests/pyanaconda_tests/modules/storage/module_storage_test.py#L293
- https://github.com/rhinstaller/anaconda/blob/master/tests/nosetests/pyanaconda_tests/modules/storage/platform_test.py#L120

sharkcz:
ok, will fix the tests too :-)

poncovka (approved):
Looks good to me. Thanks!

VladimirSlavik (approved):
Thank you!

sharkcz:
please don't merge yet, one more fix is required :-)

sharkcz:
Now I hope I got it all right from the bz.

sharkcz:
Maybe one more thing, won't be the explicit "require" for the bootloader sufficient? Allowing to the drop the platform one?

poncovka (approved):
Still looks good to me.

poncovka:
> Maybe one more thing, won't be the explicit "require" for the bootloader sufficient? Allowing to the drop the platform one?

The bootloader can be disabled and I guess that we don't install the bootloader requirements in that case.

VladimirSlavik:
One more nitpick, for commit messages we use `Resolves: rhbz#NNNN`.

sharkcz:
this can be unblocked with the last revision, IMO

poncovka (approved):
It looks great! Thanks!

PR #3112, opened 2021-01-22
network: fix bond confiuration for empty bond options (#1918744)

rvykydal:
Add another guard for addin empty ('') option as NM will now allow adding it to
a connection setting and crash later when actually adding the connection.


poncovka (dismissed):
Looks good to me. Thanks!

VladimirSlavik (approved):
Looks good to me. Thank you!

poncovka (approved):

PR #3113, opened 2021-01-22
Gate contributors by /test comment even for master PRs

jkonecny12:
We have to gate everything not created by us. This PR should solve that.

What happened here:
* Create new `owners-validate` workflow to run tests without gate for us.
* Join `validate` and `validate-rhel-8` to a new `contributors-validate`. This new one will provide gate for all PRs created by contributors for all our supported branches.
* Fix a few small issues and make the workflow files consistent.
 * Add missing `fail-fast` for `rpm-tests`
 * Fix that `rpm-tests` logs from master was overwritten by the ELN run.
 * Make the test names pretty and more consistent with owners run.
 * Use environment variables more to avoid changing the value on multiple places.

Tested:
[PR created by owner](https://github.com/Test-anaconda-org/anaconda/pull/13)
[PR created by contributor to master](https://github.com/Test-anaconda-org/anaconda/pull/14)
[PR created by contributor to rhel-8](https://github.com/Test-anaconda-org/anaconda/pull/15)


rvykydal (approved):
Looks good to me.

.github/workflows/owners-validate.yml
```diff
@@ -1,20 +1,49 @@
-name: validate
+name: Run validation tests for owners
 on: [push, pull_request_target]
+
 jobs:
+  pr-info:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Query comment author repository permissions
```
rvykydal, line 8: Seels to me that we are querying something different than comment author here IIUIC

martinpitt (approved):
There is a lot of code duplication now, I wonder if these can be factorized a bit more elegantly. But surely it's more important/urgent to make them safe.

I'm just a bit confused about the workflow file renames -- surely the main point of them is still to run tests, so how about naming them "tests-owners.yml" and "tests-contributors.yml"?

Thanks for cleaning up my mess :-)

jkonecny12:
Agree on the name change. I'll fix that.

jkonecny12:
> There is a lot of code duplication now, I wonder if these can be factorized a bit more elegantly. But surely it's more important/urgent to make them safe.

Yes, I've got to the same conclusion. It could be nice to reduce the duplication more. However, I think we have a more priority issues to solve right now. I'll make a TODO note to the workflow.

.github/workflows/owners-validate.yml
jkonecny12: Fixed, good point!

PR #3115, opened 2021-01-22
Mount the /tmp of the installed system as a tmpfs (#1306452)

poncovka:
If there is no mount point for /tmp defined in the storage configuration,
mount /tmp of the installed system system as a tmpfs before the remaining
system configuration.

Resolves: rhbz#1306452

### DIFF
--- a/pyanaconda/modules/storage/devicetree/fsset.py
+++ b/pyanaconda/modules/storage/devicetree/fsset.py
@@ -503,6 +503,27 @@ class FSSet(object):
                 else:
                     break
 
+    def collect_filesystems(self):
+        """Collect the system's filesystems.
+
+        :return: a list of devices
+        """
+        devices = list(self.mountpoints.values())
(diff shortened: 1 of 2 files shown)


jkonecny12 (commented):
Looks good to me so far.

pyanaconda/modules/storage/devicetree/fsset.py
```diff
@@ -305,7 +316,9 @@ def __init__(self, devicetree):
     @property
     def system_filesystems(self):
         if not self._system_filesystems:
-            self._system_filesystems = get_system_filesystems()
+            self._system_filesystems = get_system_filesystems(
+                self.devicetree
+            )
```
jkonecny12, line 321: I think this should be part of the previous commit.

poncovka:
Rebased and added some tests.

jkonecny12 (approved):
LGTM!

PR #3116, opened 2021-01-22
Prevent shell injection from a /kickstart-test comment

jkonecny12:
Users could inject shell code in the command of the /kickstart-test comment.
This code will prevent that.

Tested on [PR](https://github.com/Test-anaconda-org/anaconda/runs/1750151768?check_suite_focus=true). The failure is not important, parsing is happening before the failure.

### DIFF
--- a/.github/workflows/kickstart-tests.yml
+++ b/.github/workflows/kickstart-tests.yml
@@ -37,9 +37,13 @@ jobs:
 
       - name: Parse launch arguments
         id: parse_launch_args
+        # Do not use comment body directly in the shell command to avoid possible code injection.
+        env:
+          BODY: ${{ github.event.comment.body }}
         run: |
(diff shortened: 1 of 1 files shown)


M4rtinK (approved):
Looks good to me.

VladimirSlavik:
I don't understand how this could work? You just shuffle the contents around.

jkonecny12:
Kickstart tests are not meaningful to this change. I'm ignoring it here.

jkonecny12:
Properly tested [here]https://github.com/Test-anaconda-org/anaconda/runs/1750242843?check_suite_focus=true)

jkonecny12:
> I don't understand how this could work? You just shuffle the contents around.

The environment variable seems to be preventing the resolution of the command. I'll try to find out more but in general this was a suggestion from the GH security labs.

PR #3117, opened 2021-01-25
Warn users that commands without `inst.` prefix will be removed

jkonecny12:
This is the backport of warnings on master. I've adjusted the message a bit for users on RHEL.

I've tried my best to create a message for users that these won't be removed on the existing RHEL. If you have any suggestions for improvement of the message please write your ideas here.

This is a best effort solution. I don't want to change the code much just for the warnings. In general users will receive this warning on every deprecated argument not just for the ones with missing `inst.` prefix.

*Resolves: rhbz#1897657*

### DIFF
--- a/anaconda.py
+++ b/anaconda.py
@@ -301,11 +301,11 @@ if __name__ == "__main__":
     if os.path.exists("/tmp/updates"):
         log.info("Using updates in /tmp/updates/ from %s", opts.updateSrc)
 
-    # TODO: uncomment this when we're sure that we're doing the right thing
-    # with kernel_arguments *everywhere* it appears...
-    #for arg in depr:
-    #    stdout_log.warn("Boot argument '%s' is deprecated. "
-    #                   "In the future, use 'inst.%s'.", arg, arg)
+    # warn users that they should use inst. prefix all the time
+    for arg in depr:
+        stdout_log.warning("Boot argument '%s' is deprecated and will be removed in the future. "
+                           "Please use 'inst.%s' instead.",
+                           arg, arg)
 
     from pyanaconda import isys
 
--- a/anaconda.py
+++ b/anaconda.py
@@ -303,9 +303,12 @@ if __name__ == "__main__":
 
     # warn users that they should use inst. prefix all the time
     for arg in depr:
-        stdout_log.warning("Boot argument '%s' is deprecated and will be removed in the future. "
+        stdout_log.warning("Deprecated boot argument '%s' must be used with the 'inst.' prefix. "
                            "Please use 'inst.%s' instead.",
                            arg, arg)
+    if depr:
+        stdout_log.warning("Anaconda boot arguments without 'inst.' prefix have been deprecated "
+                           "and will be removed in a future major release.")
 
     from pyanaconda import isys
 
--- a/dracut/parse-anaconda-options.sh
+++ b/dracut/parse-anaconda-options.sh
@@ -45,7 +45,8 @@ set_http_header "X-Anaconda-System-Release" "$product"
 warn_renamed_arg() {
     local arg=""
     arg="$(getarg $1)" && warn "'$1=$arg'" && \
-        warn "$1 has been deprecated and will be removed. Please use $2 instead."
+        warn "$1 has been deprecated. All usage of Anaconda boot arguments without 'inst.' prefix \
+have been deprecated and will be removed in a future major release. Please use $2 instead."
 }
 
 # check for deprecated arg, warn user, and write new arg to /etc/cmdline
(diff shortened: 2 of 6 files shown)


VladimirSlavik (approved):
Looks good to me. Thank you!

jstodola:
Jirko,
I like the idea with clarifying that the boot arguments without the `inst.` prefix won't be removed from the current major RHEL version.
On the other hand, the "major release" patch is only applicable to the rhel branches and not to master - that's one more change to remember to do for the future RHEL releases.
Also, the text mentions "RHEL" - which might not be wanted for products like RHV or CentOS Stream.

The best solution from my POV would be to print the "major release" message only when the product is "Red Hat Enterprise Linux" or the base product is "Red Hat Enterprise Linux". And print the corresponding product name instead of "RHEL".

Another solution would be to make the message more general and usable for any product (Fedora/RHEL/..)
Or just keep the message as it is in the master branch.

jkonecny12:
Thanks a lot for review!

> Jirko,
> I like the idea with clarifying that the boot arguments without the `inst.` prefix won't be removed from the current major RHEL version.
> On the other hand, the "major release" patch is only applicable to the rhel branches and not to master - that's one more change to remember to do for the future RHEL releases.

This won't be backported. The master implementation is there for some time already and I even have accepted further step by FESCO to completely remove support for boot options without `.inst` prefixes.

> Also, the text mentions "RHEL" - which might not be wanted for products like RHV or CentOS Stream.
> 

Good point!

> The best solution from my POV would be to print the "major release" message only when the product is "Red Hat Enterprise Linux" or the base product is "Red Hat Enterprise Linux". And print the corresponding product name instead of "RHEL".
> 
> Another solution would be to make the message more general and usable for any product (Fedora/RHEL/..)
> Or just keep the message as it is in the master branch.

Already replied for the master branch. So I would go with general message. I think the best solution will be to just remove `RHEL` word from the message. I'll do an update of this PR.

jkonecny12:
**UPDATED:**
* Removed `RHEL` from the message.

jstodola:
It looks good to me, thanks.

anaconda.py
```diff
@@ -303,7 +303,9 @@ def setup_environment():
 
     # warn users that they should use inst. prefix all the time
     for arg in depr:
-        stdout_log.warning("Boot argument '%s' is deprecated and will be removed in the future. "
+        stdout_log.warning("Boot argument '%s' has been deprecated same as all Anaconda boot "
+                           "arguments without 'inst.' prefix. "
+                           "They will be removed with a future major release. "
```
VladimirSlavik: How about:

_Boot argument '%s' must be used with the 'inst.' prefix. Please use 'inst.%s' instead. All usage of Anaconda boot arguments without 'inst.' prefix has been deprecated and will be removed in a future major release._

That said, I wonder if you could leave out the sentence "All usage..." and print it only once at the end, if `depr` is not empty. It would drastically improve readability of the output, as the warnings would probably fit on a single line.

dracut/parse-anaconda-options.sh
```diff
@@ -45,7 +45,8 @@ set_http_header "X-Anaconda-System-Release" "$product"
 warn_renamed_arg() {
     local arg=""
     arg="$(getarg $1)" && warn "'$1=$arg'" && \
-        warn "$1 has been deprecated and will be removed. Please use $2 instead."
+        warn "$1 has been deprecated same as all Anaconda boot arguments without 'inst.' prefix. \
+They will be removed with a future major release. Please use $2 instead."
```
VladimirSlavik: Let's try to reuse as much as possible?

_$1 has been deprecated. All usage of Anaconda boot arguments without 'inst.' prefix has been deprecated and will be removed in a future major release. Please use $2 instead._

It's a bit trickier since this is shown also for options that do not fit the pattern "just add inst. in front": `kssendsn` -> `inst.ks.sendsn`. But let's not sweat it, the sentence is still true.

jkonecny12:
**UPDATED:**
* Implemented suggestion from @VladimirSlavik . They are great thanks!

jkonecny12:
Current solution looks like this:
anaconda-inst-prefix-warning
anconda-dracut-inst-warning

anaconda.py
```diff
@@ -303,9 +303,12 @@ def setup_environment():
 
     # warn users that they should use inst. prefix all the time
     for arg in depr:
-        stdout_log.warning("Boot argument '%s' is deprecated and will be removed in the future. "
+        stdout_log.warning("Deprecated boot argument '%s' must be used with the 'inst.' prefix. "
                            "Please use 'inst.%s' instead.",
                            arg, arg)
+    if depr:
+        stdout_log.warning("Anaconda boot arguments without 'inst.' prefix has been deprecated "
```
VladimirSlavik, line 310: arguments = plural -> have
(other place has usage = singular)

VladimirSlavik (approved):
Looks good to me. Thank you!

jkonecny12:
**UPDATED:**
* Fix minor grammar error.

VladimirSlavik (approved):

poncovka (approved):
It looks great! Thanks!

PR #3118, opened 2021-01-25
Fix the should_run methods of the network spoke

poncovka:
The network spoke should use the same logic for its visibility in TUI and GUI.
This might finally fix the initial setup tests in https://github.com/rhinstaller/kickstart-tests/pull/464.


VladimirSlavik (approved):
Looks good to me. Thank you!

rvykydal (approved):
Looks good to me.

M4rtinK (approved):

jkonecny12 (approved):
LGTM!

PR #3119, opened 2021-01-25
Fix root password and LUKS passphrase visibility toggle (#1911360)

M4rtinK:
Anaconda supports showing a little "eye" icon in the password/passphrase
entry field that can be used to make the entered text visible in
plaintext during entry.

For this to work correctly entry text visibility state needs to be
set correctly. This has been broken by some of the password/passphrase
entry related refactorings in the recent past, resulting in the "eye"
icon not being visible.

So put the initialization back for the root password and LUKS passphrase
entries and also extend the docstring in the user configuration spoke
to note how password visibility influences the toggle icon.

Resolves: rhbz#1911360

### DIFF
--- a/pyanaconda/ui/gui/spokes/lib/passphrase.py
+++ b/pyanaconda/ui/gui/spokes/lib/passphrase.py
@@ -90,6 +90,12 @@ class PassphraseDialog(GUIObject):
         self._checker.add_check(self._ascii_check)
         self._checker.add_check(self._empty_check)
 
+        # set the visibility of the password entries
+        # - without this the password visibility toggle icon will
+        #   not be shown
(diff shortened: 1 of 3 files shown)


M4rtinK:
So this is how it looks like with the fix:
LUKS_passphrase_visibility_toggle
root_password_spoke_visibility_toggle

poncovka (approved):
Looks good to me. Thanks!

VladimirSlavik (approved):
Looks good to me. Thank you!

rvykydal (approved):
Looks good to me

jkonecny12 (approved):
LGTM!

PR #3122, opened 2021-01-25
Avoid using DockerHub because of limits

jkonecny12:
DockerHub has a limit on number of pulls which is not great for our CI solution.
Instead use official fedora registry for the container image download.


poncovka (approved):
Thanks!

VladimirSlavik (approved):
Looks good to me. Thank you!

rvykydal (approved):

jkonecny12:
I'm skipping smoke tests here because it's completely unrelated.

PR #3124, opened 2021-01-27
Do not use cache for container build

jkonecny12:
Cache will not take dnf calls into account so no Anaconda dependencies are checked for a cache invalidation check. Let's just ignore the cache and build the container always.


VladimirSlavik (approved):
Looks good to me. Thank you!

PR #3127, opened 2021-01-27
Create real updates images

poncovka:
Don't use `/run/install/updates` and `/tmp/updates` for updates. Use the real paths.
It will allow to use updates images for the initial setup testing and debugging.

Some notes:

* Addons are installed at `/usr/share/anaconda/addons`, so they shouldn't use this support.
* So far, I haven't found any documentation that would suggest to use `/run/install/updates` in the updates image.

### DIFF
--- a/anaconda.py
+++ b/anaconda.py
@@ -103,72 +103,12 @@ def exitHandler(rebootData):
 
 def setup_python_updates():
     """Setup updates to Anaconda Python files."""
-    from distutils.sysconfig import get_python_lib
     import gi.overrides
 
     if "ANACONDA_WIDGETS_OVERRIDES" in os.environ:
         for p in os.environ["ANACONDA_WIDGETS_OVERRIDES"].split(":"):
             gi.overrides.__path__.insert(0, os.path.abspath(p))
 
-    # Temporary hack for F18 alpha to symlink updates and product directories
-    # into tmpfs.  To be removed after beta in order to directly use content
-    # from /run/install/ -- JLK
-    for dirname in ("updates", "product"):
-        if os.path.exists("/run/install/%s" % dirname):
-            if os.path.islink("/tmp/%s" % dirname):
-                # Assume updates have already been setup
-                return
-            os.symlink("/run/install/%s" % dirname,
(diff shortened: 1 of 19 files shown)


poncovka:
Rebased and tested.

poncovka:
Can I ask for a review, please?

anaconda.py
```diff
@@ -300,8 +224,9 @@ def setup_environment():
         sys.exit(0)
 
     log.info("%s %s", sys.argv[0], util.get_anaconda_version_string(build_time_version=True))
-    if os.path.exists("/tmp/updates"):
-        log.info("Using updates in /tmp/updates/ from %s", opts.updateSrc)
+
+    if opts.updates_url:
+        log.info("Using updates from: %s", opts.updates_url)
```
rvykydal, line 229: Aren't we missing the log message in case of updates defined in kickstart ?
(And if so, do we care?)

poncovka: I guess we can log both. I will fix that in a follow-up PR.

scripts/makeupdates
```diff
@@ -32,6 +32,9 @@ import tempfile
 RPM_FOLDER_NAME = os.path.expanduser("~/.anaconda_updates_rpm_cache")
 RPM_RELEASE_DIR_TEMPLATE = "for_%s"
 
+# The Python site-packages path for pyanaconda.
+SITE_PACKAGES_PATH = "./usr/lib64/python3.9/site-packages/"
```
jkonecny12, line 36: I would abstract the `python3.9` string to the separate variable to make the switch in the future easier.

jkonecny12 (approved):
Great cleanup! Thanks a lot for this.
I found just small tweak below which is not mandatory.

M4rtinK (approved):
Looks good to me - a really nice cleanup, we had *a lot* more code for the `/tmp` based updates image workflow thank I thought! Thanks a lot! :)

scripts/makeupdates
poncovka: This is a temporary solution since we have promised to provide a product-independent script that can be used also by Blivet.

PR #3129, opened 2021-01-28
fix text in iscsi loginErrorLabel not wrap

century6:
In the graphical installation mode, if the login of the iSCSI node fails, the interface will prompt an error message. If the error message text does not wrap automatically, the window will extend to the right, and the Cancel button will extend out of the screen, so that the user cannot click to exit the current interface

### DIFF
--- a/pyanaconda/ui/gui/spokes/advstorage/iscsi.py
+++ b/pyanaconda/ui/gui/spokes/advstorage/iscsi.py
@@ -417,6 +417,7 @@ class ISCSIDialog(GUIObject):
             task_proxy.Finish()
         except StorageDiscoveryError as e:
             # Login has failed, show the error.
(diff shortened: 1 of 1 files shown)


VladimirSlavik:
Ouch, something's wrong with the tests again.

Regardless, I wonder - should this be set in the glade file instead?

century6:
> Ouch, something's wrong with the tests again.
> 
> Regardless, I wonder - should this be set in the glade file instead?

You're right. I found that the problem has been fixed in iscsi.glade. It's my negligence. I'm sorry

poncovka:
I am closing this pull request based on https://github.com/rhinstaller/anaconda/pull/3129#issuecomment-771597046. Thanks for checking it!

PR #3133, opened 2021-02-01
Don't initialize the software selection if the payload is not set up (#1916114)

poncovka:
We reset the DNF base if the payload cannot be set up, so it is not ready for
the initialization and validation of the software selection.

Resolves: rhzb#1916114

### DIFF
--- a/pyanaconda/ui/tui/spokes/software_selection.py
+++ b/pyanaconda/ui/tui/spokes/software_selection.py
@@ -88,6 +88,18 @@ class SoftwareSpoke(NormalTUISpoke):
         """Initialize the spoke in a separate thread."""
         threadMgr.wait(THREAD_PAYLOAD)
 
+        # Initialize and check the software selection.
+        self._initialize_selection()
+
+        # Report that the software spoke has been initialized.
+        self.initialize_done()
+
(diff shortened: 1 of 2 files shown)


rvykydal (approved):
Looks good to me.

VladimirSlavik (approved):
Looks good to me. Thank you!

pyanaconda/ui/tui/spokes/software_selection.py
```diff
@@ -198,11 +198,6 @@ def refresh(self, args=None):
         threadMgr.wait(THREAD_PAYLOAD)
         self._container = None
 
-        if not self.payload.base_repo:
-            message = TextWidget(_("Installation source needs to be set up first."))
```
jkonecny12, line 202: Wouldn't users miss this message?

jkonecny12 (approved):
Looks good to me. I just have one question below.

pyanaconda/ui/tui/spokes/software_selection.py
poncovka: Good point. I didn't realize that, in TUI, you can enter spokes that are not ready. Fixed.

PR #3136, opened 2021-02-02
Remove support for no inst prefix

jkonecny12:
Anaconda kernel boot arguments without `inst.` prefix won't be accepted and used. Only warning will be printed that these were removed and you should use the `inst.` variant.

This will happen in Dracut and also in Anaconda.

This is implementation of [Fedora system wide change](https://fedoraproject.org/wiki/Changes/Ignore_Anaconda_kernel_boot_parameters_without_inst_prefix).

### DIFF
--- a/anaconda.py
+++ b/anaconda.py
@@ -184,7 +184,7 @@ def parse_arguments(argv=None, boot_cmdline=None):
     ap = getArgumentParser(get_anaconda_version_string(), boot_cmdline)
 
     namespace = ap.parse_args(argv, boot_cmdline=boot_cmdline)
-    return (namespace, ap.deprecated_bootargs)
+    return (namespace, ap.removed_no_inst_bootargs)
 
 def setup_python_path():
     """Add items Anaconda needs to sys.path."""
@@ -262,7 +262,7 @@ if __name__ == "__main__":
     # do this early so we can set flags before initializing logging
     from pyanaconda.flags import flags
     from pyanaconda.core.kernel import kernel_arguments
-    (opts, depr) = parse_arguments(boot_cmdline=kernel_arguments)
+    (opts, removed_no_inst_args) = parse_arguments(boot_cmdline=kernel_arguments)
 
     from pyanaconda.core.configuration.anaconda import conf
     conf.set_from_opts(opts)
@@ -304,10 +304,13 @@ if __name__ == "__main__":
         log.info("Using updates in /tmp/updates/ from %s", opts.updateSrc)
 
     # warn users that they should use inst. prefix all the time
-    for arg in depr:
-        stdout_log.warning("Boot argument '%s' is deprecated and will be removed in the future. "
-                           "Please use 'inst.%s' instead.",
+    for arg in removed_no_inst_args:
+        stdout_log.warning("Kernel boot argument '%s' detected. "
+                           "Did you want to use 'inst.%s' for the installer instead?",
                            arg, arg)
+    if removed_no_inst_args:
+        stdout_log.warning("All Anaconda kernel boot arguments are now required to use "
+                           "'inst.' prefix!")
 
     from pyanaconda import isys
 
--- a/dracut/parse-anaconda-options.sh
+++ b/dracut/parse-anaconda-options.sh
@@ -58,6 +58,12 @@ check_removed_arg() {
     fi
 }
 
+check_removed_no_inst_arg() {
+    local removed_arg="$1" new_arg="$2"
+    check_removed_arg "$removed_arg" "All usage of Anaconda boot arguments without 'inst.' prefix \
+was removed. Please use $new_arg instead."
+}
+
 check_depr_args "blacklist=" "inst.blacklist=%s"
 check_depr_arg "nofirewire" "inst.blacklist=firewire_ohci"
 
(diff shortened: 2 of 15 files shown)


anaconda.py
```diff
@@ -304,9 +304,8 @@ def setup_environment():
         log.info("Using updates in /tmp/updates/ from %s", opts.updateSrc)
 
     # warn users that they should use inst. prefix all the time
-    for arg in depr:
-        stdout_log.warning("Boot argument '%s' is deprecated and will be removed in the future. "
-                           "Please use 'inst.%s' instead.",
+    for arg in removed_no_inst_args:
+        stdout_log.warning("Boot argument '%s' was removed. Please use 'inst.%s' instead.",
```
poncovka, line 308: I don't think we can still claim these boot options as ours.

dracut/parse-anaconda-options.sh
```diff
@@ -58,6 +58,12 @@ check_removed_arg() {
     fi
 }
 
+check_removed_no_inst_arg() {
+    local removed_arg="$1" new_arg="$2"
+    check_removed_arg "$removed_arg" "All usage of Anaconda boot arguments without 'inst.' prefix \
+was removed. Please use $new_arg instead."
```
poncovka, line 64: Same here.

poncovka (commented):
Since we are dropping the support for the boot options without the `inst.` prefix, we can just remove everything that is related to that. I don't think we should continue to show warnings, because these options might belong to someone else now.

jkonecny12:
I agree but I would like to leave it there for at least some time and make another PR with complete removal in the future. It's one small thing we can do for our consumers.

poncovka (approved):
Looks good to me. Thanks! Please, open an issue for the removal of the warnings about the removal :-P

anaconda.py
```diff
@@ -262,7 +262,7 @@ def setup_environment():
     # do this early so we can set flags before initializing logging
     from pyanaconda.flags import flags
     from pyanaconda.core.kernel import kernel_arguments
-    (opts, depr) = parse_arguments(boot_cmdline=kernel_arguments)
+    (opts, removed_no_inst_args) = parse_arguments(boot_cmdline=kernel_arguments)
```
VladimirSlavik, line 265: Is this correct? There will be no more deprecated arguments?

jstodola:
Hi, this change will likely break installations of new composes in Beaker. Could you please wait before merging this PR?
Bug https://bugzilla.redhat.com/show_bug.cgi?id=1907566 tracks the Beaker change.
Thank you :-)

dracut/parse-anaconda-options.sh
```diff
@@ -77,35 +83,35 @@ check_removed_arg askmethod "Use an appropriate 'inst.repo=' argument instead."
...
 
 # lang & keymap
-warn_renamed_arg "lang" "inst.lang"
-warn_renamed_arg "keymap" "inst.keymap"
+check_removed_no_inst_arg "lang" "inst.lang"
+check_removed_no_inst_arg "keymap" "inst.keymap"
 
 # debug
-warn_renamed_arg "debug" "inst.debug"
+check_removed_no_inst_arg "debug" "inst.debug"
```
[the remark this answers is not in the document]
jkonecny12: Good point. Thanks for input!

anaconda.py
jkonecny12: It wasn't used for anything else than just `inst.` prefix so I don't see a reason to keep it universal anymore. My next step will be to remove this completely.

jkonecny12:
I'll give them a bit more time but as I wrote to bugzilla, we can't wait too long. It has to be merged before Tue 2021-02-09 because of change completion deadline.

dracut/parse-anaconda-options.sh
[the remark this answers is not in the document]
jkonecny12: Fixed.

jkonecny12:
**UPDATED:**
* Remove `debug` warning noted by @mopsfelder . Thanks!
* Add bug to commits.

VladimirSlavik (approved):

jkonecny12:
> Looks good to me. Thanks! Please, open an issue for the removal of the warnings about the removal :-P

Issue opened https://issues.redhat.com/browse/INSTALLER-2363 .

jkonecny12:
**UPDATED:**
* Update warning in stage2 to be more generic. The warning is now suggest to user instead of telling him that he's using the parameter wrong. Point is that this argument may not be ours (like in case of `debug`).

jkonecny12:
@VladimirSlavik could you please look on the new message in commit c16d17d if it makes sense to you? Or do you have a suggestion for improvement?

VladimirSlavik:
The commit message looks to me perfectly understandable.

poncovka (approved):
It looks great! Thanks!

anaconda.py
```diff
@@ -304,10 +304,13 @@ def setup_environment():
         log.info("Using updates in /tmp/updates/ from %s", opts.updateSrc)
 
     # warn users that they should use inst. prefix all the time
-    for arg in depr:
-        stdout_log.warning("Boot argument '%s' is deprecated and will be removed in the future. "
-                           "Please use 'inst.%s' instead.",
+    for arg in removed_no_inst_args:
+        stdout_log.warning("The '%s' kernel boot argument detected. "
+                           "Did you want to use 'inst.%s' instead?",
```
VladimirSlavik, line 309: Sorry, now I see what you meant. How about:

Kernel boot argument '%s' detected. Did you want to use 'inst.%s' for the installer instead?

jkonecny12: Fixed.

jkonecny12:
**UPDATED:**
* Update message as suggested by @VladimirSlavik . Thanks!

VladimirSlavik (approved):

poncovka (approved):

jkonecny12:
These coverage test failures are not because of this change.

PR #3137, opened 2021-02-03
Fix getting kernel version list for liveimg

rvykydal:
Ignore rescue kernels same as in liveos payload.

Resolves: rhbz#1919463


poncovka (approved):
Looks good to me.

VladimirSlavik (approved):
Looks good to me. Thank you!

PR #3145, opened 2021-02-04
Don't run a canceled task

poncovka:
Before running a task, make sure that the task has not been canceled.

The `run_with_signals` method needs to be able to handle this use case
as well, so reorganize the code to support it.


VladimirSlavik (approved):
Looks good to me. Thank you!

PR #3146, opened 2021-02-04
Don't run installation tasks of add-ons in a meta task

poncovka:
The UI needs be able to handle a failure of an installation task and possibly
continue with the installation if the user allows it. That is not possible if
we run all installation tasks in one meta task.

### DIFF
--- a/pyanaconda/modules/boss/boss_interface.py
+++ b/pyanaconda/modules/boss/boss_interface.py
@@ -17,22 +17,35 @@
 # License and may only be used or replicated with the express permission of
 # Red Hat, Inc.
 #
+from dasbus.client.proxy import get_object_handler
 from dasbus.server.interface import dbus_interface
+from dasbus.typing import *  # pylint: disable=wildcard-import
+
 from pyanaconda.modules.common.constants.services import BOSS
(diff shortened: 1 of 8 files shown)


pyanaconda/modules/boss/boss_interface.py
```diff
@@ -17,22 +17,35 @@
...
 from pyanaconda.modules.common.containers import TaskContainer
 from pyanaconda.modules.common.structures.kickstart import KickstartReport
 from pyanaconda.modules.common.structures.requirement import Requirement
+from pyanaconda.modules.common.typing import BusName
 
 __all__ = ["BossInterface"]
 
 
+def get_proxy_identification(proxy):
+    """Get a service name and on object path of the given DBus proxy.
```
rvykydal, line 35: A typo ? "on object path" -> "an object path"

rvykydal (approved):
Looks good to me.

VladimirSlavik (approved):
Looks good to me. Thank you!

pyanaconda/modules/boss/boss_interface.py
poncovka: Fixed.

M4rtinK (approved):
Looks good to me as well. :)

PR #3147, opened 2021-02-04
Add support for non-critical installation errors

poncovka:
The OSCAP add-on needs to be able to raise a non-critical installation error.
In that case, the users should be allowed to continue with the installation
if they want to.

### DIFF
--- a/pyanaconda/errors.py
+++ b/pyanaconda/errors.py
@@ -19,7 +19,7 @@
 from pyanaconda.core.i18n import _, C_
 from pyanaconda.flags import flags
 from pyanaconda.modules.common.errors.installation import BootloaderInstallationError, \
-    StorageInstallationError
+    StorageInstallationError, NonCriticalInstallationError
 from pyanaconda.modules.common.errors.payload import SourceSetupError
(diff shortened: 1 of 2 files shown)


pyanaconda/errors.py
```diff
@@ -205,6 +208,16 @@ def _bootloader_error_handler(self, exn):
         else:
             return ERROR_RAISE
 
+    def _non_critical_error_handler(self, exn):
+        message = _("The following error occurred during the installation:")
+        message += "\n\n" + str(exn) + "\n\n"
+        message = _("Would you like to ignore this and continue with installation?")
```
VladimirSlavik, line 214: As a translator, I would definitely prefer one string to many small pieces. If it's not inconvenient, please make it one string with format or concatenation, as in the handlers above.

VladimirSlavik (commented):
Looks good to me so far.

pyanaconda/errors.py
poncovka: Fixed.

poncovka:
uups

VladimirSlavik (approved):

PR #3151, opened 2021-02-08
Unify GRUB configuration file location across all platforms

martinezjavier:
The GRUB configuration files layout on EFI platforms isn't consistent with
other non-EFI platforms (e.g: legacy BIOS x86 and Open Firmware ppc64le).

On platforms using EFI, the GRUB config file (grub.cfg) and environment
variables block (grubenv) are stored in the EFI System Partition (ESP),
while for non-EFI platforms these are stored in the boot partition (or
/boot directory if not boot partition is used).

The reason for this is that the path where the GRUB bootloader searches
for its configuration file varies depending on the firmware interface.

For EFI the GRUB binary is located in the ESP and it expects to find its
config file in that location as well. But this creates the mentioned
inconsistency, because the GRUB configuration file has to be stored in
/boot/efi/EFI/fedora/grub.cfg while for non-EFI platforms it has to be
stored in /boot/grub2/grub.cfg.

To allow all platforms to have the GRUB config file in the same location,
only a minimal config file could be stored in the ESP and this will load
the one that is stored in /boot/grub2.

Resolves: rhbz#1918817

### DIFF
--- a/pyanaconda/modules/storage/bootloader/efi.py
+++ b/pyanaconda/modules/storage/bootloader/efi.py
@@ -35,7 +35,11 @@ class EFIBase(object):
     """A base class for EFI-based boot loaders."""
 
     @property
-    def _config_dir(self):
+    def efi_config_dir(self):
+        return "/boot/" + self._efi_config_dir
+
+    @property
+    def _efi_config_dir(self):
         return "efi/EFI/{}".format(conf.bootloader.efi_dir)
 
     def efibootmgr(self, *args, **kwargs):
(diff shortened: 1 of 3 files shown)


poncovka (commented):
The code looks good to me.

rvykydal (approved):
Looks good to me AFAICT.

gicmo (approved):
FWIW, looks good to me too. :D

poncovka (approved):
Looks good to me. Thanks!

wutaosmile:
I have a problemn with this pull.On uefi platform,when grub.cfg is unified on /boot/grub2,and grubenv is a link to /boot/grub2/EFI/efi,but when exec grubby --set-default=xxxxxx to change default boot kernel,it does not work.How to fix this problem?

PR #3153, opened 2021-02-08
network: wrap IP addresses showed in GUI for wireless devices (#1925781)

rvykydal:
Follow-up of commit fccd58ce29e69b7ed7572d8760c07f62356cbcdb

### DIFF
--- a/pyanaconda/ui/gui/spokes/network.glade
+++ b/pyanaconda/ui/gui/spokes/network.glade
@@ -1088,6 +1088,7 @@ updates available for you.</property>
                                             <property name="can_focus">False</property>
                                             <property name="label" translatable="yes">IPv4 Address</property>
                                             <property name="xalign">1</property>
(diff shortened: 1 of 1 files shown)


VladimirSlavik (commented):
Just two characters? I am not sure how that looks...

rvykydal:
> Just two characters? I am not sure how that looks...

IIRC in this context, which is IP addresses as strings (words) separated by a space it says: one IP address (word) per line.

VladimirSlavik:
No? The [official docs](https://developer.gnome.org/gtk3/stable/GtkLabel.html#GtkLabel--max-width-chars) say that this is indeed in characters, just as the raw glade XML says. But there's a [lengthy SO answer](https://stackoverflow.com/questions/27462926/how-to-set-max-width-of-gtklabel-properly#34472033) which suggests this actually works differently.

VladimirSlavik (approved):

PR #3154, opened 2021-02-08
Modify spec file to add patches required on rhel-9 rebuild (#1907566)

jkonecny12:
We made the strong requirement for the 'inst.' prefix for Anaconda kernel boot arguments. However, unfortunately beaker needs more time to adapt and we are forced to merge the changes because of Fedora System Wide Change Completion Deadline. So these patches will enable beaker to work for RHEL-9 for the required time. See the attached bug for more info.

Related: rhbz#1907566

### DIFF
--- a/anaconda.spec.in
+++ b/anaconda.spec.in
@@ -12,6 +12,17 @@ URL:     http://fedoraproject.org/wiki/Anaconda
 # make dist
 Source0: %{name}-%{version}.tar.bz2
 
+# TODO: Remove this as soon as rhbz#1907566 will be solved!
+# These patches will disable requirement of inst. prefix for RHEL-9.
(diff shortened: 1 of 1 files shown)


poncovka (approved):
Looks good to me. Thanks!

jkonecny12:
**UPDATED:**
* Use condition for RHEL-9 only to avoid ELN failures from tomorrow.

jkonecny12:
Tomorrow ELN will switch to 10 so this condition should not be satisfied which means that our ELN tests should work correctly. If not we will add 0%{?eln}.

M4rtinK (approved):
Looks good to me as well - thanks! :)

jstodola:
Thank you for this temporary workaround, Jirko!

PR #3155, opened 2021-02-10
Guess the default product name from the os-release files

poncovka:
Don't use the hard-coded default product. Use the product name from
the os-release files of the installation environment. See the commit
849a018.


rvykydal (approved):
Looks good to me.

VladimirSlavik (approved):
That's clever! Thanks!

PR #3156, opened 2021-02-10
Add a kickstart specification for the main process

poncovka:
Use a kickstart specification to define command and data overrides
for the kickstart handler used by the main process. It means that
we can define commands that will be used instead of commands that
should be ignored.


rvykydal (approved):
Looks good to me.

VladimirSlavik (approved):
Looks good to me. Thank you!

We should celebrate the victory of Useless Commands. Somehow.

poncovka:
Tested locally.

PR #3157, opened 2021-02-10
Use special defaults for kickstart tests on PR as a workaround

rvykydal:
See https://github.com/rhinstaller/kickstart-tests/issues/482
Requires:
- [ ] https://github.com/rhinstaller/kickstart-tests/pull/483
- [ ] testing


rvykydal:
After double-checking I can confirm that the issue has actually the same cause as https://github.com/rhinstaller/kickstart-tests/issues/448 and can be fixed by setting volume id of generated iso:  https://github.com/rhinstaller/anaconda/pull/3158

PR #3158, opened 2021-02-11
Set volume id for iso built for kickstart test

rvykydal:
Use the same volume id as the official rawhide boot.isos we are testing in
daily scenario.  Based on the id virt-install can select os specific
configuration which decides if slot or path based network naming scheme (ensX
vs enp0sY) is used.

See https://github.com/rhinstaller/kickstart-tests/issues/448
and https://github.com/rhinstaller/kickstart-tests/pull/452


poncovka (approved):
Looks good to me. Thanks!

VladimirSlavik (approved):
Looks good to me. Thank you!

jkonecny12 (approved):
Looks good to me.

PR #3160, opened 2021-02-11
Master wrap misc. labels

VladimirSlavik:
Based on ideas from #3010.

Blocked by beta freeze.

~~Currently blocked by the inability to start X build problems.~~

Sample results:

Screenshot_vs_main_2021-02-17_15:13:09

Screenshot_vs_main_2021-02-17_15:11:27

### DIFF
--- a/widgets/src/SpokeSelector.c
+++ b/widgets/src/SpokeSelector.c
@@ -291,6 +291,7 @@ static void anaconda_spoke_selector_init(AnacondaSpokeSelector *spoke) {
     spoke->priv->title_label = gtk_label_new(NULL);
     gtk_label_set_text_with_mnemonic(GTK_LABEL(spoke->priv->title_label), _(DEFAULT_TITLE));
     gtk_label_set_justify(GTK_LABEL(spoke->priv->title_label), GTK_JUSTIFY_LEFT);
+    gtk_label_set_line_wrap(GTK_LABEL(spoke->priv->title_label), TRUE);
     gtk_label_set_mnemonic_widget(GTK_LABEL(spoke->priv->title_label), GTK_WIDGET(spoke));
     gtk_label_set_xalign(GTK_LABEL(spoke->priv->title_label), 0.0);
     gtk_label_set_yalign(GTK_LABEL(spoke->priv->title_label), 1.0);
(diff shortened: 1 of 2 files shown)


poncovka (approved):
Looks good to me.

jkonecny12 (approved):
LGTM!

VladimirSlavik:
Update: Fixed, tested, added screenshots.

widgets/src/SpokeSelector.c
```diff
@@ -291,6 +291,8 @@ static void anaconda_spoke_selector_init(AnacondaSpokeSelector *spoke) {
     spoke->priv->title_label = gtk_label_new(NULL);
     gtk_label_set_text_with_mnemonic(GTK_LABEL(spoke->priv->title_label), _(DEFAULT_TITLE));
     gtk_label_set_justify(GTK_LABEL(spoke->priv->title_label), GTK_JUSTIFY_LEFT);
+    gtk_label_set_line_wrap(GTK_LABEL(spoke->priv->title_label), TRUE);
+    gtk_label_set_max_width_chars(GTK_LABEL(spoke->priv->title_label), 2); // don't grow the tile
```
poncovka, line 295: This is a very confusing setting. Maybe the width could be set to 25 as the status label below?

poncovka (approved):

widgets/src/SpokeSelector.c
VladimirSlavik: OK. It does not matter anyway, as I understand it. This isn't a maximum width in terms of absolutes, but maximal width the widget can expand on its own, or "unconditionally request for itself". It can still expand further, but only if there is space in the parent container. That's what happens here.

(FWIW, 25 would be consistent here, while 2 is the width Radek chose recently for other parts of the GUI. That's why I chose that value originally.)

VladimirSlavik: Actually, I can't do that, or rather there is no point. I just realized that 25 *characters* is a different width for the heading which used larger text. So the consistency would be in fact an illusion :-/

jkonecny12 (approved):
Still looks good.

VladimirSlavik:
I just realized I did not compare the width to that of before. Needs further testing.

VladimirSlavik:
Rebased before re-testing

VladimirSlavik:
This isn't enough as-is. The headings get more breaks which breaks some languages by pushing stuff below the fold.

Screenshot_vs_main_2021-03-02_14:47:15

VladimirSlavik:
Fixed - this version does not break anything, and improves some cases.

poncovka (approved):
Looks good to me.

VladimirSlavik:
Blocked by beta freeze.

PR #3162, opened 2021-02-11
Add missing space to a message

VladimirSlavik:
This very important pull request is about a single byte change. I have no idea why we have to clog all the pipes with this meta cruft, but here we go.


poncovka (approved):
Thanks! :-D

rvykydal (approved):
Looks very much to me.

jkonecny12 (approved):
We need this change NOOOOWWWWW! :D :D

PR #3164, opened 2021-02-12
Rename pyanaconda.modules.common.typing

poncovka:
We cannot have a submodule called `typing`, otherwise the build will fail with
the traceback below. Rename it to `custom_typing` instead.

What happens during the build:

* The make tool enters the `/pyanaconda/modules/common` directory.
* It means that the current directory contains our `typing` module.
* The make tool runs the `py_compile` script.
* The script tries to import the `typing` module from the standard library.
* However, Python searches modules in the current directory first.
* Our `typing` module is imported instead of the standard one.

The traceback from the build:

```
Entering directory '/builddir/anaconda-35.1/pyanaconda/modules/common'
  File "<string>", line 2, in <module>
  File "/usr/lib64/python3.9/py_compile.py", line 9, in <module>
    import importlib.util
  File "/usr/lib64/python3.9/importlib/util.py", line 2, in <module>
    from . import abc
  File "/usr/lib64/python3.9/importlib/abc.py", line 17, in <module>
    from typing import Protocol, runtime_checkable
  File "/builddir/pyanaconda/modules/common/typing.py", line 18, in <module>
    from dasbus.typing import *  # pylint: disable=wildcard-import
ModuleNotFoundError: No module named 'dasbus'
```


VladimirSlavik (approved):

jkonecny12 (approved):
LGTM! Thanks for this hotfix!

M4rtinK (approved):
Oh the wonders of Autotools!

Thanks a lot for the fix - looks good to me! :)

PR #3166, opened 2021-02-12
Save lorax-packages.log to installed system

rvykydal:
The file contains list of packages used to build the installer image.

The original motivation here is having the log in results of kickstart tests but I think it is useful in general.


VladimirSlavik (approved):
Looks good to me. Thank you!

jstodola:
That's a good idea.
But I wonder if it would not make more sense to store the file in `/var/log/anaconda/` instead of `/root/`, even though the file was not created by the installer:

1. To avoid bug reports from users complaining about having an unexpected/unwanted file in /root/ right after a fresh installation.
2. `lorax-packages.log` might be useful to have in bug reports, where the reporter is asked to provide the log files from `/var/log/anaconda/` (for problems not blocking the installation).

rvykydal:
Updated to store the file in /var/log/anaconda as @jstodola suggested.

poncovka (approved):
Looks good to me. Thanks!

PR #3167, opened 2021-02-12
Migrate from pytz to zoneinfo, list (close to) all time zones

hroncok:
https://listman.redhat.com/archives/anaconda-devel-list/2021-February/msg00020.html

This is very much untested. I've written it and never actually attempted to run it. Not yet anyway.

I found a problem thou. Python's `zoneinfo` does not have any concept of *common timezones*. I can implement my own method to determine whether a timezone is *common* (pytz uses [zone.tab](https://github.com/eggert/tz/blob/master/zone.tab) and a [short list of hardcoded values](https://github.com/stub42/pytz/blob/513450fb7e682d6d828456c661058f0e14a922cd/gen_tzinfo.py#L108)).

With pytz 2021.1, the extra zones effectively added by this PR are:

```pycon
>>> import pytz, zoneinfo
>>> pytz_zones = {tz for tz in pytz.common_timezones if '/' in tz and not tz.startswith('Etc/')}
>>> zoneinfo_zones = {tz for tz in zoneinfo.available_timezones() if '/' in tz and not tz.startswith('Etc/')}
>>> len(zoneinfo_zones) / len(pytz_zones)
1.1762013729977117
>>> zoneinfo_zones - pytz_zones
{'Africa/Asmera',
 'Africa/Timbuktu',
 'America/Argentina/ComodRivadavia',
 'America/Atka',
 'America/Buenos_Aires',
 'America/Catamarca',
 'America/Coral_Harbour',
 'America/Cordoba',
 'America/Ensenada',
 'America/Fort_Wayne',
 'America/Godthab',
 'America/Indianapolis',
 'America/Jujuy',
 'America/Knox_IN',
 'America/Louisville',
 'America/Mendoza',
 'America/Montreal',
 'America/Porto_Acre',
 'America/Rosario',
 'America/Santa_Isabel',
 'America/Shiprock',
 'America/Virgin',
 'Antarctica/South_Pole',
 'Asia/Ashkhabad',
 'Asia/Calcutta',
 'Asia/Chongqing',
 'Asia/Chungking',
 'Asia/Dacca',
 'Asia/Harbin',
 'Asia/Istanbul',
 'Asia/Kashgar',
 'Asia/Katmandu',
 'Asia/Macao',
 'Asia/Rangoon',
 'Asia/Saigon',
 'Asia/Tel_Aviv',
 'Asia/Thimbu',
 'Asia/Ujung_Pandang',
 'Asia/Ulan_Bator',
 'Atlantic/Faeroe',
 'Atlantic/Jan_Mayen',
 'Australia/ACT',
 'Australia/Canberra',
 'Australia/Currie',
 'Australia/LHI',
 'Australia/NSW',
 'Australia/North',
 'Australia/Queensland',
 'Australia/South',
 'Australia/Tasmania',
 'Australia/Victoria',
 'Australia/West',
 'Australia/Yancowinna',
 'Brazil/Acre',
 'Brazil/DeNoronha',
 'Brazil/East',
 'Brazil/West',
 'Canada/Saskatchewan',
 'Canada/Yukon',
 'Chile/Continental',
 'Chile/EasterIsland',
 'Europe/Belfast',
 'Europe/Nicosia',
 'Europe/Tiraspol',
 'Mexico/BajaNorte',
 'Mexico/BajaSur',
 'Mexico/General',
 'Pacific/Johnston',
 'Pacific/Ponape',
 'Pacific/Samoa',
 'Pacific/Truk',
 'Pacific/Yap',
 'US/Aleutian',
 'US/East-Indiana',
 'US/Indiana-Starke',
 'US/Michigan',
 'US/Samoa'}
```

That's 17.6% of extra zones.

Please let me know if this actually matters or not.

### DIFF
--- a/pyanaconda/timezone.py
+++ b/pyanaconda/timezone.py
@@ -22,9 +22,10 @@ configuration, valid timezones recognition etc.
 
 """
 
-import pytz
 import langtable
+import zoneinfo
 from collections import OrderedDict
+from functools import cache
 
 from pyanaconda.core import util
 from pyanaconda.core.constants import THREAD_STORAGE
@@ -38,8 +39,7 @@ from blivet import arch
 from pyanaconda.anaconda_loggers import get_module_logger
 log = get_module_logger(__name__)
 
-# The following zones are not in pytz.common_timezones and
-# Etc category in pytz.all_timezones includes some more,
+# The Etc category in zoneinfo.available_timezones() includes some more,
 # however confusing ones (like UCT, GMT+0, GMT-0,...)
 ETC_ZONES = ['GMT+1', 'GMT+2', 'GMT+3', 'GMT+4', 'GMT+5', 'GMT+6', 'GMT+7',
              'GMT+8', 'GMT+9', 'GMT+10', 'GMT+11', 'GMT+12',
@@ -126,6 +126,18 @@ def get_preferred_timezone(territory):
     return timezones[0]
 
 
+@cache
+def all_timezones():
+    """
+    Get all timezones, but with the Etc zones reduced. Cached.
+
+    :rtype: set
+
+    """
+    etc_zones = {"Etc/" + zone for zone in ETC_ZONES}
+    return zoneinfo.available_timezones() | etc_zones
+
+
 def get_all_regions_and_timezones():
     """
     Get a dictionary mapping the regions to the list of their timezones.
@@ -136,7 +148,7 @@ def get_all_regions_and_timezones():
 
     result = OrderedDict()
 
-    for tz in pytz.common_timezones:
+    for tz in sorted(all_timezones()):
         parts = tz.split("/", 1)
 
         if len(parts) > 1:
@@ -144,7 +156,6 @@ def get_all_regions_and_timezones():
                 result[parts[0]] = set()
             result[parts[0]].add(parts[1])
 
-    result["Etc"] = set(ETC_ZONES)
     return result
 
 
@@ -157,9 +168,7 @@ def is_valid_timezone(timezone):
 
     """
 
-    etc_zones = ["Etc/" + zone for zone in ETC_ZONES]
-
-    return timezone in pytz.common_timezones + etc_zones
+    return timezone in all_timezones()
 
 
 def get_timezone(timezone):
@@ -170,4 +179,4 @@ def get_timezone(timezone):
     :rtype: datetime.tzinfo
     """
 
-    return pytz.timezone(timezone)
+    return zoneinfo.ZoneInfo(timezone)
--- a/anaconda.spec.in
+++ b/anaconda.spec.in
@@ -108,7 +108,6 @@ Requires: subscription-manager >= %{subscriptionmanagerver}
 # which is apparently great for containers but unhelpful for the rest of us
 Requires: cracklib-dicts
 
-Requires: python3-pytz
 Requires: teamd
 %ifarch s390 s390x
 Requires: openssh
--- a/pyanaconda/isys/__init__.py
+++ b/pyanaconda/isys/__init__.py
@@ -28,7 +28,7 @@ except ImportError:
 import blivet.arch
 import time
 import datetime
-import pytz
+import zoneinfo
 
 from pyanaconda.anaconda_loggers import get_module_logger
 log = get_module_logger(__name__)
@@ -88,7 +88,7 @@ def set_system_date_time(year=None, month=None, day=None, hour=None, minute=None
 
     """
 
-    utc = pytz.UTC
+    utc = zoneinfo.ZoneInfo(key='UTC')
     # If no timezone is set, use UTC
     if not tz:
         tz = utc
@@ -104,13 +104,12 @@ def set_system_date_time(year=None, month=None, day=None, hour=None, minute=None
     minute = minute if minute is not None else now.minute
     second = second if second is not None else now.second
 
-    set_date = tz.localize(datetime.datetime(year, month, day, hour, minute, second))
+    set_date = datetime.datetime(year, month, day, hour, minute, second, tzinfo=tz)
 
     # Calculate the number of seconds between this time and timestamp 0
-    # see pytz docs, search for "Converting between timezones"
     # pylint bug here: https://github.com/PyCQA/pylint/issues/1104
     # pylint: disable=no-value-for-parameter
-    epoch = utc.localize(datetime.datetime.utcfromtimestamp(0)).astimezone(tz)
+    epoch = datetime.datetime.fromtimestamp(0, tz=utc).astimezone(tz)
     timestamp = (set_date - epoch).total_seconds()
 
     set_system_time(int(timestamp))


VladimirSlavik:
> I found a problem thou. Python's zoneinfo does not have any concept of common timezones.

I have run into at least one RFE that depends on *not* simplifying the zone list to the common ones. (The Montreal thing by @lsatenstein.) Unless we find some technical or usability reason to simplify the list, I would be most happy to see the full list.

The main doubt I had was the ability to map the zones back onto the map-of-the-world thingy.

VladimirSlavik (changes requested):
Please rebase and re-target to f34-devel? There's nothing on master so it should be no work at all.

VladimirSlavik:
Tests are mostly broken right now, so no point enabling them :(

hroncok:
You want this in f34 already? Fine by me.

hroncok:
> Tests are mostly broken right now.

How do I check easiest if this even runs? I see you use packit. Does it produce rpms I can install in a Fedora 34 VM?

hroncok:
The ELN copr build fails because https://pagure.io/releng/issue/10007

hroncok:
Before this PR:

Anaconda scsreenshot before this PR

After this PR:

Anaconda scsreenshot after this PR

The UI seems to do weird things, e.g. if I try to select Australia/Canberra, it selects Australia/Sydney etc. Or if I select United States, it gives me Americas instead and I need to to select United States 3 to 4 times in order to actually see the options there. Not sure if it is related to my PR or not.

VladimirSlavik:
I think that's the map widget, just as I feared - it doesn't know where the time zones actually lie.

https://github.com/dashea/timezonemap

VladimirSlavik:
Sorry for recurrent brain dumps, but here I go once more. I think there's some way to tell if a TZ is an alias or not - that might be the key to getting rid of the auto-switching ones. IIRC I saw code for such filtering in Anaconda, so this might be a similar but subtly different problem.

VladimirSlavik:
Oh look, here we go: https://github.com/dashea/timezonemap/blob/master/src/data/backward Explains the Canberra -> Sydney at least a bit.

pyanaconda/timezone.py
```diff
@@ -136,15 +148,14 @@ def get_all_regions_and_timezones():
 
     result = OrderedDict()
 
-    for tz in pytz.common_timezones:
+    for tz in sorted(all_timezones()):
```
VladimirSlavik, line 151: It might be better to move the sorting into function itself to take advantage of the cache.

hroncok: Technically, it could. However the other function that uses this does not need sorting and it appeared to me that `get_all_regions_and_timezones()` is usually called just once? If it is called multiple times, should that be cached as well?

M4rtinK:
@hroncok Thanks for the PR! Will try to respond to individual points bellow. :)

* extra timezones

That's a good point, I wonder if just having the extra timezones would be a problem - the list in the GUI is already long, but 17% extra zones would hardly make it worse. Also as @VladimirSlavik mentioned, there have been RFEs about missing zones.

If we really need to have some custom list of zones to include or exclude for some reason, I think it should really not be hardcoded in Anaconda source code but instead live in langtable (https://github.com/mike-fabian/langtable) like all similar language/locale/keyboard mapping things we use. Still in such a case I would arrange for that, can't really burden you with that extra work. :)

* testing the changes

Seeing the screenshots I see you got it working, but for the record - for interactive debugging like this editing the Anaconda files on the live image is likely best. It should be possible to start sshd on a booted live image and overwrite the Anaconda files as needed. The just relaunch Anaconda using the launcher icon after every edit. 

Alternatively it's possible to use the [updates image mechanism](https://fedoraproject.org/wiki/Anaconda/Updates). That's a bit more involved but can be automated, starts with clean environment every time and can be used to test the netinst image. I can get you started on this if you want. :)

* map widget issues

Let's say that this has been fragile and full of edge cases already. :P Still I guess the only thing really changing is that some timezones might be called differently & there could be more of them. And if there are already some mappings in the timezonemap projects maybe these could be adjusted to match the new names ?

hroncok:
What do I do here?

VladimirSlavik:
@hroncok I'm not sure, but we'll figure it out. Unfortunately I'm not able to start digging into this right away.

Assuming the widget really needs some dusting off and updating: Somebody would have to figure out what goes bonkers in that thing due to missing data, and get a solid idea what needs to be fixed there. Then we can start with the PRs... I am aware that that's far more than just rewriting bits of anaconda, so maybe that would be a task for me?

VladimirSlavik:
Rebased to master, force-pushed, retargeted. May need another force-push to re-run tests and clean up GH statuses, it's a total mess :(

VladimirSlavik:
Not sure what the codecov problem is, it links to some cute but generic error message.

VladimirSlavik (approved):
Forgot this bit.

jkonecny12 (approved):
LGTM!

hroncok:
Awesome!

VladimirSlavik:
Haha, you are too.

AdamWill:
I think this has caused a problem with daylight savings. openQA is showing the clock exactly an hour fast on fresh installs; our official openQA deployments are in the Eastern timezone, which uses daylight savings. This is, I think, causing two tests to fail (in non-obvious ways it took me a while to trace back to the wrong time and then to this commit). These problems started in exactly the compose this commit ultimately landed in - Fedora-Rawhide-20210520.n.1.

I'm looking through the changes now to see if I can spot the issue - I think I actually touched the `set_system_date_time` code last before this PR - but it's subtle stuff and I forget the details...

AdamWill:
OK, yeah, it's definitely this commit. Here's a test script I ginned up to compare `set_system_date_time` from before this PR and `set_system_date_time` from after it, using three timezones: UTC, US/Eastern, and Asia/Kolkata. Both versions get the right timestamp for UTC and Asia/Kolkata, because neither of those timezones has daylight savings...but the new version gets the wrong timestamp for US/Eastern, which does do daylight savings.

    import datetime
    import pytz
    import time
    import zoneinfo

    def get_timezone_new(timezone):
        """
        Return a tzinfo object for a given timezone name (new way).
        """
        return zoneinfo.ZoneInfo(timezone)

    def get_timezone_old(timezone):
        """
        Return a tzinfo object for a given timezone name (old way).
        """
        return pytz.timezone(timezone)


    def set_system_date_time_new(tz=None):
        """
        Set system date and time to "now" (new way).
        """
        utc = zoneinfo.ZoneInfo(key='UTC')
        # If no timezone is set, use UTC
        if not tz:
            tz = utc
        else:
            tz = get_timezone_new(tz)

        time.tzset()

        # get the right values
        now = datetime.datetime.now(tz)
        set_date = datetime.datetime(now.year, now.month, now.day, now.hour, now.minute, now.second, tzinfo=tz)

        # Calculate the number of seconds between this time and timestamp 0
        epoch = datetime.datetime.fromtimestamp(0, tz=utc).astimezone(tz)
        timestamp = (set_date - epoch).total_seconds()
        print(f"New timestamp: {timestamp}")

    def set_system_date_time_old(tz=None):
        """
        Set system date and time to "now" (old way).
        """

        utc = pytz.UTC
        # If no timezone is set, use UTC
        if not tz:
            tz = utc
        else:
            tz = get_timezone_old(tz)

        time.tzset()

        # get the right values
        now = datetime.datetime.now(tz)
        set_date = tz.localize(datetime.datetime(now.year, now.month, now.day, now.hour, now.minute, now.second))

        # Calculate the number of seconds between this time and timestamp 0
        epoch = utc.localize(datetime.datetime.utcfromtimestamp(0)).astimezone(tz)
        timestamp = (set_date - epoch).total_seconds()
        print(f"Old timestamp: {timestamp}")

    print("Testing UTC:")
    set_system_date_time_new()
    set_system_date_time_old()
    print("Testing US/Eastern:")
    set_system_date_time_new("US/Eastern")
    set_system_date_time_old("US/Eastern")
    print("Testing Asia/Kolkata:")
    set_system_date_time_new("Asia/Kolkata")
    set_system_date_time_old("Asia/Kolkata")

hroncok:
Thanks for the analysis and proposed fix @AdamWill!

PR #3169, opened 2021-02-12
Switch branches for branched f34

VladimirSlavik:
Goal is to make it possible to switch stuff in branch-config.mk and have the rest follow automatically.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -165,7 +165,13 @@ release-and-tag:
 	$(MAKE) tag
 
 anaconda-ci-build:
-	$(CONTAINER_ENGINE) build $(CONTAINER_BUILD_ARGS) -t $(CI_NAME):$(CI_TAG) $(CI_DOCKERFILE)
+	$(CONTAINER_ENGINE) build \
+	$(CONTAINER_BUILD_ARGS) \
+	--build-arg=git_branch=$(GIT_BRANCH) \
+	--build-arg=image=$(BASE_CONTAINER) \
+	--build-arg=copr_repo=$(COPR_REPO) \
+	-t $(CI_NAME):$(CI_TAG) \
+	$(CI_DOCKERFILE)
 
 # User has to be logged first to be able to push the image.
 # See `podman login` for more info.
@@ -173,7 +179,12 @@ anaconda-ci-push:
 	$(CONTAINER_ENGINE) push $(CI_NAME):$(CI_TAG)
 
 anaconda-rpm-build:
-	$(CONTAINER_ENGINE) build $(CONTAINER_BUILD_ARGS) -t $(RPM_NAME):$(CI_TAG) $(RPM_DOCKERFILE)
+	$(CONTAINER_ENGINE) build \
+	$(CONTAINER_BUILD_ARGS) \
+	--build-arg=git_branch=$(GIT_BRANCH) \
+	--build-arg=image=$(BASE_CONTAINER) \
+	-t $(RPM_NAME):$(CI_TAG) \
+	$(RPM_DOCKERFILE)
 
 bumpver: po-pull
 	@opts="-n $(PACKAGE_NAME) -v $(PACKAGE_VERSION) -r $(PACKAGE_RELEASE) -b $(PACKAGE_BUGREPORT)" ; \
--- a/branch-config.mk
+++ b/branch-config.mk
@@ -19,8 +19,16 @@
 # conflicts on multiple places.
 
 
-# Name of the development branch. This could be master, fXX-release, rhelX-branch, rhel-X ...
-GIT_BRANCH ?= master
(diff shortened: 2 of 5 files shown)


VladimirSlavik:
After a conversation with @jkonecny12, this is nowhere near ready.

VladimirSlavik:
OK, this now works, so let's see how we want to shape this...

VladimirSlavik:
Works, no clash between containers and branch check, tons of things in branch-config.

branch-config.mk
```diff
@@ -19,8 +19,20 @@
 # conflicts on multiple places.
 
 
-# Name of the development branch. This could be master, fXX-release, rhelX-branch, rhel-X ...
-GIT_BRANCH ?= master
+# Name of the release branch. This could be master, fXX-release, rhelX-branch, rhel-X ...
+# For branched Fedora, this must be the -release branch.
+GIT_BRANCH ?= f34-release
```
jkonecny12, line 24: I would rename this to something like ``RELEASE_GIT_BRANCH``

branch-config.mk
```diff
@@ -19,8 +19,20 @@
...
 # Directory for this anaconda branch in anaconda-l10n repository. This could be master, fXX, rhel-8 etc.
-L10N_DIR ?= master
+L10N_DIR ?= f34
+
+# Base container for our containers.
+BASE_CONTAINER ?= registry.fedoraproject.org/fedora:34
+
+# Branch with spec file for dependency installation in container. This could be master, fXX-devel, rhel-X etc.
+# For branched Fedora, this must be the -devel branch
+CONTAINER_GIT_BRANCH ?= f34-devel
```
jkonecny12, line 34: This is a bit tricky. It should be `f34-release` for release branch to be correct because spec file is thing what can change specifically for release :(.

Makefile.am
```diff
@@ -88,7 +88,7 @@ CI_DOCKERFILE ?= $(srcdir)/dockerfile/anaconda-ci
 RPM_DOCKERFILE ?= $(srcdir)/dockerfile/anaconda-rpm
 CI_NAME ?= $(CONTAINER_REGISTRY)/rhinstaller/anaconda-ci
 RPM_NAME ?= $(CONTAINER_REGISTRY)/rhinstaller/anaconda-rpm
-CI_TAG := $(or $(CI_TAG),$(GIT_BRANCH))
+CI_TAG := $(or $(CI_TAG),$(shell echo "$(GIT_BRANCH)" | sed -e 's/^\(f[0123456789]\{2\}\).*/\1/'))
```
jkonecny12, line 91: I would like to have the `fXX` in a separate variable. It could be beneficial for future changes in the Makefile.

jkonecny12 (changes requested):
This looks great. I really like these improvements!

branch-config.mk
VladimirSlavik: Extended the comment according to discussion.

Makefile.am
VladimirSlavik: Well, yes, except that is not going to be just f34. It's also master. But it makes sense to have this separately.

VladimirSlavik:
Updated for the @jkonecny12 review.

branch-config.mk
```diff
@@ -19,8 +19,17 @@
 # conflicts on multiple places.
 
 
-# Name of the development branch. This could be master, fXX-release, rhelX-branch, rhel-X ...
-GIT_BRANCH ?= master
+# Name of the expected current git branch.
+# (Expected because it might be in fact a different one by mistake.)
+# This could be master, fXX-devel, fXX-release, rhelX-branch, rhel-X ...
+GIT_BRANCH ?= f34-release
```
jkonecny12, line 25: This should be `f34-devel` now.

branch-config.mk
```diff
@@ -19,8 +19,17 @@
 # conflicts on multiple places.
 
 
-# Name of the development branch. This could be master, fXX-release, rhelX-branch, rhel-X ...
-GIT_BRANCH ?= master
+# Name of the expected current git branch.
+# (Expected because it might be in fact a different one by mistake.)
```
jkonecny12, line 23: What exactly do you mean by that?

jkonecny12 (changes requested):
Please fix the GIT_BRANCH value, otherwise it's correct! Thanks for this!

branch-config.mk
VladimirSlavik: D:

branch-config.mk
VladimirSlavik: Oh. I guess it is redundant. Basically I wrote "expected" and then thought, expected by who, where, what?

jkonecny12 (approved):

PR #3170, opened 2021-02-12
packit: generate config.h after cloning the repo

TomasTomecek:
since create-archive and post-upstream-clone are running in different
environments, paths in Makefiles written from autogen.sh and configure
are no longer valid in the create-archive action

This should resolve: https://prod.packit.dev/srpm-build/13919/logs

Full context: https://blog.tomecek.net/post/automake-in-openshift/ :)

### DIFF
--- a/.packit.yml
+++ b/.packit.yml
@@ -5,6 +5,7 @@ actions:
   post-upstream-clone:
     - ./autogen.sh
     - ./configure
+    - "make config.h"
   create-archive:
     - "make release"
     - 'bash -c "ls -1 anaconda-*.tar.bz2"'
--- a/configure.ac
+++ b/configure.ac
@@ -1,6 +1,6 @@
 # configure.ac for anaconda
 #
-# Copyright (C) 2009  Red Hat, Inc.
+# Copyright (C) 2021  Red Hat, Inc.
 #
 # This program is free software; you can redistribute it and/or modify
 # it under the terms of the GNU Lesser General Public License as published


TomasTomecek:
The ELN build indeed failed but the one for rawhide never started:

https://copr.fedorainfracloud.org/coprs/packit/rhinstaller-anaconda-3170/builds/

jkonecny12:
Wow, thanks a lot for finding this! It's totally weird... I still wonder why the runs are isolated? Seems to me like something which shouldn't be happening when it's in the same action.

Also is this PR ready for merge?

TomasTomecek:
> Wow, thanks a lot for finding this! It's totally weird... I still wonder why the runs are isolated? Seems to me like something which shouldn't be happening when it's in the same action.

It's a deficiency which we are going to fix on our side: https://issues.redhat.com/browse/PACKIT-1133

> Also is this PR ready for merge?

yup, ready to be merged, was hoping to get a green eln build but didn't happen yet

jkonecny12:
OK, thanks!

Could you please rebase this solution to `f34-devel` branch and switch target of this PR. That will get propagated to `master` then.

Also wouldn't it be better to just do everything on one line to avoid this problem?

TomasTomecek:
> OK, thanks!
> 
> Could you please rebase this solution to `f34-devel` branch and switch target of this PR. That will get propagated to `master` then.

Can do!

> Also wouldn't it be better to just do everything on one line to avoid this problem?

I hope that in a week or two, we'll change the sandbox on our side so it's consistent for all action runs. Yes, we can drop the post-upstream-clone and move everything to create-archive - that's a solution as well.

jkonecny12:
That sounds great! In that case would it be possible to add FIXME comment with link to the issue you posted here? Thanks!

TomasTomecek:
closed in favour of https://github.com/rhinstaller/anaconda/pull/3181

PR #3172, opened 2021-02-15
Drop mock-related variables from Makefile

VladimirSlavik:
We don't use mock anymore. These are not used anywhere.


poncovka (approved):
Thanks!

rvykydal (approved):
Looks good to me.

VladimirSlavik:
Kickstart tests not functional yet after branching, skipping.

PR #3173, opened 2021-02-16
Don't block the start of the Network module by the hostname service

poncovka:
When the Network module is started, it shouldn't wait for the hostname service.
It should connect to the hostname service from the loop after it is initialized
and published. Otherwise, the activation of services might time out.

Related: rhbz#1919380


rvykydal (approved):
Looks good to me, thank you.

jkonecny12 (approved):
LGTM!

VladimirSlavik (approved):
Looks good to me. Thank you!

PR #3174, opened 2021-02-16
Drop astroid hotfix patch

jkonecny12:
New version of astroid (2.5) was released so the build will fail because the patch does nothing. Let's remove it to fix the issue which, as a nice benefit, will improve a build speed of our container!


VladimirSlavik (approved):
Looks good to me. Thank you!

M4rtinK (approved):
LGTM, nice cleanup! :)

poncovka (approved):

PR #3175, opened 2021-02-16
Drop astroid hotfix patch

jkonecny12:
New version of astroid (2.5) was released so the build will fail because the patch does nothing. Let's remove it to fix the issue which, as a nice benefit, will improve a build speed of our container!

This was already merged to master but it should have go to f34 too. Sorry for this noise :(.


jkonecny12:
Backport of https://github.com/rhinstaller/anaconda/pull/3174/commits to f34-devel.

VladimirSlavik (approved):

poncovka (approved):

PR #3181, opened 2021-02-17
multiple fixes for packit

TomasTomecek:
commits:
```
packit: make tests ⊂ builds for the chroot set
```

```
packit: run all actions in a single action

to make sure that generated files from autotools have correct paths

since create-archive and post-upstream-clone are running in different
environments, paths in Makefiles written from autogen.sh and configure
are no longer valid in the create-archive action

this mainly affected the config.h in widgets
```

```
configure.ac: make the Copyright up to date

I hope this will trigger regeneration of config.h in CI
```

### DIFF
--- a/.packit.yml
+++ b/.packit.yml
@@ -2,10 +2,9 @@ specfile_path: anaconda.spec
 upstream_package_name: anaconda
 upstream_tag_template: anaconda-{version}-1
 actions:
-  post-upstream-clone:
+  create-archive:
     - ./autogen.sh
     - ./configure
-  create-archive:
     - "make release"
     - 'bash -c "ls -1 anaconda-*.tar.bz2"'
 jobs:
--- a/.packit.yml
+++ b/.packit.yml
@@ -12,12 +12,15 @@ jobs:
     trigger: pull_request
     metadata:
       targets:
-        - fedora-development
+        - fedora-34
+        - fedora-rawhide
 
   - job: copr_build
     trigger: pull_request
     metadata:
       targets:
+        - fedora-34
+        - fedora-rawhide
         - fedora-eln
 
   - job: copr_build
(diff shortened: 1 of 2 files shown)


.packit.yml
```diff
@@ -12,12 +12,15 @@ jobs:
     trigger: pull_request
     metadata:
       targets:
-        - fedora-development
+        - fedora-34
+        - fedora-rawhide
```
jkonecny12, line 16: Why not to use `fedora-development` on both places?

jkonecny12 (commented):
Looks great to me otherwise. Thanks for these changes!

.packit.yml
TomasTomecek: I wanted it to be explicit since this is for the f34-devel branch, I can change both to -development

we actually had a brief discussion in our team if eln should be part of -development alias

please let me know what you prefer :)

jkonecny12: I don't think we want to have ELN in `-development` because it's not supported by TFT so it can't be used in tests.

And when I'm thinking about it lets be explicit here. I have to remove the `34` line on master so or so. So it will be easier to remove one line instead of changing it :).

jkonecny12 (approved):
Looks great to me. Thanks!

VladimirSlavik (approved):
Looks good to me. Thank you!

.packit.yml
TomasTomecek: agreed, that would be my argument that ELN should not be in the -development alias; also many projects wouldn't even build in there.

Please let me know if I should change something in here or it's good as it is.

TomasTomecek:
oh yeah, 34 is not available in testing farm yet, that's why it's not being updated

jkonecny12:
Ahh, thanks for info!

PR #3182, opened 2021-02-17
Don't translate prompt keys (#1892830)

poncovka:
Translated prompt keys are not supported by the simpleline library, so
anaconda shouldn't translate them neither. Use constants when possible.

Resolves: rhbz#1892830


jkonecny12 (approved):
Looks good to me. Thanks for solving this!

VladimirSlavik (approved):
Thank you! I think it's ok, but don't trust me here too much.

poncovka:
Rebased.

PR #3184, opened 2021-02-17
Make the user interface context safe for the initial setup

poncovka:
The context needs to be safe to use for the initial setup, so provide
a very limited specific API that will work in both projects.

### DIFF
--- a/pyanaconda/ui/context.py
+++ b/pyanaconda/ui/context.py
@@ -18,8 +18,6 @@
 
 __all__ = ["context"]
 
-from pyanaconda.anaconda import Anaconda
-
 
 class UserInterfaceContext(object):
     """The context of the user interface.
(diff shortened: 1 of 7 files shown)


jkonecny12 (approved):
Looks good to me. Sorry for still not having KS tests for initial-setup working. :(

pyanaconda/ui/context.py
```diff
@@ -28,27 +26,26 @@ class UserInterfaceContext(object):
     of the user interface. The goal is to replace all
     these objects with this one.
 
+    WARNING: The context has to be usable by the initial setup!
```
M4rtinK, line 29: I think we can formulate this better:
```

WARNING: The UserInterfaceContext class is also used by the Initial Setup tool. 
         Please keep that in mind when doing any API breaking changes.
```

M4rtinK (approved):
Yeah, this should really make things much more robust & also looks cleaner. :)

pyanaconda/ui/context.py
poncovka: Fixed.

poncovka:
Updated.

PR #3185, opened 2021-02-17
Add support for f34 and eln branches to container refresh workflow

jkonecny12:
Change the structure so each build will be started as separate matrix with it's own checkout. This is the simplest solution and we can use defaults of the branch thanks to that.

This is a first draft for pre-review and it needs to be tested first but in theory it should work.


VladimirSlavik (approved):
Looks good to me. Thank you!

PR #3186, opened 2021-02-18
Do not try to push latest from ELN tag

jkonecny12:
This will fail because there is no master tag created and also it's not nice to push latest from master and eln builds.

Solution is not to check branch but tag we want to push.


VladimirSlavik (approved):
Looks good to me. Thank you!

VladimirSlavik:
On a second thought, is there any reason to NOT put this into f34 (too)?

PR #3187, opened 2021-02-18
Add the GetKernelVersionList method

poncovka:
Call the `GetKernelVersionList` method of a payload module to get a list of kernel
versions. The method might raise the `UnavailableValueError` exception if it is
called before the payload installation.

**Blocked by:** Beta Freeze


jkonecny12 (approved):
From the code PoV seems good to me. From the functionality it has to be tested to tell.
Thanks for this change.

M4rtinK (approved):
Looks good to me code-wise as well. :)

PR #3188, opened 2021-02-18
Add support for anaconda-rpm containers to refresh workflow

jkonecny12:
We will use power of combinations of matrix. That will create all the combinations between container-tag and container-type variables for us and we don't have to define new job for this.

### DIFF
--- a/.github/workflows/container-autoupdate.yml
+++ b/.github/workflows/container-autoupdate.yml
@@ -6,8 +6,8 @@ on:
   workflow_dispatch:
 
 jobs:
-  refresh-container:
-    name: Refresh anaconda-ci container
+  refresh-containers:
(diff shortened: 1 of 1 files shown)


VladimirSlavik (approved):
Looks good to me. Thank you!

jkonecny12:
It's failing, looking for solution right now.

jkonecny12:
**UPDATED:**
* Fixed issues found from the first version.

jkonecny12:
Here is the test run:
https://github.com/Test-anaconda-org/anaconda/runs/1927872126?check_suite_focus=true

Everything failed but that is because of missing secrets and missing blivet on ELN.

VladimirSlavik (approved):
Looks good to me. Thank you!

PR #3189, opened 2021-02-18
Enable our and blivet COPR repositories for rpm tests

jkonecny12:
This will solve the situation when rpm tests are failing because we depend on blivet which is not yet released.


M4rtinK (approved):
Looks good to me, thanks! :)

VladimirSlavik (approved):
Looks good to me. Thank you!

PR #3191, opened 2021-02-19
Add container push for rpm containers

jkonecny12:
They are now supported by container refresh so let's use that as our benefit.


poncovka (approved):
Looks good to me.

VladimirSlavik (approved):
Looks good to me. Thank you!

PR #3192, opened 2021-02-19
Remove interactive parameter from container-rpm-test

jkonecny12:
These tests are not interactive and docker does not like them so get rid of them to solve this.


VladimirSlavik (approved):
Looks good to me. Thank you!

PR #3193, opened 2021-02-19
Use docker as container engine for time required

jkonecny12:
We encountered issue that ELN images can't be used in podman. Until that is fixed fall back to docker which doesn't have this problem

https://github.com/containers/podman/issues/9442


VladimirSlavik (approved):
Looks good to me. Thank you!

PR #3194, opened 2021-02-19
Use older Ubuntu for container auto-update workflow

jkonecny12:
We need that because there is a bug which will make podman run on ELN image fail.

Older version seems to have older crun which doesn't have this issue.

See: https://github.com/containers/podman/issues/9442

### DIFF
--- a/.github/workflows/container-autoupdate.yml
+++ b/.github/workflows/container-autoupdate.yml
@@ -8,7 +8,8 @@ on:
 jobs:
   refresh-containers:
     name: Refresh anaconda containers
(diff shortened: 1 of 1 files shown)


jkonecny12:
My test fo these changes https://github.com/Test-anaconda-org/anaconda/runs/1937138521?check_suite_focus=true. Unfortunately, I really don't want to push the containers, so everything after login can't be tested there :(.

VladimirSlavik (approved):
Looks good to me. Thank you!

PR #3195, opened 2021-02-19
Fix RHEL zram conditional in spec file

M4rtinK:
Turns out a different spec file incantation needs to be used
for this conditional to work.

Thanks a lot to Chris Murphy for identifying the issue and to
Adam Williamson for suggesting how to fix it! :)


VladimirSlavik (approved):
Looks good to me. Thank you!

poncovka (approved):
Looks good to me.

PR #3196, opened 2021-02-22
Determine GRUB directory relative path to use in config file

martinezjavier:
Currently Anaconda is only checking whether /boot/ is a mount point or not
to use the proper relative path to the GRUB directory. But there are more
cases than this, for example /boot/grub2 could be a symlink or a subvolume
in a btrfs filesystem.

Use the grub2-mkrelpath tool to figure out the actual relative path that
has to be used in the GRUB config file.

Resolves: rhbz#1928588


jkonecny12:
Closing this because it will be solved by https://github.com/rhinstaller/anaconda/pull/3197/files.

PR #3198, opened 2021-02-23
Clarify how branches are merged back for contributors to pick a target

martinezjavier:
I usually get confused about which branch should be the target of a pull
request. Add some clarifications about how the branches are merged back,
for contributors to decide the branch they should target a pull-request.

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -47,12 +47,25 @@ Which is my target git branch?
 Depending on where you want to make your contribution please choose your correct branch based on the table below.
 
 +--------------------------+--------------+
-| Fedora Rawhide           | master       |
+| Fedora Rawhide only      | master       |
 +--------------------------+--------------+
-| Fedora XX                | fXX-devel    |
+| Fedora XX and Rawhide    | fXX-devel    |
++--------------------------+--------------+
+| Fedora XX only           | fXX-release  |
 +--------------------------+--------------+
 | RHEL-X / CentOS Stream X | rhel-x       |
 +--------------------------+--------------+
+
+Notice that the ``master`` branch should only be a target of a pull request if the change does not apply to any
+of the branched Fedora versions. If the change applies to both Fedora Rawhide and Fedora XX, then a ``fXX-devel``
+branch has to be targeted. All changes to ``fXX-devel`` branches are merged back into ``master``, so there's no
+need to open a separate pull-request for ``master`` in that case.
+
+If a change only applies to a Fedora XX release and the changes don't have to be merged back into ``master``,
+then the target branch should be a ``fXX-release`` relese branch.
+
+The ``rhel-x`` branches on the other hand are not merged back into ``master``, and so a separate pull request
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.rst
```diff
@@ -54,6 +54,14 @@ Depending on where you want to make your contribution please choose your correct
 | RHEL-X / CentOS Stream X | rhel-x       |
 +--------------------------+--------------+
 
+Notice that the ``master`` branch should only be a target of a pull request if the change does not apply to any
+of the branched Fedora versions. If the change applies to both Fedora Rawhide and Fedora XX, then a ``fXX-devel``
+branch has to be targeted. All changes to ``fXX-devel`` branches are merged back into ``master``, so there's no
+need to open a separate pull-request for ``master`` in that case.
+
+The ``rhel-x`` branches on the other hand are not merged back into ``master``, and so a separate pull request
+has to be opened targeting either for a ``fXX-devel`` branch or ``master``.
```
VladimirSlavik, line 63: I think you need to choose one of "targeting either" and "for", but not both at once. (So it's actually xor...)

VladimirSlavik, line 63: I'm not sure if that is within the scope here, but we use the labels "port to X" where X = Fedora or RHEL. Having that label on a PR means this should be in the other branch too, but the PR was not yet created.

Arguably, that is of little use to contributors because they can not add these labels.

VladimirSlavik (approved):
Looks good to me. Thank you!

Somewhat ironically, I guess branching doesn't matter for this very PR because people will usually see the docs from master.

jkonecny12 (changes requested):
That's great improvement, thanks!

Please just change also the table:

```
| Fedora XX                | fXX-devel    |
```
to this:

```
| Fedora XX and Rawhide                | fXX-devel    |
+---------------------------------------------+-----------------+
| Fedora XX only                              | fXX-release |
```

poncovka (approved):
It looks good to me. Thank you!

CONTRIBUTING.rst
martinezjavier: > Arguably, that is of little use to contributors because they can not add these labels.

Yeah, I always thought that this kind of labeling for PR was something handled by the Anaconda core developers as a part of the PR review process.

martinezjavier: Ok

martinezjavier:
> Please just change also the table:

Ok, will do.

jkonecny12:
@martinezjavier hi, should I re-take this PR and finish that or do you want to do that?

martinezjavier:
@jkonecny12 @VladimirSlavik @poncovka sorry for the delay, this fell of my radar.

I've addressed the issues you pointed out and pushed a new version of the commit.

VladimirSlavik (approved):
Thanks, even better.

jkonecny12 (approved):
Looks great to me now! Thanks!

poncovka (approved):
It looks great. Thank you!

PR #3199, opened 2021-02-23
The network spoke should be visible in live spins (#1932961)

poncovka:
It should be always possible to configure the hostname for the target system
in the graphical network spoke, unless the spoke was specifically disabled in
the Anaconda configuration files.

This regression was caused by the commit 6b80b7b.

Resolves: rhbz#1932961


rvykydal (approved):
Looks good to me.

jkonecny12 (approved):
LGTM!

M4rtinK (approved):

poncovka:
Rebased and added a bug number.

PR #3200, opened 2021-02-23
Pylint updates

VladimirSlavik:
New Pylint chokes on different stuff than the previous one.

The good part is that there are no new real problems, this is just whack-a-mole with false positives.

### DIFF
--- a/tests/nosetests/pyanaconda_tests/modules/boss/modules_test.py
+++ b/tests/nosetests/pyanaconda_tests/modules/boss/modules_test.py
@@ -64,7 +64,7 @@ class ModuleManagerTestCase(unittest.TestCase):
         ]
 
         task = StartModulesTask(self._message_bus, service_names, addons_enabled=False)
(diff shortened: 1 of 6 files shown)


tests/nosetests/pyanaconda_tests/modules/boss/modules_test.py
```diff
@@ -64,7 +64,7 @@ def start_one_module_test(self, gio):
         ]
 
         task = StartModulesTask(self._message_bus, service_names, addons_enabled=False)
-        (observer, ) = self._check_started_modules(task, service_names)
+        (observer, ) = self._check_started_modules(task, service_names)  # pylint: disable=unbalanced-tuple-unpacking
```
poncovka, line 67: This is weird. I think it should be `observers = ` based on other usages of the method.

poncovka (approved):

jkonecny12 (approved):
LGTM! Thanks for fixing this!

PR #3202, opened 2021-02-25
Conscious language blacklist 1

rvykydal:
Note: the main reason for picking blocklist/passlist (not denylist/allowlist) was consistency with RHEL 8 installation documentation where it is already used.

### DIFF
--- a/pyanaconda/modules/payloads/base/utils.py
+++ b/pyanaconda/modules/payloads/base/utils.py
@@ -35,19 +35,19 @@ def create_root_dir(sysroot):
     mkdirChain(os.path.join(sysroot, "root"))
 
 
-def write_module_blacklist(sysroot):
-    """Create module blacklist based on the user preference.
+def write_module_denylist(sysroot):
+    """Create module denylist based on the user preference.
 
     Copy modules from modprobe.blacklist=<module> on cmdline to
     /etc/modprobe.d/anaconda-blacklist.conf so that modules will
-    continue to be blacklisted when the system boots.
+    continue to be added to a denylist when the system boots.
     """
     if "modprobe.blacklist" not in kernel_arguments:
         return
 
     mkdirChain(os.path.join(sysroot, "etc/modprobe.d"))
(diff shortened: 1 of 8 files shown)


pyanaconda/modules/payloads/base/utils.py
```diff
@@ -35,19 +35,19 @@ def create_root_dir(sysroot):
...
     Copy modules from modprobe.blacklist=<module> on cmdline to
     /etc/modprobe.d/anaconda-blacklist.conf so that modules will
-    continue to be blacklisted when the system boots.
+    continue to be blocklisted when the system boots.
     """
     if "modprobe.blacklist" not in kernel_arguments:
         return
 
     mkdirChain(os.path.join(sysroot, "etc/modprobe.d"))
     with open(os.path.join(sysroot, "etc/modprobe.d/anaconda-blacklist.conf"), "w") as f:
```
poncovka, line 49: Don't we want to change the name of the file as well, or is it documented somewhere?

poncovka (approved):
It looks good to me. Thanks!

poncovka:
Do we want to do this in Fedora 34? Isn't it too late?

pyanaconda/modules/payloads/base/utils.py
poncovka: Never mind. I have found https://github.com/rhinstaller/anaconda/pull/3203.

VladimirSlavik (approved):
I think this is mostly fine. The word "blocklist" makes me really sad, but if that's used elsewhere too, it's better to stay consistent.

VladimirSlavik (changes requested):
On a further probing, I hear kernel dev wants to use "denylist". I think we should stay consistent with that.

I think it's actually an issue of RHEL docs choosing something else for installer.

rvykydal:
@VladimirSlavik thank you for digging into it, I pushed new version using 'denylist'

poncovka (approved):

poncovka (approved):

VladimirSlavik (approved):

PR #3207, opened 2021-03-01
network: create per-thread NMClient instance (#1931389)

rvykydal:


rvykydal:
WIP for comments from Thomas Haller.

pyanaconda/modules/network/network.py
```diff
@@ -68,17 +69,12 @@ def __init__(self):
...
-        if SystemBus.check_connection():
-            nm_client = NM.Client.new(None)
-            if nm_client.get_nm_running():
-                self.nm_client = nm_client
-                self.nm_client.connect("notify::%s" % NM.CLIENT_STATE, self._nm_state_changed)
-                initial_state = self.nm_client.get_state()
-                self.set_connected(self._nm_state_connected(initial_state))
-            else:
-                log.debug("NetworkManager is not running.")
+        self.nm_client = get_new_nm_client()
```
poncovka, line 73: Do you think this set up could be run later? Something like `_connect_to_hostname_service`.

poncovka (commented):

jkonecny12 (approved):
Looks good to me!

VladimirSlavik (approved):
Looks good to me.

jkonecny12:
I would appreciate some description in the comment of the commit message. Why we need that.

t-feng:
hello, everybody
I got same problem.
this issue has any news? :)

t-feng:
Create new NMClient instance for worker threads can fix this problem?
Applying this patch, we still got  _g_main_context_push_thread_default: assertion 'acquired_context' failed_ error. Because context is acquired in main thread.
Whether we should create new context for each worker threads？

rvykydal:
I tried to create new context but it lead to hangs, see https://bugzilla.redhat.com/show_bug.cgi?id=1931389#c10

t-feng:
@rvykydal hello
this issue has beed fixed by some certain PR?  Or fixed by high version? (anaconda-33.19 still has this problem with a very low probability)

PR #3209, opened 2021-03-01
Create a task for tearing down OSTree mount targets

poncovka:
Move the code for tearing down OSTree mount targets into a new task.


VladimirSlavik (approved):
Looks good to me. Thank you!

rvykydal (approved):
Looks good to me.

PR #3210, opened 2021-03-01
F34 devel fix owner tests

VladimirSlavik:
This makes tests work. It seems that the best course of action is to take this and comment out stuff on other branches.

Test run here: https://github.com/vslavik-test-org/anaconda/pull/3/checks?check_run_id=2005869646

In an ideal world, we could abstract away the matrix from both jobs, or construct it on demand. But I don't think we can. Oh, the wonders of Github actions black boxes, bugs and workarounds!

Potential TODOs within this PR:
- Drop push runs.
- Make names nicer.

### DIFF
--- a/.github/workflows/tests-owners.yml
+++ b/.github/workflows/tests-owners.yml
@@ -34,16 +34,33 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        # empty = release that corresponds to current branch name
-        release: ['', 'eln']
+        # * Don't forget to copypaste also the same matrix below in rpm tests!
+        # * Replace current branch in "release" with empty string. This makes name of the check for
+        #   the current branch always have the same name, so that it can be added to required
+        #   checks.
+        # * For various branches, comment out what does not apply:
+        #     on -devel, run everything
+        #     on -release, only -release
+        #     on master, only master and ELN
+        release: [master, eln, '', f34-release]
         include:
+          - release: master
+            target_branch: 'master'
+            ci_tag: 'master'
           - release: eln
-            name-postfix: ' (eln)'
+            target_branch: 'master'
+            ci_tag: 'eln'
             build-args: '--build-arg=image=quay.io/fedoraci/fedora:eln-x86_64'
+          - release: ''
+            target_branch: 'f34-devel'
+            ci_tag: 'f34-devel'
+          - release: f34-release
+            target_branch: 'f34-release'
+            ci_tag: 'f34-devel'
     env:
-      CI_TAG: '${{ matrix.release }}'
+      CI_TAG: '${{ matrix.ci_tag }}'
       CONTAINER_BUILD_ARGS: '${{ matrix.build-args }}'
-      TARGET_BRANCH_NAME: origin/master
+      TARGET_BRANCH_NAME: 'origin/${{ matrix.target_branch }}'
 
     steps:
       - name: Clone repository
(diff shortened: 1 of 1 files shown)


.github/workflows/tests-owners.yml
```diff
@@ -94,16 +108,24 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        # empty = release that corresponds to current branch name
-        release: ['', 'eln']
```
jkonecny12, line 98: That `''` was there because our current required tests setup is expecting test with name `unit-tests`. However, this way we will have unit-tests (master) which is fine but will not work. 

I guess we should drop the current required tests configuration on the repository.

.github/workflows/tests-owners.yml
```diff
@@ -34,16 +34,30 @@ jobs:
...
+        # Matrix must be built purely from includes due to bugs (as of 2021-02).
+        # Don't forget to copypaste also the same matrix below in rpm tests!
+        # For various branches, comment out what does not apply:
+        #   on -devel, run everything
+        #   on -release, only -release
+        #   on master, only master and ELN
```
jkonecny12, line 42: This is incorrect.

The correct one should be:

```
on -devel, run master, ELN and -devel
on -release, run everything
on master, only master and ELN
```
That also explains why I had there information of on what branch it should be started. If I read it correctly this will now work only for the one branch + `ELN`.

This should be fixed but let's that solve in separate PR.

jkonecny12 (approved):
Thanks a lot for fixing those. We should improve the situation but let's go with this solution and improve that in a future PR.

.github/workflows/tests-owners.yml
VladimirSlavik: Changes flow like this:
```
devel -> release
      -> master -> ELN
```
A pull request to release makes the code go nowhere else, so why would I test that on master? On the other hand, anything on devel will get everywhere, so testing it everywhere makes sense.

This PR is for devel, so the full set lands here, and we can just remove things on the other branches after merging.

.github/workflows/tests-owners.yml
VladimirSlavik: Then we need to adapt the required tests, once the actions are in place. I think it's pretty easy to set up, with rules per branch.

rvykydal:
I didn't find any reason to disapprove the PR but I don't understand the area enough to approve meaningfully. I'll try to follow and learn.

.github/workflows/tests-owners.yml
VladimirSlavik: After discussion, let's keep that mechanism.

VladimirSlavik:
Update: Simpler names for the jobs, no suffix for current-branch run to enable "required checks".

Test run: https://github.com/vslavik-test-org/anaconda/pull/3/checks?check_run_id=2015541109

.github/workflows/tests-owners.yml
jkonecny12: My fault here. I totally forgot that this is `pull_reques_target` trigger.

.github/workflows/tests-owners.yml
jkonecny12: To add more insights. It's not that simple to set specific test requirement for branches in GitHub. It's hard to match that for all future tests when the name of the test is changing, and creating new rules for each branching seems like a bad idea.

PR #3203, opened 2021-02-25
conscious lang: rename /etc/modprobe.d/anaconda-blacklist.conf

rvykydal:
Follows on https://github.com/rhinstaller/anaconda/pull/3202 but this time changing the name of a configuration file created by Anaconda on target system so proposed in a separate PR.

PR #3211, opened 2021-03-01
Extend the DBus API of the Payloads service

poncovka:
Move some methods from the payload modules to the main service.
Drop the support for the pre-installation tasks.

Some of the installation tasks don't depend on the type of the payload
module, so include them by default on the service-level.

### DIFF
--- a/pyanaconda/modules/payloads/base/initialization.py
+++ b/pyanaconda/modules/payloads/base/initialization.py
@@ -15,95 +15,15 @@
 # License and may only be used or replicated with the express permission of
 # Red Hat, Inc.
 #
-import os
-import shutil
 import traceback
-from glob import glob
-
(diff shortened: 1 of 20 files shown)


rvykydal (approved):
Looks good to me.
I've moved from blocklist to denylist in:
https://github.com/rhinstaller/anaconda/pull/3202
https://github.com/rhinstaller/anaconda/pull/3203
(it seems kernel is going to use denylist).
So maybe adapt this PR to be consistent with the above PRs (use denylist) or use the blacklist and I'll update my (conflicting) PRs after this one is merged ?
Or I can open new PR for blocklist -> denylist but that may get a bit confusing ...

VladimirSlavik (approved):
Looks good to me. Thank you!

poncovka:
Rebased and tested.

PR #3215, opened 2021-03-03
Build new crun version for ELN container refresh

jkonecny12:
This is a workaround of issue https://github.com/containers/podman/issues/9442 which broke podman on Ubuntu hosts with crun < 0.18. Build crun 0.18 from source for ELN container refresh to solve this issue.

Tested [here](https://github.com/Test-anaconda-org/anaconda/runs/2021966024?check_suite_focus=true).

### DIFF
--- a/.github/workflows/container-autoupdate.yml
+++ b/.github/workflows/container-autoupdate.yml
@@ -8,8 +8,7 @@ on:
 jobs:
   refresh-containers:
     name: Refresh anaconda containers
-    # FIXME: Switch to newer 20.04 when crun 0.18 will be available see issue https://github.com/containers/podman/issues/9442 is resolved
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-latest
     # we need to have matrix to cover all branches because schedule is run only on default branch
     # we can add here fedora branches (e.g. f33-devel) to cover their support when needed
     strategy:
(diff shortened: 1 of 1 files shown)


.github/workflows/container-autoupdate.yml
```diff
@@ -36,6 +35,29 @@ jobs:
         with:
           ref: ${{ matrix.branch }}
 
+      # FIXME: Remove when crun 0.18 will be available by default, see issue https://github.com/containers/podman/issues/9442
+      - name: Install newest crun from source (workaround for ELN)
+        if: matrix.container-tag == 'eln'
+        run: |
+          set -eux
+          echo "Currently installed crun version:"
+          crun --version
```
VladimirSlavik, line 44: Written this way, we will always use 0.18. Please make it also fail if the pre-installed version is no longer 0.17:

```
CRUN_VERSION=$(crun --version | grep "^crun version 0.17$")
echo "$CRUN_VERSION"
[ -z "$CRUN_VERSION" ] && exit 17
```

You can insert ` | sed -e "s/0.17/0.18/"` before grep to test the failure mode.

VladimirSlavik (changes requested):

.github/workflows/container-autoupdate.yml
VladimirSlavik: Alternatively, dismiss this and merge away, I can do this part as a follow-up PR.

VladimirSlavik (approved):
Alternatively, dismiss this and merge away, I can do this part as a follow-up PR.

jkonecny12:
*Updated:*
* Add check created by @VladimirSlavik , thanks!
* Just a small improvement in the check because the version is not 0.17 but 0.17.xxxx on the machine.

VladimirSlavik (approved):
Thank you!

PR #3218, opened 2021-03-03
Hide members of the DNF payload

poncovka:
Rename methods and properties that are not used outside of the DNF payload.


VladimirSlavik (approved):
Looks good to me. Thank you!

rvykydal (approved):
Looks good to me.

PR #3219, opened 2021-03-04
Fix unknown error when entering wrong NFS URL address

century6:
enables anaconda to recognize the error when inputting the wrong NFS URL address in GUI instead of throwing an unknown error

### DIFF
--- a/pyanaconda/ui/gui/spokes/installation_source.py
+++ b/pyanaconda/ui/gui/spokes/installation_source.py
@@ -1192,6 +1192,10 @@ class SourceSpoke(NormalSpoke, GUISpokeInputCheckHandler, SourceSwitchHandler):
                 else:
                     return _("Remote directory is required")
 
+            if ":" not in url_string or len(url_string.split(":")) != 2:
+                return _("NFS server must be specified as \"SERVER:/PATH\". "
+                         "Only one colon is allowed in the url string.")
+
         return InputCheck.CHECK_OK
 
     def _check_url_entry(self, inputcheck):


VladimirSlavik:
Hello @century6, thanks for the contribution! I haven't tried running the code yet, but it looks good to me.

Given that the change is in the branch of code handling NFS, I would prefer mentioning NFS in the commit message too. I had to re-read this thing a few times to understand it. At first sight, I got the impression that this handles URLs and is somehow stacked at the end of ifs because an error will blow past all of the ifs.

century6:
> Hello @century6, thanks for the contribution! I haven't tried running the code yet, but it looks good to me.
> 
> Given that the change is in the branch of code handling NFS, I would prefer mentioning NFS in the commit message too. I had to re-read this thing a few times to understand it. At first sight, I got the impression that this handles URLs and is somehow stacked at the end of the if-elif chain, because an error will blow past all of the ifs.
> 
> PS: Starting contributor tests is still technically broken, but on master the old version should still work. Not sure what outcome to expect.

Sorry, it's my fault. NFS is now mentioned in the commit message.

VladimirSlavik:
I gave this a proper test, and it works, but I'm not sure it's entirely correct.

The case where the field has no colon `:` is caught by the empty `path` check, so that is redundant. What remains is the check for more than one colon, and given the preceding operations and checks, that could be simplified as `":" in path`. Then the error message should mention that only one colon is allowed, since that's the exact problem it's solving. The current message does not actually point out the specific problem found with the input.

jkonecny12 (approved):
Looks great to me, thanks for the contribution!

century6:
> /test

century6:
> I gave this a proper test, and it works, but I'm not sure it's entirely correct.
> 
> The case where the field has no colon `:` is caught by the empty `path` check, so that is redundant. What remains is the check for more than one colon, and given the preceding operations and checks, that could be simplified as `":" in path`. Then the error message should mention that only one colon is allowed, since that's the exact problem it's solving. The current message does not actually point out the specific problem found with the input.

The specific problem found with the input have been added to the message.

VladimirSlavik (changes requested):
Please squash the commits, we won't merge your WIP changing back and forth. Alternatively, I can do that for you - just say so!

VladimirSlavik:
kickstart-test --testtype smoke

century6:
> Please squash the commits, we won't merge your WIP changing back and forth. Alternatively, I can do that for you - just say so!

OK, I have squashed the commits.

poncovka (commented):
The code looks good to me. Thank you!

VladimirSlavik (approved):
Thanks!

PR #3220, opened 2021-03-04
Remove container refresh workaround for ELN

jkonecny12:
They updated runners day before I merged this PR. So let's remove this, sight :(.

Unfortunately, not all runners have the new version but I don't think we want to solve that on our side. Just wait until they will update everything.

### DIFF
--- a/.github/workflows/container-autoupdate.yml
+++ b/.github/workflows/container-autoupdate.yml
@@ -35,33 +35,6 @@ jobs:
         with:
           ref: ${{ matrix.branch }}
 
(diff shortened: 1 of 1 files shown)


VladimirSlavik (approved):
Thank you anyway :)

jkonecny12:
Seems that GitHub runners update is a slow process. Right now, it's luck if we get new or old one. Let's leave this workaround a while and merge this later.

jkonecny12:
Last four days went with only the newer VMs. Let's merge this to remove our workaround and speed up the execution.

jkonecny12:
Rebased.

PR #3222, opened 2021-03-04
Fix contributor tests

VladimirSlavik:
Preliminary draft. It works \o/ but needs more work to be proper and tidy and all.

TODOs:

- [x] Generate JSON
- [x] Create matrix from JSON
- [x] Clean up JSON generation
- [x] Maybe reintroduce the status postfix?

Testing here: https://github.com/vslavik-test-org/anaconda/pull/4 - thanks @jkonecny12!

Note that the master and eln tests can't succeed, because this change itself, being on the repo's master, prevents rebasing. I am not sure if that is correct.

### DIFF
--- a/.github/workflows/tests-contributors.yml
+++ b/.github/workflows/tests-contributors.yml
@@ -37,30 +37,62 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
+      - name: Generate matrix for test jobs
+        id: generate_matrix
+        run: |
+          FEDVERS=34 # replace this number for next Fedora
+          TARGET_BRANCH=${{ fromJson(steps.pr_api.outputs.data).base.ref }}
+
+          JSON_MASTER='{"release": "master", "target-branch": "master", "name-postfix": " (master)"}'
+          JSON_ELN='{"release": "eln", "target-branch": "master", "name-postfix": " (eln)", "build-args": "--build-arg=image=quay.io/fedoraci/fedora:eln-x86_64"}'
+          JSON_DEVEL='{"release": "f'$FEDVERS'-devel", "target-branch": "f'$FEDVERS'-devel", "name-postfix": " (f'$FEDVERS'-devel)"}'
+          JSON_RELEASE='{"release": "f'$FEDVERS'-release", "target-branch": "f'$FEDVERS'-release", "name-postfix": " (f'$FEDVERS'-release)"}'
+
+          REF_DEVEL='f'$FEDVERS'-devel'
+          REF_RELEASE='f'$FEDVERS'-release'
+
+          # Always order things so that first array element is the current branch, because...
+          if [[ "$TARGET_BRANCH" == "$REF_DEVEL" ]] ; then
+            JSON_MERGED='['$JSON_DEVEL', '$JSON_RELEASE', '$JSON_MASTER', '$JSON_ELN']' ;
+          elif [[ "$TARGET_BRANCH" == "$REF_RELEASE" ]] ; then
+            JSON_MERGED='['$JSON_RELEASE']' ;
+          elif [[ "$TARGET_BRANCH" == "master" ]] ; then
+            JSON_MERGED='['$JSON_MASTER', '$JSON_ELN']' ;
+          else
+            echo "Branch $TARGET_BRANCH is not configured to run tests, aborting."
+            exit 0
+          fi
+
+          # ...here "name-postfix" in first array element is replaced with empty string, to make
+          # the displayed name always the same. The name with no suffix is set as the required
+          # check, so that it must always exist.
+          OUT_JSON=$(echo "$JSON_MERGED" | jq -c '.[0]."name-postfix" |= ""')
+
+          # Pretty-print the results for debug and set the outputs
+          echo "Generated JSON is:"
+          echo "$OUT_JSON" | jq '.'
+          echo "::set-output name=matrix::$OUT_JSON"
+
     outputs:
       allowed_user: ${{ steps.check_user_perm.outputs.allowed_user }}
       base_ref: ${{ fromJson(steps.pr_api.outputs.data).base.ref }}
       sha: ${{ fromJson(steps.pr_api.outputs.data).head.sha }}
+      matrix: ${{ steps.generate_matrix.outputs.matrix }}
 
-  unit-tests-master:
+  unit-tests-fedora:
     needs: pr-info
-    if: needs.pr-info.outputs.base_ref == 'master' && needs.pr-info.outputs.allowed_user == 'true'
+    if: needs.pr-info.outputs.matrix != '' && needs.pr-info.outputs.allowed_user == 'true'
     runs-on: ubuntu-20.04
     timeout-minutes: 30
     strategy:
       fail-fast: false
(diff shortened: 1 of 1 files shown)


.github/workflows/tests-contributors.yml
```diff
@@ -37,30 +37,77 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
+      - name: Generate matrix for test jobs
+        id: generate_matrix
+        run: |
+          FEDVERS=34
+          TARGET_BRANCH=${{ fromJson(steps.pr_api.outputs.data).base.ref }}
+
+          JSON_MASTER_ELN='{"release": "master", "target-branch": "master", "ci-tag": "master"}, {"release": "eln", "target-branch": "master", "ci-tag": "eln", "build-args": "--build-arg=image=quay.io/fedoraci/fedora:eln-x86_64"}'
```
jkonecny12, line 46: I know they are used together but for sake of readability I would split these two.

.github/workflows/tests-contributors.yml
```diff
@@ -37,30 +37,77 @@ jobs:
...
+          if [[ "$TARGET_BRANCH" == "$REF_DEVEL" ]] ; then
+            MERGED='{"include": ['$JSON_DEVEL', '$JSON_RELEASE', '$JSON_MASTER_ELN']}' ;
+          elif [[ "$TARGET_BRANCH" == "$REF_RELEASE" ]] ; then
+            MERGED='{"include": ['$JSON_RELEASE']}' ;
+          elif [[ "$TARGET_BRANCH" == "master" ]] ; then
+            MERGED='{"include": ['$JSON_MASTER_ELN']}' ;
+          else
+            OUT_JSON_FULL=''
+            OUT_JSON_PRIMARY=''
+            OUT_JSON_SECONDARY=''
```
jkonecny12, line 63: I guess this is just WIP. Otherwise I don't quite understand why would this else clause be here. There is no reason to create anything if it's not devel, release or master so I would just add `exit 0`.

Also these variables are not set before.

.github/workflows/tests-contributors.yml
```diff
@@ -37,30 +37,77 @@ jobs:
...
+          fi
+
+          if [[ -z $OUT_JSON_FULL ]] ; then
+            # ...here "release" in first array element is replaced with empty string to make the displayed name always the same.
+            OUT_JSON_FULL=$(echo "$MERGED" | jq -c '.include[0].release |= ""')
+            # Turn "release" values into an array to be used as the primary key for matrix. Do not use the whole JSON directly, it generates FUGLY job names.
+            OUT_JSON_PRIMARY=$(echo "$OUT_JSON_FULL" | jq -c '[.include[].release]')
```
jkonecny12, line 70: I don't think you want to complicate things by this. Add everything to include and avoid having the `release:` part completely.

genjson.sh
```diff
@@ -0,0 +1,42 @@
+#!/usr/bin/bash
```
jkonecny12, line 1: This file is here for some reason or is it just WIP?

.github/workflows/tests-contributors.yml
```diff
@@ -37,30 +37,77 @@ jobs:
...
+          echo "$OUT_JSON_SECONDARY" | jq '.'
+          echo "::set-output name=matrix-full::$OUT_JSON_FULL"
+          echo "::set-output name=matrix-primary::$OUT_JSON_PRIMARY"
+          echo "::set-output name=matrix-secondary::$OUT_JSON_SECONDARY"
+
     outputs:
       allowed_user: ${{ steps.check_user_perm.outputs.allowed_user }}
       base_ref: ${{ fromJson(steps.pr_api.outputs.data).base.ref }}
       sha: ${{ fromJson(steps.pr_api.outputs.data).head.sha }}
+      matrix-full: ${{ steps.generate_matrix.outputs.matrix-full }}
```
jkonecny12, line 88: If I see it correctly this is not used anywhere. Let's simplify the logic and everything here and have just one matrix not three.

jkonecny12 (commented):
Few notes to the current idea.

Great work in overall!

.github/workflows/tests-contributors.yml
VladimirSlavik: Thanks, done.

.github/workflows/tests-contributors.yml
VladimirSlavik: Hm, ok, I think this needs also testing on some other branch, but let's see.

.github/workflows/tests-contributors.yml
VladimirSlavik: Agreed. Done.

genjson.sh
VladimirSlavik: WIP. But I guess I don't have to keep it in git.

.github/workflows/tests-contributors.yml
VladimirSlavik: Unfortunately that does not work. So we have to choose - required checks with two matrix variables, or one matrix variable and no required checks. (I like the first option better)

Why: Name of the job is automatically created from the constant in yaml and variables in matrix. If we do just `include` then there is no "primary variable" and everything goes into the status name. There is also no way to set *job* name dynamically (job != step).

.github/workflows/tests-contributors.yml
VladimirSlavik: It works fine! Thanks!

.github/workflows/tests-contributors.yml
```diff
@@ -37,30 +37,73 @@ jobs:
...
+        # The additional values for each release value come via the include: line.
+        release: ${{ fromJson( needs.pr-info.outputs.matrix-primary ) }}
+        include: ${{ fromJson( needs.pr-info.outputs.matrix-secondary ) }}
     env:
-      CI_TAG: '${{ matrix.release }}'
+      CI_TAG: '${{ matrix.ci-tag }}'
       CONTAINER_BUILD_ARGS: '${{ matrix.build-args }}'
-      STATUS_NAME: 'unit-tests${{ matrix.name-postfix }}'
-      TARGET_BRANCH_NAME: origin/master
+      STATUS_NAME: 'unit-tests (${{ matrix.release }})'
```
jkonecny12, line 105: This will not work as expected. This will name the status on PR for main run as:

```
unit-tests ()
```

which won't match required test `unit-tests`.

jkonecny12 (approved):
Looks good to me otherwise. I'm even for merging this now if you want and do the additional change later.

.github/workflows/tests-contributors.yml
VladimirSlavik: Following the discussion: Re-introduce `name-postfix`. It will contain the name in brackets and will be the thing to clear in JSON instead of `release`. Then, the primary JSON can be dropped. The reason why nice job names do not matter is that they do *not* appear in the PR check list, we add these manually via direct API call and control the name there (`context`).

.github/workflows/tests-contributors.yml
VladimirSlavik: Resolved in next round of review below.

VladimirSlavik:
I think this is ready = it addresses most of the points above.

PR #3223, opened 2021-03-05
Add ParseOrganizationData runtime DBus task

M4rtinK:
Add a new runtime DBus task that can query information about the
organizations of a Red Hat account.

Integrate the task with the runtime subscription logic to skip auto-attach if the current organization appears to be in the SCA mode.

~TODO:~
- ~task integration~
- ~unit tests for task integrationa~

Test coverage is in place and manual testing in VM on RHEL 8.4 indicates the change works correctly under real circumstances as well.

### DIFF
--- a/pyanaconda/modules/subscription/runtime.py
+++ b/pyanaconda/modules/subscription/runtime.py
@@ -31,9 +31,12 @@ from pyanaconda.modules.common.constants.objects import RHSM_REGISTER
 from pyanaconda.modules.common.errors.subscription import RegistrationError, \
     UnregistrationError, SubscriptionError
 from pyanaconda.modules.common.structures.subscription import AttachedSubscription, \
-    SystemPurposeData
+    SystemPurposeData, OrganizationData
 from pyanaconda.modules.subscription import system_purpose
+from pyanaconda.modules.subscription.subscription_interface import \
+    ParseOrganizationDataTaskInterface
 from pyanaconda.anaconda_loggers import get_module_logger
+from pyanaconda.core.glib import Variant
 
 import gi
 gi.require_version("Gio", "2.0")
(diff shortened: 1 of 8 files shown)


pyanaconda/modules/subscription/runtime.py
```diff
@@ -570,3 +570,118 @@ def run(self):
...
+            locale = os.environ.get("LANG", "")
+            private_register_proxy = private_bus.get_proxy(RHSM.service_name,
+                                                           RHSM_REGISTER.object_path)
+            org_data_json = private_register_proxy.GetOrgs(self._username,
+                                                           self._password,
+                                                           {},
+                                                           locale)
+            log.debug("subscription: got organization data")
+
+            org_data_length = 0
```
poncovka, line 669: This line is redundant.

pyanaconda/modules/subscription/runtime.py
```diff
@@ -570,3 +570,118 @@ def run(self):
...
+            # Log how much organization data we got for debugging purposes.
+            # By only logging length, we should be able to debug cases of no
+            # or incomplete data being logged, without logging potentially
+            # sensitive subscription status detail into the installation logs
+            # stored on the target system.
```
poncovka, line 674: Is this really necessary?

poncovka (commented):

pyanaconda/modules/subscription/runtime.py
M4rtinK: We already have this for the other calls that work with JSON input and it seems to me safer to have it in place so we can do basic debugging based on log output if needed (eq. zero length JSON, very long JSON (possibly getting a HTML error page instead of the JSON document we expect, etc.).

pyanaconda/modules/subscription/runtime.py
M4rtinK: Dropped.

pyanaconda/modules/subscription/runtime.py
poncovka: In that case, could you please remove these eleven lines of code and comments and extend the debugging message above? You can do simply:

`log.debug("subscription: got organization data (%d characters)", len(org_data_json))`

It really doesn't deserve so much attention and creates the illusion, that there is something important going on.

VladimirSlavik (approved):
Looks good to me. Thank you!

pyanaconda/modules/subscription/runtime.py
```diff
@@ -570,3 +573,113 @@ def run(self):
         # return the DBus structures as a named tuple
         return SystemSubscriptionData(attached_subscriptions=attached_subscriptions,
                                       system_purpose_data=system_purpose_data)
+
+
+class ParseOrganizationDataTask(Task):
+    """Parse data about organizations the given Red Hat account is a member of.
+
+    NOTE: While most Red Hat accounts are a memeber of a single organization,
```
jkonecny12, line 581: `memeber` -> `member`

jkonecny12 (approved):
Looks good to me except the nitpick below,

pyanaconda/ui/lib/subscription.py
```diff
@@ -108,6 +111,40 @@ def check_cdn_is_installation_source(payload):
...
+    org_data_list = OrganizationData.from_structure_list(org_data_structs)
+    if org_data_list:
+        if len(org_data_list) > 1:
+            log.warning("subscription thread: more than one organization returned "
+                        "for Red Hat account currently in use")
+            log.warning(org_data_list)
+        # On hosted Candlepin there should be exactly one organization per user account,
+        # on custom Candlepin instances there could be (and often are).
+        # For now just always pick the first one as the authoritative organization.
+        return org_data_list[0].simple_content_access_enabled
```
poncovka, line 138: That's a lot of code to simply check that there is an organization with the simple content access enabled. Are there some specific plans for using `OrganizationData` outside of this use case?

pyanaconda/ui/lib/subscription.py
```diff
@@ -287,17 +332,21 @@ def register_and_subscribe(payload, progress_callback=None, error_callback=None,
...
-        error_callback(str(e))
-        return
+    if sca_enabled:
+        # skip auto-attach as it is not necessary in Simple Content Access mode
+        log.debug("subscription thread: auto attach skipped due to Simple Content Access mode")
+    else:
+        # not in SCA mode - try to auto attach subscription
+        log.debug("subscription thread: attempting to auto attach an entitlement")
+        progress_callback(SubscriptionPhase.ATTACH_SUBSCRIPTION)
+        task_path = subscription_proxy.AttachSubscriptionWithTask()
```
poncovka, line 342: So, the UI actually doesn't need to know anything about the Simple Content Access mode. It could be all handled in the `AttachSubscriptionTask` class that would internally check the mode and skip the rest. Since we want to keep the DBus API as simple as possible, I definitely prefer that solution.

poncovka (changes requested):
It looks like we don't need to extend the DBus API to support this use case. It can be handled internally in the DBus module.

rvykydal:
@M4rtinK is this PR still alive or should we close it ?

VladimirSlavik:

PR #3224, opened 2021-03-05
ostree: ignore exit code 65 for systemd-tmpfiles

gicmo:
When pre-filling /var after deploying the ostree commit, do not interpret return code 65 by systemd-tmpfiles as an error. This return code is used to indicate that some of the lines in the configuration were ignore, which could be due to a syntax error but also something as harmless as duplicated entries. Either way it the contents of `/lib/tmpfiles.d` is provided the content and thus out of the hand of the installer itself. Since there was no error executing the not-ignored lines we should not abort the installation.

### DIFF
--- a/pyanaconda/payload/rpmostreepayload.py
+++ b/pyanaconda/payload/rpmostreepayload.py
@@ -392,11 +392,26 @@ class RPMOSTreePayload(Payload):
         # matching NSS configuration inside Anaconda, and we can't "chroot" to
         # get it because that would require mounting the API filesystems in the
         # target.
+        cmd = "systemd-tmpfiles"
         for varsubdir in ('home', 'roothome', 'lib/rpm', 'opt', 'srv',
                           'usrlocal', 'mnt', 'media', 'spool', 'spool/mail'):
-            self._safe_exec_with_redirect("systemd-tmpfiles",
-                                          ["--create", "--boot", "--root=" + conf.target.system_root,
-                                           "--prefix=/var/" + varsubdir])
+            argv = ["--create", "--boot", "--root=" + conf.target.system_root,
+                    "--prefix=/var/" + varsubdir]
+
+            rc = util.execWithRedirect(cmd, argv)
+
+            # According to systemd-tmpfiles(8), the return values are:
+            #  0 → success
+            # 65 → so some lines had to be ignored, but no other errors
+            # 73 → configuration ok, but could not be created
+            #  1 → other error
+            # Therefore we ignore error 65, since this is coming from
+            # the payload itself and the actual execution of it was fine
+            if rc not in [0, 65]:
+                exn = PayloadInstallError(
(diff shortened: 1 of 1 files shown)


poncovka:
Please, report a bug in bugzilla and add `Resolves: rhbz#<bug-number>` at the end of the commit  message.

gicmo:
Sorry about that, added the `Resolves:` line.

pyanaconda/payload/rpmostreepayload.py
```diff
@@ -392,11 +392,26 @@ def _prepare_mount_targets(self):
...
+            # According to systemd-tmpfiles(8), the return values are:
+            #  0 → success
+            # 65 → so some lines had to be ignored, but no other errors
+            # 73 → configuration ok, but could not be created
+            #  1 → other error
+            # Therefore we ignore error 65, since this is coming from
+            # the payload itself and the actual execution of it was fine
+            if rc not in [0, 65]:
+                exn = PayloadInstallError(
+                    "%s %s exited with code %d" % (cmd, argv, rc))
```
VladimirSlavik, line 412: We use `format` for all new code. So `"{} {} exited with code {}".format(cmd, argv, rc)` would be better.

VladimirSlavik, line 412: I'm not 100% sure about the following, but maybe someone else can comment on that.

My own instinct would be to use here something along the lines `"{} failed to create /var/{}".format(cmd, varsubdir)` to make the message more human when it is displayed. Detailed debug output is available in the logs: All executions are logged with complete options, return code and even captured output in `program.log` and also `journal`.

VladimirSlavik (commented):
Thanks! Main logic looks good to me.

pyanaconda/payload/rpmostreepayload.py
gicmo: Sure!

gicmo: Sure, this was just taken from the function it is replacing in the intention to keep the behaviour similar, but I think it makes sense to change it to something more readable.

poncovka:
Thanks!

```
===============================================
   anaconda 33.16.4.12: tests/test-suite.log
===============================================

# TOTAL: 8
# PASS:  8
# SKIP:  0
# XFAIL: 0
# FAIL:  0
# XPASS: 0
# ERROR: 0

```

VladimirSlavik:
Looks like the test is waiting for a worker, no idea why.

rvykydal (approved):
Looks good to me.

VladimirSlavik (approved):
Oh, I know what I forgot: Looks good to me. Thank you!

poncovka:
Tested with https://bugzilla.redhat.com/show_bug.cgi?id=1935648#c4.

poncovka:
Ported to upstream in https://github.com/rhinstaller/anaconda/pull/3244.

PR #3225, opened 2021-03-05
Remove crun version abort for container refresh workflow

jkonecny12:
This abort logic works great, however, GitHub runners are not updated immediately. That means that this check will abort the workflow for some runners but and the others will work.

Lets remove this and just remove the workaround as whole when we see that the old runners are not available anymore.

Also switch to ubuntu explicit version because latest tag depends on runner you will get. It could be 18.04 or 20.04 based on documentation.

Test run is [here](https://github.com/Test-anaconda-org/anaconda/actions/runs/625003188).

### DIFF
--- a/.github/workflows/container-autoupdate.yml
+++ b/.github/workflows/container-autoupdate.yml
@@ -55,6 +55,10 @@ jobs:
           make
           sudo make install
 
+          # remove crun folder because otherwise pylint will start the check there
+          cd ..
+          rm -rf crun
+
           echo "New crun version:"
           crun --version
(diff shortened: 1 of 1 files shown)


VladimirSlavik (approved):
Let's do this!

jkonecny12:
**UPDATED**
* Ahh it was my fault that the ELN ci tests failed. Pylint tested also crun code not just us. Remove the folder.

[New run.](https://github.com/Test-anaconda-org/anaconda/actions/runs/625029754)

PR #3226, opened 2021-03-06
network: remove unneeded line wrap

rffontenelle:
The offending line has a line wrap that has no impact in the UI itself, but it raises questions for the translator (e.g. whether this should be 80-char limited line, and where to break the line), resulting in slowing down translation expending little extra-time to investigate a single string.  Besides, there is another big label text a few lines below not wrapped.

### DIFF
--- a/pyanaconda/ui/gui/spokes/network.glade
+++ b/pyanaconda/ui/gui/spokes/network.glade
@@ -418,8 +418,7 @@
                   <object class="GtkLabel" id="networkNeedLabel1">
                     <property name="visible">True</property>
                     <property name="can_focus">False</property>
-                    <property name="label" translatable="yes">We'll need network access to fetch information about your location and to make software
-updates available for you.</property>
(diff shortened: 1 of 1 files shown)


rvykydal (approved):
Looks good to me

poncovka (approved):

VladimirSlavik:
I tried to test this but could not bring up that spoke at all. @rvykydal do you have any idea what this is and how to test it?

VladimirSlavik (approved):
Tested - thanks @poncovka. Looks exactly as it should. The only drawback is that it invalidates existing translations, but that's life, I guess...

rffontenelle:
FYI: Small string changes like this will result in a new "fuzzy" string for translation, which means the previous text will be suggested to the translator, and a single newline char removal will do the job. However, If no action is applied, the English text will be displayed though.

Thanks for the merge!

VladimirSlavik:
Yes - the "no action" is exactly what I meant. There are many translations that receive updates only rarely.

PR #3227, opened 2021-03-08
Use the volume UUID to search for the GRUB config in btrfs partitions

martinezjavier:
For UEFI installs, a minimal GRUB config file is created in the EFI System
Partition that is used to load the main config file that is located in the
/boot/grub2 directory. The partition that contains the latter, is found by
the minimal config searching a device with a given filesystem UUID.

But for this to work the stage2 device must have a filesystem UUID set and
blivet doesn't set this for partitions formatted with a btrfs filesystem.

Because each btrfs volume will have its own UUID, the UUID of the volume
that's mounted according to /etc/fstab must be used to search the device.

Resolves: rhbz#1930567

### DIFF
--- a/pyanaconda/modules/storage/bootloader/efi.py
+++ b/pyanaconda/modules/storage/bootloader/efi.py
@@ -172,7 +172,13 @@ class EFIGRUB(EFIBase, GRUB2):
 
         with open(config_path, "w") as fd:
             grub_dir = self.config_dir
-            fs_uuid = self.stage2_device.format.uuid
+            if self.stage2_device.format.type != "btrfs":
+                fs_uuid = self.stage2_device.format.uuid
+            else:
+                fs_uuid = self.stage2_device.format.vol_uuid
+
+            if fs_uuid is None:
+                raise BootLoaderError("Could not get stage2 filesystem UUID")
 
             grub_dir = util.execWithCapture("grub2-mkrelpath", [grub_dir],
                                             root=conf.target.system_root)


cmurf:
@vojtechtrefny what do you think? Does Btrfs need to be special cased because a subvolume is created for any mountpoint only on Btrfs? Or could it be treated the same as ext4 and XFS when it comes to FS UUID?

martinezjavier:
For reference this was changed in blivet by the following commit storaged-project/blivet@b3042c63e93f.

I would like to understand why that change was made in the first place.

vojtechtrefny:
I'm not sure why we have a special "exception" for btrfs here. Having the UUID set correctly for every subvolume and using `format.uuid` for all filesystems should be possible. I'm not sure why the change with `vol_uuid` was implemented. @dwlehman do you remember what was the reason for it? It's possible the reason behind this is to have the UUID unique in blivet and not having having multiple devices with the same UUID (internally we treat btrfs subvolumes as devices).

martinezjavier:
> It's possible the reason behind this is to have the UUID unique in blivet and not having having multiple devices with the same UUID (internally we treat btrfs subvolumes as devices)

Yes, that was my assumption since I noticed that `BTRFSVolumeDevice` inherits from `BTRFSDevice` which in turn inherits from `StorageDevice`.

poncovka (approved):

VladimirSlavik (approved):
+1 Looks good to me. Thank you!

poncovka:
Tested.

PR #3228, opened 2021-03-09
Do not create a flat file /etc/resolv.conf during installation

keszybz:
We want a symlink, as in the live image, not a file with static contents
which will get out of date. (Right now, /mnt/sysroot/etc/resolv.conf during
installation from live image has:
```
  # This is /run/systemd/resolve/stub-resolv.conf managed by man:systemd-resolved(8).
  # Do not edit.
```
which is not suitable for being stored permanently in a config file in /etc.)

https://bugzilla.redhat.com/show_bug.cgi?id=1933454

### DIFF
--- a/pyanaconda/network.py
+++ b/pyanaconda/network.py
@@ -230,7 +230,7 @@ def copy_resolv_conf_to_root(root="/"):
     dst_dir = os.path.dirname(dst)
     if not os.path.isdir(dst_dir):
         util.mkdirChain(dst_dir)
-    shutil.copyfile(src, dst)
+    shutil.copyfile(src, dst, follow_symlinks=False)
 
 
 def run_network_initialization_task(task_path):


rvykydal:
Hm, I wonder if the code is run on LiveOS given this guard:
https://github.com/rhinstaller/anaconda/blob/9a2358fdd3d126ab9925155285d2db4caa6b8b0e/pyanaconda/installation.py#L281
https://github.com/rhinstaller/anaconda/blob/9a2358fdd3d126ab9925155285d2db4caa6b8b0e/pyanaconda/core/configuration/system.py#L176
I'll check it.

rvykydal:
So the /etc/resolv.conf is copied at the end of installation in https://github.com/rhinstaller/anaconda/blob/9a2358fdd3d126ab9925155285d2db4caa6b8b0e/pyanaconda/installation.py#L132
https://github.com/rhinstaller/anaconda/blob/9a2358fdd3d126ab9925155285d2db4caa6b8b0e/pyanaconda/modules/network/installation.py#L220
as also observation in https://bugzilla.redhat.com/show_bug.cgi?id=1933454#c15 confirms.
But I'd expect the task not be executed on LiveOS, I need to check with F33.

rvykydal:
I commented more in https://bugzilla.redhat.com/show_bug.cgi?id=1933454#c21

pyanaconda/network.py
```diff
@@ -230,8 +230,13 @@ def copy_resolv_conf_to_root(root="/"):
...
     if not os.path.isdir(dst_dir):
         util.mkdirChain(dst_dir)
-    shutil.copyfile(src, dst)
 
+    try:
+        target = os.readlink(src)
+    except OSError:
+        shutil.copyfile(src, dst)
+    else:
+        os.symlink(target, dst)  # symlink() takes symlink path as second argument
```
VladimirSlavik, line 239: I wonder why the `else` clause is needed at all? Is it because you want to conditionalize only the errors from `readlink`? In that case, why not use a plain if-else with `os.path.islink`?
```
if os.path.islink(src):
    target = os.readlink(src)
    os.symlink(target, dst)  # symlink() takes symlink path as second argument
else:
    shutil.copyfile(src, dst)
```

keszybz: That version does more syscalls and suffers from a TOCTOU condition.
But a better fix is to use `follow_symlinks=True` in any case. I didn't know shutil.copyfile() supports that.

Anywa, IIUC,  #3230 obsoletes this PR.

VladimirSlavik: Oh, I didn't consider that... thanks!

keszybz: I updated the patch to use `follow_symlinks`, just in case anyone ever looks at this again. But let's close the PR.

PR #3232, opened 2021-03-09
Move the code for space and location requirements

poncovka:
Move the `get_df_map` and `pick_mount_point` functions to the Payloads module.

Run the `PrepareDownloadLocationTask` task to find and prepare the download location.

Call the `calculate_required_space` function to get the space required for
the package installation.


VladimirSlavik (approved):
Looks good to me. Thank you!

pyanaconda/modules/payloads/payload/dnf/utils.py
```diff
@@ -219,6 +221,30 @@ def get_df_map():
     return mapping
 
 
+def get_sysroot_df_map():
```
jkonecny12, line 224: If I'm reading it correctly this doesn't use `df` at all now.

pyanaconda/modules/payloads/payload/dnf/utils.py
```diff
@@ -184,3 +187,103 @@ def get_kernel_version_list():
...
+    for mount_point, size in mount_points.items():
+
+        # Ignore mount points that are not suitable.
+        if mount_point not in suitable:
+            continue
+
+        elif size > (download_size + install_size):
+            sufficient.add(mount_point)
+
+        elif mount_point != conf.target.system_root and size > download_size:
```
jkonecny12, line 314: Shouldn't we check if `mount_point` starts with system root?

jkonecny12 (changes requested):
Looks good to me otherwise. However, could you please answer my questions below?

rvykydal (approved):
Looks good to me.

pyanaconda/modules/payloads/payload/dnf/utils.py
poncovka: It returns the same output as `get_df_map`, but we cannot call `df` in this case, because the partitioning doesn't exist yet. Both functions should use the same name. What about `get_space_usage` and `get_sysroot_space_usage`?

pyanaconda/modules/payloads/payload/dnf/utils.py
poncovka: This is the original implementation, but I had the same thought. I can fix it in another commit.

pyanaconda/modules/payloads/payload/dnf/utils.py
poncovka: Fixed.

pyanaconda/modules/payloads/payload/dnf/utils.py
poncovka: Fixed.

poncovka:
Updated and refactorized.

jkonecny12 (approved):
Look great to me now. Thanks!

PR #3233, opened 2021-03-11
Handle the module command in the Payloads module

poncovka:
Keep the data about enabled and disabled modules in the packages selection.

Call the `enable_modules` and `disable_modules` methods of the DNF manager
to enable and disable the specified modules.

### DIFF
--- a/pyanaconda/modules/payloads/payload/dnf/dnf.py
+++ b/pyanaconda/modules/payloads/payload/dnf/dnf.py
@@ -164,6 +164,17 @@ class DNFModule(PayloadBase):
         for group in data.packages.excludedGroupList:
             selection.excluded_groups.append(group.name)
 
+        for module in data.module.dataList():
+            name = module.name
+
+            if module.stream:
+                name += ":" + module.stream
+
+            if module.enable:
+                selection.modules.append(name)
+            else:
+                selection.disabled_modules.append(name)
+
         self.set_packages_selection(selection)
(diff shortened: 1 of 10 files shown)


pyanaconda/modules/payloads/payload/dnf/dnf.py
```diff
@@ -245,6 +256,25 @@ def _set_up_kickstart_packages_selection(self, data):
...
+            self._set_up_kickstart_module_data(data, name)
+
+        for name in selection.disabled_modules:
+            self._set_up_kickstart_module_data(data, name, False)
+
+    @staticmethod
+    def _set_up_kickstart_module_data(data, name, enabled=True):
+        """Set up the kickstart data for the module command."""
+        names = name.split(":", maxsplit=1)
+        names += ["", ""]
```
rvykydal, line 269: Maybe adding just [""] would be enough here IIUIC, or are two "" items safer or more readable?
Hm, sounds like friday nitpicking, feel free to ignore...

rvykydal (approved):
Looks good to me.

VladimirSlavik (approved):
Looks good to me. Thank you!

pyanaconda/modules/payloads/payload/dnf/dnf.py
poncovka: Fixed.

PR #3234, opened 2021-03-11
Use the RPM OSTree module to install the payload

poncovka:
Use the DBus API of the RPM OSTree module in the RPM OSTree payload class.

### DIFF
--- a/pyanaconda/payload/rpmostreepayload.py
+++ b/pyanaconda/payload/rpmostreepayload.py
@@ -19,13 +19,15 @@
 #
 from blivet.size import Size
 
-from pyanaconda.core.constants import PAYLOAD_TYPE_RPM_OSTREE, SOURCE_TYPE_RPM_OSTREE
+from dasbus.client.proxy import get_object_path
+
+from pyanaconda.core.constants import PAYLOAD_TYPE_RPM_OSTREE, SOURCE_TYPE_RPM_OSTREE, \
+    SOURCE_TYPE_FLATPAK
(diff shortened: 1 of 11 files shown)


pyanaconda/payload/rpmostreepayload.py
```diff
@@ -54,149 +53,57 @@ def source_type(self):
...
-        # Following up on the "remote delete" earlier, we removed the remote from
-        # /ostree/repo/config.  But we want it in /etc, so re-add it to /etc/ostree/remotes.d,
-        # using the sysroot path.
-        #
-        # However, we ignore the case where the remote already exists, which occurs when the
-        # content itself provides the remote config file.
-        #
-        # Note here we use the deployment as sysroot, because it's that version of /etc that we
-        # want.
-
```
rvykydal, line 182: Shouldn't we keep this information somewhere? Maybe in post_install_with_tasks.
Not sure if it is still accurate and comprehensible though.

rvykydal (approved):
Looks good to me.

VladimirSlavik (commented):
Code LGTM. But I don't think the comments can just go like that.

pyanaconda/payload/rpmostreepayload.py
poncovka: The comment was moved to the `ChangeOSTreeRemoteTask` class.

VladimirSlavik: Thanks!

VladimirSlavik (approved):
Looks good to me. Thank you!

PR #3237, opened 2021-03-11
Make failure in generating of dracut arguments for iSCSI device non-f…

rvykydal:
…atal.

The failure indicates a serious problem / inconsistncy.
It can be caused also eg by attaching an iSCSI device outside anaconda.
We should not crash on it.
We might want to ignore it because missing dracut boot option can be
worked around.

### DIFF
--- a/pyanaconda/modules/storage/iscsi/iscsi.py
+++ b/pyanaconda/modules/storage/iscsi/iscsi.py
@@ -220,6 +220,10 @@ class ISCSIModule(KickstartBaseModule):
 
         blivet_node = iscsi.get_node(node.name, node.address, node.port, node.iface)
 
+        if not blivet_node:
+            log.error("No iSCSI node %s found for device", node)
+            return []
+
         address = blivet_node.address
         # surround ipv6 addresses with []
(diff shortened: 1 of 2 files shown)


VladimirSlavik (approved):
Looks good to me. Thank you!

pyanaconda/modules/storage/iscsi/iscsi.py
```diff
@@ -220,6 +220,10 @@ def get_dracut_arguments(self, node):
 
         blivet_node = iscsi.get_node(node.name, node.address, node.port, node.iface)
 
+        if not blivet_node:
```
poncovka, line 223: Isn't there a unit test that could be extended?

poncovka (approved):

pyanaconda/modules/storage/iscsi/iscsi.py
rvykydal: I haven't found any (there is only a simple one for fcoe), but I'll add one.

poncovka (approved):
Thank you!

PR #3238, opened 2021-03-12
Change removing NTP servers in GUI

VladimirSlavik:
Previously, one had to untick the "use" checkbox for the server or pool and exit the dialog, thus confirming changes for the whole list. Now, there is a remove button that removes the server/pool immediately.

Resolves: [rhbz#1827683](https://bugzilla.redhat.com/show_bug.cgi?id=1827683)

### DIFF
--- a/pyanaconda/ui/gui/spokes/datetime_spoke.glade
+++ b/pyanaconda/ui/gui/spokes/datetime_spoke.glade
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<!-- Generated with glade 3.18.3 -->
+<!-- Generated with glade 3.36.0 -->
 <interface>
   <requires lib="gtk+" version="3.12"/>
   <requires lib="AnacondaWidgets" version="1.0"/>
@@ -40,6 +40,11 @@
   </object>
   <object class="GtkTreeModelFilter" id="daysFilter">
     <property name="child_model">days</property>
+  </object>
+  <object class="GtkImage" id="delImage">
+    <property name="visible">True</property>
+    <property name="can_focus">False</property>
+    <property name="icon_name">list-remove-symbolic</property>
   </object>
   <object class="GtkImage" id="downImage">
     <property name="visible">True</property>
@@ -87,8 +92,6 @@
       <column type="gboolean"/>
       <!-- column-name working -->
       <column type="gint"/>
-      <!-- column-name use -->
-      <column type="gboolean"/>
       <!-- column-name object -->
       <column type="PyObject"/>
     </columns>
@@ -147,8 +150,8 @@
           <object class="GtkLabel" id="configHeadingLabel">
             <property name="visible">True</property>
             <property name="can_focus">False</property>
-            <property name="xalign">0</property>
             <property name="label" translatable="yes">Add and mark for usage NTP servers</property>
+            <property name="xalign">0</property>
             <attributes>
               <attribute name="font-desc" value="Sans Bold 12"/>
             </attributes>
@@ -198,6 +201,26 @@
                 <property name="fill">True</property>
                 <property name="padding">4</property>
                 <property name="position">1</property>
+              </packing>
+            </child>
+            <child>
+              <object class="GtkButton" id="delButton">
+                <property name="visible">True</property>
+                <property name="can_focus">True</property>
+                <property name="receives_default">True</property>
+                <property name="image">delImage</property>
+                <signal name="clicked" handler="on_del_clicked" swapped="no"/>
+                <child internal-child="accessible">
+                  <object class="AtkObject" id="delButton-atkobject">
+                    <property name="AtkObject::accessible-name" translatable="yes">Remove NTP Server</property>
+                  </object>
+                </child>
+              </object>
+              <packing>
+                <property name="expand">False</property>
+                <property name="fill">True</property>
+                <property name="padding">4</property>
+                <property name="position">2</property>
               </packing>
             </child>
           </object>
@@ -261,8 +284,8 @@
                       <object class="GtkCellRendererText" id="hostnameRenderer">
                         <property name="editable">True</property>
                         <signal name="edited" handler="on_server_edited" swapped="no"/>
-                        <signal name="editing-started" handler="on_server_editing_started" swapped="no"/>
                         <signal name="editing-canceled" handler="on_server_editing_canceled" swapped="no"/>
+                        <signal name="editing-started" handler="on_server_editing_started" swapped="no"/>
                       </object>
                       <attributes>
                         <attribute name="text">0</attribute>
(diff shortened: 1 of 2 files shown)


VladimirSlavik:
Before:
Screenshot_vs_main_2021-03-12_13:14:32

After:
Screenshot_vs_main_2021-03-12_12:56:52

VladimirSlavik:
@AdamWill this might break OpenQA as it changes the GUI - if you do custom ntp servers, that is...

poncovka (approved):
Well, it is better. Thanks!

VladimirSlavik:
@mairin can you please take a look?

mairin:
@VladimirSlavik A few suggestions! (I hope you won't regret asking my thoughts here lol, being a designer I cant help but be annoying like this 😂 Feel free to take whatever suggestions you think are reasonable and disregard anything that is unhelpful!)

**Dialog Title**
- This isn't a new problem, but the dialog title is a little awkward. I would suggest something simpler like "Configure Network Time Servers" (instead of "Add and mark for usage NTP servers")

**Column checkbox confusion**
- Is the "Pool" column in the list indicating whether or not it is a pool and is non-editable? If it's non-editable I would suggest a plain black checkmark instead of a checked box that looks editable. (Same for the NTS column.)

**Removing Items from list**
- I think removing the "Use" column is a good idea but I don't see the mechanism for removing an entry in the screenshot. Is that a right-click / context menu action the screenshot didn't pick up? Ohhhh... is it the "-" button in the upper right? I wouldn't have linked that button as being connected to items in the list below. (It looks like it is connected to the text entry, not the list of items.) 
- The GNOME HIG suggests having an "X" button to the far right, per item in the list (see https://developer.gnome.org/hig/stable/lists.html.en )
- I know this dialog is in a unique situation that you don't want another level of popup since the dialog itself is a popup, and three layers of pop-ups in anaconda I think is too much! 
- I think the main issue here is the controls to "add" an item at the top are also being using to "edit" an item that already exists. So one way to fix this and make it more consistent is in the mockup suggestion below, which changes the order of things a little bit to give you a cleaner place to put the remove button:

image266

However I'd suggest even a better method if you were up for it, which would be to kind of emulate the HexChat add network dialog - 

Screenshot from 2021-03-18 11-02-51
Screenshot from 2021-03-18 11-03-00

Then if those in-list checkboxes are actually editable, the whole model would make more sense:

image1333

You could either double-click an entry or hit the edit button to have it change over to an editable text field. And if the checkboxes are checkable then those can also be clicked. 

Just some food for thought anyway!

M4rtinK:
This looks rather confusing to me - even though it's indeed complemetary to the + button, on a first look it now kinda looks like those +/- buttons you sometimes see next to integer entry fields that increment or decrement the value in the field.

Also while + works with value in the entry field, do I understand it correctly that - will remove the currently selected server in the list bellow ? If so, then I'm afraid that's also not very intuitive.

As for a better solution though - I'm not sure myself. :P Just two ideas/suggestions:
- maybe we could use words instead of +/-, "add server" / "remove server", that could remove some of the ambiguity ?
- possibly move the remove button to a new column to the server line in the list and use ❌, clicking it would remove the line ?

I'm sure actually UX people would have much better ideas though - maybe @andreasn ?

mairin:
>     * possibly move the remove button to a new column to the server line in the list and use x, clicking it would remove the line ?

I think this is the best solution as it's suggested by the GNOME HIG, but I didn't want to mock it up given I didn't understand how the checkboxes in the list currently operate :-/

VladimirSlavik:
Ha, thanks! I was pretty blind, I guess. Keep it coming please :)

The "use" column of checkboxes was actually a way of deleting - you had to uncheck that and confirm / exit the dialog. I agree the button is only marginally better.

Once an item is in the list, it is possible to edit all parts of it (except the working icon) - all the data faces the user in an editable form. So actually the last mockup is the best idea, I think.

From the development side, the issue with that is we'd have to use some kind of filler to have an entry, and control for that. I think that's not too much. I think it would be best if the added item actually appeared as being edited - with no name, a caret blinking there and focused and everything? I'm not sure if that's doable, just checking.

There's also the issue that the text entry gives a warning icon if the server hostname is invalid, and the adding button is disabled. So right now an invalid server can't be added. Also editing within the list just drops a change if the new server is not valid. If everything is editable within the list and users can add invalid entries, that's again something we'd have to do differently. I think it should be OK to just drop invalid entries between the dialog runs.

The only missing bit would be the hint of the server invalid warning icon. But that's a poorly discoverable thing, I guess?

mairin:
Let me think of a more GNOME-y way to do this and mock it up for you, so you can consider all the options.

mairin:
@VladimirSlavik Hey can I ask what is the implication of a host being marked as a pool or as supporting NTS? Does either being set impact how the system communicates with them?

mairin:
Here's one idea. Idea taken a little bit from how the GNOME printer dialog works:
Screenshot from 2021-03-18 17-15-51

OK here it is:






I don't know if the tags/labels are available in GTK tho. (But could be replaced with icons.) The printer gear wheel list is a tiny pop up instead of another layer of dialog which is why I thought it would suit very well here to avoid a 3rd layer window. The black things are meant to be on-hover tooltips. Hope this is good food for thought.

jstodola:
Hi, the idea in https://github.com/rhinstaller/anaconda/pull/3238#issuecomment-802366179 with popups IMO doesn't fit into anaconda, it would be another way of configuration not used anywhere else in the installer and it would make the UI even more inconsistent.
I think that the proposals in https://github.com/rhinstaller/anaconda/pull/3238#issuecomment-802009994 are much better options.

VladimirSlavik:
The last round of mockups is great! There's one significant drawback though: To make these, one would have to compose multiple controls and icons into list items, de facto creating the GUI components from the ground up. I'm afraid that is too much effort for the gain, as long as it stays isolated to one dialog. (Do feel free to teach me how it's not! I'd love that.)

So, for the time being, the hexchat-inspired picture below is about as far as we can go in terms of effort spent (IMO). To be specific, this one is ok because it keeps the list with text, checks, and icon, which is already coded - and very simple, based on stock GTK stuff.



Before working on that, I'd like to ask: Do we want to merge the current change with `[-]` button as a temporary solution? Personally I think not, because it might easily turn permanent. Alternatively, it could mess up OpenQA despite being a temporary thing. On the other hand it's at least discoverable, as opposed to unchecking "use", so it has *some* merit.

mairin:
@VladimirSlavik I can't tell you which way to go, but I can tell you that despite its difficulty I think the - button is easier to use than the use checkmark.

mairin:
@jstodola Ah, that's a shame - when I originally designed the anaconda UI it was based off of gnome-control-center. It's a good base bc it has a lot of similar config, but it's also a bad base bc it tends to have a lot of custom stuff that isnt merged upstream and it seems to change a lot. 

@VladimirSlavik I can ask some folks who work with GTK how custom this stuff is. The tag thing definitely is, but we could toss that in favor of icons. I do see this basic list with the + row in the bottom in several dialogs including language and kb selection in g-c-c right now.

VladimirSlavik:
The new PR has the good stuff, closing this.

PR #3239, opened 2021-03-12
IOError cleanup

VladimirSlavik:
I tried to eradicate `IOError` which is just an alias for `OSError`.

* Generally, most of the occurrences were to catch problems after calling `open()` or something from `os`, so the replacement was straightforward.
* There were a few more cases where we raised `IOError` in our code, so it could be replaced with a custom exception class.
* In a few cases, it's not us but Blivet raising `IOError`; in that case, I left it in our code as-is. Feel free to disagree. These were:
  * https://github.com/rhinstaller/anaconda/blob/master/pyanaconda/modules/storage/kickstart.py#L206
  * https://github.com/rhinstaller/anaconda/blob/master/pyanaconda/modules/storage/fcoe/discover.py#L46
* In one case, we have `IOError` guarding use of `Requests`, probably as a leftover from 2013 when we used `urllib`, and I have no idea if we want to keep the whole except there: https://github.com/rhinstaller/anaconda/blob/master/pyanaconda/payload/live/payload_liveimg.py#L104

### DIFF
--- a/pyanaconda/core/startup/dbus_launcher.py
+++ b/pyanaconda/core/startup/dbus_launcher.py
@@ -113,12 +113,12 @@ class AnacondaDBusLauncher(object):
         self._dbus_daemon_process = startProgram(command, stderr=self._log_file, reset_lang=False)
 
         if self._dbus_daemon_process.poll() is not None:
-            raise IOError("DBus wasn't properly started!")
+            raise RuntimeError("DBus wasn't properly started!")
 
         address = self._dbus_daemon_process.stdout.readline().decode('utf-8').strip()
 
         if not address:
-            raise IOError("Unable to start DBus session!")
+            raise RuntimeError("Unable to start DBus session!")
 
         self._bus_address = address
 
--- a/pyanaconda/modules/payloads/installation.py
+++ b/pyanaconda/modules/payloads/installation.py
@@ -117,7 +117,7 @@ class CopyDriverDisksFilesTask(Task):
         for f in glob(self.DD_FIRMWARE_DIR + "/*"):
             try:
                 shutil.copyfile(f, os.path.join(self._sysroot, "lib/firmware/"))
-            except IOError as e:
+            except OSError as e:
                 log.error("Could not copy firmware file %s: %s", f, e.strerror)
 
         # copy RPMS
@@ -129,7 +129,7 @@ class CopyDriverDisksFilesTask(Task):
         if os.path.exists(self.DD_DIR):
             try:
                 shutil.copytree(self.DD_DIR, os.path.join(self._sysroot, "root/DD"))
-            except IOError as e:
+            except (OSError, shutil.Error) as e:
                 log.error("failed to copy driver disk files: %s", e.strerror)
                 # XXX TODO: real error handling, as this is probably going to
                 #           prevent boot on some systems
--- a/pyanaconda/modules/storage/installation.py
+++ b/pyanaconda/modules/storage/installation.py
@@ -321,7 +321,7 @@ class WriteConfigurationTask(Task):
                 device.format.escrow(escrow_dir,
                                      backup_passphrase)
 
-        except (IOError, RuntimeError) as e:
+        except (OSError, RuntimeError) as e:
             # TODO: real error handling
             log.error("failed to store encryption key: %s", e)
 
--- a/pyanaconda/payload/install_tree_metadata.py
+++ b/pyanaconda/payload/install_tree_metadata.py
@@ -31,6 +31,10 @@ log = get_packaging_logger()
 MAX_TREEINFO_DOWNLOAD_RETRIES = 6
 
 
+class FileNotDownloadedError(Exception):
+    pass
+
+
 class InstallTreeMetadata(object):
     # TODO: Add tests for InstallTreeMetadata class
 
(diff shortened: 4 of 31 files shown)


pyanaconda/modules/storage/installation.py
```diff
@@ -321,7 +321,7 @@ def _write_escrow_packets(self, storage, sysroot):
                 device.format.escrow(escrow_dir,
                                      backup_passphrase)
 
-        except (IOError, RuntimeError) as e:
+        except (OSError, RuntimeError) as e:
```
VladimirSlavik, line 324: I could not find where the error would come from, so I'm not sure if the replacement is correct. Perhaps @vojtechtrefny knows more?

pyanaconda/modules/payloads/installation.py
```diff
@@ -129,7 +129,7 @@ def run(self):
         if os.path.exists(self.DD_DIR):
             try:
                 shutil.copytree(self.DD_DIR, os.path.join(self._sysroot, "root/DD"))
-            except IOError as e:
+            except (OSError, shutil.Error) as e:
```
VladimirSlavik, line 132: Added the shutil error because that can be raised too, according to docs.

pyanaconda/core/startup/dbus_launcher.py
```diff
@@ -42,6 +42,10 @@
 __all__ = ["AnacondaDBusLauncher"]
 
 
+class DBusNotStartedError(Exception):
+    pass
```
VladimirSlavik, line 46: I did not see a dasbus exception that could be used here.

VladimirSlavik (commented):
I'd like to call out a few places where I am not too sure about the change.

pyanaconda/modules/storage/installation.py
vojtechtrefny: This probably comes from `util.makedirs`, which uses `os.makedirs`, which can raise `FileExistsError` (subclass of `OSError` so it's correct here).

vojtechtrefny:
> * In a few cases, it's not us but Blivet raising `IOError`; in that case, I left it in our code as-is.

We should probably get rid of those. It doesn't really make sense to raise `IOError(_("FCoE not available"))`, that's not an input/output error.

pyanaconda/modules/storage/installation.py
VladimirSlavik: Thanks!

VladimirSlavik:
> > * In a few cases, it's not us but Blivet raising `IOError`; in that case, I left it in our code as-is.
> 
> We should probably get rid of those. It doesn't really make sense to raise `IOError(_("FCoE not available"))`, that's not an input/output error.

Should I file an issue or bug somewhere?

vojtechtrefny:
> > > * In a few cases, it's not us but Blivet raising `IOError`; in that case, I left it in our code as-is.
> > 
> > 
> > We should probably get rid of those. It doesn't really make sense to raise `IOError(_("FCoE not available"))`, that's not an input/output error.
> 
> Should I file an issue or bug somewhere?

That would be great. GitHub issue referring this PR should be enough. Thanks.

pyanaconda/ui/gui/__init__.py
```diff
@@ -171,7 +171,7 @@ def _findUIFile(self):
             if os.path.isfile(testPath) and os.access(testPath, os.R_OK):
                 return testPath
 
-        raise IOError("Could not load UI file '%s' for object '%s'" % (self.uiFile, self))
+        raise Exception("Could not load UI file '{}' for object '{}'".format(self.uiFile, self))
```
poncovka, line 174: It is not recommended to raise the general exception. I think you can use `OSError` in this case.

poncovka (approved):

pyanaconda/core/startup/dbus_launcher.py
VladimirSlavik: Thank you, done.

pyanaconda/ui/gui/__init__.py
VladimirSlavik: Thank you, done.

pyanaconda/payload/install_tree_metadata.py
```diff
@@ -120,7 +124,9 @@ def load_url(self, url, proxies, sslverify, sslcert, headers):
                 err_msg = ("Repo info download for %s failed after %d retries" %
                            (url, retry_count))
                 log.error(err_msg)
-                raise IOError("Can't get .treeinfo file from the url {}".format(url))
+                raise FileNotDownloadedError(
```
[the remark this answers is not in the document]
VladimirSlavik: I'm not sure either, merely confident :-) The first part is `requests`, which has its own errors, and then there's `_download_treeinfo_file` which is again just calls to requests wrapped in a try-except block. I'm not sure if logs could do it?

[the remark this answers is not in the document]
poncovka: Then it should be fine :-)

poncovka (approved):
It looks great. Thanks!

VladimirSlavik:
Sorry, I just realized I had a too restrictive filter set while searching. This should resolve all `IOError` cases except these named in the original description above.

poncovka (approved):

PR #3240, opened 2021-03-15
users: make root ssh login work with kickstart installation

kkang-wr:
It does not remember the user selection of checkbox "Allow root SSH
login with password". Work with patch for pykickstart to redo
enable or disable root ssh login during kickstart installation.

Signed-off-by: Kai Kang <kai.kang@windriver.com>

It requires patch for pykickstart:

https://github.com/pykickstart/pykickstart/pull/374

It fails with 
coverage/coveralls — Coverage decreased (-0.02%) to 97.03%

I do not know the root cause. But I still send out the patch of anaconda part for review first.

Thanks.


kkang-wr:
From the discussion of https://github.com/pykickstart/pykickstart/pull/374#issuecomment-800277824, the checkbox "Allow root SSH login with password" will be removed in the end, right? So no need to support such an option in pykickstart and anaconda either.

PR #3242, opened 2021-03-15
KS test workflow groups

VladimirSlavik:
Let's make the ISO build in kickstart tests easier to read and debug.

See https://docs.github.com/en/actions/reference/workflow-commands-for-github-actions#grouping-log-lines

### DIFF
--- a/.github/workflows/kickstart-tests.yml
+++ b/.github/workflows/kickstart-tests.yml
@@ -153,22 +153,28 @@ jobs:
           cd /tmp/source
 
           # install build dependencies and lorax
+          echo "::group::Install build dependencies and lorax"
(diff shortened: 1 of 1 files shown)


jkonecny12 (approved):
This is a great improvement! Thanks for learning me a new stuff :)

VladimirSlavik:
It seems this might play hell with browser output, when it automatically scrolls towards error. Not sure. I'll link some results once I have them, so we can see if it works as intended.

VladimirSlavik:
I can't reproduce the issues I encountered with this kind of output (Firefox unable to scroll, large empty spaces, mixed (un)collapsed state). It seems it's safe to use.

PR #3243, opened 2021-03-16
Drop astroid hotfix patch

VladimirSlavik:
Port of #3174 & #3175. All work by @jkonecny12 who debugged the problem and pointed out the existing fix.

This fixes the container build, which in turn enables the rhel-8 unit tests to work. Probably.


jkonecny12 (approved):
LGTM!

PR #3244, opened 2021-03-17
ostree: ignore exit code 65 for systemd-tmpfiles

poncovka:
When pre-filling /var after deploying the ostree commit, do not
interpret return code 65 by systemd-tmpfiles as an error. This
return code is used to indicate that some of the lines in the
configuration were ignore, which could be due to a syntax error
but also something as harmless as duplicated entires. Either way
it the contents of `/lib/tmpfiles.d` is provided the content and
thus out of the hand of the installer itself. Since there was no
error executing the not-ignored lines we should not abort the
installation.

(cherry-picked from a commit fbb3d70)

Related: rhbz#1935648


VladimirSlavik (approved):
Looks good to me. Thank you!

PR #3245, opened 2021-03-17
Fix the GetContainerFreeSpace method

poncovka:
If Blivet returns a negative size of the container free space, return zero.

Resolves: rhbz#1937593

### DIFF
--- a/pyanaconda/modules/storage/partitioning/interactive/scheduler_module.py
+++ b/pyanaconda/modules/storage/partitioning/interactive/scheduler_module.py
@@ -107,11 +107,15 @@ class DeviceTreeSchedulerModule(DeviceTreeModule):
     def get_container_free_space(self, container_name):
         """Get total free space in the specified container.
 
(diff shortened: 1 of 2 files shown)


poncovka:
@vojtechtrefny, I think we have already talked about this solution. Do you agree with this fix?

VladimirSlavik (approved):
Looks good to me. Thank you!

(I think there were at least two PRs in the past related to the whole negative returns thing?)

vojtechtrefny:
> @vojtechtrefny, I think we have already talked about this solution. Do you agree with this fix?

The bug looks suspiciously similar to https://bugzilla.redhat.com/show_bug.cgi?id=1868623 which should be fixed in blivet 3.3.0 https://github.com/storaged-project/blivet/pull/890 Maybe this is a different issue, but we should fix this in blivet too. But I'm ok with your workaround for anaconda in general.

poncovka:
This needs to be fixed in blivet.

PR #3246, opened 2021-03-18
network: clone connections from intramfs to persistent config

rvykydal:
In case of common connection ("Wired connection") being used for multiple
devices in initramfs (which happens in case of ip=dhcp or no ip= specified) the
connection was not used for persistent configuration of the devices and instead
a default connection created by Anaconda was used.  Let's clone the "Wired
connection" into the default connection created by Anaconda so that its
properties (like default ipv6.addr-gen-mode or
ipv4.dhcp-vendor-class-identifier) are preserved.

Originally related to rhbz#1870692

Resolves: rhbz#1910438

Port of
commit 05a80bb599d6e2ea694279ed4257a11a361b894d
commit 71ca92aa4b09b5f186b8e1a8449ca7b4dae53e80
from master

### DIFF
--- a/pyanaconda/modules/network/initialization.py
+++ b/pyanaconda/modules/network/initialization.py
@@ -23,7 +23,8 @@ from pyanaconda.modules.network.network_interface import NetworkInitializationTa
 from pyanaconda.modules.network.nm_client import get_device_name_from_network_data, \
     ensure_active_connection_for_device, update_connection_from_ksdata, \
     add_connection_from_ksdata, bound_hwaddr_of_device, get_connections_available_for_iface, \
-    update_connection_values, commit_changes_with_autoconnection_blocked, is_ibft_connection
+    update_connection_values, commit_changes_with_autoconnection_blocked, is_ibft_connection, \
+    clone_connection_sync
 from pyanaconda.modules.network.ifcfg import get_ifcfg_file_of_device, find_ifcfg_uuid_of_device, \
     get_master_slaves_from_ifcfgs
 from pyanaconda.modules.network.device_configuration import supported_wired_device_types, \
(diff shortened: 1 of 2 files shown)


poncovka (approved):
Looks good to me.

VladimirSlavik (commented):
Looks good to me, except for this bit. I'm not sure about it, so if that's fine, I'll just ack.

pyanaconda/modules/network/initialization.py
```diff
@@ -458,12 +459,17 @@ def run(self):
                               self.name, iface)
                     continue
 
-            # Devices activated in initramfs should have ONBOOT=yes
+            con = self._select_persistent_connection_for_device(device, cons)
+
             has_initramfs_con = any(self._is_initramfs_connection(con, iface) for con in cons)
```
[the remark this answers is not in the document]
rvykydal: I think the scoping is fine and I'd like to keep the code as close to master (from where I am porting the patch) as possible.

The method size is kind of legacy but I think it was a bit improved on master, esp. the bottom part (although it has a bit grown in its upper part due to fixes of some cases).

Of course feel free to tidy the method up on the master, coverage of our kickstart tests for network regressions is getting better and better ;)

VladimirSlavik (approved):

poncovka (approved):
Looks good to me.

pyanaconda/modules/network/nm_client.py
```diff
@@ -950,6 +950,39 @@ def finish_callback(nm_client, result, sync_queue):
     return ret
 
 
+def clone_connection_sync(nm_client, connection, con_id=None, uuid=None):
```
jkonecny12, line 953: I'm missing docstring here. It would be great to have an explanation here.

jkonecny12: Also I think this should be handled by a decorator somehow. If there is no available we should create it.

jkonecny12 (approved):
As far as I understand it seems good to me. I just found a few nitpicks below.

pyanaconda/modules/network/nm_client.py
rvykydal: I've added the docstring porting commit 010e5fc8dc6de9fd7ecd20d4bcd0815dfa8a1555 .
The decorator would be something for upstream, but we will to have to rework this part anyway.

PR #3161, opened 2021-02-11
Remove product-specific data from the Anaconda stylesheet

poncovka:
Anaconda doesn't provide images from the pixmaps directory, so it doesn't make
sense to use them in our stylesheet. Pixmaps are provided by the *-logos
packages, so they should provide these data as well.

**Some notes:**

* Fedora Server and Fedora Workstation already work this way. They use a custom stylesheet that is placed in a subdirectory of `pixmaps` together with their images. Everything is provided by `fedora-logos`.
* Anaconda will use the Fedora background color and no logo by default (for example, if the product configuration doesn't specify the `custom_stylesheet` option).
* Initial Setup will use the product's stylesheet after merging https://github.com/rhinstaller/initial-setup/pull/117.

**TODO:**

- [x] Move `redhat.css` to `redhat-logos` (see https://bugzilla.redhat.com/show_bug.cgi?id=1928711)
- [x] Use the RHEL-specific stylesheet from `redhat-logos` (see https://github.com/rhinstaller/anaconda/pull/3179)
- [x] Move `fedora.css` to `fedora-logos` (see https://pagure.io/fedora-logos/issue/9).
- [x] Drop the commit that adds `fedora.css` and `redhat.css` files.

PR #3247, opened 2021-03-18
Use a custom stylesheet to define Fedora-specific stylesheet data

poncovka:
Use a stylesheet provided by the `fedora-logos` package.

**Related to:** https://github.com/rhinstaller/anaconda/pull/3161


poncovka:
We are still waiting for a new release of `fedora-logos`, but we can start to load the Fedora stylesheet now. If the file doesn't exist, we will log an error and continue. Our default stylesheet still has the Fedora-specific data, so the look will be the same.

VladimirSlavik (approved):
Looks good to me. Thank you!

PR #3248, opened 2021-03-18
Document SSH root password login issues & possible workaround

M4rtinK:
Document the current situation around root password based SSH logins,
which are disabled by default. In short - don't do that, it makes you
machine vulnerable to easy root password guessing over network.

Still some users apparently still need/want this, so add a simple yet
explicit workaround for them to use to enable password based root logins
over SSH via kickstart using a %post script snippet.


poncovka (approved):
Looks good to me. Thanks!

VladimirSlavik (approved):
Looks good to me. Thank you!

PR #3249, opened 2021-03-19
Create /tmp with the right permissions (#1937626)

poncovka:
The `/tmp` directory of the target system could be created with wrong permissions,
so we need to create this directory and set up its permission right before it is
used to mount a device.

Resolves: rhbz#1937626

### DIFF
--- a/pyanaconda/modules/storage/devicetree/fsset.py
+++ b/pyanaconda/modules/storage/devicetree/fsset.py
@@ -591,6 +591,16 @@ class FSSet(object):
             if read_only:
                 options = "%s,%s" % (options, read_only)
 
+            # Create /tmp with the right permissions (rhbz#1937626).
+            # It needs to be created right before we mount anything.
+            # Call chmod to enforce the mode.
+            if device.format.mountpoint == "/tmp":
+                path = os.path.join(root_path, "tmp")
+
+                if not os.path.exists(path):
+                    os.makedirs(path)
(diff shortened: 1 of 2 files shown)


pyanaconda/modules/storage/devicetree/fsset.py
```diff
@@ -591,6 +591,15 @@ def mount_filesystems(self, root_path="", read_only=None, skip_root=False):
...
                 options = "%s,%s" % (options, read_only)
 
+            # Create /tmp with the right permissions (rhbz#1937626).
+            # It needs to be created right before we mount anything
+            if device.format.mountpoint == "/tmp":
+                path = os.path.join(root_path, "tmp")
+
+                if not os.path.exists(path):
+                    os.makedirs(path)
+                    os.chmod(path, 0o1777)
```
VladimirSlavik, line 601: I think you can call makedirs with the mode already. You actually check for the default in the test...

poncovka: Unfortunately, `makedirs` doesn't enforce the mode (see [the documentation of mkdir](https://docs.python.org/3/library/os.html#os.mkdir)), so we have to call `chmod`. I have changed the test, because it didn't check the right call.

poncovka:
Updated.

VladimirSlavik (approved):

PR #3250, opened 2021-03-19
Clean up the dependencies (#1938132)

poncovka:
The `anaconda` package shouldn't depend on `anaconda-install-env-deps`.
It was a temporary workaround when the meta package was created.

The `anaconda-gui` and `anaconda-tui` packages require `anaconda-core`, so
the `anaconda` packages doesn't need to explicitly require it.

The `anaconda-install-img-deps` package shouldn't recommend `zram-generator-defaults`.
It is required only for boot.iso and Live OS and that is handled by the lorax templates
and the `core` group.

Resolves: rhbz#1938132

### DIFF
--- a/anaconda.spec.in
+++ b/anaconda.spec.in
@@ -84,7 +84,6 @@ BuildRequires: libxml2
 Requires: anaconda-core = %{version}-%{release}
 Requires: anaconda-gui = %{version}-%{release}
 Requires: anaconda-tui = %{version}-%{release}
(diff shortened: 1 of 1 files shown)


M4rtinK (approved):
Looks good to me, although I would suggest ammending the last commit with more detailed information about where zram-generator-defaults is handled for live and boot iso.

poncovka:
> Looks good to me, although I would suggest ammending the last commit with more detailed information about where zram-generator-defaults is handled for live and boot iso.

Fixed.

PR #3251, opened 2021-03-19
Check if the mount point exists before calling statvfs (#1824357)

poncovka:

Resolves: rhbz#1824357


VladimirSlavik (approved):
Looks good to me. Thank you!

PR #3252, opened 2021-03-19
Fix crash in NTP servers GUI dialog on late network configuration

rvykydal:
Resolves: rhbz#1938168

### DIFF
--- a/pyanaconda/ui/gui/spokes/datetime_spoke.py
+++ b/pyanaconda/ui/gui/spokes/datetime_spoke.py
@@ -324,15 +324,18 @@ class NTPconfigDialog(GUIObject, GUIDialogInputCheckHandler):
         #check if we are in the same epoch as the dialog (and the serversStore)
         #and if the server wasn't changed meanwhile
         if epoch_started == self._epoch:
-            actual_hostname = self._serversStore[itr][SERVER_HOSTNAME]
+            # rhbz#1938168 - when called on Spoke enter / refresh the store can
+            # be cleared by opening NTPConfigDialog
+            if self._serversStore.iter_is_valid(itr):
+                actual_hostname = self._serversStore[itr][SERVER_HOSTNAME]
 
-            if orig_hostname == actual_hostname:
-                if server_working:
-                    set_store_value((self._serversStore,
-                                     itr, SERVER_WORKING, constants.NTP_SERVER_OK))
-                else:
-                    set_store_value((self._serversStore,
-                                     itr, SERVER_WORKING, constants.NTP_SERVER_NOK))
+                if orig_hostname == actual_hostname:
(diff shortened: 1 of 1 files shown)


rvykydal:
The crash happens when trying to update an iterator from the Gtk model for list of NTP servers after the iterator becomes invalid (by clearing the model):

- User enters DateTime Spoke -> on spoke refresh the model refresh (of ntp server state) is started: for iterator ITR_X from the model
- THREAD_Y starts to check server pointed by ITR_X with ntp_server_workig() method
- User enters NTP configuration dialog -> model is reinitalized / cleared (servers/items are added from scratch)
- THREAD1 finishes the ntp_server_working() check and tries to update the ITR which no longer exists => crash

The issue is hit when the ntp_server_working takes too long (5 secs), ie longer than the time between entering the spoke and opening the NTP configuration dialog. For example when networking is configured later in the installation (like just before NTP configuration).

The crash showed with this
https://github.com/rvykydal/anaconda/commit/212eb1cf0b293f54e350c4bb0d666a5152cc0265
RVDBG debug info added:

[anaconda.crash.log](https://github.com/rhinstaller/anaconda/files/6172501/anaconda.crash.log)

If the `ntp_server_working` function is fast, eg network is activated early the refresh on spoke entry is fast enough to prevent the crash:

[anaconda.ok.log](https://github.com/rhinstaller/anaconda/files/6172503/anaconda.ok.log)

Seems to be related also to this change https://github.com/rhinstaller/anaconda/pull/2468
In upstream the code was completely rewritten in commit 15d2b2fb568df2c1a77cfb2baa703ae9f3da0f30

I think the safe fix for rhel8-branch would be checking the validity of iterator before using it as in this PR. In case of invalid itr there is no action so I think even the result should be correct.

We could probably also remove the model refresh (https://github.com/rhinstaller/anaconda/blob/e8714d7763f462deff4fc3f47bfb9fc5786fcae3/pyanaconda/ui/gui/spokes/datetime_spoke.py#L651) from spoke refresh method because the refreshed values are used/available (I think) only when the dialog is entered, and entering the dialog does reinitialization of the model anyway. Perhaps the refresh became useless (malign) after the https://github.com/rhinstaller/anaconda/pull/2468. I haven't found any reason for its existence but we should think twice I think.

VladimirSlavik (approved):
Looks good to me. Thank you! And sorry for breaking it.

I'm actually amazed that we keep an iterator in the asynchronous call instead of just the server hostname. But I guess it's understandable how this came to be.

poncovka (approved):
Looks good to me. Thanks!

jkonecny12 (approved):
LGTM! Great work!

JeongHyeon-Kim:
Thank for solving issue and record

PR #3254, opened 2021-03-22
Don't automatically execute the default partitioning

poncovka:
We should automatically execute only a partitioning that was already
configured by the user, for example with a kickstart file.


VladimirSlavik (approved):
Looks good to me. Thank you!

rvykydal (approved):
Looks good to me.

PR #3255, opened 2021-03-22
Replace use of imp with importlib in collect()

VladimirSlavik:
Draft because I am now having doubts if the complicated  stuff is actually happening at all.

~~TODO: Test with initial setup.~~

Tested - yes, it does. See further comments.

### DIFF
--- a/pyanaconda/core/util.py
+++ b/pyanaconda/core/util.py
@@ -33,6 +33,8 @@ import sys
 import types
 import inspect
 import functools
+import importlib.util
+import importlib.machinery
 import blivet.arch
 
 import requests
@@ -1179,8 +1181,6 @@ def collect(module_pattern, path, pred):
        :param pred: function which marks classes as good to import
        :type pred: function with one argument returning True or False
     """
-    import imp
-
     retval = []
     try:
         contents = os.listdir(path)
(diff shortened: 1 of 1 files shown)


poncovka (approved):

VladimirSlavik:
Tested with initial setup. The EULA spoke apparently needs the manual dance, although I am not exactly sure why. However, we can now say with certainty that it works and `imp` can be gone...

Review above still TODO.

pyanaconda/core/util.py
```diff
@@ -1201,12 +1201,16 @@ def collect(module_pattern, path, pred):
...
-        mod_info = None
         module = None
         module_path = None
 
         try:
-            (fo, module_path, module_flags) = imp.find_module(mod_name, [path])
+            # see what can be found
+            spec = importlib.machinery.PathFinder.find_spec(mod_name, [path])
+            module_path = spec.origin
+            found_ext = "." + spec.origin.rsplit(".", 1)[1]
```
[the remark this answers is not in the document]
VladimirSlavik: Thanks for the idea, changed.

VladimirSlavik:
Tested again, it works and all. Unless the refactoring is a must, I think this scratches the deprecation itch right now.

VladimirSlavik:
It would be nice to test this with kdump, oscap and other addons too, but never mind.

pyanaconda/core/util.py
```diff
@@ -1236,15 +1243,15 @@ def collect(module_pattern, path, pred):
...
+                    spec.loader.exec_module(module)
 
             # get the filenames without the extensions so we can compare those
             # with the .py[co]? equivalence in mind
             # - we do not have to care about files without extension as the
             #   condition at the beginning of the for loop filters out those
-            # - module_flags[0] contains the extension of the module imp found
-            candidate_name = module_path[:module_path.rindex(module_flags[0])]
+            # - found_ext contains the extension of the module importlib found
             loaded_name, loaded_ext = module.__file__.rsplit(".", 1)
```
poncovka, line 1255: I think that this will remove the dot from the extension.

poncovka (approved):

pyanaconda/core/util.py
VladimirSlavik: Yes! 3 lines below: `loaded_ext = "." + loaded_ext`

PR #3256, opened 2021-03-22
Enable Codecov

poncovka:
Analyze the code coverage with Codecov.

**TODO:**

- [ ] Enable for owners.
- [ ] Enable for contributors.
- [ ] Add a badge.

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,0 +1,38 @@
+codecov:
+  # Don't wait for all other statuses to pass.
+  require_ci_to_pass: false
+
+coverage:
+  # Shift the color range.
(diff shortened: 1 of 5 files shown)


VladimirSlavik (commented):
Looks good to me so far?

VladimirSlavik (commented):
Still LGTM. I like using the action instead of custom script/package.

I'm not sure if that helps anything, but... The checkboxes you have above are not really representative of the reality. Given how things are laid out these days, I would rather suggest: master/rawhide, branched fedora, rhel(s). And it's the last item where the owner/contributor split still exists.

PR #3258, opened 2021-03-23
network: match also connections named by MAC created by NM in initramfs

rvykydal:
Related: rhbz#1910438

For example connection created by configuration:
ip=10.43.136.52::10.43.136.254:255.255.255.0:testhost.example.com:52-54-00-8e-57-91:none

Same as in case of specification by BOOTIF= boot option, the connection is
bound to interface name and renamed to the interface name.

### DIFF
--- a/pyanaconda/core/regexes.py
+++ b/pyanaconda/core/regexes.py
@@ -194,3 +194,6 @@ IPV6_ADDRESS_IN_DRACUT_IP_OPTION = re.compile(r'\[[^\]]+\]')
 
 # Octet of MAC address
 MAC_OCTET = re.compile(r'[a-fA-F0-9][a-fA-F0-9]')
+
+# Name of initramfs connection created by NM based on MAC
+NM_MAC_INITRAMFS_CONNECTION = re.compile(r'^([0-9A-Fa-f]{2}:){5}([0-9A-Fa-f]{2})$')
--- a/pyanaconda/modules/network/initialization.py
+++ b/pyanaconda/modules/network/initialization.py
@@ -16,7 +16,9 @@
 # Red Hat, Inc.
 #
 import copy
+import re
 
+from pyanaconda.core.regexes import NM_MAC_INITRAMFS_CONNECTION
 from pyanaconda.modules.common.task import Task
 from pyanaconda.anaconda_loggers import get_module_logger
 from pyanaconda.modules.network.network_interface import NetworkInitializationTaskInterface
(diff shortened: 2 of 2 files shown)


pyanaconda/core/regexes.py
```diff
@@ -194,3 +194,6 @@
 
 # Octet of MAC address
 MAC_OCTET = re.compile(r'[a-fA-F0-9][a-fA-F0-9]')
+
+# Name of initramfs connection created by NM based on MAC
+NM_MAC_INITRAMFS_CONNECTION = re.compile(r'^([0-9A-Fa-f]{2}[:]){5}([0-9A-Fa-f]{2})$')
```
VladimirSlavik, line 199: What is the `[:]` part supposed to do?

VladimirSlavik (commented):
Looks good to me but can't ack due to total ignorance of the subject :(

pyanaconda/core/regexes.py
rvykydal: Well, I think there could be also just ':' - it is a MAC octet separator.

rvykydal: I've updated the patch '[:]' -> ':'

pyanaconda/core/regexes.py
```diff
@@ -194,3 +194,6 @@
 
 # Octet of MAC address
 MAC_OCTET = re.compile(r'[a-fA-F0-9][a-fA-F0-9]')
+
+# Name of initramfs connection created by NM based on MAC
+NM_MAC_INITRAMFS_CONNECTION = re.compile(r'^([0-9A-Fa-f]{2}:){5}([0-9A-Fa-f]{2})$')
```
poncovka, line 199: It would be good to have add some unit tests for this regex, that can be done in upstream.

poncovka (approved):
Looks good to me.

pyanaconda/core/regexes.py
jkonecny12: We really should have this abstracted somewhere. This is really bad.

jkonecny12 (approved):
LGTM!

pyanaconda/core/regexes.py
VladimirSlavik: Well, this *is* the abstraction, isn't it...

rvykydal:
PR with port to Fedora: https://github.com/rhinstaller/anaconda/pull/3264

PR #3260, opened 2021-03-23
Log into container registry early

VladimirSlavik:
The login is fast, and running it early saves some time when it fails. Without upload enabled by login, the rest of the workflow is useless anyway.

(Didn't test it - too trivial.)

### DIFF
--- a/.github/workflows/container-autoupdate.yml
+++ b/.github/workflows/container-autoupdate.yml
@@ -30,6 +30,9 @@ jobs:
       CI_TAG: '${{ matrix.container-tag }}'
     timeout-minutes: 60
     steps:
+      - name: Login to container registry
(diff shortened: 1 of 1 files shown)


jkonecny12 (changes requested):
Sorry but I don't agree it's a correct way to do. If we can't login it's most of the time problem of the repository. My problem with this is that the fast-fail login could cover another more significant error during the container build which we might be able to solve.

VladimirSlavik:
What is the end goal? To have a container? Because if so, then we're not having it, regardless of failure to build or failure to upload. Somebody will have to investigate anyway.

Also, this does not save humans, it saves electricity.

VladimirSlavik:
From discussion: Login failures will happen far less than build failures. This has very little chance of having any impact. That's an argument that makes sense to me too - closing.

PR #3261, opened 2021-03-23
Check source type directly, not via DBus

M4rtinK:
Not all payloads are available via DBus in RHEL 8 Anaconda,
so use the payload instance directly when checking payload type
in the Connect to Red Hat spokes.

This should fix the issue with RHV, which uses the liveimg payload,
which is not available via DBus on RHEL 8, leading to a crash.

Resolves: rhbz#1941578


M4rtinK:
A couple notes about this PR:
- I'm not entirely sure `self.payload.type` will work for payloads that *are* available over DBus
- is Fedora modularization far enough so all payloads are now accessible via DBus, making the fix not needed there ?

pyanaconda/ui/gui/spokes/subscription.py
```diff
@@ -871,8 +871,7 @@ def _register(self):
 
         # check if the current installation source will be overriden
         # and remember it if it is the case
-        source_proxy = self.payload.get_source_proxy()
-        source_type = source_proxy.Type
+        source_type = self.payload.type
```
poncovka, line 874: It should be `self.payload.source_type`.

M4rtinK: Fixed, thanks! :)

poncovka (approved):

VladimirSlavik (approved):
LGTM, thank you!

M4rtinK:
Updated the PR to handle the case of `self.payload.source_type` being `None`, which seems to be the case on the RHV installation image.

M4rtinK:
Re-tested the updated PR on latest RHEL 8.5 & RHV images and all seems to be fine as far as I can tell.

poncovka (approved):

PR #3262, opened 2021-03-25
repair  password entry box conflicts with  root lock button when enter root_password spoke

bitcoffeeiux:
The logic of the password entry box conflicts with that of the root lock button when enter the root_password spoke
Root is locked but the password_entry box can still be entered when enter the root password spoke
It's a confusing situation

### DIFF
--- a/pyanaconda/ui/gui/spokes/root_password.py
+++ b/pyanaconda/ui/gui/spokes/root_password.py
@@ -165,6 +165,8 @@ class PasswordSpoke(FirstbootSpokeMixIn, NormalSpoke, GUISpokeInputCheckHandler)
         self._root_password_ssh_login_override.set_active(
             self._users_module.RootPasswordSSHLoginAllowed
         )
+        self.password_entry.set_sensitive(not self._lock.get_active())
+        self.password_confirmation_entry.set_sensitive(not self._lock.get_active())
         if not self._lock.get_active():
             # rerun checks so that we have a correct status message, if any
             self.checker.run_checks()


M4rtinK (approved):
Finally got to this PR - sorry for the delay. :P

In any case, a nice improvement! :)

Before it looked like this on first root password spoke entry (both entries sensitive):
locking_without_fix
And with the fix it looks like this (no entry sensitive):
locking_fixed


Code also looks good & thanks for the fix! :)

jstodola:
Although this change fixes the behavior to be more logical, I would prefer not to merge it as it is.

The user that enters the root password spoke most probably wants to set the root password. The first thing he/she notices on the spoke is a text field that's disabled and he/she needs to figure out why. IMO, this is a regression in user experience. Perhaps moving the "Lock root account" checkbox above the root password text fields could improve it. 

Not a problem of this change, but it's related:
The label "Lock root account" is a bit misleading, since it disables the root password text fields. I might want to lock the root account, but set the password so that I can unlock the account later using `passwd --unlock`.

M4rtinK:
> Although this change fixes the behavior to be more logical, I would prefer not to merge it as it is.
> 
> The user that enters the root password spoke most probably wants to set the root password. The first thing he/she notices on the spoke is a text field that's disabled and he/she needs to figure out why. IMO, this is a regression in user experience. 
That's actually a good point - it's some time since the lock checkbox has been added but now that you mention it this way I seem to faintly remember keeping the fields active was likely intentional, to keep the existing behavior as much as possible.

>Perhaps moving the "Lock root account" checkbox above the root password text fields could improve it.
Yeah, that could help - also the SSH password login override checkbox does not really have anything common with it.

> 
> Not a problem of this change, but it's related:
> The label "Lock root account" is a bit misleading, since it disables the root password text fields. I might want to lock the root account, but set the password so that I can unlock the account later using `passwd --unlock`.
That's a good point as well.

Maybe we could combine the two things:
- move the checkbox above the fields
- when user types something in the fields show a warning that the password will be set, but login with the password would not be possible without unlocking the root account first

That could address both issues, I'm a bit worried about compilicating the UI a bit though.

jkonecny12:
Moving the checkbox above the fields seems a bit illogical to me. I wouldn't change the order of the components.

bitcoffeeiux:
Is there a more reasonable setting for this function?

M4rtinK:
> Is there a more reasonable setting for this function?
Hmm, maybe make fields always writeable, make the checkbox independent and show a warning dialog if user tries to exit spoke with password set but checkbox ticked ? That way it is consistent, users are not blocked from entering password when entering the spoke and it would be now possible to set passowrd and lock it, but not to do it by accident.

- If no password would be set by user, we would make sure to lock the account at install time. 
- If a password would be set but not explicitly locked, the root account will be created with that account and not locked.
- If a password is set in spoke and checkbox ticked, root password would be set to that value and then locked.

@bitcoffeeiux @jstodola @jkonecny12 What do you think ?

VladimirSlavik:
If the first point happens, maybe a label saying something like "empty password will cause root to be locked" would be good. It's not like we lack space on this spoke...

M4rtinK:
@mairin could you possibly take a look ? I think we might be getting a bit lost in the various options with any single on of them not being ideal, so I think feedback from an actual UX person is in order - and much appreciated! :)

mairin:
@M4rtinK Sure, happy to take a look. From reading the backscroll on this I see the following main UX issues:

**Spoke Name**
The name of the spoke is "Root password" but it deals with stuff beyond the root password. So the user will click on "root password" thinking it's a dialog to set root password and see other options they might not understand, or the user will want to disable SSH for root and not realize that is an option under "root password" and won't know where to look for it (although as a required spoke one would hope they'd hit it eventually! But maybe annoyed if they searched and didn't find it until they got here :) )

All of the options deal with security around the root user account. These also seem to be very specific to "root" and not say %wheel, so I would suggest the following rename of the spoke if that's in scope:

"Root account security"

Some of our documentation and cockpit refer to "Server administrator" privileges for %wheel, but it does appear cockpit treats root account specific stuff as "root" so I would keep this as "Root account security" 

**Mapping to real-life usage**

Martin you wrote up a very helpful list of the different things to be done on this screen, copying here:

- If no password would be set by user, we would make sure to lock the account at install time.
- If a password would be set but not explicitly locked, the root account will be created with that account and not locked.
- If a password is set in spoke and checkbox ticked, root password would be set to that value and then locked.

I'm going to flip this into a list of all possible outcomes (let me know if I am missing any) with a phrase to describe what the user's intention / story might be broken into 3 categories:

**No root**
1. "I don't want any root" - No root password, no ssh access, lock root

**Root, but later**
2. "I want to set up root, but lock it for later use" - Root password, ssh access, lock root
3. "I want to set up ssh-less root, but lock it for later use" - Root password, no ssh access, lock root

**Root, now**
4. "I want to set up root with remote access" - Root password, ssh access, no lock
5. "I want to set up root without remote access" - Root password, no ssh access, no lock
6. [Nonsense option] Root password, ssh access, lock root 

** Defaults**

What is the default setting if you choose nothing? Does this vary by edition/product? What is the base default - lock root, no ssh access? (That seems the most secure, not sure though?)


**Discussion**

I have a big question here on whether or not it makes sense to expose the "root, but later" cases in the anaconda UI. Some questions I have to hopefully help figure this out:

- Is setting up a root account with a password and locking it for later usage considered a good practice (for attended or unattended installs)? Is this referred to in any documentation or manuals or is generally known as a practice? Is this functionality part of some broader workflow or specific type of deployment?
- How common is this as a practice? Let's say 2 out of 1000 attended installation users wants this functionality (is that fair / generous enough?)
- Is it possible to access "root, but later" functionality in kickstart? (I am assuming **yes**?)
- Is it possible if you had to choose between lock root, or create password, that you could create the password, then on the running system run `passwd [something]` to lock the account immediately after?
- Is it worth complicating the logic of the screen for 998 people for the benefit of 2 (or whatever the number is)? 

**Initial UX Recommendation**

This is my initial recommendation for you based on my assumptions (that this isn't commonly done, even less so for attended installs). This is just a starting point for discussion, maybe my call is wrong here, hoping this is good food for thought for you regardless:

1. **I suggest we remove the ability to both lock the account and set a password.** 

The risk of allowing for both is that it complicates the screen for the end user reading it and trying to understand what their options are, and it also risks putting the system in a state the user doesn't understand ("I entered a password and clicked lock account, now I can't log in, I don't understand why I can't log in when I set a password. Why would it ask me for a password I couldn't use?" that sort of thing.)

2. **I would have the root account be locked by default, and have an "active root account" option that requires a password to be turned on.**

This is making the assumption that locked root by default is a good, secure default that would serve most attended install users. 

3. **I would keep ssh access on/off as an option, but only visible if the root account is activated.**

By default, ssh access for root should be turned off, and you only get the option to turn it on if you activated the root account.


OK here are some quick mockups to show how this could look. Again, this is fodder for discussion, we should talk through this to make sure it makes sense:

**Default - root acct locked**
opt1-01

**Root account enabled**
opt1-02

If you enable the root account, then password is a required field... If I recall correctly there's a way to display a banner at the bottom of the screen. Would it make sense if the user selected "enable root account" to put up a warning banner that they must enter a password to enable root, and maybe grey out "done" in the upper left until they fill out the password fields correctly?

VladimirSlavik:
I have one (minor) doubt about the texts on this mockup: Should this mention that making the "normal" user an administrator allows "most" things that root can do? It's a gross simplification, but I'd imagine it could provide some comfort. ("I'm not left powerless on my own machine.") Then again, more text is more reasons to not read it...

IIRC Anaconda won't allow locked root + not an admin, so technically there isn't a hidden footgun.

mairin:
@VladimirSlavik That sounds reasonable. Let me check how we talk about this in our documentation to see what specific language is used. Does the latest anaconda currently have a screen where the user can turn sudo on for their account or does that happen silently? (I don't remember/know how that is handled these days with the new first run stuff in Fedora where the users set up their account)

mairin:
@VladimirSlavik That sounds reasonable. Let me check how we talk about this in our documentation to see what specific language is used. Does the latest anaconda currently have a screen where the user can turn sudo on for their account or does that happen silently? (I don't remember/know how that is handled these days with the new first run stuff in Fedora where the users set up their account)

VladimirSlavik:
There's a checkbox "make this user an administrator" in "create user" which is a screen similar to this one.
Screenshot_vs_main_2021-05-12_17:03:54

mairin:
@VladimirSlavik what is the logic around the "Create User" screen appearing? Is that screen hidden from workstation installs for example (where the first run tool walks those users thru creating their user account?)

mairin:
@VladimirSlavik another question re: the "Create User" screen - is there logic here that affects remote access? E.g., if I turn off "require a password to use this account", does that mean remote access for my account is turned off? If I turn on "require a password to use this account", is remote access then enabled? Or is remote access unaffected (turned off by default?)

I *think* Fedora workstation doesnt have sshd running by default?

VladimirSlavik:
Sorry, I don't know - probably @M4rtinK does?

M4rtinK:
> @VladimirSlavik what is the logic around the "Create User" screen appearing? Is that screen hidden from workstation installs for example (where the first run tool walks those users thru creating their user account?)

On Fedora Workstation installs we don't handle user creation at all and IIRC root account is locked by default. The user account creation is then handled on first boot by the Gnome Initial Setup tool.

M4rtinK:
> @VladimirSlavik another question re: the "Create User" screen - is there logic here that affects remote access? E.g., if I turn off "require a password to use this account", does that mean remote access for my account is turned off? If I turn on "require a password to use this account", is remote access then enabled? Or is remote access unaffected (turned off by default?)

So I've installed two Rawhide VMs, one with user that has a password set and one that has a user with "password not required".

The one with password set behaves as expected - you can login as that user locally with the set password. You can also login as the user with the password over SSH.

In the user-with-no-password case, I can login as the user locally and no password is requested. When trying to login via SSH it looks like sshd is running & prompts me for password for the user. So basically it looks like sshd runs always when installing from the netinst image, but we indeed not allow password less login via SSH as we aren't _insane_.

Also in both cases it is not possible to login over SSH with the root password as that's the default behaviour in Fedora since ~2019.

M4rtinK:
> @M4rtinK Sure, happy to take a look. From reading the backscroll on this I see the following main UX issues:
> 
> **Spoke Name**
> The name of the spoke is "Root password" but it deals with stuff beyond the root password. So the user will click on "root password" thinking it's a dialog to set root password and see other options they might not understand, or the user will want to disable SSH for root and not realize that is an option under "root password" and won't know where to look for it (although as a required spoke one would hope they'd hit it eventually! But maybe annoyed if they searched and didn't find it until they got here :) )
> 
> All of the options deal with security around the root user account. These also seem to be very specific to "root" and not say %wheel, so I would suggest the following rename of the spoke if that's in scope:
> 
> "Root account security"
> 
> Some of our documentation and cockpit refer to "Server administrator" privileges for %wheel, but it does appear cockpit treats root account specific stuff as "root" so I would keep this as "Root account security"

Yeah, that kinda makes sense, especially as long as the SSH override checkbox is there. We are actually considering to remove the override, as it was originally planned as a temporary migration aid:

https://fedoraproject.org/wiki/Changes/Drop_Rootpw_SSH_From_Installer

This is still being discussed (with possible quite interesting proposals from an UX perspective for easy public key injection at installation time) and FESCO has also not yet decided about it. At the moment it's possible the SSH override option either stays, is dropped or stays on some Fedora variants that need it.

> 
> **Mapping to real-life usage**
> 
> Martin you wrote up a very helpful list of the different things to be done on this screen, copying here:
> 
>     * If no password would be set by user, we would make sure to lock the account at install time.
> 
>     * If a password would be set but not explicitly locked, the root account will be created with that account and not locked.
> 
>     * If a password is set in spoke and checkbox ticked, root password would be set to that value and then locked.
> 
> 
> I'm going to flip this into a list of all possible outcomes (let me know if I am missing any) with a phrase to describe what the user's intention / story might be broken into 3 categories:
> 
> **No root**
> 
>     1. "I don't want any root" - No root password, no ssh access, lock root
> 
> 
> **Root, but later**
> 2. "I want to set up root, but lock it for later use" - Root password, ssh access, lock root
> 3. "I want to set up ssh-less root, but lock it for later use" - Root password, no ssh access, lock root
> 
> **Root, now**
> 4. "I want to set up root with remote access" - Root password, ssh access, no lock
> 5. "I want to set up root without remote access" - Root password, no ssh access, no lock
> 6. [Nonsense option] Root password, ssh access, lock root
> 
> **Defaults**
> 
> What is the default setting if you choose nothing? Does this vary by edition/product? What is the base default - lock root, no ssh access? (That seems the most secure, not sure though?)

Yep, that's the default. We always locked the root account if no password was provided and password based root login over SSH was disabled in 2019 when Fedora finally dropped it's downstream patches restoring OpenSSH default upstream behaviour.

> 
> **Discussion**
> 
> I have a big question here on whether or not it makes sense to expose the "root, but later" cases in the anaconda UI. Some questions I have to hopefully help figure this out:
> 
>     * Is setting up a root account with a password and locking it for later usage considered a good practice (for attended or unattended installs)? Is this referred to in any documentation or manuals or is generally known as a practice? Is this functionality part of some broader workflow or specific type of deployment?
> 
>     * How common is this as a practice? Let's say 2 out of 1000 attended installation users wants this functionality (is that fair / generous enough?)
> 
>     * Is it possible to access "root, but later" functionality in kickstart? (I am assuming **yes**?)
> 
>     * Is it possible if you had to choose between lock root, or create password, that you could create the password, then on the running system run `passwd [something]` to lock the account immediately after?
> 
>     * Is it worth complicating the logic of the screen for 998 people for the benefit of 2 (or whatever the number is)?

I agree, IMHO this is not really necessary - I know @jstodola suggested, but I'm not sure if he had some concrete use case in mind or was just mentioning it as a possible state in the overall state matrix of the spoke. IIRC I never saw a bug or RFE filled for Anaconda not supporting this, unless I messed something.

> 
> 
> **Initial UX Recommendation**
> 
> This is my initial recommendation for you based on my assumptions (that this isn't commonly done, even less so for attended installs). This is just a starting point for discussion, maybe my call is wrong here, hoping this is good food for thought for you regardless:
> 
>     1. **I suggest we remove the ability to both lock the account and set a password.**

I think we don't actually support that via the GUI at the moment, so that makes things easier. :)

> 
> 
> The risk of allowing for both is that it complicates the screen for the end user reading it and trying to understand what their options are, and it also risks putting the system in a state the user doesn't understand ("I entered a password and clicked lock account, now I can't log in, I don't understand why I can't log in when I set a password. Why would it ask me for a password I couldn't use?" that sort of thing.)
> 
>     1. **I would have the root account be locked by default, and have an "active root account" option that requires a password to be turned on.**
> 
> 
> This is making the assumption that locked root by default is a good, secure default that would serve most attended install users.
> 
>     1. **I would keep ssh access on/off as an option, but only visible if the root account is activated.**

Sounds good - provided we keep the SSH option.

> 
> 
> By default, ssh access for root should be turned off, and you only get the option to turn it on if you activated the root account.

Yep, that nicely maps to the current default. :)

> 
> OK here are some quick mockups to show how this could look. Again, this is fodder for discussion, we should talk through this to make sure it makes sense:
> 
> **Default - root acct locked**
> opt1-01
> 
> **Root account enabled**
> opt1-02
> 

I like this! :) It provides useful context for what the "root" thing is about & nicely makes use of all the otheriwise empty space. :)

> If you enable the root account, then password is a required field... If I recall correctly there's a way to display a banner at the bottom of the screen. Would it make sense if the user selected "enable root account" to put up a warning banner that they must enter a password to enable root, and maybe grey out "done" in the upper left until they fill out the password fields correctly?

The warning banner is already in use on the spoke to notify the user about password quality issues, but as far as I can tell this should not clash with it. Also we already do trap the user in as spoke in some case, so doing that should not really be a problem as well. :)

mairin:
@M4rtinK @VladimirSlavik Awesome thank you for the investigation / analysis. So it looks like maybe the initial UX recommends I made would make sense, the only outstanding thing I am worried about from a UX perspective is what you pointed out above @M4rtinK re: having a user with sudo:

>  Gnome Initial Setup screens for user configuration. There is no wheel/sudo/admin mentioned anywhere but it looks like they setup the used with sudo by default (as they kinda have to with root locked by default). I wonder if maybe we should do that as well if root is locked ? At the moment we have some quite complex logic that checks root or admin user has been created and prevents installation start if not. If we just always make user an admin if root is created maybe we could get ridd of that & the admin checkbox? Or hit another edge case. :D

So this would impact non-Workstation variants, right? I think there are a few cases to think about:

1. Non-workstation single-user attended install hardware or virtual, **installed by person meaning to run it**

In this case, the person performing the installation is the ultimate user of the system. They will know what the username and password for the first non-root uid 1000/5000 whatever user will be, and sudo could be turned on for it, so no need for a root account.

2. Non-workstation, multi-user attended install on hardware, **installed by one of the people meaning to run it**

For example, a server that is used / shared by a team. The person installing it may be one of the team members. I think this still falls under the same treatment for #1, because this user will know what their username will be, and can grant their usename sudo access and post-installation can set up the other team member accounts with appropriate access. There isn't a real need for a root acct in this scenario I think?

3. Non-workstation, multi-user attended install to a VM, **installed by one of the people meaning to run it**

This one might be tricky. For example, a development team could have a VM that is configured a particular way for testing, let's say it has some kind of browser testing suite, and they only have the license for one system or something like that. They throw this system together, and maybe the expectation is that everyone on the team logs in using the same account. "root" is something rather universally known, whereas if the system has a single non-root user with admin privs, what would it be named? Is it something easily guessable? How would they share the username? (I guess they can share the password if it was root - they might have a standard team password for such accounts though, so I guess they could also share a standard team username?) So there really isn't a need for a root account here either, they just have to standardiz / socialize the username on their team.

4. Non-workstation, single or multi-user attended install, **installed by someone who will then pass it onto someone else** (e.g. someone manually loading a CSB on systems for distribution)

This one is where I am the most confounded. I think mostly this scenario is something that is automated via kickstart, but I could see it potentially being a one-off scenario where a particular group of users need systems that are customized beyond what a standard image offers and it is determined better to one-off their systems. Or maybe it's meant to be a test system. Maybe the security policy is in this case you dont create a root account so root is locked, but you don't yet know what the username will be because the system hasn't been handed off yet. Does firstboot still exist for this scenario? 

So I think because of scenario #4, you couldn't make it so if root is locked, you do not allow people to proceed unless they create a first user and add sudo access to it because the scenario requires that the first account be created later. I think this scenario is why firstboot existed in the first place?

Is scenario 4 still valid, or is the correct path for that scenario to use kickstart and we don't support it via attended install?

M4rtinK:
> @M4rtinK @VladimirSlavik Awesome thank you for the investigation / analysis. So it looks like maybe the initial UX recommends I made would make sense, the only outstanding thing I am worried about from a UX perspective is what you pointed out above @M4rtinK re: having a user with sudo:
> 
> > Gnome Initial Setup screens for user configuration. There is no wheel/sudo/admin mentioned anywhere but it looks like they setup the used with sudo by default (as they kinda have to with root locked by default). I wonder if maybe we should do that as well if root is locked ? At the moment we have some quite complex logic that checks root or admin user has been created and prevents installation start if not. If we just always make user an admin if root is created maybe we could get ridd of that & the admin checkbox? Or hit another edge case. :D
> 
> So this would impact non-Workstation variants, right? I think there are a few cases to think about:
> 
>     1. Non-workstation single-user attended install hardware or virtual, **installed by person meaning to run it**
> 
> 
> In this case, the person performing the installation is the ultimate user of the system. They will know what the username and password for the first non-root uid 1000/5000 whatever user will be, and sudo could be turned on for it, so no need for a root account.
> 
>     1. Non-workstation, multi-user attended install on hardware, **installed by one of the people meaning to run it**
> 
> 
> For example, a server that is used / shared by a team. The person installing it may be one of the team members. I think this still falls under the same treatment for #1, because this user will know what their username will be, and can grant their usename sudo access and post-installation can set up the other team member accounts with appropriate access. There isn't a real need for a root acct in this scenario I think?

Yeah - and as long as they have access to the sudo user, they can either set password to the root account later on or add more users to the wheel group.

> 
>     1. Non-workstation, multi-user attended install to a VM, **installed by one of the people meaning to run it**
> 
> 
> This one might be tricky. For example, a development team could have a VM that is configured a particular way for testing, let's say it has some kind of browser testing suite, and they only have the license for one system or something like that. They throw this system together, and maybe the expectation is that everyone on the team logs in using the same account. "root" is something rather universally known, whereas if the system has a single non-root user with admin privs, what would it be named? Is it something easily guessable? How would they share the username? (I guess they can share the password if it was root - they might have a standard team password for such accounts though, so I guess they could also share a standard team username?) So there really isn't a need for a root account here either, they just have to standardiz / socialize the username on their team.

Yeah, if they have to pass a password, they might as well pass a username + password pair.

> 
>     1. Non-workstation, single or multi-user attended install, **installed by someone who will then pass it onto someone else** (e.g. someone manually loading a CSB on systems for distribution)
> 
> 
> This one is where I am the most confounded. I think mostly this scenario is something that is automated via kickstart, but I could see it potentially being a one-off scenario where a particular group of users need systems that are customized beyond what a standard image offers and it is determined better to one-off their systems. Or maybe it's meant to be a test system. Maybe the security policy is in this case you dont create a root account so root is locked, but you don't yet know what the username will be because the system hasn't been handed off yet. Does firstboot still exist for this scenario?

Yes, it's called Initial Setup these days and effectively works as a lightweight launcher for the same screens used in Anaconda. It's not used by default on the Workstation image but on all the other Fedora spins (KDE, XFCE, etc.) if you don't create a user account, Initial Setup will start and show the user configuration screen.

But I would say the main usage scenario for Initial Setup at the moment is the (ARM) image deployment. Basically, users will put a Fedora image to an SD card via the [Fedora ARM Installer](https://fedoraproject.org/wiki/Architectures/ARM/Installation) and then put that SD card into their ARM board (say a raspberry pi) and boot it. As all the images are the same, users need a way to create users, set the root password, configure timezones and keyboards, setup the network, etc.

This is where Initial Setup comes to save the day, running in the so called reconfig mode where all relevant spokes (basically anything other than storage and software selection related spokes) are shown. This works both in graphical mode (on ARM boards that support graphics on Fedora) as well as in text mode.

> 
> So I think because of scenario #4, you couldn't make it so if root is locked, you do not allow people to proceed unless they create a first user and add sudo access to it because the scenario requires that the first account be created later. I think this scenario is why firstboot existed in the first place?

Yep, that was AFAIK likely the main usecase for Initial Setup, back before the ARM board use case surfaced. IIRC we called it "OEM installs/provisioning". One of the scenarios was companies shipping computers with Fedora/RHEL pre-installed in such a way that they boot into Initial Setup once turned on, so that they user can set things up to their liking. The other related scenario was university IT either handing out pre-configured machines to students/faculty members they support or alternatively telling them how to start such a pre-configured installation (via PXE/from a pre-configured image).

There are actually provisions for it in kickstart at installation time, so that if the root password is set in kickstart, it can't be changed or disaplayed at installation time. I assume (this true origin being kinda lost to the sands of time) this is so as to enable later access to the local IT who knows the root password while allowing the users to setup the rest of the system.

> 
> Is scenario 4 still valid, or is the correct path for that scenario to use kickstart and we don't support it via attended install?

I think it's valid but it was always assumed to be handled via kickstart - either via the root-password-set-in-kickstart trick or via enabling the firstboot reconfig mode ("firstboot --enable --reconfig" and handing out such preconfigured machines booting to IS to users). It's very unlikely this would be done by some poor soul manually for any reasonably big group of users.

And if you *do* have someone doing it manually, IMHO they have options how to manually prepare the system after the installation, without us overcomplicating the Anaconda UI.

So agreed, I don't think we need to handle this usecase in a GUI-only, no-kickstart workflow at all.

mairin:
@M4rtinK OK cool, so then moving forward the idea would be to have root disabled by default on this newly renamed "root account security" screen, and on the create user screen, have the "make this user administrator" turned on by default? 

> Screenshot_vs_main_2021-05-12_17:03:54

The last thing I want to check with you... would the user creation spoke be required then, and the root account security screen not be required (since the default - disabled root account - is the most secure option and won't break anything?)

M4rtinK:
> @M4rtinK OK cool, so then moving forward the idea would be to have root disabled by default on this newly renamed "root account security" screen, and on the create user screen, have the "make this user administrator" turned on by default?
> 
> > Screenshot_vs_main_2021-05-12_17:03:54
> 
Yep, pretty much that's the idea. Using your new design for the root password spoke & making the normal user an admin by default. 

We would still need to make sure *an* admin level account is created but IMHO it would be much easier for users to achieve that without multiple roundtrips between spokes and the hub to find out what's wrong *now* or if they can finally start the installation.

And also it would make the experience on the spins & netinst images more consistent with Fedora workstation which *does* create it's user as an admin account by default.

@poncovka @jkonecny12 @bitcoffeeiux @jstodola What do you think ? Any possible pitfalls or conflicting use cases I might be missing ? :P

> The last thing I want to check with you... would the user creation spoke be required then, and the root account security screen not be required (since the default - disabled root account - is the most secure option and won't break anything?)

That's what I would expect - after entering the summary hub, the user creation spoke would be required (thus marked red). The user enter the user creation spoke, enters username and password, clicks done. 

As "Make this user an administrator" checkbox would be ticked by default, the condition that there needs to be at least one admin level account would be satisfied and both root and user configuration spokes would be marked as completed and would not block the start of the installation (not red anymore).

It would also still be possible to create a non-admin user + set root password. Just set root password and then create user with "Make this user an administrator" unchecked. Or the other way around.

bitcoffeeiux:
Do we have any new plans?

jkonecny12:
Just to be clear, would it be possible to create just `root` and avoid having user account completely?

jkonecny12:
Also I can't help myself but I see a big inconsistency between the new Root spoke and the old User Spoke. Do I understand it correctly that this is the solution we want to go now?

VladimirSlavik:
Closed in favor of #3511 which resolves the ambiguity.

PR #3264, opened 2021-03-25
network: match also connections named by MAC created by NM in initramfs

rvykydal:
This is port from rhel-8 branch.

Related: rhbz#1910438

For example connection created by configuration:
ip=10.43.136.52::10.43.136.254:255.255.255.0:testhost.example.com:52-54-00-8e-57-91:none

Same as in case of specification by BOOTIF= boot option, the connection is
bound to interface name and renamed to the interface name.


poncovka (approved):
Looks good to me. Thanks!

jkonecny12 (approved):
Looks good to me.

PR #3266, opened 2021-03-27
Master add workflow environments for contributors

jkonecny12:
Use GitHub [environments](https://docs.github.com/en/actions/reference/environments) feature for our gated PR tests. This will simplify our workflow solution and testing a lot. In general it means that selected members have to approve the execution. If this is not done the workflow will expire in 30 days.

To see how it works look here.

This is how the contributor sees it:
Screenshot from 2021-03-27 16-23-09
Screenshot from 2021-03-27 16-24-04

This is how owner sees it:
Screenshot from 2021-03-27 16-24-24
Screenshot from 2021-03-27 16-25-08

As you can see it's pretty nice solution which will stop the execution in a nice way and we don't have to have more than just one line in our workflow.

**Benefits:**
* It's clean.
* We can avoid using our testing organizations and just use our forks. We can believe that the one line shouldn't be that hard to break and thanks to the drawback below we would see that immediately if that happens. Done in the screenshots above.
* We can avoid duplication in our workflow. Basically we can use the same workflow for contributors as for owners.

**Drawbacks:**
* Problem of this solution is that it's mandatory for everyone. Even when you'll create PR you need to approve it. Unfortunately, I wasn't able to find a way how to solve this.

**Note:**
In this PR I've migrated our Rawhide tests and our kickstart tests. RHEL tests should change on ``rhel-8`` branch. 

The kickstart tests are a bit problematic because they are started by comment which is not attached to the PR itself. That means that you have to go to the `Actions` tab and find your run to be able to approve it. To solve this I've introduced commenting on the PR. After initial comment processing work is done (2 - 5 seconds) you will get comment on your PR requesting you to approve the run with the link.

Screenshot from 2021-03-27 16-38-48


jkonecny12:
Example of the run of KS tests on my fork repository https://github.com/jkonecny12/anaconda/pull/10 .

VladimirSlavik (approved):
I think it's worth it. The friction added to tests isn't that much.

PR #3269, opened 2021-03-29
Make the scale visible by default (#1943370)

poncovka:
GLib crashes with a segmentation fault if the accessibility
is enabled and a GtkScale widget is not visible by default.
We will hide the widget later if required.

Turn off wrapping of the scale values. There is not enough
space for more than one line above the scale.

Resolves: rhbz#1943370

### DIFF
--- a/pyanaconda/ui/gui/spokes/lib/resize.glade
+++ b/pyanaconda/ui/gui/spokes/lib/resize.glade
@@ -289,10 +289,9 @@
               <object class="GtkScale" id="resizeSlider">
                 <property name="visible">True</property>
                 <property name="can_focus">True</property>
+                <property name="margin_top">12</property>
                 <property name="margin_bottom">12</property>
                 <property name="adjustment">resizeAdjustment</property>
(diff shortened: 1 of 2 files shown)


poncovka:
Before:
scale_01

After:
scale_02

jstodola:
Tested on RHEL using [anabot](https://pagure.io/anabot) and [minimal.xml](https://pagure.io/anabot/blob/main/f/examples/minimal.xml) recipe, which was previously failing due to anaconda crash.
It's now possible to go through the installation and reclaim disk space.
Thanks!

rvykydal (approved):
Looks good to me, thank you!

jkonecny12 (approved):
LGTM!

VladimirSlavik (approved):
Thank you! Looks good to me.

Does this work well also when the scale is at its end?

poncovka:
> Does this work well also when the scale is at its end?

It does. It looks the same as the mark below.

PR #3270, opened 2021-03-29
Rhel8 pylint updates

VladimirSlavik:
Backport of #3200.

Not sure yet what to do wrt. bug(s) needed.

### DIFF
--- a/pyanaconda/core/configuration/storage.py
+++ b/pyanaconda/core/configuration/storage.py
@@ -38,7 +38,7 @@ class PartitioningScheme(Enum):
     def from_name(cls, value):
         """Convert the given value into a partitioning scheme."""
         try:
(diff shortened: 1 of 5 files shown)


VladimirSlavik:
I think this might indirectly prove (together with other PRs) that if the unit tests for RHEL succeed, the success propagates correctly, but somehow, a failure does not and results in a huge timeout. If that was true, then this superficially "fixes" that, as pylint made the whole suite always fail. It's not a real fix because the error is only speculated (yet), cause is not known, etc.

poncovka (commented):
Looks good to me, but I think @jkonecny12 suggested to freeze the pylint version on this branch.

VladimirSlavik:
True. I wanted to test this - it took me just a few minutes.

PR #3271, opened 2021-03-29
Test on kstest runners (rhel)

VladimirSlavik:
With this, unit tests run mostly identically to rpm tests.

~~Maybe. TODO: test this.~~


VladimirSlavik:
@jkonecny12 should this be ported to master somehow? I think not, since it conflicts with the workflows there...

jkonecny12:
I don't think we can port this. It's owners PR. Similar logic should be used for owners-tests on master if not already there but not porting this PR.

PR #3273, opened 2021-03-31
Add the default_environment property to the DNF manager

poncovka:
Choose the environment from the Anaconda configuration file
or the first available one.

Use the DNF manager to select the default environment in UI.


rvykydal (approved):
Looks good to me.

PR #3276, opened 2021-03-31
Pin version of astroid

VladimirSlavik:
The astroid library is used by pylint and can change its behavior by changing the parse tree. Pinning astroid version prevents all new detections since the pinning of pylint (in September 2020).


poncovka (approved):
Thanks!

PR #3280, opened 2021-04-06
Run RHEL8 contributor tests on kstest runners

VladimirSlavik:
Uses same code as owner tests, mostly.

With this, we should be able to get rid of the rhel-specific runner, this time for real.

~~TODO: Still testing. Ongoing PSI networking instability incident = downloading packages in the runner times out before the job can succeed. See here: https://github.com/vslavik-test-org/anaconda/pull/6~~

It works fine, see here: https://github.com/vslavik-test-org/anaconda/actions/runs/725452092


jkonecny12 (approved):
Great improvement, thanks for giving a finish blow! :)

VladimirSlavik:
COPR builds are still on these runners, so not the end, I'm afraid... but it's close!

PR #3282, opened 2021-04-07
Use the DNF manager instead of the DNF base

poncovka:
Extend the DNF manager, so it can be used instead of the DNF base.


M4rtinK (approved):
Looks good to me. :)

VladimirSlavik (approved):
Looks good to me. Thank you!

jkonecny12 (approved):
Looks really great! Thanks a lot for creating the DNF Manager the code is much cleaner!

poncovka:
Rebased.

PR #3286, opened 2021-04-09
Set up the Storage spoke in TUI on every entry

poncovka:
The setup method of the Storage spoke needs to be run every time
we enter the spoke to update its state. Use a workaround for now.


jkonecny12 (approved):
Thanks a lot for solving this. I'll try to add this functionality to simpleline 'soon'.

PR #3287, opened 2021-04-09
Use new fedora-latest alias in COPR

jkonecny12:
One more think we don't have to think about.

https://github.com/packit/packit-service/issues/986#event-4535281824


poncovka (approved):

VladimirSlavik (approved):
Thanks!

PR #3288, opened 2021-04-09
Remove patches to avoid forcing inst. prefix on RHEL 9

jkonecny12:
It's safe to remove this from Rawhide.
We still should wait for removing this from RHEL.


poncovka (approved):
Looks good to me!

M4rtinK (approved):

VladimirSlavik (approved):

PR #3289, opened 2021-04-12
Make sure rhsm.service is running at Anaconda startup

poncovka:
Make sure rhsm.service is running at Anaconda startup
to avoid issues with DBus activation of rhsm.service
timing out on systems that are slow or under heavy load.

We need both Wants to specify the anaconda.target requires
the rhsm.service as well as After to make sure it is started
before all the Anaconda services that just have Wants but
not After.

If the rhsm.service fails to start or it cannot be added to
the transaction, anaconda will start anyway.

(cherry picked from commit 1cd68dd)

Resolves: rhbz#1948371

### DIFF
--- a/data/systemd/anaconda.target
+++ b/data/systemd/anaconda.target
@@ -13,3 +13,5 @@ Wants=anaconda-sshd.service
 Wants=anaconda-pre.service
 Wants=anaconda-fips.service
 Wants=systemd-logind.service
+Wants=rhsm.service
+After=rhsm.service


jkonecny12 (approved):
Looks great to me. Thanks!

M4rtinK (approved):
Looks good to me as well, thanks!

VladimirSlavik (approved):
Looks good to me. Thank you!

poncovka:
Ported in https://github.com/rhinstaller/anaconda/pull/3333.

PR #3290, opened 2021-04-12
Allow to use no default product configuration (#1947939)

poncovka:
Don't raise an exception if we are not able to find a suitable product
configuration for the detected product name. The installer is able to
continue without it. Issue a warning instead.

Resolves: rhbz#1947939


jkonecny12 (approved):
LGTM!

VladimirSlavik (approved):
Thanks!

PR #3291, opened 2021-04-12
Run RHEL COPR builds on kstest

VladimirSlavik:
Requires a new container with the certificates, and some changes to workers. See also the [internal gitlab PR in builders](https://gitlab.cee.redhat.com/rhinstaller/builders/-/merge_requests/67) for the changes there.

Successful run at https://github.com/vslavik-test-org/anaconda/runs/2326194094?check_suite_focus=true

### DIFF
--- a/.github/workflows/daily-rhel-copr.yml
+++ b/.github/workflows/daily-rhel-copr.yml
@@ -24,33 +24,9 @@ jobs:
         run: |
           sudo podman build --no-cache -t ${{ env.CI_CONTAINER }} anaconda/dockerfile/anaconda-rhel-copr
 
-      - name: Clone copr-builder repository
-        uses: actions/checkout@v2
-        with:
-          repository: vojtechtrefny/copr-builder
-          fetch-depth: 0
-          path: copr-builder
-
-      - name: Create copr-builder configuration
-        run: |
-          cat <<EOF > copr-builder/copr-builder-rhel.conf
-          [anaconda]
-          copr_user = rhinstaller-group
-          copr_repo = Anaconda
-          package = anaconda
-          git_url = https://github.com/rhinstaller/anaconda
-          git_branch = rhel-8
-          pre_archive_cmd = ./autogen.sh && ./configure
-          archive_cmd = make po-pull && make dist
-          EOF
-
-      - name: Copy token
-        run: |
-          cp /home/github/copr-rhel-token copr-builder/copr-rhel-token
-
       - name: Run copr-builder
         run: |
-          sudo podman run -i --rm --privileged -v `pwd`/copr-builder:/copr-builder:ro ${{ env.CI_CONTAINER }} <<EOF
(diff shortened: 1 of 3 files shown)


.github/workflows/daily-rhel-copr.yml
```diff
@@ -29,5 +44,13 @@ jobs:
           archive_cmd = make po-pull && make dist
           EOF
 
+      - name: Copy token
+        run: |
+          cp /home/github/copr-rhel-token copr-builder/copr-rhel-token
```
jkonecny12, line 49: I would remove this step completely. It's not necessary.

.github/workflows/daily-rhel-copr.yml
```diff
@@ -29,5 +44,13 @@ jobs:
...
+
       - name: Run copr-builder
-        run: ./copr-builder -v -c copr-builder-rhel.conf -C /copr-rhel
+        run: |
+          sudo podman run -i --rm --privileged -v `pwd`/copr-builder:/copr-builder:ro ${{ env.CI_CONTAINER }} <<EOF
```
jkonecny12, line 53: Please use `-v ~/copr-rhel-token:...` instead. Also add `:Z` to the volume mount.

.github/workflows/daily-rhel-copr.yml
```diff
@@ -9,16 +9,31 @@ on:
...
+      - name: Build the build container
+        run: |
+          sudo podman build --no-cache -t ${{ env.CI_CONTAINER }} anaconda/dockerfile/anaconda-rhel-copr
+
+      - name: Clone copr-builder repository
         uses: actions/checkout@v2
         with:
           repository: vojtechtrefny/copr-builder
+          fetch-depth: 0
+          path: copr-builder
```
jkonecny12, line 32: I think moving the copr builder into the container will be better solution. At the end it's a dependency and also it will make debugging in the future simpler.

jkonecny12 (changes requested):
Looks good to me. I just have a few points for improvement.

.github/workflows/daily-rhel-copr.yml
VladimirSlavik: I don't think I can mount a file like that?

VladimirSlavik:
I have pushed as another (temporary) commit a version that addresses the feedback above, but can't test due to another outage.

.github/workflows/daily-rhel-copr.yml
jkonecny12: Based on my tests yes, you can :).

jkonecny12 (commented):
The new version looks great to me :)

VladimirSlavik:
Finally tested, works OK: https://github.com/vslavik-test-org/anaconda/runs/2400155738?check_suite_focus=true

VladimirSlavik:
Spoke too soon, this depends on changing the worker machines, so still blocked.

VladimirSlavik:
Workers renewed, let's have this!

PR #3292, opened 2021-04-13
Document the issue about the ignoredisk command and installation sources

poncovka:
The installer fails to find an installation media on the USB drive if the
`ignoredisk --only-use=` command is specified in a kickstart file.


VladimirSlavik (approved):
Looks good to me. Thank you!

PR #3295, opened 2021-04-16
Add a shim script to translate signals from Xorg

VladimirSlavik:
Yet another PR for the X thing, after #3107, #3132, and #3141. Thanks to @bitcoffeeiux and @poncovka.

This adds a shim script to translate the signals to make sure they are no longer the same: `SIGUSR1` - > `SIGRTMIN+8` (most probably 42). The non-responsive Xorg is left to be.

Resolves: [rhbz#1918702](https://bugzilla.redhat.com/show_bug.cgi?id=1918702)

### DIFF
--- a/scripts/anaconda-xorg-shim
+++ b/scripts/anaconda-xorg-shim
@@ -25,6 +25,7 @@ start_message = "Xorg signal shim started by {} as {}.".format(os.getppid(), os.
 subprocess.run(["systemd-cat"], input=start_message.encode("ascii"), check=True)
 params = ["/bin/Xorg"] + sys.argv[1:]
 old_handler = signal.signal(signal.SIGUSR1, handler)
+import time; time.sleep(10)
 process = subprocess.Popen(params, preexec_fn=preexec)  # pylint:disable=subprocess-popen-preexec-fn
 process.wait()
 signal.signal(signal.SIGUSR1, old_handler)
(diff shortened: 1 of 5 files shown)


bitcoffeeiux:
lgtm~

scripts/anaconda-xorg-shim
```diff
@@ -0,0 +1,29 @@
...
+    signal.signal(signal.SIGUSR1, signal.SIG_IGN)
+
+
+message = "Xorg signal shim started by {} as {}.".format(os.getppid(), os.getpid())
+subprocess.run(["systemd-cat"], input=message.encode("ascii"))
+params = ["/bin/Xorg"] + sys.argv[1:]
+old_handler = signal.signal(signal.SIGUSR1, handler)
+process = subprocess.Popen(params, preexec_fn=preexec)
+process.wait()
+signal.signal(signal.SIGUSR1, old_handler)
```
VladimirSlavik, line 29: I'm not sure if keeping the handler is the right thing to do in the shim script. The alternative would be `SIG_IGN`.

VladimirSlavik:
This should be the bash alternative for the script, except it does not work.
```
#!/usr/bin/bash

function handler() {
  echo "Xorg signal shim received SIGUSR1 from Xorg, passing on as SIGRTMIN+8." | systemd-cat
  kill -s 42 $PPID
}

echo "Xorg signal shim started by $PPID as $$." | systemd-cat
trap handler SIGUSR1
/bin/Xorg "$@"
wait $!
```

VladimirSlavik:
There is one more problem, if Xorg starts, it switches to tty6 anyway. Ugh. This just refuses to die.

VladimirSlavik:
Superseded by #3304.

PR #3297, opened 2021-04-16
Fix RPMOSTree mount to the non-existing directory (#1906735)

jkonecny12:
User defined mount points doesn't have directories created automatically which will result in a crash during the installation process.

Resolves: rhbz#1906735


VladimirSlavik:
Thank you! I'm not sure, but might this also fix stuff like mounting disks under `/data` and such? If so, there are more bugs filed for this.

VladimirSlavik (approved):
Looks good to me. Thank you!

jkonecny12:
**UPDATED:**
* Adapt existing tests and add one for dir creation.

poncovka (approved):

PR #3279, opened 2021-04-06
subscription: allow dates in ISO 8601 format

ptoscano:
Right now, the date returned by D-Bus method calls of subscription-manager have a locale-specific format, which can be hard to
parse. A planned change is to return all the dates in ISO 8601 format: https://github.com/candlepin/subscription-manager/pull/2551

Hence, make the subscription module able to cope with the future value: first try to parse the date as ISO 8601, and if it does not work try to parse as it is done today. Add small unit testing data to check that both ISO 8601 and the locale-specific formats work.

PR #3299, opened 2021-04-20
subscription: allow dates in ISO 8601 format

ptoscano:

Resolves: rhbz#1947445

Backport of #3279 to RHEL 8.


M4rtinK:
Oops, looks like I missed this PR and opened a new backport myself - sorry. :P 

I guess this could be called "pull request blindness" on my side...

poncovka (approved):

VladimirSlavik (approved):
Looks good to me. Thank you!

PR #3300, opened 2021-04-20
subscription: allow dates in ISO 8601 format

ptoscano:

Resolves: rhbz#1947447

Backport of #3279 to future RHEL releases.


jkonecny12:
I don't think there will be more releases for Fedora 34. So there is no reason to merge this to `f34-devel` branch. What do you think @M4rtinK ?

VladimirSlavik:
For master then?

ptoscano:
This was targeted for RHEL 9, which #3330 does; this PR was opened when there was no `rhel-9` branch yet.

Hence, closing.

PR #3301, opened 2021-04-20
Document known crashes under VirtualBox

VladimirSlavik:
SSIA.

### DIFF
--- a/docs/common-bugs.rst
+++ b/docs/common-bugs.rst
@@ -24,6 +24,15 @@ Missing logs
     You can find them during the installation in ``/tmp`` or on the installed system in
     ``/var/log/anaconda/``.
 
+Issue when running in Virtualbox
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+:Issue: An error is reported for Anaconda running in Virtualbox.
+:Solution: Virtualbox is known to cause problems for the whole system. Possible effects include
+    include disappearing devices, memory corruption, missing D-Bus endpoints. Is is not possible
+    to guarantee anything in such environment, including running Anaconda.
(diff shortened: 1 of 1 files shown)


jkonecny12 (approved):
Thanks for creating a known issue for that.
Interesting, do we know why is that happening?

VladimirSlavik:
Unfortunately, no.

I have seen people install on Virtualbox just fine, even make articles about that, so it's definitely not a given that Virtualbox won't work. But so far there just wasn't anyone who would know what to look for, much less have time to debug this. The apparent randomness makes it look like a task for kernel people. So we're left only with this bit of lore that weird system corruption on Virtualbox sometimes happens.

poncovka:
Hi, I think I have figured it out! Please, see my [comment](https://bugzilla.redhat.com/show_bug.cgi?id=1902464#c13). There several workarounds:

1. Increase the memory if possible.
2. Use `--pbkdf*` options in a kickstart file.
3. Change the LUKS version to `LUKS1`.
4. Disable encryption.

VladimirSlavik:
I suppose another indication separate from this would be to look for these:
- earlyoom
- WARNING:blivet:Less than [0-9]+ MiB RAM is currently free, LUKS2 format may fail.
- Activating service name='org.fedoraproject.Anaconda.Modules.Storage' (present more than once)

poncovka:
In this case, the main indication is the storage configuration warning that we show in the UI:
`The available memory is less than 128 MiB which can be too small for LUKS2 format. It may fail.`

I am not sure if there are other cases of this crash that are not related to LUKS2.

VladimirSlavik:
Superseded by #3308.

PR #3302, opened 2021-04-21
Increase the minimal suggested size for /boot

VladimirSlavik:
Formerly 200, now 512 MiB.

Resolves: [rhbz#1907245](https://bugzilla.redhat.com/show_bug.cgi?id=1907245)


poncovka (approved):
Thanks!

jkonecny12 (approved):
I thought, it's 1g these days...
Thanks for the fix.

PR #3140, opened 2021-02-03
Improve X timeout message

VladimirSlavik:
Suggested by @jstodola.

PR #3304, opened 2021-04-22
Xorg timeout give up exception handler test

VladimirSlavik:
Another attempt at the X thing. This gives up the exception handler test, and solves EVERYTHING. Maybe.

The main problem of other solutions is that once X starts, it steals the screen by going to tty6. If the exception handler test-invoking handler is not returned back, a "timeout" handler can be installed instead, which switches back to tty1.

There is still more to explore, so draft.


VladimirSlavik:
So, as it is right now, it's about 100% of my original vision.

But there's still more. I find it jarring that the options for dealing with timeout are printed with timestamps. That's because they don't come from `print()` but rather `stdout_log.warning()`. I can't tell (yet?) what's the difference, but with `print` I get the exceptions bubbling up (again!?!) and whatnot.

I am not sure what this means. The code contains `sleep(2)` which smells of some timing stuff. I traced that particular line to a commit from 2001 with no relevant information.

VladimirSlavik:
I was wrong, I can replace stdout_log with print. I just have to do that for all calls in the function.

From the history, it seems that stdout_log replaced print, and later got replaced by the logging we have now. It might be time to check if it's needed at all.

jstodola:
I gave it a try and the first attempt failed this way:
Screenshot_20210422_215854
Tested with just "inst.xtimeout=1" and updates.img in a VM.

jstodola:
And second attempt:
Screenshot_20210422_220542
This time with `inst.xtimeout=5` and limited cdrom speed:
`# virsh blkdeviotune kill_x sda --read-bytes-sec 500000`

VladimirSlavik:
Can you try without the last commit? If that doesn't work, then that's bad++

jstodola:
bad++
Without the "Improve X timeout message" commit:

Screenshot_20210423_084204
Screenshot_20210423_084537

The same reproducers as in previous attempts.

VladimirSlavik:
I see two things. a) Don't code after 9 pm, b) unwatch processes you want to kill. So maybe not so bad.

However, your tracebacks don't match what I remember seeing earlier, and then it worked just fine for me. So there might still be some other factor not yet fully understood :-/

VladimirSlavik:
@jstodola thank you for testing, if you are up for another round, this should be better. I tried your method, and it worked for me...

VladimirSlavik:
I think I finally got why the exception weirdness happened. The one part is that it was tied to killing the process, which was asynchronous, so it popped up elsewhere. The other part is that Python's `finally` does not handle the exception, it keeps propagating. With these two bits, there are no mysteries left.

pyanaconda/display.py
```diff
@@ -334,6 +344,19 @@ def setup_display(anaconda, options):
         try:
             start_x11(xtimeout)
             do_startup_x11_actions()
+        except TimeoutError as e:
+            log.warning("X startup failed: %s", e)
+            stdout_log.warning("X did not start in the expected time, falling back to text "
+                               "mode. There are multiple ways to help this issue:")
```
VladimirSlavik: Must apply @M4rtinK feedback from https://github.com/rhinstaller/anaconda/pull/3140#pullrequestreview-587841432

pyanaconda/display.py
```diff
@@ -346,20 +346,19 @@ def setup_display(anaconda, options):
             do_startup_x11_actions()
         except TimeoutError as e:
             log.warning("X startup failed: %s", e)
-            stdout_log.warning("X did not start in the expected time, falling back to text "
+            print("X did not start in the expected time, falling back to text "
```
VladimirSlavik: No idea why we used `stdout_log` in the first place. This might have some history and thus possibly reasons why this should not be changed?

pyanaconda/core/util.py
```diff
@@ -217,15 +217,19 @@ def startX(argv, output_redirect=None, timeout=X_TIMEOUT):
...
-    # Use a list so the value can be modified from the handler function
-
+    # Use a class so the values can be modified from the handler functions
     class X11Status:
         def __init__(self):
             self.started = False
+            self.timed_out = False
+
+        def needs_waiting(self):
+            return not (self.started or self.timed_out)
```
VladimirSlavik: Please verify the choice of data structure. (I'm sure `Enum` or `namedtuple` would not help anything, though.)

VladimirSlavik (commented):
Some self-conscious notes.

Apart from the below, commit "Reorganize control flow in startX" also needs amending the message to drop the sentence about exception mysteries.

jstodola:
The change seems to work nice, Xorg is killed and I can select text/vnc mode.

There is one issue and I wonder if you can also reproduce it - starting anaconda in text mode after the X timeout expired results in magnified TUI. When I choose vnc instead of text mode, the TUI to enter the vnc password looks fine. Reproduced 2 out of 2 attempts:
Screenshot_20210426_213646

pyanaconda/core/util.py
```diff
@@ -217,28 +217,46 @@ def startX(argv, output_redirect=None, timeout=X_TIMEOUT):
         :param output_redirect: file or file descriptor to redirect stdout and stderr to
         :param timeout: Number of seconds to timing out.
     """
-    # Use a list so the value can be modified from the handler function
-    x11_started = [False]
+    # Use a class so the values can be modified from the handler functions
+    class X11Status:
```
poncovka, line 221: Feel free to define the class outside of the function.

poncovka (approved):
Looks good to me. Thank you for solving this issue!

VladimirSlavik:
The resolution change happens because after X failure, we reset locale which includes running `setfont` and that destroys the resolution. Frankly, I am not sure what *should* be done here. Under "normal" text mode, we do not change fonts, which preserves the screen size *and* breaks TUI for non-ansi characters - it's all rectangles.

Screenshot_vs_main_2021-05-04_18:09:38

The tradeoff actually seems fair, in a bizarre way: Choose between the comforts of Anaconda in ýóůř ĺáňĝúàğë or above-VGA resolution.

We could presumably restore resolution with `fbset`, but I can't even find it (quickly) in Fedora.



---

[Here](https://github.com/rhinstaller/anaconda/blob/master/anaconda.py#L446) we try to set up Xorg, and `reinitialize_locale()` if that fails, which eventually calls `setup_locale()` and `set_console_font()` ([here](https://github.com/rhinstaller/anaconda/blob/master/pyanaconda/localization.py#L211)), and *that* runs `setfont <something>` ([here](https://github.com/rhinstaller/anaconda/blob/master/pyanaconda/localization.py#L585)).

jstodola (approved):
Thank you for your work on this.
I think the current solution is OK, it displays possible workarounds and allows to continue with the installation, despite the changed resolution. The resolution problem can be solved in a separate PR if necessary.

VladimirSlavik:
Thank you, let's have it then.

For the record, fbset is orphaned in Fedora and apparently nobody missed it enough for the past 3 years: https://src.fedoraproject.org/rpms/fbset

VladimirSlavik:
I addressed all of the feedback and squashed some of the commits. That's ready to merge, I think.

There was another bug left: `ask_vnc_question` did not ask, in fact. The choices appeared without the preceding message. I am almost certain it didn't work before. So I fixed that too...

PR #3305, opened 2021-04-23
Fix the broken rootless podman on Ubuntu 20.04 20210419.1

poncovka:
This is a temporary workaround for Ubuntu 20.04 20210419.1.

**See:** https://github.com/actions/virtual-environments/issues/3229
**Tested on a pull request:** https://github.com/poncovka/anaconda/pull/1

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -39,6 +39,26 @@ jobs:
       TARGET_BRANCH_NAME: 'origin/${{ matrix.target_branch }}'
 
     steps:
+      # FIXME: This is a temporary workaround for Ubuntu 20.04 20210419.1.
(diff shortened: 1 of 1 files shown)


jkonecny12 (changes requested):
Based on the bug report it doesn't happen on ` Ubuntu 18.04`. What about just switching to this version temporary and not complicate the workflow files more than needed?

VladimirSlavik:
I guess either is possible. I'll ack this anyway.

We should start some log, file, or something for infrastructure things where we know in advance that changes will be needed.

jkonecny12 (approved):
After discussion in a team we decided to go with this because it will be more robust solution than having fallback to 18.04 because that could be updated to the broken version in the future.

PR #3306, opened 2021-04-23
Add support for gnome-kiosk

rvykydal:
This is initial patch for transition from metacity to gnome-kiosk.

For a short period of time (sync between anaconda and lorax) anaconda will support both.
It is possible that this lorax patch will be required:
https://github.com/weldr/lorax/pull/1138
If so, I will do the switch by updating the dependency in anaconda.spec (metacity -> gnome-kiosk) after the lorax part is merged and built.

### DIFF
--- a/pyanaconda/display.py
+++ b/pyanaconda/display.py
@@ -182,9 +182,16 @@ def do_startup_x11_actions():
     else:
         xdg_data_dirs = datadir + '/window-manager:/usr/share'
 
-    childproc = util.startProgram(["metacity", "--display", ":1", "--sm-disable"],
-                                  env_add={'XDG_DATA_DIRS': xdg_data_dirs})
-    WatchProcesses.watch_process(childproc, "metacity")
+    vm = "gnome-kiosk"
+    try:
+        childproc = util.startProgram(["gnome-kiosk", "--display", ":1", "--sm-disable", "--x11"],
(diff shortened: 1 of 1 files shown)


pyanaconda/display.py
```diff
@@ -182,9 +182,16 @@ def do_startup_x11_actions():
     else:
         xdg_data_dirs = datadir + '/window-manager:/usr/share'
 
-    childproc = util.startProgram(["metacity", "--display", ":1", "--sm-disable"],
-                                  env_add={'XDG_DATA_DIRS': xdg_data_dirs})
-    WatchProcesses.watch_process(childproc, "metacity")
+    vm = "gnome-kiosk"
```
VladimirSlavik, line 185: I'm not sure why this is here in front of the try statement?

VladimirSlavik (commented):
Looks good to me. Hooray for no metacity needed!

I'm not sure how the dependencies will be handled, though? Will that be another PR?

poncovka:
Hi, I am not sure why we need the transition period. Lorax shouldn't install `metacity` anymore, it is the Anaconda's dependency now. And the clean-up part of Lorax's templates doesn't seem to be critical to the boot.iso build.

rvykydal:
Closing and updating https://github.com/rhinstaller/anaconda/pull/3307 instead to switch to metacity right away.

PR #3313, opened 2021-04-26
Add base repo name for CentOS Stream after repository renaming

jkonecny12:
CentOS Stream renamed the repository to make everything consistent. For that we have to keep old repository name for RHEL but also we need the new one for CentOS Stream.

Resolves: rhbz#1946347


poncovka (approved):

jkonecny12:
Already ported.

PR #3316, opened 2021-04-27
Disable ELN testing on PRs

jkonecny12:
We don't want to have red all the time. Let's disable all ELN testing except container build. We can enable that back when ELN is more stable.

[Test run](https://github.com/jkonecny12/anaconda/runs/2446174661?check_suite_focus=true).


poncovka (approved):
Looks good to me. Thanks!

PR #3317, opened 2021-04-27
Don't allow reformat if the requested file system is unsupported

poncovka:
Disable reformatting on the Manual Partitioning screen if the requested
file system is not supported by Blivet or Anaconda.

In TUI, don't allow to enter the ntfs and tmpfs file systems on the Assign
Mount Points screen. We should support the same file systems as in GUI.

### DIFF
--- a/pyanaconda/modules/storage/devicetree/utils.py
+++ b/pyanaconda/modules/storage/devicetree/utils.py
@@ -49,7 +49,8 @@ def is_supported_filesystem(fmt_type):
     return fmt.type \
         and fmt.supported \
         and fmt.formattable \
-        and (isinstance(fmt, FS) or fmt.type in ["biosboot", "prepboot", "swap"])
+        and (isinstance(fmt, FS) or fmt.type in ["biosboot", "prepboot", "swap"]) \
+        and fmt.type not in ("ntfs", "tmpfs")
 
 
 def get_supported_filesystems():
--- a/pyanaconda/modules/storage/devicetree/utils.py
+++ b/pyanaconda/modules/storage/devicetree/utils.py
@@ -21,7 +21,7 @@ import requests
 from blivet import udev
 from blivet.size import Size
 from blivet.errors import StorageError
-from blivet.formats import device_formats
+from blivet.formats import device_formats, get_format
 from blivet.formats.fs import FS
 from bytesize.bytesize import ROUND_HALF_UP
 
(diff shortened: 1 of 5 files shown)


poncovka:
@vojtechtrefny, can you have a look, please?

pyanaconda/modules/storage/devicetree/utils.py
```diff
@@ -35,23 +35,35 @@
...
+    :param fmt_type: a name of the formatting type
+    :return: True or False
+    """
+    fmt = get_format(fmt_type)
+
+    if not fmt.type:
+        return False
+
+    # Check if the filesystem is unsupported by Anaconda.
+    if fmt.type in ("ntfs", "tmpfs"):
```
vojtechtrefny, line 50: Btrfs was removed intentionally? It makes sense to support "btrfs as filesystem" not only as a volume manager, but it might be worth doing this in a separate commit for clarity.

vojtechtrefny (approved):
Looks good to me.

jstodola:
Why should anaconda not allow to re-format a filesystem like ntfs with a different supported filesystem? Are there any known problems?

poncovka:
> Why should anaconda not allow to re-format a filesystem like ntfs with a different supported filesystem? Are there any known problems?

I will have to reword the commit message. This check is for filesystems that will be created by the installer. It is not for the original formats. Therefore, it should be possible to re-format a filesystem like ntfs with a supported one.

pyanaconda/modules/storage/devicetree/utils.py
```diff
@@ -35,23 +35,35 @@
 log = get_module_logger(__name__)
 
 
+def is_supported_filesystem(fmt_type):
+    """Is the filesystem supported?
+
+    :param fmt_type: a name of the formatting type
+    :return: True or False
```
VladimirSlavik, line 42: `bool` ?

VladimirSlavik (approved):
Looks good to me. Thank you!

Conan-Kudo:
I'm confused what this change is supposed to do and what it's for? Can someone explain this to me? I see it reworks how it interacts with Btrfs, so I'm concerned...

poncovka:
I have split the code into several smaller commits to make it easier to understand. The changes are summarized in the description. Nothing else should be affected. Btrfs should work the same way as before.

pyanaconda/modules/storage/devicetree/utils.py
poncovka: Explained in a new commit.

jkonecny12 (approved):
Looks good to me!

PR #3318, opened 2021-04-27
Clarify the implementation of GetFormatTypeData

poncovka:
Move the implementation of the `GetFormatTypeData` DBus method into a separate
method and document the difference between `GetFormatData` and `GetFormatTypeData`.

### DIFF
--- a/pyanaconda/modules/storage/devicetree/viewer.py
+++ b/pyanaconda/modules/storage/devicetree/viewer.py
@@ -164,25 +164,21 @@ class DeviceTreeViewer(ABC):
     def get_format_data(self, device_name):
         """Get the device format data.
 
+        Return data about a format of the specified device.
+
+        For example: sda1
+
         :param device_name: a name of the device
         :return: an instance of DeviceFormatData
         """
         device = self._get_device(device_name)
         return self._get_format_data(device.format)
 
-    def get_format_type_data(self, format_name):
-        """Get the format type data.
-
-        For example: ext4
-
-        :param format_name: a name of the format type
-        :return: an instance of DeviceFormatData
-        """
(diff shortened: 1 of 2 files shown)


pyanaconda/modules/storage/devicetree/viewer.py
```diff
@@ -203,6 +193,29 @@ def _get_format_data(self, fmt):
...
+
+    def _get_format_type_data(self, fmt):
+        """Get the format type data.
+
+        :param fmt: an instance of DeviceFormat
+        :return: an instance of DeviceFormatData
+        """
+        data = DeviceFormatData()
+        data.type = fmt._type or ""
+        data.mountable = fmt.mountable
```
poncovka, line 215: @vojtechtrefny, I cannot use the format class, because I still need the instance to check if the format is mountable.

pyanaconda/modules/storage/devicetree/viewer.py
```diff
@@ -203,6 +193,29 @@ def _get_format_data(self, fmt):
...
+        data.description = get_blivet_translation(fmt._name) or fmt._type or ""
```
poncovka, line 216: @vojtechtrefny, some classes redefine the `name` property. Isn't is going to be a problem?

poncovka:
Btw. I have checked the places where we call `GetFormatTypeData` and we won't be able to replace them with `GetFormatData` for a specific device.

poncovka:
In general, I don't think this is a good idea.

poncovka:
I have updated the pull request.

VladimirSlavik (approved):
Looks good to me. Thank you!

jkonecny12 (approved):
LGTM!

PR #3319, opened 2021-04-27
The kernel boot argument sshd is removed and should warn user

jkonecny12:
Booting with sshd is not supported anymore. User have to use inst.sshd. Warn user in Dracut about that.

The `sshd` will not allow users to use sshd service but it won't print warning for user. This is better clean solution.


poncovka (approved):
Thanks!

VladimirSlavik (approved):
Looks good to me. Thank you!

PR #3321, opened 2021-04-27
Move post installation scripts to the end of queue for RHV-H

lveyde:
Resolves: rhbz#1953685

Signed-off-by: Lev Veyde <lveyde@redhat.com>

### DIFF
--- a/pyanaconda/installation.py
+++ b/pyanaconda/installation.py
@@ -176,10 +176,6 @@ def _prepare_configuration(payload, ksdata):
         # Join a realm. This can run only after network is configured in the target system chroot.
         configuration_queue.append_dbus_tasks(SECURITY, [security_proxy.JoinRealmWithTask()])
 
-    post_scripts = TaskQueue("Post installation scripts", N_("Running post-installation scripts"))
-    post_scripts.append(Task("Run post installation scripts", runPostScripts, (ksdata.scripts,)))
-    configuration_queue.append(post_scripts)
-
     # setup kexec reboot if requested
     if flags.flags.kexec:
         kexec_setup = TaskQueue("Kexec setup", N_("Setting up kexec"))
@@ -203,6 +199,17 @@ def _prepare_configuration(payload, ksdata):
     # only add write_configs to the main queue if we actually store some kickstarts/configs
     if write_configs.task_count:
         configuration_queue.append(write_configs)
+
+    post_scripts = TaskQueue(
+        "Post installation scripts",
+        N_("Running post-installation scripts")
+    )
+    post_scripts.append(Task(
+        "Run post installation scripts",
+        runPostScripts,
+        (ksdata.scripts,)
+    ))
+    configuration_queue.append(post_scripts)
 
     return configuration_queue
 


lveyde:
@sandrobonazzola

lveyde:
Related RHBZ: https://bugzilla.redhat.com/1953685

lveyde:
> Hello @lveyde! Thanks for updating this PR. We checked the lines you've touched for [PEP 8](https://www.python.org/dev/peps/pep-0008) issues, and found:
> 
> * In the file [`pyanaconda/installation.py`](https://github.com/rhinstaller/anaconda/blob/18453ed0a71c8d4015dc972c358bac30c1ac9c0a/pyanaconda/installation.py):
> 
> > [Line 182:100](https://github.com/rhinstaller/anaconda/blob/18453ed0a71c8d4015dc972c358bac30c1ac9c0a/pyanaconda/installation.py#L182): [E501](https://duckduckgo.com/?q=pep8%20E501) line too long (102 > 99 characters)
> > [Line 183:100](https://github.com/rhinstaller/anaconda/blob/18453ed0a71c8d4015dc972c358bac30c1ac9c0a/pyanaconda/installation.py#L183): [E501](https://duckduckgo.com/?q=pep8%20E501) line too long (101 > 99 characters)
> > [Line 211:100](https://github.com/rhinstaller/anaconda/blob/18453ed0a71c8d4015dc972c358bac30c1ac9c0a/pyanaconda/installation.py#L211): [E501](https://duckduckgo.com/?q=pep8%20E501) line too long (102 > 99 characters)
> > [Line 212:100](https://github.com/rhinstaller/anaconda/blob/18453ed0a71c8d4015dc972c358bac30c1ac9c0a/pyanaconda/installation.py#L212): [E501](https://duckduckgo.com/?q=pep8%20E501) line too long (101 > 99 characters)
> 
> ##### Comment last updated at 2021-04-27 18:00:50 UTC

Fixed

pyanaconda/installation.py
```diff
@@ -176,9 +176,19 @@ def run_generate_initramfs():
...
+        post_scripts = TaskQueue(
+            "Post installation scripts",
+            N_("Running post-installation scripts")
+        )
+        post_scripts.append(Task(
+            "Run post installation scripts",
+            runPostScripts,
+            (ksdata.scripts,)
+        ))
+        configuration_queue.append(post_scripts)
```
sandrobonazzola, line 191: If there's no special reason to write the ks earlier, maybe we can just have it done later and not differentiate between RHVH and others.

poncovka (commented):
I think that we could run `%post` scripts at the end of the installation by default. It won't have an effect on the generated kickstart file and it shouldn't cause any issues with the kexec setup. The documentation doesn't specify exactly when they should be run. What do you think, @jkonecny12, @jstodola?

jstodola:
The more installation tasks are done before `%post`, the better. So it makes sense to do this change for all products, not just RHVH.

jkonecny12:
I can't think of a way how to change kickstart data from `%post` section so I don't see a reason to have `%post` before that.

The only think that come to my mind is to change the output kickstart and attacking the future machines by this. However, I don't think this is really valid because if you can get to the `%post` here you can do much more harm.

jkonecny12:
I can't think of a way how to change kickstart data from `%post` section so I don't see a reason to have `%post` before that.

The only think that come to my mind is to change the output kickstart and attacking the future machines by this. However, I don't think this is really valid because if you can get to the `%post` here you can do much more harm.

In other words, yes please move the `%post` at the end.

poncovka:
@lveyde, could you please remove the checks for RHVH and run the `%post` scripts always at the end of the installation?

lveyde:
> @lveyde, could you please remove the checks for RHVH and run the `%post` scripts always at the end of the installation?

Done

VladimirSlavik (approved):
Looks good to me. Thank you!

poncovka (approved):
Looks good to me. Thank you!

PR #3322, opened 2021-04-27
Stop asking to load deprecated floppy driver

lveyde:
floppy driver seems to be no longer supplied with most recent
RHEL kernels, and attempt to load it shows a FATAL error
message during the boot, e.g. :

dracut-pre-udev[579]: modprobe: FATAL: Module floppy not found in
directory /lib/modules/4.18.0-304.x86_64

Signed-off-by: Lev Veyde <lveyde@redhat.com>

### DIFF
--- a/dracut/anaconda-modprobe.sh
+++ b/dracut/anaconda-modprobe.sh
@@ -24,7 +24,7 @@ done
 shopt -u nullglob
 
 if [ "$ARCH" != "s390" -a "$ARCH" != "s390x" ]; then
-    MODULE_LIST+=" floppy edd iscsi_ibft "
+    MODULE_LIST+=" edd iscsi_ibft "
 else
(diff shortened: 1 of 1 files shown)


lveyde:
Screenshot_lveyde-rhvh-4 4-bad-modules

lveyde:
@sandrobonazzola

sandrobonazzola (approved):
Thanks @lveyde !

poncovka:
The floppy driver is still available on Fedora, so I am not sure if we can stop loading it.
The fatal error was fixed in upstream in: https://github.com/rhinstaller/anaconda/commit/9f6c2300b7b4c9671275159d6355c731574120ba

VladimirSlavik:
> most recent RHEL kernels

Which ones, please? It might be just a matter of making the changes on the right branch...

PR #3324, opened 2021-04-28
Install shim also for ARMv7 EFI

martinezjavier:
Now that we have an ARMv7 shim build, we can install the shim package for
this architecture as well instead of booting GRUB directly. This makes it
consistent with other EFI arches that also contain shim in the boot path.

Suggested-by: Peter Robinson <pbrobinson@gmail.com>


martinezjavier:
Having it as a draft for now since it needs some changes in the uboot and grub2 packages as well.

martinezjavier:
Closing since after more investigation we found that this is not needed for now. I'll open a new pull-request if the need arises in the future.

PR #3326, opened 2021-04-29
Remove support for some of the buildstamp options

poncovka:
We don't use the `Arch`, `UUID` and `BugURL` options from the `.buildstamp` file.


VladimirSlavik (approved):
Looks good to me. Thank you!

PR #3327, opened 2021-05-03
Use ConditionArchitecture for s390x in systemd units

VladimirSlavik:
I'm not sure how to test this, even if it's a good idea. But it removes three TODOs...

Replaces detection based on directories.

### DIFF
--- a/data/systemd/anaconda-direct.service
+++ b/data/systemd/anaconda-direct.service
@@ -3,8 +3,7 @@ Description=the anaconda installation program
 Wants=instperf.service rsyslog.service systemd-udev-settle.service NetworkManager.service
 After=instperf.service rsyslog.service systemd-udev-settle.service NetworkManager.service anaconda-sshd.service
 Requires=anaconda.service
-# TODO: use ConditionArchitecture in systemd v210 or later
-ConditionPathIsDirectory=|/sys/hypervisor/s390
+ConditionArchitecture=|s390x
 ConditionKernelCommandLine=|inst.notmux
 
 [Service]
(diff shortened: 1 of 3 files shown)


jstodola:
Although I have not tried this PR, it might change the current behavior.
The problem is that `/sys/hypervisor/s390` exists when the system is running as a z/VM guests, but it doesn't exist when running as a KVM s390x guest.
`ConditionArchitecture=|s390x` will be true for both z/VM and KVM, resulting in sshd/anaconda-direct being executed by default on s390x KVM guests.

poncovka (commented):
Based on https://github.com/rhinstaller/anaconda/pull/3327#issuecomment-831310585, I would prefer to keep the current solution.

VladimirSlavik:
I agree, primarily because the cost for QE would be disproportionately high, compared to the gains. However, there are still some questions:

- Is it correct that the virtualization methods behave differently?
- How about removing the TODOs and documenting instead the difference?

jstodola:
- Is it correct that the virtualization methods behave differently?
The reason for running sshd by default on z/VM guests or in LPAR is that the console is not very user friendly and for example scrolling in the console can be done by whole pages, not by lines (I'm talking about x3270, as one of the recommended program for connecting to z/VM guests). Also the default resolution 80x43 characters is not too much for displaying anaconda in text mode.
KVM s390x guests on the other hand behave mostly like x86_64 KVM guests, with usable console and even a graphical adapter.
So, yes, it's expected that both virtualization methods behave differently.

poncovka:
>     * How about removing the TODOs and documenting instead the difference?

I guess we could remove the comments and mention our reasons in the commit message.

VladimirSlavik:
Replaced by #3349 which implements the comment removal described above.

PR #3330, opened 2021-05-03
subscription: allow dates in ISO 8601 format (port to RHEL 9)

M4rtinK:

Resolves: rhbz#1947447


M4rtinK:
Ported from upstream PR: #3279

VladimirSlavik (approved):

poncovka (approved):

PR #3331, opened 2021-05-04
Adapt localization part of Makefile to RHEL-9 (#infra)

jkonecny12:
Change where to get translations.

Clear other values which won't work with the original values.


rvykydal (approved):
Looks good to me.

PR #3337, opened 2021-05-05
Handle missing RHSM DBus API

bcl:


VladimirSlavik:
How is this related to #3289?

bcl:
> * How is this related to #3289?

livemedia-creator doesn't use systemd to start anaconda so it doesn't apply.

> * Should this be ported somewhere? I assume all RHELs?

Yes, it doesn't seem to be a problem in Fedora (that I have seen). I'm hitting it on RHEL9.

> 
> * The pep8 suggestion seems quite prudent.

Usually, but I have no idea what the possible range of errors are. I'm assuming that if RHSM is actually needed that it will fail later on in the process (fetching packages?) so catching everything and returning an empty list here is appropriate.

But like I said, I'm not very familiar with the new dbus structure of anaconda, so it's best if someone takes this and cleans it up the right way.

poncovka:
@bcl, the Subscription module is enabled by default on RHEL, but it is disabled on other products. You can [change the product configuration](https://anaconda-installer.readthedocs.io/en/latest/configuration-files.html#product-configuration-files) or disabled it with [a custom configuration file](https://anaconda-installer.readthedocs.io/en/latest/configuration-files.html#custom-configuration-files). The list of enabled modules is defined by:
```

[Anaconda]

kickstart_modules =
    org.fedoraproject.Anaconda.Modules.Timezone
    org.fedoraproject.Anaconda.Modules.Localization
    org.fedoraproject.Anaconda.Modules.Users
    org.fedoraproject.Anaconda.Modules.Payloads
    org.fedoraproject.Anaconda.Modules.Subscription

```


Anyway, we should probably disable the Subscription module if RHSM is not available. Maybe, it should also disabled for every dir and image installation. @M4rtinK, is the subscription supported on these types of installations?

poncovka:
This is related to: https://bugzilla.redhat.com/show_bug.cgi?id=1957063

M4rtinK:
> Maybe, it should also disabled for every dir and image installation. @M4rtinK, is the subscription supported on these types of installations?
I think we should disable subscription support for directory and image installs at least for now. The way it is implemented now it talks to the RHSM DBus service in the environment Anaconda runs in, which then manipulates the subscription state of the system. Then at the end it transfers the subscription tokens, RHSM config file and repo configuration file to the target system.

The expectation is that the physical system where Anaconda runs and the target system are tha same machine and that the installation environment Anaconda runs in is temporary and will vanish at the end of installation (but by that point all the tokens have been moved to the new rootfs so it's not a problem).

With dirinstall this could lead into confusing issues, especially if the machine running Anaconda in the directory/image installation mode is not ephemeral as you would effectively end up with multiple "clones" of the given subscription - once for the "runner" and once for every machine booting from the directory/image Anaconda created. That's definitely not the intended usecase for install time subscription support in Anaconda.

So indeed we should disable subscription support in its current for for directory and image installation workflows and only re enable it if we get clear requests for it & how should it behave to handle the system identity "clone" problem.

VladimirSlavik:
@poncovka , I assume we solved this already elsewhere?

poncovka:
> @poncovka Vendy, I assume we solved this already elsewhere?

The upstream was fixed with https://github.com/rhinstaller/anaconda/pull/3464. I am closing this pull request.

PR #3339, opened 2021-05-05
Remove github-action-run-once script from anaconda-ci (#infra)

jkonecny12:
This script is not required anymore. We are using kstest solution instead which will connect VM directly and not the container.


VladimirSlavik (approved):
Thanks!

poncovka (approved):

VladimirSlavik:
How about merging this?

PR #3343, opened 2021-05-05
The kernel boot argument sshd is removed and should warn user

jkonecny12:
Booting with sshd is not supported anymore. User have to use inst.sshd. Warn user in Dracut about that.

Backport of https://github.com/rhinstaller/anaconda/pull/3319 .

Resolves: rhbz#1954672


poncovka (approved):
Thanks!

VladimirSlavik (approved):
Thank you!

PR #3344, opened 2021-05-05
Install kbd-legacy if keyboard layout is "fi" (#1955793)

AdamWill:
fi is a special case, as described in
https://bugzilla.redhat.com/show_bug.cgi?id=1955793#c6 , where
the layout is taken from kbd-legacy even though the xkb layout
does support ASCII input. So we have to special-case it here.

### DIFF
--- a/pyanaconda/modules/localization/localization.py
+++ b/pyanaconda/modules/localization/localization.py
@@ -196,8 +196,10 @@ class LocalizationService(KickstartService):
         """
         requirements = []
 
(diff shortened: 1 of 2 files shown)


poncovka:
Unfortunately, it doesn't fix the issue, because `vc_keymap` is not always set. I will have to look into that a little more.

poncovka:
Blocked by https://github.com/rhinstaller/anaconda/pull/3372.

poncovka (approved):
Looks good to me. Thanks!

PR #3349, opened 2021-05-07
Remove comments about ConditionArchitecture

VladimirSlavik:
We do not want to use it, the current solution does exactly what we want.

More details at https://github.com/rhinstaller/anaconda/pull/3327#issuecomment-831310585 and in the commit message.


jkonecny12 (approved):
Looks good to me! I agree with your solution.

poncovka (approved):
Looks good to me. Thanks!

PR #3350, opened 2021-05-10
Add the repositories property to the DNF manager

poncovka:
Use the `repositories` property to get a list of repository identifiers.


jkonecny12 (approved):
LGTM!

PR #3351, opened 2021-05-10
Simplify the code for kickstart repositories

poncovka:
Remove the `addons` and `_enabled_repos` properties of the DNF payload and simplify the related code.


poncovka:
Rebased.

PR #3357, opened 2021-05-11
Add po-push target to Makefile on rhel 9

rvykydal:


VladimirSlavik:
Why is this needed? To keep the "interface" same for RHEL 8 & 9?

jkonecny12:
This is not required and I removed the `po-push` functionality on purpose. This is now handled by https://github.com/rhinstaller/anaconda-l10n/blob/master/.github/workflows/pot-file-update.yaml

rvykydal:
> This is not required and I removed the `po-push` functionality on purpose. This is now handled by https://github.com/rhinstaller/anaconda-l10n/blob/master/.github/workflows/pot-file-update.yaml

Ah, I thought we handle it in a different way for rhel-X branches, thank you.

PR #3359, opened 2021-05-11
Do not use pip cache when building containers (#infra)

VladimirSlavik:
Saves a bit more space.

Successful run: https://github.com/vslavik-test-org/anaconda/actions/runs/832532251
(All jobs die on login, container build and test run is successful.)

### DIFF
--- a/dockerfile/anaconda-ci/Dockerfile
+++ b/dockerfile/anaconda-ci/Dockerfile
@@ -38,7 +38,7 @@ RUN set -e; \
   rpmspec -q --requires /tmp/anaconda.spec | grep -v anaconda | xargs -d '\n' dnf install -y; \
   dnf clean all
 
-RUN pip install \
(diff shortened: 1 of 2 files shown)


jkonecny12 (approved):
Great improvement! Thanks!

VladimirSlavik:
@poncovka Vendy, do you have any idea why -0.01% in a Dockerfile fails? This is not Python code, after all...

poncovka:
@VladimirSlavik, I guess that this pull request is not rebased to the latest `master` branch, so our `.codecov.yml` file is not used. However, we can force codecov to always use the config file from the `master` branch, if we want to.

VladimirSlavik:
Ah, I see. Hm, I think I'd prefer to take the setup from master, as we set up most of our infra to disregard current branch and consider *target* branch. But that obviously needs more discussion than this comment :)

PR #3360, opened 2021-05-11
Deduplicate dependency metapackages

VladimirSlavik:
Require the anaconda-install-env-deps package from anaconda-install-img-deps, and adjust the dependencies in the latter to prevent duplication.

This simplifies the spec file somewhat.


poncovka (approved):
Thanks!

rvykydal (approved):
Looks good to me.

PR #3366, opened 2021-05-12
Tweak dependabot configuration - or not?

jkonecny12:
Add:
- automatically set 'infrastructure' label to PR
- add '(#infra):' prefix to the commit messages (don't have suffix support)

Can't be done:
We can't have support for multiple branches. Only the default branch is allowed which is a bit pain :(.

Because of the second, I'm thinking if we want to have the dependabot at all. The problem is that branches created from the main branch would be stuck on the old action version forever and if there will be a security vulnerability than someone could just create PR to the other branch with the outdated version...

So maybe we want to have dependabot support but not merging these PRs. Basically just stay with the moving tag what we have now and having dependabot just as a heads-up that there is a new version of the action available? What do you think?

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,5 +1,8 @@
 # Dependabot will check our GitHub actions and if there will be a new version of any action
 # Dependabot will create a pull request to update these.
+#
+# FIXME: Unfortunately, the Dependabot is checking only the default branch right now.
+#        Could be solved by https://github.com/dependabot/dependabot-core/issues/2511.
 
 # Set update schedule for GitHub Actions
 version: 2
@@ -10,3 +13,7 @@ updates:
     schedule:
       # Check for updates to GitHub Actions every weekday
       interval: "daily"
+    labels:
+      - "infrastructure"
+    commit-message:
+      prefix: "(#infra)"


VladimirSlavik (approved):
I think "dependabot as a heads-up" is fine. We can always re-evaluate that choice to drop it completely. However, I think it's useful for getting release notes like this. I don't have to manage all the actions one by one, subscribing to each and unsubscribing again.

poncovka (approved):
I would keep the dependabot without merging the pull requests whenever possible.

jkonecny12:
Thanks for these reactions. In that case, I wonder how to make it obvious that we don't want to merge these PRs. Maybe have a special label "won't merge"? Is that too stupid solution?

jkonecny12:
**Rebased**

VladimirSlavik:
I don't think anyone [with the power to do it] would merge such a PR - either they don't know and then they won't do it, or they do know and will not merge it.

In case we are all run over by a mob of buses and this is merged, dependabot will keep going and proposing the PRs, so presumably our clueless successors will just go with it.

jkonecny12:
> I don't think anyone [with the power to do it] would merge such a PR - either they don't know and then they won't do it, or they do know and will not merge it.
> 
> In case we are all run over by a mob of buses and this is merged, dependabot will keep going and proposing the PRs. Presumably our clueless successors will just go with it and stay safe.

It's safe for the default branch but not for other branches. However, I agree, I might starting to be a bit paranoid here.

PR #3369, opened 2021-05-13
Avoid a race condition during Connect to Red Hat spoke initialization

M4rtinK:
The Subscription spoke which provides the Connect to Red Hat screen
needs to wait for both the subscription any payload setup threads
to finish before loading valid initial values and making the spoke
available to users.

This is implemented by starting a subscription spoke initialization thread
that waits for the subscription and payload threads to finish. Once both
threads terminate, the subscription spoke initialization thread will update
Subscription spoke state, notify the hub initialization is done and then
terminate.

This unfortunately introduces a race condition as the Connect to Red Hat
spoke is only considered ready once the initialization thread terminates.
If the refresh wins the race and is processed before the thread terminates,
the spoke will be marked as not ready and will stay that way until the
next hub refresh. This is why this condition is fixed by visiting an
accessible spoke and then returning back to the hub.

To fix this issue we introduce a new variable that marks the spoke
as fully initialized and set it before we send the hub refresh signal,
avoiding the race condition and making sure the spoke is marked as ready
once the refresh signal is processed.

(cherry picked from commit 275cc2d0932dac33ae6336727e0d5865f70aa965)

Related: rhbz#1950053
Resolves: rhbz#1959383

### DIFF
--- a/pyanaconda/ui/gui/spokes/subscription.py
+++ b/pyanaconda/ui/gui/spokes/subscription.py
@@ -121,19 +121,26 @@ class SubscriptionSpoke(NormalSpoke):
         # overriden source tracking
         self._overridden_source_type = None
 
+        self._spoke_initialized = False
+
     # common spoke properties
 
     @property
     def ready(self):
-        """The subscription spoke is ready once its initialization thread finishes.
+        """The subscription spoke is ready once the spoke initialization thread finishes.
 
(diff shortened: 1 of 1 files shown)


VladimirSlavik (approved):
Looks good to me. Thank you!

jkonecny12 (approved):
LGTM!

PR #3370, opened 2021-05-13
Change logic in NFS validation

VladimirSlavik:
Now the errors are detected in the right order.


poncovka (approved):
Looks good to me. Thanks!

PR #3371, opened 2021-05-14
Add logic to the bootloader task during installation to set the password for the bootloader

bitcoffeeiux:
i wrote my own addon for the grub2 password set, the spoke in the SystemCategory along with the Storage spoke, but the bootloader password is now written to disk only when partitioning in storage spoke, not at the beginning of the installation. 
So if the storage spoke is complete, the password for the bootloader is stored in the bootloader_proxy instead of written to the storage
I think this is not reasonable


poncovka:
Thanks for the suggestion! I took your idea a little further and opened a new pull request at https://github.com/rhinstaller/anaconda/pull/3430.

PR #3372, opened 2021-05-14
Populate the missing keyboard values before the payload installation

poncovka:
Populate the missing keyboard values before the payload installation, so
the module requirements can be generated for the correct configuration.

Otherwise, the virtual console keymap doesn't have to be set yet and
the module will not generate a requirement for the kbd-legacy package.

Related: rhbz#1955793


jkonecny12 (approved):
LGTM!

VladimirSlavik (approved):
Looks good to me. Thank you!

PR #3377, opened 2021-05-17
Add a log message for successful installation

VladimirSlavik:
The message is visible only in logs of the installation environment, because logs have been already copied to the new system.

Resolves: [rhbz#1949487](https://bugzilla.redhat.com/show_bug.cgi?id=1949487)
Resolves: [rhbz#1955312](https://bugzilla.redhat.com/show_bug.cgi?id=1955312)

CC @velezd @jstodola

### DIFF
--- a/pyanaconda/installation.py
+++ b/pyanaconda/installation.py
@@ -380,3 +380,5 @@ def run_installation(payload, ksdata):
 
     # done
     progress_complete()
+    # this message is automatically detected by QE tools, do not change it lightly
+    log.info("All tasks in the installation queue are done. Installation successfully finished.")


jkonecny12 (changes requested):
Could you please add prefix or postfix to make it obvious that this message is read by an automation and shouldn't change in the future?

poncovka (approved):
Looks good to me. Thanks!

VladimirSlavik:
> Could you please add prefix or postfix to make it obvious that this message is read by an automation and shouldn't change in the future?

Can you add some example, please? I think that needs a comment, not a prefix.

poncovka:
> Could you please add prefix or postfix to make it obvious that this message is read by an automation and shouldn't change in the future?

I don't think we use these. A comment should be fine.

VladimirSlavik:
Added a comment.

pyanaconda/installation.py
```diff
@@ -380,3 +380,5 @@ def run_installation(payload, ksdata):
 
     # done
     progress_complete()
+    log.info("All tasks in the installation queue are done. Installation successfully finished.")
+    # this message is automatically detected by QE tools, do not change it lightly
```
jkonecny12, line 384: Please move the comment above the line.

By the prefix or postfix (which we don't have right now), we could make this more obvious than having a comment. 
Just to explain what I'm thinking:
 ```
log.info("REPORT: All tasks in the installation queue....")
```
The upper `REPORT` will give you a hint that the placeholder is used for something. Ideally it should be explained also in the documentation of course.

However, I don't mind any of these solutions.

jkonecny12 (approved):
Please move the comment above the message. Otherwise look great to me now.

pyanaconda/installation.py
poncovka: I guess we could have a special logger for that. It is definitely an interesting idea, but it needs to be investigated and discussed with the team and QA. That seems to be out of scope of this issue.

jstodola: IMO, do not make it more complicated than it has to be. All we care about is a message (in syslog/journal) that would be sent via the virtio port to the host system.

jstodola (approved):

VladimirSlavik:
Changed order of lines.

PR #3378, opened 2021-05-17
Don't automatically execute the default partitioning (#1954408)

poncovka:
We should automatically execute only a partitioning that was already configured by
the user, for example with a kickstart file.

Don't run the `execute` method in the kickstart installation in GUI and TUI by default.

(cherry picked from commits f5bcaf6, 31d35e9 and 9c0c7a7)

Resolves: rhbz#1954408

### DIFF
--- a/pyanaconda/ui/common.py
+++ b/pyanaconda/ui/common.py
@@ -214,8 +214,6 @@ class Spoke(object, metaclass=ABCMeta):
         self._payload = payload
         self.applyOnSkip = False
 
-        self.visitedSinceApplied = True
-
         # entry and exit signals
         # - get the hub instance as a single argument
         self.entered = Signal()
--- a/pyanaconda/ui/gui/hubs/__init__.py
+++ b/pyanaconda/ui/gui/hubs/__init__.py
@@ -188,13 +188,6 @@ class Hub(GUIObject, common.Hub):
                 self._updateCompleteness(spoke, update_continue=False)
                 spoke.selector.connect("button-press-event", self._on_spoke_clicked, spoke)
                 spoke.selector.connect("key-release-event", self._on_spoke_clicked, spoke)
-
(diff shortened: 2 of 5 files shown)


poncovka:
I had to backport more commits that I thought. Without the additional fixes, it is difficult to control whether the `execute` method should be run during the initialization or not. The change shouldn't affect anaconda, initial-setup, kdump add-on or oscap add-on. It may affect third-party add-ons if they expect their execute method to be run automatically during the kickstart installation. The workaround is to call the `execute` method explicitly from the `initialize` method.

jkonecny12 (commented):
One small point, you have a hash for a different commit (before cherry pick) in second commit. Not sure if you want to fix that or not.

About this change. I would be cautious about the hub change. It looks like something what can introduce unexpected bugs. How good is this tested on Fedora ? Are there alternative implementation?

poncovka:
> About this change. I would be cautious about the hub change. It looks like something what can introduce unexpected bugs. How good is this tested on Fedora ?

This solution was released upstream on March 30. So far, there are no reported issues related to this change. The removed code could be run only during automated installations and we have these pretty much covered.

> Are there alternative implementation?

I can make it work without the hub changes, but it won't be pretty, it won't be tested upstream and it will definitely introduce unexpected bugs.

poncovka:
Fixed commit messages.

jkonecny12 (approved):
Thanks for explanation. I guess we can go with this solution. Especially if it's tested more a month on Fedora with no issues reported. I understand this is the best option to go with.

Could we run whole RHEL-8 KS tests suite on this? Just to be sure.

Heads-up to @jstodola to know that we are landing something like.

VladimirSlavik (approved):
LGTM, thank you!

jstodola (approved):
Thanks for heads-up and I agree with this change.

PR #3379, opened 2021-05-17
Add the "Encrypt my data" checkbox to the custom partitioning spoke (#1903786)

poncovka:
Allow users to generate encrypted default partitioning to have the same
experience as with the automatic partitioning in the storage spoke.

(cherry picked from commit 8ce46f4)

Resolves: rhbz#1903786

### DIFF
--- a/pyanaconda/ui/gui/spokes/lib/accordion.py
+++ b/pyanaconda/ui/gui/spokes/lib/accordion.py
@@ -477,7 +477,8 @@ class CreateNewPage(BasePage):
         is created, it will be removed and replaced with a Page for it.
     """
     def __init__(self, title, create_clicked_cb, autopart_type_changed_cb,
-                 default_scheme, partitions_to_reuse=True):
+                 encrypted_changed_cb, default_scheme, default_encryption,
+                 partitions_to_reuse=True):
         super().__init__(title)
 
         # Create a box where we store the "Here's how you create a new blah" info.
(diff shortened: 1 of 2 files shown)


poncovka:
encrypt_data_rhel

pyanaconda/ui/gui/spokes/lib/accordion.py
```diff
@@ -551,4 +552,27 @@ def __init__(self, title, create_clicked_cb, autopart_type_changed_cb,
         combo.set_hexpand(False)
         combo.set_active_iter(default or store.get_iter_first())
         self._createBox.attach(combo, 0, 5, 2, 1)
+
+        label = Gtk.Label(
+            label=C_(
+                "GUI|Custom Partitioning|Autopart Page",
+                "_Automatically created mount points can be encrypted by default:"
```
jkonecny12, line 559: I know I should ask on the original PR but is this only me or the sentence doesn't feel right.

jkonecny12 (approved):
Other than the label it looks good to me.

pyanaconda/ui/gui/spokes/lib/accordion.py
jkonecny12: @VladimirSlavik what do you think?

VladimirSlavik: I wanted to say "looks fine", but actually, no. I think it's one of these problems where you can't see it because you already know what it means.

I think dropping the "by default" part might help, but I'd rather bring in a native speaker...

VladimirSlavik (approved):
Looks good to me. Thank you!

Obsessing over the label is optional, as Jirka says.

pyanaconda/ui/gui/spokes/lib/accordion.py
jkonecny12: Quote from IRC:
```
<sbueno> jkonecny, the sentence itself is fine. what throws me off is the text next to the checkbox after
<sbueno> makes me do a doulbe take
<sbueno> so maybe saying something like: "Encrypt automatically created mount points by default:"
```

jkonecny12 (approved):
Looks even better now. Thanks!

poncovka:
The label was changed based on @snbueno's suggestion:
encrypt_data_rhel

jkonecny12:
Backport this change to master please. Thanks!

PR #3381, opened 2021-05-17
Use daily builds for ELN containers

jkonecny12:
ELN containers had disabled daily builds repositories. I don't know exact reason, it might be because it's harder to enable them (see commit message) or because there were not available during creation of the containers.

Add these back to avoid failures with dependencies because we are waiting for a build.

Also enable ELN nightly builds for Anaconda project. Our other projects have it enabled so why not Anaconda.

### DIFF
--- a/.packit.yml
+++ b/.packit.yml
@@ -26,7 +26,7 @@ jobs:
     metadata:
       targets:
         - fedora-rawhide
-#        - fedora-eln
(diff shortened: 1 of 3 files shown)


VladimirSlavik (approved):
Looks good to me. Thank you!

poncovka (approved):

PR #3383, opened 2021-05-18
Don't set the home directory to None (#1960803)

poncovka:
None is not a valid DBus value, use an empty string instead.

Resolves: rhbz#1960803


VladimirSlavik (approved):
Looks good to me. Thank you!

M4rtinK (approved):
Looks good to me as well. :)

jkonecny12 (approved):
LGTM!

PR #3385, opened 2021-05-18
Master add rhel 9 ks tests

jkonecny12:
Add support for running KS tests on `rhel-9` branch PRs.

Test runs:
[rhel-9](https://github.com/jkonecny12/anaconda/actions/runs/853843620)
[master](https://github.com/jkonecny12/anaconda/actions/runs/853848271)

### DIFF
--- a/.github/workflows/kickstart-tests.yml
+++ b/.github/workflows/kickstart-tests.yml
@@ -55,10 +55,12 @@ jobs:
           REF_DEVEL='f'$FEDVERS'-devel'
           REF_RELEASE='f'$FEDVERS'-release'
 
-          if [ "$TARGET_BRANCH" == "$REF_DEVEL" ] || [ "$TARGET_BRANCH" == "$REF_RELEASE" ] ; then
-            echo "::set-output name=lorax_build_container::fedora:$FEDVERS" ;
-          elif [ "$TARGET_BRANCH" == "master" ] ; then
-            echo "::set-output name=lorax_build_container::fedora:rawhide" ;
+          if [ "$TARGET_BRANCH" == "$REF_DEVEL" ] || [ "$TARGET_BRANCH" == "$REF_RELEASE" ]; then
+            echo "::set-output name=lorax_build_container::fedora:$FEDVERS";
+          elif [ "$TARGET_BRANCH" == "master" ]; then
+            echo "::set-output name=lorax_build_container::fedora:rawhide";
+          elif [ "$TARGET_BRANCH" == "rhel-9" ]; then
+            echo "::set-output name=lorax_build_container::registry-proxy.engineering.redhat.com/rh-osbs/ubi9:latest";
           else
             echo "Branch $TARGET_BRANCH is not configured to run kickstart tests, aborting."
             exit 0
@@ -146,7 +148,8 @@ jobs:
 
           mkdir -p kickstart-tests/data/images
 
-      - name: Build boot.iso from Fedora and this branch
+      - name: Build boot.iso for Fedora branches
+        if: ${{ ! startsWith(needs.pr-info.outputs.base_ref, 'rhel-') }}
         run: |
           # /var/tmp tmpfs speeds up lorax and avoids https://bugzilla.redhat.com/show_bug.cgi?id=1906364
           sudo podman run -i --rm --privileged --tmpfs /var/tmp:rw,mode=1777 -v `pwd`:/source:ro -v `pwd`/kickstart-tests/data/images:/images:z ${{ env.LORAX_BUILD_CONTAINER }} <<EOF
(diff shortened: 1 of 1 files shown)


.github/workflows/kickstart-tests.yml
```diff
@@ -55,10 +55,12 @@ jobs:
...
-          if [ "$TARGET_BRANCH" == "$REF_DEVEL" ] || [ "$TARGET_BRANCH" == "$REF_RELEASE" ] ; then
-            echo "::set-output name=lorax_build_container::fedora:$FEDVERS" ;
-          elif [ "$TARGET_BRANCH" == "master" ] ; then
-            echo "::set-output name=lorax_build_container::fedora:rawhide" ;
+          if [ "$TARGET_BRANCH" == "$REF_DEVEL" ] || [ "$TARGET_BRANCH" == "$REF_RELEASE" ]; then
+            echo "::set-output name=lorax_build_container::fedora:$FEDVERS";
+          elif [ "$TARGET_BRANCH" == "master" ]; then
+            echo "::set-output name=lorax_build_container::fedora:rawhide";
+          elif [ "$TARGET_BRANCH" == "rhel-9" ]; then
+            echo "::set-output name=lorax_build_container::registry-proxy.engineering.redhat.com/rh-osbs/ubi9:latest";
```
VladimirSlavik, line 63: These changes are not needed at all, you can specify the container inline in the "duplicated" RHEL-9-specific step below.

.github/workflows/kickstart-tests.yml
```diff
@@ -55,10 +55,12 @@ jobs:
...
           REF_RELEASE='f'$FEDVERS'-release'
 
-          if [ "$TARGET_BRANCH" == "$REF_DEVEL" ] || [ "$TARGET_BRANCH" == "$REF_RELEASE" ] ; then
-            echo "::set-output name=lorax_build_container::fedora:$FEDVERS" ;
-          elif [ "$TARGET_BRANCH" == "master" ] ; then
-            echo "::set-output name=lorax_build_container::fedora:rawhide" ;
+          if [ "$TARGET_BRANCH" == "$REF_DEVEL" ] || [ "$TARGET_BRANCH" == "$REF_RELEASE" ]; then
+            echo "::set-output name=lorax_build_container::fedora:$FEDVERS";
+          elif [ "$TARGET_BRANCH" == "master" ]; then
+            echo "::set-output name=lorax_build_container::fedora:rawhide";
```
VladimirSlavik, line 61: Not critical, but - if you want to remove the spaces, maybe do so in another commit? The diff is very hard to understand because all the lines change, but there is no change to be seen in the part I highlight here. (Arguably that's a Github problem, Gitlab could display than nicely.)

.github/workflows/kickstart-tests.yml
```diff
@@ -180,6 +183,87 @@ jobs:
...
+          sudo podman run -i --rm --privileged --tmpfs /var/tmp:rw,mode=1777 -v `pwd`:/source:ro -v `pwd`/kickstart-tests/data/images:/images:z ${{ env.LORAX_BUILD_CONTAINER }} <<EOF
+          set -eux
+
+          # lorax must run on RHEL host to produce RHEL guest images :-(
+          # need to enable internal repos so that we don't need a subscription;
+          rm -f /etc/yum.repos.d/*
+          cat <<EOR > /etc/yum.repos.d/rhel9.repo
+          [RHEL-9-BaseOS]
+          name=baseos
+          baseurl=http://download.devel.redhat.com/rhel-9/nightly/RHEL-9-Beta/latest-RHEL-9/compose/BaseOS/\\\$basearch/os/
```
VladimirSlavik, line 199: I don't doubt it works as it should, but I don't understand it. Why are there three `\`s ?

VladimirSlavik (changes requested):
Thank you, this is a great improvement! I'd love to have no difference between starting kickstart tests on all branches, as well as having it all in one file.

I also very much like the idea of less workflows that pollute their history with "skipped" on every.single.comment. Which this helps!

However, I think you should decide if you're a) copying the lorax step, or b) generating parameters for some "universal" run. The way it looks now, it seems you started by extending the generator step but then hit the difference with repo files and had to add the new lorax step, and generator changes were left there anyway.

I can see only one reason to build it this way: To extend this for RHEL-8 too and have a Fedora branch and RHEL branch. The "generator logic" in `pr-info` would then create the repo file(s) too, in some way.

jkonecny12:
Still *WIP* I broke Fedora tests execution :(.

.github/workflows/kickstart-tests.yml
jkonecny12: Yep, it's a bash nonsense.

jkonecny12:
After a longer discussion with @VladimirSlavik we found a way how to proceed and avoid duplication by creating a new container for exactly this purpose. I'm closing this PR because the outcome will be completely different code.

Thanks a lot @VladimirSlavik for your help!

PR #3386, opened 2021-05-18
Introduce the docs/release-notes directory

poncovka:
Add the `docs/release-notes/` directory and document how to provide a release
note for a patch. Use `docs/release-notes/template.rst` as a template.

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -33,11 +33,13 @@ How to Contribute to the Anaconda Installer (the short version)
 
 - open a pull request for the ``<next Fedora number>-devel`` branch (f25-devel, etc.)
 - check the *Commit Messages* section below for how to format your commit messages
+- check the *Release Notes* section below for how to provide a release note
 
 2) I want to contribute to the RHEL Anaconda installer:
 
 - open a pull request for the ``<RHEL number>-branch``  branch (rhel7-branch, etc.)
 - check the *Commits for RHEL Branches* section below for how to format your commit messages
+- check the *Release Notes* section below for how to provide a release note
 
 If you want to contribute a change to both the upstream and RHEL Anaconda then follow both a) and b) separately.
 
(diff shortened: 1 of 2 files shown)


docs/release-notes/template.rst
```diff
@@ -0,0 +1,4 @@
+:Type: Kickstart / TUI / GUI / Modularization / ...
+:Summary: Summarize the change (#bug number)
+:Description: Describe the change in detail.
+:Link: Provide links to pull requests, bugs or issues.
```
VladimirSlavik, line 4: I'm not sure if more links need more `:Link:` lines, but I imagine that's irrelevant because the result will be processed by human anyway.

VladimirSlavik (approved):
Looks good to me. Thank you!

CONTRIBUTING.rst
```diff
@@ -168,6 +170,24 @@ or
 Use ``Resolves`` if the patch fixes the core issue which caused the bug.
 Use ``Related`` if the patch fixes an ancillary issue that is related to, but might not actually fix the bug.
 
+Release Notes
+^^^^^^^^^^^^^
+
+If you are submitting a patch that should be documented in the release notes, create a copy of the
+``docs/release-notes/template.rst`` file, modify its content and add the new file to your patch.
```
jkonecny12, line 177: I don't think `patch` is the correct word here. I would suggest `contribution`.

docs/release-notes/template.rst
```diff
@@ -0,0 +1,4 @@
+:Type: Kickstart / TUI / GUI / Modularization / ...
+:Summary: Summarize the change (#bug number)
+:Description: Describe the change in detail.
```
jkonecny12, line 3: I would show in this example how to do multiline description.

docs/release-notes/template.rst
jkonecny12: So maybe change this to `:Links:` and add info that there could be multiple links each on one line?

jkonecny12 (commented):
Looks good to me otherwise. Thanks for the idea and implementation!

CONTRIBUTING.rst
VladimirSlavik: Pull request?

docs/release-notes/template.rst
poncovka: Fixed.

docs/release-notes/template.rst
poncovka: Fixed.

CONTRIBUTING.rst
poncovka: It should be clear that the file needs to be part of the changes and not attached or inserted into a pull request comment or a commit message.

poncovka:
Updated.

VladimirSlavik (approved):
Looks good to me. Thank you!

PR #3387, opened 2021-05-18
Add suggestions to kickstart error message in liveinst

VladimirSlavik:
Resolves: [rhbz#1893428](https://bugzilla.redhat.com/show_bug.cgi?id=1893428)

I'm not sure if the text can reference Fedora editions, please discuss.

### DIFF
--- a/data/liveinst/liveinst
+++ b/data/liveinst/liveinst
@@ -110,7 +110,7 @@ for opt in `cat /proc/cmdline` $*; do
         ;;
     ks=*|kickstart=*|--ks=*|--kickstart=*|inst.ks=*|--inst.ks=*|inst.kickstart=*|--inst-kickstart=*)
         title="Configuration not supported"
-        text="Kickstart is not supported on live installs.  This installation will continue interactively."
+        text="Kickstart is not supported on Live ISO installs, please use netinstall or standard ISO.  This installation will continue interactively."
         if which zenity &> /dev/null; then
             zenity --warning --title="$title" --text="$text"
          else
--- a/docs/common-bugs.rst
+++ b/docs/common-bugs.rst
@@ -342,9 +342,18 @@ These issues are related to automated installations that use kickstart files.
 Automatic installation in Live OS
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-:Issue: The reporter would like to run a kickstart installation in Live OS.
+:Issue: The reporter would like to run a kickstart installation in Live OS. One of these messages
+    is displayed:
+    `Kickstart is not supported on Live ISO installs, please use netinstall or standard ISO.  This installation will continue interactively.`
+    Alternatively, before Fedora 35: `Kickstart is not supported on live installs.  This installation will continue interactively.`
+
 :Solution: Kickstart installations in Live OS are not supported. Please, run the installation with
-    ``boot.iso``.
+    one of the following types of images:
+
+    * netinstall ISO (such as the Server edition of Fedora)
+    * standard ISO
+    * ``boot.iso``
+
 :Example: `rhbz#1027160 <https://bugzilla.redhat.com/show_bug.cgi?id=1027160>`_
 
 Invalid partitioning in the output kickstart file


jkonecny12 (approved):
LGTM!

VladimirSlavik:
@Gel0bmstu @mikhailnov What do you non-Fedora-distro people think about this?

Gel0bmstu:
Hi, in ROSA Linux we use shell wrapper over _/usr/sbin/anaconda_ script, which allows user to perform kickstart install, even if hi load with livecd:

```
# --liveinst arg broke kickstart installation, so we should
# check if inst.ks setted in cmdline before begin installation.
# It's don't broke installation from GUI, because desktop icon
# run 'liveinst' script. 
LIVE_ARGS=""
if ! grep -q 'inst.ks' /proc/cmdline ; then
    LIVE_ARGS="--liveinst"
    if [ -n "$LIVE_BLOCK" ]; then
        LIVE_ARGS="$LIVE_ARGS --method=livecd:${LIVE_BLOCK}"
        # where LIVE_BLOCK is a path to /dev/mapper/live-base "file"
    fi
fi

....

exec anaconda ${LIVE_ARGS} "$@"
```

I think it is better to let the user install the system, than to confuse him with an incomprehensible message like "Kickstart is not supported on live install". However, the solution offered by @VladimirSlavik LGTM too!

VladimirSlavik:
Thank you. I was concerned mainly because the script could run anywhere, but with this change it would mentions things specific to Fedora.

poncovka (changes requested):
I have forgot to submit my review.

data/liveinst/liveinst
```diff
@@ -110,7 +110,7 @@ for opt in `cat /proc/cmdline` $*; do
         ;;
     ks=*|kickstart=*|--ks=*|--kickstart=*|inst.ks=*|--inst.ks=*|inst.kickstart=*|--inst-kickstart=*)
         title="Configuration not supported"
-        text="Kickstart is not supported on live installs.  This installation will continue interactively."
+        text="Kickstart is not supported on live installs.  This installation will continue interactively.  Use kickstart with Server, IoT, or other non-Live images."
```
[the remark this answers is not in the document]
VladimirSlavik: Thank you, that's better.

docs/common-bugs.rst
```diff
@@ -344,7 +344,7 @@ Automatic installation in Live OS
 
 :Issue: The reporter would like to run a kickstart installation in Live OS.
 :Solution: Kickstart installations in Live OS are not supported. Please, run the installation with
-    ``boot.iso``.
+    ``boot.iso`` ("netinst") for the Server edition of Fedora.
```
poncovka, line 347: I would try to expand the sentence a little more with something like:
`... or use a netinstall or standard ISO (for example, with the Server edition of Fedora).`

docs/common-bugs.rst
```diff
@@ -344,7 +344,7 @@ Automatic installation in Live OS
 
 :Issue: The reporter would like to run a kickstart installation in Live OS.
```
poncovka, line 345: The reporter mentioned that it was difficult to find a solution for the message: `Kickstart is not supported on live installs. This installation will continue interactively.` Could you maybe document this message in the issue description?

VladimirSlavik:
Oops, sorry, I did not think at all when I changed this. That's hardly better.

docs/common-bugs.rst
VladimirSlavik: Is netinstall something else than boot.iso?

docs/common-bugs.rst
VladimirSlavik: Yes, but I'll use the new one.

poncovka (approved):
Looks good to me. Thanks!

docs/common-bugs.rst
VladimirSlavik: Used both - f33 is still "active", and f34 will be, so the old message will appear more often for at least a year.

docs/common-bugs.rst
VladimirSlavik: Resolved out of band: netinstall is a subset of boot.iso. Being a boot.iso means the media is built to boot into anaconda. netinstall then has only that, while the other (unnamed) possibility is to have also packages on the media.

mikhailnov:
We in ROSA install OS using rsync via Anaconda, so _any_ ISO is a LiveCD, either booted to multi-user.target or to anaconda.target. That is why we make workarounds described by @Gel0bmstu in some places.

PR #3388, opened 2021-05-19
Replace Anaconda product configuration files with profiles

poncovka:
The support for the product configuration files was removed and replaced with profiles.

Each profile can be identified by a unique id and it can define additional options for
the automated profile detection. The profile will be chosen based on the ``inst.profile``
boot option, or based on the ``ID`` and ``VARIANT_ID`` options of the os-release files.
The profile configuration files are located in the ``/etc/anaconda/profile.d/`` directory.

The ``inst.product`` and ``inst.variant`` boot options are deprecated.

- [x] Draft the changes.
- [x] Propose [a self-contained change](https://fedoraproject.org/wiki/Changes/Replace_Anaconda_product_configuration_files_with_profiles).
- [x] Implement the changes.
- [x] Test the changes.

**Requires:** https://github.com/rhinstaller/initial-setup/pull/134
Resolves: rhbz#1974819


poncovka:
@jkonecny12 @VladimirSlavik @M4rtinK @rvykydal

VladimirSlavik (approved):
I think this should be fine.

rvykydal:
The draft looks great to me

docs/configuration-files.rst
```diff
@@ -54,71 +54,86 @@ It defines and documents all supported sections and options. The file is located
...
-The product configuration files allow to override some of the configuration options for
-specific products. The files are located at ``/etc/anaconda/product.d/``.
+The profile configuration files allow to override some of the configuration options for
+specific profiles and products. The files are located at ``/etc/anaconda/profile.d/``.
 
 .. note::
 
     Anaconda previously used so called install classes for the product-specific configuration.
     Install classes were completely removed and replaced by the product configuration files.
+    The product configuration files were rewritten into profiles.
```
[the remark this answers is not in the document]
poncovka: Fixed.

poncovka:
Rebased.

poncovka:
Replaced `product_id` with `os_id`.

docs/configuration-files.rst
```diff
@@ -54,71 +54,86 @@ It defines and documents all supported sections and options. The file is located
...
-Otherwise, it will be automatically loaded from the ``.buildstamp`` file on the installation
-media. This file is generated by ``lorax`` using the ``--product`` and ``--variant`` options.
+Profile detection
+^^^^^^^^^^^^^^^^^
 
-Based on the provided product and variant names, the installer will look up the right
-configuration file in the ``/etc/anaconda/product.d/`` directory.
+The profile can be specified by the ``inst.profile`` boot option or the ``--profile`` cmdline
+option. Based on the provided profile id, the installer will look up the right configuration
+file in the ``/etc/anaconda/profile.d/`` directory.
```
Conan-Kudo, line 81: Shouldn't these be installed into `/usr/share/anaconda/profile.d` instead?

Conan-Kudo (changes requested):

docs/configuration-files.rst
jkonecny12: Could you please specify reasoning why `/usr/share` instead of `/etc/`? I wonder what would be the benefit.

Conan-Kudo: Basically, if these are intended to be shipped as static files that users don't edit, they shouldn't be in `/etc`.

jkonecny12: Hmm, good point. However, to have that we would have to enable user to overload the configuration files from `/etc` especially now when there is Fedora Silverblue. We have something like that in `/etc/anaconda/conf.d/*.conf`. So the correct thing to do here is IMHO to move existing profiles to `/usr/share/...` and leave the `/etc/anaconda` support for overloading the conf files. And even in that case I'm not sure if user could overload the whole profile. I fear it would be a bit more work which shouldn't be part of this PR.

What do you think @poncovka ?

poncovka: I don't want to support overloading. Why users shouldn't edit existing profiles?

jkonecny12: I agree that this is work without a big benefit.
To @poncovka the main reason is that next RPM updated config wouldn't be applied because users has changed the config manually. However, in the Anaconda situation it doesn't have a big issue because there aren't many use-cases on the installed machine.

Conan-Kudo: I didn't expect overloading to be a thing, I just expect these to not be admin-editable generally.

poncovka: All our configuration files and drop dirs for configuration files are in `/etc/anaconda/`. Would you expect all of them to be in `/usr/share/`?

jkonecny12: Definitely not the `/etc/anaconda/conf.d/` users shouldn't place/edit anything in the `/usr/...` .

poncovka: In that case, I don't think that `conf.d` and `profile.d` should be at different locations. We support custom configurations for both of them, it is documented and encouraged. If we don't want to move all of these to `/usr/share/anaconda`, it might be better to keep using `/etc/anaconda/`.

Conan-Kudo:
`liveinst` will need to be reworked so that it can do the right thing for product selection and activating so-called "live" profiles.

poncovka:
> `liveinst` will need to be reworked so that it can do the right thing for product selection and activating so-called "live" profiles.

There are no live profiles anymore. See: https://github.com/rhinstaller/anaconda/pull/3390

Conan-Kudo:
There will be for CentOS Hyperscale: https://lists.centos.org/pipermail/centos-devel/2021-June/077079.html

I don't have it working just yet because of screwiness with the product definitions, but I will have Live and non-Live Workstation variants.

poncovka:
@Conan-Kudo, the workaround for Fedora Workstation Live was removed in `anaconda-35.16-1`, it was never supposed to be a support for live products/profiles. We can implement the solution I have suggested in https://github.com/rhinstaller/anaconda/pull/3390#issuecomment-844941563, but it won't be part of this pull request. Could you open a bug for that, please?

Conan-Kudo:
> @Conan-Kudo, the workaround for Fedora Workstation Live was removed in `anaconda-35.16-1`, it was never supposed to be a support for live products/profiles. We can implement the solution I have suggested in [#3390 (comment)](https://github.com/rhinstaller/anaconda/pull/3390#issuecomment-844941563), but it won't be part of this pull request. Could you open a bug for that, please?

Sure!

Conan-Kudo:
> @Conan-Kudo, the workaround for Fedora Workstation Live was removed in `anaconda-35.16-1`, it was never supposed to be a support for live products/profiles. We can implement the solution I have suggested in [#3390 (comment)](https://github.com/rhinstaller/anaconda/pull/3390#issuecomment-844941563), but it won't be part of this pull request. Could you open a bug for that, please?

Sure, should I file the bug against CentOS Stream 9, since that's where I'll need it?

Conan-Kudo:
@poncovka I've filed a BZ on Fedora Rawhide for it: https://bugzilla.redhat.com/show_bug.cgi?id=1978221

docs/configuration-files.rst
```diff
@@ -54,71 +54,86 @@ It defines and documents all supported sections and options. The file is located
...
-specific products. The files are located at ``/etc/anaconda/product.d/``.
+The profile configuration files allow to override some of the configuration options for
+specific profiles and products. The files are located at ``/etc/anaconda/profile.d/``.
 
 .. note::
 
     Anaconda previously used so called install classes for the product-specific configuration.
-    Install classes were completely removed and replaced by the product configuration files.
+    Install classes were completely removed and replaced by the profile configuration files.
+    These configuration files used to be called product configuration files for some time.
```
jkonecny12, line 67: I would remove this sentence. From the new user perspective is this information not valuable and for older users used to InstallClasses it will not be helpful too.

jkonecny12 (approved):
Looks good to me!

docs/configuration-files.rst
poncovka: It could be valuable for current users. The install classes were replaced by product configuration files in January 2019.

PR #3389, opened 2021-05-19
Remove the configuration file for CentOS Linux

poncovka:
Remove the configuration file for CentOS Linux.
This configuration file can be removed from upstream.

Rename the configuration file for CentOS Stream.
It is a base configuration for all CentOS based products.

**Related to:** https://github.com/rhinstaller/anaconda/pull/3388

### DIFF
--- a/data/product.d/centos.conf
+++ b/data/product.d/centos.conf
@@ -1,7 +1,0 @@
-# Anaconda configuration file for CentOS Linux.
-
-[Product]
-product_name = CentOS Linux
-
-[Base Product]
-product_name = CentOS Stream
--- a/data/product.d/centos.conf
+++ b/data/product.d/centos.conf
@@ -1,4 +1,4 @@
-# Anaconda configuration file for CentOS Stream.
+# Anaconda configuration file for CentOS.
 
 [Product]
 product_name = CentOS Stream
--- a/tests/nosetests/pyanaconda_tests/product_test.py
+++ b/tests/nosetests/pyanaconda_tests/product_test.py
@@ -269,11 +269,6 @@ class ProductConfigurationTestCase(unittest.TestCase):
             ENTERPRISE_PARTITIONING
         )
         self._check_default_product(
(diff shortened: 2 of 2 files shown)


jkonecny12 (approved):
LGTM!

Conan-Kudo (approved):

VladimirSlavik (commented):
Mostly LGTM, but...

I think the file should stay with the name CentOS Stream. I think the CentOS != Stream thing is still at the forefront of people's minds, and this could be misconstructed as another "attempt at rewriting the world for CentOS == Stream". (Even if realistically the "rewriting" already happened and this only reflects it.) I think the file name does not matter much and is a very small price to pay.

BTW, it's just an unfortunate coincidence, but for "Files changed", GitHub tries to be smart about the whole PR and the diff looks like we dropped Stream to sneak it into the CentOS file, making that narrative more compelling.

Feel free to feel otherwise. I'd also prefer not to have to deal with these things.

poncovka:
> I think the file should stay with the name CentOS Stream. I think the CentOS != Stream thing is still at the forefront of people's minds, and this could be misconstructed as another "attempt at rewriting the world for CentOS == Stream". (Even if realistically the "rewriting" already happened and this only reflects it.) I think the file name does not matter much and is a very small price to pay.
> 

The configuration for CentOS Linux and CentOS Stream is the same. We needed two files, because their product names in the `.buildstamp` file are different. After https://github.com/rhinstaller/anaconda/pull/3388, we won't see a difference between these two products, because their `os-release` files use the same ids. So it doesn't really make sense to keep specifically the CentOS Stream file, because it will match any product with `ID="centos"` in `/etc/os-release`.

Files from CentOS Linux:
centos-data

Files from CentOS Stream:
centos-stream-data

VladimirSlavik:
Ah, in that case it makes sense.

VladimirSlavik (approved):

PR #3390, opened 2021-05-19
Remove the configuration file for Fedora Workstation Live

poncovka:
Move the configuration for Fedora Workstation Live to the configuration file
for Fedora Workstation and use that one instead. There is no non-live version
of Fedora Workstation at this moment.

**Related to:** https://github.com/rhinstaller/anaconda/pull/3388


VladimirSlavik:
Code LGTM, but is this correct in terms of how the Fedora <insert name of group> sees things?

poncovka:
> Code LGTM, but is this correct in terms of how the Fedora sees things?

I think so. There is no product called Fedora Workstation Live. The configuration was created as a temporary workaround.

Conan-Kudo (changes requested):
This is not true. We have a Fedora Workstation netinst ISO that uses the regular configuration. Please don't do this.

poncovka:
> This is not true. We have a Fedora Workstation netinst ISO that uses the regular configuration. Please don't do this.

Ok, I have found a solution, that will allow to keep both configurations. Btw. where we can find this ISO?

Conan-Kudo:
Hmm, actually, I guess I misremembered, we stopped offering that in Fedora 31 (the last release with it was Fedora 30).

poncovka:
I have discussed it with @jkonecny12 and we would like to merge these configurations into one for now. My solution was allow these files to define a type of the expected installation environment. Then there could be a separate configuration for Fedora Workstation on Live OS. We can always implement it if needed.

PR #3391, opened 2021-05-19
Add new container specific for creating boot.iso with Anaconda

jkonecny12:
Create boot.iso with the current Anaconda could be a bit tedious task. Especially when you want to support both Fedora and RHEL. Let's move the differences to container and build the container which will allow us to simplify and easily share the workflow.

This new solution is first part of work to support kickstart-tests run on rhel branches. This could be also used on our https://github.com/rhinstaller/kickstart-tests repository IMHO.

Test [run](https://github.com/jkonecny12/anaconda/runs/2620827036?check_suite_focus=true).


jkonecny12:
Opened from wrong branch :(.

PR #3396, opened 2021-05-20
Remove unnecessary package from anaconda-ci container (#infra)

jkonecny12:
It was introduced by 473daecb1e6db643c8da22d34f281a19fff91fe1 to apply custom patch from upstream but that part is already removed from the anaconda-ci so let's remove also this package.


VladimirSlavik (approved):
Looks good to me. Thank you!

poncovka (approved):

PR #3398, opened 2021-05-21
Drop the "Allow SSH root login with password" checkbox (#1855736)

M4rtinK:
The "Allow SSH root login with password" checkbox in the root password
spoke was added as a temporary migration helper for user when Fedora
disable password based SSH login for the root account back in 2019.
This in itself was actually just restoring OpenSSH upstream default
behavior, as the OpenSSH upstream disabled password based root login
already back in 2015.

As enough time has been given for migration to more secure remote access methods
that are less prone to remote brute force password guessing (such as key based
authentication or user account with unique username) it is now time to
drop the temporary checkbox from the Anaconda GUI.

Resolves: rhbz#1855736


poncovka (changes requested):
The unit tests have failed.

M4rtinK:
@poncovka Unit tests should be fixed now. :)

poncovka (approved):
Looks good to me. Thanks!

VladimirSlavik (approved):
Looks good to me. Thank you!

poncovka:
Btw. the change needs to be in some way ported to upstream.

M4rtinK:
> Btw. the change needs to be in some way ported to upstream.

Indeed, FESCO is already voting about [the change proposal](https://fedoraproject.org/wiki/Changes/Drop_Rootpw_SSH_From_Installer) it in some manner on the [FESCO issue](https://pagure.io/fesco/issue/2612).

If I understand the preliminary votes correctly, most seem to be in favour of keeping the option at the moment. So I guess an upstream version of this would have to be parametric via a Anaconda profile switch, similar to how we hide things like the EULA notice or Blivet GUI screen on RHEL.

VladimirSlavik:
Not ported in two years, and rejected by FESCO, so removing the label.

PR #3399, opened 2021-05-21
The results returned by check_admin_user_exists() may cause problems

bitcoffeeiux:
with the user's password configuration when the following scenarios
are encountered on the text-based installation:
1. The root password is not set and is locked
2. Create user `user1` with administrator group and do not set password
3.completed() returned False, but mandatory() return false too,
This causes neither the root password nor the new user to have a password,
thus preventing users from accessing the system once the installation
 is complete.
Maybe we should check to see if the users of the wheel group have
a full password

### DIFF
--- a/pyanaconda/modules/users/users.py
+++ b/pyanaconda/modules/users/users.py
@@ -389,7 +389,10 @@ class UsersService(KickstartService):
         for user in self.users:
             if not user.lock:
                 if "wheel" in user.groups:
-                    return True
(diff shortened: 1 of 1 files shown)


VladimirSlavik:
The unit tests fail because you fail on first locked admin user and do not examine all of them. The else clause is what breaks things.

bitcoffeeiux:
Yes, I misunderstood a little bit. Thank you for pointing it out

PR #3400, opened 2021-05-21
Remove support for metacity

rvykydal:
Resolves: rhbz#1961104

### DIFF
--- a/anaconda.spec.in
+++ b/anaconda.spec.in
@@ -255,7 +255,6 @@ Requires: system-logos
 
 # Needed to compile the gsettings files
 BuildRequires: gsettings-desktop-schemas
-BuildRequires: metacity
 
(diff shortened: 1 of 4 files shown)


poncovka:
Shouldn't we remove also the fallback code that tries to run `metacity`?

rvykydal:
> Shouldn't we remove also the fallback code that tries to run `metacity`?

Oh we should, thank you:)

poncovka (approved):
Looks good to me. Thanks!

VladimirSlavik (approved):
LGTM, thank you!

jkonecny12 (approved):
Looks great to me! Thanks to this we should not need Build-root repository anymore. I'll create PR for that after this is merged!

PR #3408, opened 2021-05-25
  In text installation mode, the user selects the password but does not set the password forcibly

bitcoffeeiux:
In text mode, if you select `Use password` when you create a user and set it as an administrator group, you can start the installation without entering user password. But this goes against the logic of select the `Use password`:

create user1 eg:
```
1) [x] Create user
2) Full name
    user1
3) User name
    user1
4) [x] Use password
5) Password
6) [x] Administrator
7) Groups
    wheel
```
in the summary:
```
9) [ ] User creation
         (You must set a password)
```
The installation process is now ready to begin...

### DIFF
--- a/pyanaconda/ui/tui/spokes/user.py
+++ b/pyanaconda/ui/tui/spokes/user.py
@@ -224,7 +224,9 @@ class UserSpoke(FirstbootSpokeMixIn, NormalTUISpoke):
     @property
     def mandatory(self):
         """Only mandatory if no admin user has been requested."""
-        return not self._users_module.CheckAdminUserExists()
+        return (not self._users_module.CheckAdminUserExists() or
+                (self._use_password and not bool(self.user.password or
+                                                 self.user.is_crypted)))
 
     @property
     def status(self):


VladimirSlavik (commented):
Thank you! The code change works exactly as advertised.

I tried to simplify the condition, but it is not possible - not on its own, anyway. The spoke persists all the user "attributes" even with user switched off, so that would break in some situations. I think that's ok.

However, fixing this properly includes changing the method's docstring and a better commit message. The current commit message is "what was wrong", but normally we write also what this makes possible. And of course the first line should be a summary, according to guidelines.

@bitcoffeeiux, do you want me to fix up these things and push to your branch, or do you want to do that yourself?

bitcoffeeiux:
> Thank you! The code change works exactly as advertised.
> 
> I tried to simplify the condition, but it is not possible - not on its own, anyway. The spoke persists all the user "attributes" even with user switched off, so that would break in some situations. I think that's ok.
> 
> However, fixing this properly includes changing the method's docstring and a better commit message. The current commit message is "what was wrong", but normally we write also what this makes possible. And of course the first line should be a summary, according to guidelines.
> 
> @bitcoffeeiux, do you want me to fix up these things and push to your branch, or do you want to do that yourself?

I don't know which documents need to be modified. can you give me some guidance?

jkonecny12:
> > Thank you! The code change works exactly as advertised.
> > I tried to simplify the condition, but it is not possible - not on its own, anyway. The spoke persists all the user "attributes" even with user switched off, so that would break in some situations. I think that's ok.
> > However, fixing this properly includes changing the method's docstring and a better commit message. The current commit message is "what was wrong", but normally we write also what this makes possible. And of course the first line should be a summary, according to guidelines.
> > @bitcoffeeiux, do you want me to fix up these things and push to your branch, or do you want to do that yourself?
> 
> I don't know which documents need to be modified. can you give me some guidance?

If I understand it correctly @VladimirSlavik wants to change the doc string of the method to reflect your changes. Now you have there:
```
        """Only mandatory if no admin user has been requested."""
```
However, this is not the only condition anymore. You should reflect your changes also to the method doc string.

Another thing is the commit message, you should follow global guidelines for [git commit messages](https://gist.github.com/robertpainsi/b632364184e70900af4ab688decf6f53) otherwise tools like GitHub won't show your message correctly.

VladimirSlavik:
I edited the docstring and commit message and opened it as #3551. @bitcoffeeiux, here are the options:
1. I merge the other PR, 3551. You are author of the commit, but not PR.
2. I take the edited commit from 3551, force-push it here into this PR, and merge. You are author of commit and PR.
3. You edit this PR to satisfy the conditions for fixed docstring and commit message. You are author of commit and PR.

PR #3411, opened 2021-05-25
Revert "Drop the "Allow SSH root login with password" checkbox (#1855…

M4rtinK:
…736)"

This reverts commit 2061fa8773c142e7f1471b60f8d899562a4e1537.

Looks like FESCO will not agree with the removal in Fedora, so let's
reverse the drop of the option from RHEL as well, at least for now.

Resolves: rhbz#1855736


poncovka (approved):
Looks good to me.

jstodola (approved):
Thank you

VladimirSlavik (approved):

rvykydal (approved):
Looks good to me.

PR #3413, opened 2021-05-26
the func completed() logic not same in tui/spokes/root_password.py and gui/spokes/root_password.py

bitcoffeeiux:
Why can't we unify their logic?
gui/spokes/root_password:
def completed(self):
        return self._users_module.IsRootPasswordSet

### DIFF
--- a/pyanaconda/ui/tui/spokes/root_password.py
+++ b/pyanaconda/ui/tui/spokes/root_password.py
@@ -60,7 +60,7 @@ class PasswordSpoke(FirstbootSpokeMixIn, NormalTUISpoke):
 
     @property
     def completed(self):
-        return bool(self._users_module.IsRootPasswordSet or self._users_module.IsRootAccountLocked)
+        return self._users_module.IsRootPasswordSet
 
     @property
     def showable(self):


VladimirSlavik:
Interesting question, with no apparent answer. The code used to be the same. Commit 52bf264 (part of #1858) changes the logic in GUI to a more simple one, without any apparent reason. Its predecessor ef54072 changes TUI, without such change in logic.

bitcoffeeiux:
> predecessor

So I recommend keeping the logic consistent to make it easier for developers to understand

VladimirSlavik:
I agree, consistency is good. However, I want to understand *why* - then I can change it.

VladimirSlavik:
I think I understand now. Considered in isolation, the TUI has the correct code. However, the GUI automatically locks the root user if it's set to no password and unlocked. To do so, it must be possible to exit the spoke, which means the exit condition must be relaxed. So that is why it works that way. Note that this is how it works now, I did not check how it was when the change was made.

VladimirSlavik:
Closed. Thank you for bringing this up, anyway!

PR #3417, opened 2021-05-27
Move the support for the treeinfo metadata into a DBus module

poncovka:
The metadata is needed only during an update of the base repository,
so replace the `_install_tree_metadata` attribute of the DNF payload with
a local variable.

Call the `configure_substitution` method of the DNF manager to configure the
substitution variables. Call the `reset_substitution` method to reset these
variables.

Move the support for the treeinfo metadata into a DBus module, simplify it
and add unit tests.


pyanaconda/modules/payloads/payload/dnf/tree_info.py
```diff
@@ -0,0 +1,429 @@
...
+        :return: True or False
+        """
+        return self._type in conf.payload.enabled_repositories_from_treeinfo
+
+    @property
+    def valid(self):
+        """Is the repository valid?
+
+        This is only check if the repository is invalid.
+        If success it doesn't mean it is valid.
```
rvykydal, line 404: Funny :D, maybe we could have 'invalid' attribute then, but even that might not help. Feel free to ignore this comment.

rvykydal (approved):
Looks good to me.

pyanaconda/modules/payloads/payload/dnf/tree_info.py
```diff
@@ -0,0 +1,429 @@
...
+
+    def _get_base_repository(self):
+        """Return metadata of the base repository.
+
+        :return: an instance of TreeInfoRepoMetadata or None
+        """
+        repos = DEFAULT_REPOS
+
+        for repo_md in self.repositories:
+            if repo_md.name in repos:
```
jkonecny12, line 349: I think this constant could be in-lined. It would be even more readable that way.

jkonecny12 (approved):
Looks good to me otherwise. Great work here!

pyanaconda/modules/payloads/payload/dnf/tree_info.py
poncovka: Fixed.

pyanaconda/modules/payloads/payload/dnf/dnf_manager.py
```diff
@@ -381,6 +398,20 @@ def substitute(self, text):
             text, self._base.conf.substitutions
         )
 
+    def configure_substitution(self, release_version=None):
+        """Set up the substitution variables.
+
+        :param release_version: a string for $releasever or None
```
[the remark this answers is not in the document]
poncovka: Fixed.

pyanaconda/modules/payloads/payload/dnf/tree_info.py
poncovka: Fixed.

poncovka:
Rebased and tested.

PR #3418, opened 2021-05-27
Run tests on every push to a base branch

poncovka:
We need to update the coverage of these branches in the codecov.io,
so we can compare the pull request coverage with the base coverage.

**Tested on:** https://github.com/poncovka/anaconda/runs/2686386124

### DIFF
--- a/.github/workflows/push-tests.yml
+++ b/.github/workflows/push-tests.yml
@@ -1,0 +1,34 @@
+name: Run tests on push
+on:
+  push:
+    branches:
+      - master
+      - f[0-9]+-devel
+      - f[0-9]+-release
+
+jobs:
+  unit-tests:
+    runs-on: ubuntu-20.04
+    timeout-minutes: 30
+    steps:
+      - name: Clone repository
+        uses: actions/checkout@v2
+
+      - name: Build anaconda-ci container
+        run: make -f Makefile.am anaconda-ci-build
+
+      - name: Run tests in anaconda-ci container
+        run: |
+          # put the log in the output, where it's easy to read and link to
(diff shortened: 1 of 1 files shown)


VladimirSlavik (approved):
LGTM, I agree that this should not use the environment.

Thanks for fixing the coverage thing!

rvykydal (approved):
Looks good to me.

.github/workflows/push-tests.yml
```diff
@@ -0,0 +1,34 @@
...
+jobs:
+  unit-tests:
+    runs-on: ubuntu-20.04
+    timeout-minutes: 30
+    steps:
+      - name: Clone repository
+        uses: actions/checkout@v2
+
+      - name: Build anaconda-ci container
+        run: make -f Makefile.am anaconda-ci-build
```
jkonecny12, line 18: Do not build the container, just let it download from our quay.io repository. These will be downloaded automatically on when starting the tests. So just removing these two lines should be enough.

Or re-use this part https://github.com/rhinstaller/anaconda/blob/master/.github/workflows/tests.yml#L57 .

jkonecny12 (approved):
Looks good to me. However, I don't think we should build the containers.

jkonecny12 (changes requested):
Sorry it should have been NACK :D

jkonecny12 (approved):
After off-site discussion there is a valid reason to use this way. We are not able to detect changed files so in case there will be dependency change or Dockerfile change we won't be able to detect that.
On the other hand we are not guarded against the changes from the system. So, if some dependency will break Anaconda we would get it here but not from the repository because it's gated before the upload.

Removing NACK, this is about to choose the smaller evil in this case.

PR #3420, opened 2021-05-31
The Subscription module should be an add-on (#1957063)

poncovka:
Change the prefix of the Subscription module. Use the add-ons namespace.

The boss will find and start this module automatically. Don't start the module
if the initial conditions are not met.

The boss will run the installation tasks of this module now. Don't explicitly
schedule these tasks.

Resolves: rhbz#1957063
Replaces: https://github.com/rhinstaller/anaconda/pull/3337


poncovka:
This is one of the possible solutions of the bug https://bugzilla.redhat.com/show_bug.cgi?id=1957063. The main difference is that an add-on can refuse to start and the installer will continue without it. That allows us to automatically disable the module if the rhsm service is not available in the installation environment.

rvykydal:
Looks good and makes sense to me. I don't see any drawbacks, @M4rtinK will be better at checking if there are some.

VladimirSlavik (commented):
My inner perfectionist cringes at the idea "fix it by making it an addon", but the solution is actually valid. I guess the objection is really that having official parts of anaconda as addons seems weird. Technically, this is fine. The reduction in config file verbosity is very nice.

Maybe the files should move too? Or is that another detail that takes the "addon" thing too literally?

M4rtinK:
> My inner perfectionist cringes at the idea "fix it by making it an addon", but the solution is actually valid. I guess the objection is really that having official parts of anaconda as addons seems weird. 

That was my initial though as well, but now I see it as just changing the way how the module is started - eq. via the addons interface instead of the "regular" static module list. And I'm fine with that, especially as the functionality is already this easily available to do what we need & without changes needed in the module/addon starting functionality.

I guess over time we could still improve these things - maybe make all modules startup dynamic like this or possibly make it possible to start a "regular" module dynamically without directly using the addon namespace.

>Technically, this is fine. The reduction in config file verbosity is very nice.
> 
> Maybe the files should move too? Or is that another detail that takes the "addon" thing too literally?
I think keeping the module files where they are is correct, as we are just using the addon interface to make the module startup dynamic, while still otherwise being an important core part of Anaconda on RHEL.

pyanaconda/modules/subscription/__main__.py
```diff
@@ -17,9 +17,16 @@
...
 # Red Hat, Inc.
 #
+
+# Initialize the service.
 from pyanaconda.modules.common import init
 init()
 
+# Check the initial conditions.
+from pyanaconda.modules.subscription.initialization import check_initial_conditions
+check_initial_conditions()
```
jkonecny12, line 27: I know you probably handled this correctly but I want to check. Are we sure that this won't stay in a loop when something will start the addon again and again by dbus/systemd-service and it would kill itself all the time?

jkonecny12 (commented):
Looks great to me otherwise!

pyanaconda/modules/subscription/initialization.py
```diff
@@ -19,26 +19,47 @@
...
 
+RHSM_SYSTEMD_UNIT_NAME = "rhsm.service"
+
+
+def check_initial_conditions():
+    """Can the Subscription service run?"""
+
+    # Exclude the initial setup.
+    if PAYLOADS.service_name not in conf.anaconda.kickstart_modules:
+        log.debug("The Payloads module is required. Quit.")
```
M4rtinK, line 46: Please add a "subscription:" suffix for these messages, its used in many other places in the module and related GUI parts and it helps to quickly locate all subscription based log messages. 

Also just "Quit." sounds a bit weird to me, maybe it could be more explicit "Not starting the subscription DBus module." or "Subscription DBus module startup aborted." ?

M4rtinK (approved):
Looks good to me, just a suggestion how to possibly improve some log message wording.

Also all seems to work fine as far as I can tell from my testing:
- in normal Fedora environment subscription stuff is not started, as expected
- if I inject the RHSM packages via an updates image, the subscription stuff shows up

In any case, thanks for looking into this! :)

poncovka:
We will use a different approach.

PR #3421, opened 2021-06-01
Unmount the source if the mount point is already in use when checking it

century6:
When the service is restarted by "systemctl restart anaconda" command in tty2, anaconda will failed to mount a DBus source at a mount point generated from /run/install/source, for example /run/install/source/mount-0000-cdrom, because the mount point has been used by the anaconda process before restart and has not been unmounted before the process was terminated. We can fix it by unmounting the source when checking the mount point.

### DIFF
--- a/pyanaconda/modules/payloads/source/mount_tasks.py
+++ b/pyanaconda/modules/payloads/source/mount_tasks.py
@@ -71,15 +71,20 @@ class SetUpMountTask(Task, ABC):
     def run(self):
         """Run source setup."""
         log.debug("Mounting installation source")
-        self._check_mount()
+        if not self._check_mount():
+            task = TearDownMountTask(self._target_mount)
+            task.run()
         return self._do_mount()
 
     def _check_mount(self):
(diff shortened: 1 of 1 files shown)


poncovka:
We should probably improve the clean up when anaconda quits.

jkonecny12:
Hello,

Thank you for opening this PR but this is not the way we should solve this issue. In general, you are fixing symptoms but not the cause. With this solution you will still stay with the system with already mounted sources because last run of the Anaconda was killed.

I would suggest that we should re-use the [anconda-cleanup](https://github.com/rhinstaller/anaconda/blob/master/scripts/anaconda-cleanup) script which should already solve this issue. However, I wonder what you are trying to achieve here, are you calling this command in the boot.iso (installation environment) or on the Live media? If on the boot.iso then why do you need to call `systemctl restart anaconda`? If you have a valid use-case for this, you can try to enhance our systemd unit and call this script when the unit ends / is killed.

So or so, I'm closing this PR because we don't want to go this way.

PR #3430, opened 2021-06-04
Apply the bootloader options before the installation

poncovka:
Some of the bootloader options are not needed until the installation.
If we apply them during the scheduling of the partitions, it won't be
possible to change them after that, for example, from an add-on. Let's
apply these options later, right before the bootloader installation.

Suggested-by: bitcoffee <854182924@qq.com>


jkonecny12 (approved):
Looks good to me.

VladimirSlavik (approved):
Looks good to me. Thank you!

PR #3431, opened 2021-06-05
Use yescrypt hashing method for shadow passwords

besser82:
The yescrypt hashing method is considered to be much stronger than sha512crypt and fully supported by libxcrypt.  It is based on NIST-approved primitives and on par with argon2 in strength.

Fresh installed systems, as well as newly computed hashes for the UNIX shadow file should prefer this method.

### DIFF
--- a/pyanaconda/core/users.py
+++ b/pyanaconda/core/users.py
@@ -29,6 +29,7 @@ from pyanaconda.core.regexes import GROUPLIST_FANCY_PARSE, NAME_VALID, PORTABLE_
 import crypt
 from pyanaconda.core.i18n import _
 import re
+from random import SystemRandom as sr
 
 from pyanaconda.anaconda_loggers import get_module_logger
 log = get_module_logger(__name__)
@@ -43,13 +44,25 @@ def crypt_password(password):
     :returns: crypted representation of the original password
     :rtype: str
     """
+    # yescrypt is not supported by Python's crypt module,
+    # so we need to generate the setting ourselves
+    b64 = "./0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz"
+    setting = "$y$j9T$" + "".join(sr().choice(b64) for sc in range(24))
+
+    # and try to compute the password hash using our yescrypt setting
     try:
-        cryptpw = crypt.crypt(password, crypt.METHOD_SHA512)
-    except OSError as exc:
-        raise RuntimeError(_(
-            "Unable to encrypt password: unsupported "
-            "algorithm {}").format(crypt.METHOD_SHA512)
-        ) from exc
+        cryptpw = crypt.crypt(password, setting)
+
+    # Fallback to sha512crypt, if yescrypt is not supported
+    except OSError:
+        log.info("yescrypt is not supported, falling back to sha512crypt")
+        try:
+            cryptpw = crypt.crypt(password, crypt.METHOD_SHA512)
+        except OSError as exc:
+            raise RuntimeError(_(
+                "Unable to encrypt password: unsupported "
+                "algorithm {}").format(crypt.METHOD_SHA512)
+            ) from exc
 
     return cryptpw
 
--- a/pyanaconda/core/users.py
+++ b/pyanaconda/core/users.py
@@ -43,13 +43,13 @@ def crypt_password(password):
     :returns: crypted representation of the original password
     :rtype: str
     """
-    cryptpw = crypt.crypt(password, crypt.METHOD_SHA512)
-
-    if cryptpw is None:
+    try:
+        cryptpw = crypt.crypt(password, crypt.METHOD_SHA512)
+    except OSError as exc:
         raise RuntimeError(_(
             "Unable to encrypt password: unsupported "
             "algorithm {}").format(crypt.METHOD_SHA512)
-        )
+        ) from exc
 
     return cryptpw
 
--- a/pyanaconda/modules/common/structures/user.py
+++ b/pyanaconda/modules/common/structures/user.py
@@ -234,9 +234,18 @@ class UserData(DBusData):
 
         To create an encrypted password you can use python:
 
+        python3 -c 'import crypt; print(crypt.crypt("My Password", "$y$j9T$My Sault"))'
+
+        This will compute a hash of your password using the yescrypt hasing method and
+        your provided salt.
+
+        If the yescrypt method is not supported by your system, you can use the
+        sha512crypt hashing method:
+
         python3 -c 'import crypt; print(crypt.crypt("My Password", "$6$My Sault"))'
 
-        This will generate sha512 crypt of your password using your provided salt.
+        This will compute a hash of your password using the sha512crypt hasing method
+        and your provided salt.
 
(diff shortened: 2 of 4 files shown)


besser82:
See https://fedoraproject.org/wiki/Changes/yescrypt_as_default_hashing_method_for_shadow for the purpose of this PR.

pyanaconda/core/users.py
```diff
@@ -43,13 +45,26 @@ def crypt_password(password):
     :returns: crypted representation of the original password
     :rtype: str
     """
-    cryptpw = crypt.crypt(password, crypt.METHOD_SHA512)
+    # yescrypt is not supported by Python's crypt module,
+    # so we need to generate the setting ourselves
+    saltchars = string.ascii_letters + string.digits + "./"
```
VladimirSlavik, line 50: Pylint does not want `string`. That's why the tests fail.

VladimirSlavik (commented):
Björn, thank you for the contribution! I can't comment on the validity of the actual crypto code, so I won't ack, but otherwise LGTM. Thanks for taking care of the tests too.

I think the change of the default needs approval by FESCO anyway, and it might be useful to check if Python could support yescrypt. (@hroncok - any idea what that woudl take?) Then it might be better to just use that instead of constructing the salts manually, of course...

hroncok:
> it might be useful to check if Python could support yescrypt. (@hroncok - any idea what that woudl take?)

No idea. I am no a crypto expert :)

I'd ask @tiran instead.

tiran:
I'm against adding any new feature to crypt module. We want to reduce the scope of Python's stdlib. I explained my reasoning in https://bugs.python.org/issue44309#msg395160

The upstream PR won't help you much any way. Python 3.10 is already in feature freeze. The new feature won't be availble until Python 3.11 final (October 2022) and Fedora 37.

pyanaconda/core/users.py
besser82: > Pylint does not want `string`. That's why the tests fail.

How to solve that?  Simply like this example or does it take more work?

```
import string as _str
…
... _str.ascii_letters ...
```

I'm not a Python expert, as I'm mostly coding in c.

pyanaconda/core/users.py
```diff
@@ -43,13 +45,26 @@ def crypt_password(password):
     :returns: crypted representation of the original password
     :rtype: str
     """
-    cryptpw = crypt.crypt(password, crypt.METHOD_SHA512)
+    # yescrypt is not supported by Python's crypt module,
```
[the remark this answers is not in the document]
besser82: > I'm against adding any new feature to crypt module. We want to reduce the scope of Python's stdlib. I explained my reasoning in https://bugs.python.org/issue44309#msg395160

I think, this is valid reason to keep it as is for now.

pyanaconda/core/users.py
besser82: I've fixed that in rebased commit.

rvykydal (approved):
Looks good to me, thank you.

rvykydal:
Tested manually to be working as expected (using yescrypt).

besser82:
Rebased onto recent master.

besser82:
The failing test seems unrelated to this PR.

VladimirSlavik (approved):
Looks good to me. Thank you!

VladimirSlavik:

(This works only for org members, sorry.)

jkonecny12:
Please ignore the failed test. Should be fixed by https://github.com/rhinstaller/anaconda/pull/3445 .

besser82:
> Please ignore the failed test. Should be fixed by #3445 .

Rebased onto recent master to hopefully make the tests happy.  =)

besser82:
> /kickstart-test --testtype smoke

Thanks for kicking the test again!

FYI, the [change is likely to be accepted](https://pagure.io/fesco/issue/2623) (+6/-0) by FESCo during their upcoming meeting next week.

VladimirSlavik:
Thanks for the heads-up, just saw it too. Anyway, the man to decide is Radek, I'm just running around mashing buttons :)

jkonecny12:
Please ignore the `packit` test it seems some problem on TFT side. @thrix it's happening everywhere, could you please take a look on that?

besser82:
The SWC has been accepted in yesterdays FESCo meeting.

pyanaconda/core/users.py
```diff
@@ -43,13 +44,24 @@ def crypt_password(password):
...
+    # so we need to generate the setting ourselves
+    b64 = "./0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz"
+    setting = "$y$j9T$" + "".join(sr().choice(b64) for sc in range(24))
 
+    # and compute the password hash using our yescrypt setting
+    cryptpw = crypt.crypt(password, setting)
+
+    # Fallback to sha512crypt, if yescrypt is not supported
     if cryptpw is None:
```
poncovka, line 56: `None` as a return value is not documented in https://docs.python.org/3/library/crypt.html#crypt.crypt, so I was playing with it and it looks like `crypt.crypt` raises the OSError exception now (https://github.com/python/cpython/commit/0d3fe8ae4961bf551e7d5e42559e2ede1a08fd7c).

```
>>> crypt.crypt("password", "")
  File "/usr/lib64/python3.9/crypt.py", line 82, in crypt
    return _crypt.crypt(word, salt)
OSError: [Errno 22] Invalid argument

```

besser82: Indeed!  Thanks for the catch!

besser82: Addressed in 8781132.

besser82:
Rebased onto recent master.

poncovka (commented):
Looks good to me. Thanks!

besser82:
@VladimirSlavik, @poncovka: Rebased onto recent master branch.  I hope the packit tests will pass now.

VladimirSlavik:
The really important ones are unit tests, we make sure these work all the time. Rpm tests are expected to always work too, although there can be breakages due to other package changes.

Then kickstart, if it works. The whole system must be ok to make the boot image and run on it, so there are times where it just isn't reliable. We're early in the f35 cycle so it's very possible this could just randomly break.

Packit is more like "maybe it works if the weather is good... and you sacrificed ten bulls on the night of last full moon... and the date is palindromatic when expressed in base64..."

besser82:
> The really important ones are unit tests, we make sure these work all the time. Rpm tests are expected to always work too, although there can be breakages due to other package changes.
> 
> Then kickstart, if it works. The whole system must be ok to make the boot image and run on it, so there are times where it just isn't reliable. We're early in the f35 cycle so it's very possible this could just randomly break.

Thanks for clarification!  Good to know that.

> Packit is more like "maybe it works if the weather is good... and you sacrificed ten bulls on the night of last full moon... and the date is palindromatic when expressed in base64..."

I didn't know it's that hard to pass the Packit tests.  XD

rvykydal:
Kickstart tests passed. I've checked locally that yescrypt is used with the patch.
Thank you all, I am going to merge the PR.

besser82:
Thank you for merging!

PR #3432, opened 2021-06-07
Reduce permissions for our tests workflow (#infra)

jkonecny12:
This workflow can only download content of the repository nothing else is permitted.

Thanks to this change we could avoid environment verification because even the external contributors can't really modify anything. `GITHUB_TOKEN` for this workflow can only read the content of the repository nothing else is allowed.

Test PRs:
[Success run.](https://github.com/jkonecny12/anaconda/pull/11/checks?check_run_id=2765057988)
[Test to push to repository.](https://github.com/jkonecny12/anaconda/pull/11/checks?check_run_id=2765041562#step:4:22) - This shouldn't be allowed!

*WIP*:
- [x] Secrets could still be read, however, we can solve that by moving our secrets to the environment level and use these environments for specific workflows.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -1,9 +1,11 @@
 name: Run validation tests
 on: pull_request_target
 
+permissions:
+  contents: read
+
 jobs:
   unit-tests:
-    environment: staging
     runs-on: ubuntu-20.04
     timeout-minutes: 30
     strategy:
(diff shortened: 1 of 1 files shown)


jkonecny12:
We can't port this to RHEL-9 because these are our runners so the runners could be abused.

jkonecny12:
This PR is now prepared to be merged.

rvykydal (commented):
Looks good to me but I don't feel competent to approve (needs more eyes)

VladimirSlavik (approved):
This should work, I think.

PR #3434, opened 2021-06-08
Add brltty to boot.iso

VladimirSlavik:
This adds brltty to boot.iso and starts it with default configuration. That means:
- Some usb braille devices might be automatically detected and used.
- This works only in TUI, no GUI.
- No method of configuration is added.

https://bugzilla.redhat.com/show_bug.cgi?id=1584679

### DIFF
--- a/anaconda.spec.in
+++ b/anaconda.spec.in
@@ -223,6 +223,7 @@ Requires: gsettings-desktop-schemas
 Requires: nm-connection-editor
 Requires: librsvg2
 Requires: gnome-kiosk
+Requires: brltty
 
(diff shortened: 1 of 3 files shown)


poncovka (approved):

jkonecny12 (approved):
LGTM!

VladimirSlavik:
Asking on the brltty mailing list how to actually test this, hence the additional label.

VladimirSlavik:
Verified to some extent that it can work. Actual testing requires real devices, which we don't have.

poncovka (approved):
Thanks!

PR #3435, opened 2021-06-09
Remove unused scripts

VladimirSlavik:
I'm not sure these are used. Feel free to prove me wrong.

Tangentially related to: [rhbz#1938677](https://bugzilla.redhat.com/show_bug.cgi?id=1938677), #3277


poncovka (approved):

VladimirSlavik:
@M4rtinK @rvykydal can you please check that you don't use these tools?

jkonecny12:
You are definitely free to remove the first `merge-pr` script. However, about the second one I have two "objections":
- it's distributed
- seems usable to me (if it works) - could be beneficial to be able to add another file to an existing updates image

poncovka:
> You are definitely free to remove the first `merge-pr` script. However, about the second one I have two "objections":
> 
>     * it's distributed

I don't care. Is it documented?
 
>     * seems usable to me (if it works) - could be beneficial to be able to add another file to an existing updates image

It can live in `devel-tools` if you insist on keeping it.

VladimirSlavik:
I think `upd-updates` should go.
- As far as I can tell, it's a developer thing, you use it to *update* existing `updates.img` contents with newer files or versions of files. However it has none of the clever logic in `makeupdates`, so it will "not work" for anaconda files, it just plops them into same path as input.
- I thought it might be borderline useful for files out in the system, but nooo, for things in `/etc` it fails with an ominous message `cpio: cannot remove current <filename>: Permission denied`.
- It checks "are we the git repo" with a totally outdated URL.
- It has a grand total of 5 commits between 2007 and 2008, all by Peter Jones, except for a mass boilerplate fix.
- All of the above fails to explain why it should be considered working, be distributed, and even exist at all.
- The functionality to add external files is somewhat available via merging update images, which I wrote an year back or so for `anaconda_updates`. You can hand-create an updates image with external files and have it merged into the result.

VladimirSlavik:
To be clear, I picked that one because it was a pile of covscan failures and @M4rtinK suggested removing it instead. I think if we dug deep enough into all the scripts, we could remove more, this was just low hanging fruit.

VladimirSlavik:
I'll make a follow-up low priority task for adding files in `anaconda_updates`.

PR #3440, opened 2021-06-11
Cleanup unneeded nfs repo mount with rd.live.ram

mmatsuya:
NFS mount can sometimes cause a hang, a delay, or a timeout problem after switching root, since it needs to survive the network reconfiguration in stage2 phase. So, it's better to umount the unneeded NFS mounts before switching root.

When rd.live.ram boot parameter, the stage2 image is copied into memory from the NFS repository.
So, keeping NFS mount is not required after the copy. The NFS repository can be mounted in the stage2 phase of package installation if needed. 

This PR is to add a cleanup dracut hook to umount the NFS repository if rd.live.ram parameter is used.

### DIFF
--- a/dracut/anaconda-nfsrepo-cleanup.sh
+++ b/dracut/anaconda-nfsrepo-cleanup.sh
@@ -1,0 +1,31 @@
+#!/bin/sh
+# NFS mount can sometimes cause a hang, a delay, or a timeout problem after
+# switching root, since it needs to survive the network reconfiguration in
+# stage2 phase. So, it's better to umount the unneeded NFS mounts before
+# switching root.
+
+# When rd.live.ram boot parameter, the stage2 image is copied into memory
+# from the NFS repository. So, keeping NFS mount is not required after
+# the copy. The NFS repository can be mounted in the stage2 phase of package
+# installation if needed.
+
+type getargbool >/dev/null 2>&1 || . /lib/dracut-lib.sh
+
+if getargbool 0 rd.live.ram -d -y live_ram; then
+    while read -r src mnt fs rest || [ -n "$src" ]; do
+        if [ "$mnt" = "/run/install/repo" ]; then
+            if [ "$fs" = "nfs" ] || [ "$fs" = "nfs4" ]; then
+                umount /run/install/repo
+                break
+            fi
+        fi
+        # inst.repo=nfs://dvd.iso case
+        if [ "$mnt" = "/run/install/isodir" ]; then
+            if [ "$fs" = "nfs" ] || [ "$fs" = "nfs4" ]; then
+                umount /run/install/repo
+                umount /run/install/isodir
+                break
+            fi
+        fi
+    done < /proc/mounts
+fi
--- a/dracut/module-setup.sh
+++ b/dracut/module-setup.sh
@@ -50,6 +50,7 @@ install() {
     inst_hook pre-pivot 91 "$moddir/anaconda-copy-prefixdevname.sh"
     inst_hook pre-pivot 95 "$moddir/anaconda-set-kernel-hung-timeout.sh"
     inst_hook pre-pivot 99 "$moddir/save-initramfs.sh"
+    inst_hook cleanup 98 "$moddir/anaconda-nfsrepo-cleanup.sh"
     inst_hook pre-shutdown 50 "$moddir/anaconda-pre-shutdown.sh"
     # kickstart parsing, WOOOO
     inst_hook initqueue/online 11 "$moddir/fetch-kickstart-net.sh"
--- a/dracut/Makefile.am
+++ b/dracut/Makefile.am
@@ -50,7 +50,8 @@ dist_dracut_SCRIPTS = module-setup.sh \
                       anaconda-depmod.sh \
                       find-net-intfs-by-driver \
                       anaconda-ifdown \
-                      driver_updates.py
+                      driver_updates.py \
+                      anaconda-nfsrepo-cleanup.sh
 
 dist_dracut_DATA = driver-updates@.service
 


jstodola:
The problem with network re-configuration and hang/delay is not reproducible without `rd.live.ram=1` on the kernel command line and install.img mounted via nfs?
I'm wondering if this change fixes only the special case when `rd.live.ram=1` is used, but the users not using `rd.live.ram=1` might still hit the original problem.

mmatsuya:
I saw the problem while I'm working on a customer case. And, I could not reproduce it by myself, so I cannot confirm that here now. But, I think that it makes worse situation without rd.live.ram=1, since install.img is not copied into memory, and it's mounted as rootfs on stage2. 

When any process read a file, which makes an actual access to the NFS server, the whole system can hang up and it's not possible to enter the background shell. I saw the same situation in several customer's environments before this.  I recommended them to use rd.live.ram, and it worked in RHEL7 in almost all cases. But, I had another problem in RHEL8. systemd creates another file system namespace for some services like systemd-localed, meaning that systemd tries to mount all of existing file system on a special directory (/run/systemd/unit-root) before calling execve system call. Even when rd.live.ram is specified, statfs() and mount() system call is executed for the existing NFS file system, and systemd-localed got a timeout problem during installation. Really, it's better not to have any unneeded NFS mount before switching root.

> I'm wondering if this change fixes only the special case when rd.live.ram=1 is used, but the users not using rd.live.ram=1 might still hit the original problem.

Yes, I think the users not using rd.live.ram=1 may hit a problem. But, hopefully, I wanted to provide an effective workaround of rd.live.ram for them. But, as I wrote above, it doesn't work in RHEL8 with rd.live.ram. I wanted to change it with this PR.

rvykydal:
I've started working on it.

dracut/module-setup.sh
```diff
@@ -50,6 +50,7 @@ install() {
     inst_hook pre-pivot 91 "$moddir/anaconda-copy-prefixdevname.sh"
     inst_hook pre-pivot 95 "$moddir/anaconda-set-kernel-hung-timeout.sh"
     inst_hook pre-pivot 99 "$moddir/save-initramfs.sh"
+    inst_hook cleanup 99 "$moddir/nfsrepo-cleanup.sh"
```
rvykydal, line 53: Maybe I'd use rather
`inst_hook cleanup 98 "$moddir/anaconda-nfsrepo-cleanup.sh"`
because it is anaconda specific hook and there is also
`99-nfsroot-cleanup.sh` (https://github.com/dracutdevs/dracut/blob/master/modules.d/95nfs/nfsroot-cleanup.sh) hook in installer initramfs, and I think anaconda hook should precede it explicitly based on the number (rather than alphabetical order)

rvykydal:
The patch looks good to me, but it does not take into account nfsiso case, ie entering nfs path to installer iso (https://anaconda-installer.readthedocs.io/en/latest/boot-options.html#inst-repo):
`inst.repo=nfs:[<options>:]<server>:/<path>` where path points to installer dvd iso.
In this case the iso is nfs mounted to `/run/install/isodir` and the tree is loop mounted to `/run/install/repo`. Perhaps the patch could be updated like this for the nfsiso case:
https://github.com/rvykydal/anaconda/commit/6ce67369ef248dbb28853fdd1369e4d85e986042

rvykydal:
I've tested the patch to be working correctly for
`inst.repo=nfs://`
and
`inst.stage2=nfs://`
cases

rvykydal:
> > I'm wondering if this change fixes only the special case when rd.live.ram=1 is used, but the users not using rd.live.ram=1 might still hit the original problem.
> 
> Yes, I think the users not using rd.live.ram=1 may hit a problem. But, hopefully, I wanted to provide an effective workaround of rd.live.ram for them. But, as I wrote above, it doesn't work in RHEL8 with rd.live.ram. I wanted to change it with this PR.
Sounds like a very valuable improvement to me.

dracut/nfsrepo-cleanup.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/sh
```
jkonecny12, line 1: Could you please add explanation why is this needed here as a comment?

You can just copy the description of this PR here. It will be really valuable in the future.

mmatsuya: Thanks Jiri, I will update it.

dracut/module-setup.sh
mmatsuya: Thanks Radek, I changed the file name into anaconda-nfsrepo-cleanup.sh.

dracut/anaconda-nfsrepo-cleanup.sh
```diff
@@ -0,0 +1,31 @@
...
+        if [ "$mnt" = "/run/install/repo" ]; then
+            if [ "$fs" = "nfs" ] || [ "$fs" = "nfs4" ]; then
+                umount /run/install/repo
+                break
+            fi
+        fi
+        # inst.repo=nfs://dvd.iso case
+        if [ "$mnt" = "/run/install/isodir" ]; then
+            if [ "$fs" = "nfs" ] || [ "$fs" = "nfs4" ]; then
+                umount /run/install/repo
```
jkonecny12, line 25: I wonder if this couldn't hurt our future intentions to create source which would use Dracut repo mount? What do you think @poncovka ?
So or so it is just a future plan which may end up differently. I'm just thinking aloud, don't want to block this PR because of that.

dracut/nfsrepo-cleanup.sh
jkonecny12: Thanks a lot!

dracut/module-setup.sh
rvykydal: Don't we need to modify also the line above to:
inst_hook cleanup 98 "$moddir/anaconda-nfsrepo-cleanup.sh" ?

mmatsuya: Oops, I just changed it as well. Thanks!

rvykydal: Thank you for the fix-up.I was also proposing a change of the priority of the script (99->98 .. see my first comment in this thread) - if you agree please update it as well.

Also, could you squash all the commits in this PR into a single one so it is easier to port.

mmatsuya: Hi Radek, 
Sorry, I missed the number change. I updated it just now. Also, I squashed commits into a single one. 
Hopefully, I could do that correctly. Thanks!

dracut/anaconda-nfsrepo-cleanup.sh
poncovka: We should always set up the source defined by `inst.repo` in the stage2 phase, so that should be fine. Which installation source do we try to use with `inst.stage2=nfs://`? I don't think we fall back to `inst.stage2` in that case, so that should be fine as well.

rvykydal: For inst.stage2=nfs:// closest mirror is used (or RedHat CDN for rhel) so I think we are fine.

rvykydal (approved):
Looks good to me, thank you!

dracut/anaconda-nfsrepo-cleanup.sh
jkonecny12: Shouldn't we look on the NFS path in case ``inst.stage2=nfs://`` without `inst.repo` specified? However, when we won't find any mount in Dracut everything should be fine because we will then work with the normal NFS mount? Or we can just don't care in this case and fallback to `Closest mirror` but than we again creating some special use-case. So or so, this have to be discussed later when we focus on the feature in more detail.

PR #3442, opened 2021-06-14
Fix import of Iterable from collections

rvykydal:
Since Python 3.10 the alias was removed from collections.


jkonecny12 (approved):
LGTM!

PR #3447, opened 2021-06-17
Enable running container tests in parallel

jkonecny12:
Running container tests in parallel could be handy. Some use-cases:
- running multiple container-shell
- running container-ci together with container-rpm

With the `:z` it will not work correctly on SELinux enabled systems.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -86,7 +86,7 @@ CI_NAME ?= $(CONTAINER_REGISTRY)/rhinstaller/anaconda-ci
 RPM_NAME ?= $(CONTAINER_REGISTRY)/rhinstaller/anaconda-rpm
 ISO_CREATOR_NAME ?= $(CONTAINER_REGISTRY)/rhinstaller/anaconda-iso-creator
 CI_TAG := $(or $(CI_TAG),$(GIT_BRANCH))
-CI_TEST_ARGS ?= --rm -v $(srcdir):/anaconda:Z
+CI_TEST_ARGS ?= --rm -v $(srcdir):/anaconda:z
 CI_CMD ?= make ci
 
 SKIP_BRANCHING_CHECK ?= "false"


VladimirSlavik:
I am not sure if I understand it correctly. When you write "_will not work on SElinux-enabled systems_", does it mean that the code will not run on such systems anymore, but before it could? If so, I think that change is detrimental. We want SElinux, especially if the code runs on our workers. Alternatively, why would *tests* need SElinux off?

jkonecny12:
> I am not sure if I understand it correctly. When you write "_will not work on SElinux-enabled systems_", does it mean that the code will not run on such systems anymore, but before it could? If so, I think that change is detrimental. We want SElinux, especially if the code runs on our workers. Alternatively, why would _tests_ need SElinux off?

No, sorry for miss-understanding. It will always work on system without/disabled SELinux. This `Z`/`z` is just a hint for SELinux that you can do the mount multiple times or it has to be used just by one container at a time.

VladimirSlavik:
Thanks! The question was about enabled SElinux, though. I think that's the default.

jkonecny12:
Yes, enabled SELinux is default on Fedora based distros. And no, both variants will work. As I said before this is just a hint which will enable you to run multiple containers sharing the same mount point source. 
Right now, you can run only one container at once with the anaconda mount point. After this change you can run multiple containers sharing the anaconda mount point. In case of our GH actions it doesn't matter because we always run only one container but for local run you may want to run together rpm tests with ci tests.

VladimirSlavik:
Ah, it's not on in workers? Then it makes more sense.

So, locally, I'd have to disable SElinux to run the tests? Or just to run more of them?

poncovka:
If it is useful only for our CI, shouldn't we change `CI_TEST_ARGS` on the CI? In general, it doesn't seem to be a good idea to have this enabled for local testing.

jkonecny12:
> Ah, it's not on in workers? Then it makes more sense.
> 
> So, locally, I'd have to disable SElinux to run the tests? Or just to run more of them?

You don't have to disable SELinux. This is just a small tweak to be able to run them in parallel if you have SELinux enabled. In case you have SELinux disabled the `z`,`Z` is just ignored.

jkonecny12:
> If it is useful only for our CI, shouldn't we change `CI_TEST_ARGS` on the CI? In general, it doesn't seem to be a good idea to have this enabled for local testing.

The opposite is true, it's useful *only* when you run the tests locally. As I wrote before, our CI always execute just one container at once.

VladimirSlavik (approved):
LGTM, thank you!

poncovka (approved):
Thanks for the explanation!

VladimirSlavik:
One more thought. Is the commit message actually correct? You're changing Z to z. The commit message says that with z (lowercase) it will not work. So I read: You are making it not work with SElinux. But maybe you wanted to write uppercase Z there? Then the meaning would be reversed, and match your explanation here.

jkonecny12:
KK, the commit message is not correct at all. I'll change that. Sorry for confusion.

jkonecny12:
**UPDATED:**
* Fixed the commit message. Hopefully it won't be that confusing now.

VladimirSlavik (approved):
Thanks, that's clear!

PR #3448, opened 2021-06-17
Remove RPM_TESTS_ARGS support

jkonecny12:
Reasons for removal:
- I don't think anyone ever used this variable.
- Doesn't work reliably after container workflow switch.
- Tests are fast enough here so there is no need to execute just some.


VladimirSlavik (approved):
Looks good to me. Thank you!

poncovka (approved):

PR #3450, opened 2021-06-17
Use C.UTF-8 if the requested locale is unsupported (#1958876)

poncovka:
If there is no support for the requested locale, use `C.UTF-8` as a fallback.
The requested locale is considered to be supported if we are not able to
determine the supported locales due to missing tools.

In the `find_best_locale_match function`, skip langcodes with the POSIX variant
unless the provided locale is also a POSIX variant. Otherwise, the `en_US`
locale would match the `C` langcode and that is not desirable.

Resolves: rhbz#1958876

### DIFF
--- a/pyanaconda/localization.py
+++ b/pyanaconda/localization.py
@@ -119,47 +119,73 @@ def find_best_locale_match(locale, langcodes):
     :return: the best matching langcode from the list of None if none matches
     :rtype: str or None
     """
-    def score_value_pair(locale_value, langcode_value, weight):
-        if locale_value and langcode_value:
-            if locale_value == langcode_value:
-                # match
-                return weight
-            else:
-                # not match
-                return -weight
-        elif langcode_value and not locale_value:
-            # langcode has something the locale doesn't have
-            return -weight
-        return 0
-
+    # Parse the locale.
     if not is_valid_langcode(locale):
         return None
 
-    scores = []
+    locale_parsed = langtable.parse_locale(locale)
 
-    # get score for each langcode
+    # Get a score for each langcode.
+    scores = {}
+
     for langcode in langcodes:
         if not is_valid_langcode(langcode):
-            scores.append((langcode, 0))
+            continue
+
+        langcode_parsed = langtable.parse_locale(langcode)
+
+        # Don't match a non-POSIX locale with a POSIX langcode.
+        if langcode_parsed.variant == "POSIX" and locale_parsed.variant != "POSIX":
+            continue
+
+        score = _evaluate_locales(locale_parsed, langcode_parsed)
+
+        # Matches matching only script or encoding or both are not useful.
+        # The score of 100 requires at least territory to have matched.
+        if score <= 100:
+            continue
+
+        scores[langcode] = score
+
+    # Find the best one.
+    return max(scores.keys(), key=scores.get, default=None)
+
+
+def _evaluate_locales(locale, langcode):
+    """Compare a locale with a langcode.
+
+    :param locale: a parsed locale
+    :param langcode: a parsed langcode
+    :return: a score
+    """
+    return \
+        _evaluate_values(locale.language, langcode.language, 1000) + \
+        _evaluate_values(locale.territory, langcode.territory, 100) + \
+        _evaluate_values(locale.script, langcode.script, 10) + \
+        _evaluate_values(locale.variant, langcode.variant, 10) + \
+        _evaluate_values(locale.encoding, langcode.encoding, 1)
+
+
+def _evaluate_values(locale_value, langcode_value, weight):
+    """Compare a locale value with a langcode value.
+
+    :param locale_value: an item of the parsed locale
+    :param langcode_value: an item of the parsed langcode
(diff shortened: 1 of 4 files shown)


pyanaconda/localization.py
```diff
@@ -119,47 +119,73 @@ def find_best_locale_match(locale, langcodes):
...
+            continue
+
+        score = _evaluate_locales(locale_parsed, langcode_parsed)
+
+        # Matches matching only script or encoding or both are not useful.
+        # The score of 100 requires at least territory to have matched.
+        if score <= 100:
+            continue
+
+        scores[langcode] = score
```
jkonecny12, line 148: I just wonder if there couldn't be duplicate langcode. In that case the original solution would keep them separate. On the other hand even if they would they should have the same score so it shouldn't matter?

pyanaconda/modules/localization/installation.py
```diff
@@ -53,16 +55,50 @@ def name(self):
...
+        able to determine the supported locales due to missing tools.
+
+        :param lang: a value for the LANG locale variable
+        :return: True if the locale is supported, otherwise False
+        """
+        try:
+            output = execWithCapture("locale", ["-a"], root=self._sysroot)
+        except OSError as e:
+            log.warning("Couldn't get supported locales: %s", e)
+            return True
```
jkonecny12, line 87: I would ask on the related bug if this is good assumption or we should proceed otherwise.

jkonecny12 (approved):
From the code perspective it looks good to me. From the solution I would ask on the bug but seems reasonable.

pyanaconda/localization.py
poncovka: Duplicate langcodes will always have the same score.

pyanaconda/modules/localization/installation.py
poncovka: I have added a comment to the bug.

pyanaconda/localization.py
```diff
@@ -119,47 +119,73 @@ def find_best_locale_match(locale, langcodes):
...
-                    score_value_pair(locale_parsed.variant, langcode_parsed.variant, 10) + \
-                    score_value_pair(locale_parsed.encoding, langcode_parsed.encoding, 1)
-            scores.append((langcode, score))
-
-    # find the best one
-    sorted_langcodes = sorted(scores, key=lambda item_score: item_score[1], reverse=True)
-
-    # matches matching only script or encoding or both are not useful
-    if sorted_langcodes and sorted_langcodes[0][1] > 100:
-        # 100 = requires at least territory to have matched
```
VladimirSlavik, line 159: Are you dropping the minimum because the C locale won't have a territory?

pyanaconda/modules/localization/installation.py
```diff
@@ -53,16 +55,50 @@ def name(self):
...
+        :return: a locale
+        """
+        if self._is_language_supported(self._lang):
+            return self._lang
+
+        log.debug("The '%s' locale is unsupported.", self._lang)
+        log.debug("Using the '%s' locale as a fallback.", self.LOCALE_FALLBACK)
+        return self.LOCALE_FALLBACK
+
+    def _is_language_supported(self, lang):
```
VladimirSlavik, line 74: I'm not sure if the name is specific enough. I went to check if this does not duplicate something from localization, but then realized this is specific to the new sysroot.

pyanaconda/modules/localization/installation.py
```diff
@@ -53,16 +55,50 @@ def name(self):
...
+        """Is the language configuration supported?
+
+        The language is considered to be supported if we are not
+        able to determine the supported locales due to missing tools.
+
+        :param lang: a value for the LANG locale variable
+        :return: True if the locale is supported, otherwise False
```
VladimirSlavik, line 81: In light of the discussion below (True if can't call `locale -a`), maybe this should go the other way - "False if locale is known to be not supported, True otherwise." Or "True if locale is supported or can't be detected, ..."

VladimirSlavik (approved):
LGTM, I have only nitpicks. Thanks!

pyanaconda/localization.py
poncovka: I have changed the code, so this condition is used sooner. These langcodes are not added to `scores` anymore.

pyanaconda/modules/localization/installation.py
poncovka: Yep, I can rename it.

pyanaconda/modules/localization/installation.py
poncovka: Good idea!

pyanaconda/modules/localization/installation.py
poncovka: Fixed.

pyanaconda/modules/localization/installation.py
poncovka: Fixed.

VladimirSlavik (approved):
Thank you, all resolved for me.

PR #3459, opened 2021-06-22
Call Lorax with the Resilient Storage repository (#infra)

poncovka:
The Lorax templates require the `gfs2-utils` package that is provided by the the Resilient Storage repository.


rvykydal (approved):
Looks good to me, thank you.

VladimirSlavik (approved):
Looks good to me. Thank you!

jkonecny12:
Are we sure that it's on purpose? The point is that we have to add another repository to build the boot.iso so probably even users do which doesn't seem like something which should happen.

poncovka:
This will be fixed with https://github.com/weldr/lorax/pull/1158.

PR #3469, opened 2021-06-24
Fix time setting for daylight savings, let Python do the work

AdamWill:
Since the port from pytz to zoneinfo in #3167, anaconda has set
the system clock one hour too fast when installing for a timezone
that does daylight savings time, during daylight savings time.

There turns out to be a lot of history behind this calculation,
see https://bugzilla.redhat.com/show_bug.cgi?id=1965718#c4 for
details.

It turns out that since Python 3.3, datetime objects have the
`timestamp()` method, which does exactly what we want here. In
my tests it gives the correct answer in every problematic case,
including the one I found was broken in current code, and the
Kolkata and Aden timezones mentioned in
https://bugzilla.redhat.com/show_bug.cgi?id=1293314 . So I think
we should just use that.

Resolves: rhbz#1965718

### DIFF
--- a/dockerfile/anaconda-ci/Dockerfile
+++ b/dockerfile/anaconda-ci/Dockerfile
@@ -43,7 +43,8 @@ RUN pip install --no-cache-dir \
   coverage \
   pycodestyle \
   dogtail \
-  rpmfluff
+  rpmfluff \
+  freezegun
 
 RUN mkdir /anaconda
 
--- a/pyanaconda/isys/__init__.py
+++ b/pyanaconda/isys/__init__.py
@@ -105,14 +105,7 @@ def set_system_date_time(year=None, month=None, day=None, hour=None, minute=None
     second = second if second is not None else now.second
 
     set_date = datetime.datetime(year, month, day, hour, minute, second, tzinfo=tz)
-
-    # Calculate the number of seconds between this time and timestamp 0
-    # pylint bug here: https://github.com/PyCQA/pylint/issues/1104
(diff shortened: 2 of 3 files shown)


VladimirSlavik:
I'd take testing of this, but I have a lot of stuff already. Feel free to assign this to me, but only if this can wait.

AdamWill:
It is breaking two openQA tests; I could set those specific tests to use an updates.img but it's a bit awkward. If we don't fix them, we won't notice when anything *else* breaks in the things those tests test.

AdamWill:
OK, so I changed this to be even simpler: there is actually a `timestamp()` method for datetime objects since Python 3.3 that does precisely what we want, and seems to give the correct result in every case. I don't think we need to be compatible with Python < 3.3 any more, so let's just use that.

I also added tests, since we didn't have any. I have checked that these tests pass with both my original fix and this newer simpler fix, and the EST-during-DST tests fail as expected with the current code. So I think the test suite behaves correctly and proves that this code should be good.

poncovka:
> I'd take testing of this, but I have a lot of stuff already. Feel free to assign this to me, but only if this can wait.

I will take it, but it will have to wait till Monday. Sorry about that.

VladimirSlavik:
Actually, ditto re Monday. Let's try again then.

AdamWill:
it has unit tests now!

poncovka (approved):
Looks good to me. Thank you for fixing it!

PR #3472, opened 2021-06-24
new window in tmux to tail packaging.log

jjmakin:
Make a new window to tail packaging.log, which can be a useful way to identify problems that could affect a successful install


VladimirSlavik (approved):
Jarrod, thanks for the contribution! This seems OK from the code perspective, so it's just a question of "do we want it", and that's for the rest of the maintainers too.

One point - it would be better to put this on master. The f34 branches won't get any further releases that would matter: No Fedora image respins after GA -> anaconda changes won't get to users. If you have some reasons to want this on f34 specifically, please tell us.

poncovka:
I have changed the base to `master`. @jkonecny12, what do you think about this pull request?

jjmakin:
> Jarrod, thanks for the contribution! This seems OK from the code perspective, so it's just a question of "do we want it", and that's for the rest of the maintainers too.
> 
> One point - it would be better to put this on master. The f34 branches won't get any further releases that would matter: No Fedora image respins after GA -> anaconda changes won't get to users. If you have some reasons to want this on f34 specifically, please tell us.

Thanks Vladimir, master branch is fine, I think I misread one of the contributing docs

M4rtinK:
Personally I like the change as indeed, errors and debugging data (which packages got installed ?) can be found in the packaging log, so making it easier to access could be helpful.

BTW, what happens in installation runs that are not package based (ostree, image deployment, etc.) will the tab be simply empty ?

(While definitely out of scope for this PR, I'm thinking if we perhaps want to add something like a "payload.log" that would always have log entries regarding the current payload/-s being installed regardless of payload type.)

jkonecny12:
> BTW, what happens in installation runs that are not package based (ostree, image deployment, etc.) will the tab be simply empty ?

From the tail main page:
```
       -F     same as --follow=name --retry
       --retry
              keep trying to open a file if it is inaccessible
```
so yes, it will stay there empty. I don't have a problem with that.

VladimirSlavik:
We need an assignee. Who's going to do this?

jkonecny12 (approved):
I was thinking about this a long time already. I don't see a reason why not to have `packaging.log` there.

Thanks for the contribution!

jkonecny12:
Just please, fix the indentation of the tail command.

jkonecny12 (changes requested):
Fix the indentation before merge, please.

jkonecny12:
Also please do a rebase on `master` branch to solve the failing tests. Thanks!

jjmakin (commented):
indentation fix

jjmakin:
Hi @jkonecny12 , thanks for taking this on. I've submitted an indentation fix, although I am inexperienced with git rebase so will probably need your assistance for that.

jkonecny12:
Thanks! Commits were squashed and rebased!

jkonecny12 (approved):
Looks good to me now!

jkonecny12:
I have a vacation and will be back on next Tuesday. However, there won't be any builds before that so it shouldn't be a problem. I'm going to merge this then.

jkonecny12:
Tested and works correctly.

poncovka (approved):

PR #3473, opened 2021-06-24
Use glade tests again

VladimirSlavik:
This brings back glade tests. A few points.

- Translations turned out to be never used in past few years, so I just dropped all of that code.
- There is now a shell script that is called same as the other tests. The price is all the custom options are gone - but they were mostly for the translations, so that's ok.
- The way the code is now laid out is too bizarre for my taste. The method that is discovered and run by `unittest` is `test_*`, which mostly only has a docstring and calls a parent class method to loop over the glade files. The loop then calls `checkGlade`which is overridden in each class and implements the actual testing.
- The way glade files are loaded and stored is now very different. Without the plugin API there is no good way to plop them into the test instances. Instead they're local to the module and the intermediate class `GladeTest` can just grab them within the iteration method, as it's in the same module.
- Glade tests are meant to run same as unit tests, so no #infra because I change Makefile too.
 
I don't like the layout of the code much, so I'd welcome suggestions on how to make this better...

Docstrings are a mess, but it's a draft, I'll fix them only once things are stable.


tests/glade_tests.sh
```diff
@@ -0,0 +1,14 @@
...
+    top_srcdir="$(realpath $(dirname "$0")/..)"
+    . ${top_srcdir}/tests/testenv.sh
+fi
+
+# If no tests were selected by user or makefile, select all of them
+if [ $# -eq 0 ]; then
+    set -- "$top_srcdir"/tests/glade_tests
+fi
+
+exec python3 -m unittest discover -t $top_srcdir -v -b ${UNIT_TESTS_PATTERN:+-k $UNIT_TESTS_PATTERN} -s $@
```
poncovka, line 14: I wonder why glade tests are not part of `unit_tests`.

tests/lib/gladecheck.py
```diff
@@ -78,107 +56,16 @@ def checkGlade(self, glade_tree):
...
-            return doc.strip().split("\n")[0].strip()
-        else:
-            return testcase.test.__class__.__name__
 
-    def wantClass(self, cls):
-        # Make sure we grab all the GladeTest subclasses, and only GladeTest
-        # subclasses, regardless of name.
-        return issubclass(cls, GladeTest) and cls != GladeTest
+def load_glade_trees_from_files():
+    glade_files = testfilelist(lambda x: x.endswith('.glade'))
```
poncovka, line 67: Are you sure this is working? Usually, you need to use `global` to declare that this is a global variable.

tests/lib/gladecheck.py
```diff
@@ -28,24 +28,16 @@
...
-# pylint: disable=import-error
-import nose
 
 from filelist import testfilelist
 
 import logging
-log = logging.getLogger('nose.plugins.glade')
+log = logging.getLogger('unittest')
+
+glade_trees = []
```
poncovka, line 39: I would just set `glade_trees = load_glade_trees_from_files()`, simplify the `load_glade_trees_from_files` function  and remove `gladecheck.load_glade_trees_from_files()` from above.

poncovka (commented):
I don't mind with the suggested changes. The layout could be definitely improved, but let's focus on making the tests runnable again.

tests/lib/gladecheck.py
```diff
@@ -78,107 +56,16 @@ def checkGlade(self, glade_tree):
         """
         pass
 
-    def test_glade_file(self):
+    def check_glade_files(self):
         """Run checkGlade for each glade file."""
-        for tree in self.glade_trees:
+        for tree in glade_trees:
```
poncovka, line 61: Could you raise an error if there are no trees to check, please?

tests/lib/gladecheck.py
VladimirSlavik: The tests actually take noticeable time (hundreds on ms) when run... But your idea below is better.

tests/glade_tests.sh
VladimirSlavik: Originally I *think* the reason was the plugin, it was written pretty dirty and assumed everything is a glade test. They could be now integrated back, but I wonder how good idea that is. The thing being tested is a fundamentally different one.

tests/lib/gladecheck.py
```diff
@@ -78,107 +56,30 @@ def checkGlade(self, glade_tree):
         """
         pass
 
-    def test_glade_file(self):
+    def check_glade_files(self):
```
poncovka, line 59: It should be `_check_glade_files`. Also, I would consider to remove `checkGlade` from this class and explicitly specify it as an argument of this method to make the tests easier to understand. That will allow to have run more checks in one test case.

VladimirSlavik:
Let's do it this way: I added commits, they will be squashed, you pick how far I should actually use them. My end result is no mixin, just a function.

poncovka (approved):
It looks great. Thanks!

jkonecny12:
> Glade tests are meant to run same as unit tests, so no #infra because I change Makefile too.

Honestly, I'm still not decided if PRs like this should be #infra or not. My current decision is that if the commit is changing only files which are *not* part of the tarball, they should be marked as #infra. However, it's not that simple on RHEL where you have to mark them #infra or have a bug related to the changes.

So, when I'm thinking about it, maybe we should mark every commit which is not related to a bug as #infra? What do you think?

jkonecny12 (changes requested):
Looks good to me, however, please remove that VAR from the shell script.

VladimirSlavik:
@jkonecny you wanted to mark this as infrastructure, but the test files are not listed, so it fails. The definition is either imprecise, too strict, or correct and this is not infra-only. However I'd rather not make this depend on more of the other stuff. I think we've dug ourselves deep enough already.

Personally I think this should *not* be #infra, same as before.

jkonecny12:
As I said, I'm still not sure how to handle this correctly. Hopefully we will find the correct way to go.

One possible solution could be to not set the `infrastructure` label. That means that tests won't be started because the failure is expected... :(

jkonecny12 (approved):
Thanks for fixing the variable!

PR #3477, opened 2021-06-30
Use a custom stylesheet to define CentOS-specific stylesheet data

carlwgeorge:
This is a backport of #3179 to the rhel-8 branch.  CentOS is trying to take advantage of a custom stylesheet in centos-logos (see [rhbz#1974149](https://bugzilla.redhat.com/show_bug.cgi?id=1974149) and https://github.com/CentOS/centos-logos/issues/2).  It's not working and if I understand things correctly this is the change we need to enable it.

### DIFF
--- a/data/product.d/centos-stream.conf
+++ b/data/product.d/centos-stream.conf
@@ -29,6 +29,8 @@ default_help_pages =
     centos_help_placeholder.xml
     centos_help_placeholder.xml
 
+custom_stylesheet = /usr/share/anaconda/pixmaps/redhat.css
+
 [Payload]
 enable_closest_mirror = True
 default_source = CLOSEST_MIRROR


jkonecny12:
@poncovka could you please take a look on this PR?

VladimirSlavik (approved):
LGTM, same as original. Testing required of course.

carlwgeorge:
Just wanted to confirm that building anaconda-33.16.5.2-1.el8 with this PR as a patch resolves https://github.com/CentOS/centos-logos/issues/2.

poncovka:
This solution will start to log an error on RHEL installations, because `redhat.css` is not provided by `redhat-logos`, but `centos-logos`. The error is harmless, but it might be confusing. It would be better to change our `centos-stream.conf` file instead of `rhel.conf` to avoid the error message.

redhat_css

carlwgeorge:
Is the same change not going to happen to redhat-logos in 8?  If that's the case of course I can change the PR to what you described.

poncovka:
> Is the same change not going to happen to redhat-logos in 8? If that's the case of course I can change the PR to what you described.

We don't plan to do that in RHEL 8. The overrides for RHEL are part of `anaconda-gtk.css` and there is no reason to change that.

carlwgeorge:
I've reworked this to add custom_stylesheet to the centos-stream.conf file instead.

poncovka (approved):
Tested. It looks good to me. Thanks!

VladimirSlavik (approved):
Looks good to me. Thank you!

jkonecny12 (approved):
Looks good to me too. Thanks @carlwgeorge !

PR #3314, opened 2021-04-26
Add base repo name for CentOS Stream after repository renaming

jkonecny12:
CentOS Stream renamed the repository to make everything consistent. For that we have to keep old repository name for RHEL but also we need the new one for CentOS Stream.

Resolves: rhbz#1946347

Backport of https://github.com/rhinstaller/anaconda/pull/3313.

PR #3478, opened 2021-06-30
Rhel 9 fix centos stream repo name

jkonecny12:
CentOS Stream renamed the repository to make everything consistent. For that we have to keep old repository name for RHEL but also we need the new one for CentOS Stream.

Resolves: rhbz#1955331

Backport of https://github.com/rhinstaller/anaconda/pull/3314

### DIFF
--- a/pyanaconda/core/constants.py
+++ b/pyanaconda/core/constants.py
@@ -57,7 +57,8 @@ BASE_REPO_NAME = "anaconda"
 DEFAULT_REPOS = [productName.split('-')[0].lower(),
                  "fedora-modular-server",
                  "rawhide",
-                 "BaseOS"]
+                 "BaseOS",  # Used by RHEL
+                 "baseos"]  # Used by CentOS Stream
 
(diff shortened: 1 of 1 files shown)


poncovka (approved):
Thanks!

jstodola:
@jkonecny12 CentOS Stream repositories are available, so I think this task is no longer blocked:
http://mirror.stream.centos.org/9-stream/

jkonecny12:
Thanks for the ping. I'll take a look on this soon.

jkonecny12:
> 

I tested it but unfortunately the repo files are not part of the installer environment. Trying to find out why is that hapening.

jkonecny12:
Seems that this was my mistake. I was downloading centos stream from https://composes.stream.centos.org/test/latest-CentOS-Stream/compose/ which is without the repo files but there is also http://mirror.stream.centos.org/9-stream/ where repo file are present.

So tested and it works.

jkonecny12:
Tested locally.

M4rtinK (approved):
Looks good to me as well. :)

PR #3480, opened 2021-06-30
Adapt to Pylint 2.9.0

VladimirSlavik:
This time it's very varied.

### DIFF
--- a/tests/README.rst
+++ b/tests/README.rst
@@ -26,7 +26,7 @@ This will run all the tests, including Python test coverage reports. To run
 just some tests you can pass parameters which will replace the current one. For
 example to run just some unit tests please do this::
 
-    make -f Makefile.am container-ci CI_CMD="make tests-units-only UNIT_TESTS_PATTERN='test_layout_variant_'"
+    make -f Makefile.am container-ci CI_CMD="make tests-unit-only UNIT_TESTS_PATTERN='test_layout_variant_'"
 
 The ``UNIT_TESTS_PATTERN`` variable is passed to `unittest discover -k`_. See
 the documentation for more info.
--- a/pyanaconda/argument_parsing.py
+++ b/pyanaconda/argument_parsing.py
@@ -232,7 +232,7 @@ class AnacondaArgumentParser(ArgumentParser):
             option(self, namespace, val)
         return namespace
 
-    # pylint: disable=arguments-differ
+    # pylint: disable=arguments-renamed
     def parse_args(self, args=None, boot_cmdline=None):
         """
         Like ArgumentParser.parse_args(), but also parses the boot cmdline.
--- a/pyanaconda/core/constants.py
+++ b/pyanaconda/core/constants.py
@@ -53,7 +53,7 @@ SOURCES_DIR = MOUNT_DIR + "/sources"
 BASE_REPO_NAME = "anaconda"
 
 # Get list of repo names witch should be used as base repo
-DEFAULT_REPOS = [productName.split('-')[0].lower(),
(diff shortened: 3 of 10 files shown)


VladimirSlavik:
Pylint fails because container has not been rebuilt yet.

tests/pylint/runpylint.py
```diff
@@ -27,6 +27,7 @@ def __init__(self):
             FalsePositive(r"^E1101.*: FedoraGeoIPProvider._refresh: Instance of 'LookupDict' has no 'ok' member"),
             FalsePositive(r"^E1101.*: HostipGeoIPProvider._refresh: Instance of 'LookupDict' has no 'ok' member"),
             FalsePositive(r"^E1101.*: Method 'PropertiesChanged' has no 'connect' member"),
+            FalsePositive(r"^E1101.*: Instance of '[A-Za-z]+PartitioningModule' has no '(storage_handler|request_handler|send_request)' member"),
```
poncovka, line 30: These are our class members. Please, disable the warning at the specific places in the tests. It might be an error in other cases.

jkonecny12:
> Pylint fails because container has not been rebuilt yet.

It won't  be until this is merged because it won't get through the tests. So after you merge this please start the container rebuild manually.

tests/pylint/runpylint.py
VladimirSlavik: I will fix the commit message, but the rest is correct, I believe. The errors of this kind (no member) come and go: We had code that was "all right", then produced false positives, then no more, then again false positives... with zero change in the code itself and tests always passing. This is just another instance of that.

poncovka: You cannot mark it as a false positive unless we are sure that it is always a false positive. These member are defined only in the `BlivetPartitioningModule` class, other partitioning modules don't have them and will fail if we try to access them. The following code matches your false positive pattern, but it will fail with a traceback:

```
module = AutoPartitioningModule()
handler = module.storage_handler
```

VladimirSlavik: Actually, let's do it -  the errors come from as single test that loops over classes, so it's not as horrible as I thought.

poncovka (approved):
Looks good to me. Thanks!

tests/README.rst
```diff
@@ -26,7 +26,7 @@ This will run all the tests, including Python test coverage reports. To run
 just some tests you can pass parameters which will replace the current one. For
 example to run just some unit tests please do this::
 
-    make -f Makefile.am container-ci CI_CMD="make tests-units-only UNIT_TESTS_PATTERN='test_layout_variant_'"
+    make -f Makefile.am container-ci CI_CMD="make tests-unit-only UNIT_TESTS_PATTERN='test_layout_variant_'"
```
jkonecny12, line 29: :+1:

jkonecny12 (approved):
Looks good to me! Thanks!

tests/pylint/runpylint.py
VladimirSlavik: Ah, sorry, I missed your comment. My point was that such a line of code will be caught by unit tests, because it will not run (quite obviously). Unless the problem is in the code that has no coverage...

VladimirSlavik:
https://github.com/rhinstaller/anaconda/runs/2963441419?check_suite_focus=true

VladimirSlavik:
Needs also follow-up in pinning the pylint version in Dockerfile(s) for RHEL 9.

VladimirSlavik:
> Needs follow-up changes to disable f34 in container builds etc.

#3484 

> Needs also follow-up in pinning the pylint and astroid version(s) in Dockerfile(s) for RHEL 9.

#3490

PR #3482, opened 2021-07-01
Make cppcheck optional

VladimirSlavik:
`cppcheck` is unwanted on the enterprise distros, so let's make it easy to skip it. All that is needed to turn it off is not including the package in the container.

Note: Port to RHEL 9 must also remove the package elsewhere, unless already done.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -97,7 +97,7 @@ ANACONDA_PKG_CHECK_MODULES([LIBARCHIVE], [libarchive >= 3.0.4])
 SHUT_UP_GCC="-Wno-unused-result"
 
 # Add remaining compiler flags we want to use
-CFLAGS="$CFLAGS -Wall -Werror $SHUT_UP_GCC"
+CFLAGS="$CFLAGS -Wall -Werror $SHUT_UP_GCC -fanalyzer"
 
 
 # Perform arch related tests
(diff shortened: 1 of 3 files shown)


davidmalcolm:
To set expectations here:

- the first version of `-fanalyzer` that's worth using is in GCC 11.1 (it had a lot of false positives in GCC 10; pre-releases of GCC 11 also had a fair number)
- It's still just for "early adopters"
- It's stilll just for C code

That said, I'm actively working on improving it, and bug reports are most welcome (e.g. false positives with GCC 11.1 or later)

VladimirSlavik:
David, thank you. I think that's fine:
- The (precious little) C code that we have is mostly ancient, stable and battle-tested.
- We don't treat warnings as errors (IIRC).
- In my local run, `-fanalyzer` did not produce any additional warnings at all.
- If it fails too much, we will just disable that again.
- `cppcheck` still runs in upstream (here) and in covscan, so this is not the only check.

VladimirSlavik:
This doesn't appear to be #infra either.

poncovka (approved):
Looks good to me. Thanks!

jkonecny12 (approved):
Great and elegant solution. Thanks a lot for handling this!

PR #3462, opened 2021-06-22
Rhel 8 add infra check

jkonecny12:
Backport of https://github.com/rhinstaller/anaconda/pull/3429 https://github.com/rhinstaller/anaconda/pull/3466 and https://github.com/rhinstaller/anaconda/pull/3467.

~[Test run](https://github.com/jkonecny12/anaconda/runs/2885238403?check_suite_focus=true).~ New [test run](https://github.com/jkonecny12/anaconda/runs/2896319223?check_suite_focus=true).

PR #3461, opened 2021-06-22
Rhel 9 add infra check

jkonecny12:
Backport of https://github.com/rhinstaller/anaconda/pull/3429 https://github.com/rhinstaller/anaconda/pull/3466 and https://github.com/rhinstaller/anaconda/pull/3467.

~[Test run](https://github.com/jkonecny12/anaconda/runs/2885238403?check_suite_focus=true).~ New [test run](https://github.com/jkonecny12/anaconda/pull/13/checks?check_run_id=2896380258).

PR #3483, opened 2021-07-01
Remove .coveragerc file from tarball (#infra)

jkonecny12:
We don't need to have coverage output in gating. Making the tarball smaller is always a win!

This should be backported to https://github.com/rhinstaller/anaconda/pull/3462 and https://github.com/rhinstaller/anaconda/pull/3461 .


VladimirSlavik (approved):
Thank you!

poncovka (approved):
Thanks!

PR #3485, opened 2021-07-01
Add Satellite registration support

M4rtinK:
Add support to Anaconda to register no only to Red Hat run subscription infrastructure (usually called Hosted Candlepin) but also to custom Red Hat Satellite instances run by customers.

This is implemented by a couple new tasks and and a new library module called `satellite`.

To register to a Satellite instance instead of to Hosted Candlepin the use sets the Satellite URL in the GUI or the `--server-hostname` option of the `rhsm` kickstart command. Then at registration time Anaconda will try to fetch and execute a Satellite provisioning script from this URL.

If this is successful, the registration process will continue but the machine will end up talking to the Satellite instance instead of to Hosted Candlepin.

Also at installation time, execute the provisioning script on the target system after installation, so that the installed system is also properly provisioned for the given Satellite instance as well.

**NOTE:** Most Satellite instances use self signed certificates and only after the provisioning script has been run the chain of trust containing these self signed certificates is established. 

Due to this it is recommended only ever register to Satellite via a trusted network where an attacker can't replace the provisioning script in transit as it by necessity has to be transferred a secure SSL connection being established (as before the script is run no chain of trust to the Satellite instance exists).

**TODO**
- [x] check registration to Satellite from kickstart works as well
- [x] check the system can still work with Satellite and its repositories after installation
- [x] check Satellite provided repos can be used as installation source (needs content added to Satellite test instance)
- [x] check why install time token transfer fails (likely caused by missing entitlements on Satellite test instance)
- [x] rename "Red Hat CDN" to "Satellite" when system is registered to Satellite, roll back to "Red Hat CDN" if unregistered at installation time
- [x] ~find why source spoke status is not updated after un-registration from Satellite~ actually a race condition in payload thread, see details in a comment below
- [x] adjust and fixup unit tests to cover Satellite functionality
- [x] resolve and remove all `#FIXME` items in the code

### DIFF
--- a/pyanaconda/core/constants.py
+++ b/pyanaconda/core/constants.py
@@ -100,6 +100,9 @@ GRAPHICAL_TARGET = 'graphical.target'
 NETWORK_CONNECTION_TIMEOUT = 46  # in seconds
 NETWORK_CONNECTED_CHECK_INTERVAL = 0.1  # in seconds
 
+# Anaconda user agent
+USER_AGENT = "%s (anaconda)/%s" % (productName, productVersion)
+
 # DBus
 DEFAULT_DBUS_TIMEOUT = -1       # use default
 
--- a/pyanaconda/errors.py
+++ b/pyanaconda/errors.py
@@ -23,6 +23,7 @@ from pyanaconda.modules.common.errors.installation import BootloaderInstallation
     InsightsClientMissingError, InsightsConnectError
 from pyanaconda.modules.common.errors.payload import SourceSetupError
 from pyanaconda.modules.common.errors.storage import UnusableStorageError
+from pyanaconda.modules.common.errors.subscription import SatelliteProvisioningError
 from pyanaconda.payload.errors import PayloadInstallError, DependencyError, PayloadSetupError
 
 
@@ -114,6 +115,9 @@ class ErrorHandler(object):
             # Subscription related errors
             InsightsClientMissingError.__name__: self._insightsErrorHandler,
             InsightsConnectError.__name__: self._insightsErrorHandler,
+
+            # Satellite
+            SatelliteProvisioningError.__name__: self._target_satellite_provisioning_error_handler,
 
             # General installation errors.
             NonCriticalInstallationError.__name__: self._non_critical_error_handler,
@@ -212,6 +216,13 @@ class ErrorHandler(object):
             return ERROR_CONTINUE
         else:
             return ERROR_RAISE
+
+    def _target_satellite_provisioning_error_handler(self, exn):
+        message = _("Failed to provision the target system for Satellite.")
+        details = str(exn)
+
+        self.ui.showDetailedError(message, details)
+        return ERROR_RAISE
 
     def _non_critical_error_handler(self, exn):
         message = _("The following error occurred during the installation:"
--- a/pyanaconda/modules/common/errors/subscription.py
+++ b/pyanaconda/modules/common/errors/subscription.py
@@ -37,3 +37,9 @@ class UnregistrationError(AnacondaError):
 class SubscriptionError(AnacondaError):
     """Subscription attempt failed."""
     pass
+
+
+@dbus_error("SatelliteProvisioningError", namespace=ANACONDA_NAMESPACE)
+class SatelliteProvisioningError(AnacondaError):
+    """Failed to provision the installation environment for Satellite."""
+    pass
--- a/pyanaconda/modules/subscription/installation.py
+++ b/pyanaconda/modules/subscription/installation.py
@@ -28,6 +28,8 @@ from pyanaconda.core.subscription import check_system_purpose_set
 from pyanaconda.modules.common.task import Task
 from pyanaconda.modules.common.errors.installation import InsightsConnectError, \
     InsightsClientMissingError, SubscriptionTokenTransferError
+from pyanaconda.modules.common.errors.subscription import SatelliteProvisioningError
+from pyanaconda.modules.subscription import satellite
 
 from pyanaconda.anaconda_loggers import get_module_logger
 log = get_module_logger(__name__)
@@ -244,3 +246,58 @@ class TransferSubscriptionTokensTask(Task):
 
         # transfer the RHSM config file
         self._transfer_file(self.RHSM_CONFIG_FILE_PATH, "RHSM config file")
+
+
+class ProvisionTargetSystemForSatelliteTask(Task):
+    """Provision target system for communication with Satellite.
+
+    If the System gets registered to Satellite at installation time,
+    the provisioning is applied only to the installation environment.
+    This task makes sure it is applied also on the target system.
+
+    Run the appropriate Satellite provisioning script on the target system.
+
+    This should assure the target system has all the needed self certificates
+    installed and rhsm.conf tweaks applied.
+    """
+
+    def __init__(self, provisioning_script):
+        """Create a new task.
+
+        :param str provisioning_script: Satellite provisioning script in string form
+        """
+        super().__init__()
+        self._provisioning_script = provisioning_script
+
+    @property
+    def name(self):
+        return "Provisioning target system for Satellite"
+
+    def run(self):
+        """Provision target system for Satellite.
+
+        First check if we are actually registered to a Satellite instance
+        by checking if we got a provisioning script.
+
+        If not, do nothing.
+
+        If we are registered to a Satellite instance, run the Satellite
+        provisioning script that has been downloaded from the instance previously.
+
+        """
+        if self._provisioning_script:
+            log.debug("subscription: provisioning target system for Satellite")
+            provisioning_success = satellite.run_satellite_provisioning_script(
+                provisioning_script=self._provisioning_script,
+                run_on_target_system=True
+
+            )
+            if provisioning_success:
+                log.debug("subscription: target system successfully provisioned for Satellite")
+            else:
+                raise SatelliteProvisioningError("Satellite provisioning script failed.")
+        else:
+            # lets assume here that no provisioning script == not registered to Satellite
+            log.debug(
+                "subscription: not registered to Satellite, skipping Satellite provisioning."
+            )
--- a/pyanaconda/modules/subscription/runtime.py
+++ b/pyanaconda/modules/subscription/runtime.py
@@ -20,19 +20,27 @@ import json
 import datetime
 from collections import namedtuple
 
-from dasbus.typing import get_variant, Str
+from dasbus.typing import get_variant, Str, get_native
 from dasbus.connection import MessageBus
 from dasbus.error import DBusError
 
 from pyanaconda.core.i18n import _
+from pyanaconda.core.constants import SUBSCRIPTION_REQUEST_TYPE_USERNAME_PASSWORD, \
+    SUBSCRIPTION_REQUEST_TYPE_ORG_KEY
+from pyanaconda.core import util
+from pyanaconda.core.payload import ProxyString
+from pyanaconda.ui.lib.subscription import username_password_sufficient, org_keys_sufficient
 from pyanaconda.modules.common.task import Task
 from pyanaconda.modules.common.constants.services import RHSM
-from pyanaconda.modules.common.constants.objects import RHSM_REGISTER
+from pyanaconda.modules.common.constants.objects import RHSM_REGISTER, RHSM_REGISTER_SERVER, \
+    RHSM_UNREGISTER, RHSM_ATTACH, RHSM_ENTITLEMENT, RHSM_CONFIG, RHSM_SYSPURPOSE
 from pyanaconda.modules.common.errors.subscription import RegistrationError, \
-    UnregistrationError, SubscriptionError
+    UnregistrationError, SubscriptionError, SatelliteProvisioningError
 from pyanaconda.modules.common.structures.subscription import AttachedSubscription, \
     SystemPurposeData
-from pyanaconda.modules.subscription import system_purpose
+from pyanaconda.modules.subscription import system_purpose, satellite
+from pyanaconda.modules.subscription.constants import RHSM_SERVICE_NAME, \
+    SERVER_HOSTNAME_NOT_SATELLITE_PREFIX
 from pyanaconda.anaconda_loggers import get_module_logger
 
 import gi
@@ -176,8 +184,15 @@ class SetRHSMConfigurationTask(Task):
         # - all values need to be string variants
         # - proxy password is stored in SecretData instance and we need to retrieve
         #   its value
+        # - server host name might have a prefix indicating the given URL is not
+        #   a Satellite URL, drop that prefix before setting the value to RHSM
+
+        # drop the not-satellite prefix, if any
+        server_hostname = self._request.server_hostname.removeprefix(
+            SERVER_HOSTNAME_NOT_SATELLITE_PREFIX
+        )
         property_key_map = {
-            self.CONFIG_KEY_SERVER_HOSTNAME: self._request.server_hostname,
+            self.CONFIG_KEY_SERVER_HOSTNAME: server_hostname,
             self.CONFIG_KEY_SERVER_PROXY_HOSTNAME: self._request.server_proxy_hostname,
             self.CONFIG_KEY_SERVER_PROXY_PORT: str(self._request.server_proxy_port),
             self.CONFIG_KEY_SERVER_PROXY_USER: self._request.server_proxy_user,
@@ -310,13 +325,17 @@ class RegisterWithOrganizationKeyTask(Task):
 class UnregisterTask(Task):
     """Unregister the system."""
 
-    def __init__(self, rhsm_unregister_proxy):
+    def __init__(self, rhsm_observer, registered_to_satellite, rhsm_configuration):
         """Create a new unregistration task.
 
-        :param rhsm_unregister_proxy: DBus proxy for the RHSM Unregister object
+        :param rhsm_observer: DBus service observer for talking to RHSM
+        :param dict rhsm_configuration: flat "clean" RHSM configuration dict to restore
+        :param bool registered_to_satellite: were we registered to Satellite ?
         """
         super().__init__()
-        self._rhsm_unregister_proxy = rhsm_unregister_proxy
+        self._rhsm_observer = rhsm_observer
+        self._registered_to_satellite = registered_to_satellite
+        self._rhsm_configuration = rhsm_configuration
 
     @property
     def name(self):
@@ -324,18 +343,29 @@ class UnregisterTask(Task):
 
     def run(self):
         """Unregister the system."""
-        log.debug("subscription: unregistering the system")
+        log.debug("registration attempt: unregistering the system")
         try:
             locale = os.environ.get("LANG", "")
-            self._rhsm_unregister_proxy.Unregister({}, locale)
+            rhsm_unregister_proxy = self._rhsm_observer.get_proxy(RHSM_UNREGISTER)
+            rhsm_unregister_proxy.Unregister({}, locale)
             log.debug("subscription: the system has been unregistered")
         except DBusError as e:
-            log.exception("subscription: failed to unregister: %s", str(e))
+            log.error("registration attempt: failed to unregister: %s", str(e))
             exception_dict = json.loads(str(e))
             # return a generic error message in case the RHSM provided error message
             # is missing
             message = exception_dict.get("message", _("Unregistration failed."))
-            raise UnregistrationError(message) from None
+            raise UnregistrationError(message) from e
+
+        # in case we were Registered to Satellite, roll back Satellite provisioning as well
+        if self._registered_to_satellite:
+            log.debug("registration attempt: rolling back Satellite provisioning")
+            rollback_task = RollBackSatelliteProvisioningTask(
+                rhsm_config_proxy=self._rhsm_observer.get_proxy(RHSM_CONFIG),
+                rhsm_configuration=self._rhsm_configuration
+            )
+            rollback_task.run()
+            log.debug("registration attempt: Satellite provisioning rolled back")
 
 
 class AttachSubscriptionTask(Task):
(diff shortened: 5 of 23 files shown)


M4rtinK:
rhel_90_before_Satellite_registration
Connect to Red Hat screen before registration to a Satellite instance

rhel_90_registered_to_Satellite
Connect to Red Hat screen after registration to a Satellite instance

poncovka (commented):
The only thing that we need in UI is the `IsRegisteredToSatellite` property. Everything else could be done in the module. Yet, there are four new DBus methods and a hundred new lines in `pyanaconda.ui.lib`.

pyanaconda/ui/lib/subscription.py
```diff
@@ -177,6 +178,75 @@ def username_password_sufficient(subscription_request=None):
...
+        log.debug("subscription thread: failed to download Satellite provisioning script")
+        # Failing to download the Satellite provisioning script for a user provided server hostname
+        # is an unrecoverable error (wrong URL or incorrectly configured Satellite instance),
+        # so we end there.
+        return False
+
+    # now run the Satellite provisioning script we just downloaded, so that the installation
+    # environment can talk to the Satellite instance the user has specified via custom
+    # server hostname
+    run_script_task_path = subscription_proxy.RunSatelliteProvisioningScriptWithTask()
```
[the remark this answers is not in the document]
M4rtinK: My idea was to keep the tasks as small as possible, but yeah, lets merge it, the task will not be that bigger.

pyanaconda/modules/subscription/runtime.py
```diff
@@ -580,3 +581,122 @@ def run(self):
...
+        log.debug("subscription: downloading Satellite provisioning script")
+        download_success = satellite.download_satellite_provisioning_script(
+            satellite_url=self._satellite_url,
+            proxy_url=self._proxy_url
+        )
+        if download_success:
+            log.debug("subscription: Satellite provisioning script successfully downloaded")
+        else:
+            message = "Failed to download Satellite provisioning script."
+            raise SatelliteProvisioningError(message) from None
```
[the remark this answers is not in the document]
M4rtinK: I think I saw it in a couple places in other tasks, so I've reused that pattern:

https://github.com/rhinstaller/anaconda/blob/f0df778cc43e2494fdd0fad131aa8643590510ac/pyanaconda/modules/storage/bootloader/installation.py#L142
https://github.com/rhinstaller/anaconda/blob/f0df778cc43e2494fdd0fad131aa8643590510ac/pyanaconda/modules/storage/installation.py

pyanaconda/modules/subscription/satellite.py
```diff
@@ -0,0 +1,161 @@
...
+
+import os
+
+from requests import RequestException
+
+from pyanaconda.core import constants, util
+
+# FIXME: move this to core ? use different user agent
+#        without product version information ?
+from pyanaconda.payload.dnf.utils import USER_AGENT
```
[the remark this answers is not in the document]
M4rtinK: The current implementation does:

`USER_AGENT = "%s (anaconda)/%s" % (productName, productVersion)`

Eq. as far as I can tell nothing payload specific - what about moving the definition somewhere to core so that both payload and subscription code can use that ? 

I would expect all Anaconda interaction with web services to present the same user agent, so that should not be a problem.

pyanaconda/modules/subscription/runtime.py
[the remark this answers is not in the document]
poncovka: You need to be in the `except` block. We use it only when we want to raise a different exception and we have logged a traceback of the original one. Then the new exception will not contain info about the original exception. It helps to clean up logs. https://docs.python.org/3/tutorial/errors.html#exception-chaining

pyanaconda/modules/subscription/subscription.py
```diff
@@ -719,6 +761,77 @@ def parse_attached_subscriptions_with_task(self):
...
+    def download_satellite_provisioning_script_with_task(self):
+        """Download Satellite provisioning script with task.
+
+        :return: a DBus path of an installation task
+        """
+        # construct proxy URL needed by the task from the
+        # proxy data in subscription request (if any)
+        # (it is logical to use the same proxy for provisioning
+        #  script download as for RHSM access)
+        if self.subscription_request.server_proxy_hostname:
```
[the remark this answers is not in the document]
M4rtinK: My idea was to limit the amount of things passed to the task, and there are up to 4 values needed to construct the proxy URL. But I guess I can just pass the subscription request instead, so that removes that concern and the code can be moved.

pyanaconda/ui/lib/subscription.py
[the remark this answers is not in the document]
poncovka: You already have a library of functions that you can test separately.

pyanaconda/ui/lib/subscription.py
```diff
@@ -177,6 +178,75 @@ def username_password_sufficient(subscription_request=None):
...
+    run_script_task_proxy = SUBSCRIPTION.get_proxy(run_script_task_path)
+    try:
+        log.debug("subscription thread: running Satellite provisioning script")
+        task.sync_run_task(run_script_task_proxy)
+        log.debug("subscription thread: Satellite provisioning script has been run")
+        # unfortunatelly the RHSM service apparently does not pick up the changes done
+        # by the provisioning script to rhsm.conf, so we need to restart the RHSM systemd
+        # service, which will make it re-read the config file
+        # FIXME: find out if we can use something less radical than a full systemd service restart
+        util.restart_service(RHSM_SERVICE_NAME)
```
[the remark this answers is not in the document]
M4rtinK: Yeah, why not. :)

pyanaconda/core/constants.py
```diff
@@ -518,6 +518,9 @@ class DisplayModes(Enum):
 # - based on the default 90 second systemd service activation timeout
 RHSM_SERVICE_TIMEOUT = 90.0
 
+# name of the RHSM systemd unit
+RHSM_SERVICE_NAME = "rhsm.service"
```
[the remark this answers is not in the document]
M4rtinK: Yeah, as we move the service restart to task, the constant can stay in the module.

pyanaconda/errors.py
```diff
@@ -110,6 +111,10 @@ def _get_default_mapping(self):
             # Payload DBus errors
             SourceSetupError.__name__: self._payload_setup_handler,
 
+            # Satellite
+            SatelliteProvisioningError.__name__: self._target_satellite_provisioning_error_handler,
+
+
```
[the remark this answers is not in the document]
M4rtinK: Fixed.

pyanaconda/errors.py
```diff
@@ -208,6 +213,13 @@ def _bootloader_error_handler(self, exn):
         else:
             return ERROR_RAISE
 
+    def _target_satellite_provisioning_error_handler(self, exn):
+        message = _("Failed to provision the target system for Satellite.")
```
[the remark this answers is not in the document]
M4rtinK: I think it needs the Satellite specific error message, or is it somehow possible to provide that without a specific distinct error handler ?

pyanaconda/modules/subscription/installation.py
```diff
@@ -24,10 +24,14 @@
 from pyanaconda.core import util
 from pyanaconda.core.constants import RHSM_SYSPURPOSE_FILE_PATH
 from pyanaconda.core.subscription import check_system_purpose_set
+from pyanaconda.core.configuration.anaconda import conf
 
 from pyanaconda.modules.common.task import Task
 from pyanaconda.modules.common.errors.installation import InsightsConnectError, \
     InsightsClientMissingError, SubscriptionTokenTransferError
+from pyanaconda.modules.common.errors.subscription import SatelliteProvisioningError
+
```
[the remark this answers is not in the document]
M4rtinK: Removed.

pyanaconda/modules/subscription/installation.py
```diff
@@ -244,3 +248,50 @@ def run(self):
...
+    @property
+    def name(self):
+        return "Provisioning target system for Satellite"
+
+    def run(self):
+        """Provision target system for Satellite.
+
+        First check if we are actually registered to a Satellite instance.
+
+        If not, do nothing.
```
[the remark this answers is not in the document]
M4rtinK: Good point, adding a log message about provisioning being skipped if not registered to a Satellite instance at this point in time.

pyanaconda/ui/lib/subscription.py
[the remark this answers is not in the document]
M4rtinK: Indeed, that should cover it as well. :)

pyanaconda/ui/lib/subscription.py
```diff
@@ -273,6 +346,19 @@ def register_and_subscribe(payload, progress_callback=None, error_callback=None,
...
 
     if task_path:
+        # Now that we know we can do a registration attempt also check if custom server hostname
+        # is set, which would indicate we are most likely talking to a Satellite instance.
+        # If so, provision the insstallation environment for that Satellite instance.
+        if subscription_request.server_hostname:
+            # if custome rserver hostname is set, attempt to provision the installation
+            # environment for Satellite
+            log.debug("subscription thread: provisioning system for Satellite")
+            if provision_system_for_satellite():
```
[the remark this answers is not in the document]
M4rtinK: I can definitely look into that, but it seems to me such a change is out of scope of the Satellite work and should be done via a separate PR. I'll add an issue to our Jira so it does not get lost.

pyanaconda/modules/subscription/subscription_interface.py
```diff
@@ -202,3 +209,39 @@ def ParseAttachedSubscriptionsWithTask(self) -> ObjPath:
...
+    def RunSatelliteProvisioningScriptWithTask(self) -> ObjPath:
+        """Run Satellite provisioning script.
+
+        :return: a DBus path of an installation task
+        """
+        return TaskContainer.to_object_path(
+            self.implementation.run_satellite_provisioning_script_with_task()
+        )
+
+    def SaveRHSMConfBeforeSatelliteProvisioningWithTask(self) -> ObjPath:
```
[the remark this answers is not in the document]
M4rtinK: Oops, forgot to wire it in. :P Should be in use now. :)

pyanaconda/modules/subscription/subscription.py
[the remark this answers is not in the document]
VladimirSlavik: Sounds like a job for a "structure" type of thing?

(I haven't reviewed the actual code yet, just sampling the comments so far. So I might be wrong.)

pyanaconda/ui/lib/subscription.py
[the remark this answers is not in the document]
M4rtinK: Actually, on a more closer look, it looks doable & will clean up the API and all the interfaces quite a bit. Lets do that. :)

pyanaconda/ui/lib/subscription.py
```diff
@@ -398,7 +484,14 @@ def unregister(payload, overridden_source_type, progress_callback=None, error_ca
                 log.debug("subscription thread: restarting payload after unregistration")
                 _do_payload_restart(payload)
 
-        log.debug("Subscription GUI: unregistration succeeded")
+        # check if we were registered to a Satellite instance and roll that back as well
+        # if needed
+        if subscription_proxy.IsRegisteredToSatellite:
```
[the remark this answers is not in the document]
M4rtinK: I've kept the rollback in separate task but called it directly from `UnregisterWithTask`, to avoid the extra API problem.

M4rtinK:
**The pull request has been updated to V2 and is ready for another review round.**

Thanks for all the feedback so far! :)

pyanaconda/errors.py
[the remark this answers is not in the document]
poncovka: Let's keep it for now and solve this globally.

pyanaconda/modules/subscription/satellite.py
```diff
@@ -0,0 +1,161 @@
...
+    HTTP that client machines ca fetch and execute. This script has minimal dependencies
+    and provisions the machine to be able to talk to the one given Satellite instance
+    by installing it's self signed certificates and adjusting rhsm.conf.
+
+    NOTE: As the script is downloaded over plain HTTP is is adviced to ever only
+          provision machines from a Satellite instance on a trusted network, to
+          avoid the possibility of the provisioning script being tempered with
+          during transit.
+
+
```
poncovka, line 64: Please, remove the extra line.

pyanaconda/modules/subscription/satellite.py
```diff
@@ -0,0 +1,161 @@
...
+        result = session.get(script_url, headers=headers,
+                             proxies=proxies, verify=False,
+                             timeout=constants.NETWORK_CONNECTION_TIMEOUT)
+        if result.ok:
+            provisioning_script = result.text
+            result.close()
+        else:
+            log.debug("subscription: server returned %i code when downloading"
+                      " Satellite provisioning script", result.status_code)
+            result.close()
```
poncovka, line 107: Please, call `return False` after this statement and simplify the code below.

pyanaconda/modules/subscription/satellite.py
```diff
@@ -0,0 +1,161 @@
...
+    if os.path.exists(PROVISIONING_SCRIPT_DOWNLOAD_PATH):
+        log.debug("subscription: executing Satellite provisioning script.")
+        rc = util.execWithRedirect("bash", argv=[PROVISIONING_SCRIPT_DOWNLOAD_PATH], root=sysroot)
+        if rc == 0:
+            log.debug("subscription: satellite provisioning script executed successfully")
+            return True
+        else:
+            log.debug("subscription: satellite provisioning script executed with error")
+            return False
+        return True
```
poncovka, line 157: This is a dead code.

pyanaconda/modules/subscription/subscription.py
```diff
@@ -672,24 +673,9 @@ def unregister_with_task(self):
         # and clear attached subscriptions
         task.succeeded_signal.connect(
             lambda: self.set_attached_subscriptions([]))
```
poncovka, line 675: It might be easier to create a new callback and call all these functions from there.

pyanaconda/ui/lib/subscription.py
```diff
@@ -398,7 +391,14 @@ def unregister(payload, overridden_source_type, progress_callback=None, error_ca
...
+            log.debug("subscription thread: rolling back Satellite provisioning")
+            roll_back_satellite_provisioning()
```
poncovka, line 398: Couldn't we do this in the `UnregisterWithTask` class?

poncovka (commented):
I like it, but it needs some polishing.

pyanaconda/modules/subscription/runtime.py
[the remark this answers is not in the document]
VladimirSlavik: I agree - in case of doubt, go with Pylint. IIRC that's how the `from None` came to be in our code base, actually.

pyanaconda/modules/subscription/runtime.py
```diff
@@ -310,32 +317,47 @@ def run(self):
...
+        self._rhsm_configuration = rhsm_configuration
 
     @property
     def name(self):
         return "Unregister the system"
 
     def run(self):
         """Unregister the system."""
-        log.debug("subscription: unregistering the system")
+        log.debug("subscription thread: unregistering the system")
```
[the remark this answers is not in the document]
M4rtinK: I kept it there to so one can easily differentiate messages from a given subscription run. But I think that could be handled better - let's use "subscription attempt" instead, so one can easily find the sequence of messages for a given subscription run.

pyanaconda/modules/subscription/runtime.py
```diff
@@ -580,3 +602,335 @@ def run(self):
...
+        # before running the Satellite provisioning script we back up the current RHSM config
+        # file state, so that we can restore it if Satellite provisioning rollback become necessary
+        rhsm_config_proxy = self._rhsm_observer.get_proxy(RHSM_CONFIG)
+        backup_task = BackupRHSMConfBeforeSatelliteProvisioningTask(
+            rhsm_config_proxy=rhsm_config_proxy
+        )
+        backup_task.succeeded_signal.connect(
+            lambda: self._config_backup_callback(backup_task.get_result())
+        )
+        backup_task.run_with_signals()
```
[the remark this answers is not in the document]
M4rtinK: Indeed, thanks! :)

pyanaconda/modules/subscription/runtime.py
```diff
@@ -580,3 +602,335 @@ def run(self):
...
+        rhsm_attach_proxy = self._rhsm_observer.get_proxy(RHSM_ATTACH)
+        subscription_task = AttachSubscriptionTask(
+            rhsm_attach_proxy=rhsm_attach_proxy,
+            sla=sla
+        )
+        subscription_task.succeeded_signal.connect(
+            lambda: self._subscription_attached_callback(True)
+        )
+        try:
+            subscription_task.run_with_signals()
```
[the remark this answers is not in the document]
M4rtinK: Oh, right - if it won't get shot down by that exception being raised it will be set. And we would re-raise it anway. :)

pyanaconda/modules/subscription/satellite.py
```diff
@@ -0,0 +1,161 @@
...
+        # NOTE: we explicitely don't verify SSL certificates while
+        #       downloading the provisioning script as the Satellite
+        #       instance will have it's own self signed certs that
+        #       will only be trusted once the provisioning script runs
+        result = session.get(script_url, headers=headers,
+                             proxies=proxies, verify=False,
+                             timeout=constants.NETWORK_CONNECTION_TIMEOUT)
+        if result.ok:
+            provisioning_script = result.text
+            result.close()
```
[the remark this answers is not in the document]
M4rtinK: Nice, let's do that! :)

pyanaconda/modules/subscription/satellite.py
M4rtinK: Removed.

pyanaconda/modules/subscription/satellite.py
M4rtinK: Yep, let's do that. :)

pyanaconda/modules/subscription/satellite.py
M4rtinK: Removed.

pyanaconda/modules/subscription/subscription.py
M4rtinK: Yeah, that should reduce the boilerplate, lets do that.

pyanaconda/ui/lib/subscription.py
M4rtinK: Actually, this looks like a leftover as `UnregisterTask` now handles also Satellite provisioning rollback. So removed. :)

poncovka (changes requested):
The tests are failing.

pyanaconda/modules/subscription/satellite.py
poncovka: The code hasn't changed.

pyanaconda/ui/lib/subscription.py
poncovka: The code hasn't changed.

pyanaconda/modules/subscription/satellite.py
M4rtinK: Should be fixed now.

pyanaconda/ui/lib/subscription.py
M4rtinK: Should be fixed by the latest push as well.

M4rtinK:
Updated the PR with changes based on feedback so far & adjustments for the source spoke, to list "Satellite" as installation source when the system is registered to a Satellite instance:
rhel_90_Satellite_installation_source
rhel_90_Satellite_source_in_source_spoke

Next up is fixing the unit tests as well as implementing any additional feedback.

pyanaconda/modules/subscription/runtime.py
```diff
@@ -580,3 +602,327 @@ def run(self):
...
+        self._registered_to_satellite_callback = registered_to_satellite_callback
+        self._subscription_attached_callback = subscription_attached_callback
+        self._subscription_data_callback = subscription_data_callback
+        self._config_backup_callback = config_backup_callback
+
+    @property
+    def name(self):
+        return "Register and subscribe"
+
+    def _provision_system_for_satellite(self):
```
jkonecny12, line 772: It would be great to split this to even smaller methods. Right now, if I'm reading correctly, here are three unrelated things which shouldn't be so hard to split.
I see that you are only creating proxies and running tasks but IMHO it would be more readable having these abstracted.

pyanaconda/modules/subscription/runtime.py
```diff
@@ -580,3 +602,327 @@ def run(self):
...
+            #        systemd service restart
+            util.restart_service(RHSM_SERVICE_NAME)
+
+        except SatelliteProvisioningError as e:
+            log.debug("registration attempt: Satellite provisioning script run failed")
+            # Failing to run the Satellite provisioning script successfully,
+            # which is an unrecoverable error, so we end there.
+            raise e
+
+    def run(self):
```
jkonecny12, line 851: The same works here. I see that you can split `register_task` work from the rest. Maybe even better to have `_create_register_task` and `_run_register_task`.

pyanaconda/modules/subscription/satellite.py
jkonecny12: Still it is here... Starting to be a bit funny. It will stay here until end of Anaconda :D

pyanaconda/ui/lib/subscription.py
```diff
@@ -240,78 +239,54 @@ def register_and_subscribe(payload, progress_callback=None, error_callback=None,
...
-    else:
-        log.debug("subscription thread: credentials insufficient, skipping registration attempt")
-        error_callback(_("Registration failed due to insufficient credentials."))
-        return
 
-    # try to attach subscription
-    log.debug("subscription thread: attempting to auto attach an entitlement")
-    progress_callback(SubscriptionPhase.ATTACH_SUBSCRIPTION)
-    task_path = subscription_proxy.AttachSubscriptionWithTask()
+    # reigstration and subscripotion is handled by a combined tasks that also
```
jkonecny12, line 263: Typo `reigstration` -> `registration`.

jkonecny12 (commented):
I did a quick look through the code. I'll do a proper review after this PR is ready to be reviewed. Here are my thoughts.

pyanaconda/modules/payloads/source/cdn/cdn.py
```diff
@@ -29,6 +31,13 @@
 class CDNSourceModule(RepoFilesSourceModule):
     """The CDN source payload module."""
 
+    def __init__(self):
+        super().__init__()
+        self._subscription_module = None
+        # if the the Subscriptiopn DBus module is available, connect to it
+        if is_module_available(SUBSCRIPTION):
+            self._subscription_module = SUBSCRIPTION.get_proxy()
```
poncovka, line 39: This can be pretty unsafe.

pyanaconda/modules/payloads/source/cdn/cdn.py
```diff
@@ -44,4 +53,15 @@ def type(self):
...
+        if self._subscription_module:
+            registered_to_satellite = self._subscription_module.IsRegisteredToSatellite
+
+        # if we are currently connected to a Satelite instance, call the source "Satellite"
+        # to make it clear a repo provided by the Satellite instance is used, not the global
+        # Red Hat CDN
+        if registered_to_satellite:
+            return _("Satellite")
+        else:
+            return _("Red Hat CDN")
```
poncovka, line 67: Isn't there a general way to describe both sources since the Payloads module doesn't care at all? This is not a good idea. If you really want to have a full control over the description, we can implement a setter you can call. However, I think that the Installation Source spoke doesn't have to provide this information.

pyanaconda/modules/subscription/satellite.py
```diff
@@ -0,0 +1,161 @@
...
+
+from requests import RequestException
+
+from pyanaconda.core import constants, util
+
+# FIXME: move this to core ? use different user agent
+#        without product version information ?
+from pyanaconda.payload.dnf.utils import USER_AGENT
+
+# maybe move this as well ?
```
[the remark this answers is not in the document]
poncovka: The comment is still there.

poncovka (commented):
Please, don't try to solve every related issue in this pull request. It can be fixed later.

M4rtinK:
Reworked the way the Satellite provisioning script is run by using a temp file to make running it in both installation environment and in target system chroot easier and cleaner. 

Checked that the target system after installation can still talk to Satellite and install packages from its repositories. :)

pyanaconda/modules/subscription/satellite.py
M4rtinK: Should be gone now I guess ? ;-)

pyanaconda/ui/lib/subscription.py
M4rtinK: Fixed. :)

pyanaconda/modules/payloads/source/cdn/cdn.py
M4rtinK: I see similar pattern (checking if module is available and then connecting to it) in a couple places. Is the source somehow specific in making this less safe ? Also, any suggestions for a more safer approach ?

pyanaconda/modules/payloads/source/cdn/cdn.py
M4rtinK: We really need to be specific there - customers using the the Red Hat hosted infrastructure need to see they are indeed using our CDN as the installation source and are getting all the related benefits (global availability, up to date packages, etc.).

On the other hand customers with a Satellite instance very explicitly choose to use their self hosted subscription infrastructure so seeing "Red Hat CDN" as the installation source would be very confusing & could generate significant support burden as it would appear as something went wrong and the Satellite hosted repositories are not in use.

Also, some sources already have dynamic decription (NFS, harddrive), is it that different here ?

pyanaconda/modules/subscription/runtime.py
M4rtinK: I think splitting the proxy stuff to a separate task makes sense. Splitting the rest seems to not add much benefit for a lot of boilerplate as indeed, it's just running tasks in the correct order.

pyanaconda/modules/subscription/runtime.py
M4rtinK: What you see here is a transitional state of the registration task due to payload modularization not being fully complete (eq. other bits and pieces still have to reside elsewhere) and I would like to avoid further refactoring of the register tasks run method until after payload modularization is done.

Then all the things needed for registration can finally reside in the task and can be better re-factored.

pyanaconda/modules/payloads/source/cdn/cdn.py
poncovka: Yes, you see it in the UI, but no one communicates with the UI, so it cannot cause any deadlocks. Modules can communicate with other modules from tasks that run in threads, because that will not block the main thread that is suppose to handle DBus calls. Here you access the Boss and the Subscription module from the main thread, so the Payloads module is not able to process any new DBus calls until these calls are finished. It can work now, but it is a timing bomb.

The source module could have a new DBus method that will allow to set the type of the infrastructure or something, so the source can have all info available when it is asked for a description. You just need to update the value to the right one, but that shouldn't be a problem.

Another solution would be to rewrite these DBus calls and call them asynchronously or replace them with signal handlers, but I don't have time to work on that at this moment.

pyanaconda/modules/payloads/source/cdn/cdn.py
poncovka: The dynamic descriptions are not a problem, but you shouldn't do DBus calls from the main threads of the DBus modules. See my answer above.

poncovka (changes requested):
The tests are still failing. There are some invalid unit tests and a lot of pylint warnings.

pyanaconda/modules/payloads/source/cdn/cdn.py
M4rtinK: Good point, lets simplify the whole thing. As in the end this is just about the spoke status I've all the dynamic dropped the logic from the CDN source class and instead special cased the Spoke status string generation to return "Satellite" if satellite is in use. Should be much simpler and less dynamic than before.

pyanaconda/modules/payloads/source/cdn/cdn.py
M4rtinK: The dynamic logic has been dropped.

pyanaconda/modules/subscription/runtime.py
jkonecny12: Ok, thanks for explanation.

rvykydal:
After Vendy's and Jirka's reviews I don't have any remarks, looks good to me.

M4rtinK:
PR updated - all new and changed DBus tasks should now have full unit test coverage.

poncovka:
> PR updated - all new and changed DBus tasks should now have full unit test coverage.

Hi, there are still some failing unit tests and many pylint warnings.

VladimirSlavik (commented):
Here's what I saw in the code itself - haven't seen the tests yet. Most of that is not really about code, so maybe not so actionable. Sorry.

VladimirSlavik (commented):
It might be useful to split the test file. 1600 lines is a lot.

M4rtinK:
> Hi, there are still some failing unit tests and many pylint warnings.

Still working on the remaining test coverage & Pylint warnings. Once I'm done I'll note that in a comment and drop the draft status for the PR, marking it for hopefully final review round.

pyanaconda/modules/subscription/satellite.py
```diff
@@ -0,0 +1,175 @@
...
+    """
+    # first check we actually have the script
+    if provisioning_script is not None:
+        # next check where to run the script
+        if run_on_target_system:
+            # run in the target system chroot
+            sysroot = conf.target.system_root
+        else:
+            # run in installation environment
+            sysroot = "/"
```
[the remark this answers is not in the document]
M4rtinK: Yeah, lets do that. :)

pyanaconda/modules/common/errors/subscription.py
```diff
@@ -37,3 +37,9 @@ class UnregistrationError(AnacondaError):
 class SubscriptionError(AnacondaError):
     """Subscription attempt failed."""
     pass
+
+
+@dbus_error("SatelliteProvisioningError", namespace=ANACONDA_NAMESPACE)
+class SatelliteProvisioningError(AnacondaError):
+    """Failed to provision the installation environment for Satellite."""
```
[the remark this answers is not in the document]
M4rtinK: I'm also seeing other styles, eq. `pyanaconda/modules/common/errors/storage.py` so I would suggest to leave it for where we port this to Fedora, possibly harmonize it for all the error messages.

pyanaconda/modules/subscription/installation.py
```diff
@@ -244,3 +246,55 @@ def run(self):
 
         # transfer the RHSM config file
         self._transfer_file(self.RHSM_CONFIG_FILE_PATH, "RHSM config file")
+
+
+class ProvisionTargetSystemForSatelliteTask(Task):
+    """Provision target system for Satellite.
```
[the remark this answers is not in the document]
M4rtinK: This is about making sure the installed system will be able to talk to the Satellite instance used for the installation. I'll try to change the formulation to better convey that.

pyanaconda/modules/subscription/installation.py
```diff
@@ -244,3 +246,55 @@ def run(self):
...
+    Satellite provisioning script on the target system.
+
+    This should assure the target system has all the needed self certificates
+    installed and rhsm.conf tweaks applied.
+    """
+
+    def __init__(self, registered_to_satellite, provisioning_script):
+        """Create a new task.
+
+        :param bool registered_to_satellite: is Satellite actually in use ?
```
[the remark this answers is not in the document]
M4rtinK: Yeah, that should be enough, lets do that.

(IIRC originally the script was stored as a local file and this is likely a left over of that.)

pyanaconda/ui/gui/spokes/installation_source.py
```diff
@@ -1081,6 +1107,14 @@ def refresh(self):
         # Update the URL entry validation now that we're done messing with sensitivites
         self._update_url_entry_check()
 
+        # if subscription module is available we might need to refresh the label
+        # of the CDN/Satellite radio button
+        if self._subscription_module:
+            if self.registered_to_satellite:
+                self._cdn_button.set_label(C_("GUI|Software Source", "_Satellite"))
+            else:
+                self._cdn_button.set_label(C_("GUI|Software Source", "Red Hat _CDN"))
```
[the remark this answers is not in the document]
M4rtinK: I'll expand the docstring - basically its about the important disctinction between where the repositories are hosted for "normal" vs "Satellite" registration. 

If subscription module is available we might need to refresh the label
of the CDN/Satellite radio button, so that it properly describes what is providing the repositories available after registration.

For registration to Red Hat hosted infrastructure (also called Hosted Candlepin) the global Red Hat CDN efficiently provides quick access to the repositories to customers across the world over the public Internet.

If registered to a customer Satellite instance, it is the Satellite instance itself that provides the software repositories.

This is an important distinction as Satellite instances are often used in environments not connected to the public Internet, so seeing the installation source being provided by Red Hat CDN which the machine might not be able to reach could be very confusing.

M4rtinK:
> It might be useful to split the test file. 1600 lines is a lot.
Good point - I have split the system purpose & satellite module tests to separate files & I think we can split off more once we port this to Fedora.

M4rtinK:
**Payload thread error handling race condition**
Turns out the Anaconda payload thread status notification is not fully thread safe if an error occurs & the Satellite support code manages to reliably trigger this race condition.

Looking at the code from https://github.com/rhinstaller/anaconda/blob/master/pyanaconda/payload/manager.py#L226:
```
except (OSError, DBusError, PayloadError, DNFManagerError) as e:
    log.error("PayloadError: %s", e)
    self._error = self.ERROR_SETUP
    self._set_state(PayloadState.ERROR)
    payload.unsetup()
    return
```

This is called when an error is triggered while the payload refresh thread is running. As you can see, `_set_state()` is called before `payload.usetup()` and only then the thread returns. And this is the problem - spokes update their state based on payload state changes & source spoke will only be accessible if no payload related threads are running.

What happens there is that if Satellite is used as the installation source and the system is unregistered, the payload thread will fail (as Satellite can't
t be used as installation source without registration), trigger a state change callback reporting the error to the source spoke, including re-evaluation of the `ready` property of the source spoke.

But as the `unsetup()` method of the payload thread is still running, the thread will still be running and the source spoke will be marked as not ready, thus inaccessible. 

As there are no other callback after `unsetup()` has been called the source spoke will not "see" the thread finally terminating and will remain inaccessible until the hub is next refreshed (eq. by entering and exiting some other spoke) and the `ready` property is re-evaluated.

Ideas for a fix:
- add a new state to payload that is always the last thing being called
- somehow make the source spoke `join()` the payload thread (can't be done from the handler, as that's running in the context of the thread)
- run `unsetup()` before sending the status callback

In any case, feedback welcome!

M4rtinK:
Still, even with the payload thread issue, the Satellite support PR itself is working fine, all unit tests are passing and Pylint is happy as well.

**So I'm dropping the Draft status for this PR & it is now ready for a final review. :)**

pyanaconda/modules/subscription/satellite.py
```diff
@@ -0,0 +1,176 @@
...
+        sysroot = conf.target.system_root
+    else:
+        # run in installation environment
+        sysroot = "/"
+
+    # create the tempfile containing the script in the sysroot in /tmp, just in case
+    sysroot_tmp = util.join_paths(sysroot, "/tmp")
+    with tempfile.NamedTemporaryFile(mode="w+t", dir=sysroot_tmp, prefix="satellite-") as tf:
+        # write the provisioning script to the tempfile
+        tf.write(provisioning_script)
```
poncovka, line 163: The file is not flushed or closed when you execute the script. It might be incomplete.

pyanaconda/modules/subscription/satellite.py
```diff
@@ -0,0 +1,176 @@
...
+        # We always set root to the correct sysroot, so the script will always
+        # look like it is in /tmp. So just split the randomly generated file name
+        # and combine it with /tmp to get sysroot specific script path.
+        filename = os.path.split(tf.name)[1]
```
poncovka, line 167: Please, use `os.path.basename`. See: https://docs.python.org/3/library/os.path.html#os.path.basename

pyanaconda/modules/subscription/satellite.py
```diff
@@ -0,0 +1,176 @@
...
+        chroot_script_path = os.path.join("/tmp", filename)
+        # and execute it in the sysroot
+        rc = util.execWithRedirect("bash", argv=[chroot_script_path], root=sysroot)
+        if rc == 0:
+            log.debug("subscription: satellite provisioning script executed successfully")
```
poncovka, line 172: Could you please check journal and make sure that the new logging messages make sense? I think there will be many similar lines, that don't say anything new.

pyanaconda/modules/subscription/satellite.py
```diff
@@ -0,0 +1,176 @@
...
+            return True
+        else:
+            log.debug("subscription: satellite provisioning script executed with error")
+            return False
```
poncovka, line 176: Please, try to avoid unnecessary branching in the future. Less code, less bugs.

pyanaconda/ui/gui/spokes/installation_source.py
```diff
@@ -436,6 +436,11 @@ def __init__(self, *args, **kwargs):
         self._network_module = NETWORK.get_proxy()
         self._device_tree = STORAGE.get_proxy(DEVICE_TREE)
 
+        # connect to the Subscription module, if possible
+        self._subscription_module = None
+        if is_module_available(SUBSCRIPTION):
+            self._subscription_module = SUBSCRIPTION.get_proxy()
```
poncovka, line 442: You use this proxy only in the `refresh` method. Otherwise, you always create a new one.

poncovka (commented):
Some of the issues can be addressed in the upstream.

pyanaconda/modules/subscription/runtime.py
```diff
@@ -176,8 +184,18 @@ def run(self):
         # - all values need to be string variants
         # - proxy password is stored in SecretData instance and we need to retrieve
         #   its value
+        # - server hostname might have a prefix indicating the given URL is not#
+        #   a Satellite URL, drop that prefix before setting the value to RHSM
+        if self._request.server_hostname.startswith(SERVER_HOSTNAME_NOT_SATELLITE_PREFIX):
+            # drop the prefix
+            server_hostname = self._request.server_hostname.split(
```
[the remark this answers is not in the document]
M4rtinK: Thanks, that looks much better & simplifies the code, let
s use it! :)

pyanaconda/modules/subscription/runtime.py
```diff
@@ -310,32 +328,47 @@ def run(self):
...
             log.debug("subscription: the system has been unregistered")
         except DBusError as e:
-            log.exception("subscription: failed to unregister: %s", str(e))
+            log.exception("registration attempt: failed to unregister: %s", str(e))
             exception_dict = json.loads(str(e))
             # return a generic error message in case the RHSM provided error message
             # is missing
             message = exception_dict.get("message", _("Unregistration failed."))
-            raise UnregistrationError(message) from None
+            raise UnregistrationError(message) from e
```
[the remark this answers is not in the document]
M4rtinK: Good point, using `log.error`. :)

pyanaconda/modules/subscription/runtime.py
```diff
@@ -580,3 +613,352 @@ def run(self):
...
+        self._provisioning_script = provisioning_script
+
+    @property
+    def name(self):
+        return "Run Satellite provisioning script"
+
+    def run(self):
+        log.debug("subscription: running Satellite provisioning script"
+                  " in installation environment")
+        if satellite.run_satellite_provisioning_script(
```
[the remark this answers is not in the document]
M4rtinK: Yep, that piece of code could look better, lets do that.

pyanaconda/modules/subscription/runtime.py
```diff
@@ -580,3 +613,352 @@ def run(self):
...
+        self._rhsm_observer = rhsm_observer
+        self._subscription_request = subscription_request
+        self._system_purpose_data = system_purpose_data
+        # callback for nested tasks
+        self._registered_callback = registered_callback
+        self._registered_to_satellite_callback = registered_to_satellite_callback
+        self._subscription_attached_callback = subscription_attached_callback
+        self._subscription_data_callback = subscription_data_callback
+        self._satellite_script_downloaded_callback = satellite_script_callback
+        self._config_backup_callback = config_backup_callback
```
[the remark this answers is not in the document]
M4rtinK: Yeah, it's definitely a stopgap for now - I'm sure we will be able to come up with a better solution for upstream, possibly passing out a data structure or something.

pyanaconda/modules/subscription/runtime.py
```diff
@@ -580,3 +613,352 @@ def run(self):
...
+        # (it is logical to use the same proxy for provisioning
+        #  script download as for RHSM access)
+        if subscription_request.server_proxy_hostname:
+            proxy = ProxyString(host=subscription_request.server_proxy_hostname,
+                                username=subscription_request.server_proxy_user,
+                                password=subscription_request.server_proxy_password.value)
+            # only set port if valid in the struct (not -1):
+            if subscription_request.server_proxy_port != -1:
+                # ProxyString expects the port to be a string
+                proxy.port = str(subscription_request.server_proxy_port)
```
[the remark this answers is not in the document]
M4rtinK: OK, lets leave it like it is for now. Might as well look how we use `ProxyString` in other places when porting this upstream and unify those where possible.

pyanaconda/modules/subscription/satellite.py
M4rtinK: Oh, right - we are still inside the context manager and we have to be, so that will not do it for us. :P I think `flush()` should be enough ? I'm a bit afraid if `close()` would perhaps remove the tempfile ? :P

pyanaconda/modules/subscription/satellite.py
M4rtinK: Yep, that looks better, thanks. :)

pyanaconda/modules/subscription/satellite.py
M4rtinK: I've just checked and IMHO it seems reasonably verbose - we can always reduce the verbosity later if all turns out to work fine but for now I would like to keep it as it is.

pyanaconda/ui/gui/spokes/installation_source.py
M4rtinK: Oh, right - I wonder why, maybe some historical reasons ? Anyway, lets use the cached module, is available.

jkonecny12 (approved):
First part of my review. I'll try to do the rest but not sure when I'll be able to get to that.

pyanaconda/ui/gui/spokes/subscription.glade
```diff
@@ -725,10 +725,10 @@
                                 <property name="can_focus">False</property>
                                 <property name="spacing">4</property>
                                 <child>
-                                  <object class="GtkLabel" id="properly_subscribed_label">
+                                  <object class="GtkLabel" id="subscription_status_label">
                                     <property name="visible">True</property>
                                     <property name="can_focus">False</property>
-                                    <property name="label" translatable="yes">The system has been properly subscribed</property>
+                                    <property name="label">""</property>
```
VladimirSlavik, line 731: Should this be untranslatable?

VladimirSlavik (approved):
LGTM, as far as I can understand this...

pyanaconda/modules/subscription/satellite.py
```diff
@@ -0,0 +1,177 @@
...
+                     proxy_url, e)
+
+    with util.requests_session() as session:
+        try:
+            # NOTE: we explicitly don't verify SSL certificates while
+            #       downloading the provisioning script as the Satellite
+            #       instance will most likely have it's own self signed certs that
+            #       will only be trusted once the provisioning script runs
+            result = session.get(script_url, headers=headers,
+                                 proxies=proxies, verify=False,
```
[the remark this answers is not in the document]
M4rtinK: Yes, this is certainly something we can improve in the future - there have been some discussions with the Satellite team for example, about being able to drop in a certificate via a `%pre` script, so that it is possible to fetch the provisioning script over a secure connection. Of course then you ideally also need to provide the kickstart in a way that can't be intercepted and changed by an attacker.

In any case, the default behaviour needs to stay this way for now, due to how most customer Satellite instances are configured by default. We could add an option somewhere to enforce cert checking in case there turn out to be some Satellite instances that do use certificates that the installation images trusts.

pyanaconda/ui/gui/spokes/subscription.glade
M4rtinK: It's now set dynamically at runtime from Python code:
```
        if self._subscription_module.IsRegisteredToSatellite:
            self._subscription_status_label.set_text(
                _("The system is registered to a Satellite instance.")
            )
        else:
            self._subscription_status_label.set_text(
                _("The system is registered.")
            )
```
I guess that should work fine ?

pyanaconda/modules/subscription/runtime.py
```diff
@@ -176,8 +184,15 @@ def run(self):
         # - all values need to be string variants
         # - proxy password is stored in SecretData instance and we need to retrieve
         #   its value
+        # - server hostname might have a prefix indicating the given URL is not#
```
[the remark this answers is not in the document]
M4rtinK: Removed.

pyanaconda/modules/subscription/satellite.py
```diff
@@ -0,0 +1,177 @@
...
+    if run_on_target_system:
+        # run in the target system chroot
+        sysroot = conf.target.system_root
+    else:
+        # run in installation environment
+        sysroot = "/"
+
+    # create the tempfile containing the script in the sysroot in /tmp, just in case
+    sysroot_tmp = util.join_paths(sysroot, "/tmp")
+    with tempfile.NamedTemporaryFile(mode="w+t", dir=sysroot_tmp, prefix="satellite-") as tf:
```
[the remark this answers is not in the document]
M4rtinK: That could help but lets do it when we port this to Fedora. :)

M4rtinK:
>Close shouldn't remove the file. It will just close the descriptor what is what you want before running the script.

Yeah, looks like it works fine wit it, so lets add it. :)

pyanaconda/ui/gui/spokes/subscription.glade
VladimirSlavik: Ah, right, that makes sense, sorry.

pyanaconda/modules/subscription/satellite.py
poncovka: `If delete is true (the default), the file is deleted as soon as it is closed.` See: https://docs.python.org/3/library/tempfile.html#tempfile.NamedTemporaryFile

pyanaconda/modules/subscription/satellite.py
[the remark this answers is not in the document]
jkonecny12: ok, I agree that this should stay as the default and if we want to change it we have to introduce some sort of an API for that.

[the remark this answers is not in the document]
M4rtinK: Yeah, one idea was to look for the certificate in a dropdir or to provide an option to point to the certificate file. In both cases the idea is that a %pre script will drop it in place or possibly it would be part of a custom image built by Image Builder or similar.

But that's all preliminary - we will see what further discussions with the Satellite team will bring.

pyanaconda/modules/subscription/satellite.py
M4rtinK: Oops! Then, let's just keep `flush()` but drop `close()`. :)

Temporary files- there could not be anything complicated about those! ;-)

pyanaconda/modules/subscription/satellite.py
```diff
@@ -0,0 +1,176 @@
...
+    # now that we have something to run, check where to run it
+    if run_on_target_system:
+        # run in the target system chroot
+        sysroot = conf.target.system_root
+    else:
+        # run in installation environment
+        sysroot = "/"
+
+    # create the tempfile containing the script in the sysroot in /tmp, just in case
+    sysroot_tmp = util.join_paths(sysroot, "/tmp")
```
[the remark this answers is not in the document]
M4rtinK: Looks like the folder needs to exist or it throws an exception - thanks for catching this! I've added a `mkdirChain()`call to make sure `/tmp` exists.

poncovka (approved):

M4rtinK:
Marking not just ready-to-merge but also merging now, as discussed with other team members, so that the followup multi-org PR can be cleanly rebased & reviewed on top of rhel-9 branch.

PR #3496, opened 2021-07-07
Don't return None from is_supported_filesystem (#1979854)

poncovka:
The `is_supported_filesystem` function should never return `None`. Otherwise,
the installation might fail with a traceback for an unknown format type.

Resolves: rhbz#1979854


VladimirSlavik (approved):
Looks good to me. Thank you!

PR #3499, opened 2021-07-07
Use C.UTF-8 if the requested locale is unsupported

poncovka:
If there is no support for the requested locale, use C.UTF-8 as a fallback.
The requested locale is considered to be supported if we are not able to
determine the supported locales due to missing tools.

In the `find_best_locale_match function`, skip langcodes with the POSIX variant
unless the provided locale is also a POSIX variant. Otherwise, the `en_US`
locale would match the `C` langcode and that is not desirable.

**Ported from:** https://github.com/rhinstaller/anaconda/pull/3450

### DIFF
--- a/pyanaconda/localization.py
+++ b/pyanaconda/localization.py
@@ -119,47 +119,73 @@ def find_best_locale_match(locale, langcodes):
     :return: the best matching langcode from the list of None if none matches
     :rtype: str or None
     """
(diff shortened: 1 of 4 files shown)


jkonecny12 (approved):
LGTM!

VladimirSlavik (approved):
Thank you for porting this!

PR #3500, opened 2021-07-09
remove authconfig support

pbrezina:
Authconfig compatibility tool (from authselect-compat) will be removed from Fedora 35:
https://fedoraproject.org/wiki/Changes/RemoveAuthselectCompatPackage

### DIFF
--- a/pyanaconda/core/kickstart/commands.py
+++ b/pyanaconda/core/kickstart/commands.py
@@ -22,7 +22,6 @@
 # pylint:disable=unused-import
 
 # Supported kickstart commands.
-from pykickstart.commands.authconfig import F28_Authconfig as Authconfig
 from pykickstart.commands.authselect import F28_Authselect as Authselect
 from pykickstart.commands.autopart import F29_AutoPart as AutoPart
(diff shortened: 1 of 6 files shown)


pbrezina:
CI failed with `test_useless_command_map` is it correct to remove these commands like this (by removing the code) or should I do it differently?

VladimirSlavik:
> CI failed with `test_useless_command_map` is it correct to remove these commands like this (by removing the code) or should I do it differently?

Yes, the tests are expected to do that, you will have to change them too.

VladimirSlavik:
The kickstart tests all succeeded, except for authconfig which must fail. I think that's a success.

(Recording here because these results will disappear, once the fixed unit tests are force-pushed.)

pbrezina:
Can you give me any hint how to update the failing test? I don't know where the 'auth' keyword is comming from.

poncovka:
@pbrezina, the tests require a new pykickstart with the related changes. I will test it locally today.

poncovka (approved):
The code was tested locally. The changes look good to me. Thank you!

PR #3502, opened 2021-07-12
Fix typing errors in the Security module

poncovka:
Fix typing errors in debug messages and doc strings of the Security module.


VladimirSlavik (approved):
Thanks!

M4rtinK (approved):
Looks good to me as well!

PR #3503, opened 2021-07-12
Improve installation logs in the Security module

poncovka:
If an installation task doesn't do anything, make sure we log the reason of that.

### DIFF
--- a/pyanaconda/modules/security/installation.py
+++ b/pyanaconda/modules/security/installation.py
@@ -398,23 +398,27 @@ class ConfigureFingerprintAuthTask(Task):
         return "Configure fingerprint authentication"
 
     def _is_fingerprint_configuration_supported(self):
+        """Is the fingerprint configuration supported?"""
         return (os.path.exists(self._sysroot + PAM_SO_64_PATH) or
                 os.path.exists(self._sysroot + PAM_SO_PATH))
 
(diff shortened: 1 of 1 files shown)


pyanaconda/modules/security/installation.py
```diff
@@ -399,23 +399,27 @@ def name(self):
...
                 os.path.exists(self._sysroot + PAM_SO_PATH))
 
     def run(self):
+        """Run the task."""
         if not self._fingerprint_auth_enabled:
+            log.debug("Fingerprint configuration is not enabled. Skipping.")
             return
 
         if not self._is_fingerprint_configuration_supported():
             log.debug("Fingerprint conifguration is not supported on target system.")
```
VladimirSlavik, line 413: Maybe this line could also say Skipping, same as the case above?

VladimirSlavik (approved):
Looks good to me. Thank you!

M4rtinK (approved):
LGTMAW!

pyanaconda/modules/security/installation.py
poncovka: Fixed.

jkonecny12 (approved):
Looks great to me!

poncovka:
Rebased.

PR #3504, opened 2021-07-12
Monitor the image installation progress with a new class

poncovka:
Use the `InstallationProgress` class to monitor progress of the installation.

**TODO:**

- [x] Test LiveOS installation.
- [x] Test image installation.
- [x] Test tar installation.

### DIFF
--- a/pyanaconda/modules/payloads/payload/live_image/installation_progress.py
+++ b/pyanaconda/modules/payloads/payload/live_image/installation_progress.py
@@ -1,0 +1,123 @@
+#
+# Copyright (C) 2021  Red Hat, Inc.
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# the GNU General Public License v.2, or (at your option) any later version.
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY expressed or implied, including the implied warranties of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General
+# Public License for more details.  You should have received a copy of the
+# GNU General Public License along with this program; if not, write to the
+# Free Software Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
+# 02110-1301, USA.  Any Red Hat trademarks that are incorporated in the
+# source code or documentation are not subject to the GNU General Public
+# License and may only be used or replicated with the express permission of
+# Red Hat, Inc.
+#
+import os
+import time
+
+from blivet.size import Size
+
+from pyanaconda.anaconda_loggers import get_module_logger
+from pyanaconda.core.constants import THREAD_LIVE_PROGRESS
+from pyanaconda.core.i18n import _
+from pyanaconda.core.util import join_paths
+from pyanaconda.modules.common.constants.objects import DEVICE_TREE
+from pyanaconda.modules.common.constants.services import STORAGE
+from pyanaconda.modules.common.task.cancellable import Cancellable
+from pyanaconda.threading import threadMgr, AnacondaThread
+
+log = get_module_logger(__name__)
+
+__all__ = ["InstallationProgress"]
+
+
+class InstallationProgress(Cancellable):
+    """Progress monitor of the image installation."""
+
+    def __init__(self, sysroot, installation_size, callback):
+        """Create a new installation progress.
+
+        :param sysroot: a path to the system root
+        :param installation_size: a size of the installed payload
+        :param callback: a function for the progress reporting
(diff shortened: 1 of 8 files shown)


pyanaconda/modules/payloads/payload/live_image/installation_progress.py
```diff
@@ -0,0 +1,123 @@
...
+        # Wait for the thread to finish.
+        threadMgr.wait(self._thread_name)
+
+    def _monitor_progress(self):
+        """Monitor the amount of disk space used on the target and source."""
+        log.debug("Installing %s.", Size(self._installation_size))
+
+        # Force write everything to disk.
+        self._callback(_("Synchronizing writes to disk"))
+        os.sync()
```
M4rtinK, line 75: It's seems a bit weird to me to assure data is synced out during image install from progress monitoring thread/callback. If it was really just progress monitoring then it should not have side effect like this.

I would suggest either splitting this to a separate data sync callback/thread or renaming the combined callback/thread to something like "synchronization_and_monitoring" to make it evident it's not just read only monitoring without any side effects.

M4rtinK (changes requested):
Looks good to me overall but IMHO the write synchronization should be either decoupled from the monitoring or the callback/thread renamed to mention both synchronization and monitoring.

pyanaconda/modules/payloads/payload/live_image/installation_progress.py
poncovka: The monitoring is based on `os.statvfs` results. The synchronization ensures that the starting size bellow is up to date. If we download the image and don't synchronize the writes, the progress reporting doesn't return correct values, because the size of the downloaded image messes with the currently used size. Therefore, the synchronization is not a side effect, but part of the monitoring.

poncovka:
I was wondering if the progress reporting shouldn't be based on outputs of `rsync` and `tar`. The current solution is universal, but not very precise and probably a little expensive.

pyanaconda/modules/payloads/payload/live_image/installation_progress.py
VladimirSlavik: Just how costly is `os.sync()` ? I assume the writes need to happen at some time anyway, and we don't call this often enough to cause more writes than needed?

VladimirSlavik (approved):
I think it's good. Your point about monitoring output from the tools is a good one, though.

Also, if the thread won't exit, this hangs us. I wonder what could cause that. Somehow, "unexpected terminations" are easier to imagine. Hopefully that reflects the actual state of things.

poncovka:
> Also, if the thread won't exit, this hangs us. I wonder what could cause that. Somehow, "unexpected terminations" are easier to imagine. Hopefully that reflects the actual state of things.

The monitoring is cancelled when the actual installation is finished, so the thread shouldn't hang.

pyanaconda/modules/payloads/payload/live_image/installation_progress.py
poncovka: Yep, the writes will happen anyway.

VladimirSlavik:
Oh yes, but I meant some kind of failure, not how it works when all is ok. But never mind, this is fine.

jkonecny12 (approved):
Looks good to me!

jkonecny12:
> I was wondering if the progress reporting shouldn't be based on outputs of `rsync` and `tar`. The current solution is universal, but not very precise and probably a little expensive.

Are you able to somehow measure how expensive it is?

poncovka:
Rebased.

pyanaconda/modules/payloads/payload/live_image/installation_progress.py
M4rtinK: Oh, the `os.sync()` call is there to keep the `os.statvfs` values up to date for progress monitoring purposes. In that case I'm fine with it being called in the monitoring thread, but please note that also in in the code that it's there for monitoring not data integrity purposes. Thanks! :)

M4rtinK (approved):
Looks good to me. :)

pyanaconda/modules/payloads/payload/live_image/installation_progress.py
```diff
@@ -0,0 +1,123 @@
...
+        """Create a new installation progress.
+
+        :param sysroot: a path to the system root
+        :param installation_size: a size of the installed payload
+        :param callback: a function for the progress reporting
+        """
+        super().__init__()
+        self._sysroot = sysroot
+        self._callback = callback
+        self._installation_size = installation_size
```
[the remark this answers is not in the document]
poncovka: Fixed.

poncovka:
Rebased.

PR #3505, opened 2021-07-13
Remove not required dependencies from anaconda-ci container

jkonecny12:
Remove dependencies used to connect to GitHub as self-hosted runner (we don't support this use-case anymore).
Remove `e2fsprogs` dependency. It seemed to be installed as test dependency but I was able to run the tests locally without it.


VladimirSlavik (approved):
Thank you for the cleanup!

poncovka (approved):
Thank you!

PR #3509, opened 2021-07-14
Use the installation tasks for the image and tar payload

poncovka:
Synchronize the code with the current implementation, clean up the classes
and use the tasks in the liveimg and liveos payload classes. Add some unit tests.


pyanaconda/modules/payloads/payload/live_image/installation.py
```diff
@@ -27,37 +27,50 @@
 class InstallFromTarTask(Task):
     """Task to install the payload from tarball."""
 
-    def __init__(self, tarfile_path, dest_path):
+    def __init__(self, sysroot, tarfile):
+        """Create a new task.
```
VladimirSlavik, line 30: Very tangential, but docstgrings like this are useless. It's init of a thing called Task, I know I'm creating a task. Too bad we have these as copypasta on all tasks, IIRC... At least that means it's a collective blame thing.

On a second thought, maybe the docstring from run() is the one with the real information, in most cases.

VladimirSlavik (approved):
LGTM, thank you!

The one thing that could use some help are the first and last commit messages. I couldn't tell that the first is a temporary disable, then the last actually addresses the same thing and makes it work. Maybe using the same words in the commit messages could help link them for a future reader, or even stating that explicitly? That said, it's not clear if someone will actually review the two commits in the distant future in a way that would require linking them. So YMMV.

poncovka:
> The one thing that could use some help are the first and last commit messages. I couldn't tell that the first is a temporary disable, then the last actually addresses the same thing and makes it work.

Unfortunately, it is more complicated. We don't use the Live OS payload module at this moment. Its installation tasks were used only in tests. The clean-up commits don't enable these tasks in the DBus module. They are just doing some deduplication and run the tasks directly from the UI. The development of the image, tar and Live OS payload modules will be a little chaotic, because these modules are full of outdated half-finished code.

pyanaconda/modules/payloads/payload/live_image/installation.py
poncovka: I use the docsting of the `__init__` method to document arguments of the task. The main description should be documented in the docstring of the class and the implementation and results in the `run` method. There needs to be something written on the first line of each docstring.

I don't understand how would you like to improve it. `Create a new task` is as useless as `Create a new task for installation of the payload from tarball`, because it's init of a thing documented as `Task to install the payload from tarball`.

VladimirSlavik: Hm, true.

poncovka:
Rebased.

VladimirSlavik:
Ah, I get it now, the first commit disables it in the module, the last fixes it still as standalone. Sorry for the confusion.

PR #3510, opened 2021-07-14
Fix COPR chroots in our Dockerfiles 

jkonecny12:
Fedora Rawhide newest container has an issue that it is not successful to detect OS variant when enabling COPR repositories. Without this knowledge the build will fail. Let's make this obvious in our Dockerfiles.


VladimirSlavik (approved):
Looks good to me. Thank you!

PR #2242, opened 2019-12-03
Change the partitioning method selection (#1459930)

VladimirSlavik:
This intends to eventually solve [rhbz#1459930](https://bugzilla.redhat.com/show_bug.cgi?id=1459930) - changing the way users pick the partitioning method.

The GUI form is mostly finalized after consulting with an UX designer. Some functionality might need opinionated input - there are probably choices implicit in the implementation that have not been made consciously enough [by me]. The texts in GUI are very much WIP and need more work.

In other words, feedback is welcome!

I'd like to add that this is limited in scope: Here's one piecemeal fix for what was identified as a particularly bad spot in the current GUI. The maintainer team cannot commit to a larger scope of work at this point. Please take this into consideration when providing feedback.

Replaces #2162.

PR #3511, opened 2021-07-14
Change the root password spoke GUI design

VladimirSlavik:
This resolves ambiguity of what options are available when root is locked.

~~This is a draft because there are at least these TODOs:~~
- [x] make bold labels "clickable" 
- [x] verify the logic
- [x] find bugs etc. to link
- [x] consider renaming spoke to "Root account security"
- [x] ~~add an "empty password" banner~~ automatically switches to no password + locked instead.

See also: https://github.com/rhinstaller/anaconda/pull/3262#issuecomment-839202230
Suggested-by: Mairin Duffy <mairin@redhat.com> @mairin

### DIFF
--- a/pyanaconda/ui/gui/spokes/root_password.py
+++ b/pyanaconda/ui/gui/spokes/root_password.py
@@ -54,7 +54,7 @@ class PasswordSpoke(FirstbootSpokeMixIn, NormalSpoke, GUISpokeInputCheckHandler)
     category = UserSettingsCategory
 
     icon = "dialog-password-symbolic"
-    title = CN_("GUI|Spoke", "_Root Password")
+    title = CN_("GUI|Spoke", "_Root Account")
 
     @classmethod
     def should_run(cls, environment, data):
--- a/pyanaconda/ui/gui/spokes/root_password.py
+++ b/pyanaconda/ui/gui/spokes/root_password.py
@@ -68,8 +68,6 @@ class PasswordSpoke(FirstbootSpokeMixIn, NormalSpoke, GUISpokeInputCheckHandler)
         NormalSpoke.__init__(self, *args)
         GUISpokeInputCheckHandler.__init__(self)
         self._users_module = USERS.get_proxy()
-        self._refresh_running = False
-        self._manually_locked = False
 
     def initialize(self):
         NormalSpoke.initialize(self)
@@ -79,8 +77,10 @@ class PasswordSpoke(FirstbootSpokeMixIn, NormalSpoke, GUISpokeInputCheckHandler)
         self._password_confirmation_entry = self.builder.get_object("password_confirmation_entry")
         self._password_bar = self.builder.get_object("password_bar")
         self._password_label = self.builder.get_object("password_label")
-        self._lock = self.builder.get_object("lock")
+        self._enable_root_radio = self.builder.get_object("enable_root_radio")
+        self._disable_root_radio = self.builder.get_object("disable_root_radio")
         self._root_password_ssh_login_override = self.builder.get_object("root_password_ssh_login_override")
+        self._revealer = self.builder.get_object("password_revealer")
 
         # Install the password checks:
         # - Has a password been specified?
@@ -124,10 +124,6 @@ class PasswordSpoke(FirstbootSpokeMixIn, NormalSpoke, GUISpokeInputCheckHandler)
         self.checker.add_check(self._validity_check)
         self.checker.add_check(self._ascii_check)
 
-        # the password entries should be sensitive on first entry to the spoke
-        self.password_entry.set_sensitive(True)
-        self.password_confirmation_entry.set_sensitive(True)
-
         # Set placeholders if the password has been set outside of the Anaconda
         # GUI we either don't really know anything about it if it's crypted
         # and still would not really want to expose it if its set in plaintext,
@@ -155,31 +151,30 @@ class PasswordSpoke(FirstbootSpokeMixIn, NormalSpoke, GUISpokeInputCheckHandler)
         self.initialize_done()
 
     def refresh(self):
-        # report refresh is running
-        self._refresh_running = True
-        # set the state of the lock checkbox based on DBus data
-        # - set_active() apparently also triggers on_clicked() so
-        #   we use the _refresh_running atribute to differentiate
-        #   it from "real" clicks
-        self._lock.set_active(self._users_module.IsRootAccountLocked)
+        # set the locked/unlocked state based on DBus data
+        if self._users_module.IsRootAccountLocked:
+            control = self._disable_root_radio
+        else:
+            control = self._enable_root_radio
+        control.set_active(True)
+        self.on_root_enabled_changed(control)
+
         self._root_password_ssh_login_override.set_active(
             self._users_module.RootPasswordSSHLoginAllowed
         )
-        if not self._lock.get_active():
+        if self.root_enabled:
             # rerun checks so that we have a correct status message, if any
             self.checker.run_checks()
         # focus on the password field if it is sensitive
         if self.password_entry.get_sensitive():
             self.password_entry.grab_focus()
-        # report refresh finished running
-        self._refresh_running = False
 
     @property
     def status(self):
         if self._users_module.IsRootAccountLocked:
             # reconfig mode currently allows re-enabling a locked root account if
             # user sets a new root password
-            if is_reconfiguration_mode() and not self._lock.get_active():
+            if is_reconfiguration_mode() and not self.root_enabled:
                 return _("Disabled, set password to enable.")
             else:
                 return _("Root account is disabled.")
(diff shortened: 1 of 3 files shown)


VladimirSlavik:
@mairin, please comment on the preferred form of attribution, if you want it different.

VladimirSlavik:
Here's how it looks.
(That shall not change substantially, anymore.)

Screenshot_vs_main_2021-07-14_16:24:05

Screenshot_vs_main_2021-07-14_16:24:12


https://user-images.githubusercontent.com/15903878/125641078-e552f276-c430-48a0-bdca-459347c0b3bf.mp4

mairin:
Wow that looks wonderful, @VladimirSlavik !!! The padding / margins / spacing etc. are right on and the animation looks slick. 🎉

pyanaconda/ui/gui/spokes/root_password.py
```diff
@@ -159,13 +160,16 @@ def refresh(self):
...
-        self._lock.set_active(self._users_module.IsRootAccountLocked)
+        if self._users_module.IsRootAccountLocked:
+            self._disable_radio.set_active(True)
+        else:
+            self._enable_radio.set_active(True)
         self._root_password_ssh_login_override.set_active(
             self._users_module.RootPasswordSSHLoginAllowed
         )
-        if not self._lock.get_active():
+        if self._enable_radio.get_active():
```
poncovka, line 172: Maybe `self._enable_radio.get_active()` could be hidden in a new property.

pyanaconda/ui/gui/spokes/root_password.py
```diff
@@ -179,7 +183,7 @@ def status(self):
         if self._users_module.IsRootAccountLocked:
             # reconfig mode currently allows re-enabling a locked root account if
             # user sets a new root password
-            if is_reconfiguration_mode() and not self._lock.get_active():
+            if is_reconfiguration_mode() and not self._enable_radio.get_active():
```
poncovka, line 186: This needs to be tested with the initial setup.

poncovka (commented):
It looks great!

poncovka:
Please, test it with `rootpw --lock password`. I don't think we need to support this combination in UI, but the kickstart installation should still work. You can also run the `rootpw-lock` kickstart test.

jstodola:
It looks like a great improvement in usability of the spoke!

I would just like to express my dislike for the animation. Although it looks OK when the installation is running locally, animations in the installer in general do not look good when running a remote installation via VNC, where you can see for example just one or two frames from the animation, depending on the network speed. In some case, animations make the installation process even painful, see https://bugzilla.redhat.com/show_bug.cgi?id=1946064. No animation is good animation.

jstodola:
@AdamWill please be aware of this UI change, since it might affect your Fedora installation testing.

poncovka:
Also, add a release note for this change, please (https://anaconda-installer.readthedocs.io/en/latest/contributing.html#release-notes).

VladimirSlavik:
UI logic now works in absence of kickstart, onwards to that...

mairin:
@VladimirSlavik Hey! Taking a look now - 

Re: attribution - looks fine to me :)

Re: alignments... if the change you made to make the alignment intentionally different (e.g. the added indent on the description for each radio button option) prevents the shifting of elements you were worried about, the alignment screenshot you posted looks fine to me! I did examine the GNOME HIG and it does not account for the case where radio buttons have a header message + a non-header description, so this doesn't explicitly violate any design patterns outlined there, and I think it looks fine & clear.

Re: the animation - is there any way you could give me a preview so I could see what it looks like? Probably fade in / fade out will look fine, I just want to see how it comes off!

VladimirSlavik:
Thanks - and here you go!

https://user-images.githubusercontent.com/15903878/126683198-b5e887f1-f41e-450a-a551-6a883657eeb5.mp4

mairin:
@VladimirSlavik looks great to me, I see no issue!

jstodola:
The text boxes for entering the password are too wide - if they are as wide as possible, the screen would not look good on fullhd/4k monitors. Could the size of the password text boxes be reduced?
And one note about enabling/disabling remote access mentioned in the spoke - would it make sense to specify that it only affect remote access for the root account? Remote access for non-root users is not affected by this spoke.

VladimirSlavik:
Thanks for the reminder, I forgot about that part at some point. I changed the width to be 40 characters, not as wide as possible.

I also changed the wording - it's a separate commit so it can be squashed or dropped according to feelings about that.

Screenshot_vs_main_2021-07-23_10:58:09

VladimirSlavik:
I don't think we are tracking any Fedora bugs / issues to link. The RHEL 9 ones are technically not related until this is ported. So linking nothing.

jstodola:
It looks great to me, thanks!

pyanaconda/ui/gui/spokes/root_password.py
```diff
@@ -316,36 +316,32 @@ def _check_spoke_exit_conditions(self, unwaivable_check_failed):
...
-            # this is not a "real" click, just refresh() setting the lock check
-            # box state based on data from the DBus module
-            if not self._manually_locked:
-                # if the checkbox has not yet been manipulated by the user
-                # we can ignore this run and not adjust the fields
-                return True
-        self.password_entry.set_sensitive(not lock.get_active())
-        self.password_confirmation_entry.set_sensitive(not lock.get_active())
-        if not lock.get_active():
+    def on_disable_enable_clicked(self, control):
```
jkonecny12, line 342: This method name seems pretty confusing to me. Right now I'm confused if it should toggle radio buttons or disable some and enable some. Could you please change the name?

pyanaconda/ui/gui/spokes/root_password.py
```diff
@@ -79,8 +77,10 @@ def initialize(self):
         self._password_confirmation_entry = self.builder.get_object("password_confirmation_entry")
         self._password_bar = self.builder.get_object("password_bar")
         self._password_label = self.builder.get_object("password_label")
-        self._lock = self.builder.get_object("lock")
+        self._enable_radio = self.builder.get_object("enable_radio")
+        self._disable_radio = self.builder.get_object("disable_radio")
```
jkonecny12, line 81: I would like to change these variable names. I'm missing the target here. What to enable root account, ssh access, password field?

jkonecny12 (changes requested):
Hi, I really like the new solution, however, I would like to make the variable names a bit more obvious.

poncovka:
> Thanks for the reminder, I forgot about that part at some point. I changed the width to be 40 characters, not as wide as possible.

I think you should limit the width of the whole action area to make it more centred.

pyanaconda/ui/gui/spokes/root_password.py
VladimirSlavik: Sorry, I don't agree. It's an event handler - the name conforms to the `on_controlname_eventtype` pattern most of our code uses. You do not call it, Gtk does. And the next line is a docstring, which also says what it is: _Click handler for enable and disable radio buttons._

VladimirSlavik:
Fun fact of the day: GTK3+ does not have a way to set maximum width. Instead, you have to make sure nothing within the widget of interest will expand. In this case, it was simple, the only expanding thing were the labels. As luck has it, labels, unlike other controls, do have a "maximum width" property... set in characters. Then, the "minimal width" can be set by adding a width request to something. I feel like it's 2000 again and I am trying to vertically center a thing on the web...

Here is 60 characters:

Screenshot_vs_main_2021-07-29_22:02:06

pyanaconda/ui/gui/spokes/root_password.py
VladimirSlavik: I added "root" in the middle of the names, eg. disable_root_radio.

VladimirSlavik:
Update: Limited width, rename radio buttons.

pyanaconda/ui/gui/spokes/root_password.py
jkonecny12: In that case is `disable_enable` a good control name?

VladimirSlavik: The handler is used by two controls, one enables, one disables.

Maybe not mentioning the control names would make the name work? Eg. `on_root_enabled_changed`

jkonecny12: That seems much better to understand to me.

jkonecny12:
I don't agree with the current `enable_disable` but I don't think it's reason for blocking this PR.

jkonecny12 (commented):
I don't agree with the current `enable_disable` but I don't think it's reason for blocking this PR.

pyanaconda/ui/gui/spokes/root_password.py
VladimirSlavik: Great, I will do that, then.

VladimirSlavik: Done.

VladimirSlavik:
Update: Changed the name of the handler per @jkonecny12 suggestion - thank you!

jkonecny12 (approved):

poncovka (approved):

VladimirSlavik:
Update: Renamed to Root Account. The longer version with Security seems to still fit in English, but I wonder about all the translations.

poncovka (approved):
LGTM!

jkonecny12 (approved):
Still looks good to me!

VladimirSlavik:
@AdamWill this will break OpenQA, and #3538 too. I'm about to merge these so the breakage is imminent.

(While at it, I'd like to also get into next Fedora also some rehash of #2242. That would address the easier GUI wants. But that's only hopes so far.)

PR #3512, opened 2021-07-15
Rhel 9 disable built in help system for now

M4rtinK:
This combines a patch from @VladimirSlavik for individual help button hiding with a new config file option that globally disables the built-in help system in Anaconda.

Lastly, use this config file option in the RHEL config file to disable the built-in help system on RHEL 9 for now, due to updated help content not yet being available.

### DIFF
--- a/data/product.d/rhel.conf
+++ b/data/product.d/rhel.conf
@@ -61,6 +61,11 @@ default_help_pages =
     rhel_help_placeholder.xml
     rhel_help_placeholder.xml
 
+# FIXME: The help system is currently disabled
+# as help content has not yet been updated
+# for RHEL 9.
+help_directory =
+
 custom_stylesheet = /usr/share/anaconda/pixmaps/redhat.css
 blivet_gui_supported = False
 
--- a/pyanaconda/ui/gui/__init__.py
+++ b/pyanaconda/ui/gui/__init__.py
@@ -105,6 +105,8 @@ class GUIObject(common.UIObject):
                            given GUI object. The default value of "" indicates
                            that the object has not specific help file assigned
                            and the default help file should be used.
+
+       hide_help_button -- Hide the button for showing help.
     """
     builderObjects = []
     mainWidgetName = None
@@ -116,6 +118,7 @@ class GUIObject(common.UIObject):
 
     uiFile = ""
     helpFile = None
+    hide_help_button = False
     translationDomain = "anaconda"
 
     def __init__(self, data):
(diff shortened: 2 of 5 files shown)


VladimirSlavik (approved):
Looks good to me. Thank you!

poncovka (changes requested):
The help will be disabled only temporarily. I don't think we need a configuration option for that. The simple workaround for products is to not install the help files or to change the help directory to an empty dir.

M4rtinK:
> The help will be disabled only temporarily. I don't think we need a configuration option for that. The simple workaround for products is to not install the help files or to change the help directory to an empty dir.

This will still not be enough - without help content there will still be the Help! buttons in GUI and IIRC also the "h" shortcuts in the TUI, but they will basically try to short an empty screen due to missing help content and place-holders, which would not be a good UX.

What could be done is to check if help content is available for a given screen when it loads and only show the Help! button/"h" shortcut when there is some content. While this requires some runtime activity per every screen, on the other hand it frankly looks like a good default behavior rather than showing a useless placeholder.

And combined with making sure no help content is installed (that should be doable) we should get the needed behaviour without a config file option.

poncovka:
Still, we don't have a long-term user of this feature. If you really want to use configuration files for this, just set the `help_directory` configuration option to an empty string and use it to check if we can show the help or not. The minimization of the API applies also to the configuration files.

M4rtinK:
> Still, we don't have a long-term user of this feature. If you really want to use configuration files for this, just set the `help_directory` configuration option to an empty string and use it to check if we can show the help or not. The minimization of the API applies also to the configuration files.

Yeah, sounds good, I've updated the PR to check if `help_directory` is set instead of that new option. :)

data/product.d/rhel.conf
```diff
@@ -61,6 +61,12 @@ default_help_pages =
     rhel_help_placeholder.xml
     rhel_help_placeholder.xml
 
+
+# The help system is currently disabled
```
VladimirSlavik, line 65: Maybe add a FIXME here?

VladimirSlavik (approved):
LGTM, Thank you!

data/product.d/rhel.conf
M4rtinK: Added (and also fixed the the condition to use properly `help_dir` in rhel.conf :P).

VladimirSlavik (approved):
LGTM again.

VladimirSlavik:
No help = no help buttons feels right. Maybe we could port that to upstream too?

M4rtinK:
> No help files at all = no help buttons feels right. Maybe we could port that to upstream too?
Yeah, I think we should. :)

poncovka (approved):
Looks good to me. Thanks!

jkonecny12 (approved):
Looks good to me!

M4rtinK:
Found one more spoke (Language support) not correctly calling parent `initialize()` but after fixing that all seems fine in both GUI & TUI and in all addon screen. :)

PR #3518, opened 2021-07-16
Handle handle get_layout() method returning None (#1976526)

M4rtinK:
Looks like the GTK Scale object might return None from the
get_layout() method if no label is actually being displayed.

As we only interact with the layout to configure text wrapping,
we can just skip doing anything if None is returned as no text
will be displayed anyway.

Resolves: rhbz#1976526


VladimirSlavik (approved):
Looks good to me. Thank you!

jkonecny12 (approved):
LGTM!

poncovka (approved):

PR #3519, opened 2021-07-16
Raise kickstart errors only during kickstart parsing

poncovka:
We are not able to correctly handle kickstart errors when we don't parse
a kickstart file. For example, we would show a wrong line number.

Resolves: rhbz#1851234


VladimirSlavik (approved):
Looks good to me. Thank you!

jkonecny12 (approved):
Looks good to me.

PR #3520, opened 2021-07-16
Do not install Anaconda dependencies to iso-creator container (#infra)

jkonecny12:
We don't need Anaconda dependencies anymore because Anaconda is now build in a separate step. Now the iso-creator container is used only to build the boot.iso.

This will reduce size of the iso-creator container image from 1.1GB -> less than 400MB (based on my local testing).


VladimirSlavik (approved):
Looks good to me. Thank you!

poncovka (approved):
Thanks!

PR #3530, opened 2021-07-21
Update boot-options.rst

Ultimate-etamitlU:
There was confusion in inst.selinux=0 as,

Original : 

Enable SELinux usage in the installed system (default). Note that when used as a boot option, “selinux” and “inst.selinux” are not the same. The “selinux” option is picked up by both the kernel and Anaconda, but “inst.selinux” is processed only by Anaconda. So when “selinux=0” is used, SELinux will be disabled both in the installation environment and in the installed system, but when “inst.selinux=0” is used SELinux will only be disabled in the installed system. Also note that while SELinux is running in the installation environment by default, it is running in permissive mode so disabling it there does not make much sense.

Change: 

Enable SELinux usage in the installed system (default). Note that when used as a
boot option, "selinux" and "inst.selinux" are not the same. The "selinux" option
is picked up by both the kernel and Anaconda, but "inst.selinux" is processed
only by Anaconda. So when "selinux=0" is used, SELinux will be disabled both in
the installation environment and in the installed system, but when
"inst.selinux=0" is used SELinux will only be disabled in the installation environment.
Also note that while SELinux is running in the installation environment by
default, it is running in permissive mode so disabling it there does not make
much sense.

### DIFF
--- a/docs/boot-options.rst
+++ b/docs/boot-options.rst
@@ -716,7 +716,7 @@ boot option, "selinux" and "inst.selinux" are not the same. The "selinux" option
 is picked up by both the kernel and Anaconda, but "inst.selinux" is processed
 only by Anaconda. So when "selinux=0" is used, SELinux will be disabled both in
 the installation environment and in the installed system, but when
-"inst.selinux=0" is used SELinux will only be disabled in the installed system.
+"inst.selinux=0" is used SELinux will only be disabled in the installation environment.
 Also note that while SELinux is running in the installation environment by
 default, it is running in permissive mode so disabling it there does not make
 much sense.


jkonecny12:
Hi, thanks for the change. Just a small correction, we implemented [this fedora change](https://fedoraproject.org/wiki/Changes/Ignore_Anaconda_kernel_boot_parameters_without_inst_prefix) so we are no more reading or using kernel parameters without `inst.` prefix.

Ultimate-etamitlU:
Yes, we shall not read prefix without "inst." , the change I requested is to the confusing statement in the documentation. Red Hat documentation has the correct definition for those. It's just we need to replace "installed system" to "installed environment" as
"inst.selinux will allow to change mode of installation environment of anaconda and not the actual system which is being getting installed."

VladimirSlavik:
Thank you, the change is probably correct.

FWIW, I think copying the whole paragraph here was a mistake, now we're all hung up on the "we need prefix inst."... ;-)

rvykydal (approved):
Looks good to me, thank you for the patch.

PR #3531, opened 2021-07-21
Don't add a prefix to the dependabot pull requests (#infra)

poncovka:
Add the `(#infra)` suffix once supported as it is required by the infrastructure
check. For now, add the `blocked` label to make sure that the pull request won't
be merged without it. We could create drafts instead of pull requests, but
that's not supported as well.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -14,6 +14,7 @@ updates:
       # Check for updates to GitHub Actions every weekday
       interval: "daily"
     labels:
+      # Don't merge the pull request without #infra.
+      # TODO: Add the (#infra) suffix once supported.
       - "infrastructure"
-    commit-message:
-      prefix: "(#infra)"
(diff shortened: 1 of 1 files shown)


VladimirSlavik (changes requested):
I think the infra check is what should be changed. Who says it must be a suffix?

jkonecny12:
Honestly I'm not sure how to proceed here. One thing is that `(#infra)` could be anywhere on the other hand we are already adding bug numbers as suffix so it's a bit inconsistent to also support prefix here.

VladimirSlavik:
We don't want to merge these, so maybe it does not matter. Feel free to dismiss my nack.

jkonecny12 (approved):

VladimirSlavik (approved):
Let's do it, it doesn't matter if we are not merging the PRs...

PR #3533, opened 2021-07-22
Improve visibility of Dracut errors

jkonecny12:
When there is an error printed in Dracut it is scrolled out pretty fast (most of the time). Users could find them in journal but that is not always an easy thing to do and you have to be an advanced user to know where to look.

By thin change set, I'm trying to improve the situation. Now you are able to see the errors when Dracut starts to timeout:

anaconda_dracut_timetou_warnings

And also after a boot before Anaconda is started:

anaconda_dracut_errors_before_startup

### DIFF
--- a/anaconda.py
+++ b/anaconda.py
@@ -246,6 +246,9 @@ if __name__ == "__main__":
         stdout_log.warning("All Anaconda kernel boot arguments are now required to use "
                            "'inst.' prefix!")
 
+    # print errors encountered during boot
+    startup_utils.print_dracut_errors(stdout_log)
+
     from pyanaconda import isys
 
     util.ipmi_report(constants.IPMI_STARTED)
--- a/dracut/anaconda-error-reporting.sh
+++ b/dracut/anaconda-error-reporting.sh
@@ -1,0 +1,17 @@
+#!/bin/bash
+#
+# Print what went wrong during the boot from installation perspective because Dracut seems
+# to timeout soon!
+#
+
+warn "############# Anaconda installer errors begin #############"
+warn "#                                                         #"
+warn "It seems that the boot has failed. Possible causes include "
+warn "missing inst.stage2 or inst.repo boot parameters on the "
+warn "kernel cmdline. Please verify that you have specified "
+warn "inst.stage2 or inst.repo."
+warn "Please also note that the 'inst.' prefix is now mandatory."
+warn "#                                                         #"
+warn "############# Anaconda installer errors end ###############"
+
+sleep 1
--- a/dracut/anaconda-error-reporting.sh
+++ b/dracut/anaconda-error-reporting.sh
@@ -12,6 +12,12 @@ warn "kernel cmdline. Please verify that you have specified "
 warn "inst.stage2 or inst.repo."
 warn "Please also note that the 'inst.' prefix is now mandatory."
 warn "#                                                         #"
+warn "####     Installer errors encountered during boot:     ####"
+warn "#                                                         #"
+while read -r line; do
+    warn "$line"
+done < /run/anaconda/initrd_errors.txt
+warn "#                                                         #"
 warn "############# Anaconda installer errors end ###############"
 
 sleep 1
--- a/dracut/anaconda-netroot.sh
+++ b/dracut/anaconda-netroot.sh
@@ -83,7 +83,7 @@ case $repo in
         done
     ;;
     *)
-        warn "unknown network repo URL: $repo"
+        warn_critical "unknown network repository URL: $repo"
         return 1
     ;;
 esac
--- a/tests/unit_tests/pyanaconda_tests/core/test_startup_utils.py
+++ b/tests/unit_tests/pyanaconda_tests/core/test_startup_utils.py
@@ -1,0 +1,50 @@
+#
+# Copyright (C) 2021  Red Hat, Inc.
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# the GNU General Public License v.2, or (at your option) any later version.
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY expressed or implied, including the implied warranties of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General
+# Public License for more details.  You should have received a copy of the
+# GNU General Public License along with this program; if not, write to the
+# Free Software Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
+# 02110-1301, USA.  Any Red Hat trademarks that are incorporated in the
+# source code or documentation are not subject to the GNU General Public
+# License and may only be used or replicated with the express permission of
+# Red Hat, Inc.
+#
+# Test code from the startup_utils.py. Most of the parts are not
+# easy to test but we can try to improve that.
+#
+
+import unittest
+
+from unittest.mock import patch, mock_open, Mock
+from textwrap import dedent
+
+from pyanaconda.startup_utils import print_dracut_errors
+
+
+class StartupUtilsTestCase(unittest.TestCase):
+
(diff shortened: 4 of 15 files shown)


dracut/anaconda-error-reporting.sh
```diff
@@ -0,0 +1,23 @@
...
+# to timeout soon!
+#
+
+warn "############# Anaconda installer errors begin #############"
+warn "#                                                         #"
+warn "Seems that the boot has failed. The probable reason could"
+warn "be missing inst.stage2 or inst.repo kernel boot parameters"
+warn "on the kernel cmdline. Please verify that you have"
+warn "inst.stage2 or inst.repo specified."
+warn "Please note that the 'inst.' prefix is now mandatory!"
```
VladimirSlavik, line 13: Some suggestions for the long text.

```suggestion
warn "It seems that the boot has failed. The probable reason could"
warn "be missing 'inst.stage2' or 'inst.repo' boot parameters"
warn "on the kernel cmdline. Please verify that you have"
warn "specified the inst.stage2 or inst.repo boot option."
warn "Please note that the 'inst.' prefix is now mandatory!"
```

Even that's not perfect, but I didn't want to write it all anew. Review by a native speaker could be a good idea.

VladimirSlavik (approved):
I am not sure if this is the best approach, but in my limited knowledge, this seems fine.

I like that you started testing the "toplevel" anaconda code too.

dracut/anaconda-error-reporting.sh
jkonecny12: @snbueno if you have a time, could you please take a look how to improve the message? I suck in this :D

VladimirSlavik: TBH, so do I, usually I'm stuck in a "can't unread" state...

snbueno: Sorry this is not going to be formatted, but this is the error text I would suggest:

> The attempt to boot has failed. Possible causes include missing inst.stage2 or inst.repo boot parameters on the kernel cmdline. Please verify that you have specified inst.stage2 or inst.repo.
> Please also note that the 'inst.' prefix is now mandatory.

I replaced the "It seems..." language because people will only see this if booting has failed, and it is better to use declarative language.

I also just dislike exclamation marks. :joy: But obviously feel free to edit this as you wish. :smiley: 

What do you think?

VladimirSlavik: I'm not sure if the errors are actually always meaningful. It might be that Jirka wanted to express that uncertainty with the "it seems". But that's only a guess, he would have to confirm himself.

+1 to no exclamation marks.

anaconda.py
```diff
@@ -119,6 +119,25 @@ def parse_arguments(argv=None, boot_cmdline=None):
...
+    from pyanaconda.core.constants import DRACUT_ERRORS_PATH
+
+    try:
+        with open(DRACUT_ERRORS_PATH, "rt") as fd:
+            section_name = "Installer errors encountered during boot"
+            msg = "\n{:#^70}\n{}\n{:#^70}".format(  # add starting \n because timestamp
+                " " + section_name + " ",           # start of the section
+                "".join(fd.readlines()),            # errors from Dracut
+                " " + section_name + " end ")       # end of the section
+            stdout_logger.warning(msg)
```
VladimirSlavik, line 136: On a second thought, I am actually not sure if `stdout_logger` is the right kind of output?

dracut/anaconda-error-reporting.sh
jkonecny12: I like this text. And yes, I wanted to express "uncertainty" as you said and I think the `Possible causes...` is exactly that :).

anaconda.py
jkonecny12: I was also thinking about that, reason why I decided to use it is that we are already using it for printing errors on this place. I wanted to be consistent.

dracut/anaconda-error-reporting.sh
jkonecny12: Updated. Thanks a lot for suggestions!

jkonecny12:
**UPDATED**
* Updated error message in Dracut.

jkonecny12:
**UPDATED:**
* Fix placement of the new Dracut script. It should be in `initqueue/timeout` not `timeout`.

poncovka (commented):
I have some suggestions, but I like it in general. Nice and clean!

anaconda.py
```diff
@@ -119,6 +119,25 @@ def parse_arguments(argv=None, boot_cmdline=None):
     return (namespace, ap.removed_no_inst_bootargs)
 
 
+def print_dracut_errors(stdout_logger):
```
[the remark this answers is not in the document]
jkonecny12: Hmm, I had that in `pyanaconda.core.utils` but we want to make that place smaller so I moved it out. Any idea where to place this?

dracut/anaconda-netroot.sh
```diff
@@ -83,7 +83,7 @@ case $repo in
         done
     ;;
     *)
-        warn "unknown network repo URL: $repo"
+        warn_visible "unknown network repository URL: $repo"
```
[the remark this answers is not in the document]
jkonecny12: Sounds good to me.

anaconda.py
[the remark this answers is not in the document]
poncovka: I think that it should be in `pyanaconda.startup_utils`.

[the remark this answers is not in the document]
jkonecny12: Fixed.

jkonecny12:
**UPDATED:**
* Implemented notes from @poncovka . Thanks for suggestions!
* Improved commit messages and comments in the code.

rvykydal (approved):
Looks good to me, thank you.

anaconda.py
```diff
@@ -246,6 +248,9 @@ def setup_environment():
         stdout_log.warning("All Anaconda kernel boot arguments are now required to use "
                            "'inst.' prefix!")
 
+    # print errors encountered during boot
+    print_dracut_errors(stdout_log)
```
poncovka, line 252: Please, call `startup_utils.print_dracut_errors`.

anaconda.py
```diff
@@ -30,6 +30,8 @@
 import signal
 import pid
 
+from pyanaconda.startup_utils import print_dracut_errors
```
poncovka, line 33: Please, remove this import.

tests/unit_tests/pyanaconda_tests/core/test_startup_utils.py
```diff
@@ -0,0 +1,49 @@
...
+#
+# Test code from the startup_utils.py. Most of the parts are not
+# easy to test but we can try to improve that.
+#
+
+import unittest
+
+from unittest.mock import patch, mock_open, Mock
+
+from anaconda import print_dracut_errors
```
poncovka, line 26: Please, import the function from `pyanaconda.startup_utils`.

tests/unit_tests/pyanaconda_tests/core/test_startup_utils.py
```diff
@@ -0,0 +1,49 @@
...
+        with patch("anaconda.open", mock_open(read_data="test text")) as m:
+            print_dracut_errors(logger_mock)
+
+            m.assert_called_once_with("/run/anaconda/initrd_errors.txt", "rt")
+
+            logger_mock.warning.assert_called_once_with(
+                """
+############## Installer errors encountered during boot ##############
+test text
+############ Installer errors encountered during boot end ############""")
```
poncovka, line 43: You can indent the lines of the string and call `dedent` to fix it. The tests are more readable that way.

poncovka (changes requested):

jkonecny12:
**UPDATED:**
* Fix issues I had there and were pointed by @poncovka . Thanks!

poncovka (approved):
It looks great! Thanks!

jkonecny12:
Tested and it works correctly.

PR #3537, opened 2021-07-23
Add root_password_must_be_set config option

M4rtinK:
Add the root_password_must_be_set Anaconda config file option that
requires root password to be set during the given installation run.

This option works for both TUI and GUI and overrides the normal behavior
where setting an admin user is sufficient and where root password
configuration is not enforced for kickstart installation at all.

As this feature was requested by RHV, also enable it in the RHV and
oVirt configuration files when we are at it.

Resolves: rhbz#1981807

### DIFF
--- a/data/anaconda.conf
+++ b/data/anaconda.conf
@@ -267,6 +267,11 @@ hidden_spokes =
 # Run GUI installer in a decorated window.
 decorated_window = False
 
+# do not have hard requirement on root password being set by default
+# NOTE: this is a temporary option introduced to fix a RHV issue
+#       (see rhbz#1981807)
+root_password_must_be_set = False
+
 
 [License]
 # A path to EULA (if any)
--- a/data/product.d/ovirt.conf
+++ b/data/product.d/ovirt.conf
@@ -26,3 +26,6 @@ req_partition_sizes =
 
 [User Interface]
 help_directory = /usr/share/anaconda/help/rhv
+
+# require user to set root password during oVirt installation
+root_password_must_be_set = True
--- a/pyanaconda/ui/lib/users.py
+++ b/pyanaconda/ui/lib/users.py
@@ -16,6 +16,10 @@
 # License and may only be used or replicated with the express permission of
 # Red Hat, Inc.
 #
+from pyanaconda.flags import flags
+from pyanaconda.core.configuration.anaconda import conf
+from pyanaconda.core.i18n import N_, _
+from pyanaconda.ui.lib.services import is_reconfiguration_mode
 from pyanaconda.modules.common.structures.user import UserData
 
 from pyanaconda.anaconda_loggers import get_module_logger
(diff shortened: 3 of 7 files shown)


VladimirSlavik:
LGTM, although I have a nagging feeling that this might be a module thing...

M4rtinK:
To add some back story this is about all RHV installations being effectively kickstart installations, which disables our normal root/admin user creation enforcement code. The new config option solves the situation by requiring root password to be set regardless of any other conditions (kickstart installation, admin user being present).

This solves the issue RHV has while possibly adding a useful functionality for other cases where making user root password is set could be helpful. 

Also, it's possible there are other more elegant ways of doing what RHV needs, so feedback welcome (and also this is why I've marked the PR as draft for now)! :)

M4rtinK:
rhv_root_not_required
This is how it looks like now without any changes - root password configuration is not enforced.
rhv_root_required
With the change root password needs to be set.

data/product.d/ovirt.conf
```diff
@@ -26,3 +26,9 @@ req_partition_sizes =
 
 [User Interface]
 help_directory = /usr/share/anaconda/help/rhv
+
+[Security]
+
+# require user to set root password during oVirt installation
+root_password_must_be_set = True
```
poncovka, line 33: It feels like a UI thing. Modules don't enforce this rule, so I would keep it in the `[User Interface]` section.

pyanaconda/ui/tui/spokes/root_password.py
```diff
@@ -70,8 +78,15 @@ def showable(self):
...
-        return not self._users_module.CheckAdminUserExists()
+        """Only mandatory if no admin user has been requested.
+
+        If root_password_must_be_set is set to True in the config file,
+        root password must be set, regardless of admin user existing or not.
+        """
+        if conf.security.root_password_must_be_set:
+            return not self._users_module.IsRootPasswordSet
+        else:
+            return not self._users_module.CheckAdminUserExists()
```
poncovka, line 89: The checks could be moved to new functions in `pyanaconda.ui.lib.users`.

poncovka (commented):
I am afraid that this is just a beginning of a series of fixes to support this use case. The status doesn't make sense, the checkbox in RHEL 9 doesn't make sense and the new UI for upstream shouldn't allow to disable the root. Is it a temporary workaround or do they really want to enforce the root password in all cases (even when customers specify `rootpw --lock`)?

lveyde:
> I am afraid that this is just a beginning of a series of fixes to support this use case. The status doesn't make sense, the checkbox in RHEL 9 doesn't make sense and the new UI for upstream shouldn't allow to disable the root. Is it a temporary workaround or do they really want to enforce the root password in all cases (even when customers specify `rootpw --lock`)?

We generally want to avoid the situation when the customer forgets to set the root password, since it suddenly became a non requirement, and then needs to either perform a bit complicated root password recovery procedure or re-install from the scratch.

What to do in case the customer has specifically decided to lock the root account, is a tougher question. On one side, if the customer explicitly decides to lock the account, then it's up to him, and I think that we should appreciate his decision. On the other side, the RHV-H instance then becomes non usable in it's main function, as it becomes impossible to add it to the RHVM, so having an option to show the customer a warning in that case would be nice, just to make sure the customer really understands what they're doing, however certainly that is not a requirement.

poncovka:
> We generally want to avoid the situation when the customer forgets to set the root password, since it suddenly became a non requirement, and then needs to either perform a bit complicated root password recovery procedure or re-install from the scratch.

This could be solved by replacing `/usr/share/anaconda/interactive-defaults.ks` with your kickstart file instead of using `inst.ks`. Because of that, I suggest to mark the  `root_password_must_be_set` option as a temporary workaround.

In the future, we plan to replace `interactive-defaults.ks` with configuration files, you might be able to avoid the kickstart file completely.

lveyde:
> > We generally want to avoid the situation when the customer forgets to set the root password, since it suddenly became a non requirement, and then needs to either perform a bit complicated root password recovery procedure or re-install from the scratch.
> 
> This could be solved by replacing `/usr/share/anaconda/interactive-defaults.ks` with your kickstart file instead of using `inst.ks`. Because of that, I suggest to mark the `root_password_must_be_set` option as a temporary workaround.
> 
> In the future, we plan to replace `interactive-defaults.ks` with configuration files, you might be able to avoid the kickstart file completely.

That workaround was already suggested, but it doesn't really works in our case, since the customer may provide their own (partial) kickstart file, and we'll be again in the same situation.

That's why we basically want control over the default behavior, and this solution of configurable option works a lot better for us.

data/product.d/ovirt.conf
M4rtinK: Changed.

pyanaconda/ui/tui/spokes/root_password.py
M4rtinK: That's actually a really good idea, lots can be deduplicated this way, let's do that! :)

BTW, I think we should do this upstream in some form regardless of if the option ends up upstream as well.

M4rtinK:
Updated the PR to show a better status message:
rhv_root_required_v2

Also deduplicated a lot of the code to `ui/lib/user.py`.

pyanaconda/ui/lib/users.py
```diff
@@ -69,3 +73,69 @@ def set_user_list(users_module, user_data_list, remove_unset=False):
...
+    """
+    if users_module.IsRootAccountLocked:
+        # reconfig mode currently allows re-enabling a locked root account if
+        # user sets a new root password
+        if is_reconfiguration_mode():
+            return _("Disabled, set password to enable.")
+        # the root_password_must_be_set config file option overrides even
+        # locked account
+        elif conf.ui.root_password_must_be_set:
+            return _("Root password is not set")
```
poncovka, line 135: This status should say "Root password must be set" and it should be displayed if:
`conf.ui.root_password_must_be_set and not users_module.IsRootPasswordSet`

I don't think that the `IsRootAccountLocked` property matters in this case.

data/anaconda.conf
```diff
@@ -267,6 +267,11 @@ hidden_spokes =
 # Run GUI installer in a decorated window.
 decorated_window = False
 
+# do not have hard requirement on root password being set by default
```
poncovka, line 270: I don't understand the description. Maybe `Require the root password to be set.` might be better?

VladimirSlavik (commented):
I saw this and have no opinion in terms of ack/nack.

I like that this moves the logic away from GUI. At the same time, the assumption that the logic will not be concerned with the state of the GUI is scary. For that reason, I think this isn't really compatible with the redesign, so not upstreamable. Thus, it matters a lot if this [exact form] is temporary or not.

jstodola:
I wonder why is it even necessary to introduce a new configuration option?

On the first screenshot from Martin it is clear that there is no admin account configured, so anaconda should not allow to continue with the installation in such a case, right?

The user spoke is hidden by rhvh.conf:
```
[User Interface]
help_directory = /usr/share/anaconda/help/rhel
hidden_spokes = UserSpoke
```
so the only option to setup an admin user is to have the root account with a password - which should make the root password spoke mandatory.

rvykydal:
Should we close this as being replaced by https://github.com/rhinstaller/anaconda/pull/3701 ?

jstodola:
> The installer thinks that the root configuration was omitted intentionally in the kickstart file and the user doesn't want to change the default settings....

IMO, this is an incorrect assumption.

> Should we close this as being replaced by #3701 ?

+1

jkonecny12:
Closing this PR now.

VladimirSlavik:

PR #3538, opened 2021-07-23
Change the NTP server dialog design

VladimirSlavik:
This removes the "uncheck to remove" antipattern and adds a standard set of buttons.
    
See also: https://github.com/rhinstaller/anaconda/pull/3238#issuecomment-802814455
    
Suggested-by: Mairin Duffy @mairin

TODOs: 
- [x] Try to make it so the newly added server is being edited immediately
- [x] Solve the lots of space between buttons
- [x] Prettify / align  the add, remove, and ok buttons on the left
- [x] Link to https://bugzilla.redhat.com/show_bug.cgi?id=1827683
- [x] Fix handling of added but unchanged entries
- [x] Use non-deprecated icons
- [x] Testing and so on


VladimirSlavik:
So far it looks like this:

Screenshot_vs_main_2021-07-23_20:10:05

https://user-images.githubusercontent.com/15903878/126824070-3b0af159-9eea-44dc-b506-30abc1996878.mp4

VladimirSlavik:
I think I managed to address the most glaring problems and missing bits of functionality, so it's no longer a beta.

@mairin - a few questions if things are right this way?
- After adding a new server, the hostname is immediately being edited. Is that fine?
- After adding a new server, one can just ignore it, click away, and it will stay added but invalid. I would like to make it so that if there is an yet-unedited entry (same string), a new one won't be added, the existing one will be edited instead. Is that a good idea?
- The Edit button is not really necessary, users can click the servers/hostnames to start editing the text. But I think it's better to keep the button anyway?

poncovka:
It looks great! I would like to see a smaller space between the lines for Pool and NTS descriptions, but that's just nitpicking. Also, I was a little confused by the Edit button, because I was somehow expecting a new dialog with advanced options. Maybe one of the notes below could provide a hint about editing the hostname instead of having a special button for that.

jkonecny12:
I also like how it looks now. Great improvement.

pyanaconda/ui/gui/spokes/lib/ntp_dialog.py
```diff
@@ -171,39 +139,48 @@ def _update_rows(self):
...
+        """Handler for Add button"""
         server = TimeSourceData()
-
-        if self._poolCheckButton.get_active():
-            server.type = TIME_SOURCE_POOL
-        else:
-            server.type = TIME_SOURCE_SERVER
-
-        server.hostname = entry.get_text()
+        # make a reasonable guess abut the defaults
```
rvykydal, line 145: a typo "abut"

pyanaconda/ui/gui/spokes/lib/ntp_dialog.py
```diff
@@ -171,39 +139,48 @@ def _update_rows(self):
...
+        num = len(self._serversStore)
+        path = Gtk.TreePath(num - 1)
         itr = self._serversStore.get_iter(path)
-        old_value = self._serversStore[itr][SERVER_USE]
-        self._serversStore.set_value(itr, SERVER_USE, not old_value)
+        selection = self._serversView.get_selection()
+        selection.select_iter(itr)
+        self._serversView.grab_focus()
+
+        # sstart editing the newly added server hostname
```
rvykydal, line 163: a typo "sstart"

rvykydal (approved):
Looks good to me, thank you.

jkonecny12 (approved):
Looks good to me aside the nitpick :). Great improvement indeed.

mairin:
@VladimirSlavik 

`
After adding a new server, the hostname is immediately being edited. Is that fine?
`

Yep that seems as-designed and should be ok. 

`
After adding a new server, one can just ignore it, click away, and it will stay added but invalid. I would like to make it so that if there is an yet-unedited entry (same string), a new one won't be added, the existing one will be edited instead. Is that a good idea?`
`
So if someone clicks "Add" button and clicks away, they'll get an invalid entry that says "<fill in server [... etc...]" as the hostname because they didn't edit it at all? I think in that case, it makes sense to not add that unedited entry and just remove it. 

I don't understand the bit about how the existing one will be edited instead? 

Also, for the case that they click the "Add" button, start typing but do not finish and click away - I would not remove the entry, I would give them a chance to be able to edit it and fix it.

What happens if they fill out the hostname wrong and the "Working" column goes to a red X icon? Will this dialog be dismissable or will it nag them to fix and not let them leave dialog until they remove the invalid entry? Or will it silently remove invalid entries?

`
The Edit button is not really necessary, users can click the servers/hostnames to start editing the text. But I think it's better to keep the button anyway?
`
Yes I would keep it for accessibility. Can you add accessibility hotkeys to each of the 3 buttons as well?

VladimirSlavik:
@mairin - thanks, I think there's a misunderstanding. I've thought about this a bit, let me maybe widen the scope of this discussion to handling of invalid entries. New ones are just a special case of that.

(I mix a bit terms server, entry, hostname. In this discussion they're almost one thing - server is what user wants, it is known by the hostname, and shows as an entry in the table.)

There are several kinds of (in)validity.
- All entries are valid in terms of being able to be stored.
- Only some entries (hostname+pool?+nts?) yield actual working configuration (add a remote time source), which is the green check mark icon. If the configuration does not work, that's the red invalid icon. This can be determined only by trying to contact the server(s), so there can be a considerable time before this is known.
- There is also another kind of (in)validity, when the server/hostname does not conform to rules for hostnames, and that can be determined "offline", as it's being typed.

So, when an user adds a new entry, the defaults are a text about what to do, so by definition invalid hostname, and they must edit it. (Or delete.) The editing that immediately starts is (in my mind) meant to close the gap between "action: add" and "result: got it working".

If the entry is not edited, it remains, knowable as invalid immediately due to "not a hostname". But the defaults are very specific, so it's easy to detected entries as "newly added, yet unedited".

Important caveat: Editing a hostname in the table is checked against being a valid hostname, if it is not, then the edit is just silently dropped and the old value remains. That is how it worked already. Previously, we added new servers via the text field, which did the same validation. End result, having an invalid hostname was not possible. In this case, a byproduct is that editing newly added entries and entering invalid hostname makes the hostname stay the helpful text, so invalid.

With all of that in mind, it's clear that the entries that are `Add`-ed don't fit the situation too well and need some handling. So I propose these behaviors:
- If there is an just-added entry, it is immediately edited, to railroad the user into making it a valid one (already done). Note it's not guaranteed the result will be valid in any sense just by this alone - user can cancel editing or enter invalid hostname.
- If there is an just-added entry and user wants to add a new one, the already present one is reused and edited instead, no new entry is added (the proposal I asked about). (This would limit the amount of just-aded entries to exactly one, and exactly at last position as we add to end and don't reorder.)
- If the user exits the dialog and there is a just-added entry, it is silently removed. (This ensures all entries/servers/hostnames stored permanently, outside the dialog sessions, are valid in terms of hostnames.)
- If there are entries invalid in the sense of not yielding a working configuration, these are kept, same as before.

These workings all aim at eliminating the possibility that users add invalid hostnames. The scheme above could be considerably tightened by requiring that the initial editing either results in a valid hostname, or the new entry immediately disappears again. But I feel that's too hostile and disallows a "slow discovery" that might be typical of "weak" users...

I'm no UX expert, so this might be wrong...

VladimirSlavik:
Update: I have resolved all TODOs except testing. Switching out of draft.

jkonecny12 (approved):
Still looks good to me.

PR #3542, opened 2021-07-27
Mark a flaky pylint warning as a false positive (#infra)

poncovka:
The warning from `NetworkService.get_team_devices` about the `device_type` member of
`AttachedSubscription` class appears randomly, but it is always a false positive.


VladimirSlavik (approved):
Looks good to me. Thank you!

rvykydal (approved):
Looks good to me.

jkonecny12 (approved):
Thanks for this. I encountered this issue again today.

Could you please add TODO to the code to know that we want to solve this issue in the future somehow?

poncovka:
@jkonecny12, I think that it will disappear on its own with a new pylint, so it will be always listed in unused false positives.

VladimirSlavik:
I believe I have seen a false positive disappear and reappear, so a TODO is unneeded, I think.

PR #3545, opened 2021-07-28
Use the similar solution on RHEL-8 branch to gate contributors as on master

jkonecny12:
This will simplify our work with the workflows. It will make this workflow closer to the master version and also we could remove the external contributors specific workflow from master branch.

We don't need to have pr-info anymore because all external contributors are gated by

https://github.blog/changelog/2021-07-01-github-actions-new-settings-for-maintainers/


VladimirSlavik (approved):
Looks good to me. Thank you!

poncovka (approved):

PR #3546, opened 2021-07-28
Rhel 9 improve dracut errors visibility

jkonecny12:
Backport of https://github.com/rhinstaller/anaconda/pull/3533 .

When there is an error printed in Dracut it is scrolled out pretty fast (most of the time). Users could find them in journal but that is not always an easy thing to do and you have to be an advanced user to know where to look.

By this change set, I'm trying to improve the situation. Now you are able to see the errors when Dracut starts to timeout and also after a boot before Anaconda is started.

_Resolves: rhbz#1983098_

### DIFF
--- a/anaconda.py
+++ b/anaconda.py
@@ -245,6 +245,9 @@ if __name__ == "__main__":
         stdout_log.warning("All Anaconda kernel boot arguments are now required to use "
                            "'inst.' prefix!")
 
+    # print errors encountered during boot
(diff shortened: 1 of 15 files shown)


jkonecny12:
Tested and it works correctly.

poncovka (approved):
Thanks!

rvykydal (approved):

VladimirSlavik (approved):

PR #3547, opened 2021-07-28
Master no nosave flags

VladimirSlavik:
This is a draft, the whole thing will need re-creating by careful cherry-picking and then polishing some. Eg. docstrings, comments, switching order of if-else according to variables,...

So, a high-level review is enough. Eg. where to put the function for the task.

Fun fact: Deleting a file means it can't be part of updates image, so for now I have a placeholder ks script.


M4rtinK (commented):
Looks good to me in principle - IIRC back when I implemented this (so that containers don't contain log files with internal URLs & general unnecessary files) one had to use that post script or else the thing would not run late enough. 

It's nice that's no longer the case & we can use proper Python code triggered in normal way, that can now also be unit tested. :)

Also another possible positive side effect of this is with the option now available via the config file products could opt to not save logs and/or kickstarts after installation, which was not previously possible.

Looking forward to the regular version of this, will do a regular review of it then. :)

pyanaconda/core/configuration/anaconda.py
```diff
@@ -131,6 +131,18 @@ def optional_modules(self):
         """
         return self._get_option("optional_modules").split()
 
+    @property
+    def save_input_kickstart(self):
```
poncovka, line 135: Add docstrings, please.

data/anaconda.conf
```diff
@@ -31,6 +31,15 @@ optional_modules =
...
     org.fedoraproject.Anaconda.Addons.*
 
+# Save input kickstart to target system.
+save_input_kickstart = True
+
+# Save output kickstart to target system.
+save_output_kickstart = True
+
+# Save logs to target system.
+save_logs = True
```
poncovka, line 41: I am still not sure if the new options should be part of the `[Anaconda]` section or the `[Installation Target]` section.

data/anaconda.conf
```diff
@@ -31,6 +31,15 @@ optional_modules =
     org.fedoraproject.Anaconda.Modules.Subscription
     org.fedoraproject.Anaconda.Addons.*
 
+# Save input kickstart to target system.
+save_input_kickstart = True
```
poncovka, line 35: Do you think that the current names of these options are ok? Wouldn't be `copy_input_kickstart` better?

pyanaconda/argument_parsing.py
```diff
@@ -544,20 +543,20 @@ def __call__(self, parser, namespace, values, option_string=None):
             if values:
                 options = values.split(",")
             if "all" in options:
-                flags_instance.nosave_input_ks = True
-                flags_instance.nosave_output_ks = True
-                flags_instance.nosave_logs = True
+                setattr(namespace, "save_input_kickstart", False)
```
poncovka, line 546: I think you can do `namespace.save_input_kickstart = False`. Also, I suggest to initialize these values to `None` at the beginning of the method.

pyanaconda/core/configuration/anaconda.py
```diff
@@ -353,6 +365,14 @@ def set_from_opts(self, opts):
         if opts.debug:
             self.anaconda._set_option("debug", True)
 
+        # Set "nosave flags".
+        if hasattr(opts, "save_input_kickstart"):
```
poncovka, line 369: It would be better to initialize these values to `None` and check that.

pyanaconda/installation.py
```diff
@@ -71,6 +71,70 @@ def _writeKS(ksdata):
...
+        util.mkdirChain(conf.target.system_root + "/var/log/anaconda/")
+        # copy log files from the list
+        LOG_FILES_TO_COPY = [
+            "anaconda.log", "syslog", "X.log", "program.log", "packaging.log", "storage.log",
+            "ifcfg.log", "lvm.log", "dnf.librepo.log", "hawkey.log", "dbus.log"
+        ]
+        for logfile in LOG_FILES_TO_COPY:
+            src_path = "/tmp/" + logfile
+            dest_path = conf.target.system_root + "/var/log/anaconda/" + logfile
+            copy_file_if_exists(src_path, dest_path)
```
poncovka, line 99: Ideally, the task could have functions like `_copy_file` and `_copy_tree` and the `run` method would just contain a simple series of these calls. Also, it would be great to log what we are copying in these functions.

data/post-scripts/99-copy-logs.ks
```diff
@@ -1,47 +1 @@
...
-
-%post
-# Relabel the anaconda logs we've just coppied, since they could be incorrectly labeled, like
-# hawkey.log: https://bugzilla.redhat.com/show_bug.cgi?id=1885772.
-# Execution of this %post script will not be logged in the log files on the installed system.
-
-restorecon -ir /var/log/anaconda/
-
-%end
+# empty file
```
poncovka, line 1: This file needs to be removed. The `makeupdates` script should be able to handle removed files, but it might need some new tweaks for the post-scripts directory.

poncovka (commented):
I like it. The next step could be to replace `90-copy-screenshots.ks` and move it to the same task. We might also get rid of `80-setfilecons.ks` and replace post-scripts with something more general.  It would be great to know if someone adds custom scripts to this directory. However, that is out of scope of this pull request.

jkonecny12 (commented):
I second to @poncovka 's notes other than that this is really great improvement! Thanks!

data/post-scripts/99-copy-logs.ks
VladimirSlavik: Unfortunately that cannot be solved: An updates image can contain only new file data, which will be written. But there is no instruction in the archive format to delete files instead of writing. So no. But let's not dwell on that, this is just a minor detail of the WIP.

pyanaconda/core/configuration/anaconda.py
VladimirSlavik: Oh. My brain is still strongly typed, it seems.

VladimirSlavik:
After coming back to this and reviewing the commits, I see three somewhat independent sub-tasks in this, and will open new PRs accordingly. This will remain here until all three are done, then I'll just close this.

VladimirSlavik:
Closing in favor of #3571 and #3584.

PR #3552, opened 2021-07-31
Resolve filename into associated loop device

marmarek:
This fixes checking if given (disk image) file is mounted somewhere,
especially as /run/initramfs/repo.
PackagePayload._device_is_mounted_as_source() depends on it.

Some more details at https://github.com/QubesOS/qubes-issues/issues/6792

This is related to https://bugzilla.redhat.com/show_bug.cgi?id=1830553, although is not enough to fix it completely on Fedora 34 Server install.

cc @jkonecny12

### DIFF
--- a/pyanaconda/payload/utils.py
+++ b/pyanaconda/payload/utils.py
@@ -49,6 +49,9 @@ def get_device_path(device_name):
 
     device_tree = STORAGE.get_proxy(DEVICE_TREE)
     device_data = DeviceData.from_structure(device_tree.GetDeviceData(device_name))
+    if device_data.type == 'file' and device_data.children:
+        # resolve file path to associated loop device, if any
+        device_data = DeviceData.from_structure(device_tree.GetDeviceData(device_data.children[0]))
     return device_data.path
 
 


marmarek:
Unit test failed because `pytest` is missing. Should I rebase over https://github.com/rhinstaller/anaconda/pull/3553 ?

jkonecny12:
Yes please. Sorry for the disturbance.

jkonecny12:
There was old container used for the test run. Hopefully, it will be fine now.

jkonecny12:
I have to reproduce this and play with it a bit but I have a feeling that you are solving issue which behave differently on current Rawhide. I will reproduce this issue and find out more about it, but right now I have to take care of something else. I'll try to get to this ASAP.

marmarek:
It's hard to hit this issue on Rawhide, because another issue prevents going this far: https://bugzilla.redhat.com/show_bug.cgi?id=1830553 - specifically, a local ISO file is not tried at all and the network mirror is used (instead of trying the local ISO and failing). I haven't debugged the latter, but it sounds like you have an idea why it happens.

jkonecny12:
Hi, so I tested the issue completely and I'm convinced that your PR will not solve the issue. In short there are two variants we have now:
- booting with `inst.stage2=` and `iso-scan/filename` (your use-case)
In this case Anaconda will not even try to find the repository because we are ignoring loop devices. That is because we are using blivet (we shouldn't do that probably) to simplify the task and blivet has on purpose set the loop devices as ignored source.
For these reasons the output of Anaconda is nothing found -> `Closest Mirror` fallback.

- booting with `inst.repo=` and `iso-scan/filename`
In this case Anaconda knows that it should look for a device with LABEL set by the `inst.repo=` and won't use blivet search solution at all. This solution will work as expected and set the ISO as source of the installation.

To fix the first issue we should probably implement idea we have in our minds some time already. That is to create a source which would just re-use mount point from the Dracut and use that as the installation source. Unfortunately, we were not able to find the time to do that yet.

For the above reasons, I'm closing this PR and we will rather target the described solution above. If you are able, then you should change the `inst.stage2` in your solution to `inst.repo` and that way solve your issue.

Thanks a lot for pointing this issue out and providing all the information.

marmarek:
> change the `inst.stage2` in your solution to `inst.repo` and that way solve your issue.

Should it really be _instead of_, not _in addition to_? How would it find the installation image otherwise? Does the dracut module look for `inst.repo` too?

jkonecny12:
Hi, no you can use `inst.repo` instead of `inst.stage2`. We will automatically use `inst.repo` as source for stage2 image in case `inst.stage2` is not specified. This is work done by the Anaconda scripts in the initrd https://github.com/rhinstaller/anaconda/blob/e7a6035482d40afbf8997221dbd940de4ca09afa/dracut/parse-anaconda-repo.sh#L14 .

PR #3555, opened 2021-08-03
Add release notes for visible warnings from initrd

jkonecny12:


poncovka (approved):
Thanks!

docs/release-notes/dracut_visible_warnings.rst
```diff
@@ -0,0 +1,11 @@
+:Type: Initial RAM disk (initrd)
+:Summary: Make installer critical warnings in Dracut/initrd more visible (#1983098)
```
VladimirSlavik, line 2: I think swapping the adjectives might work better: `...critical installer...` However the adjective ordering rules don't fit too well here, so feel free to disagree.

docs/release-notes/dracut_visible_warnings.rst
```diff
@@ -0,0 +1,11 @@
...
+
+:Description:
+    Installation shows critical warnings raised in Dracut/initrd again when Anaconda is
+    starting or when Dracut starts to timeout. This should help users to resolve installation
```
VladimirSlavik, line 6: "starts to timeout" - maybe "times out" ?

jkonecny12: It's a bit tricky. The visible warnings will show up to user when retry attempts are half of the limit. So it still may be able to boot.

docs/release-notes/dracut_visible_warnings.rst
jkonecny12: Fixed. Thanks for suggestion.

jkonecny12:
**UPDATED:**
* Change Summary. Thanks for suggestion @VladimirSlavik !

docs/release-notes/dracut_visible_warnings.rst
VladimirSlavik: Ah, then it's good as is, I think. It's hard to understand, but you don't want to write a novel to explain it...

VladimirSlavik (approved):

PR #3556, opened 2021-08-03
Move post installation scripts to the end of queue

poncovka:
(cherry-picked from a commit 20001b6)

Resolves: rhbz#1983922

### DIFF
--- a/pyanaconda/installation.py
+++ b/pyanaconda/installation.py
@@ -176,10 +176,6 @@ def _prepare_configuration(payload, ksdata):
         # Join a realm. This can run only after network is configured in the target system chroot.
         configuration_queue.append_dbus_tasks(SECURITY, [security_proxy.JoinRealmWithTask()])
 
-    post_scripts = TaskQueue("Post installation scripts", N_("Running post-installation scripts"))
-    post_scripts.append(Task("Run post installation scripts", runPostScripts, (ksdata.scripts,)))
-    configuration_queue.append(post_scripts)
-
     # setup kexec reboot if requested
     if flags.flags.kexec:
         kexec_setup = TaskQueue("Kexec setup", N_("Setting up kexec"))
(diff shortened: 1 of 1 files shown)


jkonecny12 (approved):
LGTM!

jkonecny12:
Probably should be asked on the original PR, but shouldn't we have tests for this code?

poncovka:
> Probably should be asked on the original PR, but shouldn't we have tests for this code?

I don't think it is worth it at this point. These tasks should be scheduled dynamically in the future, so this implementation will disappear.

rvykydal (approved):
Looks good to me.

jstodola:
> > Probably should be asked on the original PR, but shouldn't we have tests for this code?
> 
> I don't think it is worth it at this point. These tasks should be scheduled dynamically in the future, so this implementation will disappear.

The last sentence actually sounds like 2 reasons for having this covered by a test. Even if this implementation will disappear, the functionality should be the same from the customer point of view. "Scheduled dynamically" sounds like a source of unpredictable problems for a QE person...

VladimirSlavik:
+1 on tests not worth it. What is there to test? Order of tasks in the queue? Their existence in it?

jstodola:
Based on bug 1983922, a kickstart-tests test could test that /root/anaconda-ks.cfg exists when kickstart %post scripts are executed and the file is not empty.

jkonecny12:
In that case maybe we want to cover this by KS tests not by unit tests. Unit tests are more focused on implementation.

VladimirSlavik:
Unit tests could test check the task is in the queue. Kickstart tests could check that `anaconda-ks.cfg` exists, true, but it's such a generic thing that maybe all of them should. Or there could be some special test to check the common things, or perhaps the coverage ones...

PR #3557, opened 2021-08-04
Don't run shell on every found virtualization console

poncovka:
Fix a bug that was introduced in the commit 3427a30. The fix should resolve the
issue with `anaconda-shell@.service` that is repeatedly failing with a message:
`/dev/hvc1: cannot open as standard input: No such device`

Fix the `makeupdates` script to include anaconda-generator again.


jkonecny12 (approved):
Looks great to me. Thanks for fixing so old issue in the code!

VladimirSlavik (approved):
Thank you!

rvykydal (approved):
Looks good to me, thank you.

PR #3558, opened 2021-08-05
Fix dependency on zram-generator in RHEL

sgallagher:

### DIFF
--- a/anaconda.spec.in
+++ b/anaconda.spec.in
@@ -212,7 +212,7 @@ Requires: device-mapper-multipath
 %if ! 0%{?rhel}
 Requires: zram-generator-defaults
 %else
-Requires: rust-zram-generator
+Requires: zram-generator
 %endif
 # Display stuff moved from lorax templates
 Requires: xorg-x11-drivers


VladimirSlavik:
Thank you! Does this need porting somewhere too?

sgallagher:
We will need a build with this change in Fedora and CentOS Stream 9 once https://kojihub.stream.rdu2.redhat.com/koji/buildinfo?buildID=10659 lands in the compose.

jstodola:
@sgallagher thanks for this change. Can you also create a RHEL-9 bug against anaconda?

msekletar:
I think you guys need to go ahead and build the Anaconda with this change. Otherwise we won't have any compose, as it will fail because Anaconda is still referencing the package by its old name.

poncovka (approved):
Looks good to me. Thanks!

sgallagher:
I created https://bugzilla.redhat.com/show_bug.cgi?id=1991617 for RHEL 9

jkonecny12:
Thanks a lot @sgallagher and @msekletar for solving this issue so fast!

PR #3560, opened 2021-08-10
Fix creation of user account with empty password

M4rtinK:
Make it possible to always create a user account without a password
after the "Require a password to use this account" checkbox
in the Anaconda GUI is unchecked.

Previously it was possible to get into a situation when password
validation checks would be active even if the checkbox was unchecked,
preventing user account with empty password from being created in the
Anaconda GUI.

The activation conditions of these checks has been fixed, so it should now
be possible to create a user account with empty password at all times.

Resolves: rhbz#1687315


poncovka (approved):

VladimirSlavik (approved):

jkonecny12 (approved):
Looks good to me. However, you can't merge it now. The bug don't have all the ACKs and was moved to next minor.

jkonecny12:
Interesting point. However, not sure if we need label. It's not happening often.

PR #3561, opened 2021-08-11
Don't set default of gpt option at cmdline parsing

cheese:
Value of config option is always overridden by cmdline parsing result. A
cmdling parsing option with a default value will implicitly override the
corresponding value in a config file.

This patch set set default value of gpt option at cmdline parsing to
'argparse.SUPPRESS', so it is possible to distinguish the case when gpt option
is not passed in cmdline.

Without this fix, it is not possible to override storage.gpt option with a
config file, for example, '/etc/anaconda/profile.d/some.conf', even when gpt
option is not passed at the cmdline.

It is safe to remove default of gpt option at cmdline parsing, since storage.gpt
option defaults to False in the default config.

### DIFF
--- a/pyanaconda/core/configuration/anaconda.py
+++ b/pyanaconda/core/configuration/anaconda.py
@@ -363,7 +363,8 @@ class AnacondaConfiguration(Configuration):
         # Set the storage flags.
         self.storage._set_option("dmraid", opts.dmraid)
         self.storage._set_option("ibft", opts.ibft)
-        self.storage._set_option("gpt", opts.gpt)
+        if hasattr(opts, "gpt"):
+            self.storage._set_option("gpt", opts.gpt)
         self.storage._set_option("multipath_friendly_names", opts.multipath_friendly_names)
 
         # Set up the rescue mode.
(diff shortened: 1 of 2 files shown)


pyanaconda/core/configuration/anaconda.py
```diff
@@ -363,7 +363,8 @@ def set_from_opts(self, opts):
         # Set the storage flags.
         self.storage._set_option("dmraid", opts.dmraid)
         self.storage._set_option("ibft", opts.ibft)
-        self.storage._set_option("gpt", opts.gpt)
+        if hasattr(opts, "gpt"):
```
poncovka, line 366: I did some testing and it looks like you could use `if "gpt" in opts`. It seems easier to understand.

poncovka:
Thanks for this fix! We will have to eventually do that for all our options that we use to set up the configuration, but this is great start.

jkonecny12 (approved):
Great fix. Thanks a lot for providing it!

M4rtinK (approved):
Looks good to me, thanks! :)

poncovka (approved):
Tested.

PR #3562, opened 2021-08-11
Add product config for AlmaLinux

andrewlukoshko:
AlmaLinux is Open Source, community-governed and forever-free enterprise Linux distribution, focused on long-term stability, providing a robust production-grade platform. AlmaLinux OS is 1:1 binary compatible with RHEL® and pre-Stream CentOS.
https://almalinux.org


VladimirSlavik (commented):
Andrew, thank you!

I think there is no problem adding this, but the decision ultimately depends on more people than me.

----

To modify code on RHEL, you need a bug and link it from the commit:
```
Add product config for AlmaLinux (#NNNN)

Resolves: rhbz#NNNN
```
See https://github.com/rhinstaller/anaconda/blob/rhel-8/docs/commit-log.rst for more info

----

You should add a test as well. The place to do so would be somewhere under here https://github.com/rhinstaller/anaconda/blob/rhel-8/tests/nosetests/pyanaconda_tests/product_test.py#L236 and I guess it would look mostly like this:
```
self._check_default_profile(
    "AlmaLinux", "",
    ["rhel.conf", "almalinux.conf"],
    ENTERPRISE_PARTITIONING
)
```

See https://github.com/rhinstaller/anaconda/blob/rhel-8/tests/README.rst for how to run the tests. I have not ran them manually on the rhel 8 branch for a long time, but assuming it all works, the tl;dr should be `make anaconda-ci-build && make container-ci` and go make coffee or eat your lunch, it takes 8-20 minutes. Alternatively, push here and tests will run, but somebody will have to approve them each time, so it might take far longer.

We really want tests, so as a last resort, somebody can write the tests for you.

----

Finally, is this something you'd like to see in RHEL 9 too? If so, the process is the same, except we modified things a bit so the fields would be slightly different.

jkonecny12:
> See https://github.com/rhinstaller/anaconda/blob/rhel-8/tests/README.rst for how to run the tests. I have not ran them manually on the rhel 8 branch for a long time, but assuming it all works, the tl;dr should be `make anaconda-ci-build && make container-ci` and go make coffee or eat your lunch, it takes 8-20 minutes. Alternatively, push here and tests will run, but somebody will have to approve them each time, so it might take far longer.

This will not work. You have to have access to RHEL repositories which are not released so outside collaborators unfortunately can't run RHEL-8 tests. I would like to move this / support also CentOS Stream which would work but I don't have a capacity to do that now and also waiting until CentOS Stream will stabilize a bit.

jkonecny12:
In general, thank you for the configuration file. Would be great to have it here because we can then update it to avoid breakage of your configuration file and you will just create PR on us in case you need to change something specific to your configuration.

This PR should definitely go to `master` branch too. The `master` branch should be included to have your configuration file as part of next RHELs. Where you can easily also develop the tests mentioned. 

Also as @VladimirSlavik already mentioned, you should also backport this to RHEL-9 if you want to have it there.

andrewlukoshko:
Test for AlmaLinux product is added to PR.
@jkonecny12 Could you please re-run check?

andrewlukoshko:
@VladimirSlavik sorry I forgot to mention bug https://bugzilla.redhat.com/show_bug.cgi?id=2003692 in commit message.
Updated.
I've also added PRs for rhel-9 and master branches.

VladimirSlavik:
Thank you, still a few steps though:

> To modify code on RHEL, you need a bug and link it from the commit:
> 
> ```
> Add product config for AlmaLinux (#NNNN)
> 
> Resolves: rhbz#NNNN
> ```
> 
> See https://github.com/rhinstaller/anaconda/blob/rhel-8/docs/commit-log.rst for more info

The line with `Resolves:` is actually what tools check as of late, so missing it is not optional. (Same goes for the rhel-9 PR.)

Second, this needs to be on *all* commits. I would actually advise squashing the two.

If you want, I can fix this for you.

Once that is done, the PR is in order, and the bug needs to be accepted. (Again same for rhel-9).

andrewlukoshko:
@VladimirSlavik I've squashed commits and added Resolves: line

VladimirSlavik (approved):
Thanks!

It seems this should be able to eventually get into 8.6.

PR #3563, opened 2021-08-12
Remove blivet daily builds from Anaconda dailies (#infra)

jkonecny12:
We need to remove it here because ELN is not build by blivet now. So to avoid this problem, I'm removing the COPR repository from global settings and set it only to `fedora-rawhide-x86_64` chroot only on COPR manually.

Am I able to convinced you @vojtechtrefny to enable ELN back for daily builds? :)

### DIFF
--- a/.packit.yml
+++ b/.packit.yml
@@ -32,8 +32,6 @@ jobs:
       owner: "@rhinstaller"
       project: Anaconda
       preserve_project: True
-      additional_repos:
-        - "copr://@storage/blivet-daily"
 
 #  - job: copr_build
 #    trigger: commit


VladimirSlavik (approved):
Looks good to me. Thank you!

vojtechtrefny:
> Am I able to convinced you @vojtechtrefny to enable ELN back for daily builds? :)

No. Our Copr ELN builds are failing for two months now, both daily builds and packit (btw. [yours](https://copr.fedorainfracloud.org/coprs/g/rhinstaller/Anaconda/builds/) are failing too) so I don't really see a reason to enable them now.

poncovka:
Maybe it is a sign that we should disable our ELN builds as well for a while ;-)

jkonecny12:
Tested locally and we have a successful daily build of Anaconda with current ELN.
https://copr.fedorainfracloud.org/coprs/g/rhinstaller/Anaconda/build/2525639/

PR #3564, opened 2021-08-17
Remove blivet COPR repository from ELN containers (#infra)

jkonecny12:
Blivet doesn't have ELN builds in their COPR repository now. So using this repository will fail container build because the chroot is not available.

This should solve our current `Refresh anaconda containers` workflow.


poncovka (approved):

VladimirSlavik (approved):
Looks good to me. Thank you!

PR #3565, opened 2021-08-17
Check source type directly, not via DBus

M4rtinK:
Not all payloads are available via DBus in RHEL 9 Anaconda,
so use the payload instance directly when checking payload type
in the Connect to Red Hat spokes. Also in some cases the payload
type could be None instead of a string, so handle that as well.

This should fix the issue with RHV, which uses the liveimg payload,
which is not available via DBus on RHEL 9, leading to a crash.

(cherry picked from commit d6ea3b39b839e4822ea7d1e9afea5fb81352b732)

Related: rhbz#1941578
Resolves: rhbz#1955311

### DIFF
--- a/pyanaconda/ui/gui/spokes/subscription.py
+++ b/pyanaconda/ui/gui/spokes/subscription.py
@@ -858,9 +858,8 @@ class SubscriptionSpoke(NormalSpoke):
 
         # check if the current installation source will be overriden
         # and remember it if it is the case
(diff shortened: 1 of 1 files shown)


VladimirSlavik (approved):
Looks good to me. Thank you!

poncovka (approved):

jkonecny12 (approved):
Looks good to me.

PR #3570, opened 2021-08-23
Fix issues from new pylint 2.10.2

jkonecny12:
This PR solve three pylint issues which breaking our container refresh now:
- Ignore new pylint warning which is not important for Anaconda project.
- Remove not needed 'u' prefix for strings (spoilert alert -- all the strings in Python3 are unicode :hand_over_mouth: )
- Revert our disable for a flake we had because of NetworkManager. I was looking for last 5 builds and it was always marked as false positive so let's try to get it back again and see.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -48,8 +48,8 @@ source_suffix = '.rst'
 master_doc = 'index'
 
 # General information about the project.
-project = u'Anaconda'
-copyright = u'2015, Red Hat, Inc.'      # pylint: disable=redefined-builtin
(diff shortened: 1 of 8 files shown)


poncovka (approved):
Thanks!

VladimirSlavik (approved):
Mostly LGTM, thank you!

I'm not entirely convinced about the last commit, but I might be just too paranoid. Sometimes the errors come back with new astroid that fixes a bug with a bugfix.

M4rtinK (approved):

PR #3576, opened 2021-08-25
Define a unique id in the main spokes and hubs

poncovka:
Implement the `get_screen_id` method in spokes and hubs that provide
built-in help. Eventually, all spokes and hubs should have a unique
screen id.


poncovka:
The pull request was updated. I have removed `hub` and `spoke` from the ids.

jkonecny12 (approved):
Looks even better to me now.

PR #3577, opened 2021-08-25
data/profile.d: Add profiles for KDE & Kinoite Spins

travier:
Both profiles are based on Workstation but use the generic Fedora Linux
branding until we make specific ones for those two spins.

Fedora Kinoite is a new Fedora variant similar to Fedora Silverblue,
introduced in Fedora 35.

See https://fedoraproject.org/wiki/Changes/Fedora_Kinoite

### DIFF
--- a/data/profile.d/fedora-kinoite.conf
+++ b/data/profile.d/fedora-kinoite.conf
@@ -1,0 +1,11 @@
+# Anaconda configuration file for Fedora Kinoite.
+
+[Profile]
+# Define the profile.
+profile_id = fedora-kinoite
+base_profile = fedora-kde
+
+[Profile Detection]
+# Match os-release values.
+os_id = fedora
+variant_id = kinoite
--- a/data/profile.d/fedora-kde.conf
+++ b/data/profile.d/fedora-kde.conf
@@ -1,0 +1,17 @@
+# Anaconda configuration file for Fedora KDE Spin.
+
+[Profile]
+# Define the profile.
+profile_id = fedora-kde
+base_profile = fedora
+
+[Profile Detection]
+# Match os-release values.
+os_id = fedora
+variant_id = kde
+
+[Payload]
+default_environment = kde-desktop-environment
+
+[Bootloader]
+menu_auto_hide = True
--- a/tests/unit_tests/pyanaconda_tests/core/test_profile.py
+++ b/tests/unit_tests/pyanaconda_tests/core/test_profile.py
@@ -251,6 +251,18 @@ class ProfileConfigurationTestCase(unittest.TestCase):
             WORKSTATION_PARTITIONING
         )
         self._check_default_profile(
+            "fedora-kde",
+            ("fedora", "kde"),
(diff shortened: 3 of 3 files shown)


poncovka:
Actually, the initial setup doesn't use `VARIANT_ID` to load profile configuration files. It uses only `ID`, so I don't understand how it is broken. It should use the `fedora` configuration profile for all variants of Fedora. See: https://github.com/rhinstaller/initial-setup/blob/1ba27b4d9db7de2a46e5f208cdab65efb07c927a/initial_setup/__init__.py#L110-L112

data/profile.d/fedora-kinoite.conf
```diff
@@ -0,0 +1,15 @@
+# Anaconda configuration file for Fedora Kinoite.
+
+[Profile]
+# Define the profile.
+profile_id = fedora-kinoite
+base_profile = fedora-workstation
```
Conan-Kudo, line 6: We should be based on Fedora KDE, not Fedora Workstation.

Conan-Kudo (changes requested):

M4rtinK:
> There was no use case for that so far, we need to figure out how to approach the issue. As a temporary workaround, we could reset the `hidden_spokes` option in the initial setup configuration file. See: https://github.com/rhinstaller/initial-setup/blob/master/data/10-initial-setup.conf

Agreed - maybe have initial setup specific variant of `hidden_spokes` so that both can be set separately?

And in general, I agree this is something Initial Setup should handle & I'm ready to help as Initial Setup maintainer.

travier:
OK, I've updated this one to mirror the workstation settings while keeping things independent until we figure out where we can share more.
As I don't think this will change much for F35, let's keep it for master only and I'll try to work on fixing the initial setup issue. Tracking issue: https://github.com/rhinstaller/initial-setup/issues/136

poncovka:
> Agreed - maybe have initial setup specific variant of `hidden_spokes` so that both can be set separately?

That makes zero sense. I am not going to support two versions of configuration options. If we are not able to use Anaconda's profiles in Initial Setup, then Initial Setup should have its own profiles.

poncovka:
> Actually, the initial setup doesn't use `VARIANT_ID` to load profile configuration files. It uses only `ID`, so I don't understand how it is broken. It should use the `fedora` configuration profile for all variants of Fedora. See: https://github.com/rhinstaller/initial-setup/blob/1ba27b4d9db7de2a46e5f208cdab65efb07c927a/initial_setup/__init__.py#L110-L112

It means that https://github.com/rhinstaller/anaconda/pull/3577#issuecomment-907229246 is not true and you can set `hidden_spokes` in your profiles, because Initial Setup won't use them. The current support should be sufficient for your use case.

travier:
I think this one should be ready. I will try to work on the suggestion from https://github.com/rhinstaller/initial-setup/issues/136#issuecomment-920740804 in a different PR.

poncovka (approved):
Thanks!

PR #3571, opened 2021-08-23
Copy logs in a task

VladimirSlavik:
Copy logs in a task. Split from #3547.

~~Still fighting with tests.~~

Depends on #3584

PR #3584, opened 2021-08-30
Replace nosave flags with conf

VladimirSlavik:
This implements "the other half" split off #3547, which must go before #3571.

### DIFF
--- a/data/anaconda.conf
+++ b/data/anaconda.conf
@@ -60,6 +60,15 @@ system_root = /mnt/sysroot
 # Should we install the network configuration?
 can_configure_network = True
 
+# Should we copy input kickstart to target system?
+can_copy_input_kickstart = True
+
+# Should we save kickstart equivalent to installation settings to the new system?
+can_save_output_kickstart = True
+
+# Should we save logs from the installation to the new system?
+can_save_installation_logs = True
+
 
 [Network]
 # Network device to be activated on boot if none was configured so.
--- a/pyanaconda/argument_parsing.py
+++ b/pyanaconda/argument_parsing.py
@@ -544,19 +544,27 @@ def getArgumentParser(version_string, boot_cmdline=None):
             if values:
                 options = values.split(",")
             if "all" in options:
+                namespace.can_copy_input_kickstart = False
+                namespace.can_save_output_kickstart = False
+                namespace.can_save_installation_logs = False
                 flags_instance.nosave_input_ks = True
                 flags_instance.nosave_output_ks = True
                 flags_instance.nosave_logs = True
             else:
                 if "all_ks" in options:
+                    namespace.can_copy_input_kickstart = False
+                    namespace.can_save_output_kickstart = False
                     flags_instance.nosave_input_ks = True
                     flags_instance.nosave_output_ks = True
                 else:
                     if "input_ks" in options:
+                        namespace.can_copy_input_kickstart = False
                         flags_instance.nosave_input_ks = True
                     if "output_ks" in options:
+                        namespace.can_save_output_kickstart = False
                         flags_instance.nosave_output_ks = True
                 if "logs" in options:
+                    namespace.can_save_installation_logs = False
                     flags_instance.nosave_logs = True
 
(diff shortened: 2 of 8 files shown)


VladimirSlavik:
I think something about the default values of the fields is wrong. Too late now, will fix tomorrow.

pyanaconda/core/configuration/target.py
```diff
@@ -69,3 +69,18 @@ def can_configure_network(self):
         :return: True or False
         """
         return self._get_option("can_configure_network", bool)
+
+    @property
+    def copy_input_kickstart(self):
+        """Should we copy input kickstart to the new system?"""
+        return self._get_option("copy_input_kickstart")
```
poncovka, line 76: You need to specify the `bool` type. Otherwise, the property returns a string.

tests/unit_tests/pyanaconda_tests/test_argparse.py
```diff
@@ -165,6 +165,51 @@ def test_target(self):
         assert conf.target.is_directory == True
         assert conf.target.physical_root == "/what/ever"
 
+    def test_target_nosave(self):
+        conf = AnacondaConfiguration.from_defaults()
+
+        opts, _removed = self._parseCmdline([])
+        conf.set_from_opts(opts)
+
+        assert conf.target.copy_input_kickstart == True
```
poncovka, line 174: I think you need to use `is` instead of `==`.

data/anaconda.conf
```diff
@@ -60,6 +60,15 @@ system_root = /mnt/sysroot
 # Should we install the network configuration?
 can_configure_network = True
 
+# Should we copy input kickstart to target system?
+copy_input_kickstart = True
```
poncovka, line 64: I would follow the name pattern for system and target properties:

```
can_copy_input_kickstart
can_save_output_kickstart
```

There might be more logs to copy in the future, so I would change the last one into:
```
can_copy_installation_logs
```

pyanaconda/argument_parsing.py
```diff
@@ -541,25 +540,29 @@ def __call__(self, parser, namespace, values, option_string=None):
...
+                        namespace.copy_input_kickstart = False
                     if "output_ks" in options:
-                        flags_instance.nosave_output_ks = True
+                        namespace.save_output_kickstart = False
                 if "logs" in options:
-                    flags_instance.nosave_logs = True
+                    namespace.save_logs = False
 
-    ap.add_argument("--nosave", action=ParseNosave, nargs="?", help=help_parser.help_text("nosave"))
+    ap.add_argument("--nosave", action=ParseNosave, nargs="?", default="",
```
poncovka, line 564: Could you try to use `SUPPRESS` and `if name in opts`? Then you don't have to set the values to `None`. I guess we should start to go in that direction. See: https://github.com/rhinstaller/anaconda/pull/3561/files

poncovka (commented):

tests/unit_tests/pyanaconda_tests/test_argparse.py
VladimirSlavik: That's what the conversion script gave us, and what is in the rest of the tests. I think it would use `is` if it was a better choice?

The `assert` statement takes a `bool`, therefore the comparisons are redundant - for `True` I'd just leave the value, for `False` add a `not`. That's how one should assert, ideally. Is that acceptable?

data/anaconda.conf
VladimirSlavik: Sounds good to me, will do.

tests/unit_tests/pyanaconda_tests/test_argparse.py
VladimirSlavik: Ah, yes, `is` seems to be correct.

pyanaconda/argument_parsing.py
VladimirSlavik: `SUPPRESS` will not do anything because we add these values in a custom action - if it isn't invoked, the args are not there anyway. The rest is correct though, let's try this.

poncovka (approved):
It looks great! Thanks!

PR #3585, opened 2021-09-01
Prepare for removal of product vars from constants

VladimirSlavik:
Some steps towards removing product-related variables from `pyanaconda.core.constants`. These are mostly re-imported from `pyanaconda.product` anyway.

It is not possible to fully remove them yet, because initial setup and addons depend on these too. So these are only changes inside anaconda codebase, actual removal must come later. After this is merged & released, https://github.com/OpenSCAP/oscap-anaconda-addon/pull/171 can go, and after that is handled, we can delete the product stuff in `constants` with another PR.

### DIFF
--- a/pyanaconda/core/constants.py
+++ b/pyanaconda/core/constants.py
@@ -34,11 +34,7 @@ from pyanaconda import product
 productName = product.productName
 productVersion = product.productVersion
 isFinal = product.isFinal
-
-# for use in device names, eg: "fedora", "rhel"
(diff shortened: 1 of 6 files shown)


poncovka (commented):
I think we should get rid of the `shortProductName` variable. We could use `get_os_release_value("ID")` or add a new configuration option. It looks like we use the variable only in Blivet as a device prefix, so the new option could be `device_prefix` in the `[Storage]` section. That's probably better, because it can be easily changed.

PR #3586, opened 2021-09-01
Replace the mpath flag with a configuration option

VladimirSlavik:
SSIA... another flag goes, hopefully with less fuss.

### DIFF
--- a/anaconda.py
+++ b/anaconda.py
@@ -338,7 +338,6 @@ if __name__ == "__main__":
 
     # set flags
     flags.rescue_mode = opts.rescue
-    flags.mpath = opts.mpath
     flags.eject = opts.eject
     flags.kexec = opts.kexec
 
--- a/data/anaconda.conf
+++ b/data/anaconda.conf
@@ -161,6 +161,10 @@ ibft = True
 # Do you prefer creation of GPT disk labels?
 gpt = False
 
+# Are multipath devices visible in GUI?
+# Can be overridden with the 'inst.nompath' boot option.
+allow_multipath_devices = True
+
 # Tell multipathd to use user friendly names when naming devices during the installation.
 multipath_friendly_names = True
 
--- a/pyanaconda/core/configuration/anaconda.py
+++ b/pyanaconda/core/configuration/anaconda.py
@@ -365,6 +365,7 @@ class AnacondaConfiguration(Configuration):
         self.storage._set_option("ibft", opts.ibft)
         if hasattr(opts, "gpt"):
             self.storage._set_option("gpt", opts.gpt)
+        self.storage._set_option("allow_multipath_devices", opts.mpath)
         self.storage._set_option("multipath_friendly_names", opts.multipath_friendly_names)
 
         # Set up the rescue mode.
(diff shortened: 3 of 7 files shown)


VladimirSlavik:
I hope to fix all the test comparisons in a subsequent PR just for that. I am changing too much of that file right now.

poncovka:
I did some digging in the history and I think that `inst.nompath` should work the same way as `inst.nodmraid`. Blivet actually has a flag called `multipath`, that we happily ignore and Blivet doesn't use at all. That's suspicious. It would be good to figure it out, because the option might end up in `User Interface` instead of the `Storage` section based on the current usage.

@vojtechtrefny, what do you think?

vojtechtrefny:
> I did some digging in the history and I think that `inst.nompath` should work the same way as `inst.nodmraid`. Blivet actually has a flag called `multipath`, that we happily ignore and Blivet doesn't use at all. That's suspicious.

AFAICT we don't use the `multipath` flag at all and haven't been using it for years. Looks like we stopped using it when we stopped writing our own `multipath.conf` in this [commit](https://github.com/storaged-project/blivet/commit/e35af4fc7a75764b9c7a8aa082dc0e3a3fa791d8). If `nompath` should work like `nodmraid` it most likely didn't work like that for 8+ years.

jkonecny12:
If it is not used for such long than I would suggest to remove it from Anaconda. Ideally to remove logic and leave just a boot option with deprecation notice and remove even the boot option on next-next Fedora?

VladimirSlavik:
So there are really two things to be seen here:
- There is no actual storage-related logic (de)activated by the flag. Blivet is removing the last traces of this.
- The GUI is still changed by this option.

So there are really two ways forward:
- Take this PR and amend it to use the right conf section.
- Deprecate the option so that it can be removed later, which still removes the flag which is the actual goal here.

jkonecny12:
Good point that it's changing GUI.

VladimirSlavik:
We might instead try something like #3597...

poncovka:
I would like to point out that `nompath` seems to disable the `multipathd` service. Is it possible that `nompath` without the `inst.` prefix somehow worked before we started to require the prefix? Maybe we should just support `nompath` instead of `inst.nompath`.

VladimirSlavik:
Oh, that sounds pretty probable, actually. Just wonderful. In that case I think we should instead check if the multipath thingy is turned off, it knows better than we do when that is the case.

jkonecny12:
> I would like to point out that `nompath` seems to disable the `multipathd` service. Is it possible that `nompath` without the `inst.` prefix somehow worked before we started to require the prefix? Maybe we should just support `nompath` instead of `inst.nompath`.

If this is true then we should probably also backport this solution to RHEL-9 because the issue will be even there.

poncovka:
@jstodola, do you have any insights on this topic?

VladimirSlavik:
I discussed this with Jan in person, and he thought removing the remains of mpath stuff would be best.

jstodola:
Yes, I don't believe we need to have a mechanism to hide the Multipath tab (we do not have boot arguments to hide the SAN or NVDIMM tabs either). I would prefer to remove the hiding code, show the Multipath tab always and stop supporting the `inst.nosmt` boot argument (supposing it only affect the GUI). #3597 looks like a good direction to me.

PR #3587, opened 2021-09-01
Replace ksdevice with inst ks device

rvykydal:


VladimirSlavik (approved):
Looks good to me. Thank you!

poncovka (approved):

dracut/parse-anaconda-net.sh
```diff
@@ -16,30 +16,27 @@ mac_to_bootif() {
 
 # handle ksdevice (tell us which device to use for ip= stuff later)
 export ksdevice=""
-# TODO: Remove support for ksdevice. It's deprecated a long time already
```
poncovka, line 19: It looks like we were planning to completely remove this support. Shouldn't we do that instead?

rvykydal: You mean to remove the functionality below this comment or also the other part of ksdevice functionality:
```
If the --device= option is missing on the first use of the network command, the value of the ksdevice= Anaconda boot option is used, if available. If ksdevice= is not set, link value is used. Note that this is considered deprecated behavior; in most cases, you should always specify a --device= for every network command. The behavior of any subsequent network command in the same Kickstart file is unspecified if its --device= option is missing. Make sure you specify this option for any network command beyond the first.
```
I think you mean the latter.
It has not been removed in RHEL 9 Beta, see https://bugzilla.redhat.com/show_bug.cgi?id=2000140 so it will most likely be kept for GA. Actually this PR would then apply for rhel-9 branch.
For Fedora I'd then open additional followup PR for complete removal so that the history / difference between r9 and master is more clear?

rvykydal:
Fixed the tests.

dracut/parse-anaconda-net.sh
poncovka: I don't think this replacement makes sense. If someone uses `ksdevice` and it's not working, shouldn't they start to use the supported alternatives instead of `inst.ks.device`? For RHEL 9, I would keep the support for `ksdevice` and remove the warning about the missing prefix. The warning is wrong, because `inst.ks.device` was never functional.  With other boot options, we supported both variants before the removal of the non-prefixed versions. That's different.

rvykydal: Agreed, we seem to have come to the same conclusion with @jstodola.

rvykydal: PR removing the warning: https://github.com/rhinstaller/anaconda/pull/3591

PR #3588, opened 2021-09-02
Fix Removed options inst.[product|variant] were subsections

jkonecny12:
The inst.product and `inst.variant` are accidentally one level lower than the rest of the `Removed Options` in the documentation. That will cause having them as the subsection of the `repo=nfsiso` kernel boot option.

_Suggested-by: Alois Mahdal <amahdal>_


VladimirSlavik (approved):
Looks good to me. Thank you!

M4rtinK (approved):
Looks good to me as well.

poncovka (approved):
Thanks!

PR #3590, opened 2021-09-07
we should cancel set/update time timer to avoid accidental modification error when ntp service switch on

bitcoffeeiux:
If we do not cancel the timer, the time may be incorrectly changed by the timer after the NTP service updates the time

### DIFF
--- a/pyanaconda/ui/gui/spokes/datetime_spoke.py
+++ b/pyanaconda/ui/gui/spokes/datetime_spoke.py
@@ -892,6 +892,15 @@ class DatetimeSpoke(FirstbootSpokeMixIn, NormalSpoke):
 
     def on_ntp_switched(self, switch, *args):
         if switch.get_active():
+            # clean timer
+            if self._start_updating_timer:
+                self._start_updating_timer.cancel()
+                self._start_updating_timer = None
+
+            if self._update_datetime_timer:
+                self._update_datetime_timer.cancel()
+                self._update_datetime_timer = None
+
             #turned ON
             if not conf.system.can_set_time_synchronization:
                 #cannot touch runtime system, not much to do here


rvykydal:
I was trying to look more into the issue the patch is addressing and I think this is the case:
If the NTP service is switched on in less than 2 seconds after modifying time/date manually: the time would be updated by the NTP server, then shortly after that (the 2 sec after manual setting) it will be updated by the planned callback to the wrong by the previous manual setting, and then only after rather longer time - for example about 3 minutes in my testing - by the NTP server again.
So the issue here is not critical - it happens under rather limited conditions, and the time is updated by the NTP server in the end, but the delay about 3 minutes seems to be long enough to avoid this case.

See my comments in the review, I think we need to disable only updating of the system time (_start_updating_timer), and on the contrary, make sure that UI updating (update_datetime_timer) is on.

pyanaconda/ui/gui/spokes/datetime_spoke.py
```diff
@@ -892,6 +892,15 @@ def _show_no_ntp_server_warning(self):
...
         if switch.get_active():
+            # clean timer
+            if self._start_updating_timer:
+                self._start_updating_timer.cancel()
+                self._start_updating_timer = None
+
+            if self._update_datetime_timer:
+                self._update_datetime_timer.cancel()
+                self._update_datetime_timer = None
+
```
rvykydal, line 903: Here we don't want to cancel `_update_datetime_timer` (we still want the time in the UI to be updated to the current value) but on the contrary we want to make sure it is running, because it is was cancelled in `_stop_and_maybe_start_time_updating` which planned the `_save_system_time` by` _start_updating_timer`. The `_save_system_time` would re-enable the UI update (`_update_date_timer`) again, but it is not called because the `_start_updating_timer` is cancelled in the patch. So I think the patch could be modified in this way:
```
            # cancel system time update planned by manual setting
            if self._start_updating_timer:
                self._start_updating_timer.cancel()
                self._start_updating_timer = None
            # re-enable UI update because it will not be done by system time update we've
            # just cancelled
            if not self._update_datetime_timer:
                self._update_datetime_timer = Timer()
                self._update_datetime_timer.timeout_sec(1, self._update_datetime)
```

rvykydal (changes requested):
The patch would turn update of the UI with current time off, see the inline comment.

bitcoffeeiux:
You're right. I will change it

rvykydal:
@bitcoffeeiux are you still interested in updating the PR ?

rvykydal:
Thank you for your PR.
I've opened https://github.com/rhinstaller/anaconda/pull/3691 based on this PR, which is addressing my comments.

PR #3591, opened 2021-09-07

rvykydal:
Resolves: rhbz#2000140

Support for inst.ks.device has never been implemented while the ksdevice

We will try to remove also support for this ksdevice functionality, so
it would make no sense to make inst.ks.device working at this point.


poncovka (approved):
Thank you!

jkonecny12 (approved):
Looks great to me. Thanks!

M4rtinK (approved):
Looks good to me as well. :)

rvykydal:
> @rvykydal could you please port this to master :)

https://github.com/rhinstaller/anaconda/pull/3871

PR #3593, opened 2021-09-08

rvykydal:
Resolves: rhbz#2001913

Support for inst.ks.device has never been implemented while the vice

We will try to remove also support for this ksdevice functionality, it
would make no sense to make inst.ks.device working at this point.

### DIFF
--- a/dracut/parse-anaconda-options.sh
+++ b/dracut/parse-anaconda-options.sh
@@ -105,7 +105,6 @@ warn_renamed_arg "stage2" "inst.stage2"
 
 # kickstart
 warn_renamed_arg "ks" "inst.ks"
-warn_renamed_arg "ksdevice" "inst.ks.device"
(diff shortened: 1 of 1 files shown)


jkonecny12 (approved):
LGTM!

poncovka (approved):

M4rtinK:
I think there is a typo in the commit message ?

>Support for inst.ks.device has never been implemented while the **vice**

Or I'm somehow not understanding the sentence correctly. :)

M4rtinK (approved):
Looks good to me as well code wise. :)

rvykydal:
> I think there is a typo in the commit message ?
> 
> > Support for inst.ks.device has never been implemented while the **vice**
> > option is still working as a value used to supply device in case of
> > missing kickstart network --device option.
> 
> Or I'm somehow not understanding the line correctly. :)

Fixed, thank you for the catch :)

PR #3595, opened 2021-09-08
Enable ELN ~smoke tests~ COPR builds again (#infra)

jkonecny12:
This will enable ELN COPR build on PR ~and also try to installation of the package there~.

I would like to re-enable the ELN tests because they are working more than 2 weeks without a problem. Let's try it again with the ELN.

*UPDATE*:
Unfortunately AWS images are still not solved for ELN so we can't use testing functionality of Packit for ELN.

### DIFF
--- a/.packit.yml
+++ b/.packit.yml
@@ -20,7 +20,7 @@ jobs:
     metadata:
       targets:
         - fedora-rawhide
-#        - fedora-eln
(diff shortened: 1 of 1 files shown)


VladimirSlavik (approved):
Let's try! Thanks for keeping this alive.

PR #3596, opened 2021-09-08
Few tweaks to `inst.ks.all` doc

AloisMahdal:
I found the section a hard to read -- I think the first paragraph is a bit ambiguous and does not draw clear  & full picture before moving to examples.

Here's my attempt to fix it in 3 small commits.

### DIFF
--- a/docs/boot-options.rst
+++ b/docs/boot-options.rst
@@ -237,9 +237,9 @@ For example:
 inst.ks.all
 ^^^^^^^^^^^
 
-All locations of type http, https or ftp specified with inst.ks will be used
-sequentially one by one until the kickstart file is fetched. Other locations
-will be ignored.
+All locations of type ``http``, ``https`` or ``ftp`` specified with
+``inst.ks`` will be used sequentially one by one until the kickstart file
+is fetched. Other locations will be ignored.
 
 In the following example, Anaconda will try to fetch the kickstart file at
 first from ``http://a/a.ks``, then from ``http://b/b.ks`` and finally from
--- a/docs/boot-options.rst
+++ b/docs/boot-options.rst
@@ -238,8 +238,8 @@ inst.ks.all
 ^^^^^^^^^^^
 
 Use all locations of type ``http``, ``https`` or ``ftp`` specified with
-multiple ``inst.ks`` sequentially one by one until kickstart file is fetched. Other
-locations will be ignored.
+multiple ``inst.ks`` sequentially one by one until kickstart file is fetched.
+Locations of other types (eg. ``nfs``) will be ignored.
 
 Without this option, only last location specified by ``inst.ks`` is used.
 
--- a/docs/boot-options.rst
+++ b/docs/boot-options.rst
@@ -238,7 +238,7 @@ inst.ks.all
 ^^^^^^^^^^^
 
 Use all locations of type ``http``, ``https`` or ``ftp`` specified with
-``inst.ks`` sequentially one by one until kickstart file is fetched. Other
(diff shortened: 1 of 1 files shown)


poncovka (approved):
Looks good to me. What do you think, @VladimirSlavik?

VladimirSlavik (approved):
Looks good to me. Thank you! Could be even better though ;-)

* You could also mention that the " all locations" are specified with *multiple* `inst.ks`. That information is missing, although it's easy to infer from the example.
* I think the "Other locations will be ignored." could work better as "The *remaining* locations will be ignored.".

It's a solid improvement already, so don't feel like any of that is needed to get this accepted.

AloisMahdal:
> Looks good to me. Thank you! Could be even better though ;-)
> 
> * You could also mention that the " all locations" are specified with _multiple_ `inst.ks`. That information is missing, although it's easy to infer from the example.

Good point, will do!

> * I think the "Other locations will be ignored." could work better as "The _remaining_ locations will be ignored.".

Interesting, what I actually inferred was that *Other* here is referring to other *location types* (ie. not `http`, `https`, `nfs`.)  If that's not the case -- and `inst.ks.all` indeed should work with any location types -- then I think there would not be need to list these three types specifically and the whole paragraph could be just rewritten to something simpler.

Can you please confirm?

> 
> It's a solid improvement already, so don't feel like any of that is needed to get this accepted.

Thanks :)

poncovka:
> > * I think the "Other locations will be ignored." could work better as "The _remaining_ locations will be ignored.".
> 
> Interesting, what I actually inferred was that _Other_ here is referring to other _location types_ (ie. not `http`, `https`, `nfs`.)

This is correct. What about "Locations of other types will be ignored, for example, `ftp`."? I am not sure whether _remaining_ would make sense in this sentence. @VladimirSlavik?

docs/boot-options.rst
```diff
@@ -237,9 +237,11 @@ For example:
...
 ^^^^^^^^^^^
 
-All locations of type http, https or ftp specified with inst.ks will be used
-sequentially one by one until the kickstart file is fetched. Other locations
-will be ignored.
+Use all locations of type ``http``, ``https`` or ``ftp`` specified with
+multiple ``inst.ks`` sequentially one by one until kickstart file is fetched. Other
+locations will be ignored.
+
+Without this option, only first location specified by ``inst.ks`` is used.
```
poncovka, line 244: For historical reasons, it is the last location.

AloisMahdal: thanks, fixed

AloisMahdal:
> > Interesting, what I actually inferred was that _Other_ here is referring to other _location types_ (ie. not `http`, `https`, `nfs`.)
> 
> This is correct. What about "Locations of other types will be ignored, for example, `ftp`."? I am not sure whether _remaining_ would make sense in this sentence. @VladimirSlavik?

I added one more commit, going with:

> Locations of other types (eg. ``nfs``) will be ignored.

(I made a mistake in my example, leading you to post incorrect example :)  .. `ftp` is sainted, `nfs` is correct  example of *ignored type*.)

VladimirSlavik:
> This is correct. What about "Locations of other types will be ignored, for example, ftp."? I am not sure whether remaining would make sense in this sentence. @VladimirSlavik?

Yes, I was wrong. For this logic, "other" is clearly better here.

A question might remain - what happens to the `inst.ks` values that have not been tried yet. I'm not sure if that needs clarification too.

poncovka (approved):
I like the current version.

VladimirSlavik:
I think we're all ready for merging then?

PR #3598, opened 2021-09-08
Verify the OPAL compatibility with XFS features (#1997832)

poncovka:
Shows a warning on ppc64le systems with old petitboot, when the /boot
filesystem is formatted with XFS. The system won't be bootable.

Resolves: rhbz#1997832

### DIFF
--- a/pyanaconda/modules/storage/checker/utils.py
+++ b/pyanaconda/modules/storage/checker/utils.py
@@ -20,6 +20,7 @@ gi.require_version("BlockDev", "2.0")
 from gi.repository import BlockDev as blockdev
 
 from collections import defaultdict
+from distutils.version import LooseVersion
 
 from blivet import arch, util
 from blivet.devicefactory import get_device_type
@@ -223,6 +224,66 @@ def verify_gpt_biosboot(storage, constraints, report_error, report_warning):
                                "partition to boot from a GPT disk label. "
                                "To continue, please create a 1MiB "
                                "'biosboot' type partition."))
+
+
+def verify_opal_compatibility(storage, constraints, report_error, report_warning):
+    """ Verify the OPAL compatibility.
+
+    :param storage: a storage to check
+    :param constraints: a dictionary of constraints
+    :param report_error: a function for error reporting
+    :param report_warning: a function for warning reporting
+    """
+    if arch.get_arch() == "ppc64le" and arch.is_powernv():
+        # Check the kernel version.
+        version = _get_opal_firmware_kernel_version()
+        if _check_opal_firmware_kernel_version(version, "5.10"):
+            return
+
+        # Is /boot on XFS?
+        dev = storage.mountpoints.get("/boot") or storage.mountpoints.get("/")
+        if dev and dev.format and dev.format.type == "xfs":
+            report_warning(_(
+                "Your firmware doesn't support XFS file system features "
+                "on the /boot file system. The system will not be bootable. "
+                "Please, upgrade the firmware or change the file system type."
+            ))
+
+
+def _check_opal_firmware_kernel_version(detected_version, required_version):
+    """Check the firmware kernel version for OPAL systems.
+
+    :param detected_version: a string with the detected kernel version or None
+    :param required_version: a string with the required kernel version or None
+    :return: True or False
+    """
+    try:
+        if detected_version and required_version:
+            return LooseVersion(detected_version) >= LooseVersion(required_version)
+    except Exception as e:  # pylint: disable=broad-except
+        log.warning("Couldn't check the firmware kernel version: %s", str(e))
+
+    return False
+
+
+def _get_opal_firmware_kernel_version():
+    """Get the firmware kernel version for OPAL systems.
+
+    For example: 5.10.50-openpower1-p59fd803
+
+    :return: a string with the kernel version or None
+    """
+    version = None
+
+    try:
+        with open("/proc/device-tree/ibm,firmware-versions/linux") as f:
+            version = f.read().strip().removeprefix("v")
+            log.debug("The firmware kernel version is '%s'.", version)
+
+    except IOError as e:
+        log.warning("Couldn't get the firmware kernel version: %s", str(e))
+
+    return version
 
 
 def verify_swap(storage, constraints, report_error, report_warning):
@@ -661,6 +722,7 @@ class StorageChecker(object):
         self.add_check(verify_partition_format_sizes)
         self.add_check(verify_bootloader)
         self.add_check(verify_gpt_biosboot)
+        self.add_check(verify_opal_compatibility)
         self.add_check(verify_swap)
         self.add_check(verify_swap_uuid)
         self.add_check(verify_mountpoints_on_linuxfs)
--- a/tests/nosetests/pyanaconda_tests/modules/storage/module_storage_checker_test.py
+++ b/tests/nosetests/pyanaconda_tests/modules/storage/module_storage_checker_test.py
@@ -18,15 +18,22 @@
 # Red Hat Author(s): Vendula Poncova <vponcova@redhat.com>
 #
 import unittest
-from unittest.mock import patch, Mock
+from functools import partial
+from unittest.mock import patch, Mock, mock_open, PropertyMock
 
+from blivet.devices import StorageDevice
+from blivet.formats import get_format
+from blivet.formats.fs import XFS
 from blivet.size import Size
 from dasbus.typing import get_variant, Int
 from pyanaconda.core.constants import STORAGE_MIN_RAM, STORAGE_LUKS2_MIN_RAM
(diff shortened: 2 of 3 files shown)


VladimirSlavik (approved):
Looks good to me. Thank you!

M4rtinK (approved):

jkonecny12 (approved):
Looks great to me. Thanks!
Could you please attach an image of the warning here?

pyanaconda/modules/storage/checker/utils.py
```diff
@@ -225,6 +227,52 @@ def verify_gpt_biosboot(storage, constraints, report_error, report_warning):
...
+    """ Verify the petitboot version compatibility.
+
+    :param storage: a storage to check
+    :param constraints: a dictionary of constraints
+    :param report_error: a function for error reporting
+    :param report_warning: a function for warning reporting
+    """
+    if arch.get_arch() == "ppc64le" and os.access("/sys/firmware/opal", os.F_OK):
+        # Check the version of the petitboot loader.
+        version = _get_petitboot_version()
```
jstodola, line 240: I don't think we care about the petitboot version at all. We need to ensure that the firmware kernel version (for OPAL systems) is 5.10 or newer. The kernel version can be found in `/proc/device-tree/ibm,firmware-versions/linux` or by executing `lsmcode` - see bug [1977236](https://bugzilla.redhat.com/show_bug.cgi?id=1977236).
@sharkcz , can you please confirm?

jstodola (changes requested):

pyanaconda/modules/storage/checker/utils.py
sharkcz: - @jstodola is right, it's kernel version in the OPAL firmware that matters
- I would call the function/method `verify_opal_compatibility`
- the check for arch and `/sys/firmware/opal` could be switched to `if get_platform() == 'PowerNV'`
- if you can't positively detect the kernel version, then I would assume it's "bad/old"

poncovka: Thanks for the pointers! I have got confused by https://bugzilla.redhat.com/show_bug.cgi?id=1997832#c2. Hopefully, it is fine now.

pyanaconda/modules/storage/checker/utils.py
```diff
@@ -225,6 +226,50 @@ def verify_gpt_biosboot(storage, constraints, report_error, report_warning):
...
+                "Your firmware doesn't support XFS file system features "
+                "on the /boot file system. The system will not be bootable. "
+                "Please, upgrade the firmware or change the file system type."
+            ))
+
+
+def _get_opal_firmware_kernel_version():
+    """Get the firmware kernel version for OPAL systems.
+
+    For example: 5.10.50-openpower1-p59fd803
```
jstodola, line 256: Please, be aware that the version could also be something like "v4.15.9-openpower1-p9e03417" (see https://bugzilla.redhat.com/show_bug.cgi?id=1977236#c16) - in that case comparing of versions would throw a traceback:
```
>>> LooseVersion("5.10") > LooseVersion("v4.15.9-openpower1-p9e03417")
  File "/usr/lib64/python3.9/distutils/version.py", line 64, in __gt__
    c = self._cmp(other)
  File "/usr/lib64/python3.9/distutils/version.py", line 341, in _cmp
    if self.version < other.version:
TypeError: '<' not supported between instances of 'int' and 'str'
```

sharkcz: I think there are 2 options, either check the string from `ibm,firmware-versions/linux` against `<num>.<num>.<num>-something` format or wrap the comparison in a `try` block. This is what I meant with the positive identification of 5.10+ kernel. If the installer can't be sure about the version, then it's safer to expect system without new-enough XFS support.

poncovka: I have put back the code that removes `v` from the version and moved the version comparison into a try-except block.

tests/nosetests/pyanaconda_tests/modules/storage/module_storage_checker_test.py
```diff
@@ -108,3 +115,117 @@ def lvm_verification_test(self):
...
+        mocked_arch.get_arch.return_value = "ppc64le"
+        mocked_arch.is_powernv.return_value = False
+
+        verify_opal_compatibility(None, {}, None, reporter)
+        reporter.assert_not_called()
+
+    @patch("pyanaconda.modules.storage.checker.utils._get_opal_firmware_kernel_version")
+    @patch("pyanaconda.modules.storage.checker.utils.arch")
+    def opal_verification_version_test(self, mocked_arch, version_getter):
+        """Check verify_opal_compatibility with a higher firmware version."""
```
VladimirSlavik, line 176: I don't think the test name matches the docstring.

tests/nosetests/pyanaconda_tests/modules/storage/module_storage_checker_test.py
```diff
@@ -108,3 +115,117 @@ def lvm_verification_test(self):
...
+        version_getter.return_value = "5.10.50-openpower1-p59fd803"
+
+        verify_opal_compatibility(None, {}, None, reporter)
+        reporter.assert_not_called()
+
+    @patch.object(XFS, "mountable", new_callable=PropertyMock)
+    @patch("pyanaconda.modules.storage.checker.utils._get_opal_firmware_kernel_version")
+    @patch("pyanaconda.modules.storage.checker.utils.arch")
+    def opal_verification_test(self, mocked_arch, version_getter, xfs_mountable):
+        """Check verify_opal_compatibility."""
```
VladimirSlavik, line 189: And this test is actually for a too low version, but we can't learn that here.

VladimirSlavik (commented):
I wonder what version strings are out there in the wild, expecting just "v" makes me nervous.

The code outgrew the test names and docstrings, I guess, but that's a minor thing.

tests/nosetests/pyanaconda_tests/modules/storage/module_storage_checker_test.py
poncovka: Fixed.

tests/nosetests/pyanaconda_tests/modules/storage/module_storage_checker_test.py
poncovka: Fixed.

poncovka:
> I wonder what version strings are out there in the wild, expecting just "v" makes me nervous.

The current code should be able to survive any other weird strings that and show the warning as suggested in https://github.com/rhinstaller/anaconda/pull/3598#discussion_r708975700.

VladimirSlavik (approved):
Thank you!

I agree that the error message is as good as it can get without actual data about past versions.

poncovka:
Screenshot from 2021-09-15 16-40-58

sharkcz:
Looks good, thanks. Ideally we could a link to a KBase article to the message, but it hasn't been written yet (but is planned).
How does the automatic partitioning work? Does it default to ext4 or does it require attention from the user?

poncovka:
> How does the automatic partitioning work? Does it default to ext4 or does it require attention from the user?

With the automatic partitioning, the status of the Installation Destination spoke in the Summary hub informs about warnings in the storage configuration. It is necessary to reenter the Installation Destination spoke to see the warning.

Screenshot from 2021-09-13 14-24-53

jstodola (approved):
I have not tried any manual testing, but I cannot see any other possible issues. Thanks!

sharkcz:
OK, and for kickstart installation it means manually setting up the disk layout and not relying on on `autopart`, right? I am thinking how this can affect beaker provisioning and similar situation ...

poncovka:
> OK, and for kickstart installation it means manually setting up the disk layout and not relying on on `autopart`, right? I am thinking how this can affect beaker provisioning and similar situation ...

It is possible to use `autopart --fstype ext4`. Just a note, the warning doesn't block the installation. It just informs the user about the issue, but it is possible to ignore the warning and continue with the current settings. If we want to enforce the change, we need to report an error from the verification function. However, @jkonecny12 mentioned that the warning is sufficient in this case.

jkonecny12:
Wrote a comment on the bug to make it totally clear that we are just adding non-blocking warning. To avoid confusion.

sharkcz:
This should go to the master branch as well, the problem is not specific to RHEL, it affects Fedora too.

VladimirSlavik:
> This should go to the master branch as well, the problem is not specific to RHEL, it affects Fedora too.

No worries! The `port to Fedora` label means exactly that.

sharkcz:
Thanks :-)

poncovka:
Ported to upstream in https://github.com/rhinstaller/anaconda/pull/3620.

PR #3599, opened 2021-09-09
SetUpCdromSourceTask: Allow setup source to all mountable partitions

slayercat:
see also: https://github.com/rhinstaller/anaconda/discussions/3594

### DIFF
--- a/pyanaconda/modules/payloads/source/cdrom/initialization.py
+++ b/pyanaconda/modules/payloads/source/cdrom/initialization.py
@@ -51,13 +51,13 @@ class SetUpCdromSourceTask(SetUpMountTask):
         device_name = self._choose_installation_device(device_tree, device_candidates)
 
         if not device_name:
-            raise SourceSetupError("Found no CD-ROM")
+            raise SourceSetupError("Found no CD-ROM source")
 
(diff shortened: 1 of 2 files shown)


slayercat:
As discuss in #3594 , we made some code to prove our idea.

This patch has been tested offline with `CentOS-Stream-8-aarch64-20210902-dvd1`.  (basicly)

With  make an `efi` (`FAT32`) partition, and  copies all files in the ISO file to it,  It'll boot and install in a  aarch64 machine.

you may want to view these log files:

[tmp.zip](https://github.com/rhinstaller/anaconda/files/7135285/tmp.zip)

jkonecny12:
Hi thanks for the PR, however, switching CDRom source from Optical Media to mountable device would change behavior of the CDRom source. In that case we could discover completely different sources for `inst.cdrom`. For this reason we should instead follow solution that we would add a new source which would re-use existing mount from a Dracut. However, even in that case we had to make sure that this do not break some use-cases.

jkonecny12:
Just now I'm working on a alternative solution to this. Hopefully, I'll be able to create the PR soon. Closing this PR for now.

PR #3600, opened 2021-09-09
Generate LVM devices file

VladimirSlavik:
Resolves: [rhbz#2002550](https://bugzilla.redhat.com/show_bug.cgi?id=2002550) - or not. This is the first attempt, let's see if the file is needed in initramfs too, then we'd need a task.

Do not port to Fedora.

### DIFF
--- a/data/post-scripts/70-generate-lvm-device-file.ks
+++ b/data/post-scripts/70-generate-lvm-device-file.ks
@@ -1,0 +1,11 @@
+# Generate LVM devices file
+# /etc/lvm/devices/system.devices
+# This is a temporary workaround and should be reverted once blivet supports
+# manipulation of the file.
+# See: https://bugzilla.redhat.com/show_bug.cgi?id=2002550
+
+%post
+
+[ -e /usr/sbin/vgimportdevices ] && vgimportdevices -a
+
+%end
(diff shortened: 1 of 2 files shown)


data/post-scripts/70-generate-lvm-device-file.ks
```diff
@@ -0,0 +1,9 @@
+# Generate LVM devices file
+# /etc/lvm/devices/system.devices
+# See: https://bugzilla.redhat.com/show_bug.cgi?id=2002550
+
+%post
+
+vgimportdevices -a
```
M4rtinK, line 7: I think it would be good to check the tool is there or make this non-fatal if it can't be invoked.

vojtechtrefny (commented):
Looks good. I'm still waiting for LVM team to tell us whether we need the system.devices file when creating the initramfs image or not.

vojtechtrefny (approved):
The system.devices file is not needed in the initramfs so the post script solution will work.

poncovka (commented):
I had to read the bug to understand why this change needs to be done in Anaconda and not in Blivet. From what I understand, it is a temporary workaround. It would be great to mention this in the commit message and the script.

VladimirSlavik:
Added explanations to the commit and file.

poncovka (approved):
Thanks!

M4rtinK (approved):
Looks good to me, thanks! :)

jkonecny12:
The bug has all the ACKs.

PR #3603, opened 2021-09-13
Add AlmaLinux product support (rhel-9)

andrewlukoshko:
AlmaLinux is Open Source, community-governed and forever-free enterprise Linux distribution, focused on long-term stability, providing a robust production-grade platform. AlmaLinux OS is 1:1 binary compatible with RHEL® and pre-Stream CentOS.
https://almalinux.org


andrewlukoshko:
@VladimirSlavik I've created a bug report for RHEL 9 and mentioned it in commit message

carlwgeorge:
@VladimirSlavik What CentOS Stream policies?  There have been changes merged here and in other upstreams to support CentOS needs (not just CentOS Stream) that had no relevance to RHEL.  It's only fair to do the same for Alma.  For whatever it's worth, as a CentOS maintainer I endorse merging this PR and #3562.

VladimirSlavik:
Well, so do I, as well as the rest of the dev team, so we're all on the same page here it seems :-)

As to the "policies", perhaps that was too strong a word. What I meant was that there must be some way to get in changes unrelated to RHEL proper, and CentOS stream is supposedly exactly that.

The issue with this change is that it touches code, so it's subject to more processes that we'd all like. It's certain this won't get into 9.0 Beta. We're far enough that this would need business justification, and it simply won't be vouched for across all the necessary functions. But it seems there should be a window before GA to get this in. If I'm wrong on that, 9.1 is a certainty.

VladimirSlavik (approved):

VladimirSlavik:
Setting blocked to reflect the current bug status.

carlwgeorge:
Yup, I absolutely understand the hesitation to rush it into an upcoming minor release or beta release, as you said that often gets pushback and requires business justification, and rightfully so.  What I was commenting on was your statement implying that the rhel8 PR might not get merged at all.  rhel8 is still in the "full support" phase, so I don't see why it can't be merged eventually.  To your point, I don't expect it for 8.5 this late in its cycle, but from my perspective it seems like a good fit for 8.6 or 8.7.  Same for rhel9, I don't think there is justification to get it into 9 beta, but 9.0 or 9.1 do make sense to me.

jkonecny12 (approved):
Looks good to me.

andrewlukoshko:
Thanks to all involved!

PR #3606, opened 2021-09-13
The NTP server dialog without entries shouldn't crash (#2001591)

poncovka:
Don't crash if there are no NTP servers in the dialog.

Resolves: rhbz#2001591


VladimirSlavik (approved):
Looks good to me, thank you for the fix!

poncovka:
Resolved pylint issues.

VladimirSlavik:
IMO, feel free to merge. The code did not change such that a re-review would be needed...

PR #3608, opened 2021-09-14
Fix boolean comparisons in asserts

VladimirSlavik:
These were previously changed in bulk after switch to pytest.

In other words:
```s/assert ([^\s]+) == (True|False)/assert $1 is $2/```


poncovka (approved):
Thanks!

PR #3611, opened 2021-09-17
Do not stretch NTP toggle in GUI

VladimirSlavik:
Also space out the controls a bit.

Resolves: [rhbz#1972077](https://bugzilla.redhat.com/show_bug.cgi?id=1972077)

### DIFF
--- a/pyanaconda/ui/gui/spokes/datetime_spoke.glade
+++ b/pyanaconda/ui/gui/spokes/datetime_spoke.glade
@@ -186,6 +186,7 @@
                               <object class="GtkLabel" id="cityLabel">
                                 <property name="visible">True</property>
                                 <property name="can_focus">False</property>
+                                <property name="margin_start">6</property>
                                 <property name="label" translatable="yes" context="GUI|Date and Time">_City:</property>
                                 <property name="use_underline">True</property>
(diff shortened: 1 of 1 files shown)


VladimirSlavik:
Here's where the differences are. Lines are hastily scribbled, so not entirely accurate.

**Before**

Screenshot_vs_main_2021-09-17_15:41:37

**After**

ntp gui cleanup

M4rtinK (approved):
Looks much better now, thanks! :)

pyanaconda/ui/gui/spokes/datetime_spoke.glade
```diff
@@ -259,6 +260,8 @@
                                       <object class="GtkSwitch" id="networkTimeSwitch">
                                         <property name="visible">True</property>
                                         <property name="can_focus">True</property>
+                                        <property name="valign">center</property>
+                                        <property name="margin_end">6</property>
```
poncovka, line 264: Shouldn't we disable the expansion of the widget instead of adding the margins?

VladimirSlavik: The margins are in the other direction - horizontal, while the growth is vertical.

PR #3612, opened 2021-09-20
Fix task to copy logs

VladimirSlavik:
See https://github.com/rhinstaller/anaconda/pull/3571#issuecomment-922119290

CC @AdamWill 

Note that kickstart tests are blind to this: The crash happens only if input kickstart can be copied but none was actually used.

### DIFF
--- a/pyanaconda/modules/boss/installation.py
+++ b/pyanaconda/modules/boss/installation.py
@@ -135,7 +135,7 @@ class CopyLogsTask(Task):
         """Set access bits"""
         items = glob.glob(join_paths(self._sysroot, ANACONDA_LOG_DIR, '*'))
         for item in items:
-            os.chmod(item, 0x0600)
+            os.chmod(item, 0o0600)
 
     def _copy_kickstart(self):
         """Copy input kickstart file"""
@@ -145,7 +145,7 @@ class CopyLogsTask(Task):
                 "/run/install/ks.cfg",
                 "/root/original-ks.cfg"
             )
-            os.chmod(join_paths(self._sysroot, "/root/original-ks.cfg"), 0x0600)
+            os.chmod(join_paths(self._sysroot, "/root/original-ks.cfg"), 0o0600)
         else:
             log.warning("Input kickstart will not be saved to the installed system due to the "
                         "nosave option.")
--- a/pyanaconda/modules/boss/installation.py
+++ b/pyanaconda/modules/boss/installation.py
@@ -70,7 +70,6 @@ class CopyLogsTask(Task):
         self._copy_dnf_debugdata()
         self._copy_post_script_logs()
         self._dump_journal()
-        self._chmod_logs()
 
     def _create_logs_directory(self):
         """Create directory for Anaconda logs on the install target"""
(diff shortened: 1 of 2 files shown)


M4rtinK (approved):
Looks good to me, thanks! :)

pyanaconda/modules/boss/installation.py
```diff
@@ -145,7 +145,9 @@ def _copy_kickstart(self):
                 "/run/install/ks.cfg",
                 "/root/original-ks.cfg"
             )
-            os.chmod(join_paths(self._sysroot, "/root/original-ks.cfg"), 0x0600)
+            result_ks_path = join_paths(self._sysroot, "/root/original-ks.cfg")
+            if os.path.exists(result_ks_path):
```
poncovka, line 149: This is confusing. If the original kickstart file doesn't exist, we shouldn't copy it in the first place.

pyanaconda/modules/boss/installation.py
```diff
@@ -135,7 +135,7 @@ def _chmod_logs(self):
         """Set access bits"""
         items = glob.glob(join_paths(self._sysroot, ANACONDA_LOG_DIR, '*'))
         for item in items:
-            os.chmod(item, 0x0600)
+            os.chmod(item, 0o0600)
```
poncovka, line 138: It looks like it would be useful to have this in a separate function that can handle and log the exception.

pyanaconda/modules/boss/installation.py
VladimirSlavik: The copying function checks for that itself. It's just that here we have to do more. Maybe it's a better idea to inline it all?

VladimirSlavik:
Draftifying because there's some more refactoring to do with this, and tests will be not happy for a while.

VladimirSlavik:
Ready for review now.

I think the `_copy_file_to_sysroot` function should perhaps be called `_try_copy_file_to_sysroot` to show that it will not fail for missing files. Feel free to request the change. However it's independent of the functionality.

pyanaconda/modules/boss/installation.py
```diff
@@ -158,29 +153,35 @@ def _relabel_files(self):
...
         if os.path.exists(src):
             log.info("Copying directory tree: %s -> %s", src, dest)
+            full_dest_path = join_paths(self._sysroot, dest)
             shutil.copytree(
                 src,
-                join_paths(self._sysroot, dest),
+                full_dest_path,
                 dirs_exist_ok=True
             )
+            os.chmod(full_dest_path, mode)
```
poncovka, line 187: Does it work fine if `chmod` is called only for the top directory?

pyanaconda/modules/boss/installation.py
```diff
@@ -128,24 +127,20 @@ def _copy_post_script_logs(self):
 
     def _dump_journal(self):
         """Dump journal from the installation environment"""
-        with open(join_paths(self._sysroot, ANACONDA_LOG_DIR, "journal.log"), "w") as logfile:
+        temp = "/tmp/anaconda-journal-dump"
```
poncovka, line 130: I think you can name it `journal.log`.

pyanaconda/modules/boss/installation.py
```diff
@@ -128,24 +127,20 @@ def _copy_post_script_logs(self):
...
 
     def _copy_kickstart(self):
         """Copy input kickstart file"""
         if conf.target.can_copy_input_kickstart:
             log.info("Copying input kickstart file.")
             self._copy_file_to_sysroot(
                 "/run/install/ks.cfg",
-                "/root/original-ks.cfg"
+                "/root/original-ks.cfg",
+                0o0700
```
poncovka, line 142: Is the change of the permission expected? I am not sure where it comes from.

poncovka (commented):
Usually, names of the private methods are not so important. You can keep them as they are.

pyanaconda/modules/boss/installation.py
VladimirSlavik: I wondered about this too. But it works exactly as before, because chmod in bash is not recursive by default.

pyanaconda/modules/boss/installation.py
VladimirSlavik: IIRC the bits became different with the python code doing the copying, so I added this to compensate.

jkonecny12: Honestly this seems like a bug to me. Why we want to have executable bit for kickstart. I can't see a reason for that.
So, or so, it doesn't belong to this changeset.

poncovka (approved):
Thanks!

pyanaconda/modules/boss/installation.py
VladimirSlavik: Blah, you are right. I wonder where did that 0700 come from then. Let's remove it, and with it the parameter, etc.

poncovka (approved):

PR #3613, opened 2021-09-20
Set an upper bound to entered sizes (#1992585)

poncovka:
A device size in bytes has to be in a range of the UInt64 type. Values outside
of this range will trigger the overflow error when they will be propagated on
DBus. Set an upper bound to entered sizes to prevent this error.

Resolves: rhbz#1992585


VladimirSlavik (approved):
Thank you, LGTM.

16 EiB doesn't seem to be far from recent storage sizes, but that's probably good enough for now.

jkonecny12 (approved):
Looks good to me.

PR #3617, opened 2021-09-22
Payload should wait for all storage related threads to finish

M4rtinK:
Otherwise a race condition can happen when some of the storage related
threads run at the same time as the payload thread and both interact
with the Blivet storage instance.

This storage instance is not thread safe and two threads interacting
with it at the same time can result in a deadlock.

By waiting for all storage related threads to finish before running
the payload thread we can avoid this race condition and the resulting
deadlock.

Thanks a lot to Honza Stodola and Vendula Poncova for helping me track
down this kinda tricky race condition! :)

Resolves: rhbz#2002203

### DIFF
--- a/pyanaconda/payload/manager.py
+++ b/pyanaconda/payload/manager.py
@@ -22,7 +22,8 @@ from enum import IntEnum
 
 from dasbus.error import DBusError
 from pyanaconda.core.constants import THREAD_STORAGE, THREAD_PAYLOAD, THREAD_PAYLOAD_RESTART, \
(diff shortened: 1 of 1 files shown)


poncovka (approved):
Looks good to me.

VladimirSlavik (approved):
Looks good to me. Thank you!

jkonecny12 (approved):
Great job both on solving this so fast! Looks good to me.

M4rtinK:
Port to RHEL 9: #3629
Port to Fedora: #3630

PR #3618, opened 2021-09-22
Split string helpers from pyanaconda.core.util

VladimirSlavik:
SSIA. This is mostly just shuffling things around. edit: And then removal of most checking for bytes.

~~Still TODO~~ done: 
- [x] Rename the functions to `normal_case` from `camelCase`.
- [x] More tests.
- [x] Organize the tests.
- [x] Remove `decode_bytes`
- [x] Move or remove  `ensure_str`

~~I foresee these as additional commits.~~

### DIFF
--- a/pyanaconda/core/string.py
+++ b/pyanaconda/core/string.py
@@ -92,7 +92,6 @@ def upper_ascii(s):
     :param str s: input string
     :return str: ascii-only uppercased value of s
     """
-    s = ensure_str(s)
     return str.translate(_to_ascii(s), _ascii_upper_table)
 
 
@@ -105,7 +104,6 @@ def lower_ascii(s):
     :param str s: input string
     :return str: ascii-only lowercased value of s
     """
-    s = ensure_str(s)
     return str.translate(_to_ascii(s), _ascii_lower_table)
 
 
--- a/pyanaconda/core/string.py
+++ b/pyanaconda/core/string.py
@@ -16,7 +16,7 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #
-import string  # pylint: disable=deprecated-module
+import string
 import unicodedata
 
 
--- a/pyanaconda/core/string.py
+++ b/pyanaconda/core/string.py
@@ -152,10 +152,6 @@ def have_word_match(str1, str2):
         # non-empty string cannot be found in an empty string
         return False
 
-    # Convert both arguments to string if not already
-    str1 = ensure_str(str1)
-    str2 = ensure_str(str2)
-
     str1 = str1.lower()
     str1_words = str1.split()
     str2 = str2.lower()
--- a/pyanaconda/core/string.py
+++ b/pyanaconda/core/string.py
@@ -1,0 +1,176 @@
+#
+# string.py - generic string utility functions
+#
+# Copyright (C) 2021  Red Hat, Inc.  All rights reserved.
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+import string  # pylint: disable=deprecated-module
+import sys
+import unicodedata
+
+
+def strip_accents(s):
+    """This function takes arbitrary unicode string
+    and returns it with all the diacritics removed.
+
+    :param s: arbitrary string
+    :type s: str
+
+    :return: s with diacritics removed
+    :rtype: str
+
+    """
+    return ''.join(c for c in unicodedata.normalize('NFD', s)
+                   if unicodedata.category(c) != 'Mn')
+
+
+def ensure_str(str_or_bytes, keep_none=True):
+    """
+    Returns a str instance for given string or ``None`` if requested to keep it.
+
+    :param str_or_bytes: string to be kept or converted to str type
+    :type str_or_bytes: str or bytes
+    :param bool keep_none: whether to keep None as it is or raise ValueError if
+                           ``None`` is passed
+    :raises ValueError: if applied on an object not being of type bytes nor str
+                        (nor NoneType if ``keep_none`` is ``False``)
+    """
+
+    if keep_none and str_or_bytes is None:
+        return None
+    elif isinstance(str_or_bytes, str):
+        return str_or_bytes
+    elif isinstance(str_or_bytes, bytes):
+        return str_or_bytes.decode(sys.getdefaultencoding())
+    else:
+        raise ValueError("str_or_bytes must be of type 'str' or 'bytes', not '%s'"
+                         % type(str_or_bytes))
+
+
+# Define translations between ASCII uppercase and lowercase for
+# locale-independent string conversions. The tables are 256-byte string used
+# with str.translate. If str.translate is used with a unicode string,
+# even if the string contains only 7-bit characters, str.translate will
+# raise a UnicodeDecodeError.
+_ASCIIlower_table = str.maketrans(string.ascii_uppercase, string.ascii_lowercase)
+_ASCIIupper_table = str.maketrans(string.ascii_lowercase, string.ascii_uppercase)
+
+
+def _toASCII(s):
+    """Convert a unicode string to ASCII"""
+    if isinstance(s, str):
+        # Decompose the string using the NFK decomposition, which in addition
+        # to the canonical decomposition replaces characters based on
+        # compatibility equivalence (e.g., ROMAN NUMERAL ONE has its own code
+        # point but it's really just a capital I), so that we can keep as much
+        # of the ASCII part of the string as possible.
+        s = unicodedata.normalize('NFKD', s).encode('ascii', 'ignore').decode("ascii")
+    elif not isinstance(s, bytes):
+        s = ''
+    return s
+
+
+def upperASCII(s):
+    """Convert a string to uppercase using only ASCII character definitions.
+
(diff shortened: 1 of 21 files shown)


VladimirSlavik:
I am not sure if there is any renaming that would make sense. There are `lowerASCII` and `upperASCII` but I think these names are fine. Then there are `_toASCII`, `_ASCIIlower_table`, and `_ASCIIupper_table`, and these are internals, so they don't really matter.

pyanaconda/core/string.py
```diff
@@ -0,0 +1,176 @@
...
+        raise ValueError("str_or_bytes must be of type 'str' or 'bytes', not '%s'"
+                         % type(str_or_bytes))
+
+
+# Define translations between ASCII uppercase and lowercase for
+# locale-independent string conversions. The tables are 256-byte string used
+# with str.translate. If str.translate is used with a unicode string,
+# even if the string contains only 7-bit characters, str.translate will
+# raise a UnicodeDecodeError.
+_ASCIIlower_table = str.maketrans(string.ascii_uppercase, string.ascii_lowercase)
```
poncovka, line 67: Please, use always `ascii` instead of `ASCII`.

pyanaconda/core/string.py
```diff
@@ -0,0 +1,176 @@
...
+        # compatibility equivalence (e.g., ROMAN NUMERAL ONE has its own code
+        # point but it's really just a capital I), so that we can keep as much
+        # of the ASCII part of the string as possible.
+        s = unicodedata.normalize('NFKD', s).encode('ascii', 'ignore').decode("ascii")
+    elif not isinstance(s, bytes):
+        s = ''
+    return s
+
+
+def upperASCII(s):
```
poncovka, line 85: `upper_ascii` is easier to read.

pyanaconda/core/string.py
```diff
@@ -0,0 +1,176 @@
...
+
+    :return: s with diacritics removed
+    :rtype: str
+
+    """
+    return ''.join(c for c in unicodedata.normalize('NFD', s)
+                   if unicodedata.category(c) != 'Mn')
+
+
+def ensure_str(str_or_bytes, keep_none=True):
```
poncovka, line 39: I think we can get rid of `ensure_str` and always use `decode_bytes`. The function is always called with `keep_none=True`, so we just need to handle `None`. The `ProxyString` class can have its own private method, `have_word_match` already handles `None` values and `_toASCII ` seems to handle it as well.

Please, don't forget to check all related projects.

pyanaconda/core/string.py
```diff
@@ -0,0 +1,176 @@
...
+    """
+    Returns a str instance for given string or ``None`` if requested to keep it.
```
jkonecny12, line 41: Please fix also the doc strings when you are doing these changes.

jkonecny12 (commented):
Looks good to me otherwise.

poncovka (commented):
Maybe, we could move there also `setup_translations`.

pyanaconda/core/string.py
VladimirSlavik: I can't tell what was wrong with this one. But I will change the code anyway, so it's not important.

pyanaconda/core/string.py
VladimirSlavik: It's a mess. It appears `decode_bytes` was added in parallel to `ensure_str,`, which makes some sense. The difference is handling of `None` and crucially, encoding used on bytes. So reconciling these two is a bit of apples and oranges in my opinion.

The fact that `ensure_str` is hidden in a number of calls means there are many places that can stealthily take `bytes` and make assumptions about encoding. I'll try to make sense of these first.

VladimirSlavik: I think I eliminated `have_word_match`. That leaves two others, `upper_ascii` and `lower_ascii`. A thorough review of their usage (full info below) leaves only `SimpleConfigFile`, which uses this call for dict keys, so I think it can have a private method same as `ProxyString`. Then we could probably eliminate `ensure_str` outright. That would also mean I can drop the commit adding None support to `decode_bytes`, there would be nowhere left to require that.

However, I am not sure if that is all helpful, actually. This all looks to me as an exercise in reshuffling things, while none of the actual code benefits in a tangible way. The benefit to humans is one less function to read, in a file of 175 lines. That's laughable, IMO.

---

So, it's like this for `upper_ascii`:
- `QueueFactory` in `queuefactory.py` is ok, ultimately just str.
- `SimpleConfigFile` in `simpleconfig.py` reads from file and needs conversion from default.
That's somewhat fine.

For `lower_ascii` it's more fun:
- `_normalize_checksum` in `installation.py` - str from pykickstart
- `_complete_device_factory_request` in `add_device.py` - str via D-Bus
- `QueueFactory` in `queuefactory.py` - str, same as for `upper_ascii` above
- `_set_mount_point` in `custom_storage_helpers.py` - str from Gtk

VladimirSlavik:
It seems it's actually `decode_bytes` that is redundant. Saves us all from obsessing about the right encoding, I guess.

VladimirSlavik:
/kickstart-test --skip-testtypes knownfailure coverage

Not sure if this will work, but let's see. Also, apparently all the proxy tests are knownfailure or broken, so the crucial bit is left out.

VladimirSlavik:
/kickstart-test --skip-testtypes knownfailure --testtype coverage

VladimirSlavik:
So far, all the unit tests succeeded on each commit, as well as kickstart tests (smoke + coverage - knownfailure) on the last one.

I removed use of the "don't propagate bytes" functions from most places that required them. Kickstart, DNF, Gtk, RPM, that's all only strings now. It seems this was mostly a remnant of the python2 times.

I will review python3 handling of files and see if there is perhaps a way to make the `SimpleConfigFile` do the right thing on its own. Then we could remove both functions...

VladimirSlavik:
So, `SimpleConfigParser` doesn't need this either. All removed, hooray.

Of course, `SimpleConfigParser` could use some deletionist attention too - we use it to set one line in one file, whereas it is a grandiose design with atomic operations, and its helper string functions are used elsewhere.

VladimirSlavik:
The code works and is in the target state, but the "journey" needs some work. I will reorder this so that the removals are upfront, and then a lot of the early exploration can be skipped.

pyanaconda/core/string.py
VladimirSlavik: See below, it is all deleted.

VladimirSlavik:

Breakdown of the affected code areas that need testing beyond unit and kickstart tests:
- proxy - tested manually
- SELinux config - tested manually
- RPM kernel stuff - always runs (for initramfs)
- Xkl and gettext - always runs (spoke initialize)

VladimirSlavik:
The failed coverage kstest actually succeeded except for `network-static-2-httpks` and I think even that actually ran well. The last output is when `man-db` fails to run update in postscriptlet, which doesn't seem to be related.

<details>
  <summary>Relevant log section</summary>  

  ```
  
11:26:32,988 INFO anaconda:anaconda: modules.payloads.payload.dnf.transaction_progress: Configuring (running scriptlet for): rootfiles-8.1-30.fc35.noarch 1627041603 6d7c3cf9966c2337eda16f4b099ab33a144dd81abbf3ea6d5a101effd366f619
11:26:33,430 INFO systemd:Started /usr/bin/systemctl start man-db-cache-update.
11:26:33,448 NOTICE audit:SERVICE_START pid=1 uid=0 auid=4294967295 ses=4294967295 subj=system_u:system_r:kernel_t:s0 msg='unit=run-r74520347a8954ca29118df3ea312bf06 comm="systemd" exe="/usr/lib/systemd/systemd" hostname=? addr=? terminal=? res=success'
11:26:33,568 INFO systemctl:Failed to start man-db-cache-update.service: Unit man-db-cache-update.service not found.
11:26:33,591 NOTICE systemd:run-r74520347a8954ca29118df3ea312bf06.service: Main process exited, code=exited, status=5/NOTINSTALLED
11:26:33,594 WARNING systemd:run-r74520347a8954ca29118df3ea312bf06.service: Failed with result 'exit-code'.
11:26:33,603 NOTICE audit:SERVICE_STOP pid=1 uid=0 auid=4294967295 ses=4294967295 subj=system_u:system_r:kernel_t:s0 msg='unit=run-r74520347a8954ca29118df3ea312bf06 comm="systemd" exe="/usr/lib/systemd/systemd" hostname=? addr=? terminal=? res=failed'
  ```
  
</details>

pyanaconda/core/string.py
```diff
@@ -0,0 +1,176 @@
...
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+import string  # pylint: disable=deprecated-module
```
poncovka, line 19: Could you remove `string` from `deprecated-modules` in our `pylintrc` file, please? It seems to be a Python 2 thing.

pyanaconda/core/string.py
```diff
@@ -101,28 +109,30 @@ def lowerASCII(s):
 
     The returned string will contain only ASCII characters. This function is
     locale-independent.
-    """
 
-    # XXX: Python 3 has str.maketrans() and bytes.maketrans() so we should
-    # ideally use one or the other depending on the type of 's'. But it turns
-    # out we expect this function to always return string even if given bytes.
+    :param s: Input string
```
poncovka, line 88: Please, use `:param str s: input string`. Same above.

pyanaconda/core/string.py
```diff
@@ -135,7 +145,15 @@ def upcase_first_letter(text):
...
 
 def have_word_match(str1, str2):
-    """Tells if all words from str1 exist in str2 or not."""
+    """Tells if all words from str1 exist in str2 or not.
+
+    :param str1: list of words to look for
+    :type str1: str
+    :param str2: list of words to search in
+    :type str2: str
+    :return:
```
poncovka, line 128: The return value is not specified.

poncovka (approved):
I have some smaller suggestions. Otherwise, it looks great. Thank you!

poncovka (approved):
Thank you!

VladimirSlavik:
I did not find any use of these functions in related projects - initial setup, kdump, oscap.

PR #3621, opened 2021-09-23
Fix deprecation warning about Gdk.Cursor.new

VladimirSlavik:
Uses one of the other non-deprecated constructors instead.

### DIFF
--- a/pyanaconda/ui/gui/utils.py
+++ b/pyanaconda/ui/gui/utils.py
@@ -272,7 +272,7 @@ def busyCursor():
     if not window:
         return
 
-    window.set_cursor(Gdk.Cursor(Gdk.CursorType.WATCH))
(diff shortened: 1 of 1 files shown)


jkonecny12 (approved):
Looks good to me, even thought, I don't quite get why we need to add the argument here like that.

poncovka (approved):
Thank you!

VladimirSlavik:
> Looks good to me, even thought, I don't quite get why we need to add the argument here like that.

The underlying issue is that Gtk (Gdk?) had a number of constructors for the Cursor class, and deprecated the most straightforward one - `new` - which took just one parameter. I guess it just expected defaults for the rest. And the Python wrapper used that one for `__init__`, so instantiating the class is no longer possible "normally".

 The call I used is the one I judged the least ambiguous alternative.

PR #3624, opened 2021-09-23
Make organization fetching errors non-fatal

M4rtinK:
Looks like we can't reliably detect relevant errors encountered
while fetching organization list - while with Satellite we do get
the concrete "UnauthorizedException" with Hosted Candlepin we get
"RestlibException", which could mean many things.

So instead of trying to detect if the error is relevant for fetching
the organization list (eq. wrong username and/or password) let's just
log the raw error message we get (effectively a JSON in string form)
and return an empty organization list.

Still this should not change user visible behavior, as the next task
in line - the username + password registration task, will almost
certainly encounter the same error and will properly report it to the
user.

Resolves: rhbz#2006375

### DIFF
--- a/pyanaconda/modules/subscription/runtime.py
+++ b/pyanaconda/modules/subscription/runtime.py
@@ -1076,24 +1076,18 @@ class RetrieveOrganizationsTask(Task):
                 # return the DBus structure list
                 return org_data
             except DBusError as e:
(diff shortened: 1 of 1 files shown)


poncovka (approved):

VladimirSlavik (approved):
Looks good to me. Thank you!

M4rtinK:
This can be ported to Fedora only once Satellite support has been ported.

PR #3625, opened 2021-09-24
Use curl-minimal instead of curl in dockerfiles (#infra)

M4rtinK:
Looks like there are some issues with the curl package on RHEL 9
at the moment, so try using curl-minimal instead. Also as we
don't  use advanced curl functionality we might as well keep
using just curl-minimal even after the current breakage is fixed.

### DIFF
--- a/dockerfile/anaconda-ci/Dockerfile
+++ b/dockerfile/anaconda-ci/Dockerfile
@@ -14,7 +14,7 @@ COPY *.repo /etc/yum.repos.d/
 RUN set -e; \
   dnf update -y; \
   dnf install -y \
-  curl \
+  curl-minimal \
   /usr/bin/xargs \
   rpm-build \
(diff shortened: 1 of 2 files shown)


poncovka (approved):
Looks good to me. Thanks! We might want to port this to upstream if possible.

M4rtinK:
> Looks good to me. Thanks! We might want to port this to upstream if possible.

Wanted to ask about this actually, whats the process for this if curl is not outright broken in Fedora as well. I'll add the relevant label then.

jkonecny12 (approved):
Looks a great improvement for me. Thanks!

jkonecny12:
I agree with the change and I would even port it upstream. This is a benefit on container size and everything so why not to port it.

M4rtinK:
Port to Fedora: #3628

PR #3627, opened 2021-09-29
In installer environment set static, not transient hostname

rvykydal:
It provides more stable / predictable environment for installer %post
scritpts. For example prevents hostname changes by NetworkManager upon
device (re-)activation.

Resovles: rhbz#1975349


poncovka (approved):
Looks good to me, but I would hope there is also a unit test that should be modified.

VladimirSlavik (changes requested):
LGTM, +1 on test, but most importantly you have a typo in the "Resolves" line.

(Or does it not matter?)

rvykydal:
My intention is to write a kickstart / functional test (similar to the reproducer from the BZ).
Unit test (similar as we have) somehow doesn't make sense to me - what should we test here - the name of the DBus method called? I'll think about it more.

VladimirSlavik (approved):
Thank you!

poncovka:
The idea is to have a full coverage of the DBus API. It doesn't really matter how silly the tests seem to be. Anyway, the unit test exists in upstream and will be updated there, so I no longer insist on implementing one.

rvykydal:
Fedora PR: https://github.com/rhinstaller/anaconda/pull/3633

PR #3609, opened 2021-09-17
Document workaround MGA G200e graphics card

M4rtinK:
Add a workaround for the MGA G200e graphics card to common bugs.

Related: rhbz#2000537

PR #3628, opened 2021-09-29
Use curl-minimal instead of curl in dockerfiles (#infra)

M4rtinK:
Looks like using the fill curl package makes it more likely
to hit package dependency issues so use curl-minimal - we don't
use advanced curl functionality anyway & it might even make the
container smaller.

(cherry picked from commit 47f41ba9a2993ceef42af64e27a37299bb8f1291)

### DIFF
--- a/dockerfile/anaconda-ci/Dockerfile
+++ b/dockerfile/anaconda-ci/Dockerfile
@@ -31,9 +31,11 @@ COPY ["eln.repo", "/etc/yum.repos.d"]
 RUN set -ex; \
   if grep -q VARIANT.*eln /etc/os-release; then sed -i 's/enabled=0/enabled=1/' /etc/yum.repos.d/eln.repo; fi; \
   dnf update -y; \
+  dnf install -y --allowerasing \
+  curl-minimal \
+  libcurl-minimal; \
   dnf install -y \
   'dnf-command(copr)' \
-  curl \
   /usr/bin/xargs \
   rpm-build \
   git \
--- a/dockerfile/anaconda-rpm/Dockerfile
+++ b/dockerfile/anaconda-rpm/Dockerfile
@@ -26,9 +26,11 @@ COPY ["eln.repo", "/etc/yum.repos.d"]
 RUN set -ex; \
   if grep -q VARIANT.*eln /etc/os-release; then sed -i 's/enabled=0/enabled=1/' /etc/yum.repos.d/eln.repo; fi; \
   dnf update -y; \
+  dnf install -y --allowerasing \
+  curl-minimal \
+  libcurl-minimal;\
   dnf install -y \
   'dnf-command(copr)' \
-  curl \
   python3-pip \
   /usr/bin/xargs \
   rpm-build; \


M4rtinK:
Original RHEL 9 PR: #3609 3625

VladimirSlavik:
It's conflicting with curl pulled in by something else :(

M4rtinK:
> It's conflicting with curl pulled in by something else :(

Like, that's almost funny - RHEL 9 fails due to package conflicts if you use `curl` and using `curl-minimal` fixes the issue. 

But using `curl-minimal` on Fedora _cuases_ a package conflict! Perfect! :D

BTW, any idea why `curl-minimal` conflicts with `curl` ? That does not make much sense IMHO - you can install `vim-minimal` or `gdb-minimal` while having full `vim` and `gdb` packages installed just fine.

VladimirSlavik:
It seems that the UBI9 image contains `curl-minimal` out of the box, while Rawhide has `curl`. It makes me wonder if Rawhide will eventually move to the minimal variant too.

Do we need to care about the curl packages at all? I suppose it's a good idea to make sure we have what we need. Then again, things would fail very loudly if we did not have them, so I'm fine with removing these curls instead.

No idea about the conflict.

M4rtinK:
> It seems that the UBI9 image contains `curl-minimal` out of the box, while Rawhide has `curl`. It makes me wonder if Rawhide will eventually move to the minimal variant too.
> 
> Do we need to care about the curl packages at all? I suppose it's a good idea to make sure we have what we need. Then again, things would fail very loudly if we did not have them, so I'm fine with removing these curls instead.
> 
> No idea about the conflict.

I've reached out to the curl maintainer on Fedora and apparently the issue is caused by both `curl` and `curl-minimal` providing `/usr/bin/curl` but with different content.

Apparently the correct way to request curl-minimal is to actually use `--allowerasing` & do that in a separate `dnf install` invocation to avoid surprises (eq. to avoid silently erasing a non-curl related package). The curl maintainer also suggested to also install libcurl-minimal to possibly reduce the deps even further.

So I did the changes as suggested and it seems to work, yay! :)

VladimirSlavik (commented):
In principle LGTM, but I believe this actually does not help anything. IIRC:

What do we want to achieve by replacing the package? Use less resources? Or just compatibility?

A "container" is an instance from an "image" which consists of multiple immutable "layers". The "existing" `curl` is downloaded with the base image for Rawhide, so the cost is already paid. No point replacing it, that only makes the layer doing the replacement *larger* by including files of the other package.

The way to keep layers small is to chain multiple commands in RUN statements, including cleanups, so that the resulting state is more trimmed and atomic.

(I also don't understand the consequences of `--allowerasing` well enough, but that's hardly as fundamental.)

poncovka:
I agree with @VladimirSlavik. Let's wait for Rawhide to change that first.

M4rtinK:
The main aim for this was to keep things in sync with RHEL 9 as suggested by @jkonecny12 & get more acquainted with out infrastructure. We can definitely wait for Rawhide to change things or even skip this PR, no issues with that. :)

PR #3630, opened 2021-09-29
Payload should wait for all storage related threads to finish

M4rtinK:
Otherwise a race condition can happen when some of the storage related
threads run at the same time as the payload thread and both interact
with the Blivet storage instance.

This storage instance is not thread safe and two threads interacting
with it at the same time can result in a deadlock.

By waiting for all storage related threads to finish before running
the payload thread we can avoid this race condition and the resulting
deadlock.

Thanks a lot to Honza Stodola and Vendula Poncova for helping me track
down this kinda tricky race condition! :)

Resolves: rhbz#2002203

(cherry picked from commit d69bf45e05ac892d9c3115ff9867372ad043cb7e)


M4rtinK:
Port from RHEL 8: #3629

VladimirSlavik (approved):

rvykydal (approved):

poncovka (approved):
Thanks!

PR #3631, opened 2021-09-30
anaconda-cleanup: Don't unmount '/run/rootfsbase'.

FGrose:
For live installs, keep this base filesystem mounted.


FGrose:
Closing, there should only be one commit to pull.

VladimirSlavik:
Hello @FGrose. Sorry, you added the commit on top of the `master` branch, but set the merge target to `f35-devel`. That won't work, the base and target must be the same in 99% cases. As you can see this would "bridge" the two branches and bring all the other differences.

I assume you followed the instructions in our docs - if you have any insight about what's missing there, I'd be happy to improve these.

PR #3632, opened 2021-09-30
anaconda-cleanup: Don't unmount '/run/rootfsbase'.

FGrose:
For live installs, keep this base filesystem mounted.


FGrose:
2abfed0 (CONTRIBUTING) added in response to my mistaken PR #3631 based in the master branch and @VladimirSlavik request for insight.

VladimirSlavik:
Thank you for the changes!

First, I am actually not sure how to test the changes correctly. I tried this with Fedora Workstation Live, and it worked = didn't break anything. But I didn't see that mountpoint ever being used, too, so obviously this is for a different edition or spin or something. Can you please advise on the right environment to try this?

Secondly, thank you again for the docs changes! I originally meant that just as a question, but this is better. Technically it would be correct to change the docs in a separate PR, these are really two independent changes. Also the docs PR could be just for master, because the docs are usually seen on github which defaults to master. Personally I think that's just more work for little gain, so I'm fine taking this as 2 in 1. (`fNN-devel` gets merged into `master`). Others might request this in their reviews though.

FGrose:
> Can you please advise on the right environment to try this?
This mount is needed for the [flattened-squashfs-overlayfs source image case](https://github.com/rhinstaller/anaconda/blob/f0df778cc43e2494fdd0fad131aa8643590510ac/pyanaconda/modules/payloads/source/live_os/live_os.py#L124).  See also in [anaconda-lib.sh](https://github.com/rhinstaller/anaconda/blob/e7a6035482d40afbf8997221dbd940de4ca09afa/dracut/anaconda-lib.sh#L150) and [this commit to liveinst](https://github.com/rhinstaller/anaconda/commit/62e9f94937d75af2ddd352a159c74fffc68a1ddc).

An .iso image containing a flattened root filesystem in the squashfs.img can be built with livemedia-creator using the --squashfs-only option, see [this commit](https://github.com/weldr/lorax/commit/cb91fa3c78a8003f092fd034dd3c5c0f230a1a850).

>Docs commit:  I realized too late that the separate commit to my F35-devel branch clone would cause it to join this pull request.

VladimirSlavik:
@FGrose, did you succeed in making a working image that fits these requirements? I mean, working = ran and installed. It's a new topic for me, but I'm getting a feeling that either I overlook something, or that kind of image is not supported.

More specifically, I get a message *No live image found!* It seems a code path for squashfs-only is missing? https://github.com/rhinstaller/anaconda/blob/f0df778cc43e2494fdd0fad131aa8643590510ac/pyanaconda/modules/payloads/source/live_os/live_os.py#L112

Or is my image defective? Here are the mountpoints I have...

<details>
<summary>output of `mount`</summary>

ysfs on /sys type sysfs (rw,nosuid,nodev,noexec,relatime,seclabel)
devtmpfs on /dev type devtmpfs (rw,nosuid,seclabel,size=958020k,nr_inodes=239505,mode=755,inode64)
securityfs on /sys/kernel/security type securityfs (rw,nosuid,nodev,noexec,relatime)
tmpfs on /dev/shm type tmpfs (rw,nosuid,nodev,seclabel,inode64)
devpts on /dev/pts type devpts (rw,nosuid,noexec,relatime,seclabel,gid=5,mode=620,ptmxmode=000)
tmpfs on /run type tmpfs (rw,nosuid,nodev,seclabel,size=397220k,nr_inodes=819200,mode=755,inode64)
cgroup2 on /sys/fs/cgroup type cgroup2 (rw,nosuid,nodev,noexec,relatime,seclabel,nsdelegate,memory_recursiveprot)
pstore on /sys/fs/pstore type pstore (rw,nosuid,nodev,noexec,relatime,seclabel)
none on /sys/fs/bpf type bpf (rw,nosuid,nodev,noexec,relatime,mode=700)
/dev/sr0 on /run/initramfs/live type iso9660 (ro,relatime,nojoliet,check=s,map=n,blocksize=2048,iocharset=utf8)
LiveOS_rootfs on / type overlay (rw,relatime,seclabel,lowerdir=/run/rootfsbase,upperdir=/run/overlayfs,workdir=/run/ovlwork)
rpc_pipefs on /var/lib/nfs/rpc_pipefs type rpc_pipefs (rw,relatime)
selinuxfs on /sys/fs/selinux type selinuxfs (rw,nosuid,noexec,relatime)
systemd-1 on /proc/sys/fs/binfmt_misc type autofs (rw,relatime,fd=30,pgrp=1,timeout=0,minproto=5,maxproto=5,direct,pipe_ino=17152)
tracefs on /sys/kernel/tracing type tracefs (rw,nosuid,nodev,noexec,relatime,seclabel)
mqueue on /dev/mqueue type mqueue (rw,nosuid,nodev,noexec,relatime,seclabel)
hugetlbfs on /dev/hugepages type hugetlbfs (rw,relatime,seclabel,pagesize=2M)
debugfs on /sys/kernel/debug type debugfs (rw,nosuid,nodev,noexec,relatime,seclabel)
fusectl on /sys/fs/fuse/connections type fusectl (rw,nosuid,nodev,noexec,relatime)
configfs on /sys/kernel/config type configfs (rw,nosuid,nodev,noexec,relatime)
tmpfs on /tmp type tmpfs (rw,nosuid,nodev,seclabel,size=993048k,nr_inodes=409600,inode64)
vartmp on /var/tmp type tmpfs (rw,relatime,seclabel,inode64)
tmpfs on /run/user/1000 type tmpfs (rw,nosuid,nodev,relatime,seclabel,size=198608k,nr_inodes=49652,mode=700,uid=1000,gid=1000,inode64)
gvfsd-fuse on /run/user/1000/gvfs type fuse.gvfsd-fuse (rw,nosuid,nodev,relatime,user_id=1000,group_id=1000)
</details>

Note that running the command in the code, `findmnt -n -o SOURCE <somewhere>` gives the expected output if I plug in `/run/initramfs/live` instead of `/run/rootfsbase`. Maybe I'm hitting a case which we don't support?

FGrose:
@VladimirSlavik : I made a working image using the [editliveos script in livecd-tools](https://github.com/livecd-tools/livecd-tools/blob/main/tools/editliveos), not with livemedia-creator.
The Zenity dialog:
```
The following error occurred while setting up the payload. This is a fatal error and installation will be aborted.

No live image found!
```
is exactly the error I receive before fixing `anaconda-cleanup` 

Your mount output does show a LiveOS_rootfs source on overlay for `/`, meaning the `--squashfs-only` option worked.
```
LiveOS_rootfs on / type overlay (rw,relatime,seclabel,lowerdir=/run/rootfsbase,upperdir=/run/overlayfs,workdir=/run/ovlwork)
```
so the image looks correctly formed, except that there is no mountpoint at `/run/rootfsbase`.

Did you apply the patch to not `umount` the loop device behind that mountpoint in `anaconda-cleanup` for `liveinst`?
or remount `/dev/loop0` on `/run/rootfsbase` after failing `liveinst` and then editing `/usr/bin/anaconda-cleanup`?
(The `umount` actually only hides the mountpoint, since it is already bound to the overlay device.)

jkonecny12:
If this is specific for LiveCD Tools it should be written somewhere. Could you please make a comment or update the commit message?

VladimirSlavik:
@FGrose I concur that @jkonecny12 was in a way wrong (it's not specific to that tool). Still, that mistake + my struggle to understand it both show that we won't know what this is about when we see it next time. So it would be good to amend the commit message and mention that the mount point is for the flattened squashfs+overlayfs format... please?

Alternatively, I have the same change prepared, so I can just force-push if you want less work.

I think we'll be good to merge once the commit message accounts for our emmentaler memory :)

(We missed the time frame for f35 GA, but this helps custom builds which don't depend on that, so I'm not asking to retarget this to master.)

PR #3635, opened 2021-10-01
Watch the org.freedesktop.hostname1 name

poncovka:
The hostname service deactivates itself after a while, so don't activate it
when the Network module starts and watch the DBus name instead. Once the name is
available, we can connect to the PropertiesChanged signal of the DBus service.


pyanaconda/modules/network/network.py
```diff
@@ -225,15 +256,21 @@ def _hostname_service_properties_changed(self, interface, changed, invalid):
             log.debug("Current hostname changed to %s", hostname)
 
     def get_current_hostname(self):
-        """Return current hostname of the system."""
+        """Return current hostname of the system.
+
+        It will activate the hostname service if it is deactivated.
```
VladimirSlavik, line 261: It might be useful to hint at the mechanism that does the activation, I am not able to tell how this happens...

(Same for the 3 other docstrings that mention it.)

VladimirSlavik (approved):
Looks good to me. Thank you!

rvykydal (approved):
Looks good to me, thank you.

M4rtinK (approved):
Looks good to me as well, but agreed with @VladimirSlavik that some docstrings for the the activation magic could be added. :)

PR #3636, opened 2021-10-04
Fix Satellite provisioning rollback after unregistration

M4rtinK:
After unregistering the machine from a Satellite instance, also
clear the cached provisioning script.

Otherwise the target system Satellite provisioning installation
task will run as it is triggered by the provisioning script
being available.

Resolves: rhbz#2006718


poncovka (approved):
Looks good to me.

VladimirSlavik (approved):
Looks good to me. Thank you!

PR #3638, opened 2021-10-04
ovirt: update help directory

sandrobonazzola:
on CentOS Stream 9 the help directory is located at
/usr/share/anaconda/help/en-US

Signed-off-by: Sandro Bonazzola <sbonazzo@redhat.com>

### DIFF
--- a/data/product.d/ovirt.conf
+++ b/data/product.d/ovirt.conf
@@ -26,7 +26,7 @@ req_partition_sizes =
     /boot  1  GiB
 
 [User Interface]
-help_directory = /usr/share/anaconda/help/rhel
+help_directory = /usr/share/anaconda/help/en-US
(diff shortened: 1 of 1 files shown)


VladimirSlavik (approved):
Looks good to me. Thank you!

poncovka (changes requested):
We choose the `en-US` directory automatically. It shouldn't be part of the path. I think you are trying to use the Fedora help content. Please, be aware that the built-in help was temporarily disabled in RHEL 9, because we are waiting for a new help content. See: [rhbz#1949073](https://bugzilla.redhat.com/show_bug.cgi?id=1949073), [rhbz#1995945](https://bugzilla.redhat.com/show_bug.cgi?id=1995945)

PR #3642, opened 2021-10-07
Restore contexts in a task instead of a %post script

VladimirSlavik:
This also moves our last `%post` script to python with test coverage.

This also breaks the kickstart test `selinux-contexts`.


VladimirSlavik:

I added one more commit to handle the directory.

PR #3645, opened 2021-10-11
Mention manual journal dumps for mising logs

VladimirSlavik:
Apparently this can be missing...


VladimirSlavik:
That's right, I need to reword that again. But the point is that on Live there is neither syslog nor journal.log so if Anaconda croaks and user is grabbing the files, the best part will be missing.

PR #3646, opened 2021-10-11
Don't assume Python modules are in sysconfig.get_path('purelib')

hroncok:
sysconfig.get_path('purelib') returns a location where pure Python modules
should be installed, which is not necessarily where they are loaded from.

Since Fedora 36, sysconfig uses a different install scheme for RPM packages
and for other packages (e.g. installed by pip, setuptools, distutils...).
Without additional context, the scheme for "other" is the default.
Hence, sysconfig.get_path('purelib') returns /usr/local/lib/python3.10/site-packages.
But dracut needs the files installed from RPM packages.

See the announcement about the change in Fedora's Python:
https://lists.fedoraproject.org/archives/list/python-devel@lists.fedoraproject.org/thread/AAGUFQZ4RZDU7KUN4HA43KQJCMSFR3GW/

Instead of directly asking for RPM locations,
we won't assume "where to install stuff" and "where is stuff imported from"
is the same question and we inspect sys.path instead.
To be able to reference specific Python modules,
we use importlib.util.find_spec() (available since Python 3.4).

Resolves https://bugzilla.redhat.com/show_bug.cgi?id=2012513

This change also has 2 added benefits:

 - It loads information from all Python import paths,
   not just 2, so when we add modules from /usr/lib64/python3.10/site-packages,
   this script won't need changes.

 - It does not assume where hardcoded modules are installed,
   so when requests turn into a single-file library requests.py or to an
   extension module (both unlikely, but technically possible),
   or sysconfig becomes a package with __init__.py,
   this script won't need changes.

### DIFF
--- a/dracut/python-deps
+++ b/dracut/python-deps
@@ -23,24 +23,23 @@ import copy
 import os
 import sys
 import sysconfig
+from importlib.util import find_spec
 from modulefinder import ModuleFinder
-
-sitedir = sysconfig.get_path('purelib')
-libdir = sysconfig.get_config_var('LIBDEST')
 
 # stringprep is needed by the idna encoding, and idna is needed by requests.
 # The encoding import is implicit so ModuleFinder doesn't find it right.
-alsoNeeded = {sitedir+"/requests/__init__.py": libdir+"/stringprep.py"}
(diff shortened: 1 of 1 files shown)


hroncok:
The packit failure is not related.

hroncok:
Thank you all for the super fast reviews!

PR #3650, opened 2021-10-13
Log statistics about the image installation

poncovka:
Run `rsync` with the `--stats` option to get some info about the installation.


poncovka:
rsync_stats

PR #3655, opened 2021-10-14
Unify GUI & TUI root spoke completeness conditions

VladimirSlavik:
It turns out that less is more. I also wonder how upstream missed the fix from RHEL-8.

Resolves: [rhbz#1999646](https://bugzilla.redhat.com/show_bug.cgi?id=1999646)

---

TUI showed Root spoke as completed, even if it was not. This happened because the default state is no root password = locked, and the spoke took locked root as completed too.

That behavior was not consistent with GUI, and has been fixed in RHEL 8, but for some reason the fix did not get upstreamed.

With this change:
- GUI and TUI root spokes use identical logic to determine if they are
  mandatory and completed.
- TUI shows both root and user spoke as uncompleted and mandatory on startup.
- TUI shows root spoke as uncompleted but not mandatory if admin iser exists.
- The RHEL 8 change is finally ported upstream.

### DIFF
--- a/pyanaconda/ui/tui/spokes/root_password.py
+++ b/pyanaconda/ui/tui/spokes/root_password.py
@@ -64,7 +64,7 @@ class PasswordSpoke(FirstbootSpokeMixIn, NormalTUISpoke):
 
     @property
     def completed(self):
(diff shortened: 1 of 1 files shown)


VladimirSlavik:

I added more background to the commit message and pasted it above too.

PR #3661, opened 2021-10-21
Avoid creating /foo during the test run (#infra)

jkonecny12:
The mkdirChain call wasn't patched in tests which resulted in creation of directory in '/' mount. We definitely don't want this to happen.

Another test scanning '/' for foo started to fail because of other sorting of tests during migration to pytest. Fixing this will make our test more clean, robust and solve blocking issue for pytest migration.


jkonecny12:
This test is not part of master so nothing to backport.

PR #3665, opened 2021-10-25
Show correctly that no admin user is set up

VladimirSlavik:
A quick fix for admin user detection in GUI. Previously, it was possible to satisfy the conditions just by setting root password, even if root was locked.

Code-wise, adding `and self._users_module.CheckAdminUserExists()` to `completed` of root and user spokes will make the respective spoke signal incomplete at the right time = displaying the exclamation mark and red text.

The one question is how to signal that the user has a choice of resolving this in *both* spokes. Easiest fix would be doing both, which is the initial submission here.

Resolves: [rhbz#2015508](https://bugzilla.redhat.com/show_bug.cgi?id=2015508)

### DIFF
--- a/pyanaconda/ui/gui/spokes/root_password.py
+++ b/pyanaconda/ui/gui/spokes/root_password.py
@@ -186,7 +186,10 @@ class PasswordSpoke(FirstbootSpokeMixIn, NormalSpoke, GUISpokeInputCheckHandler)
 
     @property
     def mandatory(self):
-        """Only mandatory if no admin user has been requested."""
+        """Only mandatory if no admin user has been requested.
+
+        See also doc for the property completed().
+        """
         return not self._users_module.CheckAdminUserExists()
 
     def apply(self):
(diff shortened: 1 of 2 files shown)


pyanaconda/ui/gui/spokes/root_password.py
```diff
@@ -216,7 +219,14 @@ def apply(self):
...
     def completed(self):
-        return self._users_module.IsRootPasswordSet
+        """Is the spoke completed?
+
+        For root and user, the mandatory+completed pair is a complicated hack. Having an usable
+        admin user is mandatory, but it is not clear if it should be an unlocked root, or a sudoer.
+        Thus, mandatory on both spokes checks admin user, and complete then checks again: Both the
+        spoke-specific completion condition, as well as existence of an admin.
+        """
+        return self._users_module.IsRootPasswordSet and self._users_module.CheckAdminUserExists()
```
poncovka, line 229: I think that the real issue is that the root password is still set to the previous value after the user disables the root account.

VladimirSlavik: Yes, that would solve the problem. But that is arguably a feature. I think setting password for locked users is a thing - correct me if I'm wrong.

poncovka: I wouldn't expect that behaviour based on the current UI. Locked root with a password is supported with kickstart, but I thought we don't want to support that in the UI. What do others think? @jkonecny12, @M4rtinK

jkonecny12: I agree that it's not something we want to support in GUI. In general it could be confusing for people.

VladimirSlavik: IIRC the previous GUI "supported" it, so it will be a change. But I very much like the idea, it simplifies things.

M4rtinK: Agreed with not supporting setting a root password in GUI and then locking it - that really seems pretty esoteric to me & non-issue for advanced users if supported in kickstart.

M4rtinK: Not to mention an admin level user account or service running with root privileges will be needed to unlock the root password. That thing can also set the password to a value if needed.

poncovka: Fixed at https://github.com/rhinstaller/anaconda/pull/3775.

PR #3667, opened 2021-10-27
make US keyboard layout easier to find in Anaconda

sdp5:
**Use Case**: 
As a US keyboard layout user, I would like to be able to find the US keyboard layout in my own language quickly and easily in the Anaconda installer.

a discussion around the use case is [here](https://lists.fedoraproject.org/archives/list/devel@lists.fedoraproject.org/thread/WFVVUOKIGHARF5Z2FQXRORM4RXFIMBOJ/).

### DIFF
--- a/pyanaconda/ui/gui/spokes/keyboard.py
+++ b/pyanaconda/ui/gui/spokes/keyboard.py
@@ -134,8 +134,9 @@ class AddLayoutDialog(GUIObject):
     def _initialize(self):
         common_layouts = self._xkl_wrapper.get_common_layouts()
         available_layouts = self._xkl_wrapper.get_available_layouts()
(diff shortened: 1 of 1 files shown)


sdp5:
@poncovka thanks for the review. 
We already have a separator after common layouts. 
I'm fine with having an another one after the default_layout.

Feel free to add your changes, or should I go ahead add those?

VladimirSlavik (approved):
Looks good to me. Thank you!

+1 on the separator. I think Vendy meant you would add it to this change.

poncovka:
I can do that in a follow-up pull request. Thank you!

jkonecny12:
Great improvement and even easy to implement. Thanks!

PR #3669, opened 2021-10-28
bug fix: activate connection from callback _connection_updated_cb

qyng:
Bug Fix: activate connection from callback _connection_updated_cb

this is the correct order
   ui.tui.spokes.network: updating connection ...
   ui.tui.spokes.network: updated connection ...
   ui.tui.spokes.network: activating connection ...

otherwise it could be wrong as
   ui.tui.spokes.network: updating connection ...
   ui.tui.spokes.network: activating connection ...
   ui.tui.spokes.network: updated connection ...

### DIFF
--- a/pyanaconda/ui/tui/spokes/network.py
+++ b/pyanaconda/ui/tui/spokes/network.py
@@ -404,10 +404,6 @@ class NetworkSpoke(FirstbootSpokeMixIn, NormalTUISpoke):
 
         if new_spoke.apply_configuration:
             self._apply = True
-            device = self.nm_client.get_device_by_iface(iface)
-            log.debug("activating connection %s with device %s",
-                      connection_uuid, iface)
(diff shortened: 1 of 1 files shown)


rvykydal (approved):
Looks good to me, thank you for the fix!

I wonder if we should pass nm_client instance to the ConfigureDeviceSpoke rather then getting one in the callback but I don't have strong opinion on that.

qyng:
@rvykydal, Passing the nm_client instance was also my first thinking in the beginning. But later I decided to get it on the fly to keep this fix small and interfaces clean, as the nm_client is already being well cached for us from the network module. We may overhaul this part when we have more things to pass between the two spokes.

rvykydal:
@qyng please remove the added merge commits from the branch.
I'll merge your PR then.

PR #3670, opened 2021-10-29
Use the set-up and tear-down tasks of the Live OS image source

poncovka:
* Clean up the tasks and use them in the Live OS payload class.
* Run the task for the Live OS image detection directly.


jkonecny12 (approved):
Looks good to me. Thanks!

PR #3672, opened 2021-11-01
Clean up shell code in tests

VladimirSlavik:
Fix shellcheck warnings + organize the scripts a bit.

### DIFF
--- a/.shellcheckrc
+++ b/.shellcheckrc
@@ -1,0 +1,4 @@
+# ShellCheck config file
+
+disable=SC1090  # Can't follow non-constant source. Use a directive to specify location.
+disable=SC2230  # which is non-standard. Use builtin 'command -v' instead.
--- a/tests/glade_tests.sh
+++ b/tests/glade_tests.sh
@@ -2,8 +2,8 @@
 
 # If $top_srcdir has not been set by automake, import the test environment
 if [ -z "$top_srcdir" ]; then
-    top_srcdir="$(realpath $(dirname "$0")/..)"
-    . ${top_srcdir}/tests/testenv.sh
+    top_srcdir="$(realpath "$(dirname "$0")/..")"
+    . "${top_srcdir}/tests/testenv.sh"
 fi
 
 # If no tests were selected by user or makefile, select all of them
@@ -11,4 +11,4 @@ if [ $# -eq 0 ]; then
     set -- "$top_srcdir"/tests/glade_tests
 fi
 
-exec python3 -m unittest discover -t $top_srcdir -v -b -s $@
+exec python3 -m unittest discover -t "$top_srcdir" -v -b -s "$@"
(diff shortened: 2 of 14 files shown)


poncovka (approved):
Thanks!

M4rtinK (approved):
Looks good to me, thanks!

.shellcheckrc
```diff
@@ -0,0 +1,4 @@
+# ShellCheck config file
```
jkonecny12, line 1: Please add this file to the `.structure-config` file.

tests/glade_tests.sh
```diff
@@ -2,13 +2,13 @@
 
 # If $top_srcdir has not been set by automake, import the test environment
 if [ -z "$top_srcdir" ]; then
-    top_srcdir="$(realpath $(dirname "$0")/..)"
-    . ${top_srcdir}/tests/testenv.sh
+    top_srcdir="$(realpath "$(dirname "$0")/..")"
```
jkonecny12: Are you sure that this works as expected. The parser will probably not work correctly on all these `"`. Also the original variant is wrong too. I guess the correct way to fix this should be to start escaping the inner quotes.

tests/rpm_tests.sh
```diff
@@ -2,13 +2,13 @@
 
 # If $top_srcdir has not been set by automake, import the test environment
 if [ -z "$top_srcdir" ]; then
-    top_srcdir="$(realpath $(dirname "$0")/..)"
-    . ${top_srcdir}/tests/testenv.sh
+    top_srcdir="$(realpath "$(dirname "$0")/..")"
```
jkonecny12: The same here.

tests/unit_tests.sh
```diff
@@ -2,13 +2,13 @@
 
 # If $top_srcdir has not been set by automake, import the test environment
 if [ -z "$top_srcdir" ]; then
-    top_srcdir="$(realpath $(dirname "$0")/..)"
-    . ${top_srcdir}/tests/testenv.sh
+    top_srcdir="$(realpath "$(dirname "$0")/..")"
```
jkonecny12: And here.

tests/glade_tests.sh
jkonecny12: Although the question is if the `$(..)` sections are not parsed separately.

jkonecny12 (changes requested):
Some things have to be considered, however, we definitely want that file added to the structure file list.

tests/glade_tests.sh
VladimirSlavik: Yes, it works. As you already guessed, $(...) is another "scope" - it creates a subshell where the quotes are parsed independently of the surroundings. That leads to some really bizarre things like this one here.

.shellcheckrc
VladimirSlavik: Ah, good catch, sorry!

VladimirSlavik:

Update: Added .shellcheckrc to .structure-config

tests/glade_tests.sh
jkonecny12: Thanks for verification of this.

jkonecny12 (approved):
Looks good to me now. Thanks!

PR #3675, opened 2021-11-02
Remove git hook scripts

VladimirSlavik:
They are outdated and unused.

Or so I think! Feel free to disagree, just explain why we should keep them.


scripts/githooks/pre-push-prevent-upstream-push
```diff
@@ -1,29 +0,0 @@
-#!/bin/sh
```
jkonecny12, line 1: I'm still using this one and I would honestly prefer that anyone would use this. It's a safeguard preventing you to change upstream repository. The only time when you want to push to the repository directly is when you are doing release.

And if you have the upstream named as `origin` then it's pretty easy to type `git push` instead of `git push fork`.

jkonecny12 (commented):
If no-one else using this, you are free to remove it. However, I would recommend everyone to use that one prevention mechanism.

scripts/githooks/pre-push-prevent-upstream-push
VladimirSlavik: Sounds good! It's pretty trivial to have the same security without that, though. Clone the `origin` from `https`, so you would have to provide credentials for push. Then add another remote `origin-push` for the same repo, but using `ssh`, so you can push seamlessly, but have to explicitly enter the remote.

jkonecny12: Yep, also possible, the unfortunate thing is that you have to have one more remote. However, yes, I agree.
Another solution is to pull the origin with a different remote name.

M4rtinK: I use a slightly different workflow - I never really manually use `git push` (unless I'm tagging an anaconda release), rather I always have a remote called `fork` pointing to my private fork of the given project. Then I have this shortcut i my `.gitconfig` file:

```
pf = !sh -c 'branch=$(git symbolic-ref HEAD | cut -d '/' -f 3) && git push fork' -
pff = !sh -c 'branch=$(git symbolic-ref HEAD | cut -d '/' -f 3) && git push --force fork' -
```
When working on a change I create a new branch:

`git checkout -b master-fix_stuff`

Fix things, commit them to the topic branch, then push them to fork:

`git pf`

If I need to update the remote topic branch, I run:

`git pff`

That's it - less typing, should prevent unwanted pushes to upstream and is defined via one global config for all git repos.

Worst case the git repo does not have a `fork` remote defined and the shortcut wont work, which is in itself a red flag usually indicating you have your git checkouts mixed up. :)

poncovka: I never push to upstream from repositories that I use for development. If I need to do that, I always create a temporary one in `/tmp`, so it will be deleted on next reboot.

It means that I can completely disable this functionality. I use the command below to set an invalid URL for pushing to upstream. It is just one command that works for any repository, not just Anaconda.

```
> git remote set-url --push upstream FORBIDDEN
```

Then my remotes look like this:

```
> git remote -v
origin	git@github.com:poncovka/anaconda.git (fetch)
origin	git@github.com:poncovka/anaconda.git (push)
upstream	git@github.com:rhinstaller/anaconda.git (fetch)
upstream	FORBIDDEN (push)
```

I tried to use the hook, but I don't really trust it. I would have to somehow verify that it is set up correctly and I don't want to spend time on that.

poncovka (approved):
I could imagine to have some of these checks in our GitHub workflows, but they don't make sense to me as hooks. Thanks!

VladimirSlavik:
Update: Put back the hook for Jirka.

M4rtinK (approved):
Looks good to me, thanks! :)

scripts/githooks/pre-push-prevent-upstream-push
jkonecny12: Interesting ideas. Thanks for sharing.

PR #3676, opened 2021-11-02
Migrate the Live OS payload on DBus

poncovka:
* Clean up the Live OS payload and source modules.
* Use the Payloads module to install the Live OS payload.
* Don't require implementation of `post_install_with_tasks`.

### DIFF
--- a/tests/unit_tests/pyanaconda_tests/modules/payloads/payload/test_module_payload_live_os.py
+++ b/tests/unit_tests/pyanaconda_tests/modules/payloads/payload/test_module_payload_live_os.py
@@ -17,103 +17,61 @@
 #
 # Red Hat Author(s): Jiri Konecny <jkonecny@redhat.com>
 #
+import os
+import tempfile
 import unittest
 
-from tests.unit_tests.pyanaconda_tests import check_task_creation, patch_dbus_publish_object, \
-    PropertiesChangedCallback
+from pyanaconda.core.constants import SOURCE_TYPE_LIVE_OS_IMAGE, PAYLOAD_TYPE_LIVE_OS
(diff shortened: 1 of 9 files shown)


tests/unit_tests/pyanaconda_tests/modules/payloads/payload/test_module_payload_live_os.py
```diff
@@ -17,103 +17,61 @@
...
         """Test LiveOS payload can't set multiple sources."""
-        paths = [
-            self._prepare_source(),
-            self._prepare_source()
-        ]
-
-        self.shared_tests.set_and_check_sources(paths, exception=IncompatibleSourceError)
-
-    @patch_dbus_publish_object
-    def test_set_when_initialized_source_fail(self, publisher):
```
jkonecny12, line 82: Could you please explain why you are removing these tests?

jkonecny12 (approved):
Looks good to me. I just have one question below.

tests/unit_tests/pyanaconda_tests/modules/payloads/payload/test_module_payload_live_os.py
poncovka: This is already tested in `test_set_when_initialized_source_fail` of the `PayloadBaseInterfaceTestCase` class. The tested functionality is implemented in the `PayloadBase` class and it doesn't change, so it is not necessary to test it with all types of payloads and sources.

poncovka:
There is a new commit with a fix for the `rsync` calls.

VladimirSlavik (approved):
Looks good to me. Thank you!

PR #3677, opened 2021-11-03
network: always use rd.iscsi.ibft when the need to access an iBFT device

lkundrak:
Previously, the installer would add rd.iscsi.ibft only if network during
install was configured from iBFT and not set any other network
configuration options to kernel command line.

That is not correct. The installation could've been started from a CD
media and still be done to an iSCSI volume discovered via iBFT.
Given presence of no other network configuration, the rd.iscsi.firmware
will cause dracut to still bring up networking automatically (as if
rd.needned has been passed).

That might work for most installations, but is probably not the right thing
to do: all interfaces will be brought up needlessly and things would break
in less straightforward setups, such as the iSCSI volume accessed via VLAN.

Let's just use rd.iscsi.ibft unconditionally if there's an iSCSI volume
configured via iBFT.


jkonecny12:
Seems reasonable to me. @rvykydal what do you think?

VladimirSlavik (approved):
Code-wise LGTM, can't comment on validity of what is done.

rvykydal:
Thank you for the PR, it looks good to me.
I have just one concern before approving - with your patch `_get_dracut_znet_argument_from_connection(connection)` (https://github.com/NetworkManager/anaconda/blob/8c2e8f84821c4c498b8d7f7215184841c7987c25/pyanaconda/modules/network/nm_client.py#L1127) will not be called whereas it used to be before (in case there were found an ibft connection).
Can it be a problem ?
Asking also @jstodola who is our QA for s390.

jstodola:
Good catch, Radek.
Not generating the `rd.znet=` boot argument could indeed be a problem on s390x - this needs to be tested first. I can try it later this week and report the result.

jstodola:
So, omitting the `rd.znet=` boot argument is a problem for systems with rootfs on iSCSI, the system won't boot without it.
In most cases, the `rd.znet=` boot argument is passed to the installed system by preserving the `rd.znet=` boot argument from the installation:
https://github.com/rhinstaller/anaconda/blob/master/data/anaconda.conf#L156
The `_get_dracut_znet_argument_from_connection(connection)` function is probably only necessary when `rd.znet=` was not appended on the kernel command line during the installation - I can think only about one case - when the system configuration is read from the CMS config file on z/VM guests.
The `_get_dracut_znet_argument_from_connection(connection)` function currently doesn't even work properly ([PR#3703](https://github.com/rhinstaller/anaconda/pull/3703)).

rvykydal:
With @jstodola's feedback I am OK with the change.
I am going to test it manually before merging.

rvykydal:
@lkundrak could you please resolve the conflict on the branch, it is caused by merging https://github.com/rhinstaller/anaconda/pull/3703 - just the code that is removed in this PR has changed a bit.
I am sorry for being so long with this PR, I just need to manually test the PR which I want to do shortly after the conflict resolution.

rvykydal:
Tested by booting from iscsi from ibft and from configuration in Anaconda.

rvykydal:
> @lkundrak could you please resolve the conflict on the branch, it is caused by merging #3703 - just the code that is removed in this PR has changed a bit. I am sorry for being so long with this PR, I just need to manually test the PR which I want to do shortly after the conflict resolution.

I've opened another PR with this patch rebased on top of master branch: https://github.com/rhinstaller/anaconda/pull/3750

PR #3682, opened 2021-11-04
network: apply kickstart network --nodefroute also from stage2

rvykydal:
Apply the option also when kickstart is applied in stage2, for example when
defining commands via %pre section.

Resolves: rhbz#1990145


poncovka (approved):
Looks good to me.

VladimirSlavik (approved):
Looks good to me. Thank you!

PR #3684, opened 2021-11-04
Satellite support - port to Fedora, batch nr. 1

M4rtinK:
This is a very small initial batch that basically contains all commits related but not directly dependent on Satellite related changes. Still, we won't have to care about those in the later batches if we handle them now. :)

### DIFF
--- a/tests/unit_tests/pyanaconda_tests/modules/subscription/test_system_purpose.py
+++ b/tests/unit_tests/pyanaconda_tests/modules/subscription/test_system_purpose.py
@@ -1,0 +1,193 @@
+#
+# Copyright (C) 2021  Red Hat, Inc.
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# the GNU General Public License v.2, or (at your option) any later version.
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY expressed or implied, including the implied warranties of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General
+# Public License for more details.  You should have received a copy of the
+# GNU General Public License along with this program; if not, write to the
+# Free Software Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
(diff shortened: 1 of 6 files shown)


M4rtinK:
Looking into the unit test & pylint issues. :)

M4rtinK:
Looks like this is at least partially caused by not backporting the syspurpose DBus PR first - so lets do that #3690 and then revisit this one. I'll rebase my local fork in the meantime so we can get this PR going right away once the syspurpose PR is merged.

jkonecny12:
Just write here a comment that it's prepared for review.

M4rtinK:
> Just write here a comment that it's prepared for review.

Sure - it should be ready now, rebased on top of the master branch with the syspurpose PR merged. 

Tests are now also passing as far as I can tell. :)

jkonecny12 (approved):
Looks good to me. Thanks!

tests/unit_tests/pyanaconda_tests/modules/subscription/test_system_purpose.py
```diff
@@ -0,0 +1,206 @@
...
+    """Test the system purpose data handling code."""
+
+    def test_system_purpose_valid_json_parsing(self):
+        """Test that the JSON file holding valid system purpose values is parsed correctly."""
+        # check file missing completely
+        # - use path in tempdir to a file that has not been created and thus does not exist
+        with tempfile.TemporaryDirectory() as tempdir:
+            no_file = os.path.join(tempdir, "foo.json")
+            roles, slas, usage_types = get_valid_fields(valid_fields_file_path=no_file)
+            self.assertListEqual(roles, [])
```
poncovka, line 60: Please, drop this commit and create a new one with the upstream code. All `self.assert` calls were replaced with the `assert` statement.

poncovka (changes requested):
You replace new code with the old one.

tests/unit_tests/pyanaconda_tests/modules/subscription/test_system_purpose.py
M4rtinK: Ooops, now I see it. :P Should be fixed now. :)

poncovka (approved):
Looks good to me. Thanks!

PR #3688, opened 2021-11-05
Disable container build for Fedora 35 (#infra)

jkonecny12:
Fedora 35 was already released so there is no reason to have a container for that build. Also we are seeing a pylint flake there.

I'm talking about this https://github.com/rhinstaller/anaconda/runs/4117118082?check_suite_focus=true .


VladimirSlavik (approved):
Looks good to me. Thank you!

PR #3689, opened 2021-11-08
Add changelog to GH releases

VladimirSlavik:
Let's make it human readable.

Not tested yet = blocked.

### DIFF
--- a/.github/workflows/tag-release.yml
+++ b/.github/workflows/tag-release.yml
@@ -71,9 +71,24 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           RELEASE_NAME: ${{ steps.get_version.outputs.version }}
         run: |
+          echo -e "Anaconda release $RELEASE_NAME.\n\n" > release.txt
+
+          # find version tags
+          CURRENT_VERSION=${{ github.ref }}
+          PREVIOUS_VERSION="$(git describe --tags --abbrev=0 --exclude="$CURRENT_VERSION")"
+
+          # print commit details; skip the last = release commit
+          git log \
+            --no-merges \
+            --pretty=format:"- %s (%aL)" \
+            "$PREVIOUS_VERSION..$CURRENT_VERSION~1" \
+              | grep -v "#infra" \
+              | tee -a release.txt
+
+          # create release with the release notes
           gh release create \
             ${{ github.ref }} \
             --draft \
             --title "anaconda-$RELEASE_NAME" \
-            --notes "Anaconda release $RELEASE_NAME created automatically from git tag." \
+            --notes-file release.txt \
             /tmp/results/*


M4rtinK (approved):
Nice idea, I like it! :)
Code changes look good to me as well. :)

poncovka (approved):
Looks good to me. Thanks!

.github/workflows/tag-release.yml
```diff
@@ -71,9 +71,13 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           RELEASE_NAME: ${{ steps.get_version.outputs.version }}
         run: |
+          echo "Anaconda release $RELEASE_NAME created automatically from git tag.\n\n" > release.txt
```
jkonecny12, line 74: We should remove this line. It's really not useful information to users that the release was created from the git tag automatically.

Just simple:

`Anaconda release $RELEASE_NAME.` should be enough.

.github/workflows/tag-release.yml
```diff
@@ -71,9 +71,13 @@ jobs:
...
+          # lines between %changelog and an empty line | remove the first two lines
+          sed -n '/^%changelog$/,/^$/p' anaconda.spec.in | sed '1,2d' >> release.txt
```
jkonecny12, line 76: Hmm, I wonder if this is a good approach. The point is that we want to remove `%changelog` from the spec file. It shouldn't be there and then this code will break.
I would rather like to see call of `makedump` with some option which would only give you the changelog part and nothing else.

jkonecny12 (changes requested):
Great improvement, however, I think we may want to take a bit other approach.

.github/workflows/tag-release.yml
VladimirSlavik: Yes, that makes sense.

VladimirSlavik:
Update: Do not read spec file. Instead, read the git log.

.github/workflows/tag-release.yml
VladimirSlavik: See the latest changes for an approach that uses git instead.

.github/workflows/tag-release.yml
```diff
@@ -71,9 +71,25 @@ jobs:
...
           RELEASE_NAME: ${{ steps.get_version.outputs.version }}
         run: |
+          echo "Anaconda release $RELEASE_NAME.\n\n" > release.txt
+          ## lines between %changelog and an empty line | remove the first two lines
+          #sed -n '/^%changelog$/,/^$/p' anaconda.spec.in | sed '1,2d' >> release.txt
+
+          # current state is with new version number
+          CURRENT_VERSION=$(grep -m 1 AC_INIT configure.ac | sed -E "s/AC_INIT\(\[(.*)\], \[(.*)\], \[(.*)\]\)/\2/")
+          # last commit is release = bumps version = previous commit had the previous version
+          git show HEAD~1:configure.ac > configure.ac.prev
```
jkonecny12, line 81: Are you sure that this will work also for commits like these 0c59f7f6a03070e2aa10a27a17034ec9303c2905.

jkonecny12: Instead of that please use `git describe --tags --exclude=$CURRENT_VERSION`.

VladimirSlavik: It is more neat, I could use the ref as current and this as previous, but will it work near merges?

Also, please explain what will not work? This cats the whole file, not a diff.

jkonecny12: > That's more neat, I could use the github ref as current and this as previous, but will it work near merges?

If I'm not mistaken `packit` is using the same approach so I think it should work. Also we don't merge releases so there shouldn't be tags in the merges.

> Also, please explain what will not work? This cats the whole file, not a diff.

Ohh, I miss-understood how your algorithm works. I thought it's looking for a last version commit -- my bad.
Nevertheless, I can thing of the situation where there was a mistake and the configure.ac will change the version in a different commit than the tag is. Shouldn't happen but still could happen. The proposed variant seems more robust to me.

jkonecny12:
Removing NACK. It's in shape that I don't want to block it but I still think it could be improved a bit.

.github/workflows/tag-release.yml
VladimirSlavik: The proposed command will work, with some strings attached. It finds the closest tag, so merging `fNN-devel` would not add any tags on the "other" branch behind one of the merges. That was my concern, and turns out to be fine, as long as we do not change the branching scheme. We also must not add any tags outside the releases. That seems also doable.

It seems less fragile than the manual parsing, so let's do it instead, and keep the parsing backed up in a spoiler...

<details>
  <summary>Manual parsing</summary>
  
  ```bash
        - name: Create the release
        id: create_release
        env:
          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
          RELEASE_NAME: ${{ steps.get_version.outputs.version }}
        run: |
          echo "Anaconda release $RELEASE_NAME.\n\n" > release.txt
          ## lines between %changelog and an empty line | remove the first two lines
          #sed -n '/^%changelog$/,/^$/p' anaconda.spec.in | sed '1,2d' >> release.txt

          # current state is with new version number
          CURRENT_VERSION=$(grep -m 1 AC_INIT configure.ac | sed -E "s/AC_INIT\(\[(.*)\], \[(.*)\], \[(.*)\]\)/\2/")
          # last commit is release = bumps version = previous commit had the previous version
          git show HEAD~1:configure.ac > configure.ac.prev
          PREVIOUS_VERSION=$(grep -m 1 AC_INIT configure.ac.prev | sed -E "s/AC_INIT\(\[(.*)\], \[(.*)\], \[(.*)\]\)/\2/")

          # print first log line and committer between previous and this versions' tags, except for the very last commit
          # drop #infra commits and make the lines "- first line (committer)" where everything after @ from commiter is dropped
          git log --no-merges --pretty=format:"%s %aE" "anaconda-$PREVIOUS_VERSION-1..anaconda-$CURRENT_VERSION-1~1" \
            | grep -v "#infra" | sed -E 's/(.+) ([a-zA-Z0-9_\-]+)@[a-zA-Z0-9\-\.]+$/-\1 (\2)/' >> release.txt

          # create release with the release notes
          gh release create \
            ${{ github.ref }} \
            --draft \
            --title "anaconda-$RELEASE_NAME" \
            --notes-file release.txt \
            /tmp/results/*
  ```
  
</details>

VladimirSlavik:
Update: Use packit commands + rebase.

VladimirSlavik:
Update: fix escaping echo, show extracted log so it's debuggable.

.github/workflows/tag-release.yml
VladimirSlavik: I think this is resolved.

jkonecny12 (approved):

PR #3695, opened 2021-11-09
Fix shell quotes in dracut

VladimirSlavik:
Shellcheck is more happy now: No SC2068 "Double quote to prevent globbing and word splitting."

This should not change anything, but we will see. Kickstart tests to run:
- [x] smoke
- [x] coverage
- [x] network (rvykydal says: maybe random choice, include bootopts, exclude bindtomac)
- [x] driverdisk-related ones

### DIFF
--- a/dracut/anaconda-copy-cmdline.sh
+++ b/dracut/anaconda-copy-cmdline.sh
@@ -2,6 +2,6 @@
 # Copy over cmdline(.d) files from the initrd to /run before pivot
 mkdir -p /run/install/cmdline.d
 for f in /etc/cmdline.d/*; do
-    [ -e $f ] && cp $f /run/install/cmdline.d/
+    [ -e "$f" ] && cp "$f" /run/install/cmdline.d/
 done
 [ -e /etc/cmdline ] && cp /etc/cmdline /run/install/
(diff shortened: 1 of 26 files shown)


VladimirSlavik:
/kickstart-test --testtype coverage --skip-testtypes knownfailure

VladimirSlavik:
/kickstart-test --testtype network --skip-testtypes knownfailure

VladimirSlavik:
At the volume of tests used here, flakes are to be expected, I think. The results will need more interpretation.

jkonecny12 (commented):
Looks good to me so far.

VladimirSlavik:
/kickstart-test driverdisk-disk ntp-pools lvm-1 reqpart default-desktop network-prefixdevname bond2-pre --skip-testtypes knownfailure

VladimirSlavik:
So: `driverdisk-disk` is still broken, so that had to fail, and `network-prefixdevname` which timed out, is to be skipped on fedora. In other words, all kickstart tests eventually succeeded.

I think that means we can't make it better, and can speculatively merge and let the other testing layers verify that this works.

poncovka (approved):
Looks good to me. Thanks!

rvykydal (approved):
Although I wonder why some vars are not required to be quoted - if the update makes the spellcheck happy, I am happy too.
LGTM

jkonecny12:
> Although I wonder why some vars are not required to be quoted - if the update makes the spellcheck happy, I am happy too. LGTM

I guess it has logic that quoting never hurts and could avoid issues in future modification of the code.

jkonecny12 (approved):

PR #3697, opened 2021-11-09
Unify the detection of the tar image

poncovka:
Use the `is_tar` function to detect a tar image based on the given URL.

### DIFF
--- a/pyanaconda/modules/payloads/source/utils.py
+++ b/pyanaconda/modules/payloads/source/utils.py
@@ -29,6 +29,19 @@ from pyanaconda.anaconda_loggers import get_module_logger
 log = get_module_logger(__name__)
 
 
+def is_tar(url):
+    """Is the given URL a path to the tarball?
+
+    :param url: a string with URL
+    :return: True or False
+    """
+    if not url:
+        return False
+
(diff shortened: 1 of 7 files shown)


VladimirSlavik (approved):
Looks good to me. Thank you!

pyanaconda/modules/payloads/source/utils.py
```diff
@@ -29,6 +29,19 @@
...
+def is_tar(url):
+    """Is the given URL a path to the tarball?
+
+    :param url: a string with URL
+    :return: True or False
+    """
+    if not url:
+        return False
+
+    tar_suffixes = (".tar", ".tbz", ".tgz", ".txz", ".tar.bz2", "tar.gz", "tar.xz")
```
jstodola, line 41: Dots are missing for the `tar.gz` and `tar.xz` suffixes? Otherwise http://my/tatar.gz would be considered a tarball.

VladimirSlavik: It's consistent with the previous code, though...

poncovka: Good catch! I have added a new commit with the fix and additional tests.

M4rtinK (approved):
Looks good to me, thanks!

rvykydal (approved):

jkonecny12 (approved):
Looks great to me. Thanks!

PR #3700, opened 2021-11-10
Remove the BaseLivePayload class

poncovka:
* Merge the code for tar and image installations.
* Update the kernel version list right before the installation.


jkonecny12 (approved):
Looks good to me.

PR #3701, opened 2021-11-11
Fix the condition for completing the root configuration (#1981807)

poncovka:
Revert changes in the `completed` property of the root configuration spoke
that were done in the commits e1dd6aa and 862eee2. They were never necessary.

Resolves: rhbz#1981807

### DIFF
--- a/pyanaconda/ui/gui/spokes/root_password.py
+++ b/pyanaconda/ui/gui/spokes/root_password.py
@@ -196,10 +196,7 @@ class PasswordSpoke(FirstbootSpokeMixIn, NormalSpoke, GUISpokeInputCheckHandler)
 
     @property
     def completed(self):
-        return bool(
-            self._users_module.IsRootPasswordSet or
-            (self._users_module.IsRootAccountLocked and flags.automatedInstall)
-        )
+        return self._users_module.IsRootPasswordSet
 
     @property
     def sensitive(self):
--- a/pyanaconda/ui/gui/spokes/root_password.py
+++ b/pyanaconda/ui/gui/spokes/root_password.py
@@ -31,7 +31,6 @@ from pyanaconda.ui.gui.helpers import GUISpokeInputCheckHandler
 from pyanaconda.ui.gui.utils import set_password_visibility
 from pyanaconda.ui.common import FirstbootSpokeMixIn
 from pyanaconda.ui.communication import hubQ
-from pyanaconda.ui.lib.services import is_reconfiguration_mode
 
 from pyanaconda.anaconda_loggers import get_module_logger
 log = get_module_logger(__name__)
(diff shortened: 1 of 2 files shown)


VladimirSlavik:
Looks good at a glance. However, make sure to check behavior on repeated changes from ui. The "caching" of values in the spokes and module sometimes interact weirdly.

jkonecny12:
It looks simple, elegant and correct to me. However, we really need to check if we won't break something else. Could you please run appropriate KS tests on this (maybe all of them)?

Probably a dumb question but did you tested the issue what RHEV is facing? If this will stop the installation in case the root password is not set?

poncovka:
I have tested the changes.

The good stuff that will be fixed with this pull request:

* The root spokes **blocks** the installation on RHVH with the default kickstart file.
* The root spoke is **visible** and **blocks** the installation for an empty kickstart file in TUI and GUI.
* It worked this way before the rebase and it was broken since then in RHEL 8.

The questionable stuff that will change with this pull request:

* The root spoke can be **entered** with `rootpw --lock` in a kickstart file. 
* It wasn't possible to enter the spoke in RHEL 8.
* It is possible to enter the GUI spoke in upstream and RHEL 9.
* It is possible to enter the TUI spoke in upstream, but not in RHEL 9 - this is a completely different bug.

@jkonecny12, @jstodola, are the changes acceptable?

poncovka:
RHVH without the fix:
RHVH_before

RHVH with the fix:
RHVH

jstodola:
> The questionable stuff that will change with this pull request:
> 
>     * The root spoke can be **entered** with `rootpw --lock` in a kickstart file.

The [installation guide](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/performing_an_advanced_rhel_installation/kickstart-commands-and-options-reference_installing-rhel-as-an-experienced-user#rootpw-required_kickstart-commands-for-system-configuration) answers the question:

`--lock - If this option is present, the root account is locked by default. This means that the root user will not be able to log in from the console. This option will also disable the Root Password screens in both the graphical and text-based manual installation. `

We should not change this documented behavior.

jstodola:
>     * It is possible to enter the GUI spoke in upstream and RHEL 9.

https://bugzilla.redhat.com/show_bug.cgi?id=2026916

jkonecny12:
Are we somehow able to change https://github.com/rhinstaller/anaconda/pull/3701#issuecomment-979901415 without changing other behavior? Sounds to me that it should be possible to just detect that rootpw --locked was used?

poncovka:
> Are we somehow able to change [#3701 (comment)](https://github.com/rhinstaller/anaconda/pull/3701#issuecomment-979901415) without changing other behavior? Sounds to me that it should be possible to just detect that rootpw --locked was used?

It is doable. I have a suspicion that the initial setup allows to enable a locked root account, so we might need a workaround for that. There is a new commit with suggested changes, but it needs to be tested.

VladimirSlavik (commented):
LGTM, as far as I can tell. Untangling `sensitive` from `completed` is a huge win.

poncovka:
Retested with Anaconda and Initial Setup.

VladimirSlavik (approved):
Looks good to me. Thank you!

jkonecny12 (approved):
Looks good to me. Thanks!

poncovka:
I have found another bug. See the last commit, please. I planned to remove this code anyway, so let's do it now since it doesn't work properly. @jstodola, @jkonecny12, @M4rtinK

jstodola:
Ack.

rvykydal (approved):
Looks good to me.

jkonecny12 (approved):
Thanks for this fix. Looks good to me.

PR #3704, opened 2021-11-12

M4rtinK:
… task queue

running).

Resolves: rhbz#1824422

### DIFF
--- a/pyanaconda/installation.py
+++ b/pyanaconda/installation.py
@@ -36,6 +36,7 @@ from pyanaconda.threading import threadMgr
 from pyanaconda.kickstart import runPostScripts, runPreInstallScripts
 from pyanaconda.kexec import setup_kexec
 from pyanaconda.installation_tasks import Task, TaskQueue
+from pyanaconda.progress import progressQ
 from pykickstart.constants import SNAPSHOT_WHEN_POST_INSTALL
 
 from pyanaconda.anaconda_loggers import get_module_logger
@@ -234,22 +235,17 @@ def _prepare_installation(payload, ksdata):
     installation_queue.task_completed.connect(lambda x: progress_step(x.name))
 
     # This should be the only thread running, wait for the others to finish if not.
+    #
+    # We need to do this before we start filling the task queue, or else the DBus
+    # tasks we create might have stale data due to background processing threads
+    # still running at task creation time.
     if threadMgr.running > 1:
-        # it could be that the threads finish execution before the task is executed,
-        # but that should not cause any issues
-
-        def wait_for_all_treads():
-            for message in ("Thread %s is running" % n for n in threadMgr.names):
-                log.debug(message)
-            threadMgr.wait_all()
-
-        # Use a queue with a single task as only TaskQueues have the status_message
-        # property used for setting the progress status in the UI.
-        wait_for_threads = TaskQueue("Wait for threads to finish",
-                                     N_("Waiting for %s threads to finish") % (threadMgr.running - 1))
-
-        wait_for_threads.append(Task("Wait for all threads to finish", wait_for_all_treads))
-        installation_queue.append(wait_for_threads)
+        # show a progress message
+        progressQ.send_message(N_("Waiting for %s threads to finish") % (threadMgr.running - 1))
+        for message in ("Thread %s is running" % n for n in threadMgr.names):
+            log.debug(message)
+        threadMgr.wait_all()
+        log.debug("No more threads are running, assembling installation task queue.")
 
     # Save system time to HW clock.
     # - this used to be before waiting on threads, but I don't think that's needed


pyanaconda/installation.py
```diff
@@ -234,22 +234,15 @@ def _prepare_installation(payload, ksdata):
...
-            for message in ("Thread %s is running" % n for n in threadMgr.names):
-                log.debug(message)
+        for message in ("Thread %s is running" % n for n in threadMgr.names):
+            log.debug(message)
             threadMgr.wait_all()
-
-        # Use a queue with a single task as only TaskQueues have the status_message
-        # property used for setting the progress status in the UI.
-        wait_for_threads = TaskQueue("Wait for threads to finish",
-                                     N_("Waiting for %s threads to finish") % (threadMgr.running - 1))
```
poncovka, line 249: I think we should call `progressQ.send_message` before the waiting to show the same message on the progress screen as before.

M4rtinK: Good point, added - as far as I can tell, the output now looks similar to what was shown before, possibly just in different order due to the insufficient synchronization before.

poncovka (approved):
Looks good to me. Thanks!

pyanaconda/installation.py
```diff
@@ -234,22 +235,17 @@ def _prepare_installation(payload, ksdata):
...
+    # still running at task creation time.
     if threadMgr.running > 1:
-        # it could be that the threads finish execution before the task is executed,
-        # but that should not cause any issues
-
-        def wait_for_all_treads():
-            for message in ("Thread %s is running" % n for n in threadMgr.names):
-                log.debug(message)
+        # show a progress message
+        progressQ.send_message(N_("Waiting for %s threads to finish") % (threadMgr.running - 1))
```
jkonecny12, line 244: I'm not sure if this is not a too technical message for a user. Maybe `Waiting for %s background jobs to finish` would be less technical? Not sure really.

pyanaconda/installation.py
```diff
@@ -234,22 +235,17 @@ def _prepare_installation(payload, ksdata):
...
+        for message in ("Thread %s is running" % n for n in threadMgr.names):
```
jkonecny12, line 245: I would also change the message here to something like `Waiting on thread %s to finish` to know why is the information important.

jkonecny12 (approved):
Looks good to me. Just a bit improvement on messaging. However, I won't insist on that.

pyanaconda/installation.py
```diff
@@ -234,22 +235,17 @@ def _prepare_installation(payload, ksdata):
...
+            log.debug(message)
             threadMgr.wait_all()
```
rvykydal, line 247: It seems to me that waiting for the methods should go out of the for loop ?

rvykydal, line 247: It seems to me that waiting for the threads should go out of the for loop  ?

poncovka: It should. Good catch!

pyanaconda/installation.py
poncovka: These are the same messages as before. Do we really want to change them in RHEL 8?

M4rtinK: Agreed with @poncovka, I think we want to keep this externally consistent, at least on the RHEL branches.

jkonecny12: OK, good point.

pyanaconda/installation.py
M4rtinK: Oh, right - I somehow thought this is intentional (and this code was the same before) and `wait_all()` returns every time a thread terminates, so you get a nice listing of threads as they terminate. But nope, `wait_all()` does indeed wait for all currently running non-main threads, so it really is pointless to all it like this. :D Lets fix that. :)

M4rtinK: Looking even more closely, the old method did not have `wait_all()` in the for cycle - double oops. :P

M4rtinK: Anyway should be fixed now. ;-)

rvykydal (approved):

VladimirSlavik (approved):
Looks good to me. Thank you!

poncovka (approved):

pyanaconda/installation.py
```diff
@@ -234,22 +235,17 @@ def _prepare_installation(payload, ksdata):
     installation_queue.task_completed.connect(lambda x: progress_step(x.name))
 
     # This should be the only thread running, wait for the others to finish if not.
```
poncovka, line 237: In upstream, could you move this code to a new function and call it called from `run_installation`, please?

M4rtinK: Sounds good, let's do that in the upstream version. :)

M4rtinK: changed in upstream PR: #3720

M4rtinK:
Port to Fedora: #3720

M4rtinK:
Port to RHEL 9: #3722

PR #3705, opened 2021-11-12
Reset the goal during clearing the DNF cache (#2020754)

poncovka:
We shouldn't keep data about the previous transaction. We will reset it anyway.

Resolves: rhbz#2020754


jkonecny12 (approved):
Seems to me that we should really be doing this.

VladimirSlavik (approved):
+1, thank you.

PR #3708, opened 2021-11-19
Fix double quotes in dracut package module-setup

VladimirSlavik:
Shellcheck approved.

This is the easy part. It's a separate commit and PR because the failure would manifest in yet another place.

I believe the full ISO build during kickstart tests should be enough to test this, but I'm not sure.


poncovka (approved):
Thanks!

PR #3709, opened 2021-11-19
Master dracut shellcheck level 2

VladimirSlavik:
More fixes, this time miscellaneous ones.

Blocked/draft, because ~~the last few warnings should go with this too, and they need significantly more work.~~ we want to wait until dust settles after the first batch. ~~Feel free to comment on this.~~

### DIFF
--- a/dracut/.shellcheckrc
+++ b/dracut/.shellcheckrc
@@ -1,0 +1,6 @@
+# ShellCheck config file
+
+disable=SC1090  # Can't follow non-constant source. Use a directive to specify location.
+disable=SC2230  # which is non-standard. Use builtin 'command -v' instead.
+disable=SC1091  # Not following: <some-file> was not specified as input (see shellcheck -x).
+disable=SC2154  # <variable-name> is referenced but not assigned.
--- a/dracut/anaconda-diskroot
+++ b/dracut/anaconda-diskroot
@@ -24,7 +24,7 @@ run_checkisomd5() {
                 rc=$?
                 state="inactive"
             fi
-            if [ "$rc" == "1" ]; then
+            if [ "$rc" = "1" ]; then
                 die "CD check failed!"
                 exit 1
             fi
--- a/dracut/anaconda-diskroot
+++ b/dracut/anaconda-diskroot
@@ -10,7 +10,7 @@ run_checkisomd5() {
     livedev=$1
     if getargbool 0 rd.live.check -d check; then
         [ -b "$livedev" ] && fs=$(blkid -s TYPE -o value "$livedev")
-        if [ "$fs" = "iso9660" -o "$fs" = "udf" ]; then
+        if [ "$fs" = "iso9660" ] || [ "$fs" = "udf" ]; then
             [ -x /bin/plymouth ] && /bin/plymouth --hide-splash
             if [ -n "$DRACUT_SYSTEMD" ]; then
                 p=$(dev_unit_name "$livedev")
@@ -71,7 +71,7 @@ fi
 
 # If interactive dd has been requested and this is a cdrom, wait to mount it
 # later. The user may want to swap in a driver update cdrom first.
-if [ -e /tmp/dd_interactive -a ! -e /tmp/dd.done ]; then
+if [ -e /tmp/dd_interactive ] && [ ! -e /tmp/dd.done ]; then
     if dev_is_cdrom "$dev"; then
         > /tmp/anaconda-dd-on-cdrom
         exit 0
--- a/dracut/anaconda-diskroot
+++ b/dracut/anaconda-diskroot
@@ -64,7 +64,7 @@ fi
 # retrigger this script, and we'll mount the disk as normal.
 if str_starts "$kickstart" "cdrom" && [ ! -e /tmp/ks.cfg.done ]; then
     if dev_is_cdrom "$dev"; then
-        > /tmp/anaconda-on-cdrom
+        true > /tmp/anaconda-on-cdrom
         exit 0
     fi
 fi
@@ -73,7 +73,7 @@ fi
 # later. The user may want to swap in a driver update cdrom first.
 if [ -e /tmp/dd_interactive ] && [ ! -e /tmp/dd.done ]; then
     if dev_is_cdrom "$dev"; then
-        > /tmp/anaconda-dd-on-cdrom
+        true > /tmp/anaconda-dd-on-cdrom
         exit 0
     fi
 fi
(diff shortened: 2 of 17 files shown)


dracut/anaconda-diskroot
```diff
@@ -24,7 +24,7 @@ run_checkisomd5() {
                 rc=$?
                 state="inactive"
             fi
-            if [ "$rc" == "1" ]; then
+            if [ "$rc" = "1" ]; then
```
poncovka, line 27: Please, split this commit or at least add explanations to the commit message. I like how it is done in other commits, because it is very easy to follow the changes.

poncovka (commented):
It looks great!

dracut/anaconda-lib.sh
```diff
@@ -382,7 +387,7 @@ run_kickstart() {
     # net: re-run online hooks
     if [ "$do_net" ]; then
         # If NetworkManager is used in initramfs
-        if [ -e "$hookdir"/cmdline/*-nm-config.sh ]; then
+        if ls -U "$hookdir"/cmdline/*-nm-config.sh >/dev/null 2>&1 ; then
```
jkonecny12, line 390: I wonder what  `-U` is, I can't find it in the man page on my Fedora...

dracut/anaconda-pre-shutdown.sh
```diff
@@ -1,3 +1,4 @@
+#!/bin/sh
```
jkonecny12, line 1: Shouldn't we use bash here too?

dracut/anaconda-diskroot
```diff
@@ -64,7 +64,7 @@ fi
 # retrigger this script, and we'll mount the disk as normal.
 if str_starts "$kickstart" "cdrom" && [ ! -e /tmp/ks.cfg.done ]; then
     if dev_is_cdrom "$dev"; then
-        > /tmp/anaconda-on-cdrom
+        true > /tmp/anaconda-on-cdrom
```
jkonecny12, line 67: Couldn't we rather use `touch` than this form?

dracut/fetch-driver-net.sh
```diff
@@ -5,7 +5,7 @@
 command -v warn_critical >/dev/null || . /lib/anaconda-lib.sh
 
 # initqueue/online hook passes interface name as $1
-netif="$1"
+#netif="$1"
```
jkonecny12, line 8: Why is this commented out?

jkonecny12 (commented):
Looks good to me otherwise. I agree with Vendy that spliting to commits is much easier to read.

dracut/.shellcheckrc
```diff
@@ -0,0 +1,6 @@
+# ShellCheck config file
+
+disable=SC1090  # Can't follow non-constant source. Use a directive to specify location.
+disable=SC2230  # which is non-standard. Use builtin 'command -v' instead.
+disable=SC1091  # Not following: <some-file> was not specified as input (see shellcheck -x).
+disable=SC2154  # <variable-name> is referenced but not assigned.
```
jkonecny12, line 6: Please also add this file to the https://github.com/rhinstaller/anaconda/blob/rhel-9/.structure-config file.

dracut/anaconda-diskroot
VladimirSlavik: Yes, I agree. I was afraid it will be a lot of commit spam, but initially had these changes split.

dracut/anaconda-lib.sh
VladimirSlavik: On f32 I have it? It's "do not sort before output". Technically it saves performance in cases where you have many files (millions). Perhaps it is not necessary here.

dracut/anaconda-diskroot
VladimirSlavik: Good point. I was so hung up on "fixing the issues" that I missed the obvious solutions :-/

dracut/fetch-driver-net.sh
VladimirSlavik: Sorry, that's a mistake. The variable was not used. Actually, do we want to keep the comment above or drop it too?

dracut/anaconda-pre-shutdown.sh
VladimirSlavik: That's a tough subject. Sorry I don't have a tl;dr.

I originally believed the hashbang mattered, but now I believe it does not, in dracut hooks at least. That's because dracut sources the hooks instead of calling. We do know that bashisms work, so perhaps we should go for bash everywhere, except dracut wants hooks to be POSIX compliant, so perhaps we should go for that instead?

However harmonizing the hashbangs could be an useless project.

Shellcheck changes rules depending on the hashbang, so it matters to some extent. Here I set it to the more strict mode, since the file did pass that as-is. Elsewhere I changed to bash instead, so as to make the file formally compliant and avoid more changes.

dracut/anaconda-diskroot
VladimirSlavik: No `touch` in initramfs, I'm afraid.

VladimirSlavik:
Split into more commits with more comments in commit messages.

poncovka (approved):

VladimirSlavik:
The `dns` test is broken due to the `resolv.conf` issues, so effectively everything passed.

dracut/anaconda-lib.sh
jkonecny12: Hmm, I see it now... I wonder why I missed it before. Yeah, could stay this way.

dracut/anaconda-pre-shutdown.sh
jkonecny12: I wouldn't really like ta change it everywhere or something similar. To me it's just strange that we have `bash` everywhere except this file in the commit.

dracut/anaconda-diskroot
jkonecny12: lol, ok

dracut/fetch-driver-net.sh
jkonecny12: I would rather remove it. For these situations there is git history.

dracut/anaconda-diskroot
```diff
@@ -24,7 +24,7 @@ run_checkisomd5() {
                 rc=$?
                 state="inactive"
             fi
-            if [ "$rc" == "1" ]; then
```
jkonecny12, line 27: This doesn't feel correct. However, the question here is what we would like to follow POSIX or bash? I think we should decide and use that consistently everywhere to avoid confusions. (Not saying in this PR).

jkonecny12 (approved):
Looks good to me. Just a note below.

dracut/anaconda-diskroot
VladimirSlavik: Yup, I commented on that earlier.

VladimirSlavik:

Rebased to resolve conflict in `.structure-config`.

PR #3710, opened 2021-11-19
Decorate installer window on Live

VladimirSlavik:
Live has the GNOME bar at top, and dragging down from it will "restore" the window to a nonsensical small size. Without decorations, the window cannot be maximized again except for keyboard shortcuts.

Resolves: [rhbz#1998720](https://bugzilla.redhat.com/show_bug.cgi?id=1998720)

### DIFF
--- a/data/profile.d/fedora-workstation.conf
+++ b/data/profile.d/fedora-workstation.conf
@@ -18,6 +18,7 @@ menu_auto_hide = True
 
 [User Interface]
 custom_stylesheet = /usr/share/anaconda/pixmaps/workstation/fedora-workstation.css
+decorated_window = True
 hidden_spokes =
     NetworkSpoke
     PasswordSpoke


data/profile.d/fedora-workstation.conf
```diff
@@ -18,6 +18,7 @@ menu_auto_hide = True
 
 [User Interface]
 custom_stylesheet = /usr/share/anaconda/pixmaps/workstation/fedora-workstation.css
+decorated_window = True
```
poncovka, line 21: I am afraid that this will have to be changed in the code specifically for live installations (probably in the `set_from_opts` method). The suggested solution would affect a bunch of other products and environments.

VladimirSlavik: Why not just change the value in the derived configs?

jkonecny12: That would work but in general we could break someone else use-case. I know it's not probably happening right now but in general we probably should start thinking about the backward compatibility. Please take me here with the reserve, just saying that it could be important in the future to think about that.

In the light of the above, I'm thinking if something like this (which is specific for Live only) shouldn't be a separate profile. I mean changing this profile to something like "base" profile for inheritance of other products (Workstation, SilverBlue) but next to that create a profile which would be Live specific with this option used there. Just brainstorming here.

VladimirSlavik: Ah! I meant, this change + explicitly set the value to False in all the dependent config files. The base live profile could be a good idea too. But IIRC we had this some time ago and stopped using it. Maybe Vendy knows more?

What are the config objects meant to be, anyway? I always saw them as mainly just a dumb storage. Having some logic override values loaded from the (pro)files does not sound right to me. We do have some logic to provide sane defaults, which is pretty close, but always respects the values from files.

poncovka: I don't want to have a base live profile, because you can easily get into a trap of multiple inheritance that we are trying to avoid. If we need to change something for all live environments, we can do that in the code, the same way as for image and dir installations.

In this case, we should first figure out what we are trying to achieve. I thought that we will enable decorations for all live environments. The thing is that there are several spins, some of them have configs and we cannot be sure which one are based on GNOME. Also, it looks a little weird that Anaconda runs in this semi-fullscreen mode.

VladimirSlavik: This is how I understand the bug and fixing it:
- Bug: GNOME lets users de-maximize a chromeless window.
- End goal: Let users re-maximize the window.
- Side effect: We are also letting users unmaximize the window and resize. We are OK with it because it is an option that we support and can just turn on.
- Inheritance in config files implies inheritance of changes.
- Rawhide is for changes and this is a change, business as usual.

Conan-Kudo: Doing it this way would be a problem for when the same product profile is used for both Live and non-Live setups (e.g. CentOS Hyperscale, AlmaLinux, etc.). I agree it should probably be done in the code to detect a live environment over the installer-only one.

VladimirSlavik:
This particular config field will be probably removed as it has been unused so far. Closing.

PR #3711, opened 2021-11-19
Use a task to download the image payload

poncovka:
* Use the task in the image payload class.
* Add unit tests for the image download.


pyanaconda/modules/payloads/payload/live_image/download_progress.py
```diff
@@ -0,0 +1,72 @@
...
+        """Report the progress."""
+        pct = min(int(100 * self._downloaded_size / self._total_size), 100)
+
+        if pct == self._last_pct:
+            return
+
+        self._last_pct = pct
+
+        log.debug("Downloaded %s (%s%%)", Size(self._downloaded_size), pct)
+        self._callback(_("Downloading {} ({}%)").format(self._url, pct))
```
VladimirSlavik, line 54: What do you think about unifying this? I find it fascinating that the debug call does not mention the URL, while the callback does not include the size, so you need both to get full information.

jkonecny12 (approved):
Looks good to me.

pyanaconda/modules/payloads/payload/live_image/download_progress.py
poncovka: I don't think we want to do that. The size woudn't say anything useful to the user and the URL wouldn't say anything useful to us. We would just burden everyone with unnecessary data.

PR #3713, opened 2021-11-20
Fixed the translation not taking effect

layne-yang:
The position of the sentence here is obvious, it is necessary to add a translation

Signed-off-by: yangxiaoxuan <yangxiaoxuan@uniontech.com>

### DIFF
--- a/pyanaconda/ui/gui/spokes/lib/accordion.py
+++ b/pyanaconda/ui/gui/spokes/lib/accordion.py
@@ -544,7 +544,7 @@ class CreateNewPage(BasePage):
         )
         self._createBox.attach(label, 0, 6, 2, 1)
 
-        checkbox = Gtk.CheckButton(label=C_("GUI|Storage", "Encrypt my data."))
+        checkbox = Gtk.CheckButton(label=C_("GUI|Custom Partitioning|Autopart Page", "Encrypt my data."))
         checkbox.connect("toggled", encrypted_changed_cb)
         checkbox.set_active(default_encryption)
         checkbox.set_margin_start(18)
--- a/pyanaconda/ui/gui/spokes/lib/accordion.py
+++ b/pyanaconda/ui/gui/spokes/lib/accordion.py
@@ -544,7 +544,7 @@ class CreateNewPage(BasePage):
         )
         self._createBox.attach(label, 0, 6, 2, 1)
 
-        checkbox = Gtk.CheckButton(label="Encrypt my data.")
+        checkbox = Gtk.CheckButton(label=C_("GUI|Storage", "Encrypt my data."))
         checkbox.connect("toggled", encrypted_changed_cb)
         checkbox.set_active(default_encryption)
         checkbox.set_margin_start(18)


VladimirSlavik (approved):
Looks good, thank you for catching this!

Have you considered adding a mnemonic / accelerator key to the label too? It would have to not collide with the ones that are used by controls in the right pane (`CTyLEoMVINU`). Perhaps `r` would be good enough? `"Enc_rypt my data."`

VladimirSlavik:
Ah, sorry, disregard the accelerator thing, it's already there and handled via the label above.

VladimirSlavik:
On a further examination, I see you used a different context than the surrounding code: `"GUI|Storage"` vs. `"GUI|Custom Partitioning|Autopart Page"`.

I think I can understand why - the GUI controls and their roles are identical, so it makes sense to reuse the same translation. However, it also feels as if the one "area"should use the same context. I am not sure which is better here. The gettext manuals do not offer much guidance on choice of context, and it seems "best practices" for this are not exactly easy to find.

Personally, I feel that the pragmatical choice is to use the context `"GUI|Custom Partitioning|Autopart Page"` precisely because it will not reuse the same translation. The translation contains the accelerator character, and if the same is used in multiple places, there is a chance it will break somewhere. When translating, Weblate will offer this string as a "similar one" to review and also offer the translation as a suggestion, I believe. So it is not much of a problem to copy...

However this is very much a matter of opinions, so I leave this note here so that others are aware of this potential issue and can comment.

layne-yang:
> 

The reason why I did not use the same context is that there are already the same translation items in the po files, so that the modification of the po files can be omitted.

layne-yang:
> On a further examination, I see you used a different context than the surrounding code: `"GUI|Storage"` vs. `"GUI|Custom Partitioning|Autopart Page"`.
> 
> I think I can understand why - the GUI controls and their roles are identical, so it makes sense to reuse the same translation. However, it also feels as if the one "area"should use the same context. I am not sure which is better here. The gettext manuals do not offer much guidance on choice of context, and it seems "best practices" for this are not exactly easy to find.
> 
> Personally, I feel that the pragmatical choice is to use the context `"GUI|Custom Partitioning|Autopart Page"` precisely because it will not reuse the same translation. The translation contains the accelerator character, and if the same is used in multiple places, there is a chance it will break somewhere. When translating, Weblate will offer this string as a "similar one" to review and also offer the translation as a suggestion, I believe. So it is not much of a problem to copy...
> 
> However this is very much a matter of opinions, so I leave this note here so that others are aware of this potential issue and can comment.

The reason why I did not use the same context is that there are already the same translation items in the po files, so that the modification of the po files can be omitted.

jkonecny12:
Hi, I would rather use the different context than the same. As said the Weblate will guide you and it will avoid situations where in some language it would be better to have a different string for these places. But this could be just my opinion.

poncovka:
Hi @layne-yang, I agree with @VladimirSlavik and @jkonecny12. Could you change the context as suggested, so we can merge this pull request, please?

layne-yang:
> Hi @layne-yang, I agree with @VladimirSlavik and @jkonecny12. Could you change the context as suggested, so we can merge this pull request, please?

Of course.

layne-yang:
> Hi @layne-yang, I agree with @VladimirSlavik and @jkonecny12. Could you change the context as suggested, so we can merge this pull request, please?

layne-yang:
Hi @poncovka ，

layne-yang:
Hi@poncovka ，I have changed the context as suggested, but rpm build failed. The reason is check key URLs failed。

PR #3719, opened 2021-11-24
User is admin by default

VladimirSlavik:
TODO: link Fedora change and bug here


poncovka (approved):
Thank you!

jstodola:
Would it make sense to have an anaconda configuration option for that? I can imagine that some future products might not want to create users as admins by default.

VladimirSlavik:
It isn't my first choice - growing the config api means more responsibility - but I agree that is a possibility.

jstodola:
Yes, it sounds reasonable. An anaconda config option can be added anytime in the future if necessary.

jkonecny12 (approved):
Looks good to me.

VladimirSlavik:
@mairin do you think you could help with this?

In the [emails on fedora-devel](https://lists.fedoraproject.org/archives/list/devel@lists.fedoraproject.org/thread/VAAXLGEZQCFITMTPUXKHIC3ZPKBTEVNM/), somebody said the checkbox `Make this user administrator` should have a better description. Suggestions voiced there:
- "Grant user administrator privileges (allow sudo)"
- The explanation can be even longer: maybe "(e.g. allow sudo as root, access to all logs, and other administrative actions)"

Do you have any opinion on the checkbox wording?

mairin:
@VladimirSlavik Just as a quick "competitive analysis" here's what OS X and Windows do -

**OS X** - very similar to ours, a checkbox with the text "Allow user to administer this computer"
**Windows** - very verbose and detailed explanation on a radio button "Administrators have complete control over the PC. They can change any settings and access all of the files and programs stored on the PC."

I don't think install time is the ideal place for a long detailed explanation. I think the main concern I read in the comments on fedora-devel was confusion over whether or not the resulting user would be root or have sudo access. There was also concern users who were less experienced with UNIX-like systems would know what 'sudo' meant. Therefore I would propose this text change:

"[x] Add administrative privileges to this user account (`wheel` group membership)"

I think this, in describing the account as a user account that privileges are being added to, makes it more clear that it's not making the user root (which wouldnt make sense anyway since the username isn't root) and that it's an additive property instead of swapping out the account type. If the user wants to learn more about what this means, if you search for "wheel group" on google, ddg, etc some of the obvious top hits explain it. 

This also mirrors the official documentation on this point, which says: 

" Select the Make this user administrator check box if the user requires administrative rights (the installation program adds the user to the wheel group ). "

VladimirSlavik:
Thank you, changed.

poncovka (approved):

jkonecny12 (approved):
Looks good to me.

poncovka (approved):

jkonecny12 (approved):
LGTM!

PR #3720, opened 2021-11-24

M4rtinK:
… task queue

running).

(cherry picked from commit 734befd4c4c5ce5e8bc73732455d87b672598f3c)

Related: rhbz#1824422
Related: rhbz#2017428


M4rtinK:
This is the upstream port of RHEL 8 PR #3704.

poncovka (approved):
Thanks!

VladimirSlavik (approved):
Thank you!

jkonecny12 (approved):
LGTM!

PR #3721, opened 2021-11-25
Switch lorax to RHEL-9 GA (#infra)

poncovka:
It was still using Beta links before.


rvykydal (approved):
Looks good to me, thank you.

PR #3722, opened 2021-11-25

M4rtinK:
… task queue

running).

(cherry picked from commit 734befd4c4c5ce5e8bc73732455d87b672598f3c)
(cherry picked from commit f3539304f7a5c968d94f2f4dd6f61a4887fb0f57)

Resolves: rhbz#2017428


M4rtinK:
Upstream PR: #3720
RHEL 8 PR: #3704

rvykydal (approved):
Looks good to me.

PR #3723, opened 2021-11-25
Mount and unmount the live image with tasks

poncovka:
Run a task to mount a live image. Originally, the mount point for the installation
of the image was always `/run/install/source` and we had to sometimes move mounts
to achieve that. That is too complicated. The task can just return a mount point
that should be used for the installation and that's it.

Run tasks for unmounting and deleting the live image. Don't fail if the mount
points or files don't exist.

### DIFF
--- a/pyanaconda/payload/live/payload_liveimg.py
+++ b/pyanaconda/payload/live/payload_liveimg.py
@@ -15,17 +15,15 @@
 # License and may only be used or replicated with the express permission of
 # Red Hat, Inc.
 #
-import glob
 import os
 
 from blivet.size import Size
 from pyanaconda.anaconda_loggers import get_packaging_logger
-from pyanaconda.core import util
 from pyanaconda.core.configuration.anaconda import conf
 from pyanaconda.core.constants import PAYLOAD_TYPE_LIVE_IMAGE, INSTALL_TREE, IMAGE_DIR
 from pyanaconda.modules.common.structures.live_image import LiveImageConfigurationData
 from pyanaconda.modules.payloads.payload.live_image.installation import VerifyImageChecksum, \
-    InstallFromTarTask, InstallFromImageTask, DownloadImageTask
+    InstallFromTarTask, InstallFromImageTask, DownloadImageTask, MountImageTask
 from pyanaconda.modules.payloads.payload.live_os.utils import get_kernel_version_list
 from pyanaconda.modules.payloads.payload.live_image.utils import get_kernel_version_list_from_tar
 from pyanaconda.modules.payloads.source.live_image.initialization import SetUpLocalImageSourceTask, \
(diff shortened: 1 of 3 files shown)


pyanaconda/payload/live/payload_liveimg.py
```diff
@@ -126,47 +123,14 @@ def pre_install(self):
...
-        if not os.path.exists(INSTALL_TREE + "/LiveOS"):
-            return
-
-        # Mount the first .img in the directory on INSTALL_TREE
-        img_files = glob.glob(INSTALL_TREE + "/LiveOS/*.img")
-        if img_files:
-            # move the mount to IMAGE_DIR
-            os.makedirs(IMAGE_DIR, 0o755)
-            rc = util.execWithRedirect("mount",
-                                       ["--move", INSTALL_TREE, IMAGE_DIR])
```
VladimirSlavik, line 160: There is no indication why this code is not needed anymore. Can you please describe that in the commit message?

pyanaconda/payload/live/payload_liveimg.py
```diff
@@ -156,11 +154,15 @@ def post_install(self):
             If file:// was used, just unmount it.
         """
         super().post_install()
-        payload_utils.unmount(IMAGE_DIR)
-        payload_utils.unmount(INSTALL_TREE)
+        task = TearDownMountTask(IMAGE_DIR)
+        task.run()
 
-        if os.path.exists(self.image_path) and not self.data.liveimg.url.startswith("file://"):
```
VladimirSlavik, line 202: Please explain why the second part of the condition is not needed? AFAICT it was added in response to [rhbz1164585](https://bugzilla.redhat.com/show_bug.cgi?id=1164585) in 290d46bd4c4afae3174089cdc45ac69f4930da4b.

VladimirSlavik (commented):
I have no idea why some of the code is removed without any replacement. It might be related to some earlier PR, but I could not find that.

pyanaconda/payload/live/payload_liveimg.py
poncovka: The commit message was updated.

pyanaconda/payload/live/payload_liveimg.py
poncovka: The task checks the existence of the file.

VladimirSlavik: I am asking why you want to delete the file if it was a local one. Perhaps the part I am not seeing is why the file can be deleted. Do we make a copy of it even if it's a `file:///` kind of url?

poncovka: This condition is still there. Look two lines below.

VladimirSlavik (approved):
Looks good to me. Thank you!

Sorry for missing the condition :-/

jkonecny12 (commented):
It's probably just me, but I'm confused here. Why we are not using source and instead doing the mount and download in the payload directly?

poncovka:
> It's probably just me, but I'm confused here. Why we are not using source and instead doing the mount and download in the payload directly?

The payload sources are set up before the installation is started. However, the live image is downloaded during the installation after the storage configuration to have enough space. Therefore, it has to be mounted after that. I have some plans to move these tasks to sources, but they cannot be part of set-up and tear-down tasks.

PR #3724, opened 2021-11-26
Remove unused code from the live image payload module

poncovka:
These tasks were never used in the UI and only duplicated the code. It would
be difficult to adapt them and use them for the migration of the code from
the payload class to the DBus module, so we created new ones.


jkonecny12 (approved):
Seems a bit sad to me but I agree.

PR #3726, opened 2021-11-26
Do not modify boot order on UEFI if asked

VladimirSlavik:
Resolves: [rhbz#2025953](https://bugzilla.redhat.com/show_bug.cgi?id=2025953)


jkonecny12 (approved):
Looks good to me. Thanks!

pyanaconda/modules/storage/bootloader/efi.py
```diff
@@ -69,8 +69,10 @@ def _add_single_efi_boot_target(self, partition):
         boot_disk = partition.disk
         boot_part_num = str(partition.parted_partition.number)
 
+        create_method = "-C" if self.keep_boot_order else "-c"
```
jkonecny12, line 72: You have to disable pylint for this line as it is done in on line 125.

VladimirSlavik: Oh, ouch, yes. I think there is some way to disable it for the whole file? Spamming the disables everywhere does not seem very useful...

jstodola:
FYI, QE will do manual testing on physical hardware later in the process, so I think it's enough for this PR to confirm that the right commands are run when the installation is executed in a UEFI VM.

VladimirSlavik:
Calls the right args with and without `leavebootorder`, tested.

poncovka (approved):
Thanks1

jkonecny12 (approved):
Looks great to me.

I think pylint has the disable for a file but I would rather avoid it here. The `no-member` issue could really find something.

VladimirSlavik:
I was thinking, when I port to upstream, I will add a commit that moves the disables to function level, I think - that seems better targeted than having every line be an ignore... Then we can discuss the change there.

M4rtinK (approved):
Looks good to me, thanks! :)

PR #3728, opened 2021-11-29
Disable packit smoke test on TFT (#infra)

jkonecny12:
This test is not really helpful and TFT won't be working through packit for some time.


VladimirSlavik (approved):
Looks good to me. Thank you!

jkonecny12:
Sorry for the noise. We don't need to do anything at the end.

PR #3729, opened 2021-11-30
Fix Source0 in spec to point to upstream source URL

jkonecny12:
It should be like that all the time. Also this change is required by Packit for creating downstream releases (it download tarball from there).

### DIFF
--- a/anaconda.spec.in
+++ b/anaconda.spec.in
@@ -10,7 +10,7 @@ URL:     http://fedoraproject.org/wiki/Anaconda
 # git checkout -b archive-branch anaconda-%%{version}-%%{release}
 # ./autogen.sh
 # make dist
-Source0: %{name}-%{version}.tar.bz2
+Source0: https://github.com/rhinstaller/%{name}/releases/download/%{name}-%{version}-1/%{name}-%{version}.tar.bz2
 
 # Versions of required components (done so we make sure the buildrequires
 # match the requires versions of things).


VladimirSlavik (commented):
This is technically correct. But what you are doing with this one line is making Github part of the whole release workflow. We need docs for that, too.

The comment above is now wrong, you need to push to Github. At the same time, maybe it was  broken before, it contains nothing about makebumpver etc. Or maybe the assumption is that these steps are no longer needed, because the tarball just sits at the URL? Maybe @M4rtinK should sort that out.

I think there should also be a hint how to do a 100% local build. But I'm new to packaging, so maybe the other docs already describe the local build.

Either way, the comment(s) should be fixed somehow.

jkonecny12:
> Thanks! This is technically correct. But what you are doing with this one line is making Github part of the whole release workflow. We need docs for that, too.

No, it is not. This link is used by the Packit only. However, the link should be there based on the packager guidelines because then you can verify that the tarball is really the upstream tarball. For the normal release process (without Packit) there is no change, you still have to upload the tarball to the dist-git. No documentation change is needed.

> The comment above is now wrong, you need to push to Github. At the same time, maybe it was broken before, it contains nothing about makebumpver etc. Or maybe the assumption is that these steps are no longer needed, because the tarball just sits at the URL? Maybe @M4rtinK should sort that out.

I think the comment should be removed. For these we have a guidelines in documentation. However, @M4rtinK should agree on that.

VladimirSlavik (approved):
If the release does not need github, then this is fine. Thanks!

The comment is a separate issue, I guess. Maybe it turns out to be correct after all, too?

M4rtinK:
> > Thanks! This is technically correct. But what you are doing with this one line is making Github part of the whole release workflow. We need docs for that, too.
> 
> No, it is not. This link is used by the Packit only. However, the link should be there based on the packager guidelines because then you can verify that the tarball is really the upstream tarball. For the normal release process (without Packit) there is no change, you still have to upload the tarball to the dist-git. No documentation change is needed.

Yeah, I think what's needed is to clarify that when the package is being built in Koji, there is no Internet access - it can't fetch the tarball even if it wanted to. This is on purpose to both make builds deterministic & avoid some types of mallware fetching their payload remotely. 

But this is actually not a problem, as even if there is an URL in this filed, `rpmbuild` only cares about the file at the end of the URL. If there is a valid (or not) URL in front of it does not play any important role when the actual package build happens. Still the packaging guidelines IIRC indeed recommend to use this field to point to the canonical complete upstream tarball & one in a while some sort of scan is run to check which of these URLs are actually working.

BTW, I've checked some of our packages as well as other packages I maintain and most if not all of them point to a tarball, usually hosted on GitHub, or PyPi (dasbus) or even on upstream project website (pyotherside). So it really seems like the correct way to go. :)

> 
> > The comment above is now wrong, you need to push to Github. At the same time, maybe it was broken before, it contains nothing about makebumpver etc. Or maybe the assumption is that these steps are no longer needed, because the tarball just sits at the URL? Maybe @M4rtinK should sort that out.
> 
> I think the comment should be removed. For these we have a guidelines in documentation. However, @M4rtinK should agree on that.

So I've actually tried executing the commands in the comment and there are errors in them:
- you need to `cd` to the cloned repository
- the version spec in the `git checkout` call is missing the obligatory `-1` we still have on all tags so far
- you need to also run `./configure` before you can run make dist

And even if this gives you the tarball, it does not have the translations.

As for not mentioning `makebumpver` - that's actually correct, the instructions are effectively targeting a user or distro package maintainer, who are usually interested just in getting a tarball for a given version of the software. Bumping the version is a responsibility of the upstream project maintainers and IMHO out of scope for the spec file.

Anyway, I agree that we should drop this from the spec file, especially now when we have the automatic tarball builder up and running. Ther instructions have also been more or less broken for quite a while and nobody complained, so that's another data point.

Still, we should definitely have the process of tarball creation documented in our upstream docs - I'll take it over from there, opening a PR removing these lines from the spec as well as fixing up the release docs if necessary. :)

And thanks for spotting this - I don't think I ever noticed these lines till now. ;-)

M4rtinK (approved):
Looks good to me. :)

PR #3730, opened 2021-12-01
Enable the pytest assert introspection in the helper functions

poncovka:
The assert introspection works by default only for test modules. It means that
helper functions from other modules will raise the `AssertionError` exception
without any other useful information. These modules have to be explicitly
registered with the `register_assert_rewrite` function to make the assert
introspection work.

See: https://docs.pytest.org/en/6.2.x/writing_plugins.html#assertion-rewriting

### DIFF
--- a/tests/unit_tests/__init__.py
+++ b/tests/unit_tests/__init__.py
@@ -17,6 +17,7 @@
 #
 import os
 import locale
+import pytest
 import gi.overrides
 
 # Apply overrides for the anaconda widgets.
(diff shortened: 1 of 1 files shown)


jkonecny12 (approved):
Looks great to me. Thanks!

VladimirSlavik (approved):
Looks good to me. Thank you!

VladimirSlavik:
Do you perhaps want to add "port to RHEL 9" ?

jkonecny12:
Already backported to https://github.com/rhinstaller/anaconda/pull/3664 .

PR #3731, opened 2021-12-01
Deprecate the ANA_INSTALL_PATH environment variable

poncovka:
The support for the `ANA_INSTALL_PATH` environment variable will be removed
in future releases. Please, use the `/mnt/sysroot` path instead.


VladimirSlavik:
How does this impact lorax?

poncovka:
> How does this impact lorax?

Lorax doesn't seem to use it.

poncovka:
Btw. I don't think it is documented. It might be possible that we are the only users (see https://github.com/rhinstaller/anaconda/pull/3571 and https://github.com/rhinstaller/kickstart-tests/pull/638).

VladimirSlavik (approved):

PR #3733, opened 2021-12-01
Clean up the task for the checksum verification

poncovka:
* Rename the class to `VerifyImageChecksumTask`.
* Change the arguments of the task.
* Remove a forgotten print statement.

### DIFF
--- a/pyanaconda/modules/payloads/payload/live_image/installation.py
+++ b/pyanaconda/modules/payloads/payload/live_image/installation.py
@@ -144,18 +144,19 @@ class DownloadImageTask(Task):
         progress.end()
 
 
-class VerifyImageChecksum(Task):
+class VerifyImageChecksumTask(Task):
     """Task to verify the checksum of the downloaded image."""
 
-    def __init__(self, image_path, checksum):
+    def __init__(self, configuration: LiveImageConfigurationData, image_path):
         """Create a new task.
 
+        :param configuration: a configuration of a remote image
+        :type configuration: an instance of LiveImageConfigurationData
         :param image_path: a path to the image
-        :param checksum: an expected sha256 checksum
         """
         super().__init__()
         self._image_path = image_path
-        self._checksum = checksum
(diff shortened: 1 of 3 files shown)


poncovka:
Needs a rebase.

VladimirSlavik (approved):
Looks good to me. Thank you!

pyanaconda/modules/payloads/payload/live_image/installation.py
```diff
@@ -144,18 +144,19 @@ def _stream_download(self, response, image_file):
         progress.end()
 
 
-class VerifyImageChecksum(Task):
+class VerifyImageChecksumTask(Task):
     """Task to verify the checksum of the downloaded image."""
 
-    def __init__(self, image_path, checksum):
+    def __init__(self, configuration: LiveImageConfigurationData, image_path):
```
jkonecny12, line 150: I wonder, do we want the type specified here? Isn't it enough to have it in the docstring?

jkonecny12 (approved):
Looks good to me otherwise. Just a question below.

pyanaconda/modules/payloads/payload/live_image/installation.py
poncovka: It is useful for development, if an IDE can do type checking. I can remove it, but I don't think it is so bothering.

VladimirSlavik: In Pycharm, the type checking happens also with data from docstrings - at least for me...

poncovka: That sometimes doesn't work and not all type checkers are able to use docstrings. For example, mypy that we would like to use in the future.

jkonecny12: I don't mind, I just wondered that it's just for this one argument and nothing else. Feel free to keep it here.

PR #3736, opened 2021-12-02
Migrate anaconda-devel mailing list under Fedora

jkonecny12:
Let's change the old variant to a new variant everywhere in our code base. Also fix Dockerfiles which has non-existing mailing list in the LABELs.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -18,7 +18,7 @@
 m4_define(python_required_version, 3.4)
 
 AC_PREREQ([2.63])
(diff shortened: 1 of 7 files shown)


jkonecny12:
I think changing the `master` branch should be enough? Do you agree?

VladimirSlavik (approved):
Looks good to me. Thank you!

Master is enough, docs are not generated for other branches... I think?

PR #3737, opened 2021-12-02
Fix generation of commit range for release notes (#infra)

VladimirSlavik:
Github gives us `${{ github.ref }}` as `refs/tags/***` which apparently does not work in `git describe --exclude`.

See the soft failure in "Generate release" in https://github.com/rhinstaller/anaconda/runs/4395925639?check_suite_focus=true


jkonecny12 (approved):
LGTM, we just need to change this if we decide to remove the `-1` from the version.

PR #3738, opened 2021-12-03
Introduce a download path to simplify the image payload code

poncovka:
Run the image download task for all types of the image URL. If the image is
local, the task returns its local location. Otherwise, the image is downloaded
at the specified download location and the task returns that one.

Always run the image removal task, but use the download path to make sure
that we won't remove a local image.

### DIFF
--- a/pyanaconda/modules/payloads/payload/live_image/installation.py
+++ b/pyanaconda/modules/payloads/payload/live_image/installation.py
@@ -41,18 +41,18 @@ log = get_module_logger(__name__)
 class DownloadImageTask(Task):
     """Task to download an image."""
 
-    def __init__(self, configuration: LiveImageConfigurationData, image_path):
+    def __init__(self, configuration: LiveImageConfigurationData, download_path):
         """Create a new task.
 
         :param configuration: a configuration of a remote image
         :type configuration: an instance of LiveImageConfigurationData
-        :param str image_path: a path to the downloaded image
+        :param str download_path: a path to the downloaded image
         """
         super().__init__()
         self._url = configuration.url
         self._proxy = configuration.proxy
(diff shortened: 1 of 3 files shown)


VladimirSlavik (approved):
Looks good to me. Thank you!

pyanaconda/modules/payloads/payload/live_image/installation.py
```diff
@@ -41,28 +41,39 @@
 class DownloadImageTask(Task):
```
jkonecny12, line 41: When the functionality changed this should be renamed? Maybe to something like `ProvideImageTask` or `GetImageTask`. It's not downloading anything for the local image.

jkonecny12 (commented):
Just a few ideas below.

pyanaconda/modules/payloads/payload/live_image/installation.py
```diff
@@ -41,28 +41,39 @@
...
+
+        If the image is local, we return its local location. Otherwise,
+        the image is downloaded at the specified download location and
+        we return that one.
+
+        :return: a path to the image
+        """
         log.info("Downloading the image...")
 
+        if self._url.startswith("file://"):
```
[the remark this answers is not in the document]
poncovka: There are already other `download` methods in this class, so I didn't want to introduce new ones. Also, I didn't want to hide the testing for the `file://` prefix into a function since we remove that prefix two lines later. It would be confusing.

pyanaconda/modules/payloads/payload/live_image/installation.py
poncovka: The main purpose of the task is to download an image. I don't want to hide that. It just ends earlier if the image is already "downloaded".

PR #3740, opened 2021-12-03
Fix reset of DBus containers in the unit tests

poncovka:
If we need to reset a DBus container for testing purposes, do it properly.
Otherwise, the container can stay in an inconsistent state and affect other
tests.


VladimirSlavik (approved):

jkonecny12:
@poncovka do we want to have this backported to RHEL-9?

PR #3744, opened 2021-12-07
Remove upd-kernel

VladimirSlavik:
It is not clear why anaconda repo is the place to have it, we did not use it for a long time.

CC @bcl, @jstodola, @t184256 - please yell if this was a thing of importance to you.


jstodola (approved):
jstodola has never used this script, I doubt it's working (the kernel modules are compressed now and ending with `.ko.xz`. Also not updating the kernel modules in `install.img` is often a source of troubles. Better use lorax for rebuilding the installation images.

poncovka (approved):
Thank you!

bcl:
I haven't used it in forever. IIRC it was more useful back when we had loader to debug.

PR #3746, opened 2021-12-08
Add ShellCheck to unit tests

VladimirSlavik:
- This will automatically find all shell scripts tracked in the repository.
- Files and directories can be ignored according to a configuration file.
- The test is optional: If ShellCheck is not present, test will be skipped.
- This is very lightweight, runs take < 10 seconds on a contemporary machine.
- Currently not-fixed-yet scripts are ignored with this initial configuration.

### DIFF
--- a/dockerfile/anaconda-ci/Dockerfile
+++ b/dockerfile/anaconda-ci/Dockerfile
@@ -44,7 +44,8 @@ RUN set -ex; \
   python3-lxml \
   policycoreutils \
   python3-gobject-base \
-  python3-pip; \
+  python3-pip \
+  ShellCheck; \
   if ! grep -q VARIANT.*eln /etc/os-release; then \
     BRANCH="${git_branch}"; \
     if [ $BRANCH == "master" ]; then \
--- a/tests/shellcheck/ignores.sh
+++ b/tests/shellcheck/ignores.sh
@@ -1,0 +1,8 @@
+#!/bin/bash
+# Paths to ignore in shellcheck tests; these are patterns passed to `grep -v`.
+# shellcheck disable=SC2034  # sourced by the unit test script
+ignore_patterns=(
+dockerfile/
+scripts/
+widgets/
+)
--- a/tests/shellcheck/run_shellcheck.sh
+++ b/tests/shellcheck/run_shellcheck.sh
@@ -1,0 +1,39 @@
+#!/bin/bash
+
+# If $top_srcdir has not been set by automake, detect it
+if [ -z "$top_srcdir" ]; then
+    top_srcdir="$(realpath "$(dirname "$0")/../..")"
+fi
+
+source "${top_srcdir}/tests/shellcheck/ignores.sh"
+
+# Check if this test can be run
+if ! type shellcheck > /dev/null 2>&1 ; then
+    echo "SKIP - shellcheck must be installed to run it."
+    exit 77
+fi
+
+# Identify which shellcheck version is running to debug fails due to version bump
+shellcheck --version
+rpm -q ShellCheck
+echo
+
+# Ignore files according to path patterns
+pushd "${top_srcdir}" > /dev/null || return 1
+files=$(git ls-files)
+# shellcheck disable=SC2154  # comes from the file sourced earlier
+for ignore in "${ignore_patterns[@]}" ; do
+    files=$(echo "$files" | grep -v "$ignore")
+done
+
+# Find files that are shell scripts
+files=$(echo "$files" | xargs file | grep "shell script" | awk -F ":" '{print $1}')
+
+# List files being checked
+echo -e "Files to be checked:\n--------------------\n$files\n"
+
+# Do the actual linting
+echo -e "Warnings found:\n---------------"
+if ! echo "$files" | xargs shellcheck -f gcc ; then
+  exit 1
+fi
--- a/.shellcheckrc
+++ b/.shellcheckrc
@@ -1,4 +1,5 @@
 # ShellCheck config file
 
 disable=SC1090  # Can't follow non-constant source. Use a directive to specify location.
+disable=SC1091  # Not following: <some-file> was not specified as input (see shellcheck -x).
 disable=SC2230  # which is non-standard. Use builtin 'command -v' instead.
--- a/dracut/.shellcheckrc
+++ b/dracut/.shellcheckrc
@@ -1,6 +1,6 @@
 # ShellCheck config file
 
 disable=SC1090  # Can't follow non-constant source. Use a directive to specify location.
-disable=SC2230  # which is non-standard. Use builtin 'command -v' instead.
 disable=SC1091  # Not following: <some-file> was not specified as input (see shellcheck -x).
+disable=SC2230  # which is non-standard. Use builtin 'command -v' instead.
 disable=SC2154  # <variable-name> is referenced but not assigned.
--- a/dracut/fetch-driver-net.sh
+++ b/dracut/fetch-driver-net.sh
@@ -45,7 +45,7 @@ for dd in $DD_NET; do
             mntdir=$(nfs_already_mounted "$server" "$path")
             if [ -z "$mntdir" ]; then
                 mntdir="$(mkuniqdir /run nfs_mnt)"
-                mount_nfs "$nfs:$server:$path$(options:+:$options)" "$mntdir"
+                mount_nfs "$nfs:$server:$path$(options:+:"$options")" "$mntdir"
             fi
 
             # Get and process all rpm files in the mounted directory
--- a/dracut/parse-anaconda-options.sh
+++ b/dracut/parse-anaconda-options.sh
@@ -49,7 +49,7 @@ check_depr_arg() {
 }
 check_depr_args() {
     local q=""
-    for i in $(getargs "$1"); do check_depr_arg $q "$i" "$2" && q="--quiet"; done
+    for i in $(getargs "$1"); do check_depr_arg "$q" "$i" "$2" && q="--quiet"; done
 }
 check_removed_arg() {
     local arg="$1"; shift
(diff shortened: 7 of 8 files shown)


tests/shellcheck/run_shellcheck.sh
```diff
@@ -0,0 +1,38 @@
+#!/bin/bash
+
+# If $top_srcdir has not been set by automake, import the test environment
+if [ -z "$top_srcdir" ]; then
+    top_srcdir="$(realpath "$(dirname "$0")/../..")"
+    . "${top_srcdir}"/tests/testenv.sh
```
poncovka, line 6: Do we really want to source this file? Isn't it going to affect the results of the check?

tests/shellcheck/ignores.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+# Paths to ignore in shellcheck tests; these are patterns passed to `grep -v`.
```
poncovka, line 2: Have you considered to use `shellcheck disable=` in these files?

tests/shellcheck/run_shellcheck.sh
```diff
@@ -0,0 +1,38 @@
...
+done
+
+# Find files that are shell scripts
+files=$(echo "$files" | xargs file | grep "shell script" | awk -F ":" '{print $1}')
+
+# List files being checked
+echo -e "Files to be checked:\n--------------------\n$files\n" | tee -a "$logfile"
+
+# Do the actual linting
+echo -e "Check results:\n--------------" | tee -a "$logfile"
```
poncovka, line 37: The check is not hooked to workflows, right? Could you copy-paste here the current output, please?

dockerfile/anaconda-ci/Dockerfile
```diff
@@ -44,7 +44,8 @@ RUN set -ex; \
   python3-lxml \
   policycoreutils \
   python3-gobject-base \
-  python3-pip; \
+  python3-pip \
+  ShellCheck; \
```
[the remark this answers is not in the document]
poncovka: I wouldn't worry about that now.

poncovka (approved):
Thanks!

tests/shellcheck/run_shellcheck.sh
VladimirSlavik: I thought this is a standard thing to do? The other tests appear to work that way. That said, the file does not *seem* to contain anything relevant. Right now, at least.

tests/shellcheck/run_shellcheck.sh
VladimirSlavik: Sure!

<details>
<summary>Logs</summary>

```
ShellCheck - shell script analysis tool
version: 0.7.0
license: GNU General Public License, version 3
website: https://www.shellcheck.net

Files to be checked:
--------------------
autogen.sh
data/liveinst/liveinst
data/liveinst/liveinst-setup.sh
data/systemd/anaconda-generator
dracut/anaconda-copy-cmdline.sh
dracut/anaconda-copy-dhclient.sh
dracut/anaconda-copy-ks.sh
dracut/anaconda-copy-prefixdevname.sh
dracut/anaconda-copy-s390ccwconf.sh
dracut/anaconda-depmod.sh
dracut/anaconda-diskroot
dracut/anaconda-error-reporting.sh
dracut/anaconda-hmcroot
dracut/anaconda-ifcfg.sh
dracut/anaconda-ifdown
dracut/anaconda-ks-sendheaders.sh
dracut/anaconda-lib.sh
dracut/anaconda-modprobe.sh
dracut/anaconda-netroot.sh
dracut/anaconda-nfsrepo-cleanup.sh
dracut/anaconda-pre-shutdown.sh
dracut/anaconda-set-kernel-hung-timeout.sh
dracut/driver-updates-genrules.sh
dracut/fetch-driver-net.sh
dracut/fetch-kickstart-disk
dracut/fetch-kickstart-net.sh
dracut/fetch-updates-disk
dracut/find-net-intfs-by-driver
dracut/kickstart-genrules.sh
dracut/module-setup.sh
dracut/parse-anaconda-dd.sh
dracut/parse-anaconda-kickstart.sh
dracut/parse-anaconda-net.sh
dracut/parse-anaconda-options.sh
dracut/parse-anaconda-repo.sh
dracut/repo-genrules.sh
dracut/save-initramfs.sh
dracut/updates-genrules.sh
tests/cppcheck/runcppcheck.sh
tests/gettext_tests/canary_tests.sh
tests/glade_tests/glade_tests.sh
tests/lib/testlib.sh
tests/pep8/runpep8.sh
tests/rpm_tests/rpm_tests.sh
tests/shellcheck/ignores.sh
tests/shellcheck/run_shellcheck.sh
tests/testenv.sh
tests/unit_tests/unit_tests.sh

Check results:
--------------
```
</details>

tests/shellcheck/run_shellcheck.sh
poncovka: The file helps to overcome the fact that we run tests on the source code and not on the installed code, so the files can be at different locations than expected. I don't think that any linter cares about that. It doesn't look like we use this file for running `pylint`.

VladimirSlavik:
Needs changes due to  #3751

tests/shellcheck/run_shellcheck.sh
VladimirSlavik: You're right, that seems to be redundant. Removed.

tests/shellcheck/run_shellcheck.sh
VladimirSlavik: Is it hooked to workflows? Indirectly. It runs with unit tests, same as glade tests and the others.

VladimirSlavik:
Rebased, removed sourcing of `testenv.sh`, updated to match translation-canary now passing.

tests/shellcheck/run_shellcheck.sh
poncovka: Oh, I have missed that, sorry. However, it looks like you don't propagate the exit status, because `shellcheck` found some issues, but the test passed.

```
Check results:
--------------
data/liveinst/liveinst-setup.sh:13:66: note: Not following: ./user-dirs.dirs was not specified as input (see shellcheck -x). [SC1091]
dracut/fetch-driver-net.sh:48:58: note: Double quote to prevent globbing and word splitting. [SC2086]
dracut/parse-anaconda-options.sh:52:49: note: Double quote to prevent globbing and word splitting. [SC2086]
tests/pep8/runpep8.sh:9:3: note: Not following: ./tests/pep8/config.target was not specified as input (see shellcheck -x). [SC1091]
PASS shellcheck/run_shellcheck.sh (exit status: 0)
```

VladimirSlavik: Ouch, thanks. I thought it would somehow propagate out from the piping. Back to that, I guess :(

M4rtinK (approved):
Looks good to me. :)

VladimirSlavik:
Errors don't propagate outside the "pipeline" so the test never fails, see discussion above. Needs fixing.

VladimirSlavik:
**Update**: Test now fails if there are warnings / errors, so it works correctly.

There are two more things to consider:

- How do we want to handle warnings that appear with a version update? I started using 0.7.0 but rawhide now has 0.7.2 which found a few more things. I assume this will behave like pylint, sometimes new version makes things fail.

- How do we want to manage the `.shellcheckrc` files? They are applied recursively, top-down, from current directory. That means we can have a global configuration and also tune detections in different directories. But it also means this is sensitive to where the tool / script / test runs. I propose moving all warnings to the top-level file, and considering only the unit test results. In practice that means adding a commit that merges the dracut config into the global one and forgetting about this...

jkonecny12 (approved):
Looks great to me. Thanks!

jkonecny12:
Please also add `backport rhel-9` flag.

tests/shellcheck/run_shellcheck.sh
```diff
@@ -0,0 +1,46 @@
+#!/bin/bash
+
+# If $top_srcdir has not been set by automake, detect it
+if [ -z "$top_srcdir" ]; then
+    top_srcdir="$(realpath "$(dirname "$0")/../..")"
+fi
+
+source "${top_srcdir}/tests/shellcheck/ignores.sh"
+logfile="${top_srcdir}/tests/shellcheck/shellcheck.log"
```
poncovka, line 9: I have noticed that we create the `run_shellcheck.log` and `shellcheck.log` files with pretty much the same content. What is the reasoning behind it?

tests/shellcheck/run_shellcheck.sh
```diff
@@ -0,0 +1,46 @@
...
+
+# Find files that are shell scripts
+files=$(echo "$files" | xargs file | grep "shell script" | awk -F ":" '{print $1}')
+
+# List files being checked
+echo -e "Files to be checked:\n--------------------\n$files\n" | tee -a "$logfile"
+
+# Do the actual linting
+echo -e "Warnings found:\n---------------" | tee -a "$logfile"
+echo "$files" | xargs shellcheck -f gcc | tee -a "$logfile"
```
poncovka, line 38: It looks like `shellcheck` returns useful exit codes. Couldn't we just return any nonzero code returned by `shellcheck` instead of parsing results below?

poncovka (approved):
I have a few non-blocking questions. Otherwise, it looks great. Thanks!

VladimirSlavik:
> Please also add `backport rhel-9` flag.

I am not sure if there is shellcheck on RHEL. IIRC it's in EPEL which is not there for RHEL 9. However the test won't fail, so perhaps it still makes sense to port this.

tests/shellcheck/run_shellcheck.sh
VladimirSlavik: I had no idea the output is saved to file. If I can remove the logging, that makes the script more simple.

tests/shellcheck/run_shellcheck.sh
VladimirSlavik: `tee` eats the return code. But If I can remove it - see above - then no need for the parsing...

PR #3747, opened 2021-12-08
Remove the base_repo property

poncovka:
We don't actually need to know the name of a base repository.

### DIFF
--- a/pyanaconda/payload/dnf/payload.py
+++ b/pyanaconda/payload/dnf/payload.py
@@ -204,11 +204,39 @@ class DNFPayload(Payload):
 
     def is_ready(self):
         """Is the payload ready?"""
-        return self.base_repo is not None
+        enabled_repos = self._dnf_manager.enabled_repositories
+
+        # If CDN is used as the installation source and we have
+        # a subscription attached then any of the enabled repos
+        # should be fine as the base repo.
+        # If CDN is used but subscription has not been attached
+        # there will be no redhat.repo file to parse and we
+        # don't need to do anything.
+        if self.source_type == SOURCE_TYPE_CDN:
+            return self._is_cdn_set_up() and enabled_repos
+
+        # Otherwise, a base repository has to be enabled.
+        return any(map(self._is_base_repo, enabled_repos))
+
+    def _is_cdn_set_up(self):
+        """Is the CDN source set up?"""
+        if not self.source_type == SOURCE_TYPE_CDN:
+            return False
+
+        if not is_module_available(SUBSCRIPTION):
+            return False
+
+        subscription_proxy = SUBSCRIPTION.get_proxy()
+        return subscription_proxy.IsSubscriptionAttached
+
+    def _is_base_repo(self, repo_id):
+        """Is it a base repository?"""
+        return repo_id == constants.BASE_REPO_NAME \
+            or repo_id in constants.DEFAULT_REPOS
 
     def is_complete(self):
         """Is the payload complete?"""
-        return self.source_type not in SOURCE_REPO_FILE_TYPES or self.base_repo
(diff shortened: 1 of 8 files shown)


tests/unit_tests/pyanaconda_tests/modules/payloads/payload/test_module_payload_dnf_manager.py
```diff
@@ -927,6 +927,24 @@ def test_repositories(self):
...
+    def test_enabled_repositories(self):
+        """Test the enabled_repositories property."""
+        assert self.dnf_manager.enabled_repositories == []
+
+        r1 = self._add_repo("r1")
+        r2 = self._add_repo("r2")
+        r3 = self._add_repo("r3")
+        r4 = self._add_repo("r4")
+
+        assert self.dnf_manager.repositories == ["r1", "r2", "r3", "r4"]
```
VladimirSlavik, line 939: Do you perhaps also want to test the initial state ({dis|en}abled) after adding?

VladimirSlavik (approved):
Looks good to me, thank you!

I don't know the CDN parts too well, but the diff seems to make sense alone, so I didn't dig into that further.

tests/unit_tests/pyanaconda_tests/modules/payloads/payload/test_module_payload_dnf_manager.py
poncovka: Do you mean the initial state of the added repositories? Doesn't this check test their state?

pyanaconda/payload/dnf/payload.py
```diff
@@ -204,11 +204,39 @@ def set_packages_selection(self, data: PackagesSelectionData):
...
+        # If CDN is used but subscription has not been attached
+        # there will be no redhat.repo file to parse and we
+        # don't need to do anything.
+        if self.source_type == SOURCE_TYPE_CDN:
+            return self._is_cdn_set_up() and enabled_repos
+
+        # Otherwise, a base repository has to be enabled.
+        return any(map(self._is_base_repo, enabled_repos))
+
+    def _is_cdn_set_up(self):
```
poncovka, line 221: @M4rtinK, could you review these changes, please?

tests/unit_tests/pyanaconda_tests/modules/payloads/payload/test_module_payload_dnf_manager.py
VladimirSlavik: Ah, sorry, I left out too much, again :( So. You do this:
1. Add repos
2. Test repo presence
3. Set enabled/disabled
4. Test enabled/disabled

IMO from the test's point of view, the enabled/disabled state after adding is undefined. So I'm asking if this makes sense instead:
1. Add repos
2. Test repo presence
3. *Test enabled/disabled*
4. Set enabled/disabled
5. Test enabled/disabled

poncovka: I see. Actually, I didn't like that test, because it tested DNF instead of Anaconda, so I have removed it.

pyanaconda/payload/dnf/payload.py
M4rtinK: Yep, this shduld work correctly as far as I can tell from looking at the code.

Unfortunately we can't really test this in a realistic scenario on Fedora just yet, though I have some ideas in the longer term. :)

M4rtinK (approved):
Looks good to me - nice cleanup & thanks! :)

Regarding subscription support - already when I first started working on subscription support in Anaconda it seemed to me that the concept of a base repo is rather unnecessary in many case if not always, so I'm glad to see it cleaned up. :)

jkonecny12 (approved):
LGTM!

PR #3750, opened 2021-12-09
network: always use rd.iscsi.ibft when the need to access an iBFT device

rvykydal:
Previously, the installer would add rd.iscsi.ibft only if network during
install was configured from iBFT and not set any other network
configuration options to kernel command line.

That is not correct. The installation could've been started from a CD
media and still be done to an iSCSI volume discovered via iBFT.
Given presence of no other network configuration, the rd.iscsi.firmware
will cause dracut to still bring up networking automatically (as if
rd.needned has been passed).

That might work for most installations, but is probably not the right thing
to do: all interfaces will be brought up needlessly and things would break
in less straightforward setups, such as the iSCSI volume accessed via VLAN.

Let's just use rd.iscsi.ibft unconditionally if there's an iSCSI volume
configured via iBFT.

This removes the ibft argument from get_dracut_arguments_from_connection()
because it's no longer used and thus drops the tests for it. One
situation where we are changing the behavior compared to what the tests
check is when iBFT is used along with s390 subchannels. I believe the
test was wrong: s390 doesn't use ACPI and if it did, it would make more
sense to extend the iBFT with the subchannel configuration.


poncovka (approved):
Thanks!

VladimirSlavik (approved):
Thank you!

pyanaconda/modules/network/network.py
```diff
@@ -686,24 +685,23 @@ def get_dracut_arguments(self, iface, target_ip, hostname, ibft):
...
-            connection,
-            iface,
-            target_ip,
-            hostname,
-            ibft
-        ))
+            if target_connections:
+                if len(target_connections) > 1:
+                    log.debug("Get dracut arguments: "
+                              "multiple connections found for traget %s: %s, taking the first one",
```
jkonecny12, line 691: I know this is merged but we have a typo here which should be fixed :).

 `traget` -> `target`

I know I'm such a tragéd ;).

jkonecny12: https://github.com/rhinstaller/anaconda/pull/3760

PR #3751, opened 2021-12-09
Fix ShellCheck issues in translation_canary

VladimirSlavik:
Needs also porting to the other repo with canary.


poncovka (approved):
Thanks!

PR #3753, opened 2021-12-10
Move the code for adding repositories to the DNF base

poncovka:
Call the `convert_ks_repo_to_repo_data` and `convert_repo_data_to_ks_repo` functions
to convert a kickstart repo object into its DBus representation and back.

Call the `add_repository` method of the DNF manager to add a repository to DNF.

### DIFF
--- a/pyanaconda/modules/payloads/kickstart.py
+++ b/pyanaconda/modules/payloads/kickstart.py
@@ -17,6 +17,9 @@
 # License and may only be used or replicated with the express permission of
 # Red Hat, Inc.
 #
+from pyanaconda.core.constants import URL_TYPE_BASEURL, URL_TYPE_MIRRORLIST, URL_TYPE_METALINK, \
+    DNF_DEFAULT_REPO_COST
+from pyanaconda.modules.common.structures.payload import RepoConfigurationData
 from pykickstart.errors import KickstartParseError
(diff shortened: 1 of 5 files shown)


jkonecny12 (commented):
In general looks good to me. However, I would rather see the first commit in the Clean architecture design.

Basically, have object which has the data in a unified format (just blackbox) and have `read_ks_data`, `write_ks_data` and similar function as interface for these data.

However, I understand if you don't want to complicate things with that, it would just be "nicer".

M4rtinK (approved):
Looks good to me. :)

VladimirSlavik (approved):
Looks good to me. Thank you!

poncovka:
> In general looks good to me. However, I would rather see the first commit in the Clean architecture design.
> 
> Basically, have object which has the data in a unified format (just blackbox) and have `read_ks_data`, `write_ks_data` and similar function as interface for these data.
> 
> However, I understand if you don't want to complicate things with that, it would just be "nicer".

This is a temporary solution. The kickstart repo data will be eventually processed by the DNF payload module and we will keep only the repo data objects. However, there is too much code that would have to be changed at once, so I am doing it this way to make the changes smaller.

poncovka:
Rebased and tested.

PR #3754, opened 2021-12-11
Use local spec file for container build

jkonecny12:
Some of our containers needs dependencies of Anaconda installed in it and to get these we were downloading anaconda.spec.in from the rhinstaller/anaconda repository. This works, but it has a big drawbacks like not being able to use this correctly from a fork for testing or do not test spec file changes in PRs.

Spec file couldn't be read from the repository because podman/docker do allow only resources under the build directory.

To solve this, the Makefile will copy the anaconda.spec.in file to the dockerfile(s) directory before the build, so we don't have to care about this.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -174,6 +174,9 @@ release-and-tag:
 	$(MAKE) tag
 
 anaconda-ci-build:
+	TEMP=$$(mktemp -t -d anaconda-ci-build.XXXX) && \
+	cp $(srcdir)/anaconda.spec.in $(CI_DOCKERFILE)/* $$TEMP/ && \
+	echo "Build dir is $$TEMP" && \
 	$(CONTAINER_ENGINE) build \
 	$(CONTAINER_BUILD_ARGS) \
 	$(CONTAINER_ADD_ARGS) \
@@ -181,9 +184,13 @@ anaconda-ci-build:
 	--build-arg=image=$(BASE_CONTAINER) \
 	--build-arg=copr_repo=$(COPR_REPO) \
 	-t $(CI_NAME):$(CI_TAG) \
-	$(CI_DOCKERFILE)
+	$$TEMP/ && \
+	rm -rf $$TEMP
 
 anaconda-rpm-build:
+	TEMP=$$(mktemp -t -d anaconda-rpm-build.XXXX) && \
+	cp $(srcdir)/anaconda.spec.in $(CI_DOCKERFILE)/* $$TEMP/ && \
+	echo "Build dir is $$TEMP" && \
 	$(CONTAINER_ENGINE) build \
 	$(CONTAINER_BUILD_ARGS) \
 	$(CONTAINER_ADD_ARGS) \
@@ -191,7 +198,8 @@ anaconda-rpm-build:
 	--build-arg=image=$(BASE_CONTAINER) \
 	--build-arg=copr_repo=$(COPR_REPO) \
 	-t $(RPM_NAME):$(CI_TAG) \
-	$(RPM_DOCKERFILE)
+	$$TEMP/ && \
+	rm -rf $$TEMP
 
 anaconda-iso-creator-build:
 	$(CONTAINER_ENGINE) build \
--- a/dockerfile/anaconda-ci/Dockerfile
+++ b/dockerfile/anaconda-ci/Dockerfile
@@ -27,13 +27,16 @@ LABEL maintainer=anaconda-devel@lists.fedoraproject.org
 # On ELN, BaseOS+AppStream don't have all our build dependencies; this provides the "Everything" compose
 COPY ["eln.repo", "/etc/yum.repos.d"]
 
+# The anaconda.spec.in is in the repository root. This file will be copied automatically here if
+# the build is invoked by Makefile.
+COPY ["anaconda.spec.in", "/root/"]
+
 # Prepare environment and install build dependencies
 RUN set -ex; \
   if grep -q VARIANT.*eln /etc/os-release; then sed -i 's/enabled=0/enabled=1/' /etc/yum.repos.d/eln.repo; fi; \
   dnf update -y; \
   dnf install -y \
   'dnf-command(copr)' \
-  curl \
   /usr/bin/xargs \
   rpm-build \
   git \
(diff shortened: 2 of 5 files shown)


VladimirSlavik (approved):
Looks good to me, thank you!

dockerfile/anaconda-ci/Dockerfile
```diff
@@ -58,7 +64,7 @@ RUN set -ex; \
   else \
     dnf copr enable -y ${copr_repo} fedora-eln-x86_64; \
   fi; \
-  curl -L https://raw.githubusercontent.com/rhinstaller/anaconda/${git_branch}/anaconda.spec.in | sed 's/@PACKAGE_VERSION@/0/; s/@PACKAGE_RELEASE@/0/; s/%{__python3}/python3/' > /tmp/anaconda.spec; \
```
poncovka, line 61: I would like to know why we used `curl` in the first place. Isn't there a use case that we are overlooking?

dockerfile/anaconda-rpm/Dockerfile
```diff
@@ -22,13 +25,16 @@ LABEL maintainer=anaconda-devel@lists.fedoraproject.org
 # On ELN, BaseOS+AppStream don't have all our build dependencies; this provides the "Everything" compose
 COPY ["eln.repo", "/etc/yum.repos.d"]
 
+# The anaconda.spec.in is in the repository root. This file will be copied automatically here if
+# the build is invoked by Makefile.
+COPY ["anaconda.spec.in", "/root/"]
```
poncovka, line 30: Why is the file in the `/root` directory? We don't need it after the container build. Couldn't it be in `/tmp`?

Makefile.am
```diff
@@ -174,6 +174,7 @@ release-and-tag:
 	$(MAKE) tag
 
 anaconda-ci-build:
+	cp $(srcdir)/anaconda.spec.in $(CI_DOCKERFILE)/anaconda.spec.in
```
poncovka, line 177: So, I will have multiple copies of the `anaconda.spec.in` file all over my git repo, right? That is just sad. Shouldn't we start to create a temporary directory for the build context? Or could we use a build argument to propagate the content of the spec file?

poncovka (commented):
I don't like solutions that require to create weird file at weird places.

dockerfile/anaconda-ci/Dockerfile
VladimirSlavik: tl;dr I don't think so, fortunately. The containers changed only incrementally since inception, and the use case is no longer valid.

The containers were originally written from the perspective of Github actions. That is, they would be built in a clean environment, with no context except for "for branch X". We feed all the important bits to them via arguments. These arguments bubble down from the action that has the actual config, via makefile, and only then to podman/docker. There was also the part that we would be building them daily, and reuse, so there was no context of PR, just the repo state. So straight `curl thing-from-repo` was good enough.

I also think that as we then went to use them on PRs, and needed to rebuild them if needed, and given that [back then] the tests would run always, I guess we decided to use as much from the repo as possible to minimize hackable surface. That is no longer true, all container builds directly on tests are behind the button to let them run, which is supposed to be clicked only after review of the changes.

dockerfile/anaconda-rpm/Dockerfile
VladimirSlavik: That is `/root` in the container, so it does not matter... much...

Makefile.am
VladimirSlavik: Maybe we can just remove the spec file after building, a few lines below.

dockerfile/anaconda-ci/Dockerfile
jkonecny12: I wasn't able to find any use of the curl other than this one. I run our tests locally and builds locally and both worked as expected without the curl.

dockerfile/anaconda-rpm/Dockerfile
jkonecny12: Yep, it's just _some_ folder. We can go with `/tmp` but the problem is that `COPY` is happening in a separate layer than the `RUN` where it's used. I don't want to use `/tmp` because I'm worried that sometimes (maybe even here - not tested) the `/tmp` could be tmpfs and then the content could be missing for the `RUN` phase. I would rather like to avoid using `/tmp/` during container builds if it's not used during the same `RUN` phase.

Makefile.am
jkonecny12: > So, I will have multiple copies of the `anaconda.spec.in` file all over my git repo, right? That is just sad. 

The file is in `.gitignore` so shouldn't really have to care. However, yes, we can definitely remove the file after the build finished. That is pretty easy to do, if you want.
Just a note that until recently we had a lot of files in our repo directory because of autogen and configure calls.

> Shouldn't we start to create a temporary directory for the build context?
I really don't want to complicate the Makefile by creating a temp dir for a build and copying everything there. It will be fragile and much more complex than what  we have now.
 
> Or could we use a build argument to propagate the content of the spec file?
Nope, the spec file is too big to pass the content as argument. We would probably get `line too long` from shell and it's just a wrong way of doing things in general.

poncovka: > > So, I will have multiple copies of the `anaconda.spec.in` file all over my git repo, right? That is just sad.
> 
> The file is in `.gitignore` so shouldn't really have to care. However, yes, we can definitely remove the file after the build finished. That is pretty easy to do, if you want.

I very often search my repository. It will appear there. The version of these files doesn't even have to match the branch, that I am using, so it should be at least removed.

> Just a note that until recently we had a lot of files in our repo directory because of autogen and configure calls.

Yes and I am very happy with the current state, so why do we want to change that?

> 
> > Shouldn't we start to create a temporary directory for the build context?
>
> I really don't want to complicate the Makefile by creating a temp dir for a build and copying everything there. It will be fragile and much more complex than what  we have now.

It would be cleaner and not so confusing as the current solution.

poncovka: > Honestly, I think creating the tmp dir and putting everything there would be more confusing. First it's making our Makefile bigger and harder to orient and second it will be harder to find the issue if something went wrong.

You need to copy there only the content of the `$(CI_DOCKERFILE)` directory and the spec file.

> So do I understand it correctly that you are fine with solution when the spec file will be removed after build?

Yes, it is ugly, but it will work.

VladimirSlavik: For the record - I would be happier with a better solution, too, but I would take this too. If I understand this correctly, the problem being solved affects 2/6 team members, potentially for the next month or so, so that makes me feel something should happen.

jkonecny12:
**UPDATED:**
* After a discussion with @poncovka I tried to change the solution to use temp dir for a build and it looks to be working quite nice. So, why not, it's not that complicated as I thought it will be.

poncovka (approved):
Thank you!

M4rtinK (approved):
Looks good to me, thanks! :)

PR #3756, opened 2021-12-13
Move service util functions to new module

VladimirSlavik:
Mostly SSIA. CLeans up the `util` module a bit.

The handling of roots might seem to be a mess at a glance. However there are really two or three kinds of functions, so defaults must differ, and from there on it snowballs.


poncovka (commented):
I have some suggestions.

poncovka:
Btw. could you write some unit tests for these functions, please?

pyanaconda/modules/timezone/installation.py
```diff
@@ -19,7 +19,7 @@
 import os.path
 
 from pyanaconda import ntp
-from pyanaconda.core import util
+from pyanaconda.core import service
```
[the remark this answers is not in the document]
VladimirSlavik: I will look at this, however this will need more changes and I will want to make it at least a separate commit. The import style is heavily exploited by the test, it mocks the whole module and passes the mock around.

pyanaconda/core/service.py
```diff
@@ -0,0 +1,150 @@
...
+
+    :param str service: name of the service to check
+    :return: was the service found
+    """
+    ret = _run_systemctl("status", service)
+
+    return ret == 0
+
+
+def is_service_installed(service, root=None):
```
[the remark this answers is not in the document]
VladimirSlavik: Actually, I think it might be good to require passing some root everywhere. Having a default means we can bikeshed about which will it be.

pyanaconda/core/service.py
```diff
@@ -0,0 +1,139 @@
...
+
+    :param str service: name of the service to check
+    :return: was the service found
+    """
+    ret = _run_systemctl("status", service, "/")
+
+    return ret == 0
+
+
+def is_service_installed(service, root):
```
poncovka, line 91: I was considering to make the `root` argument mandatory, but when you combine the functions with and without the `root` argument, it is just confusing and looks like a bug in the code. 

```
if is_service_installed(RHSM_SERVICE_NAME, "/"):
  start_service(RHSM_SERVICE_NAME)
```

I still think that using `root="/"` by default will more consistent with our libraries.

VladimirSlavik:
From off-site discussion: Let's make the default `root="/"` so that we do not have to specify a "local" action. That is consistent with other functions that operate on installation environment by default. It is a 180° from previous defaults, but it is a design that considers and unifies the whole module, whereas the previous state was likely just a result of isolated incremental extensions. By keeping a default, `root` also stays a kwarg.

poncovka (approved):
Thank you!

tests/unit_tests/pyanaconda_tests/modules/timezone/test_module_timezone.py
```diff
@@ -521,61 +521,49 @@ def _get_test_sources(self):
...
-        else:
-            service_util.enable_service.assert_not_called()
-            service_util.disable_service.assert_called_once_with(
-                "chronyd", root=sysroot
-            )
+    def _test_task(self, sysroot, ntp_enabled, ntp_installed, ntp_servers):
+        PREFIX = "pyanaconda.modules.timezone.installation"
+        with patch(PREFIX + ".is_service_installed") as mock_is_installed:
+            with patch(PREFIX + ".enable_service") as mock_enable:
+                with patch(PREFIX + ".disable_service") as mock_disable:
```
jkonecny12, line 548: Not sure if there is a better way to do this but this is a kind of ugly :(.

jkonecny12 (approved):
Looks good to me except the test case but I won't block on that because I don't have replacement.

tests/unit_tests/pyanaconda_tests/modules/timezone/test_module_timezone.py
VladimirSlavik: Oh, it is absolutely ugly. We discussed this part with Vendy and there are essentially three options here:

- Write a custom context manager that wraps the three patches. However that only moves the problem around, at a cost I don't want to pay.
- Give up and drop the last commit. The `module.function` call style remains.
- Nuke it from orbit and rewrite the whole test case. However at that scope I think that no longer belongs to this PR.

jkonecny12: I like the part with a nuke but I guess that is for another millennium :).

poncovka: We can drop the commit for now and fix the issue in another PR.

VladimirSlavik: Let's do that!

PR #3757, opened 2021-12-13
Remove the support for image installations from the liveinst script

poncovka:
We do not test, use or support this use case.


VladimirSlavik (approved):
Thank you!

M4rtinK (approved):

PR #3759, opened 2021-12-15
Move path util functions to new module

VladimirSlavik:
I will likely hammer away at the individual functions for some time. Each move triggers a ton of import changes, and unfortunately many times it's just `util` imported, so the IDE has a hard time doing that usefully. Tests also mock this, which is not always automatically fixed. Finally, not all of the moved functions have good coverage.


pyanaconda/core/path.py
```diff
@@ -0,0 +1,154 @@
...
+            for file_ent in (os.path.join(dir_ent, f) for f in file_items):
+                try:
+                    func(file_ent)
+                except OSError:
+                    pass
+
+        # directories under the directory entry will appear as directory entries in the loop
+
+
+def chown_dir_tree(root, uid, gid, from_uid_only=None, from_gid_only=None):
```
poncovka, line 113: I would move the `chown_dir_tree` and `dir_tree_map` functions to `pyanaconda.core.users` and make them private. Please, add `FIXME: Use 'chown -hR' instead of this function.` to their docstings.

pyanaconda/core/path.py
```diff
@@ -0,0 +1,154 @@
...
+        mkdirChain(sysroot)
+
+    # Mount the mount point.
+    rc = execWithRedirect("mount", ["--rbind", path, sysroot])
+
+    if rc != 0:
+        raise OSError("Failed to mount sysroot to {}.".format(path))
+
+
+def mkdirChain(directory):
```
poncovka, line 66: Please, don't forget to rename the function. What about `make_directories`?

pyanaconda/core/path.py
```diff
@@ -14,3 +14,50 @@
...
+    The path defined by conf.target.system_root will be bind mounted at the given
+    path, so conf.target.system_root can be used to access the root of the new OS.
+
+    We always call it after the root device is mounted at conf.target.physical_root
+    to set the physical root as the current system root.
+
+    Then, it can be used by Payload subclasses which install operating systems to
+    non-default roots.
+
+    If the given path is None, then conf.target.system_root is only unmounted.
```
M4rtinK, line 33: Shouldn't this be "is no mounted" or do I understand it wrong ? :confused:

pyanaconda/core/path.py
M4rtinK: Agreed, its a perfect opportunity to switch to a proper function name with underscores & `make_directories` sounds good to me as well. :)

pyanaconda/core/path.py
```diff
@@ -147,3 +147,8 @@ def conditional_chown(path, uid, gid, from_uid=None, from_gid=None):
         dir_tree_map(root, lambda path: conditional_chown(path, uid, gid,
                                                           from_uid_only,
                                                           from_gid_only))
+
+
+def parent_dir(directory):
```
poncovka, line 152: What about `get_parent_directory`?

pyanaconda/core/path.py
```diff
@@ -71,3 +71,40 @@ def mkdirChain(directory):
     """
 
     os.makedirs(directory, 0o755, exist_ok=True)
+
+
+def dir_tree_map(root, func, files=True, dirs=True):
+    """Apply a function to all paths under a directory.
+
+    Apply the given function to all files and directories in the directory tree
```
M4rtinK, line 79: Will the function be also applied on the root directory as well ? I think it would be good to specify that in this docstring to avoid later confusion.

M4rtinK (approved):
Looks good to me! Just added some minor improvement suggestions. :)

pyanaconda/core/path.py
```diff
@@ -147,3 +147,8 @@ def conditional_chown(path, uid, gid, from_uid=None, from_gid=None):
         dir_tree_map(root, lambda path: conditional_chown(path, uid, gid,
                                                           from_uid_only,
                                                           from_gid_only))
+
+
+def parent_dir(directory):
+    """Return the parent's path"""
+    return "/".join(os.path.normpath(directory).split("/")[:-1])
```
poncovka, line 154: Could you please try to replace the implementation with one of the solutions in https://stackoverflow.com/questions/2860153/how-do-i-get-the-parent-directory-in-python?

VladimirSlavik:
Thank you for reviews, I'm still trying to move through the whole list, so I will address the concerns after I have seen what all of the moves require.

pyanaconda/core/path.py
VladimirSlavik: +1 on rename, will do so.

pyanaconda/core/path.py
VladimirSlavik: It uses `os.walk`, so no. Will add.

pyanaconda/core/path.py
VladimirSlavik: I am not sure about the FIXME, `chown` will not use the logic that is in the function. However the rest seems a good idea, although there are more uses. Regardless, this can be trimmed significantly, half code paths are unused...

VladimirSlavik:
Things to do:
- trim down the os.walk wrappers - unused code, can be moved elsewhere, etc.
- re-check docstrings
- check addons and initial setup -> change PRs

pyanaconda/core/path.py
poncovka: About the FIXME, I was googling how to change the owner of a home directory and this is what I have found. I didn't have time to investigate why are our functions are so complicated and the commit messages are not very helpful. Feel free to reword the text, but please keep there the command as a hint.

pyanaconda/core/path.py
VladimirSlavik: This is used only in users so the implementation does not really matter as much. There are lots of rules about the parent, but it's overspecified for no reason.

VladimirSlavik:

One interactive rebase hell later, the moves and renames up to now are squashed. All of the above still TBD.

pyanaconda/core/path.py
VladimirSlavik: I tried, see non-review comment at the end.

VladimirSlavik:
I tried to address some of the feedback in the last three commits:
* Merge _dir_tree_map into _chown_dir_tree
* Remove _get_parent_directory
* Merge helper functions into users._reown_homedir() 

These stray quite far away fro mthe original idea, but this was supposed to be "cleanup" anyway.

pyanaconda/core/path.py
```diff
@@ -0,0 +1,132 @@
...
+
+    # Mount the mount point.
+    rc = execWithRedirect("mount", ["--rbind", path, sysroot])
+
+    if rc != 0:
+        raise OSError("Failed to mount sysroot to {}.".format(path))
+
+
+def make_directories(directory):
+    """ Make a directory and all of its parents. Don't fail if part or
```
poncovka, line 67: Could you fix the indentation of docstrings, please?

poncovka (approved):
It looks great! Thanks!

VladimirSlavik:
Update: Fixed the docstrings, added tests for the new home dir re-owning code.

VladimirSlavik:
Now the test really works.

I could not find any use of the changed functions outside of anaconda. Given that many of the other projects have their own "utils" collecting all kinds of things, it might be true.

poncovka (approved):
It looks great. Thanks!

PR #3762, opened 2021-12-16
Create downstream release changelog from GitHub releases (#infra)

jkonecny12:
Right now automated releases are created from a commits between the last release and the new one. That works good but if we want to play with formatting and removing something it's better to use GitHub releases than the commits.

### DIFF
--- a/.packit.yml
+++ b/.packit.yml
@@ -1,6 +1,7 @@
 specfile_path: anaconda.spec
 upstream_package_name: anaconda
 upstream_tag_template: anaconda-{version}-1
+copy_upstream_release_description: true
 actions:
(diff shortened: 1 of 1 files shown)


jkonecny12:
This PR is prepared to be merged later. Right now it's blocked because we still don't have nice GitHub releases which we would like to show in the changelog.

M4rtinK (approved):

VladimirSlavik (approved):
Thanks!

The last automated release exploded, so we'd probably want to wait at least one sucecssful release to see if it works...

jkonecny12:
Unblocked, last release looks great!

PR #3763, opened 2021-12-16
Satellite support - port to Fedora, batch nr. 2

M4rtinK:
Add a Python module with helper code for Satellite support &
untit tests for it.

(cherry picked from commit a92f8de965da69dc9603662e65d0c53d123c6d25)
(cherry picked from commit 0f20f62d5d5c40ae61da04335a77e88db5acfee2)
(cherry picked from commit 7ae12116014fbf5e2dee93ecb98eb478b114ffbd)

Related: rhbz#1951709

**NOTE**
* this is the second PR in the Satellite support upstreaming series
* this PR contains the Satellite support utility module and it's unit tests
* previous PR in the series: #3684 
* RHEL 9 Satellite support PR: #3485

### DIFF
--- a/pyanaconda/modules/subscription/satellite.py
+++ b/pyanaconda/modules/subscription/satellite.py
@@ -1,0 +1,166 @@
+#
+# Satellite support purpose library.
+#
+# Copyright (C) 2020 Red Hat, Inc.
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# the GNU General Public License v.2, or (at your option) any later version.
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY expressed or implied, including the implied warranties of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General
+# Public License for more details.  You should have received a copy of the
+# GNU General Public License along with this program; if not, write to the
+# Free Software Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
+# 02110-1301, USA.  Any Red Hat trademarks that are incorporated in the
+# source code or documentation are not subject to the GNU General Public
+# License and may only be used or replicated with the express permission of
+# Red Hat, Inc.
+#
+
+import os
+import tempfile
+
+from requests import RequestException
+
+from pyanaconda.core import constants, util
+from pyanaconda.core.path import make_directories, join_paths
+
+from pyanaconda.core.payload import ProxyString, ProxyStringError
+from pyanaconda.core.configuration.anaconda import conf
+
+from pyanaconda.anaconda_loggers import get_module_logger
+log = get_module_logger(__name__)
+
+# the well-known path of the Satellite instance URL where
+# the provisioning script should be located
+PROVISIONING_SCRIPT_SUB_PATH = "/pub/katello-rhsm-consumer"
+
+
+def download_satellite_provisioning_script(satellite_url, proxy_url=None):
+    """Download provisioning script from a Red Hat Satellite instance.
+
+    Download the provisioning script from a Satellite instance and return
+    it as a string.
+
+    Satellite instances usually have self signed certificates and also some tweaks
+    are usually required in rhsm.conf to connect to a customer run Satellite instance
+    instead of to Hosted Candlepin for subscription purposes.
+
+    Each Satellite instance thus hosts a provisioning script available over plain
+    HTTP that client machines can download and execute. This script has minimal dependencies
+    and provisions the machine to be able to talk to the one given Satellite instance
+    by installing it's self signed certificates and adjusting rhsm.conf.
+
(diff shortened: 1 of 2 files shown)


VladimirSlavik (approved):
Thanks for porting!

pyanaconda/modules/subscription/satellite.py
```diff
@@ -0,0 +1,179 @@
...
+            if result.ok:
+                provisioning_script = result.text
+                result.close()
+                log.debug("subscription: Satellite provisioning script downloaded (%d characters)",
+                          len(provisioning_script))
+                return provisioning_script
+            else:
+                log.debug("subscription: server returned %i code when downloading"
+                          " Satellite provisioning script", result.status_code)
+                result.close()
```
poncovka, line 106: I think the context manager will handle that, so you don't need to explicitly close the result. See: https://2.python-requests.org/en/master/api/#requests.Response.close

pyanaconda/modules/subscription/satellite.py
```diff
@@ -0,0 +1,179 @@
...
+        log.warning("subscription: satellite provisioning script not available")
+        return False
+
+    # now that we have something to run, check where to run it
+    if run_on_target_system:
+        # run in the target system chroot
+        sysroot = conf.target.system_root
+    else:
+        # run in installation environment
+        sysroot = "/"
```
poncovka, line 157: Please, replace the `run_on_target_system=False` argument with `sysroot="/"` and remove this block of code.

pyanaconda/modules/subscription/satellite.py
```diff
@@ -0,0 +1,179 @@
...
+        filename = os.path.basename(tf.name)
+        chroot_script_path = os.path.join("/tmp", filename)
+        # and execute it in the sysroot
+        rc = util.execWithRedirect("bash", argv=[chroot_script_path], root=sysroot)
+        if rc == 0:
+            log.debug("subscription: satellite provisioning script executed successfully")
+            return True
+        else:
+            log.debug("subscription: satellite provisioning script executed with error")
+            return False
```
poncovka, line 179: Please, merge the branches. Just log the result and return `rc == 0`.

tests/unit_tests/pyanaconda_tests/modules/subscription/test_satellite.py
```diff
@@ -0,0 +1,233 @@
...
+            timeout=NETWORK_CONNECTION_TIMEOUT
+        )
+
+    def run_satellite_provisioning_script_no_script_test(self):
+        """Test the run_satellite_provisioning_script function - no script."""
+        # if no script is provided, False should be returned
+        self.assertFalse(run_satellite_provisioning_script(provisioning_script=None))
+
+    @patch('pyanaconda.modules.subscription.satellite.util.mkdirChain')
+    @patch('tempfile.NamedTemporaryFile')
```
poncovka, line 147: You should be able to test `run_satellite_provisioning_script` for real without any mocking. We usually use `with tempfile.TemporaryDirectory() as sysroot:` to create a sysroot for testing.

poncovka (changes requested):
I have some suggestions.

tests/unit_tests/pyanaconda_tests/modules/subscription/test_satellite.py
```diff
@@ -0,0 +1,233 @@
...
+    @patch("pyanaconda.core.util.requests_session")
+    def script_download_http_test(self, get_session):
+        """Test the download_satellite_provisioning_script function - http prefix."""
+        # mock the Python Request session
+        session = get_session.return_value.__enter__.return_value
+        result = session.get.return_value
+        result.ok = True
+        result.text = "foo script text"
+        # run the download method
+        script_text = download_satellite_provisioning_script("http://satellite.example.com")
```
poncovka, line 65: Btw. you can create an additional test that will download a local file using `file://` without mocking the `requests` library. It is useful to confirm that we are using the library correctly.

pyanaconda/modules/subscription/satellite.py
M4rtinK: Sounds good, lets drop that. :)

pyanaconda/modules/subscription/satellite.py
M4rtinK: Removed. :)

pyanaconda/modules/subscription/satellite.py
M4rtinK: Yeah, that is simpler, lets do that. :)

pyanaconda/modules/subscription/satellite.py
```diff
@@ -0,0 +1,179 @@
+#
+# Satellite support purpose library.
+#
+# Copyright (C) 2020 Red Hat, Inc.
```
jkonecny12, line 4: We should probably use 2022 here now ;).

jkonecny12 (commented):
Looks good to me in general it would be just nice to have some polishing. So or so, I don't want to block on these so it's on you if you want to implement these or not. Based on that I'm not giving NAK/ACK.

PR #3764, opened 2021-12-22
Allow manual releases (#infra)

VladimirSlavik:
With this, we can go back in time and run the release process for a tag from the past... after fixing the action in the future!

Fix history and make it just flawless - at a press of a button.

Truly exciting.

However...

Obligatory meme picture

I am not sure what would happen if we tried this for a release too far back, with different `BuildRequires`. Let's not depend on this too much.

Tested here to ensure it would not explode too badly: https://github.com/VladimirSlavik/testing/actions The two green runs are manual and automatic respectively.

Thanks to @M4rtinK for inspiration by testing the previous iteration until it exploded.

### DIFF
--- a/.github/workflows/tag-release.yml
+++ b/.github/workflows/tag-release.yml
@@ -1,11 +1,23 @@
 name: Release from tags
 # Create a GitHub release when a tag is pushed.
 
+# Input can ingest any string, but it will not work with arbitrary hashes. The code expects to get
+# a tag that sits on a release commit, so that commit is automatically skipped. Similarly, it
+# expects that the tag name conforms to a particular scheme.
+
 on:
   push:
     tags:
       # this is a glob, not a regexp
       - 'anaconda-*'
+
+  workflow_dispatch:
+    inputs:
+      tag-ref:
+        # it might be possible to release from arbitrary ref, but only tags work
+        description: Name of the tag to release from, such as "anaconda-33.12-1". Branch names or arbitrary commit hashes are not supported. Make sure to trim whitespace when copying.
+        type: string
+        required: true
 
 permissions:
   contents: write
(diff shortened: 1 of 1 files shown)


VladimirSlavik:
It's all fun and memes until tests fail because our podman is out of **bear tokens**. It's not April 1st, so time to go hunt some bears so I can run the tests, I guess.

.github/workflows/tag-release.yml
```diff
@@ -1,31 +1,54 @@
...
 on:
   push:
     tags:
       # this is a glob, not a regexp
       - 'anaconda-*'
 
+  workflow_dispatch:
+    inputs:
+      tag-ref:
+        description: Tag to release from. Arbitrary hashes are not supported. Make sure to trim whitespace when copying.
```
jkonecny12, line 17: Could you please write here an example?

.github/workflows/tag-release.yml
```diff
@@ -1,31 +1,54 @@
...
+          fi
+
       - name: Extract version from the tag
         id: get_version
         # If the version does not fit the regexp, we'll just fail. TODO maybe abort instead?
         run: |
-          echo "Checking pushed tag ref ${{ github.ref }}"
-          VER=$(echo ${{ github.ref }} | perl -ne '/^refs\/tags\/anaconda-([0-9]+(?:\.[0-9]+){1,3})-1$/ && print "$1\n";')
+          echo "Checking ref ${{ steps.get_ref.outputs.ref }}"
+          VER=$(echo ${{ steps.get_ref.outputs.ref }} | perl -ne '/^(?:refs\/tags\/)?anaconda-([0-9]+(?:\.[0-9]+){1,3})-1$/ && print "$1\n";')
```
jkonecny12, line 46: Comment with what we are trying to get from what would be pretty helpful here.

jkonecny12 (changes requested):
Looks good to me but some better explanation in the code would be great.

VladimirSlavik:
Update: Tried to address @jkonecny12's feedback above.

jkonecny12 (approved):
LGTM!

.github/workflows/tag-release.yml
M4rtinK: Agreed, I think this is rather information dense and a commit explaining how it works/what it does will help. :)

M4rtinK (approved):
Thanks - looks good to me as well & should help use to make Anaconda releases more robust and flexible. :)

Just a small suggestion to add some docstrings to the info dense part of this. :)

.github/workflows/tag-release.yml
VladimirSlavik: I'm sorry, I do not understand what is the missing information so I can't add this comment. I guess I know this too well?

PR #3765, opened 2022-01-03
Always use LegacyVersion for parsing versions

poncovka:
Fix names of some tests that use `packaging.version` for parsing versions.
`LegacyVersion` instances are always ordered lower than `Version` instances,
so we should use the `LegacyVersion` class instead of the parse function.

### DIFF
--- a/tests/unit_tests/pyanaconda_tests/modules/storage/test_module_storage_checker.py
+++ b/tests/unit_tests/pyanaconda_tests/modules/storage/test_module_storage_checker.py
@@ -118,7 +118,7 @@ class StorageCheckerVerificationTestCase(unittest.TestCase):
             .format("PHYSDISK-2", "VOLGROUP-B")
         )
 
-    def get_opal_kernel_version_test(self):
+    def test_get_opal_kernel_version(self):
         """Test the function for getting the firmware kernel version."""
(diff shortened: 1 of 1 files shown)


VladimirSlavik (commented):
a) I think you might be de facto reverting some changes from earlier? I remember we did something with `LegacyVersion` already.

b) This should be two commits or explained more. I don't see how the name changes are connected to the rest.

poncovka:
Updated. Basically, the tests cannot be enabled without the `LegacyVersion` fix. The following check is failing:
```
assert check("5.10.50-openpower1-p59fd803", "5.10")
```

jkonecny12 (approved):
Looks good to me. Could you please verify if the second commit is not missing on `rhel-9` branch?

VladimirSlavik (approved):
Looks good to me. Thank you!

poncovka:
> Looks good to me. Could you please verify if the second commit is not missing on `rhel-9` branch?

Checked. It should be fine.

hroncok:
Just got here when debugging a problem with https://fedoraproject.org/wiki/Changes/Update_python-packaging_to_version_22_plus

LegacyVersions were deprecated when they were added here. Now they are removed.

PR #3770, opened 2022-01-05
Skip /etc/machine-info during live installs (#2036199)

AdamWill:
/etc/machine-info should not be installed to the system, for
the same reason we don't install /etc/machine-id: it should be
system-specific, not the same for every install from the same
live image.

This should fix live installs with systemd 250+ not booting due
to the bootloader config snippets being named based on the
KERNEL_INSTALL_MACHINE_ID from the live image /etc/machine-info,
but grub2-mkconfig expecting them to be named according to the
installed system's newly-generated /etc/machine-id.

### DIFF
--- a/pyanaconda/modules/payloads/payload/live_image/installation.py
+++ b/pyanaconda/modules/payloads/payload/live_image/installation.py
@@ -363,6 +363,7 @@ class InstallFromTarTask(Task):
             "--exclude", "./boot/loader",
             "--exclude", "./boot/efi/loader",
             "--exclude", "./etc/machine-id",
+            "--exclude", "./etc/machine-info",
             "-xaf", self._tarfile,
             "-C", self._sysroot
         ]
@@ -437,6 +438,7 @@ class InstallFromImageTask(Task):
             "--exclude", "/boot/loader/",
             "--exclude", "/boot/efi/loader/",
             "--exclude", "/etc/machine-id",
+            "--exclude", "/etc/machine-info",
             os.path.normpath(self._mount_point) + "/",
             self._sysroot
         ]
--- a/tests/unit_tests/pyanaconda_tests/modules/payloads/payload/test_module_payload_live.py
+++ b/tests/unit_tests/pyanaconda_tests/modules/payloads/payload/test_module_payload_live.py
@@ -109,6 +109,7 @@ class InstallFromImageTaskTestCase(unittest.TestCase):
             "--exclude", "/boot/loader/",
             "--exclude", "/boot/efi/loader/",
             "--exclude", "/etc/machine-id",
+            "--exclude", "/etc/machine-info",
             mount_point + "/",
             "/mnt/root"
         ])
@@ -181,6 +182,7 @@ class InstallFromTarTaskTestCase(unittest.TestCase):
             "--exclude", "./boot/loader",
             "--exclude", "./boot/efi/loader",
             "--exclude", "./etc/machine-id",
+            "--exclude", "./etc/machine-info",
             "-xaf", f.name,
             "-C", "/mnt/root"
         ])


AdamWill:
See https://bugzilla.redhat.com/show_bug.cgi?id=2036199 and https://github.com/systemd/systemd/pull/21757 for the background here.

VladimirSlavik (approved):
Looks good to me. Thank you!

martinezjavier (approved):
This looks good to me as well

poncovka (approved):
Thank you!

jkonecny12 (approved):
Looks good to me AFAIK.

AdamWill:
If this looks good, could we merge it and do a build for Rawhide today or tomorrow? Until it's merged, live installs will be broken, which kills a lot of openQA test coverage. Thanks!

keszybz:
> /etc/machine-info should not be installed to the system, for the same reason we don't install /etc/machine-id: it should be system-specific, not the same for every install from the same live image.

The general idea is that if we create files under /boot and the machine-id is used there (in particular to create a directory), then we want to "persist" this machine-id by writing it to machine-info. This way, when we update or remove kernels in the future, we can adjust the appropriate directory, even if the machine-id on the destination system is different.

In other words, this patch is OK if the machine-id string does *not* appear under /boot.

> This should fix live installs with systemd 250+ not booting due to the bootloader config snippets being named based on the KERNEL_INSTALL_MACHINE_ID from the live image /etc/machine-info, but grub2-mkconfig expecting them to be named according to the installed system's newly-generated /etc/machine-id.

grub2-mkconfig should use KERNEL_INSTALL_MACHINE_ID from /etc/machine-info if present.

AdamWill:
@keszybz the subtleties here are all about who "we" is and what's being created when. Think about how live image creation actually works. The live image itself is built by installing a bunch of packages, including a kernel. So `kernel-install` gets run and writes things to `/boot` at that time, including config snippets in `/boot/loader/entries`. At that point the ID in `/etc/machine-id`, the ID in `/etc/machine-info` and the ID used in the config snippets are all the same; if you boot a live image and check those values, they all match.

But then things go sideways when someone actually installs Fedora to their own system from that live image. As part of that install process, *before* re-generating the bootloader config snippets, we create a new machine ID for the installed system, because we do not want every system installed from the same live image to have the same machine ID. But (without this patch) the machine ID of the live image itself is still defined as the `KERNEL_INSTALL_MACHINE_ID` in `/etc/machine-info`. So when we run `kernel-install` to re-generate the config snippets - which happens next - it uses that ID. Then we run `grub2-mkconfig`, with the intent that it should adjust the `options` lines in the config snippets to be correct. But because `grub2-mkconfig` expects the config snippet filenames to contain the machine ID from `/etc/machine-id` - not the `KERNEL_INSTALL_MACHINE_ID` from `/etc/machine-info` - it doesn't find them.

So I guess the key question here is: if you install Fedora from a live image, do we want the `KERNEL_INSTALL_MACHINE_ID` for that system to be the same as the `KERNEL_INSTALL_MACHINE_ID` of the live image itself, and *different* from the system's "machine ID" as defined in `/etc/machine-id`? If so, we should reject this patch, and patch `grub2-mkconfig` to prefer `KERNEL_INSTALL_MACHINE_ID` over `/etc/machine-id` if they conflict. Or does it make more sense for the installed system to clean and re-generate both IDs so they match? Which is what this patch achieves.

I guess I'm also struggling with why it's ever useful for these two IDs to be different - what the use case for that is, why it would ever be anything but a source of bugs. I've read the upstream PR a few times but I still don't get that part.

keszybz:
> So when we run `kernel-install` to re-generate the config snippets - which happens next - it uses the `KERNEL_INSTALL_MACHINE_ID`. Then we run `grub2-mkconfig`, with the intent that it should adjust the `options` lines in the config snippets to be correct. But because `grub2-mkconfig` expects the config snippet filenames to contain the machine ID from `/etc/machine-id` - not the `KERNEL_INSTALL_MACHINE_ID` from `/etc/machine-info` - it doesn't find them.

So I think this patch is fine: during installation the boot loader and the boot loader entries get created anew, and this means that resetting the machine-id and KERNEL_INSTALL_MACHINE_ID are appropriate.

> I guess I'm also struggling with why it's ever useful for these two IDs to be different - what the use case for that is, why it would ever be anything but a source of bugs. I've read the upstream PR a few times but I still don't get that part.

The motivation is that you have systems which have some entries installed in /boot using (some) machine-id, but that machine-id is reset. One example is "golden images", distributed as a .qcow2 or other image that gets used directly — such images ship with empty /etc/machine-id, but you still want to be able to add and remove entries under /boot using the original image-id. Another example is ephemeral images where the machine-id is not stored at all, but if there some boot-loader files, you want to keep using the same identifier, instead of creating an endless series of directories on each update.

VladimirSlavik:
Tested with last Live, installed, it booted. Let's have it. Thanks!

PR #3771, opened 2022-01-06
Create LVM devices file

VladimirSlavik:
Write the LVM devices file with all PVs, to keep the previous behavior. Then, copy it to the new system.

Remove the `%post` script that did the same.

Resolves: [rhbz#2011329](https://bugzilla.redhat.com/show_bug.cgi?id=2011329)

Tests TODO, code ready.


VladimirSlavik:
I am not sure how to actually make the tests pass. The blivet version needed for this is waiting to be <something> together with anaconda, so we won't get it, so the tests will always fail. I am not even sure what could the unit tests need, currently they fail on the one import so we won't know that they are broken...

vojtechtrefny (approved):
Looks good to me.

VladimirSlavik:
Nightly is ok. Code works. Still need to run tests though.

jkonecny12 (approved):
Looks good to me. Thanks!

poncovka (approved):
Thank you!

jstodola (approved):

M4rtinK (approved):

VladimirSlavik:
Note to self: When porting to upstream, add tests. Coverage of the module with this task is only 44%.

VladimirSlavik:
Ported in #4292.

PR #3773, opened 2022-01-07
Always set system purpose from the GUI

M4rtinK:
Fix the issue of system purpose not being set if the system is re-registered - registered first without system purpose being set, unregistered and registered again with system purpose *being* set.

Also fixup some outdated system purpose related strings I've noticed when looking into this.


VladimirSlavik (approved):
Looks good to me. Thank you!

poncovka (approved):
Looks good to me.

jkonecny12 (approved):
Looks good to me.

M4rtinK:
Port to Fedora: #4314

PR #3774, opened 2022-01-07
The OPAL compatibility with XFS features is mandatory

poncovka:
If the OPAL firmware is not compatible with XFS features, show a blocking error.

(cherry-picked from a commit 1bfc649)

Related: rhbz#2008792


jkonecny12 (approved):
Looks good to me. Thanks!

VladimirSlavik (approved):

PR #3775, opened 2022-01-07
Fix the root configuration in the user interface

poncovka:
Don't allow to enter the spoke if the root account is already configured
in the kickstart file in Anaconda. Just a note, the Initial Setup in the
reconfiguration mode always allows to enter this spoke.

If the root account is locked, show the "Root account is disabled" status.
Without the fix, Anaconda shows "Disabled, set password to enable" if there
is `firstboot --reconfig` in a kickstart file.

The workaround for the Initial Setup can be removed. It was implemented in
the commit 98d7b29 for the bug 1507940, but the bug didn't require to add
a special status.

If users disable the root account, we should reset the root password.

Resolves: rhbz#2015508


VladimirSlavik (approved):
Looks good to me. Thank you!

poncovka:
Rebased.

PR #3776, opened 2022-01-10
Simplify release notes (#infra)

VladimirSlavik:
Version no longer duplicated in title and text.

### DIFF
--- a/.github/workflows/tag-release.yml
+++ b/.github/workflows/tag-release.yml
@@ -97,8 +97,6 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           RELEASE_NAME: ${{ steps.get_version.outputs.version }}
         run: |
-          echo -e "Anaconda release $RELEASE_NAME.\n\n" > release.txt
-
           # find version tags
           CURRENT_VERSION="anaconda-$RELEASE_NAME-1"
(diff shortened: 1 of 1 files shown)


.github/workflows/tag-release.yml
```diff
@@ -109,12 +107,12 @@ jobs:
             --pretty=format:"- %s (%aL)" \
             "$PREVIOUS_VERSION..$CURRENT_VERSION~1" \
               | grep -v "#infra" \
-              | tee -a release.txt
```
jkonecny12, line 112: I think that we also want a trim to remove `\n` on the last line.

jkonecny12 (changes requested):
Please verify that the new line at the end is not a problem first. Otherwise looks great.

.github/workflows/tag-release.yml
```diff
@@ -109,12 +107,15 @@ jobs:
...
+              | tee release.txt
+
+          # drop line ending
+          truncate -s -1 release.txt
```
jkonecny12, line 113: I'm thinking about more robust solution in case there won't be a ending newline but -- naah that shouldn't happen. :D

jkonecny12 (approved):
LGTM!

.github/workflows/tag-release.yml
VladimirSlavik: There was that tempting perl one-liner somewhere on stack overflow...

PR #3779, opened 2022-01-13
Correct liveinst SELinux status check

AdamWill:
b11de43 accidentally inverted the logic of this check. It now
adds `--noselinux` if SELinux is enabled, which obviously isn't
what we want.

### DIFF
--- a/data/liveinst/liveinst
+++ b/data/liveinst/liveinst
@@ -47,7 +47,7 @@ if [ -x /usr/sbin/getenforce ]; then
     /usr/sbin/setenforce 0
 fi
 
-if (sestatus | grep -q enabled) then
+if (! sestatus | grep -q enabled) then
     ANACONDA="$ANACONDA --noselinux"
 fi
 


AdamWill:
It's kinda pseudo-tested - I put all three versions of the check into a script (with the result changed to an echo, of course) like this:

    #!/bin/sh

    if [ -z "$(sestatus | grep enabled)" ]; then
        echo "Old check: SELinux disabled"
    fi

    if (sestatus | grep -q enabled) then
        echo "Current check: SELinux disabled"
    fi

    if (! sestatus | grep -q enabled) then
        echo "Fixed check: SELinux disabled"
    fi

and ran it on a live image (via `sh` not `bash`, just to make sure the exclamation point isn't a problem in sh compatibility mode or something). It prints only "Current check: SELinux disabled". Then I tweaked the checks to look for 'enabledd' instead of 'enabled' (to simulate the case where it *should* disable SELinux), ran it again, and it prints "Old check: SELinux disabled" and "Fixed check: SELinux disabled". So that seems to prove the logic of the new check is the wrong way around, and the fixed check behaves the same as the old check.

PR #3781, opened 2022-01-14
Run chown instead of os.walk-ing to re-own home dir

VladimirSlavik:
- Provides the same result,
- chown is older and widely used so bugs are less likely,
- bugs are not in our code,
- Python stack exhaustion no longer happens.

Improve also docstrings and comments.

Resolves: [rhbz#2036415](https://bugzilla.redhat.com/show_bug.cgi?id=2036415)


rvykydal (approved):
Looks good to me, thank you.

poncovka (approved):
Thank you!

PR #3785, opened 2022-01-19
po: start translating webui files (jsx) and add support for cockpit translation functions

KKoukiou:
* [x] Blocked on https://github.com/cockpit-project/cockpit/pull/16855


Makefile.am
```diff
@@ -115,7 +115,9 @@ po-pull:
 	TEMP_DIR=$$(mktemp --tmpdir -d anaconda-localization-XXXXXXXXXX) && \
 	git clone --depth 1 -b $(GIT_L10N_BRANCH) -- $(L10N_REPOSITORY) $$TEMP_DIR && \
 	cp $$TEMP_DIR/$(L10N_DIR)/*.po $(srcdir)/po/ && \
-	rm -rf $$TEMP_DIR
+	rm -rf $$TEMP_DIR && \
+	sed -i 's/X-Generator:\ Zanata 3\.5\.1//' $(srcdir)/po/kn.po && \
```
jkonecny12, line 119: We don't use Zanata so I guess we can drop this Hack?

jkonecny12 (approved):
Looks good to me otherwise.

jkonecny12:
There is typo in the commit message of https://github.com/rhinstaller/anaconda/pull/3785/commits/52bfc281a1b19056b2621a3e9a758b7f65b20b28 
`qith` -> `with`

Makefile.am
KKoukiou: The problem is that is string exists on the po files, https://github.com/rhinstaller/anaconda-l10n/blob/master/master/kn.po#L35 and makes them unparseable by cockpit's translation mechanism.

VladimirSlavik: I think the correct resolution here is to change the po files and remove the Zanata header portion. I will make such a PR for the l10n repo.

KKoukiou: makes total sense, thanks!

Makefile.am
```diff
@@ -115,7 +115,9 @@ po-pull:
...
+	find $(srcdir)/po -type f -exec sed -ri '/^\"Plural-Forms:/ s/;(\\n")$$/\1/' {} +
```
VladimirSlavik, line 120: `Plural-Forms` should have a semicolon at the end. Please verify that plurals actually work with this change.

VladimirSlavik: To expand on that: As I understand it, you consume the resulting translations with cockpit, so this tweaks the translation for the cockpit tech, which is based on Javascript. However, the GTK and text modes may work differently, as they use actual gettext binaries (`.so`?) via Python's FFI or some such. You are making a single translation file (IMO correct), so the changes for one tech stack may break the other one. That's my concern.

KKoukiou: Your concern makes sense. I can think of three ways to address this:
* Instead of symlinking po/ with ui/webui/po create the latter with this sed line to be applied extra.
* Or verify if the missing semicolon actually does not create any problem, in this case we can just leave it?
* Or find exactly the project that creates the bug linked in the commit message, and address the issue upstream, wait for a release and have it fixed for real

KKoukiou:
I was able to verify this PR by setting CockpitLang cookie to 'de' for example. Then the already translated strings (which are shared with the old UI) appeared in the new language selection.

Makefile.am
VladimirSlavik: I'd say the "verify if does not cause problems" is the right first choice here - I think the parsing may just survive that bit. Or not.

Can you do that, or should I help with this?

KKoukiou: I guess I can build the RPMS and then use the test VM I use for webui (make vm-run) to test the real installer. 

I will select some other language than english, let's say de, and check for example this string. https://github.com/rhinstaller/anaconda-l10n/blob/master/master/de.po#L4208

If it's translated in plural I guess removing the semicolon works?

KKoukiou: Or maybe better we can add a unit test for it? Do you have unit tests for the translations?

VladimirSlavik: We have a unit test, but it tests the translation files, for things like missing formats etc. So not really in the way that is needed here.

Makefile.am
KKoukiou: Removed this.

jkonecny12:
Hi Katerina, it's problem with packaging of the translation files:

```
error: Installed (but unpackaged) file(s) found:
   /usr/share/cockpit/anaconda-webui/po.af.js
   /usr/share/cockpit/anaconda-webui/po.ar.js
   /usr/share/cockpit/anaconda-webui/po.as.js
   /usr/share/cockpit/anaconda-webui/po.ast.js
   /usr/share/cockpit/anaconda-webui/po.be.js
   /usr/share/cockpit/anaconda-webui/po.bg.js
   /usr/share/cockpit/anaconda-webui/po.bn.js
   /usr/share/cockpit/anaconda-webui/po.bn_IN.js
   /usr/share/cockpit/anaconda-webui/po.ca.js
   /usr/share/cockpit/anaconda-webui/po.cs.js
   /usr/share/cockpit/anaconda-webui/po.da.js
   /usr/share/cockpit/anaconda-webui/po.de.js
   /usr/share/cockpit/anaconda-webui/po.el.js
   /usr/share/cockpit/anaconda-webui/po.en_GB.js
   /usr/share/cockpit/anaconda-webui/po.eo.js
   /usr/share/cockpit/anaconda-webui/po.es.js
   /usr/share/cockpit/anaconda-webui/po.et.js
   /usr/share/cockpit/anaconda-webui/po.eu.js
   /usr/share/cockpit/anaconda-webui/po.fa.js
   /usr/share/cockpit/anaconda-webui/po.fi.js
   /usr/share/cockpit/anaconda-webui/po.fil.js
   /usr/share/cockpit/anaconda-webui/po.fr.js
   /usr/share/cockpit/anaconda-webui/po.fur.js
   /usr/share/cockpit/anaconda-webui/po.fy.js
   /usr/share/cockpit/anaconda-webui/po.ga.js
   /usr/share/cockpit/anaconda-webui/po.gl.js
   /usr/share/cockpit/anaconda-webui/po.gu.js
   /usr/share/cockpit/anaconda-webui/po.he.js
   /usr/share/cockpit/anaconda-webui/po.hi.js
   /usr/share/cockpit/anaconda-webui/po.hr.js
   /usr/share/cockpit/anaconda-webui/po.hu.js
   /usr/share/cockpit/anaconda-webui/po.ia.js
   /usr/share/cockpit/anaconda-webui/po.id.js
   /usr/share/cockpit/anaconda-webui/po.it.js
   /usr/share/cockpit/anaconda-webui/po.ja.js
   /usr/share/cockpit/anaconda-webui/po.ka.js
   /usr/share/cockpit/anaconda-webui/po.kk.js
   /usr/share/cockpit/anaconda-webui/po.km.js
   /usr/share/cockpit/anaconda-webui/po.kn.js
   /usr/share/cockpit/anaconda-webui/po.ko.js
   /usr/share/cockpit/anaconda-webui/po.lt.js
   /usr/share/cockpit/anaconda-webui/po.lv.js
   /usr/share/cockpit/anaconda-webui/po.ml.js
   /usr/share/cockpit/anaconda-webui/po.mr.js
   /usr/share/cockpit/anaconda-webui/po.my.js
   /usr/share/cockpit/anaconda-webui/po.nb.js
   /usr/share/cockpit/anaconda-webui/po.nl.js
   /usr/share/cockpit/anaconda-webui/po.nn.js
   /usr/share/cockpit/anaconda-webui/po.oc.js
   /usr/share/cockpit/anaconda-webui/po.or.js
   /usr/share/cockpit/anaconda-webui/po.pa.js
   /usr/share/cockpit/anaconda-webui/po.pl.js
   /usr/share/cockpit/anaconda-webui/po.pt.js
   /usr/share/cockpit/anaconda-webui/po.pt_BR.js
   /usr/share/cockpit/anaconda-webui/po.ro.js
   /usr/share/cockpit/anaconda-webui/po.ru.js
   /usr/share/cockpit/anaconda-webui/po.si.js
   /usr/share/cockpit/anaconda-webui/po.sk.js
   /usr/share/cockpit/anaconda-webui/po.sq.js
   /usr/share/cockpit/anaconda-webui/po.sr.js
   /usr/share/cockpit/anaconda-webui/po.sr@latin.js
   /usr/share/cockpit/anaconda-webui/po.sv.js
   /usr/share/cockpit/anaconda-webui/po.ta.js
   /usr/share/cockpit/anaconda-webui/po.te.js
   /usr/share/cockpit/anaconda-webui/po.th.js
   /usr/share/cockpit/anaconda-webui/po.tr.js
   /usr/share/cockpit/anaconda-webui/po.uk.js
   /usr/share/cockpit/anaconda-webui/po.ur.js
   /usr/share/cockpit/anaconda-webui/po.zh_CN.js
   /usr/share/cockpit/anaconda-webui/po.zh_HK.js
   /usr/share/cockpit/anaconda-webui/po.zh_TW.js
```

Basically we need to include these files in our spec file. Ideally in the section `%files webui`, [this](https://docs.fedoraproject.org/en-US/packaging-guidelines/#_handling_locale_files) should be guide how to do that.

KKoukiou:
@VladimirSlavik I verified that the plurals did not break with my change on the semicolon. See the bottom right of the screens for the translated screen. The translation comes from https://github.com/rhinstaller/anaconda-l10n/blob/master/master/de.po#L4536

Screen Shot 2022-01-25 at 12 44 23
Screen Shot 2022-01-25 at 12 44 19

VladimirSlavik (approved):
Looks good to me. Thank you!

I don't entirely understand the last commit, but let's try.

PR #3786, opened 2022-01-19
Master add rpm mount source

jkonecny12:
*WIP* - crashes and has a bad UX but I don't have time to finish it, so if someone want to take it it's free.

This will add the new installation source RPM Mount which will enable installation from any mounted directory. Right now, the use-case is that we will use this to install from `/run/install/repo` which is mounted by Dracut part of Anaconda as the default for the media used for booting. The source will check if there is repository to be used and if yes we will use that without caring the source.

This should resolve non standard boot media with packages as the ones created by Rufus tool. Discussion [here](https://github.com/rhinstaller/anaconda/discussions/3594) and [here](https://github.com/rhinstaller/anaconda/pull/3529).


jkonecny12:
Screenshot_rpm_mount_source
Screenshot_rpm_mount_summary

pyanaconda/ui/gui/spokes/installation_source.glade
```diff
@@ -612,6 +612,23 @@
                                 <property name="position">2</property>
                               </packing>
                             </child>
+                            <child>
+                              <object class="GtkRadioButton" id="fromBootRadioButton">
```
VladimirSlavik, line 616: The radiobutton is missing left margin of 12, which is why the screenshot looks crappy.

pyanaconda/modules/payloads/constants.py
```diff
@@ -31,6 +31,9 @@
     '/etc/anaconda.repos.d'
 ]
 
+# Location of the repository directory in Dracut (may not container repository)
+DRACUT_REPO_DIR = "/run/install/repo"
```
poncovka, line 35: Please, move the constant to `pyanaconda.core.constants`. It is used outside of the module.

pyanaconda/core/constants.py
```diff
@@ -424,6 +424,7 @@ class DisplayModes(Enum):
 SOURCE_TYPE_CDROM = "CDROM"
 SOURCE_TYPE_CLOSEST_MIRROR = "CLOSEST_MIRROR"
 SOURCE_TYPE_REPO_FILES = "REPO_FILES"
+SOURCE_TYPE_RPM_MOUNT = "RPM_MOUNT"
```
poncovka, line 427: I don't like the name at all. It is confusing, because it looks like we are trying to mount RPMs. Based on the code, it looks like we don't actually care if it is a mount point, do we? What about `REPO_PATH` or just `DIR`?

pyanaconda/modules/payloads/source/rpm_mount/rpm_mount.py
```diff
@@ -0,0 +1,145 @@
...
+
+    @property
+    def type(self):
+        """Get type of this source."""
+        return SourceType.RPM_MOUNT
+
+    @property
+    def description(self):
+        """Get description of this source."""
+        return _("RPM Mount {}".format(self._path))
```
poncovka, line 65: The CDROM source uses "Local media", so I would use "Local directory". The current description is too confusing.

pyanaconda/modules/payloads/constants.py
jkonecny12: I was already thinking about that.

pyanaconda/core/constants.py
jkonecny12: `REPO_PATH` seems better to me. The `DIR` name is too generic.

pyanaconda/modules/payloads/source/rpm_mount/rpm_mount.py
jkonecny12: Agree, another thing I wanted to change - especially when I saw it on the Source spoke status.

marcosfrm:
F36 is already gone at this point I suppose. Please try to push it to F37!

marcosfrm:
Ping.

poncovka:
These changes are incorporated into https://github.com/rhinstaller/anaconda/pull/4567.

PR #3789, opened 2022-01-20
Use DBus read-write properties

poncovka:
Replace `Set.*` methods with with read-write properties to clean up
the DBus API and simplify unit testing with mocked DBus objects.

**TODO:**

- [ ] Update Initial Setup.
- [ ] Update KDUMP Add-on.
- [ ] Update OSCAP Add-on.
- [ ] Run coverage tests.

**Blocked by:** Branch Fedora Linux 36 from Rawhide


pyanaconda/modules/users/users_interface.py
```diff
@@ -100,11 +100,6 @@ def IsRootPasswordSet(self) -> Bool:
         """
         return self.implementation.root_password_is_set
 
-    @emits_properties_changed
-    def SetRootAccountLocked(self, root_account_locked: Bool):
```
KKoukiou, line 104: This is used in the webui tests. Which means they will break with this. I will trigger the tests here to verify :)
Would be nice if we can update the test in the same PR, so that we don't have the broken on master after this.

VladimirSlavik (commented):
Python LGTM, can't really comment on the webui change.

KKoukiou:
@poncovka the webui tests are still failing https://logs.cockpit-project.org/logs/pull-3789-20220124-090914-b90b5bd8-fedora-35-rawhide/log.html

I can help you debug this if the issue if not clear.

pyanaconda/modules/users/users_interface.py
poncovka: Thanks! I plan to check the code again before the final review, because there is some new code in the Web UI now. I will run every test we have before it will be merged to make sure it doesn't break anything :-)

KKoukiou: webui tests are automatically running now with each force push, so no worries about triggering these ones.

VladimirSlavik (approved):

jkonecny12 (commented):
Looks good to me so far :)

PR #3790, opened 2022-01-21
webui: parameterize ports for ssh, cockpit connection and http server

KKoukiou:
Previously no parallel test runs were possible, since there was port collision on the host.
This commit makes it possible for local test runs or CI to run in
paralle.

### DIFF
--- a/ui/webui/test/run-test
+++ b/ui/webui/test/run-test
@@ -4,13 +4,39 @@ set -e
 
 clean_up () {
     ARG=$?
-    pkill -9 -f  'python3 -m http.server'
-    virsh -c qemu:///session destroy test-updates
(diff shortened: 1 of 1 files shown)


VladimirSlavik (approved):
LGTM, thank you.

I wonder about the hardcoded "35", but that's probably more complicated.

KKoukiou:
> LGTM, thank you.
> 
> I wonder about the hardcoded "35", but that's probably more complicated.

When we start testing something else than rawhide we need to parametrize this as well.

KKoukiou:
Hmm this seems to get stuck in the CI. Locally it finishes needs debugging

jkonecny12 (approved):
LGTM! Thanks!

PR #3791, opened 2022-01-21
Test a recreation of the same thread

poncovka:
Add a test for recreation of the same thread. There is an unresolved bug
in Anaconda, so the test is marked as an expected failure.


VladimirSlavik (approved):
Looks good to me. Thank you!

PR #3794, opened 2022-01-26
packit: release: unset use_cockpit by sedding the specfile in packit script

KKoukiou:
Also, remove the unpackaged anaconda-webui files in the case `use_cockpit` is unset.

### DIFF
--- a/.packit.yml
+++ b/.packit.yml
@@ -6,6 +6,7 @@ actions:
   post-upstream-clone:
     - ./autogen.sh
     - ./configure
+    - sed -i 's/use_cockpit\ 1/use_cockpit 0/g' anaconda.spec
   create-archive:
     - "make release"
     - 'bash -c "ls -1 anaconda-*.tar.bz2"'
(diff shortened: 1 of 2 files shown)


.packit.yml
```diff
@@ -6,6 +6,7 @@ actions:
   post-upstream-clone:
     - ./autogen.sh
     - ./configure
+    - sed -i 's/use_cockpit\ 1/use_cockpit 0/g' anaconda.spec
```
jkonecny12, line 9: Honestly I think we should have "make release-cockpit" to solve this better way.

jkonecny12 (approved):
Other than the minor note below, it looks good to me.

.packit.yml
KKoukiou: Hmm, can your elaborate? I am not sure I understand.

KKoukiou: discussion IRC

M4rtinK (approved):
Looks good to me, thanks for fixing this a bit tricky thing! :)

PR #3795, opened 2022-01-26
Clean up the code for including Web UI in an updates image

poncovka:
The `makeupdates` script should check if `ui/webui/dist` exists and copy it only
once. Otherwise, the script will fail or copy the same directory multiple times.


KKoukiou (approved):
I saw this actually happening! Thanks for tracking it down!

VladimirSlavik (approved):
Thanks!

PR #3745, opened 2021-12-07
Fix the condition for entering the root configuration (#2026916)

poncovka:
Don't allow to enter the spoke if the root account is already configured
in the kickstart file in Anaconda. Just a note, the Initial Setup in the
reconfiguration mode always allows to enter this spoke.

(ported from https://github.com/rhinstaller/anaconda/pull/3701)

Resolves: rhbz#2026916

PR #3800, opened 2022-01-28
Require admin user

VladimirSlavik:
Previously, it was possible to satisfy the conditions just by setting root password, even if root was locked.

This changes both root and user spoke to signal the incomplete user.

Resolves: rhbz#2015508
Resolves: [rhbz#2047713](https://bugzilla.redhat.com/show_bug.cgi?id=2047713)

(cherry picked from commit 4cd6cd5e751c0f2aec923584c66090ced7e818fb - this backports #3665.)

### DIFF
--- a/pyanaconda/ui/gui/spokes/root_password.py
+++ b/pyanaconda/ui/gui/spokes/root_password.py
@@ -193,7 +193,10 @@ class PasswordSpoke(FirstbootSpokeMixIn, NormalSpoke, GUISpokeInputCheckHandler)
 
     @property
     def mandatory(self):
-        """Only mandatory if no admin user has been requested."""
+        """Only mandatory if no admin user has been requested.
+
+        See also doc for the property completed().
+        """
         return not self._users_module.CheckAdminUserExists()
 
     def apply(self):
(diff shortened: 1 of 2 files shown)


poncovka:
I would like to block this for now, because there are https://github.com/rhinstaller/anaconda/pull/3745 and https://github.com/rhinstaller/anaconda/pull/3775 to consider. The fix needs careful investigation.

VladimirSlavik:
CC also @jstodola - I believe we already had this conversation once.

jstodola:
Right, the new root configuration screen was not accepted to RHEL-9.0 in order to not introduce big changes after RHEL-9-Beta, also due to changes necessary in anabot to support the new screen and update of the existing anabot templates for automated testing.
I've reported bug https://bugzilla.redhat.com/show_bug.cgi?id=2048692 so that we can consider the new screen for RHEL-9.1.

jkonecny12:
After a discussion with everyone. This issue is postponed to 9.1 where we will do the correct solution by backporting the code from upstream. @VladimirSlavik I guess in that case we want to close this?

VladimirSlavik:
Yes, let's close this, and revisit the idea later. I believe we are tracking this in multiple ways both internally and externally.

PR #3802, opened 2022-01-29
Don't use progressQ in the payload classes

poncovka:
Use the `_progress_cb` method instead.

### DIFF
--- a/pyanaconda/installation.py
+++ b/pyanaconda/installation.py
@@ -374,6 +374,10 @@ def run_installation(payload, ksdata):
     queue.append(_prepare_installation(payload, ksdata))
     queue.append(_prepare_configuration(payload, ksdata))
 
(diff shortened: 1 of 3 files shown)


M4rtinK (approved):
Looks good to me. :)

VladimirSlavik (commented):
I don't understand what this does without the other pieces. Or maybe I don't understand it at all...

poncovka:
> I don't understand what this does without the other pieces. Or maybe I don't understand it at all...

I need to eventually replace `progress_message` with another callback that will be accessible only from the `pyanaconda.installation` module. This pull request makes sure, that the payload doesn't use `progress_message` directly, but it is actually `pyanaconda.installation` that sets this function as a callback.

poncovka:
Rebased and tested.

PR #3804, opened 2022-01-29
Clean up pyanaconda.installation_tasks

poncovka:
Simplify the code and remove everything we don't use.

### DIFF
--- a/pyanaconda/installation_tasks.py
+++ b/pyanaconda/installation_tasks.py
@@ -17,15 +17,12 @@
 # License and may only be used or replicated with the express permission of
 # Red Hat, Inc.
 #
-from threading import RLock
+import time
 
 from dasbus.error import DBusError
 from pyanaconda.core.signal import Signal
-from pyanaconda.core.util import synchronized
 from pyanaconda.errors import errorHandler, ERROR_RAISE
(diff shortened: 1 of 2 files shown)


poncovka:
Rebased and tested.

pyanaconda/installation_tasks.py
```diff
@@ -148,187 +119,93 @@ def status_message(self):
...
-                self._current_task_number = None
-                self._current_queue_number = None
-
-            # trigger the "completed" signals
-            self.completed.emit(self)
+    def _run(self):
+        """Run the task queue."""
+        for item in self._queue:
+            # start the item (TaskQueue/Task)
+            item.start()
```
VladimirSlavik, line 189: You start the item and connect the signals only later, I think? I don't see why this can work?

poncovka: It is a little messy, the signals are connected when you append an item to a queue.

poncovka:
Btw. this code is not going to survive modularization, but it is simpler to keep it alive in its minimal form for now.

jkonecny12 (approved):
Looks good to me, however, I wonder how is possible that you were able to remove so much of the code.

PR #3805, opened 2022-01-31
Fix anaconda-rpm container images

jkonecny12:
I'm fixing two issues here.
1) We were building anaconda-ci containers as anaconda-rpm (that is the reason why they have the same size BTW).
2) Fix failing tests because of [annobin issue](https://bugzilla.redhat.com/show_bug.cgi?id=2047148). There is a new package with the fix available but not yet in the repository.


poncovka (approved):
Thank you!

VladimirSlavik (approved):
Thank you!

PR #3806, opened 2022-02-01
Fix packit builds on COPR

jkonecny12:
Packit builds are now broken because of [annobin issue](https://bugzilla.redhat.com/show_bug.cgi?id=2047148). The fixed RPM is available but Rawhide builds are broken so we don't know when this will be officially available.
To solve this add this hotfix solution which could be used also in the future for similar issues.

### DIFF
--- a/.packit.yml
+++ b/.packit.yml
@@ -27,6 +27,10 @@ jobs:
     metadata:
       targets:
         - fedora-rawhide
+      additional_repos:
+        # This repository contains fixup of Rawhide broken environment.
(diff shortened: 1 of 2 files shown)


M4rtinK (approved):
Looks good to me, thanks! :)

poncovka (approved):
Looks good to me. Ideally, we could have additional repositories for all branches and use them also in containers, right? Then we wouldn't have to modify Dockerfiles with temporary fixes.

PR #3808, opened 2022-02-01
webui: users: do not try to use the proxy before it's ready

KKoukiou:
This fixes the flaky test which is always failing on first try as seen
here.

https://grafana-frontdoor.apps.ocp.ci.centos.org/d/ci/cockpit-ci?orgId=1&var-cockpit_project=rhinstaller%2Fanaconda

### DIFF
--- a/ui/webui/src/Common.jsx
+++ b/ui/webui/src/Common.jsx
@@ -30,7 +30,7 @@ const _ = cockpit.gettext;
 
 export const AddressContext = createContext("");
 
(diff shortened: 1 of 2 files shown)


KKoukiou:
Nope did not work, moving to draft.

KKoukiou:
Tests are passing now without retry.

M4rtinK (approved):
Looks good as far as I can tell. :)

PR #3810, opened 2022-02-02
Make Anaconda Cockpit ui as optional

jkonecny12:
Right now running `make -f ./Makefile.am container-ci` taking a long time because the new cockpit part is downloading npm packages. To make situation worse the container-ci will always remove the intermediate results so the npm packages are downloaded for every call.

To improve the situation make the `ui/webui` part optional during the build when `./configure --disable-cockpit` is added. Use this for the `container-ci` make target as default because we don't run anything web UI related there anyway.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -209,9 +209,21 @@ AC_CONFIG_FILES([Makefile
                  data/pixmaps/Makefile
                  tests/Makefile
                  ui/Makefile
-                 ui/webui/Makefile
                  utils/Makefile
                  utils/dd/Makefile])
+
+### Include Cockpit part to the make execution as optional ###
+# The main reason to have this is to speed up execution of the unit-tests by avoiding npm package download.
+AC_ARG_ENABLE([webui], AS_HELP_STRING([--disable-webui], [Build without the web ui; makes the build much faster]))
+
+# Do not include the ui/webui Makefile to the build process. Everything there will be ignored.
+AS_IF([test "x$enable_webui" != "xno"],
+      [AC_CONFIG_FILES([ui/webui/Makefile])])
+AS_IF([test "x$enable_webui" != "xno"],
+      [AC_SUBST(WEBUI_SUBDIR, "webui")],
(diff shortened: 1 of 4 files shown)


poncovka (approved):
I like it. Thank you!

configure.ac
```diff
@@ -209,9 +209,21 @@ AC_CONFIG_FILES([Makefile
                  data/pixmaps/Makefile
                  tests/Makefile
                  ui/Makefile
-                 ui/webui/Makefile
                  utils/Makefile
                  utils/dd/Makefile])
+
+### Include Cockpit part to the make execution as optional ###
+# The main reason to have this is to speed up execution of the unit-tests by avoiding npm package download.
+AC_ARG_ENABLE([cockpit], AS_HELP_STRING([--disable-cockpit], [Build without the cockpit ui; makes the build much faster]))
```
KKoukiou, line 217: s/cockpit/webui

Indeed I have used cockpit in the spec file, but that's rather an exception. Let's use webui everywhere I suggest.

KKoukiou (approved):
Aside from the naming comment, I am happy with this solution as well.

VladimirSlavik (approved):

jkonecny12:
*UPDATED:*
* Renamed all cockpit to web UI.

Also a note, for some reason this also disables download of the po files. Honestly I'm not sure why.

ui/Makefile.am
```diff
@@ -15,6 +15,7 @@
 # You should have received a copy of the GNU Lesser General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-SUBDIRS = webui
+# WebUI could be excluded by ./configure --disable-cockpit
```
KKoukiou, line 18: disable-webui in the comment.

jkonecny12: Fixed. Thanks!

jkonecny12:
Packit has a networking issue - they are trying to resolve it right now. Thus the failure.

poncovka (approved):

PR #3811, opened 2022-02-02
Use a task to write repositories on the target system

poncovka:
Run the WriteRepositoriesTask task to install .repo files generated from
the specified repo configuration data. The implementation was simplified:

* The `repo --name=updates --install` use case is no longer supported.
  It was implemented in a very weird way, that didn't really make sense.

* Only repositories that use one the http, https and ftp protocols are
  supported. It wasn't clear how the other protocols could work.

* We don't use the configuration of the DNF objects to generate the .repo
  files. The DNF objects should be in sync with the repo configuration data.


jkonecny12 (approved):
Looks good to me. I just wonder why we don't use DNF to dump the repository file instead...

poncovka:
> Looks good to me. I just wonder why we don't use DNF to dump the repository file instead...

We generate a minimal repo configuration based on the specified repo data. If I understand it correctly, DNF can only dump a complete repo configuration that might be based on other inputs (main configuration, boot options, ...).

PR #3814, opened 2022-02-03
Do not copy /etc/resolv.conf to chroot before installation

rvykydal:
It does not seem to make sense anymore given systemd-resolved is
creating its own symlink in rpm postin script that rewrites the file.

(see rhbz#2018913, rhbz#2032085)


poncovka (approved):
Yay!

VladimirSlavik (approved):
Thank you!

PR #3813, opened 2022-02-03
Fix Packit configuration for f36-release branch (#infra)

jkonecny12:
By changing the propose downstream here we should get releases on this branch to the correct downstream branch.

See https://github.com/packit/packit-service/issues/1298 for more info.

PR #3815, opened 2022-02-03
Adjust release configuration and Packit configuration

jkonecny12:
Let's do this in the PR:
* Fix Anaconda-devel COPR daily builds to f36-devel branch instead of f35-devel
* Propose downstream releases from this branch to Rawhide only. F36 releases should be solved by https://github.com/rhinstaller/anaconda/pull/3813

### DIFF
--- a/.packit.yml
+++ b/.packit.yml
@@ -13,8 +13,6 @@ actions:
 jobs:
   - job: propose_downstream
     trigger: release
-    metadata:
-      dist_git_branches: fedora-development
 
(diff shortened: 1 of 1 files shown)


jkonecny12:
This is blocked because the Fedora branching is not yet available. There is no such target now.

poncovka (approved):
Thanks!

VladimirSlavik (approved):
Thank you!

PR #3818, opened 2022-02-03
Do not copy resolv.conf to target system at the end of installation

rvykydal:
An /etc/resolv.conf symlink is created on target system by
systemd-resolved in postin rpm script.  And if it is not (bug #2018913)
it would be crated during boot of installed system (by systemd-resolved).

The only case where anaconda copying the resolv.conf could have actually
effect would be overwriting /etc/resolv.conf on target system in non-rpm
payload installations (image installations). But even in this case
Anaconda should not interfere into management of /etc/resolv.conf on the
target system. It is a business of either the image creator or a service
on starting installed system (systemd-resolved or NM).

### DIFF
--- a/pyanaconda/modules/network/installation.py
+++ b/pyanaconda/modules/network/installation.py
@@ -98,7 +98,6 @@ class NetworkInstallationTask(Task):
 
     SYSCONF_NETWORK_FILE_PATH = "/etc/sysconfig/network"
     ANACONDA_SYSCTL_FILE_PATH = "/etc/sysctl.d/anaconda.conf"
-    RESOLV_CONF_FILE_PATH = "/etc/resolv.conf"
     NETWORK_SCRIPTS_DIR_PATH = IFCFG_DIR
     PREFIXDEVNAME_CONFIG_FILE_PREFIX = "71-net-ifnames-prefix-"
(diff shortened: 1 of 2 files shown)


VladimirSlavik (approved):
Code LGTM, thank you.

poncovka (approved):
Thanks!

poncovka:
@rvykydal, could you look at https://github.com/rhinstaller/anaconda/pull/3818#issuecomment-1170497754, please?

cverna:
It seems this is also causing issues for the container base image https://bugzilla.redhat.com/show_bug.cgi?id=2106999

gbraad:
This causes issues for non-root container images based on Fedora, for example Gitpod image use the `gitpod` user to run, but now DNS resolving fails unless a workaround is used in the container definition along he lines of:

```
RUN chmod 644 /etc/resolv.conf
```
[ref](https://github.com/gbraad/devenv/blob/0a5dac43db7844ff3589060a69084a86ecaf70f9/gitpod/install.sh#L4)

dustymabe:
I still think this was the right thing for anaconda to do. I commented at https://github.com/containers/podman-wsl-fedora/issues/3#issuecomment-1190574059

gbraad:
> I still think this was the right thing for anaconda to do. 

Not arguing this. However this might now be a 'known issue' for non-root use of the container when the container runtime doesn't do the right thing.

PR #3819, opened 2022-02-03
Add anaconda-release container

jkonecny12:
Add new container `anaconda-release` to help us to do releases on Fedora of RHEL Anaconda.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -92,9 +92,11 @@ CONTAINER_ADD_ARGS ?=
 # anaconda-ci container
 CI_DOCKERFILE ?= $(srcdir)/dockerfile/anaconda-ci
 RPM_DOCKERFILE ?= $(srcdir)/dockerfile/anaconda-rpm
+RELEASE_DOCKERFILE ?= $(srcdir)/dockerfile/anaconda-release
 ISO_CREATOR_DOCKERFILE ?= $(srcdir)/dockerfile/anaconda-iso-creator
 CI_NAME ?= $(CONTAINER_REGISTRY)/rhinstaller/anaconda-ci
 RPM_NAME ?= $(CONTAINER_REGISTRY)/rhinstaller/anaconda-rpm
+RELEASE_NAME ?= $(CONTAINER_REGISTRY)/rhinstaller/anaconda-release
 ISO_CREATOR_NAME ?= $(CONTAINER_REGISTRY)/rhinstaller/anaconda-iso-creator
 CI_TAG ?= $(GIT_BRANCH)
 CI_TEST_ARGS ?= --rm -v $(srcdir):/anaconda:Z
@@ -183,6 +185,14 @@ release:
 	$(MAKE) po-pull
 	$(MAKE) dist
 
+container-release:
+	$(CONTAINER_ENGINE) run \
+	$(CONTAINER_TEST_ARGS) \
+	$(CI_TEST_ARGS) \
+	$(CONTAINER_ADD_ARGS) \
+	$(RELEASE_NAME):$(CI_TAG) \
+	sh -exc './autogen.sh && ./configure && make release'
+
 release-and-tag:
 	$(MAKE) dist
 	$(MAKE) tag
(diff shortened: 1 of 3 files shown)


jkonecny12:
@jstodola another request for PR, it's simplification for us to make releases.

M4rtinK (approved):
Looks good to me. :)

dockerfile/anaconda-release/Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+# Add missing dependencies required to do the build.
+RUN set -e; \
+  dnf install -y \
+  git \
+  python3-pip; \
+  dnf clean all;
+
+RUN pip3 install \
+  polib \
+  pocketlint
```
VladimirSlavik, line 18: I think some of the other containers run `pip3 install --upgrade pip (..)` - is that not needed on RHEL? That would be  good, of course, just not sure.

VladimirSlavik (approved):
Looks good to me. Thank you!

jstodola (approved):

dockerfile/anaconda-release/Dockerfile
poncovka: I would prefer to update and upgrade all packages.

Makefile.am
```diff
@@ -183,6 +185,14 @@ release:
 	$(MAKE) po-pull
 	$(MAKE) dist
 
+container-release:
+	$(CONTAINER_ENGINE) run \
+	$(CONTAINER_TEST_ARGS) \
+	$(CI_TEST_ARGS) \
```
poncovka, line 191: We might want to eventually reconsider names of some of the variables that are used by all our containers.

Makefile.am
```diff
@@ -183,6 +185,14 @@ release:
...
+	$(CONTAINER_ADD_ARGS) \
+	$(RELEASE_NAME):$(CI_TAG) \
+	sh -exc './autogen.sh && ./configure && make release'
```
poncovka, line 194: Could we use `run-build-and-arg`, please? Then we wouldn't have to clean up the repository. Or is it a different use case?

Makefile.am
```diff
@@ -92,9 +92,11 @@ CONTAINER_ADD_ARGS ?=
 # anaconda-ci container
 CI_DOCKERFILE ?= $(srcdir)/dockerfile/anaconda-ci
 RPM_DOCKERFILE ?= $(srcdir)/dockerfile/anaconda-rpm
+RELEASE_DOCKERFILE ?= $(srcdir)/dockerfile/anaconda-release
 ISO_CREATOR_DOCKERFILE ?= $(srcdir)/dockerfile/anaconda-iso-creator
 CI_NAME ?= $(CONTAINER_REGISTRY)/rhinstaller/anaconda-ci
 RPM_NAME ?= $(CONTAINER_REGISTRY)/rhinstaller/anaconda-rpm
+RELEASE_NAME ?= $(CONTAINER_REGISTRY)/rhinstaller/anaconda-release
```
poncovka, line 99: We should eventually start to use better names that would indicate that these are container names.

poncovka (approved):
I have some suggestions for future improvements. Otherwise, it looks good to me.

Makefile.am
jkonecny12: I don't think we want that. We need to have tarball as output and I don't really want to move the tarball around in the container to get it out. Also you should not be bothered by messing up the repository because release should always be from a new checkout as the guides said.

dockerfile/anaconda-release/Dockerfile
VladimirSlavik: That's directly at odds with the idea that the Python packages will be stable, packaged by the distro, as opposed to whatever is new on pip. The `pip --upgrade pip` thing was a hotfix, IIRC.

jkonecny12:
**UPDATED:**
* Add pip update
* Add dnf update

poncovka (approved):

jkonecny12:
**UPDATED:**
* Return back the `dnf update -y` it will work after https://github.com/rhinstaller/anaconda/pull/3892 is merged.

poncovka (approved):

rvykydal (approved):
Looks good to me.

PR #3821, opened 2022-02-03
Add anaconda-release container image

jkonecny12:
Use this container in the tag-release workflow.

Tested here:
https://github.com/jkonecny12/anaconda/runs/5301979004?check_suite_focus=true
Output:
https://github.com/jkonecny12/anaconda/releases/tag/untagged-7fbee5da484094695804

*TODO:*
- [x] Test this properly.
- [x] ~Add https://github.com/rhinstaller/anaconda/pull/3868 to here~ Not required, it's already in the anaconda-rpm on master and not it's not mandatory on f36 branches because there is no npm.

### DIFF
--- a/.github/workflows/tag-release.yml
+++ b/.github/workflows/tag-release.yml
@@ -60,36 +60,17 @@ jobs:
           ref: ${{ github.ref }}
           fetch-depth: 0
 
-      - name: Build anaconda-rpm container (for RPM build)
+      - name: Build anaconda container (to make the release)
         run: |
-          make -f Makefile.am anaconda-rpm-build
+          make -f Makefile.am anaconda-release-build
 
       - name: Run the build in the container
         run: |
           mkdir /tmp/results
-          podman run \
-            -i \
-            -v /tmp/results:/results \
-            -v `pwd`:/anaconda \
-            quay.io/rhinstaller/anaconda-rpm:master <<EOF
-
-          # prepare environment
-          set -eux
-
-          # install dependencies for translation canary
-          dnf install -y git python3-polib python3-pip
-          pip install --upgrade pip pocketlint
-
-          # build tarball
-          cd /anaconda
-          ./autogen.sh
-          ./configure
-          make release
+          make -f Makefile.am container-release
 
           # copy out stuff
-          cp anaconda-*.tar.bz2 /results
-
-          EOF
+          cp anaconda-*.tar.bz2 /tmp/results
 
       - name: Create the release
         id: create_release
--- a/Makefile.am
+++ b/Makefile.am
@@ -75,7 +75,7 @@ TEST_INST_BUILD_DIR = $(BUILD_RESULT_DIR)/02-test-install
 
 CONTAINER_ENGINE ?= podman
 # build/run tweaks for all containers, such as --cap-add
-CONTAINER_BUILD_ARGS ?= --no-cache --pull-always
+CONTAINER_BUILD_ARGS ?= --no-cache
 # HACK: bash's builtin `test -r` fails when running on Ubuntu host (GitHub) due to incompatible seccomp profile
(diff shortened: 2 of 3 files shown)


poncovka:
I have got confused by the ported pull requests and reviewed this one instead: https://github.com/rhinstaller/anaconda/pull/3819#pullrequestreview-872421050

jkonecny12:
No problem with that.

dockerfile/anaconda-release/Dockerfile
```diff
@@ -3,7 +3,8 @@
 # This will be based on our existing anaconda-rpm container because it has all the dependencies
 # required to build Anaconda.
 
-FROM quay.io/rhinstaller/anaconda-rpm:rhel-8
+ARG image
```
VladimirSlavik, line 6: Why not add a default? `ARG image=quay.io/rhinstaller/anaconda-rpm:f36`

.github/workflows/tag-release.yml
```diff
@@ -62,35 +62,16 @@ jobs:
...
-          # install dependencies for translation canary
-          dnf install -y git python3-polib python3-pip
-          pip install --upgrade pip pocketlint
-
-          # build tarball
-          cd /anaconda
-          ./autogen.sh
-          ./configure
-          make release
+          mkdir -f Makefile.am container-release
```
VladimirSlavik, line 70: Um.

VladimirSlavik (commented):
Mostly LGTM.

Cherry-picking and then changing the values seems confusing, but that's not the main problem.

dockerfile/anaconda-release/Dockerfile
jkonecny12: I would like to get rid of the default values because it gives more work during branching.

jkonecny12:
**UPDATED:**
* Small fixes to make this works.
* Add new commit to remove --pull-always from the CONTAINER_BUILD_ARGS and use that only on some places. Explanation is in commit message.

poncovka (approved):
Thanks!

dockerfile/anaconda-release/Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+# Add missing dependencies required to do the build.
+RUN set -e; \
+  dnf update -y; \
+  dnf install -y \
+  git \
+  python3-pip; \
+  dnf clean all;
+
+RUN pip install --no-cache-dir --upgrade pip; \
+    pip3 install --no-cache-dir \
```
VladimirSlavik, line 20: One runs `pip`, another `pip3`.

Makefile.am
```diff
@@ -221,7 +221,7 @@ anaconda-release-build: anaconda-rpm-build
 
 anaconda-iso-creator-build:
 	$(CONTAINER_ENGINE) build \
-	$(CONTAINER_BUILD_ARGS) \
+	$(CONTAINER_BUILD_ARGS) --pull-always \
 	$(CONTAINER_ADD_ARGS) \
```
VladimirSlavik, line 225: I think I no longer understand why we have `CONTAINER_BUILD_ARGS` and `CONTAINER_ADD_ARGS`, and why this additional parameter isn't in one of these.

(I think this can be resolved later, we want the container now.)

VladimirSlavik (approved):
Thank you, mostly LGTM, and my concerns can be addressed later.

Makefile.am
jkonecny12: You can overload that easily by make call and also it's easier to avoid missing some build when adding arguments.

dockerfile/anaconda-release/Dockerfile
jkonecny12: Fixed, thanks for pointing that out.

jkonecny12:
**UPDATED**
* Fixed note pointed out by @VladimirSlavik - thanks!

poncovka (approved):

PR #3812, opened 2022-02-03
Always pull the base image during container builds (#infra)

jkonecny12:
We want to pull the base image update to not build outdated images. Without this the base image is used from the cache if present.

PR #3823, opened 2022-02-03
Enable the Koji repository in the RPM tests

poncovka:
Use the latest updates to fix our issues with the RPM tests.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -75,7 +75,7 @@ TEST_INST_BUILD_DIR = $(BUILD_RESULT_DIR)/02-test-install
 
 CONTAINER_ENGINE ?= podman
 # build/run tweaks for all containers, such as --cap-add
-CONTAINER_BUILD_ARGS ?= --no-cache
+CONTAINER_BUILD_ARGS ?= --no-cache --pull-always
 # HACK: bash's builtin `test -r` fails when running on Ubuntu host (GitHub) due to incompatible seccomp profile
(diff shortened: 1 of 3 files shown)


poncovka:
I have no idea what is actually broken, but enabling the Koji repository seems to resolve the issues with RPM tests.

Makefile.am
```diff
@@ -75,7 +75,7 @@ TEST_INST_BUILD_DIR = $(BUILD_RESULT_DIR)/02-test-install
 
 CONTAINER_ENGINE ?= podman
 # build/run tweaks for all containers, such as --cap-add
-CONTAINER_BUILD_ARGS ?= --no-cache
+CONTAINER_BUILD_ARGS ?= --no-cache --pull-always
```
jkonecny12, line 78: Already part of https://github.com/rhinstaller/anaconda/pull/3812.

jkonecny12 (changes requested):
This solution is great for test that there is already a fix in koji but I would like to avoid having Koji as our repository. It could break any second and there is no testing of that.

jkonecny12:
Seems that this issue is now resolved even without this PR. Tested locally and I'll also run the container refresh to be sure.

jkonecny12:
Tested, we have working infra again - we can close this now.

poncovka:
Well, I am going to fix the next breakage the same way. We can discuss this solution at Monday's meeting.

PR #3824, opened 2022-02-04
dockerfile: ci: rpm: install nodejs instead of NPM directly

KKoukiou:
To solve some dependency issues that appeared on the CI recently.

This will install npm anyway.

https://developer.fedoraproject.org/tech/languages/nodejs/nodejs.html


jkonecny12 (approved):
LGTM!

PR #3822, opened 2022-02-03
Webui destructive tests support

KKoukiou:
No need to review yet, I want to look at it again tomorrow with a clear head. Feel free of course if you want.

TODO:
* [x] Update test/common tag to fetch https://github.com/cockpit-project/cockpit/pull/16914
* [x] Fix the pylint issues

PR #3826, opened 2022-02-04
Improve docs for web ui (#docs)

VladimirSlavik:
A lot of things doesn't exactly work right now, so just a draft. Ideally there would be containers everywhere they can.

### DIFF
--- a/ui/webui/README.rst
+++ b/ui/webui/README.rst
@@ -8,8 +8,43 @@ Make sure you have `npm` available (usually from your distribution package).
 These commands check out the source and build it into the `dist/` directory::
 
     git clone https://github.com/rhinstaller/anaconda.git
+    cd anaconda
+    ./autogen.sh && ./configure
     cd ui/webui
     make
+
+Note that this builds only the WebUI code. In most cases, the rest of Anaconda is needed too.
+
+Running the UI interactively
+----------------------------
+
+First, build RPM packages in the CI container, from root of the repository::
+
+    make -f ./Makefile.am anaconda-ci-build
+    make -f ./Makefile.am container-shell
+    ./autogen.sh && ./configure
+    make rpms
+
+Alternatively, you can install all the dependencies and build without using the container::
+
+    sudo ./scripts/testing/install_dependencies.sh -y
+    ./autogen.sh && ./configure
+    make rpms
+
+Note that if you went through this process already, you also need to remove the whole `ui/webui/bots/` directory before calling `make rpms`. Another pitfall is that the `bots` directory is a git repository and will not be removed with `git clean`.
+
+Then, prepare the `~/.ssh/config` file according to instructions in `<test/README.rst>`_
+
+Finally, set up the VM and start it::
+
+    cd ui/webui
+    make ../../updates.img
+    mkdir -p test/images
+    make bots
+    ./bots/image-download fedora-rawhide-boot
+    ./test/webui_testvm.py fedora-rawhide-boot
+
+Once the machine is running, instructions for connecting to it will be printed.
 
 Running eslint
 --------------
--- a/ui/webui/test/README.rst
+++ b/ui/webui/test/README.rst
@@ -13,9 +13,13 @@ Preparation and general invocation
 To run the WebUI integration tests run the following from the root of the anaconda repo.
 (do NOT run the integration tests as root)::
 
-You first need to build anaconda RPMS::
+You first need to build anaconda RPMS in the repository root::
 
+    ./autogen.sh
+    ./configure
     make rpms
+
+Note that if you went through this process already, you also need to remove the whole `ui/webui/bots/` directory before calling `make rpms`. Another pitfall is that the `bots` directory is a git repository and will not be removed with `git clean`.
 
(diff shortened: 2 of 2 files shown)


KKoukiou:
Lets block this on https://github.com/rhinstaller/anaconda/pull/3822/files

As it massive touches the docs, and many scripts moved, as for example the `make vm-run` target.

VladimirSlavik:
Updated and rebased.

ui/webui/README.rst
```diff
@@ -8,9 +8,44 @@ Make sure you have `npm` available (usually from your distribution package).
...
     make
 
+Note that this builds only the WebUI code. In most cases, the rest of Anaconda is needed too.
+
+Running the UI interactively
+----------------------------
+
+First, build RPM packages in the CI container, from root of the repository::
+
+    make -f ./Makefile.am anaconda-ci-build
```
KKoukiou, line 23: Instead of adding all these container and make commands here could we just link to https://github.com/rhinstaller/anaconda/blob/master/CONTRIBUTING.rst#how-to-run-make-commands ?

ui/webui/README.rst
```diff
@@ -8,9 +8,44 @@ Make sure you have `npm` available (usually from your distribution package).
...
+    ./autogen.sh && ./configure
+    make rpms
+
+Alternatively, you can install all the dependencies and build without using the container::
+
+    sudo ./scripts/testing/install_dependencies.sh -y
+    ./autogen.sh && ./configure
+    make rpms
+
+Note that if you went through this process already, you also need to remove the whole `ui/webui/bots/` directory before calling `make rpms`. Another pitfall is that the `bots` directory is a git repository and will not be removed with `git clean`.
```
KKoukiou, line 34: That's actually not true.

You don't need to remove `bots` in every run, in fact that's extremely inefficient.

Removing the `bots` was needed in your scenario, because I had just introduced the new image for `anaconda-webui` tests (`fedora-rawhide-boot` image). In the older `bots` checkout the image was not present, so it would just give `Image does not exist` or similar error. Now that the image exists, having an older checkout of the `bots`, just means that you could possibly run your tests against an older `rawhide boot.iso`, which is not problematic. The CI always will use the newest image, as the bots there is always created.

Of course there is the chance that the newer anaconda code does not work any more with older ISO, in that case, it will become pretty intuitive to realize that the `bots` need rebase, in order to use the new image.

ui/webui/README.rst
```diff
@@ -8,9 +8,44 @@ Make sure you have `npm` available (usually from your distribution package).
...
+
+Then, prepare the `~/.ssh/config` file according to instructions in `<test/README.rst>`_
+
+Finally, set up the VM and start it::
+
+    cd ui/webui
+    make ../../updates.img
+    mkdir -p test/images
```
KKoukiou, line 42: This is not needed, (it was needed last week indeed :)) Now there are saved actually our of the repo, see more https://github.com/cockpit-project/bots/#image-location for information about where images are stored.

ui/webui/README.rst
```diff
@@ -8,9 +8,44 @@ Make sure you have `npm` available (usually from your distribution package).
...
+
+Alternatively, you can install all the dependencies and build without using the container::
+
+    sudo ./scripts/testing/install_dependencies.sh -y
+    ./autogen.sh && ./configure
+    make rpms
+
+Note that if you went through this process already, you also need to remove the whole `ui/webui/bots/` directory before calling `make rpms`. Another pitfall is that the `bots` directory is a git repository and will not be removed with `git clean`.
+
+Then, prepare the `~/.ssh/config` file according to instructions in `<test/README.rst>`_
```
KKoukiou, line 36: This is not needed for running the `./test/webui_testvm.py fedora-rawhide-boot` command, or anything else here. 
This is needed for the `make rsync`.

ui/webui/test/README.rst
```diff
@@ -13,10 +13,14 @@ Preparation and general invocation
 To run the WebUI integration tests run the following from the root of the anaconda repo.
 (do NOT run the integration tests as root)::
 
-You first need to build anaconda RPMS::
+You first need to build anaconda RPMS in the repository root::
 
+    ./autogen.sh
```
KKoukiou, line 18: Instead I would point to https://github.com/rhinstaller/anaconda/blob/master/CONTRIBUTING.rst#how-to-run-make-commands

ui/webui/test/README.rst
```diff
@@ -134,6 +138,12 @@ connect to the test VMs by typing `ssh test-updates`::
         Hostname 127.0.0.2
         Port 2201
         User root
+        StrictHostKeyChecking=no
+        UserKnownHostsFile=/dev/null
+
+If you are creating the file, you must also set correct permissions or SSH will not start::
+
+    chmod 600 ~/.ssh/config
```
KKoukiou, line 146: I never had this problem. Do you think it's something specific to your machine?

jkonecny12: This shouldn't be required the .ssh and content should always be 600 otherwise it's a system issue.

VladimirSlavik:
Closing in favor of #4015

PR #3827, opened 2022-02-04
Improve branching documentation for Fedora

jkonecny12:
It's not simple thing to do but hopefully it's understandable this way.

### DIFF
--- a/docs/release.rst
+++ b/docs/release.rst
@@ -267,19 +267,15 @@ Create the ``-release`` branch:
     git pull
     git checkout -b f<version>-release
 
+
+Adjust Fedora specific configuration
+____________________________________
+
 Switch to f<version>-release branch for Fedora specific settings:
 
 ::
 
    git checkout f<version>-release
-
-Edit branch specific settings. This have to be done on f<version>-release branch only:
-
-::
-
-   vim ./branch-config.mk
-
-And change content according to comments in the file.
 
 
 Then correct pykickstart version for the new Fedora release by changing all occurrences of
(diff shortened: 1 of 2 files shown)


poncovka (approved):
Thanks!

docs/release.rst
```diff
@@ -310,6 +342,61 @@ Check if everything is correctly set:
...
+
+::
+
+   git log   # to find SHA of the configuration commit
+   git revert <SHA>  # please add commit message why you are doing the revert
+
+The change revert has to be done if these branch specific configuration changes. Otherwise we won't
+be testing Rawhide on master.
+
+Adjust configuration on master
```
VladimirSlavik, line 370: For the other branches you explicitly mention `git checkout ...`, so master should have this too.

docs/release.rst
```diff
@@ -302,6 +298,42 @@ in the Fedora version specific branch on Anaconda side.
...
+
+Adjust shared development branch
+________________________________
+
+
+Switch to f<version>-release branch for Fedora specific settings:
+
+::
+
+   git checkout f<version>-devel
```
VladimirSlavik, line 327: (Here too.)

VladimirSlavik (commented):
Thank you, this is very useful. However, I think the structure needs some tweaks for consistency.

Apart from these comments, the FIXME seems to be resolved by this change, I think?

jkonecny12:
Hmm, I'm more and more inclined that we should improve this instead of documenting it...

jkonecny12:
Hi @VladimirSlavik, feel free to take this over. Just for explanation, I'm not moving the documentation to the code to reduce the documentation but the main benefit should be that it won't get outdated so easily.

PR #3829, opened 2022-02-07
Don't use progressQ in tasks of the installation queue

poncovka:
Call `_progress_cb` to report installation progress of DBus tasks.
Set up this callback in the `pyanaconda.installation` module.


jkonecny12 (approved):
LGTM!

PR #3830, opened 2022-02-07
Support Btrfs-only mount points in the default partitioning

poncovka:
The `default_partitioning` configuration option now allows to specify mount
points that will be created only for the Btrfs scheme. Use the `btrfs` attribute.


Conan-Kudo (approved):

jkonecny12 (approved):
Looks great to me. Thanks!

PR #3831, opened 2022-02-07
dockerfile: stop specifying nodejs stream explicitely

KKoukiou:
This works with nodejs:16, which is default stream in f34, f35, rawhide.

Also move the installation line above with the rest, there is no need for special
treatment here.


poncovka (approved):
Thanks!

PR #3833, opened 2022-02-07
Do not fail on nonexistent fs nodes in pstore

VladimirSlavik:
The bug was introduced while removing `dir_tree_map`, in commit
8d3aa25785fe63824c3c45ae0aa756647276fda7

Related: [rhbz#2050715](https://bugzilla.redhat.com/show_bug.cgi?id=2050715)

### DIFF
--- a/pyanaconda/startup_utils.py
+++ b/pyanaconda/startup_utils.py
@@ -291,8 +291,10 @@ def clean_pstore():
     """
     for (base, dirs, files) in os.walk("/sys/fs/pstore"):  # pylint: disable=unused-variable
         for file in files:
-            os.unlink(os.path.join(base, file))
-
+            try:
+                os.unlink(os.path.join(base, file))
+            except OSError:
+                pass
 
 def print_startup_note(options):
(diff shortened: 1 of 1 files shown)


VladimirSlavik:
Looks like the f36 branch does not run tests to merge into master...

poncovka (approved):
Thank you!

poncovka:
@VladimirSlavik, have you checked other places that used to call `dir_tree_map`?

jkonecny12 (approved):
Looks good to me.

I think I saw answer to Vendy question on JIRA. In general other parts should be fine.

VladimirSlavik:
> @VladimirSlavik, have you checked other places that used to call `dir_tree_map`?

Yes, it was used in the chown-like thing that is now actual chown, so that does not apply anymore.

VladimirSlavik:
Since I am not sure how to test this, I created an updates image and attached to the BZ.

https://bugzilla.redhat.com/show_bug.cgi?id=2050715#c7

PR #3835, opened 2022-02-08
Create a task for running the current installation queue

poncovka:
Run the `RunInstallationTask` task to complete the installation.


pyanaconda/installation.py
```diff
@@ -53,355 +53,377 @@ def _writeKS(ksdata):
...
+        """Wait for background processing threads to finish.
+
+        Wait for background processing threads to finish before filling
+        the installation task queue. Otherwise installation tasks might
+        be created based on old data in DBus modules, missing data set by
+        the threads that are still running.
+        """
+
+        # This should be the only thread running, wait for the others to finish if not.
+        if threadMgr.running > 1:
```
jkonecny12, line 274: I guess that this code have to change in the future. But not something for this PR. Although, even now it will not work for Storage thread.

jkonecny12 (approved):
Looks good to me.

pyanaconda/installation.py
poncovka: Definitely. However, the Storage thread should be fine. We are still using UI threads for everything. It is one the areas that we will have to look into very soon. The plan for this task is to run it via DBus from the Web UI (with some limitations), but  GUI and TUI will still run it directly (to avoid the limitations). There is too much work to be done.

PR #3839, opened 2022-02-09
Remove undocumented installed scripts

poncovka:
Remove the following scripts:

* `/usr/bin/analog`
* `/usr/bin/restart-anaconda`

They are included in RPMs.
They are not documented.
We don't use them.

**TODO:**
- [x] Remove `move usr/share/anaconda/restart-anaconda usr/bin` from Lorax templates.
- [x] Write release notes.

**Depends on:** https://github.com/weldr/lorax/pull/1221


poncovka:
@jstodola, are you aware of any reasons why we should keep these scripts?

jstodola (approved):
I cannot think of any reasonable reason to keep the files if they are not useful. QE doesn't use them.

VladimirSlavik:
CC @t184256 maybe of interest to you?

t184256:
I don't know what these are/were.

jkonecny12:
@AdamWill I would like to make sure that you don't use these too? Honestly I'm not sure about analog script, that seems like something created for testing?

jkonecny12:
Also I would like to ask everyone about removal of the `restart-anaconda` because I was looking for exactly this. I think it could improve our current developer experience when you need to implement something. Basically you can have restarting Anaconda similar to what web UI development process have (reloading page). I wasn't aware that we have something like that and the question is if it is working but maybe we want to use that?

VladimirSlavik:
Yes, IIRC somebody asked about restarting. But the point that we do not use it is valid. Maybe we can have that in docs instead? Does that even make sense?

t184256:
[reanaconda](https://github.com/rhinstaller/devel-tools/blob/master/reanaconda/reanaconda.py) has a limited applicability with its QEMU dependency, but either qualifies or overlaps. Wanna make it watch inotify and reload when the sources change like to make it even more web-appish? =)

poncovka:
> Also I would like to ask everyone about removal of the `restart-anaconda` because I was looking for exactly this. I think it could improve our current developer experience when you need to implement something. Basically you can have restarting Anaconda similar to what web UI development process have (reloading page). I wasn't aware that we have something like that and the question is if it is working but maybe we want to use that?

We can always revive the file and include it in an updates image once we figure out how to use it for development, but I don't think we should ever install scripts for development on the boot.iso. Users might be tempted to use them.

AdamWill:
Yeah, we're not using either of these. Thanks for checking.

jkonecny12 (approved):
Looks good to me now. Thanks for a good points everyone.

jkonecny12:
> [reanaconda](https://github.com/rhinstaller/devel-tools/blob/master/reanaconda/reanaconda.py) has a limited applicability with its QEMU dependency, but either qualifies or overlaps. Wanna make it watch inotify and reload when the sources change like to make it even more web-appish? =)

The reanaconda is quite similar approach but I'm taking this as something more universal and just for the development. The reanaconda still works in the VM created by the script. This could be applicable in general.

jkonecny12:
The depends PR is merged, is there another reason why is this blocked?

poncovka:
> The depends PR is merged, is there another reason why is this blocked?

Well, it still needs to be released.

PR #3840, opened 2022-02-09
Fix the root configuration in the user interface

poncovka:
Ported from https://github.com/rhinstaller/anaconda/pull/3775.

### DIFF
--- a/pyanaconda/ui/gui/spokes/root_password.py
+++ b/pyanaconda/ui/gui/spokes/root_password.py
@@ -29,8 +29,7 @@ from pyanaconda.ui.gui.helpers import GUISpokeInputCheckHandler
 from pyanaconda.ui.gui.utils import set_password_visibility
 from pyanaconda.ui.common import FirstbootSpokeMixIn
 from pyanaconda.ui.communication import hubQ
-from pyanaconda.ui.lib.services import is_reconfiguration_mode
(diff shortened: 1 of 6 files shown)


pyanaconda/ui/gui/spokes/root_password.py
```diff
@@ -179,24 +179,26 @@ def mandatory(self):
...
+            # Unlock the root account.
+            self._users_module.SetRootAccountLocked(False)
 
-        if enabled:
+        else:
+            # Reset the root password.
+            self._users_module.ClearRootPassword()
+
+            # Lock the root account.
+            self._users_module.SetRootAccountLocked(True)
```
jkonecny12, line 194: Do we want to Lock the root if the root account is enabled but password is not set? I'm just not sure if that is the correct behavior :).

jkonecny12 (approved):
Looks good to me otherwise.

VladimirSlavik (approved):
Thank you!

pyanaconda/ui/gui/spokes/root_password.py
VladimirSlavik: I agree that porting it 1:1 is correct.

jkonecny12: Thanks for explanation.

PR #3841, opened 2022-02-09
Restore contexts also in /usr/lib

VladimirSlavik:
New RPM stores its database there. See also:
https://fedoraproject.org/wiki/Changes/RelocateRPMToUsr

Resolves: [rhbz#2052038](https://bugzilla.redhat.com/show_bug.cgi?id=2052038)

### DIFF
--- a/pyanaconda/modules/boss/installation.py
+++ b/pyanaconda/modules/boss/installation.py
@@ -244,6 +244,7 @@ class SetContextsTask(Task):
             "/etc",
             "/lib64",
             "/root",
+            "/usr/lib",
             "/usr/lib64",
             "/var/cache/yum",
             "/var/home",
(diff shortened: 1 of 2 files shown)


VladimirSlavik:
CC @t184256 maybe of interest to you?

M4rtinK (changes requested):
Looks good code wise but commit message says /var/lib while code has /usr/lib - I think that should be fixed. :)

Also I also think this should go to F36 as well as the need for this fix is most likely triggered by the following F36 change: https://fedoraproject.org/wiki/Changes/RelocateRPMToUsr

(Also I think it might be good to link to the change from the commit message as well.)

rvykydal (approved):
Looks good to me, thank you.

jkonecny12 (approved):
Looks good to me.

M4rtinK (approved):
Looks good to me, thanks! :)

PR #3842, opened 2022-02-09
Remove the support for detection of unsupported hardware

poncovka:
The detection was implemented for RHEL 8 and immediately disabled. There is some
discussion going on about this functionality, but it is very unlikely that we
will be able to enable this code again in its current form.

The removed configuration options `can_detect_unsupported_hardware` and
`can_detect_support_removed` were never documented in the RHEL documentation.

Related: rhbh#1645686


data/profile.d/rhel.conf
```diff
@@ -9,11 +9,6 @@ profile_id = rhel
 os_id = rhel
 
 [Installation System]
-# The detection is disabled since #1645686.
-# can_detect_unsupported_hardware = True
-# can_detect_support_removed = True
-
-# Show a warning if SMT is enabled.
 can_detect_enabled_smt = True
```
VladimirSlavik, line 12: I think you want to keep the last comment line?

VladimirSlavik (approved):
Looks good to me. Thank you!

data/profile.d/rhel.conf
poncovka: I don't. But I can add it back if it bothers you.

jkonecny12 (approved):
Thanks for this Vendy and we have to try hard to avoid something like this in the future.

PR #3845, opened 2022-02-09
implement /var subvolume for Fedora Silverblue and Kinoite

cmurf:
This implements Fedora change https://fedoraproject.org/wiki/Changes/VarSubvol4SilverblueKinoite, and depends on https://github.com/rhinstaller/anaconda/pull/3830


data/profile.d/fedora-kinoite.conf
```diff
@@ -9,3 +9,11 @@ base_profile = fedora-kde
 # Match os-release values.
 os_id = fedora
 variant_id = kinoite
+
+[Storage]
+default_scheme = BTRFS
+btrfs_compression = zstd:1
+default_partitioning =
```
poncovka, line 16: You are redefining `default_partitioning` from `anaconda.conf`. It looks like this:

```
default_partitioning =
    /     (min 1 GiB, max 70 GiB)
    /home (min 500 MiB, free 50 GiB)
```

I guess you want to keep the previous settings, so `fedora-kinoite` and `fedora-silverblue` should probably define something like this:

```
default_partitioning =
    /     (min 1 GiB, max 70 GiB)
    /home (min 500 MiB, free 50 GiB)
    /var  (btrfs)
```

data/profile.d/fedora-kinoite.conf
```diff
@@ -9,3 +9,11 @@ base_profile = fedora-kde
...
+    /
+    /var (btrfs)
```
poncovka, line 18: You will have to update the `ProfileConfigurationTestCase.test_fedora_profiles` test and replace `WORKSTATION_PARTITIONING` with a new list of partitions.

poncovka (changes requested):
I will change the base to `f36-devel` to get the fix to F36 and F37. Please, rebase your PR.

data/profile.d/fedora-kinoite.conf
cmurf: Thinking of future usage: is it possible for `/` and `/home` to have no attribute, resulting in the min/max/free values being inherited from anaconda.conf? That way any updates to min/max/free values in anaconda.conf just past through, and we don't have derivatives ending up with stale settings?

VladimirSlavik:
Please rebase on f36-devel...

data/profile.d/fedora-kinoite.conf
poncovka: > Thinking of future usage: is it possible for `/` and `/home` to have no attribute, resulting in the min/max/free values being inherited from anaconda.conf?

We have no control over that. We just generate a list of configuration files and give this list to `configparser`. The inheritance is very primitive and we would like to keep it that way.

> That way any updates to min/max/free values in anaconda.conf just past through, and we don't have derivatives ending up with stale settings?

We use unit tests to check this kind of dependencies between profiles.

tests/unit_tests/pyanaconda_tests/core/test_profile.py
```diff
@@ -69,6 +69,36 @@
     ),
 ]
 
+WORKSTATIONPLUS_PARTITIONING = [
```
poncovka, line 72: You can do something like `WORKSTATION_PARTITIONING + [ ... ]` to make sure that `/` and `/home` are same.

VladimirSlavik:
Replaced by #3872.

PR #3846, opened 2022-02-10
UX: clarify meaning of "additional space" checkbox

hexagonrecursion:
The phrase "I would like to make additional space available" is ambiguous. It could be interpreted in two ways:
1. I would like to delete existing partitions to reclaim space (the intended meaning)
2. I would like the automatic partitioning to leave some unused space (e.g. because I plan to install something else later)

I wanted to leave some unused space. I checked the box and was surprised when the result occupied my entire hard drive. This change removes the ambiguity.

### DIFF
--- a/pyanaconda/ui/gui/spokes/storage.glade
+++ b/pyanaconda/ui/gui/spokes/storage.glade
@@ -430,7 +430,7 @@
                                     <property name="reveal_child">True</property>
                                     <child>
                                       <object class="GtkCheckButton" id="reclaimCheckbox">
-                                        <property name="label" translatable="yes" context="GUI|Storage">I would like to _make additional space available.</property>
+                                        <property name="label" translatable="yes" context="GUI|Storage">Free up space by re_moving or shrinking existing partitions</property>
                                         <property name="visible">True</property>
                                         <property name="can_focus">True</property>
                                         <property name="receives_default">False</property>


poncovka (commented):
Btw. the space can be reclaimed also by shrinking existing partitions, but I don't know if that's relevant. @VladimirSlavik, what do you think about this pull request?

jstodola:
> Does checking the box allow me to shrink existing partitions too? I only noticed the deletion feature.

Yes, but it depends on the file system, some file systems can be shrunk - `ext{2,3,4}`, some not - `xfs`.

VladimirSlavik:
@hexagonrecursion, thank you!

I think this makes sense, but could further. I'm not @mairin, but I think the following could work:
- I feel the wording should be action-oriented, same as the encryption below, eg. "Free up space by removing or shrinking existing partitions".
- It would be very nice to keep same the accelerator `m` - needs to be `_m` somewhere in the text.
- Optional - Same as encryption, add a gray label along the lines "You will select partitions to manipulate next".
- Also optional - Any more detailed explanation could go into the pop-over hint, such as which filesystems can be shrunk, which not etc. I'm not sure if people would see it, but who knows.

Is this too ambitious?

hexagonrecursion:
> * Optional - Same as encryption, add a gray label along the lines "You will select partitions to manipulate next".
> 
>     * Also optional - Any more detailed explanation could go into the pop-over hint, such as which filesystems can be shrunk, which not etc. I'm not sure if people would see it, but who knows.
> 
> 
> Is this too ambitious?

For me - yes. I would have to set up your dev environment and figuring out how things are done in your project - not worth the effort for such a minor change.

jkonecny12:
Looks great to me but the commits should be definitely squashed together.

hexagonrecursion:
> squashed

Done

VladimirSlavik (approved):
Looks good to me. Thank you!

PR #3863, opened 2022-02-11
Ignore deps-dev commits by dependabot

VladimirSlavik:
The deps-dev commits are for development-only dependencies and not relevant for users, so they should not be in the changelog.

----

Please advise:
- should this be `#infra` or not?
- should I put this on f36 branch too?


KKoukiou (approved):
Thanks, not sure how I can verify this one.

jkonecny12:
To answer your question we don't need this on F36 because the code is not there.

PR #3868, opened 2022-02-15
Backport hotfix for the npm issue to anaconda-ci container (#infra)

jkonecny12:
We don't need that because we don't build the Anaconda in the anaconda-ci container, however, blivet using our containers for their testing and they are doing that. So or so, this will also fix the `container-shell` Makefile target in the similar cases.

The original fix is done by aae34587bc246d31eff2b229f40b1ccc87adb9b5 .

Resolves https://github.com/storaged-project/blivet/pull/1020 .

### DIFF
--- a/dockerfile/anaconda-ci/Dockerfile
+++ b/dockerfile/anaconda-ci/Dockerfile
@@ -68,6 +68,8 @@ RUN set -ex; \
   cat /root/anaconda.spec.in | sed 's/@PACKAGE_VERSION@/0/; s/@PACKAGE_RELEASE@/0/; s/%{__python3}/python3/' > /tmp/anaconda.spec; \
   rpmspec -q --buildrequires /tmp/anaconda.spec | xargs -d '\n' dnf install -y; \
   rpmspec -q --requires /tmp/anaconda.spec | grep -v anaconda | xargs -d '\n' dnf install -y; \
+  # FIXME: Install older npm version - the version from the package manager leads to timeouts
+  npm install -g npm@8.1.0; \
   dnf clean all
 
 RUN pip install --no-cache-dir --upgrade pip; \


poncovka (approved):
Thanks!

KKoukiou (commented):
Are they aware that they can just build anaconda without webui? That's what they will need for now anyway, In that case that case we don't need to add npm that we don't need in this container.

jkonecny12:
They do not know that but the question is do we want to tell them to do this? If we want to have this in master than it give sense to me that they want to test against it to know that the new UI code won't break their tests in the future, don't we?

jkonecny12:
@vojtechtrefny the question is more on you, do you want to have the npm download in the tests? We don't plan to ship this in Fedora anytime soon. However, if you want I can create a PR on your workflow to disable that.

KKoukiou:
> They do not know that but the question is do we want to tell them to do this? If we want to have this in master than it give sense to me that they want to test against it to know that the new UI code won't break their tests in the future, don't we?

I don't see how the new webui code can break any external tests. It can't affected anything, until we enable the webui by default.

vojtechtrefny:
> @vojtechtrefny the question is more on you, do you want to have the npm download in the tests? We don't plan to ship this in Fedora anytime soon. However, if you want I can create a PR on your workflow to disable that.

I think we probably shouldn't complicate this a just keep it as close to the anaconda upstream as possible, if we introduce some exceptions now, we'll probably forget to remove them in the future. It doesn't really matter the web UI isn't in Fedora, if it's in Anaconda upstream repo we should just use that.

VladimirSlavik (approved):

jkonecny12:
@vojtechtrefny after the new container-ci is created the issue on your side should be resolved:
https://github.com/rhinstaller/anaconda/actions/runs/1848015761

Feel free to ping me if not.

vojtechtrefny:
> @vojtechtrefny after the new container-ci is created the issue on your side should be resolved: https://github.com/rhinstaller/anaconda/actions/runs/1848015761
> 
> Feel free to ping me if not.

The test passed on my new PR today, thank you for the fix.

PR #3871, opened 2022-02-16

rvykydal:
Resolves: rhbz#2000140

Support for inst.ks.device has never been implemented while the ksdevice

We will try to remove also support for this ksdevice functionality, so
it would make no sense to make inst.ks.device working at this point.


rvykydal:
/kickstart-test --testtype network-device-default-ksdevice-httpks

jkonecny12 (approved):
Thanks for the backport!

VladimirSlavik (approved):
Thank you!

rvykydal:
/kickstart-test --testtype network-device-default-ksdevice-httpks

poncovka (approved):
Thanks!

PR #3872, opened 2022-02-16
Port of cmurf PR for /var subvolume on Kinoite and Silverblue

VladimirSlavik:
Fixed and cleaned up #3845.

@cmurf please let me know if the config does specify sufficient size for the /var mount point. I saw you had 15 GB in your tests, which was not in the config.

### DIFF
--- a/data/profile.d/fedora-kinoite.conf
+++ b/data/profile.d/fedora-kinoite.conf
@@ -9,3 +9,9 @@ base_profile = fedora-kde
 # Match os-release values.
 os_id = fedora
 variant_id = kinoite
+
+[Storage]
+default_partitioning =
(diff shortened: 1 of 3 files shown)


VladimirSlavik:
The `f36-release` container is missing because of missing `pykickstart` support for f36, which is why that one test fails.

Conan-Kudo (approved):

jkonecny12 (approved):
Looks good to me!

VladimirSlavik:
@cmurf sorry for the delays. It seems we will be able to address this on Monday.

poncovka (approved):
Thanks!

VladimirSlavik:
I tested the same config on a regular server boot.iso. It worked as intended, another subvolume was created for /var.

PR #3873, opened 2022-02-16
PartSpec debugging improvements

VladimirSlavik:
These changes were required to debug #3872 - pytest printed only a cryptic diff with pointers, which was totally useless, as was printing the class, because that missed some of the data.

I think it's useful to have this, especially as configs/profiles will likely stay with us for some time.

The two commits are independent.


Conan-Kudo (approved):

M4rtinK (approved):
Looks good to me as well. :)

poncovka (approved):
Thanks!

PR #3880, opened 2022-02-21
webui: checkout last release instead of main for cockpit's test library

KKoukiou:
Main now already contains code that breaks our tests, let's adjust tests
when a new cockpit release is available.

** This fixes the failing webui tests on master.

Example test failure on master:
https://logs.cockpit-project.org/logs/pull-3807-20220221-151618-6df42d11-fedora-rawhide-boot/log.html


M4rtinK (approved):
Looks good to me.

jkonecny12 (approved):
Looks good to me. I just wonder how should we keep up with the versions? Isn't there something like latest for the non API breaking releases?

VladimirSlavik (approved):

PR #3881, opened 2022-02-22
Use RHEL 9.0 repositories for tests (#infra)

rvykydal:
The rhel-9 branch is used for RHEL 9.0 development while RHEL-9
repositories point to RHEL-9.1 now.

### DIFF
--- a/dockerfile/anaconda-ci/rhel-9.repo
+++ b/dockerfile/anaconda-ci/rhel-9.repo
@@ -1,6 +1,6 @@
 [RHEL-9-CRB]
 name=crb
-baseurl=http://download.devel.redhat.com/rhel-9/nightly/RHEL-9/latest-RHEL-9/compose/CRB/$basearch/os/
+baseurl=http://download.devel.redhat.com/rhel-9/nightly/RHEL-9/latest-RHEL-9.0/compose/CRB/$basearch/os/
 enabled=1
 gpgcheck=0
 install_weak_deps=0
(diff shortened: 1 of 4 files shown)


jstodola (approved):
It looks OK to me. Just two notes:

- The same change should be done in the rhel-8 branch to keep the branches consistent.
- It should be documented that the symlinks need to be changed for the future releases (and when to do it)

poncovka (approved):
Looks good to me. Thanks!

jkonecny12:
> 

Hi, hmm, are those `.X` branches really beneficial to us? Isn't it just adding more work to ourselves without a benefit? Maybe I'm just missing something here.

jkonecny12 (approved):
LGTM!

rvykydal:
> > 
> 
> Hi, are those `.X` branches really beneficial to us? Isn't it just adding more work to ourselves without a benefit? Maybe I'm just missing something here.

It depends on the real development. I'd create the .Y branch only when we would diverge (for example development of a feature in 9.1 vs fix in stabilization of 9.0 which can happen in parallel). Also sharing the rhel-9 branch too long showed up to require hacks, mainly due to rhbz# references which were different for X.Y, X.Y+1 (automatically cloned).
So for now I'd just update the rhel-9 branch but the branch is actually a development branch for rhel-9 (ie 9.1 for now or soon) so we need to keep it in mind / track it.

M4rtinK (approved):

jkonecny12:
I'm also not sure what have to be updated after branching but I guess, let's do the branching and find that out... :(

rvykydal:
> It looks OK to me. Just two notes:
> 
>     * The same change should be done in the rhel-8 branch to keep the branches consistent.
https://github.com/rhinstaller/anaconda/pull/3893
> 
>     * It should be documented that the symlinks need to be changed for the future releases (and when to do it)
Definitely, I need to figure out how and where.

PR #3883, opened 2022-02-22
Don't report the name of the DBus task by default

poncovka:
The task shouldn't report anything by default, so we can have a total control
over the output. Otherwise, we would have to use workarounds to skip the first
progress message.

### DIFF
--- a/pyanaconda/modules/common/task/task.py
+++ b/pyanaconda/modules/common/task/task.py
@@ -105,7 +105,7 @@ class Task(AbstractTask):
             log.info("'%s' is canceled.", self.name)
             return
 
-        self.report_progress(self.name, step_number=1)
+        log.info(self.name)
         self._set_result(self.run())
 
     def _task_succeeded_callback(self):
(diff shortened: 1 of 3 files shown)


VladimirSlavik (commented):
How does this affect what we see about the queue in UI?

poncovka:
> How does this affect what we see about the queue in UI?

We might lose some initial messages from the payload tasks. Otherwise, there should be no difference. We had a workaround that discarded the first message of all DBus tasks anyway, so they were never propagated to the progress screen.

pyanaconda/modules/common/task/task.py
```diff
@@ -105,7 +105,7 @@ def _task_run_callback(self):
             log.info("'%s' is canceled.", self.name)
             return
 
-        self.report_progress(self.name, step_number=1)
+        log.info(self.name)
```
VladimirSlavik, line 108: I am wondering, what will that do? Just a random "Setting up disks" in the logs?

It's not really important, though.

VladimirSlavik (approved):
Thank you for the explanation. The rest looks good to me.

pyanaconda/modules/common/task/task.py
poncovka: I think it provides a nice context by default. I will keep it there for now. We can always changed it.

name

PR #3886, opened 2022-02-22
Remove progressQ

poncovka:
Remove the support for the progress queue.
Use progress reporting of the new task instead.


pyanaconda/ui/tui/spokes/installation_progress.py
```diff
@@ -61,74 +59,58 @@ def completed(self):
...
+        task.progress_changed_signal.connect(
+            self._on_progress_changed
+        )
         task.start()
 
-        # This will run until we're all done with the install thread.
-        self._update_progress()
+        # This will run until the task is finished.
+        while task.is_running:
+            sleep(1)
```
poncovka, line 100: @jkonecny12, is it possible implement a solution without the sleep? Does `simpleline` support something like that?

pyanaconda/ui/tui/spokes/installation_progress.py
```diff
@@ -61,74 +59,58 @@ def completed(self):
...
+
+            # Also flush the communication Queue at least once a second and
+            # process it's events so we can react to async evens (like a thread
+            # throwing an exception)
+            loop = App.get_event_loop()
+            loop.process_signals()
```
VladimirSlavik, line 106: Is this a safe thing to do?

poncovka: It is nothing new. You can find the same lines in the previous solution.

pyanaconda/ui/tui/spokes/installation_progress.py
jkonecny12: You need to wait on something so if you have a source you can use signals if I'm not mistaken. Do we have something like that?

jkonecny12: There is `return_after` parameter for this method which will end when a specific signal arrive so if you have that it should be pretty straightforward.

jkonecny12 (approved):
Looks good to me.

pyanaconda/ui/tui/spokes/installation_progress.py
poncovka: Thanks! I have used it.

poncovka:
I have updated the "[Don't use progressQ in TUI](https://github.com/rhinstaller/anaconda/pull/3886/commits/050a2865c77424861240dbcd6a38ed7d04cdee29)" commit.

jkonecny12 (approved):
I like these improvements. Thanks!

PR #3887, opened 2022-02-23
Use cache when fetching organization data from the GUI

M4rtinK:
```
Use cache when fetching organization data from the GUI

Bug 2044258 is caused by GetOrgs() call at registration time
via RetrieveOrganizationsTask triggering MultipleOrganizationsError
- the user is member of more than one organization yet no organization
has been specified.

This fails the registration attempt, which triggers cleanup tasks,
such as un-provisioning from Satellite.

As Satellite is the only known environment where users that are
members of multiple organizations exist, this causes an issue when
the GetOrgs() calling DBus task is called again from the GUI.

The system will no longer be provisioned for Satellite at this point &
the GetOrgs() call will fail - it will go to Hosted Canlepin, which
knowns nothing about this user.

The end result is the behavior described in the bug - the GUI fails to
list organizations for the user to choose from.

To fix this issue, we now always cache the org data list returned by
GetOrgs() when called via RetrieveOrganizationsTask at registration time.
For this the task is called directly, not via its DBus interface and
we make sure the cache is always cleared before calling GetOrgs().

The only case where RetrieveOrganizationsTask is called via it's DBus interface
(RetrieveOrganizationsWithTask) is when GUI needs to list organizations after
handling MultipleOrganizationsError. In this case the existing cached
value will be used when the GetOrgs() call fails due to Satellite being
no longer available.

This way if GUI calls RetrieveOrganizationsWithTask() after handling
the MultipleOrganizationsError, it will get the cached data from the
initial GetOrgs() call, which triggered the MultipleOrganizationsError.

Resolves: rhbz#2044258
```

### DIFF
--- a/pyanaconda/modules/subscription/runtime.py
+++ b/pyanaconda/modules/subscription/runtime.py
@@ -269,7 +269,8 @@ class RegisterWithUsernamePasswordTask(Task):
             org_data_task = RetrieveOrganizationsTask(
                 rhsm_register_server_proxy=self._rhsm_register_server_proxy,
                 username=self._username,
-                password=self._password
+                password=self._password,
+                reset_cache=True
             )
             org_list = org_data_task.run()
             if len(org_list) > 1:
@@ -1033,17 +1034,23 @@ class RetrieveOrganizationsTask(Task):
     Satellite instances.
     """
 
-    def __init__(self, rhsm_register_server_proxy, username, password):
+    # the cache is used to serve last-known-good data if calling the GetOrgs()
+    # DBus method can't be called successfully in some scenarios
+    _org_data_list_cache = []
+
+    def __init__(self, rhsm_register_server_proxy, username, password, reset_cache=False):
         """Create a new organization data parsing task.
 
         :param rhsm_register_server_proxy: DBus proxy for the RHSM RegisterServer object
         :param str username: Red Hat account username
         :param str password: Red Hat account password
+        :param bool reset_cache: clear the cache before calling GetOrgs()
         """
         super().__init__()
         self._rhsm_register_server_proxy = rhsm_register_server_proxy
         self._username = username
         self._password = password
+        self._reset_cache = reset_cache
 
     @property
     def name(self):
@@ -1084,6 +1091,9 @@ class RetrieveOrganizationsTask(Task):
 
         :raises: RegistrationError if calling the RHSM DBus API returns an error
         """
+        # reset the data cache if requested
+        if self._reset_cache:
+            RetrieveOrganizationsTask._org_data_list_cache = []
         log.debug("subscription: getting data about organizations")
         with RHSMPrivateBus(self._rhsm_register_server_proxy) as private_bus:
             try:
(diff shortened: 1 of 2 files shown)


pyanaconda/modules/subscription/runtime.py
```diff
@@ -265,13 +269,15 @@ def run(self):
             # an organization id being specified RHSM will not know what organization to register
             # the machine. In this throw raise a specific exception so that the GUI can react
             # accordingly and help the user fix the issue.
-
             org_data_task = RetrieveOrganizationsTask(
```
poncovka, line 272: I haven't realized you always use the same task class to retrieve the organizations. It would be better to keep the cached list as a class attribute of `RetrieveOrganizationsTask`, so you don't have to propagate it back and forth. The task could have the `reset_cache=False` argument that you could set here to True (as a replacement of `cached_org_data_list=None`).

pyanaconda/modules/subscription/runtime.py
```diff
@@ -1084,6 +1100,11 @@ def run(self):
 
         :raises: RegistrationError if calling the RHSM DBus API returns an error
         """
+        # check if we can just return cached data
+        if self._cached_org_data_list:
```
poncovka, line 1104: It seems dangerous to always prefer the cache without trying to get the actual organizations. I wouldn't trust this logic. The cache should be used as only as a fallback when you fail to retrieve the data.

poncovka (commented):
The fix can be implemented in less invasive way.

pyanaconda/modules/subscription/runtime.py
M4rtinK: Yeah, that should work and will reduce the amount of boilerplate - just wondering, do we already use something like that for other tasks ? I'm thinking it could be a surprise where the data is coming from later on if this is a one-off.

pyanaconda/modules/subscription/runtime.py
jkonecny12: Would these really change that often. I'm not sure but this seems to me like a reasonable solution?

M4rtinK:
Updated & simplified to use class variable - thanks to @poncovka for the suggestion! :)

pyanaconda/modules/subscription/runtime.py
```diff
@@ -1033,17 +1034,23 @@ class RetrieveOrganizationsTask(Task):
     Satellite instances.
     """
 
-    def __init__(self, rhsm_register_server_proxy, username, password):
+    # the cache is used to serve last-known-good data if calling the GetOrgs()
+    # DBus method can't be called successfully in some scenarios
+    org_data_list_cache = []
```
poncovka, line 1039: It is missing a `_`.

M4rtinK: Fixed.

jkonecny12 (approved):
Looks good to me.

poncovka (approved):

M4rtinK:
This can be ported to Fedora only once Satellite support has been ported.

PR #3889, opened 2022-02-23
webui: Don't save SSH key in command from VM script

VladimirSlavik:
Otherwise the key of the temporary system is saved in known_hosts and forces editing the file when the VM is started anew.


jkonecny12 (approved):
Looks good to me. Thanks!

KKoukiou (approved):
Cockpit has it here https://github.com/cockpit-project/cockpit/tree/main/test#helpful-tips 

It's independent though. Thanks!

PR #3891, opened 2022-02-23
master: move /var/tmp to its own partition

sandrobonazzola:
Move /var/tmp to its own partition on oVirt and RHV for DISA-STIG support.

Bug-Url: https://bugzilla.redhat.com/show_bug.cgi?id=2057475
Signed-off-by: Sandro Bonazzola [sbonazzo@redhat.com](mailto:sbonazzo@redhat.com)

### DIFF
--- a/data/profile.d/ovirt.conf
+++ b/data/profile.d/ovirt.conf
@@ -16,18 +16,20 @@ default_partitioning =
     /              (min 6 GiB)
     /home          (size 1 GiB)
     /tmp           (size 1 GiB)
-    /var           (size 15 GiB)
+    /var           (size 5 GiB)
     /var/crash     (size 10 GiB)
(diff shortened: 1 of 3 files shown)


sandrobonazzola:
@VladimirSlavik this is not really impacting f37 as oVirt and RHV images are not built on top of Fedora, This is going to be relevant only on next rebase for el8 / el9.

VladimirSlavik:
Thanks! Yes, but we want the PRs to have a clear indication where they will land, so the history is useful when checking bugs with the code etc.

sandrobonazzola:
Ok. I think there are a couple of test to be manually triggered here as well

poncovka (approved):
Looks good to me. Thank you!

PR #3897, opened 2022-02-24
Remove the enable_ignore_broken_packages configuration option

poncovka:
The `%packages --ignorebroken` feature is not supported on RHEL. Handle
this use case in pykickstart, so we don't have to deal with it in anaconda.

The `enable_ignore_broken_packages` configuration option was never documented
in the RHEL documentation.

Tested on F37 and RHEL9.

**Depends on:** https://github.com/pykickstart/pykickstart/pull/403

### DIFF
--- a/anaconda.spec.in
+++ b/anaconda.spec.in
@@ -38,7 +38,7 @@ Source0: https://github.com/rhinstaller/%{name}/releases/download/%{name}-%{vers
 %define libxklavierver 5.4
 %define mehver 0.23-1
 %define nmver 1.0
-%define pykickstartver 3.32-1
+%define pykickstartver 3.37-1
 %define pypartedver 2.5-2
 %define pythonblivetver 1:3.4.0-1
 %define rpmver 4.15.0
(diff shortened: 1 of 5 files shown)


VladimirSlavik (approved):
Looks good to me. Thank you!

jkonecny12 (approved):
I really like this solution but I guess we need to backport this to REHL-9 too to avoid conflict with the pyckickstart change?
Brian already requested exception to get the pykickstart part to 9.0. Wouldn't there be i collision with our current implementation on `rhel-9`?

poncovka:
> I really like this solution but I guess we need to backport this to REHL-9 too to avoid conflict with the pyckickstart change? Brian already requested exception to get the pykickstart part to 9.0. Wouldn't there be i collision with our current implementation on `rhel-9`?

I don't think we need to backport it to RHEL 9 if we don't want to. Pykickstart should fail before anaconda.

poncovka:
Rebased.

PR #3898, opened 2022-02-24
Add missing npm dependency to the install_dependencies

jkonecny12:
This will enable us to use toolbox containers pretty easily for the new UI
development.

Just do:

toolbox create anaconda
toolbox enter anaconda
sudo ./scripts/testing/install_dependencies.sh

Then you have everything you need.

### DIFF
--- a/scripts/testing/install_dependencies.sh
+++ b/scripts/testing/install_dependencies.sh
@@ -28,7 +28,7 @@
 
 set -eu
 
-dnf install $@ rpm-build /usr/bin/xargs
+dnf install $@ make rpm-build /usr/bin/xargs npm virt-install
 
 TEMP=$(mktemp /tmp/anaconda.spec.XXXXXXX)
 


VladimirSlavik (changes requested):
- I think `npm` is needed to build, so it should go into the spec.
- `make` also seems suspicious to me. How come it's not pulled in via autotools?
- The libvirt stuff is probably ok as it's a testing dependency.

jkonecny12:
> * I think `npm` is needed to build, so it should go into the spec.

@KKoukiou what do you think? 

>     * `make` also seems suspicious to me. How come it's not pulled in via autotools?

The autotools is not dependency of the Anaconda. These tools are in the buildroot by default so you don't have to have these as dependencies in the spec file.

scripts/testing/install_dependencies.sh
```diff
@@ -28,7 +28,7 @@
 
 set -eu
 
-dnf install $@ rpm-build /usr/bin/xargs
+dnf install $@ make rpm-build /usr/bin/xargs npm libvirt virt-install
```
KKoukiou, line 31: Isn't libvirt direct dependency of virt-install? 

If not, libvirt-daemon-kvm and libvirt-client are sufficient. Libvirt is a metapackage and installs a lot of uneeded things.

VladimirSlavik (approved):
Oh, sorry, that makes sense. The `spec` is second build, first we need the tarball contents.

scripts/testing/install_dependencies.sh
jkonecny12: I did not know about the libvirt-client, in that case yes, virt-install has a dependency on libvirt-client.

jkonecny12: Fixed.

jkonecny12:
**UPDATED:**
* Removed libvirt, it's dragged in by virt-install.

jkonecny12:
Rebased

poncovka (approved):

KKoukiou (approved):
the commit message just mentions npm, you add two more. But feel free to merge as is.

jkonecny12:
**UPDATED:**
* Rebased
* Change the commit message

PR #3900, opened 2022-02-25
Fix icon names to use -symbolic

VladimirSlavik:
The non-symbolic icons are no longer available.

CC @jkonecny12


jkonecny12 (changes requested):
I'm trying to also adjust the size of the icons because they are really ugly on some places but that can be done after this is resolved. Thanks.

However, rebase this to f36-devel. Based on container builds it's broken there too.

VladimirSlavik:
The point of this was to keep somewhere the list of changes - I think I found more places with icons than the tests know.

VladimirSlavik:
Yes! Thank you, that means I can use your PR...

PR #3901, opened 2022-02-25
Fix anaconda-widget crash when loaded on Wayland

jkonecny12:
The `GDK_DISPLAY_XDISPLAY` function will return `NULL` on Wayland. That will than crash when put into the `xkl_engine_get_instance()` function. To avoid the crash let's open the display here directly.

Just for clarification this is not fixing usability on Wayland system, just that it's not crashing. It still not work correctly.

*Resolves: rhbz#1959521*
*Suggested-by: Ray Strode*

**TODO:**
- [ ] Increase version of the anaconda-widgets
- [ ] Test this manually on Live
- [ ] Test this manually on netinst

### DIFF
--- a/widgets/src/LayoutIndicator.c
+++ b/widgets/src/LayoutIndicator.c
@@ -172,8 +172,7 @@ static void anaconda_layout_indicator_init(AnacondaLayoutIndicator *self) {
            class-wide stuff */
 
         /* initialize XklEngine instance that will be used by all LayoutIndicator instances */
-        display = gdk_display_get_default();
-        klass->engine = xkl_engine_get_instance(GDK_DISPLAY_XDISPLAY(display));
+        klass->engine = xkl_engine_get_instance(XOpenDisplay(NULL));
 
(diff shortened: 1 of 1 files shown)


M4rtinK (approved):
Looks good to me! :)

And thanks a lot to Jiri, Vlada & Halfline for tracking this down! :)

VladimirSlavik:
The away this is written, it seems to be as if Anaconda will never again run under Xorg. Is that really true?

jkonecny12:
> The away this is written, it seems to be as if Anaconda will never again run under Xorg. Is that really true?

I have to test this but this way it should work the same way as before. It should work on Xorg as on Wayland.

poncovka (approved):
Thanks!

jkonecny12:
Finally tested in the boot.iso and unfortunately it doesn't work. There is complete disconnect between Anaconda configuration and the widget.

Not a priority to spent more time on that now :(. Closing for now, we can play with that in the future if there will be a time.

PR #3902, opened 2022-02-25
Replace legacy adwaita icons removed in adwaita-icon-theme 42

AdamWill:
adwaita-icon-theme 42 removes a lot of the 'legacy' fullcolor
icons (but not all of them). This replaces all the removed names
with their symbolic equivalents. Even without this change GTK+
automatically gives us the symbolic icons where the non-symbolic
one is missing, but cleaning up the names seems good.

### DIFF
--- a/pyanaconda/ui/gui/helpers.py
+++ b/pyanaconda/ui/gui/helpers.py
@@ -155,7 +155,7 @@ class GUIDialogInputCheckHandler(GUIInputCheckHandler, metaclass=ABCMeta):
             inputcheck.input_obj.set_icon_tooltip_text(Gtk.EntryIconPosition.SECONDARY, "")
         else:
             inputcheck.input_obj.set_icon_from_icon_name(Gtk.EntryIconPosition.SECONDARY,
-                    "dialog-error")
+                    "dialog-error-symbolic")
             inputcheck.input_obj.set_icon_tooltip_text(Gtk.EntryIconPosition.SECONDARY,
                 inputcheck.check_status)
 
--- a/pyanaconda/ui/gui/main.glade
+++ b/pyanaconda/ui/gui/main.glade
@@ -135,7 +135,7 @@
               <object class="GtkImage" id="image1">
                 <property name="visible">True</property>
                 <property name="can_focus">False</property>
-                <property name="icon_name">application-exit</property>
+                <property name="icon_name">application-exit-symbolic</property>
                 <property name="icon_size">6</property>
               </object>
               <packing>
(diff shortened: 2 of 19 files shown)


jkonecny12:
Already done here https://github.com/rhinstaller/anaconda/pull/3900

AdamWill:
You missed one.

jkonecny12:
This should go to f36-devel branch. IIRC this happens also on Fedora 36.

jkonecny12:
Anyway, I think it's too late to do the build of Anaconda. This will go in on Monday.

jkonecny12:
Also I played with the size a bit because the new icons looks ugly on some places, but that should go as follow-up PR.

jkonecny12:
I see that you are also changing widgets, we should than correctly bump the version, but naah, let's go with the easy way here and I'll bump the version later.

M4rtinK:
> Also I played with the size a bit because the new icons looks ugly on some places, but that should go as follow-up PR.

Would it be possible to attach a screenshot showing the difference ? Here on the other PR. Thanks in advance! :)

jkonecny12:
Not really, it's too many places. Too time consuming. I'll do that in separate PR.

jkonecny12:
The fedora-rawhide-bot is not really important here, feel free to ignore it.

VladimirSlavik (approved):
Thank you for being more thorough!

VladimirSlavik:
@jkonecny12 I think the widgets version does not really matter at all here.

It will be bumped with the next release, all at once, right? The way I see it, we're doing it together because we're already grepping. But this fixes primarily the tests, which prevent other infra stuff from updating. As Adam said, the actual running code has some compatibility so even if none of these changes are there, it works. So the fix in widgets does not require the version bump.

Of course, writing this little essay led me to the the idea that the widgets version might differ somewhere so it might need bumping... And it does not. So I believe this PR is enough as-is.

poncovka (approved):
Thank you!

VladimirSlavik:
Just to keep the data here: Jirka says the widgets version needs to be changed somewhere in their configure: https://github.com/rhinstaller/anaconda/blob/master/widgets/configure.ac

jkonecny12:
Thanks everyone for your help. Merging this now to unblock stuff. 

I also contacted the Desktop team and they told me to use symbolic in most of the places. The only place where the color icons should stay is the large sizes and if we need them back we should copy them here. Honestly not a fan of that, anyway I will look on the icons what I can do with that to improve the current status.

PR #3905, opened 2022-02-28
Replace one more icon after removal from adwaita

jkonecny12:
This one was missed because it doesn't exists on Rawhide. Let's just fix the ~build~ tests quickly, no-one see this icon anyway.


M4rtinK (approved):
Looks good to me, thanks! :)

M4rtinK:
BTW, as far as I can tell we are under Beta freeze, so this will need a blocker/FE on the bug.

jkonecny12:
I would rather wait after the beta freeze. It's not blocking build.

jkonecny12:
Fix the bad description.

VladimirSlavik (approved):

PR #3879, opened 2022-02-21
build: clean up webui dist before making scratch builds

KKoukiou:
If `ui/webui/dist` is present, the `make dist` command will be skipped
for the webui subtree. This is especially problematic, if the
translation files were not present when the `ui/webui/dist` was created
in the first place, as then RPM creation will always fail on missing
translation JS files.

PR #3906, opened 2022-02-28
Move the po files download to the `make` call (#infra)

jkonecny12:
Right now if user will call:

`make && make rpms`

it will fail because webui needs to have translations and `make rpms` won't create them. To make this more optimized let's download the translations on the `make` call if they don't present already.

To do that use flag file, which will be created when the files are downloaded and removed on clean.

This is alternative solution to https://github.com/rhinstaller/anaconda/pull/3879 .

### DIFF
--- a/po/.gitignore
+++ b/po/.gitignore
@@ -1,2 +1,3 @@
 *.mo
 *.po.new
+.po_downloaded
--- a/Makefile.am
+++ b/Makefile.am
@@ -53,11 +53,6 @@ ARCHIVE_TAG   = $(PACKAGE_NAME)-$(PACKAGE_VERSION)-$(PACKAGE_RELEASE)
 # Set this to "true" if you want to have SRPM archive with test version
 TEST_BUILD	?= "false"
 
-# LOCALIZATION SETTINGS
-L10N_REPOSITORY ?= https://github.com/rhinstaller/anaconda-l10n.git
-# Branch used in anaconda-l10n repository. This should be master all the time.
-GIT_L10N_BRANCH ?= master
-
 RC_RELEASE ?= $(shell date -u +0.1.%Y%m%d%H%M%S)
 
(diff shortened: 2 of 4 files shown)


KKoukiou (approved):
I like it more than my solution.

jkonecny12:
**UPDATED:**
* Fix issue that `make release` without previous `make` call did not packaged '.po' files.

jkonecny12:
The code changed, asking for re-review.

po/.gitignore
```diff
@@ -1,2 +1,3 @@
 *.mo
 *.po.new
+.downloaded
```
VladimirSlavik, line 3: It's now `.po-downloaded` and one level up.

VladimirSlavik (commented):
Almost there! The ignore no longer matches the file location. Otherwise LGTM.

po/.gitignore
jkonecny12: Fixed.

jkonecny12:
Trying to find out why packit created srpm are broken.

VladimirSlavik (approved):

KKoukiou (approved):

KKoukiou (commented):
It think you can now drop this hack https://github.com/rhinstaller/anaconda/blob/master/anaconda.spec.in#L345

jkonecny12:
ok, if we encounter some issues we can put that back later.

jkonecny12:
**UPDATED:**
* Request localization all the time during the build.

@M4rtinK do you agree with this change?

jkonecny12:
The packit issues seems to be specific to the packit environment and race condition. Will be probably solved on the Packit side anyway because they are moving SRPM creation to the COPR.

M4rtinK (approved):
Looks good to me, thanks! :)

PR #3908, opened 2022-03-01
Fix the infrastructure check reporting

jkonecny12:
**DO NOT MERGE - This is just a testing right now**

Here is some issue which I'm not able reproduce on my fork. Trying to resolve it here.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -428,3 +428,4 @@ coverage-report:
 	$(COVERAGE) report -m --include "pyanaconda/dbus*,pyanaconda/modules/*,pyanaconda/core/*" \
 								     > tests/coverage-modular.log
 	@cat tests/coverage-report.log
+# aaa


VladimirSlavik (approved):
Thank you, looks good to me.

.github/workflows/infra-check.yml
```diff
@@ -1,7 +1,7 @@
 # Check if only infrastructure files are changed when infrastructure label is set.
 name: infrastructure-check
 on:
-  pull_request:
+  pull_request_target:
```
poncovka, line 4: Shouldn't we prefer `pull_request` for good reasons? I have a problem to find a good source that would back you up. Could you share some links with us, please?

jkonecny12: Here: https://docs.github.com/en/actions/security-guides/automatic-token-authentication#modifying-the-permissions-for-the-github_token

```
You can use the permissions key to add and remove read permissions for forked repositories, but typically you can't grant write access.
```

With combination of here: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflows-in-forked-repositories

And yes, it shouldn't be used because you can inject your code to the PR which then could be executed. However, you can't really change the workflows which are executed because these are executed from the target branch (not your PR).
That means that this check is fine. This workflow doesn't execute any code from the PR, it just invokes git commands and compare status of these. Another thing is that someone has to approve external contributor test run before the workflow is executed.
I think we should be safe here.

poncovka: Thank you!

poncovka (approved):

PR #3909, opened 2022-03-01
tests: webui: Increate timeout for accessible webui to 5 minutes

KKoukiou:
This will solve the 'Anaconda webui is not reachable' failure in the CI.

The previous timeout for the webui to become rechable in the test VM was 50seconds,
and was often leading to timeout failures.

### DIFF
--- a/ui/webui/test/machine_install.py
+++ b/ui/webui/test/machine_install.py
@@ -89,7 +89,7 @@ class VirtInstallMachine(VirtMachine):
                 f"--location {os.getcwd()}/bots/images/{self.image}"
             )
 
-            for _ in range(10):
+            for _ in range(30):
                 try:
                     requests.get(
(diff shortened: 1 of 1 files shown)


jkonecny12 (approved):
LGTM!

KKoukiou:
The infra tests are wrongly failing. @VladimirSlavik can you take a look?

VladimirSlavik:
I "fixed" it, but not really.

----
@jkonecny12 I re-ran the check, and it was the same. Then I added label because that's another trigger. That seems to have fixed that condition, somehow. But I have no idea why. Maybe the previous event was not triggered correctly, and re-running just repeated the same wrong condition?

jkonecny12:
Don't really know how to help here and also it's merged now, is it resolved?

PR #3910, opened 2022-03-01
Provide defaults for the Web UI installation

poncovka:
* Set up the partitioning before starting the Web UI.
* Set up basic error handling for the Web UI tracebacks.

### DIFF
--- a/data/interactive-defaults.ks
+++ b/data/interactive-defaults.ks
@@ -6,3 +6,5 @@
 #   This file is deprecated and will be removed in in the future.
 #   Use the Anaconda configuration files instead.
 #
+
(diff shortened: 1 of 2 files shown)


poncovka:
@KKoukiou, it needs more testing, but it should be pretty close.

poncovka:
@M4rtinK, @VladimirSlavik, I have set up some basic error handling for exceptions that are raised before the Web UI is started. It works well, but the tty stays at 6, so it has to be manually switched to 1 to see the error messages. It would be nice to switch the tty automatically in the error handler. Could you look at it, please? We won't be able to use the error handler in the Web UI anyway.

KKoukiou:
Closing in favor of https://github.com/rhinstaller/anaconda/pull/3918

PR #3912, opened 2022-03-02
Detach keyboard configuration of Anaconda from Live environment

jkonecny12:
This will hide the layout indicator and detach control of the keyboard configuration between Live environment and Anaconda. 

* Keyboard settings set in the Anaconda is not reflected to Live and just used after the installation.
* If users want to use another keyboard layout to write user names and passwords they need to set the keyboard manually in the Live environment.

This resolves issue reported in this [bug](https://bugzilla.redhat.com/show_bug.cgi?id=2016613) but also make easier to users to write passwords by keyboard layout different to what will be available after the installation. Could result in unusable system.

_Resolves: rhbz#2016613_

**WIP - not finished yet**

anaconda_live_change

### DIFF
--- a/pyanaconda/ui/gui/__init__.py
+++ b/pyanaconda/ui/gui/__init__.py
@@ -24,7 +24,7 @@ from contextlib import contextmanager
 import gi
 gi.require_version("Gdk", "3.0")
 gi.require_version("Gtk", "3.0")
-gi.require_version("AnacondaWidgets", "3.3")
+gi.require_version("AnacondaWidgets", "3.4")
 gi.require_version("Keybinder", "3.0")
 gi.require_version("GdkPixbuf", "2.0")
 gi.require_version("GObject", "2.0")
--- a/widgets/configure.ac
+++ b/widgets/configure.ac
@@ -20,7 +20,7 @@
 m4_define(python_required_version, 3.4)
 
 AC_PREREQ([2.63])
-AC_INIT([AnacondaWidgets], [3.3], [clumens@redhat.com])
+AC_INIT([AnacondaWidgets], [3.4], [clumens@redhat.com])
 
 # Disable building static libraries.
 # This needs to be set before initializing automake
--- a/pyanaconda/ui/gui/hubs/__init__.py
+++ b/pyanaconda/ui/gui/hubs/__init__.py
@@ -115,7 +115,7 @@ class Hub(GUIObject, common.Hub):
         import gi
 
         gi.require_version("Gtk", "3.0")
-        gi.require_version("AnacondaWidgets", "3.3")
+        gi.require_version("AnacondaWidgets", "3.4")
 
         from gi.repository import Gtk, AnacondaWidgets
 
--- a/pyanaconda/ui/gui/spokes/lib/accordion.py
+++ b/pyanaconda/ui/gui/spokes/lib/accordion.py
@@ -24,7 +24,7 @@ from pyanaconda.core.storage import get_supported_autopart_choices
 from pyanaconda.ui.gui.utils import escape_markup, really_hide, really_show
 
 import gi
-gi.require_version("AnacondaWidgets", "3.3")
+gi.require_version("AnacondaWidgets", "3.4")
 gi.require_version("Gtk", "3.0")
 from gi.repository import Gtk, AnacondaWidgets
 
--- a/pyanaconda/ui/gui/spokes/storage.py
+++ b/pyanaconda/ui/gui/spokes/storage.py
@@ -58,7 +58,7 @@ from pyanaconda.ui.gui.spokes.lib.storage_dialogs import NeedSpaceDialog, NoSpac
 import gi
 gi.require_version("Gtk", "3.0")
 gi.require_version("Gdk", "3.0")
-gi.require_version("AnacondaWidgets", "3.3")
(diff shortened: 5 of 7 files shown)


pyanaconda/ui/gui/__init__.py
```diff
@@ -159,6 +159,15 @@ class in the action list instead of going on to
 
         self.builder.connect_signals(self)
 
+        # Hide keyboard indicator if we can't configure the keyboard
+        # It doesn't really give you any benefit of seeing something which could
+        # give you wrong values.
+        if (not conf.system.can_configure_keyboard) and \
+           isinstance(self.window, AnacondaWidgets.BaseWindow):
```
poncovka, line 166: This is sad, but I don't see a better option at this moment. Could you please specify in the comment, that this needs to be applied to every hub and spoke, but dialogs have to be skipped?

widgets/configure.ac
```diff
@@ -20,7 +20,7 @@
 m4_define(python_required_version, 3.4)
 
 AC_PREREQ([2.63])
-AC_INIT([AnacondaWidgets], [3.3], [clumens@redhat.com])
+AC_INIT([AnacondaWidgets], [3.4], [clumens@redhat.com])
```
poncovka, line 23: We have `gi.require_version("AnacondaWidgets", "3.3")` in the code.

pyanaconda/core/configuration/system.py
```diff
@@ -125,7 +131,7 @@ def can_activate_layouts(self):
     @property
     def can_configure_keyboard(self):
         """Can we configure the keyboard?"""
-        return self._is_boot_iso or self._is_live_os or self._is_booted_os
+        return (not self._is_wayland) and (self._is_boot_iso or self._is_live_os or self._is_booted_os)
```
poncovka, line 134: I guess the proper solution would be to expose `can_configure_keyboard` in the configuration files and set it up in the `AnacondaConfiguration.set_from_opts` method if that's possible. The config would stay simple and stable. However, the current solution seems to be good enough as a temporary workaround.

jkonecny12: Not sure I understand you, I don't plan to use opts here. I want to detect Wayland and that is not doable from the opts.

AdamWill:
What's the status here?

pyanaconda/ui/gui/__init__.py
jkonecny12: Fixed

widgets/configure.ac
jkonecny12: Fixed.

pyanaconda/core/configuration/system.py
jkonecny12: I wonder wouldn't it be better to instead of changing the can_configure_keyboard in set_from_opts add is_wayland?

jkonecny12:
> What's the status here?

Sorry for missing it so long. Working on that right now. Hopefully we will have it ready soon.

jkonecny12:
**UPDATED** and ready for review

VladimirSlavik (approved):
Looks good to me. Thank you!

jkonecny12:
I tested this Change on KDE Live, Gnome Shell Live, Xfce Live and netinst, seems to be working correctly.

pyanaconda/core/configuration/system.py
```diff
@@ -122,11 +122,6 @@ def can_activate_layouts(self):
         """
         return self._is_boot_iso or self._is_booted_os
 
-    @property
-    def can_configure_keyboard(self):
```
poncovka, line 126: We should keep the property. It is still useful.

pyanaconda/keyboard.py
```diff
@@ -55,6 +56,39 @@ class InvalidLayoutVariantSpec(Exception):
...
+
+
+def can_configure_keyboard():
+    """Can we configure the keyboard?
+
+    TODO: The is_wayland logic is not part of the configuration so we would have to add it to the
+    configuration otherwise it won't be accessible in the Anaconda modules.
+    This is more a temporary workaround which will be replaced in the future.
+    """
+    if _is_xwayland() and conf.system._is_live_os:
```
poncovka, line 86: I guess we could have something like `conf.system.can_run_wayland` to avoid the private property.

poncovka: It is cheaper to check these conditions in reverse order.

pyanaconda/ui/gui/__init__.py
```diff
@@ -159,6 +160,16 @@ class in the action list instead of going on to
 
         self.builder.connect_signals(self)
 
+        # Hide keyboard indicator if we can't configure the keyboard
+        # It doesn't really give you any benefit of seeing something which could
+        # give you wrong values.
+        # This has to be applied to every spoke and hub - we have to ignore dialog and other
+        # non full screen parts.
+        if (not can_configure_keyboard()) and isinstance(self.window, AnacondaWidgets.BaseWindow):
```
poncovka, line 168: Please, remove the redundant parentheses. They are confusing.

poncovka (approved):
I have some suggestions. Otherwise, it looks good to me.

poncovka:
Updated and tested.

M4rtinK (approved):

PR #3913, opened 2022-03-02
webui: Don't check SSH key in command from VM script

VladimirSlavik:
Addendum to #3889. This makes it so the command just gives the inst. env. root prompt immediately, no confirmations, no typing, nothing.

----

@KKoukiou , there is one thing that might need clarifying before this is merged or instead dropped:

> Cockpit has it here https://github.com/cockpit-project/cockpit/tree/main/test#helpful-tips

I tried that and it did not work for me, unless I hardcode in that config the string that comes out of the VM script... And that is a variable in the script. I am not sure what that means - is the address a variable or a constant?

### DIFF
--- a/ui/webui/test/webui_testvm.py
+++ b/ui/webui/test/webui_testvm.py
@@ -32,7 +32,7 @@ def cmd_cli():
 
     print("You can connect to the VM in the following ways:")
     # print ssh command
-    print("ssh -o ControlPath=%s -o UserKnownHostsFile=/dev/null -p %s %s@%s" %
+    print("ssh -o ControlPath=%s -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no -p %s %s@%s" %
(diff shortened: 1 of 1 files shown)


M4rtinK (approved):
Looks good to me, nice improvement, thanks a lot! :)

KKoukiou:
@VladimirSlavik the exact copy of ssh/config from cockpit, won't currently work as we rely on the test VM being resolved to `test-updates`.

I  meant to rather add only that `StrictHostKeyChecking=no` line in the section for the ssh config, in https://github.com/rhinstaller/anaconda/tree/master/ui/webui/test#helpful-tips instread of the webui_testvm script as you did here. The reason is that, the tests won't use use this script for creating VMs, but you still might want to ssh to a tests, created by the tests.

VladimirSlavik:
I think we should merge this to get rid of it, and maybe later look at harmonizing the command with the cockpit SSH readme. Perhaps the messages should link there instead. There is a lot to set up for SSH to work flawlessly.

PR #3914, opened 2022-03-02
webui: implement top level wizard layout

M4rtinK:
This is a **very** early draft PR for a PatternFly wizard based top level design of the Web UI.
Just 4 screens at the moment:

Screen Shot 2022-03-09 at 15 21 56

Screen Shot 2022-03-09 at 15 22 12

Screen Shot 2022-03-09 at 15 22 19

Screen Shot 2022-03-09 at 15 22 22

Screen Shot 2022-03-09 at 15 22 25

Mostly mockup texts and sometimes buttons, but some screens, such as Installation language & Storage configuration do work with real data from the Anaconda backend obtained via its DBus interface.

Definitely not final strings and names of things as well.

**TODO**

- [x] test coverage
- [x] check if we can remove more things that are no longer needed


poncovka:
Installation progress and Installation complete shouldn't be steps in a wizzard. You don't want people to change the configuration after we start the installation. I think it should work like https://www.patternfly.org/v4/components/wizard#finished.

ui/webui/src/app.jsx
```diff
@@ -18,11 +18,19 @@ import cockpit from "cockpit";
...
 
 import {
-    Page,
+    Page, Wizard,
 } from "@patternfly/react-core";
 
 import { InstallationLanguage } from "./InstallationLanguage.jsx";
-import { Summary } from "./Summary.jsx";
+
+import { InstallationDestination } from "./storage/InstallationDestination.jsx";
```
KKoukiou, line 26: Please not line breaks between imports of the same type. 

I usually do it like this:

First import external libraries, see React.
new line
Then import externally defined components, see react-core
new line
Then import locally defined components
new line

ui/webui/src/app.jsx
```diff
@@ -33,15 +41,60 @@ export const Application = () => {
 
     useEffect(() => cockpit.file("/run/anaconda/bus.address").watch(setAddress), []);
 
+    const steps = [
+        {
+            id: 0,
+            name: "Installation language",
```
KKoukiou, line 47: Please translate the step names. You can do it like: 
_("Installation language")

ui/webui/src/app.jsx
```diff
@@ -33,15 +41,60 @@ export const Application = () => {
...
+            component: <AddressContext.Provider value={address}>
+                <InstallationLanguage />
+            </AddressContext.Provider>
+        },
+        {
+            id: 1,
```
KKoukiou, line 53: Where did you copy this from? Since that the id property here is not consumed by Patternfly.

ui/webui/src/app.jsx
```diff
@@ -33,15 +41,60 @@ export const Application = () => {
...
-            <AddressContext.Provider value={address}>
-                <Summary />
-            </AddressContext.Provider>}
+            <Wizard
+              navAriaLabel={`${title} steps`}
+              mainAriaLabel={`${title} content`}
+              titleId="wiz-modal-demo-title"
+              descriptionId="wiz-modal-demo-description"
+              title="Fedora Rawhide installation"
+              description="PRE-RELEASE/TESTING"
```
KKoukiou, line 94: Patternfly does not translate strings by itself. For this reason they expose through their properties the labels, so you can overwrite the english text.
For example here you need: 
``` 
nextButtonText={_("Next")}
backButtonText... 

etc
```

Find here https://www.patternfly.org/v4/components/wizard/#wizard all string properties that are labels, aria-labels or button texts.

ui/webui/src/app.jsx
```diff
@@ -33,15 +41,60 @@ export const Application = () => {
...
-                {!path.length > 0 && <InstallationLanguage />}
-            </AddressContext.Provider>
-            {path.length > 0 &&
-            <AddressContext.Provider value={address}>
-                <Summary />
-            </AddressContext.Provider>}
+            <Wizard
+              navAriaLabel={`${title} steps`}
+              mainAriaLabel={`${title} content`}
+              titleId="wiz-modal-demo-title"
```
KKoukiou, line 91: I would not use 'demo' here. Let's assume that ids won't change too much over time, it's pain, as tests will need to be adjusted as well.

KKoukiou (changes requested):

KKoukiou (commented):
Please drop the Header component and also stop using the PageSection wrapper. (That creates that grey background)

So essentially for all remaining files:
``` diff
[kkoukiou@sequioa webui]$ git diff
diff --git ui/webui/src/storage/InstallationDestination.jsx ui/webui/src/storage/InstallationDestination.jsx
index 01c422a41e..4e6ad3a611 100644
--- ui/webui/src/storage/InstallationDestination.jsx
+++ ui/webui/src/storage/InstallationDestination.jsx
@@ -22,10 +22,9 @@ import {
     DataList, DataListItem, DataListItemRow, DataListCheck, DataListItemCells, DataListCell,
     Form,
     Hint, HintBody,
-    PageSection
 
-import { AddressContext, Header } from "../Common.jsx";
+import { AddressContext } from "../Common.jsx";
 import { useEvent, useObject } from "hooks";
 
 const _ = cockpit.gettext;
@@ -155,21 +154,13 @@ export const InstallationDestination = () => {
     };
 
     return (
-        <>
-            <Header
-              done={onDoneClicked}
-              title={_("Installation destination")}
-            />
-            <PageSection>
-                <Form isHorizontal>
-                    <Hint>
-                        <HintBody>
-                            {_("Select the device(s) you would like to install to. They will be left untouched until you click on the main menu's 'Begin installation' button.")}
-                        </HintBody>
-                    </Hint>
-                    <LocalStandardDisks />
-                </Form>
-            </PageSection>
-        </>
+        <Form isHorizontal>
+            <Hint>
+                <HintBody>
+                    {_("Select the device(s) you would like to install to. They will be left untouched until you click on the main menu's 'Begin installation' button.")}
+                </HintBody>
+            </Hint>
+            <LocalStandardDisks />
+        </Form>
     );
 };

````

File: ui/webui/src/app.jsx [hunk 6]
```diff
@@ -33,15 +41,60 @@ export const Application = () => {
...
+            id: 0,
+            name: "Installation language",
+            component: <AddressContext.Provider value={address}>
+                <InstallationLanguage />
+            </AddressContext.Provider>
+        },
+        {
+            id: 1,
+            name: "Storage configuration",
+            component: <AddressContext.Provider value={address}>
```
KKoukiou, line 55: Please avoid code duplication and use this as a wrapper function. 

I will write the diffs for start, because I think JS syntax can be intimidating, but let me know if you prefer to let you explore it all the way

```
+    const wrapWithContext = children => {
+        return (
+            <AddressContext.Provider value={address}>
+                {children}
+            </AddressContext.Provider>
+        );
+    };
+
     const steps = [
         {
             id: 0,
             name: "Installation language",
-            component: <AddressContext.Provider value={address}>
-                <InstallationLanguage />
-            </AddressContext.Provider>
+            component: wrapWithContext(<InstallationLanguage />)
         },

....
```

KKoukiou (commented):
The Select from the installation language which can escape out of the modal can be solved with: adding the following property:
  menuAppendTo={document.body}

ui/webui/src/app.jsx
KKoukiou: Following up to our pm discussion https://github.com/patternfly/patternfly-react/issues/6993

lclay2:
Question: if you use the close button (X) in the black header area what happens?

ui/webui/src/app.jsx
M4rtinK: ACK, fixed.

ui/webui/src/app.jsx
M4rtinK: Added.

ui/webui/src/app.jsx
M4rtinK: I took the numeric ids from: https://www.patternfly.org/v4/components/wizard/react-demos#in-modal

ui/webui/src/app.jsx
M4rtinK: Adding there.

ui/webui/src/app.jsx
M4rtinK: Oh, right, that comes from the PatternFly example, I'll change it to something more sane. :)

ui/webui/src/app.jsx
M4rtinK: Much better, thanks, implemented! :)

M4rtinK:
> Installation progress and Installation complete shouldn't be steps in a wizzard. You don't want people to change the configuration after we start the installation. I think it should work like https://www.patternfly.org/v4/components/wizard#finished.

Yeah, definitely once you start the installation it makes no sense to go back. And it is nice that there is already a pre-prepared way of handling this we might be able to use - thanks for finding it! :)

Do we want to add it to this PR or maybe rather to a follow-up one ?

M4rtinK:
> Thats because Martin is using the in-wizard modal like here https://www.patternfly.org/v4/components/wizard/react-demos#in-modal, we should use instead the next in-page, as we are not in a modal that we can just close.

Sure, that the modal one was the first one found & was able to get working. ;-)

Looking at it now the in-page wizard looks like a good match. :)

ui/webui/test/check-basic
```diff
@@ -35,15 +35,10 @@ class TestBasic(MachineCase):
...
 
-        b.click('#continue-btn')
-        b.expect_load()
-
-        for page in ['keyboard', 'language', 'timedate', 'installation-source', 'software-selection', 'installation-destination', 'network-hostname', 'root-account', 'user-account']:
+        for page in ['installation-language', 'installation-destination', 'review-configuration', 'installation-progress', 'installation-complete']:
+            # with the pages basically empty of common elements (as those are provided by the top-level wizard widget)
+            # we at least iterate over them to check this works as expected
             b.click("#" + page)
-            b.wait_js_cond(f'window.location.hash === "#/{page}"')
```
KKoukiou, line 43: I would implement a URL change when the next button or the wizard nav is clicked. In this way when a user refreshes they will not be landed in the main intro page (language) but they can maintain their position in the wizard.

ui/webui/src/app.jsx
```diff
@@ -15,33 +15,88 @@
...
-    Page,
+    Page, Wizard,
 } from "@patternfly/react-core";
 
 import { InstallationLanguage } from "./InstallationLanguage.jsx";
-import { Summary } from "./Summary.jsx";
+import { InstallationDestination } from "./storage/InstallationDestination.jsx";
+import { ReviewConfiguration } from "./installation/ReviewConfiguration.jsx";
+import { InstallationProgress } from "./installation/InstallationProgress.jsx";
+import { InstallationComplete } from "./installation/InstallationComplete.jsx";
```
VladimirSlavik, line 29: Should this be one line?
`import { InstallationProgress, InstallationComplete } from "./installation/InstallationProgress.jsx";`

ui/webui/src/app.jsx
```diff
@@ -15,33 +15,88 @@
...
-            <AddressContext.Provider value={address}>
-                <Summary />
-            </AddressContext.Provider>}
+        );
+    };
+
+    const steps = [
+        {
+            id: 0,
+            name: _("Installation language"),
```
VladimirSlavik, line 54: I suppose it's fine to hardcode these now... But in the future the strings and even steps should be somehow generated, same as the grid in our current hub.

poncovka:
> > Installation progress and Installation complete shouldn't be steps in a wizzard. You don't want people to change the configuration after we start the installation. I think it should work like https://www.patternfly.org/v4/components/wizard#finished.
> 
> Yeah, definitely once you start the installation it makes no sense to go back. And it is nice that there is already a pre-prepared way of handling this we might be able to use - thanks for finding it! :)
> 
> Do we want to add it to this PR or maybe rather to a follow-up one ?

That's up to you.

ui/webui/src/app.jsx
jkonecny12: They should be probably imported from the pages directly as variables.

ui/webui/src/installation/InstallationProgress.jsx
```diff
@@ -0,0 +1,32 @@
...
+    Form,
+    PageSection
+} from "@patternfly/react-core";
+
+export const InstallationProgress = () => {
+    return (
+        <Form isHorizontal>
+            <PageSection>
+                Installing... 95% done.
+            </PageSection>
```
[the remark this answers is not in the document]
M4rtinK: Thanks! Removed the ids and rebased - not sure if I handled the notifications currently though when rebasing. :P

ui/webui/src/installation/ReviewConfiguration.jsx
```diff
@@ -0,0 +1,32 @@
...
+
+import {
+    PageSection,
+    Form
+} from "@patternfly/react-core";
+
+export const ReviewConfiguration = () => {
+    return (
+        <Form isHorizontal>
+            <PageSection>
```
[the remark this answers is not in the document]
M4rtinK: OK, dropping the page sections. :)

ui/webui/test/check-basic
M4rtinK: Good point, but seems like something for a followup PR ? Also any pointers how this can be done ?

ui/webui/src/app.jsx
M4rtinK: Hmm, yeah, makes sense, lets do that. :)

M4rtinK: Actually, no I think ? Each jsx file has one of these top-level classes/functions, so you have to select the correct jsx file to import from.

M4rtinK:
> > > Installation progress and Installation complete shouldn't be steps in a wizzard. You don't want people to change the configuration after we start the installation. I think it should work like https://www.patternfly.org/v4/components/wizard#finished.
> > 
> > 
> > Yeah, definitely once you start the installation it makes no sense to go back. And it is nice that there is already a pre-prepared way of handling this we might be able to use - thanks for finding it! :)
> > Do we want to add it to this PR or maybe rather to a follow-up one ?
> 
> That's up to you.

Then lets leave this for a follow-up PR.

M4rtinK:
Updated the PR based on the second round feedback - thanks a lot yet again for all that. :)

There are some remaining eslint issues I have not been really able to fixup so far & will have to learn a bit more about.

ui/webui/src/app.jsx
```diff
@@ -46,6 +53,46 @@ export const Application = () => {
...
+        },
+        {
+            name: _("Review"),
+            component: wrapWithContext(<ReviewConfiguration />),
+            nextButtonText: _("Begin installation"),
+            stepNavItemProps: { id: "review-configuration" }
+        },
+        {
+            name: _("Installation progress"),
+            component: wrapWithContext(<InstallationProgress />),
```
KKoukiou, line 83: This uses the onAddErrorNotification handler. Please add the property like: 
```
InstallationProgress onAddNotification={onAddNotification} />
```

ui/webui/src/app.jsx
KKoukiou: Since you keep these component in different files, you need also to import them in seperate lines as you do now. You could always have more than one components in one file, and in that has you can use this suggestion from vladimir.

ui/webui/src/app.jsx
```diff
@@ -46,6 +53,46 @@ export const Application = () => {
...
+        },
+        {
+            name: _("Storage configuration"),
+            component: wrapWithContext(<InstallationDestination />),
+            stepNavItemProps: { id: "installation-destination" }
+        },
+        {
+            name: _("Review"),
+            component: wrapWithContext(<ReviewConfiguration />),
+            nextButtonText: _("Begin installation"),
```
KKoukiou, line 78: Can you check if there is a way to disable the 'Next' button while installation is running. (Can be done also on followup)

ui/webui/src/app.jsx
```diff
@@ -71,13 +118,16 @@ export const Application = () => {
...
                 })}
             </AlertGroup>}
-            <AddressContext.Provider value={address}>
-                {!path.length > 0 && <InstallationLanguage />}
-            </AddressContext.Provider>
-            {path.length > 0 &&
-            <AddressContext.Provider value={address}>
-                <Summary onAddNotification={onAddNotification} />
-            </AddressContext.Provider>}
+            <Wizard
```
KKoukiou, line 121: Since you have the info from the URL, for which step you are at, aka http://127.0.0.2:9091/cockpit/@localhost/anaconda-webui/index.html#/installation-destination for storage etc.

You can set the correct active step from the URL. Maybe check `activeStep` property? Did not try it out.

ui/webui/src/installation/InstallationLanguage.jsx
```diff
@@ -162,19 +160,15 @@ export const InstallationLanguage = ({ onSelectLang }) => {
...
-            </Form>
-        </PageSection>
+        <Form>
+            <Title headingLevel="h2" size="1xl">
+                WELCOME TO FEDORA...
+            </Title>
+            <FormGroup label={_("What language would you like to use during the installation process?")}>
+                <LanguageSelector lang={lang} onSelectLang={setLang} menuAppendTo={document.body}/>
+            </FormGroup>
+            <ActionGroup>
```
KKoukiou, line 170: You can delete the ActionGroup since you don't use it now.

ui/webui/src/timezone/TimeDate.jsx
```diff
@@ -45,34 +42,24 @@ export const TimeDate = () => {
...
-                          isChecked={useNetworkTime}
-                          label={_("Use network time")}
-                          onChange={setUseNetworkTime}
-                        />
-                    </FormGroup>
-                    <TimeDateManual timedate={serverTime.utc_fake_now} useNetworkTime={useNetworkTime} />
-                </Form>
-            </PageSection>
-        </>
+        <Form isHorizontal>
```
KKoukiou, line 46: Remove double form.

ui/webui/test/check-basic
KKoukiou: Sure, let's handle that in follow up. Probably activeStep property.

ui/webui/src/app.jsx
M4rtinK: Added. :)

ui/webui/src/installation/InstallationLanguage.jsx
M4rtinK: Removed. :)

ui/webui/src/timezone/TimeDate.jsx
M4rtinK: Oh, now I see it! :D Removed.

ui/webui/src/app.jsx
M4rtinK: Removed.

ui/webui/src/app.jsx
M4rtinK: Lets do that in a followup, so we have this in quickly and followup work can happen more easily in parallel. :)

lclay2:
@M4rtinK , If there are there UI changes, is there a way for me to view them? :)

M4rtinK:
> @M4rtinK , If there are there UI changes, is there a way for me to view them? :)

So this is the latest state of the code on screenshots:
v3_01_installation_language
v3_02_storage_configuration
v3_03_review
v3_04_installation_progress
While the step content is empty on this screenshot due to the way I'm running the GUI during development, it should look like on the screenshots attached to this PR during a real installation: https://github.com/rhinstaller/anaconda/pull/3918

v3_05_installation_complete

VladimirSlavik:
Sorry, I think my PR caused conflicts for you :(

ui/webui/src/app.jsx
```diff
@@ -46,6 +53,46 @@ export const Application = () => {
...
+            <AddressContext.Provider value={address}>
+                {children}
+            </AddressContext.Provider>
+        );
+    };
+
+    const steps = [
+        {
+            name: _("Installation language"),
+            component: wrapWithContext(<InstallationLanguage />),
```
KKoukiou, line 67: We need to apply the language selection of the user on the onNext. This was previously done in the handleOnContinue handler.

KKoukiou:
@M4rtinK I just rebased and added two commits, one fixing the language selection handler and one for the navigation + url. I think now you just need to take a look at the tests.

KKoukiou (approved):
@M4rtinK @VladimirSlavik please take a look

KKoukiou:
> Installation progress and Installation complete shouldn't be steps in a wizzard. You don't want people to change the configuration after we start the installation. I think it should work like https://www.patternfly.org/v4/components/wizard#finished.

Updated to this comment, see the new screenshots.

ui/webui/src/app.jsx
KKoukiou: This is implemented now.

VladimirSlavik (approved):
Looks good to me. Thank you!

PR #3918, opened 2022-03-03
webui: first pass on the installation progress component

KKoukiou:
Just for the sake of showing the current screen this is like this:

Screen Shot 2022-03-03 at 15 24 52
Screen Shot 2022-03-03 at 15 24 48

Since there are not designs yet for this view, let's focus more on this PR on how we are consuming the APIs.

### DIFF
--- a/ui/webui/test/check-installation-progress
+++ b/ui/webui/test/check-installation-progress
@@ -1,0 +1,51 @@
+#!/usr/bin/python3
+#
+# Copyright (C) 2022 Red Hat, Inc.
+#
+# This program is free software; you can redistribute it and/or modify it
+# under the terms of the GNU Lesser General Public License as published by
+# the Free Software Foundation; either version 2.1 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful, but
+# WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+# Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with this program; If not, see <http://www.gnu.org/licenses/>.
+
+import os
+import sys
+
+# import Cockpit's machinery for test VMs and its browser test API
+TEST_DIR = os.path.dirname(__file__)
+sys.path.append(TEST_DIR)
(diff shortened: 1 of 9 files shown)


poncovka (approved):
From the DBus API point of view, it looks good to me. Thanks!

jkonecny12 (approved):
LGTM!

M4rtinK (approved):
Looks good to me - and a surprisingly small ammount of code for what it does! :)

ui/webui/test/check-installation-progress
```diff
@@ -0,0 +1,51 @@
...
+    MachineCase.machine_class = VirtInstallMachine
+
+    def testBasic(self):
+        b = self.browser
+        m = self.machine
+
+        wait_webui_started(m)
+        b.open("/cockpit/@localhost/anaconda-webui/index.html")
+
+        b.click('#continue-btn')
```
VladimirSlavik, line 42: Should there be perhaps some check that the progress bar is not visible before clicking the button?

VladimirSlavik (commented):
Looks good to me, thank you. I just wonder, are the Python commits part of this, or just merely required and not merged in some other PR yet?

ui/webui/test/check-installation-progress
KKoukiou: They are required.

KKoukiou: Hmm continue-btn brings us to the current summary page. Then the begin-installation buttons brings us to the progress page.

I could add a wait_not_present check for the selector just before the begin-installation button, but honestly it does not help much, since the component that shows the progress gets rendered only after the begin-installation button os clicked.

VladimirSlavik: Certainly! Do keep in mind that "no" is an answer for the original question, too...

ui/webui/test/ks.cfg
```diff
@@ -1,3 +1,4 @@
+liveimg --url="http://10.0.2.2:8000/live.tar"
```
poncovka, line 1: Btw. the tarball could be local if it would make the Web UI testing easier. Then it would be, for example:
`liveimg --url="file:///tmp/live.tar.xz"`

KKoukiou: This will however expect the file to be inside the test VM right?

poncovka: Yes.

PR #3919, opened 2022-03-03
webui: read conf from installation environment

VladimirSlavik:
This only makes conf available, it is not used yet.

### DIFF
--- a/ui/webui/src/app.jsx
+++ b/ui/webui/src/app.jsx
@@ -24,16 +24,19 @@ import {
 
 import { InstallationLanguage } from "./InstallationLanguage.jsx";
 import { Summary } from "./Summary.jsx";
-import { AddressContext } from "./Common.jsx";
+import { AddressContext, ConfContext } from "./Common.jsx";
+import { readConf } from "./helpers/conf.js";
 
 import { usePageLocation } from "hooks";
 
 export const Application = () => {
     const [address, setAddress] = useState();
     const [notifications, setNotifications] = useState({});
+    const [conf, setConf] = useState();
     const { path } = usePageLocation();
 
     useEffect(() => cockpit.file("/run/anaconda/bus.address").watch(setAddress), []);
+    useEffect(() => readConf().then(setConf, ex => console.error("Failed to parse anaconda configuration")), []);
 
     const onAddNotification = (notificationProps) => {
         setNotifications({
@@ -42,12 +45,14 @@ export const Application = () => {
         });
     };
 
-    if (!address) {
+    // Postpone rendering anything until we read the dbus address and the default configuration
+    if (!address || !conf) {
         return null;
     }
 
+    console.info("conf: ", conf);
     return (
-        <Page>
+        <Page data-debug={conf.Anaconda.debug}>
             {Object.keys(notifications).length > 0 &&
             <AlertGroup isToast isLiveRegion>
                 {Object.keys(notifications).map(idx => {
@@ -72,11 +77,15 @@ export const Application = () => {
                 })}
             </AlertGroup>}
             <AddressContext.Provider value={address}>
-                {!path.length > 0 && <InstallationLanguage />}
+                <ConfContext.Provider value={conf}>
+                    {!path.length > 0 && <InstallationLanguage />}
+                </ConfContext.Provider>
             </AddressContext.Provider>
             {path.length > 0 &&
             <AddressContext.Provider value={address}>
-                <Summary onAddNotification={onAddNotification} />
+                <ConfContext.Provider value={conf}>
+                    <Summary onAddNotification={onAddNotification} />
+                </ConfContext.Provider>
             </AddressContext.Provider>}
         </Page>
     );
--- a/ui/webui/src/helpers/conf.js
+++ b/ui/webui/src/helpers/conf.js
@@ -1,0 +1,78 @@
+/*
+ * Copyright (C) 2022 Red Hat, Inc.
+ *
+ * This program is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation; either version 2.1 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful, but
+ * WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with This program; If not, see <http://www.gnu.org/licenses/>.
+ */
+import cockpit from "cockpit";
+
+const CONF_PATH = "/run/anaconda/anaconda.conf";
+
+const isEmpty = (inStr) => {
+    // is string empty?
+    return !(inStr || inStr.trim());
+};
+
+const parseIni = (content, tag) => {
+    /*
+    Anaconda config files are a dialect of INI files based on Python's configparser. Unfortunately,
+    that parser extends the commonly understood format with multi-line values, which is treated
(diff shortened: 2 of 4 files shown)


ui/webui/src/helpers/conf.js
```diff
@@ -0,0 +1,69 @@
...
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with This program; If not, see <http://www.gnu.org/licenses/>.
+ */
+import cockpit from "cockpit";
+
+const CONF_PATH = "/run/anaconda/anaconda.conf";
+
+const isEmpty = (inStr) => {
```
KKoukiou, line 21: If you just want to check whether there's a truthy value you can do: if (inStr) { .. }
IOW is this helper really needed?

ui/webui/src/helpers/conf.js
```diff
@@ -0,0 +1,69 @@
...
+    // is string empty?
+    return !(inStr || inStr.trim());
+};
+
+export const readConf = () => {
```
KKoukiou, line 26: So none of the available NPM modules worked? Can you maybe add a comment why we ended up implementing our own parser?

ui/webui/src/helpers/conf.js
```diff
@@ -0,0 +1,69 @@
...
+                const dataStore = { loaded: true };
+                const rxHeader = /\[([^\]]+)\].*/;
+                const rxKeyValue = /(\w+) = (.*)/;
+
+                let curHeader = "";
+                let curKey = "";
+
+                for (const index in lines) {
+                    const line = lines[index];
+                    let find = null;
```
KKoukiou, line 40: let find; is enough.

ui/webui/src/helpers/conf.js
```diff
@@ -0,0 +1,69 @@
...
+    // is string empty?
+    return !(inStr || inStr.trim());
+};
+
+export const readConf = () => {
+    const confFile = cockpit.file(CONF_PATH, { superuser: "try", });
+    return confFile.read()
+            .then((content, tag) => {
+                const lines = content.split(/\r\n|\r|\n/);
+                const dataStore = { loaded: true };
```
KKoukiou, line 31: loaded: true is not needed. Till it's available, the consumers can read the default value from the AddressContext.

KKoukiou (commented):
I would be happy if we would at least have a test checking that the config file is parsed.

Do you see any info in that file that we can consume already? 

If not we can expose one information of the file in the DOM, and let the tests expect it. If we go this way I can show you how to do it tomorrow.

ui/webui/src/helpers/conf.js
VladimirSlavik: Sorry, good idea.

ui/webui/src/helpers/conf.js
```diff
@@ -0,0 +1,69 @@
...
+                const rxHeader = /\[([^\]]+)\].*/;
+                const rxKeyValue = /(\w+) = (.*)/;
+
+                let curHeader = "";
+                let curKey = "";
+
+                for (const index in lines) {
```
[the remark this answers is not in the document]
VladimirSlavik: Thanks! But then I can't break.

ui/webui/src/helpers/conf.js
VladimirSlavik: It must return false if there are only spaces.

KKoukiou: ack

ui/webui/src/helpers/conf.js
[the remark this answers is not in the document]
KKoukiou: You can use `return`, instead of `continue`.
`return` in a forEach() callback is equivalent to `continue` in a conventional for loop.

VladimirSlavik:
Most of the data is for backend, but I think `debug=true|false` looks like a good candidate for something generic and useful.

VladimirSlavik:
Added test.

jkonecny12 (approved):
Looks good to me.

ui/webui/src/app.jsx
```diff
@@ -23,24 +23,39 @@ import {
...
 
 import { usePageLocation } from "hooks";
 
 export const Application = () => {
     const [address, setAddress] = useState();
+    const [conf, setConf] = useState();
     const { path } = usePageLocation();
 
     useEffect(() => cockpit.file("/run/anaconda/bus.address").watch(setAddress), []);
+    useEffect(() => readConf().then(data => {
```
KKoukiou, line 37: I know you dont fancy shortHands, but just for your interest this is very often writen like:

``` js
    useEffect(() => readConf().then(setConf, ex => ex => console.error("Failed to parse anaconda configuration")), []);
```

KKoukiou (commented):
The test could be nice part of test/check-basic.

So I would move:
  def testConfDebug(self)

in the existing class in the test/check-basic file. I don't expect the conf file parsing to have more than one test, so having a whole test file for it is maybe an overkill?

KKoukiou (approved):
Two suggestions, but feel free to merge as is.

VladimirSlavik:
Yes, that sounds good. In fact, I wonder if the test could check if there's `[Tt]rue|[Ff]alse` in the data?

KKoukiou:
> Yes, that sounds good. In fact, I wonder if the test could check if there's `[Tt]rue|[Ff]alse` in the data?

You mean the actual file? Because you check the DOM in the tests.

KKoukiou:
Still merge conflicts - aside from the merge conflicts the code looks great.

ui/webui/src/helpers/conf.js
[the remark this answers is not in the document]
VladimirSlavik: Is there any reason to prefer one or the other approach?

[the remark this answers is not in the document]
KKoukiou: I always prefer forEach does not involve index variable setup, which you usually don't need. But I would say it's up to your preference. I just wanted to give you an alternative, feel free to keep it as is.

ui/webui/src/app.jsx
```diff
@@ -72,11 +77,15 @@ export const Application = () => {
...
+                    {!path.length > 0 && <InstallationLanguage />}
+                </ConfContext.Provider>
             </AddressContext.Provider>
             {path.length > 0 &&
             <AddressContext.Provider value={address}>
-                <Summary onAddNotification={onAddNotification} />
+                <ConfContext.Provider value={conf}>
+                    <Summary onAddNotification={onAddNotification} />
+                </ConfContext.Provider>
             </AddressContext.Provider>}
```
VladimirSlavik, line 89: Two questions abut this:
- Are these actually necessary to have in this PR? If the data is not consumed...
- Why use the context provider(s) twice?

KKoukiou: Correct you can remove it.
It's only necesary when you start consuming this 'data' in another component.  

You need to use the Provider, for each context. You have two different context variables to pass eventually, so in that case to Provider elements.

KKoukiou (approved):

ui/webui/src/helpers/conf.js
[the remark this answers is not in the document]
VladimirSlavik: Thank you! In that case I'd prefer to keep it a cycle to explicitly share the scope with the `cur*` variables. (If that's a valid choice...)

ui/webui/src/app.jsx
VladimirSlavik: Ah, sorry, I meant, I have two contexts, so two elements, ok. But why do I need to repeat the wraping for each of the actual contents? Can't I have something like...
```
<AddressContext.Provider>
    <ConfContext.Provider>
        {!path.length > 0 && <InstallationLanguage />}
        <Summary onAddNotification={onAddNotification} />
    </ConfContext.Provider>
</AddressContext.Provider>
```
...instead of repeating the wrapping for each of the innermost lines?

PR #3923, opened 2022-03-04
Test infra check (#infra)

jkonecny12:
**DO NOT MERGE**


jkonecny12:
I wonder why the response is three times but I can fix that later. Works for now.

poncovka:
I guess one check runs for an opened pull request and two for added labels. Maybe we could check if there is already a comment with the same message?

PR #3924, opened 2022-03-04
webui: rewrite language selection component to a class component

KKoukiou:
Usually I prefer using functional components, as they require less
boilerblate code, thus increasing the code reability.
However, when complicated state management needs to be done in the
inialization (in the component mounting), the usage of useEffect hook can get quite
complicated.

Let's convert the LanguageSelector component to be a class component for
this reason, and use the more straightforward componentDidMount
lifecycle method for doing the initialization.

Lastly, this fixes one bug, and thus clicking on the [x] button on the
typeahead select previously did not clear the state.

### DIFF
--- a/ui/webui/src/installation/InstallationLanguage.jsx
+++ b/ui/webui/src/installation/InstallationLanguage.jsx
@@ -15,7 +15,7 @@
  * along with This program; If not, see <http://www.gnu.org/licenses/>.
  */
 
-import React, { useContext, useEffect, useState } from "react";
+import React, { useEffect, useState } from "react";
 import cockpit from "cockpit";
 
 import {
@@ -26,8 +26,6 @@ import {
 
 import { AddressContext } from "../Common.jsx";
 
-import { useEvent, useObject } from "hooks";
-
 const _ = cockpit.gettext;
 
 const getLanguageEnglishName = lang => lang["english-name"].v;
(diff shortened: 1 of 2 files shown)


ui/webui/src/InstallationLanguage.jsx
```diff
@@ -39,105 +37,112 @@ const getLanguageNativeName = lang => lang["native-name"].v;
...
-            Promise.all(languages.map(lang => localizationProxy.GetLocales(lang))).then(res => {
-                return Promise.all(
-                    res.map((langLocales) => {
-                        return Promise.all(langLocales.map(locale => localizationProxy.GetLocaleData(locale)));
+            Promise.all(languages.map(lang => call("GetLocales", [lang])))
+                    .then(res => {
+                        return Promise.all(
+                            res.map(langLocales => {
+                                return Promise.all(langLocales.map(locale =>
+                                    call("GetLocaleData", [locale])
```
M4rtinK, line 73: This looks a little crazy I think. :) I wonder if there is perhaps some way to run this in a worker thread or something, defined as a plain sequence of commands and just return the result to a callback ?

ui/webui/src/InstallationLanguage.jsx
```diff
@@ -39,105 +37,112 @@ const getLanguageNativeName = lang => lang["native-name"].v;
...
-                        ))}
-                    </SelectGroup>
-                );
-            })
-            : []
-    );
+    }
+
+    updateDefaultSelection () {
+        const languageId = this.props.lang.split("_")[0];
```
M4rtinK, line 83: I think it would be good to mention in what format are the language strings or point to where it is defined/documented.

M4rtinK (approved):
Looks good to me function wise, added some questions/suggestions. :)

ui/webui/src/InstallationLanguage.jsx
VladimirSlavik: I don't think it can be turned into anything sequential. Once you start `Promise`-ing you can never stop, you have to chain it all together. And of course D-Bus "calls" here are promises, because they are in reality XHR to cockpit that does the actual call (I think!).

What could be done is reformat the whitespace, or use less anonymous functions, or use more intermediate variables, maybe. I think all of these would help somewhat. You could also get an editor that shows vertical lines for indentation levels and highlights matching braces etc. (except not on github of course). You could also reduce indentation to 2 spaces. Unless the linter exterminates you for trying to think about any of these and losing the spirit of modern javascript?

I guess the overall point is that some time in the mid-future, we might want to sit down and rethink how the code should be formatted. I can't really read this either, especially if there's a review chat in the middle of it.

KKoukiou: I am working now on moving the API calls as functions to seperate files, and I think that will massively increase readability.

VladimirSlavik (approved):
Looks good to me. Thank you!

KKoukiou:
Just did a clean rebase.

PR #3925, opened 2022-03-04
Remove the blivet_gui_supported configuration option

poncovka:
Automatically disable the support for Blivet-GUI if it is not installed. Or it can be
explicitly disabled using the `hidden_spokes` option of the User Interface section.

The `blivet_gui_supported` configuration option was never documented in the RHEL
documentation.


VladimirSlavik (approved):
Looks good to me. Thank you!

jkonecny12 (approved):
Nice improvement.

PR #3926, opened 2022-03-07
Fix Makefile targets using L10N Makefile variables

jkonecny12:
Move of L10N specific variables to po/Makefile.am break some make targets which are using these variables in the root Makefile.am. To fix this, move the variables to configure.ac to have them available everywhere.

This will fix issues with pot file check we have now on the localization repository.

Fixing https://github.com/rhinstaller/anaconda-l10n/runs/5441892439?check_suite_focus=true.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -99,6 +99,10 @@ SHUT_UP_GCC="-Wno-unused-result"
 # Add remaining compiler flags we want to use
 CFLAGS="$CFLAGS -Wall -Werror $SHUT_UP_GCC -fanalyzer"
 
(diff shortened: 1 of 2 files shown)


VladimirSlavik:
This means po-pull needs the whole autogen & configure, am I correct?

jkonecny12:
The `make po-pull` was removed here: https://github.com/rhinstaller/anaconda/pull/3906

VladimirSlavik (approved):
Looks good to me. Thank you!

poncovka (approved):
Looks good to me. Thanks!

PR #3888, opened 2022-02-23
ovirt: move /var/tmp to its own partition

sandrobonazzola:
Move /var/tmp and /var/crash to their own partition on oVirt and RHV for DISA-STIG support.

As CentOS Linux gone EOL, switching oVirt to CentOS Stream

Bug-Url: https://bugzilla.redhat.com/show_bug.cgi?id=2057475
Signed-off-by: Sandro Bonazzola <sbonazzo@redhat.com>

PR #3931, opened 2022-03-08
ovirt: move /var/tmp and /var/crash to own partition

poncovka:
Bug-Url: https://bugzilla.redhat.com/show_bug.cgi?id=2057475
Signed-off-by: Sandro Bonazzola <sbonazzo@redhat.com>

Resolves: rhbz#2057475
**Ported from:** https://github.com/rhinstaller/anaconda/pull/3888

### DIFF
--- a/data/product.d/ovirt.conf
+++ b/data/product.d/ovirt.conf
@@ -4,7 +4,7 @@
 product_name = oVirt Node Next
 
 [Base Product]
-product_name = CentOS Linux
+product_name = CentOS Stream
 
 [Storage]
 default_scheme = LVM_THINP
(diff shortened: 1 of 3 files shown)


rvykydal (approved):
Looks good to me.

VladimirSlavik:
We won't be getting kickstart tests here, because "Branch rhel-8.6 is not supported by kickstart tests yet!" - see https://github.com/rhinstaller/anaconda/runs/5466138667?check_suite_focus=true

rvykydal:
> We won't be getting kickstart tests here, because "Branch rhel-8.6 is not supported by kickstart tests yet!" - see https://github.com/rhinstaller/anaconda/runs/5466138667?check_suite_focus=true

Yes, the tests passed on rhel-8 branch so believe we should be ok for this PR.

jkonecny12 (approved):

M4rtinK (approved):

PR #3932, opened 2022-03-08
Make pylint survive WebUI artifacts

VladimirSlavik:
tl;dr: SSIA.

As part of the current WebUI workflow, we need some other repositories, they contain python files, and our pylint script is greedy and checks them too. To make problems worse, the repos also contain symlinks that point to nowhere, until you use the scripts from these repos. Then some symlinks start pointing to ISOs. And of course, we try to detect python scripts by reading the whole file and finding hashbangs. This PR tries to fix this bag of problems as folows:

* Skip files that can't be opened,
* ~~Skip files that are over 10 MB,~~ Read only first line up to 1024 characters when detecting hashbangs,
* Add false positive matches for the repo subdirs.

### DIFF
--- a/tests/pylint/pylintrc
+++ b/tests/pylint/pylintrc
@@ -8,7 +8,13 @@ extension-pkg-whitelist=
 # Add files or directories to the denylist. They should be base names, not
 # paths.
 # This is auto-fetched from cockpit shared library
-ignore=inotify.py
+ignore=
+
+# Add files or directories matching the regex patterns to the ignore-list.
+# The regex matches against paths.
+ignore-paths=^.*ui/webui/bots/.*$,
+             ^.*ui/webui/src/lib/.*$,
+             ^.*ui/webui/test/common/.*$
 
 # Add files or directories matching the regex patterns to the denylist. The
 # regex matches against base names, not paths.
--- a/tests/pylint/runpylint.py
+++ b/tests/pylint/runpylint.py
@@ -54,13 +54,7 @@ class AnacondaLintConfig(CensorshipConfig):
         for (root, dirnames, files) in os.walk(directory):
 
             # skip scanning of already added python modules
-            skip = False
-            for i in retval:
-                if root.startswith(i):
-                    skip = True
-                    break
-
-            if skip:
+            if any(root.startswith(i) for i in retval):
                 continue
 
             if "__init__.py" in files:
--- a/tests/pylint/runpylint.py
+++ b/tests/pylint/runpylint.py
@@ -65,7 +65,7 @@ class AnacondaLintConfig(CensorshipConfig):
                 try:
                     with open(root + "/" + f) as fo:
                         line = fo.readline(1024)
-                except UnicodeDecodeError:
+                except (UnicodeDecodeError, FileNotFoundError):
                     # If we couldn't open this file, just skip it.  It wasn't
                     # going to be valid python anyway.
                     continue
--- a/tests/pylint/runpylint.py
+++ b/tests/pylint/runpylint.py
@@ -64,7 +64,7 @@ class AnacondaLintConfig(CensorshipConfig):
             for f in files:
                 try:
                     with open(root + "/" + f) as fo:
-                        lines = fo.readlines()
+                        line = fo.readline(1024)
                 except UnicodeDecodeError:
                     # If we couldn't open this file, just skip it.  It wasn't
                     # going to be valid python anyway.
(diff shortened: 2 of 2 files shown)


tests/pylint/runpylint.py
```diff
@@ -39,6 +39,10 @@ def __init__(self):
             FalsePositive(r"E1101.*: Instance of 'int' has no 'name_from_node' member"),
             FalsePositive(r"E1101.*: Instance of 'int' has no 'generate_backup_passphrase' member"),
             FalsePositive(r"E1101.*: Instance of 'int' has no 'dasd_is_ldl' member"),
+
+            # Cockpit stuff that gets checked out inside our repo
+            FalsePositive(r"ui/webui/bots"),
```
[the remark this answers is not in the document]
KKoukiou: And once you add that directory you can remove https://github.com/rhinstaller/anaconda/blob/master/tests/pylint/pylintrc#L11

tests/pylint/runpylint.py
```diff
@@ -68,18 +68,26 @@ def _get_py_paths(self, directory):
...
 
             for f in files:
+                file_path = root + "/" + f
                 try:
-                    with open(root + "/" + f) as fo:
+                    if os.path.getsize(file_path) > 10e6:
+                        # Skip too large files - 10 MB should be enough for everyone!
+                        continue
+                    with open(file_path) as fo:
                         lines = fo.readlines()
```
poncovka, line 77: Weird. Why do we read all lines since we later work with only the first one? I think we should fix that instead of the checking the size of the file.

tests/pylint/runpylint.py
```diff
@@ -68,18 +68,26 @@ def _get_py_paths(self, directory):
...
                 except UnicodeDecodeError:
                     # If we couldn't open this file, just skip it.  It wasn't
                     # going to be valid python anyway.
                     continue
+                except FileNotFoundError:
+                    # This can happen with ui/webui/bots/images/* that is a symlink to nowhere.
+                    # These files are not Python scripts, so let's not even try.
```
KKoukiou, line 84: Is this not respected by the ignore-paths?

VladimirSlavik: No, we first scan here in this function, then pass that list to pylint, which filters by its own ignores. Yes, that's super ugly. Why? I don't know, it's already there.

tests/pylint/runpylint.py
VladimirSlavik: I agree, and have some ideas, so I'll take a look. Let's stay mindful of scope creep, though. I want to ignore some new stuff, not be forced to rewrite the whole thing because you don't like it.

tests/pylint/runpylint.py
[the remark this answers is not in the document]
VladimirSlavik: Thanks! Done.

tests/pylint/runpylint.py
KKoukiou: Doh, thanks!

tests/pylint/runpylint.py
poncovka: Definitely. However, I suggested it as an alternative to the `os.path.getsize(file_path) > 10e6` check. It seems better to fix the original issue then to add a workaround for it.

poncovka: This seems to work.

```
     def _get_py_paths(self, directory):
@@ -70,7 +69,7 @@ class AnacondaLintConfig(CensorshipConfig):
             for f in files:
                 try:
                     with open(root + "/" + f) as fo:
-                        lines = fo.readlines()
+                        line = fo.readline()
                 except UnicodeDecodeError:
                     # If we couldn't open this file, just skip it.  It wasn't
                     # going to be valid python anyway.
@@ -78,7 +77,7 @@ class AnacondaLintConfig(CensorshipConfig):
 
                 # Test any file that either ends in .py or contains #!/usr/bin/python
                 # in the first line.
-                if f.endswith(".py") or (lines and str(lines[0]).startswith("#!/usr/bin/python")):
+                if f.endswith(".py") or (line and str(line).startswith("#!/usr/bin/python")):
                     retval.append(root + "/" + f)
 
         return retval
```

VladimirSlavik: Thank you! I missed that method. That will work, as it has also a data limit parameter. I was looking for a way to limit the amount of data read, in case you have an ISO file and read it until first line end...

poncovka (approved):
Thanks!

jkonecny12:
I really like this new solution thanks!

tests/pylint/pylintrc
```diff
@@ -8,7 +8,13 @@ extension-pkg-whitelist=
 # Add files or directories to the denylist. They should be base names, not
 # paths.
 # This is auto-fetched from cockpit shared library
```
VladimirSlavik, line 10: Oh. Forgot to remove this comment, or move it below.

PR #3934, opened 2022-03-08
Remove RpmDb setup for OSTree payloads

VladimirSlavik:
Related to https://fedoraproject.org/wiki/Changes/RelocateRPMToUsr

Resolves: [rhbz#2050428](https://bugzilla.redhat.com/show_bug.cgi?id=2050428)

@cgwalters - Can you please take a look? Basically I have no idea what I'm doing, but it seems to work, except it somehow worked before too, so... `¯\_(ツ)_/¯`

### DIFF
--- a/pyanaconda/modules/payloads/payload/rpm_ostree/installation.py
+++ b/pyanaconda/modules/payloads/payload/rpm_ostree/installation.py
@@ -147,15 +147,13 @@ class PrepareOSTreeMountTargetsTask(Task):
 
         Next, run tmpfiles to make subdirectories of /var. We need this for both mounts like
         /home (really /var/home) and %post scripts might want to write to e.g. `/srv`, `/root`,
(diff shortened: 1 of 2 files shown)


cgwalters:
systemd-tmpfiles is only really relevant for content in `/var`.  Since the rpmdb is now (correctly) in `/usr`, I think we can just delete the lines of code that talk about `/var/lib/rpm` entirely.

VladimirSlavik:
Thank you, I will try to remove these and see if it still works. Given installation worked just fine both before and after the path change, I think that's correct and will work too.

VladimirSlavik:

Yes, removing "it" seems to work just fine.

M4rtinK (approved):
Looks good  to me, thanks! :)

poncovka (approved):
Thanks!

VladimirSlavik:
Oops! Sorry, fixed.

PR #3935, opened 2022-03-09
network: Handle network configuration paths not existing

AdamWill:
When installing network configuration files, we shouldn't assume
that the relevant paths (network-scripts and system-connections)
actually exist, they don't have to. NetworkManager has split
/etc/sysconfig/network-scripts off into a subpackage that is
no longer installed by default. We guarded against this in
`get_config_files_paths` already, but not in
`_copy_device_config_files`.

### DIFF
--- a/pyanaconda/modules/network/installation.py
+++ b/pyanaconda/modules/network/installation.py
@@ -252,15 +252,17 @@ Name={}
         :param root: path to the root of the target system
         :type root: str
         """
-        for config_file in os.listdir(self.NETWORK_SCRIPTS_DIR_PATH):
-            if config_file.startswith(self.NETWORK_SCRIPTS_CONFIG_FILE_PREFIXES):
-                config_file_path = os.path.join(self.NETWORK_SCRIPTS_DIR_PATH,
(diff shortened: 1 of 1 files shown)


AdamWill:
This fixes live install in current Rawhide and F36 with https://bodhi.fedoraproject.org/updates/FEDORA-2022-e675404750 . I tested.

AdamWill:
I've backported this for Rawhide as it's already broken in current Rawhide composes and we want it fixed ASAP, but I'll hold off backporting for F36 as the change is in updates-testing only, so we have time to do a proper anaconda release and build for F36. The fixed anaconda will need to be edited into https://bodhi.fedoraproject.org/updates/FEDORA-2022-e675404750 .

rvykydal (approved):
Looks good to me, thank you!

VladimirSlavik (approved):
Looks good to me. Thank you!

M4rtinK (approved):
Looks good to me as well.

AdamWill:
Changed base to f36-devel because we need this on F36 too, the NetworkManager change has been submitted there.

bengal:
Hi, any progress on this? We have a NM update on F36 that needs this change.

PR #3937, opened 2022-03-09
Do not crash on network --device link with wireless device (#2051235)

rvykydal:


VladimirSlavik (approved):
Looks good to me. Thank you!

jkonecny12 (approved):
LGTM!

poncovka:
> > Shouldn't we have a test for this?
> 
> What kind of test do you mean? Kstest would require all wired devices having no carier plus existing wireless device. I have hard time imagining unit test mocking the condition of failure.

I think Jirka means a unit test. It should be actually very simple to simulate the conditions using mock objects. I have noticed that the coverage of the Network module is lower compared to other modules. The Storage module isn't much better, but it is also four times bigger. I understand that our priorities are somewhere else at this moment, but we should eventually address that. See: https://codecov.io/gh/rhinstaller/anaconda/tree/master/pyanaconda/modules

PR #3939, opened 2022-03-09
Enable rhel-8 and rhel-9 minor branches for kickstart tests (#infra)

jkonecny12:
With this change we will match all branches starting with rhel-8.<anything> and rhel-9.<anything>.

Not tested yet.


poncovka (approved):
Thanks!

PR #3942, opened 2022-03-09
Update branching settings for rhel-9.0 branch (#infra)

rvykydal:
This PR attempts to do the same what https://github.com/rhinstaller/anaconda/pull/3936 is for 8.6.
No need to update the repos here as this already happened on rhel-9 before forking 9.0.
I am not sure it is ok, it differs from the 8.6 PR a bit. I haven't tested it.
Also it requires also creating of 9.0 branch in weblate.

PR #3943, opened 2022-03-10
Use rhel-9.0 branch in the tests workflow file (#infra)

jkonecny12:
We want to avoid having to change this but that would go to rhel-9 branch instead to improve future branching experience.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -22,8 +22,8 @@ jobs:
         release: ['']
         include:
           - release: ''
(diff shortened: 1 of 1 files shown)


VladimirSlavik (approved):

rvykydal:
I've already updated the strings - alas incompletely - in https://github.com/rhinstaller/anaconda/pull/3942 - which I have now fixed to update it on both places.

jkonecny12:
Thanks for that - closing.

PR #3944, opened 2022-03-10
Remove --ignore-broken test

jkonecny12:
This has to be backported to rhel-9.0 branch.

PR #3945, opened 2022-03-10
Remove --ignore-broken test (#test)

jkonecny12:
In the past the --ignore-broken was disabled on RHEL by Anaconda configuration file, however, in the tests the RHEL conf is not loaded so the test still worked here.

After change b24574113fce6895e32af5f8171e732a3dab134f in pykickstart this functionality was moved to the pykickstart project so it's not allowed to be used on RHEL anymore.

Anaconda code should work as expected because we are disabling this anyhow by the configuration but we have to remove the test to fix our tests right now.

(cherry picked from commit 7e611107d92a6b658586557f82de29ba0776f9b1)
Backport of https://github.com/rhinstaller/anaconda/pull/3944.


jkonecny12:
The same change is already ACKed by @jstodola in https://github.com/rhinstaller/anaconda/pull/3944.

poncovka (approved):

VladimirSlavik (approved):

jstodola (approved):
We don't ship this file in anaconda srpm. (so I approve)

PR #3946, opened 2022-03-10
Fix the --ignore-broken unit test

poncovka:
The `--ignore-broken` option of the `%packages` section is invalid on RHEL,
so make sure that the unit tests can handle that.

Tested on RHEL 9 and Fedora 37.

**Ported from:** https://github.com/rhinstaller/anaconda/pull/3944


VladimirSlavik (approved):
Looks good to me. Thank you!

PR #3947, opened 2022-03-10
 webui: set default storage configuration in the JS code

KKoukiou:
* [x] add docs strings


KKoukiou:
It's probably easier to review this one by looking at https://github.com/rhinstaller/anaconda/pull/3947/files instead of each commit.

ui/webui/src/apis/storage.js
```diff
@@ -0,0 +1,114 @@
...
+import cockpit from "cockpit";
+
+const storageClient = address => {
+    return cockpit.dbus(
+        "org.fedoraproject.Anaconda.Modules.Storage",
+        { superuser: "try", bus: "none", address }
+    );
+};
+
+export const applyPartitioning = (address, partitioning) => {
```
KKoukiou, line 26: @VladimirSlavik @M4rtinK do you think you want to have some kind of doc strings for these?

M4rtinK: Yes, please! :)

KKoukiou: Added

KKoukiou: The localization APIs are still in the react code. I will first merge https://github.com/rhinstaller/anaconda/pull/3924 and then move also these APIs to apis folder and document them.

KKoukiou:
I broke the storage configuraton with last push, will take a look in 2h.

ui/webui/src/apis/boss.js
```diff
@@ -0,0 +1,41 @@
...
+import cockpit from "cockpit";
+
+/**
+ * @param {string} address      Anaconda bus address
+ *
+ * @returns {Object}            A DBus client for the Boss bus
+ */
+export const bossClient = ({ address }) => cockpit.dbus(
+    "org.fedoraproject.Anaconda.Boss",
+    { superuser: "try", bus: "none", address }
```
M4rtinK, line 27: Just wondering, while this is likely not related to the JS code, but the DBus API itself, what does the superuser: try mean & why do we put "none" as the first tuple member we feed into `address` ? Of is it "none" fed to `bus` and address is a positional argument somehow ?

ui/webui/src/apis/storage.js
```diff
@@ -0,0 +1,168 @@
...
+    };
+    taskProxy.wait(() => {
+        addEventListeners();
+        taskProxy.Start().catch(onFail);
+    });
+};
+
+/**
+ * @param {string} address      Anaconda bus address
+ * @param {int} mode            The number of the mode
```
M4rtinK, line 134: I think it would be good to either enumerate the modes here or mention where they are defined.

ui/webui/src/apis/storage.js
```diff
@@ -0,0 +1,168 @@
...
+    return storageClient(address).call(
+        "/org/fedoraproject/Anaconda/Modules/Storage/DiskInitialization",
+        "org.fedoraproject.Anaconda.Modules.Storage.DiskInitialization",
+        "SetInitializationMode", [mode]
+    );
+};
+
+/**
+ * @param {string} address      Anaconda bus address
+ * @param {int} enabled         True if allowed, otherwise False
```
M4rtinK, line 146: So I guess 0 is False and 1 is True ? Any reason why this is not a boolean value or is this some DBus limitation ?

ui/webui/src/components/storage/InstallationDestination.jsx
```diff
@@ -118,13 +119,13 @@ const LocalStandardDisks = () => {
...
                 <DataList isCompact aria-label={_("Usable disks")}>
-                    {usableDisks && usableDisks.map(disk => (
+                    {Object.keys(disks).map(disk => (
                         <DataListItem key={disk} aria-labelledby={"local-disks-checkbox-" + disk}>
                             <DataListItemRow>
                                 <DataListCheck
                                   aria-labelledby={"local-disks-checkbox-" + disk}
-                                  onChange={value => setSelectedDisks({ ...selectedDisks, [disk]: value })}
-                                  checked={!!selectedDisks[disk]}
+                                  onChange={value => setDisks({ ...disks, [disk]: value })}
```
M4rtinK, line 127: These two lines look like quite some magic! :) What does it do actually ?

ui/webui/src/components/storage/InstallationDestination.jsx
```diff
@@ -148,15 +149,36 @@ const LocalStandardDisks = () => {
...
         </Form>
     );
 };
+
+export const applyDefaultStorage = ({ address, onAddErrorNotification }) => {
+    let partitioning;
+
+    // CLEAR_PARTITIONS_ALL = 1
+    return setInitializationMode({ address, mode: 1 })
+            .then(() => setInitializeLabelsEnabled({ address, enabled: true }))
```
M4rtinK, line 170: This is much more readable than the previous solution, thanks! :)

M4rtinK (approved):
Looks good to me, just a couple questions and some suggestions. :)

ui/webui/src/apis/boss.js
KKoukiou: For the cockpit.dbus arguments documentation read: https://cockpit-project.org/guide/latest/cockpit-dbus.html

in the section -> cockpit.dbus()

The superuser will just try to create the bus with root privileged. I guess it's actually no needed, as these processes are running in privileged mode.

"none" is for the "bus" configuration. we need to put "none" according to the docs above, so as to connect to the non-standard bus specified with the address option. This defaults to "system".

The address value in the object will be substituted by the 'address' parameter of the bossClient function.

ui/webui/src/apis/storage.js
KKoukiou: I really don't want to duplicate the API documentation.
If the documentation for the pyanaconda DBUS api was actually generated and online somewhere, I would gladly add a link to the original method that we expose here. But till then I feel it's the maximum amount of info we should add in this docstring without duplicating docs.

ui/webui/src/apis/storage.js
KKoukiou: It's my fault, it's boolean. Will update.

ui/webui/src/components/storage/InstallationDestination.jsx
KKoukiou:
``` js
value => setDisks({ ...disks, [disk]: value })

```
can be written without arrow functions syntax as:

``` js
function (value) {
   setDisks({ ...disks, [disk]: value });
}
```
And if avoid even using the object destructuring it can be written as:

```js
function (value) {
  // We want to create a new object, by using the object 'disks' and adding the new disk
  // 'disks' object is part of the state, so we should not update it directly
  // Instead we need to update the state with setDisks

  // Create a new clone of disks
  const newDisks = Object.assign({},disks);
  // Add the new disk value
  newDisks[disk] = value; 
  // Update the state
  setDisks(newDisks);
}
```

The previous can be so much nicer be done with the spread operator, to avoid creating a new variable.

KKoukiou: Does this make sense? If not let's talk it over in a call.

KKoukiou: Just to mention ths once more, the reason for this complexity here, is that we really * have to * update the state in an immutable way. Mutatiing state is a big NO in react ecosystem, so any operation here updating the 'disks' state object directly is not allowed.

ui/webui/src/apis/boss.js
M4rtinK: I get it now, thanks! :)

ui/webui/src/apis/storage.js
M4rtinK: OK, makes sense. :)

KKoukiou:
@VladimirSlavik @M4rtinK after this I plan to try to increase readability in the React component state creation, like how we create objects for react to consume from the API.

VladimirSlavik (approved):
Thank you, looks good to me.

PR #3949, opened 2022-03-10
webui: :fire:  webpack: process all assets when compressing

KKoukiou:
By default, only assets that compress better than this ratio (0.8) are
processed.

https://github.com/webpack-contrib/compression-webpack-plugin#minratio

That ended up leaving some of our translation files uncompressed, as
they contained just a few lines, resulting in minimal compression
difference.


KKoukiou:
This unbreaks master, please prioritize review @jkonecny12 @M4rtinK @VladimirSlavik

rvykydal (approved):
Looks good to me.

VladimirSlavik:
Thank you!

PR #3957, opened 2022-03-14
Skip /etc/machine-info during live installs (#2036199)

jkonecny12:
`/etc/machine-info` should not be installed to the system, for the same reason we don't install `/etc/machine-id`: it should be system-specific, not the same for every install from the same live image.

This should fix live installs with systemd 250+ not booting due to the bootloader config snippets being named based on the `KERNEL_INSTALL_MACHINE_ID` from the live image `/etc/machine-info`, but `grub2-mkconfig` expecting them to be named according to the installed system's newly-generated `/etc/machine-id`.

(based on commit fe652add9733943a3476128a83b24b9a8c63b335)

Resolves: rhbz#2059565


jkonecny12:
The bug is not ready yet.

VladimirSlavik (approved):

rvykydal (approved):
Looks good to me.

jkonecny12:
Already ported.

PR #3959, opened 2022-03-15
 webui: fix progress reporting in the installation progress step

KKoukiou:
And do some simplifications on the apis/ folder.

### DIFF
--- a/ui/webui/src/apis/boss.js
+++ b/ui/webui/src/apis/boss.js
@@ -29,6 +29,21 @@ export const bossClient = ({ address }) => cockpit.dbus(
 
 /**
  * @param {string} address      Anaconda bus address
+ * @param {string} task         DBus path to a task
+ *
+ * @returns {Promise}           Resolves the total number of tasks
+ */
+export const getSteps = ({ address, task }) => {
+    return bossClient({ address }).call(
+        task,
+        "org.freedesktop.DBus.Properties",
+        "Get",
+        ["org.fedoraproject.Anaconda.Task", "Steps"]
+    );
+};
+
+/**
+ * @param {string} address      Anaconda bus address
  *
  * @returns {Promise}           Resolves a list of tasks
  */
@@ -37,5 +52,6 @@ export const installWithTasks = ({ address }) => {
         "/org/fedoraproject/Anaconda/Boss",
         "org.fedoraproject.Anaconda.Boss",
         "InstallWithTasks", []
-    );
+    )
+            .then(ret => ret[0]);
 };
--- a/ui/webui/src/apis/boss.js
+++ b/ui/webui/src/apis/boss.js
@@ -22,19 +22,31 @@ import cockpit from "cockpit";
  *
  * @returns {Object}            A DBus client for the Boss bus
  */
-export const bossClient = ({ address }) => cockpit.dbus(
-    "org.fedoraproject.Anaconda.Boss",
-    { superuser: "try", bus: "none", address }
-);
+export class BossClient {
+    constructor (address) {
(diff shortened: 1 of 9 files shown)


ui/webui/src/apis/boss.js
```diff
@@ -27,6 +27,21 @@ export const bossClient = ({ address }) => cockpit.dbus(
     { superuser: "try", bus: "none", address }
 );
 
+/**
+ * @param {string} address      Anaconda bus address
+ * @param {string} task         DBus path to a task
+ *
+ * @returns {Promise}           Resolves a the total number of tasks
```
M4rtinK, line 46: I think there is a typo. :)

ui/webui/src/components/installation/InstallationProgress.jsx
```diff
@@ -46,18 +46,22 @@ export class InstallationProgress extends React.Component {
                         "org.fedoraproject.Anaconda.Task",
                         tasks[0]
                     );
+
                     const addEventListeners = () => {
                         taskProxy.addEventListener("ProgressChanged", (_, step, message) => {
+                            if (step === 0) {
+                                getSteps({ address: this.context, task: tasks[0] })
+                                        .then(
+                                            ret => this.setState({ steps: ret[0].v }),
```
M4rtinK, line 54: So I get that `ret` is the data returned by async call, and its apparently some array, so `ret[0]` gives you the first item from that array. But where does the `ret[0].v` come from ? :-D

ui/webui/src/apis/boss.js
```diff
@@ -43,12 +55,10 @@ export const getSteps = ({ address, task }) => {
...
 
 /**
- * @param {string} address      Anaconda bus address
- *
  * @returns {Promise}           Resolves a list of tasks
  */
-export const installWithTasks = ({ address }) => {
-    return bossClient({ address }).call(
+export const installWithTasks = () => {
+    return new BossClient().client.call(
```
M4rtinK, line 61: So this makes use of the singleton ? Interesting. :)

M4rtinK (approved):
Looks good to me, just some questions and a typo. :)

ui/webui/src/components/installation/InstallationProgress.jsx
KKoukiou: The anaconda dbus API returns properties in a variant object. So it's an object with keys, 'v' for value and 't' for type.

ui/webui/src/apis/boss.js
KKoukiou: Indeed. so now only the first call will result in an new boss interface dbus client creation, etc.

VladimirSlavik (approved):
Looks good to me. Thank you!

ui/webui/src/components/installation/InstallationProgress.jsx
KKoukiou: did the [0] part in the API method wrapper, to remove the complexity of the react side.

KKoukiou:
Updated according to the comments, and rebased.

ui/webui/src/components/installation/InstallationProgress.jsx
M4rtinK: Oh, interesting - I was thinking its either value or variant, that explains it. :)

ui/webui/src/apis/boss.js
M4rtinK: Cool! In this case its mainly for performance/efficiency but I'm sure we will hit cases where we will need to make sure something is there once and only once & this will be very handy for that as well. :)

PR #3960, opened 2022-03-15
Betanag for web ui

VladimirSlavik:
This adds a label with popup, as specified in the mockups. The label is shown or hidden dynamically, based on the contents of the `/.buildstamp` file.

The large light blue warning pane (?) at startup is not yet part of this, I will need to check how to make it appear as intended (before wizard step label/heading).

TODOs left:
- [x] Meaningful texts
- [x] Tests
- [ ] In-page warning at language selection or so

----

To summarize, this is the new thing:

betanag

----
Thank you @KKoukiou for the wizard changes!

### DIFF
--- a/ui/webui/src/components/AnacondaHeader.jsx
+++ b/ui/webui/src/components/AnacondaHeader.jsx
@@ -1,0 +1,69 @@
+/*
+ * Copyright (C) 2022 Red Hat, Inc.
+ *
+ * This program is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation; either version 2.1 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful, but
+ * WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with This program; If not, see <http://www.gnu.org/licenses/>.
+ */
+import cockpit from "cockpit";
+
+import React from "react";
+
+import {
+    Flex,
+    Label,
+    PageSection, PageSectionVariants,
+    Popover,
+    TextContent, Text, TextVariants
+} from "@patternfly/react-core";
+import { InfoCircleIcon } from "@patternfly/react-icons";
+
+const _ = cockpit.gettext;
+
+const prerelease = _("Pre-release");
+
+export const AnacondaHeader = ({ beta, title }) => {
+    let betanag;
+    if (beta) {
+        betanag = (
+            <Popover
+              headerContent={_("This is unstable, pre-release software")}
+              minWidth="40rem"
+              bodyContent={
+                  <TextContent>
+                      <Text component={TextVariants.p}>
+                          {_("Notice: This is pre-released software that is intended for development and testing purposes only. Do NOT use this software for any critical work or for production environments.")}
+                      </Text>
+                      <Text component={TextVariants.p}>
+                          {_("By continuing to use this software, you understand and accept the risks associated with pre-released software, that you intend to use this for testing and development purposes only and are willing to report any bugs or issues in order to enhance this work.")}
+                      </Text>
+                      <Text component={TextVariants.p}>
+                          {_("If you do not understand or accept the risks, then please exit this program.")}
+                      </Text>
+                  </TextContent>
+              }
+            >
+                <Label color="orange" icon={<InfoCircleIcon />} id="betanag-icon"> {prerelease} </Label>
+            </Popover>
+        );
+    }
+    return (
+        <PageSection variant={PageSectionVariants.light}>
+            <Flex spaceItems={{ default: "spaceItemsSm" }} alignItems={{ default: "alignItemsCenter" }}>
+                <TextContent>
+                    <Text component="h1">{title}</Text>
+                </TextContent>
+                {betanag}
+            </Flex>
+        </PageSection>
+    );
+};
--- a/ui/webui/src/components/app.jsx
+++ b/ui/webui/src/components/app.jsx
@@ -27,6 +27,7 @@ import {
 } from "@patternfly/react-core";
 
 import { AddressContext } from "./Common.jsx";
+import { AnacondaHeader } from "./AnacondaHeader.jsx";
 import { InstallationDestination, applyDefaultStorage } from "./storage/InstallationDestination.jsx";
 import { InstallationLanguage } from "./installation/InstallationLanguage.jsx";
 import { InstallationProgress } from "./installation/InstallationProgress.jsx";
@@ -36,6 +37,7 @@ import { readConf } from "../helpers/conf.js";
 import { BossClient } from "../apis/boss.js";
 import { LocalizationClient } from "../apis/localization.js";
 import { StorageClient } from "../apis/storage.js";
+import { readBuildstamp, getIsFinal } from "../helpers/betanag.js";
 
 import { usePageLocation } from "hooks";
 
@@ -44,6 +46,7 @@ const _ = cockpit.gettext;
 export const Application = () => {
     const { path } = usePageLocation();
     const [address, setAddress] = useState();
+    const [beta, setBeta] = useState();
     const [conf, setConf] = useState();
     const [notifications, setNotifications] = useState({});
     const [stepNotification, setStepNotification] = useState();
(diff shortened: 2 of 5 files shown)


ui/webui/src/components/app.jsx
```diff
@@ -107,9 +112,31 @@ export const Application = () => {
...
+        <Page
+          data-debug={conf.Anaconda.debug}
+          additionalGroupedContent={
+              <PageSection variant={PageSectionVariants.light}>
+                  <Flex spaceItems={{ default: "spaceItemsSm" }} alignItems={{ default: "alignItemsCenter" }}>
+                      <TextContent>
+                          <Text component="h1">{title}</Text>
+                      </TextContent><br />
+                      <Popover
+                        aria-label=""
```
KKoukiou: empty aria label cancels the purpose of aria label :) See docs, says 'Accessible label, required when header is not present' 

You have defined the header here, so it's fine to drop it.

ui/webui/src/components/app.jsx
```diff
@@ -107,9 +112,31 @@ export const Application = () => {
...
     return (
-        <Page data-debug={conf.Anaconda.debug}>
+        <Page
+          data-debug={conf.Anaconda.debug}
+          additionalGroupedContent={
+              <PageSection variant={PageSectionVariants.light}>
+                  <Flex spaceItems={{ default: "spaceItemsSm" }} alignItems={{ default: "alignItemsCenter" }}>
+                      <TextContent>
+                          <Text component="h1">{title}</Text>
+                      </TextContent><br />
```
KKoukiou, line 124: Please don't use br for spacing. That's what Flex and other Layout components from PF4 are for. If this is not working as expected, you might need to wrap your Popver in a FlexItem. will check it interactively and propose a solution.

ui/webui/src/components/app.jsx
```diff
@@ -40,10 +43,12 @@ export const Application = () => {
     const [isStorageReady, setIsStorageReady] = useState(false);
     const [notifications, setNotifications] = useState({});
     const [conf, setConf] = useState();
+    const [beta, setBeta] = useState();
     const { path } = usePageLocation();
 
     useEffect(() => cockpit.file("/run/anaconda/bus.address").watch(setAddress), []);
     useEffect(() => readConf().then(setConf, ex => console.error("Failed to parse anaconda configuration")), []);
+    useEffect(() => readBuildstamp().then(buildstamp => setBeta(!getIsFinal(buildstamp)), ex => console.error("Failed to parse anaconda configuration")), []);
```
KKoukiou, line 51: You can have one useEffect call for these.

``` js

useEffect(() => {
    cockpit.file("/run/anaconda/bus.address").watch(setAddress);
    readConf().then(setConf, ex => console.error("Failed to parse anaconda configuration");
    readBuildstamp().then(buildstamp => setBeta(!getIsFinal(buildstamp)), ex => console.error("Failed to parse anaconda configuration"));
}, []);
```

Please adjust, did not notice that there were already 2. Practically there is not difference, just less boilerplate code.

KKoukiou: Also try to avoid maybe ultra long lines.
I like to put the .then arguments in new lines; so for this.

``` js

useEffect(() => readBuildstamp().then(
    buildstamp => setBeta(!getIsFinal(buildstamp)),
    ex => console.error("Failed to parse anaconda configuration: ", ex.message))
, []);
```

ui/webui/src/components/header.jsx
```diff
@@ -0,0 +1,60 @@
...
+    Popover,
+    TextContent, Text
+} from "@patternfly/react-core";
+import { InfoCircleIcon } from "@patternfly/react-icons";
+
+const _ = cockpit.gettext;
+
+const prerelease = _("Pre-release");
+
+export const AnacondaHeader = (props) => {
```
KKoukiou, line 34: Please rename the file to capital. Since all files apart from app.jsx are capitalized. We can do the same for app.jsx.

At least to be consistent.

ui/webui/src/components/header.jsx
```diff
@@ -0,0 +1,60 @@
...
+    let betanag;
+    if (props.beta) {
+        betanag = (
+            <Popover
+              headerContent={_("Fedora ## pre-release")}
+              bodyContent={_(`Notice: This is pre-released software that is intended for development and testing purposes only. Do *not* use this software for any critical work or for production environments.
```
KKoukiou, line 40: For allowing cockpit to translate strings you need to use double quote. Stupid limitation.

ui/webui/src/components/header.jsx
KKoukiou, line 34: It's nice to keep props expanded. In this case;
instead of (props) -> ({ beta, title })

ui/webui/src/components/app.jsx
```diff
@@ -20,21 +20,22 @@ import React, { useEffect, useState } from "react";
...
+    Page,
+    Wizard
 } from "@patternfly/react-core";
 
 import { AddressContext } from "./Common.jsx";
 import { InstallationDestination, applyDefaultStorage } from "./storage/InstallationDestination.jsx";
 import { InstallationLanguage } from "./installation/InstallationLanguage.jsx";
 import { InstallationProgress } from "./installation/InstallationProgress.jsx";
 import { ReviewConfiguration } from "./installation/ReviewConfiguration.jsx";
+import { AnacondaHeader } from "./header.jsx";
```
KKoukiou, line 32: Shall we keep these imports in alphabetical order?

KKoukiou (commented):
super! Almost there? Shall we add a small check in the testBasic?

ui/webui/src/components/header.jsx
VladimirSlavik: A number of questions.
- Moving from backtick strings to double-quoted ones means I lose the ability to keep this a multi-line literal. How do I wrap it nicely? How do I keep it such that the translation extractor understands it as one literal, not three?
- Can I have 3 paragraphs instead of one? If so, can I make it somehow so that the `<p>` etc. are inside the string literal?

KKoukiou:
> I think I managed to address most of the feedback above?
> 
> Still todo:
> 
>     * I will review the current mockups to see if the large warning is still desired.
> 
>     * Find out how to make a test that also understands the file contents.

Testing that can be done as following. Use m.execute to get the line in interest.
```
pattern = m.execute("cat /file/of/interest | grep matchingKeyword")
if 'pre-release' in pattern:
   self.wait_visible("#id-pf-the-prerelease-label")
```

ui/webui/src/components/AnacondaHeader.jsx
```diff
@@ -0,0 +1,60 @@
...
+
+const prerelease = _("Pre-release");
+
+export const AnacondaHeader = ({ beta, title }) => {
+    let betanag;
+    if (beta) {
+        betanag = (
+            <Popover
+              headerContent={_("This is unstable, pre-release software")}
+              bodyContent={_(`Notice: This is pre-released software that is intended for development and testing purposes only. Do *not* use this software for any critical work or for production environments.
```
KKoukiou, line 40: I would split it into three to-translate strings.

``` js
<TextContent>
   <Text component={TextVariants.p}>
      {_("Notice: This is pre-released software that is intended for development and testing purposes only. Do *not* use this software for any critical work or for production environments.")}
  <Text>
   <Text component={TextVariants.p}>
     {_("By continuing to use this software, you understand and accept the risks associated with pre-released software, that you intend to use this for testing and development purposes only and are willing to report any bugs or issues in order to enhance this work.")}
  <Text>
   <Text component={TextVariants.p}>
      {_("If you do not understand or accept the risks, then please exit this program.`)}")}
  <Text>
<TextContent>
```

KKoukiou: For you interest cockpit also offers https://github.com/cockpit-project/cockpit/blob/main/pkg/lib/utils.jsx#L22
but with that you can translate "Visit <b>127.0.0.1</b>" but not "Visit <b>string to be translated</b>. So anything inside the html tags, wont get translated

VladimirSlavik: Thank you! Done.

VladimirSlavik:
I added a test but somehow it seems to hang locally? Let's see what happens on the PR.

ui/webui/test/check-basic
```diff
@@ -40,6 +41,11 @@ class TestBasic(MachineCase):
             ["False", "false", "True", "true"]
         )
 
+        # Check betanag
+        value = m.execute("cat /.buildstamp | grep IsFinal=")
+        if value.split("=", 1)[1] in ("False", "false"):
+            b.wait_visible("#betanag-icon")
+
```
KKoukiou, line 48: I would add an else as well.
```
else:
  b.wait_not_present("#betang-icon")
```

So that you make sure this actually covers all scenarios.
But it can be considered follow up material.

KKoukiou (commented):
Perfect :star

KKoukiou (approved):
The failing test is indeed not your fault, seems like some flakeness, (retried once) I need to take a look.

KKoukiou:
@M4rtinK please take a look as well.

M4rtinK (approved):

PR #3961, opened 2022-03-15
webui: show inline per - step alerts and stay in the same step when configuration is invalid

KKoukiou:
For example when the user has an invalid disk selection step, the page looks like this when clicking 'Next':

Screen Shot 2022-03-15 at 17 45 45

### DIFF
--- a/ui/webui/src/components/app.jsx
+++ b/ui/webui/src/components/app.jsx
@@ -39,11 +39,12 @@ import { usePageLocation } from "hooks";
 const _ = cockpit.gettext;
 
 export const Application = () => {
+    const { path } = usePageLocation();
     const [address, setAddress] = useState();
+    const [conf, setConf] = useState();
     const [isStorageReady, setIsStorageReady] = useState(false);
     const [notifications, setNotifications] = useState({});
-    const [conf, setConf] = useState();
-    const { path } = usePageLocation();
+    const [stepIdReached, setStepIdReached] = useState(path[0] || "installation-language");
 
     useEffect(() => cockpit.file("/run/anaconda/bus.address").watch(address => {
         const clients = [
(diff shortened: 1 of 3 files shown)


ui/webui/src/components/app.jsx
```diff
@@ -104,6 +108,12 @@ export const Application = () => {
         if (prevStep.prevId === "installation-destination") {
             applyDefaultStorage({ address, onAddErrorNotification, onSuccess: () => setIsStorageReady(true) });
         }
+        const stepIdx = steps.findIndex(s => s.id === stepIdReached.id);
+        const newStepIdx = steps.findIndex(s => s.id === newStep.id);
```
VladimirSlavik, line 131: `stepIdReached` should not use `.id` here, I think? Below you do `setStepIdReached(newStep.id);` which already does that part.

VladimirSlavik (commented):
LGTM except for the one thing below...

ui/webui/src/components/app.jsx
KKoukiou: Oh great catch!!

KKoukiou: Fixed! thanks for noticing!

VladimirSlavik (approved):
In that case I don't see anything else!

M4rtinK (approved):
Looks good to me! :)

We had some discussion about how to best handle the canJumpTo logic, but I'm find with this being improved in this PR or in a separate one. :)

KKoukiou:
> Looks good to me! :)
> 
> We had some discussion about how to best handle the canJumpTo logic, but I'm find with this being improved in this PR or in a separate one. :)

I will send a followup.

PR #3962, opened 2022-03-17
Update icon sizes for symbolic icons

rvykydal:
I've checked the icons modified to symbolic versions in https://github.com/rhinstaller/anaconda/pull/3902 and there are 3 places where we could or should update icon size for the symbolic version so it looks better:

1) Betanag dialog (GTK_ICON_SIZE_DIALOG 48px ->  GTK_ICON_SIZE_DND 32px): here I am not sure, I would be fine with the bigger icon as well, maybe it is even better...
2) Quit dialog: (GTK_ICON_SIZE_DIALOG 48px ->  GTK_ICON_SIZE_DND 32px) I would definitely add the border, smaller size seems a bit better to me, but shouldn't we use _DIALOG in dialog?

Actually I think for the dialogs the real problem is not too big icons but icons being too close to the border but that would be something that would need to be addressed in another PR for all dialogs consistently.

3) Network spoke: Here we should definitely use smaller icons.

Note: my goal here was not to fix the places to be looking perfect but not much worse then it used to with non-symbolic icons.
Suggestions and opinions welcome.

### DIFF
--- a/pyanaconda/ui/gui/main.glade
+++ b/pyanaconda/ui/gui/main.glade
@@ -83,6 +83,7 @@
     <child internal-child="vbox">
       <object class="GtkBox" id="quitDialog-vbox1">
         <property name="can_focus">False</property>
+        <property name="border_width">6</property>
         <property name="orientation">vertical</property>
         <property name="spacing">24</property>
         <child internal-child="action_area">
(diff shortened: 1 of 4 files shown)


rvykydal:
I'll add the screenshots shortly.

rvykydal:
1) Betanag dialog
Current state:
betanag_current
Patched:
betanag_patched

rvykydal:
2. Quit dialog
Current state:
quit_current
Patched:
quit_patched

rvykydal:
3. Network spoke
Current state:
network_current
Patched:
network_patched

VladimirSlavik (approved):
Thank you, I agree with all of your changes and ideas.

poncovka (approved):
Thanks! I agree with your approach. Do whatever you think is best.

jkonecny12 (approved):
Looks great. Thanks a lot for solving this!

PR #3963, opened 2022-03-17
webui: move wizard code out of app.jsx to a new file

KKoukiou:


ui/webui/src/components/app.jsx
```diff
@@ -43,14 +43,73 @@ import { usePageLocation } from "hooks";
...
+            canJumpTo: idx === 0 ? currentStepId == s.id : stepsVisited.includes(s.id),
+            isFinishedStep: idx === stepsOrder.length - 1
+        });
+    });
+};
+
 export const Application = () => {
+    const stepsOrder = [
+        {
+            c: InstallationLanguage,
```
M4rtinK: Even if its some extra typing I would recommend to use full names for the keys, eq. "component", "label". Otherwise I think it could be a little too magic elsewhere in the code & not clear where the names come from. :)

M4rtinK (approved):
Looks good to me, thanks! :) Just one suggestion about variable naming in one place. :)

ui/webui/src/components/app.jsx
KKoukiou: Agreed :)

M4rtinK (approved):
Looks even better now, thanks! :)

PR #3967, opened 2022-03-18
webui: start using custom webui-desktop script instead of cockpit-desktop

KKoukiou:


poncovka (approved):
Looks good to me. Thanks!

KKoukiou:
@poncovka by adding the linux-core-utils package, the installation in the test started failing earlier, when tyring to mount the root partition. This rpm indeed brings in the mount command, so it's totally related.

https://logs.cockpit-project.org/logs/pull-3967-20220318-110333-d743d8d9-fedora-rawhide-boot/TestInstallationProgress-testBasic-fedora-rawhide-boot-127.0.0.2-2601-FAIL.png

however I can't make much sense of it? do you?

KKoukiou:
The other option would be to teach lorax to leave, /bin/unshare in the boot.iso.

https://github.com/weldr/lorax/blame/3958d2aa8b28b00ed1be7362e0206525e70600c9/share/templates.d/99-generic/runtime-cleanup.tmpl#L294

I am not sure what is the better way.

poncovka:
@KKoukiou, it means that the additional dependencies cannot be added via an updates image, they have to be installed. I wonder whether we need to use unshare on the boot.iso. I understand its meaning for Cockpit, but the system on the boot.iso is already isolated since it's only purpose is to run Anaconda, isn't it? Unfortunately, I don't understand it enough to make any decisions here. @rvykydal, do you have some insights?

rvykydal:
I don't know if we need to unshare nework either, I have no experience here. Perhaps we should ask @martinpitt.
So with what we know now updating lorax template to add unshare seems to be safer (also maybe to prevent some surprises / effects in the future).

KKoukiou:
> I don't know if we need to unshare nework either, I have no experience here. Perhaps we should ask @martinpitt. So with what we know now updating lorax template to add unshare seems to be safer (also maybe to prevent some surprises / effects in the future).

I agree, so updating the lorax template to not delete random binaries would save us from the need to keep a custom cockpit-desktop script here. (not a big deal, it's pretty trivial)

However, I think we need for now this workaround, till we send a patch to lorax, and this get's released, upstreamed etc.

rvykydal:
If it works, I am OK with the workaround.

KKoukiou:
> If it works, I am OK with the workaround.

it works, but the unit tests fail for some reason. any hints welcome

poncovka:
The failing tests are weird.
```
Trying to pull registry.fedoraproject.org/fedora:rawhide...
...
Fedora 36 - x86_64                               18 MB/s |  80 MB     00:04    
Fedora 36 openh264 (From Cisco) - x86_64        3.4 kB/s | 2.5 kB     00:00    
Fedora Modular 36 - x86_64                      3.0 MB/s | 2.3 MB     00:00    
Fedora 36 - x86_64 - Updates                    295  B/s | 257  B     00:00    
Fedora Modular 36 - x86_64 - Updates            347  B/s | 257  B     00:00    
Fedora 36 - x86_64 - Test Updates               9.7 MB/s |  12 MB     00:01    
Fedora Modular 36 - x86_64 - Test Updates       336 kB/s | 320 kB     00:00 
```
It should be a Rawhide container, but it uses F36 repositories, so it has to be a F36 container. No idea why. @rvykydal, do you have time to investigate it?

KKoukiou:
The unit tests failure I hit it locally also on master.

poncovka:
Shellcheck doesn't like the new script.

```
ui/webui/webui-desktop:40:1: warning: exec_prefix appears unused. Verify use (or export if used externally). [SC2034]
ui/webui/webui-desktop:41:1: warning: libexecdir appears unused. Verify use (or export if used externally). [SC2034]
ui/webui/webui-desktop:84:12: warning: Use single quotes, otherwise this expands now rather than when signalled. [SC2064]
ui/webui/webui-desktop:84:30: warning: Use single quotes, otherwise this expands now rather than when signalled. [SC2064]
ui/webui/webui-desktop:87:8: note: Expressions don't expand in single quotes, use double quotes for that. [SC2016]
ui/webui/webui-desktop:98:99: note: Expressions don't expand in single quotes, use double quotes for that. [SC2016]
ui/webui/webui-desktop:119:21: note: Double quote to prevent globbing and word splitting. [SC2086]
ui/webui/webui-desktop:119:36: note: Double quote to prevent globbing and word splitting. [SC2086]
FAIL shellcheck/run_shellcheck.sh (exit status: 1)
```
These issues have to be fixed or they can disabled with `# shellcheck disable=`.

M4rtinK:
So there is possibly another way - change `unshare` use in upstream `cockpit-desktop` script to be conditional. Then we could call something like:

`/usr/libexec/cockpit-desktop /cockpit/@localhost/anaconda-webui/index.html --no-unshare`

By default `cockpit-desktop` would still call the `unshare` utility (I guess its there if the web view is used for untrusted content, say possibly WifFi captive portals or similar) but we could tell it not to use `unshare` as it does not really make sense in our case.

@KKoukiou @martinpitt Doe it make some sense ? I can definitely open a PR on Cockpit upstream for this if it does. :)

Still in the meantime we should definitely go with the custom-script workaround, as it will certainly take some time if we go with this.

martinpitt:
@M4rtinK : On a real system you must not *ever* run this without `unshare`. It would open up complete access to your session to *every* process and user on the same machine. This opens a new namespace which puts cockpit-ws and the browser into its own little universe. It's OK as a hack in an installer environment where we can be reasonably sure that there is no code running as another user anyway, but not as an user-visible option. So I am *very* hesitant to add an option for this.

AFAIUI the unshare as such doesn't break anything, it's just hard to get a binary into the system via the update image? would it be any easier to ship a little mock /usr/local/bin/unshare shell script which ignores its options and just executes its argument?
Or would it be possible to put /usr/bin/unshare into the update image through some other means? E.g. the cockpit-webui package itself could ship a copy of it (possibly in /usr/libexec/, and you set `$PATH`). Or put it into the update image as a plain file, not rpm?

M4rtinK:
> @M4rtinK : On a real system you must not _ever_ run this without `unshare`. It would open up complete access to your session to _every_ process and user on the same machine. This opens a new namespace which puts cockpit-ws and the browser into its own little universe. It's OK as a hack in an installer environment where we can be reasonably sure that there is no code running as another user anyway, but not as an user-visible option. So I am _very_ hesitant to add an option for this.
Interesting - agreed, it does not seems like an idea to do this in this context.


> 
> AFAIUI the unshare as such doesn't break anything, it's just hard to get a binary into the system via the update image? would it be any easier to ship a little mock /usr/local/bin/unshare shell script which ignores its options and just executes its argument? Or would it be possible to put /usr/bin/unshare into the update image through some other means? E.g. the cockpit-webui package itself could ship a copy of it (possibly in /usr/libexec/, and you set `$PATH`). Or put it into the update image as a plain file, not rpm?

So when we tried to add `unshare` via the updates image, which boils down to dumping the content of `util-linux-core` RPM to the rootfs, we hit some very strange issues. This could be either due to specific of how updates images work (just dumps RPM content in place, does not run scriptlets or resolve dependencies) or more likely due to the rather aggressive cleanup logic in Lorax at image creation time. It basically cleans up unnecessary files after it installs the requested RPM at image creation time.
Most likely this cut down rootfs is interacting badly with the content of `util-linux-core` being dumped on top of it, effectively breaking the `mount` command in some bizarre way.

I was also thinking about an updates image distributed stub to replace `unshare`, but the custom version of webview script IMHO works better as a short term hotfix.

So given how unsafe is making `unsafe` conditional in `cockpit-desktop` I think we will go the way of making sure the `unsafe` is no cleaned up by Lorax. And if it works, we can then drop the hotfix and switch back to regular `cockpit-desktop`.

martinpitt:
OOI, is this just a hack for testing? You could then just do some `sed -i s/unshare.*--net //'` somewhere in the test setup.

KKoukiou:
> OOI, is this just a hack for testing? You could then just do some `sed -i s/unshare.*--net //'` somewhere in the test setup.

I don't consider it a hack for testing. Users should still be able to boot into the webui mode (aka cockit-desktop spawning) by using the boot.iso? @M4rtinK correct me if I am wrong.

VladimirSlavik:
This is the thing that will be launched if users choose webui, which is actually default for the MVP/public demo, right? If so, using the standard cockpit thing and not erasing unshare in lorax might be the more straightforward option. It's 33K no longer saved, so not too bad, I think...

KKoukiou:
> This is the thing that will be launched if users choose webui, which is actually default for the MVP/public demo, right? If so, using the standard cockpit thing and not erasing unshare in lorax might be the more straightforward option. It's 33K no longer saved, so not too bad, I think...

I agree, but we need first a quick fix, as fixing that on lorax and waiting for release and rawhide update might take a few weeks.

VladimirSlavik:
OK, that makes sense.

Technically, this needs also https://github.com/rhinstaller/anaconda/pull/3976 to make the tests pass. But I would say the errors  fixed by that PR can be ignored here...

VladimirSlavik (approved):

KKoukiou:
Rebase to solve unit test errors.

PR #3969, opened 2022-03-22
Clarify which Anaconda profile is used by CentOS Stream

poncovka:
Extend the description of the related configuration file.


jstodola (commented):
I think I'm confused even more now :-)
My understanding was that `centos.conf` was used for CentOS.
Based on rhel-8 and rhel-9 branches, I would expect that CentOS Stream would have a configuration file `centos-stream.conf` - just to make all the branches in sync a bit. Or is there any reason for not having `centos-stream.conf` in the master branch?

VladimirSlavik (approved):

PR #3971, opened 2022-03-22
Use a task to set up an additional NFS repository

poncovka:
Set up an additional NFS repository with a task. Collect generated mount points,
so we can unmount them when needed. This is just a temporary solution, because
the source management should be eventually handled by DBus source objects.

### DIFF
--- a/pyanaconda/payload/dnf/payload.py
+++ b/pyanaconda/payload/dnf/payload.py
@@ -22,6 +22,7 @@ import dnf.repo
 
 from glob import glob
 
+from pyanaconda.core.path import join_paths
 from pyanaconda.modules.common.errors.installation import NonCriticalInstallationError, \
     InstallationError
 from pyanaconda.modules.common.errors.payload import UnknownRepositoryError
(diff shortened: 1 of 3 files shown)


pyanaconda/payload/dnf/payload.py
```diff
@@ -433,6 +434,39 @@ def _is_existing_repo_configuration(self, data):
...
+                url=self._dnf_manager.substitute(data.url)
+            )
+            mount_point = task.run()
+            data.url = "file://" + mount_point
+
+    def _get_mount_point(self, *paths):
+        """Generate a mount point from specified paths."""
+        mount_point = join_paths(*paths)
+        self._mount_points.append(mount_point)
+        return mount_point
```
VladimirSlavik, line 461: That doesn't really generate the mount point, only its path, I think. So maybe call it `_get_mount_point_path`, and change the docstring for that?

VladimirSlavik (approved):
Thank you, looks good to me.

pyanaconda/payload/dnf/payload.py
poncovka: I have changed it, however, these methods are temporary.

PR #3972, opened 2022-03-22
Use the DNF manager to gather metadata

poncovka:
Reorganize the code that collects metadata about available packages and groups.

### DIFF
--- a/pyanaconda/modules/payloads/payload/dnf/dnf_manager.py
+++ b/pyanaconda/modules/payloads/payload/dnf/dnf_manager.py
@@ -916,6 +916,26 @@ class DNFManager(object):
 
         log.info("Loaded metadata from '%s'.", url)
 
+    def load_packages_metadata(self):
+        """Load metadata about packages in available repositories.
+
+        Load all enabled repositories and process their metadata.
+        It will update the cache that provides information about
+        available packages, modules, groups and environments.
+        """
+        # Load all enabled repositories.
+        # Set up the package sack.
+        self._base.fill_sack(
+            load_system_repo=False,
+            load_available_repos=True,
+        )
+        # Load the comps metadata.
+        self._base.read_comps(
+            arch_filter=True
+        )
+
+        log.info("Loaded packages and group metadata.")
+
     def load_repomd_hashes(self):
         """Load a hash of the repomd.xml file for each enabled repository."""
         self._md_hashes = self._get_repomd_hashes()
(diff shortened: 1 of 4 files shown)


pyanaconda/payload/dnf/payload.py
```diff
@@ -369,26 +369,6 @@ def get_addon_repo(self, repo_id):
...
         return repo
 
-    def _add_repo_to_dnf_and_ks(self, ksrepo):
-        """Add an enabled repo to dnf and kickstart repo lists.
-
-        Add the repo given by the pykickstart Repo object ksrepo to the
-        system.
-
-        Duplicate repos will not raise an error.  They should just silently
-        take the place of the previous value.
```
VladimirSlavik, line 389: This is not exactly documented afterwards, but I guess you plan to refactor that further?

pyanaconda/payload/dnf/payload.py
```diff
@@ -710,7 +674,7 @@ def update_base_repo(self, fallback=True, checkmount=True):
 
         self._include_additional_repositories()
         self._disable_unwanted_repositories()
-        self._load_enabled_repositories()
+        self._validate_enabled_repositories()
```
VladimirSlavik, line 677: It seems repos are loaded when adding, and then here reloaded for validation? If so - how does the performance look? Are the metadata cached after first access, is cache enabled?

pyanaconda/payload/dnf/payload.py
poncovka: This is handled and documented in the `add_repository` method of the `DNFManager` class.

pyanaconda/modules/payloads/payload/dnf/dnf_manager.py
```diff
@@ -906,8 +906,11 @@ def load_repository(self, repo_id):
         repo = self._get_repository(repo_id)
         url = repo.baseurl or repo.mirrorlist or repo.metalink
 
+        if not repo.enabled:
+            log.debug("Cannot load metadata from '%s'. The repo is disabled.", url)
```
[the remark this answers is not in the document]
poncovka: Fixed.

pyanaconda/payload/dnf/payload.py
poncovka: I have extended the documentation of `load_repository`.

pyanaconda/modules/payloads/payload/dnf/dnf_manager.py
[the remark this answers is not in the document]
VladimirSlavik: Sorry, what I had in mind was something like: `log.debug("Skipping loading metadata from disabled repo '%s'.", url)`

I don't want to keep the whole PR blocked on that, so acking anyway.

VladimirSlavik (approved):

PR #3974, opened 2022-03-22
Fix %onerror scripts not invoked in non-interactive mode

sagarun:
When kickstart's 'cmdline' or 'text' mode is used to
automate the installation if an error is raised due to
incorrect storage configuration or hardware error, it
is not handled properly by anaconda and any scripts
in %onerror section of the kickstart file are not invoked.

This change fixes the issue by handling the exception
properly in those scenarios.

Note: This bug only affects rhel7


jkonecny12:
Hi @sagarun, thank you for the change but this will not be merged. The problem is that RHEL 7 is not releasing new ISO images so we can't propagate new changes. Closing.

PR #3976, opened 2022-03-24
ShellCheck: Quote also variables inside ${...}

VladimirSlavik:
New ShellCheck finds new error [SC2295](https://github.com/koalaman/shellcheck/wiki/SC2295): Expansions inside ${...} need to be quoted separately, otherwise they will match as a pattern.

---

This fixes the CI container build on master.


jkonecny12 (approved):
Wow the code starts to be more and more cryptographic... but I believe it's the safe and correct way. Thanks!

M4rtinK (approved):
That second one, OMG. Oh well, Bash gotta Bash... Thanks for the fix! :)

PR #3978, opened 2022-03-25
Fix failing pylint check.

rvykydal:
Pylint started to fail with used-before-assignment so initialize the
variable.

The failure in container refresh workflow:
https://github.com/rhinstaller/anaconda/runs/5685236770?check_suite_focus=true


pyanaconda/core/util.py
```diff
@@ -216,6 +216,7 @@ def sigusr1_too_late_handler(num, frame):
     def sigusr1_preexec():
         signal.signal(signal.SIGUSR1, signal.SIG_IGN)
 
+    old_sigalrm_handler = None
     try:
         old_sigusr1_handler = signal.signal(signal.SIGUSR1, sigusr1_success_handler)
```
jkonecny12, line 221: I guess we should also define this `old_sigusr1_handler`?

jkonecny12 (commented):
Could you please declare both options or did I missed the declaration?

pyanaconda/core/util.py
rvykydal: I wanted to do minimal changes, there is more such variables but pylint is not failing on them, I guess because they are  under if: clause.
Frankly, I do not dare to touch this ancient code much.

rvykydal: Well, there is also childproc and that should be it. If you insist, I can add them.

M4rtinK: Agreed with Radek - better keep this in place until we hopefully throw it all away once we switch to Wayland.

rvykydal: Updated

jkonecny12 (approved):
LGTM!

pyanaconda/core/util.py
jkonecny12: I understand that, however, this should be pretty safe change, shouldn't it?

rvykydal: Done

VladimirSlavik:
Thank you, looks good to me. (I touched most of this pretty recently...)

jkonecny12:
Merged f36-devel to f36-release to get this fix also there. All the containers are now green.

PR #3980, opened 2022-03-25
Remove the unused _noop method

poncovka:
The code that was using it was removed in a commit 19ca9b0.


jkonecny12 (approved):
It seems like the most important method in the whole project so let's get rid of it ;).

KKoukiou (approved):

PR #3981, opened 2022-03-25
Use a task to set up an additional HDD repository

poncovka:
Set up an additional HDD repository with a task.

### DIFF
--- a/pyanaconda/core/payload.py
+++ b/pyanaconda/core/payload.py
@@ -21,6 +21,25 @@ from pyanaconda.core.i18n import _
 from pyanaconda.core.regexes import URL_PARSE
 
 
+def parse_hdd_url(url):
+    """Parse HDD URL into components.
+
+    :param str url: a raw URL, including "hd:"
+    :return (str, str): a tuple with a device and a path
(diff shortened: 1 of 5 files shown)


pyanaconda/core/payload.py
```diff
@@ -21,6 +21,25 @@
...
+    """
+    # Remove the prefix.
+    url = url.removeprefix("hd:")
+
+    # Split the specified URL.
+    components = url.split(":", maxsplit=1)
+
+    # Get two components.
+    # Add empty strings if necessary.
+    device, path = (components + [""] * 2)[:2]
```
M4rtinK, line 38: This seems a bit cryptic to me - can't we possibly do this in a more conventional way to make it more readable ?

M4rtinK (approved):
Looks good to me, thanks! :)

Just a small change suggestion.

pyanaconda/core/payload.py
poncovka: Fixed at https://github.com/rhinstaller/anaconda/pull/4251.

PR #3982, opened 2022-03-25
webui: expose webui.remote boot option

KKoukiou:
Previously this was injected by the test VM runner.
However, anaconda contributors that just consume boot.isos for testing while working
on non-linux machines, can only rely on such a boot option to expose the
webui on their host's browser.

### DIFF
--- a/pyanaconda/anaconda.py
+++ b/pyanaconda/anaconda.py
@@ -228,7 +228,11 @@ class Anaconda(object):
         # from kernel boot command line like this
         if "webui" in kernel_arguments:
             from pyanaconda.ui.webui import CockpitUserInterface
-            self._intf = CockpitUserInterface(None, self.payload)
+            self._intf = CockpitUserInterface(
+                None,
+                self.payload,
+                "webui.remote" in kernel_arguments
+            )
 
             # needs to be refreshed now we know if gui or tui will take place
             # FIXME - what about Cockpit based addons ?
--- a/pyanaconda/ui/webui/__init__.py
+++ b/pyanaconda/ui/webui/__init__.py
@@ -30,7 +30,7 @@ log = get_module_logger(__name__)
 class CockpitUserInterface(ui.UserInterface):
     """This is the main class for Cockpit user interface."""
 
-    def __init__(self, storage, payload,
+    def __init__(self, storage, payload, remote,
                  productTitle="Anaconda", isFinal=True,
                  quitMessage=QUIT_MESSAGE):
         """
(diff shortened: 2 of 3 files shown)


jkonecny12 (approved):
Looks good to me.

However, I wonder, what will be the difference between the safe and unsafe solution? I guess there will be password which could be set by this option? If that is true than I would just change it to `inst.webuiremote` and if no parameters are present it's unsafe (no password) if parameter is present it will be the password used for the machine?

KKoukiou:
> Looks good to me.
> 
> However, I wonder, what will be the difference between the safe and unsafe solution? I guess there will be password which could be set by this option? If that is true than I would just change it to `inst.webuiremote` and if no parameters are present it's unsafe (no password) if parameter is present it will be the password used for the machine?

Makes sense.

jkonecny12:
> 

User with default well known password wouldn't be really better than without a password.

KKoukiou:
> > 
> 
> User with default well known password wouldn't be really better than without a password.

The 'login' will performed for the 'root' user, but with the actual credentials on the system. So if we speficy the password via the boot option, we need to let anaconda initialization tasks to change the root's password to the supplied value.

KKoukiou:
@jkonecny12 I changed the boot option name, but not it's type yet, as for the password passing, more things need to be implemented.

jkonecny12 (approved):

poncovka:
Could the boot option has an extra dot, like `inst.webui.remote`? It is a common way of grouping boot options and I find it easier to read. See: `man dracut.cmdline`. Btw. the `inst.sshd` boot option is also unsafe by default.

poncovka (approved):
Looks good to me. Thank you!

pyanaconda/anaconda.py
```diff
@@ -228,7 +228,10 @@ def initInterface(self):
         # from kernel boot command line like this
         if "webui" in kernel_arguments:
             from pyanaconda.ui.webui import CockpitUserInterface
-            self._intf = CockpitUserInterface(None, self.payload)
+            self._intf = CockpitUserInterface(
+                None, self.payload,
+                "webui.remote" in kernel_arguments
+            )
```
VladimirSlavik, line 234: It might be useful to have each parameter on a new line.

VladimirSlavik (approved):
Looks good to me. Thank you!

pyanaconda/anaconda.py
KKoukiou: Updated and repushed.

jkonecny12 (approved):

KKoukiou:
Unrelated kickstart tests faillures, also broken on master.

PR #3986, opened 2022-03-28
Reboot button for webui (+quit)

VladimirSlavik:


ui/webui/src/helpers/exit.js
```diff
@@ -0,0 +1,30 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with This program; If not, see <http://www.gnu.org/licenses/>.
+ */
+import cockpit from "cockpit";
+
+export const exitGui = () => {
+    const pidFile = cockpit.file("/run/anaconda/webui_script.pid", { superuser: "try", });
```
M4rtinK, line 20: `superuser: "try"`

Just wondering, what does this do ? I guess it tries to open the file as root, but what does the "try" mean ? Does it tries also as non root or something else ?

M4rtinK (approved):
Looks good to me - nice, simple & very readable! :)

ui/webui/test/machine_install.py
```diff
@@ -78,7 +78,7 @@ def start(self):
                 "--noautoconsole "
                 f"--graphics vnc,listen={self.ssh_address} "
                 "--extra-args "
-                f"'inst.sshd inst.nokill inst.webui inst.updates=http://10.0.2.2:{http_port}/updates.img' "
+                f"'inst.sshd inst.webui inst.updates=http://10.0.2.2:{http_port}/updates.img' "
```
KKoukiou, line 81: Good catch! I already included it here: https://github.com/rhinstaller/anaconda/pull/3982/commits/27bbd7366aedbdc1e9e1fcbdd2c7d98deb3009a6
 
So we need to rebase this or the other PR.

ui/webui/src/helpers/exit.js
KKoukiou, line 20: Actually this is indeed not needed here. We can send a follow up to remove all 'superuser: true' instances.
This makes sense when running as non-root users ,aka as admin user, which needs sudo to perform privileged operations. superuser:try is the same like running commands with sudo.

ui/webui/src/components/installation/InstallationProgress.jsx
```diff
@@ -96,7 +98,7 @@ export class InstallationProgress extends React.Component {
                     </EmptyStateBody>
                     {status === "success" &&
                     <EmptyStateSecondaryActions>
-                        <Button>{_("Reboot")}</Button>
+                        <Button onClick={exitGui}>{_("Reboot")}</Button>
```
KKoukiou, line 101: Hmm, I was actually expecting to do a real reboot. Aka, cockpit.spawn("reboot -f")

How come killing the gui process does a reboot?

ui/webui/src/helpers/exit.js
```diff
@@ -0,0 +1,30 @@
...
+    pidFile.read()
+            .then(
+                (content, tag) => {
+                    console.log("Script PID to kill:", content);
+                    cockpit.spawn(["kill", content]).then(null);
+                },
+                ex => console.error("Failed to parse anaconda configuration")
```
KKoukiou, line 27: It's not configuration file.

KKoukiou (changes requested):

ui/webui/test/machine_install.py
KKoukiou: Removed it from the other PR. Please add however a more explanatory commit message on why this is no longer needed.
Feel free to copy the commit msg from the commit above.

ui/webui/src/components/installation/InstallationProgress.jsx
VladimirSlavik: Reboot is the final effect, but before that, Anaconda needs to also run some other stuff like reporting IPMI status etc. That is the proper code path to use, regardless of how we want to call it.

As to why it works - check `CockpitUserInterface.run()`. The Anaconda UI process waits until `webui-desktop` is finished. Killing the script lets the process continue, which means it just runs to the end and exits. And that is when `exitHandler` runs and does the cleanup and actual reboot.

ui/webui/src/helpers/exit.js
```diff
@@ -0,0 +1,30 @@
...
+ */
+import cockpit from "cockpit";
+
+export const exitGui = () => {
+    const pidFile = cockpit.file("/run/anaconda/webui_script.pid", { superuser: "try", });
+    pidFile.read()
+            .then(
+                (content, tag) => {
+                    console.log("Script PID to kill:", content);
+                    cockpit.spawn(["kill", content]).then(null);
```
KKoukiou, line 25: cockpit.spawn also returns a promise, which you need to catch in case of errors. what you should better do is to return this cockpit.spawn here, and then add the exception catching, inside a seperate catch, so that you can catch failures from both the pidFile.read() and from the cockpit.spawn()

KKoukiou: Also the then(null) part seems unneeded.

jkonecny12 (commented):
Looks great to me!

ui/webui/src/components/installation/InstallationProgress.jsx
VladimirSlavik: That isn't really possible, I think. Once the script process is killed, anaconda goes through termination, shuts down system with systemctl, and that then probably kills the rest of the processes. The point is that there might be some delay, but it's not clear how much, and it's also not clear if it's possible to run more processes after that happens.

VladimirSlavik:
Rebased + fixed the above, where marked as resolved.

KKoukiou (approved):

VladimirSlavik:
Kickstart test workers are offline due to internal problems :(

PR #3987, opened 2022-03-28
webui: Initial Review screen implementation

M4rtinK:
Revies screen based on latest mockups.
review_screen_1

**TODO**

- [ ] unit tests

### DIFF
--- a/ui/webui/src/components/AnacondaWizard.jsx
+++ b/ui/webui/src/components/AnacondaWizard.jsx
@@ -19,6 +19,7 @@ import cockpit from "cockpit";
 import React, { useContext, useState } from "react";
 
 import {
+    ActionList,
     Alert,
     Button,
     Title,
@@ -30,7 +31,7 @@ import { AddressContext } from "./Common.jsx";
 import { InstallationDestination, applyDefaultStorage } from "./storage/InstallationDestination.jsx";
 import { InstallationLanguage } from "./installation/InstallationLanguage.jsx";
 import { InstallationProgress } from "./installation/InstallationProgress.jsx";
-import { ReviewConfiguration } from "./installation/ReviewConfiguration.jsx";
+import { ReviewConfiguration, ReviewConfigurationFooter } from "./installation/ReviewConfiguration.jsx";
 
 import { exitGui } from "../helpers/exit.js";
 
@@ -87,7 +88,7 @@ export const AnacondaWizard = ({ onAddErrorNotification, title }) => {
         {
             component: ReviewConfiguration,
             id: "review-configuration",
-            label: _("Review"),
+            label: _("Review and install"),
         },
         {
             component: InstallationProgress,
@@ -131,6 +132,7 @@ export const AnacondaWizard = ({ onAddErrorNotification, title }) => {
 
 const Footer = ({ setStepNotification }) => {
     const [isInProgress, setIsInProgress] = useState(false);
+    const [isNextDisabled, setIsNextDisabled] = useState(true);
 
     const goToStep = (activeStep, onNext) => {
         if (activeStep.id === "installation-destination") {
@@ -167,24 +169,30 @@ const Footer = ({ setStepNotification }) => {
                     );
 
                     return (
-                        <>
-                            <Button
-                              variant="primary"
-                              isDisabled={isInProgress}
-                              isLoading={isInProgress}
-                              onClick={() => goToStep(activeStep, onNext)}>
-                                {nextButtonText}
-                            </Button>
-                            <Button
-                              variant="secondary"
-                              isDisabled={isBackDisabled}
-                              onClick={onBack}>
-                                {_("Back")}
-                            </Button>
-                            <Button variant="link" onClick={exitGui}>
-                                {_("Quit")}
-                            </Button>
-                        </>
+                        <Stack hasGutter>
+                            {activeStep.id === "review-configuration" &&
+                            <ReviewConfigurationFooter
+                              setIsEraseChecked={value => setIsNextDisabled(!value)}
+                              isEraseChecked={!isNextDisabled} />}
+                            <ActionList>
+                                <Button
+                                  variant="primary"
+                                  isDisabled={isInProgress || (["review-configuration"].includes(activeStep.id) && isNextDisabled)}
+                                  isLoading={isInProgress}
+                                  onClick={() => goToStep(activeStep, onNext)}>
+                                    {nextButtonText}
+                                </Button>
+                                <Button
+                                  variant="secondary"
+                                  isDisabled={isBackDisabled}
+                                  onClick={onBack}>
+                                    {_("Back")}
+                                </Button>
+                                <Button variant="link" onClick={exitGui}>
+                                    {_("Quit")}
+                                </Button>
+                            </ActionList>
+                        </Stack>
                     );
                 }}
             </WizardContextConsumer>
--- a/ui/webui/src/components/installation/ReviewConfiguration.jsx
+++ b/ui/webui/src/components/installation/ReviewConfiguration.jsx
@@ -14,16 +14,125 @@
  * You should have received a copy of the GNU Lesser General Public License
  * along with This program; If not, see <http://www.gnu.org/licenses/>.
  */
-import React from "react";
+import cockpit from "cockpit";
+import React, { useEffect, useState } from "react";
 
 import {
-    Form
+    Checkbox, Title,
+    DataList, DataListItem,
+    DataListItemRow, DataListItemCells,
+    DataListCell,
+    DescriptionList, DescriptionListGroup,
+    DescriptionListTerm, DescriptionListDescription,
+    Flex, Popover,
 } from "@patternfly/react-core";
 
+import {
+    getSelectedDisks, getDeviceData,
+} from "../../apis/storage.js";
+
+import {
+    getLanguage, getLanguageData,
+} from "../../apis/localization.js";
+
+import { HelpIcon } from "@patternfly/react-icons";
+
+const _ = cockpit.gettext;
+
 export const ReviewConfiguration = () => {
+    const [deviceData, setDeviceData] = useState({});
+    const [selectedDisks, setSelectedDisks] = useState();
+    const [systemLanguage, setSystemLanguage] = useState();
+
+    useEffect(() => {
+        getLanguage()
+                .then(res => {
+                    getLanguageData({ lang: res }).then(res => {
+                        setSystemLanguage(res["native-name"].v);
+                    }, console.error);
+                }, console.error);
+        getSelectedDisks()
+                .then(res => {
+                    setSelectedDisks(res);
+                    // get detailed data for the selected disks
+                    res.forEach(disk => {
+                        getDeviceData({ disk })
+                                .then(res => {
+                                    setDeviceData(d => ({ ...d, [disk]: res[0] }));
+                                }, console.error);
+                    });
+                }, console.error);
+    }, []);
+
+    // handle case of disks not (yet) loaded
+    if (!selectedDisks || !systemLanguage) {
+        return null;
+    }
+
     return (
-        <Form isHorizontal>
-            Imagine configured values for review here.
-        </Form>
+        <>
+            <DescriptionList isHorizontal>
+                <DescriptionListGroup>
+                    <DescriptionListTerm>
+                        {_("Language")}
+                    </DescriptionListTerm>
+                    <DescriptionListDescription id="installation-review-target-system-language">
+                        {systemLanguage}
+                    </DescriptionListDescription>
+                </DescriptionListGroup>
+            </DescriptionList>
+            <Title headingLevel="h3" size="xl">
+                {_("Installation destination")}
+            </Title>
+            <DataList>
+                {selectedDisks.map(selectedDisk => (
+                    <DataListItem key={selectedDisk}>
+                        <DataListItemRow>
+                            <DataListItemCells
+                              dataListCells={[
+                                  <DataListCell key={selectedDisk} id={"installation-review-disk-label-" + selectedDisk}>
+                                      {_("Local standard disk")}
+                                  </DataListCell>,
+                                  <DataListCell key={"description-" + selectedDisk} id={"installation-review-disk-description-" + selectedDisk}>
+                                      {deviceData && deviceData[selectedDisk] && deviceData[selectedDisk].description.v + " (" + selectedDisks + ")"}
+                                  </DataListCell>,
+                                  <DataListCell key={"size-" + selectedDisk} id={"installation-review-disk-size-" + selectedDisk}>
+                                      {cockpit.format_bytes(deviceData && deviceData[selectedDisk] && deviceData[selectedDisk].size.v) + " " + _("total")}
+                                  </DataListCell>
+                              ]}
+                            />
+                        </DataListItemRow>
+                    </DataListItem>
+                ))}
+            </DataList>
+        </>
+    );
+};
+
+export const ReviewConfigurationFooter = ({ isEraseChecked, setIsEraseChecked }) => {
+    return (
+        <Flex alignItems={{ default: "alignItemsCenter" }}>
+            <Checkbox
+              label={_("Erase disks to install")}
+              isChecked={isEraseChecked}
+              onChange={setIsEraseChecked}
+              id="installation-review-disk-erase-confirm"
+            />
+            <Popover
+              headerContent={
+                  _("All data on the listed disks will be erased")
+              }
+              bodyContent={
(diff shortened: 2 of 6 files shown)


ui/webui/src/components/installation/ReviewConfiguration.jsx
```diff
@@ -14,16 +14,143 @@
...
+
 export const ReviewConfiguration = () => {
+    const [systemLanguage, setSystemLanguage] = useState("YYY");
+    const [selectedDisks, setSelectedDisks] = useState();
+    const [deviceData, setDeviceData] = useState({});
+
+    useEffect(() => {
+        getLanguage()
+                .then(res => {
+                    console.info("language code");
```
KKoukiou, line 50: Remove all console.info please.

ui/webui/src/components/installation/ReviewConfiguration.jsx
```diff
@@ -14,16 +14,143 @@
...
+                    console.info({ res });
+                    getLanguageData({ lang: res }).then(res => {
+                        console.info("language data");
+                        console.info({ res });
+                        setSystemLanguage(res["native-name"].v);
+                    }, console.error);
+                }, console.error);
+    }, []);
+
+    useEffect(() => {
```
KKoukiou, line 60: You can use a single useEffect call. Just run both apis in the body of the single useEffect, one after the other.

ui/webui/src/components/installation/ReviewConfiguration.jsx
```diff
@@ -14,16 +14,143 @@
...
+    }, []);
+
+    // handle case of disks not (yet) loaded
+    if (!selectedDisks) {
+        return null;
+    }
+
+    return (
+        <>
+            <Title headingLevel="h2" size="2xl">
```
KKoukiou, line 87: You can remove this now. The titles are auto-set in the app.jsx from the step name. So 
* remove this title from here
* rename the step from 'Review' to 'Review and install' as seen in the mockups

ui/webui/src/components/installation/ReviewConfiguration.jsx
```diff
@@ -14,16 +14,143 @@
...
+        </>
+    );
+};
+
+export const ReviewConfigurationFooter = ({ isEraseChecked, setIsEraseChecked }) => {
     return (
-        <Form isHorizontal>
-            Imagine configured values for review here.
-        </Form>
+        <>
```
KKoukiou, line 130: Unnecessary empty wrapper. Flex can be wrapper by itself.

ui/webui/src/components/installation/ReviewConfiguration.jsx
```diff
@@ -14,16 +14,143 @@
...
+} from "../../apis/localization.js";
+
+import HelpIcon from "@patternfly/react-icons/dist/esm/icons/help-icon";
+
+const _ = cockpit.gettext;
+
 export const ReviewConfiguration = () => {
+    const [systemLanguage, setSystemLanguage] = useState("YYY");
+    const [selectedDisks, setSelectedDisks] = useState();
+    const [deviceData, setDeviceData] = useState({});
```
KKoukiou, line 45: I like to keep these sorted. Sometimes I fail to stick into it, but just giving it as suggestion. since you use Vim, visually selecting the lines, and then typing `:'<,'>sort` will to the trick

ui/webui/src/components/installation/ReviewConfiguration.jsx
```diff
@@ -14,16 +14,143 @@
...
 
+import {
+    getSelectedDisks, getDeviceData,
+} from "../../apis/storage.js";
+
+import {
+    getLanguage, getLanguageData,
+} from "../../apis/localization.js";
+
+import HelpIcon from "@patternfly/react-icons/dist/esm/icons/help-icon";
```
KKoukiou, line 38: import { HelpIcon} from "@patternfly/react-icons"; please.
You should not use this exact path as there is not api guarantee for the path itself.

ui/webui/src/components/installation/ReviewConfiguration.jsx
```diff
@@ -14,16 +14,143 @@
...
+                }, console.error);
+    }, []);
+
+    // handle case of disks not (yet) loaded
+    if (!selectedDisks) {
+        return null;
+    }
+
+    return (
+        <>
```
KKoukiou, line 86: When you remove the TItle you can also remove the <> wrapper.

KKoukiou (changes requested):
Super! looks perfect. Just some cleanup comments.

ui/webui/src/components/AnacondaWizard.jsx
```diff
@@ -164,24 +166,30 @@ const Footer = ({ setStepNotification }) => {
...
-                        </>
+                        <Stack hasGutter>
+                            {activeStep.id === "review-configuration" &&
+                            <ReviewConfigurationFooter
+                              setIsEraseChecked={value => setIsNextDisabled(!value)}
+                              isEraseChecked={!isNextDisabled} />}
+                            <ActionList>
+                                <Button
+                                  variant="primary"
+                                  isDisabled={isInProgress || (["review-configuration"].includes(activeStep.id) && isNextDisabled)}
```
jkonecny12, line 177: Why is the check for includes like this? Wouldn't it be better to check if ``activeStep.id == "review-configuration``?

jkonecny12 (commented):
Looks good to me AFAICT.

ui/webui/src/components/AnacondaWizard.jsx
KKoukiou: Sure, but if more 'steps' in the future will need this kinda of guarding the 'Next' with a checkbox or dialog, this will need to be reverted to what @M4rtinK uses now.

jkonecny12: ok, it give sense.

ui/webui/src/components/installation/ReviewConfiguration.jsx
M4rtinK: Removed.

ui/webui/src/components/installation/ReviewConfiguration.jsx
M4rtinK: OK, changed. :)

ui/webui/src/components/installation/ReviewConfiguration.jsx
M4rtinK: I noticed it being used so I assumed there was some good reason for that. :)

ui/webui/src/components/installation/ReviewConfiguration.jsx
M4rtinK: Sorted. :)

ui/webui/src/components/installation/ReviewConfiguration.jsx
M4rtinK: Removed. :)

What is it doing actually ? Just a generic parent component in case there is not any specific one ?

ui/webui/src/components/installation/ReviewConfiguration.jsx
KKoukiou: When you return JSX you need to always return one element as wrapper. If you want to return two buttons for example you can do:

``` jsx
<> 
  <Button> 1 </Button>
  <Button> 2 </Button>
</>
```

If you already use a component as wrapper then you don't need this 'empty' wrapper. So this is valid code.

``` jsx
<Flex>
  <Button> 1 </Button>
  <Button> 2 </Button>
</Flex>
```

Read more about it here:
https://reactjs.org/docs/fragments.html

Note: <> is the same like React.Fragment

M4rtinK:
Updated based on feedback. Now looks like this:
review_screen_2
Modal dialog instead of the checkbox postponed to follow-up PR. :)

ui/webui/src/components/installation/ReviewConfiguration.jsx
```diff
@@ -14,16 +14,125 @@
...
+} from "../../apis/localization.js";
+
+import { HelpIcon } from "@patternfly/react-icons";
+
+const _ = cockpit.gettext;
+
 export const ReviewConfiguration = () => {
+    const [deviceData, setDeviceData] = useState({});
+    const [selectedDisks, setSelectedDisks] = useState();
+    const [systemLanguage, setSystemLanguage] = useState("YYY");
```
KKoukiou, line 45: I would just leave it unset. YYY is not any valid default value.

ui/webui/src/components/installation/ReviewConfiguration.jsx
```diff
@@ -14,16 +14,125 @@
...
+                        getDeviceData({ disk })
+                                .then(res => {
+                                    setDeviceData(d => ({ ...d, [disk]: res[0] }));
+                                }, console.error);
+                    });
+                }, console.error);
+    }, []);
+
+    // handle case of disks not (yet) loaded
+    if (!selectedDisks) {
```
KKoukiou, line 68: Maybe do 
``` 
if (!selectedDisks || !lang) {
    return null;
}
```

KKoukiou: And remote the YYY default.

ui/webui/src/components/installation/ReviewConfiguration.jsx
KKoukiou: Answered the what is it doing part in a comment above.

ui/webui/src/components/installation/ReviewConfiguration.jsx
M4rtinK: Done.

ui/webui/src/components/installation/ReviewConfiguration.jsx
M4rtinK: Oh, interesting, so essentially short-hand for Fragment. Yeah, I remember from other framework that you need to have on top level thing/parent as well but what confused me that you can just use what essentially looks like empty brackets. :)

ui/webui/src/components/installation/ReviewConfiguration.jsx
```diff
@@ -14,16 +14,125 @@
...
+    const [deviceData, setDeviceData] = useState({});
+    const [selectedDisks, setSelectedDisks] = useState();
+    const [systemLanguage, setSystemLanguage] = useState("YYY");
+
+    useEffect(() => {
+        getLanguage()
+                .then(res => {
+                    getLanguageData({ lang: res }).then(res => {
+                        setSystemLanguage(res["native-name"].v);
+                    }, console.error);
```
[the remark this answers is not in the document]
M4rtinK: That's a good question - I would say it depends on if we have a decent fallback and impact on the installation and target system. Otherwise we might spam the user with cryptic error messages that are hard to act upon.

ui/webui/src/components/installation/ReviewConfiguration.jsx
```diff
@@ -14,16 +14,125 @@
...
+};
+
+export const ReviewConfigurationFooter = ({ isEraseChecked, setIsEraseChecked }) => {
     return (
-        <Form isHorizontal>
-            Imagine configured values for review here.
-        </Form>
+        <Flex alignItems={{ default: "alignItemsCenter" }}>
+            <Checkbox
+              label={_("Erase disks to install")}
```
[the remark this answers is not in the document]
M4rtinK: Yeah, lets discuss on the mockup. :)

ui/webui/src/components/installation/ReviewConfiguration.jsx
M4rtinK: Oh, right, that's just a debugging leftover anyway. :P

ui/webui/src/components/installation/ReviewConfiguration.jsx
M4rtinK: I'm not sure its possible not to have a language set - but lets do that for consistency reasons. :)

ui/webui/src/components/installation/ReviewConfiguration.jsx
```diff
@@ -14,16 +14,125 @@
...
+              headerContent={
+                  _("All data on the listed disks will be erased")
+              }
+              bodyContent={
+                  _("Therefore it is recommended to only proceed if all data on these disks is either backed up or not important.")
+              }
+            >
+                <button
+                  type="button"
+                  className="pf-c-form__group-label-help"
```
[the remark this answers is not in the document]
M4rtinK: Looks like its needed to hide the button outline - without it it looks like this:
Snímek obrazovky 2022-03-31 v 18 35 35

ui/webui/test/check-installation-review
```diff
@@ -0,0 +1,51 @@
...
+            # with the pages basically empty of common elements (as those are provided by the top-level wizard widget)
+            # we at least iterate over them to check this works as expected
+            b.click("button.pf-m-primary")
+            if page == 'installation-language':
+                b.expect_load()
+
+        # check language is shown
+        b.wait_in_text("#installation-review-target-system-language > .pf-c-description-list__text", "English (United States)")
+
+        # check selected disks are shown
```
KKoukiou, line 48: Did you forget to add the check for the disks here?

KKoukiou (commented):
Please add the check for the selected disks, feel free to hardcode the expected vda as for now.

Also needs rebase! Super close! Lets get it in today!

ui/webui/test/check-installation-review
```diff
@@ -0,0 +1,55 @@
...
+sys.path.append(os.path.join(os.path.dirname(TEST_DIR), "bots/machine"))
+
+from testlib import MachineCase, nondestructive, test_main  # pylint: disable=import-error
+from machine_install import VirtInstallMachine
+
+@nondestructive
+class TestReview(MachineCase):
+    MachineCase.machine_class = VirtInstallMachine
+
+    def testReviewScreen(self):
```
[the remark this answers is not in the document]
M4rtinK: Sure, lets do that here as well. :)

ui/webui/test/check-installation-review
```diff
@@ -0,0 +1,53 @@
...
+
+    def testBasic(self):
+        b = self.browser
+
+        b.open("/cockpit/@localhost/anaconda-webui/index.html#/installation-language")
+
+        for page in ['installation-language', 'installation-destination']:
+            # with the pages basically empty of common elements (as those are provided by the top-level wizard widget)
+            # we at least iterate over them to check this works as expected
+            b.click("button.pf-m-primary")
```
KKoukiou, line 41: You can now use the helper classes, by doing:

i = Installer(b)

i.next()

KKoukiou: See how I import these in the other files.

ui/webui/test/check-installation-review
KKoukiou: It's good enough for now. Multi disks can be a followup test.

KKoukiou (commented):
You need to pass the helper function the 'b' variable and also include the path for the python file to be imported.

``` diff
diff --git ui/webui/test/check-installation-review ui/webui/test/check-installation-review
index 62f573c5db..82d7997ca8 100755
--- ui/webui/test/check-installation-review
+++ ui/webui/test/check-installation-review
@@ -20,6 +20,7 @@ import sys
 
 # import Cockpit's machinery for test VMs and its browser test API
 TEST_DIR = os.path.dirname(__file__)
+sys.path.append(TEST_DIR)
 sys.path.append(os.path.join(TEST_DIR, "common"))
 sys.path.append(os.path.join(TEST_DIR, "helpers"))
@@ -34,7 +35,7 @@ class TestReview(MachineCase):
 
     def testBasic(self):
         b = self.browser
-        i = Installer()
+        i = Installer(b)
 
         i.open()

```

ui/webui/src/components/installation/ReviewConfiguration.jsx
```diff
@@ -14,16 +14,125 @@
...
+    }, []);
+
+    // handle case of disks not (yet) loaded
+    if (!selectedDisks || !systemLanguage) {
+        return null;
+    }
+
+    return (
+        <>
+            <DescriptionList>
```
KKoukiou, line 74: The mockup uses the `isHorizontal` variant'

ui/webui/test/check-installation-review
M4rtinK: Oh, right, missed that, works now. :)

KKoukiou (approved):
Sehr gut!

KKoukiou (approved):

M4rtinK:
> Sehr gut!

Danke für die Inspektion und Verbesserungsideen! :)

KKoukiou (approved):

PR #3991, opened 2022-03-30
Enable fingerprint authentication using enable-feature

poncovka:
Don't select a profile to enable fingerprint authentication. Call `enable-feature`
to enable the `with-fingerprint` feature in the currently selected profile instead.

Related: rhbz#2056927
Resolves: rhbz#2069899

### DIFF
--- a/pyanaconda/modules/security/installation.py
+++ b/pyanaconda/modules/security/installation.py
@@ -415,7 +415,7 @@ class ConfigureFingerprintAuthTask(Task):
         log.debug("Enabling fingerprint authentication.")
         run_auth_tool(
             AUTHSELECT_TOOL_PATH,
-            ["select", "sssd", "with-fingerprint", "with-silent-lastlog", "--force"],
+            ["enable-feature", "with-fingerprint"],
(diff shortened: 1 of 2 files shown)


M4rtinK (approved):
Looks good to me, thanks! :)

AdamWill:
For the record, `--enable-feature` has existed since 2018 so this should be OK on all vaguely current Fedoras/RHELs, I think.

jkonecny12 (approved):
LGTM!

pbrezina:
> LGTM. Doesn't Anaconda run `authselect select sssd with-silent-lastlog --force` somewhere also without fingerprint? If yes, it should be removed. Installation of the authselect package will do this automatically in its %post.

Please note that this behavior is [new in Fedora 36](https://fedoraproject.org/wiki/Changes/Make_Authselect_Mandatory#Select_default_profile_when_authselect_is_installed), it will be probably picked up by RHEL-10 but it is not available if earlier Fedora releases nor in RHEL-8/9. So this whole PR should aim at Fedora 36+ only.

PR #3993, opened 2022-03-30
Run restorecon in chroot when handling home dirs

VladimirSlavik:
Running `restorecon` needs chroot to sysroot so as to have correct absolute paths. At the point when users are handled, payload has been already installed, so the SELinux policy is already present and we can just do it.

Related: [rhbz#2069305](https://bugzilla.redhat.com/show_bug.cgi?id=2069305) - to be ported for that, eventually. Note that some parts of this change require Python 3.9+, so this can't be ported 1:1 to RHEL 8 where the problem has been originally reported.

### DIFF
--- a/pyanaconda/core/users.py
+++ b/pyanaconda/core/users.py
@@ -327,7 +327,8 @@ def _reown_homedir(root, homedir, username):
                                         from_ids, to_ids, root + homedir])
 
         # Restore also SELinux contexts
-        util.execWithRedirect("restorecon", ["-r", root + homedir])
+        util.restorecon([homedir], root=root)
+
     except OSError as e:
         log.critical("Unable to change owner of existing home directory: %s", e.strerror)
         raise
@@ -561,4 +562,4 @@ def set_user_ssh_key(username, key, root=None):
     # Only change ownership if we created it
     if not authfile_existed:
         os.chown(authfile, int(uid), int(gid))
-        util.execWithRedirect("restorecon", ["-r", sshdir])
+        util.restorecon([sshdir.removeprefix(root)], root=root)
--- a/pyanaconda/core/util.py
+++ b/pyanaconda/core/util.py
@@ -1013,3 +1013,26 @@ def get_os_release_value(name, sysroot="/"):
     # No value found.
     log.debug("%s not found in os-release files", name[:-1])
     return None
+
+
+def restorecon(paths, root, skip_nonexistent=False):
+    """Try to restore contexts for a list of paths.
+
+    Do not fail if the program does not exist because it was not in the payload, just say so.
+
+    :param [str] paths: list of paths to restore
+    :param str root: root to run in; mandatory because we restore contexts only on the new system
+    :param bool skip_nonexistent: optionally, do not fail if some of the paths do not exist
+    :return bool: did anything run at all
+    """
+    if skip_nonexistent:
+        opts = ["-ir"]
+    else:
+        opts = ["-r"]
+
+    try:
+        execWithRedirect("restorecon", opts + paths, root=root)
+    except FileNotFoundError:
+        return False
+    else:
+        return True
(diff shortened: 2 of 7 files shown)


poncovka (approved):
Looks good to me. Thanks!

VladimirSlavik:
Fixed 2/3 test fails, need to think more about the last one, I don't want to lose testing the real effects...

jkonecny12 (approved):
Looks great to me after the test fix.

jkonecny12:
I wonder why is this not necessary on RHEL-9. I guess it wasn't easy to reproduce it there because it's harder to have bad relabeling?

@VladimirSlavik do you think we should backport this to 9 too?

VladimirSlavik:
Why didn't we see this earlier? No idea. Maybe the label is not tested anywhere. Additionally, the file is reused if it exists, so if something creates it before we try to change it, it might have the right label. Finally, restorecon *does* run, just with a wrong path to the file, so there is no hard error.

Backport to RHEL-9? Maybe. The reported bug could be a problem. (The other call affects only reused /home on "pet" type machines, and I think such users are also the ones likely to turn off SELinux and not notice.)

VladimirSlavik:
Can you please review the part with `def _filter_exec(*args, **kwargs):` ? That's the new part, and probably most controversial.

VladimirSlavik:
I just realized I might need to alter this further. In previous PRs, using `restorecon` had to be an optional thing because it might not be installed with some payloads. The previous version of code ran restorecon from the inst.env., so we knew it exists.

achilleas-k:
I added this info the bug already but I wanted to chime in here as well in case anything is still unclear.

> I wonder why is this not necessary on RHEL-9. I guess it wasn't easy to reproduce it there because it's harder to have bad relabeling?

Exactly, yes.  The edge-case that lead me to discover this was that `useradd` had a bug where it wouldn't apply the correct labels to a home directory when the home dir parameter had a trailing slash.  In RHEL 9.0 this is fixed, so Anaconda's `restorecon` isn't necessary in the general case, but probably still a good fallback for any other edge cases where a home or .ssh directory might not be labelled correctly on creation.

VladimirSlavik:
@achilleas-k Now I'm actually curious - do you know if useradd is *expected* to set the contexts? Because if so, we should drop the restorecon instead.

achilleas-k:
> @achilleas-k Now I'm actually curious - do you know if useradd is _expected_ to set the contexts? Because if so, we should drop the restorecon instead.

https://github.com/shadow-maint/shadow/blob/9cdb5251b6c30487a7d935a1a7827f493249479d/src/useradd.c#L2225-L2230
The `create_home()` function in `useradd` sets the file contexts here.

You can observer the bug I mentioned with the trailing slash pretty easily by running the following in RHEL 8.6 (or older):

```bash
#!/usr/bin/env bash

set -e

sudo useradd -m -d /home/slash/  slash
sudo useradd -m -d /home/noslash noslash

sudo -u slash   mkdir /home/slash/.ssh
sudo -u noslash mkdir /home/noslash/.ssh

sudo ls -ladZ /home/{slash,noslash}/.ssh

rpm -qv shadow-utils
```

For me this results in:
```
drwxr-xr-x. 2 noslash noslash unconfined_u:object_r:ssh_home_t:s0 6 Apr  4 11:37 /home/noslash/.ssh
drwxr-xr-x. 2 slash   slash   unconfined_u:object_r:default_t:s0  6 Apr  4 11:37 /home/slash/.ssh
shadow-utils-4.6-14.el8.x86_64
```

I'm not entirely sure when this was fixed, but in RHEL 9.0 with `shadow-utils` v4.9 this doesn't happen.
Could be https://github.com/shadow-maint/shadow/pull/323 but I haven't confirmed.

VladimirSlavik:
Thank you, that makes it clear. I can also confirm the RHEL 9 behavior is same on Rawhide.

I will split the commit into two, same as the two calls. One needs fixing, technically also on RHEL 8 although probably unused. The other call needs fixing in RHEL-8 and probably removing in upstream.

VladimirSlavik:
Here's an archived version before implementing the changes outlined above: [0001-Run-restorecon-in-chroot-when-handling-home-dirs.zip](https://github.com/rhinstaller/anaconda/files/8513046/0001-Run-restorecon-in-chroot-when-handling-home-dirs.zip)

VladimirSlavik:
I changed this to use a wrapper that solves all the conditions to ignore.

PR #3994, opened 2022-03-30
webui: tests: create some helper test classes

KKoukiou:
This will allow us in the future to share code between the tests.

### DIFF
--- a/ui/webui/test/check-basic
+++ b/ui/webui/test/check-basic
@@ -21,8 +21,10 @@ import sys
 # import Cockpit's machinery for test VMs and its browser test API
 TEST_DIR = os.path.dirname(__file__)
 sys.path.append(os.path.join(TEST_DIR, "common"))
+sys.path.append(os.path.join(TEST_DIR, "helpers"))
 sys.path.append(os.path.join(os.path.dirname(TEST_DIR), "bots/machine"))
 
+from installer import Installer
 from testlib import MachineCase, nondestructive, test_main  # pylint: disable=import-error
 from machine_install import VirtInstallMachine
 
@@ -33,8 +35,9 @@ class TestBasic(MachineCase):
     def testNavigation(self):
         b = self.browser
         m = self.machine
+        i = Installer(b)
 
-        b.open("/cockpit/@localhost/anaconda-webui/index.html#/installation-language")
+        i.open()
 
         self.assertIn(
             b.attr("#app > .pf-c-page", "data-debug"),
@@ -49,13 +52,12 @@ class TestBasic(MachineCase):
             b.wait_not_present("#betang-icon")
 
         for page in ['installation-language', 'installation-destination', 'review-configuration']:
-            # with the pages basically empty of common elements (as those are provided by the top-level wizard widget)
+            # with the pages basically empty of common elements (as those are provided by the top-level installer widget)
             # we at least iterate over them to check this works as expected
-            b.wait_js_cond(f'window.location.hash === "#/{page}"')
-            b.wait_visible("#" + page + ".pf-m-current")
+            i.wait_current_page(page)
             # Do not start the installation in non-destructive tests as this performs non revertible changes
             if page != 'review-configuration':
-                b.click("button.pf-m-primary")
+                i.next()
             if page == 'installation-language':
                 b.expect_load()
 
--- a/ui/webui/test/check-installation-language
+++ b/ui/webui/test/check-installation-language
@@ -21,8 +21,11 @@ import sys
 # import Cockpit's machinery for test VMs and its browser test API
 TEST_DIR = os.path.dirname(__file__)
 sys.path.append(os.path.join(TEST_DIR, "common"))
+sys.path.append(os.path.join(TEST_DIR, "helpers"))
 sys.path.append(os.path.join(os.path.dirname(TEST_DIR), "bots/machine"))
 
+from installer import Installer
+from language import Language
 from testlib import MachineCase, nondestructive, test_main  # pylint: disable=import-error
 
 @nondestructive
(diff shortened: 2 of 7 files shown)


ui/webui/test/check-storage
```diff
@@ -20,30 +20,33 @@ import sys
...
 @nondestructive
-class TestStorage(MachineCase):
+class TestStorage(MachineCase, CommonHelpers, StorageHelpers):
     MachineCase.machine_class = VirtInstallMachine
 
     def testLocalStandardDisks(self):
         b = self.browser
 
-        b.open("/cockpit/@localhost/anaconda-webui/index.html")
+        self.open()
```
jkonecny12, line 39: Will this work?. I mean `CommonHelpers` has the open pointing to ``"/cockpit/@localhost/anaconda-webui/index.html#/installation-language"``.

jkonecny12 (changes requested):
Please move the helper classes to a sub-directory.

ui/webui/test/check-storage
KKoukiou: This is pointing to initial page anyway, to yes, it will work.

KKoukiou:
> Please move the helper classes to a sub-directory.

Moved.

ui/webui/test/check-installation-language
```diff
@@ -21,26 +21,27 @@ import sys
...
+sys.path.append(os.path.join(TEST_DIR, "helpers"))
 sys.path.append(os.path.join(os.path.dirname(TEST_DIR), "bots/machine"))
 
+from common_helpers import CommonHelpers
+from language_helpers import LanguageHelpers
 from testlib import MachineCase, nondestructive, test_main  # pylint: disable=import-error
 
 @nondestructive
-class TestLanguage(MachineCase):
+class TestLanguage(MachineCase, CommonHelpers, LanguageHelpers):
```
poncovka, line 32: Yes, the never ending story of customized test case classes :-) We went through this iteration, settled on composition over inheritance and migrated to `pytest` to have more options in the future. We are stuck with `unittest` in this case, so let's at least use the composition.

ui/webui/test/helpers/common_helpers.py
```diff
@@ -0,0 +1,33 @@
...
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with this program; If not, see <http://www.gnu.org/licenses/>.
+
+
+class CommonHelpers():
+    # pylint: disable=no-member
+
+    def begin_installation(self):
+        self.browser.click("button:contains('Begin installation')")
```
poncovka, line 23: Since all these helper functions use the browser object anyway, wouldn't be easier to create wrappers for the browser? So instead of using methods of `CommonHelpers`, we could create `i = Installer(b)` and `l = LanguageScreen(b)` instead of `LanguageHelpers`. What do you think?

poncovka (commented):
These types of classes are difficult to write and maintain.

ui/webui/test/helpers/common_helpers.py
KKoukiou: I like this approach.

KKoukiou: However, these class instances will need to be created once per test.

ui/webui/test/check-installation-language
jkonecny12: Good point, I totally forget about that. It could result in nasty issues with the false test detection and similar other issues.

ui/webui/test/helpers/common_helpers.py
poncovka: > However, these class instances will need to be created once per test. It will slightly increase the amount of code, but it's totally fine from my side.

I think that's fine.

KKoukiou: oh let me adapt also the class names, according to the suggestion.

ui/webui/test/helpers/wizard.py
```diff
@@ -0,0 +1,34 @@
...
+# This program is distributed in the hope that it will be useful, but
+# WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+# Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with this program; If not, see <http://www.gnu.org/licenses/>.
+
+
+class Wizard():
```
poncovka, line 19: I wonder if we shouldn't try to be more abstract. Is it important for the test that this is a wizard with steps? Don't we just want to control the installation?

Here is an example of an anabot test. The example is extreme, but we could get closer to this level of abstraction:
https://github.com/rhinstaller/kickstart-tests/blob/master/anabot-1.xml

velezd: This is hard to compare, because previously we were entering and leaving spokes as defined in the xml. That should be mostly also possible in the wizard, but we still want to have the option to go through it linearly, then go back etc. (and I guess that will be the default for most e2e tests).

poncovka: I just wanted to rename the classes to something more abstract (like `Installation` or `Installer` instead of `Wizard`). That's all.

KKoukiou: nothing against that, updating.

ui/webui/test/check-basic
```diff
@@ -13,16 +13,18 @@
 # Lesser General Public License for more details.
 #
 # You should have received a copy of the GNU Lesser General Public License
-# along with this program; If not, see <http://www.gnu.org/licenses/>.
+# along with this program; If not, see <http://wwi.gnu.org/licenses/>.
```
poncovka, line 16: Typo :-)

poncovka: This typo is actually in every test.

poncovka (approved):

ui/webui/test/check-basic
KKoukiou: Oh, it was probably some vim seddery gone wrong :D

KKoukiou: Fixed.

poncovka (approved):

M4rtinK (approved):
Looks very good to me, thanks a lot! :)

This should IMHO really make more complex tests manageable in the future & make testing behaviour variants efficient. :)

jkonecny12 (approved):
Looks great to me.

PR #3995, opened 2022-03-30
Manage system repositories with the DNF manager

poncovka:
* Call `read_system_repositories` to read repositories from the installation
  environment. All these repositories will be disabled by default.
* Call `restore_system_repositories` to re-enable system repositories if the
  installation source requires to use them.
* Enable or disable updates repositories after we restored the system ones.
* Disable some of the system repositories by default.

### DIFF
--- a/pyanaconda/modules/payloads/payload/dnf/dnf_manager.py
+++ b/pyanaconda/modules/payloads/payload/dnf/dnf_manager.py
@@ -900,6 +900,10 @@ class DNFManager(object):
         """
         repo = self._get_repository(repo_id)
 
+        # Skip if the repository is already set to the right value.
+        if repo.enabled == enabled:
+            return
+
         if enabled:
             repo.enable()
             log.info("The '%s' repository is enabled.", repo_id)
--- a/tests/unit_tests/pyanaconda_tests/modules/payloads/payload/test_module_payload_dnf_manager.py
+++ b/tests/unit_tests/pyanaconda_tests/modules/payloads/payload/test_module_payload_dnf_manager.py
@@ -981,14 +981,33 @@ class DNFManagerReposTestCase(unittest.TestCase):
 
     def test_set_repository_enabled(self):
         """Test the set_repository_enabled function."""
-        self._add_repo("r1")
+        self._add_repo("r1").disable()
 
-        self.dnf_manager.set_repository_enabled("r1", True)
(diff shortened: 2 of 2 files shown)


pyanaconda/payload/dnf/payload.py
```diff
@@ -684,15 +676,39 @@ def update_base_repo(self, fallback=True, checkmount=True):
...
 
+    def _enable_system_repositories(self):
+        """Enable system repositories."""
+        self._dnf_manager.restore_system_repositories()
+
+        log.debug("Enable or disable updates repositories.")
+        self._set_repositories_enabled(conf.payload.updates_repositories, self._updates_enabled)
+
+        log.debug("Disable repositories based on the Anaconda configuration file.")
+        self._set_repositories_enabled(conf.payload.disabled_repositories, False)
```
jkonecny12, line 692: I don't think it's appropriate name of the method. It does not only enable system repositories but also disable them based on the Anaconda configuration.

I would change it more to `load_system_repositories` or something similar and mention this fact in the comment.

pyanaconda/modules/payloads/payload/dnf/dnf_manager.py
```diff
@@ -892,6 +895,44 @@ def set_repository_enabled(self, repo_id, enabled):
             repo.disable()
             log.info("The '%s' repository is disabled.", repo_id)
 
+    def read_system_repositories(self):
```
jkonecny12, line 914: Similar issue as with the next comment. I see the repositories disabling as side effect of reading the repositories. The method name clearly states that it reads the repository but it doesn't tell you anything about disabling them.

I guess the best way to solve this is to split these to `read_system_repositories` and `disable_system_repositories`?

jkonecny12 (approved):
Looks good to me except some naming.

pyanaconda/payload/dnf/payload.py
poncovka: Extended the doc string.

pyanaconda/modules/payloads/payload/dnf/dnf_manager.py
poncovka: The system repositories should be always available and disabled by default, unless they are required by the installation source or enabled by a user. Introducing additional states for the system repositories can cause problems.

PR #3998, opened 2022-03-31
webui: tests: add support for adding a second disk to the test VM

KKoukiou:
webui: add support and tests for offering more than one disks in the UI
    
    This commit adds add_disk method to the machine_install script.
    Machine.add_disk() is a method specific to our custom VM creation scipt and
    not part of MachineCase, and thus is not available for @nondestructive tests.
    
    This commit also fixes displaying more than one disks which was
    previously broken generally.
    
    Since the disk get's added after the anaconda initialization is
    finished, we also added a 'disks rescan' button in the UI, as
    implemented in the mockups, to allow us re-scan available disks.
    
    Note: This commit can be used as a reference to also add other device types on
    request to the test VM.

Screen Shot 2022-03-31 at 17 26 29

### DIFF
--- a/ui/webui/src/components/storage/InstallationDestination.jsx
+++ b/ui/webui/src/components/storage/InstallationDestination.jsx
@@ -18,13 +18,16 @@ import cockpit from "cockpit";
 import React, { useEffect, useState } from "react";
 
 import {
-    Flex,
+    Button,
+    Flex, FlexItem,
     HelperText, HelperTextItem,
     Label,
     Title,
 } from "@patternfly/react-core";
 
 import { EmptyStatePanel } from "cockpit-components-empty-state.jsx";
+import { SyncAltIcon } from "@patternfly/react-icons";
+
 import { ListingTable } from "cockpit-components-table.jsx";
 
 import {
@@ -36,7 +39,9 @@ import {
     getDiskTotalSpace,
     getUsableDisks,
     partitioningConfigureWithTask,
+    resetPartitioning,
     runStorageTask,
+    scanDevicesWithTask,
     setInitializationMode,
     setInitializeLabelsEnabled,
     setSelectedDisks,
@@ -79,6 +84,7 @@ const selectDefaultDisks = ({ ignoredDisks, selectedDisks, usableDisks }) => {
 const LocalStandardDisks = ({ onAddErrorNotification }) => {
     const [deviceData, setDeviceData] = useState({});
     const [disks, setDisks] = useState({});
+    const [refreshCnt, setRefreshCnt] = useState(0);
 
     useEffect(() => {
         let usableDisks;
@@ -94,10 +100,10 @@ const LocalStandardDisks = ({ onAddErrorNotification }) => {
                         selectedDisks: props[0].SelectedDisks.v,
                         usableDisks,
                     });
-                    setDisks(defaultDisks.reduce((acc, cur) => ({ ...acc, [cur]: true }), {}));
+                    setDisks(usableDisks.reduce((acc, cur) => ({ ...acc, [cur]: defaultDisks.includes(cur) }), {}));
 
                     // Show disks data
-                    defaultDisks.forEach(disk => {
+                    usableDisks.forEach(disk => {
                         let deviceData = {};
                         const diskNames = [disk];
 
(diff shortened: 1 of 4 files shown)


ui/webui/test/check-storage
```diff
@@ -48,5 +48,31 @@ class TestStorage(MachineCase):
...
+# We can't run this test case on an existing machine,
+# with --machine because MachineCase is not aware of add_disk method
+class TestStorageExtraDisks(MachineCase):
+    MachineCase.machine_class = VirtInstallMachine
+
+    def testLocalDisksSyncNew(self):
+        b = self.browser
+        m = self.machine
+
+        m.add_disk(5)
```
VladimirSlavik, line 60: Is this the right place to add the disk? I would expect it to be added between `b.wait_not_present("#vdb")` and `b.click("#rescan-disks")`.

ui/webui/test/machine_install.py
```diff
@@ -113,3 +113,9 @@ def kill(self):
         )
         if self.http_server:
             self.http_server.kill()
+
+    def add_disk(self, size=None):
```
M4rtinK, line 117: I think this will fail if size is not set ? I think there should either be a default value or the argument should be required, not optional.

M4rtinK (approved):
Looks god to me, just a small suggestion to improve parameter passing to add_disk(). :)

VladimirSlavik (commented):
Thank you, code LGTM except for nitpicks.

However, I think the commit name and description are "backwards". There are really three things here, and I think the order of importance is:

1) add the reload button (user facing feature),
2) add test for that,
3) extend the test framework to enable that test.

So maybe the commit name should be something like "Add support for reloading storage" ?

ui/webui/test/machine_install.py
KKoukiou: Good point, will update

ui/webui/test/check-storage
KKoukiou: The point is that even if you add the disk  before loading the page, since initialization of modules is done, it will not appear in the UI, until we click rescan.

So the current command order is testing exactly what I wanted.

I can add the wait_not_present("#vdb") before the add_disk as well, but that' expected to not be there.

VladimirSlavik: Oh, I see. Maybe add that, or just a comment? It's not clear we are testing that the other disk is missing, too.

VladimirSlavik (approved):
Thank you!

ui/webui/test/check-storage
KKoukiou: Added comment. Note, that actually it does not make any sense to call wait_not_present(#vdb) anywhere before i.next(), because only at that point, we are at the storage step in the wizard, which shows the disks.

KKoukiou:
Unrelated kickstart test failures.

ui/webui/src/components/storage/InstallationDestination.jsx
```diff
@@ -147,6 +153,24 @@ const LocalStandardDisks = ({ onAddErrorNotification }) => {
...
                 </Label>
+                <FlexItem align={{ default: "alignRight" }}>
+                    <Button
+                      aria-label={_("Rescan disks")}
+                      id="rescan-disks"
+                      onClick={() => {
+                          scanDevicesWithTask().then(res => {
+                              runStorageTask({
+                                  task: res[0],
+                                  onSuccess: () => resetPartitioning().then(() => setRefreshCnt(refreshCnt + 1), onAddErrorNotification),
```
jkonecny12, line 164: Why do we need the refreshCnt at all? Is it to inform the component to redraw?

jkonecny12 (approved):
Looks good to me, just a one newbee question.

ui/webui/src/components/storage/InstallationDestination.jsx
VladimirSlavik: Yup, its other occurrence in `useEffect(..., [refreshCnt])` means the callback will be triggered when it changes.

KKoukiou: Exactly! Thanks @VladimirSlavik <3

PR #4000, opened 2022-04-01
Run kickstart tests on PR via permian (#infra)

rvykydal:
The PR would run kickstart tests via permian.

Example of run on my local anaconda fork: https://github.com/rvykydal/anaconda/actions/runs/2076118956 (PR https://github.com/rvykydal/anaconda/pull/17)

Notably this PR imposes an interface to the comment `/kickstart-test` request, defined actually by the https://github.com/rvykydal/kickstart-tests/blob/support-for-kstest-on-PR-with-permian/scripts/generate-permian-query.py
For now it supports list of tests, --testtype, and --skip-testtypes. I think it is sufficient for current usage. We can of course add other features.

Depends on:
- [x] https://github.com/rhinstaller/kickstart-tests/pull/700
- [x] https://github.com/rhinstaller/kickstart-tests/pull/697 (only script for generating testcases in testlib)
- [x] https://gitlab.cee.redhat.com/rhinstaller/builders/-/merge_requests/107

### DIFF
--- a/.github/workflows/kickstart-tests.yml
+++ b/.github/workflows/kickstart-tests.yml
@@ -58,14 +58,16 @@ jobs:
 
           if [ "$TARGET_BRANCH" == "master" ]; then
             echo "::set-output name=skip_tests::skip-on-fedora"
+            echo "::set-output name=platform::fedora_rawhide"
           elif echo "$TARGET_BRANCH" | grep -qE "f[[:digit:]]+-(devel|release)$"; then
             echo "::set-output name=skip_tests::skip-on-fedora"
+            echo "::set-output name=platform::fedora_rawhide"
           elif echo "$TARGET_BRANCH" | grep -qE "rhel-8(\.[[:digit:]]+)?$"; then
             echo "::set-output name=skip_tests::skip-on-rhel,skip-on-rhel-8"
-            echo "::set-output name=optional_test_args::--platform rhel8"
+            echo "::set-output name=platform::rhel8"
           elif echo "$TARGET_BRANCH" | grep -qE "rhel-9(\.[[:digit:]]+)?$"; then
             echo "::set-output name=skip_tests::skip-on-rhel,skip-on-rhel-9"
-            echo "::set-output name=optional_test_args::--platform rhel9"
+            echo "::set-output name=platform::rhel9"
           else
             echo "Branch $TARGET_BRANCH is not supported by kickstart tests yet!"
             exit 1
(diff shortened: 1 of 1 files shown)


.github/workflows/kickstart-tests.yml
```diff
@@ -133,6 +135,15 @@ jobs:
...
           path: kickstart-tests
 
+      - name: Generate test cases
+        working-directory: ./kickstart-tests
+        run: scripts/generate-testcases.py -t ./testlib/test_cases/kstest-template.tc.yaml.j2 . -o ./testlib/test_cases
+
+      - name: Clone Permian repositories
+        run: |
+          git clone https://github.com/rhinstaller/permian.git
+          git clone https://github.com/rhinstaller/tclib.git
```
jkonecny12, line 145: Please use checkout action instead of clone.

jkonecny12 (changes requested):
Please use the checkout action instead of direct git clone. Other than that I like this solution.

.github/workflows/kickstart-tests.yml
rvykydal: Good point, I'll change it.

VladimirSlavik (approved):
Looks good to me. Thank you!

.github/workflows/kickstart-tests.yml
rvykydal: Done

.github/workflows/kickstart-tests.yml
```diff
@@ -133,6 +135,22 @@ jobs:
           repository: rhinstaller/kickstart-tests
           path: kickstart-tests
 
+      - name: Generate test cases
+        working-directory: ./kickstart-tests
+        run: scripts/generate-testcases.py -t ./testlib/test_cases/kstest-template.tc.yaml.j2 . -o ./testlib/test_cases
+
+      - name: Clone Permian repository
+        uses: actions/checkout@v2
```
jkonecny12, line 143: Please use `@v3` here now.

.github/workflows/kickstart-tests.yml
```diff
@@ -133,6 +135,22 @@ jobs:
...
+        with:
+          repository: rhinstaller/permian
+          path: permian
+
+      - name: Clone tclib repository
+        uses: actions/checkout@v2
```
jkonecny12, line 149: The same here.

jkonecny12 (approved):
Aside of the below, it looks good to me.

.github/workflows/kickstart-tests.yml
rvykydal: Updated

.github/workflows/kickstart-tests.yml
rvykydal: Updated

rvykydal:
Marking as blocked. We want to merge (migrate to Permian) only after it is deployed and tested with daily kstests.

jkonecny12 (approved):
Looks even better now.

PR #4001, opened 2022-04-01
webui: fix reboot & quit callback

KKoukiou:


ui/webui/test/check-basic
```diff
@@ -64,5 +64,17 @@ class TestBasic(MachineCase):
...
+class TestQuit(MachineCase):
+    MachineCase.machine_class = VirtInstallMachine
+
+    def testQuitInstallation(self):
+        b = self.browser
+        m = self.machine
+        i = Installer(b)
+
+        i.open()
+        b.click("button:contains(Quit)")
```
VladimirSlavik, line 76: Will this work with translations?

VladimirSlavik (approved):
Thank you! LGTM.

ui/webui/test/check-basic
KKoukiou: If it did change the language, no, it would not work.

That's why it better to add ids, in the react elements, and use the id selectors.

VladimirSlavik: Ah! Sorry. The button does not have an id because I didn't think I'd need one...

KKoukiou:
Something is wrong with the RPM tests, no idea what though.

KKoukiou:
Rebasing PR.

jkonecny12 (approved):
Looks good to me.

PR #4002, opened 2022-04-01
webui: tests: increase abstraction of changing pages in the wizard

velezd:


ui/webui/test/helpers/installer.py
```diff
@@ -17,18 +17,40 @@
...
+        if expecting_success:
+            self.wait_next_page(current_page)
 
-    def next(self):
+    def next(self, expecting_success=True):
+        current_page = self.get_current_page()
         self.browser.click("button:contains(Next)")
+        if expecting_success:
+            if current_page == self._start_page:
+                self.browser.expect_load()
```
KKoukiou, line 40: This will get obsolete from https://github.com/rhinstaller/anaconda/pull/3996 

I can remove it after, if this get's merged first.

ui/webui/test/helpers/installer.py
```diff
@@ -17,18 +17,40 @@
 
 
 class Installer():
+    _pages = ['installation-language',
+              'installation-destination',
+              'review-configuration',
+              'installation-progress']
+    _start_page = 'installation-language'
```
KKoukiou, line 24: You can do _start_page = _pages[0]

ui/webui/test/helpers/installer.py
```diff
@@ -17,18 +17,40 @@
...
-    def begin_installation(self):
+    def begin_installation(self, expecting_success=True):
+        current_page = self.get_current_page()
         self.browser.click("button:contains('Begin installation')")
+        if expecting_success:
+            self.wait_next_page(current_page)
 
-    def next(self):
+    def next(self, expecting_success=True):
+        current_page = self.get_current_page()
```
KKoukiou, line 36: If you return from get_current_page the page index in the _pages array, you can then remove wait_next_page method.

Instead you can do:

``` py
current_page_idx = self.get_current_page()

self.wait_current_page(current_page_idx + 1)

```

VladimirSlavik (commented):
It might be useful to eliminate all `expect_*` calls in another commit, if that's what is happening here?

ui/webui/test/check-review
```diff
@@ -50,5 +50,18 @@ class TestReview(MachineCase):
...
+        i = Installer(b)
+
+        i.open()
+
+        for page in i.steps[:-1]:
+            if i.steps[i.get_current_page_id()] != i.review_id:
+                i.next()
+            else:
+                i.begin_installation(should_fail=True, confirm_erase=False)
+                i.begin_installation()
```
KKoukiou, line 64: This test will actually start the installation, therefore it's becoming a destructive tests.
TestReview test class is marked as non-destructive, so this is also marked as non-destructive, which means it will share the VM with some other test, and make it fail. This is the case that we see in the logs.

The 'confirm' case is already tested in the TestInstallationProgress, as we need to confirm in order to start the installation.

KKoukiou (changes requested):

ui/webui/test/check-review
velezd: I've already noticed the problem with @nondestructive. Note that this test case should verify that the installation doesn't start if you don't confirm the erase.

ui/webui/test/helpers/installer.py
```diff
@@ -26,19 +26,41 @@ class Installer():
...
         self.browser = browser
 
-    def begin_installation(self):
+    def begin_installation(self, should_fail=False, confirm_erase=True):
+        current_step_id = self.get_current_page_id()
         self.browser.click("button:contains('Begin installation')")
-        self.browser.click("#installation-review-disk-erase-confirm")
 
-    def next(self):
+        self.browser.wait_visible("#installation-review-disk-erase-confirm")
```
KKoukiou, line 33: You don't need wait_visible here. 

b.click has the wait_visible check inside.

KKoukiou: https://github.com/cockpit-project/cockpit/blob/main/test/common/testlib.py#L289

ui/webui/test/helpers/installer.py
```diff
@@ -26,19 +26,41 @@ class Installer():
...
+        page = self.browser.eval_js('window.location.hash;').replace('#/', '') or self.steps[0]
+        return self.steps.index(page)
 
     def wait_current_page(self, page):
         self.browser.wait_js_cond(f'window.location.hash === "#/{page}"')
-        self.browser.wait_visible("#" + page + ".pf-m-current")
+        if page == self.progress_id:
+            self.browser.wait_visible("#installation-progress-bar")
+        else:
+            self.browser.wait_visible("#" + page + ".pf-m-current")
```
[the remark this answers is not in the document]
velezd: Right, I forgot to change it

ui/webui/test/helpers/installer.py
```diff
@@ -26,19 +26,41 @@ class Installer():
     def __init__(self, browser):
         self.browser = browser
 
-    def begin_installation(self):
+    def begin_installation(self, should_fail=False, confirm_erase=True):
```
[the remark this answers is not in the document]
velezd: Not in this case, but as you said there could be other reasons for the installation to not start. And I would like should_fail to become common parameter for functions for interacting with the webui. It will make development of future tests easier.

jkonecny12 (commented):
Aside already mentioned - looks good to me.

ui/webui/test/check-review
KKoukiou: Please read the guidelines https://cockpit-project.org/external/source/test/README
especially the section for `Guidelines for writing tests`

KKoukiou: It's fine to non-destructive tests to be very small, as their 'setup' costs nothing but the destructive tests need to be longer.
This test indeed checks that the installation won't start without the 'confirmation dialog', and till this part the test is non-destructive and therefore fine to be so small.

However, the final begin-installation() call makes the test destructive, and also offers nothing more in matters of code coverage, compared to TestInstallationProgress

ui/webui/test/helpers/installer.py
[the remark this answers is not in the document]
VladimirSlavik: That makes sense, thank you.

ui/webui/test/helpers/installer.py
```diff
@@ -26,19 +26,40 @@ class Installer():
...
+        current_step_id = self.get_current_page_id()
         self.browser.click("button:contains('Begin installation')")
-        self.browser.click("#installation-review-disk-erase-confirm")
 
-    def next(self):
+        if confirm_erase:
+            self.browser.click("#installation-review-disk-erase-confirm")
```
KKoukiou, line 34: Tihs can be parameterized: 
            self.browser.click(f"#{review_id}-disk-erase-confirm")

ui/webui/test/helpers/installer.py
```diff
@@ -26,19 +26,40 @@ class Installer():
...
+
+    def get_current_page_id(self):
+        page = self.browser.eval_js('window.location.hash;').replace('#/', '') or self.steps[0]
+        return self.steps.index(page)
 
     def wait_current_page(self, page):
         self.browser.wait_js_cond(f'window.location.hash === "#/{page}"')
-        self.browser.wait_visible("#" + page + ".pf-m-current")
+        if page == self.progress_id:
+            self.browser.wait_visible("#installation-progress-bar")
```
KKoukiou, line 63: Also can be parameterized.

KKoukiou (approved):
I found two small issues, feel free to ignore.

KKoukiou (changes requested):
The test failures are because the test is missing a cleanup. See https://github.com/rhinstaller/anaconda/blob/master/ui/webui/test/check-review#L48

ui/webui/test/helpers/installer.py
velezd: Interesting, so the ids are put together in a similar way when the page is generated?

KKoukiou: Correct. https://github.com/rhinstaller/anaconda/blob/master/ui/webui/src/components/review/ReviewConfiguration.jsx#L116

PR #4005, opened 2022-04-04
webui: port review screen confirmation from checkbox to modal dialog

KKoukiou:
Screen Shot 2022-04-04 at 10 41 34

### DIFF
--- a/ui/webui/src/components/AnacondaWizard.jsx
+++ b/ui/webui/src/components/AnacondaWizard.jsx
@@ -31,7 +31,7 @@ import { AddressContext } from "./Common.jsx";
 import { InstallationDestination, applyDefaultStorage } from "./storage/InstallationDestination.jsx";
 import { InstallationLanguage } from "./installation/InstallationLanguage.jsx";
 import { InstallationProgress } from "./installation/InstallationProgress.jsx";
-import { ReviewConfiguration, ReviewConfigurationFooter } from "./installation/ReviewConfiguration.jsx";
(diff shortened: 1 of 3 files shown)


M4rtinK (approved):
Looks good to me, thanks! :)

VladimirSlavik (commented):
Do I understand correctly that...

If the current screen is `review-configuration`, there is no going forward on button press, instead `NextWaitConfirmation` is changed to `True`, and the state change triggers a re-build (?) of the whole page, this time resolving the part `nextWaitsConfirmation && <ReviewConfigurationConfirmModal` to the modal, so it shows over the review, and then the callback from that moves the wizard forward a page?

(The part where things are triggered "passively" / "remotely" by state changes is still pretty alien to me.)

KKoukiou:
> Do I understand correctly that...
> 
> If the current screen is `review-configuration`, there is no going forward on button press, instead `NextWaitConfirmation` is changed to `True`, and the state change triggers a re-build (?) of the whole page, this time resolving the part `nextWaitsConfirmation && <ReviewConfigurationConfirmModal` to the modal, so it shows over the review, and then the callback from that moves the wizard forward a page?
> 
> (The part where things are triggered "passively" / "remotely" by state changes is still pretty alien to me.)

Everything what you describe here is exactly what happens, just one small inaccuracy.
When the state variable gets updated (nextWaitsConfirmation), just the part of the page that's rendered from that component , in this case Footer component, and its chldren get's refreshed.  *not* the whole page.

VladimirSlavik (approved):
Looks good to me. Thank you!

PR #4008, opened 2022-04-05
webui: standardize the naming of the ids and classNames

KKoukiou:


ui/webui/test/helpers/installer.py
```diff
@@ -17,6 +17,12 @@
 
 
 class Installer():
+    welcome_id = "installation-language"
+    storage_id = "installation-destination"
+    review_id = "installation-review"
+    progress_id = "installation-progress"
```
poncovka, line 23: Btw. our GUI and TUI screens also have ids, so we might want to reuse them if possible (see https://github.com/rhinstaller/anaconda/commit/ae5e6030339f5b7aa9a5d99a1c94fd99cdc439bb). They are used for showing help (that might not be relevant for Web UI) and we plan to use them for hiding screens on specific products (we use class names at this moment, see https://github.com/rhinstaller/anaconda/blob/master/data/profile.d/fedora-workstation.conf#L21). We will definitely to be able to hide screens in Web UI as well.

KKoukiou: I have nothing against using the same IDs. Since now everything is parameterized this way, it's even trivial to change it.

M4rtinK: That's a good point & the new Web UI IDs actually already look quite similar.

One possible issue though with the Wizard based Web UI could be cross screen dependencies. With hub-and-spoke the screens are reasonably independent, though you still need to know what you are doing (eq. if both user and root configuration screens are hidden, no users, including admin users, can be created).

With the Wizard layout, follow-up screens dependent on previous screens might be generated. Say software selection screen after you select RPM based installation source. Or partitioning screen after you selected top level devices & opted for some sort of custom partitioning.

So while I'm sure we need to be able to configure what is shown in the Web UI as well, it will need to be adjusted to the new layout in some way. :)

M4rtinK (approved):
Looks good to me, thanks! :)

ui/webui/test/helpers/installer.py
KKoukiou: Do you mind that we create a new issue about thinking it over and 'fixing' this as a followup?

VladimirSlavik (approved):
Looks good to me. Thank you!

ui/webui/test/helpers/installer.py
VladimirSlavik: Spokes won't always map to the wizard steps.

KKoukiou: @poncovka created a Jira so that you can track it an assign to someone in the future. https://issues.redhat.com/browse/INSTALLER-2977

PR #4010, opened 2022-04-05
webui: tests: add missing machine_class to the check-language

KKoukiou:
machine_class essentially tells the VM setUp which script to use for
creating the test VM.

Therefore, this is not used when running the tests with --machine
--browser.

It's however used when we run the test with expecting it to create
its own VM, for example:

TEST_OS=fedora-rawhide-boot test/check-language  -st


VladimirSlavik (approved):

KKoukiou:
Merging fast as it's a bug fix.

PR #4011, opened 2022-04-05
webui: tests: introduce webui code coverage

KKoukiou:
This is not ready, as we are not yet building RPMs with the NODE_ENV=developement scenario.

This will be challenging, see https://github.com/cockpit-project/cockpit-machines/pull/650/files for reference.

### DIFF
--- a/ui/webui/Makefile.am
+++ b/ui/webui/Makefile.am
@@ -81,12 +81,13 @@ $(UPDATES_IMG):
 	test/prepare-updates-img
 
 integration-test: bots test/common $(UPDATES_IMG)
-	test/run-tests
+	test/run-tests --coverage
 
 # checkout Cockpit's test API; this has no API stability guarantee, so check out a stable tag
 # when you start a new project, use the latest release, and update it from time to time
+# FIXME: workaround for including code coverage support, not yet contained in a released cockpit version
 test/common:
-	git clone -b 266 --depth=1 https://github.com/cockpit-project/cockpit.git tmp/cockpit
+	git clone -b main --depth=1 https://github.com/cockpit-project/cockpit.git tmp/cockpit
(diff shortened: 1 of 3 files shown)


KKoukiou:
Coverage links not working - this needs work.

ui/webui/Makefile.am
```diff
@@ -81,12 +81,13 @@ $(UPDATES_IMG):
...
 integration-test: bots test/common $(UPDATES_IMG)
-	test/run-tests
+	test/run-tests --coverage
 
 # checkout Cockpit's test API; this has no API stability guarantee, so check out a stable tag
 # when you start a new project, use the latest release, and update it from time to time
+# FIXME: workaround for including code coverage support, not yet contained in a released cockpit version
 test/common:
-	git clone -b 266 --depth=1 https://github.com/cockpit-project/cockpit.git tmp/cockpit
+	git clone -b main --depth=1 https://github.com/cockpit-project/cockpit.git tmp/cockpit
```
M4rtinK, line 90: Is this an intentional change that should be part of the coverage PR ? As (if I understand this correctly) this is essentially unpinning the cockpit tooling version we use for tests I think it should be at least a separate commit.

M4rtinK (approved):
Looks good to me, but please split the unpinning of cockpit to a separate commit, if intentional. :)

ui/webui/Makefile.am
KKoukiou: The issue is that testlib changes to suppor tcode coverage is not released yet in cockpit. See the comment above this line. 

This can be jsut bumped up to 267 next week that cockpit releases.

KKoukiou:
This does not work because its a production build.
Code coverage doesn't work with uglified/minified JS.

PR #4012, opened 2022-04-05
Don't log the output of journalctl -b

poncovka:
Otherwise, this output is written back to journal and that complicates debugging.


VladimirSlavik (approved):
LGTM, thank you! And sorry, I didn't realize that.

I did wonder once why the journal seems to be duplicated, though, so at least that's now clear.

M4rtinK (approved):
Looks good to me - thanks a lot for fixing this! :)

PR #4014, opened 2022-04-05
webui: set modules locale for translating messages from the backend 

KKoukiou:


KKoukiou:
> Thank you, this looks good to me.
> 
> I wonder if there's anything left to make geolocation work - I think this is at least 30% of the whole thing :-)

Ehm, I have no idea. I did not realize this will be anyhow conflicty with your work on geolocation support.

KKoukiou:
> On another note, I wonder what is the lang cookie good for, after this change. If the whole SPA got the currently set language from the backend while loading, then the language is known, no cookie needed, I think? Arguably, the other UIs do not remember the language, either.

The cookie is needed for cockpit's translation mechanism to work. In other word for cockpit to know what language the WebUI strings should be translated to.

VladimirSlavik:
> Ehm, I have no idea. I did not realize this will be anyhow conflicty with your work on geolocation support.

It does not conflict, it makes the work for me :-)

M4rtinK (approved):
Looks good to me, thanks! :)

PR #4016, opened 2022-04-05
Fix unit tests of the treeinfo support

poncovka:
* Define a valid addon variant in the tested .treeinfo example.
* Update and extend unit tests of the treeinfo support.
* Document why the addon variant is not processed.

### DIFF
--- a/pyanaconda/modules/payloads/payload/dnf/tree_info.py
+++ b/pyanaconda/modules/payloads/payload/dnf/tree_info.py
@@ -158,6 +158,7 @@ class TreeInfoMetadata(object):
                 tree_info.load(file_path)
 
             tree_info.validate()
+            log.debug("Loaded treeinfo metadata:\n%s", tree_info.dumps())
 
             # Load the release version.
             release_version = tree_info.release.version.lower()
@@ -166,6 +167,8 @@ class TreeInfoMetadata(object):
             repo_list = []
 
             for name in tree_info.variants:
+                log.debug("Processing the '%s' variant.", name)
+
                 # Get the variant metadata.
                 data = tree_info.variants[name]
 
--- a/pyanaconda/modules/payloads/payload/dnf/tree_info.py
+++ b/pyanaconda/modules/payloads/payload/dnf/tree_info.py
@@ -332,8 +332,8 @@ class TreeInfoMetadata(object):
         repo_md = self._get_base_repository()
 
         if repo_md:
-            log.debug("The treeinfo defines a base repository at: %s", repo_md.path)
-            return repo_md.path
+            log.debug("The treeinfo defines a base repository at: %s", repo_md.absolute_path)
+            return repo_md.absolute_path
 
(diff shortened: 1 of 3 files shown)


pyanaconda/modules/payloads/payload/dnf/tree_info.py
```diff
@@ -158,14 +158,18 @@ def _load_tree_info(self, root_path, file_path=None, file_content=None):
...
 
             # Load the release version.
             release_version = tree_info.release.version.lower()
 
-            # Load the repositories.
+            # Create repositories for variants and optional variants.
+            # Child variants (like addons) will be ignored.
             repo_list = []
 
             for name in tree_info.variants:
```
poncovka, line 170: @jkonecny12, your implementation of the treeinfo support skips child variants (like addons). Is it ok?

VladimirSlavik (approved):
Looks good to me. Thank you!

pyanaconda/modules/payloads/payload/dnf/tree_info.py
jkonecny12: It is.
Reason for that is that no-one requested the support + it shouldn't be used. The addons were used on RHEL-7 .treeinfo and for RHEL-8 it was decided to avoid using addons but rather go with the complete variants.

poncovka: Thanks for the confirmation!

pyanaconda/modules/payloads/payload/dnf/tree_info.py
```diff
@@ -158,6 +158,7 @@ def _load_tree_info(self, root_path, file_path=None, file_content=None):
                 tree_info.load(file_path)
 
             tree_info.validate()
+            log.debug("Loaded treeinfo metadata:\n%s", tree_info.dumps())
```
jkonecny12, line 161: I'm not sure if we want to dump the whole treeinfo. Could be a bit more information and maybe something what user don't want to have as part of the logs?

jkonecny12 (commented):
Looks good to me otherwise.

pyanaconda/modules/payloads/payload/dnf/tree_info.py
poncovka: We already dump other configurations (like DNF conf). It is hard to guess if Anaconda behaves correctly when you have no idea what's in the treeinfo file. I was really surprised that most of our installations actually work with treeinfo repositories, because there is always the base one.

jkonecny12: What do you mean by "the base one"? Like the base repository as we request to have it in the Anaconda?

poncovka: Yes, the Anaconda base repository usually has a treeinfo that defines that repository. See: http://cobra02/trees/rawhide/.treeinfo

PR #4018, opened 2022-04-06
webui: fix typo in the review configuration disk table

KKoukiou:
This was causing the list of selected disks to appear in the disk row
description instead of the disk name.


KKoukiou:
//ickstart-test --testtype smoke

M4rtinK (approved):
Thanks for the fix - wanted to do that but totally forgot. :P Looks good to me. :)

PR #4020, opened 2022-04-06
webui: bump cockpit's tag for fetching src/lib

KKoukiou:
This removes the need to kill cockpit-ws when using `make rsync`:

See [1] cockpit fix for explanation.

[1] https://github.com/cockpit-project/cockpit/commit/1c604c7b791d944b2f3f45364753372bd6ffca60


KKoukiou:
Oh wait; in the anaconda test VM:

[anaconda root@10 ~]# ls /usr/local
ls: cannot access '/usr/local': No such file or directory


@martinpitt ??

martinpitt:
@KKoukiou : That's why the [webpack plugin does an mkdir](https://github.com/cockpit-project/cockpit/commit/1c604c7b791d944b2f3f45364753372bd6ffca60#diff-2eb9911cfce8b6c9c838bdf7bb20998f71734f6b6951ad73ad601e849b02e7c7R12) first. Doesn't that work for you? (It works fine for our other cockpit projects)

KKoukiou:
@martinpitt so this syncs the code to the VM, but restarting cockpit-ws to pick up the changes is still needed for some reason. Any idea why?

VladimirSlavik (commented):
LGTM superficially, but I fear I can't really review this - I don't understand the rest of the machinery.

martinpitt (approved):
FWIW, this is fine - we should regularly update to Cockpit's latest libraries, to pick up fixes and not drift too far apart. I didn't have time to look at the "need to restart cockpit.service" part (my suspicion is that this is just intermediate -- presumably you need to log out and back in rather), but it really can't hurt to update the libs as long as they still work.

KKoukiou:
Just rebased and adjusted the commit message.

PR #4023, opened 2022-04-06
Fix `make dist` with parallel build

martinpitt:
`make dist` (and hence also `make rpms` and friends) fails with parallel
make, either with `make -j4` directly, or by having `MAKEFLAGS=-j4` set:

    NODE_ENV=production /var/home/martin/upstream/anaconda/ui/webui/node_modules/.bin/webpack
    make[6]: *** No rule to make target 'dist', needed by 'distdir-am'.  Stop.

As we dist built files, add a dependency for dist to the built webpack,
so that make waits for it to finish building before collecting the files
for the tarball.

---

Reproducer: `rm -rf ui/webui/dist/; make -j4 dist`

### DIFF
--- a/ui/webui/Makefile.am
+++ b/ui/webui/Makefile.am
@@ -31,6 +31,9 @@ UPDATES_IMG=../../updates.img
 
 export TEST_OS
 
+# we ship built files in dist/
+distdir: $(WEBPACK_TEST)
+
 dist_libexec_SCRIPTS = webui-desktop
 # makes sure it gets built as part of `make` and `make dist`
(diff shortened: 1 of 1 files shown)


martinpitt:
Meh @ [failure](https://logs.cockpit-project.org/logs/pull-4023-20220406-092809-0e29c41b-fedora-rawhide-boot/log.html) -- the autoconf setup is rather arcane. As this apparently didn't annoy anyone else, and I don't have time right now to spend a few hours on modernizing this, I'll retract this for now.

PR #4024, opened 2022-04-06
Fix translations in DBus modules (#2071098)

poncovka:
Make sure that `LANGUAGE`, `LC_ALL` and `LC_MESSAGES` are not set in the environment
of DBus modules. Only `LANG` should be specified. Otherwise, the strings are not
translated based on `LANG`, but based on one of the other variable.

See: https://docs.python.org/3/library/gettext.html#gettext.find
Resolves: rhbz#2071098


M4rtinK (approved):
Looks good to me, thanks! :)

M4rtinK:
And thanks a lot for tracking this down, it must have been really tricky! :)

jkonecny12 (approved):
Looks great to me.

I just wonder if this shouldn't go to `f36-release` because I understand it the way that this is workaround for F36 only and for Rawhide we will go with https://github.com/rhinstaller/anaconda/pull/4026 . Am I incorrect?

poncovka:
Rebased to `f36-release`.

VladimirSlavik:
This should be also ported to master, I think?

PR #4025, opened 2022-04-06
Don't set the global domain to `anaconda`

poncovka:
* We always specify the domain when we call gettext functions.
* The domain is later changed by pykickstart to `pykickstart` anyway.

See: https://github.com/pykickstart/pykickstart/issues/405


VladimirSlavik (approved):
Thank you!

PR #4026, opened 2022-04-06
Don't set LC_ALL for live installations

poncovka:
The commit 4d686f1 was created as a fix for the bug 1169019. However, the fix
was rejected and the commit was supposed to be reverted (see the comment 37).

`LC_ALL` takes precedence over `LANG` during translations. Anaconda eventually
removes `LC_ALL` from its environment to get the expected translations, but
it might still affect processes started before that (see the bug 2071098).

Related: rhbz#2071098

### DIFF
--- a/data/liveinst/liveinst
+++ b/data/liveinst/liveinst
@@ -125,12 +125,6 @@ if [ -n "$UPDATES" ]; then
 
 fi
 
-if [ -z "$LC_ALL" ]; then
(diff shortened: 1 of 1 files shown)


VladimirSlavik (approved):

M4rtinK (approved):

jkonecny12 (approved):
Wow, this was probably a hell to find. Great work!

PR #4027, opened 2022-04-06
Log the environment of DBus modules

poncovka:
Write the environment variables of DBus modules to logs to simplify debugging.


VladimirSlavik (approved):
Thank you, this will be useful!

M4rtinK (approved):
Looks good to me. :)

M4rtinK:
I guess we don't store anything sensitive in the environment variables of the modules right now, but are we sure the environment start otherwise clean & don't get stuff from say user set shell variables/bashrc or similar ? 

Otherwise there would be a chance up with sensitive data being present in logs say when a user starts Anaconda as a process for installation into a directory/image, same thing for reusable build VMs.

jkonecny12:
In that case, what about logging this only with the `inst.debug` flag?

PR #4028, opened 2022-04-06
webui: Use ProgressStepper on the installation progress screen

M4rtinK:
Use the PatternFly ProgressStepper component instead of a progress bar
on the installation progress screen.

**TODO:**
- [x] remove **Installation progress** in upper left
- [x] unit tests

TestInstallationProgress-testBasic-installation-progress-step-fail-pixels

### DIFF
--- a/ui/webui/src/components/installation/InstallationProgress.jsx
+++ b/ui/webui/src/components/installation/InstallationProgress.jsx
@@ -19,9 +19,10 @@ import cockpit from "cockpit";
 import React from "react";
 
 import {
-    Bullseye, Button,
-    Progress,
-    Text,
+    Button, Flex,
+    ProgressStepper, ProgressStep,
+    Text, TextVariants,
+    TextContent,
 } from "@patternfly/react-core";
 import { CheckCircleIcon, ExclamationCircleIcon } from "@patternfly/react-icons";
 import { EmptyStatePanel } from "cockpit-components-empty-state.jsx";
@@ -58,6 +59,23 @@ export class InstallationProgress extends React.Component {
                                             ret => this.setState({ steps: ret.v }),
                                             this.props.onAddErrorNotification
                                         );
+                            // FIXME: hardcoded progress steps
+                            //        - if ProgressStepper turns out to be viable,
+                            //          use a proper DBus API for progress tep discovery
+                            //          and switching
+                            //
+                            // storage
+                            } else if (step <= 3) {
+                                this.setState({ currentProgressStep: 0 });
+                            // payload
+                            } else if (step === 4) {
+                                this.setState({ currentProgressStep: 1 });
+                            // configuration
+                            } else if (step >= 5 && step <= 11) {
+                                this.setState({ currentProgressStep: 2 });
+                            // bootloader
+                            } else if (step >= 12) {
+                                this.setState({ currentProgressStep: 3 });
                             }
                             this.setState({ message, step });
                         });
@@ -68,7 +86,7 @@ export class InstallationProgress extends React.Component {
                             taskProxy.Finish().catch(this.props.onAddErrorNotification);
                         });
                         taskProxy.addEventListener("Succeeded", () => {
-                            this.setState({ status: "success" });
+                            this.setState({ status: "success", currentProgressStep: 4 });
                         });
                     };
                     taskProxy.wait(() => {
@@ -79,8 +97,15 @@ export class InstallationProgress extends React.Component {
     }
 
     render () {
-        const { steps, step, status, message } = this.state;
         const idPrefix = this.props.idPrefix;
+        const { steps, currentProgressStep, status, message } = this.state;
+
+        const progressSteps = [
+            { title: _("Storage configuration"), id: "installation-progress-step-storage" },
+            { title: _("Payload installation"), id: "installation-progress-step-payload" },
+            { title: _("System configuraton"), id: "installation-progress-step-configuration" },
+            { title: _("Boot loader installation"), id: "installation-progress-step-boot-loader" },
+        ];
 
         if (steps === undefined) { return null }
 
@@ -97,23 +122,46 @@ export class InstallationProgress extends React.Component {
         }
 
         return (
-            <Bullseye className={idPrefix + "-status-" + status}>
+            <div id={idPrefix + "-status-" + status}>
                 <EmptyStatePanel
                   icon={icon}
                   loading={!icon}
                   paragraph={
-                      status !== "success"
-                          ? <Progress
-                              id={idPrefix + "-bar"}
-                              label={cockpit.format("$0 of $1", step, steps)}
-                              max={steps}
-                              min={0}
-                              title={message}
-                              value={step}
-                              valueText={cockpit.format("$0 of $1", step, steps)}
-                              variant={status}
-                          />
-                          : <Text>{_("Fedora is successfully installed. To begin using it, reboot your system.")}</Text>
+                      <Flex direction={{ default: "column" }}>
+                          <ProgressStepper isCenterAligned>
+                              {progressSteps.map((progressStep, index) => {
+                                  let variant = "pending";
+                                  let ariaLabel = _("pending step");
+                                  if (index < currentProgressStep) {
+                                      variant = "success";
+                                      ariaLabel = _("completed step");
+                                  } else if (index === currentProgressStep) {
+                                      variant = status === "danger" ? status : "info";
+                                      ariaLabel = _("current step");
+                                  }
+                                  return (
+                                      <ProgressStep
+                                        aria-label={ariaLabel}
+                                        id={idPrefix + "-step-" + index}
+                                        isCurrent={index === currentProgressStep}
+                                        titleId={progressStep.id}
+                                        key={index}
+                                        variant={variant}
+                                      >
+                                          {progressStep.title}
+                                      </ProgressStep>
+                                  );
+                              })}
+                          </ProgressStepper>
+                          <TextContent>
+                              <Text
+                                component={TextVariants.p}
+                                id={idPrefix + "-status-text"}
+                              >
+                                  {message}
+                              </Text>
+                          </TextContent>
+                      </Flex>
                   }
                   secondary={
                       status === "success" &&
@@ -121,7 +169,7 @@ export class InstallationProgress extends React.Component {
                   }
                   title={title}
                 />
-            </Bullseye>
+            </div>
         );
     }
 }
--- a/ui/webui/src/components/installation/InstallationProgress.jsx
+++ b/ui/webui/src/components/installation/InstallationProgress.jsx
@@ -19,12 +19,14 @@ import cockpit from "cockpit";
 import React from "react";
 
 import {
-    Button, Flex,
+    Button,
+    ExpandableSection,
+    Flex,
     ProgressStepper, ProgressStep,
-    Text, TextVariants,
-    TextContent,
 } from "@patternfly/react-core";
 import { CheckCircleIcon, ExclamationCircleIcon } from "@patternfly/react-icons";
+import { LogViewer } from "@patternfly/react-log-viewer";
+
 import { EmptyStatePanel } from "cockpit-components-empty-state.jsx";
 
 import { AddressContext } from "../Common.jsx";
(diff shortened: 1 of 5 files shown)


M4rtinK:
Currently looks like this:
webui_progress_stepper_3

ui/webui/src/components/installation/InstallationProgress.jsx
```diff
@@ -59,6 +58,25 @@ export class InstallationProgress extends React.Component {
                                             this.props.onAddErrorNotification
                                         );
                             }
+                            console.log("MESSAGE");
```
KKoukiou, line 61: please remove console log

ui/webui/src/components/installation/InstallationProgress.jsx
```diff
@@ -59,6 +58,25 @@ export class InstallationProgress extends React.Component {
...
+                            console.log(step);
+                            console.log(message);
+                            // storage
+                            if (step <= 3) {
+                                this.setState({ currentProgressStep: 0 });
+                            }
+                            // payload
+                            if (step === 4) {
```
KKoukiou, line 69: You 'd rather use else if.

In this way if won't go over all conditions, if the first is truthy,

ui/webui/src/components/installation/InstallationProgress.jsx
```diff
@@ -79,8 +98,15 @@ export class InstallationProgress extends React.Component {
     }
 
     render () {
-        const { steps, step, status, message } = this.state;
         const idPrefix = this.props.idPrefix;
+        const { steps, currentProgressStep, status } = this.state;
+
+        const progressSteps = [
+            { title: "Storage configuration", id: "installation-progress-step-storage" },
+            { title: "Payload installation", id: "installation-progress-step-paylod" },
```
KKoukiou, line 106: s/paylod/payload

ui/webui/src/components/installation/InstallationProgress.jsx
```diff
@@ -97,31 +123,40 @@ export class InstallationProgress extends React.Component {
...
-                />
-            </Bullseye>
+            <EmptyStatePanel
+              icon={icon}
+              loading={!icon}
+              paragraph={
+                  <ProgressStepper isCenterAligned>
+                      {progressSteps.map((progressStep, index) => {
+                          let variant = "pending";
+                          const ariaLabel = "pending step";
```
KKoukiou, line 133: aria labels should get translated normally. It's read by screen readers.

ui/webui/src/components/installation/InstallationProgress.jsx
```diff
@@ -97,31 +123,40 @@ export class InstallationProgress extends React.Component {
...
+                              variant = "info";
+                          }
+                          return (
+                              <ProgressStep
+                                id={idPrefix + "-progress-step-" + index}
+                                titleId={progressStep.id}
+                                key={index}
+                                variant={variant}
+                                isCurrent={index === currentProgressStep}
+                                aria-label={ariaLabel}
```
KKoukiou, line 146: Can you sort the properties :) ? Just for the nice looks.

ui/webui/src/components/installation/InstallationProgress.jsx
```diff
@@ -97,31 +123,40 @@ export class InstallationProgress extends React.Component {
...
-                      status === "success" &&
-                      <Button onClick={exitGui}>{_("Reboot")}</Button>
-                  }
-                  title={title}
-                />
-            </Bullseye>
+            <EmptyStatePanel
+              icon={icon}
+              loading={!icon}
+              paragraph={
```
KKoukiou, line 129: Since the paragraph content is SO long, you can store it in a variable and then use:

`paragraph={installationProgressStepper}`

ui/webui/src/components/installation/InstallationProgress.jsx
```diff
@@ -69,6 +87,7 @@ export class InstallationProgress extends React.Component {
                         });
                         taskProxy.addEventListener("Succeeded", () => {
                             this.setState({ status: "success" });
```
KKoukiou, line 89: You can do:
`this.setState({ status: "success", currentProgressStep: 4 });`

KKoukiou: remember every time you call setState the component 're-renders'. React is smart enough to group setStates that are directly after each each, but if you just group them in the code as well, it's a good practice.

KKoukiou (changes requested):

ui/webui/src/components/installation/InstallationProgress.jsx
```diff
@@ -59,6 +58,25 @@ export class InstallationProgress extends React.Component {
                                             this.props.onAddErrorNotification
                                         );
                             }
+                            console.log("MESSAGE");
+                            console.log(step);
+                            console.log(message);
+                            // storage
+                            if (step <= 3) {
```
jstodola, line 65: Will these magical constants be addressed somehow in the future? If yes, there should be a FIXME comment.

ui/webui/src/components/installation/InstallationProgress.jsx
```diff
@@ -59,6 +58,25 @@ export class InstallationProgress extends React.Component {
...
+                            }
+                            // payload
+                            if (step === 4) {
+                                this.setState({ currentProgressStep: 1 });
+                            }
+                            // configuration
+                            if (step >= 5 && step <= 11) {
+                                this.setState({ currentProgressStep: 2 });
+                            }
+                            // bootloader
```
jstodola, line 76: Is the bootloader step really the last step? What about copying logs, for example?
I'm just wondering if bootloader configuration really deserves it's own step, IMO it's just one of the "configuration" substeps, like creating the users or generating initramfs.

ui/webui/src/components/installation/InstallationProgress.jsx
M4rtinK: Removed. :)

ui/webui/src/components/installation/InstallationProgress.jsx
M4rtinK: Fixed, thanks! :)

ui/webui/src/components/installation/InstallationProgress.jsx
M4rtinK: Thanks, I was wondering what all they are used for. BTW, I think we will need to add them in more places later on.

M4rtinK: Anyway, fixed. :)

ui/webui/src/components/installation/InstallationProgress.jsx
M4rtinK: Thanks for the suggestion but I think I'd like to keep the amount of indirection low in the rendering section, at least for now. :)

ui/webui/src/components/installation/InstallationProgress.jsx
M4rtinK: Oh, right, that's a really good point and really should have been noted in the commit message or a code comment. 

So basically the way how the progress stepping is implemented right now is rather tightly coupled to the MVP use case, primarily to get feedback for using the ProgressStepper in the Web UI. If the feedback is positive and we keep it, this will definitely have a proper API. 

Likely both for getting a list of phases (as those can very well change based on whats being installed) as well as proper signals when the installation process switches to the next phase.

ui/webui/src/components/installation/InstallationProgress.jsx
M4rtinK: Agreed, this is just implemented based on the mockup, which has bootloader as the last step. AFAIK we have not yet really discussed the step groupings in detail and the current progress steps should be thus considered preliminary. :)

M4rtinK:
PR updated based on feedback.

KKoukiou (commented):
A final important thing. The mpckups have the step messages as well. https://marvelapp.com/prototype/i1cb8je/screen/85836244 


This previous implementation also did, so it would be a regression to not show them now. Can you please integrate them?

M4rtinK:
> A final important thing. The mpckups have the step messages as well. https://marvelapp.com/prototype/i1cb8je/screen/85836244
> 
> This previous implementation also did, so it would be a regression to not show them now. Can you please integrate them?
Oh right, that was actually not clear to me how this is supposed to work so I have not added it yet. Basically in the mockup there are just placeholder texts for all of the steps so it was not clear to me what actually goes there. 

If this is a static "This stuff does install foo into bar." or if it shows the current status message for the step in progress ? Then I think only the current step should show the message.

M4rtinK:
So I've dropped the text in upper left and added the current step text to below the currently active step:
webui_progress_stepper_4
Looking quite nice I would say. :)

M4rtinK:
The dynamic status message under progress step turns out to be problematic - changing string lenghts make the progress steps jump around. :P

So lets try just showing the status text below the progress stepper instead:

webui_progress_stepper_5

ui/webui/src/components/installation/InstallationProgress.jsx
```diff
@@ -97,31 +121,45 @@ export class InstallationProgress extends React.Component {
...
+                              const ariaLabel = _("pending step");
+                              if (index < currentProgressStep) {
+                                  variant = "success";
+                              } else if (index === currentProgressStep) {
+                                  variant = "info";
+                              }
+                              return (
+                                  <ProgressStep
+                                    aria-label={ariaLabel}
+                                    id={idPrefix + "-progress-step-" + index}
```
KKoukiou, line 141: The idPrefix is installation-progress so the final here would be, installation-progress-progress-step-
You can remove the progress from the middle string.

ui/webui/src/components/installation/InstallationProgress.jsx
```diff
@@ -97,31 +121,45 @@ export class InstallationProgress extends React.Component {
...
+                                    titleId={progressStep.id}
+                                    key={index}
+                                    variant={variant}
+                                  >
+                                      {progressStep.title}
+                                  </ProgressStep>
+                              );
+                          })}
+                      </ProgressStepper>
+                      <Text component={TextVariants.p}>
```
KKoukiou, line 152: I am almost sure, that for the styles for the Text to be applied, you need to wrap it in a TextContent.

ui/webui/src/components/installation/InstallationProgress.jsx
```diff
@@ -97,31 +121,45 @@ export class InstallationProgress extends React.Component {
         }
 
         return (
-            <Bullseye className={idPrefix + "-status-" + status}>
```
KKoukiou, line 100: This ID was used by the custom CSS to make icons having colors https://github.com/rhinstaller/anaconda/blob/master/ui/webui/src/components/installation/InstallationProgress.scss#L8

Please add a <div id={this id } wrapper for the styles to continue working.

ui/webui/src/components/installation/InstallationProgress.jsx
```diff
@@ -79,8 +96,15 @@ export class InstallationProgress extends React.Component {
     }
 
     render () {
-        const { steps, step, status, message } = this.state;
         const idPrefix = this.props.idPrefix;
+        const { steps, currentProgressStep, status, message } = this.state;
+
+        const progressSteps = [
+            { title: _("Storage configuration"), id: "installation-progress-step-storage" },
```
KKoukiou, line 103: You can use the idPrefix here.

id: idPrefix + "-step-storage"

ui/webui/src/components/installation/InstallationProgress.jsx
M4rtinK: Oh, right, fixed. :)

ui/webui/src/components/installation/InstallationProgress.jsx
M4rtinK: Lets do that. :)

ui/webui/src/components/installation/InstallationProgress.jsx
M4rtinK: Interesting, should be fixed now. :)

M4rtinK: Or at least in theory, icon is still grey for some reason. :P

ui/webui/src/components/installation/InstallationProgress.jsx
VladimirSlavik: Bootloader currently generates something and takes forever, so it makes some sense to have it as a major thing.

KKoukiou (approved):
If the tests also agree

VladimirSlavik (approved):
Some of the commits need squashing, otherwise LGTM.

KKoukiou (commented):
Congratulations :) Great UI improvement!

PR #4030, opened 2022-04-08
packit: Fix target names for builds and tests

csomh:
While 'f36' is a valid value for 'dist_git_branches', the correct target
name for Fedora Linux 36 is 'fedora-36', as it is derived from 'copr-cli
list-chroots' output.

Signed-off-by: Hunor Csomortáni <csomh@redhat.com>

### DIFF
--- a/.packit.yml
+++ b/.packit.yml
@@ -35,19 +35,19 @@ jobs:
     trigger: pull_request
     metadata:
       targets:
-        - f36
+        - fedora-36
 
   - job: copr_build
     trigger: pull_request
     metadata:
       targets:
-        - f36
+        - fedora-36
 
   - job: copr_build
     trigger: commit
     metadata:
       targets:
-        - f36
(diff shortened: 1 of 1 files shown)


csomh:
Created packit/packit#1554.

VladimirSlavik:
LGTM, but I am not sure if this should go to the `f36-release` branch, maybe `f36-devel` needs it too. I'll wait for @jkonecny12 since he has better idea of Packit and the branching.

jkonecny12 (approved):
Looks good to me.

jkonecny12:
> LGTM, but I am not sure if this should go to the `f36-release` branch, maybe `f36-devel` needs it too. I'll wait for @jkonecny12 since he has better idea of Packit and the branching.

Ideally no, in that case we would also have to solve backporting to master. This should be enough for now.

PR #4039, opened 2022-04-13
Improve logic of the keyboard spoke complete method

jkonecny12:
It's much easier to be read than what I've created before. Thanks a lot for suggestion Vendy.

Improves code introduced in https://github.com/rhinstaller/anaconda/pull/4037.


poncovka (approved):
Thanks!

PR #4044, opened 2022-04-17
Correct message when ignoring hibernation flag

jblz:
The conditional above this else block is inclusive of the upper bound.

This change updates the message accordingly to clarify.


poncovka (approved):
Thank you!

PR #4048, opened 2022-04-19
webui - Fix pixel test failure due to string being translated

M4rtinK:
Our translation checking pixel test started to fail due to one of the
English strings being translated to German.

So we need to update the reference image in the reference image repo (done)
and also update the reference to the reference repo in Anaconda repo
(this PR).


poncovka (approved):
Thanks!

M4rtinK:
Oops, had old checkout of the submodule so it overwrote some of the files by old ones. :P Try nr. 2. :D

VladimirSlavik (approved):

M4rtinK:
Seems to be all good now, pixel test wise. :)

poncovka (approved):

PR #4049, opened 2022-04-20
Change default swap size for large-memory systems

pablomh:
It doesn't make much sense to have "almost"-32 GiB of swap in systems
with "almost"-64 GiB of RAM but limit it to 4 GiB in systems with
that amount of memory or higher, so stick to that minimum value.

Signed-off-by: Pablo Mendez Hernandez <pablomh@gmail.com>

### DIFF
--- a/pyanaconda/core/storage.py
+++ b/pyanaconda/core/storage.py
@@ -283,7 +283,7 @@ def suggest_swap_size(quiet=False, hibernation=False, disk_space=None):
     # swap(mem) = 2 * mem, if mem < 2 GiB
     #           = mem,     if 2 GiB <= mem < 8 GiB
     #           = mem / 2, if 8 GIB <= mem < 64 GiB
-    #           = 4 GiB,   if mem >= 64 GiB
+    #           = 32 GiB,  if mem >= 64 GiB
     if mem < Size("2 GiB"):
         swap = 2 * mem
 
@@ -294,7 +294,7 @@ def suggest_swap_size(quiet=False, hibernation=False, disk_space=None):
         swap = mem / 2
 
     else:
-        swap = Size("4 GiB")
+        swap = Size("32 GiB")
 
     if hibernation:
         if mem <= sixty_four_gib:


VladimirSlavik (approved):
Thank you, looks good to me.

poncovka:
It looks reasonable, but we will have to verify it. The current solution was implemented in the commit https://github.com/rhinstaller/anaconda/commit/37415063594d00c896ff3c50496582f0c4e9e3d9 to resolve the bug [744129](https://bugzilla.redhat.com/show_bug.cgi?id=744129).  The bug mentions that "the algorithm has been changed according to the chart suggested in the discussion with other teams" and contains a link to https://home.corp.redhat.com/wiki/swap-size-recommendations that doesn't work anymore. However, release notes of this bug contain the table below that corresponds to the current solution, so it is not a bug.

```

RAM        SWAP(*)       Hibernation
=====================================
<2G    +   2xRAM       =   RAM+SWAP
-------|---------------|-------------
2-8G   +   RAM         =   RAM+SWAP
-------|---------------|-------------
8-64G  +   RAM/2       =   RAM+SWAP
-------|---------------|-------------
64G up +   4G          =    SWAP
=====================================

```

@jkonecny12, any suggestions?

VladimirSlavik:
Maybe the whole algorithm should be re-evaluated. The bug from 10 years ago mentions that QE does not have access to a machine with 64 GB of RAM. But that is pretty reasonable for end-user machines these days. Servers are probably out of that chart by now...

pablomh:
My point is that we're allowing 32 GiB of swap in systems with 64 GiB-1 KiB RAM, so to be conservative, we should at least offer
that to systems with >64 GiB.

jkonecny12:
Hi @pablomh, great idea for improvement and based on my personal opinion it makes sense, on the other hand this should be a proposed change in Fedora. Probably a system wide. We should not to just tell that we think it might be fine and just change the default.

Also I'm interested of opinion of other teams @AdamWill (Fedora QE) and @keszybz (systemd) could you please tell us what do you think about this idea?

Right now, I'm setting `blocked` label because we should first start the discussion before we decide.

keszybz:
I think that in general we could and should reevaluate how much swap we assign. This would best be started by some proposal and discussion on fedora-devel.

But this PR is different: it's just a bugfix. The amount of swap created should be a monotonic function of ram, and the jump makes no sense whatsoever. (Exactly as the patch description says.) So I think this can just be merged as any other bugfix without any further ado.

VladimirSlavik:
After some research, the above seems to be compatible with all the KBs etc., as long as we don't forget addressing this completely.

AdamWill:
I don't see any problem with changing this, I agree that there's no obvious reason for the current behaviour.

keszybz:
Can we get this merged? There really isn't much so discuss, it's a clear bugfix.

jkonecny12:
After the arguments here I decided to not block this one by requiring change. Feel free to merge this if it's ready.

poncovka (approved):
Thanks!

PR #4050, opened 2022-04-20
Enable bootloader hiding on RHEL

frozencemetery:
Signed-off-by: Robbie Harwood <rharwood@redhat.com>

PR #4051, opened 2022-04-20
Enable bootloader hiding on RHEL

frozencemetery:
Signed-off-by: Robbie Harwood <rharwood@redhat.com>

[This is #4050 targeted for master so it applies to future as well.]


VladimirSlavik (approved):
Thanks! Acking here too.

PR #4052, opened 2022-04-21
Remove dependency on eslint-plugin-standard

VladimirSlavik:
See https://github.com/standard/standard/issues/1316


martinpitt (approved):
Thanks! Dropping npm requirements is a case of "it compiles, it works".

M4rtinK (approved):
It compiles, lets ship it.

PR #4053, opened 2022-04-21
Run eslint only in dev builds or when requested

VladimirSlavik:
This by default resolves to running it for webui's 'make rsync', but not for anaconda's 'make rpms'.

I'm not sure if order of plugins matters... Things appear to work locally as they should, though.

We should talk about invoking eslint in a more systematic way. I propose we don't want it to run for actual RPM builds (this change), we want it to run with unit tests (#4055).


M4rtinK (approved):
Looks good to me. :)

PR #4054, opened 2022-04-21
webui: tests: Change handling of steps and hide selectors in methods

velezd:
This should improve maintainability. JS selectors should not be used
directly in the unit tests. This PR moves most of them to helper
methods.

### DIFF
--- a/ui/webui/test/helpers/installer.py
+++ b/ui/webui/test/helpers/installer.py
@@ -14,25 +14,32 @@
 #
 # You should have received a copy of the GNU Lesser General Public License
 # along with this program; If not, see <http://www.gnu.org/licenses/>.
+from collections import UserList
+
+
+class InstallerSteps(UserList):
+    WELCOME = "installation-language"
+    STORAGE = "installation-destination"
+    REVIEW = "installation-review"
+    PROGRESS = "installation-progress"
+    _steps_list = (WELCOME, STORAGE, REVIEW, PROGRESS)
+
+    def __init__(self, initlist=_steps_list):
+        super().__init__(initlist)
 
 
 class Installer():
-    welcome_id = "installation-language"
-    storage_id = "installation-destination"
-    review_id = "installation-review"
-    progress_id = "installation-progress"
-    steps = [welcome_id, storage_id, review_id, progress_id]
-
     def __init__(self, browser, machine):
         self.browser = browser
         self.machine = machine
+        self.steps = InstallerSteps()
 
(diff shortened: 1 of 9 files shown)


ui/webui/test/helpers/installer.py
```diff
@@ -16,51 +16,55 @@
...
+    progress = "installation-progress"
+    all = [welcome, storage, review, progress]
+
 
+class Installer():
     def __init__(self, browser, machine):
         self.browser = browser
         self.machine = machine
+        self.steps = InstallerSteps()
+        self.steps_list = self.steps.all
```
VladimirSlavik, line 32: I think it's redundant to have `InstallerSteps.all` and copy it again on the parent as `steps_list`. Why not just do `self.steps.all`?

VladimirSlavik (approved):
Looks good to me. Thank you!

ui/webui/test/helpers/installer.py
poncovka: Or you can make `InstallerSteps` iterable. Then `steps` will work the same way as `steps_list`. The current solution is very confusing.

ui/webui/test/helpers/storage.py
```diff
@@ -15,16 +15,30 @@
...
     def __init__(self, browser, machine):
         self.browser = browser
         self.machine = machine
+        self._step = InstallerSteps.storage
 
     def select_disk(self, disk, selected=True):
-        self.browser.set_checked("#" + disk + " input", selected)
+        self.browser.set_checked(f"#{disk} input", selected)
+
+    def disk_selected(self, disk):
```
poncovka, line 40: We prefer function/method names that start with a verb to avoid a confusion with attributes and properties.

ui/webui/test/helpers/installer.py
```diff
@@ -16,51 +16,55 @@
...
-    welcome_id = "installation-language"
-    storage_id = "installation-destination"
-    review_id = "installation-review"
-    progress_id = "installation-progress"
-    steps = [welcome_id, storage_id, review_id, progress_id]
+class InstallerSteps():
+    welcome = "installation-language"
+    storage = "installation-destination"
+    review = "installation-review"
+    progress = "installation-progress"
```
poncovka, line 23: These are actually constants and used that way, so it would make sense to have them uppercased.

poncovka (commented):
I have some suggestions, but I like the general direction.

poncovka (approved):
Thanks!

M4rtinK (approved):

PR #4055, opened 2022-04-21
Add eslint as an unit test for webui

VladimirSlavik:
SSIA. See also #4053


M4rtinK (approved):
Looks good to me, thanks! :)

PR #4056, opened 2022-04-21
Clean up the code that adds a base repository

poncovka:
Call the `_add_base_repository` method of the DNF payload class to set up a base repository.


VladimirSlavik (approved):
Looks good to me. Thank you!

PR #4057, opened 2022-04-22
webui: Document how to fix failing pixel tests

M4rtinK:
Document how to fix pixel tests failing on PR in Cockpit CI in
a few (hopefully) easy to follow steps with examples.

### DIFF
--- a/docs/testing.rst
+++ b/docs/testing.rst
@@ -81,3 +81,77 @@ When the reason for the breackage is identified there are two options to go forw
     * If the failure comes from an intended change in behaviour adjust Anaconda or the tests
     * If the failure uncovers an actual regression, file bugs for the given Fedora components. If it does not look like the issue will be
       fixed quickly, work around in Anaconda or add a `naughty override file <https://github.com/cockpit-project/bots/tree/main/naughty/>`_, thus marking the expected failure pattern.
+
+Fix Web UI pixel tests
+----------------------
+
+So pixel tests fail on your PR - now what?
+
+There are essentially two options:
+
+    * the test uncovered a bug in your PR that needs to be fixed in the code
+    * your PR changes the visual behavior of the Web UI in a valid manner and the pixel test needs to be fixed
+
+We will cover the second option.
+
+First make sure your PR is rebased on latest revision of the target branch. This is important as the issue you are seeing might have already been fixed by another PR that has since been merged.
+
+If rebasing your PR did not fix the issue, and pixel tests still fail in CI on your PR, you will need to update one or more of the reference images.
+
+::
+
+    $ cd ui/webui
+    $ make -f Makefile.am test/reference
+    $ GITHUB_BASE=rhinstaller/anaconda ./test/common/pixel-tests pull
+
+Next find the failing test in the PR CI test results and find the individual test that is failing. The results page shows an image comparison tool and in its header links to the new screenshot that no longer matches the expected picture:
+
+
+::
+
+    New <something>-fail-pixels.png on the left, reference on the right.
+
+Download the screenshot into the `ui/webui/test/reference` folder and check it replaced an existing file - the new screenshot from the failed test will be named the same as one of the existing pictures that no longer match. For example:
+
+::
+
+    $ cd ui/webui/test/reference
+    $ git diff --stat
+    TestInstallationProgress-testBasic-installation-progress-step-fail-pixels.png | Bin 54445 -> 55628 bytes
+    1 file changed, 0 insertions(+), 0 deletions(-)
+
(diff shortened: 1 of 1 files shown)


docs/testing.rst
```diff
@@ -81,3 +81,76 @@ When the reason for the breackage is identified there are two options to go forw
...
+We will cover the second option.
+
+First make sure your PR is rebased on latest revision of the target branch. This is important as the issue you are seeing might have already been fixed by another PR that has since been merged.
+
+If rebasing your PR did not fix the issue, and pixel tests still fail in CI on your PR, you will need to update one or more of the reference images.
+
+::
+
+    $ cd ui/webui
+    $ GITHUB_BASE=rhinstaller/anaconda ./test/common/pixel-tests pull
```
poncovka, line 104: I am getting `bash: ./test/common/pixel-tests: No such file or directory`. Have you tried it with a clean repository?

VladimirSlavik: Do `make test/common` between `cd` and invoking the script.

docs/testing.rst
```diff
@@ -81,3 +81,76 @@ When the reason for the breackage is identified there are two options to go forw
...
+
+Next find the failing test in the PR CI test results and find the individual test that is failing. The results page shows an image comparison tool and in its header links to the new screenshot that no longer matches the expected picture:
+
+
+::
+
+    New <something>-fail-pixels.png on the left, reference on the right.
+
+Download the screenshot into this screenshot into the `ui/webui/test/reference` folder and check it replaced an existing file, as the new screenshot by definition will be named the same as one of the existing pictures that no longer match. For example:
```
VladimirSlavik, line 113: "Download the screenshot into this screenshot" - Should this say "download the screenshot from the page and save with the right name into the ... directory" ?

VladimirSlavik (approved):
Thank you, LGTM except for nitpicks below.

docs/testing.rst
poncovka: I don't have a Makefile by default:
`make: Nothing to be done for 'test/reference'.`

This seems to work:
`make -f Makefile.am test/reference`

VladimirSlavik: Ouch, I didn't realize that.

This feels more like investigative journalism than documenting...

M4rtinK: Yep, that works on a clean checkout - steps updated.

Still, this should not be needed. IMHO, we really should scrap these nested makefiles or at least move the most often needed targets to the top-level makefile. Also ideally, all the targets should run all necessary initialization steps if at all possible, when invoked.

Otherwise we are going to drown in "and where do I run this ?" and "what do I need to run before running this ?" issues.

poncovka: I don't really care about nested makefiles as long as there is a good documentation.

docs/testing.rst
```diff
@@ -81,3 +81,77 @@ When the reason for the breackage is identified there are two options to go forw
     * If the failure comes from an intended change in behaviour adjust Anaconda or the tests
     * If the failure uncovers an actual regression, file bugs for the given Fedora components. If it does not look like the issue will be
       fixed quickly, work around in Anaconda or add a `naughty override file <https://github.com/cockpit-project/bots/tree/main/naughty/>`_, thus marking the expected failure pattern.
+
+Fix Web UI pixel tests
+----------------------
+
+So pixel tests fail on your PR - now what ?
```
realsobek, line 88: I would remove the empty space before the question mark.

docs/testing.rst
```diff
@@ -81,3 +81,77 @@ When the reason for the breackage is identified there are two options to go forw
...
+
+There are essentially two options:
+
+    * the test uncovered a in your PR bug that needs to be fixed in the code
```
realsobek, line 92: I would change to "uncovered a bug in your PR that".

docs/testing.rst
M4rtinK: Indeed, this was a typo - should be fixed now & I've also changed a bit the structure of the sentence.

docs/testing.rst
M4rtinK: Fixed. :)

docs/testing.rst
M4rtinK: Oops, fixed, thanks! :)

VladimirSlavik (changes requested):
Please don't update the submodule in this PR.

M4rtinK:
> Please don't update the submodule in this PR.

That's actually not intended & not necessary. Should be fixed now, but definitely something to watch for & possibly adjust the tooling if we hit this again.

VladimirSlavik (approved):
Thanks, LGTM again :)

PR #4060, opened 2022-04-25
Clean up the code that includes additional repositories

poncovka:
Clean up the code that includes additional repositories:

* Remove the `_add_repo_to_dnf` method.
* Simplify the `_include_additional_repositories` method.
* Reorganize checks of invalid repo configurations.

If a base repository is invalid and the fallback to a default source is not 
enabled, raise an exception and stop the payload restart with an error.


VladimirSlavik (approved):
Looks good to me. Thank you!

PR #4061, opened 2022-04-25
Split the code for the installation source spoke

poncovka:
Move constants, functions and dialogs of the installation source spoke
to a new file.


VladimirSlavik:
Is this is only moving the code?

poncovka:
> Is this is only moving the code?

Yes.

VladimirSlavik (approved):
In that case, thank you, sounds good to me (didn't look...)

PR #4067, opened 2022-04-26
Add new dependency to our container required by copr from pip (#infra)

jkonecny12:
Without this the pip install will fail.


VladimirSlavik (approved):
Looks good to me. Thank you!

poncovka (approved):

PR #4072, opened 2022-04-28
Move more code from the Installation Source spoke

poncovka:
Move some of the methods to new functions and write some tests for them.

### DIFF
--- a/pyanaconda/ui/gui/spokes/installation_source.py
+++ b/pyanaconda/ui/gui/spokes/installation_source.py
@@ -42,7 +42,7 @@ from pyanaconda.ui.context import context
 from pyanaconda.ui.gui.spokes.lib.installation_source_helpers import validate_proxy, RepoChecks, \
     ProxyDialog, MediaCheckDialog, IsoChooser, BASEREPO_SETUP_MESSAGE, PROTOCOL_HTTP, \
     PROTOCOL_HTTPS, PROTOCOL_FTP, PROTOCOL_NFS, PROTOCOL_FILE, PROTOCOL_MIRROR, REPO_PROTO, \
-    CLICK_FOR_DETAILS, get_unique_repo_name, validate_repo_name
+    CLICK_FOR_DETAILS, get_unique_repo_name, validate_repo_name, check_duplicate_repo_names
 from pyanaconda.ui.helpers import InputCheck, InputCheckHandler, SourceSwitchHandler
 from pyanaconda.ui.lib.subscription import switch_source
 from pyanaconda.ui.gui.helpers import GUISpokeInputCheckHandler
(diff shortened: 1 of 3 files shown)


tests/unit_tests/pyanaconda_tests/ui/test_ui_source_spoke.py
```diff
@@ -135,3 +135,18 @@ def test_validate_repo_url(self):
 
         msg = "Repository my_repo required remote directory"
         assert validate_repo_url(PROTOCOL_NFS, "server:", repo_name="my_repo") == msg
+
+    def test_remove_url_prefix(self):
+        """Test the remove_url_prefix function."""
+        # Supported protocols
+        assert remove_url_prefix(PROTOCOL_HTTP, "http://path") == "path"
+        assert remove_url_prefix(PROTOCOL_HTTPS, "https://path") == "path"
+        assert remove_url_prefix(PROTOCOL_FTP, "ftp://path") == "path"
```
VladimirSlavik, line 144: I think you also need mismatch between protocol and url, eg:
```
assert remove_url_prefix(PROTOCOL_HTTP, "https://path") == "https://path"
```

pyanaconda/ui/gui/spokes/lib/installation_source_helpers.py
```diff
@@ -231,6 +231,32 @@ def validate_repo_url(protocol, path, repo_name=None):
     return InputCheck.CHECK_OK
 
 
+def remove_url_prefix(protocol, url):
+    """Remove the prefix of the given URL.
+
+    If there is a protocol in the URL, and the protocol matches the
+    combo box, just remove it. This makes it more convenient to paste
+    in URLs. It'll probably freak out people who are typing out http://
+    in the box themselves, but why would you do that?  Don't do that.
```
VladimirSlavik, line 240: I'm not sure if we want to duplicate that bit :)

pyanaconda/ui/gui/spokes/installation_source.py
```diff
@@ -811,93 +810,36 @@ def _get_selected_partition(self):
             return store[selected][0]
 
     # Input checks
-
-    # This method is shared by the checks on urlEntry and repoUrlEntry
-    def _check_url(self, inputcheck, combo):
+    def _check_url_entry(self, inputcheck):
```
VladimirSlavik, line 812: It might be useful to extend the commit message with the part where `_check_url` was split into two others which both use the new helper. It wasn't exactly clear from the diff...

pyanaconda/ui/gui/spokes/lib/installation_source_helpers.py
```diff
@@ -163,6 +163,74 @@ def check_duplicate_repo_names(repo_names):
...
+    :param repo_name: a name of the repository or None
+    :return: an error message or None
+    """
+    # If this is HTTP/HTTPS/FTP, use the URL_PARSE regex
+    if protocol in (PROTOCOL_HTTP, PROTOCOL_HTTPS, PROTOCOL_FTP):
+        if not path:
+            if repo_name:
+                return _("Repository %s has empty url") % repo_name
+            else:
+                return _("URL is empty")
```
VladimirSlavik, line 180: I think it could be better to also split the helper into two for repo and non-repo, if you're already refactoring. Having `if repo_name` for each message is weird.

VladimirSlavik (approved):
Thank you, mostly LGTM, I have only suggestions.

pyanaconda/ui/gui/spokes/lib/installation_source_helpers.py
jkonecny12: I would even split based on the top-level `if-elif-else` it's pretty hard to follow this way.

poncovka:
I dropped commits that required changes, so all the remaining ones should be ok.

PR #4074, opened 2022-04-28
Geoloc modularity

VladimirSlavik:
Modularize geolocation. This is a prerequisite for using it in webui.

Draft, remaining TODOs:

- [x] wrap the task inside the module
- [x] actual D-Bus api - `StartGeolocation()`, `GeolocationResult`
- [x] use the api
- [x] tests
- [x] retire normal geoloc
- [x] ~~add deprecation notices for `inst.geoloc` provider values~~ -> Later PR
- [x] ~~remove prefix `provider_` from constants and conf,~~ use raw url in conf
- [x] research passive waiting for network with `PropertiesChanged`


VladimirSlavik:
The commit "[Rename geoloc options](https://github.com/rhinstaller/anaconda/pull/4074/commits/f3e0c7fba01fab3fc47c7e24895aec6c32077210)" is probably useless.

pyanaconda/modules/common/structures/timezone.py
```diff
@@ -82,3 +82,124 @@ def options(self) -> List[Str]:
...
+        instance.city = city
+
+    @property
+    def status(self) -> Str:
+        """Status of the geolocation system and result
+
+        Can be one of:
+            NOT_STARTED
+            RUNNING
+            DONE
```
poncovka, line 124: What about just:

```
IDLE       The geolocation doesn't run.
DETECTING  The geolocation is running.
```

pyanaconda/modules/timezone/geoloc.py
```diff
@@ -0,0 +1,171 @@
...
+from pyanaconda.anaconda_loggers import get_module_logger, get_sensitive_info_logger
+log = get_module_logger(__name__)
+sensitive_info_log = get_sensitive_info_logger()
+
+
+class GeolocationBackend:
+    """Base class for GeoIP backends."""
+    def __init__(self):
+        self._result = GeolocationResult()
+        self._session = requests_session()
```
poncovka, line 35: This should be eventually used as `with requests_session() as session:`.

pyanaconda/modules/timezone/initialization.py
```diff
@@ -0,0 +1,43 @@
...
+from pyanaconda.modules.common.task import Task
+from pyanaconda.modules.timezone.geoloc import provider_id_to_class
+
+from pyanaconda.anaconda_loggers import get_module_logger
+log = get_module_logger(__name__)
+
+
+class GeolocationTask(Task):
+    """Run geolocation"""
+    def __init__(self, provider_id):
```
poncovka, line 27: I really think it would be useful to have the provider in the Anaconda configuration files. I could image that some products (like RHEL) might want to have special requests for the geolocation settings. Based on that, I could write a deprecation plan for dropping the provider from the boot options, so `inst.geoloc` would be just a simple boolean option in the future.

pyanaconda/modules/common/structures/timezone.py
```diff
@@ -82,3 +82,124 @@ def options(self) -> List[Str]:
     @options.setter
     def options(self, value):
         self._options = value
+
+
+class GeolocationResult(DBusData):
+    """Encapsulates the result from geolocation lookup."""
+
+    def __init__(self):
+        self._status = ""
```
poncovka, line 91: The status should be always set.

pyanaconda/modules/timezone/geoloc.py
VladimirSlavik: Even better, there is just a single request to a http target, so no need for a session...

pyanaconda/modules/common/structures/timezone.py
VladimirSlavik: That does not tell me if there is a result or not, though. If I start to rely on presence of data to detect there is something, I could just drop the status too.

pyanaconda/modules/common/structures/timezone.py
VladimirSlavik: I think part of the problem is using the same class to transport data within the module and then out. Perhaps a named tuple would be better for the first part.

However, even after that, any status will be wrong by design in most situations, since you don't know when constructing the object. The constructor would have to have an additional parameter, but that won't work for the D-Bus transport, so it would have to have some default value, and then it could be overlooked all the same because it's optional. So, I think that's a situation where we just can't have it all.

Perhaps dropping the status would be better?

pyanaconda/modules/common/structures/timezone.py
poncovka: Based on the code, we don't need to know that. We just need to know if it is worth it to give the geolocation more time.

pyanaconda/modules/common/structures/timezone.py
poncovka: I agree, the status is problematic. However, it was supposed to be used instead of `wait_for_refresh_to_finish`. What do you suggest?

VladimirSlavik: I guess the new question is, do we want to offer the consumers an option to *not* wait? If not, maybe the result query could include a timeout parameter: `GetGeolocationResult(wait_timeout=GEOLOC_GUI_TIMEOUT)`, then there's always a wait which could turn out to be immediate, and result, which might turn out to be empty. Status is not needed at all.

poncovka: I don't think we want the timeout to be part of the DBus API. It looks like an implementation detail of the UI. We might want to change it in the future. I think we should expose the task with the `GetGeolocationTask()` method. Anything else would be just a copy of some part of the task API anyway.

VladimirSlavik:
This would be the shape of the API, I think.

See top of page for todo list.

VladimirSlavik:
Functionally complete, only tests missing.

VladimirSlavik:
The preparatory work in first four commits has a separate PR now, #4081.

data/anaconda.conf
```diff
@@ -305,6 +305,13 @@ password_policies =
     user (quality 1, length 6, empty)
     luks (quality 1, length 6)
 
+# Which provider should be used to get geolocation data, if any?
+# This is used for automatic language and timezone detection.
+# Valid values:
```
poncovka, line 310: Please, add extra lines above and below this one.

data/anaconda.conf
```diff
@@ -305,6 +305,13 @@ password_policies =
...
+#   provider_fedora_geoip
+#   provider_hostip
+#
+geolocation_provider = provider_fedora_geoip
```
poncovka, line 314: This option should be in the `[Timezone]` section.

pyanaconda/core/configuration/anaconda.py
```diff
@@ -403,6 +403,11 @@ def set_from_opts(self, opts):
         if opts.noverifyssl:
             self.payload._set_option("verify_ssl", not opts.noverifyssl)
 
+        # Set geolocation provider
+        # FIXME: This will be removed once the boot option becomes a boolean
+        if "geoloc" in opts and opts.geoloc:
+            self.ui._set_option("geolocation_provider", opts.geoloc)
```
poncovka, line 409: What does happen if `inst.geoloc` is set to 0?

pyanaconda/core/configuration/ui.py
```diff
@@ -86,6 +86,14 @@ def password_policies(self):
         """
         return self._get_option("password_policies", self._convert_policies)
 
+    @property
+    def geolocation_provider(self):
+        """Geolocation provider.
+
```
poncovka, line 92: Please, copy-paste the description from the `.conf` file to this property.

pyanaconda/core/configuration/anaconda.py
poncovka, line 409: Could you set here `opts.geoloc.removeprefix("provider_")`? If you do that and update the `GEOLOC_PROVIDER_*` constants, it might be enough to get rid of that useless prefix in the configuration files.

pyanaconda/core/constants.py
```diff
@@ -135,9 +134,13 @@
...
 # - the geoloc thread is started early and in most cases will be already done
 #   when GUI finishes its initialization, so no delays will be introduced
-GEOLOC_TIMEOUT = 3
+GEOLOC_GUI_TIMEOUT = 3
 # timeout for the network connection used for geolocation (in seconds)
 GEOLOC_CONNECTION_TIMEOUT = 5
+# statuses of the result for D-Bus API
+GEOLOC_STATUS_NOT_STARTED = "NOT_STARTED"
+GEOLOC_STATUS_RUNNING = "RUNNING"
+GEOLOC_STATUS_DONE = "DONE"
```
poncovka, line 143: Leftover?

pyanaconda/core/constants.py
```diff
@@ -135,9 +134,13 @@
 # - GUI starts this count once it finishes its initialization
 # - the geoloc thread is started early and in most cases will be already done
 #   when GUI finishes its initialization, so no delays will be introduced
-GEOLOC_TIMEOUT = 3
+GEOLOC_GUI_TIMEOUT = 3
```
poncovka, line 137: Shouldn't this timeout disappear?

pyanaconda/modules/timezone/geoloc.py
```diff
@@ -0,0 +1,132 @@
...
+            else:
+                log.error("Geoloc: Fedora GeoIP API lookup failed with status code: %s",
+                          reply.status_code)
+
+        except requests.exceptions.RequestException as exc:
+            log.debug("Geoloc: RequestException for Fedora GeoIP API lookup:\n%s", exc)
+        except ValueError as exc:
+            log.debug("Geoloc: Unable to decode GeoIP JSON:\n%s", exc)
+
+        return None
```
poncovka, line 90: This is up to you, but I would always return some geolocation data even if there is nothing set. In general, it is a good practice that prevents issues with None.

pyanaconda/modules/timezone/geoloc.py
```diff
@@ -0,0 +1,132 @@
...
+
+        except requests.exceptions.RequestException as exc:
+            log.debug("Geoloc: RequestException during Hostip lookup:\n%s", exc)
+        except ValueError as exc:
+            log.debug("Geoloc: Unable to decode Hostip JSON:\n%s", exc)
+
+        return None
+
+
+def provider_id_to_class(provider_id):
```
poncovka, line 122: I am missing a verb here.

pyanaconda/modules/timezone/initialization.py
```diff
@@ -0,0 +1,75 @@
...
+        if not is_module_available(NETWORK):
+            return False
+
+        network = NETWORK.get_proxy()
+
+        waited = 0
+        while waited < timeout:
+            waited = waited + 0.3
+            time.sleep(0.3)
+            if network.Connected:
```
poncovka, line 72: You can wait for the `PropertiesChanged` signal instead of the active waiting if you want to.

pyanaconda/modules/timezone/timezone.py
```diff
@@ -216,3 +220,31 @@ def install_with_tasks(self):
...
+        # still not started
+        if not self._geoloc_task:
+            return empty_result
+
+        # still running after waiting for timeout = give up on result
+        if self._geoloc_task.is_running:
+            return empty_result
+
+        # ok, so we did get a result
+        return self._geoloc_task.get_result()
```
poncovka, line 250: This will fail when you return `None` in the task.

pyanaconda/modules/timezone/timezone.py
```diff
@@ -216,3 +220,31 @@ def install_with_tasks(self):
...
+        """Start geolocation, if not already started"""
+        if self._geoloc_task is not None:
+            # geolocation has already has started
+            return
+
+        self._geoloc_task = GeolocationTask()
+        self._geoloc_task.start()
+
+    @property
+    def geolocation_result(self):
```
poncovka, line 234: I think this should be just a passive property that should return whatever is in `_geoloc_result`. The task should handle the updates.

pyanaconda/modules/timezone/timezone_interface.py
```diff
@@ -112,3 +112,14 @@ def SetTimeSources(self, sources: List[Structure]):
         self.implementation.set_time_sources(
             TimeSourceData.from_structure_list(sources)
         )
+
+    def StartGeolocation(self):
+        """Start geolocation, if not started yet"""
+        self.implementation.start_geolocation()
+
+    @property
+    def GeolocationResult(self) -> Structure:
```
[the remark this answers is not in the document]
poncovka: There is no signal for the new property.

poncovka (commented):
I have some suggestions, but I really like where it is going. It is so much cleaner already!

pyanaconda/core/configuration/anaconda.py
VladimirSlavik: It fails, but in a way that masks it. The string "0" would be propagated all the way to the provider lookup dict in `provider_id_to_class`, which would fall back to the Fedora provider. But this is never reached because geolocation is never started - `check_if_geolocation_should_be_used` recognized the `0` and `StartGeolocation` is never called.

poncovka: It seems dangerous to rely on that. Could the option be set to an empty string instead?

pyanaconda/modules/timezone/timezone_interface.py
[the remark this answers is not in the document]
VladimirSlavik: I think naming it the same as the data class isn't a good idea. The rest applies obviously.

pyanaconda/modules/timezone/geoloc.py
VladimirSlavik: Sounds good, one more argument for a helper property to decide if GeolocationData is valid.

pyanaconda/core/constants.py
VladimirSlavik: Probably yes. I just realized keeping it "in case it's needed again" is a bad idea.

If I change the waiting for result to using a signal, as suggested somewhere around, that actually removes the need for the other timeout, too.

VladimirSlavik:
Thank you, all good ideas.

I have a feeling this has a subtitle of "let's delete"... Despite adding new classes and code, it's net negative LOC already.

pyanaconda/modules/timezone/timezone_interface.py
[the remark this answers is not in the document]
poncovka: Hmm, ok. Let's use the current solution.

pyanaconda/modules/timezone/initialization.py
VladimirSlavik: I don't see how to realize a timeout in such case. That might not be a problem, though.

poncovka: You will need a synchronization object from the `threading` library. This object can block the task thread until a timeout or until it is released by the `PropertiesChanged` callback (the callback runs in the main thread). See: https://docs.python.org/3/library/threading.html

VladimirSlavik:
This is a state halfway through the feedback above.

During the changes, I realized we don't need any factories or polymorphism, all that differs is the URL. Let's rewrite this once more.

pyanaconda/core/configuration/anaconda.py
VladimirSlavik: I'd like to leave that after the general structure is worked out, but yes, let's have it.

VladimirSlavik:
Current TODOs left:
- add deprecation notices for `inst.geoloc` provider values
- remove prefix `provider_` from constants and conf
- research passive waiting for network

data/anaconda.conf
```diff
@@ -305,7 +305,6 @@ password_policies =
     user (quality 1, length 6, empty)
     luks (quality 1, length 6)
 
-
```
poncovka, line 308: There are always two empty lines between sections.

pyanaconda/modules/timezone/initialization.py
```diff
@@ -0,0 +1,136 @@
...
+
+        if not self._wait_for_network():
+            log.error("Geoloc: no network connection")
+            return GeolocationData()
+
+        result = self._locate(url)
+        log.info("Geoloc: lookup finished in %1.1f seconds", time.time() - start_time)
+
+        if result.is_empty():
+            log.info("Geoloc: no result")
```
poncovka, line 55: This will be clear from some of the error logs in the `_locate` method. I would just log the result data here.

tests/unit_tests/pyanaconda_tests/modules/timezone/test_timezone_tasks.py
```diff
@@ -0,0 +1,351 @@
...
+
+class TimezoneTasksTestCase(unittest.TestCase):
+    """Test the D-Bus Timezone (Timezone only) tasks."""
+
+    def test_timezone_task_success(self):
+        """Test the "full success" code paths in timezone D-Bus task."""
+        self._test_timezone_inputs(input_zone="Europe/Prague",
+                                   input_isutc=False,
+                                   make_adjtime=True,
+                                   make_zoneinfo=True,
```
poncovka, line 44: Could you also fix the indentation, please?

tests/unit_tests/pyanaconda_tests/modules/timezone/test_module_timezone.py
```diff
@@ -328,321 +320,3 @@ def test_deprecated_warnings(self):
 
         assert len(report.warning_messages) == 1
         assert report.warning_messages[0].message == warning
-
-
-class TimezoneTasksTestCase(unittest.TestCase):
```
poncovka, line 333: I would open a separate PR for this change. We can merge it right away.

pyanaconda/ui/gui/spokes/welcome.py
```diff
@@ -105,22 +105,25 @@ def _apply_selected_locale(self, locale):
...
-        if flags.flags.automatedInstall and not geoloc.geoloc.enabled:
-            return
-
-        if not self._tz_module:
+        if flags.flags.automatedInstall and geoloc_result.is_empty():
             return
 
         loc_timezones = localization.get_locale_timezones(self._l12_module.Language)
-        if geoloc.geoloc.result.timezone:
+        if geoloc_result.timezone:
```
poncovka, line 122: Btw. I think that `elif loc_timezones and not self._tz_module.Timezone:` is never used, because the timezone is always set. Since there are no bug reports for that, I recommend to remove this part of the code.

data/anaconda.conf
```diff
@@ -316,3 +315,14 @@ password_policies =
...
+[Timezone]
+# Which provider should be used to get geolocation data, if any?
+# This is used for automatic language and timezone detection.
+#
+# Valid values:
+#
+#   provider_fedora_geoip
+#   provider_hostip
+#
+geolocation_provider = provider_fedora_geoip
```
poncovka, line 328: In the TODO list, I am missing the use of a URL instead of an ID.

poncovka:
> add deprecation notices for `inst.geoloc` provider values

It is not approved yet. We will discussed it on Monday. Anyway, this can be done in a separate PR later.

VladimirSlavik:
Not worth re-reviewing yet, just resolved conflicts and fixed the smaller feedback.

VladimirSlavik:
Not worth re-reviewing yet, just resolved conflicts and fixed the smaller feedback.

pyanaconda/core/configuration/anaconda.py
VladimirSlavik: If I make the conf values use only URLs, then removing the prefix becomes meaningless, it will go away eventually anyway.

pyanaconda/modules/timezone/initialization.py
VladimirSlavik: After coming back to this after a few days, I realized - one benefit of using a thread (Task) should be to write the code synchronously. So if possible, I'd leave this out, or for later.

pyanaconda/modules/timezone/initialization.py
```diff
@@ -0,0 +1,118 @@
...
+        log.info("Geoloc: starting lookup using provider: %s", url)
+        start_time = time.time()
+
+        if not self._wait_for_network():
+            log.error("Geoloc: no network connection")
+            return GeolocationData()
+
+        result = self._locate(url)
+        log.info(
+            "Geoloc: lookup finished in %1.1f seconds, got result: %s",
```
M4rtinK, line 51: In the original implementation we did not log the territory code and timezone to the normal logs that get stored on the installed system, but logged it into a "sensitive" log that was only stored to the installation environment for debugging purposes and not transferred to the target system.

This was done not to store on the machine the country the user was in when the machine was installed, which could in some circumstances be quite a sensitive information.

Given that we don't have such a sensitive log mechanism in place and given how DBus modules work it would not be very easy do I suggest not logging the result of the geolocation or at least not logging the territory code.

M4rtinK (approved):
Looks good to me, just strongly suggest not logging the result of Geolocation to the normal log file. :)

pyanaconda/core/configuration/anaconda.py
```diff
@@ -403,7 +414,33 @@ def set_from_opts(self, opts):
...
+        if "geoloc" in opts and opts.geoloc and opts.geoloc != "0":
+            self.timezone._set_option(
+                "geolocation_provider",
+                convert_geoloc_provider_id_to_url(opts.geoloc)
+            )
+
         self.validate()
 
 
+def convert_geoloc_provider_id_to_url(provider_id):
```
poncovka, line 428: Could you make it private with `_`, please?

poncovka (approved):
It looks great!

pyanaconda/modules/timezone/initialization.py
poncovka: Do you have a link to a discussion about that? I would like to point out that we dropped some really sensitive data like the public IP address and the city and we are keeping only the timezone and the territory code (which is actually the country code). The timezone is set as it is, so it is logged by the Timezone module and it was never protected anyway. The territory code is transformed to a locale and eventually logged by the Localization module. I am not sure if we really need to protect the `GB` territory code if we later happily log the `en_GB.UTF-8` locale and the `Europe/London` timezone. Anyway, we are not able to access these logs unless users attach them to a bug report or something and they can always set up the privacy. @jkonecny12, what do you think?

jkonecny12: I wonder what is the benefit of having this logged? Are we using it somewhere? If not I would avoid logging it because it's just safer and I don't think it's any use for us.

To answer the question. I see the difference between `en_GB.UTF-8` or `Europe/London` and data from geolocation service. The point is that the first two could be set by a user to anything. It doesn't necessarily means that you are in that country. However, data from geolocation are given from outside and it's strong point for location of the user.

Honestly, I don't think we should log any of those values if we don't have a strong reason to do that. We shouldn't do that just because we can. I don't think that `en_GB.UTF-8` is that much helpful in case of application crash. If it will crash because of the value you should see that in the traceback or you can ask user. Feel free to give here arguments for logging these :).

pyanaconda/modules/timezone/initialization.py
```diff
@@ -0,0 +1,137 @@
...
+        end = start + timeout
+        while time.perf_counter() < end:
+            time.sleep(interval)
+            if network.Connected:
+                return True
+
+        return network.Connected
+
+    def _convert_provider_id_to_url(self, provider_id):
+        """Convert provider ID to URL of the corresponding service.
```
jkonecny12: You have a trailing period here but nowhere else.

jkonecny12 (commented):
Going through this quickly I found just a one nitpick.

pyanaconda/modules/timezone/initialization.py
VladimirSlavik: I agree with not logging the result data, instead we can only log if the result was useful or not. The values were useful to debug this during development, but they are not really needed.

However - can we define the threat? If it is _another party can learn that when installing, the computer was geolocated in \<country-sized area>_, then even logging that any valid geolocation response happened is bad. Above, we agreed that we can tell from the other logs what values were set later. That means anyone else could, too. Is it enough to make it harder to get the data, or do we have to go all the way to make it impossible?

pyanaconda/modules/timezone/timezone.py
```diff
@@ -216,3 +221,26 @@ def install_with_tasks(self):
                 ntp_servers=self.time_sources
             )
         ]
+
+    def _set_geolocation_result(self):
+        """Set geolocation result when the task finished"""
+        self._geoloc_result = self._geoloc_task.get_result()
+        self.geolocation_result_changed.emit()
```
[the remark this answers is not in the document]
VladimirSlavik: +1 to logging, +1 on setting `None`. But I disagree with even mentioning restarting. It is not needed now. Who will do that, for what reason, in what future? Let's not dive into more hypotheticals such as "using wifi and GPS to get the location".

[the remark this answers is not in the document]
poncovka: Ok, but the DBus API needs to make some sense. If I call `StartGeolocation` and there is no geolocation task running, it is reasonable to expect that the module should start one. That doesn't mean that our UI should react on this result as well. I could imagine trying to restart the geolocation after the user sets a new network configuration. Or just for debugging and testing. Anyway, the restarting will be implemented by just setting the attribute to None, so I don't think I am asking too much.

VladimirSlavik:
Only rebase.

pyanaconda/modules/common/structures/timezone.py
```diff
@@ -82,3 +82,50 @@ def options(self) -> List[Str]:
     @options.setter
     def options(self, value):
         self._options = value
+
+
+class GeolocationData(DBusData):
```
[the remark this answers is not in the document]
VladimirSlavik: I changed the logging to use the `is_empty` helper instead.

poncovka (approved):

PR #4075, opened 2022-04-29
Test release daily

VladimirSlavik:
This would let us proactively monitor the following:
- are we able to get a release tarball using the containerized make target
- are there any translation-related failures

PR #4077, opened 2022-05-02
Test release daily

VladimirSlavik:
This would let us proactively monitor the following:
* are we able to get a release tarball using the containerized make target
* are there any translation-related failures

Revival of #4075

### DIFF
--- a/.github/workflows/try-release-daily.yml
+++ b/.github/workflows/try-release-daily.yml
@@ -1,0 +1,70 @@
+name: Try building a release daily
+# Build a release tarball for all supported branches to verify the code is releasable.
+
+on:
+  schedule:
+    - cron: 0 3 * * *
+  # be able to start this action manually from a actions tab when needed
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+jobs:
+  daily-release-test:
+    runs-on: [self-hosted, kstest]
+    strategy:
+      fail-fast: false
+      matrix:
+        branch: ['master', 'rhel-8', 'rhel-9'] # depending on season, also: fNN-release
+
+    steps:
+      - name: Check out repo
+        uses: actions/checkout@v3
+        with:
+          ref: ${{ matrix.branch }}
+
+      - name: Build anaconda container (to make the release)
+        run: |
+          make -f Makefile.am anaconda-release-build
+
+      - name: Run the build in the container
+        run: |
+          make -f Makefile.am container-release | tee release.log
+
+      - name: Check logs for malformed translations
+        continue-on-error: true
+        run: |
+          errors=$(grep -E "Unable to compile [^:]+: .*" release.log || true)  # grep exits with 1 if no lines found
+          if [ -n "$errors" ]; then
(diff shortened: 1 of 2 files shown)


VladimirSlavik:
Test run - RHEL fails because that's not on internal infra with repos etc.: https://github.com/VladimirSlavik/anaconda/actions/runs/2257969040

VladimirSlavik:
F36 fails because it doesn't have the commit adding the logs, the same will happen on RHELs.

No idea about what's happening on master, yet.

VladimirSlavik:
Now it works: https://github.com/VladimirSlavik/anaconda/runs/6325786470?check_suite_focus=true

Unfortunately, it seems our worker VMs are missing bzip2, so that's blocking, otherwise I think this is fine.

.github/workflows/try-release-daily.yml
```diff
@@ -0,0 +1,69 @@
...
+            echo "No translation compilation errors found."
+          fi
+
+      - name: Check if translation files are present in tarball
+        continue-on-error: true
+        run: |
+          tarfiles=$(tar -tf anaconda-*.tar.bz2)
+          missing=0
+          required_languages="fr ja zh_CN es pt de ru ar"
+          # This list comes from langtable's list_common_languages, which in turn reportedly comes
```
M4rtinK, line 53: If we wanted to run a RHEL specific version of this, lnovy might have a translation priority list for RHEL.

M4rtinK (approved):

.github/workflows/try-release-daily.yml
```diff
@@ -0,0 +1,69 @@
+name: Try building a release daily
+# Build a release tarball for all supported branches to verify the code is releasable.
+
+on:
+  schedule:
+    - cron: 0 3 * * *
+  # be able to start this action manually from a actions tab when needed
+  workflow_dispatch:
```
jkonecny12, line 8: I'm thinking if this should not be part of the PRs instead of daily. The main reason is that the daily runs are easy to miss and also that we could discover our mistake before merge.

jkonecny12 (approved):
Looks good to me. I just wonder about the question below.

.github/workflows/try-release-daily.yml
VladimirSlavik: ljanda says: use languages in a bullet list here: https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html-single/8.6_release_notes/index#international_languages

.github/workflows/try-release-daily.yml
VladimirSlavik: ljanda says: use languages in a bullet list here: https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html-single/8.6_release_notes/index#international_languages

VladimirSlavik:
No longer blocked, bzip2 is now available on all kstest runners.

VladimirSlavik:
Update: changed list of RHEL languages according to discussion above; added ci status badge

PR #4078, opened 2022-05-02
Remove deprecated payload exceptions

poncovka:
Remove the following exceptions:

* `PayloadSetupError`
* `PayloadInstallError`
* `PayloadError`

We use DBus errors instead.


pyanaconda/payload/utils.py
```diff
@@ -122,8 +121,6 @@ def mount(device_path, mount_point, fstype, options):
     :param str mount_point: a mount point
     :param str fstype: a filesystem type
     :param str options: mount options
+    :raise: OSError on failure
     """
-    try:
-        return blivet.util.mount(device_path, mount_point, fstype=fstype, options=options)
-    except OSError as e:
-        raise PayloadSetupError(str(e)) from e
+    return blivet.util.mount(device_path, mount_point, fstype=fstype, options=options)
```
VladimirSlavik, line 126: I guess it's an open question if we really need this function at all.

VladimirSlavik (approved):
Looks good to me. Thank you!

PR #4080, opened 2022-05-03
Rename arguments of the `restart_thread` method

poncovka:
* Rename the `onlyOnChange` argument. Use `only_on_change` instead.
* Rename the `checkmount` argument. Use `try_media` instead.

### DIFF
--- a/pyanaconda/payload/manager.py
+++ b/pyanaconda/payload/manager.py
@@ -112,7 +112,7 @@ class PayloadManager(object):
             elif event_id <= self._thread_state:
                 func()
 
-    def restart_thread(self, payload, fallback=False, checkmount=True, onlyOnChange=False):
+    def restart_thread(self, payload, fallback=False, checkmount=True, only_on_change=False):
         """Start or restart the payload thread.
 
         This method starts a new thread to restart the payload thread, so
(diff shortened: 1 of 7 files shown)


poncovka:
Rebased.

pyanaconda/payload/manager.py
```diff
@@ -112,7 +112,7 @@ def add_listener(self, event_id, func):
             elif event_id <= self._thread_state:
                 func()
 
-    def restart_thread(self, payload, fallback=False, checkmount=True, onlyOnChange=False):
+    def restart_thread(self, payload, fallback=False, checkmount=True, on_change=False):
```
VladimirSlavik: I think the "only" is needed to make sense. Alternatively, you could flip the logic and call it something like  `skip_additions`.

poncovka: I am not sure what skipping additions means. The thread will be restarted only if it looks like the repositories have changed... `verify_change`? `force_restart`?

VladimirSlavik: Point taken. Anything against keeping the "only" part?

Also - if this blocks something for you, feel free to go ahead, the explanation in doctext is good enough, even if the name does not help me personally.

poncovka: That seems to be the best option. Changed.

VladimirSlavik (approved):
Thanks!

PR #4083, opened 2022-05-04
Move functions from pyanaconda.payload.image

poncovka:
Move some of the function from `pyanaconda.payload.image`.


VladimirSlavik (approved):
Looks good to me. Thank you!

PR #4085, opened 2022-05-05
Prepare release notes for Fedora 36

poncovka:
Merge all release notes for F36 into one file.


VladimirSlavik (commented):
Here's a blurb for #3584 as requested. I can't see anything suggesting the nosave boot options are deprecated or would be, but I'm not 100% sure.

```
Configuration file changes
--------------------------

Config options for saving anaconda's data to target system
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

Anaconda configuration file format now includes additional options to control
what is saved to the target system.

The options are::

    # Should we copy input kickstart to target system?
    can_copy_input_kickstart = True

    # Should we save kickstart equivalent to installation settings to the new system?
    can_save_output_kickstart = True

    # Should we save logs from the installation to the new system?
    can_save_installation_logs = True

The default values above cause no change in behavior, the new options are
only another way to configure the behavior.
```

VladimirSlavik (approved):
Thank you, didn't check the actual PR list, but this looks good to me on its own.

jkonecny12 (approved):

poncovka:
Included notes from https://github.com/rhinstaller/anaconda/pull/4085#pullrequestreview-966383096.

PR #4088, opened 2022-05-06
Create self contained Web UI boot iso

M4rtinK:
As we need to set quite a few Lorax options to build the self contained
Web UI boot iso, lets create a new container variant for it instead of
overloading the existing iso creator container.

The new container does a couple custom things:
- makes the rootfs 5 GB to make the payload tarball fit in
- use custom Lorax templates to set boot options
- use a repo with the payload RPM
- make sure payload RPM is installed

### DIFF
--- a/.github/workflows/build-boot-iso.yml
+++ b/.github/workflows/build-boot-iso.yml
@@ -53,7 +53,7 @@ jobs:
     env:
        STATUS_NAME: boot-iso
        CONTAINER_TAG: 'lorax'
-       ISO_BUILD_CONTAINER_NAME: 'quay.io/rhinstaller/anaconda-iso-creator'
+       ISO_BUILD_CONTAINER_NAME: 'quay.io/rhinstaller/anaconda-iso-creator-webui'
     steps:
       # we post statuses manually as this does not run from a pull_request event
       # https://developer.github.com/v3/repos/statuses/#create-a-status
@@ -73,10 +73,10 @@ jobs:
           ref: ${{ needs.pr-info.outputs.sha }}
           fetch-depth: 0
 
-      - name: Build anaconda-iso-creator container image
+      - name: Build anaconda-iso-creator-webui container image
         run: |
           # set static tag to avoid complications when looking what tag is used
-          sudo make -f ./Makefile.am anaconda-iso-creator-build CI_TAG=$CONTAINER_TAG
+          sudo make -f ./Makefile.am anaconda-iso-creator-build CI_TAG=$CONTAINER_TAG ISO_CREATOR_NAME=$ISO_BUILD_CONTAINER_NAME
 
       - name: Build anaconda-rpm container (for RPM build)
         run: |
--- a/dockerfile/anaconda-iso-creator-webui/Dockerfile
+++ b/dockerfile/anaconda-iso-creator-webui/Dockerfile
@@ -1,0 +1,49 @@
+# Dockerfile to build self-contained boot.iso with Anaconda Web UI.
+# To find out how to build this container please look on the ./tests/README.rst file.
+# This container has to be started as --privileged and with precreated loop devices otherwise
+# lorax won't work correctly.
+#
+# Execution example:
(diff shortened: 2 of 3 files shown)


VladimirSlavik (commented):
Nice job!

- You need `(#infra)` in the commit message, if you want to use the label. I think that's a good idea, it should succeed as-is.
- Given how many moving parts this has, you should probably demonstrate it works, same as other PRs for workflows.
- This needs customized lorax templates from a topic branch in a your fork, which is pretty fragile. It would be nice to document that also in the workflow, so that people don't think it works with stock lorax.
- Changing the existing file would destroy the original workflow, which we don't want. Workflows invoked by comment and "manual dispatch" (the button in actions tab) are taken from `master`, so keeping the changes on a topic branch and using them is not possible.

  Personally, I think this initial stage can look like this:
  - Duplicate the workflow yaml file, same as the container, **for now**. 
  - Change also the triggering comment.
  - Make it work and prove it :-)

  Actually, duplicating the workflow is not strictly required: If we add in the workflow file another job to build the webui variant (same as the `run` part), and tweak the conditions (`if: ...`) to run these two jobs such that eg. one takes just `/boot.iso` and the other one `/boot.iso webui`, that's straightforward to make. However, that would add another layer of complexity, and make the file very long, and invite more bikeshedding anyway. So, personally I'd keep that for another PR.

dockerfile/anaconda-iso-creator-webui/Dockerfile
```diff
@@ -0,0 +1,49 @@
+# Dockerfile to build self-contained boot.iso with Anaconda Web UI.
```
jkonecny12, line 1: This new Dockerfile is a duplication of the existing one. We don't need that. Instead we should look on the script inside which you want to change.

There are two possibilities:
- adjust the script to be more universal
- add a new script which can be called instead of the original one by `--entrypoint`

.github/workflows/build-boot-iso.yml
```diff
@@ -73,10 +73,10 @@ jobs:
           ref: ${{ needs.pr-info.outputs.sha }}
           fetch-depth: 0
 
-      - name: Build anaconda-iso-creator container image
+      - name: Build anaconda-iso-creator-webui container image
         run: |
           # set static tag to avoid complications when looking what tag is used
-          sudo make -f ./Makefile.am anaconda-iso-creator-build CI_TAG=$CONTAINER_TAG
+          sudo make -f ./Makefile.am anaconda-iso-creator-build CI_TAG=$CONTAINER_TAG ISO_CREATOR_NAME=$ISO_BUILD_CONTAINER_NAME
```
jkonecny12, line 79: This will not work, you need to change the source Dockerfile not name.

However, first try to just use the old one with the an adjustment.

dockerfile/anaconda-iso-creator-webui/lorax-build
```diff
@@ -0,0 +1,45 @@
...
+INPUT_RPMS=/anaconda-rpms/
+REPO_DIR=/tmp/anaconda-rpms/
+
+# create repo from provided Anaconda RPMs
+mkdir -p $REPO_DIR
+cp -a $INPUT_RPMS/* $REPO_DIR || echo "RPM files can't be copied!"  # We could just do the build with official repositories only
+createrepo_c $REPO_DIR
+
+# clone Lorax fork with modified templates, which manipulate BIOS & EFI boot options to boot into the Web UI
+git clone --branch master-webui_boot_options https://github.com/M4rtinK/lorax/ /tmp/lorax
```
jkonecny12, line 27: Theoretically we can solve these changes by an environment variable. If there will be invocation with `-e "WEBUI=1"` as the podman call parameter then we don't need any of this duplication.

jkonecny12 (changes requested):
Hi, thanks for these changes. It needs a bit of polishing but we are on a good way here.

Please also add `(#infra)` as the postfix of the commit message.

PR #4089, opened 2022-05-09
Move the default source type on DBus

poncovka:
* Add the `DefaultSourceType` DBus property to the payload modules.
* Remove the `default_source_type` property from the payload classes.


M4rtinK (approved):
LGTM. :)

PR #4091, opened 2022-05-09
Simplify the implementation for the DBus interface for Users module

poncovka:
Call `DBusData.to_structure_list` to simplify the code.

### DIFF
--- a/pyanaconda/modules/users/users_interface.py
+++ b/pyanaconda/modules/users/users_interface.py
@@ -132,14 +132,7 @@ class UsersInterface(KickstartModuleInterface):
 
         :return: a list of user describing DBus Structures
         """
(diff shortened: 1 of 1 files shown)


poncovka:
Codecov is just a little confused.

VladimirSlavik (approved):
Looks good to me. Thank you!

VladimirSlavik:
From human perspective you replaced some special code paths with already existing universal code paths, so it's a net reduction, which is good.

Unfortunately Codecov is exactly right, but too simplistic about counting = too dumb to know what matters. From the mechanistic percent-counter perspective, replacing tested code paths with other also tested code paths decreases the amount of tested code paths, while the amount of untested code paths remains the same. So technically less % coverage is correct.

I love how this is all proven by no change to tests needed.

PR #4097, opened 2022-05-10
Check if the mount point exists before calling statvfs (#1826760)

poncovka:

(cherry-picked from a commit 5f43387)

Resolves: rhbz#1826760


VladimirSlavik (approved):
Looks good to me. Thank you!

M4rtinK (approved):
Looks good to me, thanks! :)

M4rtinK (approved):

PR #4099, opened 2022-05-12
Document how to install dependencies without any interaction

poncovka:
Run `install_dependencies.sh -y`.

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -21,9 +21,10 @@ dealing with containers, just skip this part and continue on the next one::
     toolbox create
     toolbox enter
 
-To prepare the environment in the container or on your system just run these commands::
+To prepare the environment in the container or on your system run the commands below. Install
+dependencies without the ``-y`` option to preview the transaction::
 
-    sudo ./scripts/testing/install_dependencies.sh
+    sudo ./scripts/testing/install_dependencies.sh -y
     ./autogen.sh && ./configure
 
 How to Contribute to the Anaconda Installer (the short version)


M4rtinK (approved):
LGTMAW. :)

VladimirSlavik (approved):
Thanks!

jkonecny12:
Yes, that is it. The `toolbox run` is not interactive, you can't control it. But if you follow the documentation here it will stay on the place a wait for `y/N` which could be controlled.

Strictly speaking, the change is not correct because if you follow the documentation you won't get the preview but `y/N` question as with usual `dnf` call. I guess that sentence should change to something like.

```
Use `-y` if you need non-interactive execution. Helpful in case you are using `toolbox run`.
```

VladimirSlavik:
I remember not getting the y/n prompt, only the question was printed and the script quit as if I declined the changes. I got it to work by reading the script and realizing I'm supposed to do `-y` if the changes look sane. So from my PoV this is more correct, which is why I acked it.

I never used toolbox, so if it behaves differently for you, that's a thing. and the docs will be good either for one or the other group? The different behavior could have been also without `sudo` though. So I'm not sure.,

Why can't anything ever be easy :-/

jkonecny12:
I really wonder why this is happening to you. The `toolbx` should work as local terminal without any change (almost - it has `sudo` without password). Running the `sudo ./install_dependencies.py` should behave the same way in `toolbx` as in the system.

poncovka:
I guess this line will never install anything without `-y`, because stdin is already busy:
`rpmspec -q --buildrequires $TEMP | sed 's/>=.*$//' | xargs -d '\n' dnf install $@`

VladimirSlavik:
I'd like to see the "preview functionality" preserved.

I think this can be worked around by using a variable to hold the list instead of directly piping to xargs:
```
deps=$(rpmspec -q --buildrequires $TEMP | sed 's/>=.*$//;s/\n/ /')
dnf install $deps $@
```

That would also drop the xargs dependency.

poncovka:
I guess this PR can be closed in that case.

PR #4100, opened 2022-05-12
Web UI: Reset the bootloader drive before we schedule partitions

poncovka:
The bootloader drive is automatically set during the partitioning, so
make sure we always reset the previous value before we run another one,
so it can be automatically set again based on the current disk selection.
Otherwise, the partitioning can fail with an error.


VladimirSlavik (approved):
Looks good to me. Thank you!

M4rtinK (approved):
Looks good to me, thanks! :-)

PR #4103, opened 2022-05-13
Open logical volume devices

yueyuankun:
So that os-prober can recognize the system on other hard disks.
eg.
When /dev/sda has a complete system, the new system is installed on the /dev/sdb, and the grub.cfg of the /dev/sdb disk has the start menu of /dev/sda.

### DIFF
--- a/pyanaconda/modules/storage/devicetree/model.py
+++ b/pyanaconda/modules/storage/devicetree/model.py
@@ -268,7 +268,7 @@ class InstallerStorage(Blivet):
         self.bootloader.reset()
 
         self.roots = []
-        self.roots = find_existing_installations(self.devicetree)
+        self.roots = find_existing_installations(self.devicetree, False)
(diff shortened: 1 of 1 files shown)


jkonecny12:
@poncovka what do you think about this change?

VladimirSlavik:
Just to note what I found this far:
- The parameter is not used anywhere in current code, always defaults to True.
- It's pretty hard to get any historical comments and commit messages for this because the code moved to blivet and back.
- All the tests succeeded, so perhaps it can actually work?

poncovka (changes requested):
Anaconda very early tears down all devices and keeps them that way until the actual partitioning. We do that to prevent bugs during configuration and installation. There is a plan to implement a smarter solution, but it will be risky and require excessive testing. Unfortunately, I don't think we can merge this at this moment.

PR #4107, opened 2022-05-16
webui: Add a Quit confirmation dialog

M4rtinK:
And adjust the unit tests to handle the confirm dialog.


VladimirSlavik (approved):
Looks good to me. Thank you!

Test results don't look too good though...

lclay2:
Is there a way I can check this too?

VladimirSlavik:
@lclay2, yes, if @M4rtinK attaches a screenshot.

If it helps any, I commented only on the code. The "looks" part probably sounded as if I saw the result - I didn't.

ui/webui/test/helpers/installer.py
```diff
@@ -93,3 +93,4 @@ def check_prerelease_info(self, is_expected=None):
 
     def quit(self):
         self.browser.click("#installation-quit-btn")
+        self.browser.click("#installation-quit-confirm-btn")
```
VladimirSlavik, line 96: Does this work locally? My instinct would be to wait for the button to become visible before clicking it.

M4rtinK: Yep, it does click the button, shutting down the VM. But good point, better check the dialog is visible first & only then click.

PR #4108, opened 2022-05-16
tests: Get list of supported LVM RAID levels from blivet

vojtechtrefny:
We are adding LVM RAID raid0 support to blivet so the harcoded
list of supported levels won't work anymore.

See https://github.com/storaged-project/blivet/pull/1047 for
details.

Note: Anaconda doesn't support LVM RAID so the Blivet change shouldn't affect it.

### DIFF
--- a/tests/unit_tests/pyanaconda_tests/modules/storage/partitioning/test_module_scheduler.py
+++ b/tests/unit_tests/pyanaconda_tests/modules/storage/partitioning/test_module_scheduler.py
@@ -168,8 +168,8 @@ class DeviceTreeSchedulerTestCase(unittest.TestCase):
 
     def test_get_supported_raid_levels(self):
         """Test GetSupportedRaidLevels."""
-        assert self.interface.GetSupportedRaidLevels(DEVICE_TYPE_LVM) == \
-            ['linear', 'raid1', 'raid10', 'raid4', 'raid5', 'raid6', 'striped']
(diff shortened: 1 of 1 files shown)


tests/unit_tests/pyanaconda_tests/modules/storage/partitioning/test_module_scheduler.py
```diff
@@ -169,7 +170,7 @@ def test_get_default_file_system(self):
     def test_get_supported_raid_levels(self):
         """Test GetSupportedRaidLevels."""
         assert self.interface.GetSupportedRaidLevels(DEVICE_TYPE_LVM) == \
-            ['linear', 'raid1', 'raid10', 'raid4', 'raid5', 'raid6', 'striped']
+            sorted([level.name for level in lvm.raid_levels])
```
poncovka, line 173: We need to verify that we get the expected strings to make sure that the DBus call doesn't crash and doesn't return nonsense. This test doesn't really verify any of that. Maybe we could check if we get a subset of possible strings?

vojtechtrefny: I don't understand -- it now checks that the DBus call returns the same result the Python API does. So blivet works (lets assume it does) and there is a bug in the Anaconda DBus logic, this will still find it.

But I guess I can simply change this to check for supported levels for MD RAID. We aren't changing those.

VladimirSlavik (approved):
Looks good to me. Thank you!

PR #4109, opened 2022-05-17
webui: Show installation status text on progress screen

M4rtinK:
Show the current status messages under the main text on the progress
screen. Also drop the old expandable log viewer, which will be replaced
by a global modal log viewer later.

### DIFF
--- a/ui/webui/src/components/installation/InstallationProgress.jsx
+++ b/ui/webui/src/components/installation/InstallationProgress.jsx
@@ -20,12 +20,11 @@ import React from "react";
 
 import {
     Button,
-    ExpandableSection,
     Flex,
+    TextContent,
     ProgressStepper, ProgressStep,
 } from "@patternfly/react-core";
 import { CheckCircleIcon, ExclamationCircleIcon } from "@patternfly/react-icons";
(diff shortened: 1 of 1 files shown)


M4rtinK:
Snímek obrazovky 2022-05-17 v 18 45 02

Looks like this - note the status message under the "Installation failed" label.

lclay2:
Thanks for the screenshot @M4rtinK !
Question: We have this jira for an unexpected error message: https://issues.redhat.com/browse/INSTALLER-2988. I believe it's for unrecoverable system errors. Is the toast that you're showing here for a non fatal system error? Is it directly related to the step error that you're presenting?

poncovka (approved):

M4rtinK:
> Thanks for the screenshot @M4rtinK !
No problem. :)

> Question: We have a jira for an unexpected error message. I believe it's for unrecoverable system errors. Is the toast that you're showing here for a non fatal system error? Is it directly related to the step error that you're presenting?

So this is a fatal error & directly related to the "installation failed" text to be displayed & the red exclamation. So I think this would be covered by the unexpected errors Jira eventually and the current implementation is essentially a working placeholder till then.

PR #4113, opened 2022-05-18
Inform users about the required space and the partitioning method

poncovka:
The Installation destination screen should inform users about
the required space and the partitioning method.

The required device size shouldn't be rounded down to zero,
so always round it up to the nearest MiB.


poncovka:
image

poncovka:
Updated pixel tests.

PR #4115, opened 2022-05-18
Web UI: Fix the style of paragraphs

poncovka:
* Move the content of every wizard page to the TextContent component.
* Move every main paragraph to the Text component to fix its style.

### DIFF
--- a/ui/webui/src/components/AnacondaWizard.jsx
+++ b/ui/webui/src/components/AnacondaWizard.jsx
@@ -22,10 +22,14 @@ import {
     ActionList,
     Alert,
     Button,
-    Title,
+    Modal,
+    ModalVariant,
     Stack,
-    Wizard, WizardFooter, WizardContextConsumer,
-    Modal, ModalVariant,
+    TextContent,
+    Title,
+    Wizard,
+    WizardFooter,
+    WizardContextConsumer,
 } from "@patternfly/react-core";
 
 import { AddressContext } from "./Common.jsx";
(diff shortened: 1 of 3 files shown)


M4rtinK (approved):
Looks good to me, thanks! :)

ui/webui/src/components/AnacondaWizard.jsx
```diff
@@ -55,14 +59,16 @@ const getSteps = ({
...
+                <TextContent>
+                    {stepNotification &&
+                     (stepNotification.step === currentStepId) &&
+                     <Alert
+                       isInline
+                       title={stepNotification.message}
+                       variant="danger"
+                     />}
+                    {children}
+                </TextContent>
```
VladimirSlavik, line 71: I think you should have instead wrapped only the `Alert` in `TextContent`, but it probably works this way too.

VladimirSlavik, line 71: I believe you should have instead wrapped only the `Alert` in `TextContent` and kept the logic outside, but it probably works this way too. I'm concerned mainly because of any side effects of the empty `TextContent` that is now always present, but that's because I don't know what that will really do.

To clarify:
- old was `{logic && <Alert>...`
- you made it `<TextContent>{logic && <Alert>...`
- I suggest `{logic && <TextContent><Alert>...`

ui/webui/src/components/localization/InstallationLanguage.jsx
```diff
@@ -195,14 +201,20 @@ LanguageSelector.contextType = AddressContext;
...
-                />
-            </FormGroup>
-        </Form>
+        <>
+            <Text component={TextVariants.p}>{_(
+                "Select the language you would like to use. This language will also be selected" +
+                " for your installed system.")}
+            </Text>
+            <Form>
+                <FormGroup>
```
VladimirSlavik, line 210: Can you set the `<Text>` directly to the `<FormGroup label=...` instead, or does that break the styles?

VladimirSlavik, line 210: Can you set the `<Text>` directly to the `<FormGroup label=...` instead?

VladimirSlavik (approved):
Thank you, code looks good to me.

ui/webui/src/components/localization/InstallationLanguage.jsx
poncovka: That's a completely different style.

VladimirSlavik: Makes sense, thanks.

ui/webui/src/components/AnacondaWizard.jsx
poncovka: Actually, the title should be there as well. It is not about the alert, but the whole content of the wizard step (the right part). The `TextContent` component is just a simple `div` that will fix the style of the text on that page.

VladimirSlavik (approved):
Thanks!

PR #4119, opened 2022-05-19
webui: Disable "Next" button if no disks are selected

M4rtinK:
Disable the "Next" button in the Wizard if no disks are selected
in the installation destination step.

This is using the isFormValid property, which is already used by
the language selection step.

But in this case as the installation destination step is not the first one,
we need to ale take care to reset form validation status back to default
when the step is exited.

The idea is that we always set it to default (true) in the functions
that govern step switching & each step can then do the validation
in their refresh handler (usually the useEffect() call).

As this also changes the UI workflow, adjust the unit test accordingly.

### DIFF
--- a/ui/webui/src/components/AnacondaWizard.jsx
+++ b/ui/webui/src/components/AnacondaWizard.jsx
@@ -23,6 +23,7 @@ import {
     Modal,
     ModalVariant,
     Stack,
+    Tooltip,
     Wizard,
     WizardFooter,
     WizardContextConsumer,
@@ -87,6 +88,9 @@ export const AnacondaWizard = ({ onAddErrorNotification, title }) => {
 
     const startAtStep = steps.findIndex(step => step.id === path[0]) + 1;
     const goToStep = (newStep) => {
+        // first reset validation state to default
+        setIsFormValid(true);
+
         cockpit.location.go([newStep.id]);
     };
 
@@ -95,6 +99,7 @@ export const AnacondaWizard = ({ onAddErrorNotification, title }) => {
           id="installation-wizard"
           footer={<Footer
             isFormValid={isFormValid}
+            setIsFormValid={setIsFormValid}
             setStepNotification={setStepNotification}
             isInProgress={isInProgress}
             setIsInProgress={setIsInProgress}
@@ -111,11 +116,14 @@ export const AnacondaWizard = ({ onAddErrorNotification, title }) => {
     );
 };
 
-const Footer = ({ isFormValid, setStepNotification, isInProgress, setIsInProgress }) => {
+const Footer = ({ isFormValid, setIsFormValid, setStepNotification, isInProgress, setIsInProgress }) => {
     const [nextWaitsConfirmation, setNextWaitsConfirmation] = useState(false);
     const [quitWaitsConfirmation, setQuitWaitsConfirmation] = useState(false);
 
     const goToStep = (activeStep, onNext) => {
+        // first reset validation state to default
+        setIsFormValid(true);
+
         if (activeStep.id === "installation-destination") {
             setIsInProgress(true);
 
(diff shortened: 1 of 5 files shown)


M4rtinK:
Snímek obrazovky 2022-05-19 v 14 25 37

Snímek obrazovky 2022-05-19 v 14 25 30

VladimirSlavik (approved):
Looks good to me. Thank you!

jkonecny12:
You are missing the tooltip info explaining why is the button disabled. I guess it should be part of this PR?

poncovka:
@M4rtinK, can you merge this PR, please? It will conflict with other changes.

M4rtinK:
Updated and hopefully final version with the tooltip. This is how it looks like now:
Snímek obrazovky 2022-06-08 v 14 30 50
When no disks are selected, the next button is not active & the tooltip is automatically shown:
Snímek obrazovky 2022-06-08 v 14 43 33
This is actually necessary, as PatternFly `Button` with `isDisabled` set apparently no longer gets any mouse events, to the usual on-mouse-over tooltip trigger can't be used.

ui/webui/src/components/AnacondaWizard.jsx
```diff
@@ -210,6 +219,20 @@ const Footer = ({ isFormValid, setStepNotification }) => {
...
+                                      <div>
+                                          {_("To continue, select the devices(s) to install to.")}
+                                      </div>
+                                  }
+                                  // Only show the tooltip on installation destination spoke that is not valid (no disks selected).
+                                  // NOTE: As PatternFly Button with isDisabled set apprently does not get any mouse events anymore,
+                                  //       we need to manually trigger the tooltip.
+                                  reference={() => document.getElementById("installation-next-btn")}
+                                  trigger="manual"
+                                  isVisible={activeStep.id === "installation-destination" && !isFormValid}
```
poncovka, line 234: Don't you want to actually skip the `Tooltip` component if this is a different step?

poncovka, line 234: Don't you want to actually skip the `Tooltip` component if this is a different step?

ui/webui/src/components/AnacondaWizard.jsx
```diff
@@ -144,12 +148,15 @@ export const AnacondaWizard = ({ onAddErrorNotification, title }) => {
...
 
-const Footer = ({ isFormValid, setStepNotification }) => {
+const Footer = ({ isFormValid, setIsFormValid, setStepNotification }) => {
     const [isInProgress, setIsInProgress] = useState(false);
     const [nextWaitsConfirmation, setNextWaitsConfirmation] = useState(false);
     const [quitWaitsConfirmation, setQuitWaitsConfirmation] = useState(false);
 
     const goToStep = (activeStep, onNext) => {
+        // first reset validation state to default
+        setIsFormValid(true);
```
poncovka, line 158: Doesn't this trigger rendering of the active step? I had some issues with missing disks on the Review screen.

poncovka, line 158: Doesn't this trigger rendering of the active step? I had some issues with missing disks on the Review screen.

ui/webui/src/components/storage/InstallationDestination.jsx
```diff
@@ -133,15 +133,17 @@ const LocalStandardDisks = ({ idPrefix, onAddErrorNotification }) => {
                 }, console.error);
     }, [refreshCnt]);
 
+    const totalDisksCnt = Object.keys(disks).length;
+    const selectedDisksCnt = Object.keys(disks).filter(disk => !!disks[disk]).length;
```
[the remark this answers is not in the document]
M4rtinK: It seems to work fine & its like this already:
```

    // When the selected disks change in the UI, update in the backend as well
    useEffect(() => {
        const selected = Object.keys(disks).filter(disk => disks[disk]);

        setSelectedDisks({ drives: selected }).catch(onAddErrorNotification);
    }, [disks, onAddErrorNotification]);

    const totalDisksCnt = Object.keys(disks).length;
    const selectedDisksCnt = Object.keys(disks).filter(disk => !!disks[disk]).length;

    if (totalDisksCnt === 0) {
        return <EmptyStatePanel loading />;
    }
```

Note the `useEffect` block ending right before the variable being set.

PR #4121, opened 2022-05-19
webui: Wrap Local disks heading row items in FlexItem

VladimirSlavik:
This fixes the bunching-up of the "Local standard disks" heading and the "A (of B) disks selected" label.

I'm not sure when it appeared, but it's a regression so let's fix it.

Before:
Screenshot from 2022-05-19 16-45-15

After:
Screenshot from 2022-05-19 16-38-40


poncovka:
I am changing the whole header in https://github.com/rhinstaller/anaconda/pull/4126. Please, don't touch this. Why do we have so many conflicts?

VladimirSlavik:
Ah, sorry, didn't see that change. I see that fixes the spacing too, so that's fine.

PR #4122, opened 2022-05-19
Web UI: Fix strings

poncovka:
* Fix the capital letter in `Installer`.
* Replace the `Storage configuration` title.

### DIFF
--- a/ui/webui/src/components/AnacondaWizard.jsx
+++ b/ui/webui/src/components/AnacondaWizard.jsx
@@ -101,12 +101,12 @@ export const AnacondaWizard = ({ onAddErrorNotification, title }) => {
             component: InstallationLanguage,
             id: "installation-language",
             label: _("Welcome"),
-            title: _("Welcome to Anaconda Installer")
+            title: _("Welcome to the Anaconda installer")
         },
         {
             component: InstallationDestination,
             id: "installation-destination",
-            label: _("Storage configuration"),
+            label: _("Installation destination"),
         },
         {
             component: ReviewConfiguration,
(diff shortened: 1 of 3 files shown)


ui/webui/src/components/storage/InstallationDestination.jsx
```diff
@@ -158,7 +158,7 @@ const LocalStandardDisks = ({ idPrefix, onAddErrorNotification }) => {
                   id="installation-destination-table-label"
                 >
                     {cockpit.format(
-                        cockpit.ngettext("$0 (of $1) disk selected", "$0 (of $1) disks selected", selectedDisksCnt),
+                        "$0 (of $1) disks selected",
```
VladimirSlavik, line 161: This is incorrect in most languages. See https://www.gnu.org/software/gettext/manual/html_node/Plural-forms.html after "The following rules are known at this point.".

VladimirSlavik (changes requested):
Lower i is fine, the plural form is an English-only change.

ui/webui/src/components/storage/InstallationDestination.jsx
VladimirSlavik: FWIW, this is a design problem, not a PR problem, really.

poncovka:
Updated.

VladimirSlavik (approved):
Thank you, LGTM.

You will have to change the names also in the tests, but that's expected.

jkonecny12 (approved):
Looks good to me.

lclay2:
@poncovka , I replied to your question on the slide deck.

PR #4101, opened 2022-05-12
Don't use Cockpit style overrides

poncovka:
Cockpit modify the PatternFly style, so let's drop it.

PR #4123, opened 2022-05-19
webui: Make app/page span the whole viewport

VladimirSlavik:
Fixes the most visible regression after #4101 (a61ca8dd3037fd9563b2b227aaf18a58628f2fcc).

### DIFF
--- a/ui/webui/src/components/app.scss
+++ b/ui/webui/src/components/app.scss
@@ -1,0 +1,10 @@
+// Copied from cockpit/pkg/lib/page.scss instead of including it in its entirety:
+// Let PF4 handle the scrolling through page component otherwise we might get double scrollbar
+html:not(.index-page) body {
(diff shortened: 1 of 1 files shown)


poncovka (approved):
Thanks!

poncovka:
Anyway, are we sure we want the window to be expanded? It could have fixed width and height. Didn't @M4rtinK plan to investigate something like that?

VladimirSlavik:
When would that investigation happen? The language selector on the first page is practically unusable. I think that would be good to fix asap. In other words, this could be a better starting point for that investigation.

jkonecny12 (approved):

VladimirSlavik:
Oh, no problem, I'll wait for the pixel test updates anyway. I thought you'd reply something like "that's happening next week" so this would really not be needed :)

PR #4128, opened 2022-05-23
Add missing period when concatenating password error strings (#2075419)

jkonecny12:
Without this change we have in the bottom line messages like:

"The password fails the dictionary check - it is too simplistic/systematic You will have to press Done twice to confirm it."

The missing period between sentences could be a big issue for understanding the sentence in languages like Chinese.

The fix here is not great one because we are not owners of both strings.

The first sentence comes from the libreport library the second one from us. Here we have two possible solutions. One would be to add translatable concatenation (what I'm doing here) and second is to add the period into the `...press Done twice...` message. I don't like to duplicate the second part so I rather add the new "{}. {}" for translations.

*Resolves: rhbz#2075419*

### DIFF
--- a/pyanaconda/core/constants.py
+++ b/pyanaconda/core/constants.py
@@ -236,6 +236,8 @@ SECRET_ASCII = {
 }
 PASSWORD_DONE_TWICE = N_("You will have to press <b>Done</b> twice to confirm it.")
 PASSWORD_SET = N_("Password set.")
+# TRANSLATORS: Password error message from libreport library needs to be joined with "You will have to press <b>Done</b> twice to confirm it." add a missing '.'
+PASSWORD_ERROR_CONCATENATION = N_("{}. {}")
 
 
 class SecretStatus(Enum):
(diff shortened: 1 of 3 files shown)


jkonecny12:
Before:
Screenshot_jk_test_2022-05-23_12:41:09

After:
Screenshot_jk_test2_2022-05-23_12:40:58

VladimirSlavik (changes requested):
Sorry, this won't work, or at least not as you perhaps intend it to work. If you want to use context, then the below is how you would do it.

The reason it "works" now is that the code defaults to passing through the string if not found, but would be never translated.

VladimirSlavik:
Aside from making it work, I think this is a good enough solution.

I also missed the obvious - perhaps a context is not needed after all. Context is useful when you have more "variants" of the same string literal, which is not the case here. A translator comment might be better instead. Then you have only one string to handle.

Having the string directly in the `show_warning_message` call might be better, it will be automatically detected as a format string:

```
self.show_warning_message(
  # TRANSLATORS: Concatenation of two error messages, first is lacking period (.) at end of sentence.
  _("{}. {}").format(
    error_message,
    _(constants.PASSWORD_DONE_TWICE)
  )
)
```

jkonecny12:
Adding version with the translator comment. However, still waiting for Ludek confirmation.

VladimirSlavik (approved):
Looks good to me. Thank you!

rvykydal (approved):

PR #4129, opened 2022-05-23
Port restorecon user dirs to rhel 8

VladimirSlavik:
Running restorecon needs chroot to sysroot so as to have correct absolute 
paths. At the point when users are handled, payload has been already
installed, so the SELinux policy is already present.

This is achieved by using a new restorecon utility function instead of running
restorecon directly with execWithRedirect. The helper can set root correctly.
It is also backported 1:1 from upstream and thus retains the "array of str"
parameter from upstream, even if it is not used with more than one path on
this branch.

Resolves: [rhbz#2069305](https://bugzilla.redhat.com/show_bug.cgi?id=2069305)

(cherry picked from commit a7a0713407baade6871486f7daf37863f4515c20)
(cherry picked from commit b511e61d7e5d46e79f133f767ded6f399bab9136)

(cherry picked from commit 65025ad94bb5c11005eda0fba3419563acf8c461)
(cherry picked from commit 3197d194a591fc21c2894b2da1137c425d2a7347)

Ported from #3993

### DIFF
--- a/pyanaconda/core/users.py
+++ b/pyanaconda/core/users.py
@@ -423,7 +423,7 @@ def create_user(username, password=False, is_crypted=False, lock=False,
             util.chown_dir_tree(root + homedir,
                                 int(pwent[2]), int(pwent[3]),
                                 orig_uid, orig_gid)
-            util.execWithRedirect("restorecon", ["-r", root + homedir])
+            util.restorecon([homedir], root=root)
         except OSError as e:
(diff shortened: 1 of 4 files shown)


rvykydal (approved):
Looks good to me.

M4rtinK (approved):
Looks good to me as well. :)

jkonecny12 (approved):

PR #4130, opened 2022-05-23
Port password eye fix for RHEL 8.7

VladimirSlavik:
Anaconda supports showing a little "eye" icon in the password/passphrase entry field that can be used to make the entered text visible in plaintext during entry.

For this to work correctly entry text visibility state needs to be set correctly. This has been broken by some of the password/passphrase entry related refactorings in the recent past, resulting in the "eye" icon not being visible.

So put the initialization back for the root password and LUKS passphrase entries and also extend the docstring in the user configuration spoke to note how password visibility influences the toggle icon.

Resolves: rhbz#1885879

(cherry picked from commit 8dbf7e3c7d606d0912d697fc11338698388276c3)

Port of #3119 - thanks @M4rtinK!

### DIFF
--- a/pyanaconda/ui/gui/spokes/lib/passphrase.py
+++ b/pyanaconda/ui/gui/spokes/lib/passphrase.py
@@ -90,6 +90,12 @@ class PassphraseDialog(GUIObject):
         self._checker.add_check(self._ascii_check)
         self._checker.add_check(self._empty_check)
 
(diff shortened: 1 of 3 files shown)


M4rtinK (approved):
Looks good to me, thanks! :)

jkonecny12 (approved):

PR #4137, opened 2022-05-24
Add missing period when concatenating password error strings (#2075419)

jkonecny12:
Without this change we have in the bottom line messages like:

"The password fails the dictionary check - it is too simplistic/systematic You will have to press Done twice to confirm it."

The missing period between sentences could be a big issue for understanding the sentence in languages like Chinese.

The fix here is not great one because we are not owners of both strings.

The first sentence comes from the libreport library the second one from us. Here we have two possible solutions. One would be to add translatable concatenation (what I'm doing here) and second is to add the period into the `...press Done twice...` message. I don't like to duplicate the second part so I rather add the new "{}. {}" for translations.

Resolves: rhbz#2075419

Backport of https://github.com/rhinstaller/anaconda/pull/4128.

### DIFF
--- a/pyanaconda/core/constants.py
+++ b/pyanaconda/core/constants.py
@@ -213,6 +213,8 @@ SECRET_ASCII = {
 }
 PASSWORD_DONE_TWICE = N_("You will have to press <b>Done</b> twice to confirm it.")
 PASSWORD_SET = N_("Password set.")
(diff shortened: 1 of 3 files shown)


poncovka (approved):
This is just awful and sad, but whatever.

jkonecny12:
Ideally it should be solved in the libreport library but that could break a lot of other applications which already counting on this.

Another option would be to add a new "duplicate" string for the press the done button twice with period on the begging. But that seemed even worse solution.

rvykydal (approved):

PR #4138, opened 2022-05-24
Use the IsRootAccountLocked property

poncovka:
Set the right property of the Users module.


M4rtinK (approved):
Looks good to me as well. :)

VladimirSlavik (approved):
Thanks!

PR #4139, opened 2022-05-24
Revert "Run kickstart tests on PR via permian (#infra)"

rvykydal:
This reverts commit 4b9f132126e5936b8fd41e775ebea7b15495000a.


jkonecny12 (approved):
Looks good to me. Although, I'm thinking if we shouldn't instead add a cleanup step at the end which would remove the sudo created files.

VladimirSlavik (approved):

poncovka (approved):

PR #4141, opened 2022-05-24
Install eslint during container build

VladimirSlavik:
Previously, it was installed on first test run, which with containers means always. Additionally, this installs only eslint, not all of the packages. The combined savings should be around a minute per test run, at a cost of ca. +17 MB container size.

This makes the test slightly less correct when run outside the container: Previously the version to install was taken from the config file. Now, eslint is automatically installed after invoking it via npx, but it cannot guess where the version file is, so it installs the latest. However, running the test without container is unusual, so this should be good enough.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -170,6 +170,7 @@ release-and-tag:
 anaconda-ci-build:
 	TEMP=$$(mktemp -t -d anaconda-ci-build.XXXX) && \
 	cp $(srcdir)/anaconda.spec.in $(CI_DOCKERFILE)/* $$TEMP/ && \
+	cp $(srcdir)/ui/webui/package.json $(CI_DOCKERFILE)/* $$TEMP/ && \
 	echo "Build dir is $$TEMP" && \
 	$(CONTAINER_ENGINE) build \
 	$(CONTAINER_BUILD_ARGS) --pull-always \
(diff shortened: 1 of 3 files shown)


M4rtinK (approved):
Looks good to me - nice improvement! :)

VladimirSlavik:
Interesting, the tests failed but locally it worked. I'll fix this.

VladimirSlavik:
The modules were provided incorrectly by the mounted repo. So this needs also `eslint-config-standard{,-jsx,-react}` according to [.eslintrc.json line 7](https://github.com/rhinstaller/anaconda/blob/master/ui/webui/.eslintrc.json#L7) and maybe also stuff from line 16 just below.

VladimirSlavik:
I don't understand why this doesn't work. Something about the CI build and run on the same machine, probably. 

The numbers would be roughly 65 MB instead of 352 MB.

However, this is actually **useless**. We do a full build for each CI test run, so we do download all npm packages regardless of what happens when testing/linting. The best part, though, is that we first install everything for the build before tests, then drop it all, then install everything again for the tests. Reusing that first install seems to be the right direction for improvement.

Alternatively, maybe there's some proxy/cache magic that GitHub can do about this, at network level.

PR #4144, opened 2022-05-25
Web UI: Automatically change the popover position

poncovka:
Make sure that the position of a form group label popover changes automatically.


M4rtinK (approved):

jkonecny12 (approved):
It's great to see that we don't have to do calculations of the position :).

PR #4145, opened 2022-05-25
Web UI: Jump only to the previous wizard steps

poncovka:
Fix the logic for jumping between wizard screens:

* Users have to always use the Next button to move on the next step.
* Users can jump at any point to any of the previous steps.
* It means that users can never jump forward. They have to use the button.

### DIFF
--- a/ui/webui/src/components/AnacondaWizard.jsx
+++ b/ui/webui/src/components/AnacondaWizard.jsx
@@ -50,8 +50,7 @@ const getSteps = ({
     onAddErrorNotification,
     setIsFormValid,
     stepNotification,
-    stepsOrder,
-    stepsVisited
+    stepsOrder
 }) => {
     const wrapWithContext = (children, label) => {
         return (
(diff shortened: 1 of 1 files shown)


poncovka:
@M4rtinK, @VladimirSlavik, I have made it a little more strict to avoid issues with a missing partitioning that was reported by RTT. Basically, the Web UI will always behave the same way, because we don't work with visited steps at all. What do you think?

VladimirSlavik (approved):
Thank you, I like both the behavior and code simplification.

Unfortunately I think this is yet another situation where design review will eventually take precedence.

poncovka:
> Unfortunately I think this is yet another situation where design review will eventually take precedence.

I would like to merge it to fix the current functionality. We should definitely talk to Liz once she is back, but we might not be able to implement safe jumping forward in time anyway.

M4rtinK:
> I would like to merge it to fix the current functionality. We should definitely talk to Liz once she is back, but we might not be able to implement safe jumping forward in time anyway.

Agreed - it will be also better if Liz can review the design live on a new ISO revision when she gets back. So I say lets merge this as soon as possible. :)

M4rtinK (approved):
Nicely done, thanks a lot! :)

jkonecny12 (approved):
The change looks good to me.

I just wonder, we agreed that most of the steps will be optional so use can jump forward and skip them. This is just a temporary logic that we request user to press next step or did I missed something?

poncovka:
> I just wonder, we agreed that most of the steps will be optional so use can jump forward and skip them. This is just a temporary logic that we request user to press next step or did I missed something?

Optional steps should be grouped under one step, so users should be able to skip all of them at once. Anyway, we have no optional steps so far, so it is not settled yet.

PR #4148, opened 2022-05-27
Fix tests README.rst how to run one unit test

jkonecny12:
We need to specify also directory where the shell file is the same way as it is in the Makefile.am file.

Fix unittest -> pytest and the -k seems to work only with the whole test name not just prefix.


VladimirSlavik (approved):
Looks good to me. Thank you!

rvykydal (approved):
Thank you.

PR #4152, opened 2022-05-30
Let dependency installation script ask interactively

VladimirSlavik:
The installation of build dependencies happened while piping package names to xargs, so dnf could not interactively ask for [y/n].


poncovka (approved):
Thank you!

PR #4098, opened 2022-05-11
GUI: Ask for a missing passphrase during automated installations

poncovka:
If the kickstart file defines a partitioning that requires a passphrase,
show a dialog that will allow users to provide the missing passphrase.

Resolves: rhbz#2029101

PR #4153, opened 2022-05-30
GUI: Ask for a missing passphrase during automated installations

poncovka:
If the kickstart file defines a partitioning that requires a passphrase,
show a dialog that will allow users to provide the missing passphrase.

**Resolves:** rhbz#2087218
**Ported from:** https://github.com/rhinstaller/anaconda/pull/4098


VladimirSlavik (approved):
Thanks!

rvykydal (approved):
Looks good to me

PR #4154, opened 2022-05-31
Add support for rootpw --allow-ssh

rvykydal:
Resolves: rhbz#2083269


rvykydal:
Depends on https://github.com/pykickstart/pykickstart/pull/409

pyanaconda/modules/users/users.py
```diff
@@ -87,6 +87,8 @@ def process_kickstart(self, data):
         if data.rootpw.seen:
             self.set_can_change_root_password(False)
             self._rootpw_seen = True
+            if data.rootpw.allow_ssh:
+                self.set_root_password_ssh_login_allowed(True)
```
poncovka, line 91: I think we should always set this property if we get a kickstart. Please, call the following code outside of the `if` branch: `self.set_root_password_ssh_login_allowed(data.rootpw.allow_ssh)`

poncovka (commented):

VladimirSlavik (approved):
Looks good to me. Thank you!

pyanaconda/modules/users/users.py
VladimirSlavik: I didn't comment because I thought the value can be also `None`, but given how it's parsed, the above is definitely better.

rvykydal: Done

poncovka (approved):
Thanks!

jkonecny12 (approved):

rvykydal:
Added required pykickstart version

poncovka:
Btw. I guess we can remove this common issue: https://anaconda-installer.readthedocs.io/en/latest/common-bugs.html#enabling-root-password-ssh-login-via-password

rvykydal:
> Btw. I guess we can remove this common issue: https://anaconda-installer.readthedocs.io/en/latest/common-bugs.html#enabling-root-password-ssh-login-via-password

Good catch, removed in additional patch.

poncovka (approved):

rvykydal:
@poncovka @VladimirSlavik @jkonecny12 I had to update the unit tests - change the order of rootpw command in generated kickstart of some already existing kstests. I think it can be caused just by adding the --allow-ssh option to the command in pykickstart, or do you think it is something smelly ?

VladimirSlavik (approved):
I think the test order change should be "fine". It *seems* to me that the order of ks lines is determined partially by stuff that's stored or passed around in pykickstart in "normal" dictionaries. In that case, change of order by adding a parameter seems expected.

Of course it would be better to have a stable order in the output, but that would be a different change/PR/project.

poncovka (approved):
It is a non-blocking issue. We can revert it later.

rvykydal:
> > @poncovka @VladimirSlavik @jkonecny12 I had to update the unit tests - change the order of rootpw command in generated kickstart of some already existing kstests - [641278a](https://github.com/rhinstaller/anaconda/commit/641278a847e213e9edbfb07abede76146f140994) . I think it can be caused just by adding the --allow-ssh option to the command in pykickstart, or do you think it is something smelly ?
> 
> Please, report it to `pykickstart` as a bug. I think it could be fixed in the `KickstartHandler._insertSorted` method. If `pykickstart` doesn't want to fix it, we will find a workaround in Anaconda. I don't want to change tests every time we change a kickstart command. That's stupid.

https://bugzilla.redhat.com/show_bug.cgi?id=2096871

PR #4156, opened 2022-05-31
fix typo

copperii:
remove extra at (erase at at...)


VladimirSlavik (approved):
Thanks!

M4rtinK:
> @M4rtinK I guess we need to update pixel tests for this. I would suggest to update this PR directly by you, what do you think?
Agreed, I'll handle it that way. :)

M4rtinK:
I've rebased the branch and updated the pixel test reference.

PR #4157, opened 2022-06-01
Check if the mount point exists before calling statvfs (#1826760)

poncovka:

(cherry-picked from a commit 5f43387)

Resolves: rhbz#1826760


rvykydal:
This PR has already been acked and merged as https://github.com/rhinstaller/anaconda/pull/4097 but I had to remove it from the branch to fix a mess I caused.

PR #4159, opened 2022-06-03
Check if run as root earlier (#1935470)

lukecdavidson:
Running anaconda from CLI without root will cause the subsequent attempts with root to fail due to permission errors on log files created in /tmp

Change to check if root earlier, preventing the creation of the log files.


poncovka:
Verified that it fixes https://bugzilla.redhat.com/show_bug.cgi?id=1935470.

poncovka (approved):
Thank you!

VladimirSlavik (approved):
Looks good to me. Thank you!

PR #4162, opened 2022-06-03
Web UI: Don't use a top-level text content

poncovka:
It will break styles of other components.


VladimirSlavik (approved):
Thank you, lgtm.

I wonder if the whole `<Text component={TextVariant.p}>` could be just `<p>`, but that's not the thing to solve here...

PR #4168, opened 2022-06-06
Display keyboard accelerator properly

jstodola:
"Free up space by removing or shrinking existing partitions" could not be selected by a keyboard accelerator key and was not displayed properly.


jstodola:
Before:
before
After:
after
After with "Alt" key:
after_alt

poncovka (approved):
Thank you!

PR #4172, opened 2022-06-07
Web UI: Remove unused code

poncovka:
* Remove `address` from the `AnacondaWizard` component.
* Don't try to replicate installation flags.

PR #4173, opened 2022-06-07
Web UI: Remove the wrapWithContext function

poncovka:
* Replace the `wrapWithContext` function with the `AnacondaPage` component.
* Use the `AnacondaPage` component on every page, so it can opt-out if required.
* Show the alert for the step notification only on the storage page.
* Use the `Bullseye` layout on the installation progress page.

**Depends on:** https://github.com/rhinstaller/anaconda/pull/4172 (so only the last commit is relevant)


M4rtinK (approved):
Looks good to me & the code is much nicer now. :)

PR #4176, opened 2022-06-07
Web UI: Show the "Checking disks" spinner

poncovka:
Show a spinner while we wait for the partitioning to be scheduled.


poncovka:
After users click on the Next button of the Installation destination screen:
Screenshot 2022-06-07 at 20-51-00 Anaconda Web UI

The installation progress is centered now.
Screenshot 2022-06-07 at 20-45-28 Anaconda Web UI

PR #4179, opened 2022-06-08
Integration testing

velezd:


ui/webui/test/helpers/integration.py
```diff
@@ -0,0 +1,112 @@
...
+from subprocess import CalledProcessError
+
+# import Cockpit's machinery for test VMs and its browser test API
+TEST_DIR = os.path.dirname(__file__)
+sys.path.append(os.path.join(TEST_DIR, "common"))
+sys.path.append(os.path.join(TEST_DIR, "helpers"))
+sys.path.append(os.path.join(os.path.dirname(TEST_DIR), "bots/machine"))
+
+from installer import Installer
+from language import Language
```
poncovka, line 29: Ehm, we should eventually fix imports in the Web UI tests. This will very soon start to collide with other Python packages. @VladimirSlavik  @M4rtinK

poncovka (commented):
Looks good to me. Thanks!

VladimirSlavik:
@velezd what do we do with this?

velezd:
@VladimirSlavik this is blocked. We need start using boot iso with payload for testing in Cockpit CI infra, in order to be able to finish installations. @M4rtinK should be working on that. Once that is done, I will update the PR and we can merge it.

ui/webui/test/check-integration-examples
```diff
@@ -0,0 +1,60 @@
+#!/usr/bin/python3
+#
+# Copyright (C) 2021 Red Hat, Inc.
```
jkonecny12, line 3: Please fix the year.

ui/webui/test/helpers/progress.py
```diff
@@ -0,0 +1,46 @@
+#!/usr/bin/python3
+#
+# Copyright (C) 2021 Red Hat, Inc.
```
jkonecny12, line 3: Also here.

jkonecny12 (commented):
Looks good to me otherwise.

rvykydal (approved):
Looks good to me. I wonder why the fedora-rawhide-boot test failed.

jkonecny12 (approved):

ui/webui/test/helpers/integration.py
```diff
@@ -0,0 +1,111 @@
...
+        self.installation_finished = True
+        self.helper_progress.reboot()
+        self.machine.wait_reboot()
+
+    def post_reboot(self):
+        self.machine.execute('echo "test" | passwd --stdin root') # Workaround for locked root account
+
+    def _testInstall(self):
+        self.installer.open()
+        self.language()
```
poncovka, line 98: Methods and functions should have verbs in their names, so they cannot be confused with properties and attributes. What about `configure_language`, `configure_storage` and `monitor_progress`?

ui/webui/test/helpers/integration.py
```diff
@@ -0,0 +1,111 @@
...
+
+    def setUp(self):
+        super().setUp()
+        self.installer = Installer(self.browser, self.machine)
+        self.helper_language = Language(self.browser, self.machine)
+        self.helper_storage = Storage(self.browser, self.machine)
+        self.helper_review = Review(self.browser)
+        self.helper_progress = Progress(self.browser)
+        self.installation_finished = False
+        self.logs_dir = os.path.join('./test_logs', type(self).__name__)
```
poncovka, line 48: I think we usually use `self.__class__.__name__` to get the name of the class.

ui/webui/test/helpers/integration.py
```diff
@@ -0,0 +1,111 @@
...
+from machine_install import VirtInstallMachine
+
+
+class IntegrationTest(MachineCase):
+    MachineCase.machine_class = VirtInstallMachine
+
+    def setUp(self):
+        super().setUp()
+        self.installer = Installer(self.browser, self.machine)
+        self.helper_language = Language(self.browser, self.machine)
```
poncovka, line 43: I don't really like the `helper_` part in the names of these attributes. Also, I recommend to keep all attributes and helper methods private to avoid collisions with `MachineCase` attributes and future test methods.

ui/webui/test/helpers/integration.py
```diff
@@ -0,0 +1,111 @@
...
+        self._add_public_key()
+        self._download_logs()
+        self.installation_finished = True
+        self.helper_progress.reboot()
+        self.machine.wait_reboot()
+
+    def post_reboot(self):
+        self.machine.execute('echo "test" | passwd --stdin root') # Workaround for locked root account
+
+    def _testInstall(self):
```
poncovka, line 96: Why is this in a camel case? What about `_run_installation`?

ui/webui/test/integration/default.py
```diff
@@ -0,0 +1,36 @@
...
+sys.path.append(TEST_DIR)
+sys.path.append(os.path.join(TEST_DIR, "common"))
+sys.path.append(os.path.join(TEST_DIR, "helpers"))
+sys.path.append(os.path.join(os.path.dirname(TEST_DIR), "bots/machine"))
+
+from integration import IntegrationTest  # pylint: disable=import-error
+from testlib import test_main  # pylint: disable=import-error
+
+class DefaultInstallation(IntegrationTest):
+    def testInstall(self):
```
poncovka, line 32: Please, don't use camel case names.

poncovka (approved):
I have some recommendations, otherwise it looks good to me. Thank you!

velezd:
@poncovka thanks for feedback

ui/webui/test/helpers/integration.py
M4rtinK: Agreed, while we do still have many cases of camel case in Anaconda, we strive to use underscores in new or modernised code. Though I'm not sure we have it written down somewhere nicely. :P

poncovka: https://anaconda-installer.readthedocs.io/en/latest/contributing.html#code-conventions This is a PEP8 convention.

M4rtinK (approved):
Looks good to me, thanks! :)

One minor thing I've noticed is the "WIP" in the name of the last commit - but I guess that will be dropped before this is merged. :)

ui/webui/test/helpers/integration.py
poncovka: I suggested to rename the conflicting methods in https://github.com/rhinstaller/anaconda/pull/4179#discussion_r1065527774, so these names can be taken by these attributes instead. For example, let's use `_language` to access the helper object and `_configure_language` to define the interaction with the language screen. However, maybe I am missing something.

ui/webui/test/helpers/integration.py
VladimirSlavik: It would be good to have absolute imports. But here they are probably absolute already :-/

ui/webui/test/helpers/progress.py
```diff
@@ -0,0 +1,46 @@
...
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with this program; If not, see <http://www.gnu.org/licenses/>.
+
+import os
+import sys
+import time
+
+HELPERS_DIR = os.path.dirname(__file__)
+sys.path.append(HELPERS_DIR)
```
VladimirSlavik, line 23: @KKoukiou I see this is used in cockpit and in our tests, but does that make sense? It feels like we're reinventing `import` here...

KKoukiou: It makes sense if you 're trying to use functions from other directories, other than the one the file is located in. 
This here does not make sense.

VladimirSlavik: Sorry, I should have been more specific. The question was about this kind of practice, not this specific example.

I am aware that if I have code that is invoked from a potentially random path, and it wants to import things that are not in python's library or site-packages or whatever, then it does have to manipulate path. I see that Cockpit uses this to import test stuff, at one place that I am aware of.

However, we have this in every test. So the question was actually, do we really need to do this for some deeper reason, or can we just fill the webui tests dirs with `__init__.py` so we can import without these dances everywhere?

VladimirSlavik: Eh, actually, yes. We can do the same as cockpit does with `parent.py`and have a sort of "path initializer" that primes python paths and enables imports without all the ugliness.

That also means this case would be sort-of-not-a-blocker, we'd rewrite all of that anyway.

KKoukiou: Cockpit imports these paths in all test files as well, see 'import parent'

https://github.com/cockpit-project/cockpit/blob/main/test/verify/check-storage-mdraid#L20

[kkoukiou@fedora cockpit]$ git grep 'import parent' | wc -l
72

VladimirSlavik: I think we're agreeing? :-)

KKoukiou: Yes, we can do the parent.py thing!

KKoukiou: sorry I did not read your reply https://github.com/rhinstaller/anaconda/pull/4179#discussion_r1085439839 before I submitted mine.

VladimirSlavik (approved):
Let's have it asap and fix it more later...

ui/webui/test/helpers/integration.py
VladimirSlavik: Related to the discussion below - let's overhaul the way we import these helpers, fixing names can be part of that.

PR #4183, opened 2022-06-09
dnf: apply the /etc/dnf/dnf.conf configuration file in the installer

rvykydal:
The 'gpgcheck', 'skip_if_unavailable' and 'best' are the only options
for which the default Base() config value (which anaconda has been using)
differs from the configuration in /etc/dnf/dnf.conf.

For the 'best' the goal is to use the value specified by the config file
to resolve the bug 1899494 caused by change of the Base() config default
by dnf.

For the 'gpgcheck' and 'skip_if_unavailable' the patch overrides
/etc/dnf/dnf.conf to the values keeping the current behavior. The
options are not exposed in Anaconda configuration in this patch as the
value is the same across configurations.

Resolves: rhbz#1899494

### DIFF
--- a/pyanaconda/modules/payloads/payload/dnf/dnf_manager.py
+++ b/pyanaconda/modules/payloads/payload/dnf/dnf_manager.py
@@ -116,9 +116,14 @@ class DNFManager(object):
     def _create_base(cls):
         """Create a new DNF base."""
         base = dnf.Base()
+        base.conf.read()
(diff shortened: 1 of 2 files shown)


poncovka (approved):
Looks good to me. Thanks!

VladimirSlavik (approved):

rvykydal:
/kickstart-test packages-and-groups-1 url-metalink repo-install groups-ignoremissing packages-multilib

PR #4184, opened 2022-06-09
Web UI: Add a log viewer

M4rtinK:
Add a log viewer that opens in a modal window when the
"Show logs" button in the upper right corner of the screen
is clicked.

The log viewer is currently hardcoded to follow /tmp/anaconda.log.

### DIFF
--- a/ui/webui/src/components/LogViewerModal.jsx
+++ b/ui/webui/src/components/LogViewerModal.jsx
@@ -1,0 +1,101 @@
+/*
+ * Copyright (C) 2022 Red Hat, Inc.
+ *
+ * This program is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation; either version 2.1 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful, but
+ * WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with This program; If not, see <http://www.gnu.org/licenses/>.
+ */
+import cockpit from "cockpit";
+
+import React, { useEffect, useState } from "react";
+
+import { LogViewer, LogViewerSearch } from "@patternfly/react-log-viewer";
+
+import { watchLogFile } from "../helpers/logs.js";
+
+import {
+    Button,
+    Modal, ModalVariant,
+    Toolbar, ToolbarContent, ToolbarItem,
+    Text, TextContent, TextVariants,
+    Flex,
+} from "@patternfly/react-core";
+
+const _ = cockpit.gettext;
+
+const AnacondaLogViewer = () => {
+    const [logData, setLogData] = useState("");
+
+    useEffect(() => {
+        const appendLogData = (newLogData, dataTag, error) => {
+            setLogData(l => l + newLogData);
+            if (error) {
+                console.log("Log file read failed.");
+                console.log(error);
+            }
+        };
+        watchLogFile("/tmp/anaconda.log", appendLogData);
+    }, []);
+
+    return (
+        <LogViewer
+          // see CSS file for a PF workaround needed
+          // for line numbers to work correctly
+          hasLineNumbers
+          isTextWrapped={false}
+          height={450}
+          data={logData}
+          id="log-viewer-element"
+          toolbar={
+              <Toolbar>
+                  <ToolbarContent>
+                      <ToolbarItem>
+                          <LogViewerSearch placeholder={_("Search value")} />
+                      </ToolbarItem>
+                  </ToolbarContent>
+              </Toolbar>
+          }
+        />
+    );
+};
+
+export const LogViewerModal = ({ setShowLogViewer }) => {
+    return (
+        <Modal
+          id="log-viewer-modal"
+          actions={[
+              <Button
+                id="log-viewer-exit-btn"
+                key="cancel"
+                onClick={() => setShowLogViewer(false)}
+                variant="primary">
+                  {_("Close")}
+              </Button>
+          ]}
+          isOpen
+          onClose={() => setShowLogViewer(false)}
+          title={_("Installer log")}
+          variant={ModalVariant.medium}
+        >
+            <Flex direction={{ default: "column" }}>
+                <TextContent
+                  id="anaconda-version-label"
+                >
+                    <Text component={TextVariants.p}>
+                        {_("Installer version") + " 38.5"}
+                    </Text>
+                </TextContent>
+                <AnacondaLogViewer />
+            </Flex>
+        </Modal>
+    );
+};
--- a/ui/webui/src/components/app.scss
+++ b/ui/webui/src/components/app.scss
@@ -15,3 +15,10 @@ html:not(.index-page) body {
 #installation-wizard .pf-c-wizard__main-body {
   flex: 1 1 auto;
 }
+
+// workaround for a LogViewer bug
+// - without this the line number column is too narrow
+//   and line numbers quickly become unreadable
+.pf-c-log-viewer__index {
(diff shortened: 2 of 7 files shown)


M4rtinK:
Snímek obrazovky 2022-06-09 v 22 12 14
This is basically a simplified version of the current log viewer mockup - it follows live updates to anaconda.log & supports search.
TODO (in separate PR):

- log file selection
- log levels
- head/tail buttons
- actual live version data
- printscreen support

ui/webui/src/components/LogViewer.jsx
```diff
@@ -0,0 +1,102 @@
...
+        <Modal
+          id="log-viewer-modal"
+          actions={[
+              <Button
+                id="log-viewer-exit-btn"
+                key="cancel"
+                onClick={() => {
+                    setShowLogViewer(false);
+                }}
+                variant="secondary">
```
poncovka, line 83: The button has a wrong colour. It should be primary.

ui/webui/src/components/AnacondaWizard.jsx
```diff
@@ -179,6 +182,10 @@ const Footer = ({ isFormValid, setIsFormValid, setStepNotification, isInProgress
                               exitGui={exitGui}
                               setQuitWaitsConfirmation={setQuitWaitsConfirmation}
                             />}
+                            {showLogViewer &&
+                            <LogViewerModal
```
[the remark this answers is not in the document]
M4rtinK: Looks like for some reason we have many modals defined here - might be a good idea to move those as well to some more sensible place.

ui/webui/src/components/LogViewer.jsx
```diff
@@ -0,0 +1,102 @@
...
+    // the need to disable eslint warnings
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+    }, []);
+
+    return (
+        <LogViewer
+          // TODO: find why line numbers are corrupted when displayed
+          hasLineNumbers={false}
+          isTextWrapped={false}
+          height={600}
```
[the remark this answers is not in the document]
poncovka: This seems to be fine.

ui/webui/src/components/AnacondaWizard.jsx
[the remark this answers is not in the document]
poncovka: These modals are there because they are somehow tied to the actions of the footer.

ui/webui/src/components/LogViewer.jsx
```diff
@@ -0,0 +1,102 @@
...
+          data={logData}
+          toolbar={
+              <Toolbar>
+                  <ToolbarContent>
+                      <ToolbarItem>
+                          <LogViewerSearch placeholder="Search value" />
```
poncovka, line 63: This string is not translated.

ui/webui/src/components/LogViewer.jsx
```diff
@@ -0,0 +1,102 @@
...
+            setLogData(logData + newLogData);
+            if (error) {
+                console.log("Log file read failed.");
+                console.log(error);
+            }
+        };
+        watchLogFile("/tmp/anaconda.log", appendLogData);
+    // TODO: find another way to run this only once without
+    // the need to disable eslint warnings
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
poncovka, line 49: It looks like it can be fixed with `setLogData(l => l + newLogData);`.

ui/webui/src/components/LogViewer.jsx
```diff
@@ -0,0 +1,102 @@
...
+    }, []);
+
+    return (
+        <LogViewer
+          // TODO: find why line numbers are corrupted when displayed
+          hasLineNumbers={false}
```
poncovka, line 55: It looks like it can be fixed with the following workaround:

```
.pf-c-log-viewer__index {
  width: auto;
}

```

However, it is probably a big in Patternfly. Please, report it.

ui/webui/src/components/LogViewer.jsx
```diff
@@ -0,0 +1,102 @@
...
+    Button,
+    Modal, ModalVariant,
+    Toolbar, ToolbarContent, ToolbarItem,
+    Text, TextContent, TextVariants,
+    Flex,
+} from "@patternfly/react-core";
+
+const _ = cockpit.gettext;
+
+export const AnacondaLogViewer = () => {
```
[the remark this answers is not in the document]
M4rtinK: OK, dropped `export`.

ui/webui/src/components/LogViewer.jsx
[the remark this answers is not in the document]
M4rtinK: OK, setting height to 450 for now & will report issue to PatternyFly team.

BTW, I kinda remember seeing a "modal window" PatternFly component somewhere when we first discussed the log viewer but could not find it again. Maybe @lclayton will know ?

ui/webui/src/components/LogViewer.jsx
M4rtinK: Perfect, thanks a lot! :)

M4rtinK:
PR updated - thanks a lot for all the feedback so far! :)
Snímek obrazovky 2022-06-10 v 17 56 06

ui/webui/src/components/LogViewerModal.jsx
```diff
@@ -0,0 +1,100 @@
...
+          ]}
+          isOpen
+          onClose={() => setShowLogViewer(false)}
+          title={_("Installer log")}
+          variant={ModalVariant.medium}
+        >
+            <Flex direction={{ default: "column" }}>
+                <TextContent>
+                    <Text component={TextVariants.p}>
+                        {_("Installer version") + " 37.10"}
```
VladimirSlavik, line 93: The hardcoded version isn't very nice, but let's not drag webpack macros into this yet.

ui/webui/src/components/LogViewerModal.jsx
```diff
@@ -0,0 +1,100 @@
...
+    return (
+        <Modal
+          id="log-viewer-modal"
+          actions={[
+              <Button
+                id="log-viewer-exit-btn"
+                key="cancel"
+                onClick={() => {
+                    setShowLogViewer(false);
+                }}
```
VladimirSlavik, line 80: I think you can use only `onClick={setShowLogViewer(false)}`, or at least drop the inner curly braces. Below, you have `onClose` written that way.

ui/webui/src/components/app.scss
```diff
@@ -15,3 +15,10 @@ html:not(.index-page) body {
 #installation-wizard .pf-c-wizard__main-body {
   flex: 1 1 auto;
 }
+
+// workaround for a LogViewer bug
+// - without this the line number column is too narrow
+//   and line numbers quickly become unreadable
+.pf-c-log-viewer__index {
+  width: auto;
```
VladimirSlavik, line 23: Doesn't this make the UI jump if new numbers are added and are longer? Eg. 99->100.

Perhaps also `min-width: 3rem;` or some such could be useful to prevent the log from jumping too much in the beginning when the render length of line numbers grows faster.

poncovka: I wouldn't mess with it more that we do now. It is a workaround for a bug in Patternfly anyway.

ui/webui/src/components/LogViewerModal.jsx
M4rtinK: I think eventually we want to just fetch it via DBus from the backend, right ? Or is it set somewhere else where we can reach, like the config file ?

ui/webui/src/components/AnacondaWizard.jsx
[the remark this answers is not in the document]
M4rtinK: Turns out this indeed fits well on the top level next to the help drawer - just took me some time to figure it out. :P

ui/webui/src/components/LogViewerModal.jsx
VladimirSlavik: For python we have `pyanaconda/version.py.in`, so I would expect the same...

VladimirSlavik (approved):
I think this should be fine.

PR #4187, opened 2022-06-10
dnf: apply the /etc/dnf/dnf.conf configuration file in the installer

rvykydal:
The 'gpgcheck', 'skip_if_unavailable' and 'best' are the only options
for which the default Base() config value (which anaconda has been using)
differs from the configuration in /etc/dnf/dnf.conf.

For the 'best' the goal is to use the value specified by the config file
to resolve the bug 1899494 caused by change of the Base() config default
by dnf.

For the 'gpgcheck' and 'skip_if_unavailable' the patch overrides
/etc/dnf/dnf.conf to the values keeping the current behavior. The
options are not exposed in Anaconda configuration in this patch as the
value is the same across configurations.

Resolves: rhbz#1899494

### DIFF
--- a/pyanaconda/payload/dnf/payload.py
+++ b/pyanaconda/payload/dnf/payload.py
@@ -653,9 +653,13 @@ class DNFPayload(Payload):
     def _configure(self):
         self._base = dnf.Base()
         config = self._base.conf
+        config.read()
         config.cachedir = DNF_CACHE_DIR
         config.pluginconfpath = DNF_PLUGINCONF_DIR
         config.logdir = '/tmp/'
+        # set installer defaults
+        config.gpgcheck = False
+        config.skip_if_unavailable = False
         # enable depsolver debugging if in debug mode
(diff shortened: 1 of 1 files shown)


poncovka (approved):
Thanks!

VladimirSlavik (commented):
Is it correct to set the paths only after reading the conf?

rvykydal:
> Is it correct to set the paths only after reading the conf?

Yes, I was wondering as well, but looking at the documentation https://dnf.readthedocs.io/en/latest/api_conf.html#dnf.conf.Conf.read only single file (by default /etc/dnf/dnf.conf) is read so it should be OK I think.

VladimirSlavik (approved):
Thanks, then lgtm.

rvykydal:
Kickstart test result is available in Checks ( https://github.com/rhinstaller/anaconda/pull/4187/checks?check_run_id=6829488516 )

PR #4188, opened 2022-06-10
Add support for NPIV enabled zFCP devices

jstodola:
This change allows to use NPIV-enabled zFCP devices in anaconda. Such devices can be configured just based on their device number, WWPN and LUN are useless.
Also update the width of the "Add zFCP" dialog, since with additional text it would use the whole with of the screen.

TODO:
- [x] update required pykickstart version (pykickstart  [PR-410](https://github.com/pykickstart/pykickstart/pull/410))
- [x] update required python-blivet version (blivet [PR-1035](https://github.com/storaged-project/blivet/pull/1035))

Related: rhbz#1937032

### DIFF
--- a/pyanaconda/modules/storage/zfcp/discover.py
+++ b/pyanaconda/modules/storage/zfcp/discover.py
@@ -51,18 +51,29 @@ class ZFCPDiscoverTask(Task):
         if not DASD_DEVICE_NUMBER.match(self._device_number):
             raise StorageDiscoveryError("Incorrect format of the given device number.")
 
-        if not ZFCP_WWPN_NUMBER.match(self._wwpn):
+        if self._wwpn and not ZFCP_WWPN_NUMBER.match(self._wwpn):
             raise StorageDiscoveryError("Incorrect format of the given WWPN number.")
 
-        if not ZFCP_LUN_NUMBER.match(self._lun):
+        if self._lun and not ZFCP_LUN_NUMBER.match(self._lun):
             raise StorageDiscoveryError("Incorrect format of the given LUN number.")
+
+        # Zfcp automatic LUN scan requires just the device number to be provided by the user.
+        # If zfcp auto LUN scan is not available, the user has to specify the device number, WWPN
+        # and LUN.
+        if not ((self._device_number and not self._wwpn and not self._lun)
+                or (self._device_number and self._wwpn and self._lun)):
+            raise StorageDiscoveryError(
+                "Only device number or device number with WWPN and LUN are allowed."
+            )
 
     def _sanitize_input(self):
         """Sanitize the input values."""
         try:
             self._device_number = blockdev.s390.sanitize_dev_input(self._device_number)
-            self._wwpn = blockdev.s390.zfcp_sanitize_wwpn_input(self._wwpn)
-            self._lun = blockdev.s390.zfcp_sanitize_lun_input(self._lun)
+            if self._wwpn:
+                self._wwpn = blockdev.s390.zfcp_sanitize_wwpn_input(self._wwpn)
+            if self._lun:
+                self._lun = blockdev.s390.zfcp_sanitize_lun_input(self._lun)
         except (blockdev.S390Error, ValueError) as err:
             raise StorageDiscoveryError(str(err)) from err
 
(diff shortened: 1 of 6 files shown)


pyanaconda/modules/storage/zfcp/discover.py
```diff
@@ -51,18 +51,26 @@ def _check_input(self):
...
-        if not ZFCP_WWPN_NUMBER.match(self._wwpn):
+        if self._wwpn and not ZFCP_WWPN_NUMBER.match(self._wwpn):
             raise StorageDiscoveryError("Incorrect format of the given WWPN number.")
 
-        if not ZFCP_LUN_NUMBER.match(self._lun):
+        if self._lun and not ZFCP_LUN_NUMBER.match(self._lun):
             raise StorageDiscoveryError("Incorrect format of the given LUN number.")
 
+        if not ((self._device_number and not self._wwpn and not self._lun)
+                or (self._device_number and self._wwpn and self._lun)):
```
VladimirSlavik, line 61: I think it might be more readable if we knew what's actually being tested...
```
wwpn_lun_only_one_set = bool(self._wwpn) != bool(self._lun)
if not self._device_number or wwpn_lun_only_one_set:
   raise (...)
```

I also think this might be better at the front, since it's cheaper to compute than the regexps.

poncovka: I don't think that's better. If we really want to simplify the condition, the device number can be checked separately with a message `"The required device number is missing."` or something. Anyway, I am fine with the current solution.

jstodola:
Updated required versions of blivet and pykickstart, unit tests updated and temporary fixes reverted. Ready for review.

VladimirSlavik:
@jstodola just a heads up, if #4206 is merged earlier, you will have to keep the higher pykickstart version. Given that 3.41 is now in rawhide and tests won't work without that fix, I think it's very likely this will happen.

VladimirSlavik (approved):
Thanks!

VladimirSlavik:
The unit tests are failing everywhere on master because of Python updates in Rawhide.

rvykydal (approved):
Looks good to me. Thank you.

steffen-maier (changes requested):
Hi Jan, thanks a lot for the code changes. Very much appreciated. I cannot wait to try them. Regarding user messages and code maintenance, I suggest small changes to comments, user messages, and code indentifiers. They are based on potential conceptual misunderstandings:

- The code change is about enabling users to take advantage of a Linux kernel (module) feature called zfcp automatic LUN scan.
- It is not about enabling NPIV-enabled FCP devices. (Users deploy them since NPIV support appeared in IBM Z HBAs and the zfcp device driver; using NPIV is best practice since many years)

Those are two different things.

steffen-maier (approved):

VladimirSlavik:
The first failure tells you to use the spelling "zFCP" instead of "zfcp" in the glade file.

The second says that `zfcp.allow_lun_scan=0` is interpreted as GTK accelerators. Not 100% sure how to solve that, but I see you don't have a screenshot with the string visible. Maybe try seeing if it actually looks good - GTK should interpret `_` as accelerator and you should have to use `__` (that is, repeat `_` twice) to get "_" in the visible label. I'd check that first. Using some markup for the option might be a good idea too.

jstodola:
Updated "zfcp-attached SCSI disks" to "zFCP-attached SCSI disks".
Also removed `zfcp.allow_lun_scan=0` from the label - it was causing failures of both unit tests and I couldn't figure out a quick fix for the `_` problem. (Trying HTML code `&#95;` instead of `_` didn't fool the unit test)

tests/unit_tests/pyanaconda_tests/modules/storage/test_module_zfcp.py
```diff
@@ -74,9 +74,6 @@ def test_discovery_fails(self):
         with pytest.raises(StorageDiscoveryError):
             ZFCPDiscoverTask("", "", "").run()
 
-        with pytest.raises(StorageDiscoveryError):
-            ZFCPDiscoverTask("0.0.fc00", "", "").run()
-
         with pytest.raises(StorageDiscoveryError):
             ZFCPDiscoverTask("0.0.fc00", "0x5105074308c212e9", "").run()
```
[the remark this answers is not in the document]
poncovka: Fixed at https://github.com/rhinstaller/anaconda/pull/4246.

PR #4189, opened 2022-06-10
Web UI: Fix betanag popover position

M4rtinK:
First set the popover placement priority to prefer the bottom direction
and then make it narrower so it actually fits if shown under the betanag
bubble.

### DIFF
--- a/ui/webui/src/components/AnacondaHeader.jsx
+++ b/ui/webui/src/components/AnacondaHeader.jsx
@@ -22,7 +22,7 @@ import {
     Flex,
     Label,
     PageSection, PageSectionVariants,
-    Popover,
+    Popover, PopoverPosition,
     TextContent, Text, TextVariants
 } from "@patternfly/react-core";
 import { InfoCircleIcon } from "@patternfly/react-icons";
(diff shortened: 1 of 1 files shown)


VladimirSlavik (approved):
Thank you, looks good to me.

martinpitt:
I took a deeper look at these npm install failures, see https://github.com/cockpit-project/cockpituous/pull/498 . Retried. Please retry other PRs which got affected by that.

ui/webui/src/components/AnacondaHeader.jsx
```diff
@@ -36,7 +36,8 @@ export const AnacondaHeader = ({ beta, title }) => {
         ? (
             <Popover
               headerContent={_("This is unstable, pre-release software")}
-              minWidth="40rem"
+              minWidth="30rem"
+              flipBehavior={["bottom", "right", "left", "top"]}
```
poncovka, line 40: Have you tried to set `position={PopoverPosition.auto}` instead?

M4rtinK: > Have you tried to set `position={PopoverPosition.auto}` instead?

Thanks, that works better! But still needs the `minWidth` to be set to 30rem or else it still won't fit.

VladimirSlavik (approved):

PR #4190, opened 2022-06-10
Web UI: Make it possible to close the disks alert

M4rtinK:
Make it possible to close the disks alert on the installation
destination spoke. There is one final warning about data
being wiped before the start of the installation, so this change should
be reasonably safe.


M4rtinK:
Snímek obrazovky 2022-06-10 v 18 54 37
Now you see it.
Snímek obrazovky 2022-06-10 v 18 54 42
Now you don't.

NOTE: This is just about the hardcoded info-level alert. The danger-level alerts will be more invovlved if we want to aslo make them dissmissable.

martinpitt:
I took a deeper look at these npm install failures, see https://github.com/cockpit-project/cockpituous/pull/498 . Retried. Please retry other PRs which got affected by that.

poncovka (approved):

VladimirSlavik (approved):
Thanks!

PR #4192, opened 2022-06-10
Speed up disk spoke

mikhailnov:
Fix Anaconda freeze for 6-8 sec when leaving the storage spoke
by calculating required space not by calculating size of each file inside /,
but by taking a ready to use size of the mounted squashfs image.

Idea and implementation by Oleg Solovev @Gel0bmstu 

P.S. It is a port of the patch for the rhel-9 branch (https://abf.io/import/anaconda/blob/573b765478/0030-ROSA-fix-aloithm-for-determining-required-disk-space.patch), I am not sure that I've ported it correctly, especially in the place of `return get_dir_size`, please check...

### DIFF
--- a/pyanaconda/core/constants.py
+++ b/pyanaconda/core/constants.py
@@ -257,6 +257,8 @@ CMDLINE_LIST = ["addrepo"]
 # An estimated ratio for metadata size to total disk space.
 STORAGE_METADATA_RATIO = 0.1
 
+STORAGE_GROW_RATIO = 1.1
+
 # Constants for reporting status to IPMI.  These are from the IPMI spec v2 rev1.1, page 512.
 IPMI_STARTED = 0x7          # installation started
 IPMI_FINISHED = 0x8         # installation finished successfully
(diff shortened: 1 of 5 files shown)


mikhailnov:
Hm, I think you are right, I have to minify the patch. Its initial description was confusing. Can you please comment, is return of disk_usage(DRACUT_ISODIR) correct?

poncovka (changes requested):
I am going to block this change for now. It took a lot of effort to make this code work right, so I don't want to mess with it unless it is absolutely necessary. If  `get_dir_size("/")` takes a lot of time, it shouldn't be calculated in the property, but in the `SetUpLiveOSSourceTask` class. We do something similar in `SetUpLocalImageSourceTask`. Let's start from there and see if any further changes are still needed.

poncovka:
> If `get_dir_size("/")` takes a lot of time, it shouldn't be calculated in the property, but in the `SetUpLiveOSSourceTask` class. We do something similar in `SetUpLocalImageSourceTask`. Let's start from there and see if any further changes are still needed.

This will be fixed with https://github.com/rhinstaller/anaconda/pull/4247/.

mikhailnov:
Great, thanks!

PR #4161, opened 2022-06-03
Fix pylint issues (#infra)

poncovka:
* Remove unused pylint false positives.
* Remove the `no-space-check` option from `pylintrc`.

PR #4195, opened 2022-06-14
Remove the no-space-check option from pylintrc (#infra)

VladimirSlavik:
Port of #4161 to fix tests and thus also container builds on the f36 branches.

----

The no-space-check option was removed in Pylint 2.6.

(cherry picked from commit 6142a4d9d6709bbe52ab4969fa8a0239d1cbdd23)

### DIFF
--- a/tests/pylint/pylintrc
+++ b/tests/pylint/pylintrc
@@ -253,13 +253,6 @@ max-line-length=99
 # Maximum number of lines in a module.
 max-module-lines=1000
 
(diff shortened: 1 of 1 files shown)


VladimirSlavik:
Tests fail only for master, which is a separate issue. The important part is that the f36 tests all succeeded.

poncovka (approved):
I guess we could actually disable the F36 tests.

M4rtinK (approved):
Looks good to me as well. :)

PR #4199, opened 2022-06-15
kstest on pr: disable Permian GH ReportSender temporarily (#infra)

rvykydal:
There is missing status reporting and link to details before Permian
runs or if Permian does not run at all due to an error in other job of
the workflow (ie building rpms).


VladimirSlavik (approved):
I wondered about the missing statuses. Thanks!

poncovka (approved):

PR #4201, opened 2022-06-16
Temporarily ignore the new version of the zfcp command

poncovka:
Skip the version check of the `zfcp` kickstart command.


VladimirSlavik (approved):
Thank you!

PR #4202, opened 2022-06-16
containers: Don't downgrade npm (#infra)

VladimirSlavik:
It appears that the former fix is now a bug: Now it's the older npm that fails to connect to the registry. This manifests only on Fedora, which means developer machines and kstest runners. Anything running on the Github hosted runners did not fail, though.

This de facto reverts commits:
  701539bb600ed728e7dccd229b32c9c20b3f8a40
  aae34587bc246d31eff2b229f40b1ccc87adb9b5


VladimirSlavik:
Yes, that worked. https://github.com/rhinstaller/anaconda/actions/runs/2510435116

PR #4208, opened 2022-06-22
bootloader: do not consider non-ibft iscsi disk as usable for bootloader

rvykydal:
Resolves: rhbz#2003519


VladimirSlavik (approved):
Looks good to me. Thank you!

PR #4209, opened 2022-06-23
TUI software spoke is ready with failing repos, too

VladimirSlavik:
If some repos fail to load, the PayloadManager thread will report error and base repo will not be set, so consider errors as the spoke state being resolved (ready), too.

Resolves: [rhbz#2070182](https://bugzilla.redhat.com/show_bug.cgi?id=2070182)
Resolves: [rhbz#2020299](https://bugzilla.redhat.com/show_bug.cgi?id=2020299)

(Bugs to be deduplicated if possible)

### DIFF
--- a/pyanaconda/ui/tui/spokes/software_selection.py
+++ b/pyanaconda/ui/tui/spokes/software_selection.py
@@ -303,7 +303,7 @@ class SoftwareSpoke(NormalTUISpoke):
         return not threadMgr.get(THREAD_SOFTWARE_WATCHER) \
             and not threadMgr.get(THREAD_PAYLOAD) \
             and not threadMgr.get(THREAD_CHECK_SOFTWARE) \
-            and self.payload.base_repo is not None
+            and ((self.payload.base_repo is not None) or self.errors)
 
     def apply(self):
         """Apply the changes."""


VladimirSlavik:
Tests won't pass until #4210 is merged too.

rvykydal (approved):
LGTM, makes sense to me.
I am wondering if (or why not) we do need it also in GUI, but that would be for another discussion I think.

VladimirSlavik:
>  I am wondering if (or why not) we do need it also in GUI, but that would be for another discussion I think.

Absolutely. Actually, during the quarterly planning, I saw this which seems to be that? https://bugzilla.redhat.com/show_bug.cgi?id=2001633

VladimirSlavik:
The failed kickstart tests are actually disabled, so that seems to be fine:
- repo-include = gh670
- packages-weakdeps = rhbz1960279

M4rtinK (approved):
Looks good to me. :)

jkonecny12 (approved):
Two flies with one shot! Great!

VladimirSlavik:
I think there are more duplicates, will ask rtt to check while verifying.

pyanaconda/ui/tui/spokes/software_selection.py
```diff
@@ -303,7 +303,7 @@ def ready(self):
         return not threadMgr.get(THREAD_SOFTWARE_WATCHER) \
             and not threadMgr.get(THREAD_PAYLOAD) \
             and not threadMgr.get(THREAD_CHECK_SOFTWARE) \
-            and self.payload.base_repo is not None
+            and ((self.payload.base_repo is not None) or self.errors)
```
poncovka, line 306: I am afraid this means that the Software Selection screen can be now entered even if the installation source is not properly set up. That can have unexpected consequences.

jkonecny12: @jstodola @vslavik that could be problematic. Could you please take a look on that?

VladimirSlavik: I will, once I know what to do about it...

jstodola: I didn't encounter any unexpected behavior with anaconda-34.25.1.11-1.el9 (which includes this change) when the installation source is not set or when it is set incorrectly (like having an unusable URL). The Software selection spoke just complains that "Installation source needs to be set up first." - so the same behavior like in RHEL-9.0.

poncovka: Oh, right. The Software Selection screen can actually survive that in TUI. Let's leave it as it is in that case.

VladimirSlavik:
The rewrite for DNF manager changed pretty much everything, so porting this will need more work. Practically speaking, the porting would be probably replacing ```return not self._processing_data and self._source_is_set```
with  ```return not self._processing_data and (self._source_is_set or self._errors)``` but I'm not sure if the underlying logic is broken in the same way or not.

poncovka:
Let's discuss it tomorrow. I have some questions and ideas.

PR #4216, opened 2022-06-27
Fix unit tests for python 3.11

VladimirSlavik:
With this, unit tests can succeed.

Pylint is not fixed - its most current breakage is due to interface changes in the new Python disagreeing with `dill`, whatever that is. See https://github.com/uqfoundation/dill/issues/514 and https://github.com/uqfoundation/dill/issues/480. Note that this PR does not change anything requiring rebuild of containers, so pylint appears to work here, but it's just stale container being used.

### DIFF
--- a/tests/Makefile.am
+++ b/tests/Makefile.am
@@ -48,7 +48,6 @@ dist_check_SCRIPTS = $(srcdir)/glade_tests/*.py \
 		     $(srcdir)/lib/*.sh \
 		     unit_tests/unit_tests.sh \
 		     rpm_tests/rpm_tests.sh \
-		     pylint/runpylint.py \
 		     cppcheck/runcppcheck.sh \
 		     testenv.sh \
 		     $(srcdir)/gettext_tests/*.py \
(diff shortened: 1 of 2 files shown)


hroncok:
About pylint: I don't recommend running linters on not yet released Python versions.

VladimirSlavik:
Yes, it's an enum. Thank you for the information, as well as pushing to make this visible. (I *did* read the what's new pretty carefully...)

hroncok:
> Well, is it an option to just not take Rawhide's last Python? It doesn't feel that way from here.

I don't know what are you doing (or trying to achieve), so I cannot answer your question.

- If you try to cover multiple Python versions and install tools such as pylint via pip, you can use any Python version available on Rawhide to run the lint job.
- If you try to cover multiple Fedora versions and you install tools such as pylint via official Fedora repos, I'd just not lint on Rawhide.

rvykydal (approved):
Looks good to me, thank you.

PR #4218, opened 2022-06-27
Add support for NPIV enabled zFCP devices

jstodola:
This change allows to use NPIV-enabled zFCP devices in anaconda. Such devices can be configured just based on their device number, WWPN and LUN are useless.
Also update the width of the "Add zFCP" dialog, since with additional text it would use the whole with of the screen.

Ported from https://github.com/rhinstaller/anaconda/pull/4188

Resolves: rhbz#1497086

### DIFF
--- a/anaconda.spec.in
+++ b/anaconda.spec.in
@@ -33,7 +33,7 @@ Source0: %{name}-%{version}.tar.bz2
 %define libxklavierver 5.4
 %define mehver 0.23-1
 %define nmver 1.0
-%define pykickstartver 3.16.11-1
+%define pykickstartver 3.16.15-1
 %define pypartedver 2.5-2
 %define rpmver 4.10.0
 %define simplelinever 1.1-1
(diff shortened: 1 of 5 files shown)


jstodola:
pykickstart-3.16.15-1.el8 is needed for the unit/rpm tests to pass.

rvykydal (approved):
Thank you

jstodola:
Updated comments and user visible text based on feedback from @steffen-maier in https://github.com/rhinstaller/anaconda/pull/4188#pullrequestreview-1023617370

steffen-maier:
> Should zfcp-attached be all lowercase?

I don't have a very strict opion. As zfcp developer, I kind of settled with lowercase as it's easier to type and any kind of uppercase variants such as zFCP or ZFCP seem somewhat made up to me. But then again I wasn't involved in creating the device driver so I might miss on earlier naming conventions. Feel free to change casing as you chose.

VladimirSlavik:
As a former tech writer, I'd use "zFCP" when talking about the technology in general, which is the case here.

The other "right" option is to mark it up like `zfcp` when talking about some technical item where the case matters, such as package or tool. But that's not case here, I think?

jstodola:
Just updated the label text (see https://github.com/rhinstaller/anaconda/pull/4188#issuecomment-1172878333 for details).
But unit-tests/rpm-tests are still failing with `fatal: It seems that there is already a rebase-merge directory, and I wonder if you are in the middle of another rebase....` developer's help is needed...

VladimirSlavik:
The git problem went away somehow, but the tests now hit https://github.com/rhinstaller/pocketlint/issues/45.

jkonecny12 (approved):

PR #4219, opened 2022-06-27
Media verification dialog button reflects state

VladimirSlavik:
Previously, the button was always labeled "Cancel". Now it is:
- "Cancel" when scan is in progress,
- "OK" when scan is finished.

Resolves: [rhbz#2057469](https://bugzilla.redhat.com/show_bug.cgi?id=2057469)

### DIFF
--- a/pyanaconda/ui/gui/spokes/installation_source.glade
+++ b/pyanaconda/ui/gui/spokes/installation_source.glade
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<!-- Generated with glade 3.22.1 -->
+<!-- Generated with glade 3.38.2 -->
 <interface>
   <requires lib="gtk+" version="3.0"/>
   <requires lib="AnacondaWidgets" version="1.0"/>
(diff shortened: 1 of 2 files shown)


VladimirSlavik:
I trust I don't have to add a "before" picture, the button just always said the same.

Screenshot_vs_isos_2022-06-27_19:24:57

Screenshot_vs_isos_2022-06-27_19:22:23

VladimirSlavik:
Hello @mairin, I think this is something that could use your help. The dialog shows only a text which differs only in presence of "not" in the middle of the sentence. I thought this could also use some larger indicator, such as an icon below the progress bar - hourglass when checking, red x or green check when done. What do you think?

mairin:
@VladimirSlavik Ah yes! Just one word change and it's kind of barely noticeable what happened. I think your intuition of having a large indicator like an icon below the progress bar is a great idea. I think you don't necessarily need the hourglass while checking --- having the white space get replaced with the large icon I think would draw more attention to it and the state. I would add a label below the progress bar to say "Checking media" or something like that. Just some visual doodling here, following some of the GNOME dialog icon layouts like this...

Screenshot from 2022-06-27 16-31-51
Screenshot from 2022-06-27 16-31-45

(Apologies for the code switching, I took a Czech class a long time ago and have lost most of it, haha)

mairin:
> Thanks! I'm still a bit hesitant about the icon for the "bad" state - this is `dialog-error-symbolic`, other options are `emblem-important-symbolic`, `x-circular-symbolic`, `cross-filled-symbolic`, `warning-symbolic`...

I didn't have time last night to mock it up but I had picked out warning-symbolic in the mockup for the bad media check. It's better too because it's a different shape, triangle vs circle.

VladimirSlavik:
Tests failed due to new pykickstart, unrelated.

I'd like to move the dialogs etc. out of the spoke module and gladefile, to a helper, as a last step.

@jstodola, is it acceptable to bundle also the gui changes and refactoring with the "bugfix"? I will be making these upstream anyway, the question is really about RHEL.

jstodola:
The changes look good so far, feel free to include them in the bug fix.

M4rtinK (approved):
Looks good to me, thanks! :)

VladimirSlavik:
> I'd like to move the dialogs etc. out of the spoke module and gladefile, to a helper, as one more commit.

Not really worth it - the source file is so huge that it won't matter anyway. Let's call this done.

jkonecny12:
Based on https://github.com/rhinstaller/anaconda/pull/4219#issuecomment-1180440684 adding ready to merge.

PR #4220, opened 2022-06-28
simplify TestValues enum creation and usage

ethanfurman:
Enhance #4216.


ethanfurman:
Wow, it's actually embarassing how many times I forget to `return member` at the end of `__new__`.

It should work *much* better now.  :-)

ethanfurman:
Specifying `None` on the three broken variants is a matter of taste.  I wouldn't usually do it myself, but I have no experience with your code base to know which is more consistent with its style.

As for squashing, feel free to do so as I have no idea how -- my git skills are fairly rudimentary.

VladimirSlavik:
So, I squashed it and made the `None` explicit, so it's easier to get for the next generations. Your local branch is likely "broken" for you now, unless you delete it and pull from your fork.

VladimirSlavik (approved):

PR #4217, opened 2022-06-27
Add support for NPIV enabled zFCP devices

jstodola:
This change allows to use NPIV-enabled zFCP devices in anaconda. Such devices can be configured just based on their device number, WWPN and LUN are useless.
Also update the width of the "Add zFCP" dialog, since with additional text it would use the whole with of the screen.

Ported from https://github.com/rhinstaller/anaconda/pull/4188

Resolves: rhbz#1937032

PR #4222, opened 2022-06-29
Use RHEL 9-specific ZFCP command from pykickstart 3.32.8

VladimirSlavik:
This solves the test failures on `rhel-9` branch. However, it also duplicates part of #4217.

It would be preferable if #4217 was merged immediately. Then, this could be removed.

Resolves: rhbz# ???? TODO !!!! (needs a bug in case we do need this as a standalone change)

### DIFF
--- a/anaconda.spec.in
+++ b/anaconda.spec.in
@@ -35,7 +35,7 @@ Source0: %{name}-%{version}.tar.bz2
 %define libxklavierver 5.4
 %define mehver 0.23-1
 %define nmver 1.0
-%define pykickstartver 3.32.7-1
+%define pykickstartver 3.32.8-1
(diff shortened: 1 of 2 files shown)


M4rtinK (approved):
Looks good to me code wise. :)

poncovka:
Please, never bump versions of kickstart commands without related changes in the installer. It can easily break something. You should temporarily disable the failing tests instead like in https://github.com/rhinstaller/anaconda/pull/4201.

poncovka (changes requested):
I don't think this is a good idea.

VladimirSlavik:
Fair enough. I think prioritizing #4217 is the better strategy anyway.

PR #4227, opened 2022-07-13
Fixed issues infra-check concurrency and stale results

OndrejZobal:
- Infra-check wouldn't remove it' result when the "infrastructure" tag
   was removed because it was missing the "unlabeled" trigger.
 - Infra-check would run multiple times when opening a PR. This
   happened because a single event could activate multiple triggers (like
   opening a PR would trigger "opened" and than "labeled" for each of
   it's labels). Fixed by adding a concurrency group.
 - Describe infra-check behaviour.


.github/workflows/infra-check.yml
```diff
@@ -1,13 +1,25 @@
-# Check if only infrastructure files are changed when infrastructure label is set.
+# Check if only infrastructure files were changed. Runs only when the infrastructure label is set.
+# - if changes in a commit WITH #infra in it's title CHANGES NON-INFRA FILES, the workflow FAILS.
+# - if changes in a commit WITHOUT #infra in it's title CHANGES NON-INFRA FILES, the workflow SUCCEEDS and
+# lists all non-infra files changed in a GitHub comment.
```
poncovka, line 4: Please, add some spaces at the beginning to align the line with the one above.

poncovka (commented):
Looks good to me, but it should be approved by @jkonecny12.

VladimirSlavik (approved):
Looks good to me. Thank you!

poncovka (approved):

jkonecny12 (approved):
Thanks a lot for fixing my mess here and giving it a nice explanation. Looks good to me.

This is a great improvement! You won't see multiple comments after creating of the PR. However, it will still comment on the same SHA commit when you just change any label.
If we want to improve that, it should be doable to just look on the GH Action payload and compare the current SHA of the last commit with the last successful/failed run of this workflow for this target branch.
What I'm describing above shouldn't be such a pain so I'm still happy with merging this version instead of the ideal solution.

VladimirSlavik:
I guess the question here is: Do we want to merge this fast, or wait for any potential further fixes?

poncovka:
> I guess the question here is: Do we want to merge this fast, or wait for any potential further fixes?

I am fine with a follow-up pull request.

OndrejZobal:
So I had another look at it, and I think I got it working without any useless comments. The main think is that i replaced concurrency group with a check that blocks label triggers from unrelated labels from creating comments.

I had to put this check inside of the test script and not as a workflow if, because failing that would change the status to `skipped`.

I also removed "opened" trigger as it wasn't needed and just caused duplication of comments.

Right now the only way to trigger useless duplicate comments (that I know of) is adding and removing the `infrastructure` tag repeatedly, which I don't think is a problem.

VladimirSlavik (approved):
Looks good to me. Thank you!

rvykydal (approved):
Looks good to me.

VladimirSlavik:
Are we waiting for anything here? I'd like to either see this merged or know it will be updated...

(We can ignore the kickstart test for infra.)

OndrejZobal:
@VladimirSlavik Yes, @jkonecny12 has to review it again. Than I will probably squash it or at least remove the concurrency commit.

VladimirSlavik:
Ah, ok, makes sense. Just make sure you ask him directly once he's available, he often overlooks notifications because there are too many :-/

.github/workflows/infra-check.yml
```diff
@@ -86,8 +81,14 @@ jobs:
 
           done
 
-          if [ "$failed_check" == "yes" ]; then
-            gh pr comment ${{ github.event.pull_request.number }} -F ./message.txt
+          # Label related actions should only trigger comment if the "infrastructure" label triggered it specifically.
+          skip_comment="no"
+          if ([ "${{ github.event.action }}" == "labeled" ] || [ "${{ github.event.action }}" == "unlabeled" ]) && [ "${{ github.event.label.name }}" != "infrastructure" ]; then
```
jkonecny12, line 86: I wasn't aware there is a way to find out which label was added. This is genius solution!

jkonecny12 (approved):
Great solution, thanks a lot for this! And sorry for late review.

poncovka (approved):

PR #4228, opened 2022-07-14
Enable pylint

VladimirSlavik:
Let's install pylint and some of its deps from packages. The packages in rawhide are patched to work with the new python, ahead of what's in pip, so they do work.

That brings also another round of plugging holes in pylint false positives.

Finally, let's ignore the crypt module deprecation for now.


poncovka (approved):
Thank you!

PR #4231, opened 2022-07-22
Prefer GPT instead of legacy MBR on Fedora (`inst.gpt` version)

poncovka:
**This is an alternative to the `inst.disklabel=<gpt|mbr>` solution.**
Only one of these solutions will be merged. See https://github.com/rhinstaller/anaconda/pull/4104#issuecomment-1192445360.

### DIFF
--- a/data/profile.d/fedora.conf
+++ b/data/profile.d/fedora.conf
@@ -15,6 +15,7 @@ default_on_boot = FIRST_WIRED_WITH_LINK
 efi_dir = fedora
 
 [Storage]
+gpt = True
 default_scheme = BTRFS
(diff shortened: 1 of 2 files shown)


VladimirSlavik (approved):
Tentatively approving, but I like the other version more.

vojtechtrefny:
> `fidsk` throws a warning when running it on a GPT disk.

What warning? I don't remember seeing any warning when using fdisk with GPT, fdisk had a full support for GPT for years now. We are even using it (the libfdisk library, not the CLI) in the installer for some GPT specific things that are not supported by (lib)parted (which is used for partitioning in the installer).

PR #4232, opened 2022-07-22
Prefer GPT instead of legacy MBR

poncovka:
Use the `inst.disklabel` boot option to specify a preferred disk label type. Specify
`gpt` to prefer creation of GPT disk labels. Specify `mbr` to prefer creation of MBR
disk labels if supported. The `inst.gpt` boot option is deprecated and will be removed
in future releases.

The default value of the preferred disk label type is specified by the `disk_label_type`
option in the Anaconda configuration files. The `gpt` configuration option is no longer
supported.

Fedora Linux systems installed on legacy x86 BIOS systems should get GPT partitioning by
default instead of legacy MBR partitioning. This should be a new default for all products.

See: https://fedoraproject.org/wiki/Changes/GPTforBIOSbyDefault

### DIFF
--- a/data/anaconda.conf
+++ b/data/anaconda.conf
@@ -183,7 +183,7 @@ btrfs_compression =
 #
 #    If not specified, use whatever Blivet uses by default.
 #
-disk_label_type =
+disk_label_type = gpt
 
 # Default file system type. Use whatever Blivet uses by default.
 file_system_type =
--- a/data/anaconda.conf
+++ b/data/anaconda.conf
@@ -166,9 +166,6 @@ dmraid = True
 # Enable iBFT usage during the installation.
 ibft = True
 
-# Do you prefer creation of GPT disk labels?
-gpt = False
-
 # Tell multipathd to use user friendly names when naming devices during the installation.
 multipath_friendly_names = True
 
@@ -177,6 +174,16 @@ allow_imperfect_devices = False
 
 # Btrfs compression algorithm and level. e.g. zstd:1
 btrfs_compression =
+
+# Default disk label type.
+# Valid values:
+#
+#    gpt  Prefer creation of GPT disk labels.
+#    mbr  Prefer creation of MBR disk labels.
+#
+#    If not specified, use whatever Blivet uses by default.
+#
+disk_label_type =
 
 # Default file system type. Use whatever Blivet uses by default.
 file_system_type =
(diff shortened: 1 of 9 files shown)


Conan-Kudo:
This looks good, but is there a reason we want Anaconda's core default to remain MBR?

M4rtinK (approved):
Looks good to me code-wise, thanks! :)

Conan-Kudo (approved):
Oh, I see. We now use Blivet's setting by default.

rvykydal (approved):
Looks good to me.

VladimirSlavik (approved):
Looks good to me. Thank you!

poncovka:
> This looks good, but is there a reason we want Anaconda's core default to remain MBR?

No idea. @jkonecny12? @jstodola?

jstodola:
I'm not aware of any reason why MBR should remain the default. IMO, it makes sense to make GPT the default for all future products, not just Fedora. The disk label would be unified across all systems with all disk sizes.

jkonecny12 (approved):
Looks good to me and I'm approving this. Thanks!

poncovka:
Updated and tested.

AdamWill:
This PR arguably does rather a lot more than was actually requested, and broke at least one thing. See https://bugzilla.redhat.com/show_bug.cgi?id=2092091#c6 and https://bugzilla.redhat.com/show_bug.cgi?id=2209760 . Why was this implemented this way, instead of by just tweaking the existing logic in blivet?

poncovka:
> This PR arguably does rather a lot more than was actually requested, and broke at least one thing. See https://bugzilla.redhat.com/show_bug.cgi?id=2092091#c6 and https://bugzilla.redhat.com/show_bug.cgi?id=2209760 . Why was this implemented this way, instead of by just tweaking the existing logic in blivet?

@AdamWill I don't remember all details, but this was the original proposal https://github.com/rhinstaller/anaconda/pull/4104, so we just went from there. It looks like @vojtechtrefny actually suggested to change the order of disk labels in Blivet at https://github.com/rhinstaller/anaconda/pull/4104#issuecomment-1127254303, but we never followed up on that. I don't know why.

Anyway, it should be pretty straightforward to fix this. If Blivet changes the order of disk labels, Anaconda can re-enable Blivet's defaults in the Anaconda configuration file via the `disk_label_type` option. @AdamWill @vojtechtrefny Would it make sense now?

PR #4233, opened 2022-07-25
Initialize empty disks on the Manual Partitioning screen

poncovka:
If a user removes of all partitions of a disk, the disk should be initialized
and the installer should create a new disk label on it.

**Note:** This is a quick portable fix. The upstream code would deserve an additional clean up.

Related: rhbz#2094977

### DIFF
--- a/pyanaconda/modules/storage/disk_initialization/configuration.py
+++ b/pyanaconda/modules/storage/disk_initialization/configuration.py
@@ -20,8 +20,11 @@
 import parted
 from blivet.devices import PartitionDevice
 
+from pyanaconda.anaconda_loggers import get_module_logger
 from pyanaconda.core.constants import CLEAR_PARTITIONS_DEFAULT, CLEAR_PARTITIONS_NONE, \
     CLEAR_PARTITIONS_LINUX, CLEAR_PARTITIONS_ALL, CLEAR_PARTITIONS_LIST
+
+log = get_module_logger(__name__)
 
 _all__ = ["DiskInitializationConfig"]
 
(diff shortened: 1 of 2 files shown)


pyanaconda/modules/storage/partitioning/interactive/utils.py
```diff
@@ -1090,6 +1090,7 @@ def _destroy_device(storage, device):
     # is preexisting, reinitialize the disk.
```
rvykydal, line 1090: This is unrelated, but I just can't figure out how we know that it is 'the last' partition as the comment says.

rvykydal (approved):
Looks good to me.

pyanaconda/modules/storage/partitioning/interactive/utils.py
poncovka: That magic is done by `config.can_initialize(storage, device.disk)` where we check if the disk is empty. That's not great, but it is one of the places in the installer I am afraid to touch as long as it works.

pyanaconda/modules/storage/disk_initialization/configuration.py
```diff
@@ -122,16 +125,22 @@ def can_initialize(self, storage, disk):
...
+        log.debug("Can %s be initialized?", disk.name)
+
         # Skip protected and readonly disks.
         if disk.protected:
+            log.debug("A protected disk cannot be initialized.")
             return False
 
         # Initialize disks with unrecognized formatting.
         if self.format_unrecognized and disk.format.type is None:
+            log.debug("A disk with unrecognized formatting can be initialized.")
```
M4rtinK, line 137: Just wondering (I'm afraid I don see deeply enough into use cases for this) but is this gated by user selecting the disk for installation, right ? 

Just thinking that otherwise there could be a partially corrupted disk connected to the machine or just a disk with totally custom data (IIRC some encryption systems don't have any headers on media on purpose) being written to, with possibly destructive results.

M4rtinK (approved):
Looks good to me, just asking about unrecognised formats. Thanks! :)

pyanaconda/modules/storage/disk_initialization/configuration.py
poncovka: > Just wondering (I'm afraid I don see deeply enough into use cases for this) but is this gated by user selecting the disk for installation, right ?

Of course.

VladimirSlavik (approved):
Looks good to me. Thank you!

PR #4235, opened 2022-07-26
Fix: check that the password contains the username

DustMerlin:


OndrejZobal (changes requested):
Thank you very much @DustMerlin!

The code looks good to me, but would you please add a closer description to your commit message?
You can use this:

```
 - Fix the root user creation GTK GUI not checking if a password
   contains the user name.
 - Fix the non-root user creation Text UI not checking if a password
   contains the user name.
```

Also please let me know if I misunderstood your changes.

OndrejZobal:
Sorry for the misunderstanding @DustMerlin. I meant for you to add my description after your original title.

The full commit message should look like this:
```
Fix: check that the password contains the username

 - Fix the root user creation GTK GUI not checking if a password
   contains the user name.
 - Fix the non-root user creation Text UI not checking if a password
   contains the user name.
```

OndrejZobal (approved):
Everything looks good to me. Thank you!

VladimirSlavik (approved):
Looks good to me. Thank you!

poncovka (approved):
Thanks!

PR #4237, opened 2022-07-26
Hide the keyboard layout indicator in the passphrase dialog (#2070823)

poncovka:
If the keyboard configuration is not supported in the current installation
environment, hide the keyboard layout indicator in the passphrase dialog.

Resolves: rhbz#2070823
Related: rhbz#2016613

### DIFF
--- a/pyanaconda/ui/gui/spokes/lib/passphrase.glade
+++ b/pyanaconda/ui/gui/spokes/lib/passphrase.glade
@@ -194,7 +194,7 @@
               </packing>
             </child>
             <child>
-              <object class="AnacondaLayoutIndicator" id="AnacondaLayoutIndicator1">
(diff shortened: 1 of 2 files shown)


rvykydal (approved):
Looks good to me.

M4rtinK (approved):
Looks good to me, thanks! :)

BTW, I think we could note this in release notes for the next Fedora or something like that, so that users are more aware that this is really not possible ?

VladimirSlavik (approved):
Looks good to me. Thank you!

poncovka:
> BTW, I think we could note this in release notes for the next Fedora or something like that, so that users are more aware that this is really not possible ?

I don't think it is necessary. This change should have been part of https://github.com/rhinstaller/anaconda/pull/3912 that was properly announced. It is just a bug now.

PR #4241, opened 2022-07-28
Fix the missing npm package (#infra)

poncovka:
The current solution worked, because the `npm` package is a weak dependency
of the `nodejs` package. Now it is broken due to some inconsistencies in the
Rawhide compose. It means that `npm` cannot be installed, but the CI doesn't
fail until it actually tries to use it. Let's install `npm` instead of `nodejs`.
The `nodejs` package will be installed as its required dependency.

As a temporary fix, include some missing packages into the installation.
The Rawhide compose contains incompatible versions.

### DIFF
--- a/dockerfile/anaconda-ci/Dockerfile
+++ b/dockerfile/anaconda-ci/Dockerfile
@@ -62,6 +62,9 @@ RUN set -ex; \
   dnf install -y \
   /usr/bin/xargs \
   npm \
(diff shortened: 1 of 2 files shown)


rvykydal (approved):
Thank you.

poncovka:
@M4rtinK I would like to merge this. Could you fix the Web UI tests in a separate pull request, please?

VladimirSlavik (approved):
Looks good to me. Thank you!

PR #4242, opened 2022-08-01
Web UI: Replace a newly translated string in tests

poncovka:
There is a new translation that breaks the Web UI tests.


VladimirSlavik (approved):
Thanks!

M4rtinK (approved):

PR #4196, opened 2022-06-14
Fix ELN containers somewhat

VladimirSlavik:
This fixes a large part of the ELN container builds.

Unfortunately, there are still missing `BuildRequires` so it can't be tested completely: `Error: Unable to find a match: libtimezonemap-devel >= 0.4.1-2`.

PR #4243, opened 2022-08-02
Disable ELN containers (#infra)

VladimirSlavik:
~~ELN is too disorderly at the moment.~~

This does not prevent building the containers locally, just turns off building them daily and failing.every.single.time.

### DIFF
--- a/.github/workflows/container-autoupdate.yml
+++ b/.github/workflows/container-autoupdate.yml
@@ -18,14 +18,14 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        container-tag: ['master', 'eln'] #, 'f36-devel', 'f36-release']
+        container-tag: ['master'] #, 'eln', 'f36-devel', 'f36-release']
         container-type: ['ci', 'rpm']
         include:
           - container-tag: master
             branch: master
-          - container-tag: eln
-            branch: master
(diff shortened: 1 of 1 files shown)


poncovka (approved):

sgallagher:
FWIW, the container builds were fixed yesterday.

VladimirSlavik:
So, after #4196 was merged, ELN containers build, but they don't pass the subsequent tests, because:
- Unit tests run also `pylint` which is broken due to `dill` not working on Python 3.11. These are patched in Rawhide, but don't appear in ELN.
- RPM tests blindly install `anaconda*`, which includes `anaconda-live`, which is not considered in ELN content resolver inputs. It's an open question if it should be included too or not. What that means is that `xisxwayland` is not pulled into ELN, but we don't special-case the tests for that, only the container.

If we don't want to merge this, we need to come up with some resolution to the issues above.

I considered separating the container builds into a reusable workflow and running one job with a matrix for Fedora and another for ELN, but GH actions still does not support calling reusable workflows within a matrix/strategy.

VladimirSlavik:
With #4248, we will be able to split off ELN builds to a separate workflow and have a separate status for that. That would make this change unnecessary.

VladimirSlavik:
With #4254, we will also be able to reduce the friction surface for RHEL and ELN.

I think this is no longer needed.

PR #4247, opened 2022-08-03
Fix growing installation size requirement

OndrejZobal:
Fix issue where the additional space needed for installation (shown
in the hub if the root partition didn't have enough space) would
slightly increase as time went on.

This happened because the size was calculated from the running
file system and included growing log files and other changes done in the
live session.

Delete old method for calculating installation size that uses running
file system and replace it with one that uses the payload file size.
Also modified tests accordingly.

### DIFF
--- a/pyanaconda/modules/payloads/source/live_os/initialization.py
+++ b/pyanaconda/modules/payloads/source/live_os/initialization.py
@@ -19,6 +19,8 @@ import os
 import stat
 import blivet.util
 
+from collections import namedtuple
+from blivet.size import Size
 from pyanaconda.anaconda_loggers import get_module_logger
 from pyanaconda.core.util import execWithCapture
 from pyanaconda.modules.common.task import Task
@@ -29,6 +31,8 @@ from pyanaconda.modules.common.structures.storage import DeviceData
 from pyanaconda.modules.payloads.source.mount_tasks import SetUpMountTask
 
 log = get_module_logger(__name__)
+
+SetupLiveOSResult = namedtuple("SetupLiveOSResult", ["required_space"])
 
 
 class DetectLiveOSImageTask(Task):
@@ -95,6 +99,20 @@ class SetUpLiveOSSourceTask(SetUpMountTask):
         """
         super().__init__(target_mount)
         self._image_path = image_path
+
+    def run(self):
+        """Run the task."""
+        super().run()
+
+        required_space = self._calculate_required_space()
+        return SetupLiveOSResult(required_space=required_space)
+
+    def _calculate_required_space(self):
+        """Calculate size of the live image image."""
+        source = os.statvfs(self._target_mount)
+        required_space = source.f_frsize * (source.f_blocks - source.f_bfree)
+        log.debug("Required space: %s", Size(required_space))
+        return required_space
 
     @property
     def name(self):
--- a/pyanaconda/modules/payloads/source/live_os/live_os.py
+++ b/pyanaconda/modules/payloads/source/live_os/live_os.py
@@ -19,13 +19,12 @@
 #
 from pyanaconda.core.i18n import _
 from pyanaconda.core.signal import Signal
-from pyanaconda.modules.payloads.base.utils import get_dir_size
 from pyanaconda.modules.payloads.constants import SourceType, SourceState
 from pyanaconda.modules.payloads.source.mount_tasks import TearDownMountTask
 from pyanaconda.modules.payloads.source.source_base import PayloadSourceBase, MountingSourceMixin
 from pyanaconda.modules.payloads.source.live_os.live_os_interface import LiveOSSourceInterface
 from pyanaconda.modules.payloads.source.live_os.initialization import SetUpLiveOSSourceTask, \
-    DetectLiveOSImageTask
+    SetupLiveOSResult, DetectLiveOSImageTask
 
 from pyanaconda.anaconda_loggers import get_module_logger
 log = get_module_logger(__name__)
@@ -38,6 +37,7 @@ class LiveOSSourceModule(PayloadSourceBase, MountingSourceMixin):
         super().__init__()
         self._image_path = ""
         self.image_path_changed = Signal()
+        self._required_space = 0
 
     def for_publication(self):
         """Return a DBus representation."""
@@ -68,8 +68,7 @@ class LiveOSSourceModule(PayloadSourceBase, MountingSourceMixin):
         :return: required size in bytes
         :rtype: int
         """
-        # FIXME: Unify the required space with the installation size.
-        return get_dir_size("/") * 1024
+        return self._required_space
 
     @property
     def image_path(self):
@@ -114,12 +113,17 @@ class LiveOSSourceModule(PayloadSourceBase, MountingSourceMixin):
         :return: list of tasks required for the source setup
         :rtype: [Task]
         """
-        return [
-            SetUpLiveOSSourceTask(
-                image_path=self.image_path,
-                target_mount=self.mount_point
-            )
-        ]
+        task = SetUpLiveOSSourceTask(
+            image_path=self.image_path,
+            target_mount=self.mount_point
+        )
+
+        handler = self._handle_live_os_task_result
+        task.succeeded_signal.connect(lambda: handler(task.get_result()))
+        return [task]
+
+    def _handle_live_os_task_result(self, result: SetupLiveOSResult):
+        self._required_space = result.required_space
 
     def tear_down_with_tasks(self):
         """Tear down the installation source.
--- a/tests/unit_tests/pyanaconda_tests/modules/payloads/source/test_source_live_os.py
+++ b/tests/unit_tests/pyanaconda_tests/modules/payloads/source/test_source_live_os.py
@@ -30,7 +30,7 @@ from pyanaconda.modules.payloads.constants import SourceState
 from pyanaconda.modules.payloads.source.live_os.live_os import LiveOSSourceModule
 from pyanaconda.modules.payloads.source.live_os.live_os_interface import LiveOSSourceInterface
 from pyanaconda.modules.payloads.source.live_os.initialization import SetUpLiveOSSourceTask, \
-    DetectLiveOSImageTask
+    DetectLiveOSImageTask, SetupLiveOSResult
 from pyanaconda.modules.payloads.source.mount_tasks import TearDownMountTask
 
 from tests.unit_tests.pyanaconda_tests import patch_dbus_get_proxy, patch_dbus_publish_object, \
@@ -82,9 +82,19 @@ class LiveOSSourceTestCase(unittest.TestCase):
         """Test the network_required property."""
         assert self.module.network_required is False
 
-    def test_required_space(self):
+    @patch.object(SetUpLiveOSSourceTask, "run")
+    def test_required_space(self, runner):
         """Test the required_space property."""
-        assert self.module.required_space > 0
+        assert self.module.required_space == 0
+
+        runner.return_value = SetupLiveOSResult(12345)
+
+        tasks = self.module.set_up_with_tasks()
+        for task in tasks:
+            task.run_with_signals()
+
+        runner.assert_called_once_with()
+        assert self.module.required_space == 12345
 
     @patch("os.path.ismount")
     def test_get_state(self, ismount_mock):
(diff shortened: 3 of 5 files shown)


pyanaconda/modules/payloads/source/live_os/initialization.py
```diff
@@ -96,6 +100,22 @@ def __init__(self, image_path, target_mount):
...
 
+    def run(self):
+        """Run the task.
+
+        Calculates size of the live image image.
+        """
+        super().run()
+
+        required_space = self._calculate_required_space()
+        log.debug("Size of live OS image payload: %s.", Size(required_space))
```
poncovka, line 111: This can say just `Required space: `.  In the logs, it should be clear that this is a Live OS image from the context. Please, remove the dot after the size and move the logging to the `_calculate_required_space` method.

pyanaconda/modules/payloads/source/live_os/initialization.py
```diff
@@ -96,6 +100,22 @@ def __init__(self, image_path, target_mount):
...
+
+        return SetupLiveOSResult(required_space=required_space)
+
+    def _calculate_required_space(self):
```
poncovka, line 115: Please, add a docstring here.

pyanaconda/modules/payloads/source/live_os/live_os.py
```diff
@@ -114,12 +113,17 @@ def set_up_with_tasks(self):
...
+        task = SetUpLiveOSSourceTask(
+            image_path=self.image_path,
+            target_mount=self.mount_point
+        )
+
+        handler = self._handle_live_os_task_result
+        task.succeeded_signal.connect(lambda: handler(task.get_result()))
+        return [task]
+
+    def _handle_live_os_task_result(self, result: SetupLiveOSResult):
```
poncovka, line 125: Please, add a docstring here.

tests/unit_tests/pyanaconda_tests/modules/payloads/source/test_source_live_os.py
```diff
@@ -82,9 +82,14 @@ def test_network_required(self):
         """Test the network_required property."""
         assert self.module.network_required is False
 
-    def test_required_space(self):
+    @patch("pyanaconda.modules.payloads.source.live_os.initialization"
+           ".SetUpLiveOSSourceTask.run")
```
poncovka, line 86: I think this patch is not necessary.

tests/unit_tests/pyanaconda_tests/modules/payloads/source/test_source_live_os.py
```diff
@@ -82,9 +82,14 @@ def test_network_required(self):
...
+    def test_required_space(self, task_run):
         """Test the required_space property."""
-        assert self.module.required_space > 0
+        assert self.module.required_space == 0
+
+        self.module._required_space = 12345
+        assert self.module.required_space == 12345
```
poncovka, line 92: Please, keep here only the test for zero and add a test similar to `test_handle_setup_task_result` to test other values.

tests/unit_tests/pyanaconda_tests/modules/payloads/source/test_source_live_os.py
```diff
@@ -118,6 +123,22 @@ def test_repr(self):
...
+        statvfs_result.f_bfree = 42
+        statvfs.return_value = statvfs_result
+
+        task = SetUpLiveOSSourceTask(
+                "/path/to/base/image",
+                "/path/to/mount/source/image"
+        )
+
+        assert task._calculate_required_space() == 29696
+
```
poncovka, line 141: Please, remove the extra line.

tests/unit_tests/pyanaconda_tests/modules/payloads/source/test_source_live_os.py
```diff
@@ -178,8 +199,12 @@ def test_detect_live_os_image(self, os_stat_mock, stat_mock, exists_mock, exec_m
         detected_image = task.run()
         assert detected_image == "/my/device"
 
-    def test_setup_install_source_task_name(self):
+    @patch("pyanaconda.modules.payloads.source.live_os.initialization.SetUpLiveOSSourceTask"
+           "._calculate_required_space")
```
poncovka, line 203: I think you can use `@patch.object(SetUpLiveOSSourceTask, "_calculate_required_space", return_value=0)` to make it simpler. Then you don't have to call `required_space.return_value = 0`.

tests/unit_tests/pyanaconda_tests/modules/payloads/source/test_source_live_os.py
```diff
@@ -189,14 +214,18 @@ def test_setup_install_source_task_name(self):
 
     @patch("pyanaconda.modules.payloads.source.live_os.initialization.blivet.util.mount")
     @patch("pyanaconda.modules.payloads.source.live_os.initialization.stat")
+    @patch("pyanaconda.modules.payloads.source.live_os.initialization.SetUpLiveOSSourceTask"
+           "._calculate_required_space")
     @patch("os.stat")
     @patch_dbus_get_proxy
-    def test_setup_install_source_task_run(self, proxy_getter, os_stat, stat, mount):
+    def test_setup_install_source_task_run(self, proxy_getter, os_stat, required_space,
```
poncovka, line 221: In this test, you should check the return value of `SetUpLiveOSSourceTask(...).run()`. Make sure it is an instance of `SetupLiveOSResult` and the required space is set to the expected value.

tests/unit_tests/pyanaconda_tests/modules/payloads/source/test_source_live_os.py
poncovka: Maybe use a different size than zero in these tests. You are trying to simulate the real usage.

tests/unit_tests/pyanaconda_tests/modules/payloads/source/test_source_live_os.py
```diff
@@ -259,14 +295,18 @@ def test_setup_install_source_task_invalid_block_dev(self, proxy_getter, os_stat
 
     @patch("pyanaconda.modules.payloads.source.live_os.initialization.blivet.util.mount")
     @patch("pyanaconda.modules.payloads.source.live_os.initialization.stat")
+    @patch("pyanaconda.modules.payloads.source.live_os.initialization.SetUpLiveOSSourceTask"
+           "._calculate_required_space")
     @patch("os.stat")
     @patch_dbus_get_proxy
-    def test_setup_install_source_task_failed_to_mount(self, proxy_getter, os_stat, stat, mount):
+    def test_setup_install_source_task_failed_to_mount(self, proxy_getter, os_stat,
```
poncovka, line 302: By the way, the max line length of the Anaconda code is 99. That's not great, but the consistency has a priority. I think you are using a different one. Please, check the configuration of your IDE.

pyanaconda/modules/payloads/source/live_os/initialization.py
```diff
@@ -96,6 +100,22 @@ def __init__(self, image_path, target_mount):
...
+        source = os.statvfs(self._target_mount)
+        return source.f_frsize * (source.f_blocks - source.f_bfree)
```
poncovka, line 117: Have you compared the results with the previous solution? Does it return a similar size?

poncovka (commented):
I have some suggestions for improvements, but in general the code looks good to me. Thanks!

pyanaconda/modules/payloads/source/live_os/initialization.py
OndrejZobal: Before the size estimate started approximately 440 MiB higher than than the new one and just by opening the storage spoke two more times with nearly no waiting it can grow another 80 MiB. 
I tried using this patch with one of rawhide ISOs and it installed fine.

tests/unit_tests/pyanaconda_tests/modules/payloads/source/test_source_live_os.py
```diff
@@ -118,6 +128,21 @@ def test_repr(self):
...
 class LiveOSSourceTasksTestCase(unittest.TestCase):
     """Test the tasks of the Live OS source."""
+    @patch("pyanaconda.modules.payloads.source.live_os.initialization.os.statvfs")
+    def test_live_os_image_size(self, statvfs):
+        """Test Live OS image size calculation."""
+        statvfs_result = Mock()
+        statvfs_result.f_frsize = 512
+        statvfs_result.f_blocks = 100
+        statvfs_result.f_bfree = 42
+        statvfs.return_value = statvfs_result
```
VladimirSlavik, line 138: Not sure if you like that better, but you can make this into one line:

`statvfs.return_value = Mock(f_frsize=512, f_blocks=100, f_bfree=42)`

OndrejZobal: Yes, this is much better, thanks.

tests/unit_tests/pyanaconda_tests/modules/payloads/source/test_source_live_os.py
```diff
@@ -206,17 +230,23 @@ def test_setup_install_source_task_run(self, proxy_getter, os_stat, stat, mount)
...
         ).run()
 
+        assert isinstance(result, SetupLiveOSResult)
+        assert result.required_space == 12345
+
         device_tree.ResolveDevice.assert_called_once_with("/path/to/base/image")
         os_stat.assert_called_once_with("/resolved/path/to/base/image")
 
+    @patch("pyanaconda.modules.payloads.source.live_os.initialization.SetUpLiveOSSourceTask"
+           "._calculate_required_space")
```
poncovka, line 245: Please, replace this patch in other tests as well.

OndrejZobal: I think I got them all this time, sorry about that.

poncovka (approved):
Thank you!

jkonecny12 (approved):
Needs to be tested appropriately. It's great simplification, I like it.

VladimirSlavik (approved):
Looks good to me. Thank you!

OndrejZobal:
> Needs to be tested appropriately. It's great simplification, I like it.

@jkonecny12 I am not sure how to test this better. When I tried going through with an installation to a drive of the same size as the estimate and everything went fine.

VladimirSlavik:
I think the testing needs to verify that there was a bug, and that the change fixes it. So, you would:
* start an unpatched live media on a system with tons of memory,
* verify that it calculates estimate of X,
* close anaconda,
* write a file large enough to show in the estimation,
* start anaconda again,
* verify that the estimate is now X+F,
* repeat same steps with patched anaconda and see that the estimate does *not* change.

That would be the 100% answer. Also, welcome to QE! Just kidding, of course real QE would set the bar yet higher and test also that this does not break any other case where space calculation is important...

VladimirSlavik:
There is a script called something like `apply-updates.sh`, normally it runs somewhere deep before actual anaconda python code, but you can run it manually with an url to updates image, same as you would use `inst.updates=`. The reason to do this is that live media does not run this script automatically, unlike the boot.iso kind, so the boot option does not work.

OndrejZobal:
> There is a script called something like `apply-updates.sh`, normally it runs somewhere deep before actual anaconda python code, but you can run it manually with an url to updates image, same as you would use `inst.updates=`. The reason to do this is that live media does not run this script automatically, unlike the boot.iso kind, so the boot option does not work.

Weird from my experience the `inst.updates` boot option does work on live ISOs but only if it's among the first arguments specified otherwise it gets ignored.

poncovka:
> Tested what @VladimirSlavik suggested and it proved the bug, so if nobody has any other suggestions I will go ahead and merge this.

Let's merge it! :-)

poncovka:
> > There is a script called something like `apply-updates.sh`, normally it runs somewhere deep before actual anaconda python code, but you can run it manually with an url to updates image, same as you would use `inst.updates=`. The reason to do this is that live media does not run this script automatically, unlike the boot.iso kind, so the boot option does not work.
> 
> Weird from my experience the `inst.updates` boot option does work on live ISOs but only if it's among the first arguments specified otherwise it gets ignored.

It is easier to run the `apply-updates` script, but `inst.updates` works as well with Live media.

PR #4248, opened 2022-08-03
Container rebuilds as a reusable workflow (#infra)

VladimirSlavik:
This reorganizes the code but it should stay the same.

This enables the following:
- Read the whole thing with out having to mentally parse the matrix.
- Enable and disable the `fNN-{devel|release}` containers with one line per each, no repeated commenting and uncommenting parts of the file.
- The top-level single workflow can be split into multiple ones, eg. separate ELN so we know the failures are of a different kind.
- If the templated branching idea is implemented, generating the file in this format should be far easier.

### DIFF
--- a/.github/workflows/container-autoupdate.yml
+++ b/.github/workflows/container-autoupdate.yml
@@ -9,73 +9,34 @@ permissions:
   contents: read
 
 jobs:
-  refresh-containers:
-    name: Refresh anaconda containers
-    runs-on: ubuntu-20.04
-    environment: quay.io
-    # we need to have matrix to cover all branches because schedule is run only on default branch
-    # we can add here fedora branches (e.g. f33-devel) to cover their support when needed
-    strategy:
-      fail-fast: false
-      matrix:
-        container-tag: ['master', 'eln'] #, 'f36-devel', 'f36-release']
-        container-type: ['ci', 'rpm']
-        include:
-          - container-tag: master
-            branch: master
-          - container-tag: eln
-            branch: master
-            base-container: 'quay.io/fedoraci/fedora:eln-x86_64'
-          #- container-tag: f36-devel
-          #  branch: f36-devel
-          #- container-tag: f36-release
-          #  branch: f36-release
-    env:
-      CI_TAG: '${{ matrix.container-tag }}'
-    timeout-minutes: 60
-    steps:
-      - name: Checkout anaconda repository
-        uses: actions/checkout@v3
-        with:
-          ref: ${{ matrix.branch }}
 
-      - name: Build anaconda-${{ matrix.container-type }} container
-        run: |
-          BASE_CONTAINER=${{ matrix.base-container }}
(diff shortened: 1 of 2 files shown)


jkonecny12 (approved):
I love this change. Thanks for the improvement!

.github/workflows/container-reusable.yml
```diff
@@ -0,0 +1,82 @@
...
+      base-container:
+        required: false
+        type: string
+#    secrets:
+#      token:
+#        required: true
+#      QUAY_USERNAME:
+#        required: true
+#      QUAY_PASSWORD:
+#        required: true
```
VladimirSlavik, line 21: I could find no way of verifying that these are passed with the 'inherit' keyword, so I'll leave these here until we know how to approach that:
- explicitly pass limited amount of secrets = uncomment here, add to caller
- believe it works fine, delete this block, and merge
- something else?

jkonecny12: No issue with that but please add a comment explaining why the commented code is there.

.github/workflows/container-reusable.yml
```diff
@@ -0,0 +1,82 @@
+name: Refresh container image
```
poncovka, line 1: This is basically an action, isn't it? Maybe we could keep actions in `.github/actions/` or something like that and give this file a normal name. The `container-reusable` name doesn't say anything useful.

VladimirSlavik: If you have some better idea for the file name, that's certainly doable. How about `rebuild-container-action-reusable.yml`?

For the `name:` part, that is reused in the UI, so I don't want to make it too long, or we won't know what the individual runs are. In fact, it would be good to make it even shorter, the auto-suffix ` (rpm)` is truncated already on the left side job list.

As to another directory, unfortunately the file [must be exactly where it is](https://docs.github.com/en/actions/using-workflows/reusing-workflows#creating-a-reusable-workflow).

Rant mode on: The terminology and capability mix is pretty weird. Normally, you have a _workflow_ which consists of _jobs_ which have _steps_. A step can be either an _action_ call, or a bash script. The files you put into `.github/workflows` define whole workflows, each complete with its jobs and steps, no preprocessing, no includes, no hierarchy. An action must be a separate repo, which provides some separation, but it does not feel made to be a separation measure, it mostly enables easy consumption of third party actions. A reusable workflow is a bit of a hack within that system, it's an almost-workflow, which does not run on its own but is invoked as a job (IIRC). As a corollary, it's a job, not a step, which means you can't combine it with the bash kind of step, you have to create a chain of jobs with dependencies. And you can't use matrix/strategy in workflows that call reusable workflows, only inside the reusable workflows themselves... it's all just lovely.

.github/workflows/container-reusable.yml
VladimirSlavik: Removed the commented code, added comment instead.

.github/workflows/container-reusable.yml
VladimirSlavik: Now it's `container-rebuild-action.yml`...

M4rtinK (approved):
Also looks good to me - much less noise! :)

poncovka (approved):
Thank you!

PR #4249, opened 2022-08-04
Patch pylint on ELN too (#infra)

VladimirSlavik:
This should let unit tests pass and get us to 3/4 containers building and testing correctly.

Thanks @sgallagher for the packages!


VladimirSlavik:
Tested here: https://github.com/VladimirSlavik/anaconda/actions/runs/2796460738

ELN ci fails only on login to registry = tests fixed.

poncovka (approved):

PR #4250, opened 2022-08-04
Simplify the payload manager

poncovka:
* Remove unused code.
* Simplify the implementation of the payload manager.
* Move DNF code from the payload manager to the DNF payload.

**MERGE AFTER BRANCHING!**

### DIFF
--- a/pyanaconda/payload/manager.py
+++ b/pyanaconda/payload/manager.py
@@ -40,7 +40,6 @@ class PayloadState(IntEnum):
     STARTED = 0
     WAITING_STORAGE = 1
     WAITING_NETWORK = 2
-    VERIFYING_AVAILABILITY = 3
     DOWNLOADING_PKG_METADATA = 4
     DOWNLOADING_GROUP_METADATA = 5
     FINISHED = 6
@@ -206,7 +205,6 @@ class PayloadManager(object):
         # Test if any repository changed from the last update
         if only_on_change:
             log.debug("Testing repositories availability")
-            self._set_state(PayloadState.VERIFYING_AVAILABILITY)
             if payload.dnf_manager.verify_repomd_hashes():
                 log.debug("Payload isn't restarted, repositories are still available.")
                 self._set_state(PayloadState.FINISHED)
--- a/pyanaconda/payload/manager.py
+++ b/pyanaconda/payload/manager.py
@@ -46,7 +46,7 @@ class _PayloadManager(Runnable, ProgressReporter):
     The payload thread data should be accessed using the payloadMgr object,
     and the running thread can be accessed using threadMgr with the
     THREAD_PAYLOAD constant, if you need to wait for it or something. The
-    thread should be started using payloadMgr.restart_thread.
+    thread should be started using payloadMgr.start.
     """
 
     @property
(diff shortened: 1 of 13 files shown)


VladimirSlavik (approved):
Thank you, looks good to me.

pyanaconda/payload/manager.py
```diff
@@ -60,10 +59,7 @@ class PayloadManager(object):
 
     All states except ERROR are expected to happen linearly, and adding
     a listener for a state that has already been reached or passed will
-    immediately trigger that listener. For example, if the payload thread is
-    currently in DOWNLOADING_GROUP_METADATA, adding a listener for
-    WAITING_NETWORK will immediately run the code being added
-    for WAITING_NETWORK.
+    immediately trigger that listener.
```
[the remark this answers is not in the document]
poncovka: All these special states were removed, because we don't really need them. The new implementation uses signals for monitoring the stages of the task and progress reporting for showing progress messages about the running task. That's all.

pyanaconda/ui/lib/subscription.py
```diff
@@ -86,7 +86,7 @@ def _do_payload_restart(payload):
     :param payload: Anaconda payload instance
     """
     # restart payload
-    payloadMgr.restart_thread(payload,
+    payloadMgr.start(payload,
```
jkonecny12, line 89: Seems that you broke parameter indentation by the rename here.

pyanaconda/payload/manager.py
```diff
@@ -16,152 +15,131 @@
...
         this method's return is not blocked by waiting on the previous payload
         thread. If there is already a payload thread restart pending, this method
         has no effect.
 
-        :param payload.Payload payload: The payload instance
-        :param bool fallback: Whether to fall back to the default repo in case of error
-        :param bool try_media: Whether to check for valid mounted media
-        :param bool only_on_change: Restart thread only if existing repositories changed.
-            This won't restart thread even when a new repository was added!!
+        This won't restart thread even when a new repository was added!
```
jkonecny12, line 70: I wonder this comment when two lines below is ``log.debug("Restarting payload thread")``.

pyanaconda/payload/manager.py
```diff
@@ -16,152 +15,131 @@
...
-        assert isinstance(event_id, PayloadState)
+    @property
+    def steps(self):
+        """Total number of steps."""
+        return 1
 
-        # Add the listener inside the lock in case we need to run immediately,
-        # to make sure the listener isn't triggered twice
-        with self._event_lock:
-            self._event_listeners[event_id].append(func)
```
jkonecny12, line 105: Just to be sure, we don't need this because of signals?

jkonecny12 (commented):
Looks good to me otherwise.

pyanaconda/payload/manager.py
poncovka: It is a leftover. I will remove it.

pyanaconda/ui/lib/subscription.py
poncovka: Fixed.

jkonecny12 (approved):
Definitely less confusing to me now. Thanks!

PR #4251, opened 2022-08-05
Add the `nitems` function

poncovka:
Introduce a function that returns exactly `n` items of an iterable object.


VladimirSlavik (approved):
Looks good to me, thank you.

I wonder if this is needed, the code actually works as of now. The new version is somewhat easier to read, though.

poncovka:
> I wonder if this is needed, the code actually works as of now. The new version is somewhat easier to read, though.

I promised to simplify the implementation at https://github.com/rhinstaller/anaconda/pull/3981#discussion_r837335867.

pyanaconda/core/string.py
```diff
@@ -131,3 +131,25 @@ def have_word_match(str1, str2):
     str2 = str2.lower()
 
     return all(word in str2 for word in str1_words)
+
+
+def split_in_two(value, delimiter=None):
+    """Split the given value into two strings.
```
jkonecny12, line 137: I would change the `value` as `string` or does it support something else than a string?

pyanaconda/core/string.py
```diff
@@ -131,3 +131,25 @@ def have_word_match(str1, str2):
...
+    This function is useful for safe tuple unpacking.
+    The functionality is similar to str.partition(),
+    but it supports the delimiter of str.split().
+
+    For example:
+
+        first, second = split_in_two(text)
+
+    :param value: a string to split
+    :param delimiter: a delimiter for splitting
```
jkonecny12, line 148: I'm missing information what will happen in case of `None`. Could point to the documentation.

jkonecny12 (approved):
I know it's in the `string.py` but still I think it would be better to change `value` to `string` to explicitly express that in the doc.

Anyway, even without that, looks good to me.

pyanaconda/core/string.py
VladimirSlavik: +1, actually. I didn't like the magic `None` part, but given that it's a wrapper around `string.split`, I decided it's my fault I don't know the behavior well enough.

pyanaconda/core/string.py
poncovka: Fixed.

pyanaconda/core/string.py
poncovka: Fixed.

jkonecny12 (approved):
LGTM

PR #4254, opened 2022-08-08
Do not provide the anaconda-live subpackage on RHEL

VladimirSlavik:
This should work, even if I have no idea what I'm doing...

Tested locally with rpm tests for both negated and non-negated condition.

### DIFF
--- a/anaconda.spec.in
+++ b/anaconda.spec.in
@@ -156,6 +156,9 @@ Obsoletes: booty <= 0.107-1
 The anaconda-core package contains the program which was used to install your
 system.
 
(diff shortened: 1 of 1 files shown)


jkonecny12 (approved):
Looks good to me. I just wonder about the `live-env-deps` meta-package.

I guess we probably want to leave it here because of InitialSetup?

VladimirSlavik:
Such a package does not exist. The deps metapackages we have are both used on boot.iso, so these remain.

jkonecny12:
Oh my, I can't read today. My mistake.

PR #4258, opened 2022-08-09
Improve Register button click feedback

M4rtinK:
Set the phase right after button click to avoid the registration status being outdated.

So far that has been handled by the registration thread, which could
result in the status message being out of date until the thread gets to
the status update.

Resolves: rhbz#2074636


poncovka (approved):

VladimirSlavik (approved):
Thanks!

PR #4259, opened 2022-08-09
Remove unused code

poncovka:
Remove code that we no longer use.


VladimirSlavik (approved):
Please check with @M4rtinK that the sensitive logger is not used in the still-unported registration code. Otherwise LGTM, thank you!

jkonecny12 (approved):
LGTM!

jkonecny12:
> Please check with @M4rtinK that the sensitive logger is not used in the still-unported registration code. Otherwise LGTM, thank you!

Also did we checked this these changes won't break InitialSetup in some way?

poncovka:
> Please check with @M4rtinK that the sensitive logger is not used in the still-unported registration code.

I've checked the `rhel-9` branch. The sensitive logger is used only for geolocation.

> Also did we checked this these changes won't break InitialSetup in some way?

I've checked the initial setup code. It doesn't use any of these.

VladimirSlavik (approved):
Thank you!

PR #4260, opened 2022-08-09
Remove the SimpleConfigFile class

poncovka:
The `SimpleConfigFile` class is used only in the installation task that
modifies the SELinux configuration file. Change the implementation, so
that the `SimpleConfigFile` class can be removed.


pyanaconda/kexec.py
```diff
@@ -19,20 +19,25 @@
...
-from pyanaconda.anaconda_loggers import get_module_logger
 log = get_module_logger(__name__)
 
 
 class GrubbyInfoError(Exception):
     pass
 
 
+def unquote(s):
+    return ' '.join(shlex.split(s))
```
VladimirSlavik, line 38: You might want to have also tests for this?

VladimirSlavik, line 38: You might want to test this...

tests/unit_tests/pyanaconda_tests/test_simpleconfig.py
```diff
@@ -1,204 +0,0 @@
...
-        scf.set(('KEY3', 'value3'))
-        assert scf.get('KEY3') == 'value3'
-        scf.set(('key4', 'value4'))
-        assert scf.get('KEY4') == 'value4'
-
-    def test_unset(self):
-        scf = SimpleConfigFile()
-        scf.set(('key1', 'value1'))
-        scf.unset(('key1'))
-        assert scf.get('key1') == ''
```
VladimirSlavik, line 74: ...with this?

tests/unit_tests/pyanaconda_tests/modules/security/test_module_security.py
```diff
@@ -378,92 +381,129 @@ def test_authselect_requirements(self):
...
 
-            with open(os.path.join(sysroot, "etc/selinux/config")) as f:
-                assert f.read().strip() == content.strip()
+            content = self._get_configuration(sysroot)
+            assert "\nSELINUX=enforcing\n" in content
+            assert content == self.ENFORCING_SELINUX_CONFIGURATION
+
+
+class SecurityTasksTestCase(unittest.TestCase):
+    """Test the secusrity tasks."""
```
VladimirSlavik, line 501: typo

VladimirSlavik (commented):
I like dropping the class, as well as making the SELinux tests real. However, I think these are two very different things? Otherwise I'd approve already.

tests/unit_tests/pyanaconda_tests/test_simpleconfig.py
```diff
@@ -1,204 +0,0 @@
...
-            config.read()
-            assert config.get("ESSID") == "Example Network #1"
-            assert config.get("ESSID2") == "Network #2"
-            assert config.get("COMMENT") == "Save this string"
-            assert str(config) == self.TEST_CONFIG
-
-    def test_unquote(self):
-        assert simpleconfig.unquote("plain string") == "plain string"
-        assert simpleconfig.unquote('"double quote"') == "double quote"
-        assert simpleconfig.unquote("'single quote'") == "single quote"
```
VladimirSlavik, line 51: ...with this?

pyanaconda/modules/security/installation.py
```diff
@@ -208,13 +208,30 @@ def run(self):
...
+            # Write the modified configuration.
+            with open(path, "w") as f:
+                f.writelines(lines)
+
         except OSError as msg:
             log.error("SELinux configuration failed: %s", msg)
 
+    def _process_line(self, line):
+        """Process a line from the SELinux configuration file."""
+        if line.startswith("SELINUX="):
```
jkonecny12, line 229: I think we should do the `trim` first. It could have for any reason and the config will be still valid but we won't match.

Also I'm thinking if we don't want to have this logged to know that we changed something at all?

jkonecny12 (commented):
Just a small recommendation for the code.

However, the main question is, why we are doing this? Do we need this class to be removed or it's just optimization of the code? If the second one than I wonder if it is worth the while really.

poncovka:
> However, the main question is, why we are doing this? Do we need this class to be removed or it's just optimization of the code? If the second one than I wonder if it is worth the while really.

This class used to be necessary for storing settings of ifcfg configuration files, but we no longer work with them. It had other use cases that were dropped over the years. At this moment, there is no demand for such functionality. If we ever need it again, I would prefer to use a library like  [python3-configobj](https://configobj.readthedocs.io/en/latest/) anyway (it is available in RHEL 9 AppStream, I've checked). Therefore, I recommend to remove those unnecessary 400 lines of code from our code base.

M4rtinK:
> [python3-configobj](https://configobj.readthedocs.io/en/latest/)
Just want to note that I have been using this library in some of my hobby projects and have the best experience with it. :)

poncovka:
> I like dropping the class, as well as making the SELinux tests real. However, I think these are two very different things?

I have opened a new pull request at https://github.com/rhinstaller/anaconda/pull/4264 for the tests.

pyanaconda/kexec.py
poncovka: Fixed.

tests/unit_tests/pyanaconda_tests/modules/security/test_module_security.py
poncovka: Dropped.

tests/unit_tests/pyanaconda_tests/test_simpleconfig.py
poncovka: Fixed.

pyanaconda/modules/security/installation.py
poncovka: Fixed.

jkonecny12 (approved):

PR #4261, opened 2022-08-09
Set environment in reusable workflow (#infra)

VladimirSlavik:
This fixes the container builds and propagates secrets correctly.

Turns out, this was a very simple omission. However, reusable workflows declare things exactly opposite of function calls, so that was a lesson to learn first...

Successful run with mock values for the secrets - still fails on login but you can see the masked secrets `***` actually present in the login command, as opposed to simply missing before. https://github.com/VladimirSlavik/anaconda/actions/runs/2827874013

See also:
https://colinsalmcorner.com/consuming-environment-secrets-in-reusable-workflows/
https://github.com/orgs/community/discussions/25238#discussioncomment-3247035

### DIFF
--- a/.github/workflows/container-rebuild-action.yml
+++ b/.github/workflows/container-rebuild-action.yml
@@ -33,7 +33,8 @@ on:
       base-container:
         required: false
         type: string
-     # secrets are inherited, so not explicitly listed
+    # secrets consumed: QUAY_USERNAME, QUAY_PASSWORD
+    # these are provided by using the quay.io environment below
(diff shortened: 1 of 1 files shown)


jkonecny12 (approved):
Yes true, I encapsulated the secrets into the environment to avoid it's spreading everywhere. I completely forget about that...
Thanks for solving this. Feel free to break 24 hours rule to fix our container auto updates soon.

poncovka (approved):

PR #4263, opened 2022-08-10
Split container refresh: Fedora and ELN (#infra)

VladimirSlavik:
This provides cleaner statuses. If ELN fails, we know at a glance that the Fedora containers are still OK.


poncovka (approved):
Thank you!

jkonecny12 (approved):
Looks great to me. I wish that we don't need that, maybe in the future :)

PR #4265, opened 2022-08-11
Remove release builds from CI status page (#docs)

VladimirSlavik:
The "real" releases are checked "synchronously", as they are created, so no need to have them in the status for "asynchronous" checking. Further, this workflow does not run often, so a failure status will remain on the page for far longer than the actual impact of that failure. Finally, a failure of this workflow does not mean the release did not happen - it can be done manually, too.

The status that does belong on this page is the daily release rehearsal just below.

Clarify also what the page is for.

### DIFF
--- a/docs/ci-status.rst
+++ b/docs/ci-status.rst
@@ -1,7 +1,13 @@
 CI Status
 =========
 
-This page shows current status of CI jobs that are expected to be stable.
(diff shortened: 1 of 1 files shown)


jkonecny12 (approved):
LGTM

PR #4267, opened 2022-08-11
Check periodically age of composes on cobra02 (#infra)

VladimirSlavik:
What it says. Doing this would add a red/green thingy to the CI status page for PXE compose ages.

(Not tested yet.)

### DIFF
--- a/.github/workflows/check-cobra-trees-age.yml
+++ b/.github/workflows/check-cobra-trees-age.yml
@@ -1,0 +1,41 @@
+# Check age of trees on cobra[02]
+name: Build boot.iso
+on:
+  schedule:
+    - 30 22 * * *
+  workflow_dispatch:
+
+permissions: none
+
+jobs:
+  get-data:
+    runs-on: [self-hosted, kstest]
+    steps:
+      - name: Get directory listing from cobra02
+        id: get_dir_listing
+        run: |
+          curl http://cobra02.anaconda.englab.brq.redhat.com/trees/ -o trees.html
+          rg -e '>([^<]+)</a>[\s]+</td><td align="right">([0-9]{4}-[0-9]{2}-[0-9]{2} [0-9]{2}:[0-9]{2})' -o -r '$1 $2' trees.html > pairs.txt
+          echo "::set-output name=data::$(cat pairs.txt)"
+    outputs:
+      data: ${{ steps.get_dir_listing.outputs.data }}
+
+  check:
+    runs-on: ubuntu-latest
+    needs: get-data
+    strategy:
+      fail-fast: false
+      matrix:
+        dir: ["rawhide/", "rhel8/", "rhel9/"]
+    steps:
+      - name: Check age
+        run: |
+          echo ${{ needs.get-data.outputs.data }} | tee pairs.txt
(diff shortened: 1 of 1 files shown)


jkonecny12 (changes requested):
I don't agree with this solution.

First it's completely internal. Second it's not related to the project. I wouldn't have an issue to have this separate but I don't think it should be part of the anaconda repository.

poncovka:
@jkonecny12 What do you suggest as an alternative? If the result of this check is available on a page we never visit, it's the same as not having the check at all. I guess we could be notified by e-mail or something if that's possible.

jkonecny12:
> Agreed on this being a poor fit. It's having a hammer and everything is a nail. However it's also easy to do and provides real value, which is why I even tried this.
> 
> Maybe this uncovers something important?
> 
> _As somebody watching infrastructure daily, I need a single place to monitor state of infrastructure - _all_ of it including the internal bits - so that I can save time when deciding if things need mending._
> 
> The issue is that "I" act in multiple capacities at once:
> 
> 1. upstream developer

I don't agree that upstream developer (contributor) has any benefit of knowing a state of our internal proxy repositories...
 
> 2. upstream maintainer

Not sure if there is difference between upstream and downstream maintainer in our case. However, if it would be than again it would be someone who won't be able to use our internal development environment.

> 3. downstream developer
> 4. downstream maintainer
> 
> I am not sure how to separate these well. Every person in any of these categories is also in all of them. How do we separate caring for internal and external infra, if the same people use them at the same time?

Why do you think so? I don't think there is a way of upstream developers to be eve able to use our proxy repositories.
 
> Maybe the answer is building an internal "infra dashboard", same as we have upstream? I'd be interested in that, even if it had only one item this far. I suppose that could be doable with cron jobs and building a webpage to serve statically from cobra itself, to keep it as safe as possible. But that's a lot of work up front to have that as a platform where I can just add more checks. Which brings us in a circle back here - hammering in this screw would be easier...

Benefit of your proposed solution would be a great simplification. We already have HTTP space and you would be able to just list files in the OS directly. This is honestly extremely easy to do. The only thing you need is one script which would generate static HTML.

VladimirSlavik:
I think we agree on not doing what this PR suggests, which is the good part.

At the end of it, I will happily concede these points:
- Mixing upstream and downstream things is not in the spirit of what we as Red Hat employees want to do.
- We should minimize the amount of things that could exfiltrate company-internal data.

However I have trouble understanding the "why".

Do you want to say that a contributor will be negatively affected by seeing a status that shows health of a RH-employee-only service? We already publicly display statuses for running kickstart tests on both major RHELs (more than Fedora...), building a RHEL COPR, building boot.iso for RHEL, and probably some more. These are all internal, and by the offered logic, contributors have no interest in these. So should we also remove all of that because these are not for the public? I don't think a contributor will go check state of our infrastructure, or we would have already heard some complaints. But I fail to see how this is different then.

Or is the distinction that PXE on cobra02 serves only the RH-paid developer team, while the rest of RHEL-related things hypothetically supports CentOS stream as well as second order downstreams?

VladimirSlavik:
> IMHO I think we should switch these to CentOS and work completely upstream but that is another story.

Agreed. The current workflow is de facto that, only the branches are called "rhel".

PR #4271, opened 2022-08-16
Show multiple bootloader devices on the Manual Partitioning screen

poncovka:

was removed.

Related: rhbz#2088113
Related: rhbz#2093793
Related: rhbz#1913035

### DIFF
--- a/pyanaconda/modules/storage/partitioning/interactive/utils.py
+++ b/pyanaconda/modules/storage/partitioning/interactive/utils.py
@@ -61,7 +61,7 @@ def collect_used_devices(storage):
     used_devices = []
 
     for root in storage.roots:
-        for device in list(root.mounts.values()) + root.swaps:
+        for device in root.devices:
             if device not in storage.devices:
                 continue
             used_devices.extend(device.ancestors)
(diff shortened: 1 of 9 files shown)


pyanaconda/modules/storage/partitioning/interactive/utils.py
```diff
@@ -227,29 +227,15 @@ def create_new_root(storage, boot_drive):
...
-    ]
-
     mounts = {
         d.format.mountpoint: d for d in devices
         if getattr(d.format, "mountpoint", None)
     }
 
-    for device in devices:
-        if device in bootloader_devices:
-            mounts[device.format.name] = device
```
jkonecny12, line 247: I have a bad feeling about removing this part. We don't really need it?

poncovka: Well, this is exactly that code that adds bootloader devices to a dictionary of mount points, so if you have two biosboot devices, you will end up with only one of them. Without this change, the rest of it is pointless.

rvykydal (approved):
Looks good to me.

poncovka:
Unfortunately, Blivet still contains some code that manipulates with our installation roots. The pull request uses some temporary workarounds to deal with that.

M4rtinK (approved):
Looks good to me as well. :)

VladimirSlavik (approved):
Looks good to me. Thank you!

PR #4274, opened 2022-08-18
Fix building for RHEL/ELN without live installer

sgallagher:
The conditional was excluding the `BuildRequires: desktop-file-utils`
but still attempted to execute the tool.

This patch fixes that, as well as converting to RPM bcond
conditionals for enabling the live installer feature.

### DIFF
--- a/anaconda.spec.in
+++ b/anaconda.spec.in
@@ -19,7 +19,10 @@ Source0: https://github.com/rhinstaller/%{name}/releases/download/%{name}-%{vers
 # match the requires versions of things).
 
 %if ! 0%{?rhel}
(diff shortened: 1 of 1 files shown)


sgallagher:
I think this is incomplete. It appears that the build still produces the live installer files. Probably this needs a more comprehensive solution.

VladimirSlavik:
Out of curiosity, what's the rationale for merging? I think that this streamlines the [live yes/no] mechanism in specfile, but was there any discussion about the failing tests?

VladimirSlavik:
Oh, the tests are now actually fixed... then it makes sense, sorry.

PR #4275, opened 2022-08-19
Add --hibernation option for Kickstart autopart

OndrejZobal:
Previously it wasn't possible to create a hibernation swap using
autopart.

Add a --hibernation flag that will force autopart to
create a swap partition for hibernation even if swap isn't
configured in anaconda.conf.

Corresponding changes to pykickstart: https://github.com/pykickstart/pykickstart/pull/420 

Resolves: [rhbz#882989](https://bugzilla.redhat.com/show_bug.cgi?id=882989)

### DIFF
--- a/pyanaconda/modules/storage/partitioning/automatic/automatic_partitioning.py
+++ b/pyanaconda/modules/storage/partitioning/automatic/automatic_partitioning.py
@@ -24,7 +24,8 @@ from pyanaconda.modules.storage.partitioning.automatic.noninteractive_partitioni
     NonInteractivePartitioningTask
 from pyanaconda.modules.storage.partitioning.automatic.utils import get_candidate_disks, \
     schedule_implicit_partitions, schedule_volumes, schedule_partitions, get_pbkdf_args, \
-    get_default_partitioning, get_disks_for_implicit_partitions
+    get_default_partitioning, get_part_spec, get_disks_for_implicit_partitions
+from pyanaconda.modules.storage.partitioning.specification import PartSpec
 from pyanaconda.core.storage import suggest_swap_size
 
 log = get_module_logger(__name__)
@@ -85,7 +86,7 @@ class AutomaticPartitioningTask(NonInteractivePartitioningTask):
             luks_data.pbkdf_args = pbkdf_args
 
         # Get the autopart requests.
-        requests = self._get_partitioning(storage, scheme, self._request.excluded_mount_points)
+        requests = self._get_partitioning(storage, scheme, self._request)
 
         # Do the autopart.
         self._do_autopart(storage, scheme, requests, encrypted, luks_format_args)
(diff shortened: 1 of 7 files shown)


OndrejZobal:
@pep8speaks whoops

pyanaconda/modules/storage/partitioning/automatic/automatic_partitioning.py
```diff
@@ -122,37 +123,59 @@ def _get_luks_format_args(storage, request):
...
+                spec.size = suggest_swap_size(hibernation=request.hibernation,
+                                              disk_space=disk_space)
+
+            specs.append(spec)
+
+        # When request.hibernation is set swap partition should be created even if it wasn't
+        # included in the default partitioning scheme.
+        if not will_swap_be_created and request.hibernation:
+            disk_space = storage.get_disk_free_space()
+            spec = PartSpec(
```
poncovka, line 163: This is dangerous. If we ever change the `get_default_partitioning` function, we can easily forget to update this part of the code. These bugs are tricky to find. You will have to figure out how to reuse the code of the `get_default_partitioning` function and probably create a helper function that converts `attrs` to `PartSpecs`.

pyanaconda/modules/storage/partitioning/automatic/automatic_partitioning.py
```diff
@@ -122,37 +123,59 @@ def _get_luks_format_args(storage, request):
...
-        :param excluded_mount_points: a list of mount points to exclude
+        :param request: partitioning parameters
         :return: a list of full partitioning specs
         """
-        requests = []
+        specs = []
+        will_swap_be_created = False
 
-        for request in get_default_partitioning():
+        for spec in get_default_partitioning():
```
poncovka, line 137: Feel free to split the loop. The effectivity of the code is usually not a concern for the installer.

1. Filter out unwanted specs.
2. Add a spec for swap if hibernation is requested and there is no swap.
3. Configure the specs.

Then the specs are configured the same way no matter where they come from.

poncovka (changes requested):
I have some suggestions.

VladimirSlavik (dismissed):
Thank you, looks good to me.

I can't comment on where the modification should be inserted, only that it seems ok AFAICT.

VladimirSlavik (commented):
Actually, I wonder, is the one new test enough? I don't understand this too well, but I'd expect a 2x2 matrix of cases?

tests/unit_tests/pyanaconda_tests/modules/storage/partitioning/test_module_part_automatic.py
```diff
@@ -283,6 +288,106 @@ def test_get_partitioning(self, platform, suggest_swap_size):
...
+            request=partitioning_request
+        )
+
+        swap = None
+        for spec in requests:
+            if spec.fstype == "swap":
+                swap = spec
+                break
+
+        assert swap is not None
```
poncovka, line 325: You can remove the loop and check `assert any(spec for spec in requests if spec.fstype == "swap")`. Same below. It means that there has to be at least one spec with the swap file system type.

tests/unit_tests/pyanaconda_tests/modules/storage/partitioning/test_module_part_automatic.py
```diff
@@ -283,6 +288,106 @@ def test_get_partitioning(self, platform, suggest_swap_size):
...
+        suggest_swap_size.assert_called_with(hibernation=True, disk_space=ANY)
```
poncovka, line 326: You check that the size was calculated, but that doesn't mean that it was assigned to the right spec. You can use the check from the test above to verify the sizes.

pyanaconda/modules/storage/partitioning/automatic/automatic_partitioning.py
```diff
@@ -122,37 +123,54 @@ def _get_luks_format_args(storage, request):
...
-
-        return requests
+        for spec in specs:
+            # Set the default filesystem type.
+            if spec.fstype is None:
+                spec.fstype = storage.get_fstype(spec.mountpoint)
+
+            # Detect swap.
+            if spec.fstype == "swap":
+                with open("/tmp/sus", "a") as f:
```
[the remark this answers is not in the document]
OndrejZobal: I'm really sorry about this, should have caught this.

tests/unit_tests/pyanaconda_tests/modules/storage/partitioning/test_module_part_automatic.py
OndrejZobal: True, I should check the actual spec size but checking how suggest_swap_size was called is also relevant so I will keep it there.

tests/unit_tests/pyanaconda_tests/modules/storage/partitioning/test_module_part_automatic.py
OndrejZobal: Thanks, this is much better.

pyanaconda/modules/storage/partitioning/automatic/automatic_partitioning.py
```diff
@@ -122,37 +123,48 @@ def _get_luks_format_args(storage, request):
...
                 continue
 
             # Skip excluded mount points.
-            if (request.mountpoint or request.fstype) in excluded_mount_points:
+            if (spec.mountpoint or spec.fstype) in request.excluded_mount_points:
                 continue
 
+            specs.append(spec)
+
+        for spec in specs:
```
poncovka, line 148: Let's look at this part tomorrow and see if we can simplify it a little.

poncovka (approved):
Otherwise, it looks good to me. Thank you!

poncovka (approved):
Thanks!

pyanaconda/modules/storage/partitioning/automatic/automatic_partitioning.py
```diff
@@ -122,37 +123,53 @@ def _get_luks_format_args(storage, request):
...
                 continue
 
             # Skip excluded mount points.
-            if (request.mountpoint or request.fstype) in excluded_mount_points:
+            if (spec.mountpoint or spec.fstype) in request.excluded_mount_points:
                 continue
 
+            # Detect swap.
+            if spec.fstype == "swap":
+                swap = spec
```
VladimirSlavik, line 149: Is it a given there will be only one swap?

VladimirSlavik (approved):
Looks good to me. Thank you!

pyanaconda/modules/storage/partitioning/automatic/automatic_partitioning.py
OndrejZobal: I can't imagine a scenario where autopart would create more than one swap. But maybe it would be worth it to future/fool proof this by changing this.

poncovka: The code should be fine. The `swap` variable is basically just a flag at this moment. We just need to add one of there is none. The swap size is then recalculated for all of them.

PR #4280, opened 2022-08-23
webui: Build RPMs in the latest Fedora

poncovka:
The Cockpit CI doesn't run our `run` script in the latest Fedora,
so the Anaconda RPMs might be created for a wrong version of Python
and cause a failure of the Web UI tests.

### DIFF
--- a/.cockpit-ci/run
+++ b/.cockpit-ci/run
@@ -4,4 +4,12 @@
 
 set -eu
 
-./autogen.sh && ./configure && make webui-tests
+build_cmd="
+sudo ./scripts/testing/install_dependencies.sh -y
+./autogen.sh && ./configure && make rpms
+"
+
+toolbox create --container fedora-toolbox-latest
+toolbox run --container fedora-toolbox-latest /bin/sh -c "${build_cmd}" || exit 1
+
+make webui-tests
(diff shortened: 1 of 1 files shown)


poncovka:
@M4rtinK, the mock version should eventually work. I have tested it locally with [a Cockpit image in a toolbox](https://github.com/cockpit-project/cockpituous/blob/main/tasks/README.md#using-with-toolbx). It had some complains about GPG keys, but after `sudo dnf update`, it runs fine. I guess it just needs a newer image. It didn't complain about a missing command, so `mock` should be already available.

The pull request tests currently fails with:

```
DEBUG: Creating RPM ...
The password you typed is invalid.
Please try again.
```

I have no idea why. I cannot reproduce it. The `mock` documentation mentions:

```
To use mock, a user should become a member of the mock group by adding their username to
the mock line in /etc/group. This can be done with the following command:

sudo /usr/sbin/usermod -a -G mock $USER
```

Maybe the Cockpit image needs to do something like that.

poncovka:
> There are enough bug reports to guess with high enough confidence that the error message means the user running mock is not in the mock group. I wonder if we can do that ourselves? `sudo usermod -a -G mock $(whoami)` Being able to do that would be... a hole, I guess :D

I think I tried to use `sudo` at some point and it wanted a password.

M4rtinK:
The issue has been fixed in https://github.com/rhinstaller/anaconda/pull/4287 so I think we can close this PR.

PR #4281, opened 2022-08-23
CI templates

VladimirSlavik:
This should make the templates useful, and actually valid for the other branches too, except `rhel-8`.

It's probably a bit simplistic for beginning, but works.

Templates are used to:
- build matrices where appropriate
- remove files that do not belong on a branch
- generate most of the stuff from 3 values in the one file

I am not yet sure how to extend this to Dockerfiles, if we want to do that. The current mechanisms seem to handle that fairly well as-is, but there might be room for some simplification.


.github/workflows/README.md
```diff
@@ -0,0 +1,25 @@
...
+
+- Blocks `{% ... %}` let you use conditions to select which block will be present in the output.
+  If you don't put anything else on a line except for a block "delimiter", the line will completely
+  disappear from the output. 
+
+- Whitespace handling is a complicated affair. If you stick to the two methods above, things stay
+  mostly deterministic. You can use the `+` and `-` modifiers to manage whitespace in more detail,
+  but the actual removals extend beyond the line with the statement, so these are mostly useless.  
+
+-
```
rvykydal, line 25: Empty item.

VladimirSlavik:
Note to self: Review if really the disable-able workflows do not run on the branches -> https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#push

VladimirSlavik:
I added a commit modifying our kickstart version, to see what people think about templating the python code.

VladimirSlavik:
Looks like packaging has problems with templates, so let's not do this now.

OndrejZobal (approved):
LGTM

PR #4283, opened 2022-08-24
Don't remove existing BLS entries

cmurf:
Boot Loader Specification expects this location to be shared among distros. We can't assume they can be deleted. The user can indicate they're no longer needed by reformatting the /boot volume.
http://systemd.io/BOOT_LOADER_SPECIFICATION/
Fixes RHBZ 1874724 - dual boot Fedora, 2nd installation erases 1st installation's /boot/loader/entries *conf files
https://bugzilla.redhat.com/show_bug.cgi?id=1874724

### DIFF
--- a/pyanaconda/modules/storage/bootloader/utils.py
+++ b/pyanaconda/modules/storage/bootloader/utils.py
@@ -226,12 +226,6 @@ def create_bls_entries(sysroot, storage, kernel_versions):
     if os.path.exists(sysroot + "/usr/sbin/new-kernel-pkg"):
         return
 
-    # Remove any existing BLS entries, they will not match the new system's
-    # machine-id or /boot mountpoint.
-    for file in glob(sysroot + "/boot/loader/entries/*.conf"):
-        log.info("Removing old BLS entry: %s", file)
-        os.unlink(file)
-
     # Create new BLS entries for this system
     for kernel in kernel_versions:
         log.info("Regenerating BLS info for %s", kernel)


cmurf:
Even if we don't remove the entries:

This:
https://src.fedoraproject.org/rpms/grub2/c/3e40727f723213b92defb8df7396e6c579480483?branch=rawhide
plus this:
https://github.com/rhinstaller/anaconda/blob/master/pyanaconda/modules/storage/bootloader/utils.py#L248

means that an installation will replace the unique information in the existing BLS snippets, breaking them anyway.

VladimirSlavik:
Just dumping my first thoughts.

I am not sure if that addresses the real problem. The code can be traced to 6791e7320e6c222b92746e8f228026695fa0e2a0, where the change addresses something with live images and is clearly part of the live payloads. These days the code is in `create_bls_entries` which is called by `CreateBLSEntriesTask` which in turn is invoked from `install_bootloader_with_tasks`, always. However this actually executed behind an if, which means it still runs only for live ([here](https://github.com/rhinstaller/anaconda/blob/master/pyanaconda/modules/storage/bootloader/installation.py#L160)).

One question is, does the bug happen with Server too? The comments in bugzilla don't shed more light on that, and the comment above here doesn't, either.

There also isn't anything about the original bug being fixed, which would be reintroduced...

cmurf:
The problem in [1874724](https://bugzilla.redhat.com/show_bug.cgi?id=1874724) doesn't happen with Server, all the BLS snippets are present. (However I do hit bug [2120845](https://bugzilla.redhat.com/show_bug.cgi?id=2120845) thus the entries are broken.)

VladimirSlavik:
The tar image is a valid payload. There are some plans around that for when Fedora starts using image builder, I think, but that's quite vague.

VladimirSlavik:
Tests fail because in `test_create_bls_entries`, this logic needs flipping:
```
            assert not os.path.exists(
                root + "/boot/loader/entries/fake.conf"
            )
```

I can't say if that's the only change needed, but at a glance it seems so.

jlinton:
Just as a FYI, I've got a systemd-boot subclass for anaconda (its been a WIP for the past couple months), and by default it writes the loader entries to the ESP (/boot/efi/loader/entries) so the default path here isn't right. Part of assuring that is all works properly with minimal systemd/anaconda/etc changes is assuring that the /boot/loader directory doesn't exist. So I hacked it to unlink those directories as well if the target bootloader is systemd, but that piece of code is one of the many parts I've not come up with a good alternative for yet. The grub2-mkconfig bits should for sure be in the grub2 class...

VladimirSlavik:
This seems sort of going nowhere. Let's make sure it's not lost. Unfortunately I'm no expert in this area, so I have a few questions:
- Was this intended for merging? Is it still so?
- What criteria do we define to verify this works? How do we test these?
- In absence of definitive knowledge about the effects, are we willing to test the change by deploying in rawhide?

VladimirSlavik:
Sorry, just got to the bottom of it. According to Robbie Harwood (no idea about gh handle):

An installation can or can not remove an existing system. We don't check which of the BLS entries correspond to which system. Thus we can choose one of the two evils: Either keep the old entries, likely pointing to a now-removed system and thus broken, or remove all entries, make any old system unbootable, and ensure everything in the boot menu works. And since dual-boot is not the more common case... here we are, deleting all.

So, to fix this properly, we'd need to be able to determine which entries are valid and can be deleted.

That also means this PR won't work.

PR #4285, opened 2022-08-25
Update webui test doc make rpms

rvykydal:


ui/webui/test/README.rst
```diff
@@ -70,6 +70,7 @@ After the code is changed the testing environemnt needs to be updated.
 The most robust way of doing this is (from top level directory)::
 
     rm -rf ui/webui/dist/ updates.img
+    rm result/build/01-rpm-build/anaconda-*.rpm
```
VladimirSlavik, line 78: Does it matter if the directories remain? I'd expect to see `result/` added to the line above. Or does the SRPM have to stay?

poncovka: In theory, `make clean` might help. I use the commands below, that might be too brutal:

```
git clean -fxd --exclude /.idea
rm -rf ui/webui/{bots,dist}/
```

poncovka (commented):
Thank you for documenting this!

PR #4286, opened 2022-08-25
Scan CI container pip packages with dependabot (#infra)

VladimirSlavik:
Changes to pylint and astroid have been somewhat disruptive in the past. This will let us know when a new version comes, and make it a controlled process.

Unfortunately, pylint and dill need a patched build on Rawhide and 37 at the very moment, but never mind that.

The plan is to uncomment pylint and astroid once the pip version works again, and use precise version specifications to get dependabot update PRs.


dockerfile/anaconda-ci/requirements.txt
```diff
@@ -0,0 +1,21 @@
+# pip package install list
+# storing it as an independent file enables Dependabot to bump versions with noise and attention
+# silent updates of pylint and astroid usually cause breakages
+# for format details see https://pip.pypa.io/en/stable/reference/requirements-file-format/
+
+# unit tests of all kinds
+pocketlint
```
poncovka, line 7: We don't use specific versions, so I wonder what will Dependabot do with that.

dockerfile/anaconda-ci/Dockerfile
```diff
@@ -90,16 +90,7 @@ RUN set -ex; \
...
-  pocketlint \
-  coverage \
-  pycodestyle \
-  dogtail \
-  rpmfluff \
-  freezegun \
-  pytest \
-  pyyaml \
-  jinja2
+  pip install --no-cache-dir -r /root/requirements.txt
```
poncovka, line 93: This means that changes in `requirements.txt` won't trigger a build of a new container if I remember it correctly. I think that there is a relevant code that needs to be updated as well.

poncovka (approved):

dockerfile/anaconda-ci/requirements.txt
```diff
@@ -0,0 +1,21 @@
...
+rpmfluff
+freezegun
+pytest
+
+# jinja templates
+pyyaml
+jinja2
+
+# pylint and its supporting libs
+#pylint
```
poncovka, line 20: Btw. there is a new Pylint release. https://github.com/PyCQA/pylint/releases/tag/v2.15.0

VladimirSlavik: Thanks! That works with the new python, so I actually enabled it here with the given version and dropped it from dnf in dockerfile. So the PR now makes sense.

dockerfile/anaconda-ci/Dockerfile
VladimirSlavik: Fortunately, it's a bit more clever, the check is for everything in the dockerfile directory, so this will work.

dockerfile/anaconda-ci/requirements.txt
VladimirSlavik: Oh, I assumed that dependabot will just always resolve that to latest and be quiet. But that's actually a good question.

VladimirSlavik: I can't find any mentions of this, let alone documentation. Taking a glance at the code, I *think* packages with no version are just skipped over. We can try, and revert if it's not working.

poncovka (approved):
Thanks!

PR #4289, opened 2022-08-29
Update cockpit dependency

allisonkarlitskaya:
Here's another approach to #4268 for consideration.

### DIFF
--- a/ui/webui/test/prepare-updates-img
+++ b/ui/webui/test/prepare-updates-img
@@ -9,8 +9,8 @@ set -e
     # store these separately so we don't need to download them each time
     mkdir -p tmp/extra-rpms
     cd tmp/extra-rpms
-    test -e tmp/rpms/cockpit-ws*.rpm || curl -LO https://kojipkgs.fedoraproject.org//packages/cockpit/259/1.fc35/x86_64/cockpit-ws-259-1.fc35.x86_64.rpm
-    test -e cockpit-bridge*.rpm || curl -LO https://kojipkgs.fedoraproject.org//packages/cockpit/259/1.fc35/x86_64/cockpit-bridge-259-1.fc35.x86_64.rpm
+    test -e tmp/rpms/cockpit-ws*.rpm || curl -LO https://kojipkgs.fedoraproject.org//packages/cockpit/275/1.fc38/x86_64/cockpit-ws-275-1.fc38.x86_64.rpm
+    test -e cockpit-bridge*.rpm || curl -LO https://kojipkgs.fedoraproject.org//packages/cockpit/275/1.fc38/x86_64/cockpit-bridge-275-1.fc38.x86_64.rpm
 )
 
 # build the anaconda srpm (and result RPMs go in `tmp/rpms`)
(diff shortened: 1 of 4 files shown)


allisonkarlitskaya:
Ugh.  Using `test/common/pixel-tests push` defaults to pushing to the cockpit repo (becase the `GITHUB_BASE` override is only set from the `Makefile.am`).  Because of that, I naively pushed here:

    https://github.com/cockpit-project/pixel-test-reference/commit/9ba68d177b6818582d506fe5c25ea97cf0f58816

Assuming that it would work with anaconda tests as well.  Of course it doesn't.  If someone can cherry-pick that over to the anaconda repo, it would be nice :)

allisonkarlitskaya:
force-push to `# shellcheck disable` some stuff...

VladimirSlavik:
Yay, all green. Looks like it works!

M4rtinK (approved):

ui/webui/test/prepare-updates-img
```diff
@@ -1,21 +1,28 @@
...
-    # store these separately so we don't need to download them each time
-    mkdir -p tmp/extra-rpms
-    cd tmp/extra-rpms
-    test -e tmp/rpms/cockpit-ws*.rpm || curl -LO https://kojipkgs.fedoraproject.org//packages/cockpit/259/1.fc35/x86_64/cockpit-ws-259-1.fc35.x86_64.rpm
-    test -e cockpit-bridge*.rpm || curl -LO https://kojipkgs.fedoraproject.org//packages/cockpit/259/1.fc35/x86_64/cockpit-bridge-259-1.fc35.x86_64.rpm
-)
+COCKPIT_URLS='
+    https://kojipkgs.fedoraproject.org//packages/cockpit/275/1.fc38/x86_64/cockpit-ws-275-1.fc38.x86_64.rpm
+    https://kojipkgs.fedoraproject.org//packages/cockpit/275/1.fc38/x86_64/cockpit-bridge-275-1.fc38.x86_64.rpm
+'
```
VladimirSlavik, line 11: I agree with Martin that we should just take latest here. We are no longer trying to build the very foundations, so lagging behind cockpit no longer buys us time, instead it buys us debt... That said, we can take that step once this unclogs the queue :-)

VladimirSlavik (approved):
Looks good to me. Thank you!

ui/webui/test/prepare-updates-img
M4rtinK: Agreed - we should keep this in mind but can handle it later. :)

rvykydal (approved):
Looks good to me, thank you.

ui/webui/test/prepare-updates-img
allisonkarlitskaya: It's not about not moving forward, but rather about not having "moving parts" that can change between test runs.  Part of cockpit CI design is that the test environment doesn't change in any way between runs.

I had a thought about what we might be able to do here, and one of the best ways forward would be to somehow download the cockpit dependencies at the time we download the .iso and bundle them together somehow.  That would mean that we get regular cockpit updates, as part of the normal image refreshes, but would never get a shifting environment without taking explicit action.

VladimirSlavik: Ah, that's a good idea.

It's a shame that one can't use dependabot on rpms, there are many places that would be helpful.

PR #4291, opened 2022-08-29
Silence pylint warnings in webui code

VladimirSlavik:
These warnings become visible only locally when "bots" is checked out.

Needs porting to f37 once branching etc. is sorted out.


poncovka (approved):

VladimirSlavik:
Removing porting label since we won't do much more about web ui on f37.

PR #3792, opened 2022-01-24
Write all PVs to LVM devices file

VladimirSlavik:
Resolves: [rhbz#2040302](https://bugzilla.redhat.com/show_bug.cgi?id=2040302)

PR #4292, opened 2022-08-29
Create the LVM devices file (#2011329)

VladimirSlavik:
Write the LVM devices file with all PVs, to keep the previous behavior.
Then, copy it to the new system.

Resolves: rhbz#2011329
Resolves: rhbz#2040302

(cherry picked from commit 8e3223a677a1813ba6c15faead5fb6759b0f80c0)
(cherry picked from commit 6dd7ebbd79103075061c19427737203163d78410)

----

Port of #3771 and #3792.

@vojtechtrefny, do you know if there's upstream support already? Do I need a specific Blivet version?

### DIFF
--- a/pyanaconda/modules/storage/installation.py
+++ b/pyanaconda/modules/storage/installation.py
@@ -17,7 +17,9 @@
 # License and may only be used or replicated with the express permission of
 # Red Hat, Inc.
 #
+import itertools
 import os
+import shutil
 import parted
 
 from datetime import timedelta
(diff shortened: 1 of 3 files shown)


vojtechtrefny:
> @vojtechtrefny, do you know if there's upstream support already? Do I need a specific Blivet version?

You need blivet 3.5 which is now available in Fedora 37 and rawhide, but the LVM support itself still hasn't landed in Fedora -- the devices file feature was added in 2.03.12, all versions of Fedora are still on 2.03.11 (but that shouldn't be a problem, blivet should simply set `HAVE_LVMDEVICES` to `False`).

VladimirSlavik:
Thanks! That means this will work.

Now I realize that maybe the whole code should exit early `if not HAVE_LVMDEVICES`, so that we don't copy the file if it's created by something else. That's one hole I see now and didn't before...

VladimirSlavik:
From offline discussion: It's important to not even try copying the file. If anything wrote the file without our knowledge, the installed system would stop working once it's updated to a version of LVM that actually uses the file.

vojtechtrefny (approved):
Looks good to me.

pyanaconda/modules/storage/installation.py
```diff
@@ -327,6 +334,29 @@ def _write_escrow_packets(self, storage, sysroot):
 
         log.debug("escrow: write_escrow_packets done")
 
+    def _write_lvm_devices_file(self, storage, sysroot):
```
poncovka, line 337: It would be great to have some unit tests (or kickstart tests) for this functionality. If you make this method static, you can call it directly from the tests.

poncovka (approved):
Thank you!

VladimirSlavik:
Above just a rebase to fix tests.

PR #4297, opened 2022-08-30
Port cppcheck macros to f37

VladimirSlavik:
Turns out it's broken here too.


VladimirSlavik:
Tests can't succeed because there's a number of problems that will be resolved by multiple PRs. They can't be resolved anyway

Verified locally - both rpm and unit tests pass.

poncovka (approved):
Thank you!

PR #4300, opened 2022-08-31
Bump pykickstart version for F38_AutoPart

VladimirSlavik:
Apart from using the correct pykickstart version, fix also pylint and ignore pykickstart's unreleased changes.

This should fix unit tests on master as far as possible.

### DIFF
--- a/anaconda.spec.in
+++ b/anaconda.spec.in
@@ -42,7 +42,7 @@ Source0: https://github.com/rhinstaller/%{name}/releases/download/%{name}-%{vers
 %define libxklavierver 5.4
 %define mehver 0.23-1
 %define nmver 1.0
-%define pykickstartver 3.41-1
+%define pykickstartver 3.43-1
 %define pypartedver 2.5-2
 %define pythonblivetver 1:3.4.4-3
 %define rpmver 4.15.0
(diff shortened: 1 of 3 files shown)


tests/unit_tests/pyanaconda_tests/test_kickstart_specification.py
```diff
@@ -406,6 +406,7 @@ class ModuleSpecificationsTestCase(unittest.TestCase):
 
     # Names of the kickstart commands and data that should be temporarily ignored.
     IGNORED_NAMES = {
+        "SshKeyData"
```
poncovka, line 409: Is there anything else needed on our side? Shouldn't we just bump the version of `SshKeyData`?

poncovka (approved):
Thanks!

tests/unit_tests/pyanaconda_tests/test_kickstart_specification.py
VladimirSlavik: Oh, sorry, you are right. I missed that this was released too and thought it's only in the copr nightlies. Pushed the right version now.

rvykydal (approved):

poncovka (approved):

tests/unit_tests/pyanaconda_tests/test_kickstart_specification.py
poncovka: I think we don't use `pykickstart` from the copr nightlies anymore, but I am not sure.

VladimirSlavik: That actually describes well the state of infrastructure these days: Nobody really knows \o/

PR #4301, opened 2022-08-31
Web UI: Show the help drawer next to content

OndrejZobal:
Currently the help drawer is displayed above step content while the
guidelines for its use specify it should be displayed next to the
content, so that the user sees both.


M4rtinK (approved):
Looks good to me, thanks! :)

VladimirSlavik (approved):
Thank you!

PR #4307, opened 2022-09-01
Verify a biosboot partition on all installation targets

poncovka:
Make the installation targets accessible for all types of bootloaders.
By default, there is only one installation target, but GRUB2 can have
multiple targets if stage2 is part of an array.

Don't run the `verify_gpt_biosboot` storage validation check only on the stage1
device. Check all installation targets.

Related: rhbz#2088113
**Ported from:** https://github.com/rhinstaller/anaconda/pull/4277


poncovka:
biosboot_missing_raid_one
biosboot_missing_raid

PR #4309, opened 2022-09-01
Ignore dependabot commits in changelog

VladimirSlavik:
Somehow we missed that bit, at some unspecified time earlier...


poncovka (approved):
Thank you!

PR #4310, opened 2022-09-01
Release from detected tag, not starting branch (#infra)

VladimirSlavik:
Turns out this should have been changed with the rest of the workflow...

There is some chance that the new value won't work for manual input anyway, because it would be missing `refs/tags/` prefix.


M4rtinK (approved):
Looks good to me. :)

VladimirSlavik:
Looks like the prefix was not needed after all, so it works well.

PR #4314, opened 2022-09-02
Always set system purpose from the GUI

M4rtinK:
Fix the issue of system purpose not being set if the system is re-registered - registered first without system purpose being set, unregistered and registered again with system purpose being set.

Also fixup some outdated system purpose related strings I've noticed when looking into this.

Port from RHEL 9: #3773


VladimirSlavik (approved):
Thanks!

PR #4317, opened 2022-09-05
Run yelp under liveuser with the right environment

VladimirSlavik:
The `XAUTHORITY` variable is not handled, but that is not needed as of now.

Resolves: [rhbz#2124097](https://bugzilla.redhat.com/show_bug.cgi?id=2124097)

### DIFF
--- a/pyanaconda/ui/lib/help.py
+++ b/pyanaconda/ui/lib/help.py
@@ -236,21 +236,25 @@ def _find_best_help_file(current_locale, available_files):
     return None
 
 
+# Help user data
+HelpUser = namedtuple("HelpUser", ["name", "uid"])
+
+
 def _get_help_user():
-    """Get the uid to run help viewer under.
+    """Get the user name and uid to run help viewer under.
 
     Currently, this means: Let's see if we can run yelp under liveuser on live.
     For details, see: https://bugzilla.redhat.com/show_bug.cgi?id=2118832
 
-    :return: user
-    :rtype: int | None
+    :return: user name and uid
+    :rtype: HelpUser | None
     """
     if not conf.system.provides_liveuser:
         return None
 
     try:
         user = getpwnam("liveuser")
-        return user.pw_uid
+        return HelpUser(name="liveuser", uid=user.pw_uid)
     except KeyError:
         return None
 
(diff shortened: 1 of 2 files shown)


pyanaconda/ui/lib/help.py
```diff
@@ -282,6 +282,23 @@ def show_graphical_help(help_path, help_anchor=None):
...
-    yelp_process = startProgram(["yelp", *args], reset_lang=False, user=user)
+    try:
+        user_name, user_id = _get_help_user()
+        env_prune = ("GDK_BACKEND",)
+        env_add = {
+            "XDG_RUNTIME_DIR": "/run/user/{}".format(user_id),
+            "USER": user_name,
+            "HOME": "/home/{}".format(user_name),
+        }
+    except TypeError:
```
poncovka, line 293: It is very hard to see what triggers this exception. If you don't want to check `None` and you prefer to use the exception handler, I suggest to raise an exception from `_get_help_user()`.

pyanaconda/ui/lib/help.py
```diff
@@ -237,20 +237,20 @@ def _find_best_help_file(current_locale, available_files):
...
-    """Get the uid to run help viewer under.
+    """Get the user name and uid to run help viewer under.
 
     Currently, this means: Let's see if we can run yelp under liveuser on live.
     For details, see: https://bugzilla.redhat.com/show_bug.cgi?id=2118832
 
-    :return: user
-    :rtype: int | None
+    :return: user name and uid
+    :rtype: (str, int) | None
```
poncovka, line 246: I would never recommend mixing types like this, unless there is a very good reason for it. Let's raise an exception instead of returning `None`. It will a more Pythonic solution.

poncovka (changes requested):

pyanaconda/ui/lib/help.py
VladimirSlavik: Actually, let's use some structured type to handle the data, so that it can be None as usual.

poncovka (approved):
Thanks!

VladimirSlavik:
Just a side note, I think keeping `XAUTHORITY` variable as-is is correct, because we want to run under the same session. But then, I don't really know what I'm doing with Xorg, so I might be wrong, too.

M4rtinK (approved):

PR #4319, opened 2022-09-06
packit templates

VladimirSlavik:
Let's generate packit config from templates, too.

This is to be ported to `f37-release` too.

Some actual config changes based on information in #3827 and irc conversation with @lachmanfrantisek


M4rtinK (approved):
Looks good to me, thanks! :)

PR #4323, opened 2022-09-08
Improve Register button click feedback

M4rtinK:
Set the phase right after button click to avoid the registration status being out of date.

So far that has been handled by the registration thread, which could result in the status message being out of date until the thread gets to the status update.

(cherry picked from commit 28615ae07e2a3de37c07f2d585e5b1b523b73d30)

Resolves: rhbz#2074636

Port from RHEL9: #4258


VladimirSlavik (approved):
Thanks!

PR #4325, opened 2022-09-08
gui: fix summary hub layout for Japanese translations

rvykydal:
The patch makes the spoke buttons more compact vertically (same as with other languages), making User spoke accessible without scrolling.

Resolves: rhbz#2120293


poncovka (approved):

VladimirSlavik (approved):
Thanks!

PR #4326, opened 2022-09-08
Use the Bugzilla API key in the `makebumpver` script

poncovka:
The `~/.rhbzauth` configuration file is no longer supported. Use the configuration file supported directly by `python-bugzilla` instead. The support for username and password was removed. Use the API key instead.

The `makebumpver` script will ask for the API key interactively if none is available. The key will be saved to a configuration file and reused next time. Alternatively, you can create the `~/.bugzillarc` file with the API key yourself with the following content:
```
    [bugzilla.redhat.com]
    api_key = <API-KEY>
```

### DIFF
--- a/scripts/makebumpver
+++ b/scripts/makebumpver
@@ -28,7 +28,6 @@ log.setLevel(logging.INFO)
 import bugzilla  # pylint: disable=import-error
 import datetime
 import argparse
-import getpass
 import os
 import re
 import subprocess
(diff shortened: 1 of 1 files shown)


scripts/makebumpver
```diff
@@ -247,21 +232,12 @@ class MakeBumpVer:
...
-
-            bzclass = bugzilla.Bugzilla
-            self.bz = bzclass(url=self.bzurl)
+            self.bz = bugzilla.Bugzilla(url=self.bzurl)
 
         if not self.bz.logged_in:
-            rc = self.bz.login(self.username, self.password)
-            log.debug("login rc = %s", rc)
+            msg = "Provide an API key from https://{}/userprefs.cgi?tab=apikey\n"
+            sys.stdout.write(msg.format(self.bzserver))
```
VladimirSlavik, line 239: I think it might make more sense to have the format on same line as the string literal, so that it can be checked.

VladimirSlavik (approved):
Thanks!

rvykydal (approved):
Thank you!

scripts/makebumpver
poncovka: Fixed.

poncovka:
I have tested the `_queryBug` method. It seems to work fine.

M4rtinK (approved):
Looks good to me as well. :)

PR #4328, opened 2022-09-09
F37 geoloc wait again

VladimirSlavik:
From user point of view, this makes anaconda geolocate again, with roughly the same chances.

From testing point of view, this waits again, so it fixes the bug.

From development point of view:
- The waiting is back, but moved into the anaconda code before initializing UI. That means the actual timeout is different. Previously it was some init + load gtk + load gui + timeout. Now it's only some init + timeout + load gtk + load gui. It's unclear if the different order might make a difference or not. What it makes different for debugging is that logs might report failure to geolocate but as the actual use is after loading GUI, it will still succeed. The reporting makes little sense, but read on.
- In f37, we stop here. When porting to master/rawhide, we'll move also the part setting locales etc. into the main script, so that all UIs get geolocation equally. However, this will remove all the GUI delays, so technically, weak machines on connections with very high latency might fail to geolocate on f38/rawhide, but succeed on f36.
- ~~The new Task helper is not tested. That's also to be resolved when porting to master - the tests for these parts are pretty spaghettized and need more work that is unnecessary for f37. The branch with some work is `geoloc-wait-again-to-port`.~~

### DIFF
--- a/pyanaconda/modules/common/task/__init__.py
+++ b/pyanaconda/modules/common/task/__init__.py
@@ -15,12 +15,14 @@
 # License and may only be used or replicated with the express permission of
 # Red Hat, Inc.
 #
-from time import sleep
+import math
+from time import sleep, perf_counter
 
 from pyanaconda.modules.common.task.task_interface import TaskInterface
 from pyanaconda.modules.common.task.task import Task, AbstractTask
 
-__all__ = ["sync_run_task", "async_run_task", "AbstractTask", "Task", "TaskInterface"]
+__all__ = ["sync_run_task", "async_run_task", "wait_for_task", "AbstractTask", "Task",
+           "TaskInterface"]
 
 
 def sync_run_task(task_proxy, callback=None):
@@ -67,3 +69,26 @@ def async_run_task(task_proxy, callback):
 
     task_proxy.Stopped.connect(_callback)
     task_proxy.Start()
+
+
+def wait_for_task(task_proxy, timeout=math.inf):
+    """Wait for an existing and running task with optional timeout.
+
+    If the timeout exception is raised, the task is not done. To call its Finish method and
+    receive potential errors from its run, you can attach callbacks to the tasks's signals.
+    Alternatively, you can give up re-raising the errors entirely.
+
+    :param task_proxy: a proxy of the remote task
+    :param float timeout: stop waiting after this time in seconds
+    :raise TimeoutError: when the task did not finish before timeout
+    """
+    start = perf_counter()
+    end = start + timeout
+
+    while task_proxy.IsRunning and perf_counter() <= end:
+        sleep(0.1)
+
+    if task_proxy.IsRunning:
+        raise TimeoutError()
+
+    task_proxy.Finish()
--- a/pyanaconda/modules/timezone/timezone_interface.py
+++ b/pyanaconda/modules/timezone/timezone_interface.py
@@ -21,6 +21,7 @@ from dasbus.server.property import emits_properties_changed
 from dasbus.typing import *  # pylint: disable=wildcard-import
 from dasbus.server.interface import dbus_interface
 
+from pyanaconda.modules.common.containers import TaskContainer
 from pyanaconda.modules.common.base import KickstartModuleInterface
 from pyanaconda.modules.common.constants.services import TIMEZONE
 from pyanaconda.modules.common.structures.timezone import TimeSourceData, GeolocationData
@@ -118,9 +119,14 @@ class TimezoneInterface(KickstartModuleInterface):
             TimeSourceData.from_structure_list(sources)
         )
 
-    def StartGeolocation(self):
-        """Start geolocation, if not started yet."""
-        self.implementation.start_geolocation()
+    def StartGeolocationWithTask(self) -> ObjPath:
+        """Start geolocation with task.
+
+        :return: a DBus path of the task
+        """
+        return TaskContainer.to_object_path(
+            self.implementation.start_geolocation_with_task()
+        )
 
     @property
     def GeolocationResult(self) -> Structure:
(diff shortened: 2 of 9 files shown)


pyanaconda/modules/common/task/__init__.py
```diff
@@ -67,3 +69,27 @@ def _callback():
...
+
+    :param task_proxy: a proxy of the remote task
+    :param float timeout: stop waiting after this time in seconds
+    :return bool: did the task finish or not
+    """
+    start = perf_counter()
+    end = start + timeout
+
+    while task_proxy.IsRunning and perf_counter() <= end:
+        sleep(0.1)
```
poncovka, line 88: We use `sleep(1)` for `sync_run_task`. Maybe, we should use it also here, because `task_proxy.IsRunning` is not a cheap operation.

pyanaconda/modules/timezone/timezone_interface.py
```diff
@@ -118,9 +119,11 @@ def TimeSources(self, sources: List[Structure]):
             TimeSourceData.from_structure_list(sources)
         )
 
-    def StartGeolocation(self):
-        """Start geolocation, if not started yet."""
-        self.implementation.start_geolocation()
+    def StartGeolocationWithTask(self) -> ObjPath:
+        """Start geolocation with task."""
```
poncovka, line 123: Please, document the return value in the doc string.

pyanaconda/modules/common/task/__init__.py
```diff
@@ -67,3 +69,27 @@ def _callback():
...
+
+    if not task_proxy.IsRunning:
+        task_proxy.Finish()
+        return True
+
+    else:
+        return False
```
poncovka, line 95: I prefer to raise and handle the `TimeoutError` exception in these cases. It somehow feels more standard, but this solution is also fine.

poncovka (approved):
I have some small suggestions. Otherwise, it looks good to me. Thank you!

pyanaconda/modules/common/task/__init__.py
VladimirSlavik: 1 second seems to be awfully coarse. The distant side of the D-Bus call should be performant enough to make 100000s of calls per second, so the real question is, does D-Bus take tens of ms, so that it's comparable to the wait?

pyanaconda/modules/common/task/__init__.py
VladimirSlavik: Let's do that, I have no preference.

I also wondered if there should be a handler to attach in case the Task is not finished. But at the same time, if the finished/unfinished state is communicated in other ways, I'm not sure what value that could bring.

VladimirSlavik:
I just realized that perhaps the start and finish code should be helpers in `startup_utils`... :(

rvykydal (approved):
LGTM, I don't have anything to add.

pyanaconda/modules/common/task/__init__.py
VladimirSlavik: A quick and dirty test on my test VM* shows average 3 ms per `IsRunning` call, at 25 iterations. I think that's totally fine - 3% query, 97% sleep.

\* 2 CPU, 2 GB memory - intentionally specced low

pyanaconda/modules/common/task/__init__.py
```diff
@@ -67,3 +69,25 @@ def _callback():
...
+
+def wait_for_task(task_proxy, timeout=math.inf):
+    """Wait for an existing and running task with optional timeout.
+
+    If this returns True and the task is finished, you should call the Finish method of the remote
+    task and handle the remote errors. If this returns False, either
+
+    :param task_proxy: a proxy of the remote task
+    :param float timeout: stop waiting after this time in seconds
+    :return bool: did the task finish or not
```
poncovka, line 82: The doc string needs to be updated.

pyanaconda/startup_utils.py
```diff
@@ -646,3 +646,36 @@ def check_if_geolocation_should_be_used(opts):
 
     log.info("Geolocation is enabled.")
     return True
+
+
+def start_geolocation_conditionally(opts):
+    """Start geolocation conditionally, according to the command line or boot options.
+
+    :param opts: the command line/boot options
+    :return: a DBus path of the geolocation task
```
poncovka, line 655: It returns a DBus proxy.

pyanaconda/startup_utils.py
```diff
@@ -646,3 +646,36 @@ def check_if_geolocation_should_be_used(opts):
...
+    geoloc_task_path = timezone_proxy.StartGeolocationWithTask()
+    geoloc_task_proxy = TIMEZONE.get_proxy(geoloc_task_path)
+    geoloc_task_proxy.Start()
+    return geoloc_task_proxy
+
+
+def wait_for_geolocation(geoloc_task_proxy):
+    """Wait for geolocation, if started
+
+    :param geoloc_task_proxy: path to the GeolocationTask proxy
```
poncovka, line 671: It accepts a DBus proxy. DBus path is a string. DBus proxy is a wrapper of a remote object that is specified by the DBus path.

pyanaconda/startup_utils.py
```diff
@@ -646,3 +646,36 @@ def check_if_geolocation_should_be_used(opts):
...
+    """
+    if not geoloc_task_proxy:
+        return
+
+    from pyanaconda.modules.common.task import wait_for_task
```
poncovka, line 676: I think this import can be defined on the top level.

poncovka (approved):
The doc strings need some changes. Otherwise, it looks great!

pyanaconda/modules/common/task/__init__.py
poncovka: Since we agreed that 0.1 seconds should be enough, could you open a PR for `sync_run_task` in the upstream, please?

poncovka (approved):
Thanks!

PR #4331, opened 2022-09-12
Move auditd to its own directory

VladimirSlavik:
Also, document what it is. Or not? Who knows.


utils/auditd/README.md
```diff
@@ -0,0 +1,11 @@
...
+------
+
+Q: What is this?
+A: A simple audit daemon that throws away all messages.
+
+Q: Is it used anywhere?
+A: Yes. Lorax templates replace the actual well-known auditd with our stub.
+
+Q: Why?
+A: We don't know anymore, it's a cargo cult.
```
VladimirSlavik, line 11: It would be nice to fill this hole in the knowledge.

(@bcl, would you happen to remember?)

VladimirSlavik:
As a tangent: It seems that doing this switcheroo might still make sense, because the normal `auditd` is a pain:

While its config `auditd.conf` has option `write_logs=no`, you can't control its execution manually via `systemctl` (explicitly disabled), apparently because it needs to audit it being stopped, which it can't: https://lists.freedesktop.org/archives/systemd-devel/2014-April/018608.html

On my system it also uses some 90 MB of memory which isn't something to hand-wave away for the installation environment...

utils/auditd/README.md
bcl: I had to ask pjones -- I had totally forgotten that was there. This was written in 2006 when the available space was much smaller. But it is probably still valid. auditd takes up a bunch of space, it doesn't really provide anything useful in the installation environment, and would likely clutter up the logs with unnecessary messages.

VladimirSlavik:
Linking the mailing list here too: https://lists.fedoraproject.org/archives/list/anaconda-devel@lists.fedoraproject.org/thread/U23ZN2LPXMWCOMTMRRZ4XG36TAWJJGLT/?sort=thread

Summary: Might be possible to replace with a python implementation, probably better to keep as-is because it works.

Will amend the readme.

VladimirSlavik:
Researching if dropping this completely is possible.

Notes to self:
- audit constitutes ca. 420/5600 lines in journal, similar numbers for syslog

VladimirSlavik:
Replaced by #4358 which removes this completely.

PR #4337, opened 2022-09-13
Re-enable tag release for branched fedora (#infra)

VladimirSlavik:
This is how it works: The workflow to react to a tag is taken from that tag's ref.

(cherry picked from commit 60cda265c9564a35c9f6e3ccadf5c6d8a0bc8dd5)

---

Port of the new parts in #4336. This should be the final piece to enable automatic releases on branched Fedora with just `infra-reload`. To be tested on `f37-release` with #4336 and another release.


VladimirSlavik:
This succeeded on `f37-release`, so let's have it.

PR #4339, opened 2022-09-13
Reload packit config from template (#infra)

VladimirSlavik:
Let's also write a workflow to check that no `infra-reload` is forgotten :(


M4rtinK (approved):
Looks good to me, thanks! :)

PR #4342, opened 2022-09-15
Web UI: Redesign language selection screen

OndrejZobal:
Redesign language selection screen according to our latest mockups. Select element was replaced with a menu and a search bar and an alert was added to show to current language.

### DIFF
--- a/pyanaconda/modules/localization/localization.py
+++ b/pyanaconda/modules/localization/localization.py
@@ -141,6 +141,15 @@ class LocalizationService(KickstartService):
         """
         return get_language_locales(language_id)
 
+    def get_common_locales(self):
+        """Get a list of the most commonly used locales.
+
+        For example: ["ar_EG.UTF-8", "en_US.UTF-8", "en_GB.UTF-8", ...]
+
+        :return: a list of common locale IDs
+        """
+        return langtable.list_common_locales()
+
     def get_locale_data(self, locale_id):
         """Get data about the specified locale.
 
--- a/pyanaconda/modules/localization/localization_interface.py
+++ b/pyanaconda/modules/localization/localization_interface.py
@@ -68,6 +68,15 @@ class LocalizationInterface(KickstartModuleInterface):
         """
         return self.implementation.get_locales(language_id)
 
+    def GetCommonLocales(self) -> List[Str]:
+        """Get a list of the most commonly used locales.
+
+        For example: ["ar_EG.UTF-8", "en_US.UTF-8", "en_GB.UTF-8", ...]
+
+        :return: a list of common locale IDs
+        """
+        return self.implementation.get_common_locales()
+
     def GetLocaleData(self, locale_id: Str) -> Structure:
         """Get data about the specified locale.
 
--- a/tests/unit_tests/pyanaconda_tests/modules/localization/test_module_localization.py
+++ b/tests/unit_tests/pyanaconda_tests/modules/localization/test_module_localization.py
@@ -21,6 +21,7 @@ import os
 import tempfile
 import unittest
 import pytest
+import langtable
 
 from unittest.mock import patch, Mock, call
 from textwrap import dedent
@@ -231,6 +232,14 @@ class LocalizationInterfaceTestCase(unittest.TestCase):
             "native-name": get_variant(Str, "English (United States)"),
         }
         assert data == english_us
+
+    def test_common_locales(self):
+        common_locales = self.localization_interface.GetCommonLocales()
+
+        assert isinstance(common_locales, list)
+        assert "en_US.UTF-8" in common_locales
+        assert "ja_JP.UTF-8" in common_locales
+        assert self.localization_interface.GetCommonLocales() == langtable.list_common_locales()
 
     @patch_dbus_publish_object
     def test_install_with_task(self, publisher):
--- a/ui/webui/src/components/localization/InstallationLanguage.jsx
+++ b/ui/webui/src/components/localization/InstallationLanguage.jsx
@@ -21,13 +21,16 @@ import cockpit from "cockpit";
 import {
     Form,
     FormGroup,
-    SelectGroup,
-    SelectOption,
-    Select,
-    SelectVariant,
-    Text,
-    TextVariants,
-    TextContent,
+    Title,
+    Menu,
+    MenuList,
+    MenuInput,
+    MenuItem,
+    MenuContent,
+    MenuGroup,
+    TextInput,
+    Divider,
+    Alert,
 } from "@patternfly/react-core";
 
 import { EmptyStatePanel } from "cockpit-components-empty-state.jsx";
@@ -41,6 +44,7 @@ import {
     getLocales,
     getLocaleData,
     setLanguage,
+    getCommonLocales,
 } from "../../apis/localization.js";
 
 import {
@@ -49,6 +53,8 @@ import {
     setLangCookie
 } from "../../helpers/language.js";
 import { AnacondaPage } from "../AnacondaPage.jsx";
+
+import "./InstallationLanguage.scss";
 
 const _ = cockpit.gettext;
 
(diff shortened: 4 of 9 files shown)


M4rtinK:
> * The function which is supposed to be capable of setting text to input fields doesn't work with my search field even though it's just an input field.

That's weird & possibly either Cockpit testlib or PatternFly bug.

> 
> * Cannot use wait_visible because common locales appear twice, so I need something else that will tell be at least one instance of selector is visible.

I think it should be possible to use the prefix as selector & check the locale or human readable string is as expected - lookinmg at the code, all the common locals have `option-common-` in its id & all the normal languages have `option-alpha-` in the id. So the test cold check "de_DE.UTF-8" twice, once with each of the two ids.

>   There is no documentation on this test library so I would be very happy if anyone could help me with this.
Yeah, I'm afraid I was also only able to find top level documentation for it so far - such as:

https://cockpit-project.org/external/source/test/README
https://cockpit-project.org/external/source/HACKING

What I did instead, is checking how the Cockpit project uses the test library - for example, say you want to see how  `wait_in_text` can be used, I just search for it on top of the cockpit-project organization on GitHub:

https://github.com/search?q=org%3Acockpit-project+wait_in_text&type=code

That way it looks at all the main Cockpit plugins being developped under Cockpit project and their test suites. Alternatively one can check the test suites themselves to see what functions are available:

https://github.com/cockpit-project/cockpit-machines/tree/main/test

Not ideal, but doable...

As for possible points of contact about the Cockpit test suite:

* IRC #cockpit on libera.chat
* @martinpitt (pitti on the IRC channel above)
* also @velezd from RTT has been looking into the Cockpit test API in the past & created https://github.com/rhinstaller/anaconda/tree/master/ui/webui/test/helpers with the aim of reducing code duplication in our test suite

Also it looks like the testlib itself lives under the main cockpit project:

https://github.com/cockpit-project/cockpit/blob/main/test/common/testlib.py

So if something looks broken, you can report an issue to this project - from my experience so far the Cockpit team is pretty responsive. :)

M4rtinK:
@OndrejZobal BTW, once they exist I recommend pointing to relevant related PRs for this (eq. langtable & DBus API in case it comes via separate PR) as well as adding screenshots of the current iteration, so that @lclay2 can review how the implementation of the UX mockup looks like. :)

OndrejZobal:
Current interation
obrazek

OndrejZobal:
It looks like pixel test are failing because of missing fonts :thinking: 
image

pyanaconda/modules/localization/localization_interface.py
```diff
@@ -68,6 +69,15 @@ def GetLocales(self, language_id: Str) -> List[Str]:
...
         return self.implementation.get_locales(language_id)
 
+    def GetCommonLocales(self) -> List[Str]:
+        """Get a list of the most commonly used locales.
+
+        For example: ["ar_EG.UTF-8", "en_US.UTF-8", "en_GB.UTF-8", ...]
+
+        :return: a list of common locale IDs.
+        """
+        return langtable.list_common_locales()
```
poncovka, line 79: We follow the Proxy software design pattern here (https://en.wikipedia.org/wiki/Proxy_pattern). The `*Interface` classes are just proxies between the actual implementation and a message bus. Their only job is to convert the values into something that can be send on DBus and back.

In this case, you should call `self.implementation.get_common_locales()` and use `langtable` in the `LocalizationService` class.

tests/unit_tests/pyanaconda_tests/modules/localization/test_module_localization.py
```diff
@@ -232,6 +233,9 @@ def test_locale_data(self):
         }
         assert data == english_us
 
+    def test_common_locales(self):
+        assert self.localization_interface.GetCommonLocales() == langtable.list_common_locales()
```
poncovka, line 237: This test should confirm that we are not trying to send data that are not supported by DBus. Please, make sure that the return value is a list and check if, for example, `en_US.UTF-8` and `ja_JP.UTF-8` is in this list. I would be fine with checking all values. If langtable starts to return different values, it will probably break the Web UI pixel tests anyway.

poncovka (changes requested):
The changes in the backend need some polishing.

M4rtinK:
> It looks like pixel test are failing because of missing fonts thinking
That's strange - does it happen only for the pixel tests or also if running locally ?

OndrejZobal:
@M4rtinK 
> That's strange - does it happen only for the pixel tests or also if running locally ?

Nope, locally they pass just fine.
The affected fonts are Japanese and Chinese and I think russian has a slightly different font on my machine than it does on the cockpit test.

I know in web-dev fonts are sometimes fetched from a 3rd party cdn (I don't know if this is our case but I will look into it) and if the bot VMs don't have internet than that would explain it.

M4rtinK:
> > That's strange - does it happen only for the pixel tests or also if running locally ?
> 
> Nope, locally they pass just fine. The affected fonts are Japanese and Chinese and I think russian has a slightly different font on my machine than it does on the cockpit test.
> 
> ~I know in web-dev fonts are sometimes fetched from a 3rd party cdn (I don't know if this is our case but I will look into it) and if the bot VMs don't have internet than that would explain it.~ Never mind, the Red Hat fonts arre fetch from the server.

Strange! @martinpitt Have you seen something like this in your tests ? I don't think we are doing something special here in regards to font handling.

pyanaconda/modules/localization/localization.py
```diff
@@ -141,6 +141,15 @@ def get_locales(self, language_id):
         """
         return get_language_locales(language_id)
 
+    def get_common_locales(self):
+        """Get a list of the most commonly used locales.
+
+        For example: ["ar_EG.UTF-8", "en_US.UTF-8", "en_GB.UTF-8", ...]
+
+        :return: a list of common locale IDs.
```
poncovka, line 149: Please, remove the dot at the end. It is not a sentence.

pyanaconda/modules/localization/localization_interface.py
```diff
@@ -68,6 +67,15 @@ def GetLocales(self, language_id: Str) -> List[Str]:
         """
         return self.implementation.get_locales(language_id)
 
+    def GetCommonLocales(self) -> List[Str]:
+        """Get a list of the most commonly used locales.
+
+        For example: ["ar_EG.UTF-8", "en_US.UTF-8", "en_GB.UTF-8", ...]
+
+        :return: a list of common locale IDs.
```
poncovka, line 75: Same here.

poncovka (approved):
Otherwise, the backend changes look good to me. Thanks!

martinpitt:
@OndrejZobal : I suppose you created the references on your own local machine? The cockpit/tasks container (which is what CI uses) indeed does not have any explicit fonts installed, so CJK etc. will not work. cockpit/ws only serves the RedHat fonts, and supposedly they don't cover Cyrillic or CJK.

The way how to avoid/fix that: Either do your development in a [cockpit/tasks](https://quay.io/repository/cockpit/tasks) toolbox -- then you are using *exactly* the same browser versions/fonts/etc. as CI (that's what I do). Or just run integration tests once, and update your reference images to what it spits out, after visual inspection of course.

Note: The first commit already adds reference images, but it really shouldn't. The third commit then updates them. It would be cleaner to only add them in the third commit. Unless of course you plan to squash everything on merging anyway.

velezd:
I'm not sure if it's something you want to do now, but we should definitely embed fonts into the page, otherwise there is no telling how it's going to look in users browser.

ui/webui/src/components/localization/InstallationLanguage.jsx
```diff
@@ -64,163 +70,241 @@ class LanguageSelector extends React.Component {
...
+                  key={prefix + getLocaleId(locale)}
+                  isSelected={isSelected}
+                  itemId={getLocaleId(locale)}
+                  ref={scrollRef}
+                >
+                    {getLocaleNativeName(locale)}
+                </MenuItem>
+            );
+
+            // Prevent assigning scrollRef twice to languages that are bot in common list and the alphabetical list.
```
M4rtinK, line 161: Looks like a typo, "bot" -> "both". :)

M4rtinK (approved):

Just found one typo. :)

martinpitt:
https://github.com/cockpit-project/cockpituous/pull/526 updated our tasks container to include the CJK fonts, thanks @M4rtinK ! It's rolled out, but we still need to watch out for regressions. So please re-run here and confirm that the fonts are alright now?

M4rtinK:
> [cockpit-project/cockpituous#526](https://github.com/cockpit-project/cockpituous/pull/526) updated our tasks container to include the CJK fonts, thanks @M4rtinK ! It's rolled out, but we still need to watch out for regressions. So please re-run here and confirm that the fonts are alright now?

Nice! Fingers crossed the change does not cause some undesirable side effects. :)

@OndrejZobal So I guess we can try dropping that ignore from pixel tests & see if the pictures generated by the CI task container now have the glyphs instead of boxes ? :)

VladimirSlavik (approved):
Looks good to me. Thank you!

PR #4343, opened 2022-09-19
Wait for geolocation and set result in main process

VladimirSlavik:
This makes geolocation completely separated from the UI.

Tested with both GUI and TUI.

Port of #4328 and further changes added.

### DIFF
--- a/pyanaconda/startup_utils.py
+++ b/pyanaconda/startup_utils.py
@@ -31,14 +31,15 @@ from pyanaconda.anaconda_loggers import get_stdout_logger, get_module_logger
 from pyanaconda.core import util, constants
 from pyanaconda.core.configuration.anaconda import conf
 from pyanaconda.core.constants import TEXT_ONLY_TARGET, SETUP_ON_BOOT_DEFAULT, \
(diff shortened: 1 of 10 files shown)


pyanaconda/startup_utils.py
```diff
@@ -646,3 +648,96 @@ def check_if_geolocation_should_be_used(opts):
...
+        log.info("Geoloc: skipping locale because already set")
+        return
+
+    territory = geoloc_result.territory
+    locales = get_territory_locales(territory)
+    try:
+        locale = next(l for l in locales if locale_has_translation(l))
+    except StopIteration:
+        log.info("Geoloc: detected languages are not translated, falling back to %s", DEFAULT_LANG)
+        locale = DEFAULT_LANG
```
poncovka, line 738: Maybe, we could just return in this case. The locale is probably set to `DEFAULT_LANG` anyway in most cases.

poncovka (approved):
Thanks!

VladimirSlavik:
Needs retesting after the early return added above, but not that much.

PR #4344, opened 2022-09-19
Do not require the anaconda-webui package

M4rtinK:
We don't want the package to end up on Rawhide images just yet.

Also drop the use_cockpit macro as it is not really needed now.


VladimirSlavik (approved):
Thanks!

poncovka (approved):

PR #4349, opened 2022-09-20
Web UI: Fix Esperanto breaking webui

OndrejZobal:
Esperanto's locale does not contain a country code, which prevents locale id to be converted to a language id properly causing the UI to fail.
Add another split that makes sure the charset is removed.

### DIFF
--- a/ui/webui/src/components/localization/InstallationLanguage.jsx
+++ b/ui/webui/src/components/localization/InstallationLanguage.jsx
@@ -102,12 +102,9 @@ class LanguageSelector extends React.Component {
         });
     }
 
-    updateDefaultSelection () {
-        const languageId = this.state.lang.split("_")[0];
-        const currentLangLocales = this.state.locales.find(langLocales => getLanguageId(langLocales[0]) === languageId);
-        const currentLocale = currentLangLocales.find(locale => getLocaleId(locale) === this.state.lang);
-
-        this.setState({ selectedItem: getLocaleNativeName(currentLocale) });
+    async updateDefaultSelection () {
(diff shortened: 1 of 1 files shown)


VladimirSlavik:
a) Is there perhaps a javascript library that can parse the locale strings, similar to `langtable.parse_locale`?
b) There are more languages with weird codes, such as Serbian with `sr_Latn_SR` or `sr_SR@Latn`. I wonder if these break too.

At this point, we do know we will be dealing with these, so perhaps it might be better to have a way of dealing with it all?

ui/webui/src/components/localization/InstallationLanguage.jsx
```diff
@@ -103,7 +103,8 @@ class LanguageSelector extends React.Component {
     }
 
     updateDefaultSelection () {
-        const languageId = this.state.lang.split("_")[0];
+        // First split removes the charset and the second removes optional region code.
+        const languageId = this.state.lang.split(".")[0].split("_")[0];
```
poncovka, line 107: I think you should be able to get the data with `getLocaleData` and retrieve the language id with `getLanguageId`.

OndrejZobal:
This patch works with all locales we currently support even Serbian, but I agree that this is still a hacky and potentially unstable solution. I assume this approach was chosen because it's synchronous. I will look into what @poncovka suggested.

OndrejZobal:
> Oh wait. How many times would this be called?
@VladimirSlavik It is called when a language is set.

OndrejZobal:
I will add this as a separate commit to #4342 because I am making changes to the same function there anyway and this fix is not a pressing issue

PR #4351, opened 2022-09-20
Fix duplicate alt-D accelerator on root account screen

jlinton:
The root account screen has duplicate Alt-D accelerators. One for the "Done" button that matches the rest of the anaconda screens and one for the "Disable root account" radio. Lets change that latter option to use alt-i instead so the UI remains consistent that alt-d is "done".

Signed-off-by: Jeremy Linton <jeremy.linton@arm.com>


VladimirSlavik (approved):
Looks good to me. Thank you!

poncovka (approved):

PR #4354, opened 2022-09-22
Fix Web UI VM startup on F37

M4rtinK:
Looks like some virt-install or qemu changes break our current config for starting the Web UI test VM.

The error looks like this:

ERROR    internal error: process exited while connecting to monitor:
2022-09-22T12:01:17.165149Z qemu-system-x86_64:
-device virtio-net-pci,netdev=hostnet0,id=net0,mac=52:54:00:51:4b:77,addr=0x0:
PCI: slot 0 function 0 not available for virtio-net-pci, in use by mch,id=(null)

Basically, it looks like the virtual graphical hardware now clashes with our virtual network hardware on the virtual machine virtual bus.

To fix this, we need to specify an address other than 0x0 for the network hardware, this fixing the issue.


VladimirSlavik (approved):
Looks good to me, thank you!

rvykydal (approved):
Thank you.

PR #4355, opened 2022-09-22
`set_layouts`: don't append empty variants

AdamWill:
This fixes a problem observed in
https://bugzilla.redhat.com/show_bug.cgi?id=2121106 . Kamil set the X layouts to ['cz (qwerty)', 'us']. Then when we called localed's SetX11Keyboard , this is what happened:

Changed X11 keyboard layout to 'cz,us' model '' variant 'qwerty,' options ''

Note the trailing comma on the variant string. That shouldn't be there. What that means is we literally told localed to set the variant string to "qwerty,".

I think the cause is that when we called `parse_layout_variant` for "cz (qwerty)" and then "us", the results would be:

("cz", "qwerty")
("us", "")

so then we'd wind up with `variants` as `["qwerty, ""]`. Then we do `variants_str = ",".join(variants)`, and that gives us the bad string "qwerty," because of the empty string in the list.

To avoid this happening, we should just not add empty strings to the variants list.

### DIFF
--- a/pyanaconda/modules/localization/localed.py
+++ b/pyanaconda/modules/localization/localed.py
@@ -171,7 +171,8 @@ class LocaledWrapper(object):
                 parsing_failed = True
                 continue
             layouts.append(layout)
-            variants.append(variant)
+            if variant:
+                variants.append(variant)
 
         if not layouts and parsing_failed:
             return


VladimirSlavik (approved):
Thank you!

poncovka (changes requested):
It is possible I missed something, but so far I haven't seen any proof that would support your version. For example, `man setxkbmap` says "Multiple variants can be specified as a comma-separated list and will be matched with the layouts specified with `-layout`." The example in the chapter "Even More Advanced Configuration" at https://www.x.org/releases/X11R7.5/doc/input/XKB-Config.html suggests the same thing and uses empty strings as defaults. The guide at https://wiki.archlinux.org/title/Xorg/Keyboard_configuration explicitly says "You must specify as many variants as the number of specified layouts. If you want the default variant, specify an empty string as the variant (the comma must stay)". Based on that, I think that our current implementation is correct.

AdamWill:
ah, thanks, you may well be right. I'll take a look too.

AdamWill:
OK, I agree, I think you're right and this is indeed working how it ought to, so I withdraw this PR. I'll send a follow-up for systemd to correct the line I added there.

PR #4356, opened 2022-09-23
Web UI: Make context help reusable

OndrejZobal:
Context help drawer used to only show storage options help. Added a help toggle function to AnacondaWizzard arguments that can set content of the help and can be used from any page.

### DIFF
--- a/ui/webui/src/components/HelpDrawer.jsx
+++ b/ui/webui/src/components/HelpDrawer.jsx
@@ -14,7 +14,6 @@
  * You should have received a copy of the GNU Lesser General Public License
  * along with This program; If not, see <http://www.gnu.org/licenses/>.
  */
-import cockpit from "cockpit";
 import React, { useRef } from "react";
 
 import {
@@ -25,15 +24,9 @@ import {
     DrawerContentBody,
     DrawerHead,
     DrawerPanelContent,
-    Text,
-    TextContent,
-    TextVariants,
-    Title
 } from "@patternfly/react-core";
 
-const _ = cockpit.gettext;
-
-export const HelpDrawer = ({ isExpanded, setIsExpanded, children }) => {
+export const HelpDrawer = ({ isExpanded, setIsExpanded, helpContent, children }) => {
     const drawerRef = useRef(null);
 
     const onExpand = () => {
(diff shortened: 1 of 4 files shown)


M4rtinK (approved):
Looks good to me, nice and clean! :)

OndrejZobal:
@M4rtinK Is there someone else who should be reviewing webui PRs or should I merge this?

ui/webui/src/components/HelpDrawer.jsx
```diff
@@ -25,15 +24,9 @@ import {
...
-    Text,
-    TextContent,
-    TextVariants,
-    Title
 } from "@patternfly/react-core";
 
-const _ = cockpit.gettext;
-
-export const HelpDrawer = ({ isExpanded, setIsExpanded, children }) => {
+export const HelpDrawer = ({ isExpanded, setIsExpanded, helpContent, children }) => {
```
rvykydal, line 29: It is probably not related to change in this PR, but is there a reason for passing the setIsExpanded setter to the HelpDrawer?

ui/webui/src/components/app.jsx
```diff
@@ -44,6 +44,7 @@ export const Application = () => {
     const [conf, setConf] = useState();
     const [notifications, setNotifications] = useState({});
     const [isHelpExpanded, setIsHelpExpanded] = useState(false);
+    const [helpContent, setHelpContent] = useState("");
```
rvykydal, line 47: This is more a React newbie question that stroke me, you can ignore it.
Would it make sense to keep rather some reference than the element itself in the state? Is it worth effort? If so, it does not have to be part of this PR.

ui/webui/src/components/HelpDrawer.jsx
OndrejZobal: The setter is called with `false` when you click the `X` button inside of the drawer. I think that makes sense, Setting isExpanded directly wouldn't cause rerender.

rvykydal (approved):
Looks good to me, but made me ask some questions...

ui/webui/src/components/HelpDrawer.jsx
rvykydal: Oh sure, I was just blind, it makes sense but I couldn't find it in the diff.
Sorry for the noise...

ui/webui/src/components/app.jsx
OndrejZobal: Ehm, I am not sure what you mean :flushed:

M4rtinK:
> @M4rtinK Is there someone else who should be reviewing webui PRs or should I merge this?

In general one approving review + 24 hours elapsed since PR was opened is good enough to merge. :)

ui/webui/src/components/app.jsx
OndrejZobal: Do you mean that there should be a central data structure for storing all of the JSX objects like a hash map and that the useState should only store the selector?

poncovka: We might not be owners of the help content in some cases (because they might be provided by third-party add-ons). Also, we are not sure if the content will be always static, it could eventually contain some data generated on the spot. Based on that, I would try to keep this simple for now.

PR #4357, opened 2022-09-30
rpm-ostree: set untrusted ostree pull flag

champtar:
By default ostree validate checksum when pulling from remote repos, but not for local repos. Set the untrusted ostree pull flag to always validate files.

### DIFF
--- a/pyanaconda/modules/payloads/payload/rpm_ostree/installation.py
+++ b/pyanaconda/modules/payloads/payload/rpm_ostree/installation.py
@@ -577,7 +577,10 @@ class PullRemoteAndDeleteTask(Task):
         progress = OSTree.AsyncProgress.new()
         progress.connect('changed', self._pull_progress_cb)
 
-        pull_opts = {'refs': Variant('as', [ref])}
+        # We use the UNTRUSTED flag as a way to help verification for potentially
+        # corrupted ISOs - this way ostree will validate the checksum of objects as
+        # it writes data.
+        pull_opts = {'refs': Variant('as', [ref]), 'flags': Variant('i', OSTree.RepoPullFlags.UNTRUSTED)}
         # If we're doing a kickstart, we can at least use the content as a reference:
         # See <https://github.com/rhinstaller/anaconda/issues/1117>
         # The first path here is used by <https://pagure.io/fedora-lorax-templates>
--- a/tests/unit_tests/pyanaconda_tests/modules/payloads/payload/test_rpm_ostree_tasks.py
+++ b/tests/unit_tests/pyanaconda_tests/modules/payloads/payload/test_rpm_ostree_tasks.py
@@ -22,6 +22,7 @@ import pytest
 import unittest
 from unittest.mock import patch, call, MagicMock
 
+from gi.repository import OSTree
 from pyanaconda.core.glib import Variant, GError
 from pyanaconda.modules.common.errors.installation import PayloadInstallationError
 from pyanaconda.modules.common.structures.rpm_ostree import RPMOSTreeConfigurationData
@@ -650,7 +651,7 @@ class PullRemoteAndDeleteTaskTestCase(unittest.TestCase):
         name, args, kwargs = repo_mock.pull_with_options.mock_calls[0]
         opts = args[1]
         assert type(opts) == Variant
-        assert opts.unpack() == {"refs": ["ref"]}
+        assert opts.unpack() == {"refs": ["ref"], "flags": OSTree.RepoPullFlags.UNTRUSTED}
         repo_mock.remote_delete.assert_called_once_with("remote", None)
 
     @patch("pyanaconda.modules.payloads.payload.rpm_ostree.installation.create_new_context")
@@ -679,7 +680,7 @@ class PullRemoteAndDeleteTaskTestCase(unittest.TestCase):
         name, args, kwargs = repo_mock.pull_with_options.mock_calls[0]
         opts = args[1]
         assert type(opts) == Variant
-        assert opts.unpack() == {"refs": ["ref"]}
+        assert opts.unpack() == {"refs": ["ref"], "flags": OSTree.RepoPullFlags.UNTRUSTED}
         repo_mock.remote_delete.assert_not_called()
 
     def test_pull_progress_report(self):


champtar:
For a bit more context, at my job we have an 'appliance' based on Alma 8 / rpm-ostree and out of maybe 200 deployments, we had the ISO corrupted once during transfer, the install still worked, but was unusable.

poncovka:
Hi, it is necessary to update these unit tests:
`PullRemoteAndDeleteTaskTestCase.test_run_success`
`PullRemoteAndDeleteTaskTestCase.test_run_failure`

The `opts.unpack()` call returns `{"refs": ["ref"], "flags": OSTree.RepoPullFlags.UNTRUSTED}`.

champtar:
@poncovka sorry about that, should be fixed now, can you approve the CI again ?

cgwalters:
Not opposed to this, but I think what you want to do instead is run `ostree fsck`.

Also, doesn't Anaconda already have infrastructure to self-validate the ISO integrity?  Are you not using that?

champtar:
> Not opposed to this, but I think what you want to do instead is run `ostree fsck`.

It would be a bit slower as it needs to read back from the disk, instead of just validating data while reading from the iso, and trusting the disk when it says write is ok, but if it's really the preferred way I can do that, I just don't want to have a corrupted install without errors.

> Also, doesn't Anaconda already have infrastructure to self-validate the ISO integrity? Are you not using that?

Right now I'm just appending data and patching a Alma 8 boot iso with mkksiso, I don't think it updates the hash that can be used to validate the iso, and this validation step is opt in, and means reading twice the full ISO which can be slow depending on how the iso is mounted (virtual CD over VPN).

Are the RPMs checksums validated when they are local ? If yes that makes ostree behavior more consistent, if not I'm fine keeping this as a local patch

pyanaconda/modules/payloads/payload/rpm_ostree/installation.py
```diff
@@ -577,7 +577,7 @@ def run(self):
         progress = OSTree.AsyncProgress.new()
         progress.connect('changed', self._pull_progress_cb)
 
-        pull_opts = {'refs': Variant('as', [ref])}
+        pull_opts = {'refs': Variant('as', [ref]), 'flags': Variant('i', OSTree.RepoPullFlags.UNTRUSTED)}
```
cgwalters, line 580: I think we should include the full rationale from the pull request as a comment here.  Something like:

```suggestion
        # We use the UNTRUSTED flag as a way to help verification for potentially
        # corrupted ISOs - this way ostree will validate the checksum of objects as
        # it writes data.
        pull_opts = {'refs': Variant('as', [ref]), 'flags': Variant('i', OSTree.RepoPullFlags.UNTRUSTED)}
```

cgwalters:
> It would be a bit slower as it needs to read back from the disk, instead of just validating data while reading from the iso, and trusting the disk when it says write is ok, but if it's really the preferred way I can do that, I just don't want to have a corrupted install without errors.

Yeah, fair.  It's a tricky balance as this will slow down installs for everyone doing an ISO install to catch low-probability failures.

An entirely different approach here would be for Anaconda to use dm-verity - and as a bonus, this would also cover the integrity of other data, including Anaconda itself.  I've been thinking about doing this for the Fedora CoreOS Live ISO.

> Right now I'm just appending data and patching a Alma 8 boot iso with mkksiso, I don't think it updates the hash that can be used to validate the iso, and this validation step is opt in, and means reading twice the full ISO which can be slow depending on how the iso is mounted (virtual CD over VPN).

Oh wow, virtual CD over VPN sounds flaky.  Any reason you're not doing a "minimal ISO" and fetching the target OS content over HTTP?

> Are the RPMs checksums validated when they are local ? If yes that makes ostree behavior more consistent, if not I'm fine keeping this as a local patch

I think they are actually, so this argues for this patch as is.  

In the end, installation speed doesn't matter *that* much, and gaining a bit of reliability is helpful to ensure the first boot works.

cgwalters:
> fetching the target OS content over HTTP

BTW, in case you didn't see, https://fedoraproject.org/wiki/Changes/OstreeNativeContainer and more recently https://lists.fedoraproject.org/archives/list/devel@lists.fedoraproject.org/thread/4XSCGIDMIFMP3DU6AYQOBZBMIHZ7J75Y/

pyanaconda/modules/payloads/payload/rpm_ostree/installation.py
champtar: Added your comment to my commit

champtar:
> An entirely different approach here would be for Anaconda to use dm-verity - and as a bonus, this would also cover the integrity of other data, including Anaconda itself. I've been thinking about doing this for the Fedora CoreOS Live ISO.

always on validation that doesn't cost too much would be nice

> > Right now I'm just appending data and patching a Alma 8 boot iso with mkksiso, I don't think it updates the hash that can be used to validate the iso, and this validation step is opt in, and means reading twice the full ISO which can be slow depending on how the iso is mounted (virtual CD over VPN).
> 
> Oh wow, virtual CD over VPN sounds flaky. Any reason you're not doing a "minimal ISO" and fetching the target OS content over HTTP?

My use case is a bit exotic, I work for a software editor and 99% of our deployments can't access internet, so everything (base OS, our drivers and software) is in the ISO, and the upgrade package is also the ISO (small script bind mount the ISO, run `ostree pull-local --untrusted`, and finally `rpm-ostree upgrade --cache-only`).
"virtual CD over VPN" is an extreme example, but we deploy on our customers servers using what we can / what we are provided by our customers.

cgwalters:
Disconnected isn't exotic at all.  A really key sub-reason we're doing the "container native" ostree is so that these cases users/customers can mirror the OS updates into a disconnected registry.  

So instead of doing updates via attaching a virtual media ISO, it's just `rpm-ostree upgrade` (or `rebase`) to a container image mirrored into that environment.

But anyways, this PR looks fine to me.  The extra CPU usage might be noticeable in environments like [OpenQA](https://openqa.fedoraproject.org/) but if it matters at some point we can investigate adding something like `inst.ostree.checksums=0` or so.

M4rtinK (approved):
Looks good to me & thanks for the background explanation. :)

VladimirSlavik (approved):
Code-wise LGTM, for actual effects I defer to the experts...

champtar:
> Disconnected isn't exotic at all. A really key sub-reason we're doing the "container native" ostree is so that these cases users/customers can mirror the OS updates into a disconnected registry.
> 
> So instead of doing updates via attaching a virtual media ISO, it's just `rpm-ostree upgrade` (or `rebase`) to a container image mirrored into that environment.

Sadly in my use case it's often 2 servers per site with no extra infrastructure to have an http server or a registry, and no VPN to a central location. We push updates to the servers instead of pulling, when rpm-ostree always pull.

jkonecny12 (approved):

VladimirSlavik:
Sorry I missed this! Tested with the silverblue image, works ok. Merging.

PR #4359, opened 2022-10-03
Don't check installation targets if bootloader devices are not set (#2131183)

poncovka:
If we couldn't set the bootloader devices, the bootloader configuration is invalid and there is not need to verify biosboot partitions on installation targets at that point.

Resolves: rhbz#2131183


M4rtinK (approved):
Looks good to me, thanks! :)

VladimirSlavik (approved):
Looks good to me. Thank you!

poncovka:
This was just a quick fix. Upstream will use a better solution: https://github.com/rhinstaller/anaconda/pull/4365

PR #4361, opened 2022-10-05
Webui discover disks UI

rvykydal:


rvykydal:
Single disk screenshots:
discover_disks_1_disk_1
discover_disks_1_disk_2
discover_disks_1_disk_3

rvykydal:
Three disks screenshots:
discover_disks_3_disks_1
discover_disks_3_disks_2
discover_disks_3_disks_3
discover_disks_3_disks_4

ui/webui/src/components/storage/InstallationDestination.jsx
```diff
@@ -186,6 +199,8 @@ const LocalStandardDisks = ({ idPrefix, setIsFormValid, onAddErrorNotification }
     const [disks, setDisks] = useState({});
     const [refreshCnt, setRefreshCnt] = useState(0);
     const [isDiscoveringDisks, setIsDiscoveringDisks] = useState(false);
+    const [lastDiscoveryDisks, setLastDiscoveryDisks] = useState({});
```
rvykydal, line 202: I am wondering if I should rather use useRef here?

lclay2:
Hi @rvykydal , Thanks for making the changes and for sharing this! I have just a couple of questions/comments:

- Question: When the skeleton transition is presented: 
  -  Is the gray background behind the "bulk selector" checkbox widget the "disabled" state that's offered by the component? I ask because it sort of looks like the checkbox is still enabled, so it wasn't clear to me. 
  - What happens if the user clicks the "Discover disks" button while it's still scanning?
  - Possibility a future change request: I wonder if a skeleton should be presented within each table column versus the entire row. My handoff design used a single row, so it is implemented as designed. :) But now that I view this, I'm wondering about it. It's a detail, but I thought I'd ask about it for future consideration. PF doesn't specify exactly, but I did notice this example: https://www.patternfly.org/v4/components/skeleton/html-demos/skeleton-card/  fwiw.
 
- Comment/Issue?: Some of your screenshots show the "Discover disks" button using what I believe is the hover state style (thicker border on the button) even though it's not being hovered. Is there a bug, or is that just the screenshots?   
https://www.patternfly.org/v4/components/button/react/variant-examples/

rvykydal:
> Hi @rvykydal , Thanks for making the changes and for sharing this! I have just a couple of questions/comments:
> 
>     * Question: When the skeleton transition is presented:
>       
>       * Is the gray background behind the "bulk selector" checkbox widget the "disabled" state that's offered by the component? I ask because it sort of looks like the checkbox is still enabled, so it wasn't clear to me.

It is disabled in the code based on state. Should be more apparent in screencasts I am attaching below (and sorry for not doing it in the first round, I had to resolve issues with screen recording not working in F35, and F36.)

>       * What happens if the user clicks the "Discover disks" button while it's still scanning?

Nothing happens - neither change in UI nor unexpected behavior / actions. At least when making the discovery longer to be able to click on the button by adding sleep of 4 seconds. I believe (hope) this will be the same during the real discovery.

>       * Possibility a future change request: I wonder if a skeleton should be presented within each table column versus the entire row. My handoff design used a single row, so it is implemented as designed. :) But now that I view this, I'm wondering about it. It's a detail, but I thought I'd ask about it for future consideration. PF doesn't specify exactly, but I did notice this example: https://www.patternfly.org/v4/components/skeleton/html-demos/skeleton-card/  fwiw.
> 

I think it is a good idea and it occurred to me as well, looking at the patternfly design recommendations. I didn't go that way mostly because removing the checkboxes during discovery would make change of proportions (cells layout changes) more apparent in case of split skeleton.

So I first tried to keep the checkbox in place, just making it insensitive (empty callback).

Screencast of the version of PR you reviewed: 
https://rvykydal.fedorapeople.org/webui/disks_skeleton/long_skeleton_no_checkbox.webm
Screencast of the version with the checkbox:
https://rvykydal.fedorapeople.org/webui/disks_skeleton/long_skeleton_checkbox.webm
And then tried to split the skeleton:
https://rvykydal.fedorapeople.org/webui/disks_skeleton/split_skeleton_checkbox.webm

I like the last one the most.

>     * Comment/Issue?: Some of your screenshots show the "Discover disks" button using what I believe is the hover state style (thicker border on the button) even though it's not being hovered. Is there a bug, or is that just the screenshots?
>       https://www.patternfly.org/v4/components/button/react/variant-examples/

It is just a focus staying on the button after discovery, can be seen in the screencasts.

Thank you for the review.

lclay2:
Thanks for the screencasts @rvykydal! I know what you're saying about the split skeleton view, it might be that it's more visually appealing because it breaks up the space. But now that I see it in action, I think by-row might more accurately indicate that it's a number of disks, row1 disk, row2 disk, etc. Wdyt? Also I'm assuming that one table cell of data won't be loading before another?

rvykydal:
> Thanks for the screencasts @rvykydal! I know what you're saying about the split skeleton view, it might be that it's more visually appealing because it breaks up the space. But now that I see it in action, I think by-row might more accurately indicate that it's a number of disks, row1 disk, row2 disk, etc. Wdyt? Also I'm assuming that one table cell of data won't be loading before another?

I agree.

Looking at the variants again, with a bit of distance, I don't really have a strong visual preference - maybe the unsplit skeleton (with checkboxes) looks simpler / cleaner.
And yes, regarding the content update logic, perhaps the added value in splitting would be if the split elements loaded in steps - which is not the case. (And maybe if there were some other (more uneven?) structure than the table they would follow).
The unsplit skeleton just suggests there is a list (of disks). Which is (as you indicated if I got it right) what is the most important information obtained by the detection - the number of disks. Not individual cell values - which will stay the same for the already detected disks.

Off the unsplit skeleton variants I prefer the one with checkboxes preserved (it keeps the column headers at place).

ui/webui/src/components/storage/InstallationDestination.jsx
```diff
@@ -318,6 +318,23 @@ const LocalStandardDisks = ({ idPrefix, setIsFormValid, onAddErrorNotification }
         }
     ));
 
+    const discoveringDisksColumns = localDisksColumns.map(col => ({ ...col, sortable: false }));
+
+    const discoveringDisksRow = (
+        [
+            {
+                props: { colSpan: localDisksColumns.length },
+                title: <Skeleton screenreaderText={_("Discovering disks")} />
```
VladimirSlavik, line 351: Done as it is, the text will repeat for every skeleton row, I think. But that's not too important right now.

ui/webui/src/components/storage/InstallationDestination.jsx
```diff
@@ -116,7 +116,8 @@ const DropdownBulkSelect = ({
     onSelectNone,
     onChange,
     selectedCnt,
-    totalCnt
+    totalCnt,
+    isDisabled
```
VladimirSlavik, line 129: Typically these kinds of things are formulated/named positively, as in `isEnabled`, but that's also not important since it's just an internal detail...

M4rtinK (approved):
Looks good to me. :)

VladimirSlavik (approved):
Looks good to me. Thank you!

ui/webui/src/components/storage/InstallationDestination.jsx
rvykydal: Here I followed what we set in the lower lever PF component.

rvykydal:
@M4rtinK @VladimirSlavik I've added two commits implementing the solution of per-row skeleton with the checkbox preserved (https://rvykydal.fedorapeople.org/webui/disks_skeleton/long_skeleton_checkbox.webm) which seems to be the preferred one and changing the button label from "Discover disks" to "Detect disks"
Could you please re-review?

M4rtinK (approved):
Looks good to me, thanks! :)

VladimirSlavik (approved):
Thank you!

PR #4363, opened 2022-10-05
Use correct hint for VNC password boot option

jstodola:
Resolves: rhbz#2125679


VladimirSlavik (approved):
Looks good to me. Thank you!

poncovka (approved):

PR #4365, opened 2022-10-05
There are no installation targets if bootloader devices are not set (#2131183)

poncovka:
If the stage1 and stage2 devices are not set yet (due to an invalid configuration), there are no installation targets to return.

This is an alternative to the commit 1ea4f30.

Resolves: rhbz#2131183

### DIFF
--- a/pyanaconda/modules/storage/bootloader/base.py
+++ b/pyanaconda/modules/storage/bootloader/base.py
@@ -624,7 +624,12 @@ class BootLoader(object):
     @property
     def install_targets(self):
         """List of (stage1, stage2) tuples representing install targets."""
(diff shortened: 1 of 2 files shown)


rvykydal (approved):
Looks good and better than the original solution to me.
Maybe we should mention that both stages should be defined in installation_target items in the install_targets docstring ?

VladimirSlavik (commented):
I don't know this area well enough, but this seems more explicit than the alternative. So I'd go with this.

poncovka:
Rebased.

VladimirSlavik (approved):
Looks good to me. Thank you!

PR #4366, opened 2022-10-06
Change screensaver handling to dasbus, drop safe_dbus

VladimirSlavik:
We need to keep the D-Bus session up and live all the time, or the inhibition stops. So we need to keep the proxy object somehow.


M4rtinK (approved):
Looks good to me. :)

VladimirSlavik:
I set out to rewrite this to follow the form we use in the modules. Turns out that's not a good idea.

I can have a custom bus with custom `_get_connection` that sets the euid, that's fine. So I have a `MessageBus` and `DBusServiceIdentifier`. But all calls on the bus must happen with the right euid again. Overriding `__getitem__` in an `InterfaceProxy` won't work because that is only the getting, not calling. Writing a custom class just to set euid which I will do anyway does not seem like it helps anything... So in the end only the identifier helps, because it deduplicates the DBus strings.

poncovka (approved):
Looks good to me!

poncovka:
> I set out to rewrite this to follow the form we use in the modules. Turns out that's not a feasible idea.
> 
> I can have a custom bus with custom `_get_connection` that sets the euid, that's fine. So I have a `MessageBus` and `DBusServiceIdentifier`. But all calls on the bus must happen with the right euid again. Overriding `__getitem__` in an `InterfaceProxy` won't work because that is only the getting, not calling. Writing a custom class just to set euid which I will do anyway does not seem like it helps anything... So in the end only the identifier helps, because it deduplicates the DBus strings.

Thanks for looking into that :-)

VladimirSlavik:
Yup, it's a shame that did not work :(

PR #4368, opened 2022-10-07
Provide a command line option to install systemd-boot rather than grub2 on x86_64 and arm64

jlinton:
systemd-boot is a lightweight bootloader utility that understands the Linux loader syntax and populates a menu of installed boot options. It then utilizes UEFI-provided system services to run the given image. In the case of x86, arm64, riscv and others the linux kernel can be (is on fedora/RH distros) a standalone EFI executable image. With the initrd= option, the kernel loads its own initrd as well. 

This simplifies the boot process significantly as UEFI already provides a fairly heavyweight environment and signing systemd-boot with platform-enrolled keys avoids much of the complexity of shim+grub loading linux filesystem drivers, console and keyboard mgmt, etc.

This set of patches has been tested on x86_64 and arm64, but it requires either an updated set of systemd packages (to name the bootloader, a stubby (aka systemd grubby) package which provides dnf ownership of some of the /boot/efi/ files, and removal of all the grub packages in the -comps.xml files. Once the system has been installed it should behave mostly as one would expect with 'dnf upgrade' and the like.

It's sorta RFC quality at the moment, as the root= options need some further tweaking, and there remains some stubby work for kdump and the like). I've also not included anything that modifies the default partitioning setups, but in the future, picking this option should remove the need for a dedicated /boot partition.

In theory, systemd-boot can then be used along with the systemd-stub to create unified signed images (as is being done at some large companies today!), but I've not tested anything in that path, and the short term goal is simply to clean up all the bits and pieces so that its possible in the F38/39 timeframe to clean install a fedora system without grub on x86 and arm while maintaining grub as the default bootloader.

### DIFF
--- a/data/anaconda.conf
+++ b/data/anaconda.conf
@@ -138,6 +138,7 @@ selinux = -1
 #
 #   DEFAULT   Choose the type by platform.
 #   EXTLINUX  Use extlinux as the bootloader.
+#   SDBOOT    Use systemd-boot as the bootloader.
 #
 type = DEFAULT
 
--- a/pyanaconda/modules/storage/bootloader/base.py
+++ b/pyanaconda/modules/storage/bootloader/base.py
@@ -176,6 +176,7 @@ class BootLoader(object):
     image_label_attr = "label"
     encryption_support = False
     stage2_is_valid_stage1 = False
+    stage2_required = True
 
     # requirements for stage2 devices
     stage2_device = None
@@ -644,6 +645,10 @@ class BootLoader(object):
             return False
 
         log.debug("Is %s a valid stage2 target device?", device.name)
+
+        if not self.stage2_required:
+            log.debug("stage2 not required")
+            return True
 
         if device.protected:
             valid = False
--- a/pyanaconda/modules/storage/bootloader/efi.py
+++ b/pyanaconda/modules/storage/bootloader/efi.py
@@ -20,15 +20,18 @@ import re
 
 from pyanaconda.modules.storage.bootloader.base import BootLoaderError
 from pyanaconda.modules.storage.bootloader.grub2 import GRUB2
+from pyanaconda.modules.storage.bootloader.systemd import SystemdBoot
 from pyanaconda.core import util
-from pyanaconda.core.kernel import kernel_arguments
+from pyanaconda.core.i18n import _
 from pyanaconda.core.configuration.anaconda import conf
+from pyanaconda.core.kernel import kernel_arguments
+from pyanaconda.core.path import join_paths
 from pyanaconda.product import productName
 
 from pyanaconda.anaconda_loggers import get_module_logger
 log = get_module_logger(__name__)
 
-__all__ = ["EFIBase", "EFIGRUB", "Aarch64EFIGRUB", "ArmEFIGRUB", "MacEFIGRUB"]
+__all__ = ["EFIBase", "EFIGRUB", "Aarch64EFIGRUB", "ArmEFIGRUB", "MacEFIGRUB", "Aarch64EFISystemdBoot"]
 
 
 class EFIBase(object):
@@ -200,6 +203,46 @@ class EFIGRUB(EFIBase, GRUB2):
         super().write_config()
 
 
+class EFISystemdBoot(EFIBase, SystemdBoot):
+    """EFI Systemd-boot"""
+    _packages_common = ["efibootmgr", "systemd-udev", "systemd-boot", "sdubby"]
+    _packages64 = []
+
+    def __init__(self):
+        super().__init__()
+
+        if self.get_fw_platform_size() == '32':
+            # not supported try a different bootloader
+            log.error("efi.py: systemd-boot is not supported on 32-bit platforms")
+            raise BootLoaderError(_("Systemd-boot is not supported on this platform"))
+
+    @property
+    def packages(self):
+        return self._packages64 + self._packages_common
+
+    @property
+    def efi_config_file(self):
+        """ Full path to EFI configuration file. """
+        return join_paths(self.efi_config_dir, self._config_file)
+
+    def write_config(self):
+        """ Write the config settings to config file (ex: grub.cfg) not needed for systemd. """
+        config_path = join_paths(conf.target.system_root, self.efi_config_file)
+
+        log.info("efi.py: (systemd) write_config systemd : %s ", config_path)
+
+        super().write_config()
+
+    def install(self, args=None):
+        log.info("efi.py: (systemd) install")
+        # force the resolution order, we don't want to:
+        #   efibootmgr remove old "fedora"
+        #   or use efiboot mgr to install a new one
+        # lets just use `bootctl install` directly.
+        # which will fix the efi boot variables too.
+        SystemdBoot.install(self)
+
+
 class Aarch64EFIGRUB(EFIGRUB):
     _serial_consoles = ["ttyAMA", "ttyS"]
     _efi_binary = "\\shimaa64.efi"
@@ -207,6 +250,15 @@ class Aarch64EFIGRUB(EFIGRUB):
     def __init__(self):
         super().__init__()
         self._packages64 = ["grub2-efi-aa64", "shim-aa64"]
+
+
+class Aarch64EFISystemdBoot(EFISystemdBoot):
+    _serial_consoles = ["ttyAMA", "ttyS"]
+    _efi_binary = "\\systemd-bootaa64.efi"
+
+    def __init__(self):
+        super().__init__()
+        self._packages64 = []
 
 
 class ArmEFIGRUB(EFIGRUB):
--- a/pyanaconda/modules/storage/bootloader/efi.py
+++ b/pyanaconda/modules/storage/bootloader/efi.py
@@ -147,7 +147,7 @@ class EFIBase(object):
 class EFIGRUB(EFIBase, GRUB2):
     """EFI GRUBv2"""
     _packages32 = [ "grub2-efi-ia32", "shim-ia32" ]
-    _packages_common = [ "efibootmgr", "grub2-tools" ]
+    _packages_common = ["efibootmgr", "grub2-tools", "grub2-tools-extra", "grubby" ]
     stage2_is_valid_stage1 = False
     stage2_bootable = False
 
@@ -249,7 +249,7 @@ class Aarch64EFIGRUB(EFIGRUB):
 
     def __init__(self):
         super().__init__()
-        self._packages64 = ["grub2-efi-aa64", "shim-aa64"]
+        self._packages64 = ["grub2-efi-aa64", "shim-aa64", "grub2-efi-aa64-cdboot"]
 
 
 class Aarch64EFISystemdBoot(EFISystemdBoot):
--- a/pyanaconda/modules/storage/bootloader/efi.py
+++ b/pyanaconda/modules/storage/bootloader/efi.py
@@ -31,7 +31,7 @@ from pyanaconda.product import productName
 from pyanaconda.anaconda_loggers import get_module_logger
 log = get_module_logger(__name__)
 
-__all__ = ["EFIBase", "EFIGRUB", "Aarch64EFIGRUB", "ArmEFIGRUB", "MacEFIGRUB", "Aarch64EFISystemdBoot"]
+__all__ = ["EFIBase", "EFIGRUB", "Aarch64EFIGRUB", "ArmEFIGRUB", "MacEFIGRUB", "Aarch64EFISystemdBoot", "X64EFISystemdBoot"]
 
 
 class EFIBase(object):
@@ -259,6 +259,14 @@ class Aarch64EFISystemdBoot(EFISystemdBoot):
     def __init__(self):
         super().__init__()
         self._packages64 = []
+
+class X64EFISystemdBoot(EFISystemdBoot):
+    _efi_binary = "\\systemd-bootx64.efi"
+
+    def __init__(self):
+        super().__init__()
+        self._packages64 = []
+
 
 
 class ArmEFIGRUB(EFIGRUB):
--- a/pyanaconda/modules/storage/bootloader/efi.py
+++ b/pyanaconda/modules/storage/bootloader/efi.py
@@ -42,6 +42,16 @@ class EFIBase(object):
     def _efi_config_dir(self):
         return "efi/EFI/{}".format(conf.bootloader.efi_dir)
 
+    def get_fw_platform_size(self):
+        try:
+            with open("/sys/firmware/efi/fw_platform_size", "r") as f:
+                value = f.readline().strip()
+        except OSError:
+            log.info("Reading /sys/firmware/efi/fw_platform_size failed, "
+                     "defaulting to 64-bit install.")
+            value = '64'
+        return value
+
     def efibootmgr(self, *args, **kwargs):
         if not conf.target.is_hardware:
             log.info("Skipping efibootmgr for image/directory install.")
@@ -144,14 +154,7 @@ class EFIGRUB(EFIBase, GRUB2):
         super().__init__()
         self._packages64 = [ "grub2-efi-x64", "shim-x64" ]
 
-        try:
-            f = open("/sys/firmware/efi/fw_platform_size", "r")
-            value = f.readline().strip()
-        except OSError:
-            log.info("Reading /sys/firmware/efi/fw_platform_size failed, "
-                     "defaulting to 64-bit install.")
-            value = '64'
-        if value == '32':
+        if self.get_fw_platform_size() == '32':
             self._is_32bit_firmware = True
 
     @property
--- a/pyanaconda/modules/storage/bootloader/systemd.py
+++ b/pyanaconda/modules/storage/bootloader/systemd.py
@@ -1,0 +1,147 @@
+#
+# Copyright (C) 2022 Arm
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# the GNU General Public License v.2, or (at your option) any later version.
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY expressed or implied, including the implied warranties of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General
+# Public License for more details.  You should have received a copy of the
+# GNU General Public License along with this program; if not, write to the
+# Free Software Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
+# 02110-1301, USA.  Any Red Hat trademarks that are incorporated in the
+# source code or documentation are not subject to the GNU General Public
+# License and may only be used or replicated with the express permission of
+# Red Hat, Inc.
+#
+
+from pyanaconda.modules.storage.bootloader.base import BootLoader, BootLoaderError
+from pyanaconda.core import util
+from pyanaconda.core.configuration.anaconda import conf
+from pyanaconda.core.i18n import _
+from pyanaconda.core.path import join_paths
+from pyanaconda.product import productName
+
+from pyanaconda.anaconda_loggers import get_module_logger
+log = get_module_logger(__name__)
+
+__all__ = ["SystemdBoot"]
+
+class SystemdBoot(BootLoader):
+    """Systemd-boot.
+
+    Systemd-boot is dead simple, it basically provides a second boot menu
+    and injects the kernel parms into an efi stubbed kernel, which in turn
+    (optionally) loads it's initrd. As such there aren't any filesystem or
+    drivers to worry about as everything needed is provided by UEFI. Even the
+    console remains on the UEFI framebuffer, or serial as selected by UEFI.
+    Basically rather than trying to be another mini-os (like grub) and duplicate
+    much of what UEFI provides, it simply utilizes the existing services.
+
+    Further, while we could keep stage1 (ESP) and stage2 (/boot) seperate it
+    simplifies things just to merge them and place the kernel/initrd on the
+    ESP. This requires a larger than normal ESP, but for now we assume that
+    this linux installer is creating the partitions, so where the space
+    is allocated doesn't matter.
+
+    """
+    name = "SDBOOT"
+    # oddly systemd-boot files are part of the systemd-udev package
+    # and in /usr/lib/systemd/boot/efi/systemd-boot[aa64].efi
+    # and the systemd stubs are in /usr/lib/systemd/linuxaa64.efi.stub
+    _config_file = "loader.conf"
+    _config_dir = "/loader"
+
+    # systemd-boot doesn't require a stage2 as
+    # everything is stored on the ESP
+    stage2_max_end = None
+    stage2_is_valid_stage1 = True
+    stage2_required = False
+
+    #
+    # configuration
+    #
+
+    @property
+    def config_dir(self):
+        """ Full path to configuration directory. """
+        esp = util.execWithCapture("bootctl", [ "--print-esp-path" ],
+                                   root=conf.target.system_root)
+        return esp.strip() + self._config_dir
+
+    @property
+    def config_file(self):
+        """ Full path to configuration file. """
+        return "%s/%s" % (self.config_dir, self._config_file)
+
+    # copy console update from grub2.py
+    def write_config_console(self, config):
+        log.info("systemd.py: write_config_console")
+        if not self.console:
+            return
+
+        console_arg = "console=%s" % self.console
+        if self.console_options:
+            console_arg += ",%s" % self.console_options
+        self.boot_args.add(console_arg)
+
+    def write_config(self):
+        log.info("systemd.py: write_config systemd start")
+        self.write_config_console(None)
+
+        # Rewrite the loader.conf
+        # For now we are just updating the timeout to actually
+        # implement the bootloader --timeout option
+        config_path = join_paths(conf.target.system_root, self.config_file)
+        log.info("systemd.py: write_config systemd loader conf : %s ", config_path)
+
+        with open(config_path, "w") as config:
+            config.write("timeout "+ str(self.timeout) + "\n")
+            config.write("#console-mode keep\n")
+
+        # update /etc/kernel/cmdline
+        # should look something like "root=UUID=45b931b7-592a-46dc-9c33-d38d5901ec29 ro resume=/dev/sda3"
+        config_path = join_paths(conf.target.system_root, "/etc/kernel/cmdline")
+        log.info("systemd.py: write_config systemd commandline : %s ", config_path)
+        with open(config_path, "w") as config:
+            args = str(self.boot_args)
+            log.info("systemd.py: systemd used boot args: %s ", args)
+
+            # pick up the UUID of the mounted rootfs,
+            root_uuid = util.execWithCapture("findmnt", [ "-sfn", "-oUUID", "/" ],
+                                             root=conf.target.system_root)
+            args += " root=UUID=" + root_uuid
+            config.write(args)
+
+        # rather than creating a mess in python lets just
+        # write the options above, and run a script which will merge the
+        # boot cmdline (after stripping inst. and BOOT_) with the anaconda
+        # settings and the kernel-install recovery/etc options.
+        rc = util.execWithRedirect(
+            "/usr/sbin/updateloaderentries",
+            [" "],
+            root=conf.target.system_root
+        )
+        if rc:
+            raise BootLoaderError(_("Failed to write boot loader configuration. "
+                                    "More information may be found in the log files stored in /tmp"))
+
+    #
+    # installation
+    #
+    def install(self, args=None):
+        log.info("systemd.py: install systemd boot install (root=%s)", conf.target.system_root)
+
+        # the --esp-path= isn't strictly required, but we want to be explicit about it.
+        rc = util.execWithRedirect("bootctl", [ "install", "--esp-path=/boot/efi",
+                                                "--efi-boot-option-description=" + productName.split("-")[0] ],
+                                   root=conf.target.system_root,
+                                   env_prune=['MALLOC_PERTURB_'])
+        if rc:
+            raise BootLoaderError(_("bootctl failed to install UEFI boot loader. "
+                                    "More information may be found in the log files stored in /tmp"))
+
+
+    def write_config_images(self, config):
+        return True
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -256,3 +256,13 @@ Below is a list of pure community features, their community maintainers, and mai
 * Description:
 
 ``Enable boot of the installed system from a BTRFS subvolume.``
+
+systemd-boot as a bootloader
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+* Origin: https://github.com/rhinstaller/anaconda/pull/4368
+* Bugzilla: https://bugzilla.redhat.com/show_bug.cgi?id=2135531
+* Maintainer: Jeremy Linton <jeremy.linton@arm.com>
+* Description:
+
+``Enable boot using systemd-boot rather than grub2.``
--- a/anaconda.spec.in
+++ b/anaconda.spec.in
@@ -39,7 +39,7 @@ Source0: https://github.com/rhinstaller/%{name}/releases/download/%{name}-%{vers
 %define libxklavierver 5.4
 %define mehver 0.23-1
 %define nmver 1.0
-%define pykickstartver 3.44-1
+%define pykickstartver 3.45-1
 %define pypartedver 2.5-2
 %define pythonblivetver 1:3.6.0-1
 %define rpmver 4.15.0
--- a/data/anaconda_options.txt
+++ b/data/anaconda_options.txt
@@ -225,6 +225,11 @@ Use extlinux as the bootloader. Note that there's no attempt to validate
 that this will work for your platform or anything; it assumes that if you
 ask for it, you want to try.
 
+sdboot
+Use systemd-boot as the bootloader. Note that there's no attempt to validate
+that this will work for your platform or anything; it assumes that if you
+ask for it, you want to try.
+
 nombr
 
 If nombr is specified the grub2 bootloader will be installed but the
--- a/docs/boot-options.rst
+++ b/docs/boot-options.rst
@@ -669,6 +669,15 @@ Use extlinux as the bootloader. Note that there's no attempt to validate that
 this will work for your platform or anything; it assumes that if you ask for it,
 you want to try.
 
+.. inst.sdboot:
+
+inst.sdboot
+^^^^^^^^^^^^^
+
+Use systemd-boot as the bootloader. Note that there's no attempt to validate that
+this will work for your platform or anything; it assumes that if you ask for it,
+you want to try.
+
 .. inst.leavebootorder:
 
 inst.leavebootorder
--- a/docs/release-notes/systemd-boot.rst
+++ b/docs/release-notes/systemd-boot.rst
@@ -1,0 +1,20 @@
+:Type: Kickstart Installation
+:Summary: Install an image using systemd-boot rather than grub (#2135531)
+
+:Description:
+    With this release, systemd-boot can be selected as an alternative boot
+    loader for testing and development purposes.
+
+    This can be done with 'inst.sdboot' from the grub/kernel command
+    line or with '--sdboot' in a kickstart file as part of the
+    bootloader command.  The resulting machine should be free of grub,
+    shim, and grubby packages, with all the boot files on the EFI
+    System Partition (ESP). This may mean that it is wise to dedicate
+    the space previously allocated for /boot to the ESP in order to
+    assure that future kernel upgrades will have sufficient space.
+
+    For more information, refer to the anaconda and systemd-boot documentation.
+
+:Links:
+    - https://bugzilla.redhat.com/show_bug.cgi?id=2135531
+    - https://github.com/rhinstaller/anaconda/pull/4368
--- a/dracut/parse-kickstart
+++ b/dracut/parse-kickstart
@@ -54,7 +54,7 @@ from pykickstart.commands.mediacheck import FC4_MediaCheck as MediaCheck
 from pykickstart.commands.driverdisk import F14_DriverDisk as DriverDisk
 from pykickstart.commands.network import F38_Network as Network
 from pykickstart.commands.displaymode import F26_DisplayMode as DisplayMode
-from pykickstart.commands.bootloader import F34_Bootloader as Bootloader
+from pykickstart.commands.bootloader import F39_Bootloader as Bootloader
 
 # Default logging: none
 log = logging.getLogger('parse-kickstart').addHandler(logging.NullHandler())
--- a/pyanaconda/argument_parsing.py
+++ b/pyanaconda/argument_parsing.py
@@ -591,6 +591,8 @@ def getArgumentParser(version_string, boot_cmdline=None):
                     help=help_parser.help_text("noeject"))
     ap.add_argument("--extlinux", action="store_true", default=False,
                     help=help_parser.help_text("extlinux"))
+    ap.add_argument("--sdboot", action="store_true", default=False,
+                    help=help_parser.help_text("sdboot"))
     ap.add_argument("--nombr", action="store_true", default=False,
                     help=help_parser.help_text("nombr"))
     ap.add_argument("--mpathfriendlynames", dest="multipath_friendly_names", action="store_true",
--- a/pyanaconda/core/configuration/anaconda.py
+++ b/pyanaconda/core/configuration/anaconda.py
@@ -375,6 +375,8 @@ class AnacondaConfiguration(Configuration):
         # Set the bootloader type.
         if opts.extlinux:
             self.bootloader._set_option("type", BootloaderType.EXTLINUX.value)
+        if opts.sdboot:
+            self.bootloader._set_option("type", BootloaderType.SDBOOT.value)
 
         # Set the boot loader flags.
         self.bootloader._set_option("nonibft_iscsi_boot", opts.nonibftiscsiboot)
--- a/pyanaconda/core/configuration/bootloader.py
+++ b/pyanaconda/core/configuration/bootloader.py
@@ -23,8 +23,9 @@ from pyanaconda.core.configuration.base import Section
 
 class BootloaderType(Enum):
     """Type of the bootloader."""
-    DEFAULT = "DEFAULT"
+    DEFAULT  = "DEFAULT"
     EXTLINUX = "EXTLINUX"
+    SDBOOT   = "SDBOOT"
 
 
 class BootloaderSection(Section):
@@ -38,6 +39,7 @@ class BootloaderSection(Section):
 
             DEFAULT   Choose the type by platform.
             EXTLINUX  Use extlinux as the bootloader.
+            SDBOOT    Use systemd-boot as the bootloader.
 
         :return: an instance of BootloaderType
         """
--- a/pyanaconda/core/kickstart/commands.py
+++ b/pyanaconda/core/kickstart/commands.py
@@ -25,7 +25,7 @@
 from pykickstart.commands.authselect import F28_Authselect as Authselect
 from pykickstart.commands.autopart import F38_AutoPart as AutoPart
 from pykickstart.commands.autostep import F34_AutoStep as AutoStep
-from pykickstart.commands.bootloader import F34_Bootloader as Bootloader
+from pykickstart.commands.bootloader import F39_Bootloader as Bootloader
 from pykickstart.commands.btrfs import F23_BTRFS as BTRFS
 from pykickstart.commands.cdrom import FC3_Cdrom as Cdrom
 from pykickstart.commands.clearpart import F28_ClearPart as ClearPart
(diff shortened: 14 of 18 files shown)


jlinton:
Oh, I should add that if anyone gets far enough to try this out with libvirt/virtmanager/etc that x86 UEFI machines need to have secure boot disabled otherwise it will just silently fail to execute systemd-boot/etc.

pyanaconda/modules/storage/bootloader/systemd.py
```diff
@@ -0,0 +1,190 @@
...
+log = get_module_logger(__name__)
+
+__all__ = ["SYSTEMD"]
+
+class SYSTEMD(BootLoader):
+    """Systemd-boot.
+
+    """
+    name = "SYSTEMD"
+    # oddly systemd-boot files are part of the systemd-udev package
```
martinezjavier, line 56: it's odd indeed. There has been talks with the systemd folks to split it as a separate sub-package.

pyanaconda/modules/storage/bootloader/systemd.py
```diff
@@ -0,0 +1,190 @@
...
+    stage2_is_valid_stage1 = True
+    stage2_bootable = True
+    stage2_must_be_primary = False
+
+    # requirements for boot devices
+    stage2_device_types = ["partition"]
+
+    def __init__(self):
+        super().__init__()
+        self.encrypted_password = ""
```
martinezjavier, line 72: why this is needed? I don't see being set for other bootloader classes like `ZIPL()`

pyanaconda/modules/storage/bootloader/systemd.py
```diff
@@ -0,0 +1,190 @@
...
+
+    #
+    # configuration
+    #
+
+    @property
+    def config_dir(self):
+        """ Full path to configuration directory. """
+        log.info("systemd.py: config_dir")
+        return "/boot/" + self._config_dir
```
martinezjavier, line 82: Maybe instead of hardcoding this patch you could use `bootctl --print-esp-path` and make `_config_dir = "/loader/"` instead?

That seems to be more aligned to what is expected, since AFAICT is mount point of the ESP + loader sub-dir.

pyanaconda/modules/storage/bootloader/systemd.py
```diff
@@ -0,0 +1,190 @@
...
+        log.info("systemd.py: config_file")
+        return "%s/%s" % (self.config_dir, self._config_file)
+
+    @property
+    def stage2_format_types(self):
+        log.info("systemd.py: stage2_format_types")
+        if productName.startswith("Red Hat "): # pylint: disable=no-member
+            return ["xfs", "ext4", "ext3", "ext2"]
+        else:
+            return ["ext4", "ext3", "ext2", "btrfs", "xfs"]
```
martinezjavier, line 96: Do we even need the concept of stage2 for sd-boot? Since the kernel and initrd are in the ESP, so there's a need for a boot partition.

pyanaconda/modules/storage/bootloader/systemd.py
```diff
@@ -0,0 +1,190 @@
...
+
+
+    def write_device_map(self):
```
martinezjavier, line 99: I don't think you need this function either. Since that's only used by the `GRUB2`  class.

pyanaconda/modules/storage/bootloader/systemd.py
```diff
@@ -0,0 +1,190 @@
...
+
+        # outside of resetting the default, this file should probably be moved to
+        # stubby as well?
+        config = open(config_path, "w")
+        config.write("timeout 3\n")
+        # create a default entry? No need right now..
+        # config.write("default 4142287a09f64398a50ac387bf9e1fca-*\n")
+        config.write("#console-mode keep\n")
+        config.close()
+
```
martinezjavier, line 150: I wonder if all this should be done manually or could just `bootctl install` be invoked ?

pyanaconda/modules/storage/bootloader/systemd.py
```diff
@@ -0,0 +1,190 @@
...
+        log.info("systemd.py: systemd used boot args: %s ", args)
+        # should run this through blkid to get UUID... but what is going on with the above args...
+        args += " root=/dev/mapper/fedora_fedora-root"
+        config.write(args)
+        config.close()
+        # rather than creating a mess in python lets just
+        # write the options above, and run a script which will merge the
+        # boot cmdline (after stripping inst. and BOOT_) with the anaconda settings
+        rc = util.execWithRedirect(
+            "/usr/sbin/updateloaderentries.sh",
```
martinezjavier, line 167: I see that the `ZIPL` class has an ` update_bls_args()` function that does something similar. Maybe there could be a common helper instead of having a separate script?

pyanaconda/modules/storage/bootloader/systemd.py
```diff
@@ -0,0 +1,190 @@
...
+    def installbootmgr(self, args=None):
+        log.info("systemd.py: install")
+        if args is None:
+            args = []
+
+        log.info("systemd.py: install systemd boot install (root=%s)", conf.target.system_root)
+
+        rc = util.execWithRedirect("bootctl", [ "install", "--esp-path=/boot/efi" ],
+                                   root=conf.target.system_root,
+                                   env_prune=['MALLOC_PERTURB_'])
```
martinezjavier, line 189: Ah, I see that you are installing sd-boot using bootctl install. I thought that then there's no need to manually create a loader.conf. At least I didn't have to do it when migrated my machine from grub2 to use sd-boot...

martinezjavier: Do you need to explicitly use `--esp-path=/boot/efi` ? I also didn't have to do it an somehow `bootctl install` figured out that `/boot/efi/` was the mount point for the ESP partition.

pyanaconda/modules/storage/bootloader/efi.py
```diff
@@ -204,6 +254,19 @@ def __init__(self):
         self._packages64 = ["grub2-efi-aa64", "shim-aa64"]
 
 
+class Aarch64EFISystemdBoot(EFISYSTEMD):
+    _serial_consoles = ["ttyAMA", "ttyS"]
+    _efi_binary = "\\systemd-bootaa64.efi"
+
+    def __init__(self):
+        super().__init__()
+        # we can't yet add shim-aa64 to list list because it pulls in grub2-efi-aa64, which breaks install-kernel
```
martinezjavier, line 263: that's strange, I wonder why shim would pull grub2...

pyanaconda/modules/storage/bootloader/efi.py
```diff
@@ -204,6 +254,19 @@ def __init__(self):
...
+        # and in the end I don't think we really want shim anymore
+        # we need to start enrolling the keys in the db
+        #self._packages64 = ["shim-aa64"]
```
martinezjavier, line 266: Agreed, shim should only be needed for a future `x64EFISystemdBoot` class.

pyanaconda/modules/storage/bootloader/efi.py
```diff
@@ -266,6 +266,14 @@ def __init__(self):
         #self._packages64 = ["shim-aa64"]
         self._packages64 = []
 
+class x64EFISystemdBoot(EFISYSTEMD):
```
martinezjavier, line 269: ah, you are adding that future class here :)

data/anaconda.conf
```diff
@@ -138,6 +138,7 @@ selinux = -1
 #
 #   DEFAULT   Choose the type by platform.
 #   EXTLINUX  Use extlinux as the bootloader.
+#   SYSTEMD   Use systemd-boot as the bootloader.
```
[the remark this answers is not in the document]
jstodola: Definitely +1 here, "systemd" is so confusing. The bootloader is called systemd-boot, so why not use this name?

pyanaconda/modules/storage/bootloader/systemd.py
jlinton: IIRC it needs the --esp-path when the /boot and /efi (or was it the efi variables) aren't setup yet. This bit predates the stubby package, which places the entries.srel file needed to get bootctl to place the loader files at the correct location. 

I will retest this to verify that behavior.

pyanaconda/modules/storage/bootloader/systemd.py
jlinton: Again IIRC, the default loader.conf has the timeout parm commented out, which rather than defaulting to three actually means that it defaults to 0.

data/anaconda.conf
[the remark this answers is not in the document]
jlinton: Yah, I guess people could be confused that there is an option to avoid systemd entirely. Lol

No strong opinions here, although I think "sdboot" is probably a better choice than "systemd-boot" for two reasons, one it doesn't have a "-" in the name, which will avoid possible name escaping. To a lesser extent, its less characters I have to type on a serial console when testing :) it.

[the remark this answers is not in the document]
martinezjavier: @jlinton agreed that sdboot is a better name.

pyanaconda/modules/storage/bootloader/systemd.py
```diff
@@ -0,0 +1,192 @@
...
+
+    def installbootmgr(self, args=None):
+        log.info("systemd.py: install")
+        if args is None:
+            args = []
+
+        log.info("systemd.py: install systemd boot install (root=%s)", conf.target.system_root)
+
+        rc = util.execWithRedirect("bootctl", [ "install", "--esp-path=/boot/efi",
+                                                "--efi-boot-option-description=fedora",
```
marcosfrm, line 186: Can be a non-Fedora distro...

https://github.com/rhinstaller/anaconda/blob/anaconda-38.5-1/pyanaconda/modules/storage/bootloader/efi.py#L74-L79

pyanaconda/modules/storage/bootloader/systemd.py
```diff
@@ -0,0 +1,192 @@
...
+                                                "--loader-path=/EFI/fedora/" ],
```
marcosfrm, line 187: ditto

pyanaconda/modules/storage/bootloader/systemd.py
jlinton: Thanks, this should be in the next revision.

pyanaconda/modules/storage/bootloader/systemd.py
jlinton: I've got a local change for this at the moment, but its not clear what the upstream change will be, so I will likely drop the loader-path parts from v2, under the assumption it can be added back when upstream systemd accepts a patch.

pyanaconda/modules/storage/bootloader/systemd.py
jlinton: yes, I've fixed this (and its user) up to pick it up from bootctl, although I've not tried to see what happens if the esp is mounted in /efi since i think we should attempt to retain the RH /boot/efi mountpoint even if /boot ends up mostly empty (although symbol files and the like still tend to be there).

I have been testing it without a dedicated /boot partition now, with an additional tweak to the stage2 code.

pyanaconda/modules/storage/bootloader/systemd.py
jlinton: No, I've added an additional parameter to skip the validation of a dedicated stage2 (because I couldn't thread the stage2 code in a way to avoid it altogether. And of course, that makes this redundant so it has been dropped too.

pyanaconda/modules/storage/bootloader/systemd.py
jlinton: correct is now been dropped to.

pyanaconda/modules/storage/bootloader/systemd.py
jlinton: Well, it turns out that it works fine without the --esp-path=, but the rest of anaconda (probably the packaging, too) can't deal with a /efi as the mount point, so I'm split about removing this vs being explicit about where we want it to install.

I think at the moment (as I prepare a new push) that i'm going to leave it in place.

pyanaconda/modules/storage/bootloader/systemd.py
jlinton: I double-checked this, and yes, I think, for the moment, we want to rewrite it to enable a non-zero timeout.

pyanaconda/modules/storage/bootloader/systemd.py
jlinton: This is the one piece I looked at for a half day or so, but haven't managed to merge together. I think in general you are right, both are doing something quite similar (updating existing BLS entries) but there are a few subtle differences that will need to be merged around how the command line itself is generated. Worse I suspect both have a fundamental error, in that they aren't filtering for newly installed entries which could be a problem.
I think at this point, this is probably the largest remaining issue.

jlinton:
I've updated all bits pointed out above to my own satisfaction, except for the updateloaderentries.sh bit. 
This code is generally working as I expect it at this point. The problematic issues in v1 around the second stage partitioning and root= command line selection have been fixed, and a fair number of cleanups, including most of the ones pointed out above, are in place. 

So, I would take the RFC note off, depending on whether we think the updateloaderentries.sh bit needs to be fixed before merging. (I'm on the fence here).

pyanaconda/modules/storage/bootloader/efi.py
jlinton: I took the comment out, since I think we are in agreement that shim should be out of the picture, so its redundant as shim won't be installed once comps is updated.

martinezjavier (approved):

martinezjavier:
> I've updated all bits pointed out above to my own satisfaction, except for the updateloaderentries.sh bit. This code is generally working as I expect it at this point. The problematic issues in v1 around the second stage partitioning and root= command line selection have been fixed, and a fair number of cleanups, including most of the ones pointed out above, are in place.

Thanks a lot for doing a respin and addressing all the issues I pointed out! It looks good to me and in my opinion we should just merge this. It's great to have this support and since isn't the default, people using it should know what they are doing anyways.

> 
> So, I would take the RFC note off, depending on whether we think the updateloaderentries.sh bit needs to be fixed before merging. (I'm on the fence here).

I honestly don't think this is a big issue. It can be cleaned up later and the need for this script dropped. At the end of the day, is just an implementation detail.

I've approved the changes but I'm not an Anaconda developer, just an occasional contributor. @jkonecny12 @poncovka could you folks please take a look to these patches?

pyanaconda/modules/storage/bootloader/efi.py
```diff
@@ -194,14 +195,80 @@ def write_config(self):
 
         super().write_config()
 
+class EFISYSTEMD(EFIBase, SYSTEMD):
+    """EFI Systemd-boot"""
+    _packages_common = [ "efibootmgr", "systemd-udev", "stubby"]
```
jlinton, line 200: the stubby name here is going to have to change, I was only looking at the server rpms rather than everything, and there is already a stubby package. I think the plan is to open a package request as sdubby.

poncovka:
Please, fix the pylint issues:

```
************* Module pyanaconda.modules.storage.bootloader.systemd
W0223(abstract-method):/tmp/anaconda/pyanaconda/modules/storage/bootloader/systemd.py:51,0: SYSTEMD: Method 'install' is abstract in class 'BootLoader' but is not overridden
W0223(abstract-method):/tmp/anaconda/pyanaconda/modules/storage/bootloader/systemd.py:51,0: SYSTEMD: Method 'write_config_images' is abstract in class 'BootLoader' but is not overridden
W0246(useless-parent-delegation):/tmp/anaconda/pyanaconda/modules/storage/bootloader/systemd.py:68,4: SYSTEMD.__init__: Useless parent or super() delegation in method '__init__'
W0611(unused-import):/tmp/anaconda/pyanaconda/modules/storage/bootloader/systemd.py:36,0: : Unused import os
W0611(unused-import):/tmp/anaconda/pyanaconda/modules/storage/bootloader/systemd.py:37,0: : Unused import re
W0611(unused-import):/tmp/anaconda/pyanaconda/modules/storage/bootloader/systemd.py:42,0: : Unused _ imported from pyanaconda.core.i18n
W0611(unused-import):/tmp/anaconda/pyanaconda/modules/storage/bootloader/systemd.py:43,0: : Unused open_with_perm imported from pyanaconda.core.path
************* Module pyanaconda.modules.storage.bootloader.installation
W0404(reimported):/tmp/anaconda/pyanaconda/modules/storage/bootloader/installation.py:33,0: : Reimport 'conf' (imported line 24)
************* Module pyanaconda.modules.storage.bootloader.efi
W0223(abstract-method):/tmp/anaconda/pyanaconda/modules/storage/bootloader/efi.py:198,0: EFISYSTEMD: Method 'write_config_images' is abstract in class 'BootLoader' but is not overridden
E1101(no-member):/tmp/anaconda/pyanaconda/modules/storage/bootloader/efi.py:225,15: EFISYSTEMD.packages: Instance of 'EFISYSTEMD' has no '_packages64' member
W0223(abstract-method):/tmp/anaconda/pyanaconda/modules/storage/bootloader/efi.py:257,0: Aarch64EFISystemdBoot: Method 'write_config_images' is abstract in class 'BootLoader' but is not overridden
W0223(abstract-method):/tmp/anaconda/pyanaconda/modules/storage/bootloader/efi.py:265,0: x64EFISystemdBoot: Method 'write_config_images' is abstract in class 'BootLoader' but is not overridden

************* Unused False Positives Found:
```

Also, the kickstart support should be implemented here: https://github.com/pykickstart/pykickstart

jkonecny12:
Hi all, if I have correct information (which might be wrong) this will not work on SilverBlue systems. Could you please disable this in `fedora-silverblue.conf` configuration if that is correct?

jlinton:
I'm @jkonecny12 curious about where the problem with silverblue is? If anything, this should help with the immutable rootfs/etc because the entire config/boot process is contained on the ESP, and AFAIK loader updates can/would be via bootctl.

But before anyone gets excited, this commit doesn't replace grub (or any other bootloader, it just provides the framework for testing alternate paths).

martinezjavier:
> I'm @jkonecny12 curious about where the problem with silverblue is? If anything, this should help with the immutable rootfs/etc because the entire config/boot process is contained on the ESP, and AFAIK loader updates can/would be via bootctl.

The issue that Jiri is mentioning is that `ostree` doesn't support `sd-boot`, or more generally, it requires the kernel, initrd and BLS snippets to be in a partition that support symbolic links. Because the current deployment is switched by renaming a symlink.

For context, take a look to https://github.com/ostreedev/ostree/issues/1719 that's the RFE to support `sd-boot` in `ostree`. It's pointed out there that one option could be to create two directories in the ESP and rename exchange them atomically. There's a PR opened to do that https://github.com/ostreedev/ostree/pull/1967.

But the problem was that vfat didn't have support for `renameat2(..., EXCHANGE_REPLACE)`. I've implemented that and landed in Linux v6.0:

https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=019a0c9e377c9f7bd477a0742706d93cdddaee4d
https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=da87e1725ae2136baeb9aac04c572c283afc917f

martinezjavier:
> But before anyone gets excited, this commit doesn't replace grub (or any other bootloader, it just provides the framework for testing alternate paths).

Indeed. It's still an experimental feature so I think is OK if doesn't work for all Fedora variants. I mean, if you install Silverblue with `inst.extlinux` for example I don't expect it to work as well.

jkonecny12:
OK, no problem with leaving it as experimental and potentially broken on SB for now.

cgwalters:
(General reminder that a later goal I have building on top of https://fedoraproject.org/wiki/Changes/OstreeNativeContainerStable is to squash all the weird sub-brands we have and so instead we'd just say "broken on dnf image", which would apply to all of them)

jlinton:
The kickstart pieces are here https://github.com/pykickstart/pykickstart/pull/427, sorta the equivalent of failing to check in a file... since they were in my 'updates' directory.

At the same time, I added a F38_Bootloader option to them, and synced it back to sdboot everywhere.

This update includes the pylint changes suggested above (I'm still not entirely sure how to get a completely clean make check) as well as some trivial changes here/there (dropping the .sh from updateloaders, noting in the documentation the kickstarter option is a bootloader option) which are somewhat the result of the pylint cleanups.

jkonecny12:
Hi @jlinton could you please document this as the pure community feature here: https://github.com/rhinstaller/anaconda/blob/master/CONTRIBUTING.rst#pure-community-features

This is a documentation, that Anaconda team don't have a capacity to maintain this feature so community has to be here to fix the issues in case a bug will raise. The purpose is to help us so we know who to contact and also for community that we won't be giving support for this feature.

pyanaconda/modules/storage/bootloader/systemd.py
Conan-Kudo: We should get this split out before we proceed any further here.

pyanaconda/modules/storage/bootloader/systemd.py
```diff
@@ -0,0 +1,153 @@
...
+        log.info("systemd.py: write_config systemd start")
+        self.write_config_console(None)
+
+        # Rewrite the loader.conf
+        # For now we are just updating the timeout to actually
+        # implement the bootloader --timeout option
+        config_path = "%s%s" % (conf.target.system_root, self.config_file)
+        log.info("systemd.py: write_config systemd loader conf : %s ", config_path)
+
+        config = open(config_path, "w")
```
poncovka, line 102: Please, use the `with` block again.

pyanaconda/modules/storage/bootloader/systemd.py
```diff
@@ -0,0 +1,153 @@
...
+        config.write("timeout "+ str(self.timeout) + "\n")
+        config.write("#console-mode keep\n")
+        config.close()
+
+        # update /etc/kernel/cmdline
+        # should look something like "root=UUID=45b931b7-592a-46dc-9c33-d38d5901ec29 ro resume=/dev/sda3"
+        config_path = "%s%s" % (conf.target.system_root, "/etc/kernel/cmdline")
```
poncovka, line 109: Please, use `join_paths`.

pyanaconda/modules/storage/bootloader/systemd.py
```diff
@@ -0,0 +1,153 @@
...
+        log.info("systemd.py: write_config systemd commandline : %s ", config_path)
+        config = open(config_path, "w")
+        args = " ".join(list(self.boot_args._arguments))
```
poncovka, line 112: You should be able to use `str(self.boot_args)`.

pyanaconda/modules/storage/bootloader/systemd.py
```diff
@@ -0,0 +1,153 @@
...
+        log.info("systemd.py: systemd used boot args: %s ", args)
+
+        # pick up the UUID of the mounted rootfs,
+        root_uuid = util.execWithCapture("findmnt", [ "-sfn", "-oUUID", "/" ],
```
poncovka, line 116: I would expect to get the UUID from the installer, but if it works, I am fine with that.

pyanaconda/modules/storage/bootloader/systemd.py
```diff
@@ -0,0 +1,153 @@
...
+        )
+        if rc:
+            raise BootLoaderError("failed to write boot loader configuration")
+
+    #
+    # installation
+    #
+    def install(self, args=None):
+        log.info("systemd.py: install")
+        if args is None:
```
poncovka, line 138: You can just ignore `args`.

pyanaconda/modules/storage/bootloader/systemd.py
```diff
@@ -0,0 +1,153 @@
...
+        # write the options above, and run a script which will merge the
+        # boot cmdline (after stripping inst. and BOOT_) with the anaconda
+        # settings and the kernel-install recovery/etc options.
+        rc = util.execWithRedirect(
+            "/usr/sbin/updateloaderentries",
+            [" "],
+            root=conf.target.system_root
+        )
+        if rc:
+            raise BootLoaderError("failed to write boot loader configuration")
```
poncovka, line 131: I would encourage you to provide a more detailed message if you can and mark it for translations. You can do that with: `_("This string will be translated")`. This type of an error message is a bad practice and we should finally fix that.

pyanaconda/modules/storage/bootloader/systemd.py
```diff
@@ -0,0 +1,153 @@
...
+
+        log.info("systemd.py: install systemd boot install (root=%s)", conf.target.system_root)
+
+        # the --esp-path= isn't strictly required, but we want to be explicit about it.
+        rc = util.execWithRedirect("bootctl", [ "install", "--esp-path=/boot/efi",
+                                                "--efi-boot-option-description=" + productName.split("-")[0] ],
+                                   root=conf.target.system_root,
+                                   env_prune=['MALLOC_PERTURB_'])
+        if rc:
+            raise BootLoaderError("boot loader install failed")
```
poncovka, line 149: Same here.

poncovka (changes requested):
The code needs a little clean up, but in general it looks good me. Thank you!

pyanaconda/modules/storage/bootloader/efi.py
```diff
@@ -195,13 +196,83 @@ def write_config(self):
         super().write_config()
 
 
+class EFISYSTEMD(EFIBase, SYSTEMD):
```
poncovka, line 199: I would prefer to have `SystemdBoot` instead of `SYSTEMD` in the class names.

pyanaconda/modules/storage/bootloader/systemd.py
jlinton: I'm not sure I understand why. There are a half dozen or so pieces needed to get this working, and then about another half dozen cleanups that can/will follow in order to make everything work well. This isn't one of them, as it tosses a few more complexities into the mix because its one of those places where these changes might break something that is already working (aka people already using systemd boot, if the systemd-udev package isn't then made dependent on a systemd-boot package).

Conan-Kudo: Because we need Anaconda and other things to consistently understand what package name to request when this is being used. It was _always_ wrong for it to be part of `systemd-udev`. If we don't have this fixed before this landed, we are going to have even more problems down the road.

jlinton: ? But that only matters during installation.  Along with the bootctl install flags/location, and a few other things, which presumably _might_ change in the future, it should be fairly easy to update anaconda if/when something like that changes, as it shouldn't get out of sync. 

Maybe I"m misunderstanding. 

Either way, you are probably right that the sooner the switchover in systemd happens (if its to happen at all), the better in order to avoid the problem I noted above.

q2dg:
Now there's a new web UI coming, it may be a good idea to add a combobox/selectbox/button...to let the user choose if he/she wants to install Grub or Systemd-boot (or nothing at all...), like it was in ancient times. Sorry if I'm saying a silliness.

jlinton:
@q2dg You are not wrong, but I think you're getting a little ahead of where we are with this series (and the associated commits). The initial intention is to land this as more of a "testing/expert users" type feature. Then start cleaning up the piles of things required to make it functionally on par with grub for the common use cases. That includes things like figuring out the secure boot chain, agreeing on a partitioning layout, etc.

Until that happens I expect shim/grub, despite the issues, to remain the primary boot choice for fedora/etc users.  So while I personally think anaconda should be doing a better job catering to "advanced users" (which lets face it most people installing linux from a iso are) this feature probably isn't ready for a button in the UI.

pyanaconda/modules/storage/bootloader/systemd.py
jlinton: I'm a bit at a loss for what might cause a failure like this (and would dedicate error messages if I could accurately guess what might cause them, possible ESP capacity/format, or uefi variable storage issues? Or just things missing from the initramfs?)  

This is what I've come up with:

"bootctl failed to install UEFI boot loader. "
"More information may be found in /tmp/program.log,"
"/tmp/storage.log or /tmp/anaconda.log")

pyanaconda/modules/storage/bootloader/systemd.py
jlinton: Well, given there are a bunch of systemd-boot PR's lying about, I added one of my own, slightly different in that it moves both the userspace and the UEFI executables into the same package.

https://src.fedoraproject.org/rpms/systemd/pull-request/95

pyanaconda/modules/storage/bootloader/base.py
```diff
@@ -640,6 +641,10 @@ def is_valid_stage2_device(self, device, linux=True, non_linux=False):
 
         log.debug("Is %s a valid stage2 target device?", device.name)
 
+        if self.stage2_required == False:
```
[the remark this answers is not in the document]
jlinton: I'm think i've resolved all of these, and am going to close out the ones that are trivially verified.

pyanaconda/modules/storage/bootloader/systemd.py
jlinton: I left this alone, IIRC when I initially wrote this, I tried using a UUID provided elsewhere but was having issues, so this was the trivial/self-contained fix.

pyanaconda/modules/storage/bootloader/efi.py
```diff
@@ -195,13 +196,83 @@ def write_config(self):
...
+                     "defaulting to 64-bit install.")
+            value = '64'
+        if value == '32':
+            # not supported try a diff bootloader
+            log.error("efi.py: systemd-boot not supported on 32-bit platform")
+            self._is_32bit_firmware = True
+            raise BootLoaderError("Systemd-boot not supported on this platform")
+
+    @property
+    def _efi_binary(self):
```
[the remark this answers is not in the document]
jlinton: This is gone/fixed too, as are all the renames below/etc

pyanaconda/modules/storage/bootloader/systemd.py
jlinton: This one remained the same, I simply added the "More information may be in the log files stored in /tmp" which hopefully can redirect the user at any failures should they happen. These error messages are large because of the copy/paste execWithRedirect() syle i'm copying where the failure codes are being thrown. I don't think I've seen any of them actually throw when there wasn't a bug during development.

I have put the _() around them though.

pyanaconda/modules/storage/bootloader/systemd.py
jlinton: I'm going to close this one out, mainly because the current code is honoring the anaconda bootloader timeout parameter and passing it into systemd-boot via this config file. Which is IMHO definitely the right thing to be doing and requires modifying the loader.conf file. Whether that is done as writing a new/clean copy or trying to sed/update the one placed there by bootctl I guess could be a discussion, but AFAIU, that wasn't really what your point here was.

pyanaconda/modules/storage/bootloader/systemd.py
jlinton: That discussion seems to be here: https://github.com/systemd/systemd/pull/24828 now that i've removed the arm-specific kernel-install bits, which should be redundant with the compressed EFI/kernel images.

pyanaconda/modules/storage/bootloader/systemd.py
jlinton: I'm going to resolve these too, because I think they have been fixed (the _() for sure).

pyanaconda/modules/storage/bootloader/systemd.py
keszybz: Oh, I understand what happened now. In the PR for systemd there was no mention that is related to Anaconda. The PR was doing something very close to the older PR, so I merged that one instead.

> oddly systemd-boot files are part of the systemd-udev package

Systemds has a bazillion little helpers, and if we tried to split them up by functionality, we'd end up with many tiny packages. Over the years there were a few attempts to split things up, and in the end we ended up with the current form where things are split when there is a benefit. So for example, systemd-udev.rpm has udev, rfkill, various sleep tools, timesyncd, bintfmt, fsck, and even coredump. The reason is that all those things would generally be needed on any standalone machine (either physical or vm), but not in a container. The main systemd package contains anything to boot a machine and maintain it (physical, vm, or container). systemd-journal-remote.rpm contains tools to send journal files over the network, and pulls in additional http and encryption libraries, which we want to avoid by default, hence the separate package. systemd-pam, systemd-libs, systemd-rpm-macros are also separate to not require full systemd installation in various scenarios.

And as mentioned before, various `*ctl` utilities are often in systemd.rpm. They work over the network and on containers, so they are useful without the service part being installed. And they often are useful to query if the service is actually used. Both coredumpctl and bootctl only work locally and wouldn't be useful in non-machine systems.

There's also some work in progress to add support in grub2 to the Boot Loader Interface, so bootctl will be useful even without sd-boot.

None of this is set in stone, and I'm of course happy to discuss changes, but there are reasons for the current state.

pyanaconda/modules/storage/bootloader/systemd.py
```diff
@@ -0,0 +1,146 @@
...
+        self.write_config_console(None)
+
+        # Rewrite the loader.conf
+        # For now we are just updating the timeout to actually
+        # implement the bootloader --timeout option
+        config_path = join_paths(conf.target.system_root, self.config_file)
+        log.info("systemd.py: write_config systemd loader conf : %s ", config_path)
+
+        with open(config_path, "w") as config:
+            config.write("timeout "+ str(self.timeout) + "\n")
```
keszybz, line 99: This would all look so much nicer with f-strings ;--]

pyanaconda/modules/storage/bootloader/systemd.py
jlinton: The real problem isn't this line, so much as where anaconda updates the dependency lists for the install target

https://github.com/jlinton/anaconda/blob/allow_systemd_boot_as_bootloader/pyanaconda/modules/storage/bootloader/efi.py#L206

Without a stable package name (across signed/unsigned/whatever) systemd-boot utilities+bootloaders the bootctl and EFI utilities won't be installed correctly.

keszybz: https://src.fedoraproject.org/rpms/systemd/c/732bdcb223ae95b41e37aa1ba1a3256781a51fbd?branch=rawhide adds
```
Provides:systemd-boot = …
Provides:systemd-boot%{_isa} = …
```

poncovka:
The tests pass with the following changes: https://github.com/poncovka/anaconda/commit/36f73f7b75d4b1bd0fd9ca9e7d41f2cc403ac104
Please, change also `pykickstartver` in the `anaconda.spec.in` file.

pyanaconda/modules/storage/bootloader/systemd.py
```diff
@@ -0,0 +1,146 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General
+# Public License for more details.  You should have received a copy of the
+# GNU General Public License along with this program; if not, write to the
+# Free Software Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
+# 02110-1301, USA.  Any Red Hat trademarks that are incorporated in the
+# source code or documentation are not subject to the GNU General Public
+# License and may only be used or replicated with the express permission of
+# Red Hat, Inc.
+#
+# Systemd-boot is dead simple, it basically provides a second boot menu
```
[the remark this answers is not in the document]
jlinton: yes.

Also, i'm not dead, I was planning on pushing another version today, but the F38 build changes have me rebuilding my test environment.. so hopefully early next week.

pyanaconda/modules/storage/bootloader/systemd.py
jlinton: I'm going to close this one out because I think the main complaints have been addressed, and the latest push adds the systemd-boot dependency.

pyanaconda/modules/storage/bootloader/systemd.py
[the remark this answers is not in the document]
jlinton: I think all these have been fixed in the MR below, so i'm going to resolve them..

jlinton:
I've applied the suggested changes as well as picked up test tweaks noted above, as well as adding systemd-boot to the dependencies list. I also bumped the pykickstart version in anaconda.spec.in in anticipation of the next version, but its probably causing additional failures above because the testing bot can't find a package with that version.

pyanaconda/modules/storage/bootloader/systemd.py
```diff
@@ -0,0 +1,147 @@
...
+    much of what UEFI provides, it simply utilizes the existing services.
+
+    Further, while we could keep stage1 (ESP) and stage2 (/boot) seperate it
+    simplifies things just to merge them and place the kernel/initrd on the
+    ESP. This requires a larger than normal ESP, but for now we assume that
+    this linux installer is creating the partitions, so where the space
+    is allocated doesn't matter.
+
+    """
+    name = "SYSTEMD"
```
keszybz, line 49: Is this supposed to be "SYSTMED" or "SDBOOT" ?

pyanaconda/modules/storage/bootloader/systemd.py
```diff
@@ -0,0 +1,147 @@
...
+log = get_module_logger(__name__)
+
+__all__ = ["SystemdBoot"]
+
+class SystemdBoot(BootLoader):
+    """Systemd-boot.
+
+    Systemd-boot is dead simple, it basically provides a second boot menu
+    and injects the kernel parms into an efi stubbed kernel, which in turn
+    (optionally) loads it's initrd. As such there aren't any filesystem or
```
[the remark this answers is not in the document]
jlinton: Yah I can move it around, but to be pedantic, the kernel-provided stub has this parameter.
https://elixir.bootlin.com/linux/v6.2-rc6/source/Documentation/admin-guide/kernel-parameters.txt#L2031
which is one of the ways to load the initrd with systemd-boot.

pyanaconda/modules/storage/bootloader/systemd.py
```diff
@@ -0,0 +1,147 @@
...
+    drivers to worry about as everything needed is provided by UEFI. Even the
+    console remains on the UEFI framebuffer, or serial as selected by UEFI.
+    Basically rather than trying to be another mini-os (like grub) and duplicate
+    much of what UEFI provides, it simply utilizes the existing services.
+
+    Further, while we could keep stage1 (ESP) and stage2 (/boot) seperate it
+    simplifies things just to merge them and place the kernel/initrd on the
+    ESP. This requires a larger than normal ESP, but for now we assume that
+    this linux installer is creating the partitions, so where the space
+    is allocated doesn't matter.
```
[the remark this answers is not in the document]
jlinton: Sure.

pyanaconda/modules/storage/bootloader/systemd.py
```diff
@@ -0,0 +1,147 @@
...
+        # rather than creating a mess in python lets just
+        # write the options above, and run a script which will merge the
+        # boot cmdline (after stripping inst. and BOOT_) with the anaconda
+        # settings and the kernel-install recovery/etc options.
+        rc = util.execWithRedirect(
+            "/usr/sbin/updateloaderentries",
+            [" "],
+            root=conf.target.system_root
+        )
+        if rc:
```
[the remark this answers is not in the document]
jlinton: Except that bit of python doesn't replace the ability to edit the BLS entries after the machine has been booted (per grubby behavior), nor does it solve many of the other things you commented about in the above defect.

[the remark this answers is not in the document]
jlinton: And I don't think your code is doing what the update script does anyway, which is merge and filter the current boot command line with the existing BLS entries.

pyanaconda/modules/storage/bootloader/systemd.py
```diff
@@ -0,0 +1,147 @@
...
+    # installation
+    #
+    def install(self, args=None):
+        log.info("systemd.py: install systemd boot install (root=%s)", conf.target.system_root)
+
+        # the --esp-path= isn't strictly required, but we want to be explicit about it.
+        rc = util.execWithRedirect("bootctl", [ "install", "--esp-path=/boot/efi",
+                                                "--efi-boot-option-description=" + productName.split("-")[0] ],
+                                   root=conf.target.system_root,
+                                   env_prune=['MALLOC_PERTURB_'])
```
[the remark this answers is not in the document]
jlinton: Hmm, now that is a good one. Its a copy paste job from grub2 install where it noted the dev/mapper gets garbage from the bios device names (see commit: 022b8b6b6ccc42519c3ea5aea68594c5354be2c0) if MALLOC_PERTURB is set. I think I saw this, but with one of early revisions where the BLS rootfs was garbaged up when using /dev/mapper entires. So that code has been replaced now, its probably worth retesting and removing even though its probably harmless to pull out the environment variable.

pyanaconda/modules/storage/bootloader/systemd.py
poncovka: It should be SDBOOT.

poncovka (approved):
The pull request meets all formal requirements, but the functionality requires a package that is not available in Fedora repositories at this moment (see the bug [2134972](https://bugzilla.redhat.com/show_bug.cgi?id=2134972)), so it cannot be fully tested and merged. I will mark it as blocked for now. Since the Beta Freeze of Fedora 38 started yesterday, I am affraid that this feature will have to be postponed to Fedora 39.

- [x] Unit tests and pylint pass.
- [x] Kickstart smoke tests pass.
- [x] Installations with EFIGRUB are bootable.
- [ ] `inst.sdboot` cannot be tested yet.
- [ ] `bootloader --sdboot` cannot be tested yet.

martinezjavier:
> The pull request meets all formal requirements, but the functionality requires a package that is not available in Fedora repositories at this moment (see the bug [2134972](https://bugzilla.redhat.com/show_bug.cgi?id=2134972)), so it cannot be fully tested and merged. I will mark it as blocked for now. Since the Beta Freeze of Fedora 38 started yesterday, I am affraid that this feature will have to be postponed to Fedora 39.
> 
> * [x]  Unit tests and pylint pass.
> * [x]  Kickstart smoke tests pass.
> * [x]  Installations with EFIGRUB are bootable.
> * [ ]  `inst.sdboot` cannot be tested yet.
> * [ ]  `bootloader --sdboot` cannot be tested yet.

Folks, can we please merge this if only for rawhide in the meantime since this is really an experimental feature and @jlinton has been working on it for months already. I'ts not expected to be used by default so IMO resolving https://bugzilla.redhat.com/show_bug.cgi?id=2134972 shouldn't https://bugzilla.redhat.com/show_bug.cgi?id=2134972 for this PR.

Otherwise, the changes could bit rot and cause merge conflicts with would be more work for everyone than just merging as is now.

poncovka:
@martinezjavier @jlinton I talked to the team and they agreed with merging it as an experimental feature, so I will start to work on that. The code will need some small changes to get it ready for F39, but I will handle it.

martinezjavier:
> @martinezjavier @jlinton I talked to the team and they agreed with merging it as an experimental feature, so I will start to work on that. The code will need some small changes to get it ready for F39, but I will handle it.

Perfect, thanks a lot!!

pyanaconda/modules/storage/bootloader/systemd.py
poncovka: Fixed.

poncovka:
Updated and rebased. We should have a functional CI tomorrow, but the tests pass locally.

pyanaconda/modules/storage/bootloader/efi.py
```diff
@@ -197,13 +203,70 @@ def write_config(self):
         super().write_config()
 
 
+class EFISystemdBoot(EFIBase, SystemdBoot):
+    """EFI Systemd-boot"""
+    _packages_common = ["efibootmgr", "systemd-udev", "systemd-boot", "sdubby"]
```
ausil, line 208: sdubby does not exist

jkonecny12: @jlinton should we remove the `sdubby` package or is this something special?

poncovka: The package should be mandatory, but it is still in review (https://bugzilla.redhat.com/show_bug.cgi?id=2134972). See the comments above.

ausil: https://bugzilla.redhat.com/show_bug.cgi?id=2134972
It is under review

keszybz: Formally it is under review. But I think that if the review is supposed to be more than just a rubber-stamp procedure, then IMO this particular package should not pass it. It provides a concentrated dose of bad technical choices and super super brittle workarounds for shortcomings in other packages. The good thing is that _all_ of those other packages are under our control, so we can fix things: anaconda, grub2, kernel-install, kernel scriptlets, systemd-boot. We don't need a combination of rpm packaging, bash, python, and awk to serve as glue between those other packages.

bam80:
Does it support /boot on btrfs subvolume? I didn't succeed on that.

Conan-Kudo:
> Does it support /boot on btrfs subvolume? I didn't succeed on that.

You need a btrfs EFI driver for that, like `btrfs-efi` or `efifs`.

bam80:
> You need a btrfs EFI driver for that, like `btrfs-efi` or `efifs`.

I know, but it still didn't work for me, despite the drivers worked. Anyone had the success?

bam80:
Doesn't work for Atomic Desktops, unfortunately:
https://github.com/rhinstaller/anaconda/discussions/6236

gitnohubz:
> Does it support /boot on btrfs subvolume? I didn't succeed on that.

Isnt systemdboot works without a separate /boot ?

PR #4375, opened 2022-10-14
fix keyboard-inputting character comparison issue

kklionz:
Hi,
    I guess these is a issue while comparing inputting information and I have fixed it.
    Could U help to review it?

### DIFF
--- a/pyanaconda/rescue.py
+++ b/pyanaconda/rescue.py
@@ -582,7 +582,7 @@ class RootSelectionSpoke(NormalTUISpoke):
             keyid = int(key) - 1
         except ValueError:
             # TRANSLATORS: 'c' to continue
-            if key.lower() == C_('TUI|Spoke Navigation', 'c'):
+            if C_('TUI|Spoke Navigation', key.lower()) == C_('TUI|Spoke Navigation', 'c'):
                 self.apply()
                 return InputState.PROCESSED_AND_CLOSE
             else:
--- a/pyanaconda/ui/tui/spokes/askvnc.py
+++ b/pyanaconda/ui/tui/spokes/askvnc.py
@@ -100,7 +100,7 @@ class AskVNCSpoke(NormalTUISpoke):
             return InputState.PROCESSED_AND_CLOSE
         else:
             # TRANSLATORS: 'q' to quit
-            if key.lower() == C_('TUI|Spoke Navigation', 'q'):
+            if C_('TUI|Spoke Navigation', key.lower()) == C_('TUI|Spoke Navigation', 'q'):
                 d = YesNoDialog(_(u"Do you really want to quit?"))
                 ScreenHandler.push_screen_modal(d)
                 if d.answer:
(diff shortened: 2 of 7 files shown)


poncovka:
Btw. I think this was solved in upstream with https://github.com/rhinstaller/anaconda/commit/33fcea788f3113dc2cc3ec6810c6737771d209cf.

kklionz:
This PR is different from https://github.com/rhinstaller/anaconda/commit/33fcea788f3113dc2cc3ec6810c6737771d209cf.
I guess there is a issue happened while anaconda is runing, as following:
1> setup LANG as zh_CN.UTF-8
2> while keyboard 'c' is pressed
     ```key.lower()``` is ```c```
     ```C_('TUI|Spoke Navigation', 'c')``` is ```继续``` ```继续``` is Chinese word that means ```continue```
-----------------------------------------------------------------------------------------------------------------------------------------------------------------

To above case, may is is a more reasonanle process:
Both ```c``` and ```继续``` are input, anaconda shall treat it as continue and process; and this PR is to resolve it.


Btw. Maybe there is a same issue with https://github.com/rhinstaller/anaconda/commit/33fcea788f3113dc2cc3ec6810c6737771d209cf

VladimirSlavik:
To explain why we stopped translating the inputs: Anaconda uses simpleline (another of our projects) for handling the tui input. But some of the inputs are defined in simpleline and some in anaconda. Translations of both components are independent. This made it possible to have the translations define same inputs for different actions (which did happen). Making sure translations do not give conflicting inputs proved impossible, so we decided to just make the inputs untranslateable.

https://bugzilla.redhat.com/show_bug.cgi?id=1892830
https://github.com/rhinstaller/python-simpleline/pull/90#issuecomment-783488826

kklionz:
Hello VladimirSlavik,
-------------------------------------
Your change makes it so that the input can work with both c and 继续. Can you please help us understand why?
-------------------------------------
E.g.
In a Chinese enviorment, When running anaconda in a text mode, A prompt msg will be shown as folling
```请从上方选择['b(开始)'开始安装，'h'帮助，'q'退出, 'r'刷新]```
```开始``` is Chinese word that means ```begin```
```帮助``` is Chinese word that means ```help```
```退出``` is Chinese word that means ```quit```
```刷新``` is Chinese word that means ```refresh```
image


Since this branch is not used any longer, and https://github.com/rhinstaller/anaconda/commit/33fcea788f3113dc2cc3ec6810c6737771d209cf do not allow translations to change this character, this PR will be closed.

VladimirSlavik:
Sorry if the question was not clear. What I thought about was - is it possible to type "c" easily with Chinese keyboard? If not, I would understand why you want to translate it.

kklionz:
Typing 'c' is Chinese```继续``` is not difficult, but it is indeed that it is not easy as English 'c' :)
If anaconda runs with Chinese language environment,  after English 'c' is typed, ```C_('TUI|Spoke Navigation', 'c')``` equals to the Chinese word ```继续```, but ```key.lower()``` is English ```c```,  which leads to that anaconda can not run as expected.
And so, I think ```if key.lower() == C_('TUI|Spoke Navigation', 'c')``` shall be modified as ```if C_('TUI|Spoke Navigation', key.lower()) == C_('TUI|Spoke Navigation', 'c')```

In addition, it improves the user experience: both english ```c```  and local language word ```c``` is accepted as ```continue```.

PR #4376, opened 2022-10-15
Web UI: Add a payload to ks.cfg

OndrejZobal:
Add a default payload for Web UI and expect and change check-progress to expect installation success.

On my machine the whole test takes about 5-6 minutes. I increased the time-out to 10 minutes, hopefully this will be enough.

### DIFF
--- a/ui/webui/test/check-progress
+++ b/ui/webui/test/check-progress
@@ -32,6 +32,9 @@ class TestInstallationProgress(MachineCase):
     MachineCase.machine_class = VirtInstallMachine
 
     def testBasic(self):
+        # HACK Ignore some selinux errors
+        self.allow_journal_messages(".*denied.*comm=\"tar\" name=\"/\".*")
+
         b = self.browser
         i = Installer(b, self.machine)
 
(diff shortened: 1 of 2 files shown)


VladimirSlavik (commented):
Thank you. I think this is a good kind of test to have, but I'm very hesitant to add it to the set of tests that always run. If it raises run time by tens of %, I am not sure it's worth it.

Is it possible to have some mock payload that would be better than empty tar, and thus enable "installation", but faster, because it doesn't have to produce a viable system?

M4rtinK:
> Thank you. I think this is a good kind of test to have, but I'm very hesitant to add it to the set of tests that always run. If it raises run time by tens of %, I am not sure it's worth it.
> 
> Is it possible to have some mock payload that would be better than empty tar, and thus enable "installation", but faster, because it doesn't have to produce a viable system?

That's a good point, but on the other hand we hardly test the progress screen at the moment, so at least from the validity PoV its an improvement. Also IIRC the tests on Cockcpit CI infra run in parallel, so the slowdown should at least have a lower impact.

VladimirSlavik:
Oh, right, this would be a destructive test anyway...

M4rtinK:
@velezd Also I think you had some plans with the test suite once a realistic payload is available ?

velezd:
@M4rtinK yes, once we have working integration tests we can drop check-progress completely. But at the same time we have to somehow split the test suite, as @VladimirSlavik indicated you probably don't want to run all integration tests automatically.

OndrejZobal:
@VladimirSlavik
So what is the consensus? Do we go through with this?

VladimirSlavik (approved):
Thanks!

jkonecny12 (approved):
LGTM!

PR #4377, opened 2022-10-15
Make text in custom_storage_helper more accurate

OndrejZobal:
The window for selecting the disks to be used for a mount point doesn't say clearly that only one of the selected devices will be used. Added an additional label that explains this behaviour.

These changes will be backported to rhel8 and rhel9.

Resolves: [rhbz#2063288](https://bugzilla.redhat.com/show_bug.cgi?id=2063288)

### DIFF
--- a/pyanaconda/ui/gui/spokes/lib/custom_storage_helpers.py
+++ b/pyanaconda/ui/gui/spokes/lib/custom_storage_helpers.py
@@ -409,21 +409,32 @@ class DisksDialog(GUIObject):
     mainWidgetName = "disks_dialog"
     uiFile = "spokes/lib/custom_storage_helpers.glade"
 
-    def __init__(self, data, device_tree, disks, selected_disks):
+    def __init__(self, data, device_tree, disks, selected_disks, is_md):
         super().__init__(data)
         self._device_tree = device_tree
         self._selected_disks = selected_disks
         self._disks = disks
+        self.is_md = is_md
+        self.instruction_label = self.builder.get_object("disk_selection_label")
         self._store = self.builder.get_object("disk_store")
         self._view = self.builder.get_object("disk_view")
         self._populate_disks()
         self._select_disks()
         self._view.set_tooltip_column(0)
+        self._set_instruction_label()
 
     @property
     def selected_disks(self):
         """Selected disks."""
         return self._selected_disks
+
+    def _set_instruction_label(self):
+        if self.is_md:
+            self.instruction_label.set_text(_("Select all of the drives you would like the "
+                                              "mount point to be created on."))
+        else:
+            self.instruction_label.set_text(_("Select a drive for the mount point to be created on. "
+                                              "If you select multiple, only 1 drive will be used."))
 
     def _populate_disks(self):
         for device_name in self._disks:
--- a/pyanaconda/ui/gui/spokes/custom_storage.py
+++ b/pyanaconda/ui/gui/spokes/custom_storage.py
@@ -1179,11 +1179,14 @@ class CustomPartitioningSpoke(NormalSpoke, StorageCheckHandler):
 
         self.reset_state()
 
+        is_md = self._get_current_device_type() == DEVICE_TYPE_MD
+
         dialog = DisksDialog(
             self.data,
             self._device_tree,
             self._selected_disks,
-            self._request.disks
+            self._request.disks,
+            is_md
         )
         with self.main_window.enlightbox(dialog.window):
             rc = dialog.run()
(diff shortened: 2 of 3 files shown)


VladimirSlavik (commented):
Thank you! Code changes look good to me, simple and concise. 👍

This change is about how things look or are perceived. Can you please add a screenshot so reviewers can try to judge that aspect too?

I think it's fine if you just snap a pic of all labels visible and explain which one is displayed only sometimes. Actually, if it looks the same as in glade, that is fine too.

VladimirSlavik:
One more note - @jkonecny12 and @OndrejZobal please maybe assign the bug(s) if it's certain they will be worked on. Alternatively, I have no problem handling the backporting, if needed - your call.

pyanaconda/ui/gui/spokes/lib/custom_storage_helpers.py
```diff
@@ -409,22 +409,31 @@ class DisksDialog(GUIObject):
...
 
     @property
     def selected_disks(self):
         """Selected disks."""
         return self._selected_disks
 
+    def _set_instruction_label(self):
+        if not self.is_md:
+            self.instruction_label.set_text(_("The device will be installed to "
+                                              "one of the selected drives."))
```
jstodola, line 434: I propose to change "The device will be installed to..." to "The mount point will be created on..." - to be more consistent with terms already used in the UI.

jstodola (commented):
For the RAID configuration dialog, could you also add a second line informing the user that the mount point will use all selected drives? It's not clear in the current version.

OndrejZobal:
Old device picker for both standard partitioning and RAID:
old

New device picker for standard partitioning:
normal

New device picker for RAID:
raid

jstodola (approved):
Thanks

VladimirSlavik (approved):
Looks good to me. Thank you!

pyanaconda/ui/gui/spokes/lib/custom_storage_helpers.py
wolfeallison: Hi! I'm a content designer who was asked to take a look! For the content, what do you think of making it a little more concise? Here are a couple of ideas, I'd would love to hear any feedback!
-  "Select all of the drives you would like the mount point to be created on." - wanted to make it user-focused by adding the "you" 
-  "Select 1 or more drives for the mount point to be created on."  - combined the sentences already created to make it a little easier to read and changed "one" to "1" for scalability and consistency (PatternFly recommends using numerals)

jkonecny12: Great to have you here @wolfeallison , thanks!

jkonecny12 (approved):
LGTM!

pyanaconda/ui/gui/spokes/lib/custom_storage_helpers.py
OndrejZobal: Hi @wolfeallison sorry for getting back to you this late.
Thank you very much for your advice! I am going to try to explain the situation a little better. Basically we have two use cases for this dialogue. 
- In the first, all of the selected items will be used for the mount point creation. For this I like what you have suggested: *"Select all of the drives you would like the mount point to be created on."*
- In the second, we need to emphasise that even if the user selects multiple items **only one** of them will end up being chosen to have the mount point installed. I think *"Select 1 or more drives for the mount point to be created on."* does not highlight it enough.

OndrejZobal:
## Standard partitioning:
I am very open to suggestions on this.
Standard-part

## RAID:
Suggested by @wolfeallison .
RAID

pyanaconda/ui/gui/spokes/lib/custom_storage_helpers.py
wolfeallison: @OndrejZobal Oh thank you for the explanations! 

For the second situation, what do you think of "Select a drive for the mount point to be created on. If you select multiple, only 1 drive will be used."

jkonecny12 (approved):
Still looks to me. 

I just wonder, do we really allow user to set bootable device on raid just on some devices from the raid?
Also how this works if it is raid next to the standard partitioning? What we will show?

poncovka:
> I just wonder, do we really allow user to set bootable device on raid just on some devices from the raid? Also how this works if it is raid next to the standard partitioning? What we will show?

I don't understand the questions. This is about a subset of disks that Blivet can consider to use when it tries to find the best location for a new mount point. It has nothing to do with bootable devices.

pyanaconda/ui/gui/spokes/lib/custom_storage_helpers.py
OndrejZobal: @wolfeallison I agree this wording is much better. Thank you Allison!

poncovka (approved):
Looks good to me. Thank you!

PR #4378, opened 2022-10-17
network: use separate main conext for NM client in threads

rvykydal:
Resolves: rhbz#2125256

Create a special NM client with separate main context for calling NM client from installation tasks which run in separate threads.

Based on a pull request by t.feng <t.feng94 at foxmail.com> who deserves the biggest credit, and upated with suggestions by poncovka <vponcova at redhat.com>

The created client should be used only in a limited scope as documented in nm_client_in_thread docstring. If we want to extend it and address potential issues with client instance releasing and reusing we'd need to follow recommendations from Thomas Haller's kind reviews:

<snip>

first of all, initializing a NMClient instance takes relatively long, because it makes D-Bus calls and the round trip time adds up. Btw, if you'd pass
instance_flags=NM.ClientInstanceFlags.NO_AUTO_FETCH_PERMISSIONS it can make it faster, see here. If it's too slow, then the solution would be to re-use the nmclient instance or use async initialization and do stuff in parallel. Both is more complicated however, so not necessary unless we find that it's a problem.

What is maybe more a problem is that each GMainContext consumes at least one file descriptor. When you use the sync nm_client_new() method, then NMClient has an additional internal GMainContext, so possibly there are 2 or more file descriptors involved. The way to "stop" NMClient is by unrefing it. However, with all async operations in glib, they cannot complete right away. That is because when NMClient gets unrefed, it will cancel all (internally) pending operations, but even when you cancel a glib operation, the callback still will be invoked with the cancellation error. And callbacks only get invoked by iterating/running the mainloop/maincontext. This means, if you have a short-running application (e.g. not a GUI) and a reasonable small number of NMClient instances, then you don't need to care. Otherwise, you unfortunately need to make sure that the GMainContext is still iterated just long enough, for all operations to be cancelled. That's slightly cumbersome, and you can use nm_client_get_context_busy_watcher() to find that out.

Btw, what you also cannot do, is having a NMClient instance alive and just not iterating the GMainContext anymore. NMClient will subscribe to D-Bus events, and those come (because GDBus has a separate worker thread) and will be enqueued in the GMainContext. This applies to all applications that register DBus signals via GDBus: you must iterate the context enough, so that those events get eventually processed. I think this does not apply to you here, but it would apply, if you try to keep the nmclient instance alive and reuse later.

</snip>

### DIFF
--- a/pyanaconda/core/glib.py
+++ b/pyanaconda/core/glib.py
@@ -24,34 +24,42 @@
 
 import gi
 gi.require_version("GLib", "2.0")
+gi.require_version("Gio", "2.0")
 
 from gi.repository.GLib import markup_escape_text, format_size_full, \
                                timeout_add_seconds, timeout_add, idle_add, \
                                io_add_watch, child_watch_add, \
-                               source_remove, \
+                               source_remove, timeout_source_new, \
                                spawn_close_pid, spawn_async_with_pipes, \
                                MainLoop, MainContext, \
                                GError, Variant, VariantType, Bytes, \
                                IOCondition, IOChannel, SpawnFlags, \
                                MAXUINT
+from gi.repository.Gio import Cancellable
+
+from pyanaconda.anaconda_loggers import get_module_logger
+log = get_module_logger(__name__)
+
 
 __all__ = ["create_main_loop", "create_new_context",
            "markup_escape_text", "format_size_full",
            "timeout_add_seconds", "timeout_add", "idle_add",
            "io_add_watch", "child_watch_add",
-           "source_remove",
+           "source_remove", "timeout_source_new",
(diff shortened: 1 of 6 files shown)


rvykydal:
A kickstart test batch (network tests) run on the rhel ports of the patch:
https://github.com/rvykydal/kickstart-tests/actions/runs/3293957490
This one contains also reproducer (a bindtomac-network-static-to-dhcp-pre-single flake) in nmclient-rhel9 scenario.

VladimirSlavik (approved):
Looks good to me, thank you.

I assume there is nothing to backport...?

M4rtinK:
Looks like the previously failing widget build is working now as Radek predicted, test all all green now, merging. :)

PR #4387, opened 2022-10-20
Release periodically

VladimirSlavik:
What it says.
Tested piecewise elsewhere.

The one bad thing about cron is that it won't do schedules like "bi-weekly".
Let's talk about the frequency and what days we want this to run etc.

Successful run: https://github.com/VladimirSlavik/anaconda/actions/runs/3291164651/jobs/5424966365


scripts/makebumpver
```diff
@@ -231,6 +228,10 @@ class MakeBumpVer:
 
     def _queryBug(self, bugid):
         if not self.bz:
+            sys.stdout.write("Importing bugzilla module...")
```
jkonecny12, line 231: I don't think we want it logged.

jkonecny12 (approved):
Looks good to me except the nitpick.

scripts/makebumpver
VladimirSlavik: Fair enough. Maybe we could drop more of the logging too...

VladimirSlavik:
Rebased - removed the debug message, added more comments.

PR #4390, opened 2022-10-21
dracut: read filename dhcp option from dhcpopts file

rvykydal:
The dhcp filename option will be added to the .dhcpopts file by NetworkManager.

Related: rhbz#1991516


M4rtinK (approved):
Looks good to me. :)

rvykydal:
> Thank you, LGTM. Is this blocked by an update to NM?

It is not. The feature will not work until NM is updated but the change in Anaconda should not break anything.

poncovka (approved):

PR #4391, opened 2022-10-23
Update welcome.glade

bramgn:
Remove the period at the end of a line on buttons

### DIFF
--- a/pyanaconda/ui/gui/spokes/welcome.glade
+++ b/pyanaconda/ui/gui/spokes/welcome.glade
@@ -20,7 +20,7 @@
             <property name="layout_style">end</property>
             <child>
               <object class="GtkButton" id="quitButton">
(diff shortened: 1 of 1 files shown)


VladimirSlavik (commented):
Hello Bram, thank you for the PR! Unfortunately, we're in string freeze for f37 now, so we can't take it there. There is also no point in changing that after the release, as the downloads won't be updated. So, please make the same change for master, that's always open to changes :-)

Besides that, let's get the commit message into shape. Please don't mention it changes the file, that's clear from the commit itself. Instead, mention the what (now the second line) and also add  _why_.

bramgn:
Hi, Vladimir. Thank you for your feedback! I did want to get back to it, just didn't have the time. I made the same change for master and rewrote the commit message. I partly went with what GitHub automatically filled in (the updating of the file). Hope I got it right this time!

VladimirSlavik:
Thanks! As that's another PR (#4408), I'll close this one.

PR #4392, opened 2022-10-24
infra: Use dill from pip again

VladimirSlavik:
This removes a workaround. Also removes one infra thing to check. Also, might fix the ELN containers. All good news...


jkonecny12 (approved):
LGTM!

PR #4397, opened 2022-10-25
sync_run_task: Poll proxy state faster

VladimirSlavik:
That means, now every 100 ms instead of the previous 1 second. This is possible, because getting the `IsRunning` property is fast on the backend, and the wait/exec time ratio doesn't change much. In return, we get faster return from small tasks.

This is a simple change but has a lot of potential to shake up hidden races, assumptions about timing etc.

cc @AdamWill ^^


poncovka (approved):
Thank you!

PR #4404, opened 2022-10-27
Port templating to RHEL 8

VladimirSlavik:
Same as #4402 but for -8.


rvykydal (approved):
Thank you.

VladimirSlavik:
@jstodola can you please check that this is ok? There's a lot of changes outside the infra definition...

.github/workflows/infra-check.yml
```diff
@@ -71,3 +72,54 @@ jobs:
...
+            git cherry-pick "$COMMIT"
+            make -f Makefile.am reload-infra
+            CHANGES=$(git status -s)
+            if [[ -n $CHANGES ]] ; then
+              echo "Templates out of sync after commit $COMMIT:"
+              git log -1 "$COMMIT"
+              git status
+              exit 1
+            fi
+          done
```
jkonecny12, line 125: Same as in RHEL-9 PR. Merging these could simplify this.

Also I'm thinking about adding `set -eu` to make it more "robust".

jkonecny12 (approved):
Looks good to me.

However, I'm thinking that we should also add:
https://github.com/rhinstaller/anaconda/blob/rhel-8/dockerfile/anaconda-ci/rhel-8.repo
and all the `dockerfile/*/rhel*.repo` because these should point to the correct minor to avoid testing 8.8 on rhel-8.7 branch etc.
It should stay latest on the `rhel-8` or `rhel-9` branches but should change on minor. Probably a new variable or something calculated from the distribution one?

Could be resolved as followup if you want.

jstodola (approved):

PR #4405, opened 2022-10-27
Web UI: Update language selection screen

OndrejZobal:
- Move alert above the menu
 - Selected items now have different backgrounds
 - Replace the input field with a proper search bar

language-selector
*In the screenshot I was hovering my mouse above Japanese.*

I am not able to add the pre-filling of the default language, that @lclay2 suggested, due to technical reasons.

### DIFF
--- a/ui/webui/src/components/localization/InstallationLanguage.jsx
+++ b/ui/webui/src/components/localization/InstallationLanguage.jsx
@@ -28,7 +28,7 @@ import {
     MenuItem,
     MenuContent,
     MenuGroup,
-    TextInput,
+    SearchInput,
     Divider,
     Alert,
 } from "@patternfly/react-core";
@@ -107,8 +107,8 @@ class LanguageSelector extends React.Component {
         this.setState({ commonLocales: await getCommonLocales() });
     }
 
-    async updateNativeName () {
-        const localeData = await getLocaleData({ locale: this.state.lang });
+    async updateNativeName (localeData) {
+        localeData = localeData || await getLocaleData({ locale: this.state.lang });
         this.props.getNativeName(getLocaleNativeName(localeData));
     }
 
@@ -153,6 +153,7 @@ class LanguageSelector extends React.Component {
                   isSelected={isSelected}
                   itemId={getLocaleId(locale)}
                   ref={scrollRef}
+                  style={isSelected ? { backgroundColor: "var(--pf-c-menu__list-item--hover--BackgroundColor)" } : undefined}
                 >
                     {getLocaleNativeName(locale)}
                 </MenuItem>
@@ -234,8 +235,7 @@ class LanguageSelector extends React.Component {
                                 .then(() => setLocale({ locale: getLocaleId(localeItem) }))
                                 .catch(this.props.onAddErrorNotification);
                         this.setState({ ...this.state, lang: item });
-                        this.props.getNativeName(getLanguageNativeName(localeItem));
-
+                        this.updateNativeName(localeItem);
                         window.location.reload(true);
                         return;
                     }
@@ -250,6 +250,7 @@ class LanguageSelector extends React.Component {
         }
 
         const options = this.renderOptions(this.state.search);
+        const inputRef = React.createRef(null);
 
         return (
             <Menu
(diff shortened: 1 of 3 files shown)


ui/webui/src/components/localization/InstallationLanguage.jsx
```diff
@@ -290,21 +316,26 @@ export const InstallationLanguage = ({ idPrefix, setIsFormValid, onAddErrorNotif
                 {_("Select a language")}
             </Title>
             <Form>
-                <FormGroup isRequired>
+                <FormGroup
+                  isRequired
+                >
```
jkonecny12, line 321: This is a weird split of lines. Could we move it on one line instead?

jkonecny12 (approved):
Looks good to me otherwise. Great work!

lclay2:
Thanks @OndrejZobal for working on this! 
@jkonecny12, @M4rtinK - @wolfeallison and I looked at the new screen, and we would like to ask for some minor text string changes if possible. We're concerned that having the header asking the user to "Select a language" then immediately after telling them that a language has been selected, could be confusing. We're not explaining that a language has been preselected by default. If we could modify the "Select a language" and/or info message text a little, we think that will help. :) @wolfeallison will post some microcopy later today for consideration. Thanks again!

VladimirSlavik (approved):
Looks good to me. Thank you!

lclay2:
@OndrejZobal - what do you mean by "Replace the input field with a proper search bar"? Has the behavior of the component changed? Thanks

OndrejZobal:
@lclay2 
> what do you mean by "Replace the input field with a proper search bar"? Has the behaviour of the component changed? Thanks

It's no big deal. Before, i was using a component called TextInput with the magnifying glass icon added in manually and now I have replaced it with SearchInput because it has the clear button built in.

lclay2:
Thanks @OndrejZobal , @wolfeallison and I wanted to confirm the behavior. It sounds like it behaves as it did before, but there are some style changes, thanks!

lclay2:
@OndrejZobal , @M4rtinK, @VladimirSlavik , @jkonecny12  - @wolfeallison will be posting the final microcopy that supports/reflects the recent UI changes. I created this reference image with the new microcopy in place, to help illustrate where the changes will be. 
revised-microcopy

lclay2:
Note: If the user changes the default language selection, then present the existing info alert microcopy.
IfLangChanged

OndrejZobal:
@M4rtinK Cockpit tests are failing because of some npm error

> mv tmp/cockpit/pkg/lib src/
> rm -rf tmp/cockpit
> rm -f package-lock.json #  if it exists already, npm install won't update it; force that so that we always get up-to-date packages
> env -u NODE_ENV npm install #  unset NODE_ENV, skips devDependencies otherwise
> npm ERR! code ETARGET
> npm ERR! notarget No matching version found for @babel/types@^7.20.0.
> npm ERR! notarget In most cases you or one of your dependencies are requesting
> npm ERR! notarget a package version that doesn't exist.

It persisted after after my last force push, could you please look into this?

VladimirSlavik:
@OndrejZobal perhaps this is the evergreen cockpit infra caching issue? They have their own cache for npm since their infra would hammer it pretty hard. The downside is that sometimes, things are not there... and not there yet... and still not...

lclay2:
@M4rtinK , @OndrejZobal , Based on input from @wolfeallison and @jkonecny12 - the final microcopy needed some additional revisions. The following screenshot represents the final microcopy change requests:
final_microcopy
Thanks for your help!

lclay2:
Note: Please disregard the earlier proposal posted on October 28th. The recent microcopy proposal posted today replaces that. In this updated proposal, there is only the one inline alert. Although the selected language string in the alert ["Chosen language: Language(Language)"] will need to update to reflect whatever the language the user choses.

OndrejZobal:
@lclay2 Could you please elaborate on what the "English (English)" should be. I assume that one of them should be the native name of the language and the other an English name but I am not sure which is which.

lclay2:
@OndrejZobal  - I updated the example image. I hope this is clearer. 
CR-lang-notes
It's the same as the menu formatting, which I belief is also the same as the existing installer. 
Screen Shot 2022-11-01 at 8 09 35 AM

lclay2:
@OndrejZobal , yes it looks like I made a mistake, thanks for catching that. @M4rtinK - could you please confirm, I can't find the history of the decision. It is now - the name of the language in its native language, followed by the locale in English.

M4rtinK:
> @OndrejZobal , yes it looks like I made a mistake, thanks for catching that. @M4rtinK - could you please confirm, I can't find the history of the decision. It is now - the name of the language in its native language, followed by the locale in English?

language_naming_GTK_GUI
So this is how it looks like in the "old" GTK GUI - while we do spell out the top level "language group" in English, we do always call the language(variant) using the native name. So 日本語(日本) for Japanese or say English(Denmark) or English(UK) - and IIRC we decided to basically go with whats in the right column in the Web UI & show whats in the left column as grouping label.

In short - the individual names should be:

native language name(native language variant name)

And the grouping labels should be:

native language name (English language name)

And IIRC that's how we have this in the current Web UI implementation.

lclay2:
Thanks so much for explaining and summarizing the history/decision @M4rtinK! :) 
I updated the microcopy example image, if you could take a quick look to make sure it's accurate that would be great. 
* Just a note that there is also a "Common languages" grouping title at the start of the list grouping for common languages, as represented in the mock. 
Nov1-microcopy

jkonecny12:
Still looks good to me. And thanks!

M4rtinK (approved):

PR #4407, opened 2022-10-29
dracut: include modprobed modules into initrd

mikhailnov:
anaconda-modprobe.sh tries to load some modules, but everything relies on them being included into initrd by any other ways, not directly.

It leads to warnings printed on LiveCD startup that modules have not been found. Kernel 5.15.74-generic-1rosa2021.1-x86_64 is built with CONFIG_ISCSI_TCP=m, CONFIG_ISCSI_IBFT=m, but dracut module 90multipath did not include these modules into initrd.

A better way would be to somehow avoid duplicating lists of modules in 2 files, but I don't have ideas how to implement it beautifully.

Before:

After:

### DIFF
--- a/dracut/anaconda-modprobe.sh
+++ b/dracut/anaconda-modprobe.sh
@@ -1,5 +1,6 @@
 #!/bin/bash
 # load modules needed by anaconda
+# keep in sync with module-setup.sh
 
 # load anaconda-lib for the subsequent scripts in this hook
 . /lib/anaconda-lib.sh
--- a/dracut/module-setup.sh
+++ b/dracut/module-setup.sh
@@ -16,9 +16,21 @@ depends() {
 }
 
 installkernel() {
-    case "$(uname -m)" in
+	local arch
+	arch="$(uname -m)"
+	# anaconda-modprobe.sh will load these modules
+	# (instmods in the same order as there)
+	instmods squashfs iscsi_tcp
+    case "$arch" in
         s390*) instmods hmcdrv ;;
+        * ) instmods edd iscsi_ibft ;;
     esac
+    case "$arch" in
+        ppc*) instmods spufs ;;
+    esac
+    instmods raid0 raid1 raid5 raid6 raid456 raid10 linear dm-mod dm-zero \
+             dm-mirror dm-snapshot dm-multipath dm-round-robin dm-crypt cbc \
+             sha256 lrw xts
 }
 
 install() {


jstodola:
I wonder how this can happen in the first place. `anaconda-modprobe.sh` is part of the `anaconda-dracut` RPM, which requires the `dracut-network` RPM, which contains the `95iscsi` module, where [module-setup.sh](https://github.com/dracutdevs/dracut/blob/master/modules.d/95iscsi/module-setup.sh#L164) installs both `iscsi_tcp` and `iscsi_ibft`. Or did I miss anything?

And I also do not like this duplication.

mikhailnov:
> I wonder how this can happen in the first place. `anaconda-modprobe.sh` is part of the `anaconda-dracut` RPM, which requires the `dracut-network` RPM, which contains the `95iscsi` module, where [module-setup.sh](https://github.com/dracutdevs/dracut/blob/master/modules.d/95iscsi/module-setup.sh#L164) installs both `iscsi_tcp` and `iscsi_ibft`. Or did I miss anything?

That's a good question. dracut module `anaconda` does not have a dependency from either `multipath` or `iscsi` dracut modules.  `multipath-tools` was installed, the `check()` function of `90multipath/module-setup.sh` returned true, but `open-iscsi` was not installed and the `check()` function of `95iscsi/module-setup.sh` returned false.

I have to also have `open-iscsi` installed.

mikhailnov:
Do kernel modules really have to be modprobed by anaconda dracut module?

jstodola:
The binaries required by the [check()](https://github.com/dracutdevs/dracut/blob/master/modules.d/95iscsi/module-setup.sh#L6) function from the `95iscsi` module are provided by the `iscsi-initiator-utils` package - so make sure it is installed.

mikhailnov:
iscsi-initiator-utils is called multipath-tools in ROSA Linux where it was run

VladimirSlavik:
@rhinstaller/employees any volunteer to handle this?

mikhailnov:
I am now not sure that it is a good solution, I see at least 3 possible ways to go:

1) this one
2) make anaconda dracut module depend from iscsi and multipath
3) make anaconda dracut module depend from iscsi and multipath and remove this modprobe-releated hackery

It was wrote long time ago, I am not sure that dracut, udev and kernel cannot load these modules themselves. However, I don't have a test iSCSI server, unfortunately. If it can be virtualized, it would be good to document it.

jkonecny12:
I like the third option. Just use module dependencies to get these parts in. However, it needs a lot of discovery and testing to do that step.

PR #4410, opened 2022-10-31
infra: Add runner id to kickstart test statuses

VladimirSlavik:
Let's have statuses with the runner ID, same as cockpit?

See: https://docs.github.com/en/actions/learn-github-actions/contexts#example-usage-of-the-runner-context


rvykydal (approved):
Thank you.

poncovka (approved):
Thank you!

PR #4413, opened 2022-11-02
Initialize empty disks on the Manual Partitioning screen (#2094977)

poncovka:
If a user removes of all partitions of a disk, the disk should be initialized and the installer should create a new disk label on it.

(cherry-picked from a commit 3c597a4)

Resolves: rhbz#2094977


VladimirSlavik (approved):
Thanks!

PR #4415, opened 2022-11-02
Show multiple bootloader devices on the Manual Partitioning screen (#2093793)

poncovka:

was removed.

**Ported from:** https://github.com/rhinstaller/anaconda/pull/4271
Resolves: rhbz#2093793


M4rtinK (approved):
Looks good to me. :)

VladimirSlavik (approved):
Looks good to me. Thank you!

PR #4416, opened 2022-11-02
Show multiple bootloader devices on the Manual Partitioning screen (#1913035)

poncovka:

was removed.

**Ported from:** https://github.com/rhinstaller/anaconda/pull/4271
Resolves: rhbz#1913035

### DIFF
--- a/pyanaconda/modules/storage/devicetree/root.py
+++ b/pyanaconda/modules/storage/devicetree/root.py
@@ -15,6 +15,7 @@
 # License and may only be used or replicated with the express permission of
 # Red Hat, Inc.
 #
+import copy
 import os
 import shlex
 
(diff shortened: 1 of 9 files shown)


M4rtinK (approved):
Looks good to me, thanks! :)

VladimirSlavik (approved):
Looks good to me. Thank you!

pyanaconda/modules/storage/devicetree/root.py
```diff
@@ -353,3 +354,22 @@ def name(self):
...
+        :param InstallerStorage storage: a storage model
+        :return Root: a copy of this root object
+        """
+        new_root = copy.deepcopy(self)
+
+        def _get_device(d):
+            return storage.devicetree.get_device_by_id(d.id, hidden=True)
+
+        def _get_mount(i):
+            m, d = i[0], _get_device(i[1])
```
jkonecny12, line 389: Could you please do the assignment on two lines. It's better to read (at least for me) :D.

jkonecny12 (approved):
Looks good to me. 

Although, I'm a bit concerned about removing the `swap` list. It seems like a good thing to do but I'm worried if code is ready for getting other kind of devices on all the places where `swap` list was used.
I wonder why I missed any of `if device.type == "swap"` kind of code. It's really fine to just get there everything instead of just swap devices?

PR #4418, opened 2022-11-04
webui: do not expose context menu in local Web UI

rvykydal:
The IP address used to identify local installation is set in cockpit-desktop script.

### DIFF
--- a/ui/webui/src/helpers/conf.js
+++ b/ui/webui/src/helpers/conf.js
@@ -17,6 +17,7 @@
 import cockpit from "cockpit";
 
 const CONF_PATH = "/run/anaconda/anaconda.conf";
+const LOCAL_IP_ADDRESS = "127.0.0.90";
 
 const isEmpty = (inStr) => {
     // is string empty?
@@ -78,4 +79,8 @@ export const readConf = () => {
     return confFile.read()
             .then(parseIni)
             .finally(confFile.close);
+};
+
+export const runningLocally = () => {
+    return window.origin.startsWith(`http://${LOCAL_IP_ADDRESS}`);
 };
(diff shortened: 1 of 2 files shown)


rvykydal:
I am using detection of local installation in installer app, alternative solutions could be:
- Turning the context menu off (conext-menu signal) in webview of CockpitClientWindow in cockpit-client.  Potentially depending on Anaconda specific option/profile set in cockpit-client via cockpit-desktop.
- Detection of local installation based on inst.webui.remote boot option. But currently the option is set on the boot.iso images we produce, and both remote cockpit-ws and local cockpit-ws with cockpit-client is run with the option which seems somehow unclear.

M4rtinK:
> * Detection of local installation based on inst.webui.remote boot option. But currently the option is set on the boot.iso images we produce, and both remote cockpit-ws and local cockpit-ws with cockpit-client is run with the option which seems somehow unclear.

Yeah, I agree this would not be a good indicator, especially going forward - even if we keep this option, custom installer image builds still could have it for example even if the actual user still connects locally. Also I think its likely we will drop the option and enable remote access by default once some authentication options (like type in locally displayed auto-generated password) are available.

ui/webui/src/helpers/conf.js
```diff
@@ -79,3 +80,7 @@ export const readConf = () => {
             .then(parseIni)
             .finally(confFile.close);
 };
+
+export const runningLocally = () => {
+    return window.origin.startsWith(`http://${LOCAL_IP_ADDRESS}`);
```
M4rtinK, line 85: I wonder if this could be https as well at some point when running locally? I guess we can fix that when/if it happens, though.

M4rtinK (approved):
Looks good to me, at least for now & for a solution not needing changes in the WebKit wrapper script. :)

M4rtinK (changes requested):
So actually, looking into this more closely - this would fully disable the the context menu which I'm not sure we want to do, as we would loose access to the debug console when running locally.

I guess we could conditionalize it on inst.debug or something like that, but I wonder if it would not just make it hard to use for the people actually wanting to access it.

PR #4422, opened 2022-11-08
infra: Use environment for releases

VladimirSlavik:
Github actions can not push to protected branches because it is not in the list of permitted teams and members. (Currently, it can not be so by design.)

Therefore, use a token generated by someone who can push to protected branches. Of course, the token must be generated by such a person in the first place.

The idea here is that the token is generated for the bot account, @InStallker.

----

NOTE: Not tested. I have no idea how to test this kind of change meaningfully.

### DIFF
--- a/.github/workflows/release-periodically.yml
+++ b/.github/workflows/release-periodically.yml
@@ -27,6 +27,7 @@ permissions:
 jobs:
   make-release-tag:
     runs-on: ubuntu-latest
+    environment: releases
     steps:
 
       - name: Check out repo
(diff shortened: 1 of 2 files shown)


.github/workflows/release-periodically.yml
```diff
@@ -20,13 +20,11 @@ on:
     # Tuesdays are the traditional rawhide anaconda release day, so that there are enough days to
     # unbreak things.
 
-permissions:
-  contents: write
-
```
jkonecny12, line 25: I wonder why you are removing permissions? By doing that you will increase permissions for this workflow and also you need some to Checkout the repository.

jkonecny12 (changes requested):
Looks good to me otherwise.

.github/workflows/release-periodically.yml
VladimirSlavik: Oh. I thought the token permission would override this. But you are right, the token is used only for the one step.

VladimirSlavik: Removed the removal.

"Odvolávám co jsem odvolal..."

PR #4423, opened 2022-11-09
Add special handling of NVMe over Fiber Channel to GUI

VladimirSlavik:
This removes NVMe-FC disks from the local disk list and adds a page with them to Advanced Storage spoke.

Depends on https://github.com/storaged-project/blivet/pull/1077

### DIFF
--- a/pyanaconda/ui/gui/spokes/advanced_storage.py
+++ b/pyanaconda/ui/gui/spokes/advanced_storage.py
@@ -52,14 +52,19 @@ PAGE_SEARCH = 0
 PAGE_MULTIPATH = 1
 PAGE_OTHER = 2
 PAGE_NVDIMM = 3
-PAGE_Z = 4
+PAGE_NVMEFABRICS = 4
+PAGE_Z = 5
+# The Z page must be last = highest number, because it is dynamically removed, which would reorder
+# the items and invalidate the indices hardcoded here.
 
 DiskStoreRow = namedtuple("DiskStoreRow", [
     "visible", "selected", "mutable",
     "name", "type", "model", "capacity",
     "vendor", "interconnect", "serial",
     "wwid", "paths", "port", "target",
-    "lun", "ccw", "wwpn", "namespace", "mode"
+    "lun", "ccw", "wwpn", "namespace", "mode",
+    "controllers", "transport", "transport_address",
+    "subsystem_nqn", "namespace_id"
 ])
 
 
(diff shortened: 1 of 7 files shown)


pyanaconda/ui/gui/spokes/advanced_storage.py
```diff
@@ -494,13 +497,58 @@ def get_selected_namespaces(self):
...
+    SEARCH_TYPE_SUBSYSTEMNQN = 'Subsystem NQN'
+
+    def __init__(self, builder):
+        super().__init__(builder, "nvmefcModel", "nvmefcTypeCombo")
+        self._namespace_entry = self._builder.get_object("nvmefcNamespaceEntry")
+        self._controller_entry = self._builder.get_object("nvmefcControllerEntry")
+        self._subsystemnqn_entry = self._builder.get_object("nvmefcSubsystemNqnEntry")
+
+    def is_member(self, device_type):
+        return device_type == "nvme-fabrics"
```
jstodola, line 513: The condition would also be true for NVMe/TCP devices, which should not be displayed as NVMe/FC devices.
Would it make sense to have one generic class (NVMeoFPage?) for all NVMe-over-fabric devices? In the future we want to add a support for NVMe/TCP, so NVMe/FC and NVMe/TCP devices could possibly be displayed together at the same page.

VladimirSlavik: If it's possible to handle together all "remote" NVMes, that would be good. This might be a terminology issue. I thought Fibre Channel and Fabric is the same thing?

VladimirSlavik: Just cleared it up with Vojta - for the UI it should be called "NVMe (over) Fabrics".

poncovka:
Looks good to me!

> Add nvme-cli to dependencies in some correct place. (img-deps?)

Doesn't Blivet handle that?

> Add nvme-cli to installed packages when using a NVMe-FC disk.

Usually, Blivet provides this kind of installed packages.

VladimirSlavik:
> > Add nvme-cli to dependencies in some correct place. (img-deps?)
> 
> Doesn't Blivet handle that?

I don't think so. We have all the storage stuff in there already.

VladimirSlavik:
I think I have progressed enough to start testing with a real device.

I have no idea about the tests. The failures appear unrelated to the code I changed in the GUI as well as the added package :-/

jkonecny12 (commented):
Looks good to me too.

poncovka:
image
image
image

VladimirSlavik:
Is there a reason to rename everything to fc again? It's Fabrics, not just Fibre Channel. Actually @vojtechtrefny asked for this specifically to be not just Fibre Channel, so that we don't have to rename things later...

poncovka:
> Is there a reason to rename everything to fc again? It's Fabrics, not just Fibre Channel. Actually @vojtechtrefny asked for this specifically to be not just Fibre Channel, so that we don't have to rename things later...

Ok, I will rename it. There was a mismatch of the names in the code, so I have just chosen one of them.

poncovka:
Based on the discussion about NVMe-FC attributes in Anaconda, it was decided to implement the following functionality. Anaconda will show the following columns for NVMe-FC devices:

* Namespace = nsid of the device
* Controllers = ids of the device controllers
* Transport = transport types
* Transport address = transport addresses
* Subsystem NQN = subsysnqns of the device controllers
* Namespace ID = eui64, nguid and uuid of the device

Anaconda will allow to filter NVMe-FC devices by:

* Controller ID
* Transport type
* Transport address
* Subsystem NQN
* EUI-64 / NGUID / UUID

Columns in the NVMe-FC tab (part 1), the transport addresses are supposed to be weird:
image

Columns in the NVMe-FC tab (part 2), the namespace id is not supposed to be empty, @vojtechtrefny is working on that:
image

Filtering by the controller id:
image

Filtering by the transport type and the transport address. The transport type can be undefined:
image

Filtering by the subsysnqn:
image

Filtering by EUI-64 or NGUID or UUID:
image

tbzatek:
Cc: @johnmeneghini

M4rtinK (approved):
Looks good to me, thanks! :)

VladimirSlavik (commented):
LGTM, thanks! Shame I can't somehow transfer ownership of the PR, this ways I can't make the green mark.

jkonecny12 (approved):

jkonecny12:
Thanks a lot for handling this!

PR #4424, opened 2022-11-11
webui: reduce progress animation repaints to save CPU (rhbz#1204242)

rvykydal:
We handled the spinner in the same way in Gtk GUI:

commit eead6838d69699a7f329305c119857301eab5122

Ideally, we should apply the reduction only if running
- in VM
- locally in cocpit-client/webkit

We need to figure out sane detection of these contidions, maybe passed to webui via anaconda configuration if possible ?

**NOTE** the recommendation from the desktop team is to use Firefox in kiosk mode.

### DIFF
--- a/ui/webui/src/components/animations-workaround.css
+++ b/ui/webui/src/components/animations-workaround.css
@@ -1,0 +1,33 @@
+/*
+ * Animations in cockpit-client / webkit in VMs are causing excessive CPU load.
+ * See rhbz#1204242.
+ * Workaround it by decreasing the frame rate.
+ */
+
+.pf-c-skeleton {
+  --pf-c-skeleton--after--AnimationTimingFunction: steps(4);
+}
+
+span.pf-c-spinner {
+  --pf-c-spinner--AnimationTimingFunction: steps(0);
+  --pf-c-spinner__path--AnimationTimingFunction: steps(4);
+}
+
+svg.pf-c-spinner {
+  --pf-c-spinner--AnimationTimingFunction: steps(0);
+  --pf-c-spinner__path--AnimationTimingFunction: steps(4);
+}
+
+/* These spinners look too bad with reduced frame ratio and are less exposed
+.pf-c-spinner__clipper {
+  animation-timing-function: steps(4);
+}
+
+.pf-c-spinner__lead-ball {
+  animation-timing-function: steps(4);
+}
+
+.pf-c-spinner__tail-ball {
+  animation-timing-function: steps(4);
+}
+*/
--- a/ui/webui/src/index.js
+++ b/ui/webui/src/index.js
@@ -31,6 +31,7 @@ import { Application } from "./components/app.jsx";
  */
 import "./lib/patternfly/patternfly-4-overrides.scss";
 import "./components/app.scss";
+import "./components/animations-workaround.css";
 
 document.addEventListener("DOMContentLoaded", function () {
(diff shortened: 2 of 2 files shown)


rvykydal:
Video of remote access via Firefox:

Skeleton is reduced in "Detect disks".
EmptyStateIcon in "Checking disks" is not updated in this PR (commented in) and would need some finer reduction.
Spinner is reduced in Installation Progress - (actually looks still quite cool IMO)

[animations_reduced.webm](https://user-images.githubusercontent.com/1220237/201334168-4aa47f01-77ce-4bab-8a7c-679dc429a493.webm)

rvykydal:
I've prepared updates.img image for our webui boot.iso containing simple application to play with in Instpector to fine-tune the animations:
https://rvykydal.fedorapeople.org/updates.webui-1s.img
test_app

It is located in /root/spinner/build

jstodola:
Radku, is this just a temporary workaround? Because ideally anaconda should not have to deal with such problems at all and it should be solved on cocpit-client/webkit side (or whatever is so slow).
There is no excuse for cocpit-client/webkit to eat ~150% of CPU time on a 2 CPU VM to perform a simple animation. The same VM can play 1080p youtube video in Firefox with 100% of CPU time.

rvykydal:
> **NOTE** the recommendation from the desktop team is to use Firefox in kiosk mode.

In case of using Firefox the load seems to be just moved to gnome-shell.

rvykydal:
I filed a BZ for WebKit2Gtk3: https://bugzilla.redhat.com/show_bug.cgi?id=2142779

ui/webui/src/components/animations-workaround.css
```diff
@@ -0,0 +1,33 @@
+/*
+ * Animations in cockpit-client / webkit in VMs are causing excessive CPU load.
+ * See rhbz#1204242.
+ * Workaround it by decreasing the frame rate.
+ */
+
+.pf-c-skeleton {
+  --pf-c-skeleton--after--AnimationTimingFunction: steps(4);
```
poncovka, line 8: Have you tried 24 steps as in https://github.com/rhinstaller/anaconda/commit/eead6838d69699a7f329305c119857301eab5122? It is that magical threshold when the human eye stops to see pictures and starts to see an animation. It is a kind of a standard for animations.

rvykydal: It was eating still to much CPU. I'll bring some numbers.

rvykydal: VM with 8 GB RAM, 4 CPUs:
```
              WebKitWebProcess       gnome-kiosk
original           160 (peak 210)       60
steps(4)            30                  15
steps(24)          115 (peak 130)       50
```

poncovka: Interesting. Anyway, thanks for the data!

ui/webui/src/index.js
```diff
@@ -31,6 +31,7 @@ import { Application } from "./components/app.jsx";
  */
 import "./lib/patternfly/patternfly-4-overrides.scss";
 import "./components/app.scss";
+import "./components/animations-workaround.css";
```
KKoukiou, line 34: This is for the progress component, right? 
Should be added in src/components/installation/InstallationProgress.scss or at leat in src/components/installation/ folder.

PR #4425, opened 2022-11-16
Fix infobar colors in GTK GUI

OndrejZobal:
Fix infobar CSS after GTK update broke it.

Resolves: [rhbz#2074827](https://bugzilla.redhat.com/show_bug.cgi?id=2074827)

## Screenshots
This is how it looks now! By the way I didn't change any of the color values.
### Warning
warning

### Error
Screenshot_20221116_122207

### RHEL 8
rhel8-screenshot

The info bar doesn't look 100% like in the RHEL 8. For example the icon is now monochrome and the hyperlink has a different color, the original is light grey.

Right now I just fixed the old CSS so that it now applies again. Is this fine or should I make it look 1:1 to the RHEL-8 screenshot or are we going to use a different color scheme?

### DIFF
--- a/data/anaconda-gtk.css
+++ b/data/anaconda-gtk.css
@@ -46,35 +46,35 @@ levelbar.discrete trough block.filled.high {
 @define-color error_fg_color black;
 @define-color error_bg_color rgb (237, 54, 54);
 
-infobar.info {
+infobar.info box {
     background-color: @info_bg_color;
-    border-color: darker(@info_bg_color);
+    border: none;
 }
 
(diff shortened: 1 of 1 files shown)


poncovka (approved):
Looks good to me. Thank you!

jkonecny12 (approved):
Looks great to me.

Please ignore the icon. The icon set has changed in Rawhide after RHEL-8 so it's probably because of that.

VladimirSlavik (approved):
Looks good to me. Thank you!

PR #4438, opened 2022-11-23
Fix infobar colors in GTK GUI rebase to RHEL-9

OndrejZobal:
Fix infobar CSS after GTK update broke it.

Resolves: rhbz#2074827

rhel9-patched-warn
rhel9-patched


poncovka (approved):
Looks good to me. Thank you!

jkonecny12 (approved):

PR #4440, opened 2022-11-23
Make text in custom_storage_helper more accurate rebase to RHEL-8

OndrejZobal:
The window for selecting the disks to be used for a mount point doesn't say clearly that only one of the selected devices will be used. Added an additional label that explains this behaviour.

Resolves: rhbz#2063288


poncovka (approved):
Thanks!

jkonecny12 (approved):

VladimirSlavik (approved):
Thanks!

PR #4450, opened 2022-11-29
makebumpver: ignore commits with (#test) in summary (#infra)

rvykydal:
The #test should be used to mark changes to tests, which are not part of the rpm currently. Therefore they don't need a rhbz# reference.

Port from rhel-9 branch.


poncovka (approved):

VladimirSlavik (approved):

M4rtinK (approved):

PR #4451, opened 2022-11-29
Add a block of history about the name of the installer

poncovka:
This is what I can recall from the time period as well as things learned later after coming to Red Hat.  I was working on Slackware Linux at the time for Walnut Creek CDROM but Red Hat was very much something we paid attention to.

**Reopened from**: https://github.com/rhinstaller/anaconda/pull/4428


jkonecny12 (approved):
LGTM!

PR #4463, opened 2022-12-06
add response of SIGINT with GUI and TUI

iasunsea:
when we do install, especially use TUI to install, we some time have take a mistake to put "CTRL+C", then there will be stop with traceback. And we know the main process have  shielded SIGINT, so we to shield subprocesses also.


VladimirSlavik (commented):
Hello and thank you for the contribution, @iasunsea!

For the code itself, I think that's an improvement.

However, it would be good to have only one commit, and we need useful title and description for any commits. You can use your PR comment, that's a good description. So something like:

```
Ignore SIGINT in D-Bus launcher and x11 too

When we do install, especially use TUI to install, we some time have take
a mistake to put "CTRL+C", then there will be stop with traceback. And we
know the main process have shielded SIGINT, so we to shield subprocesses also.
```

If you wish, I can change the PR to take this form.

iasunsea:
> Hello and thank you for the contribution, @iasunsea!
> 
> For the code itself, I think that's an improvement.
> 
> However, it would be good to have only one commit, and we need useful title and description for any commits. You can use your PR comment, that's a good description. So something like:
> 
> ```
> Ignore SIGINT in D-Bus launcher and x11 too
> 
> When we do install, especially use TUI to install, we some time have take
> a mistake to put "CTRL+C", then there will be stop with traceback. And we
> know the main process have shielded SIGINT, so we to shield subprocesses also.
> ```
> 
> If you wish, I can change the PR to take this form.


ok, i think it's so nice you can help me to make they as one commit. And i also like the title as you give me,  because the net is s o bad so i do it in web, so with the two commits.

VladimirSlavik:
Rebased interactively and force-pushed the new single commit.

VladimirSlavik (approved):
Looks good to me. Thank you!

VladimirSlavik:
Now released with [38.14](https://github.com/rhinstaller/anaconda/releases/tag/anaconda-38.14-1).

PR #4467, opened 2022-12-07
Delegate auto-attach handling to RHSM

M4rtinK:
At the moment both Simple Content Access and "classic" entitlement based subscription access is supported by Anaconda.

This presents an issue due to auto-attach action handling - for entitlement subscription to work correctly, we need to register and then auto-attach.

For SCA to work correctly, we should register and *not* attempt to auto-attach - the registration action does all that's needed and auto-attach afterwards is actually not a valid action.

For now this is hot-fixed on Hosted Candlepin side (and likely also on Satellite side) by ignoring these unnecessary/invalid auto-attach calls in SCA mode, but longer term the Candlepin team would like to drop this workaround.

After discussing this issue with the RHSM team, it was decided to handle this on the RHSM DBus daemon side as the daemon has all the necessary data (eq. which subscription mode is used, etc.) readily available.

All that is needed on the Anaconda side is to set an option in the RHSM Register call for RHSM to take over the auto-attach handling & drop all the auto-attach handling code.

Resolves: rhbz#2083318

### DIFF
--- a/pyanaconda/modules/subscription/runtime.py
+++ b/pyanaconda/modules/subscription/runtime.py
@@ -20,7 +20,7 @@ import json
 import datetime
 from collections import namedtuple
 
-from dasbus.typing import get_variant, Str
+from dasbus.typing import get_variant, Str, Bool
 from dasbus.connection import MessageBus
 from dasbus.error import DBusError
 
@@ -29,7 +29,7 @@ from pyanaconda.modules.common.task import Task
 from pyanaconda.modules.common.constants.services import RHSM
 from pyanaconda.modules.common.constants.objects import RHSM_REGISTER
 from pyanaconda.modules.common.errors.subscription import RegistrationError, \
-    UnregistrationError, SubscriptionError
+    UnregistrationError
 from pyanaconda.modules.common.structures.subscription import AttachedSubscription, \
     SystemPurposeData
 from pyanaconda.modules.subscription import system_purpose
@@ -247,7 +247,7 @@ class RegisterWithUsernamePasswordTask(Task):
                     organization,
                     self._username,
                     self._password,
-                    {},
+                    {"enable_content" : get_variant(Bool, True)},
                     {},
                     locale
                 )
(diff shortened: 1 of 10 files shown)


jstodola:
The required subscription-manager version should be updated in the spec file.

pyanaconda/modules/subscription/runtime.py
```diff
@@ -304,6 +304,8 @@ def run(self):
                     locale
                 )
                 log.debug("subscription: registered with organization and activation key")
+                log.debug("AAA ORG KEY REGISTRATION DATA")
+                log.debug(registration_data)
```
VladimirSlavik, line 308: LGTM except for these bits :)

M4rtinK:
Dropping the draft status, ready for regular review. :)

pyanaconda/modules/subscription/runtime.py
M4rtinK: > Code LGTM except for these bits :)

Cleaned up, thanks! :)

M4rtinK:
BTW, one can easily check if the delegation works by looking for the "Auto-attaching due to the enable_content option" in journal output coming from rhsmlib while register with a non-SCA account. :)

pyanaconda/modules/subscription/subscription.py
```diff
@@ -664,9 +664,12 @@ def register_username_password_with_task(self):
...
-        # if the task succeeds, it means the system has been registered
+        # if the task succeeds, it means the system has been registered and subscribed
+        # (with enable_content option se to True, RHSM will handle SCA/auto-attach transparently for us)
         task.succeeded_signal.connect(
             lambda: self.set_registered(True))
+        task.succeeded_signal.connect(
+            lambda: self.set_subscription_attached(True))
         # set SCA state based on data returned by the registration task
         task.succeeded_signal.connect(
             lambda: self.set_simple_content_access_enabled(
```
VladimirSlavik, line 675: I wonder if 3 lambdas could be a reason to have a method instead...

pyanaconda/modules/subscription/subscription.py
```diff
@@ -690,9 +693,11 @@ def register_organization_key_with_task(self):
...
                                                activation_keys=activation_keys)
-        # if the task succeeds, it means the system has been registered
+        # if the task succeeds, it means the system has been registered and subscription attached
         task.succeeded_signal.connect(
             lambda: self.set_registered(True))
+        task.succeeded_signal.connect(
+            lambda: self.set_subscription_attached(True))
         # set SCA state based on data returned by the registration task
         task.succeeded_signal.connect(
             lambda: self.set_simple_content_access_enabled(
```
VladimirSlavik, line 703: Same here

VladimirSlavik (approved):
LGTM, thank you. (I didn't look too closely at the tests.) The only thing I found is more of a choice than a rule...

pyanaconda/modules/subscription/subscription.py
M4rtinK: Thanks for the suggestion, but I think I would like to keep it this way and minimize the amount of changes. But looking at the RHEL 9 version, it should be cleaner there as the relevant code got simplified in the meantime. :)

M4rtinK:
Dropped two leftover log messages that implied Anaconda still handles auto-attach.

poncovka (approved):
Thanks!

PR #4468, opened 2022-12-07
Update grub2 and efibootmgr errors

frozencemetery:
Try to redirect issues requiring input from grub2 and efibootmgr more appropriately to reduce triage load on anaconda. 

[cc @poncovka , @marta-lewandowska ]

### DIFF
--- a/pyanaconda/modules/storage/bootloader/efi.py
+++ b/pyanaconda/modules/storage/bootloader/efi.py
@@ -78,8 +78,9 @@ class EFIBase(object):
             root=conf.target.system_root
         )
         if rc:
-            raise BootLoaderError("Failed to set new efi boot target. This is most "
-                                  "likely a kernel or firmware bug.")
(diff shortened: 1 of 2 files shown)


VladimirSlavik (approved):
Thank you, I like the idea.

VladimirSlavik:
Thank you, and sorry for taking this long.

After a discussion in the wider team, we are not prepared to merge this as is. The messages are factually correct, and _potentially_ help _some_ people do things more right. However, we would like to see also steps that the user can possibly take, or possibly some context, so that there's more than just "widgetnator code $9886 file bug in widgetbaz". 

Since that requires more work (communication) than just merging, and we are currently collectively overloaded, we'll try to get back to this in the second quarter of the year (CY23Q2).

(An expectation is that the stalebot will bump this in 60 days, which should be in time for the next quarter planning.)

frozencemetery:
As I understood it, this was requested by anaconda to reduce triage load.  If it is no longer desired, I have no problem with that.

PR #4470, opened 2022-12-13
Tune auto-releases

VladimirSlavik:
This fixes a lot of small kinks in the release machinery, and realizes the agreed change for dist-git as the only place to review.
- Release immediately on GH, no drafts
- Make the gChat reporting work
- Tell people in gChat to go to dist-git, since that's the first non-automated stop
- Push the new tag safely/correctly
- Do not reinvent the changelog, just propagate it
- Mention translations in changelog
- Include the auto-releases in CI dashboard

---

Tested on my fork (and messed it up really thoroughly).

Tagging action run: https://github.com/VladimirSlavik/anaconda/actions/runs/3687228579/jobs/6240709895
Resulting commit: https://github.com/VladimirSlavik/anaconda/commit/8796a7ebefdaf523fa951b9c2c2497edfc6675f6
Triggered release action run: https://github.com/VladimirSlavik/anaconda/actions/runs/3687308684/jobs/6240723227
Resulting release: https://github.com/VladimirSlavik/anaconda/releases/tag/anaconda-38.13-1

(This also shows how nothing breaks when there are no non-infra changes = empty changelog.)

Individual parts have been tested on real data, too.

### DIFF
--- a/.github/workflows/release-periodically.yml
+++ b/.github/workflows/release-periodically.yml
@@ -73,7 +73,7 @@ jobs:
             -X POST \
             -H 'Content-Type: application/json' \
             ${{ secrets.GCHAT_URL }} \
-            -d "{\"text\": \"New release tagged, please review at https://github.com/rhinstaller/anaconda/releases\"}"
+            -d "{\"text\": \"New release tagged, please review the Packit dist-git PR soon to be found at https://src.fedoraproject.org/rpms/anaconda/pull-requests"}"
 
 # This only makes a tag and pushes it, which is then picked up by another action that builds
 # the tarball and posts the actual release entity/page with it.
--- a/.github/workflows/release-periodically.yml
+++ b/.github/workflows/release-periodically.yml
@@ -72,7 +72,7 @@ jobs:
           curl \
             -X POST \
             -H 'Content-Type: application/json' \
-            ${{ secrets.GCHAT_URL }} \
+            '${{ secrets.GCHAT_URL }}' \
             -d "{\"text\": \"New release tagged, please review the Packit dist-git PR soon to be found at https://src.fedoraproject.org/rpms/anaconda/pull-requests"}"
 
 # This only makes a tag and pushes it, which is then picked up by another action that builds
--- a/.github/workflows/release-periodically.yml
+++ b/.github/workflows/release-periodically.yml
@@ -62,7 +62,12 @@ jobs:
         # the repo should have an access token from the checkout action
         # https://github.com/actions/checkout#Push-a-commit-using-the-built-in-token
         run: |
-          git push --tags origin master
+          tag_to_push=$(git describe --tags --exact-match)
+          if [ -n "${tag_to_push}" ] ; then
+            git push --atomic origin master "$tag_to_push"
+          else
+            exit 1
+          fi
 
       - name: Notify in Google chat
         if: steps.check-tag.outputs.can_bump_version
--- a/.github/workflows/release-periodically.yml.j2
+++ b/.github/workflows/release-periodically.yml.j2
@@ -56,7 +56,12 @@ jobs:
         # the repo should have an access token from the checkout action
         # https://github.com/actions/checkout#Push-a-commit-using-the-built-in-token
         run: |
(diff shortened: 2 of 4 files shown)


.github/workflows/release-periodically.yml
```diff
@@ -61,8 +61,14 @@ jobs:
         if: steps.check-tag.outputs.can_bump_version
         # the repo should have an access token from the checkout action
         # https://github.com/actions/checkout#Push-a-commit-using-the-built-in-token
+        # push first the commit to ensure there isn't anything merged before starting, and then add just the new tag
         run: |
-          git push --tags origin master
+          tag_to_push=$(git describe --tags --exact-match)
+          if [ -n "${tag_to_push}" ] ; then
+            git push origin master && git push origin "$tag_to_push"
```
poncovka, line 68: This is not atomic, we need to push the commit and the tag at the same time. See: https://github.blog/2015-04-30-git-2-4-atomic-pushes-push-to-deploy-and-more/

poncovka (approved):

.github/workflows/release-periodically.yml
VladimirSlavik: Thank you, that looks better, let's use it.

.github/workflows/release-periodically.yml
```diff
@@ -61,8 +61,14 @@ jobs:
         if: steps.check-tag.outputs.can_bump_version
         # the repo should have an access token from the checkout action
         # https://github.com/actions/checkout#Push-a-commit-using-the-built-in-token
+        # push first the commit to ensure there isn't anything merged before starting, and then add just the new tag
```
jkonecny12, line 64: I think this comment is not correct anymore.

.github/workflows/release-periodically.yml.j2
```diff
@@ -55,8 +55,14 @@ jobs:
         if: steps.check-tag.outputs.can_bump_version
         # the repo should have an access token from the checkout action
         # https://github.com/actions/checkout#Push-a-commit-using-the-built-in-token
+        # push first the commit to ensure there isn't anything merged before starting, and then add just the new tag
```
jkonecny12, line 58: The same here.

.github/workflows/tag-release.yml
```diff
@@ -78,26 +78,18 @@ jobs:
...
-          git log \
-            --no-merges \
-            --pretty=format:"- %s (%aL)" \
-            "$PREVIOUS_VERSION..$CURRENT_VERSION~1" \
-              | grep -v "#infra" \
-              | grep -v "build(deps-dev)" \
-              | tee release.txt
+          # get changelog from spec file
+          # - get lines starting with %changelog, up to but excluding first following empty line
+          # - then drop the first two lines (start at 3rd) which are %changelog and *DATE COMMITTER
```
jkonecny12, line 83: I really think that we should stop using `%changelog` in upstream spec file completely. Just generate the content by a script and consume it.

The `%changelog` entry here is not correct anyway because it doesn't have items from downstream rebuilds etc...

jkonecny12 (commented):
Looks good to me otherwise.

.github/workflows/tag-release.yml
VladimirSlavik: I agree with the direction of the idea, but it does not apply to this thing at this time.

1. This commit was mainly about establishing a single path for this data, which prevents reinventing the same thing. Once we stop doing `%changelog`, then that will change, but that's the future. And we will still want to collect these strings and propagate to the release page, somehow.

2. Downstream rebuilds are obviously downstream. But this builds the upstream tarball release and the text for its page. Does not apply.

.github/workflows/release-periodically.yml
VladimirSlavik: Thanks! good catch, fixed.

.github/workflows/tag-release.yml
VladimirSlavik: I missed the part "generate content by script". Yes, a separate script would be a preferable next step for this.

poncovka (approved):

PR #4471, opened 2022-12-13
Delegate auto-attach handling to RHSM

M4rtinK:
At the moment both Simple Content Access and "classic" entitlement based subscription access is supported by Anaconda.

This presents an issue due to auto-attach action handling - for entitlement subscription to work correctly, we need to register and then auto-attach.

For SCA to work correctly, we should register and *not* attempt to auto-attach - the registration action does all that's needed and auto-attach afterwards is actually not a valid action.

For now this is hot-fixed on Hosted Candlepin side (and likely also on Satellite side) by ignoring these unnecessary/invalid auto-attach calls in SCA mode, but longer term the Candlepin team would like to drop this workaround.

After discussing this issue with the RHSM team, it was decided to handle this on the RHSM DBus daemon side as the daemon has all the necessary data (eq. which subscription mode is used, etc.) readily available.

All that is needed on the Anaconda side is to set an option in the RHSM Register call for RHSM to take over the auto-attach handling & drop all the auto-attach handling code.

Resolves: rhbz#2083319

### DIFF
--- a/anaconda.spec.in
+++ b/anaconda.spec.in
@@ -40,7 +40,7 @@ Source0: %{name}-%{version}.tar.bz2
 %define pythonblivetver 1:3.4.0-15
 %define rpmver 4.10.0
 %define simplelinever 1.8.3-1
-%define subscriptionmanagerver 1.29.24
+%define subscriptionmanagerver 1.29.31
 %define utillinuxver 2.15.1
 
 BuildRequires: audit-libs-devel
--- a/pyanaconda/modules/common/constants/objects.py
+++ b/pyanaconda/modules/common/constants/objects.py
@@ -140,11 +140,6 @@ RHSM_UNREGISTER = DBusObjectIdentifier(
     basename="Unregister"
 )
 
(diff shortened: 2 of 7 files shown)


VladimirSlavik:
This is a port of #4467 with some changes, correct?

M4rtinK:
> This is a port of #4467 with some changes, correct?

Yep, I wondered if I should use the usual "cherry picked from" but I had to actually change quite a bit of the details due to how subscription management code changed from RHEL 8 to 9 (eq. Satellite support & code deduplication/cleanups) and in the end decided not to.

M4rtinK:
BTW, yet again this one can easily check if the delegation works by looking for the "Auto-attaching due to the enable_content option" in journal output coming from rhsmlib while registering with a non-SCA account. :)

M4rtinK:
Dropped two leftover log messages that implied Anaconda still handles auto-attach.

VladimirSlavik (approved):
Looks good to me. Thank you!

PR #4472, opened 2022-12-14
Assign remaining space to "/" in auto partitioning

OndrejZobal:
When using autopart with all the other grow partitions disabled the size cap of the root partition will be unset and fill the remaining space on the drive.

Resolves: [rhbz#1832570](https://bugzilla.redhat.com/show_bug.cgi?id=1832570#c9)

### DIFF
--- a/pyanaconda/modules/storage/partitioning/automatic/automatic_partitioning.py
+++ b/pyanaconda/modules/storage/partitioning/automatic/automatic_partitioning.py
@@ -155,7 +155,12 @@ class AutomaticPartitioningTask(NonInteractivePartitioningTask):
             specs.append(swap)
 
         # Configure specs.
+        contains_growable_partition = False
         for spec in specs:
+            # Detect if current configuration contains growable partitions apart from root.
+            if spec.grow and spec.mountpoint != "/":
+                contains_growable_partition = True
+
             # Set the default filesystem type.
             if spec.fstype is None:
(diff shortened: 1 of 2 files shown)


jstodola (commented):
This change doesn't look right to me.
1) It would resolve the problem for the reporter of bug 1832570 (although he doesn't need it any more), but it would also break deployments for the users that rely on the current behavior - so that `autopart --type=lvm --nohome` keeps free space that can be used later, when needed.
2) What would happen when `autopart --type=plain --nohome` is used on a system with 2 disks of the same size, let's say 250GB? `/boot/efi`, `/boot`, `swap` on the first disk and `/` on the second one. (I assume the current behavior would be all the point points created on the first disk, since there is enough space, and the second disk remains empty - but correct me if I'm wrong)
3) The new behavior would have to be documented, at least at https://github.com/rhinstaller/anaconda/blob/master/data/anaconda.conf#L198
4) Increasing the size of `/` on the installed system is much easier than shrinking it (if at all possible)
5) Making `/` as big as possible is simply not a recommended layout. Also re-installation would mean erasing all data from `/`. If the user really wants it this way, he/she should put an extra effort to specify in the kickstart file (with the `reqpart --add-boot` command it's not really so much work).

poncovka:
I am blocking the PR for now until I talk to @jstodola.

OndrejZobal:
Can I close this? @poncovka @jstodola

PR #4484, opened 2023-01-04
webui: pin down tag for fetching cockpit's test library

KKoukiou:


M4rtinK (approved):
Looks good to me & thanks for the fix! :)

BTW, how do you handle these pins in the cockpit project ? I guess if forgotten over time, it could perhaps break something instead due to the library being outdated. Just wait for CI to catch it or possibly some pro-active regular bumps ?

KKoukiou:
> Looks good to me & thanks for the fix! :)
> 
> BTW, how do you handle these pins in the cockpit project ? I guess if forgotten over time, it could perhaps break something instead due to the library being outdated. Just wait for CI to catch it or possibly some pro-active regular bumps ?

The latter. You need to manually now and then bump these cockpit tags.

poncovka (approved):

jkonecny12:
Looks great to me. @M4rtinK could you please add this bumping to a on-duty tutorial we have.

PR #4487, opened 2023-01-04
DEBUG: webui: bump cockpit's testlib to 282

KKoukiou:
Debugging failing CI.

### DIFF
--- a/ui/webui/Makefile.am
+++ b/ui/webui/Makefile.am
@@ -86,7 +86,7 @@ integration-test: bots test/common test/reference $(UPDATES_IMG)
 # checkout Cockpit's test API; this has no API stability guarantee, so check out a stable tag
 # when you start a new project, use the latest release, and update it from time to time
 test/common:
-	git clone -b 281 --depth=1 https://github.com/cockpit-project/cockpit.git tmp/cockpit
+	git clone -b main --depth=1 https://github.com/cockpit-project/cockpit.git tmp/cockpit
 	mv tmp/cockpit/test/common test/common
 	rm -rf tmp/cockpit
 
--- a/ui/webui/test/run-tests
+++ b/ui/webui/test/run-tests
@@ -270,9 +270,17 @@ def run(opts, image):
                     continue
                 if test_str in opts.exclude:
                     continue
-                nd = getattr(test_method, "_testlib__non_destructive", False)
-                rwa = getattr(test_method, "_testlib__retry_when_affected", True)
-                test = Test(test_id, build_command(filename, test_str, opts), test_timeout, nd, rwa)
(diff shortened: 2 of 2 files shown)


KKoukiou:
So the problem is between 282.1 and current main.
``` sh
[kkoukiou@sequioa cockpit]$ git log 282.1...HEAD test/common/
commit 379f6a45063aeed63daa7708cf1654d00281b853
Author: Martin Pitt <mpitt@redhat.com>
Date:   Tue Dec 20 16:28:17 2022 +0100

    test: Clean up decorator implementation
    
    Our cockpit test specific decorators were using wildly different
    implementations. Make them consistent by always adding a
    `_testlib__decoratorname` attribute to the test method or class.
    
    Introduce a `get_decorator()` helper which returns the decorator value on the
    test method or class. This avoids the custom code which iterates over the
    class'es methods and duplicates the value.
    
    This also debundles testlib from the `fmf_metadata` library, which we
    have never used so far, and don't really plan to either.
    
    There are still two kinds of decorators now: These with arguments, like @todo,
    or even the empty-argumented skipDistroPackage(); and the ones without
    arguments, like @nondestructive. This is still a bit confusing, but converting
    the latter is more intrusive.

commit ca75eba27786a090726766b12ab87c822ac66a80
Author: Martin Pitt <mpitt@redhat.com>
Date:   Tue Dec 20 15:28:58 2022 +0100

    test: Sort functions in testlib.py
    
    Group the global functions and the decorators together, and add a
    separator commit.

commit ebaea5592d000a41557cf73d75e16a7ae70fa844
Author: Marius Vollmer <mvollmer@redhat.com>
Date:   Wed Dec 21 10:34:00 2022 +0200

    test: Show lines starting with "WARNING:" for successful tests
    
    In the interest of saving space, we don't log all output from
    successful tests.  However, it is useful to see warnings like the
    recently introduced one for waits that get close to their timeout.
```

KKoukiou:
@martinpitt I bisected the changes, and it looks like commit 379f6a45063aeed63daa7708cf1654d00281b853 makes this repo fail. I took a look but it's not clear to me yet why.

KKoukiou:
@M4rtinK @jkonecny12 the problem in your CI came from ui/webui/test/run-tests, which is still a HACKY workaround. Th e original cockpit script needs to be adjusted and used instead. 

Do you have an item for this task?

KKoukiou:
> Thanks for finding this out @KKoukiou . We should look on this soon but it's not priority now.

I hope you understand the  consequences of not fixing this fast.

* You will be stuck with an outdated test/common checkout from commit untill this is fixed
* This means that when someone picks it up for fixing, test/common will have diverted to much, that it will be much more difficult to see the diff between your currently used script and the up to date one.

PR #4492, opened 2023-01-08
Qubes OS features

fepitre:
I propose a set of features allowing to reduce modifications we need for Qubes OS:

- Allow to specify default thin pool name and size,
- Allow to provide default partitioning based on scheme,
- Support for adding bootloader options in configuration file
- Support for setting encryption from configuration file
- Support for setting hostname from configuration file

We think that would benefit other parties, that's the reason of this PR.

### DIFF
--- a/data/anaconda.conf
+++ b/data/anaconda.conf
@@ -166,6 +166,7 @@ disable_grub_submenu = True
 # Set GRUB terminal type
 grub_terminal_type =
 
+# Additional default Grub options to be added
 additional_default_grub_options =
 
 [Storage]
--- a/data/anaconda.conf
+++ b/data/anaconda.conf
@@ -221,6 +221,11 @@ default_scheme = LVM
 #
 luks_version = luks2
 
+# Default thin pool name
+thin_pool_name =
+
+# Default thin pool size
+thin_pool_size =
 
 [Storage Constraints]
 
--- a/data/anaconda.conf
+++ b/data/anaconda.conf
@@ -74,6 +74,8 @@ can_save_installation_logs = True
 #
 default_on_boot = NONE
 
+# Set hostname
+hostname =
 
 [Payload]
 # Default package environment.
--- a/data/anaconda.conf
+++ b/data/anaconda.conf
@@ -167,6 +167,9 @@ terminal_type =
 additional_default_grub_options =
 
 [Storage]
+# Set encryption
+encrypted = False
+
 # Enable dmraid usage during the installation.
 dmraid = True
 
--- a/data/anaconda.conf
+++ b/data/anaconda.conf
@@ -161,10 +161,10 @@ preserved_arguments =
     nosmt
 
 # Disable the GRUB submenu
-disable_submenu = True
+disable_grub_submenu = True
 
 # Set GRUB terminal type
-terminal_type =
+grub_terminal_type =
 
 additional_default_grub_options =
 
--- a/data/anaconda.conf
+++ b/data/anaconda.conf
@@ -158,6 +158,13 @@ preserved_arguments =
     biosdevname ipv6.disable net.ifnames net.ifnames.prefix
     nosmt
 
+# Disable the GRUB submenu
+disable_submenu = True
+
+# Set GRUB terminal type
+terminal_type =
+
+additional_default_grub_options =
 
 [Storage]
 # Enable dmraid usage during the installation.
--- a/pyanaconda/modules/common/structures/partitioning.py
+++ b/pyanaconda/modules/common/structures/partitioning.py
@@ -34,7 +34,7 @@ class PartitioningRequest(DBusData):
         self._excluded_mount_points = []
         self._hibernation = False
 
-        self._encrypted = False
+        self._encrypted = conf.storage.encrypted
         self._passphrase = ""
         self._cipher = ""
         self._luks_version = ""
--- a/pyanaconda/modules/network/network.py
+++ b/pyanaconda/modules/network/network.py
@@ -64,7 +64,7 @@ class NetworkService(KickstartService):
         self._firewall_module = FirewallModule()
 
         self.hostname_changed = Signal()
-        self._hostname = ""
+        self._hostname = conf.network.hostname
 
         self.current_hostname_changed = Signal()
         self._hostname_service_proxy = self._get_hostname_proxy()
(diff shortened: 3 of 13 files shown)


data/anaconda.conf
```diff
@@ -158,6 +158,13 @@ preserved_arguments =
     biosdevname ipv6.disable net.ifnames net.ifnames.prefix
     nosmt
 
+# Disable the GRUB submenu
+disable_submenu = True
```
M4rtinK, line 164: Do other boot loaders have comparable sub-menus or maybe this could be also name-spaced to grub, eq. `disable_grub_submenu` ?

data/anaconda.conf
```diff
@@ -158,6 +158,13 @@ preserved_arguments =
...
+
+# Set GRUB terminal type
+terminal_type =
```
M4rtinK, line 167: Same here, especially if it uses some GRUB specific terminal type identifiers apparently, then I think it should be namespaced to grub.

data/anaconda.conf
fepitre: Yes it makes sense. For other bootloaders, I don't know particularly so I'll adapt the specific parameter for Grub here.

jkonecny12:
Thanks a lot for the contribution, it's great decision to share these changes! Given to the size of the PR from the side of feature added. It will probably take a bit more time for the review just FYI.

jkonecny12 (commented):
I look on the changes and they looks great to me. Thanks!

fepitre:
> I look on the changes and they looks great to me. Thanks!

Many thanks. Also, I'm currently getting updated on the changes that happened on anaconda since my last pass (Fedora 32). I'm about to update the `help_directory = /usr/share/anaconda/help/qubesos` part inside qubesos.conf. If you see some deprecated options I've missed, please don't hesitate to tell me.

pyanaconda/modules/storage/partitioning/automatic/utils.py
```diff
@@ -255,8 +255,7 @@ def schedule_implicit_partitions(storage, disks, scheme, encrypted=False, luks_f
 
     return devs
 
-
-def get_default_partitioning():
+def get_default_partitioning(scheme=None):
     """Get the default partitioning requests.
 
     :return: a list of partitioning specs
```
VladimirSlavik, line 261: Please add the new parameter to docstring.

pyanaconda/modules/network/network.py
```diff
@@ -64,7 +64,7 @@ def __init__(self):
         self._firewall_module = FirewallModule()
 
         self.hostname_changed = Signal()
-        self._hostname = ""
+        self._hostname = conf.network.hostname
```
VladimirSlavik, line 67: What does this do for other profiles? I assume nothing, since they have nothing set in hostname, because it is new?

pyanaconda/modules/common/structures/partitioning.py
```diff
@@ -34,7 +34,7 @@ def __init__(self):
         self._excluded_mount_points = []
         self._hibernation = False
 
-        self._encrypted = False
+        self._encrypted = conf.storage.encrypted
```
VladimirSlavik, line 37: Same question - what does this do outside Qubes OS?

VladimirSlavik (commented):
Thank you!

pyanaconda/modules/common/structures/partitioning.py
fepitre: Similar to hostname, it has no impact for other profiles because it uses default values from `anaconda.conf` where I've put the original values that was in the code.

pyanaconda/modules/storage/partitioning/automatic/utils.py
fepitre: Done.

pyanaconda/modules/network/network.py
fepitre: Please see my comment below.

pyanaconda/modules/common/structures/partitioning.py
VladimirSlavik: Thanks!

fepitre:
I see there is trivial conflicts to be solved, would you take care of them when merging or should I do it?

fepitre:
@VladimirSlavik Friendy ping for my comment above.

VladimirSlavik:
I'll leave that to @poncovka since this is assigned to her (at the moment).

fepitre:
@poncovka friendly ping.

fepitre:
Does this PR have any chance to be merged?

VladimirSlavik:
Frédéric, sorry for taking this long. I'll try to take care of this instead of Vendy.

In its current form, this PR is really a bag of miscellaneous items. What I'd like to do is consider each feature separately:

- Adding a default/initial value to conf - These are mostly fine, I think. It would be better to name them explicitly as defaults. This includes thin pools and encryption.

- Hostname default - More complicated than it seems. The value interacts with the hostname of the actual running system. The trick here is that if we set the hostname, then we do it in dracut, see `ksnet_to_dracut`. Previously there were only three sources of hostname:

   a) System default, which is taken by anaconda as the value in absence of anything else.
   b) Kickstart, which is applied in dracut so the machine boots with the correct hostname.
   c) User input, which is really first leaving the default in dracut, then reading it when anaconda starts, then later getting user input and applying it.

   All three keep the actual hostname and value we "know" the same. However, having the conf value nonempty brings a new source outside the current logic. To make matters worse, conf is not processed in dracut at all, because it can't be - conf files simply are not present in initramfs.

   So, I am uncertain: Existence of the current code indicates setting hostname in dracut has some importance. However you evidently use a hostname in conf and it works for you? Not sure which is more valid :-/ At the end, the go/nogo for this is a rabbit hole exercise for finding why we need hostname in dracut.

- Multiple default partitionings / schemes - that does not really fit how the config should be used, sorry.

- Adding the bootloader things - The capabilities are a good idea. However...
  - I'd have to find some who understands this and can verify. Note that the code was last touched 4 years earlier, when moving to the current file. This is not really something we want to touch too much.
  - As coded now, it seems some conf fields take precedence over what the code suggests. This is not a common situation, conf either has the only version of the data, or the lowest ranking one. It might be fine because the other value is on the class so it is really a fallback default - I don't know yet. In any case, getting someone to take the responsibility for saying "this is fine" is the real blocker here.

- The profile file is fine, except it mostly depends on acceptance of the other features. I'm not sure how valuable it is to have it with only some of the features? This would be up to you.

I don't have a definite opinion on the the tests yet, but I think using the qubes conf file isn't a good idea. Ideally I'd see the test independent of the configs, especially third-party ones. There's also the part where the defaults would be empty, so nothing to test really. Using a tempfile dir mocked to be the dropdir seems like a good idea for testing the values generally, but I can't tell you confidently how to do it right now.

Sorry, first this took too long and now I typed too much to read :-/

PR #4497, opened 2023-01-10
Always use `blivet.arch.is_s390()` to detect s390

VladimirSlavik:
Let's do it the same way everywhere. The (decade old) comments about import speed should not matter, we do the same in `startup_utils` and it's fine.

### DIFF
--- a/pyanaconda/argument_parsing.py
+++ b/pyanaconda/argument_parsing.py
@@ -28,6 +28,8 @@ import struct
 
 from argparse import ArgumentParser, ArgumentError, HelpFormatter, Namespace, Action, SUPPRESS
 
+from blivet.arch import is_s390
+
 from pyanaconda.core.kernel import KernelArguments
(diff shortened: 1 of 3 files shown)


poncovka (approved):
Thanks!

jstodola (approved):
It doesn't feel correct that the network module should depend on the storage library.
But it's already used this way in the timezone module as well, so...

poncovka:
> It doesn't feel correct that the network module should depend on the storage library. But it's already used this way in the timezone module as well, so...

In general, we shouldn't do that, but there are some exceptions to that rule. I am not sure if it is documented somewhere, but it should be safe to use `blivet.arch`, `blivet.size`, `blivet.udev` and `blivet.util`.

VladimirSlavik:
The failed tests all have some variation of "hardware breakage" - no usable disks, rsyslog can't write to journal file etc. That does not look related to the PR at all. Re-running.

PR #4500, opened 2023-01-10
Clean up the support for nfs additional repositories

poncovka:
**IMPORTANT:** This pull request doesn't change behaviour of nfs repositories. It cleans up the current implementation and prepares the code for possible changes in the supported nfs formats.

* Reorganize the `parse_nfs_url` function.
* Handle invalid protocols during the payload setup.
* Don't require a specific nfs format during the payload setup.

**Related to:** https://github.com/rhinstaller/anaconda/discussions/4491

### DIFF
--- a/pyanaconda/payload/dnf/payload.py
+++ b/pyanaconda/payload/dnf/payload.py
@@ -373,7 +373,7 @@ class DNFPayload(Payload):
             return
 
         # Set up the NFS source with a substituted URL.
-        if data.url.startswith("nfs://"):
+        if data.url.startswith("nfs:"):
             device_mount = self._create_mount_point(
                 constants.MOUNT_DIR,
                 data.name + "-nfs-device"
--- a/pyanaconda/payload/dnf/payload.py
+++ b/pyanaconda/payload/dnf/payload.py
@@ -365,11 +365,15 @@ class DNFPayload(Payload):
         """Set up sources for the additional repository."""
         # Check the validity of the repository.
         if not data.url:
-            msg = "The '{}' repository has no mirror, baseurl or metalink set."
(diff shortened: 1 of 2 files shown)


pyanaconda/payload/dnf/payload.py
```diff
@@ -365,11 +365,15 @@ def _set_up_additional_repository(self, data):
         """Set up sources for the additional repository."""
         # Check the validity of the repository.
         if not data.url:
-            msg = "The '{}' repository has no mirror, baseurl or metalink set."
+            msg = _("The '{}' repository has no mirror, baseurl or metalink set.")
             raise SourceSetupError(msg.format(data.name)) from None
```
rvykydal, line 369: Shouldn't we use hint for translators here? Something like:
```
            msg = _("The '{repository_name}' repository has no mirror, baseurl or metalink set.")
            raise SourceSetupError(msg.format(repository_name=data.name)) from None
```

pyanaconda/payload/dnf/payload.py
```diff
@@ -405,6 +410,11 @@ def _set_up_additional_repository(self, data):
             )
             result = task.run()
             data.url = "file://" + result.install_tree_path
+            return
+
+        # Otherwise, raise an error.
+        msg = _("The '{}' repository has an unsupported protocol.")
+        raise SourceSetupError(msg.format(data.name)) from None
```
rvykydal, line 417: The hint for translators would go also here (as above).

rvykydal (approved):
Maybe the hints for translators are worth considering,
otherwise looks good to me.

pyanaconda/payload/dnf/payload.py
poncovka: Fixed.

PR #4501, opened 2023-01-10
Show error messages and warnings from the payload setup

poncovka:
The Installation Destination screen should show the info bar and detailed
error and warnings messages if the last payload setup failed.


poncovka:
If the payload setup fails, there is now a visible info bar after revisiting the Installation Source screen.

Screenshot_vpn_test_03_2023-01-10_19:13:44

You can see specific error messages when you click on "Click for details".

Screenshot_vpn_test_03_2023-01-10_19:13:56

Previously, the info bar and the error messages were not available.

PR #4502, opened 2023-01-11
Fix missing tests in release archive

marmarek:
pyanaconda_tests has deep directory structure. Since makefile syntax doesn't support recursive wildcards, list up to 4-level deep dirs (the current layout).

Signed-off-by: Marek Marczykowski-Górecki <marmarek@invisiblethingslab.com>

### DIFF
--- a/tests/Makefile.am
+++ b/tests/Makefile.am
@@ -54,6 +54,9 @@ dist_check_SCRIPTS = $(srcdir)/glade_tests/*.py \
 		     $(srcdir)/gettext_tests/*.py \
 		     gettext_tests/canary_tests.sh \
 		     $(srcdir)/*_tests/*.py \
+		     $(srcdir)/*_tests/*/*.py \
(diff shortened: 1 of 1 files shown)


VladimirSlavik (commented):
Hello @marmarek, thank you for the contribution! Looking at what currently happens, or rather does not, it's a mess. This would help.

However, can you please explain what is the goal of this PR? The changes here would make the tarball include the code unit tests. However, the usual make targets (at least as used by the dev team) run all tests, and for that, there is a lot of missing things. Eg. pylint has one file out of three. So I believe this will definitely not enable running tests from the tarball, just move in the right direction...?

marmarek:
There might be other things missing indeed. The specific issue I hit is that patch made on git (on top of specific git tag) did not apply to release tarball (for the same release) because of the missing files.

VladimirSlavik (approved):

PR #4503, opened 2023-01-12
Move additional repositories on DBus

poncovka:
* Add the `Repositories` DBus property to the DNF module.
* Extend the `is_network_required` method of the DNF module.
* Handle the `repo` kickstart command in the DNF module.


jkonecny12 (approved):
Looks good to me. Thanks for improving the code!

PR #4505, opened 2023-01-13
webui: Fix disk refresh reference for storage screen help

M4rtinK:
Fixup the help txt to correctly refer to the Detect disks button.


VladimirSlavik (approved):
Thanks!

PR #4506, opened 2023-01-13
UI: Show error messages if the payload setup finishes early

poncovka:
When we register callbacks to signals of the payload manager in GUI and TUI, it is possible that the payload manager is already finished with the payload setup. In that case, we have to trigger the failed callback manually to set up the error messages. Otherwise, the error messages won't be visible in the UI.


VladimirSlavik (approved):
Looks good to me. Thank you!

PR #4507, opened 2023-01-13
Remove the is_complete method of the DNF payload class

poncovka:
Don't use the `is_complete` method of the DNF payload class. Instead, use
the `is_ready` method to verify if the payload is correctly set up.

Simplify the `status` and `completed` properties of the source spokes to show
a correct completeness of these spokes after the initialization. Previously,
the spoke could be shown as complete, but its status said otherwise.

I couldn't figure out why these checks should be different for automated
installations and the commit messages didn't provide any clues. There are
only three conditions we should care about:

1. All payload actions are finished (`self.ready`).
2. The latest payload setup didn't fail with an error (`self._error`).
3. The payload has some usable repositories enabled (`self.payload.is_ready()`).

### DIFF
--- a/pyanaconda/ui/tui/spokes/installation_source.py
+++ b/pyanaconda/ui/tui/spokes/installation_source.py
@@ -18,7 +18,6 @@
 #
 from pyanaconda.core.configuration.anaconda import conf
 from pyanaconda.core.payload import parse_nfs_url
-from pyanaconda.flags import flags
 from pyanaconda.modules.common.constants.objects import DEVICE_TREE
 from pyanaconda.modules.common.constants.services import STORAGE
 from pyanaconda.modules.common.structures.payload import RepoConfigurationData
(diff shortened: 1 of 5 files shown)


pyanaconda/ui/tui/spokes/installation_source.py
```diff
@@ -116,22 +116,20 @@ def _on_payload_failed(self):
...
+
+        if not self.completed:
+            return _(PAYLOAD_STATUS_INVALID_SOURCE)
+
+        source_proxy = self.payload.get_source_proxy()
+        return source_proxy.Description
 
     @property
     def completed(self):
-        if flags.automatedInstall and self.ready and not self.payload.is_ready():
```
VladimirSlavik, line 131: Yay, less flags! But how can you just stop using it? It isn't obvious what is different now. Perhaps it would be good to explain that somewhere? Otherwise it looks like just removing things randomly.

VladimirSlavik (approved):
Thank you!

pyanaconda/ui/tui/spokes/installation_source.py
```diff
@@ -116,22 +116,20 @@ def _on_payload_failed(self):
 
     @property
     def status(self):
-        if self._error:
-            return _("Error setting up software source")
```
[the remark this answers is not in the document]
poncovka: The `self._error` attribute is checked in the `completed` property and the spoke will show the "Error setting up repositories"  status if it is not complete, so the error status should be handled this way.

pyanaconda/ui/tui/spokes/installation_source.py
poncovka: I couldn't figure out why these checks should be different for automated installations and the commit messages didn't provide any clues. At the end, we should care about only three things:

1. Nothing is happening with the payload at this moment (`self.ready`).
2. The latest payload setup didn't fail with an error (`self._error`).
3. The payload has some usable repositories enabled (`self.payload.is_ready()`).

Could you explain what is missing, please?

VladimirSlavik: That sounds great! So maybe the missing bit is where this could be in the commit message?

poncovka: Updated.

PR #4510, opened 2023-01-17
ci(lint): add shell linter - Differential ShellCheck

jamacku:
A follow-up to the discussion in kickstart-tests PR:

- https://github.com/rhinstaller/kickstart-tests/pull/785

TL;DR Differential ShellCheck is a GitHub Action that performs differential ShellCheck scans on shell scripts changed via PR or in a commit and reports results directly in PR or in the Security tab.

/cc @VladimirSlavik

Based on the run in my fork, ShellCheck reports only [the following defects](https://github.com/jamacku/anaconda/security/code-scanning):

Screenshot from 2023-01-17 14-04-38

### DIFF
--- a/.github/workflows/differential-shellcheck.yml
+++ b/.github/workflows/differential-shellcheck.yml
@@ -1,0 +1,40 @@
+---
+
+name: Differential ShellCheck
+on:
+  push:
+  pull_request:
+    branches: [ master, rhel-* ]
+
+permissions:
+  contents: read
+
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+    
+    permissions:
+      security-events: write
+      pull-requests: write
+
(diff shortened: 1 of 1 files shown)


VladimirSlavik:
Thank you! Apparently it picks up the config files we already tuned for the unit test, which is _very_ good (IMO).

VladimirSlavik:
That PR should resolve the errors, the rest is warnings only so I think that might be fine.

VladimirSlavik (approved):
LGTM, thank you! Can you please make the commit message's first line start with `infra:` or end with `(#infra)` so that it gets sorted correctly?

jamacku:
Thank you, @VladimirSlavik. I have updated the commit message. Please have a look.

PR #4511, opened 2023-01-17
Ignore non-quoted array expansion in ShellCheck

VladimirSlavik:
The regular unit test does not trip on these, but the GitHub action would, once we start using it. See: https://github.com/rhinstaller/anaconda/pull/4510

Not #infra because these files are not all listed as infra.


poncovka (approved):
Thanks!

VladimirSlavik:
/kickstart-test --waive Only directives in comments, affects just tests

PR #4512, opened 2023-01-18
Use another type to make new GCC warnings go away

VladimirSlavik:
This should work.

I'm not a C type system guru, so I have only a hazy idea why this was the right thing. Feel free to point out why this is incorrect!


poncovka (approved):
Looks good to me. Thank you!

PR #4513, opened 2023-01-18
infra: Turn off infra rebuilds for Fedora 37

VladimirSlavik:
This turns off scheduled container rebuilds and release rehearsals for F37. It has been released for a few months already. Any work on these branches will be an exception and require special handling anyway.


poncovka (approved):
Thanks!

PR #4515, opened 2023-01-18
webui: Disable check for unexpected SELinux denials

martinpitt:
data/liveinst/liveinst disables SELinux for the installer. Correspondingly, disable cockpit's SELinux message checks, otherwise tests fail on tons of messages like

    avc: denied { create } for pid=4896 comm="bash" name="cockpit.conf" scontext=system_u:system_r:kernel_t:s0 tcontext=unconfined_u:object_r:user_tmp_t:s0 tclass=file permissive=1

----

See https://github.com/cockpit-project/bots/pull/4287 , this blocks the image refresh. However, I don't know if that's really the right thing to do. Please let's discuss in the image refresh PR.

### DIFF
--- a/ui/webui/Makefile.am
+++ b/ui/webui/Makefile.am
@@ -80,8 +80,10 @@ bots:
 $(UPDATES_IMG): bots
 	test/prepare-updates-img
 
+# test runs in kernel_t context and triggers massive amounts of SELinux
+# denials; SELinux gets disabled, but would still trigger unexpected messages
 integration-test: bots test/common test/reference $(UPDATES_IMG)
-	test/run-tests
+	TEST_AUDIT_NO_SELINUX=1 test/run-tests
 
 # checkout Cockpit's test API; this has no API stability guarantee, so check out a stable tag
 # when you start a new project, use the latest release, and update it from time to time


martinpitt:
@M4rtinK , @jkonecny12 : WDYT? See details in the bots PR.

martinpitt:
As we keep getting newer image refreshes, copying the question here.

@M4rtinK , @jkonecny12, do these [SELinux errors](https://cockpit-logs.us-east-1.linodeobjects.com/pull-4225-20230102-060810-4bdb672f-fedora-rawhide-boot-rhinstaller-anaconda/log.html#3) ring a bell?

> avc:  denied  { create } for  pid=4896 comm="bash" name="cockpit.conf" scontext=system_u:system_r:kernel_t:s0 tcontext=unconfined_u:object_r:user_tmp_t:s0 tclass=file permissive=1
> avc:  denied  { create } for  pid=4900 comm="mkdir" name="cockpittest" scontext=system_u:system_r:kernel_t:s0 tcontext=unconfined_u:object_r:var_lib_t:s0 tclass=dir permissive=1

These are with SELinux in permissive mode (which is not the default), and it seems they run in `kernel_t` context, which is certainly unusual. `data/liveinst/liveinst` calls `setenforce 0`, is that involved in the test run? If Anaconda is somehow known not to work with SELinux, then the integration tests should just disable the checking of unexpected messages, by calling the tests with env `TEST_AUDIT_NO_SELINUX=1`.

But I can't judge whether this would be the right thing to do, or if there's something actually wrong with Rawhide's current SELinux policy. If it is, please land this PR to unblock your testing, otherwise please describe how you run anaconda, and how selinux-policy should be adjusted. Thanks!

M4rtinK (approved):
Looks good to me!

In short - we do run with SELinux disabled on purpose during the installation. While I'm not sure anyone tried recently what would happen if we tried enabling it, the general expectation is that the Fedora SELinux policy would likely not work for the special case that is OS installation.

So thanks for the PR & I think this is the way to go, at least for now. :)

martinpitt:
Thanks @M4rtinK !

PR #4517, opened 2023-01-20
Remove the dmraid and nodmraid boot options

vojtechtrefny:
These control the blivet dmraid flag which is being removed. We are switching to mdadm for Firmware/BIOS RAID support in Fedora, see https://fedoraproject.org/wiki/Changes/UseMdadmForBIOSRAIDInAnaconda

Note that this can be merged separately from the blivet change, but it would probably be better to wait for https://github.com/storaged-project/blivet/pull/1098 to be merged first.


docs/boot-options.rst
```diff
@@ -682,19 +682,6 @@ back to a local boot.
 Storage options
 ---------------
 
-.. inst.nodmraid:
-
-inst.nodmraid
```
poncovka, line 687: I guess this should be moved to the [Removed Options](https://anaconda-installer.readthedocs.io/en/latest/boot-options.html#removed-options).

data/anaconda_options.txt
```diff
@@ -153,12 +153,6 @@ disklabel
 Prefer creation of the specified disk label type. Specify "gpt" to prefer creation of GPT disk
 labels. Specify "mbr" to prefer creation of MBR disk labels if supported.
 
-nodmraid
```
poncovka, line 156: I guess we could show a warning if `inst.nodmraid` is set. Please, look for `check_removed_arg "inst.nompath"` in our dracut code and add a new line for `inst.nodmraid`.

poncovka (approved):
Thank you!

vojtechtrefny:
One problem with the `nodmraid` flag is that it is also used outside Anaconda -- `mdadm` udev rules use it together with `noiswmd` to disable autoactivation of the BIOS/Firmware RAID arrays. But I guess this doesn't really concern Anaconda -- in Anaconda it only controls the blivet `dmraid` flag which we are removing and I assume that from Anaconda point of view it doesn't matter that `nodmraid` (without the `inst` prefix) is still a valid flag, right?

docs/boot-options.rst
vojtechtrefny: Ok, I hope the comment about the removal makes sense.

data/anaconda_options.txt
vojtechtrefny: Fixed.

VladimirSlavik (approved):
Aside from the question above, LGTM, thank you!

vojtechtrefny:
> There is one more place where this might be removed, I think. If I got it right, the ignore list does not need dmraid anymore, because blivet no longer asks us to install it...?
> 
> https://github.com/rhinstaller/anaconda/blob/master/data/profile.d/rhel.conf#L21

Yes, that's right, I removed the package. Thanks.

poncovka (approved):
Thank you!

vojtechtrefny:
The blivet PR is merged and I am doing a build for rawhide now so this can be merged too. The blivet build with the fix will be 3.6.1-3 but I don't think it's necessary to bump the required version in spec. And everything should still work fine without the Anaconda part so you don't need to do a new rawhide build right away.

PR #4521, opened 2023-01-24
infra: Release automatically only on even weeks

VladimirSlavik:


.github/workflows/release-periodically.yml
```diff
@@ -24,9 +24,24 @@ permissions:
   contents: write
 
 jobs:
+  check-date:
```
poncovka, line 27: What are the benefits of having this done in a separate job? Couldn't we do it in the steps?

.github/workflows/release-periodically.yml
```diff
@@ -24,9 +24,24 @@ permissions:
...
+            echo "is_even=true" >> $GITHUB_OUTPUT
+          fi
+    outputs:
+      is_even: ${{ steps.check.outputs.is_even }}
+
   make-release-tag:
     # Don't run scheduled workflows on forks.
-    if: github.event_name != 'schedule' || github.repository == 'rhinstaller/anaconda'
+    needs: check-date
+    if: ${{ (github.event_name != 'schedule' || github.repository == 'rhinstaller/anaconda') && needs.check-date.outputs.is_even == 'true' }}
```
poncovka, line 44: Did you test the changes? I have a feeling that `== 'true'` doesn't work for boolean values and I had to fix these conditions the last time. Anyway, you can just check `needs.check-date.outputs.is_even` (unless I am missing something). See `steps.check-tag.outputs.can_bump_version` below.

.github/workflows/release-periodically.yml
```diff
@@ -24,9 +24,24 @@ permissions:
...
 jobs:
+  check-date:
+    # run only on even weeks
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check if the week number is even
+        id: check
+        run: |
+          week_no=$(date +"%V")
+          if [ $((10#${week_no}%2)) -eq 0 ]; then
```
poncovka, line 35: Btw. this is the osbuild version, for inspiration, https://github.com/osbuild/osbuild/blob/main/.github/workflows/create-tag.yml#L15. However, it doesn't look much better.

.github/workflows/release-periodically.yml
jkonecny12: I guess the 'true' value is not taken as boolean but just a text. It's stored as true and compared to that?

Ideally we should test this.

jkonecny12 (commented):
Looks good to me, however, I like suggestions from @poncovka.

VladimirSlavik:
Sorry, I missed that it isn't a draft. Yes, this is untested. I expected more bikeshedding over the bash part...

.github/workflows/release-periodically.yml
poncovka: One additional note, if the running workflow was not scheduled, it means that it was triggered manually and it should be able to run regardless of the week parity.

VladimirSlavik:
Looks like recently we have enough to release weekly, so perhaps it's not needed at all.

PR #4522, opened 2023-01-25
Bump blivet version for nvme-of

VladimirSlavik:
We forgot one tiiiiny bit... let's not.

cc @vojtechtrefny @jstodola


vojtechtrefny (approved):

VladimirSlavik:
Blocked because it's not clear yet which if the bugs should be used, and so the commit message will need amending.

M4rtinK (approved):
Looks good to me. :)

poncovka (approved):

PR #4523, opened 2023-01-25
Templatize kickstart version

VladimirSlavik:
This is a proof of concept for using templates to do more than just infra. For a correctly filled branch-config, it imports the right kickstart version.

Feel free to discuss instead of ack/nack, I am not particularly attached to it.


M4rtinK (approved):
Looks good to me, thanks! :)

It has always been a pain to change the PyKickstart version manually during branching, so good to see this addressed. :)

pyanaconda/core/kickstart/version.py.j2
```diff
@@ -0,0 +1,29 @@
...
+# License and may only be used or replicated with the express permission of
+# Red Hat, Inc.
+#
+
+{% if distro_release == "rawhide" %}
+from pykickstart.version import DEVEL as VERSION
+{% elif distro_name == "fedora" %}
+from pykickstart.version import F{$ distro_release $} as VERSION
+{% elif distro_name == "rhel" %}
+from pykickstart.version import RHEL{$ distro_release $} as VERSION
```
poncovka, line 26: For RHEL, this won't help you. You also have to go though the `commands.py` file line by line and change the commands there. I don't think we want that in templates.

VladimirSlavik: True. Then again, this does not hurt things, I think?

PR #4525, opened 2023-01-25
Remove leftovers after the isys module removal

VladimirSlavik:
There's always more to remove.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -324,38 +324,3 @@ if not on_rtd:  # only import and set the theme if we're building docs locally
 autodoc_member_order = 'source'
 
 # otherwise, readthedocs.org uses their theme by default, so no need to specify it
-
-# This was taken directly from here:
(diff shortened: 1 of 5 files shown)


docs/conf.py
```diff
@@ -356,6 +356,6 @@ def __getattr__(cls, name):
     def __getitem__(cls, key):
         return cls.__getattr__(key)
 
-MOCK_MODULES = ['_isys']
+MOCK_MODULES = []
 for mod_name in MOCK_MODULES:
     sys.modules[mod_name] = Mock()
```
VladimirSlavik, line 361: I wonder if this is needed at all.

poncovka: I would drop that.

poncovka (approved):
Thanks!

docs/conf.py
VladimirSlavik: Thanks, agreed.

PR #4527, opened 2023-01-25
Show an error message when subscription token transfer fails

M4rtinK:
Instead of a hard failure and confusing traceback show an error message when subscription token transfer fails.

The main expected trigger for this happening is users with an expired free Red Hat developer account, so we mention that in the error message.

Also if we actually get to the token transfer attempt, there must be a working installation source and the system will be most likely functional. It will just not have a subscription - so make the error non-fatal and make it possible to continue with the installation.

Resolves: rhbz#2057472

### DIFF
--- a/pyanaconda/errors.py
+++ b/pyanaconda/errors.py
@@ -116,6 +116,7 @@ class ErrorHandler(object):
             InsightsClientMissingError.__name__: self._insightsErrorHandler,
             InsightsConnectError.__name__: self._insightsErrorHandler,
             "KickstartRegistrationError": self._kickstartRegistrationErrorHandler,
+            "SubscriptionTokenTransferError": self._subscriptionTokenTransferErrorHandler,
 
             # Satellite
             SatelliteProvisioningError.__name__: self._target_satellite_provisioning_error_handler,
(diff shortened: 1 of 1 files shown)


M4rtinK:
token_Transfer_v1
And this is how it looks like. :)

VladimirSlavik (approved):
Looks good to me. Thank you!

pyanaconda/errors.py
```diff
@@ -116,6 +116,7 @@ def _get_default_mapping(self):
             InsightsClientMissingError.__name__: self._insightsErrorHandler,
             InsightsConnectError.__name__: self._insightsErrorHandler,
             "KickstartRegistrationError": self._kickstartRegistrationErrorHandler,
+            "SubscriptionTokenTransferError": self._subscriptionTokenTransferErrorHandler,
```
poncovka, line 119: You should use the `NonCriticalInstallationError` exception.

M4rtinK: I think that's a bigger change in code than whats in scope for this bug & right next to this error handler are others that could be quite likely converted as well and are definitely not in scope.

Still, I think we could certainly use the new mechanism for the port to Fedora.

PR #4528, opened 2023-01-26
infra: Stop installing npm dependencies manually

VladimirSlavik:
It was for previous rawhide. Now it prevents proper operation, instead.

---

This makes the rpm container buildable and unblocks tests.


rvykydal (approved):
Thank you!

M4rtinK (approved):
Looks good to me as well, thanks for finding it! :)

I wonder if we perhaps could note all these pinned dependencies somewhere, so we don't loose track of them ? We could then periodically check that list & look at it when we run into some issues as the first thing as well.

jkonecny12 (approved):
That is great. Thanks!

PR #4532, opened 2023-01-27
Apply --noverifyssl option for liveimg kickstart command

jstodola:
Related: rhbz#1886985

### DIFF
--- a/pyanaconda/modules/payloads/source/live_image/initialization.py
+++ b/pyanaconda/modules/payloads/source/live_image/initialization.py
@@ -90,6 +90,7 @@ class SetUpRemoteImageSourceTask(Task):
         super().__init__()
         self._url = configuration.url
         self._proxy = configuration.proxy
+        self._ssl_verify = configuration.ssl_verification_enabled
 
     @property
     def name(self):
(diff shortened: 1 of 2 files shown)


M4rtinK (approved):
Looks good to me, thanks! :)

VladimirSlavik (approved):
Thank you, looks good to me.

(I checked the actual download / source use, and this should be the whole fix.)

poncovka (approved):
Thanks!

jkonecny12:
We are going to backport this to RHEL. So I think we should have here also the unit tests? WDYT @poncovka ?

poncovka:
> We are going to backport this to RHEL. So I think we should have here also the unit tests? WDYT @poncovka ?

Sure. Why not.

VladimirSlavik:
Normally I'd just merge this, but I can't exactly assign myself since I'm away, and it's acked but we want tests. No idea what to do here.

jkonecny12:
@jstodola are you fine with creating also a unit tests or should we handle that for you?

jstodola:
I will give it a try and ask for help if I fail.

jstodola:
A unit test added.

PR #4534, opened 2023-02-01
Fix the systemd generator for systemd 253 (#2165433)

AdamWill:
As Zbyszek explained in
https://bugzilla.redhat.com/show_bug.cgi?id=2165433#c5 , generators aren't supposed to write outside the special locations passed to them as args. Just writing the files into the first of the provided locations seems to work fine (tested that this fixes both text install and rescue mode).

### DIFF
--- a/data/systemd/anaconda-generator
+++ b/data/systemd/anaconda-generator
@@ -11,8 +11,8 @@ fi
 
 # set up dirs
 systemd_dir=/lib/systemd/system
-target_dir=$systemd_dir/anaconda.target.wants
-mkdir -p $target_dir
+target_dir="$1/anaconda.target.wants"
+mkdir -p "$target_dir"
 
 # create symlink anaconda.target.wants/SERVICE@TTY.service
 service_on_tty() {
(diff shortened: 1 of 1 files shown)


AdamWill:
I tested this by having openQA build a netinst with it included and booting it on my test box. Both `inst.text` and rescue mode boot worked fine.

poncovka:
@AdamWill, can you add quotation marks for `$target_dir` in the `mkdir -p $target_dir` statement, please?

For those, who are wondering. I think that shellcheck is too smart and it knew that the value of `$target_dir` was safe before, so it didn't need quotation marks. Now it doesn't know the value of `$1`, so the value of `$target_dir` is unknown and all occurrences need quotation marks.

AdamWill:
Yes, I know, if you look I already did one round of revisions for shellcheck errors. Just missed that one.

jkonecny12 (approved):
Looks good to me. Thanks!

VladimirSlavik (approved):
Looks good to me. Thank you!

poncovka (approved):

keszybz:
A bit late, but LGTM too. Thanks @AdamWill for handling this.

PR #4537, opened 2023-02-02
Add Sericea - ostree based Sway variant

jkonecny12:
https://fedoraproject.org/wiki/Changes/Fedora_Sway_Spin

### DIFF
--- a/data/profile.d/Makefile.am
+++ b/data/profile.d/Makefile.am
@@ -24,6 +24,7 @@ dist_config_DATA    = \
 	fedora-iot.conf \
 	fedora-kde.conf \
 	fedora-kinoite.conf \
+	fedora-sericea.conf \
 	fedora-server.conf \
 	fedora-silverblue.conf \
 	fedora-workstation.conf \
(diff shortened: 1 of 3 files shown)


data/profile.d/Makefile.am
```diff
@@ -24,6 +24,7 @@ dist_config_DATA    = \
 	fedora-iot.conf \
 	fedora-kde.conf \
 	fedora-kinoite.conf \
+	fedora-sericea.conf \
```
VladimirSlavik, line 27: Can we just wildcard these?

VladimirSlavik (approved):

data/profile.d/Makefile.am
jkonecny12: IIRC it is not recommended and we had a PR from outside to change place we had back than.

See here: https://github.com/rhinstaller/anaconda/pull/3619/commits/5c7ea9c2cee989d0b1b03cc7cfaf24f951271920

VladimirSlavik: Ah, sorry then - I thought that's only in srpms.

PR #4539, opened 2023-02-03
Sort RPM versions via rpm.labelCompare() and not via packaging.version.LegacyVersion()

hroncok:
Packaging 22+ removed the long-deprecated packaging.version.LegacyVersion class. The packaging.version library is intended to parse Python (PEP 440) versions, not arbitrary versions and definitively not RPM versions.

Even if LegacyVersion was still available, it may produce incorrect results if the version contains ~, ^ or other characters with special meaning in RPM.

This assumes the Python rpm module is always available. If that assumption is wrong
the logic from rpm.labelCompare() needs to be reimplemented instead.

--------------------

Intentionally opening this as a draft for discussion, so not following the checklist below yet.

Please check that your PR follows these rules:

* [ ] [**Code conventions**](https://anaconda-installer.readthedocs.io/en/latest/contributing.html#code-conventions). tl;dr: Follow PEP8, wrap at 100 chars, and provide docstrings.

* [ ] [**Commit message conventions**](https://anaconda-installer.readthedocs.io/en/latest/commit-log.html). tl;dr: Heading, empty line, longer explanations, all wrapped manually. If in doubt, write a longer commit message with more details.

* [ ] **Tests** pass and cover your changes. You can [run tests locally manually](https://anaconda-installer.readthedocs.io/en/latest/testing.html), or have them run as part of the PR. A team member will have to enable tests manually after inspecting the PR.
  *If you don't know how, ask us for help!*

* [ ] [**Release notes**](https://anaconda-installer.readthedocs.io/en/latest/contributing.html#release-notes) and **docs** if the PR adds something major or changes existing behavior.
  *If you don't know how, ask us for help!*

* [ ] **RHEL** rules: If your PR is for a `rhel-*` branch, pay special attention to commit messages. Make sure **all** commit messages include a line linking the commit to a bug, such as `Resolves: rhbz#123456`. For more information, see [the complete rules](https://anaconda-installer.readthedocs.io/en/latest/commit-log.html).
  *If you don't know how, ask us for help!*

### DIFF
--- a/pyanaconda/modules/common/versions.py
+++ b/pyanaconda/modules/common/versions.py
@@ -1,0 +1,31 @@
+from functools import total_ordering
+import rpm
+
+
+@total_ordering
+class RPMVersion:
+    def __init__(self, version_string):
+        # TODO do we need to parse epoch? nothing in the tests indicates that
+        self._version, _, self._release = version_string.partition('-')
+
+    def __str__(self):
+        return f'{self._version}-{self._release}'
+
+    def __repr__(self):
+        return f'<RPMVersion({str(self)!r})>'
+
+    def __hash__(self):
+        return hash((self._version, self._release))
+
+    def __eq__(self, other):
+        if not isinstance(other, RPMVersion):
+            return NotImplemented
+
+        return (self._version, self._release) == (other._version, other._release)
+
+    def __lt__(self, other):
+        if not isinstance(other, RPMVersion):
+            return NotImplemented
+
+        return rpm.labelCompare(('0', self._version, self._release),
+                                ('0', other._version, other._release)) == -1
--- a/anaconda.spec.in
+++ b/anaconda.spec.in
@@ -101,7 +101,6 @@ Requires: python3-pwquality
 Requires: python3-systemd
 Requires: python3-productmd
 Requires: python3-dasbus >= %{dasbusver}
-Requires: python3-packaging
 Requires: flatpak-libs
 %if %{defined rhel} && %{undefined centos}
 Requires: subscription-manager >= %{subscriptionmanagerver}
--- a/pyanaconda/modules/payloads/base/utils.py
+++ b/pyanaconda/modules/payloads/base/utils.py
@@ -19,8 +19,7 @@
 #
 import functools
 
-from packaging.version import LegacyVersion as parse_version
-
+from pyanaconda.modules.common.versions import RPMVersion as parse_version
 from pyanaconda.anaconda_loggers import get_module_logger
 log = get_module_logger(__name__)
 
--- a/pyanaconda/modules/payloads/base/utils.py
+++ b/pyanaconda/modules/payloads/base/utils.py
@@ -17,8 +17,6 @@
 # License and may only be used or replicated with the express permission of
 # Red Hat, Inc.
 #
-import functools
-
 from pyanaconda.modules.common.versions import RPMVersion as parse_version
 from pyanaconda.anaconda_loggers import get_module_logger
 log = get_module_logger(__name__)
@@ -26,11 +24,4 @@ log = get_module_logger(__name__)
 
 def sort_kernel_version_list(kernel_version_list):
     """Sort the given kernel version list."""
-    kernel_version_list.sort(key=functools.cmp_to_key(_compare_versions))
-
-
-def _compare_versions(v1, v2):
-    """Compare two version number strings."""
-    first_version = parse_version(v1)
-    second_version = parse_version(v2)
-    return (first_version > second_version) - (first_version < second_version)
+    kernel_version_list.sort(key=parse_version)
--- a/pyanaconda/modules/storage/checker/utils.py
+++ b/pyanaconda/modules/storage/checker/utils.py
@@ -20,7 +20,6 @@ gi.require_version("BlockDev", "2.0")
 from gi.repository import BlockDev as blockdev
 
 from collections import defaultdict
-from packaging.version import LegacyVersion as parse_version
 
 from blivet import arch, util
 from blivet.devicefactory import get_device_type
@@ -35,6 +34,7 @@ from pyanaconda.core.constants import productName, STORAGE_REFORMAT_BLOCKLIST, \
     STORAGE_MUST_NOT_BE_ON_ROOT
 from pyanaconda.core.i18n import _
 from pyanaconda.core.storage import DEVICE_TEXT_MAP
+from pyanaconda.modules.common.versions import RPMVersion as parse_version
 from pyanaconda.modules.storage.platform import platform
 
 from pyanaconda.anaconda_loggers import get_module_logger


hroncok:
:warning: I've manually run some sorts, but never tested this from within anaconda itself.

pyanaconda/modules/common/versions.py
```diff
@@ -0,0 +1,31 @@
+from functools import total_ordering
+import rpm
+
+
+@total_ordering
+class RPMVersion:
+    def __init__(self, version_string):
+        # TODO do we need to parse epoch? nothing in the tests indicates that
+        self._version, _, self._release = version_string.partition('-')
```
AdamWill, line 9: Why use `.partition` and then throw away the separator? Why not just use `.split` instead?

AdamWill:
Aside from the minor comment, this LGTM. I dug into the history here a bit, for the record. it dates all the way back to https://bugzilla.redhat.com/show_bug.cgi?id=1074358 , where we were trying to stop anaconda generating initramfs for kernels in shared `/boot` locations from other OS installs (other distros entirely, or other versions of Fedora/RHEL) and have it only generate initramfs files for kernels it had actually deployed itself. This is still pretty much what we're doing, I think, with three different payloads that have different ways of determining what kernel files they want to operate on, then passing that list to `sort_kernel_version_list`.

The commit that fixed that problem is 850063aa6a19f87a6ffd6009e00f15c3418d1111 , which changes a lot of stuff, but along the way, changed the way kernel versions were sorted. The code it dropped did this:

        try:
            import yum
        except ImportError:
            cmpfunc = cmp
        else:
            cmpfunc = yum.rpmUtils.miscutils.compareVerOnly
    ...
        versions = sorted((f.split("/")[-1][8:] for f in files if os.path.isfile(f)), cmp=cmpfunc)

so it was using a yum function for the version comparison if it could, otherwise just plain Python `cmp` (which wouldn't be very reliable). For some reason, that commit changed to using `distutils.version.LooseVersion` instead. I can't find any record of an explanation for this change; it's not explained in the commit message, nor in the discussions on the old anaconda-patches list [here](https://lists.fedorahosted.org/pipermail/anaconda-patches/2015-January/015699.html) or [here](https://lists.fedorahosted.org/pipermail/anaconda-patches/2015-January/015615.html) AFAICS.

We kept using `LooseVersion` for a long time after that, through various other changes, till September 2021, when [a bug was reported that distutils was deprecated](https://bugzilla.redhat.com/show_bug.cgi?id=2003407)...so in 25493efeaa5b3845e1b835f08a4729c2ee30d726 we switched to using `packaging.version.parse`. Then in January, that was changed to `packaging.version.LegacyVersion` in 055d767827046504b15b02d46ba1bf0632b4d7fa . The explanation given was "LegacyVersion instances are always ordered lower than Version instances, so we should use the LegacyVersion class instead of the parse function.", though that doesn't make a lot of sense to me since the two places that were changed are the *only* places where anaconda uses `packaging.version` at all AFAICS, so I don't see what comparison between a `Version` and a `LegacyVersion` we would've been worried about there.

Anyway, that's the background, in case anyone's interested. On the whole it seems pretty fine to assume that the only kernel filenames anaconda should worry about dealing with (after filtering out rescue kernels) are ones that can be parsed like RPM versions; I can't offhand think of any plausible scenario in which we'd have a kernel file that we count as part of the deployed payload which *doesn't* look like that. But if we wanted to be extra sure we could, I guess, have the comparison fall back on something like `cmp` that won't fail, if the `rpm.labelCompare` call fails...

AdamWill:
hmm, thinking it through a bit harder...I wonder if ostree payload might be an issue here? we may actually not *have* python3-rpm on ostree installer images...

pyanaconda/modules/common/versions.py
hroncok: I always forget split takes a maxsplit argument.

hroncok:
> hmm, thinking it through a bit harder...I wonder if ostree payload might be an issue here? we may actually not _have_ python3-rpm on ostree installer images...

That's what I worried about in the commit message.

pyanaconda/modules/common/versions.py
hroncok: About multiple occurrences: this now works with both simpe versions without release and also with dashes is the release, so it requires no input validation.

hroncok:
> though that doesn't make a lot of sense to me since the two places that were changed are the _only_ places where anaconda uses `packaging.version` at all AFAICS, so I don't see what comparison between a `Version` and a `LegacyVersion` we would've been worried about there.

This is where Version was created and compared with LegacyVersion:

https://github.com/rhinstaller/anaconda/blob/055d767827046504b15b02d46ba1bf0632b4d7fa/pyanaconda/modules/storage/checker/utils.py#L240

AdamWill:
So, I checked. A Silverblue installer image can `import rpm`, and indeed, [anaconda-core explicitly requires python3-rpm](https://src.fedoraproject.org/rpms/anaconda/blob/rawhide/f/anaconda.spec#_91). So we do seem fairly safe from that perspective at least at present.

pyanaconda/modules/common/versions.py
hroncok: Using split would force me to specialcase release-less versions, such as  https://github.com/rhinstaller/anaconda/blob/055d767827046504b15b02d46ba1bf0632b4d7fa/pyanaconda/modules/storage/checker/utils.py#L240

VladimirSlavik:
Just wanted to confirm that the unexplained comparison commit happened because that parse function tries to create the non-legacy class first. So yes, that comparison happened.

We should clean up the python3-packaging in specfile too, we meaning the dev team of course :)

hroncok:
> We should clean up the python3-packaging in specfile too

Done :)

VladimirSlavik:
That was fast :0

hroncok:
I want this merged fast, so we can tag packaging back in (it was untagged from rawhide when @AdamWill discovered anaconda fails).

hroncok:
Do I assume correctly the tests are running now?

VladimirSlavik:
Yes, I enabled the tests. As an external contributor you need someone to do that on every push :(

hroncok:
Will restrain from further pushes at least until we see if it works on the CI.

VladimirSlavik:
On mobile, so bad log visibility. The pylint error is new, but the nvme import was there even before, if that fails too.

hroncok:
I'll make hash return int. I assumed it needs to return anything hashable.

poncovka:
The unit tests pass when I run them locally. I think that the change in the .spec file triggers a rebuild of the testing container when run as a GitHib action and pulls in a broken version of blockdev.

VladimirSlavik:
Agreed about the specfile change.

rvykydal:
> The unit tests pass when I run them locally. I think that the change in the .spec file triggers a rebuild of the testing container when run as a GitHub action and pulls in a broken version of blockdev.
The broken blockdev is tracked here:https://github.com/rhinstaller/kickstart-tests/issues/882

hroncok:
> I'll make hash return int. I assumed it needs to return anything hashable.

Amended.

hroncok:
Note that the srt, repr and hash methods are there to make debugging easier, they are not used at all. If you wish, I can write some simple tests for them or ditch them.

hroncok:
Interestingly, the `sort_kernel_version_list` function uses the `functools.cmp_to_key` function to create a key from the `_compare_versions` cmp function which uses comparisons of `RPMVersion` (previously `LegacyVersion`) instances to return the same number (-1, 0, +1) as `rpm.labelCompare`.

I believe this could be avoided with this simple change:

```diff
diff --git a/pyanaconda/modules/payloads/base/utils.py b/pyanaconda/modules/payloads/base/utils.py
index d9b56e955c..72ee2ff4b7 100644
--- a/pyanaconda/modules/payloads/base/utils.py
+++ b/pyanaconda/modules/payloads/base/utils.py
@@ -26,11 +26,4 @@ log = get_module_logger(__name__)
 
 def sort_kernel_version_list(kernel_version_list):
     """Sort the given kernel version list."""
-    kernel_version_list.sort(key=functools.cmp_to_key(_compare_versions))
-
-
-def _compare_versions(v1, v2):
-    """Compare two version number strings."""
-    first_version = parse_version(v1)
-    second_version = parse_version(v2)
-    return (first_version > second_version) - (first_version < second_version)
+    kernel_version_list.sort(key=parse_version)
```

Let me know if I shall push that here or stop exploring :)

hroncok:
Alternatively, the entire RPMVersion class could be replaced with:

```python
@functools.cmp_to_key(cmp)
def cmp(v1, v2):
    v1, _, r1 = v1.partition('-')
    v2, _, r2 = v2.partition('-')
    return rpm.labelCompare(('0', v1, r1), ('0', v2, r2))
```

Which would be arguably much less fancy, but good enough.

VladimirSlavik:
I'm not going to be the one making the decision, but simple sounds better than fancy.

hroncok:
> I'm not going to be the one making the decision, but simple sounds better than fancy.

[Simple is better than complex](https://peps.python.org/pep-0020/)

Opened https://github.com/rhinstaller/anaconda/pull/4540 for comparison.

hroncok:
My take is: I'd like the RPMVersion better if we were designing a library API for others. But since this is internal, I prefer the simpler approach.

hroncok:
Closing in favor of https://github.com/rhinstaller/anaconda/pull/4540 due to an apparent consensus that simpler is better.

PR #4540, opened 2023-02-06
Sort RPM versions via rpm.labelCompare() and not via packaging.version.LegacyVersion() [SIMPLER]

hroncok:
Packaging 22+ removed the long-deprecated packaging.version.LegacyVersion class. The packaging.version library is intended to parse Python (PEP 440) versions, not arbitrary versions and definitively not RPM versions.

Even if LegacyVersion was still available, it may produce incorrect results if the version contains ~, ^ or other characters with special meaning in RPM.

This assumes the Python rpm module is always available. If that assumption is wrong
the logic from rpm.labelCompare() needs to be reimplemented instead.

--------------------

Intentionally opening this as a draft for discussion, so not following the checklist below yet.

Please check that your PR follows these rules:

* [x] [**Code conventions**](https://anaconda-installer.readthedocs.io/en/latest/contributing.html#code-conventions). tl;dr: Follow PEP8, wrap at 100 chars, and provide docstrings.

* [x] [**Commit message conventions**](https://anaconda-installer.readthedocs.io/en/latest/commit-log.html). tl;dr: Heading, empty line, longer explanations, all wrapped manually. If in doubt, write a longer commit message with more details.

* [x] **Tests** pass and cover your changes. You can [run tests locally manually](https://anaconda-installer.readthedocs.io/en/latest/testing.html), or have them run as part of the PR. A team member will have to enable tests manually after inspecting the PR.
  *If you don't know how, ask us for help!*

* [ ] [**Release notes**](https://anaconda-installer.readthedocs.io/en/latest/contributing.html#release-notes) and **docs** if the PR adds something major or changes existing behavior.
  *If you don't know how, ask us for help!*

* <del>[ ] **RHEL** rules: If your PR is for a `rhel-*` branch, pay special attention to commit messages. Make sure **all** commit messages include a line linking the commit to a bug, such as `Resolves: rhbz#123456`. For more information, see [the complete rules](https://anaconda-installer.readthedocs.io/en/latest/commit-log.html).</del>

### DIFF
--- a/anaconda.spec.in
+++ b/anaconda.spec.in
@@ -101,7 +101,6 @@ Requires: python3-pwquality
 Requires: python3-systemd
 Requires: python3-productmd
 Requires: python3-dasbus >= %{dasbusver}
-Requires: python3-packaging
 Requires: flatpak-libs
 %if %{defined rhel} && %{undefined centos}
 Requires: subscription-manager >= %{subscriptionmanagerver}
--- a/pyanaconda/core/payload.py
+++ b/pyanaconda/core/payload.py
@@ -16,13 +16,18 @@
 # Red Hat, Inc.
 #
 from collections import namedtuple
+from functools import cmp_to_key
 from urllib.parse import quote, unquote
+
+import rpm
 
 from pyanaconda.core.i18n import _
 from pyanaconda.core.regexes import URL_PARSE
 from pyanaconda.core.string import split_in_two
 
 NFSUrl = namedtuple("NFSUrl", ["options", "host", "path"])
+
+rpm_version_key = cmp_to_key(rpm.labelCompare)
 
 
 def parse_hdd_url(url):
--- a/pyanaconda/modules/payloads/base/utils.py
+++ b/pyanaconda/modules/payloads/base/utils.py
@@ -17,21 +17,11 @@
 # License and may only be used or replicated with the express permission of
 # Red Hat, Inc.
 #
-import functools
-
-from packaging.version import LegacyVersion as parse_version
-
 from pyanaconda.anaconda_loggers import get_module_logger
+from pyanaconda.core.payload import rpm_version_key
 log = get_module_logger(__name__)
 
 
 def sort_kernel_version_list(kernel_version_list):
     """Sort the given kernel version list."""
-    kernel_version_list.sort(key=functools.cmp_to_key(_compare_versions))
-
-
-def _compare_versions(v1, v2):
-    """Compare two version number strings."""
-    first_version = parse_version(v1)
-    second_version = parse_version(v2)
-    return (first_version > second_version) - (first_version < second_version)
+    kernel_version_list.sort(key=rpm_version_key)
--- a/pyanaconda/modules/storage/checker/utils.py
+++ b/pyanaconda/modules/storage/checker/utils.py
@@ -20,7 +20,6 @@ gi.require_version("BlockDev", "2.0")
 from gi.repository import BlockDev as blockdev
 
 from collections import defaultdict
-from packaging.version import LegacyVersion as parse_version
 
 from blivet import arch, util
 from blivet.devicefactory import get_device_type
@@ -34,6 +33,7 @@ from pyanaconda.core.constants import productName, STORAGE_REFORMAT_BLOCKLIST, \
     STORAGE_LUKS2_MIN_RAM, STORAGE_ROOT_DEVICE_TYPES, STORAGE_REQ_PARTITION_SIZES, \
     STORAGE_MUST_NOT_BE_ON_ROOT
 from pyanaconda.core.i18n import _
+from pyanaconda.core.payload import rpm_version_key
 from pyanaconda.core.storage import DEVICE_TEXT_MAP
 from pyanaconda.modules.storage.platform import platform
 
@@ -262,7 +262,7 @@ def _check_opal_firmware_kernel_version(detected_version, required_version):
     """
     try:
         if detected_version and required_version:
-            return parse_version(detected_version) >= parse_version(required_version)
+            return rpm_version_key(detected_version) >= rpm_version_key(required_version)
     except Exception as e:  # pylint: disable=broad-except
         log.warning("Couldn't check the firmware kernel version: %s", str(e))
 


hroncok:
pylint says:

```
E1120(no-value-for-parameter):/tmp/anaconda/pyanaconda/modules/storage/checker/utils.py:265,19: _check_opal_firmware_kernel_version: No value for argument 'v2' in function call
E1120(no-value-for-parameter):/tmp/anaconda/pyanaconda/modules/storage/checker/utils.py:265,56: _check_opal_firmware_kernel_version: No value for argument 'v2' in function call
```

That's rather weird, as `rpm_version_key` no longer takes 2 arguments.

I guess if `rpm_version_key` was not decorated, it might not confuse pylint?

I.e. like this:

```python
def rpm_version_cmp(v1, v2):
    # TODO do we need to parse epoch? nothing in the tests indicates that
    v1, _, r1 = v1.partition('-')
    v2, _, r2 = v2.partition('-')
    return rpm.labelCompare(('0', v1, r1), ('0', v2, r2))

rpm_version_key = cmp_to_key(rpm_version_cmp)
```

Should I try?

jkonecny12:
I like this solution a lot. Smaller is in general better. Maybe adding some debug prints to easily spot what we are comparing to be able to detect possible issue easily.

I'm a bit afraid of this change because it's indirectly touching a lot of use-cases which might not be easy to spot or test...  However, I'm more inclined to go with this version and test it by automated tests and try it in Rawhide.

The `epoch` question, I think we should definitely take the `epoch` into account. If it is not handled right now, it's probably because someone missed that use-case. WDYT @poncovka ?

jkonecny12:
Also thanks a lot @hroncok for helping us with this!

pyanaconda/modules/common/versions.py
```diff
@@ -0,0 +1,16 @@
...
+
+
+def rpm_version_cmp(v1, v2):
+    # TODO do we need to parse epoch? nothing in the tests indicates that
+    v1, _, r1 = v1.partition('-')
+    v2, _, r2 = v2.partition('-')
+    return rpm.labelCompare(('0', v1, r1), ('0', v2, r2))
+
+
+rpm_version_key = cmp_to_key(rpm_version_cmp)
```
poncovka, line 16: Please, move this code to `pyanaconda.core.payload`.

pyanaconda/modules/common/versions.py
```diff
@@ -0,0 +1,16 @@
...
+
+
+__all__ = ['rpm_version_key']
+
+
+def rpm_version_cmp(v1, v2):
+    # TODO do we need to parse epoch? nothing in the tests indicates that
+    v1, _, r1 = v1.partition('-')
+    v2, _, r2 = v2.partition('-')
+    return rpm.labelCompare(('0', v1, r1), ('0', v2, r2))
```
poncovka, line 13: I was playing around with the `rpm.labelCompare` function and it looks like it is able to handle strings. Therefore, it is not necessary to parse them and we can create the key function directly from `rpm.labelCompare`.

poncovka (changes requested):
I prefer the simpler version :-)

pyanaconda/modules/common/versions.py
hroncok: WOW. Indeed. I think it wasn't always the case, but I've checked with RPM 4.17 on Fedora 36 and it even works there.

hroncok: Bloody hell that makes the code super simple.

hroncok:
> The `epoch` question, I think we should definitely take the `epoch` into account. If it is not handled right now, it's probably because someone missed that use-case. WDYT @poncovka ?

Thanks to @poncovka's finding, epoch is now handled transparently :)

pyanaconda/modules/common/versions.py
hroncok: Moved.

poncovka:
> I like this solution a lot. Smaller is in general better. Maybe adding some debug prints to easily spot what we are comparing to be able to detect possible issue easily.

Yeah, let's not do that, please.

> I'm a bit afraid of this change because it's indirectly touching a lot of use-cases which might not be easy to spot or test... However, I'm more inclined to go with this version and test it by automated tests and try it in Rawhide.

That is not true. It is affecting only two places and we have unit tests for both of them.
* `_check_opal_firmware_kernel_version` -> `test_check_opal_firmware_kernel_version`
* `sort_kernel_version_list` -> `test_sort_kernel_version_list`

Therefore, I am not too worried about this change.

pyanaconda/modules/common/versions.py
hroncok: For clarity.

### RHEL 8

```pycon
>>> import rpm
>>> rpm.labelCompare("1", "2")
TypeError: argument 1 must be sequence of length 3, not 1
```

### RHEL 9

```pycon
>>> import rpm
>>> rpm.labelCompare("1", "2")
-1
```

jkonecny12:
> > I like this solution a lot. Smaller is in general better. Maybe adding some debug prints to easily spot what we are comparing to be able to detect possible issue easily.
> 
> Yeah, let's not do that, please.

OK, I don't have a strong opinion about this.

> 
> > I'm a bit afraid of this change because it's indirectly touching a lot of use-cases which might not be easy to spot or test... However, I'm more inclined to go with this version and test it by automated tests and try it in Rawhide.
> 
> That is not true. It is affecting only two places and we have unit tests for both of them.
> 
> * `_check_opal_firmware_kernel_version` -> `test_check_opal_firmware_kernel_version`
> * `sort_kernel_version_list` -> `test_sort_kernel_version_list`
> 
> Therefore, I am not too worried about this change.

Yes, we have unit tests. I was more talking about where the `sort_kernel_version_list` is used (all the payloads AFAIK). However, good point that we should have the use-case covered by unit tests.

BTW, I wanted to look on the failing unit tests. Unfortunately, I'm not able to build a Rawhide container because of `iputils` transaction error (broken repositories?). Will try this tomorrow again.

```
Error unpacking rpm package iputils-20221126-2.fc38.x86_64
  Verifying        : iputils-20221126-2.fc38.x86_64                                                                                                                                                                                        1/1 

Failed:
  iputils-20221126-2.fc38.x86_64                                                                                                                                                                                                               

Error: Transaction failed
```

poncovka:
> BTW, I wanted to look on the failing unit tests. Unfortunately, I'm not able to build a Rawhide container because of `iputils` transaction error (broken repositories?). Will try this tomorrow again.

The unit tests are working. The fix is available the blivet's daily builds and I have already refreshed the testing container images.

poncovka (approved):
Looks good to me. Thank you!

hroncok:
Do I backport this into Fedora as a patch or wait for a release? I'd like to tag packaging 23.0 back this week and if I can do it before branching, even better.

VladimirSlavik (approved):
Looks good to me. Thank you!

pyanaconda/modules/common/versions.py
hroncok: https://github.com/rpm-software-management/rpm/pull/1221

poncovka:
> Do I backport this into Fedora as a patch or wait for a release? I'd like to tag packaging 23.0 back this week and if I can do it before branching, even better.

I will do a new release today.

hroncok:
https://bodhi.fedoraproject.org/updates/FEDORA-2023-0540c0e02f

Thank you!

PR #4546, opened 2023-02-08
Be more indulgent when reclaiming disk space

jstodola:

Related: rhbz#1855371

### DIFF
--- a/pyanaconda/ui/gui/spokes/lib/resize.py
+++ b/pyanaconda/ui/gui/spokes/lib/resize.py
@@ -83,7 +83,6 @@ class ResizeDialog(GUIObject):
         required_size = self._device_tree.GetRequiredDeviceSize(required_space)
 
         self._required_size = Size(required_size)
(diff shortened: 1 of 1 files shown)


pyanaconda/ui/gui/spokes/lib/resize.py
```diff
@@ -419,7 +419,7 @@ def _update_action_buttons(self):
             self._delete_button.set_sensitive(False)
 
     def _update_reclaim_button(self, got):
-        self._resize_button.set_sensitive(got + self._initial_free_space >= self._required_size)
```
poncovka, line 422: It looks like you can completely remove `self._initial_free_space`, because this is the only place where we use its value.

jstodola: Right, PR updated.

poncovka (approved):
Thanks!

M4rtinK (approved):

PR #4548, opened 2023-02-09
Web UI: Make Pre-release label look clickable

OndrejZobal:
Hovering over the Pre-release label displays text selection cursor
instead of a pointer. Since PF doesn't currently implement button-like
labels, I manually assigned it the necessary CSS tags. Once PF adds them
we should switch to using them instead.

### DIFF
--- a/ui/webui/src/components/AnacondaHeader.jsx
+++ b/ui/webui/src/components/AnacondaHeader.jsx
@@ -52,7 +52,10 @@ export const AnacondaHeader = ({ beta, title }) => {
                   </TextContent>
               }
             >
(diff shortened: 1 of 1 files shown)


M4rtinK (approved):
Looks good to me, thanks! :)

VladimirSlavik (approved):
Thank you!

Can't you put the inline style on the Label itself?

OndrejZobal:
@VladimirSlavik 

> Can't you put the inline style on the Label itself?

No, when the JSX is translated into HTML the inline style gets assigned to the tiny information icon inside of the label instead of the label itself.

PR #4549, opened 2023-02-11
fix the failure to mount ISO with UDF filesystem

century6:
Blivet has added support for UDF file system, which makes it possible to install OS by using UDF ISO. However, anaconda forces the file system to be set to iso9660 when mounting /run/install/resources/mount-0000-cdrom, which will cause the mount to fail. Let's fix it.

### DIFF
--- a/pyanaconda/modules/payloads/source/cdrom/initialization.py
+++ b/pyanaconda/modules/payloads/source/cdrom/initialization.py
@@ -19,7 +19,7 @@ from pyanaconda.core.kernel import kernel_arguments
 from pyanaconda.modules.common.constants.objects import DEVICE_TREE
 from pyanaconda.modules.common.constants.services import STORAGE
 from pyanaconda.modules.common.errors.payload import SourceSetupError
-from pyanaconda.modules.common.structures.storage import DeviceData
+from pyanaconda.modules.common.structures.storage import DeviceData, DeviceFormatData
 from pyanaconda.modules.payloads.source.mount_tasks import SetUpMountTask
 from pyanaconda.modules.payloads.source.utils import is_valid_install_disk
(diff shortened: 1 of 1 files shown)


VladimirSlavik:
@century6 Thank you!

century6:
> /kickstart-test --testtype smoke

Some cases of unit-tests failed, perhaps because format_data may be None? I have submitted a new commit to fix it.

VladimirSlavik:
I think the unit tests expect specific calls to happen, so that might be the issue.

At this moment an unrelated group of tests started failing too, so disregard anything related to users and groups there.

VladimirSlavik:
@century6 I can't promise fixing the unit tests for you. Is that something you can do?

century6:
> @century6 I can't promise fixing the unit tests for you. Is that something you can do?

Well, I will try to fix it. But it may take some time.

VladimirSlavik:
@vojtechtrefny any ideas what to do about this? I see the "mount with existing format" as potentially too open. Perhaps we could check if it's in a list of known handle-able formats (realistically iso9660 & udf?) and in that case use the format, otherwise force iso9660 to preserve existing behavior?

vojtechtrefny:
Does this even work? I am looking at the code and the `devices_candidates` list comes from [find_optical_media](https://github.com/rhinstaller/anaconda/blob/7c9aa594331553e4fee8a4d66559ce0b1c88f1d4/pyanaconda/modules/storage/devicetree/utils.py#L171) which already filters out devices that are not iso9660 so udf formatted devices will be ignored anyway if I am not wrong. I'd recommend changing `find_optical_media` to work with both iso9660 and udf and then we could simply omit the filesystem type from the `mount` call.

PR #4530, opened 2023-01-27
docs: Update branching instructions

VladimirSlavik:
We now use templates to do things.

PR #4551, opened 2023-02-13
Templatize pykickstart version in dracut code

VladimirSlavik:
This is missing from templating to make a branching switch automatic.

(Probably among other things...)

Fixed thanks to @poncovka in https://github.com/rhinstaller/anaconda/pull/4530#pullrequestreview-1277125546


VladimirSlavik:
The variant where the script does `from pyanaconda.core.kickstart.version import VERSION` would add lots more to initramfs (need a file = add whole dir), so that's not desirable.

I call for a bikeshed session on the new file name 🎉

rvykydal (approved):
Looks good to me, thank you.

dracut/module-setup.sh
```diff
@@ -58,6 +58,7 @@ install() {
     inst "$moddir/fetch-kickstart-disk" "/sbin/fetch-kickstart-disk"
     inst "$moddir/fetch-updates-disk" "/sbin/fetch-updates-disk"
     inst "$moddir/parse-kickstart" "/sbin/parse-kickstart"
+    inst "$moddir/version.py" "/sbin/version.py"
```
poncovka, line 61: I am not sure if you need to do this. We don't run this code as a script. It needs to be installed as a Python library. See the end of this file.

dracut/parse-kickstart
```diff
@@ -41,7 +41,7 @@ from pykickstart.constants import *
 from collections import OrderedDict
 
 # Import the kickstart version.
-from pykickstart.version import DEVEL as VERSION
+from version import VERSION
```
poncovka, line 44: I guess it could be at least `kickstart_version` to avoid possible future conflicts.

poncovka (approved):

dracut/module-setup.sh
VladimirSlavik: Let's see how you like the one-liner needed to get a suitable directory :D

PR #4555, opened 2023-02-14
infra: Change branched fedora branch name

VladimirSlavik:
Let's change all infra to use the new fedora-NN branch name scheme.

See also: https://anaconda-installer.readthedocs.io/en/latest/release.html#branching-for-the-next-fedora-release


poncovka (approved):
Thanks!

VladimirSlavik:
FWIW, there's a lot of references to the old branching scheme in CONTRIBUTING.rst but that does not matter in making this work, which is needed asap.

PR #4557, opened 2023-02-15
infra: Fix branch name handling in container builds

VladimirSlavik:
We now have `fedora-XX` instead of `fXX-<release|devel>`. Adjust accordingly.

@M4rtinK this would be great to get before branching too, or we'd have to port it.


M4rtinK (approved):
Looks good to me. :)

poncovka (approved):
Thanks!

jkonecny12 (approved):
LGTM!

PR #4564, opened 2023-02-17
Add SYS_CHROOT capability to anaconda-ci (#infra)

jkonecny12:
We have user creation container tests which are using chroot to be able to use tooling to create users. However, from podman 4.4 they removed this capability from default capabilities set. Let's put it back for this specific container.

Podman 4.4.1 is on Fedora 38+.

See https://bugzilla.redhat.com/show_bug.cgi?id=2170568 for more info.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -276,21 +276,27 @@ ci:
 	fi
 
 # container-ci target will have disabled cockpit Makefiles by default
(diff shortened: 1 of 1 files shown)


VladimirSlavik (approved):
Thanks!

poncovka (approved):

M4rtinK (approved):
Looks good to me as well. :)

PR #4566, opened 2023-02-20
master - Move loop device pre-creation to container (#infra)

jkonecny12:
We have to pre-create loop devices because they are not available in the containers. If they are created by `losetup` than the newly created loop device is not visible inside the container. However, it still could be created manually in the container and taken correctly by the losetup tooling.

We need to merge this to all supported branches because the GitHub workflow file is always used from the `master` branch.
Needs to merged together with:
- https://github.com/rhinstaller/anaconda/pull/4579
- https://github.com/rhinstaller/anaconda/pull/4580
- https://github.com/rhinstaller/anaconda/pull/4581

### DIFF
--- a/.github/workflows/build-boot-iso.yml
+++ b/.github/workflows/build-boot-iso.yml
@@ -113,15 +113,6 @@ jobs:
           mkdir -p ./anaconda_rpms/
           cp -av ./result/build/01-rpm-build/*.rpm ./anaconda_rpms/
 
-      - name: Prepare environment for lorax run
-        run: |
-          mkdir -p images
-          # We have to pre-create loop devices because they are not namespaced in kernel so
-          # podman can't access newly created ones. That caused failures of tests when runners
-          # were rebooted.
-          sudo mknod -m 0660 /dev/loop0 b 7 0  2> /dev/null || true
-          sudo mknod -m 0660 /dev/loop1 b 7 1  2> /dev/null || true
-
       - name: Build the boot.iso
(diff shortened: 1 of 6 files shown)


jkonecny12:
@VladimirSlavik do we want to backport this to Fedora-38 too?

jkonecny12:
Tested but I would like to do more testing with a rebooted system.

VladimirSlavik (commented):
This is good, thank you! A couple notes:
- There are far more workflows that do this. Search for `sudo mknod` and there you go.
- We definitely need to port this to 38 too, because the workflows are all here on master but the containers are on that branch. (IIRC)
- I wonder what's the requirements on host to do this? Can we do it with rawhide container and a f34 host? We still have these deployed...

jkonecny12:
> This is good, thank you! A couple notes:
> 
> * There are far more workflows that do this. Search for `sudo mknod` and there you go.

Should be covered now.

> * We definitely need to port this to 38 too, because the workflows are all here on master but the containers are on that branch. (IIRC)

Good point. I should also check rhel branches.

> * I wonder what's the requirements on host to do this? Can we do it with rawhide container and a f34 host? We still have these deployed...

There are no new requirements on the host. It's the same what we were doing even before but just better way.

jkonecny12:
Seems to be working just fine based on my local testing.

VladimirSlavik (approved):
Thanks, let's try this, I think...

PR #4567, opened 2023-02-20
Simplify the default source selection of the RPM sources

poncovka:
Don't use defaults and fallbacks during the payload setup. Choose the default
source early and only once instead. If the chosen default source is not usable,
a user will have to change it via the user interface or reboot with a different
configuration.

With this change, the installer chooses a default source of the DNF payload
based on data processed in this order:

1. **Kickstart file**: If there is a source defined in a kickstart file, use it.
2. **Boot options or command line options**: If there is a source defined in boot
   options or cmdline options, use it. In the future, this may have a higher
   preference than a kickstart file to unify the experience with other boot
   options.
3. **Installation image mounted by Dracut**: If the installer is booted from an
   image with a repository (for example, the full DVD ISO), use the Dracut's
   mount point at /run/install/repo to access this repository.
4. **Anaconda configuration file**: If none of the options above is available, use
   the default source defined in the Anaconda configuration file (the closest
   mirror or CDN).

Previously, we were trying to find the installation media and mount it again
using the CDROM source instead of using the Dracut's mount point. That caused
several issues. The installation media could have been different, the new mount
could have failed or Blivet didn't have to see the media because it was marked
as hidden due to other kickstart commands.

**TODO:**
- [x] Test common use cases.
- [x] Test edge cases related to the stage2 issue.
- [x] Agree on the representation of the new source in the UI.


poncovka:
Screenshot_vpn_test_04_2023-02-17_15:13:12

These options are visible if you boot from a full DVD ISO:
Screenshot_vpn_test_04_2023-02-17_15:29:42

These options are visible if you boot from a netinst ISO:
Screenshot_vpn_test_03_2023-02-17_15:15:53

poncovka:
@jkonecny12 @jstodola Do you have any concerns/suggestions?

jstodola:
I do!
What happens if the installer is booted with `inst.stage2=nfs:<server>:<path>`? Is `Local installation source` selected by default? Or does anaconda select the `On the network` option, although the installation repository was not specified?

Is there any additional information provided when the `Local installation source` option is selected? As a user I would like to see what is the local installation source. Let's say I have an unpacked DVD.iso content placed on a disk, as well as DVD media attached to the system. The installer is booted with `inst.stage2=hd:<device>:<path>`.  How can I tell if the installer will use the "hd" source or DVD media?

VladimirSlavik (approved):
Looks good to me. Thank you!

poncovka:
> I do! What happens if the installer is booted with `inst.stage2=nfs:<server>:<path>`? Is `Local installation source` selected by default? Or does anaconda select the `On the network` option, although the installation repository was not specified?

The installer will select the `Local installation source` option.
 
> Is there any additional information provided when the `Local installation source` option is selected? As a user I would like to see what is the local installation source. Let's say I have an unpacked DVD.iso content placed on a disk, as well as DVD media attached to the system. The installer is booted with `inst.stage2=hd:<device>:<path>`. How can I tell if the installer will use the "hd" source or DVD media?

With the new solution, the installer will not try to use an unknown CD/DVD media as an installation source by default. Users have to specify the `inst.repo=cdrom` boot option or the `cdrom` kickstart option to enable this behaviour.

jstodola:
> > I do! What happens if the installer is booted with `inst.stage2=nfs:<server>:<path>`? Is `Local installation source` selected by default? Or does anaconda select the `On the network` option, although the installation repository was not specified?
> 
> The installer will select the `Local installation source` option.

So, based on the fact that the `Local installation source` option could be selected for an NFS (remote) installation source, it should be renamed. What about `Auto-detected installation source`? Unfortunately it's quite similar as `Auto-detected installation media` for removable media, so it could be confusing, but I cannot think of anything better right now.

> 
> > Is there any additional information provided when the `Local installation source` option is selected? As a user I would like to see what is the local installation source. Let's say I have an unpacked DVD.iso content placed on a disk, as well as DVD media attached to the system. The installer is booted with `inst.stage2=hd:<device>:<path>`. How can I tell if the installer will use the "hd" source or DVD media?
> 
> With the new solution, the installer will not try to use an unknown CD/DVD media as an installation source by default. Users have to specify the `inst.repo=cdrom` boot option or the `cdrom` kickstart option to enable this behaviour.

OK, sounds good.

poncovka:
> So, based on the fact that the `Local installation source` option could be selected for an NFS (remote) installation source, it should be renamed. What about `Auto-detected installation source`? Unfortunately it's quite similar as `Auto-detected installation media` for removable media, so it could be confusing, but I cannot think of anything better right now.

I don't really disagree, but I want to mention some notes before we commit to a new description.

The source is "local" from the installer's point of view. The repository is there within its reach without any additional configuration steps. It also means that there is really nothing to detect. There is no complicated search for the source, we just check one specific mount point and, if it is not it, we move on.

Also, there are two strings related to this:
* `Local installation source` in options of the Installation Source spoke
* `Local source` in a status of the Installation Summary hub

jkonecny12:
I'm personally more in favor of the @jstodola proposal. Your points are correct @poncovka but they are not really interesting to user. Users don't care how complex the logic is or if it is local from Anaconda PoV, they just need to understand it from installation PoV.

VladimirSlavik:
I'll admit that I don't follow all the details, but at the final layer that is UI, calling NFS "local" does not work for me either. Maybe it could be called "manually configured outside UI" or some such?

If the bikeshed is literally about the name on the radiobutton and nothing else, I'm tempted to call this fine and have another PR just for the cosmetics, while the functionality is being tested.

poncovka:
The source description for the Summary hub has changed:
Screenshot_vpn_test_04_2023-03-15_18:38:55

The option on the Installation Source spoke has changed:
Screenshot_vpn_test_04_2023-03-15_18:39:10

poncovka:
Updated. The strings have changed and there is one new commit. The DBus API could look better, but that can be done later. If you don't have any major objections, I would like to get these changes to Rawhide as they are for now.

jstodola (approved):
Thanks

VladimirSlavik (approved):
Thank you, the naming looks good to me.

Looking at the screenshots, what I'm missing is that I don't know what the auto-detected source is. But that's out of scope here.

PR #4568, opened 2023-02-20
infra: Use fully qualified names of exceptions in pylintrc

poncovka:
Specifying exception names in the overgeneral-exceptions option without module name is deprecated and support for it will be removed in pylint 3.0. Use fully qualified name instead.


VladimirSlavik (approved):
Thanks! I wondered about that message but didn't research it yet.

VladimirSlavik:
To state the obvious, I think the cockpit tests are totally borked by something, independent of this of course.

PR #4572, opened 2023-02-21
infra: Pylint should use the GTK 3 library to analyze our code

poncovka:
It looks like there is GTK 3 and 4 in our upstream testing containers, so make
sure that pylint uses the right version of GTK to analyse our code. Otherwise,
it will report false positives.


VladimirSlavik (approved):
Thanks!

VladimirSlavik:
We might need the same on 38 :(

PR #4575, opened 2023-02-22
Add test step logging and screenshots to WebUI tests

velezd:
Adding decorator that prints the name of a function and its parameters (test step), and optionally prints docstring and creates screenshots. Also decorating all existing helper functions that interact with the WebUI.

From now on we should use the decorator for any new helper function that we create.

### DIFF
--- a/ui/webui/test/helpers/step_logger.py
+++ b/ui/webui/test/helpers/step_logger.py
@@ -1,0 +1,48 @@
+
+class BrowserSnapshot():
+    SNAPSHOT_NUMBER = 0
+
+    @classmethod
+    def new(cls, browser):
+        browser.snapshot(f'snapshot-{browser.label}', str(cls.SNAPSHOT_NUMBER))
+        cls.SNAPSHOT_NUMBER += 1
+
+
+def log_step(snapshots=False, snapshot_before=False, snapshot_after=False, docstring=False):
+    """ Decorator for logging steps durring testing.
+
+    Decorated function needs to be part of class with self.browser.
+
+    :param snapshots: Create snapshots before and after function call, defaults to False
+    :type snapshots: bool, optional
+    :param snapshot_before: Create snapshots before function call, defaults to False
+    :type snapshot_before: bool, optional
+    :param snapshot_after: Create snapshots after function call, defaults to False
+    :type snapshot_after: bool, optional
+    :param docstring: Print docstring of the function, defaults to False
+    :type docstring: bool, optional
+    """
+    def decorator(function):
+
+        def wrapper(*args, **kwargs):
+            nice_args = ', '.join(str(a) for a in args[1:])
+            if kwargs:
+                nice_args += ', ' + ', '.join(f'{k}: {v}' for k, v in kwargs.items())
+            if nice_args:
+                nice_args = ', with ' + nice_args
+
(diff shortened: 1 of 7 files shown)


velezd:
I'm not sure what do you mean by storing. The screenshots are created by this function [/common/testlib.py#L950](https://github.com/cockpit-project/cockpit/blob/41f8035eaf59d0a0ace6b86684bd8bd5c1444694/test/common/testlib.py#L950), and whatever runs the test has to collect them. If run by cockpit CI, they are stored for example [here](https://cockpit-logs.us-east-1.linodeobjects.com/pull-4575-20230223-105157-3c18028c-fedora-rawhide-boot/index.html), if run by Permian it creates test report similar to what anabot did and saves it with other logs.

ui/webui/test/helpers/step_logger.py
```diff
@@ -0,0 +1,47 @@
...
+            if snapshots or snapshot_before:
+                BrowserSnapshot.new(args[0].browser)
+            
+            function(*args, **kwargs)
+            
+            if snapshots or snapshot_after:
+                BrowserSnapshot.new(args[0].browser)
+
+        return wrapper
+    return decorator
```
VladimirSlavik, line 46: Might be nice to have one empty line here.

VladimirSlavik (approved):
Looks good to me. Thank you!

(It took me some looking to realize the screenshots are not assertions, but just logging.)

ui/webui/test/helpers/step_logger.py
```diff
@@ -0,0 +1,47 @@
...
+            if nice_args:
+                nice_args = ', with ' + nice_args
+
+            print(f'[TEST STEP] {function.__name__}{nice_args}')
+            if docstring:
+                print(f'[DOC] {function.__doc__}')
+            if snapshots or snapshot_before:
+                BrowserSnapshot.new(args[0].browser)
+            
+            function(*args, **kwargs)
```
marusak, line 40: This function can have return value. Like for example [get_local_search](https://github.com/rhinstaller/anaconda/blob/master/ui/webui/test/helpers/language.py#L41) has. When you don't return it, python by default returns `None`. Which is why you see [here](https://github.com/rhinstaller/anaconda/pull/4575/files#diff-af148855d642cc07a588883e506315f3f74ff4f0ab87b832028ff650256ad626R69) `None`.

This simple fix makes the one failing test to pass:
```diff
--- ui/webui/test/helpers/step_logger.py
+++ ui/webui/test/helpers/step_logger.py
@@ -37,11 +37,13 @@ def log_step(snapshots=False, snapshot_before=False, snapshot_after=False, docst
             if snapshots or snapshot_before:
                 BrowserSnapshot.new(args[0].browser)
             
-            function(*args, **kwargs)
+            result = function(*args, **kwargs)
             
             if snapshots or snapshot_after:
                 BrowserSnapshot.new(args[0].browser)
 
+            return result
+
         return wrapper
     return decorator

```

velezd: Good catch, thank you.

VladimirSlavik:
@M4rtinK are you ok with merging this?

rvykydal (approved):
Looks good to me. Thank you.

PR #4576, opened 2023-02-23
tooltip text of password adapt with language

iasunsea:
When we selected other language, password tooltip text with "PASSWORD_HIDE" and "PASSWORD_SHOW" do not follow the language to change. They do not adapt the language, So to change they to adapt language.


poncovka (approved):
Looks good to me. Thank you!

VladimirSlavik (approved):
Thanks!

PR #4578, opened 2023-02-28
Fix sorting of the potfile (#infra)

jkonecny12:
Right now when we generate potfile the output file is randomly sorted. By adding --sort-output keyword it should stay sorted always the same way. That will allow us easily detect if the potfile changed.

Potfile change detection can be then used in our potfile automation and also it's much better visible in the git history.

Later we can improve potfile change detection by similar [way as I have on simpleline](https://github.com/rhinstaller/python-simpleline-l10n/blob/master/.github/workflows/pot-file-update.yml#L66).


poncovka (approved):
Thank you!

PR #4582, opened 2023-03-01
Clean up ISO option in source selection spoke

OndrejZobal:
Update ISO option in the source selection spoke as suggested by Vendula Poncova here: https://github.com/rhinstaller/anaconda/discussions/4257

obrazek

### DIFF
--- a/pyanaconda/ui/gui/spokes/installation_source.glade
+++ b/pyanaconda/ui/gui/spokes/installation_source.glade
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<!-- Generated with glade 3.38.2 -->
+<!-- Generated with glade 3.40.0 -->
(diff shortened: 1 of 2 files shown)


poncovka (approved):
Looks good me. Thank you! Btw. does the tooltip work on its own? It would be good to confirm that users can still access a full device description since that label is ellipsized now.

M4rtinK (approved):
Looks good to me as well. :)

OndrejZobal:
> Looks good me. Thank you! Btw. does the tooltip work on its own? It would be good to confirm that users can still access a full device description since that label is ellipsized now.

@poncovka  Sorry, I have goldfish memory. I updated the pull request with the tooltip code. You might want to review it again. Thanks a lot for catching this.

poncovka (approved):
Thank you!

PR #4583, opened 2023-03-01
Load treeinfo metadata with a task

poncovka:
Use the `LoadTreeInfoMetadataTask` task to load the treeinfo metadata
of an installation source and to process the collected data. Clean up
the related code to make it ready for the modularization.


M4rtinK (approved):
Looks good to me. :)

PR #4584, opened 2023-03-01
Validate the CDN source early

poncovka:
Check if the CDN source is usable in early stages of the payload setup, so we don't have deal with an invalid source later.
This doesn't change the behaviour of the spokes, because the CDN source is treated as a special case.

**TESTED ON RHEL 9**

### DIFF
--- a/pyanaconda/modules/payloads/source/cdn/initialization.py
+++ b/pyanaconda/modules/payloads/source/cdn/initialization.py
@@ -1,0 +1,51 @@
+#
+# Copyright (C) 2023 Red Hat, Inc.
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# the GNU General Public License v.2, or (at your option) any later version.
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY expressed or implied, including the implied warranties of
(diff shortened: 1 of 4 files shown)


pyanaconda/modules/payloads/source/cdn/initialization.py
```diff
@@ -0,0 +1,47 @@
...
+    def run(self):
+        """Run the task."""
+        log.debug("Validating the CDN source.")
+
+        if not is_module_available(SUBSCRIPTION):
+            raise SourceSetupError(_("Red Hat CDN is unavailable for this installation."))
+
+        subscription_proxy = SUBSCRIPTION.get_proxy()
+        if not subscription_proxy.IsSubscriptionAttached:
+            raise SourceSetupError(_("Red Hat CDN requires to connect to Red Hat."))
```
M4rtinK, line 47: I suggest a another wording:

"To access the Red Hat CDN a valid Red Hat subscription is required."

M4rtinK (approved):
Suggested a different wording for one of the messages - otherwise looks good to me. :)

pyanaconda/modules/payloads/source/cdn/initialization.py
poncovka: Fixed.

PR #4586, opened 2023-03-02
oemdrv: wait up to 5 seconds for disks to be handled

rmetrich:
OEMDRV disk is used for both Driver Update and Kickstart. When booting with the DVD and an additional OEMDRV disk, it appears that the udev rule setting the script to fetch the kickstart may not run at the time `/dev/root` was found on the DVD, causing the kickstart to not be processed at all.
This patch delays 'settled' initqueue by 5 seconds by default (tunable through `rd.disks_delay=<value>`) when no `inst.ks` is specified on the kernel command line, which is enough to process the OEMDRV disk.
Because the same issue may appear with Driver Updates, that same delay applies unless `inst.dd` is explicitly specified.
Old behavior can be restored with adding `rd.disks_delay=0` on the kernel command line.

See [RHBZ #2171811](https://bugzilla.redhat.com/show_bug.cgi?id=2171811) for details and reproducer.

### DIFF
--- a/dracut/anaconda-lib.sh
+++ b/dracut/anaconda-lib.sh
@@ -434,3 +434,18 @@ wait_for_updates() {
 wait_for_dd() {
     echo "[ -e /tmp/dd.done ]" > "$hookdir/initqueue/finished/dd.sh"
 }
+
+wait_for_disks() {
+    # Allow up to 'inst.wait_for_disks' seconds for disks to be enumerated and
+    # related udev rules to execute (defaults to 5 seconds, 0 disables the
+    # feature). This prevents dracut-initqueue from finishing early.
+    # Since a 0.5 second delay is used between two runs of dracut-initqueue, we
+    # force the latter to retry up to twice the value configured, e.g:
+    # 'inst.wait_for_disks=15' --> force looping 30 times at least
+    # 'inst.wait_for_disks=0'  --> force looping 0 times (so no wait time)
+    finished_hook="$hookdir/initqueue/finished/wait_for_disks.sh"
+    [ -e "$finished_hook" ] && return
+    DISKS_WAIT_DELAY=$(getargnum 5 0 10000 inst.wait_for_disks)
+    DISKS_WAIT_RETRIES=$((DISKS_WAIT_DELAY * 2))
+    echo "[ \"\$main_loop\" -ge \"$DISKS_WAIT_RETRIES\" ]" > "$finished_hook"
+}
--- a/dracut/README-driver-updates.md
+++ b/dracut/README-driver-updates.md
@@ -163,10 +163,14 @@ all devices have been found and declares the system "settled".
 If there's no response from any `OEMDRV` device by then, the installer starts
 normally.
 
-_If the `OEMDRV` device is present but too slow to be autodetected, the user
-can boot with an argument like `inst.dd=hd:LABEL=OEMDRV` to indicate that
-dracut should expect an `OEMDRV` device and not start the installer until it
-appears._
+_Because disks can take some time to appear, an additional delay of 5 seconds
+has been added.  This can be overridden by boot argument
+`inst.wait_for_disks=<value>` to let dracut wait up to <value> additional
+seconds (0 turns the feature off, causing dracut to only wait up to 500ms).
+Alternatively, if the `OEMDRV` device is known to be present but too slow to be
+autodetected, the user can boot with an argument like `inst.dd=hd:LABEL=OEMDRV`
+to indicate that dracut should expect an `OEMDRV` device and not start the
+installer until it appears._
 
 # DUD filesystem layout
 
--- a/dracut/driver-updates-genrules.sh
+++ b/dracut/driver-updates-genrules.sh
@@ -42,6 +42,7 @@ for dd in $DD_OEMDRV $DD_DISKS; do
             when_diskdev_appears "$(disk_to_dev_path "$dd")" \
                 driver-updates --disk "$dd_whitespace_fix" \$devnode
         fi
+        wait_for_disks
     fi
 done
 
--- a/dracut/kickstart-genrules.sh
+++ b/dracut/kickstart-genrules.sh
@@ -31,6 +31,7 @@ case "${kickstart%%:*}" in
         if [ -z "$kickstart" ] && [ -z "$(getarg inst.ks=)" ]; then
             when_diskdev_appears "$(disk_to_dev_path LABEL=OEMDRV)" \
                 fetch-kickstart-disk "\$env{DEVNAME}" "/ks.cfg"
+            wait_for_disks
         fi
     ;;
 esac


jkonecny12:
Hi @lnykryn @pvalena, what  do you think about this implementation? Is it fine to use `/proc/timer_list` this way?

dracut/anaconda-lib.sh
```diff
@@ -434,3 +434,14 @@ wait_for_updates() {
 wait_for_dd() {
     echo "[ -e /tmp/dd.done ]" > "$hookdir/initqueue/finished/dd.sh"
 }
+
+wait_for_disks() {
+    # Allow up to 'inst.wait_for_disks' seconds for disks to be enumerated and
+    # related udev rules to execute (defaults to 5 seconds, 0 disables the
+    # feature)
+    finished_hook="$hookdir/initqueue/finished/wait_for_disks.sh"
+    [ ! -e "$finished_hook" ] || return
```
lnykryn, line 443: [ -e "$finished_hook" ] && return 
is shorter and bit easier to understand

dracut/anaconda-lib.sh
```diff
@@ -434,3 +434,14 @@ wait_for_updates() {
...
+    DISKS_WAIT_DELAY=$(getargnum 5 0 10000 inst.wait_for_disks)
+    timeout=$(($(awk '/^now/ {print substr($3, 1, length($3) - 9); exit}' /proc/timer_list) + $DISKS_WAIT_DELAY))
+    echo "[ \$(awk '/^now/ {print substr(\$3, 1, length(\$3) - 9); exit}' /proc/timer_list) -ge $timeout ]" > $finished_hook
```
lnykryn, line 446: That is a really fancy way how to wait 5 seconds :-D Any reason why not just use /proc/uptime?
Also, in theory, since all the hooks are sourced you can share variables and you could do sleep 1; $((CNT++))

But this will definitely work, it is just hard to understand what the code does.

lnykryn (changes requested):

pvalena:
`cut -d'.' -f1 /proc/uptime` might be completely sufficient.

pvalena:
Alternatively there's a possibility to simply use `date +"%s%N"` (might be sensitive to time jumps).

I however agree with Lukas sleep (ideally in the background; to not block) would be a good solution also.

rmetrich:
> Alternatively there's a possibility to simply use `date +"%s%N"` (might be sensitive to time jumps).
> 
> I however agree with Lukas that `sleep` (ideally in the background; to not block) would be a good solution also.

`date` is not an option because there is no such binary :-)

dracut/anaconda-lib.sh
rmetrich: I will use `/proc/uptime`, but I don't want to sleep.

dracut/anaconda-lib.sh
```diff
@@ -434,3 +434,14 @@ wait_for_updates() {
...
 }
+
+wait_for_disks() {
+    # Allow up to 'inst.wait_for_disks' seconds for disks to be enumerated and
+    # related udev rules to execute (defaults to 5 seconds, 0 disables the
+    # feature)
+    finished_hook="$hookdir/initqueue/finished/wait_for_disks.sh"
+    [ -e "$finished_hook" ] && return
+    DISKS_WAIT_DELAY=$(getargnum 5 0 10000 inst.wait_for_disks)
+    timeout=$(($(awk -F '.' '{print $1}' /proc/uptime) + $DISKS_WAIT_DELAY))
```
jstodola, line 445: Is it guaranteed that decimal point will always be part of `/proc/uptime`? I wonder if there are platforms providing uptime as a whole number only.

lnykryn:
you could also do something like
read uptime _ < /proc/uptime
uptime=${uptime%.*}

dracut/anaconda-lib.sh
rmetrich: Actually all this is quite complicated, we may rely on `main_loop` instead

dracut/anaconda-lib.sh
```diff
@@ -434,3 +434,14 @@ wait_for_updates() {
...
 }
+
+wait_for_disks() {
+    # Allow up to 'inst.wait_for_disks' seconds for disks to be enumerated and
+    # related udev rules to execute (defaults to 5 seconds, 0 disables the
+    # feature)
+    finished_hook="$hookdir/initqueue/finished/wait_for_disks.sh"
+    [ -e "$finished_hook" ] && return
+    DISKS_WAIT_DELAY=$(getargnum 5 0 10000 inst.wait_for_disks)
+    DISKS_WAIT_RETRIES=$(($DISKS_WAIT_DELAY * 2))
```
rmetrich, line 445: Any idea what is wrong here?
~~~
dracut/anaconda-lib.sh:445:27: note: $/${} is unnecessary on arithmetic variables. [SC2004]
dracut/anaconda-lib.sh:446:54: note: Double quote to prevent globbing and word splitting. [SC2086]
~~~

jkonecny12:
Hi, I played with this today and it doesn't look to work correctly. I understand that the boot time would varied for each boot but it seems to me that the time we are presenting by the option is almost twice as much as we are telling user it is.

My tests results (the same VM):
```
I'm testing this by looking into the journal and taking time from 1st line and then look for "Starting Anaconda..." string for the second time. In brackets you have how many seconds it took to boot to Anaconda.

default:
12:42:42 -- 12:43:15 (33)
12:56:22 -- 12:56:53 (33)

timeout 0:
12:50:51 -- 12:51:15 (24)

timeout 2:
12:54:05 -- 12:54:31 (26)

timeout 4:
12:59:17 -- 12:59:45 (28)

timeout 100:
13:05:03 -- 13:07:50 (167)

150:
13:14:47 -- 13:19:36 (289)
15:00:20 -- 15:05:19 (299)
```

I don't think we need to be precise but in this case it seems to be almost twice as out which is too much to tell users these are seconds.

rmetrich:
The delay is inside initqueue loop, not the rest, so waiting for disks to
appear means up to X seconds from initqueue execution. Exactly similar to
the rd.retry value.

On Wednesday, March 29, 2023, Jiri Konecny ***@***.***> wrote:
> Hi, I played with this today and it doesn't look to work correctly. I
understand that the boot time would varied for each boot but it seems to me
that the time we are presenting by the option is almost twice as much as we
are telling user it is.
>
> My tests results (the same VM):
>
> I'm testing this by looking into the journal and taking time from 1st
line and then look for "Starting Anaconda..." string for the second time.
In brackets you have how many seconds it took to boot to Anaconda.
>
> default:
> 12:42:42 -- 12:43:15 (33)
> 12:56:22 -- 12:56:53 (33)
>
> timeout 0:
> 12:50:51 -- 12:51:15 (24)
>
> timeout 2:
> 12:54:05 -- 12:54:31 (26)
>
> timeout 4:
> 12:59:17 -- 12:59:45 (28)
>
> timeout 100:
> 13:05:03 -- 13:07:50 (167)
>
> 150:
> 13:14:47 -- 13:19:36 (289)
> 15:00:20 -- 15:05:19 (299)
>
> I don't think we need to be precise but in this case it seems to be
almost twice as out which is too much to tell users these are seconds.
>
> —
> Reply to this email directly, view it on GitHub, or unsubscribe.
> You are receiving this because you authored the thread.<
https://ci3.googleusercontent.com/proxy/OMYvWkaxc0bqDaicJKD5lm86vdea08AYaDqEhuwuK1DT3Z4w2sWB2ss_FXOPzC3oVz45cqtCBh1ycxVWAXltR0jMkogzDjWcbR8sH7W5JilJlLg1gGQ1ZGJlOYI3XhrF_-wggCjtbj14uquAQLg5Qe0nP39h8DPKu43jJTYF1BpJXHqwqvnT1QKOPOVDWCo3xD5fXslGA4XfzD51cVbKC3HnLqkiPdybE1aBHJOkq3AiMJ6mxw=s0-d-e1-ft#https://github.com/notifications/beacon/AAI4C42VWKMO6O4EI5DFYB3W6RH7ZA5CNFSM6AAAAAAVNCRNTWWGG33NNVSW45C7OR4XAZNMJFZXG5LFINXW23LFNZ2KUY3PNVWWK3TUL5UWJTSYXXRZG.gif>Message
ID: ***@***.***>

-- 
Renaud
Sent from my phone, sorry for the typos...

pvalena:
LGTM!

rmetrich:
> I'm not well informed about Anaconda development, but I wonder, why isn't the wait logic reversed? If somebody is using an "OEMDRV disk", they do it intentionally, which means they can also add a kernel cmdline option, e.g. `inst.oemdrv=1`. You can then wait for for a long time (e.g. 30 seconds) or indefinitely until the OEMDRV disk appears. With this approach it doesn't affect all the other users, their installer boot is not artificially prolonged.

This canot work, because the use of OEMDRV is to just load an additional disk and automagically the kickstart will be processed. Other methods request editing the command line, which is not possible for unattented installation.

PR #4590, opened 2023-03-03
Fix translations of the pre-release warning dialog (#2165762)

poncovka:
Move the pre-release warning dialog to a new class. Thanks to that the UI elements of the dialog will be created and translated when we need to show the dialog. The locale is already set to the right value at that point, so we don't have to retranslate the text manually.

Resolves: rhbz#2165762
**Ported from:** https://github.com/rhinstaller/anaconda/pull/4589


jkonecny12 (approved):
LGTM!

M4rtinK (approved):
Looks good to me as well. :)

VladimirSlavik (approved):
Looks good to me. Thank you!

PR #4591, opened 2023-03-03
Don't generate a repository name of the URL source

poncovka:
The generated name is never user, because we always rename it to "anaconda" during the payload setup. That is fine, because the source without a repository name is a base repository and there can be only one base repository anyway.

We cannot really use a generated name without further modifications of the code, because the validation and verification checks expect the "anaconda" name at this moment. If we ever do need generated names, we can generate them during the payload setup anyway.


VladimirSlavik (approved):
Looks good to me. Thank you!

PR #4592, opened 2023-03-03
Remove the tx_id functionality

poncovka:
Replace `tx_id` with a simple boolean property. Use `software_validation_required` to check if the validation of the software selection is required. The validation is required after every payload setup due to possible changes in configured repositories and their packages, groups and environments.


VladimirSlavik (approved):
Looks good to me. Thank you!

PR #4593, opened 2023-03-06
Clean up tasks for setting up and tearing down sources

poncovka:
Make the set up and tear down tasks easier to extend in the future.
Change some of the task names, doc strings, logs and exception messages.


VladimirSlavik (approved):
Thanks!

PR #4519, opened 2023-01-23
Add DNS search and ignore options from kickstart

VladimirSlavik:
This is the anaconda part.

Blocked by waiting on https://github.com/pykickstart/pykickstart/pull/431, ~~which in turn is blocked by [rhbz#1656662](https://bugzilla.redhat.com/show_bug.cgi?id=1656662#c9)~~.

TODOs:
- [x] pykickstart PR opened: https://github.com/pykickstart/pykickstart/pull/431
- [x] pykickstart PR merged (tests will work)
- [x] pykickstart released (can merge)
- [x] release notes
- [x] testing
- [x] add tests

PR #4594, opened 2023-03-06
Fix saving DNS search domains to kickstart

VladimirSlavik:
Resolves: [rhbz#2173219](https://bugzilla.redhat.com/show_bug.cgi?id=2173219)


VladimirSlavik:
If you're fine with that, then it goes into the next Tuesday build. (Won't make it into tomorrow's one due to missing approvals.)

rvykydal (approved):
Looks good to me.
It would be nice to have a kickstart test covering the option as I suggested in https://github.com/rhinstaller/anaconda/pull/4519#issuecomment-1422231431. Then we would just run it here in the PR.

PR #4597, opened 2023-03-07
Modernize javascript (fedora-welcome)

fmuellner:
Both the javascript standards and support in gjs/gnome have changed quite a bit since the code was originally written. Update the code to match the current best practices (without changing any of the logic or UI).

### DIFF
--- a/data/liveinst/gnome/fedora-welcome.js
+++ b/data/liveinst/gnome/fedora-welcome.js
@@ -37,9 +37,9 @@ let anacondaApp = null;
 function makeLabel(label, button) {
     let widget = new Gtk.Label();
 
-    if (button)
+    if (button) {
         widget.set_markup('<b><span size="x-large">' + label + '</span></b>');
-    else {
+    } else {
         widget.set_line_wrap(true);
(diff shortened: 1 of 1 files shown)


VladimirSlavik (approved):
Looks good to me, thank you!

VladimirSlavik:
@fmuellner did you test these changes? If not, how would one do that?

fmuellner:
> @fmuellner did you test these changes? If not, how would one do that?

Yes. I copied liveinst.desktop, changed the Exec line to `/usr/bin/true` and moved the file to `~/.local/share/applications`. Then I created this fake liveinst command in `~/.local/bin`:
```
#!/bin/sh
notify-send -h boolean:transient:true "Pretend I'm a snake"
```

After that, `fedora-welcome.js` can be run normally on the host.

M4rtinK (approved):

VladimirSlavik:
Ah, ok, thanks. I will try this also on the live media. I suspect that's just a formality, but you never know...

jkonecny12 (approved):
Thanks for your help!

PR #4600, opened 2023-03-08
Fix broken GH workflows which are using ISO builds

jkonecny12:
After this change https://github.com/rhinstaller/anaconda/pull/4566 I broke the builds.

There are two issues:
- I accidentally removed directory creation together with the mknods from the workflow.
- The `sudo` calls are not necessary (could be hurtful) because we might be missing sudo app in the container (needs to be backported to other branches)


VladimirSlavik (approved):
Thanks!

PR #4606, opened 2023-03-08
Remove sudo for mknod calls in the container (#infra)

jkonecny12:
We don't need to run sudo there. This could be even harmful if the sudo is not available in the container.

Backport of https://github.com/rhinstaller/anaconda/pull/4600.


poncovka (approved):
Thanks!

PR #4607, opened 2023-03-08
infra: Disable failing tests that call useradd and groupadd

poncovka:
These tests run in a very limited chroot environment that is probably missing some important files at this moment. Due the that the useradd and groupadd calls fail with the exit status 1 without any additional info. Let's disable them for now. We might consider replacing these unit tests with kickstart tests instead of keeping them alive.


poncovka:
Closing. The tests will be fixed with https://github.com/rhinstaller/anaconda/pull/4608.

poncovka:
Reopened. We will use it as a temporary workaround.

PR #4608, opened 2023-03-08
Stop creating files to support user creation

VladimirSlavik:
Formerly, we created `/etc/login.defs` to ensure users and groups can be created. That is not needed anymore, in fact it broke tests.

Additionally, we copied actual binary libraries to the mock root when running unit tests. That is also not needed anymore.

---

Draft only, because the "fix" is based on debugging our code, the actual reasons why things are different are still unknown.

### DIFF
--- a/pyanaconda/core/users.py
+++ b/pyanaconda/core/users.py
@@ -21,7 +21,6 @@
 import os
 import os.path
 import subprocess
-from contextlib import contextmanager
 from pathlib import Path
 
 from pyanaconda.core import util
@@ -245,30 +244,6 @@ def _getgrgid(gid, root):
     return None
 
 
-@contextmanager
-def _ensure_login_defs(root):
-    """Runs a command after creating /etc/login.defs, if necessary.
-
-    The groupadd and useradd utilities need login.defs to exist in the chroot,
-    and if someone is doing a cloud image install or some kind of --nocore thing
-    it may not. An empty one is ok, though. If it's missing, create it,
-    run the command, then clean it up.
-
-    :param str root: filesystem root for the operation
-    """
-    login_defs_path = root + '/etc/login.defs'
-    if not os.path.exists(login_defs_path):
-        open(login_defs_path, "w").close()
-        login_defs_created = True
-    else:
-        login_defs_created = False
-
-    yield
-
-    if login_defs_created:
-        os.unlink(login_defs_path)
-
-
 def create_group(group_name, gid=None, root=None):
     """Create a new user on the system with the given name.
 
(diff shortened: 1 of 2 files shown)


poncovka (approved):
Thank you!

jkonecny12 (approved):

VladimirSlavik:
So, this showed up with the Mar 8 rebuild. Looking at shadow-utils [changes in fedora](https://packages.fedoraproject.org/pkgs/shadow-utils/shadow-utils/fedora-rawhide.html), two new versions landed recently, 4.13-5 on Mar 2 and 4.13-6 on Mar 6. Perhaps that could be related? I can't get the older versions with dnf, though.

---

@ikerexxe Do you have any ideas?

What we do is that we run `useradd` and `groupadd` with the chroot option `-R`, and basically only make sure files in `/etc/` and `/etc/login.defs` are present before the tool is executed.

This stopped working recently - I can't pinpoint the exact version, but it is broken now with `shadow-utils-4.13-6.fc39.x86_64`. The tools do not create the users and groups, return error 1, and produce no output. If I stop adding the (empty) login.defs, the calls succeed.

Since the comments in code say login.defs must exist, I don't think that this is the right way to resolve this.

poncovka:
Please, don't forget to revert https://github.com/rhinstaller/anaconda/commit/e4ccf86690059ceee6316802f0fde65d49424b3a.

ikerexxe:
I'm kind of surprised that this is failing on rawhide but not on f38. The only difference between the two versions is that I added [two build dependencies that were missing](https://src.fedoraproject.org/rpms/shadow-utils/c/fd05b1cb1a183db7af622b34c9cf7c3994808475?branch=rawhide). Can it be because something else changed?

By the way, what do you do with `login.defs`?

VladimirSlavik:
> By the way, what do you do with login.defs?

Make sure that it exists (even empty) when adding users and groups. However, it's hard to tell if that's still correct. The code is from 2015, from a migration off libuser.

> Can it be because something else changed?

That's actually possible. However now that I think about it, perhaps empty `login.defs` is no longer enough.

VladimirSlavik:
So, after playing with it a bit more, I _think_ we should stop creating empty `/etc/login.defs` in actual code. Its absence does not seem to have any effect in the testing container. IMO that means empty == missing, even if that was not the case when that code was written. Further, the file comes from shadow-utils package, same as the binary tools, so it's not likely it would be missing, and if it is, that would be intentional.

For the unit tests, approximating a "normal" system by copying the file to chroot seems like a safe choice, even if that is redundant.

VladimirSlavik:
All green, so merging. Friday evening. What could go wrong?

PR #4609, opened 2023-03-09
infra: Set branched fedora for packit correctly

VladimirSlavik:
The target "fedora-latest" does not get branched fedora before beta (as of writing this), and the actual number is available for templates, so let's be explicit and get the right fedora out of the box.


poncovka (approved):
Thanks!

PR #4610, opened 2023-03-09
Modernize UI (fedora-welcome)

fmuellner:
Port to GTK4, then libadwaita, and adapt the UI to follow most recent design patterns.

Implemented on top of #4597.


fmuellner:
Obligatory screenshot:

Screenshot from 2023-03-10 00-51-28

poncovka:
Thank you! I love the new look :-)

jkonecny12:
This looks great! Do we want to get this to Rawhide only or also to F38?

kparal:
I'm not a native speaker, so perhaps I'm totally off, but the word "trial" feels weird to me here. I'm used from various proprietary services that "Continue Trial" is usually the button you want to click, because it means you get the same or similar service, but temporarily free/discounted. Unaware users might get the idea that the Trial version is the non-paid version or something. While it is just a Live environment (unfortunately this is also not an easy-to-explain concept for a newcomer). Perhaps we can find a better word/phrase than Trial? Something very basic like "_Look around first_"?

allanday:
>  the word "trial" feels weird to me here. I'm used from various proprietary services that "Continue Trial" is usually the button you want to click, because it means you get the same or similar service, but temporarily free/discounted.

That is a common use of the term, but it's not [the actual meaning of the word](https://www.merriam-webster.com/dictionary/trial). But if the "free trial" connotations are too strong, we could perhaps go with `Install Fedora…` and `Not Now` as the button labels.

If we were to do that, we might want to update the other text, which also uses the term `trial`:

`This live media can be used to install Fedora or as a temporary system. Installation can be started at any time using the install icon in Activities.`

fmuellner:
Done.

M4rtinK:
I've created a scratch build in COPR: https://copr.fedorainfracloud.org/coprs/m4rtink/anaconda_new_spash/

VladimirSlavik:
@fmuellner, what is the relationship between this and #4597? Is it sufficient to merge only this?

(For completeness sake, I'll add that we'd also need to port this to f38, but that should be just an exercise in cherry-picking.)

fmuellner:
> what is the relationship between this and #4597? Is it sufficient to merge only this?

My idea was that #4597 would land first and then this, but you are right, it would be possible/sufficient to only land this PR.

VladimirSlavik:
Tested manually on livecd - thanks @M4rtinK!

VladimirSlavik (approved):
Thank you for the contribution! Looks good to me.

VladimirSlavik:
@fmuellner I will port this to the `fedora-38` branch this week; if you want to do so faster, feel free to do so. All it means is make a PR with cherry-picked commits against that branch.

fmuellner:
You make it sound like the renaming was done willy-nilly just because I fancy the name (or even: want to screw over others).

That's not the case. GTK no longer has a way to sidestep the usual application matching and hint to an arbitrarily named .desktop file, so this is how it looks like with the old name:

image

We had to patch mutter because it looked too unprofessional opening the window off-center, so I don't see how *that* would have flown.

AdamWill:
I don't think I make it sound like that? I just said that when you rename a .desktop file, it seems like a good idea to figure out what actually runs it, and make sure that is also updated.

VladimirSlavik:
Not to dispute the point, but - I actually wonder how one would go about doing that, from a "cold start", as I apparently have to...

AdamWill:
I guess if I had to, I'd boot a live image and grep the filesystem for mentions of the old name. I mean, we know what the thing is *for*, right? It pops up on boot of a live image. Ergo there must be something on a live image that makes it happen.

Once you found that, you'd find the file was part of livesys-scripts...

PR #4614, opened 2023-03-14
Set correctly NM props for DNS kickstart options

VladimirSlavik:
I will gather here the remaining fixes for #4519


poncovka (approved):
Thanks!

rvykydal (approved):

VladimirSlavik:
A still open bug is that the changes are applied only to the "hot" devices, but not to configuration files.

No idea where to start yet, just making sure this bit is not lost.

PR #4620, opened 2023-03-17
Generate the ostreesetup kickstart command

poncovka:
Pykickstart requires to set the `seen` attribute of this kickstart command
since 3.46 to make it visible in the output kickstart file.

See: https://github.com/pykickstart/pykickstart/commit/8d85411


jkonecny12:
This will be broken even on F38.

jkonecny12 (approved):
Thanks for the fix.

VladimirSlavik (approved):
Thanks!

PR #4622, opened 2023-03-17
Tweak how we run cockpit tests locally

jkonecny12:
We do unnecessary steps when running the Cockpit tests locally and also don't have nice support for OSTree based systems. Let's improve that with this PR.

### DIFF
--- a/ui/webui/test/README.rst
+++ b/ui/webui/test/README.rst
@@ -16,14 +16,14 @@ To run the WebUI integration tests run the following from the root of the anacon
 OSTree based systems (SilverBlue etc.) can use toolbx.
 See `<../../../CONTRIBUTING.rst#how-to-run-make-commands>`_.
 
-Prepare an updates.img containing the anaconda RPMs and the cockpit dependencies::
-
-    cd ui/webui
-    make create-updates.img
-
 Then download test dependencies::
 
+    cd ui/webui
     make prepare-test-deps
+
+Prepare an updates.img containing the anaconda RPMs and the cockpit dependencies::
+
+    make create-updates.img
 
 Then download the ISO file that the test VMs will use::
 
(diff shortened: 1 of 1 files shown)


jkonecny12:
lol POSIX issue, I now understand why it's not used there :(

VladimirSlavik:
Just change the shebang to bash? Unless there's a reason to keep this as a sh script...

VladimirSlavik (approved):
Thank you very much, this is definitely a step in the right direction.

VladimirSlavik (commented):
I don't think the strict mode is necessary as optional. It should be a default instead, I think. It will find rpms that cannot be unpacked, but it will NOT detect if a rpm specified by a glob is missing.

If the problem is that `curl` in `prepare-updates-img`grabs an error webpage instead of the rpms, then that's the place to fix it.

FWIW, a real fix for a "rpm does not exist" within `makeupdates` would be changing the warning for no glob results into a failure. I'm not sure what that would break, but probably primarily manual uses. However, I see no need to even take such decision - the palce to fix this is line 16 and later in `prepare-updates-img`.

VladimirSlavik:
Here's a stub of the solution...
```diff
  for url in $COCKPIT_URLS; do
-    test -e "$(basename "${url}")" || curl -LO "$url"
+    test -e "$(basename "${url}")" || curl -fLO "$url"
  done
```
This should fail if the return status is 404 and such.

jkonecny12:
This new version is blocked by https://github.com/cockpit-project/bots/pull/4560 so far.

I'm planning to use our config rendering to select the images correctly after branching. However, I would like to first know your opinion @VladimirSlavik about this to avoid doing unnecessary stuff.

VladimirSlavik:
> We can add --fail which would kill this. However, I'm thinking about not using curl but rather call dnf download in the Anaconda build machine. We already building the rpm files so it's not that hard.

Yes please.

ui/webui/test/README.rst
```diff
@@ -129,11 +126,59 @@ Make sure to set::
...
+For updating (pushing) updated pixel test reference images you can use the available the Makefile target::
+
+    make update-reference-images
+
+How to fix failed pixel tests
+-----------------------------
+
+For all the steps below you have to be in `ui/webui` directory of the project.
+
+Locally just copy the broken tests images to the `test/refercence` directory. However, easier
```
VladimirSlavik, line 138: reference - remove the additional C

scripts/makeupdates
```diff
@@ -409,7 +411,11 @@ def add_rpms(updates_path, rpms):
     for rpm in resolved_rpms:
         cmd = "cd %s && rpm2cpio %s | cpio -dium" % (updates_path, rpm)
         sys.stdout.write(cmd+"\n")
-        os.system(cmd)
+        ret = os.system(cmd)
+        print(f"ret code: {ret}")
+
+        if STRICT_ENABLED and ret != 0:
+            raise RuntimeError(f"RPM {rpm} can't be added to the updates image!")
```
VladimirSlavik, line 418: I still think it is wrong to add this and then hide behind a flag. If you want to unpack a rpm and it does not work, that's a problem. If that problem is silently skipped, that is a bug.

ui/webui/test/README.rst
```diff
@@ -129,11 +126,59 @@ Make sure to set::
...
+If everything went well your PR should be green now.
+
+Outdated Cockpit CI image for testing
+-------------------------------------
+
+From time to time you can face an issue that the fedora-X-boot image on Cockpit side is
+missig dependency for your PR. **You should not push your PR without fixing the image first!**
+
+To update the image please ping #cockpit on IRC and they will provide a PR with the new image.
+It will look similar to ["this"](https://github.com/cockpit-project/bots/pull/4551).
```
VladimirSlavik, line 172: Does this mean that cockpit team will build a special image with the required dependencies, based on the anaconda PR? Because just grabbing newer images won't work.

jkonecny12: It should work as discussed today. Our tests depends on having the dependencies in official repositories the same as Cockpit does. At least AFAIK.

scripts/makeupdates
jkonecny12: Changed.

ui/webui/test/README.rst
jkonecny12: Fixed.

jkonecny12:
All green as I like it ;).

ui/webui/test/README.rst
VladimirSlavik: Repository != image.

PR #4628, opened 2023-03-20
WebUI: Dynamically choose Quit/Reboot button label

OndrejZobal:
If Installation_System.type = "Boot_ISO", "Reboot" will be displayed instead of "Quit" on the quit button and the quit confirmation modal.

### DIFF
--- a/ui/webui/src/components/AnacondaWizard.jsx
+++ b/ui/webui/src/components/AnacondaWizard.jsx
@@ -40,7 +40,7 @@ import { usePageLocation } from "hooks";
 
 const _ = cockpit.gettext;
 
-export const AnacondaWizard = ({ onAddErrorNotification, toggleContextHelp, title }) => {
+export const AnacondaWizard = ({ onAddErrorNotification, toggleContextHelp, title, conf }) => {
     const [isFormValid, setIsFormValid] = useState(true);
     const [stepNotification, setStepNotification] = useState();
     const [isInProgress, setIsInProgress] = useState(false);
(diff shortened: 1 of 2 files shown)


ui/webui/src/components/AnacondaWizard.jsx
```diff
@@ -104,6 +104,7 @@ export const AnacondaWizard = ({ onAddErrorNotification, toggleContextHelp, titl
             setStepNotification={setStepNotification}
             isInProgress={isInProgress}
             setIsInProgress={setIsInProgress}
+            isQuitReboot={conf["Installation System"].type == "BOOT_ISO"}
```
VladimirSlavik, line 107: Is this something that can be depended on? I'd imagine this is something we want a D-Bus call, not a conf prop...

So maybe add a TODO?

OndrejZobal: In an internal issue number `INSTALLER-3394` @M4rtinK suggested using the config directly and config was already used for something else.

M4rtinK: I think that's the idea about the config file - it's always loaded & the values should not change, so we can avoid making a DBus API for everything.

M4rtinK (approved):
Looks good to me, thanks! :)

VladimirSlavik (approved):

M4rtinK:
Is something blocking this PR or can it be merged now ? :)

PR #4617, opened 2023-03-16
F38 enable ostree containers

jkonecny12:
Backport of https://github.com/rhinstaller/anaconda/pull/4561

PR #4630, opened 2023-03-20
Update ks imports and tests for new commands in f38

VladimirSlavik:
SSIA. Most things are broken while we wait for #4617, so this is an alternative if that gets stuck.


rvykydal (approved):

VladimirSlavik:
No longer needed :)

PR #4631, opened 2023-03-21
infra: Notify to Slack about new releases

VladimirSlavik:
Let's notify to Slack, since that's where we should be.

Notice how the code is pretty much the same as the gchat notifications. Interoperability!

I'm not removing the gchat part yet, so that we have a plan B for the first run(s). However, hard-earned lessons about quoting in shell in yaml were applied right away this time.


poncovka (approved):
Thank you!

PR #4632, opened 2023-03-21
Simplify the NFS source

poncovka:
Use a new support for sources that provide access to a repository.
Update the code for the NFS source in the DNF payload, GUI and TUI.
Use the source as an internal representation of additional repositories.


VladimirSlavik (approved):
Looks good to me. Thank you!

PR #4634, opened 2023-03-22
Update pixel tests for a new cockpit-ws

jkonecny12:
Cockpit changes fonts some time ago and we are now getting a new cockpit packages after a log time so we need to fix this.


jkonecny12:
Cockpit tests are expected to fail because of missing image which is red because of this.
See https://github.com/cockpit-project/bots/pull/4551

KKoukiou:
land when fedora-rawhide-boot@bots#4551 is green in lockstep with 4551 bots pr

PR #4635, opened 2023-03-22
Don't protect unavailable devices

poncovka:
The zram deviced are excluded from the device tree by the Blivet's `ignored_device_names` filter, so it is not necessary to mark them as protected. There will be nothing to protect.

See the `enable_installer_mode` function from the Storage module:
https://github.com/rhinstaller/anaconda/blob/0a97b816536ed6af61c593130310e53f90c25ce4/pyanaconda/modules/storage/initialization.py#L72

### DIFF
--- a/pyanaconda/anaconda.py
+++ b/pyanaconda/anaconda.py
@@ -19,7 +19,6 @@
 
 import os
 import sys
-from glob import glob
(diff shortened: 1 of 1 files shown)


VladimirSlavik (approved):
Thanks!

PR #4637, opened 2023-03-23
Remove the SourceFactory class

poncovka:
This code is no longer used by the installer.

**Removal of the SourceFactory class - final part**


M4rtinK (approved):
Looks good to me, thanks!

VladimirSlavik (approved):
Thanks!

PR #4640, opened 2023-03-23
infra: Enable the workflow that prepares and tags a release commit

poncovka:
Automated releases should be disabled for branched Fedoras, but it should be still possible to manually trigger the workflow that creates a release commit via GitHub.


.github/workflows/release-periodically.yml.j2
```diff
@@ -1,4 +1,3 @@
-{% if distro_release == "rawhide" %}
 name: Release periodically
 # Create a GitHub release on schedule
```
VladimirSlavik, line 2: Maybe please add a comment that it's kept on the fedora branches with only manual trigger to make things easier. The comment is no longer the whole truth :(

VladimirSlavik (approved):
Looks good to me. Thank you!

VladimirSlavik (approved):
Looks good to me. Thank you!

jkonecny12:
I restarted the cockpit-ci test, however, if they went red again please rebase this PR to fix them.

PR #4641, opened 2023-03-23
Fix Cockpit CI tests on Fedora-38 (#infra)

jkonecny12:
This solution will solve our issue now but after Fedora 38 is released than the links will break. However, we don't care much about the released stuff so I'm taking it as good enough.


VladimirSlavik (approved):
Thanks!

jkonecny12:
Tested with https://github.com/cockpit-project/bots/pull/4555

PR #4642, opened 2023-03-23
Simplify creation of a source based on its URL

poncovka:
Call the `_create_source_from_url` method of the DNF payload object
to create a new source for the URL specified via boot options.

**Removal of the SourceFactory class - part 4**


VladimirSlavik (approved):
Looks good to me. Thank you!

PR #4643, opened 2023-03-23
move threading to core

VladimirSlavik:
SSIA, mostly.

Related changes in other places will be linked here once the PRs are opened.


anaconda.py
```diff
@@ -426,8 +426,8 @@ def _earlyExceptionHandler(ty, value, traceback):
         wait_for_connected_NM(timeout=opts.waitfornet)
 
     # In any case do some actions only after NM finishes its connecting.
-    threadMgr.add(AnacondaThread(name=constants.THREAD_WAIT_FOR_CONNECTING_NM,
-                                 target=wait_for_connecting_NM_thread))
+    thread_manager.add(AnacondaThread(name=constants.THREAD_WAIT_FOR_CONNECTING_NM,
```
poncovka, line 429: If you want to improve the API, I would consider adding support for `thread_manager.add_thread()`.

Instead of this call:

```
thread_manager.add(AnacondaThread(
   name=constants.THREAD_WAIT_FOR_CONNECTING_NM,
   target=wait_for_connecting_NM_thread
))
```

You could call something like this:

```
thread_manager.add_thread(
   name=constants.THREAD_WAIT_FOR_CONNECTING_NM,
   target=wait_for_connecting_NM_thread
)
```

anaconda.py
```diff
@@ -490,8 +490,8 @@ def _earlyExceptionHandler(ty, value, traceback):
     if not conf.target.is_directory:
         from pyanaconda.ui.lib.storage import reset_storage
 
-        threadMgr.add(AnacondaThread(name=constants.THREAD_STORAGE,
-                                     target=reset_storage))
+        thread_manager.add(AnacondaThread(name=constants.THREAD_STORAGE,
+                                          target=reset_storage))
```
poncovka, line 494: The indentation of these calls drives me crazy, but that's up to you if you want to fix it.

VladimirSlavik: Ah, yes. Actually I did that on the oscap code when I changed it later, just didn't realize this was here too.

VladimirSlavik:
I'm unsure what are the right expectations about the threading code, so the tests are so-so. Maybe think of them as documentation of what happens, instead?

Most notably, I can't seem to trigger any way to get `thread_manager.any_errors` other than zero, even if I raise an exception, catch it with a custom exception handler, and query the property in there.

----

I really dislike that `AnacondaThread` hard-codes sending statuses to `thread_manager` as instantiated in the module, which is documented to be a singleton but really isn't, you just have to treat it as one. The tests work around this by patching, but it's a quite ugly situation. Perhaps a way out would be to make `AnacondaThread` an internal detail, and rely on the fact that then the only way to add threads would be `ThreadManager.add_thread`, which would pass the manager as another `kwarg`. However, that's a more serious change, and I hesitate to turn this into a full-on refactoring...

VladimirSlavik:
/kickstart-test --testtype coverage --skip-testtypes knownfailure,manual,skip-on-fedora

VladimirSlavik:
The following tests timed out: rootpw-basic, rootpw-lock, services, onboot-activate. raid-1 also failed, but that is disabled right now.

PR #4644, opened 2023-03-24
infra: Update release workflows

poncovka:
Automated releases should be disabled for branched Fedoras, but it should
be still possible to manually trigger the workflow that creates a release
commit via GitHub.

Rename workflows that handle automated releases and update the documentation.

**Ported from:** https://github.com/rhinstaller/anaconda/pull/4640


VladimirSlavik (commented):
Thanks!

VladimirSlavik (approved):

PR #4645, opened 2023-03-24
Set timezone from user choice if applicable

VladimirSlavik:
This attempts to set timezone with the user choice on welcome screen, but only when there is no better choice.

----

Note to self: If/when porting to RHEL 9, this will drag in the geolocation modularization, and probably also the welcome spoke translation fixes.

### DIFF
--- a/pyanaconda/modules/timezone/timezone_interface.py
+++ b/pyanaconda/modules/timezone/timezone_interface.py
@@ -47,17 +47,37 @@ class TimezoneInterface(KickstartModuleInterface):
     @Timezone.setter
     @emits_properties_changed
     def Timezone(self, timezone: Str):
-        """Set the timezone.
+        """Set the timezone with maximal priority.
 
-        Sets the system time zone to timezone. To view a list of
-        available time zones, use the timedatectl list-timezones
-        command.
-
-        Example: Europe/Prague
+        See SetTimezoneWithPriority for more details.
 
         :param timezone: a string with a timezone
         """
         self.implementation.set_timezone(timezone)
+
+    @emits_properties_changed
+    def SetTimezoneWithPriority(self, timezone: Str, priority: UInt16):
+        """Set the timezone with a given priority.
+
+        Sets the system time zone to timezone, if the already stored timezone does not have higher
+        priority.
+
+        To view a list of available time zones, use the `timedatectl list-timezones` command.
+
+        Example: Europe/Prague
+
+        The priority is a positive number. Use values defined in pyanaconda.core.constants
+        as TIMEZONE_PRIORITY_* :
+            TIMEZONE_PRIORITY_DEFAULT = 0
+            TIMEZONE_PRIORITY_LANGUAGE = 30
+            TIMEZONE_PRIORITY_GEOLOCATION = 50
+            TIMEZONE_PRIORITY_KICKSTART = 70
+            TIMEZONE_PRIORITY_USER = 90
+
+        :param timezone: a string with a timezone specification in the Olson db aka tzdata format
+        :param priority: priority for the timezone; see the respective constants
+        """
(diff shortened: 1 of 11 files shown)


jkonecny12 (commented):
Looks good to me. The list of explicit priorities is something I welcome.

pyanaconda/ui/gui/spokes/datetime_spoke.py
```diff
@@ -311,7 +311,10 @@ def _initialize(self):
             log.warning("%s is not a valid timezone, falling back to default (%s)",
                         kickstart_timezone, DEFAULT_TZ)
             self._set_timezone(DEFAULT_TZ)
-            self._timezone_module.Timezone = DEFAULT_TZ
+            self._timezone_module.SetTimezoneWithPriority(
+                DEFAULT_TZ,
+                constants.TIMEZONE_PRIORITY_KICKSTART
```
poncovka, line 316: This should be handled in the module. I think you can actually remove this whole section with a kickstart timezone.

pyanaconda/ui/gui/spokes/datetime_spoke.py
```diff
@@ -352,7 +355,10 @@ def apply(self):
         if not region or not city:
             return
 
-        self._timezone_module.Timezone = region + "/" + city
```
poncovka, line 355: You don't have to change this line, the property uses the highest priority by default.

pyanaconda/ui/gui/spokes/welcome.py
```diff
@@ -260,6 +260,7 @@ def on_locale_selection_changed(self, selection):
             lang = store[selected[0]][1]
             lang = localization.setup_locale(lang)
             self._set_lang(lang)
+            self._try_set_timezone(lang)
```
poncovka, line 263: I recommend to do this in the `apply` function after the user confirms the language selection. This could generate a huge amount of log message.

pyanaconda/ui/tui/spokes/time_spoke.py
```diff
@@ -303,7 +303,10 @@ def prompt(self, args=None):
         return prompt
 
     def apply(self):
-        self._timezone_module.Timezone = self._selection
```
poncovka, line 306: You don't have to change this.

poncovka (approved):
I have some recommendations. Otherwise, it looks good to me. Thank you!

pyanaconda/modules/timezone/timezone_interface.py
```diff
@@ -47,18 +47,30 @@ def Timezone(self) -> Str:
...
-
-        Example: Europe/Prague
+        See SetTimezoneWithPriority for more details.
 
         :param timezone: a string with a timezone
         """
         self.implementation.set_timezone(timezone)
 
+    @emits_properties_changed
+    def SetTimezoneWithPriority(self, timezone: Str, priority: Int16):
```
poncovka, line 59: Maybe it could be an unsigned integer.

pyanaconda/ui/gui/spokes/datetime_spoke.py
VladimirSlavik: I think using the same function everywhere is preferable.

pyanaconda/ui/gui/spokes/datetime_spoke.py
VladimirSlavik: I'm not really sure about that. This is partial kickstart support. The module does not verify that the timezone we're setting is valid. It just passes it through without validation. That's probably ok for all-automatic installs. However the GUI might need to have some sane value when it opens. I will make this a separate commit, test and see what it does...

poncovka (approved):
It looks great. Thank you!

PR #4647, opened 2023-03-27
infra: Reorder unit tests

VladimirSlavik:
With this, the fast few-second tests come first and the slowest stuff comes last. The actual heavy tests that take 99% of time stay in the same order.

The improvement is that running tests locally now produces some output soon after the initial make bootstrapping, so it's clear that things are not stuck.


poncovka (approved):
Thank you!

PR #4649, opened 2023-03-27
Move ostreecontainer deps to install-env-deps

jkonecny12:
The dependencies should be in core package correctly but having it there will drag in these packages to Live environment and also all systems installed with Live ISO which still have Anaconda package.

Moving it to install-env-deps package will add the requirement only to boot.iso which is the main use for this.

Related: rhbz#2125655


VladimirSlavik (approved):
Thanks!

PR #4652, opened 2023-03-27
Add 'vga' to the list of preserved kernel arguments (#2176782)

AdamWill:
To address problems with KDE on Wayland on BIOS boots with "basic mode" graphics, we need to set a framebuffer mode in the bootloader (vga=791). This should also be passed through to the installed system so SDDM will work there as well.


Conan-Kudo (approved):

poncovka (approved):
Looks good to me. Thanks!

jkonecny12 (approved):

PR #4658, opened 2023-03-28
Add GH workflow for periodic webui integration tests in Permian

rvykydal:
There is a number of follow-up issues, but one of them should be handled to make the workflow useful:
- [x] Make the Permian anaconda webui workflow compatible with installer images produced by ImageBuilder (payload is not in the installer image) - or use the images compatible with the workflow.

### DIFF
--- a/.github/workflows/webui-periodic.yml
+++ b/.github/workflows/webui-periodic.yml
@@ -11,7 +11,7 @@ permissions:
 jobs:
   scenario:
     name: Webui periodic
-    runs-on: [self-hosted, kstest]
+    runs-on: [self-hosted, kstest-permian]
 
     timeout-minutes: 65
     env:
--- a/.github/workflows/webui-periodic.yml
+++ b/.github/workflows/webui-periodic.yml
@@ -1,0 +1,107 @@
+# Run webui e2e tests periodically in Permian
+name: webui-periodic
+on:
+  schedule:
+    - cron: 0 23 * * *
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+jobs:
+  scenario:
+    name: Webui periodic
+    runs-on: [self-hosted, kstest]
+
+    timeout-minutes: 65
+    env:
+      # ocp-master-xxl: 32GB RAM / 4GB RAM per VM
+      TEST_JOBS: 8
+      # The timeout should be a few minutes less then the job's timeout-minutes
+      # so that we get partial results and logs in case of the timeout.
+      LAUNCHER_TIMEOUT_MINUTES: 60
+      # URL to the unpacked installation image
+      # TODO: use preview image
+      STATIC_REPO_URL: "https://fedorapeople.org/groups/anaconda/webui_permian_tests/sources/periodic/x86_64/"
+
+    steps:
+      # self-hosted runners don't do this automatically; also useful to keep stuff around for debugging
+      # need to run sudo as the launch script and the container create root/other user owned files
+      - name: Clean up previous run
+        run: |
+          sudo podman ps -q --all --filter='ancestor=kstest-runner' | xargs -tr sudo podman rm -f
+          sudo podman volume rm --all || true
+          sudo rm -rf *
+
+      # The test library is in anaconda repository
+      - name: Clone anaconda repository
+        uses: actions/checkout@v3
+        with:
+          ref: master
+          fetch-depth: 0
+          path: anaconda
+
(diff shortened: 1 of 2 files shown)


.github/workflows/webui-periodic.yml
```diff
@@ -0,0 +1,111 @@
...
+        id: prepare_repo
+        run:
+          echo "repo_url=$STATIC_REPO_URL" >> $GITHUB_OUTPUT
+
+      - name: Create Permian settings file
+        working-directory: ./permian
+        run: |
+          cat <<EOF > settings.ini
+          [AnacondaWebUI]
+          anaconda_repo=https://github.com/rhinstaller/anaconda.git
```
velezd, line 70: This is default

.github/workflows/webui-periodic.yml
```diff
@@ -0,0 +1,111 @@
...
+          # TODO resolve using csbuild and pin to newer version
+          cockpit_branch=283
```
velezd, line 72: This is now also default, and when newer version can be used it should be changed in the default settings

.github/workflows/webui-periodic.yml
```diff
@@ -0,0 +1,111 @@
...
+          path: tplib
+
+      # Use static URL to unpacked image
+      # TODO: unpack the image as the part of the workflow
+      - name: Set the image to be tested
+        id: prepare_repo
+        run:
+          echo "repo_url=$STATIC_REPO_URL" >> $GITHUB_OUTPUT
+
+      - name: Create Permian settings file
```
velezd, line 65: Did you consider using -o options instead of creating settings file every time? If you are changing only few settings it may look cleaner.

velezd: Or, those settings are hard-coded anyway so maybe you should use static file stored in /anaconda/ui/webui/test/integration. Then you don't event have to specify it because permian looks for settings files in the library path.

.github/workflows/webui-periodic.yml
rvykydal: removed

.github/workflows/webui-periodic.yml
rvykydal: When creating GH workflow for kstests in Permian I used -o first, but then, based on a review I started to drop a file instead. It seems a bit cleaner / readable to me. And easier to modify.

rvykydal:
I've updated static source URL which has real content now.
Also for a start I'd run the workflow on a dedicated runner (kstest-permian).

.github/workflows/webui-periodic.yml
VladimirSlavik: So, where would I go to find the default?

.github/workflows/webui-periodic.yml
```diff
@@ -0,0 +1,107 @@
...
+      # TODO: use preview image
+      STATIC_REPO_URL: "https://fedorapeople.org/groups/anaconda/webui_permian_tests/sources/periodic/x86_64/"
+
+    steps:
+      # self-hosted runners don't do this automatically; also useful to keep stuff around for debugging
+      # need to run sudo as the launch script and the container create root/other user owned files
+      - name: Clean up previous run
+        run: |
+          sudo podman ps -q --all --filter='ancestor=kstest-runner' | xargs -tr sudo podman rm -f
+          sudo podman volume rm --all || true
```
VladimirSlavik, line 33: I think this would remove squid once deployed to regular runners?

.github/workflows/webui-periodic.yml
velezd: Defaults for this plugin can be found here [permian/libpermian/plugins/anaconda_webui/settings.ini](https://github.com/rhinstaller/permian/blob/devel/libpermian/plugins/anaconda_webui/settings.ini). We have documentation WIP https://github.com/rhinstaller/permian/pull/64 that talks about it and I will link it somewhere when it is done.

.github/workflows/webui-periodic.yml
rvykydal: I was in some doubts about need for this, but we are using the same for kickstart tests workflows. So if, I'd fix it globally then.

VladimirSlavik: Oh, then it probably works well enough, or we will find out the hard way. I was not sure about what volumes we have.

velezd (approved):

PR #4660, opened 2023-03-28
fsset: Catch SwapSpaceError when trying to activate swaps

vojtechtrefny:
Blivet now raises blivet.errors.SwapSpaceError when swap setup fails instead of just relaying the libblockdev exception.

We decided that raising libblockdev exceptions isn't probably the best idea so starting from blivet 3.8.0 you'll need to catch the blivet exception here. This should be the only place where libblockdev exceptions are being caught in Anaconda.

### DIFF
--- a/pyanaconda/modules/storage/devicetree/fsset.py
+++ b/pyanaconda/modules/storage/devicetree/fsset.py
@@ -25,7 +25,7 @@ from gi.repository import BlockDev as blockdev
 
 from blivet.devices import NoDevice, DirectoryDevice, NFSDevice, FileDevice, MDRaidArrayDevice, \
     NetworkStorageDevice, OpticalDevice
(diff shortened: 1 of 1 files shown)


pyanaconda/modules/storage/devicetree/fsset.py
```diff
@@ -530,7 +530,7 @@ def turn_on_swap(self, root_path=""):
                 try:
                     device.setup()
                     device.format.setup()
-                except blockdev.SwapActivateError as e:
+                except (SwapSpaceError, blockdev.SwapActivateError) as e:
```
vojtechtrefny, line 533: The `SwapSpaceError` is not new so this is backwards compatible and there's no need to wait for blivet 3.8.0 release.

VladimirSlavik (approved):
LGTM as far as I can tell (having seen the Blivet PR too).

PR #4661, opened 2023-03-28
Fedora 38 keep vga boot option

M4rtinK:
To address problems with KDE on Wayland on BIOS boots with "basic mode" graphics, we need to set a framebuffer mode in the bootloader (vga=791). This should also be passed through to the installed system so SDDM will work there as well.

Backport to F38.


VladimirSlavik (approved):

VladimirSlavik:
Can't find a reason why the tests don't run. I've seen this already. With @jkonecny12 we're somewhat certain this is a problem with GitHub. Maybe run the tests locally and just report the result?

M4rtinK:
Tested locally, all seems fine. :)

PR #4665, opened 2023-03-31
Fix wrong dracut timeout message

jkonecny12:
We added Anaconda error reporting when Dracut timeout during boot. If the reason is known issue we will print that to a user. However, in case it is not known the code failed with message:

... /run/anaconda/initrd_errors.txt: No such file or directory

To fix this just print "Reason unknown" and not execute rest of the code.


VladimirSlavik (approved):
Looks good to me, thank you!

PR #4672, opened 2023-04-03
WebUI: correct setState calls for SearchInput

jelly:
The onChange callback has two parameters, event and the string. We want to set the string not the event, furthermore simplify the clearing of state, there is no need to merge the old state, set state can work with keys.

@M4rtinK  This fixes the test failures in https://github.com/rhinstaller/anaconda/pull/4657

**Thank you** for your contribution!

Please check that your PR follows these rules:

* [x] [**Code conventions**](https://anaconda-installer.readthedocs.io/en/latest/contributing.html#code-conventions). tl;dr: Follow PEP8, wrap at 100 chars, and provide docstrings.

* [x] [**Commit message conventions**](https://anaconda-installer.readthedocs.io/en/latest/commit-log.html). tl;dr: Heading, empty line, longer explanations, all wrapped manually. If in doubt, write a longer commit message with more details.

* [x] **Tests** pass and cover your changes. You can [run tests locally manually](https://anaconda-installer.readthedocs.io/en/latest/testing.html), or have them run as part of the PR. A team member will have to enable tests manually after inspecting the PR.
  *If you don't know how, ask us for help!*

* [x] [**Release notes**](https://anaconda-installer.readthedocs.io/en/latest/contributing.html#release-notes) and **docs** if the PR adds something major or changes existing behavior.
  *If you don't know how, ask us for help!*

* [x] **RHEL** rules: If your PR is for a `rhel-*` branch, pay special attention to commit messages. Make sure **all** commit messages include a line linking the commit to a bug, such as `Resolves: rhbz#123456`. For more information, see [the complete rules](https://anaconda-installer.readthedocs.io/en/latest/commit-log.html).
  *If you don't know how, ask us for help!*

Don't forget - *if you don't know how, ask us for help!*

And now that you read this all, you can delete it and type your own description of your changes :-)

PR #4678, opened 2023-04-05
Github codeql scanning

jelly:
Github has a codeql action which can scan JavaScript, Python, C code and report issues on PR's. This is similar to shellcheck it shows up in the Security tab and as a comment in a PR.

Example alert:

image

Cockpit projects already set this up and it found a bunch of bugs / state inconsistenties.

### DIFF
--- a/ui/webui/src/components/installation/InstallationProgress.jsx
+++ b/ui/webui/src/components/installation/InstallationProgress.jsx
@@ -83,8 +83,6 @@ export class InstallationProgress extends React.Component {
                             }
                             if (message) {
                                 this.setState({ statusMessage: message });
-                            } else {
-                                this.setState({ step });
                             }
                         });
                         taskProxy.addEventListener("Failed", () => {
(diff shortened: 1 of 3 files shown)


jelly:
The remaining issues are fixed in https://github.com/rhinstaller/anaconda/pull/4672

ui/webui/src/components/installation/InstallationProgress.jsx
```diff
@@ -83,8 +83,6 @@ export class InstallationProgress extends React.Component {
                             }
                             if (message) {
                                 this.setState({ statusMessage: message });
-                            } else {
-                                this.setState({ step });
```
jkonecny12, line 87: Why is this not working? Could you please explain it to me :)?

jkonecny12 (approved):
Looks good to me. Thanks for more code scanning!

ui/webui/src/components/installation/InstallationProgress.jsx
jelly: It works fine, but nothing uses `this.state.step` so we are just setting it in state but not actually using it.

PR #4680, opened 2023-04-05
WebUI: simplify translation formatting

jelly:
ngettext is for translations which can be one byte and two bytes.

### DIFF
--- a/ui/webui/src/components/storage/InstallationDestination.jsx
+++ b/ui/webui/src/components/storage/InstallationDestination.jsx
@@ -172,7 +172,7 @@ const DropdownBulkSelect = ({
           isChecked={selectedCnt > 0 ? (selectedCnt === totalCnt ? true : null) : false}
           onChange={onChange}
         >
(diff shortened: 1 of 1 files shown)


KKoukiou (approved):

jkonecny12 (approved):

rvykydal (approved):
Thank you

jelly:
> There's literally a number in the output, so it's a message that can vary according to that number. The English string does not change, but is that true for all languages?

How would a translator figure that out? As both strings end up in the translation frontend as "$0 selected".

VladimirSlavik:
Translators know what plurals are, especially when the translation UI shows them as such...

KKoukiou:
This was correct as it was for the reason that @VladimirSlavik mentioned. Look at the italian translation for example.

[kkoukiou@sequioa anaconda]$ cat ui/webui/po/it.po | grep -A2 '0 selected' 
msgid "$0 selected"
msgid_plural "$0 selected"
msgstr[0] "$0 selezionato"
msgstr[1] "$0 selezionati"

PR #4681, opened 2023-04-05
Add missing documentation about OEMDRV (#2171811)

jkonecny12:
This is documentation part of the commit 1e33cbdbf45150308cef6b8c2b056c8cc41f0492.

Also add documentation of OEMDRV functionality for kickstart.

Extension of https://github.com/rhinstaller/anaconda/pull/4586 .

Related: rhbz#2171811

### DIFF
--- a/docs/boot-options.rst
+++ b/docs/boot-options.rst
@@ -656,6 +656,22 @@ will be used by default, if found.
 
 See the |anacondalogging|_ for more info on setting up logging via virtio.
 
+.. inst.wait_for_disks:
+
+inst.wait_for_disks
+^^^^^^^^^^^^^^^^^^^
+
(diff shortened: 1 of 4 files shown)


VladimirSlavik (approved):
LGTM, thank you!

I wonder if the values for the option can be fractional, eg. 12.56 seconds... but that is probably meaningless for the actual wait times.

jkonecny12:
> LGTM, thank you!
> 
> I wonder if the values for the option can be fractional, eg. 12.56 seconds... but that is probably meaningless for the actual wait times.

The number is not even in seconds so I don't think this idea would be helpful to anyone :).

jkonecny12:
Updated the commit name.

PR #4684, opened 2023-04-06
Move from webpack to esbuild bundler

KKoukiou:


VladimirSlavik (approved):
Looks good to me, thank you!

VladimirSlavik (approved):
Thank you!

The last commit looks like a caveat for any future code - is there something deeper to be aware of?

KKoukiou:
> Thank you!
> 
> The last commit looks like a caveat for any future code - is there something deeper to be aware of?

Actually I did not understand the reason behind this just yet. Let's sync about it on Monday.

M4rtinK (approved):
Looks good to me as well & pending the translation issues to be resolved. :)

VladimirSlavik:
> Actually I did not understand the reason behind this just yet. Let's sync about it on Monday.

No problem! Having a translation at top level of module sounds like a bad idea, anyway - it kind of feels like the translation would be applied on loading, and then when the language changes, the variable would keep the old value. Maybe it does not work that way, though.

KKoukiou:
> > Actually I did not understand the reason behind this just yet. Let's sync about it on Monday.
> 
> No problem! Having a translation at top level of module sounds like a bad idea, anyway - it kind of feels like the translation would be applied on loading, and then when the language changes, the variable would keep the old value. Maybe it does not work that way, though.

I don't understand why it worked and now it does not. but as you mention - this is an antipractice, translated strings should not be globals.

VladimirSlavik:
Re-running unit test - it should be some external problem, I canẗ think of any related change recently :-/

jelly:
> Re-running unit test - it should be some external problem, I canẗ think of any related change recently :-/

Yeah, same occurs here https://github.com/rhinstaller/anaconda/actions/runs/4793803386/jobs/8532928204

jelly (approved):

PR #4687, opened 2023-04-07
Implement EULA spoke based on initial-setup

mikhailnov:
Copied code from initial-setup and adapted it for Anaconda.

It makes sense to show EULA before installation, not just text that it is in some file after the installation.

TODO: It would be great to automatically choose license in the current language. Currently it is worked around in ROSA by creating a drop-in config: https://abf.io/import/anaconda/blob/6dc5b78efd/anaconda.sh#lc-77

Closes: https://github.com/rhinstaller/anaconda/discussions/4686

It looks like this:

Here is Russian translation of new strings (they were not inside po and pot files in initial-setup):
```po
msgid "LICENSING"
msgstr "ЛИЦЕНЗИРОВАНИЕ"

msgctxt "GUI|Spoke"
msgid "_License Information"
msgstr "О _лицензии"

msgid "License information"
msgstr "О лицензии"

msgid "License Information"
msgstr "О лицензии"

msgid "License accepted"
msgstr "Лицензия принята"

msgid "License not accepted"
msgstr "Лицензия не принята"

msgid "Read the License Agreement"
msgstr "Прочитайте лицензионное соглашение"

msgid "I accept the license agreement"
msgstr "Принимаю лицензионное соглашение"

msgid "I _accept the license agreement"
msgstr "_Принимаю лицензионное соглашение"

msgid "License Agreement:"
msgstr "Лицензионное соглашение:"
```

In some places translation is not exact because otherwise words will be too long.

ISO image (LiveCD + installer) with Anaconda with EULA spoke: https://file-store.rosalinux.ru/download/2aaf53506bc7c689fa81ab1de078619afa6db568 (ROSA_2021.1_XFCE_x86_64_8.iso)

### DIFF
--- a/pyanaconda/ui/gui/spokes/eula.py
+++ b/pyanaconda/ui/gui/spokes/eula.py
@@ -1,0 +1,107 @@
+import logging
+
+from pyanaconda.ui.common import FirstbootOnlySpokeMixIn
+from pyanaconda.ui.gui.spokes import NormalSpoke
+from pyanaconda.core.i18n import _, CN_
+from pyanaconda.core import eula
+from pyanaconda.ui.categories.eula import LicensingCategory
+from pyanaconda.anaconda_loggers import get_module_logger
+from pykickstart.constants import FIRSTBOOT_RECONFIG
+
+log = get_module_logger(__name__)
+__all__ = ["EULASpoke"]
+
+
+class EULASpoke(FirstbootOnlySpokeMixIn, NormalSpoke):
+    """The EULA spoke"""
+
+    builderObjects = ["eulaBuffer", "eulaWindow"]
+    mainWidgetName = "eulaWindow"
+    uiFile = "spokes/eula.glade"
+    icon = "application-certificate-symbolic"
+    title = CN_("GUI|Spoke", "_License Information")
+    category = LicensingCategory
+
+    @staticmethod
+    def get_screen_id():
+        """Return a unique id of this UI screen."""
+        return "license-information"
+
+    def initialize(self):
+        log.debug("initializing the EULA spoke")
+        NormalSpoke.initialize(self)
+
+        self._have_eula = True
+        self._eula_buffer = self.builder.get_object("eulaBuffer")
+        self._agree_check_button = self.builder.get_object("agreeCheckButton")
+        self._agree_label = self._agree_check_button.get_child()
+        self._agree_text = self._agree_label.get_text()
+
+        log.debug("looking for the license file")
+        license_file = eula.get_license_file_name()
+        if not license_file:
+            log.error("no license found")
+            self._have_eula = False
+            self._eula_buffer.set_text(_("No license found"))
+            return
+
+        # if there is "eula <...>" in kickstart, use its value
+        if self.data.eula.agreed is not None:
+            self._agree_check_button.set_active(self.data.eula.agreed)
+
+        self._eula_buffer.set_text("")
+        itr = self._eula_buffer.get_iter_at_offset(0)
+        log.debug("opening the license file")
+        with open(license_file, "r") as fobj:
+            # insert the first line without prefixing with space
+            try:
+                first_line = next(fobj)
+            except StopIteration:
+                # nothing in the file
+                return
+            self._eula_buffer.insert(itr, first_line.strip())
+
+            # EULA file may be preformatted for the console, we want to let Gtk
+            # format it (blank lines should be preserved)
+            for line in fobj:
+                stripped_line = line.strip()
+                if stripped_line:
+                    self._eula_buffer.insert(itr, " " + stripped_line)
+                else:
+                    self._eula_buffer.insert(itr, "\n\n")
+
+    def refresh(self):
+        self._agree_check_button.set_sensitive(self._have_eula)
+        self._agree_check_button.set_active(self.data.eula.agreed)
+
+    def apply(self):
+        self.data.eula.agreed = self._agree_check_button.get_active()
+
+    @property
+    def completed(self):
+        return not self._have_eula or self.data.eula.agreed
+
+    @property
+    def status(self):
+        if not self._have_eula:
+            return _("No license found")
+
+        return _("License accepted") if self.data.eula.agreed else _("License not accepted")
+
+    @classmethod
+    def should_run(cls, environment, data):
+        if eula.eula_available():
+            # don't run if we are in initial-setup in reconfig mode and the EULA has already been accepted
+            if FirstbootOnlySpokeMixIn.should_run(environment, data) and data and data.firstboot.firstboot == FIRSTBOOT_RECONFIG and data.eula.agreed:
+                log.debug("not running license spoke: reconfig mode & license already accepted")
+                return False
+            return True
+        return False
+
+    def on_check_button_toggled(self, *args):
+        if self._agree_check_button.get_active():
+            log.debug("license is now accepted")
+            self._agree_label.set_markup("<b>%s</b>" % self._agree_text)
+        else:
+            log.debug("license no longer accepted")
+            self._agree_label.set_markup(self._agree_text)
--- a/data/anaconda.conf
+++ b/data/anaconda.conf
@@ -316,9 +316,6 @@ password_policies =
 # If the given distribution has an EULA & feels the need to
 # tell the user about it fill in this variable by a path
 # pointing to a file with the EULA on the installed system.
-#
-# This is currently used just to show the path to the file to
-# the user at the end of the installation.
 eula =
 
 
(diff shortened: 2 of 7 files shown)


pyanaconda/ui/gui/spokes/eula.py
```diff
@@ -0,0 +1,103 @@
...
+from pyanaconda.core import eula
+from pyanaconda.ui.categories.eula import LicensingCategory
+from pyanaconda.anaconda_loggers import get_module_logger
+from pykickstart.constants import FIRSTBOOT_RECONFIG
+
+log = get_module_logger(__name__)
+__all__ = ["EULASpoke"]
+
+
+class EULASpoke(FirstbootOnlySpokeMixIn, NormalSpoke):
```
VladimirSlavik, line 15: `FirstbootOnlySpokeMixIn` practically says it runs only in initial setup. Is that really the right parent to use? I'd expect only `NormalSpoke` here...

pyanaconda/ui/gui/spokes/eula.py
```diff
@@ -0,0 +1,103 @@
...
+
+    @classmethod
+    def should_run(cls, environment, data):
+        if eula.eula_available():
+            # don't run if we are in initial-setup in reconfig mode and the EULA has already been accepted
+            if FirstbootOnlySpokeMixIn.should_run(environment, data) and data and data.firstboot.firstboot == FIRSTBOOT_RECONFIG and data.eula.agreed:
+                log.debug("not running license spoke: reconfig mode & license already accepted")
+                return False
+            return True
+        return False
```
VladimirSlavik, line 95: This needs changing a bit. All anaconda spokes work such that they are always visible, and kickstart changes only if they are completed or not.
- Since the spoke is not reused by initial setup, I think you can leave out the `FIRSTBOOT_RECONFIG` part. @M4rtinK - is that correct please?
- The spoke visibility in `should_run` should depend only on eula presence.
- The `data.eula.agreed` bit should set up the checkbox, I think in `initialize`.

pyanaconda/ui/gui/spokes/eula.py
```diff
@@ -0,0 +1,103 @@
...
+                    self._eula_buffer.insert(itr, " " + stripped_line)
+                else:
+                    self._eula_buffer.insert(itr, "\n\n")
+
+    def refresh(self):
+        self._agree_check_button.set_sensitive(self._have_eula)
+        self._agree_check_button.set_active(self.data.eula.agreed)
+
+    def apply(self):
+        self.data.eula.agreed = self._agree_check_button.get_active()
```
VladimirSlavik, line 74: Does writing to `self.data` also make the `output-ks.cfg` contain `eula --agreed` ? If so, that's a thing that needs adding to the modularization todos.

VladimirSlavik (commented):
Thanks! This looks good.

A couple of thoughts, in no particular order:

- Please add license to the code files, and don't forget to set correct year :)
- Probably needs team discussion if this should be a community feature or not. Personally I think that's fine as-is, there is very little code and it's usable primarily for RHEL anyway.
- It's debatable if the spoke should be also disabled by the default config or not. Ultimately that's a minor point, though.
- Note to self - the agreed/disagreed state needs to be stored in modularized backend, eventually.

pyanaconda/ui/gui/spokes/eula.py
mikhailnov: /root/anaconda-ks.cfg of the installed system contains `eula --agreed`:

image

That should allow initial-setup to detect agreed EULA when reusing this spoke and do not show the license if it has already been agreed with (usage scenarios may be different, however, maybe some people will want to force users to agree with it once again, they can edit `/root/anaconda-ks.cfg`).

pyanaconda/ui/gui/spokes/eula.py
mikhailnov: > The data.eula.agreed bit should set up the checkbox, I think in initialize

Do you mean running GUI/TUI with a kickstart with `eula --agreed`? What is the use case for this, who and why will do that? Is it possible without hackery?

pyanaconda/ui/gui/spokes/eula.py
VladimirSlavik: Initial setup has not been touched for a long time, is hardly a priority, and we're asking around to see if we can just drop it completely. So I'd expect it to just keep its own eula spoke and this be in anaconda only.

pyanaconda/ui/gui/spokes/eula.py
VladimirSlavik: Yes, it's actually very easy. If you don't make anaconda stop on incomplete kickstart, you can use a "partial kickstart" that does not satisfy all needed inputs, and the UIs will be pre-filled when it stops on the hub. It's a thing that is not really nice to support, but invaluable for testing. I run 99% of my anaconda debugging with this, so that I don't have to set it all every time I retry my changes :P Some spins also use this, except hidden in the "default kickstart", together with hiding spokes in config, to set up things that users can't change. It's a bit of a dark art, really.

The rest of the UI works this way, so it would be best to make it consistent.

pyanaconda/ui/gui/spokes/eula.py
VladimirSlavik: Thanks, that's perfect. I was not sure if the data attribute is writable like that.

pyanaconda/ui/gui/spokes/eula.py
M4rtinK: Yeah, that might make more sense - not that much code is needed for EULA display, so it might make sense to just duplicate it in Anaconda, rather than to keep the screen reusable, with that special mixin and the necessary Initial Setup changes.

pyanaconda/ui/gui/spokes/eula.py
M4rtinK: >Since the spoke is not reused by initial setup, I think you can leave out the FIRSTBOOT_RECONFIG part. @M4rtinK - is that correct please?

Yes, I think it really makes more sense to just add the ability to Anaconda to show the EULA without also trying to reuse that screen from Initial Setup. This nicely shows the benefits of that - we can drop the Initial setup specific mixin (which still refers to the old "Firstboot" tool, to be even more confusing) as well as the Initial Setup specific reconfig functionality. 

As for when the EULA spoke should show up - I think we should just check if an EULA is in the expected place & show it if there is something. That's how it works for Initial Setup & should work for the installer as well.

VladimirSlavik:
Note to self: The eula command would be in Payloads module, spoke under Software category. Right now, this is low priority compared to modularization for webui in f39 (hopefully).

mikhailnov:
Sorry for very long delay :(

Me and we in ROSA like the initial-setup utility. We began using it recently and like it. It has a nice architecture and nice integration with Anaconda. Its modularity (ability to add spokes) makes it be a great thing, better than e.g. GNOME's initial setup. Its UI is probably not the nicest, but it's OK, it looks good enough, and custom CSS can be written in theory.

We are now making it possible to use initial-setup as the first time configuration wizard for OEM installations: a user buys a computer with Linux preinstalled and configures it via initial-setup (user, root password, EULA, custom spokes may be added).

We are going to use the EULA spoke in both Anaconda and initial-setup, that is why I would like to avoid removing possibility to reuse it by initial-setup. but of course I can make it as you suggest here in Anaconda upstream and patch it in the downstream, but it is not he best way...

mikhailnov:
I've implemented reading "eula --agreed" from kickstart in GUI. It was implemented in TUI.

I can also remove integration with initial-setup, but I would better leave it here in Anaconda and remove EULA code from initial-setup.

mikhailnov:
Could you please give comments, should I make further changes?

VladimirSlavik:
Sorry, we were focused on the WebUI moonshot, which is sort of undecided until beta (freeze?), I believe. I'll take a look again once there's time to do so.

mikhailnov:
Ok, thanks)

Conan-Kudo:
@VladimirSlavik ping

KKoukiou:
Hi @mikhailnov, thank you for your contribution! After careful consideration during our team meeting, we believe that introducing this feature as an add-on would align better with our current goals. The anaconda team is willing to support the necessary work for porting this to an add-on solution.

We've identified a few considerations against the current approach:

1. The current PR doesn't build on a modularized API, which is a requirement for new features.
2. The EULA spoke may not be necessary for all supported distros (e.g., not needed for RHEL/Centos), as it's assumed that users confirm the license agreement when downloading the ISO.
3. Introducing new features in the GUI only, is not aligned with the team's strategy to move slowly towards the new Web UI solution.
4. We aim to maintain the existing installer behavior, allowing users to perform non-interactively installations, by configuring everything through kickstart. Introducing a new option as is in this PR might will the current behavior.

Given these factors, we suggest reconsidering this as an add-on.

Would you be open to taking this approach? The team is committed to actively supporting you during the migration process. Apologies for the delayed response, and we appreciate your understanding.

mikhailnov:
Hello!
What is the current status of GTK gui in Anaconda?

VladimirSlavik:
To be deprecated, I believe.

PR #4689, opened 2023-04-11
GUI kernel switcher for ARM 64k and Gemini

OndrejZobal:
**outdated**
Implementation for the Kernel switcher GUI inside the Software selection spoke.

Since the Gemini kernel isn't in the repos at the moment, it's kernel name is set to "kernel-rt" to make testing easier, this should change before this gets merged.

This implementation doesn't set the special kernel as default, yet.

### DIFF
--- a/pyanaconda/ui/gui/spokes/software_selection.py
+++ b/pyanaconda/ui/gui/spokes/software_selection.py
@@ -37,8 +37,9 @@ from pyanaconda.ui.gui.spokes.lib.detailederror import DetailedErrorDialog
 from pyanaconda.ui.gui.utils import blockedHandler, escape_markup
 from pyanaconda.ui.categories.software import SoftwareCategory
 from pyanaconda.ui.lib.subscription import check_cdn_is_installation_source
-from pyanaconda.ui.lib.software import FEATURE_64K, KernelFeatures, \
+from pyanaconda.ui.lib.software import FEATURE_UPSTREAM, FEATURE_64K, KernelFeatures, \
     get_kernel_from_properties, get_available_kernel_features, get_kernel_titles_and_descriptions
+
 from pyanaconda.modules.common.constants.services import SUBSCRIPTION
 from pyanaconda.modules.common.util import is_module_available
 
@@ -134,17 +135,25 @@ class SoftwareSelectionSpoke(NormalSpoke):
         self._kernel_box = self.builder.get_object("kernelBox")
         self._combo_kernel_page_size = self.builder.get_object("kernelPageSizeCombo")
         self._label_kernel_page_size = self.builder.get_object("kernelPageSizeLabel")
+        self._combo_kernel_version = self.builder.get_object("kernelVersionCombo")
+        self._label_kernel_version = self.builder.get_object("kernelVersionLabel")
 
         # Normally I would create these in the .glade file but due to a bug they weren't
         # created properly
         self._model_kernel_page_size = Gtk.ListStore(str, str)
+        self._model_kernel_version = Gtk.ListStore(str, str)
 
         kernel_labels = get_kernel_titles_and_descriptions()
         for i in ["4k", "64k"]:
             self._model_kernel_page_size.append([i, "<b>%s</b>\n%s" % \
                                                 (escape_markup(kernel_labels[i][0]),
                                                 escape_markup(kernel_labels[i][1]))])
+        for i in ["upstream", "standard"]:
+            self._model_kernel_version.append([i, "<b>%s</b>\n%s" % \
+                                              (escape_markup(kernel_labels[i][0]),
+                                              escape_markup(kernel_labels[i][1]))])
         self._combo_kernel_page_size.set_model(self._model_kernel_page_size)
+        self._combo_kernel_version.set_model(self._model_kernel_version)
         self._available_kernels = get_available_kernel_features(self.payload)
 
     # Payload event handlers
@@ -215,7 +224,10 @@ class SoftwareSelectionSpoke(NormalSpoke):
         # Select kernel
         property_64k = self._available_kernels[FEATURE_64K] and \
             self._combo_kernel_page_size.get_active_id() == FEATURE_64K
-        kernel_properties = KernelFeatures(property_64k)
+        property_upstream = self._available_kernels[FEATURE_UPSTREAM] and \
+            self._combo_kernel_version.get_active_id() == FEATURE_UPSTREAM
+
+        kernel_properties = KernelFeatures(property_upstream, property_64k)
         kernel = get_kernel_from_properties(kernel_properties)
         if kernel is not None:
             log.debug("Selected kernel package: %s", kernel)
@@ -587,6 +599,12 @@ class SoftwareSelectionSpoke(NormalSpoke):
         if show_kernels:
             self._kernel_box.set_visible(True)
             self._kernel_box.set_no_show_all(False)
+
+            # Kernel version combo
+            self._combo_kernel_version.set_visible(self._available_kernels[FEATURE_UPSTREAM])
+            self._combo_kernel_version.set_no_show_all(not self._available_kernels[FEATURE_UPSTREAM])
+            self._label_kernel_version.set_visible(self._available_kernels[FEATURE_UPSTREAM])
+            self._label_kernel_version.set_no_show_all(not self._available_kernels[FEATURE_UPSTREAM])
 
             # Arm 64k page size kernel combo
             self._combo_kernel_page_size.set_visible(self._available_kernels[FEATURE_64K])
--- a/pyanaconda/ui/gui/spokes/software_selection.py
+++ b/pyanaconda/ui/gui/spokes/software_selection.py
@@ -37,7 +37,8 @@ from pyanaconda.ui.gui.spokes.lib.detailederror import DetailedErrorDialog
 from pyanaconda.ui.gui.utils import blockedHandler, escape_markup
 from pyanaconda.ui.categories.software import SoftwareCategory
 from pyanaconda.ui.lib.subscription import check_cdn_is_installation_source
-
+from pyanaconda.ui.lib.software import FEATURE_64K, KernelFeatures, \
+    get_kernel_from_properties, get_available_kernel_features, get_kernel_titles_and_descriptions
 from pyanaconda.modules.common.constants.services import SUBSCRIPTION
 from pyanaconda.modules.common.util import is_module_available
 
@@ -129,6 +130,23 @@ class SoftwareSelectionSpoke(NormalSpoke):
         self._fake_radio = Gtk.RadioButton(group=None)
         self._fake_radio.set_active(True)
 
+        # Display a group of options for selecting desired properties of a kernel
+        self._kernel_box = self.builder.get_object("kernelBox")
+        self._combo_kernel_page_size = self.builder.get_object("kernelPageSizeCombo")
+        self._label_kernel_page_size = self.builder.get_object("kernelPageSizeLabel")
+
+        # Normally I would create these in the .glade file but due to a bug they weren't
+        # created properly
+        self._model_kernel_page_size = Gtk.ListStore(str, str)
+
+        kernel_labels = get_kernel_titles_and_descriptions()
+        for i in ["4k", "64k"]:
+            self._model_kernel_page_size.append([i, "<b>%s</b>\n%s" % \
+                                                (escape_markup(kernel_labels[i][0]),
+                                                escape_markup(kernel_labels[i][1]))])
+        self._combo_kernel_page_size.set_model(self._model_kernel_page_size)
+        self._available_kernels = get_available_kernel_features(self.payload)
+
     # Payload event handlers
     def _downloading_package_md(self):
         # Reset the error state from previous payloads
@@ -193,6 +211,16 @@ class SoftwareSelectionSpoke(NormalSpoke):
 
         for group_name in self._get_selected_addons():
             self._selection.groups.append(group_name)
+
+        # Select kernel
+        property_64k = self._available_kernels[FEATURE_64K] and \
+            self._combo_kernel_page_size.get_active_id() == FEATURE_64K
+        kernel_properties = KernelFeatures(property_64k)
+        kernel = get_kernel_from_properties(kernel_properties)
+        if kernel is not None:
+            log.debug("Selected kernel package: %s", kernel)
+            self._selection.packages.append(kernel)
+            self._selection.excluded_packages.append("kernel")
 
         log.debug("Setting new software selection: %s", self._selection)
         self.payload.set_packages_data(self._selection)
@@ -426,6 +454,7 @@ class SoftwareSelectionSpoke(NormalSpoke):
         self.refresh_addons()
         self._environment_list_box.show_all()
         self._addon_list_box.show_all()
+        self._refresh_kernel_features()
 
     def _add_addon(self, grp):
         (name, desc) = self.payload.group_description(grp)
@@ -543,6 +572,31 @@ class SoftwareSelectionSpoke(NormalSpoke):
         for child in listbox.get_children():
             listbox.remove(child)
             del(child)
+
+    def _refresh_kernel_features(self):
+        """Display options for selecting kernel features."""
+
+        # Only showing parts of kernel box relevant for current system.
+        self._available_kernels = get_available_kernel_features(self.payload)
+        show_kernels = False
+        for (_key, val) in self._available_kernels.items():
+            if val:
+                show_kernels = True
+                break
+
+        if show_kernels:
+            self._kernel_box.set_visible(True)
+            self._kernel_box.set_no_show_all(False)
+
+            # Arm 64k page size kernel combo
+            self._combo_kernel_page_size.set_visible(self._available_kernels[FEATURE_64K])
+            self._combo_kernel_page_size.set_no_show_all(not self._available_kernels[FEATURE_64K])
+            self._label_kernel_page_size.set_visible(self._available_kernels[FEATURE_64K])
+            self._label_kernel_page_size.set_no_show_all(not self._available_kernels[FEATURE_64K])
+        else:
+            # Hide the entire box.
+            self._kernel_box.set_visible(False)
+            self._kernel_box.set_no_show_all(True)
 
     @property
     def txid_valid(self):
--- a/pyanaconda/ui/lib/software.py
+++ b/pyanaconda/ui/lib/software.py
@@ -19,13 +19,16 @@ from collections import namedtuple
 from blivet.arch import is_arm
 from pyanaconda.core.i18n import _
 
+FEATURE_UPSTREAM = "upstream"
 FEATURE_64K = "64k"
-KernelFeatures = namedtuple("KernelFeatures", ["page_size_64k"])
+
+KernelFeatures = namedtuple("KernelFeatures", ["upstream", "page_size_64k"])
 
 def get_available_kernel_features(payload):
     """Returns a dictionary with that shows which kernels should be shown in the UI.
     """
     features = {
+        FEATURE_UPSTREAM: any(payload.match_available_packages("kernel-redhat")),
         FEATURE_64K: is_arm() and any(payload.match_available_packages("kernel-64k"))
     }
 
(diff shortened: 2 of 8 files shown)


data/anaconda.conf
```diff
@@ -91,6 +91,9 @@ updates_repositories =
 #
 enabled_repositories_from_treeinfo = addon optional variant
 
+# Show experimental kernel option in the menu.
+show_gemini_kernel = False
```
poncovka, line 95: Since it is related only to the user interface, I would consider to use the `[User Interface]` section.

pyanaconda/ui/gui/spokes/software_selection.py
```diff
@@ -526,6 +556,22 @@ def _get_selected_addons(self):
 
         return retval
 
+    def _get_kernel_from_properties(self):
```
poncovka, line 559: With a little effort, this method could be replaced with a function that could be tested with unit tests. We usually use `pyanaconda.ui.lib` for this kind of code.

jkonecny12 (commented):
Thanks a lot for this. It looks great! I have a few clarifications / notes.

* I would love to see screenshots so we can discuss UX part of the change.
* The gemini kernel implementation does not need (shouldn't) to be backported to Fedora. Ideally it should be independent commit so it could be excluded easily.
* ARM 64k is not supported on Fedora so it has to be hidden under configuration option to not break stuff.

OndrejZobal:
obrazek

Forgot to add screenshot from the GUI.

The text will probably need to change, but this is what I came up with. By using "kernel properties" in the heading I wanted to emphasize that you can combine these options

OndrejZobal:
Ok, so I refactored the code. 

I felt that I should involve the `get_kernel_packages` function from `pyanaconda/modules/payloads/payload/dnf/utils.py`, which previously used a hard-coded value of `kernel` as the lowest priority fall-back. But it is possible that I misunderstood the meaning of this function so please correct me if I made a mistake.

In order to run the custom kernel package name through this function, I had somehow signal that a different kernel was chosen in the UI. Using regexp isn't really an option since there are non-kernel packages that begin with `kernel-.*` such as `kernel-tools`, so **I had no other option than to add `included_kernels` property to `RepoConfigurationData` API.**

## Before merging:
- [ ] Tests need fixing
- [ ] Kickstart export is missing the kernel package

Another missing part of the implementation is TUI. I intend to implement it in a separate PR after merging this.

Kickstart tests cannot be added until `kernel-redhat` is available in the repos.

## Trying this PR
Missing `kernel-redhat` will prevent you from testing this PR. You can you around this by applying the following [patch](https://pastebin.com/MPaPiH21) which will install `kernel-rt` instead of `kernel-redhat`. And adding the RT repo.

## Screenshot
obrazek

pyanaconda/ui/gui/spokes/software_selection.py
```diff
@@ -129,6 +132,34 @@ def __init__(self, *args, **kwargs):
...
+        self.kernel_box = self.builder.get_object("kernelBox")
+        self.check_button_redhat_kernel = self.builder.get_object("redhatKernelCheckButton")
+        self.check_button_64k = self.builder.get_object("64kCheckButton")
+        self._init_kernel_properties()
+
+    def _init_kernel_properties(self):
+        """Display options for selecting kernel properties."""
+        # Only showing parts of kernel box relevant for current system.
+        show_64k = is_arm() and conf.ui.show_arm_64k_kernel()
+        show_redhat = conf.ui.show_redhat_kernel()
```
poncovka, line 145: So I checked how difficult it would be verify the availability of the packages, so you don't have to blindly rely on the configuration settings, and maybe it is worth it. Upstream has the following function that you could copy to RHEL 9 and, during the backport to upstream (if it is planned), this function can be easily be exposed via DBus API.

https://github.com/rhinstaller/anaconda/blob/ea679e9aa9d5bee788c918e4e61ba208bb23b1d6/pyanaconda/modules/payloads/payload/dnf/dnf_manager.py#L496-L507

So I am suggesting to drop the configuration options and check the availability of packages to improve the user experience. What do you think, @jkonecny12?

poncovka (commented):
I am afraid that this solution doesn't cover all use cases.

pyanaconda/ui/gui/spokes/software_selection.py
```diff
@@ -194,6 +226,14 @@ def apply(self):
         for group_name in self._get_selected_addons():
             self._selection.groups.append(group_name)
 
+        # Select kernel.
+        property_redhat_kernel = self.check_button_redhat_kernel.get_active()
+        property_64k = self.check_button_64k.get_active()
+        kernel = get_kernel_from_properties(property_redhat_kernel, property_64k)
+        if kernel is not None:
+            log.debug("Selected kernel package: %s", kernel)
+            self._selection.kernel_packages.append(kernel)
```
poncovka, line 235: Btw. I don't see code of the `refresh` method that would update the value of checkboxes based on the current packages selection.

poncovka:
The UI is killing me, but that's not you fault. This is my version, but I would ask someone from UX (Máirín?) for help with the design and someone from documentation (Sagar?) for help with the text.

image

jkonecny12:
hi @mairin , could you please take a look on the suggestions of the new feature in Anaconda above? It would be great to have a help from the designer.

pyanaconda/ui/gui/spokes/software_selection.py
jkonecny12: I like this solution. I'm always for not introducing a new API :).

mairin:
@jkonecny12 can you give me some context about the workflow here? The idea is to provide users some options for kernel selection? What is Gemini kernel, is that 64k memory pages kernel? What is RH kernel? Does "RH kernel" as a term apply for non RHEL installation as well? I just need a little more context on what's going on here.

jkonecny12:
@mairin hi,

Please don't focus on names because these will most probably change. However, the functionality is to provide users ability to select another than the default kernel for installation. Right now we will have two but another one is comming.

If users will select both of the options than we will take just one kernel which has capabilities of both. In other words there has to be (not our problem to solve) kernel variants for all the combinations. Anaconda will always install just one kernel with all the properties selected.

Right now the request is mainly for ARM kernel (will be visible just on ARM) for 64k memory pages sizes.

Hopefully, this gives you more clarification.

data/anaconda.conf
```diff
@@ -302,6 +302,10 @@ password_policies =
     user (quality 1, length 6, empty)
     luks (quality 1, length 6)
 
+# Show 64k memory page kernel on ARM.
+show_arm_64k_kernel = False
```
poncovka, line 306: Please, add an extra line between the two new options.

poncovka (commented):
The backend changes need some polishing, but they look more reasonable now. I will leave the rest to other reviewers. Thank you!

data/anaconda.conf
jkonecny12: If we are going to detect the packages than we don't need them?

mairin:
@jkonecny12 it's hard for me to make a recommendation without understanding the reasons why a user would choose one, the other, or both options.... some more questions:

-  if 64k memory pages only shows for ARM installs... what happens if the user doesn't select 64k memory pages? does the kernel you give them still work? 
- I don't understand the Red Hat kernel option either - what is this? Does this only show up on RHEL, and if so why would you not get a RHT kernel when installing RHEL? If that's not it, what is the need for this option? And why would the user want to do that?

Generically / aesthetically / layout-wise I would probably recommend something like this, with the huge caveat that without really understanding the scenarios / options here this might be a terrible experience for the user.... but this makes the controls all look consistent:

anaconda-kernel-select


But my concerns, for example (may not apply but you see where I'm thinking):

- if ARM users should be using 64k page kernels, for example, it's probably better to detect they're using ARM and give that to them automatically with no prompting. Otherwise the checkbox becomes a "break my install / don't break my install" button (assuming 64k page kernels are required for ARM)

- User has no idea what Red Hat kernel options means or what will happen if they click on it, so they don't - what is the point of the option, right?

mairin:
Hey Jiri, I did see that (that's the comment I was kind of responding to)
but it doesn't give me enough context to understand the best way to present
the info.

mairin:
(What I mocked up is fine though, but that's just visual design, and given
the entire context might not be the best solution. Up to you how you want
to go though! :-) )

jkonecny12:
Thanks  a lot for your help @mairin !

pyanaconda/modules/payloads/payload/dnf/utils.py
```diff
@@ -44,13 +45,17 @@ def get_default_environment(dnf_manager):
...
-def get_kernel_package(dnf_base, exclude_list):
+def get_kernel_package(dnf_base, include_list, exclude_list):
     """Get an installable kernel package.
 
     :param dnf_base: a DNF base
     :param exclude_list: a list of excluded packages
     :return: a package name or None
     """
+    for package in include_list:
+        if package in PACKAGE_KERNELS and package not in exclude_list:
```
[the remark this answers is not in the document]
OndrejZobal: If there is a kernel package in the include list that is not excluded, I do not need to add another kernel package, so I return `None`.

The usecase would be:

```
%packages
kernel-redhat
%end
```
The snippet will match kernel-redhat as a kernel package and will not add any other kernel packages, as only one kernel should be installed by default.

pyanaconda/core/constants.py
```diff
@@ -493,6 +493,14 @@ class DisplayModes(Enum):
     GROUP_PACKAGE_TYPE_CONDITIONAL,
 ]
 
+PACKAGE_KERNELS = [
+    "kernel",
+    "kernel-rt",
+    "kernel-redhat",
+    "kernel-64k",
+    "kernel-redhat-64k",
+]
```
Conan-Kudo, line 502: Can we please _not_ hardcode this into the Anaconda code? Aside from the fact that Fedora and RHEL aren't the only distributions that use Anaconda (and thus hardcoding kernel package names becomes really problematic), the names of package variants will differ in Fedora and RHEL/CentOS.

As an example, Fedora Asahi 16k kernel vs standard Fedora 4k kernel for Apple Silicon. Another example is CentOS Hyperscale's kernel vs the standard RHEL kernel.

We don't know what we're going to do in the ecosystem now and the future, but we _do_ know that hardcoding things is a seriously bad idea.

cc: @davide125

Conan-Kudo (changes requested):

pyanaconda/core/constants.py
OndrejZobal: @poncovka

pyanaconda/modules/payloads/payload/dnf/utils.py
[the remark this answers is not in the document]
poncovka: I have updated the list of use cases based on some comments from @jkonecny12. It looks like the current back-end support is sufficient and it won't be necessary to change the `get_kernel_package` function after all.

pyanaconda/core/constants.py
poncovka: @jkonecny12 will verify that, but it looks like we won't need this list after all. See https://github.com/rhinstaller/anaconda/pull/4689#discussion_r1188889907.

OndrejZobal:
kernel options

So here is another iteration of the GUI implementation!

I am having some issued with the TUI implementation, becase the dnf metadata seem to be uninitialized.

OndrejZobal:
Looks like I messed up splitting my changes into commits and forgot to delete some debug logs :( . Just uploaded a fix.

Conan-Kudo:
Can we make the kernel switcher fully defined from anaconda data configuration rather than hardcoding awareness of specific kernel flavors?

tests/unit_tests/pyanaconda_tests/ui/test_ui_software.py
```diff
@@ -0,0 +1,27 @@
...
+import unittest
+from pyanaconda.ui.lib.software import get_kernel_from_properties, KernelFeatures
+
+class SoftwareUITestCase(unittest.TestCase):
+    """Test the UI functions and classes of the software-selection object."""
+    def test_get_kernel_from_properties(self):
+        assert get_kernel_from_properties(KernelFeatures(upstream=False, page_size_64k=False)) is None
+        assert get_kernel_from_properties(KernelFeatures(upstream=True, page_size_64k=False)) == "kernel-redhat"
+        assert get_kernel_from_properties(KernelFeatures(upstream=False, page_size_64k=True)) == "kernel-64k"
+        assert get_kernel_from_properties(KernelFeatures(upstream=True, page_size_64k=True)) == "kernel-redhat-64k"
```
jkonecny12, line 27: Maybe a bit more tests for the coverage wouldn't be bad thing to do.

jkonecny12 (changes requested):
Thanks a lot for this. In general looks good to me. See the notes below.

Also the commit messages are not correct sometimes and you are missing rhbz numbers (Resolves / Related)

jkonecny12:
We can do that as future extension but so far we have just one requirement for that so I don't see official reason for extending the API.

jkonecny12 (changes requested):
Thanks for solving the GUI issues. Seems to be working just fine now :)

However, I found more issues on TUI side:
- When booted without repository and then the repository is set by the Source Spoke the Kernel Selection spoke is not showed (similar as issue before in the GUI)
- When booted with kernel-redhat repository and going to the Source Spoke to switch to repository without the kernel-redhat, then the Kernel spoke should be skipped but it's showed empty

pyanaconda/ui/tui/spokes/software_selection.py
```diff
@@ -93,6 +100,11 @@ def _initialize(self):
         """Initialize the spoke in a separate thread."""
         threadMgr.wait(THREAD_PAYLOAD)
 
+        self.available_kernels = get_display_kernel_features(self.payload)
```
poncovka, line 103: In general, it is good to keep the implementation of GUI and TUI as close and consistent as possible, because usually we change both at the same time. Here TUI uses `available_kernels`, but GUI uses `show_kernel_features` for the same data. Ideally, these attributes should have the same name.

poncovka (commented):
I have some recommendations.

pyanaconda/ui/tui/spokes/software_selection.py
```diff
@@ -93,6 +100,11 @@ def _initialize(self):
...
+        self.kernel_selection = dict.fromkeys(self.available_kernels, False)
+        # kernel-redhat should be the default when available
+        self.kernel_selection["upstream"] = True
```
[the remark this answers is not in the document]
OndrejZobal: Oh, that was really bad, I am sorry this got through, thanks for catching this.

pyanaconda/ui/lib/software.py
```diff
@@ -0,0 +1,58 @@
...
+from blivet.arch import is_arm
+from pyanaconda.core.i18n import _
+
+KernelFeatures = namedtuple("KernelFeatures", "upstream page_size_64k")
+
+def get_display_kernel_features(payload):
+    """Returns a dictionary with that shows which kernels should be shown in the UI.
+    """
+    features = {
+        "upstream": len(payload.match_available_packages("kernel-redhat")) != 0,
```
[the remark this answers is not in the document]
OndrejZobal: Thanks!

pyanaconda/ui/tui/spokes/software_selection.py
```diff
@@ -260,32 +299,46 @@ def _set_addons_callback(self, data):
...
         else:
             if key.lower() == Prompt.CONTINUE:
-
                 # No environment was selected, close
                 if not self._selection.environment:
                     self.close()
 
                 # The environment was selected, switch screen
-                elif args is None:
+                elif args is None and not self.shown_environment:
```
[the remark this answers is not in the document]
OndrejZobal: So, should I convert the addons screen to a spoke as well?

OndrejZobal:
All of the previous comments should be addressed in my last force push

pyanaconda/ui/tui/spokes/software_selection.py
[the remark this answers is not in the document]
jkonecny12: This seems to be still true. In this commit nor the previous ones I don't see any "upstream" kernel implementation. Also it should be renamed because it's not called upstream anywhere else.

pyanaconda/ui/tui/spokes/software_selection.py
```diff
@@ -345,3 +350,123 @@ def _check_software_selection(self):
...
+                ScreenHandler.push_screen_modal(spoke)
+            self.apply()
+            self.execute()
+            self.close()
+            return InputState.PROCESSED
+
+        return super().input(args, key)
+
+    def apply(self):
+        pass
```
jkonecny12, line 424: I wonder, do we need to call `self.apply()` when `def apply(self):` is empty?

pyanaconda/ui/lib/software.py
```diff
@@ -19,13 +19,16 @@
 from blivet.arch import is_arm
 from pyanaconda.core.i18n import _
 
+FEATURE_UPSTREAM = "upstream"
 FEATURE_64K = "64k"
-KernelFeatures = namedtuple("KernelFeatures", "page_size_64k")
+
+KernelFeatures = namedtuple("KernelFeatures", "upstream page_size_64k")
```
jkonecny12, line 25: This doesn't look correct. You are defining the `namedtuple` with one string value but then instantiating it with two values. What you probably want is:

```python
KernelFeatures = namedtuple("KernelFeatures", ["upstream", "page_size_64k"])
```

pyanaconda/ui/lib/software.py
```diff
@@ -0,0 +1,55 @@
...
+KernelFeatures = namedtuple("KernelFeatures", "page_size_64k")
+
+def get_display_kernel_features(payload):
+    """Returns a dictionary with that shows which kernels should be shown in the UI.
+    """
+    features = {
+        FEATURE_64K: is_arm() and any(payload.match_available_packages("kernel-64k"))
+    }
+
+    return features
```
jkonecny12: This name is a bit cryptic to me. Wouldn't it be better to use name what you have when you storing this value `get_available_kernel_features`?

jkonecny12 (changes requested):
Hi, thanks for these improvements. However, I think we should still a bit more polished the code.

Please feel free to write me when there is anything or we can work together on the code if you would prefer.

pyanaconda/ui/lib/software.py
OndrejZobal: Those two statements should be equivalent, see: https://realpython.com/python-namedtuple/#providing-required-arguments-to-namedtuple

pyanaconda/ui/gui/spokes/software_selection.py
```diff
@@ -194,6 +212,16 @@ def apply(self):
         for group_name in self._get_selected_addons():
             self._selection.groups.append(group_name)
 
+        # Select kernel
+        property_64k = self._available_kernels[FEATURE_64K] and \
+            self._combo_kernel_page_size.get_active_id() == FEATURE_64K
+        kernel_properties = KernelFeatures(property_64k)
```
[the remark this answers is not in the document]
OndrejZobal: I agree that this would be much better.

pyanaconda/ui/tui/spokes/software_selection.py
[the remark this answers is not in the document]
poncovka: That would be probably too much work.

pyanaconda/ui/gui/spokes/software_selection.py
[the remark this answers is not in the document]
OndrejZobal: Actually I am not sure if I fully understand it...
So, `get_kernel_property` translates a list of features in `kwargs` into a `KernelFeature` object which represents a specific Kernel, so maybe it should be called `KernelInfo` or just `Kernel`. I am not sure what do you mean by `id` is it a tuple containing the combination of requested features? 

This is what I think you meant and I think it's a better design than what I had. Storing UI text along with package name in a signle object is a great idea!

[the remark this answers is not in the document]
OndrejZobal: Or wait, did you mean for PackageFeature truly be just a representation of a feature and the package_name would contain the suffix (ex. "-64k" or "-redhat") that would be used to build the package name?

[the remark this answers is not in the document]
jkonecny12: Good point, it complicates things... In that case feel free to have it as it is right now.

pyanaconda/ui/lib/software.py
jkonecny12: Ha, interesting. I wasn't aware of that :D

jkonecny12 (commented):
You had a great reactions on the biggest points, so I will just test the functionality and set an ACK if it works.
Other pending issues are not that blocking. Even though the "upstream" is used too early is something what would be great to solve.

pyanaconda/ui/lib/software.py
OndrejZobal: I changed the argument to list to fit with the rest of the codebase

OndrejZobal:
I completed another iteration!

jkonecny12 (approved):
Tested and reviewed, all seems to be working great. Thanks for the work!

PR #4690, opened 2023-04-11
exception: only attach existent and non-empty files (#2185827)

AdamWill:
libreport barfs (in a non-fatal but scary way) if a file we try to attach does not exist or is empty. Let's make sure all the files exist and aren't empty before we try to attach them. In a current F38+ live install, for instance, both packaging.log and dnf.librepo.log are usually empty.

### DIFF
--- a/pyanaconda/exception.py
+++ b/pyanaconda/exception.py
@@ -138,6 +138,10 @@ class AnacondaExceptionHandler(ExceptionHandler):
         """
 
         log.debug("running handleException")
+        # don't try and attach empty or non-existent files (#2185827)
+        self.conf.fileList = [
+            fn for fn in self.conf.fileList if os.path.exists(fn) and os.path.getsize(fn) > 0
+        ]
         exception_lines = traceback.format_exception(*dump_info.exc_info)
         log.critical("\n".join(exception_lines))
 


AdamWill:
There are a few mysteries here (like why we get an error for packaging.log but not dnf.librepo.log, and why this just now started happening). But I think the change is safe.

VladimirSlavik (approved):
Looks good to me. Thank you!

pyanaconda/exception.py
```diff
@@ -138,6 +138,10 @@ def handleException(self, dump_info):
         """
 
         log.debug("running handleException")
+        # don't try and attach empty or non-existent files (#2185827)
+        self.conf.fileList = [
+            fn for fn in self.conf.fileList if os.path.exists(fn) and os.path.getsize(fn) > 0
```
kparal, line 143: `os.path.isfile(fn)` might be even a bit safer than `.exists()`. But that's just a nitpick.

VladimirSlavik:
@AdamWill, does this need porting to f38? The bug is reported against that...

AdamWill:
Well, theoretically yes, it's to fix an F38 blocker. But I [backported it as a downstream patch for f38](https://src.fedoraproject.org/rpms/anaconda/c/e2097295547ef395eddca808b6d1edeee98922c9?branch=f38) as we were on a tight timeframe, and we've now signed off the release, so you might consider it less important to do so at least immediately. We do need the fix to be in any future anaconda builds for F38, though, whether that's by keeping the patch downstream or merging it to the f38 branch upstream.

VladimirSlavik:
@M4rtinK, @jkonecny12, @rvykydal - we probably want to backport that, right?

jkonecny12:
Removing port to Fedora label because F38 is released already, we can't do the backport now anyway.

PR #4693, opened 2023-04-12
WebUI: avoid conflicting import with python-installer

jelly:
When python-installer is installed, the tests fail with:

ImportError: cannot import name 'Installer' from 'installer' (/usr/lib/python3.10/site-packages/installer/__init__.py)

I suppose ideally everything would use helpers for imports.

### DIFF
--- a/ui/webui/test/check-language
+++ b/ui/webui/test/check-language
@@ -15,23 +15,15 @@
 # You should have received a copy of the GNU Lesser General Public License
 # along with this program; If not, see <http://www.gnu.org/licenses/>.
 
-import os
-import sys
-
-# import Cockpit's machinery for test VMs and its browser test API
-TEST_DIR = os.path.dirname(__file__)
-sys.path.append(os.path.join(TEST_DIR, "common"))
-sys.path.append(os.path.join(TEST_DIR, "helpers"))
-sys.path.append(os.path.join(os.path.dirname(TEST_DIR), "bots/machine"))
+import anacondalib
 
 from installer import Installer
 from language import Language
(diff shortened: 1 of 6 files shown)


ui/webui/test/check-language
```diff
@@ -24,7 +24,7 @@ sys.path.append(os.path.join(TEST_DIR, "common"))
 sys.path.append(os.path.join(TEST_DIR, "helpers"))
 sys.path.append(os.path.join(os.path.dirname(TEST_DIR), "bots/machine"))
 
-from installer import Installer
+from helpers.installer import Installer
```
poncovka, line 27: Btw. `helpers` don't look like a safe namespace either. I would consider a truly unique namespace or relative imports. Or maybe just prepend these Python paths instead of appending them. Otherwise, this will happen again.

VladimirSlavik (approved):
Thank you!

ui/webui/test/check-language
jelly: Good point, let's fix it for real.

jelly: I've done some more general re-factoring by moving all the common imports to `anacondalib` (I couldn't make up a better name...) and introduced a new MachineCase subclass which avoids other code duplication.

As a side-effect I now prepend `helpers` so having `python-installer` installed is no longer an issue.

jelly:
> Thank you! I guess this actually creates something similar to cockpit's parent.py or whatever is that called?

Yes, this is a similar approach as `parent.py` from cockpit.

M4rtinK (approved):

PR #4695, opened 2023-04-12
Trim whitespaces from /kickstart-tests GH comment (#infra)

jkonecny12:
Right now the tests will fail in weird way when the comment on ends up with \n. Let's fix that by trimming not only start of the command but also end of it.

Should avoid issue like in https://github.com/rhinstaller/anaconda/actions/runs/4677570787/jobs/8285239264.

### DIFF
--- a/.github/workflows/kickstart-tests.yml
+++ b/.github/workflows/kickstart-tests.yml
@@ -53,7 +53,7 @@ jobs:
           BODY: ${{ github.event.comment.body }}
         run: |
           # extract first line and cut out the "/kickstart-tests" first word
-          LAUNCH_ARGS=$(echo "$BODY" | sed -n '1 s/^[^ ]* *//p')
+          LAUNCH_ARGS=$(echo "$BODY" | sed -n '1 s/^[^ ]* *//p' | sed 's/[[:space:]]*$//')
           echo "launch arguments are: $LAUNCH_ARGS"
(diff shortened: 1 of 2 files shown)


jkonecny12:
Comment is trimmed in the GitHub UI (here) but you are getting a raw form into the GH Action with a lot of new lines and whitespaces at the end.

jkonecny12:
However, the trimming can't be tested here because workflow file is taken from Rawhide branch. Failure is expected :(.

If everyone is fine with that I would go with "tested locally in bash" and merge it?

VladimirSlavik (commented):
Thank you!

Personally I'm fine with merging and fixing again if need be.

jkonecny12:
Add also part for waive - the same issue and tested here: https://github.com/jkonecny12/anaconda/actions/runs/4690604718/jobs/8313927024

Tested the kickstart test run here (failure expected):
https://github.com/jkonecny12/anaconda/actions/runs/4690633873/jobs/8313999477

The kickstart test run failed (not having self-hosted runners) but I'm able to verify by looking on the test output in run job. On [failed tests](https://github.com/rhinstaller/anaconda/actions/runs/4677570787/jobs/8285239264) you can see step with name "Generate query arguments for ..." which is missing ending quotes. This step is however named correctly in the run linked above.

jkonecny12:
/kickstart-test --waive not able to test (workflow is not from PR)

VladimirSlavik (approved):
Thanks!

PR #4698, opened 2023-04-13
docs: Document the distribution component

poncovka:
Bugs related to changes in package groups and environments should be reassigned to the `distribution` component instead of the `comps` component.

### DIFF
--- a/docs/common-bugs.rst
+++ b/docs/common-bugs.rst
@@ -118,7 +118,8 @@ Changes in package groups and environments
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 :Issue: The reporter wants a new package to be installed by default.
-:Solution: Reassigning to comps.
+:Solution: The package groups and environments are defined by comps (see
+    https://pagure.io/fedora-comps/). Reassigning to distribution.
 :Example: `rhbz#1787018 <https://bugzilla.redhat.com/show_bug.cgi?id=1787018>`_
 
 Corrupted ISO


docs/common-bugs.rst
```diff
@@ -118,7 +118,7 @@ Changes in package groups and environments
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 :Issue: The reporter wants a new package to be installed by default.
-:Solution: Reassigning to comps.
+:Solution: Reassigning to comps or distribution.
```
jstodola, line 121: The commit summary says to use "distribution instead of comps", but the change suggests to use "comps or distribution". I (as a lazy user) would pick the first listed component.
In addition to that, https://fedoraproject.org/wiki/How_to_use_and_edit_comps.xml_for_package_groups#All_other_groups says to use the comps component - it looks like the current text is correct.

jstodola (changes requested):

docs/common-bugs.rst
poncovka: The `comps` component doesn't seem to exist in Bugzilla anymore and the bug, that we needed to recently reassign, ended up in the `distribution` component. There is only the `fedora-comps` repository at https://pagure.io/fedora-comps with their own issue tracker. It seemed better to keep there all variants in case that this will change again in the future, but I can change it.

jstodola: Since the `comps` component cannot be selected when creating a new bug, it should not be mentioned in the documentation - we do not want to confuse the users.
I've asked for clarification and updating the "How to use and edit comps.xml for package groups" wiki page:
https://lists.fedoraproject.org/archives/list/packaging@lists.fedoraproject.org/thread/KTZW557TLXXNEDSEBJFR77APHUSDIYOT/

jkonecny12 (approved):
Looks great to me. Thanks!

PR #4703, opened 2023-04-17
Fix virt-install cockpit run on fedora-X images

jkonecny12:
The issue is that your system (the one running virt-install) needs to known the fedora you want to test.

Example:
My system is Fedora 37
I want to run fedora-38 test
\=
virt-install doesn't know fedora-38 because it's fedora-37 virt-install

So instead of giving them the link let's virt-install be smart and try to detect the os variant.

(cherry picked from commit 9487223ce2f9afb6b7e823efa3274a9288d1ba7b)


VladimirSlavik (approved):

M4rtinK (approved):
Looks good to me as well. :)

PR #4705, opened 2023-04-17
Workaround to packit srpm builds npm issue (#infra)

jkonecny12:
Right now srpm packit build failing because of missing npm tool.

Fedora should be standard install nodejs-npm but in COPR for some reason it installs nodejs20-npm. That seems to provoke an error during build that npm tool is missing.

Require nodejs-npm directly as solution for now.


VladimirSlavik (approved):
Thanks!

PR #4707, opened 2023-04-18
.github: add Cockpit library update action

jelly:
This action works similar to the npm-update workflow, it creates a PR every Thursday (usually after the Cockpit release) to update the COCKPIT_REPO_COMMIT to the latest version.

---

Tested this on my fork:

https://github.com/jelly/anaconda/actions/runs/4732456425/jobs/8398650535

Which creates the following PR

https://github.com/jelly/anaconda/pull/2

### DIFF
--- a/.github/workflows/cockpit-lib-update.yml
+++ b/.github/workflows/cockpit-lib-update.yml
@@ -1,0 +1,32 @@
+name: cockpit-lib-update
+on:
+  # Runs every Thursday at 2 AM
+  schedule:
+    - cron: '0 2 * * 4'
+  # can be run manually on https://github.com/rhinstaller/anaconda/actions
+  workflow_dispatch:
+jobs:
+  cockpit-lib-update:
+    permissions:
+      contents: write
+      pull-requests: write
+    runs-on: ubuntu-latest
+    steps:
+      - name: Set up dependencies
+        run: |
+          sudo apt update
+          sudo apt install -y make
+
+      - name: Set up configuration and secrets
+        run: |
+          printf '[user]\n\tname = github-actions\n\temail=github-actions@github.com\n' > ~/.gitconfig
+          echo '${{ secrets.GITHUB_TOKEN }}' > ~/.config/github-token
+
+      - name: Clone repository
+        uses: actions/checkout@v3
+
+      - name: Run cockpit-lib-update
+        run: |
+          make -f Makefile.am bots # HACK: to skip running autogen.sh/configure
+          bots/cockpit-lib-update Makefile.am
+        working-directory: ui/webui
--- a/docs/ci-status.rst
+++ b/docs/ci-status.rst
@@ -37,6 +37,10 @@ Anaconda
    :alt: Run unit and RPM tests daily
    :target: https://github.com/rhinstaller/anaconda/actions/workflows/tests-daily.yml
 
+.. |cockpit-lib-update| image:: https://github.com/rhinstaller/anaconda/actions/workflows/cockpit-lib-update.yml/badge.svg
+   :alt: Updates Cockpit library
+   :target: https://github.com/rhinstaller/anaconda/actions/workflows/cockpit-lib-update.yml
+
 .. _Dependabot: https://github.com/rhinstaller/anaconda/network/updates
 
 |container-autoupdate-fedora|
@@ -57,6 +61,9 @@ Anaconda
 
 |tests-daily|
   Runs unit and RPM tests every day, independent of any changes to code or containers.
+
+|cockpit-lib-update|
+  Updates the COCKPIT_REPO_COMMIT in ui/webui/Makefile.am and opens a pull request.
 
 Dependabot_
   Checks Anaconda dependencies and opens pull requests for new versions.


.github/workflows/cockpit-lib-update.yml
```diff
@@ -0,0 +1,32 @@
+name: cockpit-lib-update
+on:
+  schedule:
+    - cron: '0 2 * * 4'
```
VladimirSlavik, line 4: Maybe please consider adding a comment what is the schedule for those less cron-literate?

.github/workflows/cockpit-lib-update.yml
```diff
@@ -0,0 +1,32 @@
...
+  # can be run manually on https://github.com/cockpit-project/starter-kit/actions
```
VladimirSlavik, line 5: Not the right URL :)

.github/workflows/cockpit-lib-update.yml
```diff
@@ -0,0 +1,32 @@
...
+  workflow_dispatch:
+jobs:
+  cockpit-lib-update:
+    permissions:
+      contents: write
+      pull-requests: write
+      statuses: write
```
VladimirSlavik, line 12: Does this need statuses? I don't see anything setting a status... If it's needed for the github api in script from bots, please add a comment so it makes sense at a glance.

.github/workflows/cockpit-lib-update.yml
```diff
@@ -0,0 +1,32 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Set up dependencies
+        run: |
+          sudo apt update
+          sudo apt install -y make
+
+      - name: Set up configuration and secrets
+        run: |
+          printf '[user]\n\tname = Cockpit Project\n\temail=cockpituous@gmail.com\n' > ~/.gitconfig
```
VladimirSlavik, line 22: We set up identity in other actions like this:
```
          git config user.name github-actions
          git config user.email github-actions@github.com
```
Alternatively, our bot identity is @InStallker, although we use it very little.

Would that work instead of cockpituous?

VladimirSlavik (commented):
As I understand it, this is a dependabot-alike that tries to bump the cockpit hash and gives us a PR that runs the tests and shows if that is a good idea. Thank you, this would be quite useful!

How would this handle different branches? Or, does this expect that it runs only from the default branch? I never remember what's the github rules for this :-(

VladimirSlavik:
A few more thoughts, not directly affecting PR acceptance.
- There's a page `docs/ci-status.rst` that gives us a crude "control center" with the red/green badges for scheduled jobs. Would be nice to have it there.
- The scheduled workflows are usually disabled by templating on branches outside master (rawhide).

VladimirSlavik:
The infra commit message fails because we expect the `infra:` prefix or `(#infra)` postfix.

.github/workflows/cockpit-lib-update.yml
jelly: Ah, it probably doesn't. So we can drop that copy-pasta.

.github/workflows/cockpit-lib-update.yml
jelly: Sure! This is just the default Cockpit uses.

jelly:
> As I understand it, this is a dependabot-alike that tries to bump the cockpit hash and gives us a PR that runs the tests and shows if that is a good idea. Thank you, this would be quite useful!

That's it!
 
> How would this handle different branches? Or, does this expect that it runs only from the default branch? I never remember what's the github rules for this :-(

It does not handle different branches, I never thought about that. I guess the checkout action could checkout a different branch but I am not sure if the Cockpit code allows for making a pr to a different master/main.

VladimirSlavik:
IMO leaving stable branches as stable should not be too much of a problem, so running just on master sounds fine to me.

M4rtinK (approved):
Looks good to me & IMHO a nice improvement to the current situation. :)

I'm always a bit concerned when we pin something to a concrete version - even though it makes stuff more stable and dependable in the short term, it will bite us later if we forget about it. Having an automated action for this could certainly help, while keeping most of the benefits. :)

docs/ci-status.rst
```diff
@@ -35,6 +35,8 @@ Anaconda
 
 .. _Dependabot: https://github.com/rhinstaller/anaconda/network/updates
 
+.. _CockpitLibUpdate: https://github.com/rhinstaller/anaconda/actions/workflows/cockpit-lib-update.yml
+
```
VladimirSlavik, line 39: You want to do it like rest above, the Dependabot bit is the only different one because there's no standard status for its config parsing.

VladimirSlavik: Ouch, that was probably too little information. Gotta love ReST... You need two blocks of code for each of the badges on the resulting page. First, which you already got now, is a definition of the image. The second part (below) is where you actually use the image in the doc/page. Sorry about being unclear :(

jelly: Woops :) Fixed it

VladimirSlavik (approved):
Thank you!

VladimirSlavik:
@jelly is this blocked by anything elsewhere?

jelly:
> @jelly is this blocked by anything elsewhere?

Not from my side, I'll rebase it.

VladimirSlavik:
Yes, that's harmless. The purpose is to prevent changes that would affect the actual application code. It's an imperfect and imprecise thing, by definition. Basically there's a list of things that are guaranteed to be always only infra, and then the rest needs human intelligence to decide, so it is reported with the comment.

This particular file is technically documentation, but it should be part of the infra list, because it is related to that. I have a PR for this change.

PR #4708, opened 2023-04-19
WebUI: run eslint in CI

jelly:


jelly:
This should fail...

jelly:
```
[jelle@t14s][~/projects/anaconda/ui/webui]%npm run eslint

> eslint
> eslint --ext .js --ext .jsx src/


/home/jelle/projects/anaconda/ui/webui/src/components/AnacondaWizard.jsx
  168:60  error  Expected '===' and instead saw '=='  eqeqeq

✖ 1 problem (1 error, 0 warnings)

[jelle@t14s][~/projects/anaconda/ui/webui]%
```

jelly:
It did fail here too in CI https://cockpit-logs.us-east-1.linodeobjects.com/pull-4708-20230419-151654-99c36147-fedora-rawhide-boot/log.html

So we never linted in CI.

VladimirSlavik:
Eslint runs as one of the anaconda unit tests already, so it just ran elsewhere. However, it's super heavy and takes forever. What would be the "cost" of this? It might be the preferable option...

jelly:
> Eslint runs as one of the anaconda unit tests already, so it just ran elsewhere. However, it's super heavy and takes forever. What would be the "cost" of this? It might be the preferable option...

Oh, why didn't it catch the eslint error then?

The PR which introduced it was https://github.com/rhinstaller/anaconda/pull/4628 

The overhead is quite minimal as it's linted as part of the build, example.

```
[jelle@t14s][~/projects/anaconda/ui/webui]%time npm run eslint

> eslint
> eslint --ext .js --ext .jsx src/


/home/jelle/projects/anaconda/ui/webui/src/components/AnacondaWizard.jsx
  168:60  error  Expected '===' and instead saw '=='  eqeqeq

✖ 1 problem (1 error, 0 warnings)

npm run eslint  3.68s user 0.21s system 141% cpu 2.757 total
```

VladimirSlavik:
That'đ a great question, and I'm not entirely sure, actually.

The file is here: https://github.com/rhinstaller/anaconda/blob/master/tests/webui_eslint/run_webui_eslint.sh and it runs eslint as `npx eslint ...`, no longer certain why exactly. But if I had to guess, probably because it's suggested [here](https://eslint.org/docs/latest/use/getting-started) and because the default output format looked like crap in logs.

When I run it as you do, `npm run eslint`, it does discover the error, and dies on any arguments other than the directory.

VladimirSlavik:
Ah, now that I know what to search for, I found also how to run it with the options...

jelly:
Ok, I can take a look why that uni test is either not triggered or does not fail.

jelly:
Fails now https://github.com/rhinstaller/anaconda/actions/runs/4756440056/jobs/8451985594?pr=4708 now re-pushed with the fix so the PR should be green :)

VladimirSlavik:
Thank you! LGTM, though I can't speak for those working on WebUI...

tests/webui_eslint/run_webui_eslint.sh
```diff
@@ -27,5 +27,5 @@ npm install
 echo
 
 echo "Linting:"
-npx eslint --format stylish --no-color src/
+npx eslint --format stylish --no-color --ext .js --ext .jsx src/
```
KKoukiou, line 30: Maybe just run the script from package.json? 

`npm run eslint`

KKoukiou (approved):

tests/webui_eslint/run_webui_eslint.sh
VladimirSlavik: I think that would become `npm run eslint src/ -- --format stylish --no-color` then...?

Maybe we want to drop one of these two :)

VladimirSlavik (approved):
Oops, sorry, forgot to ack :)

jkonecny12 (approved):

PR #4710, opened 2023-04-24
Update review screen for v1 of autopart

rvykydal:
This is a follow-up task [INSTALLER-3445](https://issues.redhat.com/browse/INSTALLER-3445) of https://github.com/rhinstaller/anaconda/pull/4599.
It updates the review screen to include information about the selected autopartitioning scenario and encryption choice. Also the list of disks used is expandable now. The warnings, button labels and warning dialog for the newly added autopartioning option (Use free space) was added.

Notes and issues needing UX expert feedback / suggestions:

- The list of selected disk is automatically expanded in case there is only a single disk.
- The on-screen warning is there also in case of "Use free space" scenario. I think advice to backup is valid even in this case.
- The button label and warning dialog microcopy for "Use free space..." can be perhaps improved.
- The button label ("Erase disks and install") and scenario description ("Erase devices and install" which is the same as the Storage Configuration screen option label) should be the unified. (Moreover there is "Erase data and install" in play suggested in microcopy task [INSTALLER-3406](https://issues.redhat.com/browse/INSTALLER-3406) but I'd prefer to address the "Erase data" suggestion separately not to block this PR)

TODO:
- [x] tests for review screen

Screenshots:

For "Erase devices and install" option:
https://rvykydal.fedorapeople.org/webui/autopart_pr1/v1/screenshots/review_screen/erase_all/
erase_all4
erase_all5

For "Use free space for the installation" option:
https://rvykydal.fedorapeople.org/webui/autopart_pr1/v1/screenshots/review_screen/use_free_space/
use_free_space3
use_free_space4

For multiple disks:
https://rvykydal.fedorapeople.org/webui/autopart_pr1/v1/screenshots/review_screen/multiple_disks/
multiple_disks4
multiple_disks5

- [implementation] For modification of button labels and warning dialogs we need to keep the choice of autopart scenario in the AnacondaWizard component (`storageScenarioId`). So it could make sense not to update the scenario in the backend right on the Storage Configureation screen (`setInitializationMode`) but instead keep the value in the `storageScenarioId` component state and update the backend only when the whole partitioning is applied (at the end of the Installation destination steps - on Next from the Disk Encryption step). We do something similar with disk encryption options (keeping state in AnacondaWizard `storageEncryption`) but in that case the main reason is we are not able to keep these settings in the backend.

### DIFF
--- a/ui/webui/src/components/review/ReviewConfiguration.jsx
+++ b/ui/webui/src/components/review/ReviewConfiguration.jsx
@@ -25,12 +25,17 @@ import {
     DataListCell,
     DescriptionList, DescriptionListGroup,
     DescriptionListTerm, DescriptionListDescription,
+    ExpandableSection,
     Modal, ModalVariant,
     Alert
 } from "@patternfly/react-core";
 
 import {
-    getSelectedDisks, getDeviceData,
+    getSelectedDisks,
+    getDeviceData,
+    getAppliedPartitioning,
+    getPartitioningRequest,
+    getInitializationMode,
 } from "../../apis/storage.js";
 
 import {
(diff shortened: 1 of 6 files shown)


ui/webui/src/components/review/ReviewConfiguration.jsx
```diff
@@ -25,26 +25,55 @@ import {
...
+import { scenarioForInitializationMode, getScenario } from "../storage/StorageConfiguration.jsx";
+
 const _ = cockpit.gettext;
 
-export const ReviewConfiguration = ({ idPrefix }) => {
+export const ReviewDescriptionList = ({ children }) => {
+    return (
+        <DescriptionList
+          isHorizontal
+          horizontalTermWidthModifier={{
```
M4rtinK, line 54: I wonder if these should maybe somehow got to the CSS file if possible ?

rvykydal: Why do you think it should go to a CSS file? Seems to me like customization of a component by setting its properties. Following this example: https://www.patternfly.org/v4/components/description-list/#horizontal-using-custom-term-width-modifier.

rvykydal:
Thank you for spotting the regression @lclay2. It was caused by a typo in an object attribute name so the title was not applied. I updated the PR to contain the titles.

Still I agree that we can think about easing / removing the warning for "Use free space" option although given we warn suggesting backup on the screen, keeping the warning also here is not so bad.

Screenshots of updated dialogs:
Screenshot from 2023-04-28 13-52-14
Screenshot from 2023-04-28 13-52-54

rvykydal:
Updated with change of Back buttons from "secondary" variant to "link"
https://rvykydal.fedorapeople.org/webui/autopart_pr1/v1/screenshots/review_screen/warning_dialog_back_link/
Screenshot from 2023-04-28 15-22-10
Screenshot from 2023-04-28 15-22-59

jkonecny12 (approved):
I really like this new version. Thanks!

rvykydal:
Based on review by @lclay2 and Allison I updated the "Use free space dialog" to be less scary and more assuring.
use_free_no_warning

lclay2:
Thanks for the dialog updates @rvykydal , and for catching the typo! LGTM :)

rvykydal:
Updated tests.

lclay2 (approved):
Thanks for making the dialog updates! LGTM :)

PR #4711, opened 2023-04-24
Increase the minimum EFI System Partition (ESP) size to 500MiB

hughsie:
Currently the default allocated space for the EFI System Partition (ESP) is a minimum of 200 MiB and a maximum of 600MiB. This smaller size is more than enough for storing the EFI binaries just needed to boot the machine.

The ESP is also used to deploy firmware updates which need 2x SPI flash size (typically 64MB, soon 128MB) as free space. We're also soon going to modernize how the bootloader is deployed, and this also is going to need more space again.

Microsoft mandates OEMs to allocate a minimum of 500MiB and so I think we should just do the same.

### DIFF
--- a/pyanaconda/modules/storage/platform.py
+++ b/pyanaconda/modules/storage/platform.py
@@ -224,8 +224,8 @@ class EFI(Platform):
         return PartSpec(
             mountpoint="/boot/efi",
             fstype="efi",
-            size=Size("200MiB"),
-            max_size=Size("600MiB"),
+            size=Size("500MiB"),
+            max_size=Size("2GiB"),
             grow=True
         )
 
--- a/docs/release-notes/bigger-esp.rst
+++ b/docs/release-notes/bigger-esp.rst
@@ -1,0 +1,14 @@
+:Type: Bootloader
+:Summary: Make the EFI System Partition at least 500MiB in size
+
+:Description:
+    The size of the EFI System Partition (ESP) created by Anaconda has changed from 200 MiB to 500 MiB.
+
+    The reasons for this change include:
+    - This partition is used to deploy firmware updates. These updates need free space of twice the SPI flash size, which will grow from 64 to 128 MiB in near future and make the current partition size too small.
+    - The larger space enables future changes to how the bootloader is deployed on Fedora.
+    - The new minimum is identical with what Microsoft mandates OEMs allocate for the partition.
+
+:Links:
+    - https://fedoraproject.org/wiki/Changes/BiggerESP
+    - https://github.com/rhinstaller/anaconda/pull/4711
--- a/tests/unit_tests/pyanaconda_tests/modules/storage/test_platform.py
+++ b/tests/unit_tests/pyanaconda_tests/modules/storage/test_platform.py
@@ -181,7 +181,7 @@ class PlatformTestCase(unittest.TestCase):
 
         self._check_partitions(
             PartSpec(mountpoint="/boot/efi", fstype="efi", grow=True,
-                     size=Size("200MiB"), max_size=Size("600MiB")),
+                     size=Size("500MiB"), max_size=Size("2GiB")),
             PartSpec(mountpoint="/boot", size=Size("1GiB"))
         )
 
@@ -256,7 +256,7 @@ class PlatformTestCase(unittest.TestCase):
 
         self._check_partitions(
             PartSpec(mountpoint="/boot/efi", fstype="efi", grow=True,
-                     size=Size("200MiB"), max_size=Size("600MiB")),
+                     size=Size("500MiB"), max_size=Size("2GiB")),
             PartSpec(mountpoint="/boot", size=Size("1GiB"))
         )
 
@@ -293,7 +293,7 @@ class PlatformTestCase(unittest.TestCase):
 
         self._check_partitions(
             PartSpec(mountpoint="/boot/efi", fstype="efi", grow=True,
-                     size=Size("200MiB"), max_size=Size("600MiB")),
+                     size=Size("500MiB"), max_size=Size("2GiB")),
             PartSpec(mountpoint="/boot", size=Size("1GiB"))
         )
 


VladimirSlavik:
```
=========================== short test summary info ============================
FAILED unit_tests/pyanaconda_tests/modules/storage/test_platform.py::PlatformTestCase::test_aarch64_efi
FAILED unit_tests/pyanaconda_tests/modules/storage/test_platform.py::PlatformTestCase::test_arm_efi
FAILED unit_tests/pyanaconda_tests/modules/storage/test_platform.py::PlatformTestCase::test_efi
```
These tests all expect the size to be 200, like `size=Size("200MiB")` - you'll want to increase that to 500.

hughsie:
> These unit tests all expect the size to be 200

Aha, I wasn't sure if those were defining the size or testing. Fixed, thanks.

VladimirSlavik:
I just let the tests run again, but... I'll note that you bumped the default for normal platforms and the 3 tests for that, and also the one mac_efi test but not the actual value for that platform... ;-)

But we're getting there! Thank you!

VladimirSlavik:
I can write a release note later, if you want to skip that part...  If not, see docs/release-notes/template.rst

hughsie:
> I just let the tests run again, but... I'll note that you bumped the default for normal platforms and the 3 tests for that, and also the one mac_efi test but not the actual value for that platform... ;-)

Do we also want the `macefi` size to be 500MB? There is not going to be firmware updates for that hardware, but I think it makes sense to be consistent.

hughsie:
Ack, okay. Fixed.

pyanaconda/modules/storage/platform.py
```diff
@@ -224,7 +224,7 @@ def _bootloader_partition(self):
         return PartSpec(
             mountpoint="/boot/efi",
             fstype="efi",
-            size=Size("200MiB"),
+            size=Size("500MiB"),
             max_size=Size("600MiB"),
```
berrange, line 228: With the default bumped all the way to 500,  keeping 600 for the maximum is not really facilitating much enlargement. IOW should max_size also be increased, perhaps to 1GiB ?

poettering:
I would prefer if this would grow even larger. We already know that gfx driver people want to include their drivers in initrds in a future no so far away, and that means 100mb+ in an initrd. if you stick to 500M this is going to require a bumping very soon again I am sure. And it's really hard retroactviely growing the ESP.

(see here for a discussion about that: https://mastodon.social/@karolherbst@chaos.social/110066943704825190)

I'd really be more aggressive here. i.e. on a large disk go for 2g at least, and on the tiniest of systems I'd still go for 800m or so. (if people ant to go smaller, allow them via some form of override, sure)

poettering:
(and it's not just that some drivers will grow to 100M+ but that we want to go for a future with pre-build kernels, and that likely means shipping multiple of them in the ESP at the same time, even on systems that don't need them.)

poettering:
This well-respected page from 2015 suggests 550M as a minimum btw: https://www.rodsbooks.com/linux-uefi/ 

ArchLinux recommends 512M minimum https://wiki.archlinux.org/title/EFI_system_partition (and that's a pretty old recommendation too)

debian also defaults to 500M since a longer time.

I think in 2023 we shouldn't really default to a default that 8 years ago seemed appropriate, in particular with the clear perspective pre-built UKIs and graphics drivers in the initrd will boost our disk space needs in future.

pyanaconda/modules/storage/platform.py
hughsie: Yes, agreed, I've modified the patch to allow scaling up to 2GiB now.

hughsie:
> I would prefer if this would grow even larger.

I think a lot of people working on UKIs would agree, and that can certainly be a future change -- but this hopefully non-controversial change is about aligning the value as used by Microsoft and to fix future firmware updates. Your suggested 2GB is going to be a significant percentage of space on IoT devices and needs much more careful consideration -- for instance deleting the /boot and expanding /boot/efi when we do the switch to UKIs, or maybe defaulting to the larger "maximum" value when the target disk is much larger. That's deliberately not covered in this proposal, although I have changed the maximum size to 2GiB to be helpful.

hughsie:
> The release note text could use some more work, I'd suggest something like this:

I've modified it to your text, thanks.

poettering:
> > I would prefer if this would grow even larger.
> 
> I think a lot of people working on UKIs would agree, and that can certainly be a future change -- but this hopefully non-controversial change is about aligning the value as used by Microsoft and to fix future firmware updates. Your suggested 2GB is going to be a significant percentage of space on IoT devices and needs much more careful consideration -- for instance deleting the /boot and expanding /boot/efi when we do the switch to UKIs, or maybe defaulting to the larger "maximum" value when the target disk is much larger. That's deliberately not covered in this proposal, although I have changed the maximum size to 2GiB to be helpful.

Well so far the only controversy your proposal triggered was that people (such as me) saying you are not bold enough, and go higher. I think the noise in the other direction was rather muted, no?

hughsie:
> saying you are not bold enough

You're completely free to submit another PR and Fedora Change Proposal which supersedes this one. I'm changing the default by 300MB for two easy to justify reasons, you're proposing changing it by 1800 for a different reason entirely.

VladimirSlavik:
If there's a Fedora Change associated with this, I'd love to see it mentioned in the commit message as well as release note links :-)

hughsie:
It's here: https://fedoraproject.org/wiki/Changes/BiggerESP

Maybe don't link to it just yet as we're figuring out if it's a self contained change or a systemwide change.

nl6720:
> ArchLinux recommends 512M minimum https://wiki.archlinux.org/title/EFI_system_partition (and that's a pretty old recommendation too)

Arch recommends 300 MiB which matches what, AFAIK, Windows Setup creates on 4Kn drives. 512 MiB is only recommended for "early and/or buggy UEFI implementations". And for systems with that will have multiple kernel packages installed, the Arch recommendation is at least 1 GiB.

jkonecny12:
Hi thanks for this contribution. I see two things we should resolve before merging this. Give it visibility for a few people (I'll ping them) and have the change accepted by FESCO (it is not right now).

Also @hughsie if you can, please add the change link to the Release notes.

Hi @jstodola @frozencemetery, are you fine with this change?

jkonecny12:
Blocked until we have ACKs from everyone.

jkonecny12 (approved):
Looks good to me.

hughsie:
> Also @hughsie if you can, please add the change link to the Release notes.

Done.

frozencemetery:
> Hi @jstodola @frozencemetery, are you fine with this change?

Yes, bootloader folks are fine with it.  Thanks for checking

jstodola (approved):
It looks good to me.

VladimirSlavik:
Hello @hughsie, do you have any idea when this will be revisited, roughly? I don't want to miss that just because the final commitment happens elsewhere than this PR...

hughsie:
This was approved as a fedora feature. I think it's good to go now, actually.

AdamWill:
So, the first compose we got through with this change shows two failures in openQA tests that are probably related to it (and a third, but that one probably just needs adjustments to the test to handle the larger ESP). I didn't dig into them yet as it's the weekend, but if anyone else wants to, feel free. anaconda logs and journal are on the assets tab:

* https://openqa.fedoraproject.org/tests/1971210 (UEFI install of Workstation package set to a 13GiB disk)
* https://openqa.fedoraproject.org/tests/1971142#step/disk_guided_delete_partial/7 (test which takes a 10GiB disk with two 5GiB partitions, deletes one of them, and does a minimal install to the freed-up space)

AdamWill:
I've filed the various issues openQA ran into after this change and marked them as blocking the Change bug tracker:

* https://bugzilla.redhat.com/show_bug.cgi?id=2212121
* https://bugzilla.redhat.com/show_bug.cgi?id=2214329
* https://bugzilla.redhat.com/show_bug.cgi?id=2214342

For now we have worked around the aarch64 image size issue by simply increasing the image sizes, on the assumption that we really do want the ESP on those images to be 2G in size; if we don't, we'll have to make more adjustments somewhere.

I will probably fiddle with the openQA stuff to work around the other two bugs so those tests don't fail every day, but the underlying issues do still need to be looked at.

PR #4712, opened 2023-04-24
WIP: Store dracut kickstart comands in boss

VladimirSlavik:
Mostly trying to see which parts would work. The code probably needs a different layout, maybe a child module, not really sure yet.

### DIFF
--- a/pyanaconda/modules/boss/kickstart_manager/kickstart_manager.py
+++ b/pyanaconda/modules/boss/kickstart_manager/kickstart_manager.py
@@ -35,11 +35,28 @@ class KickstartManager(object):
 
     def __init__(self):
         self._module_observers = []
+        self._direct_observer = None
+        self._direct_observer_name = ""
 
     @property
     def module_observers(self):
         """Get all module observers for kickstart distribution."""
         return self._module_observers
+
+    @property
+    def direct_observer(self):
+        """Get the direct observer for kickstart distribution."""
+        return self._direct_observer
+
+    @property
+    def direct_observer_name(self):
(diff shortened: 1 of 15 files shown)


pyanaconda/modules/boss/kickstart_manager/kickstart_manager.py
```diff
@@ -157,6 +230,18 @@ def _generate_from_modules(self):
...
+    def _generate_from_direct(self):
+        """Generate kickstart from the direct observer.
+
+        :return: kickstart strings
+        """
+        if not self._direct_observer:
+            return ""
+
+        super_instance = super(type(self._direct_observer), self._direct_observer)
+        kickstart = super_instance.generate_kickstart()
```
VladimirSlavik, line 242: This in particular is just wrong, don't do this :-) It doesn't really work.

VladimirSlavik:
This now works, but the actual organization of the code is still unfinished.

jkonecny12 (commented):
Looks good to me so far.

VladimirSlavik:
Still needs additional tests for `KickstartManager._distribute_to_direct()`, and splitting more things from the main commit.

I am not sure if this is fine or if there should be another layer of abstraction to mate the ks manager with both the dbus and direct modules. My plan is to finish the above, completing the approach without this abstraction layer, and try the other form as further changes on top of this.

VladimirSlavik:
The last two commits are a rough outline of how "unified observers" might look in terms of code.

VladimirSlavik:
(Now it also works.)

PR #4719, opened 2023-04-28
localization: allow to turn off geolocation by configuration

rvykydal:
New Localization section with use_geolocation boolean option is added to
Anaconda configuration.

Intended use: On Fedora Workstation (Gnome) the language will be set by
environment initial setup and Anaconda will take over this value.
This option set to False in the Workstation configuration profile would
prevent overriding the value from environment with geolocation.

INSTALLER-3472

### DIFF
--- a/docs/release-notes/lang-configuration-setting.rst
+++ b/docs/release-notes/lang-configuration-setting.rst
@@ -1,0 +1,11 @@
+:Type: Configuration
+:Summary: Allow to turn off geolocation for language selection (#INSTALLER-3472)
+
+:Description:
+New Localization section with use_geolocation option is added to Anaconda
+configuration.
+
+The option allows to turn off geolocation for language selection.
+
+:Links:
+    - https://github.com/rhinstaller/anaconda/pull/4719
--- a/data/anaconda.conf
+++ b/data/anaconda.conf
@@ -334,3 +334,8 @@ eula =
 # If left empty, geolocation does not run.
 #
 geolocation_provider = https://geoip.fedoraproject.org/city
+
+[Localization]
+# Should geolocation be used when setting the language ?
+#
+use_geolocation = True
--- a/pyanaconda/core/configuration/anaconda.py
+++ b/pyanaconda/core/configuration/anaconda.py
@@ -34,6 +34,7 @@ from pyanaconda.core.configuration.base import Section, Configuration, Configura
 from pyanaconda.core.configuration.profile import ProfileLoader
 from pyanaconda.core.configuration.ui import UserInterfaceSection
 from pyanaconda.core.configuration.timezone import TimezoneSection
+from pyanaconda.core.configuration.localization import LocalizationSection
 from pyanaconda.core.constants import ANACONDA_CONFIG_TMP, ANACONDA_CONFIG_DIR, \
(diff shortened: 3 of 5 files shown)


jkonecny12 (approved):
Looks great to me. Thanks!

jkonecny12:
I wonder why the rpm-tests are failing? npm issue?

docs/release-notes/lang-configuration-setting.rst
```diff
@@ -0,0 +1,17 @@
...
+    New Localization section with lang setting is added to Anaconda
+    configuration.
+
+    The setting corresponds to the kickstart lang option.
+
+    Intended use: taking over configuration provided by other tools in Anaconda
+    configuration drop directory (/etc/anaconda/conf.d), eg Gnome Inital Setup.
+
+    Priority of the configuration:
+    inst.lang > kickstart lang > conf.localization.lang > geolocation
```
poncovka, line 14: This is stressing me out. The configuration file always has the lowest priority, because it provides defaults. It is documented here: https://anaconda-installer.readthedocs.io/en/latest/configuration-files.html#runtime-configuration-file So it cannot be never ever higher than a kickstart file or boot options. I strongly recommend to follow this practice. You already struggle with mixed priorities of kickstart commands and boot options.

I looked up the Jira issue and I don't fully understand the use cases you are trying to cover. Don't you target Fedora Workstation? Why are kickstart and boot options dragged into this? The comment at https://github.com/rhinstaller/anaconda-webui/issues/1#issuecomment-1516406850 says that the pre-selected language will be applied to the Live environment. Anaconda should be actually able to pick up this value automatically if the language is not set:

https://github.com/rhinstaller/anaconda/blob/ea679e9aa9d5bee788c918e4e61ba208bb23b1d6/pyanaconda/localization.py#L647-L656

Btw. the language seems to be a good candidate for introducing someting like `SetLanguageWithPriority` (see `SetTimezoneWithPriority`) and get rid of `LanguageKickstarted`, but that is probably out of scope of this issue.

poncovka (commented):
This doesn't follow the documentation nor the conventions.

docs/release-notes/lang-configuration-setting.rst
rvykydal: > This is stressing me out. The configuration file always has the lowest priority, because it provides defaults. It is documented here: https://anaconda-installer.readthedocs.io/en/latest/configuration-files.html#runtime-configuration-file So it cannot be never ever higher than a kickstart file or boot options. I strongly recommend to follow this practice. You already struggle with mixed priorities of kickstart commands and boot options.

I am not following, the priority scheme wanted to say that the priority is the same as before with the configuration option just overriding the default or geolocation.
But it really seems we are misusing configuration to do something it should not to. We should then rather turn off geolocation for our use case.

poncovka: > I am not following, the priority scheme wanted to say that the priority is the same as before with the configuration option just overriding the default or geolocation.

Right, I have got confused by the order. My mistake. Still, it doesn't make sense to prioritize the configuration option over the geolocation result.

> We should then rather turn off geolocation for our use case.

Agreed.

poncovka (approved):
Looks good to me. Thank you!

jkonecny12 (approved):
Looks great to me.

VladimirSlavik (approved):
Looks good to me. Thank you!

PR #4720, opened 2023-04-28
RHEL-9 wait for OEMDRV

jkonecny12:
OEMDRV disk is used for both Driver Update and Kickstart.
When booting with the DVD and an additional OEMDRV disk, it appears that
the udev rule setting the script to fetch the kickstart may not run at
the time /dev/root was found on the DVD, causing the kickstart to not be
processed at all.
This patch delays 'settled' initqueue by 5 seconds by default when no
'inst.ks' is specified on the kernel command line, which is enough to
process the OEMDRV disk.
Because the same issue may appear with Driver Updates, that same delay
applies unless 'inst.dd' is explicitly specified.
Old behavior can be restored with adding 'inst.wait_for_disks=0' on the
kernel command line.

Backport of https://github.com/rhinstaller/anaconda/pull/4586

*Resolves: rhbz#2171811*

### DIFF
--- a/docs/boot-options.rst
+++ b/docs/boot-options.rst
@@ -644,6 +644,22 @@ will be used by default, if found.
 
 See the |anacondalogging|_ for more info on setting up logging via virtio.
 
+.. inst.wait_for_disks:
+
+inst.wait_for_disks
+^^^^^^^^^^^^^^^^^^^
+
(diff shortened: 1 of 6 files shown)


M4rtinK (approved):
Looks good to me. :)

VladimirSlavik (approved):
Looks good to me. Thank you!

PR #4722, opened 2023-04-28
RHEL-9: Improve driver updates debugging

jkonecny12:
Debugging of the driver_updates script could be a pain. To improve that we need much more logs in the whole file in a way that it won't bother users during boot. To solve that we are adding debug logs and showing them on the console only in case that inst.debug or rd.debug is used.

Backport of https://github.com/rhinstaller/anaconda/pull/4149.

*Resolves: rhbz#2154906*


M4rtinK (approved):
Looks good to me. :)

VladimirSlavik (approved):
Thanks!

PR #4723, opened 2023-04-28
Rhel 8 improve driver updates debugging

jkonecny12:
Debugging of the driver_updates script could be a pain. To improve that we need much more logs in the whole file in a way that it won't bother users during boot. To solve that we are adding debug logs and showing them on the console only in case that inst.debug or rd.debug is used.

Backport of https://github.com/rhinstaller/anaconda/pull/4149.

*Resolves: rhbz#2154904*

**TODO:**
- [x] Fix tests

### DIFF
--- a/tests/nosetests/dracut_tests/test_driver_updates.py
+++ b/tests/nosetests/dracut_tests/test_driver_updates.py
@@ -26,11 +26,13 @@ import tempfile
 import shutil
 import collections
 
+import logging
+
 import sys
 sys.path.append(os.path.normpath(os.path.dirname(__file__)+'/../../dracut'))
 
-from driver_updates import copy_files, move_files, iter_files, ensure_dir
-from driver_updates import append_line, mkdir_seq
+from driver_updates import copy_files, move_files, iter_files, ensure_dir, append_line, \
+    mkdir_seq, is_debug_mode_enabled, setup_log
 
 
 def touch(path):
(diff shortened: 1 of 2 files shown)


M4rtinK (approved):
Looks good to me, thanks! :)

tests/nosetests/dracut_tests/test_driver_updates.py
```diff
@@ -84,6 +86,95 @@ def makefiles(self, *paths):
...
+
+        calls = mock_log.addHandler.mock_calls
+        calls = list(map(repr, calls))
+
+        # with debug mode disabled
+        # DEBUG logs go to journal only
+        self.assertIn("call(<SysLogHandler (DEBUG)>)", calls)
+        # and INFO logs to the console
+        self.assertNotIn(
+            "call(<StreamHandler <_io.FileIO name=8 mode='rb+' closefd=True> (DEBUG)>)",
```
VladimirSlavik, line 141: I think it would be better to just use the calls as objects, not converting to strings...

jkonecny12: I was thinking about that but the issue is that you then comparing two python objects, so you have to deal with identity.

I played with that today and found out:
this call:
```
        self.assertIn(mock.call(syslog), calls)
```
will give you this error:
```
AssertionError: call(<SysLogHandler (DEBUG)>) not found in [call(<StreamHandler <stderr> (INFO)>), call(<SysLogHandler (DEBUG)>)]
```

as you can see `SysLogHandler (DEBUG)` is there but it's not the same object so it's false statement. For that reason I decided to compare with repr which will solve this identity issue for you.

Do you have an idea how to resolve this issue?

VladimirSlavik: Ohhhh. I get it now. Yes, you can't compare the objects like that. That's why `Mock` has things like [`assert_any_call`](https://docs.python.org/3/library/unittest.mock.html?highlight=assert_any_call#unittest.mock.Mock.assert_any_call).

jkonecny12: Yes, but the issue here is that the call argument is the object I want to check. So it will again compare just objects and you need to find out what object type it was.

There is solution, you have to unpack all the calls and use `instanceof` and check logLevel. But given to fact that you are testing one of many is present it is much more code and issues than just comparing output of the repr call.

You would have to do something like.

```
self.assertTrue(any(filter(custom_check_func(map(unpack, calls)))
```
and you have to define tho `custom_check_func` which would basically check `instanceof` and `logLevel`. Honestly, I think using `repr` is better.


Another option when I'm thinking about that. is to mock the whole `SysLogHandler` and similar classes. Benefit is that you can check your mock was created. Downside - more mocking. Still not sure if it is worth the trouble to make this working.

jkonecny12: I guess the best approach here would be to change the code to abstract away handler creation so I can check the function which creates these objects. That would work well but it's more an upstream work than something to do here. At the end there is not a big chance that the repr output will change on RHEL-8.

jkonecny12:
Stop of over-engineering solutions and just simply fix the tests ;)

jkonecny12:
This PR tests are blocked by: https://github.com/rhinstaller/anaconda/pull/4723 because of new version of pykickstart.

VladimirSlavik (approved):
Thanks! I agree, maybe it's not worth the time to make the tests perfect for rhel 8.

PR #4725, opened 2023-05-02
conf: Missing geolocation provider URL disables it

VladimirSlavik:
SSIA. This adds a quick and simple way of disabling geolocation.


jkonecny12 (approved):
This is great. Thanks!

PR #4726, opened 2023-05-03
anaconda-ci: Dockerfile: Force nodejs 18 on Rawhide

KKoukiou:
nodejs 20 crashes left and right, see
https://bugzilla.redhat.com/show_bug.cgi?id=2190075

Install the previous nodejs 18 for now, and update the /usr/bin/node symlink to point to 18.

### DIFF
--- a/dockerfile/anaconda-ci/Dockerfile
+++ b/dockerfile/anaconda-ci/Dockerfile
@@ -51,7 +51,6 @@ RUN set -ex; \
   # Install rest of the dependencies
   dnf install -y \
   /usr/bin/xargs \
-  npm \
   rpm-build \
   git \
   bzip2 \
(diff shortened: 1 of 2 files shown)


martinpitt:
:+1: Same fix as in https://github.com/cockpit-project/cockpit/pull/18735 . Unfortunately this is spreading fast..

VladimirSlavik:
@sgallagher mentioned on fedora-devel creating a build that temporarily drops optimizations and avoids the crashes. But that will take time to propagate anyway...

VladimirSlavik (commented):
Hah, nice symlink. For me that was the point where I decided I don't want to hack this too much, but if it works, fine.

However, I think you still need npm, not only node... `nodejs18-npm`.

I wonder how Packit works without this change, but maybe not worth touching right now.

Overall, I think this is ok as a "patch" to revert once things work again.

(Not acking yet because tests didn't succeed.)

PR #4727, opened 2023-05-04
webui: create disk images for VirtInstallMachine in advance

rvykydal:
I run the tests 7 times on this PR, also doing "RETRY X (test affected tests 3 times)" in some, and all the tests passed without issues.

### DIFF
--- a/ui/webui/test/machine_install.py
+++ b/ui/webui/test/machine_install.py
@@ -20,6 +20,7 @@ import socket
 import subprocess
 import sys
 import time
+import tempfile
 
 ANACONDA_ROOT_DIR = os.path.normpath(os.path.dirname(__file__)+'/../../..')
 WEBUI_DIR = f'{ANACONDA_ROOT_DIR}/ui/webui'
@@ -54,6 +55,12 @@ class VirtInstallMachine(VirtMachine):
                 return port
             port = port + 1
 
+    def _create_disk_image(self, size=15):
+        name = f"disk-anaconda-{self.label}"
+        _, image = tempfile.mkstemp(suffix='.qcow2', prefix=name, dir=self.run_dir)
+        subprocess.check_call(["qemu-img", "create", "-q", "-f", "qcow2", image, str(size)+"G"])
+        return image
+
     def start(self):
         update_img_file = os.path.join(ANACONDA_ROOT_DIR, "updates.img")
         if not os.path.exists(update_img_file):
@@ -66,6 +73,8 @@ class VirtInstallMachine(VirtMachine):
         http_port = self._get_free_port(8000)
         self.http_server = subprocess.Popen(["python3", "-m", "http.server", str(http_port)])
         os.chdir(WEBUI_DIR)
+
+        disk_image = self._create_disk_image(15)
 
         try:
             self._execute(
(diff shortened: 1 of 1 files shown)


VladimirSlavik (commented):
Can you please also explain why create the file directly, and not with `qemu-img create ...`?

ui/webui/test/machine_install.py
```diff
@@ -87,7 +96,7 @@ def start(self):
                 "-device virtio-net-pci,netdev=hostnet0,id=net0,addr=0x4' "
                 f"--initrd-inject {os.getcwd()}/test/ks.cfg "
                 "--extra-args 'inst.ks=file:/ks.cfg' "
-                "--disk size=15,format=qcow2 "
+                f"--disk path={disk_image},bus=virtio,cache=unsafe,format=qcow2 "
```
KKoukiou, line 99: Did you consider just keeping the code as is and instead catching the exception and retrying it 2-3 times in case it collides with another VM image creation?
Just an alternative suggestion to consider, not review request.

rvykydal: Yes, I considered it, but given Cockpit is pre-creating the image as well in virtual_machine.py, and we are pre-creating the image in Anaconda kickstart tests I was thinking this is a better way to go.

rvykydal:
> Can you please also explain why create the file directly, and not with `qemu-img create ...`? All the answers for "create qcow2 manually" use this utility.

Huh, I am using `qemu-img create`, am I not?

ui/webui/test/machine_install.py
```diff
@@ -54,6 +55,12 @@ def _get_free_port(self, start_port):
                 return port
             port = port + 1
 
+    def _create_disk_image(self, size=15):
+        name = f"disk-anaconda-{self.label}"
+        _, image = tempfile.mkstemp(suffix='.qcow2', prefix=name, dir=self.run_dir)
```
KKoukiou, line 60: I guess you use mkstemp to get some completely random name for the disk?
Generally self.label should be unique, and it is in cockpit, it's a bug in anaconda-webui test code that this is not unique. @jelly was taking a look on this.


Maybe add a comment to revert to using self.label as prefix for the disk instead of name when this will be fixed?

ui/webui/test/machine_install.py
KKoukiou, line 99: Since it's not the virt-install creating the image anymore, you can remove format parameter for the --disk.

ui/webui/test/machine_install.py
rvykydal: I'd prefer to be safe here using random name. Also I can imagine creating more than 1 disk this way for a test / vm (I am already using it locally when developing). Maybe we should even use this in `add_disk()` to be able to add more then a single disk in scope of a single vm? I think self.label is per-vm thing?

ui/webui/test/machine_install.py
rvykydal: I'll fix that.

VladimirSlavik:
>> Can you please also explain why create the file directly, and not with qemu-img create ...? All the answers for "create qcow2 manually" use this utility.

> Huh, I am using qemu-img create, am I not?

Yes, you are, and I'm blind. I thought it was done with the temporary files.

ui/webui/test/machine_install.py
rvykydal: Done

ui/webui/test/machine_install.py
rvykydal: Maybe we can consolidate creating disk(s) for testing vm and adding disks with add_disk: https://github.com/rhinstaller/anaconda/pull/4731 ?

rvykydal: @KKoukiou would you be OK with this version, or https://github.com/rhinstaller/anaconda/pull/4731 ?

PR #4730, opened 2023-05-04
Add Runtime module

VladimirSlavik:
Add the Runtime module. Store dracut-only kickstart commands in a submodule of that.

Replaces #4712. This is the better choice than trying to twist Boss into dual use - the patterns are the same as the other modules.

(Eventually, store also all the remaining things in there instead of Boss, changing the plan slightly.)

TODOs:

- [x] Tests
- [x] Better naming suggestions for the _dracut_only_ thing are highly welcome.
- [x]  Add a release note that there is a new module that must not be disabled.
- [x] Inform Image builder about the same.

### DIFF
--- a/docs/release-notes/runtime-module.rst
+++ b/docs/release-notes/runtime-module.rst
@@ -1,0 +1,19 @@
+:Type: Modularization
+:Summary: New Runtime module
+
+:Description:
+    Anaconda now has a new D-Bus module called ``Runtime``. This module stores run-time
+    configuration of the installer and provides methods for the overall installer flow control.
+
+    .. TODO: Clarify the text as further changes are added in subsequent PRs.
+
+    Warning: This module must always run, or anaconda crashes. Users of the following
+    configuration file entries must adapt to this change:
+
+    - ``kickstart_modules``
+    - ``activatable_modules``
+    - ``forbidden_modules``
+    - ``optional_modules``
+
+:Links:
+    - https://github.com/rhinstaller/anaconda/pull/4730


pyanaconda/modules/runtime/dracut_only_commands/dracut_only_commands.py
```diff
@@ -0,0 +1,83 @@
...
+    def __init__(self):
+        super().__init__()
+        self._driverdisks = []
+        self._ssh_pws = []
+        self._mediacheck = False
+        self._updates_url = ""
+
+    def process_kickstart(self, data):
+        """Process the kickstart data."""
+        for ks_dd_data in data.driverdisk.driverdiskList:
```
poncovka, line 45: I would be fine with keeping and reusing a copy of the `DriverDiskData` list if you want to simplify the code. The data objects shoudn't have a reference to the kickstart handler (unlike the command objects). Same for `SshPwData`.

poncovka (approved):
I am so glad we have decided to do this. It is so much better. Thank you!

pyanaconda/modules/runtime/dracut_only_commands/dracut_only_commands.py
VladimirSlavik: I'm conflicted. Keeping the data as such did not appear to be a good idea in my previous testing. On the other hand, missing potential new options here would be a problem, too.

poncovka (approved):
Thanks!

PR #4738, opened 2023-05-05
infra: Drop RHEL 8 COPR nightly builds

VladimirSlavik:
Up for discussion - not to be blindly merged before team discussion.

Why this is desirable:
- These are not used anywhere
- Mostly remembered only when the token runs out every quarter
- RHEL 8 is in the late lifecycle stage and receives few changes
- Daily release rehearsals include RHEL 8
- Less things running on the internal runners = closer to migrating off these
- One less action explicitly using RH-internal data (certs, token)

What's lost with this:
- Builds using the actual RH internal infra
- Usage of vtrefny's copr_builder


M4rtinK (approved):
Looks good to me. :)

PR #4740, opened 2023-05-08
Load translations on the fly - eval approach

jelly:
Cockpit's translations are served as `po.js` which returns the language depending on the set browser cookie. The served `po.js` is a JavaScript file which executes `cockpit.locale({ ...locale data })`.

To make the application reload on the fly three things need to happen:
- set the relevant cookie
- fetch the new translations and execute them. As well it's a JavaScript file.
- force React to re-render the whole application

The `eval()` here is quite evil, even serving po's as json won't be an easy task as it's the locale data content is a JavaScript Object with a function which cannot be serialised to JSON. There is a function in the translation data as `cockpit.ngettext` needs to figure out which variant to pick which can have multiple forms for example in Polish it is `"plural-forms": (n) => n==1 ? 0 : n%10>=2 && n%10<=4 && (n%100<10 || n%100>=20) ? 1 : 2,`

### DIFF
--- a/ui/webui/src/components/localization/InstallationLanguage.jsx
+++ b/ui/webui/src/components/localization/InstallationLanguage.jsx
@@ -34,7 +34,7 @@ import {
 } from "@patternfly/react-core";
 
 import { EmptyStatePanel } from "cockpit-components-empty-state.jsx";
-import { AddressContext } from "../Common.jsx";
+import { AddressContext, LanguageContext } from "../Common.jsx";
 import { setLocale } from "../../apis/boss.js";
 
 import {
@@ -299,6 +299,7 @@ LanguageSelector.contextType = AddressContext;
 
 export const InstallationLanguage = ({ idPrefix, setIsFormValid, onAddErrorNotification }) => {
     const [nativeName, setNativeName] = React.useState(false);
+    const { setLanguage } = React.useContext(LanguageContext);
 
     return (
         <AnacondaPage title={_("Welcome to the Anaconda installer")}>
@@ -326,6 +327,7 @@ export const InstallationLanguage = ({ idPrefix, setIsFormValid, onAddErrorNotif
                       setIsFormValid={setIsFormValid}
                       onAddErrorNotification={onAddErrorNotification}
                       getNativeName={setNativeName}
+                      reRenderApp={setLanguage}
                     />
                 </FormGroup>
             </Form>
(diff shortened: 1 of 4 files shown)


ui/webui/src/components/localization/InstallationLanguage.jsx
```diff
@@ -236,7 +236,19 @@ class LanguageSelector extends React.Component {
...
+                        fetch("po.js").then(response => response.text())
+                                .then(body => {
+                                    // always reset old translations
+                                    cockpit.locale(null);
+                                    // en_US is always null
+                                    if (body.trim() === "") {
+                                        cockpit.locale(null);
+                                    } else {
+                                        // eslint-disable-next-line no-eval
+                                        eval(body);
```
M4rtinK, line 248: I wonder - can the contents of this be influenced via say crafted translation strings ? Eq. someone injecting JS instead of a translation in Weblate & it then being executed via this eval ?

M4rtinK (approved):
Looks good to me as far as I can tell - just adding a questions about how the eval() nterats with user contributed translation strings. :)

ui/webui/src/components/localization/InstallationLanguage.jsx
VladimirSlavik: We have checks for something like that in Weblate, but it's unclear how much that would protect us.

jelly:
> @jelly So... if I understand correctly, the risk does not really change with this, it just becomes more explicit. The usual route for the data is to have the page link the script like `<script src="po.js" />`, which evaluates it at load time. So then what this does is evaluate the file manually, which overwrites the values and thus bypasses the need to reload the page. Correct?

Yes, we load the po.js manually and evaluate it (execute it as the browser would normally load it) after that we need to do some more special sauce to actually use the new translation strings.

The bigger downside is that with this PR we now allow `eval` in the whole application, previously this was denied by [CSP](https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP).

VladimirSlavik:
Oh, so, while the code LGTM, I don't feel qualified to judge the policy... CPS thing... which seems to me the main point for a yes/no decision.

Can we allow eval only in the one place? It's a SPA so probably the JSX granularity has no relationship with the actual browser granularity?

jelly:
> Oh, so, while the code LGTM, I don't feel qualified to judge the policy... CPS thing... which seems to me the main point for a yes/no decision.
> 
> Can we allow eval only in the one place? It's a SPA so probably the JSX granularity has no relationship with the actual browser granularity?

I'm afraid that is not possible, the Content security policy is a header which is applied per origin (think "domain"). So any code served in the whole WebUI can now call `eval`.

KKoukiou (approved):

PR #4741, opened 2023-05-08
WebUI: improve handling of removal of testvm's

jelly:
When we call machine.start() we already create a libvirt domain so we need to handle a user pressing ctrl+c. Additionally move the SIGTERM handler up to after the creation of the VM.

Fixes https://issues.redhat.com/browse/INSTALLER-2945

### DIFF
--- a/ui/webui/test/webui_testvm.py
+++ b/ui/webui/test/webui_testvm.py
@@ -30,31 +30,31 @@ def cmd_cli():
     args = parser.parse_args()
 
     machine = VirtInstallMachine(image=args.image)
-    machine.start()
-
-    print("You can connect to the VM in the following ways:")
-    # print ssh command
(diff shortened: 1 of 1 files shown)


VladimirSlavik (approved):
Looks good to me, thank you!

KKoukiou (commented):
This script is essentially a copy from https://github.com/cockpit-project/bots/blob/main/machine/machine_core/cli.py. 

And I think these changes now diverge these a bit. Can you check the initial script and if so make the changes in both places ?

jkonecny12 (approved):
Thanks for the improvement.

KKoukiou:
@jelly did you try to check the original script? https://github.com/cockpit-project/bots/blob/main/machine/machine_core/cli.py

As mentioned above this commit is literally a copy over, and fixing things only here feels weird.

PR #4745, opened 2023-05-09
infra: Fix release rehearsal status reporting

VladimirSlavik:
As it turns out, continue-on-error also marks things with green check, so the overall status could never become red.

~~TODO: Test on my fork.~~ "Successful" run that correctly ran both tests and failed both: https://github.com/VladimirSlavik/anaconda/actions/runs/4926254342/jobs/8801552796


rvykydal (approved):
Thank you.

PR #4748, opened 2023-05-09
WebUI: close embedded panel when clicking prev/next

jelly:
When the Storage drawer is opened, automatically close it when clicking previous or next.

Resolves: INSTALLER-3405

### DIFF
--- a/ui/webui/test/check-progress
+++ b/ui/webui/test/check-progress
@@ -18,6 +18,7 @@
 import anacondalib
 
 from installer import Installer
+from storage import Storage
 from testlib import test_main  # pylint: disable=import-error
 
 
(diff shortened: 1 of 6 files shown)


ui/webui/test/check-progress
```diff
@@ -31,11 +31,19 @@ class TestInstallationProgress(anacondalib.VirtInstallMachineCase):
...
 
         i.open()
+        drawer_open = False
 
         for page in i.steps[:-1]:
             # with the pages basically empty of common elements (as those are provided by the top-level installer widget)
             # we at least iterate over them to check this works as expected
             i.wait_current_page(page)
+            if drawer_open:
+                self.assertFalse(b.is_present(".pf-c-drawer__panel-main"))
```
rvykydal, line 41: I think the common practice in our tests is to use the element ids via helper functions in test/helpers/*.py. For example check_ or click_ funcitons in test/helpers/storage.py. So maybe it is worth following it also here, like having check_drawer_open function somewhere?

M4rtinK (approved):
Looks good to me, thanks! :) 

Also agreed with Radek on ideally accessing this via the library function in tests. :)

ui/webui/test/check-progress
jelly: Added helper functions

rvykydal (approved):
Looks good to me.

VladimirSlavik (approved):
Looks good to me. Thank you!

PR #4749, opened 2023-05-10
Save selected storage scenario in state

jelly:
This is preparation work for the custom mount storage configuration, to not make a giant PR with a bunch of changes I've split this up into it's own PR.

Additionally as the initialisation method was stored by the Anaconda daemon, I wanted retain that behaviour in some form as a user could accidentally refresh the browser tab. I've opted to store the storage configuration in sessionstorage instead of localstorage as localstorage is more or less permanent. SessionStorage survives a refresh, but should go away once the user closes their browser session.

### DIFF
--- a/ui/webui/src/components/review/ReviewConfiguration.jsx
+++ b/ui/webui/src/components/review/ReviewConfiguration.jsx
@@ -35,7 +35,6 @@ import {
     getDeviceData,
     getAppliedPartitioning,
     getPartitioningRequest,
-    getInitializationMode,
 } from "../../apis/storage.js";
 
 import {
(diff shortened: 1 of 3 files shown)


ui/webui/src/components/review/ReviewConfiguration.jsx
```diff
@@ -35,15 +35,14 @@ import {
     getDeviceData,
     getAppliedPartitioning,
     getPartitioningRequest,
-    getInitializationMode,
```
rvykydal, line 38: Maybe we could also remove getInitializationMode function (https://github.com/rhinstaller/anaconda/blob/7779a2812059d42a7554e80f5b3e9ca9a785f9a3/ui/webui/src/apis/storage.js#L291) which is no more used at this point.

VladimirSlavik (approved):
Thank you!

PR #4750, opened 2023-05-10
Use the latest Read the Docs theme (rhel-8)

VladimirSlavik:
Backport from upstream.

Fixes readthedocs.


VladimirSlavik:
@jkonecny12, @jstodola, this is de facto infrastructure, do you agree with merging this without a bug etc.?

jstodola (approved):
I agree with merging this without a bug etc.

PR #4751, opened 2023-05-10
Use the latest Read the Docs theme

VladimirSlavik:
Backport from upstream.

Fixes readthedocs.


VladimirSlavik:
@jkonecny12, @jstodola, this is de facto infrastructure, do you agree with merging this without a bug etc.?

jkonecny12 (approved):
LGTM! Thanks for the fix!

PR #4756, opened 2023-05-11
WebUI: document pixel-test fetch command

jelly:
For example:

```
GITHUB_BASE=rhinstaller/anaconda ./test/common/pixel-tests fetch "https://cockpit-logs.us-east-1.linodeobjects.com/pull-4728-20230505-082156-923abdb2-fedora-rawhide-boot/log.html"
```

Fetches the changed pixels for you and you only need to call `make update-reference-images` afterwards

### DIFF
--- a/ui/webui/test/README.rst
+++ b/ui/webui/test/README.rst
@@ -139,11 +139,11 @@ Locally just copy the broken tests images to the `test/reference` directory. How
 option to deal with this is to use automation which will download all the broken images from
 fail test on PR::
 
-    ./test/common/pixel-tests fetch <link to HTML with failed tests>
+    GITHUB_BASE=rhinstaller/anaconda ./test/common/pixel-tests fetch <link to HTML with failed tests>
 
 Example of such a call::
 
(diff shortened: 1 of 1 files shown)


VladimirSlavik (approved):
Thank you!

jkonecny12:
Hi @jelly this is already documented a few lines below. Could you please avoid duplication of this documentation? I'm fine with removing the old one if you think that is better.

jkonecny12 (changes requested):
Duplication of documentation should be avoided first.

VladimirSlavik:
The example below does not use `GITHUB_BASE` - what's the difference?

jkonecny12:
I think it's safer to have that specified, maybe even required. I have a feeling I wrote that without having place to test it.

jelly:
@jkonecny12 GITHUB_BASE is required as otherwise you would push pixel tests to cockpit-project, I've made it explicit now.

jkonecny12 (approved):
Looks great to me now. Thanks for the improvement!

VladimirSlavik (approved):
Thanks!

I'd say looks good to me, but apparently I'm getting blind :D

PR #4768, opened 2023-05-16
Add readme for the conf.d drop dir

VladimirSlavik:
What it says.


VladimirSlavik:
It seems that the `.txt` extension goes against the convention of calling it just a `README`, so I changed that.

PR #4771, opened 2023-05-16
RHEL 9 geolocation and timezone

VladimirSlavik:
This selects timezone based on selected language, when geolocation did not provide the timezone data.

As a side effect, it drags in modularized geolocation running in a task, in the main process startup sequence, so that the geolocation result applies also to TUI.

~~WIP, right now this is just identification of the commits that should constitute the port.~~

(I hope it's clear why the RHEL-8 port is close to impossible.)

https://bugzilla.redhat.com/show_bug.cgi?id=2127473


VladimirSlavik:
Note to self: Need to research if commit "Don't fall back to default timezone in welcome spoke" introduced a problem and was fixed later, or not.

VladimirSlavik:
This needed slight modification to make TUI work correctly for RHEL. The changes make geolocation apply to TUI too, but it has no way to change the language, so it becomes stuck in the geolocated language, no matter what.

The last commit is new on RHEL and solves this.

It might be good to consider if we want to support fonts and languages in TUI at all - kernel breaks resolution when we set fonts, everything has been disabled for quite some time, and the code is needlessly complicated while unused.

VladimirSlavik:
Re-running succeeded, so I consider the tests to be 100% passed.

pyanaconda/core/configuration/anaconda.py
```diff
@@ -404,9 +408,30 @@ def set_from_opts(self, opts):
...
 
         self.validate()
 
 
+def _convert_geoloc_provider_id_to_url(provider_id):
+    """Convert provider ID to URL of the corresponding service.
+
+    :param str provider_id: id of the geolocation provider service
+    :return str: URL to use
+    """
```
jkonecny12, line 424: If I'm reading it correctly. It would be valuable to have a note that this is only kept for backward compatibility. Or am I incorrect?

pyanaconda/ui/gui/spokes/datetime_spoke.py
```diff
@@ -534,14 +534,6 @@ def _initialize(self):
...
         if is_valid_timezone(kickstart_timezone):
             self._set_timezone(kickstart_timezone)
-        elif not flags.flags.automatedInstall:
-            log.warning("%s is not a valid timezone, falling back to default (%s)",
-                        kickstart_timezone, DEFAULT_TZ)
-            self._set_timezone(DEFAULT_TZ)
-            self._timezone_module.SetTimezoneWithPriority(
-                DEFAULT_TZ,
-                constants.TIMEZONE_PRIORITY_KICKSTART
-            )
```
jkonecny12: Could you please explain me why the correction from kickstart is not needed anymore? Is that because it is already set from the module?

jkonecny12 (approved):
This changeset is big but I really like the changes. It's small chunks easy to review one by one. Great work and thanks for this!

jkonecny12:
> Re-running succeeded, so I consider the tests to be 100% passed.

Maybe re-running the whole suite so it's nice a green?

pyanaconda/core/configuration/anaconda.py
VladimirSlavik: Sorry, I missed this aspect, both back then and now. Yes, that's correct.

In fact, we should probably officially deprecate the possibility to use `inst.geoloc=<provider-id>` and ask to use `[Timezone] geolocation_provider = <url>` in conf. Both for RHEL 9 and Fedora 39.

We already shipped this in f37, but never added the deprecation notice to docs and release notes, it seems. I can't find any PR that would satisfy the TODO there: https://github.com/rhinstaller/anaconda/pull/4074#issue-1219112128

pyanaconda/ui/gui/spokes/datetime_spoke.py
VladimirSlavik: Yes.

VladimirSlavik:
7 timeouts now. Doesn't look better to me :-/

M4rtinK (approved):
Quite a big one, but looks good to me as far as I can tell. :)

Just a little questions/suggestions bout a docstring. :)

pyanaconda/startup_utils.py
```diff
@@ -661,16 +670,69 @@ def start_geolocation_conditionally(opts):
...
         log.debug("Geolocation timed out. Exceptions will not be logged.")
+        return
+    else:
+        apply_geolocation_result(display_mode)
+
+
+def apply_geolocation_result(display_mode):
+    """Apply geolocation result.
+
+    This does not check for kickstart installations, because in that case geolocation is not even
```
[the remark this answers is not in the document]
VladimirSlavik: Ah, I see. The comment really is a bit misleading. The thing is, `apply_geolocation_result`(this) is one of the multiple parts involved, and the last one, so it does not get to choose much :-)

Here's the whole picture:
- First, you call `start_geolocation_conditionally` that handles the starting, and damage control in case modules are not available etc.
- That actually calls `check_if_geolocation_should_be_used` to determine if geolocation should be started, which is where the kickstart status and boot option is actually considered.
- Then, you have `wait_for_geolocation_and_use` that actually handles the task and its finalization, and calls `apply_geolocation_result`.
- Once you are finally in `apply_geolocation_result` which you're commenting here, you just apply the result, the decision to start or not was already taken earlier.

I will think about describing this better.

VladimirSlavik:
kickstart-test --testtype coverage

VladimirSlavik:
@M4rtinK, sorry, I added two more commits...

jkonecny12 (approved):
Looks great to me. Thanks for the fixes.

jkonecny12:
Removing ready_to_merge because of the conflicts.

jkonecny12:
I did the rebase and solved one minor conflict in the `constants.py` file to unblock this change for a release.

Sorry for not asking (you are not available), @VladimirSlavik hope you don't mind.

PR #4774, opened 2023-05-19
infra: Rebuild container during RPM test only when needed

VladimirSlavik:
What it says.

Needs testing live in GHA.


VladimirSlavik:
Tested successfully here: https://github.com/VladimirSlavik/anaconda/pull/93

First commit test run did not rebuild the container, because the changes did not contain anything on the list of paths. Second commit changes something relevant and the rebuild happens.

.structure-config
```diff
@@ -13,3 +13,10 @@ scripts/testing/
 dracut/.shellcheckrc
 docs/ci-status.rst
 )
+
+RPM_REBUILD_PATHS=(
+anaconda.spec.in
+dockerfile/anaconda-rpm
+.github/workflows/tests.yml
+branch-config.mk
+)
```
VladimirSlavik, line 22: I am not sure if this list has everything, but it seems like it could be a good start.

I considered also including Makefile, but only the top level one applies, because it starts the container with tests. Current pattern matching would take all makefiles.

It's trivial to make it so that only a top level Makefile can be specified - change
```diff
- if [[ "$iter_f" =~ "$rebuild_f" ]]; then
+ if [[ "./$iter_f" =~ "$rebuild_f" ]]; then
```
and then files can be specified relative to repo root as eg. `./Makefile`

M4rtinK (approved):

.github/workflows/tests.yml
```diff
@@ -127,7 +127,29 @@ jobs:
           git log --oneline -1 ${{ env.TARGET_BRANCH_NAME }}
           git rebase ${{ env.TARGET_BRANCH_NAME }}
 
+      - name: Check if container rebuild is needed
+        id: need_rebuild
+        run: |
+          changed_files=$(git diff --name-only ${{ env.TARGET_BRANCH_NAME }}..HEAD)
+          echo -e "Changed files: \n$changed_files\n"
+
+          . .structure-config
```
jkonecny12, line 136: This could be a potential attack source.

In your PR you can just change the `.structure-config` file and you can do anything to the workflow here.

However, it's probably not that problematic because we have to confirm tests for 3rd party contributors. Just want to point this out.

jkonecny12 (commented):
Sorry for the late review. Thanks a lot for solving this!

.github/workflows/tests.yml
VladimirSlavik: Ah, right - we do the same but on `pull_request_target` so it's really different https://github.com/rhinstaller/anaconda/blob/master/.github/workflows/infra-check.yml#L63 Thanks for the note!

PR #4780, opened 2023-05-19
WIP: Logging brace style

VladimirSlavik:
Change logging format string style from percent to braces.

There's a "fun" part where blivet logs to our functions somehow, that is not resolved yet. I don't really understand why we do that. Keyword: `log_exception_info`.

### DIFF
--- a/anaconda.py
+++ b/anaconda.py
@@ -220,15 +220,15 @@ if __name__ == "__main__":
     if startup_utils.prompt_for_ssh(opts):
         sys.exit(0)
 
-    log.info("%s %s", sys.argv[0], util.get_anaconda_version_string(build_time_version=True))
+    log.info("{} {}", sys.argv[0], util.get_anaconda_version_string(build_time_version=True))
 
     if opts.updates_url:
-        log.info("Using updates from: %s", opts.updates_url)
+        log.info("Using updates from: {}", opts.updates_url)
 
     # warn users that they should use inst. prefix all the time
     for arg in removed_no_inst_args:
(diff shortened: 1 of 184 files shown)


VladimirSlavik:
What a... Friday thing. Turns out the docs don't really lie, it's just one sentence that says this is wrong. Basically, the `style="{"` thing on `logging.Formatter` changes only how `fmt` and `datefmt`parameters are consumed, right there next to the style. But the actual `log.info` calls etc. always use %-formatting. Changing that is a completely different thing, done via `logging.setLogRecordFactory` and some other things.

VladimirSlavik:
So, these are the actual options:
- Set the [factory](https://docs.python.org/3/library/logging.html#logging.setLogRecordFactory)  - here's [the most bare bones example](https://stackoverflow.com/questions/13131400/logging-variable-data-with-new-format-string/42823461#42823461). This has the problem that logging uses the factory globally, and anything we import and tries to log will also get our stuff, and then break. Such as dnf and Blivet. So, that does not sound possible, unless we resort to some madness like trying to determine if the string uses {} or % formatting, or detecting if the caller is from pyanaconda.
- Use something else to log messages - [logbook](https://logbook.readthedocs.io/en/stable/) seems to be a good choice, as it is already packaged for Fedora. However, we'd have to maintain that. There is also the question how to integrate with journal and other facilities.
- Give up brace formatting in logs. The string style is not worth the cost at this moment. We want less, not more.

VladimirSlavik:
As an afterthought - I don't really understand why the factory couldn't be set per logger, which would solve the problem. Oh well.

PR #4782, opened 2023-05-22
Move the User Interface module under Runtime

VladimirSlavik:
With this, Boss will again only coordinate other modules.


M4rtinK (approved):
LGTM. :)

PR #4785, opened 2023-05-23
Verify LUKS passphrase minimum length for FIPS mode (RHEL-9)

VladimirSlavik:
Resolves: [rhbz#2163497](https://bugzilla.redhat.com/show_bug.cgi?id=2163497)


VladimirSlavik:
Looks like tests should go to `StorageInterfaceTestCase`.

VladimirSlavik:
> Code is likely final and reviewable.

Famous last words...

pyanaconda/input_checking.py
```diff
@@ -435,6 +436,27 @@ def run(self, check_request):
...
+        if len(check_request.password) < constants.FIPS_PASSPHRASE_MIN_LENGTH:
+            self.result.success = False
+            self.result.error_message = _(
+                "In FIPS mode, the passphrase must be at least {} letters long."
+            ).format(constants.FIPS_PASSPHRASE_MIN_LENGTH)
+        else:
+            self.result.success = True
+            self.result.error_message = ""
+
+
```
rvykydal, line 459: I'd maybe organize the code here a bit differently but it is not important.

rvykydal (approved):
Looks good to me, thank you.

pyanaconda/input_checking.py
VladimirSlavik: Changed.

pyanaconda/modules/storage/kickstart.py
```diff
@@ -52,6 +54,34 @@ def get_device_names(specs, disks_only=False, msg="{}", lineno=None):
...
+
+    if not kernel_arguments.is_enabled("fips"):
+        return
+
+    if len(luks_passphrase) >= FIPS_PASSPHRASE_MIN_LENGTH:
+        return
+
+    raise KickstartParseError(
+        _("Passphrase given in the {} command is too short in FIPS mode. "
+          "Please use at least {} characters.").format(command_name, FIPS_PASSPHRASE_MIN_LENGTH),
```
VladimirSlavik, line 80: Martin said something like: This message does not say that the FIPS mode is on, and users can't understand that the error is conditional on that.

I think that's true.

pyanaconda/input_checking.py
rvykydal: Something similar, thank you:)

PR #4787, opened 2023-05-24
infra: Always try to make artefacts cleanable

VladimirSlavik:
Timeout of the test step can make the job fail, and without this the files become uncleanable, and the next job can't clean them when initializing.


rvykydal (approved):
Looks good to me, thank you!

PR #4791, opened 2023-05-25
Verify LUKS passphrase minimum length for FIPS mode (port to Rawhide)

VladimirSlavik:
Preliminary port of #4785.

(I'd prefer to do this downstream first, as that would get dedicated QE work.)


tests/unit_tests/pyanaconda_tests/modules/storage/test_module_storage.py
```diff
@@ -1611,3 +1660,46 @@ def test_validation_failed(self, storage_checker):
...
+
+        with pytest.raises(KickstartParseError):
+            fips_check_luks_passphrase("weak", "command name", 0)
+
+        with pytest.raises(KickstartParseError):
+            fips_check_luks_passphrase("short", "command name", 0)
+
+        fips_check_luks_passphrase("whatever", "command name", 0)
+
+        fips_check_luks_passphrase("king kong strong", "command name", 0)
```
M4rtinK, line 1705: :+1:

M4rtinK (approved):
Looks good to me, thanks! :)

rvykydal (approved):

VladimirSlavik:
Release notes are not required, because it seems we don't have reports of this bug on Fedora. Either nobody does FIPS in Fedora, or the underlying requirement causing the crash is not present (yet).

PR #4794, opened 2023-05-26
Rhel 9 fix inst proxy

jkonecny12:
Fix proxy is not used when downloading stage2 image.

This fix is handling two cases:
- kickstart has `url --proxy` which should be used also to download stage2 image with proxy
- `inst.proxy` is not used by dracut to download stuff (like stage2 image)

*Resolves: rhbz#2177219*

To do:
- [ ] try to create some tests


VladimirSlavik (approved):
Looks good to me. Thank you!

rvykydal (approved):

PR #4795, opened 2023-05-26
Don't override blivet's preferred disk label type by default

AdamWill:
https://github.com/rhinstaller/anaconda/pull/4232 made anaconda prefer GPT disk labels by default. This was intended to implement the Change called "Install Using GPT on x86_64 BIOS by Default": https://fedoraproject.org/wiki/Changes/GPTforBIOSbyDefault https://bugzilla.redhat.com/show_bug.cgi?id=2092091 However, it actually does more than that Change requested. It prefers GPT in other cases - it's quite hard to parse out every possible case this changes, but for example, it also prefers GPT on ppc64le installs, which previously preferred MBR. It's also just an odd approach, when blivet already goes to some lengths specifically to provide a list of supported label types in a preferred order. It makes much more sense for anaconda to keep respecting the order of the list blivet provides, and adjust blivet to list GPT first on x86_64. The config setting and `inst.disklabel` boot option can still be used to force a preference; this just makes the default config value unset rather than 'gpt'.

This goes with this blivet PR that implements preferring GPT on x86_64 (even on BIOS):
https://github.com/storaged-project/blivet/pull/1132

### DIFF
--- a/anaconda.spec.in
+++ b/anaconda.spec.in
@@ -41,7 +41,7 @@ Source0: https://github.com/rhinstaller/%{name}/releases/download/%{name}-%{vers
 %define nmver 1.0
 %define pykickstartver 3.47-1
 %define pypartedver 2.5-2
-%define pythonblivetver 1:3.6.0-1
+%define pythonblivetver 1:3.7.1-4
 %define rpmver 4.15.0
 %define simplelinever 1.9.0-1
 %define subscriptionmanagerver 1.26
--- a/data/anaconda.conf
+++ b/data/anaconda.conf
@@ -181,7 +181,7 @@ btrfs_compression =
 #
 #    If not specified, use whatever Blivet uses by default.
 #
-disk_label_type = gpt
+disk_label_type =
 
 # Default file system type. Use whatever Blivet uses by default.
 file_system_type =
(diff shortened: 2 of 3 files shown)


poncovka (approved):
Thank you!

M4rtinK (approved):

VladimirSlavik (approved):
LGTM, thank you.

VladimirSlavik:
This needs the blivet patch too, so right now it's blocked, I think?

AdamWill:
yeah, we should merge the blivet one first, and also we probably ought to plan to do a release and build of blivet before merging this and building anaconda...

vojtechtrefny:
> yeah, we should merge the blivet one first, and also we probably ought to plan to do a release and build of blivet before merging this and building anaconda...

The blivet change is merged now and available in rawhide in *python-blivet-3.7.1-4*.

VladimirSlavik:
Thank you! So, I will do this:
- [ ] bump the required blivet version and force-push here
- [ ] re-run all the tests
- [ ] merge

VladimirSlavik:
Looks like that build is not really available yet...

VladimirSlavik:
> It will be available...

"Are we there yet?"

No.

AdamWill:
Hmm? We should be:

https://kojipkgs.fedoraproject.org/compose/rawhide/Fedora-Rawhide-20230605.n.0/compose/Everything/x86_64/os/Packages/p/python3-blivet-3.7.1-4.fc39.noarch.rpm

AdamWill:
oh, although, I belive @nirik said he had to untag the last anaconda build to get Rawhide composes working again, so that's a factor here - even if we merge this we need to fix whatever the problem was before we'll actually get this into a build that makes it into Rawhide...

PR #4800, opened 2023-05-30
If system uses nano as default editor, do the same

VladimirSlavik:
This affects meh and libreport in text mode.

It is not possible to source the bash file, because we run python, not bash. The file is hardcoded because there is no way to walk over the whole system and find if nano is being used.

Doing it this way also preserves the ability to not have the file, and use vi as before.

Resolves: [rhbz#1889674](https://bugzilla.redhat.com/show_bug.cgi?id=1889674)

### DIFF
--- a/anaconda.py
+++ b/anaconda.py
@@ -149,6 +149,11 @@ def setup_environment():
     else:
         os.environ["DISPLAY"] = ":%s" % constants.X_DISPLAY_NUMBER
 
+    # We mostly don't run from bash, so it won't load the file for us, and libreport will then
+    # show vi instead of nano. Resolves https://bugzilla.redhat.com/show_bug.cgi?id=1889674
+    if "EDITOR" not in os.environ and os.path.isfile("/etc/profile.d/nano-default-editor.sh"):
+        os.environ["EDITOR"] = "/usr/bin/nano"
+
 
 if __name__ == "__main__":
     # check if the CLI help is requested and return it at once,


VladimirSlavik:
For review - please evaluate if we want to do it this way. It does not feel correct to me, even if it fixes the problem.
- The file is normally sourced by bash.
- We run directly under tmux and systemd, so there's no bash on the way.
- It's possible to hardcode the variable into the systemd units, but that makes it impossible to switch.
- We don't want to run bash and see what is the value of the variable.
- We could do some regexp over the file to see if it really sets nano, but the file name itself is explicit enough, and in the installation environment, it is 
- The file name is the weak point, but I don't see any better alternative.

anaconda.py
```diff
@@ -149,6 +149,11 @@ def setup_environment():
     else:
         os.environ["DISPLAY"] = ":%s" % constants.X_DISPLAY_NUMBER
 
+    # We run directly under systemd and tmux, so bash won't load the file for us, and libreport
```
M4rtinK, line 152: IIRC in some scenarios (s390) we don't run via TMUX and also sometimes without systemd (container & Live image generation - but this might be non-interactive so out of scope). That still should not break this code though. :)

M4rtinK (approved):
Looks good to me - as long as vi remain available for manual invocation on the boot.iso I'm fine. ;-)

anaconda.py
VladimirSlavik: Ah, thank you! Yes, it should work, I think. I will change the comment, so it makes sense.

PR #4801, opened 2023-06-01
Add WebUI image to anaconda-iso-container (#infra)

jkonecny12:
It behaves almost the same as the current boot.iso with difference that we need to patch the Lorax templates before executing the build.

To run WebUI ISO build you have to add `--entrypoint /lorax-build-webui` as the `podman run` argument, otherwise it will build standard ISO.

Next step would be to adjust workflow to build ISO from a PR.

### DIFF
--- a/dockerfile/anaconda-iso-creator/lorax-build-webui
+++ b/dockerfile/anaconda-iso-creator/lorax-build-webui
@@ -1,11 +1,17 @@
 #!/bin/bash
 #
-# Build a boot.iso by lorax. The boot.iso will be stored in the `/images/` directory.
+# Build a Web UI boot.iso by lorax. The boot.iso will be stored in the `/images/` directory.
 # We have to build the RPMs files of Anaconda first and then add them as volume
 # mount to /anaconda-rpms to the container (could be RO mount).
 #
+# Compared to the normal boot.iso we need to patch the templates but other than
+# that it's mostly the same.
+#
+# To run this please do:
+#
 #   sudo make -f ./Makefile.am container-rpms-scratch
-#   sudo podman run -i --rm --privileged --tmpfs /var/tmp:rw,mode=1777 -v `pwd`/result/build/01-rpm-build:/anaconda-rpms:ro -v `pwd`/output-dir:/images:z quay.io/rhinstaller/anaconda-iso-creator:master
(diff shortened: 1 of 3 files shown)


dockerfile/anaconda-iso-creator/lorax-build-webui
```diff
@@ -0,0 +1,54 @@
...
+
+# build boot.iso with our rpms
+. /etc/os-release
+# The download.fedoraproject.org automatic redirector often selects download-ib01.f.o. for GitHub's cloud, which is too unreliable; use a mirror
+# The --volid argument can cause different network interface naming: https://github.com/rhinstaller/kickstart-tests/issues/448
+lorax -p Fedora -v "$VERSION_ID" -r "$VERSION_ID" \
+      --volid Fedora-S-dvd-x86_64-rawh \
+      --sharedir ./templates.d/99-generic/ \
+      -s http://dl.fedoraproject.org/pub/fedora/linux/development/rawhide/Everything/x86_64/os/ \
+      -s https://fedorapeople.org/groups/anaconda/webui_new_payload/repo/ \
```
VladimirSlavik, line 48: In this repo, there is only one huge package. How do I know what it is and how to make it? Or will this change so soon that it's meaningless to document?

VladimirSlavik (approved):
Thank you, LGTM. I love the patching, this is exactly how I thought it should work.

I'd use an environment variable as the switch, and just overwrite installed templates with the patch, but this works too.

dockerfile/anaconda-iso-creator/lorax-build-webui
jkonecny12: Yes, that is taken from @M4rtinK tarball. It's test payload wrapped into RPM so it could be installed by Lorax.

Ideally we want this file minimal and don't care about it. We need to drop this in the future anyway.

jkonecny12:
I was also thinking about the environment variable but then I decided to go this way because:
- putting `--environment` or `--entrypoint` seems to me like the same
- separate script allows us to be easily removed when in future the webui will be the default

About the patching, that is good idea to patch directly files in the system, however, I don't think it's worth doing the change now :)

dockerfile/anaconda-iso-creator/lorax-build-webui
M4rtinK: Its not just the payload tarball - it also has an alternative `interactive-defaults.ks` file, that actually makes the boot.iso use the tar payload without user interaction.

This is because we still don't have a config file support for setting installation payload.

Alternatively, when modularization is done, we could also use RPM + closest mirror, but that unlike the current image will not be network independent, which could be tricky as we don't have any means to configure networking on the image.

M4rtinK (approved):
Looks good to me. :)

PR #4807, opened 2023-06-02
Do not print error on inst.ks=cdrom|hd (#2077045)

jkonecny12:
In Dracut we have two places where the kickstart file is processed:

- fetch-kickstart-disk
- fetch-kickstart-net.sh

These two places are started independently.

What happens is that fetch-kickstart-disk detects that there is kickstart file to process (inst.ks=) but because it's not on network the script doesn't recognize this and print error.

Let's solve this by early return in case it's cdrom or hd localization of the kickstart file.

*Resolves: rhbz#2077045*


VladimirSlavik (approved):
Maybe indent the return more so it also looks correct, otherwise LGTM. Thank you!

PR #4809, opened 2023-06-02
webui: change the language direction when language changes

KKoukiou:
Resolves https://issues.redhat.com/browse/INSTALLER-3389


VladimirSlavik (commented):
I think this won't unset the direction once it is set. In other words, make test switch to Hebrew and then back.

jelly (approved):

rvykydal (approved):

PR #4811, opened 2023-06-02
webui: Date and time screen

M4rtinK:
Add the Date & time configuration screen, in the wizard flow it sits between the welcome screen & installation destination screen.

Screenshot 2023-06-02 at 19-41-33 Anaconda Web UI
Screenshot 2023-06-02 at 19-43-04 Anaconda Web UI

All the major controls  are there UI wise, the main missing bits are test coverage & some remaining plumbing.

**followup PR/s**
- detailed timezone information for the timezone details popup (timezone offset & timezone pretty name)
  - look like this will need backend API changes/extension
- look into date & time display formatting & cockpit compatibility
- look into timezone display - _ vs " " in timezone names

**TODO**

- [x] backend test coverage
- [ ] frontend test coverage
- [x] pre-select region & city based on backend data
- [x] date & time configuration plumbing

- [x] find out why DBus calls from the Save button handler don't work

### DIFF
--- a/pyanaconda/modules/timezone/timezone.py
+++ b/pyanaconda/modules/timezone/timezone.py
@@ -17,6 +17,9 @@
 # License and may only be used or replicated with the express permission of
 # Red Hat, Inc.
 #
+
+import datetime
+
 from pykickstart.errors import KickstartParseError
 
 from pyanaconda.core.i18n import _
@@ -36,6 +39,7 @@ from pyanaconda.modules.timezone.installation import ConfigureHardwareClockTask,
     ConfigureNTPTask, ConfigureTimezoneTask
 from pyanaconda.modules.timezone.kickstart import TimezoneKickstartSpecification
 from pyanaconda.modules.timezone.timezone_interface import TimezoneInterface
+from pyanaconda.timezone import get_timezone, set_system_date_time, get_all_regions_and_timezones
 
 from pyanaconda.anaconda_loggers import get_module_logger
 log = get_module_logger(__name__)
@@ -167,6 +171,20 @@ class TimezoneService(KickstartService):
         self.timezone_changed.emit()
         log.debug("Timezone is set to %s.", timezone)
 
+    def get_timezones(self):
+        """Get all valid timezones.
+
+        :return: list of valid timezones
+        :rtype: list of str
+        """
+        timezone_dict = get_all_regions_and_timezones()
+        # convert to a dict of lists for easier transfer over DBus
+        # - change the nested sets to lists
+        new_timezone_dict = {}
+        for region in timezone_dict:
+            new_timezone_dict[region] = list(timezone_dict[region])
+        return new_timezone_dict
+
     @property
     def is_utc(self):
         """Is the hardware clock set to UTC?"""
@@ -259,3 +277,30 @@ class TimezoneService(KickstartService):
         :return GeolocationData: result of the lookup, empty if not ready yet
         """
         return self._geoloc_result
+
+    def get_system_date_time(self):
+        """Get system time as a ISO 8601 formatted string.
+
+        :return: system time as ISO 8601 formatted string
+        :rtype: str
+        """
+        # convert to the expected tzinfo format via get_timezone()
+        return datetime.datetime.now(get_timezone(self._timezone)).isoformat()
+
+    def set_system_date_time(self, date_time_spec):
+        """Set system time based on a ISO 8601 formatted string.
+
+        :param str date_time_spec: ISO 8601 time specification to use
+        """
+        log.debug("Setting system time to: %s, with timezone: %s", date_time_spec, self._timezone)
+        # first convert the ISO 8601 time string to a Python date object
+        date = datetime.datetime.fromisoformat(date_time_spec)
+        # set the date to the system
+        set_system_date_time(
+            year=date.year,
+            month=date.month,
+            day=date.day,
+            hour=date.hour,
+            minute=date.minute,
+            tz=self._timezone
+        )
--- a/pyanaconda/modules/timezone/timezone_interface.py
+++ b/pyanaconda/modules/timezone/timezone_interface.py
@@ -79,6 +79,15 @@ class TimezoneInterface(KickstartModuleInterface):
         """
         self.implementation.set_timezone_with_priority(timezone, priority)
 
+    def GetTimezones(self) -> Dict[Str, List[Str]]:
+        """Get valid timezones.
+
+        Return a dictionary, where keys are region ids and values are lists of timezone names in the region.
+
+        :return: a dictionary of timezone lists per region
+        """
+        return self.implementation.get_timezones()
+
     @property
     def IsUTC(self) -> Bool:
         """Is the hardware clock set to UTC?
@@ -157,3 +166,17 @@ class TimezoneInterface(KickstartModuleInterface):
         return GeolocationData.to_structure(
             self.implementation.geolocation_result
         )
+
+    def GetSystemDateTime(self) -> Str:
+        """Get the current local date and time of the system.
+        The timezone set via the Timezone property affects the returned data.
+        :return: a string representing the date and time in ISO 8601 format
+        """
+        return self.implementation.get_system_date_time()
+
+    def SetSystemDateTime(self, date_time_spec: Str):
+        """Set the current local date and time of the system.
+        The timezone set via the Timezone property will be applied to the received data.
+        :param date_time_spec: a string representing the date and time in ISO 8601 format
+        """
+        self.implementation.set_system_date_time(date_time_spec)
--- a/ui/webui/src/apis/timezone.js
+++ b/ui/webui/src/apis/timezone.js
@@ -1,0 +1,141 @@
+/*
+ * Copyright (C) 2023 Red Hat, Inc.
+ *
+ * This program is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation; either version 2.1 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful, but
+ * WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with This program; If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import cockpit from "cockpit";
+
+export class TimezoneClient {
+    constructor (address) {
+        if (TimezoneClient.instance) {
+            return TimezoneClient.instance;
+        }
+        TimezoneClient.instance = this;
+
+        this.client = cockpit.dbus(
+            "org.fedoraproject.Anaconda.Modules.Timezone",
+            { superuser: "try", bus: "none", address }
+        );
+    }
+
+    init () {
+        this.client.addEventListener("close", () => console.error("Timezone client closed"));
+    }
+}
+
+/**
+ * @returns {Promise}           The current system Timezone
+ */
+export const getTimezone = () => {
+    return (
+        new TimezoneClient().client.call(
+            "/org/fedoraproject/Anaconda/Modules/Timezone",
+            "org.freedesktop.DBus.Properties",
+            "Get",
+            [
+                "org.fedoraproject.Anaconda.Modules.Timezone",
+                "Timezone"
+            ]
+        )
+                .then(res => res[0].v)
+    );
+};
+
+/**
+ * @param {string} timezone         Timezone id
+ */
+export const setTimezone = ({ timezone }) => {
+    return new TimezoneClient().client.call(
+        "/org/fedoraproject/Anaconda/Modules/Timezone",
+        "org.freedesktop.DBus.Properties",
+        "Set",
+        [
+            "org.fedoraproject.Anaconda.Modules.Timezone",
+            "Timezone",
+            cockpit.variant("s", timezone)
+        ]
+    );
+};
+
+/**
+ * @returns {Promise}           Resolves a list of timezones
+ */
+export const getTimezones = () => {
+    return new TimezoneClient().client.call(
+        "/org/fedoraproject/Anaconda/Modules/Timezone",
+        "org.fedoraproject.Anaconda.Modules.Timezone",
+        "GetTimezones", []
+    );
+};
+
+/**
+ * @returns {Promise}           NTP enabled
+ */
+export const getNtpEnabled = () => {
+    return (
+        new TimezoneClient().client.call(
+            "/org/fedoraproject/Anaconda/Modules/Timezone",
+            "org.freedesktop.DBus.Properties",
+            "Get",
+            ["org.fedoraproject.Anaconda.Modules.Timezone", "NTPEnabled"]
+        )
+                .then(res => res[0].v)
+    );
+};
+
+/**
+ * @param {bool} enabled - enable/disable NTP
+ *
+ * @returns {Promise}           FIXME: what does it return in this case ??
+ */
+
+export const setNtpEnabled = ({ enabled }) => {
+    return (
+        new TimezoneClient().client.call(
+            "/org/fedoraproject/Anaconda/Modules/Timezone",
+            "org.freedesktop.DBus.Properties",
+            "Set",
+            [
+                "org.fedoraproject.Anaconda.Modules.Timezone",
+                "NTPEnabled",
+                cockpit.variant("b", enabled)
+            ]
+        )
+    );
+};
+
+/**
+ * @returns {Promise}           Resolves the DBus path to the partitioning
+ */
+export const getSystemDateTime = () => {
+    return new TimezoneClient().client.call(
+        "/org/fedoraproject/Anaconda/Modules/Timezone",
+        "org.fedoraproject.Anaconda.Modules.Timezone",
+        "GetSystemDateTime", []
+    );
+};
+
+/**
+ * @param {string} datetimespec date time specification in ISO 8601 format
+ *
+ * @returns {Promise}           FIXME: what does it ret
+ */
+export const setSystemDateTime = ({ datetimespec }) => {
+    return new TimezoneClient().client.call(
+        "/org/fedoraproject/Anaconda/Modules/Timezone",
+        "org.fedoraproject.Anaconda.Modules.Timezone",
+        "SetSystemDateTime", [datetimespec]
+    );
+};
--- a/ui/webui/src/components/time/DateAndTime.jsx
+++ b/ui/webui/src/components/time/DateAndTime.jsx
@@ -1,0 +1,500 @@
+/*
+ * Copyright (C) 2023 Red Hat, Inc.
+ *
+ * This program is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation; either version 2.1 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful, but
+ * WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with This program; If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import React, { useState, useEffect } from "react";
+
+import {
+    ActionGroup,
+    Button,
+    Card,
+    CardTitle,
+    CardBody,
+    DatePicker,
+    Divider,
+    Flex,
+    FlexItem,
+    Form,
+    FormGroup,
+    FormSection,
+    Grid,
+    GridItem,
+    Text,
+    TextContent,
+    TextVariants,
+    TimePicker,
+    Popover,
+    PopoverPosition,
+    Icon,
+    Label,
+    Switch,
+    Select,
+    SelectOption,
+    SelectVariant,
+} from "@patternfly/react-core";
+
+import {
+    InfoCircleIcon,
+    HelpIcon,
+} from "@patternfly/react-icons";
+
+import cockpit from "cockpit";
+
+import * as timeformat from "timeformat.js";
+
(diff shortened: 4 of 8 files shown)


pyanaconda/modules/timezone/timezone.py
```diff
@@ -167,6 +175,26 @@ def set_timezone_with_priority(self, timezone, priority):
...
         log.debug("Timezone is set to %s.", timezone)
 
+    def get_timezones(self):
+        """Get all valid timezones.
+
+        :return: list of valid timezones
+        :rtype: list of str
+        """
+        task = GetTimezonesTask()
+        timezone_dict = task.run()
```
VladimirSlavik, line 185: You don't need the task, this does not take time. Just call `get_all_regions_and_timezones()` here.

pyanaconda/modules/timezone/timezone.py
```diff
@@ -167,6 +175,26 @@ def set_timezone_with_priority(self, timezone, priority):
...
+        # FIXME: remove
+        log.debug("ALL TIMEZONES")
+        log.debug(timezone_dict)
+        # convert to a dict of lists for easier transfer over DBus
+        # - change the nested sets to lists
+        new_timezone_dict = {}
+        for region in timezone_dict:
+            new_timezone_dict[region] = list(timezone_dict[region])
```
VladimirSlavik, line 193: I remember this problem. Nice to see it solved!

pyanaconda/modules/timezone/timezone.py
```diff
@@ -259,3 +293,28 @@ def geolocation_result(self):
...
+        # FIXME: remove
+        log.debug("AAA TIME - CURRENT TIMEZONE")
+        log.debug(self._timezone)
+        # FIXME: check if this is correct to do ?
+        if self._timezone is None:
+            timezone = "America/New_York"
+        else:
+            timezone = self._timezone
+        task = GetSystemDateTimeTask(get_timezone(timezone))
+        return task.run()
```
VladimirSlavik, line 312: Same here. The getting is fast.

pyanaconda/modules/timezone/timezone.py
```diff
@@ -259,3 +293,28 @@ def geolocation_result(self):
...
+
+    def set_system_date_time(self, date_time_spec):
+        """Set system time based on a ISO 8601 formatted string.
+
+        :param str date_time_spec: ISO 8601 time specification to use
+        """
+        task = SetSystemDateTimeTask(date_time_spec=date_time_spec, timezone=self._timezone)
+        task.run()
```
VladimirSlavik, line 320: The setting might take longer, but I think it's fine without the task too. Even so, this would not be the way to do it: You just run the task here synchronously, so there is no benefit.

VladimirSlavik (commented):
I've looked only at the backend. Mostly LGTM. Good news! You don't need any of these tasks.
- These that are implemented, you run them all synchronously and they are very short, so it's all pointless. The original GTK GUI usage is synchronous as well and it's good enough. Just take the task's `run()` and put it where you call it.
- I'd consider perhaps the time setting as taskifiable, because that runs a program, but then you should return the task via D-Bus, not run it in the backend. My first instinct would be to not use a task here though.
- If you're not implementing the NTP status part, please don't add any of the code, just have the frontend with the control disabled or something. Unless that will be part of this PR, too? In that case it's too early to say.

ui/webui/src/components/time/DateAndTime.jsx
```diff
@@ -0,0 +1,476 @@
...
+              setCurrentSelection(selection);
+              // then close the Select
+              setIsOpenCitySelect(false);
+          }}
+          isOpen={isOpenCitySelect}
+          variant={SelectVariant.typeahead}
+          placeholderText={_("City")}
+          selections={currentSelection}
+        >
+            {}
```
rvykydal, line 190: Some leftover?

ui/webui/src/components/time/DateAndTime.jsx
```diff
@@ -0,0 +1,476 @@
...
+const EditSettingsCard = ({
+    idPrefix,
+    setEditModeEnabled,
+    automaticTimeEnabled,
+    setAutomaticTimeEnabled,
+    timezones,
+    regions,
+    selectedRegion,
+    setSelectedRegion,
+    regionSelected,
```
rvykydal, line 207: Perhaps not wired in RegionSelect yet (or should selectedRegion be used instead below in RegionSelect?)

rvykydal, line 207: regionSelected does not seem to be used anywhere, maybe selectedRegion should be passed to RegionSelect component below?
Ah, using selectedRegion and selectedCity in RegionSelect and CitySelect are in TODO...

rvykydal:
Looks good to me so far.

ui/webui/src/components/time/DateAndTime.jsx
```diff
@@ -0,0 +1,490 @@
...
+}) => {
+    const [selectedDate, setSelectedDate] = useState(null);
+    const [selectedTime, setSelectedTime] = useState(null);
+    return (
+        <Card>
+            <CardBody>
+                <Flex direction={{ default: "column" }} spacer={{ default: "spaceItemsXs" }}>
+                    <FlexItem>
+                        <Text className="time-edit-heading">
+                            {_("Timezone")}
```
KKoukiou, line 223: Timezone should be CardTitle like 'Current settings"

ui/webui/src/components/time/DateAndTime.jsx
```diff
@@ -0,0 +1,490 @@
...
+                        </Text>
+                    </FlexItem>
+                    <FlexItem spacer={{ default: "spacerXs" }} />
+                    <FlexItem>
+                        <Flex direction={{ default: "row" }} alignSelf={{ default: "alignItemsCenter" }}>
+                            <Flex direction={{ default: "column" }}>
+                                <FlexItem>
+                                    <Text className="time-edit-subheading">
```
KKoukiou, line 231: This should be all reimplemented to be a form. See https://www.patternfly.org/v4/components/form/#grid-form

KKoukiou (commented):
Please reimplement the edit timezone/time view to be a form. This will massive change the code so I will continue the review from there.

M4rtinK:
Updated the PR - now the edit card is `Form` based and I was able to reduce the complexity somewhat. Main issue right now it that apparently calling the DBus API from the Save button handler does not seem to actually call any DBus methods. :P

ui/webui/src/components/time/DateAndTime.jsx
M4rtinK: This has been removed in the end. :)

ui/webui/src/components/time/DateAndTime.jsx
M4rtinK: Should be resolved now. :)

ui/webui/src/components/time/DateAndTime.jsx
M4rtinK: The code is using the form layout now. :)

M4rtinK:
Latest screenshots:
Screenshot 2023-06-21 at 15-50-11 Anaconda Web UI
Screenshot 2023-06-21 at 15-50-41 Anaconda Web UI

pyanaconda/modules/timezone/timezone_interface.py
```diff
@@ -79,6 +79,15 @@ def SetTimezoneWithPriority(self, timezone: Str, priority: UInt16):
         """
         self.implementation.set_timezone_with_priority(timezone, priority)
 
+    def GetTimezones(self) -> Dict[Str, List[Str]]:
+        """Get valid timezones.
+
+        Return a dictionary, where keys are region ids and values are lists of timezone names in the region.
+
+        :return: a list of timezones
```
poncovka, line 87: The return type is wrong now, because you have changed the proposed API. Pinging @VladimirSlavik to look at it.

pyanaconda/modules/timezone/timezone.py
```diff
@@ -259,3 +277,35 @@ def geolocation_result(self):
...
+        :return: system time as ISO 8601 formatted string
+        :rtype: str
+        """
+        # FIXME: check if this is correct to do ?
+        if self._timezone is None:
+            timezone = "America/New_York"
+        else:
+            timezone = self._timezone
+        # convert to the expected tzinfo format via get_timezone()
+        return datetime.datetime.now(get_timezone(timezone)).isoformat()
```
poncovka, line 293: If the timezone is invalid, this will cause an error. Maybe the timezone setter should check the validity of the new timezone and reject it if it is not valid to make sure that the these calls always succeed. @VladimirSlavik, what do you think?

poncovka (commented):
It would be nice to see some tests for the DBus backend to know that I am not reviewing a broken code. Also, the DBus backend can be easily handled in a separate pull request. There is even a separate Jira issue for this.

pyanaconda/modules/timezone/timezone.py
```diff
@@ -259,3 +293,28 @@ def geolocation_result(self):
...
+        """Get system time as a ISO 8601 formatted string.
+
+        :return: system time as ISO 8601 formatted string
+        :rtype: str
+        """
+        # FIXME: remove
+        log.debug("AAA TIME - CURRENT TIMEZONE")
+        log.debug(self._timezone)
+        # FIXME: check if this is correct to do ?
+        if self._timezone is None:
```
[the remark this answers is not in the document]
M4rtinK: Perfect - so I will drop it. Thanks! :)

pyanaconda/modules/timezone/timezone.py
```diff
@@ -259,3 +293,28 @@ def geolocation_result(self):
...
+            timezone = "America/New_York"
```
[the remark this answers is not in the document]
M4rtinK: ACK. :)

pyanaconda/modules/timezone/timezone.py
VladimirSlavik: Setter rejecting invalid values is a possibility, and probably even a good idea. However, there's more to consider.

Timezone is taken verbatim from kickstart, so there's always a way to sneak in something invalid. That's why I didn't make the setter check before, IIRC. I wonder if fallback-ifying invalid timezone in kickstart would be an option. Then the check would be somewhere in `TimezoneService.set_timezone_with_priority()` and apply to everything.

Theoretically, the install env should have the same data as the payload, so timezone validity should be identical. Additionally, this is slow-moving data defined by a standard, and the actual changes are mostly in the time rules, not zone names. So I think that approach might be fine.

I guess this would be more of a Steve B. topic, but anyway, any thoughts @jstodola?

(Enforcing timezone validity in kickstart should have a release note though, so that we don't forget about it.)

pyanaconda/modules/timezone/timezone_interface.py
VladimirSlavik: IMO this isn't a big problem, actually. The data needs to come from somewhere anyway, in some form. If we already collate this in the backend, it might be advantageous to just use it.

It's true the original suggestion is more UI agnostic, though...

pyanaconda/modules/timezone/timezone.py
VladimirSlavik: Ah, I just realized something. The only way to set time was on time spoke. So this probably worked such that the kickstart data was taken as-is, and then if the spoke was actually entered, the bad zone was fixed on entry so there was no problem setting the time.

(Past tense because the changes around welcome screen, geolocation etc. might have removed this.)

ui/webui/src/components/time/DateAndTime.jsx
```diff
@@ -0,0 +1,505 @@
...
+            </FlexItem>
+            <FlexItem>
+                <Text className="automatic-time-label">
+                    {automaticTimeEnabled ? _("Automatic time is on.") : _("Automatic time is off.")}
+                </Text>
+            </FlexItem>
+            <FlexItem>
+                <Popover
+                  bodyContent={_(
+                      "Synchronize your device with Network Time Protocol (NTP) server. " +
```
KKoukiou, line 393: This imperative text sounds like the user should perform some action when they should not.

ui/webui/src/components/time/DateAndTime.jsx
```diff
@@ -0,0 +1,505 @@
...
+            </CardBody>
+        </Card>
+    );
+};
+
+const AutomaticTimeNotification = ({ automaticTimeEnabled }) => {
+    return (
+        <Flex direction={{ default: "row" }} spaceItems={{ default: "spaceItemsSm" }} alignSelf={{ default: "alignItemsCenter" }}>
+            <FlexItem>
+                <Icon status="info">
```
KKoukiou, line 381: The mockup is using an info alert.

<Alert variant="info" isInline isPlain title="Info inline alert title" />

https://www.patternfly.org/v4/components/alert#plain-inline-alert-variants

ui/webui/src/components/time/DateAndTime.jsx
```diff
@@ -0,0 +1,505 @@
...
+                    <InfoCircleIcon />
+                </Icon>
+            </FlexItem>
+            <FlexItem>
+                <Text className="automatic-time-label">
+                    {automaticTimeEnabled ? _("Automatic time is on.") : _("Automatic time is off.")}
+                </Text>
+            </FlexItem>
+            <FlexItem>
+                <Popover
```
KKoukiou, line 391: The helper icon has somehow different color in the mockup. Looks like more like the variant from https://www.patternfly.org/v4/components/form#basic

Try copying that one.

ui/webui/src/components/time/DateAndTime.jsx
```diff
@@ -0,0 +1,505 @@
...
+                      <Text component={TextVariants.p}>
+                          {_("Timezone")}
+                      </Text>
+                  </Flex>
+                  <Flex spaceItems={{ default: "spaceItemsSm" }}>
+                      <Text component={TextVariants.p}>
+                          {getTimezoneName(timezoneRegion, timezoneCity)}
+                      </Text>
+                      <Text component={TextVariants.p}>
+                          {_("pretty region name (pretty timezone name)")}
```
KKoukiou, line 94: These are still unfinished. Commenting so that we don't forget.

ui/webui/src/components/time/DateAndTime.jsx
```diff
@@ -0,0 +1,505 @@
...
+                            <GridItem span={4}>
+                                <FormGroup
+                                  label={_("Automatic time")}
+                                  labelIcon={
+                                      <Popover
+                                        bodyContent={_(
+                                            "To edit date and time, turn off automatic time."
+                                        )}
+                                      >
+                                          <Icon iconSize="sm">
```
KKoukiou, line 300: Please copy the info code from the form example. This code you wrote - looks correct but it produces some really small icon with wrong color.

pyanaconda/modules/timezone/timezone.py
```diff
@@ -167,6 +175,26 @@ def set_timezone_with_priority(self, timezone, priority):
         self.timezone_changed.emit()
         log.debug("Timezone is set to %s.", timezone)
 
+    def get_timezones(self):
```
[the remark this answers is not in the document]
M4rtinK: Unit tests are now in place. :)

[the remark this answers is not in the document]
M4rtinK: (for the backend code so far)

ui/webui/src/apis/timezone.js
```diff
@@ -0,0 +1,156 @@
...
+        "/org/fedoraproject/Anaconda/Modules/Timezone",
+        "org.fedoraproject.Anaconda.Modules.Timezone",
+        "GetSystemDateTime", []
+    );
+};
+
+/**
+ * @param {string} datetimespec date time specification in ISO 8601 format
+ *
+ * @returns {Promise}           FIXME: what does it return in this case ??
```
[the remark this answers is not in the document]
M4rtinK: OK, interesting & makes sense. Lets drop the FIXME. :)

ui/webui/src/components/time/DateAndTime.jsx
```diff
@@ -0,0 +1,505 @@
...
+        setTimezoneRegion(region);
+        // clear city selection if region changes to avoid
+        // setting nonsense timezones to the backend
+        setTimezoneCity(null);
+    };
+    const handleCityChange = city => {
+        setCitiesOpen(false);
+        setTimezoneCity(city);
+    };
+    const handleNTPChange = NTPvalue => {
```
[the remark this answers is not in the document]
M4rtinK: Yeah, that indeed looks weird- lets change it. :)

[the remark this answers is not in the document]
M4rtinK: I changed it everywhere when I was at it - that way we also get rid of stuff like `setNTPEnabled` which kinda break camel case.

ui/webui/src/components/time/DateAndTime.jsx
```diff
@@ -0,0 +1,505 @@
...
+        <Card>
+            <CardBody>
+                <Form>
+                    <FormSection title={_("Timezone")}>
+                        <Grid hasGutter>
+                            <GridItem span={6}>
+                                <FormGroup
+                                  label={_("Region")}
+                                  isRequired
+                                  fieldId="time-date-edit-region"
```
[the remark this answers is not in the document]
M4rtinK: Oh, right, changed. :)

ui/webui/src/components/time/DateAndTime.jsx
```diff
@@ -0,0 +1,505 @@
...
+                                >
+                                    <Select
+                                      id={idPrefix + "select-region"}
+                                      variant={SelectVariant.typeahead}
+                                      placeholderText={_("Select region")}
+                                      isOpen={regionsOpen}
+                                      selections={timezoneRegion}
+                                      onToggle={(_, isOpen) => setRegionsOpen(isOpen)}
+                                      onSelect={(event, value) => handleRegionChange(value)}
```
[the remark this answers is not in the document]
M4rtinK: Yep, will do that. BTW, I know its something else, but its a bit unfortunate the "now in use" placeholder looks the same as the regular translation function :P

ui/webui/src/components/time/DateAndTime.jsx
```diff
@@ -0,0 +1,505 @@
...
+                                  fieldId="time-date-edit-city"
+                                >
+                                    <Select
+                                      id={idPrefix + "select-city"}
+                                      variant={SelectVariant.typeahead}
+                                      placeholderText={_("Select city")}
+                                      isOpen={citiesOpen}
+                                      selections={timezoneCity}
+                                      onToggle={(_, isOpen) => setCitiesOpen(isOpen)}
+                                      onSelect={(event, value) => handleCityChange(value)}
```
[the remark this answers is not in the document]
M4rtinK: Fixed. :)

ui/webui/src/components/time/DateAndTime.jsx
```diff
@@ -0,0 +1,505 @@
...
+                                          </Icon>
+                                      </Popover>
+                                  }
+                                  isRequired
+                                  fieldId="time-date-edit-NTP-enabled"
+                                >
+                                    <Switch
+                                      label={_("On")}
+                                      labelOff={_("Off")}
+                                      id={idPrefix + "switch-automatic-time"}
```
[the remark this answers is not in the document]
M4rtinK: Oh, right, fixed.

ui/webui/src/components/time/DateAndTime.jsx
```diff
@@ -0,0 +1,505 @@
...
+                                          value={timeformat.date(systemDate.valueOf())}
+                                          onChange={handleDateChange}
+                                        />
+                                    </FormGroup>
+                                    <FormGroup
+                                      label={_("Time")}
+                                      isRequired
+                                      fieldId="time-date-edit-time"
+                                    >
+                                        <TimePicker
```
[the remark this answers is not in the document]
M4rtinK: Added.

ui/webui/src/components/time/DateAndTime.jsx
```diff
@@ -0,0 +1,505 @@
...
+                          onClick={handleSave}
+                        >
+                            {_("Save")}
+                        </Button>
+                        <Button
+                          id={idPrefix + "-btn-cancel"}
+                          variant="link"
+                          onClick={handleCancel}
+                        >
+                            Cancel
```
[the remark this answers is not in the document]
M4rtinK: Indeed! Fixed.

ui/webui/src/components/time/DateAndTime.jsx
```diff
@@ -0,0 +1,505 @@
...
+    // system time
+    const [systemDate, setSystemDate] = useState("");
+    const [systemDateString, setSystemDateString] = useState("");
+    const [systemTimeString, setSystemTimeString] = useState("");
+
+    // fetch timezone data
+    useEffect(() => {
+        const fetchTimezoneData = async (setTimezones, setTimezoneRegion, setTimezoneCity) => {
+            // fetch timezone listing
+            const res = await getTimezones();
```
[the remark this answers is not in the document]
M4rtinK: Adding a simple `.catch(console.error)` for now, like we do elsewhere.

ui/webui/src/components/time/DateAndTime.jsx
```diff
@@ -0,0 +1,505 @@
...
+            const timezoneDict = res[0];
+            setTimezones(timezoneDict);
+            // fetch current timezone
+            const currentTimezone = await getTimezone();
```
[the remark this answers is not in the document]
M4rtinK: Here as well.

ui/webui/src/components/time/DateAndTime.scss
```diff
@@ -0,0 +1,22 @@
...
+}
+.time-edit-subheading {
+    font-size: var(--pf-global--FontSize--sm);
+    font-weight: var(--pf-global--FontWeight--bold);
+}
+.time-edit-automatic-hint {
+    font-size: var(--pf-global--FontSize--sm);
+    color: var(--pf-global--Color--dark-200);
+}
+
```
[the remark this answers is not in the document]
M4rtinK: Cleaned up. :)

ui/webui/src/components/time/DateAndTime.jsx
M4rtinK: Oh, right - looking at the mockup it says "Synchronizes your device with Network Time Protocol (NTP) server." - I must have typoed that. Lets fix it. :)

PR #4813, opened 2023-06-06
Import BlockDev from blivet instead of gi

vojtechtrefny:
We are planning to release a new major version of libblockdev soon and because version needs to be specified when importing from gi it would make the import unnecessarily complicated if we wanted to keep it backwards compatible. Anaconda is already using blivet so importing BlockDev from it instead of gi shouldn't change anything.

### DIFF
--- a/pyanaconda/modules/storage/checker/utils.py
+++ b/pyanaconda/modules/storage/checker/utils.py
@@ -15,13 +15,9 @@
 # License and may only be used or replicated with the express permission of
 # Red Hat, Inc.
 #
-import gi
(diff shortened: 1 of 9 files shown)


VladimirSlavik (approved):
Thanks!

vojtechtrefny:
Can this be merged? We'd like to release libblockdev 3.0 in ~2 weeks so we need this in Fedora before this (I'd like to avoid adding anaconda to the chain build).

vojtechtrefny:
Thanks!

PR #4815, opened 2023-06-06
Add easy way to build ISO from Anaconda

jkonecny12:
This is using our existing containers to build the ISO with a make call. It has to run as sudo to work because of the loop devices used by Lorax.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -325,6 +325,28 @@ container-rpms:
 container-rpms-scratch:
 	$(MAKE) -f ./Makefile.am CONTAINER_ADD_ARGS="-e TEST_BUILD=true" container-rpms
 
+# Build ISO from Anaconda RPM files build here
+container-iso-build:
+	@sudo -nv 2>/dev/null || echo "You will be prompted for sudo password because of loop device mounting in Lorax!"
+
+	@if ! ls $(srcdir)/result/build/01-rpm-build/anaconda-*.rpm >/dev/null 2>/dev/null; then \
+		echo "You need to have anaconda RPMs build first. Please run 'make -f ./Makefile.am container-rpms-scratch' before this command."; \
+		exit 1; \
+	fi
+
+	mkdir -p result/iso
+	sudo $(CONTAINER_ENGINE) run \
+	--rm --privileged  --tmpfs /var/tmp:rw,mode=1777 \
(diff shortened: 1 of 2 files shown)


M4rtinK (approved):
Looks good to me, should be very useful, thanks! :)

Makefile.am
```diff
@@ -325,6 +325,23 @@ container-rpms:
 container-rpms-scratch:
 	$(MAKE) -f ./Makefile.am CONTAINER_ADD_ARGS="-e TEST_BUILD=true" container-rpms
 
+# Build ISO from Anaconda RPM files build here
+container-iso-build-scratch:
+	@echo "To build an ISO we need to have root privilidges because of loop device mounting in Lorax!"
+	@echo "You need to have anaconda RPMs build first. Please run 'make -f ./Makefile.am container-rpms-scratch' before this command."
```
VladimirSlavik, line 331: Why don't you actually test for this? I'd imagine something like `if ! sudo -v ; then echo "..."; exit 1; fi` and for the RPMs something like `if compgen -E "./result/build/01-rpm-build/*.rpm" > /dev/null ; then echo "..."; exit 1; fi`.

M4rtinK: Indeed, that looks like it should not be that hard to do & could make the whole thing much more user friendly. :)

jkonecny12: I can definitely add that but not sure if this makes sense to complicate just the `echo` call?

If you insist on this I'm fine with adding the logic but right now it's seems to me just as more complication (lower readability) for not a big benefit.

jkonecny12:
**UPDATED:**
* Rename the targets to (remove -scratch)
* Add checks for the echo prints

Makefile.am
jkonecny12: I added the checks.

VladimirSlavik: Thank you, but I would like also the `exit 1` part, so that we don't have to hunt for these lines on top of error messages from lorax in podman. There are clear prerequisites, you can technically check for them, so why not do it? Just telling the user there will be errors is pointless if you can also stop doing the thing that won't work.

jkonecny12: I'll add `exit 1` for missing RPMs but it doesn't give sense for the `We will ask for password` message.

VladimirSlavik (commented):
Thank you!

VladimirSlavik (approved):

PR #4816, opened 2023-06-07
Revert "Download cockpit rpms during build"

KKoukiou:
This reverts commit 326cda0817ed2521716db6d1bc3dd90c165bfb23.

This commit that I revert here was hindering reproducability of the updates.img.

We need to get the same version of anaconda-webui dependencies RPMs every time, instead of these randomly/confusingly changing just because a new update passed bodhi.

The build-rpms script will likely be obsolete soon, when anaconda will start building their own ISO files, which will contains the anaconda-webui deps. Till then let's have:

* The dependencies gated
* The downloaded rpms cached

The only downside of the change is that now we need to manually update the versions of the deps, but it's much safer to gate this and the only guarantee for a green CI.

### DIFF
--- a/ui/webui/test/prepare-updates-img
+++ b/ui/webui/test/prepare-updates-img
@@ -2,6 +2,25 @@
 
 set -eu
 
+# FIXME boot.iso on rawhide does not currently contain the new cockpit dependencies
+# This will change once we include this changes upstream and start building boot.iso with the new dependencies
+# Then we can safely remove this workaround
+ANACONDA_WEBUI_DEPS_URLS='
+    https://kojipkgs.fedoraproject.org//packages/cockpit/292/1.fc39/x86_64/cockpit-ws-292-1.fc39.x86_64.rpm
+    https://kojipkgs.fedoraproject.org//packages/cockpit/292/1.fc39/x86_64/cockpit-bridge-292-1.fc39.x86_64.rpm
+    https://kojipkgs.fedoraproject.org/packages/firefox/114.0/1.fc39/x86_64/firefox-114.0-1.fc39.x86_64.rpm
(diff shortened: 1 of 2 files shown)


ui/webui/test/prepare-updates-img
```diff
@@ -2,18 +2,37 @@
 
 set -eu
 
+# FIXME boot.iso on rawhide does not currently contain the new cockpit dependencies
+# This will change once we include this changes upstream and start building boot.iso with the new dependencies
+# Then we can safely remove this workaround
+ANACONDA_WEBUI_DEPS_URLS='
+    https://kojipkgs.fedoraproject.org//packages/cockpit/292/1.fc39/x86_64/cockpit-ws-292-1.fc39.x86_64.rpm
```
M4rtinK, line 9: So the reason for no longer getting these RPMs from direct links was that the links eventually go dead and everything breaks. Are we sure this change is temporary enough so this will not be an issue or perhaps some other solution will be better ?

Maybe we can download the PRMs to a folder on fedorapeople and fetch from there ?

M4rtinK (approved):
I don't really like reverting back to the direct links (that will eventually go dead), but if this is really temporary, then I'm fine with it in its current form.

jkonecny12 (changes requested):
Hi all.

In general I'm seeing this as an issue. This was implemented because we've got to the place that the the rpms were so outdated that they were removed from the koji. I think it's better to go with Rawhide than with having something to get outdated.

In other words, I'm happy to merge this but we need to somehow solve periodic updates. Otherwise I don't think this is improvement of the current state.

jkonecny12:
I'm not aware of such a thing. However, true is that Firefox worked when you tried it in the past and now when we want to switch to it, it does not work. In other words, that would be a case if we did the switch before.

PR #4817, opened 2023-06-07
webui: enable linter for webui CI runs

KKoukiou:


jstodola:
Shouldn't be `ui/webui/src/index.js` changes in a separate commit?

jstodola:
I should have provided a reasoning why I think that separate commits are better:
When linter for webui CI runs needs to be disabled for whatever reason in the future, it would not be possible to do it by reverting a single commit that enabled it without re-introducing the issues found in `ui/webui/src/index.js`.

KKoukiou:
> I should have provided a reasoning why I think that separate commits are better: When linter for webui CI runs needs to be disabled for whatever reason in the future, it would not be possible to do it by reverting a single commit that enabled it without re-introducing the issues found in `ui/webui/src/index.js`.

But there is not  scenario where we would want to disable a linter in CI runs. linters are an extra tool to ensure code quality.

jkonecny12 (approved):

M4rtinK (approved):
Looks good to me - but please split the fixups to a separate commit. Thanks! :)

PR #4820, opened 2023-06-08
Suggest cockpit toolbox for the developent of the webui

KKoukiou:


jkonecny12 (approved):
Looks great to me otherwise. Just we should be consistent and call it `toolbx`.

VladimirSlavik (approved):
LGTM, thank you!

jkonecny12 (approved):
LGTM!

PR #4821, opened 2023-06-08
WebUI: Remove centering from loading screen

OndrejZobal:
Remove centering from the loading screen that appears after finishing installation destination as per [PF design guidelines.](https://www.patternfly.org/v4/components/empty-state/design-guidelines#full-page-empty-states).

obrazek
*pardon my French*

### DIFF
--- a/ui/webui/src/components/storage/DiskEncryption.jsx
+++ b/ui/webui/src/components/storage/DiskEncryption.jsx
@@ -19,7 +19,6 @@ import cockpit from "cockpit";
 import React, { useState, useEffect } from "react";
 
 import {
-    Bullseye,
     Button,
     Checkbox,
     EmptyState,
@@ -206,19 +205,17 @@ const getRuleLength = (password) => {
 const getRuleConfirmMatches = (password, confirm) => (password.length > 0 ? (password === confirm ? "success" : "error") : "indeterminate");
 
 const CheckDisksSpinner = (
-    <Bullseye>
-        <EmptyState id="installation-destination-next-spinner">
-            <EmptyStateIcon variant="container" component={Spinner} />
-            <Title size="lg" headingLevel="h4">
-                {_("Checking storage configuration")}
-            </Title>
-            <TextContent>
-                <Text component={TextVariants.p}>
-                    {_("This may take a moment")}
-                </Text>
-            </TextContent>
-        </EmptyState>
-    </Bullseye>
+    <EmptyState id="installation-destination-next-spinner">
+        <EmptyStateIcon variant="container" component={Spinner} />
+        <Title size="lg" headingLevel="h4">
+            {_("Checking storage configuration")}
+        </Title>
+        <TextContent>
+            <Text component={TextVariants.p}>
+                {_("This may take a moment")}
+            </Text>
+        </TextContent>
+    </EmptyState>
 );
 
 export const DiskEncryption = ({


ui/webui/src/components/storage/DiskEncryption.jsx
```diff
@@ -206,19 +208,17 @@ const getRuleLength = (password) => {
 const getRuleConfirmMatches = (password, confirm) => (password.length > 0 ? (password === confirm ? "success" : "error") : "indeterminate");
 
 const CheckDisksSpinner = (
-    <Bullseye>
```
KKoukiou, line 209: You need to remove also the redundant import now :
22:5  error  'Bullseye' is defined but never used  no-unused-vars

OndrejZobal: Thanks!

KKoukiou (commented):
So although I see it's written in the guidelines using px is very much an antipattern. I am fine with merging this as is, but ideally, I would like to reach out to mcoker in patternfly team and ask what's the suggested layout component to use to create this padding on top. 

If not components available, we should generally use rem or em units for spacing.

KKoukiou (approved):

ui/webui/src/components/storage/DiskEncryption.scss
```diff
@@ -0,0 +1,5 @@
+#installation-destination-next-spinner {
```
KKoukiou, line 1: So i tool a better look and realized this is not correct. Generally it's not adviced to add custom CSS, we should rely on patternfly paddings and use layout components from spacing. 

In this case, we should try to follow https://www.patternfly.org/v4/components/empty-state/design-guidelines/#empty-states-inside-tables-cards-or-wizards, so we want 24px padding on the top. 

The empty state component has default top padding 32 px, and the wizard main body has default padding 24 px. So we are currently, without your overrides and by removing the Bullseye at 56. It's more than double of the advised padding.

So if you want to be  100% correct, you keep the padding from the wizard body, which is 24px (1.5rem) and you reset to 0 the empty state component's padding.

KKoukiou (changes requested):

ui/webui/src/components/storage/DiskEncryption.scss
OndrejZobal: I feel like the 24px rule is generally meant for elements that are embedded within a larger page and since the wizard is full-screen in our case, I think the full-page rule fits it better.

Thanks for your patience, by the way.

KKoukiou: Sorry for taking long time to reply. I don't have strong opinion between the in-page vs wizard, indeed that's a in-page wizard so that's a bit fuzzy which guidelines to follow. But also it's not so important.

Regardless, we should avoid using hardcoded pixels in CSS stylesheets. Instead we need to rely on the paddings of the existing elements. Removing the Bullseye and keeping the rest of elements gives top padding 56px which is fine I believe.

It's slightly less than the 64px suggestion for in-page empty states, and ,more than the 24px suggestion for the in-wizard empty states. But since we actually have a wizard header, I think it's an acceptable padding for our case.

OndrejZobal: Ok, I removed the hard-coded padding.

KKoukiou (approved):

PR #4824, opened 2023-06-09
Allow/improve displaying passwords in plain text (#2013190)

jstodola:
The root and user spokes already allowed to display the passwords in plain text, but they stayed displayed in plain text after re-entering the spokes. Fix this by always using the invisible characters by default when the entry widget is displayed.

Also allow to display plain text passwords on the subscription and installation source spokes.

Resolves: rhbz#2013190

Ported from https://github.com/rhinstaller/anaconda/pull/4625. Since the source spoke design is slightly different in RHEL-9, the commit for proxy passwords for repositories on the installation source spoke was not possible to easily cherry-pick and was re-implemented. So that's the biggest change comparing to the original PR.


VladimirSlavik (approved):
Looks good to me. Thank you!

M4rtinK (approved):

PR #4825, opened 2023-06-09
Be more indulgent when reclaiming disk space (#2187371)

jstodola:

Resolves: rhbz#2187371

Ported from https://github.com/rhinstaller/anaconda/pull/4546


VladimirSlavik (approved):
Thanks!

jkonecny12 (approved):

PR #4826, opened 2023-06-09
Be more indulgent when reclaiming disk space (#1855371)

jstodola:

Resolves: rhbz#1855371

Ported from https://github.com/rhinstaller/anaconda/pull/4546


VladimirSlavik (approved):
Thank you!

jkonecny12 (approved):

PR #4828, opened 2023-06-12
Fix inst.proxy configuration in dracut to allow stage2 image download

jkonecny12:
Fix proxy is not used when downloading stage2 image.

This fix is handling two cases:

* kickstart has `url --proxy` which should be used also to download stage2 image with proxy
* `inst.proxy` is not used by dracut to download stuff (like stage2 image)
* do not request `inst.` prefix for `proxy` kernel boot argument because that is valid option for Dracut (and maybe others) 

*Resolves: rhbz#2177219*

Backport of: https://github.com/rhinstaller/anaconda/pull/4794

### DIFF
--- a/docs/release-notes/fix-proxy-for-downloading-installation-image.rst
+++ b/docs/release-notes/fix-proxy-for-downloading-installation-image.rst
@@ -1,0 +1,12 @@
+:Type: Kickstart
+:Summary: The installation program now correctly processes the proxy configuration (#2177219)
+
+:Description:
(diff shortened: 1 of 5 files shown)


jkonecny12:
**REBASED** to get rid of the annoying red test (should be resolved but won't be executed here again).

VladimirSlavik:
Is there anything preventing us from merging this? I don't think so?

jkonecny12:
Missing KS test. Otherwise it's ready.

M4rtinK (approved):
Looks good to me. :)

PR #4830, opened 2023-06-12
webui: adding PRETTY_NAME to use in title instead of anaconda generic title

acruzgon:
According to UX, "_Anaconda_" should never show to people to avoid confusion. This PR is to start using `PRETTY_NAME` instead of the generic Anaconda title.

Before:
image

After:
image

### DIFF
--- a/ui/webui/src/components/app.jsx
+++ b/ui/webui/src/components/app.jsx
@@ -35,8 +35,7 @@ import { PayloadsClient } from "../apis/payloads";
 
 import { readBuildstamp, getIsFinal } from "../helpers/betanag.js";
 import { readConf } from "../helpers/conf.js";
-
-const _ = cockpit.gettext;
+import { getOsReleaseByKey } from "../helpers/product.js";
 
 export const Application = () => {
     const [address, setAddress] = useState();
@@ -46,6 +45,7 @@ export const Application = () => {
     const [notifications, setNotifications] = useState({});
     const [isHelpExpanded, setIsHelpExpanded] = useState(false);
     const [helpContent, setHelpContent] = useState("");
+    const [prettyName, setPrettyName] = useState("");
 
     useEffect(() => {
         cockpit.file("/run/anaconda/bus.address").watch(address => {
@@ -68,6 +68,10 @@ export const Application = () => {
         readBuildstamp().then(
             buildstamp => setBeta(!getIsFinal(buildstamp)),
             ex => console.error("Failed to parse anaconda buildstamp file")
+        );
+
+        getOsReleaseByKey("PRETTY_NAME").then(
+            setPrettyName
         );
     }, []);
 
@@ -95,7 +99,7 @@ export const Application = () => {
     }
     console.info("conf: ", conf);
 
-    const title = _("Anaconda installer");
+    const title = cockpit.format("$0 installation", prettyName);
 
     const page = (
         <Page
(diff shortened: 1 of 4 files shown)


ui/webui/src/components/localization/InstallationLanguage.jsx
```diff
@@ -317,9 +319,16 @@ LanguageSelector.contextType = AddressContext;
 export const InstallationLanguage = ({ idPrefix, setIsFormValid, onAddErrorNotification }) => {
     const [nativeName, setNativeName] = React.useState(false);
     const { setLanguage } = React.useContext(LanguageContext);
+    const [prettyName, setIsPrettyName] = useState("");
+
+    useEffect(() => {
+        readPrettyName().then(
```
KKoukiou, line 325: Can be nicely written as one linter. readPrettyName().then(setIsPrettyName);

KKoukiou (changes requested):
Some small fixups needed. Also tests :)

garrett (changes requested):
Thanks for making this change!

There's a minor difference what what I intended: I was thinking of having the full distribution name + version at the top and the welcome to message only have the distribution name (without version).

VladimirSlavik (commented):
1) I'm not sure this is the right place where the data should come from. Take a look at `pyanaconda/product.py` - that loads `/.buildstamp` and environment variables. This should work similarly, in fact we already get `isFinal` from buildstamp in `betanag.js`.

2) I don't think `conf.js` is the right place for the helper. Maybe some reorganization is in order, so that the betanag stuff and anything else reading buildstamp are together?

ui/webui/src/components/app.jsx
```diff
@@ -69,6 +70,10 @@ export const Application = () => {
             buildstamp => setBeta(!getIsFinal(buildstamp)),
             ex => console.error("Failed to parse anaconda buildstamp file")
         );
+
+        readPrettyName().then(
```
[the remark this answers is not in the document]
acruzgon: as talked, this is done by running `npm run eslint:fix`

acruzgon:
@garrett it would like this:
image
is that okay?

ui/webui/src/components/app.jsx
```diff
@@ -69,6 +69,10 @@ export const Application = () => {
             buildstamp => setBeta(!getIsFinal(buildstamp)),
             ex => console.error("Failed to parse anaconda buildstamp file")
         );
+
+        getOsReleaseByKey("PRETTY_NAME").then(
+            setIsPrettyName,
```
KKoukiou, line 74: Please remove comma, this is just not failing because - JS - but it does not belong there.

ui/webui/src/components/localization/InstallationLanguage.jsx
```diff
@@ -317,9 +318,16 @@ LanguageSelector.contextType = AddressContext;
 export const InstallationLanguage = ({ idPrefix, setIsFormValid, onAddErrorNotification }) => {
     const [nativeName, setNativeName] = React.useState(false);
     const { setLanguage } = React.useContext(LanguageContext);
+    const [distributionName, setIsDistributionName] = useState("");
```
KKoukiou, line 321: the setter name needs fix. s/setIsDistributionName/setDistributionName
(remove the Is)

KKoukiou: It's common to use 'Is' as part of the state var if it's boolean. For example:
const [isValid, setIsValid] = useState(true);

KKoukiou (changes requested):

KKoukiou (commented):
Please don't include merge commits and/or empty commits in your PRs.

PR #4835, opened 2023-06-14
Improve btrfs support in tui mountpoint assignment

rvykydal:
This PR adds support for btrfs subvolumes reformatting/clearing in mount point assignment with the outlook of using mount point assignment for /home reuse use case on Fedora Workstation (would require reformatting of boot partitions and 'reformatting' / clearing of the root btrfs subvolume). The support is added to Anaconda but some bits could be perhaps moved to Blivet. The PR is assuming there will be no substantial update of device abstractions to accomodate for btrfs volume / subvolume characteristics in near future.

Possible issues and follow-ups:
1) The support for btrfs volume/subvolume device specification resolution should be moved to Blivet, and proper specification (currently it is UUID=uuid@subvolume) should be agreed on (it is pretty arbitrary now).
2) Maybe we could have 'clear' action support in blivet which would replace the _recreate_device function.
3) The display of subvolumes could be better (including a volume id?)
4) We may want to handle also volumes:
  - only leaf volumes ? or warning with non-leaf volumes ?
  - allow to change format of reformatted volume ?
  - support handling volumes residing on single / multiple devices ?
5) possible follow up improvements in GUI (allow reformatting of btrfs subvolumes in Custom Partitioning GUI: https://bugzilla.redhat.com/show_bug.cgi?id=2186158
6) using existing mount options, like compression (compress=zstd:1)

Screenshot of mount-point assignment with /home preserved:
btrfs_tui_reuse_home

### DIFF
--- a/pyanaconda/modules/storage/devicetree/utils.py
+++ b/pyanaconda/modules/storage/devicetree/utils.py
@@ -290,3 +290,24 @@ def find_unconfigured_luks(storage):
         devices.append(device)
 
     return devices
+
+def resolve_device(storage, dev_spec):
+    """Return the device matching the provided device specification.
+
+    A wrapper of Blivet's resolve_device function supporting also
+    resolution of btrfs volumes specified by UUID of the volume
+    and name of the subvolume, as:
+    'UUID=2252ec30-1fce-4f8e-bdef-c50c3a44ede4@root'
+
+    :param storage: an instance of Blivet
+    :param dev_spec: a string describing a block device
+    :returns: the device
+    :rtype: :class:`~.devices.StorageDevice` or None
+    """
+
+    options = None
+    if dev_spec.startswith("UUID") and "@" in dev_spec:
+        dev_spec, _, subvol = dev_spec.partition("@")
+        options = "subvol={}".format(subvol)
+
+    return storage.devicetree.resolve_device(dev_spec, options=options)
--- a/pyanaconda/modules/storage/devicetree/viewer.py
+++ b/pyanaconda/modules/storage/devicetree/viewer.py
@@ -28,7 +28,7 @@ from pyanaconda.modules.common.errors.storage import UnknownDeviceError
 from pyanaconda.modules.common.structures.storage import DeviceData, DeviceActionData, \
     DeviceFormatData, OSData
 from pyanaconda.modules.storage.devicetree.utils import get_required_device_size, \
-    get_supported_filesystems
+    get_supported_filesystems, resolve_device
 
 log = get_module_logger(__name__)
 
@@ -347,13 +347,16 @@ class DeviceTreeViewer(ABC):
         The spec can be anything from a device name (eg: 'sda3') to a
         device node path (eg: '/dev/mapper/fedora-root') to something
         like 'UUID=xyz-tuv-qrs' or 'LABEL=rootfs'.
+        For btrfs subvolumes the subvolume can be specified by the UUID of
+        the volume and the name of the subvolume, as:
+        'UUID=2252ec30-1fce-4f8e-bdef-c50c3a44ede4@root'
 
         If no device is found, return an empty string.
 
         :param dev_spec: a string describing a block device
         :return: a device name or an empty string
         """
-        device = self.storage.devicetree.resolve_device(dev_spec)
+        device = resolve_device(self.storage, dev_spec)
 
         if not device:
             return ""
--- a/pyanaconda/modules/storage/partitioning/manual/manual_module.py
+++ b/pyanaconda/modules/storage/partitioning/manual/manual_module.py
@@ -134,7 +134,7 @@ class ManualPartitioningModule(PartitioningModule):
 
         for device in self.storage.devicetree.devices:
 
-            if not device.isleaf and not device.raw_device.type == "btrfs subvolume" and not device.raw_device.type == "btrfs volume":
+            if not device.isleaf and not device.raw_device.type == "btrfs subvolume":
                 continue
 
             # Is the device usable?
--- a/pyanaconda/modules/storage/partitioning/manual/manual_module.py
+++ b/pyanaconda/modules/storage/partitioning/manual/manual_module.py
@@ -170,11 +170,23 @@ class ManualPartitioningModule(PartitioningModule):
         request.device_spec = self._btrfs_device_spec(device) or device.path
         request.format_type = device.format.type or ""
         request.reformat = False
+        request.mount_options += self._btrfs_mount_options(device)
 
         if device.format.mountable and device.format.mountpoint:
             request.mount_point = device.format.mountpoint
 
         return request
+
+    def _btrfs_mount_options(self, device):
+        """Get btrfs subvolume mount options.
+
+        :param device: a Blivet's device object
+        :return: a btrfs subvolume mount options
+        """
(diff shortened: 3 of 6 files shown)


rvykydal:
@vojtechtrefny could you take a look please?
What to you think about 1. and 2. from the first comment?

pyanaconda/modules/storage/devicetree/utils.py
```diff
@@ -290,3 +290,24 @@ def find_unconfigured_luks(storage):
...
+    'UUID=2252ec30-1fce-4f8e-bdef-c50c3a44ede4@root'
+
+    :param storage: an instance of Blivet
+    :param dev_spec: a string describing a block device
+    :returns: the device
+    :rtype: :class:`~.devices.StorageDevice` or None
+    """
+
+    options = None
+    if dev_spec.startswith("UUID") and "@" in dev_spec:
```
poncovka, line 309: @vojtechtrefny Couldn't this be part of Blivet or is it too Anaconda specific?

pyanaconda/modules/storage/devicetree/viewer.py
```diff
@@ -347,13 +347,16 @@ def resolve_device(self, dev_spec):
...
+        the volume and the name of the subvolume, as:
+        'UUID=2252ec30-1fce-4f8e-bdef-c50c3a44ede4@root'
 
         If no device is found, return an empty string.
 
         :param dev_spec: a string describing a block device
         :return: a device name or an empty string
         """
-        device = self.storage.devicetree.resolve_device(dev_spec)
+        device = resolve_device(self.storage, dev_spec)
```
poncovka, line 359: It kind of scares me to have two different versions of `resolve_device`. I would be fine with overriding the `DeviceTree` class with our own and using the new version everywhere (if it is safe to do so).

pyanaconda/modules/storage/partitioning/manual/manual_module.py
```diff
@@ -162,15 +167,65 @@ def _create_request_for_device(self, device):
...
+        if device.raw_device.type in ("btrfs volume", "btrfs subvolume"):
+
+            uuid = self._get_btrfs_volume_uuid(device)
+            if uuid:
+                device_spec = "UUID={}".format(uuid)
+            else:
+                log.error("volume UUID of a btrfs device not found")
+
+            if device.raw_device.type == "btrfs subvolume":
+                device_spec = device_spec + "@{}".format(device.name)
```
poncovka, line 211: @vojtechtrefny And this?

poncovka (commented):
@rvykydal The Storage module has a lot of unit tests that use pre-created and partially mocked device tree. Could you look at them and try to cover the new code, if possible? Also I should have read your list of possible issues first. I see that you have already mentioned a lot of stuff there.

M4rtinK (approved):
Looks good to me, thanks! :)

jkonecny12:
Yes, thanks for writing this here.

So my points, there are tools like [dnf snapper plugin](https://dnf-plugins-extras.readthedocs.io/en/latest/snapper.html) which creates snapshot on each DNF transaction. Thanks to these tools you can easily get on standard workstation into a hundreds of snapshots. This has a few issues we need to take into account:

- How to show so many subvolumes in the mount point selection?
  - Maybe just to ignore snapshots?
- If we re-create the device (proposed reformat solution) it will destroy all the dependent devices too AFAIK.
  - What if some of these devices are also usable in the mount point selection (thus visible to user) and user want to use that. I guess in this situation we should to strikeout the not usable? In that case the logic has to be there.

I wonder is there any reason why would someone have dependent subvolumes which are not snapshots in BTRFS?

pyanaconda/modules/storage/devicetree/utils.py
vojtechtrefny: I think we could add an extra argument for `resolve_device` in blivet, something like `subvolspec` just for btrfs.

pyanaconda/modules/storage/partitioning/manual/manual_module.py
```diff
@@ -162,15 +167,65 @@ def _create_request_for_device(self, device):
         :return: an instance of MountPointRequest
         """
         request = MountPointRequest()
-        request.device_spec = device.path
+        request.device_spec = self._btrfs_device_spec(device) or device.path
```
vojtechtrefny, line 170: Would it be possible to change `MountPointRequest` here to have more information (like adding a `request.subvolspec` here)? I don't like the idea of taking Blivet device, reducing it to one string (path or uuid + subvolume name in case of btrfs) just to use `devicetree.resolve_devspec` later to get the Blivet device back. I understand that a Blivet device cannot be send over DBus, but inventing a special "device specification string" just for btrfs subvolumes seems like a very complicated solution.

pyanaconda/modules/storage/partitioning/manual/manual_module.py
```diff
@@ -162,15 +167,65 @@ def _create_request_for_device(self, device):
...
+            return "subvol={}".format(device.name)
+        else:
+            return ""
+
+    def _btrfs_device_spec(self, device):
+        """Get btrfs device specification of the device.
+
+        Examples:
+        'UUID=2252ec30-1fce-4f8e-bdef-c50c3a44ede4'
+        'UUID=2252ec30-1fce-4f8e-bdef-c50c3a44ede4@root'
```
vojtechtrefny, line 196: Using `@` as delimiter here might be problematic -- naming subvolumes like `@home` is a common practice.

pyanaconda/modules/storage/partitioning/manual/manual_module.py
rvykydal: I preferred it in the beginning as well, but then I realized that if I don't want to change API of https://github.com/rvykydal/anaconda/blob/31db7b8e5f6b423ea8a3ec2663b42c0720a771e9/pyanaconda/modules/storage/devicetree/viewer_interface.py#L104 by adding kind of special argument, I need to work only with device_spec and then having another request attribute would be redundant. Moreover, we specify all the devices just by the spec so far in the request (ie reduce the device to the device specification) so it would be only consistent to use just the spec also for btrfs files.
Constructing the device_spec as and arbitrary string and single argument for anaconda's ResolveDevice (and blivet's resolve_device?) seemed more future proof than adding arguments to me.

pyanaconda/modules/storage/devicetree/utils.py
rvykydal: I am not sure we want to add it also to https://github.com/rvykydal/anaconda/blob/31db7b8e5f6b423ea8a3ec2663b42c0720a771e9/pyanaconda/modules/storage/devicetree/viewer_interface.py#LL104C7-L104C7. I am not sure if we want to diverge here though (device specificaiton for Anaconda's ResolveDevice and blivet's resolve_device). See also my other comment about adding a special device specification attribute to MountPointRequest.
@poncovka what do you think?

pyanaconda/modules/storage/partitioning/manual/manual_module.py
rvykydal: Sure, that was a silly first stab, let's discuss it after we decide if we want it at all (device specification string for a btrfs subvolume).

rvykydal:
Replaced by https://github.com/rhinstaller/anaconda/pull/4913

PR #4836, opened 2023-06-14
webui: increasing timeout for download cockpit rpms

acruzgon:
This PR is to increase the timeout for download cockpit rpms to avoid an error like the following example:
```
File "/home/acruzgon/anaconda/anaconda/ui/webui/bots/machine/machine_core/timeout.py", line 41, in handle_timeout
    raise RuntimeError(self.error_message)
RuntimeError: Timed out on 'dnf download --downloaddir /var/tmp/build/ cockpit-ws cockpit-bridge firefox dbus-glib'
make[1]: *** [Makefile:743: ../../updates.img] Error 1
make[1]: Leaving directory '/home/acruzgon/anaconda/anaconda/ui/webui'
make: *** [Makefile:747: create-updates.img] Error 2
```

Also to avoid any inconvenient with contributors with slower network connections.

### DIFF
--- a/ui/webui/test/build-rpms
+++ b/ui/webui/test/build-rpms
@@ -35,7 +35,7 @@ def build_rpms(srpm, image, verbose, quick):
         # FIXME boot.iso on rawhide does not currently contain the new cockpit dependencies
         # This will change once we include this changes upstream and start building boot.iso with the new dependencies
         # Then we can safely remove this workaround
-        machine.execute(f"dnf download --downloaddir /var/tmp/build/ {missing_packages}", stdout=sys.stdout)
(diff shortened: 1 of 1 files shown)


KKoukiou (changes requested):
As mentioned on slack this needs commit message imporovements:
something like:

webui: increase timeout for downloading anaconda-webui dependencies

The current timeout is very tight for contributors with slow connections.

M4rtinK:
BTW, we actually have some guidelines for commit messages: 
https://anaconda-installer.readthedocs.io/en/latest/contributing.html#guidelines-for-commits

KKoukiou (commented):
You need to squash the two commits. In anaconda we don't squash on merge, so commits should be already squashed when sending.

KKoukiou (approved):

PR #4837, opened 2023-06-14
webui: show partitions of local standard disks

KKoukiou:
The more information we show for the available devices the easier will be for the users to remove ambiguity for which disk(s) is the desired installation target.


M4rtinK (approved):
Looks good to me, thanks! :)

I think this can really help users choose the right disk & being in an expander, it should not be too overwhelming.

M4rtinK:
BTW, more advanced storage technologies still technically exist on partitions - eq. LVM will have its physical volumes placed on partitions, the same with BTRFS (not sure if it has a technical term for its constituent partitions ?) and software RAID.

So maybe it could be enough to just show "their" partitions with something like "PV of LVM VG abc" or "part of BTRFS layout  123".

BTW, in some rare cases, filesystems, software RAID or LVM PVs (not sure about BTRFS now) can exist also directly on the disk, without a partition table. Its not seen that often, especially outside of enterprise/servers, but we will have to handle it for the disk selection info eventually.

rvykydal:
Yes, I think having an option for partitions to be shown is definitely an improvement. And especially having UI bits there as a base for future improvements I see very valuable.

I can imagine something like mountable devices (https://github.com/rhinstaller/anaconda/blob/4c94371113afa5d3901dd73d18e0a8eb8373c0e6/pyanaconda/modules/storage/devicetree/handler_interface.py#L117) can be more descriptive in some cases (like btrfs layout maybe). Or even systems found on the disk (https://github.com/rhinstaller/anaconda/blob/4c94371113afa5d3901dd73d18e0a8eb8373c0e6/pyanaconda/modules/storage/devicetree/handler_interface.py#L124) - but it is a bigger task, that would perhaps require also more modifications of UI layout / design and also support detection of Windows systems.

rvykydal:
> Yes, I think having an option for partitions to be shown is definitely an improvement. And especially having UI bits there as a base for future improvements I see very valuable.
> 
> I can imagine something like mountable devices (
> 
> https://github.com/rhinstaller/anaconda/blob/4c94371113afa5d3901dd73d18e0a8eb8373c0e6/pyanaconda/modules/storage/devicetree/handler_interface.py#L117
> ) can be more descriptive in some cases (like btrfs layout maybe). 

Well, that would actually do similar thing it is mountable partitions (not devices). So in the end, it can be something close to what mount point assignment does:
https://github.com/rhinstaller/anaconda/blob/4c94371113afa5d3901dd73d18e0a8eb8373c0e6/pyanaconda/modules/storage/partitioning/manual/manual_module.py#L127.
But that would be for next steps.

rvykydal (approved):
Looks good to me, thank you.

PR #4840, opened 2023-06-17
Add build of anaconda-iso-creator autobuild (#infra)

jkonecny12:
Autobuild container image for building of the ISO.

This is just a small draft quickly created to find out if we want to have such a thing.

Question to answer **are we fine to have this without tests?** The issue with running the build test is that you need to have KVM access which means not a GH runners.

### DIFF
--- a/.github/workflows/container-rebuild-action.yml
+++ b/.github/workflows/container-rebuild-action.yml
@@ -53,7 +53,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        container-type: ['ci', 'rpm']
+        container-type: ['ci', 'rpm', 'iso-creator']
     env:
       CI_TAG: '${{ inputs.container-tag }}'
     timeout-minutes: 60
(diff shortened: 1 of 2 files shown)


VladimirSlavik:
There are no tests to run in the container, I think, so no problem not running them :)

However - why? The reason I ask is that the daily release rehearsals ("try release daily") build the container and actually use it, on all 3 main branches, so it's verified to be functional or raise an alarm.

If it's to maybe also push the container for master, that sounds like a good thing to try.

jkonecny12:
That is great idea!

To answer the why question. Having this will simplify building of the images and make those more accessible. If you look on the discussion happening yesterday, people are confused that it's starting the download of the image and fail.

M4rtinK:
> To answer the why question. Having this will simplify building of the images and make those more accessible. If you look on the discussion happening yesterday, people are confused that it's starting the download of the image and fail.

Yeah, that was super confusing - I was just following the docs & hit it right away. :P Nice to see this being addressed. :)

M4rtinK (approved):
LGTM. :)

VladimirSlavik:
Maybe consider pushing the container after the release test instead?

PR #4841, opened 2023-06-19
Revert "WebUI: explicitly specify GITHUB_BASE for pixel test commands"

KKoukiou:
This reverts commit dc9b3a5389f70bed4afc55548120741d0507a77a.

When pulling pixel test references with the full link GITHUB_BASE is not needed. This would be needed if we would use only the PR number.


jelly (approved):
Technically this is correct, but it also doesn't hurt very much.

PR #4842, opened 2023-06-19
webui: do not apply  log_step decorator for non permian tests

KKoukiou:


velezd:
I would suggest renaming the tests and class to EndToEnd tests or something like that. Because they can be executed without Permian and Permian has workflows for other kinds of tests not just for the WebUI.

ui/webui/test/helpers/step_logger.py
```diff
@@ -23,6 +24,8 @@ def log_step(snapshots=False, snapshot_before=False, snapshot_after=False, docst
     :type docstring: bool, optional
     """
     def decorator(function):
+        if not os.environ.get('PERMIAN', False):
```
velezd, line 27: The ENV variable name should be different, it has nothing to do with Permian. And I think it should only have an effect on the creation of the screenshots. Logging what steps are being performed is always useful and AFAIK has no negative effects.

velezd (changes requested):

KKoukiou:
So removing the step logger decorator changed the timing and revealed some flakiness, which makes 2 tests now consistently fail on this PR. Will debug this here.

ui/webui/test/helpers/step_logger.py
M4rtinK: Yes & more detailed logs are especially useful when we hit race conditions.

ui/webui/test/helpers/step_logger.py
```diff
@@ -34,13 +34,16 @@ def wrapper(*args, **kwargs):
             print(f'[TEST STEP] {function.__name__}{nice_args}')
             if docstring:
                 print(f'[DOC] {function.__doc__}')
-            if snapshots or snapshot_before:
-                BrowserSnapshot.new(args[0].browser)
+
+            if os.environ.get('END2END', False):
```
velezd, line 38: I suggest doing it like this:
```
end2end = os.environ.get('END2END', False)
if end2end and (snapshots or snapshot_before):
    ...
if end2end and (snapshots or snapshot_after):
    ...
```

velezd (changes requested):

ui/webui/test/helpers/step_logger.py
KKoukiou: Adjusted.

velezd (approved):

KKoukiou:
testing farm is broken because of dnf5 https://redhat-internal.slack.com/archives/C04MU29TES1/p1687419798774249

PR #4757, opened 2023-05-11
Custom mount point

jelly:
Draft PR of a first version of the mount point assignment feature https://issues.redhat.com/browse/INSTALLER-3508

Some things which still need work:

- [ ] Texts / labels
- [ ] Switching between manual / automated partitioning? This is probably broken
- [ ] Validation / Error handling
- [ ] Conditionals for storage configuration page, when do we disable the feature?
- [ ] Automated Tests
- [ ] Generalise applyDefaultStorage?

Note: this PR depends on https://github.com/rhinstaller/anaconda/pull/4749

PR #4843, opened 2023-06-19
webui: mount point assignment support

KKoukiou:
Enhanced version of https://github.com/rhinstaller/anaconda/pull/4757 

I made many changes that I felt maybe force pushing into the existing branch is bad idea. Let me know if you prefer.

@jelly kept your commit an d fixed the issue with the custom mountpoint option still showing plain disks. 

Wanna take a look?

### DIFF
--- a/ui/webui/src/components/storage/CustomMountPoint.jsx
+++ b/ui/webui/src/components/storage/CustomMountPoint.jsx
@@ -1,0 +1,254 @@
+/*
+ * Copyright (C) 2023 Red Hat, Inc.
+ *
+ * This program is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation; either version 2.1 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful, but
+ * WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with This program; If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import cockpit from "cockpit";
+import React, { useState, useEffect } from "react";
+
+import {
+    Alert,
+    Checkbox,
+    Flex,
+    HelperText,
+    HelperTextItem,
+    Popover,
+    Select,
+    SelectOption,
+    SelectVariant,
+    TextContent,
+} from "@patternfly/react-core";
+import { HelpIcon } from "@patternfly/react-icons";
+
+import { ListingTable } from "cockpit-components-table.jsx";
+import { EmptyStatePanel } from "cockpit-components-empty-state.jsx";
+import { AnacondaPage } from "../AnacondaPage.jsx";
+
+import {
+    createPartitioning,
+    setBootloaderDrive,
+    setManualPartitioningRequests,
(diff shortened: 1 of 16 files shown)


ui/webui/src/components/storage/CustomMountPoint.jsx
```diff
@@ -0,0 +1,264 @@
...
+                    if (selection === "/") {
+                        newRow.reformat = true;
+                    }
+
+                    newRow["mount-point"] = selection;
+                }
+                return newRow;
+            });
+        });
+        setManualPartitioningRequests({ partitioning, requests: requestsToDbus(requests) }).catch(onAddErrorNotification);
```
KKoukiou, line 190: This is using the old requests object and therefore is wrong.

ui/webui/src/components/storage/InstallationDestination.jsx
```diff
@@ -212,8 +213,9 @@ const LocalStandardDisks = ({ deviceData, idPrefix, setIsFormValid, onAddErrorNo
                         usableDisks,
                     });
                     setDisks(usableDisks.reduce((acc, cur) => ({ ...acc, [cur]: defaultDisks.includes(cur) }), {}));
-                }, console.error);
-    }, [refreshCnt]);
+                })
+                .then(refreshDeviceData, console.error);
+    }, [refreshCnt, refreshDeviceData]);
```
rvykydal, line 218: Student question: this dependency is the reason why the refreshDeviceData is using useCallback, right?

jelly:
Some UI things I did not address, which can be a follow up:

# Allows selection when you have no partitions

image

image

# Allows selection, when you have a partition but no Filesystem

image

image

Note that every time you change something with the partitions you need to go back and click `Detect disks`, adding a button everywhere to `detect disks` is a solution but not super nice if you ask me.

ui/webui/src/components/storage/InstallationDestination.jsx
KKoukiou: Right, if refreshDeviceData was not using useCallback this would be re-triggered unnecessarily.

KKoukiou:
> Would it perhaps make sense to split the PR into more commits ? Say one commit for the JS API code calling DBus, one for the review screen & similar ?

But we need the front end to make the JS APIs call for dbus testable.

ui/webui/src/components/storage/CustomMountPoint.jsx
```diff
@@ -0,0 +1,276 @@
...
+                "device-spec": cockpit.variant("s", row["device-spec"]),
+                "format-type": cockpit.variant("s", row["format-type"]),
+                "mount-point": cockpit.variant("s", row["mount-point"]),
+                reformat: cockpit.variant("b", row.reformat),
+            };
+        });
+    };
+
+    const handleOnSelect = (selection, device) => {
+        const newRequests = requests.map(row => {
```
KKoukiou, line 172: i would like to emit the requests object in the state and read it from the global state instead.

KKoukiou:
@jelly i need some support on this. Checked your review comments from above:

- [ ] no custom mountpoint section is disabled if no partitions on the device
- [ ] when no f ilesystem on the partition I still show it - and we get then the error when clicking next as you suggest. Mind if we fi this in followup? 
- [ ] errors don't get discared currently when switching between pages indeed. Since a change in this approach is affecting the whole app, not just this feature let's please check this as well seperately 
- [ ] Parsing the errors to make this inline can be considered - but let's keep it as well for enhancement.

M4rtinK (approved):
Looks good to me - quite the piece of work @KKoukiou & @jelly! :)

Just a couple minor remarks. :)

PR #4845, opened 2023-06-21
WebUI: Convert InstallationProgress to function

OndrejZobal:
Convert InstallationProgress from a class based react component to a function component.


M4rtinK (approved):
Looks good to me, thanks a lot! Looking much better now! :)

KKoukiou (commented):
If you repush please also include in your commit message that you dropped AddressContext because it was not used. 

We had it there for a reason, in case the dbus address changes we should get notifed but I guess it got broken over time and there is no test.

I will fix it in separate PR.

PR #4847, opened 2023-06-21
webui: start using useReducer for managing complicated state objects centrally

KKoukiou:


rvykydal (approved):
Thank you. I think it is definitely useful to have such managed central state. We may need to be careful about what should go to the state, to what level and scope it could reflect/cache the state of the backend and when/if it is better to communicate directly with the backend on lower level? (especially storage may be tricky if we have more features, screens, etc?). But these are rather vague concerns that may arise from me not grasping the concept of the reducer with thunk fully yet.

KKoukiou:
> Thank you. I think it is definitely useful to have such managed central state. We may need to be careful about what should go to the state, to what level and scope it could reflect/cache the state of the backend and when/if it is better to communicate directly with the backend on lower level? (especially storage may be tricky if we have more features, screens, etc?). But these are rather vague concerns that may arise from me not grasping the concept of the reducer with thunk fully yet.


So the idea if that you can trigger actions to update the store (global state) from anywhere in the project. And then every component has access to the same store. That solved the problem that different pages might use different version of the backend data.

rvykydal (approved):
Looks good to me.

KKoukiou:
Still dnf5 breakage on testing farm.

PR #4850, opened 2023-06-22
WebUI: Switch Quit/Reboot button in Progress spoke

OndrejZobal:
Dynamically switch between Quit and Reboot buttons in the InstallationProgress spoke after the installation is finished.

I am not sure if passing everything down as a prop is an elegant solution, so I will be happy if anybody has suggestions on how to do this differently.

### DIFF
--- a/ui/webui/src/components/AnacondaWizard.jsx
+++ b/ui/webui/src/components/AnacondaWizard.jsx
@@ -52,6 +52,9 @@ export const AnacondaWizard = ({ onAddErrorNotification, toggleContextHelp, hide
     const [showPassphraseScreen, setShowPassphraseScreen] = useState(false);
     const [storageScenarioId, setStorageScenarioId] = useState(window.sessionStorage.getItem("storage-scenario-id") || getDefaultScenario().id);
 
+    // On live media rebooting the system will actually shut it off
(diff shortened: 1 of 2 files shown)


ui/webui/src/components/AnacondaWizard.jsx
```diff
@@ -52,6 +52,9 @@ export const AnacondaWizard = ({ onAddErrorNotification, toggleContextHelp, hide
     const [showPassphraseScreen, setShowPassphraseScreen] = useState(false);
     const [storageScenarioId, setStorageScenarioId] = useState(window.sessionStorage.getItem("storage-scenario-id") || getDefaultScenario().id);
 
+    // True if using the quit button will reboot the system
+  const isQuitReboot = conf["Installation System"].type === "BOOT_ISO";
```
KKoukiou, line 56: Can you maybe rename the isQuiteReboot to something more self explanatory? isBootIso? or invert the logic and make it isLiveMedia? 

The you can also improve the comment so that people understand what's going on here./
/* On live media rebooting the system will actually shut it off */

Or whatever the case is. And if you can support it with some link so a documentation that would be idea.

KKoukiou (approved):

PR #4851, opened 2023-06-22
Revert "Add our repository with the new annobin package (#infra)"

KKoukiou:
This reverts commit bf5bccc6040835f7e3754f861cb31394e05240b9.

The bugzilla is long time fixed.

### DIFF
--- a/.packit.yml
+++ b/.packit.yml
@@ -49,10 +49,6 @@ jobs:
     trigger: pull_request
     targets:
       - fedora-rawhide
-    additional_repos:
-      # This repository contains fixup of Rawhide broken environment.
-      # Mainly useful when there is a package which is not yet in Rawhide but build is available.
-      - "https://fedorapeople.org/groups/anaconda/repos/anaconda_fixup_repo/"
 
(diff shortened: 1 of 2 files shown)


KKoukiou:
This is currently blocked on dnf5 breakage.

VladimirSlavik:
Sounds a bit contradictory. To fix the "bz no longer a problem", you want to remove the package from the repo, not remove the repo, I think. Maybe we don't need the repo as well, but then it might be better to write that instead?

VladimirSlavik (commented):
Thank you, lgtm. I'm not in the chat so I will leave acking to somebody more in the loop, but no objections from me.

VladimirSlavik:
@KKoukiou are you still interested in this? I just randomly stumbled on the same repo in packit config...

jkonecny12 (approved):
We did not used this option much so I guess we should remove it instead of keeping it around.

VladimirSlavik:
It might be useful to grep for the repo in docs, I think.

PR #4854, opened 2023-06-22
Fix Anaconda startup on Live media via custom webview launcher  script

M4rtinK:
This should fix the startup on Live images for now.

The first commit reverts the revert & switches Anaconda to again use the custom version of cockpit-desktop, called webui-desktop.

The second commit updates the script to the version found in the latest cockpit-ws package.

### DIFF
--- a/tests/shellcheck/ignores.sh
+++ b/tests/shellcheck/ignores.sh
@@ -5,4 +5,5 @@ ignore_patterns=(
 dockerfile/
 scripts/
 widgets/
+webui-desktop  # This is just a temporary file, needed to run the Web UI on the Live image, to be replaced with a long term solution eventually
 )
--- a/ui/webui/Makefile.am
+++ b/ui/webui/Makefile.am
@@ -33,6 +33,7 @@ UPDATES_IMG=../../updates.img
 export GITHUB_BASE
 export TEST_OS
 
+dist_libexec_SCRIPTS = webui-desktop
 # makes sure it gets built as part of `make` and `make dist`
 dist_noinst_DATA = \
 	$(WEBPACK_TEST) \
@@ -57,6 +58,7 @@ install-data-hook: $(WEBPACK_TEST)
 	cp -r dist/* $(DESTDIR)/usr/share/cockpit/$(PACKAGE_NAME)
 	mkdir -p $(DESTDIR)/usr/share/metainfo/
 	cp org.cockpit-project.$(PACKAGE_NAME).metainfo.xml $(DESTDIR)/usr/share/metainfo/
+	cp webui-desktop $(DESTDIR)/usr/libexec/
 
 EXTRA_DIST = dist src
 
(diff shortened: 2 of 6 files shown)


jkonecny12 (commented):
Looks good to me otherwise.

tests/shellcheck/ignores.sh
```diff
@@ -5,4 +5,5 @@ ignore_patterns=(
 dockerfile/
 scripts/
 widgets/
+webui-desktop  # This is just a temporary file, to be replaced with cockpit-desktop once 'unshare' is not filtered out by lorax
```
KKoukiou, line 8: Reasoning is wrong, unshare is now present
https://github.com/weldr/lorax/pull/1222

KKoukiou (commented):
Can you also pls write in the commit message what was the problem with the unshare command.

ui/webui/Makefile.am
```diff
@@ -57,6 +58,7 @@ install-data-hook: $(WEBPACK_TEST)
 	cp -r dist/* $(DESTDIR)/usr/share/cockpit/$(PACKAGE_NAME)
 	mkdir -p $(DESTDIR)/usr/share/metainfo/
 	cp org.cockpit-project.$(PACKAGE_NAME).metainfo.xml $(DESTDIR)/usr/share/metainfo/
+	cp webui-desktop $(DESTDIR)/usr/libexec/
```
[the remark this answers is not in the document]
M4rtinK: This is just a re-revert of working code - i suggest we handle non-critical improvements in separate PR.

M4rtinK:
> Can you also pls write in the commit message what was the problem with the unshare command.

So I have this in the first commit message:

`Unfortunately it turns out the unshare command prevents the Web UI from
being started on Live images, so we need to put this temporary hotfix
back in.`

As for why it does not work - we have not investigated further so far. It was actually the Image Builder team finding out it start to work if we drop the unshare command from the `cockpit-desktop` script.

tests/shellcheck/ignores.sh
M4rtinK: Updated the reasoning.

M4rtinK:
Updated the PR.

M4rtinK:
@martinpitt @allisonkarlitskaya @ondrejbudai @supakeen So basically during early Image Builder built Live image we found out the Anaconda Web UI won't start as long as `unshare` is used by the cockpit-desktop script. Once we manually removed it, Web UI started as expected. So for now we decided to go back to the forked `cockpit-desktop` script temporarily, but it would be certainly good to think about a more permanent solution eventually. Suggestions Welcome. :)

jkonecny12:
Feel free to ignore these issues:
- cppcheck is failing on code from gettext-devel package: reported [here](https://bugzilla.redhat.com/show_bug.cgi?id=2217542)
- ShellCheck is failing on `webui-desktop` which is Cockpit code and should be working
- Testing Farm is failing on DNF5 incompatibility issues

jkonecny12 (approved):
Looks good to me as hotfix.

KKoukiou (approved):

PR #4855, opened 2023-06-23
webui: Fix WebUI end2end tests after recent renaming and introducing of screenshot limit

velezd:
Looks like we missed some things in the recent changes of WebUI tests. I guess nobody tried to run them.

### DIFF
--- a/ui/webui/test/end2end/default.tc.yaml
+++ b/ui/webui/test/end2end/default.tc.yaml
@@ -8,7 +8,7 @@ priority: 9
 execution:
   type: anaconda-webui
   automation_data:
-    script_file: ./ui/webui/test/integration/default.py
+    script_file: ./ui/webui/test/end2end/default.py
     test_case: DefaultInstallation
 instructions:
   setup:
--- a/.github/workflows/webui-tests.yml
+++ b/.github/workflows/webui-tests.yml
@@ -5,7 +5,7 @@
 # ======================================
 # The template is located in: webui-tests.yml.j2
 
-# Run webui integration tests in a PR
+# Run webui end2end tests in a PR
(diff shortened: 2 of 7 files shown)


velezd:
There seem to be more issues with the end2end tests, I will look into it next week.

velezd:
The last one was issue with my setup.

ui/webui/test/end2end/default.tc.yaml
```diff
@@ -8,7 +8,7 @@ priority: 9
 execution:
   type: anaconda-webui
   automation_data:
-    script_file: ./ui/webui/test/integration/default.py
+    script_file: ./ui/webui/test/end2end//default.py
```
KKoukiou, line 11: Double //

KKoukiou (approved):

KKoukiou (commented):
I think in anaconda they have this convention that commit changing infra files like .github directory, need the infra prefix in the commit message. that's why the first test is failing for this PR. can you please adjust for the test to pass?

velezd:
> I think in anaconda they have this convention that commit changing infra files like .github directory, need the infra prefix in the commit message. that's why the first test is failing for this PR. can you please adjust for the test to pass?

I've changed it but the check doesn't seem to be working correctly, or I'm missing something.
`infra: webui: Fix end2end tests after renaming --> Is missing 'infra:' prefix or '(#infra)' postfix`

KKoukiou:
@velezd it's because the check is very naive, and wants pRs with 'infrastructure label' to have ALL commits starting with infra which is wrong. I suggest split the PR into 2 PRs, one with the infra commit and one without.

jkonecny12 (approved):
Looks good to me. I think it's fine to ignore the infra check here.

PR #4856, opened 2023-06-23
webui: store system language information in the global store

KKoukiou:
Before having fetched the system language just show a spinner, to avoid having glitch in the UI when default system language is not english.


KKoukiou:
testing farm failing because of new dnf5: fix at https://gitlab.com/testing-farm/tests/-/merge_requests/24

jelly (approved):

PR #4858, opened 2023-06-26
Port GUI kernel switcher for ARM 64k

OndrejZobal:
Port https://github.com/rhinstaller/anaconda/pull/4689 to upstream.

Adds a combo box allowing user to select kernel-64k in the software selection spoke

### DIFF
--- a/data/anaconda.conf
+++ b/data/anaconda.conf
@@ -310,6 +310,9 @@ password_policies =
     user (quality 1, length 6, empty)
     luks (quality 1, length 6)
 
+# Should kernel options be shown in the software selection spoke?
+show_kernel_options = True
+
 [License]
 # A path to EULA (if any)
 #
--- a/data/profile.d/rhel.conf
+++ b/data/profile.d/rhel.conf
@@ -39,6 +39,7 @@ swap_is_recommended = True
 [User Interface]
 help_directory = /usr/share/anaconda/help/rhel
 custom_stylesheet = /usr/share/anaconda/pixmaps/redhat.css
+show_kernel_options = True
 
 [License]
 eula = /usr/share/redhat-release/EULA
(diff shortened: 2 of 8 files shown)


M4rtinK (approved):
Looks good to me. :)

jkonecny12 (approved):
Looks good to me. Thanks!

OndrejZobal:
I am sorry there were some issues with the TUI port and the test were missing. In my latest force push I have replaced is_arm with is_aarch64. It should all be good now.

jkonecny12 (approved):
Thanks for spotting that and fixing.

OndrejZobal:
Oh, what made those kickstart tests sad?

jkonecny12:
I'm not able to test this reasonably on Fedora because the required kernel is not available there. However, kickstart tests are successful and we have a unit tests.

AdamWill:
I think this broke composes. `match_available_packages` is a method of the anaconda `DNFManager` class, but this seems to try to use it on an instance of the `DNFPayload` class, causing this crash:

aarch64crash

this caused https://pagure.io/releng/failed-composes/issue/5212 .

AdamWill:
Reverted downstream, we have time to consider a proper fix upstream. Please don't cut a new release and build it in Fedora without fixing this, though.

jkonecny12:
@VladimirSlavik please don't do a new release before we have revert or fix.

jkonecny12:
The issue is that the original RHEL-9 code don't have payload as Anaconda module but master have. So we need to use module for this.

jkonecny12:
@AdamWill I have a revert for now: https://github.com/rhinstaller/anaconda/pull/4958
We don't have time now for a proper fix.

I wonder why this wasn't caught by our kickstart tests...

VladimirSlavik:
+1 for dismay at the kickstart test success.

AdamWill:
well, for a start, this will only happen on aarch64. also, the affected code is in the UI stuff, so if you're doing unattended kickstart installs without a display attached, maybe the UI code just doesn't trigger?

AdamWill:
btw, if you're wondering why openQA update tests didn't catch it - we don't run the update tests on aarch64 because I don't have enough aarch64 worker capacity :(

VladimirSlavik:
> maybe the UI code just doesn't trigger

Sounds like the root cause.

PR #4860, opened 2023-06-27
Fix cppcheck failure on gettext.h

jkonecny12:
The `gettext.h` is not our code. It's comming from gettext-devel package and we copy that into our code during autogen execution. However, new cppcheck version 2.11 started to gave errors from this header file. It's filed here

https://bugzilla.redhat.com/show_bug.cgi?id=2217542

To not block ourselves on an error from another library let's suppress that error for now.


jkonecny12:
**REBASED** to fix tests

PR #4863, opened 2023-06-27
Fix DNF5 argument change in build-rpms script

jkonecny12:
We are downloading packages from Rawhide and Rawhide switched to DNF5 so which is not backward compatible. Let's fix issue raised from this change.


KKoukiou:
Closing in favor of https://github.com/rhinstaller/anaconda/pull/4862

PR #4864, opened 2023-06-27
webui: state update logic fixes in the reducer 

KKoukiou:
Split out of the custom mountpoint PR.


M4rtinK (approved):

PR #4865, opened 2023-06-27
storage: no need in HFS+ on Apple Macs

mikhailnov:
Tests have shown that:
* HFS-specific code in grub-install utility contains silly mistakes (https://abf.io/import/grub2/blob/1145a32d77/0001-grub-install-fix-creating-mach_kernel-on-x86-mac.patch)
* even whith that patch, that code is broken and registers an incorrect bootloader entry via efibootmgr in UEFI (we in ROSA patch Anaconda to call grub2-install directly instead of calling efibootmgr inside Anaconda for unification)
* macOS installer does not allow to create an HFS+ partition for EFI and creates a FAT partition by default

It is not clear what for the removed code existed. This patch fixes installation on x86 macs.
Did not test on ARM-based macbooks (I think they should also simply follow the UEFI specification and do not require removed hacks).

Co-authored with @temaps


jkonecny12:
We don't have Mac for testing so I would go with just smoke testing in this case.

jkonecny12:
So a few notes to this:

it seems that this code is here for a long time. It was originally named hfs+ and introduced here: https://github.com/rhinstaller/anaconda/commit/383a190ddb0648b78a0d80a9903348a64987438a in RHEL-5. So there is low chance we will find more about it :D

About the PR, please remove all the occurrences of the class - see the tests.

VladimirSlavik:
Doesn't this effectively mean we stop supporting the intel macs as an installation target?

mikhailnov:
> Doesn't this effectively mean we stop supporting the intel macs as an installation target?

I does not because Intel Macs understand FAT16 and FAT32, and macOS there itself makes a FAT UEFI partition, but cannot create an HFS+ UEFI partition. This will allow to reuse the EFI with macos.

mikhailnov:
Both variants work:

* reuse the FAT UEFI partition from macOS
* create another FAT partition

mikhailnov:
We tested with macbooks from 2009, 2010 and probably from 2008 and a bit later ones. Is is possible to create a Fedora LiveCD with Anaconda with this patch so that I would test Fedora on them with this change? We tested ROSA.

mikhailnov:
Another user tested it (https://abf.io/platforms/rosa2021.1/products/316/product_build_lists/47297) on Macbook Air 13 from 2016 and reported that Anaconda works correctly there.

jkonecny12 (approved):
Looks good to me.

mikhailnov:
Great, thanks! 😊

PR #4866, opened 2023-06-28
WebUI: Change help drawer width

OndrejZobal:
- Set max width to half of viewport,
- increase minimal width to prevent the
  "x" button being pushed by the help title,
- convert default size to rem.


KKoukiou (approved):
This will go away with garrets new designs i think, but the change looks fine.

We also need to unify this help drawer with the ones from storage configuration page

PR #4868, opened 2023-06-28
Fix Web UI iso missing grub boot options (#infra)

jkonecny12:
Our custom builds of Web UI boot.iso add correct boot options on the command line for EFI but we forgot to check grub. Let's fix this now.


M4rtinK (approved):

PR #4870, opened 2023-06-29
Split write network and hostname configuration

temaps:
When the parameter can_configure_network is disabled in the configuration, the file /etc/hostname is not written to the installed system.

### DIFF
--- a/pyanaconda/installation.py
+++ b/pyanaconda/installation.py
@@ -153,9 +153,9 @@ class RunInstallationTask(InstallationTask):
 
         configuration_queue.append(os_config)
 
+        overwrite = payload.type in PAYLOAD_LIVE_TYPES
         # schedule network configuration (if required)
         if conf.target.can_configure_network and conf.system.provides_network_config:
-            overwrite = payload.type in PAYLOAD_LIVE_TYPES
             network_config = TaskQueue(
                 "Network configuration",
                 _("Writing network configuration")
@@ -166,6 +166,18 @@ class RunInstallationTask(InstallationTask):
                 (overwrite, )
             ))
             configuration_queue.append(network_config)
+        # schedule hostname configuration (if required)
+        if conf.system.can_change_hostname:
+            hostname_config = TaskQueue(
+                "Hostname configuration",
+                _("Writing hostname configuration")
+            )
+            hostname_config.append(Task(
+                "Hostname configuration",
+                network.write_configuration_hostname,
+                (overwrite, )
+            ))
+            configuration_queue.append(hostname_config)
 
         # add installation tasks for the Users DBus module
         if is_module_available(USERS):
--- a/pyanaconda/network.py
+++ b/pyanaconda/network.py
@@ -54,7 +54,7 @@ _nm_client = None
 __all__ = ["get_supported_devices", "status_message", "wait_for_connectivity",
            "wait_for_connecting_NM_thread", "wait_for_network_devices", "wait_for_connected_NM",
            "initialize_network", "prefix_to_netmask", "netmask_to_prefix", "get_first_ip_address",
-           "is_valid_hostname", "check_ip_address", "get_nm_client", "write_configuration"]
+           "is_valid_hostname", "check_ip_address", "get_nm_client", "write_configuration", "write_configuration_hostname"]
 
 
 def get_nm_client():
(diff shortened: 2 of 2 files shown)


mikhailnov:
A bit of context why this change was made. A derivative of ROSA - MOS (ROSA-based distro for schools) - sets can_configure_network=False by default, because every teacher/student in Moscow schools has a personal password for the local WiFi network. Someone's password was made visible to every user by can_configure_network=True.

But we in ROSA generate random hostnames on start of the LiveCD (https://abf.io/import/anaconda/blob/rosa2023.1/anaconda-livecd-init.sh) and then inherit them into the installed system (see also: https://abf.io/import/anaconda/blob/6b6591efa3/0046-Copy-current-hostname-into-installed-system.patch , I think this could also be upstreamized, what do you think?). But without the changes in the PR both a randomly and manually (in the network spoke) set hostname was not copied into the installed system.

Setting random hostnames helps to identify machines inside network.

VladimirSlavik:
@rvykydal Radku, can you please take a look at this?

pyanaconda/installation.py
```diff
@@ -166,6 +166,18 @@ def _prepare_configuration(self, payload, ksdata):
...
             configuration_queue.append(network_config)
+        # schedule hostname configuration (if required)
+        elif conf.system.can_change_hostname:
+            hostname_config = TaskQueue(
+                "Hostname configuration",
+                _("Writing hostname configuration")
+            )
+            hostname_config.append(Task(
+                "Hostname configuration",
+                network.network.write_configuration_hostname,
```
rvykydal, line 177: Anaconda crashes here, there should be `network.write_configuration_hostname`

rvykydal (changes requested):
I think the split of hostname configuration makes sense.

In the PR can_configure_network still implies can_configure_hostname which seems reasonable and should prevent us from possible breaking of some current use cases. 
But I'd prefer to split the hostname configuration from write_configuration completely and call it separately after "Network confgiration" task (if can_configure_network or can_configure_hostname) so we are more clear and explicit in the tasks and logs in what is happening.

mikhailnov:
> But I'd prefer to split the hostname configuration from write_configuration completely and call it separately after "Network confgiration" task (if can_configure_network or can_configure_hostname) so we are more clear and explicit in the tasks and logs in what is happening

+1 , code flow is not obvious otherwise

rvykydal:
> But we in ROSA generate random hostnames on start of the LiveCD (https://abf.io/import/anaconda/blob/rosa2023.1/anaconda-livecd-init.sh) and then inherit them into the installed system (see also: https://abf.io/import/anaconda/blob/6b6591efa3/0046-Copy-current-hostname-into-installed-system.patch , I think this could also be upstreamized, what do you think?). 

I am afraid making this default behavior would be in conflict with https://github.com/rhinstaller/anaconda/pull/643 where we stopped using the hostname from installer environment for target system.

mikhailnov:
> > But we in ROSA generate random hostnames on start of the LiveCD (https://abf.io/import/anaconda/blob/rosa2023.1/anaconda-livecd-init.sh) and then inherit them into the installed system (see also: https://abf.io/import/anaconda/blob/6b6591efa3/0046-Copy-current-hostname-into-installed-system.patch , I think this could also be upstreamized, what do you think?).
> 
> I am afraid making this default behavior would be in conflict with #643 where we stopped using the hostname from installer environment for target system.

Hm... Commit message there says: "Doing so, the installed system would not use new transient hostname if it boots in different environment". But the current behaviour is to set the same static hostname, but not skip setting it, isn't it?

mikhailnov:
The script generates a random hostname only if another hostname, including a transient one via DHCP, was not set:
https://abf.io/import/anaconda/blob/d94b36a3eb/anaconda-livecd-init.sh#lc-21

pyanaconda/installation.py
temaps: https://github.com/rhinstaller/anaconda/pull/4870/files#diff-335d1ad6ab4d87d2febae93c61ef93bb406a9913b4fef5503e337481c28ee697R57
https://github.com/rhinstaller/anaconda/pull/4870/files#diff-335d1ad6ab4d87d2febae93c61ef93bb406a9913b4fef5503e337481c28ee697R279

temaps: That's exactly what was done. First, the operation is divided into two functions, and then they are called in turn depending on can_change. Moreover, the formulation of the name configuration task comes just after the network configuration task.

temaps: > Anaconda crashes here, there should be `network.write_configuration_hostname`

fixed typo :-)

temaps:
The tests are confused by the name "hostname" (write-hostname-configuration), there don't seem to be any more problems. Is this an obstacle?

PR #4871, opened 2023-06-29
webui: Start Web UI when the anaconda-webui package is installed

M4rtinK:
We do this by checking the /usr/share/cockpit/anaconda-webui path exists. If it exists, Web UI will be launched, regardless of the inst.webui option being present.

### DIFF
--- a/pyanaconda/anaconda.py
+++ b/pyanaconda/anaconda.py
@@ -202,7 +202,7 @@ class Anaconda(object):
 
         # this boot option is meant to be temporary, so just check it directly
         # from kernel boot command line like this
-        if "webui" in kernel_arguments:
+        if "webui" in kernel_arguments or os.path.exists("/usr/share/cockpit/anaconda-webui"):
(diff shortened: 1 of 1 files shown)


M4rtinK:
So this unconditionally launches the Web UI if the anaconda-webui package is installed on the media - this should work on Live media (just tested that) and will likely also work on the boot.iso.

Right now there is no override, but we should likely come up with one for the Live image fallback UI use case - once it has been agreed which form it will take.

With this change in place, there will be a couple benefits:
- you can install anaconda-webui on a regular Live image & the Web UI will start instead of the GTK GUI, without the need for further interaction
- the Image Build created live images can drop the inst.webui option from their bootloader setup, if they want (boot options were never really heavily used on live images & it is kinda weird to use one on here)
- it would be very easy to create Live images using Lorax - just add `anaconda-webui` to the packages section of your input kickstart

KKoukiou (approved):

jkonecny12 (approved):
LGTM!

PR #4878, opened 2023-07-04
WIP: Check for webui with property

VladimirSlavik:
This is the same as other UIs.

Refactor of #4871.

WIP since there is potentially far more to do about this - see usages of `DisplayMode`.


M4rtinK (approved):
Looks good to me. :)

VladimirSlavik:
After some research and talking with Martin, let's keep `DisplayModes` aside for now. It's intertwined with pykickstart, so that's a whole new level of pain and refactoring to tackle. That means it's not feasible as a side project.

PR #4879, opened 2023-07-05
Handle subprocess disallowing preexec during shutdown

AdamWill:
In Python 3.12, you cannot pass a preexec_fn to subprocess.Popen during interpreter shutdown:

https://github.com/python/cpython/pull/104826

This avoids the problem by giving startProgram an arg to not use a preexec_fn, and passing that all the way from anaconda's exitHandler through execWithRedirect and _run_program.

### DIFF
--- a/anaconda.py
+++ b/anaconda.py
@@ -89,15 +89,15 @@ def exitHandler(rebootData):
                     util.dracut_eject(device_path)
 
         if flags.kexec:
-            util.execWithRedirect("systemctl", ["--no-wall", "kexec"])
+            util.execWithRedirect("systemctl", ["--no-wall", "kexec"], do_preexec=False)
             while True:
                 time.sleep(10000)
         elif rebootData.action == KS_SHUTDOWN:
-            util.execWithRedirect("systemctl", ["--no-wall", "poweroff"])
+            util.execWithRedirect("systemctl", ["--no-wall", "poweroff"], do_preexec=False)
         elif rebootData.action == KS_WAIT:
-            util.execWithRedirect("systemctl", ["--no-wall", "halt"])
+            util.execWithRedirect("systemctl", ["--no-wall", "halt"], do_preexec=False)
         else:  # reboot action is KS_REBOOT or None
-            util.execWithRedirect("systemctl", ["--no-wall", "reboot"])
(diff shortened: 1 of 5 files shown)


AdamWill:
This works OK in openQA testing.

A possible alternative is just to have the exit handler call subprocess directly; I don't think all these wrappers around it are *doing* much of any use on this path. But that seemed a slightly bigger change than this, so I went with this to start with.

AdamWill:
This (or a different fix) must be merged before the next release, because Python 3.12 has been merged into Rawhide. If a new release is made without this bug fixed, it'll be built in Rawhide by packit, and that build will be broken. (Fortunately, now gating is enabled, it should be blocked by gating rather than getting in and breaking the compose).

Also, today's compose testing indicates this needs extending to the VNC shutdown code:

https://openqa.fedoraproject.org/tests/2005161#step/_boot_to_anaconda/5

I'm on vacation without a laptop, so if someone else can do that, it'd be great.

VladimirSlavik:
Just a note, I thought of the following alternatives constrained to changes only in `startProgram`:
- Check dynamically if preexec should happen based on `sys.is_finalizing()` . This appears as a silent side effect, so not really desirable.
- Check if preexec is really needed - the function body is all conditionals - by testing all the clauses. Unfortunately this does not solve the problem, because `reset_handlers` is always `True` in the call path from `execWithRedirect`.
- Propagate the `reset_handlers` flag instead of `do_preexec` - mostly the same problem as above, does not really explain that it's preexec that should not happen.

So I think this is the only way.

VladimirSlavik (approved):

VladimirSlavik:
Needs:
- fix the VNC code path above too
- tests

Working on that.

AdamWill:
Thanks! What did you think of the 'just have the exit handlers call subprocess directly' idea?

VladimirSlavik:
Ah, I missed that one. I think this way is better - we have all the logging and such in place for these calls, and will likely expect the logs to look the same for this.

I also feel that worrying too much about this might be premature optimization. The exit-ing code paths need modularization, soonish, and the exec* family of functions is also one of the areas that are known to need rewriting for contemporary Python. So this might not live too long anyway...

PR #4881, opened 2023-07-05
webui: if device selection changed since last partitioning request redo the partitioning

KKoukiou:


jelly (approved):
Nice catch / solution.

PR #4882, opened 2023-07-06
webui: correct the last partitioning getter

KKoukiou:
Previously we would read just the last character of the path, which would give incorrect data if more than 9 partitionings would exist.

Now we parse the digits in the end of the path with a regexp.

### DIFF
--- a/ui/webui/src/components/AnacondaWizard.jsx
+++ b/ui/webui/src/components/AnacondaWizard.jsx
@@ -56,7 +56,7 @@ export const AnacondaWizard = ({ dispatch, storageData, localizationData, onAddE
     const [showPassphraseScreen, setShowPassphraseScreen] = useState(false);
     const [storageScenarioId, setStorageScenarioId] = useState(window.sessionStorage.getItem("storage-scenario-id") || getDefaultScenario().id);
     const lastPartitioning = useMemo(() => {
-        const lastPartitioningKey = Object.keys(storageData.partitioning || {}).find(path => parseInt(path[path.length - 1]) === Object.keys(storageData.partitioning).length);
+        const lastPartitioningKey = Object.keys(storageData.partitioning || {}).find(path => parseInt(path.match(/\d+$/)[0]) === Object.keys(storageData.partitioning).length);
 
         return storageData.partitioning?.[lastPartitioningKey];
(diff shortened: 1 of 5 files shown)


ui/webui/src/components/AnacondaWizard.jsx
```diff
@@ -56,7 +56,7 @@ export const AnacondaWizard = ({ dispatch, storageData, localizationData, onAddE
     const [showPassphraseScreen, setShowPassphraseScreen] = useState(false);
     const [storageScenarioId, setStorageScenarioId] = useState(window.sessionStorage.getItem("storage-scenario-id") || getDefaultScenario().id);
     const lastPartitioning = useMemo(() => {
-        const lastPartitioningKey = Object.keys(storageData.partitioning || {}).find(path => parseInt(path[path.length - 1]) === Object.keys(storageData.partitioning).length);
+        const lastPartitioningKey = Object.keys(storageData.partitioning || {}).find(path => parseInt(path.match(/\d+$/)[0]) === Object.keys(storageData.partitioning).length);
```
jelly, line 59: Would be nice to have a test for this, it still feels a bit fragile.

jelly (approved):

ui/webui/src/components/AnacondaWizard.jsx
KKoukiou: Good point - will add one.

VladimirSlavik (approved):
Looks good to me. Thank you!

vojtechtrefny (approved):

ui/webui/test/check-storage
```diff
@@ -234,6 +234,53 @@ class TestStorage(anacondalib.VirtInstallMachineCase):
...
+
+        # Read partitioning data before we do anything
+        created_partitioning = s.dbus_get_created_partitioning()
+
+        # Go to Review step
+        i.open()
+        i.next()
+        i.next()
+        i.next()
+        i.next()
```
M4rtinK, line 255: It seems to me that we might need to have a library function for reaching a given screen in the future, as the screen order ḿight be more dynamic in the future.

M4rtinK (approved):
Looks good to me as well. :)

VladimirSlavik (approved):

PR #4884, opened 2023-07-06
webui: tests: switch some storage tests to nondestructive

KKoukiou:
For the two tests affected in this commit it was enough to use add_ram_disk test storage helper, which has includes a teardown cleanup.

### DIFF
--- a/ui/webui/test/anacondalib.py
+++ b/ui/webui/test/anacondalib.py
@@ -27,12 +27,26 @@ sys.path.append(os.path.join(os.path.dirname(TEST_DIR), "bots/machine"))
 from machine_install import VirtInstallMachine
 from testlib import MachineCase  # pylint: disable=import-error
 
+from storage import Storage
+
 
 class VirtInstallMachineCase(MachineCase):
     MachineCase.machine_class = VirtInstallMachine
 
     def setUp(self):
+        # FIXME: running this in destructive tests fails because the SSH session closes before this is run
+        if self.is_nondestructive():
+            self.addCleanup(self.rescanStorage)
+
         super().setUp()
 
         self.machine.execute("systemctl restart cockpit")
         self.allow_journal_messages('.*cockpit.bridge-WARNING: Could not start ssh-agent.*')
+
+    def rescanStorage(self):
+        # Ensures that anaconda has the latest storage configuration data
+        m = self.machine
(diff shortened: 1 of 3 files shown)


ui/webui/test/helpers/storage.py
```diff
@@ -91,6 +91,18 @@ def wait_no_disks_detected(self):
...
+        task = self.machine.execute(f'dbus-send --print-reply --bus="{self._bus_address}" \
+            --dest={STORAGE_SERVICE} \
+            {STORAGE_OBJECT_PATH} \
+            {STORAGE_INTERFACE}.ScanDevicesWithTask')
+        task = task.splitlines()[-1].split()[-1]
+
+        self.machine.execute(f'dbus-send --print-reply --bus="{self._bus_address}" \
+            --dest={STORAGE_SERVICE} \
+            {task} \
+            org.fedoraproject.Anaconda.Task.Start')
```
jelly, line 104: I would recommend to switch new things over to busctl as that's part of systemd so always available. However this works so it's fine :)

jelly (approved):

rvykydal (approved):
Looks good to me.

VladimirSlavik (approved):
Looks good to me, thank you!

ui/webui/test/anacondalib.py
```diff
@@ -27,11 +27,25 @@
...
+from storage import Storage
+
 
 class VirtInstallMachineCase(MachineCase):
     MachineCase.machine_class = VirtInstallMachine
 
     def setUp(self):
+        # FIXME: running this in destructive tests fails because the SSH session closes before this is run
+        if self.is_nondestructive():
+            self.addCleanup(self.rescanStorage)
```
M4rtinK, line 39: Maybe we could print/log a warning when someone calls this by mistake in destructive mode ?

M4rtinK (approved):
Looks good to me as well. :)

ui/webui/test/anacondalib.py
KKoukiou: Hmm i don't see how I can know if rescanStorage will be added with an addCleanup in the future

KKoukiou:
Permian tests are failing https://github.com/rhinstaller/anaconda/actions/runs/5541527226

PR #4885, opened 2023-07-06
webui: pin down cockpit-* packages versions for gating purposes 

KKoukiou:
# MAIN IS BROKEN WITHOUT THIS 

This is caused because of different python version between fedora-rawhide-boot iso and current fedora-rawhide RPM builders. This can be reverted once we refresh the rawhide-boot image ISO https://github.com/cockpit-project/bots/pull/4976

### DIFF
--- a/ui/webui/test/build-rpms
+++ b/ui/webui/test/build-rpms
@@ -11,7 +11,7 @@ import sys
 BOTS_DIR = os.path.realpath(f'{__file__}/../../bots')
 sys.path.append(BOTS_DIR)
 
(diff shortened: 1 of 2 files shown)


VladimirSlavik (approved):
LGTM, thank you.

That said, I think the repository feels like a trashbin as we wildly flip left and right how to do this.

KKoukiou:
This is caused because of different python version between fedora-rawhide-boot iso and current fedora-rawhide RPM builders. This can be reverted once we refresh the rawhide-boot image ISO https://github.com/cockpit-project/bots/pull/4976

PR #4887, opened 2023-07-07
Revert "webui: pin down cockpit-* packages versions for gating purposes"

KKoukiou:


VladimirSlavik (approved):

jkonecny12 (approved):

VladimirSlavik:
It would be kind of nice to the rest of the team to note here what this really is - the counterpart to #4885 for when things work again. What does it mean, work again? Nobody knows. Instead we get _No description provided_ and have to read the internal chat to maybe find out some bits of this...

KKoukiou:
> It would be kind of nice to the rest of the team to note here what this really is - the counterpart to #4885 for when things work again. What does it mean, work again? Nobody knows. Instead we get _No description provided_ and have to read the internal chat to maybe find out some bits of this...


I don't think you will find any information in the internal chat, as nobody aside from me was working when this broke CI. 
All the information that are  available is in the original PR commit (the PR that this one reverts).

KKoukiou:
Actually I don't want to revert this - because having these gated is the only way we can ensure stable CI. And exactly this two lines have broken us at least once the past month. 

In 1 month with f39 release the webui dependencies will be inside the isos, and the cockpit deps will not need to be installed separately.

PR #4888, opened 2023-07-07
webui: simplify usage of dbus API for running tasks

KKoukiou:


KKoukiou:
Oh i see why this is not possible.

PR #4889, opened 2023-07-07
webui: Hide the language selection screen on Live images

adamkankovsky:


ui/webui/src/components/AnacondaWizard.jsx
```diff
@@ -75,12 +75,6 @@ export const AnacondaWizard = ({ dispatch, storageData, localizationData, onAddE
     const isBootIso = conf["Installation System"].type === "BOOT_ISO";
 
     const stepsOrder = [
-        {
```
KKoukiou, line 78: I would prefer inline. unshift works fine but the changed order makes this harder to read.

See here how to conditionally have an item in an array.
```
const arr = [
  ...(cond ? ['a'] : []),
  'b',
];
```

ui/webui/src/components/AnacondaWizard.jsx
```diff
@@ -144,7 +147,7 @@ export const AnacondaWizard = ({ dispatch, storageData, localizationData, onAddE
     const flattenedStepsIds = getFlattenedStepsIds(stepsOrder);
 
     const { path } = usePageLocation();
-    const currentStepId = path[0] || "installation-language";
+    const currentStepId = !isBootIso ? path[0] || "installation-language" : path[1] || "storage-devices";
```
KKoukiou: path contains the currnt URL path. path[1] is wrong here.  You still need path[0].

ui/webui/src/components/AnacondaWizard.jsx
```diff
@@ -331,7 +334,7 @@ const Footer = ({
             <WizardContextConsumer>
                 {({ activeStep, onNext, onBack }) => {
                     const isBackDisabled = (
-                        activeStep.id === "installation-language"
+                        activeStep.id === "installation-language" || (activeStep.id === "storage-devices" && isBootIso)
```
KKoukiou, line 337: Please create a const variable isFirstScreen.

KKoukiou (changes requested):

ui/webui/src/components/AnacondaWizard.jsx
adamkankovsky: I tried to do it this way too, but any empty value causes inconsistency in the UI. So I find another way to do this.

KKoukiou:
I see one issue with this PR. @adamkankovsky currently we only test against the boot.iso. With your PR language page goes away for the boot.iso image. So we will stop testing the language page basically. That's a no go - so we need to think of a workaround.

PR #4891, opened 2023-07-10
webui: Hide the language selection screen on Live images

adamkankovsky:


KKoukiou (commented):
There is one remaining issue still https://github.com/rhinstaller/anaconda/pull/4889#issuecomment-1625492741


We need to fake /run/anaconda/anaconda.conf for the language tests  so that ` const isBootIso = conf["Installation System"].type === "BOOT_ISO";` will be truthy.

KKoukiou (commented):
So as you correctly pointed out - language screen still appears for the tests are they are all running against the boot.iso.

What you can do is add a new test, which tweaks the conf file to mock the boot.iso so that the code thinks it's a live image.

ui/webui/src/components/AnacondaWizard.jsx
```diff
@@ -330,8 +334,11 @@ const Footer = ({
         <WizardFooter>
             <WizardContextConsumer>
                 {({ activeStep, onNext, onBack }) => {
+                    const isFirstScreen = (
+                        activeStep.id === "installation-language" || (activeStep.id === "storage-devices" && isLiveOS)
+                    );
                     const isBackDisabled = (
-                        activeStep.id === "installation-language"
+                        isFirstScreen
                     );
```
VladimirSlavik, line 342: You can actually have just `const isBackDisabled = isFirstScreen;`. So then - do we need both `isBackDisabled` and `isFirstScreen` when only one is used later in the code...?

VladimirSlavik (commented):
From my perspective as a mostly backend person, LGTM...

ui/webui/src/components/AnacondaWizard.jsx
```diff
@@ -73,14 +73,17 @@ export const AnacondaWizard = ({ dispatch, storageData, localizationData, onAddE
 
     // On live media rebooting the system will actually shut it off
     const isBootIso = conf["Installation System"].type === "BOOT_ISO";
+    const isLiveOS = conf["Installation System"].type === "LIVE_OS";
```
KKoukiou, line 76: Are there more options than liveos and bootiso? If not we just need one variable - not two.

adamkankovsky: There is a BOOTED_OS, but it cant happen in the webui. So i can remove isLiveOS line.

ui/webui/test/check-language
```diff
@@ -114,5 +114,20 @@ class TestLanguage(anacondalib.VirtInstallMachineCase):
...
+    def testLanguageScreenHiding(self):
+        self.restore_file('/run/anaconda/anaconda.conf')
+        b = self.browser
+        m = self.machine
+        i = Installer(b, m)
+
+        m.execute("sed -i 's/type = BOOT_ISO/type = LIVE_OS/g' /run/anaconda/anaconda.conf")
+
+        # Need to wait for storage-devices step because installation-language was hidden
+        i.open(step="storage-devices")
```
KKoukiou, line 126: So you want to test that opening without specifying which step which automatically pick the storage-devices step. So do i.open(step='')

ui/webui/test/check-language
```diff
@@ -114,5 +114,20 @@ class TestLanguage(anacondalib.VirtInstallMachineCase):
...
+
+        # Expect Storage devices page
```
KKoukiou, line 128: Instead of checking the contents of the page it is cleaner to:
self.assertEqual(self.get_current_page(), i.steps.STORAGE_DEVICES)

ui/webui/test/helpers/installer.py
```diff
@@ -88,6 +88,16 @@ def back(self, should_fail=False, subpage=False):
         else:
             self.wait_current_page(self.steps[current_step_id-1])
 
+    @log_step()
```
KKoukiou, line 91: Please don't log this, it's adding unnecessary noise.

ui/webui/test/helpers/installer.py
```diff
@@ -88,6 +88,16 @@ def back(self, should_fail=False, subpage=False):
...
+    def check_back_disabled(self, disabled=True):
```
KKoukiou, line 92: I don't see this beign re-usable as it's something we don't need to test again. I would not add it to helpers, just add this one line check in the test.

KKoukiou (changes requested):

ui/webui/src/components/AnacondaWizard.jsx
```diff
@@ -330,9 +332,10 @@ const Footer = ({
         <WizardFooter>
             <WizardContextConsumer>
                 {({ activeStep, onNext, onBack }) => {
-                    const isBackDisabled = (
```
KKoukiou, line 333: isBackDisabled and isFirstScreen are identical. PLs keep just one.

KKoukiou (changes requested):

ui/webui/test/check-language
adamkankovsky: First line of this function run well but ```self.wait_current_page(step)``` cant work with empty string. But we need to wait for the page there. Maybe original ```i.open(step="storage-devices")``` is there right because we cant wait at BootIso for this page without clicking on the next button.

KKoukiou: Since this fakes the media to looks like live and not boot.iso, there is never the next to click next as there is not welcome page shown. 

Please sent the latest version of the code and I can take a look how it fails.

adamkankovsky: Its its too difficult for this type of test update helpers. If i use ```i.open(step="storage-devices")``` it can only end with timeout if its not right.

ui/webui/test/check-language
adamkankovsky: I send my latest code here, but i dot do this step in the code. Because if we use original i.open i cant be anything else than storage-devices or timeout.

KKoukiou (approved):

PR #4893, opened 2023-07-10
Fix unit tests for python 3.12

VladimirSlavik:
This updates the parts of unit tests that get broken by rawhide's update to Python 3.12.

The breakage is not visible on PRs since these run the old container with Python 3.11. The breakage is visible in container daily updates.

Pylint is not included, because that is its own thing. That will likely require using patched pylint and astroid from rawhide rpms.

~~Translation canary needs its own PR and such, see https://github.com/rhinstaller/translation-canary/pull/7~~

### DIFF
--- a/tests/unit_tests/pyanaconda_tests/modules/payloads/payload/test_flatpak_manager.py
+++ b/tests/unit_tests/pyanaconda_tests/modules/payloads/payload/test_flatpak_manager.py
@@ -349,12 +349,12 @@ class FlatpakTest(unittest.TestCase):
         flatpak.replace_installed_refs_remote("cylon_officer")
 
         # test that every file is read and written
-        open_mock.has_calls([
+        open_mock.assert_has_calls([
             call("/path/app/org.test/x86_64/stable/active/deploy", "rb"),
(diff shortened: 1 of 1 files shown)


translation-canary/translation_canary/translatable/__init__.py
```diff
@@ -36,14 +36,17 @@
 
 # Gather tests from this directory
 import pkgutil
+import importlib
```
vojtechtrefny, line 39: These changes needs to go to https://github.com/rhinstaller/translation-canary, the translation-canary folder here is just a git subtree and we want this fix for other projects that use it too.

vojtechtrefny (changes requested):

translation-canary/translation_canary/translatable/__init__.py
VladimirSlavik: Done: https://github.com/rhinstaller/translation-canary/pull/7

vojtechtrefny: Thanks!

vojtechtrefny (approved):

PR #4894, opened 2023-07-10
infra: Group dependabot updates

VladimirSlavik:
See https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#groups


M4rtinK (approved):
Looks good to me, thanks! :)

Still sucks it needs to be manually set up & can't somehow figure it out automatically, but still an improvement. :)

PR #4900, opened 2023-07-11
webui: tests: restart cockpit-ws between tests

KKoukiou:
It has been noticed that cockpit-ws might present the login page now and then after multiple test runs against the same test VM.


M4rtinK (approved):
Looks good to me.

PR #4903, opened 2023-07-11
infra: Temporarily use pylint and astroid from RPM

VladimirSlavik:
Pip has too old versions that don't work with Rawhide Python 3.12.

----

Draft because this is how a potential fix looks like, except the updated packages are not available yet.

There's also some unclear information about the problem (which problem?) being in Python instead of pylint.

PS: As of submitting the PR, pylint is the only thing that fails on a freshly built CI container.

### DIFF
--- a/anaconda.py
+++ b/anaconda.py
@@ -85,6 +85,8 @@ def exitHandler(rebootData):
 
         if flags.eject or rebootData.eject:
             for device_path in optical_media:
+                # path is already imported in __main__ as other things used here -> disable pylint
+                # pylint: disable=undefined-loop-variable
                 if path.get_mount_paths(device_path):
                     util.dracut_eject(device_path)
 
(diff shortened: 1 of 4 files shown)


jkonecny12:
I've updated the change and finished the rest of the issues. With this update pylint should work as expected.

jkonecny12:
This behaves differently than what I'm running locally. I'll find out more about it.

VladimirSlavik:
Thank you! Almost there.

A few ideas:
- My first choice was to just install rpm over the pip install, but if it doesn't work, maybe try commenting out the entries in requirements.txt to get only the RPMs?
- Perhaps print also version of dill, if you can. That was a problem in the past too. It's an astroid dependency.
- Pylint issues here on Github mention a Python interpreter bug as a cause of some astroid crashes. It might be that we're seeing that :-/

PR #4904, opened 2023-07-11
webui: pin down cockpit-ws and cockpit-bridge versions before the python bridge release

KKoukiou:


M4rtinK (approved):
Looks good to me - seems like the best option for now.

PR #4907, opened 2023-07-12
infra: Clone repo for cockpit lib update with token

VladimirSlavik:
This addresses https://github.com/rhinstaller/anaconda/pull/4790#issuecomment-1564673883

It's not clear if this solves the problem.

Currently blocked because we can't update.

### DIFF
--- a/.github/workflows/cockpit-lib-update.yml
+++ b/.github/workflows/cockpit-lib-update.yml
@@ -5,6 +5,7 @@ on:
     - cron: '0 2 * * 4'
   # can be run manually on https://github.com/rhinstaller/anaconda/actions
   workflow_dispatch:
+
 jobs:
   cockpit-lib-update:
     permissions:
@@ -24,6 +25,8 @@ jobs:
 
       - name: Clone repository
         uses: actions/checkout@v3
+        with:
+          token: ${{ secrets.INSTALLKER_TOKEN }}
 
       - name: Run cockpit-lib-update
         run: |


.github/workflows/cockpit-lib-update.yml
```diff
@@ -24,6 +25,8 @@ jobs:
 
       - name: Clone repository
         uses: actions/checkout@v3
+        with:
+          token: ${{ secrets.INSTALLKER_TOKEN }}
```
KKoukiou, line 29: I think you have a typo here? INSTALLKER?

Also the fact that we can't update is not making this PR blocked. We can nicely generate the PRs just not merge them.

KKoukiou (commented):
All in all once this is tested I am ok that it get's merged even though we should not update for the time being.

.github/workflows/cockpit-lib-update.yml
VladimirSlavik: That's the bot name, so its token has its name.

VladimirSlavik:
Yup, not a high priority, just wanted to dump it here as a PR instead of a random comment hidden somewhere.

M4rtinK (approved):
Looks good to me. :)

VladimirSlavik:
I think let's try and see what this does. Usually, we catch the issues with opening PRs later after we deploy the code.

The interaction with other tokens used throughout the whole process is unclear, so I don't see testing it as economical...

PR #4909, opened 2023-07-12
infra: Enable @storage/udisks-daily Copr repo for CI images

vojtechtrefny:
The blivet-daily Copr repository contains latest release of libblockdev which cannot be installed together with udisks2 (which Anaconda depends on) so the result is that incompatible versions of blivet, libblockdev and udisks2 are installed together in the image. Adding the repository that also includes daily builds of udisks2 should fix this.

### DIFF
--- a/dockerfile/anaconda-ci/Dockerfile
+++ b/dockerfile/anaconda-ci/Dockerfile
@@ -43,6 +43,7 @@ RUN set -ex; \
     BRANCH=${BRANCH#fedora-}; \
     dnf copr enable -y ${copr_repo} fedora-${BRANCH}-x86_64; \
     dnf copr enable -y @storage/blivet-daily fedora-${BRANCH}-x86_64; \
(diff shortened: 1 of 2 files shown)


jkonecny12 (approved):
Looks good to me. Thanks for the improvement.

jkonecny12 (changes requested):
OK, you need to change also https://github.com/rhinstaller/anaconda/blob/master/dockerfile/anaconda-rpm/Dockerfile#L47

vojtechtrefny:
> OK, you need to change also https://github.com/rhinstaller/anaconda/blob/master/dockerfile/anaconda-rpm/Dockerfile#L47

Fixed.

M4rtinK (approved):
Looks good to me. :)

VladimirSlavik (approved):
Thank you!

I wonder if packit and tmt need this too, but I think not. They don't certainly have the other coprs we use.

PR #4910, opened 2023-07-12
webui: fix disk selection not working when moving back to the disk selection screen

KKoukiou:


ui/webui/src/components/storage/InstallationDestination.jsx
```diff
@@ -151,7 +151,7 @@ const DropdownBulkSelect = ({
           key="select-multiple-split-checkbox"
           id="select-multiple-split-checkbox"
           aria-label={_("Select multiple disks")}
-          isChecked={selectedCnt > 0 ? (selectedCnt === totalCnt ? true : null) : false}
+          isChecked={selectedCnt === totalCnt}
```
rvykydal, line 154: So we want this UI change - to get rid of this bulk selection checkbox state (some disks (neither all nor none) selected in this PR)?
Screenshot from 2023-07-13 09-47-37

rvykydal (commented):
Looks good to me if the UX change is intentional.

ui/webui/src/components/storage/InstallationDestination.jsx
KKoukiou: oops good catch, how the tests even passed here? I will remove this commit.

KKoukiou: I think the checkbox should be indeed only checked if all are checked.

rvykydal (approved):
Thank you.

ui/webui/src/components/storage/InstallationDestination.jsx
rvykydal: > oops good catch, how the tests even passed here? I will remove this commit.

I am afraid I just didn't write a test for this case. I'll add it.

PR #4913, opened 2023-07-13
Improve btrfs support in tui mountpoint assignment v2

vojtechtrefny:
This is intended as replacement of #4835

The biggest change is we use device name instead of path as a unique identifier, this way we can display btrfs subvolumes without any additional identifier and changes.


rvykydal:
Looks good to me.

PR #4914, opened 2023-07-13
Enable ruff pylint checks

VladimirSlavik:
This enables lots of checks for ruff, hopefully bringing it closer to what we have with pylint.

See also https://beta.ruff.rs/docs/rules/

The list of ignored checks is partially a masked TODO - there are some valid detections with these. However, fixing such detections is either a too wide change, or would modify behaviour enough to warrant retesting the application.


vojtechtrefny (approved):
LGTM

VladimirSlavik:
rebased again...

PR #4915, opened 2023-07-13
webui: tests: port helpers to busctl and unselect all disks on teardown

KKoukiou:


KKoukiou:
I consider we should create some abstraction layer over dbus calls from python test code in the future as it seems these are very useful.

VladimirSlavik (approved):
Thank you, looks good to me.

I am starting to have second thoughts about manipulating the backend so freely, but that's not the change here, so never mind.

rvykydal (approved):
Thank you

PR #4916, opened 2023-07-14
infra: Check container age in testing make targets

VladimirSlavik:
With this, users have some chance to realize what's wrong when the container is too old.

### DIFF
--- a/scripts/testing/check-container-age.sh
+++ b/scripts/testing/check-container-age.sh
@@ -1,0 +1,23 @@
+#!/usr/bin/bash
+# container image name, eg. "quay.io/rhinstaller/anaconda-ci:master"
+container_name="$1"
+# when is an image too old - string for `date --date=`
+too_old_description="60 hours ago"
+
+container_created=$(podman image inspect --format '{{.Created}}' "$container_name")
+# dates in unix epoch seconds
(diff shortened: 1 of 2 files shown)


scripts/testing/check-container-age.sh
```diff
@@ -0,0 +1,19 @@
...
+# when is an image too old - string for `date --date=`
+too_old_description="60 hours ago"
+
+# dates in unix epoch seconds
+ts_container_created=$(date --date="$(podman image inspect --format '{{.Created}}' $container_name | awk '{print($1 " " $2)}')" +%s)
+ts_too_old=$(date +%s --date="$too_old_description")
+
+if [ "$ts_too_old" -gt "$ts_container_created" ] ; then
+  echo "====================================================================="
+  echo "WARNING: Container is too old: $container_name"
```
vojtechtrefny, line 13: It might good to add the container age here, something like *WARNING: Container is too old: $container_name ($age)*

vojtechtrefny (approved):
Looks good to me.

M4rtinK (approved):
Looks good to me as well - a really nice improvement, thanks! :)

scripts/testing/check-container-age.sh
VladimirSlavik: Unfortunately this is almost impossible. Apparently I would have to calculate this myself, or install something.

PR #4917, opened 2023-07-14
Add possibility to run pylint-only test

jkonecny12:
Pylint was recently removed from the default test suite because it caused issues on python 3.12. To be able to verify if that issue was resolved and in general improve state of running just this test this change introduce a possibility to run just pylint test in container or without the container.

Usually this should be solved by `make check TESTS=<pylint>`, unfortunately this is not possible for pylint for unknown reasons. The same situation works for unit tests but when pylint is used this way it not even try to start the tests. Seems to me like some automake magic fails here. Even automake substition doesn't work here. Seems that pylint test has to be in the TESTS variable hardcoded which can't happen if it is not part of the default runs.

Anyhow, to resolve this issue a new script was introduced to set environment variables correctly to the pylint start so it's easy and consistent to start the pylint tests.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -299,6 +299,9 @@ container-shell:
 	--cap-add=SYS_CHROOT \
 	$(CI_NAME):$(CI_TAG)
 
+container-pylint-only:
+	$(MAKE) -f ./Makefile.am container-ci CI_CMD="make tests-pylint"
+
 container-rpm-test:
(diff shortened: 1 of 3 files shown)


VladimirSlavik (approved):
Thank you very much!

jkonecny12:
The test failures are not related - merging.

PR #4918, opened 2023-07-14
webui: Add Firefox theme for use on Live media

M4rtinK:
The theme upstream is here: https://github.com/garrett/firefox-theme-frame

Its main aim is to shape the Firefox window to the most suitable state on Live images:
- has a header bar for window switching
- has no tab/status/URL bars
- has no browser controls

As Firefox needs the profile folder to be writable, we need to copy the profile template to a writable tempdir & use that as the profile path for Firefox.

To simplify things, we also use the same theme on non Live images for now, with the expectation that a tweaked theme will provided later.

**TODO**
- [x] check the license if fine (https://github.com/garrett/firefox-theme-frame/blob/main/LICENSE) - is this compatible with our GPLv2 + MIT ?
- [x] drop the status bubble
- [x] drop the tab bar
- [x] fix the content not loading

### DIFF
--- a/ui/webui/webui-desktop
+++ b/ui/webui/webui-desktop
@@ -39,7 +39,6 @@ set -eu
 # exec_prefix= is set because the default /usr/libexec contains "${exec_prefix}"
 exec_prefix="/usr"
 libexecdir="/usr/libexec"
-
 if [ -z "${1:-}" ]; then
     echo "Usage: $0 <Cockpit path> [ssh host]" >&2
     exit 1
@@ -61,7 +60,27 @@ case "$1" in
         ;;
 esac
 
-BROWSER="/usr/bin/firefox --kiosk"
+# prepare empty firefox profile dir with theme based on the passed profile id
+FIREFOX_THEME_DIR="/usr/share/anaconda/firefox-theme"
+FIREFOX_PROFILE_PATH="/tmp/anaconda-firefox-profile"
+THEME_ID=$2
+
+# make sure the profile directory exists and is empty
+if [ -d ${FIREFOX_PROFILE_PATH} ]
+then
+    echo "Cleaning up existing Anaconda Firefox profile directory."
+    rm -rf ${FIREFOX_PROFILE_PATH}
+fi
+mkdir -p ${FIREFOX_PROFILE_PATH}
+
+# populate the profile directory with our custom Firefox theme
+# - theme id is passed as the second argument of this script
+THEME_PATH="${FIREFOX_THEME_DIR}/${THEME_ID}"
+
+cp -a "${THEME_PATH}/." ${FIREFOX_PROFILE_PATH}
+
+# FIXME: is this hardcoded resolution necessary ?
+BROWSER="/usr/bin/firefox --new-instance --window-size 1024,768 --profile ${FIREFOX_PROFILE_PATH}"
 
 # start browser in a temporary home dir, so that it does not interfere with your real one
 BROWSER_HOME=$(mktemp --directory --tmpdir cockpit.desktop.XXXXXX)
--- a/anaconda.spec.in
+++ b/anaconda.spec.in
@@ -441,6 +441,15 @@ rm -rf \
 %{_datadir}/cockpit/anaconda-webui/manifest.json
 %{_datadir}/metainfo/org.cockpit-project.anaconda-webui.metainfo.xml
 %{_datadir}/cockpit/anaconda-webui/po.*.js.gz
+%dir %{_datadir}/anaconda/firefox-theme
+%dir %{_datadir}/anaconda/firefox-theme/default
+%dir %{_datadir}/anaconda/firefox-theme/default/chrome
+%{_datadir}/anaconda/firefox-theme/default/user.js
+%{_datadir}/anaconda/firefox-theme/default/chrome/userChrome.css
+%dir %{_datadir}/anaconda/firefox-theme/live
+%dir %{_datadir}/anaconda/firefox-theme/live/chrome
+%{_datadir}/anaconda/firefox-theme/live/user.js
+%{_datadir}/anaconda/firefox-theme/live/chrome/userChrome.css
 %{_libexecdir}/webui-desktop
 
 %files gui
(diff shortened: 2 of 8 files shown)


M4rtinK:
So far I've noticed a couple issues with this that will have to be resolved.

On Live images built via the Image builder pipeline, the Web UI shows up ~like we want it:
live_image_custom_theme_1

There are still a couple issues:

- you can notice there is a tab bar
- some sort of status bubble at the bottom
- language list is not populated (we have seen this in regular kiosk mode as well, so this is unrelated)
- the page is not clickable (possibly stuck in loading state ? as the status bubble seems to imply something is still being loaded from the server)

This was using the rather early post Python 3.12 live image: https://fedorapeople.org/groups/anaconda/webui_preview_image/x86_64/fedora-preview-live-installer-2023-07-11.iso

Thinking this might be some image related breakage, I've also tried using a regular Fedora Rawhide live image from July 12. This time the page content does not load at all (but its loading the correct page):

live_image_custom_theme_2
live_image_custom_theme_3
live_image_custom_theme_4

**HOW TO TEST**
The best option is to fetch a Live image - either one from the Image Builder pipeline or even a regular one.

Then you can add the following COPR which has packages corresponding to this PR as of right now:
https://copr.fedorainfracloud.org/coprs/m4rtink/anaconda-webui-firefox-theme/

Once you add the COPR, something like this should replace the on-media Anaconda with the one corresponding to this PR:

`sudo dnf install anaconda-999999999 --allowerasing`

Alternatively you can fetch the RPMs from the COPR (or build them yourself from the PR), `scp` them to the machine running your Live image & install them via DNF:

`sudo dnf install *.rpm --allowerasing`

Then start Anaconda in the usual way - either via its icon in the quick launch menu or via the `liveinst` script in `PATH`.

VladimirSlavik:
Why does this contain two themes?

KKoukiou (commented):
The default user and live user js code are mostly shared. Can't you create some common file and import it?

M4rtinK:
> The default user and live user js code are mostly shared. Can't you create some common file and import it?

They are identical right now, but the expectation is the themes will diverge as we adjust the `default` (essentially boot.iso) theme, which for example does not need a header bar.

As for sharing code with the theme - I've just copied the theme created by @garrett in https://github.com/garrett/firefox-theme-frame & don't really have any experience with how it is done. So someone with a deeper knowledge of Firefox theming will need to answer this. :)

M4rtinK:
> Why does this contain two themes?
The `live` theme is to be used for the Live images and the `default` theme is meant for everything else (but mainly boot.iso). Right now the themes are the same for both, as we don't have a boot.iso specific theme yet. In the future the `default` theme should be adjusted accordingly.

M4rtinK:
Updated the PR to move most of the Firefox theme handling logic to the `webui-desktop` script, which seems to be a more suitable place for it.

M4rtinK:
Unfortunately still stuck at the Web UI not (fully ?) loading:

webui_live_still_not_working

garrett:
It looks like there are two title bars. First: window manager provided titlebar. Second: CSD with tabs. The theme was made with CSD (client-side decorations) in mind, not the window manager titlebar.

We have two options:

1. Turn off tab bar entirely and rely on the window manager. However, we could not disable, remove, or hide the close button.
2. Force CSD mode to remove the window manager titlebar. (But I thought the theme was doing this already? I need to double-check the `user.js` file, I guess?)

> This time the page content does not load at all (but its loading the correct page):

That looks like it's due to a problem in the JS in Anaconda web UI, not in the theme.

---

I've updated the theme to remove the button, fix the titlebar, and to try harder to make new windows show up.

But after playing with it more, I think a more proper approach would be to hide the tab bar and use the window manager titlebar... but we'd have a close button (unless we use the semi-restricted environment that GNOME Initial Setup does... which _may_ disable the X icon).

KKoukiou:
The crash happens because the buildstamp file is not there. 
liveimg-crash


Is this file supposed to be always there? Should we handle it differently?

garrett:
:ballot_box_with_check: I relicensed the theme to MIT, as asked.

VladimirSlavik:
After thinking a bit about the themes, I realized the problem I have is that one is called "default". Perhaps "bootiso" would be better?

M4rtinK:
> After thinking a bit about the themes, I realized the problem I have is that one is called "default". Perhaps "bootiso" would be better?

Well, the way it is we kinda need to always have a theme to pick - the live image theme is specifically picked for the Live image & the "default" theme will be used for everything else (which indeed is mostly the boot iso).

M4rtinK:
> **TL;DR**: I've updated the theme. It should work much better now. (I hope!) wink

Thanks! Just updated the PR your the latest version of your theme - will post screenshots in a bit (& will generate a new boot.iso).

M4rtinK:
live_image_new_theme

M4rtinK:
Looking good IMHO. :)

garrett:
Whew. I'm happy to see the theme working as intended @ https://github.com/rhinstaller/anaconda/pull/4918#issuecomment-1651902869

> It's a spaghetti monster, but it's flying!

:rofl:

ui/webui/webui-desktop
```diff
@@ -61,7 +60,27 @@ case "$1" in
...
-BROWSER="/usr/bin/firefox --kiosk"
+# prepare empty firefox profile dir with theme based on the passed profile id
+FIREFOX_THEME_DIR="/usr/share/anaconda/firefox-theme"
+FIREFOX_PROFILE_PATH="/tmp/anaconda-firefox-profile"
+THEME_ID=$2
+
+# make sure the profile directory exists and is empty
+if [ -d ${FIREFOX_PROFILE_PATH} ]
+then
+    echo "Cleaning up existing Anaconda Firefox profile directory."
```
KKoukiou, line 71: In what scenario would the profile directory on /tmp already exist?

KKoukiou (approved):
I have no comments on the css applied. Please attach to the PR the latest screenshot for both live and boot iso.

ui/webui/webui-desktop
M4rtinK: If you launch the installer multiple times on Live media - quite a frequent occurrence during debugging & we know from logs attached to bug reports, that users do it quite often in real life, usually when encountering an error and trying again.

In any case, we really should start with empty profile or else I would expect the behavior on followup starts being non-deterministic given the sheer amount of stuff Firefox dumps to the originally empty folder.

PR #4919, opened 2023-07-14
webui: simplify webui-desktop script

KKoukiou:
Stop using coproc, nesting bash and calling unshare. All these are needed for the namespace isolation which we don't need in anaconda.

### DIFF
--- a/ui/webui/webui-desktop
+++ b/ui/webui/webui-desktop
@@ -40,42 +40,6 @@ set -eu
 exec_prefix="/usr"
 libexecdir="/usr/libexec"
 
-# find suitable browser, unless already set by $BROWSER
-# We can't use xdg-open, it does too much magic behind the back to connect to
-# existing instances (outside of our namespace) and does not allow us to reduce
-# the UI, or pass options like chromium's --no-sandbox.
-detect_browser()
-{
-    [ -z "${BROWSER:-}" ] || return 0
-
-    # First choice, but it depends on gi.repository WebKit2, so check it
-    if /usr/libexec/cockpit-client --help >/dev/null 2>/dev/null; then
-        BROWSER="/usr/libexec/cockpit-client --disable-uniqueness --no-ui --external-ws"
-        return 0
-    fi
-
-    for browser in chromium-browser chromium google-chrome; do
-        if type $browser >/dev/null 2>&1; then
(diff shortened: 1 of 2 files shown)


KKoukiou:
This fixes the crash but it seems that still the VM does not shutoff.

ui/webui/webui-desktop
```diff
@@ -97,28 +61,17 @@ case "$1" in
         ;;
 esac
 
-detect_browser
-
-# start the bridge; this needs to run in the normal user session/namespace
-coproc ${2:+ssh "$2"} cockpit-bridge
-trap "kill $COPROC_PID; wait $COPROC_PID || true" EXIT INT QUIT PIPE
+BROWSER="/usr/bin/firefox --kiosk"
```
M4rtinK, line 64: So I'm using this variable on my Firefox theme PR: https://github.com/rhinstaller/anaconda/pull/4918/files#diff-4384704581bd02e11beb389b584be16b175b80679194ac91422ed2701d8bd3bdR139

I wonder - now that we started changing it maybe this script is a better place for the theme handling than the UI handling Python code where I put it in my PR ? What do you think ? 

We might have to solve live/boot.iso detection somehow, but otherwise it should work.

KKoukiou: Yes it probably makes more sense that the theming gets applied here.

jkonecny12 (approved):
Less is most of the time better. Thanks, I love this fix!

ui/webui/webui-desktop
```diff
@@ -97,35 +61,24 @@ case "$1" in
...
 
 # start ws and browser in a detached network namespace
```
martinpitt, line 66: This comment is obsolete and wrong now.

martinpitt (commented):
Thanks! LGTM (mostly), but it still seems to [fail](https://cockpit-logs.us-east-1.linodeobjects.com/pull-4919-20230717-090841-25aecdfc-fedora-rawhide-boot/log.html)? Or is that unrelated?

KKoukiou:
fedora-rawhide-boot@bots#4976  is passing now.

ui/webui/webui-desktop
M4rtinK: OK, so I'll adjust https://github.com/rhinstaller/anaconda/pull/4918 accordingly. :)

PR #4923, opened 2023-07-17
webui: language information should be fetched also for live media

KKoukiou:


M4rtinK (approved):
Oh, so this fixes the case where on Live media the language list is not populated ? Nice! 

(With all the ongoing stuff, it sometimes hard for me to find which PR refers to which issue/effort due to lack of context.) 

Looks good to me code wise. :)

KKoukiou:
> Oh, so this fixes the case where on Live media the language list is not populated ? Nice!
> 
> (With all the ongoing stuff, it sometimes hard for me to find which PR refers to which issue/effort due to lack of context.)
> 
> Looks good to me code wise. :)

On live media the language screen is hidden, but we need still to load the languages because of the review screen. Prevoysly the review screen would just crash on live media - because of the missing localization data.

M4rtinK:
> On live media the language screen is hidden, but we need still to load the languages because of the review screen. Prevoysly the review screen would just crash on live media - because of the missing localization data.

Oh, right - thanks for the explanation! :)

VladimirSlavik (approved):
Looks good to me. Thank you!

PR #4924, opened 2023-07-18
webui: allow critical errors also in the first load of the application

KKoukiou:


M4rtinK (approved):
Looks good to me - nice to see early stage better covered. :)

ui/webui/src/components/app.jsx
```diff
@@ -105,7 +105,7 @@ export const Application = () => {
     };
 
     // Postpone rendering anything until we read the dbus address and the default configuration
-    if (!address || !conf || !beta || !prettyName || !storeInitilized) {
+    if (!criticalError && (!address || !conf || !beta || !prettyName || !storeInitilized)) {
```
rvykydal, line 108: In case of criticalError (set in the useEffect hook) it is possible that conf in isBootIso definition is undefined below (Uncaught TypeError: conf is undefined).

PR #4925, opened 2023-07-18
webui: design adjustment on the disk encryption screen

KKoukiou:


M4rtinK:
Also adding the screenshots from pixel there for convenience:

Old way
old_1
old_2

New way
new_1
new_2

M4rtinK (approved):
Looks good to me - nice improvements! :)

garrett (approved):
Thanks! Looks good!

I'd suggest we might make the passhprase entry a bit longer (not the full width, but wider than the after). Some people have long (and sometimes even very long) passphrases!

rvykydal (approved):
Looks good to me.
I wish I could remember why the design used 2 steps back then. I guess we were expecting more options on the first screen, like default autopartitioning scheme (btrfs, plain, ...) was in play maybe at some point, but nevermind, we may have been trying to see too far.

PR #4927, opened 2023-07-18
Handle missing buildstamp on live media

M4rtinK:
This PR handles the issue of /.buildstamp not existing on Live images by introducing a DBus interface for product information & Web UI then using this interface.

~The backend commits from Vlada are preliminary at this points & might be rolled into a separate PR.~


M4rtinK:
So only a small issue with this - it does not work & I can't really figure out why so far.

Basically, our early initialization hook reacted to the crash & displayed a nice error dialog:
Screenshot 2023-07-18 at 19-25-53 Anaconda Web UI

Unfortunately that is also the extent of information I get at this point - there is nothing else in Journal or in JS console. :P 

That makes it a bit hard to find out what is wrong right now, but it also a good data for for the ongoing error reporting work. :)

ui/webui/src/components/app.jsx
```diff
@@ -71,18 +72,18 @@ export const Application = () => {
             startEventMonitorStorage({ dispatch });
 
             startEventMonitorLocalization({ dispatch });
+
+            getIsFinal().then(
```
rvykydal, line 76: Is setting Beta moved to the bus.address change handler intentionally?

KKoukiou (commented):
You can remove the data-debug from the test. (see failures) Its not much interesting anyhow.

ui/webui/src/apis/runtime.js
```diff
@@ -0,0 +1,61 @@
...
+ * @returns {Promise}           Reports if the given OS release is considered final
+ */
+export const getIsFinal = () => {
+    return (
+        new RuntimeClient().client.call(
+            "/org/fedoraproject/Anaconda/Modules/Runtime/UIInterface",
+            "org.freedesktop.DBus.Properties",
+            "Get",
+            [
+                "org.fedoraproject.Anaconda.Modules.Runtime.UIInterface",
```
rvykydal, line 55: UIInterface -> UserInterface

ui/webui/src/apis/runtime.js
```diff
@@ -0,0 +1,61 @@
...
+}
+
+/**
+ *
+ * @returns {Promise}           Reports if the given OS release is considered final
+ */
+export const getIsFinal = () => {
+    return (
+        new RuntimeClient().client.call(
+            "/org/fedoraproject/Anaconda/Modules/Runtime/UIInterface",
```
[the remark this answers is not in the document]
rvykydal: https://github.com/M4rtinK/anaconda/blob/69a08b857af217756ecf3980f6282d2f7ffa1417/pyanaconda/modules/common/constants/objects.py#L28

rvykydal:
With the name update the getIsFinal fails with
```
do not specify "name" option in dbus channel when "internal"
```
(which is not handled as CriticalError because of this issue: https://github.com/rhinstaller/anaconda/pull/4924#discussion_r1267847065)

rvykydal:
> With the name update the getIsFinal fails with
> 
> ```
> do not specify "name" option in dbus channel when "internal"
> ```
> 


Ah, that is probably the reason why the getIsFinal was called from the bus.address change handler (after the client is initialized).

ui/webui/src/components/app.jsx
M4rtinK: I noticed other DBus calls there, so I put it there. Placing it where the "old" bootstamp checking code was did not work at all - I guess that place might be run somehow before the DBus machinery is initialized ?

M4rtinK:
Updated the PR to only contain the Web UI part & switching from draft to regular PR. Still, for whatever reason, I end up with a white screen - it even affects the generated image (you can see for your self). worse still, there is nothing relevant in either Journal or browser console. :P

Any hints welcome at this point...

ui/webui/src/components/app.jsx
```diff
@@ -36,15 +36,15 @@ import { BossClient } from "../apis/boss.js";
...
 
-import { readBuildstamp, getIsFinal } from "../helpers/betanag.js";
 import { readConf } from "../helpers/conf.js";
 import { useReducerWithThunk, reducer, initialState } from "../reducer.js";
 
 export const Application = () => {
     const [address, setAddress] = useState();
     const [criticalError, setCriticalError] = useState();
-    const [beta, setBeta] = useState();
+    const [beta, setBeta] = useState(null);
```
KKoukiou, line 47: No need for null, it's by default 'undefined'

ui/webui/src/components/app.jsx
```diff
@@ -105,7 +106,7 @@ export const Application = () => {
     };
 
     // Postpone rendering anything until we read the dbus address and the default configuration
-    if (!criticalError && (!address || !conf || !beta || !prettyName || !storeInitilized)) {
+    if (!criticalError && (!address || !conf || beta == null || !prettyName || !storeInitilized)) {
```
KKoukiou, line 109: === undefined and you don't need the override.

ui/webui/src/components/app.jsx
KKoukiou: Indeed - the place that this was before  is not guaranteed to run after address is set.

KKoukiou (approved):

PR #4929, opened 2023-07-19
Accept conf being undefined (yet) during a CriticalError

rvykydal:


KKoukiou (commented):
I find the webui prefix in commit messages - rather nice to use. I would suggest to keep it for the future.  (Or any kind of prefix that shows it's webui related)

PR #4930, opened 2023-07-19
webui: tests: Fix WebUI end2end tests

velezd:
This should fix e2e webui tests.


KKoukiou (commented):
I don't really understand this change. There is a single button containing 'Reboot' string on the page, so why you need to make the selector more specific.

velezd:
There isn't a single button. This is sort of a workaround. The page contains footer hidden by css that has Back, Next and Reboot buttons in it.

M4rtinK (approved):
Looks good to me. :)

PR #4931, opened 2023-07-19
Publish isFinal on D-Bus

VladimirSlavik:
This is a temporary change to make this available to webui until the "product stuff" is properly modularized.

Blocks #4927.

PR #4932, opened 2023-07-19
Product cleanups

VladimirSlavik:
Found while working on #4931

edit: Oops, the UI constructor parameters are used by initial setup. Lovely.

### DIFF
--- a/pyanaconda/core/constants.py
+++ b/pyanaconda/core/constants.py
@@ -33,11 +33,7 @@ ADDON_PATHS = ["/usr/share/anaconda/addons"]
 from pyanaconda import product
 productName = product.productName
 isFinal = product.isFinal
-
-# for use in device names, eg: "fedora", "rhel"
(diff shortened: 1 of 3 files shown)


pyanaconda/core/constants.py
```diff
@@ -32,7 +32,6 @@
 # common string needs to be easy to change
 from pyanaconda import product
 productName = product.productName
```
KKoukiou, line 34: You can also delete this variable then?

KKoukiou (approved):

pyanaconda/core/constants.py
VladimirSlavik: Not really. There are dependencies on this stuff all over the place, including initial setup. I filed a new task which details how to address all of this, but actually doing it needs to wait a bit.

rvykydal (approved):

PR #4940, opened 2023-07-21
webui: tests: make some more tests non-destructive

KKoukiou:


ui/webui/test/check-storage
```diff
@@ -438,8 +441,8 @@ class TestStorageMountPoints(anacondalib.VirtInstallMachineCase):
         sgdisk --new=0:0:+10GiB {disk}
         sgdisk --new=0:0:0 {disk}
         mkfs.ext4 {disk}2
```
jelly, line 443: Hmm `mkfs.ext4` does not require an `-f`. I would suspect `wipefs` already takes care of everything so no `--force` should be required right?

KKoukiou: It does not even provide the -f .

KKoukiou: mkfs.ext4: invalid option -- 'f'

KKoukiou: ah it would need -F, but anyhow it does not fail. Whilst the other mkfs would fail even after wipefs

jelly: Aha! TIL I thought all mkfs options where equal.

jelly (approved):

PR #4942, opened 2023-07-24
Use Critical error dialog with context

rvykydal:
The draft is adding some context for the user to Critical Error dialog addressing https://issues.redhat.com/browse/INSTALLER-3562. I am opening it to check if the direction and approach makes sense at all (I feel there already must be a common way how to do such things.). It is a very early draft, don't miss the notes below.

The goal here is to inform user more than the exception text says. In this PR I add:
- context: what was the acction performed when the error ocured
- hint: (more optional) - what can user do (usually not much in case of critical errors?)

Notes:
- The wording / presentation of the messages is to be discussed (like using just normal phrases). For example: "Initialize storage model" -> "Reading information about disks." (closer to user pov). ~~Or "While reading information about disks this error appeared: "Error: TEST DBUS EXCEPTION". This should not happen ..."~~
- In the PR there are both the new `onCritFail` and existing `onAddErrorNotification` (Just exception message in notification). I think in most cases we want to replace the notification with Critical Error dialog (needs to be audited place by place). ~~Here I am replacing it just in progress component as an example.~~
- ~~There is a rough stab on passing the handler to Actions.~~
- We would like to add also some information for the developer to find the error origin easily, like source file reference if it is possible to do in automated way (the user context messages are translated so they might be hard to find).
- And add a link to BZ reporting to the dialog but it is another independent issue: UPDATE: this is done later below in this PR

Screenshot from 2023-07-24 10-09-54


ui/webui/src/components/app.jsx
```diff
@@ -67,26 +73,32 @@ export const Application = () => {
...
+            const errorContext = {
+                context: _("Initialize storage model."),
+                hint: _("This should not happen, the installation is doomed."),
+            };
+
+            const errorHandler = onCritFail(errorContext);
+
             Promise.all([
-                initDataStorage({ dispatch }),
+                initDataStorage({ dispatch, errorHandler }),
```
KKoukiou, line 84: I don't see the reason to pass the 'errorHandler' here. Any exception that happens inside this function, if the promises are correctly chained should be possible to catch in the end of the chain, with a catch or as a second argument to 'then', as it's already done is line 91.

ui/webui/src/components/Error.jsx
```diff
@@ -48,11 +48,30 @@ export const CriticalError = ({ exception, isBootIso }) => {
                   </Button>
               </>
           }>
+            {exception.contextData?.context &&
             <TextContent>
                 <Text component={TextVariants.p}>
-                    {exception.message}
+                    {cockpit.format("Action: $0", exception.contextData.context)}
```
KKoukiou, line 54: @garrett should take a look at this with an example exception data.

KKoukiou: Also these should be translated.

ui/webui/src/components/app.jsx
rvykydal: The point is to pass the handler to the getDeviceDataAction so that the exceptions caught here: https://github.com/rvykydal/anaconda/blob/a415a9f093d1661e6a9df8fe31f6699018584d84/ui/webui/src/actions/storage-actions.js#L61 are handled by it instead by console.error. (Or we need to remove the catch from the action).

KKoukiou: I think we should remove it then. If we agree that this action if failed is a critical error , it should never have the plain 'console.error' handler.

rvykydal: Ah, it didn't work for me before https://github.com/rhinstaller/anaconda/commit/8ef0b9d508b120433cbc8c039424bbfda2cf0602

rvykydal:
I've added BZ report link to the dialog. All the layout and UI is TBD, I just wanted to demonstrate what information would be available at this point.
Action: and Hint: are added in the source code, Error: is the exception message, Report: points to the generated link (can be seen in href or the alert window.
Screenshot from 2023-07-26 11-42-07
Href does not seem to work in the dialog, click to look at the URL:
Screenshot from 2023-07-26 11-42-33

BZ form opened with the link (component, summary, details):
Screenshot from 2023-07-26 11-57-15
snip
Screenshot from 2023-07-26 11-57-30

rvykydal:
Added onCritFail to installation progress:
Screenshot from 2023-07-27 10-57-37


Previous state:
Screenshot from 2023-07-27 10-55-12

garrett:
So, a few points:

1. The reporting action shouldn't be a link, but a button.
2. We should have a way to preview the data that would be sent.
3. We probably should let someone know that this will send information over the network.
4. There should be an entry in the context of the error to report what was happening, and the text would be send over to the full report.

Something like this:

---

Fatal bug report (1)

---

Fatal bug report (2)

Clicking on the preview action would toggle the modal dialog into a preview mode, where the information is read-only, in a scrollable viewport. There would be a way to toggle back to edit to be able to correct typos.

(I'm not certain of the toggle as an action button in the modal. I'm still exploring ideas. But I wanted to share what I have so far and the reasoning behind it.)

---

Fatal bug report (4)

When sending the report, the button should have a spinner. All input fields and buttons should be in a disabled state while sending.

It would use the progress variant of the button: https://www.patternfly.org/v4/components/button#progress-indicators

---

Fatal bug report (3)

When done, there should be a confirmation prompt with a link to the bug.

---


Looking forward to your feedback!

garrett:
Also: That dark grey background ("backdrop") should be transparent; you should be able to see the screen behind the modal, similar to the design doc @ https://www.patternfly.org/v4/components/modal/design-guidelines#center-aligned-modal-default

M4rtinK:
@garrett Thanks for the mockups! :) I think both look good, but given the available time and tooling, I think we can only implement number 2 for F39. Number 1 would require automated error reporting tooling that is jut not there right not function wise right now. :P

garrett:
For the previews, we might want to consider making the modal larger (especially vertically), as there might be a lot of information, such as logs. Browsing through that in a tiny scrollable area isn't very ideal. :wink: So the mockups might not be so accurate in terms of size.

rvykydal:
@garrett thank you for the mockups!

rvykydal:
Fixed link to Bugzilla report:
Screenshot from 2023-07-28 15-37-16

rvykydal:
New update, again with layout to be refined and improved. Should demonstrate what we want (are able) to communicate in the dialog and check the UX approach:
- Detail of the error.
- The error / exception itself (should be perhaps less prominent).
- Information about log.
  - I think we will use output of journalctl -a generated by the handler.
  - I guess we want the user to review it, but maybe having another button opening a bigger window would be more appropriate.
  - Do we want the log reviewing (and even attaching it) so mandatory (or upfront) so that we risk user rather not reporting the issue? Perhaps yes because issue without log is not of much use for us. I am pretty sure wording and UI around the log attachment can be done better then in my draft.
  - The name of the log file to be attached is communicated only in the Bugzilla form, in the Details field.
  - The "hint" context property in the code of the PR does not make much sense with this version and can be removed.
Screenshot from 2023-07-31 10-15-19

Prefiled BZ report now also with log attachment mentioned:
Screenshot from 2023-07-31 10-00-59

rvykydal:
> Also: That dark grey background ("backdrop") should be transparent; you should be able to see the screen behind the modal, similar to the design doc @ https://www.patternfly.org/v4/components/modal/design-guidelines#center-aligned-modal-default

This is not a matter of Backdrop but of how we display the dialog - either dialog or the background: https://github.com/rhinstaller/anaconda/blob/e43ad74a31dcf11aec6a50622467a4cd2f15de60/ui/webui/src/components/app.jsx#L118 . I'll look at it.

ui/webui/src/components/installation/InstallationProgress.jsx
```diff
@@ -80,13 +81,16 @@ export const InstallationProgress = ({ onAddErrorNotification, idPrefix, isBootI
...
+                                refStatusMessage.current = message;
                             }
                         });
                         taskProxy.addEventListener("Failed", () => {
                             setStatus("danger");
                         });
                         taskProxy.addEventListener("Stopped", () => {
-                            taskProxy.Finish().catch(onAddErrorNotification);
+                            taskProxy.Finish().catch(onCritFail({
+                                context: cockpit.format(_("Installing the system: $0"), refStatusMessage.current),
```
KKoukiou: Why don't you use the 'message' directly here? I guess I am missing something?

ui/webui/src/components/AnacondaWizard.jsx
```diff
@@ -292,14 +294,14 @@ const Footer = ({
...
-    const goToPreviousStep = (activeStep, onBack) => {
+    const goToPreviousStep = (activeStep, onBack, errorHandler) => {
         // first reset validation state to default
         setIsFormValid(true);
         onBack();
         if (activeStep.id === "installation-review") {
             resetPartitioning().then(() => {
                 debug("resetPartitioning");
-            }, console.error);
+            }, errorHandler);
```
KKoukiou, line 304: I am not convinced this should be consiredered critical error. We should block going back in case this fails, but it's probable that the user can do something about it? Maybe try again?

ui/webui/src/components/installation/InstallationProgress.jsx
rvykydal: Message is not available here. I didn't want to use statusMessage because I would have to add it to the hook dependencies and the hook (getting tasks, registering handlers, starting the queue) would be called on every message change?

rvykydal:
@garrett I like the labels, error description and details updates - now definitely more human readable. I am going to update the PR, and also build iso so you can play with it live, especially with the flow with following BZ link opened.

Regarding the button icons and log review I have some questions and notes.

1) Iconography of button for reporting:

> Please also note the names of labels (including button labels) and use of iconography (the reprot issue does not have an icon, until submission happens, then it has a spinner and is disabled (as is quit and reboot) while submitting).


I was using the button for sending report proposed in the alternative solution by you (opening the BZ report) above. We do just open the BZ report so I am not sure we have enough information / events to be able to update button with proper icons (in progress, done), ie: the user submitted the form in bugzilla (you can see the form in the screenshot above), the user attached the log - he has to do it manually based on the instructions in the BZ form (see Note: below) and our hint in the description (see Details:, the /tmp/webui.log is the file reviewed by the user in the previous step).
attach

So here I would rather go just with the simple stateless button. This way the user can even try again if he for some reason already closed the window with the BZ report form.

2) Attaching the log is an action that happens later based on the instructions in BZ report form (generic for Fedora component) and our prefiled Details so it is even kind of optional, although we definitely want to have the logs. That is why I was thinking about having another button [Review log attachment] opening another modal for editing the file as large as we need. But maybe it is better to enforce a review (and then attaching) a bit by having it upfront.

garrett:
> I like the labels, error description and details updates - now definitely more human readable. 

:+1: 

> I am going to update the PR, and also build iso so you can play with it live, especially with the flow with following BZ link opened.

Awesome! Thank you!

> I was using the button for sending report proposed in the alternative solution by you (opening the BZ report) above.

Haha, whoops. I see that above. I didn't see it in the Penpot file when I redid it and was context shifting from dev work on Cockpit (PatternFly stuff, mainly around browser compatibility and CSS), as it was on a separate page and I was basing the newer dialogs from the page with the interactive ones. Oops!

> So here I would rather go just with the simple stateless button. 

That would only work if it has immediate actions... which I guess it does?

(That is, using `rel="noopener" target="_blank"
`, like so: <a href="https://link.to.site/to/report/a/bug?with=values&you=are+reporting" rel="noopener" target="_blank">Report it!</a>` — but this would have to be a link that is styled like a button. Otherwise, if it's a button (which it will likely be with PF React, then you need to use `window.open()` with a target and such. https://developer.mozilla.org/en-US/docs/Web/API/Window/open `window.open(site, target, windowFeatures)` where target is also `_blank` and features would include `noopener` or `noreferrer`.)

> This way the user can even try again if he for some reason already closed the window with the BZ report form.

Normally dialogs close, but I guess it makes sense to do this here. They'll need an option to reboot, which is in the modal, and if they accidentally close the window, like you're saying, then they need a way to get back. Since we don't have the (X) modal and there's nothing else to do other than report or quit/reboot, then it does make sense to treat this modal differently.

(In other words: I agree; this makes sense.)

---

I wonder if we really should use the word "Bugzilla" though. I was probably wrong suggesting that a while back. The icon probably still makes sense though, as it would open a new window.

Here's a revised mockup that changes the title, drops "Bugzilla", and has the external icon (nothing else changed):

Fatal bug report_ simple, details, larger

rvykydal:
Updated to address most of the suggestions from https://github.com/rhinstaller/anaconda/pull/4942#issuecomment-1661722007.
I (or someone who would be faster at that) still need to figure out:
- using monospace font for log and error detail content
- bigger (vertically) log textarea by default (it is just vertically resizable now).

Screenshot from 2023-08-02 13-40-30

I am going to build live iso shortly.

ui/webui/src/components/Error.jsx
```diff
@@ -16,43 +16,128 @@
...
+                        </Text>
+                    </TextContent>
+                </FormGroup>
+                <FormGroup
+                  label={_("Log")}
+                >
+                    <TextArea
+                      value={logContent}
+                      onChange={setLogContent}
+                      aria-label="review-attached-log"
```
KKoukiou, line 124: aria-label should be a normal text like "Review attached log"
In this case there is already the label just above so you just need to add an id in the TextArea, and the 'fieldId' property in the FormGroup

rvykydal:
> I am going to build live iso shortly.

Videos from running the iso (before running the installation anaconda is patched to raise exception after hitting "Erase data and install"):
https://rvykydal.fedorapeople.org/webui/error_handling/Fedora-Workstation_PR4942-2.webm
continued after logging into bugzilla:
https://rvykydal.fedorapeople.org/webui/error_handling/Fedora-Workstation_PR4942-2_continued.webm

rvykydal:
The tests are failing on page reload (https://github.com/rvykydal/anaconda/blob/ccd015f75cdb7f09464582ee9606ba566bc50721/ui/webui/test/check-basic#L140) because the commit keeping background for the error dialog (https://github.com/rhinstaller/anaconda/pull/4942/commits/f7edce12cb2f483fe6bd808b064c4e2c50d66a51). The mechanism raising the testing exception is too destructive to allow (background) page reload. I think I'll try to find something more lightweight.

rvykydal:
> @KKoukiou any advice here? I'll try to investigate why the shutdown fails in the new test but maybe I should I just remove page reload from the existing test and extend it with testing some other features / properties of the current dialog (like review of the log)?

Ok, the shutdown fail should be fixed. It was caused by using cleanup handler inappropriately in destructive test.

ui/webui/src/components/Error.jsx
```diff
@@ -30,7 +30,34 @@ import { exitGui } from "../helpers/exit.js";
...
+    reportURL.pathname = "enter_bug.cgi";
+    Object.keys(queryData).map(query => reportURL.searchParams.append(query, queryData[query]));
+    return reportURL.href;
+};
+
+const addExceptionDataToReportURL = (url, exception) => {
+    const newUrl = new URL(url);
+    newUrl.searchParams.append(
+        "short_desc",
+        "WebUI: " + exception.message
```
M4rtinK: I would suggest we also include the hint & context in the content we forward to Bugzilla - that's the best pointer to code we have & we could easily search for the hint/context string to find where the error comes from in our source code.

Also I guess some DBus error messages might be the same, but might have a different context/hint, so without those being included we might not be able to tell them apart ?

ui/webui/src/components/Error.jsx
```diff
@@ -30,7 +30,34 @@ import { exitGui } from "../helpers/exit.js";
...
+    };
+
+    const reportURL = new URL(baseURL);
+    reportURL.pathname = "enter_bug.cgi";
+    Object.keys(queryData).map(query => reportURL.searchParams.append(query, queryData[query]));
+    return reportURL.href;
+};
+
+const addExceptionDataToReportURL = (url, exception) => {
+    const newUrl = new URL(url);
```
M4rtinK, line 55: I guess we don't check for maximum URL length ? Could be something for a followup PR, possibly with corresponding policy for what to cut if we reach the maximum length.

ui/webui/src/components/AnacondaWizard.jsx
M4rtinK: I guess at least on the live media they might just run the installer again ? I think @vojtechtrefny or @poncovka might want to chime in if the state of the storage backend makes it possible to attempt with the installation run without restarting Anaconda.

ui/webui/src/components/Error.jsx
```diff
@@ -52,16 +58,23 @@ const addExceptionDataToReportURL = (url, exception) => {
...
+    const [logContent, setLogContent] = useState("");
+
+    useEffect(() => {
+        const syslogFile = cockpit.file("/tmp/syslog");
+        syslogFile.read().then(content => setLogContent(content));
+    }, []);
 
     const openBZIssue = (reportURL) => {
-        window.open(reportURL, "_blank");
+        cockpit.file("/tmp/webui.log").replace(logContent)
```
M4rtinK, line 77: On first look this looks like something that the backend would generate, but I guess that if we want to include data from the front end, this might be a better way. 

Or we might end up with a hybrid  eventually where a backend & frontend generated files get merged.

ui/webui/src/components/AnacondaWizard.jsx
rvykydal: > I am not convinced this should be consiredered critical error. We should block going back in case this fails, but it's probable that the user can do something about it? Maybe try again?

If resetPartitioning fails with exception, the installer is in undefined state IMO, so blocking going back would not help.


> I guess at least on the live media they might just run the installer again ? I think @vojtechtrefny or @poncovka might want to chime in if the state of the storage backend makes it possible to attempt with the installation run without restarting Anaconda.

Yes, and restarting the installation is a consequence of Critical Error I think. Here the hint (another context element I am removing in later commit because I didn't find a use yet, although I wasn't looking too hard) could be useful. OTOH we have just Reboot option in the dialog (and I think it is on purpose at this point, we want to avoid restarting anaconda after critical error because we are not sure if it would work) so we would also need some logic allowing just Quit.
In any case something to be handled separately as a follow-up I think.
And I can remove the patch from the set if you prefer.

ui/webui/src/components/Error.jsx
rvykydal: Maybe I should name it journal.log because in fact it just contains the content of journal. If we add some more information we can name it accordingly. Or is it better to have some specific name (like webui.log) from the beginning... not sure.

ui/webui/src/components/Error.jsx
rvykydal: Thank you, updated.

ui/webui/src/components/app.jsx
```diff
@@ -68,26 +72,32 @@ export const Application = () => {
 
             setAddress(address);
 
+            const errorContext = {
+                context: _("Initialize information about system."),
+                hint: _("Please report the error."),
```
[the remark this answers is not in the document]
rvykydal: Added  webui: translate critical error context only in the dialog

ui/webui/src/components/Error.jsx
rvykydal: Added ` webui: Add error context to Bugzilla report summary and detail `

rvykydal:
I've added commit that translates the message only in Dialog (not in BZ reprort) based on @M4rtinK's  feedback.

I think the PR would be ready for merging if approved, but one thing that must be handled before is flaky test:

https://cockpit-logs.us-east-1.linodeobjects.com/pull-4942-20230804-085733-3a98dc78-fedora-rawhide-boot/log.html

When the test is fixed, there are possible follow-up tasks to be addressed:

Tests (maybe required for this PR?):
- test that clicking on `Report issue` genrates the expected log file (/tmp/webui.log) or even
- test that editing the log preview and clicking on `Report issue` generates the expected file (/tmp/webui.log) with expected content (edited log)
- page reload check does not work: [webui: reload does not work for Critical Error dialog](https://github.com/rhinstaller/anaconda/pull/4942/commits/3a98dc788c14f457fe4747c2bc1f89fd7a702873)

Layout issues:
- using monospace font for log and error detail content
- bigger (vertically) log textarea by default (it is just vertically resizable now).

Report Detail issue:
- I'll expand on it in a separate comment

rvykydal:
> Report Detail issue:
> 
>     * I'll expand on it in a separate comment

*Recently* the Details textarea prefiled based on &comment= URL search parameter is not visible at least for me locally. As can be seen with this example:
https://bugzilla.redhat.com/enter_bug.cgi?product=Fedora&version=rawhide&component=anaconda&short_desc=SHORT_DESC&comment=COMMENT 
Screenshot from 2023-08-04 10-54-32

It *is* present in the source page:
```
    <div><textarea name="comment"
            rows="12"
          cols="80">COMMENT</textarea>
  
      <p>Expand on the Summary. Please be as specific as possible about what is
      wrong.</p>

      <p>
        <kbd>
          <font color="#990000">Bad example</font>: The program simply will not work.
        </kbd>
        <br>
        <kbd>

          <font color="#009900">Good example</font>: When I use a particular
          option to the<code>ls</code> command, I do not get the proper output
          as expected. The option I was trying was trying was <i>blah</i> and
          the exact output was <i>blah</i>.

        </kbd>
      </p>

      <p><b>Note:</b> If you need to enter large amounts of text such as logfiles,
      sample code, patches, etc. then please submit them as an attachment to the
      bug instead of placing them in the description field. You will
      be given the opportunity to make attachments after submitting the
      bug report.</p>

    </div>
  </div>

```

Strangely the URL with `&component=anaconda` removed shows the Details:
https://bugzilla.redhat.com/enter_bug.cgi?product=Fedora&version=rawhide&short_desc=SHORT_DESC2&comment=COMMENT2
Screenshot from 2023-08-04 10-54-55

rvykydal:
> > Tests (maybe required for this PR?):
> > ```
> > * test that clicking on `Report issue` genrates the expected log file (/tmp/webui.log) or even
> > ```
> 
> Added a commit [webui: add testing of log existence in critical error handling](https://github.com/rhinstaller/anaconda/pull/4942/commits/a86b13cab3feba5f5bccdde74fbf1317eb44db97) testing that (it is using sleep though, maybe there is a better waiting (testlib?) for the log.)

This test seems to be very flaky in the end. I think I'll remove it from this set.
Also moved s.udevadm_settle after the partition uuid setting.

ui/webui/src/components/Error.jsx
```diff
@@ -16,43 +16,131 @@
...
               </>
           }>
-            <TextContent>
-                <Text component={TextVariants.p}>
-                    {exception.message}
-                </Text>
-            </TextContent>
+            <Form>
+                <FormGroup fieldId="critical-error-review-attached-log" label={_("Error details")}>
+                    <TextContent>
```
KKoukiou, line 115: the fieldid from the formgroup should be used as id here.

ui/webui/src/components/Error.jsx
```diff
@@ -16,43 +16,131 @@
...
+                        <Text component={TextVariants.p}>
+                            {exception.name + ": " + exception.message}
+                        </Text>
+                    </TextContent>
+                </FormGroup>
+                <FormGroup
```
KKoukiou, line 121: Missing fieldId. This is a problem for accessibility.,

ui/webui/test/check-basic
```diff
@@ -126,20 +128,48 @@ class TestQuit(anacondalib.VirtInstallMachineCase):
...
+
+        with b.wait_timeout(50):
+            b.wait(lambda: self.file_exists(logfile))
+
+        # Reviewed log contains more than 100 lines
+        wc_output = m.execute(f"wc -l {logfile}")
+        lines = int(wc_output.partition(" ")[0])
+        self.assertTrue(lines > 100)
+
+        sleep(3)
```
KKoukiou, line 171: Pleae do above: testlib.wait(lambda: int(m.execute(f"wc -l {logfile} | cut -d" " -f1")).strip()) > 100)

KKoukiou: did not test the exact command but should be something like this

ui/webui/src/components/AnacondaWizard.jsx
M4rtinK: > About the possibility to recover from unexpected errors, I don't recommend to do that. They are bugs that need to be fixed. The recovery could lead to weirder bugs and errors that can be difficult to debug and fix.

Totally agreed & I will also add that another reason to stop when we hit a bug is to possibly prevent some irreversible issues caused by the bug. Not being able to install sucks, machine that does not boot or lost data sucks more.

ui/webui/test/check-basic
rvykydal: Used

ui/webui/src/components/Error.jsx
rvykydal: Added

ui/webui/src/components/Error.jsx
rvykydal: Fixed

ui/webui/src/components/app.jsx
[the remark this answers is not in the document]
M4rtinK: Perfect, thanks! :)

ui/webui/src/components/Error.jsx
M4rtinK: Created a Jira issue for that BTW. :)

M4rtinK (approved):
Still looks good to me. :)

KKoukiou:
Split two commits https://github.com/rhinstaller/anaconda/pull/5027 as these can already be merged. To get this a bit leaner.

KKoukiou (approved):

ui/webui/test/check-basic
```diff
@@ -139,6 +139,7 @@ class TestBasic(anacondalib.VirtInstallMachineCase):
         b.wait_visible("#critical-error-modal")
 
         # Test review of logs
+        self.addCleanup(m.execute, "rm -f /tmp/webui.log")
```
rvykydal, line 142: I didn't clean up the log file because this is a destructive test, but it may be better to clean up anyway

ui/webui/src/components/Error.jsx
```diff
@@ -80,8 +80,7 @@ export const CriticalError = ({ exception, isBootIso, reportLinkURL }) => {
     const openBZIssue = (reportURL) => {
         setPreparingReport(true);
         cockpit
-                .file("/tmp/webui.log")
-                .replace(logContent)
```
rvykydal: Aren't we losing the changes from the review with this?

ui/webui/src/components/Error.jsx
```diff
@@ -80,8 +80,7 @@ export const CriticalError = ({ exception, isBootIso, reportLinkURL }) => {
...
+                .spawn(["bash", "-c", "journalctl -a > /tmp/webui.log"])
```
rvykydal, line 83: Aren't we losing the changes from the review with this?

rvykydal: Ok, it is nondestructive now.

KKoukiou (approved):

ui/webui/src/components/Error.jsx
KKoukiou: Yes sorry

ui/webui/src/components/Error.jsx
```diff
@@ -16,55 +16,135 @@
...
           titleIconVariant="danger"
-          variant="small"
+          variant={ModalVariant.large}
           footer={
               <>
+                  {reportLinkURL &&
+                  <Button
+                    variant="primary"
+                    isLoading={preparingReport}
+                    isDisabled={preparingReport}
```
KKoukiou, line 109: You should disable this button instead of adding sleep in the test. So isDisabled={!content || preparingReport}

rvykydal: I'll open a followup PR.

PR #4944, opened 2023-07-24
Master add live iso support

jkonecny12:
Add support for local build of Live ISO with the latest packages.

In case the `anaconda-webui` package is part of the build (by default it is) it will run web UI automatically.

This solution is based on container running QEMU. Virtual Machine will get 80% of your computation resources to run Anaconda and build the ISO.

The kickstart file for creating the Live will be taken from `fedora-kickstarts` upstream project, flattened, and stored to the container. This will save time and could gate some issues.


jkonecny12:
**UPDATED:**
Removed trap from the HTTP server. It was incorrectly used and the container will kill everything at the end anyway so we don't have to care about this.

CONTRIBUTING.rst
```diff
@@ -130,11 +130,18 @@ Build Anaconda RPM files with our container::
...
+  sudo make -f ./Makefile.am anaconda-iso-creator-build # to build the container if it doesn't exists already
+  sudo make -f ./Makefile.am container-iso-build
 
 or for Web UI image run::
 
-  make -f ./Makefile.am container-webui-iso-build
+  sudo make -f ./Makefile.am anaconda-iso-creator-build # to build the container if it doesn't exists already
+  sudo make -f ./Makefile.am container-webui-iso-build
+
+ro for Web UI in Live media run::
```
vojtechtrefny, line 141: Typo: **or** for Web UI in Live media run

rvykydal (approved):

VladimirSlavik (approved):
Mostly LGTM, thank you.
- I don't know if it makes any sense to get the kickstarts in container build. Perhaps doing that in the script might be easier to debug? The amount of data you save is negligible.
- Installing `lorax-lmc-virt` gives me almost a gigabyte of files, but there's a lot of things for architectures we won't ever use. Is it possible to trim this down to something more manageable? (I guess that's rather for @bcl?)

bcl:
> * Installing `lorax-lmc-virt` gives me almost a gigabyte of files, but there's a lot of things for architectures we won't ever use. Is it possible to trim this down to something more manageable? (I guess that's rather for @bcl?)

There's really nothing that can be done there. All it does is install qemu, qemu-kvm and edk2-ovmf, so any extra dependencies are coming from there.

VladimirSlavik:
Thanks for confirming! I'm starting to get the _why_ of lorax. If only it could build containers...

dockerfile/anaconda-live-iso-creator/Dockerfile
```diff
@@ -0,0 +1,61 @@
...
+  dnf install -y \
+# required for ksflatten
+  pykickstart; \
+  mkdir /fedora-kickstarts; \
+  git clone https://pagure.io/fedora-kickstarts.git /fedora-kickstarts; \
+  cd /fedora-kickstarts; \
+  ksflatten -o /lorax/workstation.ks -c fedora-live-workstation.ks; \
+# cleanups of parts which don't need to be part of the container
+  dnf remove -y pykickstart; dnf clean all; \
+  rm -rf /fedora-kickstarts
```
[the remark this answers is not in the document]
VladimirSlavik: It's probably ok, time will tell :)

PR #4945, opened 2023-07-24
webui: rename custom mountpoint assignment step to mountpoint mapper

KKoukiou:
This affect the wizard step title, component name, selectors, ids etc.


KKoukiou:
> From storage point of view, I'd prefer something like _mount point mapping_ instead of _partition mapping_ -- users will assign mount points to different types of devices, not necessarily only to partitions so it might be confusing.

@garrett ^^

garrett:
Oh, that makes sense, especially since we changed it to be mount point focused instead of being partition focused... Thanks! :+1:

VladimirSlavik (commented):
Code-wise LGTM, can't comment on correctness of the renaming since I did not keep up with this.

PR #4948, opened 2023-07-25
webui: Prevent the Anaconda window from being closed by keyboard shortcuts

adamkankovsky:
**The changes in this PR do not prevent closing the application window, but only add a confirmation window before quitting.**
In most browsers this behavior is prevented because it could be used to attack https://stackoverflow.com/questions/30781070/prevent-onbeforeunload-to-close-page-in-any-case. 
Therefore, this seems to me to be the only possible solution at the moment.

### DIFF
--- a/ui/webui/src/components/app.jsx
+++ b/ui/webui/src/components/app.jsx
@@ -55,6 +55,8 @@ export const Application = () => {
     const [storeInitilized, setStoreInitialized] = useState(false);
 
     useEffect(() => {
+        // Before unload ask the user for verification
+        window.onbeforeunload = e => "";
         cockpit.file("/run/anaconda/bus.address").watch(address => {
             setCriticalError();
             const clients = [
--- a/ui/webui/test/check-language
+++ b/ui/webui/test/check-language
@@ -112,10 +112,6 @@ class TestLanguage(anacondalib.VirtInstallMachineCase):
         l.check_selected_locale("he_IL")
         b.wait_attr("html", "dir", "rtl")
 
-        # Language direction should persist after reload
-        b.reload()
-        b.wait_attr("html", "dir", "rtl")
-
         l.select_locale("de_DE")
         l.check_selected_locale("de_DE")
         b.wait_attr("html", "dir", "ltr")


ui/webui/src/components/app.jsx
```diff
@@ -55,6 +55,8 @@ export const Application = () => {
     const [storeInitilized, setStoreInitialized] = useState(false);
 
     useEffect(() => {
+        // Before unload ask the user for verification
+        window.onbeforeunload = function (e) { return "" };
```
VladimirSlavik, line 59: Just out of curiosity, can't you do something like this?
```
window.onbeforeunload = e => "";
```
While I personally like the keyword variant more, I guess consistency would be more valuable...

adamkankovsky: Yes, it can be written like that too

KKoukiou (commented):
This gives the default dialog from firefox.

@garrett can we keep this or we need a custom dialog? 

Also @adamkankovsky can we please have a test for this?

garrett:
> @garrett can we keep this or we need a custom dialog?

Unfortunately, I think we're stuck with this.

You used to be able to customize it, but all browsers removed support for this a few years ago: https://stackoverflow.com/questions/38879742/is-it-possible-to-display-a-custom-message-in-the-beforeunload-popup

> The changes in this PR do not prevent closing the application window, but only add a confirmation window before quitting.

We want it specifically for closing the application window, including Ctrl-W, Ctrl-Q, and the (X) icon in the top-right. Bummer.

We _do not_ want this prompt when quitting from the UI. That can and should be handled in-page with a PatternFly modal.

So does that mean this PR doesn't do what we want it to?

garrett:
There are apparently a bunch of custom settings one can toggle in Firefox to adjust prompting on closing: https://www.reddit.com/r/firefox/comments/si9in4/comment/hv8hd7v/?context=3 — I can add some of these to the Firefox theme's `user.js`. I wonder if any of those would do what we want. (Time to experiment...)

garrett:
@KKoukiou What did you do to  get this dialog?

KKoukiou:
> @KKoukiou What did you do to get this dialog?

Ctrl + w

garrett:
I tried closing a Firefox window with just one tab... Penpot... and got this modal:

image

So this _should_ theoretically work for Anaconda too, for the (X) icon.

(This is Firefox; I'm just using a custom theme that makes it look more like GNOME.)

adamkankovsky:
@garrett Thanks to the fact that it is done via window.onbeforeunload which captures all window closures from the user side. I've tried it and even on the x button it works.

garrett:
> captures all window closures from the user side. I've tried it and even on the x button it works.

:tada: Excellent to hear!

garrett:
Hi. Any reason this hasn't been merged yet? If there's anything that needs to be done still, can we merge this and split whatever it might be out to a follow-up? Thanks!

KKoukiou:
> Hi. Any reason this hasn't been merged yet? If there's anything that needs to be done still, can we merge this and split whatever it might be out to a follow-up? Thanks!

so I havent managed to make the required changes for https://github.com/rhinstaller/anaconda/pull/4948#issuecomment-1668080941 tests to work

You can remove the tests - merge it and create a new PR - mark it as blocked on my cockpit PR

KKoukiou (approved):

KKoukiou (changes requested):
These need to be squashed

PR #4949, opened 2023-07-25
infra: Fix permissions for Permian GitHub reporter

velezd:
This should fix Permian not being able to report results of webui e2e tests to PR


VladimirSlavik (approved):
Thanks!

VladimirSlavik:
I think this is small enough to just try as-is.

PR #4950, opened 2023-07-25
webui: Disable minification     

M4rtinK:
Disable minification when we build our JS code, so that the code we ship in our RPMs can be debugged.

This should improve the debugging of runtime issues on any media built from RPM packages as well as for any cases where we apply updates images built from RPM packages.

### DIFF
--- a/ui/webui/build.js
+++ b/ui/webui/build.js
@@ -63,7 +63,7 @@ const context = await esbuild.context({
         ".js": "jsx",
         ".py": "text",
     },
-    minify: production,
+    minify: false,
     nodePaths,
     outdir,
     target: ['es2020'],


M4rtinK:
Same with (PR generated) boot ISO:
boot_iso_no_minification

M4rtinK:
Local development / `make rsync` also seems to be working fine with this change. :)

KKoukiou (commented):
I think instead of passing production=false you can write `minify=false` 
https://esbuild.github.io/api/#minify

Also it would be  noteworthy to mention in the PR the size diff of the RPM (or just the dist/* directory) before and after this change.

KKoukiou (commented):
Ok I think you can disable minification for production in general. But keep the production flag on otherwise? What do you think? 

The size difference is 200KB, I consider it acceptable.
``` diff
[kkoukiou@sequioa webui]$ make
make[1]: Entering directory '/home/kkoukiou/repos/anaconda/ui/webui'
NODE_ENV=production ./build.js
▲ [WARNING] Using direct eval with a bundler is not recommended and may cause problems [direct-eval]

    src/components/localization/InstallationLanguage.jsx:225:40:
      225 │                                         eval(body);
          ╵                                         ~~~~

  You can read more about direct eval and bundling here: https://esbuild.github.io/link/direct-eval

14:22:42: Build finished
make[1]: Leaving directory '/home/kkoukiou/repos/anaconda/ui/webui'
[kkoukiou@sequioa webui]$ du -sh dist/
640K	dist/
[kkoukiou@sequioa webui]$ vim build.js 
[kkoukiou@sequioa webui]$ git diff
diff --git ui/webui/build.js ui/webui/build.js
index 4930072a9b..7d44114c08 100755
--- ui/webui/build.js
+++ ui/webui/build.js
@@ -63,7 +63,7 @@ const context = await esbuild.context({
         ".js": "jsx",
         ".py": "text",
     },
-    minify: production,
+    minify: false,
     nodePaths,
     outdir,
     target: ['es2020'],
[kkoukiou@sequioa webui]$ rm -rf dist/
[kkoukiou@sequioa webui]$ make
make[1]: Entering directory '/home/kkoukiou/repos/anaconda/ui/webui'
NODE_ENV=production ./build.js
▲ [WARNING] Using direct eval with a bundler is not recommended and may cause problems [direct-eval]

    src/components/localization/InstallationLanguage.jsx:225:40:
      225 │                                         eval(body);
          ╵                                         ~~~~

  You can read more about direct eval and bundling here: https://esbuild.github.io/link/direct-eval

14:23:25: Build finished
make[1]: Leaving directory '/home/kkoukiou/repos/anaconda/ui/webui'
[kkoukiou@sequioa webui]$ du -sh dist/
820K	dist/

```

M4rtinK:
@KKoukiou So for now I've just opted for a minimal change - keep minification ON, but make sure source maps are generated & shipped in the RPM. This seems to fix the main issue - eq. being able to debug Anaconda JS code installed from RPMs. 

This is about +600 kB at RPM level for the source map data, some more once installed (looks like the source maps are not compressed for some reason, unlike all the other files).

And if the source maps are not fetched unless the user is debugging the over-the-wire seize will still be small for remotely controlled installations.

If we find some scenario where the source maps don't work - we can always also turn off minification.

M4rtinK:
Hmm, the source maps have +4.2 MB of installed size (they are not compressed for some reason) - lets try disabling minification + source map generation next.

M4rtinK:
So back to just disabling minification - hopefully the last iteration for now. :)

KKoukiou (approved):
The size diff (200KB) is negligible, I am ok with this change for now at least.

jelly:
> Hmm, the source maps have +4.2 MB of installed size (they are not compressed for some reason) - lets try disabling minification + source map generation next.

Sourcemaps should not be loaded by a browser by default. Only when opening Developers tools so it's quite harmless.

KKoukiou:
> > Hmm, the source maps have +4.2 MB of installed size (they are not compressed for some reason) - lets try disabling minification + source map generation next.
> 
> Sourcemaps should not be loaded by a browser by default. Only when opening Developers tools so it's quite harmless.

We care about the RPM size increase as all this will go even on the tiny boot.iso

PR #4951, opened 2023-07-25
Add a sub-module manager

VladimirSlavik:
This makes it easier to keep track of modules, even if very simple.

Funnily enough, the manager itself needs no tests, it gets almost full coverage just by being used.

----

~~Draft, because I'm not 100% sure about the storage module changes. Exchanging variables for strings seems to relax the correctness.~~

### DIFF
--- a/pyanaconda/modules/common/submodule_manager.py
+++ b/pyanaconda/modules/common/submodule_manager.py
@@ -55,3 +55,6 @@ class SubmoduleManager:
             requirements.extend(module.collect_requirements())
 
         return requirements
+
+    def __iter__(self):
+        return iter(self._modules)
--- a/pyanaconda/modules/common/submodule_manager.py
+++ b/pyanaconda/modules/common/submodule_manager.py
@@ -1,0 +1,57 @@
+#
+# Submodule manager for kickstart services.
+#
+# Copyright (C) 2023 Red Hat, Inc.
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# the GNU General Public License v.2, or (at your option) any later version.
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY expressed or implied, including the implied warranties of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General
+# Public License for more details.  You should have received a copy of the
+# GNU General Public License along with this program; if not, write to the
+# Free Software Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
+# 02110-1301, USA.  Any Red Hat trademarks that are incorporated in the
+# source code or documentation are not subject to the GNU General Public
+# License and may only be used or replicated with the express permission of
+# Red Hat, Inc.
+#
+__all__ = ["SubmoduleManager"]
+
+class SubmoduleManager:
+    """Kickstart sub-module manager.
+
+    This class helps manage multiple kickstart service sub-modules.
+    """
+
+    def __init__(self):
+        self._modules = []
+
+    def add_module(self, module):
+        """Add a module."""
+        self._modules.append(module)
+
+    def publish_modules(self):
+        """Publish the modules."""
+        for module in self._modules:
+            module.publish()
+
+    def process_kickstart(self, data):
+        """Process the kickstart data in all modules."""
+        for module in self._modules:
+            module.process_kickstart(data)
+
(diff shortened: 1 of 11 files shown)


pyanaconda/modules/storage/storage.py
```diff
@@ -263,10 +235,10 @@ def scan_devices_with_task(self):
         storage = self.storage.copy()
 
         # Set up the storage.
-        storage.ignored_disks = self._disk_selection_module.ignored_disks
-        storage.exclusive_disks = self._disk_selection_module.exclusive_disks
-        storage.protected_devices = self._disk_selection_module.protected_devices
-        storage.disk_images = self._disk_selection_module.disk_images
+        storage.ignored_disks = self._modules["disk_selection"].ignored_disks
```
poncovka, line 238: I am worried that using a dictionary to get a module means that pylint or IDE won't be able to recognize incorrect attributes or hint the right ones. Would it be possible to use the manager just for managing the list of modules, but keep the instances for the direct access? You could drop the names in that case.

pyanaconda/modules/storage/storage.py
```diff
@@ -71,87 +72,60 @@ def __init__(self):
...
+        self._modules.add_module("fcoe", FCOEModule())
+        self._modules.add_module("iscsi", ISCSIModule())
+        self._modules.add_module("nvdimm", NVDIMMModule())
+        self._modules.add_module("dasd", DASDModule())
+        self._modules.add_module("zfcp", ZFCPModule())
 
         # Connect modules to signals.
         self.storage_changed.connect(
-            self._device_tree_module.on_storage_changed
+            self._modules["device_tree"].on_storage_changed
```
poncovka, line 90: Btw. I planned to introduce a new class that would define the `on_storage_changed` method to replace most of these statements with a single loop.

pyanaconda/modules/common/submodule_manager.py
```diff
@@ -0,0 +1,63 @@
...
+    """
+
+    def __init__(self):
+        self._modules = OrderedDict()
+
+    def add_module(self, name, module):
+        """Add a module."""
+        self._modules[name] = module
+
+    def __getitem__(self, item):
```
poncovka, line 37: It would be great if you could also iterate over the list of modules.

poncovka (commented):
It is a great improvement. Thank you! :-) I would just reconsider the pros and cons of the direct access.

pyanaconda/modules/storage/storage.py
VladimirSlavik: Yes, I think that's the better way to do this, too. About the names, they make some sense, but perhaps let's not treat them as a "primary key", but rather property of the managed module(s). Then, everything still has a name.

pyanaconda/modules/common/submodule_manager.py
VladimirSlavik: Yes, I considered that too. Ultimately, I think that's something to add only when needed. After all, `__getitem__` is there only to support the getting, I would just dump it otherwise. Also, having access to the underlying list opens the can of worms that is direct access to the "managed" list. I would prefer keeping this as simple as possible, until needed.

In other words, if there are some places where having an iterable like that makes sense, feel free to share, that's a data-based reason to have it.

VladimirSlavik: I now realize the other comment above might be related to this. IMO the point still stands - let's add this once needed.

pyanaconda/modules/storage/storage.py
VladimirSlavik: Ah, yes, that makes sense. How about:

1. Define another mixin `StorageSubscripter` with abstract method `on_storage_changed`
2. Add this as another parent to all the submodules of interest.
3. Loop over all submodules, filter by the mixin class, connect the signals.

VladimirSlavik: Actually, no, that would not be a mixin but a regular ancestor class. I got confused after looking at `DeviceTreeModule` with its rich ancestry.

VladimirSlavik: Never mind, pushed here, too.

VladimirSlavik:
The `raid-1` ks test is disabled due to some problems, so it's ok that it failed.

pyanaconda/modules/storage/storage.py
poncovka: I think the mixin would make a lot of sense. Btw. I don't know why `DeviceTreeModule` isn't just `BaseModule`.

poncovka (approved):
Thank you!

PR #4952, opened 2023-07-25
Update pixel tests broken from localization

jkonecny12:
The back button is now localized so we need to update the tests.


KKoukiou (commented):
I am ok with the change but the commit message is wrong. It's the opposite happening. The back button used to be localized and now it's not.

KKoukiou (approved):

KKoukiou (commented):
Actually pls fix the commit message - it's confusing.

KKoukiou (approved):

jkonecny12:
Ahh, ok. Still getting used to the pixel tests, I understood it incorrectly.

PR #4953, opened 2023-07-26
webui: Quit button does not work on the Live image

adamkankovsky:


KKoukiou (commented):
This needs some explanation. ;)

KKoukiou (commented):
I will ask @martinpitt if he can take a look into this one. As I am a bit noob with the best practices on bash scripting.

ui/webui/webui-desktop
```diff
@@ -86,4 +86,9 @@ else
     sleep 3
 fi
 
-HOME="$BROWSER_HOME" $BROWSER http://127.0.0.90"$URL_PATH"
+trap 'kill $B_PID; exit 0' INT TERM
```
martinpitt, line 89: The signal list is incomplete; you should at least also terminate on `EXIT`, but adding `QUIT` and `PIPE` can't hurt either.

Also, please drop the `exit 0` -- running into that trap *already* means that your script is exiting.

martinpitt (changes requested):
Certainly a little convoluted, but if the Quit button emits SIGTERM or so, then that's an okay way to kill the browser with it.

ui/webui/webui-desktop
```diff
@@ -86,4 +86,9 @@ else
     sleep 3
 fi
 
-HOME="$BROWSER_HOME" $BROWSER http://127.0.0.90"$URL_PATH"
+trap 'kill $B_PID' INT TERM EXIT
```
martinpitt, line 89: Oh, there's already a trap in line 76. I'm not actually sure if they are additive, or the second one would replace the first (my hunch is the latter?) Can you just move this kill into line 76 then?

adamkankovsky: So I finally modified it to just one trap and created a cleanup function to make the code clearer.

martinpitt (approved):
Looks good now, cheers!

PR #4954, opened 2023-07-26
webui: debounce password quality checks to be done only once per 300ms

KKoukiou:
Otherwise this call is performed for each user key stroke, resulting in a pretty inefficient password check.

Will fix the following flake: https://cockpit-logs.us-east-1.linodeobjects.com/pull-4946-20230725-160918-c9d5fa06-fedora-rawhide-boot/log.html#11


rvykydal (approved):
Thank you.

VladimirSlavik (commented):
Thank you, also note that GTK GUI does the same for a number of things, so there will be more of this.

PR #4958, opened 2023-07-27
Revert 64k ARM switcher

jkonecny12:
Revert of https://github.com/rhinstaller/anaconda/pull/4858 .

Unfortunately, this changeset is breaking composes now by https://github.com/rhinstaller/anaconda/pull/4858#issuecomment-1652180161 .

The issue is that this changeset was written for RHEL-9 branch but RHEL-9 don't have modularized payload. So all the calls to the payload (or most of them) are wrong. We need bigger changes to this feature before we can get it back.

We can take a look on this after the deadlines we have now. This code is not a priority for now.

### DIFF
--- a/data/anaconda.conf
+++ b/data/anaconda.conf
@@ -302,9 +302,6 @@ password_policies =
     user (quality 1, length 6, empty)
     luks (quality 1, length 6)
 
-# Should kernel options be shown in the software selection spoke?
-show_kernel_options = True
-
 [License]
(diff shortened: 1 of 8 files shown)


VladimirSlavik (approved):
Thank you, let's do this.

VladimirSlavik:
The `hmc` test is manual-only, so the "payload" run passed.

VladimirSlavik:
Manual testing seems ok.

PR #4962, opened 2023-07-27
webui: minor code cleanups

KKoukiou:


vojtechtrefny (approved):

VladimirSlavik (approved):
Thanks! Not sure about some of the better names, but you know that better.

KKoukiou:
> Thanks! Not sure about some of the better names, but you know that better.

open for suggestions! Which one you don't like?

M4rtinK (approved):
Looks good to me. :)

rvykydal (approved):
Looks good to me. Given we already use "Scenario" term it is way more consistent now.

(I wouldn't try to find something better at this point. Now it maybe close to "Method" but in the future we might really get to implementing guided partitioning scenarios? (or how to call it), like "Reuse home", "Install along existing system"...

PR #4963, opened 2023-07-27
webui: unpack GetDiskFreeSpace and GetDiskTotalSpace from array

skobyda:
GetDiskFreeSpace and GetDiskTotalSpace returns an array with an actual value at first index.
We then use this value to check if there is enough free space, e.g.: "diskFreeSpace < requiredSize". While it may not have caused any issues until now, since in JS comparison of array and value (e.g. [99] < 100) still evaluates correctly. But this is just an invitation for bugs in the future and needs to be fixed nevertheless

### DIFF
--- a/ui/webui/src/actions/storage-actions.js
+++ b/ui/webui/src/actions/storage-actions.js
@@ -47,15 +47,15 @@ export const getDeviceDataAction = ({ device }) => {
                 .then(free => {
                     // Since the getDeviceData returns an object with variants as values,
                     // extend it with variants to keep the format consistent
-                    devData.free = cockpit.variant(String, free[0]);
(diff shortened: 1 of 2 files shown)


KKoukiou (approved):
Javascript is weird. thanks for catching that!

KKoukiou (commented):
There is something off here with the tests - taking a look

KKoukiou (changes requested):
You also need to adjust src/actions/storage-actions.js:

devData.free = cockpit.variant(String, free[0])
devData.total = cockpit.variant(String, total[0]);

skobyda:
Fixed
And also applied the same change to getFormatData API to be consistent

PR #4966, opened 2023-07-27
Pin translations by SHA instead of always grabbing latest

VladimirSlavik:
Let's have controlled translation updates.

Plan:

- [x] Use a variable in a separate file (Makefile include) to hold the commit-ish value
- [x] Use this variable instead of using the branch
- [x] Create a Makefile target to update this value
- [x] Create a GH action/workflow to bump the variable to SHA of the l10n branch HEAD and make a PR with it


VladimirSlavik:
Side note - getting one arbitrary commit from a remote repo is a funny dance. Onwards to the action...

VladimirSlavik:
So, as it stands now:
- The first two commits prepare the works, but change nothing. Everything is still ungated.
- The next commit adds the update magic and the one after that actually locks the translation to a specific version. Gating is on.
- The next two commits add an action/workflow that does this for us automatically.

VladimirSlavik:
Confirmed to work:
- first run that created PR https://github.com/VladimirSlavik/anaconda/actions/runs/5693307080/job/15432116381
- PR it made https://github.com/VladimirSlavik/anaconda/pull/106
- second run that correctly saw everything up to date https://github.com/VladimirSlavik/anaconda/actions/runs/5693369743/job/15432325244

VladimirSlavik:
As an aside, this lets us also delete folders for older Fedoras, since old branches (starting now) will keep the pinned sha where the folders existed. The commits themselves will not go away, since they are all part of the single branch in the l10n repo.

KKoukiou (approved):

jkonecny12 (commented):
Thanks for implementing this!

